Issue #10890, opened 2021-08-23
Add `list` to `jupyter lab workspace` to see all workspaces


baggiponte:
### Problem

Switching from one workspace to another is not super easy unless you save a `.jupyterlab-workspace` within the project directory. Problem: sometimes said file is exactly in the directory you want to open.

Opening workspaces is straightforward, as outlined in the [docs](https://jupyterlab.readthedocs.io/en/stable/user/urls.html#managing-workspaces-ui). However, sometimes one might not exactly recall the workspace name - I usually rely on `jq` to list the number of workspaces I have.

### Proposed Solution

Add the CLI subcommand `jupyter lab workspaces list` to see all workspaces (and location?). In this way, one can simply copy-paste the name inside the URL bar.

karthikayan4u:
Hello @agoose77 , I wish to work on this issue. Should I be assigned for this issue to work on the same? I am a complete beginner to open source.

ariso-sac:
Hi Team !!
I would like to work on this issue.
Could you help me with code architecture and place where this shell commands are placed ?

agoose77:
Hi all - 
JupyterLab is split into a "front end" in TypeScript, and a "back end" in Python. The Python-side uses the `traitlets` library to implement configurable, typed, reactive objects. Traitlets is *also* used to implement (through layers of inheritance) a command-line API. The classes that implement the workspace commands are in `jupyterlab/labapp.py`, e.g. the `LabWorkspaceExportApp`
https://github.com/jupyterlab/jupyterlab/blob/05c2482bd9cb85cd134dff4bcb047a3c7d5e2326/jupyterlab/labapp.py#L261-L294


Information about how `traitlets` is used to implement commandline programs can be found [in its documentation](https://traitlets.readthedocs.io/en/stable/config.html#the-main-concepts)

Now, the actual workspace reading logic for the JupyterLab API (which the frontend talks to over HTTP) is in a separate repo: 
https://github.com/jupyterlab/jupyterlab_server/blob/900ee8a4547749cd46fcc5d1649c1f5b9da675e5/jupyterlab_server/workspaces_handler.py#L134-L139

This calls `_list_workspaces` - https://github.com/jupyterlab/jupyterlab_server/blob/master/jupyterlab_server/workspaces_handler.py#L21-L46

[...]

PR #10923, opened 2021-08-25
Feature: Added a CLI subcommand 'list' to list all the workspaces

karthikayan4u:
## References

This PR is with respect to the issue [#10890](https://github.com/jupyterlab/jupyterlab/issues/10890).
No other PRs available for the same.

## Code changes

Changes are made to [jupyterlab/labapp.py](https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/labapp.py) as below,
1). Created a class  **LabWorkspaceListApp** within which all the workspaces are collected and displayed with the help of helper functions below.
2). Added a helper function **_list_available_workspaces** which returns a list comprising of all available workspaces.
3). Added another helper function **_load_with_file_times** which will return the respective workspace's **json** from the disk with its stats overwritten accordingly.
4). Made changes to class **LabWorkspaceApp** in such a way that it can handle and call the above class when **list** subcommand is used.

### DIFF
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -314,7 +314,8 @@ class LabWorkspaceListApp(JupyterApp):
             for workspace in workspaces:
                 print(json.dumps(workspace))
         else:
-            print(*items, sep="\n")
+            for workspace in workspaces:
+                print(workspace['metadata']['id'])
                 
 
 def _list_available_workspaces(directory, items):
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -258,6 +258,28 @@ class LabPathApp(JupyterApp):
         print('Workspaces directory: %s' % get_workspaces_dir())
 
 
+class LabWorkspaceListApp(JupyterApp):
+    version = version
+    description = """
+    Print name of all the workspaces available alongside their respective path
+    """
+    def start(self):
+        app = LabApp(config=self.config)
+        directory = app.workspaces_dir
+        app_url = app.app_url
+
+        items = [item
+                for item in os.listdir(directory)
+                if item.endswith(WORKSPACE_EXTENSION)]
+        items.sort()
+
+        for slug in items:
+            workspace_path = os.path.join(directory, slug)
+            if os.path.exists(workspace_path):
+                print(slug, workspace_path, json.dumps(dict(data=dict(), metadata=dict(id=app_url))))
+                
+
+
 class LabWorkspaceExportApp(JupyterApp):
     version = version
     description = """
@@ -396,9 +418,9 @@ class LabWorkspaceImportApp(JupyterApp):
 class LabWorkspaceApp(JupyterApp):
     version = version
     description = """
-    Import or export a JupyterLab workspace
+    Import or export a JupyterLab workspace or list all the JupyterLab workspaces
 
-    There are two sub-commands for export or import of workspaces. This app
+    There are three sub-commands for export, import or list of workspaces. This app
         should not otherwise do any work.
     """
     subcommands = dict()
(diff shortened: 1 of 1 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/introduce-yourself/17/265

baggiponte:
Not the best place to add this here (I should just add this below #10890) but it helps with my last reply.

How about adding a "metadata.projectRoot" element to the `jupyterlab-workspace`? The value would be the directory where the workspace file is saved to. Then this element might be fetched by `jupyterlab workspace list` as well.

agoose77:
> Not the best place to add this here (I should just add this below #10890) but it helps with my last reply.
> 
> How about adding a "metadata.projectRoot" element to the `*.jupyterlab-workspace`? The value would be the directory where the workspace file is saved to. Then this element might be fetched by `jupyterlab workspace list` as well.

It sounds like you want to customize where you store workspaces. You can do this with `jupyter lab --LabServerApp.workspaces_dir=...`. 
Otherwise, I'm not quite sure what you're asking. In any case, I think it's probably out of scope for this command.

@krassowski what do you think of this PR?

baggiponte:
> It sounds like you want to customize where you store workspaces. You can do this with `jupyter lab --LabServerApp.workspaces_dir=...`.
> Otherwise, I'm not quite sure what you're asking. In any case, I think it's probably out of scope for this command.

Sorry if I was not being clear, I was further elaborating on this point:

> @baggiponte what did you mean by "location"? Did you mean location on disk, or the URL?

I am not thinking of the directory where workspaces are stored. If I understood correctly, workspaces are stored in $JUPYTER_CONFIG_DIR/lab/workspaces and we can also save them in our project directory, as seen below.

image

 What I was referring to was storing the project directory information in the workspace file, so that we are brought in the right directory when workspaces are opeedn. In other words, the lab navigates to the right url where the project is - like with vscode workspaces and Rstudio Rproj files.

This cannot be done with this command - but would eventually be some useful info to display:

```bash
$ jupyter lab workspaces list
my-awesome-proj   /path/to/my-awesome-proj
```

agoose77:
OK, I understand what you're looking to achieve. Workspaces are more like "sessions" that you can distinguish from one another. As such, they record the last directory that was active in the file manager. Is this behaviour not sufficient?

jupyterlab/labapp.py
```diff
@@ -258,6 +260,91 @@ def start(self):
...
+        directory = app.workspaces_dir
+
+        items = [item
+                for item in os.listdir(directory)
+                if item.endswith(WORKSPACE_EXTENSION)]
+        items.sort()
+
+        workspaces = _list_available_workspaces(directory, items)
+        if self.jsonlist:
+            print(json.dumps(workspaces))
```
bollwyvl, line 312: would want to use `encoding='utf-8'`, probably

jupyterlab/labapp.py
```diff
@@ -258,6 +260,91 @@ def start(self):
...
+            {"LabWorkspaceListApp": {"jsonlist": True}},
+            _i18n("Produce machine-readable JSON list output."),
+        ),
+        json=(
+            {"LabWorkspaceListApp": {"json": True}},
+            _i18n("Produce machine-readable JSON object on each line of output."),
+        ),
+    )
+
+    jsonlist = Bool(
```
bollwyvl, line 283: perhaps `jsonl` or `jsonlines`, and reference https://jsonlines.org so the format is _slightly_ more specified

bollwyvl:
On the main, my gut says we should be trying to add _less_ in `labapp`. And indeed, as this primarily works with primitives from `jupyterlab_server`, we may want to move 90% of this implementation over there and/or encapsulate the opinions (e.g. encoding, sorting, etc) as a `WorkspaceManager ` over there. This would _still_ require a PR here, once that got merge usptream, as was needed when we added the licenses CLI/API.

jupyterlab/labapp.py
```diff
@@ -14,6 +14,8 @@
 from jupyter_server._version import version_info as jpserver_version_info
 from jupyter_server.serverapp import flags
 from jupyter_server.utils import url_path_join as ujoin
+from jupyter_server import _tz as tz
+from jupyter_server.transutils import _i18n
```
fcollonval, line 18: JupyterLab is not using the `jupyter_server` translation tool. Could you remove this?

fcollonval (changes requested):

krassowski:
As a general +1 for having this in: it would have helped to debug #11265 a lot. @karthikayan4u would you like to give it one more push to get it through the finish line? Do you need any help in addressing the review?

jupyterlab/labapp.py
fcollonval: It does not seem to be supported: https://docs.python.org/3/library/json.html#json.dumps

PR #10796, opened 2021-08-11
Add Galata in JupyterLab

fcollonval:
## References

Fixes https://github.com/jupyterlab/galata/issues/67
Fixes #10480

Features:
- mock state (user can set an initial state)
- mock settings (user can set initial values)
- create unique folder for each test on the server
- automatically go to JupyterLab page, hook Galata Inpage code and wait for the app to be started
- Stop terminals and sessions created during a test
- Playwright HTML report:

playwright_html_report

## Code changes

:warning: @types/node was frozen to ^14.6.1 in `package.json` to resolve typedoc [error related to jest types](https://github.com/facebook/jest/issues/11640)

This imports galata as a first party package in JupyterLab core to ensure the helpers are in sync with JupyterLab packages.

This also convert galata from a tool to a set of fixtures and helpers on top of Playwright Test Runner - this will allow to catch up quickly all the nice features already and coming into Playwright.

It changes also the logic to fit test framework aka isolation of `test` functions to be able to select a file or a function only as proposed by Playwright Test Runner. To allow that the state and the settings are mocked in memory to avoid changes on the hard drive. The sessions and terminals created during a test are automatically deleted. A unique temporary path can be created for a test to be use on the server to ensure files are not shared between tests.

> ~~For now the HTML report has been removed, it will be restored at the next Playwright release that will include a [default HTML report](https://github.com/microsoft/playwright/pull/7994).~~ HTML report is available in 1.14.0. This turn that feature on (although it is not documented - so probably not considered as stable; I encountered crash by time to time).

> ~~For now using multiple workers does not work (I did not get why but when using multiple workers some tests are even not started?!).~~ It seems than when using multiple workers, if a test failed it will result in additional test failures.

Tasks:

- [x] Fix https://github.com/jupyterlab/galata/issues/69 - add a fixture for base app path.
- [x] Migrate `ui-tests`
- [x] Update documentation

## User-facing changes

Packages using galata will need to update their tests to use Playwright Test Runner directly

## Backwards-incompatible changes

Galata is no more a tool but a set of helpers and fixtures for Playwright Test Runner.

PR #10929, opened 2021-08-26
Upgrade `jlpm` cmd to yarn 3.x

telamonian:
I followed the migration guide, and suprisingly it mostly seems to work. Needs way more testing/poking at


jtpio:
Thanks for starting this!

Looks like there a couple of conflicts to fix for the CI to be triggered.

telamonian:
up to 42/63 checks passing as of 4388e02. Woo!

blink1073:
Needs another rebase from #10796

telamonian:
> Needs another rebase from #10796

that's gonna happen a lot. Both `package.json` and `yarn.lock` underwent large scale formatting changes, so any other PR that touches either of those is going to trigger the need for yet another rebase (YAR?)

blink1073:
YAR indeed.  Happy to merge this sooner than later to avoid the churn

telamonian:
@blink1073 There's still a couple of chewy issues that probably boil down to pathing but aren't obvious how to fix:

1. win and mac tests fail at first run of `ensure-buildutils.js` (which happens as part of `pip install`), while linux tests run `ensure-buildutils.js` just fine
  - specifically, the first call to `npm run build` in the `buildutils` pkg (where the relevant script is `"build": "tsc"`) fails on win and mac
  - error:
    ```
           ➤ YN0000: │ @jupyterlab/repo-top@workspace:. STDERR sh: tsc: command not found
    ```

2. the "dynamic" installation of `playwright` that happens as part of `browser_check.py` is no longer producing a functional install
  - the existing code for this is here:
    https://github.com/jupyterlab/jupyterlab/blob/93c79b6e26bb982ee6ec66ec3e24d8839e68fba0/jupyterlab/browser_check.py#L147-L157
  - running a browser check will eventually result in: 
    ```
    Error: Cannot find module 'playwright'`
    ```
which is raised from a `require('playwright');` call in `browser_check.js`

blink1073:
ACK, I can take a look but probably not until Tuesday

telamonian:
I seem to have been able to fix the `tsc` bug by changing various scripts such as `"build": "tsc"` to `"build": "yarn run tsc"`. Similarly, I was able to fix the  `browser_check.py` bug by changing the subprocess call from `node ...` to `jlpm node ...`:

```python
await run_async_process(["jlpm", "node", "browser-test.js", url], cwd=target) 
```

telamonian:
I've solved the issues above, but I've run into another major snag: `jupyter lab build` hangs the CI. It'll run for 30+ minutes and do nothing. I've tried adding the `--minimize=False` flag and a bunch of other tweaks, but it just doesn't work.

On the other hand, `jupyter lab build` runs just fine locally. Somehow it seems to be panicking the CI container

telamonian:
> I've solved the issues above, but I've run into another major snag: `jupyter lab build` hangs the CI ...On the other hand, `jupyter lab build` runs just fine locally.

Okay, so I finally got to the bottom of the mysterious silent CI-only test failure. Turns out that Yarn v2.x+ sets some extra strict config ONLY in CI environments, not on the desktop, specifically:

```
➤ YN0000: enableImmutableInstalls       true
➤ YN0000: enableInlineBuilds            true
```

Immutable installs set to true means that `yarn.lock` files are not allowed to change when `yarn` or `yarn install` is run. The weird part is that that wasn't the error message, instead there was no error message, the `yarn install` command that runs as part of `jupyter lab build` simply hung for 30+ minutes.

Anyhow, setting the env var `YARN_ENABLE_IMMUTABLE_INSTALLS=0`, or just properly generating the `yarn.lock` file for the app dir so that running `jupyter lab build` doesn't cause any changes to said lockfile, fixes the issue

telamonian:
I've now burnt down all of the test failures. All CI tests are now passing, aside from an unrelated issue that I've already posted a fix for (jupyterlab/jupyterlab_server#207).

So this is now technically working! Yaaay

telamonian:
from @blink1073 on gitter:

> we shouldn't use `yarn` for any commands since a yarn executable could exist
> should be `jlpm run`

Basically, the issue is  whether we should use `foo: yarn run <bin>` or `foo: jlpm run <bin>` in our many, many `package.json` scripts. I've been thinking about this, and here's the pros and cons:

- `foo: yarn run <cmd>`:
  - pro:
    - if called as `yarn run foo`, the script can potentially still work even before `jlpm` is installed (which happens when `pip install jupyterlab` is first run)
  - con:
    - could potentially lead to `yarn` path/executable confusion
      - this is mostly mitigated by the `yarnPath` setting in the `.yarnrc.yml` and `jupyterlab/staging/.yarnrc.yml` config files
- `foo: jlpm run <cmd>`
  - pro:
    - (might?) help to ensure that the same yarn executable is called in all circumstances
  - con:
    - would create a restriction that `pip install .` has to be run before most `package.json` scripts
      - this restriction does not exist in current master for many top level packages such as `buildutils`

So I think really we could go either way with `yarn run` vs `jlpm run`. Guess I'll see what people think at the dev meeting this wednesday

PR #10951, opened 2021-08-29
Bump major Galata version

fcollonval:
Follow up of #10796 
Bump major version of @jupyterlab/Galata as the interpreted package to core as completely changed the API and usage.

> + Rename the workflow to follow original UI tests naming


blink1073 (approved):
👍🏼

PR #10954, opened 2021-08-29
Improve release notes for 3.1

krassowski:
The "New features added" section was missing a lot of the added features. I also described in one-two sentences the the user-facing changes which are not so obvious without knowing about them in the first place. I guess should be done earlier, but better late than never.


blink1073 (approved):
Thanks very much!

blink1073:

Issue #9976, opened 2021-03-22
Improve read-only behaviour for Markdown cells


agoose77:
### Problem
TL;DR
> The default view for read-only Markdown cells is a read-only CodeMirror editor. This is not particularly useful, and in some cases it might be useful to prevent the user from seeing the Markdown source.

A number of content authors are using Jupyter Notebooks as guided workbooks. These notebooks are often pre-populated with Markdown cells which guide the user through a particular set of tasks. When used in this manner, users are not expected to edit the Markdown cells, which should be read-only. 

The notebook specification defines metadata fields that render cells as non-deleteable and read-only. However, JupyterLab implements read-only views on cells as a read-only CodeMirror editor. In the context of the given use case, this means that users can accidentally double click on Markdown cells, which hides the rendered markup. This is not always a desired behaviour, and it would be nice if there were a mechanism by which authors could opt-in to disable this.

### Proposed Solution
@jasongrout proposed adding a setting to prevent showing a read-only cell content editor, and enabling this behaviour by default.

@bollwyvl suggested that one could add a "locked" icon with an `aria-label` to the active cell. This might be an orthogonal change to the setting suggested by @jasongrout.

Mockup:
![image](https://user-images.githubusercontent.com/1248413/111999482-f02cb280-8b14-11eb-9ee9-e4dbc2c6a593.png)

jasongrout:
> @jasongrout proposed adding a setting to prevent showing a read-only cell content editor, and enabling this behaviour by default.

I'd even in be in favor of not having a setting, and just saying that read-only markdown cells do not ever change to show the source. If you want to see the source, first change the cell to editable. My (untested) guess is that in the vast majority of the cases, people accidentally double-click on markdown cells and do not want to see the source.

As a check: what does classic Notebook do?

agoose77:
The behaviour in classic notebook matches that of JupyterLab

goanpeca:
>The behaviour in classic notebook matches that of JupyterLab. Personally a 👍 on your suggestion to never show an editor for read-only Markdown cells.

👍🏼 on this!

bollwyvl:
So, down in the guts:
- A: `packages/cells/src/widget.ts`
  - there's `rendered` and `readOnly`, both getter/setters
- B: `packages/notebook/src/widget.ts`
  - there's a `mode` which sets `rendered` iff it detects a `MarkdownCell`

does it make sense to: 

[...]

PR #10956, opened 2021-08-29
Enable not showing editor for read-only Markdown cells

krassowska:
## References

Task 1 in #9976

## Code changes

- The value passed to `set rendered(value: boolean)` setter is overwritten if cell is read-only and user set `showEditorForReadOnlyMarkdown` to `false`.
- Added optional `showEditorForReadOnlyMarkdown?: boolean` to `MarkdownCell.IOptions`, and `INotebookConfig.IOptions` to pass the settings down to the markdown cell. This setting updates on change (the user does not need to reload JupyterLab)

## User-facing changes

Editor does not show up when clicking on read-only Markdown cell (a cell where metadata include `"editable": true`) if the notebook setting `showEditorForReadOnlyMarkdown` is set to `false`. Default for `"showEditorForReadOnlyMarkdown"` setting is `true`.

None


krassowska:
@meeseeksdev tag enhancement

packages/cells/src/widget.ts
```diff
@@ -1486,6 +1486,10 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     });
     this.renderCollapseButtons(this._renderer!);
     this.renderInput(this._renderer!);
+    this._showEditorForReadOnlyMarkdown =
+      typeof options.showEditorForReadOnlyMarkdown === 'undefined'
+        ? MarkdownCell.defaultShowEditorForReadOnlyMarkdown
+        : options.showEditorForReadOnlyMarkdown;
```
fcollonval, line 1492: You could simplify this with the nullish operator:

```suggestion
      options.showEditorForReadOnlyMarkdown ?? MarkdownCell.defaultShowEditorForReadOnlyMarkdown;
```

packages/cells/src/widget.ts
```diff
@@ -1762,6 +1770,7 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
   private _rendered = true;
   private _prevText = '';
   private _ready = new PromiseDelegate<void>();
+  private _showEditorForReadOnlyMarkdown: boolean = true;
```
fcollonval, line 1773: This is strange that it was not flagged by the linter. But the rule is to not specify type for variable with default value and basic type (see e.g. `_rendered` or `_prevText` above):

```suggestion
  private _showEditorForReadOnlyMarkdown = true;
```

packages/notebook-extension/schema/tracker.json
```diff
@@ -878,6 +878,12 @@
       "type": "number",
       "default": 50
     },
+    "showEditorForReadOnlyMarkdown": {
+      "title": "Show editor for read-only Markdown cells",
+      "description": "Should an editor be shown for read-only markdown",
+      "type": "boolean",
+      "default": true
+    },
```
fcollonval, line 886: > This is more for discussion with core devs and don't need to be addressed by this PR.

It feels weird to have sections like `MarkdownCellConfig`, `CodeCellConfig`,... but to get `maxNumberOutputs` and `showEditorForReadOnlyMarkdown` outside those sections (aka because these `CellConfig` are actually `Cell`**Editor**`Config`). It would be good if the settings and code structure reflect that better.

packages/notebook/src/widget.ts
```diff
@@ -842,6 +845,7 @@ export class StaticNotebook extends Widget {
   private _placeholderCellRendered = new Signal<this, Cell>(this);
   private _observer: IntersectionObserver;
   private _renderedCellsCount = 0;
+  private _showEditorForReadOnlyMarkdown?: boolean;
```
fcollonval, line 848: We don't need that attribute as it is part of `this._notebookConfig`.

```suggestion
```

packages/notebook/src/widgetfactory.ts
```diff
@@ -95,7 +95,9 @@ export class NotebookWidgetFactory extends ABCWidgetFactory<
       notebookConfig: source
         ? source.content.notebookConfig
         : this._notebookConfig,
-      translator
+      translator,
+      showEditorForReadOnlyMarkdown: this._notebookConfig
+        .showEditorForReadOnlyMarkdown
```
fcollonval, line 100:
```suggestion
      translator
```

packages/notebook/src/widget.ts
```diff
@@ -663,13 +664,15 @@ export class StaticNotebook extends Widget {
...
+    const showEditorForReadOnlyMarkdown = this._showEditorForReadOnlyMarkdown;
     const options = {
       editorConfig,
       model,
       rendermime,
       contentFactory,
       updateEditorOnShow: false,
-      placeholder: false
+      placeholder: false,
+      showEditorForReadOnlyMarkdown
```
fcollonval, line 675:
```suggestion
      showEditorForReadOnlyMarkdown: this._notebookConfig.showEditorForReadOnlyMarkdown
```

packages/notebook/src/widget.ts
```diff
@@ -663,13 +664,15 @@ export class StaticNotebook extends Widget {
     const rendermime = this.rendermime;
     const contentFactory = this.contentFactory;
     const editorConfig = this.editorConfig.markdown;
+    const showEditorForReadOnlyMarkdown = this._showEditorForReadOnlyMarkdown;
```
fcollonval, line 667:
```suggestion
```

packages/notebook/src/widget.ts
```diff
@@ -208,6 +208,7 @@ export class StaticNotebook extends Widget {
       options.notebookConfig || StaticNotebook.defaultNotebookConfig;
     this._updateNotebookConfig();
     this._mimetypeService = options.mimeTypeService;
+    this._showEditorForReadOnlyMarkdown = options.showEditorForReadOnlyMarkdown;
```
fcollonval, line 211:
```suggestion
```

packages/notebook/src/widget.ts
```diff
@@ -889,6 +893,11 @@ export namespace StaticNotebook {
      * The application language translator.
      */
     translator?: ITranslator;
+
+    /**
+     * Show editor for read-only Markdown cells.
+     */
+    showEditorForReadOnlyMarkdown?: boolean;
```
fcollonval, line 900:
```suggestion
```

fcollonval (changes requested):
Thanks @krassowska I have a couple of suggestions to simplify the code.

One additional point that would be good too is to react on the change of the settings value. For now if you will need to close and reopen a notebook to see the setting change effect.
You could do that by extending the `StaticNotebook._updateNotebookConfig` that gets called when the notebook config changes:

https://github.com/jupyterlab/jupyterlab/blob/master/packages/notebook/src/widget.ts#L814

packages/cells/src/widget.ts
```diff
@@ -1574,6 +1581,13 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     }
   }
 
+  set showEditorForReadOnlyMarkdown(value: boolean) {
```
fcollonval, line 1584: Could you either use a method `setShowEditorForReadOnly` or a getter? And add a small doc string.

Defining only a setter is not a good practice because it feels strange to be able to do `myObj.a = value` but not `const value = myObj.a`.

> It is ok to drop `Markdown` as we are in the `MarkdownCell`

fcollonval (changes requested):
Thanks for the modifications @krassowska 

I have a last minor suggestion.

krassowska:
Thank you @fcollonval for review!

fcollonval:

PR #10973, opened 2021-08-31
Restore test for kernel that does not support debugger

### LINKED ISSUE #10525: Restore test for kernel that does not support debugger
## Description

`js-debugger` tests [started failing](https://github.com/jupyterlab/jupyterlab/pull/10519/checks?check_run_id=2957338995#step:9:31729) as can be seen on #10519 and #10524 on "Debugging support" → `#isAvailable` → `should return false for kernels that do not have support for debugging`. This is likely due to `ipykernel` 6 release:

https://github.com/jupyterlab/jupyterlab/blob/dbdefed9db9332381fe4104bdf53ec314451951f/packages/debugger/test/service.spec.ts#L95-L98

I wonder if we have another another kernel that does not support debugging that we could use here?

fcollonval:
## References

Fix #10525

## Code changes

N/A

## User-facing changes

N/A

N/A

### DIFF
--- a/packages/debugger/test/service.spec.ts
+++ b/packages/debugger/test/service.spec.ts
@@ -92,7 +92,7 @@ describe('Debugging support', () => {
       expect(enabled).toBe(true);
     });
 
-    it.skip('should return false for kernels that do not have support for debugging', async () => {
+    it('should return false for kernels that do not have support for debugging', async () => {
       const enabled = await service.isAvailable(ipykernel);
       expect(enabled).toBe(false);
     });


jtpio:
Thanks @fcollonval.

Wondering whether the test should actually still fail, since the latest `ipykernel` 6 now supports debugging?

jtpio:
Or maybe it passes because the kernel spec is mocked here without the `debugger` key in the metadata:

https://github.com/jupyterlab/jupyterlab/blob/32ccdf162706726531265e1fb14a7c9f2fe04beb/testutils/src/start_jupyter_server.ts#L262-L266

fcollonval:
> Or maybe it passes because the kernel spec is mocked

Yes this is because of the mock. This is why I though it was good to restore the test. The reason of failures was not related to the debug ability introduced in ipykernel 6.

jtpio:
Alright, thanks for the confirmation :+1:

blink1073:

jtpio:
Maybe a follow-up could be to add a comment to the tests to mention this corresponds to a mocked spec of `ipykernel`. Or we rename the `ipykernel` variable in the tests for clarity.

Issue #10955, opened 2021-08-29
Galata failures on master, lack of artifacts, empty report


krassowski:
## Description

- [x] After merging #10796 there are failures on master
- [ ] I tried to download `ui-test-output` artifacts as instructed in https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests but this artifact is no longer there; the two artifacts that are downloadable contain:
   - `galata-test-assets`: it only contains `.webm` movies, and I cannot easily find the folder for the movie for the failing test case
   -  `galata-report`: `.webm` movies and some `.png` files, but those are not helpful due to their names
     ![Screenshot from 2021-08-29 17-27-02](https://user-images.githubusercontent.com/5832902/131257972-76300945-30d6-4f83-92c4-573bc338c7d6.png)
- [ ] the report is empty for me, even though there are many files (as shown above):
   ![report](https://user-images.githubusercontent.com/5832902/131257944-d798bc01-d110-426d-b307-e8008215791e.gif)

fcollonval:
> After merging #10796 there are failures on master

I'm working on #10950 to make them more robust

> I tried to download ui-test-output artifacts as instructed in https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests but this artifact is no longer there. The two artifacts that are downloadable contain:
galata-test-assets: it only contains .webm movies, and I cannot easily find the folder for the movie for the failing test case
galata-report: .webm movies and some .png files, but those are not helpful due to their names

The two artifacts are now:
- galata-test-assets: Raw artifacts with one folder per tests - artifacts are videos and some images if an error occurred when comparing screenshots. Each folder is built using the same pattern `<path-to-test-file>-<dashed-test-title>`  
We can change the default settings to retain only the failing test as done previously by galata.
- galata-report: The report needs to be started with a server within the extracted folder. `npx http-server -a localhost -o <path-to-report>` - this is required as the report data is in JSON format loaded using `fetch`.

> I opened an issue upstream to at least display an error in that case (https://github.com/microsoft/playwright/issues/8549)

A possible workaround would be to build or own report template (I tried to avoid that to reduce our maintenance load).

krassowski:
> I'm working on #10950 to make them more robust

:tada:

> We can change the default settings to retain only the failing test as done previously by galata.

[...]

PR #10980, opened 2021-08-31
Improve Galata documentation and setup

fcollonval:
## References

Address some points raised in #10955

## Code changes

Galata tests will by default not keep uploaded and created files even in case of failure.

Use `retries` equals 1 to mitigate flaky UI tests without retrying too much when something is really broken.

Raw assets kept are only for failures tests (but all are kept in the other artifact with the report).

in-page code make use only of JupyterLab app object (replace `NotebookActions` calls by commands calls) to reduce the injected JS artifact size from 4Mo to 120ko.

## User-facing changes

N/A

N/A

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -530,8 +530,8 @@ Visual Regression and UI Tests
 ------------------------------
 
 As part of JupyterLab CI workflows, UI tests are run with visual regression checks.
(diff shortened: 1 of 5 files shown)


blink1073:
Looks like a formatting error: `/home/runner/work/jupyterlab/jupyterlab/docs/source/developer/contributing.rst:562: WARNING: Inline literal start-string without end-string.`

jtpio (approved):
CI all green, thanks!

PR #10373, opened 2021-06-09
Build context menu from settings for easy customization

fcollonval:
## References

Follow-up of #10254 to define context menu from settings.

## Code changes

- Context menu is built from a data definition set in application-extension settings
- `JupyterLabMenu` has been renamed `RankedMenu` and moved to ui-components package (it is still exported with the old name from the mainmenu package for backward compatibility)
- Integration test for the context menu have been added
- :warning: the code to build dynamically the _Open With_ submenu has been changed to not use a subclass of Menu to fit the use of a data backed item definition.

## User-facing changes

N/A

But the context menu can now be customize from the settings.

## Backwards-incompatible changes

N/A

PR #10992, opened 2021-09-02
Move RankedMenu test to ui-components

fcollonval:
## References

Follow-up of #10373

## Code changes

Refactor test to reflect new code structure introduce in #10373

## User-facing changes

N/A

N/A


fcollonval:
Ja I don't why the mock is not working properly. It probably needs more mock to avoid the need of a real server running (at least for the simple debugging feature availability test).

PR #11003, opened 2021-09-03
fix #10997 - increase max_message_size of websocket messages

### LINKED ISSUE #10997: With RTC enabled, large notebooks to fail on save and whole UI becomes extremely sluggish.
## Description

When Lab is started with `--collaborative`, certain large notebooks fail to save or take inordinately long to do so.

## Reproduce

[This Gist](https://gist.github.com/06cc8cab195a7ebaa97bf42ff77dd141.git) contains a notebook that illustrates the problem. To reproduce, clone the gist and run side-by-side JupyterLab with and without `--collaborative`. 

I tested on a Mac Mini with Big Sur, Python 3.9, 64GB of RAM and JupyterLab 3.1.10.

The version with `--collaborative` on will either not save at all, or take easily 10 minutes to do a save. In addition, the whole UI gets very sluggish, and the browser occasionally pops warnings about the page consuming too many resources (seen on Firefox, Chrome and Safari).

## Additional context

We spent some time tracking down this problem [over at the Jupyter meets the Earth repo](https://github.com/pangeo-data/jupyter-earth/issues/78). Flagging that thread for reference, but all relevant info is here.

dmonad:
## References

Closes #10997. As described in the ticket, with collaborative editing enabled the websocket connection closes when the client sends a message larger than 10mb. This ultimately leads to performance problems and being unable to save files. 

## Code changes

I increased the `max_message_size` of websocket messages to 1GB. We might want to increase this even further if we expect files larger than 1GB (e.g. because of large images).

## User-facing changes

None.

None.

### DIFF
--- a/jupyterlab/handlers/yjs_echo_ws.py
+++ b/jupyterlab/handlers/yjs_echo_ws.py
@@ -36,6 +36,11 @@ class YjsRoom:
 class YjsEchoWebSocket(WebSocketHandler):
     rooms = {}
 
+    # Override max_message size to 1GB
+    @property
+    def max_message_size(self):
+        return 1024 * 1024 * 1024
+
     def open(self, guid):
         #print("[YJSEchoWS]: open", guid)
         cls = self.__class__


blink1073 (approved):
Thank you!

blink1073:

Issue #10737, opened 2021-07-29
Gaps in translation coverage


krassowski:
## Description

The Chinese language pack has 100% coverage so everything that does not display in the UI as translated is a bug:

Remaining:
- ~~`Locals`, `Arguments` and `registers` in debugger (I cannot find them - are those served by the debugger itself? If so, there could be a map on frontend)~~ -> Not coming from JupyterLab code.
- [ ] `Build failed with`, `Build aborted` etc in `builder/index.src`  
  > Will be though to address and is less critical with the wider spread of pre-built extensions
- [ ] do shortcuts (which are handled by lumino) need translation? (there were voices saying that its not needed on CrowdIn)
    ![Screenshot from 2021-07-29 19-16-33](https://user-images.githubusercontent.com/5832902/127544918-c6eb314e-bcda-4357-b522-82be2a2027b8.png)

Done:
- [x] "Show more outputs": https://github.com/jupyterlab/jupyterlab/pull/18021    https://github.com/jupyterlab/jupyterlab/blob/358e470dd7bf37806827b4ad4e2c4c4fb234d1f3/packages/outputarea/src/widget.ts#L1354-L1355
- [x] top-level menu bar is not translated and menus with submenues are not translated (fixed, released in 3.1.8, see #10739)
    ![Screenshot from 2021-07-29 19-16-18](https://user-images.githubusercontent.com/5832902/127544921-65b12cfc-004f-4c82-9139-649371bcc193.png)
- [x] <s>"show last modified column" in file browser</s> new in 3.1, strings updated on CrowdIn on 2021-08-17
    ![Screenshot from 2021-07-29 19-26-57](https://user-images.githubusercontent.com/5832902/127546158-d53c1554-521b-4a12-9c52-979c42333c56.png)
- [x] <s>Merge Cell Below/Above commands</s> new in 3.1, strings uploaded on CrowdIn on 2021-08-17
    ![Screenshot from 2021-07-29 19-27-50](https://user-images.githubusercontent.com/5832902/127546344-06ee0ed7-b273-45ae-9a22-ec809ba933b3.png)
- [x] Kernel status in status bar (fixed in 3.1.6, see #10738)
    ![Screenshot from 2021-07-29 19-29-39](https://user-images.githubusercontent.com/5832902/127546581-267cae98-779c-4a30-9673-6c96340e58b7.png)
[...]

PR #11006, opened 2021-09-03
Translate factory names by adding `label`

krassowski:
## References

Fixes a translation point from https://github.com/jupyterlab/jupyterlab/issues/10737#issuecomment-910619040

## Code changes

- clarify that `name` acts as a de-facto identifier, remove translation from `Workspace loader` name (ideally `name` would be renamed to `id`, but this would be a breaking change for quite a few extensions).
- interface `IWidgetFactoryOptions` has an optional (for now) `label` member

## User-facing changes

Open with menu entries will now be translated; also the CSV/TSV viewers will have a proper label:

Before

Screenshot from 2021-09-03 21-27-40

After

Screenshot from 2021-09-03 21-41-45

I am never sure what makes a cut - I would say this can be backported, but I did add an optional dependency on translator to `vdom-extension` and the interface `IWidgetFactoryOptions` has an optional `label` member.

### DIFF
--- a/packages/json-extension/src/index.tsx
+++ b/packages/json-extension/src/index.tsx
@@ -97,6 +97,7 @@ const extensions: IRenderMime.IExtension | IRenderMime.IExtension[] = [
     dataType: 'json',
     documentWidgetFactoryOptions: {
       name: 'JSON',
+      // TODO: how to translate label of the factory?
       primaryFileType: 'json',
       fileTypes: ['json', 'notebook', 'geojson'],
       defaultFor: ['json']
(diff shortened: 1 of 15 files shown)


fcollonval (approved):
LGTM

I don't have any idea how to tackle the case of rendermime extension.

packages/json-extension/src/index.tsx
```diff
@@ -97,6 +97,7 @@ const extensions: IRenderMime.IExtension | IRenderMime.IExtension[] = [
     dataType: 'json',
     documentWidgetFactoryOptions: {
       name: 'JSON',
+      // TODO: how to translate label of the factory?
```
fcollonval, line 100: That one sounds really tricky. I presume we can trick the string extraction algorithm with:

```
const trans = {__: (s: string) => s}

// ...
  name: 'JSON',
  label: trans.__('JSON'),
```
Then translate the label when the plugin is generated. But this is really ugly.

PR #11010, opened 2021-09-04
Fix ignored promise leading to incorrect initial tooltip position

krassowski:
## References

While working on improved signature display in https://github.com/krassowski/jupyterlab-lsp/pull/671 I noticed that the first position of tooltip is incorrect, but upon any user action (scroll, keypress , anything which triggers lumino update message) it gets automatically fixed. I figured it is because the position is calculated prior to the content being set by the model rendering (so the proper size information is not yet available).

## Code changes

This PR fixes the issue by properly handling the promise returned by `renderModel()` in the tooltip widget constructor (setting geometry after it resolves, and showing error if anything goes wrong).

## User-facing changes

Tooltips will be positioned better on first render. This allows to consistently place LSP signature tooltip above the line:

Screenshot from 2021-09-04 13-18-07

None

### DIFF
--- a/packages/tooltip/src/widget.ts
+++ b/packages/tooltip/src/widget.ts
@@ -69,7 +69,10 @@ export class Tooltip extends Widget {
     }
 
     this._content = this._rendermime.createRenderer(mimeType);
-    void this._content.renderModel(model);
+    this._content
+      .renderModel(model)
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

blink1073:

PR #11022, opened 2021-09-07
Configure sphinx for gettext

fcollonval:
## References

Follow-up on translation of the UI. This introduces small tweak on the sphinx configuration to deal with internationalization

## Code changes

Don't generate TS API when dealing with gettext subcommand

## User-facing changes

N/A

N/A


blink1073 (approved):
Thank you!

Issue #10852, opened 2021-08-16
Real-time collaborative editing is broken 


dclong:
## Description

I  visited the same JupyterLab server (with RTC enabled) from 2 computers. RTC keeps adding copies of existing cells into the notebook. And the worst thing was that the latest change from a computer was lost. 

## Reproduce

I do not know how to reproduce this issue exactly but I can describe my situation here. 
Hopefully, someone can confirm the same observation later. 

1. Started a JupyterLab service with RTC enabled. 
2. Worked on a notebook from a Windows PC for a while.
3. Visit the notebook from a Mac Book (authenticated using the JupyterLab sever's token) while someone is still working on the notebook on the Windows PC. 
4. Put the Mac Book to sleep (by close its lid). 
5. Done working on the notebook on the Windows PC but leave the browser and notebook tabs open. 
6. Several hours later I found that the notebook on the Windows PC has been altered. The most recent changes in the notebook was gone. And existing cells was duplicated and pasted into the notebook. When I visited the JupyterLab server on the Mac Book again, the duplicating existing cells becomes even more obvious. I literally see new cells being added into the notebook. 

## Expected behavior

RTC editing should work well (like Google docs). 

## Context

- JupyterLab 3.1.6 was launched in a Docker container and Docker container was launch from WSL 2 (Ubuntu 20.04).
- Browser One: Chrome 92.0.4515.131 on macOS Big Sur (11.5)
- Browser Two: Chrome on Windows 10

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

hbcarlos:
Hi @dclong, this is an issue we have been trying to reproduce for a while without success.

Maybe this could be related to #10816, which I'm working on right now.

I'll keep updating this thread with new findings.

dmonad:
This is probably related to the locking mechanism that we implemented. When many users connect, there is a chance that a user with a slow network connection writes a document with a timed-out lock. I recommend that we first increase the lock-timeout and then also check locally whether the lock has expired before overwriting a document.

[...]

PR #11026, opened 2021-09-08
Retain the rtc lock until the user releases it

hbcarlos:
In the JupyterLab weekly meeting, we sometimes observe the meeting notes being erased at the begging of the meeting when everyone is connecting to the binder instance at the same time.

When a client needs to save the document to disk, the client requires a lock to the server, and once it is assigned to him, then writes the content to disk. We suspect the issue comes from the lock that the server assigns to the client. If the server doesn't receive a new request from the client in a certain timeout, the server assumes that the client is gone and releases the lock.

**We have not been able to reproduce the issue. It's possible that this PR doesn't solve it.**

## References
Tries to solves #10962 and #10852.

## Code changes
The solution we are implementing in this PR increases the timeout on the server by the number of connected clients when the client requests the lock. In addition, the client keeps sending request messages to acquire the lock until it's done with it and ready to release it. This way, the knows that the client is not gone and doesn't release the lock.

## User-facing changes
N/A

N/A

Pining @dmonad for a review.

### DIFF
--- a/jupyterlab/handlers/yjs_echo_ws.py
+++ b/jupyterlab/handlers/yjs_echo_ws.py
@@ -61,17 +61,23 @@ class YjsEchoWebSocket(WebSocketHandler):
         room = cls.rooms.get(room_id)
         if message[0] == ServerMessageType.ACQUIRE_LOCK:
             now = int(time.time())
-            if room.lock is None or now - room.lock > 15: # no lock or timeout
+            if room.lock is None or now - room.lock > (10 * len(room.clients)) : # no lock or timeout
                 room.lock = now
+                room.lock_holder = self
                 # print('Acquired new lock: ', room.lock)
                 # return acquired lock
                 self.write_message(bytes([ServerMessageType.ACQUIRE_LOCK]) + room.lock.to_bytes(4, byteorder = 'little'), binary=True)
+            
+            elif room.lock_holder == self :
+                room.lock = now
+
         elif message[0] == ServerMessageType.RELEASE_LOCK:
             releasedLock = int.from_bytes(message[1:], byteorder = 'little')
             # print("trying release lock: ", releasedLock)
             if room.lock == releasedLock:
                 # print('released lock: ', room.lock)
                 room.lock = None
+                room.lock_holder = None
         elif message[0] == ServerMessageType.REQUEST_INITIALIZED_CONTENT:
             # print("client requested initial content")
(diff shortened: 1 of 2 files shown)


packages/docprovider/src/yprovider.ts
```diff
@@ -195,7 +195,7 @@ export class WebSocketProviderWithLocks
     }
     this._sendMessage(new Uint8Array([127]));
     // try to acquire lock in regular interval
-    const intervalID = setInterval(() => {
+    this._intervalID = setInterval(() => {
```
dmonad, line 198: You should make sure to clear the current `_intervalID` before creating a new one. Maybe we should also rename this to `this._requestLockInterval`

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -56,17 +56,23 @@ def on_message(self, message):
...
             now = int(time.time())
-            if room.lock is None or now - room.lock > 15: # no lock or timeout
+            if room.lock is None or now - room.lock > (10 * len(room.clients)) : # no lock or timeout
                 room.lock = now
+                room.lock_holder = self
                 # print('Acquired new lock: ', room.lock)
                 # return acquired lock
                 self.write_message(bytes([ServerMessageType.ACQUIRE_LOCK]) + room.lock.to_bytes(4, byteorder = 'little'), binary=True)
+            
+            elif room.lock_holder == self :
```
dmonad, line 66: Did you make sure that the current implementation supports renewing their lock?

dmonad (changes requested):

jupyterlab/handlers/yjs_echo_ws.py
hbcarlos: Yes, I test it and looks like it works.

packages/docprovider/src/yprovider.ts
hbcarlos: I added a commit with the changes.

dmonad (approved):
Looks good :+1:

blink1073:
@hbcarlos, can you please run `jlpm run lint` and push the changes so we can get this into today's release?

blink1073:
Proceeding with today's release, we can get this into the next one

hbcarlos:
Sorry @blink1073, I missed your message. We can add this to the next release.

blink1073:
Thanks!  I'm actually waiting until tomorrow for the 3.1.x release, so we can get this in.

blink1073:

hbcarlos:
Thank you soo much!

PR #11036, opened 2021-09-08
Check changes on translatable strings

fcollonval:
## References

Follow-up [jupyterlab/team-compass#128 (comment)](https://github.com/jupyterlab/team-compass/issues/128#issuecomment-910424790)

## Code changes

Add test for the translatable strings by using `juyterlab-translate` to extract the strings and create a hash from the entries of the POT file. The test fails if the hash changes. And the error message request to update the hash if the target branch is master otherwise it displays the policy:

```
Translatable strings can be changed only for major or minor version.
```

> To fix the test, the hash needs to be updated locally:

```
python -m pip install jupyterlab-translate
python scripts/i18n_check.py -u
```

## User-facing changes

N/A

N/A


fcollonval:
Question:
- Should we raise an error instead of commenting on the PR? If so, what should be the workflow target (something like all pull_request on `\d+\.\d+\.\d+` branches)?

blink1073:
I think a warning on master and an error on stable branches makes sense

fcollonval:
Actually it needs to always fail because if the strings are changing in the legit case, the hash needs to be updated.

krassowski:
CodeQL compares two last runs instead. Would it be an option here?

fcollonval:
> CodeQL compares two last runs instead. Would it be an option here?

I'm not sure to get what you are proposing. Do you mean to only check the changes between the target branch and the pr head?

> The check done for now is to use `juyterlab-translate` to extract the strings and create a hash from the entries of the POT file.

goanpeca (approved):
This is awesome @fcollonval ! 🚀

fcollonval:

PR #11043, opened 2021-09-09
Fix prettier error

fcollonval:
## References

since the release of v4.0.0a8 prettier is complaining. And indeed the tsconfig.json file of that package is not _pretty_.
Seems that `prettier` should be run by the releaser.

## Code changes

Make the file pretty again

## User-facing changes

N/A

N/A

### DIFF
--- a/packages/vdom-extension/tsconfig.json
+++ b/packages/vdom-extension/tsconfig.json
@@ -4,9 +4,7 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": [
-    "src/*"
-  ],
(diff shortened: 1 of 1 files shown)


jtpio:
>  Seems that `prettier` should be run by the releaser.

Maybe we can run it as a hook? (probably `after-bump-version`)

https://github.com/jupyterlab/jupyterlab/blob/15d51ef6c521b9680584dc429bee82f53f46215c/pyproject.toml#L39-L43

blink1073 (approved):
Thanks!  Merging to get us un-stuck

PR #10936, opened 2021-08-27
Add benchmark tests

fcollonval:
Use playwright to benchmark performance in CI

This builds upon #10796 and the [benchmarks](https://github.com/jupyterlab/benchmarks) repository.

### Code changes

:warning: Upgrade `vega-lite` from v4 to v5 to solve [`null` type issue](https://github.com/vega/vega-lite/pull/7352).

- [x] Update comment instead of appending a new report each time

PR #11047, opened 2021-09-09
Fix benchmark PR commenting for forks

fcollonval:
## References

follow-up on #10936

Commenting on PR needs to be done in two steps so that it can happen for PR from forks.

## Code changes

N/A

## User-facing changes

N/A

N/A


jtpio (approved):
Thanks!

PR #11031, opened 2021-09-08
Use posix paths explicitly

Mithil467:
## References
Fixes #11030 

## Code changes

1. Using `path.posix.join` instead of `path.join` (only where needed).
2. Using `.replace('\\', '/')` to convert win path to posix.
3. Commit updated format for the json files not regarding paths.

Turns out there is no inbuilt function in Node.js path module to convert win paths to posix. Hence used replace.

PR #11048, opened 2021-09-09
use path.posix explicitly for URLs

mbektas:
JupyterLab Desktop App (https://github.com/mbektas/jupyterlab_app/tree/jlab3_upgrade) uses webpack target `electron-renderer` and it bundles JupyterLab packages as part of app's js bundle. When Desktop App is built for Windows, during runtime, the URLs generated by `URLExt.join` end up having path components with backward slash (i.e. http://localhost:8888\api\kernels). It is caused by path.join using Windows specific win32.join. This PR enforces using `posix.join` for URL generation.

## Code changes

istead of using `path.join` which is platform dependent, now using `path.posix` for URL path component joins.

## User-facing changes

none

none

### DIFF
--- a/packages/coreutils/src/url.ts
+++ b/packages/coreutils/src/url.ts
@@ -2,7 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { PartialJSONObject } from '@lumino/coreutils';
-import * as posix from 'path';
+import { posix } from 'path';
 import urlparse from 'url-parse';
 
 /**


jasongrout:
To expand: we probably didn't see a problem before because we are using the browser substitute, path-browserify, which *only* contains the posix implementation: https://github.com/browserify/path-browserify/blob/872fec31a8bac7b9b43be0e54ef3037e0202c5fb/index.js#L527-L529

However, is it true that in this electron case, we are actually using the Node path module (since it is on the desktop), and so it is, as [documented in Node](https://nodejs.org/api/path.html#path_path_join_paths), using os-dependent path separators?

jasongrout:

jasongrout:
This is similar to #11031

mbektas:
> However, is it true that in this electron case, we are actually using the Node path module (since it is on the desktop), and so it is, as [documented in Node](https://nodejs.org/api/path.html#path_path_join_paths), using os-dependent path separators?

yes @jasongrout , electron app is resolving to use os-dependent node path. I tried to alias it in webpack.config to use `path-browserify` but it didn't work.

Issue #6957, opened 2019-08-06
Comments should be indented


technic:
**Describe the ~~bug~~ feature**
JupyterLab toggle comment action should create indented comments i.e.
```python
def f(x):
    # comment
    return x**2
```
instead of 
```python
def f(x):
#    comment
    return x**2
```

This was discussed before for the notebook https://github.com/jupyter/notebook/issues/954, but I didn't find any info regarding JupyterLab.

**To Reproduce**
Type some code, press `Ctrl`+`/` button

**Expected behavior**
Correct indentation like almost any IDE or text editor will do.

**Desktop (please complete the following information):**

- OS: [Windows 10]
- Browser [chrome 75.0]
- JupyterLab [Version 1.0.2]

lukelbd:
Hoping someone has time to look into this soon. The current block-commenting behavior results in code that does not conform to the increasingly popular ["black" format](https://github.com/psf/black), and therefore conflicts with the [code-formatter plugin](https://jupyterlab-code-formatter.readthedocs.io/en/latest/).

For example, suppose I have the following python cell: 

```python
if condition:
    do_something_0
    do_something_1
    do_something_2
    do_something_3
```

and comment out code with `Cmd + /` as follows:

```python
if condition:
    do_something_0
#     do_something_1
#     do_something_2
    do_something_3
```

If I then automatically format the cell with `black` (which is the *default* formatter for the [code-formatter extension](https://jupyterlab-code-formatter.readthedocs.io/en/latest/)), the code looks like this:

```python
if condition:
    do_something_0
    #     do_something_1
    #     do_something_2
    do_something_3
```

and uncommenting the code with `Cmd + /` again results in invalid python code:

```python
if condition:
    do_something_0
        do_something_1
        do_something_2
    do_something_3
```

Plus, the current `Cmd + /` block-commenting behavior just looks plain strange to me.

If it helps, it looks like this issue was addressed for jupyter notebooks in jupyter/notebook#954.

lukelbd:
@blink1073 Who would be best equipped to address this? This issue isn't critical but has been a year with no activity, not sure if it's on anyone's radar. Was just hoping to get acknowledgement from the core jupyter contributors.

technic:
I would like to have this fixed. Please let us know if you need any help with it.

jasongrout:
[...]

PR #11063, opened 2021-09-13
Indent comments (#6957)

josephrocca:
## References

https://github.com/jupyterlab/jupyterlab/issues/6957

## Code changes

Pass `{indent:true}` option object to `toggleComment` (on `Ctrl + /` and `Cmd + /` keypress) so that it indents properly, per @jasongrout's [comment](https://github.com/jupyterlab/jupyterlab/issues/6957#issuecomment-655199796) on #6957.

Please `Ctrl+f` for "comment.js" on [this page](https://codemirror.net/doc/manual.html) so see the relevant part of the CodeMirror manual. Note that the comment.js addon adds the "toggleComment" command (which is what is currently used) as a shortcut for option-less usage, but it also adds the `.toggleComment(opts)` method to the CodeMirror instance, which is what's being in the new code introduced by this pull request.

## Edit:
I tried to use the binder test link provided by jupyterlab-probot, but after seeming to successfully build, it logged "Failed to connect to event stream":
```
Pushing image
Pushing image
Pushing image
Successfully pushed gesiscss/binder-r2d-g5b5b759-josephrocca-2djupyterlab-48e787:0bbdd9526cc16beafa5d0a04c93f98a515db2f93Built image, launching...
Launching server...
Failed to connect to event stream
```
I tried a second time and the same thing happened.

### DIFF
--- a/packages/codemirror/src/factory.ts
+++ b/packages/codemirror/src/factory.ts
@@ -25,8 +25,8 @@ export class CodeMirrorEditorFactory implements IEditorFactoryService {
         'Cmd-Left': 'goLineLeft',
         Tab: 'indentMoreOrinsertTab',
         'Shift-Tab': 'indentLess',
-        'Cmd-/': cm => cm.toggleComment({indent:true}),
-        'Ctrl-/': cm => cm.toggleComment({indent:true}),
+        'Cmd-/': cm => cm.toggleComment({ indent: true }),
+        'Ctrl-/': cm => cm.toggleComment({ indent: true }),
         'Ctrl-G': 'find',
         'Cmd-G': 'find'
       },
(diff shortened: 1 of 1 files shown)


packages/codemirror/src/factory.ts
```diff
@@ -25,8 +25,8 @@ export class CodeMirrorEditorFactory implements IEditorFactoryService {
         'Cmd-Left': 'goLineLeft',
         Tab: 'indentMoreOrinsertTab',
         'Shift-Tab': 'indentLess',
-        'Cmd-/': 'toggleComment',
-        'Ctrl-/': 'toggleComment',
+        'Cmd-/': (cm) => cm.toggleComment({indent:true}),
```
blink1073, line 28: I ran the linter locally, and these lines should be formatted as `cm => cm.toggleComment({ indent: true }),`, mind updating them?

blink1073 (approved):
Thanks!  Just needs a lint pass

fcollonval (approved):
Thanks a lot @josephrocca 

It works even for multi-lines.

The only corner case I found is the follow-up.

```py
def f():
    print("a")

    print("b")
```

Will be commented as:

```py
def f():
#    print("a")

#    print("b")
```

This is due to the empty line having no white spaces. But anyway this is a limitation of CodeMirror and this will bring a good improvements.

fcollonval:
@josephrocca you need also to place white space around `indent: true`. But to be sure to fix the linter, the best is to run locally `jlpm run prettier` then commit the changes.

josephrocca:
@fcollonval Sorry, I'm just using the web UI to edit, please feel free to push to my branch.

blink1073:

jasongrout:
@meeseeksdev please backport to 3.0.x

Issue #11038, opened 2021-09-09
collaboration and tags


djangoliv:
## Description

When we share a notebook with tags, the tags are removed.

## Reproduce

0. jupyterlab --collaborative
1. Create a notebook with cell and tags
2. Save it
3. Use jupyterlab-link-share to copy the share link
4. Open the link in private mode (incognito-mode)
5. The file is rewrite without the tags

## Expected behavior

Keep the tags.

## Context

- firefox: 91.0.2 (64 bits)
- JupyterLab 3.1.10

Regards

hbcarlos:
Hi @djangoliv, thanks for opening this issue.

The tags are stored in the cell's metadata, and by default, the metadata is not shared.
This is an ongoing discussion because there are some attributes in the metadata (like `kernelspec`) that we are not sure if should be shared or not.

hbcarlos:
@meeseeksdev tag "Real Time Collaboration"

hbcarlos:
@meeseeksdev tag "tag:Real Time Collaboration"

blink1073:
Oh, so the issue was that the tag name didn't exist, I thought something might have been wrong with our [config](https://github.com/jupyterlab/jupyterlab/blob/66aabb72bf0fc26021868eb0d1a0555ef5b759ac/.meeseeksdev.yml#L5)

djangoliv:
@hbcarlos I would prefer the tags to be shared, but the real problem is removing them from the original file.

krassowski:
From the given the `editable` is a metadata field as well, and some extensions (like citation manager) use metadata for storing information essential for their functioning, I would expect the metadata to be shared. If we do not want specific items of metadata to be shared we can create a blacklist.

hbcarlos:
> Oh, so the issue was that the tag name didn't exist. I thought something might have been wrong with our [config](https://github.com/jupyterlab/jupyterlab/blob/66aabb72bf0fc26021868eb0d1a0555ef5b759ac/.meeseeksdev.yml#L5)

Yes, it was my fault. Thanks!

hbcarlos:
> I would expect the metadata to be shared. If we do not want specific items of metadata to be shared, we can create a blocklist.

Agree. We should, by default, share the metadata and remove the attributes we don't want to share.

hbcarlos:
Fixed in #11064

djangoliv:
@hbcarlos :
[...]

PR #11064, opened 2021-09-13
Share notebook's metadata

hbcarlos:
As discussed in #11038, by default the metadata is not shared because there are some attributes that we do not want to share. This makes every use of the metadata in the JupyterLab, not collaborative. In the case you want to share the metadata, you need to explicitly add the content to the shared model instead of modelDB. With this PR, we listen to changes in modelDB and apply them to the shared model to make everything collaborative.

In this PR we are facing one small problem. ModelDB allows modifying the attributes independently while the shared model doesn't. To update the modelDB on a remote client we need to update the whole metadata because is the update we receive from the shared model. Maybe we should change how we update the metadata in the shared model.

## References
Solves #11038

## Code changes
Listening for changes on modelDB metadata to update the shared model. The problem with this approach is that we need to update the modelDB on other clients and that triggers the change event. For this reason, we create some sort of a mutex to avoid updating again the client.

## User-facing changes
Now the metadata is shared between clients.

### DIFF
--- a/packages/notebook/src/model.ts
+++ b/packages/notebook/src/model.ts
@@ -447,11 +447,13 @@ close the notebook without saving it.`,
     change: IObservableMap.IChangedArgs<ReadonlyPartialJSONValue | undefined>
   ): void {
     console.debug('_onMetadataChanged');
-    console.debug(metadata.toJSON());
     console.debug(change);
-    this._modelDBMutex(() => {
-      this.sharedModel.updateMetadata(metadata.toJSON());
-    });
+    if (change.key !== 'kernelspec' && change.key !== 'language_info') {
+      this._modelDBMutex(() => {
+        console.debug(metadata.toJSON());
+        this.sharedModel.updateMetadata(metadata.toJSON());
+      });
+    }
     this.triggerContentChange();
   }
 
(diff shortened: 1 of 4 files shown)


packages/notebook/src/model.ts
```diff
@@ -428,6 +429,27 @@ close the notebook without saving it.`,
...
+        });
+      });
+    }
+  }
+
+  private _onMetadataChanged(
+    metadata: IObservableJSON,
+    change: IObservableMap.IChangedArgs<ReadonlyPartialJSONValue | undefined>
+  ): void {
+    if (change.key !== 'kernelspec' && change.key !== 'language_info') {
```
fcollonval, line 447: I would set that as constant at the module level:

```suggestion
    if ( !UNSHARED_KEYS.includes(change.key) ) {
```

fcollonval (commented):
I have only one suggestion to set the _unshared_ keys in a constant tuple at the module level.

packages/notebook/src/model.ts
hbcarlos: Hey @fcollonval, thanks for the suggestion. I applied the changes on my last commit.

fcollonval (approved):
LGTM

blink1073:

blink1073:

meeseeksmachine:

https://discourse.jupyter.org/t/jupyterhub-collaboration-and-tags/11259/3

PR #11065, opened 2021-09-13
Remove status bar item flickering

### LINKED ISSUE #9910: Log console statusbar icon classes are applied with a delay causing flickering
## Description

The statusbar icons added to the right part of the statusbar get moved to the right for a fraction of a second when switching sidebar tabs because of the icon with `jp-LogConsoleStatusItem` class getting its classes updated with a noticeable delay. Initially discovered in https://github.com/jupyterlab/jupyterlab-git/pull/893#issuecomment-792081091.

## Reproduce

1. Open the binder from https://github.com/jupyterlab/jupyterlab-git/pull/893
2. Follow the steps from the GIF below which demonstrates the issue and then force-hides the culprit icon with display: none to demonstrate that it is this icon that is causing the problem.

the-problem

## Expected behavior

The `jp-LogConsoleStatusItem` classes should be updated in the same frame so that no flickering of the position of icons from extensions is introduced when switching sidebar tabs.

## Context

Probably an issue with model/class switching logic in `render()` of https://github.com/jupyterlab/jupyterlab/blob/master/packages/logconsole-extension/src/status.tsx

- Browser and version: Chrome 89
- JupyterLab version: 3.0.7

fcollonval:
## References

Fix #9910

## Code changes

Set more appropriately `isActive` on status bar items.

## User-facing changes

No more flickering of the status bar items (except from kernel status changes)

N/A

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -303,7 +303,7 @@ export const savingStatusPlugin: JupyterFrontEndPlugin<void> = {
     statusBar.registerStatusItem(savingStatusPlugin.id, {
       item: saving,
       align: 'middle',
-      isActive: () => true,
+      isActive: () => saving.model !== null && saving.model.status !== null,
       activeStateChanged: saving.model!.stateChanged
     });
   }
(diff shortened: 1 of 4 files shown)


blink1073 (approved):
Thank you!

blink1073:

PR #11067, opened 2021-09-13
Split settings schema for inclusion in documentation

### LINKED ISSUE #11037: Menu item schema extraction in documentation is off
See "Item must follow this definition:" in https://jupyterlab.readthedocs.io/en/stable/extension/extension_points.html#context-menu being shifted off by several lines. It is trivial to updated, but opening the issue to get any suggestions on how to handle it better. This also affects the toolbar item definitions:

https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html#toolbar-registry (after "Add the toolbar item must follow this definition:")

fcollonval:
## References

Fixes #11037

## Code changes

Split the full settings schema to avoid positional changes in the file.

## User-facing changes

N/A

N/A


blink1073 (approved):
Thanks!

PR #11052, opened 2021-09-10
Constrain ipykernel version on CI

fcollonval:
## References

Fix CI UI tests

## Code changes

Downgrade ipykernel to <6.4.0

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

PR #11076, opened 2021-09-14
Revert "Constrain ipykernel version on CI"

jtpio:
Reverts jupyterlab/jupyterlab#11052 as this pin should not be needed anymore.


blink1073 (approved):
Thank you!

Issue #8148, opened 2020-04-01
UI UX Exploration: Settings User Interface.


tgeorgeux:
We currently have an advanced settings editor, but not a 'normal' one.

There is a team from UCI who's working on this with me, we'll post updates and questions here.

telamonian:
We absolutely need this. I like vscode's approach to presenting core and extension settings together in one scroll. I think anything we have should have roughly the same capabilities, if not exactly the same design:

![image](https://user-images.githubusercontent.com/2263641/78167547-ac7cfc80-741c-11ea-9cf4-1765d667ccd3.png)

telamonian:
One blocker on the technical side has been "how do we generate appealing forms from our existing settings .json files?" We currently don't have the necessary zoo of nice input components internally.


There was a brief push for us to adopt [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form):

#5892
#7250

@ian-r-rose actually got it working in an experimental branch, but it was never merged:

https://github.com/ian-r-rose/jupyterlab/tree/settings-form

In theory, react-jsonschema-form can generate fairly nice looking forms directly from our existing .json schema files:

![image](https://user-images.githubusercontent.com/2263641/78168779-82c4d500-741e-11ea-9bac-e46bea543651.png)

The upstream issues that stopped those earlier efforts have seemingly been resolved, so now may be a good time to revisit that work. If we do adopt react-jsonschema-form, it will still likely require a great deal of theming work in order to actually look nice.

fcollonval:
Hey just sharing a recent discovery here:
https://github.com/vazco/uniforms

It looks more interesting than react-jsonschema-form due to it's higher modularity especially regarding theming.

And this also brings back the question of a dedicated UI components library for jupyter.

telamonian:
@fcollonval uniforms looks neat

bollwyvl:
Howdy folks! Very glad this conversation has come back up.

The work that ian started was blatantly stolen by me and eventually became [jupyterlab-rjsf](https://github.com/deathbeds/jupyterlab-starters/tree/master/packages/jupyterlab-rjsf), extracted from the [parent repo](https://github.com/deathbeds/jupyterlab-starters) into its own [installable](https://www.npmjs.com/package/@deathbeds/jupyterlab-rjsf) package. I've shipped a couple products with it, and seen some folk do some pretty mind-blowing but alas, not open source, things with it.

[...]

PR #11079, opened 2021-09-14
Add settings UI

marthacryan:
## References
Uses #8148 as a design spec (with adjustments)

## Questions
- I added the shortcuts settings UI to show that you can register a new renderer but I think it's a great UI - should we just merge that with this PR? 
- Not sure about naming for this settings editor vs. the json settings editor (I'm keeping both because I think some people would want the classic JSON UI). Right now I'm calling it simple settings editor and json settings editor. 

## Todo
- [x] fix integrity
- [x] Add tests
- [x] Re-renders all of JL every time edits are made to the theme settings
- [x] Clean up validation
- [x] Check types and documentation
- [x] Move raw editor into pop out

## Todo in the future (but not blocking this PR)
- [ ] Add granular reset setting
- [x] Add tracker
- [ ] Add ui-schema to the settings schema
- [ ] Add better descriptions / titles for some fields
- [ ] separate settings by namespace
- [ ] Make renderer for main menu editor
- [ ] Add better searching

## Code changes
* Adds to the ui-components package:
  * A form component registry which allows for adding new renderers for fields in a form (based on an id)

* Adds `SettingsFormEditor` and `SettingsPanel` components to handle preparing the setting registry schema for each plugin to be input for the `Form` (package `rsjf`)

* Updates the schema to include default values so that the modified indicator doesn't show up for un-edited fields

## User-facing changes
* Adds a search bar to the left panel
* Separates modified plugins from un-touched plugins in the left sidebar

image

Allows for easily switching to the raw editor using a switch in the bottom left corner:
image

Validation errors highlight the field in red and display a message below:
image

Modified fields show a blue bar to the left of the field:
image

And plugins that contain a modified setting have a "restore to default" button next to the name of the plugin:
image

Doesn't change any settings API's


isabela-pf (commented):
@marthacryan and @ajbozarth thank you so much for your hard work on this! I'm really excited to see the settings UI in motion!

## Your questions
> The design spec above wants to separate extensions and core - how to distinguish?

This is a great question. Do you not want to put them in their own section (like the mockups [on this comment on #8148](https://github.com/jupyterlab/jupyterlab/issues/8148#issuecomment-683112403)) because extensions settings may not be so separate from the main settings in practice? Or is there some other blocker?

> Design spec doesn’t use adjustable width sidebar but my current implementation does, should I change that?

I like the adjustable side bar just fine. In fact, I think it might even be good because JupyterLab's responsiveness is always a little wacky (or I'm doing something wrong) and this might help people make sure it can be used better on a variety of screen sizes.

## Other thoughts

I know this is a draft PR, so please ignore these other comments if you are already aware.
- Some elements crashing together. I think a little more vertical spacing overall would solve a lot of what I'm finding.

Examples:
Notebook section of the JupyterLab settings with text overlapping in input boxes
Code Console section of the JupyterLab settings with text overlapping in input boxes


- Some text, mostly the `ADD ITEM`,  doesn't look like buttons until you interact with it. Would it be possible to add an icon (maybe Material Design's `add`) to try and make it clearer?

I'm talking about this button, in case you aren't sure:
Transparent Add Item button in JupyterLab settings

- I was really impressed by how changing a setting worked! Watching it pop up into the `Modified` section and making it so easy to reset felt good to me. Thanks for paying attention to these details. 💕 

JupyterLab settings with settings modified in the Terminal section and a Reset to Defaults button

## Wish list
I think these things might be difficult to make happen, and I'd much rather have a settings UI even as I'm seeing it in a draft state than have it blocked by any of these.
- Is there any way to have the shortcuts listed as a more user-friendly summary (like `Copy: ⌘ + C` format) by default, and then go to the view you have now when editing? Same for main menu items as well.

I'm talking about this as the current view:
JupyterLab settings keyboard shortcuts list as input fields

- Would it be reasonable to show the interface language with the language's full name instead of an abbreviation? 
- I need to give a closer review for accessibility here, but I'm happy to say color contrast is looking good so far. I did notice I can tab through sections of the settings—like input fields— just fine, but I couldn't manage to get my focus on the navigation or moving across sections/headers.
__

Thanks again for all your work. What I'm seeing here now is already a huge improvement!

goanpeca:
Thanks so much for working on this @marthacryan. Looks awesome. 

> @goanpeca How should the translator be used in the settings editor? Plugin names? Currently I have it translating the label, category, and the header, am I missing anything there?

I think you have got all things covered, but to be sure, I would advice to install the latest Chinese language pack, which is the most complete so far.

```
pip install jupyterlab-language-pack-zh-CN
```

Then we can be sure thing are working as expected.

Any other thoughts @fcollonval & @krassowski ?

bollwyvl:
Looking great! From the meeting: recommend adding a top-level `uiSchema` to the schemas which would allow a plugin to tweak their stuff without _having_ to know deep react via the [`rjsf` uischema](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/).

Presumbly, anything in the registry would be available, which is kinda fun.

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -49,178 +89,117 @@ export class PluginList extends Widget {
...
 
   /**
    * Check the plugin for a rendering hint's value.
    *
    * #### Notes
    * The order of priority for overridden hints is as follows, from most
    * important to least:
-   * 1. Data set by the end user in a settings file.
+   * import { showDialog } from '@jupyterlab/apputils';
+1. Data set by the end user in a settings file.
```
fcollonval, line 198: Sporious copy-paste?

```suggestion
   * 1. Data set by the end user in a settings file.
```

packages/settingeditor-extension/src/index.ts
```diff
@@ -45,6 +50,88 @@ namespace CommandIDs {
  * The default setting editor extension.
  */
 const plugin: JupyterFrontEndPlugin<ISettingEditorTracker> = {
+  id: '@jupyterlab/settingeditor-extension:simple-plugin',
```
fcollonval, line 53: We need a better name.

```suggestion
  id: '@jupyterlab/settingeditor-extension:form-ui',
```

packages/settingeditor/src/settingseditor.tsx
```diff
@@ -0,0 +1,99 @@
...
+import React from 'react';
+import { PluginList } from './pluginlist';
+import { SettingsPanel } from './settingspanel';
+import { Message } from '@lumino/messaging';
+import { Dialog, showDialog } from '@jupyterlab/apputils';
+
+export class SimpleSettingsEditor extends Widget {
+  translator: any;
+  //   private _settings: any[];
+  //   private _settingsPanel: any;
```
fcollonval, line 15:
```suggestion
```

packages/settingeditor/src/settingseditor.tsx
```diff
@@ -0,0 +1,99 @@
...
+  private _list: any;
+  private _panel: SplitPanel;
```
fcollonval, line 17: Could you place the attribute declaration at the bottom of the class to be consistent with JLab convention.

packages/settingeditor/src/settingseditor.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ITranslator } from '@jupyterlab/translation';
+import { IFormComponentRegistry, ReactWidget } from '@jupyterlab/ui-components';
+import { PanelLayout, SplitPanel, Widget } from '@lumino/widgets';
+import React from 'react';
+import { PluginList } from './pluginlist';
+import { SettingsPanel } from './settingspanel';
+import { Message } from '@lumino/messaging';
+import { Dialog, showDialog } from '@jupyterlab/apputils';
+
+export class SimpleSettingsEditor extends Widget {
```
fcollonval, line 12: We probably want to rename to something like `FormSettingsEditor` or `SettingsEditor`.

packages/settingeditor/src/settingseditor.tsx
```diff
@@ -0,0 +1,99 @@
...
+          super.onCloseRequest(msg);
+        }
+      });
+    } else {
+      super.onCloseRequest(msg);
+    }
+  }
+}
+
+interface IOptions {
```
fcollonval, line 77: Could you wrap this in a namespace that matches the name of the widget class to follow JLab convention?

packages/settingeditor/src/settingspanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+|----------------------------------------------------------------------------*/
+
+import { IFormComponentRegistry } from '@jupyterlab/ui-components';
+import { Settings } from '@jupyterlab/settingregistry';
+import { ISignal } from '@lumino/signaling';
+import React from 'react';
+import { SettingsFormEditor } from './SettingsFormEditor';
+import { PluginList } from './pluginlist';
+
+interface ISettingsPanelProps {
```
fcollonval, line 13: should be exported
```suggestion
export interface ISettingsPanelProps {
```

packages/shortcuts-extension/src/index.ts
```diff
@@ -50,17 +73,25 @@ import { DisposableSet, IDisposable } from '@lumino/disposable';
  */
 const shortcuts: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlab/shortcuts-extension:shortcuts',
-  requires: [ISettingRegistry, ITranslator],
+  requires: [ISettingRegistry, IFormComponentRegistry, ITranslator],
```
fcollonval, line 76: It is probably a good idea to make `IFormComponentRegistry` optional.

packages/shortcuts-extension/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": ["src/*"],
+  "include": ["src/*", "src/*/*"],
```
fcollonval, line 7: Small simplification

```suggestion
  "include": ["src/**/*"],
```

packages/ui-components-extension/src/index.ts
```diff
@@ -24,4 +28,17 @@ const labiconManager: JupyterFrontEndPlugin<ILabIconManager> = {
   }
 };
 
-export default labiconManager;
+/**
+ * Sets up the component registry to be used by the FormEditor component.
+ */
+const registryPlugin: JupyterFrontEndPlugin<IFormComponentRegistry> = {
+  id: '@jupyterlab/settingeditor-extension:registry-plugin',
```
fcollonval, line 35: I would use `form` instead of `plugin`

```suggestion
  id: '@jupyterlab/settingeditor-extension:form-registry',
```

packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -0,0 +1,195 @@
...
+          <h3> {settings.schema.description} </h3>
+        </div>
+        {isModified ? (
+          <button className="jp-RestoreButton" onClick={reset}>
+            {' '}
+            Restore to Defaults{' '}
+          </button>
+        ) : undefined}
+      </div>
+      {hidden ? undefined : (
```
fcollonval, line 173: Prefer something like this:

```suggestion
      {!hidden && (
```

or return `null` not `undefined`

packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -0,0 +1,195 @@
...
+        <div className="jp-SettingsTitle">
+          <h2> {settings.schema.title} </h2>
+          <h3> {settings.schema.description} </h3>
+        </div>
+        {isModified ? (
+          <button className="jp-RestoreButton" onClick={reset}>
+            {' '}
+            Restore to Defaults{' '}
+          </button>
+        ) : undefined}
```
fcollonval, line 171: Prefer something like this or return `null` (you should avoid returning `undefined` explicitly)

```suggestion
        {isModified && (
          <button className="jp-RestoreButton" onClick={reset}>
            {' '}
            Restore to Defaults{' '}
          </button>
        )}
```

packages/settingeditor/src/plugineditor.ts
```diff
@@ -18,7 +18,7 @@ import { Message } from '@lumino/messaging';
 import { ISignal, Signal } from '@lumino/signaling';
 import { StackedLayout, Widget } from '@lumino/widgets';
 import { RawEditor } from './raweditor';
-import { SettingEditor } from './settingeditor';
+import { SettingEditor } from './jsonsettingeditor';
```
fcollonval, line 21: It would be good to rename `SettingEditor` to `JSONSettingEditor` to make things more obvious

packages/settingeditor-extension/src/index.ts
```diff
@@ -45,6 +50,88 @@ namespace CommandIDs {
...
+
+      editor.id = namespace;
+      editor.title.icon = settingsIcon;
+      editor.title.label = trans.__('Settings');
+
+      const main = new MainAreaWidget({ content: editor });
+      void tracker.add(main);
+      shell.add(main);
+    },
+    label: trans.__('Simple Settings Editor')
```
fcollonval, line 125: Let's call it `Settings Editor` as this is certainly not simple :wink: 

```suggestion
    label: trans.__('Settings Editor')
```

fcollonval (changes requested):
Hey @marthacryan thanks a lot for this PR. Here are some comments.

It would go through the code and add documentation where missing. Do you need some help?

Big question: could we avoid explicitly setting attributes with default values (for shortcuts, menus,...)? This will be pretty annoying for extensions if required.

Otherwise, I think we should merge quickly to get it in an alpha. So that people can play with it and improve from there.

packages/metapackage/package.json
```diff
@@ -12,16 +12,22 @@
   },
   "license": "BSD-3-Clause",
   "author": "Project Jupyter",
-  "sideEffects": false,
+  "sideEffects": [
```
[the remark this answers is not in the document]
marthacryan: The integrity tests were automatically generating style files and then failing if they weren't listed in the side effects. Not sure why that was happening but I can try removing it to see if it happens again

packages/settingeditor-extension/src/index.ts
marthacryan: Agreed - any ideas?

marthacryan:
@fcollonval Thank you for all the feedback! I addressed most of it (still working on adding better documentation) but I'm not sure how to fix the integrity tests without including the style metapackage changes. Let me know if you're familiar with this issue / know what's going on there. 

Also, I removed the changes to schema defaults and I'm trying to update the isModified function to account for defaults that are being marked as modified right now. The issue is coming up with arrays with items that are complex types because the form editor is automatically adding values for the optional fields, but the default isn't. It hasn't been straightforward for me to check for those optional fields when comparing the default to the user defined values, I'm definitely open to help if you have suggestions there.

packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -0,0 +1,207 @@
...
+import React from 'react';
+import Form, { FieldTemplateProps, IChangeEvent, UiSchema } from '@rjsf/core';
+import { JSONSchema7 } from 'json-schema';
+import { JSONExt, ReadonlyPartialJSONObject } from '@lumino/coreutils';
+import { reduce } from '@lumino/algorithm';
+import { PluginList } from './pluginlist';
+import { ISignal } from '@lumino/signaling';
+interface IProps {
+  settings: Settings;
+  renderers: { [id: string]: React.FC };
```
fcollonval, line 17: Could you make the signature more precise? From rjsf it seems that the type is `React.FunctionComponent<FieldProps> | React.ComponentClass<FieldProps>`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ce3c84e10e30a3efa6078d5cda57f80c1d539886/types/react-jsonschema-form/index.d.ts#L216

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -49,178 +89,117 @@ export class PluginList extends Widget {
...
-   */
-  const ICON_LABEL_KEY = 'jupyter.lab.setting-icon-label';
 
   /**
    * Check the plugin for a rendering hint's value.
    *
    * #### Notes
    * The order of priority for overridden hints is as follows, from most
    * important to least:
+   * import { showDialog } from '@jupyterlab/apputils';
```
fcollonval, line 197: I think this is a noisy copy-paste

```suggestion
```

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -249,67 +228,140 @@ namespace Private {
...
+            ? 'jp-mod-selected jp-PluginList-entry'
+            : 'jp-PluginList-entry'
+        } ${this._errors[id] ? 'jp-ErrorPlugin' : ''}`}
+        data-id={id}
+        key={id}
+        title={itemTitle}
+      >
+        {id === this.selection || this._errors[id] ? (
+          <div className="jp-SelectedIndicator" />
+        ) : undefined}
```
fcollonval, line 275: Please try never setting `undefined` but prefer `null` to distinguish when we explicitly set missing value versus it comes from non-existent item resulting from JS _freedom_

```suggestion
        ) : null}
```

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -249,67 +228,140 @@ namespace Private {
...
+        <LabIcon.resolveReact
+          icon={icon || (iconClass ? undefined : settingsIcon)}
```
fcollonval, line 277:
```suggestion
          icon={icon || (iconClass ? null : settingsIcon)}
```

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -249,67 +228,140 @@ namespace Private {
...
+          useFuzzyFilter={true}
+          placeholder={'Search...'}
+          forceRefresh={false}
+        />
+        {modifiedItems.length > 0 ? (
+          <div>
+            <p className="jp-PluginList-header">Modified</p>
+            <ul>{modifiedItems}</ul>
+          </div>
+        ) : undefined}
```
fcollonval, line 312:
```suggestion
        {modifiedItems.length > 0 && (
          <div>
            <p className="jp-PluginList-header">Modified</p>
            <ul>{modifiedItems}</ul>
          </div>
        )}
```

packages/ui-components/src/FormComponentRegistry.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    // Any information needed to render a component (i.e. label, field type, etc.)
+    uihints?: any;
+  }
+}
+
+/**
+ * A registry for rendering fields used in the FormEditor component.
+ */
+export class FormComponentRegistry implements IFormComponentRegistry {
```
fcollonval, line 27: Could you avoid using `any` and add a doc string on all public elements here?

packages/ui-components/src/FormComponentRegistry.tsx
```diff
@@ -0,0 +1,77 @@
...
+  getRenderer(
+    id: string
+  ): (props: FormComponentRegistry.IRendererProps) => any {
+    return this._renderers[id];
+  }
+
+  private _renderers: { [id: string]: (props: any) => any } = {};
+}
+
+export interface IFormComponentRegistry {
```
fcollonval, line 63: Could you add documentation? - It can be duplicated from the class.

packages/ui-components/src/FormComponentRegistry.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ) => void;
+  getRenderer: (
+    id: string
+  ) => (props: FormComponentRegistry.IRendererProps) => any;
+
+  renderers: { [id: string]: (props: any) => any };
+}
+
+export const IFormComponentRegistry = new Token<IFormComponentRegistry>(
+  '@jupyterlab/settingeditor:ISettingEditorRegistry'
```
fcollonval, line 76: Missing rename following file move to `ui-components`

```suggestion
  '@jupyterlab/ui-components:ISettingEditorRegistry'
```

packages/ui-components/src/components/search.tsx
```diff
@@ -1,17 +1,16 @@
...
+import { InputGroup } from './inputgroup';
+import { ReactWidget } from './vdom';
 import { StringExt } from '@lumino/algorithm';
 import React, { useEffect, useState } from 'react';
-import { DirListing } from './listing';
 
 /**
  * The class name added to the filebrowser crumbs node.
  */
 export interface IFilterBoxProps {
```
fcollonval, line 12: Could you add documentation on the props?

fcollonval (changes requested):
Thanks for addressing the review comments.

> Also, I removed the changes to schema defaults and I'm trying to update the isModified function to account for defaults that are being marked as modified right now. The issue is coming up with arrays with items that are complex types because the form editor is automatically adding values for the optional fields, but the default isn't. It hasn't been straightforward for me to check for those optional fields when comparing the default to the user defined values, I'm definitely open to help if you have suggestions there.

Could that feature of ajv be helpful?  
https://ajv.js.org/options.html#usedefaults  
https://ajv.js.org/guide/modifying-data.html#assigning-defaults

It is already used in the settings registry: https://github.com/jupyterlab/jupyterlab/blob/master/packages/settingregistry/src/settingregistry.ts#L233

Regarding the documentation, the public API (hence interfaces, React component properties,...) should be documented as much as possible.

packages/metapackage/package.json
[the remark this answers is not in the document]
fcollonval: Actually the good way to fix this is to update the CSS skip list in https://github.com/jupyterlab/jupyterlab/blob/f65faadc14ee18b45d5ea06de6472d1b04cab5a0/buildutils/src/ensure-repo.ts#L145-L226

Thanks @jtpio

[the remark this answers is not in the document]
marthacryan: Ah that worked!! Thank you!

packages/settingeditor/src/pluginlist.tsx
marthacryan: Oh yeah I think my IDE might have done that

packages/settingeditor/src/pluginlist.tsx
marthacryan: This doesn't compile because of a type error

marthacryan:
```
Type 'string | LabIcon | null' is not assignable to type 'IMaybeResolvable'.
  Type 'null' is not assignable to type 'IMaybeResolvable'.ts(2322)
```

marthacryan:
Thank you for more feedback - I think I addressed most of it. The integrity seems to be working now, thank you!!

> Could that feature of ajv be helpful? https://ajv.js.org/options.html#usedefaults https://ajv.js.org/guide/modifying-data.html#assigning-defaults
The issue is that I think I'm already using the default that ajv is creating there when it creates the schema? I'm not totally sure though

packages/settingeditor/src/pluginlist.tsx
fcollonval: Ok then let be it `undefined`

fcollonval (changes requested):
@marthacryan i played around with the interface - I really think we should merge this one. Could you tackle the test failures? You will need to update how to change the settings in the one failing or to open the old JSON editor instead of the new one (but I think it will be better to take the chance of testing the new interface).

Other than that I have seen the following problems:

settings-ui

- Check box are not checked when clicking on the highlighted area. It would be good to switch the CSS `cursor` to `pointer` on the area on which the click is working.
- There is something to fix on the editor configuration for the notebook cells because only the subset of default that are different per cell types is shown; it looks like a bad idea to mix `$ref` and `properties` on the same object. The solution is maybe to let the `$ref` (removing the default values it contains) and provide a `default` value for the all properties of that object in each cell type.

krassowski (changes requested):
Thanks! It is a huge improvement for UX. I would like to see the following issues fixed before we merge it:

- [ ] Clicking on `HTML Viewer`, `CSV Viewer`, or `Keyboard Shortcuts` changes them instantly even though no user action was taken and also triggers a dialog; this should not happen (GIF 1)
- [ ] Clicking on items in the sidebar once items got moved to `Modified` does not select the item under the cursor (GIF 1)
- [ ] "Toolbar customization has changed" is too aggressive - it should have a debouncer/throttler or require a confirm action (GIF 2)
- [ ] Sidebar overrides do not allow to set key which is essential to add the override - or it is not obvious (GIF 3)

Question: is the old Advanced Settings Editor still available somewhere or was it removed?

Minor things which I believe can be addressed in follow up PRs (if at all):
- I personally would prefer to only display one item in the left panel rather than have this slightly odd combination of scrollable and clickable interface as this can be confusing at times.
- the margins/paddings and font sizes seem too large for me - this appears to be not very usable on smaller screens; we can iterate on this one in the future
- typestyle: recently bits of it were removed in other PRs; it is completely fine to merge it as is but we will need to discuss if we want to keep it as typestyle or maybe use some more developer-friendly backend-side CSS pre-processor instead (this is a matter of preference of course, nothing was decided yet)

GIF 1:
click-error

GIF 2:
invasive-info

GIF 3:
sidebar-cannot-choose-a-key

Regarding fixing up galata tests, in addition to the meeting with @fcollonval, please also feel free to me ping me on gitter if you have any specific issues - I ma be able to answer some of the simpler questions.

fcollonval (commented):
:tada: @marthacryan on getting all tests green on this one.

I pushed two minor changes (revert setting playwright worker = 1 and remove a unused variable). And we are good to go with it (if tests pass :crossed_fingers:)

marthacryan:
@krassowski Thank you for the feedback!! 
> * [ ]  Clicking on `HTML Viewer`, `CSV Viewer`, or `Keyboard Shortcuts` changes them instantly even though no user action was taken and also triggers a dialog; this should not happen (GIF 1)

This is a more complicated issue that I've been trying to work on but haven't quite figured out how to fix - basically for arrays of objects, if optional fields aren't manually added to each item in the default array (set in the schema) it registers the initial load of the form values (which its setting to the user value in the setting registry) as different from the default. For example in the HTML Viewer settings, the user value is 
```
"toolbar":[
    {
        "name":"refresh",
        "args":{},
        "command":"",
        "disabled":false,
        "rank":10
    },
    {
        "name":"trust",
        "args":{},
        "command":"",
        "disabled":false,
        "rank":20
    }
]
```
Whereas the default value, while functionally the same, is this:
```
"toolbar": [
        {
            "name": "refresh",
            "rank": 10
        },
        {
            "name": "trust",
            "rank": 20
        }
    ]
```
Because it's set to that here: https://github.com/jupyterlab/jupyterlab/blob/master/packages/htmlviewer-extension/schema/plugin.json#L5 
Any ideas are welcome!

> * [ ]  Clicking on items in the sidebar once items got moved to `Modified` does not select the item under the cursor (GIF 1)

I'll look into this, the scrolling has been a little weird since I added the accordion style, because it hasn't expanded the editor when it's scrolling to it. 

> * [ ]  "Toolbar customization has changed" is too aggressive - it should have a debouncer/throttler or require a confirm action (GIF 2)

I did a little research and it looks like it's popular to use lodash for debouncer/throttlers with react functional components - do you think that's the best route to take? 

> * [ ]  Sidebar overrides do not allow to set key which is essential to add the override - or it is not obvious (GIF 3)

I'm a little confused on this, according to the documentation this should be working: https://react-jsonschema-form.readthedocs.io/en/latest/usage/objects/#additional-properties but I can't get it to display the key name input. I'll keep looking into that though

> Question: is the old Advanced Settings Editor still available somewhere or was it removed?

Yes - it's under "JSON Settings Editor" in the command palette, and `cmd + shift + ,` should open it.

marthacryan:
@fcollonval Thank you! I think I addressed all of the feedback except for the last point in your comment:
> * There is something to fix on the editor configuration for the notebook cells because only the subset of default that are different per cell types is shown; it looks like a bad idea to mix `$ref` and `properties` on the same object. The solution is maybe to let the `$ref` (removing the default values it contains) and provide a `default` value for the all properties of that object in each cell type.

I'm not sure if I understand what you're describing, do you mind linking parts of the schema you're talking about?

krassowski:
> I did a little research and it looks like it's popular to use lodash for debouncer/throttlers with react functional components - do you think that's the best route to take?

Can `Debouncer`/`Throttler` from [`@lumino/polling`](https://lumino.readthedocs.io/en/latest/api/polling/globals.html) be used with function components of React? If not, what would we need to modify them to support that use case?

> Yes - it's under "JSON Settings Editor" in the command palette, and cmd + shift + , should open it.

That's great. Should we add a prominent button to the new Settings for now? We can remove it in the future, but especially given some suggestions that this could potentially be backported to 3.3.

Since there is a fallback to the old editor available, I'm happy with addressing the issue with the sidebar overrides in a follow up PR.

marthacryan:
> Can `Debouncer`/`Throttler` from [`@lumino/polling`](https://lumino.readthedocs.io/en/latest/api/polling/globals.html) be used with function components of React? If not, what would we need to modify them to support that use case?

Oh I wasn't aware of those - I'll try adding that in!

> That's great. Should we add a prominent button to the new Settings for now? We can remove it in the future, but especially given some suggestions that this could potentially be backported to 3.3.

Sure - so just to be clear, you're saying I should add a button into the JSON settings editor that opens the new settings editor? Not the other way around? 

> Since there is a fallback to the old editor available, I'm happy with addressing the issue with the sidebar overrides in a follow up PR.

Cool! I'll keep looking into it though, this might be a bug in the

krassowski:
>  just to be clear, you're saying I should add a button into the JSON settings editor that opens the new settings editor?

I meant the other way round (so that if something in the new settings does not work, e.g. for third-party extensions, the user can still easily go back to the old interface); sorry for poor wording.

fcollonval:
> > 
> 
> Can `Debouncer`/`Throttler` from [`@lumino/polling`](https://lumino.readthedocs.io/en/latest/api/polling/globals.html) be used with function components of React?

It seems that it is possible using `useCallback` to cache the callback definition between render update: https://thewebdev.info/2021/03/13/how-to-use-throttle-or-debounce-with-react-hook/

fcollonval:
> I'm not sure if I understand what you're describing, do you mind linking parts of the schema you're talking about?

If you look in the `Notebook` section, in the new UI only `lineWrap` and `matchBrackets` settings for every cell type are displayed. But the list of settings is longer:

image

It seems the bug is coming from the definition of default values for the only two parameters (that are different depending of the cell type) in
https://github.com/jupyterlab/jupyterlab/blob/fda5809404dd790b439ee48b05b9c5e0544004d6/packages/notebook-extension/schema/tracker.json#L707-L713

bollwyvl:
>  user can still easily go back to the old interface

Right: also, copy-paste of rich forms sounds horrible, unless it's already supported upstream... text is much cleaner and easier.

This is relevant as I assume this PR does _not_ provide an import/export (nor should it). 

Perhaps, though, it _might_ be worth leaving a (few) places for a future (third-party) extensions to to get closer to the #8148 mocks by registering additional UI elements, preferably as lumino widgets: 

- around the search box?
- below the sidebar list?
- each list item row?
- next to "restore defaults"

>  default values for the only two parameters

ruh roh:

> ```
> "lineWarp": { 
>     "default": "off" 
>  }, 
> ```

LET'S DO THE LINE WARP AGAAAIAIIAN!

bollwyvl:
As the status-quo was _Advanced Setting**s** Editor_, would this not just be the _Settings Editor_? Or maybe it could be made more active, e.g. _Edit all Settings..._

Screenshot from 2021-12-17 08-46-28

bollwyvl:
Looking at this next to the file manager search box, perhaps applying the same amount of padding would be appropriate?
Screenshot from 2021-12-17 08-49-31


> Ick, though, it's hard-coded pixels:

```
    margin: 8px 12px 0px 12px;
```

bollwyvl:
Further looking into the CSS: 
- it doesn't seem like the rules need to be hanging off an id `#setting-editor`
  - could be the already-created `.jp-SettingsForm`, or the not-yet-named sidebar 
- each of the settings forms needs an `idPrefix` so that URL anchors for deep links _could_ work
- `#setting-editor p ` should probalby _not_ hide any content, ever: if someone took the time to write it, it should get shown
  - one way to make slightly better use of vertical space would be to make simple group fields flex-row
  - there are a _ton_ of hard-coded pixel values:
    - some are base 4, some are base 5/10..
    - these could probably either 
      - use existing variables
      - be typeographically-derived (e.g. `0.5em`)
  - the use of the hover making the font bold on title is distracting
  - form buttons don't look great in dark themes
    -  long con, we need an rjsf theme that uses lab class names

marthacryan:
Added a button to open JSON editor: 
image

packages/settingeditor/style/base.css
```diff
@@ -148,96 +202,283 @@
...
+}
+.jp-SettingsPanel .array-item-move-up::after {
+  content: 'Move Up';
+}
+.jp-SettingsPanel .array-item-move-down::after {
+  content: 'Move Down';
+}
+.jp-SettingsPanel .array-item-remove::after {
+  content: 'Remove';
+}
```
krassowski, line 297: This is not blocking (if you do not plan to address it before we merge, let's create a new issue to track it), but:

> _Using an `::after` pseudo-element to add content is discouraged, as it is not reliably accessible to screen readers._

https://developer.mozilla.org/en-US/docs/Web/CSS/::after#accessibility_concerns

CC @isabela-pf

krassowski: Also, we won't be able to translate these strings if they are defined in CSS.

marthacryan:
@bollwyvl Thanks for the feedback! 
> * it doesn't seem like the rules need to be hanging off an id `#setting-editor`
>   
>   * could be the already-created `.jp-SettingsForm`, or the not-yet-named sidebar

Done

> * each of the settings forms needs an `idPrefix` so that URL anchors for deep links _could_ work

Done

> * `#setting-editor p ` should probalby _not_ hide any content, ever: if someone took the time to write it, it should get shown

Done

>   * there are a _ton_ of hard-coded pixel values:
>     
>     * some are base 4, some are base 5/10..
>     * these could probably either
>       
>       * use existing variables
>       * be typeographically-derived (e.g. `0.5em`)

I'll work on that

>   * the use of the hover making the font bold on title is distracting

Done

>   * form buttons don't look great in dark themes

Done

>     * long con, we need an rjsf theme that uses lab class names

We can make an issue for this

packages/settingeditor/style/base.css
marthacryan: Oh great point - there is a way around this, I can try to get that into this PR but if not I can definitely make an issue

marthacryan:
@fcollonval I was able to reproduce the issue with fields not showing up in the editor config fields - just pushed a fix, let me know if that's not working. Thanks! BTW I'm still looking at the throttler stuff but I'm hitting some issues

packages/settingeditor/style/base.css
marthacryan: @krassowski Just pushed a change that should address this issue. Let me know if there are any issues

marthacryan:
@fcollonval I figured out the issue with the debouncer! Maybe you could try it out and make sure it's working as expected?

marthacryan:
Bug I just noticed: not sure how to reproduce, but there have been a couple of times when I was unable to open the json settings editor because the tracker was registering the widget as already being open. I've fixed it by resetting my workspace but this should be investigated

marthacryan:
Would people be against removing the main menu editor from the new settings editor for now? It takes so long to load that I would worry that people could accidentally click it and have to wait for it to load.

marthacryan:
Just pushed a commit that hard-codes removing the context menu/ main menu settings in the new settings editor because of the issue described above^^. Users can still edit these settings in the json editor.

isabela-pf:
I know it's been a while since I last gave this a review, but I'm still surprised by how much work has been done. It's looking really clean! For the most part, It looks good to me. There are a few things I still wanted to ask about.

Selected Theme in the JupyterLab Settings Editor as a freeform input box.

Most of the input component choices made sense to me. It seems like it'd be very easy to have an error in typing out a theme name, though. Is there a reason this isn't a drop down of installed themes?

Dialog requesting user reload the page to update JupyterLab toolbars.

I kept getting this dialog (even when I didn't think I was changing the toolbars) when trying this out on Binder. I see there's been some toolbar-related work discussed above, so it may be related to that? If it is supposed to be something a user will see, it'd be nice to have a more descriptive title like `Reload page` or `Toolbar customized`. 

A non-default setting in the JupyterLab Settings Editor

I really like having a non-default setting labelled at the smallest unit with the blue bar. I don't remember this from last time I checked, and I liked to see that edited settings were easy to find from the high-level section all the way to the setting-by-setting level.

This is the kind of thing where so much has been done I expect to keep finding edge cases for a while, so I don't have any blocking feedback. Thanks for sticking with this work!

marthacryan:
@isabela-pf Thank you for the feedback! I think I addressed it but it'd be great if you could check again to make sure you're not seeing any of those same issues. Also I added a theme dropdown!

isabela-pf:
Thanks for the ping @marthacryan! Here's what I'm seeing
- [x] Settings editor/Theme/Selected Theme shows a drop down of theme options.
- [ ] ??? Update the dialog requesting a reload. This has a question mark because I can't get it to happen again. I think this means the underlying issue may have been solved. I left the question marks since I can't confirm, but happy to check this off.

From my earlier comment:
- [x] Make it so that UI no longer overlaps
- [x] Add item looks more like a button
- [x] Shortcuts listed to match UI (not a JSON list)
- [ ] Full language name instead of abbreviation
- [x] Keyboard navigation and color. A manual and not-too-long review of this seems like it accounts for some of the big accessibility things I keep watch for. I can navigate with my keyboard (though there's no visible focus, but that's a larger JLab problem) and color contrast is being met with interactive components. Things are labelled properly and persistently.

In short, this looks good to me. Nothing that's left here should block merging this PR to me. We can always return to this as we get user feedback once this is merged. 

It's not good practice accessibility-wise for me to flood this comment with celebratory emojis, but know that I want to! Thanks for taking all this feedback and cleaning this up so well! I am in awe! 🎉 ✨ 🌈 🌻 Thank you thank you thank you!

marthacryan:
@isabela-pf Thank you! I checked off the dialog thing because I added a fix for that.

fcollonval:
@marthacryan 

I went ahead a push some small improvements:
- Fix issue with some sections icons (see for example the console icon below)
- Wrap the editors in a main area widget (to better integrated in term of look and feel in the dockpanel)
- Translate more strings (pushing the translator deeper in the setting editor and the shortcut editor)
- Move the open JSON editor to a button toolbar and use the status bar for the save notification (so that the editor feels better integrated)

| before | after |
| --- | --- |
| before | after |

fcollonval:
UI test failure is not related

Check release failure is tracked in #11933

fcollonval (approved):
Thanks a lot @marthacryan

jasongrout:
Thank you so much, Martha and everyone else involved in this! This is an incredible amount of work.

When I looked at the UI for the first time, I didn't understand that it was an accordion-like widget and I was opening sections of the settings as I clicked on things. Would you consider adding some small disclosure triangles, like in the HTML details element, to indicate that each settings heading can be expanded? Something like

```diff
diff --git a/packages/settingeditor/src/SettingsFormEditor.tsx b/packages/settingeditor/src/SettingsFormEditor.tsx
index 30d74ccf4c..7f0be7a07c 100644
--- a/packages/settingeditor/src/SettingsFormEditor.tsx
+++ b/packages/settingeditor/src/SettingsFormEditor.tsx
@@ -340,7 +340,7 @@ export class SettingsFormEditor extends React.Component<
           }}
         >
           <div className="jp-SettingsTitle">
-            <h2> {this.props.settings.schema.title} </h2>
+            <h2> {this.state.hidden ? "▶" : "▼"} {this.props.settings.schema.title} </h2>
             <h3> {this.props.settings.schema.description} </h3>
           </div>
           {this.state.isModified && (
```
(except those triangles seem too big, and they aren't quite the same width so the text shifts slightly as we toggle it).

fcollonval:
That is a good suggestion @jasongrout 

i would suggest reusing the caret icon we have for homogeneity with the UI.

```ts
const icon = this.state.hidden ? caretRightIcon : caretDownIcon;

// ...

<h2> <icon.react tag="span" elementPosition="center" /> {this.props.settings.schema.title} </h2>
```

The svg size will certainly needs to be adapted.

marthacryan:
Thanks @jasongrout! I can definitely add that

marthacryan:
Updated:
image
image

krassowski (approved):
There was a huge progress on this PR, thank you @marthacryan. I have three questions on this PR. I would appreciate answers before we merge it, but I am **not** requesting any changes (merging as-is and iterating later is fine for me):

### 1. Accordion - do we need it? Why?

I also had the same experience as described by @jasongrout:

> When I looked at the UI for the first time, I didn't understand that it was an accordion-like widget and I was opening sections of the settings as I clicked on things.

The addition of the caret icons does help, but I would like to take a step back: why is there an accordion-like widget in the first place? I remember that it was discussed on one of the meetings, but I could not piece it back together now.

It feels to me that the accordion is sub-optimal and deteriorating the UX:
- clicking on an item in the sidebar does scroll to the relevant section but depending on the section position the section may be on top, or in the middle, or bottom of the screen making it difficult to find the section we clicked
- we render a lot more widgets than we need to render at a time
- it is difficult to navigate the settings because everything is on the screen and a lot of scrolling is needed; user may find themselves in a section other than they intended because a lot of scrolling is needed (see also point 2).

### 2. Styling: could we compactify everything and strengthen the visual hierarchy?

On the styling side I feel that the sections, margins and layout taking a lot more space than it needs:
- the margins are wide
- fonts seem bigger,
- input fields seem larger

when compared with other sections of the JupyterLab UI.

It also feels like the settings are not dense enough - I have to scroll a lot to find the option of choice, even on a very big 4K screen. To demonstrate this, on the same vertical space as the file browser shows 36 files, the "Theme" section shows barely 10 settings.

Screenshot from 2022-01-29 15-36-20

Obviously I am not advocating to make it as compact as in the file browser, but just a little bit. Given how much whitespace remains unused to the right:

| on 4K screen | on full HDscreen |
|---|---|
| Screenshot from 2022-01-29 15-39-28 |   Screenshot from 2022-01-29 15-40-03-1 |

what do you think about reducing the margins and making the forms responsive (side-by-side layout) on larger screens?

Further, the visual hierarchy is confusing making the navigation in the accordion layout difficult: the legend "Theme CSS Overrides" subtitle is as big as the "Theme" section title (both are `font-size: var(--jp-content-font-size3)` but the legend which is lower in hierarchy has stronger weight (it is bold) making it completely confusing user when accordion sections are expanded. This problem goes away if we decide that accordion is not needed (point 1), but otherwise we could just make the legend use `size2`.

For example the mockup below (implemented via very simple CSS changes) would make a huge difference for UX in my humble opinion:

Screenshot from 2022-01-29 16-14-50

This is a strong opinion. I feel strongly that compressing the UI and following the [visual hierarchy](https://en.wikipedia.org/wiki/Visual_hierarchy) principles will lead to substantial improvement in the UX.

### 3. The search bar: should we hide it for now?

The search bar currently is quite limited. It only searches among the section titles. I expected it to search for particular setting names too. Based on the PR description I understand that the intention is to add better search in the future. If we are going to backport this to 3.3, a misleading feature (search which works contrary to user expectation and will incorrectly tell the user that there is no setting they were looking for) will likely lead to worse UX than no search bar at all. As long as the PR was only targeting 4.x I had no issue with this, but since we want to ship it soon, what do you think about hiding the search bar for now?

This is **not** a strong opinion - just a suggestion for playing it safe.

### 4. Minor styling notes (this is all slightly subjective):
- the rounded corners: I think that we do not use them elsewhere on input fields. Should we keep them straight or reduce the radius to keep the look-and-feel more consistent across application?
- the fields are in bold but checkboxes are in normal; should we stick to one weight (as on my mock-up)?
- keyboard shortcuts UX feel jittery because of how width-constrained it is; could we make it full width?
- This piece of keyboard shortcuts UI feels wrong due to the width (/flex) constraint too
   Screenshot from 2022-01-29 16-21-06
   on the upstream repo it used to render as:
   Screenshot from 2022-01-29 16-24-27
   which I take was intended to be an explanation of the names and symbols but it kind of feels like meaningless hieroglyphs when squished like that; we need to give it some horizontal breathing space,

bollwyvl:
> Accordion - do we need it? Why?

The initial reason for the accordion was due to performance: we can't confidently render approaching 100 rjsf instances (the menu ones, in particular) take a few seconds.

> The search bar: should we hide it for now?

I feel like to be useful, it will eventually have search on anything that might appear, but might not even be rendered yet, such as schema `title`, `description`, _and_ raw names (as some "humane" fields... aren't yet). Further, the actual values, both in defaults, and user values. This is _more_ complex than a notebook document.

> Styling: could we compactify everything and strengthen the visual hierarchy?

Many of the space issues come from the basically un-themed state of the "no opinion" theme from rjsf. Really taking control of this, as has been done for material and bootstrap (or even blueprint, shudder), is rather a large undertaking, and while a `@jupyterlab/rjsf-theme` would be nice at some point, it represents a _lot_ of investment and on-going maintenance, and perhaps more react-centricity than we care to have.

In the near term, it's possible to get a somewhat consistent look by styling the fairly well-differentiated HTML underneath, a la [`jupyterlab-rjsf`](https://github.com/deathbeds/jupyterlab-starters/tree/master/packages/jupyterlab-rjsf/style).

krassowski:
> The initial reason for the accordion was due to performance: we can't confidently render approaching 100 rjsf instances (the menu ones, in particular) take a few seconds.

Thanks for reminding! In that case rendering only one section (and hiding all other sections and the accordion altogether) would still solve the initial problem, and potentially lead to a better UX?

fcollonval:
> In that case rendering only one section (and hiding all other sections and the accordion altogether) would still solve the initial problem, and potentially lead to a better UX?

Thanks @krassowski I pushed 24902ee to open only one plugin at a time. This indeed reduces the length that need to be scrolled.

This may be revisited when the search box will target the parameters too.

fcollonval:
Merging this one. So we can get a fully featured 3.3 pre-release.

fcollonval:

PR #11091, opened 2021-09-16
Fetch translations via the `ServerConnection.ISettings`

jtpio:
## References

Other distributions like [JupyterLite](https://github.com/jupyterlite/jupyterlite/) use the `ServerConnection.ISettings` to route requests to a different place than the regular Jupyter Server backend. In the case of JupyterLite they go to an alternative Jupyter Server implementation running in the browser.

This change ensures the translation requests also go through the `ServiceManager`'s `ServerConnection.ISettings`, so alternate downstream application can have more control on the behavior and add support for localization.

## Code changes

Make the translation plugin fetch translation data using the `ServerConnection.ISettings`.

## User-facing changes

Enable users of downstream distributions like JupyterLite to switch their UI to other languages.

New parameters are added as optional arguments, so this should be backward compatible so we can backport to 3.1.x.

### DIFF
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -50,9 +50,11 @@ const translator: JupyterFrontEndPlugin<ITranslator> = {
     const displayStringsPrefix: boolean = setting.get('displayStringsPrefix')
       .composite as boolean;
     stringsPrefix = displayStringsPrefix ? stringsPrefix : '';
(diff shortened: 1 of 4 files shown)


jtpio:
> New parameters are added as optional arguments, so this should be backward compatible so we can backport to 3.1.x.

Adding to the 3.1.x milestone based on this observation.

blink1073 (approved):
Thanks!

blink1073:

PR #11092, opened 2021-09-16
Update the lock after every request

hbcarlos:
Follows #11026

Yesterday during the JupyterLab meeting, the binder instance for the meeting notes didn't work.
When we tried to open the document with the meeting notes, the document widget showed a spinner for a long time while trying to retrieve the document's content. This is due to not releasing the lock properly.

The problem comes from #11026, where I added a new request to update the lock when a client is too slow, but I was not returning the new lock to the client.

## Code changes
Returns the lock to the client and updates the lock

## User-facing changes
They won't get stuck trying to retrieve the content of the document.

N/A

### DIFF
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -60,11 +60,11 @@ export class WebSocketProviderWithLocks
       messageType
     ) => {
       // acquired lock
-      const timestamp = decoding.readUint32(decoder);
+      this._currentLock = decoding.readUint32(decoder);
       const lockRequest = this._currentLockRequest;
       this._currentLockRequest = null;
       if (lockRequest) {
-        lockRequest.resolve(timestamp);
+        lockRequest.resolve(this._currentLock);
       }
     };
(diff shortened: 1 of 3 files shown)


packages/docprovider/src/yprovider.ts
```diff
@@ -190,6 +190,8 @@ export class WebSocketProviderWithLocks
    * Returns a Promise that resolves to the lock number.
    */
   acquireLock(): Promise<number> {
+    // keep two locks the first and the second
```
dmonad, line 193: Not sure about this comment. What do you mean?

hbcarlos: Ouch. Nothing some notes I wrote for me, and I forgot to remove them.

blink1073:
Let's merge and iterate on this

blink1073 (approved):
Thanks!

blink1073:

hbcarlos:
Thanks, @blink1073!!!

PR #11108, opened 2021-09-21
Improve the reactive toolbar

### LINKED ISSUE #11102: Toolbar popup is not cleaned
## Description

When loading the v4 alpha10 (but also on master), the toolbar popup widget is not remove from the DOM:

image

Moreover, each time you open a new main area panel, a new div element will be added to the DOM.

## Reproduce

1. Launch JupyterLab v4 alpha or master
2. Open the DOM tree
3. Open any main area panel
4. See that a new node `<div class="lm-Widget p-Widget jp-Toolbar-responsive-popup lm-mod-hidden p-mod-hidden"></div>` appeared

## Expected behavior

Toolbar popup are present only for the main area widgets and are removed when the main widgets are close.

fcollonval:
## References

Fixes #11102
Fix also a small glitch seen when the pop up is opened and the user switches tabs:

| Before | After |
| --- | --- |
| toolbar_hidden_before | toolbar_hidden_after |

## Code changes

Create `ReactiveToolbar` class that subclass `Toolbar` to easily choose between a toolbar with a pop up or not; the small toolbars like in the debugger or in the filebrowser are `Toolbar` but the ones for document widget are `ReactiveToolbar`.

Dispose the pop-up with the opener button.

This splits a class but it was not released on the 3.1.x branch. So we are fine.

### DIFF
--- a/packages/apputils/src/mainareawidget.ts
+++ b/packages/apputils/src/mainareawidget.ts
@@ -2,7 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
-import { Spinner, Toolbar } from '@jupyterlab/ui-components';
+import { ReactiveToolbar, Spinner, Toolbar } from '@jupyterlab/ui-components';
 import { Message, MessageLoop } from '@lumino/messaging';
 import { BoxLayout, BoxPanel, Widget } from '@lumino/widgets';
 import { DOMUtils } from './domutils';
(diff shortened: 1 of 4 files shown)


fcollonval:
@jtpio would you mind reviewing this PR?

jtpio (approved):
Looks good on Binder, thanks!

3coins:
@fcollonval 
Thanks for making the responsive toolbar better. I am working on moving responsive toolbar changes to 3.1 and 3.2, I will add these updates as well.

blink1073:
We can't target 3.1 since the reactive toolbar involves API changes, but I support 3.2

3coins:
> We can't target 3.1 since the reactive toolbar involves API changes, but I support 3.2

Got it. I am working on 3.2 then. Thanks.

PR #11112, opened 2021-09-22
Relax `@playright/test` dependency in Galata

jtpio:
## References

So we can pick up minor releases of `@playwright/test`.

## Code changes

Dependency update in the galata package.

## User-facing changes

None

None


jtpio:
cc @fcollonval looks like it should be fine to relax that dependency?

fcollonval:
> cc @fcollonval looks like it should be fine to relax that dependency?

The error are coming from the fix of a bug I reported on the unique name and are expected - some reference screenshots need to be updated.

Pinning the version was done for two reasons:
- Playwright is highly active and some regressions like the ones seen here are to be expected when changing version.
- The browsers are changing with Playwright version. So this can be a source of regression too.

I'm not against relaxing the constrain.

jtpio:
Thanks for the context.

If this brings too much trouble with sudden regressions, then I guess we could stick to a strict pin. Or allow for patch releases (`~`).

fcollonval (approved):
LGTM - just need to update the reference screenshots as the unique name changes in Playwright 1.15

jtpio:
> The error are coming from the fix of a bug I reported on the unique name and are expected

For reference and extra context, this issue and PR seem relevant: 

- https://github.com/microsoft/playwright/issues/8276
- https://github.com/microsoft/playwright/pull/8282

jtpio:
Not sure it's a regression in `1.15.0`, but the report shows the same image for the actual, expected and diff screenshots too:

https://user-images.githubusercontent.com/591645/134369201-a01396c6-f924-4d32-909d-d36bd7a21d17.mp4

Or it might be a misconfiguration on our side.

jtpio:
The UI Tests check should now be passing.

fcollonval:
Thanks a lot @jtpio 

This looks like a bug on their side as the other two images are not copied in the report folder.

fcollonval:
Should we wait for 1.16 (the bug has been added to that milestone)?

jtpio:
> Should we wait for 1.16 (the bug has been added to that milestone)?

Sounds good. I'll convert the PR to draft in the meantime.

Linking to the upstream issue for reference: https://github.com/microsoft/playwright/issues/9104

Issue #10874, opened 2021-08-19
Auto-close brackets does not work in console


gioarma:
## Description

After turning on `Auto Close Brackets`, the feature works in the notebook, but not in the console.

## Reproduce

1. Open a new notebook
2. Click on `Settings -> Auto Close Brackets` to turn on the feature
3. Type `(` in the notebook and `()` automatically appears
4. Open a new console for the notebook and repeat 3.  `()` does not appear

## Expected behavior

Auto Close Brackets works also in console

## Context

- Operating System and version: macOS 10.15.7
- Browser and version: Chrome 92.0.4515.159
- JupyterLab version: 3.1.7

eemilhaa:
Was able to reproduce this with:
- Operating system: Linux, Pop_os 21.04
- Browser and version: Firefox 91.0 (64-bit)
- JupyterLab versions 3.1.4 and 3.1.7

Also noticed that disabling/enabling the setting once the console window has been opened seems to work

Dharanikanna:
Hello All,
I'm Dharani Kanna. I have interested to Contribute and resolve this issue. I'm just beginner for open source contribution. Please some one guide me to catch the path of a open source contributor.

jasongrout:
@Dharanikanna - our new contributor guide helps you get a development environment set up: https://jupyterlab.readthedocs.io/en/stable/developer/contributing.html

Dharanikanna:
Thank you @jasongrout sir.

lucabarcelos:
I would like to work on this issue.

jasongrout:
Excellent! Please post a comment here with the PR link when you submit the PR.

ohrely:
I will work on this issue.

gioarma:
With version 3.2.3 the issue is solved only for the first time a cell is executed in the console.

Once the first cell is executed, if I try to type something new in the console I see the exact same problem. Do you see this as well?

fcollonval:
I can confirm that it works only on the first executed block. Then the setting is not effective any longer.

--> re-opening

pree-T:
@fcollonval  May I work on this? Can anyone help me a bit?

fcollonval:
> FYI you don't need to ask to start working on an issue. The best is to open early a draft PR with what you are doing and ask for review and/or help.

For this one, the path I would take is to save the editor options as a new attribute in the `CodeConsole` class. 

[...]

PR #11137, opened 2021-09-24
Fix auto close brackets for console

ohrely:
## References

https://github.com/jupyterlab/jupyterlab/issues/10874

## Code changes

Adds listener to apply settings when a console is created.  Additionally, modifies `updateSettings` to update just one widget when specified.

n/a

### DIFF
--- a/packages/console-extension/src/index.ts
+++ b/packages/console-extension/src/index.ts
@@ -394,9 +394,7 @@ async function activateConsole(
     if (panel) {
       setWidgetOptions(panel);
     } else {
-      tracker.forEach(widget => {
-        setWidgetOptions(widget);
-      });
+      tracker.forEach(setWidgetOptions);
     }
   }
 
--- a/packages/console-extension/src/index.ts
+++ b/packages/console-extension/src/index.ts
@@ -391,6 +391,11 @@ async function activateConsole(
   });
   await updateSettings();
 
+  // Apply current settings when a console is created.
+  tracker.widgetAdded.connect((sender, panel) => {
+    updateSettings();
+  });
+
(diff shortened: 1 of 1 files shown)


packages/console-extension/src/index.ts
```diff
@@ -391,6 +391,11 @@ async function activateConsole(
   });
   await updateSettings();
 
+  // Apply current settings when a console is created.
+  tracker.widgetAdded.connect((sender, panel) => {
+    updateSettings();
```
jasongrout, line 396: Thanks! You probably noticed that `updateSettings` cycles through each widget and updates settings for each one. What do you think about changing the code so that when a single widget needs an update, like in this case where we have one widget that was added, we only update that single widget rather than updating every widget?

ohrely: Testing that change now 😄

packages/console-extension/src/index.ts
```diff
@@ -374,26 +374,42 @@ async function activateConsole(
...
+    const setWidgetOptions = (widget: ConsolePanel) => {
       widget.console.node.dataset.jpInteractionMode = interactionMode;
       setOption(widget.console.promptCell?.editor, promptCellConfig);
-    });
+    };
+
+    if (panel) {
+      setWidgetOptions(panel);
+    } else {
+      tracker.forEach(widget => {
```
jasongrout, line 397: Small suggestion, take it or leave it: can we just use the function directly here and simplify this to `tracker.forEach(setWidgetOptions)`?

blink1073 (approved):
Thank you!

blink1073:

PR #11141, opened 2021-09-24
Normalize notebook cell line endings to \n

### LINKED ISSUE #9465: Code Editor Out of Sync error while trying to edit some code in a cell in Jupyterlab
Hello,

I used Google Colab to write some code, I downloaded Jupyterlab a couple hours ago and opened the same code in jupyter lab. However, when I try to edit a cell, an error message comes up saying "Code Editor Out of Sync". This is the error that prints on the javascript console.

(I know the "google.colab.patches" is probably not available, I was trying to delete that part, but the error kind of interferes with it.)

{"model":"#show (and save for report) images and show the beginning (first 10) of the flattened image arrays\r\n\r\nfor i in range(0, len(imagesDS t\n ry: t\n      cv2.imshow(imagesDS[]#)\r\n  exep #t:\r\n    #I added this part because I am using google collaboratory and they disabled cv2.imsho#w()\r\n    #it will not be run if not using google collabora#tory\r\n    from google.colab.patches import cv2_i#mshow\r\n    cv2_imshow(imagesDS[i])\r\n  #save the downsampled images for report\r\n  cv2.imwrite(imageNames[i].split(\".\")[0]+\"2.png\", imagesDS[i])\r\n","view":"#show (and save for report) images and show the beginning (first 10) of the flattened image arrays\n\nfor i in range(0, len(imagesDS)\n try: t\n      cv2.imshow(imagesDS[]#)\n  exep #t:\n    #I added this part because I am using google collaboratory and they disabled cv2.imsho#w()\n    #it will not be run if not using google collabora#tory\n    from google.colab.patches import cv2_i#mshow\n    cv2_imshow(imagesDS[i])\n  #save the downsampled images for report\n  cv2.imwrite(imageNames[i].split(\".\")[0]+\"2.png\", imagesDS[i])\n","selections":[{"uuid":"a21f53ac-f30f-4d09-85e2-a076b21eacbd","start":{"line":3,"column":2},"end":{"line":3,"column":2},"style":{"className":"","displayName":"","color":"black"}}],"cursor":{"line":3,"column":2},"lineSep":null,"mode":"text/x-ipython","change":{"from":{"line":3,"ch":1,"sticky":null},"to":{"line":3,"ch":1,"sticky":null},"text":["t"],"origin":"+input","removed":[""]}}

jasongrout:
## References

Fixes #9465

## Code changes

This normalizes any cell’s text to have \n line endings instead of \r\n line endings. Limitations:

1. This commit silently changes line endings without indicating that the file will be changed when saved.
2. This commit does not handle \r line endings, which I think were only default on old mac computers? See https://en.wikipedia.org/wiki/Newline.

### DIFF
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -196,10 +196,11 @@ export class CellModel extends CodeEditor.Model implements ICellModel {
     trusted.set(!!cell.metadata['trusted']);
     delete cell.metadata['trusted'];
 
+    // Set the text value, normalizing line endings to \n
     if (Array.isArray(cell.source)) {
-      this.value.text = (cell.source as string[]).join('');
+      this.value.text = cell.source.map(s => s.replace(/\r\n/g, '\n')).join('');
     } else {
-      this.value.text = cell.source as string;
+      this.value.text = cell.source.replace(/\r\n/g, '\n');
     }
     const metadata = JSONExt.deepCopy(cell.metadata);
     if (this.type !== 'raw') {


jasongrout:
Not sure which milestone to put this in (master is on 4.0, right?). It would be great if it can be backported to at least 3.0.x

jasongrout:
Making draft since I'd love some feedback on:

1. Are there unintended consequences with converting line endings right here, where cells models are created, especially with the new RTC stuff, or is this the right place to do it? Another place to do this conversion is at https://github.com/jupyterlab/jupyterlab/blob/05a2fa047c5416ae1dd6a6cdac165c13a3513f47/packages/notebook/src/model.ts#L283 but I figured that since the problem here is a limitation of how our cell model and widget are interacting, it's probably better to normalize at the cell creation.
2. I'd love to figure out how to show the file is dirty. Even if we set the model to dirty in the above place, leftover code at https://github.com/jupyterlab/jupyterlab/blob/05a2fa047c5416ae1dd6a6cdac165c13a3513f47/packages/docregistry/src/context.ts#L687 automatically sets the dirty state to false even after we do a fromJSON. Perhaps in the above location, we could do a toJSON immediately after a fromJSON to check to see if the file immediately has changes?

hbcarlos:
Hi @jasongrout, thanks for working on this.

Regarding 1, I think it is okay in both places, but I would do it in the `fromJSON()` method since we have the entire content there, and we can also check the metadata.

Regarding 2, I'm not sure; I always have trouble with the `dirty` attribute when working with document models.

jasongrout:
@ohrely and I dove deep into both the issues listed above (where the conversion should happen, and how to deal with dirty state) and both of us feel that the current code (converting in the cell model, and not dealing with dirty state issues) is the right choice for this PR. Here is our reasoning:

1. Where should the conversion happen? Because the cell is where the actual problem is (i.e., where the mismatch between locations in the model and editor cause a problem), and because the cell npm package can be used outside of JupyterLab (like on a page by itself as a single cell), we feel the fix should be at the cell level, not at the notebook document serialization level.
2. How to deal with dirty state? We dove into it, and realized there are quite a few issues with the dirty state management. For example, code at https://github.com/jupyterlab/jupyterlab/blob/05a2fa047c5416ae1dd6a6cdac165c13a3513f47/packages/docregistry/src/context.ts#L687 insists that no matter what happens at in the fromJSON function, the model will always be clean when creating it, and when we comment that line out, the last line in the notebook fromJSON function insists that the model is always dirty! Commenting even that out leads to multiple times when the cell operations force the model state to be dirty on creation, and an end result is that the model state is *always* considered dirty. In any case, the dirty state messaging issues are much deeper problems with larger ramifications, and I think they deserve their own PR and discussion. For now, we feel like this bugfix for a current show-stopping issue outweighs possible confusion about dirty state not being updated in this automated conversion.

In other news, the VS Code bug on this at https://github.com/microsoft/vscode-jupyter/issues/4576 was closed today, so hopefully this issue in the wild will be a moot point. I still think it is important to merge this, though, for the next client that does this...

blink1073 (approved):
Thanks!

blink1073:

blink1073:

jasongrout:
Since we've been getting lots of reports about this recently, can we also backport to 3.0.x? I'm happy to do a 3.0.x release if needed.

blink1073:
@meeseeksdev please backport to 3.0.x

jasongrout:
> Since we've been getting lots of reports about this recently, can we also backport to 3.0.x? I'm happy to do a 3.0.x release if needed.

3.0.18 has now been released with this bugfix.

The-Compiler:
> 2\. This commit does not handle \r line endings, which I think were only default on old mac computers? See https://en.wikipedia.org/wiki/Newline.

FWIW I had a case of a Jupyter file triggering the mentioned bug which had a mixture of `\r\n`, `\n` and only `\r`. This was a formerly correct file edited in VS Code on Windows. Unfortunately I don't have more information, as it happened on the machine of a student of mine...

jasongrout:
Thanks. I have a follow-up patch that I need to submit as a PR that also handles \r, and does a bit more to handle things at the codemirror level.

PR #11149, opened 2021-09-26
Fix the "Edit on GitHub" link

### LINKED ISSUE #9755: Edit on GitHub button does not work in documentation (stable)
When on https://jupyterlab.readthedocs.io/en/stable/# the `Edit on GitHub` button goes to `https://github.com/jupyterlab/jupyterlab/blob/origin/3.x/docs/source/index.rst` which does not exist; the origin part is spurrious, the correct URL would be: `https://github.com/jupyterlab/jupyterlab/blob/3.x/docs/source/index.rst`.

Probably trivial to fix, but opening the issue anyways, i.e. how could we make sure that it works in the future. Would it be worth to set up an automated check for dead links in documentation?

krassowski:
## References

May solve #9755 (but requires testing on production I guess)

## Code changes

- workaround the issue by replacing surplus `origin/`,
- update the template using upstream [breadcrumbs.html](https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/breadcrumbs.html) which now requires to use `page_source_suffix` instead of `suffix` (which is missing `.rst` in new Sphinx versions)

## User-facing changes

None

None

### DIFF
--- a/docs/source/_templates/breadcrumbs.html
+++ b/docs/source/_templates/breadcrumbs.html
@@ -15,24 +15,24 @@
         <!-- User defined GitHub URL -->
         <a href="{{ meta['github_url'] }}" class="fa fa-github"> {{ _('Edit on GitHub') }}</a>
       {% else %}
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

blink1073:

blink1073:

PR #11157, opened 2021-09-27
Update changelog for 3.0.17

blink1073:


jasongrout:
I'm going through the bugfixes seeing if there are any other big ones we can backport to 3.0.x. Did you generate this by hand? If so, I'll add another commit here. If not, is it easy to regenerate?

blink1073:
(which is forked from https://github.com/jupyter-server/jupyter_releaser/blob/master/.github/workflows/generate-changelog.yml).  You can run it on the source repo since it doesn't require auth

Issue #8929, opened 2020-08-28
Fix FIXMEs in testing kernel mocks


jasongrout:
In https://github.com/jupyterlab/jupyterlab/pull/8921/commits/ea33b6d942a005265fefb9d0cfde96236d5efdbc, there were three FIXMEs that were ignoring typing errors. We should fix these properly.

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/11159

PR #11159, opened 2021-09-27
Resolve typing errors in kernel mocks

ph-ph:
## References
Addresses issue https://github.com/jupyterlab/jupyterlab/issues/8929

## Code changes
I'll comment on specific changes in the diff

## User-facing changes
None

None that I'm aware of

### DIFF
--- a/testutils/src/mock.ts
+++ b/testutils/src/mock.ts
@@ -170,6 +170,10 @@ export const KernelMock = jest.fn<
     ...jest.requireActual('@jupyterlab/services'),
     ...options,
     ...model,
+    model,
+    serverSettings: ServerConnection.makeSettings(
+      options.serverSettings as Partial<ServerConnection.ISettings>
+    ),
     status: 'idle',
     spec: Promise.resolve(spec),
     dispose: jest.fn(),
@@ -235,7 +239,7 @@ export const KernelMock = jest.fn<
         KernelMessage.IExecuteReplyMsg
       >;
     })
-  } as any; // FIXME: fix the typing error this any cast is ignoring
+  };
   // Add signals.
   const iopubMessageSignal = new Signal<
     Kernel.IKernelConnection,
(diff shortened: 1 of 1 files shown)


testutils/src/mock.ts
```diff
@@ -170,6 +170,10 @@ export const KernelMock = jest.fn<
     ...jest.requireActual('@jupyterlab/services'),
     ...options,
     ...model,
+    model,
+    serverSettings: ServerConnection.makeSettings(
+      options.serverSettings as Partial<ServerConnection.ISettings>
```
ph-ph, line 175: Treating `RecursivePartial` as `Partial` should be fine, I guess?

testutils/src/mock.ts
```diff
@@ -271,6 +275,7 @@ export const SessionConnectionMock = jest.fn<
   const name = kernel?.name || options.model?.kernel?.name || DEFAULT_NAME;
   kernel = kernel || new KernelMock({ model: { name } });
   const model = {
+    id: UUID.uuid4(),
```
ph-ph, line 278: Moved here from line 282. This way, we guarantee that `model` has `id`, but allow `options.model` to override it.

testutils/src/mock.ts
```diff
@@ -388,10 +395,9 @@ export const SessionContextMock = jest.fn<
     path: session.path,
     type: session.type,
     name: session.name,
-    kernel: session.kernel,
```
ph-ph, line 391: No such property in `ISessionContext`

testutils/src/mock.ts
```diff
@@ -826,7 +832,7 @@ namespace Private {
   export function changeKernel(
     kernel: Kernel.IKernelConnection,
     partialModel: Partial<Kernel.IModel>
-  ): Promise<Kernel.IModel> {
+  ): Promise<Kernel.IKernelConnection> {
```
ph-ph, line 835: All happy paths return one of the elements in `RUNNING_KERNELS`, and that array is of type `IKernelConnection`. Fixing return type resolves several type errors above.

fcollonval: :+1:

testutils/src/mock.ts
fcollonval: :+1:

testutils/src/mock.ts
fcollonval: :+1:

fcollonval (approved):
LGTM thanks @ph-ph 

> Failing test is unrelated

Issue #10561, opened 2021-07-07
Update debugger variable panel instead of closing it.


fcollonval:
### Problem

For now, when an user opens variable panel, it is closed directly if the kernel state changes. It will be more interesting to let it open a refresh it to let the user see the evolution of variable (like it is done for a watch expression).

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/10299

### Proposed Solution

Instead of closing the panel on variable changed event, a new request to update the variable should be sent to the kernel.

https://github.com/jupyterlab/jupyterlab/blob/603966586d9cbd46dc7d38100cc2158605e5f7a3/packages/debugger-extension/src/index.ts#L325

### Additional context

Special care will be needed to handle widgets as they automatically update as the view in the panel will be notified through direct kernel messages.

MMesch:
The triaging agent has identified this issue as potentially resolved but from https://github.com/jupyterlab/jupyterlab/pull/11171 it seems that what is left is fixing this when debugging code _with breakpoints_ as well, correct @fcollonval ? Any other updates on the state of this issue?

krassowski:
Would it be related to https://github.com/jupyterlab/jupyterlab/issues/17916?

Issue #9765, opened 2021-02-09
Cannot "trust" a notebook with blank cells


jasongrout:
## Description

In https://discourse.jupyter.org/t/jupyterlab-not-remembering-trust-for-notebooks/7869, a user told about how they had difficult persisting trust state in a new notebook. I tried trusting a new notebook, and could not get JLab to do so.

## Reproduce

In JupyterLab 3.0.7 in a clean environment (but also in latest JLab 2 or JLab 1, from my experiments):

1. Create a new notebook and evaluate a few cells and have a few blank cells
2. Check the trust status in the status bar. For me, it says for example "Notebook not trusted: 3 cells of 4 cells trusted", where I have 4 cells and one of them is blank.
3. Run the "Trust notebook" command from the command palette, and see that the trust status bar icon still says the same thing.
4. Close and shutdown the notebook, then open it again. None of the cells will be trusted (which IIRC is expected behavior if at least one cell was untrusted when saved)
5. Run all the cells and delete the empty cells. Then the trusted status bar icon shows the notebook is trusted.

## Expected behavior

Running all the cells in a notebook should trust all the cells (and thus trust the notebook as a whole). Probably blank cells (both input and output) should be considered trusted by default.

## Context

- Operating System and version: macOS Catalina
- Browser and version: Firefox 85
- JupyterLab version: 3.0.7

It does not look like there are any relevant errors in the js console or in the server log.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/jupyterlab-not-remembering-trust-for-notebooks/7869/2

echarles:
> In JupyterLab 3.0.7 in a clean environment (but also in latest JLab 2 or JLab 1, from my experiments):
> Create a new notebook and evaluate a few cells and have a few blank cells
> Check the trust status in the status bar. For me, it says for example "Notebook not trusted: 3 cells of 4 cells trusted", where I have 4 cells and one of them is blank.
> Run the "Trust notebook" command from the command palette, and see that the trust status bar icon still says the same thing.
> Close and shutdown the notebook, then open it again. None of the cells will be trusted (which IIRC is expected behavior if at least one cell was untrusted when saved)
> Run all the cells and delete the empty cells. Then the trusted status bar icon shows the notebook is trusted.

Tried this scenario on latest 3.1, and can not reproduce it. The trust is showing fine when closing and reopening.

[...]

PR #11171, opened 2021-09-28
Update variable renderer panels

### LINKED ISSUE #11009: Rendering on "special variables" and "function variables" should not be possible/break the debugger
## Description

The new variable renderer for the debugger should not show when hovering over "special variables" or "function variables".
Clicking on "function variables" leads to a traceback in xeus-kernel, while clicking on "special variables" freezes the GUI

## Reproduce

1. Open JupyterLab 4.0a
2. Activate the debugger
3. Set a breakpoint and run the cell with the breakpoint
4. See that  the button for "rendering" is visible for special categories "special variables" and "function variables" and not only the actual variables

Screenshot from 2021-09-04 12-55-06

## Expected behavior

- do not show the loop icon on hover over "special variables" and "function variables"
- kernels should do some sanity check on the variables to avoid syntax errors

## Context

- JupyterLab version: 4.0.0a7

```
Error processing internal command.
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 419, in evaluate_expression
    compiled = compile(_expression_to_evaluate(expression), '<string>', 'eval')
  File "<string>", line 1
    function variables_repr_data,function variables_repr_metada = (get_ipython().display_formatter.format(function variables))
             ^
SyntaxError: invalid syntax

During handling of the above exception, another exception occurred:

  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/pydevd.py", line 1676, in process_internal_commands
    int_cmd.do_it(self)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 542, in do_it
    self.method(dbg, *self.args, **self.kwargs)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 1331, in internal_set_expression_json
    result = pydevd_vars.evaluate_expression(py_db, frame, exec_code, is_exec=True)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 371, in new_func
    return _run_with_unblock_threads(original_func, py_db, curr_thread, frame, expression, is_exec)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 339, in _run_with_unblock_threads
    return _run_with_interrupt_thread(original_func, py_db, curr_thread, frame, expression, is_exec)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 310, in _run_with_interrupt_thread
    return original_func(py_db, frame, expression, is_exec)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 421, in evaluate_expression
    Exec(_expression_to_evaluate(expression), updated_globals, frame.f_locals)
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<string>", line 1
    function variables_repr_data,function variables_repr_metada = (get_ipython().display_formatter.format(function variables))
             ^
SyntaxError: invalid syntax
```

fcollonval:
## References

Addresses #10561 for variables exploration - does not fix it when debugging code with breakpoints.
Fixes #11009

## Code changes

- Trigger update of the variable renderer when the variables model changes.
- Keep the variable renderer window open but display an info message if the variable is not available in the current context
- Move the variable selection to the variables model instead of accessing it from the view (this is ok as that API has not yet been released except in v4 alphas).
- ~~Don't trust variables if the notebook is not trusted - unsure about this one as the variable needs to be in the kernel memory for it to be displayed.~~ Trust unconditionally as the user must execute some code to get the variable in-memory.

## User-facing changes

The variable renderer panels stay open when running cells or changing the active cell.

The variable renderer widget takes a `dataLoader` function and not a promise to allow data refresh.

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -46,7 +46,6 @@ import {
 import { Session } from '@jupyterlab/services';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITranslator } from '@jupyterlab/translation';
-import { toArray } from '@lumino/algorithm';
 
 /**
  * A plugin that provides visual debugging support for consoles.
@@ -453,23 +452,9 @@ const variables: JupyterFrontEndPlugin<void> = {
 
         const variablesModel = service.model.variables;
 
-        // Trust the console panel by default
-        let isTrusted =
-          handler.activeWidget instanceof ConsolePanel ? true : false;
-        if (
-          handler.activeWidget instanceof NotebookPanel &&
-          handler.activeWidget.model
-        ) {
-          // Trust variable view if the notebook is trusted
-          isTrusted = !toArray(handler.activeWidget.model.cells).some(
-            cell => cell.trusted === false
-          );
-        }
-
         const widget = new Debugger.VariableRenderer({
           dataLoader: () => service.inspectRichVariable(name!, frameId),
           rendermime: activeRendermime,
-          isTrusted,
           translator
         });
         widget.addClass('jp-DebuggerRichVariable');
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -460,6 +460,7 @@ const variables: JupyterFrontEndPlugin<void> = {
           handler.activeWidget instanceof NotebookPanel &&
           handler.activeWidget.model
         ) {
+          // Trust variable view if the notebook is trusted
           isTrusted = !toArray(handler.activeWidget.model.cells).some(
             cell => cell.trusted === false
           );
@@ -474,7 +475,8 @@ const variables: JupyterFrontEndPlugin<void> = {
         widget.addClass('jp-DebuggerRichVariable');
         widget.id = id;
         widget.title.icon = Debugger.Icons.variableIcon;
-        widget.title.label = `${service.session?.connection?.name} - ${name}`;
+        widget.title.label = name;
+        widget.title.caption = `${name} - ${service.session?.connection?.name}`;
         void trackerMime.add(widget);
         const disposeWidget = () => {
           widget.dispose();
@@ -482,21 +484,7 @@ const variables: JupyterFrontEndPlugin<void> = {
           activeWidget?.disposed.disconnect(disposeWidget);
         };
         const refreshWidget = () => {
-          if (
-            name &&
-            variablesModel.scopes
-              .map(s => s.variables.map(v => v.name))
-              .findIndex(variables => variables.includes(name!)) >= 0
-          ) {
-            widget.refresh();
-          } else if (
-            !frameId &&
-            !service.model.callstack.frames.map(f => f.id).includes(frameId!)
-          ) {
-            // Dispose the widget if the frame in which the variable is defined
-            // does not exists any longer.
-            disposeWidget();
-          }
+          widget.refresh();
         };
         widget.disposed.connect(disposeWidget);
         variablesModel.changed.connect(refreshWidget);
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -325,14 +325,13 @@ const sources: JupyterFrontEndPlugin<IDebugger.ISources> = {
 const variables: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlab/debugger-extension:variables',
   autoStart: true,
-  requires: [IDebugger, IDebuggerHandler, ITranslator, IDebuggerSidebar],
+  requires: [IDebugger, IDebuggerHandler, ITranslator],
   optional: [IThemeManager, IRenderMimeRegistry],
   activate: (
     app: JupyterFrontEnd,
     service: IDebugger,
     handler: Debugger.Handler,
     translator: ITranslator,
-    sidebar: IDebugger.ISidebar,
     themeManager: IThemeManager | null,
     rendermime: IRenderMimeRegistry | null
   ) => {
@@ -441,7 +440,10 @@ const variables: JupyterFrontEndPlugin<void> = {
           return;
         }
 
-        const id = `jp-debugger-variable-mime-${name}`;
+        const id = `jp-debugger-variable-mime-${name}-${service.session?.connection?.path.replace(
+          '/',
+          '-'
+        )}`;
         if (
           !name || // Name is mandatory
           trackerMime.find(widget => widget.id === id) || // Widget already exists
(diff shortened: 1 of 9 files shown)


packages/debugger-extension/src/index.ts
```diff
@@ -459,9 +461,24 @@ const variables: JupyterFrontEndPlugin<void> = {
...
+          variablesModel.changed.disconnect(refreshWidget);
+          activeWidget?.disposed.disconnect(disposeWidget);
         };
-        service.model.variables.changed.connect(disposeWidget);
+        const refreshWidget = () => {
+          if (
+            name &&
+            variablesModel.scopes
+              .map(s => s.variables.map(v => v.name))
+              .findIndex(variables => variables.includes(name!)) >= 0
```
trungleduc, line 472: might it be useful to put this logic into a `VariablesModel` method?

packages/debugger/src/panels/variables/mimerenderer.ts
```diff
@@ -19,30 +20,55 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
...
-          if (mimeType) {
-            const widget = rendermime.createRenderer(mimeType);
-            const model = new MimeModel(data);
-            await widget.renderModel(model);
+    if (data.data) {
+      const hash = murmur2(JSON.stringify(data), 17);
+      if (this._dataHash !== hash) {
+        if (this.content.layout) {
+          this.content.widgets.forEach(w => {
+            this.content.layout?.removeWidget(w);
```
trungleduc, line 45: `this.content.layout` nullish is already checked.

jtpio: This might indeed be improved in TS 4.4 (but may need to assign to a tmp variable to do the check): https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#cfa-aliased-conditions

blink1073:
Can you please run `jlpm bump:js:major debugger debugger-extension` to update the versions?

meeseeksmachine:

https://discourse.jupyter.org/t/how-to-use-debugger-in-jupyterlab/11275/14

fcollonval:
@krassowski I'm back to finish this one. I updated the PR quite substantially to handle properly the trust of a variable (see demo) with the ability to trust single variable even if the notebook is not trusted.
The panel is now long living (it will be disposed only if the user wants it or if the associated panel is closed). If the variable, does not make sense in the current context, the variable renderer display a message. This is coherent with the behavior of debugger in other IDEs (like the browsers, VS Code,...).

Would you mind reviewing this PR (... again)?

demo_render_mime_variable

> Note: it confirms that the trust feature of a notebook is quite broken in JupyterLab.... as discussed https://github.com/jupyterlab/jupyterlab/issues/9765.

fcollonval:
Finally I removed the trust button in the viewer as the user needs to make explicitly an action for the variable to be available in-memory and then displayed.

krassowski:
UX feedback:

1. The variable seems to be identified by its name. Should it be identified by its name **and** source notebook? On the GIF below you can see that when switching between two notebooks with different `x` variables the value updates instead of staying constant. It also precludes comparing two identically named variables from different notebooks side-by-side because no new view for identically named variable can be created.
2. Can we hide the loop icon on "special variables" and "function variables" or is it impossible to reliably distinguish those from others? Is it worth hard-coding for now?

render-variables

fcollonval:
@krassowski thanks for the review. I updated the PR so that the variables are linked to their source widget (and update only if it is the active widget) and I hard coded skipping the groups; this is probably not really satisfactory as they are specific to [python](https://github.com/microsoft/debugpy/blob/cf0d684566edc339545b161da7c3dfc48af7c7d5/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py#L359) - but as this is the primary target of notebook I assume it is ok.

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -257,35 +250,43 @@ const VariableComponent = (props: IVariableComponentProps): JSX.Element => {
...
-              .execute(Debugger.CommandIDs.renderMimeVariable, {
+      {service.model.hasRichVariableRendering &&
+        // Don't add rich display for special entries
+        // debugpy: https://github.com/microsoft/debugpy/blob/cf0d684566edc339545b161da7c3dfc48af7c7d5/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py#L359
+        ![
+          'special variables',
+          'protected variables',
+          'function variables',
+          'class variables'
+        ].includes(variable.name) && (
```
krassowski, line 261: Ultimately, would it make sense to move it to settings so that users can customise it, or should we create a new field to the `inspectVariables` response and move the responsibility of knowing which variables are scopes without mime rendering support? I am happy to open a new issue for that.

packages/debugger/src/tokens.ts
```diff
@@ -616,36 +616,22 @@ export namespace IDebugger {
...
   /**
    * Debugger sidebar interface.
    */
-  export interface ISidebar extends Panel {
-    /**
-     * Debugger variables explorer.
-     */
-    variables: IVariablesPanel;
-  }
+  export interface ISidebar extends Panel {}
```
krassowski, line 634: Should `ISidebar` be removed?

```suggestion
```

packages/debugger/src/panels/variables/mimerenderer.ts
```diff
@@ -12,37 +18,77 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
...
+            this.content.layout!.removeWidget(w);
+          });
+        }
 
-            content.addWidget(widget);
-            loaded.resolve();
-          } else {
-            loaded.reject('Unable to determine the preferred mime type.');
-          }
+        // We trust unconditionally the data has the user is required to
```
krassowski, line 66:
```suggestion
        // We trust unconditionally the data as the user is required to
```

packages/debugger/src/panels/variables/tree.tsx
fcollonval: From the user point of view, it would be better if it does not need to know. So I'm :+1: to get them from the DAP connector.

packages/debugger/src/tokens.ts
fcollonval: In 3.x, `ISidebar` is providing a public API. This is the reason I kept the `Panel` API. I assume the idea is for third-party extension to be able to add widgets in that panel.

https://github.com/jupyterlab/jupyterlab/blob/f4ef57707b65121888c3c87482eb3f3f9ab63425/packages/debugger/src/tokens.ts#L550-L565

fcollonval:
Thanks @krassowski for the review

krassowski (approved):
LGTM, thank you!

fcollonval:

Issue #6890, opened 2019-07-25
Remove dependency on third-party library for `ui-components`


fcollonval:
Based on work that we did for JupyterLab 1.0 and for accessibility, my
sense right now is that using an existing, general purpose UI toolkit
won't solve our problems. Here is what we are seeing:

* Accessibility isn't a fundamental property of a UI toolkit, but a
relationship between the UI components and the overall application.
What I mean by this is that simple adopting Material UI or Blueprint
doesn't make JupyterLab accessible - even if those toolkits are
accessible as possible. Example: Blueprint has well designed
accessible buttons, but they use an exterior offset border to indicate
focus that doesn't work in many contexts where JupyterLab buttons are
in tight layouts and those outlines are cut off by overflow: hidden.
* Complexity. It is orders of magnitude more complex to build/use a
general purpose UI component versus building a UI component that does
just what we need it to. The general purpose nature of these UI
toolkits means that each object has dozens of props, children, styling
options. We don't need to be able to render an exponential number of
buttons - a very small, with a small number of props will do.
* Bugs. The complexity leads to bugs and accessibility problems. With
highly flexible and general components, it becomes easy to use them in
a way that disrupts accessibility or creates bugs. Example: right now,
we are (mis)-using Blueprint buttons by shoving the Icon into the
wrong place, leading to an awkward DOM hierarchy that will likely
cause accessibility problems. This happened even though the person
using Blueprint is one of the most experienced react.js devs in the
project. It is caused by the APIs being overly flexible and complex.
* Difficulty of theming. The reliance on statically compiled themes is
a huge problem for us, as we need to re-theme dynamically (we use
custom CSS properties, but that is an implementation detail).
* Decoupling: the reliance on traditional CSS is killing us, as it is
creating hidden dependencies between packages that make it easy for us
to break one thing while trying to fix another in a separate package.
We haven't decided on which CSS in JS framework we will use, but I
don't see any way we can not have this. This is maybe a softer
requirement in that Material UI and Blueprint may be compatible with
this.

_Originally posted by @ellisonbg in https://github.com/jupyterlab/jupyterlab/pull/6828#issuecomment-513887853_

So from there, the direction for the core packages tends towards building the minimal set of UI components needed. And therefore extension authors should be aware that blueprintjs shipped with the core packages will likely be removed in the future. And so if they need an UI toolkit, they should not choose blueprintjs just because its currently included.

[...]

Issue #6590, opened 2019-06-16
Styling shipped third-party React UI toolkit (i.e. Blueprint for now)


fcollonval:
This issue follows some discussion in PR https://github.com/jupyterlab/jupyterlab/pull/6389
@ellisonbg  @echarles 

**Describe the bug**

Currently JupyterLab ships with Blueprint JS as React-based UI toolkit. But the theme is not overwritten to fit JupyterLab theme. Starting from the hypothesis that introducing a third-party library as base for React UI component is a positive thing, I think it will harm JupyterLab project if that third-party toolkit is not modified to match JupyterLab style by default.

The current status is:
* when we use Blueprint component, styling them to fit JupyterLab style request to use explicitly Blueprint class names; for e.g. in the [PR](https://github.com/jupyterlab/jupyterlab/pull/6389/files#diff-edfc32797ef53c10b2d490a0dd9eccc5) or in the [`ui-components` package](https://github.com/jupyterlab/jupyterlab/blob/79e03992b2964063a917184072be3c8819cefa19/packages/ui-components/style/base.css#L59). Updating Blueprint may break it as such usage is not [advised](https://blueprintjs.com/docs/#core/classes) by Blueprint authors.
* author extensions will need to define style for standard blueprint components to be correctly integrated in JupyterLab. This could results in duplicating style or defining incoherent one.

This highlights the need to style the React UI toolkit within JupyterLab.

Looking at some of the [toolkits available](https://hackernoon.com/23-best-react-ui-component-libraries-and-frameworks-250a81b2ac42) out there, they are two strategies regarding styling: 

* using a theme component around the full react app or locally to overwritten the default theme (like [Material UI](https://material-ui.com/customization/themes/) or [Grommet](https://v2.grommet.io/))
* using SaSS or Less to define the toolkit style; like Blueprint, Ant or [Material Design for Web](https://material.io/develop/web/)

As JupyterLab is not a React application, only the second category seems the way to go.

In the aforementioned PR #6389, I propose to regenerate the Blueprint CSS after modifying the SaSS variables to match more closely the JupyterLab style. But it is not included in the final state as this needs discussion in this issue ;)

echarles:
Thx @fcollonval for starting this discussion. Before going into the potential solutions, a description of the existing situation may be interesting (correct the following, I am sure it is approximative and contains errors and sorry if this bring more questions/confusion).

[...]

Issue #7574, opened 2019-12-02
Future of ui-components package: drop Blueprint; should we use Material-UI?


telamonian:
Refs:
#6828
#6890

There's currently an ongoing discussion about the future of the `ui-components` package in Jupyterlab core:

- There's no particular reason to keep the Blueprint dependency around. We should refactor out the sparse use of it in `ui-components` and drop the dep
    - This should be easy, and it will be an API breaking change, so it should maybe go in for 2.0

- Should we replace Blueprint with Material-UI?
    - Pro:
        - Does a lot of nice stuff that would be too much work to do on our own
        - Really nice dialog boxes
    - Con:
        - Does a lot of complex stuff that we don't need
        - May be counterproductive for accessibility stuff

vidartf:
Duplicate of https://github.com/jupyterlab/jupyterlab/issues/6890 ?

telamonian:
We're discussing issue triage for the 2.0 RC. This is meant to supercede #6890. Basically, the first part of #6890 (removing Blueprint) is easy, and should prob go in 2.0, whereas the second part (should we use Material-UI/something else/nothing?) requires more discussion (and also @ellisonbg).

saulshanabrook:
Hey Max. Did you look into doing a custom theme for material UI? https://material-ui.com/customization/theming/

That way we could wrap our top level react render in the react widget with this context... And then users can just use native material UI components and they will be themed.

It would be nice to not *have* to wrap all their components... Instead just maybe export our own if we really need a new one, but otherwise just relying on theming theirs.

cc @echarles who is also exploring this

saulshanabrook:
One thing to be aware of is MUI's internal plans on what library they are using, There is a long running issue to switch to styled components from their own internal system (https://github.com/mui-org/material-ui/issues/6115). It seems like in their next version, they will switch to styled components and maybe also let this configurable, so that you could use it with multiple CSS in JS solutions (https://github.com/mui-org/material-ui/issues/20012).

echarles:
Just did a quick scan and not sure of the impacts. But whatever internal solution and compatibility option MUI  will provide, I would expect they will keep a theming layer like https://material-ui.com/customization/theming/ (maybe with breaking changes)?

saulshanabrook:
@echarles Yes that's true! It only matters if we care about which one it uses internally, which I am not sure if we.

saulshanabrook:
[...]

PR #11173, opened 2021-09-28
Remove Blueprint

fcollonval:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/6890
Closes https://github.com/jupyterlab/jupyterlab/issues/6590
Quid of https://github.com/jupyterlab/jupyterlab/issues/7574

## Code changes

- Drop Blueprint JS
- Add some React components to replace those from Blueprint:
  - `Button`
  - `Collapse`
  - `InputGroup`
- Remove unused components: `Select` and `Checkbox`
- Remove unused exported `Intent`
- [x] Bump major version of `@jupyterlab/ui-components`

Additional changes:
- Example tests are now using Playwright instead of puppeteer
- ServerConnection has been modified to tolerant undefined `process.env` (following removing a webpack patch for Blueprint in the examples)
- Explicitly resolve react to v17 because storybook is requiring v16. This results in two versions of react in `node_modules` that generates error on unit tests rendering a react LabIcon.
- [x] Start 4.0 migration guide

## User-facing changes

N/A (Hopefully - although the components are simpler and may lack some advanced transition or shadow effects).

Downstream extensions using the proxied Blueprint component will need to explicitly use the Blueprint component as only the part of the API used in JupyterLab core was added to the remaining components.

Style changes:
- Shift of 1px of icon in button due to the removal of the blueprint icon wrapper (did not find where that additional pixel is coming). But the size of the button itself is not changed.
- BlueprintJS was defining a default font size of 14px on `body`. This has been replace by a font size of `var(--jp-ui-font-size1)` that equals 13px. This impacts the text in the inspector panel when the current widget is not a notebook for example. But at least all texts are themed. So I considered that regression as acceptable.


krassowski:
From screenshots it seems that there are significant changes in top/bottom margin/padding of buttons and other elements (like TOC, where I think we should keep more compact layout). Are you planning to address those in this PR?

fcollonval:
> From screenshots it seems that there are significant changes in top/bottom margin/padding of buttons and other elements (like TOC, where I think we should keep more compact layout). Are you planning to address those in this PR?

Yes I think it makes more sense to have almost no visual regression on this one. I need to look a bit deeper in the change you are mentioning as they don't come from the explicitly used components but from general style rules leaking outside Blueprint component that are no more imported. Hopefully they are not too many of them...

fcollonval:
There is a one-pixel change on toolbar button icon:

code-cell-diff

In Blueprint js there is an additional `span` element wrapping the icon that get a height of _icon height + 1px_. But I cannot figure out where this additional pixel is coming from. As this does not change the height of the button itself, I updated the screenshot assuming this is an acceptable visual regression.

fcollonval:
> FYI Impact on the bundle size using the info from the check release step

master 4c8bcf7

assets by info 7.62 MiB [immutable]
  assets by chunk 4.75 MiB (id hint: vendors)
    assets by status 3.32 MiB [big] 7 assets
assets by chunk 2.79 MiB (auxiliary id hint: vendors)
  assets by path *.svg 1.73 MiB 3 assets
  assets by path *.eot 363 KiB 3 assets
  assets by path *.ttf 362 KiB 3 assets
  assets by path *.woff 203 KiB 3 assets
  assets by path *.woff2 164 KiB
    asset 1551f4f60c37af51121f106501f69b80.woff2 76.4 KiB [emitted] (auxiliary id hint: vendors)
    asset d878b0a6a1144760244ff0665888404c.woff2 74.9 KiB [emitted] (auxiliary id hint: vendors)
    asset 7a3337626410ca2f40718481c755640f.woff2 12.9 KiB [emitted] (auxiliary id hint: vendors)
asset third-party-licenses.json 275 KiB [emitted] [big]
asset index.html 1.56 KiB [emitted]
orphan modules 5.08 MiB [orphan] 1288 modules
runtime modules 111 KiB 14 modules
built modules 14.2 MiB (javascript) 4.59 KiB (share-init) 4.59 KiB (consume-shared) [built]
  modules by path ./node_modules/ 14.2 MiB
    javascript modules 14.2 MiB 1047 modules
    json modules 33.2 KiB 6 modules
  provide-module modules 4.59 KiB 112 modules
  consume-shared-module modules 4.59 KiB
    modules by path consume shared module (default) @jupyterlab/ 3.65 KiB 41 modules
    modules by path consume shared module (default) @lumino/ 588 bytes 14 modules
  modules by path ./build/ 23.5 KiB
    ./build/bootstrap.js 3.86 KiB [built] [code generated]
    ./build/index.out.js + 1 modules 19.7 KiB [built] [code generated]


---

This PR c7be037

assets by info 6.41 MiB [immutable]
  assets by chunk 3.54 MiB (id hint: vendors)
    assets by status 2.11 MiB [big] 6 assets
assets by chunk 2.79 MiB (auxiliary id hint: vendors)
  assets by path *.svg 1.73 MiB 3 assets
  assets by path *.eot 363 KiB 3 assets
  assets by path *.ttf 362 KiB 3 assets
  assets by path *.woff 203 KiB 3 assets
  assets by path *.woff2 164 KiB 3 assets
asset third-party-licenses.json 205 KiB [emitted]
asset index.html 1.56 KiB [emitted]
orphan modules 4.47 MiB [orphan] 1156 modules
runtime modules 110 KiB 14 modules
built modules 12.4 MiB (javascript) 4.59 KiB (share-init) 4.59 KiB (consume-shared) [built]
  modules by path ./node_modules/ 12.4 MiB
    javascript modules 12.4 MiB 963 modules
    json modules 33.2 KiB 6 modules
  provide-module modules 4.59 KiB 112 modules
  consume-shared-module modules 4.59 KiB
    modules by path consume shared module (default) @jupyterlab/ 3.65 KiB 41 modules
    modules by path consume shared module (default) @lumino/ 588 bytes 14 modules
  modules by path ./build/ 23.5 KiB
    ./build/bootstrap.js 3.86 KiB [built] [code generated]
    ./build/index.out.js + 1 modules 19.7 KiB [built] [code generated]

krassowski:
> In Blueprint js there is an additional span element wrapping the icon that get a height of icon height + 1px. But I cannot figure out where this additional pixel is coming from. As this does not change the height of the button itself, I updated the screenshot assuming this is an acceptable visual regression.

It looks better to me after, so I would call it an improvement ;)

krassowski:
Would it be a good idea to start writing extension migration guide for 4.0 in this PR to document this change?

jasongrout:
> Explicitly resolve react to v17 because storybook is requiring v16. This results in two versions of react in `node_modules` that generates error on unit tests rendering a react LabIcon.

What advantage do we have from storybook? I've had problems with storybook in the past when trying to upgrade packages as well. Is anyone using storybook in JupyterLab?

fcollonval:
Personally I never used storybook in JupyterLab. So I'm fine to drop it (although I would suggest doing it in a separate PR).

krassowski:
The new implementation of `InputGroup` does not implement style for `disabled` attribute; on 3.x:

Screenshot from 2021-10-29 19-18-46

On this PR:

Screenshot from 2021-10-29 19-18-53

The blueprint styles for the disabled state for input group:

```css
    background: rgba(206, 217, 224, 0.5);
    -webkit-box-shadow: none;
    box-shadow: none;
    color: rgba(92, 112, 128, 0.6);
    cursor: not-allowed;
    resize: none;
```

Let's add at least cursor and background (does not have to be exactly the same)

krassowski (commented):
Looks great, I would suggest two changes before merging:
- adding the missed style for `disabled` state of `InputGroup`
- reviewing suggestions on the documentation grammar and the missing 3.0 → 3.1 heading

The visual tests show three changes:
- The toolbar icons moved one pixel up, which looks better than before (visually more centered) so I would like to keep this change as mentioned earlier
- There is also some (sub-?)pixel change in top/bottom margins of outputs after images, visible on `galata/test/galata/notebook.spec.ts-snapshots/example-run-galata-linux.png` and `galata/test/jupyterlab/notebook-run.test.ts-snapshots/notebook-panel-0-jupyterlab-linux.png` but I could not figure it out when comparing the elements with inspector. I believe that it is acceptable, just documenting the change here.
- Very minor and non-blocking, the font size in inspector got reduced (`galata/test/jupyterlab/sidebars.test.ts-snapshots/opened-sidebar-jp-property-inspector-jupyterlab-linux.png`) by 1px; it is not critical to change, I think that  larger font is better here because the centered text may be missed on larger monitors.

fcollonval:
Thanks for reviewing @krassowski 

>     * Very minor and non-blocking, the font size in inspector got reduced (`galata/test/jupyterlab/sidebars.test.ts-snapshots/opened-sidebar-jp-property-inspector-jupyterlab-linux.png`) by 1px; it is not critical to change, I think that  larger font is better here because the centered text may be missed on larger monitors.

BlueprintJS was defining a default font size of 14px on body. This has been replaced by a font size of var(--jp-ui-font-size1) that equals 13px. This impacts the text in the inspector panel when the current widget is not a notebook for example. But at least all texts are themed. So I considered that regression as acceptable.

krassowski (approved):
LGTM, thank you!

jtpio:
>  FYI Impact on the bundle size using the info from the check release step

This is also a nice side-effect!

PR #11175, opened 2021-09-28
Fix kernelspec logo handling

jtpio:
## References

Fix the handling of kernelspec logos to allow for arbitrary URLs.

This came up in https://github.com/jupyterlite/jupyterlite/pull/352 where we use the federated extension mechanism to load federated kernels in the browser. The kernels bring their own resources such as the logos, which are then served under a different URL that does not necessarily include `kernelspecs` in it (or could be absolute URLs to external sites like CDNs).

## Code changes

Looks like the handling of the base url is already done in `jupyter_server`:

https://github.com/jupyter-server/jupyter_server/blob/ac93dd8f60de54db99167016f0396959826b8ed9/jupyter_server/services/kernelspecs/handlers.py#L33

## User-facing changes

None

Since this logic already lives in Jupyter Server, this should be backwards compatible and can be backported to 3.1.x and 3.2.x.

### DIFF
--- a/packages/console-extension/package.json
+++ b/packages/console-extension/package.json
@@ -41,7 +41,6 @@
     "@jupyterlab/apputils": "^3.3.0-alpha.11",
     "@jupyterlab/codeeditor": "^3.3.0-alpha.11",
     "@jupyterlab/console": "^3.3.0-alpha.11",
-    "@jupyterlab/coreutils": "^5.3.0-alpha.11",
     "@jupyterlab/filebrowser": "^3.3.0-alpha.11",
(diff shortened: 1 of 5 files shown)


jtpio:
> Looks like the handling of the base url is already done in `jupyter_server`:

This seems to indeed be the case, for example the `baseUrl` is correctly added as a prefix when testing on Binder:

image

fcollonval (approved):
LGTM thanks @jtpio

blink1073:

jtpio:
Just opened https://github.com/jupyterlab/jupyterlab/pull/11183 to do the manual backport to `3.2.x`

PR #11180, opened 2021-09-29
Add new package "user" to represent a connected users

hbcarlos:
As described in #10965, JupyterLab doesn't have any representation of a user. With the incorporation of real-time collaboration, we introduced a multiuser feature that needs to identify connected users somehow and show them as collaborators.

## References
#10965

## Code changes
This PR adds two new packages, "@jupyterlab/user", and "@jupyterlab/user-extension". These packages introduce three new plugins:

* ICurrentUser that provides a representation of the connected user, the idea is that some extensions will require this plugin to modify the information (an extension for Hub to fetch the user info from the backend), and other extensions will require the user just to read the information (docprovider to use this identity in RTC, or and extension to add comments in a notebook)
* IUserMenu that adds a new menu to the JupyterLab toolbar with the current user's name allows anonymous users to rename themself and provides a lumino menu where extensions can add different buttons to actions related to the user (Login with GitHub, Logout, settings, ...).
* IUserPanel (Maybe needs to be renamed) is a side panel to add info about the current user and the shared document (who is collaborating, what permissions they have, etc...)

## Demo

https://user-images.githubusercontent.com/26092748/135240635-70a89ee3-f299-4b3d-88e6-99b113d38e37.mp4


blink1073:
Thanks @hbcarlos!  I marked this as 4.0, but we can talk about whether this fits into the 3.2 timeline at today's meeting

hbcarlos:
Hey Steven, thanks!
Yes, this goes to 4.0 milestone. I will bring it to the meeting today to get some feedback!

isabela-pf:
Hooray! This is great, and demo video was so helpful! Thanks for making this so easy to provide feedback on, @hbcarlos. 

I know this PR is in draft state still, so I won't give too much feedback, but I'm also excited to see someone working on polishing the user experience around collaboration. Almost everything you showed looked like what I'd expect, so I only have two questions.

- Is it intentional that default/unnamed anonymous user does not appear in the list of collaborators? 
- Does the rename/user's self-assigned name carry over between sessions, across notebooks, or any other interactions? Or is it something the user resets every time they open JupyterLab with collaboration enabled. (I'd guess there's some technical limitations dictating the experience here). 

I'm really looking forward to seeing how this work progresses! I think this work is critical to a responsible collaboration experience.

hbcarlos:
Hi @isabela-pf, thanks for looking at this PR!

> Is it intentional that default/unnamed anonymous user does not appear in the list of collaborators?

The list of collaborators is the people working in the same document. The list changes when moving between documents showing the people that are collaborating in the focused notebook. Should we change the title to something like "Notebook's collaborators"?

> Does the rename/user's self-assigned name carry over between sessions, across notebooks, or any other interactions? Or is it something the user resets every time they open JupyterLab with collaboration enabled. (I'd guess there's some technical limitations dictating the experience here.)

Yes, the name of the user is stored using `StateDB`. If there is no name saved when starting JupyterLab, we pick one anonymous name from the list and save it.

At the moment, I'm working on the user menu and creating a settings page to change the name and the color. I'm making a new page and not using the JupyterLab settings because the idea is to allow the user to update the name only when it is anonymous. Otherwise, the identity comes from the backend, and I don't think the user should change the name. I'm also creating a visual color picker using the HTML tag `input.`

The user menu has only one entry (settings). The idea is to allow other extensions to add their login and logout buttons because every deployment uses different authentication mechanisms.

A small demo of the user menu and settings page (feedback is wellcome):

https://user-images.githubusercontent.com/26092748/136713996-6494cc2a-5cdf-49da-b5c7-9df3ea6019c8.mp4

hbcarlos:
I am closing this PR to split the different features into multiple PR. This way would be easier to disscuse and review.
Also, at this point I changed so many times the code that's easier to open a new PR.

I will continue in #11443 plus other PRs for the list of collaborators and the rename of the anonymous users.

PR #11011, opened 2021-09-04
Restore Copy shareable link use of shareUrl

fcollonval:
## References

Fix #10369

## Code changes

Add new option on `PageConfig.getUrl` to use `shareUrl` instead of `baseUrl`.
Change default mode behavior to ensure it target `multi-document` on default workspace
Create a _shareable_ link with the current mode (don't force `single-document`).

## User-facing changes

Shareable link open in the same mode than the active mode
Shareable link use `shareUrl` if it is defined

## Backwards-incompatible changes

N/A

PR #11188, opened 2021-09-30
Backport PR #11011 on branch 3.2.x (Restore Copy shareable link use of shareUrl)

fcollonval:


fcollonval:
Ok I figure out why the internationalization is not working. I miss the point that the comment with the string code positions are part of the string used to generate the hash like:

```
#: packages/filebrowser-extension/src/index.ts:978
#: packages/mainmenu-extension/src/index.ts:526
msgid "Shut Down Kernel"
msgstr ""
```

I'll push a PR to keep only the `msgid` and `msgctxt`.

blink1073:
> Ok I figure out why the internationalization is not working

Great!

blink1073 (approved):
Thanks!

PR #11190, opened 2021-09-30
Use only context and id to check i18n

### LINKED ISSUE #11189: Internationalization check is raising false positive
I miss the point that the comment with the string code positions are part of the string used to generate the hash like:

```
#: packages/filebrowser-extension/src/index.ts:978
#: packages/mainmenu-extension/src/index.ts:526
msgid "Shut Down Kernel"
msgstr ""
```

I'll push a PR to keep only the `msgid` and `msgctxt`.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab/issues/11188#issuecomment-931307654_

fcollonval:
## References

Fixes #11189

## Code changes

Use only context and id to check i18n
Sort the entries

## User-facing changes

N/A

N/A

### DIFF
--- a/scripts/i18n_check.py
+++ b/scripts/i18n_check.py
@@ -20,8 +20,10 @@ if __name__ == "__main__":
         )
 
     hash = sha256()
-    for entry in pot:
-        hash.update(str(entry).encode("utf-8"))
+    # Use only the context and the id as the position may changed without impact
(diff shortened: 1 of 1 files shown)


blink1073:

blink1073:

PR #11196, opened 2021-10-01
Clean up Frontend Typings

blink1073:
## References
N/A.  Additions needed to help when writings tests for extensions.

## Code changes
- Allow alternate `ServiceManager` implementations in `JupyterFrontEnd`.  Specifically this allows the use of a mock manager.
- Make the `JupyterLab.IOptions` more permissive: do not require a shell when options are given and allow a shell that meets the `ILabShell` interface
- Change the `language` used in the mock `irkernel` to get some diversity in the kernel data for testing purposes

## User-facing changes
None

Consumers relying on the concrete `ServiceManager` have to be updated.


jtpio:
Nice, thanks!

Pushing a bit on this and putting some thoughts about it here, should `ServiceManager.IManager` allow for optional services? For example:

```typescript
readonly sessions?: Session.IManager;
readonly contents?: Contents.IManager;
readonly terminals?: Terminal.IManager;
// ...
```

In alternate lab-based applications, some of these services are not needed (for example `terminals` in Voila). Having them optional would indicate they might or might not be available. Plugins would then have to check whether the functionality is available or not before using it.

However this would constitute quite of a major change. That would mean updating quite a lot of code in core lab, but also in third-party extensions.

This question is somehow related to initiatives like [jupyverse](https://github.com/jupyter-server/jupyverse), where the backend is a set of plugins where some functionality might be disabled.

---

For now the simpler might be to say the `ServiceManager.IManager` always provides these services, even if they don't do anything. Alternate lab-based applications can provide mocks as mentioned in the top comment of this PR. For example for the `Terminal.IManager` service it would mean implementing no-op operations for most of the interface:

https://github.com/jupyterlab/jupyterlab/blob/36aa21f1037da0a5df7323e3b1d64fa029cc7630/packages/services/src/terminal/terminal.ts#L105-L196

blink1073:
> Looks like this would require a manual major bump of the `@jupyterlab/application` package?

Yes, which would update all of our extensions by major a major version as well (and anything else that depends on it).

> Were you planning to add more changes to this PR? 

I can make those changes on this PR and then bump those versions.  I'll wait and bring this up at next week's meeting for consensus before making any more changes.

blink1073:
I updated the API of `ServiceManager` to accept and provide interfaces for all the sub-managers and added no-op implementations of the kernel, session, and terminal managers.  I'll follow up with tests for the new managers and the major version bumps.

blink1073:
Code changes are done, I'll come back and sort out the rebase.

blink1073:
I'm too burned out to finish this PR.

jtpio:
No problem, thanks a lot @blink1073 for all the work :heart: 

I opened https://github.com/jupyterlab/jupyterlab/pull/11537 to continue there.

Issue #11116, opened 2021-09-22
renaming and a notebook (a.ipynb -> b.ipynb) and creating a new one (a.ipynb) lets edits appear in both 


pfarndt:
## Description

Edits in one notebook show up in another as well.

## Reproduce

1. Create a new (empty) subfolder and change into it
2. Create a new notebook - it will be named "Untitled.ipynb"
3. Create some cells and save it - keep it open - rename it in the file browser to e.g. "t1.ipynb"
4. Create yet another new (empty) notebook - it will be again named "Untitled.ipynb".
5. Create some cell in it - these cells will now also show up in "t1.ipynb", i.e. in two notebooks at the same time.

This is for sure not expected and I lost some large chunks of work stumbling into this issue. This stops as soon as I rename "Untitled.ipynb" to something else.

- Operating System and version: Linux
- Browser and version: Safari (from remote) 
- JupyterLab version: 3.1.12

pfarndt:
Some more factss:
- this bug is persistent in JupyterLab 3.1.14. 
- it also happens if the notebook is mv'ed on the command line.
- it happens also for other filenames - not just 'Untitled.ipynb'
- it happens also if the kernel of the nb is shut down

I find this bug very disturbing, because it messes with notebooks which I do not edit at the moment but had the same name earlier. Lowers my confidence enormously.

jasongrout:
I could not reproduce this just launching JupyterLab with `jupyter lab`. However, I was able to reproduce all sorts of problems around renaming files when launching with the experimental `jupyter lab --collaborative` flag. Please note that the collaborative features are still experimental, and not recommended for production. (Side note: I notice that the `--collaborative` command line option does not note that it is experimental, but probably should).

@dmonad - Can you take a look at this? Here is a screencast of some of the issues happening for me. If I had to guess, it seems that a rename in the file browser is not propagating to open files to change the collaboration models for those open notebook widgets.

![renamingproblems](https://user-images.githubusercontent.com/192614/135658647-8b5a8b10-c3b3-4e61-8233-84949c0bce85.gif)

jasongrout:
@pfarndt - can you confirm my findings above, i.e., that you see the problems only with the `--collaborative` flag, but running without the `--collaborative` flag is fine?

krassowski:
There was an issue in similar vein earlier on master https://github.com/jupyterlab/jupyterlab/issues/10436 before the rename dialog was postponed. It might have a common cause.

[...]

PR #11128, opened 2021-09-23
Run comparative benchmark

fcollonval:
## References

Follow-up of the discussion at the JuypterLab dev meeting https://github.com/jupyterlab/team-compass/issues/128#issuecomment-920282463

## Code changes

Add interval change computation from `jupyterlab/benchmarks` repository
Homogeneous interface for the report and the graph creation functions in `BenchmarkReporter`.

Change the logic for benchmark test:
- Compute the _reference_ and the _challenger_ in the same job to be hardware independent
- [ ] Set job to run only on approved PR review
- Set job to run every sunday to evaluate changes introduced by the last week commits

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

PR #11197, opened 2021-10-02
Fix renaming issue in collaborative mode

dmonad:
## References

Fixes the issue described in #11116. I briefly described the problem in the mentioned ticket. The y-websocket provider distributes updates through two channels: via the websocket connection to the jupyter server and via broadcastchannel - a technology for cross-window communication. The problem was that we didn't properly unsubscribe from messages of the broadcastchannel when a file is renamed. Hence we still received messages from the "Untitled" document.

## Code changes

Properly unsubscribe and resubscribe to messages from the new document-path.

## User-facing changes

None.

None.

### DIFF
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -135,10 +135,6 @@ export class WebSocketProviderWithLocks
   setPath(newPath: string): void {
     if (newPath !== this._path) {
       this._path = newPath;
-      // The next time the provider connects, we should connect through a different server url
-      this.bcChannel =
-        this._serverUrl + '/' + this._contentType + ':' + this._path;
-      this.url = this.bcChannel;
       const encoder = encoding.createEncoder();
       encoding.write(encoder, 123);
       // writing a utf8 string to the encoder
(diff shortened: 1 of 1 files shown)


jasongrout:
I haven't looked deeply into this code recently, but it is interesting that we are escaping the path in this function, but then using the path directly in constructing this.url. Do we not need to worry about escaping the path, for example if the path has spaces, etc?

dmonad:
I might misunderstand, but we are not escaping the path. We construct a unique URL for our document in the form `wss://{server}/{"file" | "notebook"}:{path}`. 

We probably don't have to worry about spaces because the `path` variable already contains an escaped path. It is then automatically url-encoded.

blink1073:
I'm afk, but the binder link is in collaborative mode so it should be possible to check a path with spaces using the original repro instructions

SylvainCorlay:
Thanks for fixing it!

SylvainCorlay:
🎉

jasongrout:
This seems to work great when I test it, and my worries about escaping paths seem to be unfounded - the url is url-encoded appropriately when actually used, like Kevin says.

Thanks!

jasongrout:
What is the current status of the benchmark tests? From the errors, it seems like they are broken, but looking back at other PRs, it seems like they were passing recently.

jasongrout (approved):
Thanks!

blink1073:
> What is the current status of the benchmark tests? From the errors, it seems like they are broken, but looking back at other PRs, it seems like they were passing recently.

@fcollonval is working on making them more robust in #11128

blink1073:

blink1073:

PR #11198, opened 2021-10-02
Rename "JupyterLab Theme" to "Theme"

jtpio:
## References

Drop the "JupyterLab" part of "JupyterLab Theme" which is visible in the settings menu.

In downstream distributions and custom applications, the word "JupyterLab" might not correspond to the `appName` and might sound a bit off.

Keeping only "Theme" sounds more succinct and should be enough.

"JupyterLab Light" and "JupyterLab Dark" can stay the same since they correspond to themes provided by the upstream JupyterLab repo.

## Code changes

Rename the "JupyterLab Theme" menu entry.

### Before

image

### After

image

None


jasongrout (approved):
+1 on the change. I suppose we have to make this a 4.0 change because we've frozen the strings for translation?

blink1073:
I thought we had said minor releases can have new strings?  It was only point releases that should be stable.

blink1073:
cc @fcollonval @krassowski

krassowski:
Yes, the translations are pinned down to minor. I think it can go into 3.2. Interesting that the check_i18n did not fail here.

fcollonval:
> Yes, the translations are pinned down to minor. I think it can go into 3.2. 

:+1:

> Interesting that the check_i18n did not fail here.

This is expected as the target branch is `master`. In that case, that job will always pass. The i18n check will fail only on PRs targeting `#.#.x` branches as they will result in new patch versions.

The job log computes a change in hashes  for this PR is:

```
+ [[ 86c1d58208e78399cc3ef85bd8f7f630cd6111d5fe3f9e87ab3b5e4a1feabf66 != \1\8\1\1\d\a\8\8\4\f\2\3\2\c\1\3\6\f\7\3\b\9\a\0\7\6\f\b\0\1\b\a\e\f\4\1\1\5\8\2\f\e\b\1\a\b\f\9\e\5\6\4\b\5\b\d\5\b\7\4\d\3\8\a ]]
```

blink1073:

PR #11202, opened 2021-10-05
refactor window.open to make it work also in desktop app

mbektas:
Export Notebook downloads are not working in Desktop App as reported in https://github.com/jupyterlab/jupyterlab_app/issues/270 . Because, in Electron `window.open` is returning null and setting a URL later on fails.

## Code changes

Instead of creating popup with no URL first and then setting a URL, now download URL is set directly in `window.open` call.

## User-facing changes

none

none


blink1073:

PR #11205, opened 2021-10-05
Handle case when JupyterHub returns 424 for not running server

yuvipanda:
## References

Companion to https://github.com/jupyterhub/jupyterhub/pull/3636,
should be merged only if that is.

## Code changes

Treat a 424 response from the server as 'server is not running' as well, not just a 503.

## User-facing changes

Users will see a pop-up about restarting their server correctly once
https://github.com/jupyterhub/jupyterhub/pull/3636 lands.

If the change in JupyterHub lands, I think people using previous versions of JupyterLab
will not see a popup asking them to restart their server. However, once this change lands,
JupyterLab can detect this problem in old and new versions of JupyterHub. So in some ways,
this is a backwards compatibility break in JupyterHub.

### DIFF
--- a/packages/services/src/kernel/manager.ts
+++ b/packages/services/src/kernel/manager.ts
@@ -239,11 +239,13 @@ export class KernelManager extends BaseManager implements Kernel.IManager {
     try {
       models = await listRunning(this.serverSettings);
     } catch (err) {
-      // Check for a network error, or a 503 error, which is returned
-      // by a JupyterHub when a server is shut down.
+      // Handle network errors, as well as cases where we are on a
+      // JupyterHub and the server is not running. JupyterHub returns a
+      // 503 (<2.0) or 404 (>2.0) in that case.
       if (
         err instanceof ServerConnection.NetworkError ||
-        err.response?.status === 503
+        err.response?.status === 503 ||
+        err.response?.status === 404
       ) {
         this._connectionFailure.emit(err);
       }
--- a/packages/services/src/kernel/manager.ts
+++ b/packages/services/src/kernel/manager.ts
@@ -241,11 +241,11 @@ export class KernelManager extends BaseManager implements Kernel.IManager {
     } catch (err) {
       // Handle network errors, as well as cases where we are on a
       // JupyterHub and the server is not running. JupyterHub returns a
-      // 503 (<2.0) or 404 (>2.0) in that case.
+      // 503 (<2.0) or 424 (>2.0) in that case.
       if (
         err instanceof ServerConnection.NetworkError ||
         err.response?.status === 503 ||
-        err.response?.status === 404
+        err.response?.status === 424
       ) {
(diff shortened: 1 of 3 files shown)


minrk:
I'm envious of that benchmark report! I want to add this for the JupyterHub API!

packages/services/src/kernel/manager.ts
```diff
@@ -239,11 +239,13 @@ export class KernelManager extends BaseManager implements Kernel.IManager {
...
-      // Check for a network error, or a 503 error, which is returned
-      // by a JupyterHub when a server is shut down.
+      // Handle network errors, as well as cases where we are on a
+      // JupyterHub and the server is not running. JupyterHub returns a
+      // 503 (<2.0) or 404 (>2.0) in that case.
       if (
         err instanceof ServerConnection.NetworkError ||
-        err.response?.status === 503
+        err.response?.status === 503 ||
+        err.response?.status === 404
```
minrk, line 248: Since 404 can mean other things, I think it's worth further checking that we're in JupyterHub somehow. One option would be `if status === 404 and "JupyterHub" in response body`. I don't know if there are other signals in lab that the context happens to be JupyterHub.

Another negotiation with the server could be to say "if this gives a 404, double-check with a request to the most basic `/api/status` and if *that* fails, signal a connection error.

Alternately, it could also check for the redirect that normally only occurs with JupyterHub `(err.response?.status === 404 && error.response?.redirected && error.response?.url.indexOf("/hub/") >= 0)`. I'm not sure if I want to guarantee from the jupyterhub side that there will always be a redirect here, though.

Following discussion in https://github.com/jupyterhub/jupyterhub/pull/3636, though, I think it will be better if we can find a *non-*404 error code to use for this.

blink1073: We have [metadata](https://github.com/jupyterlab/jupyterlab/blob/6e655233fd745cfd9087f3b177fd57f0e27bc9b3/packages/hub-extension/src/index.ts#L43) available when we are run under hub, which are set in `pageconfig` [here](https://github.com/jupyterlab/jupyterlab/blob/6e655233fd745cfd9087f3b177fd57f0e27bc9b3/jupyterlab/labapp.py#L784)

yuvipanda: Based on conversation in https://github.com/jupyterhub/jupyterhub/pull/3636, we're trying 424 status code now - the less lab knows of hub internals the better!

minrk: Nice. Looks like we'll be going with `424` for the error code, which I think is unusual enough that just treating 424 like 503 ought to be enough, I think.

blink1073 (approved):
Thank you!  The release_check failure is being addressed in #11218

blink1073:

blink1073:

yuvipanda:
Thanks a lot, @blink1073 and @minrk!

PR #7596, opened 2019-12-05
Add null fileformat

telamonian:
## References

fixes  #5885

## Code changes

This adds a `null` `FileType`, and ensures that when getting such files their contents is not fetched. Previously, when getting a file there was no way to do anything other than get the entire contents eagerly.

This potentially solves a big issue for file types that want to fetch contents lazily. I'll start this PR out as WIP and test it against the hdf5 filetype in [jupyterlab/jupyterlab-hdf5](https://github.com/jupyterlab/jupyterlab-hdf5). I did inspect the contents stack (on both the frontend and the `notebook` backend) through which the `Contents.IFetchOptions` gets passed, and it does indeed appear that the minimal changes in this PR are sufficient to ensure that only metadata gets fetched when `FileType` is `null`

## User-facing changes

None

## Backwards-incompatible changes

None

PR #11229, opened 2021-10-05
Remove format from fetching options if null

### LINKED ISSUE #10757: `null` fileFormat raise `Format is invalid` error 400
## Description

Playing around with the new `null` fileFormat introduced in #7596, I noticed that it is in fact not usable in its current state. The url to get model contents (https://github.com/jupyterlab/jupyterlab/blob/master/packages/services/src/contents/index.ts#L1057) translates the option `format: null` into `...&format=null` (`null` is serialized as string by `objectToQueryString`: https://github.com/jupyterlab/jupyterlab/blob/e59aed79d0e61c53088f24b0e0eff8e64557690d/packages/coreutils/src/url.ts#L93).

This throws an error in the backend as the option `'null'` is not in the expected values (`'text'`, `'base64'` or `None`): https://github.com/jupyter-server/jupyter_server/blob/a0f98a0f1a3bc6567ca0a4d64954d1ac5ab1f127/jupyter_server/services/contents/handlers.py#L103.

## Reproduce

Make a GET request to `/api/contents` with `format=null`

## Expected behavior

I see two possbilities for the request to go through:
- `format` should be removed from the URL if `null`
- `format=null` should be converted to `None` on the server-side.

## Context

- Operating System and version: Linux Ubuntu 20.04
- Browser and version: Firefox 90.0
- JupyterLab version: 3.1.1
- `jupyter_server` version: 1.10.1

<pre>
[W 2021-08-02 12:54:20.531 ServerApp] 400 GET /api/contents/simple.h5?format=null&type=file&content=0&1627901660525 (127.0.0.1): Format 'null' is invalid
[W 2021-08-02 12:54:20.532 ServerApp] Format 'null' is invalid
</pre>
</details>

loichuder:
## References

Fix #10757 

Follows #7596

## Code changes

Format is removed from the fetching options if `null`. This avoids `format` to be misinterpreted (parsed as string `null`) on the server side.

## User-facing changes

None.

None.

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -660,9 +660,11 @@ export class Context<
    */
   private _revert(initializeModel: boolean = false): Promise<void> {
     const opts: Contents.IFetchOptions = {
-      format: this._factory.fileFormat,
       type: this._factory.contentType,
-      content: this._factory.fileFormat !== null
+      content: this._factory.fileFormat !== null,
+      ...(this._factory.fileFormat !== null
+        ? { format: this._factory.fileFormat }
+        : {})
     };
     const path = this._path;
     const model = this._model;


packages/docregistry/src/context.ts
```diff
@@ -660,9 +660,11 @@ export class Context<
    */
   private _revert(initializeModel: boolean = false): Promise<void> {
     const opts: Contents.IFetchOptions = {
-      format: this._factory.fileFormat,
       type: this._factory.contentType,
-      content: this._factory.fileFormat !== null
+      content: this._factory.fileFormat !== null,
+      ...(this._factory.fileFormat !== null
+        ? { format: this._factory.fileFormat }
+        : {})
```
loichuder, line 667: In the end, I choose to handle it here (rather than [here](https://github.com/jupyterlab/jupyterlab/issues/10757#issuecomment-915784160))  to keep the `null` fileFormat logic at the same place.

blink1073 (approved):
Thank you!

Issue #11042, opened 2021-09-09
Add more UI tests for the debugger


jtpio:
### Problem

We are seeing a couple of regressions for the debugger related to the new features, or caused by updates in the `ipykernel` and `xeus-python` kernels:

https://github.com/jupyterlab/jupyterlab/issues?q=is%3Aissue+is%3Aopen+label%3Apkg%3Adebugger++label%3Abug+

### Proposed Solution

Let's add more UI tests for the debugger.

### Additional context

This was initially started in https://github.com/jupyterlab/jupyterlab/pull/10406, but should be rebooted.

3coins:
@jtpio 
Does this require Galata based UI tests? Can you assign this to me.

jtpio:
Yes ideally we can use Galata to check the sidebar renders properly for example.

Thanks @3coins, I've assigned to you!

3coins:
@jtpio, @fcollonval 
Do you know if there is a way to change the kernel in galata tests? I believe there is no page object/helper for the kernel changer right now, is this something I should add as a part of these tests?

fcollonval:
Hey @3coins there is no such helper indeed. It can be interesting to add an helper indeed.

PR #11250, opened 2021-10-07
UI tests for debugger

3coins:
## References
This PR fixes [#11042](https://github.com/jupyterlab/jupyterlab/issues/11042)


3coins:
@jtpio 
Can you review the changes. Also, it seems like the galata is adding "darwin" to the snapshots, and other tests have "linux" in the snapshot names. How do I resolve this? Should I wait for the CI to kick in and pick and save snapshots from that run?

jtpio:
Thanks @3coins for this!

> Should I wait for the CI to kick in and pick and save snapshots from that run?

Yes we usually take the snapshots from CI and check them into the repo, since generating them from a local environment might not generate the same result for everyone.

Issue #11260, opened 2021-10-07
Button color styles do not follow theme styling


3coins:
## Description
Button styling is inconsistently applied within the Jupyter notebooks. e.g., inside the kernel switcher dialog, the **Submit** and **Cancel** buttons have hardcoded colors that do not change when a new theme is installed. In addition, the material colors selected for normal (500) and hover (600) states are not accessible with a white font color.

*Notice the difference in background colors for the toolbar button at the top left and the primary button in the notebook.*

<img width="796" alt="buttons-theme-monokai" src="https://user-images.githubusercontent.com/289369/136457365-92b46296-03df-4596-86dd-60e2c1675654.png">

## Reproduce

1. Install a new theme
```bash
jupyter labextension install @arbennett/base16-monokai
```
2. Open a new notebook, and add this in a markdown cell
```html
<button class="jp-mod-accept jp-mod-styled">Button (Accept)</button><br/><br/>
<button class="jp-mod-warn jp-mod-styled">Button (Warn)</button><br/><br/>
<button class="jp-mod-reject jp-mod-styled">Button (Reject)</button>
```
3. Change theme to monokai, **Settings** > **Theme** > **base16-monokai**

## Context

- Operating System and version: All
- Browser and version: All
- JupyterLab version: All

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

PR #11264, opened 2021-10-07
Added button theme variables, updated button style

3coins:
## References
This PR fixes [#11260](https://github.com/jupyterlab/jupyterlab/issues/11260)

## Code changes

- [x] Introduce new theme variables for button styles in all variants (accept, reject, warn) and states (normal, hover, active)
- [x] For default values, use a sequence of material design colors that are WCAG accessible.
- [x] Update focus state with a 1px border color to match the non-hover/active state.

#### Before
buttons-before

#### After
buttons-after

### DIFF
--- a/packages/application/style/buttons.css
+++ b/packages/application/style/buttons.css
@@ -17,67 +17,67 @@ button {
 }
 
 button:focus-visible {
-  border: 1px solid var(--jp-button-accept-active-color, var(--jp-brand-color1));
+  border: 1px solid var(--jp-accept-color-active, var(--jp-brand-color1));
 }
 
 button.jp-mod-styled.jp-mod-accept {
-  background: var(--jp-button-accept-normal-color, var(--md-blue-500));
+  background: var(--jp-accept-color-normal, var(--md-blue-500));
   border: 0;
   color: white;
 }
 
 button.jp-mod-styled.jp-mod-accept:hover {
-  background: var(--jp-button-accept-hover-color, var(--md-blue-600));
+  background: var(--jp-accept-color-hover, var(--md-blue-600));
 }
 
 button.jp-mod-styled.jp-mod-accept:active {
-  background: var(--jp-button-accept-active-color, var(--md-blue-700));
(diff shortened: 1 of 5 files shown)


fcollonval:
Why introducing new variables instead of updating the basic color for `brand` and `warn`?
What about taking the chance of this PR to change all default values to themed variables (dropping using hard coded color that breaks theme for hover, focus and active state)?

3coins:
@fcollonval 
Are these variables acceptable?
```css
--jp-accept/brand-color-normal: var(--md-blue-700);
--jp-accept/brand-color-hover: var(--md-blue-800);
--jp-accept/brand-color-active: var(--md-blue-900);

--jp-warn/brand-color-normal: var(--md-red-700);
--jp-warn/brand-color-hover: var(--md-red-800);
--jp-warn/brand-color-active: var(--md-red-900);

--jp-reject/brand-color-normal: var(--md-grey-600);
--jp-reject/brand-color-hover: var(--md-grey-700);
--jp-reject/brand-color-active: var(--md-grey-800);
```

fcollonval:
Thanks @3coins those sound good to me (assuming the `/brand` is dropped).

blink1073:
> Check Release failure doesn't seem to be related to this change. One of the link check is failing, but I am able to open the link directly.

I kicked the build

3coins:
> I think we should see if we can fix this without new CSS variables. If the colors are the same in the light/dark theme, we can probably just update them to the more accessible colors for now. While that doesn't allow theme extensions to customize the button colors, it does address the "bug" dimension of this that the existing colors are not accessible.
> […](#)

This will move more in the direction of hard coded colors for all button styles than using the css variables. Is that an acceptable short term fix?

blink1073:
> with a bump for the theme packages minor version

All of the packages already have a minor bump in master

fcollonval:
> All of the packages already have a minor bump in master

Thanks @blink1073 I forgot about that one. So this one is ready for me.

PR #11268, opened 2021-10-08
Add missing changelog entry from 4.0.0a13

blink1073:
## References
https://github.com/jupyterlab/jupyterlab/pull/11267

## Code changes
None

## User-facing changes
Adds missing changelog entry

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ github_url: 'https://github.com/jupyterlab/jupyterlab/blob/master/CHANGELOG.md'
 
 ### Bugs fixed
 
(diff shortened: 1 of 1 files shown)


blink1073:
Best I can tell, some entries are now missing from the GitHub API.  When we ran the query to generate the changelog for [3.2.0rc0](https://github.com/blink1073/jupyter_releaser/runs/3828614538?check_suite_focus=true#step:6:337), this entry was there.  Now when I run the same query that was used in the draft release run (`github-activity jupyterlab/jupyterlab --since @jupyterlab/example-services-outputarea@3.1.9 --branch 3.2.x`), the entry is not there.  


When I run the query `github-activity jupyterlab/jupyterlab --since v4.0.0a12` (with no branch filter), none of the PRs associated with this change are available.

This loss of data is what forced me to make a manual release of 3.2.0rc0 yesterday, and this manual change today.

PR #11270, opened 2021-10-08
Clarify sidebar switching settings

joelostblom:
I had issues when trying to modify the sidebar settings because I didn't know the names of the different icons and they don't follow a predicatable pattern. This PR updates the outdated example and includes a note about the right-clicking of side bar icons, which I only found out myself after searching the GitHub repo and stumbling upon the changelog entry for when that was introduced.


blink1073:

Issue #9431, opened 2020-12-04
Memory leak with notebooks and ANSI colors in Firefox 


freud14:
Hi, I already did an issue in the Jupyter Notebook repo (https://github.com/jupyter/notebook/issues/5897) but it seems that the repo is more in a maintenance mode so I don't know if this will be addressed. Since this bug also affects Jupyter Lab, I told myself that I might as well make the issue for it anyway.

## Description

I seems that under specific circumstances, using ANSI colors repeatedly causes a huge memory leak in Firefox. The reproducing example does not trigger any memory use increase in Chromium.

Here is the screenshot when running the reproducing code:
![image](https://user-images.githubusercontent.com/1090012/101065001-9e125200-3562-11eb-93fd-014207c16a41.png)

## Reproduce

Here is the code to reproduce the bug.
```python
import time
import sys
for i in range(1, 101):
    for j in range(1, 1001):
        sys.stdout.write(f'\r\x1b[35mEpoch: \x1b[36m{i}/100 \x1b[35mStep: \x1b[36m{j}/1000 \x1b[35m100.00% \x1b[35m|\x1b[35mETA: \x1b[32m0.00s \x1b[35mloss:\x1b[94m 2.295335\x1b[35m acc:\x1b[94m 15.625000\x1b[0m')
        sys.stdout.flush()
        time.sleep(0.005)
    print()
```
For the first 20 iterations of the main loop (for i in range), memory usage is normal. At the 40th iteration, the memory usage of Firefox has increased for as much as 5GB. From there, it still increases until OOM.

Without the colors, the memory usage is normal. The example prints multiple lines. When printing a single line repeatedly, the bug is not triggered. My educated guess is that it has something to do with removing and adding nodes in the DOM.

## Expected behavior

Obviously, a constant memory usage as in Chromium would be preferable.

## Context

- Operating System and version: Ubuntu 20.04
- Browser and version: Firefox 83 and Chromium 87
- JupyterLab version: 2.2.9

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/home/fredy/opt/node-v12.19.0-linux-x64/bin
	/home/fredy/bin
	/home/fredy/.venv/base/bin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin
	/usr/games
	/usr/local/games
	/snap/bin

sys.path:
	/home/fredy/.venv/base/bin
[...]

PR #11273, opened 2021-10-10
Fix memory leak with notebooks and ANSI colors in Firefox (#9431)

freud14:
## References

Issue #9431

## Code changes

Do `/\x1b\[([^@-~]*?)([@-~])/g` instead of `/\x1b\[(.*?)([@-~])/g`

## User-facing changes

None

No breaking change.


freud14:
Hi @mgeier, thank you for your response. You're right that the question mark seems unnecessary.  I made the commit so that we can test it in binder. Will probably test it later. Not sure it will completely solve the bug tho.

freud14:
Hi @mgeier, I just tested without the question mark in binder but that didn't fix the bug.

Issue #11277, opened 2021-10-12
Extension upgrade script throws exception for certain files


ammgws:
## Description

Attempting to fix the `pydeck` extension so that it works on JupyterLab v3. 
(Currently it only works on v2.3, see issue: https://github.com/visgl/deck.gl/issues/5615)

Running the upgrade scipt as per the [extension migration guide](https://jupyterlab.readthedocs.io/en/stable/extension/extension_migration.html#jupyterlab-2-x-to-3-x) results in an exception.

**Disclaimer** I am a bit lost here so not sure if this is an issue with the pydeck extension itself having a broken (but somehow working) package file, or if it is something else. However after observing various `package.json` files on GitHub it does seem like the upgrade script makes assumptions that aren't always true hence leading to it throwing exceptions.

## Reproduce

1. pip install jupyterlab jupyter-packaging cookiecutter
2. git clone https://github.com/visgl/deck.gl
3. cd deck.gl/modules/jupyter-widget/
4. python -m jupyterlab.upgrade_extension .
5. Press enter and accept the default values at the prompts that show up
6. See error:
```
author_name [<author_name>]: 
author_email []: 
labextension_name [@deck.gl/jupyter-widget]: 
python_name [deck.gl_jupyter_widget]: 
project_short_description [Jupyter widget for rendering deck.gl in a Jupyter notebook]: 
has_settings [n]: 
has_server_extension [n]: 
has_binder [n]: 
repository [https://github.com/visgl/deck.gl.git]: 
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/ammgws/.virtualenvs/temp-jupterlab/lib/python3.9/site-packages/jupyterlab/upgrade_extension.py", line 210, in <module>
    update_extension(args.path, args.branch, args.no_input==False)
  File "/home/ammgws/.virtualenvs/temp-jupterlab/lib/python3.9/site-packages/jupyterlab/upgrade_extension.py", line 100, in update_extension
    data['devDependencies'][key] = value
KeyError: 'devDependencies'
```

## Expected behavior

Upgrade script completes without throwing Exceptions

## Context

jupyter-client      7.0.6
jupyter-core        4.8.1
jupyter-packaging   0.10.6
jupyter-server      1.11.1
[...]

PR #11278, opened 2021-10-12
Extension upgrade script: Avoid throwing exceptions for certain package.json files

ammgws:
## References

Resolves the exceptions thrown in #11277

**Disclaimer:** I am unable to test whether this actually produces a valid package file, as I have not figured out how to successfully build even the current version of the `pydeck` extension mentioned in the above issue. However looking at various `package.json` files around the internet, the script does seem to make assumptions that may not necessarily be true. Other sections of the upgrade script might also have similar bugs in it.

## Code changes

- upgrade script assumes `sideEffects` in `package.json` is a `list`, where as in pydeck and others I found on the internet it can also just be a `bool`.
- upgrade script assumes `devDependencies` exists

## User-facing changes

None

None


blink1073:

Issue #3275, opened 2017-11-23
Consider making Jupyter Lab mobile friendly


DamienPetrilli:
Hi,

I just discovered and started messing around with Jupyter Lab so I might have missed something. However, while Jupyter Notebooks mostly worked on touch based screen, Jupyter Lab is not usable on an iPad 12.9”. It seems the UI relies on double clicks and other mouse driven events which are not available (for example I can’t open a folder, one tap selects it, double tap trigger the rename folder).

I am not sure if it’s because it’s an Alpha or if it’s because touch screens weren’t in the priority list.

More and more people are using tablets nowadays and it would be great if Jupyter Lab supported touch screen based devices. Taking a tablet and connect to a Jupyter Notebook remotely is super useful.

Cheers

dwt:
Some more info of things that don‘t work well / workarounds

* Navigating the files tab via the arrow keys does not work - opening files with enter does. That means that a mix of gestures and keyboard does get me around
* in notebooks, arrow keys to change the selected cell don‘t work (if you are not in editing mode)
* in notebooks: arrow keys Do Not allow you to navigate to a different cell
* cannot use escape to leave the focus of a cell (as written above)

There is probably more, but I haven‘t found it yet.

juniorrojas:
What repos / directories / documentation should one get familiar with in order to be able to extend the interface (to make it work on mobile, for instance). Would it require making changes in external repos like phosphorjs?

chrism2671:
I’m trying to use it on my iPad today and I can’t seem to open files. Anyone got any suggestions?

bollwyvl:
Try a long tap to trigger the context menu, then "open". Works on chrome,
at least...

saptarshiguha:
Another thing not working is dragging window panes to resize. It would be really nice to make Jupyterlab mobile friendly making it truly device free.

dentarthur:
It seemed ok for very briefly running mybinder sample on Samsung 10.5" Android Tablet.

This environment with keyboard is quite suitable for most things despite small screen and no mouse. But it is normal to zoom the screen with Android finger movements both to see details more clearly and to push buttons more easily.

Result of doing so is that control bar with run button at top disappears. (Returns when unzoom).

Should default to freezing the control bar at top of screen despite any zooming of entire screen done at Android end. (Or provide easily noticed instructions on how to configure this).

[...]

PR #11281, opened 2021-10-12
Update the latest `@lumino` packages

jtpio:
## References

Since the latest release adds basic support for touch events: https://github.com/jupyterlab/lumino/releases/tag/v2021.9.30

So this might help improve the lab experience on mobile devices (#3275)

## Code changes

Dependency update with:

```
jlpm run update:dependency --regex @lumino/ latest
jlpm
jlpm run integrity --force
```

## User-facing changes

Can be tested with the Binder link: https://mybinder.org/v2/gh/jtpio/jupyterlab/update-lumino?urlpath=lab-dev


blink1073:
Thanks!  I think it is too late to add this to 3.2, so it will have to wait for 4.0.

Issue #7560, opened 2019-11-26
Kernel info request is blocked after a long execute followed by a page reload


jtpio:
For now the `kernel_info_request` message is sent on the `shell` channel.

This can lead to the following behavior when a cell is executed, keeping the kernel in a `busy` state for some time, and the page reloaded right after:

1. Open a notebook with `ipykernel` as the kernel
2. Execute a cell that takes some time to execute. For example:

```python
import time
time.sleep(99999)
```

3. Reload the page
4. After the page is reloaded, JupyterLab sends a new `kernel_info_request` on the `shell` channel.
5. However since the kernel is currently executing the code from the previous `execute_request`, it can't process the `kernel_info_request` and send the `kernel_info_reply`
6. The kernel is then restarted every 10 seconds, and a new WebSocket is recreated.

![kernel-info-blocked](https://user-images.githubusercontent.com/591645/69548402-93a89a00-0f97-11ea-8d45-6457f3fcc070.gif)

This is also visible in the server logs:
```
[I 15:14:33.500 LabApp] Adapting from protocol version 5.1 (kernel f391e165-dad1-43b0-8f8d-ea426dfa4077) to 5.3 (client).
[I 15:14:43.532 LabApp] Adapting from protocol version 5.1 (kernel f391e165-dad1-43b0-8f8d-ea426dfa4077) to 5.3 (client).
[I 15:14:53.563 LabApp] Adapting from protocol version 5.1 (kernel f391e165-dad1-43b0-8f8d-ea426dfa4077) to 5.3 (client).
[I 15:15:03.571 LabApp] Adapting from protocol version 5.1 (kernel f391e165-dad1-43b0-8f8d-ea426dfa4077) to 5.3 (client).
[I 15:15:13.601 LabApp] Adapting from protocol version 5.1 (kernel f391e165-dad1-43b0-8f8d-ea426dfa4077) to 5.3 (client).
```

One solution would be to send the `kernel_info_request` on the `control` channel instead of `shell`. According to the ["Messaging in Jupyter" documentation](https://jupyter-client.readthedocs.io/en/latest/messaging.html):

> Control: This channel is identical to Shell, but operates on a separate socket to avoid queueing behind execution requests. The control channel is used for shutdown and restart messages, as well as for debugging messages.

jasongrout:
We do the kernel info request for several reasons:

1. It triggers a kernel status message, so that we can get into a known kernel state.
[...]

PR #11285, opened 2021-10-14
Add a usage request/reply on the Kernel Control channel

echarles:
This is an experimental PR to engage a discussion a how we should address the addition of new Kernel messages.

> This PR is not meant to be merged.

## Context

Third party developer may need to add optional/extension kernel messages and manage those message from JupyterLab. For example, in the case of a Remote Kernel, we can imagine the need to show the user the CPU and Memory usage for that specific Kernel (not the CPU/Memory of the server, but of each of the Kernels). A bit like https://github.com/jupyter-server/jupyter-resource-usage or https://github.com/jtpio/jupyterlab-system-monitor, but on a per- remote-kernel basis.

I have discussed with @Zsailer, @kevin-bates, @jess-x, @mlucool and @mwakaba2 options like OpenTelementry, but at the end, I have POCed another approach where a new Kernel protocol message (`usage_request` and `usage_reply`) would be transmitted on the non-blocking Control channel (with a new message type implemented in ipykernel).

## usage_request POC on JupyterLab

The `@jupyterlab/services` package is not meant to be extensible (as far as I understand). Therefor, I had to add a few types and methods (see the changes in this PR) in the `services` source to be able to query the Kernel usage from an extension and generate this view.

Screenshot 2021-10-14 at 08 00 13

## [DISCUSS] Allow extension authors to add kernel messages

Let's forget for now the question related to the validity for extension authors to add new kernel messages..

What would it mean for JupyterLab to support a more general case `Allow extension authors to add kernel messages`? 

cc/  @jasongrout @blink1073 @jtpio @vidartf @afshin @Carreau

### DIFF
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -761,6 +761,30 @@ export class KernelConnection implements Kernel.IKernelConnection {
     >;
   }
 
+  requestUsage(
+    content: KernelMessage.IUsageRequestMsg['content'],
+    disposeOnDone: boolean = true
+  ): Kernel.IControlFuture<
+    KernelMessage.IUsageRequestMsg,
+    KernelMessage.IUsageReplyMsg
+  > {
+    const msg = KernelMessage.createMessage({
+      msgType: 'usage_request',
+      channel: 'control',
+      username: this._username,
+      session: this._clientId,
+      content
+    });
+    return this.sendControlMessage(
+      msg,
+      true,
+      disposeOnDone
+    ) as Kernel.IControlFuture<
+      KernelMessage.IUsageRequestMsg,
+      KernelMessage.IUsageReplyMsg
+    >;
+  }
+
   /**
    * Send an `is_complete_request` message.
    *
--- a/packages/services/src/kernel/kernel.ts
+++ b/packages/services/src/kernel/kernel.ts
@@ -325,6 +325,14 @@ export interface IKernelConnection extends IObservableDisposable {
     KernelMessage.IDebugReplyMsg
   >;
 
+  requestUsage(
+    content: KernelMessage.IUsageRequestMsg['content'],
+    disposeOnDone?: boolean
+  ): IControlFuture<
+    KernelMessage.IUsageRequestMsg,
+    KernelMessage.IUsageReplyMsg
+  >;
+
   /**
    * Send an `is_complete_request` message.
    *
--- a/packages/services/src/kernel/messages.ts
+++ b/packages/services/src/kernel/messages.ts
@@ -109,6 +109,10 @@ export function createMessage<T extends IDebugRequestMsg>(
   options: IOptions<T>
 ): T;
 
+export function createMessage<T extends IUsageRequestMsg>(
+  options: IOptions<T>
+): T;
+
 /**
  * @hidden
  * #### Notes
@@ -117,6 +121,10 @@ export function createMessage<T extends IDebugRequestMsg>(
  * part of the public API, and may change without notice.
  */
 export function createMessage<T extends IDebugReplyMsg>(
+  options: IOptions<T>
+): T;
+
+export function createMessage<T extends IUsageReplyMsg>(
   options: IOptions<T>
 ): T;
 
@@ -183,7 +191,11 @@ export type ShellMessageType =
  * kernel message specification. As such, debug message types are *NOT*
  * considered part of the public API, and may change without notice.
  */
-export type ControlMessageType = 'debug_request' | 'debug_reply';
+export type ControlMessageType =
+  | 'usage_request'
+  | 'usage_reply'
+  | 'debug_request'
+  | 'debug_reply';
 
 /**
  * IOPub message types.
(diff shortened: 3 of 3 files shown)


jtpio:
Thanks @echarles for checking and reporting.

This issue of messages being queued and blocked is similar to how the kernel info request is sent at the moment (on `shell`): https://github.com/jupyterlab/jupyterlab/issues/7560

For the debugger this was solved by sending the messages on `control`, which are then processed in a different thread.

echarles:
We have discussed this experiment during the last developer meeting and it was agreed to move forward with this, adding a method to send new usage request on the control channel.

I have undrafted this PR and changed the name to state that intent.

This is ready for review.

echarles:
cc/ @jasongrout

echarles:
An usage requrest/request implementation example for ipykernel is available on https://github.com/ipython/ipykernel/pull/805

jtpio:
> in the case of a Remote Kernel

Just to be sure to understand the need for this: it looks like remote kernels are the main motivation for these additional kernel messages?

In other approaches like https://github.com/jupyter-server/jupyter-resource-usage the server does the reporting. assuming kernels are started on the same machine so the memory and cpu usage can be inspected with `psutil`. Which is indeed an issue when the kernel is not located next to the server.

jtpio:
It would be interesting to check how Colab handles this since they display RAM / Disk usage:

image

echarles:
> It would be interesting to check how Colab handles this since they display RAM / Disk usage:

@blois ?

echarles:
This is ready for review.

Getting this merged will benefit extension developers running remote kernels.
We can also let https://github.com/jupyter-server/jupyter-resource-usage evolve to use this new feature.

Happy to go through an question/comment on this.

davidbrochart:
Maybe we should take a step back and reconsider if comms could transit through the control channel, and this way have a generic solution? I have the feeling this is addressing a specific need and that it would need to be better specified in the kernel protocol, since it would add a new message type.

echarles:
> Maybe we should take a step back and reconsider if comms could transit through the control channel, and this way have a generic solution?

Comm on the control channel would allow to have a non-blocking solution, which can be very useful. We can certainly open discussion around that.

This should however not block this PR as the implemented methods are there to hide those details to the extension developers who doesn't know if through which channels the request/reply are flowing.

davidbrochart:
But before proceeding with the implementation, shouldn't we request an addition to the kernel protocol, and specify the new message type?
https://github.com/ipython/ipykernel/pull/805 looks like a specification of this message type, although it's not very clear (e.g. `cpu_percent` not always present in the reply).

echarles:
> But before proceeding with the implementation, shouldn't we request an addition to the kernel protocol, and specify the new message type?

What would be the process for that?

davidbrochart:
Maybe a JEP, like for the [debugger protocol](https://github.com/jupyter/enhancement-proposals/blob/cc60f08a991054a8a429d527a95bb8f207eb7eaa/47-jupyter-debugger-protocol/jupyter-debugger-protocol.md)? @JohanMabille may know better.

JohanMabille:
@echarles you need to open a PR with your proposal on the [Jupyter Enhancement Proposal repo](https://github.com/jupyter/enhancement-proposals). You can use [this JEP](https://github.com/jupyter/enhancement-proposals/blob/cc60f08a991054a8a429d527a95bb8f207eb7eaa/47-jupyter-debugger-protocol/jupyter-debugger-protocol.md) as a template.

echarles:
Another approach would be to make the jupyterlab services more extensible, at least on the control channel level, so that extension authors can transit custom control messages. The kernel would be responsible to reply (or not) to those custom message.

What do people think of that approach?

JohanMabille:
I think that even in this case we would need an additional message in the protocol (`user_request` / `custom_request`) as a placeholder for extension messages (meaning the content would not be specified). This would allow kernels to distinguish between extension messages, and messages they should not handle at all. Besides, this combined with additional fields in thje `kernel_info_reply` would allow the frontend to dynamically discover what features the kernel supports.

We also have to be careful with the kind of requests an extension may send, the `control` channel should not process request that may block it for a non ngligible amount of time.

fcollonval:
This is part of JupyterLab 3.3 wish list, do you think it will be complete by the end of January?

echarles:
No, this feature is being now developed in an external extension to get something quickly without waiting on a potential JEP.

I have removed this issue from the list on https://github.com/jupyterlab/jupyterlab/issues/11643

We can leave this open for further discussion on how to get that into core at some point.

Carreau:
One thing to keep in mind, especially WRT your screenshot is values as a function of time. 
I'm fine having a request for a single value to the kernel but you may want to consider having a different endpoint ofr the server to keep records.

Carreau:
No objections for me, I think this might need a rebase to fix test though you did not seem to have checked "allow edit form maintainer", I don't seem to be able to push a rebased branch

echarles:
I will close this one as we are not going to go further with this specific implementation. A potential more generic follow-up can be discussed in https://github.com/jupyterlab/jupyterlab/issues/12117

jweill-aws:
Closing per @echarles at triage.

jasongrout:
Should we also close https://github.com/jupyterlab/jupyterlab/issues/11455 ?

PR #11291, opened 2021-10-15
Fix for terminal theme style

3coins:
## References
Reinstate terminal style fix from PR [#8994](https://github.com/jupyterlab/jupyterlab/pull/8994) with additional updates to accommodate terminal theme selection.

## Code changes

- [x] Add a data attribute to terminal widget to track the selected theme
- [x] Add a border to the terminal canvas elements that match the resolved theme selection
- [x] Add UI test for terminal in light and gray themes 

#### Before

Terminal

#### After

Terminal

None

### DIFF
--- a/galata/test/jupyterlab/terminal.test.ts
+++ b/galata/test/jupyterlab/terminal.test.ts
@@ -1,4 +1,4 @@
-import { IJupyterLabPageFixture, test } from '@jupyterlab/galata';
+import { test } from '@jupyterlab/galata';
 import { expect } from '@playwright/test';
 
 test.describe('Terminal Tests', () => {


3coins:
> @3coins would you mind adding screenshots before VS after for this one? It would also be good to take screenshots of the terminal in galata tests (to be created) for both themes. So we can track changes on the look and feel.

This was reported by Brian and is a finicky bug to reproduce; I can add snapshots from Brian's original PR for context. Will also add Galata tests to test this change.

3coins:
> Should the border be removed?
> 
> In light theme, I'm not seeing any border. But I see one with the dark theme because the background is not black:
> 
> image
> 
> This does not look pretty. I would expect no border.
> 
> > I'm using Firefox 93

I see the problem in Firefox. I can fix this by matching the border color to the terminal background color; this doesn't seem scalable for custom themes, so might have to look at applying this dynamically for a permanent solution. Here is the original issue from xterm repo that discusses the issue, and similar solutions people have applied.
https://github.com/xtermjs/xterm.js/issues/1974

PR #10244, opened 2021-05-14
Completer: expose `activeIndex` and `indexChanged`

hbcarlos:
## References

Solves #10243

## Code changes

Adds a new signal triggered when the `_activeIndex` in the `Completer` changes.
Makes public `_activeIndex' (read-only).
Makes the `connector` in `CompletionHandler` more generic.

## User-facing changes

N/A

## Backwards-incompatible changes

Maybe changing types of `get/set connector` in `CompletionHandler`, not sure.

PR #11298, opened 2021-10-16
Emit `indexChanged` on model state updates

krassowski:
## References

Active index can be changed not only by cycling, but also by typing (or other model updates).

Follow up after #10244, discovered when debugging a docpanel issue in LSP.

## Code changes

- emit `indexChanged` on model state updates.

## User-facing changes

None

None.


blink1073:

PR #11334, opened 2021-10-21
ensure debug ability is enabled

eastonsuo:
## References
User will be confused if there is a debug button (for JupyterLab >=3.0), yet not able to click
image

## Code changes

change requirements

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,7 @@ install_requires =
     jupyter_server~=1.4
     nbclassic~=0.2
     jinja2>=2.1
+    ipykernel~=6.0
 
 [options.extras_require]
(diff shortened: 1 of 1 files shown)


jtpio:
Maybe we'll need to add `ipykernel` as an explicit dependency if we want users to have at least one kernel installed by default, as noticed in https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1024011925.

Raising the question here for visibility.

jtpio:
Closing as `ipykernel` is now an explicit dependency in `jupyterlab` (added in https://github.com/jupyterlab/jupyterlab/pull/11894):

https://github.com/jupyterlab/jupyterlab/blob/2eb7328478604fa799969ed2b588a0f1a4fc25dc/setup.cfg#L36

Thanks @eastonsuo!

Issue #10452, opened 2021-06-19
Show pending/executing in TOC


goerz:
The [classic notebook TOC2 extension](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/toc2/README.html) would show a color highlight to indicate which section was being edited/pending/executing. This was extremely useful for "Run all", as the TOC would serve as a nice progress indicator.

In the latest version of the jupyterlab TOC, in the form that is now built-in to Jupyterlab 3 (I'm running 3.1.0a13 specifically), the TOC already shows a color indicator for the active cell. I'm very impressed with all the work that has gone into the version 3 user interface. A big thanks to everyone who's been putting work into this! The TOC has definitely much progressed since the early versions of the [third-party-plugin](https://github.com/jupyterlab/jupyterlab-toc). It's *almost* at the functionality of the classic TOC. So: Would it be feasible to extend the "active" indicator with color-coded indicators for pending/executing, like in the classic TOC?

I'd also love the see an option to skip the main title in the TOC numbering, cf https://github.com/jupyterlab/jupyterlab-toc/issues/66 and https://github.com/jupyterlab/jupyterlab-toc/issues/97.

krassowski:
Nice ideas!

> I'd also love the see an option to skip the main title in the TOC numbering

There is a PR open for it, which is almost ready (basically waiting to be merged): https://github.com/jupyterlab/jupyterlab/pull/9994; would you like to have a quick look and see if this implements the expected functionality well?

Minor detail: did you wish to cross-reference issues from another repository?

goerz:
> > I'd also love the see an option to skip the main title in the TOC numbering
> 
> There is a PR open for it, which is almost ready (basically waiting to be merged): #9994; would you like to have a quick look and see if this implements the expected functionality well?

The screenshots in the PR look exactly like what I'd be hoping for, as far as section numbering is concerned 👍 

> Minor detail: did you wish to cross-reference issues from another repository?

You mean https://github.com/jupyterlab/retrolab/issues/168? I did want to crosslink from that issue to here, yes. Seems like Github creates the backlink automatically.

krassowski:
No, I mean issues 66 (enable mathjax) and 97 (clarify extension installation language).

goerz:
[...]

PR #11356, opened 2021-10-27
Toc running cell indicator

andrewfulton9:
## References

Related to #10452

TODO:

- [x] Update doc string
- [x] Update screencast
- [x] Add UI tests

## Code changes

- Make the notebook toc generator as a class so it clarify it
- Store a running cells array in that new generator class (as it does not make really sense in the options container)  
  It got updated by listening to the `NotebookActions.executionScheduled/executed` signals.
- Use a three state `isRunning` attribute on the heading to distinguish: _Idle_, _Scheduled_, _Running_
- Set `data-running` attribute on the toc item with the `isRunning` status
- Use pure CSS to define the indicator, so that it can be tuned easily.
- Add '*' to the scheduled/running prompts if code cell are displayed.

## User-facing changes

Adds indicator to ToC headings if they or a cell in their section is scheduled to run, or running.

toc_running_status

`Registry.IOptionsManager` is now a _interface_ not a _abstract class_ to be coherent with the naming (and that nothing was defined in it).


fcollonval (dismissed):
From the implementation point of view, we should avoid crowding the metadata; especially with transient information that have no meaning in the file. So I would recommend using the execution signals as used by the execution indicator: https://github.com/jupyterlab/jupyterlab/blob/e4e323992d24f6c5e48d181381e23c547b665b15/packages/notebook/src/actions.tsx#L96

andrewfulton9:
Thanks for the feedback!

> * What is displayed if multiple cells are queued for execution?

There will be a circle on each section that has pending or executing cells. I couldn't figure out a way to differentiate between the two unfortunately. It could be assumed the top most cell if the one that's running, but only if cells are run in order. 

> * If the running cell(s) are collapsed, is there an execution indicator?

Yes, the execution indicator will still be on the section in the ToC. If the ToC sections are collapsed, the indicator will be on the highest visible level.

> From the implementation point of view, we should avoid crowding the metadata; especially with transient information that have no meaning in the file. So I would recommend using the execution signals as used by the execution indicator:

I just pushed up some edits that remove the use of metadata, using a new `OptionsManager` attribute that holds an array of running cells to track which ones are scheduled for execution. It responds a little slower, but it keeps metadata clear. What do you think about this method, @fcollonval?

fcollonval:
The new approach looks great @andrewfulton9 
Regarding how to deal with queued vs currently running cell, I will play a bit with this PR. But I'm convinced we can figure out the status by looking at the order in the running cells array. 

I'll look to that this week.

fcollonval:
Thanks Andrew for merging my changes. I listed the 3 remaining tasks and updated the initial comment to reflect the current status. Do you want to tackle adding UI tests?

andrewfulton9:
I have tested the new changes and everything looks good to me. Thanks for the additions @fcollonval!

fcollonval:
@jweill-aws would you mind reviewing this one?

jweill-aws:
@fcollonval I've reviewed this including on Binder. I'm not authorized to merge this PR, but it looks good to me. Thanks @andrewfulton9 !

andrewfulton9:
@meeseeksdev please backport to 3.3.x

echarles:

PR #11358, opened 2021-10-27
Fix Handling of WebSocket Startup Errors

blink1073:
## References
Required for slow starting kernels work proposed https://github.com/jupyter-server/jupyter_server/issues/592.
The linked proposal fixes a race condition in the server for slow starting kernels, but it changes the behavior of 
how an error during kernel process start is handled.  The error is surfaced on WebSocket connect instead of 
as a response to a POST to `/api/sessions`.  The change in this PR detects those errors and displays them to the user.
This is also a feature parity with the classic notebook, which had similar handling for errors seen during initial WebSocket
[connect](https://github.com/jupyter/notebook/blob/2cfff07a39fa486a3f05c26b400fa26e1802a053/notebook/static/services/kernels/kernel.js#L479).

## Code changes
Add handling of WebSocket startup errors in the `Kernel` class and display associated errors to the user in `SessionContext`.

## User-facing changes
Surface kernel WebSocket startup errors to the user.

None

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -652,12 +652,19 @@ export class SessionContext implements ISessionContext {
    */
   async restartKernel(): Promise<void> {
     const kernel = this.session?.kernel || null;
+    if (this._isRestarting) {
+      return;
+    }
     this._isRestarting = true;
     this._isReady = false;
     this._statusChanged.emit('restarting');
-    await this.session?.kernel?.restart();
+    try {
+      await this.session?.kernel?.restart();
+      this._isReady = true;
+    } catch (e) {
+      console.error(e);
+    }
     this._isRestarting = false;
-    this._isReady = true;
     this._statusChanged.emit(this.session?.kernel?.status || 'unknown');
     this._kernelChanged.emit({
       name: 'kernel',
@@ -832,9 +839,7 @@ export class SessionContext implements ISessionContext {
       this._pendingKernelName = model.name;
     }
 
-    if (this._session && !this._isTerminating) {
-      await this._shutdownSession();
-    } else if (!this._session) {
+    if (!this._session) {
       this._kernelChanged.emit({
         name: 'kernel',
         oldValue: null,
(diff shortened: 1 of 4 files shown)


blink1073:
A few issues I noticed while testing this:

- Restarting while kernel is starting gets you stuck in "restarting" status despite getting a 500 response.
- Without pending kernels, restart requests stack up on each other while a kernel is starting.
- Changing kernels results in a DELETE and a POST  on `/api/sessions` instead of a PATCH, which means we can't take advantage of the quick switching logic introduced in the [server PR](https://github.com/jupyter-server/jupyter_server/pull/593/files#diff-7d5f5e4dc0efd2904e11278697317826d4871a19e775c333e2321699cf512c7dL147).

These should be handled within the `SessionContext` object.

blink1073:
Fixed the restart and change kernel errors in the last commit.  That commit can be backported to 3.2.x

blink1073:
As discussed in today's meeting, it might make sense to target this for 3.2.x since the only API change is optional fields on an interface, and it allows folks to start experimenting with pending kernels.

jtpio:
As discussed in the Jupyter Server meeting today, it looks like we can get this change in `master` so it's part of the next `4.0` alpha release, and can be more easily tested with the other PRs in `jupyter_server` and `jupyter_client` :+1:

jtpio (approved):
Thanks!

The change looks sensible.

Just did a quick smoke test on Binder and the default behaviors such as starting, restarting and stopping the kernel seem to be working fine.

It would make sense to land this PR so it can be tested against the other components using the next `4.0.0ax` release.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -848,6 +853,17 @@ export class SessionContext implements ISessionContext {
       this._initStarted.resolve(void 0);
     }
 
+    // If we already have a session, just change the kernel.
+    if (this._session && !this._isTerminating) {
```
jtpio, line 857: Is there any side-effect to expect, now that the order of this condition has changed here?

blink1073: The desired effect was that we don't DELETE and POST, but instead PATCH to change kernels.  I confirmed this in my testing.

Zsailer:
@meeseeksmachine backport to 3.3.x

Zsailer:
@meeseeksmachine backport to 3.3.x

Zsailer:

Issue #11362, opened 2021-10-28
Add VSCode recommendation for Prettier extension


jweill-aws:
### Problem

Jupyter recommends the Prettier extension and invokes it as a pre-commit hook to style code. For people who use Visual Studio Code, we can install such an extension locally, but the Jupyter package does not [recommend](https://code.visualstudio.com/docs/editor/extension-marketplace#_extensions-view-filters) that developers do so.

### Proposed Solution

Add a `.vscode/extensions.json` file to recommend that the Prettier extension be installed. This will yield a non-modal popup if a VSCode user opens the repository and doesn't have the extension installed.

### Additional context

Microsoft extensions documentation: https://code.visualstudio.com/docs/editor/extension-marketplace#_extensions-view-filters

jtpio:
Fixed by #11363, thanks!

PR #11363, opened 2021-10-28
Adds recommendation to install Prettier extension

jweill-aws:
## References

#11362 

## Code changes

Adds an extension recommendation for the Prettier extension

## User-facing changes

Affects the developer experience: Visual Studio Code users who do not have the Prettier extension will be asked in a non-modal dialog whether they want to add it.

None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,6 +121,5 @@ junit.xml
 # ms IDE stuff
 *.code-workspace
 .history
-.vscode
 
 .jupyter_releaser_checkout
--- a/.gitignore
+++ b/.gitignore
@@ -121,5 +121,6 @@ junit.xml
 # ms IDE stuff
 *.code-workspace
 .history
+.vscode/settings.json
 
 .jupyter_releaser_checkout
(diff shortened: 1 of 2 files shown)


jasongrout:
I'm +1 on this.

.gitignore
```diff
@@ -121,6 +121,5 @@ junit.xml
 # ms IDE stuff
 *.code-workspace
 .history
-.vscode
```
jtpio, line 124: How will this removal affect VS Code users that might have a custom `settings.json`? For example to store `python.pythonPath` pointing to their virtual environment.

Looks like they might commit this file by mistake if it is not ignored anymore?

jweill-aws: Added `.vscode/settings.json` as a more specific exclusion.

jtpio:
Folks can also setup a global `.gitignore` on their machine (for example in `~/.gitignore`) to avoid committing other files in the `.vscode` folder (or from other IDEs).

jtpio (approved):
Thanks!

Issue #11325, opened 2021-10-20
Export or move PanelWithToolbar from debugger to apputils or ui-components


hbcarlos:
I'm using the accordion panel in another package in Jupyterlab, and I would like to reuse the PanelWithToolbar from the debugger to create the widgets inserted in the AccordionPanel.

The problem is that this widget is not exported. Since it is generic and could be used everywhere, I would like to move it to `apputils` or `ui-components`. Another possibility is to move it upstream to lumino widgets.

### Additional context
I'm talking about moving out of the debugger these classes:
[PanelWithToolbar](https://github.com/jupyterlab/jupyterlab/blob/8c67f5949af19503e447a83584885f919e115c47/packages/debugger/src/panels/panelwithtoolbar.ts#L16)
[Renderer](https://github.com/jupyterlab/jupyterlab/blob/8c67f5949af19503e447a83584885f919e115c47/packages/debugger/src/panels/panelbody.ts#L11)

cc @trungleduc @jtpio

fcollonval:
Closing as resolved by #11369

PR #11369, opened 2021-10-29
Creates a new accordion panel with toolbar in the ui-components package

hbcarlos:
## References
Solves #11325

## Code changes
This PR Creates a new widget based on the `AccordionPanel` that uses panels with toolbar in the accordion widget.
Moves part of the already existing logic in the `@jupyterlab/debugger` package to the `@jupyterlab/ui-components` package so we can reuse it everywhere. It also removes some fixes that were moved upstream.

## User-facing changes

Styling is more consistent with the other panels:

image

```ts
export interface ISidebar extends Panel
```

This implies that it exposes `addWidget`, `insertWidget`, `widgets` instead of `addItem`, `insertItem` and `items`. To be more coherent to a Lumino `Panel` widget API. It should not impact downstream extension, as they are unlikely to add panels to the debugger sidebar.


fcollonval:
@hbcarlos most of the logic I want is in. But I still have some issues with the latest commit:
- [ ] Click on toolbar items trigger (un-)collapse
- [ ] Styling of the header
- [ ] Check styling in dark mode

I'll look into it tomorrow.

hbcarlos:
Wow, looks fantastic. Thanks for working on this @fcollonval

fcollonval (approved):

fcollonval:
@trungleduc would you mind reviewing this PR as I contributed quite extensively on this one?

trungleduc:
Hello @fcollonval, I will do it.

trungleduc (commented):
It looks neat @hbcarlos @fcollonval! I only add some minor comments above.

fcollonval:
Thanks for the review @trungleduc

Issue #11005, opened 2021-09-03
"orig_nbformat: 1" in content metadata?


davidbrochart:
## Description

Saving a notebook sends `orig_nbformat: 1` in the content metadata.

## Reproduce

1. Launch Jupyter Lab
2. Create a new notebook
3. Save the notebook
4. In the network monitor of the browser, you can see a `PUT` at e.g. `http://localhost:8888/api/contents/Untitled.ipynb?1630687142498`, with a request payload of e.g.:

```
{"type":"notebook","format":"json","content":{"metadata":{"orig_nbformat":1,"language_info":{"name":"python","version":"3.9.6","mimetype":"text/x-python","codemirror_mode":{"name":"ipython","version":3},"pygments_lexer":"ipython3","nbconvert_exporter":"python","file_extension":".py"},"kernelspec":{"name":"python3","display_name":"Python 3 (ipykernel)","language":"python"}},"nbformat_minor":5,"nbformat":4,"cells":[{"cell_type":"code","source":"","metadata":{},"execution_count":null,"outputs":[],"id":"aa426ee8-3467-40bc-90db-2ab4d83935da"}]},"path":"Untitled3.ipynb"}
```

Note the `"content":{"metadata":{"orig_nbformat":1` in it.

The metadata is retrieved from here:
https://github.com/jupyterlab/jupyterlab/blob/29b052e5a699b30ba7450e998283ba599d561d59/packages/notebook/src/model.ts#L265
And set by the shared model here:
https://github.com/jupyterlab/jupyterlab/blob/2a99d596bc259f7fe294c8d911e27a9dffa6f239/packages/shared-models/src/ymodels.ts#L329

## Expected behavior

I don't think `orig_nbformat: 1` should be present in the metadata.

## Context

- Operating System and version: Linux Ubuntu 21.04
- Browser and version: Chrome 92.0.4515.159
- JupyterLab version: 4.0.0a7

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/home/david/mambaforge/envs/jupyverse-dev/bin
	/home/david/mambaforge/condabin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin
	/usr/games
	/usr/local/games
	/snap/bin

sys.path:
	/home/david/mambaforge/envs/jupyverse-dev/bin
	/home/david/mambaforge/envs/jupyverse-dev/lib/python39.zip
[...]

PR #11370, opened 2021-10-29
Make orig_nbformat optional #11005

nanoant:
According to https://github.com/jupyter/nbformat/blob/master/nbformat/v4/
schemas orig_nbformat is not required, and is only present when nbformat
upgrades the notebook from earlier major version. Therefore JupyterLab
should make it optional on its side and not set it anywhere by default.

## References

- Issue #11005

## Code changes

This PR addresses issue making `orig_nbformat` optional according to nbformat spec rather than setting it to 4 unconditionally.

## User-facing changes

None

None

### DIFF
--- a/galata/src/galata.ts
+++ b/galata/src/galata.ts
@@ -383,8 +383,7 @@ export namespace galata {
             nbconvert_exporter: 'python',
             pygments_lexer: 'ipython3',
             version: '3.8.0'
-          },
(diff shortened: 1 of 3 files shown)


nanoant:
> Thanks @nanoant
> 
> The failing UI test is not related (flaky test)

Thank you @fcollonval

fcollonval:

PR #11371, opened 2021-10-29
[3.2.x] Relax `@playright/test` dependency in Galata

jtpio:
## References

Manual backport of #11112


jtpio:
Seems like the UI tests fails so we might need to backport some other changes too.

jtpio:
According to the video in the galata report it seems to be failing to create a new terminal, and then times out:

https://user-images.githubusercontent.com/591645/140068600-acb40834-977e-4372-b622-9ca67d04362d.mp4

fcollonval:
The fix to return 200 for POST request when creating a terminal was not backported. I pushed cd90968. That should fix the tests.

jtpio:
Ah nice thanks @fcollonval. I was in the middle on applying this patch, as well as switching to `await request.allHeaders()` like in https://github.com/jupyterlab/jupyterlab/pull/11112/commits/08f5e7e6309f9bdfccde8bdd54ece26935d79274

PR #10729, opened 2021-07-29
Toc: Run nested code cells directly from markdown headings

jess-x:
## References

Improves Toc run cells feature https://github.com/jupyterlab/jupyterlab/issues/10581

## Code changes

- Removed runNestedCodeCells function
- Added new run cells logic, which references nested code cells from notebookTracker, not Toc headings

## User-facing changes

- Enable run nested cells feature when code cells headings are hidden in Toc

## Backwards-incompatible changes

N/A

Edit: 
Thx @andrewfulton9 for suggesting `notebookTracker` :)

PR #11375, opened 2021-10-29
Backport PR #10729 on branch 3.2.x (Toc: Run nested code cells directly from markdown headings)

jess-x:
… headings

## References
https://github.com/jupyterlab/jupyterlab/pull/10729#issuecomment-944992145
Backport PR #10729 on branch 3.2.x (Toc: Run nested code cells directly from markdown headings)

### DIFF
--- a/packages/toc/src/utils/headings.ts
+++ b/packages/toc/src/utils/headings.ts
@@ -101,7 +101,7 @@ const isNotebookHeading = (heading: any): heading is INotebookHeading => {
  * @param headings - list of headings
  * @param heading - heading
  */
-const runNestedCodeCells = (headings: IHeading[], heading: IHeading) => {
+ const runNestedCodeCells = (headings: IHeading[], heading: IHeading) => {
   let h: INotebookHeading;
   let i: number;
 
@@ -150,3 +150,4 @@ export { IHeading };
 export { INumberedHeading };
 export { INotebookHeading };
 export { runNestedCodeCells };
+export { isNotebookHeading };
--- a/packages/toc/src/utils/headings.ts
+++ b/packages/toc/src/utils/headings.ts
@@ -101,7 +101,7 @@ const isNotebookHeading = (heading: any): heading is INotebookHeading => {
  * @param headings - list of headings
  * @param heading - heading
  */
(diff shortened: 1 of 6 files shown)


packages/toc/src/utils/headings.ts
```diff
@@ -94,59 +93,10 @@ const isNotebookHeading = (heading: any): heading is INotebookHeading => {
...
-  }
-};
-
 /**
  * Exports.
  */
 export { IHeading };
 export { INumberedHeading };
 export { INotebookHeading };
-export { runNestedCodeCells };
```
krassowski, line 152: Should we keep this function (and its export) in this backport for backward compatibility?

jess-x: Good point!! let me see how to add those back

jess-x: Added them back manually :)

krassowski: Thanks!

krassowski (approved):
Thank you! I will highlight the added `@jupyterlab/cells` dependency for anyone reviewing after me before merging (I assume it is ok). I don't believe the failing visual regression test is related, but I also do not understand why it is failing. Seems to be related to https://github.com/jupyterlab/jupyterlab/pull/11360 in a way...

jess-x:
@krassowski thanks for the review!! 
Agree, the visual regression test doesn't look related to me either. JS flaky tests are also failing though  - I don't know it that is related, but remains me that I have noticed, even when my kernel is idle, `Run Cells` sometimes works intermittently. I might investigate that a bit separately, if more folk are also seeing that behavior.

jtpio:
There were a couple of UI tests fixes in https://github.com/jupyterlab/jupyterlab/pull/11112.

Which is being backported to 3.2.x in https://github.com/jupyterlab/jupyterlab/pull/11371. So maybe landing that backport PR first will help.

jtpio:
Just rebased to grab the latest fixes on the `3.2.x` branch.

jtpio (approved):
Thanks!

jess-x:
Thanks @jtpio!!

PR #11378, opened 2021-10-30
Fix links

krassowski:
## References

Follow-up on #11357 + changes links which used http to https.

I wonder if we should have:

```
grep -r -P 'http(s):(?:.*?)>`([^_])' docs/
```

somewhere in CI

## Code changes

None

## User-facing changes

None

None


jtpio (approved):
Thanks!

jtpio:

PR #11379, opened 2021-10-30
Delay cursor activity handling to fix completer edge case, add completer tests

### LINKED ISSUE #11377: Completer closes when typing a letter after dot in Python
When using code completion in **JupyterLab**, it currently works as follows:

- **Image 1**: You type e.g. `os.` and then press `TAB` to get all available suggestions. So far so good.
- **Image 2**: You press another letter e.g. `g` to further filter the suggestions. Unfortunately, the dropdown menu just disappears.
- **Image 3**: You need to press `TAB` a **second time** to get the filtered suggestions back.

image

I believe this behaviour was different a few versions back. Could we revert to the (I'd argue) "correct" behaviour, where the dropdown menu does not disappear, i.e. one has to press `TAB` only once?

---

System Details:
- Windows 10
- Python 3.8.7
- JupyterLab 3.2.1

krassowski:
## References

Fixes #11377.

## Code changes

Delay handling of `cursorActivity` event so that the updated value of the model (source code) is available for the handler.
The current version is a workaround at best, waiting for @dmonad and @hbcarlos to hear if they have better ideas (see https://github.com/jupyterlab/jupyterlab/issues/11377#issuecomment-955188655).

## User-facing changes

Completer will not close unexpectedly when typing.

None.


hbcarlos:
Hi @krassowski, thank you so much for debugging this.
I'm looking into this, but I'm a bit lost, and I need some time to figure out the best way to solve it.

galata/README.md
```diff
@@ -79,7 +79,7 @@ c.LabApp.expose_app_in_browser = True
 Then start the server with:
 
 ```bash
-jupyter lab --config jupyter_server_test_config.py
+jupyter lab --LabApp.config jupyter_server_test_config.py
```
jtpio, line 82: As mentioned by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/11391#issuecomment-957384056, do we need the explicit `LabApp`?

krassowski:
Superseded by a proper fix in #11398 :tada:

Issue #9615, opened 2021-01-13
Create notebook from (public ) URL


ptitzler:
### Problem

As a user, I commonly explore notebooks that are stored in a public location, for example on GitHub or S3. Currently there does not appear to be an (obvious) way to create a notebook in JupyterLab from an external source without "downloading" it first to the local machine.

Example:
A user would like to try the JupyterLab debugger example notebook, which is located here: https://github.com/jupyterlab/debugger/blob/master/examples/index.ipynb
To use the notebook, the user currently either has to clone the entire Git repository (using the git extension or outside of JupyterLab), or download the notebook source to the local machine (e.g. using a web browser).

### Proposed Solution

General workflow:
 - User provides a URL as an input (e.g. `https://raw.githubusercontent.com/jupyterlab/debugger/master/examples/index.ipynb`)
 - An attempt is made to retrieve the URL using an HTTP GET request.
 - If the request succeeded, a new notebook is created using the content of the downloaded file.

Potential integration points:
- Extend the "File" > "New Notebook" dialog to accept an optional text input. 
- Create new "File" > "New Notebook from URL" command

jasongrout:
For a GitHub-specific solution, you can look at https://github.com/jupyterlab/jupyterlab-github - it adds a new filebrowser and works with notebooks directly from GitHub.

Having a more general "Download and open this notebook" would be nice. I suppose it would be created in whatever the current filebrowser directory is.

krassowski:
And there is also: https://github.com/IBM/jupyterlab-s3-browser for S3

ptitzler:
> And there is also: https://github.com/IBM/jupyterlab-s3-browser for S3

I did test drive that extension a while ago for a different project. It's designed to provide access to protected S3 resources, which makes perfect sense. However, it's unfortunately not addressing the public resource scenario, where there's no need for an S3 client to access the files. (I'm not saying that the extension should!)

jtpio:
Maybe it's possible to do it client-side by fetching the content with the browser `fetch`, and upload via the contents manager from `@jupyterlab/services`. With the current directory being tracked by the file browser.

Linking to https://github.com/jupyterlab/jupyterlab/issues/10570 which expresses a similar idea.

Issue #10570, opened 2021-07-07
Open new notebook directly from gist / URL


TomNicholas:
### Problem

I would like it if I could open someone's else's notebook gist from github into my jupyterlab session, as a new notebook.

I regularly share work with my co-workers and want to execute their code, but currently the only way I know to do this is to copy-and-paste individual cells into my jupyterlab session, which is frustrating.

### Proposed Solution

Add an option to point to a github gist url (or even any url containing an .ipynb file) and automatically open a new notebook with the contents of those cells.

mlucool:
Related work: https://ricardodeazambuja.com/jupyter/2019/04/06/jupyter-loading-from-web/

jtpio:
Thanks for submitting the idea. 

This sounds like a good idea and it could be implemented in a third-party extension. Which could then live under the `jupyterlab-contrib` org if more people find it useful: https://github.com/jupyterlab-contrib

jtpio:
There is now:

- a new "Open from URL" entry in the file menu: added in https://github.com/jupyterlab/jupyterlab/pull/11387
- https://github.com/jupyterlab-contrib/jupyterlab-open-url-parameter : to allow specifying the URL as a search param

Closing this issue as fixed. Feel free to comment here if needed, if thanks!

PR #11387, opened 2021-11-01
Add a command to open a file from a URL

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/9615

This is particularly useful in lab-based frontends without an application server, for example JupyterLite: https://github.com/jupyterlite/jupyterlite/

Another use case for this will be for third-party extensions to be able to pass URLs to files as URL parameters when opening JupyterLab, and use the command to then download and open them.

This should also fix the use case of opening from a gist: https://github.com/jupyterlab/jupyterlab/issues/10570. The user would have to first retrieve the raw url of the file.

## Code changes

Add a new command to open a file from a URL.

Similar to `Open from Path`.

## User-facing changes

Open a notebook:

https://user-images.githubusercontent.com/591645/139747406-a77255e2-1fb4-41bc-ab7f-fd16269777ee.mp4

Open a csv file:

https://user-images.githubusercontent.com/591645/139747647-207bb305-7f9a-448f-9547-a08a7c018060.mp4

Open in the current folder:

https://user-images.githubusercontent.com/591645/139749051-e34476b0-473c-4768-b5d4-8f24f6149fdb.mp4

None


jtpio:
Maybe this can be moved to a separate plugin so it's easier to disable it if needed,

fcollonval:
Thanks @jtpio. This looks great!

> Maybe this can be moved to a separate plugin so it's easier to disable it if needed,

I think it is a good idea.

packages/filebrowser-extension/src/index.ts
```diff
@@ -691,6 +693,74 @@ export const launcherToolbarButton: JupyterFrontEndPlugin<void> = {
...
+    palette: ICommandPalette | null
+  ) => {
+    const { commands } = app;
+    const trans = translator.load('jupyterlab');
+    const { defaultBrowser: browser } = factory;
+    const command = CommandIDs.openUrl;
+
+    commands.addCommand(command, {
+      label: args =>
+        args.path ? trans.__('Open %1', args.path) : trans.__('Open from URL…'),
```
jtpio, line 717: Since the commands uploads the content to the server, we might want to rename the label to something like `Download and Open from URL...`

jtpio:
Marking as ready for review if folks would like to give a try on Binder:

https://mybinder.org/v2/gh/jtpio/jupyterlab/open-from-url?urlpath=lab-dev

You can try opening the following files as examples:

- https://raw.githubusercontent.com/jupyterlab/jupyterlab-demo/master/data/iris.csv
- https://raw.githubusercontent.com/jupyterlab/jupyterlab-demo/master/notebooks/Lorenz.ipynb

Adding to 4.0 for consideration. There is no breaking change so this could also be backported to `3.2.x`.

packages/filebrowser-extension/src/index.ts
fcollonval: Pinging @jasongrout for his input on the wording.

williamstein:
Just saw you demo this in the meeting.  +1 to this.  

Some trivial thoughts:

Often when people implement this sort of thing, they implement a mapping from user-friendly pages to the underlying raw content.   I did that a little bit when implementing this in CoCalc long ago, but only did a tiny amount.  I wonder if there is a generic library out there already that implements such a mapping (e.g., from a page for gist to the raw content)?

My other thought is -- what happens when the user tries to create a file and pastes a url in (e.g., they make the filename be a url).  In CoCalc it falls back to open the file using the url download functionality.

krassowski:
UX-wise, maybe we should merge it with `Open From Path`? The URL should be easy to distinguish from a relative fileystem path. In that case, I would like to have this in the core.

williamstein:
> UX-wise, maybe we should merge it with `Open From Path`?

The only issue I've seen with this is that many users might not know that they can paste a URL there (or not know which URL's are valid, a gist or only the raw gist?).  Adding a separate menu option documents the fact the functionality exists.

jtpio:
We probably don't want to support opening gists in core JupyterLab.

However a third-party extension could implement that mapping and use the `filebrowser:open-url` command (added in this PR) to open the files.

packages/filebrowser-extension/src/index.ts
```diff
@@ -691,6 +693,85 @@ export const launcherToolbarButton: JupyterFrontEndPlugin<void> = {
...
+      caption: args =>
+        args.path ? trans.__('Open %1', args.path) : trans.__('Open from URL'),
+      execute: async args => {
+        let url: string | undefined = (args?.url as string) ?? '';
```
fcollonval, line 721: In the label and caption, you are using `args.path` and here you are using `args.url` is there a reason to not use the same argument?

packages/filebrowser-extension/src/index.ts
```diff
@@ -691,6 +693,85 @@ export const launcherToolbarButton: JupyterFrontEndPlugin<void> = {
...
+        if (!url) {
+          url =
+            (
+              await InputDialog.getText({
+                label: trans.__('Url'),
+                placeholder: 'https://example.com/path/to/file',
+                title: trans.__('Open Url'),
```
fcollonval, line 728: We should probably use `URL` instead of `Url`

```suggestion
                label: trans.__('URL'),
                placeholder: 'https://example.com/path/to/file',
                title: trans.__('Open URL'),
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -691,6 +693,85 @@ export const launcherToolbarButton: JupyterFrontEndPlugin<void> = {
...
+        // upload the content of the file to the server
+        try {
+          const name = PathExt.basename(url);
+          const file = new File([blob], name, { type });
+          const model = await browser.model.upload(file);
+          return commands.execute('docmanager:open', {
+            path: model.path
+          });
+        } catch (error) {
+          trans._p('showErrorMessage', 'Upload Error'), error;
```
fcollonval, line 761: Missing a `showErrorMessage` call?

packages/filebrowser-extension/src/index.ts
```diff
@@ -86,6 +86,8 @@ namespace CommandIDs {
 
   export const openPath = 'filebrowser:open-path';
 
+  export const openUrl = 'filebrowser:open-url';
```
fcollonval, line 89: Probably better to stick with URL as in the standard API

```suggestion
  export const openURL = 'filebrowser:open-url';
```

fcollonval (changes requested):
Thanks @jtpio 

I mainly have comment on the variable naming.

packages/filebrowser-extension/src/index.ts
jtpio: Ah thanks for catching this. These are leftovers of bootstrapping the command from the "Open from Path" one.

Pushing the fix.

packages/filebrowser-extension/src/index.ts
jtpio: Looks like we use more often `Url` than `URL` in the code base when combined in a variable name? For example with `baseUrl` and `kernelIconUrl`.

No strong opinion on that one, although it reads a little bit more easily I think. 

> standard API

What does the standard API refer to?

fcollonval: Hey method like [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). But indeed if there are other use cases already in the code happy to keep it like this.

fcollonval (approved):
Thanks @jtpio

packages/filebrowser-extension/src/index.ts
jtpio: Good point. Maybe we could do a pass on the code base in a separate PR to unify all of this.

fcollonval:
@jtpio - this needs a rebase

jtpio:
Rebased and updated the reference snapshots.

fcollonval (approved):

yuvipanda:
Note that this doesn't respect the filename set in content-disposition header, so you might get a valid .ipynb file but with a nonsense filename. These open in just the editor view. 
image

jtpio:
Thanks @yuvipanda for reporting, this is now tracked in #11531

PR #11388, opened 2021-11-02
Set default ui font to `system-ui`

jasongrout:
## Code changes

Set the default ui and content font to `system-ui`, overriding the existing `-apple-system` and other system defaults. `system-ui` is a more modern cross-browser and cross-OS alternative to the existing default fonts listed before `Helvetica` in our font listing. See https://developer.mozilla.org/en-US/docs/Web/CSS/font-family or https://caniuse.com/?search=system-ui.

I also included a short Python script that can take a galata-report directory and update the snapshot images. I used this to do the large number of snapshot updates this PR requires.

I also added a new step to the galata workflow to generate and upload updated image snapshots if there was a test failure, which should make it easier to update images when needed. (@fcollonval - feel free to make that smarter!)

## User-facing changes

Update the default ui and content fonts to `system-ui`.

Some articles on using system fonts (like we are) in web pages:

- https://medium.design/system-shock-6b1dc6d6596f#.iw7sldtd6
- https://css-tricks.com/snippets/css/system-font-stack/

Both of those seem to be written before `system-ui` was an option.


fcollonval:
It seems the release test is hit by a bug of pip and Python 3.10 (I hit it in the gator project). It should be fixed by forcing upgrading to the latest pip version.

For ref the error emitted is:
```
ResolvePackageNotFound: 
  - python=3.1
```

with Python 3.10, old pip version may ends up creating a `lib/python3.1/site-packages` folder instead of `lib/python3.10/site-packages`.

jasongrout:
> It seems the release test is hit by a bug of pip and Python 3.10 (I hit it in the gator project). It should be fixed by forcing upgrading to the latest pip version.

I think we are using pip 21.3.1 at that stage? See https://github.com/jupyterlab/jupyterlab/runs/4076720108?check_suite_focus=true#step:10:7396

fcollonval:
Yes digging up it is a bug in conda itself not in pip: https://github.com/conda/conda/issues/10969

I opened #11390 that skips python 3.10.0.

jasongrout:
We are seeing differences in the font used for the UI. The new font in tab titles from the galata tests, for example, seems to be not quite as compressed, more spread out.

fcollonval:
To give a quick feeling on the change (before this PR is left, after is right):

image

jasongrout:
Another possibility is to change the ui-font, but leave the content font the way it currently is. That would get us slightly more readable menu items, etc., but without increasing the width of markdown text.

fcollonval:
> Another possibility is to change the ui-font, but leave the content font the way it currently is. That would get us slightly more readable menu items, etc., but without increasing the width of markdown text.

I think it makes sense to different both.

jasongrout:
After reading the above articles, actually, I think it's probably okay to go back to the original commit and make all ui and content fonts `system-ui`. For example, GitHub renders the vast majority of its text in the system font (San Francisco on macOS).

jasongrout:
Actually, I think there is probably no change on mac and windows - they should already be getting their own system fonts based on the precedence list. I think the major change here is for Linux, which now would be getting the system font whereas before it was getting Helvetica or Arial. Since our UI tests are run on Linux, that's why we're seeing the difference there.

jasongrout:
> The TOC test runs unfortunately in serial due to history effects. I would suggest running that particular test locally to update the screenshots. Otherwise you will do 3-4 more commits to get all screenshots...

Thanks. It won't do any good to run it locally since I'm on macOS, not Linux, so the screenshots would be different anyway :). Doing several more iterations is less effort than spinning up a Docker container.

However, is it easy to patch the test workflow to run all the way to completion and ignore errors, just to get the updated images?

fcollonval:
> However, is it easy to patch the test workflow to run all the way to completion and ignore errors, just to get the updated images?

I played around a bit - unfortunately when forcing it to regenerate the screenshots, they are not kept in the test results folder that is used to create the artifacts. Can I update your PR to add a conditional job producing the reference screenshots that will run if somebody comment on a PR like `please update the screenshots`?

galata/update_snapshots.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import hashlib
+from pathlib import Path
+import json
+import pprint
```
jasongrout, line 7:
```suggestion
```

jasongrout:
All tests pass! I think this is ready for review.

fcollonval (approved):
Thanks @jasongrout !

> I push a commit to solve the conflict on MANIFEST.in

jtpio:
> Can I update your PR to add a conditional job producing the reference screenshots that will run if somebody comment on a PR like `please update the screenshots`?

This would be really useful!

jasongrout:
The current pr generates the updated screenshots if there was a test failure.

fcollonval:
> The current pr generates the updated screenshots if there was a test failure.

Actually it depends on the test structure. For instance if a test is comparing multiple screenshots (that should be avoided) or if a series of test is set as serial because the result of a step depend on the previous one(s), only the first screenshot that is different will be produced. To be sure to regenerate all screenshots in one go, the tests should be run with the `--update-snapshots` option. As you mentioned, when running non-linux OS, it is not possible to do that locally. So that step (if failure occured) should probably be added.

jasongrout:
> To be sure to regenerate all screenshots in one go, the tests should be run with the --update-snapshots option. 

I think that is what my change is doing, if there is a test failure.

fcollonval:
Yo ja sorry I miss that change. We probably want to rename the artifact by prefixing it with `jupyterlab-` following #11405

.github/workflows/galata.yml
```diff
@@ -88,6 +88,19 @@ jobs:
...
+        run: |
+          cd galata
+          jlpm run test:update
+
+      - name: Upload updated snapshots
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: galata-updated-snapshots
+          path: galata/test/jupyterlab
```
fcollonval, line 102: Select snapshots from all projects and only snapshots & apply prefix `jupyterlab-`

```suggestion
          name: jupyterlab-galata-updated-snapshots
          path: galata/test/**/*-snapshots/*.*
```

fcollonval:
The new steps to generate the updated snapshots work as expected - thanks a lot @jasongrout for those :tada: 

But I cannot figure out what is the reason of the discrepancy.

jasongrout:
I rebased on master

jasongrout:
@fcollonval - tests are now passing, if you want to review and merge

fcollonval (approved):
Thanks a lot @jasongrout 

Let's merge this one. Do you think the update snapshots logic should be backported to 3.2.x?

fcollonval:
Backporting this one only to get the update snapshot logic.

fcollonval:

PR #11389, opened 2021-11-02
PoC Instantiate only visible editors

fcollonval:
## References

This is a performance related PR. Its goal is to experiment with instantiating a small number of editors. As it has been analyzed that codemirror editor refresh induces overhead.

For now try to instantiate only the editor on the active cell if in edit mode.

## User-facing changes

Goal is no visual impact

Target  is no backward incompatible changes. This probably will require to adopt an approach similar to Colab by instantiating all visible editors.


fcollonval (commented):
please run benchmark

fcollonval:
| Test file | large_code_notebook | large_md_notebook |
| --- | --- | --- |
| **open** | | |
| `chromium` | | |
| actual | 1868 <- [1914 - 1931 - 1947] -> 2038 | 2095 <- [2128 - 2143 - 2156] -> 2267 |
| expected | 4485 <- [6005 - 6249 - 6438] -> 7199 | 2895 <- [3124 - 3179 - 3427] -> 3671 |
| Mean relative change | -68.3% ± 0.6% | -34.0% ± 0.8% |

@afshin Good news this direction looks promising to improve by a good chunk the opening time of a notebook. The final numbers will be less impressive as the goal is to instantiate codemirror editors for all visible editors in the viewport so that we support cursor display for RTC and consistent look and feel coming from exotic codemirror extensions.

fcollonval:
This path should be discarded because extensions such LSP need to access all codemirror editors; although they don't have to be in the DOM.

I'll focus on #11574 preserving the instantiation of codemirror editors.

---

> But I guess even if notebook never gets fully rendered, there is a time at which all CodeMirror Editor objects are fully instantiated? That's the only thing we would want to know for LSP ("can we start interacting with the editors, even if they are not attached to the DOM").

@krassowski commented on [gitter](https://gitter.im/jupyterlab/jupyterlab?at=61a80704a9c8eb44c40b3377)

krassowski:
@fcollonval I don't think that the current LSP requirement is an ultimate blocker for this one (It is possible to refactor LSP to work with a mix of instantiated and non-instantiated editors). Of note the current implementation of spellchecker would need to be reworked for this one too (but we want to do that either way to support spellchecking code cells).

So while I agree that #11574 might be an easier way forward, I don't think we have to give up on this one just yet :)

fcollonval:
> I don't think we have to give up on this one just yet :)

:smile: 

Interested people are welcomed to take over this one then. I focus on #11574.

PR #11390, opened 2021-11-02
Fix release_test

fcollonval:
## References

There is a problem with conda and Python 3.10. When resolving the environment, conda is looking for `python=3.1`. 

Xref: https://github.com/conda/conda/issues/10969

## Code changes

Don't use python 3.10

## User-facing changes

N/A

N/A


jasongrout:
Thanks!

jtpio:
Looks like we can backport to 3.2.x since the issue is also happening on that branch: https://github.com/jupyterlab/jupyterlab/runs/4078192664?check_suite_focus=true

jtpio:

Issue #10706, opened 2021-07-27
Writing Tests for JupyterLab Extensions


subhamagrawal7:
1. Can this library [Galata](https://github.com/jupyterlab/galata) be used for testing lab extensions and in general be used outside the core team??
2. If the tests can be written in Galata, it would be really helpful if you could provide some documentation for the same as the repository does not seem to have them (apart from the Readme.md which doesn't give all the functions that the API has).
3. If the tests cannot be written in Galata, then, could you give some recommendations for writing the tests in [Cypress](https://github.com/cypress-io/cypress) to ensure they are robust (e.g. don't break on classname changes)

fcollonval:
Hey @subhamagrawal7 


> Can this library Galata be used for testing lab extensions and in general be used outside the core team??

Yes you can. Jupyter widgets are tested using it for example: https://github.com/jupyter-widgets/ipywidgets/tree/master/ui-tests

> If the tests can be written in Galata, it would be really helpful if you could provide some documentation for the same as the repository does not seem to have them (apart from the Readme.md which doesn't give all the functions that the API has).

Having use it, lately the experience is not as great as it could be. There are some plans to build it a bit differently to exploit more features from Playwright.

> If the tests cannot be written in Galata, then, could you give some recommendations for writing the tests in Cypress to ensure they are robust (e.g. don't break on classname changes)

So I would recommend using [Playwright](https://playwright.dev) rather than Cypress. For having use Cypress, Playwright is really above especially in term of default selectors determine by the test code generator tool.

You can see demonstration of integration test with Playwright on the official examples repository: https://github.com/jupyterlab/extension-examples  
Every example has a `ui-tests` folder that contains the integration tests. The readme in that folder describes the main commands to use to create, execute and debug tests.

bollwyvl:
For a lot of extensions I work on, i prefer the workflow offered by [robotframework](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html) and [seleniumlibrary](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html).

Some of the lab/notebook/(old) nteract patterns are wrapped up in the `pip`/`conda-forge` installable [robotframework-jupyterlibrary](https://robotframework-jupyterlibrary.readthedocs.io) (disclaimer: author). 

[...]

PR #11391, opened 2021-11-02
Improve documentation on galata setup

fcollonval:
## References

Improve documentation on galata setup: #10706 and Gitter question.

## Code changes

Rename server configuration file for ui tests to match the name used in the documentation.

## User-facing changes

N/A

N/A


jtpio (approved):
Thanks!

fcollonval:
@krassowski I saw in #11379 you changed some of the documentation. It would make sense to bring it in this repository, what do you think?

> I don't get the rename the `--config` option in #11379. It is an alias that should work:

```
--config=<Unicode>
    Full path of a config file.
    Default: ''
    Equivalent to: [--JupyterApp.config_file]
```

krassowski:
Thanks! Please ignore my changes, I will undo them. I was tired and it looks like the issue was elsewhere. Sorry about it!

Would it be a good idea to add:

```python
ServerApp.port_retries = 0
```

so that we don't accidentally run tests on a different instance which happens to run on the default port? (This might have been my problem).

Possibly another improvement would be to change the default port to reduce a chance of a collision, but this might need more discussion.

fcollonval:
> Would it be a good idea to add:
> 
> ```python
> ServerApp.port_retries = 0
> ```

Thanks this is a great suggestion

> Possibly another improvement would be to change the default port to reduce a chance of a collision, but this might need more discussion.

I fear the ship have sailed on this one as this means changing the default value in galata fixture. So that an upgrade of current user may induce unexpected broken configuration.

fcollonval:

PR #11398, opened 2021-11-03
fix #11377 & bump Yjs dependencies & fix modeldb overwriting yjs content

### LINKED ISSUE #11377: Completer closes when typing a letter after dot in Python

dmonad:
## References

This PR fixes #11377. While debugging I found other unintended effects which I fixed in this PR as well. Furthermore I bumped Yjs dependencies because they include some fixes that might be relevant to JupyterLab.

## Code changes

As described in #11377 the cursor event is fired before content is propagated to modeldb. Events in codemirror are fired in this order when typing a character: `beforeChange`, `cursorActivity`, `change`, `changes`. y-codemirror listens to the `changes` event to reflect the changes in the Yjs data model. This means that the cursor event is triggered before the changes are actually applied. In the case of #11377 this lead to problems.

I introduced a new cursor event in y-codemirror that fixes the order of events to `beforeChange`, `change`, `changes`, `cursorActivity`, which IMHO is how it should be.

---

While debugging this issue, I found that 1. there are several instances of the CodeEditorModel and CellModels. It seems that these models are never cleaned up - this is a potential memory problem. 2. modeldb completely replaces the content of the `shared-model` - another potential performance problem. This issue is happening because there are several instances of the CodeEditorModel and we can't use the `mutex` approach to filter events. I removed the `_mutex` property and made it a global variable which should fix the issue. We can revert to the old approach once we fixed 1.

## User-facing changes

Only bugfixes.

None.

Mentioning @hbcarlos @krassowski @phil-zxx


krassowski:
Hi @dmonad could you please cherry pick a40dfc25a84ed5180b15847e7fb14db84546f650 and 01e8b73195dc16b9ec1f88c5607331c99fed280a from #11379 so we get the visual test cases in, please? Feel free to squash/create a new commit if easier.

dmonad:
Oh absolutely, thanks!

fcollonval:
This looks good to me but the CI failure seems legit - could it come from `yjs` not being update in `yarn.lock`?

dmonad:
Oh right, thanks for the hint @fcollonval ! This should be fixed now.

krassowski (approved):
Thank you!

fcollonval (approved):

jtpio:
Thanks all!

krassowski:
Is it fine to backport this one to 3.2.x?

jtpio:
I think so. Looks like this doesn't have any breaking change and is mostly a bug fix.

jtpio:

PR #11405, opened 2021-11-05
Add `jupyterlab` prefix to the Galata artifacts

jtpio:
## References

A small naming change to add the `jupyterlab-` prefix to the Galata test assets.

This is useful when working on multiple projects that have a Galata setup, and downloading the test report locally.

## Code changes

CI workflow update.

## User-facing changes

None

None


fcollonval (approved):
Good idea, thanks @jtpio

PR #11413, opened 2021-11-07
perf: scroll active toc item into view

### LINKED ISSUE #11263: Table of contents should scroll to bring relevant heading into the view
### Problem

Navigating larger documents can be difficult. Table of contents nicely shows the current heading, but it does not automatically scroll it into the view.

Tagging this as a good first issue, but it is likely a good first issue for someone who already worked with React rather than for a complete novice.

toc-does-not-scroll

### Proposed Solution

`scrollIntoView()` should be called on the heading node (these have `jp-tocItem` class), or its contents (`toc-entry-holder`) in the ToC when the state of the heading changes to active. Bottom-up, the state of the cell is currently compared with the heading in the renderer:

https://github.com/jupyterlab/jupyterlab/blob/d48e0c04efb786561137fb20773fc15788507f0a/packages/toc/src/generators/notebook/render.tsx#L85-L101

we will either want to store a reference to the heading node at this level, so that we can call `scrollIntoView()` on it after rendering is completed, or even perform it at the render level (which is possible with react, but feels like a hack to me). As a reference I would recommend checking some of the solutions proposed in: https://stackoverflow.com/questions/43441856/how-to-scroll-to-an-element

Importantly, we only want to scroll after user navigated to a different cell; the relevant signal gets received by ToC already in:

https://github.com/jupyterlab/jupyterlab/blob/d48e0c04efb786561137fb20773fc15788507f0a/packages/toc/src/generators/notebook/index.ts#L73-L77

It would need to either invoke the the scroll action, or just pass information on the change lower down so it is available in a post-render action.

Further, we would want to use [`ElementExt.scrollIntoViewIfNeeded`](https://lumino.readthedocs.io/en/latest/api/domutils/modules/elementext.html#scrollintoviewifneeded) polyfil from lumino (it is already used for scrolling in notebook).

Similar, but distinct problems were recently solved:
- https://github.com/jupyterlab/jupyterlab/issues/8252
- https://github.com/jupyterlab/jupyterlab/issues/9740

### Additional context

- LibreOffice Writer (and I believe other office suites as well) scrolls to the relevant heading in the outline when user navigates the document.

skyetim:
## References

Fix #11263 .

## Code changes

Used a slightly different approach from proposed solution in the issue. When widget is called to update, it will find active cell and scroll into view if needed. 

## User-facing changes

Before:

before

After: 

after

None.

### DIFF
--- a/packages/toc/package.json
+++ b/packages/toc/package.json
@@ -53,7 +53,6 @@
     "@jupyterlab/translation": "^3.3.0-alpha.14",
     "@jupyterlab/ui-components": "^4.0.0-alpha.14",
     "@lumino/coreutils": "^1.10.0",
-    "@lumino/domutils": "^1.7.0",
     "@lumino/messaging": "^1.9.0",
     "@lumino/signaling": "^1.9.0",
     "@lumino/widgets": "^1.28.0",
--- a/packages/toc/package.json
+++ b/packages/toc/package.json
@@ -53,6 +53,7 @@
     "@jupyterlab/translation": "^3.3.0-alpha.14",
     "@jupyterlab/ui-components": "^4.0.0-alpha.14",
     "@lumino/coreutils": "^1.10.0",
+    "@lumino/domutils": "^1.7.0",
     "@lumino/messaging": "^1.9.0",
     "@lumino/signaling": "^1.9.0",
     "@lumino/widgets": "^1.28.0",
--- a/packages/toc/src/generators/notebook/index.ts
+++ b/packages/toc/src/generators/notebook/index.ts
@@ -104,7 +104,7 @@ function createNotebookGenerator(
    * @returns rendered item
    */
   function renderItem(item: INotebookHeading, toc: INotebookHeading[] = []) {
-    return render(options, tracker, item, toc);
+    return render(options, tracker, widget, item, toc);
   }
 
   /**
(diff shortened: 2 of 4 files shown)


fcollonval (changes requested):
Thanks for working on this @skyetim 

As this is related to the notebook toc, we should define a logic that is enclosed in the notebook toc generator.

To do so, you can take inspiration of https://www.carlrippon.com/scrolling-a-react-element-into-view/

The idea here is to modify the `render` function:

https://github.com/jupyterlab/jupyterlab/blob/114159269a76c4a8d4c1e725e882d7c8a7791869/packages/toc/src/generators/notebook/render.tsx#L23

And create a new element `NotebookHeading` instead of the simple `div` that will have the feature of scrolling into view; something like:

```ts
function NotebookHeading(props): JSX.Element {
  const itemRef = React.useRef<HTMLDivElement>(null);
  const { isActive } = props;
    React.useEffect(() => {    if (isActive && itemRef.current) {      itemRef.current.scrollIntoView();    }  }, [isActive]);
    return <div ref={itemRef} className={[props.className, (isActive ? 'toc-active-cell' : '')].join(' ')} >{props.children}</div>
}
```

skyetim:
Thanks for the suggestion. `scrollIntoView` will make the scroll quite weird as it will scroll the element even though it's already in view (as the following gif). I replaced it with `ElementExt.scrollIntoViewIfNeeded`. 

scrollintoview

fcollonval (changes requested):
Thanks for quickly applying the changes. I have a last enhancement suggestion. And it will be good to go.

fcollonval (approved):

fcollonval:
Thanks @skyetim this looks good to me

PR #11415, opened 2021-11-08
Better editor initialization

fcollonval:
## References

Improve cell editors initialization by setting mimetype at init.


fcollonval (commented):
please run benchmark

PR #11427, opened 2021-11-10
Handle relative paths to `themePath` and `schemaDir`

jtpio:
## References

Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3307

The `@jupyter-widgets/jupyterlab-manager` extension uses `./schema` as the `schemaDir` here:

https://github.com/jupyter-widgets/ipywidgets/blob/6dff84d5fec69d95c860b3039024fc688cf0c413/python/jupyterlab_widgets/package.json#L95

Which makes the string comparison fail here when the extension is installed with `jupyter labextension install`:

https://github.com/jupyterlab/jupyterlab/blob/2b8a2cedcdb7cd4d3aac556692b146e8712a0e8f/jupyterlab/commands.py#L2078

## Code changes

Normalize the path to `schemaDir` and `themePath` to drop the leading `./` if any.

## User-facing changes

Should fix installing some extensions with `jupyter labextension install`.

None

### DIFF
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -2071,10 +2071,10 @@ def _validate_extension(data):
     if mime_extension and mime_extension not in files:
         messages.append('Missing mimeExtension module "%s"' % mime_extension)
 
-    if themePath and not any(f.startswith(themePath) for f in files):
+    if themePath and not any(f.startswith(str(Path(themePath))) for f in files):
         messages.append('themePath is empty: "%s"' % themePath)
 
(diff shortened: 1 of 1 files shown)


jasongrout (approved):
This actually goes further and drops any `.` components in the path, which I think is good. Thanks!

When we support Python 3.9 or above, it's probably better to use Path.is_relative_to (https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to), as that handles other path idiosyncrasies, but I think this current PR is great for now.

jtpio:

PR #11431, opened 2021-11-10
Galata: add `acceptDownloads` to the browser context

jtpio:
## References

This should help test downloading content from the UI such as notebooks, and check they are correctly formatted.

The Playwright docs mention this flag must be explicitly set to `true`:

image

https://playwright.dev/docs/api/class-download/

## Code changes

Add `acceptDownloads` to the default galata config.

## User-facing changes

None

None

Can be backported to 3.2.x

### DIFF
--- a/galata/src/playwright-config.ts
+++ b/galata/src/playwright-config.ts
@@ -17,6 +17,9 @@ module.exports = {
     viewport: { width: 1024, height: 768 },
 
     // Artifacts
-    video: 'retain-on-failure'
+    video: 'retain-on-failure',
+
+    // Allow downloading files
+    acceptDownloads: true
   }
 } as PlaywrightTestConfig;


galata/src/galata.ts
```diff
@@ -207,7 +207,7 @@ export namespace galata {
     tmpPath: string,
     waitForApplication: (page: Page, helpers: IJupyterLabPage) => Promise<void>
   ): Promise<IJupyterLabPageFixture> {
-    const context = await browser.newContext();
+    const context = await browser.newContext({ acceptDownloads: true });
```
jtpio, line 210: Actually it looks like we might not be using this `newPage` helper at the moment.

fcollonval: Indeed it is not.

i think the better way is to use the test options: https://playwright.dev/docs/api/class-testoptions#test-options-accept-downloads
And therefore updating https://github.com/jupyterlab/jupyterlab/blob/master/galata/src/playwright-config.ts

jtpio: Ah nice, thanks for the hint!

jtpio:
Updated to pass the flag as part of the `playwright-config.ts` options.

However since it can easily be specified by downstream projects, it would also be fine to not add it here for now. Or at least until we start testing downloads in the jupyterlab tests.

jtpio:
> Or at least until we start testing downloads in the jupyterlab tests.

Closing for now until (or if) we start testing features related to downloads.

Issue #11412, opened 2021-11-07
In JupyterLab 3.2.2 can't use Arrows shortcuts


sntgluca:
## Description
Hi all,
I apologise in advance if this was already under your radar. I searched for a similar issue but can't find any. 

In JupyterLab 3.2.2, I can't make use of Arrows to execute cell commands. 
For instance, moving between cells using ArrowUp or ArrowDown doesn't work, and I need to use the J/K keys combination instead.

The ArrowUp/ArrowDown events scroll the page instead.

## Reproduce
Open a notebook, and use ArrowUp or ArrowDown to move between existing cells, or Shift-ArrowUp/Shift-ArrowDown to select multiple cells.

The equivalent with J/K keys would work.

## Expected behavior
I expect to be able to use/modify Arrows shortcuts to execute commands such as selecting cells and activating cells.
 
In JupyterLab 3.2.1 Arrows are working well.

## Context

- Operating System and version: Tested on CentOS7 and MacOS Big Sur
- Browser and version: Chrome v 95 and Microsoft Edge Version 95 on Mac OS
- JupyterLab version: 3.2.2

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

krassowski:
Thank you for reporting! It is also tracked in https://github.com/jupyterlab/lumino/pull/256; the root issue is https://github.com/jupyterlab/lumino/pull/256 and it was already fixed in lumino and the fix should be included in the next JupyterLab release :)

sntgluca:
Lovely, thanks!

sntgluca:
Duplicate of  https://github.com/jupyterlab/jupyterlab-desktop/issues/320

jasongrout:
Have we upgraded the lumino version here yet? Master dependencies still shows lumino commands 1.18, but the fix is in 1.19.

michaelaye:
Highly annoying. Every navigation currently requires an extra Enter to go back to edit mode, just to reach other cells. :/ Can't wait for that bugfix release!

jasongrout:
[...]

PR #11433, opened 2021-11-10
Use the root yarn.lock in staging when making a release.

### LINKED ISSUE #11429: Released JupyterLab uses different dependencies than the master yarn.lock specifies
Recently there was a bug in `@lumino/commands` version 1.18.0 which prevented arrow shortcut keys from working (see https://github.com/jupyterlab/lumino/issues/255, https://github.com/jupyterlab/lumino/issues/151, and https://github.com/jupyterlab/lumino/pull/258; jlab issue at https://github.com/jupyterlab/jupyterlab/issues/11412).

It turns out that `@lumino/commands` 1.15.0 is what yarn.lock is locked to (this is the root yarn.lock file at the 3.2.2 release commit):

https://github.com/jupyterlab/jupyterlab/blob/4852e50b431a491b154a69acbf95b948e69e4d89/yarn.lock#L2352-L2353

However, something in the release process actually released JLab 3.2.2 using `@lumino/commands` 1.18.0 (this is the yarn.lock file in our 3.2.2 release staging directory at the same commit):

https://github.com/jupyterlab/jupyterlab/blob/4852e50b431a491b154a69acbf95b948e69e4d89/jupyterlab/staging/yarn.lock#L1731-L1732

The end result is that we are releasing something that we are not seeing or testing in our dev branches. That concerns me, in that we can pick up regressions in dependencies that we won't catch in our normal dev workflow, like we did for 3.2.2.

jasongrout:
## References

Fixes #11429

## Code changes

Essentially, since we were creating a new yarn.lock when making a release, we were possibly upgrading dependencies in a release that we were not using in our dev branches. This meant we could introduce changes or regressions right when making a release that we would not see in our dev workflow. This change locks the release to the versions of packages we had been using in the dev process.

I think this change is risky enough (given that is changing how release dependencies have been done since pre-1.0) that we should not backport it.

### DIFF
--- a/buildutils/src/update-core-mode.ts
+++ b/buildutils/src/update-core-mode.ts
@@ -57,7 +57,11 @@ const notice =
   }
 });
 
-// Create a new yarn.lock file to ensure it is correct.
+// Copy the root yarn.lock, then update and deduplicate to prune it.
+fs.copySync(
+  path.join('.', 'yarn.lock'),
+  path.join('.', 'jupyterlab', 'staging', 'yarn.lock')
+);
 utils.run('jlpm', { cwd: staging });
 try {
   utils.run('jlpm yarn-deduplicate -s fewer --fail', { cwd: staging });


blink1073 (approved):
Thank you!  I agree we should not backport this change.

jtpio:
Thanks!

For reference the dist files (built by the releaser) are uploaded as GitHub artifacts as part of the check release workflow, and can be inspected here: https://github.com/jupyterlab/jupyterlab/actions/runs/1445095685

image

krassowski:
As previously suggested in https://github.com/jupyterlab/jupyterlab/issues/11429#issuecomment-974696314, and based on multiple breakages in JupyterLab due to dependency changes:
- https://github.com/jupyterlab/jupyterlab/issues/11473: klona update affecting sanitize-html breaking CSS styling in notebooks
- https://github.com/jupyter/notebook/issues/6270: the same as above in Notebook (since v6 includes some JupyterLab code now)
- https://github.com/jupyterlab/jupyterlab/issues/11412: the keyboard shortcuts issue in lumino
- https://github.com/jupyterlab/jupyterlab/issues/11856: recent lumino updates
- https://github.com/jupyterlab/jupyterlab/issues/11854: more recent recent lumino updates

I believe we should backport this to 3.3.x and 3.2.x and make a new hotfix release on 3.2.x with pinned lumino to fix the new issues.

jtpio:
Thanks @krassowski for bringing this up.

> I believe we should backport this to 3.3.x and 3.2.x and make a new hotfix release on 3.2.x with pinned lumino to fix the new issues.

This also sounds like the benefits (fixing the breakage) would outweigh the risk.

cc @jasongrout @blink1073 if you have more input about this. Otherwise we can move forward with this, merge https://github.com/jupyterlab/jupyterlab/pull/11858 and make a new `3.2.8` release right after.

blink1073:
I'm in favor of backporting given new evidence.

jtpio:
OK then let's merge https://github.com/jupyterlab/jupyterlab/pull/11858 and I'll start a new release right after.

jtpio:
For reference `3.2.8` is out with this change: https://github.com/jupyterlab/jupyterlab/releases/tag/v3.2.8

jasongrout:
> I'm in favor of backporting given new evidence.

Same here.

PR #11440, opened 2021-11-11
Update the release documentation to recommend using the Jupyter Releaser

jtpio:
## References

Now that we are using the Jupyter Release for making releases.

## Code changes

Documentation.

## User-facing changes

None

None

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -27,7 +27,7 @@ The workflow takes a couple of input parameters, for example:
 | Version Spec | next                  |
 | Since        | v4.0.0a15             |
 
-The version spec follows the specification documented bellow in the [Bump Version](#bump-version) section.
+The version spec follows the specification documented below in the [Bump Version](#bump-version) section.
 
 We can use `next` when making a `patch` release or a `build` pre-release.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,10 +2,65 @@
 
 This document guides a contributor through creating a release of JupyterLab.
 
+## Automated Releases with the Jupyter Releaser
+
+The recommended way to make a release is to use [`jupyter_releaser`](https://github.com/jupyter-server/jupyter_releaser#checklist-for-adoption).
+
+### Workflow
+
+#### Draft Changelog
+
+The first step is to generate a new changelog entry for the upcoming release.
+
+We use the "Draft Changelog" workflow as documented here: https://jupyter-releaser.readthedocs.io/en/latest/get_started/making_first_release.html#draft-changelog
+
+The workflow takes a couple of input parameters, for example:
+
+| Input  | Value |
+| ------------- | ------------- |
+| Target | jupyterlab/jupyterlab  |
+| Branch  | master  |
+| Version Spec | next |
+| Since | v4.0.0a15 |
+
+The version spec follows the specification documented bellow in the [Bump Version](#bump-version) section.
+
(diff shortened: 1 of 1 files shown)


jtpio:
cc @jasongrout if you want to have a look and check the information is accurate.

For now this change includes a couple of details and screenshots for a smoother transition to the releaser.

Later on we can also drop this and just refer to the Jupyter Releaser documentation directly: https://jupyter-releaser.readthedocs.io/en/latest/get_started/making_first_release.html#making-your-first-release

RELEASE.md
```diff
@@ -2,18 +2,78 @@
...
+The workflow takes a couple of input parameters, for example:
+
+| Input        | Value                 |
+| ------------ | --------------------- |
+| Target       | jupyterlab/jupyterlab |
+| Branch       | master                |
+| Version Spec | next                  |
+| Since        | v4.0.0a15             |
+
+The version spec follows the specification documented bellow in the [Bump Version](#bump-version) section.
```
jasongrout, line 30:
```suggestion
The version spec follows the specification documented below in the [Bump Version](#bump-version) section.
```

RELEASE.md
```diff
@@ -2,18 +2,78 @@
...
+
+We can use `next` when making a `patch` release or a `build` pre-release.
+
+Here is an example screenshot of what it should look like:
+
+![draft-changelog-example](https://user-images.githubusercontent.com/591645/141277239-571c2e6a-e8d5-4af9-aee4-dd52a7dae8ba.png)
```
jasongrout, line 36: Should this image be included in the repo so the documentation is stand-alone, or perhaps the fields should just be listed out in text instead of a screenshot that will get out of date?

I think instead of a screenshot, a listing of each field and what text is in it is more sustainable as instructions.

RELEASE.md
```diff
@@ -2,18 +2,78 @@
...
+
+We then wait for:
+
+1. the PR to be created on the repo. Example: https://github.com/jupyterlab/jupyterlab/pull/11422
+2. Tests to pass
+3. Merge the changelog PR
+
+### Full Release
```
jasongrout, line 44: Can you add a note about making sure to set up up NPM and PYPI tokens before running the release workflows?

jasongrout (commented):
Thanks! A few comments inline...

RELEASE.md
jtpio: As mentioned above the releaser docs already have screenshots anyway: https://jupyter-releaser.readthedocs.io/en/latest/get_started/making_first_release.html#making-your-first-release

So we can stick to plain text here.

I find screenshots really useful because we can immediately get the idea of what to do. Agree they can easily get out of date though.

RELEASE.md
jtpio: Done in https://github.com/jupyterlab/jupyterlab/pull/11440/commits/bfae1222f03369f9eb5c87ee0479daec3c388db0

blink1073:
> Later on we can also drop this and just refer to the Jupyter Releaser documentation directly

We've been keeping the manual instructions around in case something is up with releaser and you still need to make a release.

blink1073:
Nice, thank you!

jtpio:
> We've been keeping the manual instructions around in case something is up with releaser and you still need to make a release.

Also if for some reason there is an issue with GitHub Actions or GitHub is down, we can still use the `jupyter_releaser` CLI manually.

Issue #11355, opened 2021-10-27
Add RBAC support to JupyterLab


hbcarlos:
### Problem
If we want to restrict actions to collaborators, we need to set different roles.
I'm already working on a [PR](https://github.com/jupyterlab/jupyterlab/pull/11180) to add a new package that represents the active user by introducing a new interface `IUser`, a new token `ICurrentUser` and an extension to implement a basic model `User` that stores the user info in the browser. The interface `IUser` has the attribute role defining four new roles ADMIN, READ, WRITE, RUN.

For now, In #11180, I just added the attribute, but I'm not using it in any other place (I think we should open another PR to restrict the actions of a collaborator).

The main problem is that Jupiter Server doesn't have RBAC. For that reason, the new extension creates anonymous users with the role ADMIN. Every user can do any action.

### Proposed Solution
Make core extensions depend on the new package user and limit the different actions by roles.

krassowski:
How does this proposal and PR fit together with the RBAC support implemented by JupyterHub team in 2.0 release (https://github.com/jupyterhub/jupyterhub/issues/2245 and https://github.com/jupyterhub/jupyterhub/pull/3438) and with the ongoing work in jupyter server in: https://github.com/jupyter-server/jupyter_server/pull/165 and in https://github.com/jupyter-server/jupyter_server/pull/540? Sorry if this is a trivial question, I would just like to understand the landscape a little bit more (and cross-reference the issues so we can navigate easily between projects). Is there a coordination activity/document so that someone who is not actively working in this domain could catch up? Thanks!

hbcarlos:
Hey @krassowski! It is not a trivial question. It is the right one.
The idea is to follow what JupyterHub and Jupyter Server do, to have the exact implementation across projects, so everyone benefits from it.
For now, there is no coordination. I think we can start defining the roles for the user and the scopes for the different actions based on the work happening in [here](https://github.com/jupyter-server/jupyter_server/pull/165) and JupyterHub's documentation [here](https://jupyterhub.readthedocs.io/en/rbac/rbac).

fcollonval:
Although this is tagged for Real-Time Collaboration, I did not move it to `jupyter_collaboration` repository as it is likely to require work in core (as the identity and authorization are part of Jupyter Server).

krassowski:
With the past changes in jupyter-server and JupyterHub is there any more work that needs to be done on JupyterLab side?

krassowski:
[...]

PR #10411, opened 2021-06-14
RTC: Anonymous user names based on the moons of Jupyter

dmonad:
## References

Currently, the collaborative-editing feature doesn't show any user-names. In the previous Jupyter Notebook meeting somebody requested adding anonymous user-names and random colors to distinguish the users somehow. This PR assigns each user an anonymous name based on the moons of Jupyter. A user will be rendered as, for example, `Anonymous Metis` or `Anonymous Amalthea`. I also added some usercolors that I use in my projects. You can keep using the colors or change them in a future PR.

Screenshot from 2021-06-14 17-15-37

@williamstein Requested that it should be possible to set your own user-name and user-color. This is now also possible by adding `?username=MyName&usercolor=007007` Where `usercolor` is a six-digit hexadecimal string without the hash.

## Code changes

The user-names are added by the docprovider. Once the `docprovider` supports authentication, it can use real usernames instead.

## User-facing changes

None.

## Backwards-incompatible changes

None.

PR #9296, opened 2020-11-06
Move the Single Document Switch to the status bar

jtpio:
## References

#9117 and #9105

This is more of an experimental PR to have an idea of what it would look like having the Single Document Switch in the status bar instead of the top area. And also be easily tested on Binder.

Since the status bar already has icons, some of the ideas mentioned in #9117 about using icons as an alternative to the switch might be relevant.

## Code changes

Move the single document switch to the status bar, still in its own plugin.

## User-facing changes

To play with the 2 options on Binder:

- Using JupyterLab 3.0.0rc7 with the SDM switch in the top area: https://mybinder.org/v2/gist/jtpio/8d392846c2f3ee6fab38cdfab5a364ec/master?urlpath=/lab
- Using the PR with the Binder dev mode, with the SDM switch in the status bar: https://mybinder.org/v2/gh/jtpio/jupyterlab/sdm-switch?urlpath=lab-dev

### Before (switch upper right)

UyvprsSO1j

### After (switch in status bar)

sdm-switch-statusbar

## Backwards-incompatible changes

None

PR #11443, opened 2021-11-11
Add a user package to represent the current connected user

hbcarlos:
This PR was started in #11180, I decided to close the other one to separate every  feature in a different PR.

This PR adds two new packages, "@jupyterlab/user", and "@jupyterlab/user-extension". These packages introduce two new plugins:

#### ICurrentUser:
Provides a representation of the connected user. The idea is that some extensions will require the ICurrentUser to read the information (docprovider to use this identity in RTC, or an extension to add comments in a notebook, etc.). At the moment, all the users are anonymous since we don't have any identity provider, but the idea is that other extensions can swap the default implementation of the current user to fetch the information from a third-party identity provider.
This interface includes the property role. For now, this property is useless since any other package supports RBAC to limit the actions that users can do. There is already an [issue](https://github.com/jupyterlab/jupyterlab/issues/11355) open to discuss what roles should we have.

#### IUserMenu:
Provides a new menu added to the JupyterLab toolbar with the current user's name and icon. The idea behind this widget is to provides a lumino menu where extensions can add different buttons to actions related to the user (Login/Logout buttons, settings, etc.).

## References
solves #10965

## Code changes
Adds two new package, "@jupyterlab/user", and "@jupyterlab/user-extension".

## User-facing changes
Now users will see his name and icon at the right side of the toolbar.

image


jtpio:
Thanks!

> Adds two new package, "@jupyterlab/user", and "@jupyterlab/user-extension".

I think some CI checks are expected to fail because these packages are not published to npm yet.

hbcarlos:
Thanks Jeremy!! I updated the images of the UI test to include the new menu.

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * A namespace for command IDs.
+ */
+export namespace CommandIDs {
+  export const settings = '@jupyterlab/user-extension:settings:open';
+}
+
+/**
+ *
```
fcollonval, line 30: Could you add a small docstring?

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  autoStart: true,
+  requires: [IStateDB],
+  provides: ICurrentUser,
+  activate: (app: JupyterFrontEnd, state: IStateDB): User => {
+    return new User(state);
+  }
+};
+
+/**
+ *
```
fcollonval, line 43: Same here

packages/user/src/components.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { User } from './model';
+import { IUser } from './tokens';
+import { getInitials } from './utils';
+
+/**
+ * React component to render the user icon.
+ *
+ * @param user User information as IUser or JSON object.
+ */
+export const getUserIcon = (user: IUser | IUser.User) => {
```
fcollonval, line 18: Prefer that structured otherwise the react component will be anonymous.

```suggestion
export function getUserIcon(user: IUser | IUser.User): JSX.Element {
```

packages/user/src/components.tsx
```diff
@@ -0,0 +1,89 @@
...
+  handlerColor: (event: any) => void;
+};
+
+/**
+ * React component to render a custom color picker.
+ *
+ * @param user User information as JSON object.
+ * @param handlerColor callback to call when color changes.
+ */
+export const ColorPickerComponent: React.FC<Props> = props => {
```
fcollonval, line 62:
```suggestion
export function ColorPickerComponent(props: Props): JSX.Element {
```

packages/user/src/components.tsx
```diff
@@ -0,0 +1,89 @@
...
+          tag="span"
+          width="25px"
+          height="25px"
+        />
+      </div>
+    );
+  }
+};
+
+type Props = {
```
fcollonval, line 51: Please add some doc and export it as the component using it is exported

packages/user/src/tokens.ts
```diff
@@ -0,0 +1,352 @@
...
+ */
+export const ICurrentUser = new Token<IUser>('@jupyterlab/user:ICurrentUser');
+
+/**
+ * The user menu token.
+ *
+ * NOTE: Requirer this token in your extension to access the user menu
+ * (top-right menu in JupyterLab's interface).
+ */
+export const IUserMenu = new Token<Menu>('@jupyterlab/user:IUserMenu');
```
fcollonval, line 27: You should create an interface IUserMenu that describe the API of the token and not using a default type directly (expose only what is needed).

packages/user/src/tokens.ts
```diff
@@ -0,0 +1,352 @@
...
+ * NOTE: The left side bar panel. Useful to add user related widgets.
+ */
+export const IUserPanel = new Token<AccordionPanel>(
+  '@jupyterlab/user:IUserPanel'
+);
+
+/**
+ * An interface describing an user.
+ */
+export interface IUser {
```
fcollonval, line 41: Do we really want to list attributes like `gender`, `birthDate`, `honorificPrefix`,... ?
I would suggest extending on ReadOnlyJSONObject and define only what is required (name, username,...) or used by the code (cursor position).

Moreover it seems duplicated of type `IUser.User`. So you should probably linked the two: `interface IUser extends IUser.User`.

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+ * A namespace for command IDs.
+ */
+export namespace CommandIDs {
+  export const settings = '@jupyterlab/user-extension:settings:open';
+}
+
+/**
+ *
+ */
+const userPlugin: JupyterFrontEndPlugin<User> = {
```
fcollonval, line 32: Plugin template should match its contract:

```suggestion
const userPlugin: JupyterFrontEndPlugin<ICurrentUser> = {
```

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  id: '@jupyterlab/user-extension:user',
+  autoStart: true,
+  requires: [IStateDB],
+  provides: ICurrentUser,
+  activate: (app: JupyterFrontEnd, state: IStateDB): User => {
```
fcollonval, line 37: Same here:

```suggestion
  activate: (app: JupyterFrontEnd, state: IStateDB): ICurrentUser => {
```

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    return new User(state);
+  }
+};
+
+/**
+ *
+ */
+const userMenuPlugin: JupyterFrontEndPlugin<Menu> = {
```
fcollonval, line 45: You need to define the minimal interface needed by downstream plugin - avoid providing the full menu API if possible (see later comment).

```suggestion
const userMenuPlugin: JupyterFrontEndPlugin<IUserMenu> = {
```

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  id: '@jupyterlab/user-extension:userMenu',
+  autoStart: true,
+  requires: [ICurrentUser],
+  provides: IUserMenu,
+  activate: (app: JupyterFrontEnd, user: User): Menu => {
+    const { shell, commands } = app;
+
+    const spacer = new Widget();
+    spacer.id = 'jp-topbar-spacer';
+    shell.add(spacer, 'top', { rank: 1000 });
```
fcollonval, line 55: This is gonna be a problem for the two following reasons:

- Not enough rank between 1000 and 1002: if user wants to place more element on that top bar align right but keeping the user at the far right we should offer more empty slots
- If the plugin is not loaded then the positioning of a third party extension will change if it wants it to be rightly aligned depending if this plugin is loaded or not. So I would always add the spacer in the top bar.

packages/user/style/menu.css
```diff
@@ -0,0 +1,52 @@
+/*-----------------------------------------------------------------------------
+| Copyright (c) Jupyter Development Team.
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+
+.jp-MenuBar-label {
+  margin-left: 25px;
+}
+
+.jp-MenuBar-anonymousIcon {
```
fcollonval, line 10: Don't hesitate to merge the definition if they must match to highlight their link if modified later:

```suggestion
.jp-MenuBar-anonymousIcon,
.jp-MenuBar-imageIcon {
```

Same for the inner elements

packages/user/test/user.spec.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { JupyterServer } from '@jupyterlab/testutils';
```
fcollonval, line 4: Could you please drop the test part if none are implemented?

fcollonval (changes requested):
Thanks @hbcarlos

please find some suggestions for this PR.

packages/metapackage/package.json
```diff
@@ -12,16 +12,22 @@
   },
   "license": "BSD-3-Clause",
   "author": "Project Jupyter",
-  "sideEffects": false,
+  "sideEffects": [
```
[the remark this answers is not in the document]
blink1073: Yeah I wouldn't have expected this change.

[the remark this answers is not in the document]
hbcarlos: I did this change, this is what I wanted to ask you @fcollonval. I do not remember exactly but or the test where failing or I could not push the changes because the integrity command was asking for this change.

packages/user-extension/src/index.ts
hbcarlos: May the menubar should be 2000?

packages/user/test/user.spec.ts
hbcarlos: What do you mean? removing the test completely (user.spec.ts file)?

packages/metapackage/package.json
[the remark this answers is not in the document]
fcollonval: I tried locally adding a new package and these changes are created when running the integrity test. So I assume that without them the CI test will fail.

packages/user/src/tokens.ts
hbcarlos: > Do we really want to list attributes like gender, birthDate, honorificPrefix,... ?

I tried to make it as generic as possible selecting some of the attributes listed in https://schema.org/Person that I thought could be useful for others.
But as you said, maybe its better to extend from ReadOnlyJSONObject and everyone can define whatever they need.

packages/user-extension/src/index.ts
hbcarlos: Should this be the interface `IUser` instead of `ICurrentUser` and the same on the next comment?

galata/test/jupyterlab/general.test.ts
```diff
@@ -6,11 +6,15 @@ import { expect } from '@playwright/test';
```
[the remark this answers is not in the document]
hbcarlos: Do you mean to solve the problem of having the same username and color for all the tests?

packages/metapackage/package.json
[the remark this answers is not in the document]
fcollonval: Actually the good way to fix this is to update the CSS skip list in https://github.com/jupyterlab/jupyterlab/blob/f65faadc14ee18b45d5ea06de6472d1b04cab5a0/buildutils/src/ensure-repo.ts#L145-L226

Thanks @jtpio for this one

packages/user/test/user.spec.ts
martinRenou: I suppose you should remove this file yes. We can add it back when there are actually tests.

packages/user/src/tokens.ts
martinRenou: I agree with @fcollonval that we should keep it very simple (`id`, `name`, `color`) and then add what the code needs (`cursor`, and maybe more, like `currentNotebook`, `currentCell`...)

packages/user-extension/src/index.ts
fcollonval: Currently the elements added to the top bar have rank 100. The default rank is 900. So i proposed to set the spacer at the default rank of 900 and to set the new menubar at 1000.

packages/metapackage/package.json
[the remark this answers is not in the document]
hbcarlos: I added the user and user-extension packages to the list and I still have the error when pushing commits.

```=bash
$ node scripts/ensure-buildutils.js && node buildutils/lib/ensure-repo.js
{
  "@jupyterlab/metapackage": [
    "Style files not included in sideEffects in /home/carlos/Documents/jp-user/jupyterlab/packages/metapackage"
  ],
```

galata/test/jupyterlab/general.test.ts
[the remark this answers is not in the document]
fcollonval: No to speed thing up because by explicitly calling `page.goto`, you are asking for a page reload.

packages/metapackage/package.json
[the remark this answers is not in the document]
fcollonval: Did you set `sideEffects` to `false`?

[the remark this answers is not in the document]
fcollonval: You need to revert all changes in metapackage in addition to add the CSS skip list update.

packages/user/test/user.spec.ts
hbcarlos: I can not remove the tests. The ensure repo script complains.

packages/metapackage/package.json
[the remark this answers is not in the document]
hbcarlos: Yes, I set sideEffects to false.

I reverted all the changes in metapackage, but when running the ensure repo script adds all the CSS even with the user and user-extension packages in the list of skip CSS.

jtpio:
Thanks @hbcarlos for the continued effort on this :+1: 

> Now users will see his name and icon at the right side of the toolbar.

Posting here a summary of a discussion with Carlos on another channel.

At the moment the plugin adds the name and icon to the top area, even if the user does not plan to use any of the RTC features to collaborate with others. For example a single user only using JupyterLab to do some work locally.

Since this was the subject of a discussion at the time when we added the Simple Interface toggle (see https://github.com/jupyterlab/jupyterlab/pull/9100, https://github.com/jupyterlab/jupyterlab/issues/9105 and https://github.com/jupyterlab/jupyterlab/pull/9296 for more context), I would like to raise a concern about having this name / icon indicator visible at all time in the top area.

This is also related to some user feedback (for example from https://github.com/jupyter/notebook/issues/6210) about the JupyterLab UI having too many distractions, and a new visual indicator in the top area might add some more noise.

Possible alternatives we discussed could be to place to indicator in the bottom left corner (similar to VS Code):

image

image

Not sure the left area handler is currently able to pull items down. Otherwise it could also be a good opportunity to add that feature (in a separate PR).

hbcarlos:
Thanks for commenting on the user menu @jtpio. I'll remove the widget and we can tackle that on another PR.

fcollonval:
I fixed the integrity test; in fact in addition to the new user packages, the `docprovider` package style needed to be skipped because it has now style (due to its dependency on the new user package).

Regarding the very good point raised by @jtpio, my understanding from the JLab team meeting was that no UI elements should be displayed if JupyterLab was not executed with collaborative edition. This means that the collaboration state needs to be stored in the pageconfig object.

hbcarlos:
> I fixed the integrity test; in fact in addition to the new user packages, the docprovider package style needed to be skipped because it has now style (due to its dependency on the new user package).

Thanks, @fcollonval! I did not realize about the `docprovider`.

> This means that the collaboration state needs to be stored in the pageconfig object.

That's actually a good idea. I can remove the menu if the collaborative flag is false.

fcollonval:
> That's actually a good idea. I can remove the menu if the collaborative flag is false.

:+1:

jtpio:
> Note regarding retrolab: it does not have sidebars. So the current proposition à la Google doc sounds better for me.

Since the plugin adding the UI element to the top area is quite small, RetroLab could also define its own if needed.

But if we keep the current one behind the `collaborative` flag, then it should work automatically in RetroLab (might have to tweak the ranks there though)

trungleduc:
👍  on the ideal of @jtpio, we can place the user widget to the bottom right corner and use it to indicate both the user information and the list of connected users.

hbcarlos:
I rebased to have the latest changes from master and remove the merge commit. Sorry, @fcollonval I could not save your commit, but I added those changes to the review commit.

Probably the UI-test will fail because now the user menu will only be created if the collaborative flag is true.
I'm not sure this is the best way of making the user menu plugin optional, if soo, I have to change the documentation to specify that the plugin is optional.

isabela-pf:
I'm glad to see this UI only appearing when it is actually in-use (collaborative flag true)! 

I am curious about the placement of the indicator. I don't think I understand why this belongs in the lower right corner since I don't think it will be as visible there. I'd also be curious what happens if the sidebars are full, since I didn't think there was an equivalent overflow `…` to account for that. Is there a reason the lower right corner fits well with the rest of the interface.

krassowski:
Should we integrate the conversation about placement with the one in https://github.com/jupyterlab/jupyterlab/issues/11528?

jtpio:
> Should we integrate the conversation about placement with the one in #11528?

Yes let's keep the conversation in one place if possible :+1:

hbcarlos:
I pushed some more commits, and this is ready now.

* I restored the integrity changed o the metapackage.
* I removed 3 commits, restoring the UI tests. I changed the UI tests to support the menu on the top right corner, but now the menu only appears when the collaborative flag is true soo I restored the test.
* I removed two unused react components.
* Added test for the User class and the user menu.

hbcarlos:
> We'll also want to start having UI tests for RTC, starting simple for now, that we could do independently of this PR.

Yes, that's another task on my to-do list.
I have to add tests to the shared-models and UI tests for RTC.

packages/user/src/tokens.ts
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * NOTE: Requirer this token in your extension to access the
+ * current connected user information.
+ */
+export const ICurrentUser = new Token<IUser>('@jupyterlab/user:ICurrentUser');
+
+/**
+ * The user menu token.
+ *
+ * NOTE: Requirer this token in your extension to access the user menu
```
jtpio, line 24:
```suggestion
 * NOTE: Require this token in your extension to access the user menu
```

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  provides: ICurrentUser,
+  activate: (app: JupyterFrontEnd, state: IStateDB): IUser => {
+    return new User(state);
+  }
+};
+
+/**
+ * Jupyter plugin providing the IUserMenu.
+ */
+const userMenuPlugin: JupyterFrontEndPlugin<IMenu | undefined> = {
```
jtpio, line 41: Two questions here:

- for clarity and coherence with the rest of the code base, should this be providing a `IUserMenu` instead of `IMenu`?
- probably the plugin should always provide `IUserMenu` instead of `IMenu | undefined`. Then another plugin would be responsible for adding it to the shell by checking the `collaborative` option. This would be similar to the pattern used for enabling the debugger sidebar here:

https://github.com/jupyterlab/jupyterlab/blob/1bf5fdff171f8a9e91755821efd376d5742778ba/packages/debugger-extension/src/index.ts#L565-L582

packages/user-extension/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@jupyterlab/user-extension",
+  "version": "3.3.0-alpha.15",
+  "description": "JupyterLab - User Widget",
```
jtpio, line 4:
```suggestion
  "description": "JupyterLab - User Extension",
```

packages/user-extension/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  activate: (app: JupyterFrontEnd, user: IUser): IMenu | undefined => {
+    const { shell, commands } = app;
+
+    if (PageConfig.getOption('collaborative') !== 'true') {
+      return undefined;
+    }
+
+    const spacer = new Widget();
+    spacer.id = 'jp-topbar-spacer';
+    shell.add(spacer, 'top', { rank: 900 });
```
jtpio, line 55: It's fine to have the spacer here for now, but maybe we should check how to make the top area more well defined so extensions can more easily add items there without stepping on each others toes.

For example [jupyterlab-topbar](https://github.com/jupyterlab-contrib/jupyterlab-topbar) also adds a spacer to align the items to the right. Which means there would be two spacers if the user menu and the top bar extensions (or any other extensions) are activated on the page.

So maybe we should just add that spacer independently of these plugins, by default in JupyterLab, and document using a rank < 900 to place items to the left, and rank > 900 to place them to the right.

packages/user-extension/src/index.ts
hbcarlos: > for clarity and coherence with the rest of the code base, should this be providing a IUserMenu instead of IMenu

I'm lost in this part.
The `IUserMenu` is the name of the token that the plugin provides, And the `IMenu` is the interface of the object provided by the plugin.
So, the plugin provides an `IUserMenu` and returns an `IMenu`, right?

> probably the plugin should always provide IUserMenu instead of IMenu | undefined. Then another plugin would be responsible for adding it to the shell by checking the collaborative option. This would be similar to the pattern used for enabling the debugger sidebar here:
Thanks for the clarification. I had no idea how to proceed in this case.

hbcarlos:
I pushed two more commits applying the changes from the review:
* Changed the name of the interface from IUser to ICurrentUser since it refers to the currently connected user.
* Renamed IUser to ICurrentUser everywhere.
* Renamed the IMenu interface to IUserMenu to specify that this menu is related to the current user.
* Split the user menu plugin into two plugins, one providing the menu and the other adding it to the top bar.

packages/user/src/tokens.ts
```diff
@@ -0,0 +1,244 @@
...
+    readonly anonymous: boolean;
+
+    /**
+     * User's role.
+     *
+     * A user can have the following roles:
+     *  - ADMIN: can
+     *  - READ:
+     *  - WRITE:
+     *  - RUN:
```
jtpio, line 112: Maybe we can leave that part of the docs out for now until this is implemented.

packages/user-extension/src/index.ts
jtpio: Opened https://github.com/jupyterlab/jupyterlab/issues/11589

jtpio:
@hbcarlos looks like there is one more conflict to fix. But looking with the recent changes, thanks for the patience!

Also, how about bringing that PR to the weekly call tomorrow one more time. And then we can move forward with it, merge and iterate?

This is targeting 4.0 so there will be time to make tweaks to the UI elements based on user feedback and address other concerns related to the top area (https://github.com/jupyterlab/jupyterlab/issues/11589).

As a follow-up of the merge, we should also make a new `4.0.0ax` release. Normally the releaser should be able to publish these new packages automatically but we can double check this will be the case (cc @blink1073 for confirmation).

blink1073:
Yeah, because we are setting `access: public` and the packages aren't private, they should get picked up by the workspace logic.

hbcarlos:
Thank you soo much for the review @jtpio and @fcollonval. I'll bring it to the team meeting today.

fcollonval (approved):
Thanks for the patience @hbcarlos 

@jtpio I let you have a final look and merge this one.

ellisonbg:
I will try to have a look today as well, if I don't comment by the end of the day, please move forward.

packages/user/src/model.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * User's role.
+   *
+   * NOTE: Jupyter Server and JupyterLab doesn't implement a role-base access control (RBAC) yet.
+   * This attribute is here to start introducing RBAC to JupyterLab's interface. At the moment every
+   * user has the role ADMIN.
+   */
+  get role(): IUser.ROLE {
```
ellisonbg, line 91: While I think it is the right direction to start adding roles, I don't know that this is the right place to put this. The reason is that there is little in place to ensure that the role returned here is the authoritative role given to the user by the system. For, example, an attacking extension could disable `ICurrentUser` and provide its own implementation that escalates the role. Thoughts on removing for now until we work through those details?

packages/user/src/model.ts
```diff
@@ -0,0 +1,200 @@
...
+   * User's ID.
+   */
+  get id(): string {
+    return this._id;
+  }
+
+  /**
+   * User's name.
+   */
+  get name(): string {
```
ellisonbg, line 54: It isn't clear what name this is. Is this mean to be full name, first name, etc. My advice would be for us to follow the convention of something like the schema.org person type, which uses `familyName` and `givenName`. That way as we build UI around this we can know how to construct short and long variants.

packages/user/src/model.ts
```diff
@@ -0,0 +1,200 @@
...
+        this._isReady = true;
+        this._ready.emit(true);
+      })
+      .catch(e => console.error(e));
+  }
+
+  /**
+   * User's ID.
+   */
+  get id(): string {
```
ellisonbg, line 47: It is not clear what the difference is between `id` and `username`. Are both needed? Can we document them in the interface?

packages/user/src/utils.ts
```diff
@@ -0,0 +1,174 @@
...
+ * This function returns names like "Anonymous Io" or "Anonymous Metis".
+ */
+export const getAnonymousUserName = (): string =>
+  'Anonymous ' +
+  moonsOfJupyter[Math.floor(Math.random() * moonsOfJupyter.length)];
+
+/**
+ * Predefined colors for users
+ */
+export const userColors = [
```
ellisonbg, line 139: Can we instead use a categorical color scheme that has been well tested for usability, constrast, etc. Something like category20 is a good starting point. We are also using this in other parts of JupyterLab (coloring autocompletion results).

https://vega.github.io/vega/docs/schemes/#category20

packages/user/src/utils.ts
```diff
@@ -0,0 +1,174 @@
...
+ */
+export const getRandomColor = (): string =>
+  userColors[Math.floor(Math.random() * userColors.length)];
+
+/**
+ * Extract the initials from the name of the user.
+ */
+export const getInitials = (name: string, familyName?: string): string => {
+  let initials = '';
+  const tmpName = name.split(' ');
```
ellisonbg, line 161: This type of logic won't generally be reliable and would be unneeded if we have `givenName` and `familyName` field in `ICurrentUser`

packages/user/src/utils.ts
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Get a random color from the list of colors.
+ */
+export const getRandomColor = (): string =>
+  userColors[Math.floor(Math.random() * userColors.length)];
+
+/**
+ * Extract the initials from the name of the user.
+ */
+export const getInitials = (name: string, familyName?: string): string => {
```
ellisonbg, line 159: Also, even if we have `givenName` and `familyName` in `ICurrentUser` it may be be clear how to build initials. Maybe it makes sense to have an `initials` field in `ICurrentUser` that can be set or which defaults to using the first initials of `givenName` and `familyName`.

ellisonbg (changes requested):
In this review I focus on the API and in particular `ICurrentUser`. Because this interface will be a public API that is used by a a lot of other extensions, I think we need to be considered about its design.

ellisonbg:
Question: does this PR introduce UI changes? If so, can you post an updated screenshot of the UI?

packages/user-extension/src/index.ts
ellisonbg: Yep, good point.

packages/user/src/model.ts
hbcarlos: Thanks for reviewing the PR @ellisonbg.

You are right. This is not the right place to ensure the authorization. All the authentication and authorization have to be handled by the backend by adding RBAC to all the end-points to return `401` when the user doesn't have the proper permissions.

Nevertheless, we need to somehow show different UI components depending on the user's role. For example, a user with `read` permission should not see the notebook in edit mode. And sure, any extension would be able to change the ICurrentUser plugin and assign any role to the user, but that doesn't mean that the user can edit the document. When the user tries to edit the document, the backed will respond with a `401`.

hbcarlos: The other reason to start adding the role already is that at the moment, all the users are `admin` (have the higher role). If we start to modify the UI deactivating buttons for other roles (for example, a user with role `read` can not see the button to open settings). This won't affect the actual behavior of JupyterLab because every user is `admin`.

This property doesn't alter the frontend but allows us to start adding RBAC to the front end without waiting until the backend is done.

packages/user/src/model.ts
hbcarlos: Agree, that is probably the best solution. The reason I removed the `familyName` is that we were not using it on the actual implementation, but I think I can change it and use `givenName` and `familyName`. For example for `Anonymous Metis` Anonymous could be the name and Metis the family name.

packages/user/src/model.ts
hbcarlos: The reason for adding the id is that at the moment, every user is anonymous, and we do not have a unique username or email. Having the id allows us to create a unique UUID for every user.

packages/user/src/utils.ts
hbcarlos: Yes, choose the colors, and I'll add them.
The colors we are using are the ones @dmonad uses in YJS, and they were introduced in the RTC PR.

ellisonbg: Here are the category 20 colors (as a concatenated string, I can separate these): https://github.com/vega/vega/blob/master/packages/vega-scale/src/palettes.js#L67

packages/user/src/model.ts
ellisonbg: Ahh, thanks for the clarification. Although doesn't this mean that `username` is not necessarily unique. Given how `username` is usually tied to Auth systems, I think that is an important constraint for us to have as the username may be used in other systems or extensions as a unique key. In my mind a username as the following properties: 1) it is unique, 2) is more human-readable than a uuid, but 3) possibly less human readable than a first + family name. What if we removed the `id` field and instead made sure that the username was unique even when anonymous.

packages/user/src/utils.ts
krassowski: The current colors we have were already checked for contrast: https://github.com/jupyterlab/jupyterlab/pull/10411#discussion_r651960093 and I haven't seen any complaints (does not mean those are the best they could be). @ellisonbg is there any colour on the current list that does not work well for you?

Vega's `category20` colors have excellent contrast against black text, but of the 20 colors from only 7 pass the contrast check for WCAG AA for large text when used as background for **white** text, which is the case when showing the user cursor; showing it with black text would be distracting (and depart from the convention adopted by Google Docs, Deepnote and many others). Many in `category20` fail badly (contrast ratio < 2.5):

```js
[
  '#1f77b4',   # 4.82
  '#aec7e8',   # 1.73 fail
  '#ff7f0e',   # 2.53 fail
  '#ffbb78',   # 1.66 fail
  '#2ca02c',   # 3.40
  '#98df8a',   # 1.58 fail
  '#d62728',   # 5.02
  '#ff9896',   # 2.06 fail
  '#9467bd',   # 4.25
  '#c5b0d5',   # 1.99 fail
  '#8c564b',   # 5.91
  '#c49c94',   # 2.46 fail
  '#e377c2',   # 2.73 fail
  '#f7b6d2',   # 1.66 fail
  '#7f7f7f',   # 4.00
  '#c7c7c7',   # 1.69 fail
  '#bcbd22',   # 2.01 fail
  '#dbdb8d',   # 1.44 fail
  '#17becf',   # 2.25 fail
  '#9edae5'    # 1.54
]
```

`category20b` would be a better starting point - only 8 failing contrast with white text:

```js
[
  '#393b79',   # 10.11
  '#5254a3',   # 6.65
  '#6b6ecf',   # 4.41
  '#9c9ede',   # 2.51 fail
  '#637939',   # 4.85
  '#8ca252',   # 2.83 fail
  '#b5cf6b',   # 1.73 fail
  '#cedb9c',   # 1.47 fail
  '#8c6d31',   # 4.83
  '#bd9e39',   # 2.58 fail
  '#e7ba52',   # 1.81 fail
  '#e7cb94',   # 1.57 fail
  '#843c39',   # 7.81
  '#ad494a',   # 5.48
  '#d6616b',   # 3.63
  '#e7969c',   # 2.27 fail
  '#7b4173',   # 7.37
  '#a55194',   # 4.99
  '#ce6dbd',   # 3.2
  '#de9ed6'    # 2.1 fail
]
```

There is also `category20c` is not good either:


```js
[
  '#3182bd',   # 4.15
  '#6baed6',   # 2.42 fail
  '#9ecae1',   # 1.75 fail
  '#c6dbef',   # 1.42 fail
  '#e6550d',   # 3.70
  '#fd8d3c',   # 2.31 fail
  '#fdae6b',   # 1.83 fail
  '#fdd0a2',   # 1.42 fail
  '#31a354',   # 3.23
  '#74c476',   # 2.12 fail
  '#a1d99b',   # 1.62 fail
  '#c7e9c0',   # 1.32 fail
  '#756bb1',   # 4.67
  '#9e9ac8',   # 2.65 fail
  '#bcbddc',   # 1.83 fail
  '#dadaeb',   # 1.38 fail
  '#636363',   # 6.00
  '#969696',   # 2.95 fail
  '#bdbdbd',   # 1.87 fail
  '#d9d9d9'    # 1.41 fail
]
```

The truth is there is only so many colors you can show with good contrast against text and with good contrast against each other. From experience, it is much less than 20.

You can use if you want to test another pallette: https://webaim.org/resources/contrastchecker/

packages/user/src/model.ts
hbcarlos: I guess we can remove `id` and create a UUID as a `username` for anonymous users. But, we will have to use the `givenName` and `familyName` when showing the collaborators on the screen.

I believe this is much better than the actual implementation. Thanks for the help, I'll implement it.

packages/user/src/model.ts
minrk: Coming from JupyterHub, the main use of roles is to resolve to permissions (scopes). All permission checks are then direct scope checks, and roles are purely a point of storage, resolved immediately to scopes and never used directly thereafter. In an environment like JupyterHub, JupyterLab (client-side) wouldn't have any use for roles because it is the responsibility of the server to map request credentials to a set of scopes. _But_ duplicating permission scope checks server & client side might be a maintenance pain!

But if your definition of 'role' is a little different (i.e. they are a few fixed bundles of permissions, not arbitrary named sets of permissions), then each JupyterLab 'role' would map just fine to a single JupyterHub 'scope', and that's no problem.

hbcarlos: Hi @minrk, thank you soo much for the feedback.

First of all, it would be great if we can all set a meeting to sync between JupyterHub, Jupyter-Server and JupyterLab to discuss and decide how to proceed with the implementation of RBAC across projects. I suggest to use the RTC meeting every two weeks.

You (JupyterHub team) have done an amazing job in 2.0 and I think we should follow the same pattern.

> In an environment like JupyterHub, JupyterLab (client-side) wouldn't have any use for roles because it is the responsibility of the server to map request credentials to a set of scopes

But what about fine grade permission to the content, and kernels API on Jupyter Server? For what I understood from [the documentation](https://jupyterhub.readthedocs.io/en/rbac/rbac/index.html) correct me if I'm worng, JupyterHub assign permissions to manage servers, but what about allowing multiple users to access a server with different permissions? For example, both can write documents but only one can send a execute request to the kernel.

hbcarlos:
> does this PR introduce UI changes? If so, can you post an updated screenshot of the UI?

I'm sorry, @ellisonbg, I missed this question. The UI is still, as mentioned in the first comment. There is only a small widget on the top bar with the user's icon and name. In addition, this widget is visible only when running JupyterLab with the collaborative flag to true.

packages/user/src/model.ts
minrk: > I suggest to use the RTC meeting every two weeks.

Yes, that makes sense. I will try to come once in a while, but that meeting is at a very difficult time for me, so I cannot come on a regular basis. I think I can come next week, though.

> But what about fine grade permission to the content, and kernels API on Jupyter Server? 

For JupyterHub, I believe that should be implemented as custom scopes for JupyterHub, which 2.0 doesn't _yet_ support. But I plan to, once we figure out more about what it should look like here. Part of my goal for this collaboration is to establish:

1. what do scopes look like for Jupyter Server, and
2. how can they be represented in JupyterHub

I see two ways to do it - express the full fine-grained scopes in JupyterHub, _or_ express only a short list of roles in JupyterHub, and then map them to permissions in Jupyter serer. The latter case only really works if single-server roles are relatively static, though.

But for designing permissions in JupyterLab, one thing you should assume is the ability to store your own scopes and role assignments in JupyterHub. Even if that feature doesn't exist yet.

packages/user/src/utils.ts
ellisonbg: Actually, I was thinking a bit more about the color scheme. I think it is important for the user colors to be something that are in the theme, so that themes an modify the colors used.

ellisonbg: Yep, you are right. The categoryX color sequences fair pretty poorly as a background upon which to render text. At the same time, I think there are two contrast dimensions this sequence needs to cover: 1) the contrast between the colors, so users are distinguishable by color, 2) the contrast of text (white or black) on the color.

My initial thought is the generate a sequence using something like the following: 

* Work in HCL colorspace, and pick a single value for the Luminance. Because the Luminance is the primary dimension that determines the contrast against white/black text, we should pick a L that gives the needed WGAC constrast numbers.
* Pick a mind-ranged Chrome, to give enough color to give a strong visual effect, but not so high as to run into colors that our out of the RGB gamut.
* Then generate the sequence by rotating through the huge values.

I have used the R [colorspace](https://cran.r-project.org/web/packages/colorspace/vignettes/colorspace.html) library before and its [Python port](https://python-colorspace.readthedocs.io/en/stable/hclcolorspace.html).

One question: how were the current colors generated.

packages/user/src/model.ts
hbcarlos: > I will try to come once in a while, but that meeting is at a very difficult time for me, so I cannot come on a regular basis. I think I can come next week, though.

We changed the time since it was late for Europe. Now it is Monday 5:00 pm CET (the next one is next week, Monday 13th)

> But I plan to, once we figure out more about what it should look like here

Then we should work together to come out with the shape of RBAC in JupyteLab.

Pining @davidbrochart since he might be interested in the discussion.

packages/user-extension/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@jupyterlab/user-extension",
+  "version": "3.3.0-alpha.15",
```
jtpio, line 3: Let's update to `4.0.0-alpha.1` as a follow-up to https://github.com/jupyterlab/jupyterlab/pull/11537:

```suggestion
  "version": "4.0.0-alpha.1",
```

packages/user/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@jupyterlab/user",
+  "version": "3.3.0-alpha.15",
```
jtpio, line 3: Let's update to `4.0.0-alpha.1` as a follow-up to https://github.com/jupyterlab/jupyterlab/pull/11537:

```suggestion
  "version": "4.0.0-alpha.1",
```

hbcarlos: Thanks, @jtpio! I'm on it.

hbcarlos:
This is now ready for review again. I added three more commits:

- Remove StateDB and use local storage, because StateDB persists content by workspace.
- Applied changes requested by @ellisonbg, (Regarding the colors I'm waiting for a final decision).
- Rebased and updated packages version to solve conflicts.

packages/user/src/model.ts
minrk: > Now it is Monday 5:00 pm CET (the next one is next week, Monday 13th)

Yeah, 5-7 CET will always be hard, every day for me. But I can move things around once in a while. There's no way around it, I just can't participate regularly in synchronous meetings with the US (I think it's a good practice for all community projects to minimize synchronized meetings).

FWIW, over in jupyter-server, there are no plans to implement roles (https://github.com/jupyter-server/jupyter_server/pull/165 and https://github.com/jupyter-server/jupyter_server/pull/540) only scopes. That's all that's needed for fine-grained permissions. Roles are a convenient way for managing assignment of scopes to users, but not part of the actual enforcement of permissions.

So if there is something to store here, I would probably say it should be `scopes` or `permissions` rather than `role`. And it should be a collection, not a single value (that would be true for roles, too, btw). That could be a dict of `{resourceName: permissionLevel}` (probably easiest to work with, matches current [plan](https://github.com/jupyter-server/jupyter_server/pull/165) in jupyter-server), or a list of string scopes.

> Then we should work together to come out with the shape of RBAC in JupyterLab.

That's my plan! As much as possible via GitHub issues, to maximize visibility, participation, etc.

packages/user/src/utils.ts
SylvainCorlay: We could add a categorical color scale to the theme, so that we can override it in custom themes?

packages/user/src/utils.ts
hbcarlos: I added the `initials` field to the `ICurrentUser`. I'm keeping the function as convenient to extract the initials from `givenName` and `familyName` in multiple places for anonymous users without repeating the code. However, this is now done when fetching the user information so every organization can change the logic of creating the initials.

The logic of this function has into account composed names and names with two last names (which is the case of the Spanish culture). For example, "Jose Carlos" could be a `givenName` and "Gonzáles García" is the last name.
Having this in mind, this function will extract the following initials depending on the input:

- `Jose Carlos`, `Gonzáles García` -> `JG`
- `Jose Carlos`, `` -> `JC`
- `Jose`, `Gonzáles` -> `JG`
- `Jose`, `` -> `J`

packages/user/src/model.ts
hbcarlos: I'm going to remove the property `role` from the `ICurrentUser` interface, this way we can merge the PR.
Meanwhile, I will look into the Jupyter Server and open another PR for RBAC later.
For now, there is an issue open to discuss how to implement RBAC in the frontend and I think we should continue the conversation there:
* https://github.com/jupyterlab/jupyterlab/issues/11355

packages/user/test/model.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { User } from './../src/model';
+
+describe('user/model', () => {
+  describe('User', () => {
+    const stateDB = new StateDB();
+    let user: User;
+
+    it('Should emit user ready', async () => {
+      user = new User(stateDB);
```
jtpio, line 15: Looks like the `stateDB` doesn't need to be passed when instantiating a new `User` here in the tests?

(caught by the CodeQL check)

packages/user/src/utils.ts
krassowski: > 1) the contrast between the colors, so users are distinguishable by color, 2) the contrast of text (white or black) on the color.

and also 3) contrast of the cursor buble with the background; this can be worked around by adding a thin 1-px (or even sub-pixel) outline for dark themes, but yes ideally we would have different variations per theme, but possibly we would not want to change the colours completely (only their Luminance) so that users can still refer to their colours regardless of the theme ("can you see my green cursor?")

packages/user/src/utils.ts
vidartf: Does this handle non-latin names?

hbcarlos: It is meant to handle common names (name, last name) plus composed names.
I updated the table to clarify.
Do you have an example of the name you are worried about?

SylvainCorlay:
Let's get this in!

ellisonbg (approved):
Looks good, thanks for addressing the feedback!

jtpio:
Looks like the integrity check failure is relevant: https://github.com/jupyterlab/jupyterlab/runs/4462279685?check_suite_focus=true

hbcarlos:
Thanks Jeremy.
I rebased to pick up the changes from the TypeScript update.

krassowski:
Is user menu empty supposed to be empty by default? Should the dropdown indicator be removed when it is empty, or maybe a disabled placholder item "(empty)" added?

user-menu

jtpio:
> Edit: feel free to move it out to an issue if you really want to merge soon, but let's address it before a release.

@krassowski do you think this should be addressed here? Or could it be done in a follow-up PR? When we merge this PR, we'll have to make a release right after to fix the CI.

packages/user/src/utils.ts
vidartf: I mean e.g. chinese, japanese, hebrew, arabic, etc names written in their native script. Common assumptions about latin names do not always translate to such scripts (e.g. picking and capitalizing the first character of some parts of the name, split by whitespace, might become completely intelligible, and maybe downright rude).

krassowski:
Follow-up PR would be fine.

packages/user/src/utils.ts
vidartf: A reasonable resource: https://www.w3.org/International/questions/qa-personal-names

vidartf: To be clear: It might be completely ok to leave this as it is right now. I was mainly trying to stimulate the questions "What do we do if the name is not in latin script? Will the code fail with an exception? Does the concept of initials make sense in those cases? What do we do if it does not?"

hbcarlos: Those are good questions. Thanks for raising them.
Unfortunately, I do not have the answer.
In a case not contemplated here, the code won't fail. It could be a case where this function returns initials that don't make sense in other languages.

hbcarlos:
> Follow-up PR would be fine.

Thanks, @krassowski!
I'm already working on it, I'll add on a follow-up PR as soon as it is done.

hbcarlos:
I'm pushing one more commit to tag the `ICurrentUser` token and interface as experimental.
This will give us more flexibility for future changes.

jtpio:
Thanks!

Let's merge and address the remaining items in follow-up issues and PRs.

jtpio:
Making a new alpha release now.

martinRenou:

martinRenou:
Oh come on, be nice Meeseeks

fcollonval:

martinRenou:
Thanks for trying @fcollonval, but I also realized there would be conflicts. I made a big PR manually backporting all the IUser and collaboration PRs https://github.com/jupyterlab/jupyterlab/pull/13242 (still draft for now)

PR #11445, opened 2021-11-12
Makes restorer parameter optional in toc-extension

### LINKED ISSUE #11444: Make ILayoutRestorer optional in toc-extension
### Problem

`IRetroShell` in RetroLab cannot use the `toc-extension` because the required parameter `restorer` (of type `ILayoutRestorer`) is not accessible. Error:

```
Plugin '@jupyterlab/toc:plugin' failed to activate. index.es6.js:292:25
Error: No provider for: @jupyterlab/application:ILayoutRestorer
```

### Proposed Solution

Make the `restorer` parameter optional in `toc-extension`.

### Additional context

See also PR #11420 which made `labShell` optional in this extension.

jweill-aws:
## References

Fixes #11444 .

## Code changes

Makes the `restorer` parameter in `toc-extension` optional so that shells that do not implement a layout restorer, such as `IRetroShell`, can use this extension.

## User-facing changes

None.

None.

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -52,26 +52,26 @@ namespace CommandIDs {
  * @private
  * @param app - Jupyter application
  * @param docmanager - document manager
- * @param editorTracker - editor tracker
- * @param markdownViewerTracker - Markdown viewer tracker
- * @param notebookTracker - notebook tracker
  * @param rendermime - rendered MIME registry
  * @param translator - translator
+ * @param editorTracker - editor tracker
  * @param restorer - application layout restorer
  * @param labShell - Jupyter lab shell
+ * @param markdownViewerTracker - Markdown viewer tracker
+ * @param notebookTracker - notebook tracker
  * @param settingRegistry - setting registry
  * @returns table of contents registry
  */
 async function activateTOC(
   app: JupyterFrontEnd,
   docmanager: IDocumentManager,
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

fcollonval:
I think we can go a step further and make the widget trackers optional; in particular in retrolab, only one tracker is available per current document, isn't it?

```
    IEditorTracker,
    IMarkdownViewerTracker,
    INotebookTracker,
```

> I'll open a PR because actually the logic should be inverted. The table of content should provide the registry and the generator should be defined in the appropriate packages (INotebookTracker -> notebook, IEditorTrakcer -> fileeditor,...).

jtpio:
Good point :+1: 

Yes let's make a bigger pass on that `toc-extension` to put more dependencies as optional if possible (which can be done as a separate PR)

> The table of content should provide the registry

Making one of the toc plugins return the registry sounds good. Looking at the current state the following logic should probably be in its own plugin too:

https://github.com/jupyterlab/jupyterlab/blob/5755ea86fef3fdbba10cd05b23703b9d60b53226/packages/toc-extension/src/index.ts#L199-L219

jweill-aws:
Per @jtpio and @fcollonval , I've updated this PR to make additional parameters optional.

fcollonval (approved):
LGTM thanks @jweill-aws

fcollonval:

PR #11446, opened 2021-11-12
Close answered issues more quickly if there are no updates.

jasongrout:
## Code changes

This updates the answered issue workflow to simplify the configuration and close issues more quickly if there are no updates. It now marks answered issues for closure after a week of no updates, and then closes them 5 days later if there are still no updates. We deliberately set the previous timeline as very conservative, and while the process went well, it doesn't seem like leaving answered issues open for a month helped us.

### DIFF
--- a/.github/workflows/answered.yml
+++ b/.github/workflows/answered.yml
@@ -14,13 +14,11 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
I'm :+1: on this one - thanks @jasongrout 

> CI error is not related.

PR #11447, opened 2021-11-12
Do not update contextual help inspector if there would be no change.

### LINKED ISSUE #11410: Avoid refreshing contextual help just to show the same thing again
### Problem
The contextual help pane updates its view, even if it is just to re-display the same thing. This can reset the scroll and disrupt reading in the context help pane. 

### Proposed Solution

Avoid updating/refreshing the contextual help pane if it would be just for showing the same docs it's already displaying.

jasongrout:
## References

Fixes #11410

## Code changes

Keep the last inspection reply data and do not reset the inspector if the new reply is the same as the old one.

## User-facing changes

This means that slight changes in the editor cursor which give the same inspection output (such as moving the cursor inside the name of a function call) won’t reset the help content (thus resetting the scroll position, losing your place, etc.).

To test, execute the following code:

```python
with open('a', 'w') as f:
    f.write('a')
```
Then open the contextual help panel, put your cursor in the word `open`. Then scroll the contextual help panel to see the bottom of the text there.

Now, click anywhere else in the word `open` (or move your cursor). Before this PR, the contextual help panel would reset, losing the scroll position. After this PR, the contextual help panel will not be reset, preserving the scroll position in the panel.

### DIFF
--- a/packages/inspector/src/handler.ts
+++ b/packages/inspector/src/handler.ts
@@ -5,7 +5,7 @@ import { CodeEditor } from '@jupyterlab/codeeditor';
 import { Text } from '@jupyterlab/coreutils';
 import { IRenderMimeRegistry, MimeModel } from '@jupyterlab/rendermime';
 import { IDataConnector } from '@jupyterlab/statedb';
-import { ReadonlyJSONObject } from '@lumino/coreutils';
+import { JSONExt, ReadonlyJSONObject } from '@lumino/coreutils';
 import { IDisposable } from '@lumino/disposable';
 import { Debouncer } from '@lumino/polling';
 import { ISignal, Signal } from '@lumino/signaling';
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
LGTM

bluss:
Thanks a lot! :slightly_smiling_face:  The binder thing didn't quite manage to build and start in reasonable time unfortunately. Just curious, I guess at this layer we'll have to do the query and can only ignore if the answer is the same, it would be too much to assume we always get the same answer for the same query?

<details>
<summary> *Another thing about inspector and jank* </summary>

I'll mention this here. I promise I'm not interested in moving goalposts, but maybe we can open another issue.

If I write this in jupyterlab, python kernel:  sns.relplot( | )  and place the cursor at the `|`, I get the inspector pane showing the docs. That's great.
When typing with the cursor there, we can notice that the kernel is working to update the inspector pane. I often notice some jank coming from keeping the inspector open when working, typing - then I'll have to close it, so that it doesn't spend time looking up stuff under the cursor all the time.

Has anyone else seen this?
</details>

fcollonval:
> I guess at this layer we'll have to do the query and can only ignore if the answer is the same, it would be too much to assume we always get the same answer for the same query?

Yes

---

For your other point I opened a new issue #11458 - thanks for sharing your experience.
FYI the inspector requests are throttled to avoid requesting update too often. But from your experience it may not be sufficient (or there is a flaw to discover).

fcollonval:

PR #11432, opened 2021-11-10
Enforce ascii-only identifiers

jasongrout:
## Code changes

The article at https://certitude.consulting/blog/en/invisible-backdoor/ and the HackerNews comment https://news.ycombinator.com/item?id=29172889 inspired adding this rule. Essentially, this guards against having tricky unicode identifiers (like invisible characters) create a backdoor that is not visible during code review.

Our code currently passes with this check.

PR #11449, opened 2021-11-12
Enforce ascii-only identifiers

jasongrout:
## References

This backports #11432 to 3.2.x.

## Code changes

The article at https://certitude.consulting/blog/en/invisible-backdoor/ and the HackerNews comment https://news.ycombinator.com/item?id=29172889 inspired adding this rule. Essentially, this guards against having tricky unicode identifiers (like invisible characters) create a backdoor that is not visible during code review.


fcollonval (approved):
Thanks @jasongrout

PR #11450, opened 2021-11-13
Refactor status bar

fcollonval:
## References

Due to history (the status bar was first an extension outside of core), default widgets are defined in the status bar package. This makes it not easily reusable for remix application.

This removes all default widgets from the status bar. So it focuses on providing the application status bar in the spirit of composing advanced application from fundamental plugins.

## Code changes

Running sessions and kernel status are moved in `apputils`
Code editor position is moved in `codeeditor`

Other extensions (codemirror, console and notebook) are registering provider via new tokens (`IPositionModel` and `IKernelStatusModel`) to connect to the status bar items.

`HoverBox` is moved to `ui-components`.

## User-facing changes

N/A

Some status bar widgets have moved to new packages. Their id have been updated to reflect their new ownership.

Line form widget styles are using regular classes (dropping `typestyle`).


packages/application-extension/src/index.tsx
```diff
@@ -1060,6 +1063,63 @@ const topSpacer: JupyterFrontEndPlugin<void> = {
   }
 };
 
+/**
+ * The simple interface mode switch in the status bar.
+ */
+const modeSwitchPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/statusbar-extension:mode-switch',
```
jtpio, line 1070: Looks like we could also rename the plugin id, for example:

```suggestion
  id: '@jupyterlab/application-extension:mode-switch',
```

packages/apputils/src/kernelstatuses.tsx
```diff
@@ -11,7 +8,37 @@ import {
...
 import { JSONArray, JSONExt } from '@lumino/coreutils';
 import React from 'react';
-import { interactiveItem, TextItem } from '..';
+import { ISessionContext } from './sessioncontext';
+
+/**
+ * Helper function to translate kernel statuses mapping by using
+ * input translator.
+ *
+ * @param translator - - Language translator.
```
3coins, line 17: Are the double dashes expected here?

packages/codeeditor/src/lineCol.tsx
```diff
@@ -125,15 +120,19 @@ class LineFormComponent extends React.Component<
                 this._textInput = input;
               }}
             />
-            <div className={lineFormButtonDiv}>
+            <div className="jp-baseLineForm jp-lineFormButtonDiv">
```
3coins, line 123: **Nit**: Use container or wrapper instead of a tag name div in the class name.

packages/codeeditor/style/linecol.css
```diff
@@ -0,0 +1,74 @@
...
+  border: none;
+  border-radius: 0px;
+  position: absolute;
+  background-size: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  outline: none;
+}
+
+.jp-lineFormButtonDiv {
```
3coins, line 26: **Nit**: See my comment above, use "jp-lineFormButtonWrapper"

3coins (commented):
@fcollonval Added some minor comments, looks good otherwise. Does this change have any visual changes?

packages/application-extension/src/index.tsx
```diff
@@ -1060,6 +1063,63 @@ const topSpacer: JupyterFrontEndPlugin<void> = {
...
+    }
+    const trans = translator.load('jupyterlab');
+    const modeSwitch = new Switch();
+    modeSwitch.id = 'jp-single-document-mode';
+
+    modeSwitch.valueChanged.connect((_, args) => {
+      shell.mode = args.newValue ? 'single-document' : 'multiple-document';
+    });
+    shell.modeChanged.connect((_, mode) => {
+      modeSwitch.value = mode === 'single-document';
```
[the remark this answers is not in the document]
fcollonval: Thanks for the suggestion

packages/apputils/src/kernelstatuses.tsx
fcollonval: Good catch

packages/codeeditor/src/lineCol.tsx
fcollonval: :+1:

3coins (approved):
Thanks for making the changes. LGTM!

3coins:
Failing UI tests should be fixed after rebase from master, fixed by @jtpio [here](https://github.com/jupyterlab/jupyterlab/commit/90d933cd808af3c4e7b882d38e475d03d0de6490).

Issue #11452, opened 2021-11-13
executionScheduled should be emitted after the code cell's execute() is called


xiaohk:
## Description

Right now the `executionScheduled` signal is emitted before `CodeCell.execute` is called.

https://github.com/jupyterlab/jupyterlab/blob/5755ea86fef3fdbba10cd05b23703b9d60b53226/packages/notebook/src/actions.tsx#L2026-L2027

It can cause execution racing in widget development. See the example below.

## Reproduce

Say we are developing a widget that automatically executes a code cell A whenever any other code cell is executed. We can listen to the `executionScheduled` signal:

```ts
// Listen to the notebook execution events so we can auto-run cell A
NotebookActions.executionScheduled.connect((_, args) => {
  if (args.cell.node !== this.cell.node) {
    this.execute();
  }
});
```

Our custom `execute` in this widget is defined as:

```ts
/**
 * Run the code cell. The implementation logic is: (1) change the original cell
 * as the active cell, (2) use the NotebookActions to run the current active
 * cell
 */
execute = () => {
  // Find the cell index of the original cell
  // https://github.com/jupyterlab/jupyterlab/blob/5755ea86fef3fdbba10cd05b23703b9d60b53226/packages/notebook/src/widget.ts#L1803
  const cellIndex = ArrayExt.findFirstIndex(
    this.notebook.content.widgets,
    widget => widget.node === this.originalCell.node
  );

  // Change the active cell to the original cell
  this.notebook.content.activeCellIndex = cellIndex;

  // Run the active cell
  NotebookActions.run(
    this.notebook.content,
    this.notebook.context.sessionContext
  );
};
```

Now we run cell B; it would trigger cell A to run as well.

## Expected behavior

We would expected cell B to execute before cell A. However, cell A would actually run before cell B, because executionScheduled is emitted [before we actually call cell B's `execute` function](https://github.com/jupyterlab/jupyterlab/blob/5755ea86fef3fdbba10cd05b23703b9d60b53226/packages/notebook/src/actions.tsx#L2026-L2027).

A workaround would be setting a timeout:

```ts
// Listen to the notebook execution events so we can auto-run cell A
NotebookActions.executionScheduled.connect((_, args) => {
  if (args.cell.node !== this.cell.node) {
      setTimeout(() => {
        cd.execute();
      }, 1);
  }
});
```

## Context

- Operating System and version: macOS 12.0.1
- Browser and version: Firefox Developer 95.0b6
[...]

PR #11453, opened 2021-11-13
Fix executionScheduled emit order issue

xiaohk:
## References

#11452 

## Code changes

The order of `emit()` and `execute()` is swapped.

## User-facing changes

No user-facing changes

There should be no backwards-incompatible issues

### DIFF
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2023,10 +2023,14 @@ namespace Private {
             return Promise.resolve(false);
           }
           const deletedCells = notebook.model?.deletedCells ?? [];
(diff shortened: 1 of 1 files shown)


fcollonval:
We should certainly not do that permutation as mentioned in https://github.com/jupyterlab/jupyterlab/issues/11452#issuecomment-968620967

`executionScheduled` means _The execution of a cell has been planed at a later time_. It promises that your cell is not yet executed but will be later.

If we do the permutation suggested in #11453 (aka emitting `executionScheduled` after starting the execution). We can end up in a race condition because there is now way to ensure all listeners will receive `executionScheduled` before `executed` started to be emitted.

xiaohk:
Thanks for the feedback! I will look into the kernel message instead.

Issue #11456, opened 2021-11-14
JupyterLab fails to build


amzon-ex:
## Description

JupyterLab fails to build using `jupyter lab build`. I read a few issues relating to this in the repo, but they don't seem to answer my issue.

## Context

JupyterLab installed with pip.

- Operating System and version: WSL Ubuntu 20.04
- Python version: 3.8.10
- Browser and version: Edge 95 
- JupyterLab version: 3.2.3
- Nodejs version: 17.1.0

## Error Log
```
[LabBuildApp] > node /home/amzon-ex/.local/lib/python3.8/site-packages/jupyterlab/staging/yarn.js run build:prod:minimize
[LabBuildApp] yarn run v1.21.1
$ webpack --config webpack.prod.minimize.config.js
[webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve './base.css' in '/home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/@mohirio/jupyterlab-horizon-theme/style'
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/Compilation.js:2011:28
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:795:13
    at eval (eval at create (/home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:10:1)
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:275:22
    at eval (eval at create (/home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:431:22
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:124:11
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:667:25
    at /home/amzon-ex/.local/share/jupyter/lab/staging/node_modules/webpack/lib/NormalModuleFactory.js:852:8
[...]

Issue #10554, opened 2021-07-06
Improve the user experience and awareness about prebuilt extensions


krassowski:
### Problem

Many users find themselves trying to install extensions from source; this is not surprising especially given it was the only way for a long time and that the Extension Manager uses this approach.

Several extension authors who migrated to pre-built extensions either abandoned publishing source releases to NPM, or are unable to support troubleshooting errors in the built step following installation from source (which is often the reason for only publishing to PyPI/conda). However, many users do not know about the pre-built extensions and are stuck trying to install the source version and failing either because the new version compatible with JupyterLab 3 was never published to NPM, or because they encounter build errors (and there are not enough volunteers to help them troubleshoot). This leads to frustrating user experience; there are multiple examples:

<details>
- ["JupyterLab failed to build"](https://discourse.jupyter.org/t/jupyter-lab-failed-to-build/9745), discourse (problem with building after installing `@lckr/jupyterlab_variableinspector` solved by `conda install`)
- ["hate-hate relationship between me and jupyterlab"](https://twitter.com/psychemedia), Tony Hirst on Twitter (`jupyterlab-skip-traceback` 4.0 with JupyterLab 3.x support published on PyPI but not on NPM, I suggested pip install)
- ["Unable to install jupyterlab-execute-time extension"](https://stackoverflow.com/q/67390716/6646912), StackOverflow (`jupyterlab-execute-time` 2.x with JupyterLab 3.x support published on PyPI but not on NPM)
- ["Can't install any jupyterlab extensions in docker image"](https://stackoverflow.com/q/67599080/6646912), StackOverflow (hours spent on trying to troubleshoot problems with installing `@jupyter-widgets/jupyterlab-manager` despite of a simple solution of `pip install jupyterlab_widgets`, actually appears to be the same as:  10272)
- ["How to match the versions of jupyterlab-nvdashboard and JupyterLab"](https://stackoverflow.com/q/67817972/6646912), StackOverflow (`jupyterlab-nvdashboard` not released on NPM but available on PyPI)
- ["RuntimeError: JupyterLab failed to build"](https://stackoverflow.com/q/68276261/6646912), StackOverflow (`jupyterlab-lsp` build error that I cannot and don't want to investigate as extension author as a) the poster did not provide sufficient information b) I won't have time for that)
[...]

PR #11476, opened 2021-11-16
Recommend trying prebuilt extension version in the build failure dialog

krassowski:
## References

Addresses point 3 in #10554 and the discussion in #11456. I would really appreciate suggestions on the wording.

## Code changes

Expand the build failed dialog message.

## User-facing changes

TBD

None - this message is not currently translated (which we should fix, but it's a separate matter - I added a note in #10737)

### DIFF
--- a/packages/services/src/builder/index.ts
+++ b/packages/services/src/builder/index.ts
@@ -82,7 +82,11 @@ export class BuildManager {
         throw new ServerConnection.ResponseError(response, 'Build aborted');
       }
       if (response.status !== 200) {
-        const message = `Build failed with ${response.status}, please run 'jupyter lab build' on the server for full output`;
+        const message = `Build failed with ${response.status}.
+
+        If you are experiencing the build failure after installing an extension (or trying to include previously installed extension after updating JupyterLab) please check the extension website for new installation instructions as many extensions migrated to the prebuilt extensions system which no longer requires rebuilding JupyterLab (but uses a different installation procedure, typically involving a package manager such as 'pip' or 'conda').
+
+        If you specifically intended to install a source extension, please run 'jupyter lab build' on the server for full output.`;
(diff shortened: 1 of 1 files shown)


packages/services/src/builder/index.ts
```diff
@@ -82,7 +82,11 @@ export class BuildManager {
         throw new ServerConnection.ResponseError(response, 'Build aborted');
       }
       if (response.status !== 200) {
-        const message = `Build failed with ${response.status}, please run 'jupyter lab build' on the server for full output`;
+        const message = `Build failed with ${response.status}.
+
+        If you are experiencing the build failure after installing an extension (or trying to include previously installed extension after updating JupyterLab) please check the extension website for new installation instructions as many extensions migrated to the prebuilt extensions system which no longer requires rebuilding JupyterLab (but uses a different installation procedure, typically involving a package manager such as 'pip' or 'conda').
```
jtpio, line 87: Agree that extensions can be distributed in many ways, but probably they will also be on public GitHub repositories?

```suggestion
        If you are experiencing the build failure after installing an extension (or trying to include previously installed extension after updating JupyterLab) please check the extension repository for new installation instructions as many extensions migrated to the prebuilt extensions system which no longer requires rebuilding JupyterLab (but uses a different installation procedure, typically involving a package manager such as 'pip' or 'conda').
```

jtpio:
Thanks, this will already make the error message more informative :+1:

packages/services/src/builder/index.ts
krassowski: Well the only thing we know is they have an NPM website, which might be well outdated. I agree that repository is better because it should be up-to-date even if authors no longer publish to NPM.

fcollonval (approved):
Thanks @krassowski

jtpio:

PR #11488, opened 2021-11-17
Fix Tex highlights affecting Markdown with standalone `$`

### LINKED ISSUE #11482: Markdown syntax highlighting broken when standalone dollar (`$`) present
## Description

Similar to #6774, but not due to code blocks but incorrect highlighting rule for `$math mode$` in markdown: math mode is only parsed if both opening and closing dollars are in the same line but the current implementation does not have this restriction causing valid GFM markdown to be incorrectly highlighted.

Observed out in the wild when following a statistical genetics practical - together with #10861 (I already sent a PR yesterday) made it very hard to follow.

## Reproduce

Paste the following in a markdown cell:

~~~markdown
Example command:

```sh
$ ls
```

runs `ls`; the $ is the command prompt (the text after dollar sign should NOT be highlighted)

```sh
$ cd some/path
```

And list contents again: (text in this line should NOT be highlighted, greyed out, nor in italics)

```sh
$ ls
```
~~~

See incorrect highlighting:

Screenshot from 2021-11-17 13-22-33

## Expected behavior

Should look like this:

Screenshot from 2021-11-17 13-43-53

## Context

- JupyterLab version: 3.2.3

I will send a PR after hours.

krassowski:
## References

Fixes #11482

## Code changes

- Adds a regular expression lookahead to the opening conditions for inline tex highlighting.
- Adds visual regression test

## User-facing changes

Before:

Screenshot from 2021-11-17 19-29-57

After:

Screenshot from 2021-11-17 19-27-00

None


fcollonval (approved):
Thanks @krassowski !

fcollonval:

PR #11496, opened 2021-11-18
Enforce labels on PRs

blink1073:
## References
Use the new action from [`maintainer-tools`](https://github.com/jupyterlab/maintainer-tools#enforce-labels).

## Code changes
Adds a workflow to ensure PRs are labeled with a triage label.

## User-facing changes
None

None

### DIFF
--- a/.github/workflows/enforce-label.yml
+++ b/.github/workflows/enforce-label.yml
@@ -1,0 +1,11 @@
+name: Enforce PR label
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


blink1073:
I verified that the build fails when there is no label and passes when there is a label.

fcollonval (approved):
Awesome thanks a lot @blink1073 

> Approving as the new job is green and the others are not relevant here.

blink1073:

PR #11499, opened 2021-11-18
Update screenshots and text for user interface docs

krassowski:
## References

Addresses https://discourse.jupyter.org/t/items-missing-from-the-left-sidebar/10794

## Code changes

None

## User-facing changes

Docs include updated screenshots and UI descriptions in a few (but not all) places.

None


krassowski:
I think that "Enforce PR label" ( https://github.com/jupyterlab/jupyterlab/pull/11496) should wait for the "Pull Request Labeler" or be re-triggered after labeller finishes. CC @blink1073.

fcollonval (approved):
Thanks a lot @krassowski - especially for taking the time to update the screenshot (we should definitely automated them...).

jtpio:
It would be great to have a small Galata / Playwright script to take these screenshots. Although we would need a way to add overlays (mouse pointer, arrow)

blink1073:
> I think that "Enforce PR label" ( #11496) should wait for the "Pull Request Labeler" or be re-triggered after labeller finishes. CC @blink1073.

That's a good point, it could be modified to be a dependent workflow, but then both would fail on PRs that end up trying to label all of the sub packages, since the labeler build fails.

fcollonval:
Would it work to set the two jobs in the same workflow file so that it can be triggered even if the labeller fails?

meeseeksmachine:

https://discourse.jupyter.org/t/items-missing-from-the-left-sidebar/10794/4

fcollonval:

fcollonval:

PR #11457, opened 2021-11-14
Only show the head of the outputs and ensure iopub outputs are correctly displayed

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/10045

## Code changes

We need to take into account the iopub messages when stripping the output to only show the defined max number of outputs.

This PR takes those messages into account. To achieve this, we also had to remove the display of the tail.

## User-facing changes

The outputs are correctly displayed when running e.g.

```python
for i in range(100):
    for x in range(10):
        print(x, flush=True)
    for x in range(10):
        print(x, file=sys.stderr, flush=True)
```

the tail is not more displayed (all outputs can still be seen if the user clicks on the message `Click on this message to get the complete output.`)

## Backwards-incompatible changes

None

PR #11502, opened 2021-11-19
Backport PR #11457 on branch 3.2.x (Only show the head of the outputs and ensure iopub outputs are correctly displayed)

fcollonval:


echarles (approved):
LGTM. The i18n CI failure have happened before on backport and is not related to this PR.

echarles:
Thx @fcollonval

krassowski:
> The i18n CI failure have happened before on backport and is not related to this PR.

It is related (the text of setting description was modified), but I would say that this is an acceptable to break translations here (i.e. users of fully-translated languages will see a regression in the form of missing translation for this string) because this is an important bugfix PR, and not a minor stylistic adjustment.

echarles:
true, it is related, but acceptable to be broken. Thx for that precision.

Issue #11490, opened 2021-11-17
Make the hidden cells button optional


schmidi314:
### Problem

In large notebooks with many headings that are usually collapsed, the hidden-cells buttons next to a collapsed heading take quite some space. I would like the display of this button to be optional.

### Proposed Solution

* Add an option "showHiddenCellsButton" to TOC
* Basically add dependency on this option in maybeCreateOrUpdateExpandButton() to
https://github.com/jupyterlab/jupyterlab/blob/7aba384baa09393e98432619b3355f57d7529012/packages/cells/src/widget.ts#L1604-L1608
(But it would probably be better to rework the if/else structure slightly)

### Open questions

* Should this option actually be located in TOC? Notebook might also be a good place for this.

schmidi314:
More specific solution proposal:
* Option `showHiddenCellsButton` goes to Notebook settings (ToC hasn't much to do with the collapsing of headings.)
* Option is communicated to `MarkdownCell` widget --> `MarkdownCell._showHiddenCellsButton`
* In `MarkdownCell.maybeCreateOrUpdateExpandButton()` this property is used to decide whether or not the button shall be displayed.

And this is what it looks like:

![optional_hidden_cells_button](https://user-images.githubusercontent.com/74381262/142507691-3c375a9f-0d7d-4490-9972-8c399c4822d0.gif)

Available in schmidi314/jupyterlab branch feature/optional_hidden_cells_button

...not sure what to do now. Pull request?

krassowski:
> ...not sure what to do now. Pull request?

 :+1: Opening a draft pull request is always a good idea once you are at a working prototype stage and would like to get feedback :)

I would suggest that there should always be an indication of cells being hidden even when `showHiddenCellsButton` is set to `false`. I would suggest when `showHiddenCellsButton == false` the little arrow which shows on headers when hovering over should be persisted as a subtle indication. Maybe even something more prominent (but obviously not as prominent as the current button which as you note takes a lot of space).

Another thing that we like to have before adding any more features/configuration options to notebook is more visual tests. The chances of the PR getting approved will rise significantly if it will come with a visual regression test (or a few) using [galata](https://github.com/jupyterlab/jupyterlab/tree/master/galata).

schmidi314:
Thanks for the input. Making the little arrow persistent sounds good to me. Also making the little arrow somewhat less little would be a general suggestion of mine. So, how about:
* make the arrow persistent if `showHiddenCellsButton == false`
[...]

PR #11504, opened 2021-11-19
Feature/optional hidden cells button

schmidi314:
# References
First prototype for issue #11490 

Todo: Test scripts fail (probably I have to update them?) and not yet looked into galata.

## Code changes
* Added option `showHiddenCellsButton` to Notebook Settings and communicated it to `MarkdownCell` widget. Default = true.
* Made behavior of hidden cells button (show or not) and collapse arrow-head buttons (hover or not) dependent on this option .
* Added two somewhat larger arrow head svgs. Not sure about design, though.

## User-facing changes
* With the default `showHiddenCellsButton=true` option, the only noticeable difference are somewhat larger arrow head buttons (left of the headings).
* With `showHiddenCellsButton=false`, the arrow heads are always displayed (not only on hover) and the hidden cells button won't be displayed.

optional_hidden_cells_button

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -153,8 +153,15 @@ const MARKDOWN_OUTPUT_CLASS = 'jp-MarkdownOutput';
 
 export const MARKDOWN_HEADING_COLLAPSED = 'jp-MarkdownHeadingCollapsed';
 
+/**
+ * The class names added to heading collapse button.
+ */
+const HEADING_COLLAPSER_HOVER_CLASS = 'jp-collapseHeadingButtonHover';
 const HEADING_COLLAPSER_CLASS = 'jp-collapseHeadingButton';
 
+/**
+ * The class names added to hidden cells button.
+ */
 const SHOW_HIDDEN_CELLS_CLASS = 'jp-showHiddenCellsButton';
 
 /**
@@ -1531,14 +1538,14 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
       this.model.metadata.delete(MARKDOWN_HEADING_COLLAPSED);
     }
     const collapseButton = this.inputArea.promptNode.getElementsByClassName(
-      HEADING_COLLAPSER_CLASS
+      this._showHiddenCellsButton ? HEADING_COLLAPSER_HOVER_CLASS : HEADING_COLLAPSER_CLASS
     )[0];
     if (collapseButton) {
       collapseButton.setAttribute(
         'style',
         `background:
       ${
-        value ? 'var(--jp-icon-caret-right)' : 'var(--jp-icon-caret-down)'
+        value ? 'var(--jp-icon-caret-right-large-emph)' : 'var(--jp-icon-caret-down-large)'
       } no-repeat center`
       );
     }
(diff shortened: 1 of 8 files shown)


fcollonval (changes requested):
Thanks for starting this @schmidi314 

> For this first review, I will focus on the implementation logic.

We can simplify the logic as the setting is part of the notebook config by toggling a class on the notebook panel instead of propagating the value down to all markdown cells (also because there is no need to set that flag per markdown cells). You should do that in the `StaticNotebook` constructor and in [_updateNotebookConfig](https://github.com/schmidi314/jupyterlab/blob/b5832321a7a383cdcdfbef59ba3b85dbd5e958a5/packages/notebook/src/widget.ts#L782) instead of looping on all cells.  
The name of the class should be `jp-mod-...` to highlight that it depends on the state/settings.

Regarding the icon, I would not use a different design for homogeneity. But it would indeed be good if the icon could be bigger.
Could you look how it looks for deeper title (like heading level 5 or 6) ? We don't want the icon to be bigger than the title ;-)

Let me know if you need pointers.

fcollonval:
To scale the background image, you can find more information on https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Resizing_background_images

schmidi314:
Thanks a lot @fcollonval for the review. I learned a lot. And, phew, took me a while to figure out what to do, but I think I know now what you meant. Well, it was high time to start learning CSS anyway :)

* I think it is sufficient to toggle the class in `_updateNotebookConfig` only since this method is called on construction anyway. Also, the `jp-mod-scrollPastEnd` is only set in in `_updateNotebookConfig()`. So, either both or none class toggles should go to the constructor, right?
* Admittedly, this method of showing/unshowing buttons is a lot leaner than what I had written. But it is also a lot harder to grasp for beginners like me that are not so familiar with CSS. I would have had a hard time to understand why the `maybeCreateOrUpdateExpandButton` creates a button that doesn't show up in some cases. So: I'd suggest to add a comment in `maybeCreateOrUpdateExpandButton` 
* In the present pull request I also cleaned up (IMHO) the `maybeCreateOrUpdateExpandButton` method a bit. Basically it was about renaming variables, removing duplicates and writing the button text to the right property. Should I redo this or is the code better as it is? I'll add a single commit for this and then you can decide if you want to keep it.

Since only the first commit of the present PR will survive, I'd like to close this one and add a new one. Any objections?

fcollonval:
> Thanks a lot @fcollonval for the review. I learned a lot. And, phew, took me a while to figure out what to do, but I think I know now what you meant. Well, it was high time to start learning CSS anyway :)

Don't hesitate to bring question on the PR or on Gitter if comments are unclear or to concise. We are always eager to help new contributors - and thank you for that.

>     * I think it is sufficient to toggle the class in `_updateNotebookConfig` only since this method is called on construction anyway. Also, the `jp-mod-scrollPastEnd` is only set in in `_updateNotebookConfig()`. So, either both or none class toggles should go to the constructor, right?

Indeed both should be toggle to in the constructor. So calling `_updateNotebookConfig` there sounds good.

>     * Admittedly, this method of showing/unshowing buttons is a lot leaner than what I had written. But it is also a lot harder to grasp for beginners like me that are not so familiar with CSS. I would have had a hard time to understand why the `maybeCreateOrUpdateExpandButton` creates a button that doesn't show up in some cases. So: I'd suggest to add a comment in `maybeCreateOrUpdateExpandButton`

Go ahead, better documentation is always good.
FYI this kind of CSS tricks is common in Web UI. A good practice when figure out what happens is to introspect the DOM and the CSS rules applied on a particular element. Then to search in the code where is the style set (usually through setting a class or attributes typically).
 
>     * In the present pull request I also cleaned up (IMHO) the `maybeCreateOrUpdateExpandButton` method a bit. Basically it was about renaming variables, removing duplicates and writing the button text to the right property. Should I redo this or is the code better as it is? I'll add a single commit for this and then you can decide if you want to keep it.

Better code wording is better documentation; so feel free to do so (you should only be careful not to change the API).

> Since only the first commit of the present PR will survive, I'd like to close this one and add a new one. Any objections?

:+1:

Issue #11442, opened 2021-11-11
Heading collapse not working in second notebook view


schmidi314:
This is my first bug report. Tried my best to find something related, but I think this is a new one.

But first of all a big thanks to the developers. I've been a huge fan of jupyter notebook and now jupyter lab is even better. Hopefully I can help with fixing some bugs or contributing some features in the future, but I'm still reading through your codebase. So for now:

## Description

Collapsing headings works well for a single notebook view. If multiple views for a notebook are opened, headings that have existed prior to opening the new view cannot be collapsed in the new view but only in the primary view. (Note: I collapse by clicking the small triangle left of the heading)

[I believe this should be tagged with TOC, but I don't get how I can add this label..??]

## Reproduce

1. Create a fresh Notebook with a heading (level 1) cell and at least one content cell. Do this __before__ you open a second view for the notebook.
2. Open a second view (right click on the tab --> "New View for Notebook")
3. Try to collapse the heading. Works in the primary view, doesn't work in the new view.

Note: If you create the heading while the second view is already open, collapsing works as expected.

## Expected behavior

One should be able to collapse and expand headings in all views. 

In my opinion: There should be no synchronization of collapse state between the different views. But maybe this depends on personal taste and might be a good thing for an option.

## Context

- Operating System and version: macOS 11.6 (but also observed on Linux and Windows)
- Browser and version: Chrome Version 95.0.4638.69 (Official Build) (x86_64)
- JupyterLab version: 3.2.3

schmidi314:
I think I have an idea what the problem might be.

https://github.com/jupyterlab/jupyterlab/blob/ee1436736daf6f856d99ab125cc2b83592f6d196/packages/notebook-extension/src/index.ts#L1516-L1530

The `toggleCollapsedSignal` of the __widget__ `MarkdownCell` is connected to the right handler (l.1526) whenever the cell lists in the __model__ changes. For the first view of the notebook, there is only one widget "structure" for the model and the signal connecting probably happens as the cell list is populated initially (during loading). But for a second view, the **model** is already populated and remains unchanged, while new **widgets** for this model are generated. However, these widgets' signals aren't connected properly.

So: The collapse signal should be connected whenever a new **widget** is created.

schmidi314:
![second_view_heading_collapse](https://user-images.githubusercontent.com/74381262/142625044-cf62cfbe-07ea-461b-aa9f-334da31d49ba.gif)

[...]

PR #11505, opened 2021-11-19
Connecting toggleCollapsedSignal to handler right at creation of Mark…

schmidi314:
…downCell.

## References

#11442 

## Code changes

Connected the toggleCollapsedSignal of MarkdownCell to the handler in the _createMarkdownCell method.


packages/notebook/src/widget.ts
```diff
@@ -647,6 +647,14 @@ export class StaticNotebook extends Widget {
     const cell = this.contentFactory.createMarkdownCell(options, this);
     cell.syncCollapse = true;
     cell.syncEditable = true;
+    // Connect collapsed signal for each markdown cell widget
+    if(this instanceof Notebook) {
```
fcollonval, line 651: Could you elaborate on the need for that test? As the method is `private` it should not be required.

packages/notebook-extension/src/index.ts
```diff
@@ -1518,6 +1518,7 @@ function addCommands(
           list: IObservableUndoableList<ICellModel>,
           args: IObservableList.IChangedArgs<ICellModel>
         ) => {
+          /* probably this can go away
```
fcollonval, line 1521: Yes please delete it.

fcollonval (changes requested):
Thanks @schmidi314 I get only two small comments on code style.

packages/notebook/src/widget.ts
schmidi314: Hm, seems this is a remnant of me trying to cast in type script. Sorry, new to this language. I'll fix that.

packages/notebook-extension/src/index.ts
schmidi314: Done.

schmidi314:
@meeseeksdev tag bug

krassowski:
@schmidi314 I think you accidentally uploaded `highlight-latex-jupyterlab-linux.png` instead of `do-not-highlight-not-latex-jupyterlab-linux.png`.

krassowski:
Also, there are some pixel changes in the editor positioning in these tests - https://github.com/jupyterlab/jupyterlab/pull/11507 will fix it once ready, so no need to worry about it too much.

schmidi314:
Ah, damn, sorry. Thanks for the hint.

OK, good to know about #11507. But let me fix it anyway, for practicing reasons :)

fcollonval (approved):
Thanks @schmidi314

fcollonval:

PR #11508, opened 2021-11-19
Attempt removing use of flexbox in the notebook DOM

SylvainCorlay:
This will also make it possible to make use of `break-inside: avoid` CSS media properties for nbconvert webpdf exports.

Besides, it may also improce rendering performances of notebooks.

### DIFF
--- a/packages/cells/style/inputarea.css
+++ b/packages/cells/style/inputarea.css
@@ -9,18 +9,19 @@
 
 /* All input areas */
 .jp-InputArea {
-  display: flex;
-  flex-direction: row;
+  display: table;
   overflow: hidden;
-}
-
-body[data-format='mobile'] .jp-InputArea {
-  flex-direction: column;
 }
 
 .jp-InputArea-editor {
-  flex: 1 1 auto;
+  display: table-cell;
+  width: 100%;
   overflow: hidden;
+  vertical-align: top;
+}
+
+body[data-format='mobile'] .jp-InputArea-editor {
+  display: table-row;
 }
 
 .jp-InputArea-editor {
@@ -35,7 +36,9 @@ body[data-format='mobile'] .jp-InputArea-editor {
 }
 
 .jp-InputPrompt {
-  flex: 0 0 var(--jp-cell-prompt-width);
+  display: table-cell;
+  vertical-align: top;
+  min-width: var(--jp-cell-prompt-width);
   color: var(--jp-cell-inprompt-font-color);
   font-family: var(--jp-cell-prompt-font-family);
   padding: var(--jp-code-padding);
@@ -58,6 +61,6 @@ body[data-format='mobile'] .jp-InputArea-editor {
 }
 
 body[data-format='mobile'] .jp-InputPrompt {
-  flex: 0 0 auto;
+  display: table-row;
   text-align: left;
 }
--- a/packages/cells/style/widget.css
+++ b/packages/cells/style/widget.css
@@ -76,11 +76,9 @@
 }
 
 .jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-prompt {
-  flex: 0 0
-    calc(
-      var(--jp-cell-prompt-width) -
-        var(--jp-private-cell-scrolling-output-offset)
-    );
+  min-width: calc(
+    var(--jp-cell-prompt-width) - var(--jp-private-cell-scrolling-output-offset)
+  );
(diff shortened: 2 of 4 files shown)


fcollonval (commented):
please run benchmark

jtpio:
Thanks @SylvainCorlay for opening the PR.

Some of the failing UI tests look relevant (caret not vertically centered):

image

fcollonval:
This impact the _mobile_ styles. It would be good to add UI tests for that to check we are doing ok.

jtpio:
> This impact the _mobile_ styles. It would be good to add UI tests for that to check we are doing ok.

Yes, it seems there is already an issue with 3.2.4 and the mobile styling: https://github.com/jupyterlab/jupyterlab/issues/11516

SylvainCorlay:
The test failure is due to the fact that I use the "inline-box" display to mimick the horizontal flex box instead of an horizontal table.

SylvainCorlay:
I made the following diagrams of the DOM structure (happy to contribute them to the docs).

Screenshot from 2021-11-22 11-14-00

Currently, horizontal flexboxes are use for
 - inputWrapper
 - outputWrapper
 - inputArea
 - outputArea-child

However, in the nbconvert case, there is not collapser, so the only flexboxes are inputArea and outputArea-child, so we could start there.

SylvainCorlay:
please run benchmark

SylvainCorlay (commented):
please run benchmark

SylvainCorlay (commented):
please run benchmark

SylvainCorlay (commented):
please run benchmark

SylvainCorlay:
Visual regression tests are passing.

fcollonval (approved):
Thanks @SylvainCorlay

jtpio:
> I made the following diagrams of the DOM structure (happy to contribute them to the docs).

It could be interesting to have that diagram in this section of the docs for example: https://jupyterlab.readthedocs.io/en/stable/extension/notebook.html

Although it might become outdated at some point. In that case we could mention "as of JupyterLab 3.2.x" or "4.0.0" depending on where this lands.

gabalafou:
Hi, hope you don't mind me dropping in on this thread out of the blue. 

I leave the following remarks only in a spirit of spreading knowledge, raising awareness, and asking questions; I am not requesting any changes or making recommendations.

I heard about this PR in today's JupyterLab team meeting, and it raised a red flag in my mind, because as a general rule of thumb, the CSS `display` property affects HTML semantics, which in turn is important for accessibility. However, I did a little digging and luckily, in the specific case of `display: table`, I don't think the general rule of thumb applies (in most cases... there are always caveats in this line of work). More on the topic at Adrian Roselli's blog for anyone who wants the nitty-gritty on [tables, CSS display properties, and ARIA.](https://adrianroselli.com/2018/02/tables-css-display-properties-and-aria.html) 

tl;dr-- display:table is probably not an accessibility issue in this PR but please be careful when applying display properties in a non-standard way. 

That said, I can't help but wonder if display:table is an anti-pattern, though perhaps a desirable and/or necessary one for the `break-inside: avoid` declaration. I also wonder if an @media rule would be an appropriate solution here, though I'm not sure I would prefer it, given that I think it is extremely desirable to reduce markup and style variants as much as possible, but I thought I should at least mention it. 

:)

SylvainCorlay:
> Hi, hope you don't mind me dropping in on this thread out of the blue.

Quite the contrary. Many thanks for your feedback.

> That said, I can't help but wonder if display:table is an anti-pattern, though perhaps a desirable and/or necessary one for the break-inside: avoid declaration. I also wonder if an @media rule would be an appropriate solution here, though I'm not sure I would prefer it, given that I think it is extremely desirable to reduce markup and style variants as much as possible, but I thought I should at least mention it.

I have the exact same thought. There is a companion PR in nbconvert https://github.com/jupyter/nbconvert/pull/1679 that overrides the CSS but I would rather have as little as possible in the `@media` rule. The reason is that I fear it may not be as thoroughly tested as the general case and will be more likely to break. More generally, maintaining two versions will be more difficult.

krassowski:
This PR has introduced two significant regressions:
- https://github.com/jupyterlab/jupyterlab/issues/14564
- https://github.com/jupyterlab/jupyterlab/issues/14625

SylvainCorlay:
I am working on a fix for #14564.

PR #11510, opened 2021-11-20
Update `sanitize-html` pin to `3.5.3`

### LINKED ISSUE #11473: HTML inline style not shown in trusted notebook's markdown cell
## Reproduce
- Open jupyterlab v3.2.3, create a new notebook and add a markdown cell:
```html
<span style="font-size:42px">Irrelavant text.</span>
```
- The style does not apply
https://stackoverflow.com/questions/68815487/html-inline-style-not-applying-in-jupyter-notebook-cells-anymore
## Expected behavior
html should be rendered correctly just like in latest jupyter notebook v6.4.5

krassowski:
Updates `sanitize-html` from `3.3.3` to `3.5.3`.

## References

Fixes #11473.

## Code changes

https://github.com/apostrophecms/sanitize-html/compare/2.3.3...2.5.3

<details>
<summary>Upstream changelog</summary>

```
## 2.5.3 (2021-11-02):

- Fixed bug introduced by klona 2.0.5, by removing klona entirely.

## 2.5.2 (2021-10-13):

- Nullish HTML input now returns an empty string. Nullish value may be explicit `null`, `undefined` or implicit `undefined` when value is not provided. Thanks to Artem Kostiuk for the contribution.
- Documented that all text content is escaped. Thanks to Siddharth Singh.

## 2.5.1 (2021-09-14):
- The `allowedScriptHostnames` and `allowedScriptDomains` options now implicitly purge the inline content of all script tags, not just those with `src` attributes. This behavior was already strongly implied by the fact that they purged it in the case where a `src` attribute was actually present, and is necessary for the feature to provide any real security. Thanks to Grigorii Duca for pointing out the issue.

## 2.5.0 (2021-09-08):

- New `allowedScriptHostnames` option, it enables you to specify which hostnames are allowed in a script tag.
- New `allowedScriptDomains` option, it enables you to specify which domains are allowed in a script tag. Thank you to [Yorick Girard](https://github.com/yorickgirard) for this and the `allowedScriptHostnames` contribution.
- Updates whitelist to allowlist.

## 2.4.0 (2021-05-19):
- Added support for class names with wildcards in `allowedClasses`. Thanks to [zhangbenber](https://github.com/zhangbenber) for the contribution.
```

</details>

## User-facing changes

Styles will work correctly in markdown again.

None.

### DIFF
--- a/packages/apputils/package.json
+++ b/packages/apputils/package.json
@@ -62,7 +62,7 @@
     "@lumino/widgets": "^1.28.0",
     "@types/react": "^17.0.0",
     "react": "^17.0.1",
-    "sanitize-html": "~2.3.3",
+    "sanitize-html": "~2.5.3",
     "url": "^0.11.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -11398,11 +11398,6 @@ kleur@^3.0.3:
   resolved "https://registry.yarnpkg.com/kleur/-/kleur-3.0.3.tgz#a79c9ecc86ee1ce3fa6206d1216c501f147fc07e"
   integrity sha512-eTIzlVOSUR+JxdDFepEYcBMtZ9Qqdef+rnzWdRZuMbOywu5tO2w2N7rqjoANZ5k9vywhL6Br1VRjUIgTQx4E8w==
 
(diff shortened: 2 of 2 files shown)


fcollonval:

Issue #11406, opened 2021-11-05
[RTC] Collaborator cursor/selection not displayed for markdown cells


fcollonval:
## Description

In collaborative mode, cursor and selection of users are displayed in editors. But when it comes to markdown cells, the other user cursor/selection is not displayed if the cells is rendered.

![rtc_cursor_md](https://user-images.githubusercontent.com/8435071/140488279-8defa1cb-ae3a-4a90-b3a4-812b268374fb.gif)

## Reproduce

1. Launcher JupyterLab with `--collaborative` editing
2. Open the same notebook in two browsers
3. Edit a markdown cell in one browser
4. The cursor/selection is not appearing in the other notebook.

## Expected behavior

There are some sense of edition of markdown cells by other users (maybe the markdown cells need to be switched automatically to edition mode when another user is editing it).

## Context

- Operating System and version: 
- Browser and version: 
- JupyterLab version: 

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

hbcarlos:
Thanks for opening this issue, Frédéric!
Adding the cursor to the rendered cell might be a little bit difficult, but maybe we can force the cell to switch to edit mode when a collaborator is editing it.

martinRenou:
Assigning myself, if that's fine :)

martinRenou:
As suggested by @hbcarlos, what we want to do is to force the rendered markdown cell back to the edit mode if a collaborator is currently editing it. I believe that makes a lot of sense.

Though currently, we only have the information of **what the other collaborators's cursor position is**, we don't really know what they are doing exactly. We don't want to force the rendered cell back to edit mode just because a collaborator's mouse went over it.

We should probably not try to fix this issue before we have more information about what other collaborators are doing (editing cell number #, working on notebook #.ipynb, ...), unless someone has an idea on how to fix it properly with the very limited knowledge we have.

PR #11518, opened 2021-11-22
RTC: Sync Notebook view information (activeCell, current mode)

martinRenou:
## References

Fix https://github.com/jupyterlab/jupyterlab/issues/11406

## Code changes

With RTC, we sync the Notebook model (its actual content), but some things related to the view are not synced. This leads to the issue mentioned above.

This PR adds a new event `uiChanged` to the YJS shared model that lets us implement UI related sharing.

## User-facing changes

markdown

None?

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -16,6 +16,7 @@ import { IChangedArgs } from '@jupyterlab/coreutils';
 import * as nbformat from '@jupyterlab/nbformat';
 import { IObservableList, IObservableMap } from '@jupyterlab/observables';
 import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
+import { NotebookUIChange, YNotebook } from '@jupyterlab/shared-models';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { ArrayExt, each, findIndex } from '@lumino/algorithm';
 import { MimeData, ReadonlyPartialJSONValue } from '@lumino/coreutils';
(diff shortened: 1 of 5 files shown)


packages/notebook/src/widget.ts
```diff
@@ -2387,6 +2401,20 @@ export class Notebook extends StaticNotebook {
     }
   }
 
+  /**
+   * Handle UI change from other collaborators
+   */
+  private _onUiChanged(sharedModel: YNotebook, change: NotebookUIChange) {
+    // Force edit mode for the markdown cells on which other collaborators are working
```
martinRenou, line 2408: Wondering if we should have a finer solution, where we have per-cell awareness on the `rendered` state.

hbcarlos: Yes, it is probably better to have a per-cell awareness.
And regarding the new signal, I think would be better to not add a new signal but instead add a new property stateChanged to the existing signal. I just changed my mind because I saw that the notebook is using stateChanged to notify that the mode changed https://github.com/jupyterlab/jupyterlab/blob/bd07b0779760e63c2fc147a5544db78e8b9a7282/packages/notebook/src/widget.ts#L1147. I think we should follow this pattern on the shared cell model.

martinRenou: Ok :) Let's do that. Thanks for the review!

martinRenou:
Closing this PR. We won't add another signal, instead it might be nicer to add the `mode` property to the `YNotebook`'s `ystate`?

PR #11519, opened 2021-11-22
Feature/optional hidden cells button (v2)

schmidi314:
# References
Second try for issue #11490 

## Code changes
* Adding option `showHiddenCellsButton` to Notebook Settings.
* Makes visibility of "hidden cells" and "heading collapser" buttons dependent on this option.
* Increases size of heading collapser/expander buttons (arrows left of the heading)

## User-facing changes
* With the default `showHiddenCellsButton=true` option, the only noticeable difference are somewhat larger arrow head buttons (left of the headings).
* With `showHiddenCellsButton=false`, the arrow heads are always displayed (not only on hover) and the hidden cells button won't be displayed.

optional_hidden_cells_button


schmidi314:
@meeseeksdev tag enhancement

fcollonval (commented):
Thanks @schmidi314 this is a clean nice PR :bow: 

As you are improving the collapsing buttton, could you include in this PR two more enhancements:
- Make the collapse button always visible for the active cell (displaying elements only on hover is bad for user to discover them)
- Add translation for the hidden button label `N cells hidden`

For the last point, you need to store the `trans` variable (defined [there](https://github.com/jupyterlab/jupyterlab/blob/24cf571f0433d0e4087b47d93c1aa7471698b86e/packages/cells/src/widget.ts#L1455)) as private class attribute. Then you need to set the text like this:

```
  buttonText = trans._n('%1 cell hidden', '%1 cells hidden', this._numberChildNodes);
```

schmidi314:
Thanks @fcollonval 

* Regarding the collapse button visibility: I'd never have made the button invisible in the first place. I only tried to reduce the user-facing changes to a minimum for the default setting. At least in my workflow it practically never happens that I collapse a heading that happens to be an active cell. (Actually, I used to collapse/expand with the arrow keys in jupyter notebook, which doesn't work yet in lab, but that's a story for another PR). So, would you be OK with the arrows *always* being visible? Or even with the default `showHiddenCellsButton` being false?

* trans... done, but why in the world is
image
prettier than 
image
? :)

And a further questions, regarding testing:
* @krassowski mentioned a while ago that I should add some galata tests. I haven't really tried this yet, but I'd probably do something along the lines of the gif I've put in the PR description. Any thoughts before I start?

fcollonval:
> Regarding the collapse button visibility: I'd never have made the button invisible in the first place.

Sorry about the wording, I know you did not change the current behavior (what is very good usually). But as the current behavior is not great, I would like to add that tweak (of displaying the collapse button for the active cell) in this PR.
My statement of saying displaying thing only on hover is only to justify my request.

> would you be OK with the arrows always being visible?

I'll bring the subject at today's Lab team meeting; it will be great if [you can join](https://hackmd.io/Y7fBMQPSQ1C08SDGI-fwtg).

> trans... done, but why in the world is...

Most of code formatters have a rule of max. number of characters per lines. That number is usually between 70 and 100 with the idea that everything should be rendered on a terminal (because lots of dev' are still using tools like vim or emacs). Hence that long line gets split into smaller ones.

> UI testing

If you can add some UI tests, this will be awesome. You can look in `galata/test/jupyterlab/notebook-run.test.ts` for an example of opening a notebook. Then the trick is to figure out how to select the button, to emulate click on it. You can read on that in the [Playwright doc]() that we use.

To deal with the settings, you can change the mocked settings like in that example: https://github.com/jupyterlab/jupyterlab/blob/202a7a6349166e417da812d92d0f0afac823178f/galata/test/galata/fixture.spec.ts#L25

fcollonval:
> would you be OK with the arrows always being visible? Or even with the default showHiddenCellsButton being false

To move forward on, let's keep `showHiddenCellsButton` as `true` by default. But it will be nice to add a CSS rule to always display the collapser for the active cell.

schmidi314:
@fcollonval thanks for your galata advice. I'm starting to work on it. And I hope I haven't given rise to the impression that I've been somehow unhappy with your reply. I really appreciate your time and help a lot. Maybe I should add more smileys :) Ah and sorry for not joining the team meeting. I'd love to join at some point, but these days I'm quite busy with work... not the best point in time to start jupyterlab development, if I think about it :)

Anyway, I have changed the `.jp-MarkdownCell:hover` class selector to `:is(.jp-MarkdownCell:hover, jp-mod-active)`. In principle I like that one better than duplicating the rule. But I read that the :is selector is sort of new. I tested it on Firefox and Chrome (Linux). Let me know if I should better expand it again.

fcollonval (approved):

fcollonval:
Thanks a lot for this PR and taking the time to add UI test :rocket: 

> thanks for your galata advice. I'm starting to work on it. And I hope I haven't given rise to the impression that I've been somehow unhappy with your reply. I really appreciate your time and help a lot. Maybe I should add more smileys :) 

:smile: That's the problem of writing - feeling are harder to guess.

> Ah and sorry for not joining the team meeting. I'd love to join at some point, but these days I'm quite busy with work... not the best point in time to start jupyterlab development, if I think about it :)

Don't worry, we really appreciate you are taking the time to contribute and improve JLab; no pressure or rules to attend it. 

FYI There is also a monthly [Jupyter community meeting](https://discourse.jupyter.org/t/jupyter-community-calls/668); it targets the broader Jupyter community and it is a good place to show new features or extensions.

> Anyway, I have changed the `.jp-MarkdownCell:hover` class selector to `:is(.jp-MarkdownCell:hover, jp-mod-active)`. In principle I like that one better than duplicating the rule. But I read that the :is selector is sort of new. I tested it on Firefox and Chrome (Linux). Let me know if I should better expand it again.

It looks well supported according [canIuse](https://caniuse.com/?search=%3Ais()). So let's keep it like this.

PR #11525, opened 2021-11-23
Sync dirty property between clients

hbcarlos:
Solves a minor bug with the dirty property when saving the document.

## References
I think we never opened an issue, and I can not find the issue/PR where this bug was detected, but @jasongrout detected a bug in the dirty property while working on something else.

## Code changes
Sync the dirty property using the SharedDocument and improve the initialization.

## User-facing changes
Now when someone saves the document, this change is notified to every client.

Not sure. Added a new property in SharedDocument and removed a private one from DocumentWidget.

[@fcollonval] This is ok as you added something to the public API (no changes, nor deletion)

### DIFF
--- a/packages/docregistry/src/default.ts
+++ b/packages/docregistry/src/default.ts
@@ -29,6 +29,9 @@ export class DocumentModel
     const filemodel = new models.YFile() as models.ISharedFile;
     this.switchSharedModel(filemodel, true);
     this.value.changed.connect(this.triggerContentChange, this);
+
+    (this.sharedModel as models.YFile).dirty = false;
+    this.sharedModel.changed.connect(this._onStateChanged, this);
   }
 
   /**
@@ -49,15 +52,14 @@ export class DocumentModel
    * The dirty state of the document.
    */
   get dirty(): boolean {
-    return this._dirty;
+    return this.sharedModel.dirty;
   }
   set dirty(newValue: boolean) {
-    if (newValue === this._dirty) {
+    const oldValue = this.sharedModel.dirty;
+    if (newValue === oldValue) {
       return;
     }
-    const oldValue = this._dirty;
-    this._dirty = newValue;
-    this.triggerStateChange({ name: 'dirty', oldValue, newValue });
+    (this.sharedModel as models.YFile).dirty = newValue;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


fcollonval (approved):
Thanks @hbcarlos

krassowski:
@hbcarlos could you add a gif demo? I think it did not work for me on binder yesterday.

hbcarlos:
Sure, I recorded a screencast from binder:

https://user-images.githubusercontent.com/26092748/143221211-ba5c4f5c-91ea-401f-a7b8-3c52870ad155.mp4

1 - When saving the document from a client, it changes for every client.
2 - When opening a file the dirty property stays false.
3 - When creating a new file the dirty property is true because the metadata changes (kernel-related stuff).

krassowski:
Sorry about the delay here. So it does not seem to work for me on Binder for a notebook which was created before sharing the link:

does-not-work-with-outputs


Nor for a new notebook - it only updates indicator for one of the clients:

does-not-update-one-of-the-states

The behaviour does not appear consistent - sometimes it updates indicator on one side sometimes on the other, and once it even worked for both - but in any case it looks like something is wrong here.

krassowski:
Please ignore the repro above it was tainted by the use of unconstrainted `git log` (see edit above - I should have used `git log | head`). The correct repro is below:

correct-repro

Sorry about the bad recording quality. Maybe the state is not synced properly when opening the notebook? It looks like the [1] prompt is missing in the newly opened notebook.

hbcarlos:
Wow! Thanks for caching that one @krassowski!
I'll take a look, to see what is happening.

hbcarlos:
The problem was that the dirty className was added multiple times to the title and removed once.
I added multiple checks to make sure we add it only once.

krassowski (approved):
Looks good on binder now, thanks!

fcollonval:

packages/docregistry/src/default.ts
```diff
@@ -29,6 +29,9 @@ export class DocumentModel
     const filemodel = new models.YFile() as models.ISharedFile;
     this.switchSharedModel(filemodel, true);
     this.value.changed.connect(this.triggerContentChange, this);
+
+    (this.sharedModel as models.YFile).dirty = false;
```
dmonad, line 33: @hbcarlos We should avoid parsing to the concrete `Y*` implementations because that is non-ideomatic. 

In general, when you work with an abstraction, you shouldn't cast to the concrete implementation just to manipulate properties of the concrete implementation because that will eventually lead to issues.

In this case, we explicitly don't want to initialize shared properties whenever a user joins a session. Properties on the concrete Yjs implementation must only be initialized on the `create` and `createStandalone` methods.

PR #11526, opened 2021-11-23
Support file type extension with upper case

fcollonval:
## References

I hit that issue when adding file type for `.Rmd` in https://github.com/mwouts/jupytext/pull/877

## Code changes

Compare path extension with file type extension after converting both to lower case

## User-facing changes

Some files may be shown with a different type.

N/A

### DIFF
--- a/packages/docregistry/src/registry.ts
+++ b/packages/docregistry/src/registry.ts
@@ -1431,7 +1431,7 @@ export namespace DocumentRegistry {
         name: 'r',
         displayName: trans.__('R File'),
         mimeTypes: ['text/x-rsrc'],
-        extensions: ['.r'],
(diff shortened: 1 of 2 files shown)


krassowski:
Should we also change the default extension for R files to `.R` from current `.r`? This is what all the R style guidelines recommend.

fcollonval:
> Should we also change the default extension for R files to `.R` from current `.r`? This is what all the R style guidelines recommend.

Thanks for the suggestion, I updated the PR accordingly

jtpio (approved):
Looks good, thanks!

jtpio:
The CI failure is not related and should be fixed now.

jtpio:

PR #11529, opened 2021-11-23
Cell YModel: Fix setAttachment method

martinRenou:
## References

None

This looks like an issue. Though, as far as I can tell, this method isn't used anywhere (yet?).

## Code changes

Remove the `attachment` value from the cell `ymodel` if it's null, instead of adding it.

## User-facing changes

None

None

### DIFF
--- a/packages/shared-models/src/ymodels.ts
+++ b/packages/shared-models/src/ymodels.ts
@@ -731,9 +731,9 @@ export class YBaseCell<Metadata extends models.ISharedBaseCellMetadata>
   public setAttachments(attachments: nbformat.IAttachments | undefined): void {
     this.transact(() => {
       if (attachments == null) {
-        this.ymodel.set('attachments', attachments);
-      } else {
         this.ymodel.delete('attachments');
+      } else {
+        this.ymodel.set('attachments', attachments);
       }
     });
   }


hbcarlos:
Thanks, @martinRenou!
Yes, this is an issue, but this method is not used anywhere.
The test failing is also failing on master (at least locally). It is probably related to the service test failing on every PR.

dmonad:
Thanks so much for finding this @martinRenou!

The "Visual Regression Test" might be failing because the correct implementation brought up a new issue. Can we rerun the test?

@hbcarlos It would be interesting to find out what the `attachment` field is used for and why we are not syncing it using the shared-model.

martinRenou:
I suppose the attachment field represents the files you can embed in a Markdown cell by doing a drag and drop: see https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Working%20With%20Markdown%20Cells.html#Markdown-attachments

I guess it makes sense to synchronise it, if YJS can handle such file?

hbcarlos:
It is a JSON object, see https://github.com/jupyterlab/jupyterlab/blob/202a7a6349166e417da812d92d0f0afac823178f/packages/nbformat/src/index.ts#L62-L79

We should sync them through the shared models.

dmonad:
Yes, we should definitely sync files. Yjs should be able to sync files as long as they are smaller than 4GB.

fcollonval (approved):
Thanks @martinRenou for catching this one.

_attachment_ primarily (only use) is indeed to drop an markdown media within the cells (like you can do in GitHub comments). So that information needs to be stored somewhere :wink: 

Regarding  the UI test, this is a flaky test not link to this PR

dmonad (approved):

jtpio:
The UI tests check passed after restarting.

`js-services` failure is tracked in #11548 

Thanks!

jtpio:

PR #11530, opened 2021-11-23
Trigger TOC update on app started

jtpio:
## References

Attempts at making the TOC extension more reusable in other lab-based apps such as RetroLab (https://github.com/jupyterlab/retrolab/pull/275).

## Code changes

Rework TOC plugin logic.

## User-facing changes

None for lab users

None

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -51,12 +51,18 @@ namespace CommandIDs {
   export const showPanel = 'toc:show-panel';
 }
 
+//
+/**
+ * The id for the Table of Contents plugin.
+ * TODO: rename to `@jupyterlab/toc-extension:registry` for consistency.
+ */
+const PLUGIN_ID = '@jupyterlab/toc:plugin';
+
 /**
  * A plugin to provide a registry for the table of contents.
  */
 const registry: JupyterFrontEndPlugin<ITableOfContentsRegistry> = {
-  // TODO: rename to `@jupyterlab/toc-extension:registry` for consistency.
-  id: '@jupyterlab/toc:plugin',
+  id: PLUGIN_ID,
   autoStart: true,
   provides: ITableOfContentsRegistry,
   requires: [IDocumentManager, IRenderMimeRegistry, ITranslator],
@@ -150,13 +156,14 @@ const registry: JupyterFrontEndPlugin<ITableOfContentsRegistry> = {
 
     if (restorer) {
       // Add the ToC widget to the application restorer:
-      restorer.add(toc, '@jupyterlab/toc:plugin');
+      restorer.add(toc, PLUGIN_ID);
     }
 
     // Attempt to load plugin settings:
     let settings: ISettingRegistry.ISettings | undefined;
     if (settingRegistry) {
       try {
+        // TODO: use PLUGIN_ID
         settings = await settingRegistry.load(
           '@jupyterlab/toc-extension:plugin'
         );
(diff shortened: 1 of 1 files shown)


packages/toc-extension/src/index.ts
```diff
@@ -42,212 +52,196 @@ namespace CommandIDs {
...
 };
 
 /**
- * Exports.
+ * Export the plugins as default.
  */
-export default extension;
+const plugins: JupyterFrontEndPlugin<any>[] = [registry];
+
+export default plugins;
```
fcollonval, line 247: Minor optional suggestion, we don't need to export an array.

```suggestion
export default registry;
```

fcollonval (commented):
LGTM I have a minor suggestion.

packages/toc-extension/src/index.ts
jtpio: Sure. I was looking into splitting the main plugin into multiple ones at first, but happy to switch to a more concise export.

packages/toc-extension/src/index.ts
```diff
@@ -42,212 +52,194 @@ namespace CommandIDs {
...
-  optional: [
-    IEditorTracker,
-    ILayoutRestorer,
-    ILabShell,
-    IMarkdownViewerTracker,
-    INotebookTracker,
-    ISettingRegistry
-  ],
-  activate: activateTOC
+    app.started.then(updateCurrent);
```
jtpio, line 236: Need to check the order of events (widget added to the main area, app started) is as expected, so this line is relevant.

jtpio: This might work in the case of RetroLab which should add the widget on startup to the main area via the document opener.

However this might not be robust enough in more complex scenarios where widgets are added a later time. Maybe this toc plugin should then be reworked to allow for more flexibility like the file browser widget.

jtpio: Checking in RetroLab with a local build of https://github.com/jupyterlab/retrolab/pull/275:

image

This change is unfortunately not enough, and being able to listen to `currentChanged` seems to be required.

One way forward would be to decouple the widget and registry to allow for better reuse of the plugin(s).

krassowski: While at it - I was thinking whether LSP should be able populate the TOC widget with document outline in script files, or whether it should be a separate widget. I wonder if the decoupling you mention could help that use case.

jtpio: Probably yes. For LSP that could mean adding a new generator similar to the existing ones?

https://github.com/jupyterlab/jupyterlab/blob/51d8f101f7ab5df7551dcd32ec4cb47c10f2fb4d/packages/toc-extension/src/index.ts#L155-L197

jweill-aws:
I was taking a look at https://github.com/jupyterlab/retrolab/pull/275 earlier. Is this item still a work in progress? Is it still intended for release in 4.0? Just want to make sure I have the most current info about this. Thanks.

jtpio:
@jweill-aws this PR was left open for visibility, but I am not actively working on it. So happy to close it for now.

We can track the changes that need to happen for 4.0 in a separate issue.

jweill-aws:
I'm OK with leaving this open for 4.0 work. Not sure that we need to open a separate issue.

jtpio:
Closing in favor of https://github.com/jupyterlab/jupyterlab/pull/12374, where the table of contents refactor is now happening.

PR #11538, opened 2021-11-24
RTC shared model: Rename createCellFromType into createCellFromModel

martinRenou:
## References

None

## Code changes

Rename `createCellFromType` function into `createCellFromModel`. This function creates a new YBaseCell given the ymodel, so this new name makes more sense

## User-facing changes

None

It's an exported function, so this could be backward incompatible

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -38,7 +38,7 @@ bumped their major version (following semver convention):
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    The ``NotebookPanel._onSave`` method is now ``private``.
 - ``@jupyterlab/shared-models`` from 3.x to 4.x
-  The ``createCellFromType`` functoin is now named ``createCellModelFromSharedType``
+  The ``createCellFromType`` function has been renamed to ``createCellModelFromSharedType``
 
 JupyterLab 3.0 to 3.1
 ---------------------
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -37,6 +37,8 @@ bumped their major version (following semver convention):
    The ``CodeEditor.ICoordinate`` interface now extends ``DOMRectReadOnly`` instead of ``JSONObject, ClientRect``.
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    The ``NotebookPanel._onSave`` method is now ``private``.
+- ``@jupyterlab/shared-models`` from 3.x to 4.x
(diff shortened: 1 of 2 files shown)


dmonad:
When we talk about shared data structures in Yjs, we usually call them *shared types*. The idea of a *model* is something that comes from the Jupyter project. Every package seems to have a "model" that contains all the data (this term probably originates from the concept MVC or MVVM). The *shared notebook model* is an adaption of the existing concept which uses Yjs' shared types to build a collaborative model of the notebook. Since the function accepts a shared type and returns a shared model, the function should have been called `createCellModelFromSharedType`.

martinRenou:
Thanks for the clarification 👍🏽 I will rename it to `createCellModelFromSharedType` as you suggested

jtpio:
Thanks Martin.

> It's an exported function, so this could be backward incompatible

I think so. In that case it needs a major version bump: https://github.com/jupyterlab/jupyterlab/blob/master/RELEASE.md#js-major-releases

Which is likely to be a command like the following: `jlpm bump:js:major shared-models`

Also do we actually want to `export` that function? Even if we drop the export this will still be a breaking change.

jtpio (approved):
Thanks!

jtpio:
> In that case it needs a major version bump

@martinRenou mind doing a quick rebase of this PR? The versions have been bumped so we don't have to do it in this PR.

jtpio:
> It's an exported function, so this could be backward incompatible

Also there is a wip extension migration guide to cover changes in the JS packages here: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#api-breaking-changes

So we can add a new entry for that rename.

martinRenou:
Done :)

docs/source/extension/extension_migration.rst
```diff
@@ -37,6 +37,8 @@ bumped their major version (following semver convention):
    The ``CodeEditor.ICoordinate`` interface now extends ``DOMRectReadOnly`` instead of ``JSONObject, ClientRect``.
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    The ``NotebookPanel._onSave`` method is now ``private``.
+- ``@jupyterlab/shared-models`` from 3.x to 4.x
+  The ``createCellFromType`` functoin is now named ``createCellModelFromSharedType``
```
jtpio, line 41: Small typo fix:

```suggestion
  The ``createCellFromType`` function has been renamed to ``createCellModelFromSharedType``
```

martinRenou:
Thanks for the reviews!

jtpio:
Waiting for the CI and this should be good to go.

jtpio:
Thanks!

Issue #11376, opened 2021-10-29
Side-by-side rendering feedback


mlucool:
Hi,

What do people think of adding side-by-side rending as global setting? I suspect anyone who likes it will want to be able to make it their default. I like the ability to toggle it per notebook as well (as is there now).

I also find it a bit hard to tell where which cell has which output on a wide screen since you have to focus a bit more as your eye moves across the screen. I wonder if we can do more to help here from a UX POV. 
![image](https://user-images.githubusercontent.com/1813603/139509807-32c30fde-079f-4204-a038-eff3327b11cb.png)

CC @jess-x

krassowski:
> I also find it a bit hard to tell where which cell has which output on a wide screen since you have to focus a bit more as your eye moves across the screen. I wonder if we can do more to help here from a UX POV.

- a simple fix would be increasing the top/bottom margin by a little bit in the side-by-side mode
- one could experiment with guide lines or even-odd colouring but it might be a bit distracting
- I wonder if we should consider swapping the collapser (the blue bar + spacing) location in side-by-side mode (currently it is always on the left, but maybe it should be on the right?)?

jess-x:
Hi, I read this thread along with the feedback from https://github.com/jupyterlab/jupyterlab/issues/11308.
To help improve the UX of side-by-side, I'm proposing the following:

- [ ] Increase the right/left margins to help with eyeballing outputs with relating cells
- [ ] Side-by-side rendering as global setting to enable setting as default
- [ ] The ability to toggle per notebook
- [ ] Click and drag to the left to minimize the size of input cells
- [ ] Enable setting ratio option greater than 1
- [ ] Option to switch between drag resizing the one cell vs all cells

While there will likely be changes, my plan is to open PRs accordingly
If you feel strongly about any item above, thoughts and suggestions will be appreciated!!

echarles:
Thx @krassowski for reopening (was closed via a PR...)

I'd like to get some feedback on the following feature: What do user think about adding settings to define the margin
- for the input
- for the output

This would be values applied to all cells of all notebooks in case of side-by-side rendering.

krassowski:
I took the liberty of ticking the boxes in @jess-x comment instead, it is very satisfying to tick things of the list ;)

[...]

PR #11539, opened 2021-11-25
Increase notebook-cell margin in side-by-side mode

jess-x:
## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/11376

## Code changes

When in side-by-side mode, increase top/bottom margins to `3em` & left/right margins to `5%`

## User-facing changes

Before:

https://user-images.githubusercontent.com/22648112/143334708-b1787149-063e-4e10-a603-331b88a71512.mov

After: 

https://user-images.githubusercontent.com/22648112/143334661-2a8c01b1-ce5d-4b5a-b336-61ab16df382b.mov

### DIFF
--- a/packages/notebook/style/base.css
+++ b/packages/notebook/style/base.css
@@ -304,7 +304,10 @@
 }
 
 .jp-mod-sideBySide.jp-Notebook .jp-Notebook-cell {
-  margin: 2%;
+  margin-top: 3em;
+  margin-bottom: 3em;
+  margin-left: 5%;
+  margin-right: 5%;
 }
 
 .jp-mod-sideBySide.jp-Notebook .jp-CodeCell {
--- a/packages/notebook/style/base.css
+++ b/packages/notebook/style/base.css
@@ -303,6 +303,10 @@
   --jp-side-by-side-resized-cell: var(--jp-side-by-side-output-size);
 }
 
+.jp-mod-sideBySide.jp-Notebook .jp-Notebook-cell {
+  margin: 3em;
+}
+
 .jp-mod-sideBySide.jp-Notebook .jp-CodeCell {
   display: grid;
   grid-template-columns: minmax(0, 1fr) min-content minmax(
--- a/packages/notebook/style/base.css
+++ b/packages/notebook/style/base.css
@@ -304,7 +304,7 @@
 }
 
 .jp-mod-sideBySide.jp-Notebook .jp-Notebook-cell {
-  margin: 3em;
+  margin: 2%;
 }
 
 .jp-mod-sideBySide.jp-Notebook .jp-CodeCell {


echarles:
This looks better with the margins. 2 comments:

1. I see 2 collapsers (the blue vertical lines). I am no more sure if this was already there before. Is this intented?
2. You have set the margin to `margin: 3em;`. Have you tried a '%' instead of a fixed value? Just asking to see if we could have a margin depending on the actual width of the notebook panel.

Screenshot 2021-11-25 at 07 50 08

jess-x:
Screen Shot 2021-11-26 at 2 03 01 PM
This is how `2%` notebook-cell margins look on my screen. I cannot actually decide between this and `3em` by eyeballing the minimal difference they have on my screen, but using '%' does make more intuitive sense to me. Thanks @echarles

jess-x:
> I see 2 collapsers (the blue vertical lines). I am no more sure if this was already there before. Is this intended?

Hmm, the left most ones are the input collapsers and the ones on the right side are output collapsers, I believe. Visually they can be confusing, I see. I could address this separately

echarles:
> Hmm, the left most ones are the input collapsers and the ones on the right side are output collapsers, I believe. Visually they can be confusing, I see. I could address this separately

Yeah, I have been fooled by the input and output collapsers. Let's leave it as they are for now.

echarles:
> This is how 2% notebook-cell margins look on my screen. I cannot actually decide between this and 3em by eyeballing the minimal difference they have on my screen, but using '%' does make more intuitive sense to me. Thanks @echarles

Prolly just set only `margin-left` and `margin-right` instead of `margin`(margin will set the top and bottom, which we don't need here)

I have trid with `margin-left: 5%` and `margin-right: 5%` and sounds to me like giving a better display, especially when you resize the browser to a small viewport.

jess-x:
> Prolly just set only margin-left and margin-right instead of margin(margin will set the top and bottom, which we don't need here)
I have tried with margin-left: 5% and margin-right: 5% and sounds to me like giving a better display, especially when you resize the browser to a small viewport.

I think right now in side-by-side mode, the cells are too clustered together, that's why users are having a harder time focusing outputs with their input cells. I am aware by increasing `margin`, `margin-top` and `margin-bottom` are both increased under the hood. That is an intentional UI/UX decision from me, and I'd argue that's better than just increasing left/right margins. In terms of  `2%` vs `5%`, again I don't have a strong opinion/argument as I don't have access to various screens about different sizes. So it's I think A/B testing that with power users might let us make a more educated decision (if that's possible). With that said, I personally like the current implementation (`margin: 2%`), but feel free to push to this branch if needed :)

echarles:
> I think right now in side-by-side mode, the cells are too clustered together, 

Thx for the reminder. In that case, increasing top and bottom makes sense.

> And I think A/B testing that with power users might let us make a more educated decision (if that's possible). 

True. You have spent more time than me trying the different options. 2% was ok for me also.

Let's wait a bit before merging for anyone wiling to give it a try and provide feedback.

fcollonval:
I tried it on my ultra wide screen. I would not use a percentage for the vertical margin. Even 2% starts to be a lot. And if you close or not a side panel (see below), this changes the displayed cells although the height did not change. So I would use a constant vertical margin not a variable one.

| With filebrowser open | Without file browser |
| --- | --- |
| side-by-side_ | side_by_side_no_filebrowser |

echarles:
> So I would use a constant vertical margin not a variable one.

Makes sense.

jess-x:
@fcollonval @echarles thanks!! I changed top/bottom margins to `3em`, left/right margins to `5%`. On my screen, looks like this right now 
Screen Shot 2021-11-29 at 7 01 49 PM

jtpio:
Thanks @jess-x.

> left/right margins to 5%

@fcollonval how does the 5% look on the wide display?

fcollonval:
> @fcollonval how does the 5% look on the wide display?

| With sidebar | Without sidebar |
| --- | --- |
|wide_screen_with_sidebar | wide_screen_no_sidebar|

fcollonval (approved):
Thanks @jess-x

echarles:
Thx @jess-x

echarles:

PR #7103, opened 2019-08-28
Map filetype to  widget factory by the `mimeTypes` property of each

telamonian:
## References

fixes #7099

PR #11540, opened 2021-11-25
Allow to link factory to file type when adding it

### LINKED ISSUE #7099: Add a new filetype to an existing widget factory
I asked a question on [gitter](https://gitter.im/jupyterlab/jupyterlab?at=5d65a279a498b67237a87328):

> If I register a new filetype in an extension, is there a way to make it so that the JSON viewer shows up in the "Open With" menu when you right click on an item of the new filetype in filebrowser?

Apparently the answer is no. A widget factory's filetypes are used to determine what goes into the "Open With" context menu of a filebrowser listing. Currently, you can specify the filetypes when you create a new document widget factory, but are out of luck otherwise:

https://github.com/jupyterlab/jupyterlab/blob/5a8ceaab45d2cb70e096304ad7d5581fff8c9ca5/packages/docregistry/src/registry.ts#L962-L965

There are some cases where it would be good to be able to use an existing widget factory with a new filetype. For example, for files in a JSON format it would be nice to be able to allow optionally opening the file using the JSON document viewer.

fcollonval:
## References

Fix #7099

And fix bug that not all file types for a path were returned

## Code changes

- Add optional factory names option to `addFileType`; if no default factory for the new file type is found, add the first given factory as default.
- Find all matching extensions for a given path in the file types list

## User-facing changes

N/A

Extension authors will be able to registered more easily new file types that should be opened with existing factories (such as Jupytext)

N/A

### DIFF
--- a/packages/docregistry/src/registry.ts
+++ b/packages/docregistry/src/registry.ts
@@ -275,14 +275,21 @@ export class DocumentRegistry implements IDisposable {
   /**
    * Add a file type to the document registry.
    *
-   * @params fileType - The file type object to register.
+   * @param fileType - The file type object to register.
+   * @param factories - Optional factories to use for the file type.
    *
    * @returns A disposable which will unregister the command.
    *
    * #### Notes
    * These are used to populate the "Create New" dialog.
+   *
+   * If no default factory exists for the file type, the first factory will
+   * be defined as default factory.
    */
-  addFileType(fileType: Partial<DocumentRegistry.IFileType>): IDisposable {
+  addFileType(
+    fileType: Partial<DocumentRegistry.IFileType>,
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks!

Looks like there is a small conflict to fix.

jtpio:
Looks like this should also close https://github.com/jupyterlab/jupyterlab/pull/7103? (which was opened as fixing #7099)

jtpio (approved):
Thanks!

fcollonval:
PR #7103 sounds like a more generic alternative to this one as file type may become automatically compatible with existing factory through their mime type.

fcollonval:

Issue #10875, opened 2021-08-19
Changing default viewer for PDF in User Preferences does not work


imcovangent:
## Description

I want to change the default PDF viewer from the standard PDF viewer to PDFJS. The options in the Settings tab clearly indicates that this is possible:

![image](https://user-images.githubusercontent.com/15325525/130033651-b770f417-2cc2-4991-8feb-790eed7f1cdb.png)

However, when I apply this setting. My PDFs still open in the standard PDF viewer and not in PDFJS.

## Reproduce

Install jupyterlab latex extension to make the PDFJS viewer available.

Go to Settings -> Advanced Settings Editor -> Document manager

Change User Preferences to:

![image](https://user-images.githubusercontent.com/15325525/130033804-2dbfa362-7c75-439b-b0ec-6ce45f05dc9a.png)

I do get this message in the browser console:

![image](https://user-images.githubusercontent.com/15325525/130033878-96f50ff6-6b56-45c4-9ee8-cb71895b459a.png)

More surprisingly, I even get this message when I set it to PDF instead of PDFJS:

![image](https://user-images.githubusercontent.com/15325525/130034064-45934d18-605c-41c9-bddb-62fcdcd56d8f.png)

![image](https://user-images.githubusercontent.com/15325525/130034030-d9d35a12-0cab-4013-b917-47d591ed87d2.png)

## Expected behavior

I would expect PDFs to be opened with the PDFJS viewer.

## Context

- Operating System and version: Ubuntu 18.04
- Browser and version: Chrome 92.0.4515.159
- JupyterLab version: 3.1.6

krassowski:
Thank you for reporting this issue .This is a duplicate of https://github.com/jupyterlab/jupyterlab/issues/7776 which was fixed in https://github.com/jupyterlab/jupyterlab/pull/11541; currently the fix is scheduled to be included in 4.0 but maybe we can backport it.

fcollonval:
I don't think this is a duplicate issue. But I don't know about that viewer `PDFJS`. Could you point to the extension providing it? 

From the error displayed in the console, it looks like the viewer factory is not properly set to handle `PDF` file type.

fcollonval:
I got an idea. Could you try setting: `{"pdf": "PDFJS"}` (with the key being lower case)?

LiberaVeritas:
I have the exact same issue. I tried setting both lowercase and uppercase, as well as on the override json. 

[...]

PR #11541, opened 2021-11-25
User defined default viewer take precedence for rendered factory

### LINKED ISSUE #7776: Cannot override the default rendered viewer
In https://github.com/jupyterlab/jupyterlab/pull/6813, @ian-r-rose wrote a nice setting for overriding the default editor for a file. This is used in preferredWidgetFactories to allow a user to control a default editor for the file: https://github.com/jupyterlab/jupyterlab/blob/57cd4545b93606177950e7189d03394b26c63b0c/packages/docregistry/src/registry.ts#L334-L339

However, this override is *not* taken into account when we ask for the default rendered widget factory: https://github.com/jupyterlab/jupyterlab/blob/57cd4545b93606177950e7189d03394b26c63b0c/packages/docregistry/src/registry.ts#L406-L421

This means that we can't set the default renderer for clicking on a file in a rendered markdown, for example. This is confusing since, for example, the setting name is "defaultViewers" (and not defaultEditors or something).

Should we allow a separate override for the default rendered widget?

This came up in https://github.com/mwouts/jupytext/issues/271

fcollonval:
## References

Fix #7776

## Code changes

Look first in user overrides list for rendered widget so it takes precedence.

## User-facing changes

Some link may open with a different editor if they change the default viewer settings.

N/A

### DIFF
--- a/packages/docregistry/src/registry.ts
+++ b/packages/docregistry/src/registry.ts
@@ -401,22 +401,33 @@ export class DocumentRegistry implements IDisposable {
    * file types and there is a match in that set, this returns that.
    * Otherwise, this returns the same widget factory as
    * [[defaultWidgetFactory]].
+   *
+   * The user setting `defaultViewers` took precedence on this one too.
    */
   defaultRenderedWidgetFactory(path: string): DocumentRegistry.WidgetFactory {
     // Get the matching file types.
-    const fts = this.getFileTypesForPath(PathExt.basename(path));
+    const ftNames = this.getFileTypesForPath(PathExt.basename(path)).map(
+      ft => ft.name
+    );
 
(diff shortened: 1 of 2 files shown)


jtpio (approved):
Thanks!

krassowski:
This fixes https://github.com/jupyterlab/jupyterlab/issues/10875 - should we backport it?

fcollonval:
> This fixes #10875 - should we backport it?

I don't think it fixes that issue because this addresses a corner case: the viewer to pick when a local file is opened from a link in a markdown cell for example). Before this PR, the user settings was ignored when looking for the viewer is that case.

fcollonval:

PR #11544, opened 2021-11-25
RTC: Sync markdown cell  "rendered" state

### LINKED ISSUE #11406: [RTC] Collaborator cursor/selection not displayed for markdown cells

martinRenou:
## References

Fix #11406, follow up of https://github.com/jupyterlab/jupyterlab/pull/11518

## Code changes

We now sync more cell state between collaborators

## User-facing changes

https://user-images.githubusercontent.com/21197331/143414840-f3fee846-149f-4122-ab2a-e9a73b6eae6a.mp4

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -80,6 +80,12 @@ import { ResizeHandle } from './resizeHandle';
 
 import { Signal } from '@lumino/signaling';
 import { addIcon } from '@jupyterlab/ui-components';
+import {
+  CellChange,
+  ISharedBaseCellMetadata,
+  ISharedMarkdownCell,
+  YMarkdownCell
+} from '@jupyterlab/shared-models';
 
 /**
  * The CSS class added to cell widgets.
@@ -1465,6 +1471,11 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     // Stop codemirror handling paste
     this.editor.setOption('handlePaste', false);
 
+    // Initialize rendered to true
+    if (this.rendered === null) {
+      (this.model.sharedModel as YMarkdownCell).rendered = true;
+    }
+
     // Check if heading cell is set to be collapsed
     this._headingCollapsed = (this.model.metadata.get(
       MARKDOWN_HEADING_COLLAPSED
@@ -1476,7 +1487,7 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
       timeout: RENDER_TIMEOUT
     });
     this._monitor.activityStopped.connect(() => {
-      if (this._rendered) {
+      if (this.rendered) {
         this.update();
       }
     }, this);
@@ -1486,6 +1497,17 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     });
     this.renderCollapseButtons(this._renderer!);
     this.renderInput(this._renderer!);
+
+    (this.model.sharedModel as YMarkdownCell).changed.connect(
+      this._onSharedModelChanged,
+      this
+    );
+    this.model.sharedModelSwitched.connect(() => {
+      (this.model.sharedModel as YMarkdownCell).changed.connect(
+        this._onSharedModelChanged,
+        this
+      );
+    }, this);
   }
 
   /**
@@ -1557,21 +1579,15 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
   /**
    * Whether the cell is rendered.
    */
-  get rendered(): boolean {
-    return this._rendered;
+  get rendered(): boolean | null {
+    return (this.model.sharedModel as YMarkdownCell).rendered;
   }
-  set rendered(value: boolean) {
-    if (value === this._rendered) {
+  set rendered(value: boolean | null) {
+    if (value === this.rendered) {
       return;
     }
-    this._rendered = value;
-    this._handleRendered();
-    // Refreshing an editor can be really expensive, so we don't call it from
-    // _handleRendered, since _handledRendered is also called on every update
-    // request.
-    if (!this._rendered) {
-      this.editor.refresh();
-    }
+
+    (this.model.sharedModel as YMarkdownCell).rendered = value;
   }
 
   protected maybeCreateCollapseButton(): void {
@@ -1688,10 +1704,35 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
   }
 
   /**
+   * Handle a change to the cell shared model.
+   */
+  protected _onSharedModelChanged(
+    sender: ISharedMarkdownCell,
+    change: CellChange<ISharedBaseCellMetadata>
+  ): void {
+    if (change.stateChange) {
+      change.stateChange.forEach(stateChange => {
+        switch (stateChange.name) {
+          case 'rendered':
+            this._handleRendered();
+
+            // Refreshing an editor can be really expensive, so we don't call it from
+            // _handleRendered, since _handledRendered is also called on every update
+            // request.
+            if (!this.rendered) {
+              this.editor.refresh();
+            }
+            break;
+        }
+      });
+    }
+  }
+
+  /**
    * Handle the rendered state.
    */
   private _handleRendered(): void {
-    if (!this._rendered) {
+    if (!this.rendered) {
       this.showEditor();
     } else {
       // TODO: It would be nice for the cell to provide a way for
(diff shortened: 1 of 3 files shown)


packages/shared-models/src/ymodels.ts
```diff
@@ -696,6 +696,25 @@ export class YBaseCell<Metadata extends models.ISharedBaseCellMetadata>
...
+      if (
+        ['metadata', 'execution_count', 'source'].find(value => value === key)
+      ) {
+        return;
+      }
+
+      const change = modelEvent.changes.keys.get(key);
+      stateChange.push({
+        name: key,
+        oldValue: change?.oldValue ? change!.oldValue : 0,
```
fcollonval, line 712: Returning `0` as default value is dangerous because we don't know the type of the state element. I would suggest to stick with the same behavior than the one of execution count or metadata (aka `undefined` if unknown)

```suggestion
        oldValue: change?.oldValue,
```

packages/shared-models/src/ymodels.ts
```diff
@@ -1031,6 +1050,26 @@ export class YMarkdownCell
       attachments: this.getAttachments()
     };
   }
+
+  /**
+   * Gets whether or not the cell is rendered.
+   *
+   * @returns The cell "rendered" state.
```
fcollonval, line 1057: Getter/setter are not documented with `@returns` and `@params`.  Just one doc string describing the attribute on the getter.

```suggestion
   * Whether or not the cell is rendered.
```

packages/shared-models/src/ymodels.ts
```diff
@@ -1031,6 +1050,26 @@ export class YMarkdownCell
...
+   */
+  get rendered(): boolean {
+    return this.ymodel.get('rendered');
+  }
+
+  /**
+   * Sets the cell "rendered" state
+   *
+   * @param rendered: The cell rendered state.
+   */
```
fcollonval, line 1067:
```suggestion
```

packages/shared-models/src/ymodels.ts
```diff
@@ -1031,6 +1050,26 @@ export class YMarkdownCell
       attachments: this.getAttachments()
     };
   }
+
+  /**
+   * Gets whether or not the cell is rendered.
+   *
+   * @returns The cell "rendered" state.
+   */
+  get rendered(): boolean {
```
fcollonval, line 1059: I would change the type to be `boolean` or `null` as only markdown cells have that state.

```suggestion
  get rendered(): boolean | null {
```

packages/cells/src/widget.ts
```diff
@@ -1564,7 +1578,10 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     if (value === this._rendered) {
       return;
     }
+
     this._rendered = value;
```
fcollonval, line 1582: I suggest removing the private attribute `_rendered` to have a single source of truth in the model (that will avoid troubles of duplicated information).

fcollonval (changes requested):
Thanks @martinRenou I have some suggestions.

packages/shared-models/src/ymodels.ts
martinRenou: Thanks! That's a bad copy-paste

packages/shared-models/src/ymodels.ts
```diff
@@ -505,6 +505,10 @@ export class YBaseCell<Metadata extends models.ISharedBaseCellMetadata>
     this.ymodel = ymodel;
     const ysource = ymodel.get('source');
     this._prevSourceLength = ysource ? ysource.length : 0;
+
+    this.ystate = new Y.Map();
```
dmonad, line 509: What is the difference between `ystate` and `ymodel`?

packages/shared-models/src/ymodels.ts
```diff
@@ -1031,6 +1055,18 @@ export class YMarkdownCell
...
   }
+
+  /**
+   * Whether or not the cell is rendered.
+   */
+  get rendered(): boolean | null {
+    return this.ystate.get('rendered');
+  }
+  set rendered(rendered: boolean | null) {
+    this.transact(() => {
```
dmonad, line 1066: This shouldn't be tracked by the undomanager. So it would be best to do this without a transaction.

dmonad (changes requested):
I have mixed feelings about this. On the one hand, I can see that this is an improvement to the user experience when people are actively collaborating. On the other hand, I feel that a user should be able to leave the Markdown in render-mode while another user is editing it. 

One case where the current behavior might be preferable is when a single user opens the Markdown in two different browser windows: The one browser window is for editing and the other for rendering Markdown. The markdown should update in real time while the user is editing it.

We should probably first discuss this before it is merged.

In any case, good job on the implementation :+1: I added a few comments.

packages/cells/src/widget.ts
```diff
@@ -1486,6 +1497,17 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     });
     this.renderCollapseButtons(this._renderer!);
     this.renderInput(this._renderer!);
+
+    (this.model.sharedModel as YMarkdownCell).changed.connect(
+      this._onSharedModelChanged,
+      this
+    );
+    this.model.sharedModelSwitched.connect(() => {
+      (this.model.sharedModel as YMarkdownCell).changed.connect(
```
martinRenou, line 1506: Would be good to disconnect from the old shared model here

packages/shared-models/src/ymodels.ts
martinRenou: Thanks, fixing it :)

packages/shared-models/src/ymodels.ts
```diff
@@ -505,6 +505,10 @@ export class YBaseCell<Metadata extends models.ISharedBaseCellMetadata>
     this.ymodel = ymodel;
     const ysource = ymodel.get('source');
     this._prevSourceLength = ysource ? ysource.length : 0;
+
+    this.ystate = new Y.Map();
+    this.ymodel.set('state', this.ystate);
```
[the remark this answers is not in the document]
martinRenou: Thanks for the review! It was indeed problematic to overwrite the `ystate` here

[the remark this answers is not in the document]
martinRenou: So you're saying we should make `ystate` an extra constructor argument?

packages/shared-models/src/ymodels.ts
martinRenou: `ystate` would be extra state here, state that is not part of the actual json model. Though `ystate` is still an element of the `ymodel` in the implementation, so it's debatable.

martinRenou:
> On the other hand, I feel that a user should be able to leave the Markdown in render-mode while another user is editing it.

Yep, that's a good point.

Maybe a better UX would be to have a visual indicator showing that a collaborator is editing the markdown cell when it's rendered. What do you think?

martinRenou:
@dmonad I opened #11555 for discussion. Let's make this PR draft.

krassowski:
Maybe we could avoid unrendering the markdown cell and show the cursor in rendered markdown by feeding a custom tag prior to marked.js parsing (I previously mentioned it in https://github.com/jupyterlab/jupyterlab/issues/11555#issuecomment-979888354 but that was not the best palce to post it)? We already do that for Tex rendering:

https://github.com/jupyterlab/jupyterlab/blob/ffc85e41cd4852b84786f476d63024c92d8df528/packages/rendermime/src/renderers.ts#L320-L350

We would just add a cursor at the correct character position as a special escape sequence that would be _very unlikely_ to be used by the user (same as we do for Tex) and then substitute it with our custom cursor after Markdown rendering.

dmonad:
Thank you for the suggestion @krassowski. I think that would be a much better approach.

https://github.com/jupyterlab/jupyterlab/issues/11555 Might also take care of this issue. At least users will now be able to see that clients are editing the markdown cell. While the markdown cell is in rendering modus, there is not really a need to see exactly where users are typing.

fcollonval:
> there is not really a need to see exactly where users are typing.

Actually when doing RTC to explain things up, I like to be able to set my cursor and/or select the relevant element. This is the equivalent to pointing your finger to the screen in pair programming; very useful.

martinRenou:
I personally believe both #11555 and @krassowski's suggestion should be addressed, they are kind of complementary IMO.

martinRenou:
Let's close this PR. I guess we all agree we won't be taking this approach of syncing the rendered state.

Issue #11548, opened 2021-11-25
Failing `js-services` test on CI


jtpio:
## Description

`js-services` tests have started to fail more frequently on CI lately.

This might be related to the latest `jupyter_server` released a couple of days ago: https://github.com/jupyter-server/jupyter_server/releases/tag/v1.12.0

Which includes support for pending kernels (just a guess but sounds related)

## Reproduce

Latest run on `master`: https://github.com/jupyterlab/jupyterlab/runs/4321818438?check_suite_focus=true

jtpio:
This is the failing test: 

![image](https://user-images.githubusercontent.com/591645/143607026-89ff2ad3-459b-421a-9452-10cdfb38c5d5.png)

jtpio:
Fixed by https://github.com/jupyter-server/jupyter_server/pull/617

PR #11552, opened 2021-11-25
postcss 8.4.0 breaks integrity 2 CI test

### LINKED ISSUE #11550: postcss 8.4.0 breaks integrity 2 CI test
The newly released postcss 8.4.0 https://github.com/postcss/postcss/releases due to that change:

```
Lazy Parsing

Previously, we found that many tools run PostCSS even if the developer didn’t pass any PostCSS plugins. Parsing is the most expensive step in CSS processing. It led to a waste of resources without any reason.
```

We are exactly doing that as we only check that the files can be parsed with postcss.

I'm not sure what is the best workaround here. Should we run postcss with some dummy plugin just to ensure it parses the files?

fcollonval:
## References

Fixes #11550

## Code changes

N/A

## User-facing changes

N/A

N/A

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,0 +1,21 @@
+/* -----------------------------------------------------------------------------
+| Copyright (c) Jupyter Development Team.
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+
+// Since postcss 8.4.0, parsing the CSS is done only if some plugin are defined
+// As we only want to check that the CSS files can be processed by postcss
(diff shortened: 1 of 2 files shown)


jtpio:
> So we can add it to the list here:

Done in https://github.com/jupyterlab/jupyterlab/pull/11552/commits/c75ce70591dc4a49694f638bccd55ced00dfc98e

jtpio (approved):
Thanks!

jtpio:
Remaining CI failure tracked in #11548

jtpio:

Issue #10139, opened 2021-04-22
[UX] Debugger Feedback


mlucool:
### Problem

The debugger is an amazing addition (thanks to all who made it happen!). I gave it a try and want to propose some ways we can improve the experience. This thread is to start a dicussion on various ideas. Some of these may end up being big enough to spin into their own issues and others may be more of a true feature request then a simple UX enhancement.

I am using jupyterlab (3.1.0a5) and ipykernel (6.0.0a4). 

### Proposed Solution

1. When the debugger starts, any step is taken (next, step in, step out), or a level in the callstack is selected, the source should scroll current LOC is on screen (i.e. the orange line).
2. Clicking on a breakpoint should scroll to that breakpoint in the corresponding source file
2. Clicking on clear breakpoints should have a confirmation box
2. It should be possible to easily open any file in a mode that the debugger can set breakpoints. Today, you can the open a source that is being debugged in the main area and set additional breakpoints. I'd like to also open other source files and set breakpoints in them before the kernel loads them. Even a weaker version that requires the kernel to have loaded it would be a big help!
  a. Something like chrome's ctrl+o menu would be immensely helpful. As an idea, if I typed `foo.bar` it would open `foo.bar.__file__` and let me debug.
![image](https://user-images.githubusercontent.com/1813603/115774039-11382b80-a37f-11eb-8783-0d0d2817d0ee.png)
  b. It should be possible to see all loaded sources in a tree form to navigate and set a breakpoint in any file.
![image](https://user-images.githubusercontent.com/1813603/115773722-a2f36900-a37e-11eb-9685-46384acbf555.png)
5. There should be a trivial way to copy a variable into the kernel from the UI
![image](https://user-images.githubusercontent.com/1813603/115773860-d0401700-a37e-11eb-97f9-99fb4292d323.png)
10. Make it easier to copy simple types (numbers, Booleans, objects) from the variable inspector
6. Have a toggle that sets a breakpoint as soon as any exception is thrown
7. Include filenames next to each level in the callstack
8. Maybe this is a user setting, but it feels like the various panels should use a smaller font or maybe just less spacing to make this more data dense
![image](https://user-images.githubusercontent.com/1813603/115774938-35483c80-a380-11eb-9b87-b7c651f62e8d.png) vs. ![image](https://user-images.githubusercontent.com/1813603/115774966-3d07e100-a380-11eb-992b-ab4fcc6885e0.png)
[...]

PR #11566, opened 2021-11-28
Show the kernel sources as a debugger tab and allow the user to break in kernel sources

echarles:
## References

` It should be possible to easily open any file in a mode that the debugger can set breakpoints. Today, you can the open a source that is being debugged in the main area and set additional breakpoints. I'd like to also open other source files and set breakpoints in them before the kernel loads them. Even a weaker version that requires the kernel to have loaded it would be a big help!`

## Code changes

Add a new tab.

## User-facing changes

143607893-c307780b-ff45-4261-a21d-5dfe6dce523f

None

## Todo

[ ] Ensure a  breakpoint set in the kernel sources has some effect
[ ] Remove the created kernel variable
[ ] Show the kernel sources as a tree
[ ] Polish, polish...


echarles:
This is a first attempt to show the user the kernel sources, open any kernel source and allow the user to set breakpoint.

As shown in the screenshot, a breakpoint set in a kernel source has no effect.  Any idea or hint for this issue? @JohanMabille @jtpio @fcollonval

fcollonval:
One idea: it can require a configuration to be sent to debugpy. For example in VS Code the debugger is only tracking files in the workspace and a parameter needs to be set for it to work on any files.

echarles:
> One idea: it can require a configuration to be sent to debugpy. For example in VS Code the debugger is only tracking files in the workspace and a parameter needs to be set for it to work on any files.

Yeah, this sounds like a great hint. I have googled to find the debugpy setting to add and looked in https://code.visualstudio.com/docs/python/debugging#_command-line-debugging but did not find it. Maybe a lazy Sunday morning here :)

Any link to that?

echarles:
On https://code.visualstudio.com/docs/python/debugging#_justmycode 

"When omitted or set to true (the default), restricts debugging to user-written code only. Set to false to also enable debugging of standard library functions."

More also on e.g. https://github.com/microsoft/debugpy/issues/159

Debugpy test on https://github.com/microsoft/debugpy/blob/2ae7dc96ecc4ca73b119c2056284d203afa10afd/tests/debugpy/test_justmycode.py#L28

echarles:
I have opened https://github.com/ipython/ipykernel/issues/811 on ipykernel repository, as it seems to me that the kernel debugger must be started with configuration options.

jtpio:
Linking to https://github.com/jupyterlab/debugger/issues/263 where this was previously discussed.

echarles:
Did some more experiment with the `just_my_code`. On my env, to get the stacktrace with ipykernel shown in the UI, I had to manually change some `_vendored/pydevd/_pydevd_bundle/...` sources. With those changes, I was able to nicely see the stack in the UI like this, but adding breakpoints in the non-user sources remains non effective (the debugger does not stop on those breakpoints).

Screenshot 2021-11-29 at 10 02 24

These are the changes I made in the vendored `debubpy` code.

Screenshot 2021-11-29 at 10 02 30

echarles:
> With those changes, I was able to nicely see the stack in the UI like this, but adding breakpoints in the non-user sources remains non effective (the debugger does not stop on those breakpoints).

Finally, with those changes, the breakpoints have effect if you put them in the body of a method (I was setting at field level, which looks like having no effect). The following gif shows that you can pause at a kernel file level.

The next step will be to understand how to setup the debug session with the "justMyCode" property (for the doc and source, it is not clear ATM).

Untitled

echarles:
> The next step will be to understand how to setup the debug session with the "justMyCode" property (for the doc and source, it is not clear ATM).

I have opened https://github.com/ipython/ipykernel/pull/812 in ipykernel to pass the needed flag when attaching the debug session. With that change, this PR allows users to break in any code.

packages/debugger/src/panels/kernelSources/body.ts
```diff
@@ -0,0 +1,215 @@
...
+                label: key,
+                tooltip: value as string
+              });
+              layout.addWidget(button);
+            });
+          };
+        }
+      );
+      const code = `
+import sys, re, json
```
jtpio, line 93: Since this is Python specific, is there another language agnostic way for retrieving the sources?

jtpio: Or is this mostly a workaround for `supportsLoadedSourcesRequest` as mentioned in https://github.com/jupyterlab/jupyterlab/issues/10139#issuecomment-872169353?

echarles: It is indeed a workaround on the lack of support for `SourcesRequest`. I have tried to use Requests_Modules https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Modules but it returns only a partial list of the expected modules, not sure why (in my case 13 modules, missing a lot of the existing and loaded modules).

I guess the Kernel sources could be activated in first instance for python kernels only. Not sure how to identify the kernel is python...?

jtpio: There is `language_info` as part of the kernel info reply. Checking that and whether debugging is available could be enough to activate the kernel sources.

Another way would be to improve the extensibility of the core jupyterlab debugger / sidebar so it's easy to add new panels, if that's not the case already. And implement the Python specific kernel sources logic in a separate extension.

echarles: > There is language_info as part of the kernel info reply. Checking that and whether debugging is available could be enough to activate the kernel sources.

Thx, will implement that in this PR.

> Another way would be to improve the extensibility of the core jupyterlab debugger / sidebar so it's easy to add new panels, if that's not the case already. And implement the Python specific kernel sources logic in a separate extension.

I have opened https://github.com/jupyterlab/jupyterlab/issues/11588 to discuss that. For this specific case, I think users want it soon in core and we are already far enough to further progress with this adhoc implementation. I think we also need more toughts and cases to correctly decide on to create the flexible and extensible solution.

jtpio: > 

Were you able to try overriding the handling of that specific request on the kernel side? Instead of forwarding the message to `debugpy`. This might allow running the custom Python code currently sent as a Comm message directly on the kernel instead.

For reference there was something similar in `xeus-python` to be able to open sources: https://github.com/jupyter-xeus/xeus-python/pull/168

echarles: > Were you able to try overriding the handling of that specific request on the kernel side? Instead of forwarding the message to debugpy. This might allow running the custom Python code currently sent as a Comm message directly on the kernel instead.

I haven't tried to override the Request_Modules, but it sounds something worth to try. Will have a look.

JohanMabille: >  I have tried to use Requests_Modules https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Modules but it returns only a partial list of the expected modules, not sure why (in my case 13 modules, missing a lot of the existing and loaded modules).

I guess this is because debupy capability `supportsModulesRequest` is [False](https://github.com/microsoft/debugpy/blob/main/src/debugpy/adapter/servers.py#L241), although I don't understand why debugpy returns something instead of throwing an exception.

I think all these requests (ModuleRequest, LoadedSourceRequest) should be implemented on the kernel side. JupyterLab sould only check whether the requets was successful (kernels not supporting these requests should return a response with the success field set to `False`).

echarles:
The last commit uses the Modules request served by a new features still to be merged in ipykernel with https://github.com/ipython/ipykernel/pull/816

jtpio:
Looking good, thanks @echarles for working on this :+1:

echarles:
I have a weird issue with this new ModuleRequest approach. For "some" reason, a few seconds after finishing my "debug session", the debugger pauses on that `selectors.py`. I didn't have that behavior using the Comm. For now, I have no clue where the issue is coming from.

Maybe relevant (or not?), when the UI suddenly shows the following, I see some errors in the log relate to the kernel python trying to deliver a response for the variables request (*)

Screenshot 2021-12-01 at 14 53 12

(*)

```
  File "/Users/echarles/datalayer/repos/jupyter-kernel-python/ipykernel/kernelbase.py", line 281, in process_control
    await result
  File "/Users/echarles/datalayer/repos/jupyter-kernel-python/ipykernel/kernelbase.py", line 857, in debug_request
    reply_content = await reply_content
  File "/Users/echarles/datalayer/repos/jupyter-kernel-python/ipykernel/ipkernel.py", line 426, in do_debug_request
    return await self.debugger.process_request(msg)
  File "/Users/echarles/datalayer/repos/jupyter-kernel-python/ipykernel/debugger.py", line 610, in process_request
    reply = await handler(message)
  File "/Users/echarles/datalayer/repos/jupyter-kernel-python/ipykernel/debugger.py", line 457, in variables
    [var for var in reply['body']['variables'] if self.accept_variable(var['name'])]
KeyError: 'body'
```

JohanMabille:
I actually wonder if the code asking for the loaded modules should not be executed on the shell instead of the control since they now run on different threads. Comm messages are executed on the shell.

JohanMabille:
Hi @echarles , what are the steps to reproduce the issue? I've built and installed your branches of JupyterLab and ipykernel, and I've played a bit with them, and I could not reproduce your issue. Also I noticed that starting the debugger was super slow.

echarles:
@JohanMabille 

> what are the steps to reproduce the issue? 

On my env:

- Wait for the `Kernel Source` tab is populated with the kernel sources
- Open 'os.py'
- Set a breakpoint in the `makedirs` method
- Run the following cell with debug enabled.

```
import os
# os.removedirs('tmp_dir')
os.makedirs('tmp_dir')
print('done')
```

- start the cell with debug enabled
- pause in os.py
- continue
- cell is correctly executed
- +/- 1 second after the cell execution, I break suddenly and without reason in `selector.py``

>  Also I noticed that starting the debugger was super slow.

We should load the modules in async way. I will double check if it is not already done.

JohanMabille:
The kernel sources panel does not get populated. However, as soon as I open a kernel file in the main area (by sliding the source file from the debugger panel for instance), I can reproduce the issue.

echarles:
> The kernel sources panel does not get populated.

Thx for letting me know. Is it with ipykernel or xeus? I will double check what is going on once we have a solution for the other issue.

> However, as soon as I open a kernel file in the main area (by sliding the source file from the debugger panel for instance), I can reproduce the issue.

Good that you can reproduce. Just to make sure you have all details in hands, that was not happening when I was pulling the kernel sources with a Comm (code in my previous commits, no more in the latest version of this PR).

JohanMabille:
With ipykernel

echarles:
Thx a lot @JohanMabille for your help and investigations on this. I have also spent time and can confirm your findings.

I don't see for now a way to understand that, so can not imagine a fix.

As comparison point and as fallback solution, I have opened https://github.com/jupyterlab/jupyterlab/pull/11692 which uses a Comm and does not suffer from that side-effect. Maybe you can confirm it also works on your env.

echarles:
@JohanMabille Have you an idea where the issue is coming from. Otherwise, I'd like to progress to a merge with the alternative solution opened in https://github.com/jupyterlab/jupyterlab/pull/11692

JohanMabille:
Sorry I didn't have the time to investigate more. I think we should not merge #11692: we should investigate and solve the bug in the front-end rather than working around it with a bad change in the protocol that will make it weaker. Debug messages should be sent on the control channel only, whatever the message is. I'll try to investigate more tomorrow.

mlucool:
I tried with #11692 just to be able to test. Overall I find this feature a strong net positive and am very excited to have it!

**Issues**
- Restarting a kernel does not clear sources list
- Values should be sorted
- The list should appear as soon as the debugger is activated (and not first breakpoint). From there, I could see limiting how often it is refreshed so that we aren't slowing everything down fetching this all the time

**UX Suggestions:**
- Make double click for sources not single slick. This makes copy/paste possible for example
- It's hard to find values. I think a tree may be more clicks than we need right now. How about we add a regexp based filter and allow a user setting so that we can set a good default (e.g. we may set ours to be deshaw)
- [Low priority] I think things with breakpoints already set should stand out in some way

FWIW with a sample notebook, I saw ~250kb of data sent to populate this list. In my environment that's not a big deal.

echarles:
I am not sure anymore where I have read, but I think @jtpio or @fcollonval (?) has proposed to update the debugger extension so that extension authors could add their own panel.

Unless we can find a solution to the sideffect, and because the Comm solution does not get buy-in so far, I may start going to that direction so this feature can be developed as an extension. Any feedback?

fcollonval:
Today when reviewing the pause on exception, I have seen that the break on uncaught exception is set to true by default on ipykernel:

image

Could it be related?

---

> I am not sure anymore where I have read, but I think @jtpio or @fcollonval (?) has proposed to update the debugger extension so that extension authors could add their own panel.

The `ISidebar` token has methods to add widgets into it (note it will change in 4.0 to inherit directly from `Panel`).

echarles:
> Today when reviewing the pause on exception, I have seen that the break on uncaught exception is set to true by default on ipykernel - Could it be related?

Thx for the hint. I have looked more at the code and the docs and it sounds it like this reply informs that one can add breakpoints on uncaught exception, but as far as you don't ask to add that breakpoint, it should not be there (this reply is just "informative" about what you can or can not do).

.... but I agree that the behavior really looks like an exception breakpoint would have been set, so I will investigate more on that direction.

If we can not find something, I will prolly go the extension road and implement this feature as an extension adding a panel via `ISidebar`.

JohanMabille:
> Today when reviewing the pause on exception, I have seen that the break on uncaught exception is set to true by default on ipykernel:
> 
> image
> 
> Could it be related?

I don't think so, the debugger does not fire any stop event.

fcollonval:
This PR is listed in the JLab 3.3 wish list. Are you still thinking it could be completed by the end of January?

echarles:
Let's leave it in the list for now. I am working on UI feedback listed on https://github.com/jupyterlab/jupyterlab/pull/11566#issuecomment-1002179731

Then we can see if we have a fix for the unexpected "side effect".

JohanMabille:
I have experienced the same issue (i.e. code suddenly breaking on selectors.py) with the JupyterLab 3.2.6 and ipykernel 6. The culprit seems to be the "DebugStandardLib" option (which implies "justMyCode = False") added in https://github.com/ipython/ipykernel/pull/812

echarles:
>  have experienced the same issue (i.e. code suddenly breaking on selectors.py) with the JupyterLab 3.2.6 and ipykernel 6. The culprit seems to be the "DebugStandardLib" option (which implies "justMyCode = False") added in ipython/ipykernel#812

You mean without this PR, and vanilla jlab 3.2.6 and ipykernel 6 ?

It is well possible that `"justMyCode = False"` has introduced that effect.
- How can we confirm?
- How to find the place of the root cause?
- If confirmed, we prolly should revert and tackle the root cause wherever it is.

FYI I continue working on this PR to take into account the UI feedbacks listed on https://github.com/jupyterlab/jupyterlab/pull/11566#issuecomment-1002179731.

JohanMabille:
> You mean without this PR, and vanilla jlab 3.2.6 and ipykernel 6 ?

Yes.

> It is well possible that "justMyCode = False" has introduced that effect.

I don't think so, `justMyCode` is false by default if you don't provide any value for it. I'm still investigating, but it might be a bug from Jupyter Lab (it seems that the kernel does not fire any stop event, it simply receives a `stackTrace` request at some point.)

echarles:
> I don't think so, justMyCode is false by default if you don't provide any value for it. 

I am not sure it is False by default. Maybe in the scope of VSCode, but depending on how you launch the debugger services, I remember various path in debugpy that were taking into account env var, and other constructs (don't remember the details as those constructs were not intuitive and not documented).

In the current jlab/kernel/debugpy constructs, the justMyCode was set based on the given `DebugStandardLib` and was true by default, hence the need to add that param to allow to break inside the standard python libraries (e.g. in os.py).

> but it might be a bug from Jupyter Lab (it seems that the kernel does not fire any stopped event, it simply receives a stackTrace request at some point.)

I trust you on this.

JohanMabille:
My bad, I read justMyCode=True (and write the opposite :s). To be precise:

- The argument `['debugOptions'] = [ 'DebugStdLib' ]` is equivalent to `justMyCode=False`
- setting `justMyCode` to `False` leads to the bug described in `https://github.com/ipython/ipykernel/issues/832`. Omitting it (meaning using the deafult value which is `True`) fixes the bug in the context of ipykernel + jupyterlab.
- there is a bug (which seems to come from JupyterLab) as soon as you step into an external file, where the code seems to break somewhere in ipykernel. Setting `justMyCode` to `False` or `True` just changes where the code breaks.

I opened a PR to revert `https://github.com/ipython/ipykernel/pull/812` in order to fix the first issue and I am now investigating the second issue. For the first one, I am affraid that we need to implement a more sophisticated mechanism to send `next` commands ot the debugger until the execution breaks on user code and not kernel code.

echarles:
@mlucool I have updated your comments and ticked the things that have been fixed so far.

Regarding the modules request lifecycle (when do issue that request):

- The current code in iypkernel returns the modules having been loaded so far by the kernel. This means that if the notebook does `import numpy`, numpy will not be shown at the initial module request. I had experimented code like `help(modules)` that loaded all available libraries, but that was quite heavy and time consuming.
- To have the latest and most accurate version of modules, we could issue the modules request on each cell run, but that would be overkill
- A middle ground approach is adding a refresh button near the (to be implemented) filter box.

mlucool:
> returns the modules having been loaded so far by the kernel

That seem right

> A middle ground approach is adding a refresh button near the (to be implemented) filter box.

That seems fine, but maybe if the notebook had any cells run since the last update also have a very slow poll (10s? 30s?)?

echarles:
I have removed the draft status. The latest commits ensure the list is loaded from the kernel on session load and removed on session unload. A refresh button is added and the request is only issued if the kernel supports the modulesRequest.

I still need to add a filter, but as we'd like this to go in 3.3, it is not too early to ask review and feedbacks.

Running with ipykernel latest main branch works fine (except that you can not break in the stdlib source files, which is expected until we find a fix for that - You can still break in any user library, like eg numpy...).

echarles:
> I still need to add a filter, 

Done as a simple string setting. I will convert it to a regexp.

echarles:
Screenshot 2022-01-18 at 08 42 58

mlucool:
> How about we add a regexp based filter and allow a user setting so that we can set a good default (e.g. we may set ours to be deshaw)

I imagined it to look more like:
image

The setting would set the default value of that filter.

echarles:
Adding the ui filter in the toolbar will be imho difficult because of [1] the heigh is small (or we need to lower the font, or make the toolbar heigh bigger, but would break the overall look-and-feel) and [2] the toolbar allows for now to add icons, but not widgets (this can be changed).

What about adding that ui filter as first item of the list? (so at the place of `IPython.core.alias` in the previous image)?

mlucool:
> What about adding that ui filter as first item of the list? (so at the place of IPython.core.alias in the previous image)?

If that gives the user a quick filter which works similar to the file filter, I suspect that will be fine.

echarles:
This is a filter box added in the kernel sources panel. WDYT?

Untitled

echarles:
I will wait a few more days before fixing the CI. Happy to get reviews on this one.

packages/debugger-extension/src/index.ts
```diff
@@ -537,6 +537,9 @@ const sidebar: JupyterFrontEndPlugin<IDebugger.ISidebar> = {
         if (kernel && filters[kernel]) {
           sidebar.variables.filter = new Set<string>(filters[kernel]);
         }
+        const kernelSourcesFilter = setting.get('kernelSourcesFilter')
```
fcollonval, line 540:
```suggestion
        const kernelSourcesFilter = setting.get('defaultKernelSourcesFilter')
```

packages/debugger/src/panels/kernelSources/filter.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * The class name added to the filebrowser crumbs node.
+ */
+export interface IFilterBoxProps {
+  model: IDebugger.Model.IKernelSources;
+  filter: string;
+}
+
+const FilterBox = (props: IFilterBoxProps) => {
```
fcollonval, line 18: The source filter truth should only be in one place (and preferably in the model).

It would better to define the filter box with props: `filter` and `onFilterChange`. And create a ReactWidget like:

```ts
<UseSignal signal={props.model.changed} initArgs={model.filter}>
    {(model) => <FilterBox filter={model.filter} onFilterChange={(e) => {model.filter = (e.target as HTMLInputElement).value}} />}
</UseSignal>
```

packages/debugger/src/panels/kernelSources/body.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  set filter(filter: string) {
+    (this.layout as PanelLayout).removeWidget(this._filenameSearcher);
+    this._filenameSearcher = KernelSourcesFilter({
+      model: this._model,
+      filter: filter
+    });
+    this._filenameSearcher.addClass(FILTERBOX_CLASS);
+    (this.layout as PanelLayout).insertWidget(0, this._filenameSearcher);
+  }
```
fcollonval, line 83: This could be removed (see filter.tsx)

packages/debugger/src/panels/kernelSources/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    );
+
+    this.addClass('jp-DebuggerKernelSources-header');
+    this.addWidget(this._body);
+    this.addClass('jp-DebuggerKenelSources');
+  }
+
+  public set filter(filter: string) {
+    this._model.filter = filter;
+    this._body.filter = filter;
```
fcollonval, line 56: This could be removed

```suggestion
```

packages/debugger/src/panels/kernelSources/model.ts
```diff
@@ -0,0 +1,114 @@
...
+      this._filteredKernelSources.sort(compare);
+    } else {
+      this._kernelSources = new Array<IDebugger.KernelSource>();
+      this._filteredKernelSources = new Array<IDebugger.KernelSource>();
+    }
+    this._changed.emit(this._filteredKernelSources);
+  }
+
+  private _kernelSources: IDebugger.KernelSource[] | null;
+  private _filteredKernelSources: IDebugger.KernelSource[] | null;
```
fcollonval, line 92: Lets explicitly set the initial value.

```suggestion
  private _kernelSources: IDebugger.KernelSource[] | null = null;
  private _filteredKernelSources: IDebugger.KernelSource[] | null = null;
```

fcollonval (changes requested):
Thanks @echarles 

I have some suggestions to improve the code.

echarles:
Thx a lot @fcollonval for the review. I have merged your changes and will implement your suggestions tomorrow 👍

packages/debugger/src/session.ts
```diff
@@ -32,6 +34,13 @@ export class DebuggerSession implements IDebugger.ISession {
     return this._isDisposed;
   }
 
+  /**
+   * Returns the initialize response .
+   */
+  get initializeResponse(): DebugProtocol.InitializeResponse | null {
```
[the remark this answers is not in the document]
echarles: Done. Thx!

packages/debugger/src/panels/kernelSources/filter.tsx
```diff
@@ -0,0 +1,47 @@
...
+    props.model.filter = target.value;
+  };
+
+  return (
+    <InputGroup
+      type="text"
+      rightIcon="ui-components:search"
+      onChange={handleChange}
+      value={filter}
+    />
```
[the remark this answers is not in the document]
echarles: done. thx.

packages/debugger/src/panels/kernelSources/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+      new ToolbarButton({
+        icon: refreshIcon,
+        onClick: async (): Promise<void> => {
+          this._model.kernelSources = [];
+          service.displayModules();
+        },
+        tooltip: trans.__('Refresh kernel sources')
+      })
+    );
+
```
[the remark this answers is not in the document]
echarles: Done. Thx.

packages/debugger/src/panels/kernelSources/filter.tsx
echarles: Done. Thx!

packages/debugger/src/panels/kernelSources/body.ts
echarles: Done. Thx!

echarles:
> Thx a lot @fcollonval for the review. I have merged your changes and will implement your suggestions tomorrow 👍

@fcollonval My latest commits implements your reviews.

packages/debugger/src/service.ts
```diff
@@ -332,6 +332,22 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+    if (!this.session) {
+      throw new Error('No active debugger session');
+    }
+
+    const modules = await this.session.sendRequest('modules', {});
+    this._model.kernelSources.kernelSources = modules.body.modules.map(
+      module => {
+        return {
+          name: Object.keys(module)[0],
+          path: Object.values(module)[0]
```
fcollonval, line 345: I was not able to test it locally with [commit 57a0fce](https://github.com/datalayer-externals/jupyter-kernel-python/tree/feat/debugModulesRequest) because the reply is a list of dictionaries. So this leads to name equals `id` and path equals _<id value>_.

Example of reply:

```json5
{
	"body": {
		"modules": [
			{
				"id": 0,
				"name": "__main__",
				"path": "/tmp/ipykernel_51649/730864761.py",
				"package": "None"
			},
			{
				"id": 1,
				"name": "IPython.core.interactiveshell",
				"path": ".../lib/python3.9/site-packages/IPython/core/interactiveshell.py",
				"package": "IPython.core"
			},
			{
				"id": 2,
				"name": "IPython.core.async_helpers",
				"path": ".../lib/python3.9/site-packages/IPython/core/async_helpers.py",
				"package": "IPython.core"
			},
			{
				"id": 3,
				"name": "ipykernel.zmqshell",
				"path": ".../lib/python3.9/site-packages/ipykernel/zmqshell.py",
				"package": "ipykernel"
			},
                        // ...
		]
	}
}
```

fcollonval (commented):
Thanks for applying the suggestions @echarles 

I tried it locally but I get a mismatch between the reply format of the _modules_ request and what this PR is expected.

packages/debugger/src/panels/kernelSources/body.ts
```diff
@@ -0,0 +1,117 @@
...
+          });
+          button.node.addEventListener('dblclick', () => {
+            this._debuggerService
+              .getSource({
+                sourceReference: 0,
+                path: path
+              })
+              .then(source => {
+                this._model.open(source);
+              });
```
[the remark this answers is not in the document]
echarles: Commited. Thx.

packages/debugger/src/panels/kernelSources/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    this.toolbar.addItem(
+      'refresh',
+      new ToolbarButton({
+        icon: refreshIcon,
+        onClick: async (): Promise<void> => {
+          this._model.kernelSources = [];
+          service.displayModules();
```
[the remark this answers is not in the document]
echarles: Implemented. Thx.

packages/debugger/src/service.ts
echarles: I have double checked and it works fine on my env. Can you install ipykernel with `pip install git+https://github.com/datalayer-externals/jupyter-kernel-python@feat/debugModulesRequest#egg=ipykernel` and try again?

echarles:
This is now rebased on master.

packages/debugger/src/service.ts
fcollonval: :+1: it works

fcollonval:
This is looking good. The UI tests failing are relevant as the accordion panel items are a bit smaller due to the new panel.

echarles:
@fcollonval  UI test is now green.
The 2 failing tests are expected check release failures tracked in https://github.com/jupyterlab/jupyterlab/issues/11933

fcollonval (approved):
Thanks @echarles

echarles:

echarles:
Thx @fcollonval and @JohanMabille for your continuous support on this one. I will manually backport to 3.3.

packages/debugger/src/service.ts
krassowski: Following up on this in https://github.com/jupyterlab/jupyterlab/issues/11966. Two thoughts:
- if we can support the old format (=do not require users to upgrade ipykernel etc), let's support it
- is the new format really valid in DAP?

Issue #11563, opened 2021-11-27
Check Release failing on CI - unpin `vega-embed` dependency


jtpio:
## Description

One more failing check left on CI.

The Check Release is now failing, for example this run on the latest commit: https://github.com/jupyterlab/jupyterlab/runs/4342119386?check_suite_focus=true

Which is picking up the latest `vega-embed@6.20.4`: 

![image](https://user-images.githubusercontent.com/591645/143715417-41cdb2b5-6738-4493-aafa-0693a829da1e.png)

This PR sounds related: https://github.com/vega/vega-embed/pull/802

fcollonval:
I opened an issue upstream: https://github.com/vega/vega-embed/issues/806

fcollonval:
Upstream issue has been fixed and a new vega-embed version has been published - I opened #11578 to revert the temporary fix.

PR #11571, opened 2021-11-29
Temporary fix for release checker pinning vega-embed

fcollonval:
## References

#11563

## Code changes

Forbid using version of vega-embed more recent than 6.20.2

## User-facing changes

N/A

N/A

### DIFF
--- a/packages/vega5-extension/package.json
+++ b/packages/vega5-extension/package.json
@@ -36,7 +36,7 @@
     "@lumino/coreutils": "^1.10.0",
     "@lumino/widgets": "^1.28.0",
     "vega": "^5.20.0",
-    "vega-embed": "^6.2.1",
+    "vega-embed": "6.2.1-6.20.2",
     "vega-lite": "^5.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


jtpio:
Nice, thanks @fcollonval. CI is now green.

According to https://github.com/vega/vega-embed/issues/806, this looks like an issue that should be fixed upstream and not here in the `jupyterlab` repo? If so then we can then keep the temporary pin.

fcollonval:
I agree it should be fixed upstream. But I don't know how quick they will figure out the way forward.

jtpio:
Probably the switch to `npx` prevents looking up the `node_modules` from the top-level monorepo.

jtpio:
OK let's get that one in for now to have a green CI and not block the merge of other PRs.

I have updated the title of https://github.com/jupyterlab/jupyterlab/issues/11563 so keep track of the pin

jtpio:

fcollonval:
Fixed upstream (and reverted in #11578) before backport so no need to backport

PR #11575, opened 2021-11-29
Fix markdown benchmark snapshot

fcollonval:
## References

Follow-up #11519
In the mentionned PR, the collapser button is now always displayed for active cell (for better discovery). That changes requires the benchmark snapshot to be updated.

## Code changes

Fix syntax in benchmark test
Support running not enough tests for statistical distribution computation in benchmark tests.
Run the benchmark tests once as part of the UI tests to detect a snapshot regression.

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

PR #11580, opened 2021-11-30
Explicitly build JupyterLab in dev-mode

fcollonval:
Otherwise it was hidden in the waiting time for JupyterLab to be available

## References

Follow-up #11575

## Code changes

Explicitly call `jlpm run build` on UI tests so the dev version is not built indirectly when waiting for JupyterLab server to be available.

## User-facing changes

N/A

N/A


fcollonval:

PR #11537, opened 2021-11-24
Clean up Frontend Typings

jtpio:
## References

This is a continuation of #11196.

## Code changes

- [x] Allow alternate `ServiceManager` implementations in `JupyterFrontEnd`.  Specifically this allows the use of a mock manager.
- [x] Make the `JupyterLab.IOptions` more permissive: do not require a shell when options are given and allow a shell that meets the `ILabShell` interface
- [x] Change the `language` used in the mock `irkernel` to get some diversity in the kernel data for testing purposes
- [x] Document breaking changes in https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-3-x-to-4-x. **edit**: docs preview available here: https://jupyterlab--11537.org.readthedocs.build/en/11537/extension/extension_migration.html#api-breaking-changes

## User-facing changes

None

## Backwards-incompatible changes

Consumers relying on the concrete `ServiceManager` have to be updated.

PR #10993, opened 2021-09-02
Update to TypeScript 4.4

jtpio:
## References

Try updating to the latest TypeScript stable version.

Blog post: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/

## Code changes

Dependency update.

## User-facing changes

None

## Backwards-incompatible changes

Probably declarations have changed, so good to do the update for 4.0.

PR #11594, opened 2021-11-30
Update to TypeScript 4.5

jtpio:
## References

Reviving the effort started in https://github.com/jupyterlab/jupyterlab/pull/10993 since we are a couple of versions behind now.

## Code changes

TypeScript update:

- [x] ~Switching to another polyfill for `resize-observer` based on the discussions in https://github.com/que-etc/resize-observer-polyfill/issues/83~ -> **edit**: drop the polyfill based on the discussion in https://github.com/jupyterlab/jupyterlab/pull/11594#pullrequestreview-819423470
- [x] `ClientRect` has been removed in 4.4: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029#issuecomment-869224737
- [x] Update `CodeEditor.ICoordinate` interface
- [x] More `null` checks caught by the compiler
- [x] ~Bump major version of the packages affected by the breaking changes~ -> done in https://github.com/jupyterlab/jupyterlab/pull/11537
- [x] Document interface changes

## User-facing changes

None

- `lib.d.ts`
- Update `CodeEditor.ICoordinate` interface

### DIFF
--- a/packages/outputarea/package.json
+++ b/packages/outputarea/package.json
@@ -63,7 +63,7 @@
     "rimraf": "~3.0.0",
     "ts-jest": "^26.3.0",
     "typedoc": "~0.21.2",
-    "typescript": "~4.1.3"
+    "typescript": "~4.5.2"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/outputarea/package.json
+++ b/packages/outputarea/package.json
@@ -54,7 +54,7 @@
     "@lumino/properties": "^1.8.1",
     "@lumino/signaling": "^1.10.1",
     "@lumino/widgets": "^1.30.0",
-    "resize-observer-polyfill": "^1.5.1"
+    "resize-observer": "^1.0.4"
   },
   "devDependencies": {
     "@jupyterlab/testutils": "^4.0.0-alpha.1",
--- a/packages/outputarea/package.json
+++ b/packages/outputarea/package.json
@@ -53,8 +53,7 @@
     "@lumino/messaging": "^1.10.1",
     "@lumino/properties": "^1.8.1",
     "@lumino/signaling": "^1.10.1",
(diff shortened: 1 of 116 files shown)


packages/outputarea/package.json
```diff
@@ -54,7 +54,7 @@
     "@lumino/properties": "^1.8.1",
     "@lumino/signaling": "^1.10.1",
     "@lumino/widgets": "^1.30.0",
-    "resize-observer-polyfill": "^1.5.1"
+    "resize-observer": "^1.0.4"
```
jtpio, line 57: Switching to another polyfill based on the discussions in https://github.com/que-etc/resize-observer-polyfill/issues/83

jtpio:
CI is now passing. The remaining failing check is expected (more than 100 GitHub labels to apply).

jtpio:
>  Bump major version of the packages affected by the breaking changes

Gonna check if we can get https://github.com/jupyterlab/jupyterlab/pull/11537 in first, which affects all packages and already bumps the major versions.

fcollonval (approved):
Thanks @jtpio

packages/outputarea/package.json
fcollonval: We could probably drop the polyfill, no?

https://caniuse.com/?search=ResizeObserver

jtpio: Yes that would be great.

I left it for now just to check if we have any commitment related to the minimal browser versions, and which browsers we support.

jtpio: The docs say: https://jupyterlab.readthedocs.io/en/stable/getting_started/installation.html?#supported-browsers

image

So we could indeed just drop the polyfill.

fcollonval:
Let's merge

PR #11599, opened 2021-12-01
Save document from the backend using y-py

davidbrochart:
## Code changes

When collaborative mode is enabled, the save command still goes through the HTTP endpoint, but the document source is computed in the back-end using `y-py`.
When collaborative mode is disabled, the previous behavior is kept: the document content is sent from the front-end to the back-end through the HTTP endpoint.
The back-end raises an error if:
- the document content is sent from the front-end through the HTTP endpoint and there is a Yjs shared "room" corresponding to that document,
- no document content is sent from the front-end through the HTTP endpoint and no Yjs shared "room" corresponds to that document.

Otherwise, the back-end gets the document content either from the HTTP endpoint or computes it from the Yjs shared "room" corresponding to that document.

cc @dmonad

None

### DIFF
--- a/jupyterlab/handlers/yjs_echo_ws.py
+++ b/jupyterlab/handlers/yjs_echo_ws.py
@@ -7,10 +7,20 @@ import time
 import uuid
 from enum import IntEnum
 
+import pkg_resources
+import y_py as Y
 from jupyter_server.base.handlers import JupyterHandler
 from tornado import web
 from tornado.ioloop import IOLoop
 from tornado.websocket import WebSocketHandler
+
+YDOCS = {}
+for ep in pkg_resources.iter_entry_points(group="jupyter_ydoc"):
+    YDOCS.update({ep.name: ep.load()})
+
+YFILE = YDOCS["file"]
+
+ROOMS = {}
 
 ## The y-protocol defines messages types that just need to be propagated to all other peers.
 ## Here, we define some additional messageTypes that the server can interpret.
@@ -32,16 +42,20 @@ class ServerMessageType(IntEnum):
 
 
 class YjsRoom:
-    def __init__(self):
+    def __init__(self, type):
+        self.type = type
         self.lock = None
         self.timeout = None
         self.lock_holder = None
         self.clients = {}
         self.content = bytes([])
+        self.ydoc = YDOCS.get(type, YFILE)()
+
+    def get_source(self):
+        return self.ydoc.source
 
 
 class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
-    rooms = {}
 
     # Override max_message size to 1GB
     @property
@@ -54,24 +68,23 @@ class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
             raise web.HTTPError(403)
         return await super().get(*args, **kwargs)
 
-    def open(self, guid):
-        # print("[YJSEchoWS]: open", guid)
-        cls = self.__class__
+    def open(self, type_path):
+        # print("[YJSEchoWS]: open", type_path)
+        type, path = type_path.split(":", 1)
         self.id = str(uuid.uuid4())
-        self.room_id = guid
-        room = cls.rooms.get(self.room_id)
+        self.room_id = path
+        room = ROOMS.get(self.room_id)
         if room is None:
-            room = YjsRoom()
-            cls.rooms[self.room_id] = room
+            room = YjsRoom(type)
+            ROOMS[self.room_id] = room
         room.clients[self.id] = (IOLoop.current(), self.hook_send_message, self)
         # Send SyncStep1 message (based on y-protocols)
         self.write_message(bytes([0, 0, 1, 0]), binary=True)
 
     def on_message(self, message):
-        # print("[YJSEchoWS]: message, ", message)
-        cls = self.__class__
+        # print("[YJSEchoWS]: message,", message)
         room_id = self.room_id
-        room = cls.rooms.get(room_id)
+        room = ROOMS.get(room_id)
         if message[0] == ServerMessageType.ACQUIRE_LOCK:
             now = int(time.time())
             if room.lock is None or now - room.timeout > (
@@ -80,7 +93,7 @@ class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
                 room.lock = now
                 room.timeout = now
                 room.lock_holder = self.id
-                # print('Acquired new lock: ', room.lock)
+                # print('Acquired new lock:', room.lock)
                 # return acquired lock
                 self.write_message(
                     bytes([ServerMessageType.ACQUIRE_LOCK])
@@ -89,14 +102,14 @@ class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
                 )
 
             elif room.lock_holder == self.id:
-                # print('Update lock: ', room.timeout)
+                # print('Update lock:', room.timeout)
                 room.timeout = now
 
         elif message[0] == ServerMessageType.RELEASE_LOCK:
             releasedLock = int.from_bytes(message[1:], byteorder="little")
-            # print("trying release lock: ", releasedLock)
+            # print("trying release lock:", releasedLock)
             if room.lock == releasedLock:
-                # print('released lock: ', room.lock)
+                # print('released lock:', room.lock)
                 room.lock = None
                 room.timeout = None
                 room.lock_holder = None
@@ -110,24 +123,27 @@ class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
             room.content = message[1:]
         elif message[0] == ServerMessageType.RENAME_SESSION:
             # We move the room to a different entry and also change the room_id property of each connected client
-            new_room_id = message[1:].decode("utf-8")
+            new_room_id = message[1:].decode("utf-8").split(":", 1)[1]
             for client_id, (loop, hook_send_message, client) in room.clients.items():
                 client.room_id = new_room_id
-            cls.rooms.pop(room_id)
-            cls.rooms[new_room_id] = room
+            ROOMS.pop(room_id)
+            ROOMS[new_room_id] = room
+            # send rename acknowledge
+            self.write_message(bytes([ServerMessageType.RENAME_SESSION, 1]), binary=True)
             # print("renamed room to " + new_room_id + ". Old room name was " + room_id)
         elif room:
+            if message[0] == 0:  # sync message
+                read_sync_message(self, room.ydoc.ydoc, message[1:])
             for client_id, (loop, hook_send_message, client) in room.clients.items():
                 if self.id != client_id:
                     loop.add_callback(hook_send_message, message)
 
     def on_close(self):
         # print("[YJSEchoWS]: close")
-        cls = self.__class__
-        room = cls.rooms.get(self.room_id)
+        room = ROOMS.get(self.room_id)
         room.clients.pop(self.id)
-        if len(room.clients) == 0:
-            cls.rooms.pop(self.room_id)
+        if not room.clients:
+            ROOMS.pop(self.room_id)
             # print("[YJSEchoWS]: close room " + self.room_id)
 
         return True
@@ -138,3 +154,71 @@ class YjsEchoWebSocket(WebSocketHandler, JupyterHandler):
 
     def hook_send_message(self, msg):
         self.write_message(msg, binary=True)
+
+
+message_yjs_sync_step1 = 0
+message_yjs_sync_step2 = 1
+message_yjs_update = 2
+
+
+def read_sync_step1(handler, doc, encoded_state_vector):
+    message = Y.encode_state_as_update(doc, encoded_state_vector)
+    message = bytes([message_yjs_sync_step2] + write_var_uint(len(message)) + message)
+    handler.write_message(message, binary=True)
+
+
+def read_sync_step2(doc, update):
+    try:
+        Y.apply_update(doc, update)
+    except Exception:
+        raise RuntimeError("Caught error while handling a Y update")
+
+
+def read_sync_message(handler, doc, message):
+    message_type = message[0]
+    message = message[1:]
+    if message_type == message_yjs_sync_step1:
+        for msg in get_message(message):
+            read_sync_step1(handler, doc, msg)
+    elif message_type == message_yjs_sync_step2:
+        for msg in get_message(message):
+            read_sync_step2(doc, msg)
+    elif message_type == message_yjs_update:
+        for msg in get_message(message):
+            read_sync_step2(doc, msg)
+    else:
+        raise RuntimeError("Unknown message type")
+
+
+def write_var_uint(num):
+    res = []
+    while num > 127:
+        res += [128 | (127 & num)]
+        num >>= 7
+    res += [num]
+    return res
+
+
+def get_message(message):
+    length = len(message)
+    i0 = 0
+    while True:
+        msg_len = 0
+        i = 0
+        while True:
+            byte = message[i0]
+            msg_len += (byte & 127) << i
+            i += 7
+            i0 += 1
+            length -= 1
+            if byte < 128:
+                break
+        i1 = i0 + msg_len
+        msg = message[i0:i1]
+        length -= msg_len
+        yield msg
+        if length <= 0:
+            if length < 0:
+                raise RuntimeError("Y protocol error")
+            break
+        i0 = i1
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -20,6 +20,7 @@ from jupyterlab_server import (
     WorkspaceListApp,
 )
 from notebook_shim.shim import NotebookConfigShimMixin
+from tornado import web
 from traitlets import Bool, Instance, Unicode, default
 
 from ._version import __version__
@@ -49,7 +50,7 @@ from .handlers.extension_manager_handler import (
     ExtensionManager,
     extensions_handler_path,
 )
-from .handlers.yjs_echo_ws import YjsEchoWebSocket
+from .handlers.yjs_echo_ws import ROOMS, YjsEchoWebSocket
 
 DEV_NOTE = """You're running JupyterLab from source.
 If you're working on the TypeScript sources of JupyterLab, try running
@@ -628,6 +629,30 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
             self.template_paths = [self.templates_dir]
 
     def initialize_settings(self):
+        def hook(model, path, **kwargs):
+            if "type" in model and model["type"] == "directory":
+                pass
+            elif "content" in model and model["content"] is not None:
+                # content sent through HTTP, it must not be an RTC session
+                if path in ROOMS:
+                    raise web.HTTPError(
+                        409,
+                        "Document content cannot be present both in RTC session and HTTP request",
+                    )
+                # keep the content sent through HTTP
+            else:
+                # no content sent through HTTP, it must be an RTC session
+                if path in ROOMS:
+                    # we found the RTC session as expected
+                    # set the document content from y-py
+                    model["content"] = ROOMS[path].get_source()
+                else:
+                    # RTC session not available, shouldn't happen
+                    raise web.HTTPError(
+                        410, "Could not find an RTC session corresponding to this document"
+                    )
+
+        self.serverapp.contents_manager.register_pre_save_hook(hook)
         super().initialize_settings()
 
     def initialize_handlers(self):
--- a/packages/docprovider/src/tokens.ts
+++ b/packages/docprovider/src/tokens.ts
@@ -36,6 +36,12 @@ export interface IDocumentProvider {
   releaseLock(lock: number): void;
 
   /**
+   * Returns a Promise that resolves when renaming is ackownledged.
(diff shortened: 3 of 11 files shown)


jtpio:
Thanks @davidbrochart for exploring this.

packages/docprovider/src/tokens.ts
```diff
@@ -22,6 +22,11 @@ export interface IDocumentProvider {
    */
   putInitializedState(): void;
 
+  /**
+   * Put the document source.
+   */
+  putSource(source: string): void;
```
jtpio, line 28: Since this adds a new (non-optional) method to the `IDocumentProvider` interface, I think this should be considered as a backward incompatible change.

Would you mind adding that to the "Backwards-incompatible changes" section of the PR template so we can keep track of it? Thanks!

davidbrochart: Sure, although we could make the method optional, and fall back to the previous behavior if it doesn't exist?
Also, this PR:
- makes it mandatory to be in collaborative mode, but I guess we could support non-collaborative mode too.
- requires a change in the server. I made the change [in jupyverse](https://github.com/jupyter-server/jupyverse/pull/127), but it should also be done in `jupyter-server`.

dmonad: I think that we can remove this method eventually as we will be able to compute the source using y-py. This is just meant as an intermediate step, right?

The HTTP API shouldn't be able to overwrite the current RTC state. Hence, the PUT request shouldn't contain any content when RTC is enabled and simply grab the source from the RTC server and then save it to disk.

To maintain a bit of backwards compatibility we could still make the `content` value in the PUT request optional:

* If there currently exists an RTC session, and `content` is empty. Calculate the source based on the RTC state and save it to disk (for now we will use the stored `source` value that David implemented here).
* If there currently exists an RTC session, and `content` is **not** empty. Return a 409 error code.
* If there is no active RTC session for that room, and `content` is **not** empty. Save the `content` as usual.
* Lastly. if there is no active RTC session, and content is not empty, return an error code. This scenario should never happen as the sending client is clearly an RTC client, but for some reason the RTC session is not available anymore. Maybe a 410 might be appropriate.

packages/docregistry/src/context.ts
```diff
@@ -595,23 +595,26 @@ export class Context<
...
     const options = {
       type: this._factory.contentType,
-      format: this._factory.fileFormat,
-      content
+      format: this._factory.fileFormat
     };
     try {
       let value: Contents.IModel;
       await this._manager.ready;
+      this._provider.putSource(source);
```
dmonad, line 617: We have to be aware that the websocket message might arrive after the HTTP PUT request (so you will safe the old content). But since this is just an intermediate step, this is fine (although not production-ready yet).

dmonad (commented):
Thank you David, this is great! 

This gives us a good baseline to integrate y-py. I'm mildly concerned about HTTP requests arriving before the websocket messages (see my comment). So let's keep this PR in the draft state until we integrated y-py.

packages/docprovider/src/tokens.ts
davidbrochart: For the last point, do you mean "if there is no active RTC session, and content is empty"?

davidbrochart:
This last commit implements saving the document from the back-end entirely, by computing the document source using `y-py`. I included the changes from #11698, although that PR will change. Also, `y-py` is not installed, but I thought I'd share the code anyway, so people can have a look at how it looks like.

davidbrochart:
In this last commit we set the model content from the RTC session or the HTTP endpoint, as discussed by @dmonad in https://github.com/jupyterlab/jupyterlab/pull/11599#discussion_r766690940.
Will need https://github.com/jupyter-server/jupyter_server/pull/643 to be merged.

meeseeksmachine:

https://discourse.jupyter.org/t/plans-on-bringing-rtc-to-jupyterhub/9813/14

davidbrochart:
This PR is now ready for review.
cc @dmonad

jupyterlab/labapp.py
```diff
@@ -586,6 +588,23 @@ def initialize_templates(self):
 
 
     def initialize_settings(self):
+        def hook(model, path, **kwargs):
```
blink1073, line 591: Nice!

davidbrochart:
I updated [the description of the PR](https://github.com/jupyterlab/jupyterlab/pull/11599#issue-1068406324).

davidbrochart:
@dmonad do you think we can merge this PR?

damianavila:
> I updated the description of the PR.

@davidbrochart was there any previous discussion about the sequential approach you are taking here?
It would be nice (for people looking from far away like me 😉) to get some info, in the top message, about future steps to persist the document content when collaborative is not enabled (maybe a link to a previous discussion is just enough).
Thanks!!

davidbrochart:
> @davidbrochart was there any previous discussion about the sequential approach you are taking here?

Do you mean the fact that we look if the content is sent through HTTP or if there is an RTC session for the document? If so, I don't think there has been any previous discussion about that, except maybe [this comment](https://github.com/jupyterlab/jupyterlab/pull/11599#discussion_r766690940).

> It would be nice (for people looking from far away like me wink) to get some info, in the top message, about future steps to persist the document content when collaborative is not enabled (maybe a link to a previous discussion is just enough).

I'm not sure to understand, when collaborative is disabled we do as before, we pass the content through the HTTP endpoint.
Maybe you mean something else?

davidbrochart:
Kindly pinging @dmonad.

setup.cfg
```diff
@@ -36,9 +36,10 @@ install_requires =
     tornado>=6.1.0
     jupyter_core
     jupyterlab_server~=2.10
-    jupyter_server~=1.4
+    jupyter_server>=1.13.4
```
jtpio, line 39: Let's keep the upper-bound, for example?

```suggestion
    jupyter_server~=1.13
```

SylvainCorlay:
As discussed in person, I think that we should replace this:

```python
if type == "file":
elif type == "notebook":
```

with a dictionary lookup (`y[de]serializers.get(type, default_[de]serializer)`), in which [de]serializers for new document types can be registered with an entry point.

Ideally, we should be able to reuse the same entry points in jupyverse...

setup.cfg
jtpio: We can also switch to `jupyter_server>=1.13.4,<2` :+1:

davidbrochart:
Not sure about [the CI failures](https://github.com/jupyterlab/jupyterlab/runs/5021462944?check_suite_focus=true#step:5:8176) ("TypeError: Cannot read property 'scheduledCellNumber' of undefined").

hbcarlos:
I pushed a commit to solve the tests.

davidbrochart:
Thanks!

davidbrochart:
OK to merge?

setup.cfg
```diff
@@ -36,9 +36,10 @@ install_requires =
     tornado>=6.1.0
     jupyter_core
     jupyterlab_server~=2.10
-    jupyter_server~=1.4
+    jupyter_server>=1.13.4,<2
     nbclassic~=0.2
     jinja2>=2.1
+    y-py>=0.2.2
```
jtpio, line 42: Maybe we should add an upper bound here too?

jupyterlab/labapp.py
```diff
@@ -586,6 +588,23 @@ def initialize_templates(self):
...
+                # no content sent through HTTP, it must be an RTC session
+                if path in ROOMS:
+                    # we found the RTC session as expected
+                    # set the document content from y-py
+                    model["content"] = ROOMS[path].get_source()
+                else:
+                    # RTC session not available, shouldn't happen
+                    raise web.HTTPError(410, "Could not find an RTC session corresponding to this document")
+
+        self.serverapp.contents_manager.pre_save_hook = hook
```
vidartf, line 607: Will this not conflict with other users of this hook?

vidartf: I.e. maybe we should try to create a more flexible hook system now while this is still fresh, so that we don't have jupyter_hub, lab, and X other extensions all fighting for the `pre_save_hook slot`.

davidbrochart: Good point, maybe jupyter_server could implement a setter for the hooks, that would append them to a list of hooks under the hood? All hooks of the list would then be called. That has the advantage of keeping the same API.
I can open a PR there if that seems reasonable.

damianavila:
> I'm not sure to understand, when collaborative is disabled we do as before, we pass the content through the HTTP endpoint.
Maybe you mean something else?

I am probably misunderstanding things and probably talking without enough context, but let me try 😉 

> about future steps to persist the document content when collaborative is not enabled

Why this PR is conditional (besides backward compatibility)? I mean, why we are not doing the collaborative mode "mandatory" if that brings us a server-side save for free? People have been asking for that for a long time AFAIK... Is there any penalization in terms of performance preventing this from happening? Something else I should understand first?

Thanks!!

davidbrochart:
Thanks for clarifying @damianavila.
I don't think the aim of this PR is to make collaborative editing mandatory. This decision will probably be made once we have enough added value in RTC support, and enough confidence in it, but right now I guess it is still a work in progress as we are figuring out things. In particular, RTC will become much more useful when we have authentication and permission support (see [this PR in jupyter_server](https://github.com/jupyter-server/jupyter_server/pull/671) and [that PR in JupyterLab](https://github.com/jupyterlab/jupyterlab/pull/11443)).

davidbrochart:
Trigger CI.

davidbrochart:
Trigger CI

jtpio:
For reference `y-py` has now been packaged for conda:

- https://github.com/conda-forge/staged-recipes/pull/17076
- https://anaconda.org/conda-forge/y-py

davidbrochart:
Trigger CI.

davidbrochart:
Ready to merge!

davidbrochart:
The CI failures are caused by https://github.com/jupyterlab/jupyterlab_server/pull/239#discussion_r828082056.

afshin:
Otherwise, this is a :+1: from me!

davidbrochart:
The "Test Minimum Versions" workflow fails because the required version of jupyterlab-server [from jupyterlab/maintainer-tools](https://github.com/jupyterlab/maintainer-tools/blob/main/.github/actions/install-minimums/action.yml) is `2.10`, and we need `2.11.1` here.
Should we update maintainer-tools @blink1073 ?

davidbrochart:
Ah, I see it's just an inconsistency between the install requirements and the test requirements here, I will update.

davidbrochart:
The CI is green again, let's merge!

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -101,30 +114,94 @@ def on_message(self, message):
...
     def hook_send_message(self, msg):
         self.write_message(msg, binary=True)
+
+message_yjs_sync_step1 = 0
+message_yjs_sync_step2 = 1
+message_yjs_update = 2
+
+def read_sync_step1(handler, doc, encoded_state_vector):
+    message = Y.encode_state_as_update(doc, encoded_state_vector)
+    message = bytes([message_yjs_sync_step2] + get_len(message) + message)
```
dmonad, line 152: This is not correct. The length of the message must be encoded using variable-length encoding. I sent you some sample code earlier.

davidbrochart: Could you be more specific? I believe the length of the message is encoded using variable-length encoding.

jtpio:
Thanks @davidbrochart. That's good progress and it should already help fix some data loss issues folks have been experiencing.

Since it introduces the new `y-py` package and changes the behavior as to how files are saved, maybe we should keep this for `4.0` only and not backport to `3.3.x`?

cc @dmonad who might want to review this too

afshin:
If it is truly backward-compatible, I propose we keep it in 3.x because data loss is a critical bug and it has already burned early RTC users.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -52,43 +66,42 @@ async def get(self, *args, **kwargs):
...
-        #print("[YJSEchoWS]: message, ", message)
-        cls = self.__class__
+        # print("[YJSEchoWS]: message,", message)
         room_id = self.room_id
-        room = cls.rooms.get(room_id)
+        room = ROOMS.get(room_id)
         if message[0] == ServerMessageType.ACQUIRE_LOCK:
             now = int(time.time())
-            if room.lock is None or now - room.timeout > (10 * len(room.clients)) : # no lock or timeout
+            if room.lock is None or now - room.timeout > (10 * len(room.clients)): # no lock or timeout
```
dmonad, line 88: Do we still need room-locks? Or are you keeping them for compatibility sake?

dmonad (changes requested):
Thank you, this is a very solid implementation. I think we might need to resolve a final issue regarding session management to avoid potential duplication of content. 

I do like the idea to get this already into 3.x.

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: Thanks, I will rename this function.

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: Indeed I don't think we need them anymore, I will remove them.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -101,30 +114,94 @@ def on_message(self, message):
...
     def on_close(self):
         # print("[YJSEchoWS]: close")
-        cls = self.__class__
-        room = cls.rooms.get(self.room_id)
+        room = ROOMS.get(self.room_id)
         room.clients.pop(self.id)
-        if len(room.clients) == 0 :
-            cls.rooms.pop(self.room_id)
+        if not room.clients:
+            ROOMS.pop(self.room_id)
```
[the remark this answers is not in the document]
davidbrochart: I would say that this kind of issues already exist (before this PR), right? If you agree, could we tackle them in another PR? I think this PR has been around long enough and it's time to merge it, since it will solve [a critical issue](https://github.com/jupyterlab/jupyterlab/issues/12154).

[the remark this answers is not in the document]
dmonad: Yeah, that's true. Although we did solve *case 1* with the lock approach. However, there are regular disconnects in the beginning of a session. So I suggest that we at least debounce the destruction of the room after all clients disconnected. This shouldn't be much work. WDYT?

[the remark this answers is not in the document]
dmonad: Aside from that, I don't see any issues with this PR.

davidbrochart:
I found an issue with the "save as" command: [an HTTP request is sent to save the document with the new path](https://github.com/jupyterlab/jupyterlab/blob/633f2eea4f7f9ea32c6d418a324172ffe3c0c645/packages/docregistry/src/context.ts#L885), and [a message is sent to the Yjs websocket to rename the document to this new path](https://github.com/jupyterlab/jupyterlab/blob/633f2eea4f7f9ea32c6d418a324172ffe3c0c645/packages/docregistry/src/context.ts#L886) on the next line. Because we need to have the new room ID before actually saving the file, renaming must happen before saving. But because they go through the network and are asynchronous functions, calling the rename function before the save function is not enough: we must ensure that the renaming completes before saving.
https://github.com/jupyterlab/jupyterlab/pull/11599/commits/264f4399d2cd378c3de6fe996175cc346130a269 implements a rename acknowledge, sent by the back-end so that the front-end knows when the renaming has completed.

packages/docprovider/src/yprovider.ts
```diff
@@ -21,15 +21,15 @@ import { IDocumentProvider, IDocumentProviderFactory } from './tokens';
  * We specify custom messages that the server can interpret. For reference please look in yjs_ws_server.
  *
  */
-export class WebSocketProviderWithLocks
-  extends WebsocketProvider
+export class WebSocketProvider
```
jtpio, line 24: This renaming and the removal of `acquireLock` and `releaseLock` from the interfaces make this PR not backward compatible anymore.

davidbrochart: It also makes the tests fail and I don't understand why: https://github.com/jupyterlab/jupyterlab/runs/5597604691?check_suite_focus=true.
Let's keep the room locks for now and remove them in another PR.

davidbrochart:
Trigger CI.

jupyterlab/handlers/yjs_echo_ws.py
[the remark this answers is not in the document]
davidbrochart: Let's do the debouncing in another PR.

davidbrochart:
Not sure the [Visual Regression Documentation](https://github.com/jupyterlab/jupyterlab/runs/5598609300?check_suite_focus=true) failures are related.

jtpio:
> Not sure the [Visual Regression Documentation](https://github.com/jupyterlab/jupyterlab/runs/5598609300?check_suite_focus=true) failures are related.

Looks like these are new and also happening on other PRs and latest commits on master. Tracked in https://github.com/jupyterlab/jupyterlab/issues/12233

davidbrochart:
Then I think it's good to merge.

jtpio:
> Not sure about the failure in [macOS Tests / macOS (python, 3.8)](https://github.com/jupyterlab/jupyterlab/runs/5656976929?check_suite_focus=true):
> 
> ```
> ./scripts/ci_script.sh: line 17: py.test: command not found
> ```

This one should now be fixed by https://github.com/jupyterlab/jupyterlab/pull/12262

packages/docprovider/src/tokens.ts
```diff
@@ -35,6 +35,11 @@ export interface IDocumentProvider {
    */
   releaseLock(lock: number): void;
 
+  /**
+   * Returns a Promise that resolves when renaming is ackownledged.
+   */
+  readonly renameAck: Promise<boolean>;
```
jtpio, line 41: This addition to the new `IDocumentProvider` will probably break downstream implementations if this gets backported to 3.3.x and they update to the new packages, since this field is not optional?

For example JupyterLite implements the interface here: https://github.com/jupyterlite/jupyterlite/blob/41da1cb7ba64708284a07ac381d296466d0f0fdd/packages/application-extension/src/index.tsx#L66

jtpio: We could say this is fine, if we consider there are not that many custom implementations out there yet.

davidbrochart: I think we could make this field optional. It is used by the JupyterLab implementation in this PR but I guess other implementations would not care to know if/when renaming actually completes.

afshin: We could make it optional with a `TODO` to make it required for the 4.0 version?

davidbrochart:
Again.

davidbrochart:
Not sure about this [CI failure](https://github.com/jupyterlab/jupyterlab/runs/5667077054?check_suite_focus=true):
```
./scripts/ci_script.sh: line 301:  3849 Terminated: 15          ./test_install/bin/jupyter lab --no-browser
Error: Process completed with exit code 143.
```

jtpio:
Restarted the CI and it now passes. Looks like this is good to go now.

cc @afshin who might want to have another look. Otherwise I would suggest we merge this and make a new release so it can be more easily tested.

Thanks @davidbrochart for your patience!

hbcarlos:
Thanks for working on this @davidbrochart! LGTM!

cc @jtpio

davidbrochart:
I squashed everything because it was becoming to hard to rebase after the Black formatting.
In order to fix the Black failures [we need to use Black 22.3.0](https://github.com/psf/black/issues/2964):
```
  File "/home/runner/.cache/pre-commit/repo_9lbscd0/py_env-python3.10/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/home/runner/.cache/pre-commit/repo_9lbscd0/py_env-python3.10/lib/python3.10/site-packages/black/__init__.py", line 1423, in patched_main
    patch_click()
  File "/home/runner/.cache/pre-commit/repo_9lbscd0/py_env-python3.10/lib/python3.10/site-packages/black/__init__.py", line 1409, in patch_click
    from click import _unicodefun
ImportError: cannot import name '_unicodefun' from 'click' (/home/runner/.cache/pre-commit/repo_9lbscd0/py_env-python3.10/lib/python3.10/site-packages/click/__init__.py)
```

afshin (approved):
Thank you, @davidbrochart!

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -138,3 +154,71 @@ def check_origin(self, origin):
...
+
+
+message_yjs_sync_step1 = 0
+message_yjs_sync_step2 = 1
+message_yjs_update = 2
+
+
+def read_sync_step1(handler, doc, encoded_state_vector):
+    message = Y.encode_state_as_update(doc, encoded_state_vector)
+    message = bytes([message_yjs_sync_step2] + write_var_uint(len(message)) + message)
```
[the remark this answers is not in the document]
dmonad: Yep, you are right .. It's weird that this didn't throw an error on the client..

Issue #8960, opened 2020-09-04
Jlab could provide a way to select predefined layouts where widgets are always docked together in a certain way


goanpeca:
### Problem

The issue can be phrased like:

Is there a way of having some sort of predefined layouts so I can assign a given type of:
- Documents
- Widget
- Output
- Viewer
- Terminal

 to a fixed position in JLab?

Maybe I want all editors to be tabbed together on the left side, and all terminals tabbed in the bottom and outputs that result in some sort of plot (is this a mimerenderer?) always in the bottom right and the contextual help and csv viewer on the top right.

Having to open JLab and organize it in a certain way every time I open it, and hope that workspaces are going to do the right thing (this is often not the case) can be very frustrating and time-consuming.

![Jlab](https://user-images.githubusercontent.com/3627835/92282066-84f5f880-eec2-11ea-87f1-1c012d455105.gif)

Why did the notebook I opened got tabbed with the terminal? Because it was the active/focused widget?

Having things in a given place is a time tested pattern that can be seen over and over in many IDEs like:

#### Matlab

![image](https://user-images.githubusercontent.com/3627835/92281495-5c213380-eec1-11ea-8699-ddd6535de12b.png)

#### RStudio

![image](https://user-images.githubusercontent.com/3627835/92281515-65aa9b80-eec1-11ea-8032-3a56c9ac432b.png)

#### Spyder
![image](https://user-images.githubusercontent.com/3627835/92281564-778c3e80-eec1-11ea-91cb-2f3c103d4ed9.png)

### Eclipse

![image](https://user-images.githubusercontent.com/3627835/92281604-896de180-eec1-11ea-8003-4aa69a9eb0d2.png)

etc...

### Proposed Solution

Maybe I want to have something like this and call it the Rstudio layout (or something)

![image](https://user-images.githubusercontent.com/3627835/92282277-e027eb00-eec2-11ea-9aea-06d92c94beaa.png)

I could use a GridLayout so something like the image can be expressed as:

```typescript
rstudioLayout = GridLayout();
layout.addArea(row=0, column=0, name= “Top Left”, default=true,  [“notebook”, “editor”], default=true);
layout.addArea(row=0, column=1, name=“Top Right”,  [“contextual_help”]);
layout.addArea(row=1, column=0, name=“Bottom left”, [“console”]);
layout.addArea(row=1, column=1, name=“Bottom Right”, [“terminal”]);

layoutManager.register("RStudio", rstudioLayout);
```

[...]

PR #11608, opened 2021-12-02
Toggle side-by-side rendering for current notebook

echarles:
## References

https://github.com/jupyterlab/jupyterlab/issues/11376

## Code changes

Add a command, a shortcut to toggle the side-by-side rendering only for the current notebook.

## User-facing changes

Untitled

None


fcollonval (commented):
LGTM

One question, would it make sense to be easier to switch on a per notebook basis and more complex to do it globally?
Say otherwise should this new shortcut fall to `Shift R` and the older to `Ctrl Shift R`?

echarles:
> Say otherwise should this new shortcut fall to Shift R and the older to Ctrl Shift R?

Makes a lot sense to me. May also this should be added to the menu, but I fear some kind of `menu congestion`... I will wait a bit more for more feedbacks on these.

Screenshot 2021-12-03 at 10 05 20

fcollonval:
We probably want to keep only the action for the active notebook in the menu as this is already the case for items like _show line numbers_, _match brackets_,...

echarles:
@fcollonval I have pushed the needed change to toogle for the current notebook with `Shift R`.

A few questions.

## 1. Notebook metadata

The notebook reminds the side-by-side state via metadata. For now, after save, a new metadata is persisted. It is not clear to me from https://nbformat.readthedocs.io/en/latest/format_description.html if this is allowed/accepted. I don't know either if we want that. I can see advantages to it, as the next notebook author designs its content to fit a defined presentation, so the next reader get benefit from that persisted stat

Screenshot 2021-12-06 at 08 10 57
e

## 2. RTC

I have not tested with RTC, but I can imagine your notebook rendering suddenly changing if a remote collaborator toggles its side-by-side state. Do we want to prevent that or it is a welcome feature? 

## 3. Menu label

For now, the label in the menu is a bit long, though descriptive `Toggle Side-by-side rendering for current Notebook`. Any comment on that? 

Screenshot 2021-12-06 at 08 11 56

fcollonval:
> @fcollonval I have pushed the needed change to toogle for the current notebook with `Shift R`.

Thanks

> ## 1. Notebook metadata
> 
> The notebook reminds the side-by-side state via metadata. For now, after save, a new metadata is persisted. It is not clear to me from https://nbformat.readthedocs.io/en/latest/format_description.html if this is allowed/accepted. I don't know either if we want that. I can see advantages to it, as the next notebook author designs its content to fit a defined presentation, so the next reader get benefit from that persisted stat

I don't know. But it used to be done like this by lots of extensions in the classical notebook.

I realize also that the convention goes more towards using `_` rather than `-`.


> ## 2. RTC
> 
> I have not tested with RTC, but I can imagine your notebook rendering suddenly changing if a remote collaborator toggles its side-by-side state. Do we want to prevent that or it is a welcome feature?

As far as I understand RTC, only a subset of explicitly given metadata are synchronized between clients. cc @hbcarlos 

> ## 3. Menu label
> 
> For now, the label in the menu is a bit long, though descriptive `Toggle Side-by-side rendering for current Notebook`. Any comment on that?

What about dropping the `for current Notebook` and use `Toggle Side-by-Side Rendering` (using capitalized word for homogeneity)? The command will not be enable on other document, isn't?

hbcarlos:
@fcollonval thanks for pinging me.

Since #11064 all the metadata is shared except for `kernelspec`, `language_info`.

fcollonval:
Thanks Carlos

Then indeed synchronizing those view related settings may be pretty annoying as they are more a question of taste.

echarles:
@fcollonval @hbcarlos thx for the reviews and comments. I have pushed those changes

1. Notebook metadata: Nows uses `render_side_by_side` (with underscores)
2. RTC: Added `render_side_by_side` to the `UNSHARED_KEYS` to prevent unexpected UI changes
3. Label: Uses `Toggle Side-by-side Rendering` (short and correctly defines the action)

fcollonval:
Thanks @echarles - for me you only need to update the UI test snapshot and we are good to go.

krassowski:
> 2. RTC: Added `render_side_by_side` to the `UNSHARED_KEYS` to prevent unexpected UI changes

- Which one gets ultimately saved? The one of the host? If collaborator reopens they will see different layout, right?
- What if I want to have two views of the same notebook, do they have to be both in the same mode (side-by-side or normal)?

I think we should discuss whether this should be per-view or per-notebook setting; for majority of use cases they will behave the same in a single session, except for the RTC corner cases and multiple views of the same notebook). The drawback of having it per view/widget is that we cannot save the state (well we can save it in layout restorer but it won't be persistent beyond current workspace).

I can see an argument against storing it along: we do not store scroll position in notebook itself (although it could be useful) because it is the responsibility of the viewer to remember the state (in our case of the layout restorer).

There is also an analogy with other documents (such as Word): do they store the view state in the documents themselves (e.g. web view, print view, normal view; zoom level)? I think that some of these things are stored, but I am not sure if all.

krassowski:
Instead of `render_side_by_side` we should **consider** using something more generic like `rendering_layout: "side_by_side"` or `arrange: "side_by_side"` (is there a preference for underscore/hyphen in values too?), so that we can add more mutually exclusive modes in the future without having to worry about what to do when both `render_side_by_side` and say `render_detached` (or whichever we will explore next) mean when both are set.

Some users like to run notebooks one cell at a time; I could imagine a view in which the active cell is always centered and output is displayed in dedicated output widget/panel as in RStudio. Or just a view identical as in RStudio (so input cells are mostly as they are but output is always shown in a separate tab which can be moved around). But then those views are more like user-preference layouts (as discussed in #8960) and we probably would not want to save them in the notebook anyways (?).

Another question is whether we should distinguish left-to-right and right-to-left (i.e. whether outputs are on the left or on the right) positioning; this could be an extra metadata field to be added in the future, or just different values like `layout: "side-by-side-outputs-on-left"`. This is important from authoring perspective (if users write "see the result on the right but someone will be viewing this in a localized LTR view in the future they might be confused).

echarles:
@krassowski 

1. If we don't use the notebook metadata to store the `render_side_by_side` rendering, I guess we will need to store it at the Panel/Widget level. Do you see other options.
2. We can move to a more `rendering_layout` naming.

echarles:
@krassowski @fcollonval @hbcarlos I have pushed more to generalize the rendering layout notion and not depend on the notebook metadata anymore. This means you can toggle now the notebook views independently.

Untitled

echarles:
@krassowski I hope I have addressed your points with my last commit. The menu is not changed (still using the `Toogle side-by-side rendering`) as this is the only rendering mode for now, but the implementation supports adding more rendering modes.

fcollonval:
I played with this PR on binder and found three issues (one not directly related to this PR but to side-by-side feature):

bug_side_by_side

1. The screencast displays a notebook with a second view on the same notebook. If I hit `Shift R` on the left notebook, then focus on the right notebook; the right notebook switch to side-by-side rendering without pressing the shortcut... This is unexpected. It should either switch directly or not at all.
2. Long text in code-mirror do not reduce their width as other cells (look at the second code cell).
3. If the screen width is small (like in the screencast) the sides are overlapping. I would forbid using side by side mode if the width is too small (although _too small_ is a complex notion to translate in numbers...).

echarles:
Thx @fcollonval I see indeed non-coheren behavior when hitting `Shift-R`. I don't see a strict reproducible pattern, and have not hit that during the dev of that feature. I will look at that.

Your points 2. and 3. are IMHO not related to this feature.

echarles:
> Thx @fcollonval I see indeed non-coherent behavior when hitting Shift-R. I don't see a strict reproducible pattern, and have not hit that during the dev of that feature. I will look at that.

@fcollonval Latest commit fixes point 1. We can open separated issues for points 2. and 3. (we need to discuss how the global settings act in relationship with the toggling for such small screen cases)

fcollonval:
> We can open separated issues for points 2. and 3. (we need to discuss how the global settings act in relationship with the toggling for such small screen cases)

I'm fine with that being tackle in another PR (just wanted to point it out).

fcollonval:
The CI error seems relevant

```
Error: test/widget.spec.ts(54,3): error TS2741: Property 'renderingLayout' is missing in type '{ showHiddenCellsButton: boolean; scrollPastEnd: boolean; defaultCell: nbformat.CellType; recordTiming: boolean; numberCellsToRenderDirectly: number; renderCellOnIdle: boolean; observedTopMargin: string; observedBottomMargin: string; maxNumberOutputs: number; disableDocumentWideUndoRedo: boolean; }' but required in type 'INotebookConfig'.
```

echarles:
> I'm fine with that being tackle in another PR (just wanted to point it out).

Just opened 2 follow-up issues for that:

- https://github.com/jupyterlab/jupyterlab/issues/11652
- https://github.com/jupyterlab/jupyterlab/issues/11653

echarles:
Test is fixed. Last CI run shows 3 failing which I think are not introduced by this PR.

fcollonval (approved):
LGTM

> UI test failure is not related.

echarles:
Thx @fcollonval for the reviews and approval.

@krassowski anything missing to get this merged?

jtpio:
@echarles I restarted the failing checks just to be sure

echarles:
@jtpio Result is `4 cancelled and 62 successful checks`

packages/notebook-extension/src/index.ts
```diff
@@ -2347,29 +2353,36 @@ function addCommands(
   });
 
   commands.addCommand(CommandIDs.toggleRenderSideBySide, {
-    label: trans.__('Render Side-by-side'),
+    label: trans.__('Toggle Side-by-side rendering setting'),
```
krassowski, line 2356:
```suggestion
    label: trans.__('Render Side-by-side by Default'),
```

Or some variation? Like I don't feel like it's much better, but the current version is not distinct enough from the per-notebook command IMO.

echarles: The `toggleRenderSideBySide` is responsible to toggle the setting. I am not sure that `Render Side-by-side by Default` expresses that action, especially that the sentence remains the same whatever the setting state (true/false) is.

krassowski: Yes, my proposal is not great. But I do feel like "toggle setting" will not be obvious to the user. Maybe something like "Toggle default notebook layout: side-by-side"?

echarles: I have no strong opinion on that. Maybe input from @fcollonval or @jtpio would help.

packages/notebook/test/widget.spec.ts
```diff
@@ -43,7 +43,8 @@ const notebookConfig = {
   observedTopMargin: '1000px',
   observedBottomMargin: '1000px',
   maxNumberOutputs: 50,
-  disableDocumentWideUndoRedo: true
+  disableDocumentWideUndoRedo: true,
+  renderingLayout: 'default' as 'default' | 'side-by-side'
```
jtpio, line 47: Maybe `'default' | 'side-by-side'` should be made a `type` for improved readability and maintainability?

packages/notebook-extension/src/index.ts
jtpio: Another option could be `Enable Side-by-side Rendering by Default`?

packages/notebook/test/widget.spec.ts
echarles: Done. Thx @jtpio

packages/notebook-extension/src/index.ts
echarles: I have removed that command which had been there in the previous iteration of the side-by-side, but looking closer, was not changing for real the user setting, so was not doing the job it was supposed to do.

Also we have generalized the the notion of rendering layout, so it become difficult anyway to find a good fit for that commande.

fcollonval: Earlier I proposed to use `Render Side-by-Side` (using the :white_check_mark: mark to distinguish the status) to be aligned with for example `Show line numbers` (that is also per document) or `Wrap words`.

echarles: Hi, thx for the help here. To wrap up, we have a setting `renderingLayout` which can be `default` or `side-by-side` and we are talking about 2 commands

- The one that changes toggle the layout of the current notebook and is available via the menu. The name is now `Toggle Side-by-side Rendering`. I can change it to `Render Side-by-Side` with the ✅.
- The one that was there before and I have removed which was supposed to toggle the user setting. The former implementation had just a in-memory flag that was used to apply the toggle the rendering on all the notebooks, but was note changing the user setting. I don't think a command can, or should change a user setting. This is why I have dropped it

echarles:
My last commit implements what we have been discussed on https://github.com/jupyterlab/jupyterlab/pull/11608#discussion_r770284417

The menu shows a mark to indicate the side-by-side toggling state 

Untitled

I think we are good to go now.

fcollonval:
I'm :+1: with the latest version. The reference snapshots for the UI tests need to be updated. Then this is good to me.

echarles:
I have updated the UI snapshot. UI Tests are failing with the following (before it was explicitly failing with the wrong snapshots. I will relaunch to see what it gives.

```
    test-results/test-jupyterlab-debugger-Debugger-Tests-Start-debug-session-jupyterlab/cc0cedab382e97654b8e1332220a4967.webm
    ------------------------------------------------------------------------------------------------

    Retry #1 ---------------------------------------------------------------------------------------

    page.evaluate: TypeError: Cannot read properties of null (reading 'type')

        at GalataInpage.waitForCellRun (<anonymous>:3790:37)
```

echarles:
RTD is still building with Error Concurrency limit reached (2), retrying in 5 minutes.

Otherwise all green.

echarles:
@fcollonval CI is now fully green. Can we merge this?

echarles:
Thx @fcollonval @krassowski @jtpio for the help.

echarles:

Issue #6013, opened 2019-02-21
Schema for Workspaces and Commands?


bollwyvl:
First off, thanks @afshin et al for making the workspace stuff what it is today! 

Over on the venerable https://github.com/jupyterlab/jupyterlab/issues/1640#issuecomment-465839781, I've broached using workspace JSON as a format for encoding things like dashboards and presentations. UI aside, for this and other reasons, I'd like to see the workspace format become more intentionally-managed, with the whole shooting match of:
- versioned JSON Schema
- reference validators/migrators in python and typescript

I think we're sitting at a `v0` right now...
<details>
<summary>...with something like...</summary>
<pre>
{
  "type": "object",
  "required": ["data", "metadata"],
  "properties": {
    "data": {
      "type": "object",
      "patternProperties": {
        ".*": {
          "type": "object"
        }
      }
    },
    "metadata": {
      "type": "object",
      "properties": {
        "id": {
          "type": "string"
        }
      }
    }
  }
}
</pre>
</details>

... which doesn't really say much. All the "good stuff" is in the definitions of each of the keys under `data`, and _those_, based on my reading of the [docs](http://jupyterlab.github.io/jupyterlab/application/classes/layoutrestorer.html), are basically the args of each of the plugin's commands.

So to round out `v0`, we'd need to be able to build up a synthetic schema for _your_ lab+dog. Now, of course, some commands won't be able to _be_ typed, but I think for a lot of them we could. So, overloading `schemaDir` to include `commands.json`, we could specify partial schema there...
<details>
<summary>... like for restoring notebooks, <code>notebook.v0.schema.json</code></summary>
<pre>
{
  "version": "0",
  "type": "object",
  "patternProperties": {
    "^notebook:.*$": {
      "type": "object",
      "required": ["data"],
      "properties": {
        "data": {
          "type": "object",
          "required": ["factory", "path"],
          "properties": {
            "factory": {
              "enum": ["Notebook"]
            },
            "path": {
              "type": "string"
            }
          }
        }  
      }
    }
  }
}
</pre>
</details>

...though the real action really fires up on the dockpanel, etc. 

Anyhow, then we can construct an in-memory schema that looks at all of the installed extensions, make a huge `allOf`, and Bob's-your-uncle. 

[...]

PR #11614, opened 2021-12-03
Customize layout from settings

### LINKED ISSUE #8960: Jlab could provide a way to select predefined layouts where widgets are always docked together in a certain way

fcollonval:
## References

Fix #8960

demo_customize_layout


fcollonval:
cc @jtpio @ellisonbg @bollwyvl here is a proposition to get widget layout customization. I think we could use such approach in RetroLab too to overcome supporting extension panel positionning in it. For example, in retrolab we could allow three state for sidebar: `left`, `right`, `treeView` (depending if the panel needs to be positioned on a notebook sidebar or in the initial tree view).

bollwyvl:
fascinating, will take a look!

bollwyvl:
One thought on this: at the end of it, I would hope there wouldn't be much "special" stuff that was not introspectable from outside that couldn't be resolved until there was a JS interpreter and hot DOM.

By that I mean: instead of the magic string `terminals` it might be something namespaced by `@jupyterlab/terminals`, as one of the more (accidentally) powerful patterns we have is that `@{org}/{pkg}:{plugin}` is configurable if and only if there exists a file called `$PREFIX/labextensions/${org}/${pkg}/{schemaDir}/{plugin}.json` (ignoring for the moment that you have to look up `schemaDir`). There, schema has _also_ given us predictable human-readable things to know, as well as lots of configuration. 

A less-discoverable language we already have is [lumino commands](https://github.com/jupyterlab/lumino/issues/58#issuecomment-963103017). This is also, semi-accidentally, surfaced by the [workspace](https://github.com/jupyterlab/jupyterlab/issues/6013). It suffers, however, from the commands _not_ being registered by a controlled namespace.

I imagine the final form of this at some point in the future being an abstract data shell that is a mapping of (the possibility of a) components to a set of regions, with a descriptive-enough language that it would be straightforward to build a visual tool that could discover, configure, and re-combine all available screen elements into:

- a dashboard
- a [single-cell](https://github.com/jupyterlite/jupyterlite/issues/148)
- a presentation
 
...where any of them could include self-modifying tools, such as a command palette (with the now-hidden group metadata restored, and expanded).

Practically, I imagine this would change `shell.add` to _not_ specify (or at least require/always respect) `area`, as the user-configured shell would decide, based on the widget type, where that-which-is-being-added should go, especially if the chosen `area` didn't exist. 

also cc @ajbozarth... this isn't exactly the launcher issue (never made the issue, sorry), but it's a related concern, as launchers would probably have to be aware of this kind of system.

fcollonval:

fcollonval:

fcollonval:
CI failure tracked in https://github.com/jupyterlab/jupyterlab/pull/12552

jtpio:
Nice, thanks @fcollonval for iterating on this!

Looks like there are a few conflicts to fix.

docs/source/user/interface_customization.rst
```diff
@@ -0,0 +1,298 @@
...
+- TSV Viewer: *TSV Viewer* -> *toolbar*
+
+Those settings are accessible through the interactive *Settings Editor*. But you
+may find more comfortable to use the *Advanced Settings Editor* (accessible by clicking
+on *JSON Settings Editor* button in the top right corner of the *Settings Editor*).
+
+Your ``toolbar`` customization will be merged with the default menu bar definition (this
+is a different behavior as the other settings that are overriding the default settings).
+So to remove a default item, you will need to disable it. To do that, you will need
+to set the toolbar item ``disabled`` attribute to ``true``; a item is uniquely identified
```
trungleduc, line 109:
```suggestion
to set the toolbar item ``disabled`` attribute to ``true``; an item is uniquely identified
```

docs/source/user/interface_customization.rst
```diff
@@ -0,0 +1,298 @@
...
+.. _menu-bar-customization:
+
+Menu Bar
+--------
+
+The top menu bar definition can be customized through the *Advanced Settings Editor*. You can
+access it by opening the *Settings Editor* from the *Settings* menu, then click on the
+*JSON Settings Editor* in the top right corner.
+
+In the *Advanced Settings Editor*, you will need to the select the *Main Menu* section in
```
trungleduc, line 158:
```suggestion
In the *Advanced Settings Editor*, you will need to select the *Main Menu* section in
```

docs/source/user/interface_customization.rst
```diff
@@ -0,0 +1,298 @@
...
+        // Add new terminals in the down area in simple mode
+        "Terminal": { "area": "down" }
+      },
+      "multiple": {
+        // Add new terminals in the right sidebar in default mode
+        "Terminal": { "area": "right" }
+      }
+    }
+
+The above example, will results in the following changes:
```
trungleduc, line 74:
```suggestion
The above example will result in the following changes:
```

packages/application-extension/src/index.tsx
```diff
@@ -1356,50 +1405,13 @@ namespace Private {
...
-        labShell.layoutModified.connect(update);
-        update(labShell, initial);
-      }
-    );
-
     // Add a command to switch a side panels's side
     app.commands.addCommand(CommandIDs.switchSidebar, {
       label: trans.__('Switch Sidebar Side'),
-      execute: () => {
+      execute: async () => {
```
trungleduc, line 1414: Do we not need `async` here?

trungleduc (commented):
Thanks @fcollonval, minor suggestions on the doc then it looks great to me!

packages/application-extension/src/index.tsx
fcollonval: Good catch

fcollonval:
Thanks @trungleduc for the review

jtpio:
Nice this is looking good :+1: 

Just tested on Binder and it looks like the terminal is still opened in main area when created from the launcher:

https://user-images.githubusercontent.com/591645/169058846-bbd9e4cb-158c-4cfb-b40a-e474babd0d29.mp4

jtpio:
Nice thanks!

Apart from that it's looking good. I think we can get this in so it can be released in the next 4.0 pre-release and tested in Notebook v7.

jtpio:
Check again on Binder and looking good:

https://user-images.githubusercontent.com/591645/169076381-eec06930-2e26-408a-b0b9-e3fcda03042e.mp4

jtpio (approved):
Thanks!

fcollonval:
Thanks for the reviews

Issue #10785, opened 2021-08-07
Feature: Collapse/Uncollapse cells with LEFT and RIGHT keyboard's arrows keys.


willemavjc:
### Search the opened issues for the same idea

Done: None found.

### Problem

I'm faster with keyboard's shortcuts than the mouse so that I use `UP` and `DOWN` keyboard's arrows keys to move up or down in my notebooks. I cannot do *collapsing* or *uncollapsing* of cells though either than clicking with the mouse on this small UX arrow icon.

### Proposed Solution

It would be great - if possible - to add 2 more shortcuts that would allow  *collapsing* or *uncollapsing* of the cells using the keyboard and not only the mouse.

`UP` and `DOWN` keyboard's arrows keys would be the most convenient as they are natural and this behavior tends to be found on other apps in the wide I.T. world.

agoose77:
You can already implement this with a custom keyboard shortcut!

```json
{
    "shortcuts": [
        {
            "command": "apputils:run-all-enabled",
            "keys": [
                "ArrowRight"
            ],
            "args": {
                "commands": [
                    "notebook:hide-cell-code",
                    "notebook:hide-cell-outputs"
                ],
                
                "args": [
                    {},
                    {}
                ]
            },
            "selector": "body"
        },
        
        {
            "command": "apputils:run-all-enabled",
            "keys": [
                "ArrowLeft"
            ],
            "args": {
                "commands": [
                    "notebook:show-cell-code",
                    "notebook:show-cell-outputs"
                ],
                
                "args": [
                    {},
                    {}
                ]
            },
            "selector": "body"
        }
    ]
}
```

willemavjc:
Oh, that's interesting! Shouldn't that be bundled in by default? Wouldn't that make sense?

agoose77:
> Oh, that's interesting! Shouldn't that be bundled in by default? Wouldn't that make sense?

Perhaps! This might be where other developers with more of an eye for feature creep might weigh in :)

willemavjc:
@agoose77 Oh, I took the time to try out your tip! How could I make it work for every cell? (including markdown etc.)

schmidi314:
[...]

PR #11615, opened 2021-12-03
Keyboard shortcuts related to collapsible headings

### LINKED ISSUE #11401: Keyboard shortcut parity (with notebook) for collapsible headings
The classic collapsible headings have some keyboard shortcuts that are a key part of the user experience and would be a positive addition to JupyterLab:

> ... lab doesn't have the keyboard shortcuts, which is the key thing - since with the keyboard shortcuts one can jump to start and end of sections so easily, collapsing them actually becomes much less important. Note also that collapsible headings integrates with the TOC2 extension to allow them to optionally sync collapsed sections.

(@jph00 -- https://github.com/jupyter/notebook/issues/6210#issuecomment-960256975)

### Problem

@jph00 continues, describing the behavior in the classic extension that does not exist in JupyterLab:

> Here's what the docs say about the keyboard shortcuts:
> * Command-mode keyboard shortcuts, (enabled by default, and set to left and right arrow keys to collapse/expand sections, or go to the previous/next heading, plus shift-right to select a heading cell's section, shift-a/b to insert a heading above/below the current cell, ctrl-shift-left and ctrl-shift-right to collapse/uncollapse all headings). Bindings are also configurable from the config page
> 
> Details: https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/collapsible_headings/readme.html

### Proposed Solution
Let's add these shortcuts by default.

### Additional context
* Original discussion thread: https://github.com/jupyter/notebook/issues/6210#issuecomment-960256975
* Notebook v7 issue: https://github.com/jupyter/notebook/issues/6220

cc: @fcollonval @ajbozarth @marthacryan -- If no one has claimed this issue by next week, could you please raise it at the weekly JupyterLab call?

schmidi314:
## References
#10785 
Fixes #11401 

Adding commands and callbacks for features similar to Collapsible Headings in classic notebook:

* notebook:move-cursor-heading-above-or-collapse
* notebook:move-cursor-heading-below-or-expand
* notebook:insert-heading-above
* notebook:insert-heading-below

## Code changes

IMHO [packages/notebook/src/actions.tsx ](https://github.com/jupyterlab/jupyterlab/blob/568ccde8e985a5cadfdad8c31b29c66a1b7d31c9/packages/notebook/src/actions.tsx) could use some more structure, at least if it comes to the private functions. For the moment, I have created a new namespace `Private.Headings` that contains all new functions related to headings. I think this one:
 https://github.com/jupyterlab/jupyterlab/blob/a6c19ed240c230379e89703d2f9c54d7c01164ba/packages/notebook/src/actions.tsx#L1622
should be moved into this namespace as well... or, since `findNearestParentHeader` seems to be buggy (fails if cell is a markdown cell), it should be replaced by the new
https://github.com/jupyterlab/jupyterlab/blob/a6c19ed240c230379e89703d2f9c54d7c01164ba/packages/notebook/src/actions.tsx#L2425
...still investigating

## User-facing changes

* Default shortcuts for ArrowLeft (jump to heading above or collapse) and ArrowRight (jump to heading below or expand)

left_right_collapse_expand

* Default shortcuts for Shift+A (new heading of same level above) and Shift+B (new heading of same level below)

new_headings

* Better cursor placing when starting to edit a heading cell.

cursor_placement


schmidi314:
@meeseeksdev tag enhancement

fcollonval:
Thanks for starting this @schmidi314 

A comment about GitHub, when referring to an issue, could you set the special word _fix_ or _fixes_ or _closes_ in front of the `#<ref>`? This will automatically link the issue with the PR (and close it when the PR is merged).

This PR addresses also partly https://github.com/jupyterlab/jupyterlab/issues/11401 (actually to fix that issue the only thing missing are shortcuts Ctrl Shift Left/Right to (un)collapse all headings).

packages/notebook/src/actions.tsx
```diff
@@ -995,9 +1113,9 @@ export namespace NotebookActions {
             notebook.lastClipboardInteraction === 'cut' &&
             typeof cell.id === 'string'
           ) {
-            let cell_id = cell.id as string;
+            let cellId = cell.id as string;
```
fcollonval, line 1116: :rocket: thanks for making it compliant with the naming convention

packages/notebook/src/actions.tsx
```diff
@@ -2292,4 +2410,155 @@ namespace Private {
...
+    /** Find the heading that is parent to cell.
+     *
+     * @param childCell - The cell that is child to the sought heading
+     * @param notebook - The target notebook widget
+     * @param includeChildCell [default=false] - if set to true and childCell is a heading itself, the childCell will be returned
+     * @param returnIdx [default=false] - if set to true, the cell idx is returned rather than the cell object.
+     *
+     * @returns the (index | Cell object) of the parent heading or (-1 | null) if there is no parent heading.
+     */
+    export function findParentHeading(
```
fcollonval, line 2425: @marthacryan is there already some helper functions to get the parent or the sibling header?

> It does not look like it. But I'm not familiar with the heading collapser feature.

fcollonval (commented):
This looks awesome @schmidi314; I tested it on binder and this works as expected.

I pinged Martha that introduces the heading collapser feat. about avoiding code duplication and having common helper functions (I'm not familiar with that part of the code).

Would you mind adding some test for it?

> about the keyboard test with Playwright, you can emulate key event with [`page.keyboard`](https://playwright.dev/docs/api/class-keyboard/)

packages/notebook/src/actions.tsx
schmidi314: Actually I think there is. It is the `findNearestParentHeader` (see PR description). But my first impression is that it is buggy so I didn't use it. But I'm not sure if the reaction on markdown cells is intended.

schmidi314:
> Would you mind adding some test for it?

Sure. I was hesitant to do that because the UI tests already take ages. But I'll add some soon.

> This PR addresses also partly #11401 (actually to fix that issue the only thing missing are shortcuts Ctrl Shift Left/Right to (un)collapse all headings).

Then I probably should add them as well.

fcollonval:
> > Would you mind adding some test for it?
> 
> Sure. I was hesitant to do that because the UI tests already take ages. But I'll add some soon.

The value of those tests is huge. But I know that designing them takes time. One tip on that front is to run only one test file with:

```sh
jlpm run test tests/jupyterlab/workspace.test.ts
```

Or even a single test with the `-g` option:

```sh
jlpm run test -g "should reset the workspace"
```

> > This PR addresses also partly #11401 (actually to fix that issue the only thing missing are shortcuts Ctrl Shift Left/Right to (un)collapse all headings).
> 
> Then I probably should add them as well.

If you have the time, it would be awesome.

schmidi314:
> The value of those tests is huge. But I know that designing them takes time.

Yes, sure. :) But what I meant is that the UI test on github runs for about an hour already. At some point, I could imagine, automated tests become very annoying if you have to wait for half a day until completion. Maybe one should think about parallelization...?

And for some reason the snapshots on my computer don't agree with the tests on github. Maybe this is related to a non-100% scaling of my monitor. Any experience in this regard?

schmidi314:
The Ctrl+Shift+Left/Right handlers and commands were already there, so that's an easy one. 

I noticed that the notebook action `collapseAll` and the commands `Collapsible_Headings:XXX` don't quite match the scheme. I have marked the commits as optional to show my proposal to change them. Not sure if this actually improves things...

fcollonval:
> Yes, sure. :) But what I meant is that the UI test on github runs for about an hour already. At some point, I could imagine, automated tests become very annoying if you have to wait for half a day until completion. 

GitHub won't let us reach that state as a CI job must complete in less than 6 hours :wink: (what is ok in comparison with the usual review time).

> Maybe one should think about parallelization...?

When introducing the new galata based on Playwright fixture, I had more flaky test with multiple runner (and even sometimes it took longer...). But sure we should definitely try again especially as the number of tests grows.
An alternative can be to group tests and run some only when a PR is ready for review or something like that.
 
> And for some reason the snapshots on my computer don't agree with the tests on github. Maybe this is related to a non-100% scaling of my monitor. Any experience in this regard?

Which OS are you using? Could you please provide a list of the failing tests?  
Some tests are expected to fail (like the one related to the terminals) as they are influenced by the OS configuration. But we should reduce that list if possible.

schmidi314:
> Which OS are you using? Could you please provide a list of the failing tests?

It would probably be easier to provide a list of those that don't fail :)

But maybe this helps: I think it is related to fonts (forget my last comment about resolution scaling). Typical cases:

fail1
fail2

In principle I'm on Ubuntu 20.04. But it is a customized OS shipped with a tuxedo laptop.

image

My impression is that raw and code cells are OK. For instance, the notebook-toolbar/cut-cell is fine (one code and one raw cell in the snapshot).

krassowski:
I guess that it could be related to https://github.com/jupyterlab/jupyterlab/pull/11388 - I think that since then the snapshots depend on system font settings (including font family, size but possibly even subpixel rendering settings). Ideally we would just override the fonts for tests (without requiring to run a docker which was a tad high barier of entry).

schmidi314:
@krassowski thanks for the hint. Setting the fonts made the github results consistent with my local results. For the moment I have merely done a test-specific `mockSettings` for `content-font-family` and `content-font-size`. I understood you wanted to set this globally, right? I would very much appreciate this because my present workflow (upload without snapshots, let github produce the snapshots, download, commit snapshots, push, wait for test ...) is a real pain.

Anyway, for this PR I wouldn't mix this in but rather start a new PR with globally setting fonts all test snapshots updated. What do you think?

krassowski:
> Anyway, for this PR I wouldn't mix this in but rather start a new PR with globally setting fonts all test snapshots updated. What do you think?

Will we need to change any snapshots? I think that current CI screenshots are using some variant of Ubuntu font (I might be wrong here), which has the benefit of being free (/though there are fonts with more free licences) and available on the CI; the downside is that developers on Mac and Windows will need to install it. We could ship _a_ font file in `galata/` to streamline the developer experience. We should ask on the dev meeting if there are other preferences, but I think this is a reasonable solution.

fcollonval:
Thanks a lot for figuring out how to remove that font system impact. As you figure out what to change in the default settings, would you mind opening a PR?

The default mocked settings are (for now they avoid cursor blinking) at https://github.com/jupyterlab/jupyterlab/blob/b80475fbd409385eae2017fbb82c72c638fd6959/galata/src/galata.ts#L27

Regarding shipping a font, I don't have strong opinion. The only sure point is that we already _don't care_ about running tests on Mac and Windows because reference snapshots are tagged by OS (playwright is automatically appending the `_linux` suffix to them) and we do not provide them.  
Could it enough to use _Arial (sans-serif)_ (https://www.w3schools.com/cssref/css_websafe_fonts.asp)?

docs/source/extension/extension_migration.rst
```diff
@@ -36,7 +36,11 @@ bumped their major version (following semver convention):
    The ``anchor`` parameter of ``HoverBox.IOptions` is now a ``DOMRect` instead of ``ClientRect``.
    The ``CodeEditor.ICoordinate`` interface now extends ``DOMRectReadOnly`` instead of ``JSONObject, ClientRect``.
 - ``@jupyterlab/notebook`` from 3.x to 4.x
```
fcollonval, line 38: Thanks for adding that doc

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    mockSettings: {
+      '@jupyterlab/apputils-extension:themes': {
+        overrides: {
+          'content-font-family': 'Arial',
+          'content-font-size1': '12pt'
+        }
+      }
+    }
+  });
+
```
fcollonval, line 136: Let's remove this in the final version and use the snapshot produced by the CI. The snapshots could then be all updated at once when changing the default font family.

```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+  // create an empty notebook for each test
+  test.beforeEach(async ({ page }) => {
+    await page.notebook.createNew(fileName);
+  });
```
fcollonval, line 140: It seems you run the those two actions for each test. So it could be moved here:

```suggestion
    await populateNotebook2(page);
    await page.notebook.run();
  });
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+
+  test('Jump to Previous Header', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 144:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.notebook.selectCells(6);
+    await page.keyboard.press('ArrowLeft');
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('jump_previous_header.png');
+  });
+
+  test('Collapse Previous Header', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 154:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('ArrowLeft');
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('collapse_previous_header.png');
+  });
+
+  test('Collapse Previous Headers', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 165:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('ArrowLeft');
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('collapse_previous_headers.png');
+  });
+
+  test('ReExpand Headers 01', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 178:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.keyboard.press('ArrowRight');
+    await page.keyboard.press('ArrowLeft');
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('reexpand_headers_01.png');
+  });
+
+  test('ReExpand Headers 02', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 196:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('ArrowRight');
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('reexpand_headers_02.png');
+  });
+
+  test('Add Header Below 01', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 211:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.notebook.selectCells(2);
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('add_header_below_01.png');
+  });
+
+  /** Check that header below adds header at end of present-level section. */
+  test('Add Header Below 02', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 228:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.keyboard.press('Shift+Enter');
+    await page.notebook.selectCells(0);
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('add_header_below_02.png');
+  });
+
+  test('Add Header Below 03', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 246:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.notebook.selectCells(2);
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('add_header_below_03.png');
+  });
+
+  /** Checks also if the cursor is at the right position when adding heading */
+  test('Add Header Above 01', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 266:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.waitForTimeout(200);
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('add_header_above_01.png');
+  });
+
+  /** Checks also if the cursor is at the right position when adding heading */
+  test('Add Header Above 02', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 278:
```suggestion
```

galata/test/jupyterlab/collapsible-headings.test.ts
```diff
@@ -111,3 +111,188 @@ test.describe('Collapsible Headings; no_showHCB', () => {
...
+    await page.waitForTimeout(200);
+    expect(
+      await (await page.notebook.getNotebookInPanel()).screenshot()
+    ).toMatchSnapshot('add_header_above_02.png');
+  });
+
+  /** Checks also if the cursor is at the right position when adding heading */
+  test('Add Header Above 03', async ({ page }) => {
+    await populateNotebook2(page);
+    await page.notebook.run();
```
fcollonval, line 290:
```suggestion
```

fcollonval (changes requested):
Thanks a lot for adding all the tests @schmidi314 

I have a small suggestion to refactor a bit the test code. And I would only like to see the font fix dropped for now from this PR.

schmidi314:
Could you do me a favor and check which fonts are used in your environment? For me it is Noto Sans. But I don't know how to figure out which font is used on github.

fcollonval:
It is _DejaVu Sans_

galata/test/jupyterlab/collapsible-headings.test.ts
schmidi314: Thanks for the hint. Did all of that and similar streamlining in the first two tests I did previously.

schmidi314:
Thanks @fcollonval , setting to this font at least partially fixes my problem: if I mock the per-test settings to those fonts, my local snapshots are consistent with github's. But globally setting them (in `galata.DEFAULT_SETTINGS` didn't work that well), as doing a `test.use(mockSettings:...)` overwrites all the `DEFAULT_SETTINGS`, it seems.

fcollonval:
> doing a test.use(mockSettings:...) overwrites all the DEFAULT_SETTINGS, it seems.

Yes [they do](https://github.com/jupyterlab/jupyterlab/blob/2e0a48382b16af4145ce8f1d39220b63ec80f417/galata/src/fixtures.ts#L85), because we should not block the ability to test the settings defined in `DEFAULT_SETTINGS`. But you can always merge the default with new ones.

fcollonval (approved):
Thanks a lot for pushing this through @schmidi314

krassowski (commented):
This is great, thank you. One question: how should insert new heading below behave when executed above the first heading/in empty notebook? The current behaviour seems a bit strange to me:

inser-heading

schmidi314:
Ah, that was a refactoring bug. Thanks for noticing! ...so much for the test suite.

Nevertheless, since this is a corner case in my view, I wouldn't add tests for that.

fcollonval:
Thanks @schmidi314 
> CI failure is not related.

packages/notebook/src/actions.tsx
```diff
@@ -2292,4 +2410,193 @@ namespace Private {
...
+      let baseHeadingLevel = Private.Headings.determineHeadingLevel(
+        baseCell,
+        notebook
+      );
+      if (baseHeadingLevel == -1) {
+        baseHeadingLevel = 1; // if no heading level can be determined, assume we're on level 1
+      }
+
+      console.log('A', baseHeadingLevel);
+
```
krassowski, line 2467:
```suggestion
```

packages/notebook/src/actions.tsx
```diff
@@ -2292,4 +2410,193 @@ namespace Private {
...
   }
+
+  /** Functionality related to collapsible headings */
+  export namespace Headings {
+    /** Find the heading that is parent to cell.
+     *
+     * @param childCell - The cell that is child to the sought heading
+     * @param notebook - The target notebook widget
+     * @param includeChildCell [default=false] - if set to true and childCell is a heading itself, the childCell will be returned
+     * @param returnIdx [default=false] - if set to true, the cell idx is returned rather than the cell object.
```
krassowski, line 2421:
```suggestion
     * @param returnIdx [default=false] - if set to true, the cell index is returned rather than the cell object.
```

packages/notebook/src/actions.tsx
```diff
@@ -2292,4 +2410,193 @@ namespace Private {
...
+        cellIdx--;
+      }
+      return returnIdx ? -1 : null;
+    }
+
+    /** Find heading above with leq level than baseCell heading level.
+     *
+     * @param baseCell - cell relative to which so search
+     * @param notebook - target notebook widget
+     * @param returnIdx [default=false] - if set to true, the cell idx is returned rather than the cell object.
```
krassowski, line 2449:
```suggestion
     * @param returnIdx [default=false] - if set to true, the cell index is returned rather than the cell object.
```

packages/notebook/src/actions.tsx
```diff
@@ -2292,4 +2410,193 @@ namespace Private {
...
+        cellIdx++;
+      }
+      return returnIdx ? -1 : null;
+    }
+
+    /** Find next heading.
+     *
+     * @param baseCell - cell relative to which so search
+     * @param notebook - target notebook widget
+     * @param returnIdx [default=false] - if set to true, the cell idx is returned rather than the cell object.
```
krassowski, line 2523:
```suggestion
     * @param returnIdx [default=false] - if set to true, the cell index is returned rather than the cell object.
```

krassowski (approved):
This looks amazing, thank you. Two points on the code:
- [ ] I spotted a debugging `console.log` - let's remove it
- [ ] I would personally avoid using `idx` abbreviation in any public API (e.g. I would rename `returnIdx` → `returnIndex`), but this is not a blocker; however, at a minimum please expand `idx` to `index` in documentation strings

schmidi314:
Sorry, I've been away for a while. Anything left to do for me here?

fcollonval:
Thanks @schmidi314 (especially for the ping), merging it now.

PR #11421, opened 2021-11-09
Backport PR #11420 on branch 3.2.x (Makes ILabShell optional in toc extension)

jweill-aws:
## References

This backports the change from PR #11420 to address issue #11419 in the 3.2.x branch.

## Code changes

See prior PRs

## User-facing changes

See prior PRs

## Backwards-incompatible changes

See prior PRs

PR #11617, opened 2021-12-04
Restore accidentally removed ToC context menu

### LINKED ISSUE #11584: Run groups of cells under header
### Problem

Often I group cells under a header to divide my notebook into sections. These *sections* may contain numerous cells, but I would not like to re-run them all by hand when I need to execute all the code in a certain *section*. These *sections* may be cells I always need to run *before* I can run more important sections of my notebook and the task gets very tedious. Also, marking them as "Initialization Cells" is not probably a solution because the freedom to execute sections as and when I want is important. 

---

#### Example:

My notebook has sections **A**, **B** and **C**.
**A** and **B** contain certain sections I need to run. 
  - For scenario 1, I run only section **A**. 
  - For scenario 2, I run only section **B**.
  - For scenario 3, I run both.
Finally, once I've executed one of these 3 scenarios, I execute section **C**.

---

This seems like a pretty common way to use the notebooks in an organised manner.

### Proposed Solution

Add an option to run entire sections (under a heading) at once. 

(A similar feature exists in Google Colab, which is super handy and hence seems like it would be possible to implement.)

krassowski:
## References

Fixes #11584. This PR is targeting 3.2.x branch because the code was only removed in manual backport.

## Code changes

Restores code accidentally removed in #11421. 

## User-facing changes

Context menu opens again in table of contents

None

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -125,6 +125,11 @@ async function activateTOC(
     label: trans.__('Run Cell(s)')
   });
 
+  app.contextMenu.addItem({
+    selector: '.jp-tocItem',
+    command: CommandIDs.runCells
+  });
+
   if (restorer) {
     // Add the ToC widget to the application restorer:
     restorer.add(toc, '@jupyterlab/toc:plugin');


jtpio (approved):
Thanks @krassowski for catching this!

jweill-aws:
Thanks for fixing this!

Issue #11553, opened 2021-11-25
Rename the table of content plugin id


jtpio:
This is a small issue to track renaming:

https://github.com/jupyterlab/jupyterlab/blob/1bf5fdff171f8a9e91755821efd376d5742778ba/packages/toc-extension/src/index.ts#L235

To:

```ts
'@jupyterlab/toc-extension:registry'
```

So it's consistent with the rest of the code base.

### Additional context

This came as an observation when looking at https://github.com/jupyterlab/jupyterlab/pull/11530

Adding to 4.0 as a potential breaking change, since the rename can affect users who were using it for disabling the extension for instance.

PR #11621, opened 2021-12-06
Revert "Restore accidentally removed ToC context menu"

fcollonval:
Reverts jupyterlab/jupyterlab#11617

In fact the trouble is the plugin name of the toc that does not respect JupyterLab pattern => the settings containing the menus definition are not loaded.

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -125,11 +125,6 @@ async function activateTOC(
     label: trans.__('Run Cell(s)')
   });
 
-  app.contextMenu.addItem({
-    selector: '.jp-tocItem',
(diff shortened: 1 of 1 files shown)


krassowski:
So that should probably be addressed on master first and then back-ported?

fcollonval:
Actually it is probably a bad idea to backport it then to 3.2.x as it requires changing the plugin id.

jtpio:
The TOC extension needs a pass to use consistent naming for the plugin and settings: https://github.com/jupyterlab/jupyterlab/issues/11553

But that will indeed be a 4.0 change.

fcollonval:
I opened #11622 to push a fix on master.

PR #1927, opened 2017-03-15
Create a client session object

blink1073:
Fixes #1856.  Fixes #1835.

Formalizes the logic that the context was internally using for a session that may or may not be active on the server into a consistent interface that can be used across the different components, including the console and toolbar.  Also consolidates the kernel dialog helpers.

PR #11626, opened 2021-12-07
Remove leading slash from console path

davidbrochart:
## References

https://github.com/jupyterlab/retrolab/issues/300

## Code changes

This change ensures that a console path doesn't start with `/`, otherwise this leads to the console URL having a double slash e.g. in RetroLab:
```
http://localhost:8888/retro/consoles//console-1-33a9d4ee-3873-4611-9edd-c1ff7afee29e
```

## User-facing changes

None

None

### DIFF
--- a/packages/console/src/panel.ts
+++ b/packages/console/src/panel.ts
@@ -55,7 +55,9 @@ export class ConsolePanel extends MainAreaWidget<Panel> {
       options.contentFactory || ConsolePanel.defaultContentFactory);
     const count = Private.count++;
     if (!path) {
-      path = `${basePath || ''}/console-${count}-${UUID.uuid4()}`;
+      path = `${
+        basePath ? basePath + '/' : ''
+      }console-${count}-${UUID.uuid4()}`;
     }
 
     sessionContext = this._sessionContext =
--- a/packages/console/src/panel.ts
+++ b/packages/console/src/panel.ts
@@ -8,7 +8,7 @@ import {
   sessionContextDialogs
 } from '@jupyterlab/apputils';
 import { IEditorMimeTypeService } from '@jupyterlab/codeeditor';
-import { PathExt, Time } from '@jupyterlab/coreutils';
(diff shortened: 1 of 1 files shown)


packages/console/src/panel.ts
```diff
@@ -55,7 +55,9 @@ export class ConsolePanel extends MainAreaWidget<Panel> {
       options.contentFactory || ConsolePanel.defaultContentFactory);
     const count = Private.count++;
     if (!path) {
-      path = `${basePath || ''}/console-${count}-${UUID.uuid4()}`;
+      path = `${
+        basePath ? basePath + '/' : ''
```
davidbrochart, line 59: I'm not sure if `basePath` can have a trailing `/`, and if we should check for it.

jtpio: Does the git history say anything about why the `/` was introduced in the first place?

davidbrochart: Looks like it was introduced in https://github.com/jupyterlab/jupyterlab/commit/50a771bc9affed9e2dbaf5bbe459f4f7b52ddad6. [The PR](https://github.com/jupyterlab/jupyterlab/pull/1927) doesn't say much about it.

davidbrochart: Actually, `basePath` is the working directory where the console is launched, and I checked that it doesn't have a trailing slash, so this should be good.

krassowski: Maybe the join should be done with `URLExt.join` from `@jupyterlab/coreutils` instead? I think it can handle trailing slashes.

davidbrochart: Indeed, thanks!

jtpio:
Thanks @davidbrochart!

The change looks good, especially since there doesn't seem to be a reason for having the leading `/` in the first place.

jtpio (approved):

fcollonval:

PR #11635, opened 2021-12-08
Specifying print CSS rules for notebooks

SylvainCorlay:
## References

This CSS change is meant to prevent page breaks in the middle of outputs when printing notebooks (webpdf export in nbconvert).

## Code changes

Adding `@media print` CSS rules to:

 - hide collapsers
 - display the input wrappers and output wrapper with `display: block` (because flex prevents the use of `break-inside` CSS property in inner HTML).
 - set the `break-inside` property to `avoid-page` for `jp-OutputArea-child` elements.

## Screenshots (Before | After)

before after

### DIFF
--- a/packages/cells/style/collapser.css
+++ b/packages/cells/style/collapser.css
@@ -23,3 +23,19 @@
   top: 0px;
   bottom: 0px;
 }
+
+/*-----------------------------------------------------------------------------
+| Printing
+|----------------------------------------------------------------------------*/
+
+/*
+Hiding collapsers in print mode.
+
+Note: input and output wrappers have "display: block" propery in print mode.
+*/
+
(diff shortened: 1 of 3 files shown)


jtpio:
Diff looks good, thanks!

Mind posting a quick screenshot of what it looks like when printed if you have one around? So we have it as reference here in this PR.

SylvainCorlay:
> Diff looks good, thanks!
>
> Mind posting a quick screenshot of what it looks like when printed if you have one around? So we have it as reference here in this PR.

I added a before / after screenshot going over several pages with bqplot figures.

jtpio:
Nice, thanks!

jtpio (approved):

Issue #11598, opened 2021-12-01
Server console disappears when starting terminal in browser


Qrox:
## Description
I usually start jupyter lab server on a remote computer and connect to the server via the local computer. It has been working fine until I recently updated to jupyter lab 3.2.4 from 3.0.12. Now when I start jupyter lab server and start a terminal in the browser, the server console window disappears. I can still connect to the server from my local computer, but I cannot see the server console output because the window is hidden.

## Reproduce

1. Start jupyter lab server with command `jupyter lab --no-browser --ip 0.0.0.0 --port 8888 --certfile foo.pem --keyfile foo.key` (the command is executed as a .bat script)
2. Start a terminal session in the browser (it will start a powershell session)
3. Server console disappears

## Expected behavior

The server console is always shown, unless configured to be hidden.

## Context

- Operating System and version: Windows 10 Professional 20H2
- Browser and version: Chrome 96.0.4664.45
- JupyterLab version: 3.2.4

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
        D:\Python39\tensorflow\Scripts
        D:\NVIDIA\CUDA\bin
        D:\NVIDIA\CUDA\extras\CUPTI\lib64
        D:\NVIDIA\CUDA\include
        D:\Python39\tensorflow\Scripts
        D:\Python39\Scripts
        D:\NVIDIA\CUDA\bin
        D:\NVIDIA\CUDA\libnvvp

        D:\Git\mingw64\bin
        C:\Windows\system32
        C:\Windows
        C:\Windows\System32\Wbem
        C:\Windows\System32\WindowsPowerShell\v1.0\
        C:\Windows\System32\OpenSSH\
        C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common      
        C:\Program Files\NVIDIA Corporation\NVIDIA NvDLISR
        D:\GitExtensions\
        E:\matlab\runtime\win64
        E:\matlab\bin
        E:\matlab\polyspace\bin
        C:\WINDOWS\system32
        C:\WINDOWS
        C:\WINDOWS\System32\Wbem
        C:\WINDOWS\System32\WindowsPowerShell\v1.0\
        C:\WINDOWS\System32\OpenSSH\
        C:\Program Files\NVIDIA Corporation\Nsight Compute 2020.3.0\
        C:\Users\pc15465\AppData\Local\Microsoft\WindowsApps        

sys.path:
[...]

PR #11637, opened 2021-12-08
Add `closeOnExit` terminal option

davidbrochart:
## References

https://github.com/jupyterlab/retrolab/issues/299

## Code changes

This PR adds a new `closeOnExit` option for terminals, which defaults to `true`. If set to `false`, the terminal widget won't be disposed when exiting the terminal, which is similar to the classic notebook behavior.

## User-facing changes

No change when set to `true` (which is the default). When set to `false`, the widget is not disposed:

image

None

### DIFF
--- a/packages/services/src/terminal/terminal.ts
+++ b/packages/services/src/terminal/terminal.ts
@@ -145,14 +145,16 @@ export interface IManager extends IBaseManager {
   /**
    * Create a new terminal session.
    *
-   * @param name - The name of the target terminal.
+   * @param options - The options used to create the terminal.
    *
    * @returns A promise that resolves with the terminal connection instance.
    *
    * #### Notes
    * The manager `serverSettings` will be always be used.
    */
-  startNew(name?: string): Promise<ITerminalConnection>;
+  startNew(
+    options?: Omit<ITerminalConnection.IOptions, 'serverSettings'>
+  ): Promise<ITerminalConnection>;
 
   /*
    * Connect to a running session.
--- a/packages/services/src/terminal/terminal.ts
+++ b/packages/services/src/terminal/terminal.ts
@@ -145,23 +145,21 @@ export interface IManager extends IBaseManager {
   /**
    * Create a new terminal session.
    *
-   * @param options - The options used to create the session.
+   * @param name - The name of the target terminal.
    *
-   * @returns A promise that resolves with the terminal instance.
+   * @returns A promise that resolves with the terminal connection instance.
    *
    * #### Notes
    * The manager `serverSettings` will be always be used.
    */
-  startNew(
-    options?: ITerminalConnection.IOptions
-  ): Promise<ITerminalConnection>;
+  startNew(name?: string): Promise<ITerminalConnection>;
 
   /*
    * Connect to a running session.
    *
-   * @param name - The name of the target session.
+   * @param options - The options used to connect to the terminal.
    *
-   * @returns A promise that resolves with the new session instance.
+   * @returns The new terminal connection instance.
    */
   connectTo(
     options: Omit<ITerminalConnection.IOptions, 'serverSettings'>
--- a/packages/services/src/terminal/terminal.ts
+++ b/packages/services/src/terminal/terminal.ts
@@ -16,6 +16,15 @@ import { IManager as IBaseManager } from '../basemanager';
 import { IModel, isAvailable } from './restapi';
 export { IModel, isAvailable };
 
+export namespace ITerminal {
+  export interface IOptions {
+    /**
+     * Terminal name.
+     */
+    name: string;
+  }
+}
+
 /**
  * An interface for a terminal session.
  */
@@ -152,9 +161,7 @@ export interface IManager extends IBaseManager {
    * #### Notes
    * The manager `serverSettings` will be always be used.
    */
-  startNew(
-    options?: Omit<ITerminalConnection.IOptions, 'serverSettings'>
-  ): Promise<ITerminalConnection>;
+  startNew(options?: ITerminal.IOptions): Promise<ITerminalConnection>;
 
   /*
    * Connect to a running session.
--- a/packages/terminal-extension/schema/plugin.json
+++ b/packages/terminal-extension/schema/plugin.json
@@ -81,9 +81,9 @@
       "type": "boolean",
       "default": false
     },
-    "disposeOnExit": {
-      "title": "Dispose on exit",
-      "description": "Dispose the widget when exiting the terminal.",
+    "closeOnExit": {
+      "title": "Close on exit",
+      "description": "Close the widget when exiting the terminal.",
       "type": "boolean",
       "default": true
     },
--- a/packages/terminal-extension/schema/plugin.json
+++ b/packages/terminal-extension/schema/plugin.json
@@ -81,6 +81,12 @@
       "type": "boolean",
       "default": false
     },
+    "disposeOnExit": {
+      "title": "Dispose on exit",
+      "description": "Dispose the widget when exiting the terminal.",
+      "type": "boolean",
+      "default": true
+    },
     "pasteWithCtrlV": {
       "title": "Paste with Ctrl+V",
       "description": "Enable pasting with Ctrl+V.  This can be disabled to use Ctrl+V in the vi editor, for instance.  This setting has no effect on macOS, where Cmd+V is available",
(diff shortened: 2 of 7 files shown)


davidbrochart:
I can switch the behavior from the "Advanced Settings Editor" in the JupyterLab UI, but I don't know how to change it programmatically.

krassowski:
It might need to be added as a special case in `setOption()` handling too, like does `shutdownOnClose` (but not 100% sure of it):

https://github.com/jupyterlab/jupyterlab/blob/4f2ff432d0ddf7c44226d8038636a711775f3331/packages/terminal/src/widget.ts#L149-L151

jtpio:
Adding to `3.2.x` since this is additive.

packages/terminal/src/widget.ts
```diff
@@ -176,8 +177,10 @@ export class Terminal extends Widget implements ITerminal.ITerminal {
         });
       }
     }
-    this._term.dispose();
-    super.dispose();
+    if (this.getOption('disposeOnExit')) {
+      this._term.dispose();
+      super.dispose();
+    }
```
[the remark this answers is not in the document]
davidbrochart: For instance, when we exit a terminal with `disposeOnExit=false`, and restart JupyterLab, the workspace is restored with a terminal widget that has no terminal in the back-end:

image

We have this message in the back-end:
```
[W 2021-12-08 16:26:28.658 ServerApp] 404 GET /terminals/websocket/1 (127.0.0.1) 30.48ms referer=None
```

Maybe the behavior should be to not reopen this widget?

packages/terminal-extension/schema/plugin.json
```diff
@@ -81,6 +81,12 @@
       "type": "boolean",
       "default": false
     },
+    "disposeOnExit": {
```
fcollonval, line 84: Is _dispose_ a good term for the end user? What about `closeOnExit`?

packages/terminal/src/widget.ts
fcollonval, line 183: This is not the right place to do that because it will skip clearing the terminal resources.

This should be handle on line 74 when reacting to `session.disposed`.

fcollonval (changes requested):
Thanks @davidbrochart 

You need to place the logic in another place (see comment). I also have a small concern about the word `dispose` that may not be clear for user.

@jtpio I'm a bit reluctant to include it in a patch version because this is not a bug fix.

> notebook v7 will be based on JLab 4. So at some points the main branch of retrolab should be based on JLab 4.

packages/terminal/src/widget.ts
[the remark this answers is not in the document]
jtpio: Maybe the `disposeOnExit` check should be done on `session.disposed` here:

https://github.com/jupyterlab/jupyterlab/blob/5590af614f4585138cbd19cffe98905faf35bf24/packages/terminal/src/widget.ts#L74

To avoid preventing the calls to the parent `dispose`.

jtpio:
> I'm a bit reluctant to include it in a patch version because this is not a bug fix.

Fair enough. I thought we had already backported some features to a previous `3.2.x` release, and this could have been an additional one.

Let's target 4.0 then, unless a `3.3` comes up in the meantime.

packages/terminal-extension/schema/plugin.json
davidbrochart: `closeOnExit` sounds better, and is more aligned with the already-existing `shutdownOnClose`.

jtpio:
> notebook v7 will be based on JLab 4. So at some points the main branch of retrolab should be based on JLab 4.

Yes that makes sense. Opened https://github.com/jupyterlab/retrolab/issues/303 to track this.

davidbrochart:
With this last commit, the unclosed widget is still restored without a corresponding terminal in the back-end, when we relaunch JupyterLab.
Maybe we should adopt the same behavior as for notebooks: when we open a notebook and shut down its kernel, then restart JupyterLab, a kernel is automatically launched for the restored notebook.
Should we automatically launch a terminal in the back-end for the restored terminal widget?

fcollonval:
> Should we automatically launch a terminal in the back-end for the restored terminal widget?

:+1:  Doing that at the restoration of the UI makes sense to me.

davidbrochart:
The problem is that there is already a widget restorer, and it tries to reconnect to a terminal with the same name. That is used e.g. when you create a terminal in a JupyterLab window, and open a new window with the same server URL. In that case, the terminal widgets in both windows are in sync, any character entered in one window is visible in the other one.
In our case we get a `404` when trying to reconnect, but I'm not sure how to catch that and start a new terminal.

krassowski:
If we can find a way to restore the terminal contents it would also fix https://github.com/jupyterlab/jupyterlab/issues/11598 (I think).

davidbrochart:
Thanks @krassowski, I'll try to understand better how this can be done then.

davidbrochart:
I think this is good to go.

packages/terminal/src/widget.ts
```diff
@@ -71,7 +71,9 @@ export class Terminal extends Widget implements ITerminal.ITerminal {
     this.title.label = this._trans.__('Terminal');
 
     session.messageReceived.connect(this._onMessage, this);
-    session.disposed.connect(this.dispose, this);
+    if (this.getOption('closeOnExit')) {
+      session.disposed.connect(this.dispose, this);
+    }
```
fcollonval, line 76: If the settings is changed during a session, the new value is not taken into consideration. We need to look for the option when disposing the session.

```suggestion
    session.disposed.connect(
      () => {
        if (this.getOption('closeOnExit')) {
          this.dispose();
        }
      },
      this
    );
```

packages/services/src/terminal/terminal.ts
```diff
@@ -145,23 +145,21 @@ export interface IManager extends IBaseManager {
...
-   * @returns A promise that resolves with the terminal instance.
+   * @returns A promise that resolves with the terminal connection instance.
    *
    * #### Notes
    * The manager `serverSettings` will be always be used.
    */
-  startNew(
-    options?: ITerminalConnection.IOptions
-  ): Promise<ITerminalConnection>;
+  startNew(name?: string): Promise<ITerminalConnection>;
```
fcollonval, line 155: The options are indeed not used in the implementation. But I would recommend still passing an option object (with a new interface to define). To be coherent with other APIs in JupyterLab. Something like:

```ts
startNew(options?: ITerminal.IOptions): Promise<ITerminalConnection>;

export namespace ITerminal {
  export interface IOptions {
    name?: string;
  }
}
```

fcollonval (changes requested):
Thanks @davidbrochart to figure out how to restart a session.

I have 2 additional suggestions.

packages/terminal/src/widget.ts
davidbrochart: Thanks!

packages/services/src/terminal/terminal.ts
davidbrochart: In https://github.com/jupyterlab/jupyterlab/pull/11637/commits/3e83c95e63b22b59191075bb22d3abe5e413f110, I used the existing `ITerminalConnection.IOptions` as it's done for `connectTo`. What do you think?

fcollonval: I'm not a fan because it make the API unclear. You are passing a fake model just to get a name. But from the 3e83c95 API one would wonder, why do I need a model to start a new terminal? Should I first instantiate some model?

davidbrochart: I agree, but we could also say the same for `connectTo` (we create a "fake" model just to pass a name). Do you think we should also refactor the `connectTo` API?

fcollonval: No

Conceptually, sessions and terminals are similar. You start a new one that generates a model. And you can use that model to connect to the corresponding server resources.

We should stick with passing the model in `connectTo` even if the model is really simple for terminals.

davidbrochart: Thanks, then https://github.com/jupyterlab/jupyterlab/pull/11637/commits/f4bd0ec62ad6903f381bc26bd6d75f1b7d1cca92 should be enough.

fcollonval (approved):
Thanks for pushing this through @davidbrochart

davidbrochart:
OK to merge?

fcollonval:

Issue #10380, opened 2021-06-10
Drop the `CodeEditor` abstraction


jtpio:
In the [weekly dev meeting yesterday](https://github.com/jupyterlab/team-compass/issues/117#issuecomment-858386990), we discussed whether we should drop the [`CodeEditor`](https://github.com/jupyterlab/jupyterlab/tree/master/packages/codeeditor) abstraction. The overall sentiment was positive, with ~15 people in the call. 

This would be a [4.0](https://github.com/jupyterlab/jupyterlab/milestone/61) change.

Some core packages already rely on editors being Codemirror editors, for example the debugger extension.

Focusing on single code editor (Codemirror) would make it easier to fully leverage all the available APIs.

Related: #10370 

@jupyterlab/committers, extension authors and anyone else: feel free to leave a :+1: or :-1: to get a clear idea of the general sentiment. Thanks!

jasongrout:
Thanks for opening this. Several people in the meeting also said we should move directly to CodeMirror 6 in 4.0, and that it is pretty stable at this point.

fcollonval:
> move directly to CodeMirror 6
:+1: 

@vidartf I know that CM author decided to not port the merge view feature of CodeMirror 5 to 6: https://github.com/codemirror/codemirror.next/issues/327

This will presumably require some efforts to get nbdime and jupyterlab-git compatible with 4.0. But this is not a blocker.

bollwyvl:
Definitely a big :+1: on CM6. I think if we fully engage with it, there will be some good opportunities for using things like [DecorationSet](https://codemirror.net/6/docs/ref/#view.DecorationSet) to make huge challenges of [jupyterlab-lsp](https://github.com/krassowski/jupyterlab-lsp/issues/500) just... disappear. And the accessibility impact cannot be overstated.

However: while we're taking a breath...

I would posit that ProseMirror document editor (esp. the [code editor](https://prosemirror.net/examples/codemirror/) and [track changes](https://prosemirror.net/examples/track/) examples) show depth of integration (and composability) on top of a more-broadly-familiar UI (e.g. word processing) that we might really not want to try to re-build. Indeed, this Document-first view might end up being what a lot of different audiences might want first, and by default, even for notebooks that were _authored_ in a traditional cell view.

If these could be made to fit the existing API, then `CodeEditor` might still be useful to keep.

[...]

PR #11638, opened 2021-12-08
Migrate to Codemirror 6

### LINKED ISSUE #10370: Update Codemirror to version 6
### Problem

Currently codemirror (5.x) presents some performance and accessibility issues which have been addressed in the new version CodeMirror Next ( https://codemirror.net/6/)

### Proposed Solution

Update to use the new version

### Additional context

This version is backwards incompatible, so this would go into JLab 4.x

### Performance

@echarles worked on some performance benchmarks:

>CodeMirror Next brings performance boost but is backwards incompatible (operation + contain workarounds bring the best improvement with stable version)

codemirror-comparison

>Luminio resize event seems to be the cause of the performance degradation (disabling that event brings back the jlab cells in lumino panel in-pair with plain codemirrors).

exp

JohanMabille:
This PR is a first pass on the migration to codemirror 6. The code does not build yet, a lot is missing, however I think it could be useful to share this work at this very early stage. Many CodeMirror APIs and concepts have changed, and we have to take decisions that will impact the design of JLab:

- there is no more commands central registry in CodeMirror6. This means that we can either recreate it internally, or decide to follow the design of CodeMirror and work with `Command` and `StateCommand` in the API (see the `execCommand` function)
- there is no more options central registry, every option is a Facet/Extension in CodeMirror 6. Therefore, like for commands, we can decide to recreate an option mapping internally, or stick with Facet/Extension/Compartment in the API.
- A lot of APIs operating on the Editor have been removed, and replaced with Extensions, like the gutters. In the debugger package, we add and remove gutters whenever we toggle on / off the debugger. Dybnamically adding and removing extensions in CodeMirror6 is possible but painful and potentially costly. Besides, setting up precedence between extension can be complicated, especially when external pugins can add their own exensions.
- Given all we have to expose from the editor, I wonder if providing a getter to the CodeMirror EditorView would not be better than duplicating the API in the Codemirror package of JupyterLab.

~~EDIT: this PR needs to be rebased, please consider the last four commits only.~~

---

Fix #10370


JohanMabille:
The code now builds. And free benefit: no more bug in your notebooks! :laughing: 
no_more_bugs_in_notebook

JohanMabille:
Bug fixed, and YJS bindings added back.

fcollonval:
> * Given all we have to expose from the editor, I wonder if providing a getter to the CodeMirror EditorView would not be better than duplicating the API in the Codemirror package of JupyterLab.

This is one [objective](https://github.com/jupyterlab/jupyterlab/issues/10380) for 4.0. So for your comment could it be worth to drop the JupyterLab abstraction in a preliminary PR? Then pushing CodeMirror 6 instead of packing everything in this PR.

JohanMabille:
I think it's easier to remove the abstraction after migrating to CodeMirror 6 (this way the migration is somehow contained). I will drop the abtrsaction in a dedicated PR when this one is finished and merged.

echarles:
> I think it's easier to remove the abstraction after migrating to CodeMirror 6 (this way the migration is somehow contained). I will drop the abtrsaction in a dedicated PR when this one is finished and merged.

Sounds reasonable having 2 steps 👍

JohanMabille:
The last issue I have is that I would like to use the [mkPython](https://github.com/codemirror/legacy-modes/blob/main/mode/python.js#L27) function for the ipython syntax highlighting, but it is not declared in the [python.d.ts file](https://github.com/codemirror/legacy-modes/blob/main/mode/python.d.ts).

I tried to "forward declare it" (yeah, be indulgent, I'm a C++ developer ^^), with something like `declare let mkPython: any;` but it fails at runtime.

Any help would be appreciated.

krassowski:
It looks not too far off. Maybe https://stackoverflow.com/a/53098695/6646912 could help?

fcollonval:
You should be able to fix this by overriding the type definition.

```ts
// packages/codemirror/src/typings.d.ts

declare module '@codemirror/legacy-modes/mode/python' {
  import { StreamParser } from '@codemirror/stream-parser';

  export const python: StreamParser<unknown>;
  
  export const cython: StreamParser<unknown>;

  export function mkPython(parserConf: any): StreamParser<unknown>;
}
```

Then in `packages/codemirror/src/codemirror-ipython.ts`

```diff
+ import { mkPython } from '@codemirror/legacy-modes/mode/python';

- // Stub for the mkPython function.
-  declare let mkPython: any;

**Comment by fcollonval** [comment]
> Side note you could also open a PR upstream to fix the typing :wink:

**Comment by JohanMabille** [comment]
> > Side note you could also open a PR upstream to fix the typing wink

Yes but not sure it will be accepted, and I want to have something to show quickly ;) But I will definitely open a PR upstream.

**Comment by JohanMabille** [comment]
syntax_highlighting
Syntax hilghlighting is back!

**Comment by jasongrout** [comment]
Note: this would fix https://github.com/jupyterlab/jupyterlab/issues/10370

**Comment by jtpio** [comment]
Looks like CodeMirror 6.0 has been released:

- https://marijnhaverbeke.nl/blog/codemirror-6.html
- https://discuss.codemirror.net/t/codemirror-6-0-has-been-released/4498

Which is good news since this was one of the points raised during the [Migration to CodeMirror 6 presentation](https://www.youtube.com/watch?v=Jtl7mLP1LKU) that could have been an issue for adopting it for JupyterLab 4.0.

**Review by fcollonval — COMMENTED** [review]
please run benchmarks

**Review by fcollonval — COMMENTED** [review]
please run benchmark

**Comment by jtpio** [comment]
Just pushed a commit to fix the integrity check: https://github.com/jupyterlab/jupyterlab/pull/11638/commits/a485d34cd24924486aa5860511b84a6310c5c3e5

image

It looks like there were some references to the `@codemirror/basic-setup` package defined as a dependency, although not used anywhere in the code.

Also the Binder for this PR seems to be building fine, which means folks can already start testing it using this link: https://mybinder.org/v2/gh/JohanMabille/jupyterlab/codemirror?urlpath=lab-dev

image

**Comment by williamstein** [comment]
Found another one -- markdown cells aren't syntax highlighted.  Code is fine, but not markdown cells:

image

**Comment by aiqc** [comment]
So one of the things I was looking forward to in codemirror 6 is indentation lines - [discussion](https://stackoverflow.com/questions/67991836/how-to-get-indentation-lines-in-jupyter-lab-jupyter-notebook)

image

Is it possible to enable them by default?

**Comment by JohanMabille** [comment]
@williamstein thanks for reporting! I'm actually fixing the selection marks, will jump on this bugs after!
@aiqc I think I had it working in some previous commits of this PR, so this looks like a regression, I'll have a look.

**Comment by williamstein** [comment]
@JohanMabille  you're very welcome.  I'm likely to be doing a similar migration for Cocalc in the not-to-distant future, and greatly appreciate you doing all this work now, since I'm sure I'll learn a lot from it!

**Comment by JohanMabille** [comment]
@williamstein the last commit fixes the cursor issues.

**Comment by krassowski** [comment]
Some things that do not seem to work on binder:
- [ ] indenting code with <kbd>tab</kbd> does not work
- [ ] expanding to full height available of the editor
Screenshot from 2022-06-29 18-36-01
- [ ] scrolling does not work for me in files (paste a long text and try to scroll)

Super impressed that grammarly just works, though I think we should tell browser to not spell check in code cells:

Screenshot from 2022-06-29 18-35-27

Surely a lot of work needed on accessible colors in dark mode:

Screenshot from 2022-06-29 18-38-40

The line numbers are not coloured in dark mode:

Screenshot from 2022-06-29 18-39-44

**Comment by JohanMabille** [comment]
@krassowski thanks for your review. The last commits should fix all your issue but the dark theme. I think we can fastly iterate on the css and the themes when this is merged.

File: packages/codemirror/src/codemirror-ipythongfm.ts [hunk 1]
```diff
@@ -1,11 +1,30 @@
...
+// load: () => {
+//   return import('@codemirror/lang-markdown').then(m => m.markdown({extensions: latex}));
+// }
+
+/*import CodeMirror from 'codemirror';
 import 'codemirror/addon/mode/multiplex';
 import 'codemirror/mode/gfm/gfm';
 import 'codemirror/mode/stex/stex';
-
+*/
```
krassowski, line 27: Why not import other modes, including stex from `codemirror/legacy-modes`?

JohanMabille: These modes are imported to be combined into a new one thanks to `CodeMirror.multiplexingMode` which is not available anymore in CodeMirror 6. Modes should now be based on Lezer and I am not sure we can reuse the legacy multiplex addon.

JohanMabille:
The example failure is unrelated to this PR; other PR changing the yarn.lock file have experienced the same issue, it is due to importing twice yjs, not sure where the problem comes from.

The failures of `check_release` is due to a connection failure, not related to this PR.

Only the visual regression failures are expected.

JohanMabille:
please update snapshots

packages/codemirror/src/editor.ts
```diff
@@ -662,124 +666,75 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
...
-    this.doc.replaceSelection(text);
-  }
-
-  /**
-   * Get a list of tokens for the current editor text content.
-   */
-  getTokens(): CodeEditor.IToken[] {
-    let tokens: CodeEditor.IToken[] = [];
-    for (let i = 0; i < this.lineCount; ++i) {
-      const lineTokens = this.editor.getLineTokens(i).map(t => ({
```
fcollonval, line 674: Seems that we needs to loop on `syntaxTree` to get something similar

packages/codemirror/src/editor.ts
```diff
@@ -802,24 +757,21 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
   /**
    * Clean selections for the given uuid.
    */
+  // TODO: CM6 migration see Mark Text section of the migration guide
```
fcollonval, line 760: Done?

packages/codemirror/src/editor.ts
```diff
@@ -1369,262 +1106,130 @@ export namespace CodeMirrorEditor {
...
-        lines.style.maxWidth = `${wordWrapColumn}ch`;
-      }
-      host.appendChild(el);
-    }, bareConfig);
+    config: CodeMirrorEditor.IConfig,
+    ybinding: IYCodeMirrorBinding | null,
+    editorConfig: Configuration.EditorConfiguration,
+    additionalExtensions: Extension[]
+  ): EditorView {
+    let extensions = editorConfig.getInitialExtensions(config);
```
fcollonval, line 1142:
```suggestion
    const extensions = editorConfig.getInitialExtensions(config);
```

packages/codemirror/src/editor.ts
```diff
@@ -1369,262 +1106,130 @@ export namespace CodeMirrorEditor {
...
-/**
- * Add a CodeMirror command to delete until previous non blanking space
- * character or first multiple of tabsize tabstop.
- */
-CodeMirrorEditor.addCommand(
-  'delSpaceToPrevTabStop',
-  Private.delSpaceToPrevTabStop
-);
+    eq(other: CaretWidget) {
+      return other.collaborator.sessionId == other.collaborator.sessionId;
```
fcollonval, line 1194: I don't get this - is that not always true.

packages/codemirror/src/editor.ts
```diff
@@ -1369,262 +1106,130 @@ export namespace CodeMirrorEditor {
...
+      // TODO: check if this should be replaced with
+      // a global event handler
+      caret.onmouseenter = () => {
+        this.callbacks.setHoverId(this.collaborator.sessionId);
+        const rect = caret.getBoundingClientRect();
+        // Construct and place the hover box.
+        const hover = document.createElement('div');
+        hover.className = COLLABORATOR_HOVER_CLASS;
+        hover.style.left = String(rect.left) + 'px';
+        hover.style.top = String(rect.bottom) + 'px';
```
fcollonval, line 1211:
```suggestion
        hover.style.left = `${rect.left}px`;
        hover.style.top = `${rect.bottom}px`;
```

packages/codemirror/src/editorconfiguration.ts
```diff
@@ -0,0 +1,497 @@
...
+
+    /**
+     * Defines the mouse cursor appearance when hovering over the selection.
+     * It can be set to a string, like "pointer", or to true,
+     * in which case the "default" (arrow) cursor will be used.
+     */
+    selectionPointer: boolean | string;
+  }
+
+  interface IExtensionBuilder {
```
fcollonval, line 164: Docuemntation for all those classes will be more than welcome

packages/codemirror/src/editorconfiguration.ts
```diff
@@ -0,0 +1,497 @@
...
+     */
+    scrollPastEnd?: boolean;
+
+    /**
+     * Whether to give the wrapper of the line that contains the cursor the class
+     * CodeMirror-activeline, adds a background with the class
+     * CodeMirror-activeline-background, and adds the class
+     * CodeMirror-activeline-gutter to the line's gutter space is enabled.
+     */
+    styleActiveLine: boolean; //| CodeMirror.StyleActiveLine;
```
fcollonval, line 147: Is the doc string still valid?

```suggestion
    styleActiveLine: boolean;
```

packages/codemirror/src/editorconfiguration.ts
```diff
@@ -0,0 +1,497 @@
...
+      );
+
+      let themeOverload = this.updateThemeOverload(config);
+      extensions.push(
+        this._themeOverloader.of(themeOverload),
+        insertExt,
+        keymapExt
+      );
+
+      Mode.ensure('text/x-python')
```
fcollonval, line 425: Why is that string encoded? Should it be `config.mode` instead?

packages/codemirror/src/factory.ts
```diff
@@ -19,7 +22,15 @@ export class CodeMirrorEditorFactory implements IEditorFactoryService {
...
-      extraKeys: {
+      extraKeys: [
+        {
+          key: 'Tab',
+          run: CodeMirrorEditor.indentMoreOrInsertTab,
+          shift: indentLess
+        },
+        ...searchKeymap
+      ],
+      /*extraKeys: {
```
fcollonval, line 33: dead code?

packages/codemirror/src/factory.ts
```diff
@@ -29,20 +40,33 @@ export class CodeMirrorEditorFactory implements IEditorFactoryService {
...
+          shift: indentLess
+        },
+        {
+          key: 'Shift-Enter',
+          run: (target: EditorView) => {
+            return true;
+          }
+        }
+      ],
+      /*extraKeys: {
```
fcollonval, line 61: Dead code?

packages/codemirror/src/index.ts
```diff
@@ -11,11 +11,11 @@ import { CodeMirrorMimeTypeService } from './mimetype';
 
 export * from './editor';
 export * from './factory';
+export * from './searchprovider';
 export * from './mimetype';
 export * from './mode';
-export * from './searchprovider';
 export * from './syntaxstatus';
-export * from './tokens';
+//export * from './tokens';
```
fcollonval, line 18:
```suggestion
```

packages/codemirror/src/mode.ts
```diff
@@ -3,199 +3,1603 @@
...
+  // Code mirror uses two similar structures, a plain object with optional fields,
+  // and a class with the same fields but all mandatory. Maybe adopting the same
+  // pattern would be less confusing (although far more verbose)
+  function makeSpec(spec: ISpec): ISpec {
+    let res = LanguageDescription.of(spec) as unknown as ISpec;
+    res.mime = spec.mime;
+    return res;
+  }
+
+  const goldorak = [
```
fcollonval, line 55: Let's use a more appropriate name :wink:

packages/codemirror/src/mode.ts
```diff
@@ -3,199 +3,1603 @@
...
+          if (info.mime[j] == mime) {
+            return info;
+          }
+        }
+      } else if (info.mime == mime) {
+        return info;
+      }
+    }
+    if (/\+xml$/.test(mime)) return findByMIME('application/xml');
+    if (/\+json$/.test(mime)) return findByMIME('application/json');
```
fcollonval, line 1474: Could you add a comment why those two are needed a specific treatment?

packages/codemirror/src/searchprovider.ts
```diff
@@ -291,10 +377,9 @@ export class CodeMirrorSearchProvider implements IBaseSearchProvider {
    */
   readonly isReadOnly = false;
 
-  private _onDocChanged(_: any, changeObj: CodeMirror.EditorChange): void {
-    // If we get newlines added/removed, the line numbers across the
-    // match state are all shifted, so here we need to recalculate it
-    if (changeObj.text.length > 1 || (changeObj.removed?.length ?? 0) > 1) {
+  private _onDocChanged(view: ViewUpdate): void {
+    // TODO: CM6 migration
```
fcollonval, line 381: TODO

packages/codemirror/test/editor.spec.ts
```diff
@@ -362,10 +357,15 @@ describe('CodeMirrorEditor', () => {
   describe('#getPositionForCoordinate()', () => {
     it('should get the window coordinates given a cursor position', () => {
       model.value.text = TEXT;
-      const coord = editor.getCoordinateForPosition({ line: 10, column: 1 });
+      const coord = editor.getCoordinateForPosition({ line: 10, column: 2 });
       const newPos = editor.getPositionForCoordinate(coord)!;
-      expect(newPos.line).toBeTruthy();
-      expect(newPos.column).toBeTruthy();
+      if (typeof process !== 'undefined') {
```
fcollonval, line 362: what is `process`?

packages/codemirror/test/mode.spec.ts
```diff
@@ -1,28 +1,31 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
-/// <reference types="../typings/codemirror/codemirror" />
+
+// TODO CM6:
```
fcollonval, line 4: TODO we should provide and test adding a new mode.

packages/completer/src/default/contextprovider.ts
```diff
@@ -52,12 +52,13 @@ namespace Private {
   export function contextHint(
     editor: CodeEditor.IEditor
   ): CompletionHandler.ICompletionItemsReply {
+    // TODO: CM6
```
fcollonval, line 55: TODO

packages/debugger-extension/src/index.ts
```diff
@@ -808,10 +808,11 @@ const main: JupyterFrontEndPlugin<void> = {
           if (breakpoint && typeof breakpoint.line !== 'undefined') {
             results.forEach(editor => {
               if (editor instanceof CodeMirrorEditor) {
-                (editor as CodeMirrorEditor).scrollIntoViewCentered({
+                // TODO: CM6
```
fcollonval, line 811: TODO

packages/debugger/src/factory.ts
```diff
@@ -28,13 +28,13 @@ export class ReadOnlyEditorFactory {
...
+    const { content /*, mimeType, path*/ } = source;
     const factory = this._services.factoryService.newInlineEditor;
-    const mimeTypeService = this._services.mimeTypeService;
+    //const mimeTypeService = this._services.mimeTypeService;
     const editor = new CodeEditorWrapper({
       model: new CodeEditor.Model({
-        value: content,
-        mimeType: mimeType || mimeTypeService.getMimeTypeByFilePath(path)
+        value: content
+        //mimeType: mimeType || mimeTypeService.getMimeTypeByFilePath(path)
```
fcollonval, line 37: Is this missing migration or API changes?

packages/debugger/src/handlers/editor.ts
```diff
@@ -111,15 +180,50 @@ export class EditorHandler implements IDisposable {
...
+
+    this._uglyCM6Workaround();
     this._addBreakpointsToEditor();
+  }
 
+  // There is currently a bug in the gutter of CM6, gutter
+  // elements without any marker don't get a DOM class assigned.
+  // The bug is fixed in the master branch, we can remove this
+  // when it is released.
+  private _uglyCM6Workaround(): void {
```
fcollonval, line 213: Could you add a link to the PR or a mention of the version that will contain the bug fix

packages/debugger/test/debugger.spec.ts
```diff
@@ -346,6 +346,7 @@ describe('Debugger', () => {
       items.forEach(item => {
         // TODO: replace by toEqual when there is an alternative to the rtl
         // breakpoint display
+        console.log(item.innerHTML);
```
fcollonval, line 349:
```suggestion
```

packages/docregistry/src/default.ts
```diff
@@ -252,7 +252,7 @@ export class TextModelFactory implements DocumentRegistry.CodeModelFactory {
    */
   preferredLanguage(path: string): string {
     const mode = Mode.findByFileName(path);
-    return mode && mode.mode;
+    return mode ? mode.name : '';
```
fcollonval, line 255:
```suggestion
    return mode?.name ?? '';
```

packages/docregistry/test/default.spec.ts
```diff
@@ -539,8 +539,8 @@ describe('docregistry/default', () => {
     describe('#preferredLanguage()', () => {
       it('should get the preferred kernel language given an extension', () => {
         const factory = new TextModelFactory();
-        expect(factory.preferredLanguage('.py')).toBe('python');
-        expect(factory.preferredLanguage('.jl')).toBe('julia');
+        expect(factory.preferredLanguage('.py')).toBe('Python');
```
fcollonval, line 542: Should we lower case the language to reduce behavior changes?

packages/fileeditor/src/searchprovider.ts
```diff
@@ -28,8 +28,8 @@ export class FileEditorSearchProvider
    * @param widget File editor panel
    */
   constructor(widget: FileEditorPanel) {
-    super();
-    this.editor = widget.content.editor as CodeMirrorEditor;
+    super(widget.content.editor as CodeMirrorEditor);
+    //this.editor = widget.content.editor as CodeMirrorEditor;
```
fcollonval, line 32:
```suggestion
```

fcollonval (dismissed):
I did a first review. Let me know how I can help with the todos.

Note: we definitely need a follow-up to provide a token to extend the default list of CM extension, add mode and add theme. This can be part of the follow-up PR dropping CodeEditor.

packages/codeeditor/src/editor.ts
```diff
@@ -685,12 +685,12 @@ export namespace CodeEditor {
...
      * Gets the token at a given position.
      */
-    getTokenForPosition(position: IPosition): IToken;
+    //getTokenForPosition(position: IPosition): IToken;
 
     /**
      * Gets the list of tokens for the editor model.
      */
-    getTokens(): IToken[];
+    //getTokens(): IToken[];
```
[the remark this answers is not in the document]
JohanMabille: Yes, CM6 provide an API to iterate over syntax tree that can be used directly from the `state` object. Building an array of Token from that iterator would be inefficient, so let's drop this API.

packages/codemirror-extension/src/index.ts
```diff
@@ -268,29 +269,30 @@ function activateEditorCommands(
     keyMap = (settings.get('keyMap').composite as string | null) || keyMap;
 
     // Lazy loading of vim mode
-    if (keyMap === 'vim') {
+    // TODO: CM6 migration
```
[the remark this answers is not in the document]
JohanMabille: it did but in an external package. I haven't created a mapping "keymap" <-> kybindings in the editorconfiguration object, maybe we will need to do it (as I did for the themes).

packages/codemirror-extension/src/index.ts
```diff
@@ -268,29 +269,30 @@ function activateEditorCommands(
...
 
     scrollPastEnd =
       (settings.get('scrollPastEnd').composite as boolean | null) ??
       scrollPastEnd;
     styleActiveLine =
-      (settings.get('styleActiveLine').composite as
-        | boolean
-        | CodeMirror.StyleActiveLine) ?? styleActiveLine;
+      (settings.get('styleActiveLine').composite as boolean) ??
+      /*| CodeMirror.StyleActiveLine*/ styleActiveLine;
```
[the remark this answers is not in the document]
JohanMabille: nope, the type `StyleActiveLine` does not exist anymore. I'll remove the comment.

packages/codeeditor/src/editor.ts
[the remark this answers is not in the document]
fcollonval: But that will bring a regression on the completer, isn't it?

packages/codemirror-extension/src/index.ts
```diff
@@ -466,6 +457,7 @@ function activateEditorCommands(
   });
 
   if (mainMenu) {
+    // TODO: CM6 migration
```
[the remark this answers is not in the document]
JohanMabille: I forgot to remoe the TODO, the migration is actually done.

packages/codemirror/src/editor.ts
```diff
@@ -312,12 +332,13 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
     this.host.removeEventListener('focus', this, true);
     this.host.removeEventListener('blur', this, true);
     this.host.removeEventListener('scroll', this, true);
-    if (this._yeditorBinding) {
+    /*if (this._yeditorBinding) {
       this._yeditorBinding.destroy();
-    }
+    }*/
```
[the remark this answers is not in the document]
JohanMabille: Yes they are.

packages/codemirror/src/editor.ts
```diff
@@ -420,21 +440,21 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
...
-    this._editor.getInputField().blur();
+    this._editor.contentDOM.blur();
   }
 
   /**
    * Repaint editor.
    */
   refresh(): void {
-    this._editor.refresh();
+    //this._editor.refresh();
```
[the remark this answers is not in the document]
JohanMabille: This should not be required anymore and CM6 does not provide any API for that.

packages/codemirror/src/editor.ts
JohanMabille: yes ;)

packages/codemirror/src/mode.ts
JohanMabille: Indeed this was "review quality indicator" :D

packages/debugger/src/handlers/editor.ts
JohanMabille: The fix has been released, this code is not needed anymore, I will remove it.

packages/codemirror/test/mode.spec.ts
JohanMabille: This can be done in a dedicated PR.

JohanMabille:
please update snapshots

JohanMabille:
please update snapshots

JohanMabille:
please update snapshots

JohanMabille:
I don't understand why the Visula Regression Documentation still fails after uploading the screenshots.

jtpio:
> I don't understand why the Visual Regression Documentation still fails after uploading the screenshots.

I just checked the report available here: https://github.com/jupyterlab/jupyterlab/actions/runs/2677483924

And there seems to be only 1 failure. Not sure really this is relevant or whether it should be regenerated as there doesn't seem to be major differences?

image

fcollonval:

fcollonval:
The failure is probably not relevant - there are only 

> 2 pixels (ratio 0.01 of all image pixels) [...] different

packages/codemirror/test/editor.spec.ts
fcollonval: Bumping

packages/debugger/style/breakpoints.css
```diff
@@ -55,25 +55,34 @@
...
 
-[data-jp-debugger='true'].jp-Editor .CodeMirror-gutter-wrapper:hover::after,
+.cm-gutter {
+  cursor: default;
+}
+
+.cm-breakpoint-gutter .cm-gutterElement {
+  color: var(--jp-error-color1);
+  padding-left: 5px;
+  font-size: 20px;
```
fcollonval, line 71: Should it be related to code font size?

fcollonval (changes requested):
From tests, here are items that feel bad:

- Line height is larger
- Breakpoint indicator is bigger and it is not vertically aligned with the line number.
- Markdown highlight for code block and inline block is missing.
  See [that example to implement it](https://codemirror.net/try/?example=Markdown%20code%20block%20highlighting)
- [known bug] LaTeX highlighting is missing.
- Python built-in functions like `print`, `open`, `range`,... are not highlighted (missing in codemirror lexer)
- In a raw cell and text file some words like `is` or `in` get highlighted in bold green and first sentence word after a dot are highlighted in blue.
- Go to line editor UI test is broken (the text is not inserted)
- Go to line from status bar is not scrolling to the target line
- Some options are not existing - we need to list them somewhere
  - rulers UI test is broken (rulers lines are not appearing)
  - cursor blink rate seems missing
  - trailing white spaces seems missing

packages/codemirror/src/editor.ts
```diff
@@ -146,82 +226,24 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
...
-    );
+    this._yeditorBinding = {
+      text: sharedModel.ysource,
+      awareness: sharedModel.awareness,
+      undoManager: sharedModel.undoManager
+    };
   }
 
+  save: () => void;
+
```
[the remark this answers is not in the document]
JohanMabille: It i required to get [this](https://github.com/jupyterlab/jupyterlab/blob/master/packages/codemirror-extension/src/index.ts#L210) working

packages/debugger/style/breakpoints.css
JohanMabille: This was an attempt to get the breakpoint aligned. Otherwise, the breakpoint is in the upper left corner.

JohanMabille:
please update snapshots

SylvainCorlay:
This is a major migration that is required for JupyterLab 4.0 - and we will keep finding improvements to make to it, but there are over 150 files modified and each rebase gets in the way of working on other aspects of the codebase. I am in favor of merging this as soon as possible.

jtpio:
Linking to https://github.com/jupyterlab/jupyterlab/issues/12812 where follow-up fixes and improvements can be tracked.

Most of them should be self-contained and indeed easier to pick up by others as well once this first (big) PR is merged.

Issue #11556, opened 2021-11-26
`experimentalDisableDocumentWideUndoRedo` breaks undo of completer


krassowski:
## Description

When `experimentalDisableDocumentWideUndoRedo` (#10791) is enabled the undo does not work on text inserted by completer.

## Reproduce

1. Go to settings -> `Notebook` and set:
    ```js
    {
        "experimentalDisableDocumentWideUndoRedo": true
    }
    ```
2. Open a new notebook
3. Type `f<tab>`, select "from", type "a" select "abc", type "im" select "import", type "A" select "ABCMeta"
4. See `from abc import ABCMeta`
4. Press ctrl + Z three times
5. See `fromabcimportABCMeta` - spaces were removed but the text that was auto-completed was not

![problem-with-undo](https://user-images.githubusercontent.com/5832902/143573393-b59689f4-2389-472d-ac26-4f8fc43c49d5.gif)

## Expected behavior

Undo removes completed text when document-wide undo/red is disabled.

## Context

- Browser and version: Chrome 96
- JupyterLab version: 3.2.4

cyaris:
Same issue for me.

suganya-sk:
Hello! This issue is still present in version 3.4.3. Any updates on a timeline here please?

nishikantparmariam:
+1 for this, I encountered it while writing an extension. Looking forward to https://github.com/jupyterlab/jupyterlab/pull/14063!

I am using below API which completer also uses:

https://github.com/jupyterlab/jupyterlab/blob/b806333290052c3d549876213191a1e2094793ee/packages/completer/src/handler.ts#L170

But, running into the same problem as above. Is there a cheap workaround here other than setting `experimentalDisableDocumentWideUndoRedo` to `false`?

Issue #10791, opened 2021-08-09
Allow to disable the document-wide history tracking


krassowski:
#10694 highlighted some issues with the document-wide history tracking relating to outputs, but in the discussion it also came up that it might be a good idea to allow to disable this feature entirely. @mlucool noted that this behavior is detrimental to the workflow of some users (https://github.com/jupyterlab/jupyterlab/issues/10694#issuecomment-891390472).

During the weekly call (https://github.com/jupyterlab/team-compass/issues/128#issuecomment-892826198) @williamstein provided very useful feedback:
1. (against) teachers may want to have full history tracking available
2. (in favour) users sometimes mistakenly paste secretes into notebooks

The first issue does not apply to online sessions where a concern of some disruption to the lesson/vandalism could be a considerations since *I think* we would only want to allow to disable the tracking when not in RTC mode. Still there are some settings (exams?) where instructors may want to keep the history tamper-proof, should an ability to view it be introduced in the future.

The second issue could indeed be a substantial problem once #10544 gets implemented.

I would add another point related to (1):

3. There are regulatory settings (e.g. statistical analysis of clinical trials, see 21 CFR Part 11) where an audit trial record is required; some statistical analysis software products implement it by setting a flag on startup which persists through the session, this is the history tracking cannot be disabled without restarting the application (which clears the history/restarts the signature keys making it tamper-proof on the sufficient level to comply with the regulations).

Given the points (1) and (3) I think we can allow to opt-out of the document-wide history tracking, and it would make sense to require a server restart (so the setting should be implemented on server level).

mlucool:
FWIW, I think its perfectly reasonable to always have document level history so long as users can continue to have cell level redo/undo as in previous version of Jupyter. The problem users like being able to think of cells closer to how IDEs treat files. This means sometimes undo between files has meaning at times (e.g. undo a rename), but ctrl-z inside a file (i.e. in a codecell) means undo my last action in this block.

As shown here, I push ctrl+z twice in the top cell but start removing code from another cell (see the +2 get deleted)!
![cross-cell-undo](https://user-images.githubusercontent.com/1813603/128782023-0fd0ff05-421d-4f18-8d1b-9cfb2651f933.gif)

Even worse, at times it seems ctrl+z removes larger swaths of work including deleting cells:
![buggy-undo](https://user-images.githubusercontent.com/1813603/128782150-e467dc8a-7015-4d71-a92f-a7dbab46c575.gif)

[...]

PR #11640, opened 2021-12-08
[WIP] cell-based undo manager.

dmonad:
We switched to a global undo-redo manager for Jupyter notebooks when we integrated shared editing. This lead to some frustration for several users that were used to the local (cell-based) undo manager. https://github.com/jupyterlab/jupyterlab/issues/10791

#10949  aims to revert back to the old behavior by disabling the current collaborative undo manager so that CodeMirror's native undo manager is used. However, when the collaborative undo manager is disabled, you will be able to undo changes from remote users. Aside from that, it seems that disabling the collaborative undo manager leads to some problems (https://github.com/jupyterlab/jupyterlab/issues/11556). 

I've been working on a move feature in Yjs for quite some time now. It allows us to move cells (or ranges of cells) from one place to another while keeping the associated undo-manager to that cell. The undo-history will be retained even when a cell is moved from one place to another. 

In short, this PR implements a proper cell-based undo manager in JupyterLab. In ticket #10791 it was mentioned that switching to a global undo manager was a breaking change. So we should probably make this behavior the default again.

I also would like to add a flag "globalUndo" flag that will replace the current experimental feature "disableDocumentWideUndoRedo".

This PR closes #10949, #11556, & https://github.com/jupyterlab/jupyterlab/issues/10791#issuecomment-943552144 

## Code changes

Currently, when we move a cell from one place to another, we delete the old cell in shared-model and copy it to a new position. This PR uses the new move feature in Yjs to move the cell. This retains the editing history for that cell.

## User-facing changes

Switching back to a local undo manager. 

This change is backward compatible, but not forward compatible. The PR implements a new feature in shared-model. Old versions of shared-model don't understand the move-semantics in Yjs and won't be able to read the Yjs content. However, I don't think that this will be a problem since there is no reason to use two different versions of shared-model in the same project.

---------

This PR is still in an experimental state. It uses a prerelease of Yjs that still has some known bugs (It's unlikely that you will catch them). I want to invite others to use this PR and give feedback.

@tonyfast

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -121,7 +121,7 @@
     "@lumino/widgets": "^1.30.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "yjs": "^13.5.17"
+    "yjs": "next"
   },
   "dependencies": {
     "@jupyterlab/application": "~3.3.0-alpha.15",
--- a/jupyterlab/staging/package.json
+++ b/jupyterlab/staging/package.json
@@ -121,7 +121,7 @@
     "@lumino/widgets": "^1.28.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "yjs": "^13.5.17"
+    "yjs": "next"
   },
   "dependencies": {
     "@jupyterlab/application": "~3.3.0-alpha.15",
--- a/packages/docprovider/package.json
+++ b/packages/docprovider/package.json
@@ -42,7 +42,7 @@
     "@lumino/coreutils": "^1.11.1",
     "lib0": "^0.2.42",
     "y-websocket": "^1.3.15",
-    "yjs": "^13.5.17"
+    "yjs": "next"
   },
   "devDependencies": {
     "@jupyterlab/testutils": "^3.3.0-alpha.15",
--- a/packages/docregistry/package.json
+++ b/packages/docregistry/package.json
@@ -59,7 +59,7 @@
     "@lumino/messaging": "^1.10.1",
     "@lumino/signaling": "^1.10.1",
     "@lumino/widgets": "^1.30.0",
-    "yjs": "^13.5.17"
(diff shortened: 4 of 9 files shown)


jtpio:
Thanks @dmonad for this!

jtpio:
Looks like there are a couple of conflicts to fix.

Also setting this one for `4.0` for now, although we should be able to backport it to 3.3.x.

dev_mode/package.json
```diff
@@ -121,7 +121,7 @@
     "@lumino/widgets": "^1.30.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "yjs": "^13.5.17"
+    "yjs": "next"
```
jtpio, line 124: Looks like there has been a couple of `yjs` releases since.

Is it already possible to update to `^13.5.27` for instance?

jtpio: For reference the move feature is being implemented in Yjs in this PR: https://github.com/yjs/yjs/pull/357

dmonad: The moving feature is currently only available in the `next` release which points to the Yjs alhpa release that contains experimental features.

The new move feature seems to work well so far. I collected feedback and some suggestions for changes and need another day of work to finish this and merge it into the Yjs main. I want to wait for funding and a new contract before I continue working on this.

I recommend against merging this PR as it points towards the alpha release of Yjs.

krassowski:
What is the status of this PR?

krassowski:
@hbcarlos do you know if this could be revived?

hbcarlos:
Hi @krassowski, sorry I missed your previous comment.

This PR is outdated. We need to implement it in https://github.com/jupyter-server/jupyter_ydoc. Besides, now we have ypy in the backend to load documents. That means we need the move feature in both yjs and ypy. We are working on it.

PR #11642, opened 2021-12-09
pyproject.toml: clarify build system version

adamjstewart:
See https://github.com/jupyter/jupyter-packaging/issues/120

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires=["jupyter_packaging~=0.9,<2"]
+requires=["jupyter_packaging~=0.9"]
 build-backend = "jupyter_packaging.build_api"
 
 [license]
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @adamjstewart

jtpio:
A follow-up to this would be to update `jupyter_packaging>=0.10,<2` to as mentioned in https://github.com/jupyter/jupyter-packaging/pull/121

adamjstewart:
> A follow-up to this would be to update `jupyter_packaging>=0.10,<2` to as mentioned in [jupyter/jupyter-packaging#121](https://github.com/jupyter/jupyter-packaging/pull/121)

Unless the build no longer works with 0.9 or requires features introduced in 0.10, I see no reason to change the lower bound. As for the upper bound, I'm happy to change that if you want me to, was just trying to keep the behavior the same until told otherwise.

jtpio:
Thanks @adamjstewart.

I pushed a small commit to update to `jupyter_packaging>=0.9,<2`

jtpio (approved):
Thanks!

PR #11644, opened 2021-12-09
Create first_program.rst

betrabes:
In the JupyterLab docs page - https://jupyterlab.readthedocs.io/en/latest/index.html - under the section User Guide, there is no section that walks the user through actually creating a useful program. My documentation tries to solve this problem by creating a page called “First Program” that walks a beginner user through the process of doing something useful in JupyterLab, in this case creating some plots of some data.

### DIFF
--- a/docs/source/user/first_program.rst
+++ b/docs/source/user/first_program.rst
@@ -1,0 +1,77 @@
+.. first-program:
+
+First Program
+=======================
+
+Setting up a new program is always difficult at first. Hopefully, this tutorial will guide you
+through a common use of Jupyter Lab. This tutorial will go over graphing some data using matplotlib. 
+When you are done with this tutorial, you will be able to get all started with JupyterLab. 
+
+
+Start by running 'pip install jupyterlab' in the command line
+
+Then go to https://fred.stlouisfed.org/series/T5YIE and download the data in the csv format.
+
+Open the jupyterlab software by running: jupyter notebook
+
+Open a new python file by clicking on the new dropdown menu
+
+Import the csv file that you downloaded into jupyterlab
+
+Paste the following code:
+
+/////////////////////////////////////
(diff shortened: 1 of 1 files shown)


fcollonval:
Thanks @betrabes could you clarify the goal you would like to achieve or the pain point you are trying to remove?

Regarding the technical aspect, documentation should be contained in the project (so the PR should include the illustrations). And you need to format the text using [ReStructuredText](https://docutils.sourceforge.io/docs/user/rst/quickref.html) syntax.

In particular, see thoses examples:

- code snippet: https://github.com/jupyterlab/jupyterlab/blame/master/docs/source/extension/extension_tutorial.rst#L227
- image: https://github.com/jupyterlab/jupyterlab/blame/master/docs/source/extension/extension_tutorial.rst#L377
- URL link: https://github.com/jupyterlab/jupyterlab/blame/master/docs/source/extension/extension_tutorial.rst#L41

krassowski:
An alternative to adding text (which needs to be maintained, written for multiple languages etc) would be to add links to actual notebooks in [jupyterlab-demo](https://github.com/jupyterlab/jupyterlab-demo) repository running on Binder; we already have example notebooks for Julia, Python, R and C++; we could add a direct link in the docs to these notebooks on Binder (and avoid using the word Binder at first because complete beginner will not know what a binder link and possibly ignore such a link).

krassowski:
@betrabes are you interested in reworking this PR to integrate with existing resources, reduce reliance on external resources (such as stlouidsfed page or screenshots outside of this repository) and generalisation to other languages, or would you prefer to close it to signal that someone else can pick this by starting from scratch?

krassowski:
Closing due to lack of follow up. If anyone is interested, they are welcome to open revive this work in a new PR :)

Issue #7126, opened 2019-08-29
Run CI on mulitple Python versions


vidartf:
As exemplified by #7124, we could probably benefit from running the CI python tests on mulitple versions of Python. Ideally, the full minor span from oldest to newest (3.5, 3.6, 3.7 currently), but the minimum should probably be oldest and newest (3.5, 3.7) to pick up the largest feature span. If we do not want to add more entries to the build, we could e.g. do the oldest on Windows, and the newest on Linux.

jasongrout:
I support this.

saulshanabrook:
This sounds good! I would support running them all on linux. If we end up needing more concurrency we can always try to ping the pipelines team.

telamonian:
I'm a little concerned about adding more test suites to our already only "mostly reliable" CI. If the Python suite ever has any intermittent failures, doing it in triplicate will increase the failure rate for the CI overall. This is good if the intermittent problem is due to a bug that we can fix, but we also get a lot of intermittent CI failures for spurious reasons (eg flood protection during the documentation link check).

As an alternative, maybe we could just lint check for post 3.5 features in a pre-commit hook? I was planning on adding a hook to prettify Jlab's Python. It probably wouldn't be too hard to get some static code checks in there as well.

vidartf:
My impression is that it is mostly the JS tests that have spurious test timeouts? Others might have a more complete picture than me though. Here are Python test failures from last 30 days, and I think most of them are relevant failures: https://dev.azure.com/jupyterlab/jupyterlab/_test/analytics?definitionId=1&contextType=build

vidartf:
(filtering to master branch is most likely to show spurious failures, as most relevant ones are dealt with before merging).

PR #11646, opened 2021-12-09
Drop testing Python 3.6, test on Python 3.10

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/7126

## Code changes

- [x] `setup.cfg` and CI updates
- [x] ~Update migration guide~ -> this will appear in the changelog under "API and Breaking Changes"

## User-facing changes

~JupyterLab 4.0 will not be usable with Python 3.6.~

None

~Tagging `api-change` as this will be breaking for Python 3.6 users.~

None for now

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,16 +20,16 @@ classifiers =
     Intended Audience :: Science/Research
     License :: OSI Approved :: BSD License
     Programming Language :: Python
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
 
 [options]
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.6
+python_requires = >=3.7
 install_requires =
     ipython
     packaging
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,7 @@ classifiers =
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.7
+python_requires = >=3.6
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -20,16 +20,16 @@ classifiers =
...
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
 
 [options]
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.6
+python_requires = >=3.7
```
fcollonval, line 32: Do we really want to forbid installing JupyterLab on older system? Or just mentioned through the specifiers it is not supported.

Forbidding it may be too stringent on some servers that ships with older stable OS. More over there is no code used that requires Python >= 3.7

jtpio: Maybe we can also wait for `jupyter_server` (and / or other dependencies) to drop it first.

jtpio: Based on the discussions from yesterday's meeting (https://github.com/jupyterlab/team-compass/issues/128#issuecomment-995035053), I pushed https://github.com/jupyterlab/jupyterlab/pull/11646/commits/088a672075ae3cd837d86fa500c4e78db24e0106 to keep `python_requires = >=3.6` for now.

So we can move forward with this PR and do the real drop at a later stage.

fcollonval (approved):
Thanks

fcollonval:

jtpio:

jtpio:
Only backporting this one but not the drop of Python 3.6 on `3.2.x`. To hopefully fix the windows CI on `3.2.x` PRs.

PR #10755, opened 2021-08-01
Bump `url-parse~1.5.4`

krassowski:
## References

#10753. Jumping on this one just in case to make sure it gets updated to 1.5.3+ and I don't have to make downstream patch releases with workarounds for bugs in <1.5.3.

## Code changes

- [x] bump `url-parse": "~1.5.4"` for `coreutils` (1.5.3 was breaking tests due to https://github.com/unshiftio/url-parse/issues/219)
  - [ ] fix tests?
- [x] we are still stuck with url-parse 1.4.x because of `storybook` 5.2.8. We would probably need to move to 6.x. Any concerns about bumping the major version here? CC @jasongrout who was updating it last month for 3.1.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #10516, opened 2021-06-30
Refresh yarn.lock in preparation for 3.1

jasongrout:
## Code changes

This refreshes the yarn.lock file to the latest versions of requested dependencies. I created these changes by deleting the yarn.lock and then installing with `pip install -ve .`.

The idea here is that before doing a minor, we should upgrade to the latest versions of our dependencies to pick up bugfixes and stay current on our dependencies.

PR #11649, opened 2021-12-09
Remove Storybook

jtpio:
## References

In https://github.com/jupyterlab/jupyterlab/pull/11648 we noticed some dependencies like `nth-check` are difficult to update and come from the dependency on `storybook`.

Storybook seems to be coming back often when we try to update dependencies, for example:

- https://github.com/jupyterlab/jupyterlab/pull/10516
- https://github.com/jupyterlab/jupyterlab/pull/10755

I'm not sure if anyone is currently using it? Which is why this PR proposes we just remove it. If someone is still using, then we should consider keeping it.

## User-facing changes

None

Should be none.


jtpio:
cc @telamonian if you have any input on this

fcollonval:
Thanks for starting this.

We spoke briefly about that in https://github.com/jupyterlab/jupyterlab/pull/11173#issuecomment-935444879 @jasongrout  and I.

The agreement was indeed to drop it as no active developers use it (this is probably the reason it was never updated :wink:).

jtpio:
Marking as ready as I'm not planning to make any more changes.

We can also leave it open for a bit so folks have time to comment. This can also be discussed at the next weekly meeting.

jtpio:
Also the diff cleans up the `yarn.lock` quite a bit:

image

fcollonval (approved):
Thanks @jtpio

fcollonval:
Sorry I closed it a bit too quickly. I will raise the point at today's team meeting. And revert if needed.

PR #11650, opened 2021-12-09
Remove border from user menu when is empty

hbcarlos:
Follow-up for https://github.com/jupyterlab/jupyterlab/pull/11443#issuecomment-989664692

## References
#11443

## Code changes
Added a data attribute to the title when the menu is empty and remove the border from CSS.
I also removed the caret down from the widget when the menu is empty.

Screencast:

https://user-images.githubusercontent.com/26092748/145483961-f7743758-94ca-4aea-8151-f6ff9a6aa323.mp4

cc @krassowski

### DIFF
--- a/packages/user/src/menu.ts
+++ b/packages/user/src/menu.ts
@@ -120,7 +120,6 @@ export class UserMenu extends Menu {
     this.title.icon = caretDownIcon;
     this.title.iconClass = 'jp-UserMenu-caretDownIcon';
     this.title.dataset = { empty: 'true' };
-    //this.dataset.empty = 'true';
     this._user.ready.connect(this._updateLabel);
     this._user.changed.connect(this._updateLabel);
   }
@@ -152,7 +151,6 @@ export class UserMenu extends Menu {
    * The index will be clamped to the bounds of the items.
    */
   insertItem(index: number, options: Menu.IItemOptions): Menu.IItem {
-    //this.dataset.empty = 'false';
     this.title.dataset = { ...this.title.dataset, empty: 'false' };
     return super.insertItem(index, options);
   }
@@ -168,7 +166,6 @@ export class UserMenu extends Menu {
   removeItemAt(index: number): void {
     super.removeItemAt(index);
     if (this.items.length === 0) {
-      //this.dataset.empty = 'true';
       this.title.dataset = { ...this.title.dataset, empty: 'true' };
     }
   }
@@ -178,7 +175,6 @@ export class UserMenu extends Menu {
    */
   clearItems(): void {
     super.clearItems();
-    //this.dataset.empty = 'true';
     this.title.dataset = { ...this.title.dataset, empty: 'true' };
   }
 }
--- a/packages/user/src/menu.ts
+++ b/packages/user/src/menu.ts
@@ -35,20 +35,15 @@ export class RendererUserMenu extends MenuBar.Renderer {
     let dataset = this.createItemDataset(data);
     let aria = this.createItemARIA(data);
 
-    if ('empty' in dataset && dataset.empty === 'true') {
-      return h.li(
-        { className, dataset, tabindex: '0', onfocus: data.onfocus, ...aria },
-        this._createUserIcon(),
-        this.renderLabel(data)
-      );
-    } else {
-      return h.li(
-        { className, dataset, tabindex: '0', onfocus: data.onfocus, ...aria },
-        this._createUserIcon(),
-        this.renderLabel(data),
-        this.renderIcon(data)
-      );
+    const children = [this._createUserIcon(), this.renderLabel(data)];
+    if (dataset.empty !== 'true') {
+      children.push(this.renderIcon(data));
     }
+
+    return h.li(
+      { className, dataset, tabindex: '0', onfocus: data.onfocus, ...aria },
+      ...children
+    );
   }
 
   /**
--- a/packages/user/src/menu.ts
+++ b/packages/user/src/menu.ts
@@ -120,7 +120,7 @@ export class UserMenu extends Menu {
     this.title.icon = caretDownIcon;
     this.title.iconClass = 'jp-UserMenu-caretDownIcon';
     this.title.dataset = { empty: 'true' };
-    this.dataset.empty = 'true';
+    //this.dataset.empty = 'true';
     this._user.ready.connect(this._updateLabel);
     this._user.changed.connect(this._updateLabel);
   }
@@ -152,7 +152,7 @@ export class UserMenu extends Menu {
    * The index will be clamped to the bounds of the items.
    */
   insertItem(index: number, options: Menu.IItemOptions): Menu.IItem {
-    this.dataset.empty = 'false';
+    //this.dataset.empty = 'false';
     this.title.dataset = { ...this.title.dataset, empty: 'false' };
     return super.insertItem(index, options);
   }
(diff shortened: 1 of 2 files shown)


packages/user/src/menu.ts
```diff
@@ -34,12 +34,21 @@ export class RendererUserMenu extends MenuBar.Renderer {
...
+        this.renderLabel(data)
+      );
+    } else {
+      return h.li(
+        { className, dataset, tabindex: '0', onfocus: data.onfocus, ...aria },
+        this._createUserIcon(),
+        this.renderLabel(data),
+        this.renderIcon(data)
+      );
+    }
```
fcollonval, line 51: We could refactor it to enforce consistency:
```suggestion
    const children = [this._createUserIcon(), this.renderLabel(data)]
    if (dataset.empty !== 'true') {
      children.push(this.renderIcon(data));
    }
    
    return h.li(
      { className, dataset, tabindex: '0', onfocus: data.onfocus, ...aria },
      ...children
    );
```

> I remove the test `'empty' in dataset` as in that case `dataset.empty === undefined` -> we are covered

packages/user/src/menu.ts
```diff
@@ -115,6 +124,8 @@ export class UserMenu extends Menu {
     this.title.label = this._user.isReady ? name : '';
     this.title.icon = caretDownIcon;
     this.title.iconClass = 'jp-UserMenu-caretDownIcon';
+    this.title.dataset = { empty: 'true' };
+    this.dataset.empty = 'true';
```
fcollonval, line 128: It sounds to me that the source of truth is `this.title.dataset`. So could we remove modifying `this.dataset` in the `UserMenu`?
Changing the title dataset will trigger a change signal on the menu that will trigger an update on the menubar. So this should not be needed.

https://github.com/jupyterlab/lumino/blob/6da4c5899de06b9ed857b0c4355592e12f8c3476/packages/widgets/src/title.ts#L338-L344

```suggestion
```

packages/user/src/menu.ts
```diff
@@ -132,6 +143,49 @@ export class UserMenu extends Menu {
...
+   *
+   * @param options - The options for creating the menu item.
+   *
+   * @returns The menu item added to the menu.
+   *
+   * #### Notes
+   * The index will be clamped to the bounds of the items.
+   */
+  insertItem(index: number, options: Menu.IItemOptions): Menu.IItem {
+    this.dataset.empty = 'false';
```
fcollonval, line 160: See previous comment

```suggestion
```

packages/user/src/menu.ts
```diff
@@ -132,6 +143,49 @@ export class UserMenu extends Menu {
...
+   *
+   * @param index - The index of the item to remove.
+   *
+   * #### Notes
+   * This is a no-op if the index is out of range.
+   */
+  removeItemAt(index: number): void {
+    super.removeItemAt(index);
+    if (this.items.length === 0) {
+      this.dataset.empty = 'true';
```
fcollonval, line 176:
```suggestion
```

packages/user/src/menu.ts
```diff
@@ -132,6 +143,49 @@ export class UserMenu extends Menu {
...
+      this.title.dataset = { ...this.title.dataset, empty: 'true' };
+    }
+  }
+
+  /**
+   * Remove all menu items from the menu.
+   */
+  clearItems(): void {
+    super.clearItems();
+    this.dataset.empty = 'true';
```
fcollonval, line 186:
```suggestion
```

packages/user/style/menu.css
```diff
@@ -3,8 +3,23 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
+.lm-MenuBar-item[data-empty='true'],
+.p-MenuBar-item[data-empty='true'] {
```
fcollonval, line 7: Let's drop the old phosphor class as we know we are using recent version

```suggestion
.lm-MenuBar-item[data-empty='true'] {
```

packages/user/style/menu.css
```diff
@@ -3,8 +3,23 @@
...
+  border: none !important;
+}
+
+.lm-MenuBar-item[data-empty='true']:hover,
+.p-MenuBar-item[data-empty='true']:hover {
```
fcollonval, line 12:
```suggestion
.lm-MenuBar-item[data-empty='true']:hover {
```

packages/user/style/menu.css
```diff
@@ -3,8 +3,23 @@
...
+  background: none !important;
+}
+
+.lm-MenuBar-menu[data-empty='true'],
+.p-MenuBar-menu[data-empty='true'] {
```
fcollonval, line 17:
```suggestion
.lm-MenuBar-menu[data-empty='true'] {
```

fcollonval (changes requested):
Thanks @hbcarlos 

I left some suggestions to simplify the code.

packages/user/src/menu.ts
hbcarlos: Hey @fcollonval, thanks for reviewing this PR!

We need to add the attribute to both datasets because the menu and the title are two different elements on the DOM.
The title is added to the menu bar while the menu is added to `body`, and both have a small border to delimiter the menu. The title shows the border and changes the background in the menu bar to differentiate it from other menus. In contrast, the menu adds a border to the menu element, which is positioned `absolute` or `relative` (I do not remember correctly) below the top bar in the same position as the title.

fcollonval: I tested it locally and I go it now. Without it you get a pixel glitches from the empty menu:

image

So I would still not do things like this. I would set pointer events to `none` on the menu bar item if it is empty. This will prevent the menu appearance.

> I update the suggestion accordingly

packages/user/style/menu.css
```diff
@@ -3,8 +3,20 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
+.lm-MenuBar-item[data-empty='true'] {
+  border: none !important;
```
fcollonval, line 7:
```suggestion
  border: none !important;
  /* Ensure the menu won't be displayed if there are no items */
  pointer-events: none;
```

packages/user/style/menu.css
```diff
@@ -3,8 +3,20 @@
...
+}
+
+.lm-MenuBar-item[data-empty='true']:hover {
+  background: none !important;
+}
+
+.lm-MenuBar-menu[data-empty='true'] {
+  border: none !important;
+}
+
```
fcollonval, line 17: Unneeded

```suggestion
```

hbcarlos:
I can not reproduce locally the UI test that's failing.

jtpio:
> I can not reproduce locally the UI test that's failing.

This should now have been fixed in https://github.com/jupyterlab/jupyterlab/pull/11758

jtpio:
Thanks @hbcarlos for working on this.

The related PR in Lumino has been merged: https://github.com/jupyterlab/lumino/pull/277

Does this current PR in JupyterLab need a new Lumino release to move forward?

hbcarlos:
Yes, I'm waiting for a release in lumino to finish the PR.

jtpio:
Looks like a new version of Lumino is out: https://github.com/jupyterlab/lumino/releases/tag/v2022.1.10

hbcarlos:
Yep, I'm on it! Thanks, @jtpio for the reminder!

hbcarlos:
Closing since this was fixed in #11823

PR #11654, opened 2021-12-10
Move the top area spacer to a different plugin

### LINKED ISSUE #11589: Improve the Top Area as an extension point
### Problem

In https://github.com/jupyterlab/jupyterlab/pull/11443 we add a spacer to the top area to align items to the top right corner of the page.

Some extensions already add a spacer to the top bar to align the items to the right. Which means there could be multiple spacers if the user menu and the other extensions are activated on the page.

It might be fine for now, but maybe we should check how to make the top area more well defined so extensions can more easily add items there without stepping on each others toes.

### Proposed Solution

We could just add that spacer independently of these plugins, by default in JupyterLab, and document using a rank < 900 to place items to the left, and rank > 900 to place them to the right.

Another approach would be to make the space to the right of the menu in the top bar a `Toolbar` like extension point. Which is what [jupyterlab-topbar](https://github.com/jupyterlab-contrib/jupyterlab-topbar) started experimented with.

image

### Additional context

Discussed in https://github.com/jupyterlab/jupyterlab/pull/11443

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/11443 tracked in https://github.com/jupyterlab/jupyterlab/issues/11589.

This PR moves the spacer widget in the top area to a separate plugin in `@jupyterlab/application-extension`.

## Code changes

Move the `#jp-topbar-spacer` to a different plugin.

This is the simplest "fix" for #11589. With this, we should also document how to position widgets to the left (rank < 900) or to the right (rank > 900) of the spacer.

However we could also consider doing some more elaborate, such as reusing the `Toolbar` API so it's easier to position elements. Similar to the notebook toolbar. This could then also be made configurable via the settings.
Not sure yet whether we want to do this to fully fix #11589. If so this could be done as a separate PR.

## User-facing changes

None

None

### DIFF
--- a/docs/source/extension/extension_points.rst
+++ b/docs/source/extension/extension_points.rst
@@ -424,7 +424,7 @@ declaring default keyboard shortcuts for a command:
 
 Shortcuts added to the settings system will be editable by users.
 
-From Jupyterlab version 3.1 onwards, it is possible to execute multiple commands with a single shortcut. 
+From Jupyterlab version 3.1 onwards, it is possible to execute multiple commands with a single shortcut.
 This requires you to define a keyboard shortcut for ``apputils:run-all-enabled`` command:
 
 .. code:: json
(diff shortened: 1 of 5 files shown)


jtpio:
> However we could also consider doing some more elaborate, such as reusing the `Toolbar` API so it's easier to position elements. Similar to the notebook toolbar. This could then also be made configurable via the settings.

One proposal could be to use parts (probably without the drag and drop) of the `jupyterlab-topbar` extension https://github.com/jupyterlab-contrib/jupyterlab-topbar.

fcollonval (approved):
Thanks @jtpio

fcollonval:
I switch the label to enhancement because most extension dev are not paying attention to the maintenance section.

Issue #10911, opened 2021-08-25
Visual indicator that a file filter is in effect


vidartf:
### Problem

Currently, users might have a file filter active in the file browser, and not realize it (particularly since it is retained even after browsing into a different folder, which is probably unexpected). They will then be annoyed when they cannot find their files, and instead only see empty folders (with a few files if they happen to match the filter).

### Proposed Solution

A number of proposals might be considered independent of each other:
1. Add some clear visual queue that a filter is currently in effect. Either by a border, overlay, or icon (doing some research on what other applications use would be good).
2. Clear the file filter when you browse to a different folder. Very rarely does the filter you applied in one folder translate well to another, and very rarely is that what you would want.
3. Adding a clear button to the input ("X" button) to quickly clear the filter. Most filters are temporary in nature, e.g. used to help find a file in a folder with a lot of entries. Once the file has been found, in most cases the filter is no longer necessary.

Note that proposals 1 and 3 might be combined, similarly to what vscode does:

With a filter of "rc" (on both folders and files in vscode's case), a overlay is shown, and matches are highlighted:
![image](https://user-images.githubusercontent.com/510760/130767832-660cfb95-9a69-4c42-a95d-c21d04aee047.png)

On mouse hover on overlay, a clear button is shown (in addition to other filter options):
![image](https://user-images.githubusercontent.com/510760/130767855-c74fa5aa-21af-438d-a19f-7a2d112290a1.png)

krassowski:
Also, we already have a quick find option (just start typing in the listing), and I believe that the filter field is redundant with respect to that. I think that the quick find on listing is the thing that should receive an indicator (to show what prefix is being used for quick find). I will follow up after hours.

vidartf:
@krassowski Yes, I do not see that the quick-find functionality was mentioned in the [original issue](https://github.com/jupyterlab/jupyterlab/issues/8157) or [PR](https://github.com/jupyterlab/jupyterlab/pull/8615).

treyhunner:
I teach many students who use JupyterLab as their Python environment. These are first-time Python and Jupyter users. The biggest issue we tend to run into is an accidental filter on the file browser. This typically happens when students type a search in to find the folder they're looking for, double click the folder name, and then see that the folder appears to be unexpectedly empty. The folder isn't *actually* empty, but the filter does not clear when navigating into a subdirectory in the file browser so what the (now very narrow) filter makes the folder appear empty.

[...]

Issue #2735, opened 2017-07-24
Blinking cursor is masking code in high contrast mode


huguesv:
In Windows settings, select High Contrast #1
![image](https://user-images.githubusercontent.com/1696845/28536830-b9c2d696-705d-11e7-9bca-b1f4770e4c9f.png)

Open a notebook using Microsoft Edge (Chrome doesn't respect high contrast)

Place the cursor in the middle of some code, notice that the character that follows the cursor disappears every time the cursor blinks.

See the 'persisted' word:
![image](https://user-images.githubusercontent.com/1696845/28536912-fce1a95c-705d-11e7-8ba0-5827e527ab72.png)

sccolbert:
The right solution for this is to probably have a "high contrast" jlab theme, similar to how we have light and dark themes. That way it works across all browsers.

Your specific issue may be related to code mirror, and may not be something we can easily fix.

huguesv:
Chrome (for reference):
![image](https://user-images.githubusercontent.com/1696845/28537527-11f611b4-7060-11e7-8b41-1d4642dff9e6.png)

Firefox (no repro for this bug, but notice the toolbar is wrong like in Edge):
![image](https://user-images.githubusercontent.com/1696845/28537546-20db7ab6-7060-11e7-90b5-738faa1a0122.png)

PR #11658, opened 2021-12-10
Increase color contrast in input boxes

isabela-pf:
## References

Related to #8832 (but will not close it). Specifically, this PR makes changes to the file browser’s filter UI so that it meets [minimum non-text contrast accessibility standards](https://www.w3.org/TR/WCAG21/#non-text-contrast), makes the search button more clear, and is consistent with the command palette’s similar UI. This was discussed further in [this comment on #8832](https://github.com/jupyterlab/jupyterlab/issues/8832#issuecomment-758898894).

## Code changes

As a list, this PR
- Darkens the border color for input boxes in the light theme, and lightens the border color in dark theme. This meant changing `--jp-input-border-color` to Material Design’s Grey 600, the smallest change in value that also meets contrast standards for a UI element. All other borders (ones that aren’t interactive, like menu dividers) have stayed the same.
- Changes the styling of the file browser’s filter to match the command palette’s blue button design.
- Changes the New Launcher button in the file browser to the main blue used elsewhere (the fact that they were barely not the same color is more visible in the dark theme). This means it is now `--jp-brand-color1`.

## User-facing changes

Interactive borders (like in search fields) are slightly higher contrast, the file browser search matches the appearance of the command palette, and the New Launcher button will always match the main accent blue used. 

Before:
JupyterLab light theme file browser and command palette with lower contrast.
JupyterLab dark theme file browser and command palette with lower contrast.

After:
JupyterLab light theme file browser and command palette with increased search and button contrast.
JupyterLab dark theme file browser and command palette with increased search and button contrast.

The changes are pretty minimal from a user perspective since they are meant to give a boost of contrast and consistency rather than overhaul the look.

None of these changes are backwards-incompatible.

---

Questions I’d like feedback on (please!):
- I noticed that the New Launcher button has a white plus icon, so no matter the theme its icon remains white. Similar UI, like the command palette’s search button, uses the layout color variables and switch per theme. I’d prefer it all of these match, but which method do you all prefer? (Usually I’d recommend using the variables.)
- `--jp-input-border-color` looks like it’s used in the command palette (shown here), file browser (shown here), notebook styling, and code editor styling. I wasn’t able to test and take a screen shot of how this change impacts the notebook and code editor, so could someone either post a screen shot or give me advice on how to test it myself?

Thanks in advance for the review! 🌻 

P.S. I expect the UI tests to fail at first, but I only know how to fix them once they fail here so I'm going to have to let that happen.


krassowski:
Thank you for starting this, great catch with the plus button using the old colour.

I think the problem with changes like this is that strong colors can be very distracting (and I already feel affected by the previous changes to the colour scheme). This is discussed on https://www.w3.org/WAI/GL/WCAG3/2021/how-tos/visual-contrast-of-text/:

> Some users find too much contrast in large/bold elements may be distracting or overwhelming due to glare or neurological issues.

I am a big supporter of having a good contrast, but there is a point at which I believe we should focus more on addressing the issues reported by users using high contrast mode, like: https://github.com/jupyterlab/jupyterlab/issues/2735. I do not know if this particular issue was brought up by any user, but if it was not then maybe we should reconsider what are the implications of going thought with this change?

In this particular PR I would like to reconsider the search box "button" and border in the file browser:
- we should follow the visual hierarchy principle. The loop icon/"button" is symbolic - it does not have a function attached (clicking on it does not do anything) therefore it should not be prominent in the interface. Even if it had a function it should only be prominent when the search input (in the file browser) is focused, not at all times (because it is distracting).
- when on it we should probably consider how it would fit with the request to include a clear button in the search field: https://github.com/jupyterlab/jupyterlab/issues/342
- the search field in the file browser is redundant and it appears that it was introduced without a proper consideration for the UX and ignored a pre-existing quick-search implementation (which is why it is redundant), see https://github.com/jupyterlab/jupyterlab/issues/10911#issuecomment-905502582; moreover it is annoying due to its persistence and lack of filtering for directories. As you can see from the previous sentence and the linked issue I am not a big fan of that search bar, especially because as a user I cannot hide it (I would prefer to make it hidden and only show up as a small pop-up mirror the text typed into the quick-search functionality, as intended in the original issue). Making this field more prominent (both the button and outline) exacerbates the issue for me. I mean the stronger border proposed in this PR would be fine for me in the command palette if we agree that it will target 4.0 and also rework the search bar from the file browser. I am happy to write the code for this one.

isabela-pf:
Thanks for always being so on-top of reviewing PRs, @krassowski. It’s good to get your feedback on  always (and doubly so on accessibility efforts). 

I think your critiques make sense. Let me address them point by point so I don’t miss anything.

- I agree that higher contrast colors have consequences and in some cases might be less accessible when it isn’t serving low vision needs. The problem is that I want the default JupyterLab themes to pass any WCAG. And while compliance isn’t everything, I do think it is a standard worth respecting since it is the most widely accepted and researched. Far more so than I could do myself or even what I think we could do as a team within the Jupyter community. Maybe there’s a middle ground here, though it may make sense to pull that discussion back to #8832 where these changes were last discussed. 
- As for high contrast mode, my lack of focus there for the moment has been because I believe having a separate theme to maintain isn’t the best approach since it can easily push accessibility efforts off to the side instead of keeping it a part of the default application state. Still, it’s probably a good long term goal to have either more themes or an easier way to make themes (like from the soon-to-come settings interface). 
- I’m open to removing the blue that makes the icon look like a button. That was my misunderstanding of how the button works and a really good point (I honestly thought maybe I was doing something wrong when it wasn't clicking :laughing:). I’ll have review the issue about the filter UX; removing it would indeed remove the need for these accessibility concerns, though I’m sure that would also require a little more notice (and skills beyond my own).
- When I checked again, I don’t think the command palette is a true button either. Should blue button be removed from both then? (This is probably a bigger question; I may bring it up in this week’s team meeting.)
- A clear button sounds fine to add to the filter to me. I don’t know if this was included in your volunteering to write code for it, but I would definitely need to look for help with this if not.

In general, you’re right to say that I haven’t been browsing other open issues to group into accessibility efforts as we re-evaluate some of the areas in the interface. I’ve been keeping a more narrow scope (basically on anything labeled accessibility/what shows up in the [accessibility GitHub project](https://github.com/orgs/jupyterlab/projects/1)) both because of time and my own limited skills. Thanks for pointing out these relevant issues.

For this PR, would you say it’d make sense to
- [ ] revert blue button changes to the file browser filter and redesign in another PR based on existing feedback in both issues (and browsing for others). (And maybe make sure the command palette matches, where relevant?)
- [ ] keep the corrected new launcher button color (what do we think about the button icon being `white` and not `--jp-layout-color1` like other interface elements?)
- [ ] keep the updated `--jp-input-border-color` for now since it is okay with the command palette (which we plan to keep)
?

Let me know if I’ve misinterpreted anything or missed something. Thanks again.

krassowski:
Yes, that all makes sense. I consider the _clear_ button a good first issue so I did not plan on working on this in the near future. The filtering improvements are of course out of scope for this PR;  I hope I will be able to help with this one.

I really hope my comment was not too overwhelming. I mostly wanted to bring in the conversation about the big picture and how different enhancements will interact and fit together.

isabela-pf:
I shared this in this week's accessibility meeting too ([notes for this meeting are in this jupyter/accessibility PR]()! I wanted to add some of the things we discussed there:
- Defining `--jp-input-border-color` as `--md-grey-600` may not be best practice to match elsewhere. Maybe add another border color (proposed `--jp-inverse-border-color` class since all the other borders are light and this needs to be darker) to that larger group then use that variable to define `--jp-input-border-color`.
- Long term, having more themes to support a range of needs and making themes more easily customizable (without code) will be key. That's outside the scope of this PR, but worth noting since we discussed that one theme will not be able to perfectly serve all needs.
- Having non-buttons styled to look like buttons seems like bad practice (in the file browser and command palette). It may need to be it's own issue to discuss changes to this that aren't just color, though.

isabela-pf:
I think I've made all the [changes I listed in the above comment](https://github.com/jupyterlab/jupyterlab/pull/11658#issuecomment-994140372). I still need to check for or write issues for some of the larger issues like not-actually-a-button styling in the command palette.

Let me know if I've missed anything or there's any other feedback!

krassowski:
Thanks @isabela-pf! It looks like the snapshots need updating.

krassowski (approved):
I believe the CI failures are not related, should go away after merging master in/rebasing.

isabela-pf:
Okay, so I thought I rebased but this looks completely wrong. Is there a clear fix for this to someone else, or is it easier for review if I make a fresh PR?

jess-x:
hey @isabela-pf, rebase's annoying, been there. Let me know if I can help

fcollonval:
@isabela-pf I try to fix your rebase. Could you please check the modifications in the current state are the one expected?

isabela-pf:
@fcollonval thank you thank you thank you so much! It looks correct to me.

fcollonval:
Let's backport it to 3.3.x as it is a great enhancement

fcollonval:

Issue #11663, opened 2021-12-11
TOC extension should have mandatory notebook tracker


jweill-aws:
## Description

This is a partial revert of changes made in #11445. The Table of Contents (TOC) extension's `INotebookTracker` parameter should be mandatory, not optional, as without the tracker the TOC is not useful. RetroLab continues not to provide `INotebookTracker` in its service map, but the proper recourse should be to change RetroLab's service map, not to make the parameter optional in the TOC extension.

## Reproduce

Attempt to use TOC extension in RetroLab

## Expected behavior

Extension appears, but is non-functional.

krassowski:
But does it work for Python files? I'm not saying the proposed solution is wrong, just trying to highlight that ToC also supports use cases other than notebook.

jweill-aws:
Prior to #11445, the `INotebookTracker` parameter was mandatory, as were `IEditorTracker`, `IMarkdownViewerTracker`, and `ILayoutRestorer`. Of these three, I believe only `ILayoutRestorer` should genuinely remain optional, as RetroLab does not yet support restorable layouts. RetroLab supports editing Python files and Markdown files, so perhaps the other trackers should be reinstated as required as well.

https://github.com/jweill-aws/jupyterlab/commit/3c1aca7403f8276d5413a3761674d5f77b06e821#diff-9e7264f868473a01274aaa24956ff8182819b297555fe6d533b403601b4ddb10

PR #11664, opened 2021-12-11
Make INotebookTracker parameter mandatory in TOC extension, as it had been

jweill-aws:
## References

Affects issue #11663. A partial rollback of changes made in #11445.

## Code changes

Reinstates `INotebookTracker` as a mandatory parameter in the TOC extension, as without a notebook tracker, the TOC extension displays nothing.

## User-facing changes

Required to make TOC extension functional in RetroLab. Should have no effect on JupyterLab.

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -62,11 +62,11 @@ async function activateTOC(
   docmanager: IDocumentManager,
   rendermime: IRenderMimeRegistry,
   translator: ITranslator,
+  notebookTracker: INotebookTracker,
   editorTracker?: IEditorTracker,
   restorer?: ILayoutRestorer,
   labShell?: ILabShell,
   markdownViewerTracker?: IMarkdownViewerTracker,
-  notebookTracker?: INotebookTracker,
   settingRegistry?: ISettingRegistry
 ): Promise<ITableOfContentsRegistry> {
   const trans = translator.load('jupyterlab');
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @jweill-aws for looking into this.

Unfortunately this won't fix the Table of Contents being empty. As mentioned in https://github.com/jupyterlab/retrolab/pull/275#issuecomment-989600616, RetroLab does include the plugin that provides `INotebookTracker` on the notebook page. So even if `INotebookTracker` is marked as `optional` in the plugin definition of the JupyterLab package, it will still be provided and available in RetroLab. So this won't change anything.

Instead we need to rework the TOC extension so it's more modular, and decouple the `TableOfContents` widget from `ITableOfContentsRegistry`. https://github.com/jupyterlab/jupyterlab/pull/11530 provides more context about this, but still need to be finished. Also this will probably have to be a 4.0 change.

---

As a general guideline, it's better to not make services `requires` when they don't have to. Instead making them `optional` allow for greater plugin re-usability and this is what made it possible to build RetroLab in the first place. If we make them `requires` like this change suggests, it will be more difficult for other lab-based applications to reuse the plugins as is. So https://github.com/jupyterlab/jupyterlab/pull/11445 is still a good change and a general pattern to follow.

As mentioned elsewhere, we can imagine a simple lab-based app with a code editor on the right and the table of contents on the left side of the page. This app would not provide the `INotebookTracker` but will still want to use the functionalities of the TOC extensions depending on `IEditorTracker` for example. If `INotebookTracker` is required, the TOC plugin will fail to activate and it will not be possible to reuse any functionality of the plugin.

jtpio:
For more information about `requires` and `optional`: https://jupyterlab.readthedocs.io/en/latest/extension/extension_dev.html#tokens

Issue #11555, opened 2021-11-26
Show the currently focused cell by other collaborators


martinRenou:
Similar to https://github.com/jupyterlab/jupyterlab/issues/11528 

### Problem

We currently have a way to see the cursor of other collaborators. But this only works when the other collaborators are in "edit" mode. This results in this issue https://github.com/jupyterlab/jupyterlab/issues/11406

We should maybe think of a way to indicate that another collaborator has focus on one cell, whether they are in "edit" mode or "command" mode. Similar to what google does for showing the focused cells in spreadsheets:
![spreadsheet](https://user-images.githubusercontent.com/21197331/143556116-264748ec-aaad-4434-83b6-cddbc75d3c36.png)

### Proposed Solution

Show an outline colored with the collaborator's color on the focused cell.

### Additional context

This would fix https://github.com/jupyterlab/jupyterlab/issues/11406

hbcarlos:
Maybe we can show a list of collaborators in each cell to see where is everyone.
Something like this, but with smaller icons.

![rendered-markdown](https://user-images.githubusercontent.com/26092748/143570264-17c084ed-35d2-4139-8b62-320d277cdd53.png)

martinRenou:
It's similar to what I wanted to suggest in https://github.com/jupyterlab/jupyterlab/issues/11528#issuecomment-976651851. Though if we already have the list of contributors somewhere else in the UI it might be too much redundant information.

Another idea could be to just have a similar visual indication to the "active" state:
![collab](https://user-images.githubusercontent.com/21197331/143571505-983e643c-1a24-4eff-817f-893a974b3eee.png)

krassowski:
We could even show the cursor. It is technically feasible - while figuring out where the cursor should be after rendering is difficult, the trick is to insert the cursor before rendering so that it is reserved by marked.js.

We do this already for rendering Tex (we use custom markers to substitute stuff after rendering).

I think that rendering the cursor would be the best UX possible.

Out of the two proposals above I slightly prefer the icons over colours (you know colourblind people etc).

martinRenou:
> I think that rendering the cursor would be the best UX possible. (of course this is for edit mode, for command mode the icons or colors are fine).

We're doing this already in edit mode. Or did I misunderstand what you mean by cursor?

krassowski:
I mean when person A is in edit mode but person B has markdown rendered, we are currently not showing the cursor of A in the rendered markdown of person B.

[...]

PR #11676, opened 2021-12-13
Cell collaborators list

martinRenou:
## References

https://github.com/jupyterlab/jupyter_collaboration/issues/109

## Code changes

- Add a new CollaboratorsList react component.
- Synchronise current users between collaborators

## User-facing changes

There are two different options I explored here. 

---

### Over the cell

The first one is showing the list in the top-right of the cell:

collabicons

---

### In the cell header

The second one (current state of the PR) is showing the list in the cell header:

collab2

None


martinRenou:
I didn't explore the right-gutter solution discussed in https://github.com/jupyterlab/jupyterlab/issues/11555#issuecomment-989687674, as it will not work in the case of rendered Markdown cells, which would be unfortunate.

I have a personal preference for the first option (over the cell), it takes less space, and is easier to read.
It might hide the code content, but we could technically fix this by hiding the collaborator list if the user mouse goes over it (when the user wants to see/select code).

hbcarlos:
> I have a personal preference for the first option (over the cell), it takes less space, and is easier to read.

I prefer the first option too. The downside is that we could cover code/text, but I think this will happen in a few cases since, for example, markdown cells tend to start with a title (it is rare that the title takes the whole line).

martinRenou:
As discussed with @SylvainCorlay we might be able to improve the cell-header proposition by changing the style of the header itself (slightly different background color, solid border etc).

martinRenou:
It might also be possible to show the collaborator list as a column at the right of the cell, though the amount of space for small cells is limited

jtpio:
Looks like this PR could be closed and ported over to https://github.com/jupyterlab/jupyter_collaboration?

Issue #11611, opened 2021-12-03
Encourage extension authors to use `optional` dependencies


jtpio:
### Problem

In the current extension tutorial, we show how to add a command to the command palette by requiring a `ICommandPalette`:

```typescript
const extension: JupyterFrontEndPlugin<void> = {
  id: 'jupyterlab_apod',
  autoStart: true,
  requires: [ICommandPalette],
  activate: activate
};
```

This is however not the way `ICommandPalette` is usually consumed across the `jupyterlab` code base and other third-party extensions, since the command palette might or might not be available.

### Proposed Solution

We should instead encourage extension authors to use `optional` whenever possible. And explain how this helps make an extension usable in many different lab-based applications.

### Additional context

Extension tutorial: https://jupyterlab.readthedocs.io/en/latest/extension/extension_tutorial.html#show-an-empty-panel

PR #11677, opened 2021-12-13
Make `ILayoutRestorer` optional in the extension tutorial

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11611.

The motivation here is to make extension authors more aware of the `optional` field of the plugin definition, so they can build extensions that can be used in more JupyterLab based applications by default.

## Code changes

Documentation changes.

- Make `ILayoutRestorer` an `optional` to cover `optional` in the extension tutorial
- Keep `ICommandPaletter` as `requires` to also cover tokens specified as `requires`.

## User-facing changes

None for end users.

Documentation changes can be previewed on RTD: https://jupyterlab--11677.org.readthedocs.build/en/11677/extension/extension_tutorial.html

None

### DIFF
--- a/docs/source/extension/extension_tutorial.rst
+++ b/docs/source/extension/extension_tutorial.rst
@@ -820,7 +820,7 @@ Finally, rewrite the ``activate`` function so that it:
 
 1. Declares a widget variable, but does not create an instance
    immediately.
-2. Adds the global ``LayoutRestorer`` as the thid parameter of the ``activate`` function.
+2. Adds the global ``LayoutRestorer`` as the third parameter of the ``activate`` function.
    This parameter is declared as ``ILayoutRestorer | null`` since the token is specified as ``optional``.
 3. Constructs a ``WidgetTracker`` and tells the ``ILayoutRestorer``
    to use it to save/restore panel state.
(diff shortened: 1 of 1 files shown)


jtpio:
A follow-up to this would be to update the apod repo to take these changes into account: https://github.com/jupyterlab/jupyterlab_apod

docs/source/extension/extension_tutorial.rst
```diff
@@ -792,30 +792,43 @@ entire list of import statements looks like the following:
...
+
+    You can learn more about ``requires`` and ``optional`` in the :ref:`tokens` section
+    of the Extension Developer Guide.
+
 Finally, rewrite the ``activate`` function so that it:
 
 1. Declares a widget variable, but does not create an instance
    immediately.
-2. Constructs a ``WidgetTracker`` and tells the ``ILayoutRestorer``
+2. Adds the global ``LayoutRestorer`` as the thid parameter of the ``activate`` function.
```
fcollonval, line 823:
```suggestion
2. Adds the global ``LayoutRestorer`` as the third parameter of the ``activate`` function.
```

fcollonval (commented):
Thanks @jtpio 

I caught a typo. Otherwise LGTM

docs/source/extension/extension_tutorial.rst
jtpio: Good catch, thanks!

fcollonval (approved):
LGTM

fcollonval:
I think we should wait to solve jupyterlab/jupyterlab_apod#10 before backporting it to 3.2.x

PR #11680, opened 2021-12-13
Testing effect of DEFAULT_SETTINGS: this should make tests fail on gi…

schmidi314:
## References

Scratch PR for testing `galata.DEFAULT_SETTINGS` behaviour on github UI tests

(See #11615)

### DIFF
--- a/galata/src/galata.ts
+++ b/galata/src/galata.ts
@@ -29,9 +29,14 @@ export namespace galata {
       editorConfig: { cursorBlinkRate: 0 }
     },
     '@jupyterlab/notebook-extension:tracker': {
-      codeCellConfig: { cursorBlinkRate: 0 },
-      markdownCellConfig: { cursorBlinkRate: 0 },
-      rawCellConfig: { cursorBlinkRate: 0 }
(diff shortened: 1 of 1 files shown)


galata/src/galata.ts
```diff
@@ -29,9 +29,14 @@ export namespace galata {
       editorConfig: { cursorBlinkRate: 0 }
     },
     '@jupyterlab/notebook-extension:tracker': {
-      codeCellConfig: { cursorBlinkRate: 0 },
-      markdownCellConfig: { cursorBlinkRate: 0 },
-      rawCellConfig: { cursorBlinkRate: 0 }
+      //codeCellConfig: { cursorBlinkRate: 0 },
+      //markdownCellConfig: { cursorBlinkRate: 0 },
+      //rawCellConfig: { cursorBlinkRate: 0 },  // this should make some tests fail (checking if this makes a difference)
```
fcollonval, line 34: FYI The `cursorBlinkRate` impact is hard to catch as this makes the tests flaky. Every tests could still be passing if you are lucky enough that all snapshots got generated when the cursor was visible (we had that before).

PR #11682, opened 2021-12-13
Enforcing fonts for galata

schmidi314:
Trying to make fonts consistent over different Linux systems, including the github test.

Testing the github galata tests...
--> This commit should not affect the github workflow.

## References

See discussion in #11615

## Code changes

* Only in galata tests

## User-facing changes

None

None

### DIFF
--- a/galata/src/galata.ts
+++ b/galata/src/galata.ts
@@ -32,6 +32,9 @@ export namespace galata {
       codeCellConfig: { cursorBlinkRate: 0 },
       markdownCellConfig: { cursorBlinkRate: 0 },
       rawCellConfig: { cursorBlinkRate: 0 }
+    },
+    '@jupyterlab/apputils-extension:themes': {
+      overrides: { 'content-font-family': 'DejaVu Sans' }
     }
   };
 


fcollonval (approved):
LGTM

schmidi314:
> LGTM

Well, unfortunately most of my tests still fail. My feeling is that we should do this when there is at least one other person that has problems with local UI tests. Otherwise this PR would be only about fixing my local dev environment issues.

krassowski:
> Well, unfortunately most of my tests still fail. My feeling is that we should do this when there is at least one other person that has problems with local UI tests. Otherwise this PR would be only about fixing my local dev environment issues.

I do. My super inefficient workflow was to open a PR on my own fork to get the CI to kick in, download artifacts and then cleanly push the correct screenshot in the PR to this repo. Having majority of the tests pass locally would be super useful for good developer experience; the final 20% may not be worth the effort, but if we can get 80% to pass that would be a big step forward.

What things still cause problems on your setup?

schmidi314:
@krassowski that sounds familiar to me. Only that I have polluted the tests on the main repo (sorry) ... so let's get on with this.

1) the commit a7f5c5d should fix the collapsible-headings test. @krassowski can you confirm this locally?
2) the commit d306dbc shows something interesting. It definitely improves the contextmenu test. However:
antialias
Is this antialiasing?

fcollonval:
I guess I'm lucky. Running on stock Debian 11, running the 158 tests on 4 workers:

```
  12 failed
    [jupyterlab] › test/jupyterlab/general.test.ts:8:3 › General Tests › Launch Screen =============
    [jupyterlab] › test/jupyterlab/general.test.ts:13:3 › General Tests › Enter Simple Mode ========
    [jupyterlab] › test/jupyterlab/general.test.ts:28:3 › General Tests › Toggle Dark theme ========
    [jupyterlab] › test/jupyterlab/notebook-toolbar.test.ts:78:3 › Notebook Toolbar › Run cell =====
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:60:3 › Sidebars › Capture File Browser on right 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:21:5 › Terminal › Theme › Light theme terminal inherit 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:33:5 › Terminal › Theme › Light theme terminal light 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:43:5 › Terminal › Theme › Light theme terminal dark 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:53:5 › Terminal › Theme › Dark theme terminal inherit 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:66:5 › Terminal › Theme › Dark theme terminal light 
    [jupyterlab] › test/jupyterlab/terminal.test.ts:77:5 › Terminal › Theme › Dark theme terminal dark 
    [jupyterlab] › test/jupyterlab/toc.test.ts:217:3 › Table of Contents › Open context menu =======
  2 flaky
    [jupyterlab] › test/jupyterlab/debugger.test.ts:33:3 › Debugger Tests › Start debug session ====
    [jupyterlab] › test/jupyterlab/notebook-create.test.ts:41:3 › Notebook Create › Create a Code cell 
  144 passed (9m)
```

Terminal related tests are expected to fail. The general tests and the sidebar one are failing because I'm using `nb_conda_kernels` and so all my environments show up in the launcher (we could mock kernelspec endpoint to improve that one).

Regarding the font for menu, here are the details:

image

Regarding the discrepancy on the font, I don't know if it could come from anti-aliasing. But I'm a bit afraid of using higher threshold (the [default seems to be 0.2](https://github.com/microsoft/playwright/blob/5ba7903ba098586a13745e0d7ac894f1d55d47aa/packages/playwright-test/src/matchers/golden.ts#L69)). The small discrepancy like the difference between triangle icon showing collapsed/expanded may not be catch. But it is a risk we need to evaluate against dev' comfort.

fcollonval:
@schmidi314 in your screencast with font diff, was the test ran twice? If so was the diff identical in the two attempts?

schmidi314:
> in your screencast with font diff, was the test ran twice? If so was the diff identical in the two attempts?

Just re-ran it twice manually. Yes they are the same.

schmidi314:
> But I'm a bit afraid of using higher threshold (the [default seems to be 0.2](https://github.com/microsoft/playwright/blob/5ba7903ba098586a13745e0d7ac894f1d55d47aa/packages/playwright-test/src/matchers/golden.ts#L69)).

Yes, I know what you mean. But with many of my tests failing with minor font issues (the italics stuff I mean), I'm gaining more confidence with the 0.3 threshold.

schmidi314:
@fcollonval can you please check `fc-list | grep 'Menlo\|Consolas'`

fcollonval:
> fc-list | grep 'Menlo\|Consolas'

This returns no match

fcollonval:
@krassowski @jtpio would you mind running this PR locally and report the test results? If possible could you test it a second time not specifying the snapshot threshold in galata/src/playwright-config.ts?

schmidi314:
Another observation:

completer

Shouldn't the result be like "actual", rather than "expected"?

jtpio:
I was going through issues added to the `3.2.x` milestone and found this one.

Is it still relevant, and if so should it be bumped to `3.3.x` or `4.0`? Thanks!

fcollonval:
Moving to 4.0, it is a maintenance PR for dev comfort (but it will be a good one to get...).

> Instead of fighting with the OS fonts, another path could be to override the theme font (removing the system-ui one) for the UI tests.

schmidi314:
How shall we proceed here? For me this PR makes testing a little easier. But I haven't managed to get the italics issue under control. So it would be interesting how it works for the other devs to choose which commits to keep.

krassowski:
Apologies that no one responded here @schmidi314. I would suggest to keep the part changing fonts but not the part relaxing threshold and merge this after resolving conflicts. If it helps even a bit then it's good to go - does not need to be perfect.

krassowski:
@fcollonval would your docker image PR supersede this PR?

fcollonval:
> @fcollonval would your docker image PR supersede this PR?

Yes it will

JasonWeill:
@fcollonval If #15916 supersedes this, and you're planning to merge this in during the next month, would you prefer to close this PR and tag #15916 with the 4.3.0 milestone?

krassowski:
Closing in favour of work-in-progress for fixing all fonts in:
- https://github.com/jupyterlab/jupyterlab/pull/18535
- https://github.com/jupyterlab/jupyterlab/pull/18542
- https://github.com/jupyterlab/jupyterlab/pull/18541
- https://github.com/jupyterlab/jupyterlab/pull/18568

The major shift in approach is to use a font installed via npm to ensure that an exact font version gets used regardless of the OS as the same font can have different versions and render differently depending on the delivery format.

Issue #5893, opened 2019-01-21
Theme extension made with create:theme command doesn't work


telamonian:
I made a new theme extension with the `jlpm run create:theme` command yesterday. At first it didn't work at all, but after fixing up `tsconfig.json` (the starting one copied from `theme-light-extension` isn't valid for a standalone extension) I was able to follow along with the instructions in the latest version of the [extension authoring docs](https://jupyterlab.readthedocs.io/en/latest/developer/extension_dev.html#extension-authoring). I built and installed it by running the following commands in the extension's repo:

```
npm install
npm run build
jupyter labextension install
```

Everything seems to be fine, and the new theme even shows up in the "Jupyterlab Theme" submenu. Which is awesome! However, when I try to select the new theme I get an error message:

```
Error Loading Theme
Stylesheet failed to load: /lab/api/themes/@telamonian/theme-darcula/index.css
```

In spite of the error message, Jupyterlab creates a new blank file at `/lab/api/themes/@telamonian/theme-darcula/index.css` and tries to use that as the theme. At that point, Jupyterlab is completely borked (100% non-interactive) until I manually reset the theme in `~/.jupyter/lab/user-settings/\@jupyterlab/apputils-extension/themes.jupyterlab-settings`.

It seems like everything in my `style` dir (including `style/index.css`) should be getting copied over to `themes/@telamonian/theme-darcula`, but that isn't happening. I'm also assuming that whatever is going on is somehow related to all of the changes to themes that were made in #5505.

Also, I don't know if this is related, but whenever I start up Jupyterlab I see a warning about the metadata of the theme extension:

```
[W 06:34:54.072 LabApp] Failed to fetch package metadata for '@telamonian/theme-darcula': <HTTPError 404: 'Not Found'>
```

telamonian:
You can find the complete code for the extension at https://github.com/telamonian/theme-darcula. I'll post the relevant seeming bits here.

Here's my `tsconfig.json`:

```
{
  "compilerOptions": {
    "allowSyntheticDefaultImports": true,
    "composite": true,
    "declaration": true,
    "noImplicitAny": true,
    "noEmitOnError": true,
    "noUnusedLocals": true,
    "esModuleInterop": true,
    "preserveWatchOutput": true,
    "module": "commonjs",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "target": "es2015",
    "lib": ["dom", "es2015"],
    "jsx": "react",
    "types": [],

[...]

PR #11683, opened 2021-12-14
Remove the `create:theme` script

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/5893

## Code changes

Remove the `create:theme` utility script from the code base.

I'm not sure this script was actually used by anyone so I propose we skip the deprecation and drop it entirely for 4.0.

The theme cookiecutter has been updated for JupyterLab 3.0 and has already been used to bootstrap new third-party themes.: https://github.com/jupyterlab/theme-cookiecutter

Also it's unlikely that new themes will be added to JupyterLab core anytime soon.

## User-facing changes

None for end users.

Theme authors should use the theme cookiecutter instead, as documented in https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#cookiecutters

`create:theme` script has been removed from the `buildutils` package.


fcollonval (approved):
This sounds reasonable. Thanks @jtpio

PR #11574, opened 2021-11-29
Windowed (Virtual) notebook

fcollonval:
## References

This is a Proof Of Concept to show the potential performance gain of using a fully windowed notebook.

PR #6824, opened 2019-07-13
WIP: Fix/slow search

krassowski:
## References

Fixes #6756

## Code changes

The major culprit of the slowness is re-rendering of the CodeMirror cells. This PR addresses this issue by:
- eliminating use of `rendered` property setter at search time setter which has a side-effects of refreshing the editor (instead introducing `renderedDirty` which does not refresh editor)
- removing unnecessary refreshing operations for search overlay in-between of search operations (by introducing `removeOverlay` argument to `endQuery`)
- disabling all layout recalculations during the actual search and then re-rendering cells gradually in a non-blocking manner, which eliminates interface freezes (implemented by hiding the Notebook widget for the search time and then refreshing editors of cells in background)

## User-facing changes

Positive:
- the JupyterLab interface no longer freezes for 10+ seconds when searching in a large notebook
- the code cell with the first result is now always available and rendered immediately

Negative (can be addressed later or maybe as a part of this PR):
- the re-rendering is sometimes noticeable:
   - in the first search after opening the notebook, when the markdown cells have empty editors
  -  in cells around the active cell when the active cell is close to the end of the notebook

## Backwards-incompatible changes

None.

### Before:
In large noteboks the UI has been freezing for tens of seconds (here showing response time 14s):
Screenshot from 2019-07-14 00-21-40

Which was caused by consecutive layout recalculations (cell-by-cell), repainting the entire notebook:
Screenshot from 2019-07-13 23-54-14

Each individuall cell (Code or Markdown) was limitted by the known bottlnecks of CodeMirror: 
Screenshot from 2019-07-13 20-43-23

### After:

The actual search is performed quickly (here in 1.5s), while the repainting of cells is spread in time allowing for full responsiveness of the user interface:
Screenshot from 2019-07-14 00-23-12

The first task repaints entire notebook once; second tasks repaints the first cells with matches (activated by `_stepNext`) and consecutive tasks are running in the background, refreshing editors for all cells with matches:
Screenshot from 2019-07-13 23-56-42

Issue #6756, opened 2019-07-03
Find/replace is very slow for large notebooks


krassowski:
The new notebook-wise find/replace widget is very slow for large notebooks and not fast enough for medium notebooks. Particularly detrimental to the user experience is the first "indexing" step which re-renders most of the cells when we begin to type the search phrase.

I have noticed it on my collection of real-life notebooks which I cannot share but, here are some 
test cases of similar size from [the gallery of interesting Jupyter Notebooks](https://github.com/jupyter/jupyter/wiki/A-gallery-of-interesting-Jupyter-Notebooks):
- medium notebook: [AEC.04 - Evolutionary Strategies and Covariance Matrix Adaptation.ipynb](https://raw.githubusercontent.com/lmarti/evolutionary-computation-course/master/AEC.04%20-%20Evolutionary%20Strategies%20and%20Covariance%20Matrix%20Adaptation.ipynb)
- large notebook: [not_so_obvious_python_stuff.ipynb](https://raw.githubusercontent.com/rasbt/python_reference/master/tutorials/not_so_obvious_python_stuff.ipynb)

**To Reproduce**

1. download and start the medium-size and lage notebook
2. search for "t"
3. observe the notebook freeze for ~2 seconds for the medium notebook and 10+ seconds for the large notebook
4. if you have enough patience, try typing "test" in the searchbox of the large notebook.

In the wild it happens for me on the same notebooks which cause #4292 (very irritating one as well).

**Expected behavior**
The search in medium-sized (~50 code cells) notebooks takes less than one second and search in large notebooks (200 code cells+) takes less than two seconds. The interfaces should not freeze during the search.

I would be happy to sacrifice the initial number of matches for responsiveness - for example get first 50 matches to start with and get the remaining ones loaded in the background later). Because who really cares about 3095 "t"s in the notebook when trying to type "test"?

**Desktop (please complete the following information):**

- OS: Ubuntu 19.04
- Browser: Version 75.0.3770.100 (Official Build) (64-bit)
- JupyterLab: 1.0.0

**Additional context**

When recording with the dev-tools profiler in Chrome the exact time for "t" search in the large notebook was 12.4 seconds; the bump on the left (0.5 sec) is opening of the find/replace widget (for comparison):

![Screenshot from 2019-07-03 16-06-00](https://user-images.githubusercontent.com/5832902/60603036-e4cbad80-9dac-11e9-8798-97bd7675a345.png)
 
[...]

PR #11689, opened 2021-12-14
Notebook search based on data model

fcollonval:
## References

This is a preliminary work for https://github.com/jupyterlab/jupyterlab/pull/11574 that switches the search feature in the notebook from view (aka CodeMirror and DOM nodes) search to data model (aka modelDB) for the inputs. The search on outputs is disabled by default and will looked at the rendered DOM tree if activated (when using virtual rendering this will trigger a warning that it may takes some time as outputs rendering will be needed).

## Code changes

- Refactor `@jupyterlab/documentsearch`:
  - Simplify the public API
  - Standardize how to define filters (only boolean type supported)
  - Move search provider to their respective package (notebook in @jupyterlab/notebook, text editor in @jupyterlab/fileeditor)
- Update search when cell list changes
- Don't change hidden state of input cells nor rendered state of markdown cells (this improve user experience and reduce performance related issue due to CodeMirror refresh).  
  On Markdown cells, they are unrendered when the user is highlighting next or previous match that are in a markdown cell.
- Simplify the `OutputArea` widget (and correct some bugs) as if trimmed the output area will only show the first outputs (following #11457)

## User-facing changes

image

`ISearchProvider` API is modified.


fcollonval:
Performance of this PR has been tested manually on the _large_ notebook [not_so_obvious_python_stuff.ipynb](https://raw.githubusercontent.com/rasbt/python_reference/master/tutorials/not_so_obvious_python_stuff.ipynb) mentioned in #6756

search_large_nb

In the issue, it was mentioned that the gain for the tested notebook is (presumably gain obtained by the associated PR #6824):

> I managed to cut the time of searching for `t` in the large notebook from about 14s to 7s

In the screencast above we can see that the search for `t` or `test` takes a couple of seconds. So I consider we are good to go regarding performance.

echarles:
Awesome @fcollonval 

Will a mimetype renderer be obliged to support this new search capability? In other words, what are the impacts for existing renderers such as e.g. https://github.com/jupyterlab/jupyter-renderers?

fcollonval:
> Will a mimetype renderer be obliged to support this new search capability? In other words, what are the impacts for existing renderers such as e.g. https://github.com/jupyterlab/jupyter-renderers?

No, the additional API has two folds:

- In rendermime interfaces, new optional API:  
```ts
    /**
     * Render a mime model.
     *
     * @param model - The mime model to render.
     * @param highlights - The highlights to render.
     *
     * @returns A promise which resolves when rendering is complete.
     *
     * #### Notes
     * This method may be called multiple times during the lifetime
     * of the widget to update it if and when new data is available.
     */
    renderModel(model: IMimeModel, highlights?: IHighlight[]): Promise<void>;

    /**
     * Highlights text fragment on the widget.
     *
     * @param highlights - The highlights to render.
     *
     * @returns A promise which resolves when rendering is complete.
     *
     * #### Notes
     * The highlights are sorted from the lowest to the highest position.
     *
     * This method may be called multiple times during the lifetime
     * of the widget to update it if and when new data is available.
     */
    renderHighlights?(highlights: IHighlight[]): Promise<void>;
```
- Register a search engine for additional mimetype:
```ts
/**
 * A plugin providing search engine for cell outputs.
 */
const search: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/rendermime-extension:search',
  requires: [ISearchProviderRegistry],
  activate: (app: JupyterFrontEnd, searchRegistry: ISearchProviderRegistry) => {
    textRendererFactory.mimeTypes.forEach(mimeType => {
      searchRegistry.registerMimeTypeSearchEngine(mimeType, TextSearchEngine);
    });

    [
      ...markdownRendererFactory.mimeTypes,
      'text/x-ipythongfm',
      'text/x-markdown',
      'text/x-gfm'
    ].forEach(mimeType => {
      searchRegistry.registerMimeTypeSearchEngine(
        mimeType,
        MarkdownSearchEngine
      );
    });

    htmlRendererFactory.mimeTypes.forEach(mimeType => {
      searchRegistry.registerMimeTypeSearchEngine(
        mimeType,
        HTMLStringSearchEngine
      );
    });
  },
  autoStart: true
};
```

---

Test using `jupyterlab-geojson` (only requires upgrading the dependencies to v4 alpha):

image

fcollonval:
Comments from the performance meeting:
- Document decision done: like ignoring the link markdown
    william: encourage you to document the choices related to the subtle things like links, math in markdown, ansi output messages. I would be thrilled if there was an official jupyter specification of what search should do, and then client implementors could try to follow that spec.
- Keep the GenericSearch for not supported outputs
- Look at other tools like VSCode to see what they are doing:
  - VSCode: search in the editor source only
    - No search on outputs
    - If there is a hit in markdown cell, it is switched from rendered to the editor when the user is pressing next/previous button

fcollonval:

fcollonval:
UI failure is not related.

packages/documentsearch/src/tokens.ts
```diff
@@ -13,26 +14,300 @@ export const ISearchProviderRegistry = new Token<ISearchProviderRegistry>(
...
+   */
+  supportReplace: boolean;
+}
+
+/**
+ * Type of filters
+ *
+ * TODO support additional filter type
+ */
+export interface IFiltersType {
```
afshin, line 44: I would propose called this `IFilters` and removing the word "Type" from the interface name.

One other approach could be making explicit `type` declarations like `BooleanFilter`, etc. and then having `get filters(): Array<BooleanFilter | FooFilter | BarFilter>

But probably just `interface IFilters` is good.

packages/documentsearch/src/tokens.ts
```diff
@@ -13,26 +14,300 @@ export const ISearchProviderRegistry = new Token<ISearchProviderRegistry>(
...
+
+  /**
+   * The text in the replace entry
+   */
+  replaceText: string;
+
+  /**
+   * Whether or not the replace entry row is visible
+   */
+  replaceEntryShown: boolean;
```
afshin, line 113: I would suggest `replaceEntryVisible` to be more in line with some other APIs that have `isVisible` attributes.

packages/documentsearch/src/tokens.ts
afshin: (I imagine you wanted to avoid having `IFilter` and `IFilters` but I think the addition of the word `Type` for an `interface` makes it suboptimal)

packages/documentsearch/src/tokens.ts
```diff
@@ -13,26 +14,300 @@ export const ISearchProviderRegistry = new Token<ISearchProviderRegistry>(
...
+
+  /**
+   * Stop a search and clear any internal state of the provider
+   */
+  endQuery(): Promise<void>;
+
+  /**
+   * Clear currently highlighted match.
+   */
+  clearHighlight(): void;
```
afshin, line 229: Will this always be trivial to implement synchronously?

Since the other functions return promises, it's low-cost to make this one return a promise if it'll ever be useful. But if it's always safe to assume this can be done synchronously then fair enough :+1:

packages/documentsearch/src/tokens.ts
```diff
@@ -13,26 +14,300 @@ export const ISearchProviderRegistry = new Token<ISearchProviderRegistry>(
...
+
+  /**
+   * The current index of the selected match.
+   */
+  readonly currentMatchIndex: number | null;
+
+  /**
+   * The number of matches.
+   */
+  readonly matchesSize: number | null;
```
afshin, line 282: I would suggest `matchCount`, here.

packages/fileeditor/src/searchprovider.ts
```diff
@@ -0,0 +1,76 @@
...
+export class FileEditorSearchProvider
+  extends CodeMirrorSearchProvider
+  implements ISearchProvider {
+  /**
+   * Constructor
+   * @param widget File editor panel
+   */
+  constructor(widget: FileEditorPanel) {
+    super();
+    this.editor = widget.content.editor as CodeMirrorEditor;
```
afshin, line 31: I am looking forward to a future where we don't have to cast, here 🚀

packages/outputarea/src/model.ts
```diff
@@ -14,12 +14,14 @@ import { ISignal, Signal } from '@lumino/signaling';
  */
 export interface IOutputAreaModel extends IDisposable {
   /**
-   * A signal emitted when the model state changes.
+   * A signal emitted when the output item changes.
+   *
+   * The number is the index of the output that changed.
    */
-  readonly stateChanged: ISignal<IOutputAreaModel, void>;
+  readonly stateChanged: ISignal<IOutputAreaModel, number | void>;
```
afshin, line 21: Is the `| void` for backward compatibility or is it necessary here?

packages/outputarea/src/widget.ts
```diff
@@ -429,52 +450,49 @@ export class OutputArea extends Widget {
...
+
+    if (!this._outputTracker.has(output)) {
+      void this._outputTracker.add(output);
+    }
+
+    const layout = this.layout as PanelLayout;
+    layout.insertWidget(index, output);
+
+    if (index === this._maxNumberOutputs) {
+      output.node.addEventListener('click', () => {
```
afshin, line 490: Can this event listener be added to the `OutputArea` class in the normal Lumino way instead of on-the-fly here?

packages/documentsearch/style/base.css
```diff
@@ -61,8 +61,8 @@
   margin: 2px;
 }
 
-.jp-DocumentSearch-focused-input {
-  border: var(--jp-border-width) solid var(--jp-cell-editor-active-border-color);
+.jp-DocumentSearch-input-wrapper:focus-within {
```
afshin, line 64: Neat 🤓

packages/rendermime-extension/src/index.ts
```diff
@@ -44,9 +44,9 @@ const plugin: JupyterFrontEndPlugin<IRenderMimeRegistry> = {
 };
 
 /**
- * Export the plugin as default.
+ * Export the plugins as default.
  */
-export default plugin;
+export default [plugin];
```
afshin, line 49: Is this necessary? Will it require a major version bump of the package?

afshin:
This is awesome and a monumental undertaking, thank you @fcollonval! I made a few suggestions primarily focusing on the public APIs. I hope we ship this one soon because it's awesome.

fcollonval:

packages/cells/src/searchprovider.ts
```diff
@@ -0,0 +1,717 @@
...
+import {
+  GenericSearchProvider,
+  IBaseSearchProvider,
+  IFilters,
+  ISearchMatch,
+  TextSearchEngine
+} from '@jupyterlab/documentsearch';
+import { IObservableString } from '@jupyterlab/observables';
+import { OutputArea } from '@jupyterlab/outputarea';
+import { Signal } from '@lumino/signaling';
```
afshin, line 18:
```suggestion
import { ISignal, Signal } from '@lumino/signaling';
```

packages/cells/src/searchprovider.ts
```diff
@@ -0,0 +1,717 @@
...
+    this._changed = new Signal<IBaseSearchProvider, void>(this);
+    this.cmHandler = new CodeMirrorSearchHighlighter(
+      this.cell.editor as CodeMirrorEditor
+    );
+  }
+
+  /**
+   * Changed signal to be emitted when search matches change.
+   */
+  get stateChanged(): Signal<IBaseSearchProvider, void> {
```
afshin, line 48:
```suggestion
  get stateChanged(): ISignal<IBaseSearchProvider, void> {
```

packages/documentsearch/src/searchmodel.ts
```diff
@@ -0,0 +1,308 @@
...
+    if (this._searchExpression) {
+      this.endQuery().catch(reason => {
+        console.error(
+          `Failed to end query '${this._searchExpression}.`,
+          reason
+        );
+      });
+    }
+
+    this.searchProvider.stateChanged.disconnect(this.refresh, this);
```
afshin, line 180: Does this cover everything (i.e., disconnection from the state changed signal and also disconnecting all listeners on the disposed signal)?
```suggestion
    Signal.clearData(this);
```

packages/documentsearch/src/searchprovider.ts
```diff
@@ -0,0 +1,152 @@
...
+   */
+  constructor(protected widget: T) {
+    this._changed = new Signal<this, void>(this);
+    this._disposed = false;
+  }
+
+  /**
+   * Signal indicating that something in the search has changed, so the UI should update
+   */
+  get stateChanged(): Signal<this, void> {
```
afshin, line 26:
```suggestion
  get stateChanged(): ISignal<this, void> {
```

packages/documentsearch/src/searchprovider.ts
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { Signal } from '@lumino/signaling';
```
afshin, line 4:
```suggestion
import { ISignal, Signal } from '@lumino/signaling';
```

packages/notebook/src/searchprovider.ts
```diff
@@ -0,0 +1,571 @@
...
+   * It is undefined behavior to use any functionality of the object
+   * after it has been disposed unless otherwise explicitly noted.
+   */
+  dispose(): void {
+    if (this.isDisposed) {
+      return;
+    }
+    super.dispose();
+
+    this.widget.content.placeholderCellRendered.disconnect(
```
afshin, line 144: Same question re: clearing signal data:
```suggestion
    Signal.clearData(this);
```

afshin:
Minor bookkeeping updates, otherwise, I think let's merge and iterate! :+1:

fcollonval:

afshin (approved):
Thank you @fcollonval! This is awesome 🚀

packages/rendermime/style/base.css
```diff
@@ -613,3 +613,15 @@ h6:hover .jp-InternalAnchorLink {
...
+  background-color: var(
+    --jp-search-unselected-match-background-color
+  ) !important;
+  color: var(--jp-search-unselected-match-color) !important;
+}
+
+.jp-mod-selected.jp-mimeType-highlight {
+  background-color: var(--jp-search-selected-match-background-color) !important;
+  color: var(--jp-search-selected-match-color) !important;
+}
```
krassowski, line 627: @fcollonval it seems to me that `.jp-mimeType-highlight` is not used, is this correct or am I missing something?

fcollonval: This is correct - this is an artifact of my trial for searching the unrendered output.

Issue #11633, opened 2021-12-08
Reusable boilerplate for federated extensions


jtpio:
### Problem

There is an increasing number of lab-based applications now reusing the federated extensions system to pick up prebuilt extensions automatically.

For example:

- JupyterLab Desktop: https://github.com/jupyterlab/jupyterlab-desktop
- RetroLab: https://github.com/jupyterlab/retrolab
- JupyterLite: https://github.com/jupyterlite/jupyterlite
- RISE extension for JupyterLab (https://github.com/damianavila/RISE/pull/605)
- Voila (soon to be a lab-based app, https://github.com/voila-dashboards/voila/pull/846)

However they all ship their own copy of the boilerplate code to enable this. Which mostly corresponds to the following files in the `federated_extensions` example:

**bootstrap.js**

https://github.com/jupyterlab/jupyterlab/blob/2688ce5a9038e94de8ab77e836b3c11f1d91c455/examples/federated/core_package/bootstrap.js#L1-L117

**index.js**

https://github.com/jupyterlab/jupyterlab/blob/2688ce5a9038e94de8ab77e836b3c11f1d91c455/examples/federated/core_package/index.template.js#L1-L181

**createShare** in `webpack.config.js`

https://github.com/jupyterlab/jupyterlab/blob/2688ce5a9038e94de8ab77e836b3c11f1d91c455/examples/federated/core_package/webpack.config.js#L72-L168

### Proposed Solution

Probably some of this bootstrap code could be placed in `@jupyterlab/builder`. Which could then be consumed by downstream packages by importing from `@jupyterlab/builder`.

jtpio:
Just closed https://github.com/jupyterlab/jupyterlab/pull/11693 as outdated, and also because this is something that should now be implemented in https://github.com/jupyterlab/jupyter-builder directly.

Leaving that issue opened for now for visibility.

Issue #11336, opened 2021-10-21
Drop managing source extensions with `jupyter labextension`?


jtpio:
### Problem

Having two ways to refer to extensions can be confusing to users and extension authors.

It also requires writing documentation (and keeping it up-to-date) about the differences:

- Managing Extension: https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#managing-extensions-with-jupyter-labextension
- https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#source-extensions
- https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#prebuilt-extensions

It also holds complexity within the JupyterLab code base.

### Proposed Solution

We drop support for managing source extensions with `jupyter labextension`:

Pros:

- Remove the extension manager, or update it to make it work with prebuilt extensions only -> less confusing to the end user who might not know the difference between the two types
- Remove quite a bit of code from the lab code base (`install`, `update`, `uninstall` commands) -> better maintenance
- Easier to reason about how extensions work in JupyterLab
- Focus efforts on a single way to provide extensions to JupyterLab. We can then build more focused tools for extension authors wanting to quickly hack around (a la classic notebook, also related #11075), without worrying about breaking the second way of installing extensions with `jupyter labextension install` 
- Maybe a good opportunity to streamline the `bootstrap.js` / `index.js` boilerplate and make it easier to reuse in other lab-based apps like retrolab, jupyterlite, voila, quetz-frontend
- Most of the other JupyterLab distributions reusing JupyterLab extensions do not support managing source extensions with `jupyter labextension install / update /uninstall`. They only load prebuilt extensions at startup:
  - JupyterLab Desktop
  - RetroLab
  - JupyterLite
  - Voila (soon to be a lab-based app)

Cons:

- in some cases it can be useful to be able bundle everything together. Dropping support for source extensions would also drop support for this

### Additional context

This came up during casual discussions. Also something we have been thinking about over in [JupyterLite](https://github.com/jupyterlite/jupyterlite). Trying to imagine what it would look like if everything was a federated extension, even the core plugins.

Also many third-party extensions have now been updated to be prebuilt extensions since it's a much better experience for the end user. There should now be a lot less users installing extensions with `jupyter labextension install`.

---

Opening this issue now to check whether this is something we would like to consider for 4.0.

It was also something briefly considered for 3.0, but since the prebuilt extension system was still new at the time it was better to keep both system for the 3.0 release.

[...]

PR #11693, opened 2021-12-15
Reusable util functions to enable federated extensions in other JupyerLab distributions

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11633

Move some utility functions to `@jupyterlab/builder` so they are more easily reusable in downstream JupyterLab distributions.

## Code changes

- [x] Move `createShare`
- [ ] Move `bootstrap` functions
- [ ] Make more webpack config reusable, like `ModuleFederationPlugin`
- [ ] Type the utility functions
- [ ] Export at the top-level of `@jupyterlab/builder`, or under something else like `federation`?
- [ ] Make JupyterLab (`dev_mode`) and the `federated` example depend on this. Both already depend on `@jupyterlab/builder` in their `devDependencies`. Most of the other JupyterLab distributions follow a similar structure as the `federated` example.
- [ ] Document this for distribution authors? Could also be done in a separate PR. Contnrary to extension authors, distribution authors can be aware of internals such as webpack.


builder/src/utils.ts
```diff
@@ -0,0 +1,103 @@
+// TODO: remove
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+
+/**
+ * Create the webpack ``shared`` configuration
+ */
+export function createShared({ extensions, resolutions, singletonPackages }) {
```
bollwyvl, line 8: I agree with the linter: a documented `IOptions` and return value would be lovely

builder/src/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      let oldConfig = mergedShare[pkg];
+
+      // if the old one has import: false, use the new one
+      if (oldConfig.import === false) {
+        mergedShare[pkg] = config;
+      }
+    }
+  }
+
+  Object.assign(shared, mergedShare);
```
bollwyvl, line 86: would be more idiomatic as a `{...shared, ...mergedShare}`?

builder/src/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // Transform any file:// requiredVersion to the version number from the
+  // imported package. This assumes (for simplicity) that the version we get
+  // importing was installed from the file.
+  for (let [pkg, { requiredVersion }] of Object.entries(shared)) {
+    if (requiredVersion && requiredVersion.startsWith('file:')) {
```
bollwyvl, line 92: this smells like something that might need to be extensible... `npm` allows _all kidns_ of icky things...

builder/src/index.ts
```diff
@@ -5,3 +5,4 @@
 
 export * from './build';
 export * from './webpack-plugins';
+export * from './utils';
```
bollwyvl, line 8: `util` kinda under-sells this... perhaps `./sharing`?

builder/src/utils.ts
jtpio: Yes the whole set of utils needs proper typing.

builder/src/utils.ts
jtpio: Right, we might still need to keep that for now since it seems to be handling source extensions. Which we'll hopefully drop at some point in the future (in https://github.com/jupyterlab/jupyterlab/issues/11336 or a follow-up)

fcollonval:
Part of this may be tackled during the GSoC project on creating a [standalone builder](https://github.com/jupyterlab/jupyterlab-builder). For now Ronan is focusing on the Python code. But we may have time to work on the JavaScript assets too.

PR #11698, opened 2021-12-15
Share execution count, nbformat and metadata

davidbrochart:
## References

#11599

## Code changes

When we save a notebook from the back-end using `y-py`, we will need data that is currently not set in the shared model e.g. when a notebook is loaded.
- Populate every cell shared model with `execution_count` when a notebook is loaded.
- Populate the notebook shared model with `nbformat` and `nbformat_minor` when a notebook is created/loaded, and with `metadata` whenever it changes. For `metadata`, we didn't share `kernelspec` and `language_info`, I'm not sure why.

## User-facing changes

None

None

### DIFF
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -632,10 +632,12 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
     const trusted = this.trusted;
     const cell = options.cell as nbformat.ICodeCell;
     let outputs: nbformat.IOutput[] = [];
+    let execution_count: number | null = null;
     const executionCount = this.modelDB.createValue('executionCount');
     if (!executionCount.get()) {
       if (cell && cell.cell_type === 'code') {
-        executionCount.set(cell.execution_count || null);
+        execution_count = cell.execution_count || null;
+        executionCount.set(execution_count);
         outputs = cell.outputs ?? [];
         // If execution count is not null presume the input code was the latest executed
         // TODO load from the notebook file when the dirty state is stored in it
@@ -654,6 +656,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
     globalModelDBMutex(() => {
       const sharedCell = this.sharedModel as models.ISharedCodeCell;
       sharedCell.setOutputs(outputs);
+      sharedCell.execution_count = execution_count;
     });
     this._outputs = factory.createOutputArea({ trusted, values: outputs });
     this._outputs.changed.connect(this.onGenericChange, this);
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -118,6 +118,8 @@ export interface ICodeCellModel extends ICellModel {
    * Clear execution, outputs, and related metadata
    */
   clearExecution(): void;
+
+  sharedModel: models.ISharedCodeCell;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


hbcarlos:
We decided not to share these attributes because collaborators could use different kernels.

Maybe we should open the discussion again and see if we should share the information but allow some collaborators to use other kernels.

davidbrochart:
> For `metadata`, we didn't share `kernelspec` and `language_info`, I'm not sure why.

@hbcarlos mentioned that we currently allow users to work on the same notebook with different kernels. If is true that requiring every user to use the same kernel would be problematic anyway, because they might work on a machine where this kernel cannot be installed. Yet they could perfectly execute the notebook if they had e.g. Python 3.9 instead of Python 3.10.
One option is to not populate the kernel information in the notebook metadata, but that would mean each time a user opens a notebook, they will be asked to choose one.
Another option is to set the kernel information in the shared model, but only at creation/loading time. Any subsequent change would not be stored, so if a user opens a notebook which requires a kernel that is not installed on his machine, they will be asked to choose another one but this won't change the kernel information in the saved notebook. We would probably need a way to change the kernel though (in the saved notebook), with a more explicit action required from the user (like selecting "Save kernel information" in the menu).

dmonad:
What we are currently doing really is a bit inconsistent. Because we are simply not listening to the event when changing kernels. That means that we are storing the default kernel, and then we are unable to switch to another kernel. 

Conceptually it also makes more sense to me to force all users to use the same kernel. Eventually, we could allow switching kernels without forcing other clients to also use that kernel. 

I would argue that in almost all cases you want to enforce that everyone uses the same kernel. Otherwise, you'd need to convince every single user to switch to kernel X to use a specific feature.

packages/notebook/src/model.ts
```diff
@@ -116,6 +114,8 @@ export class NotebookModel implements INotebookModel {
 
     (this.sharedModel as models.YNotebook).dirty = false;
     this.sharedModel.changed.connect(this._onStateChanged, this);
+    (this.sharedModel as models.YNotebook).nbformat = this._nbformat;
```
dmonad, line 117: I feel that `nbformat` should be initialized in the `shared-model` package. If we initialize it here, we will set `nbformat` every time this class is instantiated (i.e. every time a new user joins).

The same is true for the `dirty` flag above.

Here, we should instead only check whether `this.nbformat === sharedModel.nbformat` and throw an error if this is not true.

packages/notebook/src/model.ts
```diff
@@ -447,11 +447,9 @@ close the notebook without saving it.`,
     metadata: IObservableJSON,
     change: IObservableMap.IChangedArgs<ReadonlyPartialJSONValue | undefined>
   ): void {
-    if (!UNSHARED_KEYS.includes(change.key)) {
```
dmonad, line 450: This is fine if we really want to enforce shared kernels.

packages/cells/src/model.ts
```diff
@@ -654,6 +656,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
     globalModelDBMutex(() => {
       const sharedCell = this.sharedModel as models.ISharedCodeCell;
       sharedCell.setOutputs(outputs);
+      sharedCell.execution_count = execution_count;
```
dmonad, line 659: Same here, we should never initialize `execution_count` in the constructor of some class. Content should be initialized in the `create`/`createStandalone` methods of the shared model. Otherwise, we will always initialize to 0 when a new client joins the session.

I'm a bit puzzled why we initialize the outputs above..

dmonad (changes requested):

packages/notebook/src/model.ts
hbcarlos: I can open another PR to initialize everything in the shared model

hbcarlos:
> Another option is to set the kernel information in the shared model, but only at creation/loading time. Any subsequent change would not be stored, so if a user opens a notebook which requires a kernel that is not installed on his machine, they will be asked to choose another one but this won't change the kernel information in the saved notebook. We would probably need a way to change the kernel though (in the saved notebook), with a more explicit action required from the user (like selecting "Save kernel information" in the menu).

I like this idea.

packages/notebook/src/model.ts
dmonad: This is kind of a big issue. `nbformat` must only be initialized in the `create` and `createStandalone` methods. Can you please revert the changes in this PR? 

I added some comments about why we should avoid casting to the concrete implementation whenever possible:
https://github.com/jupyterlab/jupyterlab/pull/11525/files#r771420793

Please also check other places where you casted to the concrete Yjs implementation.

davidbrochart:
Thanks Carlos, I copied these changes locally and checked that it works well with https://github.com/jupyterlab/jupyterlab/pull/11599.
You might want to revert the change in `dev_mode/package.json`.

hbcarlos:
Done! I removed that integrity commit and rebased with the master branch.

fperez:
Mmh, btw - is it possible that this PR will address a weird behavior I was noticing in the hub referenced in #12154 (cc @yuvipanda)? I was noticing that the hub kept constantly asking me to re-select my kernel for notebooks that I had only saved in that same hub, over and over.  I couldn't figure out why it wasn't remembering my kernel choice.

Now that I see this issue it sounds like they could be related: at least since disabling RTC, that annoyance seems also gone...

fcollonval:
@davidbrochart @hbcarlos I think this can be closed as out-dated, isn't it?

PR #11685, opened 2021-12-14
Remove the `add:sibling` script

jtpio:
## References

This removes the top-level `add:sibling` script, and the associated script in `@jupyterlab/builtutils`. And the test code in `ci_script.sh` commented out.

## Code changes

The `add:sibling` script was already removed from the documentation in https://github.com/jupyterlab/jupyterlab/pull/10024.

Instead folks can use the federated extension approach with `--extensions-in-dev-mode`, or `--splice-source` (https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html?#development-workflow-for-source-extensions).

## User-facing changes

None

## Backwards-incompatible changes

Removed the `add-sibling` script from `@jupyterlab/buildutils`.

PR #8037, opened 2020-03-12
Add ability to include third party extension in dev mode

saulshanabrook:
This PR adds the ability to run dev mode with third party extensions enabled.

Currently the only way to do this is with add-sibling which really messes up the working directory and also requires you to build the other extension with JuptyerLab itself. This causes all sorts of problems if the extensions are not written in exactly the same ES6/webpack way that JupyterLab itself is.

This PR adds the option to instead just install extensions as normal npm packages and require them to bring them in. I haven't tried it yet for local extensions, I mainly wanted this to test published extensions against master and not have the ability to edit them. 

## References

Possibly fixes https://github.com/jupyterlab/jupyterlab/issues/6911  and https://github.com/jupyterlab/jupyterlab/pull/7179.

Built to let me test plotly widgets against core for benchmarking in this PR https://github.com/jupyterlab/jupyterlab/pull/8020 to help address long load times for notebooks https://github.com/jupyterlab/jupyterlab/issues/4292

## Code changes

Adds a new key `externalExtensions` to the `jupyterlab` object in the `dev_mode/package.json`. By adding package to version mappings here you can make JupyterLab require them in dev mode and add them to the `dependencies`.

## User-facing changes

None.

## Backwards-incompatible changes

None.

PR #11705, opened 2021-12-16
Drop support for `externalExtensions`

jtpio:
## References

As we try to encourage folks to use federated extensions (https://github.com/jupyterlab/jupyterlab/issues/11336), we should limit the options for installing and using source extensions in a dev build of JupyterLab.

The recommended alternative would be to use `--extensions-in-dev-mode` to enable federated extensions: https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html?#developing-a-prebuilt-extension

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/11685

## Code changes

`externalExtensions` is currently documented in the Extension Developer Guide: https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html?#development-workflow-for-source-extensions

- [x] Remove support for `externalExtensions`, which consisted in manually editing `dev_mode/package.json` to include local extensions in a dev build.
- [x] Document the change in case some were relying on this behavior

## User-facing changes

None

Remove `externalExtensions` from `@jupyterlab/application-top`.

Probably not a real breaking change since this was meant for development purposes.

### DIFF
--- a/buildutils/src/ensure-repo.ts
+++ b/buildutils/src/ensure-repo.ts
@@ -413,11 +413,7 @@ function ensureJupyterlab(): string[] {
   corePackage.jupyterlab.extensions = {};
   corePackage.jupyterlab.mimeExtensions = {};
   corePackage.jupyterlab.linkedPackages = {};
-  // start with known external dependencies
-  corePackage.dependencies = Object.assign(
-    {},
-    corePackage.jupyterlab.externalExtensions
-  );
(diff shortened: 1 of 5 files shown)


jtpio:
For more context `externalExtensions` predates the federated extension system and was added in https://github.com/jupyterlab/jupyterlab/pull/8037.

fcollonval (approved):
Thanks @jtpio

PR #1631, opened 2017-02-06
Refactor completer plugin into a standalone plugin.

afshin:

PR #11707, opened 2021-12-16
Updated typedoc

hbcarlos:
It looks like typedoc v0.21.2 doesn't support TypeScript 4.5.

There is a new release, typedoc v0.22.10 with support for TypeScript v4.5, see: https://github.com/TypeStrong/typedoc/releases/tag/v0.22.10

## References
Follow-up #11594

## Code changes
Update to typedoc v0.22.100

cc @jtpio


bollwyvl:
If we do update the theme, it would be lovely if it was published as a package... or perhaps moved up to lumino? these things are super-annoying to fix everywhere.

hbcarlos:
Do you mean moving/publishing it somewhere to reuse across projects?

I can do it. I'm not sure what would be the best place, though.

jtpio:
Probably this could be done as a follow-up so we can get this fix in first?

bollwyvl:
Yes, definitely a "some day".

Another tack altogether: it might even be time to take another look at
sphinx-jsdoc, though it also had a tightly pinned version of typedoc last
time I looked. But getting the whole ts/py API into one addressable,
searchable site would be superb.

fcollonval:
Thanks @hbcarlos it looks good on the CI: https://jupyterlab--11707.org.readthedocs.build/en/11707/api/index.html

@bollwyvl do you mean that project from Mozilla? https://github.com/mozilla/sphinx-js

bollwyvl:
>  that project from Mozilla? https://github.com/mozilla/sphinx-js

Yep, that's the one. It's used over on [pyodide](https://github.com/pyodide/pyodide/blob/b401350a2a972ad48a80d8def635c310fa9acbb0/docs/sphinx_pyodide/sphinx_pyodide/jsdoc.py) for [example](https://pyodide.org/en/stable/usage/api/js-api.html). 

Looks like the version of `typedoc` (and therefore `typescript`) it supports hasn't been updated in some time. Last i checked it didn't like our liberal use of eponymous `class`/`namespace` and the "last in" would win.

I think if we ever offered a more full, out-of-the-box solution for first- and third-party apps and extensions to document things, e.g. `jupyterlab-sphinx`, we'd want to figure out how to make that work... sure, typedoc looks nice, and one _can_ google search it, but sphinx is still the least-worst system for hitting all the quandrants of multi-language documentation, to my knowledge... and `myst` takes much of the pain of `rst` away, without losing too much power.

hbcarlos:
> Thanks @hbcarlos it looks good on the CI: https://jupyterlab--11707.org.readthedocs.build/en/11707/api/index.html

Yes, but is because I changed the theme from our custom one to the default one from typedoc. Looks like in typedoc 0.22 they decided to change the way custom themes are build from handlebars to typescript plugins.
I'm still investigating how to update the theme. see:
https://github.com/TypeStrong/typedoc/releases?page=2

>HTML output is now rendered with JSX instead of Handlebars, closes #1631.
This change provides major performance benefits, reducing rendering time by up to 10x for several benchmarked projects.
It also allows themes to be easily type checked, preventing mistakes when creating custom themes.
Removing Handlebars also fixed memory leaks when --watch was specified due to Handlebar's caching mechanism.
This change breaks all existing custom themes, so a theme created for v0.21 or earlier will not work in v0.22.
See internal-docs/custom-themes.md for documentation on how to create a custom theme in v0.22.

hbcarlos:
I updated the theme creating a typedoc plugin as indicated [here](https://github.com/TypeStrong/typedoc/blob/master/internal-docs/custom-themes.md) and following [this](https://github.com/Gerrit0/typedoc-custom-theme-demo) example. Unfortunately, typedoc uses a custom JSX module for the theme that I can not figure out how it works, see https://typedoc.org/api/modules/JSX.html. I'm able to build the plugin (new custom theme) but when building the API documentation of Jupyterlab (`jlpm run docs`) there are a few errors that I think come from a conflict between the react dependency of jupyterlab and typedoc.

jtpio:
> when building the API documentation of Jupyterlab (`jlpm run docs`) there are a few errors that I think come from a conflict between the react dependency of jupyterlab and typedoc.

Is this the error in question? https://github.com/jupyterlab/jupyterlab/runs/4594634160?check_suite_focus=true

hbcarlos:
> Is this the error in question? https://github.com/jupyterlab/jupyterlab/runs/4594634160?check_suite_focus=true

Hey @jtpio, thanks for looking into this PR.

No, it is not that error. That error is because now we need to build the theme before running `jlpm run docs`. I'm talking about the error when running `jlpm run docs` after building the theme. Let me update the CI.

jtpio:
OK thanks.

I was wondering that since this PR is already a fix, maybe we can merge it and fix the remaining theme related issues as follow-ups?

hbcarlos:
> I was wondering that since this PR is already a fix, maybe we can merge it and fix the remaining theme related issues as follow-ups?

Yes, I'll remove the theme for now.

jtpio (approved):
Thank you!

jtpio:
> Yes, I'll remove the theme for now.

Let's open a new issue to track this if it still needs a fix?

hbcarlos:
Thanks, @jtpio!

> Let's open a new issue to track this if it still needs a fix?

I opened a new issue for the theme.
https://github.com/jupyterlab/jupyterlab/issues/11788

PR #11711, opened 2021-12-17
Removing early bail out in ToC widget update

### LINKED ISSUE #11709: ToC Heading numbering not rendered on open notebook
## Description

The heading numbering is not rendered after the notebook is opened. Only if the ToC widget is activated, the 

## Reproduce

heading_numbering

## Expected behavior

The heading numbering should be displayed right after the notebook is opened.

schmidi314:
## References

Fixes #11709 

## Code changes

Removing early bail out in ToC widget update to trigger rendering of notebook header numbering also if ToC widget is hidden.

## User-facing changes

Headings now numbered correctly.

none

### DIFF
--- a/packages/toc/src/toc.tsx
+++ b/packages/toc/src/toc.tsx
@@ -122,11 +122,6 @@ export class TableOfContents extends Widget {
    * @param msg - message
    */
   protected onUpdateRequest(msg: Message): void {
-    if (this.isHidden) {
-      // Bail early
-      return;
-    }
-
     let title = this._trans.__('Table of Contents');
     if (this._current) {
       this._headings = this._current.generator.generate(


fcollonval:
I will start a large rework of the notebook ToC as part of v4 Roadmap to use the notebook model rather than the current approach. So I'm in favor of either merging this as is (although it has some performance impact) to be able to release it as part of 3.2.x or not doing anything as this will be reworked for v4.

schmidi314:
> I will start a large rework of the notebook ToC as part of [v4 Roadmap](https://github.com/jupyterlab/jupyterlab/issues/9647#issuecomment-965612826) to use the notebook model rather than the current approach. So I'm in favor of either merging this as is (although it has some performance impact) to be able to release it as part of 3.2.x or not doing anything as this will be reworked for v4.

I don't know if that was a question to me :) But since v4 is still like half a year from now, I'd of course love to see it in 3.2.x.

jtpio:
> although it has some performance impact

Is this bug fix worth the performance impact?

Adding to `3.2.x` so it can indeed be released earlier than 4.0.

fcollonval:
In fact those lines where introduced in https://github.com/jupyterlab/jupyterlab/pull/11356 and not backported to 3.2.x. So this is a v4 bug. So we could merge it as is. And thanks @schmidi314 for pointing that out - it will be critical to check when I start reworking the ToC extension.

jtpio:
Thanks!

Issue #11710, opened 2021-12-16
Autocomplete broken for console


mlucool:
## Description

I'm pretty sure autocomplete used to work in console as well. Now using tab adds a space.

## Reproduce
![ipython-autocomplete](https://user-images.githubusercontent.com/1813603/146465854-526fde37-0365-4b7e-a632-728bfb1e250a.gif)

## Expected behavior

Autocomplete works with tab press in console.

## Context

Lab@3.2.5

echarles:
I have quickly tried on 3.1.0 and console autocomplete works well with the default builtin completer but does not work with LSP (which looks like the one you are showing @mlucool )

<img width="433" alt="Screenshot 2021-12-17 at 07 45 19" src="https://user-images.githubusercontent.com/226720/146501285-b3712cbc-5589-4ac2-bf41-79066280dbba.png">

<img width="817" alt="Screenshot 2021-12-17 at 07 55 04" src="https://user-images.githubusercontent.com/226720/146502460-1b117301-2e63-4a33-bd30-bbfe53e5db00.png">

mlucool:
This seems unrelated to LSP. With a fresh install, I can see this works as expected with 3.2.0 and does not work with 3.2.5

vosak:
I experience the same. I found a kinda annoying workaround - to switch to a different tab and back to the console, the completer then works until a command execution. The tab switch is needed after each command execution in the console.

krassowski:
Thanks @vosak this is a useful observation.

It worked fine in 3.2.2 but it seems like it regressed in 3.2.3. This was a small release, and contained a bugfix related to completer (https://github.com/jupyterlab/jupyterlab/pull/11398), but we indeed did not add a test for console.

@dmonad could it be possible that we also need to make prompt cells creation y-js aware for the completer to work after the changes to `cursorActivity` event and/or mutex (note: the prompt cells are based on code cell; code cells in notebook are y-js aware and work fine; I wonder what is different about the prompt cells that they bail)? The relevant place in the code is:

https://github.com/jupyterlab/jupyterlab/blob/52e4ae9d959838023043853c66bfa5732513e263/packages/console/src/widget.ts#L588-L596

echarles:
[...]

PR #11533, opened 2021-11-23
Add side-by-side rendering as global setting

jess-x:
## References
https://github.com/jupyterlab/jupyterlab/issues/11376

## Code changes

- Added `sideBySideRendering` to notebook setting schema
- Made `renderSideBySide` in `notebook-extension` listen to `sideBySideRendering` to notebook setting

## User-facing changes
- Enabled setting side-by-side rendering as default 

## Backwards-incompatible changes
I am not aware of any.

PR #11718, opened 2021-12-18
Backport/3.2.x/11608 Toggle side-by-side rendering for current notebook

echarles:
Manual Backport to 3.2.x branch of Toggle side-by-side rendering for current notebook https://github.com/jupyterlab/jupyterlab/pull/11608


echarles (commented):
CI green except check_i18n and enforce-label which I have seen failing also in previous backport PR. @fcollonval Can you confirm it is OK to merge?

krassowski:
IMO this is 3.3 material. I wouldn't mind the translations experience breakage as much as the settings key change.

echarles:
> IMO this is 3.3 material.

Correct. The way we will release 3.3 is to branch from `3.2.x`, this is why this PR targets `3.2.x`

krassowski:
So I guess we should create the 3.3.x branch soon so we could still have the same workflow when backporting bug fixes to 3.2.x? I think we will need at least one more 3.2.x to fix the console completer regression.

echarles:
I have added https://github.com/jupyterlab/jupyterlab/issues/11710 to the list of PR to go in 3.3.

I think the 3.3.x branch would typically be created just after the 3.3 release

I will merge this backport PR (green CI) unless anyone says otherwise.

fcollonval:
> I think the 3.3.x branch would typically be created just after the 3.3 release

Actually this is not correct because, we need to be able to create patch release (for potential security fix).
Following the discussion at this week meeting, I'll create the 3.3.x branch.

> Thanks @krassowski for pointing this was merged.

fcollonval:
To be able to rebase 3.3.x on 3.2.x up to the release date, I will revert this PR on 3.2.x and then re apply it targeting 3.3.x

echarles:
> Actually this is not correct because, we need to be able to create patch release (for potential security fix).

I have read the discussion, but not completely clear on that.

After the 3.2 release, we have merged many features in 3.2.x branch. Is this one the only one to be reverted or do we have to do that for all the PRs merged in 3.2.x?

fcollonval:
The trouble of this PR is that is changes a setting key name => hence it should be part of 3.3

As far as of 3.1.x and 3.2.x ([changelog link](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html)), there has been very few new features brought in patch releases. The only one I can spot that is actually not related to some fixes is https://github.com/jupyterlab/jupyterlab/pull/11533 that this PR is altering.

If we don't revert this one on 3.2.x and need to publish a new 3.2.# release, then this will break the behavior introduced in 3.2.5 as the side by side rendering setting is renamed.

fcollonval:
> do we have to do that for all the PRs merged in 3.2.x?

So no only this one.

echarles:
Sounds good. Thx for the explanations.

PR #11734, opened 2021-12-22
Parse settings as JSON 5 in services

fcollonval:
## References

Fixes https://github.com/jupyterlab/jupyterlab_server/issues/217

## Code changes

Parse response from in settings service as JSON 5 not standard JSON.

## User-facing changes

Settings can be JSON 5 - in particular `Infinity` can now be used.

N/A

### DIFF
--- a/packages/services/package.json
+++ b/packages/services/package.json
@@ -55,6 +55,7 @@
     "@lumino/disposable": "^1.10.1",
     "@lumino/polling": "^1.9.1",
     "@lumino/signaling": "^1.10.1",
+    "json5": "^2.1.1",
     "node-fetch": "^2.6.0",
     "ws": "^7.4.6"
   },
--- a/packages/services/src/setting/index.ts
+++ b/packages/services/src/setting/index.ts
@@ -7,6 +7,8 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 
 import { DataConnector } from '@jupyterlab/statedb';
 
+import * as json5 from 'json5';
+
 import { ServerConnection } from '../serverconnection';
 
 /**
@@ -60,7 +62,8 @@ export class SettingManager extends DataConnector<
     }
 
     // Assert what type the server response is returning.
-    return response.json() as Promise<ISettingRegistry.IPlugin>;
+    const content = await response.text();
+    return json5.parse(content) as ISettingRegistry.IPlugin;
   }
 
   /**
@@ -80,7 +83,8 @@ export class SettingManager extends DataConnector<
       throw new ResponseError(response);
     }
 
-    const json = await response.json();
+    const content = await response.text();
+    const json = json5.parse(content);
     const values: ISettingRegistry.IPlugin[] =
       json?.['settings']?.map((plugin: ISettingRegistry.IPlugin) => {
         plugin.data = { composite: {}, user: {} };


fcollonval:
Kicking the CI

jasongrout:
I think it would be better if the server did not send nonstandard JSON (for example, it could send the setting just as raw text to be interpreted as JSON5 on the frontend). If we send nonstandard JSON in response to a settings request, it breaks the expectation of the rest api that the response is valid JSON.

fcollonval:
> I think it would be better if the server did not send nonstandard JSON (for example, it could send the setting just as raw text to be interpreted as JSON5 on the frontend). If we send nonstandard JSON in response to a settings request, it breaks the expectation of the rest api that the response is valid JSON.

I'm not sure to get it. You would like to do something like this on the server:

```
settings = json5.load(file)
# Test it is a valid json
settings = json.loads(json.dumps(settings))
```

And similarly on the frontend to ensure user settings sent are valid json?

So that we will allow json5 syntax but reject specific things like `Infinity` or `NaN`?

jasongrout:
> And similarly on the frontend to ensure user settings sent are valid json?

No. I would send the json5 source as a raw unparsed string, and then parse that string on the frontend as json5. That way what comes over the wire is actually valid JSON.

jtpio:
Bumping from `3.2.x` to `3.3.x` for consideration. But could also be added to `4.0` to not block the 3.3 release.

fcollonval:
@jasongrout I have a trouble with this one. If I got it correctly, the `settings` attributes in the object returned by the settings REST endpoint will be a raw string instead of a JSON object.

So how do we achieve changing `jupyterlab_server` API without breaking older JupyterLab versions that will assume that the `settings` attribute is an object.

fcollonval:
Should we test for the content type header?

Should we use `application/json5` (xref https://github.com/json5/json5/issues/80)

jasongrout:
I was proposing sending settings as a raw string as a breaking change in jlab server.

jasongrout:
I think using the content type would technically work, but it would be the first rest endpoint that didn't return JSON, which I think is a bigger architecture decision. I think I'm -0.5 on sending something other than JSON, just because there is a very standard convention to send JSON for rest endpoints. If it keeps sending JSON, then it keeps being broken. If we stop sending JSON, then there will be lots of confusion about it, I think.

In my mind, I was imagining a breaking change in jlab server where we send *different* JSON (the raw string), and us depending on it in jlab 4.0, and making a clean cutover then. Iirc, we already have this pattern of sending a raw string, so we'd just be applying that existing pattern here.

fcollonval:
Sorry we misunderstood. I had in mind to send a raw string as you propose but using the content type to flag that the attribute is a raw string that need json5 parsing. But this is not a great idea (hence our misunderstanding) as by looking at the content type people will think the body they got does not need an additional parsing.

So closing this as the approach will be different.

jasongrout:
> I had in mind to send a raw string as you propose but using the content type to flag that the attribute is a raw string that need json5 parsing. But this is not a great idea (hence our misunderstanding) as by looking at the content type people will think the body they got does not need an additional parsing.

Ah, I see what you were saying. Yes, I agree that we'd be misapplying the json5 content type in this situation (well, technically not because IIRC json is a subset of json5, but it would still be confusing).

PR #11739, opened 2021-12-23
Update `lighthouse` dependency in `@jupyterlab/testutils`

jtpio:
## References

Update to a newer version of `lighthouse` in `@jupyterlab/testutils`.

## Code changes

As a result this also continues cleaning up old dependencies in `yarn.lock` such as `xmldom`.

## User-facing changes

None (used for dev only).

None

### DIFF
--- a/testutils/package.json
+++ b/testutils/package.json
@@ -68,7 +68,7 @@
     "@types/jest": "^26.0.10",
     "@types/node-fetch": "^2.5.4",
     "jest-retries": "^1.0.1",
-    "lighthouse": "6.3.0",
+    "lighthouse": "9.2.0",
     "typescript": "~4.5.2"
   },
   "publishConfig": {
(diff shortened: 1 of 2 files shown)


testutils/package.json
```diff
@@ -68,7 +68,7 @@
     "@types/jest": "^26.0.10",
     "@types/node-fetch": "^2.5.4",
     "jest-retries": "^1.0.1",
-    "lighthouse": "6.3.0",
+    "lighthouse": "9.2.0",
```
jtpio, line 71: Maybe this could also be switched to `^`?

```suggestion
    "lighthouse": "^9.2.0",
```

jtpio: We don't seem to be running `lighthouse` checks on CI at the moment, and might then be missing issues related to such upgrades.

fcollonval (approved):
Thanks @jtpio

PR #11742, opened 2021-12-24
Remove deprecated `KarmaTestApp` and `JestApp`

jtpio:
## References

The `KarmaTestApp` and `JestApp` were marked as `removed_version=4` here:

https://github.com/jupyterlab/jupyterlab/blob/9ad40ceed4f74470f2ea1d99f12c883f8d460a16/jupyterlab/tests/test_app.py#L327

## Code changes

- Remove the deprecated `KarmaTestApp` and `JestApp` test apps.
- Remove mention of `karma` in the extension developer docs

## User-facing changes

None

Breaking for those that were still relying on importing `KarmaTestApp` and `JestApp` from `jupyterlab.tests.test_app`.


jtpio:
Pushed one more commit to remove `JestApp` which was also marked as deprecated.

fcollonval (approved):
Thanks @jtpio 

Optionally there are import not used in `jupyterlab/tests/test_app.py` that would be good to clean as well.

jtpio:
Thanks @fcollonval.

I just pushed one more commit to remove the unused imports.

Issue #11743, opened 2021-12-24
Python check is failing on CI


jtpio:
## Description

This is an issue to track the current CI failure which seems to be happening on all PRs for the past couple of days. With the following:

```
=================================== FAILURES ===================================
_____________________ TestExtension.test_enable_extension ______________________

self = <jupyterlab.tests.test_jupyterlab.TestExtension testMethod=test_enable_extension>

    def test_enable_extension(self):
        options = AppOptions(app_dir=self.tempdir())
        assert install_extension(self.mock_extension, app_options=options) is True
        assert disable_extension(self.pkg_names['extension'], app_options=options) is True
        assert enable_extension(self.pkg_names['extension'], app_options=options) is True
        info = get_app_info(app_options=options)
        assert '@jupyterlab/notebook-extension' not in info['disabled']
        name = self.pkg_names['extension']
>       assert name not in info['disabled']
E       AssertionError: assert '@jupyterlab/mock-extension' not in {'@jupyterlab/mock-extension': False}

jupyterlab/tests/test_jupyterlab.py:566: AssertionError
...
FAILED jupyterlab/tests/test_jupyterlab.py::TestExtension::test_enable_extension
```

## Reproduce

Example run: https://github.com/jupyterlab/jupyterlab/runs/4625267017?check_suite_focus=true

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/11744

PR #11744, opened 2021-12-24
Fix handling of disabled extensions

jtpio:
## References

Attempt at fixing #11743 

## Code changes

Fix handling of disabled extensions when they are explicitly set to `false`:

```json
{
  "@jupyterlab/mock-extension": false
}
```

## User-facing changes

Should fix issues with `disabledExtensions` when using `jupyterlab_server>=2.10`.

### DIFF
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -1152,6 +1152,8 @@ class _AppHandler(object):
         page_config = get_page_config(labextensions_path, app_settings_dir=app_settings_dir, logger=self.logger)
 
         disabled = page_config.get('disabledExtensions', {})
+        # handle disabledExtensions specified as a list (jupyterlab_server < 2.10)
+        # see https://github.com/jupyterlab/jupyterlab_server/pull/192 for more info
         if isinstance(disabled, list):
             disabled = { extension: True for extension in disabled }
 
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -602,8 +602,6 @@ class _AppHandler(object):
         # Do this last since it relies on other attributes
         self.info = self._get_app_info()
 
-
-
     def install_extension(self, extension, existing=None, pin=None):
         """Install an extension package into JupyterLab.
 
@@ -1153,7 +1151,11 @@ class _AppHandler(object):
 
         page_config = get_page_config(labextensions_path, app_settings_dir=app_settings_dir, logger=self.logger)
 
-        info['disabled'] = page_config.get('disabledExtensions', [])
+        disabled = page_config.get('disabledExtensions', {})
+        if isinstance(disabled, list):
+            disabled = { extension: True for extension in disabled }
+
+        info['disabled'] = disabled
 
         disabled_core = []
         for key in info['core_extensions']:
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -2238,7 +2238,7 @@ def _is_disabled(name, disabled={}):
     """Test whether the package is disabled.
     """
     for pattern, value in disabled.items():
-        # skip packages explicitely marked as disabled
+        # skip packages explicitly marked as not disabled
         if value == False:
             continue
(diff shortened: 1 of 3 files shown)


jtpio:
The temp commit (https://github.com/jupyterlab/jupyterlab/pull/11744/commits/3f44faf83933777727c046409dbfe115d6b92d33) pinning on `jupyterlab_server~=2.9.0` seems to be passing:

image

jtpio:
cc @vidartf who might have some insights, since this looks related to https://github.com/jupyterlab/jupyterlab_server/pull/192

jtpio:
According to https://github.com/jupyterlab/jupyterlab/pull/11744/commits/ed2317b8ff9a5dc8d3a4b5e19deaf55e1c55677d, this appeared to be a mishandling of `disabledExtensions`, which can take the form of:

```json
{
  "@jupyterlab/mock-extension": false
}
```

And until now the lab logic would only check for the key to be in the dictionary, without checking for the value that could be set to `false`.

jtpio:
Looks like this is actually a bug fix, as the following steps of enabling and disabling extensions would otherwise leave the extension disabled:

```
mamba create -n tmp python jupyterlab=3 ipywidgets=7.6 -y
jupyter labextension list
jupyter labextension disable @jupyter-widgets/jupyterlab-manager
jupyter labextension list
jupyter labextension enable @jupyter-widgets/jupyterlab-manager
jupyter labextension list
``` 

This is with `jupyterlab==3.2.5` and `jupyterlab_server==2.10.1`.

image

jupyterlab/commands.py
```diff
@@ -2230,10 +2230,13 @@ def noop(x, y, z):
     return return_value
 
 
-def _is_disabled(name, disabled=[]):
+def _is_disabled(name, disabled={}):
```
jtpio, line 2233: Need to double check whether a `list` or a `dict` is expected here, based on the change in https://github.com/jupyterlab/jupyterlab_server/pull/192

jtpio:
Converting back to a draft for now, as there are still a couple of things to double check based on the list to dict late conversion added in https://github.com/jupyterlab/jupyterlab_server/pull/192.

jtpio:
Updated the PR so it should now handle both `jupyterlab_server<2.10` and `jupyterlab_server>=2.10`

jupyterlab/commands.py
```diff
@@ -2230,10 +2234,13 @@ def noop(x, y, z):
...
     """Test whether the package is disabled.
     """
-    for pattern in disabled:
+    for pattern, value in disabled.items():
+        # skip packages explicitely marked as disabled
```
krassowski, line 2241: Should it be _not_ disabled?
```suggestion
        # skip packages explicitly marked as not disabled
```

jtpio: Yes, thanks for catching this!

jupyterlab/commands.py
fcollonval: [Minor comment] It is a bad practice to set a mutable as default value.

fcollonval (approved):
Thanks @jtpio

jtpio:
> [Minor comment] It is a bad practice to set a mutable as default value.

Agreed. Pushing a new commit now so we can improve that at the same time.

fcollonval:
@jtpio Is there some additional work/info missing or can we go ahead and merge this one to fix the CI?

jtpio:
Normally this is good to go now, since https://github.com/jupyterlab/jupyterlab/pull/11744/commits/21ba4264ef38141cb8a492ece526e403efdf3d10 should also handle  `jupyterlab_server<2.10`.

jtpio:
Merging as this fixes the CI and the use of disable extensions with `jupyterlab_server>=2.10`.

Let's check if other pending PRs can also be merged and we can make new `4.0.0ax` and `3.2.x` releases with this fix tomorrow for example.

jtpio:

PR #11751, opened 2021-12-27
Remove `Install`, `Update` and `Uninstall` buttons from the extension manager UI

jtpio:
## References

Incremental step towards https://github.com/jupyterlab/jupyterlab/issues/10554 and https://github.com/jupyterlab/jupyterlab/issues/11336.

To improve and unify the user experience around installing extensions, which is at the moment confusing.

And help reduce the number of "JupyterLab failed to build" issues as mentioned in the `<details>` section of https://github.com/jupyterlab/jupyterlab/issues/10554#issue-938314948.

## Code changes

This PR:

- removes the `Install`, `Update` and `Uninstall` buttons in the extension manager
- makes it possible to `Enable` and `Disable` prebuilt extensions

### Before

image

image

### After

image

image

None (UI only)

### DIFF
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -282,41 +282,6 @@ function ListEntry(props: ListEntry.IProperties): React.ReactElement<any> {
             {entry.description}
           </div>
           <div className="jp-extensionmanager-entry-buttons">
-            {!entry.installed &&
-              entry.pkg_type == 'source' &&
-              !entry.blockedExtensionsEntry &&
-              !(!entry.allowedExtensionsEntry && listMode === 'allow') &&
-              ListModel.isDisclaimed() && (
-                <Button
-                  onClick={() => props.performAction('install', entry)}
-                  minimal
-                  small
-                >
-                  {trans.__('Install')}
-                </Button>
-              )}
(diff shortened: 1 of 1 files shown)


jtpio:
If we decide to move forward with this for 4.0, we can then open follow-up PRs to remove more code handling the source extensions (installing, updating, uninstalling)

fcollonval:
Thanks @jtpio for starting this.


> removes the Install, Update and Uninstall buttons in the extension manager

I'm unsure we want to remove the buttons per se. The target is to provide an API that an extension could fulfill as extension provider. So I would suggest that the provider should only care about listing/installing/uninstalling extensions with the UI provided by the core extension.

> makes it possible to Enable and Disable prebuilt extensions

:heart:

jtpio:
> The target is to provide an API that an extension could fulfill as extension provider

That would indeed be the best long term goal. But since it's unclear if anyone is working on this and the extension manager for 4.0, this PR proposes to solve the more pressing issue of being able to install source extensions and confusing end users. Which seems to have also been a recurring topic at the weekly meetings (for example: https://github.com/jupyterlab/team-compass/issues/128#issuecomment-978091990)

fcollonval (approved):
Thanks for this one. Your argument makes sense.

PR #10298, opened 2021-05-26
Add pause on exception to debugger

goanpeca:
## References

https://github.com/jupyterlab/jupyterlab/issues/9544

## User-facing changes

jlab

IPykernel 6.0rc1

image

Issue #9544, opened 2020-03-05
Enter debug on exception


e-dorigatti:
This extension is gold, thank you!

I would like to propose a new feature, namely to enter debug when an uncaught exception occurs. This would allow the user to directly investigate the crime scene, check variables, and so on. It would really help development, especially when long-running computations are involved.

This would work just like when you run a script through `python -m ipdb -c c script.py`, and get a pdb prompt in case of troubles.

goanpeca:
Pinging @mlucool

krassowski:
Closing since #9544 was merged.

PR #11752, opened 2021-12-27
Pause on exception

andrewfulton9:
## References

This is a continuation of PR draft #10298.
It references issue #9544

## User-facing changes
chrome-capture (3)


packages/debugger-extension/src/index.ts
```diff
@@ -695,13 +696,34 @@ const main: JupyterFrontEndPlugin<void> = {
       }
     });
 
+    commands.addCommand(CommandIDs.pause, {
+      label: trans.__('Enable / Disable pausing on exceptions'),
```
fcollonval, line 700: You could improve it by using the status:
```suggestion
      label: service.isPausingOnExceptions ? trans.__('Disable pausing on exceptions') : trans.__('Enable pausing on exceptions'),
```

packages/debugger/src/panels/callstack/index.ts
```diff
@@ -81,6 +81,15 @@ export class Callstack extends PanelWithToolbar {
       })
     );
 
+    this.toolbar.addItem(
```
fcollonval, line 84: I would prefer the button to appear on the breakpoints panel not on the callstack one.

packages/debugger/src/panels/callstack/index.ts
```diff
@@ -128,6 +137,11 @@ export namespace Callstack {
      * The evaluate command ID.
      */
     evaluate: string;
+
+    /**
+     * The puse command ID.
```
fcollonval, line 142:
```suggestion
     * The pause command ID.
```

packages/debugger-extension/src/index.ts
```diff
@@ -695,13 +696,34 @@ const main: JupyterFrontEndPlugin<void> = {
...
+      caption: trans.__('Enable / Disable pausing on exceptions'),
+      className: 'jp-PauseOnExceptions',
+      icon: Debugger.Icons.pauseOnExceptionsIcon,
+      isToggled: () => {
+        return service.isPausingOnExceptions;
+      },
+      isEnabled: () => {
+        return !!service.isStarted;
```
fcollonval, line 708: What about kernel not supporting that type of exceptions?

packages/debugger/src/service.ts
```diff
@@ -498,6 +513,34 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+  async pauseOnExceptions(enable: boolean): Promise<void> {
+    if (!this.session?.isStarted) {
+      return;
+    }
+    const exceptionBreakpointFilters = this.session.exceptionBreakpointFilters;
+    this.session.pausingOnExceptions = enable;
+    const filters: string[] = [];
+    const exceptionOptions: DebugProtocol.ExceptionOptions[] = [];
+    const breakMode = enable ? 'userUnhandled' : 'never';
+    for (let filter_dict of exceptionBreakpointFilters ?? []) {
```
fcollonval, line 530: `filter_dict` is not a valid identifier. You should use camel case.

packages/debugger/src/session.ts
```diff
@@ -80,12 +82,42 @@ export class DebuggerSession implements IDebugger.ISession {
...
+  /**
+   * Whether to pause on exceptions
+   */
+  get pausingOnExceptions(): boolean {
+    return this._pausingOnExceptions;
+  }
+
+  /**
+   * Sets pause on exception
+   */
```
fcollonval, line 100: When you define a getter and a setter only one doc string is required.

```suggestion
```

packages/debugger/src/tokens.ts
```diff
@@ -46,11 +51,19 @@ export interface IDebugger {
    */
   readonly sessionChanged: ISignal<IDebugger, IDebugger.ISession | null>;
 
+  /**
+   * TODO: Signal emitted upon session changed.
```
fcollonval, line 55: `TODO` to resolve

packages/debugger/src/tokens.ts
```diff
@@ -46,11 +51,19 @@ export interface IDebugger {
...
+   */
+  readonly stateRestored: ISignal<IDebugger, null>;
+
   /**
    * Removes all the breakpoints from the current notebook or console
    */
   clearBreakpoints(): Promise<void>;
 
+  /** TODO */
```
fcollonval, line 64: `TODO` to resolve

packages/debugger/style/icons.css
```diff
@@ -25,3 +25,13 @@ button.jp-Button.jp-mod-minimal.jp-TableView:hover {
 .jp-ViewModeSelected {
   background-color: var(--jp-layout-color2);
 }
+
+/* Pause on exceptions */
+button.jp-Button.bp3-button.bp3-minimal.jp-PauseOnExceptions:hover {
```
fcollonval, line 30: Blueprint has been removed

```suggestion
button.jp-Button.jp-mod-minimal.jp-PauseOnExceptions:hover {
```

packages/debugger-extension/src/index.ts
```diff
@@ -695,13 +696,34 @@ const main: JupyterFrontEndPlugin<void> = {
...
+    });
+
     service.eventMessage.connect((_, event): void => {
       commands.notifyCommandChanged();
       if (labShell && event.event === 'initialized') {
         labShell.activateById(sidebar.id);
       }
     });
 
+    service.stateRestored.connect(_ => {
```
fcollonval, line 723: (Question due to my lack of knowledge)

Is this needed? Is the connector below not doing the job already?

fcollonval (changes requested):

packages/debugger-extension/src/index.ts
andrewfulton9: You are right. I'll get that taken out.

packages/debugger-extension/src/index.ts
andrewfulton9: At worst, I think the kernel would just run until it hit the the exception instead of breaking immediately before. That being said, pauseOnExceptions sets the breakpoint argument options via the debug protocol so my assumption is that it should be kernel agnostic assuming their is a debug protocol for it. I have only tried running it with the ipython and xeus python kernels though.

fcollonval: I understand but the debug protocol does specify an exchange of capabilities at start-up. Example for ipykernel (server reply):

image

So it seems, we do have a way to determine if the kernel supports it.

andrewfulton9: Those are currently used to build the options object that gets sent to debugger. So if none are defined for a kernel then pause on exception would never pause. Otherwise it should stop on those breakpoint filters.

fcollonval: From the user point of view, this means the button will not do anything what should be avoided. This is why it will be nice if the `isEnabled` or perhaps the `isVisible` attribute of the command will be set accordingly.

andrewfulton9:
@fcollonval, I think I have addressed everything from your previous review, and fixed an issue with the pause button not updating when switching tabs. Could you give it another review when you get a chance? Thank you!

fcollonval (changes requested):
This looks great @andrewfulton9 (tested on Binder)

I have two minor comments and it should be good to go.

andrewfulton9:
@fcollonval, I added the docstring. Let me know if there is anything else I should do. I think this should be good to go otherwise. I rebased on master. I think thats where all the auto tags are coming from. I'm not sure if they are supposed to be added.

fcollonval:
@andrewfulton9 would you mind rebasing this PR? It looks like something went wrong hence the amount of changes (and all those tags).

andrewfulton9:
@fcollonval, looks like re-basing again did the trick. I'm not sure what happened on the first one. Thanks for the suggestion!

packages/debugger/package.json
```diff
@@ -73,7 +73,8 @@
     "@lumino/widgets": "^1.30.0",
     "@vscode/debugprotocol": "^1.51.0",
     "codemirror": "~5.61.0",
-    "react": "^17.0.1"
+    "react": "^17.0.1",
+    "vscode-debugprotocol": "~1.51.0"
```
fcollonval, line 77: This probably slips in following the rebase. `vscode-debugprotocol` has been renamed `@vscode/debugprotocol` and [that change](https://github.com/jupyterlab/jupyterlab/pull/11812) is already on master.

```suggestion
    "react": "^17.0.1"
```

packages/debugger/src/session.ts
```diff
@@ -9,6 +9,8 @@ import { PromiseDelegate } from '@lumino/coreutils';
 
 import { ISignal, Signal } from '@lumino/signaling';
 
+import { DebugProtocol } from 'vscode-debugprotocol';
```
fcollonval, line 12:
```suggestion
import { DebugProtocol } from '@vscode/debugprotocol';
```

packages/debugger/src/service.ts
```diff
@@ -859,6 +950,7 @@ export class DebuggerService implements IDebugger, IDisposable {
   private _sessionChanged = new Signal<IDebugger, IDebugger.ISession | null>(
     this
   );
+  private _stateRestored = new Signal<IDebugger, null>(this);
```
fcollonval, line 953: It looks like `stateRestored` signal is not read in the final version. So let's get ride of it.

```suggestion
```

fcollonval (changes requested):
@andrewfulton9 sorry two more changes are required. Following the rebase `vscode-debugprotocol` package is now named `@vscode/debugprotocol`. And it looks like we can remove `stateRestored` signal as it is not read anywhere.

andrewfulton9:
@fcollonval, thanks for catching those! I committed all your suggestions.

fcollonval:
@andrewfulton9 the *Linux Tests / Linux (release_test, 3.10) (pull_request)* is legit. The yarn.lock needs to be updated (most likely to remove the old `vscode-debugprotocol` package). Could you please update the PR?

andrewfulton9:
@fcollonval, thanks for pointing me in the right direction. I got that fixed. That last failed isn't something to be fixed in this PR, right?

fcollonval (approved):
Thanks @andrewfulton9 for pushing this through

jtpio:
Thanks!

CI is now all green.

jtpio:

andrewfulton9:
@jtpio, Yeah I'll try to do that today

krassowski:
I updated the top-level comment with recent screenshots (as the button was moved to a different location). Was there a discussion about having the caption differ from label?

```ts
       label: service.isPausingOnExceptions
        ? trans.__('Disable pausing on exceptions')
        : trans.__('Enable pausing on exceptions'),
      caption: trans.__('Enable / Disable pausing on exceptions'),
```

It would be easier for me to know if the option is on/off if it was separated.

fcollonval:
> It would be easier for me to know if the option is on/off if it was separated.

Good point I missed it when reviewing. I opened #11970

PR #11761, opened 2021-12-29
Add text on how to run it in a dir other than home

TheOtherRealm:
## References
https://stackoverflow.com/questions/46755604/how-to-navigate-to-a-different-directory-in-jupyter-notebook/46762386

### DIFF
--- a/docs/source/getting_started/starting.rst
+++ b/docs/source/getting_started/starting.rst
@@ -11,7 +11,7 @@ Start JupyterLab using:
 
 JupyterLab will open automatically in your browser.
 
-If the location of your notebook files are not in your home directory (more often the case in Windows, or if you have an external drive) you can use the `--notebook-dir` added parameter to start JupyterLab in that directory.  Be very careful though, never run it in your root volume (ex: `c/` in Windows, `/` in Linux) unless you know what you are doing and are prepared for the consequences! 
+If your notebook files are not in the current directory, you can pass your working directory path as argument when starting JupyterLab. Avoid running it from your root volume (e.g. `C:\` on Windows or `/` on Linux) to limit the risk of modifying system files.
 
 .. code:: bash
 
--- a/docs/source/getting_started/starting.rst
+++ b/docs/source/getting_started/starting.rst
@@ -11,6 +11,15 @@ Start JupyterLab using:
 
 JupyterLab will open automatically in your browser.
 
+If the location of your notebook files are not in your home directory (more often the case in Windows, or if you have an external drive) you can use the `--notebook-dir` added parameter to start JupyterLab in that directory.  Be very careful though, never run it in your root volume (ex: `c/` in Windows, `/` in Linux) unless you know what you are doing and are prepared for the consequences! 
+
+.. code:: bash
+
+    #Windows Example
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/starting.rst
```diff
@@ -11,6 +11,15 @@ Start JupyterLab using:
...
 JupyterLab will open automatically in your browser.
 
+If the location of your notebook files are not in your home directory (more often the case in Windows, or if you have an external drive) you can use the `--notebook-dir` added parameter to start JupyterLab in that directory.  Be very careful though, never run it in your root volume (ex: `c/` in Windows, `/` in Linux) unless you know what you are doing and are prepared for the consequences! 
+
+.. code:: bash
+
+    #Windows Example
+    jupyter lab --notebook-dir E:/
+    #Linux Example
+    jupyter lab --notebook-dir /var/www/html/
```
fcollonval, line 21: Actually there is a shortcut for that option:

```suggestion
    jupyter lab <path/to/working/directory>
```

docs/source/getting_started/starting.rst
```diff
@@ -11,6 +11,15 @@ Start JupyterLab using:
 
 JupyterLab will open automatically in your browser.
 
+If the location of your notebook files are not in your home directory (more often the case in Windows, or if you have an external drive) you can use the `--notebook-dir` added parameter to start JupyterLab in that directory.  Be very careful though, never run it in your root volume (ex: `c/` in Windows, `/` in Linux) unless you know what you are doing and are prepared for the consequences!
```
fcollonval, line 14: The default command opens the current folder. So I would rephrase the sentence (+ taken into account the next comment).

```suggestion
If your notebook files are not in the current directory, you can pass your working directory path as argument when starting JupyterLab. Avoid running it from your root volume (e.g. `C:\` on Windows or `/` on Linux) to limit the risk of modifying system files.
```

fcollonval (changes requested):
Thanks for the suggestion @TheOtherRealm 

I left some comments.

docs/source/getting_started/starting.rst
krassowski: Also, `--notebook-dir` is deprecated, but `--root-dir` and `--preferred-dir` might be worth mentioning. Combining the two allows to set root to e.g. `/home/user/projects/` but open the browser or `/home/user/projects/current_project/`.

TheOtherRealm: `--root-dir' and `--preferred-dir` do not seem to work but `--app_dir` and `--preferred_dir` do. I'm running v3.2.1 .

TheOtherRealm: I updated the pending changes to reflect the new parameters

fcollonval (approved):
Thanks @TheOtherRealm

jtpio:
Looks like this PR is targeting the `3.2.x` branch?

Maybe we can update it to target `master`, and then backport to `3.2.x`.

jtpio:
> Maybe we can update it to target master, and then backport to 3.2.x.

Let's just backport back to `master` when merged.

Thanks!

jtpio:

Issue #11467, opened 2021-11-15
Issues with breakpoints and debugger


subhamagrawal7:
Don't know if this is related, but it seemed quite close to #10090 that I came across these 2 observations. Let me know if these are expected or not?? -
1) Adding a break point to an empty cell adds it to other cells as well but does not show up in the debugger. Like adding cells after adding the first cell and the breakpoint does not show the other cells breakpoints
![breakpoints_empty](https://user-images.githubusercontent.com/34346812/141802268-2c75bc93-60ca-4787-a856-3e840356ff0e.gif)

2) Delete Cells or Cut Cells, neither of the 2 options removes the breakpoints [this is even after the cell was executed]
**With delete cells -** 
Actions - 
    1) Create the cell with reqd. function
    2) Add a breakpoint
    3) Execute the cell
    4) It stops at breakpoint, continue by pressing Play button besides CallStack
    5) Goto Edit -> Delete Cells

![delete_cells_breakpoint](https://user-images.githubusercontent.com/34346812/141803362-2ebf41b5-fb4f-4963-9598-19135a842501.gif)

  **With cut cells -**
  Actions - 
      1) Create the cell with reqd. function
      2) Add a breakpoint
      3) Execute the cell
      4) It stops at breakpoint, continue by pressing Play button besides CallStack
      5) Click on Scissor Icon from toolbar

![cut_cells_breakpoint](https://user-images.githubusercontent.com/34346812/141806271-38ebeb42-76a6-4076-b49f-f26d308d9d8d.gif)

Thanks,
Subham

fcollonval:
> Adding a break point to an empty cell adds it to other cells as well but does not show up in the debugger.

One of the trouble with the current approach to create cell dump is the use of a content hash. This implies that two cells with the same content will have two identical hashes and breakpoints will be set on all identical cells. However since notebook format version 4.5, cells have an unique id. It will be good to extend the dump cell filename to include it to deal with uniqueness mapping between cells and breakpoint definitions.

cc @JohanMabille @afshin @jtpio to get their inputs as they have a better background on the debugger design.

> Delete and cut cells

I will open a PR to address the two other cases that should be easier to fix as they don't need changes outside JupyterLab.

PR #11764, opened 2021-12-30
Clear breakpoints when disposing an editor

fcollonval:
## References

Partialy address #11467

## Code changes

When disposing an editor handler, the associated breakpoints are cleared.

Minor changes:
- Fix usage of translation
- Use non-nullish operator instead of or operator

## User-facing changes

This clear the breakpoints when a cell is cut or deleted.

N/A

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -716,7 +716,7 @@ const main: JupyterFrontEndPlugin<void> = {
     shell.add(sidebar, 'right');
 
     commands.addCommand(CommandIDs.showPanel, {
-      label: translator.load('jupyterlab').__('Debugger Panel'),
+      label: trans.__('Debugger Panel'),
       execute: () => {
         shell.activateById(sidebar.id);
       }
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @fcollonval for looking into this.

Wondering if this is a behavior we want for the debugger? I believe it used to be possible to undo a cell deletion are retrieve the previous breakpoints automatically, which was considered convenient. When testing on the Binder for this PR this does not seem to be the case anymore?

https://user-images.githubusercontent.com/591645/147765294-b620d682-4961-4300-a479-23a4483e4823.mp4

jtpio:
> I believe it used to be possible to undo a cell deletion are retrieve the previous breakpoints automatically

Testing on `3.2.5` with this gist: https://gist.github.com/jtpio/6ce26381703355e0ef1da4af742b7f72

Shows that it was indeed the case:

https://user-images.githubusercontent.com/591645/147765460-0ff0e7ed-cc4e-4a71-a19e-e6058049a08b.mp4

jtpio:
Should this be bumped to `3.3.x`?

fcollonval:
Closing as the current approach is conflicting with another feature

PR #11770, opened 2021-12-30
Fix integrity failure on CI

jtpio:
## References

Investigate the integrity failure that as seen on recent PRs but not locally.

## Code changes

Make the `getPythonVersion` function more robust when parsing the output of `python setup.py --version`.

## User-facing changes

None

None

### DIFF
--- a/buildutils/src/utils.ts
+++ b/buildutils/src/utils.ts
@@ -75,10 +75,6 @@ export function writePackageData(
 ): boolean {
   const text = JSON.stringify(sortPackageJson(data), null, 2) + '\n';
   const orig = fs.readFileSync(pkgJsonPath, 'utf8').split('\r\n').join('\n');
-  console.log('NEW');
-  console.log(text);
-  console.log('ORIG');
-  console.log(orig);
   if (text !== orig) {
     fs.writeFileSync(pkgJsonPath, text, 'utf8');
     return true;
--- a/buildutils/src/utils.ts
+++ b/buildutils/src/utils.ts
@@ -75,6 +75,10 @@ export function writePackageData(
 ): boolean {
   const text = JSON.stringify(sortPackageJson(data), null, 2) + '\n';
   const orig = fs.readFileSync(pkgJsonPath, 'utf8').split('\r\n').join('\n');
+  console.log('NEW');
+  console.log(text);
+  console.log('ORIG');
+  console.log(orig);
   if (text !== orig) {
     fs.writeFileSync(pkgJsonPath, text, 'utf8');
     return true;
--- a/buildutils/src/utils.ts
+++ b/buildutils/src/utils.ts
@@ -191,7 +191,8 @@ export function checkStatus(cmd: string): number | null {
  */
 export function getPythonVersion(): string {
   const cmd = 'python setup.py --version';
-  return run(cmd, { stdio: 'pipe' }, true);
+  const lines = run(cmd, { stdio: 'pipe' }, true).split('\n');
+  return lines[lines.length - 1];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


jtpio:
Looks like the `ensure-repo` script tries to write a new `dev_mode/package.json` with the following in `jupyterlab.version`:

image

jtpio:
Turns out this check is currently not hit on CI: 

https://github.com/jupyterlab/jupyterlab/blob/602b05399b0ca762613c8f560a49b15abdefee39/scripts/ci_install.sh#L8-L13

Fixing it so with `linux-gpu` (https://github.com/jupyterlab/jupyterlab/pull/11770/commits/04b031dd91e3800b6e6342949cd52a3bf08f82b1) makes the `release_test` check fail, for example: https://github.com/jupyterlab/jupyterlab/runs/4668362330?check_suite_focus=true

image

Something we should track and investigate separately with this commit: https://github.com/jupyterlab/jupyterlab/pull/11770/commits/04b031dd91e3800b6e6342949cd52a3bf08f82b1

jtpio:
Looks like CI is now green.

buildutils/src/utils.ts
```diff
@@ -187,7 +187,8 @@ export function checkStatus(cmd: string): number | null {
  */
 export function getPythonVersion(): string {
   const cmd = 'python setup.py --version';
-  return run(cmd, { stdio: 'pipe' }, true);
+  const lines = run(cmd, { stdio: 'pipe' }, true).split('\n');
+  return lines[lines.length - 1];
```
jtpio, line 191: Not sure what changed compared to the commit that passed on `master` yesterday with this commit: https://github.com/jupyterlab/jupyterlab/commit/602b05399b0ca762613c8f560a49b15abdefee39

But regardless this should be a better way to handle the output of the command, since we are only interested in the last line. Similar to this change in `jupyter_releaser`: https://github.com/jupyter-server/jupyter_releaser/pull/37

jtpio:
> Looks like the ensure-repo script tries to write a new dev_mode/package.json with the following in jupyterlab.version:

We should probably investigate why we now get this on CI, but not locally.

The `Yarn configuration loaded.` message comes from this place in the JupyterLab code base:

https://github.com/jupyterlab/jupyterlab/blob/602b05399b0ca762613c8f560a49b15abdefee39/jupyterlab/commands.py#L1985

So probably something is now triggering that code path, and the debug message ends up in the output of `python setup.py --version`.

jtpio:
The `release_test` check failure from another PR (https://github.com/jupyterlab/jupyterlab/pull/11771) seems to be surfacing some `distutils` issue: https://github.com/jupyterlab/jupyterlab/runs/4668341149?check_suite_focus=true

image

Which might be related to the issue being investigated here.

jtpio:
The latest `setuptools` release from yesterday might be relevant:

image

jtpio:
> The latest `setuptools` release from yesterday might be relevant:

Yes, upgrading to the latest locally with `pip install setuptools --U` makes it possible to reproduce the issue:

https://user-images.githubusercontent.com/591645/147791327-a02cec2a-1e20-47ce-bb83-dcf9b9ebc631.mp4

Converting back to draft to investigate more.

jtpio:
This seems to be the related change in the latest (`60.2.0`) release of `setuptools`: https://github.com/pypa/setuptools/pull/2974

jtpio:
Since the `python setup.py --version` and `python setup.py --name` commands are currently used in several places, for example:

https://github.com/jupyterlab/jupyterlab/blob/602b05399b0ca762613c8f560a49b15abdefee39/jupyterlab/federated_labextensions.py#L384

https://github.com/jupyterlab/jupyterlab/pull/11770/commits/dbef4fa88bd0f2cf27564162f37eac9df52b4306 delays the import of `LabApp`, so the default traits are not created which prevents the log messages from appearing.

fcollonval (approved):
Thanks a lot @jtpio for fixing this one.

fcollonval:

PR #11785, opened 2022-01-05
Fix overlapped shadow for scrolling output cell

### LINKED ISSUE #11583: Make scrolled output shadow visible at the bottom
The current behavior matches the one of the classical notebook (aka content is below top shadow but above bottom shadow - see https://github.com/berkeley-dsep-infra/datahub/issues/2993). Should the content be below the bottom shadow too?

image

I personnally in favor of applying the same trick to have the content below the bottom shadow too as it should emphasize the content is enclosed within a scrollable container.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab/issues/11478#issuecomment-973830299_

thesinepainter:
## References

Fixes #11583

## Code changes

Moved `box-shadow:` property from `.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea` to it's own `::after` pseudo class so the shadow is displayed after the conent.

```css
.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after {
  content: ' ';
  box-shadow: inset 0 0 6px 2px rgb(0 0 0 / 30%);
  width: 100%;
  height: 100%;
  position: sticky;
  bottom: 0;
  top: 0;
  margin-top: -50%;
  float: left;
  display: block;
  pointer-events: none;
}
```
Moved `padding-top:` property to selector `.jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-child` to fix shadow's height alignment.

```css
.jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-child {
  padding-top: 6px;
}
```

## User-facing changes

The scrolling output's content overlapped the inset shadow of the scrolling output box due to the contents being drawn above the shadow.

To address this, the shadow was made to be drawn after the content using the `::after` pseudo class for the selector 
`.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea`.

This however caused a height misalignment of the inset shadow such that only either top or bottom of the shadow was visible simultaneously. To remedy this, the `padding-top:` property was moved to another selector 
`.jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-child`

The `::after` selector also causes the inset shadow to be drawn to next to the scroll bar instead of over it.

Before: Content Overlapping Shadows
scroll shadow 1

scroll shadow 2

After: Shadows Overlapping Content
scroll shadow 3

scroll shadow 4


krassowski:
Thank you for starting work on this one! Could we give a try to a different method, which would actually move the inset over all elements, regardless of their type (table, image, or something else)? One idea would be to use `::after` pseudo-element, roughly like:

```css
.lm-Widget.p-Widget.jp-OutputArea.jp-Cell-outputArea::after {
  content: ' ';
  box-shadow: inset 0 0 6px 2px rgb(0 0 0 / 30%);
  width: 100%;
  height: 100%;
  position: sticky;
  top: 0;
  margin-top: -50%; /* account for flow adjustment */
  float: left; /* remove from DOM flow */
  display: block;
  pointer-events: none; /* ignore user interactions */
}
```

There will be more adjustments necessary (moving `padding-top` one level lower, maybe adding some `position: relative`).

thesinepainter:
Thanks a lot for the info! Sure! Initially, I'd tried to do that as well. 
I used `::after` on  `.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea` but I needed to set the ```position:absolute``` for the shadow to show up, which caused the shadow to stay in one place inside the scrolled output area.

scroll shadow fix 4

This is the class:
```css
.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after{
  content: ' ';
  box-shadow: inset 0 0 6px 2px rgba(0, 0, 0, 0.3);
  left: 0;
  top: 0;
  width: 100%;
  height: 100%;
  position: absolute;
  pointer-events: none;
}
```

Then I'd tried using the above for `.jp-Cell:not(.jp-mod-noOutputs) .jp-Cell-outputWrapper::after` as well but the scrolling shadow is displayed even if scrolling output is disabled. 

scroll shadow fix 5

I'm getting the same problem with your suggestion too, the shadow's displayed even when scrolling is disabled.

Would you know how to get the shadow to appear when scrolled outputs is enabled?

krassowski:
Sorry, I missed one important line in my example: `bottom: 0`, that should make it work with `position: sticky`. I am not sure what you mean by shadow still showing up, could you commit the code so we can take a look at Binder?

thesinepainter:
Sorry for being unclear. What i meant was that, when disable scrolling for outputs is selected the usual behavior is this flat display.

scroll shadow fix 6

Note that there is no shadow over the output cell. However, using 
`.lm-Widget.p-Widget.jp-OutputArea.jp-Cell-outputArea::after` I get this.

scroll shadow fix 5

The shadow is still displayed even though scrolling outputs is disabled.

krassowski:
Right I was suggesting wrong selector, sorry. I think that `.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after` should work well.

packages/cells/style/widget.css
```diff
@@ -71,11 +71,51 @@
 .jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea {
   overflow-y: auto;
   max-height: 24em;
-  box-shadow: inset 0 0 6px 2px rgba(0, 0, 0, 0.3);
+  /*box-shadow: inset 0 0 6px 2px rgba(0, 0, 0, 0.3);*/
   margin-left: var(--jp-private-cell-scrolling-output-offset);
   padding-top: 6px;
 }
```
krassowski, line 77: let's try to move the padding to the child

```suggestion
}

.jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-child {
  padding-top: 6px;
}
```

packages/cells/style/widget.css
```diff
@@ -71,11 +71,51 @@
...
+  box-shadow: inset 0 0 6px 2px rgba(0, 0, 0, 0.3);
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  pointer-events: none;
+}*/
+
+.lm-Widget.p-Widget.jp-OutputArea.jp-Cell-outputArea::after {
```
krassowski, line 103:
```suggestion
.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after {
```

krassowski:
If after the suggested change it all works, let's clean it up and maybe revert the changes to the row/even odds to keep it simple?

thesinepainter:
Yeah, will do! Thanks for the help.

packages/cells/style/widget.css
```diff
@@ -71,8 +71,21 @@
...
+  width: 100%;
+  height: 98%;
+  position: sticky;
+  bottom: 0;
+  top: 0;
+  margin-top: -50%;
+  float: left;
+  display: block;
+  pointer-events: none;
   padding-top: 6px;
```
krassowski, line 89:
```suggestion
  height: 100%;
  position: sticky;
  bottom: 0;
  top: 0;
  margin-top: -50%;
  float: left;
  display: block;
  pointer-events: none;
```

thesinepainter:
Using the below pseudo class resolved the issue and allows the inset shadow to now appear over all content in scrolling output. 

```css
.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after {
  content: ' ';
  box-shadow: inset 0 0 6px 2px rgb(0 0 0 / 30%);
  width: 100%;
  height: 98%;
  position: sticky;
  bottom: 0;
  top: 0;
  margin-top: -50%;
  float: left;
  display: block;
  pointer-events: none;
  padding-top: 6px;
}
```

Also cleaned up the widget.css as well as reverted all changes from the first solution. Thanks again @krassowski !

packages/cells/style/widget.css
```diff
@@ -71,8 +71,21 @@
 .jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea {
   overflow-y: auto;
   max-height: 24em;
-  box-shadow: inset 0 0 6px 2px rgba(0, 0, 0, 0.3);
   margin-left: var(--jp-private-cell-scrolling-output-offset);
+}
```
krassowski, line 75:
```suggestion
}

.jp-CodeCell.jp-mod-outputsScrolled .jp-OutputArea-child {
  padding-top: 6px;
}
```

krassowski:
Thanks! I left a small suggestion which comes down to moving the `padding-top` to another selector which should allow to set height back to `100%`.

thesinepainter:
Done!

krassowski (approved):
Thank you! Works well in Chrome and Firefox.

Would you mind updating the description in the top level comment, please?

jtpio:
Thanks!

This also looks good with Figures when testing on Binder, for example:

image

jtpio:
Looks like there is small side-effect with this change related to the position of the scrollbar:

**Before**

image

**With this PR**

image

I think this is fine and the new version actually looks nicer overall?

jtpio (approved):

krassowski:
Good catch. We _could_ force it by setting `width` > 100% but I am not sure if we _should_.

thesinepainter:
Thanks! @jtpio and @krassowski  Re-did the top comment, detailing all the updated changes and their effects. I like the change to the scroll bar's shadow as well.

jtpio:

Issue #11602, opened 2021-12-01
Remove document's content from ModelDB


hbcarlos:
Follows #11434

> I'll be updating this comment until we find the definitive API

## Problem
Currently, we have the document's content triplicated in the editor, ModelDB and YJS. I.e., the same content is mirrored in different data structures. This means that every time one of the models changes, we must update all other models to keep them in sync. This approach is very complex and prone to bugs. As commented several times, we need to remove the content from ModelDB and use only the shared models to manipulate it. This will allow us to simplify the existing code. For extension authors, this means that the API will change.

## Proposed Solution
The idea is to open multiple PRs to remove attributes from ModelDB and keep them only in the shared models. Some of the changes are:

#### Remove document's content:
* More information in its corresponding PR #11601

#### Remove document's metadata:
* PR: none yet
* API changes:
    * Document's metadata are stored in ModelDB as an `IObservableJSON` named `metadata`. We want to remove this attribute from ModelDB to use its equivalent from the shared model:
        * Notebook:
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/notebook/src/model.ts#L108
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/shared-models/src/ymodels.ts#L472
        * Cells:
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/cells/src/model.ts#L185-L187
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/shared-models/src/ymodels.ts#L625
    * Regarding the models the change would be `readonly metadata: IObservableJSON` -> `readonly metadata: nbformat.INotebookMetadata`
        * Notebook:
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/notebook/src/model.ts#L65
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/notebook/src/model.ts#L165-L167
        * Cells:
https://github.com/jupyterlab/jupyterlab/blob/014091f21d00bd1c940105744d945a61a7b23e6a/packages/cells/src/model.ts#L65
[...]

Issue #11434, opened 2021-11-10
Next steps on RTC


hbcarlos:
I'm opening this issue to keep track of the next steps we plan for real-time collaboration in JupyterLab 4.0.

### Deduplicate content
Currently, we have the document's content triplicated in the view, ModelDB and YJS. I.e., the same content is mirrored in different data structures. This means that every time one of the models changes, we must update all other models to keep them in sync. This approach is very complex and prone to bugs. As commented several times, we need to remove ModelDB and use only the shared models to manipulate content. This will allow us to simplify the existing code. For extension authors, this means that they will no longer have access to ModelDB. Instead, they should use the new shared-models API, which provides better guarantees. The switch should be fairly straight-forward.
* jupyterlab/jupyterlab#11602 

### Add tests
* To the shared models
* To the username: https://github.com/jupyterlab/jupyterlab/pull/11852#issuecomment-1011994165

### Changes on the Shared-models
* Undo manager by cell jupyterlab/jupyterlab#10791
* jupyterlab/jupyterlab#11640

### ICurrentUser
* jupyterlab/jupyterlab#10965
* jupyterlab/jupyterlab#11443
* jupyterlab/jupyterlab#11678

### UI features
* List of collaborators: jupyterlab/jupyterlab#11528
* Current focused cell: jupyterlab/jupyter_collaboration#109
* Rename anonymous users

### Add RBAC
* jupyterlab/jupyterlab#11355

### Use the python port of YJS on the backend
The locking approach used now does not work well when there are many collaborators, we could improve it to make it work decently, but it will always be prone to fail.

The best approach is to adopt the new python bindings for YJS on the backend. This implies removing the content REST API and accessing the document's content through YJS. At the same time, we would need to remove the save button/command since the document's content would always be synced to disk; a la google docs. Using YJS on the backend has some other benefits that could be tackled later by extension, like using YJS versions to see the list of changes between versions; a la git. 
* jupyterlab/jupyterlab#11599

In addition, it allows to tackle the following issues:
* jupyterlab/jupyterlab#2833
* jupyterlab/jupyterlab#9621
* jupyterlab/jupyterlab#10544

cc @dmonad

martinRenou:
> At the same time, we would need to remove the save button/command since the document's content would always be synced to disk

Maybe JupyterLab could always work on file copies and only overwrite the original on save?

hbcarlos:
I updated the first comment with new PRs and issues.

[...]

PR #11790, opened 2022-01-05
Remove metadata from modelDB

hbcarlos:
This PR depends on #11601 

## References
* #11434
* #11602


jtpio:
Thanks @hbcarlos :+1: 

> This PR depends on https://github.com/jupyterlab/jupyterlab/pull/11601

How do you see we could move this forward? Should we try to look into https://github.com/jupyterlab/jupyterlab/pull/11601 first so it can be built upon?

PR #10523, opened 2021-07-01
[WIP] Update completer API

hbcarlos:
This PR is a draft to discuss and prototype a new completer API for JupyterLab 4.0.

## References
* Discussion issue: #9763
* This work is based on: https://github.com/krassowski/jupyterlab-lsp/pull/600
* Previous work:
    * #10348
    * #8080
    * #8140
    * #8141

### Major changes:
* Add a new class `DefaultProvider` with its interface `ICompletionProvider` which is a refactor of `CompletionConnector` to enable extensions to add diferent implementations.
* Support for multiple providers.

### Allow multiple providers:
To allow extensions to add new providers that request completions from different sources, we need to make it asynchronous to cause sure providers don't block each other. We need to discuss the best way to merge completion items from multiple sources.

Examples of providers:
* Completions from kernel
* Completions from language server
* https://github.com/krassowski/jupyterlab-lsp/issues/208

### Language server protocol:
With the actual implementation, completions come from kernels, and they decide how much information to send. Some extensions would like to add a provider that requests completions from a language server and choose what information to add and how to add it (for example, lazy loading of documentation for better performance). The problem is that the completer has no idea what is the language of the document. In the case of notebooks, we can obtain the information from the kernel. Still, the file editor doesn't have a kernel, there are other ways of getting the language, but there are some edge cases where it is difficult or is not assured.

There is already a proposal for a [Language Server Protocol adoption](https://github.com/jupyter/enhancement-proposals/pull/72).

cc @krassowski

PR #11795, opened 2022-01-06
Refactor completer and completer-extension package

trungleduc:
## References

This PR is a continuation of  https://github.com/jupyterlab/jupyterlab/pull/10523 with a lot of inspiration from https://github.com/jupyter-lsp/jupyterlab-lsp/pull/600 to prototype a new completer API for JupyterLab 4.0. 

These breaking changes are my experiment to facilitate the integration of external completer provides, since I do not have many experiences with other completers (`
jupyterlab-lsp`, `tabnine`, `kite`....), all points here need further discussions! 

### Replace `Connector` with a new provider interface.
The old `Connector` concept is removed (motived by https://github.com/jupyterlab/jupyterlab/pull/11795#issuecomment-1011350045 ) and replaced by a more flexible provider concept. To create a completer provider for JupyterLab, users need to implement the interface `ICompletionProvider`
```typescript
export interface ICompletionProvider<
  T extends CompletionHandler.ICompletionItem = CompletionHandler.ICompletionItem
> {
  /**
   * Unique identifier of the provider
   */
  identifier: string;

  /**
   * Is completion provider applicable to specified context?
   * @param request - the completion request text and details
   */
  isApplicable(context: ICompletionContext): Promise<boolean>;

  /**
   * Fetch completion requests.
   *
   * @param request - the completion request text and details
   */
  fetch(
    context: ICompletionContext
  ): Promise<CompletionHandler.ICompletionItemsReply<T>>;

  /**
   * Renderer for provider's completion (optional).
   */
  renderer: Completer.IRenderer | null | undefined;
}
```

- `isApplicable` is used to check if the provider can be used for the current completion context.
- `fetch` will be used by the internal handler to get completer results. Provider authors are free to decide the additional items in the result.
- if `renderer` is not defined, the default renderer of jupyterlab will be used.

### New default providers
The old default completer which combines the results from context and kernel completer is now split into two independent provides `ContextCompleterProvider` and `KernelCompleterProvider`

### New manager to register providers
A token `ICompletionProviderManager` is created to help users register new providers implemented with `ICompletionProvider`. The current JupyterLab completer providers are registered as follow:

```typescript
const defaultProvider: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/completer-extension:base-service',
  requires: [ICompletionProviderManager],
  autoStart: true,
  activate: async (
    app: JupyterFrontEnd,
    serviceManager: ICompletionProviderManager
  ): Promise<void> => {
    serviceManager.registerProvider(new ContextCompleterProvider());
    serviceManager.registerProvider(new KernelCompleterProvider());
  }
};
```

One advantage of the new system is that completer provider authors do not need to care about creating/disposing providers on different events (notebook added, cell switched ...)

## User-facing changes

https://user-images.githubusercontent.com/4451292/150306043-b6f97ebe-f303-4058-8dda-be4f03f5af32.mp4

### New setting item for `Code completion`
```javascript
{
    // Code Completion
    // @jupyterlab/completer-extension:tracker
    // Code Completion settings.
    // ***************************************

    // Installed completion providers
    "availableProviders": {
        "CompletionProvider:context": 100,
        "CompletionProvider:kernel": 200,
    },

    // Default timeout for a provider
    "providerTimeout": 1000
}
```
- `availableProviders`:  Configuration for installed providers, the keys of this dictionary are the identifier of providers, the values are the priority rank of the provider. Providers with higher rank will be shown before the ones with lower rank, providers with negative rank are disabled.
- `providerTimeout`: If a provider can not return the response for a completer request before `timeout`, the result of this provider will be ignored. Value is in ms

Since `ICompletionManager` is removed, this PR will break all existing completer extensions.


trungleduc:
CC @hbcarlos @krassowski @edzkite

SylvainCorlay:
This is amazing! Wow.

trungleduc:
Hi @jasongrout, this list is restored automatically every time users start Jupyter Lab so changes made by users on that list do not have any effect.

packages/completer/src/connectorproxy.ts
```diff
@@ -0,0 +1,30 @@
...
+    request: CompletionHandler.IRequest
+  ): Promise<Array<{ [id: string]: CompletionHandler.ICompletionItemsReply }>> {
+    let promises: Promise<{
+      [id: string]: CompletionHandler.ICompletionItemsReply;
+    }>[] = [];
+    for (const [id, connector] of this._connectorMap.entries()) {
+      let promise = connector.fetch(request).then(reply => ({ [id]: reply }));
+      promises.push(promise.catch(p => p));
+    }
+    const combinedPromise = Promise.all(promises);
```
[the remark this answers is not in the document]
krassowski: This would be alleviated by https://github.com/jupyterlab/jupyterlab/pull/11831 :)

packages/completer/src/default/connector.ts
```diff
@@ -73,31 +73,22 @@ namespace Private {
...
    */
   export function mergeReplies(
-    kernel: CompletionHandler.IReply,
-    context: CompletionHandler.IReply
-  ): CompletionHandler.IReply {
-    // If one is empty, return the other.
-    if (kernel.matches.length === 0) {
+    kernel: CompletionHandler.ICompletionItemsReply,
+    context: CompletionHandler.ICompletionItemsReply
+  ): CompletionHandler.ICompletionItemsReply {
```
krassowski, line 78: I feel that we should remove this special handling/merging of replies and just have separate completion providers for `context` and `kernel`.

packages/completer/src/connectorproxy.ts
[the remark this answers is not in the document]
trungleduc: Hi @krassowski, thanks for your review!

> * have a timeout for kernel completions (because kernel can take a long time to reply): this can be simply implemented by wrapping each promise with `promise = Promise.race(promise, resolveOnTimeout(settings.connectorTimeout))`
> * do not ask kernel for completions in the first place when the kernel is busy; similarly, the LSP connection may be temporarily disconnected due to network issues so this provider would be registered but temporarily unavailable. This requires changes to the interface; we would need to either:
>   
>   * reject in the base provider when kernel is busy (which means we would want to split the base provider into `contextProvider` and  `kernelProvider`), or
>   * have a method on provider interface to ask the provider "hey, provider are you available to provide completions now?", say `provider.isAvailable`
> 

these are good ideas, as I asked above, I'm not entirely sure about the provider interface `ICompletionProvider`, any suggestions about the attributes you need for a provider are very helpful.

krassowski:
That's great work! I have two major points:
- [ ] let's simplify our the codebase by removing the `CompletionConnector` which agglomerates `KernelConnector` and `ContextConnector` and instead make them two just providers; this should remove quite a few lines of code, reduce duplication and make sure that the provider's API exposes at least a bare minimum to implement the behaviour we already have (i.e. preferring completions from one provider - kernel - over the other - context)
- [ ] I would propose to remove the `Connector` concept (details below); this might be a bit more controversial because it will be a bigger API change, but please see the rationale below and let me know what you think

----

I would like if instead of keeping the `Connector` entity (and introducing the proposed `connectorFactory`) we could instead remove the `connector` completely and go for a simpler API. My grudge against the concept of connector is that it seems to be a forced use of `IDataConnector` interface which does not belong here; the documentation says:

https://github.com/jupyterlab/jupyterlab/blob/6631799510a90938a0ed8bc6fd02b0fc55a067a0/packages/completer/src/handler.ts#L49-L53

And its definition is:

https://github.com/jupyterlab/jupyterlab/blob/6631799510a90938a0ed8bc6fd02b0fc55a067a0/packages/completer/src/handler.ts#L614-L619

So downstream ends up implementing dummy methods which do [absolutely nothing](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/ed88a08288a8f203131b67d59aa98e04124a6af1/packages/jupyterlab-lsp/src/features/completion/completion_handler.ts#L617-L627):

```js
  list(query: string | undefined): Promise<any> {
    return Promise.resolve(undefined);
  }

  remove(id: CompletionHandler.IRequest): Promise<any> {
    return Promise.resolve(undefined);
  }

  save(id: CompletionHandler.IRequest, value: void): Promise<any> {
    return Promise.resolve(undefined);
  }
```

Further, there is no need to perform any preparatory work on the context anymore because if a new physical connection needs to be established, provider now can take care of it (since now it is a thing), so the need for a dedicated `Connector` object goes away (in my eyes - I might be wrong here!).

Instead I would propose that the `Provider` gets to implement a new `async fetch()` method, roughly:

```ts
 fetch(
    context: ICompletionContext
  ): Promise<CompletionHandler.ICompletionItemsReply<T>>;
```

I am happy to be proven wrong on this one if anyone believes that we should keep the connector API (but in that case let's make it _not_ inherit from `IDataConnector` as it is unidirectional).

For reference this is the `IProvider` interface from my [earlier experiments](https://github.com/jupyter-lsp/jupyterlab-lsp/pull/600/files#diff-dcfa4a28b487787de12cb977690666f574be0f533e0bcd55ed544cab5a221a66R95-R137):

<details>

```ts
export interface ICompletionProvider<
  T extends IExtendedCompletionItem = IExtendedCompletionItem
> {
  /**
   * Unique identifier of the provider
   */
  identifier: string;

  /**
   * Is completion provider applicable to specified context?
   * @param request - useful because we may want to disable specific sources in some parts of document (like sql code cell in a Python notebook)
   * @param context
   */
  isApplicable(
    request: ICompletionRequest,
    context: ICompletionContext
  ): Promise<boolean>;

  /**
   * Renderer for provider's completions (optional).
   */
  renderer?: ICompleterRenderer<T>;

  /**
   * Fetch completion requests.
   *
   * @param request - the completion request text and details
   */
  fetch(
    request: ICompletionRequest,
    context: ICompletionContext
  ): Promise<ICompletionsReply<T>>;

  /**
   * Given an incomplete (unresolved) completion item, resolve it by adding all missing details,
   * such as lazy-fetched documentation.
   *
   * @param completion - the completion item to resolve
   */
  resolve?(completion: T): Promise<T>;
}
```

</details>

packages/completer-extension/schema/tracker.json
```diff
@@ -0,0 +1,24 @@
+{
+  "title": "Code Completion",
+  "description": "Code Completion settings.",
+  "jupyter.lab.setting-icon-label": "Code Completion settings",
+  "properties": {
+    "availableProviders": {
```
fcollonval, line 6: As this is purely informative, a better way would be to not add this properties and to dynamically update the description of `selectedProviders` to include a list of available values. You can look at how this is done for listing the command in shortcuts: https://github.com/jupyterlab/jupyterlab/blob/master/packages/shortcuts-extension/src/index.ts#L51

Let me know if you need help to understand that piece of code. This is a tricky part; but the idea is to define a transformation of the settings (using `registry.transform(<id>, callbacks)`) between the value obtained from the server and what the user will see.

trungleduc: Thanks @fcollonval, indeed this `transform` method is what I'm looking for!

trungleduc:
@krassowski thank you so much for these pointers, I'm completely in line with what you proposed about the `Connector`. I keep the old `Connector ` just to make the immigration to the new provider system easier since the extension authors can re-use all internal parts and change only the way they register their providers

If it's ok for everyone, I'll apply your suggestions.

trungleduc:
Hi all, this PR is ready for review.

fcollonval (changes requested):
Thanks @trungleduc here are some suggestions

galata/test/jupyterlab/debugger.test.ts
```diff
@@ -15,6 +15,7 @@ async function openNotebook(page: IJupyterLabPageFixture, tmpPath, fileName) {
 test.describe('Debugger Tests', () => {
   test.afterEach(async ({ page }) => {
     await page.debugger.switchOff();
+    await page.waitForTimeout(500);
```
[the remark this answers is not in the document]
trungleduc: I had this test fails randomly with matching screenshot, then I realized the error comes from the `afterEach` call where we turn off the debugger and close the notebook too fast. Adding a small pause between these two operations helps eliminate the flaky result.
Do you have suggestions to improve this condition?

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,187 @@ namespace CommandIDs {
...
+    settings: ISettingRegistry
+  ): ICompletionProviderManager => {
+    const AVAILABLE_PROVIDERS = 'availableProviders';
+    const PROVIDER_TIMEOUT = 'providerTimeout';
+    const manager = new CompletionProviderManager();
+    const updateSetting = (settingValues: ISettingRegistry.ISettings): void => {
+      const providersData = settingValues.get(AVAILABLE_PROVIDERS);
+      const timeout = settingValues.get(PROVIDER_TIMEOUT);
+      manager.setTimeout(timeout.composite as number);
+      const selectedProviders = providersData.user || providersData.composite;
```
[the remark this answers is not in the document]
trungleduc: Yes, I think that most of the time, users only want to use one provider. By merging the result, users need to explicitly set the rank to negative to disable all other providers instead of just having the provider they want to use in the custom setting panel.

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,187 @@ namespace CommandIDs {
...
+
+        if (userData) {
+          const newData: PartialJSONObject = {};
+          for (const [key, value] of Object.entries(userData)) {
+            if (availableProviders.includes(key)) {
+              newData[key] = value;
+            }
+          }
+          void settingValues.set(AVAILABLE_PROVIDERS, newData);
+        }
```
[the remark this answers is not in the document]
trungleduc: Indeed I forgot this use case of having multiple environments. I'll fix it.

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,187 @@ namespace CommandIDs {
...
-      };
+    app.restored.then(() => {
+      const availableProviders = [...manager.getProviders().keys()];
+      settings.transform(COMPLETION_MANAGER_PLUGIN, {
+        fetch: plugin => {
+          const schema = plugin.schema.properties!;
+          availableProviders.forEach((item, index) => {
+            schema[AVAILABLE_PROVIDERS]['properties']![item] = {
+              type: 'number',
+              default: (index + 1) * 100
```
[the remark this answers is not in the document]
trungleduc: The default rank here is computed from the order of the provider's registration, which I suppose to be the same as the order of installation. So users always get their newly installed provider on top of the completer. For the default rank, It would be confusing for the external provider authors to know which rank they should choose because they do not have control over the rank of other providers.

packages/completer/src/handler.ts
```diff
@@ -353,64 +322,29 @@ export class CompletionHandler implements IDisposable {
...
-          throw new Error(`Invalid request: ${request}`);
+    return this._connector.fetch(request).then(replies => {
+      let start = 0;
+      let end = 0;
+      let items: CompletionHandler.ICompletionItem[] = [];
+      for (const data of replies) {
+        if (data) {
+          items = items.concat(data.items);
+          start = start || data.start;
+          end = end || data.end;
```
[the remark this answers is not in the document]
trungleduc: indeed, I missed that corner case.

SylvainCorlay (approved):

jtpio:
> Since ICompletionManager is removed, this PR will break all existing completer extensions.

Since this is a breaking change, let's document that change in this section of the docs? 

https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#api-breaking-changes

trungleduc:
> > Since ICompletionManager is removed, this PR will break all existing completer extensions.
> 
> Since this is a breaking change, let's document that change in this section of the docs?
> 
> https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#api-breaking-changes

👍 I'll do it.

packages/completer-extension/src/index.ts
```diff
@@ -10,20 +10,16 @@ import {
...
-import { IConsoleTracker } from '@jupyterlab/console';
 import { IEditorTracker } from '@jupyterlab/fileeditor';
 import { INotebookTracker } from '@jupyterlab/notebook';
-import { Session } from '@jupyterlab/services';
-import { find, toArray } from '@lumino/algorithm';
-import { Widget } from '@lumino/widgets';
-
+import { IConsoleTracker } from '@jupyterlab/console';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
+// import { PartialJSONObject } from '@lumino/coreutils';
```
krassowski, line 22:
```suggestion
```

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,175 @@ namespace CommandIDs {
...
- */
-const notebooks: JupyterFrontEndPlugin<void> = {
-  id: '@jupyterlab/completer-extension:notebooks',
-  requires: [ICompletionManager, INotebookTracker],
+const manager: JupyterFrontEndPlugin<ICompletionProviderManager> = {
+  id: COMPLETION_MANAGER_PLUGIN,
+  requires: [
+    INotebookTracker,
+    IEditorTracker,
+    IConsoleTracker,
```
fcollonval, line 64: This gonna be a large change API wise (and the late catch). But this must be moved. JupyterLab infrastructure is based on composition of plugins. So in that spirit, this plugin should indeed provide the manager. But it is then in the notebook, file editor and console extensions that it should be used/connected to the respective editors.

This is particularly important for simplified application.

fcollonval (changes requested):
Sorry @trungleduc for the late catch. But the connection between the completer manager and the other editor plugins must be refactored to better fulfill plugin compositions.

packages/completer-extension/src/index.ts
trungleduc: Thanks @fcollonval, I got the point. Since I'm having a second PR in the pipeline based on this PR, can we defer this change to the second one? It would simplify the rebase work once this PR is merged.

trungleduc: I applied the suggestion here https://github.com/jupyterlab/jupyterlab/pull/11969/commits/625e8ec03aef73b35f22cd75e42c0f31b93f949f

fcollonval:
@trungleduc do you mind fixing the conflict so we can merge this one?

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,175 @@ namespace CommandIDs {
...
-  requires: [ICompletionManager, IConsoleTracker],
+const defaultProvider: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/completer-extension:base-service',
+  requires: [ICompletionProviderManager],
   autoStart: true,
   activate: (
     app: JupyterFrontEnd,
-    manager: ICompletionManager,
-    consoles: IConsoleTracker
+    serviceManager: ICompletionProviderManager
```
jtpio, line 52: Maybe `serviceManager` can be renamed to `completionManager` for clarity?

Since `serviceManager` usually refers to the `ServiceManager.IManager`.

packages/completer-extension/src/index.ts
```diff
@@ -45,337 +41,175 @@ namespace CommandIDs {
...
-/**
- * An extension that registers notebooks for code completion.
- */
-const notebooks: JupyterFrontEndPlugin<void> = {
-  id: '@jupyterlab/completer-extension:notebooks',
-  requires: [ICompletionManager, INotebookTracker],
+const manager: JupyterFrontEndPlugin<ICompletionProviderManager> = {
+  id: COMPLETION_MANAGER_PLUGIN,
+  requires: [
+    INotebookTracker,
```
jtpio, line 62: Do we need all the trackers to be specified as `requires`?

Before they were split into separate plugins so downstream lab-based distributions could cherry-pick the plugins they wanted.

So maybe we can make them `optional` here now.

trungleduc: Indeed, these trackers are removed in my second PR https://github.com/jupyterlab/jupyterlab/commit/625e8ec03aef73b35f22cd75e42c0f31b93f949f

packages/completer/src/connectorproxy.ts
```diff
@@ -0,0 +1,66 @@
...
+    return combinedPromise;
+  }
+
+  private _providers: Array<ICompletionProvider>;
+  private _context: ICompletionContext;
+  private _timeout: number;
+}
+
+export namespace ConnectorProxy {
```
jtpio, line 61: A docstring would be useful here since the namespace is exported.

packages/completer/src/default/kernelprovider.ts
```diff
@@ -0,0 +1,72 @@
...
+import { ICompletionContext, ICompletionProvider } from '../tokens';
+import { CompletionHandler } from '../handler';
+import { JSONObject } from '@lumino/coreutils';
+
+export const KERNEL_PROVIDER_ID = 'CompletionProvider:kernel';
+/**
+ * A kernel connector for completion handlers.
+ */
+export class KernelCompleterProvider implements ICompletionProvider {
+  async isApplicable(context: ICompletionContext): Promise<boolean> {
```
jtpio, line 14: Let's add a docstring here too and to other definitions of `isApplicable`?

packages/completer/src/handler.ts
```diff
@@ -609,19 +457,26 @@ export namespace CompletionHandler {
      * Indicates if the item is deprecated.
      */
     deprecated?: boolean;
+
+    /**
+     * {Provider id will be added automatically.
```
jtpio, line 462: Extra curly brace?

```suggestion
     * Provider id will be added automatically.
```

packages/completer/src/connectorproxy.ts
trungleduc: Adressed in https://github.com/jupyterlab/jupyterlab/commit/625e8ec03aef73b35f22cd75e42c0f31b93f949f

packages/completer/test/handler.spec.ts
```diff
@@ -301,23 +289,23 @@ describe('@jupyterlab/completer', () => {
           line,
           column: column + 6
         });
-        console.warn(editor.getCursorPosition());
+        // console.warn(editor.getCursorPosition());
```
jtpio, line 292: Should these be kept or removed?

trungleduc:
> @trungleduc do you mind fixing the conflict so we can merge this one?

Done

fcollonval (approved):
Merging so we can move on with the follow-up.

PR #11801, opened 2022-01-07
Fix `OSTYPE` check in `ci_install.sh`

jtpio:
## References

As noticed in https://github.com/jupyterlab/jupyterlab/pull/11770#issuecomment-1003129065, the check against `$OSTYPE` in the `ci_install.sh` script doesn't seem to be hit at the moment.

## Code changes

Update CI install script to see if this has any affect on the CI checks.

## User-facing changes

None

None

### DIFF
--- a/buildutils/src/local-repository.ts
+++ b/buildutils/src/local-repository.ts
@@ -32,10 +32,10 @@ async function startLocalRegistry(out_dir: string, port = DEFAULT_PORT) {
 
   // Get current registry values
   let prev_npm = utils.run('npm config get registry', { stdio: 'pipe' }, true);
-  let prev_yarn = '';
+  let prev_jlpm = '';
   try {
-    prev_yarn = utils.run(
-      'yarn config get npmRegistryServer',
+    prev_jlpm = utils.run(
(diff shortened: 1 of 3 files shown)


jtpio:
So both `check_release` and `release_test` seem to be passing if we use `jlpm` instead of `yarn`.

Taking a second look at it, it seems we are using `jlpm` in the other `@jupyterlab/buildutils` scripts.

cc @blink1073 just to confirm such change would be appropriate.

blink1073:
Yes that sounds correct.

jtpio:
cc @jupyterlab/release for awareness

This should be good to go now.

krassowski (approved):
I think this makes sense, though I do not have the full context here.

jtpio:
Thanks @krassowski and @krassowski for the review!

> though I do not have the full context here.

This is mostly to fix this condition on CI noticed in https://github.com/jupyterlab/jupyterlab/pull/11770#issuecomment-1003129065.

jtpio:
Merging as CI is now all passing.

Issue #11696, opened 2021-12-15
[JLab v4a17] CodeMirror width is not limited when text overflow


fcollonval:
## Description

As can displayed in the screenshot below, codemirror editor is expanding outside the notebook panel when it contains a long line.
The horizontal scrollbar is displayed on the notebook panel. It used to be a fix width with a local (= at editor level) horizontal scrollbar when text overflow.

![image](https://user-images.githubusercontent.com/8435071/146201711-b6ebebc2-2df0-43cc-9f23-c5d5d0ff5df0.png)

## Reproduce

1. Open a notebook
2. Type a very long line (longer that the notebook panel width)
3. The horizontal scrollbar appears at the bottom of the notebook panel not in the editor.

## Expected behavior

Codemirror width is identical and the scrollbar appears at its bottom. Like in 3.2:

![image](https://user-images.githubusercontent.com/8435071/146202412-fcd9f16f-eeac-4847-96c1-f8c89b8decf6.png)

The changes made in https://github.com/jupyterlab/jupyterlab/pull/11508 could be the source of this bug.

## Context

- Operating System and version:  Debian 11
- Browser and version:  Firefox 95
- JupyterLab version:  v4 alpha 17

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

VijayaRaghavanM:
@fcollonval I'd like to take up this task

rfox12:
I think Frederic is correct about changes made in #11508.  I can't see an easy fix to this; but when I toggle off "display: table" from .jp-InputArea then the cell looks to be the correct width.

pree-T:
I want to work on this. Can someone help me a bit with this?

fcollonval:
Hey @pree-T 

I think the change that breaks the style is the following:

![image](https://user-images.githubusercontent.com/8435071/147650174-61dc78a0-79e0-4d20-a336-936704a8df47.png)

You could try to set some constrain on the `width` or `max-width` to see if you can prevent the cell input growth.

rfox12:
[...]

PR #11805, opened 2022-01-08
fix for codemirror width

pree-T:
#11696  a possible fix for this. Kindly review

### DIFF
--- a/packages/cells/style/inputarea.css
+++ b/packages/cells/style/inputarea.css
@@ -11,11 +11,13 @@
 .jp-InputArea {
   display: table;
   overflow: hidden;
+  table-layout: fixed;
+  width: 100%;
 }
 
 .jp-InputArea-editor {
   display: table-cell;
-  width: 100%;
+ 
   overflow: hidden;
   vertical-align: top;
 }
@@ -40,6 +42,7 @@
   font-size: var(--jp-code-font-size);
   border: var(--jp-border-width) solid transparent;
   opacity: var(--jp-cell-prompt-opacity);
+  width: var(--jp-cell-prompt-width);
   /* Right align prompt text, don't wrap to handle large prompt numbers */
   text-align: right;
   white-space: nowrap;
@@ -59,6 +62,7 @@
   .jp-InputArea-editor {
     display: table-row;
     margin-left: var(--jp-notebook-padding);
+    
   }
 
   .jp-InputPrompt {
--- a/packages/cells/style/inputarea.css
+++ b/packages/cells/style/inputarea.css
@@ -10,14 +10,13 @@
 /* All input areas */
 .jp-InputArea {
   display: table;
-  overflow: hidden;
   table-layout: fixed;
-  width: 100%;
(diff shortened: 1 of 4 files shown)


packages/cells/style/inputarea.css
```diff
@@ -59,6 +62,7 @@
   .jp-InputArea-editor {
     display: table-row;
     margin-left: var(--jp-notebook-padding);
+
```
fcollonval, line 65: Unneeded empty line

```suggestion
```

packages/cells/style/inputarea.css
```diff
@@ -40,6 +42,7 @@
   font-size: var(--jp-code-font-size);
   border: var(--jp-border-width) solid transparent;
   opacity: var(--jp-cell-prompt-opacity);
+  width: var(--jp-cell-prompt-width);
```
fcollonval, line 45: Could you remove the `min-width` rule?

packages/cells/style/inputarea.css
```diff
@@ -11,11 +11,13 @@
...
   display: table;
   overflow: hidden;
+  table-layout: fixed;
+  width: 100%;
 }
 
 .jp-InputArea-editor {
   display: table-cell;
-  width: 100%;
+
```
fcollonval, line 20: Unneeded empty line

```suggestion
```

tsconfigdoc.json
```diff
@@ -20,274 +20,274 @@
   },
   "references": [
     {
-      "path": "./packages/application"
+      "path": "./packages\\application"
```
fcollonval, line 23: Would you mind reverting the change in this file please?

fcollonval:
Open question: this changes the width of the input prompt to a constant width. Is this acceptable? Do we need to apply the same change in the output area (I would say yes)?

fcollonval (changes requested):
Thanks @pree-T I made some suggestions. And they are one question regarding the need to apply similar changes to the output area that is likely to be confirmed by others.

fcollonval:
I looked again at the problematic PR https://github.com/jupyterlab/jupyterlab/pull/11508/files

Actually the error comes from the style `flex: 0 0 <width>` that have been replaced by a `min-width` constrain when they should have been replaced by `width` constraint.

So could you please replace `min-width` by `width` in both:

https://github.com/jupyterlab/jupyterlab/blob/4a9c1d52c760bde51e0af0ec60a46ac3053cb78c/packages/outputarea/style/base.css#L31

https://github.com/jupyterlab/jupyterlab/blob/4a9c1d52c760bde51e0af0ec60a46ac3053cb78c/packages/cells/style/widget.css#L96

And set that table as `table-layout: fixed;` as you did for the input area.

https://github.com/jupyterlab/jupyterlab/blob/4a9c1d52c760bde51e0af0ec60a46ac3053cb78c/packages/outputarea/style/base.css#L22

pree-T:
Sure Sir, I'll make all the necessary changes. Thanks for the guidance .

PR #11804, opened 2022-01-07
Backport PR #11356: Toc running cell indicator

andrewfulton9:
## References
Backports #11356 to 3.3.x

Issue #11643, opened 2021-12-09
JupyterLab 3.3 Release


echarles:
Hi, I am opening this to discuss and gather inputs on a potential 3.3. release from the current 3.2 branch before the 4.0 release. We have a few ongoing PR that, if merged, would deserve a release before the 4.0 release.

The idea would be to get a 3.3. release in January 2022. Of course, if the 4.0 release is achievable by that time, we could skip that intermediary step.

List of pending PR to consider for 3.3 - Please edit this comment to add yours.

- [x] https://github.com/jupyterlab/jupyterlab/pull/11752
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/pull/11608  
       https://github.com/jupyterlab/jupyterlab/pull/11718  
       https://github.com/jupyterlab/jupyterlab/pull/11794
  -  Check if PR for resizing is already backported in 3.x
- [x] https://github.com/jupyterlab/jupyterlab/pull/11566  
  https://github.com/jupyterlab/jupyterlab/pull/11954
  https://github.com/jupyterlab/jupyterlab/pull/12068
- [x] https://github.com/jupyterlab/jupyterlab/issues/11735. 
  https://github.com/jupyterlab/jupyterlab/pull/11909
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/issues/11719
  https://github.com/jupyterlab/jupyterlab/pull/11948
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/issues/11700
  https://github.com/jupyterlab/jupyterlab/pull/11929
- [x] https://github.com/jupyterlab/jupyterlab/issues/11710
  https://github.com/jupyterlab/jupyterlab/pull/11956
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/pull/11646  
  https://github.com/jupyterlab/jupyterlab/pull/11867
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/pull/11740  
  https://github.com/jupyterlab/jupyterlab/pull/11866
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/pull/10299  
  https://github.com/jupyterlab/jupyterlab/pull/11874
- [x] :ballot_box_with_check: https://github.com/jupyterlab/jupyterlab/pull/11171  
  https://github.com/jupyterlab/jupyterlab/pull/11874
[...]

Issue #7237, opened 2019-09-19
CI check that ensures PRs are not making changes to jupyterlab/staging


jasongrout:
The `jupyterlab/staging` directory is only modified at publish time by the release script. We should have a CI check on PRs making sure that no PRs are modifying the staging directory.

jasongrout:
I suppose changing the staging/yarn.js file is something done in a PR, so maybe we just whitelist a few specific files allowed to change in a PR.

jtpio:
Maybe this action triggered on `pull_request` only could work? https://github.com/tj-actions/changed-files

galathinius:
Hi @jasongrout I'd like to try working on this issue, could you please tell me which files would need to be whitelisted?

jtpio:
Thanks @galathinius!

> which files would need to be whitelisted

Maybe we can start with `jupyterlab/staging/yarn.js`? https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/staging/yarn.js
Ideally this list can be configured easily, so we could add more files to it if needed, for example: https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/staging/.yarnrc

galathinius:
I just opened a pr with my implementation, new files could be added in the workflow yaml. 
Or the list of files could be written to a separate file and read from there

PR #11810, opened 2022-01-10
[3.3.x] Bump version for the 3.3 prerelease

jtpio:
## References

As mentioned in https://github.com/jupyterlab/jupyterlab/issues/11643#issuecomment-1007493038

This should fix CI issues on the `3.3.x` branch as noticed in https://github.com/jupyterlab/jupyterlab/pull/11804

## Code changes

- Bump the JS versions to `3.3.0-alpha.15`
- Keep `@jupyterlab/ui-components` at `3.3.0-alpha.14` because it was bumped independently as part of https://github.com/jupyterlab/jupyterlab/pull/11173
- Bump the Python version to `3.3.0a0`

This bump is done manually because some of the JS versions have already been used and released (during the `4.0.0ax` series) on `master` when the major version bump had not happened yet. For example: 

https://github.com/jupyterlab/jupyterlab/blob/d1244eaa11eada494d98826e16ca42b31b9476f5/packages/application/package.json

## User-facing changes

The first 3.3 prerelease will then be `3.3.0a1` since the releaser will bump the version during the release process

None


jtpio:
The failing check is expected (more than 100 labels applied).

jtpio:
Pending the remaining CI checks, this should now be good to go.

Ideally we would have done the bump and release with the releaser directly. But because of some `3.3.0-alpha.x` packages already used and published (detailed in https://github.com/jupyterlab/jupyterlab/issues/11643#issuecomment-1007493038), the script would have bumped versions from `3.2.x` to `3.3.0-alpha.0`.

jupyterlab/staging/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jupyterlab/application-top",
-  "version": "3.2.5",
+  "version": "3.3.0-alpha.15",
```
jtpio, line 3: Here we manually update the `staging` files just to be able to realign the versions.

Normally these files should not be updated in regular PRs: https://github.com/jupyterlab/jupyterlab/issues/7237

jtpio:
Let's merge to make sure this fixes the CI. 

We can rebase some of the opened backport PRs targeting `3.3.x` right after.

fcollonval (approved):
Thanks a lot @jtpio

PR #11813, opened 2022-01-10
Remove `url` dependency from `@jupyterlab/apputils`

jtpio:
## References

Attempt at removing the `url` dependency from `@jupyterlab/apputils` which was added in https://github.com/jupyterlab/jupyterlab/pull/8659

## Code changes

Try to remove the `url` dependency since it is now showing the following warning:

```
warning @jupyterlab/apputils > url > querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
```

## User-facing changes

None

None


jtpio:
CI is all green.

fcollonval (approved):
Thanks @jtpio

PR #11803, opened 2022-01-07
DOCS: Convert theme to pydata-sphinx-theme

choldgraf:
## References

closes https://github.com/jupyterlab/jupyterlab/issues/10834

## Documentation changes

This PR converts the documentation of the JupyterLab theme to `pydata-sphinx-theme`. Since @fcollonval noted that there was consensus that this was a good idea, and because I noticed that the jupyter-server docs have been converted to this theme as well by @Zsailer et al.

This doesn't change the structure of the documentation much, but the landing page of each section is now a top-level link in the navigation bar. There's one extra "landing page" for the user guide added, because there wasn't a natural landing page that was already there.

I've also added a logo image + some icons to other jupyterlab spaces.

PR #11821, opened 2022-01-11
Use playwright to generate documentation screenshots

fcollonval:
## References

https://github.com/jupyterlab/jupyterlab/pull/11499#issuecomment-973906425

## Code changes

Add playwright tests to generate documentation screenshots.

Add the debugger tutorial directly into the documentation to have its screenshots updated.
Rework the extension manager documentation

Screenshots not automated: RTC screenshots - I don't know if that will be possible; maybe by creating two browsers instances.

> Some screenshots are generated for the documentation without being part of the tests because some visual artifacts keep changing (like the kernel PID appearing in the debugger temporary file names) - those are in `galata/test/documentation/screenshots`. An updated version is generated by running the tests.

## User-facing changes

N/A

Try to get as close as possible to the manually taken screenshots


jtpio:
This is looking great, thanks @fcollonval for starting this!

fcollonval:
This is ready for review
[Documentation preview link](https://jupyterlab--11821.org.readthedocs.build/en/11821/index.html)

fcollonval:
We get the same flaky test on the documentation for setting the breakpoint gutter. The waiting condition before taking the snapshot is probably not robust enough as the gutter does appear when looking at the test video.

jtpio:
> This is ready for review [Documentation preview link](https://jupyterlab--11821.org.readthedocs.build/en/11821/index.html)

Thanks! Should this get a rebase to test with the new theme added in https://github.com/jupyterlab/jupyterlab/pull/11803?

fcollonval:
@jtpio this is ready for review and rebased to use the new sphinx template.

.github/workflows/galata.yml
```diff
@@ -84,3 +85,121 @@ jobs:
...
+          set -ex
+          # Install chinese font
+          sudo apt-get update
+          sudo apt-get install fonts-dejavu fonts-noto
+
+          bash ./scripts/ci_install.sh
+          # Freeze the packages to ensure consistent look and feel
+          # IPython is frozen because its version is displayed in
+          # the console header
+          pip install altair==4.2.0 ipython==8.0.0 ipywidgets==7.6.5 jupyterlab-geojson==3.2.0 jupyterlab-language-pack-zh-CN==3.2.post7 matplotlib==3.5.1 pandas==1.3.5 scipy==1.7.3 vdom==0.6 vega_datasets==0.9.0
```
jtpio, line 132: Wondering whether we should extract these frozen dependencies somewhere else in a separate file, so it's easier to keep track of them?

jtpio (approved):
This is great stuff, thanks @fcollonval!

Just did a quick tour on the RTD preview and it looks good :tada: 

image

jtpio:
As a follow-up and future idea, I wonder if it will be possible to replace the YouTube embedded videos by new videos generated with Playwright?

image

fcollonval:
> As a follow-up and future idea, I wonder if it will be possible to replace the YouTube embedded videos by new videos generated with Playwright?

This is technically doable. The major issues are:
- Need for displaying a mouse pointer and emulate its _slow_ displacement
- Need for highlighting mouse event (like clicking) and may be keyboard event
- How do you verify the video is good? Is it ok to for instance only do a check on the final state (like for the screenshots)?
- [The big :question:] How to do you slice the test; especially to remove the application start?  
  They are at least two paths here: 
  - Slicing the (all?) videos with ffmpeg for example (but this means we need to configure for each videos what is a good start time (and end time?))
  - Running serial Playwright test to avoid restarting a new session (and the associated loading time). But this may be more annoying for developers because one intermediate failing test may hide multiple errors.

.github/workflows/galata.yml
fcollonval: What about adding an additional extra in setup.cfg? We could probably drop `docs/environment.yml` and add two extras like: `doc` and `doc-screenshots`.

jtpio: Sounds like extras would work.

Don't we need to keep `docs/environment.yml` for ReadTheDocs? 

https://github.com/jupyterlab/jupyterlab/blob/1903d0dd09b2415a998dd46de15b54a5c8988db5/readthedocs.yml#L4-L5

fcollonval: We could switch to [pip](https://docs.readthedocs.io/en/stable/config-file/v2.html#python) and specify that nodejs is a need build tool.

jtpio:
Looks like the Visual Regression Documentation check is still a bit flaky: https://github.com/jupyterlab/jupyterlab/runs/4969825698?check_suite_focus=true

fcollonval:
> Looks like the Visual Regression Documentation check is still a bit flaky: https://github.com/jupyterlab/jupyterlab/runs/4969825698?check_suite_focus=true

This PR was impacted by #11566 that added a panel in the debugger.

fcollonval:
@jtpio I think I reach a more robust point.

fcollonval:
Merging this one

jtpio:
Thanks, looks great!

Issue #10837, opened 2021-08-15
Links in dialogs are not visually distinct


krassowski:
Looking into #10836 remained me that the links in popup dialogs are not visually distinct:

![Screenshot from 2021-08-15 17-13-28](https://user-images.githubusercontent.com/5832902/129485059-a49fc82c-888e-42cf-8919-5ac9245673d1.png)

"the the Jupyter security documentation" is a link.

Should we change the colours of links and maybe add some other visual indicators?

Here is another example of how a link in a popup is displayed, but with the text equal to the URL:

![Screenshot from 2021-08-15 17-16-13](https://user-images.githubusercontent.com/5832902/129485142-215816f1-c507-4fce-abd2-0e4330b04763.png).

By the way I think that we should avoid using URLs as the text of a link as this might be annoying for screen readers.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/10836#issuecomment-899075385_

jweill-aws:
Call me old-school, but I still prefer having links be a different color and underlined to emphasize that they are present in hypertext.

jweill-aws:
This is what the dialog would look like with link styling overrides removed.
![image](https://user-images.githubusercontent.com/93281816/149033690-4cca2971-3e5e-421f-bf26-ead75133bf78.png)

jweill-aws:
Incidentally, the message says "Selecting **trust** will …" (emphasis added) but the action button in the dialog reads "Ok".

Jessie-Newman:
Looks like this was fixed in https://github.com/jupyterlab/jupyterlab/pull/11827

jasongrout:
Awesome, confirmed. Thanks for linking the PR!

PR #11827, opened 2022-01-11
Trust dialog link styling, text changes

jweill-aws:
Updates the link styling in UI components such as the Trust dialog. Updates the text in the Trust dialog's body. Updates the primary action button to be "Trust" instead of "Ok" to match text of dialog box.

## References

See #10837.

## Code changes

Removes link styling overrides that make links visually indistinct from text. Updates dialog body and button text.

## User-facing changes

After change, note:
1. "Trust" (in quotes) in the dialog body
2. Primary action button renamed from "Ok" to "Trust"
3. "the Jupyter security documentation" link is now visually distinct from the rest of the dialog box's body.

image

None.


packages/ui-components/style/base.css
```diff
@@ -51,16 +51,6 @@ strong {
   font-weight: unset;
 }
 
-/* Override Blueprint's _typography.scss styles */
-a {
-  text-decoration: unset;
```
fcollonval, line 56: Did you do a quick search to check there are no `a`-tags in core that are use for buttons and so that probably should not be decorated?

Styling rules, it seems the default behavior of browser is to break the underline when needed (as seen in your screenshot). This is recommended [there](https://www.a11y-101.com/development/underlined-links) too. But should we [enforce that rule](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink) or let the web browser behaves as it wants?

krassowski:
One backward-compatible way would be to add a selector only for dialogs and use the same styling as we already use in rendered markdown (blue colour but no underline).

packages/ui-components/style/base.css
jweill-aws: I modified the styling to leave most `a` tags in the application styled as they have been. I would prefer to leave the underlining behavior up to the browser.

packages/ui-components/style/base.css
```diff
@@ -51,12 +51,13 @@ strong {
   font-weight: unset;
 }
 
-/* Override Blueprint's _typography.scss styles */
-a {
+/* Override Blueprint's _typography.scss styles outside of dialog boxes */
+a:not(.jp-VisibleLink) {
```
krassowski, line 55: Could we just use `.jp-Dialog-content` here rather than introducing a new class (which needs to be manually added)? I am sure there are links in dialogs of extensions (and possibly in other dialogs in core) which suffer from lack of distinction.

packages/ui-components/style/base.css
```diff
@@ -51,12 +51,13 @@ strong {
   font-weight: unset;
 }
 
-/* Override Blueprint's _typography.scss styles */
-a {
+/* Override Blueprint's _typography.scss styles outside of dialog boxes */
```
krassowski, line 54: There is no Blueprint anymore, this is outdated

```suggestion
/* Disable native link decoration styles everywhere outside of dialog boxes */
```

packages/ui-components/style/base.css
jweill-aws: Reverted the new `jp-VisibleLink` class and added two new CSS rules to override the `a` and `a:hover` rules.

packages/ui-components/style/base.css
fcollonval: FYI I realized that for example firefox has a setting to toggle underlining links. So indeed best to let user do what they like through the browser.

fcollonval (approved):
LGTM

PR #11831, opened 2022-01-12
Experiment sending the complete_request message on the control channel.

jasongrout:
## Code changes

With ipykernel 6.0 or later, this enables the complete request to be processed in parallel with execution, so we can still get completions while the kernel is busy.

This is an experiment to see if we can improve autocomplete responsiveness. Luckily, it seems that ipykernel accepts any shell message on the control channel, so this should work with any ipykernel version 6 or later. The messaging spec implies that sending completion requests to the control channel is not officially supported.

## User-facing changes

In one cell, execute a sleep, like: `import time; time.sleep(20)`

Immediately after that, while it is still busy, try doing autocomplete (like `pri<tab>`). The autocomplete should work.

### DIFF
--- a/packages/completer/src/kernelconnector.ts
+++ b/packages/completer/src/kernelconnector.ts
@@ -42,7 +42,7 @@ export class KernelConnector extends DataConnector<
       cursor_pos: request.offset
     };
 
-    const msg = await kernel.requestComplete(contents);
+    const msg = await kernel.requestCompleteControl(contents);
     const response = msg.content;
 
     if (response.status !== 'ok') {
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -611,7 +611,7 @@ export class KernelConnection implements Kernel.IKernelConnection {
   }
 
   /**
-   * Send a `complete_request` message.
+   * Send a `complete_request` message on the shell channel.
    *
    * #### Notes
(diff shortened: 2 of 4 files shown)


jasongrout:
@krassowski mentions in the dev meeting that jedi, when executed in threads, can cause huge issues. @krassowski, can you elaborate?

jasongrout:
The [Jedi docs](https://jedi.readthedocs.io/en/latest/docs/api.html?#script) include this warning:

> By default jedi.settings.fast_parser is enabled, which means that parso reuses modules (i.e. they are not immutable). With this setting Jedi is not thread safe and it is also not safe to use multiple Script instances and its definitions at the same time.
> If you are a normal plugin developer this should not be an issue. It is an issue for people that do more complex stuff with Jedi.
> This is purely a performance optimization and works pretty well for all typical usages, however consider to turn the setting off if it causes you problems. See also [this discussion](https://github.com/davidhalter/jedi/issues/1240).

jasongrout-db:
> The problem would be only if this change would allow to send two completion requests simultaneously for concurrent execution.

This could happen if a client sent a completion request on the control channel while another client sent a completion request on the shell channel.

jasongrout:
> This could happen if a client sent a completion request on the control channel while another client sent a completion request on the shell channel.

Of course, the easy thing here is to introduce a coordination mechanism in ipykernel between the control and main thread to not process completion messages concurrently.

PR #11841, opened 2022-01-12
Protocol alignment

davidbrochart:
## Code changes

See https://github.com/jupyter-server/jupyter_server/pull/657

## User-facing changes

Improved scalability.

None, since we fall back to the default protocol if the new kernel wire protocol over websocket is not supported by the backend.

### DIFF
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -1614,7 +1614,7 @@ export class KernelConnection implements Kernel.IKernelConnection {
    * Websocket to communicate with kernel.
    */
   private _ws: WebSocket | null = null;
-  private _supported_protocols: string[] = ['v1.websocket.jupyter.org'];
+  private _supported_protocols: string[] = ['v1.kernel.websocket.jupyter.org'];
   private _username = '';
   private _reconnectLimit = 7;
   private _reconnectAttempt = 0;
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -1618,7 +1618,9 @@ export class KernelConnection implements Kernel.IKernelConnection {
   private _reconnectLimit = 7;
   private _reconnectAttempt = 0;
   private _reconnectTimeout: any = null;
-  private _supportedProtocols = KernelMessage.supportedKernelWebSocketProtocols;
+  private _supportedProtocols: string[] = Object.values(
+    KernelMessage.supportedKernelWebSocketProtocols
+  );
 
   private _futures = new Map<
     string,
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -1614,7 +1614,7 @@ export class KernelConnection implements Kernel.IKernelConnection {
    * Websocket to communicate with kernel.
    */
   private _ws: WebSocket | null = null;
-  private _supported_protocols: string[] = ['0.0.1'];
+  private _supported_protocols: string[] = ['v1.websocket.jupyter.org'];
   private _username = '';
   private _reconnectLimit = 7;
   private _reconnectAttempt = 0;
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -1224,7 +1224,7 @@ export class KernelConnection implements Kernel.IKernelConnection {
   /**
    * Create the kernel websocket connection and add socket status handlers.
    */
-  private _createSocket = (use_protocols = true) => {
+  private _createSocket = (useProtocols = true) => {
     this._errorIfDisposed();
 
     // Make sure the socket is clear
@@ -1256,11 +1256,12 @@ export class KernelConnection implements Kernel.IKernelConnection {
     }
 
     // Try opening the websocket with our list of subprotocols.
-    // If the server doesn't handle subprotocols, the accepted protocol will be ''.
+    // If the server doesn't handle subprotocols,
+    // the accepted protocol will be ''.
     // But we cannot send '' as a subprotocol, so if connection fails,
     // reconnect without subprotocols.
-    const supported_protocols = use_protocols ? this._supported_protocols : [];
-    this._ws = new settings.WebSocket(url, supported_protocols);
+    const supportedProtocols = useProtocols ? this._supportedProtocols : [];
+    this._ws = new settings.WebSocket(url, supportedProtocols);
 
     // Ensure incoming binary messages are not Blobs
     this._ws.binaryType = 'arraybuffer';
@@ -1532,14 +1533,14 @@ export class KernelConnection implements Kernel.IKernelConnection {
   private _onWSOpen = (evt: Event) => {
     if (
       this._ws!.protocol !== '' &&
-      !this._supported_protocols.includes(this._ws!.protocol)
+      !this._supportedProtocols.includes(this._ws!.protocol)
     ) {
       console.log(
         'Server selected unknown kernel wire protocol:',
         this._ws!.protocol
       );
       this._updateStatus('dead');
-      throw new Error('Unkown kernel wire protocol: ' + this._ws!.protocol);
+      throw new Error(`Unknown kernel wire protocol:  ${this._ws!.protocol}`);
     }
     this._ws!.onclose = this._onWSClose;
     this._ws!.onerror = this._onWSClose;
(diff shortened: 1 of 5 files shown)


davidbrochart:
This PR is now ready for review.
It adds support for a new kernel wire protocol over websocket, as discussed in https://github.com/jupyter-server/jupyter_server/pull/657.
Jupyter Lab now has a [list of supported protocols](https://github.com/jupyterlab/jupyterlab/blob/efe440cd11138d3f5ae88614afae627064b00b3d/packages/services/src/kernel/default.ts#L1595) (not including the legacy protocol). We first try and connect with our list of supported protocols, here ['0.0.1']. We must handle the case of a legacy server that doesn't handle subprotocols, which will return a protocol named `''`. But we cannot have `''` in our list of subprotocols, so the connection will fail. When we reconnect, we drop our list of subprotocols, so the connection will succeed, and we will use the default protocol.
If the server selects one of the protocols in our list at the first connection attempt, then the connection succeeds and we will use that protocol.

davidbrochart:
The new protocol name is now `v1.kernel.websocket.jupyter.org`.

davidbrochart:
The CI failures:
```
error >> WebSocket connection to 'ws://localhost:8888/foo/api/kernels/e17c5184-f72d-4ffc-a45b-09e26528c204/channels?session_id=30274aab-2172-4df0-b4d1-86f630402b6d' failed: Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received
```
are actually caused by the way we handle protocols:
- The front-end first tries to open the websocket with a list of supported protocols. Currently, this only consists in `['v1.kernel.websocket.jupyter.org']`. The legacy protocol has no name, it is implicitly supported.
- If the back-end accepts one of the protocols in the list, everything is fine.
- If the back-end doesn't support any of the protocols in the list (as it is the case in the CI), this generates an error in the websocket connection. The front-end catches the error and retries to open the websocket, this time with an empty list of protocols.
- The back-end must accept the websocket connection, because it uses the legacy procotol, which is always supported.

The drawbacks of this approach are the following:
- If the websocket connection is not opened on the first try, for any other reason than the back-end not supporting the new protocol (e.g. time-out), then we fall back to the legacy protocol in subsequent retries. But I guess we can live with that for a while, until JupyterLab requires a jupyter_server version that supports the new protocol.
- It generates an error in the browser console.

For the CI failures, now that https://github.com/jupyter-server/jupyter_server/pull/657 is merged, and once it's released, we could require this version in the tests.

davidbrochart:
CI failure seems unrelated.

davidbrochart:
OK to merge?

SylvainCorlay:
cc @afshin for discussion - plan to merge?

packages/services/src/kernel/default.ts
```diff
@@ -1524,6 +1530,18 @@ export class KernelConnection implements Kernel.IKernelConnection {
    * Handle a websocket open event.
    */
   private _onWSOpen = (evt: Event) => {
+    if (
+      this._ws!.protocol != '' &&
+      !this._supported_protocols.includes(this._ws!.protocol)
+    ) {
+      console.log(
+        'Server selected unkown kernel wire protocol:',
```
afshin, line 1538:
```suggestion
        'Server selected unknown kernel wire protocol:',
```

packages/services/src/kernel/default.ts
```diff
@@ -1524,6 +1530,18 @@ export class KernelConnection implements Kernel.IKernelConnection {
...
+        this._ws!.protocol
+      );
+      console.log('Supported protocols are:', this._supported_protocols);
```
afshin, line 1541: We should probably remove this `console.log` before shipping.

packages/services/src/kernel/serialize.ts
```diff
@@ -3,13 +3,156 @@
 
 import * as KernelMessage from './messages';
 
+/**
```
afshin, line 6: Could we please use `namespace Private` like we do in other packages for the private functions and all can we use the TypeScript camelCase conventions that we use elsewhere?

packages/services/src/kernel/default.ts
```diff
@@ -1596,6 +1614,7 @@ export class KernelConnection implements Kernel.IKernelConnection {
    * Websocket to communicate with kernel.
    */
   private _ws: WebSocket | null = null;
+  private _supported_protocols: string[] = ['v1.kernel.websocket.jupyter.org'];
```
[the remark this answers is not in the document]
davidbrochart: I moved it to [packages/services/src/kernel/messages.ts](https://github.com/jupyterlab/jupyterlab/blob/7ab6480d67d437b89b4fc6a18a5a7904b2cf9744/packages/services/src/kernel/messages.ts).

davidbrochart:
Thanks for the review @afshin, I made the suggested changes.

davidbrochart:
Thanks @afshin for finishing this!

jtpio:
We should check whether the `js-services` failure is relevant: https://github.com/jupyterlab/jupyterlab/runs/5193491592?check_suite_focus=true

And it would also be preferable to have the CI fully fixed (#12043) before merging this one.

jasongrout:
Just curious, what is sent currently (before these protocol changes) on the zmq channel when parent header is not given? Is an empty dict sent?

jasongrout:
> This last commit fixes the issue by sending `'{}'` when no parent header is present in the message.

In other words, is this change backwards-compatible, producing the same zmq messages that used to be sent?

From https://jupyter-client.readthedocs.io/en/stable/messaging.html#parent-header, it appears that request messages, for example, don't have parent headers, but on the other hand, that paragraph also says:

> If there is no parent, an empty dict should be used.

so I can see having an empty dict as the "right" thing.

davidbrochart:
> Just curious, what is sent currently (before these protocol changes) on the zmq channel when parent header is not given? Is an empty dict sent?

Because we sent the whole message without chunking it up into its parts (except for the binary buffers, if present), we sent whatever is present in `msg`. If there is no parent header, there won't be any `parent_header` dictionary in there.

> is this change backwards-compatible, producing the same zmq messages that used to be sent?

It doesn't produce the same zmq messages that used to be sent, so I guess you're right, this change is not backwards-compatible.

> From https://jupyter-client.readthedocs.io/en/stable/messaging.html#parent-header, it appears that request messages, for example, don't have parent headers, but on the other hand, that paragraph also says:
> 
> > If there is no parent, an empty dict should be used.
> 
> so I can see having an empty dict as the "right" thing.

Thanks, I hadn't seen that. So it looks like we are now doing what the spec recommends. Should it still be seen as a backwards-incompatible change then?

davidbrochart:
> Just curious, what is sent currently (before these protocol changes) on the zmq channel when parent header is not given? Is an empty dict sent?

I didn't answer your question in my previous comment. On the zmq channel [we send an empty dictionary if no `parent_header` is present in the message](
https://github.com/jupyter/jupyter_client/blob/10decd25308c306b6005cbf271b96493824a83e8/jupyter_client/session.py#L663). So I guess we will be more coherent after this change.

jasongrout:
> I didn't answer your question in my previous comment. On the zmq channel [we send an empty dictionary if no parent message exists](https://github.com/jupyter/jupyter_client/blob/10decd25308c306b6005cbf271b96493824a83e8/jupyter_client/session.py#L663). So I guess we will be more coherent after this change.

Thanks for the link. If I'm reading the code correctly and understanding you correctly, the same zmq messages are being generated after this change, and in fact we are encoding this default parent_header from the spec in the frontend with this change. Awesome.

davidbrochart:
That is correct.

davidbrochart:
#12043 was merged but the CI still fails for documentation, I'm not sure it's related to this PR.

jtpio:
> https://github.com/jupyterlab/jupyterlab/pull/12043 was merged but the CI still fails for documentation, I'm not sure it's related to this PR.

There is a new attempt at fixing the check in https://github.com/jupyterlab/jupyterlab/pull/12123

davidbrochart:
OK to merge?

packages/services/src/kernel/messages.ts
```diff
@@ -1256,3 +1256,14 @@ export interface IInputReplyMsg extends IStdinMessage<'input_reply'> {
...
+// ///////////////////////////////////////////////
+// Message (de)serialization
+// ///////////////////////////////////////////////
+
+/**
+ * The list of supported kernel wire protocols over websocket.
+ */
+export const supportedKernelWebSocketProtocols: string[] = [
+  'v1.kernel.websocket.jupyter.org'
+];
```
fcollonval, line 1269: Could you switch to a [enumeration](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) with text value? This will clarify the code.

davidbrochart: :+1:

fcollonval (approved):
Thanks @davidbrochart

Issue #11820, opened 2022-01-11
`username` query parameter does not support escaping.


HoseonRyu:
### Problem
Currently, I'm using `jupyterlab 3.2.5`.

An issue was discovered while using the **collaborative edition mode**. When I put some Korean username in the `username` parameter like this `... /Untitled.ipynb?username=사용자`, it looks like the picture below to others.

![image](https://user-images.githubusercontent.com/31889783/148915487-c38062ad-f50e-4322-be86-c47a4f439475.png)

Also, I tried to include a space to `username` via this url : `... /Untitled.ipynb?username=space space`.

![image](https://user-images.githubusercontent.com/31889783/148912464-9a209084-8bdb-429f-bc6e-ac05dacd047a.png)
The percent encoded name appears, again.

In the `/packages/user/src/model.ts`: line 157, `getParam` gives parsed `username` parameter. However, this function just splits with `&` character as you can check in https://github.com/dmonad/lib0/blob/main/environment.js .
I hope it will be solved for the use of special characters including space and for global users.

### Proposed Solution
It can be solved by simply adding percent decoding codes.

Always thank you for providing great services!

hbcarlos:
Hi @HoseonRyu! Thank you for opening the issue.

Indeed this parameter only supports limited use cases, this was a temporary solution until we built a proper UI to rename anonymous users.

Since you have already debugged the issue and know a solution, do you want to open a PR to solve it?

hbcarlos:
> In the /packages/user/src/model.ts: line 157, getParam gives parsed username parameter.

I think you are looking at the master branch. The master branch has already some features for v4.0. In v3.2.x, we had the same functionality in another package, see:
https://github.com/jupyterlab/jupyterlab/blob/d8e9192baf193c55bbee07c4acef3e262e033391/packages/docprovider/src/yprovider.ts#L45-L46

HoseonRyu:
Hello, @hbcarlos  , thanks for your quick reply!

Then, I'll start to work on `3.2.x` branch. Should I also provide fixes for other versions like 3.3 or 4.0?

hbcarlos:
> Then, I'll start to work on 3.2.x branch.

I'm not sure how to do it in this case, because we move it to another package. Usually, we make the fix in master and then backport it automatically but in this case, I'm not sure if the backport will work. Let me ping someone with more experience.
@jtpio @fcollonval.

> Should I also provide fixes for other versions like 3.3 or 4.0?

[...]

PR #11852, opened 2022-01-13
Add percent decoding to `username` & `initials`

HoseonRyu:
## References
See this issue - [#11820](https://github.com/jupyterlab/jupyterlab/issues/11820).

## Code changes
- Added the function `decodeURIComponent` at the definition of `name`.
- Applied samely to `initials`.
- Edited comments.

### DIFF
--- a/packages/user/src/model.ts
+++ b/packages/user/src/model.ts
@@ -153,10 +153,10 @@ export class User implements ICurrentUser {
    * the user as anonymous if doesn't exists.
    */
   private _fetchUser(): void {
-    // Read username and color from URL
-    let name = env.getParam('--username', '');
+    // Read username, color and initials from URL
+    let name = decodeURIComponent(env.getParam('--username', ''));
     let color = env.getParam('--usercolor', '');
-    let initials = env.getParam('--initials', '');
+    let initials = decodeURIComponent(env.getParam('--initials', ''));
 
     const { localStorage } = window;
     const data = localStorage.getItem(USER);


fcollonval (approved):
LGTM @HoseonRyu thanks for the fix

hbcarlos:
Looks great! Thank you @HoseonRyu!

@fcollonval Can we backport this to 3.1, 3.2, and 3.3?

krassowski:
1. @hbcarlos Is there a way to add a test case for this one?
2. Do we really need `lib0/environment` dependency here, or could we use standard way of parsing URL? (Which could possibly simplify the test case as we would not need to mock `lib0/environment`).

hbcarlos:
> Is there a way to add a test case for this one?

Not sure yet how, but yes we should add a test for this one.
I will take care of all the tests once we have the final API for documents. I added a reference to your comment on the RTC issue.

> Do we really need `lib0/environment` dependency here, or could we use standard way of parsing URL?

Yes, we can use a standard way to parse the URL. Can we keep it as it is for now and I'll change I for 4.0?

fcollonval:
> Can we backport this to 3.1, 3.2, and 3.3?

The policy is to backport only on the current stable branch (to not increase extensively the maintenance bandwidth). The only exception is a security fix.

HoseonRyu:
@fcollonval Then... do I have to open another PR to apply this to `3.2`? That was my first goal.

fcollonval:
> @fcollonval Then... do I have to open another PR to apply this to `3.2`? That was my first goal.

I got my answer on the chat. We will backport this to 3.2.x and 3.3.x. And you don't need to do it manually I'll trigger a bot (that usually succeeds at picking up the changes).

Thanks again for the fix.

fcollonval:

fcollonval:

PR #11875, opened 2022-01-14
Make use of current user id for YJS room

SylvainCorlay:
- Rename `guid` parameter to `room_id` since it is not a guid, but rather a string of the form `notebook:Untitled.ipynb`
- Reuse user id for the id of the user in the YJS room.

### DIFF
--- a/jupyterlab/handlers/user.py
+++ b/jupyterlab/handlers/user.py
@@ -1,0 +1,26 @@
+from jupyter_server.base.handlers import APIHandler
+import json
+
+
+class CurrentUserHandler(APIHandler):
+    """A Handler returning the current user information"""
+
+    @web.authenticated
+    @gen.coroutine
+    def get(self):
+        """GET query returns info on the current user"""
+        yield json.dumps({
+            'name': 'Victor Hugo'
+        })
+
+class ConnectedUsersHandler(APIHandler):
+    """A Handler returning the list of connected users"""
+
+    @web.authenticated
+    @gen.coroutine
+    def get(self):
+        """GET query returns the list of connected users"""
+        yield json.dumps([])
+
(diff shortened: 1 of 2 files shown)


jupyterlab/handlers/user.py
```diff
@@ -0,0 +1,26 @@
...
+    """A Handler returning the list of connected users"""
+
+    @web.authenticated
+    @gen.coroutine
+    def get(self):
+        """GET query returns the list of connected users"""
+        yield json.dumps([])
+
+# The path for lab extensions handler.
+extensions_handler_path = r"/lab/api/extensions"
```
davidbrochart, line 26: This file doesn't seem to be used?

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -52,11 +51,11 @@ async def get(self, *args, **kwargs):
...
         return await super().get(*args, **kwargs)
 
-    def open(self, guid):
-        #print("[YJSEchoWS]: open", guid)
+    def open(self, room_id):
+        #print("[YJSEchoWS]: open", room_id)
         cls = self.__class__
-        self.id = str(uuid.uuid4())
-        self.room_id = guid
+        self.id = self.get_current_user()
```
davidbrochart, line 57: I'm wondering how this plays with the same user opening the same notebook in two different browsers. We still need Yjs to see them as two different peers, right? If this is true, then we should keep an UUID for `self.id`.

davidbrochart: I just checked that if the same value is given to `self.id`, opening the same notebook in two different browsers breaks the collaborative mode (the cursor of the other user doesn't display anymore).

Issue #11735, opened 2021-12-22
Simple mode by default


mlucool:
### Problem

Simple mode gives users a notebook like experience with lab features. Some users prefer this and would like to always use simple mode. Can we have a setting which defaults new workspaces? E.g. https://my-instance.example.com/lab/workspaces/auto-Z?reset starts in simple mode.

### Proposed Solution

Add a new user setting that allows a user to default to simple mode.

jtpio:
Thanks @mlucool for the suggestion.

At the moment one way to default to simple mode at startup is to redirect users to the `/doc` endpoint. Maybe that's good enough in most cases.

Otherwise there could indeed be a setting that would be activated on startup, and trigger the switch to single mode if users land on `/lab`.

PR #11881, opened 2022-01-15
Add startMode setting to define the startup mode

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11735

## Code changes

Add a `startWithSimpleMode` setting.

## User-facing changes

Untitled

None.

### DIFF
--- a/packages/statusbar-extension/schema/plugin.json
+++ b/packages/statusbar-extension/schema/plugin.json
@@ -30,6 +30,12 @@
       "title": "Status Bar Visibility",
       "description": "Whether to show status bar or not",
       "default": true
+    },
+    "startWithSimpleMode": {
+      "type": "boolean",
+      "title": "Start with simple mode",
+      "description": "Whether to start with the simple mode or not",
+      "default": false
     }
   },
   "additionalProperties": false,
--- a/packages/statusbar-extension/schema/plugin.json
+++ b/packages/statusbar-extension/schema/plugin.json
@@ -31,11 +31,11 @@
       "description": "Whether to show status bar or not",
       "default": true
     },
-    "startWithSimpleMode": {
-      "type": "boolean",
-      "title": "Start with simple mode",
-      "description": "Whether to start with the simple mode or not",
-      "default": false
+    "startMode": {
+      "type": "string",
+      "title": "Start mode: ``, `single`or `multiple`",
+      "description": "The mode under which JupyterLab should start. If empty, the mode will be imposed by the URL",
+      "default": ""
     }
   },
   "additionalProperties": false,
--- a/packages/statusbar-extension/schema/plugin.json
+++ b/packages/statusbar-extension/schema/plugin.json
@@ -32,7 +32,7 @@
       "default": true
     },
     "startMode": {
-      "type": "string",
+      "enum": ["", "single", "multiple"],
       "title": "Start mode: ``, `single`or `multiple`",
       "description": "The mode under which JupyterLab should start. If empty, the mode will be imposed by the URL",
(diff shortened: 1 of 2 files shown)


packages/statusbar-extension/src/index.ts
```diff
@@ -355,6 +357,29 @@ const modeSwitch: JupyterFrontEndPlugin<void> = {
...
+          shell.mode = 'single-document';
+        }
+      };
+
+      Promise.all([loadSettings, app.restored])
+        .then(([settings]) => {
+          updateSettings(settings);
+          settings.changed.connect(settings => {
+            updateSettings(settings);
+          });
```
fcollonval, line 376: This is a special settings that acts once. So do we want to react to a change? What I mean is, if the user is in multi documents mode and turn the setting on, it will be switched to single document. But that may not be the intent.

echarles: agree. I have changed that setting and the visible mode was toggled. As the name is `startWithSimpleMode` it should only be applied on start. I will remove that update block.

fcollonval (changes requested):

packages/statusbar-extension/src/index.ts
```diff
@@ -355,6 +357,26 @@ const modeSwitch: JupyterFrontEndPlugin<void> = {
...
     });
+
+    if (settingRegistry) {
+      const loadSettings = settingRegistry.load(STATUSBAR_PLUGIN_ID);
+      const updateSettings = (settings: ISettingRegistry.ISettings): void => {
+        const startWithSimpleMode = settings.get('startWithSimpleMode')
+          .composite as boolean;
+        shell.mode = startWithSimpleMode
+          ? 'single-document'
+          : 'multiple-document';
```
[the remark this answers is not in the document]
echarles: I can try again with that option, but this was generating some issues (difficult to explain here) related to toogle state not reflecting the option.

However, I was thinking that instead of a boolean, we could go with `startingMode` which could be blank (then the url or whatever latest state comes in the picture), `single` or `multiple` in which case the current code would act.

If you like this approach, I can give it a try.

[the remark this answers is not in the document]
echarles: My last implements implements that approach which works fine for me. Use can force the startup mode via setting, or if that setting is blank, the URL enforce the startup mode.

packages/statusbar-extension/schema/plugin.json
```diff
@@ -30,6 +30,12 @@
       "title": "Status Bar Visibility",
       "description": "Whether to show status bar or not",
       "default": true
+    },
+    "startMode": {
+      "type": "string",
```
fcollonval, line 35: Let's use an enumeration.

```suggestion
      "enum": ["", "single", "multiple"],
```

fcollonval (commented):
Thanks the last implementation with the three values is great. I have a suggestion to improve the settings.

echarles:
The CI failure on js-apputils is not relevant for this PR and is tracked in https://github.com/jupyterlab/jupyterlab/pull/11887.

fcollonval (approved):
Thanks a lot @echarles 

> Do you want this one to be backported to 3.3?

echarles:

echarles:
Thx for the reviews @fcollonval I have merged and asked meeseeksdev to backport (tracked in https://github.com/jupyterlab/jupyterlab/pull/11881, I will merge it if CI is green)

I have also updated https://github.com/jupyterlab/jupyterlab/issues/11643 to add this feature (as a few pending others).

Issue #11700, opened 2021-12-15
Autosaving documents stacks duplicates file changed dialogs


mlucool:
## Description

If you have a file open in two browser tabs and edit a file, autosave in the non-edited tab will create a File Changed popup. These continue to stack every time it autosaves. If you come back to a tab after a while (e.g. overnight) you could end up with 100 copies of this.

## Reproduce
1. Create a new notebook and save it (tab1)
2. In a new browser tab open it and change something in it then save (tab2)
3. Go back to tab1 and add a character to a cell but don't save. Now wait for at least a couple autosaves (pro tip: change `autosaveInterval` in settings to see this happen faster)
4. Cancel on the File Changed dialog and you'll see many more copies of this

![autosave](https://user-images.githubusercontent.com/1813603/146251609-74cf150c-1770-456f-a656-e8df664f70b2.gif)

## Expected behavior

At most one file changed dialog per notebook.

## Context

Lab@3.2.3

PR #11880, opened 2022-01-15
Add side-by-side margin override in the notebookConfig

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11719

## Code changes

Add 2 settings in the notebookConfig

- sideBySideLeftMarginOverride
- sideBySideRightMarginOverride

## User-facing changes

The user can define its own margins when rendering in side-by-side.

## Backwards-incompatible changes

None.

PR #11883, opened 2022-01-15
ensure a single modal is opened in case of time conflict savings

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11700

## Code changes

Add a `timeConflictModalIsOpen` flag in the context.

## User-facing changes

Only one modal is opened and they don't stack anymore as described on https://github.com/jupyterlab/jupyterlab/issues/11700#issue-1081421554

None.

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -808,6 +808,11 @@ export class Context<
         `while the current file seems to have been saved ` +
         `${tDisk}`
     );
+    if (this._timeConflictModalIsOpen) {
+      return new Promise(() => {
+        return;
+      });
+    }
     const body = this._trans.__(
       `"%1" has changed on disk since the last time it was opened or saved.
 Do you want to overwrite the file on disk with the version open here,
@@ -818,11 +823,13 @@ or load the version on disk (revert)?`,
     const overwriteBtn = Dialog.warnButton({
       label: this._trans.__('Overwrite')
     });
+    this._timeConflictModalIsOpen = true;
     return showDialog({
       title: this._trans.__('File Changed'),
       body,
       buttons: [Dialog.cancelButton(), revertBtn, overwriteBtn]
     }).then(result => {
+      this._timeConflictModalIsOpen = false;
       if (this.isDisposed) {
         return Promise.reject(new Error('Disposed'));
       }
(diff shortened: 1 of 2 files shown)


echarles:
The Linux JS Flaky Tests fails with the following, like on https://github.com/jupyterlab/jupyterlab/pull/11880. I don't think this is related to this PR

```
test/sessioncontext.spec.ts
  @jupyterlab/apputils
    SessionContext
      #changeKernel()
        ✕ should handle multiple requests
```

echarles:
Apart from the expected failures, CI is happy. This is ready for review.

packages/docregistry/src/context.ts
```diff
@@ -808,6 +808,11 @@ export class Context<
         `while the current file seems to have been saved ` +
         `${tDisk}`
     );
+    if (this._timeConflictModalIsOpen) {
+      return new Promise(() => {
+        return;
+      });
```
krassowski, line 814: This creates a promise which will never get resolved. This means that the code waiting for the `_timeConflict` promise will be stuck forever. I feel uneasy about this because it feels like an undefined behaviour: will it be garbage collected? maybe. Will it contribute to performance problems? maybe. I may be wrong - I just could not find a specification for this.

Would it be safer to just reject here with a special error message (and handle it in `_maybeSave`, or even better propagate it to the auto-save logic)?

echarles: Thx for the review @krassowski. I have pushed a change that uses now `return Promise.reject(new Error('Modal is already displayed'));` like used in the rest of the code file.

krassowski: Should we add handling of this message to the exception handling in context's `save` (https://github.com/jupyterlab/jupyterlab/blob/e671140fe7ed03b53de09e935cf8ac399e939cae/packages/docmanager/src/savehandler.ts#L136-L140) or autosave's `save` (https://github.com/jupyterlab/jupyterlab/blob/6d147b45ff340aa2517927ffce6e9761c3fcff38/packages/docregistry/src/context.ts#L634-L639) (btw it looks like the check in autosave's `save` is redundant)?

echarles: Makes sense. I have added that condition in those 2 catch blocks.

fcollonval (approved):
LGTM

echarles:
Thx for the approval @fcollonval. I will wait a bit more in case @krassowski like to shim in and confirms he has no more comment.

krassowski:
LGTM, thanks!

echarles:

Issue #11612, opened 2021-12-03
Question: remove classic notebook server traitlets shim?


Zsailer:
In JupyterLab 3.x, we switched from the classic Notebook Tornado server to Jupyter Server's Tornado server. When doing this, the underlying traitlets `Configurable` object that holds the server's traits changed names (i.e. `NotebookApp` to  `ServerApp`). Anyone who knows traitlets also knows that this is kind of transition is a pain, because it forces all users to change the trait prefixes in all of their configuration files. 

To help smooth this transition, we added a [shim layer](https://github.com/jupyterlab/jupyterlab/blob/295443beccdc1b0d19834e1ae60a13733f4e655b/jupyterlab/labapp.py#L478) in JupyterLab 3.x (coming from nbclassic) that *did its best* to automatically port traits with the wrong prefix name to the correct object. This layer raises a bunch of warnings every time it shims a trait. 

**Now that we're nearing a 4.x release, should we remove this shim layer?**

Removing this layer requires a major release. The question is should we remove it in 4.x or 5.x.

jtpio:
Thanks @Zsailer for opening this :+1: 

I would be in favor of dropping it for 4.0, which is most likely still a few months away before being released.

We can communicate about breaking changes early enough so folks have time to process them and update their config, ideally early enough in the beta phase.

jtpio:
I opened https://github.com/jupyterlab/jupyterlab/pull/11894 to consider removing it for `4.0`.

echarles:
What does it cost to keep it as it is as long as possible? nbclassic has done a great work for some extensions I have been using. Not sure if they are migrated now.

If there is a cost to maintain nbclassic, I guess a scan of server extensions that are not migrated would help to take a decision.

jtpio:
Linking to https://github.com/jupyterlab/team-compass/issues/133#issuecomment-989018555, https://github.com/jupyterlab/nbclassic/issues/81 and related issues for more context.

One of the main reasons for this would be to allow installing JupyterLab 4.0 with Notebook 7 as part of the Notebook 7 JEP: https://jupyter.org/enhancement-proposals/79-notebook-v7/notebook-v7.html.

At the moment `nbclassic` has `notebook<7` as a dependency: https://github.com/jupyterlab/nbclassic/blob/623b8fa136031b0d32aee0385327df82f0060ac5/setup.cfg#L30

PR #11894, opened 2022-01-18
Drop `nbclassic`, depend on `notebook_shim`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11612

Keeping as a draft for now as this is still being discussed.

## Code changes

- [x] Removed `nbclassic` as a dependency
- [x] Remove the "Launch Classic Notebook" menu entry and command. This functionality can be added back from the Notebook v7 repo as a prebuilt JupyterLab extension.
- [x] Depend on https://github.com/jupyterlab/notebook_shim for the server shims

## User-facing changes

Stock install of `jupyterlab` will not bring `nbclassic` and the classic notebook (v6) interface anymore.

None w.r.t to JupyterLab API.

However this is a big enough change that should carefully be documented.


jtpio:
One of the side-effects of this is `ipykernel` not being installed by default anymore:

image

jtpio:
TODO: depend on https://github.com/jupyterlab/notebook_shim for the server shims

mlucool:
Will nbclassic be able to be optionally installed? Especially due to performance concerns, many users still prefer classic.

jtpio:
> Especially due to performance concerns

By performance concerns do you mean performance issues when rendering a notebook in JupyterLab? Is there an issue that we could link to for reference?

echarles:
@jtpio A user installing notebook@7 (based on retrolab), will not be able to install notebook@6.

My understanding is that a user installing jupyterlab@4 (and later on jupyterlab@5...) will be able to install notebook@6 and will be able to switch from `/lab` to `/tree`. Is this correct?

jtpio:
Right, Notebook v7 is just another major version of `notebook`. Like any other package, users have to choose which version they want to install.

Currently `nbclassic` depends on `notebook<7`, which is why this PR drops `nbclassic` and will depend on `notebook_shim` (https://github.com/jupyterlab/notebook_shim) instead.

jtpio:
>  Remove the "Launch Classic Notebook" menu entry and command

Done in https://github.com/jupyterlab/jupyterlab/pull/11894/commits/5ae4e594e66901eb6441ee9ca83f4e62e8cb1741.

Instead this functionality can be added back from Notebook v7 as a JupyterLab prebuilt extension if needed.

jtpio:
> One of the side-effects of this is `ipykernel` not being installed by default anymore:

Before we had `jupyterlab` -> `nbclassic` -> `notebook` -> `ipykernel`, with the dependency on `ipykernel` defined here:

https://github.com/jupyter/notebook/blob/26626343384195a1f4f5461ba42eb3e133655976/setup.py#L125

image

Just like the comment says above, maybe we'll still want to bless `ipykernel` for a while? In that case we can add it as an explicit dependency, like in this PR: https://github.com/jupyterlab/jupyterlab/pull/11334

jtpio:
> This way we don't block the first `7.0.0ax` `notebook` prerelease. There is still time to add `notebook_shim` as a dependency to `jupyterlab` before `4.0` is released.

This will make it possible to make a new `4.0.0ax` prerelease for `jupyterlab` that dropped the `nbclassic` dependency, and update Notebook 7 to use the prerelease and be able to move forward.

Zsailer:
Hm, I would prefer to get notebook_shim released and added as a dependency in this PR so there is not change in behavior for users. If we temporarily drop it, there's a regression for users who might still have traits living in old config files.

jtpio:
OK then the next step would be to work on making a new `notebook_shim` release and finish https://github.com/jupyterlab/notebook_shim/pull/4

So we can unblock this PR and the work on Notebook v7 and its first prerelease.

jtpio:
Thanks @Zsailer!

Zsailer:
It looks like my commits to your branch adding notebook_shim as a dependency here were removed? Was that intentional?

Zsailer:
I'm guessing this was dropped by accident after a rebase on master. I went ahead and pushed this commit again. 😃

jtpio:
@Zsailer I think they were dropped by mistake yes. I probably forgot to sync the latest state of the branch before rebasing. Thanks for catching this!

Zsailer (approved):

Zsailer:
This looks good to me :+1: :rocket:

echarles:
As stated in the initial comment, replace `nbclassic` by `notebook_shim`. The result will be that `notebook` being previously pulled by `nbclassic` as transitive dependency will not be pulled anymore.

I have test on my laptop, and jupyterlab works fine.

I have tested on binder and it also works fine (jupyterlab and jupyterhub startup config have been historically touchy). Lab is launched, and classic notebook is available under `tree` which is prolly linked to the binder preRequisites (should this change?)
 
Coming back to the question on https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1023482506 replied on https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1023499441: I have installed on my laptop `nbclassic` and `notebook` on top of this branch. When I launch `jupyter lab` or `jupyter nbclassic` or `jupyter notebook` and I go to `http://localhost:8888/tree`, I receive a blank page (with `ServerApp] 404 GET /static/notebook/components/...`.

Is this the expected behavior with this PR? Some users will still need to let JupyterLab and the classic Notebook coexist.

Zsailer:
> Is this the expected behavior with this PR? Some users will still need to let JupyterLab and the classic Notebook coexist.

No, this is not the expected behavior. Let me look into this a little further.

You should be able to install JupyterLab 4.x, Notebook 6.x, and NBClassic with no issues and run classic Notebook side-by-side with JLab 4.x.

As soon as you upgrade to Notebook 7.x, you should get a pip dependency warning saying you can't run nbclassic (and thus, the classic notebook experience) anymore.

echarles:
I have quickly tested after Zach push and before our weekly meeting. I typically double and triple check, which I had not the chance to. Can someone test on his local env?

jtpio:
@echarles I just tried locally in a fresh new environment.

```
python -m pip install -e . && jlpm install && jlpm run build && jlpm run build:core && jupyter lab build
python -m pip install notebook
jupyter serverextension enable jupyterlab
```

Then the server must be started with `jupyter notebook` (and not `jupyter lab`).

Since we drop `nbclassic`, the Jupyter Server extension to serve the notebook frontend is then also dropped.

Probably it works on Binder because the environment has `nbclassic` in it.

echarles:
> Then the server must be started with jupyter notebook (and not jupyter lab).

Launching will `jupyter notebook` will not work for jupyterlab (the user will not be able to choose `/lab` or `/tree` like now.

I think the goal is to serve from the same running server process both `/lab` and `/tree`.

jtpio:
Right, but then you can still install `nbclassic` to get the jupyter server extension, which actually also brings `notebook<7`.

https://user-images.githubusercontent.com/591645/155372489-65ec10bd-7911-416d-baf9-0f6e8d8dcf1a.mp4

jtpio:
So as a summary of this testing session locally:

- Installing JupyterLab 4.0 alone will not serve Classic Jupyter Notebook v6
- Install `nbclassic` with JupyterLab 4 will bring `notebook<7`, and serve the classic notebook v6 just like before with JupyterLab 3
- But if you install JupyterLab 4.0 and Notebook v7, then it will not be possible to install `nbclassic` since it depends on `notebook<7`. Or maybe it will be but the environment might be broken

Zsailer:
> * Installing JupyterLab 4.0 alone will not serve Classic Jupyter Notebook v6
> * Install nbclassic with JupyterLab 4 will bring notebook<7, and serve the classic notebook v6 just like before with JupyterLab 3
> * But if you install JupyterLab 4.0 and Notebook v7, then it will not be possible to install nbclassic since it depends on notebook<7. Or maybe it will be but the environment might be broken

Yes, this is the expected behavior. Thank you for summarizing Jeremy.

@echarles, keep in mind, this PR drops the menu items in JupyterLab that redirects you to the classic notebook UI. Users will just have to manually change the URL to see the notebook UI. You can, of course, create a small JupyterLab extension to add these menu items back for your usecase.

echarles:
> jupyter serverextension enable jupyterlab
> Then the server must be started with jupyter notebook (and not jupyter lab).

This has done the job. Launching with notebook with preliminary jupyterlab extension enabled worked.

Thx for your support on this.

jtpio:
> You can, of course, create a small JupyterLab extension to add these menu items back for your usecase.

There is a chance Notebook v7 will add it back as a JupyterLab prebuilt extension, installed automatically when a user installs `notebook>=7`. In fact RetroLab currently already adds some menu items to do this interface switch more easily:

image

Zsailer:
This is ready to merge IMO. Thanks for leading this, @jtpio!

jtpio:
> With the need to launch via `jupyter notebook`, this sounds like it will not be possible anymore. Any input on that?

It should be possible, if `nbclassic` is installed and the server started with `jupyter lab` (https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049021393)

echarles:
I have quickly tested and the settings seems to be accepted. However the JupyterLab flag are not as expected eg `jupyter notebook --expose-app-in-browser`.

Let me look more at that tomorrow to be sure the CLI settings are OK.

WDYT about not supporting anymore the flags. Actually, we are reverting the mecansim. Before lab was the primary, and notebook the fallback, and it worked fine as Lab was a superset of classic in terms of conf.. With this PR, this is the reverse.

jtpio:
> I have quickly tested and the settings seems to be accepted. However the JupyterLab flag are not as expected eg jupyter notebook --expose-app-in-browser.

Not sure this was supposed to work before, since `-expose-app-in-browser` is a JupyterLab flag?

echarles:
> Not sure this was supposed to work before, since -expose-app-in-browser is a JupyterLab flag?

That is my point. Users willing to migrate to JupyterLab 4 while still offering Notebook V6 will have to forget the jupyterlab flag.

jtpio:
Well they should still be able to do:

```
pip install jupyterlab==4
pip install nbclassic

jupyter lab --expose-app-in-browser
```

And they have JupyterLab with the flags and classic notebook 6 like before.

echarles:
> Well they should still be able to do:
> pip install jupyterlab==4
> pip install nbclassic
> jupyter lab --expose-app-in-browser
> And they have JupyterLab with the flags and classic notebook 6 like before.

As discussed earlier, starting with `jupyter lab` does not work for me for the `/tree` endpoint.

To recap, on my env, I have built this branch (which install  `notebook_shim`), then manually installed `nbclassic` and `notebook`.

- `jupyter lab` works fine for `/lab` and returns 404 when hitting `/tree`.
- `jupyter notebook` works fine for both `/lab` and `/tree`. You can pass traits via CLI (e.g. --LabApp.collaborative=True) but the flag `--collaborative` breaks the launch (expected as this is a flag for lab, not for notebook).

I have also tried to change the default url to have lab show on startup instead of notebook with `---LabApp.default_url=/lab` without luck.

If my test are confirmed, it looks to me an important move to be discussed some deployments (the ones willing to upgrade to jupyterlab4 while keeping notebook v6) will have to change the scripts...

jtpio:
> I have quickly tested and the settings seems to be accepted. However the JupyterLab flag are not as expected eg `jupyter notebook --expose-app-in-browser`.

Just to confirm this, I went ahead and checked locally in a new environment with JupyterLab 3:

```
mamba create -n tmp jupyterlab=3 -y
mamba activate tmp
jupyter notebook --expose-app-in-browser
```

Gives the following error:

```
$ jupyter notebook --expose-app-in-browser
usage: jupyter-notebook [-h] [--debug] [--show-config] [--show-config-json] [--generate-config] [-y]
                        [--no-browser] [--no-mathjax] [--allow-root] [--autoreload] [--script]
                        [--no-script] [--log-level NotebookApp.log_level]
                        [--config NotebookApp.config_file] [--ip NotebookApp.ip] [--port NotebookApp.port]
                        [--port-retries NotebookApp.port_retries] [--sock NotebookApp.sock]
                        [--sock-mode NotebookApp.sock_mode] [--transport KernelManager.transport]
                        [--keyfile NotebookApp.keyfile] [--certfile NotebookApp.certfile]
                        [--client-ca NotebookApp.client_ca] [--notebook-dir NotebookApp.notebook_dir]
                        [--browser NotebookApp.browser] [--pylab [NotebookApp.pylab]]
                        [--gateway-url GatewayClient.url]
                        [extra_args ...]
jupyter-notebook: error: unrecognized arguments: --expose-app-in-browser
```

https://user-images.githubusercontent.com/591645/155477719-8e24f70f-6b63-42c0-8f30-8718c33e9713.mp4

So it seems using lab flags with the notebook entry point was already not supported before.

jtpio:
> To recap, on my env, I have built this branch (which install notebook_shim), then manually installed nbclassic and notebook.

> - jupyter lab works fine for /lab and returns 404 when hitting /tree.

This behavior is not meant to change with this PR. As stated in https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049026078, this should still work the same as with JupyterLab 3 like it is now.

Can you confirm the server extensions are correctly listed and activated in your local environment?

The output of  `jupyter server extension list` should look like the following:

```
Config dir: /home/jtp/miniforge3/envs/jupyterlab/etc/jupyter
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 4.0.0a19 OK
    nbclassic enabled
    - Validating nbclassic...
      nbclassic  OK
    notebook_shim enabled
    - Validating notebook_shim...
      notebook_shim  OK
```

echarles:
> So it seems using lab flags with the notebook entry point was already not supported before.

True, and with this PR you would be obliged to use the notebook entrypoint, with the result that the jupyterlab flags are not available anymore

> The output of jupyter server extension list should look like the following:

```
 jupyter server extension list
Config dir: /Users/echarles/.jupyter

Config dir: /Users/echarles/miniconda3/envs/datalayer/etc/jupyter
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 4.0.0a19 OK
    nbclassic enabled
    - Validating nbclassic...
      nbclassic  OK
    notebook_shim enabled
    - Validating notebook_shim...
      notebook_shim  OK

Config dir: /usr/local/etc/jupyter
```

jtpio:
> with this PR you would be obliged to use the notebook entrypoint to have jupyterlab and notebookV6 at the same time

That's not what happens in my local environment (and also on Binder), where both JupyterLab and Notebook v6 can still coexist fine just like before this PR, when `nbclassic` is installed (see summary posted above: https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049026078).

Users are not obliged to switch to the notebook entrypoint and can also launch `jupyter lab` if they want to. `nbclassic` takes care of adding the proper Jupyter Server extension. They can also launch `jupyter notebook` and in that case JupyterLab is served as a Classic Notebook Server extension.

> with the result that the jupyterlab flags are not available anymore

This comment suggests this was already the case before this PR: https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049569885

---

To put it another way: this PR doesn't change things that much. It just removes `nbclassic` from being installed by default. If users want notebook v6, they install `nbclassic` and it's just the same as before.

echarles:
> That's not what happens in my local environment (and also on Binder), where both JupyterLab and Notebook v6 can still coexist fine just like before this PR, when nbclassic is installed (see summary posted above: https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049026078).

Then something must be wrong on my env. I have built from scratch again, enabled with `jupyter server extension enable jupyterlab` and `jupyter server extension enable jupyterlab` (note the subtle difference, `server extension` and `serverextension`), and still no luck with I start with `jupyter lab`: `/tree` is not available. Let's pause for now, I will try again later and maybe we can join on a call to see what is going on.

jtpio:
> note the subtle difference, `server extension` and `serverextension`)

Yes, and for `jupyter lab` it's `server extension` that is important (starting a Jupyter Server).

@echarles since this PR is a blocker for further development on Notebook v7, what do you think about opening a new issue to track this use cases? In the issue, we would then list all the various use cases we should make sure to cover.

We are still in the 4.0 alpha cycle. We can release and iterate. And make further fixes before the beta.

echarles:
@jtpio I have started from a completely blank env and now `/tree` and `/lab` work fine with `jupyter lab`. Strangely the steps you have given in https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049014680 (enable jupyterlab extension and launch with jupyter notebook) does not work anymore when you .

> @echarles since this PR is a blocker for further development on Notebook v7, what do you think about opening a new issue to track this use cases? In the issue, we would then list all the various use cases we should make sure to cover.
> We are still in the 4.0 alpha cycle. We can release and iterate. And make further fixes before the beta.

Sure, we can do merge this. I was just trying to make sure we were covered and did not face major regression difficult to catch-up.

echarles:
> Strangely the steps you have given in https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1049014680 (enable jupyterlab extension and launch with jupyter notebook) does not work anymore when you .

Actually if you run `jupyter serverextension enable jupyterlab` you can access `/lab` when you launch jupyter notebook.

Overall this is a bit confusing (`nbclassic` vs `notebook_shim`, `server extension` vs `serverextension`) and more documentation will be helpfull

echarles (approved):
Thx @jtpio and @Zsailer. This looks good and works on my env and binder. I have opened https://github.com/jupyterlab/jupyterlab/issues/12115 to follow-up with more doc on this.

jtpio:
Thanks @echarles for the extensive testing and your help finding potential issues :+1:

PR #11899, opened 2022-01-19
Fix plugin id for the toolbar registry plugin

jtpio:
## References

For consistency with the other `@jupyterlab` plugins.

Noticed when updating RetroLab to the latest `@jupyterlab` packages: https://github.com/jupyterlab/retrolab/pull/319

## Code changes

Rename plugin id.

## User-facing changes

None

None (4.0 still in alpha cycle)


fcollonval (approved):
Good catch! Thanks @jtpio. We will need to backport it to 3.3.x after https://github.com/jupyterlab/jupyterlab/pull/11873 is in.

jtpio:
> I added this as part of #11873 for 3.3

Sounds good, thanks :+1:

PR #11900, opened 2022-01-19
Add spacer widget via `app.shell` instead of `ILabShell`

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/11654

## Code changes

The `@jupyterlab/application-extension:top-spacer` plugin should not require a `ILabShell`, so it can be reused in downstream distributions.

## User-facing changes

None

None


jweill-aws (approved):
Changes look good! Thanks.

fcollonval (approved):
thanks @jweill-aws for reviewing

PR #11901, opened 2022-01-19
fix(docprovider): fix issue with empty notebook

entropitor:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/10962

## Code changes

The response of the initialContent request was only applied after the promise callbacks for the promise that is resolved on line 80. 

So the original order is:
1. Request received
2. Promise resolved
3. Triggering the [save](https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/docregistry/src/context.ts#L264-L270) (if there is any data in the intialContent) (and the [first thing this does](https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/docregistry/src/context.ts#L597-L605) is getting the content as known at that time)
4. Updated applied in the setTimeout

By dropping the setTimeout, we first apply the initialContent before we resolve the `_initialContentRequest`.

Which basically means that IF there is data in the intialContent, a save happens of the current state without the actual initial content applied. Basically relying on the fact that before [initialize](https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/docregistry/src/context.ts#L262) is called, the sync is already started and hopefully finished, if however it isn't, the notebook ends up being saved as an empty notebook without cells and if multiple people are joining at the same time, this update is also propagated over RTC

You can replicate by dropping / delaying some of the yjs sync messages on the client and then refreshing 2 different clients at almost the same time.
```typescript
    const originalHandler = this.messageHandlers[0];
    this.messageHandlers[0] = (
      encoder,
      decoder,
      provider,
      emitSynced,
      messageType,
    ) => {
      const type = decoding.peekVarInt(decoder);
      if (type === 2) {
        // drop updates
        return;
      }
      if (type === 1) {
        // make applying syncstep 2 slow
        setTimeout(() => {
          originalHandler(encoder, decoder, provider, emitSynced, messageType);

          if (encoding.length(encoder) > 1) {
            this.ws?.send(encoding.toUint8Array(encoder));
          }
        }, 2000);
      } else {
        // normal syncstep 1
        originalHandler(encoder, decoder, provider, emitSynced, messageType);
      }
    };
```

## User-facing changes

None other than no empty notebooks

None

### DIFF
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -70,9 +70,7 @@ export class WebSocketProviderWithLocks
       const initialContent = decoding.readTailAsUint8Array(decoder);
       // Apply data from server
       if (initialContent.byteLength > 0) {
-        setTimeout(() => {
-          Y.applyUpdate(this.doc, initialContent);
-        }, 0);
+        Y.applyUpdate(this.doc, initialContent);
       }
       const initialContentRequest = this._initialContentRequest;
       this._initialContentRequest = null;


hbcarlos:
Hi @entropitor! Thanks for opening the PR, LGTM!

hbcarlos:
The failing test doesn't seem to be related to this PR.

jtpio:
Thanks @entropitor!

jtpio:
@hbcarlos do we know if there was a reason for having the `setTimeout` in the first place?

hbcarlos:
Not sure, @dmonad wrote that part. We reviewed the PR yesterday in the RTC meeting. It is OK to merge.
We did not merge it because we don't have commits right.

dmonad (approved):

jtpio:
Thanks all!

jtpio:

entropitor:
@dmonad 

> However, I don't think that this will fix anything. Before we save or reset a file, we always acquire a lock from the server. During this time the setTimeout will already be fired.

This is not true, see https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/docregistry/src/context.ts#L262. The thing is that you already request the lock before you request init content and so when you receive init content, you have the lock

The bug was as follows:

1. initialize is called
2. Lock is requested
3. ....
4. lock is acquired
5. init content is requested
6. ...
7. init content is returned
8. promise resolved
9. save
10. init content applied

Usually however, before 7 the provider was already synced through the sync protocol so we didn't notice this. With this fix in place, 10 happens before 8 and 9 and so it's also okay.

With the code in the PR description I was able to consistently reproduce the issue and this code fixes the issue.

WilliamJudge94:
Does anyone know when this will be added to the PyPi version of jupyterlab? I am still getting random file deletions during RTC.

jtpio:
@WilliamJudge94 this should be available in JupyterLab 3.3.0rc0: https://pypi.org/project/jupyterlab/3.3.0rc0

Which can be installed with:

```
pip install -U --pre "jupyterlab==3.3.0rc0"
```

3.3 final will be released in the coming days.

PR #11904, opened 2022-01-20
Updates code, documentation to use new standard terms for cell output

jweill-aws:
Updates code and documentation to use new standard terms: an *output* is one of potentially multiple results of running a cell, a *cell output* is the collection of all outputs for one cell, and *all cell output* is the collection of all outputs for all cells.

## References

#11878 

## Code changes

Updates menus (string changes only) using updated terms, adding captions where they can add clarify

## User-facing changes

Updates strings in menus — no placement changes

None


docs/source/developer/contributing.rst
```diff
@@ -908,17 +908,17 @@ Or with ``jlpm``:
...
    The author of the documentation is not sitting next to the user, so
    using "we" can lead to frustration when things don't work as
    expected.
 -  Avoid words that trivialize using JupyterLab such as "simply" or
    "just." Tasks that developers find simple or easy may not be for
    users.
--  Write in the active tense, so "drag the notebook cells..." rather
-   than "notebook cells can be dragged..."
--  The beginning of each section should begin with a short (1-2
+-  Write in the active tense. Use "drag the notebook cells…" rather
```
3coins, line 919: Rather than breaking the sentence, would it make sense to use e.g.,?

Write in the active tense e.g., use "drag the notebook cells.." rather than "notebook cells can be dragged..."

docs/source/developer/contributing.rst
```diff
@@ -908,17 +908,17 @@ Or with ``jlpm``:
...
+   than "notebook cells can be dragged…".
+-  The beginning of each section should begin with a short (1–2
```
3coins, line 921: Seems like an unintended long dash was added here instead of a hyphen.

docs/source/developer/contributing.rst
```diff
@@ -930,36 +930,45 @@ User Interface Naming Conventions
...
 """""""""""""
 
--  The generic content area of a tabbed UI is a panel, but prefer to
-   refer to the more specific name, such as “File browser.” Tab bars
-   have tabs which toggle panels.
--  The menu bar contains menu items, which have their own submenus.
--  The main work area can be referred to as the work area when the name
+-  The generic content area of a tabbed UI is a *panel*. Refer to a panel
+   by its most specific name, such as “File browser.” *Tab bars*
+   have *tabs* which let a user view different panels.
```
3coins, line 964: Use "that" instead of "which".

packages/notebook-extension/src/index.ts
```diff
@@ -1718,7 +1718,8 @@ function addCommands(
     isEnabled
   });
   commands.addCommand(CommandIDs.restartClear, {
-    label: trans.__('Restart Kernel and Clear All Outputs…'),
+    label: trans.__('Restart Kernel and Clear All Cell Outputs…'),
```
3coins, line 1721: Is there any special meaning for the three dots at the end?

docs/source/developer/contributing.rst
```diff
@@ -930,36 +930,45 @@ User Interface Naming Conventions
...
+not related to a file, such as terminals, consoles or the inspector.
+
+Notebook Cells
+""""""""""""""
+
+A notebook contains *cells*, each of which have *input* and one or more 
+*outputs*. When the user runs a cell, the kernel reads and executes the
+input and generates outputs. The notebook then displays the cell's output.
+The term *output* describes one of possibly multiple results of running a
+cell. The term "all output" is ambiguous; it could refer either to all
```
3coins, line 954: Is there any reference to "all output" anywhere else in this doc? 

It seems like there is a recommendation to use "cell output" or "all cell outputs", so referring to "all outputs" seems redundant.

docs/source/developer/contributing.rst
```diff
@@ -975,8 +984,8 @@ The majority of names are written in lower case. These names include:
 -  cell inspector
 -  code console
```
3coins, line 985: Is there an indentation mismatch here?

packages/notebook-extension/src/index.ts
krassowski: Good point! I think that three dots make sense for "Change Kernel..." because it indicates that a pop-up dialog with more options will be shown (and for "Go to line..." and for "Find..." as all these require additional user action), but it indeed does not seem necessary here (or for any other "Restart Kernel" option).

I think we should document that ellipsis should be only added when there is a follow up choice/action need from the user.

docs/source/developer/contributing.rst
jweill-aws: I prefer "For example" to "e.g." for clarity. Microsoft's style guide also prefers "for example" or "examples": https://docs.microsoft.com/en-us/style-guide/procedures-instructions/writing-step-by-step-instructions

docs/source/developer/contributing.rst
jweill-aws: That's an [en dash](https://en.wikipedia.org/wiki/Dash#En_dash), which indicates a range of values among other purposes.

packages/notebook-extension/src/index.ts
jweill-aws: Added a section to the style guide on command names.

docs/source/developer/contributing.rst
jweill-aws: Is there? I see each line starting with `-  ` (hyphen and two spaces)

docs/source/developer/contributing.rst
jweill-aws: This comment is making me want to use `Outputs of All Cells` instead: https://github.com/jupyterlab/jupyterlab/issues/11878#issuecomment-1018817833

docs/source/developer/contributing.rst
3coins: I checked out the file and it looks fine. Looks good.

packages/notebook-extension/src/index.ts
```diff
@@ -1718,7 +1718,8 @@ function addCommands(
     isEnabled
   });
   commands.addCommand(CommandIDs.restartClear, {
-    label: trans.__('Restart Kernel and Clear All Outputs…'),
+    label: trans.__('Restart Kernel and Clear Outputs of All Cells…'),
```
3coins, line 1721: Was there a consensus on the recommendation from @krassowski about removing the ellipsis? Here is the original comment:

>I think that three dots make sense for "Change Kernel..." because it indicates that a pop-up dialog with more options will be shown (and for "Go to line..." and for "Find..." as all these require additional user action), but it indeed does not seem necessary here (or for any other "Restart Kernel" option).

jweill-aws: From the [Apple Human Interface Guidelines on Menu Anatomy](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/), "Use an ellipsis whenever choosing a menu item requires additional input from the user. The ellipsis character (…) means a dialog or separate window will open and prompt the user for additional information or to make a choice."

If the kernel is running, all "Restart Kernel" options will spawn a dialog box asking for confirmation (Restart) or to abort (Cancel). The dialog box also provides additional information, that all variables will be lost, that may help a user make a more informed decision.

The ellipsis also conveys to the user that it is safe for them to click the menu item even if they're not totally sure what it will do, as they should expect to see a dialog box appear. This encourages users to explore the UI and to discover what all the different menu items do.

3coins: Sounds good. Thanks for explaining.

3coins (approved):
Ship it! 🚢

fcollonval (approved):
Thanks for improving the doc

PR #11907, opened 2022-01-21
Remove unused `isInit` param in `sessionContext.changeKernel()`

jtpio:
## References

Remove the unused `isInit` param in the private `sessionContext.changeKernel()` method.

Caught while looking at the code.

## Code changes

Remove unused parameter.

## User-facing changes

None

None since `changeKernel` is `private`.


fcollonval (approved):
Thanks @jtpio

Issue #11911, opened 2022-01-21
Explicitly specify issue triage label


jweill-aws:
### Problem

Per comments in  https://github.com/jupyterlab/jupyterlab-probot/pull/11 (https://github.com/jupyterlab/jupyterlab-probot/pull/11#issuecomment-1018725789) the bot is tagging all issues with `status:Needs Triage`, even issues in repositories without a triage process.

### Proposed Solution

Explicitly specify the triage label so that jupyterlab-probot can avoid using a default value for this property. Repos without a triage label will not be tagged with a default triage label.

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/11912

PR #11912, opened 2022-01-21
Explicitly specifies triageLabel parameter

jweill-aws:
## References

Addresses #11911 by explicitly specifying the triageLabel option.

Right now this option has a default value in `jupyterlab-probot`, but I plan to remove this default value.

## Code changes

Affects `jupyterlab-probot` config only.

## User-facing changes

None

None


fcollonval (approved):
Thanks @jweill-aws

PR #11812, opened 2022-01-10
Update `vscode-debugprotocol` to `@vscode/debugprotocol`

jtpio:
## References

As noticed when installing dependencies locally.

The `vscode-debugprotocol` package has been renamed to `@vscode/debugprotocol`: https://www.npmjs.com/package/@vscode/debugprotocol

## Code changes

Update dependency.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #11923, opened 2022-01-24
Backport PR #11752: Pause on exception

andrewfulton9:
## References

Backport PR #11752

### DIFF
--- a/packages/debugger/package.json
+++ b/packages/debugger/package.json
@@ -71,6 +71,7 @@
     "@lumino/messaging": "^1.4.3",
     "@lumino/signaling": "^1.4.3",
     "@lumino/widgets": "^1.19.0",
+    "@vscode/debugprotocol": "~1.51.0",
     "codemirror": "~5.61.0",
     "react": "^17.0.1",
     "vscode-debugprotocol": "^1.37.0"
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -517,9 +517,15 @@ const sidebar: JupyterFrontEndPlugin<IDebugger.ISidebar> = {
       evaluate: CommandIDs.evaluate
     };
 
+    const breakpointsCommands = {
+      registry: commands,
+      pause: CommandIDs.pause
+    };
+
     const sidebar = new Debugger.Sidebar({
       service,
       callstackCommands,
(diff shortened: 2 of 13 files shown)


fcollonval (commented):
Thanks @andrewfulton9 the linter and js-debugger errors are legit. Would you mind looking into those?

andrewfulton9:
@fcollonval, Im looking into these issues. I'll let you know when I think I have it fixed.

andrewfulton9:
@fcollonval, I think this is good to go. Do I need to push up the tsconfidoc.json too? It looks like the user and user-extension packages are added on my machine after running `jlpm run integrity`

fcollonval:
> Do I need to push up the tsconfidoc.json?

Indeed you don't. The integrity script is scanning your local directory tree. And unfortunately as the directories are not tracked by git, when switching branches it finds non existing extensions.

fcollonval (approved):

packages/debugger/package.json
```diff
@@ -71,6 +71,7 @@
     "@lumino/messaging": "^1.4.3",
     "@lumino/signaling": "^1.4.3",
     "@lumino/widgets": "^1.19.0",
+    "@vscode/debugprotocol": "~1.51.0",
```
jtpio, line 74: Looks like this adds `@vscode/debugprotocol` in addition to `vscode-debugprotocol` defined a few lines below.

Maybe we can clean that up to only depend on the new `@vscode/debugprotocol`, as it is the case on `master` now:

https://github.com/jupyterlab/jupyterlab/blob/1903d0dd09b2415a998dd46de15b54a5c8988db5/packages/debugger/package.json#L74

And switch to `^1.51.0` instead of `~1.51.0`

jtpio: Or if `@vscode/debugprotocol` is not used on the `3.3.x` branch, then we can just remove it here and in the `yarn.lock`.

jtpio: For reference the update was done in https://github.com/jupyterlab/jupyterlab/pull/11812

fcollonval: I saw it too. I'm on it :wink:

jtpio: Alright thanks!

Issue #11754, opened 2021-12-28
Check the application includes the plugin and is activated before reconciling settings


jtpio:
## Description

At the moment settings for a plugin seem to be picked up even if the plugin is not loaded on the page.

This also seems to be the case when a plugin is loaded but fails to activate.

## Reproduce

This can be reproduced by creating a new prebuilt extension with settings from the cookiecutter:

1. `cookiecutter https://github.com/jupyterlab/extension-cookiecutter-ts`
2. answer `y` to `has_settings`
3. do the dev setup
4. edit `schema/plugin.json` with the following:

```json
{
  "jupyter.lab.shortcuts": [
    {
      "command": "documentsearch:start",
      "keys": ["Accel F"],
      "selector": ".jp-mod-searchable",
      "disabled": true
    }
  ],
  "title": "myextension",
  "description": "myextension settings.",
  "type": "object",
  "properties": {},
  "additionalProperties": false
}
```
5. edit the plugin definition to `require` a token not available in JupyterLab, for example:

```ts
import {
  JupyterFrontEnd,
  JupyterFrontEndPlugin
} from '@jupyterlab/application';

import { IRetroShell } from '@retrolab/application';

import { ISettingRegistry } from '@jupyterlab/settingregistry';

/**
 * Initialization data for the myextension extension.
 */
const plugin: JupyterFrontEndPlugin<void> = {
  id: 'myextension:plugin',
  autoStart: true,
  requires: [IRetroShell],
  optional: [ISettingRegistry],
  activate: (app: JupyterFrontEnd, shell: IRetroShell, settingRegistry: ISettingRegistry | null) => {
    console.log('JupyterLab extension myextension is activated!');

    if (settingRegistry) {
      settingRegistry
        .load(plugin.id)
        .then(settings => {
          console.log('myextension settings loaded:', settings.composite);
        })
        .catch(reason => {
          console.error('Failed to load settings for myextension.', reason);
        });
    }
  }
};

export default plugin;
```
6. rebuild the extension and check the shortcuts in the Advanced Settings editor:

![image](https://user-images.githubusercontent.com/591645/147553340-742f9256-9636-4260-8a9b-f299f103794b.png)

## Expected behavior

The settings for a given plugin should not be picked up if the plugin is not loaded on the page or failed to activate.

Maybe the plugin should checked with `app.hasPlugin(pluginId)` before reconciling its settings?

## Context

[...]

PR #11938, opened 2022-01-26
Filter settings with `app.hasPlugin()`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11754.

The main motivation is to load the settings for the plugins that are loaded in the app only. This check is then performed on the frontend.

An alternative would be to filter the settings on the backend as discussed in https://github.com/jupyterlab/jupyterlab/issues/11754.

## Code changes

Check the `JupyterFrontEnd` application has the plugin before loading its settings.

**TODO**

- [x] Menus
- [x] Context menus
- [x] Toolbars
- [x] Shortcuts

## User-facing changes

Testing this locally with the dev install of JupyterLab:

- drop a `plugin.json` file in `packages/application-extension/schemas` with the following content:

```json
{
  "jupyter.lab.menus": {
    "context": [
      {
        "command": "help:about",
        "selector": "body",
        "rank": 0
      }
    ]
  },
  "jupyter.lab.toolbars": {
    "Notebook": [
      { "name": "save", "rank": 1000 }
    ]
  },
  "title": "myextension",
  "description": "myextension settings.",
  "type": "object",
  "properties": {},
  "additionalProperties": false
}
```

- at the moment there is no plugin named `@jupyterlab/application-extension:plugin` in JupyterLab
- run `jlpm run build`
- reload the page

**Before this PR (master)**

image

**After**

image

None

### DIFF
--- a/packages/csvviewer-extension/src/index.ts
+++ b/packages/csvviewer-extension/src/index.ts
@@ -116,13 +116,14 @@ function activateCsv(
     );
 
     if (settingRegistry) {
-      toolbarFactory = createToolbarFactory(
+      toolbarFactory = createToolbarFactory({
         toolbarRegistry,
         settingRegistry,
-        FACTORY_CSV,
-        csv.id,
-        translator
-      );
+        factoryName: FACTORY_CSV,
+        pluginId: csv.id,
+        translator,
+        isPluginAvailable: (pluginId: string) => app.hasPlugin(pluginId)
+      });
     }
   }
 
@@ -260,13 +261,14 @@ function activateTsv(
     );
 
     if (settingRegistry) {
-      toolbarFactory = createToolbarFactory(
+      toolbarFactory = createToolbarFactory({
         toolbarRegistry,
         settingRegistry,
-        FACTORY_TSV,
-        tsv.id,
-        translator
-      );
+        factoryName: FACTORY_TSV,
+        pluginId: tsv.id,
+        translator,
+        isPluginAvailable: (pluginId: string) => app.hasPlugin(pluginId)
+      });
     }
   }
 
--- a/packages/csvviewer-extension/src/index.ts
+++ b/packages/csvviewer-extension/src/index.ts
@@ -116,14 +116,13 @@ function activateCsv(
     );
 
     if (settingRegistry) {
-      toolbarFactory = createToolbarFactory({
+      toolbarFactory = createToolbarFactory(
         toolbarRegistry,
         settingRegistry,
-        factoryName: FACTORY_CSV,
-        pluginId: csv.id,
-        translator,
-        isPluginAvailable: (pluginId: string) => app.hasPlugin(pluginId)
-      });
+        FACTORY_CSV,
+        csv.id,
+        translator
+      );
     }
   }
 
@@ -261,14 +260,13 @@ function activateTsv(
     );
 
     if (settingRegistry) {
-      toolbarFactory = createToolbarFactory({
+      toolbarFactory = createToolbarFactory(
         toolbarRegistry,
         settingRegistry,
-        factoryName: FACTORY_TSV,
-        pluginId: tsv.id,
-        translator,
-        isPluginAvailable: (pluginId: string) => app.hasPlugin(pluginId)
-      });
+        FACTORY_TSV,
+        tsv.id,
+        translator
+      );
     }
   }
 
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -40,6 +40,7 @@ import { IStatusBar } from '@jupyterlab/statusbar';
 import { ITranslator, TranslationBundle } from '@jupyterlab/translation';
 import {
   buildIcon,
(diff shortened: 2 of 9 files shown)


jtpio:
cc @fcollonval I updated the top comment with an example of how this would work in practice with plugins not loaded on the page.

However it's unclear how to fix this for toolbar items, since they seem to be constructed in a different with the toolbar factory which makes it less convenient to access the underlying `JupyterFrontEnd`.

jtpio:
> However it's unclear how to fix this for toolbar items, since they seem to be constructed in a different way with the toolbar factory which makes it less convenient to access the underlying `JupyterFrontEnd`.

One way could be to pass the `JupyterFrontEnd` (could be made optional) to `createToolbarFactory()` so it can be used at the lower layers here for filtering:

https://github.com/jupyterlab/jupyterlab/blob/8a786a25c113aeeb162f8e523dc36944e1bef8e9/packages/apputils/src/toolbar/factory.ts#L73-L74

packages/csvviewer-extension/src/index.ts
```diff
@@ -116,13 +116,14 @@ function activateCsv(
...
         toolbarRegistry,
         settingRegistry,
-        FACTORY_CSV,
-        csv.id,
-        translator
-      );
+        factoryName: FACTORY_CSV,
+        pluginId: csv.id,
+        translator,
+        isPluginAvailable: (pluginId: string) => app.hasPlugin(pluginId)
```
jtpio, line 125: Maybe Lumino should also expose whether a plugin is *activated*? Not just whether it's in the `_pluginMap`.

In that case that function could be modified to also check for `app.isPluginActivated(pluginId)` (in case such function would exist in Lumino at some point).

jtpio:
Some UI tests are now failing because the sidebar settings are manually retrieved here:

https://github.com/jupyterlab/jupyterlab/blob/64c8e4ef35cd327a5cdd29f70259a5f2f5091ee1/packages/application-extension/src/index.tsx#L1356

And `@jupyterlab/application-extension:sidebar` is not a plugin so its corresponding file with the context menu item gets filtered out:

https://github.com/jupyterlab/jupyterlab/blob/64c8e4ef35cd327a5cdd29f70259a5f2f5091ee1/packages/application-extension/schema/sidebar.json#L5-L13

jtpio:
Updated the test settings in the top comment https://github.com/jupyterlab/jupyterlab/pull/11938#issue-1114899414 so it can also be tested with toolbar items:

**Before**

image


**After**

image

jtpio:
This should be ready for review.

cc @fcollonval @krassowski

fcollonval:
@jtpio to be sure to get it correctly, this PR will result in loading only activated plugin initially. But if an extension manually request a setting from its id, it will get loaded.

So for instance if someone replace a plugin by its own implementation but still load the setting from the original plugin this will work - the original plugin settings will not be loaded initially but when the new extension will explicitly load them.

jtpio:
> to be sure to get it correctly, this PR will result in loading only activated plugins initially. But if an extension manually request a setting from its id, it will get loaded - even if it is part of the not active plugin list.

Yes, and this is for example the case with the `@jupyterlab/application-extension:sidebar` plugin mentioned above (https://github.com/jupyterlab/jupyterlab/pull/11938#issuecomment-1076055419) that adds the following context menu:

image

> So for instance if someone replace a plugin by its own implementation but still load the setting from the original plugin this will work - the original plugin settings will not be loaded initially but when the new extension will explicitly load them.

That's the intended behavior. Testing with the [jlab-enhanced-launcher](https://github.com/jupyterlab-contrib/jlab-enhanced-launcher) (providing the [`disabledExtensions` override](https://github.com/jupyterlab-contrib/jlab-enhanced-launcher/blob/5161b3f2f908f95db936adf6a3f133d88863cc66/package.json#L83-L85)) with this branch seems to behave properly:

image

fcollonval (approved):
Thanks @jtpio

jtpio:
> loading only activated plugins initially

To be more precise, it still loads *registered* plugins even if they fail to *activate*. Since `app.hasPlugin()` only checks if a plugin is in the `_pluginMap`, not its state:

https://github.com/jupyterlab/lumino/blob/d289e27ee3d701c2ca32c04deea56dfc23cea58e/packages/application/src/index.ts#L170-L172

krassowski (approved):
Thanks! I see you already commented on a related topic I was thinking of in https://github.com/jupyterlab/lumino/issues/278#issuecomment-1076346718.

jtpio:
Thanks @fcollonval and @krassowski for the reviews :+1: 

Then I would suggest we merge this and I will report here in case there are any issues after testing this change with Notebook v7.

This does not need to be backported and can target 4.0 only.

jtpio:
> Then I would suggest we merge this and I will report here in case there are any issues after testing this change with Notebook v7.

For reference this seems to be doing the job pretty nicely after testing with Notebook v7 in https://github.com/jupyter/notebook/pull/6336.

PR #11952, opened 2022-01-28
Fix for kernel reconnect

### LINKED ISSUE #11534: Kernel auto re-connect stops after 7 attempts
### Problem
Current kernel connection makes only 7 attempts to re-connect after a network connection fails; this amounts to about 1-2 minutes in actual notebook sessions. Many users leave their machines, close them or engage in other activities after they submit a long running task. If a network connection is severed during this time for any reason, although the long running task continues to execute on the kernel, the UI stops re-connecting after 1-2 minutes and shows a state of "Disconnected". When the users return to the notebook and even if network connection is resumed, there is no attempt to re-connect to the kernel and so the task output is stuck at the initial state. 

Although, manually reconnecting to kernel (Kernel -> Reconnect to Kernel) re-hydrates the task output to correct state, most users end up refreshing the notebook and re-submitting the task. This ends up being an immense waste of resources and a frustrating experience for users especially for long running tasks. 

### Proposed Solution

* After UI has exhausted connecting to the kernel, update the dialog with an action button/link to re-connect to kernel. This should bring the option to re-connect front and center to the users and help alleviate this issue.
* Make reconnect limit configurable, so this can be controlled at runtime or through settings. See [here](https://github.com/jupyterlab/jupyterlab/blob/master/packages/services/src/kernel/default.ts#L1464-L1465)  for more context.
* Auto-close the "Server Connection Error" dialog if the kernel connection is resumed. This has to be manually closed today even after the connection to the kernel is resumed.
* Backport kernel re-connect API to version 1.2.x. Lab users for this version doesn't have an option to re-connect to kernel.

#### Reproducing the issue locally
1. Start Jupyter Lab - `jupyter lab`
2. Setup port forwarding to port 8889 - `ssh -L 8889:localhost:8888 localhost`
3. Open lab on port 8889 - `http://localhost:8889/lab`
4. Open a new ipython notebook, paste this code and execute
```python
import time

step = 1

while step < 10:
    print(f"Step {step}")
    time.sleep(30)
    step = step + 1
    
print('Done')
``` 
5. Stop local port forwarding - `kill $(ps aux | awk '/localhost/ && !/awk/ {print $2}')`
6. Wait for the kernel status to become "Disconnected" in the status bar, this should take 1-2 mins.
7. Re-setup the port forwarding - `ssh -L 8889:localhost:8888 localhost`
8. No more output is seen from the cell, refreshing the notebook will not re-connect to the same kernel session.

3coins:
## References
Fixes #11534. 

## Code changes

Added a reconnect for `getKernelModel`, so for network failures it will be called again until the connection works or the server responds with a "dead" kernel response.

## User-facing changes

For long running kernel tasks, if network connection is dropped and then resumes, JupyterLab will be able to auto-reconnect  with the kernel and resume the output of the task.

None

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -28,7 +28,7 @@ import { Dialog, showDialog } from './dialog';
  * #### Notes
  * The current session connection is `.session`, the current session's kernel
  * connection is `.session.kernel`. For convenience, we proxy several kernel
- * connection and and session connection signals up to the session context so
+ * connection and session connection signals up to the session context so
  * that you do not have to manage slots as sessions and kernels change. For
  * example, to act on whatever the current kernel's iopubMessage signal is
  * producing, connect to the session context `.iopubMessage` signal.
--- a/packages/services/src/kernel/default.ts
+++ b/packages/services/src/kernel/default.ts
@@ -1262,17 +1262,10 @@ export class KernelConnection implements Kernel.IKernelConnection {
 
     let alreadyCalledOnclose = false;
 
-    const earlyClose = async (evt: Event) => {
-      // If the websocket was closed early, that could mean
-      // that the kernel is actually dead. Try getting
-      // information about the kernel from the API call --
-      // if that fails, then assume the kernel is dead,
-      // otherwise just follow the typical websocket closed
-      // protocol.
-      if (alreadyCalledOnclose) {
+    const getKernelModel = async (evt: Event) => {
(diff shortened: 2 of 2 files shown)


blink1073 (approved):
LGTM!

blink1073:
Backporting to 3.3.x since this is a bug fix following #11358

blink1073:

3coins:
@blink1073 
Thanks for reviewing & merging.

PR #11955, opened 2022-01-28
Remove tslint comment and move `Token` to `tokens.ts`

fcollonval:
## Code changes

This removes all _tslint_ comments as we use eslint now.

It moves three `Token` to `tokens.ts` file (as they should be for three shaking reason):
- `ILabStatus` in `@jupyterlab/application`
- `ILauncher` in `@jupyterlab/launcher`
- `ITableOfContentsRegistry` in `@jupyterlab/toc`

This forces the refactor of the launcher and the toc to use only interfaces that are implemented by the corresponding classes.

## User-facing changes

N/A

Developer using the toc will be affected by the rename of `TableOfContentsRegistry` to `ITableOfContentsRegistry`.


jtpio (approved):
Thanks for cleaning this up!

PR #11963, opened 2022-01-31
Support dynamic toolbar definition

fcollonval:
## References

This addresses mainly the error seen in
https://github.com/jupyterlab/retrolab/pull/319#issuecomment-1017371450

it also allows to change the toolbar definitions from the settings without reloading

## Code changes

The toolbar factory for document widget factory can return either a list of toolbar items or an observable list. When an observable list is returned, the toolbar is updated accordingly.

The toolbar definition built using the settings are returning an observable list.

## User-facing changes

Toolbar definitions can be changed without reloading.

> When changing the definition, toolbar items not added through the settings (like the debug button) won't be positioned as expected - a reload will be needed for that.

N/A

### DIFF
--- a/packages/apputils/src/toolbar/factory.ts
+++ b/packages/apputils/src/toolbar/factory.ts
@@ -272,9 +272,7 @@ export function createToolbarFactory(
           );
           break;
         case 'remove':
-          change.oldValues.forEach(() =>
-            toolbar.remove(change.oldIndex)
-          );
+          change.oldValues.forEach(() => toolbar.remove(change.oldIndex));
           break;
         case 'set':
           change.newValues.forEach(item =>
--- a/packages/apputils/src/toolbar/factory.ts
+++ b/packages/apputils/src/toolbar/factory.ts
@@ -1,7 +1,7 @@
 import { IObservableList, ObservableList } from '@jupyterlab/observables';
 import { ISettingRegistry, SettingRegistry } from '@jupyterlab/settingregistry';
 import { ITranslator, TranslationBundle } from '@jupyterlab/translation';
-import { findIndex, toArray } from '@lumino/algorithm';
+import { toArray } from '@lumino/algorithm';
 import { JSONExt, PartialJSONObject } from '@lumino/coreutils';
 import { Widget } from '@lumino/widgets';
 import { Dialog, showDialog } from '../dialog';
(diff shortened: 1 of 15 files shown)


fcollonval:
I'll like to target 3.3.x for this one as the toolbar definition from settings has been backported there.

fcollonval:
cc @jtpio

jtpio:
> I'll like to target 3.3.x for this one as the toolbar definition from settings has been backported there.

Sounds good, thanks!

krassowski (commented):
Changing ranks works beautifully but I could not get the `disabled` option to work - the item did not disappear neither after saving settings nor after reloading. I see that a filter on `item.disabled` was removed - was this intentional, or just an oversight when refactoring?

fcollonval:
This was indeed an unfortunate oversight. I improved the unit tests to cover rank and disabled status update.

fcollonval:
@echarles would you mind reviewing this one? This is the one missing before releasing the 3.3 beta.

echarles:
ranking and disabled work for me. Removing the complete item does not produce the expected result (item being removed), is it expected?

Untitled4

fcollonval:
> ... but I may be misunderstanding the feature. Maybe the settings aims to mutate (rank, disable) the existing toolbar, but is not there to remove any of the existing items?

The settings for the toolbar is taking a similar approach as for the menus and the shortcuts; aka user settings is merged with the default one. Hence the behavior when the user setting is empty.

echarles (approved):

echarles:

echarles:
@fcollonval I have merged this PR in master. Backporting to 3.3.x is giving conflict.

fcollonval:
Thanks I'll do the backport

jtpio:
This still seems to be an issue with `3.3.0rc0`, noticed when updating RetroLab to use the latest `3.3.0-rc.0` packages in https://github.com/jupyterlab/retrolab/pull/351: https://github.com/jupyterlab/retrolab/pull/351/commits/dd750da7814043b24404314016970f0b6e4489ef

jtpio:
We still have the workaround in the test on master too: 

https://github.com/jupyterlab/jupyterlab/blob/c7b14eee863dc2b764378be5e9533ea2e7714d89/packages/apputils/test/toolbar.spec.ts#L310-L314

fcollonval:
I tested locally without the workaround. And it worked. But I presume the trouble is on the CI, isn't it?

jtpio:
The tests seem to be passing on the JupyterLab CI without the workaround: https://github.com/jupyterlab/jupyterlab/pull/12096

jtpio:
But normally this is reproducible locally with retro with https://github.com/jupyterlab/retrolab/pull/351 and without the workaround (https://github.com/jupyterlab/retrolab/pull/351/commits/dd750da7814043b24404314016970f0b6e4489ef)

PR #11969, opened 2022-02-01
Update completer part 2

trungleduc:
## References
Follow-up on  #11795 

## Code changes
- Refactor the connection between the completer manager and the other editor plugins (https://github.com/jupyterlab/jupyterlab/pull/11795#discussion_r795650361)
- The completer can be activated with the `Tab` button or on user typing, 
- New properties in `ICompletionProvider`: 
  - `resolve` allows the completer to resolve the incomplete completion items.
  -  `shouldShowContinuousHint` allows the completer to check if it should be shown in the current context. 
- `ConnectorProxy` now will cancel ongoing `fetch` promises if a new one is called. 
- The default kernel provider `KernelCompleterProvider` will provide docstrings for the completer document panel.

## User-facing changes
- New items in `Code completion` setting : `showDocumentPanel` and `continuousHinting`

https://user-images.githubusercontent.com/4451292/151948266-2cf36f8c-f079-413b-8af9-f84ac671b38b.mp4

- Document content is lazy loaded with loading bar

https://user-images.githubusercontent.com/4451292/151949036-6e57cb5d-f6fa-4445-8db7-d2dba2744731.mp4

N/A


trungleduc:
Hi all, this PR is ready for review

packages/completer/src/handler.ts
```diff
@@ -92,6 +93,10 @@ export class CompletionHandler implements IDisposable {
     return this._isDisposed;
   }
 
+  set continuousHinting(value: boolean) {
```
fcollonval, line 96: I would define a getter too. I found always weird an API that defines a setter but no getter. Alternatively you could convert this to a method `setContinuousHinting`.

packages/completer/src/tokens.ts
```diff
@@ -8,6 +8,28 @@ import { Session } from '@jupyterlab/services';
...
+
+  export const invokeFile = 'completer:invoke-file';
+
+  export const select = 'completer:select';
+
+  export const selectConsole = 'completer:select-console';
+
+  export const selectNotebook = 'completer:select-notebook';
+
+  export const selectFile = 'completer:select-file';
```
fcollonval, line 31: They should be defined in the respective document extensions.

packages/completer/src/widget.ts
```diff
@@ -126,6 +128,10 @@ export class Completer extends Widget {
     }
   }
 
+  set showDocsPanel(showDoc: boolean) {
```
fcollonval, line 131: Same comment about setter without getter.

packages/completer/src/widget.ts
```diff
@@ -126,6 +128,10 @@ export class Completer extends Widget {
...
+    this._showDoc = showDoc ?? true;
```
fcollonval, line 132: Either showDoc could be undefined (the signature needs to be corrected) or the nullish operator is not needed.

```suggestion
    this._showDoc = showDoc ?? true;
```

trungleduc:
Thanks a lot @fcollonval ! I applied suggestions in the last commit

packages/completer/src/tokens.ts
```diff
@@ -65,12 +66,33 @@ export interface ICompletionProvider<
...
 
   /**
-   * Given an incomplete (unresolved) completion item, resolve it by adding all missing details,
-   * such as lazy-fetched documentation.
+   * Given an incomplete (unresolved) completion item, resolve it by adding
+   * all missing details, such as lazy-fetched documentation.
    *
    * @param completion - the completion item to resolve
+   * @param context - The context of the completer
+   * @param path - The text which will be injected if the completion item is
```
fcollonval, line 74:
```suggestion
   * @param patch - The text which will be injected if the completion item is
```

packages/completer/src/model.ts
```diff
@@ -459,6 +458,52 @@ export class CompleterModel implements Completer.IModel {
...
+      if (completionItem.insertText) {
+        patch = this.createPatch(completionItem.insertText);
+      }
+      resolvedItem = completionItem.resolve(patch);
+    } else {
+      resolvedItem = Promise.resolve(completionItem);
+    }
+    return resolvedItem
+      .then(activeItem => {
+        Object.assign(completionItem, activeItem);
```
fcollonval, line 493: This is flagged by the code security scan. Would you mind changing the statement? Probably by something like `Object.keys(activeItem).forEach(key => { completionItem[key] = activeItem[key] });`

packages/notebook-extension/src/index.ts
```diff
@@ -750,6 +751,13 @@ const lineColStatus: JupyterFrontEndPlugin<void> = {
   autoStart: true
 };
 
+const completerPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/notebook-extensions:completer',
```
fcollonval, line 755: You should rename the settings schema `completer.json` (insteand of `notebooks.json`) to be linked with this plugin.

packages/notebook-extension/src/index.ts
```diff
@@ -1572,6 +1581,70 @@ function activateNotebookHandler(
...
+function activateNotebookCompleterService(
+  app: JupyterFrontEnd,
+  notebooks: INotebookTracker,
+  manager?: ICompletionProviderManager
+): void {
+  if (!manager) {
+    return;
+  }
+  const invokeNotebook = 'completer:invoke-notebook';
+  const selectNotebook = 'completer:select-notebook';
```
fcollonval, line 1596: should be defined in `CommandIDs` namespace line 97.

packages/notebook-extension/src/index.ts
```diff
@@ -1572,6 +1581,70 @@ function activateNotebookHandler(
...
+    selector: '.jp-Notebook .jp-mod-completer-active'
+  });
+
+  notebooks.widgetAdded.connect(async (_, notebook) => {
+    const completerContext = {
+      editor: notebook.content.activeCell?.editor ?? null,
+      session: notebook.sessionContext.session,
+      widget: notebook
+    };
+    await manager.attachWidget(notebook, completerContext);
```
fcollonval, line 1628: There is an inconsistency in the API the widget is passed twice: once as first argument and in the context object.

You should pass it only once. And to align with most JLab API I would keep only the object as single argument for `attachWidget`.

> This likely means that `widget` is not optional but mandatory.

packages/notebook-extension/src/index.ts
```diff
@@ -1572,6 +1581,70 @@ function activateNotebookHandler(
...
+    notebook.content.activeCellChanged.connect((_, cell) => {
+      const newCompleterContext = {
+        editor: cell.editor,
+        session: notebook.sessionContext.session,
+        widget: notebook
+      };
+      manager.updateHandler(notebook.id, newCompleterContext);
```
fcollonval, line 1635: Same inconsistency as for `attachWidget`

packages/tooltip-extension/src/index.ts
```diff
@@ -283,6 +283,7 @@ namespace Private {
    */
   export function fetch(options: IFetchOptions): Promise<JSONObject> {
     const { detail, editor, kernel } = options;
+
```
fcollonval, line 286: If this is a manual changes, let's remove this change to avoid confusion as tooltip is not modified by this PR

packages/metapackage/package.json
```diff
@@ -21,7 +21,9 @@
   "files": [
     "lib/*.d.ts",
     "lib/*.js.map",
-    "lib/*.js"
+    "lib/*.js",
+    "style/index.css",
+    "style/index.js"
```
fcollonval, line 26: Those change should not be happening.

```suggestion
    "lib/*.js"
```

packages/fileeditor-extension/src/index.ts
```diff
@@ -373,3 +385,101 @@ function activate(
...
+  app: JupyterFrontEnd,
+  editorTracker: IEditorTracker,
+  manager?: ICompletionProviderManager
+): void {
+  if (!manager) {
+    return;
+  }
+
+  const selectFile = 'completer:select-file';
+  const invokeFile = 'completer:invoke-file';
```
fcollonval, line 402: Should be defined in `CommandIDs` namespace at commands.ts#L47.

packages/console-extension/src/index.ts
```diff
@@ -847,3 +857,68 @@ async function activateConsole(
...
+function activateConsoleCompleterService(
+  app: JupyterFrontEnd,
+  consoles: IConsoleTracker,
+  manager?: ICompletionProviderManager
+): void {
+  if (!manager) {
+    return;
+  }
+  const invokeConsole = 'completer:invoke-console';
+  const selectConsole = 'completer:select-console';
```
fcollonval, line 873: should be defined in `CommandIDs` namespace L52

packages/completer/src/tokens.ts
```diff
@@ -81,12 +103,71 @@ export const ICompletionProviderManager = new Token<ICompletionProviderManager>(
...
+   */
+  select(id: string): void;
+
+  /**
+   * Attach completion service to a widget.
+   *
+   * @param widget - the widget which requests completion service.
+   * @param completerContext - current completion context.
+   */
+  attachWidget(
```
fcollonval, line 133: I would change the wording to

```suggestion
  attachCompleter(
```

packages/completer/src/tokens.ts
```diff
@@ -81,12 +103,71 @@ export const ICompletionProviderManager = new Token<ICompletionProviderManager>(
...
+    widget: Widget,
+    completerContext: ICompletionContext
+  ): Promise<void>;
+
+  /**
+   * Invoke the completer in the widget with provided id.
+   *
+   * @param id - the id of notebook panel, console panel or code editor.
+   */
+  updateHandler(
```
fcollonval, line 143: And reflect the wording here:

```suggestion
  updateCompleter(
```

packages/completer/src/tokens.ts
```diff
@@ -81,12 +103,71 @@ export const ICompletionProviderManager = new Token<ICompletionProviderManager>(
...
+   */
+  attachWidget(
+    widget: Widget,
+    completerContext: ICompletionContext
+  ): Promise<void>;
+
+  /**
+   * Invoke the completer in the widget with provided id.
+   *
+   * @param id - the id of notebook panel, console panel or code editor.
```
fcollonval, line 141: The doc string does not seems correct.

fcollonval (changes requested):
Thanks a lot for the refactoring this is a much cleaner separation @trungleduc 

I still have to test it (binder is failing to spin on this PR...) But here are some additional suggestions.

packages/completer/src/model.ts
trungleduc: I tried to use `Object.key` loop before but it becomes quite ugly to make TS compile 

```typescript
(Object.keys(activeItem) as Array<
  keyof CompletionHandler.ICompletionItem
>).forEach(
  <Key extends keyof CompletionHandler.ICompletionItem>(key: Key) => {
    completionItem[key] = activeItem[key];
  }
);
```

packages/metapackage/package.json
trungleduc: Thanks for catching it twice 😃 I reintroduced it with the rebase.

packages/notebook-extension/src/index.ts
trungleduc: I merged `attachWidget` into  `updateHandler` with simplified signature.

krassowski:
Let's discuss naming settings:
- `continuousHinting` → "continuous hinting" phrase was taken from notebook's hitnerland extension, but it turns out it is not too common elsewhere; I wonder if we should use something else, possibly more standard (if there is anything)?
- `showDocumentPanel` → `showDocumentationPanel`?

We could bring it up tomorrow.

packages/completer-extension/schema/tracker.json
```diff
@@ -20,10 +20,22 @@
...
+      "title": "Default timeout for a provider.",
       "description": "If a provider can not return the response for a completer request before timeout, the result of this provider will be ignored. Value is in millisecond",
       "type": "number",
       "default": 1000
+    },
+    "showDocumentPanel": {
+      "title": "Show the document panel.",
+      "description": "Show or hide the document panel of completer widget, this configuration is applied to all providers. If a provider does not provide the document content, the panel will be hidden for this provider.",
+      "type": "boolean",
+      "default": true
```
krassowski, line 32: Not sure if we want it default to true. Would be good to bring up on dev meeting, but it would probably be fine to merge as default and flip to `false` just before 4.0 release if we find it lacks consensus for enabling it as new default :).

krassowski:
For some reason `onTextChanged` does not get called for the first character of any line (first character in the cell) and continuous hinting does not work for the first letter typed ever. I guess somewhere upstream there is a check which incorrectly bails if ch = 0?

trungleduc:
Thanks @krassowski, I will bring these points up at the dev meeting today.

packages/console-extension/src/index.ts
```diff
@@ -192,6 +197,14 @@ const lineColStatus: JupyterFrontEndPlugin<void> = {
   autoStart: true
 };
 
+const completerPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/console-extensions:completer',
```
fcollonval, line 201: Typo

```suggestion
  id: '@jupyterlab/console-extension:completer',
```

packages/fileeditor-extension/src/index.ts
```diff
@@ -162,13 +165,22 @@ const lineColStatus: JupyterFrontEndPlugin<void> = {
   autoStart: true
 };
 
+const completerPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/fileeditor-extensions:completer',
```
fcollonval, line 169:
```suggestion
  id: '@jupyterlab/fileeditor-extension:completer',
```

packages/notebook-extension/src/index.ts
```diff
@@ -750,6 +755,13 @@ const lineColStatus: JupyterFrontEndPlugin<void> = {
   autoStart: true
 };
 
+const completerPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/notebook-extensions:completer',
```
fcollonval, line 759:
```suggestion
  id: '@jupyterlab/notebook-extension:completer',
```

yarn.lock
```diff
@@ -11676,9 +11676,9 @@ simple-concat@^1.0.0:
   integrity sha512-cSFtAPtRhljv69IK0hTVZQ+OfE9nePi/rtJmw5UjHeVyVroEqJXP1sFztKUy1qU+xvz3u/sfYJLa947b7nAN2Q==
 
 simple-get@^3.0.3:
-  version "3.1.1"
-  resolved "https://registry.yarnpkg.com/simple-get/-/simple-get-3.1.1.tgz#cc7ba77cfbe761036fbfce3d021af25fc5584d55"
-  integrity sha512-CQ5LTKGfCpvE1K0n2us+kuMPbk/q0EKl82s4aheV9oXjFEz6W/Y7oQFVJuU6QG77hRT4Ghb5RURteF5vnWjupA==
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/simple-get/-/simple-get-3.1.0.tgz#b45be062435e50d159540b576202ceec40b9c6b3"
+  integrity sha512-bCR6cP+aTdScaQCnQKbPKtJOKDp/hj9EDLJo3Nw4y1QksqaovlW/bnptB6/c1e+qmNIDHRK+oXFDdEqBT8WzUA==
```
fcollonval, line 11681: This probably needs to be reverted

fcollonval (changes requested):
@trungleduc There is a typo in the plugin names (it is `-extension` in singular not in plural) and the change in the yarn.lock should likely be reverted as unneeded. Then this is good to go

PR #11970, opened 2022-02-01
Split caption for pausing on exceptions

fcollonval:
## References

https://github.com/jupyterlab/jupyterlab/pull/11752#issuecomment-1026010018

## Code changes

Split the caption to highlight the status of the toggleable command
use function for label and caption so it gets properly updated when the command state changes

## User-facing changes

Caption is clarified on the command status

N/A

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -721,12 +721,14 @@ const main: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.pause, {
-      label: () => service.isPausingOnExceptions
-        ? trans.__('Disable pausing on exceptions')
-        : trans.__('Enable pausing on exceptions'),
-      caption: () => service.isPausingOnExceptions
-        ? trans.__('Disable pausing on exceptions')
-        : trans.__('Enable pausing on exceptions'),
+      label: () =>
+        service.isPausingOnExceptions
+          ? trans.__('Disable pausing on exceptions')
+          : trans.__('Enable pausing on exceptions'),
+      caption: () =>
+        service.isPausingOnExceptions
+          ? trans.__('Disable pausing on exceptions')
+          : trans.__('Enable pausing on exceptions'),
       className: 'jp-PauseOnExceptions',
       icon: Debugger.Icons.pauseOnExceptionsIcon,
       isToggled: () => {
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -721,10 +721,12 @@ const main: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.pause, {
-      label: service.isPausingOnExceptions
+      label: () => service.isPausingOnExceptions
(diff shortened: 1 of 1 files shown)


jasongrout:
I thought our convention was that the toggle status indicated state, and the language didn't change. It's confusing for me to see a command checked and the language indicating the negative of the status.

krassowski:
I think that we are looking at a different part of the UI @jasongrout. My comment was about the tooltip which currently shows `Enable / Disable pausing on exceptions`:

enable-disable-button

But it seems you were thinking about title in the command palette, which already uses the format which you say is undesirable:

Screenshot from 2022-02-01 21-36-33
Screenshot from 2022-02-01 21-36-47

I agree that for command palette the current state (i.e prior to this PR) is suboptimal. I think it should be showing "Pause on exceptions" with a tick or without a tick. It would be also useful to use the consistent capitalisation here (also the pause icon needs the css highlight class for contrast or to be removed).

As for the button we should re-consider it's style also because while in 3.3 the background is grey and active button is white, in 4.0 it is reversed (the background is white and active button is grey) which will lead to confusion on future migration.

3.3 vs 4.0 changes apart, it is hard to know if the button is enabled or waiting to be pressed. Changing tooltip as in this PR will help a little (and possibly improve accessibility), but we can improve it a bit more I believe:

> Side note: regarding the enable/disable confusion, there is some more ideas on UX stackexchange [here](https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan) and [here](https://ux.stackexchange.com/questions/5069/should-toggle-buttons-show-what-they-do-or-the-system-status). One suggestion is to avoid buttons for what could be checkboxes.

Let's see what others do:
- Chrome/Edge has tooltips for two states with tooltip "Pause on exceptions" when disabled (indicates action) and "Don't pause on exceptions" when enabled; when action is enabled, an extra checkbox "Pause on caught exceptions" shows up and the button is blue
  Screenshot from 2022-02-01 21-31-53
  Screenshot from 2022-02-01 21-45-28
- Firefox has a checkbox:
   Screenshot from 2022-02-01 21-30-44
- VSCode has a checkbox too:
   Screenshot from 2022-02-01 21-48-33

fcollonval:
Thanks for the research @krassowski 

d4860d1I updated the wording to `Pause on exceptions` for _label_ and to a switching caption (so the button title will be switched).

krassowski (approved):
Thank you!

packages/debugger-extension/src/index.ts
```diff
@@ -721,10 +721,11 @@ const main: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.pause, {
-      label: service.isPausingOnExceptions
-        ? trans.__('Disable pausing on exceptions')
-        : trans.__('Enable pausing on exceptions'),
-      caption: trans.__('Enable / Disable pausing on exceptions'),
+      label: trans.__('Pause on exceptions'),
```
krassowski, line 724: For consistency with other items in palette we may prefer:

```suggestion
      label: trans.__('Pause on Exceptions'),
```

but this is not a strong opinion.

PR #11973, opened 2022-02-01
Adjust z-index of execution progress tooltip

### LINKED ISSUE #11972: The execution progress tooltip shown on hover is behind search bar
## Description

This is a minor usability issue: the execution progress tooltip shown on hover (to be introduced in 3.3) lands behind search bar:

overlap2

## Reproduce

1. Install from 3.3 branch
2. Create several cells with `from time import sleep; sleep(2)`
3. Open search bar
4. Select the added cells and run them all
5. Hover over the kernel status indicator

## Expected behavior

The tooltip shown on hover should be not be obscured by the searchbar.

## Context

- Browser and version: Chrome 97
- JupyterLab version: 3.3 dev

Sync271:
## References

Fixes #11972

## Code changes

z-index of the search bar is 7
https://github.com/jupyterlab/jupyterlab/blob/f995e5d8be9a31b40fc08d8a210ec32bc849b0bf/packages/documentsearch/style/base.css#L19
and tooltip was 1
https://github.com/jupyterlab/jupyterlab/blob/f995e5d8be9a31b40fc08d8a210ec32bc849b0bf/packages/apputils/style/toolbar.css#L23

This was causing the search bar to overshadow the tooltip so changed the z-index of tooltip to 8 so that it stays on top
```diff
-  z-index: 1;
+  z-index: 8;
```
### Before
image
### After
image

### DIFF
--- a/packages/apputils/style/toolbar.css
+++ b/packages/apputils/style/toolbar.css
@@ -20,7 +20,7 @@
   background: var(--jp-toolbar-background);
   min-height: var(--jp-toolbar-micro-height);
   padding: 2px;
-  z-index: 1;
+  z-index: 8;
   overflow-x: hidden;
 }
 


fcollonval (approved):
Thanks a lot @Sync271

jtpio:
Is it already fixed on master or should it then be backported?

trungleduc:
> Is it already fixed on master or should it then be backported?

No, the fix is not on master yet.

jtpio:
Alright thanks!

jtpio:

krassowski:
Late to this party, but what do you all think about the way the search bar now appears (i.e. shadowed over by the toolbar)? Better? Worse? More distracting? Nicer? Just noting that there was this extra change and I don't know how I feel about it.

fcollonval:
Good catch @krassowski I would say I like the border separation. But the shadow makes it awkward.

In VSCode in default light theme, they use a different background color (v1.64):

image

A possibility could be to move it slightly below (like in the JSON viewer but maybe not so much).

PR #11949, opened 2022-01-27
Fix autocomplete in console

fcollonval:
## References

Fix #11710

## Code changes

Add awareness to standalone cell to restore cursor activity event

## User-facing changes

Completer and cursor position are back in the console

## Backwards-incompatible changes

N/A

PR #11975, opened 2022-02-02
Backport PR #11949 on branch 3.2.x (Fix autocomplete in console)

### LINKED ISSUE #11466: Tab completion not working in console
## Description

The tab completion works as expected in the notebook, suggesting all possibilities when pressing the tab key after the `.`
Example: 
`object.<TAB KEY>` works as expected giving all possible attributes and methods of `object`.

However **in the console** this does not work and the tab key simply works as a tab spacer.

I noticed that instead tab completion works fine in the console when pressing TAB while typing a file path inside a string. In this case it correctly suggests all file contained in the folder. So it seems a problem with tab completion after the `.`

I attach a video to clear this out

- Operating System and version: macOS 10.15.7
- Browser and version: Chrome 95.0.4638.69
- JupyterLab version:  3.2.3  

https://user-images.githubusercontent.com/59478987/141798550-b608f42b-fbd7-42de-ad6f-39c04a26dcc2.mov

meeseeksmachine:
Backport PR #11949: Fix autocomplete in console

Fixes #11466


jtpio:
CI is now green.

PR #11998, opened 2022-02-03
Fix debugger UI tests

fcollonval:
## References

UI tests on _master_ are failing due to the debugger panel not opening when the notebook debugger is opened.

The bug is related to ipykernel 6.8.0 (I succeed to reproduce it after upgrading from ipykernel 6.7.0 to 6.8.0). Still looking for the root cause.

The [initialized](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Initialized) event does not appear on the kernel comm.

### DIFF
--- a/scripts/ci_install.sh
+++ b/scripts/ci_install.sh
@@ -24,7 +24,7 @@ pip install -q --upgrade pip --user
 pip --version
 # Show a verbose install if the install fails, for debugging
 pip install -e ".[test]" || pip install -v -e ".[test]"
-pip install "git+https://github.com/ipython/ipykernel.git@97cf7bdac07e84d7ff796539d8b7c3117f567555#egg=ipykernel"
+pip install "git+https://github.com/ipython/ipykernel.git@main#egg=ipykernel"
 jlpm versions
 jlpm config current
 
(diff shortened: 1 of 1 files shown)


fcollonval:
@JohanMabille do you have any idea why the event _initialized_ would not be emitted in ipykernel?

fcollonval:
Using `ipykernel < 6.8.0` solves the issue. So definitely something to do with the latest ipykernel release.

echarles:
These are the ipykernel PR merged between 6.7.0 and 6.8.0

https://github.com/ipython/ipykernel/pull/816 and https://github.com/ipython/ipykernel/pull/844 related to modules request (I have done them but don't see how it could impact)

https://github.com/ipython/ipykernel/pull/849

https://github.com/ipython/ipykernel/pull/850

fcollonval:
Commit b99e162 is testing the fix https://github.com/ipython/ipykernel/pull/855

fcollonval:
No idea why b99e162 does not work - it works locally. So waiting for a patch release on ipykernel

echarles:
It should work but I remember having weird behavior with explictit git revision number. You could try "git+https://github.com/ipython/ipykernel.git@master#egg=ipykernel"

echarles:
main, not master: git+https://github.com/ipython/ipykernel.git@main#egg=ipykernel

fcollonval:
Fixed by upstream ipykernel [v6.9.0 release](https://github.com/ipython/ipykernel/releases/tag/v6.9.0)

PR #12002, opened 2022-02-03
docs: fix shell command with unquoted '>'

ErikBjare:
## Code changes

Without this fix, some (most?) shells may interpret it as: redirect stdout into the file '=6'

### DIFF
--- a/docs/source/user/debugger.rst
+++ b/docs/source/user/debugger.rst
@@ -28,7 +28,7 @@ Here is an example of how to install ``ipykernel`` and ``xeus-python`` in a new 
 
 .. code:: bash
 
-   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 'ipykernel>=6' xeus-python
+   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 "ipykernel>=6" xeus-python
    conda activate jupyterlab-debugger
 
 Usage
--- a/docs/source/user/debugger.rst
+++ b/docs/source/user/debugger.rst
@@ -28,7 +28,7 @@ Here is an example of how to install ``ipykernel`` and ``xeus-python`` in a new 
 
 .. code:: bash
 
-   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 ipykernel>=6 xeus-python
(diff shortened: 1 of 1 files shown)


ErikBjare:
Looks like the "Enforce PR label" CI frontran the tag bot :shrug: 

```
Error: Select at least one triage label ['bug','enhancement','feature','maintenance','documentation']
```

bollwyvl:
It's a good fix! 

Frankly, though, when we document these things, we should probably be encouraging folks to build up `environment.yml`/`requirements.txt` and then `conda env update --file`/`pip install -r` them to help get atomic success... or failure and better log messages. 

I can't count the amount of hours our users (and maintainers) have spent with peoples' frankenvironments that come out of mixed interactive package management, and the (lack of) pin metadata that come from e.g. `pip[extras]`, for example, are basically ignored on subsequent solves.

docs/source/user/debugger.rst
```diff
@@ -28,7 +28,7 @@ Here is an example of how to install ``ipykernel`` and ``xeus-python`` in a new
 
 .. code:: bash
 
-   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 ipykernel>=6 xeus-python
+   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 'ipykernel>=6' xeus-python
```
fcollonval, line 31: Thanks for catching this one

let's use double quote (for those using Windows CMD shell):

```suggestion
   conda create -n jupyterlab-debugger -c conda-forge jupyterlab=3 "ipykernel>=6" xeus-python
```

fcollonval (changes requested):

docs/source/user/debugger.rst
ErikBjare: Ah right, forgot about that difference on Windows! Committed.

fcollonval (approved):
Thanks @ErikBjare

fcollonval:
Small note for future contribution: Prefer targeting the `master` branch rather than version branches. Our process is to push on master (dev branch) and backport when needed to production branch

fcollonval:

bollwyvl:
Another future angle: we could consider adding [sphinxcontrib-shellcheck](https://github.com/pmacosta/sphinxcontrib-shellcheck)... though this would not have caught the quote issue.

PR #12005, opened 2022-02-04
show pause on exception button when not available and change caption …

### LINKED ISSUE #11965: Pause on exception button does not show up
Giving a spin to 3.3 alpha because I very much need the "pause on exception" debugger feature. I noticed that the button is missing. I guess a specific kernel version is required for it to show up based on:

https://github.com/jupyterlab/jupyterlab/blob/f7aa928c48fb6cef71d21e661cb8832185aaaac1/packages/debugger-extension/src/index.ts#L733-L735

### Problem

Hiding buttons when features are not available makes the features less discoverable. When we added debugger many users did not realise that they just needed to update ipykernel/xeus kernel to get it working and we got a lot of questions from confused users, exacerbated by the fact that some of them were genuine issues with how debugger capability was detected.
 
### Proposed Solution

- [ ] Always show the pause button; when it is not available make it disabled add a tooltip "Your kernel version does not support pausing on exception yet"
- [ ] Document what kernels are know to work with this feature in user-facing documentation

### Context

We discussed this problem for the debugger bug icon in: https://github.com/jupyterlab/jupyterlab/pull/10727#issuecomment-892534609 and https://github.com/jupyterlab/jupyterlab/pull/10727#issuecomment-892819037

We also use a condition like this for rich variable rendering:

https://github.com/jupyterlab/jupyterlab/blob/f7aa928c48fb6cef71d21e661cb8832185aaaac1/packages/debugger-extension/src/index.ts#L417-L419

andrewfulton9:
…to reflect status

## References

Fixes #11965
Closes #11970

I used PR #11970 as my basis for this.
## Code changes

changes the caption for the pause_on_exception to reflect if a kernel does not support pausing.
## User-facing changes

the caption for pause is changed depending on the kernel and debugger extension state.

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -721,19 +721,21 @@ const main: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.pause, {
-      label: service.isPausingOnExceptions
-        ? trans.__('Disable pausing on exceptions')
-        : trans.__('Enable pausing on exceptions'),
-      caption: trans.__('Enable / Disable pausing on exceptions'),
+      label: trans.__('Enable / Disable pausing on exceptions'),
+      caption: () =>
+        service.isStarted
+          ? service.pauseOnExceptionsIsValid()
+            ? service.isPausingOnExceptions
+              ? trans.__('Disable pausing on exceptions')
+              : trans.__('Enable pausing on exceptions')
+            : trans.__('Kernel does not support pausing on exceptions.')
+          : trans.__('Enable / Disable pausing on exceptions'),
       className: 'jp-PauseOnExceptions',
       icon: Debugger.Icons.pauseOnExceptionsIcon,
       isToggled: () => {
         return service.isPausingOnExceptions;
       },
       isEnabled: () => {
-        return !!service.isStarted;
-      },
-      isVisible: () => {
         return service.pauseOnExceptionsIsValid();
       },
       execute: async () => {


packages/debugger-extension/src/index.ts
```diff
@@ -721,19 +721,21 @@ const main: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.pause, {
-      label: service.isPausingOnExceptions
-        ? trans.__('Disable pausing on exceptions')
-        : trans.__('Enable pausing on exceptions'),
-      caption: trans.__('Enable / Disable pausing on exceptions'),
+      label: trans.__('Enable / Disable pausing on exceptions'),
```
fcollonval, line 724: From #11970 review

```suggestion
      label: trans.__('Pausing on Exceptions'),
```

packages/debugger-extension/src/index.ts
```diff
@@ -721,19 +721,21 @@ const main: JupyterFrontEndPlugin<void> = {
...
+      caption: () =>
+        service.isStarted
+          ? service.pauseOnExceptionsIsValid()
+            ? service.isPausingOnExceptions
+              ? trans.__('Disable pausing on exceptions')
+              : trans.__('Enable pausing on exceptions')
+            : trans.__('Kernel does not support pausing on exceptions.')
+          : trans.__('Enable / Disable pausing on exceptions'),
```
fcollonval, line 732: I would drop the `service.isStarted` to focus on providing the info if the kernel does not support the feature or the button state.

```suggestion
        service.pauseOnExceptionsIsValid()
          ? service.isPausingOnExceptions
            ? trans.__('Disable pausing on exceptions')
            : trans.__('Enable pausing on exceptions')
          : trans.__('Kernel does not support pausing on exceptions.'),
```

fcollonval (changes requested):
Thanks @andrewfulton9

I have some suggestions to improve the logic and some wording.

packages/debugger-extension/src/index.ts
andrewfulton9: I added in the `service.isStarted` logic because otherwise, if someone hovers over the button when the debugger is turned off, it will show the the kernel does not support pausing on exceptions which may not be true and could be a misdirection in case they turn on the debugger and don't hover again to see that the kernel actually does support it. I'm happy to change this, but thought it was worth mentioning. I went back and forth on it a few times when I wrote this part.

fcollonval:

Issue #11850, opened 2022-01-12
Use path comparisons instead of string comparisons in initialize_templates()


kellyyke:
### Problem
We have a rather special setup where we're using `Bazel` and `rules_python`, resulting in a folder structure that looks like this:
```
pypi__jupyterlab
├── jupyterlab
├── jupyterlab-3.2.5.data
│   └── data
│       ├── share
│       │   ├── jupyter
│       │   │   ├── lab
│       │   │   │   ├── schemas
│       │   │   │   ├── static
│       │   │   │   ├── themes
```
We are able to launch jupyterlab fine, but it  keeps launching in `core_mode`, meaning extensions are disabled.
```
[I 2021-11-30 09:15:04.680 LabApp] Running JupyterLab in core mode
[I 2021-11-30 09:15:04.681 LabApp] JupyterLab extension loaded from /app/tools/notebooks_lab/notebooks_lab.runfiles/canva_pip_ml/pypi__jupyterlab/jupyterlab
[I 2021-11-30 09:15:04.681 LabApp] JupyterLab application directory is /app/tools/notebooks_lab/notebooks_lab.runfiles/canva_pip_ml/pypi__jupyterlab/jupyterlab-3.2.4.data/data/share/jupyter/lab
```

After some debugging, it seems that [this line](https://github.com/jupyterlab/jupyterlab/blob/e3edc9daacdd474bc58a765742dc3af88af700bf/jupyterlab/labapp.py#L553), combined with our folder structure, will always evaluate to `True`, resulting in `core_mode` being set automatically.

```
# HERE='/app/tools/notebooks_lab/notebooks_lab.runfiles/canva_pip_ml/pypi__jupyterlab/jupyterlab'
# self.app_dir='/app/tools/notebooks_lab/notebooks_lab.runfiles/canva_pip_ml/pypi__jupyterlab/jupyterlab-3.2.4.data/data/share/jupyter/lab'

def initialize_templates(self):
        # Determine which model to run JupyterLab
        if self.core_mode or self.app_dir.startswith(HERE):
            self.core_mode = True
            self.log.info('Running JupyterLab in core mode')
        ...

```

### Proposed Solution
[...]

PR #12024, opened 2022-02-08
use path comparison instead of string comparisons

kellyyke:
## References
https://github.com/jupyterlab/jupyterlab/issues/11850

## Code changes
Use path comparisons between `HERE` and `app.self_dir` in `initialize_templates()`

## User-facing changes
None

None


fcollonval:

PR #12027, opened 2022-02-08
fix run cells breaking on non-header markdown cells

### LINKED ISSUE #12026: ToC Run Cells doesn't run nested cells after non-header markdown cells
## Description
If I try to run code cells under a ToC header with the run cells command, the cells below a non-header markdown cell will not be run.

chrome-capture (4)
Notice in the above screen shot that when I use run cells on the "Text" header, the code cell below the header, defining and printing text is not run. This is because of the markdown cell immediately before it. However when I use run cells on the "HTML" header, the cells below do run because there are no non-header markdown cells between them and their section header.

## Reproduce
create a notebook with three cells. The first cell make a markdown header, the second cell make a markdown cell and fill with text without any markdown headers. the 3rd cell, fill with python code. 

1) From the ToC, right click on header
2) click "Run Cell(s)"
3) notice the third code cell is not run

## Expected behavior

I would expect any code cells nested under a header to be run when I use the Run Cells command

## Context

- Operating System and version: Linux Ubuntu 18.04
- Browser and version: Chrome 97.0.4692.99
- JupyterLab version: master

andrewfulton9:
## References
Fixes #12026
If this could be backported to 3.3.x, that would be great too.

## Code changes
in the toc-extension `runCells` command, the logic of when to break out of a loop running each code cell is amended so that the loop is not broken if `MarkdownCell.headingInfo.level` is equal to -1 (markdown cell is not a header)

## User-facing changes
Now code cells under a ToC section are run after runCells is called even if they are under a non-header markdown cell

chrome-capture (5)

None that I am aware

### DIFF
--- a/packages/toc-extension/src/index.ts
+++ b/packages/toc-extension/src/index.ts
@@ -111,7 +111,11 @@ async function activateTOC(
         let level = activeCell.headingInfo.level;
         for (let i = cells.indexOf(activeCell) + 1; i < cells.length; i++) {
           const cell = cells[i];
-          if (cell instanceof MarkdownCell && cell.headingInfo.level <= level) {
+          if (
+            cell instanceof MarkdownCell &&
+            cell.headingInfo.level <= level &&
+            cell.headingInfo.level > -1
+          ) {
             break;
           }
 


krassowski:
Worth adding a test case?

fcollonval:
@andrewfulton9 you can add it in `galata/test/jupyterlab/toc-running.test.ts` as it test that feature. You should be able to test it by changing the generated notebook (see `beforeEach`). And to test cell output content, you could look at the test _Check cell output #_ in `galata/test/jupyterlab/notebook-run.test.ts`

andrewfulton9:
@fcollonval, thanks for the extra context for updating the test. I'll try to get that added today.

fcollonval:

PR #11762, opened 2021-12-30
Restore compact notebook layout on mobile

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11516

## Code changes

- [x] Format input and output prompts based on the `only screen and (max-width: 760px)` CSS media query that was removed in https://github.com/jupyterlab/jupyterlab/pull/9831.
- [x] Add UI Test

## User-facing changes

Restore back the more compact notebook layout on mobile:

https://user-images.githubusercontent.com/591645/147733322-9914d32c-6690-4cfe-96c2-f8d68944f258.mp4

## Backwards-incompatible changes

None

PR #12028, opened 2022-02-08
Creates cell-toolbar, cell-toolbar-extension packages and populates toolbar

jweill-aws:
Created blank `cell-toolbar` and `cell-toolbar-extension` packages using `jlpm run create:package`. Also added `jupyterlab: { coreDependency: true }` (with strings quoted) in each package's `package.json` file. Split code from [jlab-enhanced-cell-toolbar](https://github.com/jupyterlab-contrib/jlab-enhanced-cell-toolbar) into two packages, `cell-toolbar` and `cell-toolbar-extension`, for use with JupyterLab.

## References

#9733

## Code changes

Incorporates code from jlab-enhanced-cell-toolbar into two new packages, `cell-toolbar` and `cell-toolbar-extension`. The CSS classes are modified to be distinct from jlab-enhanced-cell-toolbar so that both could be used.

## User-facing changes

Cell toolbar appears on every cell in a notebook unless it would interfere with the input (code/Markdown) or output (Markdown only) of the cell.

Cell toolbars do not appear if they would occupy more than 50% of a cell's input area width.

Minor style changes, such as adding a 1-pixel border to markdown cells (transparent if unselected, gray if selected), which will affect automated UI tests. 

Users should be able to disable `@jupyterlab/cell-toolbar-extension` if they would prefer to use `jlab-enhanced-cell-toolbar`, which has more features and more customization options.


packages/cell-toolbar-extension/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "jest": "^26.4.2",
+    "rimraf": "~3.0.0",
+    "ts-jest": "^26.3.0",
+    "typescript": "~4.5.2"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "jupyterlab": {
+    "coreDependency": true
```
3coins, line 51: Instead of core dependency, this should be an extension.
```
"extension": true
```

packages/cell-toolbar/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "jest": "^26.4.2",
+    "rimraf": "~3.0.0",
+    "ts-jest": "^26.3.0",
+    "typescript": "~4.5.2"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "jupyterlab": {
+    "coreDependency": true
```
3coins, line 51: Don't think this is required. You will be able to pull this in from `cell-toolbar-extension` package.

3coins (commented):
You might need to add the core-toolbar-extension package as a dependencies in application-extension, so webpack can pull it in.

packages/cell-toolbar-extension/src/index.ts
```diff
@@ -0,0 +1,8 @@
+/* -----------------------------------------------------------------------------
+| Copyright (c) Jupyter Development Team.
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+/**
+ * @packageDocumentation
+ * @module cell-toolbar-extension
+ */
```
3coins, line 8: You might need to export a simple plugin here. 

```typescript
const cellToolbar: JupyterFrontEndPlugin<void> = {
  id: "@jupyterlab/cell-toolbar-extension:plugin",
  requires: [ITranslator],
  optional: [ILabShell, ISettingRegistry, ICommandPalette],
  autoStart: true,
  activate: (
    app: JupyterFrontEnd,
    translator: ITranslator,
    labShell: ILabShell | null,
    settingRegistry: ISettingRegistry | null,
    palette: ICommandPalette | null
  ) => {
    console.log('cell toolbar extension works...');
  };
};

export default cellToolbar;
```

jweill-aws:
Changes are failing to push (unless forced) due to the following error locally. Running `jlpm run integrity` does not fix the problem.

```
{
  "@jupyterlab/metapackage": [
    "Style files not included in sideEffects in /Users/jweill/git/jupyterlab/packages/metapackage"
  ]
}


Please run `jlpm run integrity` locally and commit the changes
```

fcollonval:
> Changes are failing to push (unless forced) due to the following error locally. Running `jlpm run integrity` does not fix the problem.
> 
> ```
> {
>   "@jupyterlab/metapackage": [
>     "Style files not included in sideEffects in /Users/jweill/git/jupyterlab/packages/metapackage"
>   ]
> }
> 
> 
> Please run `jlpm run integrity` locally and commit the changes
> ```

@jweill-aws to fix that one you need to update the ensure repo script (see https://github.com/jupyterlab/jupyterlab/pull/11079#discussion_r750195045)

packages/metapackage/package.json
```diff
@@ -21,7 +21,9 @@
   "files": [
     "lib/*.d.ts",
     "lib/*.js.map",
-    "lib/*.js"
+    "lib/*.js",
+    "style/index.css",
+    "style/index.js"
```
fcollonval, line 26:
```suggestion
    "lib/*.js"
```

Left over that can be cleaned.

jweill-aws:
Updates for today:
* Adds and removes the toolbar from cells individually (should avoid cases where newly-loaded cells don't have toolbars)
* Suppresses toolbar from code and unrendered Markdown cells where the top line is so long that it runs into the toolbar.

Not done yet:
* Showing/hiding the toolbar as the edit box gets edited (logic only applies when focusing a cell)
* Showing/hiding the toolbar based on the size of rendered markdown content

meeseeksmachine:

https://discourse.jupyter.org/t/event-for-listening-to-changes-to-a-code-markdown-cells-input-area/13132/1

aiqc:
- A button to change cell type (markdown vs code) would be hugely helpful. It is such a pain to go up to the top toolbar for that so frequently.
- I prefer the colab overlapping top right approach significantly more

image

jweill-aws:
Updated the PR to auto-detect overlap on every change, using a `MutationObserver` on the active cell.

In the video below, I have a few words' worth of sample text on the clipboard and I paste it in to fill the first line. I then change the text area so that the first line does or does not overlap with the toolbar. When any overlap occurs, the toolbar disappears entirely.
https://user-images.githubusercontent.com/93281816/155432812-558a884f-c841-42f2-82d6-6bfff3df65af.mov

jweill-aws:
Today's changes add a `MutationObserver` for code cells and unrendered Markdown cells and a `ResizeObserver` for rendered Markdown cells. Each observer triggers a check for toolbar overlap. If there is overlap, the toolbar will disappear.

aiqc:
Fired up the binder. I like that this will help new users feel more comfortable. However, in a real use case, not just adding cells to a blank page, due to the overlap of content, the feature is just not there half the time:

image

Regarding the buttons, you can already move cells with a mouse using the left side of the cell. On hover, a 4-way arrow drag-n-drop icon on the left side already indicates to the user that the ui allows for moving cells. Perhaps that icon could appear there without a hover.

image

If we run w the left side of the cell as a place to manage cell position, then similarly you could put + icons at the top and bottom of the left side of the cell

It's weird to have a copy button without a paste button, which is why I would suggest making it duplicate, not copy.

Again, I'd advocate for a button that changes cell type as the primary value this toolbar could add.

jweill-aws:
Changes in today's updates:

* Replaces "copy" with "duplicate" in the cell toolbar (thanks @aiqc for the feedback and thanks @WangWanyue for the icon)
* Fixes a bug where in a multi-pane view, on initial load, the toolbar overlap detection code didn't run on all visible notebooks' initially active cells.

fcollonval:
Thanks @jweill-aws for working on this one.

I concur with @aiqc that having no indication in case of text overlap brings a strange feeling for the user. This should be the primary concern. And I would reiterate my suggestion of making the position configurable - I don't agree that taking more vertical space is a bad idea (vscode and colab are now both placing a cell toolbar on the cell border); it is a question of taste and user confidence (junior users will likely prefer it to be visible as it will help using the application - senior users will certainly be annoyed by any additional vertical space taken by it).

Regarding accessibility, when a cell is focused in command mode, pressing tab switches the focus on the first cell toolbar button (this is good). However pressing tab a second time is not entering the cell editor but focusing the second button in the toolbar. This is a bug in the toolbar container. Navigation within a toolbar must be with arrow keys not tab (see [aria practices](https://www.w3.org/TR/wai-aria-practices-1.2/#toolbar)).

Regarding performance, this feature must be easy to turn off - it worsens the performance as adding even more nodes to the DOM: see https://github.com/jupyterlab-contrib/jlab-enhanced-cell-toolbar/issues/45

It may require some testing with the virtual rendering as seen in https://github.com/jupyterlab-contrib/jlab-enhanced-cell-toolbar/issues/33

jweill-aws:
@fcollonval I confirmed that `jupyter labextension disable @jupyterlab/cell-toolbar-extension` disables the cell toolbar with no side effects. This is now in the documentation in this branch. I will look at the keyboard shortcuts to let users navigate through the toolbar with arrow keys instead of just `TAB`.

jweill-aws:
The `ArrowLeft` and `ArrowRight` keys aren't getting picked up by a `keydown` event handler on toolbar button components. After taking a look through the source, it looks like these are already handled on selected cells in command mode — see excerpt below.

If keyboard navigation on the toolbar is necessary, I'll need to use some other keybinding besides :arrow_left: and :arrow_right:.

https://github.com/jupyterlab/jupyterlab/blob/9688c8ef5e5b945518a842e871dfca71dde566e5/packages/notebook-extension/schema/tracker.json#L519-L528

jweill-aws:
@aiqc Here's a screen shot of what it looks like to pull the "move up" and "move down" buttons out of the cell toolbar and move them to the far left. What do you think?
image (2)

aiqc:
@jweill-aws So there is already functionality in the UI to move cells around via the mouse, but it just isn't apparent. When you hover over the little landing zone to the left side of the cell, a tiny 4-way arrow appears. It lets you click and drag the cell up/ down to reposition it. the nice thing about this is that to move a cell up 4 cells you don't have a click a button 4 times and the window scrolls with the drag.
image

You could use a single (permanent, not hover) icon to indicate that the UI allows for this. It would only appear when a cell is selected. Here are a few icons. I like the one of the left. 
image

jweill-aws:
Filed new issues for left/right arrow keyboard navigation within toolbars (not just the cell toolbar) and for adding icons for the cell move handle. These are #12187 and #12188 respectively.

After discussing with @ellisonbg earlier, I'm going to keep the up and down arrows in the cell toolbar for now. Future work will iterate on this design.

packages/cell-toolbar-extension/schema/plugin.json
```diff
@@ -0,0 +1,177 @@
...
+        },
+        "top": {
+          "title": "Top distance in pixels",
+          "type": "number",
+          "minimum": 0,
+          "default": 0
+        }
+      },
+      "additionalProperties": false,
+      "type": "object"
```
krassowski, line 174:
```suggestion
      "additionalProperties": false,
      "title": "Above cell (floating)",
      "type": "object"
```

packages/cell-toolbar-extension/schema/plugin.json
```diff
@@ -0,0 +1,177 @@
...
+      "type": "number",
+      "minimum": 0,
+      "default": 0
+    },
+    "rightMenu": {
+      "title": "List of left menu items",
+      "description": "An item is defined by a 'command' name and an 'icon' name + optionally a 'tooltip' and the 'cellType' on which it applies",
+      "oneOf": [
+        {
+          "type": "null"
```
krassowski, line 112: Do we really need the `null` type here? Would not check for an empty list suffice?

```suggestion
          "type": "null",
          "title": "Right menu disabled"
```

krassowski (commented):
Thank you, UX-wise this PR is in a very nice state already. Two small improvements would be welcome:

1. settings schema requires polishing, it is difficult to understand from the GUI what specific options are (see comments)
2. the positioning requires improvement for "mobile" (=very narrow, cf [#11762](https://github.com/jupyterlab/jupyterlab/pull/11762)) layout
  Screenshot from 2022-03-10 16-31-05
  I would propose to move the toolbar to the same position as the execution counter (`[ ]:`) and to make it always visible for selected cell when in the mobile/compact mode.
3. when multiple cells are selected, the toolbar acts on the entire selection rather than on a single cell (which is good!), however the duplicate action produces different results depending on which way the selection was created:
   - top-bottom (correct - this is what I expected):
      top-bottom
   - bottom-top (IMO incorrect, selected block of cells gets mixed up):
      bottom-top

packages/cell-toolbar-extension/schema/plugin.json
```diff
@@ -0,0 +1,177 @@
...
+        {
+          "command": "notebook:delete-cell",
+          "icon": "@jupyterlab/cell-toolbar:delete",
+          "tooltip": "Delete Cell"
+        }
+      ]
+    },
+    "leftSpace": {
+      "title": "Cell toolbar empty left space size",
+      "description": "Size of the empty space left of the cell toolbar in px. Ignored if `floatPosition` is not null.",
```
krassowski, line 102: To make life of translators easier:

```suggestion
      "description": "Size of the empty space left of the cell toolbar in pixels. Ignored if `floatPosition` is not null.",
```

packages/cell-toolbar/style/icons/addabove.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_137_19492)">
```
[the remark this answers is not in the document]
jweill-aws: This icon was created by @WangWanyue on my team. Any recommendations for tools or settings to create SVG files that match those already in JupyterLab?

jasongrout:
> when multiple cells are selected, the toolbar acts on the entire selection rather than on a single cell (which is good!), however the duplicate action produces different results depending on which way the selection was created: 

My guess is what is happening here is that the new cells are added after the active cell, so in your second example the two new cells are positioned right after the active top cell, but above the bottom selected cell. If that is the case, you're right that instead, the new cells should be added after the last selected cell. This is likely a problem in the underlying command, not the button in this toolbar.

jweill-aws:
> This is likely a problem in the underlying command, not the button in this toolbar.

@jasongrout This may be true, but this PR _introduces_ the command for duplicating cells, so it seems like a decent place to introduce a fix for this issue as well. 😄

jweill-aws:
Updates on reported flaws by @krassowski:

1. Streamlined the schema to reduce test cases. This includes the elimination (at least for now) of the "helper buttons" that would appear to the left of the input area in desktop browsers.
2. Mobile cell toolbar: Modified CSS for small-screen layouts to position the toolbar above the cell and to show it irrespective of cell content.
3. Duplicate cells bug: the new `notebook:duplicate-below` command calls `NotebookActions.copy` and `NotebookActions.paste(..., 'below')` in that order. For this reason, the bug will occur even in the case where you highlight two cells from bottom to top, copy them, and paste them. I'll file a separate bug about this.

jweill-aws:
@krassowski @jasongrout I updated the "Duplicate Cells Below" command to not use the clipboard and to add copies of the selected cells _below the current selection_ irrespective of which cell is active.

blink1073:
I'm +1 on a manual publish beforehand, that makes things less guess-and-check

krassowski:
Thanks for the changes! I also concur with the point raised earlier that the disappearing toolbar can feel weird. I would have expected it to instead collapse to an "expand" arrow button as in the first alternative design idea in https://github.com/jupyterlab/jupyterlab/issues/9733#issuecomment-1028275057:

Screenshot from 2022-03-15 22-41-19

(if we are not including the on-hover positioning from the primary design in https://github.com/jupyterlab/jupyterlab/issues/9733#issuecomment-1028275057. Did you settle on a solution earlier/during a team meeting or something?

jweill-aws:
> I'm +1 on a manual publish beforehand, that makes things less guess-and-check

@blink1073 I didn't find info about publishing a new package in the JupyterLab docs. Do I need special permission to publish to NPM, or is there a shared credential I should use?

blink1073:
What is your `npm` username?  I will add you to the org.

jweill-aws:
@blink1073 It's jweill-aws, same as my Github username. Just created my NPM account. Thanks for your help!

blink1073:
Invite sent.  Since we have the publish config set in the packages (see below), all you need to do is make sure the packages are built, and then run `npm publish`.

```
  "publishConfig": {
    "access": "public"
  },
```

jtpio:
Nice, we should follow a similar approach and publish the packages to npm first to get the CI to run, when we add new packages in the future (for example for https://github.com/jupyterlab/jupyterlab/pull/12095)

jtpio:
Looks like one failing UI test is relevant:

image

packages/cell-toolbar-extension/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@jupyterlab/cell-toolbar-extension",
+  "version": "4.0.0-alpha.6",
```
jtpio, line 3: FYI `4.0.0a22` has been released, so this (and other references) can be re-aligned to:

```suggestion
  "version": "4.0.0-alpha.7",
```

jtpio: However it shouldn't be needed to publish again to npm.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+ */
+const TOOLBAR_OVERLAP_CLASS = 'jp-toolbar-overlap';
+
+/**
+ * Icons for use in toolbar.
+ *
+ * These are copied from icon.ts, which is not part of the webpack bundle
+ * because nothing is imported from it.
+ */
+export const addAboveIcon = new LabIcon({
```
fcollonval, line 47: Let's move the icons to the @jupyterlab/ui-components package: packages/ui-components/src/icon/iconimports.ts

You will need to update the extension id.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    const cells = this._panel?.context.model.cells;
+    if (cells) {
+      cells.changed.disconnect(this.updateConnectedCells, this);
+      each(cells.iter(), model => this._removeToolbar(model));
+    }
+
+    this._panel = null;
+  }
```
fcollonval, line 141: Let's call the clear data to be sure.

```suggestion
    Signal.clearData(this);
  }
```

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+    this._isDisposed = true;
+
+    if (this._settings) {
+      this._settings.changed.disconnect(this._onSettingsChanged, this);
+    }
+
+    const cells = this._panel?.context.model.cells;
+    if (cells) {
+      cells.changed.disconnect(this.updateConnectedCells, this);
+      each(cells.iter(), model => this._removeToolbar(model));
```
fcollonval, line 137: If I get the code right, you optimize it to create the toolbar only for the active cell isn't it?

If so we should be able to be smarter than looping on all cells here.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (!activeCell) {
+      return;
+    }
+
+    if (
+      this._previousActiveCell !== null &&
+      this._previousActiveCell !== undefined
+    ) {
+      this._removeToolbar(this._previousActiveCell.model);
+    }
```
fcollonval, line 113: We should always remove the toolbar on the previous cell to be safe in case we change the logic that is possible to have no active cell.

I also simplify the test on the previous active cell.

```suggestion
    if (this._previousActiveCell) {
      this._removeToolbar(this._previousActiveCell.model);
    }
    
    const activeCell = notebook.activeCell;
    if (!activeCell) {
      return;
    }
```

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+    this._panel = null;
+  }
+
+  /**
+   * Callback to react to cells list changes
+   *
+   * @param cells List of notebook cells
+   * @param changed Modification of the list
+   */
+  updateConnectedCells(
```
fcollonval, line 149: Is listing to the cells list (and this callback) still need as you react to the active cell change?

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+
+// icon svg import statements
+import addAboveSvg from '../style/icons/addabove.svg';
+import addBelowSvg from '../style/icons/addbelow.svg';
+import deleteSvg from '../style/icons/delete.svg';
+import duplicateSvg from '../style/icons/duplicate.svg';
+import moveDownSvg from '../style/icons/movedown.svg';
+import moveUpSvg from '../style/icons/moveup.svg';
+
+const DEFAULT_LEFT_MENU: ICellMenuItem[] = [];
```
fcollonval, line 29: It would make sense to set the default list matches the one in the settings (in case of simpler remix not using the settings plugin).

packages/cell-toolbar/src/celltoolbarwidget.ts
```diff
@@ -0,0 +1,21 @@
+/* -----------------------------------------------------------------------------
+| Copyright (c) Jupyter Development Team.
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+import { CommandRegistry } from '@lumino/commands';
```
fcollonval, line 5: File to be removed

packages/cell-toolbar/src/tokens.ts
```diff
@@ -0,0 +1,31 @@
...
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+import { LabIcon } from '@jupyterlab/ui-components';
+
+export const EXTENSION_ID = '@jupyterlab/cell-toolbar';
+
+/**
+ * Menu action interface
+ */
+export interface ICellMenuItem {
```
fcollonval, line 12: Unneeded interface

packages/cell-toolbar/src/toolbarbutton.tsx
```diff
@@ -0,0 +1,99 @@
+/* -----------------------------------------------------------------------------
```
fcollonval, line 1: Unneeded file

packages/cell-toolbar/src/utils.ts
```diff
@@ -0,0 +1,15 @@
+/* -----------------------------------------------------------------------------
```
fcollonval, line 1: To be removed - looks unused.

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -583,6 +583,7 @@ export namespace ToolbarButtonComponent {
     iconLabel?: string;
     tooltip?: string;
     onClick?: () => void;
+    onClickWithEvent?: (e: Event) => void;
```
fcollonval, line 586: I would support adding it to the existing `onClick` to avoid yet another properties (as we already have `onClick` and `actualOnClick` - let's not add a third one).

But anyway this will need to be revisited when the `Toolbar` widget is used for the cell toolbar.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,372 @@
...
+
+  private _addToolbar(model: ICellModel): void {
+    const cell = this._getCell(model);
+
+    if (cell) {
+      const { leftMenu } = (this._settings?.composite as any) ?? {};
+
+      const leftMenu_ = leftMenu === null ? [] : leftMenu ?? DEFAULT_LEFT_MENU;
+
+      const toolbar = new CellToolbarWidget(this._commands, leftMenu_);
```
fcollonval, line 173: Let's reuse the `Toolbar` widget and use the following logic to populate it: https://github.com/jupyterlab/jupyterlab/blob/master/packages/docregistry/src/default.ts#L432

```suggestion
      const toolbar = new Toolbar();
```

fcollonval (changes requested):
@jweill-aws I did a first review of the code. There are some refactoring possible by reusing existing code (not used in the extension) and due to some optimization as only one toolbar for the active cell is displayed at once.

packages/cell-toolbar-extension/schema/plugin.json
```diff
@@ -0,0 +1,76 @@
...
+          "icon": "@jupyterlab/cell-toolbar:delete",
+          "tooltip": "Delete Cell"
+        }
+      ]
+    }
+  },
+  "additionalProperties": false,
+  "type": "object",
+  "definitions": {
+    "menuItem": {
```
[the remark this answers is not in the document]
jweill-aws: Is there a way to refer to the `notebook-extension`'s definition of `toolbarItem`? If not, should I copy over the definition as it exists in this file?

packages/ui-components/src/components/toolbar.tsx
jweill-aws: Using `actualOnClick` appears to work. Reverted the `onClickWithEvent` prop. Thanks!

jweill-aws:
@fcollonval If you have time this morning, can you please take a look at the related WIP PR https://github.com/jweill-aws/jupyterlab/pull/4 ? I spent some time last week trying to figure out how to make the `jupyter.lab.toolbars` key work with the `cell-toolbar` code, but I'm not sure how to access the contents of the new `"Cell"` toolbar. Thanks.

fcollonval:
Let me know if you want to discuss about it.

jweill-aws:
@fcollonval Thank you very much! I'll take a look at your branch and incorporate the changes. I really appreciate your help.

jweill-aws:
Merged in my and @fcollonval 's changes to use the `toolbarFactory` pattern

packages/cell-toolbar/src/celltoolbartracker.ts
jweill-aws: I agree that there should be only one cell with a toolbar, but just in case the extension got into a weird state (e.g., due to another extension's conflict) I decided that `cell-toolbar` should loop over all cells only in the `dispose` method.

jweill-aws:
Now seeing an error with the "Duplicate below" command: when duplicating a cell, the duplicate appears at the very beginning of the notebook, not below the selection.

fcollonval:

jweill-aws:
The following documentation images do not show the cell toolbar, but might need to:
file_formats_html_display.png
file-formats-html-display.png
includeOutput.gif
rtc_shared_cursors.png
shared_cursors.png
step.gif
sync_editors.png
syncCollapseState.gif
toc.gif
tutorial-notebook.png

I don't have the resources to make animated GIFs like the ones above, but I can try to update the static images.

jweill-aws:
please update snapshots

jweill-aws:
All snapshots have been updated — anything else to modify?

packages/cell-toolbar/src/svg.d.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Jupyter Development Team.
```
fcollonval, line 1: This file can be removed

packages/cell-toolbar/src/toolbarbutton.tsx
fcollonval, line 1: This file can be removed

packages/cell-toolbar/test/celltoolbartracker.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      commands.addCommand('notebook:move-cell-up', {
+        execute: args => null
+      });
+      commands.addCommand('notebook:move-cell-down', {
+        execute: args => null
+      });
+
+      settings = new TestSettings();
+
+      extension = new CellBarExtension(commands, settings);
```
fcollonval, line 62: This should be updated to reflect the new signature of `CellBarExtension`

packages/cell-toolbar/test/celltoolbartracker.spec.ts
```diff
@@ -0,0 +1,116 @@
...
```
fcollonval, line 95: Same here

fcollonval (changes requested):
Almost there @jweill-aws - see the latest comments

You should also revert the snapshots `galata/test/documentation/internationalization.test.ts-snapshots/language-chinese-documentation-linux.png`. It should not be updated (the new one is wrong as the bottom has a gray area below the status bar).

jweill-aws:
@fcollonval Conflict resolved. Thanks!

fcollonval (approved):
Thanks a lot for the patience @jweill-aws

fcollonval:

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  private _cellToolbarLeft(activeCell: Cell<ICellModel>): number | null {
+    const toolbarWidgets = this._findToolbarWidgets(activeCell);
+    if (toolbarWidgets.length < 1) {
+      return null;
+    }
+    const activeCellToolbar = toolbarWidgets[0].node;
+
+    return activeCellToolbar.getBoundingClientRect().left;
```
krassowski, line 268: I wonder if `getBoundingClientRect` could be cached (if line has not changed) to prevent potential performance degradation in a mechanism of long layout recalculation as discussed in https://github.com/jupyterlab/jupyterlab/issues/9757. I meant to open an issue but went directly to this PR to see if we caught anything in the performance benchmarks (likely it would not be picked up as it requires user to focus a cell which may not be sufficiently covered by the current scenarios), but I do not see reference values in https://github.com/jupyterlab/jupyterlab/pull/12028#issuecomment-1097874513; is this expected @fcollonval ?

fcollonval: Good catch - the second run (on the PR verison) is missing the update of the reference snapshots. I will correct that. Unfortunately the [weekly benchmark](https://github.com/jupyterlab/benchmarks/issues/85#issuecomment-1100791897) will not provide much insight for this specific PR because the [PR turn off virtual rendering](https://github.com/jupyterlab/jupyterlab/pull/12258) was also merged the same week.

Issue #12029, opened 2022-02-09
Debounce the kernel sources filter


echarles:
### Problem

https://github.com/jupyterlab/jupyterlab/pull/11566 has introduced a new debugger panel that show the available sources in the kernel. That list can be filtered, but as the list is typically large, every keystrokes requests a refresh of the panel, which can be slow and not user-friendly

### Proposed Solution

Debounce the keystrokes in the filter box.

krassowski:
I think we can improve rendering time without introducing debouncer; we still need to do that for #12012. I believe that we should just render a custom list widget with `<li>` items (like in file manager) instead of using tons of individual widgets (as those have too much overhead).

I feel that debouncer should be more like a last resort solution when something cannot be made snappy with other means because it takes away from interface responsibility by design.

krassowski:
> but still a debouncer for a not-light action (filtering and displaying potentially thousands of entries) deserves a debouncer IMHO

Possibly. For now, could we settle on making it 250ms or 500ms instead of 1000ms?

echarles:
> Possibly. For now, could we settle on making it 250ms or 500ms instead of 1000ms?

I have now explicitly set to 500ms which is the default IIRC

PR #12030, opened 2022-02-09
Debounce kernel sources filter

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12029

## Code changes

Add a debouncer that update the kernelsources after a filter change with a 1 second latency on inactivity.

## User-facing changes

Untitled4

None.

### DIFF
--- a/packages/debugger/src/panels/kernelSources/filter.tsx
+++ b/packages/debugger/src/panels/kernelSources/filter.tsx
@@ -3,7 +3,7 @@
 
 import { InputGroup, ReactWidget, UseSignal } from '@jupyterlab/ui-components';
 
-import React, { useState } from 'react';
+import React from 'react';
 
 import { IDebugger } from '../../tokens';
 
@@ -15,21 +15,16 @@ export interface IFilterBoxProps {
 }
 
 const FilterBox = (props: IFilterBoxProps) => {
-  const [filter, setFilter] = useState('');
-  props.model.filterChanged.connect((_, filter) => {
-    setFilter(filter);
-  });
   const onFilterChange = (e: any) => {
     const filter = (e.target as HTMLInputElement).value;
     props.model.filter = filter;
-    setFilter(filter);
   };
   return (
     <InputGroup
       type="text"
       onChange={onFilterChange}
       placeholder="Filter the kernel sources"
-      value={filter}
+      value={props.model.filter}
     />
   );
 };
@@ -40,8 +35,8 @@ const FilterBox = (props: IFilterBoxProps) => {
 export const KernelSourcesFilter = (props: IFilterBoxProps): ReactWidget => {
   return ReactWidget.create(
     <UseSignal
-      signal={props.model.changed}
-      initialArgs={props.model.kernelSources}
+      signal={props.model.filterChanged}
+      initialArgs={props.model.filter}
     >
       {model => <FilterBox model={props.model} />}
     </UseSignal>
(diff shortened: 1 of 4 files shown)


packages/debugger/src/panels/kernelSources/filter.tsx
```diff
@@ -12,24 +12,24 @@ import { IDebugger } from '../../tokens';
...
 }
 
 const FilterBox = (props: IFilterBoxProps) => {
-  /**
-   * Handler for search input changes.
-  const handleChange = (e: React.FormEvent<HTMLElement>) => {
-    const target = e.target as HTMLInputElement;
-    setFilter(target.value);
-    props.model.filter = target.value;
+  const [filter, setFilter] = useState('');
```
fcollonval, line 18: You could keep the previous implementation and change the `UseSignal` L42 to listen for the new signal `filterChanged`. That will avoid duplicating filter state in the React component and in the model.

packages/debugger/src/panels/kernelSources/model.ts
```diff
@@ -19,6 +21,11 @@ const compare = (a: IDebugger.KernelSource, b: IDebugger.KernelSource) => {
  * The model to keep track of the current source being displayed.
  */
 export class KernelSourcesModel implements IDebugger.Model.IKernelSources {
+  constructor() {
+    this.refresh = this.refresh.bind(this);
+    this._debouncer = new Debouncer(this.refresh, 500);
```
fcollonval, line 26: It would be nicer to set a constant with explicit name rather than the timeout value.

fcollonval (changes requested):
Thanks @echarles here are some suggestions

packages/debugger/src/panels/kernelSources/filter.tsx
echarles: I remember I have done that to begin and came into issues where the filter content was not rended. I have suspected some weird interaction between the debouncer and the react.js refresh mecanism. I have spent some time to understand and try to fix that. I am going to try again with fresh eyes and will report. I have not found similar cases in the jupyterlab code base so it may be the first time this case is tried out.

echarles: oh rereading you comment... yes I had to create for a new signal otherwise the react component was not refreshing. I am going to try to use it to make the implementation cleaner.

echarles: React component state is cleaner using that new signal. Done, Thx.

packages/debugger/src/panels/kernelSources/model.ts
echarles: Done. Thx.

echarles:
Thx @fcollonval, I have implemented your suggestions.

CI failure (UI Tests / Visual Regression Documentation and [Linux (release_test, 3.10) failing with + python -m pip install
ERROR: You must give at least one requirement to install (see "pip help install")) are not relevant to this PR IMHO.

This is ready for review/merge.

echarles:
> This is ready for review/merge.

@fcollonval Would be great if you could review/merge this. Once done, I will need to backport to 3.3.x branch. Thx in advance.

fcollonval (approved):
Thanks @echarles

fcollonval:

echarles:
Thx @fcollonval for the merge and the backport PR 👍

PR #12043, opened 2022-02-12
Stabilize documentation CI tests

fcollonval:
## References

The integration tests for the documentation are flaky mainly due to the width computation of the main split layout panel (containing the two sidebars and the main dockpanel).

A first test forcing the width of the sidebars did not work as the main dock panel width may then be smaller during some frames. This is translated in screenshot by larger separator spaces than expected.
Second workaround is to wait for two frames before taking the snapshots. Two frames are at least required because in one frame, all the widgets are inserted. Then only the split layout computes the final width of its children as a function of the size constrains and the stretch parameters.


jtpio:
Thanks @fcollonval for starting this :+1:

galata/test/documentation/debugger.test.ts
```diff
@@ -21,6 +21,16 @@ test.describe('Debugger', () => {
       [generateCaptureArea({ top: 62, left: 1050, width: 190, height: 28 })]
     );
 
+
+    let toResolve;
+    const awaitTimeout = new Promise(resolve => {
+      toResolve = resolve;
+    });
+
+    requestAnimationFrame(() => {
```
jtpio, line 30: Looks like this gives the following:

```
ReferenceError: requestAnimationFrame is not defined
```

Maybe need to be wrapped in a `page.evaluate()`?

jtpio:
Many thanks @fcollonval for the patience.

Looks like CI is green on the latest run? Maybe we should kick it just to double check?

There is also a small conflict to fix, but otherwise this could be good to go.

jtpio:
@fcollonval I pushed a commit to fix a minor typo, and CI looks green now.

jtpio (approved):
Diff looks good, happy to undraft and merge.

Thanks a lot for working on this!

Issue #12042, opened 2022-02-11
Shutdown from SessionContext doesn't handle errors


Zsailer:
## Description

If an error occurs when calling `SessionContext.shutdown()`, the status indicator for the kernel still says `terminating` even though kernel is not in a properly terminating state.

That's because the status is changed *before* the server request is made:

https://github.com/jupyterlab/jupyterlab/blob/c5df985383964b027e475e667d69a7e8facf42bb/packages/apputils/src/sessioncontext.tsx#L768-L769

This is a problem for pending kernels, i.e. `use_pending_kernels=True` in Jupyter Server. If one attempts to shutdown a kernel in a pending state, it raises a `RuntimeError` and blocks the shutdown request. (The kernel can be shutdown once it has finished its pending state.) The status indicators still say "terminating" after the request is blocked, but that's  not accurate. 

Instead if the shutdown request returns an error, the status should go back to its original status before the shutdown request was made.

## Reproduce

Reproducing takes a bit of work :laughing:.

1. Clone the jupyter server repo.
2. Artificially add an Exception to the "DELETE" session handler, preventing the shutdown request from happening. 
3. Open JupyterLab and start a kernel
4. Go to the Kernel > "Shutdown Kernel" menu item
5. Observe that the kernel didn't shutdown, but the status indicator still says "terminating".

## Expected behavior

If the shutdown request returns an error, the status should go back to its original status before the shutdown request was made.

## Context

- Operating System and version: Mac OS
- Browser and version: Chrome 97.x
- JupyterLab version: 3.2.x and above

PR #12048, opened 2022-02-14
Handle shutdown error

Zsailer:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12042

## Code changes

Catch errors raised when attempting to shutdown a kernel through the SessionContext interface. If an error is raised, the SessionContext is reset back to its original status. 

## User-facing changes

This won't affect most users. This really becomes important in a remote kernel context, where kernel shutdown failures are more common.

Specifically, if [pending kernels](https://jupyter-client.readthedocs.io/en/stable/pending-kernels.html) are enabled, an [error is raised when a "pending kernel" is shutdown](https://github.com/jupyter/jupyter_client/blob/10decd25308c306b6005cbf271b96493824a83e8/jupyter_client/multikernelmanager.py#L260-L261). Without this change, the UI shows that the kernel is terminating when it's actually in a pending state.

None 

This change should be backported to 3.3.x branch.

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -763,6 +763,9 @@ export class SessionContext implements ISessionContext {
    */
   private async _shutdownSession(): Promise<void> {
     const session = this._session;
+    // Capture starting values in case an error is raised.
+    const isTerminating = this._isTerminating;
+    const isReady = this._isReady;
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM!

fcollonval (approved):
Thanks @Zsailer

fcollonval:

PR #11814, opened 2022-01-10
Fix for Code Mirror width. Indentation changes made

pree-T:
Fix for the Code Mirror Width  issue #11696 . Indentation changes implemented.

Issue #4208, opened 2018-03-19
Keep horizontal scroll bar position in output fixed.


ma-ji:
Currently, the horizontal scroll bar in notebook output is affixed to the end of the output, so the user needs to scroll vertically to the end to be able to move horizontally. Not convenient especially when you have a very long output.

jasongrout:
Can you post a screenshot showing where you think it makes more sense to have the horizontal scrollbar?

timkpaine:
duplicate of https://github.com/jupyterlab/jupyterlab/issues/2992

timkpaine:
which was closed by @ellisonbg but is still an issue

jasongrout:
> but is still an issue

Are you saying it's still an issue when someone enables scrolling for an output?

jasongrout:
> Can you post a screenshot showing where you think it makes more sense to have the horizontal scrollbar, versus where it is now?

Maybe instead, can you post instructions for reproducing this? Is it indeed the same as #2992?

ma-ji:
> Can you post a screenshot showing where you think it makes more sense to have the horizontal scrollbar, versus where it is now?
>Maybe instead, can you post instructions for reproducing this? Is it indeed the same as #2992?

Probably the same. Instruction for reproducing:
```Python
import pandas as pd
pd.read_csv('https://raw.githubusercontent.com/ma-ji/RICF/v2015.1.0/2015/2015_financial_activities.tsv', sep='\t', header=1)
```
See the GIF replication.
![scroll_horizontal](https://user-images.githubusercontent.com/11808231/37742989-6ab266fe-2d3e-11e8-9bca-5c67c097e028.gif)

jasongrout:
Ah, thanks. Yes, I see how that's really annoying!

jasongrout:
Possibly the same issue, but with the fasta plugin:

<img width="1036" alt="screen shot 2018-07-07 at 1 51 15 pm" src="https://user-images.githubusercontent.com/192614/42414602-1ce9362e-81ed-11e8-80c3-5df7a52d8473.png">

Really, the bottom scrollbar shouldn't be there - the fasta plugin should be sized to the width of the notebook output.

jasongrout:
I looked at this for a while. I think this is going to require a more serious redesign of the notebook, given our current nesting structure.

dstanner:
[...]

PR #12077, opened 2022-02-17
Add scroll bar for horizontally overflowing, unshrinkable outputs

### LINKED ISSUE #12070: Horizontal overflow unreachable
## Description

Generally, JupyterLab tries to make sure everything is visible without the need for scrolling horizontally. But when scrolling horizontally is still necessary, e.g. to reach critical parts of the output, there are no horizontal scroll bars. This is especially prone to happening in the new fancy side-by-side renderer due to reduced width. See also: https://github.com/matplotlib/ipympl/issues/431

## Reproduce

```bash
pip install ipympl
```

```python
import matplotlib.pyplot as plt
%matplotlib widget
plt.figure(figsize=(20,8))
plt.plot()
```

Resulting figure will be too large, but also the resize handle will be unreachable, since it is in the hidden, unreachable overflow.

## Expected behavior

Show horizontal scroll bars when necessary to prevent overflow.

## Workaround solution

Set CSS option

```css
.jp-OutputArea-child {
  overflow-x: auto;
}
```

Note that this will not apply the scrolling at the same place as the vertical scroll bars in the DOM. A better solution would be to somehow get the overflow to propagate through all the flexboxes to the full `.jp-Cell-outputArea` where also the scroll bars in case of vertical scrolling option are, but I don't know enough CSS to get that to work.

## Context

- Browser and version: Firefox 96
- JupyterLab version: 3.2.8

mueslo:
## References

Fixes #12070

## Code changes

`.jp-OutputArea-child` overflow `auto` instead of `hidden`.

## User-facing changes

Adds a horizontal scroll bar for the case that horizontal overflow occurs.
image

In the general case of no overflow, no visual changes are present.

image

If output scrolling is enabled, the horizontal scrollbar will appear only once scrolled to the bottom (as it is in a different place in the DOM).

image
image

None.

### DIFF
--- a/packages/outputarea/style/base.css
+++ b/packages/outputarea/style/base.css
@@ -22,6 +22,7 @@
 .jp-OutputArea-child {
   display: flex;
   flex-direction: row;
+  overflow-x: auto;
 }
 
 body[data-format='mobile'] .jp-OutputArea-child {


fcollonval:
Thanks for opening this PR @mueslo 

The way changes are handled in JupyterLab is by always targeting the master branch and then backporting to production branch. This improves consistency and allows to backport easily to multiple branches if needed.

Could you please bring a PR against the master branch?

mueslo:
Sorry - if this procedure was mentioned somewhere, I must have missed it.
To me it looked like the master branch targeted 4.x. The master branch has
reworked the `outputarea` css (grid instead of flexbox, see #11508), but I
don't see it being backported. What course of action would you suggest?

fcollonval:
No problem - and thanks for pointing out that indeed this is not mentioned - I'll correct that.

As those CSS changes are not altering the look of the notebook, for this peculiar case we can backport #11508, #11814 and the up coming one.

mueslo:
> No problem - and thanks for pointing out that indeed this is not mentioned - I'll correct that.
> 
> As those CSS changes are not altering the look of the notebook, for this peculiar case we can backport #11508, #11814 and the up coming one.

Thanks. Once #11508 is backported I think a different method can be used to bring the outputs back which will show the horizontal scroll bar at the same location in the DOM as the vertical scroll bar, which would also solve #4208

Is there any timeline of backporting those changes? Or would I need to initiate it?

krassowski:
It looks like closing this PR and opening a new one against master branch is the way to go. @mueslo are you still interested?

PR #12080, opened 2022-02-17
Log error on open document widget.

trungleduc:
## References

On document opening, the widget is disposed if the context initialization fails.  Since no error is logged, debugging the issue from the external extension is very hard.

## Code changes
Log the error in the try-catch block.

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/docmanager/src/manager.ts
+++ b/packages/docmanager/src/manager.ts
@@ -592,6 +592,7 @@ export class DocumentManager implements IDocumentManager {
 
     // If the initial opening of the context fails, dispose of the widget.
     ready.catch(err => {
+      console.error(err)
       widget.close();
     });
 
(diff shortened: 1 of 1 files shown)


packages/docmanager/src/manager.ts
```diff
@@ -592,6 +592,7 @@ export class DocumentManager implements IDocumentManager {
 
     // If the initial opening of the context fails, dispose of the widget.
     ready.catch(err => {
+      console.error(err)
```
fcollonval, line 595: Adding some context to the error would be nice

```suggestion
      console.error(`Failed to initialize the context with '${factory.name}' for ${path}`, err)
```

fcollonval (changes requested):

fcollonval:
Could you fix the linter @trungleduc ?

trungleduc:
Done

jtpio:

Issue #11776, opened 2022-01-01
Enable extensions to use (modal) selector


krassowski:
### Problem

Multiple extensions would benefit from the ability to open a modal selector. While an option to select an option already exists using `InputDialog.getItem()`, it is limited in its capabilities to simple `<select>` widget which is less usable (does not support search, formatting, icons, etc). Having a unified selector would ensure consistent UX and reduce the duplication of effort between extensions.

> <img width="600px" src="https://user-images.githubusercontent.com/5832902/147851507-7e652917-e6a1-4f62-abf6-c0253fffccc3.gif">
>
> `InputDialog.getItem` does not allow me to dynamically render the context of the references in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/739: because I need to retrieve contents of files without blocking the UI in order to quickly show the jump targets I need a selector more advanced than `<select>` and an ability to update the contents asynchronously; ideally I could render larger context dynamically for the selected option.

Currently the core implements the modal selector for command palette basing on the lumino internals. However, the underlying selection logic is not available to extension. It leads to certain peculiar behaviours like https://github.com/jupyterlab/jupyterlab/issues/9659.

The following extensions would benefit from the proposed selector capabilities:
- [`jupterlab-citation-manager`](https://github.com/krassowski/jupyterlab-citation-manager): for selecting citations, references, and citation styles,
- [`jupyterlab-lsp`](https://github.com/jupyter-lsp/jupyterlab-lsp): for selecting jump targets (references), for selecting for navigation symbols,
- [`jupyterlab-search-replace`](https://github.com/jupyterlab-contrib/search-replace): for opening search matches across files,
- [`jupyterlab-quickopen`](https://github.com/parente/jupyterlab-quickopen): for opening files from search results

### Proposed Solution

I propose we include the `Selector` and `ModalSelector` described below in `ui-components`, and re-write the modal palette using to use `ModalSelector`.

[...]

PR #12082, opened 2022-02-17
Add support for filtering by field names in setting editor

### LINKED ISSUE #11990: Add searching by individual setting names in settings editor search
The search bar in the settings editor currently only searches by the name of the plugins but users will probably want to search based on the titles (and maybe descriptions?) of individual settings within a plugin setting. Adding that would help people navigate / discover settings much more easily.

marthacryan:
**Note**: Some fields that are within more complex object fields still aren't showing up. Search "unique" to test - should show the notebook panel settings, but nothing shows up. This probably isn't blocking / I might find the issue. 

**Todo**: 
- [ ] Make the part of the entry that matches the search be in bold font. 

## Questions
* Should we search by description as well as field title? 

## References
Fixes #11990. Currently the search in the plugin list only filters by name of plugin, but this adds filtering by names of fields.

## Code changes
* Adds a signal to update the settings fields shown when the search value updates. 
* Adds logic to find matching field names when the filter updates. 

## User-facing changes
* Adds a list of matching fields under each entry in the plugin list. 
* Shows the corresponding filtered fields in the actual settings editor display. 
* **Note**: If the name of the plugin matches the search, it doesn't check for matching fields. 
image

None


fcollonval:
> Note: If the name of the plugin matches the search, it doesn't check for matching fields.

I would search for matching field in all cases

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -237,10 +250,112 @@ export class PluginList extends ReactWidget {
...
+    }
+
+    // Make sure props is non-empty before calling reduce
+    if (Object.keys(props).length === 0) {
+      return [];
+    }
+
+    // Iterate through the properties and check for titles / descriptions that match search.
+    return reduce(
+      Object.keys(props),
```
fcollonval, line 304: You could use `Array.reduce` directly

image

```suggestion
    // Iterate through the properties and check for titles / descriptions that match search.
    return Object.keys(props).reduce(
```

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -237,10 +250,112 @@ export class PluginList extends ReactWidget {
...
+  }
+
+  /**
+   * Updates the filter when the search bar value changes.
+   * @param filter Filter function passed by search bar based on search value.
+   */
   setFilter(filter: (item: string) => boolean): void {
-    this._filter = (value: ISettingRegistry.IPlugin) => {
-      return filter(value.schema.title?.toLowerCase() ?? '');
+    this._filter = (plugin: ISettingRegistry.IPlugin): string[] | boolean => {
```
fcollonval, line 347: I would use a signature `(plugin: ISettingRegistry.IPlugin): string[] | null`

So you don't have to test for the return type.

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -298,20 +424,26 @@ export class PluginList extends ReactWidget {
...
-        plugin =>
-          !this._modifiedPlugins.includes(plugin) && this._filter(plugin)
-      )
+      .filter(plugin => {
+        const filtered = this._filter(plugin);
+        return (
+          !this._modifiedPlugins.includes(plugin) &&
+          (typeof filtered === 'boolean' ? filtered : filtered.length > 0)
+        );
+      })
```
fcollonval, line 439: Is it possible to change this part to avoid filtering twice a plugin.

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -321,19 +453,34 @@ export class PluginList extends ReactWidget {
...
+        {otherItems.length > 0 && (
+          <div>
+            <h1 className="jp-PluginList-header">{trans.__('Settings')}</h1>
+            <ul>{otherItems}</ul>
+          </div>
+        )}
+        {modifiedItems.length === 0 && otherItems.length === 0 && (
+          <h1 className="jp-PluginList-noResults">
+            {' '}
+            No items match your search.{' '}
```
fcollonval, line 465: This needs to be wrapped through `trans.__` to be translated

```suggestion
            {trans.__('No items match your search.)'}
```

packages/settingeditor/src/settingspanel.tsx
```diff
@@ -79,6 +94,25 @@ export const SettingsPanel: React.FC<ISettingsPanelProps> = ({
     [id: string]: boolean;
   }> = React.useRef({});
 
+  // When filter updates, only show plugins that match search.
+  updateFilterSignal.connect(
```
fcollonval, line 98: You need to wrap this in a `useEffect` (you can reuse the one below).

The current code will be called at every react update cycle. So it will create multiple connections without cleaning them. Use `useEffect` with empty array as second argument (as the one L116) will connect the signal only when mounting the component.

> The `useEffect` should return a callback that is disconnecting the signal. That callback is triggered when unmounting the component (as the second argument of `useEffect` is an empty array).

packages/settingeditor/src/settingspanel.tsx
```diff
@@ -107,7 +141,9 @@ export const SettingsPanel: React.FC<ISettingsPanelProps> = ({
 
   return (
     <div className="jp-SettingsPanel" ref={wrapperRef}>
-      {settings.map(pluginSettings => {
+      {shownPlugins.map(pluginSettings => {
+        // Pass filtered results to SettingsFormEditor to only display filtered fields.
+        const filtered = filterPlugin(pluginSettings.plugin);
```
fcollonval, line 146: This is suboptimal because you are requesting again the plugins although the filter that produces `shownPlugins` has already been called earlier.

You should change the logic to store the filtered fields per plugin map as state. So you don't need to apply the filter except when `updateFilterSignal` is emitted.

fcollonval (changes requested):
Thanks @marthacryan 

Here are some code structure comments.
Would you mind adding some tests.

fcollonval:
>  Make the part of the entry that matches the search be in bold font.

As much as I don't like the current search highlighting style, it may be good to use it here to have a more homogeneous look and feel.

pinging @isabela-pf

krassowski (changes requested):
Thank you @marthacryan. I left a few suggestions the code.

In addition:
- I think that we should also search in property keys like `numberCellsToRenderDirectly` and maybe also display those stylised as code; there are hundreds of GitHub issues, Discourse posts and Stack Overflow answers referencing these keys and users may expect to find them (yes, there is the "JSON Settings Editor" button, but it may still frustrate user if they do not find the setting they are looking for immediately). What do you think?
- There is jitter when selecting items:
   jitter
   it appears to be due to https://github.com/jupyterlab/jupyterlab/blob/047fbe7c2772bc34503207b7724216f9e0bbcfc3/packages/settingeditor/style/base.css#L130-L132 which is there to offset `jp-SelectedIndicator` width; if it would simplify matters to always display the `jp-SelectedIndicator` I would be happy with such a solution.
- for highlighting matching fragments we could use [`StringExt`](https://jupyterlab.github.io/lumino/algorithm/modules/stringext.html)`.highlight` and `StringExt.matchSumOfSquares` from `@lumino/algorithm` [like in the completer](https://github.com/jupyterlab/jupyterlab/blob/76d6b1ed0e16e348363228dbf018623f68922081/packages/completer/src/model.ts#L392-L431).

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -279,39 +365,54 @@ export class PluginList extends ReactWidget {
...
-        ) : null}
-        <LabIcon.resolveReact
-          icon={icon || (iconClass ? undefined : settingsIcon)}
-          iconClass={classes(iconClass, 'jp-Icon')}
-          title={iconTitle}
-          tag="span"
-          stylesheet="settingsEditor"
-        />
-        <span>{title}</span>
+        <div>
```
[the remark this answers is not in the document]
marthacryan: Oh ok I'll remove `width: -webkit-fill-available`, not sure why that was in either. 

Are there accessibility considerations for using `div` instead of the `button` tag? Should we add an aria role of some kind?

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -237,10 +224,108 @@ export class PluginList extends ReactWidget {
...
+      return [];
+    }
+
+    // Iterate through the properties and check for titles / descriptions that match search.
+    return Object.keys(props).reduce((acc: string[], value: any) => {
+      // If this is the base case, check for matching title / description
+      const subProps = props[value] as PartialJSONObject;
+      if (
+        !subProps &&
+        (filter((props.title as string)?.toLocaleLowerCase() ?? '') ||
```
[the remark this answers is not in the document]
marthacryan: Sorry I think I might not understand what you're looking for - are you thinking that we'd add a flag to the `FilterBox` component to indicate if the search should be case sensitive?

marthacryan:
Thanks @krassowski! Still working through some of your review but I pushed a couple fixes

> * I think that we should also search in property keys like `numberCellsToRenderDirectly` and maybe also display those stylised as code; there are hundreds of GitHub issues, Discourse posts and Stack Overflow answers referencing these keys and users may expect to find them (yes, there is the "JSON Settings Editor" button, but it may still frustrate user if they do not find the setting they are looking for immediately). What do you think?

I believe I already am (let me know if I'm missing something) - however if it matches the property key, I display the title for that property if one exists, for readability. See pluginlist.tsx line 284. 

> * There is jitter when selecting items:
>   jitter
>   it appears to be due to https://github.com/jupyterlab/jupyterlab/blob/047fbe7c2772bc34503207b7724216f9e0bbcfc3/packages/settingeditor/style/base.css#L130-L132
>    which is there to offset `jp-SelectedIndicator` width; if it would simplify matters to always display the `jp-SelectedIndicator` I would be happy with such a solution.

Just pushed a fix for the jitter - let me know if that looks ok?

marthacryan:
Thank you for the review @fcollonval! Do the updates look okay to you?

marthacryan:
> * for highlighting matching fragments we could use [`StringExt`](https://jupyterlab.github.io/lumino/algorithm/modules/stringext.html)`.highlight` and `StringExt.matchSumOfSquares` from `@lumino/algorithm` [like in the completer](https://github.com/jupyterlab/jupyterlab/blob/76d6b1ed0e16e348363228dbf018623f68922081/packages/completer/src/model.ts#L392-L431).

@krassowski I'm having trouble implementing this because I only have access to the filter function, not the actual search query. Is this something I should change in the `FilterBox` component?

packages/settingeditor/src/pluginlist.tsx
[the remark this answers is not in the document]
krassowski: Yes, I think that adding an aria role would be appropriate when replacing the `button` with a `div`. I think that [`tab` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) is appropriate here.

krassowski:
> > * for highlighting matching fragments we could use [`StringExt`](https://jupyterlab.github.io/lumino/algorithm/modules/stringext.html)`.highlight` and `StringExt.matchSumOfSquares` from `@lumino/algorithm` [like in the completer](https://github.com/jupyterlab/jupyterlab/blob/76d6b1ed0e16e348363228dbf018623f68922081/packages/completer/src/model.ts#L392-L431).
> 
> @krassowski I'm having trouble implementing this because I only have access to the filter function, not the actual search query. Is this something I should change in the `FilterBox` component? 

This is interesting, I was not fully aware how strictly `FilterBox` separates concerns. If we are targeting 4.0 I think we should consider improvements to `FilterBox` API. I see that `FilterBox` does already use `StringExt.matchSumOfDeltas` under the hood for `fuzzySearch`:

https://github.com/jupyterlab/jupyterlab/blob/2e4411ea4efd0bbf524bb927062a19eae6bfbb6e/packages/ui-components/src/components/search.tsx#L52

but the result is discarded rather than exposed for rendering:

https://github.com/jupyterlab/jupyterlab/blob/2e4411ea4efd0bbf524bb927062a19eae6bfbb6e/packages/ui-components/src/components/search.tsx#L120-L125

Somewhat related, I previously proposed an API which facilitates working with scores metadata in https://github.com/jupyterlab/jupyterlab/issues/11776 - that would be the simplest approach IMO; the implementation in [jupyterlab-citation-manager's selector](https://github.com/krassowski/jupyterlab-citation-manager/blob/15215df002d7b3b22e968980530d99e32329282f/src/components/citationSelector.tsx#L23-L36) also uses the matches for highlighting.

As an incremental improvement for the `FilterBox` API the current `updateFilter` signature:

https://github.com/jupyterlab/jupyterlab/blob/2e4411ea4efd0bbf524bb927062a19eae6bfbb6e/packages/ui-components/src/components/search.tsx#L16

could be changed to:

```typescript
updateFilter: (filterFn: (item: string) => IScore | null) => void; 
```

I would however ask @fcollonval and @echarles for a second opinion though.

echarles:
> I would however ask @fcollonval and @echarles for a second opinion though.

I have implemented a filter box on the file browser and in the debugger kernel sources. At that time, I had to manually implement the fuzzy search, but now it looks that the filter component provides it.

In general, I would love to have more of those components having well documented API that you can reuse. 

At some point, the interactions between the model and the component update was giving me difficulties. How the signals would trigger an update (and a refresh) is something to think about when designing such components.

@marthacryan For this specific PR, is there anything remaining to solve?

echarles:
> In general, I would love to have more of those components having well documented API that you can reuse.

To give example, the filter box should have properties to define the type of search (normal, fuzzy...), the debouncing wanted, mask (for date...), formatting (render with currency)...

packages/settingeditor/src/pluginlist.tsx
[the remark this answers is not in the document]
marthacryan: Added it - thanks!

marthacryan:
@krassowski I just pushed a commit that adds highlighting and shows the key as well as the title of the property if it matches the search result. I updated the API of the FilterBox to add an optional `query` parameter to the `updateFilter` function that passes the filter back to the parent component. 

I removed the description matching but I can add that back in if you think it'd be helpful. I wasn't sure how to display the matching parts of the description so if you think that should be included, suggestions on the UI would be welcome. I think I addressed the rest of your feedback but let me know if I missed anything. Thanks!

Here's the updated look: 
image

packages/settingeditor/src/settingspanel.tsx
```diff
@@ -79,6 +93,18 @@ export const SettingsPanel: React.FC<ISettingsPanelProps> = ({
...
+  React.useEffect(() => {
+    // When filter updates, only show plugins that match search.
+    updateFilterSignal.connect(
+      (
+        list: PluginList,
+        newFilter: (plugin: ISettingRegistry.IPlugin) => string[] | null
+      ) => {
+        setFilter(() => newFilter);
+      }
+    );
```
fcollonval, line 105: You can move it in the below `useEffect`. And please set up a disconnect in the returned callback (see example below).

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -279,61 +389,82 @@ export class PluginList extends ReactWidget {
...
+        {otherItems.length > 0 && (
+          <div>
+            <h1 className="jp-PluginList-header">{trans.__('Settings')}</h1>
+            <ul>{otherItems}</ul>
+          </div>
+        )}
+        {modifiedItems.length === 0 && otherItems.length === 0 && (
+          <h1 className="jp-PluginList-noResults">
+            {trans.__('No items match your search.')}
+          </h1>
```
fcollonval, line 450: Do we want that to be a `h1` element. I would switch for a `p` element.

packages/settingeditor/src/settingspanel.tsx
marthacryan: Good point! Done

krassowski:
Looking nice! I think we can iterate on improvements later. The only major inconvenience I encountered is problems with lowercase/uppercase handling:

1. Filtering excludes items if the letter case does not match:

    Screenshot from 2022-03-15 22-19-34

    Screenshot from 2022-03-15 22-19-24

    I would argue that we should allow any lettercase here

2. On the contrary, if we use the same letter case as in the titles the highlighting does not work:

    Screenshot from 2022-03-15 22-21-15

    Screenshot from 2022-03-15 22-21-53

marthacryan:
@krassowski Oh good catch! I just pushed a fix, let me know if that works.

packages/settingeditor/style/base.css
```diff
@@ -444,16 +461,32 @@ ul.jp-PluginList li.jp-mod-selected span.jp-PluginList-icon.jp-FileIcon {
...
   text-overflow: ellipsis;
   white-space: nowrap;
-  width: -webkit-fill-available;
+}
+
+.jp-PluginList-entry li {
+  margin-left: 27px;
+  margin-top: 5px;
+  color: var(--jp-ui-font-color1);
+  width: fit-content;
```
krassowski, line 485:
```suggestion
  overflow-x: hidden;
  text-overflow: ellipsis;
```

| Before | After |
| --- | --- |
| Screenshot from 2022-03-20 00-47-07 | Screenshot from 2022-03-20 00-46-41 |

packages/settingeditor/style/base.css
```diff
@@ -444,16 +461,32 @@ ul.jp-PluginList li.jp-mod-selected span.jp-PluginList-icon.jp-FileIcon {
...
+
 .jp-PluginList-entry {
   display: flex;
-  flex-direction: row;
+  flex-direction: column;
   border: 1px solid transparent;
   background: transparent;
   overflow: hidden;
   padding: 4px 0 4px 4px;
   text-overflow: ellipsis;
```
krassowski, line 477: In addition to the suggestion in https://github.com/jupyterlab/jupyterlab/pull/12082/files#r830543410, this `text-overflow: ellipsis;` needs to be moved to the span element which is inside of `.jp-pluginList-entry-label` or otherwise it has no effect.

Let's add:

```css
.jp-pluginList-entry-label span {
  text-overflow: ellipsis;
  overflow-x: hidden;
  white-space: nowrap
}
```

and remove it from here.

```suggestion
```

| Before (ellipsis has no effect, `jp-SelectedIndicator` is squashed) | After |
| --- | --- |
| Screenshot from 2022-03-20 00-54-46 | Screenshot from 2022-03-20 00-53-54 |

krassowski (approved):
Thank you! I added two tiny suggestions on styling.

fcollonval:
Thanks @marthacryan 

The CI error is relevant:

```
Error: packages/settingeditor/src/settingspanel.tsx:80:11 - error TS7011: Function expression, which lacks return-type annotation, implicitly has an 'any' return type.

80   >(() => (plugin: ISettingRegistry.IPlugin) => {
```

marthacryan:
@fcollonval Fixed it, thanks!

isabela-pf (approved):
This looks really good to me! A great addition with the UX and styling I expected. Thank you so much for continuing to improve the settings editor @marthacryan 🎉 

I have two non-blocking comments:
- I'm glad to find it mostly matches the behavior (visually and UX) of the filter in the file browser. (And if I pretend like I can read typescript effectively, I think it's using some of the same components? Yay!) 
- Since this filter also shares behavior with the command palette (the bold text for filtered matches), I would like to have a similar highlight/visual focus for navigating the list of results in the future. I won't block this PR over it because the file browser is currently missing the same thing, so it feels like a separate fix.

Thanks again!

fcollonval:

PR #12122, opened 2022-02-25
Remove duplicated shortcuts

### LINKED ISSUE #12103: Many "Skipping this default shortcut because it collides with another default shortcut." warnings
## Description

Opening JupyterLab results in dozens of warnings like:

Screenshot from 2022-02-23 18-32-57

The shortcuts it complains about are:
- `Accel Enter` for `notebook:run-cell` added in #11942 CC @jtpio; I am not sure about this one, and would appreciate if you could look into this
- `Shift Accel ,` for `settingeditor:open-json` added in #11079 CC @marthacryan - this conflicts with new settings UI which also has the same shortcut assigned and the same selector; we need to change it; up to discussion how.

It could be either:
- genuine mistake in the PRs mentioned above
- a regression in logic as a collateral of changes added in #11079.

I am not sure which it is. This warning and logic was added in #9858 (CC @jasongrout).

Note: we are currently working around this check in #12091.

## Reproduce

1. Open [3.3.0rc0](https://mybinder.org/v2/gh/jupyterlab/jupyterlab/136121b05a5f739a1e4e78e44857d6bbe3780a01?urlpath=lab/tree)
2. Open developer console
3. See errors

## Expected behavior

No warnings; all shortcuts get activated.

## Context

- Browser and version: Chrome 98
- JupyterLab version: [3.3.0rc0](https://mybinder.org/v2/gh/jupyterlab/jupyterlab/136121b05a5f739a1e4e78e44857d6bbe3780a01?urlpath=lab/tree) and master

fcollonval:
## References

Fixes #12103

## Code changes

Remove shortcut for JSON editor (it was ignored as already defined).
Convert `Accel Enter` shortcut to `Cmd Enter` to run cells (this will be available only on MacOS)
Ignore shortcuts that contains explicitly `Cmd` on non-MacOS platforms. This will avoid creating unwanted shortcuts. Otherwise the `Cmd Enter` is converted to `Enter` shortcut on Windows and Linux.

## User-facing changes

`Cmd Enter` and `Ctrl Enter` should work on Mac OS.

N/A

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -558,16 +558,14 @@
     },
     {
       "command": "notebook:run-cell",
-      "linuxKeys": ["Ctrl Enter"],
-      "macKeys": ["Cmd Enter", "Ctrl Enter"],
-      "winKeys": ["Ctrl Enter"],
+      "keys": ["Ctrl Enter"],
+      "macKeys": ["Ctrl Enter", "Cmd Enter"],
       "selector": ".jp-Notebook:focus"
     },
     {
       "command": "notebook:run-cell",
-      "linuxKeys": ["Ctrl Enter"],
-      "macKeys": ["Cmd Enter", "Ctrl Enter"],
-      "winKeys": ["Ctrl Enter"],
+      "keys": ["Ctrl Enter"],
+      "macKeys": ["Ctrl Enter", "Cmd Enter"],
       "selector": ".jp-Notebook.jp-mod-editMode"
     },
     {
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -558,12 +558,12 @@
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Accel Enter"],
+      "keys": ["Accel Enter", "Ctrl Enter"],
       "selector": ".jp-Notebook:focus"
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Accel Enter"],
+      "keys": ["Accel Enter", "Ctrl Enter"],
       "selector": ".jp-Notebook.jp-mod-editMode"
     },
     {
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -558,22 +558,24 @@
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Ctrl Enter"],
+      "macKeys": ["Ctrl Enter"],
+      "keys": [],
       "selector": ".jp-Notebook:focus"
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Cmd Enter"],
-      "selector": ".jp-Notebook:focus"
-    },
-    {
-      "command": "notebook:run-cell",
-      "keys": ["Ctrl Enter"],
+      "macKeys": ["Ctrl Enter"],
+      "keys": [],
       "selector": ".jp-Notebook.jp-mod-editMode"
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Cmd Enter"],
+      "keys": ["Accel Enter"],
+      "selector": ".jp-Notebook:focus"
+    },
+    {
+      "command": "notebook:run-cell",
+      "keys": ["Accel Enter"],
       "selector": ".jp-Notebook.jp-mod-editMode"
     },
     {
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -559,13 +559,21 @@
     {
       "command": "notebook:run-cell",
       "keys": ["Ctrl Enter"],
-      "macKeys": ["Ctrl Enter", "Cmd Enter"],
+      "selector": ".jp-Notebook:focus"
+    },
+    {
+      "command": "notebook:run-cell",
+      "keys": ["Cmd Enter"],
       "selector": ".jp-Notebook:focus"
     },
     {
       "command": "notebook:run-cell",
       "keys": ["Ctrl Enter"],
-      "macKeys": ["Ctrl Enter", "Cmd Enter"],
+      "selector": ".jp-Notebook.jp-mod-editMode"
+    },
+    {
+      "command": "notebook:run-cell",
+      "keys": ["Cmd Enter"],
       "selector": ".jp-Notebook.jp-mod-editMode"
     },
     {
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -558,16 +558,6 @@
     },
     {
       "command": "notebook:run-cell",
-      "keys": ["Ctrl Enter"],
-      "selector": ".jp-Notebook:focus"
-    },
-    {
-      "command": "notebook:run-cell",
-      "keys": ["Ctrl Enter"],
-      "selector": ".jp-Notebook.jp-mod-editMode"
-    },
-    {
-      "command": "notebook:run-cell",
       "keys": ["Accel Enter"],
       "selector": ".jp-Notebook:focus"
     },
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -558,12 +558,16 @@
     },
     {
       "command": "notebook:run-cell",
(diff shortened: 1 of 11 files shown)


jtpio:
> Remove `Ctrl Enter` shortcut to use only `Accel Enter` to run cells (Accel will be set to Cmd or Ctrl depending on the OS) [documentation](https://jupyterlab.readthedocs.io/en/stable/extension/extension_points.html#keyboard-shortcuts)

The idea behind https://github.com/jupyterlab/jupyterlab/pull/11942 was to have *both* shortcuts, to match the behavior of the classic notebook on Mac (feature parity)

fcollonval:
> The idea behind https://github.com/jupyterlab/jupyterlab/pull/11942 was to have both shortcuts, to match the behavior of the classic notebook on Mac (feature parity)

Thanks for providing that info. I updated the PR to support both in 201b48f - do you have a Mac to test it... ?

jtpio:
No Mac around at the moment, but we can ping folks that have one so they can try on Binder.

echarles:
Launching binder now from my Mac (FR-BE keyboard). I am always confused with those namings, and keyboard layouts. What do I need to try?

fcollonval:
> Launching binder now from my Mac (FR-BE keyboard). I am always confused with those namings, and keyboard layouts. What do I need to try?

Thanks @echarles 
Could you try to run notebook cells using the shortcut `Ctrl + Enter` and `Cmd + Enter`. It should work on Mac OS with this PR.

echarles:
`Shift+Enter` runs the cell and goes to the next cell
However, `Ctrl+Enter` and `Cmd+Enter` has no effect.
Strangely `Option+Enter` runs the cell and goes to the next cell

Screenshot 2022-02-25 at 11 17 12

We can hop on a call to double check if you need.

fcollonval:
Indeed it breaks also the shortcut on Linux.

fcollonval:
@echarles the latest commit is working on Linux. Could you try it again on MacOS please?

echarles:
I have clicked on the binder button, build has run, and Ctrl+Enter and Cmd+Enter do not work.

I am now building on my local env to be sure and will report.

echarles:
@fcollonval on my local env macOs, latest commit does not work. Nor Ctrl+Enter nor Cmd+Enter run the cell. Opt+Enter does run the cell.

echarles:
@fcollonval If you give me a few hints, I can hack to make it work on my local env. It will be much faster to iterate. We can also have a call if it helps, the shortcut settings are not familiar to me, especially since the recent changes.

fcollonval:
It sounds like there is a bug in [Lumino](https://github.com/jupyterlab/lumino/blob/aae9de3a042b91bd54ba7f80b2ff7034a879fe4e/packages/commands/src/index.ts#L498) as setting two shortcuts like `Ctrl + Enter` and `Cmd + Enter` breaks the shortcut detection.

echarles:
> It sounds like there is a bug in [Lumino](https://github.com/jupyterlab/lumino/blob/aae9de3a042b91bd54ba7f80b2ff7034a879fe4e/packages/commands/src/index.ts#L498) as setting two shortcuts like Ctrl + Enter and Cmd + Enter breaks the shortcut detection.

Does that mean that a temporary workaround would be to only define `Ctrl+Enter` to run a cell (without moving to the next one?)?

Where those 2 shortcuts already defined before (I guess not)?

packages/notebook-extension/schema/tracker.json
```diff
@@ -559,21 +559,13 @@
...
-    },
-    {
-      "command": "notebook:run-cell",
-      "keys": ["Accel Enter"],
-      "selector": ".jp-Notebook:focus"
-    },
-    {
-      "command": "notebook:run-cell",
-      "keys": ["Accel Enter"],
+      "macKeys": ["Ctrl Enter", "Cmd Enter"],
```
jasongrout, line 568: This means that you need to press the chord `Ctrl Enter, Cmd Enter`, i.e., you need to press both key combinations in succession for the shortcut to trigger. To define two separate shortcuts, you need to make two completely separate entries.

jasongrout: I would suggest reverting this part of the PR - it was very intentional that there were two shortcuts, one for ctrl enter and one for Accel enter. If you wanted to clean things up, you could make the Ctrl-enter shortcut only a mackeys one, but I don't think it is necessary either.

fcollonval: Thanks @jasongrout I did not understand the list thingy properly. But then the shortcuts setting loader does not either :wink: as seen in https://github.com/jupyterlab/jupyterlab/issues/12103 defining multiple shortcuts for the same command is not possible.

I will look at correcting the loader.

packages/settingregistry/src/tokens.ts
```diff
@@ -592,7 +592,12 @@ export namespace ISettingRegistry {
     disabled?: boolean;
 
     /**
-     * The key combination of the shortcut.
+     * The key sequence of the shortcut.
+     *
+     * ### Notes
+     *
+     * If this is a list like `[Ctrl A, B]`, the user needs to press
```
jasongrout, line 599:
```suggestion
     * If this is a list like `['Ctrl A', 'B']`, the user needs to press
```

packages/notebook-extension/schema/tracker.json
fcollonval: This works only partially, we still have one warning (instead of two):

```
{ command: "notebook:run-cell", keys: (1) […], selector: ".jp-Notebook:focus" }
```

That comes from `Cmd Enter` being converted as `Enter` that conflict with another existing shortcuts. Actually it seems wrong to even accept that shortcut on non Mac platform as it can lead to unexpected behavior.

jasongrout: Can you put the `Cmd Enter` in the macKeys argument?

jasongrout: It looks like `Cmd` is just ignored as a modifier on non-mac platforms. I agree that this seems problematic - the shortcut should probably be rejected if it has Cmd on a non-mac:

https://github.com/jupyterlab/lumino/blob/aae9de3a042b91bd54ba7f80b2ff7034a879fe4e/packages/commands/src/index.ts#L1146-L1148

jasongrout: Or you can leave the Accel Enter shortcut, and make a new one that is `Ctrl Enter` in the macKeys, so it just applies for macs.

fcollonval: > Can you put the Cmd Enter in the macKeys argument?

I'm not sure to get that suggestion. `keys` cannot be [undefined](https://github.com/jupyterlab/jupyterlab/blob/aef2a162ece7f2e983bb118675669920c6bed6f9/packages/shortcuts-extension/schema/shortcuts.json#L40). And if I set the command with:

```
"keys": ["Ctrl Enter"],
"macKeys": ["Cmd Enter"]
```

It will override the "Ctrl Enter" shortcut as [lumino](https://github.com/jupyterlab/lumino/blob/aae9de3a042b91bd54ba7f80b2ff7034a879fe4e/packages/commands/src/index.ts#L1164) is taking `macKeys || keys`. But we want to have both "Ctrl Enter" and "Cmd Enter" on MacOS.

jasongrout: keys can be an empty list, though.

fcollonval:
Thanks for the snippet @jasongrout 

It does not work due to that [schema](https://github.com/jupyterlab/jupyterlab/blob/aef2a162ece7f2e983bb118675669920c6bed6f9/packages/settingregistry/src/plugin-schema.json#L209) being enforced for settings schema metadata.

```
Validating @jupyterlab/notebook-extension:tracker failed:
0 - schema @ #/definitions/shortcut/properties/keys/minItems, data @ ['jupyter.lab.shortcuts'][39].keys
{minItems} should NOT have fewer than 1 items
```

fcollonval:
@jasongrout would you mind having a second look at this one?

jasongrout:
Thanks. I pushed one more commit here that restores the Accel shortcut (both because Accel is probably good practice, and also because I think we want to change shortcuts as minimally as possible, so that people's overrides continue to work). This necessitated tweaking the schema that you pointed out to allow `keys` to be an empty list.

fcollonval:
Thanks @jasongrout 

Merging as CI is green and the warning are gone (see example app in Linux test CI).

fcollonval:

PR #12123, opened 2022-02-25
Set left sidebar width through mouse motion

### LINKED ISSUE #12023: Visual Regression Documentation CI check is failing
## Description

It looks like the main Visual Regression check now passes with the latest commit on master after restarting the CI:

https://github.com/jupyterlab/jupyterlab/actions/runs/1805191540

image

However the "Visual Regression Documentation" check seems to be failing:

image

We should double check whether this failure is relevant or not, since this seems related to the default side bar width.

## Reproduce

The latest commit on master: https://github.com/jupyterlab/jupyterlab/actions/runs/1805191540

fcollonval:
## References

Fixes #12023

## Code changes

Remove `waitForFrames` as this kind of trick should be avoided (and at the end proves not useful for us)

`waitForFrames` has not yet been published. So no.


fcollonval:

jtpio:
Thanks @fcollonval for looking into this :+1: 

We need to check whether the text selection is now happening because of the manual drag of the side panel handle:

image

image

Otherwise the other checks look good.

galata/test/documentation/utils.ts
```diff
@@ -56,3 +58,19 @@ export function positionMouse(position: { x: number; y: number }): string {
...
+): Promise<void> {
+  const splitHandle = await page.$('.lm-SplitPanel-handle');
+  const handleBBox = await splitHandle.boundingBox();
+
+  await page.mouse.move(
+    handleBBox.x + 0.5 * handleBBox.x,
+    handleBBox.y + 0.5 * handleBBox.height
+  );
+  await page.mouse.down();
+  await page.mouse.move(33 + width, handleBBox.y + 0.5 * handleBBox.height);
```
jtpio, line 74: Do we need to move along the `y` axis?

Wondering if this might be causing the selection mentioned in https://github.com/jupyterlab/jupyterlab/pull/12123#issuecomment-1054018426.

fcollonval: The position is the absolute one. So there is no y motion. But the selection makes me wonder if we really are catching the handle.

jtpio: Ah right. Then yes it's possible the handle is indeed not caught. Posting the Playwright captured video here for reference:

https://user-images.githubusercontent.com/591645/156006004-4e557e45-480c-4969-8ea2-fb1c42890b8f.mp4

galata/test/documentation/utils.ts
```diff
@@ -56,3 +58,19 @@ export function positionMouse(position: { x: number; y: number }): string {
...
+
+export async function setLeftSidebarWidth(
+  page: Page,
+  width = 251
+): Promise<void> {
+  const splitHandle = await page.$('.lm-SplitPanel-handle');
+  const handleBBox = await splitHandle.boundingBox();
+
+  await page.mouse.move(
+    handleBBox.x + 0.5 * handleBBox.x,
```
jtpio, line 70: Looks like we might need the `width` here? (not tested though)

```suggestion
    handleBBox.x + 0.5 * handleBBox.width,
```

jtpio:
https://github.com/jupyterlab/jupyterlab/pull/12123/commits/b11ea52da5a5b7e31a6c533afcb2e39321eb8508 seems to help with the text selection issue, the last run shows only 2 errors:

Looks like we also need to update the screenshot logic for the debugger sidebar:

image

fcollonval:
Sounds we are finally there - although they are still some flaky tests I think we have now a more robust way to deal with the sidebar width. Thanks a lot @jtpio

jtpio (approved):
Nice, many thanks @fcollonval!!

PR #12130, opened 2022-02-28
Fix typos in `@jupyterlab/console-extension` plugins

jtpio:
## References

Fixes the name of some plugins added in https://github.com/jupyterlab/jupyterlab/pull/11450

## Code changes

Rename plugins.

## User-facing changes

None

None since https://github.com/jupyterlab/jupyterlab/pull/11450 was a 4.0 change.


jasongrout:
Thanks!

PR #12141, opened 2022-03-02
Build UMD module for @jupyterlab/services

### LINKED ISSUE #12140: UMD module for `@jupyterlab/services` are not published since 3.1.6
## Description

A colleague noticed that the @jupyterlab/services package no longer includes a UMD module in the /dist directory when it is published. It seems that it stopped being provided between JupyterLab 3.1.6 and 3.1.8, perhaps as a consequence of jupyterlab/jupyterlab#10891 ? Has anyone else noticed this?

from @jasongrout on Gitter

fcollonval:
Fixes #12140

### DIFF
--- a/packages/services/package.json
+++ b/packages/services/package.json
@@ -35,6 +35,7 @@
   ],
   "scripts": {
     "build": "tsc -b",
+    "build:all": "npm run build && webpack",
     "build:test": "tsc --build tsconfig.test.json",
     "clean": "rimraf lib && rimraf tsconfig.tsbuildinfo",
     "docs": "typedoc src",


fcollonval:
I checked that we got back the dist folder

The strange thing is that the produced asset is not `dist/index.js` but `dist/dist/index.js`. I checked that it was also the case in `@jupyterlab/services@6.1.0`. So I presume for this minor release we should keep it like this.

> Maybe we should have something similar for the files.

Indeed it will be nice to check that there is at least a match for the globs defined.

jtpio (approved):
Thanks!

jasongrout:
Thanks! CC @tkrabel

PR #12156, opened 2022-03-03
Export KernelConnection

### LINKED ISSUE #12155: Expose KernelConnection publicly
### Affiliation Disclaimer

I work for Databricks, and my request is related to work we currently do there. 

### TLDR

We want to support ipywidgets on our platform while reusing as many open source components possible. Due to architectural differences (e.g. we run all third party FE code in iframes and we have more Databricks specific components between the Browser and the Python kernel), we cannot simply use the `KernelManager` as provided in the `examples/` folder. instead, we need to reuse the `KernelConnection` directly.

### Problem

* I would like the KernelConnection to be accessible from a public AMD module because I need to use this components inside Databricks.

### Proposed Solution

* Export `KernelConnection`

tkrabel-db:
## References

Fixes #12155

## Code changes

Export `KernelConnection` so that it's accessible when loading the @jupyterlab/services AMD bundle

## User-facing changes

none

none

### DIFF
--- a/packages/services/src/kernel/index.ts
+++ b/packages/services/src/kernel/index.ts
@@ -5,6 +5,7 @@
 import * as Kernel from './kernel';
 import * as KernelMessage from './messages';
 import * as KernelAPI from './restapi';
+import { KernelConnection } from './default';
 
 export * from './manager';
-export { Kernel, KernelMessage, KernelAPI };
+export { Kernel, KernelMessage, KernelAPI, KernelConnection };


tkrabel-db:
CC @jasongrout

fcollonval:
I can approve this PR - but as I don't have strong history it will be nice if some _older_ maintainer can approve it.

I agree with Jason that it is ok to add this to 3.3.x.

jtpio:
> ots of projects are using these underlying libraries at various levels to talk with kernels, Jupyter Server, etc., and there is quite a bit of logic in KernelConnection that would be great to be able to reuse in third-party packages

Yes. One example is the Voila Frontend using it here:

https://github.com/voila-dashboards/voila/blob/88565ebb07ad136b543ebdba96fc042d2232c8f2/packages/voila/src/kernel.ts#L14

With:

```ts
import { KernelConnection } from '@jupyterlab/services/lib/kernel/default';
```

jasongrout:
> Yes. One example is the Voila Frontend using it here:

I didn't realize Voila was already having to import KernelConnection from deep in the package. I think this indicates we should have exported this long ago.

jasongrout (approved):

fcollonval:

Issue #10933, opened 2021-08-27
[DISCUSS] Notebook actions while the notebook is not yet fully rendered by the virtual notebook


echarles:
This is a followup of https://github.com/jupyterlab/jupyterlab/issues/10840

I looks also that the reported case https://github.com/jupyterlab/jupyterlab/issues/10840#issuecomment-903130229 still gives failures

The initial idea of the virtual notebook was to have 2 cell "areas":

1. The first one at the top of the notebook which are rendered directly.
2. The second one after the first one, which are populated by placeholders, and then lazily rendered when the browser is idle.
We were using in the second area an index which has shown to be broken after #10793. This PR fixes that.

Still, there is the assumption that during the rendering of the second area, the user will not take actions to change the structure of the notebook. It is fine for him to run an existing cell, write in existing cell, save the notebook, restart the kernel... but I can see an issue if he inserts or deletes a cell, as the position of the cells to be rendered should be changed. We have 2 options to takes this behavior into account:

1. Disallow any structural change to the cells indices while the notebook is not fully rendered... in other words the "toRenderCell" map is empty
2. On insert/delete, update the indices of the "toRenderMap" according to the change, which is possible, but a bit cumbersome to write in code.

@krassowski made goof points on https://github.com/jupyterlab/jupyterlab/pull/10898#issuecomment-904416206 which I copy here after.

In my opinion if we will have a visual cue of loading, this could be an acceptable UX (barely, but acceptable). On the other hand, I am not convinced whether making good visual cues and prohibiting changes from triggering any susceptible code path will be easier than tracking the changes. I mean that a good UX would only prohibit very specific actions and only on a subset of cells: by the time we add the logic to handle all those actions we already covered the same places as we would if we instead implemented the index tracking/rebuilding (which is obviously more difficult in itself, but the extend places in code that need to be touched will not be so much different). Can you raise this tomorrow on the dev meeting (unfortunately I cannot attend it tomorrow)?

Mithil467:
#10998 has an example that still causes this issue on latest master.
<details>
<summary>ipynb file</summary>
<pre>
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "id": "e93a11d0-9c02-4550-812e-b88208da40e0",
   "metadata": {},
   "outputs": [],
   "source": [
    "print(\"a\")"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": null,
[...]

PR #12166, opened 2022-03-04
renderCellOnIdle improvements

rtrg:
Currently on page load, all placeholder cells' widget are being rendered immediately irrespective of whether user is interacting with UI or not.

This is because currently on every cell's widget (numberCellsToRenderDirectly) insert, there is a requestIdleCallback registered to render a placeholder cells' widget (which is enabled when observer is present) and its timeout is specified as 1 second. On page load, initial cells' widget (numberCellsToRenderDirectly) gets rendered and each of them will trigger a renderPlaceholderCells in a short span and given all of them have timeout: 1000 ms, they are going to render almost together and more over it is not using idle time left which means even if 100 cells are queued then irrespective of idle time left (which is generally <50ms during active rendering) it will try to render all the 100 cells which might hinder actual user interaction.
For each scheduled place holder cell's widget rendered it will register more callbacks which is making all the cells to render as soon as page loads making the numberCellsToRenderDirectly setting not so useful.

For large notebooks this causes user experience degradation (like tab switch being slow) as browser is busy rendering all the cells.

## References

https://github.com/jupyterlab/jupyterlab/issues/9757

## Code changes

To fix this I have tried rendering place holder cell's widget during [full idle frames](https://w3c.github.io/requestidlecallback/#idle-periods) (deadline.timeRemaining() will be 50) i.e., when there are no UI updates/user inputs queued. Increased timeout to avoid lot of cells getting idle callback at closer intervals and in case of timeout (meaning it couldn't find real idle time) instead of trying to render placeholder cell's widget, rescheduling to find another idle period.

These changes fixes sluggish UI for large notebooks.

Below is the benchmark report for [generated-1000cells-0.9ratio-2loc.ipynb](https://github.com/jupyterlab/benchmarks/blob/master/examples/from-script/generated-1000cells-0.9ratio-2loc.ipynb) notebook with 1K cells.

image

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -603,15 +603,28 @@ export class StaticNotebook extends Widget {
       this.onCellInserted(index, widget);
     }
 
+    this._scheduleCellRenderOnIdle();
+  }
+
+  private _scheduleCellRenderOnIdle() {
     if (this._observer && this.notebookConfig.renderCellOnIdle) {
       const renderPlaceholderCells = this._renderPlaceholderCells.bind(this);
       (window as any).requestIdleCallback(renderPlaceholderCells, {
-        timeout: 1000
+        timeout: 60000
       });
     }
   }
 
   private _renderPlaceholderCells(deadline: any) {
+    const timeRemaining = deadline.timeRemaining();
+
+    // In case this got triggered because of timeout or when there are screen updates (https://w3c.github.io/requestidlecallback/#idle-periods),
+    // avoiding the render and rescheduling the place holder cell rendering.
+    if (deadline.didTimeout || timeRemaining < 50) {
+      this._scheduleCellRenderOnIdle();
+      return;
+    }
+
     if (
       this._renderedCellsCount < this._cellsArray.length &&
       this._renderedCellsCount >=
@@ -623,6 +636,11 @@ export class StaticNotebook extends Widget {
   }
 
   private _renderPlaceholderCell(cell: Cell, index: number) {
+    // We don't have cancel mechanism for scheduled requestIdleCallback(renderPlaceholderCells),
+    // adding defensive check for layout in case tab is closed.
+    if (!this.layout) {
+      return;
+    }
     const pl = this.layout as PanelLayout;
     pl.removeWidgetAt(index);
     pl.insertWidget(index, cell);


fcollonval:
Thanks @rtrg this looks really promising.

Looking at the benchmark results, I guess this is certainly well working. As rendering the cells on idle is further delayed (as expected), you are probably not having the notebook fully loaded when carrying out the actions; hence the impressive improvements.

The main flaw with the current approach is the assumption that the user will not change the cells list (by deletion, motion or insertion) before the notebook is fully rendered. Would you mind trying to alleviate that assumption? Otherwise I'll be quite reluctant to merge this one as with the additional delay, the risk to break that assumption is greater.

echarles:
> The main flaw with the current approach is the assumption that the user will not change the cells list (by deletion, motion or insertion) before the notebook is fully rendered. Would you mind trying to alleviate that assumption? Otherwise I'll be quite reluctant to merge this one as with the additional delay, the risk to break that assumption is greater.

That risk exists and is highlighted in https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605. A mitigation would be to disallow any structural change to the cells indices while the notebook is not fully rendered... in other words the "toRenderCell" map is empty

rtrg:
Thanks @fcollonval and @echarles for reviewing!

> The main flaw with the current approach is the assumption that the user will not change the cells list (by deletion, motion or insertion) before the notebook is fully rendered. Would you mind trying to alleviate that assumption? Otherwise I'll be quite reluctant to merge this one as with the additional delay, the risk to break that assumption is greater.

As pointed out by @echarles,[ this assumption](https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605) is coming from virtual notebook feature itself and not newly introduced.

Today even without these changes virtual notebook seems to have [mentioned issues](https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605). Agree that with these changes risk of hitting these issues might increase as rendering might get delayed further but when users opt to `renderCellOnIdle`, expecting all the cells to be there on load (before notebook actions are triggered) seems not correct just for notebook actions to work correctly, especially when rendering all cells is causing huge performance degradation for large notebooks' common use cases like tab switch, scrolling, cell execution,....


> That risk exists and is highlighted in [#10933 (comment)](https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605). A mitigation would be to disallow any structural change to the cells indices while the notebook is not fully rendered... in other words the "toRenderCell" map is empty

Agree with this, IMO if people opt for virtual notebook we can restrict few actions like these as these use cases might be not so frequent.

These changes goes inline with `renderCellOnIdle` feature expectation and strictly improves user interaction experience for common use cases like tab switch, scroll, cell execution of large notebooks.

Overall I feel that irrespective of these changes, [notebook actions interfering with virtual notebook implementation](https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605) should be solved either by restricting to not allow them or to update the virtual notebook implementation. Once solved these changes should automatically get out of notebook actions concern.

Please let me know your opinion about merging these changes to improve user experience greatly for common use cases of large notebooks and for notebook actions concerns raised until [fix is added to virtual notebook feature](https://github.com/jupyterlab/jupyterlab/issues/10933#issue-980883605), if notebook actions are important then recommendation can be they should not enable virtual notebook instead of enabling virtual notebook and expecting all cells to be there.

echarles:
@rtg Looking at the changes of this PR, could we define the timeout as a setting so that each user can set up their own value based on their habits.

A follow-up of this PR would be to disallow cells changes (move, insert, delete, maybe a few other actions tbd).

rtrg:
> @RTG Looking at the changes of this PR, could we define the timeout as a setting so that each user can set up their own value based on their habits.

@echarles With these changes, `timeout` doesn't guarantee that cells will render after `timeout` expires as I am rescheduling to find another idle callback in case of timeout. My intention here is to render cells only when truly idle and timeout shouldn't force which effects user interactions. I have increased `timeout` just to give enough time for `idlecallbacks` to scatter instead of triggering at close intervals.

I can make this `timeout` as a setting but do you also want it to force render instead of rescheduling on timeout? If not then providing this as a setting is not useful. I am ok to change the timeout behavior back to force render as users can set it very high number in case of large notebooks.

echarles:
> I can make this timeout as a setting but do you also want it to force render instead of rescheduling on timeout? If not then providing this as a setting is not useful. I am ok to change the timeout behavior back to force render as users can set it very high number in case of large notebooks.

We have to be very cautious with the virtual rendering to avoid un-excepted side-effects. I can understand the advantage of rescheduling the rendering, but this will for sure need more battle-testing before merging. This leads maybe to add one-more setting to enable or disable that rescheduling. The default values for all these settings should be defensive. To avoid one-more setting, you can e.g. say you reschedule if the timeout is larger than 5s (or any defined and documented threshold). The defensive values would be the current 1s timeout (without rescheduling).

fcollonval:
I agree with @echarles that we should be defensive on those changes; that is the reason I'm reluctant on merging this as is even if it will bring improvement for displaying a notebook.

Regarding configuration, for me the one that needs customization is more the `50` time slot than the timeout. Requesting a full idle cycle, implies that at best one cell will be rendered per idle cycle otherwise it will be rescheduled (if I got the code correctly).
We could imagine that having that setting equals to `0` will fallback to the current behavior - otherwise it will use rescheduling.

Two comments on handling file closed before finishing rendering.
- You should not reschedule if the notebook panel `isDisposed()`.
- Would it be possible to use [cancelIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelIdleCallback) for browsers supporting it?

rtrg:
> We have to be very cautious with the virtual rendering to avoid un-excepted side-effects. I can understand the advantage of rescheduling the rendering, but this will for sure need more battle-testing before merging. This leads maybe to add one-more setting to enable or disable that rescheduling. The default values for all these settings should be defensive. To avoid one-more setting, you can e.g. say you reschedule if the timeout is larger than 5s (or any defined and documented threshold). The defensive values would be the current 1s timeout (without rescheduling).

In general agree on being defensive to not break existing use cases.


> Requesting a full idle cycle, implies that at best one cell will be rendered per idle cycle otherwise it will be rescheduled (if I got the code correctly).

I overlooked this, intention is to allow cells to render only during full idle cycle but not to restrict to only one cell per idle cycle.



> Regarding configuration, for me the one that needs customization is more the `50` time slot than the timeout.

This will be very useful, thanks for the suggestion! Instead of currentTimeRemaining < 50, we can use this setting and default setting can be the time taken for a single cell to render (which can vary with upgrades and this default setting can be updated as required).



> We could imagine that having that setting equals to `0` will fallback to the current behavior - otherwise it will use rescheduling.

yes, we can get rid of timeout as setting with this.



> Two comments on handling file closed before finishing rendering.
> 
> * You should not reschedule if the notebook panel `isDisposed()`.
> * Would it be possible to use [cancelIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelIdleCallback) for browsers supporting it?

These are doable.

PR #12179, opened 2022-03-08
Add pygments-extension

martinRenou:
Add a pygments extension that contains the CSS generated from https://github.com/jupyterlab/jupyterlab_pygments

This allows theme-aware syntax highlighting in IPython's inspection:

https://user-images.githubusercontent.com/21197331/157206327-a2c9454d-f084-4f98-bf0e-6f3f86a97fb6.mp4

## Code changes

New `pygments` and `pygments-extension` packages

## User-facing changes

Syntax highlighting in inspect requests HTML outputs

None


fcollonval (changes requested):
Thanks @martinRenou 

There is no reason to add two packages. You should have only  `@jupyterlab/pygments-extension`.

You should not make add that package to the `@jupyterlab/application-extension` - it is not related to it. Having it simply separated is fine.

You should remove the jest and babel config files as they are only needed if a package is defining some tests.

martinRenou:
Thanks for the review. I'll make those updates

martinRenou:
@fcollonval I did as you suggested, but the extension does not seem to load anymore (the CSS is not in the JupyterLab page).

This was happening before when I tried making a `pygments-extension` without a `pygments` one (and this is why I tried making two sub-packages as it's done for other packages).

martinRenou:
Actually it's not the `pygments` package removal that made it stop working, it's the removing from `@jupyterlab/application-extension` dependencies.

I added it back and now it works perfectly.

What do you think I should do @fcollonval? Should I keep it as a `@jupyterlab/application-extension` dependency or is there another way?

jtpio:
> Add a pygments extension that contains the CSS generated 

@martinRenou is this new package supposed to be loaded by default in JupyterLab as an extension?

jtpio:
> @martinRenou is this new package supposed to be loaded by default in JupyterLab as an extension?

If so maybe we need the `jupyterlab.extension` metadata in the `package.json`?

martinRenou:
> @martinRenou is this new package supposed to be loaded by default in JupyterLab as an extension?

Yes I guess that's the idea. It would be nice if RetroLab picked it up automatically as well (or any other remix that could benefit from it), if that is possible.

> If so maybe we need the jupyterlab.extension metadata in the package.json?

I'll try that, do you mean in the `package.json` of the `pygments-extension`?

jtpio:
> I'll try that, do you mean in the `package.json` of the `pygments-extension`?

Right, for example the other extensions have this:

https://github.com/jupyterlab/jupyterlab/blob/f64fed3ce3ef3ecb10725e3d94725d6d9fa0f48a/packages/htmlviewer-extension/package.json#L53-L56

martinRenou:
Ok that makes sense! So it's basically a core federated extension?

jtpio:
> Ok that makes sense! So it's basically a core federated extension?

Not federated, but a core extension yes if we want to include it by default in JupyterLab.

Also I'm wondering if this is the structure we would like to use. Right now it looks like the `pygments-extension` does not have any plugins? So if it's just CSS that should be included by default, maybe we can then just include the CSS in one of the existing extensions?

martinRenou:
> So if it's just CSS that should be included by default, maybe we can then just include the CSS in one of the existing extensions?

The problem with this approach is that we're not sure which extension to include it into.

The pygments CSS is useful for inspection output, which can either be in console outputs, notebook outputs or the inspection panel.

Also, we probably don't want to add that CSS to the main app, otherwise RetroLab would not get it.

jtpio:
> A nice side effect of publishing it as a separate package will be its availability to external project.

Similar to the `nbconvert-css` package: https://github.com/jupyterlab/jupyterlab/tree/master/packages/nbconvert-css

packages/pygments-extension/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "author": "Project Jupyter",
+  "sideEffects": [
+    "style/*.css",
+    "style/index.js"
+  ],
+  "style": "style/index.css",
+  "files": [
+    "style/**/*.{css,js,eot,gif,html,jpg,json,png,svg,woff2,ttf}"
+  ],
+  "scripts": {},
```
fcollonval, line 23: Let's add build and clean scripts

> We need to add `jupyterlab_pygments` and `pygments` as Python dev dependencies as they are needed by the generator scripts

```suggestion
  "scripts": {
    "build": "python generate_css.py",
    "build:all": "npm run build",
    "clean": "rimraf style/base.css",
  },
```

typedoc.js
```diff
@@ -63,6 +63,7 @@ const packages = [
   'outputarea',
   'pdf-extension',
   'property-inspector',
+  'pygments-extension',
```
fcollonval, line 66: Unneeded

```suggestion
```

fcollonval (changes requested):
Here are some simplifications

packages/pygments-extension/package.json
martinRenou: That will make jupyterlab dependent on jupyterlab_pygments and pygments at build time

martinRenou:
Thanks a lot for the review!

jtpio:
Thinking about this more: why can't we keep everything in the https://github.com/jupyterlab/jupyterlab_pygments repo and the `jupyterlab_pygments` package?

Having to generate the CSS with a Python script that depends on the external `jupyterlab_pygments` package sounds like we would split this functionality into two places? Which may be difficult to maintain over time.

If we want to enable this CSS in JupyterLab then the `jupyterlab_pygments` package can also distribute a prebuilt JupyterLab extension. Downstream distributions like RetroLab and other custom apps could then just depend on `jupyterlab_pygments`.

fcollonval:
If this is moved to `jupyterlab_pygments` (no counter argument), I would suggest to publish the package on npm to ease inclusion in other tools. And the question about shipping it by default or not to highlight code block in documentation remains.

martinRenou:
> And the question about shipping it by default or not to highlight code block in documentation remains.

And would you make `jupyterlab_pygments` a JupyterLab dependency?

fcollonval:
> And would you make jupyterlab_pygments a JupyterLab dependency?

I would depend on the npm package but not on the python one; as the thing that matters is the style sheet.

martinRenou:
Replacing this PR with https://github.com/jupyterlab/jupyterlab_pygments/pull/11

Issue #4185, opened 2018-03-15
Double-clicking the dark-grey empty space next to tabs should create a Launch screen


choldgraf:
I find myself wanting to "double-click" the grey areas next to a tab to create the "new tab" window. Does this sound like a reasonable pattern to support with JupyterLab? Alternatively, if there were a little floating button to the right of the right-most tab for creating a launch window, I think that'd work as well.

timkpaine:
Or just a small + sign like the add tab on most browsers

timkpaine:
This is a browsery UX thing, I like it

sccolbert:
Just an anecdote on why the "+" isn't *already* there, is because it's a bit of a tricky UX issue. Since tabs can be moved around and reordered, the question arises as to what to do with the "+" when the user starts moving tabs (since it would be in the way). Chrome simply hides the "+", and that's an option. But we also thought that we might want more buttons to the right in the future, perhaps minimize/maximize/settings/etc., and didn't really have a good answer about a "+" would interact with those.

So, we're not opposed to putting something there, but exactly *what* to put there is not yet fully clear.

PR #12195, opened 2022-03-10
Add a "New Tab" button that opens the launcher

### LINKED ISSUE #12186: More intuitive way to open the Launcher
*As discussed in the [9th March 2022 Jupyter Lab meeting](https://hackmd.io/Y7fBMQPSQ1C08SDGI-fwtg?view#Attendees)*

---

</br>

### Problem

The other day I couldn't figure out how to navigate to the **Launcher** in order to open the command line.

Since the **Launcher** is essentially an app catalog (e.g. tensorboard), it doesn't really make sense that it accessed from a button in the "file" explorer. We shouldn't do aware with this existing navigation pattern, but we should provide a more intuitive way to get to the **Launcher**.

image

</br>

---

</br>

### Proposed Solution

@ajbozarth  suggested having a 'new tab' button that opens the **Launcher** since people are used to this behavior in web browsers. [did this exist previously?]

image

</br>

---

</br>

### Additional Context

- @afshin had a cool idea of being able to pin **Launcher** apps as shortcuts on the left bar [sorry if I misinterpreted]
- The **Launcher** could be its own button on the left bar like a home button [but I like alex's idea better]

ajbozarth:
## References

Fixes #12186
Fixes https://github.com/jupyterlab/jupyterlab/issues/4185

## Code changes

Enables the `addButton` on tab bars in the main area that opens
the launcher widget.

This is a more intuitive way of opening the launcher than the current
methods of closing all tabs or the blue plus button in the file
browser sidebar widget

## User-facing changes

Screenshots for both themes and mouse-over

Screen Shot 2022-03-14 at 2 07 21 PM
Screen Shot 2022-03-14 at 2 07 36 PM
Screen Shot 2022-03-14 at 2 07 47 PM
Screen Shot 2022-03-14 at 2 07 58 PM

## To Do

- [ ] Finish styling
   - [x] Move add button next to tabs (instead of far right)
   - [x] Improve visual look of the add button
   - [ ] Add accessibility styling
- [x] Open launcher in tab bar that the add button is clicked
   - Currently the new launcher is opened in the tab bar of whichever tab is currently in focus, not the tab bar whose add button was clicked. (When multiple widgets are open side by side or above/below each other)
- [ ] ~Add tests~


packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -20,6 +20,7 @@ export class TabBarSvg<T> extends TabBar<T> {
   constructor(options: TabBar.IOptions<T> = {}) {
     options.renderer = options.renderer || TabBarSvg.defaultRenderer;
     super(options);
+    this.addButtonNode.innerHTML = addIcon.svgstr;
```
fcollonval, line 23: You should prefer the `.element` method:

```ts
addIcon.element({ container: this.addButtonNode });
```

> see [documentation](https://jupyterlab.readthedocs.io/en/stable/extension/ui_components.html#how-to-render-an-icon-into-a-dom-node)

jtpio:
There also seems to be an issue slightly related to this, with the following comment: https://github.com/jupyterlab/jupyterlab/issues/4185#issuecomment-373472577

> Just an anecdote on why the "+" isn't already there, is because it's a bit of a tricky UX issue. Since tabs can be moved around and reordered, the question arises as to what to do with the "+" when the user starts moving tabs (since it would be in the way). Chrome simply hides the "+", and that's an option. But we also thought that we might want more buttons to the right in the future, perhaps minimize/maximize/settings/etc., and didn't really have a good answer about how a "+" would interact with those.

packages/ui-components/src/icon/widgets/tabbarsvg.ts
ajbozarth: thanks, I was pretty sure there was a better way

ajbozarth:
Thanks for that reference @jtpio the addButton itself was only added to Lumino last year https://github.com/jupyterlab/lumino/pull/108 thus making this a pretty easy update on our side.

jtpio:
> Thanks for that reference @jtpio the addButton itself was only added to Lumino last year [jupyterlab/lumino#108](https://github.com/jupyterlab/lumino/pull/108) thus making this a pretty easy update on our side.

Ah nice! This will fix https://github.com/jupyterlab/jupyterlab/issues/4185 then :+1:

ajbozarth:
I've pushed an update fixing the issue where the launcher isn't opened in the same tab bar as it's clicked. I also did some work on the css styling around the add button to make it match the tab look and feel.

I'm unsure if this change necessitates any tests, there are no tests related to the tab bar or dock panel in ui-components currently and no tests in the launcher packages at all.

@isabela-pf if you could double check if the accessibility of this feature is solid (or ping someone pertinent)? I'm pretty sure it's ok, but I'm not confident about it.

Lastly a design issue: as it is implemented in lumino the add button is locked to the right end of the tab bar. This is due to the tabs actually being list items in a `ul` element, which is a sibling element to the add element. As such there's no way to get the add button to sit/float next to the tabs without changes to the feature in lumino, which I'm unsure is worth the hassle.

ajbozarth:
Pending any followup to my points above I've moved this out of WIP and updated the screenshots in the description

fcollonval:
To move the `+` button next to the tabs, you must remove the `flex: 1 1 auto` on tab content:

```
.p-TabBar-content, .lm-TabBar-content {
	margin: 0;
	padding: 0;
	display: flex;
	/* flex: 1 1 auto; */
	list-style-type: none;
}
```

If it is possible to hide the button when there is a tab motion, that will look better indeed.

ajbozarth:
> If it is possible to hide the button when there is a tab motion, that will look better indeed.

@fcollonval I looked into this and afaik right now lumino does not support this. We would need to add some signals for drag start/stop, which we could do, but I don't think we actually need this feature. I checked how FireFox and Chrome handle the add button when dragging tabs and they don't hide it, in fact they behave very similar to how this PR currently behaves.

fcollonval:
please update snapshots

> All ui tests failure are expected - updating the snapshots with the new button

packages/application/src/shell.ts
```diff
@@ -689,6 +690,27 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
...
+  /*
+   * Whether the add buttons for each main area tab bar are enabled.
+   */
+  get addButtonEnabled(): boolean {
+    return this._dockPanel.addButtonEnabled;
+  }
+
+  /*
+   * Set whether the add buttons for each main area tab bar are enabled.
+   */
```
fcollonval, line 702: The practice is to document only once for getter/setter pair.

```suggestion
```

packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -17,9 +18,16 @@ export class TabBarSvg<T> extends TabBar<T> {
...
-  constructor(options: TabBar.IOptions<T> = {}) {
+  constructor(options: TabBarSvg.IOptions = {}) {
     options.renderer = options.renderer || TabBarSvg.defaultRenderer;
     super(options);
+    const trans = ((options && options.translator) || nullTranslator).load(
+      'jupyterlab'
+    );
+    addIcon.element({
+      container: this.addButtonNode,
+      title: trans.__('New Launcher')
```
fcollonval, line 29: On Firefox the working is `Open a new tab`. I think we should also be more explicit here.

```suggestion
      title: trans.__('Open a New Launcher')
```

I don't know if it is better to say _a new launcher_ or _a new tab_. The latter is probably more common and easily to be understood by the users.

packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -54,6 +62,13 @@ export namespace TabBarSvg {
   }
 
   export const defaultRenderer = new Renderer();
+
+  export interface IOptions extends TabBar.IOptions<any> {
```
fcollonval, line 66: Let's keep the templating

```suggestion
  export interface IOptions<T> extends TabBar.IOptions<T> {
```

fcollonval (changes requested):
Thanks @ajbozarth 

I have three suggestions.

packages/ui-components/src/icon/widgets/tabbarsvg.ts
ajbozarth: Safari and Chrome only have `New Tab` so I personally think simpler is better. I'm ok switching to tab instead of launcher, but one of the reasons I used this exact text is it's the tool tip for the current button in the file browser and should already be translated.

packages/ui-components/src/icon/widgets/tabbarsvg.ts
ajbozarth: thanks it was giving me errors, so I removed it, adding it to `IOptions` fixes that

packages/application/src/shell.ts
ajbozarth: done

fcollonval:
@ajbozarth what about moving it to the shell and add a settings with the ordered list of commands to trigger when the button is hit (the default being the filebrowser and the launcher one)?

isabela-pf:
Thanks for summoning me for review! This is an interesting feature.

Based on my current knowledge, this looks good to go. Thanks for tackling a long standing issue, @ajbozarth!

Things I looked for
- [x] How does it fit when the space is filled with tabs? It stays visible, always at the end of the list.
- [x] What happens when transitioning to `simple mode`.  It disappears along with the tabs. Nice!
- [x] Does it have an appropriate tool tip? Yes: "New Launcher"
- [x] Does it have high enough contrast? Yes. It meets with the WCAG 2 color measurements and the drafted WCAG 3 APCA.
- [x] What kinds of labels does it have? title: New Launcher
- [x] Can I get to it with my keyboard only? No, but I also think this is an issue larger than this PR. I can't easily navigate tabs at all right now.
- [x] Where does focus go when you open a new launcher? It does switch to the new launcher. That's good!
- [x] Is there any hard coded styling? Nope. Fantastic!

In the [JupyterLab team meeting (notes linked)](https://github.com/jupyterlab/team-compass/issues/135#issuecomment-1069381431), Alex mentioned that this shares behavior with the New Launcher button in the file browser. That sounds like a wise implementation to me for a number of reasons, but especially because it means the behavior will be consistent.

Long term (non-blocking) feedback, this makes me wonder if there's some unneeded redundancy in having the same button so prominent in both places. I haven't had much time to think about this, so I could probably convinced either direction.

ajbozarth:
> what about moving it to the shell and add a settings with the ordered list of commands to trigger when the button is hit (the default being the filebrowser and the launcher one)? 

I originally loved this idea, but I realized it has a snag, we need to derive the `ref` to pass to the launcher commands, so we couldn't "just" trigger a command based on a list in settings, since those commands might need args derived from the signal (which give access to the dock panel and tabor that triggered the signal). So after factoring that in the result would be to have user wishing to include the button just enable it and connect to the signal, which is what we do here. 

This does mean that any implementation of launcher would need to "copy" this chunk of code to enable the add button, but they also need to copy the command in the index file too so I don't think this is a large issue if we document it properly in the change log.

ajbozarth:
I've pushed the update to move the css into ui-components where it belongs.

With this I've finished my planned updates to the PR pending further review.

fcollonval (approved):
Thanks for looking at the setting scenario @ajbozarth 

This looks good to me - CI failures are not related.

ajbozarth:
Pending any further feedback I'll merge this after the dev meeting tomorrow

fcollonval:

nscozzaro:
Since this design is essentially a browser-within-a-browser, it makes me wonder if it would make sense to offer a version of jupyterlab that opens as a desktop application.

jtpio:
> Since this design is essentially a browser-within-a-browser, it makes me wonder if it would make sense to offer a version of jupyterlab that opens as a desktop application.

@nscozzaro maybe you will want to check out the `jupyterlab-desktop` project: https://github.com/jupyterlab/jupyterlab-desktop

PR #12204, opened 2022-03-12
Add argument line and column to codemirror go to line command

fcollonval:
## References

Trying to add action to open a file and go to a specific line from an external extension is not easy as calling the go to line command.

Ref https://github.com/jupyterlab-contrib/search-replace/issues/22

## Code changes

Support optional `line` and `column` arguments to the command `codemirror:go-to-line`

## User-facing changes

N/A

N/A


bollwyvl:
How about a column arg, too? then we can stop doing it over on `jupyterlab-lsp` :laughing:

bollwyvl:
Cool stuff. For reference, here's all the stuff we have for [editors](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/0cff6b9cd7dab4d44f877cf7bacc91bf72c6ec4e/packages/code-jumpers/src/jumpers/fileeditor.ts#L34)... of course, it's a lot more _interesting_ for notebooks...

ajbozarth (approved):
code and test code lgtm, this looks like a solid add. I didn't test it locally though since there's no use-case in core

fcollonval:

PR #12208, opened 2022-03-14
Fix sdist editable install and add tests

blink1073:
## References
In the `maintainer-tools` [downstream test action](https://github.com/jupyterlab/maintainer-tools#test-downstream-libraries) we install in editable mode from an sdist by default.  In JupyterLab this was breaking because `scripts/ensure-buildutils.js` was not in the sdist, but really we shouldn't be running node at all.

## Code changes
Cleans up installing from sdist and adds a test for the sdist, as well as a test with the minimum supported dependencies using the new action from `maintainer-tools`.

## User-facing changes
Can now install in editable mode from an sdist.

None

### DIFF
--- a/.github/workflows/linuxtests.yml
+++ b/.github/workflows/linuxtests.yml
@@ -77,7 +77,7 @@ jobs:
           name: ${{ matrix.group }} ${{ github.run_number }}
           path: ./build/${{ matrix.group }}_output
 
-  test_miniumum_verisons:
+  test_minimum_versions:
     name: Test Minimum Versions
     timeout-minutes: 30
     runs-on: ubuntu-latest
--- a/.github/workflows/linuxtests.yml
+++ b/.github/workflows/linuxtests.yml
@@ -87,6 +87,8 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
         with:
           python_version: "3.7"
+      - name: Update Setuptools
+        run: python -m pip install -U setuptools
       - name: Install miniumum versions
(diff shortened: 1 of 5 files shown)


.github/workflows/linuxtests.yml
```diff
@@ -76,3 +76,65 @@ jobs:
         with:
           name: ${{ matrix.group }} ${{ github.run_number }}
           path: ./build/${{ matrix.group }}_output
+
+  test_miniumum_verisons:
```
jtpio, line 80: Minor typo spotted:

```suggestion
  test_minimum_versions:
```

setup.cfg
```diff
@@ -35,10 +35,10 @@ install_requires =
     packaging
     tornado>=6.1.0
     jupyter_core
-    jupyterlab_server~=2.10
-    jupyter_server~=1.4
+    jupyterlab_server>=2.10
+    jupyter_server>=1.8
```
fcollonval, line 39: Do we want to allow future major versions to be installed with older JupyterLab?

blink1073: This comes from the advice in a [blog post](https://iscinumpy.dev/post/bound-version-constraints/) about dependency handling in Python, cf https://github.com/jupyter-server/jupyter_server/issues/699.

blink1073: I take your point that JupyterLab is an application, and tightly coupled to its server, I put the version constraints back in for those.

fcollonval: Thanks @blink1073

fcollonval (approved):
LGTM

PR #12218, opened 2022-03-15
Fix state restoration in the notebook extension

jtpio:
## References

As reported in https://github.com/jupyterlab/retrolab/issues/358, where some commands are not available in downstream applications like RetroLab since they don't include a `ILayoutRestorer`.

## Code changes

Fix the guard checking if the `restorer` is available, so it only surrounds the relevant logic.

## User-facing changes

Should fix some behavior in downstream applications reusing `@jupyterlab/notebook-extension:tracker` as is.

None


fcollonval (approved):
LGTM @jtpio

jtpio:

PR #12219, opened 2022-03-15
Update completer part 3

trungleduc:
## References

This PR allows completer providers to customize the completer widget model and fixes an issue of activating completer before all providers are registered.

## Code changes

- Add a new signal (`providersActivated`), that is emitted when all providers are registered.
- Add an optional `modelFactory` method to the provider interface.

## User-facing changes

N/A

N/A

### DIFF
--- a/packages/completer/src/manager.ts
+++ b/packages/completer/src/manager.ts
@@ -25,13 +25,13 @@ export class CompletionProviderManager implements ICompletionProviderManager {
   constructor() {
     this._providers = new Map();
     this._panelHandlers = new Map();
-    this._providersActivated = new Signal<ICompletionProviderManager, void>(
+    this._activeProvidersChanged = new Signal<ICompletionProviderManager, void>(
       this
     );
   }
 
-  get providersActivated(): ISignal<ICompletionProviderManager, void> {
-    return this._providersActivated;
+  get activeProvidersChanged(): ISignal<ICompletionProviderManager, void> {
+    return this._activeProvidersChanged;
   }
 
   /**
@@ -102,7 +102,7 @@ export class CompletionProviderManager implements ICompletionProviderManager {
       this._activeProviders.add(KERNEL_PROVIDER_ID);
       this._activeProviders.add(CONTEXT_PROVIDER_ID);
     }
-    this._providersActivated.emit();
+    this._activeProvidersChanged.emit();
   }
 
   /**
@@ -216,6 +216,7 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     }
 
     const completer = new Completer({ model, renderer });
+    completer.showDocsPanel = this._showDoc;
     completer.hide();
     Widget.attach(completer, document.body);
     const connectorProxy = await this.generateConnectorProxy(completerContext);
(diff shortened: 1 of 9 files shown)


trungleduc:
Restart CI

packages/completer/src/manager.ts
```diff
@@ -24,6 +25,13 @@ export class CompletionProviderManager implements ICompletionProviderManager {
   constructor() {
     this._providers = new Map();
     this._panelHandlers = new Map();
+    this._providersActivated = new Signal<ICompletionProviderManager, void>(
+      this
+    );
+  }
+
+  get providersActivated(): ISignal<ICompletionProviderManager, void> {
```
fcollonval, line 33: Design question, are you interested by listening to the changes in the active providers or in the readiness of the manager at load time?

If this is the second, I would use a `ready` attribute:

```ts
get ready(): Promise<void>
```

If not, the signal is good, I would use anoterh name like `activeProvidersChanged` to better reflect when the signal is emitted (as it will be emitted on settings update too).

> Look for example at `jupyterlab/jupyterlab/packages/apputils/src/sessioncontext.tsx`

fcollonval (changes requested):
I got a question about the intent.

packages/completer/src/manager.ts
```diff
@@ -241,4 +259,10 @@ export class CompletionProviderManager implements ICompletionProviderManager {
    * Flag to enable/disable continuous hinting.
    */
   private _autoCompletion: boolean;
+
+  /**
+   * Signal emitted when active providers list is changed.
+   *
+   */
```
fcollonval, line 266: Let's place the doc on the public getter

```suggestion
```

packages/completer/src/manager.ts
```diff
@@ -24,6 +25,13 @@ export class CompletionProviderManager implements ICompletionProviderManager {
   constructor() {
     this._providers = new Map();
     this._panelHandlers = new Map();
+    this._activeProvidersChanged = new Signal<ICompletionProviderManager, void>(
+      this
+    );
+  }
+
+  get activeProvidersChanged(): ISignal<ICompletionProviderManager, void> {
```
fcollonval, line 33:
```suggestion
  /**
   * Signal emitted when active providers list is changed.
   */
  get activeProvidersChanged(): ISignal<ICompletionProviderManager, void> {
```

packages/completer/src/tokens.ts
```diff
@@ -61,9 +67,12 @@ export interface ICompletionProvider<
   ): Promise<CompletionHandler.ICompletionItemsReply<T>>;
 
   /**
-   * Renderer for provider's completions (optional).
+   * This method is called to customize the model of a completer widget.
+   * If it is not provided, the default model will be used.
+   *
+   * @param context - additional information about context of completion request
```
fcollonval, line 73:
```suggestion
   * @param context - additional information about context of completion request
   * @returns The completer model
```

packages/completer/src/tokens.ts
```diff
@@ -130,6 +139,12 @@ export interface ICompletionProviderManager {
    * @param newCompleterContext - The completion context.
    */
   updateCompleter(newCompleterContext: ICompletionContext): Promise<void>;
+
+  /**
+   * Signal emitted when active providers list is changed.
+   *
```
fcollonval, line 145:
```suggestion
```

fcollonval (changes requested):
Minor suggestions on the doc and it is good to go for me.

trungleduc:
Thanks @fcollonval for the review!

fcollonval (approved):
Thanks @trungleduc

PR #12224, opened 2022-03-16
Backport PR #12208 on branch 3.3.x (Fix sdist editable install and add tests)

blink1073:


setup.cfg
```diff
@@ -36,12 +36,12 @@ install_requires =
     tornado>=6.1.0
     jupyter_core
     jupyterlab_server~=2.10
-    jupyter_server~=1.4
+    jupyter_server~=1.8
```
fcollonval, line 39: Let's not change the dependency for a patch version

```suggestion
    jupyter_server~=1.4
```

setup.cfg
```diff
@@ -36,12 +36,12 @@ install_requires =
...
     nbclassic~=0.2
-    jinja2>=2.1
+    jinja2>=3.0.3
```
fcollonval, line 41:
```suggestion
    jinja2>=2.1
```

blink1073: We'll have to pin `markupsafe` then I believe

blink1073:
Holy moly, we can now just re-run failed jobs only!

blink1073:
jinja2 relies on an old API of setuptools, I think we need to just drop the min versions test in 3.3.x.   What do you think @fcollonval?

fcollonval:
> jinja2 relies on an old API of setuptools, I think we need to just drop the min versions test in 3.3.x. What do you think @fcollonval?

:+1: @blink1073

PR #12232, opened 2022-03-18
Make `IStatusBar` optional in the plugin providing `IPositionModel`

jtpio:
## References

Noticed when updating to the latest lab packages for Notebook v7: https://github.com/jupyter/notebook/pull/6314

## Code changes

The definition of `IPositionModel` doesn't indicate anything related to the status bar:

https://github.com/jupyterlab/jupyterlab/blob/d33de1537004414bd2242661a4797d53d2e699cd/packages/codeeditor/src/tokens.ts#L32-L57

But the plugin definition of the plugin providing `IPositionModel` still requires `IStatusBar`.

It looks like these two should be independent. Or that `IStatusBar` should at least be `optional`, if some other plugins want to require `IPositionModel`.

## User-facing changes

None for lab users.

None


jtpio:
The alternative would be to provide the status bar item in a different plugin.

jtpio:
Visual Regression Documentation failure is tracked in https://github.com/jupyterlab/jupyterlab/issues/12233

Issue #12233, opened 2022-03-18
Visual Regression Documentation check is failing


jtpio:
## Description

The [Visual Regression Documentation](https://github.com/jupyterlab/jupyterlab/runs/5597486569?check_suite_focus=true) check seems to be failing recently.

## Reproduce

Example run: https://github.com/jupyterlab/jupyterlab/actions/runs/2003019798

![image](https://user-images.githubusercontent.com/591645/158993352-d3b51326-d46f-4ca0-a413-4627b7939189.png)

## Expected behavior

Checks should pass.

This might require an update to the tests.

## Context

Seen on the latest commits on `master` and also in new PRs.

jtpio:
Investigating this locally with JupyterLab 3.3.2, and it looks like the `jupyterlab-geojson` also doesn't show up in the extension manager (but the widget manager extension does):

```
$ jupyter labextension list
JupyterLab v3.3.2
PREFIX/share/jupyter/labextensions
        @jupyter-widgets/jupyterlab-manager v3.1.0 enabled OK (python, jupyterlab_widgets)
        @jupyterlab/geojson-extension v3.2.0 enabled OK (python, jupyterlab-geojson)
```

![image](https://user-images.githubusercontent.com/591645/159027743-e0b738f7-5c22-493c-8e15-9a3a8c068507.png)

jtpio:
So it looks like the extension manager starts with a `searchResult` with 250 entries and filters the extensions here:

https://github.com/jupyterlab/jupyterlab/blob/d33de1537004414bd2242661a4797d53d2e699cd/packages/extensionmanager/src/widget.tsx#L835-L839

The filter seems to only be keeping the widgets extension:

![image](https://user-images.githubusercontent.com/591645/159033336-8b02ca1f-98c1-4705-bf38-eda74933e4cb.png)

So this random change could be explained by the request to npm returning a different set of extensions not including `jupyterlab-geojson`.

jtpio:
Which means this is most likely an issue with the extension manager not displaying all installed extensions.

Just installed another third-party (prebuilt) extension in the same environment for testing and this one seems to be showing up:

![image](https://user-images.githubusercontent.com/591645/159034036-b10a73d1-010e-4602-bd49-899540f9385c.png)

PR #7423, opened 2019-10-27
Search installed extensions

jtpio:
## References

Is feels more natural to be able to use the same search bar to search for installed extensions. Especially when a lot of third-party extensions are installed.

This makes it easier to disable an installed extension for instance.

## Code changes

Use the `searchResult` array to also search for installed extensions.

### Before

no-search-installed-extensions

### After

search-installed-extensions

### Note

If there is no search result (for example no internet connection), all the installed extensions will be shown.

## Backwards-incompatible changes

None

PR #12234, opened 2022-03-18
Fix installed extensions in the extension manager

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12233

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/7423

## Code changes

Fixes the logic to list installed extensions in the extension manager.

## User-facing changes

With the following prebuilt extensions installed and JupyterLab dev mode started with `--extensions-in-dev-mode`:

```bash
$ jupyter labextension list
JupyterLab v4.0.0a22
PREFIX/share/jupyter/labextensions
        @jupyter-widgets/jupyterlab-manager v3.1.0 enabled  X (python, jupyterlab_widgets)
        @jupyterlab/geojson-extension v3.2.0 enabled  X (python, jupyterlab-geojson)
```

#### Before

image

#### After

image

### DIFF
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -832,7 +832,7 @@ risks or contain malicious code that runs on your machine.`)}
             listMode={model.listMode}
             viewType={'installed'}
             entries={
-              model.searchResult.length > 0
+              model.query && model.searchResult.length > 0
                 ? model.installed.filter(
                     entry => model.searchResult.indexOf(entry) > -1
                   )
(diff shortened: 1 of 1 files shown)


jtpio:
Looks like this fixes the check :tada: 

image

jtpio:
Added to `3.3.x` so it can be backported, since it's likely to be failing on the `3.3.x` branch too.

fcollonval:
I'll propose another approach because as you mentioned in the issues the search is paginated. So we should not rely on the search results to filter the installed packages. I opened #12249 as alternative that uses the query term to filter the installed list.

jtpio:
OK sounds good, closing in favor of https://github.com/jupyterlab/jupyterlab/pull/12249

PR #12238, opened 2022-03-19
Add aria progressbar role and data-status for testing in extensions

### LINKED ISSUE #12174: No way to easily detect busy state of the notebook from DOM for testing
## Description

The progress indicator introduced in #10730 removed the way to detect if the kernel is busy from DOM. Previously it was easy to test for scenarios where kernel is busy with a simple CSS selector like:

```robot
${KERNEL_BUSY_INDICATOR}    css:.jp-NotebookPanel-toolbar div[title="Kernel Busy"]
```

However the new progress circle hides the title if active:

https://github.com/jupyterlab/jupyterlab/blob/ffc85e41cd4852b84786f476d63024c92d8df528/packages/notebook/src/executionindicator.tsx#L79

This is because we now have a custom tooltip which enables more customised behaviour as compared to a simple title.

This however:
- makes testing slightly trickier
- might have short-term accessibility implications: the custom-built hover tooltip may not be as accessible as the `title` attribute

## Reproduce

1. `from time import sleep; sleep(20)`
2. See no title without hover
3. See no easy way to detect state (requires parsing text)

## Expected behavior

- I propose that we add (pseudocode) `aria-label={showProgress ? kernelStatus : null}` when  which would both improve accessibility and allow for easier testing
- we should probably replace `''` with `null` in the `title` so that the attribute is completely removed rather than hanging as empty

## Context

- needed for https://github.com/jupyter-lsp/jupyterlab-lsp/pull/770
- JupyterLab version: 3.3.0

krassowski:
## References

Fixes #12174

## Code changes

- exposes kernel status in DOM as a data attribute for easier downstream testing
- adds aria attributes for progress bar

## User-facing changes

Screen reader users may notice the progress updates being read out depending on the screen reader and it settings.

None

### DIFF
--- a/packages/notebook/src/executionindicator.tsx
+++ b/packages/notebook/src/executionindicator.tsx
@@ -77,6 +77,7 @@ export function ExecutionIndicatorComponent(
     <div
       className={'jp-Notebook-ExecutionIndicator'}
       title={showProgress ? '' : titleFactory(kernelStatuses[status])}
+      data-status={status}
     >
       {circle}
       <div
--- a/packages/statusbar/src/components/progressBar.tsx
+++ b/packages/statusbar/src/components/progressBar.tsx
@@ -32,10 +32,16 @@ export namespace ProgressBar {
  * A functional tsx component for a progress bar.
  */
 export function ProgressBar(props: ProgressBar.IProps): JSX.Element {
-  const { width, ...rest } = props;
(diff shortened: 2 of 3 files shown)


jtpio (approved):
Thanks!

fcollonval:

PR #12240, opened 2022-03-20
Fix settings with `null` default not getting marked as modified

### LINKED ISSUE #12092: Settings UI sections not marked as modified when default is `null`
## Description

The sections are not properly marked as modified if the default is null.

## Reproduce

1. Go to "Settings"
2. Click on Markdown Viewer
3. Edit font family
4. See it does not jump to "Modified" section

Markdown-viewer

## Expected behavior

Edits in fields with default "null" should be recognised.

## Context

- JupyterLab version: 3.3

krassowski:
## References

Fixes #12092.

## Code changes

Remove lines which were ignoring `null` so that changes in setting if the user-set value, or the default values is `null` are recognised as modifications.

## User-facing changes

Settings UI properly marks sections with modified settings when only fields with `null` where modified (such as Markdown Viewer → Line Width).

| Before (`Restore to Defaults` not shown, even though settings were modified) | After (shown correctly) |
| --- | --- |
| Screenshot from 2022-03-20 02-05-55 | Screenshot from 2022-03-20 02-06-23 |

None.

### DIFF
--- a/packages/settingregistry/src/settingregistry.ts
+++ b/packages/settingregistry/src/settingregistry.ts
@@ -728,9 +728,7 @@ export class Settings implements ISettingRegistry.ISettings {
       const defaultValue = this.default(key);
       if (
         value === undefined ||
-        value === null ||
         defaultValue === undefined ||
-        defaultValue === null ||
         JSONExt.deepEqual(value, JSONExt.emptyObject) ||
         JSONExt.deepEqual(value, JSONExt.emptyArray)
       ) {


krassowski:
It is likely that the comparison to `undefined` could (/should?) be removed as well. @marthacryan, as this was added in #11079 do you happen to remember why these conditions were added?

marthacryan (approved):
Sorry about the delay on this review - I refactored this function a few times because I was having issues with arrays of complex types so I'm not completely sure why I did this. I think I was getting an error when `value` or `defaultValue` was undefined and I was using it as an input for `JSONExt.deepEqual`? I might have just extended that logic to null values as well.

Issue #11716, opened 2021-12-17
Restart Kernel dialog is missing context


krassowski:
### Problem

https://github.com/jupyterlab/jupyterlab/issues/11547 proposes to remove the notebook/file name from the status bar.

In https://github.com/jupyterlab/jupyterlab/issues/11547#issuecomment-979081182 and https://github.com/jupyterlab/jupyterlab/issues/11547#issuecomment-979362622 I argued that the notebook name needs to stay because no sufficient context is present for some of the operations. On of those is restarting kernel. Currently even when many notebooks are open side-by-side, the restart (and re-run) function shows a modal which does not say which kernel will get restarted:

![Screenshot from 2021-12-17 16-32-54](https://user-images.githubusercontent.com/5832902/146577761-98438f0c-8248-4293-9f22-f419c5520758.png)

Restarting a wrong kernel can lead to loss of work or hours of computations. Because the dialog can overlay with the tab bar it is sometimes impossible to determine the currently active tab (notebook/console) without looking at the statusbar.

### Proposed Solution

The dialog modal should include the name of the kernel and of the notebook/console to allow user to double-check what they are restarting.

### Additional context

While adding the context to the modal is IMO a prerequisite for removing the notebook name from statusbar (or at least hiding it by default or if it does not fit), it would be a useful UX improvement even if we decide to keep it in the status bar.

jweill-aws:
I like this idea — the current dialog does not have enough information to indicate what the red "Restart" button will do.

echarles:
What if the kernel is binded (via a sessions) to a console?

What if the kernel is binded via multiple sessions to multiple notebooks and console ?

I do that frequently.

krassowski:
Very good point. In that case I suggest we list all notebooks and consoles it is bound to. Does it sound good to you?

echarles:
> Does it sound good to you?
 
notebook, consoles... and files.. sure.

krassowski:
> notebook, consoles... and files.. sure.

Happy to list everything :) but do files have kernels? I may be missing something, but the common use case of creating a console for editor binds an existing kernel to a console for file, not to directly to a file:

![files-dont-have-kernels](https://user-images.githubusercontent.com/5832902/146674011-d78617cb-0053-4c72-b148-a5fc9eea4a8e.gif)

echarles:
[...]

PR #12241, opened 2022-03-20
Show kernel name on restart to avoid data loss on misclick

krassowski:
## References

A simple solution to #11716.

## Code changes

Update text in the modal to show the name of the kernel that is getting restarted.

Grammar suggestions welcome.

## User-facing changes

| Before | After |
| --- | --- |
|  image | Screenshot from 2022-03-20 02-35-42 |

Changes translated string so should not be backported to already released 3.3.x.

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -1353,7 +1353,8 @@ export const sessionContextDialogs: ISessionContext.IDialogs = {
     const result = await showDialog({
       title: trans.__('Restart Kernel?'),
       body: trans.__(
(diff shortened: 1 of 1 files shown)


krassowski:
A less subtle change would be to make it:

> **Restart Untitled.ipynb Kernel?**
> Do you want to restart the current kernel? All variables will be lost.
> Kernel path: path/to/Untitled.ipynb

Which would allow to disambiguate the case of several files with the same name but in different sub-directories, and make the short name more pronounced in general.

jtpio (approved):
Thanks!

This already looks like a nice improvement :+1:

jtpio:
CI failures are not related with fixes incoming in separate PRs.

Issue #1366, opened 2016-12-04
Open terminal in the cwd of the file browser


dclong:
When opening a new terminal, I think the best working directory should be the current directory in the file explore or the current working directory of the notebook. However, it seems that neither of them of used. The working directory of a new terminal is confusing to me.

blink1073:
Hi @dclong, this issue was being tracked in #782, but I like your title better ;).

ellisonbg:
This could be implemented on top of #2855

ellisonbg:
Then, no server change is required.

Madhu94:
I'd like to try fixing this.

nathancarter:
When using JupyterLab for education, this is a very annoying missing feature.  It means that if I want to teach even something as simple as how to run a Python script with "python file_name.py" I have to first teach

 * `ls`/`pwd`/`cd` to the students who brought Mac laptops,
 * `dir`/`cd`/`cd` to the students who brought PC laptops,
 * and quoting/escaping folder names for those whose paths contain spaces,

in total a huge distraction from the actual educational goals of the lesson.

Seems like a small feature, but ticket open for two years...perhaps it could be reprioritized?  Thank you!

vidartf:
@nathancarter contributions welcome!

vidartf:
More constructively: The efforts of the core contributors are quite thinly stretched, and there are a *lot* "small" issues. This means that the core contributors have no chance of clearing all of them, so we are dependent on other contributors for helping carry the load. If you think that this feature is worthwhile to you, and the feature is indeed a small feature, then hopefully this is something you could help us with. We are more than happy to spend some time in helping new contributors, in the hope that they will also end up contributing further fixes later on.

nathancarter:
Hey @vidartf ...I think we had a beer in Luxembourg together last year.  Hi again!

This seems like a small enough issue that I might be able to attack it.  Right now I'm swamped, but I spent a little time this morning looking over the jupyterlab source for the first time, and I think I at least see where to start, which I'll record here for myself or anyone else later:

 * [Here](https://github.com/jupyterlab/jupyterlab/blob/master/packages/terminal-extension/src/index.ts#L293) is the line where new terminals are created.  This would seem to be the line to tweak, passing in a cd to the new directory as the initial command.
[...]

PR #12250, opened 2022-03-21
Open terminal in cwd from launcher

rccern:
This PR is a draft, I am posting this now in order to receive any feedback and to see if there is any interest in further developing it.

## References

Fix https://github.com/jupyterlab/jupyterlab/issues/1366

## Code changes

I added a parameter, `cwd`, to `ITerminal.IOptions`, that is passed to the server and from there to Terminado. 

## User-facing changes

The terminals started from the launcher now open in the folder open in the filebrowser. Does this behavior need to be hidden behind some configuration option (especially for 3.x)?

I'm not sure it's a change to public APIs, but the terminals started from the launcher now open in the folder open in the filebrowser.


fcollonval (changes requested):
This looks great @rccern I left some suggestion (mainly to add missing documentation).

Then some tests will certainly fail (at least the integration tests). But this is expected. If you need help to fix them let us know.

rccern:
I didn't add any test, but it seems that there a no test for the creation behavior in  `packages/terminal`. I'm having issues runnin `jlpm test` from the root of the repo, with failures in the  `docregistry` package. 
Please let me know if I should add tests (and some info on how to get started in case ;) )

packages/services/src/terminal/terminal.ts
```diff
@@ -19,9 +19,10 @@ export { IModel, isAvailable };
 export namespace ITerminal {
   export interface IOptions {
     /**
-     * Terminal name.
+     * Terminal name and directory it will start in.
      */
-    name: string;
+    name?: string;
+    cwd?: string | undefined;
```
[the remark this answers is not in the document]
rccern: you are right, my comment https://github.com/jupyterlab/jupyterlab/pull/12250#discussion_r831202428 is wrong... the problem, then, is that this PR doesn't work when starting notebook in the root folder, because the request we send to terminado has issues with `''`  as cwd. There are several places where this can be fixed, but maybe the easiest is to go back to my original test `if (cwd)`?

packages/services/src/terminal/restapi.ts
```diff
@@ -34,17 +34,20 @@ export interface IModel {
...
-  name?: string
+  name?: string,
+  cwd?: string
 ): Promise<IModel> {
   Private.errorIfNotAvailable();
   const url = URLExt.join(settings.baseUrl, TERMINAL_SERVICE_URL);
   const init = {
     method: 'POST',
-    body: JSON.stringify({ name })
+    body: JSON.stringify({ name, cwd })
```
fcollonval, line 50: Let's be forgiven as the type is `string`:

```suggestion
    body: JSON.stringify({
      name,
      cwd: cwd ? cwd : undefined
    })
```

rccern:
I investigated this a bit, and it seems that there is no problem with spaces or `-` . 
What happens is that if you start lab when you are in a dir "/home/user/lab", but you pass e.g. `--ServerApp.root_dir=/tmp`  (which is basically what the server_config for galata tests here https://github.com/jupyterlab/jupyterlab/blob/master/galata/README.md does with mkdtemp), the terminal will be started in "/home/user/lab".

With my change, if i go to `/tmp` , create a `test` folder, open the launcher there and try to start a terminal, it will try to open it in `/home/user/lab/test`  which doesn't exist, hence the failure.

In my understanding, I would need to append the value of "cwd" to the value of `ServerApp.root_dir`, but I guess that info is not available in the lab frontend? Where is the best place to add this?  jupyter_server [here](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/terminal/api_handlers.py#L27)?

fcollonval:
Thanks for looking into that @rccern 

Indeed the best place is the one you are referring to in jupyter-server. You should be able to access the root directory withing the handler with `self.settings["server_root_dir"]`. Would you mind opening a PR for that in Jupyter Server as this PR will break otherwise without an easy way to notify the user on the reason for the failure?

fcollonval:
Let's finish this one @rccern now that Jupyter Server [1.16.0](https://github.com/jupyter-server/jupyter_server/releases/tag/v1.16.0) is out

fcollonval:
I don't have rights to push to your branch. Would you mind applying the following patch?

```diff
diff --git a/galata/test/jupyterlab/terminal.test.ts b/galata/test/jupyterlab/terminal.test.ts
index cf6b51e3b4..e694c24f7a 100644
--- a/galata/test/jupyterlab/terminal.test.ts
+++ b/galata/test/jupyterlab/terminal.test.ts
@@ -84,3 +84,14 @@ test.describe('Terminal', () => {
     });
   });
 });
+
+test('Terminal should open in Launcher cwd', async ({ page, tmpPath }) => {
+  await page.waitForSelector(`.jp-Launcher-cwd > h3:has-text("${tmpPath}")`);
+
+  await page.locator('[role="main"] >> p:has-text("Terminal")').click();
+
+  const terminal = page.locator(TERMINAL_SELECTOR);
+  await terminal.waitFor();
+
+  expect(await terminal.screenshot()).toMatchSnapshot('launcher-term.png');
+});
diff --git a/setup.cfg b/setup.cfg
index e3d97025b7..bac41e5e95 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     tornado>=6.1.0
     jupyter_core
     jupyterlab_server~=2.10
-    jupyter_server~=1.8
+    jupyter_server~=1.16
     notebook_shim>=0.1
     jinja2>=3.0.3
```

rccern:
I changed a bit setup.cfg (there is a `>=1.15.6` in master and i changed that to 1.16.0) and rebased

fcollonval:

rccern:

galata/test/jupyterlab/terminal.test.ts
```diff
@@ -84,3 +84,14 @@ test.describe('Terminal', () => {
...
 });
+
+test('Terminal should open in Launcher cwd', async ({ page, tmpPath }) => {
+  await page.waitForSelector(`.jp-Launcher-cwd > h3:has-text("${tmpPath}")`);
+
+  await page.locator('[role="main"] >> p:has-text("Terminal")').click();
+
+  const terminal = page.locator(TERMINAL_SELECTOR);
+  await terminal.waitFor();
+
```
fcollonval, line 95: Let's print current directory to check we are at the expected position

```suggestion

    await page.keyboard.type('pwd');
    await page.keyboard.press('Enter');
```

rccern:
I tried to generate the snapshots locally but i'm failing at using the Dockerfile in the repo

fcollonval:
Thanks for pushing this @rccern 

Actually I miss the fact that the path won't be reproducible as the random temporary folder is appearing in the output:

image

And as the terminal is a `canvas` we cannot introspect the text. So let's drop the test and go ahead merging this one.

galata/test/jupyterlab/terminal.test.ts
```diff
@@ -84,3 +84,16 @@ test.describe('Terminal', () => {
...
+  await page.keyboard.type('pwd');
```
krassowski, line 96: What about using `basename "$PWD"` instead to just get the name of the directory (and not the full path)? @fcollonval

fcollonval: It would indeed be good to test that.

fcollonval:
> I tried to generate the snapshots locally but i'm failing at using the Dockerfile in the repo

The bot is unable to push the new snapshots due to GitHub rights on the branch. It is though to reproduce snapshots locally as the terminal is highly dependent of the OS setup.

The best solution here is to download the `jupyterlab-galata-test-assets` archive from the UI tests (for [example](https://github.com/jupyterlab/jupyterlab/actions/runs/2076027819)). And then copy the file `/test-results/test-jupyterlab-terminal-Terminal-should-open-in-Launcher-cwd-jupyterlab-retry1/launcher-term-actual.png` to `galata/test/jupyterlab/terminal.test.ts-snapshots/launcher-term-jupyterlab-linux.png`

rccern:
I added two sleeps, maybe they are not needed when the shell starts more rapidly than it does on my machine (heavy bashrc)

packages/services/src/terminal/terminal.ts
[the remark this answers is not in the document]
fcollonval: Thanks a lot for pushing the test through @rccern. Would you mind adding the above suggestion? After that we are good to go.

fcollonval (approved):
Thanks a lot @rccern for the patience

> This will be part of 3.4.0 (ETA end of April).

jtpio:
Thanks!

rccern:
Thanks for your help @fcollonval !

nathancarter:
This is an important feature - thank you for getting it in!!

fcollonval:

Issue #7394, opened 2019-10-21
HTML tables from pandas: adjust font size?


allefeld:
## Description

I made the font sizes in Jupyter Lab larger through the Settings (Theme / User Preferences):

```
{
    // Theme
    // @jupyterlab/apputils-extension:themes
    // Theme manager settings.
    // *************************************

    // Theme CSS Overrides
    // Override theme CSS variables by setting key-value pairs here
    "overrides": {
        "code-font-size": "12pt",
        "content-font-size1": "12pt",
        "ui-font-size1": "10pt",
    }
}
```

However, the font size in pandas-generated HTML tables doesn't seem to be affected by that:

![Screenshot_20191021_194032](https://user-images.githubusercontent.com/1738353/67233611-797d1880-f43b-11e9-8737-87a7b4128ce2.png)

Is there a way to adjust these fonts, too?

I understand that pandas itself has a lot of styling options. However, I'm looking for a way to change the default appearance, without extra programmatic effort in every notebook.

- Operating System and version: Debian 10.1 (buster)
- Browser and version: Chromium 76.0.3809.100
- JupyterLab version: 1.1.4

vidartf:
If you "inspect" the table text in the browser web tools, you should be able to find an option "get computed styles" (or something similar. This often shows which are the current styling options in effect, *and where they are coming from*. Hopefully this can help in tracking down which bit of styling is controlling the font size.

allefeld:
@vidartf thanks for replying. This is what I find:

![image](https://user-images.githubusercontent.com/1738353/67399162-b401c500-f5a3-11e9-8f43-c10b9ef52c68.png)

So apparently the content font size modified by me is overridden. The 12px appears to be hardcoded in one of the many many `<style>` sections of the page:

![image](https://user-images.githubusercontent.com/1738353/67400215-3ccd3080-f5a5-11e9-93b3-a42496fd04fe.png)

It is however not present in the original page source, but seems to be inserted via Javascript. I tried to find the corresponding css file in the repository, but without luck.

Anyway, I think this should be fixed. There is seldomly a good reason for hard-coded CSS overrides, and in this case, having `.jp-RenderedHTMLCommon table` to simply inherit thre font size from `.jp-RenderedHTMLCommon` appears to me to be the right thing.

Sorry for the images, I wasn't able to copy text from the Inspector.

vidartf:
Thanks! This seems to be the origin of that style:

[...]

PR #12255, opened 2022-03-22
Use css variable for font size.

Carreau:
Those are minor change of size (12px to 13px), and 0.8em to 0.86,
thigh they will fix issues, like nbconvert slides trying to bump up the
size by updating the variables values.

### DIFF
--- a/packages/rendermime/style/base.css
+++ b/packages/rendermime/style/base.css
@@ -413,7 +413,7 @@
   border-spacing: 0;
   border: none;
   color: var(--jp-ui-font-color1);
-  font-size: 12px;
+  font-size: var(--jp-ui-font-size1);
   table-layout: fixed;
   margin-left: auto;
   margin-bottom: 1em;
@@ -600,7 +600,7 @@ h6:hover .jp-InternalAnchorLink {
   border-radius: 3px;
   box-shadow: inset 0 -1px 0 rgba(0, 0, 0, 0.25);
   display: inline-block;
-  font-size: 0.8em;
+  font-size: var(--jp-ui-font-size0);
   line-height: 1em;
   padding: 0.2em 0.5em;
 }


fcollonval:

fcollonval:
Edit bot commit to remove bad snapshot update

fcollonval (approved):
Thanks @Carreau

krassowski:
1. This closes https://github.com/jupyterlab/jupyterlab/issues/7394 and is relevant to the discussion in there
2. What do you think about preserving the old font size by using `calc`, e.g. `calc(12/13 * var(--jp-ui-font-size1))`? This way we get the benefit of customization, but keep the status quo of the font size in table being slightly smaller.

krassowski:
This was my concern at first too, but I tested on Chrome and it works well. Also had a quick glance at known issues for cros--browser support at https://caniuse.com/calc and I do not see a reason for that to be a problem, but we probably should also test on Safari (because Safari).

jtpio:
Looks like this picked up some conflicts.

packages/rendermime/style/base.css
```diff
@@ -413,7 +413,7 @@
   border-spacing: 0;
   border: none;
   color: var(--jp-ui-font-color1);
-  font-size: 12px;
+  font-size: var(--jp-ui-font-size1);
```
krassowski, line 416:
```suggestion
  font-size: calc(12/13 * var(--jp-ui-font-size1));
```

krassowski (commented):
This should allow to eliminate the visual changes to table size (for the most part).

packages/rendermime/style/base.css
Carreau: the main question is do we _want_ to have table be 12/13 of the normal font size just for the sake of keeping a value that was chosen 6 years ago ? 

I haven't been involved in JupyterLab design in a long time, so I don't want my opinion to weight too much on this, but I believe that uniformity is better, and would personally avoid calc even if it make a slight change of theme

fcollonval:
As we are planning a 3.4 release, I suggest to acknowledge the change from 12px to `jp-font-size1` (aka 13px per default) in that release.

Carreau:
> As we are planning a 3.4 release, I suggest to acknowledge the change from 12px to `jp-font-size1` (aka 13px per default) in that release.

What would be the right place to do so ?

fcollonval:
Merging in master...

>  What would be the right place to do so ?

I added this PR in the roadmap for 3.4.0 #12268

Carreau:
Thanks !

fcollonval:

PR #12257, opened 2022-03-22
Update links to create new issues in README.md

jtpio:
## References

cc @jweill-aws 

Maybe this can explain why many issues are being created without the proper `status:Needs Triage` label. Which is mitigate by the JupyterLab probot afterwards.

Maybe there is also another link somewhere else to create issues with the `bug` or `enhancement` labels, but without the `status:Needs Triage` label

## Code changes

Update links to create new issues in README.md

## User-facing changes

Uses should more often open issues with the proper labels from the start.

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,11 @@ We encourage you to ask questions on the [Discourse forum](https://discourse.jup
 
 ### Bug report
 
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Good catch thanks @jtpio

jweill-aws:
Looks good — thank you for adding this!

PR #12258, opened 2022-03-22
Default is no virtual rendering + Relax virtual notebook rendering and ensure no structural change until rendering is completed

echarles:
## References

Continuation of https://github.com/jupyterlab/jupyterlab/pull/12166 PR done by @rtrg

Closes https://github.com/jupyterlab/jupyterlab/pull/12166

## Code changes

The rendering virtual notebook rendering of the cells is rescheduled if no idle time can be found by the browser.
The notebook actions which change the structure of the notebook are disabled and an information message is shown to the user.
To ensure user is not impacted, the default setting for the cells to be rendered directly is set to a high number (99999) so that this feature will be an opt-in.

## User-facing changes

Untitled4

None.

### DIFF
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2000,12 +2000,21 @@ namespace Private {
     activeCell: Cell | null;
   }
 
-  export function isNotebookRendered(notebook: Notebook): boolean {
+  export function isNotebookRendered(
+    notebook: Notebook,
+    translator?: ITranslator
+  ): boolean {
+    translator = translator || nullTranslator;
+    const trans = translator.load('jupyterlab');
+
     if (notebook.remainingCellToRenderCount !== 0) {
       showDialog({
-        body: `Notebook is still rendering and has for now ${notebook.remainingCellToRenderCount} remaining cells to render.
+        body: trans.__(
+          `Notebook is still rendering and has for now (%1) remaining cells to render.
 
 Please wait for the complete rendering before invoking that action.`,
+          notebook.remainingCellToRenderCount
+        ),
         buttons: [Dialog.okButton({ label: 'Ok' })]
       }).catch(reason => {
         console.error(
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2015,7 +2015,7 @@ namespace Private {
 Please wait for the complete rendering before invoking that action.`,
           notebook.remainingCellToRenderCount
         ),
-        buttons: [Dialog.okButton({ label: 'Ok' })]
+        buttons: [Dialog.okButton({ label: trans.__('Ok') })]
       }).catch(reason => {
         console.error(
           'An error occurred when displaying notebook rendering warning',
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2000,9 +2000,7 @@ namespace Private {
     activeCell: Cell | null;
   }
 
-  export function isNotebookRendered(
-    notebook: Notebook,
-  ): boolean {
+  export function isNotebookRendered(notebook: Notebook): boolean {
     const translator = notebook.translator;
     const trans = translator.load('jupyterlab');
 
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2003,7 +2003,7 @@ namespace Private {
   export function isNotebookRendered(
     notebook: Notebook,
   ): boolean {
-    translator = notebook.translator;
+    const translator = notebook.translator;
     const trans = translator.load('jupyterlab');
 
     if (notebook.remainingCellToRenderCount !== 0) {
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -2002,9 +2002,8 @@ namespace Private {
 
   export function isNotebookRendered(
     notebook: Notebook,
-    translator?: ITranslator
   ): boolean {
-    translator = translator || nullTranslator;
+    translator = notebook.translator;
     const trans = translator.load('jupyterlab');
 
     if (notebook.remainingCellToRenderCount !== 0) {
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -149,6 +149,9 @@ export namespace NotebookActions {
       return;
     }
 
+    if (!Private.isNotebookRendered(notebook)) {
+      return;
+    }
     const state = Private.getState(notebook);
 
     notebook.deselectAll();
@@ -244,6 +247,9 @@ export namespace NotebookActions {
       return;
     }
 
+    if (!Private.isNotebookRendered(notebook)) {
+      return;
+    }
     const state = Private.getState(notebook);
     const toMerge: string[] = [];
     const toDelete: ICellModel[] = [];
@@ -340,6 +346,9 @@ export namespace NotebookActions {
       return;
     }
 
+    if (!Private.isNotebookRendered(notebook)) {
+      return;
+    }
     const state = Private.getState(notebook);
 
     Private.deleteCells(notebook);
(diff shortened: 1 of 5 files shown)


echarles:
@fcollonval This is the PR I have mentioned during our last performance meeting. As you said, @krassowski will probably like it as it sets the default value for the cells to be rendered directly to a very high number.

packages/notebook/src/widget.ts
```diff
@@ -841,6 +856,10 @@ export class StaticNotebook extends Widget {
     this._renderedCellsCount++;
   }
 
+  public get cellsToRender(): Map<string, { index: number; cell: Cell }> {
+    return this._toRenderMap;
+  }
```
fcollonval, line 861: I would rather return the remaining number of cells rather than exposing that sensitive map publicly.

packages/notebook/src/actions.tsx
```diff
@@ -1966,6 +2000,19 @@ namespace Private {
...
   }
 
+  export function isNotebookRendered(notebook: Notebook): boolean {
+    if (notebook.cellsToRender.size !== 0) {
+      showDialog({
+        body: `Notebook is still rendering and has for now ${notebook.cellsToRender.size} remaining cells to render.
+
+Please wait for the complete rendering before invoking that action.`,
+        buttons: [Dialog.okButton({ label: 'Ok' })]
+      }).then(() => undefined);
```
fcollonval, line 2010: It makes more sense to add the `catch` clause than an empty `then` clause.

```suggestion
      }).catch((reason) => { console.error('An error occurred when displaying notebook rendering warning', reason); });
```

packages/notebook/src/actions.tsx
```diff
@@ -1966,6 +2000,19 @@ namespace Private {
     activeCell: Cell | null;
   }
 
+  export function isNotebookRendered(notebook: Notebook): boolean {
+    if (notebook.cellsToRender.size !== 0) {
+      showDialog({
+        body: `Notebook is still rendering and has for now ${notebook.cellsToRender.size} remaining cells to render.
+
+Please wait for the complete rendering before invoking that action.`,
```
fcollonval, line 2008: This needs to be translated.

packages/notebook/src/actions.tsx
```diff
@@ -1966,6 +2000,19 @@ namespace Private {
...
+        buttons: [Dialog.okButton({ label: 'Ok' })]
```
fcollonval, line 2009: _Ok_ needs translation too.

fcollonval (changes requested):
Thanks @echarles 

Does this closes https://github.com/jupyterlab/jupyterlab/pull/12166?

If so, would you mind addressing https://github.com/jupyterlab/jupyterlab/pull/12166#issuecomment-1059157762?

packages/notebook/src/widget.ts
echarles: I was hesitant on that one while implementing. That should be considered as sensitive, but on the other side the model is anyway exposed, which is even more sensitive. I was thinking that extension developer could benefit from more details to show visuals on unrendered cells. But yes, exposing a simple number sounds safer and simpler, I have implemented that now. Thx.

packages/notebook/src/actions.tsx
echarles: I also faced that question while implementing. The translator is not given in the method signature of the NotebookActions class, and it looks overkill to me to pass that in all the static methods. Any idea/advice on that?

echarles:
@fcollonval Thx a lot for the reviews and inputs.

> Does this closes https://github.com/jupyterlab/jupyterlab/pull/12166?

Yes. I have updated the initial comment.

> If so, would you mind addressing https://github.com/jupyterlab/jupyterlab/pull/12166#issuecomment-1059157762?

These are your comments:

> Regarding configuration, for me the one that needs customization is more the 50 time slot than the timeout. Requesting a full idle cycle, implies that at best one cell will be rendered per idle cycle otherwise it will be rescheduled (if I got the code correctly). We could imagine that having that setting equals to 0 will fallback to the current behavior - otherwise it will use rescheduling.

Done

> You should not reschedule if the notebook panel isDisposed().

Done.

> Would it be possible to use [cancelIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelIdleCallback) for browsers supporting it?

Done.

packages/notebook/src/actions.tsx
echarles: I have added the default translator in the isNotebookRendered

packages/notebook/src/actions.tsx
echarles: I have added the default translator in the isNotebookRendered

echarles:
@fcollonval I don't think the CI failures are related to this PR. Before looking at that, could you check if the changes I have pushed address your comments? Thank you.

fcollonval (commented):
Thanks @echarles 

The error on `js-notebook` and integration tests are legit.

It seems that a cell is not set to active when a notebook is opened. This is the reason for the integration tests failures and at least some of the js-notebook.

packages/notebook/src/actions.tsx
```diff
@@ -1966,6 +2000,33 @@ namespace Private {
     activeCell: Cell | null;
   }
 
+  export function isNotebookRendered(
+    notebook: Notebook,
+    translator?: ITranslator
+  ): boolean {
+    translator = translator || nullTranslator;
```
fcollonval, line 2007: You can get the translator from the notebook widget:

```suggestion
  ): boolean {
    translator = notebook.translator;
```

echarles:
I have rebased on latest master, and the js-notebook tests are failing with the same errors on th CI. Those errors does not look to me like being caused to the changes of this PR, but they don't occur in other PRS...

On my local env, the js-notebook tests are successfull with `cd packages/notebook && jlpm run test:cov` or somethimes fail (falkiness?) on the following unrelated test, which I also encounter with the master branch.

````
test/executionindicator.spec.tsx
  @jupyterlab/notebook
    ExecutionIndicator
      executedAllCell
        ✕ should count correctly elapsed time
```

echarles:
update from my local env: ` ✕ should count correctly elapsed time` is only on master, this branch returns all green test for `js-notebook`.

```
Test Suites: 10 passed, 10 total
Tests:       436 passed, 436 total
Snapshots:   0 total
Time:        42.994 s, estimated 55 s
Ran all test suites.

All tests passed
```

echarles:
I have rebased this PR on latest master and less ui regression are reported (4 instead of 7). I put here 2 regression diffs. How can this PR generate such diffs?

opened-menu-run-diff

copy-paste-cell-diff

fcollonval:
I don't know the link between this PR and the reason. But the errors are related to the non-activation of cells (for example the menu items are dis/enabled depending on the active cell state).

fcollonval:
You should fix the `js-notebook` unit tests as I'm pretty sure they have the same root cause.

echarles:
> You should fix the js-notebook unit tests as I'm pretty sure they have the same root cause.

Great call! js-notebook and Visual regression tests are green.

This is ready for review. Thx again.

fcollonval:
Once prettified this is good to go.

echarles:
Apart from the expect Documentation Regression, all is now green. Ready to be merged.

fcollonval (approved):
Thanks @echarles

echarles:
@fcollonval Should we rerun the benchmark? The previous run leads to slower large notebook opening, which is probably expected as we are have no virtual notebook by default.

echarles:
I have changed the title of this PR to clearly state `Default is no virtual rendering` which makes the loading slower as expected.

I will leave this opened a few days for any feedback before mergin.

echarles:
@fcollonval Once merged to master, could we backport this to 3.3.x or should we rather wait on 3.4.x? On my side, I don't see any breaking change that would not allow to backport to 3.3.x.

echarles:
Thx @fcollonval . The backport to 3.3.x is open on https://github.com/jupyterlab/jupyterlab/pull/12394

Issue #10405, opened 2021-06-14
Disable upload button


kwon14:
Hello.!
Thank you very much for your hard work.

Disable the upload feature as a usage policy.
As a result of browsing several websites
jupyterlab/packages/filebrowser/src/browser.ts
I've seen that you can just delete the "this.toolbar.addItem('upload', uploader);" line of code.

the question is
1. Is there a way to find out what the browser.ts file is?
2. Is there a way to block the upload button or function?

As always, thank you for your reply
Thank yo
![스크린샷 2021-06-14 오후 4 43 17](https://user-images.githubusercontent.com/64061288/121857358-80ebe680-cd30-11eb-97db-7fd65179b2bb.png)
u.!!

jtpio:
Thanks @kwon14.

Maybe this functionality could then be moved to its own plugin so it can be disabled with the `jupyter labextension disable` command?

Similar to the download UI and feature: https://github.com/jupyterlab/jupyterlab/pull/10065

jtpio:
Somewhat related: https://github.com/jupyterlab/jupyterlab/issues/7451

Jose-Matsuda:
Hello @jtpio, I've been going through issues in this and the notebook repo looking for more information on blocking the upload of files via drag and drop into jupyter but haven't found anything solid in terms of solutions or next steps. Would you happen to have any suggestions for that? 

Something we tried was blocking the PUT network call to the path of notebook/notebookName/api/contents/ (via an [Istio Authorization policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/)) but as it turns out that the PUT to that path is the same that is used when jupyter saves a file so that wouldn't work.

jtpio:
@Jose-Matsuda right, the drag and drop upload is implemented in the file browser directly, so it's not as easy as disabling a core JupyterLab plugin unfortunately.

https://github.com/jupyterlab/jupyterlab/pull/17676 explores adding a new option to disable file uploads via drag and drop, although it would also need a way to prevent users from going to the settings editor and changing that option.

Otherwise for the original issue of disabling the upload button, it should be possible to put something like the following in an `overrides.json` file:

```json
{
  "@jupyterlab/filebrowser-extension:widget": {
    "toolbar": [
      {
        "name": "uploader",
        "disabled": true
      }
    ]
  }
}
```

MMesch:
[...]

Issue #12196, opened 2022-03-11
Customize the file browser toolbar via the settings


jtpio:
### Problem

Just like the notebook toolbar or file editor toolbar, it would be great if the file browser toolbar could also be configured via the settings.

![image](https://user-images.githubusercontent.com/591645/157823576-3c50ccf3-3919-48c4-867f-5c0815f08e57.png)

### Proposed Solution

Maybe we can follow a similar approach as the other factories, and place the toolbar items in the settings definition.

### Additional context

This could also prove useful for customizing the notebook v7 toolbar (see https://github.com/jupyterlab/retrolab/issues/260#issuecomment-971202618 for more context):

![image](https://user-images.githubusercontent.com/591645/157824014-30a4b4bd-874e-4b88-95e7-4d1571659344.png)

jtpio:
Adding to 4.0 for consideration.

jtpio:
Looks like the current toolbar customization applies to the *document* factories.

Not sure it will be possible to adapt the current pieces such as `createToolbarFactory` so they work on generic widgets with a toolbar.

cc @fcollonval who might have some input on this.

jtpio:
Ideally in the case Notebook v7 it would be possible to register a new factory like it is already done here for the interface switcher in the notebook toolbar:

https://github.com/jupyter/notebook/blob/30310847d6804294cbeb274eeb91d126c86669f0/packages/lab-extension/src/index.ts#L140-L152

And define overrides in a custom downstream settings file, for example to be able to make the toolbar look a bit more like classic if needed (disabling some items, inserting spacers to pull content to the right):

![image](https://user-images.githubusercontent.com/591645/160128286-d107f6a7-05d1-4438-9c79-939a877c726b.png)

jtpio:
One thing that could be tricky is the fact there can be multiple file browsers. It's not clear whether it will be possible to have something generic enough without making it too complicated.

Issue #7451, opened 2019-11-01
Disable file upload


cometta:
May I know how to disable file up load/ drap an drop file to upload for Jupyterlab ?  package/filebrowser is not an extension.

vidartf:
Here it would be useful to understand what problem you are trying to solve. E.g. why are you trying to disable file uploading?

cometta:
We do not want user to upload their own notebook into Jupiter hub. Instead only allow to run existing notebooks on the workspace

vidartf:
Then this should be done on the server side, to ensure that users cannot work around it. Do you still want users to be able to edit existing notebooks?

Jose-Matsuda:
I apologize for commenting on an older issue, would there be any suggestions to turning off drag and drop upload functionality?

As we serve notebooks via k8s, something we tried was blocking the PUT network call to the path of notebook/notebookName/api/contents/ (via an [Istio Authorization policy](https://istio.io/latest/docs/reference/config/security/authorization-policy/)) but as it turns out that the PUT to that path is the same that is used when jupyter saves a file so that wouldn't work, as we want users to still be able to work in these notebooks and save their work.

PR #12275, opened 2022-03-25
Move the file browser upload button to a separate plugin so it can be disabled

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/10405

Should also help with https://github.com/jupyter/notebook/issues/5770

However this does not prevent dragging and dropping files to the listing to perform an upload.

## Code changes

Move the file upload button to a separate plugin so it can be disabled with:

```
jupyter labextension disable @jupyterlab/filebrowser-extension:file-upload-button
```

## User-facing changes

After running:

```
jupyter labextension disable @jupyterlab/filebrowser-extension:file-upload-button
```

**Before**

image

**After**

image

None

### DIFF
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -22,7 +22,6 @@ import { PanelLayout, Widget } from '@lumino/widgets';
 import { BreadCrumbs } from './crumbs';
 import { DirListing } from './listing';
 import { FilterFileBrowserModel } from './model';
-import { Uploader } from './upload';
 
 /**
  * The class name added to file browsers.
@@ -92,7 +91,6 @@ export class FileBrowser extends Widget {
       },
       tooltip: this._trans.__('New Folder')
     });
-    const uploader = new Uploader({ model, translator: this.translator });
 
     const refresher = new ToolbarButton({
       icon: refreshIcon,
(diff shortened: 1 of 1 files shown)


jtpio:
> However this does not prevent dragging and dropping files to the listing to perform an upload.

Not linking to https://github.com/jupyterlab/jupyterlab/issues/7451 because the drag and drop is not disabled (but we should look into being able to disable it in a separate change).

jtpio:
Looks like the CI failures are also happening on master: https://github.com/jupyterlab/jupyterlab/runs/5691286994?check_suite_focus=true

Needs investigation...

krassowski:
I wonder if it would be better to move the buttons to schema so that they can be disabled by overrides. For example, one may want to disable the blue plus button as we now have + on the tabbar (https://github.com/jupyterlab/jupyterlab/pull/12195). This would also give us an extra benefit of ability to re-arrange buttons in settings editor. This does not preclude moving the upload button to a separate plugin of course (but then we would not want to move all buttons to separate plugins, right?).

jtpio:
> move the buttons to schema so that they can be disabled by overrides

Yes that's a good point, and would fix https://github.com/jupyterlab/jupyterlab/issues/12196.

Maybe we should go with this approach directly then, will look into it.

jtpio:
Closing in favor of https://github.com/jupyterlab/jupyterlab/pull/12281

PR #12276, opened 2022-03-25
Allow linear and radial gradient

### LINKED ISSUE #12093: allow linear-gradient in background-image html styling within Markdown cells
## Description

Styling such as `style="background-image:linear-gradient(green,blue);"` within the styling of `div` or other element specified within a markdown cell is not rendered. Plain background images are rendered.

Unless there is a security issue I am unaware of, I think gradient styling should be allowed.

## Reproduce

Enter the following in a markdown cell and run it to render:
```
<div style = "height: 20px; width:100%; background-image:linear-gradient(green,blue);"></div>
<div style = "height: 20px; width:100%; background-color:blue;"></div>
```
The first will not be shown, but the second will.

## Expected behavior

The expected behavior is that the gradient will be rendered.
## Context

- Operating System and version: all
- Browser and version: at least Chrome and Firefox
- JupyterLab version: 3.1.1 (latest I could get to launch in Binder from the repositories).

krassowski:
## References

Fixes #12093.

Known limitation: because we use "simplified" background (allowing only one layer), in order to set multiple overlying gradients users need to use `background-image` instead of `background`. We could rewrite the regex to accept multi-layer background in `background` but that is out of scope for this PR.

## Code changes

- [`linear-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient) and [`radial-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient) are now accepted by sanitizer using definitions per MDN.
- regular expression for validating floats in range 0-1 now accepts `.x` notation (omission of the leading zero)

## User-facing changes

Users should have an easier time migrating notebooks from Jupyter Notebook v6 (which previously was allowing for gradients).

Test code:

```html
<div style="background: linear-gradient(lightgreen, red); height: 100px; width:100px">Simple linear gradient in background</div>
<div style="background-image: linear-gradient(217deg, rgba(255,0,0,.8), rgba(255,0,0,0) 70.71%), linear-gradient(127deg, rgba(0,255,0,.8), rgba(0,255,0,0) 70.71%), linear-gradient(336deg, rgba(0,0,255,.8), rgba(0,0,255,0) 70.71%); height: 100px; width:100px">Compounded linear gradient in background-image</div>
<div style="background-image: radial-gradient(ellipse at top, #e66465, transparent), radial-gradient(ellipse at bottom, #4d9f0c, transparent); height: 100px; width:100px">Compounded radial gradient in background-image</div>
```

| Before | After |
| --- | --- |
| Screenshot from 2022-03-25 19-17-43 | Screenshot from 2022-03-25 19-18-31 |

None


fcollonval (approved):
Thanks @krassowski 

CI failure are not related

Issue #12272, opened 2022-03-24
Switch to Pre-Commit


blink1073:
### Problem
There are several checks that need to by run on the code base.  Some of them are run using `husky`, and slower ones are run on CI.  Additionally, if there is a lint or integrity error, a maintainer or the original PR author has to manually run the command and push the commit.

### Proposed Solution
Now that `meeseeks` supports running `pre-commit`, we should switch to using it. 

- We can add external or local repo checks for existing husky checks
- We would use manual hook for `integrity` and `eslint:check:typed` so it only runs in CI and by meeseeks.
- We could add `pre-commit  install` to editable install since `husky` is currently installed there
- We would add everything used by [`jupyter_server`](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/.pre-commit-config.yaml) except `black`
- We could add `black` and other auto-formatters as a follow up commit

- Then we would add flake8 support and and add a [.git-blame-ignore-revs](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/.git-blame-ignore-revs) file in a follow up PR
- Finally, we would run the auto-formatters on previous branches to make backports easier

krassowski:
:+1: Adding `.git-blame-ignore-revs` would also close #10607; it seems that some kind of GitHub support for ignoring revisions landed recently (as an experimental function? I cannot find any announcement, but saw a notification on GitHub)

blink1073:
Oh sweet!  https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

fcollonval:
Closing this one as done

PR #12279, opened 2022-03-28
Use pre-commit

blink1073:
## References
First step of #12272 

## Code changes
- Switches from `husky` to `pre-commit` while running the same checks and same local user experience (auto-install git hooks).
- Adds auto-formatting and checks used by `jupyter_server` (except flake8 which will follow).
- Runs the autoformatter as a separate commit that can be ignored 
- Updates contributing guide

## User-facing changes
None

None


blink1073:
That was easy.  😅

setup.py
```diff
@@ -1,66 +1,85 @@
...
-# Distributed under the terms of the Modified BSD License.
-from os.path import join as pjoin
 import json
 import os
 import os.path as osp
+import subprocess
 import sys
 
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
```
fcollonval, line 11: This got move by mistake I think

fcollonval (commented):
Small lines motion with the license notice. LGTM otherwise

setup.py
blink1073: Fixed

fcollonval (approved):
Thanks @blink1073

bollwyvl:
From a downstream perspective, adding `pre-commit` as part of the `build-system` seems super heavy, as noted on https://github.com/conda-forge/jupyterlab-feedstock/pull/333

vidartf:
@blink1073 This PR switched the lockfile from using yarnpkg to using npmjs (https://github.com/jupyterlab/jupyterlab/pull/12742). Was this intentional? If so, we can switch the default. If it was an accident we can revert it, as currently proposed in that PR.

blink1073:
> Was this intentional?

No, it was not intentional

PR #12293, opened 2022-03-29
Use `importlib` instead of `pkg_resources`

fcollonval:
## Code changes

Use entrypoints instead of pkg_resources as it is more efficient

## User-facing changes

N/A

N/A


blink1073:
There's a couple other usages of `pkg_resources` that probably should be replaced.  It is an implicit runtime dependency on `setuptools`:

```
./tests/test_app.py:18:import pkg_resources
./tests/test_app.py:71:    themes = pkg_resources.resource_string(schema_package, schema_path)
./federated_labextensions.py:426:    import pkg_resources
./federated_labextensions.py:429:        pkg_resources.get_distribution(package)
./federated_labextensions.py:430:    except pkg_resources.DistributionNotFound:
./handlers/yjs_echo_ws.py:10:import pkg_resources
./handlers/yjs_echo_ws.py:18:for ep in pkg_resources.iter_entry_points(group="jupyter_ydoc"):
```

We can replace `resource_string` with:

```
In [1]: from importlib.resources import path
In [2]: path('jupyterlab', '.yarn.js')
Out[2]: PosixPath('/Users/silvester/workspace/jupyterlab/jupyterlab/.yarn.js')
```

We can replace get_distribution with:

```

In [1]: from importlib.metadata import version
In [2]: version('jupyter_server')
Out[2]: '1.16.0'
```

It will raise `importlib.metadata.PackageNotFoundError` on failure.

blink1073:
Note that `importlib_metadata` would have to be used on Python 3.7.

fcollonval:
Thanks for the suggestion @blink1073 - this would be indeed good to replace with `importlib_metadata` all usages of `pkg_resources` now that 3.6 is not longer supported.

blink1073 (approved):
Nice!  You might want to update the title

setup.cfg
```diff
@@ -31,6 +31,8 @@ include_package_data = True
 packages = find:
 python_requires = >=3.7
 install_requires =
+    importlib-metadata>=4.4;python==3.7
```
blink1073, line 34:
```suggestion
    importlib-metadata>=4.4; python_version=="3.7"
```

setup.cfg
```diff
@@ -31,6 +31,8 @@ include_package_data = True
...
+    importlib-resources>=1.4;python<3.9
```
blink1073, line 35:
```suggestion
    importlib-resources>=1.4; python_version<"3.9"
```

setup.cfg
blink1073: cf https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

fcollonval: Thanks @blink1073 - bad memory :s

bollwyvl:
as noted over on https://github.com/jupyterlab/jupyterlab_server/pull/264#discussion_r847346624 there are still a lot of gotchas on the `importlib_*` family of packages and APIs which have not been fun to support: it's pretty hard to reason about their feature history, and there are a lot of bad pins floating around out there which can make them difficult to use, or even install correctly together.

part of me wants to just keep using `entrypoints`, deprecated as it may be, until 3.10 is the oldest supported python.

blink1073 (approved):
Thanks!

PR #12295, opened 2022-03-30
Fix Pre-Commit CI

blink1073:
## References
Follow up to #12279 

## Code changes
- Install ourselves when running `pre-commit` manually, so the local build scripts work.
- Skip the `jsonschema` check since it does not work on pre-commit.ci

## User-facing changes
None

None


blink1073:
Hmm, it seems like this won't work because we can't access the outside internet to install from `pip`/`npm` to get the scripts working on pre-commit.ci.  I'll disable it for this repo.

Issue #12098, opened 2022-02-22
Filebrowser - clicking the folder icon should set the current directory to preferred directory


mlucool:
### Problem

When setting `root_dir` and `preferred_dir`, clicking the folder button in the file browser brings you to `root_dir` instead of `preferred_dir`.

![image](https://user-images.githubusercontent.com/1813603/155197503-bfceeccd-a2cb-4913-8519-0dd5d5a9d990.png)

When preferred directory was added in https://github.com/jupyterlab/jupyterlab/pull/10667, the intent was to make this directory feel like root_dir did in the UI, but allow for users to navigate to a wider range of files.  Therefore, I think it's more correct to change this button to update the current directory to `preferred_dir` (when set).

### Proposed Solution

Per above, clicking on the folder icon should bring the user back to their preferred directory. If no preferred directory is set, it should work as it does today.

### Additional context

N/A

fcollonval:
The trouble by changing that behavior it will make it complex to navigate to the `root_dir`. Some maybe an additional button or making it configurable.

fcollonval:
Another possibility is to use the [jupyterlab-favorites](https://github.com/jupyterlab-contrib/jupyterlab-favorites) extension.

mlucool:
Favorite looks pretty neat.

Humm, you are right, getting to `root_dir` would be hard. Maybe we add a home icon in addition in cases when a user have preferred dir set?

echarles:
An additional icon sounds better, but this needs to be challenged in terms of design and accessibility. I have quickly tried a few attempts to add an icon that would go to the `preferred_dir `, but usually I fail to make it right, and in this case, I fear that it will clutter an already crowded area. Any alternative cc/ @isabela-pf @andrewfulton9  

<img width="988" alt="Screenshot 2022-03-03 at 09 28 41" src="https://user-images.githubusercontent.com/226720/156526402-15a1c4ab-b59b-4224-838f-446d37f04293.png">

<img width="1097" alt="Screenshot 2022-03-03 at 09 29 11" src="https://user-images.githubusercontent.com/226720/156526414-f0cfe470-e314-4e4a-ac55-2f3bd1e8f397.png">

<img width="978" alt="Screenshot 2022-03-03 at 09 25 23" src="https://user-images.githubusercontent.com/226720/156526438-2a2b8c03-5afb-495d-9386-b41e83ce261e.png">

mlucool:
[...]

PR #12281, opened 2022-03-28
Customize the file browser toolbar via the settings

fcollonval:
## References

Fixes #12196

## Code changes

- New settings for `@jupyterlab/filebrowser-extension:widget` to define the browser toolbar definition
- Remove plugin to isolate create launcher button
- Use `SidePanel` as base widget for the filebrowser - it will ease adding widget to it. But use a `Panel` as default content widget and not a `AccordionPanel` to avoid displaying the file browser title
- Add a wrapper widget around the filter, the crumbs and the listing
- Add a `data-command=<id>` to toolbar button to allow CSS selection
- Create a new command refresh file browser
- Make create new launcher command homogeneous with the removed add launcher button plugin

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

> The widgets hierarchy has changed due to the introduction of a wrapper around the filter, the crumb and the listing.

PR #12297, opened 2022-03-30
Add a preferred-dir icon to the file browser crumbs

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12098

## Code changes

Filebrowser crumbs loads an icon and displays it.

## User-facing changes

Add a clickable to the preferred-dir in the file browser crumbs.

Untitled4

None


echarles:
@fcollonval I just read the note of yesterday meeting `Add note to the documentation for core contribution acknowledging that is should be in ui-components.` I can move the new icon of this PR in the ui-components. There are subfolders there. e.g. I remember I added on subfolder for the `listings` but not sure if it was a good idea. My question (and also for the doc to be written): should be icons be names on their specific usage (here : preferred-dir`) or with a more generic names (here: `slash`)?

fcollonval:
> I can move the new icon of this PR in the ui-components. 

:+1: 

> There are subfolders there. e.g. I remember I added on subfolder for the `listings` but not sure if it was a good idea. My question (and also for the doc to be written): should be icons be names on their specific usage (here : `preferred-dir`) or with a more generic name (here: `slash`)?

I would use generic name. So the name won't feel weird when reusing it. And it seems aligned with the current status as the folder structure is actually not reflected when exported the icons (aka the list is flatten).

mlucool:
The UX here feels a bit backwards. We are adding this icon so that you can get to `/`. I think when you set a preferred directory, the folder mean "go home" which is your preferred directory and clicking on `/` mean go to `/`. Can we flip these?

fcollonval:
pinging @isabela-pf for her input on this one.

echarles:
My last push ensures that no ui change is done if no preffered_dir is given.

I just had a discussion with @fcollonval on this.

> Can we flip these?

As the `preferred_dir` has to be a subfolder of `root_dir`, it sounds like the icon for `root_dir` should be first, followed by the icon for `preferred_dir`.

> the folder mean "go home" which is your preferred directory 
> pinging @isabela-pf for her input on this one.

While discussing with @fcollonval  we may need to revisit the icons, we were thinking that the `preferrred` notion ws close from the `home`, leaning towards the usage of a home for that icon. Here are a few options from the top of my head

```
root_dir preferred_dir ... / ...
📁 🏠 .. /...
```

```
root_dir preferred_dir ... / ...
📁 ♥️ .. /...
```

```
root_dir preferred_dir / ... / ...
📁 / .. /...
```

```
root_dir preferred_dir ... / ...
📁 / / .. /...
```

mlucool:
> As the preferred_dir has to be a subfolder of root_dir, it sounds like the icon for root_dir should be first, followed by the icon for preferred_dir.

I don't fully agree with that. Our general idea with creating root_dir + preferred_dir was to allow jupyter to access all files (i.e. start from `/`) but don't break user experience of opening to a logical location to the user (e.g. the directory they started juptyer in).

The problem in https://github.com/jupyterlab/jupyterlab/issues/12098 that @fcollonval rightfully pointed out was that making the folder icon preferred_dir meant that you could not ever get to root_dir (e.g. `/`). I love that in this PR we have an icon for `/` so now the user can navigate to the root_dir intuitively and preferred_dir feels like the old root_dir. Personally, I like the home icon best and would do this ideally (both the home and first / are clickable):
```
preferred_dir root_dir.. / ...
🏠 /.. /...
```

No preferred_dir set (both the home and first / are clickable and go to the same place):
```
root_dir root_dir.. / ...
🏠 /.. /...
```
Or maybe (no home icon, first / is always clickable - or maybe we make all / clickable?):
```
root_dir.. / ...
/.. /...
```

isabela-pf:
Apologies for the late review (since I was pinged on the original issue a while ago). I do think having options to switch to the preferred or root directory can work together, and I am okay with having them back to back at the top of the file browser. I do not think the `/` is a good choice, though, because
- None of the other `/`s will be interactive (and I'm not sure what they would do if they were).
- `/` doesn't describe the interaction at all. Even with a tool tip, I don't think we should use icons that don't describe what they do.

I agree with @mlucool that using a home icon would be clearer. [Material Design's home icon](https://fonts.google.com/icons?selected=Material%20Icons%3Ahome%3A) seems like a good choice since I don't think we're using it anywhere else yet. 

I would prefer a more descriptive icon for the preferred directory since the generic folder icon seems vague when it's no longer the root. Maybe [`folder special`](https://fonts.google.com/icons?selected=Material%20Icons%3Afolder_special%3A)?

echarles:
Thx for your inputs. Based on the discussions, here are 2 options.

Screenshot 2022-04-02 at 16 51 31

Screenshot 2022-04-02 at 16 54 00

echarles:
I have added the needed colors to the icons. My favorite one is the home icon.

Screenshot 2022-04-02 at 16 59 41
.

echarles:
The UI Tests for the visual regression are failing, note sure why. The diff does not seem to be directly related to these chagnes.

echarles:
Trying to understand why the ui regression tests are failing, I run locally the tests (first time doing that) and get a lot of failures on my local env

```
  ✓  [galata] › test/galata/benchmarkReporter.spec.ts:84:3 › BenchmarkReporter › should generat
  ✓  [galata] › test/galata/jupyterlabpage.spec.ts:11:3 › appPath › should have non-default app
  ✘  [galata] › test/galata/notebook.spec.ts:11:3 › Notebook Tests › Create New Notebook (24s)
  ✓  [galata] › test/galata/contents.spec.ts:11:3 › Contents API Tests › Upload directory to server (26s)
  ✓  [galata] › test/galata/fixture.spec.ts:8:3 › page › should return a JupyterLabPage (22s)
  ✘  [jupyterlab] › test/jupyterlab/completer.test.ts:16:5 › Completer › Notebook › Open completer on notebook (24s)
  ✓  [galata] › test/galata/test.spec.ts:8:1 › should display the launcher (23s)
  ✘  [jupyterlab] › test/jupyterlab/collapsible-headings.test.ts:25:3 › Collapsible Headings; showHCB › Show Collapser Unselected;
```

I have rerun a few times the regression tests on a few PR and I still need to check if the different runs returns the same failing tests.

The diff taken from the report show some difference on e.g. the launcher button, not sure how these changes can impact that. Also not all failing tests report some diff image.

I could ask to the bot `please update galata snapshots` as documented on https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests,  but we first need to understand the cause of those diffs.

fcollonval:
Looking at the test errors, the change in the breadcrumb bar is breaking some galata helpers. In particular that one:

https://github.com/jupyterlab/jupyterlab/blob/be752503f1f9fce88d02da944d58ef0634d8c484/galata/src/helpers/filebrowser.ts#L125

The DOM structure being different the test function checking the home directory is the current opened one is not working any more.

> the launcher button

I have seen the filebrowser toolbar icons not loaded in tests since https://github.com/jupyterlab/jupyterlab/pull/12281. But I have not been able to find the reason the icon command is not resolved.

> not all failing tests report some diff image.

You get the output depending on the test failure position. In particular you will get the diff if it fails at the image comparison AND if the image have the same size.


Note about local testing - some tests are expected to fail locally (in particular the ones related to the terminal) - as some parts of JupyterLab are deeply connected to the underlying OS configuration.

echarles:
Visual regression tests are now ok (documentation ui fails, which is expected for now). Usage fails, but should note be related to this PR.

Ready for review/merge.

fcollonval (changes requested):

packages/filebrowser/src/crumbs.ts
```diff
@@ -25,7 +29,12 @@ const BREADCRUMB_CLASS = 'jp-BreadCrumbs';
 /**
  * The class name for the breadcrumbs home node
  */
-const BREADCRUMB_HOME_CLASS = 'jp-BreadCrumbs-home';
+const BREADCRUMB_ROOT_CLASS = 'jp-BreadCrumbs-root';
```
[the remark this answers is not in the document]
echarles: I have changed that name as it was confusing to me to assign a `home` name to the icon which is not a home icon, while having a home icon (for the preferred dir) which did not named `home`.

Any solution has drawbacks. I am happy to revert (including the test) the name. WDYT?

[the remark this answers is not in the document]
echarles: @fcollonval I have pushed the change you have proposed to avoid breaking changes. This will also allow to backport to 3.3.x

echarles:
@fcollonval `All checks have passed`  This is good to go. Thx for all the inputs and follow-ups on this 👍

fcollonval (approved):
Thanks @echarles

echarles:
Merged. Thx @fcollonval

echarles:
I guess the boxplot need to be taken into account. @fcollonval may have more to say on that based on the work he has done.

For the changes of this PR, I do not see how it could impact the performance.

fcollonval:
The white line is the median and the limits of the boxplot limits are the first and third quartiles. The gray lines display the extension from min to max.

Compare with usual results, those are inline.

echarles:
From which snapshot is the expected valued defined (a few months ago?).
What is the habit to update the expected values (are they freezed or subject to update?).

fcollonval:
> From which snapshot is the expected valued defined (a few months ago?).
> What is the habit to update the expected values (are they freezed or subject to update?).

Not sure to get the questions.

Due to uncontrolled hardware, the expected values are not stored. For every benchmark report, both the actual and the expected results are computed on the same GitHub job. So those results are always meaningful for comparison analysis.

echarles:
> Due to uncontrolled hardware, the expected values are not stored. For every benchmark report, both the actual and the expected results are computed on the same GitHub job. So those results are always meaningful for comparison analysis.

I was thinking the expected was hardcoded. Computing the expected and actual values is good to see a single PR. I can also see advantages comparing against a previous version, in which case we cover more general changes of many PRs.

packages/filebrowser/src/crumbs.ts
```diff
@@ -64,6 +73,11 @@ export class BreadCrumbs extends Widget {
     this.addClass(BREADCRUMB_CLASS);
     this._crumbs = Private.createCrumbs();
     this._crumbSeps = Private.createCrumbSeparators();
+    this._hasPreferred =
+      PageConfig.getOption('preferredPath') !== '/' ? true : false;
```
jtpio, line 77: Shouldn't this compared to an empty string, in case `preferredPath` is not set?

jtpio: Otherwise it seems to be assuming that `preferredPath` should be part of the page config.

jtpio: This seems to be where the default `/` is set, for applications using the `LabHandler`:

https://github.com/jupyterlab/jupyterlab_server/blob/93703e2242788c6b8083b351e25ec8113930420d/jupyterlab_server/handlers.py#L89-L99

Issue #12298, opened 2022-03-30
New SettingsUI not updating when setting changed from Menu or Command Palette


matthewgasbarro:
I have a boolean setting that is added to a new file menu as well as the command pallette. When I change the setting on either menu or command pallette, the value is changed correctly in both places and in the JSON settings editor but is not updated in the new settings UI. Is there a notify event that I have not registered to get the setting updated in the new UI?

`const sqSettingsPlugin: JupyterFrontEndPlugin<void> = {
  id: Plugins.sqSettings,
  autoStart: true,
  requires: [ISettingRegistry, ICommandPalette],
  activate: (app: JupyterFrontEnd, settingRegistry: ISettingRegistry, commandPalette: ICommandPalette) => {
    console.log(`extension ${Plugins.sqSettings} is activated!`);
    const { commands } = app;
    let labView = false;

    /**
     * Load the settings for this extension
     *
     * @param settings Extension settings
     */
    function onSettingsUpdated(settings: ISettingRegistry.ISettings): void {
      // Get current labView setting
      const currentLabView = settings.get('labView').composite as boolean | null;
      labView = currentLabView === true || currentLabView === false ? currentLabView : false;
      app.commands.notifyCommandChanged(CommandIDs.toggleLabView);

      console.log(`extension ${Plugins.sqSettings}: labView is '${labView}'`);
    }

    // Wait for the application to be restored and
    // for the settings for this plugin to be loaded
    Promise.all([app.restored, settingRegistry.load(Plugins.sqSettings)])
      .then(([, settings]) => {

        // add the command
        commands.addCommand(CommandIDs.toggleLabView, {
          label: 'JupyterLab Lab View as default',
          caption: 'Enable JupyterLab Lab View as default view"',
          isToggled: () => labView,
          execute: () => {
            // Programmatically change a setting
            Promise.all([
                settings.set('labView', !labView),
              ])
              .then(() => {
                const newLabView = settings.get('labView').composite as boolean;
                console.log(`extension ${Plugins.sqSettings}: labView set to '${newLabView}'`);
              })
              .catch((reason) => {
                console.error(`extension ${Plugins.sqSettings}: error setting labView: ${reason}`);
              });
          },
        });

        // Add the command to the command palette
        commandPalette.addItem({ command: CommandIDs.toggleLabView, category: 'Settings' });

        // Read the settings
        onSettingsUpdated(settings);

        // Listen for plugin setting changes using Signal
[...]

Issue #3046, opened 2017-10-03
File paths containing bidi (RTL) characters are displayed in a wrong sequence in Console


edithkk:
For example, when writing a path containing a parent folder and a sub  folder, the expected display is:
C:/ parent folder/sub folder

When folder names contain bidi characters the display is not as expected.
Parent folder: 		אבג 
Sub folder:		דהו

Actual display in Console is:
C:/אבג/דהו

Expected display is:
C:/דהו/אבג

![image](https://user-images.githubusercontent.com/8215501/31127332-4406b206-a857-11e7-88d8-1e8d121792a8.png)

tomerm:
The correct display would be:

blink1073:
Hi @edithkk and @tomerm, thanks for starting the conversation!  I'd prefer to take the route used in the [classic notebook](https://github.com/jupyter/notebook/pull/2357/files): create a bidi package, and use it where appropriate in the other packages, given the low-level overrides that bidi requires.  I think the actual implementation should wait until after upcoming refactor of our document models, tracked by https://github.com/jupyterlab/jupyterlab/issues/2540.

edithkk:
Hi @blink1073 , thanks for your answer. I'm afraid I don’t quite understand what you mean.
In the code sample in #3047 I already created a bidi package.
The architecture in classic notebook is different so I'm having a hard time understanding where bidi code should be in JupyterLab. Should it not be under packages? Or perhaps it should be used differently?
I would appreciate if you could clarify this for me.
Thanks

blink1073:
Hi @edithkk, I meant that I think we should treat `bidi` as a core feature rather than as an extension, perhaps even in `apputils`.  Then we could write code like the the following:

```typescript
import {
  Bidi
} from '@jupyterlab/apputils';

....
let filePath = Bidi.filePath(...)
```

edithkk:
@blink1073, Thank you, I will move forward with your suggested change.

edithkk:
Hello @blink1073,
I moved bidi code to a package under apputils. I uploaded the code in ticket #3316 for review.
Would this be the acceptable approach?
Are there any other issues I should address before I move on to testing?
(I noticed that the blocking ticket #2540 is still open but I would like to make as much progress as I can while waiting for the refactoring work do be done).
Thanks

jasongrout:
See also #1163, which is a much more comprehensive issue for all of jlab.

tomerm:
[...]

Issue #1163, opened 2016-10-26
Support right-to-left mirroring of UI for rtl languages


jasongrout:
See https://github.com/jupyter/notebook/issues/1852 for many helpful suggestions.

AliaNaguib:
Hi, this is Alia Naguib from IBM. Here is a proposal for the mirroring support. 
[JupyterLab Mirroring Design.docx](https://github.com/jupyterlab/jupyterlab/files/668674/JupyterLab.Mirroring.Design.docx)

DearRude:
Any updates on this?

PR #468, opened 2016-07-19
File browser interaction cleanup

blink1073:
Fixes #441.

Context Menus:
- Right clicking on an item selects it before opening the context menu
- Right clicking on an unselected item when a context menu is open selects a new item and opens the context menu
- A left click anywhere when there is a context menu clears the menu but keeps the selection

Soft Selection:
- Unmodified mouse down an item in an existing selection is a "soft selection" for dragging or context menu
- Mouse up on a soft selection clears other selection unless there is an active context menu

Drag-drop:
- Fixes a bug when dragging multiple items - the original node text was being changed
- Uses a file icon for all multi-item drags

There is overall less spurious triggering of file name editing as well.

fix-file-selection

PR #12299, opened 2022-03-30
Add checkboxes to file browser

gabalafou:
## References

Partially addresses https://github.com/jupyterlab/retrolab/issues/260, which
seeks to minimize UI differences between classic Notebook (which has file
checkboxes) and the upcoming Notebook v7 (which currently does not).

## Code changes

- Style changes in `filebrowser` CSS to make space for the checkbox and handle
  visibility
- Changes to `DirListing` class and default renderer to create checkboxes,
  handle clicks on them, and update state
- New test file, `listing.spec.ts`, to test checkbox functionality
- Fix for `ServiceManagerMock` so that when you ask for the contents of a
  directory you don't get the directory itself within that list

## User-facing changes

- Similar to the Windows file browser, this adds checkboxes that allows the user
  to select and unselect files by clicking the checkboxes. This provides another
  method to select multiple files, which should make it easier for some users.
- A "check all" checkbox.
- Some whitespace changes to help make room for the checkboxes 
- An option to turn off (or on) the file checkboxes. In the UI, this can be
  modified via the settings editor and the context menu that opens on right
  click on the header above the files (the header with sortable columns "Name"
  and "Last Modified").

Light and dark theme screenshots:

JupyterLab light theme, file browser with checkboxes

JupyterLab dark theme, file browser with checkboxes

Context menu with option to turn off checkboxes:

Context menu with option to turn off checkboxes

I'm not sure that there's a way to guarantee that this is 100%
backwards-compatible because it modifies two base classes, `DirListing` and
`DirListing.Renderer` that downstream code can (and does) inherit from and
extend. However, I made efforts to minimize the chances that these code changes
will break downstream code. (I checked it against `jupyterlab-unfold`.)
Furthermore, in the case of some UI incompatibility, extensions should be able
to turn off the file checkboxes. To that end, I made every effort to avoid
touching existing styles and only added new style rules that apply when the file
checkboxes are added to the DOM.


packages/filebrowser/style/base.css
```diff
@@ -215,13 +215,36 @@
...
+  padding: 4px;
+}
+
+.jp-DirListing-checkboxWrapper.jp-mod-visible {
+  visibility: visible;
+}
+
+/* For devices that support hovering, hide checkboxes until hovered, selected...
+*/
+@media (hover: hover) {
```
gabalafou, line 237: I tested that this works on the following:

- iPad -> checkboxes always visible
- MacBook -> checkboxes visible on hover
- Windows laptop -> checkboxes visible on hover
- Windows laptop in tablet mode -> checkboxes always visible

packages/filebrowser/style/base.css
```diff
@@ -145,7 +145,7 @@
 }
 
 .jp-DirListing-headerItem {
-  padding: 4px 12px 2px;
+  padding: 5px 12px 2px 6px;
```
gabalafou, line 148: These were the pixel-by-pixel adjustments that looked best to my eye

packages/filebrowser/style/base.css
```diff
@@ -154,11 +154,11 @@
 }
 
 .jp-DirListing-headerItem.jp-id-name {
-  flex: 1 0 84px;
+  flex: 1 0 94px;
```
gabalafou, line 157: +10px

packages/filebrowser/style/base.css
```diff
@@ -154,11 +154,11 @@
...
 }
 
 .jp-DirListing-headerItem.jp-id-modified {
-  flex: 0 0 112px;
+  flex: 0 0 102px;
```
gabalafou, line 161: -10px

packages/filebrowser/style/base.css
```diff
@@ -215,13 +215,36 @@
 .jp-DirListing-item {
   display: flex;
   flex-direction: row;
-  padding: 4px 12px;
+  align-items: center;
+  padding-right: 12px;
```
gabalafou, line 219: Is there a right-to-left mode for JupyterLab that I should worry about?

packages/filebrowser/style/base.css
```diff
@@ -241,19 +264,19 @@
 
 .jp-DirListing-itemIcon {
   flex: 0 0 20px;
-  margin-right: 4px;
+  margin: 0 4px;
 }
 
 .jp-DirListing-itemText {
-  flex: 1 0 64px;
+  flex: 1 0 74px;
```
gabalafou, line 271: +10px

packages/filebrowser/style/base.css
```diff
@@ -241,19 +264,19 @@
...
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
   user-select: none;
 }
 
 .jp-DirListing-itemModified {
-  flex: 0 0 125px;
+  flex: 0 0 115px;
```
gabalafou, line 279: -10px

packages/filebrowser/src/listing.ts
```diff
@@ -1824,6 +1905,16 @@ export namespace DirListing {
...
 
+    /**
+     * Get the checkbox input element node.
+     *
+     * @param node A node created by [[createItemNode]] or
+     * [[createHeaderItemNode]]
+     *
+     * @returns The checkbox node.
+     */
+    getCheckboxNode?: (node: HTMLElement) => HTMLInputElement;
```
gabalafou, line 1916: I wasn't sure if I should modify this interface (`DirListing.IRenderer`) in a backwards-incompatible way, so I made this method optional. I was also careful in the `DirListing` methods to not assume that the instance's renderer comes with checkboxes, which is why you see a lot of the code wrapped in `if (renderer.getCheckboxNode)`.

gabalafou (commented):
self review

aiqc:
Does `shift+click` and `ctrl+click` (cmd+click on mac) aka multi-select apply the checkboxes?

jtpio:
Many thanks @gabalafou for starting this, it looks great!

> Does `shift+click` and `ctrl+click` (cmd+click on mac) aka multi-select check the checkboxes?

Just checked on Binder: https://mybinder.org/v2/gh/gabalafou/jupyterlab/file-checkbox?urlpath=lab-dev

And it does not seem to be the case yet. But could be tracked in a separate issue.

jtpio:
> And it does not seem to be the case yet. But could be tracked in a separate issue.

Although these keyboard shortcuts already work when clicking on the file and directories:

https://user-images.githubusercontent.com/591645/160888421-9fadb1be-1e80-4e98-a642-88f6669b3dfb.mp4

krassowski:
Can we make it toggle-able as is the "Last Modified" column?

packages/filebrowser/style/base.css
isabela-pf: Right now, I believe the answer is no. Long term this is something we'd like to support (and have had [RTL support requested via issues](https://github.com/jupyterlab/jupyterlab/issues/1163)). (And [further discussion on this issue](https://github.com/jupyterlab/jupyterlab/issues/3046).)

gabalafou: Thanks for that context!

gabalafou (commented):
Good observation re: shift-click!

So I'm not entirely sure how this should be handled. 

In the Windows file browser, which has checkboxes, the shift modifier is ignored when the click happens on top of the checkbox, as opposed to on the file name. It doesn't select multiple files. It behaves like a normal click.

I can understand the desire for a shift-click on the checkbox to behave just like a shift-click anywhere else in the row; however, I think it becomes unintuitive when you have a block of files already selected and then you shift-click on a selected checkbox and nothing happens. My intuition is that if I shift-click on a checked box, it should uncheck. The only time I think it makes sense to ignore the click on the checkbox is when it's a right-click on an item that is already selected. But perhaps I've just gotten used to the Windows behavior.

I've summarized my proposed behavior in the following table:

| initial checkbox state | after left-click | after right-click |
| --------------------- | -------------- | --------------- |
| unchecked | checked | checked | 
| checked | unchecked | checked (no change) |

(Apologies for the terminology if you use a left-handed mouse!) 

You might wonder why the asymmetry: why not ignore all right-clicks, whether or not the file is selected? That's because a change was made explicitly to support [selecting the file on right-click](https://github.com/jupyterlab/jupyterlab/pull/468). This makes sense because the context menu that opens up on right-click operates on the selected file. If you don't select a file on right-click, then the action you take in the context menu (such as opening the file) would operate on the file you last selected and not the file you just right-clicked, which is counter-intuitive. Combine that with a principle that the checked state of the checkbox should strictly mirror the selected state of the item, and then the right-click asymmetry is unavoidable.

gabalafou:
@krassowski I think we could make it togglable. Out of curiosity, may I ask what kind of use cases you have in mind for that?

gabalafou:
@krassowski does the fact that the checkboxes only appear on hover have any bearing on the rationale for making the entire column of checkboxes togglable?

krassowski:
> @krassowski does the fact that the checkboxes only appear on hover have any bearing on the rationale for hiding the column?

No. This is a new addition which mostly benefits Notebook, but much less existing JupyterLab users (multi-select is already implemented, it does not add a new functionality, just a new way to interface with existing one) and has a non-zero cost to the real estate:

- this PR add a new potentially distracting element (the checkbox on top); users often ask for more focused interface with less bells and whistles
- checkboxes even when not visible take up precious space which could allow to distinguish filenames which have the same suffix or display more of the notebook(s)

Also, we should test if this will interfere with existing extensions like jupyterlab-unfold (https://pypi.org/project/jupyterlab-unfold/)

gabalafou:
Thanks @krassowski! Super valuable feedback. I tested my code changes against `jupyterlab-unfold` today. It breaks it.

I need to fix that but I'm not sure how. I may need an hour of pair-programming time with someone to figure out how to implement this in a way that doesn't break downstream code like `jupyterlab-unfold`. This is something I don't have a lot of experience with.

In the meantime, until I figure out a way forward, I'm going to put this PR in draft mode.

gabalafou (commented):
Since last update:

- Feature now lives behind (new) `showFileCheckboxes` setting, yay!
- Doesn't break `jupyterlab-unfold` extension

I need to check the latest updates on Windows, iPad, etc, but I'm still going to take this PR out of draft mode for now

gabalafou:
Looks good on iPad:

screenshot iPad

Looks good on Windows desktop:

screenshot Windows desktop

packages/filebrowser-extension/schema/browser.json
```diff
@@ -126,6 +126,11 @@
         "selector": ".jp-DirListing-header",
         "rank": 14
       },
+      {
+        "command": "filebrowser:toggle-file-checkboxes",
+        "selector": ".jp-DirListing-header",
+        "rank": 14.5
```
jtpio, line 132: Wondering whether we should keep integers here.

Even though decimal values are supported, integers are probably more readable and easier to reason about? And we could increase the ranks of the items below.

packages/filebrowser-extension/schema/browser.json
```diff
@@ -227,6 +232,12 @@
       "title": "Show hidden files",
       "description": "Whether to show hidden files. The server parameter `ContentsManager.allow_hidden` must be set to `True` to display hidden files.",
       "default": false
+    },
+    "showFileCheckboxes": {
+      "type": "boolean",
+      "title": "Use checkboxes to select items",
+      "description": "Whether to show checkboxes next to files and folders",
+      "default": true
```
jtpio, line 240: Should we make this an opt-in?

```suggestion
      "default": false
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -1172,6 +1175,22 @@ function addCommands(
...
 
+  commands.addCommand(CommandIDs.toggleFileCheckboxes, {
+    label: trans.__('Show File Checkboxes'),
+    isToggled: () => browser.showFileCheckboxes,
+    execute: () => {
+      const value = !browser.showFileCheckboxes;
+      const key = 'showFileCheckboxes';
+      if (settingRegistry) {
+        return settingRegistry
+          .set('@jupyterlab/filebrowser-extension:browser', key, value)
```
jtpio, line 1186: Optional for this PR: we could take the opportunity to hoist `@jupyterlab/filebrowser-extension:browser` in a `const FILE_BROWSER_PLUGIN_ID` variable at the module level, since this seems to be used in several places now.

packages/filebrowser/src/browser.ts
```diff
@@ -404,6 +420,7 @@ export class FileBrowser extends SidePanel {
   private _showLastModifiedColumn: boolean = true;
   private _useFuzzyFilter: boolean = true;
   private _showHiddenFiles: boolean = false;
+  private _showFileCheckboxes: boolean = true;
```
jtpio, line 423: Similar to above: should this be opt-in or opt-out?

```suggestion
  private _showFileCheckboxes: boolean = false;
```

packages/filebrowser/src/listing.ts
```diff
@@ -1824,6 +1909,16 @@ export namespace DirListing {
...
 
+    /**
+     * Get the checkbox input element node.
+     *
+     * @param node A node created by [[createItemNode]] or
+     * [[createHeaderItemNode]]
+     *
+     * @returns The checkbox node.
+     */
+    getCheckboxNode?: (node: HTMLElement) => HTMLInputElement | void;
```
jtpio, line 1920: Since this will mostly be useful for Notebook v7, we could also only target JupyterLab 4.0.

It means this method does not have to be optional, which can make the code more readable in other places.

And we can add it to the extension migration guide: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#api-breaking-changes

jtpio: That would mean:

```suggestion
    getCheckboxNode: (node: HTMLElement) => HTMLInputElement | void;
```

packages/filebrowser/src/listing.ts
```diff
@@ -2329,4 +2495,12 @@ namespace Private {
       container.className = HEADER_ITEM_ICON_CLASS;
     }
   }
+
+  /**
+   * An event handler that just prevents the event's default action from
+   * happening (such as checking a checkbox after a click).
+   */
+  export function preventDefault(event: Event): void {
```
jtpio, line 2503: Do we need this extra function? Looks like it is used in one place only for now?

packages/filebrowser/style/base.css
jtpio: Thanks for the extensive testing!

packages/filebrowser/test/listing.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+          left: 0,
+          right: 10,
+          top: 0,
+          bottom: 10
+        });
+        const checkbox = dirListing.renderer.getCheckboxNode!(
+          itemNode
+        ) as HTMLInputElement;
+        const item = dirListing.sortedItems().next()!;
+        dirListing.selectItemByName(item.name);
```
jtpio, line 186: Missing `await` here?

```suggestion
        await dirListing.selectItemByName(item.name);
```

jtpio (commented):
Thanks, this is looking good!

Left a couple of inline comments.

jtpio:
Looks like the UI tests are failing because of the new column.

But they should pass if we make this feature opt-in instead.

packages/filebrowser-extension/src/index.ts
```diff
@@ -220,7 +222,8 @@ const browser: JupyterFrontEndPlugin<void> = {
               navigateToCurrentDirectory: false,
               showLastModifiedColumn: true,
               useFuzzyFilter: true,
-              showHiddenFiles: false
+              showHiddenFiles: false,
+              showFileCheckboxes: true
```
gabalafou, line 226:
```suggestion
              showFileCheckboxes: false
```

packages/filebrowser-extension/schema/browser.json
gabalafou: Since I have heard that JupyterLab settings involve a lot of merging settings from different places, I was afraid to touch existing rank values for fear that it might move the menu item in some downstream configuration of the UI.

But I went ahead and used your suggestion.

Please double check

packages/filebrowser/src/listing.ts
gabalafou: No it's not needed. Probably did it out of old habit, not wanting to create a multiple function instances when the event callbacks could all point to the same function in memory.

packages/filebrowser/src/listing.ts
gabalafou: I committed your suggestion but one thing to think about: the jupyterlab-unfold extension, which renders the directory as a tree. I'm not sure it will ever make sense for that extension, which implements the `DirListing.IRenderer` interface, to implement a `getCheckboxNode` function.

I will wait on your reply here before I update the parts of the PR that do `this.getCheckboxNode?.(node)`.

gabalafou:
Thanks @jtpio! I implemented almost all of your suggestions. There's just a question about `getCheckboxNode` being optional or not

packages/filebrowser/src/listing.ts
gabalafou: I'm guessing that's probably a premature optimization these days

gabalafou:
I think the UI tests are still failing because of the CSS `align-items: center` rule that I added. This is the one CSS change that doesn't toggle with the `showFileCheckboxes` setting, so the snapshots still fail even now that the feature is off by default.

That said, I think `align-items: center` is perhaps a rule that should been there all along

gabalafou:
For example:

JupyterLab screenshot diff

jtpio:
> That said, I think align-items: center is perhaps a rule that should been there all along

Sounds good. Then let's update the reference snapshots?

packages/filebrowser-extension/schema/browser.json
jtpio: Good point, this might indeed affect some extensions that add new menu items.

But since this change is targeting 4.0, it sounds fair to expect slight changes in the menu. In that case it's still minor since it's bumped by 1.

packages/filebrowser/src/listing.ts
jtpio: Another alternative could be:

```ts
getCheckboxNode: (node: HTMLElement) => HTMLInputElement | null;
```

So downstream implementation can return `null` if they don't implement that feature.

gabalafou (commented):
Self-reviewed again. Looks good to me

packages/filebrowser/src/listing.ts
```diff
@@ -2114,6 +2272,21 @@ export namespace DirListing {
...
+     * @param node A node created by [[createItemNode]] or
+     * [[createHeaderItemNode]]
+     *
+     * @returns The checkbox node.
+     */
+    getCheckboxNode(node: HTMLElement): HTMLInputElement | null {
+      const checkbox = node.querySelector(
+        `.${CHECKBOX_WRAPPER_CLASS} input[type=checkbox]`
+      ) as HTMLInputElement | void;
+      return checkbox || null;
```
jtpio, line 2287: Minor styling we could use for consistency:

```suggestion
      const checkbox = node.querySelector(
        `.${CHECKBOX_WRAPPER_CLASS} input[type=checkbox]`
      ) as HTMLInputElement | null;
      return checkbox ?? null;
```

jtpio (commented):
Thanks this is looking good.

Left a minor suggestion, and also tried locally.

It appeared that changing the setting via the settings editor didn't have any effect. It would be nice if changing the setting would update the file browser without having to refresh the page, like when using the context menu.

https://user-images.githubusercontent.com/591645/164224864-d11ea5bb-3a20-4f73-ba87-e8e520dd256c.mp4

Happy to double check on Binder when it's available, just to make sure it's an issue with the local environment.

gabalafou:
Thank again @jtpio! 

Your screencast seems to show several distinct bugs:

1. The checkboxes UI in the file browser does not update when you toggle the corresponding setting in the Settings Editor
2. The setting does not persist after a page refresh
3. Turning the checkbox UI on/off via the right-click context menu updates the file browser UI but does not update the state of the setting in the Settings Editor

I am unable to reproduce 1 and 2 either locally or on Binder.

However, I can reproduce 3. This seems like a [bug in the Settings Editor](https://github.com/jupyterlab/jupyterlab/issues/12298), and there's at least one issue in GitHub on it. I would be happy to work on fixing it. Do I just assign the issue to myself in GitHub and then go for it?

packages/filebrowser/src/listing.ts
gabalafou: Oh right! querySelector's return signature is actually [Element | null](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector).

I fixed it, thanks for pointing it out

gabalafou:
@jtpio I also built and deployed my branch to a server, so you can verify the behavior there instead of, or in addition to Binder:

http://gabalafou.com:9000/lab
password hints: starts with an i, was predecessor to jupyter, but all in lowercase this time

jtpio:
> However, I can reproduce 3. This seems like a [bug in the Settings Editor](https://github.com/jupyterlab/jupyterlab/issues/12298), and there's at least one issue in GitHub on it. I would be happy to work on fixing it. Do I just assign the issue to myself in GitHub and then go for it?

OK thanks for checking :+1: 

I think we can indeed continue testing this once released in a 4.0 pre-release, and track follow-up work in new issues. So we don't block this PR and can already start using in the Notebook v7 pre-releases.

jtpio (approved):
Thanks!

PR #12320, opened 2022-03-31
Pin exact version of y-py

davidbrochart:
## References

[y-py](https://github.com/y-crdt/ypy) is still in development, and we should expect the API to break at every release until we reach v1.0.0. Next release won't be backward-compatible for sure.

## Code changes

Pin exact version of y-py until it reaches v1.0.0.

## User-facing changes

None.

None.


davidbrochart:
Closing in favor of #12326.

PR #12326, opened 2022-04-01
Use y-py 0.3.0

davidbrochart:
## References

Supersedes #12320.

## Code changes

Require `y-py==0.3.0` and use its new API.

## User-facing changes

None.

None.

### DIFF
--- a/jupyterlab/handlers/ydoc.py
+++ b/jupyterlab/handlers/ydoc.py
@@ -21,8 +21,7 @@ class YFile(YBaseDoc):
 
     @property
     def source(self):
-        with self._ydoc.begin_transaction() as t:
-            return self._ysource.to_string(t)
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @davidbrochart

jtpio:
Thanks!

Looks like saving in the latest 4.0 pre-release (4.0.0a23) is currently broken, and this should fix it.

jtpio:
Just restarted the failed CI check.

Expecting one failure for "Visual Regression Documentation", which will be fixed by https://github.com/jupyterlab/jupyterlab/pull/12337.

jtpio:
> Expecting one failure for "Visual Regression Documentation",

Looks like that's the case:

image

So let's get this in, thanks!

PR #12329, opened 2022-04-01
Remove cell id before saving the notebook

hbcarlos:
I think there is a small bug from #11599

Before returning the content of the notebook, the NotebookModel checks the nbformat version and removes the cell id for versions previous 4.5.
https://github.com/jupyterlab/jupyterlab/blob/eeaa3f395cd755e34045611d3d9cc3947626b822/packages/notebook/src/model.ts#L260-L263

Should we do the same check when saving from the backend?

cc @davidbrochart


davidbrochart:
I think you're right, thanks for catching that.

jtpio (approved):
Thanks!

PR #12333, opened 2022-04-03
Allow for an optional label for the uploader widget

jtpio:
## References

Allow passing an optional `label` to the `Uploader` component, which can then be used in downstream distribution like Notebook v7 or third-party extensions.

## Code changes

Add an optional `label` to the `Uploader.IOptions` interface.

## User-facing changes

If provided, the button could then show a label next to the icon: 

image

This is however *not* enabled by default in JupyterLab, which still shows the icon only:

image

None

### DIFF
--- a/packages/filebrowser/src/upload.ts
+++ b/packages/filebrowser/src/upload.ts
@@ -20,6 +20,7 @@ export class Uploader extends ToolbarButton {
   constructor(options: Uploader.IOptions) {
     super({
       icon: fileUploadIcon,
+      label: options.label,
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
LGTM!

> The error in check releaser is not related but sounds legit, isn't it?

jtpio:
> The error in check releaser is not related but sounds legit, isn't it?

Yes this one is tracked in https://github.com/jupyterlab/jupyterlab/issues/12336

Issue #12328, opened 2022-04-01
Visual Regression Documentation CI check is failing


jtpio:
## Description

Similar to the previous time: https://github.com/jupyterlab/jupyterlab/issues/12233

This looks like an issue with items not being sorted in the extension manager:

https://user-images.githubusercontent.com/591645/161276234-31aa75e9-9adb-4ec6-84ff-82759ec2014b.mp4

Thus depending on the order returned from the query to npm.

## Reproduce

Example run: https://github.com/jupyterlab/jupyterlab/runs/5788364548?check_suite_focus=true

## Expected behavior

The tests should pass, and the screenshot should be stable.

## Context

https://github.com/jupyterlab/jupyterlab/issues/12233 and https://github.com/jupyterlab/jupyterlab/pull/12249 for reference.

jtpio:
Probably the fix is to sort the extensions in the extension manager.

jtpio:
> Probably the fix is to sort the extensions in the extension manager.

Looks like this is the related code:

https://github.com/jupyterlab/jupyterlab/blob/eeaa3f395cd755e34045611d3d9cc3947626b822/packages/extensionmanager/src/model.ts#L953-L956

Which depends on the sort from the npm API.

So either we consider this new sort the reference, and we update the snapshots accordingly. Or we ensure our own sorting, for example by name.

echarles:
I also see failure for the ui-test visual regression (not the documentation one, the other), see e.g. my comment on https://github.com/jupyterlab/jupyterlab/pull/12297#issuecomment-1086715618

PR #12337, opened 2022-04-04
Update reference snapshots for the extension manager

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12328

## Code changes

~Update sorting logic in the extension manager.~

Update the reference snapshots for the visual regression documentation checks.

## User-facing changes

~The order of the extensions in the extension manager might change and would not depend on the result from the npm query anymore.~

None

None


jtpio:
This can then be backported to 3.4.x and 3.3.x if needed.

jtpio:
This approach has the downside of discarding the `score` returned by npm:

image

For example if an extension has more downloads it might be more relevant, and it would make sense to keep it higher in the list.

jtpio:
So maybe we can instead just update the reference snapshots for now, and see if the scores change again later.

fcollonval:
Thanks for looking into this @jtpio 

I'm :+1: to update the snapshots for now. A future stabilization would be to dump a npm request result and to serve it as mock answer for the tests.

jtpio:
Just pushed the new snapshots, let's see if the check passes again now.

jtpio:
The check is back to green:

image

> A future stabilization would be to dump a npm request result and to serve it as mock answer for the tests.

Sounds good, we can track that in a separate issue.

fcollonval (approved):
Thanks a lot @jtpio

jtpio:
CI all green :tada:

PR #12339, opened 2022-04-04
Disable `fail-fast` for the check release workflow

jtpio:
## References

This will help prevent failures happening in the `check_release` check to cause the `link_check` job to be cancelled. 

So the link check can still run without being impacted by `check_release`.

## Code changes

CI workflow update.

## User-facing changes

None

None


fcollonval (approved):
Thanks @jtpio

PR #9677, opened 2021-01-26
Move flaky tests to a separate workflow on CI

jtpio:
## References

The long term motivation is to be able to reach a full green CI some time soon.

For now, this change moves some of the known flaky tests to their own workflow, so they can easily be restarted without re-running everything else. The current list includes:

- `js-services`
- `js-apputils` 

image

## Code changes

CI

## User-facing changes

None

## Backwards-incompatible changes

None

PR #12341, opened 2022-04-04
Remove the CI workflow for flaky tests

jtpio:
## References

Remove the extra `linuxjs-flaky-tests` CI workflow, which was introduced in https://github.com/jupyterlab/jupyterlab/pull/9677 to avoid restarting all the checks if only some known flaky ones were failing.

But GitHub now has a built-in way to restart failed job so this workflow is not needed anymore:

image

## Code changes

Remove unneeded CI workflow.

## User-facing changes

None

None


jtpio:
The `js-apputils` and `js-services` checks are now run as "Linux JS Tests`" checks:

image

fcollonval (approved):
LGTM

PR #12344, opened 2022-04-04
Allow LauncherModel to be more extendable

ajbozarth:
## References

No tracking issue

## Code changes

Currently when a 3rd party extension wants to override launcher
they don't have the option of overriding the LauncherModel since
the items list is private and not accessable. This updates the private
variable to be protected so the value can be accessed by extensions of
the model class.

I also had to change the var name since protected vars do not prepend
a _ but there's already a public function called items in the class.

## User-facing changes

none

IIUC since this is surfacing a previously private var this API change is back-compatable

### DIFF
--- a/packages/launcher/src/widget.tsx
+++ b/packages/launcher/src/widget.tsx
@@ -52,11 +52,11 @@ export class LauncherModel extends VDomModel implements ILauncher.IModel {
     // Create a copy of the options to circumvent mutations to the original.
     const item = Private.createItem(options);
 
-    this._items.push(item);
+    this.itemsList.push(item);
     this.stateChanged.emit(void 0);
 
     return new DisposableDelegate(() => {
-      ArrayExt.removeFirstOf(this._items, item);
+      ArrayExt.removeFirstOf(this.itemsList, item);
       this.stateChanged.emit(void 0);
     });
   }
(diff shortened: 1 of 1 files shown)


ajbozarth:
For a bit more context, we add new editors for python and r files, but the new py file and new r file launcher cards cause our users confusion. Currently there's no easy way to remove those tile in a third party extension. We also override the launcher so allowing the editing of the model by 3rd party extensions which replace the launcher would be the easiest way to deal with our issue.

I've tagged @fcollonval and @afshin as the most familiar with my recent work and this chunk of code.

Also if this is approved I would love to grab it in 3.4

packages/launcher/src/widget.tsx
```diff
@@ -68,7 +68,7 @@ export class LauncherModel extends VDomModel implements ILauncher.IModel {
     return new ArrayIterator(this._items);
   }
 
-  private _items: ILauncher.IItemOptions[] = [];
+  protected _items: ILauncher.IItemOptions[] = [];
```
jtpio, line 71: Normally the convention would be to drop the leading `_` when the attributes are not private anymore:

```suggestion
  protected items: ILauncher.IItemOptions[] = [];
```

jtpio: Ah good point, searching the code base for `protected _` returns quite a few results.

> also worth noting that there's already a function called items() in the class

In that case it indeed looks like keeping `_items` would be the safest for now :+1:

jtpio: Nice that change looks good.

jtpio (approved):
Thanks!

jtpio:
cc @fcollonval @afshin who might want to have a look too since they were initially tagged for review

fcollonval (approved):
LGTM

fcollonval:

Issue #12345, opened 2022-04-05
'get' and 'set' accessor must have the same type for showDocsPanel


echarles:
## Description

Error is thrown using the completer package in external libraries

```
node_modules/@jupyterlab/completer/lib/widget.d.ts:56:9 - error TS2380: 'get' and 'set' accessor must have the same type.
56     set showDocsPanel(showDoc: boolean | null);

~~~~~~~~~~~~~


node_modules/@jupyterlab/completer/lib/widget.d.ts:57:9 - error TS2380: 'get' and 'set' accessor must have the same type.


57     get showDocsPanel(): boolean;

~~~~~~~~~~~~~
```

## Reproduce

Follow the contribute steps documented on https://github.com/datalayer/jupyter-react#contribute

PR #12346, opened 2022-04-05
showDocsPanel should receive a boolean

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12345

## Code changes

showDocsPanel should now receives a `boolean` instead of `boolean | null`.

cc/ @krassowski

## User-facing changes

None

None.


trungleduc (approved):

fcollonval (approved):
Thanks @echarles 

Thanks for the review @trungleduc

echarles:
👍 @trungleduc @fcollonval

Issue #12225, opened 2022-03-16
Drop deprecated url-loader and file-loader


fcollonval:
### Problem

From a point raised by @bollwyvl at the weekly meeting, 

> webpack: lab 4.0 should evaluate replacing url-loader and file-loader with assets 

See [documentation](https://webpack.js.org/guides/asset-modules/)

### Proposed Solution

Use new asset loader of webpack 5 for Lab 4.

### Additional context

Xref https://github.com/jupyterlite/jupyterlite/pull/557

fcollonval:
Seems like `raw-loader` is deprecated too.

bollwyvl:
thanks for making this!

Some notes: 
- to my understanding, with the non-inline `asset/*` types, webpack doesn't even bring the bytes in.nto memory, which might help the top-line memory of a build: we have hundreds of JSON, svg, fonts, etc. files.
- `style-loader` and `css-loader` major versions should be bumped to support some of these changes.
- this change should probably _not_ be backported `@jupyterlab/builder@3.x`, as it would be a fairly significant departure from the existing build API surface, and would be hard to intrepret the errors during `jupyter lab build`

fcollonval:
Xref: [Meeting notes](https://github.com/jupyterlab/team-compass/issues/135#issuecomment-1069381431)

blink1073:
This is also related to https://github.com/jupyterlab/jupyterlab/issues/11248

jtpio:
Some work in progress started in https://github.com/jupyterlab/jupyterlab/pull/12350

Issue #11359, opened 2021-10-28
Build failure with Node 17


jweill-aws:
## Description

On macOS using Node v17.0.1, the current version installed by Homebrew, I get an error when I run `jlpm run build`.

## Reproduce

1. Check out source
2. Satisfy `canvas` dependencies with `brew install pkg-config cairo pango libpng jpeg giflib librsvg`
3. Run `pip install -e .` (successfully)
3. Run `jlpm run build`
4. Compilation error as described below in Command Line Output section.

With Node 16.13.0 or Node 14.18.1, as installed by `nvm`, the compilation step completes successfully. Node 17.0.1 as installed by `nvm` demonstrates the error below as well.

## Expected behavior

Compilation completes successfully.

## Context

- Operating System and version: macOS Big Sur 11.6
- Browser and version:N/A
- JupyterLab version: Building from head of source tree (4 alpha)

<details><summary>Troubleshoot Output</summary>
<pre>
N/A
</pre>
</details>

<details><summary>Command Line Output</summary>
Compilation error:
<pre>
yarn run v1.21.1
$ jlpm run build:dev
$ jlpm run integrity && jlpm run build:all && cd dev_mode && jlpm run build
$ node scripts/ensure-buildutils.js && node buildutils/lib/ensure-repo.js
node:internal/errors:464
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/jweill/git/jupyterlab/node_modules/@playwright/test/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:416:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:669:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
 *at requirePackage (/Users/jweill/git/jupyterlab/buildutils/lib/utils.js:312:16) {
*  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v17.0.1
[...]

Issue #11248, opened 2021-10-07
Webpack Hashing Problems


blink1073:
## Description
We have seen a bug in [CI](https://github.com/jupyterlab/jupyterlab/runs/3808333301?check_suite_focus=true#step:10:3450) related to hashing and Webpack that affects the ability to build JupyterLab and source extensions. 

The underlying issue is that for versions of OpenSSL built with [FIPS compliance](https://docs.oracle.com/cd/E36784_01/html/E54953/fips-notok-1.html), the default hashing algorithm used by Webpack and loaders is not available (`md4`).

We started seeing failures recently because the latest version of `openssl` on `conda-forge` was built with [FIPS compliance](https://github.com/conda-forge/openssl-feedstock/pull/72/files#diff-d7075654874cb08007a21aaab3ecd4b3453a9087e7505d034d548b8938b599bcR7).

The solution suggested by the [Webpack developers](https://github.com/webpack/webpack/issues/13572#issuecomment-923742724) is to use `output.hashFunction`.  However, this setting is not respected by [loaders](https://github.com/webpack/loader-utils/blob/d9f4e23cf411d8556f8bac2d3bf05a6e0103b568/lib/getHashDigest.js#L43), causing the build to fail for loaders even with the setting.  Loaders are also deprecated in favor of [asset modules](https://webpack.js.org/guides/asset-modules/).

For now we need to use the hard [workaround](https://github.com/webpack/webpack/issues/13572#issuecomment-923736472) over overriding the built-in `crypto.hashFunction` to prevent the usage of `md4`.  This fix will have to be applied to all of our usages of `webpack` config files, including the one in [`notebook`](https://github.com/jupyter/notebook/blob/76d051eb5ed031f33664c0d1bd1f946bf6cd6542/webpack.config.js).

Longer term, we should migrate away from using loaders in favor of `asset modules`.  It would take some work to handle the loaders configured in our [base config](https://github.com/jupyterlab/jupyterlab/blob/master/builder/src/webpack.config.base.ts), especially the third-party [`svg-url-loader`](https://github.com/bhovhannes/svg-url-loader).

A workaround for `conda` users in general when facing this bug is to install an older version of `openssl`:  `conda install -c conda-forge "openssl<3"`

## Reproduce

1. Create and activate new conda environment
2. Install nodejs: `conda install nodejs=16`
2. Install JupyterLab: `pip install jupyterlab==3.1.17`
[...]

PR #12350, opened 2022-04-05
webpack: switch to the asset modules

jtpio:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/12225
- Fixes https://github.com/jupyterlab/jupyterlab/issues/11248
- Fixes https://github.com/jupyterlab/jupyterlab/issues/11359

Documentation for the webpack 5 Asset Modules: https://webpack.js.org/guides/asset-modules

## Code changes

- [x] Replace the deprecated `url-loader`, `raw-loader` and `file-loader` by the corresponding asset modules: https://webpack.js.org/guides/asset-modules/
- [x] Update `style-loader`, `css-loader`, `svg-url-loader`
- [x] Replace `svg-url-loader`? Or in a separate PR? Considering switching to `mini-svg-data-uri` with a custom `generator`: https://webpack.js.org/guides/asset-modules/#custom-data-uri-generator
- [x] Check with new openssl
- [x] Clean up dependencies
- [x] Update examples
- [x] Update galata
- [x] Update extra webpack config used for themes: 

https://github.com/jupyterlab/jupyterlab/blob/a7ef139ee50ad598a231e90b76a0959b3b4b692d/builder/src/build.ts#L205-L218

As mentioned in https://github.com/jupyterlab/jupyterlab/issues/12225#issuecomment-1069397270, this might also speed up the build and / or reduce memory usage.

## User-facing changes

None

None


jtpio:
OK so this also seems to be building fine locally:

- Node 17.8.0
- openssl 3.0.2 installed with conda: https://anaconda.org/conda-forge/openssl
- without the md4 workaround: https://github.com/jupyterlab/jupyterlab/pull/12350/commits/be9c5da7b90da4aeb194026e524df7057201ff69

jtpio:
Marking as ready for review since CI is now all green, and Binder seems to be running fine:

image

jtpio:
cc @blink1073 for review since this removes the workaround added in https://github.com/jupyterlab/jupyterlab/pull/11249, and may fix https://github.com/jupyterlab/jupyterlab/issues/11248?

jtpio:
also cc @fcollonval @bollwyvl as part of the discussion in https://github.com/jupyterlab/jupyterlab/issues/12225

yarn.lock
```diff
@@ -10412,6 +10412,15 @@ postcss@^8.2.15, postcss@^8.3.11, postcss@^8.4.5:
     picocolors "^1.0.0"
     source-map-js "^1.0.1"
 
+postcss@^8.4.7:
```
bollwyvl, line 10415: looks like this could be deduped with the above?

yarn.lock
```diff
@@ -10412,6 +10412,15 @@ postcss@^8.2.15, postcss@^8.3.11, postcss@^8.4.5:
...
+  version "8.4.12"
+  resolved "https://registry.yarnpkg.com/postcss/-/postcss-8.4.12.tgz#1e7de78733b28970fa4743f7da6f3763648b1905"
```
bollwyvl, line 10417: all the new packages are pulling from `yarnpkg.com`, and the status quo links are all `npmjs.com`... i guess for consistency the `npmjs` is better?

maybe reset to the starting point, `jlpm cache clean` to be sure?

yarn.lock
jtpio: Thanks for catching this, should be fixed now.

yarn.lock
jtpio: Indeed it looks like both were mixed up. https://github.com/jupyterlab/jupyterlab/pull/12350/commits/4eabb6e1b45d89baa3ef2af995abce0d553e4bfe normalized on `npmjs.org`.

https://github.com/jupyterlab/jupyterlab/pull/12350/commits/b7ce2f402730c8bcb89e15a7a81c64d9e86f0a6b tried to add the default registry to `.yarnrc` but it somehow made the `release_test` check fail.

So we can defer that to a separate issue / PR since it's not strictly related to this change.

bollwyvl:
One _minor_ thing while looking over the dep changes: `webpack 5.72` dropped _today_ which closes out this issue:

https://github.com/webpack/webpack/issues/13764

Since this seems like kind of drastic change, it may be worth a little bake-off of `5.71` and `5.72` defaults to see which outcome we like... it's always hard to assess when file count/size will make a difference.

jtpio:
> Since this seems like kind of drastic change

Agree. Maybe we should do the webpack update in a separate PR then? Just so it's easier to troubleshoot later in case this brings some issues.

fcollonval (approved):
Thanks @jtpio

jtpio:
> raising the bottom pin so we can rely on that behavior

Maybe we could start using `~` for the webpack dependency like we do for `typescript`, if webpack starts bringing risky changes in minor releases.

jtpio:
OK so let's get that in, and we can continue in https://github.com/jupyterlab/jupyterlab/pull/12423 before making a new 4.0 pre-release with these changes.

PR #12351, opened 2022-04-05
Add CI check that fails when files in the staging directory are modified

galathinius:
## References

Fixes #7237 

## Code changes

Add a CI check when files in jupyterlab/staging are modified.
Fails if any files in staging other than the whitelisted ones are modified.
Currently whitelisted files:
 - yarn.js

## User-facing changes

None

## Backwards-incompatible changes

None

PR #12360, opened 2022-04-06
Load/auto-save document from the back-end using y-py

davidbrochart:
## References

Follow-up of #11599.

## Code changes

We don't fetch a file content through HTTP anymore, the file is loaded when the Yjs websocket is opened.
The front-end doesn't request or put an initial content through the Yjs websocket anymore, it just syncs the Y document directly.
Pressing Ctrl-S in collaborative mode doesn't send an HTTP request, as every change to the document is auto-saved in the back-end.

## User-facing changes

When a user changes a document, it is marked as dirty for one second then automatically saved.
A change made to a file in the back-end is automatically propagated to every user.

`requestInitialContent` and `putInitializedState` have been removed from the `IDocumentProvider` interface in the `@jupyterlab/docprovider` package.


davidbrochart:
I also get [this error](https://github.com/y-crdt/ypy/issues/28) back-end, so it might be related.

jupyterlab/handlers/ydoc.py
```diff
@@ -52,7 +61,25 @@ def source(self):
...
+        with self._ydoc.begin_transaction() as t:
+            ycells = []
+            for cell in value["cells"]:
+                ycell = Y.YMap()
+                ycell.set(t, "source", cell["source"])
+                ycell.set(t, "metadata", cell["metadata"])
+                ycell.set(t, "cell_type", cell["cell_type"])
+                ycell.set(t, "id", cell["id"])
+                ycells.append(ycell)
+            self._ycells.push(t, ycells)
```
davidbrochart, line 80: We should probably delete all prior cells before pushing, even though it is only initialized once so it should not be a problem.

jupyterlab/handlers/ydoc.py
```diff
@@ -23,6 +27,11 @@ def __init__(self):
     def source(self):
         return str(self._ysource)
 
+    @source.setter
+    def source(self, value):
+        with self._ydoc.begin_transaction() as t:
+            self._ysource.push(t, value)
```
davidbrochart, line 33: We should probably delete all prior characters before pushing, even though it is only initialized once so it should not be a problem.

jupyterlab/handlers/ydoc.py
dmonad: I think it would be better to log an error and do nothing if this happens. Overwriting means basically dataloss. We expect that the document is empty.

davidbrochart: We also want to support changes that happen in the back-end, for instance when changing to another git branch changes the whole content of the file. We might want to ask the user what they want to do in that case (ignore the file change and save their changes, or loose their changes and revert to the saved file), but I think that we should support overwriting the Y model.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -79,6 +79,8 @@ async def open(self, type_path):
         self.saving_document = None
         self.id = str(uuid.uuid4())
         self.room_id = file_path
+        self.file_type = file_type
+        self.file_format = "text"
```
davidbrochart, line 83: Leftover?

jupyterlab/handlers/ydoc.py
```diff
@@ -141,18 +141,29 @@ def source(self, value):
...
+
                 ycell.set(t, "source", cell["source"])
                 ycell.set(t, "metadata", cell["metadata"])
                 ycell.set(t, "cell_type", cell["cell_type"])
-                ycell.set(t, "id", cell["id"])
+
+                if cell["cell_type"] == "code":
+                    ycell.set(t, "execution_count", 0)
+                    ycell.set(t, "outputs", cell.get("outputs", []))
+
```
davidbrochart, line 159: Wouldn't this just work?
```py
ycell = Y.YMap(cell)
```

jupyterlab/handlers/yjs_echo_ws.py
hbcarlos: The `self.file_format = "text"` yes. `self.file_type = file_type` no

davidbrochart:
I still need to fix the tests, but this PR is ready for review as I'm satisfied with the user experience.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +32,170 @@
...
+        self.room.document.observe(self.maybe_save_document)
+        if self.saving_document is not None and not self.saving_document.done():
+            # the document is being saved, cancel that
+            self.saving_document.cancel()
+            self.saving_document = None
+        self.saving_document = asyncio.create_task(self.save_document())
+
+    async def save_document(self):
+        # save after 1 second of inactivity to prevent too frequent saving
+        await asyncio.sleep(1)
```
fcollonval, line 179: I would move the rate limiter in `maybe_save_document`. So the `save_document` method does what it says and nothing else.

fcollonval (commented):
Thanks @davidbrochart 

Here are some comments.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +32,170 @@
...
+                self.room.document.ydoc.initialized.set()
+                self.room.watcher = asyncio.create_task(self.watch_file())
+                # save the document when changed
+                self.room.document.observe(self.maybe_save_document)
+
+    async def watch_file(self):
+        # can only watch file changes if local file system
+        # if/when https://github.com/jupyter-server/jupyter_server/pull/783 gets in,
+        # we will check for a contents manager ability to watch file changes
+        if False:  # isinstance(self.contents_manager, (FileManagerMixin, AsyncFileManagerMixin)):
```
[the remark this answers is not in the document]
davidbrochart: :+1: (if we keep watchfiles, maybe we will just poll)

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +32,170 @@
...
+            abs_path = Path(self.file_path).resolve()
+
+            def filter(change, path):
+                return Path(path).resolve() == abs_path
+
+            async for _ in awatch(abs_path.parent, watch_filter=filter):
```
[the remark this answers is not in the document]
davidbrochart: True, I have tried several things as I had unexpected issues (like file changes not being detected).

> The `debounce` option should be set to a configurable value to use in the alternative while loop too.

I don't understand why we would need to debounce in the alternative while loop, it shouldn't be need, right? As we're polling at regular intervals.

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: :+1:

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +32,170 @@
...
+                await self.maybe_load_document()
+        else:
+            while True:
+                await asyncio.sleep(1)
```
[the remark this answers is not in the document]
davidbrochart: As a side note, we check if the [file was changed just before saving](https://github.com/jupyterlab/jupyterlab/blob/c059bbc921dd66868bc457407ab8d3d938f78548/jupyterlab/handlers/yjs_echo_ws.py#L188-L193), so the check interval could even be `None` (never check, except when saving).

davidbrochart:
I created https://github.com/davidbrochart/jupyterlab/tree/watchfiles which has watchfiles support (but it doesn't work :smile: ), I will remove watchfiles from this PR.

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: The problem is that the `save_document` task can be cancelled (if a new document change happens less than one second later), so this won't work.
I'll just rename `maybe_save_document` to `on_document_change`, and `save_document` to `maybe_save_document`. Feedback welcome!

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +25,163 @@
...
-# Here, we define some additional messageTypes that the server can interpret.
-# Messages that the server can't interpret should be broadcasted to all other clients.
 
+class JupyterRoom(YRoom):
+    def __init__(self, type):
+        super().__init__(has_internal_ydoc=True)
+        self.type = type
+        self.cleaner = None
+        self.watcher = None
+        self.document = YDOCS.get(type, YFILE)(self.ydoc)
```
dmonad, line 37: Can you please walk me through this python sorcery? How does it initialize a Yjs document exactly?

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -25,170 +25,163 @@
...
+            self.room.cleaner.cancel()
+
+        if not self.room.ydoc.initialized.is_set():
+            model = await ensure_async(
+                self.contents_manager.get(self.file_path, type=self.file_type)
+            )
+            self.last_modified = model["last_modified"]
+            # check again if initialized, because loading the file can be async
+            if not self.room.ydoc.initialized.is_set():
+                self.room.document.source = model["content"]
```
dmonad, line 96: This seems to handle all documents as a text document, is that correct?

dmonad (commented):
Thank you @davidbrochart! 

Can you please point me where you transform the enml document to a Yjs document? Or is this currently only text document transformation?

It might make sense to add some tests to ensure that the enml->yjs transformation works as expected.

I'm proposing that we generate a few enml documents with all kinds of edge-cases (different cell types, different types of outputs, a variety of associated metadata). We transform these documents to Yjs using `y-py`, then we read the document using `shared-model` and see whether they generate the same enml document.

davidbrochart:
> Can you please point me where you transform the enml document to a Yjs document? Or is this currently only text document transformation?

I'm not sure what you mean by "enml", but the "Jupyter Y documents" was just moved to https://github.com/davidbrochart/jupyter_ydoc.

> It might make sense to add some tests to ensure that the enml->yjs transformation works as expected.
> 
> I'm proposing that we generate a few enml documents with all kinds of edge-cases (different cell types, different types of outputs, a variety of associated metadata). We transform these documents to Yjs using `y-py`, then we read the document using `shared-model` and see whether they generate the same enml document.

That sounds good, I guess these tests would be part of the repository I mentioned above.

jupyterlab/handlers/yjs_echo_ws.py
```diff
@@ -3,15 +3,15 @@
 # Copyright (c) Jupyter Development Team.
```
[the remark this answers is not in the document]
davidbrochart: Yes :smile: 
What about `ydoc_handler.py`?

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: [Jupyter Y documents](https://github.com/davidbrochart/jupyter_ydoc) are registered through [entry points](https://github.com/davidbrochart/jupyter_ydoc/blob/3ec7528c8c5d714bbcdabcd7f9ec4305f891c4d4/setup.cfg#L30-L33). We currently have two built-in ydocs:
- [YFile](https://github.com/davidbrochart/jupyter_ydoc/blob/3ec7528c8c5d714bbcdabcd7f9ec4305f891c4d4/jupyter_ydoc/ydoc.py#L49-L70): a generic text document.
- [YNotebook](https://github.com/davidbrochart/jupyter_ydoc/blob/3ec7528c8c5d714bbcdabcd7f9ec4305f891c4d4/jupyter_ydoc/ydoc.py#L73-L155): a Jupyter notebook.

They are high-level documents that use Ypy's `YDoc` underneath, but with methods to get text/JSON source, set from text/JSON, and observe changes. They must of course mimic their equivalent structures in the front-end.
Here `YDOCS` is just a dictionary of all the available Jupyter documents, e.g.:
```py
{'file': <class 'jupyter_ydoc.ydoc.YFile'>, 'notebook': <class 'jupyter_ydoc.ydoc.YNotebook'>}
```
We create a Jupyter document corresponding to the type requested by the front-end, defaulting to a `YFile`.

jupyterlab/handlers/yjs_echo_ws.py
davidbrochart: Not exactly, for a generic file the returned source is a text indeed, but for a notebook it is a JSON object (as a Python dictionary).

jupyterlab/handlers/yjs_echo_ws.py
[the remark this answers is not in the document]
dmonad: Yeah, sounds good!

jupyterlab/handlers/yjs_echo_ws.py
dmonad: Thank you, this is exactly what I was looking for!

It makes a lot of sense to outsource this to a separate package. This is basically the equivalent of `shared-models` package in JupyterLab. What do you think about naming this package `shared-models` as well?

davidbrochart: I'm not sure, `shared-model` is a good name in the context of JupyterLab, but the current `jupyter_ydoc` could be used outside JupyterLab. For instance, I'm planning to use it in [jpterm](https://github.com/davidbrochart/jpterm), a Python Jupyter client that allows to work collaboratively with e.g. JupyterLab:

https://user-images.githubusercontent.com/4711805/166236284-320b5275-79c0-4a76-9aa5-e2ad09964812.mp4

:

setup.cfg
```diff
@@ -42,6 +42,8 @@ install_requires =
     notebook_shim>=0.1
     jinja2>=3.0.3
     y-py>=0.4.6,<0.5.0
+    ypy-websocket>=0.1.2
+    jupyter_ydoc>=0.1.0
```
fcollonval, line 46: The question about where that code should live is good. But I'm reluctant to multiply the downstream dependencies here. I would prefer to keep them in this repo or move them to `jupyterlab_server` for now if the goal is to allow reuse without having to link to the big `jupyterlab` package.

And whatever is the final choice, that code must be part of one Jupyter GitHub organization.

fcollonval (changes requested):

setup.cfg
davidbrochart: The plan is to move:
- `ypy-websocket` to https://github.com/y-crdt,
- `jupyter_ydoc` to https://github.com/jupyter-server.

davidbrochart: We can move `jupyter_ydoc` to `jupyter-server`, but is it an issue that `ypy-websocket` is not part of a Jupyter GH org? It's the same for `y-py`, these libraries are not tied to Jupyter IMO.

fcollonval: y-crdt is definitely a good candidate too.

dmonad:
Just tried this out locally. Works really well for me :confetti_ball: I feel the loading times are much faster now.

davidbrochart:
FYI `jupyter_ydoc` has been transfered to https://github.com/jupyter-server/jupyter_ydoc.

davidbrochart:
I'm not sure about the CI failure for [Reject changes to staging](https://github.com/jupyterlab/jupyterlab/runs/6349452715?check_suite_focus=true). It is probably caused by [the Yjs version update](https://github.com/jupyterlab/jupyterlab/pull/12360/files#diff-60ebdb5ae7c29a3b2bc12680ee42ef7291899fd9af7887d9399a08fd18e3e841R129), but I think it is needed.
Other CI failures seem unrelated.

fcollonval (approved):
Thanks @davidbrochart 

The UI doc error is not related.
Regarding the `jupyterlab/staging/package.json` file, it should not be modified as it is gonna be copied from `dev_mode` at publication time. So only changing `dev_mode/package.json` is required.

jupyterlab/labapp.py
```diff
@@ -538,6 +537,14 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
 
     collaborative = Bool(False, config=True, help="Whether to enable collaborative mode.")
 
+    collab_file_poll_interval = Int(
```
jtpio, line 540: How about a more explicit name for this public trait?

```suggestion
    collaborative_file_poll_interval = Int(
```

jtpio: A shorter `file_poll_interval` could also work.

packages/docprovider/src/tokens.ts
```diff
@@ -12,16 +12,6 @@ export const IDocumentProviderFactory = new Token<IDocumentProviderFactory>(
  * An interface for a document provider.
  */
 export interface IDocumentProvider {
-  /**
-   * Resolves to true if the initial content has been initialized on the server. false otherwise.
-   */
-  requestInitialContent(): Promise<boolean>;
```
jtpio, line 18: Mind adding a note to the extension migration about these methods being gone from the interface?

jtpio:
> Regarding the `jupyterlab/staging/package.json` file, it should not be modified as it is gonna be copied from `dev_mode` at publication time. So only changing `dev_mode/package.json` is required.

Normally this should be caught by the CI, since such check was added in https://github.com/jupyterlab/jupyterlab/pull/12351. Looks like it might not have been triggered.

davidbrochart:
@jtpio I took your suggestions into account in this last commit.

jupyterlab/handlers/ydoc_handler.py
```diff
@@ -0,0 +1,177 @@
...
+        self.watcher = None
+        self.document = YDOCS.get(type, YFILE)(self.ydoc)
+
+
+class JupyterWebsocketServer(WebsocketServer):
+    def get_room(self, path: str) -> JupyterRoom:
+        file_type, file_path = path.split(":", 1)
+        if file_path not in self.rooms.keys():
+            self.rooms[file_path] = JupyterRoom(file_type)
+        return self.rooms[file_path]
```
hbcarlos, line 35: I think I'm having an issue here. When opening a notebook as a plain text document closing it and opening the same notebook with the Notebook editor the content disappears.

I think the problem is that we are using the same key to store both rooms for two different types of files.
I believe we should use `type + path` as a key to store the room, this way we will create different rooms when opening the same document with different editors.

```suggestion
        file_type, file_path = path.split(":", 1)
        if path not in self.rooms.keys():
            self.rooms[path] = JupyterRoom(file_type)
        return self.rooms[path]
```

hbcarlos (commented):
I

jupyterlab/handlers/ydoc_handler.py
davidbrochart: Thanks Carlos, good catch!
I had never tried opening the same file with different editors. Now I'm seeing issues with the dirty flag "bouncing" between the windows where the file is opened as different types. I need to investigate.

davidbrochart: The issue is that when we have e.g. a notebook opened as such and also as a text document, modifying the notebook doesn't sync the text document, because they are actually two different (incompatible) Y documents. What happens is that when the notebook is changed, it is saved to disk after one second. Then the text document watches the file change and reverts to the file content, setting its dirty flag. After one second, it is saved to disk and the notebook watches the file change, and so on.
I think that for the long term fix, we need a kind of "Y bridge" between two different Y documents of the same file, so that we can sync the two in real time. But it is not going to be easy, if possible at all, because the text file is "unstructured" and it may be impossible to map character changes to higher-level notebook changes. Or maybe we could recreate the notebook from scratch, but then it would not be incremental updates.
For now, let's just break this circle by not actually saving if what we want to save is already on disk. The two opened documents will sync "through the file system", and not immediately, but I think it is fine.
What do you think @hbcarlos @fcollonval @jtpio @dmonad ?

hbcarlos:
I'm trying out this PR and from a UX perspective, having the save button is a bit confusing.
Should we deactivate it (leave it there but greyed out) when the collaboration is activated and add a tooltip explaining that the document is saved automatically?
This way we can use the tooltip to notify the users that the save button doesn't work in the collaborative mode.

packages/docregistry/src/context.ts
```diff
@@ -637,6 +637,43 @@ export class Context<
...
+      ...(this._factory.fileFormat !== null
+        ? { format: this._factory.fileFormat }
+        : {})
+    };
+    const path = this._path;
+    return this._manager.ready
+      .then(() => {
+        return this._manager.contents.get(path, opts);
+      })
+      .then(contents => {
```
hbcarlos, line 656: I'm seeing another issue here. The image doesn't render into the screen when opening a picture because the widget needs the attribute `format` from the context.
I believe the problem is coming from [here](https://github.com/jupyterlab/jupyterlab/blob/13ca0e0caea5ad943f08627f0c2ed447bc79d829/packages/docregistry/src/context.ts#L654) when requesting the metadata of the file. The server doesn't return the file format.
I've been investigating it, but I can not figure out why.

davidbrochart: But we don't support collaborative images :smile: 
I mean, `_loadContext` is called when in collaborative mode because we need the metadata of the file without its content, but for an image we actually want the content, right?
Is there some condition to insert [here](https://github.com/jupyterlab/jupyterlab/blob/13ca0e0caea5ad943f08627f0c2ed447bc79d829/packages/docregistry/src/context.ts#L265) for types that are not supported in collaborative mode?

davidbrochart: I pushed https://github.com/jupyterlab/jupyterlab/pull/12360/commits/d579a7feafe7fc793208a4c383cec6d4ae9c2093, let me know if that's the correct way to do it @hbcarlos.

hbcarlos: We are sharing every file at the moment. Actually, the implementation that we have in master, is also sharing every file.
For example, an image (`.jpg`) has the type `file` and uses the `DocumentModel` as a model for the data.

The problem is that we have only three types `file`, `notebook` or `directory`. Using the `this._factory.contentType` we can not differentiate between them since a `.txt` and a `.jpg` have the same type. We can not use `this._factory.fileFormat` ("json" | "text" | "base64")  either because I think an image can be loaded as `text` and rendered using its mimetype, see https://github.com/jupyterlab/jupyterlab/blob/a41b9816fdedb936be1589dbc98363f62f5f4574/packages/imageviewer/src/widget.ts#L193-L197

The only solution I can see is by extension, but we will need to explicitly say what extensions should be shared, and I don't think it is a good idea.

davidbrochart: I pushed https://github.com/jupyterlab/jupyterlab/pull/12360/commits/5708d02d2388b68d24231f351f51e49d41265be5 where we set the room name as `format:type:path`, because the RTC handler needs to load/save with the correct file format.

hbcarlos: LGTM

davidbrochart: I'm seeing issues when renaming a notebook a lot of times, it ends up not saving the changes. I need to investigate, maybe it's a Ypy issue with observing changes.

davidbrochart: Fixed in https://github.com/jupyterlab/jupyterlab/pull/12360/commits/c4c6aa4146eda3d77c22d29bbb3492100471f7c4.

davidbrochart: There are still small issues with the dirty flag sometimes not being cleared when opening a file. I'm going to add:
```py
self.room.document.dirty = False
```
Right [after setting the source](https://github.com/jupyterlab/jupyterlab/blob/c4c6aa4146eda3d77c22d29bbb3492100471f7c4/jupyterlab/handlers/ydoc_handler.py#L93), but we need https://github.com/jupyter-server/jupyter_ydoc/pull/16 first.

davidbrochart:
I think there is some redundancy in the way we observe changes to a document, because we observe changes to both the document "source" *and* the dirty flag. There should be only one indicator that the document was changed, and it is the dirty flag. I'm going to clean that up.

hbcarlos:
Are you talking about the `DocumentModel` and `NotebookModel` in the front-end?

There is some redundancy in the signals that these classes emit, `contentChanged` and `sourceChanged` for example. But the dirty flag is part of the state of the document and I don't think it should replace the `contentChanged` or `sourceChanged` signals.

davidbrochart:
I'm talking about the `dirty` field of the `state` YMap.

davidbrochart:
> There should be only one indicator that the document was changed, and it is the dirty flag.

That's not exactly true, because the front-end sets the dirty flag when the document has changed, but in the back-end we also want to be notified when the document keeps on changing after it has initially changed, so that we wait another second before actually saving.
With https://github.com/jupyterlab/jupyterlab/pull/12360/commits/be4c4968f509d2de82802391df139cb4ca67a580 I believe that we are good.

hbcarlos:
Got it! You are right. For saving the document, we should listen for changes to the content and not the state.
In my opinion, in the documents, we should split the observe function into state and "content" (source for the YFile and metadata and cells for the YNotebook).
https://github.com/jupyter-server/jupyter_ydoc/blob/72f1c7bb7f8ea08b67ece07151c380e4ad657913/jupyter_ydoc/ydoc.py#L70
https://github.com/jupyter-server/jupyter_ydoc/blob/72f1c7bb7f8ea08b67ece07151c380e4ad657913/jupyter_ydoc/ydoc.py#L142
And in the YDocWebSocketHandler, listen for changes in the content.

davidbrochart:
We also want to observe the dirty flag, because for instance a notebook opened as a text document will have its dirty indicator set from the front-end when the notebook opened as such changes. In that case we don't want to save the notebook again from the text document but we want to clear its dirty flag.

> In my opinion, in the documents, we should split the observe function into state and "content" (source for the YFile and metadata and cells for the YNotebook).

We can also have a unique `observe` method for the whole document, and see what changed in the callback. That works well too.

davidbrochart:
That sound good, the notebook's metadata is actually stored in `meta["metadata"]` so we can move `nbformat` and `nbformat_minor` to `meta["nbformat"]` and `meta["nbformat_minor"]`.
I'll make the changes in `jupyter_ydoc`, we also need to change the front-end.

hbcarlos:
I'll make the changes in the frontend.

davidbrochart:
Thanks, the problem is that we would need `@jupyterlab/shared-models` to be released, so this has to be done in a separate PR.

davidbrochart:
> Thanks, the problem is that we would need `@jupyterlab/shared-models` to be released, so this has to be done in a separate PR.

Actually, it's only to have tests pass in https://github.com/jupyter-server/jupyter_ydoc/pull/21, so maybe we don't need a release.

dmonad:
> We also want to observe the dirty flag, because for instance a notebook opened as a text document will have its dirty indicator set from the front-end when the notebook opened as such changes. In that case we don't want to save the notebook again from the text document but we want to clear its dirty flag.

It might make sense to only modify the shared `dirty` property on the backend. If server observes a change (content, metadata, or otherwise), then the document state is dirty. The server can also set a variable "last_modified" that it maintains locally. If the document state has been dirty for more than X seconds, the server can store the document and set `dirty=false` again.

A potential problem is that a client modifies a document (setting `dirty=true` flag) at the same time as the server stores a document and sets `dirty=false`. These kinds of problems can be avoided if `dirty` is only modified by the server.

davidbrochart:
I was thinking about that too, but @afshin mentioned that we still want the dirty flag to work when the front-end looses connection, and this is a good point.
But I agree things would be simpler if the dirty flag was only set/cleared from the back-end.

dmonad:
Hmm.. that makes sense. There could be a local `dirty` variable on the client as well set is set to true in anticipation that the server will soon set the real `dirty` flag to true. This way the client can still set a `local` dirty flag, but prefers the remote state. The getter for this function could look like this:

```js
class SharedModel {
  constructor () {
    ..
    this._localDirty = false
    this._ymeta = ydoc.getMap('meta')
    this._ymeta.observe(event => {
      if (event.changed.has('dirty')) {
        this._localDirty = this._ymeta.get('dirty')
        // potentially fire dirty-changed observer here
      }
    })
    ..
  }

  get dirty () {
    return this._localDirty
  }

  set dirty () {
    this._localDirty = true
  }
}
```

hbcarlos:
Yep, I agree we should only modify it on the backend and keep a local variable in the frontend.
I'll make the changes

davidbrochart:
@hbcarlos I thought that when the front-end changes the document, it would only set a local dirty flag, and let the back-end set the shared dirty flag when the document changes, but with your changes I see that the dirty flag is seen by all front-ends, and I've not changed the back-end yet to set the dirty flag.
Am I missing something?

davidbrochart:
Mmmm I'm seeing the same thing after this new commit @hbcarlos.

hbcarlos:
Yes, it is. the sharedModel is still initializing the dirty property.
https://github.com/jupyterlab/jupyterlab/blob/d8ee1cb51d2ac73ff00976d43189421e1de69ed9/packages/shared-models/src/ymodels.ts#L155
https://github.com/jupyterlab/jupyterlab/blob/d8ee1cb51d2ac73ff00976d43189421e1de69ed9/packages/shared-models/src/ymodels.ts#L377

I'm on it 🤣

davidbrochart:
This is ready to be merged!

fcollonval:
Let's move on with this.

damianavila:
@davidbrochart, just to confirm, you need to have collaborative mode enabled to make this work as described in the top message, right?

davidbrochart:
@damianavila yes, this only works in collaborative mode.

PR #12364, opened 2022-04-06
type-only and lazy imports of settings widgets

### LINKED ISSUE #12363: Lazy load rjsf assets
### Problem

The rjsf bundle is _large_: larger than the jupyterlab core bundle, at least in dev mode.

### Proposed Solution

Only load the tokens at start up.

Defer the loading of the top-level settings widget (form or JSON) until a settings form is actually requested.

Tree shaking _should_ take care of the rest. If not, it might  be required to make a `createSettingsForm` or something that is cheap to import which wraps the heavy duty one.

### Additional context

On [jupyterlab-rjsf](https://github.com/deathbeds/jupyterlab-starters/blob/master/packages/jupyterlab-rjsf/src/async-component/index.tsx) we offered a whole async loading component, cargo-culted from somewhere.

bollwyvl:
## References

- fixes #12363

## Code changes

- [x] use type imports for heavy widgets from `settingseditor`
- [x] lazy import when first used `SettingsEditor` and `JsonSettingEditor`

## User-facing changes

- should not load the rjsf bundle up front

- n/a

### DIFF
--- a/galata/test/jupyterlab/texteditor.test.ts
+++ b/galata/test/jupyterlab/texteditor.test.ts
@@ -25,6 +25,7 @@ test.describe('Text Editor Tests', () => {
 
     await page.menu.clickMenuItem('Settings>Settings Editor');
 
+    await page.waitForSelector('text=Text Editor');
     await page.click('text=Text Editor');
 
     // Add two rulers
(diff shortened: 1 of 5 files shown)


fcollonval (approved):
Thanks @bollwyvl 

> CI error is unrelated.

fcollonval:

bollwyvl:
Thanks!

PR #12365, opened 2022-04-07
Add CodeViewer widget and openCodeViewer command

### LINKED ISSUE #12361: Introduce a transient Code Viewer to core
### Problem

Currently if an extension wants to display code in an editor UI int requires persisting the contents as a file.

### Proposed Solution

I recently published the code viewer extension in Elyra, I purpose merging that implementation in core. This would consist of moving the widget code into the code-editor package and the command declaration code into the file editor extension. 

### Additional context

This feature would allow developers to open transient "files" for viewing without the need to persist them to the file system.

Current code: https://github.com/elyra-ai/elyra/tree/master/packages/code-viewer

This feature was demoed and pitched at the April 6th 2022 dev meeting. During that meeting many potential additions to functionality were purposed that could be handled in followup issues such as:
- Allowing for users to save/persist the transient file (this would require re-opening it in an actual editor)
- Add open viewers to the restorer so they persist on page refresh
- This has a similar functionality as the read-only editor in the debugger, but that requires a persisted file that's just outside the workspace.

ajbozarth:
Code Viewer is a transient readonly editor that displays the given code without persisting it to the file system.

## References

Fixes #12361

## Code changes

Adds the CodeViewer widget and the command to open a code viewer.

The widget is added to the codeeditor package as all its imports are in that package. The command is added in the file editor-extension package as that extension was the closest in functionality of the extensions dealing with the codeeditor package.

## Screenshots

Here is a screen record showing a use case of the Code Viewer in Elyra. In it we open a "file" returned by an api call without persisting it to the file system.

https://user-images.githubusercontent.com/13952758/164115722-b663b49f-7b33-4edd-a4f9-3140cd515209.mov

---
Edited: Add keyword in from of issue to link it

### DIFF
--- a/packages/codeeditor/src/viewer.ts
+++ b/packages/codeeditor/src/viewer.ts
@@ -11,11 +11,13 @@ export class CodeViewerWidget extends Widget {
    */
   constructor(options: CodeViewerWidget.IOptions) {
     super();
+    this.content = options.content;
 
     this.model = new CodeEditor.Model({
       value: options.content,
       mimeType: options.mimeType
     });
+    this.mimeType = this.model.mimeType;
 
     const editorWidget = new CodeEditorWrapper({
       factory: options.factory,
@@ -28,6 +30,8 @@ export class CodeViewerWidget extends Widget {
     layout.addWidget(editorWidget);
   }
 
+  content: string;
+  mimeType: string;
   model: CodeEditor.IModel;
   editor: CodeEditor.IEditor;
 }
--- a/packages/codeeditor/src/viewer.ts
+++ b/packages/codeeditor/src/viewer.ts
@@ -24,7 +24,7 @@ export class CodeViewerWidget extends Widget {
     layout.addWidget(editorWidget);
   }
 
-  static getCodeViewer(
+  static createCodeViewer(
     options: CodeViewerWidget.INoModelOptions
   ): CodeViewerWidget {
     const model = new CodeEditor.Model({
@@ -34,8 +34,13 @@ export class CodeViewerWidget extends Widget {
     return new CodeViewerWidget({ factory: options.factory, model });
   }
 
-  getContent = (): string => this.model.value.text;
-  getMimeType = (): string => this.model.mimeType;
+  get content(): string {
+    return this.model.value.text;
+  }
+
+  get mimeType(): string {
+    return this.model.mimeType;
+  }
 
   model: CodeEditor.IModel;
   editor: CodeEditor.IEditor;
--- a/packages/codeeditor/src/viewer.ts
+++ b/packages/codeeditor/src/viewer.ts
@@ -11,17 +11,11 @@ export class CodeViewerWidget extends Widget {
    */
   constructor(options: CodeViewerWidget.IOptions) {
     super();
-    this.content = options.content;
-
-    this.model = new CodeEditor.Model({
-      value: options.content,
-      mimeType: options.mimeType
-    });
-    this.mimeType = this.model.mimeType;
+    this.model = options.model;
 
     const editorWidget = new CodeEditorWrapper({
       factory: options.factory,
-      model: this.model
+      model: options.model
     });
     this.editor = editorWidget.editor;
     this.editor.setOption('readOnly', true);
@@ -30,8 +24,19 @@ export class CodeViewerWidget extends Widget {
     layout.addWidget(editorWidget);
   }
 
-  content: string;
-  mimeType: string;
+  static getCodeViewer(
+    options: CodeViewerWidget.INoModelOptions
+  ): CodeViewerWidget {
+    const model = new CodeEditor.Model({
+      value: options.content,
+      mimeType: options.mimeType
+    });
+    return new CodeViewerWidget({ factory: options.factory, model });
+  }
+
+  getContent = (): string => this.model.value.text;
+  getMimeType = (): string => this.model.mimeType;
+
   model: CodeEditor.IModel;
   editor: CodeEditor.IEditor;
 }
@@ -44,6 +49,21 @@ export namespace CodeViewerWidget {
    * The options used to create an code viewer widget.
    */
   export interface IOptions {
+    /**
+     * A code editor factory.
+     */
+    factory: CodeEditor.Factory;
+
+    /**
+     * The content model for the viewer.
+     */
+    model: CodeEditor.Model;
+  }
+
+  /**
+   * The options used to create an code viewer widget without a model.
+   */
+  export interface INoModelOptions {
     /**
      * A code editor factory.
      */
--- a/packages/codeeditor/src/viewer.ts
+++ b/packages/codeeditor/src/viewer.ts
@@ -1,0 +1,57 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { StackedLayout, Widget } from '@lumino/widgets';
+import { CodeEditor, CodeEditorWrapper } from './';
+
+export class CodeViewerWidget extends Widget {
+  /**
+   * Construct a new code viewer widget.
+   */
+  constructor(options: CodeViewerWidget.IOptions) {
+    super();
+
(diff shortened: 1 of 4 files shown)


ajbozarth:
As an initial step I've moved the current code over from https://github.com/elyra-ai/elyra/tree/master/packages/code-viewer without changing it functionality. I did clean up a few minor things that were found due to the different lint rules, but no functional differences.

Based on discussion in #12361 I believe making the viewer restorable is blocker for merging this, but any other additions or updates could be handled in followup PRs.

ajbozarth:
Updated the description with a screen record as was requested in last weeks dev meeting

packages/codeeditor/src/viewer.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+    this.editor = editorWidget.editor;
+    this.editor.setOption('readOnly', true);
+
+    const layout = (this.layout = new StackedLayout());
+    layout.addWidget(editorWidget);
+  }
+
+  public model: CodeEditor.IModel;
+  public editor: CodeEditor.IEditor;
```
fcollonval, line 31: The rules is usually to drop the `public` visibility. Should they be readonly?

```suggestion
  model: CodeEditor.IModel;
  editor: CodeEditor.IEditor;
```

packages/codeeditor/src/viewer.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { StackedLayout, Widget } from '@lumino/widgets';
+import { CodeEditor, CodeEditorWrapper } from './';
+
+export class CodeViewerWidget extends Widget {
```
fcollonval, line 7: I saw that FileEditor has some logic to ensure focus on `mousedown`. Do you think it is needed here?

packages/codeeditor/src/viewer.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { StackedLayout, Widget } from '@lumino/widgets';
+import { CodeEditor, CodeEditorWrapper } from './';
```
fcollonval, line 5: The practice is to provide the explicit local file for import

```suggestion
import { CodeEditor, CodeEditorWrapper } from './widget';
```

fcollonval (commented):
Thanks @ajbozarth 

I have minor suggestions/questions.

fcollonval:
That CI error is legit:

```
../fileeditor-extension/src/commands.ts(4,10): error TS6133: 'ILayoutRestorer' is declared but its value is never read.
```

packages/codeeditor/src/viewer.ts
ajbozarth: yeah I can change that, I copied this format from the FileEditor and did think it was odd

packages/codeeditor/src/viewer.ts
ajbozarth: again something I copied from widget.ts in the same directory and though was odd, I'll fix it

packages/codeeditor/src/viewer.ts
ajbozarth: It is not needed here since this is readonly. That functionality makes sure that the editor get keyboard focus for typing, which you can't doing in readonly

ajbozarth:
pushed an update fixing the items raised by @fcollonval 

still (back to) working on restoration and potentially a way to use a model to allow live updates (as suggested by @bollwyvl ). I hope to have some stuff pushed by end of week

packages/codeeditor/src/viewer.ts
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { StackedLayout, Widget } from '@lumino/widgets';
+import { CodeEditor } from './editor';
+import { CodeEditorWrapper } from './widget';
+
+export class CodeViewerWidget extends Widget {
```
jtpio, line 8: Since this will be part of the public API, a question about the exported name.

Is this functionality specific to code? Or should we indicate this can be used for any text content? If so maybe the "Code" part in `CodeViewerWidget` could be misleading.

ajbozarth: This was actually a point of discission when I first created this widget in Elyra. I had originally called it `TextViewerWidget` but didn't quite like it and after some discussion in a Elyra dev call we settled on `CodeViewerWidget`.

Our reasoning was that this wraps the CodeEditor and specifically enabled code highlighting based on mime type, so even though it can be used for any text, its purpose is for displaying code. I'm not too stuck on the name, but I definitely prefer `CodeViewerWidget` more than `TextViewerWidget`

ajbozarth:
I just pushed a couple of commits that add requested functionality:

- 6c060b7 adds restoration support as requested
- 1575222 update the widget to take a model and move the creation of the model from the previous args to a static creator function.

> also offer a model instead of just a string so it can be live updated
@bollwyvl does 1575222 address this functionality you mentioned in the issue? I don't have a use case to test it

If so then this completes all the requested work on this PR

packages/fileeditor-extension/src/commands.ts
```diff
@@ -1192,4 +1201,60 @@ export namespace Commands {
...
+        widget.id = args.widgetId;
+      }
+      const main = new MainAreaWidget({ content: widget });
+      await tracker.add(main);
+      app.shell.add(main, 'main');
+      return widget;
+    };
+
+    app.commands.addCommand(CommandIDs.openCodeViewer, {
+      execute: (args: any) => {
```
fcollonval, line 1255: You need to add a label.

There is a new test on master to check that and to create documentation automatically for all commands as they can be used in settings to tune menus and toolbars.

packages/codeeditor/src/viewer.ts
```diff
@@ -28,6 +24,19 @@ export class CodeViewerWidget extends Widget {
...
+  ): CodeViewerWidget {
+    const model = new CodeEditor.Model({
+      value: options.content,
+      mimeType: options.mimeType
+    });
+    return new CodeViewerWidget({ factory: options.factory, model });
+  }
+
+  getContent = (): string => this.model.value.text;
+  getMimeType = (): string => this.model.mimeType;
```
fcollonval, line 38: The practice in the code is more to use getter than method for this.
```suggestion
  get content(): string { return this.model.value.text; }
  get mimeType(): string { return this.model.mimeType; }
```

packages/codeeditor/src/viewer.ts
```diff
@@ -28,6 +24,19 @@ export class CodeViewerWidget extends Widget {
     layout.addWidget(editorWidget);
   }
 
+  static getCodeViewer(
```
fcollonval, line 27: Factories in the code are usually starting with `create` rather than `get`.

```suggestion
  static createCodeViewer(
```

fcollonval (commented):
Thanks @ajbozarth this is great.

I have ome styling suggestions and a fix needed for the command label.

packages/fileeditor-extension/src/commands.ts
ajbozarth: good to know, I had a label on that originally and removed it because I didn't think it was needed

packages/codeeditor/src/viewer.ts
ajbozarth: will do, I went back and forth on this trying to decide which to use (since both are found in core)

packages/codeeditor/src/viewer.ts
ajbozarth: thanks, I had trouble naming this

ajbozarth:
@fcollonval just pushed your recommended changes

ajbozarth:
IIUC test failure is unrelated to my changes

fcollonval (approved):
LGTM thanks @ajbozarth

PR #12368, opened 2022-04-07
Toolbar items may not act on the proper target

### LINKED ISSUE #12177: Notebook toolbar buttons do not respect context
## Description

When clicking a button on the notebook toolbar, JupyterLab does not update the current document from the click event. This means that one can restart the kernel of second open notebook tab if it is the current active document, and one then directly clicks the first notebook toolbar.

## Reproduce
Clicking "Restart and Run All" on the right notebook acts on the left notebook kernel, if that notebook body was most recently interacted with.
image

## Context

- JupyterLab version: 3.3.0

fcollonval:
## References

Fixes #12177

## Code changes

Due to the definition of toolbar items from a data description, the primarly actions are calling a command. To figure out the targeted widget for a command, a tracker is commonly used. Therefore when displaying side-by-side multiple documents, clicking on a toolbar button in the non-active document needs first to turn the widget own as the active widget. As CommandToolbarButton action is triggered on `mousedown` event to avoid focusing the button itself (and losing the focus on the working cell for example), the fix in this PR is using `mousedown` event listener at capture phase (that occurs before the classical bubbling phase) to focus the widget if it contains the event target and if it does not contain the active element (best guess for the current focused element).

## User-facing changes

Toolbar button should act on the expected widget

N/A

### DIFF
--- a/galata/test/jupyterlab/notebook-toolbar.test.ts
+++ b/galata/test/jupyterlab/notebook-toolbar.test.ts
@@ -13,7 +13,6 @@ async function populateNotebook(page: IJupyterLabPageFixture) {
     '## This is **bold** and *italic* [link to jupyter.org!](http://jupyter.org)'
   );
   await page.notebook.addCell('code', '2 ** 3');
-  // await page.notebook.runCell(2, true);
 }
 
 test.describe('Notebook Toolbar', () => {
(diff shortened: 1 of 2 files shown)


fcollonval:
@afshin @jasongrout I know that the focus mechanism is sensible. Would you mind having a look at this one?

fcollonval:
There has been some discussion on a PR against this branch PR: https://github.com/fcollonval/jupyterlab/pull/4

fcollonval:

PR #12373, opened 2022-04-07
Update command to `pip install -e ".[test]"` in the contribution documentation

jtpio:
## References

Update the command to install JupyterLab in editable mode, so it also installs extra dependencies like `pre-commit`.

## Code changes

Contributing documentation update.

## User-facing changes

None

None

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -300,7 +300,7 @@ Then use the following steps:
 
    git clone https://github.com/<your-github-username>/jupyterlab.git
    cd jupyterlab
-   pip install -e .
+   pip install -e ".[test]"
    jlpm install
    jlpm run build  # Build the dev mode assets (optional)
    jlpm run build:core  # Build the core mode assets (optional)


fcollonval (approved):
Thanks @jtpio

docs/source/developer/contributing.rst
```diff
@@ -300,7 +300,7 @@ Then use the following steps:
 
    git clone https://github.com/<your-github-username>/jupyterlab.git
    cd jupyterlab
-   pip install -e .
+   pip install -e ".[test]"
```
gabalafou, line 303: Just FYI, this is going to mess with most Mac users (like me) since zsh became default and brackets inside double quotes have some meaning in zsh apparently. I had to run:

    pip install -e '.[test]'

jtpio: @gabalafou do you know if this is specific to Mac?

I also use zsh but on Ubuntu Linux, and brackets inside double quotes work fine.

gabalafou: Glad you asked. I think I was wrong. 

I remembered it as being a double versus single quotes issue, but digging in this morning, I think it was actually an issue of quotes versus no quotes. I found a longer discussion about this issue where it looks like the conclusion is to [use double quotes for maximum cross-system compatibility](https://github.com/mu-editor/mu/issues/852#issuecomment-499870171)

gabalafou: So I think the command as you wrote it with double quotes is good and we should leave it as is. I tried it on my Mac and it worked fine

PR #12376, opened 2022-04-07
Add parent header to input reply kernel message

### LINKED ISSUE #12355: parent header for input_reply
### Problem

input_reply doesn't have parent header and when implementing custom Jupyter Kernel Gateway server it is tricky to get input_reply matched to one of multiple pending input_requests.  

### Proposed Solution

Probably input_reply could have corresponding input_request as a parent header?

davidbrochart:
## References

Fixes #12355.

## Code changes

The [kernel messaging protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#parent-header) says that "`_reply` messages MUST have a `parent_header`". This PR adds the parent header to `input_reply` messages, that was missing.

## User-facing changes

None.

None.

### DIFF
--- a/packages/outputarea/test/widget.spec.ts
+++ b/packages/outputarea/test/widget.spec.ts
@@ -379,7 +379,7 @@ describe('outputarea/widget', () => {
             parent_header: {
               date: '',
               msg_id: '',
-              msg_type: 'input_request',
+              msg_type: 'input_request' as const,
               session: '',
               username: '',
               version: ''
--- a/packages/outputarea/test/widget.spec.ts
+++ b/packages/outputarea/test/widget.spec.ts
@@ -376,6 +376,14 @@ describe('outputarea/widget', () => {
           const factory = new OutputArea.ContentFactory();
           const future = kernel.requestExecute({ code: CODE });
           const options = {
+            parent_header: {
+              date: '',
+              msg_id: '',
+              msg_type: 'input_request',
+              session: '',
+              username: '',
+              version: ''
+            },
             prompt: 'hello',
             password: false,
             future
--- a/packages/services/test/kernel/ikernel.spec.ts
+++ b/packages/services/test/kernel/ikernel.spec.ts
@@ -97,7 +97,17 @@ describe('Kernel.IKernel', () => {
       defaultKernel.pendingInput.connect((sender, args) => {
         if (!called) {
           called = true;
-          defaultKernel.sendInputReply({ status: 'ok', value: 'foo' });
+          defaultKernel.sendInputReply(
+            { status: 'ok', value: 'foo' },
+            {
+              date: '',
+              msg_id: '',
+              msg_type: 'input_request',
+              session: '',
+              username: '',
+              version: ''
+            }
+          );
         }
       });
       const code = `input("Input something")`;
@@ -306,7 +316,17 @@ describe('Kernel.IKernel', () => {
           expect(direction).toBe('send');
         }
       });
-      kernel.sendInputReply({ status: 'ok', value: 'foo' });
+      kernel.sendInputReply(
+        { status: 'ok', value: 'foo' },
+        {
+          date: '',
+          msg_id: '',
+          msg_type: 'input_request',
+          session: '',
+          username: '',
+          version: ''
+        }
+      );
       await emission;
     });
   });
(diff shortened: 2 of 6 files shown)


packages/outputarea/test/widget.spec.ts
```diff
@@ -376,6 +376,14 @@ describe('outputarea/widget', () => {
           const factory = new OutputArea.ContentFactory();
           const future = kernel.requestExecute({ code: CODE });
           const options = {
+            parent_header: {
```
jasongrout, line 379: This doesn't have any effect, right? The options passed to createStdin are very well specified, and do not include the parent header (and the parent header should be the header from the request message anyway).

It seems like the kernel connection sendInputReply method needs an option for the parent header, and then that needs to be threaded through calls to make its way down to it.

packages/services/test/kernel/ikernel.spec.ts
```diff
@@ -97,7 +97,17 @@ describe('Kernel.IKernel', () => {
       defaultKernel.pendingInput.connect((sender, args) => {
         if (!called) {
           called = true;
-          defaultKernel.sendInputReply({ status: 'ok', value: 'foo' });
+          defaultKernel.sendInputReply(
+            { status: 'ok', value: 'foo' },
+            {
```
jasongrout, line 102: sendInputReply doesn't take a second argument, so I don't think passing a second argument in is going to do anything.

jasongrout:
> The [kernel messaging protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#parent-header) says that "_reply messages MUST have a parent_header". This PR adds the parent header to input_reply messages, that was missing.

+1 to adding the parent header to reply messages. However, I think we actually need to change sendInputReply to take the parent header and then thread that info down to calls to sendInputReply.

packages/outputarea/test/widget.spec.ts
davidbrochart: > The options passed to createStdin are very well specified, and do not include the parent header

I added `parent_header` to the options passed to `createStdin` [here](https://github.com/jupyterlab/jupyterlab/pull/12376/files#diff-4d0f96dfb060e341e538a737407cb92ad9a1003d88037f844a289e11134837abR954), but I'm not sure it's the way to go.

packages/services/test/kernel/ikernel.spec.ts
davidbrochart: Yes, I added a second argument `parent_header` to `sendInputReply` [here](https://github.com/jupyterlab/jupyterlab/pull/12376/files#diff-1acd3b371270660fed12e4595862cfb74ab7e973dd853b9caac92dcb34b6ca99R821).

packages/outputarea/test/widget.spec.ts
```diff
@@ -376,6 +376,14 @@ describe('outputarea/widget', () => {
...
+              date: '',
+              msg_id: '',
+              msg_type: 'input_request',
```
davidbrochart, line 382: Not sure why `msg_type: 'input_request'` generates this error:
```
Error: test/widget.spec.ts(391,38): error TS2345: Argument of type '{ parent_header: { date: string; msg_id: string; msg_type: string; session: string; username: string; version: string; }; prompt: string; password: boolean; future: IShellFuture<IExecuteRequestMsg, IExecuteReplyMsg>; }' is not assignable to parameter of type 'IOptions'.
  The types of 'parent_header.msg_type' are incompatible between these types.
    Type 'string' is not assignable to type '"input_request"'.
```
I know it's related to [StdinMessageType](https://github.com/jupyterlab/jupyterlab/blob/ffbb016cd8d072654d00285bf0404cafa627a476/packages/services/src/kernel/messages.ts#L214), but I don't know why it is not accepted.

packages/services/test/kernel/ikernel.spec.ts
jasongrout: Ah, sorry, maybe I was just looking at one commit (from an email notification) rather than the whole PR. I'll rescind my comments and make sure I'm looking at the whole PR.

fcollonval (approved):
Thanks @davidbrochart 

When backporting, I will change the signature to have `parent_header` optional as we must not break API.

fcollonval:

PR #12398, opened 2022-04-12
Add ability to open settings editor to specific plugin's settings

### LINKED ISSUE #12180: Add command to open settings editor to specific plugin / search value
### Problem
There are many settings available, and if extensions want to link to specific plugin settings, they can only link to the entire settings editor. 

### Proposed Solution
Add an optional arg to the open settings editor command that allows for specifying a plugin settings to open or specific search results.

marthacryan:
## References
Fixes #12180. 

**Note:** For the purpose of reviewing this, this includes an example of how this could be used by an extension but I plan to remove the example once this has received some review. 

## Code changes
Adds an extra arg to the command to open the settings editor to specify a specific plugin's settings to open. Also adds an arg to specify the label when invoked (in case the caller wants to add a label like "Open <plugin> settings". 

## User-facing changes
Note: wouldn't change any user's experience as-is, this is for extensions that might want to create a button or palette option to open their plugin's settings. 

This opens the settings editor with only the specified plugin's settings displayed in the settings panel. All plugins and the search bar are still displayed in the left side panel and if the user clicks them, the settings panel shows all plugins again and is reset to the regular editor. 

None

### DIFF
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -17,5 +17,5 @@ test('Open the settings editor with a specific search query', async ({
     await page.locator('.jp-PluginList .jp-FilterBox input').inputValue()
   ).toEqual('Command Palette');
 
-  expect(await page.locator('.jp-SettingsForm')).toHaveCount(1);
+  await expect(page.locator('.jp-SettingsForm')).toHaveCount(1);
 });
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -2,13 +2,15 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { test } from '@jupyterlab/galata';
+import { expect } from '@playwright/test';
 
 test('Open the settings editor with a specific search query', async ({
   page
 }) => {
-  await window.jupyterapp.commands.execute('settingeditor:open', {
-    label: 'Open command palette settings',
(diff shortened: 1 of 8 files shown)


fcollonval (changes requested):
I would not implement that feature like that because it makes the UI inconsistent.

Now that #12082 is in, the command argument should set the filter on the settings and it should expand the first plugin of settings (if it exists).

It will also be good to add that feature to the JSON editor.

fcollonval (changes requested):
Thanks @marthacryan I have two additional comments

galata/test/jupyterlab/settings.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    await window.jupyterapp.commands.execute('settingeditor:open', {
+      query: 'Command Palette'
+    });
+  });
+
+  expect(
+    await page.locator('.jp-PluginList .jp-FilterBox input').inputValue()
+  ).toEqual('Command Palette');
+
+  expect(await page.locator('.jp-SettingsForm')).toHaveCount(1);
```
fcollonval, line 20:
```suggestion
  await expect(page.locator('.jp-SettingsForm')).toHaveCount(1);
```

marthacryan:
@fcollonval Any idea what's happening with the CI failures? Looks like there's an issue with cell toolbar?

fcollonval (approved):
Thanks @marthacryan

fcollonval:

PR #12407, opened 2022-04-14
Fix GitHub link

fcollonval:
Fix link_check

See for example: https://github.com/jupyterlab/jupyterlab/runs/6020621739


fcollonval:

PR #12411, opened 2022-04-15
Uses dark theme for Vega when JupyterLab theme is dark

### LINKED ISSUE #10617: VegaLite theme doesn't track JupyterLab theme
## Description

VegaLite allows themes to be provided, but we are currently not using them in a way that syncs the theme of VegaLite with that of JupyterLab. Thus, when a user changes themes in JupyterLab, VegaLite charts are not themed appropriately.

## Reproduce

1. Create a VegaLite based chart by opening a `.vl` file in the file browser.
2. Change to the dark theme.

## Expected behavior

The VegaLite theme should adjust to the light and dark backgrounds of JupyterLab.

## Context

We should probably enable a JupyterLab theme to provide the data for the VegaLite theme that is should be used. That way a JupyterLab theme can provide all the information required.

jweill-aws:
## References

Fixes #10617.

## Code changes

Modifies the `vega5-extension` to use a dark theme when the active JupyterLab theme is dark. Changing the theme _does not_ re-render all cells due to concerns about performance and side effects.

## User-facing changes

Rendered content with light theme:
image

After switching to dark theme, but before rerendering:
image

After rerendering in dark theme:
image

None.


jtpio:
Thanks!

Should we put this under visual regression testing with a new reference snapshot?

jweill-aws:

jweill-aws:
@jtpio I tried adding tests to run Vega code, but the output area when run from within Galata is empty on both my local machine and when run remotely. See screen shots below from Linux runs via GitHub CI.

run-cells-vega-actual

dark-theme-vega-actual

fcollonval:
@jweill-aws this is probably due to the loading time of the Javascript for vega to inject the graph. You can add a statement like

```
await page.waitForSelector('<a meaningful selector for the graph>')
```

before taking the screenshot; see [selectors documentation](https://playwright.dev/docs/selectors).

> It would also be better to take the screenshot only for the graph. So the test is more robust to change in input cell modification.

fcollonval:
Looking at the videos, typing the code is really slow and this is a lot of code - so I think you are hitting the 1min limit for a test. So I would change a bit the logic and upload a notebook that contains the cell with the needed code like that:

https://github.com/jupyterlab/jupyterlab/blob/f087c5207b62051b4fdca60674c41c041edaed23/galata/test/galata/contents.spec.ts#L27

jweill-aws:

jweill-aws:
@fcollonval Thanks! This commit has updated the screen shots (both input and output) for light and dark mode visuals: https://github.com/jupyterlab/jupyterlab/pull/12411/commits/5f910f32702bbe1c1cb24f4a13a334df56109ef6

fcollonval:
@jweill-aws this looks good but could you remove the old snapshots?

fcollonval (approved):
Thanks @jweill-aws

fcollonval:

PR #12288, opened 2022-03-29
Remove Yjs room locks

davidbrochart:
## Code changes

Remove the Yjs room locks.

## User-facing changes

None

## Backwards-incompatible changes

Rename `WebSocketProviderWithLocks` to `WebSocketProvider` in `@jupyterlab/docprovider`.

PR #12427, opened 2022-04-18
In RTC, document save operation causes deadlock

a3626a:
## Introduction
In Jupyter Lab 3 with RTC, document save operation causes deadlock, and further save attempts are ignored (deadlocked)
(I check for version 3.2, 3.3 (base branch for this  PR))

### Steps to reproduce

1. Launch Jupyter Lab with RTC
`jupyter lab --collaborative`

2. Open notebook document

3. Try to save the document very frequently. Click + (Ctrl+S)(shortcut) helps a lot to cause deadlock.
(try 10 clicks / second)

4. From now on, save operation does nothing. Autosave, shortcut keys, button, everything.

Status on the bottom("saving completed") is very useful. When deadlocked, no more messages are noticed.
image

### Investigation

Save operation is defined here. `acquireLock()` and `releaseLock()`  is called around the actual save operation
https://github.com/jupyterlab/jupyterlab/blob/7d43b7ad23460a80f80a21c00a00d74c19b88008/packages/docregistry/src/context.ts#L298-L317

`acquireLock` and `releaseLock` is defined here.
https://github.com/jupyterlab/jupyterlab/blob/7d43b7ad23460a80f80a21c00a00d74c19b88008/packages/docprovider/src/yprovider.ts#L155-L181

https://github.com/jupyterlab/jupyterlab/blob/7d43b7ad23460a80f80a21c00a00d74c19b88008/packages/docprovider/src/yprovider.ts#L183-L198

`this._sendMessage(new Uint8Array([127]));` actually causes the promise to release the thread. And it is periodically executed by `this._requestLockInterval`. I found that `this._requestLockInterval` is not well managed.

In some cases, `releaseLock()` can accidently clear `this._requestLockInterval`  which is created by other thread. Check this chart.
```
Timeline
================================================
Thread State
Thread 1             |      | acquireLock() |               | releaseLock() |
Thread 2             |      |               | acquireLock() |               | deadlocked (interval is not registed, nothing can release it)
================================================
Memory State 
_requestLockInterval | null | number        | number        | number        |
intervalRegistered   | no   | yes                           |  no           |
================================================
```

## Code changes

Prevent deadlock by counting request lock number

## User-facing changes

Deadlock on save is very annoying bug. fixed :)

No

## General Opinion

Lock implementation in yprovider looks very fragile. I already found that it is migrated to server in jupyter lab 4. So putting too much effort on this feels like a waste. So I don't reimplment all yprovider, just introduce simple variable to fix the bug.

### DIFF
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -95,6 +95,8 @@ export class WebSocketProviderWithLocks
     this._isInitialized = false;
     this._onConnectionStatus = this._onConnectionStatus.bind(this);
     this.on('status', this._onConnectionStatus);
+
+    this._requestLockNumber = 0;
   }
 
   setPath(newPath: string): void {
@@ -158,6 +160,8 @@ export class WebSocketProviderWithLocks
    * Returns a Promise that resolves to the lock number.
    */
   acquireLock(): Promise<number> {
+    this._requestLockNumber += 1;
+
     if (this._currentLockRequest) {
       return this._currentLockRequest.promise;
     }
@@ -193,7 +197,9 @@ export class WebSocketProviderWithLocks
     encoding.writeUint32(encoder, lock);
     // releasing lock
     this._sendMessage(encoding.toUint8Array(encoder));
-    if (this._requestLockInterval) {
+
+    this._requestLockNumber -= 1;
+    if (this._requestLockNumber === 0 && this._requestLockInterval) {
       clearInterval(this._requestLockInterval);
     }
   }
@@ -246,6 +252,7 @@ export class WebSocketProviderWithLocks
     reject: () => void;
   } | null = null;
   private _initialContentRequest: PromiseDelegate<boolean> | null = null;
+  private _requestLockNumber: number;
 }
 
 /**


davidbrochart:
Thanks @a3626a, the locking approach has always been problematic and is probably responsible for some issues like the one you are describing.
We are getting away from this locking approach, see:
- #12288
- #12360

So I'm not sure we should try and fix it, or wait for #12360 to get in.

a3626a:
> 

Thank you for your reply.
I hope RTC problems to be fixed in `3.x`. Will `y-py` related features be merged into `3.x`? If it will, this PR is redundant and I am ok to close this.

davidbrochart:
I don't know, maybe @dmonad has an opinion on this?

dmonad:
Thank you @a3626a for testing out the experimental RTC feature!

I think in one of our last PRs we broke the locking approach even further.  As David said, the plan is to move away from the locking approach. Any fixes for the locking approach would just be removed in a couple of weeks.  So I recommend to disable RTC for now until we have something working again.

aditya211935:
@dmonad @a3626a I know that this issue is no longer applicable after the removal of locks, but just out of curiosity wouldn't a better solution to this be only this extra condition on releaseLock()?

```
releaseLock(lock: number): void {
    const encoder = encoding.createEncoder();
    // reply with release lock
    encoding.writeVarUint(encoder, 126);
    encoding.writeUint32(encoder, lock);
    // releasing lock
    this._sendMessage(encoding.toUint8Array(encoder));
    if (this._requestLockInterval) {
      console.log("Lola: release2", this._requestLockInterval);
      clearInterval(this._requestLockInterval);
    }
    if (this._currentLockRequest) {
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore
      this._currentLockRequest.reject("Lola: Bye bye");
      this._currentLockRequest = null;
    }
  }
```

This would ensure that every time lock for the document is released, we also reject any _currentLockRequest and set it to null, as after `releaseLock()` is called, there's very little chance that _currentLockRequest is resolved (it can only be resolved if by chance the backend responded with `127` message). But by explicitly rejecting _currentLockRequest, we can prevent a deadlock scenario by failing the current save attempt.
How does this sound?

### Test
I verified this by setting absurdly large numbers to setInterval for `_requestLockInterval`, and by introducing an arificial delay in the finally block of Context's save() function like this 
```
 const finally_ = async () => {
      console.log("Lola: quik release ⚡️");
      await new Promise((res, rej) => setTimeout(res, 4000));
      this._provider.releaseLock(lock);
    };
```
Then tested it by waiting for `Lola: quik release ⚡️` message to appear and then immediately hitting save. Also, I removed the immediate `this._sendMessage(new Uint8Array([127]));`  on line 195. Thus by setting absurdly large number (10s) to `_requestLockInterval`, it'll ensure that we always wait for `releaseLock()` to clear intervals of the new lock request.

PR #12439, opened 2022-04-19
Removes FileEditorCodeWrapper

hbcarlos:
## References
Follow up:
- #12338
- #12381

## Code changes
Removes the FileEditorCodeWrapper class.

Removes the class FileEditorCodeWrapper, instead, you can use the CodeEditorWrapper from `@jupyterlab/codeeditor`.

### DIFF
--- a/packages/fileeditor/src/widget.ts
+++ b/packages/fileeditor/src/widget.ts
@@ -42,13 +42,13 @@ export class FileEditor extends Widget {
     const context = (this._context = options.context);
     this._mimeTypeService = options.mimeTypeService;
 
-    const editorWidget = (this.editorWidget = new CodeEditorWrapper({
+    const editorWidget = (this._editorWidget = new CodeEditorWrapper({
       factory: options.factory,
       model: context.model
     }));
-    this.editorWidget.addClass('jp-FileEditorCodeWrapper');
-    this.editorWidget.node.dataset[CODE_RUNNER] = 'true';
-    this.editorWidget.node.dataset[UNDOER] = 'true';
+    this._editorWidget.addClass('jp-FileEditorCodeWrapper');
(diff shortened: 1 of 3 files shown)


packages/fileeditor/src/widget.ts
```diff
@@ -242,11 +161,12 @@ export class FileEditor extends Widget {
     );
   }
 
-  private editorWidget: FileEditorCodeWrapper;
+  private editorWidget: CodeEditorWrapper;
   public model: CodeEditor.IModel;
   public editor: CodeEditor.IEditor;
   protected _context: DocumentRegistry.Context;
   private _mimeTypeService: IEditorMimeTypeService;
+  private _ready = new PromiseDelegate<void>();
```
fcollonval, line 169: Let's fix the visibility (`context` has public getter so no need to have `_context` protected as it should not be set outside the constructor) and follow the naming convention (`editorWidget` -> `_editorWidget`):

```suggestion
  model: CodeEditor.IModel;
  editor: CodeEditor.IEditor;
  private _context: DocumentRegistry.Context;
  private _editorWidget: CodeEditorWrapper;
  private _mimeTypeService: IEditorMimeTypeService;
  private _ready = new PromiseDelegate<void>();
```

fcollonval (changes requested):
Thanks @hbcarlos 

Would you mind applying the suggestion and adding a note in the [migration doc](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst)

hbcarlos:
Thanks for the review @fcollonval! I applied your suggestions, let me know if it is ok.

fcollonval (approved):
Thanks @hbcarlos

jtpio:
Thanks!

PR #12448, opened 2022-04-20
Use more explicit name for the releaser dist files uploaded as artifacts

jtpio:
## References

For now all dist files artifacts uploaded as part of the check release CI run are named `dist-files`:

image

This updates the name so it also includes `jupyterlab` and the GitHub run number. This is useful when downloading these files for this repo or other repos using the releaser, so it's easier to differentiate them.

## Code changes

CI workflow update.

## User-facing changes

None

None


fcollonval (approved):
Thanks @jtpio

Issue #12213, opened 2022-03-14
services makeSettings assumes a global declared process object


jasongrout:
## Description

In https://github.com/jupyterlab/jupyterlab/blob/7a2e386ff2e0fb9368cf8d79b18f4d0272eb5a0a/packages/services/src/serverconnection.ts#L247, we attempt to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to write code that works even if `process` is not defined, but as MDN states:

> Optional chaining cannot be used on a non-declared root object, but can be used with an undefined root object.

I think we should instead do something like the following, which checks first to make sure process is declared.

```typescript
((typeof process !== 'undefined') && process?.env?.JEST_WORKER_ID !== undefined)
```

## Reproduce

1. Use `@jupyterlab/services` makeSettings in an environment where `process` is not declared.
2. Get an "Uncaught ReferenceError: process is not defined"

## Expected behavior

We should be able to use `@jupyterlab/services` without defining the `process` global when in a context with a global window object.

## Context

- Operating System and version: 
- Browser and version: Chrome 99
- JupyterLab version: @jupyterlab/services 6.3.2 (jlab 3.3.2)

PR #12454, opened 2022-04-21
Check if process is declared before optional chaining in makeSettings

xiaochen-db:
This fixes https://github.com/jupyterlab/jupyterlab/issues/12213.

### DIFF
--- a/packages/services/src/serverconnection.ts
+++ b/packages/services/src/serverconnection.ts
@@ -244,7 +244,7 @@ namespace Private {
       appUrl: PageConfig.getOption('appUrl'),
       appendToken:
         typeof window === 'undefined' ||
-        process?.env?.JEST_WORKER_ID !== undefined ||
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks!

Looks like this needs a small link fix:

image

Which should have been fixed by `pre-commit.ci` normally, but it doesn't seem to be the case here.

xiaochen-db:
I pair-programed with @jasongrout and we tested the current `makeSettings` change with `process` being removed from the `builder/src/webpack.config.base.ts` (see https://github.com/jupyterlab/jupyterlab/pull/12461), and we found JupyterLab can still launch correctly. We decided to go with the current change in this pull request but remove `process` in future as it's deemed more controversial

fcollonval (approved):
LGTM

fcollonval:

PR #12290, opened 2022-03-29
Add git-blame-ignore-revs file

blink1073:
## References
Follow up to #12279

## Code changes
Ignore the auto-formatting commit for git blame.

## User-facing changes
Easier to browse git history.

## Backwards-incompatible changes
None

PR #10344, opened 2021-06-04
Add linter rule for sorting import

fcollonval:
## References

Fixes #10319

## Code changes

Add the following eslint rule

```javascript
'sort-imports': [
      'error',
      {
        ignoreCase: true,
        ignoreDeclarationSort: true,
        ignoreMemberSort: false,
        memberSyntaxSortOrder: ['none', 'all', 'multiple', 'single'],
        allowSeparatedGroups: false
      }
    ]
```

The file changes result from:
- applying VSCode sort import tools on the ~~project~~ _packages_ folder (except the _debugger_)
- applying the linter on the project.

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

PR #12458, opened 2022-04-23
Ignore the commit introducing import sort during blame

### LINKED ISSUE #10607: Create .git-blame-ignore-revs?
Git allows to ignore certain commits for blame. This could be handy for looking at history ignoring reformatting commits like https://github.com/jupyterlab/jupyterlab/pull/10344.

There is a very nice write up on this here: https://akrabat.com/ignoring-revisions-with-git-blame/

Currently everyone has to configure git to use this file locally; there is an open ticket for GitHub to support this naively so it would also work in the web UI: https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256/25

Edit: GitHub now supports it naively too: https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

Just bringing this up to ask if anyone else would use it if added.

krassowski:
## References

Closes #10607

## Code changes

Just added dbdefed9db9332381fe4104bdf53ec314451951f commit hash for #10344 as the file itself was already created in #12290

## User-facing changes

None

None

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,4 @@
 # Run pre-commit with black: https://github.com/jupyterlab/jupyterlab/pull/12279
 6df5e0b6e8ec740ffa2f84f198820d1968e4d74d
+# Add linter rule for sorting import: https://github.com/jupyterlab/jupyterlab/pull/10344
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
LGTM

PR #9636, opened 2021-01-18
fix: use process/browser module as real polyfill

maartenbreddels:
For the ipywebrtc/jupyter-webrtc widget, I need the real process polyfill. Would this be ok?

PR #12461, opened 2022-04-25
Issue 12213 remove process from webpack

xiaochen-db:

### DIFF
--- a/builder/src/webpack.config.base.ts
+++ b/builder/src/webpack.config.base.ts
@@ -2,7 +2,6 @@
 // Distributed under the terms of the Modified BSD License.
 
 import * as path from 'path';
-import * as webpack from 'webpack';
 import miniSVGDataURI from 'mini-svg-data-uri';
 
 const rules = [
@@ -98,8 +97,7 @@ module.exports = {
       url: false,
       buffer: false,
       // See https://github.com/webpack/webpack/blob/3471c776059ac2d26593ea39f9c47c1874253dbb/lib/ModuleNotFoundError.js#L13-L42
-      path: require.resolve('path-browserify'),
-      process: require.resolve('process/browser')
+      path: require.resolve('path-browserify')
     }
   },
   watchOptions: {
@@ -108,10 +106,5 @@ module.exports = {
   },
   output: {
     hashFunction: 'sha256'
-  },
-  plugins: [
-    new webpack.ProvidePlugin({
-      process: 'process/browser'
-    })
-  ]
+  }
 };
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks for this clean-up. Let me know when this is ready for review.

krassowski:
@xiaochen-db I know that it was a while (over 2 years!) but I just wanted to check if you still are interested in finishing this PR or if we should close it. As per https://github.com/jupyterlab/jupyterlab/pull/12454#issuecomment-1108118424 I understand that the immediate problem was solved, but you still wanted to remove `process` in the future (future is now).

jasongrout:
It looks like the change to serverconnection.ts was actually merged in https://github.com/jupyterlab/jupyterlab/pull/12454, but the change to webpack.config.base.ts was not merged.

Now that we've moved to rspack, I might try to do the webpack.config.base.ts change again and see what happens. I forget why it was controversial a few years ago.

jasongrout:
From https://github.com/jupyterlab/jupyterlab/pull/9636, it seems we needed the process polyfill because blueprint and the `path` polyfill required it.

However, I'm removing it now (and the path polyfill as well) and it appears everything is working:

```patch
diff --git a/builder/src/webpack.config.base.ts b/builder/src/webpack.config.base.ts
index 5b47f12b58..6e09e940db 100644
--- a/builder/src/webpack.config.base.ts
+++ b/builder/src/webpack.config.base.ts
@@ -1,6 +1,5 @@
-import * as rspack from '@rspack/core';
 import miniSVGDataURI from 'mini-svg-data-uri';
 
 const rules = [
@@ -72,10 +71,7 @@ module.exports = {
     fallback: {
       url: false,
       buffer: false,
-      crypto: false,
-      // See https://github.com/webpack/webpack/blob/3471c776059ac2d26593ea39f9c47c1874253dbb/lib/ModuleNotFoundError.js#L13-L42
-      path: require.resolve('path-browserify'),
-      process: require.resolve('process/browser')
+      crypto: false
     }
   },
   watchOptions: {
@@ -84,10 +80,5 @@ module.exports = {
   },
   output: {
     hashFunction: 'xxhash64'
-  },
-  plugins: [
-    new rspack.ProvidePlugin({
-      process: 'process/browser'
-    })
-  ]
+  }
 };
```

jasongrout:
I've created a new PR with just the removal of the polyfills from the webpack config at https://github.com/jupyterlab/jupyterlab/pull/18207. @xiaochen-db - I gave you co-author credit on that commit. I'll close this PR in favor of the updated PR.

PR #12462, opened 2022-04-25
Fix update snapshot for 3.4.x

fcollonval:
Workflow triggered on issues/PR comments are always from the default branch. This fixes the 3.4.x to be compatible with master workflow for updating snapshots.


fcollonval:
please update snapshots

fcollonval:
The update worked - the workflow finished with error because they were nothing to commit.

So merging this one.

Issue #12283, opened 2022-03-28
Settings editor should remember last mode


mlucool:
### Problem

Not everyone finds the form-based settings editor better than the JSON version. Those users need to click on JSON each time, adding a step to get to their editor of choice.

### Proposed Solution

The settings editor should remember if the user was in form-based or JSON-based last and open itself in that mode.

### Additional context

N/A

jweill-aws:
Triage notes: There is a command to open the JSON settings editor immediately, although the UI only includes a command to open the form-based editor. Accepted as an enhancement.

jtpio:
The default menu entry could be replaced to use the `settingeditor:open` command instead, via the settings system: https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html#settings-defined-menu

![image](https://user-images.githubusercontent.com/591645/165249884-630a082a-1f49-44e6-8bd2-9e840b06bf4f.png)

Issue #12340, opened 2022-04-04
More robust UI tests


fcollonval:
It would be nice to dump a npm request result for the extension manager tests and to serve it as mock answer for the tests to stabilize UI tests.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab/issues/12337#issuecomment-1087286925_

jtpio:
Looks like a similar failure is happening on `master` and other PRs:

For example on https://github.com/jupyterlab/jupyterlab/commit/d55621fa2d4d088be1ce767998c9712ad1663214: https://github.com/jupyterlab/jupyterlab/runs/6269416295?check_suite_focus=true

![image](https://user-images.githubusercontent.com/591645/166421165-5db6527f-9b47-497f-975b-b37f3a7043e7.png)

jtpio:
> mock answer for the tests to stabilize UI tests.

Reopening as it looks like this would still be needed.

dlqqq:
I think the snapshot needs another update; looks like the owner of @deathbeds/jupyterlab-drawio added a new icon.

https://user-images.githubusercontent.com/44106031/172951843-e905795a-0cba-4b81-81ec-e00f725e0da1.mov

dlqqq:
I'm not very familiar with Galata/Playwright, but it seems like this just requires us to stub out the network request fetching these search results via `page.route`: https://playwright.dev/docs/api/class-page#page-route

fcollonval:
Yes this is the way.

I did not do it because I would like to refactor that part of JupyterLab but have not found the time yet to finish it.

dlqqq:
@fcollonval See #12714. Do you get notifications if I submit a PR that will fix this issue? 👀

fcollonval:
> @fcollonval See #12714. Do you get notifications if I submit a PR that will fix this issue? 👀

Thanks @dlqqq I received notifications for all PRs. But feel free to request a PR review from me too when you need it.

PR #12466, opened 2022-04-25
Setting to use the advanced setting editor for the settings

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12283

## Code changes

Add a setting `useJsonEditor` in the `settingeditor-extension` package. If that setting is `true, the advanced json editor is use to edit the settings.

## User-facing changes

Untitled4

None.


fcollonval (changes requested):
I would use an enumeration instead of a boolean as setting; something like

Default settings editor: ['UI', 'json']

jtpio:
Also, is this new option needed? Or could the menu customization via the settings be used to replace the menu entry by `"command": "settingeditor:open-json"`?

echarles:
> I would use an enumeration instead of a boolean as setting; something like Default settings editor: ['UI', 'json']

Thx @fcollonval. Yes, we can do that.

> Also, is this new option needed? Or could the menu customization via the settings be used to replace the menu entry by "command": "settingeditor:open-json"?

Thx @jtpio. With  `option`, are you meaning the new setting, or the new command? I guess you mean the latter. I have thought replacing the menu entry (which is the visual menu show here below) based on the user setting ediotr (ui or json), but I wonder if it is possible to achieve the replace without the full browser refresh. Once the menu is built, is it mutable?

Screenshot 2022-04-26 at 10 23 48

fcollonval:
I like a lot the idea of @jtpio - but I think it is nice to still combine it with a boolean settings as menu customization is not easy for the end users.

So to layout the changes needed, you would not add a new command, but extend the `open` one to receive arguments.

If the argument defines the type of settings editor, use the argument otherwise fallback to the settings value. And I think that for 4.0 it will make sense to drop the command `'settingeditor:open-json'`. But when backporting this enhancement we should keep it around.

echarles:
> So to layout the changes needed, you would not add a new command, but extend the open one to receive arguments. If the argument defines the type of settings editor, use the argument otherwise fallback to the settings value. And I think that for 4.0 it will make sense to drop the command 'settingeditor:open-json'. But when backporting this enhancement we should keep it around.

@fcollonval I have implemented your suggestion in my latest commit. Maybe you could have a look and confirm this was the intent. If confirmed, I will then implement your suggestion `I would use an enumeration instead of a boolean as setting; something like Default settings editor: ['UI', 'json']`.

echarles:
> and I think that for 4.0 it will make sense to drop the command 'settingeditor:open-json'

I have not been able to achieve that as the execution of the `open` and `open-json` commands are defined in separated plugins.

packages/settingeditor-extension/src/index.ts
```diff
@@ -101,64 +101,77 @@ function activate(
...
+      });
     },
     label: trans.__('Settings Editor')
   });
 
+  if (palette) {
+    palette.addItem({
+      category: trans.__('Settings'),
+      command: CommandIDs.open,
+      args: { useJsonEditor: true }
```
fcollonval, line 171:
```suggestion
      args: { useJsonEditor: false }
```

packages/settingeditor-extension/src/index.ts
```diff
@@ -101,64 +101,77 @@ function activate(
...
-      void tracker.add(editor);
-      shell.add(editor);
+  commands.addCommand(CommandIDs.open, {
+    execute: async ({ useJsonEditor }: { useJsonEditor: boolean }) => {
+      registry.load(plugin.id).then(settings => {
+        if (useJsonEditor) {
+          commands.execute(CommandIDs.openJSON);
+          return;
+        }
+        (settings.get('useJsonEditor').composite as boolean)
```
fcollonval, line 159: That's the spirit

```suggestion
    execute: async ({ useJsonEditor }: { useJsonEditor?: boolean }) => {
      registry.load(plugin.id).then(settings => {
        useJsonEditor ?? (settings.get('useJsonEditor').composite as boolean)
```

fcollonval (commented):
Thanks @echarles  yes this is the idea.

echarles:
@fcollonval I have merged with latest master and resolved the conflicts.

echarles:
Galata is failing on `Open the settings editor with a specific search query `

I am not familiar with the recent setting search query. I run `openUi({ query: '' });` What should I do?

echarles:
Did a rebase on latest master and UI test is now green. A few other flows are failing mostly on splice feature but should not berelated to this PR.

echarles:
The UI regression test fails on ` Extension Manager › Search` which is not related to this PR.

@fcollonval I think this is ready to be merged.

jtpio:
> The UI regression test fails on ` Extension Manager › Search` which is not related to this PR.

Yes also noticed in other places: https://github.com/jupyterlab/jupyterlab/issues/12340#issuecomment-1115833463

echarles:
Just solved recent conflicts with master and CI is still green (exception of the not-revelant Extension Manager › Search).

echarles:
@fcollonval @jtpio I have rebased and solved the conflicts. Happy to get your reviews so we can merge and backport.

fcollonval (approved):
Thanks @echarles LGTM

> CI failure is not related

echarles:

echarles:
Thx @fcollonval 

The backport PR is https://github.com/jupyterlab/jupyterlab/pull/12529

Issue #12160, opened 2022-03-03
v3.3.0 settings new UI


miwojc:
## Description

The new UI for settings (Advanced settings editor). When i type a font name into the editor after each character it saves the file and exits the input field. So i have to repeat the process for each character or the font.

## Reproduce

1. Go to `Settings` -> `Advanced Settings Editor`
2. Click on `Notebook`
3. Scroll down to `Font Family`
4. Start typing in the Font Family input filed
5. After each character it saves the settings and exits the input field

Was changing font.

## Expected behavior

It should be possible to enter the full font name in one go.

## Context

- Operating System and version:  macOS 12.0.1
- Browser and version:  Firefox 97.0.1
- JupyterLab version:  3.3.0

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

mlucool:
All fields have this problem, which makes the new UI pretty unusable for input boxes (numbers, strings).

echarles:
@marthacryan The focus lost cause is the rerendering of the React.js component onChange. Disable the call to `setState` fixes the focus lost, the ui is still updated, but the save does not work anymore as the state does not contain the changes. I'd like to dig more into searching a solution, and will be very happy to get your thoughts and advices on this as you have implemented that feature with the reac-json-schema library. The focus lost in case of rerending is a know issue, see e.g. https://stackoverflow.com/questions/52924447/custom-widget-with-input-loses-focus-in-react-jsonschema-form-when-formdata-is-p

echarles:
I am trying to find the currently focussed widget an put it in the state, to focus on it after. Nor `document.querySelector(":focus");` nor `document.activeElement` work for now. (`document.activeElement` returns <body...> which looks like a well-known issue)

echarles:
Maybe a `onChange` handler on the props could avoid using the react state.

echarles:
@marthacryan I have opened https://github.com/jupyterlab/jupyterlab/pull/12470 to fix this.

PR #12470, opened 2022-04-26
Setting form editor has a formState to avoid focus lost

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12160

## Code changes

Use a dedicated class field to maintain the form state instead of the standard react state. This avoid rerendering and focus lost.

## User-facing changes

Focus is no more lost.

Untitled4

None.

### DIFF
--- a/packages/settingeditor/src/SettingsFormEditor.tsx
+++ b/packages/settingeditor/src/SettingsFormEditor.tsx
@@ -332,10 +332,11 @@ export class SettingsFormEditor extends React.Component<
   SettingsFormEditor.IProps,
   SettingsFormEditor.IState
 > {
+  private formState: any;
   constructor(props: SettingsFormEditor.IProps) {
     super(props);
     const { settings } = props;
-    this.state = {
+    this.formState = {
       formData: settings.composite,
       isModified: settings.isModified
     };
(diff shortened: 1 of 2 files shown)


packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -332,10 +332,11 @@ export class SettingsFormEditor extends React.Component<
   SettingsFormEditor.IProps,
   SettingsFormEditor.IState
 > {
+  private formState: any;
   constructor(props: SettingsFormEditor.IProps) {
     super(props);
     const { settings } = props;
-    this.state = {
+    this.formState = {
```
fcollonval, line 339: The formData pass to `Form` is indeed only the [initial value](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/form-props/#formdata). So it makes sense to extract `formData` from the state. However `isModified` should be kept in the state otherwise the _Restore to Defaults_ button does not appear when needed.

fcollonval (changes requested):

packages/settingeditor/src/SettingsFormEditor.tsx
echarles: Putting back `isModified` in the react state generate rendering and looses the focus (the **Restore to Defaults** button is show again).

I have tried alternative approaches https://github.com/jupyterlab/jupyterlab/issues/12160#issuecomment-1108897013 without success.

fcollonval: I opened https://github.com/datalayer-externals/jupyterlab/pull/3 with a better approach that only update props appropriately avoiding unnecessary component rerendering that were creating the focus lost.

echarles: Awesome. I have merged your changes. Thx a bunch for this.

fcollonval (changes requested):

packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -88,13 +88,40 @@ export namespace SettingsFormEditor {
     /**
      * The current form values being displayed in the editor.
      */
-    formData: any;
+    // formData: any;
```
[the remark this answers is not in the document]
echarles: Done. Thx.

fcollonval (approved):
Thanks @echarles

echarles:

echarles:
@fcollonval I am now testing master branch with this merge, and I face an issue. The settings editor does not open anymore (the json editor does). 

I wonder how this is related to https://github.com/jupyterlab/jupyterlab/issues/12485 As mentioned on that issue, I faced the problem once those last days, but today it is 100% reproducible.

It looks like this code block is responsible for the UI not showing up. If I comment-out the block the editor is shown correctly (with the focus being not lost, which is good). If I refresh the page, the editor opens back again.

```
      if (tracker.currentWidget) {
        shell.activateById(tracker.currentWidget.id);
        return;
      }
```

PR #12493, opened 2022-04-29
Inverse dependencies link between translation and rendermime interfaces

### LINKED ISSUE #11748: Don't bump version for @jupyterlab/rendermime-interfaces
## Description

The JS packages version bump to v4 is probably to severe. In particular, the rendermime interfaces did not change. And that package is very likely to be used by numerous downstream extensions. It would be good to not bump it.

> The bump occured because it explicitly depends on @jupyterlab/translation. We should probably cut that dependency (as I think the idea was to have this package provide a long-term stable API) as this is done for all the other interfaces (like `ISanitizer`, `IRenderMime`, ...). And we should depends on this package in the other ones to ensure we are not breaking those stable interfaces (for now we do not check if we break or not the interfaces copied in `rendermime-interfaces`...).

## Expected behavior

We should be able to install extension like plotly that provides a rendermime extension on JupyterLab v4 without requiring an update of the extension.

fcollonval:
## References

Fixes #11748

## Code changes

The rendermime interfaces package should be independent of the other packages.

Remove `IRenderer.translator?` as a renderer has only two contracts: 
- Be a `Widget`
- Have a `renderModel` method

> It receives the translator object at construction if it needs it.

## User-facing changes

None

Remove `IRenderer.translator?`

### DIFF
--- a/packages/rendermime-interfaces/src/index.ts
+++ b/packages/rendermime-interfaces/src/index.ts
@@ -7,7 +7,6 @@
  * @module rendermime-interfaces
  */
 
-import { ITranslator } from '@jupyterlab/translation';
 import { ReadonlyPartialJSONObject } from '@lumino/coreutils';
 import { Widget } from '@lumino/widgets';
 
@@ -278,11 +277,6 @@ export namespace IRenderMime {
    */
   export interface IRenderer extends Widget {
     /**
-     * The application language translator.
-     */
-    translator?: ITranslator;
-
-    /**
      * Render a mime model.
      *
      * @param model - The mime model to render.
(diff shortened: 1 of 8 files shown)


packages/rendermime-interfaces/src/index.ts
```diff
@@ -450,4 +444,157 @@ export namespace IRenderMime {
...
   }
+
+  // ********************** //
+  // Translation interfaces //
+  // ********************** //
+
+  /**
+   * Bundle of gettext-based translation functions for a specific domain.
+   */
+  export type TranslationBundle = {
```
jtpio, line 455: Was this snippet copied manually?

Wondering whether there is a risk of it going out of sync later. Although it's probably unlikely this type will be changing any time soon, but it could.

Maybe we should add some logic to the integrity script to make sure they don't drift apart. Haven't looked into how feasible this is, but we could also track that separately if it's too much effort.

jtpio:
Thanks @fcollonval :+1: 

Should we also "unbump" the package, so it's back to `^3.x.x`?

packages/rendermime-interfaces/src/index.ts
fcollonval: As `@jupyterlab/translation::TranslationBundle` is proxying this one (as well as `ITranslator`). I don't think there is a risk of divergence.

packages/rendermime-interfaces/src/index.ts
```diff
@@ -277,11 +276,6 @@ export namespace IRenderMime {
    * A widget which displays the contents of a mime model.
    */
   export interface IRenderer extends Widget {
-    /**
-     * The application language translator.
-     */
-    translator?: ITranslator;
```
fcollonval, line 283: If this change is fine to left the package at v3.x, I'm definitely :100: for avoiding downstream mimetype extension to require bumping their package for JLab v4.

I'll raise the question at the next meeting.

packages/rendermime-interfaces/src/index.ts
jtpio: Ah indeed looks good :+1:  The declaration has been updated to:

```ts
export interface ITranslator extends IRenderMime.ITranslator {}
```

packages/rendermime-interfaces/src/index.ts
jtpio: Sounds good.

jtpio (approved):
Thanks!

Issue #11897, opened 2022-01-18
Error is thrown when Dialog has no buttons


mnowacki-b:
## Description
Seems there are insufficient checks being performed in `apputils` Dialog class. When a dialog with no buttons is created it results in an error. The dialog is still visible but in the console you can see an error log.

## Reproduce

1. Create a new dialog without any buttons:
```ts
import {]
  Dialog
} from '@jupyterlab/apputils';

const dialog = new Dialog({
    title: 'title',
    body: createPreformattedTextWidget('body'),
    buttons: [],
    hasClose: false,
  });
```
2. launch the dialog:
```ts
dialog.launch()
```
3. Observe the console log:
error:
```
TypeError: Cannot read properties of undefined (reading 'focus')
    at Dialog.onAfterAttach (dialog.js:212:1)
    at Dialog.Widget.processMessage (index.es6.js:1445:1)
    at invokeHandler (index.es6.js:444:1)
    at Object.sendMessage (index.es6.js:180:1)
    at Function.attach (index.es6.js:1838:9)
    at dialog.js:130:13
 ```

## Expected behavior

There should be no errors.

## Context

- Operating System and version: macOS Big Sure
- Browser and version: Chrome 97.0.4692.7
- JupyterLab version: 3.2.5

**Comment by mnowacki-b** [comment]
Maybe adding a check here:
https://github.com/jupyterlab/jupyterlab/blob/80a00385b776ce2bafbac0e65e5ef6aa36421023/packages/apputils/src/dialog.tsx#L252
to see if `_primary` is set, ignoring the focus if it is not set?
```ts
if (this._primary) {
  this._primary.focus();
}
```
Or perhaps here:
https://github.com/jupyterlab/jupyterlab/blob/80a00385b776ce2bafbac0e65e5ef6aa36421023/packages/apputils/src/dialog.tsx#L123
if no button is present default to focus on the body?
```ts
this._primary = this._buttonNodes[this._defaultButton] || body;
```

mnowacki-b:
Another area that would require some change is here:
https://github.com/jupyterlab/jupyterlab/blob/80a00385b776ce2bafbac0e65e5ef6aa36421023/packages/apputils/src/dialog.tsx#L395

anngedeus:
Hello, I would like to work on this issue. If it is still available

mnowacki-b:
I was planing on working on it but haven't started yet. So if you like to work on it go for it.

[...]

PR #12496, opened 2022-04-29
Handled new dialog creation with no buttons

Jnnamchi:
NOTE: The git pre-commit and pre-push scripts fail at the flake8 installation with the following error:
```
stderr:
      error: subprocess-exited-with-error

      × python setup.py egg_info did not run successfully.
      │ exit code: 1
      ╰─> [40 lines of output]
          /Users/jnnamchi/.cache/pre-commit/repotdj18xwv/py_env-default/lib/python3.10/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
```

## References

Pull request related to the following issue: https://github.com/jupyterlab/jupyterlab/issues/11897

## Code changes

Added handling to dialog class when initialized with no buttons. Code checks the presence of a primary button before invoking the focus() method on it, if defined.

## User-facing changes

No visual changes.

No backwards incompatible changes

### DIFF
--- a/packages/apputils/src/dialog.tsx
+++ b/packages/apputils/src/dialog.tsx
@@ -249,9 +249,7 @@ export class Dialog<T> extends Widget {
         this._primary = el as HTMLElement;
       }
     }
-    if (this._primary) {
-      this._primary.focus();
-    }
+    this._primary?.focus();
   }
 
   /**
@@ -394,9 +392,7 @@ export class Dialog<T> extends Widget {
     const target = event.target as HTMLElement;
     if (!this.node.contains(target as HTMLElement)) {
       event.stopPropagation();
-      if (this._buttonNodes && this._buttonNodes[this._defaultButton]) {
-        this._buttonNodes[this._defaultButton].focus();
-      }
+      this._buttonNodes[this._defaultButton]?.focus();
     }
   }
 
--- a/packages/apputils/src/dialog.tsx
+++ b/packages/apputils/src/dialog.tsx
@@ -249,7 +249,9 @@ export class Dialog<T> extends Widget {
         this._primary = el as HTMLElement;
       }
     }
-    this._primary.focus();
+    if (this._primary) {
+      this._primary.focus();
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/apputils/src/dialog.tsx
```diff
@@ -392,7 +394,9 @@ export class Dialog<T> extends Widget {
     const target = event.target as HTMLElement;
     if (!this.node.contains(target as HTMLElement)) {
       event.stopPropagation();
-      this._buttonNodes[this._defaultButton].focus();
+      if (this._buttonNodes && this._buttonNodes[this._defaultButton]) {
+        this._buttonNodes[this._defaultButton].focus();
+      }
```
fcollonval, line 399: `this._buttonNodes` is always defined although it may be an empty array in this case. So we could take advantage of the nullish operator:

```suggestion
      this._buttonNodes[this._defaultButton]?.focus();
```

packages/apputils/src/dialog.tsx
```diff
@@ -249,7 +249,9 @@ export class Dialog<T> extends Widget {
         this._primary = el as HTMLElement;
       }
     }
-    this._primary.focus();
+    if (this._primary) {
+      this._primary.focus();
+    }
```
fcollonval, line 254: We can take advantage of the nullish operator

```suggestion
    this._primary?.focus();
```

FYI, in JavaScript this will be translated to the condition you have explicitly written.

fcollonval (approved):
Thanks @Jnnamchi 

I added two styling suggestions but otherwise it looks good.

> CI failure can be ignored.

fcollonval:
Kicking the CI after fix

Jnnamchi:
@fcollonval let me know if you need anything else from me! I do not have merging priviledges. This is my first open source contribution so Im not fully aware of the whole process.

fcollonval (approved):
Thanks @Jnnamchi and congrats on the first PR.

fcollonval:

Issue #12478, opened 2022-04-26
The file browser checkbox is checked by default when there is no file 


jtpio:
## Problem

When `showFileCheckboxes` is turned on it shows the box checked by default.

Using JupyterLab `4.0.0a24`:

![image](https://user-images.githubusercontent.com/591645/165350128-b4956954-292f-4241-b3ed-c88a17db52b2.png)

## Proposed solution

Maybe the default should be to leave the box unchecked starting from an empty directory?

For reference the classic notebook leaves the box empty when there is no files:

![image](https://user-images.githubusercontent.com/591645/165349829-ff48a5dd-38ae-4b1c-b0ef-ae7b62ca24f0.png)

## Context

Checkboxes were added in https://github.com/jupyterlab/jupyterlab/pull/12299 as opt-in.

cc @gabalafou who might have some input on this, thanks!

PR #12510, opened 2022-05-03
Check-all checkbox should be unchecked when current directory is empty

gabalafou:
## References

Fixes issue #12478.

## Code changes

Fixes some logic (edge case error), adds regression test.

## User-facing changes

See issue  #12478.

Same as PR #12299.


fcollonval:
Kicking the CI

fcollonval (approved):
Thanks @gabalafou 

> CI failure is not related

PR #12519, opened 2022-05-03
Make package importable in node.js

### LINKED ISSUE #12518: Use JupyterLab packages in node
### Problem

I would like to use JupyterLab packages in node. But if I run node with the following file (after installing `@jupyterlab/shared-models` with `npm`):
```js
import { YNotebook } from '@jupyterlab/shared-models'
```
I get:
```
(node:151939) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/david/github/davidbrochart/jupyter_ydoc/tests/foo.js:1
import { YNotebook } from '@jupyterlab/shared-models'
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
If I add `"type": "module"` in `package.json`, I get:
```
import { YNotebook } from '@jupyterlab/shared-models'
         ^^^^^^^^^
SyntaxError: Named export 'YNotebook' not found. The requested module '@jupyterlab/shared-models' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@jupyterlab/shared-models';
const { YNotebook } = pkg;
```
If I change the file with:
```js
import pkg from '@jupyterlab/shared-models'
```
I get:
```
(node:151807) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/david/github/davidbrochart/jupyter_ydoc/tests/node_modules/@jupyterlab/shared-models/lib/index.js:9
export * from './api';
^^^^^^

SyntaxError: Unexpected token 'export'
```
If I add `"type": "module"` in `node_modules/@jupyterlab/shared-models/package.json`, I get:
```
node:internal/errors:465
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/david/github/davidbrochart/jupyter_ydoc/tests/node_modules/@jupyterlab/shared-models/lib/api' imported from /home/david/github/davidbrochart/jupyter_ydoc/tests/node_modules/@jupyterlab/shared-models/lib/index.js
    at new NodeError (node:internal/errors:372:5)
    at finalizeResolution (node:internal/modules/esm/resolve:405:11)
    at moduleResolve (node:internal/modules/esm/resolve:964:10)
    at defaultResolve (node:internal/modules/esm/resolve:1173:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:604:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```
If I change `node_modules/@jupyterlab/shared-models/lib/index.js` to the following (adding `.js`):
```js
export * from './api.js';
export * from './ymodels.js';
export * from './utils.js';
```
I get:
```
import pkg from '@jupyterlab/shared-models'
       ^^^
SyntaxError: The requested module '@jupyterlab/shared-models' does not provide an export named 'default'
```
If I change back to:
```js
import { YNotebook } from '@jupyterlab/shared-models'
```
It works!

### Proposed Solution

It looks like the needed changes were:
- add `"type": "module"` to each package's `package.json`
- add `.js` to every `export * from './x';`

But I don't know what I'm doing :smile:

davidbrochart:
## References

Closes #12518.

## Code changes

- Update integrity check to add type: "module" for packages targeting esModule
  Add tool to add extension in import/export statement (will be better when tsx are covered too with https://github.com/antongolub/tsc-esm-fix/pull/75)
- Add `"type": "module"` to all esModule packages
- Add `.js` extension to import and export statements

- [x] Check CI validation works

## User-facing changes

None.

None.


davidbrochart:
This [Stack Overflow question](https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat) seems to confirm the needed changes.

davidbrochart:
Actually it looks like the `.js` is not necessary if we use node's [--experimental-specifier-resolution=node](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm). Thanks @wolfv !

fcollonval:
Thanks @davidbrochart 

I would like to raise the point at the weekly meeting as it could be nice to extend that feature to other packages.

For reference, TypeScript is working on some experimental feature related to that: https://www.typescriptlang.org/docs/handbook/esm-node.html

davidbrochart:
> Actually it looks like the `.js` is not necessary if we use node's [--experimental-specifier-resolution=node](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm). Thanks @wolfv !

@wolfv also mentioned this TypeScript plugin to automatically add `.js`: https://github.com/viT-1/systemjs-ts-es6/blob/master/src/tsconfig-dev.json

fcollonval:
We should not backport to 3.4.x as it changes the Jest configuration in @jupyterlab/testutils. That may break jest tests in extensions.

packages/application/src/connectionlost.ts
```diff
@@ -4,7 +4,7 @@
 import { showErrorMessage } from '@jupyterlab/apputils';
 import { ServerConnection, ServiceManager } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
-import { IConnectionLost } from './tokens';
+import { IConnectionLost } from './tokens.js';
```
jtpio, line 7: Looking at the diff, these `.js` suffixes looks a bit odd from `.ts` files.

fcollonval: There is no other alternatives offered by TypeScript for now.

fcollonval:
Typescript 4.7 is out with [ESM module support](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#ecmascript-module-support-in-node-js). It requires to add the extension of the transpiled file in the import statement.

jtpio:
> Probably we could also update to TS 4.7 in a separate PR.

Started in https://github.com/jupyterlab/jupyterlab/pull/12683

fcollonval:
> Wondering if they plan to support the `.ts` suffixes

I don't think it will happen as if it was the case they should have introduced it in 4.7. Moreover they are mentioning that the tooling like path completion and auto import will add the `.js` extension. So switching to `.ts` in a latter version will be quite annoying.

jtpio:
Quick update:

- JupyterLab `master` is now using TypeScript 3.7: https://github.com/jupyterlab/jupyterlab/pull/12683
- There might be more needs for this PR based on the feedback from https://github.com/jupyterlab/jupyterlab/issues/13429
- So we could update / rebase / fix conflicts for this PR and consider getting it in. And maybe discuss it one more time at one of the weekly meetings.

fcollonval:
Closing as stalled for now - we may revisit this at a later stage.

PR #12520, opened 2022-05-03
Make selected text translucent so the cursor is visible in vim mode

### LINKED ISSUE #8207: Vim cursor not visible on search highlights
The positioning of vim cursor (below the tokens) does not work great when highlights (achieved by means of background) are applied to tokens. This could be worked around by bumping the cursor z-index and making it semi-transparent or outline only (but neither is perfect..). Observed when investigating https://github.com/krassowski/jupyterlab-lsp/issues/240 (where it is a bigger problem).

I *think* that this is an upstream (codemirror) issue, but I do not see how they could fix this without a major refactor - the cost-benefit tradeoff does not look great. Just looking for ideas to address this here.

## Reproduce

1. Enable vim mode: Settings → Text Editor Key Map → vim
2. Search for a string in the file editor
3. Move cursor over the highlighted region
4. See the cursor obscured by the highlight

highlight_obscures_vim_cursor

## Expected behavior

The cursor remains visible at the top of the highlight, but below the token text.

## Context

- JupyterLab version: 2.1

z-index for the vim cursor is set in: `.cm-fat-cursor div.CodeMirror-cursors`,
```css
.cm-fat-cursor div.CodeMirror-cursors {
    z-index: 1;
}
```

fill for the vim cursor is set in `.cm-fat-cursor .CodeMirror-cursor`

```css
.cm-fat-cursor .CodeMirror-cursor {
    width: auto;
    border: 0 !important;
    background: #7e7;
}
```

Jessie-Newman:
## References

Fixes #8207

## Code changes

When text is highlighted in vim mode we set a background color to it. Because the cursor appears behind the background, the cursor is entirely hidden if it is on a highlighted character. This sets the background color to be slightly transparent so the cursor is still visible. Works in both light and dark mode.

## User-facing changes
Light mode
Screen Shot 2022-05-03 at 1 39 14 PM

Dark mode
Screen Shot 2022-05-03 at 1 39 01 PM

N/A

---

Edited to link the associated issue

### DIFF
--- a/packages/theme-dark-extension/style/variables.css
+++ b/packages/theme-dark-extension/style/variables.css
@@ -209,8 +209,7 @@ all of MD as it is not optimized for dense, information rich UIs.
    * theme these would go from dark to light.
    */
 
-  /* Separated rbg values to allow passing into rgba. */
-  --jp-inverse-layout-color0: 155, 155, 155; /* white */
+  --jp-inverse-layout-color0: white;
   --jp-inverse-layout-color1: white;
   --jp-inverse-layout-color2: var(--md-grey-200);
   --jp-inverse-layout-color3: var(--md-grey-400);
@@ -389,8 +388,8 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-search-toggle-on-opacity: 1;
   --jp-search-selected-match-background-color: rgb(255, 225, 0);
   --jp-search-selected-match-color: black;
-  --jp-search-unselected-match-background-color: rgba(
-    var(--jp-inverse-layout-color0), 0.8
+  --jp-search-unselected-match-background-color: var(
+    --jp-inverse-layout-color0
   );
   --jp-search-unselected-match-color: var(--jp-ui-inverse-font-color0);
 
--- a/packages/theme-dark-extension/style/variables.css
+++ b/packages/theme-dark-extension/style/variables.css
@@ -209,7 +209,8 @@ all of MD as it is not optimized for dense, information rich UIs.
    * theme these would go from dark to light.
    */
 
-  --jp-inverse-layout-color0: white;
+  /* Separated rbg values to allow passing into rgba. */
+  --jp-inverse-layout-color0: 155, 155, 155; /* white */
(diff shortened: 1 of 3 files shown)


Jessie-Newman:
The actual correct way to fix this would be to, instead of showing the cursor as a div that moves around the page and sits behind the text, to update the background-color on the selected character. This would give us more flexibility in terms of color. Rather than using a transparent background, we could have selected text and the cursor not overlay at all. We could also specify the color of the selected character (right now for example, if the green cursor is on a green character, it's mostly unreadable).

However the cursor logic is inside of the CodeMirror package and would require changes to that directly. Additionally, changing the cursor from a div to a character background-color would be a very extensive change to their code.

So this translucent background-color solution, while less than perfect, is much smaller and requires no changes to other codebases.

packages/theme-dark-extension/style/variables.css
```diff
@@ -209,7 +209,8 @@ all of MD as it is not optimized for dense, information rich UIs.
    * theme these would go from dark to light.
    */
 
-  --jp-inverse-layout-color0: white;
+  /* Separated rbg values to allow passing into rgba. */
+  --jp-inverse-layout-color0: 155, 155, 155; /* white */
```
fcollonval, line 213: We should not touch those generic variables as they may impact extensions styling. We should aim at something specific for the vim mode.

Looking at the DOM, when the VIM mode is activated the `cm-fat-cursor` is added to `CodeMirror cm-s-jupyter CodeMirror-wrap` in the top container.

I tried to add that CSS rule and it works as expected - feel free to tune the styling change this is more to point out a good CSS selector to fix the issue:

```css
.cm-fat-cursor .cm-overlay.cm-searching {
	opacity: 0.5;
}
```

fcollonval (changes requested):

fcollonval:
Thanks @Jessie-Newman I have a recommendation to fix this using a slightly different and more specific change.

fcollonval (approved):

fcollonval:

PR #12532, opened 2022-05-04
Add "Open in Simple Mode" contextMenu option

### LINKED ISSUE #12325: Context Menu option to open a notebook in simple mode
### Problem

Currently there is no easy method to open a notebook in simple mode (and in a new tab).
### Proposed Solution

Add a context menu option to open a notebook in simple mode (and in a new tab).

divyansshhh:
## References

Fix #12325 

## Code changes

Adds a new `openBrowserTabInSimpleMode` command and plugin.

## User-facing changes

This adds a "Open in Simple Mode" contextMenu option to
filebrowser-extension

---
Edited to link the issue

### DIFF
--- a/packages/filebrowser-extension/src/index.ts
+++ b/packages/filebrowser-extension/src/index.ts
@@ -101,6 +101,9 @@ namespace CommandIDs {
 
   export const openBrowserTab = 'filebrowser:open-browser-tab';
 
+  export const openBrowserTabInSimpleMode =
+    'filebrowser:open-browser-tab-in-simple-mode';
+
   export const paste = 'filebrowser:paste';
 
   export const createNewDirectory = 'filebrowser:create-new-directory';
@@ -635,25 +638,44 @@ const openBrowserTabPlugin: JupyterFrontEndPlugin<void> = {
     const { tracker } = factory;
 
     commands.addCommand(CommandIDs.openBrowserTab, {
-      execute: () => {
+      execute: args => {
         const widget = tracker.currentWidget;
 
         if (!widget) {
           return;
         }
 
+        const mode = args['mode'] as string | undefined;
+
         return Promise.all(
           toArray(
             map(widget.selectedItems(), item => {
-              return commands.execute('docmanager:open-browser-tab', {
-                path: item.path
-              });
+              if (mode === 'single-document') {
+                const url = PageConfig.getUrl({
+                  mode: 'single-document',
+                  treePath: item.path
+                });
+                const opened = window.open();
+                if (opened) {
+                  opened.opener = null;
+                  opened.location.href = url;
+                } else {
+                  throw new Error('Failed to open new browser tab.');
+                }
+              } else {
+                return commands.execute('docmanager:open-browser-tab', {
+                  path: item.path
+                });
+              }
             })
           )
         );
       },
       icon: addIcon.bindprops({ stylesheet: 'menuItem' }),
-      label: trans.__('Open in New Browser Tab'),
+      label: args =>
+        args['mode'] === 'single-document'
+          ? trans.__('Open in Simple Mode')
+          : trans.__('Open in New Browser Tab'),
       mnemonic: 0
     });
   }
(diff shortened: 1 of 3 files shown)


fcollonval (changes requested):
Thanks a lot @divyansshhh 

I would propose to make the current command `filebrowser:open-browser-tab` configurable for this feature. And not display the option by default to avoid overpopulating the context menu.

Any command can be called with a JSON-serializable object. In this case we could image passing an argument `mode`. Merging your code with the existing command, the final solution will look like

```ts
    commands.addCommand(CommandIDs.openBrowserTab, {
      execute: args => {
        const widget = tracker.currentWidget;

        if (!widget) {
          return;
        }

        const mode = args["mode"] as string | undefined;

        return Promise.all(
          toArray(
            map(widget.selectedItems(), item => {
              if (mode === 'single-document') {
                const url = PageConfig.getUrl({
                  mode: 'single-document',
                  treePath: item.path
                });
                const opened = window.open();
                if (opened) {
                  opened.opener = null;
                  opened.location.href = url;
                } else {
                  throw new Error('Failed to open new browser tab.');
                }
              } else {
                return commands.execute('docmanager:open-browser-tab', {
                  path: item.path
                });
              }
            })
          )
        );
      },
      icon: addIcon.bindprops({ stylesheet: 'menuItem' }),
      label: args => args['mode'] === 'single-document' ? trans.__('Open in Simple Mode') : trans.__('Open in New Browser Tab'),
      mnemonic: 0
    });
```

To add the new entry in the context menu, users will need to set:

```json
      {
        "command": "filebrowser:open-browser-tab",
        "args": { "mode": "single-document" },
        "selector": ".jp-DirListing-item[data-isdir=\"false\"]",
        "rank": 1.6
      }
```

in _Settings_ -> _Application Context Menu_ -> `contextMenu`.

divyansshhh:
Thanks for the suggestion @fcollonval . I agree that adding `contextMenu` options for all users wouldn't scale well but is there a place where we can document that "Open in Simple Mode" is an option that is available in case someone requires it?

packages/filebrowser-extension/src/index.ts
```diff
@@ -101,6 +101,9 @@ namespace CommandIDs {
 
   export const openBrowserTab = 'filebrowser:open-browser-tab';
 
+  export const openBrowserTabInSimpleMode =
+    'filebrowser:open-browser-tab-in-simple-mode';
+
```
fcollonval, line 106: This is not needed any longer.

```suggestion
```

fcollonval:
Thanks for quickly updating the PR.

> is there a place where we can document that "Open in Simple Mode" is an option that is available in case someone requires it?

I would suggest adding a new file `ui_customization.rst` in https://github.com/jupyterlab/jupyterlab/tree/master/docs/source/user

The layout would be

```rst
.. _interface-customization:

Interface Customization
=======================

.. _context-menu-customization:

Context Menu
------------

File Browser
^^^^^^^^^^^^

<place the doc here>
```

Don't worry about filling the others sections.

> You will need to add the new file to the docs/source/user/index.md to be integrated in the documentation.

fcollonval (changes requested):

packages/filebrowser-extension/src/index.ts
divyansshhh: Yeah I missed that. Thanks

fcollonval (approved):
Thanks a lot @divyansshhh 

> CI failures are not related

packages/filebrowser-extension/src/index.ts
fcollonval, line 106:
```suggestion
```

fcollonval:

jtpio:
Thanks!

While trying this out on Binder, I noticed the advanced settings editor seems to not be displayed the default context menu entries correctly:

image


This is not related to this change, probably tracked in a separate issue?

fcollonval:
> This is not related to this change, probably tracked in a separate issue?

Yes I have seen that for configuration that are array of objects.

PR #12533, opened 2022-05-05
Allow users to yarn link @jupyterlab/builder

ajbozarth:
Currently the compatable version check for the builder will fail
even if a user has used `yarn link` to use the local instance of
@jupyterlab/builder without also changing the version in thier
package.json

This moves that check farther into the function and allows the case
where the version of @jupyterlab/builder found in `node_modules` is
valid, despite the version in the package.json

## References

https://github.com/jupyterlab/team-compass/issues/135

A follow up to my discussion point at the May 4th 2022 meeting

## Code changes

Moves an error raise farther along in the ensure_builder function and adds a new valid case where the version of @jupyterlab/builder found in node_modules has a valid version despite the wrong version in the package.json

## User-facing changes

none

### DIFF
--- a/jupyterlab/federated_labextensions.py
+++ b/jupyterlab/federated_labextensions.py
@@ -275,12 +275,6 @@ def _ensure_builder(ext_path, core_path):
     if "/" in depVersion2:
         with open(osp.join(ext_path, depVersion2, "package.json")) as fid:
             depVersion2 = json.load(fid).get("version")
-    overlap = _test_overlap(depVersion1, depVersion2, drop_prerelease1=True, drop_prerelease2=True)
-    if not overlap:
-        raise ValueError(
-            "Extensions require a devDependency on @jupyterlab/builder@%s, you have a dependency on %s"
-            % (depVersion1, depVersion2)
-        )
     if not osp.exists(osp.join(ext_path, "node_modules")):
         subprocess.check_call(["jlpm"], cwd=ext_path)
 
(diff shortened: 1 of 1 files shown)


ajbozarth:
A few notes on this in reference to my questions during the dev meeting:

I found that my issue with `disabledExtensions` not working was due to using the deprecated method as [outlined in the docs](https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#enabling-and-disabling-extensions). Updating it solved that issue. 

I then used the method of adding a `path` in the tsconfig as [detailed in the docs](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#developing-a-prebuilt-extension) but I had issues getting @jupyterlab/builder to work since in a monorepo that package has to be `nohoist` so it is available at each packages root. I solved this by using yarn link to point each package at the local version, which is where I ran into the problem addressed in this PR.

Relatedly this PR should probably be back ported to recent minor versions as well to allow extension development on those branches.

blink1073 (approved):
👍🏼

fcollonval (approved):
Thanks @ajbozarth and @blink1073

fcollonval:
> Relatedly this PR should probably be back ported to recent minor versions as well to allow extension development on those branches.

@ajbozarth would you mind backporting the PR and carrying out the associated releases?

ajbozarth:

PR #12534, opened 2022-05-05
Integrate `jupyterlab-lsp` into `jupyterlab`

trungleduc:
## References

This PR improves the LSP integration in the `JupyterLab` frontend to make LSP features available to other extensions. It will lay the infrastructure for handling and communicating with language servers.

This PR does not contain any frontend features like code completion, code refactoring, or jump to..., it will be the subject of future work. An example of an LSP code completion provider is available [here](https://gist.github.com/trungleduc/166d6cd915558840d9ea51c4282f90ee) 

`JupyterLab` will not come with any language servers, users still need to install language servers manually.  

## Code changes

- On the frontend, the core elements of the [jupyterlab-lsp](https://github.com/jupyter-lsp/jupyterlab-lsp) extension that handles the language server initialization and communication are moved into the `lsp` package of `JupyterLab`.

src_diagram4
<p align = "center">
<b>Diagram of components taken into JupyterLab</b>
</p>

- `lsp-extension` package is created to provide tokens that allow other extensions to interact with the language servers:
		- `ILSPDocumentConnectionManager`: the token used to manage the connection to language servers for the notebook and file editor.
		-  `ILSPFeatureManager`: the token used to register features with language servers.
		- `ILSPCodeExtractorsManager`: the token used to manage the code extractors, this will allow the creation of multiple language server connections on one notebook.

- The backend parts, for now, are provided by [`jupyter-lsp`](https://github.com/jupyter-lsp/jupyterlab-lsp/tree/master/python_packages/jupyter_lsp) package.

### Running language server panel
A new language server section is added to the running panel. If the language server for the language of the opened notebook is available, `JupyterLab` will create a connection to this server for the notebook. The language server section will show the running server and allow users to terminate these servers.

lsp2
### Setting interface for the language server
`JupyterLab` will detect automatically the installed language servers and will propose the settings for these servers in the settings UI.
lsp3
## Remaining tasks

 - [x] Migration direction for the backend component https://github.com/jupyterlab/jupyterlab_server/issues/240
 - [x] Move [`lsp-ws-connection`](https://github.com/jupyter-lsp/jupyterlab-lsp/tree/master/packages/lsp-ws-connection) into `JupyterLab`
 - [x] Handle the json schema  <-> typescript interface generation.
 - [x] ~Documentation for language servers installation and configuration.~
 - [x] ~Guide for creating `LSP` extension.~
 
I will address the last 2 tasks in the follow-up PR. 

N/A

cc @krassowski @bollwyvl 

 [Binder](https://mybinder.org/v2/gh/trungleduc/jupyterlab/lsp?urlpath=lab-dev)


jtpio:
Thanks @trungleduc for starting this, and the very well detailed write-up!

> This PR does not contain any frontend features like code completion, code refactoring, or jump to..., it will be the subject of future work. An example of an LSP code completion provider is available [here](https://gist.github.com/trungleduc/166d6cd915558840d9ea51c4282f90ee)

Is the plan to move more features and components currently in https://github.com/jupyter-lsp/jupyterlab-lsp to JupyterLab core? If so, should we open a new issue in the team compass to increase visibility and follow the usual process, similar to https://github.com/jupyterlab/team-compass/issues/145?

trungleduc:
@jtpio ok,  I will open an issue in the team compass.

trungleduc:
Please update galata references

trungleduc:
please update snapshots

trungleduc:
please update snapshots

jtpio:
FYI @trungleduc I published `alpha.11` versions of the `lsp` packages to `npm` which should help fix the CI.

jtpio:
So most of the CI is now passing.

@trungleduc mind doing a quick rebase, which should hopefull help make the visual regression documentation check pass as fixed by https://github.com/jupyterlab/jupyterlab/pull/12714?

trungleduc:
Thanks @jtpio, I'll get back to this PR next week.

jtpio:
> make the visual regression documentation check pass as fixed by https://github.com/jupyterlab/jupyterlab/pull/12714?

Actually the diff for this PR shows that some of the extension manager snapshots have been updated, so these might have to be reverted:

image

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,274 @@
...
+  id: '@jupyterlab/lsp-extension:feature',
+  requires: [ISettingRegistry, ITranslator],
+  provides: ILSPFeatureManager,
+  autoStart: true
+};
+
+const codeExtractorManagerPlugin: JupyterFrontEndPlugin<ILSPCodeExtractorsManager> =
+  {
+    id: ILSPCodeExtractorsManager.name,
+    requires: [],
```
jtpio, line 60: Probably can be omitted if empty?

```suggestion
```

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,274 @@
...
+
+function activateFeature(
+  app: JupyterFrontEnd,
+  settingRegistry: ISettingRegistry,
+  translator: ITranslator
+): ILSPFeatureManager {
+  const featureManager = new FeatureManager();
+
+  return featureManager;
+}
```
jtpio, line 198: Maybe this one can be inlined at the plugin definition directly if it just `return new FeatureManager();`?

packages/notebook-extension/src/index.ts
```diff
@@ -1751,6 +1770,34 @@ function activateNotebookCompleterService(
...
+ * Activate the language server for notebook.
+ */
+function activateNotebookLanguageServer(
+  app: JupyterFrontEnd,
+  notebooks: INotebookTracker,
+  connectionManager?: ILSPDocumentConnectionManager,
+  featureManager?: ILSPFeatureManager,
+  codeExtractorManager?: ILSPCodeExtractorsManager
+): void {
+  if (!connectionManager || !featureManager || !codeExtractorManager) {
```
jtpio, line 1783: Similar comment as above.

jtpio (commented):
Looking good!

Did a first pass focused on the plugin definitions mostly which resulted in a couple of comments.

packages/lsp-extension/schema/plugin.json
```diff
@@ -0,0 +1,62 @@
...
+          "description": "When multiple servers match specific document/language, the server with the highest priority will be used",
+          "type": "number",
+          "default": 50,
+          "minimum": 1
+        }
+      }
+    }
+  },
+  "properties": {
+    "language_servers": {
```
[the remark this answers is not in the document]
trungleduc: Yes, I want to keep compatibility with the extension, but indeed it's better to update  these variables name

packages/lsp/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+export * from './adapters/adapter';
+export * from './positioning';
+export * from './virtual/document';
+export * from './utils';
+export * from './connection';
+export * from './feature';
+export * from './extractors/manager';
+export * from './extractors/text_extractor';
+export * from './extractors/types';
+export * from './plugin';
```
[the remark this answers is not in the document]
trungleduc: Since the notebook and file adapters are now generated in the perspective packages, I need these modules to be exported.

trungleduc:
please update snapshots

packages/lsp/src/virtual/document.ts
```diff
@@ -0,0 +1,945 @@
...
+  private updateLock: boolean = false;
+
+  protected isDisposed = false;
+
+  /**
+   * Signal emitted by the editor that triggered the update, providing the root document of the updated documents.
+   */
+  private documentUpdated: Signal<UpdateManager, VirtualDocument>;
+  public blockAdded: Signal<UpdateManager, IBlockAddedInfo>;
+  updateDone: Promise<void> = new Promise<void>(resolve => {
```
fcollonval, line 843: Seems it should be readonly for the public API.

packages/notebook-extension/src/index.ts
```diff
@@ -1751,6 +1770,30 @@ function activateNotebookCompleterService(
...
+ */
+function activateNotebookLanguageServer(
+  app: JupyterFrontEnd,
+  notebooks: INotebookTracker,
+  connectionManager: ILSPDocumentConnectionManager,
+  featureManager: ILSPFeatureManager,
+  codeExtractorManager: ILSPCodeExtractorsManager
+): void {
+  notebooks.widgetAdded.connect(async (_, notebook) => {
+    const adapter = new NotebookAdapter(
```
fcollonval, line 1784: Should the `adapter` be disposed when the notebook is disposed?

packages/fileeditor-extension/src/index.ts
```diff
@@ -534,3 +555,24 @@ function activateFileEditorCompleterService(
...
+
+function activateFileEditorLanguageServer(
+  app: JupyterFrontEnd,
+  notebooks: IEditorTracker,
+  connectionManager: ILSPDocumentConnectionManager,
+  featureManager: ILSPFeatureManager,
+  extractorManager: ILSPCodeExtractorsManager
+): void {
+  notebooks.widgetAdded.connect(async (_, notebook) => {
+    const adapter = new FileEditorAdapter(
```
fcollonval, line 567: Should the adapter be disposed when the editor is disposed?

packages/notebook/src/notebooklspadapter.ts
```diff
@@ -0,0 +1,360 @@
...
+      let cellWidget = this.widget.content.widgets.find(
+        cell => cell.model.id === cellModel.id
+      );
+      if (!cellWidget) {
+        console.warn(
+          `Widget for removed cell with ID: ${cellModel.id} not found!`
+        );
+        continue;
+      }
+      this.knownEditorsIds.delete(cellWidget.editor.uuid);
```
fcollonval, line 255: I don't see the removal of the cell in the _ceEditorToCell_ mapping

fcollonval (changes requested):
Generally speaking this PR will benefit of getting more doc strings and conform to the [code style guidelines](https://github.com/jupyterlab/jupyterlab/wiki/TypeScript-Style-Guide).

On the API front, some classes are missing the disposable feature - required by the use signals. Those should be added and disposal of adapters, virtual document,... should be more careful crafted - for instance I did not see any explicit call to dispose object when the document widget is disposed.
Another point that needs to be addressed is to remove the link to `nbformat.CellType`. We should support a generic `type: string` to allow more flexibility and avoid adding explicit dependency to nbformat in the lsp and fileeditor packages.

fcollonval:
From binder, it will be nice to change the setting icon; refresh does not seem appropriate:

image

jtpio:
@trungleduc I'm thinking that maybe it would make sense to update the staging dependencies as done initially, which should fix the `splice_source` check.

It would be fine to have the `staging` check fail temporarily for this PR since new packages are being added.

trungleduc:
> @trungleduc I'm thinking that maybe it would make sense to update the staging dependencies as done initially, which should fix the `splice_source` check.
> 
> It would be fine to have the `staging` check fail temporarily for this PR since new packages are being added.

👍 I'll do it

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,265 @@
...
+    // TODO: if priorities changed reset connections
+    connectionManager.updateConfiguration(languageServerSettings);
+    connectionManager.updateServerConfigurations(languageServerSettings);
+    connectionManager.updateLogging(
+      options.logAllCommunication,
+      options.setTrace
+    );
+  };
+  languageServerManager.sessionsChanged.connect(() => {
+    settingRegistry.transform(plugin.id, {
```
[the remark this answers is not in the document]
trungleduc: I need this pattern because the transform callback will read the language server session list,  and we don't have control over the availability of this list. 
> If that pattern is needed, you should store the disposable object output by settingRegistry.transform and dispose the previous one before adding a new transform callback.

Applied

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,265 @@
...
+  managers.add({
+    name: trans.__('Language servers'),
+    running: () => {
+      const connections = new Set([...lsManager.connections.values()]);
+      return [...connections].map(
+        conn => new RunningLanguageServers(conn, lsManager)
+      );
+    },
+    shutdownAll: () => {
+      /** */
```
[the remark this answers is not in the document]
trungleduc: Implementation  added

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,265 @@
...
+    },
+    refreshRunning: () => {
+      /** */
```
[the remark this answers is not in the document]
trungleduc: This is no-op

packages/lsp/src/positioning.ts
```diff
@@ -0,0 +1,86 @@
...
+  isEditor: true;
+}
+
+export interface IVirtualPosition extends IPosition {
+  isVirtual: true;
+}
+
+export interface IRootPosition extends ISourcePosition {
+  isRoot: true;
+}
```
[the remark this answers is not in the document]
trungleduc: According to the comment
`is_* attributes are there only to enforce strict interface type checking`

packages/lsp/src/schema.ts
```diff
@@ -0,0 +1,235 @@
...
+ * via the `patternProperty` ".+".
+ */
+export type InstallHelp = string;
+/**
+ * languages supported by this Language Server
+ *
+ * This interface was referenced by `JupyterLspServerStatusResponse`'s JSON-Schema
+ * via the `definition` "language-list".
+ */
+export type LanguageList = [string, ...string[]];
```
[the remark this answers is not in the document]
trungleduc: This interface is generated from the schema https://github.com/jupyter-server/jupyter-lsp/blob/main/jupyter_lsp/schema/schema.json, I think it is to enforce the `minItems` value

packages/lsp/src/schema.ts
```diff
@@ -0,0 +1,235 @@
...
+ */
+export type DisplayName = string;
+/**
+ * known extensions that can contribute to the Language Server's features
+ */
+export type Extensions = LanguageServerExtension[];
+/**
+ * list of MIME types supported by the language server
+ */
+export type MIMETypes = [string, ...string[]];
```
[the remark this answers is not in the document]
trungleduc: Same as above

packages/lsp/src/schema.ts
```diff
@@ -0,0 +1,235 @@
...
+export type SpecSchemaVersion = 2;
+/**
+ * This interface was referenced by `JupyterLspServerStatusResponse`'s JSON-Schema
+ * via the `definition` "env-var".
+ *
+ * This interface was referenced by `EnvironmentVariables`'s JSON-Schema definition
+ * via the `patternProperty` "[^ ]+".
+ */
+export type AnEnvironmentVariableMayContainPythonStringTemplateEvaluatedAgainstTheExistingEnvironmentEG$HOME =
+  string;
```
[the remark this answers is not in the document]
trungleduc: It is generated automatically from the schema file.

packages/lsp-extension/src/index.ts
```diff
@@ -0,0 +1,273 @@
...
+export class RunningLanguageServers implements IRunningSessions.IRunningItem {
+  constructor(
+    connection: ILSPConnection,
+    manager: ILSPDocumentConnectionManager
+  ) {
+    this._connection = connection;
+    this._manager = manager;
+  }
+  /**
+   * This i no-op because we do not do anything on server click event
```
trungleduc, line 202:
```suggestion
   * This is no-op because we do not do anything on server click event
```

fcollonval:
> @fcollonval I still have failed Ui tests, does it relate to #12819? Otherwise, this PR is ready for another review round.

Indirectly yes, but it is not related to this PR (and is fixed by #12821).

...doing another review right now.

packages/notebook/src/notebooklspadapter.ts
```diff
@@ -0,0 +1,467 @@
...
+      });
+    }
+    this._activeEditorChanged.emit({
+      editor: cell.editor
+    });
+  }
+
+  /**
+   * Get the cell at the cursor position of the virtual document.
+   * @param {IVirtualPosition} pos
```
fcollonval, line 443:
```suggestion
   * @param pos
```

packages/lsp-extension/src/renderer.tsx
```diff
@@ -0,0 +1,483 @@
...
+  [key: string]: ISettingProperty;
+}
+interface ISettingProperty {
+  property: string;
+  type: 'boolean' | 'string' | 'number';
+  value: any;
+}
+const SETTING_NAME = 'languageServers';
+const SERVER_SETTINGS = 'configuration';
+function debounce<Params extends any[]>(
```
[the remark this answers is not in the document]
fcollonval: Let's use the coming new version of Lumino

fcollonval (changes requested):
Here it comes.

packages/lsp/src/adapters/adapter.ts
```diff
@@ -0,0 +1,708 @@
...
+ * it operates on the virtual document and must not attempt to infer the language dependencies
+ * as this would make the logic of inspections caching impossible to maintain, thus the WidgetAdapter
+ * has to handle that, keeping multiple connections and multiple virtual documents.
+ */
+export abstract class WidgetLSPAdapter<T extends IDocumentWidget> {
+  // note: it could be using namespace/IOptions pattern,
+  // but I do not know how to make it work with the generic type T
+  // (other than using 'any' in the IOptions interface)
+  constructor(public widget: T, protected options: IAdapterOptions) {
+    this.app = options.app;
```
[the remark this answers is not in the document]
trungleduc: `app` is removed and `docRegistry` is added as an option for the file editor adapter.

packages/lsp/src/adapters/adapter.ts
```diff
@@ -0,0 +1,708 @@
...
+
+  /**
+   * Promise that resolves once the document is updated
+   */
+  updateFinished: Promise<void>;
+
+  /**
+   * Promise that resolves once the adapter is initialized
+   */
+  initialized: Promise<void>;
```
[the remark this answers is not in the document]
afshin: I agree with you, `ready` is more idiomatically JupyterLab 👍

trungleduc:
@krassowski, we a finalizing this PR and the remaining discussion is the scope of unstable APIs. I tagged all newly added tokens and APIs as `alpha` (https://github.com/jupyterlab/jupyterlab/pull/12534/commits/783056f17de6ee64f9525eb23e7a4ec97c036509). Do we need to aware of any old APIs coming from the extension?

packages/fileeditor/src/fileeditorlspadapter.ts
```diff
@@ -0,0 +1,190 @@
+import { DocumentRegistry } from './../../docregistry/src/registry';
```
krassowski, line 1:
```suggestion
import { DocumentRegistry } from '@jupyterlab/docregistry';
```

?

packages/fileeditor/src/fileeditorlspadapter.ts
```diff
@@ -0,0 +1,190 @@
...
+    // when MIME type is not known it defaults to 'text/plain',
+    // so if it is different we can accept it as it is
+    if (codeMirrorMimeType != 'text/plain') {
+      return codeMirrorMimeType;
+    } else if (contentsModel) {
+      // a script that does not have a MIME type known by the editor
+      // (no syntax highlight mode), can still be known by the document
+      // registry (and this is arguably easier to extend), so let's check it
+      // just in case; this is also how the "Klingon" language for testing
+      // gets registered, so we need it for tests too.
```
krassowski, line 67:
```suggestion
      // registry (and this is arguably easier to extend).
```

I think there are no integration tests with Klingon in core, so let's cut it.

packages/lsp-extension/schema/plugin.json
```diff
@@ -0,0 +1,57 @@
...
+        ".*": {
+          "$ref": "#/definitions/languageServer"
+        }
+      }
+    },
+    "setTrace": {
+      "title": "Ask servers to send trace notifications",
+      "enum": ["off", "messages", "verbose"],
+      "default": "off",
+      "description": "Whether to ask server to send logs with execution trace (for debugging). To see these messages, set loggingLevel to debug or log. Accepted values are: \"off\", \"messages\", \"verbose\". Servers are allowed to ignore this request."
```
krassowski, line 54: Either implement `loggingLevel` filtering, or remove mention of it.

```suggestion
      "description": "Whether to ask server to send logs with execution trace (for debugging). Accepted values are: \"off\", \"messages\", \"verbose\". Servers are allowed to ignore this request."
```

packages/lsp-extension/schema/plugin.json
```diff
@@ -0,0 +1,57 @@
...
+      "title": "Language Server",
+      "description": "Language-server specific configuration, keyed by implementation",
+      "type": "object",
+      "default": {},
+      "patternProperties": {
+        ".*": {
+          "$ref": "#/definitions/languageServer"
+        }
+      }
+    },
```
krassowski, line 49: Should we also expose `logAllCommunication` as a setting as it is in the extension?

packages/lsp/src/connection.ts
```diff
@@ -0,0 +1,529 @@
+// Disclaimer/acknowledgement: Fragments are based on LspWsConnection, which is copyright of wylieconlon and contributors and ISC licenced.
+// ISC licence is, quote, "functionally equivalent to the simplified BSD and MIT licenses,
+// but without language deemed unnecessary following the Berne Convention." (Wikipedia).
+// Introduced modifications are BSD licenced, copyright JupyterLab development team.
```
krassowski, line 4: I think there is nothing left of the upstream `LspWsConnection` in this file, we should be fine using standard JupyterLab copyright header.

packages/lsp/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "@jupyterlab/coreutils": "^6.0.0-alpha.11",
+    "@jupyterlab/docregistry": "^4.0.0-alpha.11",
+    "@jupyterlab/services": "^7.0.0-alpha.11",
+    "@jupyterlab/translation": "^4.0.0-alpha.11",
+    "@lumino/coreutils": "^1.12.0",
+    "@lumino/disposable": "^1.10.1",
+    "@lumino/signaling": "^1.10.1",
+    "codemirror": "~5.61.0",
+    "lodash.mergewith": "^4.6.1",
+    "lsp-ws-connection": "~0.7.0",
```
krassowski, line 58: I think we should remove this dependency, it was getting axed anyways (I think this is incorrectly ticked as done in the PR description). It only provides `registerServerCapability` and `unregisterServerCapability` and `LspWsConnection`; let's merge [`LspWsConnection`](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/240f2cbda77f574ab98ae65dafea8b694ecb13f5/packages/lsp-ws-connection/src/ws-connection.ts#L30-L33) into `LSPConnection` removing `@deprecated` methods (everything which uses `events.EventEmitter` or old `get`/`send` API).

krassowski (changes requested):
Excited to see this so close to the finish line. Leaving some feedback in answer to the question about old interfaces: basically `lsp-ws-connection` has a lot of deprecated stuff which we should cut to avoid bringing useless dependencies (like events) and dev dependencies.

Also, I noticed that standard copyright headers are missing:

```
```

and sometimes there are headers in another format. I would suggest to use the standard lab format for consitency, it will be easier for us if we decide to grep-change them later.

fcollonval:
Regarding the license header, I'll try to add an GitHub action to reach homogeneity. I found the promising: https://github.com/marketplace/actions/license-eye-header

trungleduc:
Thank @krassowski for your review, I removed `lsp-ws-connection` in https://github.com/jupyterlab/jupyterlab/pull/12534/commits/1ccb7513bad40ecb54919c4aa252bd76658ce8db. I did not merge `LspWsConnection` into `LSPConnection` since I want to separate vendor codes.

packages/lsp/src/tokens.ts
```diff
@@ -0,0 +1,625 @@
...
+  AnyLocation,
+  IDocumentInfo,
+  ILspConnection,
+  ILspOptions
+} from './ws-connection/types';
+
+import type * as rpc from 'vscode-jsonrpc';
+import type * as lsp from 'vscode-languageserver-protocol';
+export { IDocumentInfo };
+export type TLanguageServerId =
```
fcollonval, line 27: can it be something else? Would it block us if a new server needs to be added?

packages/lsp/src/tokens.ts
```diff
@@ -0,0 +1,625 @@
+// Copyright (c) Jupyter Development Team.
```
fcollonval, line 1: Could you add a doc strings for all exported types and interfaces in this file as they will be the primary API used by developers.

fcollonval (commented):
Only two minor comments otherwise the code looks good.

packages/lsp/src/tokens.ts
krassowski: It doesn't matter at runtime, a few example values are used kind of as a sentinel to get typing errors if a different kind of ID (language ID, also a string) was to be mistakenly passed. But no strong feelings on converting it to just a string.

fcollonval: Thanks @krassowski happy to keep it that way for now.

krassowski (commented):
Some quick coffee-break comments, will take another look later.

SylvainCorlay:
Hey @krassowski let us know if we could address more things in follow up PRs. It would be really nice to get this in (in a pre-release) so that we could start working on the follow-up changes in the extension. We will most certainly find more things to fix as we adapt jupyterlab-lsp to use this code, and codemirror 6.

krassowski (approved):
I'm happy for it to be merged and iterated on later (though let's address the two minor things from my previous review so that they are not lost once we merge).

SylvainCorlay:
Thanks ! Lets fix these up an iterate!

fcollonval (approved):
Thanks a lot @trungleduc

PR #12537, opened 2022-05-05
Fix code output font size to be affected by content-font-size1

### LINKED ISSUE #12178: Output font size not changing for code cells
**Version: 3.3.0**

Hello,

It seems that this has already been implemented but I can't get it to work for output  of code cells. My theme settings are:

```json
{
    // Theme
    // @jupyterlab/apputils-extension:themes
    // Theme manager settings.
    // *************************************

    // Theme CSS Overrides
    // Override theme CSS variables by setting key-value pairs here
    "overrides": {
        "code-font-family": "JetBrains Mono",
        "ui-font-size1": "14px",
        "content-font-size1": "18px",
        "content-font-family": "Noto Sans"
    },
    "theme": "JupyterLab Dark",
}

It does affect the output of Markdown cells, but the output of other code cells is still smaller. Am I missing another setting?

galathinius:
## References

 
Fixes #12178 

## Code changes

Add the jp-RenderedHTMLCommon CSS class (which gets the font size from content-font-size1) to RenderedText

## User-facing changes

Before:
image

After:
image


krassowski:
Thank you for the PR @galathinius  I will update your branch with the main branch and update snapshots.

I am not certain if we should proceed the way proposed. As I understand, this PR changes the default font size from `--jp-code-font-size` to `--jp-code-font-size1`. By some opinionated users, this might be considered a regression. A safer approach to solve #12178 could could be allowing to customise `--jp-code-font-size` in addition to `--jp-code-font-size1`. I will also tag this as needs review to get a second opinion.

krassowski:

krassowski (changes requested):
The updated snapshots show that on top of changing the font size, addition of the proposed class would also break the background for exceptions:

Screenshot from 2022-08-11 20-46-06

PR #12538, opened 2022-05-05
Fix file browser search highlighting bug

### LINKED ISSUE #12267: Search results not getting highlighted in filebrowser
## Description
In jupyterlab 3.3.x, the search results in the filebrowser no longer get highlighted. Here are some screenshots - 
In lab 3.3.x - 
image
In lab 3.2.4 - 
image

## Reproduce
Enter a search query in the filebrowser search.

## Expected behavior
Matching search results should be highlighted.

## Context

- Operating System and version: Linux Ubuntu 21.04 
- Browser and version: Chrome 92 
- JupyterLab version: 3.3.2

marthacryan:
## References
Fixes #12267. Due to changes made to the `FilterBox` component introduced in #11079, the file browser highlighting stopped working, this adds back that functionality. 

## Code changes
Changes the filter function returned by the `FilterBox` component to include the matching indices (if there are any). Changes the components that use the `FilterBox` component (settings editor plugin list, and the filebrowser model) to use the updated filter function. Sets the indices in the filebrowser model when the filter updates. 

## User-facing changes
Adds back in highlighting in the file browser search. 

Changes the API of the `FilterBox` to take a different `updateFilter` function (returns `{ match: boolean; indices?: number[] }` instead of `boolean`.


fcollonval (commented):
Thanks @marthacryan would it be possible to fix the bug without break the API? Would it be possible to use `filter(i: string) => boolean | {match: boolean; indices?: number[]}`. So we can backport this safely.

fcollonval (changes requested):
Diving into previous discussion, I think we should move to the suggestion of @krassowski : https://github.com/jupyterlab/jupyterlab/pull/12082#issuecomment-1049148451

So the new filter signature will be `filter(item: string) => Partial<IScore> | null` where `IScore` is the interface of `fuzzySearch`.

So testing `if(filter(query))` will work as before `Partial<IScore>` <-> `true` and `null` <-> `false`.

For filter returning simply a boolean, you can return and empty object when true; null otherwise.

marthacryan:
Thanks @fcollonval! Do my changes look like what you were suggesting?

packages/ui-components/src/components/search.tsx
```diff
@@ -53,16 +56,16 @@ interface IScore {
   /**
    * The indices of the text matches.
    */
-  indices: number[] | null;
+  indices?: number[];
```
fcollonval, line 59: What is the reason to change this type?

fcollonval (commented):
Thanks @marthacryan I have one question.

packages/ui-components/src/components/search.tsx
marthacryan: Oh oops didn't mean to leave that change in, it was leftover from a different approach. Just reverted it

packages/ui-components/src/components/search.tsx
```diff
@@ -59,10 +62,10 @@ interface IScore {
 /**
  * Perform a fuzzy search on a single item.
  */
-function fuzzySearch(source: string, query: string): IScore | null {
+export function fuzzySearch(source: string, query: string): IScore | null {
   // Set up the match score and indices array.
   let score = Infinity;
-  let indices: number[] | null = null;
+  let indices: number[] | undefined;
```
fcollonval, line 68:
```suggestion
  let indices: number[] | null = null;
```

fcollonval (approved):
Thanks @marthacryan

fcollonval:

PR #12543, opened 2022-05-06
fixed shouldOverwrite is never called when rename target exists

### LINKED ISSUE #12541: Rename via right click on notebook ignores rename if target already exists
## Description

If you right click on a jupyterlab notebook tab and try to give the clicked notebook the name of an already existing notebook, nothing happens. If you do the same thing in the file browser on the left you get asked if you want to overwrite the existing notebook.

The cause for this is that the error returned from the notebook server has a different structure than expected. The status code '409' is no longer included in `error.message` but in `error.response.status` and therefore the check in  [line 86](https://github.com/jupyterlab/jupyterlab/blob/43f784a35b5ec474f5a94433c5d2d4d6d712cbc8/packages/docmanager/src/dialogs.ts#L86) is always true and the shouldOverwrite function is never called.

## Reproduce

1. Create a new notebook named `foo.ipynb` and save it
2. Create a second notebook and right click on the `Untitled.ipynb` name in the notebook tab
3. Choose to rename the notebook to `foo`
4. Nothing happens

## Expected behavior

You should have been asked whether you want to overwrite `foo.ipynb`.

## Context

- Operating System and version: macOS 12.3.1
- Browser and version: Safari 15.4 / Chrome 100.0.4896.127
- JupyterLab version: 4.0.0a24

ephes:
## References

Fix #12541

## Code changes

The main fix is to not look at `error.message` to determine if the error was cause by an already existing notebook with this name but `error.response.status`.

I also added 3 tests to make sure `shouldOverwrite` is called when `error.response.status` is `409`. And that other errors are re-thrown and don't cause `shouldOverwrite` to be called.

I'm a javascript newbie and had trouble to mock `shouldOverwrite`. Therefore I added a callback parameter to `renameFile` to be able to pass a mock callback in the tests. But probably somebody with more javascript experience should have a look at the tests, there's probably an easier way.

## User-facing changes

Nothing really changed besides the missing dialog is now shown.

---
Edited to link the issue

### DIFF
--- a/packages/docmanager/test/dialog.spec.ts
+++ b/packages/docmanager/test/dialog.spec.ts
@@ -1,0 +1,84 @@
+import { DocumentRegistry } from '@jupyterlab/docregistry';
+import { Contents, ServiceManager } from '@jupyterlab/services';
+import { Widget } from '@lumino/widgets';
+import * as Mock from '@jupyterlab/testutils/lib/mock';
+import { DocumentManager, renameFile } from '../src';
+
+function shouldOverwriteFalse(path: string): Promise<boolean> {
+  return Promise.resolve(false);
+}
+
+describe('docregistry/dialog', () => {
+  let manager: DocumentManager;
+  let services: ServiceManager.IManager;
+  let alreadyExistsError: any = {};
+
+  beforeAll(() => {
+    const registry = new DocumentRegistry({});
+    services = new Mock.ServiceManagerMock();
+    manager = new DocumentManager({
+      registry,
+      manager: services,
+      opener: {
+        open: (widget: Widget) => {
+          // no-op
+        }
+      }
+    });
+  });
+
+  beforeEach(() => {
+    alreadyExistsError = {
+      name: 'file already exists',
+      message: 'File already exists: bar.ipynb',
+      response: {
+        status: 409
+      }
+    };
+    manager.rename = (oldPath, newPath) => {
+      const renamePromise = new Promise(function (resolve, reject) {
+        throw alreadyExistsError;
+      }) as Promise<Contents.IModel>;
+      return renamePromise;
+    };
+  });
+
+  describe('@jupyterlab/docmanager', () => {
+    describe('#renameFile()', () => {
+      it('should show overwrite dialog when file is already existing', async () => {
+        alreadyExistsError.response.status = 409;
+        await renameFile(
+          manager,
+          'foo.ipynb',
+          'bar.ipynb',
+          shouldOverwriteFalse
+        ).catch(error => {
+          // error should be 'File not renamed'
+          expect(error).not.toBe(alreadyExistsError);
+        });
+      });
+      it('should throw error on no status', async () => {
+        alreadyExistsError.response = {};
+        await renameFile(
+          manager,
+          'foo.ipynb',
+          'bar.ipynb',
+          shouldOverwriteFalse
+        ).catch(error => {
+          expect(error).toBe(alreadyExistsError);
+        });
+      });
+      it('should throw error on not 409 status', async () => {
+        alreadyExistsError.response.status = 408;
+        await renameFile(
+          manager,
+          'foo.ipynb',
+          'bar.ipynb',
+          shouldOverwriteFalse
+        ).catch(error => {
+          expect(error).toBe(alreadyExistsError);
+        });
+      });
+    });
+  });
+});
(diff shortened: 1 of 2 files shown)


packages/docmanager/test/dialog.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { DocumentRegistry } from '@jupyterlab/docregistry';
+import { Contents, ServiceManager } from '@jupyterlab/services';
+import { Widget } from '@lumino/widgets';
+import * as Mock from '@jupyterlab/testutils/lib/mock';
+import { DocumentManager, renameFile } from '../src';
+
+function shouldOverwriteFalse(path: string): Promise<boolean> {
+  return Promise.resolve(false);
+}
```
fcollonval, line 9: We are using Jest for testing. The documentation for mocking is [there](https://jestjs.io/docs/26.x/mock-functions#mocking-partials).

Something like the following should to the trick

```suggestion
import { DocumentManager, renameFile } from '../src';

jest.mock('../src', () => {
  const originalModule = jest.requireActual('../src');

  //Mock the shouldOverwrite
  return {
    __esModule: true,
    ...originalModule,
    shouldOverwrite: jest.fn().mockReturnValue(false),
  };
});
```

packages/docmanager/test/dialog.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        status: 409
+      }
+    };
+    manager.rename = (oldPath, newPath) => {
+      const renamePromise = new Promise(function (resolve, reject) {
+        throw alreadyExistsError;
+      }) as Promise<Contents.IModel>;
+      return renamePromise;
+    };
+  });
```
fcollonval, line 44: You should be able to take advantage of the [jest.spyOn](https://jestjs.io/docs/26.x/jest-object#jestspyonobject-methodname) feature with a code like:

```suggestion
    alreadyExistsError = {
      name: 'file already exists',
      message: 'File already exists: bar.ipynb',
      response: {
        status: 409
      }
    };
    // spyRename can be store as variable as manager
    spyRename = jest.spyOn(manager, 'rename');
    spyRename.mockRejectedValue(alreadyExistsError);
  });

  afterEach(() => {
    spyRename.mockReset();
  });
```

fcollonval (changes requested):
Thanks @ephes 

I added some pointers on how to use proper Jest mock (the JS test library used by JupyterLab).
Let me know if you need more help.

packages/docmanager/test/dialog.spec.ts
ephes: Thanks, I didn't know about mocking partials, which seems to be very useful. But in this case it doesn't work, because it only mocks shouldOverwrite if you call it from the test. But it's called from within `dialog.ts` and stays unmocked there. A lot of people seem to have run [into this](https://github.com/facebook/jest/issues/936) (even found a comment from myself 2 years ago in this thread 😬).
 
 I tried [option 1](https://github.com/facebook/jest/issues/936#issuecomment-545080082) but ended up with the typescript error message `TypeError: Cannot redefine property: shouldOverwrite`.

packages/docmanager/test/dialog.spec.ts
ephes: Cool, this is much simpler, thanks :). Changed in 370655d5fc.

packages/docmanager/test/dialog.spec.ts
fcollonval: Thanks for trying that path @ephes 

I look a bit more at the code of `shouldOverwrite`. And as this is triggering a dialog, you could make use of the test helper `dimissDialog` (as you are emulating a rejection). See that example:

https://github.com/jupyterlab/jupyterlab/blob/43f784a35b5ec474f5a94433c5d2d4d6d712cbc8/packages/apputils/test/sessioncontext.spec.ts#L578

Something like the following code should do the trick:

```ts
        await Promise.all([dismissDialog(), renameFile(
          manager,
          'foo.ipynb',
          'bar.ipynb'
        )]).catch(error => {
          // error should be 'File not renamed'
          expect(error).not.toBe(alreadyExistsError);
        });
```

fcollonval (approved):
Thanks @ephes

fcollonval:
please backport to 3.4.x

fcollonval:

packages/docmanager/test/dialog.spec.ts
ephes: Oh wow, this is much better.

ephes:
> Thanks @ephes

Thank you too @fcollonval :)

PR #12556, opened 2022-05-10
Fix loading toolbar factory twice

fcollonval:
## References

Seen in https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/32

## Code changes

- Tolerate to create a toolbar factory multiple times from the same plugin (like multiple filebrowser with the same buttons).
- For late plugin registration, we reload all plugin toolbar definitions as for the shortcuts.
- Compare only the toolbar item name to determine if definitions must be merged

## User-facing changes

N/A

N/A

### DIFF
--- a/packages/apputils/src/toolbar/factory.ts
+++ b/packages/apputils/src/toolbar/factory.ts
@@ -105,54 +105,58 @@ async function setToolbarItems(
   }
 
   // Transform the plugin object to return different schema than the default.
-  registry.transform(pluginId, {
-    compose: plugin => {
-      // Only override the canonical schema the first time.
-      if (!canonical) {
-        canonical = JSONExt.deepCopy(plugin.schema);
-        populate(canonical);
+  if (!Private.transformedPlugin.includes(pluginId)) {
+    registry.transform(pluginId, {
+      compose: plugin => {
+        // Only override the canonical schema the first time.
+        if (!canonical) {
+          canonical = JSONExt.deepCopy(plugin.schema);
+          populate(canonical);
+        }
+
+        const defaults =
+          ((canonical.properties ?? {})[propertyId] ?? {}).default ?? [];
+        // Initialize the settings
+        const user: PartialJSONObject = plugin.data.user;
+        const composite: PartialJSONObject = plugin.data.composite;
+        // Overrides the value with using the aggregated default for the toolbar property
+        user[propertyId] =
+          (plugin.data.user[propertyId] as ISettingRegistry.IToolbarItem[]) ??
+          [];
+        composite[propertyId] = (
+          SettingRegistry.reconcileToolbarItems(
+            defaults as ISettingRegistry.IToolbarItem[],
+            user[propertyId] as ISettingRegistry.IToolbarItem[],
+            false
+          ) ?? []
+        ).sort(
+          (a, b) =>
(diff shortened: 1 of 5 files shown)


jtpio:
Thanks!

Changes look good. Curious if you were able to try it with https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/32 and the `--extensions-in-dev-mode` flag?

fcollonval:
> Changes look good. Curious if you were able to try it with [jupyterlab-contrib/jupyterlab-filesystem-access#32](https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/32) and the `--extensions-in-dev-mode` flag?

Yes it works.

jtpio (approved):
Thanks!

packages/apputils/src/toolbar/factory.ts
```diff
@@ -170,41 +185,34 @@ async function setToolbarItems(
...
-            (a, b) =>
-              (a.rank ?? DEFAULT_TOOLBAR_ITEM_RANK) -
-              (b.rank ?? DEFAULT_TOOLBAR_ITEM_RANK)
-          );
-          transferSettings(newList);
+  if (listenPlugin) {
+    registry.pluginChanged.connect(async (sender, plugin) => {
+      // As the plugin storing the toolbar definition is transformed using
+      // the above definition, if it changes, this means that a request to
+      // reloaded was triggered. Hence the toolbar definitions from the other
```
afshin, line 192:
```suggestion
      // reload was triggered. Hence the toolbar definitions from the other
```

packages/apputils/src/toolbar/factory.ts
```diff
@@ -170,41 +185,34 @@ async function setToolbarItems(
...
-              (a.rank ?? DEFAULT_TOOLBAR_ITEM_RANK) -
-              (b.rank ?? DEFAULT_TOOLBAR_ITEM_RANK)
-          );
-          transferSettings(newList);
+  if (listenPlugin) {
+    registry.pluginChanged.connect(async (sender, plugin) => {
+      // As the plugin storing the toolbar definition is transformed using
+      // the above definition, if it changes, this means that a request to
+      // reload was triggered. Hence the toolbar definitions from the other
+      // plugins has been automatically reset during the transform step.
```
afshin, line 193:
```suggestion
      // plugins have been automatically reset during the transform step.
```

packages/settingregistry/src/settingregistry.ts
```diff
@@ -1195,39 +1195,18 @@ export namespace SettingRegistry {
...
-          const refIndex = items.findIndex(ref => ref.name === item.name);
-          if (refIndex < 0) {
-            items.push({ ...item });
-          } else {
-            if (warn) {
-              console.warn(`Toolbar item '${item.name}' is duplicated.`);
-            }
-            items[refIndex] = { ...items[refIndex], ...item };
-          }
+      // Name must be unique so it sufficient to only compare it
```
afshin, line 1198:
```suggestion
      // Name must be unique so it's sufficient to only compare it
```

jtpio:
@afshin looks like all the comments have been addressed, and this should be good to go now?

fcollonval:

PR #12571, opened 2022-05-12
Building extensions fail if not using latest patch

ajbozarth:
Due to my changes in https://github.com/jupyterlab/jupyterlab/pull/12533
which were included in v3.4.1 extension builds may fail without an explicit
upgrade to the new patch (should not be necessary for patches).

This is because my change will compare against the exact version installed
rather than the range defined in the package.json, this update makes the
new check conditional to the original failure it was addressing.

## References

Original source of the bug https://github.com/jupyterlab/jupyterlab/pull/12533

### DIFF
--- a/jupyterlab/federated_labextensions.py
+++ b/jupyterlab/federated_labextensions.py
@@ -288,9 +288,13 @@ def _ensure_builder(ext_path, core_path):
 
     overlap = _test_overlap(depVersion1, depVersion2, drop_prerelease1=True, drop_prerelease2=True)
     if not overlap:
-        with open(osp.join(target, "node_modules", "@jupyterlab", "builder", "package.json")) as fid:
+        with open(
+            osp.join(target, "node_modules", "@jupyterlab", "builder", "package.json")
+        ) as fid:
             depVersion2 = json.load(fid).get("version")
-        overlap = _test_overlap(depVersion1, depVersion2, drop_prerelease1=True, drop_prerelease2=True)
+        overlap = _test_overlap(
+            depVersion1, depVersion2, drop_prerelease1=True, drop_prerelease2=True
+        )
 
     if not overlap:
         raise ValueError(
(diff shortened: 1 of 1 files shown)


ajbozarth:
@fcollonval we should get this into a v3.4.2 asap to prevent any dev thrash. Currently if you have Jupyterlab 3.4.0 pip installed and then build your extension the extension will pull in 3.4.1 dependencies during the build and `@juyterlab/builder` will error since 3.4.0 is installed even though that difference should be fine because it checks against 3.4.0 specifically instead of ^3.4.0 or ~3.4.0 that are usually set in the dependencies.

jtpio:
Thanks for the quick fix!

Also noticed this in https://github.com/voila-dashboards/voila/issues/1155 and https://github.com/jupyterlite/xeus-python-kernel/pull/5 today.

Backporting to `3.4.x` and making a new `3.4.2` release right after sounds good :+1:

jtpio:
@ajbozarth do you feel like making the `3.4.2` release with the releaser? Otherwise no problem it could be another time.

ajbozarth:
Yeah I can do the 3.4.2 release, it'll be my first time doing it myself (rather than watching someone doing it on a call), but iirc it's pretty straightforward and I should have all the correct perms.

If you approve this I can merge it once the CI finishes then backport it and merge that after it's CI passes then I'll start the release. (I should have enough time left in my day to do all that)

jtpio:
Sounds good.

For reference the diff in https://github.com/jupyterlab/jupyterlab/pull/12533/files shows the following CodeQL warning, which could have given a hint (not sure the warning was there at the time of the PR though):

image

jtpio (approved):
Thanks!

ajbozarth:
yeah I saw that warning too, it seems to have shown up after we merged the pr

ajbozarth:
The test failures are unrelated, merging

ajbozarth:

PR #12581, opened 2022-05-13
Unbump the major version of rendermime-interfaces

fcollonval:
## References

As discussed at JupyterLab meeting

## Code changes

Unbump @jupyterlab/rendermime-interfaces to 3.8.0 to keep some space for minor releases.

Make LabIcon and DocumentRegistry extends those interfaces to reduce possible future breakage.

## User-facing changes

N/A

N/A


jtpio (approved):
Thanks!

PR #12585, opened 2022-05-16
Invert relationship between launcher and filebrowser

fcollonval:
## Code changes

Make the relationship from filebrowser to launcher rather than the opposite. This is more consistent for UI without launcher.

## User-facing changes

N/A

Drop redundant `filebrowser:create-main-launcher` command; use `launcher:create` instead.


3coins (changes requested):

fcollonval:

ajbozarth (approved):
Code LGTM, and it worked as expected when run locally

3coins (approved):

PR #12586, opened 2022-05-16
Use customizable toolbar for topbar

fcollonval:
## References

Use customizable toolbar for topbar

## Code changes

Inject a customizable toolbar next to the main menu. It only contains a spacer by default.

## User-facing changes

Top bar is now more easily customizable

N/A


jtpio:
Thanks for starting this!

Linking to https://github.com/jupyterlab/jupyterlab/issues/11589 for reference.

packages/application-extension/src/topbar.ts
```diff
@@ -0,0 +1,48 @@
...
+  setToolbar
+} from '@jupyterlab/apputils';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
+import { ITranslator, nullTranslator } from '@jupyterlab/translation';
+import { Toolbar } from '@jupyterlab/ui-components';
+
+const TOPBAR_FACTORY = 'TopBar';
+
+/**
+ * The default JupyterLab application shell.
```
jtpio, line 17:
```suggestion
 * A plugin adding a toolbar to the top area.
```

packages/application-extension/style/base.css
```diff
@@ -8,7 +8,12 @@
   width: calc(var(--jp-private-sidebar-tab-width) + var(--jp-border-width));
 }
 
-#jp-topbar-spacer {
-  flex-grow: 1;
-  flex-shrink: 1;
+#jp-top-bar {
+  --jp-private-toolbar-height: var(--jp-private-menu-panel-height);
```
jtpio, line 12: Should this be moved to a `:root` definition, similar to other private variables like:

https://github.com/jupyterlab/jupyterlab/blob/ce09e0daabd48e4e90c13234137cbe4c6783eea0/packages/application/style/menus.css#L10-L13

packages/user-extension/schema/user-menu-bar.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "User Menu Bar",
+  "description": "User Menu Bar settings.",
+  "jupyter.lab.toolbars": {
+    "TopBar": [
+      {
+        "name": "user-menu",
+        "rank": 100
```
jtpio, line 8: Wondering if the rank should be `>900` here? If the user menu is supposed to be positioned in the top right corner.

jtpio: Will double check on Binder once it has finished building.

jtpio: Ah it's the toolbar that is at rank `900`. But the spacer is at rank `50`.

Probably we'll want to update that part of the documentation as well then: https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html?#top-area

jtpio:
The "Visual Regression Documentation" check seems to be failing with other PRs as well and doesn't look related:

image

jtpio (approved):
Thanks!

fcollonval:
> The "Visual Regression Documentation" check seems to be failing with other PRs as well and doesn't look related:
> 
> image

Seems we get a new kind of behavior, the thumbnails seems to not be downloaded.

image

packages/application-extension/style/base.css
fcollonval: I did not move those because we don't want to modify all the styles but only overwrite the default for that specific toolbar.

PR #12591, opened 2022-05-17
Use error names instead of messages for validation

afshin:
This PR has client code check for `Error.prototype.name` values instead of `Error.prototype.message` in order to be less brittle if an error message is changed or translated in the future.

## References
N/A

## Code changes
Updates business logic to check against specified error names.

## User-facing changes
N/A

N/A


fcollonval (approved):
LGTM

PR #12593, opened 2022-05-17
Remove deprecated SplitPanel

afshin:
Removes deprecated `SplitPanel` that had a handle moved signal in favor of the upstream version newly released in Lumino.

## References
https://github.com/jupyterlab/lumino/pull/301

## Code changes
- Bumps `@lumino/application`, `@lumino/datagrid`, and `@lumino/widgets`
- Removes custom `SplitPanel` from setting editor package

## User-facing changes
N/A

N/A


fcollonval (approved):
Thanks @afshin

packages/settingeditor/src/splitpanel.ts
```diff
@@ -1,26 +0,0 @@
...
-|----------------------------------------------------------------------------*/
-
-import { ISignal, Signal } from '@lumino/signaling';
-import { SplitPanel as SPanel } from '@lumino/widgets';
-
-/**
- * A deprecated split panel that will be removed when the phosphor split panel
- * supports a handle moved signal. See https://github.com/phosphorjs/phosphor/issues/297.
- */
-export class SplitPanel extends SPanel {
```
jtpio, line 13: Since this was exported before, maybe we should tag this PR as `api-change` and mention it in https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-3-x-to-4-x?

afshin: It was not exported in the `index.ts` for `settingeditor`, it was not intended to be public API. A user would have had to import the file directly from `lib` or something ... I don't consider this a public API, though. What do you think?

The intention was always to remove this so it wasn't exposed.

jtpio: Ah indeed, I just checked in the code base and it is indeed the case.

Thanks for the context!

jtpio:
Thanks!

PR #12595, opened 2022-05-18
Fix collaborative document cleanup

davidbrochart:
## Code changes

In collaborative mode:
- fixed the condition to know if the room is empty,
- added a `collaborative_document_cleanup_delay` option to set the delay to keep a document in memory in the back-end after all clients disconnect (`None` to keep the document forever, defaults to 60 seconds).

## User-facing changes

The new `--LabApp.collaborative_document_cleanup_delay=1` option.

None.


fcollonval (approved):
Thanks @davidbrochart

PR #12604, opened 2022-05-20
Fix `documentsearch-extension` plugin id

jtpio:
## References

To follow the same convention as for the other plugins.

## Code changes

Rename `@jupyterlab/documentsearch:labShellWidgetListener` to `@jupyterlab/documentsearch-extension:labShellWidgetListener`.

## User-facing changes

None

A plugin rename means that some previous commands to disable plugins would need to be updated.


fcollonval (commented):
LGTM

We probably want to mention it in the migration guide.

jtpio:
> We probably want to mention it in the migration guide.

Sounds good, will update the PR with this.

fcollonval (approved):
Thanks @jtpio

jtpio:
Thanks @fcollonval for the fix.

CI failure (check links) is not related:

```
Error: /home/runner/work/jupyterlab/jupyterlab/.jupyter_releaser_checkout/docs/source/user/extensions.rst: https://2.python-requests.org/en/v2.7.0/api/#requests.request
```

PR #12606, opened 2022-05-21
Switch to hatch backend

blink1073:
## References
cf https://github.com/jupyterlab/team-compass/issues/146

## Code changes
Updates python packaging infrastructure.

LSP packages and related plugins are explicitly removed from examples _app_ and _federated extensions_.
The skip `lsp/status` request in logger for examples has been removed.

## User-facing changes
None

None

## Steps

- [x] Move `hatch_jupyter_builder` to `jupyterlab` [org](https://github.com/jupyterlab/team-compass/issues/146).
- [x] Get `hatch_jupyter_builder` on [`conda-forge`](https://github.com/conda-forge/staged-recipes/pull/19042).
- [x] Write a helper script similar to the extension upgrade script to migrate packages.
   - The script will use `hatch init --new` to do the heavy lifting
   - It will also move `flake8` config from `setup.cfg` to a new `.flake8` file
   - It will migrate `jupyter-packaging` config in `pyproject.toml` to equivalent `hatch_jupyter_builder` config
   - It will look for `npm_builder` config in `setup.py` and convert it to `editable-build-kwargs` config for `hatch_jupyter_builder`
   - It will look for ignore patterns in `.gitignore` that are in the package folder and suggest them as `artifacts` config for inclusion (commented out with a note)
   - Make that script available as part of `hatch_jupyter_builder`.  It should operate in a temporary virtual environment so it doesn't add any requirements to `hatch_jupyter_builder`.  [See docs](https://github.com/jupyterlab/hatch-jupyter-builder#migration)
- [x] Use that script to migrate `extension-cookiecutter-ts`
- [x] Test this PR against a federated extension made from the [migrated cookiecutter extension](https://github.com/jupyterlab/extension-cookiecutter-ts/pull/222).  Also make sure new extensions work with existing versions of `jupyterlab` - need a shim `setup.py` for older JupyterLabs.
- [x] Prepare a PR for [`extension-cookiecutter-js`](https://github.com/jupyterlab/extension-cookiecutter-js/pull/41)
- [x] Prepare a PR for [`widget-ts-cookiecutter`](https://github.com/jupyter-widgets/widget-ts-cookiecutter/pull/117)
- [x] Get this PR passing
- [ ] Draft a blog post about the switch and how to migrate existing extensions
- [ ] Merge all open PRs and publish the blog post
- [ ] Deprecate Jupyter-Packaging, pointing to the blog post


blink1073:
@fcollonval I'm not super familiar with the Galata tests, any idea what is going wrong there?

blink1073:
I think it might have been due to differences in files including in the sdists.  I have one more issue to work though - the `.js.map` files are getting included in the `data` folder, reported as https://github.com/pypa/hatch/issues/265.

blink1073:
WIP migration [script](https://gist.github.com/blink1073/e77218a824af4ace9711f6de5b226092)

blink1073:
Update: it looks like this approach is going to work.  I've tested `ipywidgets` and an extension made from the [cookiecutter PR](https://github.com/jupyterlab/extension-cookiecutter-ts/pull/222) against both this branch and 3.4.2.

Next steps as I see them are to:

- Create draft PRs to update `extension-cookiecutter-js` and `widget-ts-cookiecutter`
- Draft a blog post about the switch and how to migrate existing extensions
- Merge the cookiecutter PRs and publish the blog post

jupyterlab/federated_labextensions.py
```diff
@@ -424,18 +425,28 @@ def _get_labextension_metadata(module):
     except Exception as exc:
         errors.append(exc)
 
+    # Try to get the package name
+    package = None
+
+    # Try getting the package name from pyproject.toml
+    if os.path.exists(os.path.join(mod_path, "pyproject.toml")):
```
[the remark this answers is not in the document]
blink1073: I ended up adding a setup.py shim to projects to avoid needing to backport fixes

blink1073:
Moving this out of draft, as the blog post and cookiecutter changes are dependent on this change but the change itself only affects JupyterLab.

blink1073:
Actually, the cookiecutters don't depend on this change since they include a `setup.py` shim, so really it is about merging this so it doesn't keep getting merge conflicts as `setup.cfg` changes.

jtpio:
Just rebased the branch to fix the conflict with `setup.cfg`.

fcollonval:
It seems that @jupyterlab/lsp is bringing troubles:

```
Failed at the @jupyterlab/lsp@4.0.0-alpha.11 build script.
```

pyproject.toml
```diff
@@ -2,45 +2,224 @@
...
+
+[project.scripts]
+jlpm = "jupyterlab.jlpmapp:main"
+jupyter-lab = "jupyterlab.labapp:main"
+jupyter-labextension = "jupyterlab.labextensions:main"
+jupyter-labhub = "jupyterlab.labhubapp:main"
+
+[project.urls]
+Homepage = "https://jupyter.org"
+
```
fcollonval, line 63: Let's add more URL (ref https://pypi.org/project/links-demo/):

```suggestion
Changelog = "https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html"
Documentation = "https://jupyterlab.readthedocs.io"
Source = "https://github.com/jupyterlab/jupyterlab"
Issues = "https://github.com/jupyterlab/jupyterlab/issues/new/choose"
Gitter = "https://gitter.im/jupyterlab/jupyterlab"
Pypi = "https://pypi.org/project/jupyterlab"

```

fcollonval (commented):
Thanks @blink1073 

The code looks good (I have an optional suggestion to add more URLs).

fcollonval:
Thanks @jtpio for fixing the CI

jtpio:
More errors that might also be related to lsp:

```
[W 2022-08-04 09:02:37.426 ServerApp] 404 GET /foo/lsp/status?1659603757274 (127.0.0.1) 138.87ms referer=http://localhost:8888/foo/lab?reset=
Waiting for #browserTest selector...
Waiting for application to start...
Parsed an error from text content: undefined Could not fetch sessions
```

fcollonval:
For the examples a skip condition was implemented:

https://github.com/jupyterlab/jupyterlab/blob/a650dd8ac4fefe42d9a3b38a702b8660adf94bc3/examples/example.spec.ts#L21

I don't know why we are being it in browser_check now. Probably the request is done faster than before or we got lucky before and Playwright was successful before the request was made.

blink1073:
Perhaps the failure is related to the failures I started seeing in [update-snapshots](https://github.com/jupyterlab/maintainer-tools/runs/7664161390?check_suite_focus=true) in `maintainer-tools`.

fcollonval:
@trungleduc in the CI the schema is failing to generate (it falls back to the default schema content); see https://github.com/jupyterlab/jupyterlab/runs/7674913771?check_suite_focus=true. Do you know what could be done for fixing it? Does the fallback is appropriate; should it skips updating `schema.ts` if it exists but the generation script failed to get the reference `schema.json`?

pyproject.toml
```diff
@@ -2,45 +2,230 @@
...
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+]
+dependencies = [
+    "importlib-metadata>=3.6;python_version<\"3.10\"",
+    "importlib-resources>=1.4;python_version<\"3.9\"",
+    "ipykernel",
+    "jinja2>=3.0.3",
+    "jupyter_core",
```
trungleduc, line 41: @fcollonval  the lsp server extension is missing.

```suggestion
    "jupyter_core",
    "jupyter-lsp>=1.5.1",
```

trungleduc:
> For the examples a skip condition was implemented:
> 
> https://github.com/jupyterlab/jupyterlab/blob/a650dd8ac4fefe42d9a3b38a702b8660adf94bc3/examples/example.spec.ts#L21
> 
> I don't know why we are being it in browser_check now. Probably the request is done faster than before or we got lucky before and Playwright was successful before the request was made.

We should be able to remove this check after adding `jupyter-lsp` to the dependency list

fcollonval:
Thanks Trung - let's see if the CI is happy. Then I'll clean up the exceptions for `lsp/status` request.

trungleduc:
The issue with the schema file is that the build process builds the frontend packages before installing the python dependencies. So the schema file is generated without `jupyter-lsp` installed.

packages/lsp/package.json
```diff
@@ -29,7 +29,7 @@
     "style/index.js"
   ],
   "scripts": {
-    "build": "jlpm build:schema && tsc -b",
+    "build": "npm run build:schema && tsc -b",
```
trungleduc, line 32: Removing the build schema step in `build` would help. Since we already have a test for obsolete schemes on CI, I think we don't need to re-generate these files on every build.
```suggestion
    "build": "tsc -b",
```

jtpio:
CI is now passing.

fcollonval:
@jtpio would you mind having a look? It is good to go for me. But as I did some changes I'll would appreciate another review.

jtpio:
Just checked the releaser built assets and they look good.

However binder fails to build: 

image

fcollonval:
I pushed a fix for the binder configuration.

fcollonval (approved):
Thanks a lot @blink1073

jtpio:
> I pushed a fix for the binder configuration.

Nice thanks @fcollonval :+1:

jtpio:
We can make a new 4.0 pre-release right after merging this, since there are also other (rather big) changes that have not been released yet.

Issue #12157, opened 2022-03-03
Add a Side-by-side ratio to the settings


echarles:
Follow-up of https://github.com/jupyterlab/jupyterlab/issues/12106

The ratio as set via the command palette is not persistent. We should convert it to a setting.

echarles:
That setting should accept any demical number to allow input bigger than then output. The input form should be adapted accordingly.

andrewfulton9:
> That setting should accept any demical number to allow input bigger than then output. **The input form should be adapted accordingly.**

@echarles, Does this mean that there should still be a command ID to update the ratio? So it would be able to be updated via the standard settings page, and also through the command ID form that comes up now, meaning the command id would also have to update the setting for persistence sake?

echarles:
@andrewfulton9 thinking a bit more, I don't know any setting being updated via a command. So we could just drop the command, and rely on the setting mecanism. This will be simpler and more standard.

krassowski:
Some thoughts:
- depends if you want to have this modifiable per-notebook or not. I can imagine that someone may want to have different ratios in different notebooks; if you only change it in settings then you cannot have it per notebook.
- if we keep the dialog we could have a checkbox `[ ] safe as default`
- it would be cool if the input was a slider with a midpoint indicating 50%; probably easier to accomplish in a dialog than in settings (but not impossible)

echarles:
> ...  I don't know any setting being updated via a command. So we could just drop the command, and rely on the setting mecanism. This will be simpler and more standard.

@andrewfulton9 this is not the case, there are plenty of commands that change the setting (exposed via menu or via the command palette), see e.g. the theme change.

![Untitled4](https://user-images.githubusercontent.com/226720/159856312-aa688373-e192-43d5-92f4-a1580bd70268.gif)

> depends if you want to have this modifiable per-notebook or not. I can imagine that someone may want to have different ratios in different notebooks; if you only change it in settings then you cannot have it per notebook.

The idea was I think to just have a global single value for all notebooks. I can understand the advantages to override that user setting value with a value that would be stored in the notebook ipynb metadata. However, it my value is a good fit for my screen, it may not be the case for your screen, so there are also disavantages to that approach.

> if we keep the dialog we could have a checkbox [ ] safe as default

or `[ ] save in my settings`.

> it would be cool if the input was a slider with a midpoint indicating 50%; probably easier to accomplish in a dialog than in settings (but not impossible)

Do we already have such a slider widget to be reused?

PR #12608, opened 2022-05-23
Add a persistent side-by-side ratio setting

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12157

## Code changes

Add a `sideBySideOutputRatio` in the notebook setting and use that value in the popup form when user wants to change the ratio via the command palette.

## User-facing changes

Untitled5

None


echarles:
CI is green. This is ready for review.

packages/notebook-extension/schema/tracker.json
```diff
@@ -922,6 +922,12 @@
       "description": "Side-by-side right margin override.",
       "type": "string",
       "default": "10px"
+    },
+    "sideBySideOutputRatio": {
+      "title": "Side-by-side output ratio",
+      "description": "For the side-by-side rendering, the side-by-side output ratio defines the width of the output vs the input. Set 1 for same size, > 1 for larger output, < 1 for smaller output.",
+      "type": "number",
+      "default": 1
```
fcollonval, line 930: Minor enhancement:

```suggestion
      "default": 1,
      "exclusiveMinimum": 0
```

echarles:
@fcollonval I have merged your suggestion and now it was failing with

```
[W 2022-05-24 08:27:47.689 BrowserApp] Failed validating schema (@jupyterlab/notebook-extension:tracker): 0 is not of type 'boolean'
    
    Failed validating 'type' in metaschema['properties']['properties']['additionalProperties']['properties']['exclusiveMinimum']:
        {'default': False, 'type': 'boolean'}
    
    On schema['properties']['sideBySideOutputRatio']['exclusiveMinimum']:
        0
```


So I made it a boolean, and now I have another error:

````
[W 2022-05-24 08:46:20.558 ServerApp] 500 GET /foo/lab/api/settings?1653381980433 (127.0.0.1): Failed validating schema (@jupyterlab/notebook-extension:tracker): 'minimum' is a dependency of 'exclusiveMinimum'

...
   On schema['properties']['sideBySideOutputRatio']:
        {'default': 1,
         'description': 'For the side-by-side rendering, the side-by-side '
                        'output ratio defines the width of the output vs the '
                        'input. Set 1 for same size, > 1 for larger output, < '
                        '1 for smaller output.',
         'exclusiveMinimum': True,
         'title': 'Side-by-side output ratio',
```

fcollonval:
That means we are using Draft 4 :sob: 
So I advice to bump ajv to the latest version and stick with `exclusiveMinimum: number`. But if you plan to backport this one, I would argue not to bump ajv as it will be on a patch release. And therefore you need to set `minimum: number` and `exclusiveMinimum: boolean` - see: https://json-schema.org/understanding-json-schema/reference/numeric.html

echarles:
> That means we are using Draft 4 😭

The reported errors are from the GitHub CI jobs.

echarles:
> So in fact you need to set the exclusiveMinimum to a boolean (and therefore set minimum) because the JSON schema constrained comes from jupyterlab_server:

I have pushed and the previous errors dissapear. CI was green, but with that changes it is constantly failing on 5+ flows and I can make sense of what is going wrong based on the log. 

```
      "minimum": 0,
      "exclusiveMinimum": true
```

fcollonval:
I suspect the json schema validator is not consistent at some place and so it uses either draft04 at some place and more recent version at others. So I would suggest setting only `minimum` attribute.

echarles:
@fcollonval I have pushed your suggestion. I see now 2 failing tests, which are unrelated IMHO.

fcollonval:
Thanks @echarles 
Do you know why this PR is showing so much changes? Would you mind to rebase it?

echarles:
This branch has issues. Closing in favor of follow-up https://github.com/jupyterlab/jupyterlab/pull/12633

PR #12374, opened 2022-04-07
Table of content refactor

fcollonval:
## References

This is a preliminary work for https://github.com/jupyterlab/jupyterlab/pull/11574 that switches the search feature in the notebook from view (aka CodeMirror and DOM nodes) search to data model (aka modelDB) for the inputs. The search on outputs is disabled by default and will looked at the rendered DOM tree if activated (when using virtual rendering this will trigger a warning that it may takes some time as outputs rendering will be needed).

Closes https://github.com/jupyterlab/jupyterlab/issues/11663
Fixes https://github.com/jupyterlab/jupyterlab/issues/10457 - add base numbering support and widget for edition

## Code changes

Rework fully the table of contents feature to be purely based on the document models for extracting the headings.

- This uses a registry to which ToC model for widget can be registered.
- The model provides the heading list (no more possibility to provide custom element or html string) - only text is supported
  - Models are defined in the document package (fileeditor, notebook and markdownviewer).
- Style
  - Items are layout in an ordered list of ordered lists
  - Use tree view (with left arrow to indicate collapsing status) - the click on the blue line and the ellipsis button were not aligned with the behavior used within the notebook for collapsing headers.
- All documents:
  - New settings to control ToC depth (4 by default)
- notebook ToC:
  - Dropping display of Markdown content or source content
  - Dropping tag filtering (this will be implemented in a companion PR to apply on the notebook panel directly).
  - Dropping synchronization with notebook metadata (view state should not be stored as metadata - it won't play nicely with RTC)
  - Add support for notebook metadata `toc.base_numbering` for compatibility with classical nb toc2
  - Add support to skip heading if it contains a html tag with class `jp-toc-ignore` or `tocSkip` (for compatibility with nb toc2)
- python ToC:
  - Fix heading level to respect code nesting structure

TODO
- [x] Fix heading run cells
- [x] Update documentation
- [x] Fix integration tests

## User-facing changes

- Consistent styling for all table of contents whatever the type of document
- Consistent set of feature from the table of contents

## Backwards-incompatible changes

The API is totally new.

PR #12612, opened 2022-05-24
Break loop activeCell -> activeHeading

fcollonval:
## References

Break the triggering of `activeHeadingChanged` following `activeCellChanged`

## Code changes

Replace `activeHeading` setter by `setActiveHeading` with optional argument to trigger `activeHeadingChanged`.

## User-facing changes

Stop scrolling to another cell than the active one.

N/A


fcollonval:

jtpio (approved):
Thanks!

packages/toc/src/tokens.ts
```diff
@@ -225,6 +225,14 @@ export namespace TableOfContents {
      */
     isActive: boolean;
 
+    /**
+     * Set a new active heading.
+     *
+     * @param heading The new active heading
+     * @param emitSignal Whether to emit the activeHeadingChanged signal or not.
+     */
+    setActiveHeading(heading: H | null, emitSignal?: boolean): void;
```
jtpio, line 234: Note to self: since this is a follow-up to https://github.com/jupyterlab/jupyterlab/pull/12374 we don't need to document the change in the public interface.

PR #12619, opened 2022-05-27
Fix kernel protocol serialization

davidbrochart:
## References

Fixes https://github.com/jupyter-server/jupyter_server/issues/857.

## Code changes

In the serialization of kernel messages over WebSocket `v1.kernel.websocket.jupyter.org`, we were computing offsets directly on the `JSON.stringified` objects, but then we use a `TextEncoder` to write the objects in binary. The problem is that with the former, emojis are kept as-is (not encoded), and with the latter they are encoded, resulting in a bigger size:
```js
'✅'.length
// 1
```
```js
encoder = new TextEncoder();
encoder.encode('✅').length
// 3
```
In this PR we compute the offsets on the encoded objects.

## User-facing changes

None.

None.

cc @echarles @jasongrout

### DIFF
--- a/packages/services/src/kernel/serialize.ts
+++ b/packages/services/src/kernel/serialize.ts
@@ -109,11 +109,41 @@ namespace Private {
     let offsets: number[] = [];
     offsets.push(8 * (1 + offsetNumber));
     offsets.push(msg.channel.length + offsets[offsets.length - 1]);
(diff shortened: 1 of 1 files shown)


echarles (approved):
LGTM Tested wit jupyter-server HEAD and works fine. The errors don't arise when printing an emoji.

Issue #11958, opened 2022-01-28
Drop the dependency on `ipykernel`


jtpio:
### Problem

Currently `jupyterlab` doesn't **directly** depend on `ipykernel`.

The upcoming drop of the `nbclassic` dependency for JupyterLab 4.0 started in https://github.com/jupyterlab/jupyterlab/pull/11894 has the side-effect of also dropping the transitive `ipykernel` dependency:

`jupyterlab` -> `nbclassic` -> `notebook` -> `ipykernel`

This can also be inspected in a new environment:

```bash
mamba create -n tmp -c conda-forge jupyterlab=3 pipdeptree -y
conda activate tmp
pipdeptree -ar
```

Gives:

![image](https://user-images.githubusercontent.com/591645/151561989-03a2039a-faff-4b0c-a757-61ae733cea9d.png)

### Proposed Solution

There are two options:

1. We add `ipykernel` as an explicit dependency for `jupyterlab`. This means users can expect to find a kernel ready to use on `pip install jupyterlab`
2. We don't add `ipykernel` as an explicit dependency for `jupyterlab`. End users will have to install `ipykernel` manually when installing `jupyterlab`, for example with:

```
pip install jupyterlab ipykernel
mamba install -c conda-forge jupyterlab ipykernel
```

Since Jupyter is language agnostic, option 2 is cleaner and lets users choose the kernel(s) they want from the start. For example they could be using `xeus-python` instead of `ipykernel` for the Python language.

Option 1 is however more convervative, which might be expected by users who want things "that just work".

### Additional context

Noticed when dropping `nbclassic`: https://github.com/jupyterlab/jupyterlab/pull/11894#issuecomment-1024011925

Related: https://github.com/jupyterlab/jupyterlab/pull/11334

echarles:
I like things "that just work".

Zsailer:
I think, for historical reasons, we have to require ipykernel.

jtpio:
`ipykernel` has been added as an explicit dependency in https://github.com/jupyterlab/jupyterlab/pull/11894 for now

vidartf:
+1. It makes sense for me that the JupyterLab *python* package includes ipykernel. Low cost, high reward 👍

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/11894, since `ipykernel` is now an explicit dependency in JupyterLab:

[...]

PR #12625, opened 2022-05-30
Install either jupyter-server or jupyverse

### LINKED ISSUE #11101: Make jupyterlab_server optional
### Problem

JupyterLab currently depends on [jupyterlab server](https://github.com/jupyterlab/jupyterlab_server). For alternative servers to the JupyterLab frontend, such as [jupyverse](https://github.com/jupyter-server/jupyverse), this dependency is really not necessary, and makes the installation more problematic.

### Proposed Solution

`jupyterlab_server` should either be an optional dependency, that would be installed e.g. with:

```bash
pip install jupyterlab[jupyter_server]  # would install jupyterlab_server
```

or we could keep it as the default server (i.e. `jupyterlab_server` would be installed when `pip install jupyterlab`), but it wouldn't be installed when another server is specified:

```bash
pip install jupyterlab[jupyverse]  # would not install jupyterlab_server, but would install jupyverse instead
```

davidbrochart:
## References

Closes #11101.

## Code changes

Now that we have an official alternative Jupyter server ([jupyverse](https://github.com/jupyter-server/jupyverse)), JupyterLab should not strictly depend on [jupyter-server](https://github.com/jupyter-server/jupyter_server). The user should be able to install the back-end of their choice. This PR makes `jupyter-server` and `jupyverse` optional dependencies.

## User-facing changes

JupyterLab should be installed with:
```bash
pip install jupyterlab[jupyter-server]
# or
pip install jupyterlab[jupyverse]
```

Installing with `pip install jupyterlab` won't be enough, and if no Jupyter server is installed, JupyterLab will show at startup:
```
No server installed, please install either jupyter-server or jupyverse.
```

### DIFF
--- a/.github/workflows/galata-update.yml
+++ b/.github/workflows/galata-update.yml
@@ -100,7 +100,7 @@ jobs:
           # Freeze the packages to ensure consistent look and feel
           # IPython is frozen because its version is displayed in
           # the console header
-          pip install .[docs-screenshots]
+          pip install .[jupyter-server, docs-screenshots]
           bash ./scripts/ci_install.sh
 
           # Build dev-mode
--- a/.github/workflows/galata.yml
+++ b/.github/workflows/galata.yml
@@ -114,7 +114,7 @@ jobs:
           # Freeze the packages to ensure consistent look and feel
           # IPython is frozen because its version is displayed in
           # the console header
-          pip install .[docs-screenshots]
+          pip install .[jupyter-server, docs-screenshots]
           bash ./scripts/ci_install.sh
 
           # Build dev-mode
--- a/.github/workflows/linuxtests.yml
+++ b/.github/workflows/linuxtests.yml
@@ -132,7 +132,7 @@ jobs:
           mkdir test
           tar --strip-components=1 -zxvf *.tar.gz -C ./test
           cd test
-          pip install -e .[test]
+          pip install -e .[jupyter-server, test]
           pip install pytest-github-actions-annotate-failures
       - name: Run Test
         run: |
@@ -147,5 +147,5 @@ jobs:
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
       - name: Install Jupyterlab
-        run: pip install -e .
+        run: pip install -e .[jupyter-server]
       - uses: pre-commit/action@v2.0.0
--- a/jupyterlab/__init__.py
+++ b/jupyterlab/__init__.py
@@ -4,14 +4,17 @@
 # Distributed under the terms of the Modified BSD License.
 
 from ._version import __version__  # noqa
-from .serverextension import load_jupyter_server_extension  # noqa
 
+try:
+    from .serverextension import load_jupyter_server_extension  # noqa
 
-def _jupyter_server_extension_paths():
-    return [{"module": "jupyterlab"}]
+    def _jupyter_server_extension_paths():
+        return [{"module": "jupyterlab"}]
 
+    def _jupyter_server_extension_points():
+        from .labapp import LabApp
 
-def _jupyter_server_extension_points():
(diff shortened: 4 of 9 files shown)


jtpio:
Thanks @davidbrochart for proposing this.

Not being able to run JupyterLab anymore after a single `pip install jupyterlab` sounds a bit disruptive, similar to the previous question about whether to drop the dependency on `ipykernel`: https://github.com/jupyterlab/jupyterlab/issues/11958

davidbrochart:
The other option I can think of is having a new package `jupyterlab-frontend`, that would mostly be the current `jupyterlab` package, but without the dependency to `jupyter-server`. The new `jupyterlab` package would depend on `jupyter-server` and `jupyterlab-frontend`. That way `jupyverse` could depend on `jupyterlab-frontend` without depending on `jupyter-server`.
What do you think?

jtpio:
Decoupling the frontend from the backend would indeed allow for more flexibility here.

Just wondering whether introducing new packages might make the setup more complicated and increase the number of combinations. This also sounds similar to a couple of other ideas:

- extracting the build tools from jupyterlab to a separate package: https://github.com/jupyterlab/jupyterlab/issues/11336#issuecomment-962016778
- introducing metapackages to install jupyterlab distribution with specific sets of features

davidbrochart:
It would increase the number of combinations for sure (and that's the goal), but at least the "default" JupyterLab (`jupyterlab-frontend` + `jupyter-server`) would remain as easy to install.

krassowski:
@davidbrochart during JupyterLab meeting yesterday we triaged it as something that could possibly go into 4.2 if rewritten in a way that there are no backward compatible changes (this is installing `jupyterlab` still gives a working jupyter-server, and importing from `jupyterlab` and subpackages works as it used to), which could be achieved by having the proposed `jupyterlab-frontend` package (which I would strongly prefer to name `jupyterlab-core` given that it would include CLI, extension management and other non-frontend stuff) with shim imports.

davidbrochart:
But a `jupyterlab-core` with CLI and other jupyter-server-specific things will still not be a clean separation, right? In my mind, a `jupyterlab-frontend` with nothing backend-specific makes more sense.

krassowski:
> with CLI and other jupyter-server-specific things

CLI should not intrinsically jupyter-server specific, nor should be the extension manager etc. The fact that today these might import something from jupyter-server does it mean they should to nor need to. Things like configuration (traitlets or whatever) and contribution of assets (extensions/module federation) can be defined agnostically of the server and there is no reason why these should be in yet another place.

echarles:
I talked with a few users and they still want need to just `pip install jupyterlab` and have a fully featured installation with all the existing extensions and traitlets configuration, including the current jupyterlab and jupyter-server extensions.

davidbrochart:
Depends what we mean by "all the extensions" and "traitlets configuration". Users surely want configuration, but I doubt they want it to use traitlets (and its custom file format) or a standard, widely used format like YAML. Jupyter-server probably has more extensions that Jupyverse at the moment, but who is going to use Tornado in a few years and in which position are we going to find ourselves if we don't embrace new technology?
I think @krassowski is right when saying that we should define an agnostic format for defining configuration and extensions. Then users can install the server of their choice, depending on which server supports which extension.

jtpio:
@davidbrochart maybe a temporary solution could be to document in the JupyterLab docs how to use `jupyverse`? 

It looks like this would otherwise involve more discussions on how to organize the different packages (and potentially the creation of metapackages).

davidbrochart:
Thanks, I opened #16190.

krassowski:
What is the status here @davidbrochart - should we close this one since https://github.com/jupyterlab/jupyterlab/pull/16190 was merged?

davidbrochart:
Yes, let's close it.

PR #517, opened 2016-07-22
Implement input request handling

blink1073:
Fixes #495.   I'm going to hold of off writing more tests that involve the mock kernel for now, so tests will have to wait.

screen shot 2016-07-22 at 9 46 24 am

input_request

screen shot 2016-07-22 at 9 58 19 am

Latest with padding added after the prompt:
screen shot 2016-07-22 at 2 12 42 pm

PR #12659, opened 2022-06-07
Make password inputs not give away how many characters were typed

jasongrout:
## References

Adjusts the original password input implementation from #517 to not give away how many characters were typed.

## Code changes

Make password inputs not give away how many characters were typed

This aligns with classic Jupyter Notebook behavior (always show 8 dots), and is generally a good security idea.

Also, this decouples the implementation of the HTML input widget and the notion of the input being a password input by storing the password field as a private attribute of the stdin control, rather than relying on how the input was created.

## User-facing changes

Always show 8 dots as the value of the password input

Screen Shot 2022-06-06 at 18 31 35

### DIFF
--- a/packages/outputarea/src/widget.ts
+++ b/packages/outputarea/src/widget.ts
@@ -920,6 +920,7 @@ export class Stdin extends Widget implements IStdin {
     this._future = options.future;
     this._parentHeader = options.parent_header;
     this._value = options.prompt + ' ';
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @jasongrout 

> CI failure is due to flaky test and is unrelated.

fcollonval:

Issue #12616, opened 2022-05-26
Notebook execution indicator gets confused by Comms


vidartf:
## Description

The notebook execution indicator added in https://github.com/jupyterlab/jupyterlab/pull/10730 gets confused when using ipywidgets. When the kernel reacts to ipywidget activity, it will do a busy/idle cycle that is unrelated to any cell execution, yet the indicator will still react to it and count it as an execution.

## Reproduce

1. Install ipywidgets and the jupyterlab widgets manager. `pip install ipywidgets jupyterlab_widgets`
2. Create a new notebook and execute:
   ```python
   import ipywidgets
   ipywidgets.IntSlider()
   ```
3. Note the state of the execution indicator (Idle), and then move the slider around. The kernel indicator becomes busy as it should, but never returns to "Idle". Instead it now shows "Executed 0/N requests", where `N` depends on how much sliding you did.

## Expected behavior

The activity by ipywidgets should trigger a normal busy/idle cycle without considering them an "execution".

## Context

- Operating System and version: Windows 10
- Browser and version: Chrome 101.0.4951.67
- JupyterLab version: 3.4.2

vidartf:
CC some people previously involved: @trungleduc, @fcollonval, and @jasongrout (https://github.com/jupyterlab/jupyterlab/pull/10730#issuecomment-894019369)

vidartf:
Note: The status in the statusbar correctly returns to "Idle" after interacting.

vidartf:
Having looked a little closer, I think that the issue is that the new `echo_update` message triggers this conditional branch: https://github.com/jupyterlab/jupyterlab/blob/76fdc0ba76d4978c0538495a6507008d671aeb03/packages/notebook/src/executionindicator.tsx#L301

So the repro might require you to add the `--pre` flag to get the prerelease of ipywidgets 8.

vidartf:
I think an invalid assumption was made that "all Comm messages without "status" in ("request_state", "update")" == execution request.

trungleduc:
Thanks @vidartf for reporting the issue. I will take care of it.

jasongrout:
"request_state" and "update" are very specific messages from ipywidgets. Those values should not be hardcoded in JupyterLab - that would be very brittle and prone to breaking if/when we change the ipywidgets protocol (such as when we added the echo_update message), and it gives special treatment to ipywidgets where we ought not have favorites.

trungleduc:
[...]

PR #12665, opened 2022-06-07
Remove ipywidgets message count in the execution indicator model

trungleduc:
## References
Closes https://github.com/jupyterlab/jupyterlab/issues/12616

## Code changes

- Remove `ipywidgets` execution message count in the indicator model
- Fall back to  the kernel status if the scheduled cell list is empty

## User-facing changes

- Executions invoked by widgets are no longer counted in the execution indicator, but the kernel status is still shown.

indicator

N/A

### DIFF
--- a/packages/notebook/src/executionindicator.tsx
+++ b/packages/notebook/src/executionindicator.tsx
@@ -117,7 +117,7 @@ export function ExecutionIndicatorComponent(
     return reactElement('busy', progressBar(percentage), [
       <span key={0}>
         {trans.__(
-          `Executed ${executedCellNumber}/${scheduledCellNumber} requests`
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @trungleduc 

> CI failure is not related.

jasongrout:
Just curious, how does it behave if you combine the two execution situations, i.e., execute some cells while that button execution is still processing, or move the slider while cells are still executing?

trungleduc:
> execute some cells while that button execution is still processing

The indicator will switch to the cells counter mode once the button's callback finishes. 

widget+cell

> move the slider while cells are still executing

slider+cell

PR #12667, opened 2022-06-07
Add events service

afshin:
This PR adds support for the [newly added `api/events` endpoints in `jupyter_server`](https://github.com/jupyter-server/jupyter_server/pull/851) in `@jupyterlab/services`.

**⚠️ Note:** An earlier version of this PR added some new concepts for retrieving the service manager without retrieving it from the `JupyterFrontEnd` instance, but they will create more confusion than benefit. We should revisit the idea of how to swap out the `ServiceManager` from an extension in another PR.

## References
https://github.com/jupyter-server/jupyter_server/issues/780

## Code changes
- Exports `Event` namespace in `@jupyterlab/services` that contains only types, so `window.Event` can still be instantiated as a `new Event(...)` even if the `Event` namespace has been imported.
- Adds `EventManager` class that implements the `Event.IManager` interface to broadcast stream of emissions from the event bus.
- Fails subscription to `api/events` gracefully since `jupyter_server` 2 is not a hard requirement.
- Uses a private implementation of a stream that is both a Lumino `ISignal<Event.IManager, Event.Emission>` and an `AsyncIterable<Event.Emission>` so this can be back-ported to 3.6 and will switch to the [Lumino 2 implementation of `Stream` for 4.0](https://github.com/jupyterlab/lumino/pull/462/)

## User-facing changes
There are no user-visible changes in this PR.

There are no backward-incompatible changes.

#### Example: [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) loop
```ts
const iterable: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/application-extension:iterable',
  activate: (app: JupyterFrontEnd) => {
    void (async () => {
      const { events } = app.serviceManager;
      for await (const emission of events.stream) {
        console.log('iterable emission', emission);
      }
    })();
  },
  autoStart: true
};
```
#### Example [`AsyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols)
```ts
const iterator: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/application-extension:iterator',
  activate: ({ serviceManager }) => {
    (async () => {
      const it = serviceManager.events.stream[Symbol.asyncIterator]();
      let emission;
      while (!(emission = await it.next()).done) {
        console.log('iterator emission', emission.value);
      }
    })();
  },
  autoStart: true
};
```
#### Example: [Lumino `ISignal#connect`](https://lumino.readthedocs.io/en/1.x/api/signaling/interfaces/isignal.html)
```ts
const signal: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/application-extension:signal',
  activate: (app: JupyterFrontEnd) => {
    app.serviceManager.events.stream.connect((_, emission) => {
      console.log('signal emission', emission);
    });
  },
  autoStart: true
};
```

## Remaining `TODO` follow-on PRs
- [ ] In `EventManager#_subscribe()`, make the test emission a request with a registered schema. This requires registering a schema with the back-end application's event bus.
- [ ] Remove capability probing for `api/events` in `EventManager#_subscribe()` after `jupyter_server` 2 becomes a hard requirement.
- [ ] Write tests for `EventManager` in the `@jupyterlab/services` test suite.


afshin:
@jtpio After thinking it through and discussing it with @blink1073, I think a better model is the following (I'll update the top-level comment as well):

* `ServiceManager.getInstance(): ServiceManager.IManager | null`
  You get either the currently existing service manager or `null`.
* `ServiceManager.setInstance(manager: ServiceManager.IManager): ServiceManager.IManager`
  You can set the service manager instance (and trigger any already existing one to `dispose()`
* Since the `ServiceManager.IManager` interface is an observable disposable, this allows the `JupyterFrontEnd` to discover that its service manager has been disposed and then set it to a new service manager:
  ```typescript
  this._serviceManager =
  options.serviceManager ||
  ServiceManager.getInstance() ||
  ServiceManager.setInstance(new ServiceManager());
  this._serviceManager.disposed.connect(() => {
  this._serviceManager =
    ServiceManager.getInstance() ||
    ServiceManager.setInstance(new ServiceManager());
  });
  ```

afshin:
> Does that mean we will expect extensions and plugins to use ServiceManager.getInstance() to access the service manager? And do we encourage this practice, or just provide an alternative?

This is just another alternative to try to simplify the structure of some components and not require them to thread down the `JupyterFrontEnd#serviceManager` but it isn't prescriptive about how somebody *should* get a service manager.

afshin:
> Also wondering how this would work when a custom ServiceManager is provided to the JupyterLab application (or to a different JupyterFrontEnd). Would calling ServiceManager.getInstance() return the custom service manager instance?

Yes, in this model it would return the custom one. The idea is that perhaps an extension might want to globally replace the service manager the application uses.

packages/services/src/manager.ts
```diff
@@ -80,78 +91,78 @@ export class ServiceManager implements ServiceManager.IManager {
...
+   * The contents manager instance.
    */
-  get isDisposed(): boolean {
-    return this._isDisposed;
-  }
+  readonly contents: Contents.IManager;
 
   /**
-   * Dispose of the resources used by the manager.
+   * The nbconvert manager instance.
```
blink1073, line 99:
```suggestion
   * The event manager instance.
```

jtpio:
> **EDIT**: ah this already seems to be the case on the latest commit +1

Since this is already the case the following TODO can probably be marked as completed? 

> - [ ] Remove capability probing for api/events in EventManager#_subscribe() after jupyter_server 2 becomes a hard requirement.

afshin:
This version of the PR can be merged into the 3.6 branch, but a follow-on
PR should remove the capability probing block and should also remove the
`Private.Stream` implementation.

jtpio:
ah ok, adding to the 3.6 milestone then if it was meant to be backported.

fcollonval (changes requested):

fcollonval:
> Although the 3.6 branch does not include the changes from https://github.com/jupyterlab/jupyterlab/pull/11537.
> 
> So it's not clear whether it can still be backward compatible because that will likely require changing the `ServiceManager.IManager` interface.

The events manager can be added as optional in the interface in 3.6.x. so it stays backward compatible.

afshin:
> Although the 3.6 branch does not include the changes from #11537.
> 
> So it's not clear whether it can still be backward compatible because that will likely require changing the `ServiceManager.IManager` interface.

Maybe we should make `events` optional on that interface?

jtpio (approved):
Thanks!

jtpio:
Based on the discussions above here are a couple of follow-ups to track in separate issues and PRs:

- [ ] In `EventManager#_subscribe()`, make the test emission a request with a registered schema. This requires registering a schema with the back-end application's event bus.
- [ ] Remove capability probing for `api/events` in `EventManager#_subscribe()` after `jupyter_server` 2 becomes a hard requirement.
- [ ] Write tests for `EventManager` in the `@jupyterlab/services` test suite.
- [ ] Make `events` optional when backporting to 3.6.x

afshin:
Maybe making `EventManager` optional should be in *this* PR since it's basically already got outstanding tasks to make it 4.0-ready.

afshin:
@fcollonval Thank you for helping with the test failures!

I added a couple commits that switch to checking `PageConfig` for the server version instead of making a request (cc: @blink1073 @Zsailer) and I switched to using a `Poll` instance to manage backing off when socket connections fail so that we don't flood re-connection attempts.

packages/services/src/event/index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+      // Cause the poll to tick a rejection and back off if the socket closes.
+      socket.onclose = () => reject(new Error('EventManager socket closed'));
+      socket.onmessage = msg => msg.data && stream.emit(JSON.parse(msg.data));
+    });
+  }
+
+  // TODO: Remove this check for the `jupyter_server` version.
+  // It is only necessary in JupyterLab < 4.
+  private _isDisabled = 2 > PageConfig.getNotebookVersion()[0];
```
jtpio, line 127: Wondering whether that would work in a non-browser environment or in other frontends, since `@jupyterlab/services` can be reused elsewhere?

afshin: We already use `PageConfig` in `@jupyterlab/services` in a similar way in:

https://github.com/jupyterlab/jupyterlab/blob/9ee65ca27712b19cb8a7a2f4e99453ef1439a525/packages/services/src/serverconnection.ts#L222-L223

I don't think this PR's usage changes the requirements we already imposed in a material way. What do you think, @jtpio?

jtpio: Ah right thanks for the extra context!

fcollonval (approved):

fcollonval:

PR #12673, opened 2022-06-08
prefer new `.jp-icon-mono` class for monochromatic icons

dlqqq:
## References

- https://github.com/jupyterlab/jupyterlab/issues/12041

## Code changes

- prefer new `.jp-icon-mono` class for all monochromatic action icons
  - identical to existing `.jp-icon3` class

- adds new `.jp-icon-mono-inverse` class to handle inverse colors (e.g. the shell prompt within the terminal icon).

- replaces Python icon with one more consistent with Python's style guide (from https://github.com/elyra-ai/elyra/pull/2705)

- replaces Markdown icon with icon from dcurtis/markdown-mark

## User-facing changes

- new Markdown icon is used, along with a more standard "sky-blue" color shared by VS Code.
- new Python icon is used with correct styling according to Python style guide
- standardizes color for monochromatic action icons
- updated documentation to encourage theme developers to prefer `.jp-icon-mono` over `.jp-iconX` classes, which have a much more limited scope as most icons are monochromatic

Before:

Screen Shot 2022-06-08 at 12 41 23 PM

After:

Screen Shot 2022-06-08 at 12 41 01 PM

Most action icons had their previous CSS class removed and replaced with `.jp-icon-mono`. This will break theme extensions that rely on this. However, standardizing the CSS class we are using for monochromatic icons will also make it significantly easier for theme developers to maintain consistent icon theming throughout the JupyterLab UI.


dlqqq:
@ellisonbg Good question. There were a few icons that I chose not to touch because they were not monochromatic, so these should probably be reviewed on a case-by-case basis on how we should replace the classes associated with them. The scope of this PR is mainly for monochromatic icons and to fix the most glaring icon CSS issues visible from the JL launcher; a separate issue and PR should be made to address the larger issue of an excessive number of CSS classes being misused throughout JL.

## Icons still using old CSS classes

Uses `.jp-icon0`:

- `packages/ui-components/style/debug/bad.svg`
- `packages/ui-components/style/icons/filetype/html5.svg`

Uses `.jp-icon1`:

- `packages/ui-components/style/icons/filetype/vega.svg`
- `packages/ui-components/test/bar.svg`
- `packages/ui-components/test/foo.svg`

Uses `.jp-icon2`:

- `packages/ui-components/style/icons/jupyter/jupyterlab-wordmark.svg`

Uses `.jp-icon3`:

- `packages/ui-components/style/icons/jupyter/jupyter.svg`
- `packages/ui-components/style/unused/text-editor-alt.svg`

Uses `.jp-icon4`:

None.

Uses `.jp-icon-accent*`:

- `packages/ui-components/style/unused/text-editor-alt.svg`

Uses `.jp-icon-brand*`:

- `packages/ui-components/style/icons/filetype/image.svg`
- `packages/ui-components/style/icons/filetype/react.svg`

Uses `.jp-icon-warn*`:

- `packages/ui-components/style/icons/filetype/json.svg`
- `packages/ui-components/style/icons/filetype/notebook.svg`
- `packages/ui-components/style/icons/jupyter/jupyter-favicon.svg`
- `packages/ui-components/style/icons/jupyter/jupyter.svg`
- `packages/ui-components/style/icons/jupyter/jupyterlab-wordmark.svg`

Uses `.jp-icon-contrast*`:

- `packages/ui-components/style/icons/filetype/r-kernel.svg`
- `packages/ui-components/style/icons/filetype/spreadsheet.svg`
- `packages/ui-components/style/icons/filetype/yaml.svg`

## Key observations

- Many classes (like `.jp-icon4`) aren't even being used at all by any icon.
- These classes are not being used contextually, i.e. `.jp-icon-warn` is not being used in "warning" icons, but is instead used to just color icons orange. Same with `.jp-icon-contrast0` for coloring icons purple.

## Open question

Can we remove some (if not all) of these CSS classes and replace them with ones that are directly related to the icon?

  - Right now, a theme developer changing the color of `.jp-icon-warn*` would expect only warnings to be changed, when in fact it actually just changes the color of the Jupyter logo! This is very confusing.
  - Instead of having several poorly-named and frequently unused `.jp-icon*` classes, we can instead create new classes for each special icon that requires its own color palette. For example, we add the `.jp-icon-jupyter-logo` class to the Jupyter SVG, and then apply our desired colors via that class. This way, theme developers don't accidentally end up changing the color of icons that are intended to have their own theme-independent color.

I personally fail to see a reason we should continue exposing so many `.jp-icon*` classes and variables that don't seem to be getting much usage. I think it's worth considering migrating away from this approach and have developers explicitly specify which icons they wish to override rather than forcing disparate icons to share the same CSS classes.

fcollonval:
Thanks for tackling this @dlqqq 



> Can we remove some (if not all) of these CSS classes and replace them with ones that are directly related to the icon?
> 
>     * Right now, a theme developer changing the color of `.jp-icon-warn*` would expect only warnings to be changed, when in fact it actually just changes the color of the Jupyter logo! This is very confusing.
> 
>     * Instead of having several poorly-named and frequently unused `.jp-icon*` classes, we can instead create new classes for each special icon that requires its own color palette. For example, we add the `.jp-icon-jupyter-logo` class to the Jupyter SVG, and then apply our desired colors via that class. This way, theme developers don't accidentally end up changing the color of icons that are intended to have their own theme-independent color.

I'm :100: to reduce the number of generic classes and to use more semantic names for specific icons.

I'm also wondering if we could not simply drop using a generic class for most icons and set `fill` or `stroke` to [`currentColor`](https://www.w3.org/TR/css-color-3/#currentcolor) to simplify the handling of icon next to text (e.g. dealing with selectable icon); it brings the ease of web font icons.

dlqqq:
Another action item to add to the wider goal of eliminating excessive generic classes: as we have gathered the styles for monochromatic icons under the `.jp-icon-mono` and `.jp-icon-mono-inverse` selectors, we can actually get rid of `.jp-icon-selectable` as well from all the icons. I've tested this locally and can't notice any obvious differences. This will have to come after if/when we migrate away from the existing icon classes, since some icons (as mentioned above) still use the old `.jp-iconX` classes.

dlqqq:
Small change that had to be made: it looks like the way we're currently rendering SVG icons means that we can't directly attach a `class` attribute to the root SVG element. The fix for this was pretty simple, just had to nest child elements under a new `<g class="..." />` element.

Screen Shot 2022-06-09 at 1 37 44 PM

dlqqq:
List of CSS hacks being used that can be fixed with the `.jp-icon-override` class I suggested earlier in this thread (noticed by fixing UI regressions):

- `packages/launcher-extension/style/base.css:12` (create launcher icon)
- `packages/toc-extension/style/base.css:10` (TOC icon)
- `packages/apputils/style/commandpalette.css:105` (command palette search icon)
- `packages/ui-components/style/icons.css:{291,320}` ("busy" unsaved changes icon beneath close tab icon)
- `packages/debugger/style/icons.css:25` (debugger icon)

fcollonval:
Thanks for all the tests and report

> The icons get set to pure black because that's the default color of text in light mode. We want the icon to be a different color while unselected and have the icon be the same color while selected.

Thanks for trying.

> Rather than attaching another class to each icon, I suggest that we attach a class to the parent and use a separate selector to target icon descendants of the parent instead. We can provide a generic `.jp-icon-override` class for elements that need inline icons to always match color. In `icons.css`: 
> It's ambiguous as to how mono inverse color should be handled (e.g. the shell prompt of the terminal icon). I suggest making it transparent as above by default. If greater specificity is needed, developers can forego the `.jp-color-override` class and write their own overriding parent class to control both mono and mono inverse colors.

As you demonstrate, if we are to use a single `jp-icon-mono` class, it becomes easy to switch to `currentColor` by using a parent selector. So I would not add the additional `jp-icon-override` and let developer switch the style where they see fit. So adding your above snippet in the documentation sounds good enough.

> TL;DR: delegate responsibility of control to the parent, rather than trying to implement the functionality in the icon styles itself.

:100: 

> Small change that had to be made: it looks like the way we're currently rendering SVG icons means that we can't directly attach a class attribute to the root SVG element. The fix for this was pretty simple, just had to nest child elements under a new <g class="..." /> element.

Did you try setting `className` instead of `class`? Class is never used in JS because of its ambiguity with the language keyword `class`.

> List of CSS hacks being used that can be fixed with the .jp-icon-override class I suggested earlier in this thread

I don't get how introducing `jp-icon-override` will simplify cover those cases. Would you mind providing the code snippet?

> Comment on the code change

You should not commit `docs/source/developer/ui_components.rst` as it is generated.

dlqqq:
> I would not add the additional jp-icon-override and let developer switch the style where they see fit. So adding your above snippet in the documentation sounds good enough.
> I don't get how introducing jp-icon-override will simplify cover those cases.

Yeah, after thinking about this more, it does seem unnecessary as all it does is really just save a few lines and requires the developer to know that the class exists before using it. I think to keep our default CSS classes minimal, we should strike off this idea.

I can briefly explain the approach I was thinking of. The idea is to add the `.jp-icon-override` class to the parent element of the icon, and then set `color: whatever` on styles attached to the parent element. Thus we are delegating the responsibility of controlling icon color to the parent rendering the icon rather than the icon itself. But again, we can just have the component owners write the CSS selectors themselves.

> You should not commit docs/source/developer/ui_components.rst as it is generated.

@fcollonval Are you sure about this? The `docs/source/developer` has other files underneath it as well, so I followed the convention by committing them to the repo. 🤔 

Screen Shot 2022-06-13 at 10 52 31 AM

dlqqq:
Reviewed all failing UI snapshots, and there are no remaining unintentional regressions.

ajbozarth:
I'll take a deep look at this as soon as I have time (prob not until next Monday though)

fcollonval:
> >    You should not commit docs/source/developer/ui_components.rst as it is generated.
> 
> @fcollonval Are you sure about this? The docs/source/developer has other autogenerated files underneath it as well, so I followed the convention by committing them to the repo. thinking

@dlqqq you are indeed closer to the solution - the script copying the file from ui-components package needs to be updated:

https://github.com/jupyterlab/jupyterlab/blob/989683cb7f83a9fbbe6caea1e1b0cfb9096f8d16/packages/ui-components/docs/build.sh#L10

The path is not `source/developer` but `source/extension`. So the existing file should be modified and not brand new as for now.

HaudinFlorence:
Hi, I am wondering if the logics of monochromatic icons could be extended to cases when the background colors are colored. Like on this example snapshot:
Screenshot from 2022-10-07 16-55-50
This may be useful for creating new themes.

PR #12695, opened 2022-06-13
Remove modeldb

dmonad:
This PR implements a lightweight alternative to #12359. As we discussed last year, we wanted to remove modeldb as it maintains the same information as `shared-model`. This PR does just that.

As we can remove modeldb, we can make the shared-models package the only source of truth, which will greatly simplify double-binding logic and can help us to avoid the initialization problem (concurrent initializations by different peers can lead to users overwriting content). There should be no syncing back-and-forth between, for example, the notebook widgets and shared models. When somebody wants to add a new widget, they need to modify the shared notebook model, which will trigger the same event handlers as if there would be a remote insertion of a new cell. This simplifies the codebase + we can throw out duplicate implementations of the same functionality (overall this PR removes 1300 lines of code).

The most important contribution of this PR is that it completely avoids the initialization problem. It is important for us that there is only one way to initialize content (via shared-models). The shared-model package initializes a document safely with a single cell. That means users don't have to wait for the server to respond before editing. This PR removes all initialization code that we previously used, which was fairly problematic in shared editing scenarios.

Future PRs could move the shared models defined here again into separate packages (e.g. into the `cells` package as #12359 does). Although I recommend keeping the shared state separate from the application-specific state (e.g. models containing `dirty` flag, `mimeType`, selections, and another application-specific state like the editor instance that renders the model). This also enables others to reuse the shared-models package.

## Code changes

Throughout the codebase, there are many places where we follow a fairly complicated event flow. This PR simplifies the event flow and enforces that the shared model is the only source of truth for all editor content. This leads to a significant reduction in the number of lines as we can avoid a lot of the complicated "double-binding" logic (syncing one state with another).

This PR retains the double-binding logic for `metadata` (which is an `ObservableMap`). It gives users the ability to have fine-grained access to metadata that many depend on. Ideally, we move the `metadata` field into the `shared-model` package and remove the double-binding logic. This can be done in a future PR.

## User-facing changes

None

## Backwards-incompatible 

We remove the modeldb API, which might be relevant to extension authors.

----

@hbcarlos @davidbrochart @SylvainCorlay @echarles


davidbrochart:
Thanks @dmonad.
It looks like you need to rebase.

dmonad:
This PR is ready for review.

> I see a lot of calls like ...model.sharedModel... (eg editor.model.sharedModel.getSource()). At some point, we could think to add shortcut methods to get rid of that call and simply expose editor.model.getSource() if possible. The whole idea would be to have a model hosted in the shared-models package that can be access by the other packages.

@echarles Thanks for your feedback. I think ideally we simplify the codebase even further so that there are fewer (nested) models. I want to encourage users to use the shared-models package instead directly, whenever they want to manipulate the document.

dmonad:
Thank you for the feedback @SylvainCorlay.

> The hierarchy of cell types do not match the schemas. For example, the base YCell already has attributes such as attachments, which don't make sense for code cells (see [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/shared-models/src/ymodels.ts#L883)). The single CellChange object also includes keys for execution count and outputs, which don't make sense for markdown cells (see [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/shared-models/src/api.ts#L498)).

That is very valuable feedback. Indeed, the YCell shouldn't maintain attachments, this should be implemented only in `YMarkdownCell` and `YRawCell`. The same goes for `CellChange`. I'm not sure why we implemented it like this. I'll fix this.

> Another example of a misuse of shared models is that the CodeEditorModel makes use of a YCodeCell (see [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/codeeditor/src/editor.ts#L217)) for its shared document, which has many attributes that don't correspond to anything in the code editor, such as id, cell type, metadata, outputs, execution count, and attachments.

I think you misunderstood that. CodeEditor only depends on the `ISharedText` - which is a general abstraction for any kind of text content. You can supply a `YNotebook` as well, and it will work just fine. 

Background: A `CodeEditor` can be initialized from a cell model or from a file content. The other PR solves these two cases in an `if` clause:
https://github.com/jupyterlab/jupyterlab/blob/4112b8635537f6405a133e156e86c1c82dbb0c88/packages/codeeditor/src/editor.ts#L846-L878
https://github.com/jupyterlab/jupyterlab/blob/4112b8635537f6405a133e156e86c1c82dbb0c88/packages/codeeditor/src/editor.ts#L240-L253

I think the abstraction of `ISharedText` is ideal in this scenario, as it can be implemented by other things as well (e.g. a different type of cell that has multiple text objects). We probably don't want to handle all different cases as `if` conditions (jupyter notebook vs raw text content, vs other potential types of content).

For simplicity, we a basic structure (`Cell`) and cast it immediately to an `ISharedText`.

> The YNotebook and YFile models are both extending YDocument (see [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/shared-models/src/ymodels.ts#L226-L228) and [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/shared-models/src/ymodels.ts#L100-L102)), which has a public source attribute (see [here](https://github.com/jupyterlab/jupyterlab/blob/e3621f17ac367f58bb01f1b0d0d5f653ba1e5628/packages/shared-models/src/ymodels.ts#L91)), which does not exist at the top level of the notebook schema. Hence the YDoc (even in the backend) has a source top-level attribute, that we are simply skipping when saving to disk to respect the notebook format.

You are absolutely right about that. `source` shouldn't be a property of `YDocument`. This should be private to `YFile`. I have no idea how this landed there. The idea was exactly to get away from this type of representation. Thanks for the feedback.

---

The idea of the models in shared-models is exactly to give an exact representation of what different kinds of cells are. Unfortunately, there are still some things lacking. I'll fix it.

---

> Fixing these issues is definitely possible - but this illustrates my argument that maintaining parallel class hierarchies is a slippery slope in terms of design.

We already have multiple hierarchies of models.

 Here are the three hierarchies as I understand them:

* `shared-models` - This module is responsible for modeling shared data. For example, it defines how we sync a shared notebook `YNotebook` to the backend and to other clients. Others can reuse this package as a baseline to enable collaborative editing in their applications. This module defines, for example, `SharedMarkdownCell`.
* `cells/models` - This module is responsible for modeling application state (e.g. `trusted` fields, selections, ..) that users don't generally want to share with other clients. This package is not reusable by third-party applications. However, it would make sense if `cells` would inherit from the models defined in `shared-models`. This module defines, for example, `MarkdownCellModel` that uses `SharedMarkdownCell`.
* `cells/widgets` - This module is responsible for rendering the content of the `cells/model` to the dom. This package defines, for example, widgets like `MarkdownCell` that use `MarkdownCellModel`.

Previous to our work last year (and in the alternative PR) we still have a separation between application state and shared state. However, shared-state is represented as untyped Key-Value stores (aka modeldb). Hence the majority of logic for the shared content lives in `cells/model`. 

Multiple hierarchies always existed because it makes it easier to deal with specific problems. Modeling a schema for shared content is definitely one of the problems that deserve to live in its own package. It is really hard to get right. At least, it should be put in a separate module (file). But the separation between *application state* and *shared state* should still exist.

### Todos

- [ ] `CellChange` is too generic
- [ ] `YCell` has attachments
- [ ] Remove `source` field from `YDocument`

echarles:
> Multiple hierarchies always existed because it makes it easier to deal with specific problems

Would you name that "separation of concerns"?

echarles:
>  the application that consumes @jupyterlab/shared-models doesn't ship all of @jupyterlab/notebook with it

The build process will pull a lot of transitive deps, which is a minor issue.

> Why do you see it as not adding value, @echarles?

In my view, the whole point of shared-models is to have a single place to control how CRDT behaves. Reexporting is just sugar, but the expected core functionality of that single place will be lost because there is no more single place.

I you have discussed your proposal on the RTC call and you all agree with your solution, please process.

afshin:
@dmonad I am happy to rebase your PR and implement https://github.com/jupyterlab/jupyterlab/issues/12708 if you would prefer. Is that alright with you?

@echarles this issue is one of the two that we put to a vote because we were stuck. It's a compromise that I hope we can implement and iterate on. (cf. https://github.com/jupyterlab/team-compass/issues/149)

packages/shared-models/src/ymodels.ts
```diff
@@ -190,6 +195,24 @@ export class YFile
...
+ * Note that initialization must not be done dynamically. Hence you need to create a static update message that doesn't change.
+ *
+ * You may creates templates by creating an empty shared notebook model, insert some content (e.g. an intial code cell), and then encoding the update to base64:
+ *
+ * ```js
+ * template = buffer.toBase64(Y.encodeStateAsUpdateV2(model.ydoc))
+ * ```
+ */
+const yCodeCellTemplate =
+  'AAAGyKjKnwUFAgEEAAcHACcAKAMnPDRjZWxsc3NvdXJjZW1ldGFkYXRhY2VsbF90eXBlaWRleGVjdXRpb25fY291bnRvdXRwdXRzBQYICQIPBwMBAAADAQIAAkECAQcAdgB3BGNvZGV3JGU1MTNlNzk2LTM4NmItNDg3Zi1hNGFjLTQ5MjZkNTcyZGJiNX0AAA==';
```
echarles, line 210: Can we document the decoded version of those base64 templates?

echarles:
Just pulled your latest commit @dmonad and RTC functionality works find on my env. In the meantime, the changes for codemirror6 have been merged, giving conflicts to resolve. Looks like the impacts is still limited, but that gives more work on this PR to fix that.

packages/shared-models/src/ymodels.ts
dmonad: The above comment explains how to create these "templates". These are base64-encoded Yjs documents - it's not really possible to decode this in a human-readable comment. However, I extended the above comment and added code that shows how these templates are created.

dmonad:
I rebased and made a quick usability check (checking whether collaboration still works). Let's wait for the tests to succeed.

packages/shared-models/src/ymodels.ts
echarles: Well, it would still be useful to indicate what is the empty document the user will start with. I guess it is a json payload with a few metadata and a cell list have a single empty cell. Could that json payload be added in the typedoc?

dmonad: Makes sense. Will do.

echarles:
@dmonad just tried locally. The RTC functionality does not work (well at some point it synced a cell with a lot of latency). The cursor are not show and I have in the logs the following messages

```
[I 2022-07-22 14:40:31.537 LabApp] Generating new user for token-authenticated request: 6358a42b0ef64887bb6f401c626d74fd
[I 2022-07-22 14:40:31.816 ServerApp] Generating new user for token-authenticated request: fe4a4f5bb79e4b30b7a372155eb31a14
[I 2022-07-22 14:40:31.821 LabApp] Generating new user for token-authenticated request: 6e7196d02f0d412a8a9b93a71e32ae97
[I 2022-07-22 14:40:31.834 ServerApp] Generating new user for token-authenticated request: 9524dc53aa374884838f8b0117c5a371
[I 2022-07-22 14:40:31.908 LabApp] Generating new user for token-authenticated request: ef54a5168c364c32ba560b8dc098f236
[I 2022-07-22 14:40:32.945 LabApp] Generating new user for token-authenticated request: 66c73fca13a74b628efd08abf7064eb1
[I 2022-07-22 14:40:39.827 TerminalsExtensionApp] Generating new user for token-authenticated request: 3ef086d6f10b4eb78e9cd64840ce0f51
[I 2022-07-22 14:40:39.828 ServerApp] Generating new user for token-authenticated request: 65188a38125344ac9674deab5ba368c9
[I 2022-07-22 14:40:39.830 ServerApp] Generating new user for token-authenticated request: ce82fe116a594d4883fd0e350b3559ab
[I 2022-07-22 14:40:39.831 ServerApp] Generating new user for token-authenticated request: f0ca6a3352e64c07801eedd288baa2f5
[I 2022-07-22 14:40:39.832 LabApp] Generating new user for token-authenticated request: c93caef1e73541ce8567de0833de0e29
```

dmonad:
Thanks for checking @echarles. I just tried that out again and it does work for me. I collaborated in two different browsers and I didn't notice any error messages.

Can you please make sure that you made a fresh installation? You might need to remove `node_modules` and reinstall.

On another note: The `js-services` test works for me locally. I think it just failed because of some timeout issue in the GitHub actions service.

echarles:
I have relaunched the js-services test and it is now green.

I was on a wrong env. Started again from source with tag `v4.0.0a26` with valid RTC.

Same env, with this branch and your latest commits, when I got to the second with a second browser, I end-up n the console with the following error.

```
  File "/Users/echarles/opt/miniconda3/envs/datalayer/lib/python3.10/asyncio/tasks.py", line 234, in __step
    result = coro.throw(exc)
  File "/Users/echarles/opt/miniconda3/envs/datalayer/lib/python3.10/site-packages/tornado/websocket.py", line 1104, in wrapper
    raise WebSocketClosedError()
tornado.websocket.WebSocketClosedError
```

dmonad:
We didn't specify a `ypy-websocket` version. When I did a fresh installation, I got very different issues (I was running 0.1.9, even though the latest version is 0.2.0). I fixed `ypy-websocket` to `0.2.0` in `setup.cfg` which fixed my issues. Can you please try again @echarles?

@davidbrochart Which version are we supposed to use?

echarles:
Just tried again. The RTC changes are not shown in the second browser. The second browser hangs when running cell. 
When I stop the server, it exits with `RuntimeError: Cannot enter into task <Task pending name='Task-14' coro=<HTTP1ServerConnection._server_request_loop()`.

```
y-py                     0.5.3
ypy-websocket            0.2.0
```

BTW the `YDocWebSocketHandler` has just moved to jupyter-server with today merge of https://github.com/jupyterlab/jupyterlab/pull/12772/files

davidbrochart:
> @davidbrochart Which version are we supposed to use?

When #12852 gets in, you will get `jupyter_server_ydoc>=0.1.7`, which in turn requires [ypy-websocket>=0.2.0](https://github.com/jupyter-server/jupyter_server_ydoc/blob/fd7c0927a62c0b9e2aaf28d5f48a226a9321287d/pyproject.toml#L25).

davidbrochart:
> Just tried again. The RTC changes are not shown in the second browser. The second browser hangs when running cell.

I also had a weird behavior when trying RTC with this branch. It might also be caused by the switch to CodeMirror 6.

> When I stop the server, it exits with `RuntimeError: Cannot enter into task <Task pending name='Task-14' coro=<HTTP1ServerConnection._server_request_loop()`.

I think this is unrelated (caused by Tornado's latest release).

echarles:
> I also had a weird behavior when trying RTC with this branch. It might also be caused by the switch to CodeMirror 6.

That is very much possible. This branch has to deal with a lot of moving parts (server, cm6...)

> I think this is unrelated (caused by Tornado's latest release).

Prolly unrelated indeed. I just had that message once and thought to report it.

dmonad:
@davidbrochart There is something wrong in the master branch of jupyterlab. I get the following error when doing a fresh installation of jupyterlab.

image

dmonad:
@davidbrochart I can't get the `master` to run anymore.

After a fresh installation, I get the following error.

image

I still don't understand why we don't specify a version for `ypy-websocket` in `jupyterlab/setup.cfg`:

```
ypy-websocket
jupyter_ydoc==0.1.10
```

Implicit versions seem dangerous. Why don't we use semver?

davidbrochart:
> ```
> ypy-websocket
> jupyter_ydoc==0.1.10
> ```

Are you on JupyterLab HEAD? There is no such thing in [setup.cfg](https://github.com/jupyterlab/jupyterlab/blob/a177a6dca561d4aac4ecc6e25146e4ff895c7702/setup.cfg).

dmonad:
Sorry about that @davidbrochart. Indeed, I was still on an older commit.

@echarles This works now for me as well after a fresh install. The tests are also running..

jupyterlab/staging/package.json
```diff
@@ -126,7 +126,7 @@
     "@lumino/widgets": "^1.32.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "yjs": "^13.5.34"
+    "yjs": "^13.5.40"
```
jtpio, line 129: For this PR it looks like this change to `jupyterlab/staging/package.json` can be omitted.

Staging is updated automatically during a release by this script:

https://github.com/jupyterlab/jupyterlab/blob/3291708449f7fe008daf0ba1be63c65f3dbc2bb1/buildutils/src/update-core-mode.ts#L34

Undoing this change should also make the `staging-check` CI check pass.

echarles:
I have tried on my local env and the cells run fine, but the RTC feature is not shown (the text does not sync, the shared cursor is not shown). I end-up with a new `.jupyter_ystore.db` file which is expected. I think.

I have tried on binder with this link https://hub-binder.mybinder.ovh/user/dmonad-jupyterlab-s5a2dof0/lab-dev/lab. The link asks authentication for another browser, so I have tried with 2 windows of the same browser and get weird behavior (jiggling cursor, no text sync, see image)

Untitled5

jupyterlab/staging/package.json
dmonad: Thanks! I wasn't aware of that.

dmonad:
I reverted the change to `staging/package.json` as @jtpio suggested.

@echarles Thanks for bringing up this issue. However, I'm pretty sure that this is unrelated to the changes in this branch. This branch only removes modeldb with some refactors on how we manage cells and such. The bug you described shows that the clients only partially synced. This is likely to be a bug with the new ypy websocket server. 

Unless there are other issues, I suggest that we merge this PR.

davidbrochart:
> This is likely to be a bug with the new ypy websocket server.

Not sure, there is also the move to CodeMirror 6.

dmonad:
I offer these changes to the gods of Jupyter. 

I have no idea how to get the "Visual Regression **Documentation**" running. Some changes to the vdom are supposed to happen as this PR removes a lot of code.

I will not be working on this PR any longer.

SylvainCorlay:
The bug with the cursor seems to happen on master.

fcollonval:
> UI tests failure must be resolved prior to merge.

In the present state, the removal of ModelDB and associated changed signals is breaking callback as seen in the UI tests; e.g.

image

afshin:
Thanks for pushing hard on this @dmonad. We'll take a look at the CI and get it in as soon as we can. This is a universally awaited piece of functionality.

SylvainCorlay:
Johan (@JohanMabille) is looking into the cursor/codemirror 6 bug that is also affecting the main branch.

The callback issue that Frederic showed above seems to be legitimately related to this PR.

davidbrochart:
I tried locally with a prior version of ypy-websocket, and the cursor issue is still present, so I would say that it doesn't come from ypy-websocket.
I've seen HEAD working fine (including RTC sync) and then cursors start to jump around, but most of the time the cursor bug happens immediately and the browsers freeze.

davidbrochart:
I reverted Johan's PR on CodeMirror 6, and cherry-picked my commits that came after, and it works fine (in master).
I think that means the bug comes from CodeMirror 6.

fcollonval:
> I think that means the bug comes from CodeMirror 6.

:sob:

SylvainCorlay:
It works with a single cell, and starts failing when there is more than one. This may be useful to track down where the issue is.

JohanMabille:
https://github.com/jupyterlab/jupyterlab/pull/12877 should fix it.

packages/codemirror/package.json
```diff
@@ -80,9 +79,9 @@
     "@lumino/widgets": "^1.33.0",
     "react": "^17.0.1",
     "style-mod": "~4.0.0",
-    "y-codemirror.next": "~0.3.0",
```
jtpio, line 83: Curious if there is a specific reason for switching from `~` to `^` for that particular package?

Or maybe because it should have been `^` from the beginning?

jtpio: Probably fine either way, was mostly to know the extra context if any.

dmonad: In the JavaScript ecosystem you usually use ^ to get all bugfixes. 

^13.0.0 will use the latest release (13.x.x).

~13.0.0 only accepts patches (13.0.x). So it will never use 13.2.7 (which is a later release with a lot of bugfixes).

In this specific case (0.3.0) it doesn't make a difference.

jtpio: OK then for that particular dependency it makes sense.

The JupyterLab code base sometimes pins on patches for some dependencies like webpack loaders or `typescript`, or for those known to break on minors.

Marking as resolved.

jtpio:
Looks like there are some conflicts, preventing the CI to be triggered.

fcollonval:
I'm still working on this one. There is at least one trouble that remains. The execution_count for new cell is set to zero what does not respect nbformat. That is the root cause of UI test failures because it breaks the dirty state (turning it on even if the cell was never run). I'll rebase once locally it looks good.

jtpio:
OK sounds good, thanks @fcollonval for the update :+1:

fcollonval:

fcollonval:
Kicking CI with the latest snapshots

fcollonval (approved):

SylvainCorlay:
Many thanks to @fcollonval for completing this!

fcollonval:
Important note the memory leak job highlight that a notebook is not cleared from memory when closed. This should be fixed before merging.

fcollonval (commented):
bot please run benchmarks

fcollonval:

PR #12349, opened 2022-04-05
Provide `IDocumentManager` in a separate plugin

jtpio:
## References

This should help with https://github.com/jupyter/notebook/issues/6344.

## Code changes

Move the instantiation of the `DocumentManager` and the way the `IDocumentManager` token is provided to a separate plugin.

This should make it easier for downstream applications to provide a different `IDocumentManager`, while still reusing the default commands and settings.

## User-facing changes

None

## Backwards-incompatible changes

- Add `lastModifiedCheckMargin` to the `IDocumentManager` interface for consistency
- Some plugin restructuring.

PR #12701, opened 2022-06-15
Split the Document Manager extension into multiple plugins

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/12349

Fixes https://github.com/jupyterlab/jupyterlab/issues/12965

This will help with https://github.com/jupyter/notebook/pull/6446 to avoid duplicating too much logic in the Notebook v7 code base.

## Code changes

Split some of the docmanager extension logic into multiple plugins so they are easier to reuse and compose downstream.

New plugins:

- `@jupyterlab/docmanager-extension:opener`: a plugin to open document widgets
- `@jupyterlab/docmanager-extension:contexts`: handle dirty contexts

### TODO

- [x] Split plugins
- [x] Add docstrings
- [x] Fix tests
- [x] Fix examples
- [x] Add to extension migration guide

## User-facing changes

This fixes an issue introduced in https://github.com/jupyterlab/jupyterlab/pull/12349 not being able to create a new view for a document widget:

image

This was because the name `opener` was used in the code, but was not referring to the widget opener but implicitly to `window.opener` instead: https://developer.mozilla.org/en-US/docs/Web/API/Window/opener

~None at the API level~

- The `IDocumentWidgetOpener` interface also now defines an `opened` signal that is emitted when a widget is opened.
- Some plugins have their dependencies changed.


jtpio:
cc @fcollonval for review, since you also had a look at this in https://github.com/jupyterlab/jupyterlab/pull/12966

packages/docmanager-extension/src/index.tsx
```diff
@@ -93,12 +94,77 @@ namespace CommandIDs {
...
+        this.opened.emit(widget);
+      },
+      get opened() {
+        if (this._signal) {
+          return this._signal;
+        }
+        this._signal = new Signal<IDocumentWidgetOpener, IDocumentWidget>(this);
+        return this._signal;
+      }
+    };
```
fcollonval, line 129: Let's freeze the object to forbid tempering:

```suggestion
    return Object.freeze({
      open(widget: IDocumentWidget, options?: DocumentRegistry.IOpenOptions) {
        if (!widget.id) {
          widget.id = `document-manager-${++Private.id}`;
        }
        widget.title.dataset = {
          type: 'document-title',
          ...widget.title.dataset
        };
        if (!widget.isAttached) {
          shell.add(widget, 'main', options || {});
        }
        shell.activateById(widget.id);
        this.opened.emit(widget);
      },
      get opened() {
        if (this._signal) {
          return this._signal;
        }
        this._signal = new Signal<IDocumentWidgetOpener, IDocumentWidget>(this);
        return this._signal;
      }
    });
```

fcollonval (changes requested):
Thanks @jtpio

I have a minor suggestion. Otherwise this is good to go.

fcollonval:
@jtpio sorry I miss the instantiation of `this._signal`. So instead of `Object.freeze` an anonymous class would be better.

jtpio:
ah right, at least CI caught it.

I'll push a fix.

packages/docmanager-extension/src/index.tsx
```diff
@@ -117,16 +117,17 @@ const openerPlugin: JupyterFrontEndPlugin<IDocumentWidgetOpener> = {
...
-          return this._signal;
-        }
-        this._signal = new Signal<IDocumentWidgetOpener, IDocumentWidget>(this);
-        return this._signal;
+        return this._opened;
       }
-    });
+
+      private _opened = new Signal<IDocumentWidgetOpener, IDocumentWidget>(
+        opener
```
fcollonval, line 128: It should be this, isn't it?

```suggestion
        this
```

jtpio: yes fixed in https://github.com/jupyterlab/jupyterlab/pull/12701/commits/098660f8e321662c88994436e59849f161f750ce

fcollonval (approved):
Thanks

Issue #12647, opened 2022-06-02
Side-by-Side overlaps with cell tools


mlucool:
## Description

The cell tools are always placed on the right most part of the cell. This does not work for side by side mode where the input and output share the same line.

## Reproduce

Turn on side-by-side mode via the command pallete. Then output something like `'abcdef' * 15`:

![image](https://user-images.githubusercontent.com/1813603/171696110-ca7cf649-9e47-4c40-924c-85978d2e03a3.png)

## Expected behavior

The cell tools overlay the input only.

## Context
Lab@3.4.1

CC @echarles

jweill-aws:
Could also be put in visual regression testing

peytondmurray:
I'll take this issue on; I think it should be possible to check whether any output text overlaps with the toolbar. We'll hide the toolbar if there's overlap, and display it if not. For any other kind of output (images, etc) we can hide the toolbar.

peytondmurray:
This issue has been addressed in https://github.com/jupyterlab/jupyterlab/pull/12710 and can be closed.

PR #12710, opened 2022-06-17
Fix cell toolbar overlap in side-by-side render mode

peytondmurray:
## References
https://github.com/jupyterlab/jupyterlab/issues/12647

## Code changes
* When the notebook's `renderingLayout === 'side-by-side'`, a new check has been introduced which ensures that the output(s) of each cell do not overlap with the cell toolbar. If they do, the toolbar is hidden (as is currently done for `renderingLayout === 'default'`. If not, the toolbar will be shown.
* Added a new `renderingLayoutChanged` signal to the `StaticNotebook`, and made the cell toolbar re-render upon receiving this signal. This handles an edge case where if the user is in `default` render layout mode the toolbar can be visible; but if they switch to `side-by-side` render layout, it may overlap with cell output. This check ensures this doesn't happen.

## User-facing changes

* This PR changes how the cell toolbar is shown or is hidden according to the current rendering layout.

### Before
before

### After
after

None

### DIFF
--- a/packages/cell-toolbar/src/celltoolbartracker.ts
+++ b/packages/cell-toolbar/src/celltoolbartracker.ts
@@ -3,7 +3,7 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 import { createDefaultFactory, ToolbarRegistry } from '@jupyterlab/apputils';
-import { Cell, ICellModel, MarkdownCell } from '@jupyterlab/cells';
+import { Cell, CodeCell, ICellModel, MarkdownCell } from '@jupyterlab/cells';
 import { DocumentRegistry } from '@jupyterlab/docregistry';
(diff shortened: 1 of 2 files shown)


jtpio:
Nice thanks @peytondmurray for working on this :+1: 

cc @jweill-aws who might want to have a look

jweill-aws (approved):
Changes look good — thanks! Sorry for the delay in getting back to you.

jtpio:
Thanks!

The remaining CI failure has been fixed in https://github.com/jupyterlab/jupyterlab/pull/12714.

jtpio:

jtpio:
@peytondmurray or @jweill-aws it looks like the automated backport to 3.4 didn't succeed.

Mind opening a PR against the `3.4.x` branch to port these changes? Thanks!

PR #12714, opened 2022-06-17
stub extension search in UI test

### LINKED ISSUE #12340: More robust UI tests

dlqqq:
## References

Fixes #12340. Lays groundwork for further stubbing (if necessary) in the extension manager UI tests.

## Code changes

Adds stub for extension search. To stub the GH avatars in the search results, I decided to use the Jupyter logo since the avatars in the original snapshot were those of QuantStack and another GitHub user (deathbeds@). Wanted to avoid copyright/trademark/legal/privacy issues.

Open to feedback regarding the stub data I've used. Now is the time to change it if we want to.

## User-facing changes

None.

None known.


galata/test/documentation/extension_manager.test.ts
```diff
@@ -81,6 +86,8 @@ test.describe('Extension Manager', () => {
     expect(
       await page.screenshot({ clip: { y: 31, x: 0, width: 283, height: 600 } })
     ).toMatchSnapshot('extensions_search.png');
+
+    await unstubExtensionsSearch(page);
```
dlqqq, line 90: Yeah, it turns out `page.route` doesn't remove the route stub after a test completes, so it's required to call `page.unroute` manually after each test. If anybody knows an alternative to this (that doesn't make it so easy for devs to accidentally share stubs across tests), please LMK!

fcollonval: @dlqqq how did you come to that conclusion? The `.route` is linked to the page object that is unique for each test (except when using some specific describe block). So I'm a bit surprised it needs to be undone explicitly.

fcollonval:
Thanks a lot for working on this @dlqqq

galata/test/documentation/extension_manager.test.ts
dlqqq: I noticed this when adding `page.route` caused the other snapshot tests to fail. I'm also surprised, since I also assumed that the page object is unique per test. This probably warrants further investigation since ideally page routings shouldn't get shared across tests.

galata/test/documentation/data/extensions-search.json
```diff
@@ -0,0 +1,190 @@
...
+          "homepage": "https://github.com/QuantStack/jupyterlab-drawio",
+          "repository": "https://github.com/wolfv/jupyterlab-drawio",
+          "bugs": "https://github.com/QuantStack/jupyterlab-drawio/issues"
+        },
+        "author": {
+          "name": "Wolf Vollprecht"
+        },
+        "publisher": {
+          "username": "hbcarlos",
+          "email": "carlosherrerocontact@gmail.com"
```
jtpio, line 27: Although it's true they are public on npm, maybe we could use placeholder email addresses here to avoid making them more visible and searchable via the GitHub UI?

dlqqq: Good thinking. Done. 👍

jtpio (approved):
Thanks!

Issue #12711, opened 2022-06-17
Support user-defined `y_doc` document type 


trungleduc:
### Problem

The current y_doc handler of `JupyterLab` uses the `file_type` of the document to create the corresponding y_doc document, but [JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/b4540166b6afaba3454492591122ca963dbb3350/packages/services/src/contents/index.ts#L121) and [jupyter_server](https://github.com/jupyter-server/jupyter_server/blob/7b5815e0176c079d17c539c3baf6b497197f58a7/jupyter_server/services/contents/handlers.py#L101) hard-coded the file types to `file | notebook | directory` so any new file type registered with  `jupyter_ydoc` can not be used.

### Proposed Solution

Add an optional parameter to the `path` property of `IDocumentProviderFactory.IOptions` to specify the type of  `y_doc`
 document.

trungleduc:
cc @hbcarlos @davidbrochart

davidbrochart:
Thanks @trungleduc, yes that is needed to support collaborative editing for new document types.

trungleduc:
Fixed by https://github.com/jupyter-server/jupyter_server/pull/895.

PR #12719, opened 2022-06-20
Add `yDocumentType` to `IModelFactory`

trungleduc:
## References

Closes https://github.com/jupyterlab/jupyterlab/issues/12711

## Code changes

Add `yDocumentType` property to `DocumentRegistry.IModelFactory` to define the type of the Y document. This is an optional property and `WebSocketProvider` will fall back to the content type if `yDocumentType` is not defined.

## User-facing changes

N/A

N/A

### DIFF
--- a/jupyterlab/handlers/ydoc_handler.py
+++ b/jupyterlab/handlers/ydoc_handler.py
@@ -29,9 +29,9 @@ class JupyterRoom(YRoom):
 
 class JupyterWebsocketServer(WebsocketServer):
     def get_room(self, path: str) -> JupyterRoom:
-        file_format, file_type, file_path = path.split(":", 2)
+        file_format, file_type, file_path, ydoc_type = path.split(":", 3)
         if path not in self.rooms.keys():
-            self.rooms[path] = JupyterRoom(file_type)
+            self.rooms[path] = JupyterRoom(ydoc_type)
         return self.rooms[path]
 
 
@@ -58,7 +58,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
 
     def get_file_info(self) -> Tuple[str]:
         room_name = self.websocket_server.get_room_name(self.room)
-        file_format, file_type, file_path = room_name.split(":", 2)
+        file_format, file_type, file_path, ydoc_type = room_name.split(":", 3)
         return file_format, file_type, file_path
 
     def set_file_info(self, value: str) -> None:
--- a/packages/docprovider/src/tokens.ts
+++ b/packages/docprovider/src/tokens.ts
@@ -47,8 +47,22 @@ export namespace IDocumentProviderFactory {
      * The name (id) of the room
      */
     path: string;
+
+    /**
+     * The content type of the file.
+     */
     contentType: string;
+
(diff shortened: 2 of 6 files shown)


jupyterlab/handlers/ydoc_handler.py
```diff
@@ -29,9 +29,9 @@ def __init__(self, type):
 
 class JupyterWebsocketServer(WebsocketServer):
     def get_room(self, path: str) -> JupyterRoom:
-        file_format, file_type, file_path = path.split(":", 2)
+        file_format, file_type, file_path, ydoc_type = path.split(":", 3)
         if path not in self.rooms.keys():
-            self.rooms[path] = JupyterRoom(file_type)
+            self.rooms[path] = JupyterRoom(ydoc_type)
```
davidbrochart, line 34: I don't quite understand what other information `ydoc_type` provides, compared to `file_type`?
I thought that what we wanted was for the front-end to send a configurable `file_type`, that is currently hard-coded to either `file` or `notebook`.
Am I missing something?

trungleduc: Currently, `jupyter-server` only serves *directory*, *file*, and  *notebook* `file_type` (https://github.com/jupyter-server/jupyter_server/blob/7b5815e0176c079d17c539c3baf6b497197f58a7/jupyter_server/services/contents/handlers.py#L101), if we accept any `file_type`, we will need to update the content handler and managers of `jupyter-server`. I don't know enough about `jupyter-server` to judge the relevance of these changes so I went this way for the `ydoc` type.

davidbrochart: But you might be introducing a new parameter which is redundant with an already existing one.
Do you see `ydoc_type` taking a different value than `file_type`, if we could configure `file_type` to not be restricted to *directory*, *file* or *notebook*?
Maybe a document could be a *notebook* (for its `file_type`) but it's Y type not necessarily a *YNotebook* (for its `ydoc_type`), for instance. That would be a valid situation.

hbcarlos: > Yes, we can open the same file extension in different views by adding widget factories for this file extension, and in each widget factory, we can define the document model with DocumentRegistry.IModelFactory interface. By adding yDocumentType to this interface we can have different Y-types for the same file extension.

Yes, but this is what the Model does. At the moment we define one model (schema) to load the data, and then we can use this model in multiple widgets. I believe the WidgetFactory doesn't specify a `file_type` but the model's name.
I'm with David here, having a new attribute `ydoc_type` is redundant. I think we should make the actual attribute `file_type` more flexible and use it to specify the schema of the document.

trungleduc: Thanks @hbcarlos for the clarification, I agree that a fix on `jupyter_server` would be a better approach. I had a direct discussion with David that leads to https://github.com/jupyter-server/jupyter_server/issues/884.

trungleduc:
This PR is not needed anymore with the fix (https://github.com/jupyter-server/jupyter_server/pull/895) on the server side.

PR #12732, opened 2022-06-26
Translate "Default: " and "Remove" in custom fields

### LINKED ISSUE #12730: "Default: " in settings is not translated
## Description

The new "Default: " hint added in https://github.com/jupyterlab/jupyterlab/pull/12470 is missing a translation wrapper:

https://github.com/jupyterlab/jupyterlab/blob/c2907074e58725942946a73a823fc60e1795da39/packages/settingeditor/src/SettingsFormEditor.tsx#L345

## Reproduce

1. Install language pack, change language
2. Go to Setting editor UI
3. Modify a value
4. See that "Default: " hint is not translated

## Expected behavior

`CutomTemplate` should be wrapped in a factory function allowing to pass the translator (as is the case with `ObjectFieldTemplate` and `ArrayFieldTemplate`).

## Context

- JupyterLab version: 3.4 and 4.0 alpha

krassowski:
## References

Fixes #12730

## Code changes

- Custom field templates were wrapped in a factory.
- I am intending to build upon this change to subsequently remove hard-coded `id != 'jp-SettingsEditor-@jupyterlab/shortcuts-extension:shortcuts_shortcuts';` comparison.
- Also changed factory display names to make downstream debugging easier.

## User-facing changes

"Remove" and "Default: %1" will get translated.

None.


fcollonval:

PR #12738, opened 2022-06-27
Remove unstubExtensionsSearch

fcollonval:
## References

Follow-up #12714

## Code changes

Don't unstub routing


fcollonval:
CI failure unrelated

@dlqqq I tried locally and did not encounter any conflict for per test routing definition. The CI is also green - so let's keep thing simple.

jtpio (approved):
Thanks @fcollonval!

Indeed CI is green after restarting the failing check, so we should be able to go with this change.

PR #12740, opened 2022-06-27
Add resize handle for scrolled cell outputs

### LINKED ISSUE #12100: Resize scrolled output
### Problem

Scrolled output is very useful at times, but it's sometimes too small. It would be helpful if users could resize it on demand.

image

### Proposed Solution

Add CSS to allow for this widget to be resized ([ref](https://developer.mozilla.org/en-US/docs/Web/CSS/resize)).

peytondmurray:
## References

Closes #12100.

## Code changes

This PR adds (vertical only) resizing of cell outputs when the "Scrolled Output" option is enabled in a notebook. Only a small CSS change was needed to add this.

## User-facing changes

A resize handle was added to cell outputs when the "Scrolled Output" option is enabled (see the bottom right of the cell output for the new handle).

### Before
default

### After
resizeable

None - this is a pure CSS change.

### DIFF
--- a/packages/cells/style/widget.css
+++ b/packages/cells/style/widget.css
@@ -73,6 +73,11 @@
   overflow-y: auto;
   max-height: 24em;
   margin-left: var(--jp-private-cell-scrolling-output-offset);
+  resize: vertical;
+}
+
+.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea[style*='height'] {
+  max-height: unset;
 }
 
 .jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea::after {


packages/cells/style/widget.css
```diff
@@ -71,8 +71,9 @@
 
 .jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea {
   overflow-y: auto;
-  max-height: 24em;
+  height: 24em;
```
krassowski, line 74:
```suggestion
  max-height: 24em;
```

When users enable scrolling for outputs for all cells in the notebook they would end up with some outputs taking up more space than before.

There is a neat CSS trick allowing to retain `max-height` here, which relies on the resize handler adding `style="height: {userSetValue}px"` once user starts resizing, so we can target this state with a separate CSS selector:

```css
.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea[style*=height] {
    max-height: unset;
}
```

This should be safe to rely on as this behaviour is standardised in the spec:

> When an element is resized by the user, the user agent sets the [width](https://drafts.csswg.org/css-sizing-3/#propdef-width) and [height](https://drafts.csswg.org/css-sizing-3/#propdef-height) properties to px unit length values of the size indicated by the user, in the element’s [style attribute](https://www.w3.org/TR/css-style-attr/#style-attribute) DOM, replacing existing property declaration(s), if any, without !important, if any.

(https://drafts.csswg.org/css-ui/#resize)

packages/cells/style/widget.css
```diff
@@ -71,8 +71,9 @@
...
   margin-left: var(--jp-private-cell-scrolling-output-offset);
+  resize: vertical;
```
krassowski, line 76:
```suggestion
  resize: vertical;
}

.jp-CodeCell.jp-mod-outputsScrolled .jp-Cell-outputArea[style*=height] {
  max-height: unset;
```

(might require a lint pass)

packages/cells/style/widget.css
peytondmurray: Oh, that's really nifty - thanks for the advice!

fcollonval (approved):
Thanks @peytondmurray

PR #12428, opened 2022-04-18
Backport PR #11074 on branch 3.4.x (Use transform to quickly switch between tabs.)

fcollonval:
Backport PR #11074 on branch 3.4.x 

And upgrade lumino package dependencies

PR #12742, opened 2022-06-28
Fix staging/yarn.lock registry

vidartf:
## References

When [switching to use precommit](#12279) on the master branch, all the `resolved` entries in the yarn.lock file changed from using yarnpkg to using npmjs. This later spread (partially) to the 3.4.x branch in a backport https://github.com/jupyterlab/jupyterlab/pull/12428 (first released in 3.4.0a0), where the registry used for lumino packages changed. This causes an issue for those using custom package mirrors, since JupyterLab relies on the hard-coded value of `yarnpkg` to modify the lockfile:

https://github.com/jupyterlab/jupyterlab/blob/6eb9ae6558dc234d427ae8de50908bba3f6887e0/jupyterlab/commands.py#L1338-L1345

## Code changes

This adds an integrity check that the `resolved` values all go to yarnpkg mirror. This is a minimum solution for this issue, but other additional points could also be addressed:
1. We could ensure that the integrity script reads out the default from the `commands.py` file.
2. We could change the default to npmjs.
3. We could make the logic in commands.py handle both yarnpkg *and* npmjs as "default" sources.

These are mutually compatible with each other.

## User-facing changes

No visual changes.

None. If anything it fixes an backwards incompatibility issue.

### DIFF
--- a/buildutils/src/ensure-repo.ts
+++ b/buildutils/src/ensure-repo.ts
@@ -801,7 +801,7 @@ export async function ensureIntegrity(): Promise<boolean> {
     .getCorePaths()
     .filter(pth => !tsConfigDocExclude.some(pkg => pth.includes(pkg)))
     .map(pth => {
-      return { path: './' + path.relative('.', pth).replace('\\/g', '/') };
+      return { path: './' + path.relative('.', pth).replace(/\\/g, '/') };
     });
   utils.writeJSONFile(tsConfigdocPath, tsConfigdocData);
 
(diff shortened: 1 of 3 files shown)


jtpio:
Thanks @vidartf for catching this and the fix!

The `staging-check` CI failure is expected and in that particular case it makes sense to explicitly update the files in the `staging` directory manually.

fcollonval (approved):
Thanks @vidartf 

As [Steve's comment](https://github.com/jupyterlab/jupyterlab/pull/12279#issuecomment-1168819905) allow this one, I'm gonna merged. So we can release a new version as soon as possible.

fcollonval:
> Documentation CI failure is not related.

fcollonval:

PR #12746, opened 2022-06-28
Optimize debugger editor `eachLine` loops

vidartf:
When modifying each line of a large document it is very important that we do not update the DOM after each update. In CodeMirror this is done by wrapping loops in operations.

## References

None.

## Code changes

For `clearHighlight`, we wrap the loop in an operation.

For `clearGutter` we just call the already defined method in CM, which uses an operation by default. It might also be slightly more optimized in other ways.

## User-facing changes

No visual changes expected. Improvement to speed when navigating away from large files in the debugger editor.

## Backwards-incompatible changes

None.

Issue #12012, opened 2022-02-04
Debugger is very slow to start (21 seconds)


krassowski:
## Description

On a powerful desktop, enabling the debugger freezes the UI for way too long. This is a regression in 3.3 alpha.

## Reproduce

1. Install 3.3 in an environment with quite some number of packages installed (`pip list | wc -l` 337)
2. Open JupyterLab
3. Open a notebook with 195 code cells
4. Click on the bug icon
4. See freeze of about 20-30 seconds

## Expected behavior

The debugger starts within reasonable time without freezing UI, 2-3 seconds max.

## Context

- Operating System and version: Ubuntu
- Browser and version: Chrome 97
- JupyterLab version: 3.3.0a3

## Analysis

There are two issues here:

1. [ ] `set kernelSources` adds a huge number of widget nodes (~2 seconds); ideally we would make it 
2. [ ] [`_onCellsChanged`](https://github.com/jupyterlab/jupyterlab/blob/4dc0ec21fef81c18a912ea358800f5fd7e568dfa/packages/debugger/src/handlers/notebook.ts#L41) refreshes every single editor forcing hundreds of reflows via CodeMirror's `setOption` (~19 seconds)

Subsequently style recalculation takes additional 2-3 seconds due to big changes in the DOM.

![Untitled](https://user-images.githubusercontent.com/5832902/152522351-743aef56-0353-47da-b40b-de8793a7960f.png)

<details>

Zoom-in on `kernelSources`:

![Screenshot from 2022-02-04 11-44-40](https://user-images.githubusercontent.com/5832902/152523732-97514efe-6b6a-4c73-823e-66e7fe1523a7.png)

Zoom-in on `_onCellsChanged` in details:

![Screenshot from 2022-02-04 11-45-59](https://user-images.githubusercontent.com/5832902/152523867-7e4b86da-d46f-4d2c-bd47-9f14e1327616.png)

</details>

echarles:
> There are two issues here:
> - set kernelSources adds a huge number of widget nodes (~2 seconds); ideally we would make it
> -  [_onCellsChanged](https://github.com/jupyterlab/jupyterlab/blob/4dc0ec21fef81c18a912ea358800f5fd7e568dfa/packages/debugger/src/handlers/notebook.ts#L41) refreshes every single editor forcing hundreds of reflows via CodeMirror's setOption (~19 seconds)

[...]

PR #12751, opened 2022-06-29
Debugger: Make kernel source list react based

vidartf:
This improves performance when updating of the kernel sources, which happens whenever a document that is being debugged gets focused. This includes when switching from the notebook document to a kernel source editor. Ideally the kernel sources would not be refreshed when switching between tabs that share the same debug context.

Possible future improvements:
- Disconnect the source entry click event from further processing once we start the network request. This would avoid hanging the UI while waiting for the request to complete.
- Avoid refreshing sources when switching between tabs that share the same context (avoid a large "catch-all" update method if we know what has changed).
- Use a virtualized list/view of the kernel source list. 

## References

Follow up to #12746. Further progress on #12012.

## Code changes

Changes the debugger's kernel sources list to be react based. This improves performance because:
- We avoid large, iterative changes to the DOM that triggered many recalucations etc.
- The entries are keys on the name (+ dedupe id, as some name/paths can be duplicated!) so react can efficiently diff entries.
- The DOM manipulation happens after the signal processing has completed, meaning that we are less likely to "hang" on processing the original click event. I.e. the UI should still be responsive.

Also made the kernel source filter search bar be "sticky" so that it is always visible even if you scroll.

## User-facing changes
More responsive UI. Changed user interaction to open a kernel source from double click to single click. Filter search entry is now sticky:

sticky

None. AFAIU, the signatures changed are not public.


fcollonval (approved):
Thanks a lot @vidartf 

This looks great (tested on Binder).

When backporting it, we should keep the double click to not break current behaviour. Do you know if we have a good place to mention that the action switches from double click to single click in the documentation?

> CI failure are not related

fcollonval:

PR #12753, opened 2022-06-29
Backport PR #12742 on branch 3.4.x (Fix staging/yarn.lock registry)

fcollonval:
Backport PR #12742


tsconfigdoc.json
```diff
@@ -280,6 +286,12 @@
     {
       "path": "./packages/ui-components-extension"
     },
+    {
+      "path": "./packages/user"
+    },
+    {
+      "path": "./packages/user-extension"
+    },
```
vidartf, line 294: Do these bits belong here? They are not from the original PR.

fcollonval: Good catch - they comes from running the integrity check locally on my folder. I will remove them.

PR #12755, opened 2022-06-30
Rename Notebook from title: fix drive issue

martinRenou:
## References

cc. @jtpio 

Fix issue mentioned in https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/34#issuecomment-1156452964

Renaming from the title bar was not taking the current `drive` into account, failing with:
```
ContentsManager: renaming files must occur within a Drive
```

## User-facing changes

Users can now rename files in custom drives from the title bar

None

### DIFF
--- a/packages/docmanager/src/dialogs.ts
+++ b/packages/docmanager/src/dialogs.ts
@@ -43,10 +43,11 @@ export function renameDialog(
 ): Promise<Contents.IModel | null> {
   translator = translator || nullTranslator;
   const trans = translator.load('jupyterlab');
+  const localPath = manager.services.contents.localPath(oldPath);
 
   return showDialog({
     title: trans.__('Rename File'),
-    body: new RenameHandler(oldPath),
+    body: new RenameHandler(localPath),
     focusNodeSelector: 'input',
     buttons: [
       Dialog.cancelButton({ label: trans.__('Cancel') }),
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks for trying! Do you see from where this is happening? Is this when renaming from the title bar or is it coming from somewhere else?

jtpio:
It's coming from here:

https://github.com/jupyterlab/jupyterlab/blob/5690e6f8aa05c1d34fd667e19c70fd956b1c7c0a/packages/services/src/contents/index.ts#L799

Happening when renaming from the title bar and the file browser context menu:

https://user-images.githubusercontent.com/591645/176659581-c2c3c2fd-0441-4fef-97eb-409933717ffc.mp4

martinRenou:
Thanks. So it looks like JupyterLab is wildly trying to rename the file when it shouldn't.

Note that your first rename was a success, it's just an issue with the filebrowser not refreshing the view. It's unclear to me it it's an issue on the JupyterLab side or on the jupyterlab-filesystem-access side.

martinRenou:
I am not able to reproduce the errors you see locally.

jtpio:
Maybe by installing the prebuilt `jupyterlab-filesystem-access` extension and loading it with the `--extensions-in-dev-mode` flag?

martinRenou:
> Maybe by installing the prebuilt jupyterlab-filesystem-access extension in the same environment as the lab dev install and loading it with the --extensions-in-dev-mode flag?

Yes that is what I do

PR #12767, opened 2022-07-04
Pin nbclassic below 0.4.0

### LINKED ISSUE #12766: nbclassic 0.4.0 breaks 3.x CI
## Description

nbclassic 0.4.0 is dropping notebook as dependency to prepare for notebook v7:

https://github.com/jupyter/nbclassic/blob/859bf813e9564d1848a051f8d4d04325df7e6a0a/setup.cfg#L30

This removes notebook as a dependency of JupyterLab v3 => seen in [CI errors](https://github.com/jupyterlab/jupyterlab/runs/7175603541?check_suite_focus=true).

fcollonval:
## References

Fix #12766

## Code changes

Pin nbclassic below 0.4.0 to keep notebook as dependency of JLab v3

## User-facing changes

N/A

N/A

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,8 @@ install_requires =
     jupyter_core
     jupyterlab_server~=2.10
     jupyter_server~=1.16
-    nbclassic~=0.2,<0.4.0
+    nbclassic
+    notebook<7
     jinja2>=2.1
 
 [options.extras_require]
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ install_requires =
     jupyter_core
     jupyterlab_server~=2.10
     jupyter_server~=1.16
-    nbclassic~=0.2
+    nbclassic~=0.2,<0.4.0
     jinja2>=2.1
 
 [options.extras_require]


jtpio:
cc @echarles for awareness.

I believe we still want nbclassic to be functional with JupyterLab 3.x and not break the notebook dependency?

echarles:
> I believe we still want nbclassic to be functional with JupyterLab 3.x and not break the notebook dependency?

Yes. JupyterLab could depend on latest `nbclassic` and specifically add `notebook<7` as additional dependency. That should work.

setup.cfg
```diff
@@ -37,7 +37,7 @@ install_requires =
     jupyter_core
     jupyterlab_server~=2.10
     jupyter_server~=1.16
-    nbclassic~=0.2
+    nbclassic~=0.2,<0.4.0
```
echarles, line 40:
```suggestion
    nbclassic
    notebook<7
```

fcollonval:
Thanks @echarles for the review

fcollonval:
@echarles do you know why the usage job are failing? In particular why do we get:

```
[C 07:34:19.709 NotebookApp] Bad config encountered during initialization: The 'contents_manager_class' trait of a NotebookApp instance expected a subclass of notebook.services.contents.manager.ContentsManager or jupyter_server.contents.services.managers.ContentsManage, not the LargeFileManager LargeFileManager.
```

echarles:
@fcollonval Looks like the complete log is

```
[W 07:58:51.315 NotebookApp] Config option `nbserver_extensions` not recognized by `NotebookApp`.  Did you mean one of: `reraise_server_extension_failures, server_extensions`?
[C 07:58:51.317 NotebookApp] Bad config encountered during initialization: The 'contents_manager_class' trait of a NotebookApp instance expected a subclass of notebook.services.contents.manager.ContentsManager or jupyter_server.contents.services.managers.ContentsManage, not the LargeFileManager LargeFileManager.
```

The first like drives me to https://github.com/jupyterlab/jupyterlab/blob/6bf3f3b285f238b28e8e3b47e3f3a65249c41513/jupyter-config/jupyter_notebook_config.d/jupyterlab.json#L3 which prolly does not make sense anymore without notebook installed. I am not sure also if it was needed before.

I don't have a clear answer atm, but would like to step back asking what is the goal? I guess we want to ensure that the command `jupyter lab` allows the user to fall back to `/tree` and get the classic experience. I would be inclined to remove that `jupyter_notebook_config.d` folder and add to https://github.com/jupyterlab/jupyterlab/blob/6bf3f3b285f238b28e8e3b47e3f3a65249c41513/jupyter-config/jupyter_server_config.d/jupyterlab.json#L4 an entry for `nbclassic` as nbclassic ui is now served as server extension

fcollonval:
Although I get your fix proposal, it seems a too large deviation in behavior for a JupyterLab patch release. I'll bring this point in today's call.

echarles:
> Although I get your fix proposal, it seems a too large deviation in behavior for a JupyterLab patch release. I'll bring this point in today's call.

Yes, I was thinking the same thing. jlab3 could pin nbclassic < 0.4.0. The proposal in https://github.com/jupyterlab/jupyterlab/pull/12767#issuecomment-1175831332 could target jlab4.

echarles:
I have opened https://github.com/jupyter/nbclassic/pull/123 that fixes the issue. `python -m jupyterlab.browser_check --notebook` runs fine on my env. The idea is to not shim the notebook package if notebook<7 is available, which is the case for juypterlab3.

We would need to release a new nbclassic to try this out in this PR.

echarles:
The tests are running fine.

PR #12737, opened 2022-06-27
Update dependencies, remove `markdown-loader-jest`

NextFire:
## References

Trying to fix #12736.

## Code changes

- `yarn upgrade`
- bump `lighthouse` to 9.3.0 and `package-json` to ^7.0.0 to upgrade `got`
- remove `markdown-loader-jest` dependency, ~~adding this 3-line loader directly in src~~ (unused)

## User-facing changes

None

## Backwards-incompatible changes

None (?)

Issue #12736, opened 2022-06-27
CVE vulnerabilities in some dependencies


NextFire:
## Description

- `testutils` uses [`markdown-loader-jest`](https://github.com/gpincheiraa/markdown-loader-jest), a 5 years old lib that locks `marked` to ^0.3.9, vulnerable to the severe https://github.com/advisories/GHSA-rrrm-qjm4-v8hf
- Old versions of `package-json` in `buildutils` and `lighthouse` in `testutils` locks `got` to ^9.6.0, vulnerable to https://github.com/advisories/GHSA-pfrx-2q88-qq97
- (Out of scope for now: JLab uses `verdaccio`, which one of its dependancies locks `jsdom` to 15.2.1, vulnerable to https://github.com/advisories/GHSA-f4c9-cqv8-9v98, [open pr in their end](https://github.com/verdaccio/monorepo/pull/592))

Context: I had these vulnerabilities reported by Dependabot on one of my JupyterLab extensions projects.

## Reproduce

N/A

## Expected behavior

Updated dependancies to patch these vulnerabilities.

## Context

- Operating System and version: 
- Browser and version: 
- JupyterLab version:  989683cb7f83a9fbbe6caea1e1b0cfb9096f8d16

fcollonval:
Closing as verdaccio was updated as side effect of https://github.com/jupyterlab/jupyterlab/pull/12095

willarmiros:
@fcollonval is there any chance that the PRs fixing this issue, specifically #12776, will get applied to the `3.6.x` branch as these are security issues?

PR #12776, opened 2022-07-06
Removes lighthouse and markdown-loader-jest

fcollonval:
## References

Partially addresses #12736
Superseeds #12737

## Code changes

- Drop lighthouse (not use by regular maintainers, benchmark tests are more meaningful).
- Drop unused markdown-jest-loader

## User-facing changes

N/A

N/A


jtpio:
Thanks!

Makes sense to clean that up if it's not used anymore. 

There seems to be some dangling references to it in the contributing docs: https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#performance-testing. Maybe we can also take the opportunity to remove them?

PR #12779, opened 2022-07-06
Updates ydoc

hbcarlos:
Updates jupyter-ydoc

cc @davidbrochart @krassowski

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ install_requires =
     notebook_shim>=0.1
     jinja2>=3.0.3
     ypy-websocket
-    jupyter_ydoc==0.1.12
+    jupyter_ydoc==0.1.13
 
 [options.extras_require]
 docs =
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     notebook_shim>=0.1
     jinja2>=3.0.3
     ypy-websocket
-    jupyter_ydoc==0.1.10
+    jupyter_ydoc==0.1.12
```
davidbrochart, line 45: Did you mean v0.1.13?
```suggestion
    jupyter_ydoc==0.1.13
```

hbcarlos: yep

davidbrochart: Another option would be `jupyter_ydoc>=0.1.13,<0.2.0`, so that we don't update so often in JupyterLab.
And we make sure that each time we change the document structures in jupyter-ydoc, we do a minor bump.

jtpio (approved):
Thanks!

PR #12785, opened 2022-07-08
Use Vega SVG renderer to drop canvas dependency

### LINKED ISSUE #12726: Galata fails to install due to node-canvas
## Description
When using a VPN, the installation of Galata fails due to the [canvas](https://github.com/jupyterlab/jupyterlab/blob/c2907074e58725942946a73a823fc60e1795da39/galata/package.json#L62) dependency failing to install. The install of canvas fails because canvas tries to [download from a link](https://github.com/Automattic/node-canvas/blob/1f2b156a2c1da29118d55c2a1747a04a87e02d7a/package.json#L38) during the install process (It seems reasonable that a VPN would block this kind of thing). We have created an [issue with canvas](https://github.com/Automattic/node-canvas/issues/2017) which has not had a response since it was opened a few months ago.

Without canvas making any changes, this seems like room to improve Galata to avoid situations like this. Canvas is only used in the [`benchmarkReporter.ts`](https://github.com/jupyterlab/jupyterlab/blob/master/galata/src/benchmarkReporter.ts) which uses playwright to benchmark performance in CI. Given that "Galata is a set of helpers and fixtures for JupyterLab UI Testing," it does not seem like benchmark performance test should be shipped with Galata. This will also guarantee that what is being shipped with Galata is as light weight as possible for the people using it. 

Another, but less ideal, approach would be to add canvas as an optional dependency in the `package.json`, so that the install of canvas can fail silently. However, I would want to think that shipping Galata without the `benchmarkReporter.ts` would be better. 

Thanks all for the help with this :)

fcollonval:
## References

Fix #12726

## Code changes

Use Vega SVG renderer.

## User-facing changes

N/A

N/A


fcollonval (commented):
please run benchmark

fcollonval:

fcollonval:

PR #12791, opened 2022-07-11
update tab name after file rename

### LINKED ISSUE #12705: Browser tab title does not update on notebook rename
## Description

The browser tab title does not update on Notebook rename.

## Reproduce

1. Create a new Notebook. The browser title should be set to "Untitled.ipynb - JupyterLab"
2. Right click on the JupyterLab tab for the notebook, rename notebook.
3. The notebook is renamed, but the browser title does not update. 

Additionally, if you switch the active notebook to another one and back, it will update correctly.

## Expected behavior

When you rename a notebook, the browser tab's title should update automatically.

## Context

- Operating System and version: MacOS Monterrey 12.4
- Browser and version: Chrome 102
- JupyterLab version: 3.4.3

Here's a screenshot: 

image

RobbyPratl:
Fixes #12705

No other pull requests have been made to address this issue thus far.
## Code changes

https://github.com/jupyterlab/jupyterlab/blob/dc55eca669b2730ec33025e8e9bbde91c9b0629c/packages/apputils-extension/src/index.ts#L336-L345 
Since currentFile is only declared when the tabname should be the name of a file, thus I would propose changing: `let currentFile: string = PathExt.basename(current.split(':')[1]); ` to `let currentFile: string = PathExt.basename(window.location.href);`
So if the user renamed a file, the tab title would update based on the last part of the path to the new file, instead of the old file name.

## User-facing changes

Previously the browser tab did not update when the name of a file was changed as the image below shows
newtab
Afterwards, it would look like this, with both the file name and tab name matching.
previoustab

---

Edited to link the issue

### DIFF
--- a/packages/apputils-extension/src/index.ts
+++ b/packages/apputils-extension/src/index.ts
@@ -342,7 +342,7 @@ async function updateTabTitle(workspace: string, db: IStateDB, name: string) {
     }`;
   } else {
     // File name from current path
-    let currentFile: string = PathExt.basename(current.split(':')[1]);
+    let currentFile: string = PathExt.basename(window.location.href);
     // Truncate to first 12 characters of current document name + ... if length > 15
     currentFile =
       currentFile.length > 15


fcollonval:

jiewangjs:
I found there is name length limit, around 10 digits. Can this limit be extensively extended, say, to dozens ?

Issue #12019, opened 2022-02-05
Open a notebook with no kernel or an existing kernel


davidbrochart:
### Problem

Currently, opening a notebook automatically launches a new kernel. It would be nice to be able to:
- not launch a kernel, i.e. just open the notebook for viewing/editing, but not for executing.
- connect to an existing kernel, even if this kernel "belongs to" a different notebook, or a console. The existing kernel should be compatible with the one specified in the notebook.

### Proposed Solution

The `Open With` context menu looks like the right place for this. Currently, right-clicking on a notebook in the file browser shows an `Open With` menu with `Notebook`, `Editor` and `JSON`. I'm not sure if it's possible, but having the mouse over `Notebook` could show another menu with `No Kernel`, `New Kernel`, and the list of running kernels.

### Additional context

See https://github.com/jupyter/notebook/issues/3170

davidbrochart:
Is it possible to have several levels of submenus which display on mouse over? Something like:
```
Open With
    Notebook
        Don't Launch Kernel
        Launch New Kernel
        Use Last Launched Kernel
        Use Same Kernel As
            Untitled.ipynb
            Untitled1.ipynb
            Untitled2.ipynb
    Editor
    JSON
```

davidbrochart:
Actually, the popup window that appears when no kernel is set in the notebook has everything we need:

![image](https://user-images.githubusercontent.com/4711805/152875578-b1e0c3ae-59b7-4adf-9c5a-fa73a32c7957.png)

We can choose `No Kernel` or a kernel from another session.
So why not showing this window even when a kernel is set in the notebook file? It makes opening the notebook one click further though.

jasongrout:
> So why not showing this window even when a kernel is set in the notebook file? It makes opening the notebook one click further though.

That's why - the assumption was that the vast majority of the time, a user wants to open the notebook with the same kernel it was saved with.


> Is it possible to have several levels of submenus which display on mouse over? Something like:

Keep in mind that in theory any viewer can require a kernel, not just the notebook. Selecting a kernel is sort of orthogonal to selecting a viewer (though in practice, not many viewers request a kernel).

jasongrout:
One way of handling just the notebook situation (i.e., not other viewers that may request kernels) that I think is friendlier than submenus is adding a single `Notebook (no kernel)` viewer option in Open With... menu. Once you open a notebook with no kernel, you can always change the kernel to get the kernel selection box.

[...]

PR #12793, opened 2022-07-11
Add `autoStartDefault ` to `getKernelPreference`

aditya211935:
## Code changes
`getDefaultKernel` function requires `autoStartDefault` argument to return `defaultKernel`. This in turn is used to start the notebook with default kernel if no kernel is specified.

However, `autoStartDefault` property is not defined anywhere execept in the original interface. https://github.com/jupyterlab/jupyterlab/blob/b57ec962a7b8cda65bb89a72223f195564f8387a/packages/apputils/src/sessioncontext.tsx#L274

I've added `autoStartDefault` property to `NotebookWidgetFactory` as `true`. This fixes the problem.

## User-facing changes
Previously, on clicking `New Notebook` from file explorer's context menu, the notebook would be opened without any kernel. 
With these changes, the notebook opens with the default kernel.

https://user-images.githubusercontent.com/23213686/178343393-59bae348-9255-4f5e-a4a0-250bfad7dc27.mov


aditya211935:
@jasongrout Any updates?

aditya211935:
@afshin @fcollonval Any updates?

krassowski:
This would also alleviate a problem currently faced for users of https://github.com/krassowski/jupyterlab-voice-control/issues/5 extension. I am tagging it as a `enhancement` and adding to 4.0 milestone (so we have more time in case if someone comes back with a strong case for the existing behaviour).

@aditya211935 thank you for your contribution. Next time to get the PRs faster through the review please first open an issue describing the problem that you are trying to fix, as recommended and explained in the [contributing guideline](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#submitting-a-pull-request-contribution).

krassowski:
I think that this PR is compatible with other requests to improve the kernel startup like in: https://github.com/jupyterlab/jupyterlab/issues/12019 in principle. Recently a new draft PR appeared which touches similar ideas in a different user experience path: https://github.com/jupyterlab/jupyterlab/pull/12858 by @a3626a.

a3626a:
@krassowski 
> I think that this PR is compatible with other requests to improve the kernel startup like in: #12019 in principle. Recently a new draft PR appeared which touches similar ideas in a different user experience path: #12858 by @a3626a.

I changed my draft to PR, now. :)

krassowski:
Would you mind updating with the latest master branch (rebase or merge) so we can make sure the tests are passing? At the time of last commit there was something failing with the UI tests (unrelated to this PR).

krassowski:
Thanks! It looks like the visual regression tests are failing and I believe that this is relevant to this change. This is because they await for a dialog window about kernel selection, which is being skipped with this PR.

krassowski:
I think that `autoStartDefault` should be exposed in notebook settings. For now we could default it to `false` (this way tests would not need updating) and if we have more interest, change the default to `true` in a future release.

aditya211935:
@krassowski Sure, will add `autoStartDefault` to notebook settings and default it to false.

PR #12799, opened 2022-07-12
Update `verdaccio`, start registry on `0.0.0.0` instead of `localhost`

jtpio:
## References

Some CI troubleshooting since the `release_test` check has been failing recently.

## Code changes

- [x] Update the `verdaccio` dependency in `@jupyterlab/buildutils`
- [x] Start the local registry on `0.0.0.0` instead of `localhost`

## User-facing changes

None

None

### DIFF
--- a/buildutils/package.json
+++ b/buildutils/package.json
@@ -57,7 +57,7 @@
     "semver": "^7.3.2",
     "sort-package-json": "~1.53.1",
     "typescript": "~4.7.3",
-    "verdaccio": "^5.2.2"
+    "verdaccio": "^5.13.3"
   },
   "devDependencies": {
     "@types/fs-extra": "^9.0.1",
--- a/buildutils/src/local-repository.ts
+++ b/buildutils/src/local-repository.ts
@@ -26,6 +26,12 @@ async function startLocalRegistry(out_dir: string, port = DEFAULT_PORT) {
     fs.mkdirSync(out_dir);
   }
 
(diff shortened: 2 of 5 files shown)


jtpio:
So `playwright` can be installed when passing the registry explicitly with `--registry https://registry.npmjs.org`.

image


Now need to check why the `@jupyterlab` are now failing to install:

```
error An unexpected error occurred: "http://localhost:4873/@jupyterlab%2flauncher: connect ECONNREFUSED ::1:4873".
```

fcollonval:
I'm wondering if we don't get trouble because of the IPv6 addressing. But I don't know if we can go around that. Or maybe we need to use the same trick as for docker using 0.0.0.0 instead of localhost.

jtpio:
> Or maybe we need to use the same trick as for docker using 0.0.0.0 instead of localhost.

Thanks @fcollonval for this it seems like it did the trick: 

image

I'm gonna clean up the other changes and maybe we can still keep the `verdaccio` update in the diff.

fcollonval (approved):
Thanks @jtpio

jtpio:
Looks like the [failing UI test](https://github.com/jupyterlab/jupyterlab/runs/7412720327?check_suite_focus=true) seems to be coming from the merge of https://github.com/jupyterlab/jupyterlab/pull/11638

fcollonval:
I'm merging this one and going to address the UI tests separately.

PR #12826, opened 2022-07-20
Add readonly indicator

### LINKED ISSUE #7980: UX: Clearer Read Only Status
# Description
Read only notebooks in lab do not have any indication that save is failing nor that they are read only.

In Notebook there are both indications:
image

The left tag comes when you try and save and the right icon is always visible on read only notebooks.

## Proposal:
We add these back in a similar spot (next to the kernel name).

I do not have a good suggestion on what to do for non-notebooks files as there is no real space in the editor. The bottom info bar (where line/col numbers are) seems like the wrong place.

## Context
- JupyterLab version: 1.2.3

telamonian:
## References

Fixes #7980

## Code changes

Adds a slightly somewhat (visually) loud indicator for the readonly status of the current document. My goal with this PR is to implement some version of UI proposals `1)` and `2)` made by @isabela-pf in #7980

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -13,8 +13,6 @@ import {
   JupyterLab
 } from '@jupyterlab/application';
 import {
-  addCommandToolbarButtonClass,
-  CommandToolbarButtonComponent,
   Dialog,
   ICommandPalette,
   InputDialog,
@@ -33,7 +31,7 @@ import {
   SavingStatus
 } from '@jupyterlab/docmanager';
 import { IDocumentProviderFactory } from '@jupyterlab/docprovider';
-import { DocumentRegistry } from '@jupyterlab/docregistry';
+import { DocumentRegistry, DocumentWidget } from '@jupyterlab/docregistry';
 import { Contents, Kernel } from '@jupyterlab/services';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { IStatusBar } from '@jupyterlab/statusbar';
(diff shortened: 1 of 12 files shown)


telamonian:
Currently working, albeit with wonky styling (aligning raw svg elements in the toolbar is arcane):

image

I think what I'd like to do to improve the look and readability of the above is to merge the readonly label into the `save` button next door. So instead of just showing up with the typical lumino `disabled` style, when the doc is readonly the `save` button will have a crossed-out icon, a blue background, and a "readonly document" label.

I'm also open to alternative proposals. Pinging @isabela-pf @mlucool @jasongrout

ellisonbg:
I am talking to a team on Friday that has built an extension that has an improved read-only notebook UI. I haven't seen it, but can report back once I learn more. 

If we do decide to put the visual indicator in the notebook toolbar, I would probably put it on the right side to keep with our current pseudo-separation between button on the L and indicators on the R (I realize that some elements on the R are not purely visual indicators, but I think they function that way mostly).

fcollonval:
> I would probably put it on the right side

:+1: 

> aligning raw svg elements in the toolbar is arcane

If you look at the toolbar implementation the label is place next to the icon and not within it:

https://github.com/jupyterlab/jupyterlab/blob/master/packages/ui-components/src/components/toolbar.tsx#L665

I would also not use `addCommandToolbarButtonClass` to avoid any sense of actionable item.

krassowski:
@telamonian I tagged this for 4.0 but I am not sure if you plan to work on this prior to the feature freeze date? If not then this can likely still go in 4.1 (at least in the current state it seems to be additive change which does not break API).

krassowski:
Closing in favour of #14600.

Issue #8604, opened 2020-06-23
Does the use-sys-dir work?


afshin:
Does the `--use-sys-dir` flag work?

## Description
 I'm trying to run JupyterLab from a specified application directory `--app-dir=./foo` and to *not* use the system directory's extensions with `--use-sys-dir=False` and I'm getting on JupyterLab 2.x and also JupyterLab 1.x:

```
[C 13:39:29.634 LabApp] Bad config encountered during initialization:
[C 13:39:29.634 LabApp] Invalid argument: '--use-sys-dir=False'
```

## Reproduce
```
jupyter lab --app-dir=./foo --use-sys-dir=False
```
## Expected behavior
This flag should be accepted as valid configuration and ignore the system directory.

## Context

cf. https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/commands.py#L314-L317

- Operating System and version: macOS and also Linux
- Browser and version: N/A
- JupyterLab version: 1.x and 2.x

blink1073:
cc @vidartf who has more insight into `AppOptions` usage.

vidartf:
I added that as an escape hatch on the `AppOptions` object so that the option was available (either for other apps reusing the commands, or for Lab to use in the future), but it was not wired through to the relevant application entry points. There are quite a few entry points that might be relevant in ordinary lab usage (the main entry `jupyter lab` is the most important one, but I think it will be relevant for all the `labextension` and `build` entry points as well, maybe more). All of those entry points would need to add the trait + alias and then parse that into the `AppOptions` object they create. There might be a more DRY method to expose various options on a set of entry points, but that would then need to be written.

Sorry if the presence of this option gave you false hope 😞 

A more efficient way to achieve what you want in the interim is probably to set the environment variable `JUPYTERLAB_DIR` as that will set both the appdir *and* the sys dir.

afshin:
Cheers; thanks for the clarification @vidartf!

PR #12844, opened 2022-07-22
Add an option to enable "fast checks" of the jupyter lab build.

### LINKED ISSUE #12797: Allow skipping NPM download / hash checks when doing build check.
### Problem

When launching a jupyterlab instance in our JupyterHub cluster, the `build_check` request can take as much as a minute despite the fact that the image that we have built is in a known-good state.

### Proposed Solution

Add an option to `AppOptions` named something like `fast_build_check` (or something else; this is just a first proposal) that a user or admin can set to default to using fast build-checks (with the option at any location that calls `build_check` to override this default).

I have a first cut of a solution here: https://github.com/TileDB-Inc/jupyterlab/commit/068b8e41f091e6d3a0334f1e60aeb99ee4a9ee87 but am open to changes or other approaches.

Thank you!

thetorpedodog:
This adds the `fast_build_check` option to the `AppOptions`
configuration. A user can set this option to `true` to avoid performing
expensive work like downloading data from NPM and computing hashes when
the install is in a known-good state or the user cannot actually trigger
a rebuild.

## References

Fixes #12797.

## Code changes

[see change description above]

## User-facing changes

No user-facing changes.

No backwards-incompatible changes.

### DIFF
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -352,6 +352,14 @@ class AppOptions(HasTraits):
 
     splice_source = Bool(False, help="Splice source packages into app directory.")
 
+    skip_full_build_check = Bool(
+        False,
+        help=(
+            "If true, perform only a quick check that the lab build is up to date."
+            " If false, perform a thorough check, which verifies extension contents."
+        ),
+    )
+
     @default("logger")
     def _default_logger(self):
         return logging.getLogger("jupyterlab")
@@ -608,6 +616,7 @@ class _AppHandler(object):
         self.labextensions_path = options.labextensions_path
         self.kill_event = options.kill_event
         self.registry = options.registry
+        self.skip_full_build_check = options.skip_full_build_check
 
         # Do this last since it relies on other attributes
         self.info = self._get_app_info()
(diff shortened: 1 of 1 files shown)


fcollonval (changes requested):
Thanks @thetorpedodog I would go for a wording like `skip_build_check` that highlights a risk of turning it on.

Otherwise this is looking good to me.

thetorpedodog:
> I would go for a wording like `skip_build_check` that highlights a risk of turning it on.

I went with `skip_full_build_check` to reflect that there still is *a* build check going on, but not a complete check. If you think `skip_build_check` is a better name, I will gladly change it, though.

Thanks for the review!

thetorpedodog:
@fcollonval a gentle bump in case this fell off your radar

fcollonval (approved):
Thanks for the bump and the follow-up @thetorpedodog

fcollonval:

jhgoebbert:
This reads like a great feature. But how is it supposed to be turned on 🤔   

Looking at the code it reads for me as if it is not yet exposed to some entry points like `jupyter lab`.  
It seems to be non-accessible same as this option: https://github.com/jupyterlab/jupyterlab/issues/8604#issuecomment-648287294

PR #12849, opened 2022-07-24
Fixes renaming files from title while using a custom drive

hbcarlos:
Fixes #12755
cc @jtpio @martinRenou 

Yes, the rename dialog now receives the context instead of the path.

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -564,7 +564,7 @@ export class Context<
     if (drivName) {
       newPath = `${drivName}:${newPath}`;
     }
-    
+
     await this._manager.contents.rename(this.path, newPath);
     await this.sessionContext.session?.setPath(newPath);
     await this.sessionContext.session?.setName(newName);
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -557,10 +557,14 @@ export class Context<
    * @param newName - the new name for the document.
    */
   private async _rename(newName: string): Promise<void> {
-    const splitPath = this.path.split('/');
+    const splitPath = this.localPath.split('/');
     splitPath[splitPath.length - 1] = newName;
-    const newPath = splitPath.join('/');
-
+    let newPath = PathExt.join(...splitPath);
+    const drivName = this._manager.contents.driveName(this.path);
+    if (drivName) {
+      newPath = `${drivName}:${newPath}`;
+    }
+    
     await this._manager.contents.rename(this.path, newPath);
     await this.sessionContext.session?.setPath(newPath);
     await this.sessionContext.session?.setName(newName);
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -560,9 +560,9 @@ export class Context<
     const splitPath = this.localPath.split('/');
     splitPath[splitPath.length - 1] = newName;
     let newPath = PathExt.join(...splitPath);
-    const drivName = this._manager.contents.driveName(this.path);
-    if (drivName) {
-      newPath = `${drivName}:${newPath}`;
+    const driveName = this._manager.contents.driveName(this.path);
+    if (driveName) {
+      newPath = `${driveName}:${newPath}`;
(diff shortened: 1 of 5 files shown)


martinRenou:
Thanks! If I understand correctly it should close https://github.com/jupyterlab/jupyterlab/pull/12755 ?

packages/docregistry/src/context.ts
```diff
@@ -557,9 +557,13 @@ export class Context<
...
   private async _rename(newName: string): Promise<void> {
-    const splitPath = this.path.split('/');
+    const splitPath = this.localPath.split('/');
     splitPath[splitPath.length - 1] = newName;
-    const newPath = splitPath.join('/');
+    let newPath = PathExt.join(...splitPath);
+    const drivName = this._manager.contents.driveName(this.path);
+    if (drivName) {
+      newPath = `${drivName}:${newPath}`;
+    }
```
jtpio, line 566: Looks like a minor typo:

```suggestion
    const driveName = this._manager.contents.driveName(this.path);
    if (driveName) {
      newPath = `${driveName}:${newPath}`;
    }
```

jtpio:
Thanks @hbcarlos for working on this.

Just tried on Binder after installing the `jupyterlab-filesystem-access` extension (https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access) and there seems to be some cases where the new name is not propagated (to the file browser or the main area widget title):

https://user-images.githubusercontent.com/591645/181223372-0a52a407-a62a-464b-bb37-82da90fd3278.mp4

Not sure yet but this might also be related to the extension itself.

hbcarlos:
Thanks, @jtpio! Let me check what is going on with the update of the title.

> Not sure yet but this might also be related to the extension itself.

Yes, I believe when renaming from the file browser, the title is not updated because the file browser doesn't emit the signal `fileChanged` (or something similar).

But I'm not sure what is going on when renaming from the main menu. I need to check that.

hbcarlos:
@jtpio I just tried locally, and if you wait a bit, the file gets renamed.
The remaining issue comes from the [jupyterlab-filesystem-access](https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access) extensions. As you can see here:
https://github.com/jupyterlab/jupyterlab/blob/8066d229cabe84ae92af5dd9db03a56bf92c548d/packages/services/src/contents/index.ts#L1189-L1212
the default drive from lab, emits a signal after renaming the file but that's not the case with the extension, see:
https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/blob/3c7d326a8de6b79c7f506d02ead949bd54eca384/src/drive.ts#L211-L221
and the issue we are seeing in your screencast is that the file browser doesn't update the list of files until it fetches them.

We can merge this PR and fix the remaining issues in the extension.

jtpio:
> Version `0.5.2` of the `jupyterlab-filesystem-access` extension was just released: https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/releases/tag/v0.5.2

Will have a second look on Binder with this new version.

jtpio:
Arf Binder seems to be broken (looks unrelated to this PR):

image

jtpio:
This is likely because of the `ypy-websocket` issue:

```
ImportError: cannot import name 'YDoc' from 'ypy_websocket.websocket_server'
```

jtpio:
Rebasing to make sure we grab the latest state of master.

jtpio:
Looking better after the rebase. Just checked on Binder and it seems to be behaving correctly:

https://user-images.githubusercontent.com/591645/182403691-8cae6f0d-43c7-40d8-a639-9f6560c2899f.mp4

jtpio:
> Yes, the rename dialog now receives the context instead of the path.

Just pushed a commit to add this change to https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-3-x-to-4-x.

image

hbcarlos:
Thanks, @jtpio!

jtpio (approved):
Thanks!

PR #12379, opened 2022-04-08
Pop up select kernel dialog when run a cell without kernel

a3626a:
## Introduction

Experienced users can manage their kernels and resources, but beginners like students are not familiar with managing resources. I am implementing useful features for them, including [limiting number of running kernels](https://github.com/jupyter/notebook/issues/615) and this PR.

In some cases, kernels can be killed while the file tabs are still open.
1) When users kill their kernel manually,
image
2) By [custom kernel managers](https://github.com/jupyter/notebook/issues/615#issuecomment-1092382232)

In such cases, Jupyter Lab does nothing when users try to execute cells.
_no kernel, before execution_
image

_no kernel, after execution_
image

I think poping up select kernel dialog is much better user experience especially for beginners.
_with this PR, after execution_
image

## Code changes

I've found some cases that pop up dialogs before cell execution, like
image

This feature is implemented in `packages/notebook/src/actions.tsx`

I just added a new condition and a dialog there. I am not sure about what to return `Promise.resolve(true)` or `Promise.resolve(false)`

## User-facing changes

_before, no response from the Lab_
image

_after, select kernel dialog pops up_
image

## Backwards-incompatible changes

No.

Issue #4040, opened 2018-02-27
Open Notebook without kernel


davidchall:
There are times when opening a Notebook does not require a new kernel to be started (e.g. reading or only editing Markdown cells). Currently these functions are still possible after you have shutdown the kernel.

Although starting up a kernel is usually inexpensive, there are times when this isn't the case (e.g. connecting a PySpark kernel to YARN). There may also be times when you want to view a Notebook but don't have the right kernel installed.

Perhaps we could have a new option in the "Open With" menu for this purpose? I suggest it might be called "Notebook (no kernel)".

thincal:
+1

ian-r-rose:
@davidchall Note that you can achieve this by opening a new notebook from the file menu and selecting "No Kernel" from the selection dropdown.

thavidu:
Seems like this could be an opposite default of the existing
```
    // Shut down kernel
    // Whether to shut down or not the kernel when closing a notebook.
    "kernelShutdown": false,
```
setting

I wish there was something like:
```
// Whether to automatically start a kernel when opening a notebook
"kernelAutostart": true,
```

tilusnet:
> @davidchall Note that you can achieve this by opening a new notebook from the file menu and selecting "No Kernel" from the selection dropdown.

I am not sure I follow. The OP wants to open an existing notebook without kernel, not a new notebook.

rvalieris:
I will go a step further and say that instead of an option, it should behave like on google colab, where a kernel is only started if the user tries to execute a code cell.

very often I will open a old notebook just to check what I did last time, if I am not careful I end up with a bunch of idle kernels open because of this,  I need to manually close each one.

there are also other bugs related to this behavior, such as: https://github.com/jupyterlab/jupyterlab/issues/8574, and related discussions: https://github.com/jupyterlab/jupyterlab/issues/5241

PR #12858, opened 2022-07-26
Run kernel on cell execution when no kernel

a3626a:
## Background
There are several requests about opening notebook file without kernel. `KernelPreference` already have `shouldStart` boolean flag. This flag can be set through jupyter server argument `--LabServerApp.notebook_starts_kernel=False`. This flag works as intended, so it prevents kernels to be launched automatically. 

Nevertheless, we need more improvement here. When users execute a code cell in a notebook without kernel, it shows a dialog:
image
(It was done by this PR https://github.com/jupyterlab/jupyterlab/pull/12379)
When `shouldStart` is false, user must select kernels for every notebook file opened. It is very annoying, and bad user experience.

So I suggest a new behaviors for launching a new kernel on code execution.
- _Requirement_ **A**: Automatically selects kernel if possible
- _Requirement_ **B**: Keep executing code after starting the kernel
  - or select the kernel (and start).

## References
- https://github.com/jupyterlab/jupyterlab/issues/4040
- https://github.com/jupyterlab/jupyterlab/issues/12019
  - Especially https://github.com/jupyterlab/jupyterlab/issues/12019#issuecomment-1145841583

This closes https://github.com/jupyterlab/jupyterlab/pull/12793 as it set the `autoStartDefault`.

### SessionContext
- Extract kernel start logic from `_startIfNecessary` into a new public method `startKernel`
- Loose restriction for `getDefaultKernel`
  - don't care about `shouldStart`
    - it is backwards-incompatible changes, but I think backwards-compatible changes are possible. Help needed
  - `startKernel` depends on `getDefaultKernel` to automatically select a kernel.
- `SessionContextDialogs` default global are removed - you should use the token `'@jupyterlab/apputils-extension:sessionDialogs'` (This is a similar change to the sanitizer or the content factory to be more token based)
- To be able to set `autoStartDefault` only for notebook (with a settings in notebook-extension package) with a check box in the kernel selection dialog, the logic is a complex. But basically, if `autoStartDefault` is not `undefined` a checkbox will be displayed. The checkbox value will be used to update the session context kernel preference. That update will trigger a new `kernelPreferenceChanged` in `ISessionContext`. That signal is listen to in the notebook extension to be able to update the settings.

### Actions
- runCell
  - if no kernel, call `startKernel`
    - it satisfies _Requirement_ **A**
  - work on promise to satisfy _Requirement_ **B**

## User-facing changes

When a notebook is opened without kernel and a user tries to execute a code cell, the user don't have to select a kernel. Additionally, code is executed immediately after the kernel executed.

When the user is opening a new notebook, if the new settings is true, the associated kernel will be started directly.

### Before (JLab 4.0.0a34)

jlab400a34

### After

thispr

global `sessionContextDialogs` from `@jupyterlab/apputils` does not exist any longer.

~In my initial draft, the semantic of `getDefaultKernel` changed. It is not a serious but a backwards-incompatible change. I think it is possible to satify _Requirement_ **A** without editing `getDefaultKernel`, but some level of code duplication is inevitable.~

~I need help to choose the direction, 1) ignoring small level of backwards-incompatibility or 2) accepting a little bit of duplication.~

(Edited) `getDefaultKernel` is a private function. I made every function which calls `Private.getDefaultKernel` keep old behavior.

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -32,6 +32,7 @@ bumped their major version (following semver convention). We want to point out p
    * Rename ``IToolbarWidgetRegistry.registerFactory`` to ``IToolbarWidgetRegistry.addFactory``
    * ``ISanitizer`` and ``ISanitizer.IOptions`` are deprecated in favor of ``IRenderMime.ISanitizer`` and
      ``IRenderMime.ISanitizerOptions`` in ``@jupyterlab/rendermime-interfaces``.
+   * Global ``sessionContextDialogs`` is removed; you should request the token ``ISessionContextDialogs`` (from ``@jupyterlab/apputils``).
 - ``@jupyterlab/attachments`` from 3.x to 4.x
    Removed ``modelDB`` from ``IAttachmentsModel.IOptions``.
 - ``@jupyterlab/buildutils`` from 3.x to 4.x
@@ -198,6 +199,7 @@ bumped their major version (following semver convention). We want to point out p
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    * The ``NotebookPanel._onSave`` method is now ``private``.
    * ``NotebookActions.collapseAll`` method renamed to ``NotebookActions.collapseAllHeadings``.
+   * ``NotebookActions.run`` and other run commands now require ``ISessionContext`` and ``ISessionContextDialogs``.
    * Command ``Collapsible_Headings:Toggle_Collapse`` renamed to ``notebook:toggle-heading-collapse``.
    * Command ``Collapsible_Headings:Collapse_All`` renamed to ``notebook:collapse-all-headings``.
    * Command ``Collapsible_Headings:Expand_All`` renamed to ``notebook:expand-all-headings``.
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -199,7 +199,7 @@ bumped their major version (following semver convention). We want to point out p
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    * The ``NotebookPanel._onSave`` method is now ``private``.
    * ``NotebookActions.collapseAll`` method renamed to ``NotebookActions.collapseAllHeadings``.
-   * ``NotebookActions.run`` and other run commands now require ``ISessionContext`` and ``ISessionContextDialogs``.
+   * ``NotebookActions.run`` and other run commands now require ``ISessionContext``.
    * Command ``Collapsible_Headings:Toggle_Collapse`` renamed to ``notebook:toggle-heading-collapse``.
    * Command ``Collapsible_Headings:Collapse_All`` renamed to ``notebook:collapse-all-headings``.
    * Command ``Collapsible_Headings:Expand_All`` renamed to ``notebook:expand-all-headings``.
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -197,9 +197,9 @@ bumped their major version (following semver convention). We want to point out p
    * ``IMainMenu.addMenu`` signature changed from ``addMenu(menu: Menu, options?: IMainMenu.IAddOptions): void``
      to ``addMenu(menu: Menu, update?: boolean, options?: IMainMenu.IAddOptions): void``
 - ``@jupyterlab/notebook`` from 3.x to 4.x
+   * ``NotebookWidgetFactory.IOptions`` has no ``sessionDialogs`` option any more.
    * The ``NotebookPanel._onSave`` method is now ``private``.
    * ``NotebookActions.collapseAll`` method renamed to ``NotebookActions.collapseAllHeadings``.
-   * ``NotebookActions.run`` and other run commands now require ``ISessionContext``.
    * Command ``Collapsible_Headings:Toggle_Collapse`` renamed to ``notebook:toggle-heading-collapse``.
    * Command ``Collapsible_Headings:Collapse_All`` renamed to ``notebook:collapse-all-headings``.
    * Command ``Collapsible_Headings:Expand_All`` renamed to ``notebook:expand-all-headings``.
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -13,10 +13,11 @@ import {
   ISessionContext,
   ISessionContextDialogs,
   KernelStatus,
-  RunningSessions
+  RunningSessions,
+  SessionContextDialogs
 } from '@jupyterlab/apputils';
 import { IStatusBar } from '@jupyterlab/statusbar';
-import { ITranslator } from '@jupyterlab/translation';
+import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { Title, Widget } from '@lumino/widgets';
 
 /**
@@ -25,23 +26,26 @@ import { Title, Widget } from '@lumino/widgets';
 export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
   id: '@jupyterlab/apputils-extension:kernel-status',
   autoStart: true,
-  requires: [IStatusBar, ISessionContextDialogs, ITranslator],
+  requires: [IStatusBar],
   provides: IKernelStatusModel,
-  optional: [ILabShell],
+  optional: [ISessionContextDialogs, ITranslator, ILabShell],
   activate: (
     app: JupyterFrontEnd,
     statusBar: IStatusBar,
-    sessionDialogs: ISessionContextDialogs,
-    translator: ITranslator,
+    sessionDialogs: ISessionContextDialogs | null,
+    translator: ITranslator | null,
     labShell: ILabShell | null
   ): IKernelStatusModel => {
+    translator = translator ?? nullTranslator;
+    sessionDialogs =
+      sessionDialogs ?? new SessionContextDialogs({ translator });
     // When the status item is clicked, launch the kernel
     // selection dialog for the current session.
     const changeKernel = async () => {
       if (!item.model.sessionContext) {
         return;
       }
-      await sessionDialogs.selectKernel(item.model.sessionContext);
+      await sessionDialogs!.selectKernel(item.model.sessionContext);
     };
 
     // Create the status item.
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -32,20 +32,20 @@ export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
   activate: (
     app: JupyterFrontEnd,
     statusBar: IStatusBar,
-    sessionDialogs: ISessionContextDialogs | null,
-    translator: ITranslator | null,
+    sessionDialogs_: ISessionContextDialogs | null,
+    translator_: ITranslator | null,
     labShell: ILabShell | null
   ): IKernelStatusModel => {
-    translator = translator ?? nullTranslator;
-    sessionDialogs =
-      sessionDialogs ?? new SessionContextDialogs({ translator });
+    const translator = translator_ ?? nullTranslator;
+    const sessionDialogs =
+      sessionDialogs_ ?? new SessionContextDialogs({ translator });
     // When the status item is clicked, launch the kernel
     // selection dialog for the current session.
     const changeKernel = async () => {
       if (!item.model.sessionContext) {
         return;
       }
-      await sessionDialogs!.selectKernel(item.model.sessionContext);
+      await sessionDialogs.selectKernel(item.model.sessionContext);
     };
 
     // Create the status item.
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -13,8 +13,7 @@ import {
   ISessionContext,
   ISessionContextDialogs,
   KernelStatus,
-  RunningSessions,
-  sessionContextDialogs
+  RunningSessions
 } from '@jupyterlab/apputils';
 import { IStatusBar } from '@jupyterlab/statusbar';
 import { ITranslator } from '@jupyterlab/translation';
@@ -26,14 +25,14 @@ import { Title, Widget } from '@lumino/widgets';
 export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
   id: '@jupyterlab/apputils-extension:kernel-status',
   autoStart: true,
-  requires: [IStatusBar, ITranslator],
+  requires: [IStatusBar, ISessionContextDialogs, ITranslator],
   provides: IKernelStatusModel,
-  optional: [ISessionContextDialogs, ILabShell],
+  optional: [ILabShell],
   activate: (
     app: JupyterFrontEnd,
     statusBar: IStatusBar,
+    sessionDialogs: ISessionContextDialogs,
     translator: ITranslator,
-    sessionDialogs: ISessionContextDialogs | null,
     labShell: ILabShell | null
   ): IKernelStatusModel => {
     // When the status item is clicked, launch the kernel
@@ -42,10 +41,7 @@ export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
       if (!item.model.sessionContext) {
         return;
       }
-      await (sessionDialogs ?? sessionContextDialogs).selectKernel(
-        item.model.sessionContext,
-        translator
-      );
+      await sessionDialogs.selectKernel(item.model.sessionContext, translator);
     };
 
     // Create the status item.
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -41,7 +41,7 @@ export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
       if (!item.model.sessionContext) {
         return;
       }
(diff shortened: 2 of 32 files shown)


packages/notebook/src/actions.tsx
```diff
@@ -2273,51 +2273,67 @@ Please wait for the complete rendering before invoking that action.`,
             });
             return Promise.resolve(false);
           }
+          let promise = Promise.resolve();
```
a3626a, line 2276: need an any promise instance to chain it to `startKernel` or `CodeCell.execute`

packages/notebook/src/actions.tsx
```diff
@@ -2273,51 +2273,67 @@ Please wait for the complete rendering before invoking that action.`,
...
+              // Session has still no kernel, so we can't execute the cell.
+              return false;
+            }
 
-              if (!reply) {
-                return true;
-              }
-              if (reply.content.status === 'ok') {
-                const content = reply.content;
+            let deletedCells = notebook.model?.deletedCells ?? [];
```
a3626a, line 2290: code below is just wrapping it inside promise. Git diff is a bit confusing, but I do not edit details.

a3626a:
I have found that `getDefaultKernel` is a private function (is located inside `Privte` namespace) So this is not a backward incompatible change. I think this PR is OK to be merged.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -817,31 +853,7 @@ export class SessionContext implements ISessionContext {
...
-        await this._changeKernel(options);
-        return false;
-      } catch (err) {
-        /* no-op */
-      }
-    }
-
-    // Always fall back to selecting a kernel
-    return true;
+    return this.startKernel();
```
a3626a, line 856: _startIfNecessary already checked `preference.shouldStart === false` So _startIfNecessary is backward compatible

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1265,6 +1277,13 @@ export namespace SessionContext {
    * Get the default kernel name given select options.
    */
   export function getDefaultKernel(options: IKernelSearch): string | null {
+    const { preference } = options;
+    const { shouldStart } = preference;
+
+    if (shouldStart === false) {
+      return null;
+    }
+
```
a3626a, line 1286: I added this to make public `getDefaultKernel` be backward compatible.
Honestly, I don't like this change... It is very hard to understand this code without knowing history.

packages/notebook/src/actions.tsx
```diff
@@ -2201,51 +2201,67 @@ namespace Private {
...
-            .then(reply => {
-              deletedCells.splice(0, deletedCells.length);
-              if (cell.isDisposed) {
-                return false;
+            promise = sessionContext.startKernel().then(shouldSelect => {
+              if (shouldSelect) {
+                return sessionContextDialogs.selectKernel(sessionContext);
               }
+            });
+          }
```
a3626a, line 2211: Now we conditionally change promise to start new default kernel or prompt dialogs. Code will be executed after the kernel selection

fcollonval (commented):
Sorry for not looking at this earlier and thanks a lot for the PR. I understand and like the idea. But some organization prefer to be able to pick a kernel. Recently the dialog has been extended to display a checkbox in the footer: see [documentation](https://jupyterlab.readthedocs.io/en/latest/extension/ui_helpers.html#generic-dialog).

Do you think it could be possible to use it on the kernel dialog selection to add an option _Always select the preferred kernel_? That will be stored in a settings in order to allow your use case A or not. And in your deployed installation, that settings could be set to `true` to fit your need.

a3626a:
@fcollonval 
> Sorry for not looking at this earlier and thanks a lot for the PR. I understand and like the idea. But some organization prefer to be able to pick a kernel. Recently the dialog has been extended to display a checkbox in the footer: see [documentation](https://jupyterlab.readthedocs.io/en/latest/extension/ui_helpers.html#generic-dialog).
> 
> Do you think it could be possible to use it on the kernel dialog selection to add an option _Always select the preferred kernel_? That will be stored in a settings in order to allow your use case A or not. And in your deployed installation, that settings could be set to `true` to fit your need.

Sorry for my late reply. Your suggestion about checkbox is good. 

For the first, I added new setting `selectPrefferedKernel`. But updating setting by clicking checkbox is more complicated, it is not done, yet. I am checking `confirmClosingDocument` related codes.

fcollonval:
Hey sorry too for the slow reply. My two cents here is that we should drop the _default_ `sessionContextDialogs` object and only allow to get them through the token.

So the actions are:

1. at that line, we should instantiate and return the session dialogs.

https://github.com/jupyterlab/jupyterlab/blob/31886fdee1092646a782feb947877937e30c2170/packages/apputils-extension/src/index.ts#L557

2. Replace that constant object by a class

https://github.com/jupyterlab/jupyterlab/blob/31886fdee1092646a782feb947877937e30c2170/packages/apputils/src/sessioncontext.tsx#L1275

It will take the `translator: ITranslator` object - so we can drop it from `SessionContext.IOptions`.

3. Add an attribute `selectPreferredKernel` to that class.

4. Pass to the new class the `settings: ISettingsRegistry.ISettings` . Use it to get `selectPreferredKernel` and to save it when it is set to a different value.

5. Modify the dialog to bypass it or not depending of `selectPreferredKernel` and add a checkbox to set `selectPreferredKernel` depending of the checkbox status.

Does it make sense?

a3626a:
> Hey sorry too for the slow reply. My two cents here is that we should drop the _default_ `sessionContextDialogs` object and only allow to get them through the token.
> 
> So the actions are:
> 
> 1. at that line, we should instantiate and return the session dialogs.
> 
> https://github.com/jupyterlab/jupyterlab/blob/31886fdee1092646a782feb947877937e30c2170/packages/apputils-extension/src/index.ts#L557
> 
> 2. Replace that constant object by a class
> 
> https://github.com/jupyterlab/jupyterlab/blob/31886fdee1092646a782feb947877937e30c2170/packages/apputils/src/sessioncontext.tsx#L1275
> 
> It will take the `translator: ITranslator` object - so we can drop it from `SessionContext.IOptions`.
> 
> 3. Add an attribute `selectPreferredKernel` to that class.
> 4. Pass to the new class the `settings: ISettingsRegistry.ISettings` . Use it to get `selectPreferredKernel` and to save it when it is set to a different value.
> 5. Modify the dialog to bypass it or not depending of `selectPreferredKernel` and add a checkbox to set `selectPreferredKernel` depending of the checkbox status.
> 
> Does it make sense?

I finally understand what Token is. It is actually a Dependency Injection system. Thank you for your reply, it helps a lot.



> It will take the translator: ITranslator object - so we can drop it from SessionContext.IOptions

Should `translator` be dropped in this PR? It will make this PR not backward compatible.

fcollonval:
> Should translator be dropped in this PR? It will make this PR not backward compatible.

Indeed this is a good point - we can keep it to ease backporting it.

a3626a:
I have finished dropping old default sessionContextDialogs. And I found that it is somewhat impossible to make it backward compatible. So many functions have to be changed. To make it compatible, We should keep old default.

I will check tests now. It looks like Github actions are failing with some other reason during installation.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -198,6 +199,18 @@ export interface ISessionContext extends IObservableDisposable {
    */
   readonly specsManager: KernelSpec.IManager;
 
+  /**
+   * Session dialogs for selecting a kernel.
+   */
+  readonly dialogs: ISessionContext.IDialogs;
```
a3626a, line 205: I added dialogs to SessionContext. Without this, bunch of additional code changes are required to inject ISessionContext.IDialogs.

ISessionContext.IDialogs and ISessionContext is closely related, so I think it is okay to defined `dialogs` member inside ISessionContext. However in some perspective like MVC pattern, someone might think it should be avoided.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1265,14 +1304,26 @@ export namespace SessionContext {
...
   }
 }
 
 /**
  * The default implementation of the client session dialog provider.
  */
-export const sessionContextDialogs: ISessionContext.IDialogs = {
+export class SessionContextDialogs implements ISessionContext.IDialogs {
+  constructor(settings?: ISettingRegistry.ISettings, translator?: ITranslator) {
+    this._settings = settings || null;
```
a3626a, line 1323: SessionContextDialogs works with `null` settings. This makes preparing testing environment a lot easier.

packages/apputils/src/sessioncontext.tsx
fcollonval: Sorry to be annoying about this one. But indeed I prefer not to provide the `dialogs` from the context.

If this is okay with you, I could push a commit to take care of that.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1297,13 +1348,26 @@ export const sessionContextDialogs: ISessionContext.IDialogs = {
...
+        caption: trans.__(
+          'Remember my choice and always select the preferred kernel'
+        ),
+        checked: selectPreferredKernel
+      }
     });
 
     const result = await dialog.launch();
+
+    this._settings?.set('selectPreferredKernel', result.isChecked);
```
fcollonval, line 1369: We need a `catch` here.

```suggestion
    this._settings?.set('selectPreferredKernel', result.isChecked).catch(reason => { console.error("Failed to set 'selectPreferredKernel';\n", reason) });
```

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1297,13 +1348,26 @@ export const sessionContextDialogs: ISessionContext.IDialogs = {
...
       body: new Private.KernelSelector(sessionContext, translator),
-      buttons
+      buttons,
+      checkbox: {
+        label: trans.__('Always select the preferred kernel'),
+        caption: trans.__(
+          'Remember my choice and always select the preferred kernel'
+        ),
+        checked: selectPreferredKernel
+      }
```
fcollonval, line 1364: We should not show the checkbox if there is no settings plugin.

```suggestion
      checkbox: this._settings && {
        label: trans.__('Always select the preferred kernel'),
        caption: trans.__(
          'Remember my choice and always select the preferred kernel'
        ),
        checked: selectPreferredKernel
      }
```

fcollonval (changes requested):

packages/apputils/src/sessioncontext.tsx
a3626a: Thats completely OK. I added you as a collaborator to our fork.

packages/apputils/src/sessioncontext.tsx
a3626a: Actually, Setting Plugin is required for ISessionDialogs plugin

```
/**
 * The default session context dialogs extension.
 */
const sessionDialogs: JupyterFrontEndPlugin<ISessionContextDialogs> = {
  id: '@jupyterlab/apputils-extension:sessionDialogs',
  provides: ISessionContextDialogs,
  requires: [ISettingRegistry, ITranslator],
  autoStart: true,
  activate: async (
    app: JupyterFrontEnd,
    settingRegistry: ISettingRegistry,
    translator: ITranslator
  ) => {
    const settings = await settingRegistry.load(
      '@jupyterlab/notebook-extension:tracker'
    );
    return new SessionContextDialogs(settings, translator);
  }
};
```

I think it is better to make it optional, because just it is possible.

a3626a: However I have found that `@jupyterlab/apputils-extension:settings` is required by so many other plugins. So making is optional for '@jupyterlab/apputils-extension:sessionDialogs' is not meaningful.

a3626a: It is still possible to make it optional, but it cannot be tested by hand (or by some kind of integration test)

a3626a: This is what i got when disabling settings plugin
image

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -26,14 +25,14 @@ import { Title, Widget } from '@lumino/widgets';
...
   autoStart: true,
-  requires: [IStatusBar, ITranslator],
+  requires: [IStatusBar, ISessionContextDialogs, ITranslator],
   provides: IKernelStatusModel,
-  optional: [ISessionContextDialogs, ILabShell],
+  optional: [ILabShell],
   activate: (
     app: JupyterFrontEnd,
     statusBar: IStatusBar,
+    sessionDialogs: ISessionContextDialogs,
```
jtpio, line 34: Probably `ISessionContextDialogs` could still be optional?

jtpio: And use `sessionDialogs.selectKernel()` only if available.

packages/debugger-extension/src/index.ts
```diff
@@ -180,13 +180,14 @@ const files: JupyterFrontEndPlugin<void> = {
 const notebooks: JupyterFrontEndPlugin<IDebugger.IHandler> = {
   id: '@jupyterlab/debugger-extension:notebooks',
   autoStart: true,
-  requires: [IDebugger, INotebookTracker, ITranslator],
+  requires: [IDebugger, INotebookTracker, ISessionContextDialogs, ITranslator],
```
jtpio, line 183: Similar to the command above: do we need `ISessionContextDialogs` to be required?

packages/notebook/src/actions.tsx
```diff
@@ -678,7 +708,9 @@ export namespace NotebookActions {
    */
   export function runAllAbove(
     notebook: Notebook,
-    sessionContext?: ISessionContext
+    sessionContext: ISessionContext,
+    sessionDialogs: ISessionContextDialogs,
```
jtpio, line 712: Since the underlying `runCell` seems to be accepting `sessionDialogs` according to the current state of the PR (`sessionDialogs?`), maybe we could make it optional here as well?

jtpio: And in other `NotebookActions` functions like `runAll`, `runAndAdvance` and so on.

packages/debugger-extension/src/index.ts
```diff
@@ -213,7 +217,7 @@ const notebooks: JupyterFrontEndPlugin<IDebugger.IHandler> = {
         }
 
         const { content, sessionContext } = widget;
-        const restarted = await sessionDialogs.restart(sessionContext);
+        const restarted = await sessionDialogs!.restart(sessionContext);
```
jtpio, line 220: Looks like we might be able to avoid the `!` by using a different name for the service above?

For example: `const sessionDialogs = sessionContextDialogs ?? new SessionContextDialogs({ translator });`

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -26,26 +26,26 @@ import { Title, Widget } from '@lumino/widgets';
...
     // selection dialog for the current session.
     const changeKernel = async () => {
       if (!item.model.sessionContext) {
         return;
       }
-      await (sessionDialogs ?? sessionContextDialogs).selectKernel(
-        item.model.sessionContext,
-        translator
-      );
+      await sessionDialogs!.selectKernel(item.model.sessionContext);
```
jtpio, line 48: Similar to https://github.com/jupyterlab/jupyterlab/pull/12858#pullrequestreview-1330378877 we can probably avoid the `!` here.

packages/notebook-extension/src/index.ts
```diff
@@ -1420,14 +1423,14 @@ function activateNotebookHandler(
   app: JupyterFrontEnd,
   factory: NotebookWidgetFactory.IFactory,
   extensions: IEditorExtensionRegistry,
+  sessionDialogs: ISessionContextDialogs,
```
jtpio, line 1426: This one could probably be made `optional` as well.

jtpio: And we would need to change the use of the `!` operator below.

docs/source/extension/extension_migration.rst
```diff
@@ -189,6 +190,7 @@ bumped their major version (following semver convention). We want to point out p
 - ``@jupyterlab/notebook`` from 3.x to 4.x
    * The ``NotebookPanel._onSave`` method is now ``private``.
    * ``NotebookActions.collapseAll`` method renamed to ``NotebookActions.collapseAllHeadings``.
+   * ``NotebookActions.run`` and other run commands now require ``ISessionContext``.
```
jtpio, line 193: This does seem to be accurate anymore according to the latest state of the diff?

```suggestion
```

fcollonval: It is accurate - only the UI related elements (dialogs and translator) are optional:

image

jtpio: ah I thought we could get around without requiring a `ISessionContext` as this level as well.

jtpio: Why is it required in `NotebookActions.run` here:

https://github.com/jupyterlab/jupyterlab/blob/d3ae4678aca111b209e392ed0914cb565b7c4ee3/packages/notebook/src/actions.tsx#L502-L507

If it is not in the underlying `Private.runSelected` here:

https://github.com/jupyterlab/jupyterlab/blob/d3ae4678aca111b209e392ed0914cb565b7c4ee3/packages/notebook/src/actions.tsx#L2149-L2154

(links taken from the latest state of this PR)

jtpio (approved):
Thanks!

Just checked the latest state of the PR and it looks good.

Not sure about the failing license header CI check though.

jtpio:
> Not sure about the failing license header CI check though.

There is also a failing UI tests check we might want to look into or restart.

Issue #12812, opened 2022-07-18
Migration to CM6 follow up


JohanMabille:
This issue tracks the remaining parts to fix or to reimplement when https://github.com/jupyterlab/jupyterlab/pull/11638 is merged:

- [ ] LaTeX highlighting. A parser can be implemented as explained [here](https://codemirror.net/examples/lang-package/)
- ~~Multiplex mode. An example of mixed-language parsing is provided [here](https://codemirror.net/examples/mixed-language/)~~ won't be implemented.
- [x] Once multiplexing and LaTeX highlighting are fixed the visual regression tests for markdown multiplex should look [like before the CM6 merge](https://user-images.githubusercontent.com/5832902/183522825-8221f4e1-d3b9-4554-a305-c4f0454553ef.png).
- ~~Breakpoints gutters should be reworked to be smaller and the breakpoint indicator should be aligned with the line number~~
- [x] The codemirror package should provide a singleton that allow to register new themes and modes.
- [x] Python console header have syntax highlight but it should not
- [x] Source files opened from the debugger kernel source panel don't have syntax highlight
- [x] We need to remove settings that are not supported any longer (see next list)
- Many configuration options are not available anymore or should be implemented a different way,, this shouId be done when https://github.com/jupyterlab/jupyterlab/pull/11638 is merged:
  - [x] cursorBlinkRate: this can be done thanks to another [extension](https://codemirror.net/docs/ref/#view.drawSelection), however this requires some work in the CSS
  - [x] fontFamily: done
  - [x] fontSize: done
  - [x] lineHeight: done
  - [x] lineNumber: done
  - [x] lineWrap: done
  - [x] readOnly: done
  - [x] tabSize: done
  - [x] insertSpaces: done
  - [x] matchBtackets: https://github.com/jupyterlab/jupyterlab/pull/12861
  - [x] autoClosingBrackets: done
  - [x] handlePaste: TODO
  - [x] wordWrapColumn: done
  - [x] rulers: N/A ?
  - [x] codeFolding: https://github.com/jupyterlab/jupyterlab/pull/12861 - ok for file editor but does not work for Python code cells; it may be related to the ipython mode not providing the required information
  - [x] showTrailingSpaces: Probably need to be able to display white space akin to VS Code ?
  - [x] mode: done
  - [x] theme: done
  - smartIndent: N/A
  - electricChar: N/A
  - [x] keyMap: done
  - [x] extraKeys: done
  - [x] gutters: must be injected
[...]

PR #12861, opened 2022-07-26
Moved theme Map and added API to use it from other dependencies

JohanMabille:
This fixes some issues reported in #12812:

- matchBtackets
- smartIndent
- codeFolding
- foldGutter

This also adds a token for modes and themes registry.
This PR is not finished but a review would be appreciated to know if it goes in the right direction. If so I would add missing doc strings and redirection.

I also wonder if we should expose more methods from Mode via the IEditorModeRegistry interface.


fcollonval (changes requested):
Using tokens is nice. But this requires changing a bit the current structure to not use static object. In particular the `editorThemeRegistry`. They should be propagated to the `EditorConfiguration`. Or the logic should be similar to the `EditorModeRegistry` that modifies a static list.

Otherwise, providing a different theme registry overriding the token will not have the expected outcome.

If we want to align more with other APIs, the generic pattern are:

- tokens that define registries
- registries provide factories
- factories instantiate objects

But that would require more work for the code editor.

JohanMabille:
@fcollonval I removed the token API for now, so we don't have backward incompatible changes when we add it later.

> Or the logic should be similar to the EditorModeRegistry that modifies a static list.

Not sure to understand why this would work. The `EditorConfiguration` works with function that modifies the static list, so even if another extension defines another EditorModeRegistry, this later won't be used by the `EditorConfiguration` object and we'll run into the same issue as with the previous definition ot editorThemeRegistry.

My understanding is that we would also need to propagate the mode registry object to the editor configuration object to get the expected behavior. But that can be done after 4.0.

packages/codemirror/src/editortheme.ts
```diff
@@ -76,3 +79,23 @@ export const jupyterTheme: Extension = [
   jupyterEditorTheme,
   syntaxHighlighting(jupyterHighlightStyle)
 ];
+
+export namespace Theme {
```
jtpio, line 83: It would be nice to add some basic docstrings here for consistency.

fcollonval:
I'm taking over this PR.

I added a test for adding a mode (not yet pushed)
I was testing adding a theme but actually the reconfiguration of the editors are not happening. So I'm working to fix that.

fcollonval:
bot please update snapshots

fcollonval:
bot please update snapshots

fcollonval:
bot please update snapshots

fcollonval:

fcollonval:
Kicking the CI

fcollonval (approved):

PR #12865, opened 2022-07-27
Fix kernel in the statusbar does not match the actual

### LINKED ISSUE #12592: Kernel name mismatch bug between the status bar and in-notebook kernel indicator
## Description

Kernel names in status bar and the one in notebook don't match after selecting a kernel in one notebook and switching to the other one right after.

Bug

## Reproduce

1- Open 2 notebooks
2- Select kernels on both notebooks
3- Switch the kernel on one of them 
4- Switch to the other notebook right after
5- Wait until the status bar updates to show the other notebook's kernel.

## Expected behavior

I would expect the status bar to show the same kernel name as the one in notebook panel.

## Context

- Operating System and version: Linux
- Browser and version: Chrome 101
- JupyterLab version: 3.3

<pre>
</pre>
</details>

<pre>
</pre>
</details>

<pre>

</pre>
</details>

hsuanxyz:
fix #12592

## References
#12592

## Code changes

The `disconnect` be check the `thisArg`, will result in the previous callback function not being disconnected.
~~In this case, both functions are arrow functions, we don't need the `thisArg`.~~
As described in https://github.com/jupyterlab/jupyterlab/pull/12865#issuecomment-1197766873, now use the class style to call these methods

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/apputils/src/kernelstatuses.tsx
+++ b/packages/apputils/src/kernelstatuses.tsx
@@ -191,9 +191,17 @@ export namespace KernelStatus {
     }
     set sessionContext(sessionContext: ISessionContext | null) {
       this._sessionContext?.statusChanged.disconnect(
-        this._onKernelStatusChanged
+        this._onKernelStatusChanged,
+        this
       );
-      this._sessionContext?.kernelChanged.disconnect(this._onKernelChanged);
+      this._sessionContext?.connectionStatusChanged.disconnect(
+        this._onKernelStatusChanged,
+        this
+      );
+      this._sessionContext?.kernelChanged.disconnect(
+        this._onKernelChanged,
+        this
+      );
 
       const oldState = this._getAllState();
       this._sessionContext = sessionContext;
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @hsuanxyz for the fix :+1: 

This `KernelStatus.Model` class currently mixes the style of method definition which can indeed lead to issues like this one.

Wondering whether we should take the opportunity to make `_onKernelStatusChanged` and `_onKernelChanged` not be arrow functions, but follow the style of the other class methods instead?

That way we could also keep the `this` when connecting the signals.

hsuanxyz:
> Thanks @hsuanxyz for the fix 👍
> 
> This `KernelStatus.Model` class currently mixes the style of method definition which can indeed lead to issues like this one.
> 
> Wondering whether we should take the opportunity to make `_onKernelStatusChanged` and `_onKernelChanged` not be arrow functions, but follow the style of the other class methods instead?
> 
> That way we could also keep the `this` when connecting the signals.

OK, I've updated it, please review.

jtpio (approved):
Thanks!

jtpio:

PR #12866, opened 2022-07-27
New extension manager

fcollonval:
## References

Related to https://github.com/jupyterlab/jupyterlab/issues/11336

## Code changes

- The extension manager is fully on the backend
  - Extension manager is picked from a entry point `jupyterlab.extension_manager_v1` that must implement a callback receiving optionally `AppOptions`, `ExtensionsOption` (in dictionary format) and `LabApp` and returning and `ExtensionManager`. Two default implementations are provided:
    readonly = jupyterlab.extensions:get_readonly_manager
    pypi = jupyterlab.extensions:get_pypi_manager
  - Allowed and blocked list feature is implemented in the abstract `ExtensionManager` but third-party may override it.
  - Listing installed extensions, enabling and disabling extensions is implemented in the abstract `ExtensionManager` (but could be overridden)
  - `ExtensionManager` inherits from `traitlets.config.LoggingConfigurable` to be configurable. It receives the config from `LabApp` by passing it as `parent` of the configurable.
  - Abstract methods that must be implemented are: 
        - `metadata`: Returns extension manager metadata that will be saved in `PageConfig['extension_manager']` to tune the extension manager panel
          - `name`: Name of the extension manager
          - `can_install`: whether the extension manager can (un-)install extensions
          - `install_path`: Extensions installation path; e.g. Python virtual environment
        - `get_latest_version`: Get the latest version of an extension
        - `list_packages`: List packages for a given query
        - `install`: Install a package
        - `uninstall`: Uninstall a package
   - Actions _install_, _uninstall_, _enable_ and _disable_ return a json object with a `status` (ok, warning, error), optional `message` and a list of follow-ups restart actions (server, kernels, frontend).
- The frontend 
  - presents a generic listing of extensions installed or resulting from a search query
  - Use the generic `SidePanel` widget with `AccordionPanel` for UI homogeneity

- PyPI manager
  - Get the list of extensions from the classifier _Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt_
  - No check is performed to ensure the extension is compatible with the current JupyterLab version
  - No download statistics can be retrieve from PyPI directly. So the extensions are sorted in alphabetical order.
  - The name of the Python package for an extension is extracted from the `install.json` file. If it is missing, it _normalized_ the npm package name. But there is no guarantee that it will be the pip package name => some packages may fail to get uninstall.
  - To know if a server or kernels restart are required, the `jupyterlab.discovery` metadata in package.json are used. They may not be accurate.
  - PyPI manager configurable; the base URL, the timeouts

- Read-only manager:
  - Allows users to enable/disable installed extensions
  - Does not support un-/installing extensions.

> Upgrading @lumino/widgets in staging was required for the splice-source test to succeed.

## User-facing changes

By default they can installed or uninstalled pre-built extensions via pip.

The two default managers:

| PyPI | Read-only |
| --- | --- |
| Screenshot from 2022-08-25 12-09-58 | Screenshot from 2022-08-25 12-08-26 |

Break the api as most features are now in the backend.


fcollonval:
CI failure requires https://github.com/jupyterlab/lumino/pull/321 to be reviewed, merged and published. But as this brings lots of changes reviews are welcomed.

agoose77:
@fcollonval this looks really cool. Thanks for all of the hard work! 

Thinking about the `ExtensionManager` class, is there a reason not to have it inherit from `traitlets.Configurable`? Although third-party extension managers might not need this, for the default ones might it be useful down the road if we add configuration e.g. for `BASE_URL`?

fcollonval:
> Thinking about the `ExtensionManager` class, is there a reason not to have it inherit from `traitlets.Configurable`? Although third-party extension managers might not need this, for the default ones might it be useful down the road if we add configuration e.g. for `BASE_URL`?

I mimic what was done in [the pull-requests](https://github.com/jupyterlab/pull-requests/blob/master/jupyterlab_pullrequests/managers/manager.py) extension - dropping the configuration for now. But yes there is surely high value of adding configurable capability.
Do you know if a Configurable can also inherit from abstract class? I have the feeling inheriting from Configurable will require dropping inheriting from abstract class.

agoose77:
> Do you know if a Configurable can also inherit from abstract class?

I don't believe so - both implementations leverage a metaclass (which probably could be chained, but it's messy).

If you want to keep this PR constrained, given that it doesn't add a configuration mechanism it's probably OK to hold off! :)

jtpio:
> CI failure requires [jupyterlab/lumino#321](https://github.com/jupyterlab/lumino/pull/321) to be reviewed, merged and published.

Looks like https://github.com/jupyterlab/lumino/pull/321 has now been merged. Since it is on `main` which also includes other major changes for Lumino 2, do we want to wait for Lumino 2 to be released, or can this be backported to the [`1.x` branch](https://github.com/jupyterlab/lumino/tree/1.x) and a new release made from the `1.x` branch?

fcollonval:
@agoose77 `ExtensionsManager` is now a `LoggingConfigurable`.

fcollonval:
bot please update snapshots

fcollonval:
Kicking CI

docs/source/extension/extension_migration.rst
```diff
@@ -60,6 +60,11 @@ bumped their major version (following semver convention). We want to point out p
    The search provider API has been fully reworked. But the logic is similar, for new type of documents
    you will need to register a ``ISearchProviderFactory`` to the ``ISearchProviderRegistry``. The
    factory will build a ``ISearchProvider`` for the document widget.
+- ``@jupyterlab/extensionmanager`` from 3.x to 4.x
+   The frontend API has been drastically reduce to fetch all information from the backend. And it is now advised
```
jtpio, line 64:
```suggestion
   The frontend API has been drastically reduced to fetch all information from the backend. And it is now advised
```

docs/source/extension/extension_migration.rst
```diff
@@ -60,6 +60,11 @@ bumped their major version (following semver convention). We want to point out p
...
+   to provide a custom ``ExtensionsManager`` class for your need rather than overriding the frontend plugins.
+   See ``jupyterlab/extensions/pypi.py`` for an example using PyPI.org and pip. You can then register your manager
+   defining an entry point in the Python package defining it; see ``pyproject.toml::project.entry-points."jupyterlab.extension_manager_v1"``.
```
jtpio, line 67: Maybe something like the following to avoid repeating "defining"?

```suggestion
   by defining an entry point in the Python package; see ``pyproject.toml::project.entry-points."jupyterlab.extension_manager_v1"``.
```

docs/source/user/extensions.rst
```diff
@@ -152,13 +152,22 @@ Managing Extensions Using the Extension Manager
...
 .. _extension_manager:
 
-You can use the Extension Manager in JupyterLab to manage extensions.
+You can use the Extension Manager in JupyterLab to manage extensions. That feature
+may have been disabled by your system administrator for security reasons.
 
-.. warning::
+.. note::
+
+   Since JupyterLab v4, the default manager use PyPI.org as source for the available
```
jtpio, line 160:
```suggestion
   Since JupyterLab v4, the default manager uses PyPI.org as source for the available
```

docs/source/user/extensions.rst
```diff
@@ -152,13 +152,22 @@ Managing Extensions Using the Extension Manager
...
+   extensions and ``pip`` to install them.
 
-   Since JupyterLab v4, the core manager does not handle searching, installing or
-   uninstalling extensions. This is due to the multiple managers available to
-   distribute extensions.
+   An extension will be listed if the Python package has the classifier:
+   `Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt <https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Prebuilt>`__
+   Moreover it is advised for extension author to provide the ``install.json`` file
+   to specified the exact Python package name.
```
jtpio, line 166:
```suggestion
   to specify the exact Python package name.
```

docs/source/user/extensions.rst
```diff
@@ -152,13 +152,22 @@ Managing Extensions Using the Extension Manager
...
+
+.. warning::
+   There are no verification that the extension is compatible with the current JupyterLab
```
jtpio, line 169:
```suggestion
   There is no check to ensure the extension is compatible with the current JupyterLab
```

jupyterlab/extensions/pypi.py
```diff
@@ -0,0 +1,408 @@
...
+        base_url + f"/{name}/{latest_version}/json",
+        headers={"Content-Type": "application/json"},
+    )
+    data = json.loads(response.body).get("info")
+
+    # Keep minimal information to limit cache size
+    return {k: data.get(k) for k in ["summary", "home_page", "package_url", "author", "license"]}
+
+
+class PyPiExtensionsManager(ExtensionsManager):
```
jtpio, line 45: Maybe this could be renamed to the following for consistency with "PyPI" and the fact we usually refer to the "extension manager":

```suggestion
class PyPIExtensionManager(ExtensionsManager):
```

jupyterlab/extensions/manager.py
```diff
@@ -0,0 +1,547 @@
...
+    Attributes:
+        cache: Extension list per page
+        last_page: Last available page result
+    """
+
+    cache: Dict[int, Optional[Dict[str, ExtensionPackage]]] = field(default_factory=dict)
+    last_page: int = 1
+
+
+class ExtensionsManager(LoggingConfigurable):
```
jtpio, line 137: Maybe this class could be renamed to `ExtensionManager` without the "s", since we usually refer to the "extension manager"?

jtpio (commented):
Thanks @fcollonval for working on this!

I left a couple of comments after a (partial) first look at the diff.

fcollonval:

fcollonval:

docs/source/extension/extension_migration.rst
```diff
@@ -60,6 +60,11 @@ bumped their major version (following semver convention). We want to point out p
    The search provider API has been fully reworked. But the logic is similar, for new type of documents
    you will need to register a ``ISearchProviderFactory`` to the ``ISearchProviderRegistry``. The
    factory will build a ``ISearchProvider`` for the document widget.
+- ``@jupyterlab/extensionmanager`` from 3.x to 4.x
+   The frontend API has been drastically reduced to fetch all information from the backend. And it is now advised
```
agoose77, line 64:
```suggestion
   The frontend API has been drastically reduced to fetch all information from the backend. It is now advised
```

docs/source/extension/extension_migration.rst
```diff
@@ -60,6 +60,11 @@ bumped their major version (following semver convention). We want to point out p
...
+   to provide a custom ``ExtensionManager`` class for your need rather than overriding the frontend plugins.
```
agoose77, line 65:
```suggestion
   to implement a custom ``ExtensionManager`` class for your needs rather than overriding the frontend plugins.
```

docs/source/user/extensions.rst
```diff
@@ -249,10 +254,7 @@ it by clicking the "Install" button of the extension list entry.
     avoid installing extensions you do not trust, and watch out for
     any extensions trying to masquerade as a trusted extension.
 
-
-A short while after starting the install of an extension, a drop-down should
-appear under the search bar indicating that the extension has been
-downloaded. The newly installed extension may require to restart JupyterLab.
+The newly installed extension may require to restart JupyterLab.
```
agoose77, line 257:
```suggestion
The newly installed extension may require JupyterLab to be restarted.
```

agoose77 (commented):
This is super cool @fcollonval! I looked over the RST and added a couple of comments, but the meaning is still clear without these changes, so feel free to ignore them.

fcollonval:
Thanks @jtpio and @agoose77 for the reviews

fcollonval:
CI failure are not related

fcollonval:
Comment from JLab meeting:

> Should we show the package manager name and installation path (environment) in the user interface? Users might get a breaking envionment change after installation, so we should be explicit about what action will be taken. 

_from @krassowski_

krassowski:
We also discussed whether we could bring back the "trust" indicator for extensions from Project Jupyter (currently seen as removed):

Screenshot from 2022-08-24 17-54-01

and about ranking extensions by number of users etc.

fcollonval:
@krassowski the latest commit adds support for custom displayed name and installation path. I also added a new attribute `approved` on package that if `true` will display the jupyter icon. But for now nothing is approved as we likely need to set up our own packages safe list - a good follow-up.

fcollonval:
bot please update snapshots

packages/services/src/manager.ts
```diff
@@ -221,6 +221,8 @@ export namespace ServiceManager {
   interface IManagers {
     /**
      * The builder for the manager.
+     *
+     * @deprecated will be removed in v5
```
blink1073, line 225:
```suggestion
     * @deprecated will be removed in JupyterLab v5
```

blink1073:
When the "Warning" portion is expanded we can't see any of the other sections.  It took me a minute to realize that I needed to collapse it.  Perhaps it should collapse when we click "Enable"?

blink1073 (approved):
Also, I had expected the panel to be expanded when I refreshed the page after uninstalling an extension.

I read through the code itself, and other than the above suggestions, this looks good to me, thanks @fcollonval!

blink1073:
Tried it out again, looks great!

jupyterlab/extensions/pypi.py
```diff
@@ -0,0 +1,405 @@
...
+
+from jupyterlab.extensions.manager import (
+    ActionResult,
+    ExtensionManager,
+    ExtensionManagerMetadata,
+    ExtensionPackage,
+)
+
+
+@alru_cache(maxsize=500)
```
krassowski, line 33: 1. While there are 167 extensions with `FRAMEWORK :: JUPYTER :: JUPYTERLAB :: EXTENSIONS :: PREBUILT` modifier, there are 1565 packages matching `jupyterlab` on NPM. After release of the new extension manager we can expect developers to port more extensions to prebuilt. Raising maxsize to 1500 should not hurt (we can revisit later):

```suggestion
@alru_cache(maxsize=1500)
```

2. Can we make it configurable by keeping the standalone function `_fetch_package_metadata` in global scope without decorator and add a decorator dynamically in `__init__` of `PyPiExtensionManager` based on a value from configurable tratitlet? Something like:

```python
class PyPiExtensionManager(ExtensionManager):
   metadata_cache = Integer(1500, config=True, help="The cache size for package metadata")
   def __init__(self):
       super.__init()
       this._fetch_package_metadata = alru_cache(maxsize=self.metadata_cache)(_fetch_package_metadata)

**Review by krassowski — DISMISSED** [review]
I left some suggestions on URL handling for PyPI manager.

The biggest blocker for me is "Discover" section UX and the load that we will be putting on PyPI: I noticed that loading the "Discover" panel for the first time can take quite a while (10-20 seconds after each startup).
1. this is too long for patience of an average user
2. even, worse it ends up displaying hundreds of irrelevant "toy" extensions, so they waited a long time for nothing
3. it puts quite a lot of pressure on PyPI servers; they have disabled search APIs before due to too many requests; we do NOT want this to happen to JupyterLab leaving us with a broken extension manager, so let's be cautious with the number of requests we send.

Note: the problem will only get worse as the ecosystem of prebuilt extensions grows; based on the number of NPM extensions we can expect it to get 10x bigger soon (not mentioning the Notebook v7 impact).

(2) can be solved by a popularity/quality metric to sort by, or by having a list of featured extensions to show on top (like extensions from jupyterlab org and then from jlab-contrib).

(1) can be alleviated by changing the text from `Updating extensions list…` to `Updating extensions list… Type a search term in the bar above to narrow down results`

(3) and (1) can be alleviated by fetching and display the first "page" straight away and stopping there with an additional message saying `Showing first X results; Type a search term in the bar above to narrow down results, or scroll down for more` (or similar). Note: the pagination is already implemented on frontend, so I am just asking if we can make use of it for PyPI manager.

I am happy for the "Discover" section UX and PyPI load issue to be addressed in a follow-up PR if that's easier, as long as it happens before 4.0 release.

File: packages/extensionmanager/src/model.ts [hunk 13]
```diff
@@ -791,48 +424,67 @@ export class ListModel extends VDomModel {
...
+            title: trans.__('Information'),
+            body: trans.__(
+              'You will need to restart JupyterLab to apply the changes.'
+            ),
+            buttons: [Dialog.okButton({ label: trans.__('Ok') })]
+          });
+        } else {
+          const followUps: string[] = [];
+          if (reply.needs_restart.includes('frontend')) {
+            followUps.push(trans.__('refresh the web page'));
```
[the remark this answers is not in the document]
krassowski: We could either detect electron based on user agent (and customize the message), or use a neutral wording something like:

```suggestion
            followUps.push(trans.__('reload JupyterLab (refresh the web page if using a browser)'));
```

blink1073:
Thoughts on better handling of discovery:

- We could add weighting based on known github orgs
- We could use https://github.com/hugovk/pypistats to get a download weighting
- We could store the fetched stats in the JupyterLab app directory and not update them often

fcollonval:

> (1) can be alleviated by changing the text from `Updating extensions list…` to `Updating extensions list… Type a search term in the bar above to narrow down results`
> 
> (3) and (1) can be alleviated by fetching and display the first "page" straight away and stopping there with an additional message saying `Showing first X results; Type a search term in the bar above to narrow down results, or scroll down for more` (or similar). Note: the pagination is already implemented on frontend, so I am just asking if we can make use of it for PyPI manager.

Unfortunately PyPI RPC API does not support pagination. So searching with a query string does not help as we are forced to request all extensions based on classifier and then filter it.

> We could add weighting based on known github orgs
> We could use https://github.com/hugovk/pypistats to get a download weighting
> We could store the fetched stats in the JupyterLab app directory and not update them often (maybe look for new packages once a day and update existing weights once a day).

Only the later solution will be really good because if we list from the organization, we will end up will for example the _pull-request_ extension to be higher rated that the LSP one for example. I looked at pypistats when working on the PR but they have a throttling API rate. So we will end up storming their micro service and hitting a lot the rate limitation. Therefore the only good solution for me is the last one. But it requires some additional work - setting up a GitHub repo with bot action to maintain a list of package with popularity. So I decided it could be done in a second step with this first step focusing on the new API and extensibility capability.

So I propose to address the code comments from Mike in this PR. Then to iterate to build a small provider (probably statically serving a json file from raw github content is enough) that will improve responsiveness, add ranking and approved status.

krassowski:
> Unfortunately PyPI RPC API does not support pagination. So searching with a query string does not help as we are forced to request all extensions based on classifier and then filter it.

I know. What I'm suggesting is to download full list from RPC and then paginate on downloading metadata which is one request per package and we can define page size as we wish. Does it make sense?

fcollonval:
> > Unfortunately PyPI RPC API does not support pagination. So searching with a query string does not help as we are forced to request all extensions based on classifier and then filter it.
> 
> I know. What I'm suggesting is to download full list from RPC and then paginate on downloading metadata which is one request per package and we can define page size as we wish. Does it make sense?

Yes this is definitely a good idea.

jupyterlab/extensions/pypi.py
```diff
@@ -0,0 +1,405 @@
...
+
+        for name, group in groupby(filter(lambda m: query in m[0], matches), lambda e: e[0]):
+            _, latest_version = list(group)[-1]
+            data = await _fetch_package_metadata(name, latest_version, self.base_url)
+
+            normalized_name = self._normalize_name(name)
+            extensions[normalized_name] = ExtensionPackage(
+                name=normalized_name,
+                description=data.get("summary"),
+                url=data.get("home_page", data.get("package_url")),
```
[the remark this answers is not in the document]
fcollonval: I will display only one URL - targeting the homepage. It may be confusing for the user to get two links. And the homepage should be the better one to accompany a new user.

packages/extensionmanager/src/model.ts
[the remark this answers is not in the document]
fcollonval: I opened https://github.com/jupyterlab/jupyterlab-desktop/pull/492 to add an easy test in JupyterLab desktop.

For reference, the matter of detecting execution in Electron is discussed in https://github.com/electron/electron/issues/2288

fcollonval:

blink1073:
The pagination looks good, thanks!  I think we should merge and iterate once tests pass, this PR is awkward to review given the number of commits and comments.

fcollonval:

fcollonval:
Merging as CI is green (remain failure is not related)

Issue #12792, opened 2022-07-11
Add setting `trustbydefault` to `htmlviewer-extension`


oscar6echo:
As mentioned by  @jtpio in [jupyterlite issue #708](https://github.com/jupyterlite/jupyterlite/issues/708#issuecomment-1180053413).

I think is would be useful to have a new setting that allows to trust `.html` files by default upon opening them.  
The default would be `false` of course, to change nothing of the current safe behavior.  
I would be particularly useful in the context of jupyterlite, I think.

I have written a an extension to do that, but it really feels like overkill given the tiny modification: 
https://github.com/oscar6echo/jupyterlab-trust-html

What do you think ?

jtpio:
Thanks @oscar6echo for suggesting this.

Maybe an example for adding this to the `htmlviewer-extension` would be to define a setting like `flash` in the `logconsole-extension`: 

https://github.com/jupyterlab/jupyterlab/blob/2bd99333aefdc43a442ebe84b4b23d3ac01a011a/packages/logconsole-extension/schema/plugin.json#L37-L42

Which is then used here: 

https://github.com/jupyterlab/jupyterlab/blob/2bd99333aefdc43a442ebe84b4b23d3ac01a011a/packages/logconsole-extension/src/index.tsx#L285

And still default to `false` to keep the existing behavior of not trusting HTML files.

There are other examples across the code base that define a `boolean` setting like this. Happy to provide more guidance if needed.

oscar6echo:
@jtpio Thx for the pointers.  
Getting inspiration from them I created this PR: #12868.  

It works as expected if I set the `trustbydefault` default value to true.  
But I realize that I do not know how to set this param upon jupyterlab launch, neither in the context of jupyterlite, so could not test to set it.  

I do not know where/if the documentation must be modified.  

What do you recommend as next steps ?

jtpio:
Thanks @oscar6echo! Let's continue on the PR then.

PR #12868, opened 2022-07-27
Add trustbydefault param to htmlviewer-extension

oscar6echo:
## References

PR suggested by @jtpio in #12792.
 

## Code changes

+ Added property `trustbydefault` in `packages/htmlviewer-extension/schema/plugin.json`
+ Slightly changed function `addWidgetFactory()` to read and set the new param

## User-facing changes

None unless the user launches jupyterlab with `htmlviewer-extension` param `trustbydefault=true`

None that I know.


oscar6echo:
One CI error is a missing tag among: [ 'bug', 'enhancement', 'feature', 'maintenance', 'documentation' ]   
I guess it should be enhancement, but it seems I do not have the rights to add a tag.

packages/htmlviewer-extension/src/index.tsx
```diff
@@ -131,23 +133,34 @@ function activateHTMLViewer(
...
-    });
-    // Notify the application when the trust state changes so it
-    // can update any renderings of the trust command.
-    widget.trustedChanged.connect(() => {
-      app.commands.notifyCommandChanged(CommandIDs.trustHTML);
-    });
+  // Fetch settings if possible.
+  const fetchSettings = settingRegistry
+    ? settingRegistry.load(HTML_VIEWER_PLUGIN_ID)
+    : Promise.reject(
```
jtpio, line 139: Since `ISettingRegistry` is declared as `optional` at the plugin level definition, a common pattern in the JupyterLab code base (and in extensions) is to first check for `settingRegistry` before using it.

Some alternative lab-based applications might be reusing the plugin without providing a `ISettingRegistry`.

For example in the `documentsearch-extension`:

https://github.com/jupyterlab/jupyterlab/blob/3291708449f7fe008daf0ba1be63c65f3dbc2bb1/packages/documentsearch-extension/src/index.ts#L113-L130

oscar6echo: Indeed this common pattern makes sense.  
I pushed a commit to try to implement it in this PR.  
Is it ok ?

jtpio (approved):
Thanks @oscar6echo this looks good!

Just tested locally and it works fine. I left a couple of comments about the naming of the setting and variables.

packages/htmlviewer-extension/src/index.tsx
```diff
@@ -131,6 +133,26 @@ function activateHTMLViewer(
     });
   }
 
+  let trustedAtOpen = false;
+
+  if (settingRegistry) {
+    const loadSettings = settingRegistry.load(HTML_VIEWER_PLUGIN_ID);
+    const updateSettings = (settings: ISettingRegistry.ISettings): void => {
+      trustedAtOpen = settings.get('trustbydefault').composite as boolean;
```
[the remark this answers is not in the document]
oscar6echo: Yes, just pushed the suggested changes.

jtpio:
Thanks @oscar6echo!

This is a useful setting especially when working with HTML files that are known to be trusted. The fact that it is disabled by default means it doesn't affect the current behavior of not trusting the files.

Merging this in, if someone has any concern please comment here or open a new issue, thanks!

PR #12870, opened 2022-07-27
Add more ways to copy the currently opened file

### LINKED ISSUE #12099: Easier copying of the current file
### Problem

We should make it easier to copy the file you are currently working on. Today you can only do this (as far as I can tell) by finding the file in the file browser then right clicking on it then duplicate.

### Proposed Solution
Do some of:
1. Add a menu item like nbclassic (File->Make a Copy)
2. Right click on the tab and "Make a copy"
3. Add a Command Palette item to allow for this + hotkey opt-in support

peytondmurray:
## References

Closes #12099.

## Code changes

- Adds a new `docmanager:duplicate` command for duplicating a file

## User-facing changes

Adds 3 new ways to duplicate the currently opened file to the UI:
- Through the context menu that opens when right clicking on an opened tab
- Via a new entry in the `File` menu
- Using the command palette

which are the 3 ways of duplicating an opened file proposed in the original issue.

**Context menu entry:**
image

**File menu entry:**
image

**Command palette entry:**
palette

None


krassowski:
bot please update snapshots

krassowski:
Two of the screenshots got updated but should not have:
- `debugger-kernel-documentation-linux.png`
- `file-formats-nteract-vdom-documentation-linux.png`

@peytondmurray I will merge your branch with master and re-run snapshot bot to see if these were due to something which was already fixed.

peytondmurray:
:+1: Thanks @krassowski, it looks like that did the trick!

krassowski:
Well I think that, c3f6667c32af38ea95738a1ce79a487a3fbcaadd did it. Surprisingly the visual regression tests for documentation work about as well with and without that commit. It guess that it could be because of lenient retry settings: the snapshots are taken by the bot when the kernel is still busy / SVG comic has not finished loading and they pass the regression tests in one of the allowed retries; the tracked snapshots are taken after loading finished and they also pass the tests in one of the retries.

packages/docmanager-extension/schema/plugin.json
```diff
@@ -85,6 +89,11 @@
         "command": "docmanager:show-in-file-browser",
         "selector": "[data-type=\"document-title\"]",
         "rank": 23
+      },
+      {
+        "command": "docmanager:duplicate",
+        "selector": "[data-type=\"document-title\"]",
+        "rank": 24
```
fcollonval, line 96: Should we move the new item to be displayed after the rename action as on the file menu?

peytondmurray: @fcollonval That would be fine by me, though @krassowski [has noted that the File menu is getting pretty long](https://github.com/jupyterlab/jupyterlab/issues/12099#issuecomment-1207525834). What do you think of removing the option and modifying the behavior of the `Save As...` option? I'm happy to go either way here.

peytondmurray: :100: Done.

fcollonval: @peytondmurray may it be that you forgot to push a commit?

fcollonval:

peytondmurray:
bot please update snapshots

fcollonval:
Kicking CI

krassowski (approved):
LGTM

PR #12875, opened 2022-07-29
Set `Contents.ContentType` to string

trungleduc:
## References
The current `y_doc` handler of JupyterLab uses the `file_type` property of the document to create the corresponding `y_doc` document. Currently, [JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/b4540166b6afaba3454492591122ca963dbb3350/packages/services/src/contents/index.ts#L121) hard-codes the file types to `file | notebook | directory` so any new file type registered with jupyter_ydoc can not be used. 
Recently, `jupyter_server` removes this constraint by allowing [any kind of file type](https://github.com/jupyter-server/jupyter_server/pull/895). This PR does the same thing for the frontend part.

## Code changes

Set `Contents.ContentType` of `@jupyterlab/services` to `string`

N/A

### DIFF
--- a/packages/services/src/contents/index.ts
+++ b/packages/services/src/contents/index.ts
@@ -116,9 +116,11 @@ export namespace Contents {
   }
 
   /**
-   * A contents file type.
+   * A contents file type. It can be anything but `jupyter-server`
+   * has special treatment for `notebook` and `directory` types.
+   * Anything else is considered as `file` type.
    */
-  export type ContentType = 'notebook' | 'file' | 'directory';
+  export type ContentType = string;
 
   /**
    * A contents file format.


fcollonval (commented):
Can it breaks things if jupyter server does not have https://github.com/jupyter-server/jupyter_server/pull/895? If so, we should probably pin a minimal version of the server containing the PR (not yet published).

trungleduc:
The new type definition covers the old one so it does not break anything. We need to pin `jupyter_server` in the future if we implement new file types.

fcollonval:
> The new type definition covers the old one so it does not break anything. We need to pin `jupyter_server` in the future if we implement new file types.

Thanks then it a green for me.

fcollonval (approved):

afshin:
Cheers, @trungleduc!

fcollonval:

> This is required to add new shared document type in >=3.6

PR #12876, opened 2022-07-29
Bump jupyter_server_ydoc>=0.1.9

davidbrochart:
## References

Closes #12874.

## Code changes

Bump `jupyter_server_ydoc` minimum version to `0.1.9`.

## User-facing changes

Fixes Binder.

## Backwards-incompatible changes

None.

PR #12877, opened 2022-07-29
Upgraded y-codemirror.next

JohanMabille:
This PR fixes an issue in collaborative mode; prior to this change, when opening a notebook with many cells, the cursor of collaborators was jumping from one cell to another one quite fast. This was due to a backward incompatible change in CM6 (between the alpha and the official release) that y-codemirror.next hadn't taken into account yet. The last release fixes it.

### DIFF
--- a/packages/codemirror/package.json
+++ b/packages/codemirror/package.json
@@ -80,7 +80,7 @@
     "@lumino/widgets": "^1.33.0",
     "react": "^17.0.1",
     "style-mod": "~4.0.0",
-    "y-codemirror.next": "~0.3.0",
+    "y-codemirror.next": "^0.3.2",
     "y-protocols": "^1.0.5",
     "yjs": "^13.5.34"
   },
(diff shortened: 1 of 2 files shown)


davidbrochart:
I checked locally, it works.
To check with the Binder link, we first need #12876 to be merged.

SylvainCorlay:
Let's get this in!

jtpio:
FYI @JohanMabille I opened https://github.com/jupyterlab/jupyterlab/pull/12880 as an alternative, which has a much small diff for the `yarn.lock` (just what is needed).

Also cc @afshin since you seem to have forced pushed to the branch recently.

I have added both of you as committers to the commit.

jtpio:
Also over https://github.com/jupyterlab/jupyterlab/pull/12880 I keep `~` to specify the dependency instead of switching to `^` to keep the change minimal.

For reference I asked in https://github.com/jupyterlab/jupyterlab/pull/12695#pullrequestreview-1055926969 if there was reason to change that.

afshin:
Closed in favor of https://github.com/jupyterlab/jupyterlab/pull/12880
Thanks @JohanMabille @SylvainCorlay @jtpio

PR #12892, opened 2022-08-01
Fix JSON Settings Editor

### LINKED ISSUE #12891: JSON editor broken on 3.4.4
## Description

As reported in https://stackoverflow.com/questions/73191368/json-settings-editor-in-jupyterlab-not-working JSON Settings Editor stopped working. I was able to independently reproduce it on 3.4.4. It works perfectly on 3.4.3.

## Reproduce

1. Open Settings Editor
2. JSON Settings Editor
3. See nothing opens
4. See TypeError in browser console:

```
Uncaught (in promise) TypeError: setting getter-only property "handleMoved"
```

it is raised from transplied:

https://github.com/jupyterlab/jupyterlab/blob/f5ecbf1c0c3209892f4bf6ede4b4c8bc15d8b9e1/packages/settingeditor/src/splitpanel.ts#L13-L17

which at runtime ends up being:

```js
class M extends s.SplitPanel {
    constructor() {
        super(...arguments);
        this.handleMoved = new i.Signal(this)
    }
    handleEvent(e) {
        super.handleEvent(e);
        if (e.type === "mouseup") {
            this.handleMoved.emit(undefined)
        }
    }
}
```

and likely a consequence of merging https://github.com/jupyterlab/lumino/pull/301 which had an accompanying PR on the development branch (https://github.com/jupyterlab/jupyterlab/pull/12593) but not a compatibility fix on 3.4 branch. It was not immediately needed as the `@lumino/widgets` version was not bumped on 3.4.x up until https://github.com/jupyterlab/jupyterlab/pull/12777 which was meant to pull in memory leak fixes. 

## Expected behavior

JSON Settings Editor opens.

## Context

- JupyterLab version: 3.4.4

krassowski:
## References

Fixes #12891. Backward-compatible version of #12593 which was not backported.

## Code changes

Re-export the new `SplitPanel` from `@lumino/widgets` 1.32.0+ from the same file it was previously defined in.

## User-facing changes

JSON settings editor opens.

If downstream extension authors were overriding `handleMoved` via setter they will get an error, but they are getting this error in 3.4.4 right now anyways so I would say "none" for practical purposes.

### DIFF
--- a/packages/settingeditor/src/splitpanel.ts
+++ b/packages/settingeditor/src/splitpanel.ts
@@ -3,24 +3,10 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
-import { ISignal, Signal } from '@lumino/signaling';
-import { SplitPanel as SPanel } from '@lumino/widgets';
+import { SplitPanel } from '@lumino/widgets';
 
 /**
- * A deprecated split panel that will be removed when the phosphor split panel
- * supports a handle moved signal. See https://github.com/phosphorjs/phosphor/issues/297.
+ * @deprecated alias for `SplitPanel` from `@lumino/widgets`, will be removed in JupyterLab 4.0.
+ * Please use `import { SplitPanel } from '@lumino/widgets';` instead.
  */
-export class SplitPanel extends SPanel {
-  /**
-   * Emits when the split handle has moved.
-   */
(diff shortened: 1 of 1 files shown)


jtpio:
> If downstream extension authors were overriding `handleMoved` via setter they will get an error, but they are getting this error in 3.4.4 right now anyways so I would say "none" for practical purposes.

Yes that makes sense :+1:

jtpio (approved):
Thanks!

PR #12909, opened 2022-08-04
Log launcher error to console

trungleduc:
## Reference

If a launcher fails to start, the dialog shows only the error message, not the full stack. It will be easier to debug if the error stack is logged to the console.

## Code changes

Log the error of the command of launchers to the console.

## User-facing changes

- Before 

launcher-before

- After

launcher-after

N/A


jtpio (approved):
Thanks!

jtpio:
Looking at the screencast, maybe the message in the dialog could also be more user friendly to those not familiar with JavaScript errors.

fcollonval:

PR #12910, opened 2022-08-04
Remove @lumino/coreutils dependency from @jupyterlab/buildutils

afshin:
Removes the `@lumino/coreutils` dependency from `@jupyterlab/buildutils`. This is desirable because Lumino relies on `@jupyterlab/buildutils` in its infrastructure but `buildutils` relies on Lumino. This PR removes that circular dependency and uses native and `node` functionality instead.

### DIFF
--- a/buildutils/src/utils.ts
+++ b/buildutils/src/utils.ts
@@ -10,7 +10,8 @@ import fs from 'fs-extra';
 import childProcess from 'child_process';
 import { DepGraph } from 'dependency-graph';
 import sortPackageJson from 'sort-package-json';
-import { JSONExt, JSONObject } from '@lumino/coreutils';
+
+const assert = require('assert');
 
 type Dict<T> = { [key: string]: T };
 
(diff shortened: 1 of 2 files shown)


buildutils/src/utils.ts
```diff
@@ -123,11 +127,15 @@ export function writeJSONFile(filePath: string, data: JSONObject): boolean {
   } catch (e) {
     // no-op
   }
-  if (!JSONExt.deepEqual(data, orig)) {
+
+  // If values do not match, overwrite file.
+  try {
+    assert.deepEqual(data, orig);
```
jasongrout, line 133: This seems to now be deprecated in favor of assert.deepStrictEqual: https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message

jasongrout (approved):

buildutils/src/utils.ts
afshin: I am using strict deep equal by importing it like this:
```typescript
const assert = require('assert').strict;
```

jasongrout: Yes, I know you're using it. My point is that the name is an alias and deprecated, according to the docs. Minor point, I know, but thought I would point it out.

afshin: I misunderstood the `.strict` import to mean it was not deprecated. I've updated to use `.deepStrictEquals`

jtpio:
Thanks!

Issue #12942, opened 2022-08-12
Debugger variable explorer columns not resizable


firai:
## Description

In the table view of the debugger variable explorer panel, hovering over a border between columns changes the cursor to the resize cursor, but dragging the mouse doesn't resize the columns, even though the PR incorporating Lumino DataGrids into the debugger shows the columns to be resizable (https://github.com/jupyterlab/debugger/pull/391). It seems like the debugger is still using Lumino DataGrids (as merged in that PR), so I’m not sure if there’s a problem with JupyterLab ~~, Lumino or the browser~~.

Edit: Resizing of DataGrid rows and columns in the Lumino sample Binders at https://mybinder.org/v2/gh/jupyterlab/lumino/main?urlpath=lab/tree/examples works, so the bug seems to be in the JupyterLab debugger.

## Reproduce

1. Open the debugger sidebar.
2. Change the variable subpanel to the table view.
3. Hover over a border between columns in the variable explorer subpanel.
4. Click and drag the mouse.

## Expected behavior

Dragging borders between columns in the variable explorer resizes them.

## Context

- Operating System and version: Windows 11
- Browser and version: Chrome (104.0.5112.81) and Firefox (103.0.2)
- JupyterLab version: 3.4.5 (I understand that @krassowski has reproduced this in the latest dev version as of posting per the Discourse link below.)

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
        C:\SIMULIA\Commands
        C:\Windows\system32
        C:\Windows
        C:\Windows\System32\Wbem
        C:\Windows\System32\WindowsPowerShell\v1.0\
        C:\Windows\System32\OpenSSH\
        C:\Program Files\NVIDIA Corporation\NVIDIA NvDLISR
        C:\Program Files\Microsoft SQL Server\120\Tools\Binn\
        C:\Program Files\Microsoft VS Code\bin
        C:\Program Files\GTK3-Runtime Win64\bin
        C:\Program Files\dotnet\
        C:\WINDOWS\system32
        C:\WINDOWS
        C:\WINDOWS\System32\Wbem
        C:\WINDOWS\System32\WindowsPowerShell\v1.0\
        C:\WINDOWS\System32\OpenSSH\
        C:\Program Files\Pandoc\
        C:\Program Files\MATLAB\MATLAB Compiler Runtime\v84\runtime\win64
        C:\Program Files\MATLAB\MATLAB Runtime\v93\runtime\win64
[...]

PR #12943, opened 2022-08-13
Fix resizing and selection of debugger variable explorer grid

firai:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12942.

## Code changes

Propagate debugger variable explorer grid `onMouseDown` event from the custom handler, which does not handle grid resize and selection, to the `@lumino/datagrid` `BasicMouseHandler` default `onMouseDown` (https://github.com/jupyterlab/lumino/blob/f7b9915cb74e6836adaa86696446a9c1cac27208/packages/datagrid/src/basicmousehandler.ts#L129-L410), so that the latter can handle resize and selection requests.

## User-facing changes

Debugger variable explorer table view columns and rows become resizable, and cells/rows/columns become selectable.

None

### DIFF
--- a/packages/debugger/src/panels/variables/grid.ts
+++ b/packages/debugger/src/panels/variables/grid.ts
@@ -508,23 +508,6 @@ namespace Private {
     }
 
     /**
-     * Handle the mouse down event for the data grid.
-     *
-     * @param grid - The data grid of interest.
-     *
-     * @param event - The mouse down event of interest.
-     */
-    onMouseDown(grid: DataGrid, event: MouseEvent): void {
-      // Unpack the event.
-      let { clientX, clientY } = event;
-
-      // Hit test the grid.
-      let hit = grid.hitTest(clientX, clientY);
-
-      this._selected.emit(hit);
-    }
-
-    /**
      * Handle the context menu event for the data grid.
      *
      * @param grid - The data grid of interest.
--- a/packages/debugger/src/panels/variables/grid.ts
+++ b/packages/debugger/src/panels/variables/grid.ts
@@ -508,6 +508,26 @@ namespace Private {
     }
 
     /**
+     * Handle the mouse down event for the data grid.
+     *
+     * @param grid - The data grid of interest.
+     *
+     * @param event - The mouse down event of interest.
+     */
+    onMouseDown(grid: DataGrid, event: MouseEvent): void {
+      // Unpack the event.
+      let { clientX, clientY } = event;
+
+      // Hit test the grid.
+      let hit = grid.hitTest(clientX, clientY);
+
+      this._selected.emit(hit);
+
+      // Propagate event to Lumino DataGrid BasicMouseHandler.
+      super.onMouseDown(grid, event);
+    }
+
+    /**
      * Handle the context menu event for the data grid.
      *
      * @param grid - The data grid of interest.


firai:
Seems like the link test is failing because the link on this line is unreachable, unrelated to this PR:
https://github.com/jupyterlab/jupyterlab/blame/master/docs/source/extension/internationalization.rst#L58
It's a bit unusual, but the whole LaBRI domain seems to have been unreachable for several days now. Not sure if it's going to persist and whether a replacement intro article to `gettext` should be sought.

Not sure what happened with the `js-testutils` test. Seems to be a network issue, and maybe the test just needs to be restarted?

krassowski (changes requested):
Thank you for diagnosing the problem and a quick fix! Looking at the code path, the selection update that happens `onMouseDown` is still needed. Here is why: the `selected` signal updates the `selectedVariable` state of the model:

https://github.com/jupyterlab/jupyterlab/blob/451a00265893eb70ed723e18e22321f2161ec444/packages/debugger/src/panels/variables/grid.ts#L143-L151

And subsequently, this is used when the user issues a command to `Inspect variable`:

https://github.com/jupyterlab/jupyterlab/blob/8fb10e4855da0ba30eac7d978f2fb88a9613f77b/packages/debugger-extension/src/index.ts#L340-L359

While the variable to inspect might be passed directly to the command, there are times when it will not be the case, for example when invoked from command palette or via shortcut; these paths might be the only way to access some functions for users of assistive technologies.

I think that instead of removing the event handler, we need to propagate the event back to the lumino's `BasicMouseHandler` (via `super.onMouseDown(event)`).

firai:
Thanks for the review. I can keep the current emitter and invoke `super.onMouseDown(event)` as suggested. I suppose there isn’t a way to avoid hit testing twice.

In the longer term, should the signal emission be moved upstream instead? I'm not familiar with how TS inheritance works, but I would imagine that it's possible to have `BasicMouseHandler` functions emit the signal and just overload the signal handler? I assume that this isn’t the only application of `datagrid` with assistive needs. For example, I understand that the `csvviewer` extension just uses the `BasicMouseHandler`, so it seems like it may not be able to feed the selected data to assistive services based on your description.

krassowski:
> In the longer term, should the signal emission be moved upstream instead? [...]  I assume that this isn’t the only application of datagrid with assistive needs

Yes, you are right.

> For example, I understand that the csvviewer extension just uses the BasicMouseHandler, so it seems like it may not be able to feed the selected data to assistive services based on your description?

Yes. The `datagrid` does not have perfect accessibility support yet, but we can make breaking changes in needed in lumino v2 ([soon](https://github.com/jupyterlab/lumino/issues/341)). 

> I understand that the `csvviewer` extension just uses the BasicMouseHandler, so it seems like it may not be able to feed the selected data to assistive services based on your description?

In fact it handles accessibility needs better because it does not try to re-implement selection handling as the debugger does.

The solution that I propose for debugger indeed only helps users who can still use the mouse pointer - `selectedVariable` does not gets updated/triggered with keyboard navigation in the debugger table (we just use `BasicKeyHandler`). So I agree that customization of of `BasicMouseHandler.onMouseDown` should ultimately go away.

Instead, we would need to re-wire the `selectedVariable` to be updated when [`SelectionModel.changed`](https://github.com/jupyterlab/lumino/blob/578509cae6af5567975a6ed2e6bc9d8a1927a078/packages/datagrid/src/selectionmodel.ts#L99-L104) fires up (in our context `BasicSelectionModel.changed`) and delegate handling of mouse and keyboard to lumino's data grid. This would help because the `changed` fires for both mouse *and* keyboard events.

krassowski:
> Instead, we would need to re-wire the `selectedVariable` to be updated when [`SelectionModel.changed`](https://github.com/jupyterlab/lumino/blob/578509cae6af5567975a6ed2e6bc9d8a1927a078/packages/datagrid/src/selectionmodel.ts#L99-L104) fires up (in our context `BasicSelectionModel.changed`) and delegate handling of mouse and keyboard to lumino's data grid. This would help because the `changed` fires for both mouse _and_ keyboard events.

Just to clarify, I am not requesting pursing such solution in this PR (but if you attempt it I am happy to review)  :)

firai:
I won't attempt the more elaborate rewire in this PR unless I have to. Since there's a simple fix available and since the impact of the bug/regression is significant (at least for me as a user), I propose that we implement the simple fix for now so that it can also be backported. It seems like the more complete refactor could benefit from additional discussions and coordination with the Lumino project anyway.

firai:
Just so the refactoring that you proposed doesn't get lost in the mix, I have added a new issue at #12946. Feel free to edit or let me know if I've mischaracterized anything.

fcollonval (approved):
Thanks @firai

fcollonval:

PR #12950, opened 2022-08-16
Bump lerna to 5.x

fcollonval:
## References

Bump lerna to 5.x

## Code changes

N/A

## User-facing changes

N/A

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "stylelint-prettier": "^2.0.0"
   },
   "devDependencies": {
-    "lerna": "^4.0.0",
+    "lerna": "^5.0.0",
     "typedoc": "~0.22.10",
     "yarn": "1.22.17"
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -119,7 +119,7 @@
     "stylelint-prettier": "^2.0.0"
   },
   "devDependencies": {
-    "lerna": "^4.0.0",
+    "lerna": "^5.0.0",
```
jtpio, line 122: Maybe we could set the version to `^5.4.2` here for consistency with the new version resolved in `yarn.lock`?

```suggestion
    "lerna": "^5.4.2",
```

jtpio: Actually a `5.4.3` was released a couple of hours ago and we could also update to it: https://github.com/lerna/lerna/releases/tag/v5.4.3

jtpio (approved):
Thanks!

fcollonval:
I opened #12957 to fix the regression on the documentation snapshot

jtpio:
> I opened #12957 to fix the regression on the documentation snapshot

Thanks :+1:. This is indeed not related to the change in this current PR.

PR #12968, opened 2022-08-18
Add scrolling to `debugger` variable renderer

### LINKED ISSUE #12964: Debugger MIME renderer panel does not allow scrolling for overflowing contents
## Description

When a variable with overflowing contents, such as a large DataFrame or image, is rendered from the debugger, the MIME renderer panel does not allow scrolling and does not seem to have any options analogous to the "enable scrolling for output" that's available in the notebook panels.

Animation

## Reproduce

1. Create or load a medium or large sized variable, such as a DataFrame or an image.
2. Enable the debugger and open the debugger sidebar.
3. Render the variable via the right click menu or the tree view buttons. Resize the MIME renderer panel as necessary so that the contents overflow the size of the panel.
4. Attempt the usual methods of scrolling (mouse wheel, keyboard arrows, etc.)

## Expected behavior

Scrollable panel when the contents overflow. Just to illustrate that the contents shown above do overflow and can be scrolled otherwise, I changed `overflow` of the MIME renderer panel (see below) to `auto` in the browser tools.

Animation2

## Context

- Browser and version: Chrome (104.0.5112.102) and Firefox (103.0.2)
- JupyterLab version: 3.4.5 and latest development version

## Potential solution

The panel and its child seem to directly use the `.p-Widget` and `.lm-Widget` classes, which have `overflow` set to `hidden`. Akin to what I did in the screenshot in "Expected behavior" section, the panel can be changed to a new class with `overflow: auto`, which seems to be how other scrollable panels are set up.

It seems however that the setup of the panel causes the horizontal scrollbar to be at the bottom of the overflowing content (#4208) even if this is done, so the horizontal scrollbar would be hidden until the panel is scrolled to the bottom. In addition, as shown in the screen cap in the "Expected behavior" section, it seems that the horizontal arrow keys could not be used to horizontally scroll the panel, but `Shift` + scroll wheel seems to work.

firai:
## References

Fixes #12964.

## Code changes

- Create and add `.jp-VariableRendererPanel` class to `VariableMimeRenderer.content` with `overflow: auto` to allow vertical scrolling
- Create and add `.jp-VariableRendererPanel-renderer` class to `VariableMimeRenderer.renderMime` `widget` with `overflow: auto` to allow horizontal scrolling
    - Set `height: 100%` to set horizontal scrollbar at the bottom of parent, rather than bottom of content

## User-facing changes

`debugger` variable renderer becomes scrollable.

Animation

None

### DIFF
--- a/packages/debugger/src/panels/variables/mimerenderer.ts
+++ b/packages/debugger/src/panels/variables/mimerenderer.ts
@@ -32,7 +32,6 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
       content,
       reveal: Promise.all([dataLoader, loaded.promise])
     });
-    this.addClass(MIME_RENDERER_CLASS);
     this.trans = (translator ?? nullTranslator).load('jupyterlab');
     this.dataLoader = dataLoader;
     this.renderMime = rendermime;
@@ -77,6 +76,7 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
 
         if (mimeType) {
           const widget = this.renderMime.createRenderer(mimeType);
+          widget.addClass(MIME_RENDERER_CLASS);
           const model = new MimeModel({ ...data, trusted: true });
           this._dataHash = hash;
           await widget.renderModel(model);
--- a/packages/debugger/src/panels/variables/mimerenderer.ts
+++ b/packages/debugger/src/panels/variables/mimerenderer.ts
@@ -15,7 +15,7 @@ import { Panel } from '@lumino/widgets';
 import { murmur2 } from '../../hash';
 import { IDebugger } from '../../tokens';
 
-const MIME_RENDERER_CLASS = 'jp-MimeRenderer';
+const MIME_RENDERER_CLASS = 'jp-VariableRenderer';
 
 /**
  * Debugger variable mime type renderer
--- a/packages/debugger/src/panels/variables/mimerenderer.ts
+++ b/packages/debugger/src/panels/variables/mimerenderer.ts
@@ -15,6 +15,8 @@ import { Panel } from '@lumino/widgets';
 import { murmur2 } from '../../hash';
 import { IDebugger } from '../../tokens';
 
+const MIME_RENDERER_CLASS = 'jp-MimeRenderer';
+
 /**
  * Debugger variable mime type renderer
  */
@@ -30,6 +32,7 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
       content,
       reveal: Promise.all([dataLoader, loaded.promise])
     });
+    this.addClass(MIME_RENDERER_CLASS);
     this.trans = (translator ?? nullTranslator).load('jupyterlab');
     this.dataLoader = dataLoader;
     this.renderMime = rendermime;
--- a/packages/debugger/src/panels/variables/mimerenderer.ts
+++ b/packages/debugger/src/panels/variables/mimerenderer.ts
@@ -15,7 +15,8 @@ import { Panel } from '@lumino/widgets';
 import { murmur2 } from '../../hash';
 import { IDebugger } from '../../tokens';
 
-const MIME_RENDERER_CLASS = 'jp-VariableRenderer';
+const RENDERER_PANEL_CLASS = 'jp-VariableRendererPanel';
+const RENDERER_PANEL_RENDERER_CLASS = 'jp-VariableRendererPanel-renderer';
 
 /**
  * Debugger variable mime type renderer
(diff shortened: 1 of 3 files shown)


jweill-aws:
This is looking good! Ready to merge?

packages/debugger/src/panels/variables/mimerenderer.ts
```diff
@@ -15,6 +15,8 @@ import { Panel } from '@lumino/widgets';
 import { murmur2 } from '../../hash';
 import { IDebugger } from '../../tokens';
 
+const MIME_RENDERER_CLASS = 'jp-VariableRenderer';
```
krassowski, line 18: The [pattern](https://jupyterlab.readthedocs.io/en/stable/developer/css.html#css-class-naming-conventions) elsewhere is to have something which describes the parent class for the container (`jp-NotebookPanel`) and then something which gives semantic meaning to the child widgets (` jp-NotebookPanel-toolbar`, `jp-NotebookPanel-notebook`):

```suggestion
const RENDERER_PANEL_CLASS = 'jp-VariableRendererPanel';
const RENDERER_PANEL_RENDERER_CLASS = 'jp-VariableRendererPanel-renderer';
```

IMO the _MIME_ part is an implementation detail, and I think we could (in another PR, for 4.0) rename `VariableMimeRenderer` to `VariableRendererPanel`; pinging @fcollonval who I think authored it for an opinion.

packages/debugger/src/panels/variables/mimerenderer.ts
```diff
@@ -30,6 +32,7 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
       content,
       reveal: Promise.all([dataLoader, loaded.promise])
     });
+    this.content.addClass(MIME_RENDERER_CLASS);
```
krassowski, line 35:
```suggestion
    this.content.addClass(RENDERER_PANEL_CLASS);
```

packages/debugger/src/panels/variables/mimerenderer.ts
```diff
@@ -74,6 +77,7 @@ export class VariableMimeRenderer extends MainAreaWidget<Panel> {
 
         if (mimeType) {
           const widget = this.renderMime.createRenderer(mimeType);
+          widget.addClass(MIME_RENDERER_CLASS);
```
krassowski, line 80: Having the same class for container and child can lead to a lot of funny side-effects when modifying padding/margin and be annoying to debug. Also, see the comment in CSS file.

```suggestion
          widget.addClass(RENDERER_PANEL_RENDERER_CLASS);
```

packages/debugger/style/variables.css
```diff
@@ -114,6 +114,10 @@
   color: var(--jp-content-font-color0);
 }
 
+.jp-VariableRenderer {
+  overflow: auto;
+}
+
```
krassowski, line 120:
```suggestion
.jp-VariableRendererPanel {
  overflow: auto;
}

.jp-VariableRendererPanel-renderer {
  overflow: auto;
  height: 100%
}

```

Without adding height 100% to the child widget, the scrollbar gets positioned in the middle of the panel:

Screenshot from 2022-08-19 14-10-59

with:

Screenshot from 2022-08-19 14-11-03

krassowski (changes requested):
Some minor improvements on naming, scrollbar position and separating class of parent and child.

packages/debugger/style/variables.css
fcollonval: I was looking at the same time to this PR. Simply using `visible` for the variable view seems enough (at least on Firefox).

```suggestion
.jp-VariableRendererPanel {
  overflow: auto;
}
.jp-VariableRendererPanel-renderer {
  overflow: visible;
}

```

firai:
And semicolon applied

packages/debugger/style/variables.css
```diff
@@ -114,6 +114,15 @@
   color: var(--jp-content-font-color0);
 }
 
+.jp-VariableRendererPanel {
+  overflow: auto;
+}
+
+.jp-VariableRendererPanel-renderer {
+  overflow: auto;
+  height: 100%;
```
krassowski, line 123: @firai just highlighting the suggestion which you might have missed from https://github.com/jupyterlab/jupyterlab/pull/12968#discussion_r950181192. I can confirm that `overflow: visible` works well on Chromium too, so I think we can go that way for simplicity.

```suggestion
  overflow: visible;
```

krassowski: Regardless of whether you prefer overflow auto + height 100% or overflow visible, please update the top-level comment to reflect the changes applied after review before we merge :)

firai: Thanks for the suggestion, @fcollonval, and thanks for pointing this out, @krassowski. I did miss the additional comment. `auto` + `100%` height might be easier to understand in the future; this particular `visible` behavior seems non-obvious? What do you all think?

Top level info has been updated, though the screenshot is from a local test build. Binder has been giving me server-spawning errors in the last few hours for some reason.

krassowski (approved):
Thanks!

fcollonval:

Issue #12973, opened 2022-08-19
Emacs in terminal is broken in JL > 3.4.3


athornton:
## Description

In Jupyterlab 3.4.3, you could (as expected) say `emacs foo.txt` on the command line, and Emacs would indeed open the file foo.txt.

As of 3.4.4 (and continuing into 3.4.5), emacs startup (even if done with `-q` to bypass startup files) behaves as if Control-G were inserted almost immediately: no file is opened, and the `*Messages*` buffer has a `Quit` in it right after the invocation.

I suspect this is tied to the xterm.js update in 3.4.4, and that suggests that the fix is something we'd need to take upstream to them.

## Reproduce

So, here's the thing.  I've only (so far) seen this in the Rubin Science Platform.  If no one else can reproduce this, then this is clearly not the real issue I'm having.

1. Install jupyterlab 3.4.4 or later
2. Ensure that emacs is available in the Terminal environment
3. use `emacs <file>` to open an existing file.
4. You get `Quit` on the bottom and no file opened, rather than emacs opening the file and putting you in that buffer.

## Expected behavior

The file would open in emacs.

## Context

- Operating System and version: lsstsqre/sciplat-lab container (w_2022_30 and later exhibit this behavior)
- JupyterLab version: > 3.4.3

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

athornton:
Ah.  Looks like this: https://github.com/xtermjs/xterm.js/issues/3700

athornton:
So I think upgrading the JL dependency to xterm 4.19 should do the trick.

athornton:
https://github.com/jupyterlab/jupyterlab/pull/12974
https://github.com/jupyterlab/jupyterlab/pull/12975

afshin:
Thanks for the background research and finding a fix, @athornton!

I can confirm this behavior exists in JupyterLab 3.4.5 and **does not exist** in 3.4.0.

athornton:
I can confirm that updating to xtermjs 4.19 fixes the Emacs behavior.

[...]

PR #12974, opened 2022-08-19
update xterm.js dependency

athornton:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12973

(addresses xterm.js bug fixed in 4.19.0: https://github.com/xtermjs/xterm.js/issues/3700)

Needs backport to 3.4.x as well.

## Code changes

Use 4.19.0 of xterm.js

## User-facing changes

Hopefully, emacs in a terminal will work again.

PR #12975, opened 2022-08-19
Update xterm.js dependency

athornton:
## References

https://github.com/jupyterlab/jupyterlab/issues/12973

(addresses xterm.js bug fixed in 4.19.0: https://github.com/xtermjs/xterm.js/issues/3700)

## Code changes

Use 4.19.0 of xterm.js

## User-facing changes

Hopefully, emacs in a terminal will work again.


fcollonval:
FYI @athornton, we have a special workflow for backporting PRs. So you can simple open a PR targeting master and mention in the description that it should be backported.

=> closing this one in favor of meeseek workflow based on #12974

PR #12981, opened 2022-08-22
Support subprotocols in kernel restart

davidbrochart:
## References

In #11841 we introduced subprotocols for kernels websocket, with a reconnection mechanism to fall back to "no subprotocol" when the server doesn't support them.

## Code changes

We should not reconnect with "no subprotocol" if the reconnection is the result of a kernel restart. In this case we should use subprotocols if the server previously supported them.

## User-facing changes

Better performances, as we will most likely use the new kernel wire protocol over websocket when restarting a kernel.

None.


fcollonval (approved):
Thanks @davidbrochart

Issue #12982, opened 2022-08-22
Update JupyterLab 4 to Lumino 2


afshin:
Lumino 2 deprecates some APIs, sheds the `@lumino/datastore` package and updates several APIs to use native types.

These changes are documented in a work-in-progress [Lumino 1 to Lumino 2 migration guide](https://lumino.readthedocs.io/en/latest/migration.html).

PR #12992, opened 2022-08-23
Update to Lumino 2

afshin:
This PR migrates the JupyterLab 4 code base to use Lumino 2.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/12982


afshin:
I think the code has all been updated. I could use some advice about CI failures.

afshin (commented):
please run benchmark

packages/application-extension/src/index.tsx
```diff
@@ -147,15 +147,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         return shell.currentWidget;
       }
 
-      const matches = toArray(shell.widgets('main')).filter(
-        widget => widget.id === node.dataset.id
-      );
-
-      if (matches.length < 1) {
-        return shell.currentWidget;
+      for (const widget of shell.widgets('main')) {
```
fcollonval, line 150: [optional] Alternatively we could use `find`.

packages/application-extension/src/index.tsx
```diff
@@ -170,15 +168,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
-            value = iterator.next();
-            if (value) {
-              tab = findTab(value, widget);
-            }
-          } while (!tab && value);
+          let item = it.next();
+          while (!tab && !item.done) {
+            tab = findTab(item.value, widget);
+            item = it.next();
+          }
```
fcollonval, line 177: Could we use `find` to ease readability?

packages/docregistry/src/registry.ts
```diff
@@ -401,34 +393,34 @@ export class DocumentRegistry implements IDisposable {
...
 
     // Construct the return list, checking to make sure the corresponding
     // model factories are registered.
     const factoryList: DocumentRegistry.WidgetFactory[] = [];
-    factories.forEach(name => {
+    for (const name of factories) {
       const factory = this._widgetFactories[name];
       if (!factory) {
-        return;
+        break;
```
fcollonval, line 417: I think we need to introspect all possible factories.

```suggestion
        continue;
```

packages/fileeditor-extension/src/commands.ts
```diff
@@ -1247,9 +1246,11 @@ export namespace Commands {
...
 
       // Get the fileType based on the mimetype to determine the icon
-      const fileType = toArray(app.docRegistry.fileTypes()).find(fileType => {
-        return mimetype ? fileType.mimeTypes.includes(mimetype) : undefined;
-      });
+      const fileType = Array.from(app.docRegistry.fileTypes()).find(
+        fileType => {
+          return mimetype ? fileType.mimeTypes.includes(mimetype) : undefined;
+        }
+      );
```
fcollonval, line 1253: Could we use `find` directly?

```suggestion
      const fileType = find(app.docRegistry.fileTypes(),
        fileType => {
          return mimetype ? fileType.mimeTypes.includes(mimetype) : undefined;
        }
      );
```

packages/notebook/test/model.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('@jupyterlab/notebook', () => {
         const cell = model.contentFactory.createCodeCell({});
         model.cells.push(cell);
         model.fromJSON(utils.DEFAULT_CONTENT);
-        expect(ArrayExt.firstIndexOf(toArray(model.cells), cell)).toBe(-1);
+        expect(ArrayExt.firstIndexOf(Array.from(model.cells), cell)).toBe(-1);
```
fcollonval, line 70: Could we replace it by `findIndex`?

```suggestion
        expect(findIndex(model.cells, c => c === cell)).toBe(-1);
```

https://github.com/jupyterlab/lumino/blob/0e652ad10d8f0e90d96aca781cc691facd28a50b/packages/algorithm/src/find.ts#L88

jupyterlab/staging/package.json
```diff
@@ -114,18 +114,18 @@
...
+    "@lumino/commands": "^2.0.0-alpha.1",
+    "@lumino/coreutils": "^2.0.0-alpha.1",
+    "@lumino/disposable": "^2.0.0-alpha.1",
+    "@lumino/domutils": "^2.0.0-alpha.1",
+    "@lumino/dragdrop": "^2.0.0-alpha.1",
+    "@lumino/messaging": "^2.0.0-alpha.1",
+    "@lumino/properties": "^2.0.0-alpha.1",
+    "@lumino/signaling": "^2.0.0-alpha.1",
+    "@lumino/virtualdom": "^2.0.0-alpha.1",
+    "@lumino/widgets": "^2.0.0-alpha.1",
```
fcollonval, line 128: You should undo those changes. They are the reason the usage and interop test are failing.

fcollonval (changes requested):
Thanks @afshin 

here follows specific point and other suggestions are in the code directly:

We should get ride of _phosphorAlias_ as the Lumino API will not match any more.

https://github.com/jupyterlab/jupyterlab/blob/c099b6b9c3a2b6fe52b0dc40b06f2738e2a76c8f/builder/src/webpack.config.base.ts#L67

buildutils/src/ensure-package.ts
```diff
@@ -146,17 +146,19 @@ export async function ensurePackage(
     });
   }
 
-  let names: string[] = Array.from(new Set(imports)).sort();
```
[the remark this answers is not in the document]
afshin: Good point, that is a mistake on my part.

fcollonval:
Searching for `p-` these matches that are relevant. In particular the first one is the reason the integration tests are failing. 

image

packages/application-extension/src/index.tsx
afshin: That's a good idea. This is a recursive function that should probably be refactored. I think I'd rather put that in a different PR, though, if you're okay with that.

afshin (commented):
please update snapshots

fcollonval:
I was looking at the remaining CI failure:
- Labeler: expected too many labels
- Linux Tests / Linux (splice_source): expected staging is not updated (and will not because otherwise it breaks usage and interop jobs)
- Integration tests:
  - Broken `isRunning` helper (need to be updated for standard iterator syntax?) https://github.com/jupyterlab/jupyterlab/blob/c099b6b9c3a2b6fe52b0dc40b06f2738e2a76c8f/galata/src/helpers/kernel.ts#L24
  - ipywidgets==7.6.5 is used to produce the welcome page. But it does not work with the new Lumino => we need to fix that
  - Some commands `isEnabled` methods have a different behavior. In particular things that concern _All something_. The one reported for now are _Save All_, _Shutdown All Kernels_, _Close All Other Tabs_ => need to check the enabled test to see what is wrong.

packages/application-extension/src/index.tsx
```diff
@@ -243,10 +236,11 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
           return;
         }
         const { id } = widget;
-        const otherWidgets = toArray(shell.widgets('main')).filter(
-          widget => widget.id !== id
-        );
-        closeWidgets(otherWidgets);
+        for (const widget of shell.widgets('main')) {
```
fcollonval, line 239: We can mark an untouched line. But if you look up at line 231, the condition is redundant :wink: 

And it seems that some of the remaining error are due to call to `next()` not followed by a test on `done`.

afshin: The condition wasn't redundant, but it *was* wrong for not checking `done`. It specifically wants to make sure there are at least **two** widgets:
```ts
// Ensure there are at least two widgets.
const iterator = shell.widgets('main');
return !iterator.next().done && !iterator.next().done;
```

afshin: I went with this so it won't look like a mistake:
```ts
// Ensure there are at least two widgets.
return some(shell.widgets('main'), (_, i) => i === 1);
```

fcollonval:
bot please update snapshots

afshin:
It looks like in the new snapshots "Save All" is enabled but in the old snapshots it was disabled. I *think* the code in this PR is actually correct, though. This is a bit confusing.

blink1073 (approved):
LGTM other than the things @fcollonval highlighted.  Thanks!

fcollonval (approved):
Thanks a lot @afshin.

> Regarding the _Save All_ feature, this PR does indeed correct something as there is a writable document opened in the snapshots.

afshin (commented):
please update snapshots

blink1073:
please update snapshots

blink1073:
Thanks @afshin!

PR #13009, opened 2022-08-26
Fix customize expected reference

fcollonval:
## References

_expected_ is not customized as seen in https://github.com/jupyterlab/benchmarks/actions/runs/2919374023

## Code changes

_expected_ label can be customized through environment variable `BENCHMARK_EXPECTED_REFERENCE`

## User-facing changes

N/A

N/A


blink1073 (approved):
LGTM!

PR #13011, opened 2022-08-26
Conditional call to waitIsReady in reload

### LINKED ISSUE #12924: Galata test tool adds unhelpful async/await conditions to page.reload method, rendering it far less useful.
## Description

The Playwright method for refreshing the browser page has been overwritten in Galata in the following file:
https://github.dev/jupyterlab/jupyterlab/blob/master/galata/src/jupyterlabpage.ts#L479

The method page.reload() now expects the Launcher to be visible, meaning any tests wanting to refresh the browser can no longer do so in any reasonable way. 

image

## Expected behaviour

I don't see any reason for the reload method to be modified. 
So the expected behaviour is that page.reload() should only reload the page. 
It should not reload the page, then check that the splash screen and the launcher are visible. 

### My workaround for anyone who may see this issue

You can get access to original playwright reload method with the following `context.pages()[0].reload();`. Just make sure to pass `context` into your tests.

fcollonval:
## References

Fix #12924

## Code changes

Add optional attribute to `JupyterLabPage.reload` to skip waiting for `waitIsReady`

## User-facing changes

N/A

N/A

### DIFF
--- a/galata/src/jupyterlabpage.ts
+++ b/galata/src/jupyterlabpage.ts
@@ -492,16 +492,26 @@ export class JupyterLabPage implements IJupyterLabPage {
      * - `'domcontentloaded'` - consider operation to be finished when the `DOMContentLoaded` event is fired.
      * - `'load'` - consider operation to be finished when the `load` event is fired.
      * - `'networkidle'` - consider operation to be finished when there are no network connections for at least `500` ms.
+     * - `'commit'` - consider operation to be finished when network response is received and the document started loading.
      */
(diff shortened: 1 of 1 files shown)


pauldruce:
Ahh thank you for this, I spent yesterday trying to get a local dev environment working, but only managed to do so this morning. 

I'll remove my PR for this, as yours seems to be more in line with the style of the reload call.

fcollonval:
Sorry for that @pauldruce 

Thanks for giving it a try.

fcollonval:

PR #13012, opened 2022-08-26
Adding default parameter to reload method in Galata 

pauldruce:
Galata has overridden the page.reload method of Playwright and as a result, has introduced assertions that are only true if the JupyterLab Launcher is visible after reload.

## References

This was by the suggestion of @fcollonval on the issue: https://github.com/jupyterlab/jupyterlab/issues/12924

## Code changes

A boolean-valued default parameter named `waitForLauncher` has been introduced, which is set to true by default. If you wish to bypass the assertions that the Launcher is visible, you can set it equal to false.

## User-facing changes

None

Should be backwards compatible, so none.

### DIFF
--- a/galata/src/jupyterlabpage.ts
+++ b/galata/src/jupyterlabpage.ts
@@ -474,34 +474,44 @@ export class JupyterLabPage implements IJupyterLabPage {
    * - the splash screen to have disappeared
    * - the launcher to be visible
    *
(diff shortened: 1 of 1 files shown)


pauldruce:
Seems @fcollonval beat me to this by a few minutes, and his approach seems better.

Issue #12951, opened 2022-08-16
Narrow the scope of JupyterLab's simple mode and refactor it


afshin:
JupyterLab has a "simple" mode that can be toggled in the main UI to switch the application from multiple-document to single-document view. This mode also changes the URL from `/lab` to `/doc` to indicate that the user really is focusing on one single document (most of the time, but non-document tabs can also be the focus, e.g. a Code Console).

Because this mode evolved over several iterations, it has some idiosyncrasies, bugs, and unexpected behaviors:

### Outstanding problems (as of `4.0.0a27`)
```
🪲 = bug | 🚀 = enhancement
```
- [ ] (🪲) The simple mode URL shows both `tree` and `doc` in the URL. This seems superfluous. We should just have `doc` in the URL because it is implicit that if a path follows it should be loaded.
- [ ] (🪲) The file name *disappears* from the URL once the document has finished loading.
- [ ] (🚀) The `doc`/simple mode should *not* be attached to a workspace; @mlucool lays out a compelling case for [some aspects that go wrong with simple mode loading and writing to a workspace in this issue](https://github.com/jupyterlab/jupyterlab/issues/12644#issuecomment-1172884607).
- [ ] (🪲) When a simple mode URL (i.e. when the URL contains `doc` instead of `lab`) is loaded, the path that is loaded opens with the *default* viewer even if the user had originally loaded a secondary view of the document:
  - Right-click an `.html` file in the file browser and in the "Open with" sub-menu, select "Editor"
  - Switch to simple mode
  - Refresh the page

Here is a screen recording of the default viewer bug:

https://user-images.githubusercontent.com/159529/184880634-c0260627-0b9a-4934-bf70-b1f8f92fca7e.mov

### Proposal
Let's simplify the scope of simple mode by making it a state-less view on a document (or a functional tab like a Code Console).
1. When a user is in the multi-document mode of JupyterLab and they toggle the "simple" toggle, we can likely leave everything as is in terms of user experience.
2. When a user loads a `doc` URL, we should *not* restore the layout of a multi-document workspace.
3. We should remove `workspaces/foo` from the `doc` URLs and allow `doc` URLs to be orthogonal and independent of workspaces.
4. User actions in `doc`/simple mode should not *write* to any workspace.

### Some relevant threads
- cf. https://github.com/jupyterlab/jupyterlab/issues/12644
- cf. https://github.com/jupyterlab/jupyterlab/issues/9869

Issue #12644, opened 2022-06-02
Workspace loaded when opening a simple view link directly


mlucool:
## Description

If you open a file in simpleview with a direct link, behind the active notebook, the rest of the workspace loads. This creates a problem both in terms wasted performance as well as  user interaction issues. One example is users will get popups about "File Changed" for the hidden files if they are working on them in another tab in their workspace. Can we only load the file listed in tree view?

## Reproduce
https://user-images.githubusercontent.com/1813603/171521092-f1861c78-91a1-469b-85a5-88f2b68aaa5e.mov

Per above, I use the `tree` link to load a file in simple view, but then when you uncheck simple, you can see the rest of the workspace did load.

## Expected behavior

Only the file you navigated to loaded.

## Context

Lab@3.4.2

jweill-aws:
This is the expected behavior. When opening JupyterLab, even in Simple mode, your session is restored.

mlucool:
Per above, this causes a pretty serious degradation in user experience. Can we reopen to discuss?

jweill-aws:
If you add `reset=true` to your URL, your session should be cleared of other tabs besides the active one. Is there another action that we can take to improve the experience?

mlucool:
That would work. Can loading a URL in simple mode default to that (even if we make this opt in)?

mlucool:
As a note, reset clears the workspace so opening a new tab after no longer will show the last used workspace.

afshin:
I have been thinking about this issue and I think there are several improvements we could consider and should begin to build:

---
1. **The `MainAreaWidget` should *optionally* allow its contents to be lazily loaded.** This could be accomplished by accepting a JSON serializable argument to pass into a command that generates its content pane. If JupyterLab *only* fully renders the currently active tab at initial load time this would improve both startup performance and the resources consumed by simple mode. (cc: @jasongrout, @jtpio)
2. **When the layout is restored at initial startup, JupyterLab should not render every widget if it is going to be in simple mode.** It should either short-circuit that part of layout restoration (which might require a full page refresh if the user chooses to switch back to full mode). This means that instead of simply re-hydrating all widgets in the workspace, the initialization and restoration of the application should be more nuanced. (cc: @ellisonbg, @jasongrout)
[...]

PR #13037, opened 2022-09-01
Avoids restoring widget in dock panel when first loading in 'single-document' mode

brichet:
When loading JupyterLab in *single-document* mode (`/doc` instead of `/lab` in URL), the workspace is restored. This workspace includes the restoration of all opened tab in the DockPanel, which is useless in *single-document* mode.

This PR fixes it by delaying the restoration of DockPanel widgets until the mode changes to *multiple-document* using the switch.

## References

Related to #12644 and #12951

## Code changes

If the mode is *single-document* at start, the `LayoutRestorer` does not call the `WidgetTracker.restore()` method. Instead the `WidgetTracker` stores the options to be able to restore its widgets later.
When switching to *multiple-document* mode, the shell requests the list of non restored tracker and restores them.

## User-facing changes

Not applicable

None

## Remains to fix
- currently when widgets are restored, the last restored is activated instead of keeping the one from *single-document* mode.

EDIT : fixed in https://github.com/jupyterlab/jupyterlab/pull/13037/commits/351be506e1d6aa1b6981deab7e94d47e8575e0fe

- when starting in *single-document* mode, the workspace without any widget is saved. Strangely that seems not affect the restoring if starting an other instance in 'multiple-document', even if the workspace file is emptied...

EDIT : that is expected as only the main content is emptied from the workspace file -> *"layout-restorer:data"* part. Each opened document remains in the workspace file in its own part (eg: *"notebook:test.ipynb"* or *"application-mimedocuments:plugin.json:JSON"*, so each document will be reloaded anyway when starting in 'multiple-document'. The only thing which could be lost is the order of documents in main panel and the current active widget.


brichet:
The `launcher-extension` creates an launcher widget if the main area is empty when the file browser is restored :
https://github.com/jupyterlab/jupyterlab/blob/1d19ababaa004247a5566f585f701a4537169c1b/packages/launcher-extension/src/index.ts#L118-L129
On the other hand, the document from URL is also opened after the file browser is restored :
https://github.com/jupyterlab/jupyterlab/blob/1d19ababaa004247a5566f585f701a4537169c1b/packages/filebrowser-extension/src/index.ts#L1301-L1307

At the end, even if there is a document opened from URL, a launcher remains in the main area.

https://github.com/jupyterlab/jupyterlab/pull/13037/commits/044baa9da25ce4eebbb15132b053c0b51b62ae3e removes the launcher widget if it exists when opening the document from URL.

Perhaps the `launcher-extension` could wait for a different `promise` than the `browser.restored()`, to make sure that all documents have been restored. This would avoid creating and disposing of the launcher.

fcollonval (commented):
@brichet could you please add integration tests for this?

fcollonval (commented):
@brichet would you mind rebasing and removing the unused variable please?

brichet:
Thanks @afshin

packages/filebrowser-extension/src/index.ts
```diff
@@ -1300,10 +1302,17 @@ namespace Private {
...
         await browser.model.restore(browser.id, false);
         if (paths.file) {
-          await commands.execute(CommandIDs.openPath, {
+          let widget = await commands.execute(CommandIDs.openPath, {
             path: paths.file,
             dontShowBrowser: true
           });
+          if (widget !== undefined) {
+            for (let w of app.shell.widgets('main')) {
+              if ((w as MainAreaWidget).content instanceof Launcher) {
```
afshin, line 1311: Maybe we should add a command in `launcher-extension` to close all launcher instances instead of doing this here. What do you think?

brichet: Yes, it would be better than the current implementation.

But I think a better solution could be to not open the Launcher at all before the main area is fully restored (including the file from URL).
Currently both the Launcher extension and the document from URL are awaiting the same promise `browser.restored()`. Maybe the creation of the Launcher could move from `launcher-extension` to `filebrowser-extension`. then we could wait for `browser.restored()` -> opening file from URL -> and finally opening a launcher if there is no widget in the main area.

Let me know which solution is better, I can add it to this PR.

afshin: Hm, both of those sound like good solutions. Avoiding the creation of a launcher when it's not necessary is more efficient and performant.

brichet: In fact I think  this is now useless and this part can be removed.

It was useful when:
- the saved workspace was empty
- we start in `doc` mode with a document in URL (a `Launcher` was also created)
- then we switch to `lab` mode, the opened document remains, with the launcher

Currently the opened document is closed when we switch back to `lab` mode, so:
- if the saved workspace was empty we need a `Launcher`
- if the saved workspace was not empty it is restored without the `Launcher`

If we remove it, when starting in `doc` mode there will be 2 widgets in main area (the document and a `Launcher`), but only the document will be visible...

Do you think we can remove it ?

brichet: > Hm, both of those sound like good solutions. Avoiding the creation of a launcher when it's not necessary is more efficient and performant.

OK, thanks for your reply, I can handle it.

brichet:
I think it can be merged, failing tests look like flaky tests.

brichet:
> please update snapshots

Looks like it was updated in master branch

brichet:
Nice, the last commit from bot broke the test... this is endless

jtpio:
Was this visual change intended?

image

packages/application/src/layoutrestorer.ts
```diff
@@ -35,11 +35,29 @@ export interface ILayoutRestorer extends IRestorer {
...
   restored: Promise<void>;
 
+  /**
+   * Whether full layout restoration is deferred and is currently incomplete.
+   *
+   * #### Notes
+   * This flag is useful for tracking when the application has started in
+   * 'single-document' mode and the main area has not yet been restored.
+   */
+  isDeferred: boolean;
```
jtpio, line 45: Since these new fields are added to the interface (not optional), it looks like downstream extensions / apps will need to handle them now. How about adding them to the migration guide?

afshin: You're right. In order to make a back-port of this feature, can we make it `isDeferred?: boolean;` for the `3.x` version so that clients in that version check `restorer.isDeferred?` instead of `restorer.isDeferred` to opt-in to this functionality?

afshin:
> Was this visual change intended?
> 
> image

Hm, it is possible this fell out of this PR but I'm not sure. It wasn't an intentional outcome I don't think. It's possible this is a bug? cc: @brichet

brichet:
> > Was this visual change intended?
> > image
> 
> Hm, it is possible this fell out of this PR but I'm not sure. It wasn't an intentional outcome I don't think. It's possible this is a bug? cc: @brichet

No it wasn't. It comes from a bot commit https://github.com/jupyterlab/jupyterlab/pull/13037/commits/f1145838aae10c297eabb4ffad972b1ad607194b that added at least one other unexpected change (which I have since restored). I don't know how this can be related to this commit, I'll have a look.

packages/application/src/layoutrestorer.ts
jtpio: According to the diff should this also apply to `fetch()` and `restoreDeferred()` since they were also added?

afshin: Hm, I guess we have to stretch pretty far if we want this in 3.x -- it is possible but maybe not the most elegant if we have to add three optional items to layout restorer.

afshin: For `3.x` we could create an `IDeferrableRestorer` that is `@deprecated` and use that in the places where we want this functionality and leave `ILayoutRestorer` alone and only upgrade it for `4.x` the way we do in this PR?

jtpio:
> EDIT: In fact, this change causes the tests to fail https://github.com/jupyterlab/jupyterlab/actions/runs/3312988462 Don't understand what happened with that bot commit... I'll revert it.

This might now be stabilized by https://github.com/jupyterlab/jupyterlab/pull/13287

packages/application/src/layoutrestorer.ts
brichet: As these functionalities are only used by the shell currently, would it be a bad practice to declare `LabShell._layoutrestorer` as `LayoutRestorer` instead of `ILayoutRestorer` for 3.x (without changing the  interface)
- here https://github.com/jupyterlab/jupyterlab/blob/0082cb4ee538e8f11b39359cd41e163b9eb08ad8/packages/application/src/shell.ts#L1065
- and here https://github.com/jupyterlab/jupyterlab/blob/0082cb4ee538e8f11b39359cd41e163b9eb08ad8/packages/application/src/shell.ts#L1720

and to update the interface for the 4.x.

afshin: This is a good idea, @brichet. Perhaps this is the right move in general, instead of modifying `ILayoutRestorer`. I'll think some more about this. What do others think?

afshin: I opened a PR to do this: https://github.com/jupyterlab/jupyterlab/pull/13300

brichet: Thanks @afshin

fcollonval:

brichet:
> Owee, I'm MrMeeseeks, Look at me.
> 
> There seem to be a conflict, please backport manually. Here are approximate instructions:

I'll try to handle it.

PR #13040, opened 2022-09-02
Reorder of webpackConfig merge

### LINKED ISSUE #13032: Prepend webpack rules with webpackConfig to catch `.vue.html` files
### Problem

I'm attempting to build a jupyterlab extension that utilizes VueJS SFCs in the form of `.vue` files.  My understanding -- which, I confess freely, may not be complete or total -- is that the most straightforward way to integrate these in the extension is to have webpack manage them using `vue-loader`.  I've used the `jupyterlab.webpackConfig` key in `package.json` to supply an alternate set of rules, specifically:

```
    module: {
        rules: [
            {
                test: /\.vue$/,
                loader: 'vue-loader'
            }
        ]
    },
```

What `vue-loader` seems to do, however, is to verify that a loader works if it's able to match `.vue.html` such that the *first* rule to hit is the one that has `loader: 'vue-loader'`.  If it doesn't, we get this nice error message:

```
/home/mturk/cis/jupyterlab_nodeeditor/node_modules/vue-loader/dist/pluginWebpack5.js:89
            throw new Error(`[VueLoaderPlugin Error] No matching use for vue-loader is found.\n` +
                  ^

Error: [VueLoaderPlugin Error] No matching use for vue-loader is found.
Make sure the rule matching .vue files include vue-loader in its use.
    at VueLoaderPlugin.apply (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/vue-loader/dist/pluginWebpack5.js:89:19)
    at createCompiler (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/webpack.js:73:12)
    at /home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/webpack.js:44:48
    at Array.map (<anonymous>)
    at createMultiCompiler (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/webpack.js:44:33)
    at create (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/webpack.js:125:16)
    at webpack (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/webpack.js:158:32)
    at Object.f [as default] (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/webpack/lib/index.js:64:16)
    at Command.<anonymous> (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/@jupyterlab/builder/lib/build-labextension.js:88:39)
    at Command.listener [as _actionHandler] (/home/mturk/cis/jupyterlab_nodeeditor/node_modules/commander/index.js:422:31)
```

I believe that this is because `extensionConfig.ts` merges in the order `baseConfig`, then the custom one that includes `.html` mapping to `file-loader`, and then `webpackConfig`.  This means that the first rule to match on `.vue.html` would be the `file-loader`, which then `vue-loader` isn't pleased with.

I'm not entirely sure if my use case is unreasonable here, but it would be very nice to be able to preempt the `file-loader` rule somehow to make `vue-loader` happy.  For instance, reordering (in https://github.com/jupyterlab/jupyterlab/blob/3.4.x/builder/src/extensionConfig.ts#L253 ) succeeds at fixing this specific problem:

```
  const config = [
    merge(
      baseConfig,
      webpackConfig
      {
        mode,
        devtool,
        entry: {},
        output: {
          filename,
          path: staticPath,
          publicPath: staticUrl || 'auto'
        },
        module: {
          rules: [{ test: /\.html$/, use: 'file-loader' }]
        },
        plugins
      }
    )
  ].concat(extras);
```

but I am wary that this will break other use cases.

Would it be possible to provide a second key, or otherwise specify the order of `merge`-ing, so that the vue rule is first?  (I'm happy to submit a PR to do this, but felt asking first might work better.)

matthewturk:
The merge order of the different webpack config items currently pre-empts any
`.html`-catching rules.

## References

 - #13032

## Code changes

This changes the webpackConfig merge order to allow it to pre-empt the rule for
.html files.  Putting the `.html` rule after the inclusion of `webpackConfig`
allows extension developers to override that rule, which is helpful for
situations such as VueJS extensions that use `vue-loader`.  Previously,
`vue-loader` was testing if `.vue.html` would be caught by it.  The order of
config merging prevented the `.vue.html` from being executed.  This change
shouldn't affect any situations where `.html` was *not* being specified in the
`webpackConfig` override.

Closes #13032.

## User-facing changes

This should not have any user-facing changes, and any developer-facing changes
should only arise in the case that a `webpackConfig` rule was specified and
*not* caught by the overridden configuration, which would arguably not be
intended behavior anyway.

There should be no changes.

### DIFF
--- a/builder/src/extensionConfig.ts
+++ b/builder/src/extensionConfig.ts
@@ -262,12 +262,13 @@ function generateConfig({
           path: staticPath,
           publicPath: staticUrl || 'auto'
         },
-        module: {
-          rules: [{ test: /\.html$/, type: 'asset/resource' }]
-        },
         plugins
       },
-      webpackConfig
+      webpackConfig, {
+      module: {
+        rules: [{ test: /\.html$/, type: 'asset/resource' }]
+      }
+    }
     )
   ].concat(extras);
 
--- a/builder/src/extensionConfig.ts
+++ b/builder/src/extensionConfig.ts
@@ -264,11 +264,12 @@ function generateConfig({
         },
         plugins
       },
-      webpackConfig, {
-      module: {
-        rules: [{ test: /\.html$/, type: 'asset/resource' }]
+      webpackConfig, 
+      {
+        module: {
+          rules: [{ test: /\.html$/, type: 'asset/resource' }]
+        }
       }
-    }
     )
   ].concat(extras);
 
--- a/builder/src/extensionConfig.ts
+++ b/builder/src/extensionConfig.ts
@@ -264,7 +264,7 @@ function generateConfig({
         },
         plugins
       },
-      webpackConfig, 
+      webpackConfig,
       {
         module: {
           rules: [{ test: /\.html$/, type: 'asset/resource' }]


builder/src/extensionConfig.ts
```diff
@@ -262,12 +262,13 @@ function generateConfig({
...
-          rules: [{ test: /\.html$/, type: 'asset/resource' }]
-        },
         plugins
       },
-      webpackConfig
+      webpackConfig, {
+      module: {
+        rules: [{ test: /\.html$/, type: 'asset/resource' }]
+      }
+    }
```
fcollonval, line 271: Let's make the linter happy :wink: 

```suggestion
      webpackConfig, 
      {
        module: {
          rules: [{ test: /\.html$/, type: 'asset/resource' }]
        }
      }
```

fcollonval (changes requested):
Thanks a lot for the PR @matthewturk 

I have a suggestion to fix the linter error otherwise this is looking good.

You can add co-authors to a commit in GitHub with a special expression in the commit description: see [documentation](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). But I'm find not being a co-author.

fcollonval:
> Don't worry about the license header error.

matthewturk:
@fcollonval Thank you for the linter help!  I thought I had correctly set it up in vscode, but ... looks like I hadn't.  :)

fcollonval (approved):
Thanks @matthewturk

fcollonval:

PR #13046, opened 2022-09-02
Keep completer visible when anchor is horizontally scrolled out of view 

### LINKED ISSUE #13034: Long completion of long token on narrow screens results in completer hidden as "out of view"
## Description

When completing a long (but not short) token with long completion candidates the completer widget gets placed "out of view" which results in the geometry algorithm bailing early and completer being hidden.

This came up when investigating #12830 (thanks @mlucool).

## Reproduce

1. Create empty notebook with Python kernel
2. Paste:
    ```python
    # long completion candidates:
    very_very_very_very_very_very_very_very_very_very_very_1 = 1
    very_very_very_very_very_very_very_very_very_very_very_2 = 2
    # long token, does not work:
    very_very_very_very_very_very_very_very_very_very_very_  # press tab with cursor after `_`
    # short token, works ok:
    ve # press tab with cursor after `e`
    ```
3. Resize your browser window so that scrollbar appears (but close side panels)
4. Position cursor after long token as indicated by comment in the code
5. See that completer does not appear
6. Repeat for short token, see completer appear

After step (5) see that completer node has `jp-mod-outofview` modifier class added (which has a single CSS rule: `display: none`); remove the CSS class manually and see the completer appears as expected.

repro_completer

This behaviour comes from logic in `HoverBox.setGeometry()`:

https://github.com/jupyterlab/jupyterlab/blob/7f87d89dc88ad86e84fb2245c873139ce60dec3c/packages/ui-components/src/hoverbox.ts#L105-L111

It does not happen 100% of the time, as the placement is not recalculated after resizing.

## Expected behaviour

Completer shows up even if it is partially out of view.

## Context

- Browser and version: both Chrome and Firefox are affected
- JupyterLab version: 3.4.5

krassowski:
## References

Fixes #13034.

## Code changes

- adds optional `outOfViewDisplay` (better name suggestions welcome) to `HoverBox.IOptions` with option to show/hide the hover box when the inner/outer edge exceeds the host boundary
- changes the host of the completer hover box from editor to main area widget content
- sets `outOfViewDisplay: {left: 'stick-inside', right: 'stick-outside'}` for completer with (`bottom` and `top` set to `hidden-inside`)
- ensures that hover box does not exceed left edge of the window (when not in sticky-inside mode) - analogous to existing check for right edge.

## User-facing changes

Completer stays visible when anchor gets horizontally scrolled out of view:

completer_snap_demo

Hopefully none.

### DIFF
--- a/packages/ui-components/src/hoverbox.ts
+++ b/packages/ui-components/src/hoverbox.ts
@@ -204,7 +204,7 @@ export namespace HoverBox {
 
     // Move left to fit in the window.
     const rect = node.getBoundingClientRect();
-    const hostRect = host.getBoundingClientRect(); // Query together to avoid extra layout shift
+    const hostRect = host.getBoundingClientRect(); // Query together to avoid extra layout recalculation
     const right = rect.right;
     if (right > window.innerWidth) {
       left -= right - window.innerWidth;
--- a/packages/ui-components/src/hoverbox.ts
+++ b/packages/ui-components/src/hoverbox.ts
@@ -11,6 +11,12 @@ const HOVERBOX_CLASS = 'jp-HoverBox';
  */
 const OUTOFVIEW_CLASS = 'jp-mod-outofview';
 
+type OutOfViewDisplay =
+  | 'hidden-inside'
+  | 'hidden-outside'
+  | 'stick-inside'
+  | 'stick-outside';
+
 /**
  * A namespace for `HoverBox` members.
  */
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/hoverbox.ts
```diff
@@ -187,10 +203,170 @@ export namespace HoverBox {
     }
 
     // Move left to fit in the window.
-    const right = node.getBoundingClientRect().right;
+    const rect = node.getBoundingClientRect();
+    const hostRect = host.getBoundingClientRect(); // Query together to avoid extra layout shift
```
krassowski, line 207:
```suggestion
    const hostRect = host.getBoundingClientRect(); // Query together to avoid extra layout recalculation
```

fcollonval (approved):
Thanks @krassowski 

> Regarding the naming the only alternative I can think of is `outOfViewBehavior`

fcollonval:

Issue #12733, opened 2022-06-26
Hiding schema-derived descriptions for custom setting UI forms


krassowski:
### Problem

Authors of extensions (and JupyterLab) want to have a detailed description for more complex setting objects in the JSON settings editor. This descriptions are often auto-generated to include all necessary options. However, once a custom setting UI form is added these descriptions are no longer needed, and in fact confusing to the user of UI editor as they do not correspond to options in the UI. There are a few examples where a way to hide a description would be useful:
- JupyterLab shortcuts UI (currently hard-coded to hide description)
- jupyterlab-lsp server configuration options https://github.com/jupyter-lsp/jupyterlab-lsp/pull/778
- in the future: custom forms for edition of the menus and toolbars

Currently shortucuts UI has description hidden thanks to hard-coded override:

https://github.com/jupyterlab/jupyterlab/blob/c2907074e58725942946a73a823fc60e1795da39/packages/settingeditor/src/SettingsFormEditor.tsx#L282-L286

which was fine as a quick workaround, but does not allow extension authors to take advantage of this option.

### Proposed Solution

Currently custom form components can be registered by adding a `renderer` function:

https://github.com/jupyterlab/jupyterlab/blob/c2907074e58725942946a73a823fc60e1795da39/packages/ui-components/src/FormComponentRegistry.tsx#L13-L26

I propose to wrap the `renderer` into a component-describing interface so that we can pass additional option to custom field templates:

```ts
/**
 * Form component options
 */
export interface IFormComponent {
  /**
   * A function that takes props and returns a rendered component.
   */
  renderer: Field;
  /**
   * Whether the description derived from schema should be hidden. Default is false.
   */
  hideDescription?: boolean
}
```

In backward-compatible fashion two methods will be added:
- `addComponent(id: string, component: IFormComponent): boolean`
- `components(): { [id: string]: IFormComponent }` to match existing `renderers`

Following methods will be deprecated:
- `addRenderer(id: string, renderer: Field): boolean`
- `getRenderer(id: string): Field` - currently not used in JupytyterLab codebase

Subsequently code handling description will be added to the `CustomTemplate` (`Factory`, see  https://github.com/jupyterlab/jupyterlab/pull/12732)

### Additional context

This approach will allow to expand `IFormComponent` in the future with additional attributes and methods without introduction of breaking changes.

krassowski:
[...]

Issue #11989, opened 2022-02-02
Add uiSchema to settings schema to enable custom settings rendering


marthacryan:
Follow-up to [this comment](https://github.com/jupyterlab/jupyterlab/pull/11079#issuecomment-947838238) - adding a top-level `uiSchema` to the settings schemas which would allow a plugin to tweak their stuff without _having_ to know deep react via the [`rjsf` uischema](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/).

bollwyvl:
The irony of this: the `uiSchema` doesn't have a schema.

PR #13056, opened 2022-09-05
Create a form to read and write the metadata of Notebook's cells

brichet:
The purpose of this PR is to allow any extension to easily build a form to interact with the metadata of the Notebook cells.

It relies on [React JSON Schema Form](https://react-jsonschema-form.readthedocs.io/en/latest/) to include a form in the `NotebookTools` panel using the schema of other extensions.

To do this the external extension must embed a schema with the key `jupyter.lab.metadataforms`.

Features :
- [x] build a form from extension schema
- [x] do not write the metadata if value is default, to keep metadata as light as possible
- [x] metadata may or may not be available by cell type
- [x] custom widgets can be used in the form
- [x] [uiSchema](https://react-jsonschema-form.readthedocs.io/en/docs/api-reference/uiSchema/) from *RJSF* should be usable

TODO :
- [x] add conditions on metadata to display or not some fields depending on previous answers ?
- [x] add Notebook level metadata

## Code changes

It adds a new package in JupyterLab (*metadataform-extension*) to build forms from the extension schema, and include these forms in the `NotebookTools` panel.

It also adds a registry to allow extensions to register their own widgets.

## how to test
An example on how to test it can be found at https://github.com/jupyterlab/extension-examples/pull/208

EDIT:
## preview
image


packages/ui-components/src/FormWidgetRegistry.tsx
```diff
@@ -0,0 +1,77 @@
...
+   * @param id - The unique id for the renderer.
+   * @returns - A function that takes props and returns a rendered component.
+   */
+  getRenderer: (id: string) => Widget;
+
+  /**
+   * Returns all registered renderers in dictionary form.
+   * @returns - A dictionary that maps an id to a renderer.
+   */
+  renderers: { [id: string]: Widget };
```
krassowski, line 72: I would propose to wrap the second argument in an interface like:

```ts
interface IFormWidget {
  renderer: Widget;
}
```

so that we can add new properties without breaking compatibility in the future. We regretted not doing it for the settings form registry, please see https://github.com/jupyterlab/jupyterlab/issues/12733

packages/ui-components-extension/src/index.ts
```diff
@@ -41,4 +43,17 @@ const registryPlugin: JupyterFrontEndPlugin<IFormComponentRegistry> = {
...
+/**
+ * Sets up the widget registry to be used by a RJSF form.
+ */
+const widgetRegistryPlugin: JupyterFrontEndPlugin<IFormWidgetRegistry> = {
+  id: '@jupyterlab/ui-components-extension:widget-registry',
+  provides: IFormWidgetRegistry,
+  autoStart: true,
+  activate: (app: JupyterFrontEnd): IFormWidgetRegistry => {
+    const editorRegistry = new FormWidgetRegistry();
+    return editorRegistry;
```
krassowski, line 55:
```suggestion
    const formRegistry = new FormWidgetRegistry();
    return formRegistry;
```

krassowski (commented):
Just a quick advice on the `IFormWidgetRegistry` API from experience. I would love to see a screenshot of what the resulting form would look like to get a better feel of what this PR is attempting to do.

packages/metadataform-extension/src/form.tsx
```diff
@@ -0,0 +1,316 @@
...
+     * The default values for each key.
+     */
+    defaultValues: IDefaultValues;
+  }
+
+  /**
+   * Template to allow for custom buttons to re-order/remove entries in an array.
+   */
+  export const CustomArrayTemplateFactory =
+    (): React.FC<ArrayFieldTemplateProps> => {
```
krassowski, line 111: We probably would not want to have the custom templates duplicated here and in `packages/settingeditor/src/SettingsFormEditor.tsx` but instead I would prefer to move the common code to `ui-components` or a new package dedicated to customising rjsf (not sure on this idea though); there are bugs we need to fix and changes coming with the new rjsf version and having to keep the templates copies in sync manually will be cumbersome and error-prone.

packages/ui-components/src/FormWidgetRegistry.tsx
brichet: Agree, thanks for the suggestion

packages/metadataform-extension/src/form.tsx
bollwyvl: >  have the custom templates duplicated here

Agreed. Indeed, having a proper `ui-components/lib/schemaform` (_not_ imported by default) with base styling/etc. which multiple providers can subclass is clutch.

>  new rjsf version

We should take a hard look at `rjsf5`, as it might make custom templates far easier.

I would also _really_ like to see us support the data-driven `uiSchema` (#11989) directly, as the current "whole custom widget" approach is really unwieldy, and doesn't integrate particularly well with third-party tools.

packages/ui-components/src/FormWidgetRegistry.tsx
brichet: Changes included in https://github.com/jupyterlab/jupyterlab/pull/13056/commits/24115612f74af847ec7d706983084c939ccb81bd

bollwyvl:
A clutch performance thing: it's extremely important to _not_ draw/update these forms if the sidebar is not actually open. Indeed, when making this change, the existing CodeMirror editor should also be hidden/disabled, if not already done, as it is extremely expensive to do all the round-tripping to just JSON.

So, likely, each extension needs to be its own, collapsible form, and all collapsed by default.

Regarding the JSON editor: perhaps the right play now would be to make this new set of forms the default UI, and offer to open a _Raw Metadata JSON_ main area widget, much like the form-based settings editors will do, for getting some heavy config work done.

packages/metadataform-extension/src/form.tsx
brichet: > We should take a hard look at `rjsf5`

Agreed. At least the `v4` (current stable) to include conditional fields, look like it is not available in the `v3`.

brichet:
> it's extremely important to _not_ draw/update these forms if the sidebar is not actually open

Thanks for suggestion, done in https://github.com/jupyterlab/jupyterlab/pull/13056/commits/3137280e2d8282af60e0f83bd79e02b50f5168f3

brichet:

packages/metadataform-extension/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@jupyterlab/metadataform-extension",
+  "version": "4.0.0-alpha.14",
```
jtpio, line 3: Probably this should be bumped to be aligned with the latest release:

```suggestion
  "version": "4.0.0-alpha.15",
```

packages/metadataform-extension/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "clean": "rimraf lib && rimraf tsconfig.tsbuildinfo",
+    "watch": "tsc -w --listEmittedFiles"
+  },
+  "dependencies": {
+    "@jupyterlab/application": "^4.0.0-alpha.14",
+    "@jupyterlab/metadataform": "^4.0.0-alpha.14",
+    "@jupyterlab/notebook": "^4.0.0-alpha.14",
+    "@jupyterlab/settingregistry": "^4.0.0-alpha.14",
+    "@jupyterlab/translation": "^4.0.0-alpha.14",
+    "@jupyterlab/ui-components": "^4.0.0-alpha.29",
```
jtpio, line 46: These can be bumped as well:

```suggestion
    "@jupyterlab/application": "^4.0.0-alpha.15",
    "@jupyterlab/metadataform": "^4.0.0-alpha.15",
    "@jupyterlab/notebook": "^4.0.0-alpha.15",
    "@jupyterlab/settingregistry": "^4.0.0-alpha.15",
    "@jupyterlab/translation": "^4.0.0-alpha.15",
    "@jupyterlab/ui-components": "^4.0.0-alpha.30",
```

packages/metadataform/package.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "@jupyterlab/metadataform",
+  "version": "4.0.0-alpha.14",
```
jtpio, line 3:
```suggestion
  "version": "4.0.0-alpha.15",
```

packages/metadataform/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "watch": "tsc -w --listEmittedFiles"
+  },
+  "dependencies": {
+    "@jupyterlab/apputils": "^4.0.0-alpha.14",
+    "@jupyterlab/nbformat": "^4.0.0-alpha.14",
+    "@jupyterlab/notebook": "^4.0.0-alpha.14",
+    "@jupyterlab/observables": "^5.0.0-alpha.14",
+    "@jupyterlab/settingregistry": "^4.0.0-alpha.14",
+    "@jupyterlab/translation": "^4.0.0-alpha.14",
+    "@jupyterlab/ui-components": "^4.0.0-alpha.29",
```
jtpio, line 53:
```suggestion
    "@jupyterlab/apputils": "^4.0.0-alpha.15",
    "@jupyterlab/nbformat": "^4.0.0-alpha.15",
    "@jupyterlab/notebook": "^4.0.0-alpha.15",
    "@jupyterlab/observables": "^5.0.0-alpha.15",
    "@jupyterlab/settingregistry": "^4.0.0-alpha.15",
    "@jupyterlab/translation": "^4.0.0-alpha.15",
    "@jupyterlab/ui-components": "^4.0.0-alpha.30",
```

packages/metadataform/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "@lumino/coreutils": "^2.0.0-alpha.6",
+    "@lumino/messaging": "^2.0.0-alpha.6",
+    "@lumino/widgets": "^2.0.0-alpha.6",
+    "@rjsf/core": "^4.2.0",
+    "json-schema": "^0.4.0",
+    "react": "^17.0.1"
+  },
+  "devDependencies": {
+    "@jupyterlab/testutils": "^4.0.0-alpha.14",
```
jtpio, line 62:
```suggestion
    "@jupyterlab/testutils": "^4.0.0-alpha.15",
```

jtpio (commented):
Left a couple of inline suggestions to re-align the version numbers as a follow-up to the latest pre-release.

brichet:
Thanks @jtpio

packages/metadataform-extension/src/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  id: PLUGIN_ID,
+  autoStart: true,
+  requires: [
+    INotebookTools,
+    ITranslator,
+    IFormWidgetRegistry,
+    IFormComponentRegistry
+  ],
+  optional: [ISettingRegistry],
+  provides: IMetadataFormProvider,
```
fcollonval, line 295: I don't understand the need for the `IMetadataFormProvider` it is not consume by any other plugins.

packages/metadataform/src/form.tsx
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A ReactWidget with the form itself.
+ */
+export class FormWidget extends ReactWidget {
+  /**
+   * Constructs a new FormWidget.
+   */
+  constructor(props: MetadataForm.IProps, pluginId?: string) {
+    super();
+    this.addClass('jp-ReactWidget');
```
fcollonval, line 26: I would change the name of the class because it is misleading on the component it is applied to.

```suggestion
    this.addClass('jp-FormWidget');
```

packages/metadataform/src/form.tsx
```diff
@@ -0,0 +1,62 @@
...
+        idPrefix={`jp-MetadataForm-${this.pluginId}`}
+        onChange={(e: IChangeEvent<ReadonlyPartialJSONObject>) => {
+          this._props.metadataFormWidget.updateMetadata(e.formData);
+        }}
+      />
+    );
+  }
+
+  private _props: MetadataForm.IProps;
+  private pluginId?: PartialJSONValue;
```
fcollonval, line 60: Let's follow code practice and use a leading `_`

```suggestion
  private _pluginId?: PartialJSONValue;
```

packages/metadataform/src/metadataform.ts
```diff
@@ -0,0 +1,473 @@
...
+  /**
+   * Construct an empty widget.
+   */
+  constructor(
+    metadataSchema: MetadataForm.IMetadataSchema,
+    metaInformation: MetadataForm.IMetaInformation,
+    uiSchema: MetadataForm.IUiSchema,
+    pluginId?: string,
+    translator?: ITranslator,
+    showModified?: boolean
```
fcollonval, line 44: In the code base, we tend to use an object `options` rather than long list of arguments to be able to easily extend it and skip non-mandatory argument.

> This is a general comment, I'm fine with keeping things as is.

packages/metadataform/src/token.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Current data of the form.
+     */
+    formData: ReadonlyPartialJSONObject | null;
+
+    // /**
+    //  *  Errors on the form.
+    //  */
+    // errors?: {[keyError: string]: {__errors: string}};
```
fcollonval, line 110: Dead code

```suggestion
```

packages/notebook/src/notebooktools.ts
```diff
@@ -78,11 +78,15 @@ export class NotebookTools extends Widget implements INotebookTools {
...
+    this._commonTools.title.label = this._trans.__('Common Tools');
     this._advancedTools = new RankedPanel<NotebookTools.Tool>();
+    this._advancedTools.id = 'advancedToolsSection';
     this._advancedTools.title.label = this._trans.__('Advanced Tools');
 
+    this._extendedTools = [];
+
     const layout = (this.layout = new PanelLayout());
-    layout.addWidget(this._commonTools);
+    layout.addWidget(new Collapser({ widget: this._commonTools }));
```
fcollonval, line 89: In a follow-up, it can be interesting to switch to a `SidePanel` to get a better overall look and feel for the side bars.

fcollonval (dismissed):
I left some comments @brichet

packages/metadataform-extension/src/index.ts
brichet: The token is necessary to update schema after the application started.

For example if we aim to rebuild the *common tools* and *advanced tools* using the *metadataform*, the `RawNBConvert` would need it. The full list of export formats is known after the application started:
https://github.com/jupyterlab/jupyterlab/blob/26999c5df85dd8a088580b5dafbc5404ef9b4a44/packages/notebook-extension/src/index.ts#L934-L943

packages/metadataform/src/metadataform.ts
brichet: Right thanks, I'll change it.

brichet: Done in https://github.com/jupyterlab/jupyterlab/pull/13056/commits/77b9bd39ea421b43fbba692eed8f942124d3a695

packages/metadataform-extension/src/index.ts
fcollonval: Thanks @brichet

In that case, could you create an interface for the provided object and encapsulate the map within an class that will have only allow to add and get the items? A good example for an simple interface like that is `IRunningSessionManagers`:

https://github.com/jupyterlab/jupyterlab/blob/219200ed47ff84c3f98355f500ac92bb48792338/packages/running/src/index.tsx#L81

brichet: The relevant changes are included in https://github.com/jupyterlab/jupyterlab/pull/13056/commits/3f61a754ad43f82e869b7290c9d3e1603c08470e

jtpio:
Looks like there are a couple of conflicts to fix.

We can also publish the current versions (from this PR) of the `@jupyterlab/metadataform` and `@jupyterlab/metadataform-extension` packages to help make the CI pass.

brichet:
> We can also publish the current versions (from this PR) of the `@jupyterlab/metadataform` and `@jupyterlab/metadataform-extension` packages to help make the CI pass.

I agree, if the PR is approved. It look like the last tests are failing because of that.

packages/metadataform/src/metadataform.ts
```diff
@@ -0,0 +1,496 @@
...
+   */
+  protected onActiveCellMetadataChanged(msg: Message): void {
+    if (!this._updatingMetadata && this.isVisible) this._update();
+  }
+
+  protected onActiveNotebookPanelChanged(msg: Message): void {
+    // Do not use notebook metadata if model is null.
+    let notebook = this.notebookTools.activeNotebookPanel;
+    if (notebook === null || notebook.model === null) {
+      console.warn('Notebook model is null, its metadata cannot be updated.');
```
afshin, line 339:
```suggestion
```

afshin (approved):
I think it's a good time to merge this and iterate if there is any fall out from it. I think the suggestion to publish the newly introduced packages makes sense.

packages/metadataform/src/metadataform.ts
```diff
@@ -0,0 +1,429 @@
...
+
+  /**
+   * Set properties to a metadataKey.
+   *
+   * @param metadataKey - metadataKey (string).
+   * @param properties - the properties to add or modify.
+   */
+  setProperties(metadataKey: string, properties: PartialJSONObject): void {
+    Object.entries(properties).forEach(([key, value]) => {
+      this._metadataSchema.properties[metadataKey][key] = value;
```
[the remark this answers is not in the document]
afshin: I don't think we need to worry about this. I am not dismissing the alert so that others can weigh in, but in this circumstance, I think we're safe.

jtpio:
> I think it's a good time to merge this and iterate if there is any fall out from it. I think the suggestion to publish the newly introduced packages makes sense.

Yeah they'll have to be published before merging or the CI will keep failing on master until the next release.

brichet:
Thanks @jtpio

brichet:

fcollonval:
Kicking the CI

fcollonval:
@brichet the integration test failure is relevant. That snapshot should not be updated: galata/test/documentation/general.test.ts-snapshots/running-layout-documentation-linux.png

could you revert the change? Then we are good to go with this one.

fcollonval (approved):
Thanks @brichet

Issue #12997, opened 2022-08-23
Blurry icons in the Launcher at 400% Zoom


steff456:
## Description

I'm starting to work in the accessibility grant and did some audits for start identifying the issues in JupyterLab with 400% Zoom. One of them is that the icon for `Python 3` for both the Notebook and Console Launcher starts to blur with a lot of zoom,

**Default**

<img width="116" alt="image" src="https://user-images.githubusercontent.com/20992645/186281799-65a8a3f6-6d6b-4553-a35c-d187675a2220.png">

**200%**

<img width="221" alt="image" src="https://user-images.githubusercontent.com/20992645/186281775-4e29a21c-3af8-4504-980c-0f682b3b8721.png">

**400%**

<img width="1507" alt="image" src="https://user-images.githubusercontent.com/20992645/186281681-ee4e884a-5a71-408d-b97b-748c207e601d.png">

## Reproduce

1. Open the Launcher
2. Zoom in and check the Python 3 icon for the Notebook and Console 

## Expected behavior

I would expect to not have a blurry icon.

I want to make the contribution to fix this issue, but I was unable to find the asset that is loaded for this icon. In my investigation I discovered that apparently it is trying to load the icon from `/kernelspecs/python3/logo-64x64.png` and this one is different from the blue version that is at `packages/ui-components/style/icons/filetype/python.svg`. 

I hope you can guide me to make the PR and fix this issue! 

## Context

- Operating System and version: MacOS Catalina
- Browser and version: Chrome 104.0.5112.79 
- JupyterLab version: Latest master version

tonyfast:
@steff456 very cool you are working on the zoom problem.

> I want to make the contribution to fix this issue, but I was unable to find the asset that is loaded for this icon.

the source images for kernels are actually in the kernel providers repository. the specific images for the python logo are in the `ipython/ipykernel` repo in this folder
https://github.com/ipython/ipykernel/blob/main/ipykernel/resources/

krassowski:
This also affects other kernels notably IRkernel and Julia kernel:

![Screenshot from 2022-08-24 21-42-41](https://user-images.githubusercontent.com/5832902/186520724-3427d622-4598-4825-972c-df9d19340cc3.png)

PR #13057, opened 2022-09-05
Fix blurry icons in Launcher at 400% Zoom

steff456:
## References

This PR fixes https://github.com/jupyterlab/jupyterlab/issues/12997

## Code changes

The code change gives preference to loading the svg asset, if not found the png asset will be loaded.

Note: Please note that I made PRs in the Python, Julia and R kernels to add the svg asset and it should be loaded without having to do any change in [jupyter-server](https://github.com/jupyter-server/jupyter_server/blob/7239666c231c733cb49ed2f5c18369051c255e9a/jupyter_server/services/kernelspecs/handlers.py#L34). We may need to wait until an actual release on each of the kernels for the users to see the new asset.

## User-facing changes

Given that the new asset is `svg`, it will not get blurry regardless of zoom

This PR doesn't introduce any backwards-incompatible change

### DIFF
--- a/packages/console-extension/src/index.ts
+++ b/packages/console-extension/src/index.ts
@@ -281,12 +281,8 @@ async function activateConsole(
         for (const name in specs.kernelspecs) {
           const rank = name === specs.default ? 0 : Infinity;
           const spec = specs.kernelspecs[name]!;
-          let kernelIconUrl;
-          try {
-            kernelIconUrl = spec.resources['logo-svg'];
-          } catch (error) {
-            kernelIconUrl = spec.resources['logo-64x64'];
-          }
+          const kernelIconUrl =
+            spec.resources['logo-svg'] || spec.resources['logo-64x64'];
           disposables.add(
             launcher.add({
               command: CommandIDs.create,
--- a/packages/console-extension/src/index.ts
+++ b/packages/console-extension/src/index.ts
@@ -281,7 +281,12 @@ async function activateConsole(
         for (const name in specs.kernelspecs) {
           const rank = name === specs.default ? 0 : Infinity;
           const spec = specs.kernelspecs[name]!;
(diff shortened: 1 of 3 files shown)


packages/console-extension/src/index.ts
```diff
@@ -281,7 +281,12 @@ async function activateConsole(
         for (const name in specs.kernelspecs) {
           const rank = name === specs.default ? 0 : Infinity;
           const spec = specs.kernelspecs[name]!;
-          let kernelIconUrl = spec.resources['logo-64x64'];
+          let kernelIconUrl;
+          try {
+            kernelIconUrl = spec.resources['logo-svg'];
+          } catch (error) {
+            kernelIconUrl = spec.resources['logo-64x64'];
+          }
```
krassowski, line 289: Thank you for starting this!

1) It seems that the change would also need to be applied in `helper-extension`, and `notebook-extension`
2) the code needs to be adjusted as `spec.resources['logo-svg']` does not rise; it just returns an undefined, which leads to no fallback icon currently:
image

I would suggest:

```suggestion
          const kernelIconUrl = spec.resources['logo-svg'] || spec.resources['logo-64x64'];
```

krassowski (dismissed):

packages/console-extension/src/index.ts
krassowski: Also for reference if someone would like to add an icon for xeus-python (since we are already testing against it), it would go into: https://github.com/jupyter-xeus/xeus-python/tree/master/share/jupyter/kernels/

steff456: I'll make a PR there too :)

fcollonval (approved):
CI failures are not related

fcollonval:

PR #13062, opened 2022-09-07
Bump to Lumino 2ᵅ⁵

afshin:
Updates Lumino 2 from `alpha.1` to `alpha.5`


jtpio:
> Just restarted the UI tests check to make sure they still pass: https://github.com/jupyterlab/jupyterlab/runs/8248244462?check_suite_focus=true

Another "cancelled" after the restart, so there might be something to look into:

image

packages/shortcuts-extension/package.json
```diff
@@ -44,13 +44,13 @@
...
-    "@lumino/commands": "^2.0.0-alpha.1",
-    "@lumino/coreutils": "^2.0.0-alpha.1",
-    "@lumino/disposable": "^2.0.0-alpha.1",
-    "@lumino/domutils": "^2.0.0-alpha.1",
+    "@lumino/algorithm": "^2.0.0-alpha.4",
+    "@lumino/commands": "^2.0.0-alpha.4",
+    "@lumino/coreutils": "^2.0.0-alpha.4",
+    "@lumino/disposable": "^2.0.0-alpha.4",
+    "@lumino/domutils": "^2.0.0-alpha.4",
     "@lumino/keyboard": "^1.8.1",
```
fcollonval, line 52: It seems we missed that one in the first PR:

```suggestion
    "@lumino/keyboard": "^2.0.0-alpha.4",
```

afshin:
D'oh! Thanks @fcollonval!

fcollonval:
@afshin I'm gonna make another release of Lumino 2-alpha that include the use of `setTimeout`. So we can test it before backporting it in Lumino 1 and JLab 3.x as @krassowski mentions it may introduce some race conditions.

fcollonval:
The remaining integration failure is certainly related to this PR - the tab bar of the main area is displayed when switching to simple mode.

fcollonval:
@afshin there is a bug with the tab bars update. If you run this PR, open two tabs and then close one of them, somehow the tab bar for the closed item and the add button are not removed.

I did not find the root cause. But I would bet for virtualdom because the tab title is correctly removed from the tabbar list.

afshin:
I still haven't resolved the tab bar bug, but this is another bug that affects the current Lumino alpha, so we should not merge this PR without it:
- https://github.com/jupyterlab/lumino/pull/411

fcollonval:
@afshin PR merged, new alpha.6 release done and this PR has been updated.

fcollonval (approved):
Thanks @afshin

PR #13066, opened 2022-09-08
Added mimeType for .webp image files

### LINKED ISSUE #12961: .webp files do not open in 3.4.5
## Description

When I try to open a .webp file in JupyterLab 3.4.5, I get the following error -

image

(This is the same issue as reported in https://github.com/jupyterlab/jupyterlab/issues/12832)

The changelog for 3.4.5 mentions that the webp filetype has been added to the docRegistry; should we wait for any more changes to come in before such files can be viewed? Please advise.

## Context

- Browser and version: Chrome 103
- JupyterLab version: JupyterLab 3.4.5

alec-kr:
This commit resolves #12961, by adding support for .webp image files.

## Code changes

The ```image/webp``` mimeType was added to line 23 of ```jupyterlab/packages/rendermime/src/factories.ts``` to enable MIME support for WebP images in JupyterLab.

## User-facing changes
In issue #12961, users encountered the following error when attempting to add certain .webp files:
webp-error

This patch resolves the issue by adding support for WebP files. Hence, users can import these image files, and will no longer be greeted with the error above.

### DIFF
--- a/packages/rendermime/src/factories.ts
+++ b/packages/rendermime/src/factories.ts
@@ -21,12 +21,12 @@ export const htmlRendererFactory: IRenderMime.IRendererFactory = {
 export const imageRendererFactory: IRenderMime.IRendererFactory = {
   safe: true,
   mimeTypes: [
-  'image/bmp',
-  'image/png',
-  'image/jpeg',
-  'image/gif',
-  'image/webp'
-],
+    'image/bmp',
+    'image/png',
+    'image/jpeg',
+    'image/gif',
+    'image/webp'
+  ],
   defaultRank: 90,
   createRenderer: options => new widgets.RenderedImage(options)
 };
--- a/packages/rendermime/src/factories.ts
+++ b/packages/rendermime/src/factories.ts
@@ -20,7 +20,13 @@ export const htmlRendererFactory: IRenderMime.IRendererFactory = {
  */
 export const imageRendererFactory: IRenderMime.IRendererFactory = {
   safe: true,
-  mimeTypes: ['image/bmp', 'image/png', 'image/jpeg', 'image/gif', 'image/webp'],
+  mimeTypes: [
+  'image/bmp',
+  'image/png',
+  'image/jpeg',
+  'image/gif',
+  'image/webp'
+],
   defaultRank: 90,
   createRenderer: options => new widgets.RenderedImage(options)
 };
--- a/packages/rendermime/src/factories.ts
+++ b/packages/rendermime/src/factories.ts
@@ -20,7 +20,7 @@ export const htmlRendererFactory: IRenderMime.IRendererFactory = {
  */
 export const imageRendererFactory: IRenderMime.IRendererFactory = {
   safe: true,
-  mimeTypes: ['image/bmp', 'image/png', 'image/jpeg', 'image/gif'],
+  mimeTypes: ['image/bmp', 'image/png', 'image/jpeg', 'image/gif', 'image/webp'],
   defaultRank: 90,
   createRenderer: options => new widgets.RenderedImage(options)
 };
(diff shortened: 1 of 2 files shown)


packages/rendermime/src/factories.ts
```diff
@@ -20,7 +20,7 @@ export const htmlRendererFactory: IRenderMime.IRendererFactory = {
  */
 export const imageRendererFactory: IRenderMime.IRendererFactory = {
   safe: true,
-  mimeTypes: ['image/bmp', 'image/png', 'image/jpeg', 'image/gif'],
+  mimeTypes: ['image/bmp', 'image/png', 'image/jpeg', 'image/gif', 'image/webp'],
```
fcollonval, line 23: Prettier is not happy :wink: 

```suggestion
  mimeTypes: [
    'image/bmp',
    'image/png',
    'image/jpeg',
    'image/gif',
    'image/webp'
  ],
```

fcollonval (changes requested):
Thanks @alec-kr 

Would you mind fixing the test at packages/rendermime/test/factories.spec.ts:L428 by adding the new type there too?

alec-kr:
> Would you mind fixing the test at packages/rendermime/test/factories.spec.ts:L428 by adding the new type there too?

Sure! I'll get that fixed.

packages/rendermime/src/factories.ts
alec-kr: I have updated the test at packages/rendermime/test/factories.spec.ts:L428, and I have also resolved the Prettier error is also resolved now. @fcollonval

fcollonval:
Testing on binder, it is still not working because actually the server is not sending the correct mimetype for .webp file. See this comparison between the metadata from .webp and .png files:

image

The trouble is that the mimetype is not supported by the Python library `mimetypes` used in Jupyter Server:

https://github.com/jupyter-server/jupyter_server/blob/15310fd760d21be84a1509c0f875970c0c0faaef/jupyter_server/services/contents/filemanager.py#L311

Adding webp was refused because it is not official register as IANA type:
- https://bugs.python.org/issue11362
- https://bugs.python.org/issue45639
- https://bugs.python.org/issue38902

But it got eventually merged recently: https://github.com/python/cpython/pull/29259 - to be released in Python 3.11

I would suggest opening a PR to jupyter server to be able to add additional mimetypes. Ideally through a configurable like `Dict[str, List[str]]` where the keys are mimetypes and the list is all supported extensions; e.g. {'image/webp': ['.webp']}. The mapping will be inverted within the server for quicker look up.

fcollonval (approved):
Thanks @alec-kr 

I'm gonna merge this as it is required. But this is not yet enough see previous comment.

fcollonval:

alec-kr:
Thank you for merging my first PR @fcollonval 😄.  

As for the issue with the ```mimetypes``` library: Why wouldn't the addition of the webp mimetype in [python/cpython#29259](https://github.com/python/cpython/pull/29259) be reflected in Jupyter Server since it uses the same library?

fcollonval:
> Thank you for merging my first PR

Thank you for contributing :wink:

> Why wouldn't the addition of the webp mimetype in https://github.com/python/cpython/pull/29259 be reflected in Jupyter Server since it uses the same library?

It will. But if the Python PR is not backported in older Python versions, you will need to wait for Python 3.11 to be released to have that extension recognized by the server.

alec-kr:
I've had a look at the ```mimetypes``` library, and noticed that ```mimetypes.add_type('image/webp', '.webp')``` seems to solve this issue. 

Could this possibly be implemented on Jupyter Server to add custom mimetypes, or would using the ```Dict[str, List[str]]``` as you suggested be more advantageous?

fcollonval:
> I've had a look at the mimetypes library, and noticed that mimetypes.add_type('image/webp', '.webp') seems to solve this issue.

This is probably the best solution indeed - feel free to open a PR in [jupyter_server repo](https://github.com/jupyter-server/jupyter_server).

You can open an issue prior to open the PR if you don't know the best place for adding that statement. @Zsailer will be more able than me to help you.

alec-kr:
Thanks for the tip! That was quite helpful. I will open an issue over on that repo shortly.

PR #13068, opened 2022-09-08
Raise ceiling on jupyter_server dependency to < 3

Zsailer:
## References

Discuss in today's [Jupyter Server meeting](https://github.com/jupyter-server/team-compass/issues/15#issuecomment-1241023014).

Also see https://github.com/jupyterlab/jupyterlab_server/pull/302. While it's not strictly necessary to merge+release that PR, first, this PR isn't useful until jupyterlab_server relaxes its pin.

## Code changes

This raises the ceiling on JupyterLab's jupyter_server dependency.

This is a temporary change. I think we want JupyterLab 4.0 to depend on Jupyter Server 2.0, i.e. `jupyter_server~=2.0` once Jupyter Server 2.0 reaches final release. JupyterLab 4.0 won't work with Jupyter Server 1.0 (since RTC will likely rely on Jupyter Server 2.0 features).

This change should also be backported to a JLab 3.5 branch/release, but we'll keep this change as-is going forward—users can always use either JS 1.x and 2.x in JLab >= 3.5 

## User-facing changes

This enables folks to test Jupyter Server 2.0 with JupyterLab 4.0.
 

None.

cc @fcollonval


Zsailer:
@fcollonval, woah sorry I checked in the setup.cfg! Apparently I was using an old copy of jupyterlab 😆

fcollonval (approved):
Thanks @Zsailer for both PRs

> woah sorry I checked in the setup.cfg! Apparently I was using an old copy of jupyterlab laughing

Don't worry about it

Zsailer:
We also needed to relax the dependency in notebook_shim: https://github.com/jupyter/notebook_shim/pull/12

I'll cut a release in the next hour or so.

Zsailer:
> This change should also be backported to a JLab 3.5 branch/release, but we'll keep this change as-is going forward—users can always use either JS 1.x and 2.x in JLab >= 3.5

@fcollonval Do you want to create 3.5.x branch now? I can backport this to PR to that branch.

fcollonval:
@Zsailer thanks for proposing - I would like to wait for the list of PRs to be frozen before doing so (Sept 21st). That leaves us the ability to do patch release on 3.4.x without increasing maintenance load by backporting things twice.

fcollonval:

Issue #12972, opened 2022-08-19
Running terminal and kernels - Scrollable headers when sections are empty


firai:
## Description

The section headers in the "Running terminals and kernels" panel are slightly scrollable when the corresponding sections are empty. When the browser is set to use non-overlay scrollbars, scrollbars are visible. The scrollability of the headers and the scrollbars go away if the section is populated. Seems like the behavior might be caused by the hardcoded header height combined with `overflow: auto` in the section container.

**Edit:** This seems to only happen consistently in Firefox with default settings. It happened to me for particular builds inconsistently in Chrome with default font sizes (haven't figured out the pattern), but it also appears when the UI font size is increased.

![image](https://user-images.githubusercontent.com/23555/185653052-fa2c0aed-b50a-4a36-a7a2-9961ac550068.png)

## Reproduce

Open "Running terminal and kernels" with clean start-up.

## Expected behavior

No scrollbars at the headers.

## Context

- Operating System and version: Windows 11
- Browser and version: Chrome (104.0.5112.102) and Firefox (103.0.2)
- JupyterLab version: 3.4.5 and latest development version

firai:
@meeseeksdev tag pkg:running

jweill-aws:
Note for resolver: You should be able to fix this by increasing the `height` of the headers. Please test in Windows 11 using the browsers noted above.

firai:
Rendered obsolete by #13074. It doesn't seem like anyone else noticed the issue anyway.

PR #12899, opened 2022-08-02
LSP follow-up

trungleduc:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/12534
 - [x] Option to enable/disable LSP service
 - [x] Notebook adapter memory leak issue
 - [x] Documentation for language servers installation and configuration.
 - [x] Guide for creating `LSP` extension.

## Code changes
- Add new methods to the `LanguageServerManager` class to enable/disable language servers 

## User-facing changes
New option in the setting page allowing users to enable/disable LSP service

lsp

PR #13074, opened 2022-09-09
Running panel - switch to SidePanel

### LINKED ISSUE #12983: Running terminals and kernels - Large number of open tabs and/or kernels pushes subsequent sections out of visibility
## Description

There appears to be no minimum height to the sections in the Running Terminals and Kernels panel when other sections grow. When there are a large number of open tabs and/or kernels, the items in subsequent sections (i.e. language servers and terminals) become not visible and inaccessible, since the sections also aren't resizable.

**Edit:** Updated screenshot at https://github.com/jupyterlab/jupyterlab/commit/ff37c4f4324fbc6640f5c35777a79d64552696e2. Section header doesn't seem to completely disappear like the earlier screenshot anymore, but the running terminal list is still only visible through a sliver of a box in this example.

image

## Reproduce

1. Open a large number of tabs and/or kernels.
2. Open the "Running Terminals and Kernels" panel.

## Expected behavior

All sections of the "Running Terminals and Kernels" panels are visible and accessible even with a large number of items in one or more sections.

## Context

- Browser and version: Chrome (104.0.5112.102) and Firefox (103.0.2)
- JupyterLab version: 3.4.5 and latest development version (the former does not have the "language servers" section)

## Potential solutions

1. Impose minimum heights on the lists in a `:has`, and/or
2. Make the panels resizable (latter is obviously more involved)

fcollonval:
## References

Fixes #12983

## Code changes

Use `SidePanel` to render the running side bar
Add `added` signal to `IRunningSessionManagers` to notify view a new manager is available.

## User-facing changes

- More homogeneous look and feel with others sidebars
- Each list is resizable
- Each list is scrollable when overflow occurs

Before | After
--- | ---
before | after

`RunningSessions` is now a `SidePanel` not a `ReactWidget`
We could consider that as a safe change in case of backport.


fcollonval:
@firai you can test this PR via the binder link above. Feel free to comment back here.

firai:
Thanks, @fcollonval! You can also add #12972 to the list of issues that this PR fixes.

The "Language Servers" section label seems to overflow its box on Firefox (see below). This appears to be occurring in your screenshot above as well. Maybe this can be mitigated by either removing the `overflow: hidden` from `.lm-AccordionPanel-titleLabel` or by widening the minimum/default sidebar width? The issue does not seem to exist on Chrome with default font sizes.
image

A couple of other observations that are related but not directly resulting from this PR, so these might need to be opened as separate issues/PRs:

- Each section initializes with equal height, which is the same behavior before the PR, but practically speaking the number of tabs and kernels would likely be far more than the number of language servers and terminals. I'm not sure if it's practical to make the first two sections longer than the other two by default, or if the even section height aesthetic is actually preferred (though the debugger sections aren't exactly evenly distributed by default either).
- The "Close All" and "Shutdown All" button labels seem to have excessive side margins compared to icon-style toolbar buttons, such as the refresh button at the top of the Running panel or those in the debugger. Since the button already has padding, I'm not sure if the label margins are even necessary.

fcollonval:
Thanks for the feeback @firai 

Regarding the ellipsis, I keep it like this because it is gonna be tricky to shrink the button label within the toolbar. But as I removed the margin following your suggestion, it should be better (I unfortunately don't have the time to try with Firefox on Debian).

Regarding the size of the panels, I also keep it as is (aka each section have an even space). The reason for it is that additional plugins can add panels to that sidebar. So trying to get different ratio that matters will be very tricky. A better follow-up would be to save the panel sizes. So it gets restored when opening JLab.

firai:
Thanks, @fcollonval! Looks like getting rid of the label margins did the trick on Windows Firefox.
image

fcollonval:
bot please update snapshots

fcollonval:
Kicking CI

jtpio (approved):
Nice thanks!

jtpio:
@fcollonval do you know if the failing memory leak check is relevant? https://github.com/jupyterlab/jupyterlab/actions/runs/3142433060/jobs/5107702133

(it failed a second time after restarting it)

fcollonval:
> do you know if the failing memory leak check is relevant?

@jtpio the memory leak is requiring some fixes in https://github.com/jupyterlab/jupyterlab/pull/12899/commits/bf4edb785f00652456bd1ea79dd3dc85978464d7 part of https://github.com/jupyterlab/jupyterlab/pull/12899

Thanks for the review, I'll merge this one.

Issue #13018, opened 2022-08-26
Debugger side bar not minimized when disabling the debugger


karlaspuldaro:
### Problem
When the debugger is enabled, the debugger side panel is expanded.
However, if one disables the debugger from the toolbar the side panel is currently not minimized, forcing the user to do that manually.

![image](https://user-images.githubusercontent.com/25207344/186985792-247f415c-de37-4bf8-9bfd-c49123e5cc87.png)

### Proposed Solution
Minimize the debugger side bar when the user disables the debugger.

ohrely:
Closing because #13088 fixes this issue.

PR #13088, opened 2022-09-14
Collapse debugger panel when disabling debugger

yanmulin:
## References

Addressing the issue: https://github.com/jupyterlab/jupyterlab/issues/13018

## Code changes

listen to Debugger `terminated` event and collapse the debugger panel if it is visible.

## User-facing changes

Before
Recording 2022-09-13 at 19 22 11

After
Recording 2022-09-13 at 19 20 12

None


fcollonval (changes requested):
Thanks for opening this PR @yanmulin 

Would you mind not closing the tab if multiple debug sessions are active (see screencast) or place this behind a setting?

multiple-debug-sessions

fcollonval:
I don't think we can figure out if there are other opened widgets with an active debug session. @afshin is that correct?

krassowski:
Performance and UX wise:
- closing the panel will cause layout shift and freeze apps if there are too many nodes (not that big of a problem with windowed rendering, but if someone has huge SVG on screen it still might happen)
- for the same reason it will shift the content of all open editors/widgets, possibly leaving the user scrolling around to find where they were.

So _ideally_ I would see the behaviour as switching to most recently used non-debugger sidebar widget on that side first and only if debugger is the only tab collapsing the right sidebar.

yanmulin:
@fcollonval  A quick solution is to add a global variable to track the number of active debugger session. But not sure if it is a good idea.

yanmulin:
@krassowski  Couldn't trigger the layout shift on my side with a random SVG found online. The cells just expand horizontally.

Recording 2022-09-20 at 01 55 34

fcollonval:
> A quick solution is to add a global variable to track the number of active debugger sessions. It increments when a session is started and decrements when a session is stopped. But not sure if it is a good idea.

I agree this is not an elegant solution - let's add a settings for this new feature. Do you need pointer for that?

> Couldn't trigger the layout shift on my side with a random big SVG found online. The cells just expand horizontally. Is that because the SVG not big enough?

I think @krassowski was making a general comment that changing the width of a notebook will trigger style recalculation (and layout shift as the content distribution will change). This can freeze the UI if the notebook is large and/or contains highly complex views.

yanmulin:
Recording 2022-09-20 at 22 44 53 (1)

fcollonval (changes requested):
Sorry for the late feedback @yanmulin.

Would you mind accepting the suggestion to make the settings follow the format name?

krassowski:
@yanmulin here the layout shift has two meanings:
1) browser's need to re-compute the layout after a DOM/CSS change that could have shifted the layout
2) user-perceivable movement of elements

I mentioned both together, but you do not the (2) to have the (1). The performance consequences of layout shift in the sense of (1) are more severe when there are many nodes; SVG visualisations can have thousands or even millions of nodes, but this is not a precondition of layout shift occurring.

Here is an example of (1) - in visible in the debugger and (2) - visible because the print output with expanded/collapsed side panel takes either one or two lines:

layout_shift

Even if we remove the `print()` call, the layout shift in the sense of (1) still occurs:

layout_shift_2

fcollonval (approved):
Thanks @yanmulin

fcollonval:

PR #13093, opened 2022-09-15
Properly reload the code mirror editor on shared model switched

a3626a:
## Introduction

After https://github.com/jupyterlab/jupyterlab/pull/11638, CodeMirror Editor is not properly reloaded after the shared model switch. This breaks code mirror editor. It seems like built-in editors(file editor and notebook editor) are not affected by this bug, but my custom document editor is affected. 

## References

https://github.com/jupyterlab/jupyterlab/pull/11638

## Code changes

Originally, `_initializeEditorBinding` is connected to `sharedModelSwitched` signal. However `_initializeEditorBinding` just creates an object(`_yeditorBinding`), which does not change the editor at all.
```
  private _initializeEditorBinding(): void {
    const sharedModel = this.model.sharedModel as models.IYText;
    this._yeditorBinding = {
      text: sharedModel.ysource,
      awareness: sharedModel.awareness,
      undoManager: sharedModel.undoManager
    };
    // It just creates an object. and do nothing.
  }
```

To property change yeditorBinding, the state of the editor must be changed. This can be achieved by calling `setState` function to the `this._editor`.
```
    editor.setState(
      // editor state is required when the editor is created and updated. So I introduced a new function for that, 'createEditorState'
      createEditorState(config, ybinding, editorConfig, additionalExtensions)
    );
```

I introduced two new function `Private.createEditorState` and `Private.updateEditor`.

### `Private.createEditorState`
I extracted editor state construction logic from `Private.createEditor`. This function is used by `Private.createEditor` and `Private.updateEditor`

### `Private.updateEditor`
It updates editor state, so the editor is properly reloaded according to newly switched shared model. This function is called by `sharedModelSwitched` signal.

## User-facing changes

No

No

### DIFF
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -252,7 +252,21 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
     );
 
     // every time the model is switched, we need to re-initialize the editor binding
-    this.model.sharedModelSwitched.connect(this._initializeEditorBinding, this);
+    this.model.sharedModelSwitched.connect(() => {
+      this._initializeEditorBinding();
+      Private.updateEditor(
+        this._editor,
+        fullConfig,
+        this._yeditorBinding,
+        this._editorConfig,
+        [
+          this._markField,
+          Prec.high(domEventHandlers),
+          updateListener,
+          translation
+        ]
+      );
+    }, this);
 
     this._onMimeTypeChanged();
     this._onCursorActivity();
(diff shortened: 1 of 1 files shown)


packages/codemirror/src/editor.ts
```diff
@@ -252,7 +252,21 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
...
+        this._editor,
+        fullConfig,
+        this._yeditorBinding,
+        this._editorConfig,
+        [
+          this._markField,
+          Prec.high(domEventHandlers),
+          updateListener,
+          translation
+        ]
```
a3626a, line 267: These arguments are identical to the arguments for `Private.createEditor` See L242~L251

packages/codemirror/src/editor.ts
```diff
@@ -1187,13 +1201,12 @@ export namespace CodeMirrorEditor {
  * The namespace for module private data.
  */
 namespace Private {
-  export function createEditor(
-    host: HTMLElement,
+  export function createEditorState(
```
a3626a, line 1204: Git compare is not quite clear. We still have `createEditor` below.

a3626a:
I have found that 'setState' on Editor can cause unexpected result. Using CodeMirror compartment and dispatch function is much better. `y-codemirror.next` provides `ySyncFacet` for this purpose. However it is fetching the facet only once during construction. So dynamically changing `ySyncFacet` has no effect. 

I have created an issue for this.

dmonad:
This issue would also be fixed by #12695 which removes the `sharedModelSwitched` functionality.

fcollonval:
Closing as #12695 has been merged - that removes the use of `sharedModelSwitched`

PR #13109, opened 2022-09-20
Avoid menus overflowing in small screens

steff456:
## References

Fixes https://github.com/jupyterlab/lumino/issues/418

## Code changes

This code changes avoid forcing the `x` value in menus, which avoid the overflowing in small screens or high zoom.

## User-facing changes

The change is perceptible just in small screens or when high zoom is present like the following screenshots,

### New behavior

image

### Old behavior

image

In 100% zoom it continues working without any change. It would be interesting to know if there's some idea why the `x` was enforced given that this is not the default Lumino value.

There's no backwards incompatible change.

### DIFF
--- a/packages/mainmenu/src/mainmenu.ts
+++ b/packages/mainmenu/src/mainmenu.ts
@@ -24,7 +24,8 @@ export class MainMenu extends MenuBar implements IMainMenu {
    * Construct the main menu bar.
    */
   constructor(commands: CommandRegistry) {
-    super();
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @steff456 

> During the weekly call, nobody raises concerns about changing this behavior. And people agree that it will improve the user experience.

fcollonval:

krassowski:
Just for the record I believe that it will not cause a performance degradation as all DOM querying and modifications happen regardless of this switch (as currently implemented in lumino).

PR #13111, opened 2022-09-22
Add a setter to TagWidget's parent

brichet:
A getter has been added to the `TagWidget`'s *parent* attribute in `v4.0.0a28`, without setter (which should rely to its parent class).

It look like Typescript does not allow overriding the getter without the setter and vice versa.

At the end, it is not possible to add a new tag to a cell : 

image

## Code changes

Add a setter to `TagWidget`'s *parent* attribute.

None

### DIFF
--- a/packages/celltags/src/widget.ts
+++ b/packages/celltags/src/widget.ts
@@ -23,10 +23,13 @@ export class TagWidget extends Widget {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @brichet 

> CI failures are not related.

PR #13113, opened 2022-09-22
Remove xeus-python installation for debugger test

fcollonval:
## References

CI is failing because the `xeus-python` package pin in `ci_install.sh` script is too old for the new default Python version.

As that pin predate the ability of ipykernel to support debugger protocol, I'm simply dropping that specific installation.

## Code changes

Drop installing xeus-python for _js-debugger_

## User-facing changes

N/A

N/A

### DIFF
--- a/scripts/ci_install.sh
+++ b/scripts/ci_install.sh
@@ -37,8 +37,3 @@ if [[ $GROUP == nonode ]]; then
     sudo rm -rf $(which node)
     ! node
 fi
-
-# The debugger tests require a kernel that supports debugging
-if [[ $GROUP == js-debugger ]]; then
-    pip install xeus-python">=0.9.0,<0.10.0"
-fi


krassowski:
Not opposing this PR, but I think that it would be nice to have at least some tests with an alternative kernel (IRKernel, xeus, whatever) even if that would be a separate job with different version constraints.

(the above could be moved to a new issue if going ahead with merging this one as-is)

fcollonval:

jtpio:
> (the above could be moved to a new issue if going ahead with merging this one as-is)

Right the `xeus-python` CI dependency was indeed added to originally test debugging on CI before `ipykernel` had support for it.

Makes sense to remove it now, although it would indeed be interested to test the debugger with more kernels.

SylvainCorlay:
Bumping to a more recent version could have worked too.

fcollonval:
> Bumping to a more recent version could have worked too.

It would have worked too. But as `ipykernel` is ship by default with JupyterLab, it makes more sense to test against it.

PR #13114, opened 2022-09-22
Fix js-debugger test using xeus-python

brichet:
Update `xeus-python` dependency to `">=0.14.0,<0.15.0"` for `js-debugger` test.
The current required version (`">=0.9.0<0.10.0"`) is not available anymore on Pypi ???

None


brichet:
:+1:  I just seen it

PR #13124, opened 2022-09-26
Pin jupyter_ydoc to 0.2

hbcarlos:
At the moment, RTC is broken in master because the schema of the notebook between the frontend and the backend doesn't match. This PR pins `jupyter_ydoc` to a version with the same schema as the frontend.


pyproject.toml
```diff
@@ -42,7 +42,8 @@ dependencies = [
     "jupyter_core",
     "jupyter-lsp>=1.5.1",
     "jupyter_server>=1.16.0,<3",
-    "jupyter_server_ydoc>=0.1.9,<0.2.0",
+    "jupyter_ydoc>=0.2.0,<0.3.0",
```
davidbrochart, line 45: I think it should be:
```suggestion
    "jupyter_ydoc>=0.3.0a0,<0.4.0",
```

hbcarlos: No, not yet. Right now, the schema we use in the frontend needs v0.2.0.
We need to merge this PR and backport it to JupyterLab v0.3.x.
And later, I will work on a PR to change the schema to support jupyter_ydoc v0.3.0. But that is a bigger PR.

davidbrochart: > We need to merge this PR and backport it to JupyterLab v0.3.x.

I guess you mean JupyterLab v3.4.x. Why not opening a PR on https://github.com/jupyterlab/jupyterlab/tree/3.4.x directly?

hbcarlos: > I guess you mean JupyterLab v3.4.x.

Yes, sorry.

> Why not opening a PR on https://github.com/jupyterlab/jupyterlab/tree/3.4.x directly?

Because, right now, the master branch needs jupyter_ydoc 0.2.0

davidbrochart: > Because, right now, the master branch needs jupyter_ydoc 0.2.0

The master branch used to work with the "new" schema (the one in `jupyter_ydoc` v0.3.0a0). Do you mean #12695 reverted that?

hbcarlos: Yes

hbcarlos: Actually, Lab 3.4.x is not using `jupyter_ydoc`, and we don't need to backport it.

davidbrochart: But it will, right? See #13131.

fcollonval (approved):
Thanks @hbcarlos

PR #13154, opened 2022-09-29
Remove `width: 100%` of `jp-WindowedPanel-window`

jtpio:
## References

Noticed while iterating on the JupyterLab update for Notebook 7 in https://github.com/jupyter/notebook/pull/6539.

This extra `width: 100%` causes some layout issues in Notebook 7 and does not seem necessary in JupyterLab. 

## Code changes

Update CSS for `jp-WindowedPanel-window`.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #13157, opened 2022-09-30
Remove `width: 100%` of `jp-WindowedPanel-inner`

jtpio:
## References

Similar to https://github.com/jupyterlab/jupyterlab/pull/13154.

## Code changes

Cleanup some CSS which does not seem required in JupyterLab, but can cause issues in downstream applications reusing the notebook component (https://github.com/jupyter/notebook/pull/6539)

## User-facing changes

None

None


fcollonval (approved):
Thanks @jtpio

PR #13168, opened 2022-10-03
Improve shared-models API

fcollonval:
## References

Follow-up on https://github.com/jupyterlab/jupyterlab/pull/12695

Fix https://github.com/jupyterlab/jupyterlab/issues/13207

Improve API and prepare for removal of modelDB in a follow-up

## Code changes

- Improve shared-models API
- Avoid creating YCell outside YNotebook to protect against inconsistency
- Add signals to metadataChange and cellsChanged
- Allow partial get/set of metadata.

## User-facing changes

None

Yes API of shared models are changed to improve consistency and align closer to internal JLab mechanism to listen for partial changes.


packages/shared-models/src/api.ts
```diff
@@ -426,115 +519,236 @@ export interface ISharedMarkdownCell
...
+
+  /**
+   * Item(s) were removed from the list.
+   */
+  | 'remove'
+
+  /**
+   * An item was set in the list.
+   */
+  | 'set';
```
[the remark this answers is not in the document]
fcollonval: Good point I'll limit to _add_ and _remove_

packages/shared-models/src/api.ts
```diff
@@ -71,28 +80,32 @@ export interface ISharedText extends ISharedBase {
...
    * The changed signal.
    */
-  readonly changed: ISignal<this, TextChange>;
+  readonly changed: ISignal<this, SourceChange>;
+
   /**
-   * Gets cell's source.
+   * Text
+   */
+  source: string;
```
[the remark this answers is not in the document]
fcollonval: Using properties instead of getter and setter functions is a more common pattern in JLab.
I did not remote the getter/setter though to ease backport to 3.x.

packages/shared-models/src/api.ts
```diff
@@ -394,14 +477,14 @@ export interface ISharedCodeCell
...
-export interface ISharedMarkdownCell
-  extends ISharedBaseCell<ISharedBaseCellMetadata> {
+export interface ISharedAttachmentsCell
+  extends ISharedBaseCell<nbformat.IBaseCellMetadata> {
   /**
-   * String identifying the type of cell.
+   * Cell attachments
    */
-  cell_type: 'markdown';
+  attachments?: nbformat.IAttachments;
```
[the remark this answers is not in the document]
fcollonval: See reply for `source`

packages/shared-models/src/api.ts
```diff
@@ -362,6 +440,11 @@ export interface ISharedCodeCell
    */
   execution_count: nbformat.ExecutionCount;
 
+  /**
+   * Cell outputs
+   */
+  outputs: Array<nbformat.IOutput>;
```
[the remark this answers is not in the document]
fcollonval: See reply for `source`

packages/shared-models/src/api.ts
```diff
@@ -328,18 +389,35 @@ export interface ISharedBaseCell<Metadata extends ISharedBaseCellMetadata>
...
+   * If only one argument is provided, it will override all cell metadata.
+   * Otherwise a single key will be set to a new value.
+   *
+   * @param metadata Cell's metadata or key.
+   * @param value Metadata value
+   */
+  setMetadata(
+    metadata: Partial<Metadata> | string,
+    value?: PartialJSONValue
+  ): void;
```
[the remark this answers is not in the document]
fcollonval: I kept the global get/set to ease backport to 3.x

packages/shared-models/src/api.ts
```diff
@@ -127,23 +167,36 @@ export interface ISharedNotebook extends ISharedDocument {
...
    *
-   * @param metadata: Notebook's metadata.
+   * @param metadata All Notebook's metadata or the key to set.
+   * @param value New metadata value
    */
-  setMetadata(metadata: nbformat.INotebookMetadata): void;
+  setMetadata(
+    metadata: nbformat.INotebookMetadata | string,
+    value?: PartialJSONValue
+  ): void;
```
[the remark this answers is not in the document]
fcollonval: Same comment as for cell metadata

packages/cells/src/model.ts
```diff
@@ -185,7 +185,11 @@ export function isRawCellModel(model: ICellModel): model is IRawCellModel {
  */
 export class CellModel extends CodeEditor.Model implements ICellModel {
   constructor(options: CellModel.IOptions<ISharedCell> = {}) {
-    super(options);
+    super({
+      sharedModel: createStandaloneCell({ cell_type: 'raw', id: options.id }),
+      ...options
+    });
+    this._standaloneModel = typeof options.sharedModel === 'undefined';
```
hbcarlos, line 192: I'm not sure about this. What happens if we create a standalone Code cell?
In this case, `options.sharedModel` won't be `undefined` and `this_standalone` will be false, so when disposing the cell, we will not dispose the sharedModel, right?

I tried the console, and the cells there are not standalone.

packages/settingeditor/src/raweditor.ts
```diff
@@ -49,11 +48,7 @@ export class RawEditor extends SplitPanel {
 
     // Create read-only defaults editor.
     const defaults = (this._defaults = new CodeEditorWrapper({
-      model: new CodeEditor.Model({
-        sharedModel: createStandaloneCell({
-          cell_type: 'code'
-        })
-      }),
+      model: new CodeEditor.Model(),
```
hbcarlos, line 51: In this use case, when disposing the `CodeEditor.Model` we are not disposing the sharedModel because the `CodeEditor.Model` leaves that responsibility to its child (The class that extends it, DocumentModel or CellModel).

I think we should do the same trick we do in the CellModel:
`this._standaloneModel = typeof options.sharedModel === 'undefined';`

packages/settingeditor/src/raweditor.ts
```diff
@@ -62,11 +57,7 @@ export class RawEditor extends SplitPanel {
 
     // Create read-write user settings editor.
     const user = (this._user = new CodeEditorWrapper({
-      model: new CodeEditor.Model({
-        sharedModel: createStandaloneCell({
-          cell_type: 'code'
-        })
-      }),
+      model: new CodeEditor.Model(),
```
hbcarlos, line 60: Same here.

packages/debugger/src/factory.ts
```diff
@@ -34,10 +33,7 @@ export class ReadOnlyEditorFactory {
     const mimeTypeService = this._services.mimeTypeService;
     const editor = new CodeEditorWrapper({
       model: new CodeEditor.Model({
-        sharedModel: createStandaloneCell({
-          cell_type: 'code',
-          source: content
-        }),
+        sharedModel: YFile.create(content),
```
hbcarlos, line 36: > In this use case, when disposing the CodeEditor.Model we are not disposing the sharedModel because the CodeEditor.Model leaves that responsibility to its child (The class that extends it, DocumentModel or CellModel).

> I think we should do the same trick we do in the CellModel:
this._standaloneModel = typeof options.sharedModel === 'undefined';

Maybe that trick is not a good idea, since in this case we will not dispose the sharedModel.

packages/codeeditor/src/viewer.ts
```diff
@@ -29,8 +28,7 @@ export class CodeViewerWidget extends Widget {
     options: CodeViewerWidget.INoModelOptions
   ): CodeViewerWidget {
     const model = new CodeEditor.Model({
-      mimeType: options.mimeType,
-      sharedModel: createStandaloneCell({ cell_type: 'code' })
+      mimeType: options.mimeType
     });
```
hbcarlos, line 32: Same here. This will never be disposed.

packages/codeeditor/src/jsoneditor.ts
```diff
@@ -78,9 +74,7 @@ export class JSONEditor extends Widget {
     this.node.appendChild(this.headerNode);
     this.node.appendChild(this.editorHostNode);
 
-    const model = new CodeEditor.Model({
-      sharedModel: createStandaloneCell({ cell_type: 'code' })
-    });
+    const model = new CodeEditor.Model();
```
hbcarlos, line 77: Same here.

hbcarlos (commented):
Hi @fcollonval! I looked at the latest changes, and I left some comments.

packages/debugger/src/factory.ts
fcollonval: Good point, I will set the source after creating the model.

packages/cells/src/model.ts
fcollonval: > That part was pushed in a separate PR addressing existing notebook memory leaks

hbcarlos (approved):
Thanks, @fcollonval! LGTM.

fcollonval:
CI failure is not related

PR #11272, opened 2021-10-10
Fix ANSI vs URL conflict, prefix `www.` with `https://`

krassowski:
## References

Fixes #10948.

## Code changes

- `autolink` now operates on `textContent` rather than the HTML string, allowing it to properly capture the URL independent of any decorations/highlights from ANSI added as span elements in the previous step
- `autolink` now returns an array of nodes (text nodes or anchor elements) instead of transformed text with embedded HTML, which guarantees that we will not break the DOM even if the implementation is not perfect
- `renderText` now re-conciliates the arrays of nodes:
   - from the `<pre>` (which should only include text nodes and shallow span elements; the text nodes are expected due to sanitization and the span elements are added by ANSI substitution),
   - with the nodes returned by `autolink` (which should only include text nodes and shallow anchor elements).
- `https://` prefix is added when needed to prevent creation of malformed URLs
- a test-case for #10948 was added

## User-facing changes

- the ANSI should no longer interfere with URL and vice versa
- auto-link URLs for addresses missing protocol but starting with `www.` will will now include `https://` preventing creation of incorrect links such as `https://localhost:8888/www.example.com`

For manual testing:

```
print('\x1b[48;2;185;0;129mwww.python.\x1b[0m\x1b[48;2;113;0;119morg\x1b[0m')
raise ValueError('www.example.com')
```

Before:

before

After:

after

## Backwards-incompatible changes

None (`autolink` changes signature, but it is not exported).

PR #10301, opened 2021-05-26
Add Yjs as a singleton package

dmonad:
## References

This addresses #10266. Extensions that define custom shared modules should share the Yjs dependency with JupyterLab. This reduces dependencies but is also needed for technical reasons as explained in the ticket.

I confirmed with @hbcarlos that this PR fixes the mentioned problem and that the yjs package is now shared with extensions.

## Code changes

This PR adds Yjs as a singleton package and updates Yjs-related dependencies. Yjs and related modules are exported as EcmaScript modules and as CommonJS modules. I made sure that from now on all packages consume the EcmaScript module version of Yjs which should result in smaller bundles.

## User-facing changes

None.

## Backwards-incompatible changes

None.

PR #13202, opened 2022-10-07
Optimize text mimerenderer: ansi vs autolink 

vidartf:
## References

Fixes JS tests on Windows by removing some backslashes that were added in #10301.

Optimizes ansi color coding vs auto-linking in plain text rendermime. #11272

## Code changes

This optimizes the rendering for cases where there are many more elements of one type than the other. E.g. for a large syntax highlighted source file. In this case there can be many (tens of) thousands of span elements, but likely only a few links. In this case, multiple span elements will all correspond to a single text node in the autolink output. In these cases, we can simply use the nodes from the span element list, without splitting / cloning any nodes. This greatly reduces the time spent.

## User-facing changes

UI hangs less. For an extremely symptomatic file in the inspector, time spent is reduced from ~40s to ~4s.

Old, 41s:
renderText-old-41s

New, 4s:
renderText-new-3s

Should only change internal details. Ideally this would be backported to 3.x , but then we would likely need to rewrite using lumino iterators.

### DIFF
--- a/packages/rendermime/src/renderers.ts
+++ b/packages/rendermime/src/renderers.ts
@@ -684,7 +684,7 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
       if (!nodes[0]) {
         combinedNodes.push(nodes[1]);
         inAnchorElement = nodes[1].nodeType !== Node.TEXT_NODE;
-        continue
+        continue;
       } else if(!nodes[1]) {
         combinedNodes.push(nodes[0]);
         inAnchorElement = false;
--- a/packages/rendermime/src/renderers.ts
+++ b/packages/rendermime/src/renderers.ts
@@ -554,14 +554,99 @@ function splitShallowNode<T extends Node>(
   at: number
 ): { pre: T; post: T } {
   const pre = node.cloneNode() as T;
-  pre.textContent = node.textContent?.substr(0, at) as string;
+  pre.textContent = node.textContent?.slice(0, at) as string;
   const post = node.cloneNode() as T;
-  post.textContent = node.textContent?.substr(at) as string;
+  post.textContent = node.textContent?.slice(at) as string;
   return {
-    pre: pre,
-    post: post
+    pre,
+    post
   };
 }
+
+
+/**
+ * Iterate over some nodes, while tracking cumulative start and end position.
+ */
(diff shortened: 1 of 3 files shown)


packages/rendermime/src/renderers.ts
```diff
@@ -554,15 +554,100 @@ function splitShallowNode<T extends Node>(
...
+function* alignedNodes<T extends Node, U extends Node>(a: T[], b: U[]): IterableIterator<[T, null] | [null, U] | [T, U]> {
+  let iterA = nodeIter(a);
+  let iterB = nodeIter(b);
+  let nA = iterA.next();
+  let nB = iterB.next();
+  while (!nA.done && !nB.done) {
+    let A = nA.value;
+    let B = nB.value;
+
+    if (A.isText && A.start <= B.start && A.end >= B.end) {
```
krassowski, line 601: This is neat, thank you! I think it would be great to also add a test case for each of the conditions in `packages/rendermime/test/factories.spec.ts`.

packages/rendermime/src/renderers.ts
```diff
@@ -593,43 +679,18 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
-            linkNode,
-            preLen
-          );
-          linkedNodes.unshift(postpone);
-          linkNode = keep;
-        }
+      if (!nodes[0]) {
+        combinedNodes.push(nodes[1]);
+        inAnchorElement = nodes[1].nodeType !== Node.TEXT_NODE;
+        continue
```
fcollonval, line 687:
```suggestion
        continue;
```

fcollonval (approved):
Thanks @vidartf for this. Is it complete?

packages/rendermime/src/renderers.ts
vidartf: Thanks, I added enough tests to get the coverage to 100%. It unveiled a log error! 👍

vidartf:
@fcollonval it should be now, but let's see if CI has anything to say on the matter 😆

fcollonval (approved):
Thanks @vidartf

fcollonval:
CI failure is tracked in #13243

fcollonval:

PR #13216, opened 2022-10-10
Store original path as returned from contents API in the `Contents.IModel`

### LINKED ISSUE #13155: Content Manager Path to notebook overridden by URL path in lab
Consider the following dispatched contents manager created using a snippet from jupyter/notebook#3233.
```python
from notebook.services.contents.filemanager import ContentsManager
from jupyter_server.services.contents.largefilemanager import LargeFileManager
import inspect
VIRTUAL_PREFIX = 'virtual'

def make_dispatched(self, function):
    """
    This function inspects a function and dispatches it out by finding which
    arg is the path, then finding a manager to manage that path and forwarding
    this on
    """
    # functions passed here have self in it, when it's called its bound so -1 for no self
    path_idx = inspect.getargspec(function)[0].index('path') - 1

    def wrapper(*args, **kwargs):
        path = kwargs['path'] if 'path' in kwargs else args[path_idx]
        for manager in self.managers:
            if manager.manages_path(path):
                fn = getattr(manager, function.__name__)
                return fn(*args, **kwargs)
        # This self is jupyter's
        self.log.error("Could not find manager for path %s", path)
        raise Exception('Could not find manager for path "%s"' % path)

    return wrapper


class DispatchedContentsManager(LargeFileManager):
    def __init__(self, **kwargs):
        _managers = [VirtualContentsManager, DefaultContentsManager]
        self.managers = []
        for manager in _managers:
            # Init each manger with the same params
            inst = manager(**kwargs)
            if not hasattr(inst, 'manages_path'):
                raise RuntimeError('Count not find method manages_path in manager!')
            self.managers.append(inst)

        # FIXME: add support for rename_file! This does not follow the pattern below
        for name, method in inspect.getmembers(self,
                                               predicate=inspect.ismethod):
            if (not name.startswith('_') and
                        'path' in inspect.getargspec(method)[0]):
                setattr(self, name, make_dispatched(self, method))


class DefaultContentsManager(LargeFileManager):
    """
    Base contents manager that manages all paths on the filesystem
    """
    def manages_path(self, path):
        return True

class VirtualContentsManager(LargeFileManager):
    """
    Virtual Filesystem that can be used to serve files from any host 
    """

    def manages_path(self, path: str):
        path = path.strip('/')
        return path.startswith(VIRTUAL_PREFIX)

    def get(self, path: str, *args, **kwargs):
        return super().get(path.strip('/')[len(VIRTUAL_PREFIX):], *args, **kwargs)

c.ServerApp.contents_manager_class = DispatchedContentsManager
```

## Description
In the above setup, the `VirtualContentsManager` is just a pass through (in our usecase it used to get files from a different host but here it just returns the files from the same host). 

If lab is launched with the above config and we prefix a notebook path with `/virtual` (eg, `/virtual/home/$USER/Untitled.ipynb`) and inspect the requests, we can see that the path returned in response is `/home/$USER/Untitled.ipynb` which is expected.

Now, if the value of  `tracker.currentWidget.context.path` is checked, it is set to `virtual/home/$USER/Untitled.ipynb` instead of what is returned by contents manager (which ends up being used for any operations like `save`).

Due to this, we need to send a request to the contents manager to get what the actual path of a notebook is.

Also, this works fine in notebook(`nbclassic`). ie. the path used for any requests is the path returned from the contents manager not the path from the url.

The main issue here is how can we get the path returned by the contents manager in the front end apis. 

Here are a few screenshots to show this - 
1. Request sent to `/virtual/home/$USER/Untitled.ipynb` (expected) - 
image
2. Response has the correct path (ie. `home/$USER/Untitled.ipynb`) (expected) -
image

## Reproduce

1. Paste the above code snippet in `~/.jupyter/jupyter_server_config.py`
2. Navigate to any notebook
5. Reload the page by prefixing the path with `/virtual`
6. Try saving the notebook.
7. Repeat the same in notebook to see the difference in behavior

## Expected behavior
I would expect the path from the contents manager to be given higher priority. 

## Context
I am writing a front end extension where I need the actual path of the notebook (the one returned from the contents manager). To do this, I'm unable to use `tracker.currentWidget.context.path` and instead I need to send a request to the contents manager which introduces delay in the extension.

krassowski:
## References

Fixes #13155 using solution (b) as proposed solutions in https://github.com/jupyterlab/jupyterlab/issues/13155#issuecomment-1264770051. I have moderate doubts on whether this is the best solution; (c) seems cleaner but I am not convinced if getting the cleaner solution is worth introducing a breaking change, see discussion in #13155 (whereas this PR could still be backported to 3.5).

## Code changes

Adds optional `serverPath` attribute to `Contents.IModel` and populates it in `ContentsManager` methods by path returned by the server.

Note: an example of the issue was actually in our tests all along: `DEFAULT_FILE` has path `foo/test` which is correctly returned when creating a file:

https://github.com/jupyterlab/jupyterlab/blob/376e52c2fbeae00849a836533d6621ce1b724197/packages/services/test/contents/index.spec.ts#L329-L335

but when a file is queried with `get`:

https://github.com/jupyterlab/jupyterlab/blob/376e52c2fbeae00849a836533d6621ce1b724197/packages/services/test/contents/index.spec.ts#L242-L249

then `other:foo` instead of `other:foo/test` gets returned. This PR adds `foo/test` on extra property, `serverPath` (which for now is optional in the interface).

## User-facing changes

None.

None

### DIFF
--- a/packages/services/src/contents/index.ts
+++ b/packages/services/src/contents/index.ts
@@ -50,6 +50,14 @@ export namespace Contents {
     readonly path: string;
 
     /**
+     * The path as returned by the server contents API.
+     *
+     * #### Notes
+     * Differently to `path` it does not include IDrive API prefix.
+     */
+    readonly serverPath?: string;
+
+    /**
      * The type of file.
      */
     readonly type: ContentType;
@@ -716,12 +724,14 @@ export class ContentsManager implements Contents.IManager {
         return {
           ...contentsModel,
           path: this._toGlobalPath(drive, localPath),
-          content: listing
+          content: listing,
+          serverPath: contentsModel.path
         } as Contents.IModel;
       } else {
         return {
           ...contentsModel,
-          path: this._toGlobalPath(drive, localPath)
+          path: this._toGlobalPath(drive, localPath),
+          serverPath: contentsModel.path
         } as Contents.IModel;
       }
     });
@@ -759,7 +769,8 @@ export class ContentsManager implements Contents.IManager {
         .then(contentsModel => {
           return {
             ...contentsModel,
-            path: PathExt.join(globalPath, contentsModel.name)
+            path: PathExt.join(globalPath, contentsModel.name),
+            serverPath: contentsModel.path
           } as Contents.IModel;
         });
     } else {
@@ -798,7 +809,8 @@ export class ContentsManager implements Contents.IManager {
     return drive1.rename(path1, path2).then(contentsModel => {
       return {
         ...contentsModel,
-        path: this._toGlobalPath(drive1, path2)
+        path: this._toGlobalPath(drive1, path2),
+        serverPath: contentsModel.path
       } as Contents.IModel;
     });
   }
@@ -825,7 +837,11 @@ export class ContentsManager implements Contents.IManager {
     return drive
       .save(localPath, { ...options, path: localPath })
       .then(contentsModel => {
-        return { ...contentsModel, path: globalPath } as Contents.IModel;
+        return {
+          ...contentsModel,
+          path: globalPath,
+          serverPath: contentsModel.path
+        } as Contents.IModel;
       });
   }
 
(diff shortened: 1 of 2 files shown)


fcollonval:
@krassowski thanks for pointing out the inconsistency.

I would be in favor to force `IModel.path` to be consistent - it sounds to me like a bug to have a new file path not matching the one obtained for getting the same file.

Otherwise I'm :+1: to add the `serverPath` to allow backport.

fcollonval:
CI failures are not relevant

fcollonval:

Issue #13209, opened 2022-10-10
"Test Minimum Versions" CI check is failing


jtpio:
## Description

The "Test Minimum Versions" CI check is now failing.

## Reproduce

This is happening on (at least) 2 different PRs:

- https://github.com/jupyterlab/jupyterlab/actions/runs/3219766730/jobs/5265944142
- https://github.com/jupyterlab/jupyterlab/actions/runs/3218127960/jobs/5265863414

## Expected behavior

The check should pass.

## Context

Might be related to the recent releases of `jupyterlab/maintainer-tools`: https://github.com/jupyterlab/maintainer-tools/releases

Since the JupyterLab workflow uses `jupyterlab/maintainer-tools/.github/actions/install-minimums@v1`.

PR #13219, opened 2022-10-11
Update `pytest` and `pytest-check-links` dependencies

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13209

## Code changes

Update `test` dependencies:

- `pytest`
- `pytest-check-links`

## User-facing changes

None

None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,8 +101,8 @@ docs-screenshots = [
 ]
 test = [
     "jupyterlab_server[test]>=2.11.1,<3",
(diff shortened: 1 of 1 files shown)


jtpio:
Haven't investigated more than just bumping some `pytest` dependencies, but this seems to be fixing https://github.com/jupyterlab/jupyterlab/issues/13209 so marking as ready.

fcollonval (approved):
Thanks for looking into this @jtpio

PR #13230, opened 2022-10-12
Removes triggering an event when replacing pasted text

### LINKED ISSUE #13206: Pasting in markdown cells pastes twice (regression in 4.0)
## Description

Pasting in markdown cells pastes twice

## Reproduce

1. Open JupyterLab 4.0 alpha
2. Create markdown cell
3. Paste some text (using a shortcut or shift + right click context menu)
4. See it occur twice

pastes-twice-in-markdown
pastes-twice-bis

## Expected behavior

Pastes once.

## Context

hbcarlos:
## References
Fixes #13206

## Code changes
On AttachmentsCells, we listen for `paste` events and replace the pasted text with the attachment if necessary. When this occurs, we trigger a new change event from codemirror. That's why we were pasting the text twice. This PR prevents triggering a second event.

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -666,7 +666,7 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
    * @param text The text to be inserted.
    */
   replaceSelection(text: string): void {
-    this.editor.dispatch(this.state.replaceSelection(text));
+    this.state.replaceSelection(text);
   }
 
   /**


fcollonval (approved):
Thanks @hbcarlos

JasonWeill:
@hbcarlos Should this be backported to 3.5.x? #13731, filed by @bsdz, notes that the "double paste" bug is present in 3.5.1.

hbcarlos:
Good catch @JasonWeill, thanks for pining for me.
Indeed, it seems to be the same issue. We should backport it to 3.5.x and 3.6.x.

hbcarlos:

JasonWeill:
The relevant function in the `3.5.x` branch is using `this.doc.replaceSelection`:

https://github.com/jupyterlab/jupyterlab/blob/f2db90135429c0e80599e976633a2de5b3aedcc3/packages/codemirror/src/editor.ts#L663-L665

Should I replace this with `this.state.replaceSelection` as in this PR? [The file as exists in `master`](https://github.com/hbcarlos/jupyterlab/blob/e9842aca62b133fd6ae86fd5648f21b8142e5620/packages/codemirror/src/editor.ts) doesn't seem to have this function at all anymore.

@hbcarlos @fcollonval

fcollonval:
I'm not convinced this is the fix. I did not succeed to reproduce #13731 actually.

Issue #13049, opened 2022-09-05
Improve the Gitpod setup


jtpio:
Opening a new issue to track follow-up improvements to the Gitpod setup.

Previous iterations:

- https://github.com/jupyterlab/jupyterlab/pull/9883
- https://github.com/jupyterlab/jupyterlab/pull/12697
- https://github.com/jupyterlab/jupyterlab/pull/13030

@trallard originally posted in #13030:

- [ ] **Add to the developer documentation**

> Also - while there is this work being done to add Gitpod (and it would reduce the friction for Jupyterlab dev and contributions), I would like to suggest adding docs to the developer docs such as https://numpy.org/devdocs/dev/development_gitpod.html
I can confirm that adding Gitpod + documentation has been a game-changer for contributions, sprints and even maintainers workflows for NumPy and SciPy, so worth a shot.

- [ ] **Re-add comments to document prebuilds**

https://github.com/jupyterlab/jupyterlab/pull/13030#discussion_r959324113

https://github.com/jupyterlab/jupyterlab/blob/5ea1bc727db6a79ecba78033a423d731d1d14837/.gitpod.yml#L2-L10

- [ ] **Use a custom Dockerfile**

> Since all these extras to use mamba are being added, I would highly recommend specifying a custom Dockerfile (see https://www.gitpod.io/docs/config-docker#configure-a-custom-dockerfile) this would help keep this config file cleaner and also removes the need to add later the `echo "source /workspace/bin/activate-env.sh" >> ~/.bashrc` command in line 40. 
> Plus, we can ensure that the corresponding environment is always activated by default and reduce the need to have all the `gp sync` related commands. 
> This is the approach currently used in NumPy and SciPy. This setup and leveraging prebuilds ensures that the Gitpod instance is pretty much ready to use once someone starts a new workspace. 

- [ ] **Add a couple more extensions**

> I'd also suggest adding a couple of extensions (i.e. Python language support, prettier) - as it also helps with developer experience (see https://github.com/numpy/numpy/blob/main/.gitpod.yml for a number of extensions added during the prebuild)

trallard:
I can take a stab at this. From the previous conversation:

> Just checked the config on the numpy repo: 
> [https://github.com/numpy/numpy/blob/main/.gitpod.yml](https://github.com/numpy/numpy/blob/main/.gitpod.yml?rgh-link-date=2022-09-02T09%3A02%3A30Z), and it looks very clean.

[...]

PR #13239, opened 2022-10-13
Enable RTC by default when starting JL in the Gitpod setup

firai:
## References

https://github.com/jupyterlab/jupyterlab/issues/13049#issuecomment-1277740779

## Code changes

Enable RTC by default when JupyterLab is started as part of Gitpod setup. The users sidebar shows up in JL in the test environment, and other RTC features can be tested once the 8888 port is made public in Gitpod.

## User-facing changes

None

None


jtpio (approved):
Thanks!

PR #12548, opened 2022-05-09
Fixes user package

hbcarlos:
The tests in the user package don't work and are not activated in the CI.

This PR fixes the tests.

PR #12095, opened 2022-02-21
RTC left panel

martinRenou:
## References

Rework some parts of https://github.com/jupyterlab/jupyterlab/pull/11180, somewhat related to https://github.com/jupyterlab/jupyterlab/pull/11676

cc. @hbcarlos @SylvainCorlay @wolfv 

## Code changes

Add two new plugins in the user-extension: 

- rtcGlobalAwareness: adds a global YJS awareness for tracking who is connected and who is working on which Notebook/file
- rtcLeftPanel: adds a left panel showing the current user and who is connected 

## User-facing changes

collaborators

## Backwards-incompatible changes

None

PR #12046, opened 2022-02-14
Parse URL parameters in user model

hbcarlos:
Parse URL parameters without lib0 as suggested by @krassowski here: https://github.com/jupyterlab/jupyterlab/pull/11852#issuecomment-1011994165

## References
follow-up #11852

## Code changes
Removes lib0 from the user package.

PR #11774, opened 2021-12-31
Drop familyName from the ICurrentUser interface

hbcarlos:
This PR implements the solution commented by @ellisonbg in https://github.com/jupyterlab/jupyterlab/issues/11657#issuecomment-996338877

## References
Solves #11657 

## Code changes
Changed `givenName` and `familyName` to `name` and `displayName`. Also removed the `getInitials` function instead, for the default anonymous user's name for example "Anonymous Metis" the initials are calculated as "AM", If users manually change their names they will have to introduce the initials.

PR #13242, opened 2022-10-14
Backporting IUser/Collaboration PRs

martinRenou:
## References

Backports of all PRs concerning real-time collaboration UI changes (left panel, top menu) #11443 #11605 #11703 #11774 #12046 #12175 #12548 #12095


jtpio:
Probably need https://github.com/jupyterlab/jupyterlab/pull/12586 or related to align the user menu to the right. Although this might be breaking for some folks / extensions already using the top area in 3.x.

martinRenou:
You're commenting at the right time, I was looking for this! Thanks.

Do you think there is a way to do this in a non-breaking way?

jtpio:
Probably the only "breaking" change would be visually with the addition of the spacer widget in the top area.

This might affect extensions that add widgets to the top area, similar to https://github.com/jtpio/jupyterlab-system-monitor.

Not sure we should do anything clever to handle this, or just document this change in a `3.5.x` -> `3.6.x` migration (similar to https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-3-0-to-3-1).

Another option would be to leave the user menu indicator on the left hand side as is and not backport https://github.com/jupyterlab/jupyterlab/pull/12586.

martinRenou:
CI failing with `error Couldn't find any versions for "@jupyterlab/collaboration" that matches "~3.5.0-beta.0"` as expected because it's not available yet.

tsconfigdoc.json
```diff
@@ -280,6 +286,9 @@
     {
       "path": "./packages/ui-components-extension"
     },
+    {
+      "path": "./packages/user-extension"
+    },
```
fcollonval, line 291: False positive integrity change

```suggestion
```

packages/application-extension/schema/top-bar.json
```diff
@@ -0,0 +1,75 @@
+{
+  "title": "Top Bar",
+  "description": "Top Bar settings.",
+  "jupyter.lab.toolbars": {
+    "TopBar": [
+      {
+        "name": "spacer",
+        "type": "spacer",
+        "rank": 50
+      }
```
fcollonval, line 10: I would not add the spacer here. And prefer to register a new widget in the collaboration-extension to the toolbar that create a spacer to be added with the user menu only in collaborative mode. This should help not impacting the UI in non-collaborative mode.

```suggestion
```

fcollonval (changes requested):
Thanks @martinRenou 

I have some suggestion

packages/application-extension/schema/top-bar.json
jtpio: > only in collaborative mode

Right that would be a good compromise.

jtpio:
> CI failing with `error Couldn't find any versions for "@jupyterlab/collaboration" that matches "~3.5.0-beta.0"` as expected because it's not available yet.

Probably we can publish the package from this branch as is to make the CI pass.

jtpio:
Just published `@jupyterlab/collaboration` and `@jupyterlab/collaboration-extension` to npm to help make the CI pass:

- https://www.npmjs.com/package/@jupyterlab/collaboration
- https://www.npmjs.com/package/@jupyterlab/collaboration-extension

jtpio:
Thanks Marin for doing all these backports!

PR #13249, opened 2022-10-15
Fix illegible white on yellow text of stacktrace in dark theme

### LINKED ISSUE #13248: In dark theme highlighted error stacktrace is white-on-yellow which is practically unreadable
## Description

In dark theme highlighted error stacktrace is white-on-yellow which is practically unreadable.

image

## Reproduce

1. Choose dark theme Settings -> Theme -> JupyterLab Dark
2. Create new notebook
3. Create code cell with function which raises error
4. Execute cell

## Expected behavior

Highlighted stacktrace should be readable, e.g.

image

NikolayXHD:
## References

Fixes #13248

## User-facing changes

image

### DIFF
--- a/packages/theme-dark-extension/style/theme.css
+++ b/packages/theme-dark-extension/style/theme.css
@@ -15,3 +15,8 @@ pre {
   font-size: var(--jp-code-font-size);
   line-height: var(--jp-code-line-height);
 }
+
+/* fix illegible yellow backround in exception stacktrace */
+.jp-RenderedText pre .ansi-yellow-bg {
+  background-color: #554422!important;
+}
--- a/packages/theme-dark-extension/style/theme.css
+++ b/packages/theme-dark-extension/style/theme.css
@@ -17,6 +17,8 @@ pre {
 }
 
 /* fix illegible yellow backround in exception stacktrace */
-:where(.jp-RenderedText[data-mime-type='application/vnd.jupyter.stderr'] pre .ansi-yellow-bg) {
+:where(.jp-RenderedText[data-mime-type='application/vnd.jupyter.stderr']
+    pre
+    .ansi-yellow-bg) {
   color: black;
 }
--- a/packages/theme-dark-extension/style/theme.css
+++ b/packages/theme-dark-extension/style/theme.css
@@ -17,6 +17,6 @@ pre {
 }
 
 /* fix illegible yellow backround in exception stacktrace */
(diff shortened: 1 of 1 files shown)


krassowski:
Thanks for looking into this.  While this fixes the problem of white text on yellow background:
- the proposed selector will match outputs other than stacktrace and distort the expected behaviour (of yellow markers producing yellow background),
    - instead we could target the traceback with `jp-RenderedText[data-mime-type="application/vnd.jupyter.stderr"]` selector
- it introduces a problem of red text on red background (see screenshot below)
   - instead we could keep the yellow as yellow, but make the text black if in yellow highlight in dark mode; the problem is that our selector would then have a precedence over foreground colour classes erasing their effect
     - therefore we need to use `:where` selector to reduce the specificity to zero

Screenshot from 2022-10-15 14-26-13

So, a plausible solution would be:

```css
:where(.jp-RenderedText[data-mime-type="application/vnd.jupyter.stderr"] pre .ansi-yellow-bg) {
  color: black;
}
```

which would produce:

Screenshot from 2022-10-15 14-45-22


This seems better to me. Yet, I am not fully convinced on that one either; an alternative would be to swap the yellow markers for another colour when parsing the stacktrace.

Test code:

<details>
```python
class Test:
    def __init__(self, *args, **kwargs):
        raise Exception()

Test(
    a=11111,
    b="can you read me?",
    c=True is False,
    d=[i for i in range(1)],
    e={**{}},
    f=Exception
)
````

</details>

File: packages/theme-dark-extension/style/theme.css [hunk 1]
```diff
@@ -15,3 +15,8 @@ pre {
   font-size: var(--jp-code-font-size);
   line-height: var(--jp-code-line-height);
 }
+
+/* fix illegible yellow backround in exception stacktrace */
+.jp-RenderedText pre .ansi-yellow-bg {
+  background-color: #554422!important;
+}
```
krassowski, line 22:
```suggestion
:where(.jp-RenderedText[data-mime-type='application/vnd.jupyter.stderr'] pre .ansi-yellow-bg) {
  color: black;
}
```

krassowski (changes requested):
As per comment above.

NikolayXHD:
Agree, white -> black foreground solution looks much better overall. Also thank you for figuring out the issue of specificity.

fcollonval (approved):
Thanks @NikolayXHD

fcollonval:
CI failures are not related

PR #13255, opened 2022-10-16
Remove some unused CSS styles and fix icon alignment in plugin list

krassowski:
## References

Fixes #13001 (misaligned icons)

Removes unused styles:
- `.jp-DirListing-deadSpace` unused since https://github.com/jupyterlab/jupyterlab/commit/10537f4c8c69c0c9edf7cc94cb5f5c737e750822
- `.jp-openJSONSettingsEditor` unused since https://github.com/jupyterlab/jupyterlab/commit/94cc691d9eda5d9bae5c2abd6eb53d2c7cd53f87
- `.jp-OutputArea-stdin` unused since https://github.com/jupyterlab/jupyterlab/commit/c7b8f418cdcb9c245513b84e8cd864701f40db8a
- `.jp-PluginList .jp-SettingsHeader` unused since https://github.com/jupyterlab/jupyterlab/commit/6dc35dac7779da22f99a295016fa97484a208550
- `.jp-PluginList-Searcher` unused since https://github.com/jupyterlab/jupyterlab/commit/3f65d5519892f600032ab4c0d7a6176481b65496
- `.jp-RedirectForm` unused since https://github.com/jupyterlab/jupyterlab/commit/555d84b4175840f675cdfe0d99ea1c56ba39ed21
- `.lm-TabBar-WindowTabLabel` unused since (? it was added in 2016)

Removed rules to without traceable use:
- `.jp-Notebook-render *`
- `.jp-mimeType-highlight`

## Code changes

Fixes lettercase in `jp-pluginList-entry-label` class name (to `jp-PluginList-entry-label`).

## User-facing changes

| Before | After |
|------|--------|
| Screenshot from 2022-10-17 00-13-20 | Screenshot from 2022-10-17 00-12-57 |

## Backwards-incompatible changes

Class name with lettercase error could be retained on backport or not.

PR #13256, opened 2022-10-16
Check for unused file browser style rules in Galata

krassowski:
## References

This is a proposal to start testing for presence of unused style rules at runtime in Galata, follow up to #13255.

Limitations:
- since I only see a moderate number of unused styles in the codebase, this might not be needed
- for now I only added one test case to demonstrate the capability
- rules matching pseudo-elements cannot be tested with this approach
- rules matching pseudo-classes are time consuming to write tests for (we would need to set hover/focus states and re-run the checks) and are therefore not tested by default;
   - for similar reasons rules including `lm-mod-` or `jp-mod-` classes are not tested by default

## Code changes

- Add new galata page helper (`StyleHelper`) with `findUnusedStyleRules(options)` and `collectAllSelectors()` methods.
- On runtime check for presence of unused styles (the test case is expected to fail until [#13255](https://github.com/jupyterlab/jupyterlab/pull/13255) is merged)

## User-facing changes

None

None


fcollonval (approved):
Thanks a lot @krassowski 

This will definitely help keeping the code in good shape. And it will allow extension authors to reuse it.

fcollonval:
@krassowski should we backport it to 3.6.x as it may flag additional unused styles or is it too noisy due to Blueprint dependency for example?

PR #13268, opened 2022-10-18
Set `isUntitled` to false on document path changes

### LINKED ISSUE #13263: Rename dialog shows up on save even if file was already renamed
## Description

The rename dialog introduced in #12953 shows up regarldess of whether user already renamed the file or not.

## Reproduce

1. Create a notebook or file
2. Rename the file (e.g. from file browser or tab)
3. Introduce a change in the file
4. Save the `Rename file` dialog

## Expected behavior

`Rename file` dialog only shows up if file was not already renamed.

First guess is that `isUntitled` does not get set to `false` after rename triggered by other code paths.

## Context

- Browser and version: Chrome
- JupyterLab version: HEAD fc3f6ab716ef82944ab6c4e5d1a06418daad49dc (4.0.0a30+); since this feature was backported to 3.5 it would be great to fix it before release if it is just a quick fix

fcollonval:
## References

Fix #13263

## Code changes

Set `isUntitled` to false as soon as the path changes

## User-facing changes

Renaming an untitled file outside of the dialog prompt will avoid prompting for renaming at first save.

N/A

### DIFF
--- a/packages/docregistry/src/default.ts
+++ b/packages/docregistry/src/default.ts
@@ -562,6 +562,8 @@ export class DocumentWidget<
     path: string
   ): void {
     this.title.label = PathExt.basename(sender.localPath);
+    // The document is not untitled any more.
+    this.isUntitled = false;
   }
 
   /**
@@ -591,6 +593,15 @@ export class DocumentWidget<
   }
 
   readonly context: DocumentRegistry.IContext<U>;
+
+  /**
+   * Whether the document has an auto-generated name or not.
+   *
+   * #### Notes
+   * A document has auto-generated name if its name is untitled and up
+   * to the instant the user saves it manually for the first time.
+   */
+  isUntitled?: boolean;
 }
 
 export namespace DocumentWidget {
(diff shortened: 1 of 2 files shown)


fcollonval:

packages/docregistry/src/default.ts
```diff
@@ -562,6 +562,8 @@ export class DocumentWidget<
     path: string
   ): void {
     this.title.label = PathExt.basename(sender.localPath);
+    // The document is not untitled any more.
+    this.isUntitled = false;
```
krassowski, line 566: Good idea. Though when user moves the new file to another directory (using file browser, without renaming it) it will also set `isUntitled` to `false`. Maybe the following would work?

```js
const newName = PathExt.basename(sender.localPath)
if (newName !== this.title.label) {
  this.isUntitled = false;
}
```

The current fix is sufficient for 3.5 release IMO (better for the new feature to not show up at edge cases than for it to show up when not needed annoying the user).

krassowski: Tracking in #13271.

fcollonval:

PR #13273, opened 2022-10-19
Fix notebook trust in RTC

davidbrochart:
## References

This is for JupyterLab v4.
See #12756.

## Code changes

Propagate cell trust to/from shared model.

## User-facing changes

The notebook trust mechanism now works in RTC.

None.


jtpio (approved):
Thanks!

PR #13279, opened 2022-10-21
Define file or activity icons color as static

HaudinFlorence:
## Code changes

Minor changes in svg files to redefine the fill colors.


krassowski:
+1, but possibly it would be better implemented by adding a new class to allow overriding in themes and avoid exposing `--md` variables which are implementation detail. Maybe even two classes?

HaudinFlorence:
> +1, but possibly it would be better implemented by adding a new class to allow overriding in themes and avoid exposing `--md` variables which are implementation detail. Maybe even two classes?

Thanks for your reply. And what about keeping the orange color that was provided initially instead of using `--md`variables ? About creating new class, the point here was also that we don't want that these icons change when creating new themes (since they are logo at least for the jupyter svg). Can you please tell me more about what you think about when you suggest creating two classes ? Thanks in advance.

krassowski:
> And what about keeping the orange color that was provided initially instead of using `--md` variables ? 

I _think_ that would be better.

> the point here was also that we don't want that these icons change when creating new themes (since they are logo at least for the jupyter svg).

I would rather say that the objective should be to:
- discourage accidental overriding of the logo icon (which should not be overridden unless absolutely necessary, e.g. in a dedicated accessible high-contrast theme)
- decouple the warning class from file icon class to allow tuning the warn class easily (no reason for forbidding to override file icons in themes, I think?)

> Can you please tell me more about what you have in mind when you suggest creating two classes?

Sorry for vague comment. I was thinking one semantic class for notebook and JSON icon and once class for jupyter icon. An alternative would be to create two variables, which would proxy the `--md-orange-700`.

packages/ui-components/style/icons/jupyter/jupyter-favicon.svg
```diff
@@ -1,5 +1,5 @@
 <svg width="152" height="165" viewBox="0 0 152 165" version="1.1" xmlns="http://www.w3.org/2000/svg">
-  <g class="jp-icon-warn0" fill="#F37726">
+   <g class="jp-icon-selectable" fill="#F37726">
```
krassowski, line 2:
```suggestion
   <g class="jp-icon-selectable jp-jupyter-icon-color" fill="#F37726">
```

Was the change to selectable intended here? Was omission of `jp-jupyter-icon-color` intended? I am not saying it is wrong, just asking for rationale. Otherwise it looks good to me.

HaudinFlorence: Thanks for the comment. Both changes were intended. In the last commit I removed the `jp-icon selectable` to set back to the initial case (to be honnest I am not sure to understand what it adds but I had bugs with having only fill and the color value so I had a look at patterns working for other icons) and `jp-jupyter-icon-color` was only introduced in commit [6c470c8](https://github.com/jupyterlab/jupyterlab/pull/13279/commits/6c470c808e13385fce5ff54d0e9d6c161f48e195) in which the ccs variable has been added in the `variables.cs`. I hope it answers your question.

krassowski: Thanks for a quick answer. I still see the `jp-icon selectable` in the last commit. Did you push the last commit? Just to make sure we are on the same page, I see `jp-jupyter-icon-color` added to `jupyter.svg`, but not to `jupyter-favicon.svg`.

HaudinFlorence: Sorry. After checking it appears that I forgot to save my local changes for `jupyter-favicon.svg`so the pushed file was not updated even if it looked ok on my computer. I think that now it should be ok.

packages/theme-light-extension/style/variables.css
```diff
@@ -412,4 +412,9 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-jupyter-icon-color: var(--md-orange-900);
   --jp-notebook-icon-color: var(--md-orange-700);
   --jp-svg-icon-color: var(--md-orange-700);
+  --jp-console-icon-color: var(--md-blue-700);
+  --jp-terminal-icon-color: var(--md-grey-800);
+  --jp-terminal-icon-background_color: var(--md-grey-200);
```
fcollonval, line 417: The convention is to use dash as separator:

```suggestion
  --jp-terminal-icon-background-color: var(--md-grey-200);
```

packages/ui-components/style/icons/filetype/console.svg
```diff
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="16" viewBox="0 0 200 200">
-  <g class="jp-icon-brand1 jp-icon-selectable" fill="#0288D1">
+  <g class="jp-console-icon-color jp-icon-selectable" fill="#0288D1">
```
fcollonval, line 2: We could probably use the same structure as for the terminal icon:
`--jp-console-icon-background-color` for this
`--jp-console-icon-color` for the white parts

packages/ui-components/style/icons/filetype/json.svg
```diff
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="16" viewBox="0 0 22 22">
-  <g class="jp-icon-warn1 jp-icon-selectable" fill="#F9A825">
+  <g class="jp-svg-icon-color jp-icon-selectable" fill="#F9A825">
```
fcollonval, line 2: I guess is more appropriate to use `json` instead of `svg`?

```suggestion
  <g class="jp-json-icon-color jp-icon-selectable" fill="#F9A825">
```

fcollonval (changes requested):
Thanks @HaudinFlorence

packages/ui-components/style/icons/filetype/console.svg
HaudinFlorence: That's right. `jp-console-icon-background-color`has been added and used in `console.svg` in a similar way to what is done with the terminal icon.

packages/theme-light-extension/style/variables.css
HaudinFlorence: This has been corrected in https://github.com/jupyterlab/jupyterlab/pull/13279/commits/87d925f82e1a371d202d8832855287c8b63dd634.

krassowski:
Thank you for adding the semantic variables for other file type icons as well! I think that the last bit remaining is to either:
- adding CSS classes using the added CSS variables (it either was not done, not committed, or I am looking in a wrong plaec), or
- using the `var()` directly in SVG (which would mean that icons would be colourless until the variables are loaded - so we could not due that for that favicon icon - but then this could reduce the number of classes that we have which is a performance consideration)

fcollonval (commented):
Small comment, I would invert `...-color` and `...-background-color`. It feels more natural to say that the colored squared is using `background-color` and the pictogram is using `color`

HaudinFlorence:
> Thank you for adding the semantic variables for other file type icons as well! I think that the last bit remaining is to either:
> 
> * adding CSS classes using the added CSS variables (it either was not done, not committed, or I am looking in a wrong place), or
> * using the `var()` directly in SVG (which would mean that icons would be colourless until the variables are loaded - so we could not due that for that favicon icon - but then this could reduce the number of classes that we have which is a performance consideration)

Thanks for the comment. The first option has been chosen and is implemented in 
6d15db0

HaudinFlorence:
> Small comment, I would invert `...-color` and `...-background-color`. It feels more natural to say that the colored squared is using `background-color` and the pictogram is using `color`

This is done in commit 
6d15db0.

HaudinFlorence:
bot please update snapshots

packages/theme-light-extension/style/variables.css
```diff
@@ -407,4 +407,15 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-reject-color-normal: var(--md-grey-600);
   --jp-reject-color-hover: var(--md-grey-700);
   --jp-reject-color-active: var(--md-grey-800);
+
+  /* File or activity icons semantic variables */
```
krassowski, line 411: We also need to make it work with dark theme, as currently we get a regression:

Screenshot from 2022-10-28 21-22-54

At a minimum we need to:
- add the same set of variables defined for the built-in dark theme
- add an entry in the changelog/extension migration guide documenting that these variables must be defined for themes for JupyterLab 4.0

HaudinFlorence:
bot please update snapshots

packages/theme-light-extension/style/variables.css
HaudinFlorence: Thnask for the comment. An equivalent set of variables has been added in [2528ff6](https://github.com/jupyterlab/jupyterlab/pull/13279/commits/2528ff6350beb733f0239e42896b6af5acf6b1b5)

HaudinFlorence:
Do you think this PR can be merged or there is other things to add from your point of view ?

HaudinFlorence:
bot please update snapshots

krassowski (approved):
Looks good to me, thank you!

HaudinFlorence:
> Looks good to me, thank you!

Ok thanks. I am just adding a small change to take into account the switch that was also defined using `--jp-border-color1`.

fcollonval:

fcollonval:
The last commit change the style of the switch in dark mode:

image

I'm fine with it as it improve contrast for the switch. What do you think @krassowski ?

fcollonval (approved):

krassowski:
Looks ok to me. If users say will say that the new colour is distracting then we can revert to something in between.

fcollonval:
> CI failure is not related

jtpio:
Wondering if this could have potentially introduced breaking changes to third-party themes, as reported in https://github.com/jupyterlab/jupyterlab/issues/13461.

I went over the changelog for `4.0.0a31` and this PR looked the most relevant.

krassowski:
I think it did and therefore we should not include it in 3.6 as-is. I assumed that this is ok to break themes for major releases. Should we be more conservative here?
Another side-effect of this PR seems to be that the Jupyter logo in splash is black for a few seconds, but it is not clear to me why this happens.

HaudinFlorence:
I can make changes that may fix the breaking if we can identify more precisely what's wrong with the PR

jtpio:
> I think it did and therefore we should not include it in 3.6 as-is

Yes the backport should at least be reverted for now. Just tested with the latest 3.6 alpha and it's also an issue there:

image

> I assumed that this is ok to break themes for major releases. Should we be more conservative here?

Right, I was under the impression that theme extension should indeed not break unless very necessary, like mime render extensions.

fcollonval:
We don't need to revert it. @HaudinFlorence could fix this by using a fallback value as done in https://github.com/jupyterlab/jupyterlab/pull/11264/files

fcollonval:
The fallback value should be the one from the team (that was set in the SVG). For example:

```css
.jp-console-icon-background-color[fill] {
-  fill: var(--jp-console-icon-background-color);
+  fill: var(--jp-console-icon-background-color, --jp-brand-color1);
}
```

as the console icon had the class `jp-icon-brand1`

Issue #12928, opened 2022-08-09
Improve cursor placement, display and consistency in documentation screenshots


krassowski:
Mouse cursor on screenshots helps to convey the action depicted.

Currently:
- a few examples have mouse cursor on screenshots
- the cursor is too big (about three lines of text) and not very appealing visually; it should be big-ish but not bigger then 1.5 line of text (the target user with low vision will be zooming to see text anyways)
- the cursor is placed manually to fixed coordinates; instead it could be placed semi-automatically as suggested below:

> I was thinking if we should make the signature something like `position: { x: number | ElementHandle, y: number | ElementHandle }` to allow setting one of the coordinates manually while the other would be in the middle of BBox of the element. We can extract this to a new issue (along with changing the cursor to something nicer, and adding it to more screenshots).

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/12919#discussion_r940709221_

VIVEKANAND-SINHA:
hi, there i am very new to open source and after searching a lot I reached out to this issue as good first issue and tried to make my first contribution, but  i need some help to understand what changes i have to make in this problem and where the changes have to be made as this is a very big repository, kindly help me doing my first contribution

krassowski:
`positionMouse` is defined at:

https://github.com/jupyterlab/jupyterlab/blob/912aceb5a0a08fda578375567b6b2fee936d86c4/galata/test/documentation/utils.ts#L40-L45

Note: after opening the file on GitHub you can click on function names to jump between definition and usage(s).

You could then:
- adjust this function to accept either a number of an `ElementHandle` for each of the coordinates, and if an element is passed, then take the middle of its position (using the `boundingBox` described in the initial post), and/or
- update the SVG string to a smaller cursor image (ideally in public domain) or just decrease the size of the existing one slightly

You can read more about contributing to JupyterLab in the [contributing guide](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html). In particular you may be interested a chapter about [Visual Regression and UI Tests](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests) and [UI Testing developer documentation](https://github.com/jupyterlab/jupyterlab/blob/master/galata/README.md) and [Playwright documentation](https://playwright.dev/docs/intro).

PR #11517, opened 2021-11-22
Allow cross-file anchors with leading number

loichuder:
## References

Fix #11374. Note that this issue was also present for MD files.

## Code changes

The resolution of anchors between files goes through `document.querySelector(anchor)`. This raises errors when the anchor name starts with a number (`2. Test` in #11374)

A [SO thread](https://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers) describes several strategies to go around this. I implemented something using [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) but this is tagged as experimental technology so I am open to suggestions.

The ternary might be simplified: I was unsure if `fragment` was always an anchor (?)

## User-facing changes

Anchors starting with numbers pointing to another file should work for notebooks and markdown files.

## Backwards-incompatible changes

None that I know of.

PR #13285, opened 2022-10-23
Scroll to cell by ID based on hash fragment

krassowski:
## References

Reference implementation for https://github.com/jupyter/nbformat/issues/317.

Fixes a 4.0-only regression in jumping to headings which had CSS-escapeable characters.

Closes https://github.com/jupyterlab/jupyterlab/issues/12928.

## Code changes

- `setFragment` is now always called for notebook on route change, whether in full windowing mode or not
- removes `CSS.escape` which was no longer correct and breaking jumping to headings with characters that would be escaped (it was useful back then when we were using `querySelector`, see #11517, but not since we migrated to manual searching for heading in #12554)

## User-facing changes

- Links to `#cell-id=my-cell-id` will scroll to the cell with such an id.

Links to headings starting with `cell id=` will now need to be updated with `heading=` prefix (I don't think there are many headings like that used in the wild).


krassowski:

krassowski:

krassowski:

krassowski:
Good point. Documentation added, ready for another round of review/merge.

fcollonval (commented):
Thanks a lot @krassowski 

The cell properties side panel snapshot is not sized correctly when updated.

> Lately we are having again side panel sizing flakiness :cry: I think this is linked to https://github.com/jupyterlab/jupyterlab/pull/13037 that changes the workspace restoration logic.

krassowski:

fcollonval:
Kicking the CI

krassowski:
I did not find a quick way to improve that one, so just reverted the assets. The visual tests are passing now but `js-services` are failing - looks unrelated at first glance?

fcollonval:
> looks unrelated at first glance?

I confirm this is not related.

Thanks Mike - merging now.

PR #13284, opened 2022-10-23
Use settings icons for 6 plugins

krassowski:
## References

It became a bit difficult to navigate the settings manager lately as too many entries have the generic cog icon. This PR sets icons for plugins which have an associated icon already (no new icons added).

Follow-up/good first issue would be adding icons for remaining entries (Language settings is a low-hanging fruit).

| Before | After |
|-----|---|
| Screenshot from 2022-10-23 19-22-31 | Screenshot from 2022-10-23 19-23-05 |

## Code changes

None

## User-facing changes

Easier to navigate settings.

## Backwards-incompatible changes

None

PR #13287, opened 2022-10-24
Reduce CI test flake due to kernel switching to busy

krassowski:
## References

An attempt to tackle documentation snapshot failures seen in #13284 and elsewhere earlier.

## Code changes

Await for the Idle > Busy > Idle steps to complete. Ideally we would not have this jitter at all, but that's a bigger undertaking.

## User-facing changes

None

None


jtpio (approved):
Thanks!

PR #13300, opened 2022-10-25
Reduce ILayoutRestorer API surface area

afshin:
This PR reduces the API surface area of `ILayoutRestorer` that was added in #13037 while still keeping the deferred restoration feature, as suggested by @brichet.

## References
Follow on to https://github.com/jupyterlab/jupyterlab/pull/13037

## User-facing changes
N/A

N/A


fcollonval (approved):
Thanks @afshin

fcollonval:
CI failures are not related

PR #13305, opened 2022-10-25
Set font sizes with css variables

HaudinFlorence:
Set some font sizes with css variables instead of having fixed numerical values in a perspective of resizing items when changing the font size, when building a theme editor (as shown on the following gif) . Ideally the user can choose a base font size from which other font sizes are calculated through scale factors. Resizing of the icons or card labels is also performed. First commit replaced the fixed fontsize and line height in the `jp-LauncherCard-label p`  by `var(--jp-ui-fontsize1)`.
card_label_size_changing


packages/launcher/style/base.css
```diff
@@ -163,8 +163,8 @@
   word-break: break-word;
   text-align: center;
   color: var(--jp-ui-font-color1);
-  line-height: 14px;
-  font-size: 12px;
+  line-height: var(--jp-ui-font-size1); /*initially was 14px*/
```
fcollonval, line 166: I would change the `line-height` to be proportional with the `font-size` to keep the default ratio. See https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

fcollonval (changes requested):
@HaudinFlorence would you mind rebasing this PR? and modifying the size so that there is not visual regression (or at least not as large as currently):

image

For now the PR is increasing the size of the cards.

fcollonval (changes requested):
Almost there.

I guess the height or the line height is not proper because the card label is shifted down so much that the second line for the kernel name is half cut. Could you fix that please?

image

fcollonval:
Thanks the last commit makes the change imperceptible.

bot please update snapshots

HaudinFlorence:

fcollonval (approved):

PR #13259, opened 2022-10-17
Don't dispose the notebook metadata editor on active cell change

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/13229

The metadata editor is disposed before loading new values (to avoid polluting undo history). But the changes are disposing the notebook metadata editor on active cell changes (what should not be the case). 

## Code changes

Don't dispose the notebook metadata editor on active cell changes.

## User-facing changes

Notebook metadata editor are not cleared when active cell changes

## Backwards-incompatible changes

N/A

PR #13308, opened 2022-10-25
Updates JSONEditor's source only when there is an active cell or an active notebook panel

hbcarlos:
## References
Follow-up of #13259

## Code changes

Updates JSONEditor's source only when there is an active cell or an active notebook panel.

When removing a cell or closing a notebook, the active cell or active notebook panel changes and the editor tries to set a new source into a disposed editor.


fcollonval (approved):
Thanks @hbcarlos

fcollonval:

PR #13309, opened 2022-10-25
Avoids use of @deprecated to refer to a parameter

### LINKED ISSUE #12937: resolveReact and resolveElement functions in LabIcon incorrectly appear as deprecated in VS Code
In Visual Studio Code, the static `LabIcon.resolveReact` function and the static `resolveElement` function appear struck out, suggesting that they are deprecated.

image

These functions are not deprecated, but their `fallback` parameter is. The `@deprecated` annotation in the docs above a function (see example below) applies to the whole function ([see docs](https://tsdoc.org/pages/tags/deprecated/)) and not to one parameter.
 
```
  /**
   * Resolve an icon name or a {name, svgstr} pair into a DOM element.
   * If icon arg is undefined, the function will fall back to trying to render
   * the icon as a CSS background image, via the iconClass arg.
   * If both icon and iconClass are undefined, this function will return
   * an empty div.
   *
   * @param icon - optional, either a string with the name of an existing icon
   * or an object with {name: string, svgstr: string} fields
   *
   * @param iconClass - optional, if the icon arg is not set, the iconClass arg
   * should be a CSS class associated with an existing CSS background-image
   *
   * @deprecated fallback - don't use, optional, a LabIcon instance that will
   * be used if neither icon nor iconClass are defined
   *
   * @param props - any additional args are passed though to the element method
   * of the resolved icon on render
   *
   * @returns a DOM node with the resolved icon rendered into it
   */
```

There is no TSDoc annotation for "deprecated parameter". [This StackOverflow thread](https://stackoverflow.com/questions/29763919/jsdoc-param-together-with-deprecated) suggests putting a "Deprecated" message in the `@param` annotation.

jweill-aws:
## References

Fixes #12937.

## Code changes

Removes the `@deprecated` annotation in JSDoc, which was meant to indicate that a _parameter_ was deprecated, but actually caused the _function_ in question to be shown as deprecated in Visual Studio Code.

## User-facing changes

None.

None.

### DIFF
--- a/packages/ui-components/src/icon/labicon.tsx
+++ b/packages/ui-components/src/icon/labicon.tsx
@@ -91,7 +91,7 @@ export class LabIcon implements LabIcon.ILabIcon, VirtualElement.IRenderer {
    * @param iconClass - optional, if the icon arg is not set, the iconClass arg
    * should be a CSS class associated with an existing CSS background-image
    *
-   * @deprecated fallback - don't use, optional, a LabIcon instance that will
+   * @param fallback - DEPRECATED, optional, a LabIcon instance that will
    * be used if neither icon nor iconClass are defined
    *
    * @param props - any additional args are passed though to the element method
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @jweill-aws

fcollonval:
CI failures are not related

fcollonval:

PR #13317, opened 2022-10-26
Store document info in the state not in a separate context map out of the document interface.

fcollonval:
## References

To deal with path renaming, a out of interface map was added to the document. It is a bad practice to add new map to the document outside of shared models. So this uses the existing state map for storing those information.

It also extends ISharedDocument interface to access and modify the state.

## Code changes

- Use state Y.Map to store the document `path` and `last_modified` information
- Remove adding out of the model a new map to the document.

The expected side effect is that y-js is only required in `shared-models`, `codemirror` (for the editor binding) and `collaboration` (as it uses a global awarness).

## User-facing changes

N/A

You need to listen to `stateChange` instead of `contextChange` on a `ISharedDocument`.

### DIFF
--- a/packages/docregistry/package.json
+++ b/packages/docregistry/package.json
@@ -58,8 +58,7 @@
     "@lumino/disposable": "^2.0.0-alpha.6",
     "@lumino/messaging": "^2.0.0-alpha.6",
     "@lumino/signaling": "^2.0.0-alpha.6",
-    "@lumino/widgets": "^2.0.0-alpha.6",
-    "yjs": "^13.5.40"
+    "@lumino/widgets": "^2.0.0-alpha.6"
   },
   "devDependencies": {
     "@jupyterlab/testutils": "^4.0.0-alpha.15",
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -23,7 +23,7 @@ import {
   ServerConnection,
   ServiceManager
 } from '@jupyterlab/services';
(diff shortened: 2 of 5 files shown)


davidbrochart:
@fcollonval do you think we should add `last_modified` also in [jupyter_ydoc](https://github.com/jupyter-server/jupyter_ydoc/blob/6dcc9ec61f4b022ec43bdaa7916bd58de481c761/jupyter_ydoc/ydoc.py#L14)'s `state`? Currently it only consists of `dirty` and `path`.

davidbrochart:
Since this PR will conflict with #13246, how do you feel about merging #13246 first and then rebasing this one?

fcollonval:
> @fcollonval do you think we should add `last_modified` also in [jupyter_ydoc](https://github.com/jupyter-server/jupyter_ydoc/blob/6dcc9ec61f4b022ec43bdaa7916bd58de481c761/jupyter_ydoc/ydoc.py#L14)'s `state`? Currently it only consists of `dirty` and `path`.

It sounds right to align the two schemas.

> Since this PR will conflict with https://github.com/jupyterlab/jupyterlab/pull/13246, how do you feel about merging https://github.com/jupyterlab/jupyterlab/pull/13246 first and then rebasing this one?

I left a question there. Then we can merge it before this one.

davidbrochart:
I just tested it, works fine with file renaming in RTC.

davidbrochart (approved):

fcollonval:

Issue #12386, opened 2022-04-09
Deprecate installing source extensions with `jupyter labextension install`


jtpio:
Opening a separate issue to track some of the points mentioned in https://github.com/jupyterlab/jupyterlab/issues/11336.

### Problem

Based on the discussion in https://github.com/jupyterlab/jupyterlab/issues/11336, having to manage both source and prebuilt extensions adds extra maintenance costs.

Also installing prebuilt extension has been a better experience to users, since it doesn't require Node to be installed on the end user machine, and no extra JupyterLab build.

Because of this we should encourage explicitly installing prebuilt extensions to move towards the changes proposed in https://github.com/jupyterlab/jupyterlab/issues/11336.

### Proposed Solution

This issue is about adding the deprecation warnings to the commands.

Adding deprecation warnings in 4.0 will make it possible to drop the feature in the next 5.0 major version.

### Additional context

Refer to https://github.com/jupyterlab/jupyterlab/issues/11336 for more context.

PR #13321, opened 2022-10-26
Deprecate managing source extensions with `jupyter labextension`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12386

## Code changes

Log a deprecation warning to the console when using the following commands:

- `jupyter labextension install`
- `jupyter labextension update`
- `jupyter labextension uninstall`

Also includes other formatting changes since `black` was run on the file.

## User-facing changes

Users managing extensions via the command line with the `jupyter labextension` commands will get the following warning:

image

None

### DIFF
--- a/jupyterlab/labextensions.py
+++ b/jupyterlab/labextensions.py
@@ -204,8 +204,7 @@ class InstallLabExtensionApp(BaseExtensionApp):
         self.deprecation_warning(
             """Installing extensions with the jupyter labextension install command is now deprecated and will be removed in a future major version of JupyterLab.
 
-Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to
-distribute their extensions as Python packages.
+Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to distribute their extensions as Python packages.
 """
         )
         pinned_versions = self.pin.split(",")
@@ -334,8 +333,7 @@ class UpdateLabExtensionApp(BaseExtensionApp):
         self.deprecation_warning(
             """Updating extensions with the jupyter labextension update command is now deprecated and will be removed in a future major version of JupyterLab.
 
-Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to
-distribute their extensions as Python packages.
+Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to distribute their extensions as Python packages.
 """
         )
         if not self.all and not self.extra_args:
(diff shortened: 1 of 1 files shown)


jtpio:
Adding to the `3.6.0` milestone for consideration. If we plan to remove the `install / update / uninstall` commands in JupyterLab 5, the deprecation notice would then be in 3.6 and 4.x spanning 2 major versions.

fcollonval:
This PR was discuss at the JupyterLab weekly meeting.

No objection were raised to move forward introducing deprecation warning.

jupyterlab/labextensions.py
```diff
@@ -304,6 +330,12 @@ class UpdateLabExtensionApp(BaseExtensionApp):
     all = Bool(False, config=True, help="Whether to update all extensions")
 
     def run_task(self):
+        self.deprecation_warning(
+            """Updating extensions with the jupyter labextension update command is now deprecated and will be removed in a future major version of JupyterLab.
+
+Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to distribute their extensions as Python packages.
```
krassowski, line 336:
```suggestion
Users should manage prebuilt extensions with package managers like pip and conda, and extension authors are encouraged to distribute their extensions as prebuilt packages.
```

I would try to avoid "Python extensions" phrase as this gives an incorrect impression that "JupyterLab only supports Python". For example IHaskell distributes their prebuilt extensions without any Python at all:  https://github.com/IHaskell/IHaskell/issues/1238#issuecomment-907658217, which was the intent behind having the extensions system support arbitrary source of prebuilt packages.

This also applies to the same part below and and above. For consistency, the repetitive part could possible be moved to a constant variable to make adjustments easier.

jtpio: Thanks for the suggestion. I think "Python" was used here because the new extension manager in 4.0 now shows up as "PyPI manager" by default.

Agree we can switch to prebuilt and make some of these strings easier to update later on :+1:

krassowski (approved):
Thanks!

jtpio:

jtpio:
ok meeseeks, will do the backport manually.

bollwyvl:
Fire. Glad this statement has been made.

I still think 4.x would be an excellent opportunity to split all the nodejs-relevant commands/features into a sidecar package, e.g. `jupyterlab-nodejs-tools`, which would be a hard dependency, with shims, during the 4.x line, and an opt-in/`[extra]` in 5.x. 

This would also solve some chicken-and-egg problems like https://github.com/jupyterlab/jupyterlab_pygments/issues/23.

jtpio:
Thanks @bollwyvl.

I opened https://github.com/jupyterlab/jupyterlab/issues/13456 as a follow-up since it might indeed be possible to address this for 4.0.

PR #13335, opened 2022-10-28
Add mdformat plugins

blink1073:
## References
Follow up to #13322, cf https://github.com/jupyterlab/jupyterlab/pull/13322#pullrequestreview-1160139441

## Code changes
Add some mdformat plugins to allow GitHub syntax.

## User-facing changes
None

None

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,6 +51,8 @@ repos:
     rev: 0.7.16
     hooks:
       - id: mdformat
+        additional_dependencies:
+          [mdformat-gfm, mdformat-frontmatter, mdformat-footnote]
 
   - repo: https://github.com/sirosen/check-jsonschema
     rev: 0.19.1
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -51,6 +51,8 @@ repos:
     rev: 0.7.16
     hooks:
       - id: mdformat
+        additional_dependencies:
+          [mdformat-gfm, mdformat-frontmatter, mdformat-footnote]
```
jtpio, line 55: Wondering if it's possible to add these as `dev` dependencies.

With `prettier` it was possible to format markdown files automatically in IDEs like VS Code which was quite handy. I don't know if it's also done automatically with `mdformat` or if it needs additional IDE configuration?

blink1073: Perhaps we should stick with Prettier for all but the changelog?

jtpio: Yeah maybe. Anyway we can have a look separately.

jtpio (approved):
Thanks!

PR #13337, opened 2022-10-28
Backport PR #12926 on branch 3.6.x (User service)

fcollonval:
Backport PR #12926


packages/services/src/manager.ts
```diff
@@ -134,6 +138,11 @@ export class ServiceManager implements ServiceManager.IManager {
    */
   readonly terminals: TerminalManager;
 
+  /**
+   * Get the user manager instance.
+   */
+  readonly user: UserManager.IManager;
```
jtpio, line 144: Shouldn't this be `UserManager` for consistency with the other managers above?

packages/services/src/manager.ts
```diff
@@ -221,6 +230,11 @@ export namespace ServiceManager {
      */
     readonly terminals: Terminal.IManager;
 
+    /**
+     * The user manager for the manager.
+     */
+    readonly user: UserManager.IManager;
```
jtpio, line 236: Shouldn't this be optional to be backward compatible?

fcollonval: I'll do that when backporting #13428

Issue #7384, opened 2019-10-17
Zotero://select     one-click local reference database access


jangenoe:
It is convenient (when preparing research papers) to have quick access (i.e. one-click access) to all references and sources (and the links between them) in my publication database.
I use [Zotero](http://www.zotero.org) as publication database. A typical local one-click link to a paper uses **"zotero://select"**.

This one-click reference access to Zotero works perfect from:
* html files in Firefox
* Markdown files rendered in Firefox using [Keith Robertsons Firefox extension](https://github.com/KeithLRobertson/markdown-viewer) 
* Zotero select links all microsoft office apps (Word, PowerPoint, ...)
* Links in code rendered by Notepad++
*  ...

However, in jupyter lab, the corresponding markdown links **do not work** in Markdown cells  or Markdown files using:
* ` [PriorArt2007](zotero://select/library/items/MLW6CEM6)   `
* ` <a href="zotero://select/library/items/MLW6CEM6">PriorArt2007</a>  `

It does work however when I embed the zotero-select URI in a HTML code-cell in jupyter lab
```
%%HTML
<a href="zotero://select/library/items/MLW6CEM6">PriorArt2007</a>
```

The root cause seems to be the blocking done during the sanatizing of the html, I guess for security reasons in outdated browsers, but can security be the argument, when there is a bypass using `%%HTML`?

Would it be possible to allow users to add additional allowed uri schemes to their local jupyter lab preferences?

Related issues:
* Others have reported similar problems with [papers2](https://github.com/jupyter/notebook/issues/189)  (the papers2 software now integrated in the ReadCube reference management software)
* Once the document goes public, it is easy to replace the link with the corresponding reference using [cite2c](https://github.com/takluyver/cite2c) @takluyver

jasongrout:
Indeed, it looks like we already set some relevant options: https://github.com/jupyterlab/jupyterlab/blob/1d03a3616613da84b9dc9dbe8341d9f898048ea0/packages/apputils/src/sanitizer.ts#L956-L965

I wouldn't be opposed to a user setting listing additional schemes that would be okay.

jangenoe:
Hi Jason,
Thanks for supporting this.

The following line inserted in front of the [line](https://github.com/jupyterlab/jupyterlab/blob/1d03a3616613da84b9dc9dbe8341d9f898048ea0/packages/apputils/src/sanitizer.ts#L964) does the job for me:
```
[...]

PR #9873, opened 2021-02-25
Turn HTML sanitizer into a plugin

ohrely:
## References

https://github.com/jupyterlab/jupyterlab/issues/7384

## Code changes

- Creates a sanitizer plugin in `apputils-extension`.
- Provides new sanitizer plugin to rendermime registry.

## User-facing changes

No visual changes anticipated.

## Backwards-incompatible changes

No backwards-incompatible changes anticipated.

PR #13341, opened 2022-10-30
Add user configuration for additional schemes for the sanitizer plugin

kostyafarber:
## References
This looks to close out #7384 by adding user configuration for the sanitizer plugin for the purpose of adding additional schemes to the sanitizer.

## Code changes

* Add settings to the sanitizer plugin
* Add function to default sanitizer to add user configuration to additional schemes

## User-facing changes

A new option in the advanced settings allowing users to add additional schemes.

localhost_8817_lab

No backwards-incompatible changes anticipated.

### DIFF
--- a/packages/apputils-extension/src/index.ts
+++ b/packages/apputils-extension/src/index.ts
@@ -611,10 +611,34 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
  * The default HTML sanitizer.
  */
 const sanitizer: JupyterFrontEndPlugin<ISanitizer> = {
-  id: '@jupyter/apputils-extension:sanitizer',
+  id: '@jupyterlab/apputils-extension:sanitizer',
   autoStart: true,
   provides: ISanitizer,
-  activate: () => {
+  requires: [ISettingRegistry],
+  activate: (app: JupyterFrontEnd, settings: ISettingRegistry) => {
+    const loadSetting = (setting: ISettingRegistry.ISettings): void => {
+      const schemesArray = setting.get('additionalSchemes')
+        .composite as Array<string>;
+
+      schemesArray.forEach(scheme => {
+        if (defaultSanitizer.addScheme) {
+          defaultSanitizer.addScheme(scheme);
+        }
+      });
+    };
+
+    // Wait for the application to be restored and
+    // for the settings for this plugin to be loaded
+    Promise.all([
+      app.restored,
+      settings.load('@jupyterlab/apputils-extension:sanitizer')
+    ]).then(([, setting]) => {
+      // Read the settings
+      loadSetting(setting);
+
+      // Listen for your plugin setting changes using Signal
+      setting.changed.connect(loadSetting);
+    });
     return defaultSanitizer;
   }
 };
--- a/packages/apputils/src/sanitizer.ts
+++ b/packages/apputils/src/sanitizer.ts
@@ -447,6 +447,12 @@ export class Sanitizer implements ISanitizer {
     return sanitize(dirty, { ...this._options, ...(options || {}) });
   }
 
+  addScheme(scheme: string): void {
+    if (Array.isArray(this._options.allowedSchemes)) {
+      this._options.allowedSchemes.push(scheme);
+    }
+  }
+
   private _options: sanitize.IOptions = {
     // HTML tags that are allowed to be used. Tags were extracted from Google Caja
     allowedTags: [
@@ -952,7 +958,8 @@ export class Sanitizer implements ISanitizer {
     // 'src' Attributes are validated to be URIs, which does not allow for embedded (image) data.
     // Since embedded data is no longer deemed to be a threat, validation can be skipped.
     // See https://github.com/jupyterlab/jupyterlab/issues/5183
-    allowedSchemesAppliedToAttributes: ['href', 'cite']
+    allowedSchemesAppliedToAttributes: ['href', 'cite'],
+    allowedSchemes: []
   };
 }
 
--- a/packages/apputils-extension/schema/sanitizer.json
+++ b/packages/apputils-extension/schema/sanitizer.json
@@ -1,0 +1,17 @@
+{
+  "title": "HTML Sanitizer",
+  "description": "HTML Sanitizer settings.",
+  "jupyter.lab.setting-icon": "ui-components:html5",
+  "additionalProperties": false,
+  "properties": {
+    "additionalSchemes": {
+      "title": "Additional Scheme",
+      "description": "Add an additional scheme to the sanitizer.",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    }
+  },
+  "type": "object"
+}
--- a/packages/apputils/src/tokens.ts
+++ b/packages/apputils/src/tokens.ts
@@ -206,6 +206,7 @@ export interface ISanitizer {
    * @returns The sanitized string.
    */
   sanitize(dirty: string, options?: ISanitizer.IOptions): string;
+  addScheme?(scheme: string): void;
 }
 
 /**


packages/apputils-extension/src/index.ts
```diff
@@ -611,10 +611,34 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
  * The default HTML sanitizer.
  */
 const sanitizer: JupyterFrontEndPlugin<ISanitizer> = {
-  id: '@jupyter/apputils-extension:sanitizer',
+  id: '@jupyterlab/apputils-extension:sanitizer',
```
krassowski, line 614: While it is fixing what seems to be an unintentional typo (which was there since this plugin was introduced in https://github.com/jupyterlab/jupyterlab/pull/9873), it is a potentially breaking change. We may need to target 4.0 because of that.

I checked for other occurrences of this issue, and only found one more in tests: https://github.com/jupyterlab/jupyterlab/blob/83b12ea7359a1f547585b2ea2b16b48ad0c10e27/packages/notebook/test/widget.spec.ts#L161.

kostyafarber: Okay fair enough. Is there anything else I need to do on my end?

packages/apputils-extension/src/index.ts
```diff
@@ -611,10 +611,34 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
...
-  activate: () => {
+  requires: [ISettingRegistry],
+  activate: (app: JupyterFrontEnd, settings: ISettingRegistry) => {
+    const loadSetting = (setting: ISettingRegistry.ISettings): void => {
+      const schemesArray = setting.get('additionalSchemes')
+        .composite as Array<string>;
+
+      schemesArray.forEach(scheme => {
+        if (defaultSanitizer.addScheme) {
+          defaultSanitizer.addScheme(scheme);
```
fcollonval, line 625: We need something slightly different here because if the user changes the schemes from the settings this will append the full new list. That means, if a scheme is preserved, it will be added an additional time and if it is removed, it won't actually be as the previous values are kept.

We could rather define a `setSchemes`. The trouble is that since this is public API of `ISanitizer` another plugin can mess up with the settings.

So I would go for implementing a `setSchemes` and to not add anything to the public API (aka the interface `ISanitizer` in the file `token.ts`)

packages/apputils-extension/src/index.ts
fcollonval: In my opinion we should correct the id name on master (aka for 4.0.0) and not do it when backporting this PR.

fcollonval (commented):
Thanks @kostyafarber I left some comments

packages/apputils-extension/src/index.ts
kostyafarber: Ah okay fair enough. Sorry I'm a bit new to JuypterLab development. Would implementing `setSchemes` mean creating a new class that extends `Sanitizer` so we wouldn't have to modify the interface?

kostyafarber: ??

krassowski: This should be possible without creation new class. The interface specifies a subset of class API, if a class has a method which is not in the interface it is ok

kostyafarber: If I try to add `setSchemes` in `Sanitizer` at `https://github.com/jupyterlab/jupyterlab/blob/master/packages/apputils/src/sanitizer.ts` and call it in `https://github.com/jupyterlab/jupyterlab/blob/master/packages/apputils-extension/src/index.ts` I get a:  

```ts
Property 'setScheme' does not exist on type 'ISanitizer'.ts(2339)
```

(sorry if I'm missing something super obvious)

krassowski: No, it's me who missed that `defaultSanitizer` is casted to `ISanitizer` in:

https://github.com/jupyterlab/jupyterlab/blob/a059144fd7946f1f953ffec8d9b56baf283d5cf7/packages/apputils/src/sanitizer.ts#L962

Some ways I could quickly think of:
- add a new interface with only `setSchemes` method, say `ISchemaConfigurable` and mark `defaultSanitizer` as implementing a union of `ISchemaConfigurable` and `ISanitizer`, or
- cast the `defaultSanitizer` to `Sanitizer` type with `as` (I think we should avoid it if possible)
- add an alias for `defaultSanitizer` in `apputils/src/sanitizer.ts` which would not be casted to `ISanitizer`

Maybe @fcollonval has a better idea though.

kostyafarber: Ah okay. Thanks for the quick reply. I'll look into implementing something on the back off those points and `await` @fcollonval response if there are any other suitable ideas. Thanks! @krassowski

fcollonval: For this PR targeting `master`, we should get ride of `defaultSanitizer` in favor of the token based sanitizer - otherwise we will get inconsistency for people using a different plugin or (as it will happen with this PR) non-default settings.

You will instantiate a `new Sanitizer` (after removing the default one) in this `activate` function. So that you get access the full API (including the `setSchemes`) but the token consumer should not rely on it as it is not exposed in the interface.

---

Unfortunately the default sanitizer is use in the code in two places:

1. in the `RenderMimeRegistry`:

https://github.com/jupyterlab/jupyterlab/blob/60491d8185d86597d3f1882b277824635b60fc93/packages/rendermime/src/registry.ts#L41

There you can simply replace it with a new instance of the sanitizer

```patch
- options.sanitizer ?? defaultSanitizer;
+ options.sanitizer ?? new Sanitizer();
```

1. in the completer

https://github.com/jupyterlab/jupyterlab/blob/60491d8185d86597d3f1882b277824635b60fc93/packages/completer/src/widget.ts#L1013

That one is more tricky because you gonna need to provide the sanitizer as option to the completer constructor and pass it from the token in `updateCompleter` calls in three extensions: `console-extension`, `notebook-extension` and `fileeditor-extension` (look for `updateCompleter` to find them).

kostyafarber: Thanks @fcollonval I'll try have a crack at it.

kostyafarber: @fcollonval Had a crack at it. Wasn't exactly sure how to pass it in the `updateCompleter` calls but was able to add it in the constructor to the `Completer`.

kostyafarber: Leaving this here. Let me know if there is anything else I need to do. Thanks.

fcollonval (changes requested):
@kostyafarber thanks for pushing forward.

Here is the last big piece. I tried to describe in details what is needed. Let me know if some parts are unclear.

packages/completer/src/widget.ts
```diff
@@ -65,8 +65,14 @@ export class Completer extends Widget {
     this.model = options.model || null;
     this.editor = options.editor || null;
     this.addClass('jp-Completer');
+    this.sanitizer = options.sanitizer ?? new Sanitizer();
```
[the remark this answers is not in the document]
kostyafarber: @fcollonval thanks for the reply. I really appreciate the support on this one and the detailed points!

I've implemented the changes. As always let me know if I need to add anything.

packages/completer/src/widget.ts
```diff
@@ -1010,7 +1021,8 @@ export namespace Completer {
       const matchNode = document.createElement('code');
       matchNode.className = 'jp-Completer-match';
       // Use innerHTML because search results include <mark> tags.
-      matchNode.innerHTML = defaultSanitizer.sanitize(result, {
+      const sanitizer = new Sanitizer();
+      matchNode.innerHTML = sanitizer.sanitize(result, {
```
krassowski, line 1025: Can we use the newly added `this.sanitizer` here to avoid creating thousands of instances of Sanitizer?

krassowski (commented):
In addition to code changes, we should add a short note to [extension migration guide](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst) in documentation to highlight the changes around sanitizer (importantly, how to replace the old `defaultSanitizer`):

packages/completer/src/widget.ts
kostyafarber: Sure thing on adding a note to the [extension migration guide](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst). I'll try to write something up.

Will change to `this.sanitizer`, it isn't ideal to keep instantiating new sanitizer's. Thanks for the pickup.

kostyafarber: > Can we use the newly added `this.sanitizer` here to avoid creating thousands of instances of Sanitizer?

`this` in this instance refers to 'IRenderer' which does not have a `sanitizer` attribute. Should I define one there?

krassowski: Yes, please do add it to `IRenderer` as well.

packages/notebook-extension/src/index.ts
```diff
@@ -1777,7 +1778,8 @@ function activateNotebookCompleterService(
     const completerContext = {
       editor: notebook.content.activeCell?.editor ?? null,
       session: notebook.sessionContext.session,
-      widget: notebook
+      widget: notebook,
+      sanitizer: new Sanitizer()
```
krassowski, line 1782: Can we hoist the constructor here (and similar cases) a level higher so that we do not create a new instance of sanitizer on each `updateCompleter()` (etc) call? I think that only one instance of the Sanitizer should suffice for the entire notebook-extension, one instance for fileeditor-extension, and the same for console.

kostyafarber: Sure thing sorry about that. Bad practise. I'll update it.

packages/apputils-extension/src/index.ts
```diff
@@ -611,11 +611,38 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
...
+  requires: [ISettingRegistry],
+  activate: (app: JupyterFrontEnd, settings: ISettingRegistry) => {
+    const userSanitizer: Sanitizer = new Sanitizer();
+    const loadSetting = (setting: ISettingRegistry.ISettings): void => {
+      const schemesArray = setting.get('additionalSchemes')
+        .composite as Array<string>;
+
+      if (userSanitizer.setSchemes) {
+        userSanitizer.setSchemes(schemesArray);
+      }
```
krassowski, line 626: This is almost complete now! The visual regression tests are failing because valid links are no longer shown and instead get sanitised away. This is because the schemes which were previously implicitly allowed by default (`[ 'http', 'https', 'ftp', 'mailto', 'tel' ]`) are no longer allowed. I would suggest to rename  `additionalSchemes` to `allowedSchemes` and in the JSON schema add a default of `[ 'http', 'https', 'ftp', 'mailto', 'tel' ]`; I would also suggest setting this array on `allowedSchemes` instead of an empty one.

packages/apputils/src/sanitizer.ts
```diff
@@ -952,11 +958,7 @@ export class Sanitizer implements ISanitizer {
     // 'src' Attributes are validated to be URIs, which does not allow for embedded (image) data.
     // Since embedded data is no longer deemed to be a threat, validation can be skipped.
     // See https://github.com/jupyterlab/jupyterlab/issues/5183
-    allowedSchemesAppliedToAttributes: ['href', 'cite']
+    allowedSchemesAppliedToAttributes: ['href', 'cite'],
+    allowedSchemes: []
```
krassowski, line 962:
```suggestion
    allowedSchemes: ['http', 'https', 'ftp', 'mailto', 'tel']
```

(also see https://github.com/jupyterlab/jupyterlab/pull/13341/files#r1014741922).

packages/apputils-extension/src/index.ts
krassowski: In case if you wonder, this comes from upstream from [here](https://github.com/apostrophecms/sanitize-html/blob/ba3a2f67e3fdcefb7212b8776111ad6f248b52cb/index.js#L814)

krassowski: The alternative is to keep `additionalSchemes` as is but make `setSchemes` concatenate the default (from an immutable source of truth) with the additional. In that case it would probably make sense to name it `setAdditionalSchemes`. The former approach however also allows users to remove existing schemes giving them more control (e.g. I can see why someone could remove `http` but keep `https`).

packages/apputils/src/sanitizer.ts
kostyafarber: Awesome nearly done! Thanks for making those changes. In this case we are going to extend off the defaults and restrict user configurability if I understand correctly (pushed my changes below).

In terms of the note in [extension migration guide](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst) not sure exactly what level of detail is needed to explain the changes made to the default sanitizer (as they were quite a lot of changes but we want to keep the note short).

packages/console-extension/src/index.ts
```diff
@@ -907,7 +908,7 @@ function activateConsoleCompleterService(
   }
 
   const trans = (translator ?? nullTranslator).load('jupyterlab');
-
+  const sanitizer = new Sanitizer();
```
fcollonval, line 911: Ok we are almost there, now we should not instantiate a new sanitizer (except as fallback) but we should request the application sanitizer.

You need to add an argument `sanitizer: ISanitizer | null` to `activateConsoleCompleterService`:

https://github.com/jupyterlab/jupyterlab/blob/2a689157ebddad2fb32d0d611b349cccd10cb5b6/packages/console-extension/src/index.ts#L900-L905


```suggestion
  const sanitizer = sanitizer ?? new Sanitizer();
```

You will pass the application sanitizer by requesting its token in the plugin:

https://github.com/jupyterlab/jupyterlab/blob/2a689157ebddad2fb32d0d611b349cccd10cb5b6/packages/console-extension/src/index.ts#L211-L217

To do that, modify `optional` to be

```diff
- optional: [ICompletionProviderManager],
+ optional: [ICompletionProviderManager, ITranslator, ISanitizer],
```

> `ITranslator` needs to be added because it is missing (bug :cry:).

You need to do similar chagnes for the file editor and the notebook plugins too.

fcollonval (changes requested):

packages/console-extension/src/index.ts
kostyafarber: Ah okay thanks for the review. I've added those changes now (glad we could pick up some bugs along the way 😆)

fcollonval:
@kostyafarber I pushed a fix for the CI and change a bit the logic to set the full scheme array. So this PR also allows to reduce the scope of the sanitize library.

fcollonval (approved):

kostyafarber:
@fcollonval thank you! Glad we could get this one done. Appreciate all the support on this one @krassowski 🎉

fcollonval:
Documentation UI test failing is relevant -> require a new release of `jupyterlab-geojson` with https://github.com/jupyterlab/jupyter-renderers/pull/287

fcollonval:

kostyafarber:
Thanks for pushing this one forward @fcollonval thanks for your help and review @krassowski 🎉

PR #13346, opened 2022-10-31
Check a core path is actually a package

fcollonval:
## References

This will fix the issue for maintainer that an empty folder is added to `tsconfigdoc.json` although it should not like when switching between Git branches (that only cares about files and not folders).

## Code changes

Check that a `package.json` file exists in a folder to consider it as a core path.

## User-facing changes

N/A

N/A


fcollonval:

PR #13364, opened 2022-11-02
Backport document dirty logic

davidbrochart:
## References

Backport the refactoring of the `dirty` state handling missing in https://github.com/jupyterlab/jupyterlab/pull/13222

> This backport will align the behavior in collaboration mode of 3.6.0 and master.

## Code changes

The state `dirty` of a Y-backed document should only be set by the backend not the frontend.

This removes the public `dirty` attribute from the shared document API.


davidbrochart:
I just tested https://github.com/jupyterlab/jupyterlab/pull/13364/commits/0b04ce9f44ce88a4a0d9db063ab257cdb8ca9eb9, it does fix the issue :+1:

PR #13370, opened 2022-11-03
Add jupyter-server-fileid

fcollonval:
## References

Renaming a file in Binder on master is not working properly as in collaborative mode, we need a file ID service to be robust.

> Needs https://github.com/conda-forge/staged-recipes/pull/20986 

## Code changes

Add the jupyter-server-fileid package to binder.

## User-facing changes

More robust experience when renaming file.

N/A

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - jupyterlab-link-share=0.2
+  - jupyter_server_fileid
   - jupyter-server-proxy
   - matplotlib-base
   - nodejs=14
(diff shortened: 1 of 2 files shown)


binder/environment.yml
```diff
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - jupyterlab-link-share=0.2
+  - jupyter_server_fileid
```
davidbrochart, line 6: BTW, `jupyter_server_fileid` will be a required dependency of `jupyter_server_ydoc`, so this won't be needed: https://github.com/jupyter-server/jupyter_server_ydoc/pull/51.

PR #13390, opened 2022-11-07
Create renderer to render errors

divyansshhh:
## References
jupyterlab/jupyterlab##13277

## Code changes
Creates a renderer to render errors. (Currently WIP)

## User-facing changes
Paths in tracebacks will be clickable for the user and clicking on them will open the file where the error occurred.


packages/rendermime/src/renderers.ts
```diff
@@ -511,38 +522,94 @@ function autolink(content: string): Array<HTMLAnchorElement | Text> {
...
+  anchor.rel = 'noopener';
+  anchor.target = '_blank';
+  anchor.appendChild(document.createTextNode(url.slice(0, len)));
+  return anchor;
+}
+
+function convertToPathAnchor(path: string): HTMLAnchorElement {
+  const anchor = document.createElement('a');
+  anchor.innerText = path.slice(0, path.length);
+  anchor.href = path;
```
krassowski, line 609: UX-wise this may lead to a number of "broken" links because:
- the path may be outside of the JupyterLab root
- there will be false positive matches

One solution would be to initially render the path anchors as `<span>`s which later asynchronously get swapped with anchors after checking that a file under such path can be opened.

packages/fileeditor/src/widget.ts
```diff
@@ -204,6 +204,37 @@ export namespace FileEditor {
...
+    const parseFragments = fragment.split('=');
+
+    // TODO: expand to allow more schemes of Fragment Identification Syntax
+    // reference: https://datatracker.ietf.org/doc/html/rfc5147#section-3
+    if (parseFragments[0] !== '#line') {
+      return;
+    }
+
+    // A range of lines can be provided, only respect first for now.
+    let firstLine = parseFragments[1].split(',')[0];
```
fcollonval, line 224: Reading the norm, we should distinguish the single position and the range case. Because in the range case, the examples state:

>   http://example.com/text.txt#line=10,20
>
>   This URI identifies lines 11 to 20 of the text.txt MIME entity.  If
>   the MIME entity has fewer than 11 lines, it identifies the position
>   after the last line.  If the MIME entity has less than 20 but at
>   least 11 lines, it identifies the range from line 11 to the last line
>   of the MIME entity.
>
>   https://example.com/text.txt#line=,1
>
>   This URI identifies the first line.

So in case of a range the first line should be incremented by one and if undefined 0 should be taken.

fcollonval (commented):
Thanks @divyansshhh 

I have only one major comment about the handling of the range lines fragment.

Thanks a lot for this addition.

packages/fileeditor/src/widget.ts
krassowski: Thank you for catching it. I addressed the undefined range start in 439d735623fc5674be0a7ee5579057e734f48731 and  switched to 0-based positions in 9723fdf45aee993cc4908c8e3b7557db16e1b1c2. After checking the spec I  think there is no difference between the single position vs range, both are 0-based.

packages/rendermime/src/registry.ts
```diff
@@ -386,6 +386,53 @@ export namespace RenderMimeRegistry {
...
+     * Returns `null` if there is no file matching provided path in neither
+     * kernel nor jupyter-server contents manager.
+     */
+    async resolvePath(
+      path: string
+    ): Promise<IRenderMime.IResolvedLocation | null> {
+      // TODO: a clean implementation would be server-side and depends on:
+      // https://github.com/jupyter-server/jupyter_server/issues/1280
+
+      const rootDir = PageConfig.getOption('rootUri').replace('file://', '');
```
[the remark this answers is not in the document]
krassowski: It does look similar, the difference is that `rootUri` is a proper URL (`file:///` prefix), which keeps expanded home and takes care of weird Windows issues, while `serverRoot` does not.

Both derive from `ContentsManager.root_dir`, but `serverRoot` is post-processed  variant coming from `jupyter-server` which forces [collapsing of home path](https://github.com/jupyter-server/jupyter_server/blob/c0833098e8c54b302a865a5fb9e63e0a8d7b2ebc/jupyter_server/serverapp.py#L335-L339) which was a UX choice/compromise worked out in https://github.com/jupyter/notebook/pull/2234

Solving https://github.com/jupyter-server/jupyter_server/issues/1280 would relieve us of this dilemma.

krassowski (approved):
Self-approving despite having contributed to this pull request substantially as this seems to had enough time for discussion and if we want to trial this in 4.1 it is high time to merge.

krassowski:
@fcollonval just checking if we can merge it or if you wanted to have another look?

fcollonval (approved):
Let's merge!

Thanks or the ping @krassowski

linlol:
hello @divyansshhh 

may I confirm that if our code's source path differs with jupyter's root dir

For example, we start jupyter with/root

While our code is stored at/src

Can we still enable this feature?

On my side, I can see the file becomes the a link, but it is not clickable.

krassowski:
@linlol this is discussed in https://github.com/jupyter-server/jupyter_server/issues/1280. There are pull requests open against jupyter-server and jupyter-client implementing an API that would allow frontend to point to right place. Recently jupyter-server team suggested to instead pursue https://github.com/jupyter/jupyter_client/issues/1006 but there was not much follow up. Any help on nudging the jupyter-server team on these issues would be appreciated.

PR #13276, opened 2022-10-19
Link the background colors and the ui font colors in the relevant css.

HaudinFlorence:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13152
This PR replaces #13173
## Code changes

Colors in the interface may change slightly.

PR #13400, opened 2022-11-09
Define css variables for the font colors of inactive elements

HaudinFlorence:
Define css variables for the inactive elements like some of the menu.

This PR is the follow up of PR  #13276 that did not take into account the color font of the menu : the logics applied in PR #13276 is to have the same index in the css variables for the background color and the font color in the interface but it was not applicable to the menu with a satisfying rendering. All these changes are proposed in a perspective of making it easier to change the theme in an automatic way.

## Code changes

In variable.css, new css variables are defined for both the dark and light theme, then used menu.css.

### DIFF
--- a/packages/theme-dark-extension/style/variables.css
+++ b/packages/theme-dark-extension/style/variables.css
@@ -125,6 +125,13 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-ui-font-color1: rgba(255, 255, 255, 0.87);
   --jp-ui-font-color2: rgba(255, 255, 255, 0.54);
   --jp-ui-font-color3: rgba(255, 255, 255, 0.38);
+  --jp-ui-font-color4: rgba(255, 255, 255, 0.24);
+  --jp-ui-font-color5: rgba(255, 255, 255, 0.12);
+
+  /* Font colors for disabled items like for instance in the menu */
+  --jp-ui-font-disabled-color0: var(--jp-ui-font-color3);
+  --jp-ui-font-disabled-color1: var(--jp-ui-font-color4);
+  --jp-ui-font-disabled-color2: var(--jp-ui-font-color5);
 
   /*
    * Use these against the brand/accent/warn/error colors.
@@ -436,8 +443,4 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-switch-color: var(--md-grey-400);
   --jp-switch-true-position-color: var(--md-orange-700);
   --jp-switch-cursor-color: rgba(0, 0, 0, 0.8);
-
-  /* Variables for the menu font color : to highlight the active items */
-  --jp-menu-enabled-font-color: rgba(255, 255, 255, 1);
-  --jp-menu-disabled-font-color: rgba(255, 255, 255, 0.38);
 }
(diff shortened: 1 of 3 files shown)


packages/theme-dark-extension/style/variables.css
```diff
@@ -436,4 +436,8 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-switch-color: var(--md-grey-400);
   --jp-switch-true-position-color: var(--md-orange-700);
   --jp-switch-cursor-color: rgba(0, 0, 0, 0.8);
+
+  /* Variables for the menu font color : to highlight the active items */
+  --jp-menu-active-font-color: var(--jp-ui-font-color0);
+  --jp-menu-inactive-font-color: var(--jp-ui-font-color3);
```
krassowski, line 442: The naming does not seem accurate to me. The active item is `.lm-Menu-item.lm-mod-active`.

```suggestion
  /* Variables for the menu font color */
  --jp-menu-enabled-font-color: var(--jp-ui-font-color0);
  --jp-menu-disabled-font-color: var(--jp-ui-font-color3);
```

packages/theme-dark-extension/style/variables.css
```diff
@@ -125,6 +125,13 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-ui-font-color1: rgba(255, 255, 255, 0.87);
   --jp-ui-font-color2: rgba(255, 255, 255, 0.54);
   --jp-ui-font-color3: rgba(255, 255, 255, 0.38);
+  --jp-ui-font-color4: rgba(255, 255, 255, 0.24);
+  --jp-ui-font-color5: rgba(255, 255, 255, 0.12);
+
+  /* Font colors for disabled items like for instance in the menu */
+  --jp-ui-font-disabled-color0: var(--jp-ui-font-color3);
+  --jp-ui-font-disabled-color1: var(--jp-ui-font-color4);
+  --jp-ui-font-disabled-color2: var(--jp-ui-font-color5);
```
krassowski, line 134: +1 I like the generic scope better than the previous menu-specific one.

packages/theme-dark-extension/style/variables.css
```diff
@@ -125,6 +125,13 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-ui-font-color1: rgba(255, 255, 255, 0.87);
   --jp-ui-font-color2: rgba(255, 255, 255, 0.54);
   --jp-ui-font-color3: rgba(255, 255, 255, 0.38);
+  --jp-ui-font-color4: rgba(255, 255, 255, 0.24);
+  --jp-ui-font-color5: rgba(255, 255, 255, 0.12);
```
krassowski, line 129: Are those used anywhere?

HaudinFlorence: Sorry for the late answer. After retesting this PR, the conclusion is that the visual result is not good enough and a bit counter intuitive in some cases with the inactive elements more visible than active ones, like for instance on this screen shot (elements in dark blue grey are inactive ones):

Screenshot from 2023-01-11 17-01-19.

So I propose to close this PR.

Issue #9399, opened 2020-11-25
Accessibility Issues Needing Addressing for WCAG 2.1 compliance (As of Version 2.2.6)


manfromjupyter:
# Accessibility Issues Needing Addressing for WCAG 2.1 compliance
Tested on **JupyterLab V. 2.2.6**, because couldn’t grab current most up to date version. The command `pip3 install --pre jupyterlab` to grab upcoming release’s version worked but still used 2.2.6 for whatever reason. Maybe because I did a `pip install jupyterlab` first, oh well. Found I could use `pip install jupyterlab==3.0.0rc10` but it was after I had already completed this, oh well. Next time.

## JupyterLab Accessibility Summary by Handicap for Version 2.2.6
-	**Blind Users** – completely inaccessible. Grade F.
-	**Low Vision** – hard to use to completely inaccessible. Grade D.
-	**Color Blind** – Out-of-the-box functionality is for the most part accessible with exception of coding which is mostly inaccessible. With the setting of additional contrast settings, it does get better but still needs addressing. Grade C.
-	**Ambulatory** - hard to use to completely inaccessible. Grade D.
-	**Deaf/Hard of Hearing** - completely supported. JupyterLab is 100% compliant as far as I know. I saw videos that get pulled into the app through the documentation sections but they wouldn’t play so don’t know if they have audio-descriptive closed captions. Just leaving at passed for now because in my test, technically, deaf users and non-deaf users would have had the exact same user experience. Grade A.
-	**Cognitive** - completely supported. JupyterLab is 100% compliant. Grade A.
-	**Seizures** - completely supported. JupyterLab is 100% compliant. Grade A.

### Keyboard/Gestures
Issue Area # 1
- **Expectation**: Keyboard can be used to navigate all necessary tasks instead of the mouse. (WCAG Criteria 2.1.1 & 2.1.2 (A))
- **Observed**:
  - Getting into code block won’t let screenreader users leave. Tab adds tab to code block and escape leaves but tabbing again puts them right back in it.
  - Top nav, sidebars, footer nav,  and sub-nav within the File Browser section are not in the tab order and thus can’t be selected using tabs, arrows keys, etc.

Issue Area # 2
- **Expectation**: All actions can be consistently executed by using either the Enter key or Spacebar. (WCAG Criteria 2.1.1 & 2.1.2 (A)).
[...]

PR #13415, opened 2022-11-14
Make focus visible (mostly CSS)

gabalafou:
## References

I'm not sure there is a single issue that corresponds to this PR, but it's related to:

- https://github.com/jupyterlab/jupyterlab/issues/9399

## Code changes

The biggest change introduced by this PR is that I delete the bit of CSS that disables the browser's default focus-visible indicator. [This is a major accessibility anti-pattern.](https://www.a11yproject.com/posts/never-remove-css-outlines/)

After restoring the focus-visible outline, I went about fixing parts of the UI that that change messed up.

I also search-replaced the CSS, replacing `border` with `outline` for focus-visible. [For several reasons, `outline` is a better choice.](https://www.sarasoueidan.com/blog/focus-indicators/#what-exactly-is-a-focus-indicator%3F)

## User-facing changes

Every element that can receive focus should now have a visible outline. The following videos show screen casts of me pressing the tab key in JupyterLab. Each element that receives focus has a visible indicator.

### Safari

https://user-images.githubusercontent.com/317883/201770381-b4a57d7f-b439-4feb-918f-5a533ce9dadf.mov

### Chrome

https://user-images.githubusercontent.com/317883/201770411-eec462ab-c7ed-4746-9256-e2798d3945bd.mov

### Firefox

https://user-images.githubusercontent.com/317883/201770419-e9dfe8f2-3e24-4a0a-8ddb-04f05ba87c6e.mov

The only potentially backward-incompatible change is where I remove the `tabindex` attribute from the main area widget. All the other changes are just style tweaks.


packages/ui-components/style/base.css
```diff
@@ -36,13 +36,6 @@ a:hover {
   color: unset;
 }
 
-/* Disable default focus outline */
-:focus {
```
gabalafou, line 40: There is a risk that this exposes the focus-visible indicator in ways that some users find ugly. But I would prefer to have an ugly focus-visible indicator that prompts aesthetic interventions from users and contributors rather than take away a key accessibility feature from users that need it.

packages/apputils/src/mainareawidget.ts
```diff
@@ -67,7 +67,6 @@ export class MainAreaWidget<T extends Widget = Widget>
     if (!content.id) {
       content.id = DOMUtils.createDomID();
     }
-    content.node.tabIndex = 0;
```
gabalafou, line 70: @marthacryan could you take a look at this change?

After playing for some time with the JupyterLab UI, I couldn't find any good reason to give the main area widget a tab index. Every main area widget that I could find has some element within it that can receive focus, so there's no reason that I can think of to make the container receive focus. I thought maybe it had to do with widgets that only have content and no editing functions (like contextual help). I thought maybe putting a tab index on the container would help the user to use page up and page down keyboard keys to scroll the content, but in fact (on my Mac at least), when the container element has focus I cannot use the page up and page down keys to scroll.

packages/application/style/menus.css
```diff
@@ -38,6 +38,24 @@
...
+}
+
+.lm-MenuBar-menu:focus-visible,
+.lm-MenuBar-item:focus-visible,
+.lm-Menu-item:focus-visible {
+  outline: unset;
+  outline-offset: unset;
+  -moz-outline-radius: unset;
+}
+
```
gabalafou, line 58: With the changes here, the tabbing/focus user experience for the Lumino menubar is not great. The menubar UX needs to be changed at the Lumino level. But my intention with this PR is not to fix all keyboard navigation issues. My intention with this PR is to make focus visible. There are elements that shouldn't receive focus but do, and vice versa. But even if the tabbing order or the set of elements that can receive focus doesn't make sense, that's not a reason to hide the focus indicator; it's a reason to fix which elements receive focus and in which order. At least that's how I see it.

gabalafou (commented):
Done self-reviewing.

packages/application/style/menus.css
krassowski: Would this changes be applicable to lumino too therefore be better included as default lumino styles?

> The menubar UX needs to be changed at the Lumino level.

Does https://github.com/jupyterlab/lumino/pull/465 address the changes needed? Would you be able to help reviewing it?

packages/launcher/style/base.css
```diff
@@ -113,6 +113,8 @@
 .jp-LauncherCard:focus:not(:active) {
   border: 1px solid var(--jp-brand-color1);
   box-shadow: var(--jp-elevation-z6);
+  outline: unset; /* if outline is not unset, then depending on which browser you use, the
+    border change is either hidden behind the outline or visually combined with it */
```
krassowski, line 117: What is the rationale for adding the outline if we already have another visual indicator of focus?

krassowski:
This surfaces an issue with the file browser ignoring focused elements. I think that we would either need to rewrite file browser listing to use browser focus (may be hard), or make the items unfocusable and handle it all programatically (as we currently do).

Actually since the file browser listing can have 100s of files, those should probably be only focused (regardless of implementation detail) only after explicit user confirmation. The alternative is adding a skip-link. I think that the former would be better UX.

packages/application/style/menus.css
gabalafou: Oh neat! I'll take a look at that PR today, thanks @krassowski!

The problem with the Lumino menubar is that it renders an HTML structure like this:

```html
<div tabindex="0">
  <div tabindex="0">
    <ul>
      <li tabindex="0">
```

So the user has to tab through an outer container, then an inner container, and then finally each item in the menu bar (File, Edit, etc.). I think that ideally the menubar component should output a single DOM node with tabindex=0 so that the end user can navigate to the menubar, and then from there use arrow keys to access the different parts of the menubar. I'm hoping that's what jupyterlab/lumino#465 does.

packages/launcher/style/base.css
gabalafou: I'm not sure I follow. I'm removing the outline here. This was one of the places that I had to fix as a result of removing the bit of CSS that disables the outline:

```css
/* Disable default focus outline */
:focus {
  outline: unset;
  outline-offset: unset;
  -moz-outline-radius: unset;
}
```

gabalafou:
@krassowski thanks! You are 100% spot on. My intention was to roll out a number of follow-up PRs to this one that address some of the issues that this PR surfaces (such as the file browser ignoring focussed elements). I wanted to keep this PR fairly small and focussed on making focus visible, separate from any other issues around focus. But do you think I should update this PR to address some of these other issues instead of doing it in a separate PR?

For the file browser in particular, my preference would be to make it possible for the user to tab to the directory listing, but then remove the ability to tab to individual files within the list. Instead the user would just use the arrow keys, which is already supported, whereas pressing the tab key would move the user away from the directory listing to another part of the JupyterLab UI. What do you think? Should I go ahead and make that change in this PR or in a follow-up?

krassowski:
I like your plan for the file browser listing. I am happy to approve it as-is (though let's ensure the UI tests pass) as it is already an improvement.

No strong preference for a single large PR vs multiple smaller ones, especially if you are targeting 4.0.

packages/application/style/menus.css
krassowski: I like the idea of simplifying the DOM output of menubar as it could help with performance too.

krassowski:
bot please update snapshots

packages/apputils/src/mainareawidget.ts
krassowski: I wonder why the visual tests fail (and updates cannot be generated https://github.com/jupyterlab/jupyterlab/actions/runs/3503291407). Since all other changes are CSS-only, my suspicion is that galata may depend on `MainAreaWidget` being focusable. This is usually required in order to receive keyboard events. Could we try with

```
content.node.tabIndex = -1;
```

?

fcollonval:
bot please update snapshots

packages/apputils/src/mainareawidget.ts
krassowski: @gabalafou I opened https://github.com/gabalafou/jupyterlab/pull/12. In particular tests hang on:

https://github.com/jupyterlab/jupyterlab/blob/72f89c9e7015dc3a79c6b5e2d6e2e479d9a8b926/galata/src/helpers/activity.ts#L91-L97

and this is visible in the UI as the tabs do not get the blue highlight the way they used to:

focus-problem

krassowski:
I took liberty of switching milestone as I would love to see this in 3.6.

https://github.com/gabalafou/jupyterlab/pull/12 should solve the visual testing regression (which is relevant).

packages/apputils/src/mainareawidget.ts
marthacryan: Oh sorry thought I responded here! If you can't find a good reason to set that tabindex to 0 that's totally fine to remove. I went through and tried to remove a bunch of tabindex being set to -1 and I may have been setting them to 0 for no good reason. Thanks for working on this!

krassowski:
bot please update snapshots

krassowski:
The documentation code needs updating to ensure there is no focus on irrelevant elements as that will confuse the documentation readers. It would also be nice if this PR did not change the dimensions of buttons (as seen on the popup dialog galata documentation screenshot)

fcollonval:
In c5339c5, I updated the documentation test to remove noisy focus highlight for documentation snapshot. I did not modify the sizing change for the button as actually this fixes the jitter of the button switching to having a `0` border to `1px` border when focus due to CSS rules like:

https://github.com/jupyterlab/jupyterlab/blob/c5339c535cd918e2cbc5753e348fe9558258ccb8/packages/application/style/buttons.css#L31

fcollonval (approved):

packages/ui-components/style/base.css
brichet: @gabalafou this changes the settingseditor's input:focus appearance, replacing the light blue gradient with a strong dark blue, similar to the following diff (new on left):



Was that expected or just a consequence ?
I'm asking this to know if I should open a PR to fix this for the settingseditor.

gabalafou: No, that was not specifically intended. Sorry about that! (If you're short on time, just let me know, and I can submit a PR to fix that. I'm quite familiar now with configuring focus states in CSS, so it shouldn't take me much time.)

By the way, do you know why I cannot tab-key to the drop-downs in the settings editor?

brichet: Thanks, I had already something about that. I opened https://github.com/jupyterlab/jupyterlab/pull/13554.

> By the way, do you know why I cannot tab-key to the drop-downs in the settings editor?

No idea, it seems to work on my end, if you mean reaching a drop-down with tab-key.

Issue #9109, opened 2020-09-30
Breakpoints UI


jtpio:
The breakpoints view now shows the breakpoints with the name of their corresponding temporary files:

![image](https://user-images.githubusercontent.com/591645/68696891-1b48de80-057e-11ea-96e5-f8aba0c82d39.png)

Some points to consider:

- Should the breakpoints be grouped per cell, and show something from the cell UI such as the execution count or some text (the first line of the cell for example), if possible?
- Should the breakpoints for previously dumped cells be shown in the UI, even if these cells are not reachable anymore? It might not be obvious to do so, since the files exist on disk and corresponding breakpoints are set in `ptvsd`. Sometimes we can also map back to the same files (editing followed by successive `Ctrl-Z`).

PR #13433, opened 2022-11-17
Allows to pause the execution during debug

brichet:
This PR allows to pause the execution during debugging.

## References

## Code changes

- Rename the previous existing  command `pause` to `pauseOnException`, as this was its intended action
- Add a command `pause` that sends a *pause* request on the control channel.

## User-facing changes

- Add a pause button in the `CALLSTACK` tab of the debugger panel.

   The button is enabled while the script is running and disabled when it is paused (after the button has been pressed or a breakpoints has been reached).

   

   

- Change the `pauseOnException` icon from   to   to avoid confusion with the other one.

None


packages/debugger-extension/src/index.ts
```diff
@@ -682,6 +683,17 @@ const main: JupyterFrontEndPlugin<void> = {
       }
     });
 
+    commands.addCommand(CommandIDs.pause, {
+      label: trans.__('Pause'),
+      caption: trans.__('Pause'),
```
fcollonval, line 688: Let's be more descriptive

```suggestion
      caption: trans.__('Pause the current thread if possible.'),
```

fcollonval (changes requested):
Thanks @brichet 

What would it take to use an approach similar to VS Code:

image

aka using a checkbox in the breakpoints panel to state if we want to pause on exceptions or not.

I understand the need to bring another icon. But that icon does not feel describing really its purpose. Moreover the pause within the triangle is having very thin lines at some places that makes it hard to grasp the symbol.

jtpio:
> What would it take to use an approach similar to VS Code:

If not in this PR this could also be discussed in https://github.com/jupyterlab/jupyterlab/issues/9109

brichet:
> What would it take to use an approach similar to VS Code:

That could be interesting. In VS Code these options are mixed with the existing breakpoints that can be unchecked.
I would prefer a drop down menu to check/uncheck the exception options, not mixed with the breakpoints.

I'm not sure about the way it works currently. It seems that the same `ExceptionOption` is send several times to debugpy:
https://github.com/jupyterlab/jupyterlab/blob/3e8f4e5f87ac315bf95c7c959a9073b18b493778/packages/debugger/src/service.ts#L593-L600
for each existing `exceptionBreakpointFilters` (received from the server), `breakMode` is set the `'userUnhandled' : 'never'` depending on the `enable` status. That result with an array with same object several time...

Having checkbox for each option would clarify (and probably fix) this. I'll have a look at it.

brichet:
> What would it take to use an approach similar to VS Code:

After exploring a bit more, this involves non trivial changes, including a modal to select options.
Perhaps this should be done in a follow-up PR, as that was not the original intent of this one.

For this one:
- I can revert the `pauseOnException` icon for now if the new one is non relevant. Since the buttons are in 2 different tabs the confusion is minor.
- The `pause` button could just replace the `run` button when the script is running, instead of having 2 distinct buttons.

What do you think ?

fcollonval:
> For this one:
> 
>     * I can revert the `pauseOnException` icon for now if the new one is non relevant. Since the buttons are in 2 different tabs the confusion is minor.
> 
>     * The `pause` button could just replace the `run` button when the script is running, instead of having 2 distinct buttons.
> 
> 
> What do you think ?

If you can do the button icon switch let's do that. Otherwise keep this version but improve a bit the icon by taking as much as possible of the SVG size (the triangle could be bigger and then the vertical pause bars could be separated a  bit more).

packages/debugger/src/debugger.ts
```diff
@@ -100,12 +100,14 @@ export namespace Debugger {
...
   export namespace CommandIDs {
-    export const debugContinue = 'debugger:continue';
+    export const pauseOrContinue = 'debugger:pause-continue';
 
     export const terminate = 'debugger:terminate';
 
     export const next = 'debugger:next';
 
+    export const pause = 'debugger:pause';
+
```
fcollonval, line 110: Could you restore the `debugger:continue` command?
And I think the `pause` removal is missing from the latest changes.

This will ease backport as the command ids are part of the public API.

fcollonval (changes requested):

packages/debugger/src/debugger.ts
brichet: Sure, thank for review @fcollonval . I restored the command name in https://github.com/jupyterlab/jupyterlab/pull/13433/commits/093de33867084d6659be9a5f9233fcde93118a7b, but kept the pause feature in this command.

packages/debugger-extension/schema/main.json
```diff
@@ -40,7 +40,7 @@
       "selector": "body"
     },
     {
-      "command": "debugger:continue",
+      "command": "debugger:pause-continue",
```
fcollonval, line 43:
```suggestion
      "command": "debugger:continue",
```

This is highlighted by the failing test.

brichet: Thanks @fcollonval

packages/debugger/src/debugger.ts
```diff
@@ -120,7 +120,7 @@ export namespace Debugger {
 
     export const restartDebug = 'debugger:restart-debug';
 
-    export const pause = 'debugger:pause';
+    export const pauseOnExceptions = 'debugger:pause-on-exceptions';
```
fcollonval, line 123: Final one (sorry for it), could you add a note about this api change in the migration guide:

https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst

fcollonval (approved):

fcollonval:

brichet:
Thanks @fcollonval.
I can backport it manually if necessary

Issue #13440, opened 2022-11-18
enable document model specific collaboration


dlqqq:
### Problem

Custom extensions that add their own file types to the document registry often break when RTC is enabled. It would be much nicer if developers could instead "opt-in" to collaboration by explicitly setting a flag that their file type (and its corresponding widgets and models) support collaboration. That way we preserve the existing behavior for custom file types when RTC is enabled.

### Proposed Solution

Add a `collaborative` optional property to `IFileType`, defaulting to `false`. Then for all native file types that ship with JupyterLab (mainly under `docregistry/src/registry.ts`), we set these to `true`.

### Additional context

None.

dlqqq:
It makes more sense to place the `collaborative` optional on the document model rather than the file type, since different file types often share the same document model (which is guaranteed to either support or not support collaboration). See comments in #13441.

PR #13441, opened 2022-11-18
enable filetype-specific collaboration

dlqqq:
cc @ellisonbg @3coins @davidbrochart 

## References

- https://github.com/jupyterlab/jupyterlab/issues/13440

## Code changes

Forces collaboration to explicitly enabled per file type, even when the collaboration global flag is enabled. This prevents custom extensions that add file types to the document registry from breaking, and forces extension developers to be explicit about enabling collaboration for their widgets.

### Testing

1. Run JupyterLab with the collaboration flag and verify RTC + autosave still works for files.
2. Remove the `collaborative: true` line from `getDefaultNotebookFileType()` in `docregistry/src/registry.ts`.
3. Rebuild JupyterLab and restart the server.
4. Verify that RTC is enabled for text files but not for notebook files.

## User-facing changes

None known.

None known.

### DIFF
--- a/packages/docmanager/src/manager.ts
+++ b/packages/docmanager/src/manager.ts
@@ -2,7 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { ISessionContext, sessionContextDialogs } from '@jupyterlab/apputils';
-import { PathExt } from '@jupyterlab/coreutils';
+import { PageConfig, PathExt } from '@jupyterlab/coreutils';
 import { IDocumentProviderFactory } from '@jupyterlab/docprovider';
 import {
   Context,
@@ -39,7 +39,6 @@ export class DocumentManager implements IDocumentManager {
     this.translator = options.translator || nullTranslator;
     this.registry = options.registry;
     this.services = options.manager;
-    this._collaborative = !!options.collaborative;
     this._dialogs = options.sessionDialogs || sessionContextDialogs;
     this._docProviderFactory = options.docProviderFactory;
     this._isConnectedCallback = options.isConnectedCallback || (() => true);
@@ -487,7 +486,8 @@ export class DocumentManager implements IDocumentManager {
   private _createContext(
     path: string,
     factory: DocumentRegistry.ModelFactory,
-    kernelPreference?: ISessionContext.IKernelPreference
+    kernelPreference: ISessionContext.IKernelPreference | undefined,
+    collaborative: boolean | undefined
   ): Private.IContext {
     // TODO: Make it impossible to open two different contexts for the same
     // path. Or at least prompt the closing of all widgets associated with the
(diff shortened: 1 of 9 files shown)


packages/docregistry/src/registry.ts
```diff
@@ -932,6 +932,12 @@ export namespace DocumentRegistry {
      */
     readonly ready: Promise<void>;
 
+    /**
+     * Returns whether this document context is writable. True only if the
+     * contents model is writable and collaboration is disabled.
+     */
+    readonly writable: boolean;
+
```
dlqqq, line 940: Does adding this constitute a breaking API change? We need to backport this to 3.6.x. Yes, it causes TypeScript errors for plugins providing their own custom document context, but I'm not sure if this should count.

packages/docregistry/src/registry.ts
```diff
@@ -1347,71 +1360,81 @@ export namespace DocumentRegistry {
...
         displayName: trans.__('Image'),
         mimeTypes: ['image/svg+xml'],
         extensions: ['.svg'],
         icon: imageIcon,
-        fileFormat: 'base64'
+        fileFormat: 'base64',
+        collaborative: true
       },
       {
         name: 'tiff',
```
dlqqq, line 1440: I explicitly chose not to enable collaboration on the image formats below. Let me know if you don't think this makes sense.

packages/docmanager/src/manager.ts
```diff
@@ -584,6 +584,21 @@ export class DocumentManager implements IDocumentManager {
       return undefined;
     }
 
+    let fileType: DocumentRegistry.IFileType | undefined = undefined;
+    for (const ft of this.registry.fileTypes()) {
+      if (ft.name === modelName) {
+        fileType = ft;
+        break;
+      }
+    }
```
dlqqq, line 593: @davidbrochart Here I'm assuming that a widget factory's `modelName` is always equivalent to the file type. This seems to hold for all the widget factories I've seen, and it seems to work at least for notebooks and text files. Is this a safe assumption? If not, how else can I retrieve the "current" file type from the created/opened document?

3coins:
@dlqqq 
Thanks for making this change.
Tested this on my local, and I was able to change the `collaborative` flag for "text" files; this correctly disabled collaborative mode for text files, and allowed collaborative mode for notebooks. I will do some more testing, and report back if I see any issues.

fcollonval (commented):
Thanks @dlqqq 

I understand the issue but I'm not convinced this is the correct way of fixing it (I need more info on your use case).

In your use case, are you only adding a new file type or also a model factory?

For custom model factory (and maybe widget factory), it makes sense that the collaborative capability should be provided indeed as the model may not be YJs backed. It makes less sense to place the flag at the file type level; e.g. if the file type is reusing a _default_ model factory from JupyterLab core, why should the extension requires change for RTC.

If you only add a filetype, would you mind describing more specifically 

> often break when RTC is enabled

What errors are you seeing? Or at lease from which part of the code.

---

That said if the only path forward is the one proposed here (aka flag at the file type level), my main comment on the code is the lookup for the file type from `modelName`. They are independent (although it may works it is not a requirement). A better path would be to use `DocumentRegistry.getFileTypesForPath` - but the output may not be a single element (proving that point above that `collaborative` should not be an attribute of the file type).

https://github.com/jupyterlab/jupyterlab/blob/9e94df7feb1408685a4441f442e8a9f897dc0514/packages/docregistry/src/registry.ts#L700

PR #13365, opened 2022-11-02
Add announcements

fcollonval:
## References

Fix https://github.com/jupyterlab/jupyterlab/issues/13110

## Code changes

Add a new plugin to fetch announcements from Atom feed and update check.

## User-facing changes

Add notifications at start up from announcements feed.

announcements

## Backwards-incompatible changes

N/A

PR #13443, opened 2022-11-18
Use default URL in test mock-up

fcollonval:
## References

[Follow-up of #13365](https://github.com/jupyterlab/jupyterlab/pull/13365#discussion_r1025242234)

## Code changes

Nothing

> I updated the date of last modification for the private policy to tomorrow because the policy will change for JupyterLab 4

## User-facing changes

N/A

N/A

### DIFF
--- a/docs/source/privacy_policies.rst
+++ b/docs/source/privacy_policies.rst
@@ -1,7 +1,7 @@
 Privacy policies
 ================
 
-Last modified: November 17th, 2022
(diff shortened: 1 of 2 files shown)


jtpio (commented):
Thanks, this will also make it easier when searching the code base for `jupyterlab.github.io`.

jtpio (approved):

Issue #13439, opened 2022-11-17
JupyterLab extension tutorial error, on open/close/reopen empty panel


ericsnekbytes:
## Description

Possible bug in the [JupyterLab extension tutorial](https://jupyterlab.readthedocs.io/en/stable/extension/extension_tutorial.html#extension-tutorial), during the blank tab close/reopen step at this exact passage in the tutorial page (can copy/paste to find in this spot in the tutorial):

> You should see a new, blank panel appear with the tab title Astronomy Picture. Click the x on the tab to close it and activate the command again. The tab should reappear.

## Reproduce

1. Follow the JupyterLab extension tutorial up to and including the passage "Click the x on the tab to close it and activate the command again. The tab should reappear."
2. Open the javascript console and observe the error: "TypeError: Cannot read properties of null (reading 'widgets')"
3. Observe that a blank tab reappears (presumably the APOD tab?), but is missing both the tab name and the tab close button (X)

![image](https://user-images.githubusercontent.com/104786633/202549322-067b760d-b1b3-4ce4-87f9-89916564d5db.png)

More console output is provided below.

## Expected behavior

The APOD tab should reappear, the tab name should be present, and the tab close button (X) should appear on the tab.

## Context

- Operating System and version: MacOS 12.4
- Browser and version: Version 107.0.5304.110 (Official Build) (arm64)
- JupyterLab version: Version 3.5.0

<details><summary>Troubleshoot Output</summary>
<pre>
/Users/username/workspace_ana/armconda/bin/python: No module named pip
$PATH:
	/Users/username/workspace_ana/armconda/bin
	/Users/username/workspace_ana/armconda/bin
	/Users/username/workspace_ana/armconda/bin
	/opt/homebrew/bin
	/Users/username/.cargo/bin
	/Users/username/workspace_ana/armconda/envs/jupyterlab-ext/bin
	/Users/username/workspace_ana/armconda/condabin
	/usr/bin
	/bin
	/usr/sbin
	/sbin

sys.path:
	/Users/username/workspace_ana/armconda/bin
	/Users/username/workspace_ana/armconda/lib/python39.zip
	/Users/username/workspace_ana/armconda/lib/python3.9
	/Users/username/workspace_ana/armconda/lib/python3.9/lib-dynload
	/Users/username/workspace_ana/armconda/lib/python3.9/site-packages

sys.executable:
	/Users/username/workspace_ana/armconda/bin/python

sys.version:
[...]

PR #13453, opened 2022-11-20
Check if `contentHeader` is not disposed before accessing widgets

krassowski:
## References

When disposed widget gets attached to the main area the error message is misleading for extension developer as it fails on `contentHeader` (see https://github.com/jupyterlab/jupyterlab/issues/13439). This PR fixes that message which surfaces more useful errors.

## Code changes

Check if `contentHeader` is not disposed before accessing `contentHeader.widgets.length`

## User-facing changes

None

None


fcollonval (approved):
Thanks @krassowski

fcollonval:
Should this be backported?

PR #13454, opened 2022-11-21
Fix token based completions, restore deduplication, follow up on completer refactor

### LINKED ISSUE #13386: Token-based completions do not work anymore
## Description

The kernel-less completions in files do not work anymore on 4.0 alpha. This is likely related to CodeMirror 6 migration.

## Reproduce

1. Open a new Python file
2. Enter the following text:
  ```python
  test1 = 1
  test2 = 2
  te
  ```
4. Place cursor at the last character of last line
5. Invoke completer and see no completions

However, if we set cursor on `test2` and press <kbd>tab</kbd> we see a single completion `test2 = 2` (which is useless). We can conclude that the polyfill for tokens in CodeMirror 6 behaves differently than tokens in CM5 and prevents baic completion from working in files when there is no kernel/lsp.

## Expected behavior

Token-based completions work as in 3.x. They do not have to cover all the cases completely since CodeMirror 6 does not give as much data as CM5 did and because we are close to having LSP, but while it is not ready we should not introduce a downgrade for this crucial component.

## Context

Related to https://github.com/jupyterlab/jupyterlab/issues/12812.

- Browser and version: Chrome
- JupyterLab version: 4.0.0a30

krassowski:
## References

Fixes #13386, addresses #13496

## Code changes

Fix behaviour of `CodeMirrorEditor.getTokens` and `CodeMirrorEditor.getTokenAt`:
- only return leaf tokens
- add unit tests
- add galata test

Restore backward compatibility with `ICompletionItemsConnector` and make merging completions the responsibility of `IProviderReconciliator` (previously `IConnectorProxy`):
- replace `IConnectorProxy` with `IProviderReconciliator`,
  - use `IOptions` pattern for `IProviderReconciliator` (allows backward-compatible additions and better docs)
  - remove unused and previously deprecated `IReply`
  - deprecate `ICompletionItemsResponseType`
  - deprecate `ICompleterConnecterResponseType`
  - deprecate `ICompletionItemsConnector`
  - deprecate `connector` setter
  - deprecate `connector` option of `CompletionHandler`
- fix and add unit test for a scenario when one of the providers fails
- de-duplicate replies from multiple providers and add unit tests
 
We may decide to remove the `connector` references entirely for 4.0 release, but let's do that in a separate PR to have a proper discussion and an easy way to revert if needed.

Rename `setShowDocumentFlag` method of `ICompletionProviderManager` to `setShowDocumentationPanel`.

<details>

Previously the results included unwanted non-tokens like:

```json
{
  "offset": 0,
  "type": "AssignStatement",
  "value": "foo = \"a\"",
},
{
  "offset": 10,
  "type": "AssignStatement",
  "value": "bar = 1",
},
{
  "offset": 0,
  "type": "Script",
  "value": "foo = \"a\"\nbar = 1",
}
```

</details>

## User-facing changes

Token-based completions will work again in 4.0 alpha

None, actually fixing breaking changes. All other changes are to non-released code of refactored completer.


packages/codemirror/src/editor.ts
```diff
@@ -692,13 +696,27 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
...
+          // If it has a child, it is not a leaf, but we still want to enter
+          if (node.node.firstChild !== null) {
+            return true;
+          }
+          if (offset >= node.from && offset <= node.to) {
+            token = {
+              value: this.state.sliceDoc(node.from, node.to),
+              offset: node.from,
+              type: node.name
+            };
```
krassowski, line 712:
```suggestion
            };
            // We have just found the relevant leaf token, no need to iterate further
            return false;
```

krassowski:
Of note, the tests are failing because the token completions started working, but the new completion mechanism introduced in completer refactor #11795 does not do a good job de-duplicating:

Screenshot from 2022-11-26 12-15-09

Actually, it does not de-duplicate at all and wrongly assumes that everything will have the same start/end:

https://github.com/jupyterlab/jupyterlab/blob/198d01058290155f7cbbc91519f5d3c15452c6aa/packages/completer/src/handler.ts#L354-L364

krassowski:

krassowski:

krassowski:
Examples failure is unrelated (#13481).

fcollonval:
Kicking the CI

packages/completer/src/reconciliator.ts
```diff
@@ -0,0 +1,241 @@
...
+              : insertText;
+            return item;
+          })
+        };
+      });
+    }
+    return replies;
+  }
+
+  private async _mergeCompletions(
```
trungleduc, line 102: Could you move these two private methods down behind the public methods?

packages/completer/src/reconciliator.ts
```diff
@@ -0,0 +1,241 @@
...
+    for (const reply of replies) {
+      reply.items.forEach(item => {
+        // IPython returns 'import' and 'import '; while the latter is more useful,
+        // user should not see two suggestions with identical labels and nearly-identical
+        // behaviour as they could not distinguish the two either way.
+        let text = (item.insertText || item.label).trim();
+        if (insertTextSet.has(text)) {
+          return;
+        }
+        item.type;
```
trungleduc, line 147:
```suggestion
        
```

trungleduc (commented):
Thanks @krassowski for the fix, it looks really good to me, and I have only some minor suggestions above.

packages/completer/src/reconciliator.ts
```diff
@@ -0,0 +1,241 @@
...
+import { CompletionHandler } from './handler';
+import {
+  ICompletionContext,
+  ICompletionProvider,
+  IProviderReconciliator
+} from './tokens';
+import { Completer } from './widget';
+
+/**
+ * The connector which is used to fetch responses from multiple providers.
```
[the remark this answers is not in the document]
krassowski: I changed `completer providers` → `completion providers`.

packages/completer/src/reconciliator.ts
krassowski: Done in 80286edd99eb15ea656802f4d7c87eeaf1fcec77.

krassowski:
@trungleduc thanks for the review. I applied suggested changes.

trungleduc (approved):

fcollonval (approved):
Thanks @krassowski and @trungleduc

PR #13458, opened 2022-11-21
enable document model specific collaboration

dlqqq:
cc @ellisonbg @3coins @davidbrochart 

## References

- https://github.com/jupyterlab/jupyterlab/issues/13440

## Code changes

Forces collaboration to explicitly enabled per document model, even when the collaboration global flag is enabled. This prevents custom extensions that add document models to the document registry from providing a broken experience by default when RTC is enabled globally, and forces extension developers to be explicit about enabling collaboration for their document models.

### Testing

1. Run JupyterLab with the `--collaborative` flag and verify RTC + autosave still works for files.
2. Comment out the `get collaborative()` method in `packages/notebook/src/model.ts`.
3. Rebuild JupyterLab and restart the server.
4. Verify that RTC is enabled for text files but not for notebook files.

## User-facing changes

None known.

None known.

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -611,12 +611,6 @@ function addCommands(
       return false;
     }
     const context = docManager.contextForWidget(currentWidget);
-    console.log({
-      context,
-      writable: !!context?.writable,
-      // @ts-ignore
-      collaborative: context?._collaborative
-    });
     return !!context?.writable;
   };
 
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -191,7 +191,6 @@ const manager: JupyterFrontEndPlugin<IDocumentManager> = {
       setBusy: (status && (() => status.setBusy())) ?? undefined,
       sessionDialogs: sessionDialogs || undefined,
       translator: translator ?? nullTranslator,
-      collaborative: true,
       docProviderFactory: docProviderFactory ?? undefined,
       isConnectedCallback: () => {
         if (info) {
@@ -612,11 +611,13 @@ function addCommands(
       return false;
     }
     const context = docManager.contextForWidget(currentWidget);
-    return !!(
-      context &&
-      context.contentsModel &&
-      context.contentsModel.writable
-    );
+    console.log({
+      context,
+      writable: !!context?.writable,
+      // @ts-ignore
+      collaborative: context?._collaborative
+    });
+    return !!context?.writable;
   };
 
   // If inside a rich application like JupyterLab, add additional functionality.
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -611,7 +611,7 @@ function addCommands(
       return false;
     }
     const context = docManager.contextForWidget(currentWidget);
-    return !!context?.writable;
+    return !!context?.contentsModel?.writable;
   };
 
   // If inside a rich application like JupyterLab, add additional functionality.
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -59,13 +59,18 @@ export class Context<
     const localPath = this._manager.contents.localPath(this._path);
     const lang = this._factory.preferredLanguage(PathExt.basename(localPath));
     this._model = this._factory.createNew(lang);
+    this._collaborative = !!(
+      PageConfig.getOption('collaborative') === 'true' &&
+      this._model.collaborative
+    );
     const docProviderFactory = options.docProviderFactory;
     this._provider = docProviderFactory
       ? docProviderFactory({
           path: this._path,
           contentType: this._factory.contentType,
           format: this._factory.fileFormat!,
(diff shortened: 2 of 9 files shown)


dlqqq:
This will need to be backported to 3.6.x in an alpha release.

packages/docregistry/src/context.ts
```diff
@@ -59,13 +59,18 @@ export class Context<
     const localPath = this._manager.contents.localPath(this._path);
     const lang = this._factory.preferredLanguage(PathExt.basename(localPath));
     this._model = this._factory.createNew(lang);
```
fcollonval, line 61: We should pass here a new flag telling the factory if collaboration is enabled -> the model should use that information to set the appropriate `model.collaborative` properties.

```suggestion
    this._model = this._factory.createNew(lang, PageConfig.getOption('collaborative') === 'true');
```

packages/docregistry/src/default.ts
```diff
@@ -102,6 +102,10 @@ export class DocumentModel
     return this._defaultLang;
   }
 
+  get collaborative(): boolean {
```
fcollonval, line 105:
```suggestion
  /**
   * Whether the model is collaborative or not.
   */
  get collaborative(): boolean {
```

The returned value should be the value of the to-be-added `collaborationEnabled` argument to be passed to the constructor.

packages/notebook/src/model.ts
```diff
@@ -231,6 +231,10 @@ export class NotebookModel implements INotebookModel {
     return info?.name ?? '';
   }
 
+  get collaborative(): boolean {
```
fcollonval, line 234:
```suggestion  
  /**
   * Whether the model is collaborative or not.
   */
  get collaborative(): boolean {
```

Same about the value here - it should depend on the value of `collaborationEnabled` pass to the constructor.

fcollonval (changes requested):

packages/docregistry/src/context.ts
```diff
@@ -219,6 +224,14 @@ export class Context<
     return this._readyPromise;
   }
 
+  /**
+   * Returns whether this document context is writable. True only if the
+   * contents model is writable and collaboration is disabled.
+   */
+  get writable(): boolean {
+    return !!(this._contentsModel?.writable && !this._collaborative);
+  }
```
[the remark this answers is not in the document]
dlqqq: The problem is that `_contentsModel` is of type `Contents.IModel`, while `_model` is of type `DocumentRegistry.IModel`. So because only the DocumentContext has access to both of these models, it should be the single source of truth for whether this document is writable, which is why I think it makes sense to keep this property here.

packages/docregistry/src/registry.ts
```diff
@@ -932,6 +938,12 @@ export namespace DocumentRegistry {
      */
     readonly ready: Promise<void>;
 
+    /**
+     * Returns whether this document context is writable. True only if the
+     * contents model is writable and collaboration is disabled.
+     */
+    readonly writable: boolean;
+
```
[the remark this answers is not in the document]
dlqqq: See above.

dlqqq:
@fcollonval I addressed all your other comments, but I still think we should consider keeping `writable` as a property on DocumentContext. It is the only object with references to both the document model and content model, and this condition is used in multiple locations throughout the code. I think it's better to keep a single source of truth on the "writability" of a document, rather than imperatively check every condition everywhere.

packages/docregistry/src/context.ts
```diff
@@ -219,6 +223,14 @@ export class Context<
     return this._readyPromise;
   }
 
+  /**
+   * Returns whether this document context is writable. True only if the
+   * contents model is writable and collaboration is disabled.
+   */
+  get writable(): boolean {
```
fcollonval, line 230: Ok then can we make it `protected` to not increase the surface API (as it is used only locally) and could we use another name to avoid misunderstanding with the `this._contentsModel.writable` attribute that is closer to the purpose; something like:

```suggestion
   * Whether the document support saving through the Contents REST API or not.
   */
  protected get isSavingWithContents(): boolean {
```

> I'm not great at wording so feel free to propose something else.

packages/docregistry/src/registry.ts
fcollonval, line 946: Unneeded once we change the visibility

```suggestion
```

packages/docmanager-extension/src/index.tsx
```diff
@@ -612,11 +611,7 @@ function addCommands(
       return false;
     }
     const context = docManager.contextForWidget(currentWidget);
-    return !!(
-      context &&
-      context.contentsModel &&
-      context.contentsModel.writable
-    );
+    return !!context?.writable;
```
fcollonval, line 614: Changed not needed because the `writable` state of the content model is _corrected_ in `Context._updateContentsModel`:

```suggestion
    return !!context?.contentsModel?.writable;
```

fcollonval (changes requested):
Thanks @dlqqq 

I made some additional suggestions to not increase the surface API.

fcollonval:

PR #13465, opened 2022-11-22
Backport #12667 on branch 3.6.x (Add events service)

fcollonval:
Backport #12667

### DIFF
--- a/tsconfigbase.json
+++ b/tsconfigbase.json
@@ -16,7 +16,7 @@
     "resolveJsonModule": true,
     "sourceMap": true,
     "strictNullChecks": true,
-    "target": "es2017",
+    "target": "es2018",
     "types": []
   }
 }
--- a/buildutils/src/create-test-package.ts
+++ b/buildutils/src/create-test-package.ts
@@ -29,6 +29,7 @@ if (require.main === module) {
   const jsonPath = path.join(dest, 'package.json');
   const data = utils.readJSONFile(jsonPath);
   if (name.indexOf('@jupyterlab/') === -1) {
+    // eslint-disable-next-line no-global-assign
     name = '@jupyterlab/test-' + name;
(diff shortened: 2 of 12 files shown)


afshin (approved):

tsconfigbase.json
```diff
@@ -16,7 +16,7 @@
     "resolveJsonModule": true,
     "sourceMap": true,
     "strictNullChecks": true,
-    "target": "es2017",
+    "target": "es2018",
```
fcollonval, line 19: @afshin are we ok to change that for a minor release?

afshin:
I am not sure I fully understand the ramifications. I don’t *think* this is
backward-incompatible but I’m not 100% confident.

tsconfigbase.json
afshin: I am not sure I fully understand the ramifications. I don’t *think* this is backward-incompatible but I’m not 100% confident.

fcollonval: This is backward incompatible in the sense that it raises the required version of web browsers.

afshin: I think in that sense we might be okay because all the [browsers we support are covered](
https://caniuse.com/?feats=mdn-javascript_builtins_regexp_dotall,mdn-javascript_builtins_regexp_lookbehind_assertion,mdn-javascript_builtins_regexp_named_capture_groups,mdn-javascript_builtins_regexp_property_escapes,mdn-javascript_builtins_symbol_asynciterator,mdn-javascript_functions_method_definitions_async_generator_methods,mdn-javascript_grammar_template_literals_template_literal_revision,mdn-javascript_operators_destructuring_rest_in_objects,mdn-javascript_operators_spread_spread_in_destructuring,promise-finally).

Do you think we have a problem?

PR #13470, opened 2022-11-22
Fix syntax highlighting for JSON viewer in Jupyter Notebook

kostyafarber:
## References
This PR addresses https://github.com/jupyter/notebook/issues/6567 in Jupyter Notebook, by making an upstream change here in JupyterLab.

## Code changes

The changes are to the `json-extension` which mount `jupyterHighlightStyle` manually.

## User-facing changes
The users of Jupyter Notebook will see the same as what they see in JupyterLab when opening a JSON file.

### Before
image

### After
image

No anticipated backwards incompatible changes.

### DIFF
--- a/packages/json-extension/src/component.tsx
+++ b/packages/json-extension/src/component.tsx
@@ -9,7 +9,7 @@ import { JSONArray, JSONExt, JSONObject, JSONValue } from '@lumino/coreutils';
 import * as React from 'react';
 import Highlight from 'react-highlighter';
 import { JSONTree } from 'react-json-tree';
-
+import { StyleModule } from 'style-mod';
 /**
  * The properties for the JSON tree component.
  */
@@ -37,6 +37,8 @@ export interface IState {
 function getStyle(tag: Tag): string {
   return jupyterHighlightStyle.style([tag]) ?? '';
 }
+
+StyleModule.mount(document, jupyterHighlightStyle.module as StyleModule);
 
 /**
  * A component that renders JSON data as a collapsible tree.
--- a/packages/json-extension/src/component.tsx
+++ b/packages/json-extension/src/component.tsx
@@ -38,8 +38,6 @@ function getStyle(tag: Tag): string {
   return jupyterHighlightStyle.style([tag]) ?? '';
 }
 
-StyleModule.mount(document, jupyterHighlightStyle.module as StyleModule);
-
 /**
  * A component that renders JSON data as a collapsible tree.
  */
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @kostyafarber for working on this and tracking the issue down :+1:

packages/json-extension/src/component.tsx
```diff
@@ -9,7 +9,7 @@ import { JSONArray, JSONExt, JSONObject, JSONValue } from '@lumino/coreutils';
 import * as React from 'react';
 import Highlight from 'react-highlighter';
 import { JSONTree } from 'react-json-tree';
-
+import { StyleModule } from 'style-mod';
```
jtpio, line 12: Looks like the `style-mod` dependency should be added to the `package.json`, as suggested by the following check failure:

https://github.com/jupyterlab/jupyterlab/actions/runs/3525395365/jobs/5912584760

krassowski:
Is all this stylesheet does adding the green bold colour for keys and a few more colours for numbers/strings/null? Maybe we could just have a dedicated style in this extension rather than bringing another dependency and another copy of the stylesheet?

fcollonval:
> Is all this stylesheet does adding the green bold colour for keys and a few more colours for numbers/strings/null?

FYI: I think it was done to get an identical styles for JSON in CodeMirror editor and in the viewer - that said I don't have cons to move to a new stylesheet.

> Maybe we could just have a dedicated style in this extension rather than bringing another dependency and another copy of the stylesheet?

FYI: this is a new direct dependency. But it is not a new dependency as that package was created for and is needed by CodeMirror 6 (by the same author). Moreover it is carefully designed to not add stylesheet that are not necessary. And as in this case the theme is identical to the default Jupyter CodeMirror theme, it will not add a new stylesheet in JupyterLab page.

packages/json-extension/src/component.tsx
```diff
@@ -38,6 +38,8 @@ function getStyle(tag: Tag): string {
   return jupyterHighlightStyle.style([tag]) ?? '';
 }
 
+StyleModule.mount(document, jupyterHighlightStyle.module as StyleModule);
```
fcollonval, line 41: This is not a good practice as it will mount the stylesheet even if the component is not used. So you should do this in the `componentDidMount` of `Component` (to be added):

```typescript
componentDidMount(): void {
  StyleModule.mount(document, jupyterHighlightStyle.module as StyleModule);
}
```

See [documentation](https://reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class) if you want to know more about React component life cycle

fcollonval (changes requested):

packages/json-extension/src/component.tsx
kostyafarber: Ah okay sorry I missed that. Thanks for the reference to the docs. Will change.

packages/json-extension/src/component.tsx
kostyafarber: Ah my bad your right. Will add.

fcollonval (approved):
Thanks @kostyafarber

jtpio (approved):
Thanks!

The `examples` CI failure is tracked in https://github.com/jupyterlab/jupyterlab/issues/13481

Issue #13463, opened 2022-11-22
[3.6.0a4] Can't create a notebook with RTC enabled


jtpio:
## Description

With `3.6.0a4` on Binder it's not  possible to create a new notebook from the launcher:

https://user-images.githubusercontent.com/591645/203251342-4f744a8d-e506-45a0-bb43-17d2dce533ea.mp4

## Reproduce

Using this gist on Binder which has RTC enabled: https://gist.github.com/jtpio/d063489b0843547ceeaaf5ee9832e69d

## Expected behavior

It should be possible to create new notebooks.

## Context

cc @davidbrochart @hbcarlos

davidbrochart:
Thanks for reporting @jtpio.
It looks like `PUT /api/yjs/roomid/{path}` returns a `404`. I will look into it.

davidbrochart:
I opened https://github.com/jupyter-server/jupyter_server_ydoc/pull/61.

jtpio:
Nice thanks @davidbrochart!

davidbrochart:
https://github.com/jupyter-server/jupyter_server_ydoc/pull/61 was merged and released in `jupyter-server-ydoc` v0.5.0.

davidbrochart:
JupyterLab does a `GET` request to the created notebook before the `PUT /api/yjs/roomid/{path}`, so this is not the issue.

davidbrochart:
So the reply to `PUT /api/yjs/roomid/{path}` is an HTML document with:
```
Jupyter Notebook requires JavaScript.
Please enable it to proceed.
```
@jtpio maybe a clue to what's going on?

jtpio:
> I yanked `jupyter-server-ydoc` v0.5.0 in PyPI

Should the GitHub Release be deleted then? https://github.com/jupyter-server/jupyter_server_ydoc/releases/tag/v0.5.0

davidbrochart:
> Should the GitHub Release be deleted then? https://github.com/jupyter-server/jupyter_server_ydoc/releases/tag/v0.5.0

Thanks, done.

> Is this on Binder? Could it be that it is hitting the classic notebook server and not the proxied JupyterLab Server (backed by a Jupyter Server)?

Yes, on Binder only. It works locally even when I change the server `base_url`.
Yes it could be hitting the classic notebook server, as in the previous release `jupyter-server-ydoc` was previously handling the base URL manually and was compatible with the classic notebook server.

fcollonval:
[...]

PR #13472, opened 2022-11-23
Forbid using RTC mode without Jupyter Server v2.

fcollonval:
## References

Follow-up of RTC rework for 3.x

It will resolve https://github.com/jupyterlab/jupyterlab/issues/13463

## Code changes

Stop initialization of the server in collaborative mode if Jupyter Server is not v2+

## User-facing changes

Forbid using RTC mode without Jupyter Server v2.

Collaboration is conditioned on a new dependency (that don't break other features in non-RTC mode).

### DIFF
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -6,6 +6,7 @@
 
 import json
 import os
+import sys
 from os.path import join as pjoin
 
 from jupyter_core.application import JupyterApp, NoStart, base_aliases, base_flags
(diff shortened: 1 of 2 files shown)


jupyterlab/labapp.py
```diff
@@ -764,6 +765,16 @@ def initialize(self, argv=None):
         """Subclass because the ExtensionApp.initialize() method does not take arguments"""
         super().initialize()
 
+        if self.collaborative and jpserver_version_info < (2, 0, 0):
+            jpserver_version = ".".join(filter(lambda p: p, map(lambda p: str(p), jpserver_version_info)))
```
fcollonval, line 769:
```suggestion
            jpserver_version = ".".join(filter(lambda p: len(p) > 0, map(lambda p: str(p), jpserver_version_info)))
```

hbcarlos (approved):
Thanks, @fcollonval!

Issue #13429, opened 2022-11-17
Enable jupyterlab client packages to be imported in server side code


stevejpurves:
### Problem

Currently is it not possible of difficult to include some `juptyerlab` packages within a code base that use different build systems, or web frameworks. For example,`remix.run` and `next.js` or additionally systems that use alternatives to `webpack` for bundling, such as `esbuild`

Frameworks like this are typically geared up for server side rendering (SSR), which is clearly beyond the scope of `juptyerlab` to support. When using `jupyterlab` components in these applications, they are marked as client side only and rendered in the browser in any case. However, the build process becomes complicated when attempting to include these as dependencies in a build.

The problem to solve / objective here would be to enable easy inclusion of jupyterlab packages as dependencies at build time. i.e. it should be possible to `import` `jupyterlab` packages in a [node] server environment without issues.

This would make it significantly easier for (non lumino based) application developers to simply include the dependencies in their build.

Problems that prevent this include unguarded references to `window`, `document`, `navigator` at "import" time, typically at global file scope or in immediately executed functions. There are instances. of these in:

* `@juptyerlab/apputils`
* `@juptyerlab/notebook`
* `@juptyerlab/ui-components`

as well as dependencies (which may have the be addressed else where, or could possibly be addressed through non static loading here (thinking of codemirror)

* `@lumino/domutils`
* `@juptyer-widgets/base`
* `codemirror` 

Resolving these issues would make it easier to statically import dependencies into their builds.

### Proposed Solution

Add guards to any uses of `window`, `document`, `navigator` that have to happen at file scope, providing appropriate default values/behaviours or restructure code such that initialization depending on these happens later, on invocation of an initialize function etc...

stevejpurves:
I have local changes based on the `3.5.x` branch addressing some of this - happy to open a PR, but should the be on `3.6.x` or `main`?

jtpio:
Thanks @stevejpurves for reporting.

Linking to https://github.com/jupyterlab/jupyterlab/pull/12519 which should help address some of these points.



> should the be on `3.6.x` or `main`?

Usually PRs target `master` (not renamed to `main` yet), and are then backported to other branches.

stevejpurves:
Opened a draft PR to help discussion on this 👆🏻

[...]

PR #13475, opened 2022-11-23
Add guards for browser globals accessed at import time

stevejpurves:
## References

This relates to issue #13429 and is meant to be illustrative / assist triage.

## Code changes

The code changes here are minimal and were required in order to build `thebe` in a web application built on `remix.run`. Remix has a closed `esbuild` based build process and performs SSR. The objective is here is not to enable SSR with jupyterlab components, but just to allow them to be built/packages by remix (or similar build processes/platforms) with minimum friction.

The code changes here may not be suitable for server side use of the components, as tsome code will not be called when `window` or `document` is not present.

The changes here stem from initialization of objects at import time / file scope instantiation of variables / objects -- so a longer term / better change would be to address how the classes / objects here are initialized and move that initialisaiton to happen at application start rather than import.

## User-facing changes

none

Unknown, but there may be impact from not executing certain code when expected.

### DIFF
--- a/packages/apputils/src/windowresolver.ts
+++ b/packages/apputils/src/windowresolver.ts
@@ -110,6 +110,7 @@ namespace Private {
    * Start the storage event handler.
    */
   function initialize(): void {
+    if (typeof window === 'undefined') return;
     // Listen to all storage events for beacons and window names.
     window.addEventListener('storage', (event: StorageEvent) => {
       const { key, newValue } = event;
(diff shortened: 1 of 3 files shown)


krassowski:
> so a longer term / better change would be to address how the classes / objects here are initialized and move that initialisaiton to happen at application start rather than import.

Agreed.

@stevejpurves would you like to proceed with this PR adding workarounds, or is it meant to remain for illustrative purposes? We could consider it for 4.0 (and revert if any concerns arise from the side of extension authors during porting), but it might be harder to get this (either a workaround or refactor of initialisation) in a point release.

krassowski:
If the workarounds were to be merged each line with a workaround would need a comment describing why this happens and maybe even emit a `console.warn`/debug/info (not sure) message for ease of debugging if someone relied on this working without `window` being defined.

stevejpurves (commented):
added comments to explain addition guards on browser-only api objects

stevejpurves:
@krassowski do those comments suffice or are they too simplistic?

PR #13476, opened 2022-11-23
Copy variable in kernel from debugger panel

brichet:
This PR adds a contextual menu entry to copy a local variable to the global scope from the debugger panel.

## References

> There should be a trivial way to copy a variable into the kernel from the UI

## Code changes

- adds a `setVariableInGlobal` method in `DebuggerService` to create a variable in the global scope of the kernel.
- adds a command and a context menu entry (on non-global variables only) in `debugger-extension` to use this method.
- adds tests on this feature.

None


packages/debugger/src/service.ts
```diff
@@ -329,6 +329,36 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+   */
+  async setVariableInGlobal(name: string, value: string): Promise<void> {
+    if (!this.session) {
+      throw new Error('No active debugger session');
+    }
+    const frames = this.model.callstack.frames;
+    this.session
+      .sendRequest('setExpression', {
+        expression: name,
+        value: value,
```
fcollonval, line 346: Candid question but is this answering the need?

I tried with a complex object and it failed:

image

Note: I have no idea if it is possible to copy complex object (especially in a language independent way).

galata/test/documentation/debugger.test.ts
```diff
@@ -157,15 +157,48 @@ test.describe('Debugger', () => {
...
+    // Expect the context menu for local variables to match snapshot
+    const variable = await page
+      .locator('.jp-DebuggerVariables-body li:first-child')
+      .textContent();
+    await page.click('.jp-DebuggerVariables-body li:first-child', {
+      button: 'right'
+    });
+    expect(await page.locator('.lm-Menu-content').screenshot()).toMatchSnapshot(
+      'debugger_variables_context-locals.png'
+    );
```
fcollonval, line 181: Could you test for text content rather than snapshot? It will make the test more robust and faster.

galata/test/documentation/debugger.test.ts
```diff
@@ -157,15 +157,48 @@ test.describe('Debugger', () => {
...
+    await page.locator('select[aria-label="Scope"]').selectOption('Globals');
+    await expect(
+      page.locator(`.jp-DebuggerVariables-body li:text("${variable}")`)
+    ).toHaveCount(1);
+
+    // Expect the context menu for global variables to match snapshot
+    await page.click(`.jp-DebuggerVariables-body li:text("${variable}")`, {
+      button: 'right'
+    });
+    expect(await page.locator('.lm-Menu-content').screenshot()).toMatchSnapshot(
```
fcollonval, line 198: Same here

fcollonval (changes requested):

packages/debugger/src/service.ts
brichet: Candid answer: maybe not :smile: 
I haven't tried it with a complex object, but in this implementation the value is sent in a request, so it will probably never work with complex objects. I'll see in the `debugpy` API if there is something similar on the kernel side.

brichet: I haven't found anything like that in the DAP specifications.
I understand that the only way would be to evaluate some code, which would not be generic for languages.
This PR can be closed because it doesn't actually meet the need.
Sorry for your time @fcollonval

krassowski: One way to think about it is that we are already making addition to DAP (e.g. `inspectVariables` see https://jupyter-client.readthedocs.io/en/latest/messaging.html#additions-to-the-dap), but copying variables is a generic operation which ideally would be supported in the main DAP.
It looks that DAP only supports special evaluation with aim of copying to clipboard (`clipboard` context); this PR could be reworked to support that use case instead (which would still be an improvement for some scenarios, but not as good as the ability to copy to global namespace).

Can you open an issue over at https://github.com/microsoft/debug-adapter-protocol to see if DAP would be happy to add a "copy to another frame" request and link back from here so we can track it?

brichet: Thanks for suggestion @krassowski. I opened an issue [here](https://github.com/microsoft/debug-adapter-protocol/issues/360)

galata/test/documentation/debugger.test.ts
brichet: Thanks, fixed in https://github.com/jupyterlab/jupyterlab/pull/13476/commits/abdd2cdd68c86bb20c3d126852722b490f515a5c

galata/test/documentation/debugger.test.ts
brichet: Fixed in https://github.com/jupyterlab/jupyterlab/pull/13476/commits/abdd2cdd68c86bb20c3d126852722b490f515a5c

packages/debugger/src/service.ts
brichet: I converted the PR to draft as it will now wait for https://github.com/ipython/ipykernel/pull/1055 to be merged and released (an probably also https://github.com/jupyter-xeus/xeus-python/pull/562), which both wait for the next release of `debugpy`, which should integrate https://github.com/microsoft/debugpy/pull/1147.

brichet: > It looks that DAP only supports special evaluation with aim of copying to clipboard (`clipboard` context); this PR could be reworked to support that use case instead (which would still be an improvement for some scenarios, but not as good as the ability to copy to global namespace).

Thanks, I'll probably open an other PR for this feature.

packages/debugger/src/service.ts
```diff
@@ -327,6 +327,35 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+   *
+   * @param name The name of the variable.
+   */
+  async copyToGlobals(name: string): Promise<void> {
+    if (!this.session) {
+      throw new Error('No active debugger session');
+    }
+    const frames = this.model.callstack.frames;
+    this.session
+      .sendRequest('copyToGlobals', {
```
krassowski, line 341: Since this relies on non-standard method, we need to be very careful here and do one or both:
- detect support (adding `supportsCopyToGlobals`) and suggest this feature upstream and only condition `isEnabled` on support
- enhance the error handling beyond just printing reason to console; we need to tell user that this is requires `copyToGlobals` method which is implemented in IPython >{version number} and xeus-python and to use it they need to upgrade or ask the third-party kernel developer to consider adding support for it.

We also need to document it (in the repository where all other custom methods are listed).

brichet: Thanks @krassowski.
I already opened a PR to document it: https://github.com/jupyter/jupyter_client/pull/932
It seems difficult to handle the error as the request is always successful (status=200), even if the copy does not occur.
The first suggestion is probably better. I'll see on `ipykernel` and `xeus-python` how to add this information.

brichet: I opened https://github.com/jupyter/enhancement-proposals/pull/93

packages/debugger/src/tokens.ts
```diff
@@ -601,6 +611,12 @@ export namespace IDebugger {
       variables: DebugProtocol.VariablesResponse;
     };
 
+    export interface ICopyToGlobalsArguments {
```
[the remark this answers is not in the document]
brichet: Thanks @krassowski , I added docstring in https://github.com/jupyterlab/jupyterlab/pull/13476/commits/eda2f4ba55e8bebfb91f2c8a08bab3438df88afb

packages/debugger/src/service.ts
brichet: I added a flag on the `copyToGlobals` support from the kernel.
I also opened https://github.com/ipython/ipykernel/pull/1099 and https://github.com/jupyter-xeus/xeus-python/pull/582.
We will have to wait until the `ipykernel` one is released to be able to make the tests.

brichet:
The tests do not depend on the kernel version (and its support to *copyToGlobals*) anymore.
This should be ready for merge now, except if we want to wait for https://github.com/jupyter/enhancement-proposals/pull/93.

brichet:
The UI documentation test failed on the input prompt of the `activeCellTool`, it should not be related to the current changes.

galata/playwright.config.js
```diff
@@ -31,7 +28,12 @@ module.exports = {
     {
       name: 'jupyterlab',
       testMatch: 'test/jupyterlab/**',
-      testIgnore: '**/.ipynb_checkpoints/**'
+      testIgnore: '**/.ipynb_checkpoints/**',
+      use: {
+        contextOptions: {
+          permissions: ['clipboard-read', 'clipboard-write']
+        }
```
krassowski, line 35: The alternative - as used in https://github.com/jupyterlab/jupyterlab/pull/14345 - is to enable them on per-test basis:

```js
const browserContext = page.context();
await browserContext.grantPermissions(['clipboard-read']);
```

I guess either way is fine.

krassowski (approved):
Thank you!

PR #13482, opened 2022-11-24
Use same key for saving user info in local store

hbcarlos:
Since we are using the username as a key to saving the user's identity in local storage, and the username is a randomly generated UUID, every time we remove the cookie, the server will generate a new anonymous user, and instead of removing the old identity and adding the new one to local storage, we are creating a new entry.

See:
Screenshot from 2022-11-24 17-51-28

## Code changes
It uses a service ID as a key for saving the user identity in local storage.

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/services/src/user/index.ts
+++ b/packages/services/src/user/index.ts
@@ -23,6 +23,13 @@ import { BaseManager, IManager as IBaseManager } from '../basemanager';
 const SERVICE_USER_URL = 'api/me';
 
 /**
+ * The service's ID.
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @hbcarlos 

One question, does that implies that the user name will be kept between two sessions?

hbcarlos:
> One question, does that implies that the user name will be kept between two sessions?

Yes

fcollonval:

Issue #13457, opened 2022-11-21
Events service - follow-up for JLab v4


fcollonval:
## Remaining `TODO` follow-on PRs
- [x] ~~In `EventManager#_subscribe()`, make the test emission a request with a registered schema. This requires registering a schema with the back-end application's event bus.~~
- [x] Remove capability probing for `api/events` in `EventManager#_subscribe()` after `jupyter_server` 2 becomes a hard requirement.
- [ ] Write tests for `EventManager` in the `@jupyterlab/services` test suite.

_Originally posted by @afshin in https://github.com/jupyterlab/jupyterlab/issues/12667#issuecomment-1316504650_

PR #13485, opened 2022-11-24
Follow-on to events service

afshin:
Follow-on PR for events service

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/13457


packages/services/src/event/index.ts
```diff
@@ -58,10 +56,9 @@ export class EventManager implements IDisposable {
    * Dispose the event manager.
    */
   dispose(): void {
-    if (this.isDisposed) {
+    if (this._poll.isDisposed) {
```
afshin, line 59:
```suggestion
    if (this.isDisposed) {
```

jtpio (approved):
Thanks!

PR #13491, opened 2022-11-25
Fixes sharing metadata

### LINKED ISSUE #12052: Metadata sharing in RTC sharing UNSHARED_KEYS
## Description

Metadata is being shared except for UNSHARED_KEYS (https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/notebook/src/model.ts#L36) but if you look at the code at https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/notebook/src/model.ts#L450-L454 you see that it is being shared in full if any other fields is edited

## Reproduce
Edit any metadata field, also the UNSHARED_KEYS are being sent over yjs to other clients

## Expected behavior

If the UNSHARED_KEYS have any meaning (this is unclear from the PR), they should never be shared, not even when other metadata changes.
If they do not have any meaning, the whole UNSHARED_KEYS thing should be removed

### LINKED ISSUE #12053: Deleting metadata keys doesn't propagate over RTC
## Description

When a metadata key is being delete, it's not propagated over RTC

## Reproduce

Delete a metadata key while in RTC mode, the other notebook still has the old values and will never receive the update

## Expected behavior

Other clients in RTC mode also see the removal of the metadata field

## Context

When the metadata is updated, https://github.com/jupyterlab/jupyterlab/blob/90ff82bb9023be8d09b895d8e8eedac45f82eba7/packages/notebook/src/model.ts#L446-L456 is called which calls https://github.com/jupyterlab/jupyterlab/blob/4575107d5d737c7909ec028c4c9b0a507629d10b/packages/shared-models/src/ymodels.ts#L363-L366 with the current metadata (so nothing about the deleted field) but the latter thinks it's receiving a partial metadata object and so it's appending the known state before it which means the deleted field is added again and so the sharedModel still has the deleted field and nothing is shared over RTC

hbcarlos:
fixes #12053
fixes #12052

## References
Fixes sharing metadata for JupyterLab 3.1.x - 3.5.x

## Code changes
Deletes and sets the new metadata.

### DIFF
--- a/packages/notebook/src/model.ts
+++ b/packages/notebook/src/model.ts
@@ -437,6 +437,7 @@ close the notebook without saving it.`,
     if (changes.metadataChange) {
       const metadata = changes.metadataChange.newValue as JSONObject;
       this._modelDBMutex(() => {
+        this.metadata.clear();
         Object.entries(metadata).forEach(([key, value]) => {
           this.metadata.set(key, value);
         });
@@ -450,7 +451,7 @@ close the notebook without saving it.`,
   ): void {
     if (!UNSHARED_KEYS.includes(change.key)) {
       this._modelDBMutex(() => {
-        this.sharedModel.updateMetadata(metadata.toJSON());
+        this.sharedModel.setMetadata(metadata.toJSON());
       });
     }
     this.triggerContentChange();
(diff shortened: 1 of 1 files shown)


fcollonval:
@hbcarlos do you have a clue why the example CI is failing?

fcollonval:
Thanks Carlos for looking at it

fcollonval (approved):

PR #13492, opened 2022-11-25
jupyter_server_ydoc>=0.6.0,<0.7.0

davidbrochart:
## References

https://github.com/jupyter-server/jupyter_server_ydoc/pull/64
https://github.com/y-crdt/ypy-websocket/pull/57

## Code changes

Metadata is not stored anymore. It used to consist of a timestamp, which is now built-in the YStore, so no need to duplicate data.

## User-facing changes

Smaller YStore.

YStore is incompatible.


fcollonval (approved):
Thanks

PR #13494, opened 2022-11-25
Backport #13433 (Allows to pause the execution during debug)

brichet:
Backport #13433

### DIFF
--- a/packages/debugger/src/debugger.ts
+++ b/packages/debugger/src/debugger.ts
@@ -118,7 +118,7 @@ export namespace Debugger {
 
     export const restartDebug = 'debugger:restart-debug';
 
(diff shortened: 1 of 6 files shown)


packages/debugger/src/debugger.ts
```diff
@@ -118,7 +118,7 @@ export namespace Debugger {
 
     export const restartDebug = 'debugger:restart-debug';
 
-    export const pause = 'debugger:pause';
```
krassowski, line 121: should we keep command names on 3.x?

brichet: Thanks @krassowski I forgot that one.

fcollonval (approved):
Thanks @brichet

PR #13468, opened 2022-11-22
Set fallback values for icons colors.

HaudinFlorence:
## References

This PR is a follow up of PR https://github.com/jupyterlab/jupyterlab/pull/13279 and should fix issue https://github.com/jupyter/notebook/issues/6633 reporting that icons are not visible with the jupyterlab-night theme

## Code changes

In packages/ui-components/style/icons.css, add fallback values for the css variables related to the colors of the icons in the launcher.

PR #13500, opened 2022-11-28
Set corrections to icons and switch colors

HaudinFlorence:
This PR fixes some issues on the colors of the icons that were improperly changed in PR #13279 and adds fall back variables for the switch that were missing in PR #13468.

## References

This PR is a follow up of PRs #13279 and #13468

## Code changes

The changes are in the variable.css files of both themes and in packages/ui-components/style/swtich.css.


fcollonval (dismissed):
Thanks @HaudinFlorence

HaudinFlorence:
bot please update snapshots

packages/theme-dark-extension/style/variables.css
```diff
@@ -424,10 +424,10 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-reject-color-active: var(--md-grey-800);
 
   /* File or activity icons and switch semantic variables */
-  --jp-jupyter-icon-color: var(--md-orange-900);
+  --jp-jupyter-icon-color: var(--md-orange-700);
```
krassowski, line 427: I see that this is only changed in dark theme and not the other. Is the intention to have two different colours depending on theme? If this is intended, please update the PR description to make this explicit for future reference.

As per [Jupyter brand guide](https://github.com/jupyter/design/blob/master/brandguide/brand_guide.pdf) the colour should be `#F37626` which is in between `--md-orange-700` (`#f57c00`) and `--md-orange-900` (`#e65100`) (but closer to 700). I am **not** requesting changes, just wanted to make sure you are aware that we do not currently comply with own brand guidelines. Suggestions on how to fix that are welcome, though it may be better discussed in a new issue :)

fcollonval: I realized that the color warn# and brand# are actually not the same in the light and the dark themes. As the icon was using those - this change is to avoid introducing a change. But you are making a good point, it sound better to set the official color for the Jupyter icon.

HaudinFlorence: I can update the description. Indeed most of the changes are for the dark theme except a correction for notebook icon color in light theme. Initially (before PR #13279) I thought from `variable.css` file that the color for jupyter and notebook were defined through :

-  `--jp-warn-color0: var(--md-orange-900)` for the light theme

-  `--jp-warn-color0: var(--md-orange-700)` for the dark theme.

I may be missing some transformation that turns jupyter color from `--md-orange-900` to  `#F37626` (for light theme).

jtpio (commented):
Looks like the updated snapshot has an issue:

image

packages/theme-dark-extension/style/variables.css
krassowski: Thanks! Just for context, about a year ago we changed `--jp-warn-color0` to use a darker colour to improve contrast of buttons. Change of the logo colour was not intended, but just happened because they were using the same variable. All I am saying is that we were not compliant with the brand colour for some time now, but your PR separating these variables did enable us to fix it so it is a good time to address it one way or the other).

jtpio:
For some reasons CI is not triggered when the bot pushes the snapshots commit: https://github.com/jupyterlab/jupyterlab/pull/13500/commits/0cc33343796d12fc5ed200c751012f5e3e748ffc

image

Not sure if this is an issue with GitHub or whether there is something to tweak in the GitHub Actions workflow.

packages/theme-dark-extension/style/variables.css
HaudinFlorence: > Thanks! Just for context, about a year ago we changed `--jp-warn-color0` to use a darker colour to improve contrast of buttons. Change of the logo colour was not intended, but just happened because they were using the same variable. All I am saying is that we were not compliant with the brand colour for some time now, but your PR separating these variables did enable us to fix it so it is a good time to address it one way or the other).

Ok thanks for giving me these elements of context (that I did not have). So I am even more happy to give back the "true" branc color to these icons :-)

HaudinFlorence:
> For some reasons CI is not triggered when the bot pushes the snapshots commit: [0cc3334](https://github.com/jupyterlab/jupyterlab/commit/0cc33343796d12fc5ed200c751012f5e3e748ffc)
> 
> image
> 
> Not sure if this is an issue with GitHub or whether there is something to tweak in the GitHub Actions workflow.

I  don't know how to fix this.

jtpio:
Opened https://github.com/jupyterlab/jupyterlab/issues/13505 to track it.

packages/theme-dark-extension/style/variables.css
HaudinFlorence: I found this in base.css of the launcher 
```
.jp-LauncherCard[data-category='Notebook'] .jp-LauncherCard-noKernelIcon {
  /* This color is copied from the notebook icon. */
  color: #ef6c00;
}
```
I will propose to change it to the #F37626. If you know other places where this may be corrected. Please let me know.

HaudinFlorence:
bot please update snapshots

fcollonval:
@HaudinFlorence would you mind not changing the launcher card styles in this PR?

packages/launcher/style/base.css
```diff
@@ -141,8 +142,8 @@
 }
 
 .jp-LauncherCard[data-category='Notebook'] .jp-LauncherCard-noKernelIcon {
-  /* This color is copied from the notebook icon. */
-  color: #ef6c00;
+  /* This color is the Jupyter brand's one. */
+  color: #f37626;
```
fcollonval, line 146: Let's use the CSS property here:

```suggestion
  color: var(--jp-jupyter-icon-color);
```

HaudinFlorence: > @HaudinFlorence would you mind not changing the launcher card styles in this PR?

Sorry for the changes committed by mistakes. They should have been removed in commit d8c9878

HaudinFlorence: > Let's use the CSS property here:

This has been done in commit d8c9878 .

packages/theme-light-extension/style/variables.css
```diff
@@ -419,6 +419,6 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-text-editor-icon-color: var(--md-grey-700);
   --jp-inspector-icon-color: var(--md-grey-700);
   --jp-switch-color: var(--md-grey-400);
-  --jp-switch-true-position-color: var(--md-orange-700);
+  --jp-switch-true-position-color: var(--md-orange-900);
   --jp-switch-cursor-color: rgba(255, 255, 255, 1);
```
fcollonval, line 423: I did not find any use of `--jp-switch-cursor-color`. If it is indeed unused could your remove it from the two themes please.

HaudinFlorence: Thanks. You are right. This css variable has been removed in commit 
cd90e73.

fcollonval (approved):
Thanks @HaudinFlorence

fcollonval:

PR #13510, opened 2022-11-29
Sanitize notification message

fcollonval:
## References

Follow up of https://github.com/jupyterlab/jupyterlab/pull/13365 and [JupyterLab weekly meeting](https://github.com/jupyterlab/team-compass/issues/152#issuecomment-1325457152)

## Code changes

Agressively sanitize the notification message

> I tried the iframe encapsulation. But it sounds to much to apply it per notification messages. And to have all notifications wrapped within an iframe, this requires quite some rework and it will also bring question about accessibility (how do you notify of a new alert within a iframe).

## User-facing changes

Most likely none

None


fcollonval:
Friendly ping to @vidartf and @jasongrout

jasongrout:
If we're okay with starting very conservative, I would suggest having messages be only plain text (set with textContent), with a maximum size, with a single url to the user can click on to read more. Starting simple is a feature, in that the simpler the message is forced to be, the less distracting it is in general for users.

jasongrout:
Perhaps RSS is a good model to follow: an item has three required fields:

* title
* description
* link

I'd encourage that title/descriptions be shortened to a max number of displayed characters to encourage best practices of not having overly long notifications, and to protect user screen space.

fcollonval:
bot please update snapshots

packages/apputils-extension/src/announcements.ts
```diff
@@ -150,7 +159,10 @@ export const announcements: JupyterFrontEndPlugin<void> = {
               if (!state.dismissed) {
                 options.actions = [
                   {
-                    label: trans.__('Do not show to me again'),
+                    label: trans.__('Hide'),
+                    caption: trans.__(
+                      'If pressed, this notification will never be showed again.'
```
jweill-aws, line 164:
```suggestion
                      'Never show this notification again.'
```

krassowski:
Looking across the board before we ship this:

- The privacy policy does not render the command to disable ("The plugin can also be disabled by executing:"):
  Screenshot from 2022-12-11 08-34-02
- Currently the link-button behaviour is not user-friendly because users will not know if clicking a button will open a new browser or perform an action in JupyterLab. This is because links are implemented using `window.open` and therefore the link information will not show on hover (which will annoy some users). Possible solutions are to:
  - use proper links styled like buttons and/or
  - add the external-website icon or
  - add the URL link in the new line in the title attribute of the button which acts as a link
- When I click "open privacy policy" it opens the policy and closes the popup; the popup should stay on until accepted or rejected.
- We could consider opening the privacy policy as an iframe in JupyterLab; this has benefit with respect to UX
- Minor suggestion on `Do you want to receive official Jupyter news?` → `Would you like to receive official Jupyter news?`
- Maybe the (x) button should explain what it does better (currently it says "Close notification"). As far as I understand it acts as "notify me later"/"I don't want to make a decision now".

fcollonval:

The changes made are:
- Fix code block in the privacy policies page
- Add display type to the actions that will style the button (available values are: _default_, _accent_, _warn_ or _link_). If the case of link, the button is identical to default but the text is colored and underlined with the link content color. I tested added a icon but if it feels strange.
I did not change the HTML tree because I want to avoid coding something specific and I want to be sure the notification message is displayed as HTML text node only for security.
- I added a way to not close the notification if an action is triggered.
- The privacy policies page is now opened in a tab within JupyterLab. I'm half happy about it as the documentation website looks partly broken (and lots of errors are prompted in the web console) due to the cross-origin restriction.
- Wording switched
- Wording for the cross button from _Close notification_ to _Hide notification_.

fcollonval:

jupyterlab/handlers/announcements.py
```diff
@@ -75,11 +81,13 @@ class CheckForUpdate(CheckForUpdateABC):
         logger - logging.Logger: Server logger
     """
 
-    async def __call__(self) -> Awaitable[Optional[str]]:
+    async def __call__(self) -> Awaitable[Tuple[str, Tuple[str, str]]]:
         """Get the notification message if a new version is available.
 
         Returns:
-            The notification message or None if there is no update.
+            None if there is not update.
```
jweill-aws, line 88:
```suggestion
            None if there is no update.
```

jupyterlab/handlers/announcements.py
```diff
@@ -118,11 +125,13 @@ class NeverCheckForUpdate(CheckForUpdateABC):
         logger - logging.Logger: Server logger
     """
 
-    async def __call__(self) -> Awaitable[Optional[str]]:
+    async def __call__(self) -> Awaitable[None]:
         """Get the notification message if a new version is available.
 
         Returns:
-            The notification message or None if there is not update.
+            None if there is not update.
```
jweill-aws, line 132:
```suggestion
            None if there is no update.
```

jweill-aws (approved):
Made a couple of small cosmetic changes and opened #13594 to revise privacy docs; otherwise, this looks good

fcollonval:
Merging this one to move forward with 3.6 beta release (CI failures are not related).

@krassowski if I release the beta this week. The RC will probably land before Christmas. But I won't be doing the final release before next year (probably waiting for the first January 4th meeting). Does that leave enough time for testing the 3.6.0?

fcollonval:

PR #13517, opened 2022-12-01
RTC: Move user name to user panel

martinRenou:
## References

This frees a bit of space in the top bar, and display the full user name in the RTC panel. 

### Before

Screenshot from 2022-12-01 09-56-43

### After

Screenshot from 2022-12-01 09-52-36


packages/collaboration/src/components.tsx
```diff
@@ -26,7 +26,7 @@ export const UserIconComponent: React.FC<Props> = props => {
       >
         <span>{user.initials}</span>
       </div>
-      <h3>{user.displayName}</h3>
```
martinRenou, line 29: This was actually a mistake apparently

jtpio:
Looks like the `js-collaboration` failure might be relevant? https://github.com/jupyterlab/jupyterlab/actions/runs/3591011716/jobs/6045047541

hbcarlos (approved):
Thanks, @martinRenou! Looks great.

I see you removed the tests. It makes sense since they were for the old user package, not for the user service but it would be great to add UI tests. Could you add some tests in galata?

fcollonval (approved):

martinRenou:
Thanks fred!

> I see you removed the tests. It makes sense since they were for the old user package, not for the user service, but it would be great to add UI tests. Could you add some tests in galata?

I agree we should add Galata tests for this. We would have caught the issue that prevented the name to be displayed in the left panel if we had some. I will try to come up with another PR for that.

PR #13526, opened 2022-12-01
Sets whether the model is collaborative or not when registering its factory

hbcarlos:
Follow-up #13458 

Hi @dlqqq, @ellisonbg and @fcollonval .

I like the idea of specifying whether the document is collaborative or not. Furthermore, I believe we should make it more robust by setting it as collaborative when registering the model factory.

### DIFF
--- a/packages/notebook/src/modelfactory.ts
+++ b/packages/notebook/src/modelfactory.ts
@@ -17,6 +17,7 @@ export class NotebookModelFactory
   constructor(options: NotebookModelFactory.IOptions) {
     this._disableDocumentWideUndoRedo =
       options.disableDocumentWideUndoRedo || false;
+    this._collaborative = options.collaborative ?? true;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/notebook/src/modelfactory.ts
```diff
@@ -74,7 +82,7 @@ export class NotebookModelFactory
   ): INotebookModel {
     return new NotebookModel({
       languagePreference,
-      collaborationEnabled,
+      collaborationEnabled: collaborationEnabled && this.collaborative,
```
dlqqq, line 85: You mean:

```suggestion
      collaborative: collaborationEnabled && this.collaborative,
```

Right?

dlqqq (changes requested):

packages/notebook/src/modelfactory.ts
hbcarlos: No, the attribute is called `collaborationEnabled`. I did not change them. See:

https://github.com/jupyterlab/jupyterlab/blob/49c5c0984b1ead0f7cd2ebac4a13378d3c7f8858/packages/notebook/src/model.ts#L522

https://github.com/jupyterlab/jupyterlab/blob/49c5c0984b1ead0f7cd2ebac4a13378d3c7f8858/packages/docregistry/src/default.ts#L27

dlqqq (approved):

fcollonval:

PR #8425, opened 2020-05-14
File editor right-click

karlaspuldaro:
## References
#2778 - Improve the right-click menu of the editor

## Code changes
New commands added to file editor right-click menu:
- Undo
- Redo (new icon added)
- Cut
- Copy
- Paste
- Select all

## User-facing changes
When right clicking on a file editor with a text selection:
image

When there is no text selection, cut and copy buttons are disabled

PR #13535, opened 2022-12-04
Add copy and paste commands to terminal context menu

### LINKED ISSUE #3011: Add Copy/Paste to the terminal right-click menu
Currently, we need to use shift to get the system right-click menu in order to copy/paste in the terminal. This functionality is hidden, so it would be interesting if copy and paste were added to the JupyterLab's right click menu.

krassowski:
## References

Closes #3011

Re-uses the same interface to Clipboard API that we use in `FileEditor`'s context menu for over two years now (#8425, JupyterLab 2.2).

## Code changes

- add new methods to `ITerminal` interface: `getSelection`,  `hasSelection` and `paste` following the underlying naming of `xterjm.js` implementation
- implement the new methods in `Terminal` by passing them over to `xterm.js` methods
- add new commands: `terminal:paste` and `terminal:copy`
- add the new commands to context menu

## User-facing changes

- 'Refresh terminal' command now uses refresh icon when in context menu
- 'Copy' and 'Paste' context menu actions added

Screenshot from 2022-12-04 14-44-37

When pasting for the first time the browser may (depending on browser/settings) ask for confirmation to allow access to the clipboard:

Screenshot from 2022-12-04 14-51-32

All the caveats of this not working if browser restricts Clipboard API apply, see [browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility).

<details>

For example, Firefox does not permit copy in non-secure contexts (`https` is ok) and does not allow paste at all. This is also the case for file editor, so this is not new. Ideally we would detect support and disable the command if support is not available but this is tricky for browsers which allow support after user consent.

</details>

Adds new non-optional methods to the interface.

If would like to backport to 3.6 if time permits, making the new methods optional in backport (but required for 4.0).

### DIFF
--- a/packages/terminal-extension/schema/plugin.json
+++ b/packages/terminal-extension/schema/plugin.json
@@ -4,9 +4,19 @@
   "jupyter.lab.menus": {
     "context": [
       {
-        "command": "terminal:refresh",
+        "command": "terminal:copy",
         "selector": ".jp-Terminal",
         "rank": 1
+      },
+      {
+        "command": "terminal:paste",
+        "selector": ".jp-Terminal",
+        "rank": 2
+      },
+      {
+        "command": "terminal:refresh",
+        "selector": ".jp-Terminal",
+        "rank": 3
       }
     ]
   },
--- a/packages/terminal-extension/src/index.ts
+++ b/packages/terminal-extension/src/index.ts
@@ -11,6 +11,7 @@ import {
   JupyterFrontEndPlugin
 } from '@jupyterlab/application';
 import {
+  Clipboard,
   ICommandPalette,
   IThemeManager,
   MainAreaWidget,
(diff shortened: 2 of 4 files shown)


fcollonval (approved):
Thanks @krassowski 

> Do you have time to make the backport this week. I'll like to get a beta out this week for 3.6; so we can focus on 4.0.0?

fcollonval:

> It will likely fail but that will set the bot label

dhirschfeld:
> *For example, Firefox does not permit copy in non-secure contexts (`https` is ok) and does not allow paste at all. This is also the case for file editor, so this is not new. Ideally we would detect support and disable the command if support is not available but this is tricky for browsers which allow support after user consent.*

Thanks for the heads-up. No paste is a bit disappointing 😢 From the link you posted it seems you also need to opt-in to the read feature on Firefox:

image

PR #13541, opened 2022-12-05
Add lumino with support for plugin deactivation

fcollonval:
## References

Update lumino to support https://github.com/jupyterlab/lumino/issues/278

## Code changes

N/A

## User-facing changes

N/A

N/A

### DIFF
--- a/builder/package.json
+++ b/builder/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "@jupyterlab/buildutils": "^3.6.0-alpha.4",
     "@lumino/algorithm": "^1.9.0",
-    "@lumino/application": "^1.27.0",
+    "@lumino/application": "^1.31.0",
     "@lumino/commands": "^1.19.0",
     "@lumino/coreutils": "^1.11.0",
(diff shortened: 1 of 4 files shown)


jtpio (approved):
Nice this is going to be useful for https://github.com/jupyterlab/jupyterlab-plugin-playground :+1: 

According to https://github.com/jupyterlab/lumino/pull/377#issuecomment-1274601424 this was initially not meant to be backported to Lumino 1.x, but I guess this might be related to the recent change of the TypeScript target, or other reasons?

fcollonval:
> Nice this is going to be useful for https://github.com/jupyterlab/jupyterlab-plugin-playground +1

This is exactly the reason I pushed forward.

> According to [jupyterlab/lumino#377 (comment)](https://github.com/jupyterlab/lumino/pull/377#issuecomment-1274601424) this was initially not meant to be backported to Lumino 1.x, but I guess this might be related to the recent change of the TypeScript target, or other reasons?

The backport was requiring some changes that I did not get the time to do what was necessary. And at that time I did not think it would take so long to get 3.6.0 out.

PR #5291, opened 2018-09-11
expand JupyterLab.IOptions to allow for custom DocumentRegistry, CommandLinker, ServiceManager to be passed into app

adobs:
expand JupyterLab.IOptions to allow for custom DocumentRegistry, CommandLinker, ServiceManager to be passed into app

PR #13558, opened 2022-12-08
Allow the application service manager to be set by a plugin

afshin:
This PR allows a new minimal plugin type, the `ServiceManagerPlugin` to set the application's default service manager. 

This plugin cannot have any `requires` or `optional` dependencies and its `activate` function's only argument is a Lumino `Application` instance (*not* a `JupyterLab` instance or even a `JupyterFrontEnd` instance).

This plugin must *synchronously* provide a `ServiceManager.IManager`.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/13631

## Code changes
This PR allows a plugin to override the default service manager. A minimal such plugin would look like this:
```ts
const services: ServiceManagerPlugin = {
  id: '@jupyterlab/application-extension:services',
  provides: ServiceManager.IManager,
  activate: () => new ServiceManager()
};
```

## User-facing changes
N/A

N/A


echarles:
Trying to understand the intent, if I develop an extension which declares/injects such a new service manager, will all notebooks use that new service manager?

What if 2 extensions declares/injects different service managers?

afshin:
> Trying to understand the intent, if I develop an extension which declares/injects such a new service manager, will all notebooks use that new service manager?

All notebooks and the entire application will use your service manager. So in their `activate` functions, when they access `app.serviceManager`, they will be accessing the one provided by your plugin.

> What if 2 extensions declares/injects different service managers?

Currently there is no explicit check for this. The first such plugin will win the race. However, like any other plugin ... we have an expectation that they won't clobber one another. I think writing a runtime error might be a good idea for this PR.

afshin:
@echarles I added a runtime warning if multiple service manager plugins exist, the warning will show which ones were ignored.

afshin:
Cheers, Eric! Other plugin failures don't stop the application from loading, so my intuition is to not disrupt the entire application.

echarles:
>  Other plugin failures don't stop the application from loading

This is a different case here. If 2 extensions aims to inject a service manager, and only one of them is randomly picked, the user does not know which low level service he will have while opening a notebook. I see it as a functional and security risk.

afshin:
Right, but this isn't the sort of plugin an end user will typically install. This is much more the sort of plugin an administrator would install for a deployment. It's not really an end user piece of functionality, it's lower level than that.

echarles:
> Right, but this isn't the sort of plugin an end user will typically install. This is much more the sort of plugin an administrator would install for a deployment. It's not really an end user piece of functionality, it's lower level than that.

Prolly in most case, but "better safe than sorry"

afshin:
I don't think there is going to be a circumstance where a user installs two different service manager plugins that connect to two different `jupyter-server`-compatible REST+WebSocket API servers and then has doubt which one their application is connected to.

I think treating this plugin the way other plugins work: whichever loads first "wins" and the other one gets a console warning ... is a sound starting point. Because this particular sort of plugin is one that basically *no* end user would be installing, it will pretty much always be an administrator installing it for a deployment.

I wonder what other people think. We can bring it up at the next call, or perhaps people while chime in here, as well.

echarles:
>. We can bring it up at the next call, or perhaps people while chime in here, as well.

I think the place to discuss that is the Security meeting as silently changing the service (so potentially the server on which the code is executed) for the user (no user is looking nor understanding those king of logs) is a security issue.

afshin:
I am happy to discuss this functionality with the security team, but I think the security implication of this change is very different than what you seem to believe it is.

*Any* plugin a user installs can add a set of API endpoints, write and execute arbitrary code on the user's machine, etc.

This particular functionality does not create a new security vector that didn't already exist.

echarles:
> Any plugin a user installs can add a set of API endpoints, write and execute arbitrary code on the user's machine, etc.

True, but with this proposal, my understanding is that an extension can silently replace the core service manager that is used by all notebooks, which is effectively adding a new security issue vector.

afshin:
> True, but with this proposal, my understanding is that an extension can silently replace the core service manager that is used by all notebooks, which is effectively adding a new security issue vector.

You have long had the ability to run a custom contents manager which creates the *exact* same risk. You have long had the ability to install an extension that silently overwrites every `.ipynb` file on your system. This PR does not introduce a *new* security vector at all, in my opinion, because the level of risk you introduce with these extensions is the same level of risk as any other extension can create.

echarles:
> You have long had the ability to run a custom contents manager

The content manager is governed by the startup configuration, here we are talking about users installing extensions driving unpredictable behavior for the users (not guarantee which service manager, which governs code execution... is finally used for all the notebook). Let's review the risks in a meeting.

afshin:
I'm absolutely happy to have a conversation with the security team.

But as a data point, here is a currently possible (in all versions of JupyterLab) front-end plugin that replaces the service manager. This has *always* been possible, it's just not a nice API:

```ts

const insecure: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/application-extension:insecure',
  autoStart: true,
  activate: (app) => {
    console.log('replacing the application service manager');
    ((app as unknown) as any).serviceManager = new ServiceManager();
  }
};
```

echarles:
>  This has always been possible, it's just not a nice API:

Correct, hacking around has always been possible, and I am sure there are many other ways an extension developer can fool the user. Those kind of hacks deserve to be listed and solved with additional means in the core of JupyterLab. That would be a nice action to take to make JupyterLab more secured than it is now.

I am just concerned that making such things "official" is not a good move, unless enough visual feedback and warning are clearly shown to the user with explicit consent from him so he agrees to be connected with another server that is not the one he things he is when opening his notebook.

afshin:
I understand your position. I disagree with it. There is *no new security* vector for a malicious actor to operate in, the risk profile after this PR is the exact same risk profile as before this PR.

But since we disagree, I believe you're right that we should probably involve other people who have thought about security in trying to resolve this.

echarles:
The nuance is subtle. Agree that there is not new security vector, but I see that vector being now "officialised" (in other words "supported", "burned in the API", "advertised") and given for usage to the extension developers. Once "given" and "used", it will not be possible to take that "official" security risk back. Thinking loud, I can see two potential enhancements towards this move

1. Have a predictable way to identify the winner service manager.
2. Have a visual mechanism to collect consent from the user in case of service manager change.

Let's hear what others think about this.

jtpio:
> Have a visual mechanism to collect consent from the user in case of service manager change.

Isn't it already the purpose of the existing disclaimer in the extension manager?

image

jasongrout:
Update: we had a good discussion of some points in the JupyterLab Dev meeting (recording will be posted online).

At one point several years ago, a group I was with needed to replace the service manager. We considered writing something like this, but we decided in the end that it really did make sense to either subclass the application to customize this fundamental part of the application, or to write some custom code to instantiate the application with our own service manager as a constructor argument. We contributed https://github.com/jupyterlab/jupyterlab/pull/5291 to make the service manager customizable and then we were able to change the service manager by passing it in as an option to the application.

bollwyvl:
A bit late to the party, but my personal take: more opportunities where behavior can be composed than winner-takes-all would lead to better user outcomes than introducing a new, special class of plugin.

Speaking from the cited example in the linked issue: jupyerlite is actually three separate, custom `Applications`... probably only the `lab` one would benefit from any kind of new plugin thing... but is still pretty much guaranteed to _have_ to be a custom app.

As we've found folk (e.g. [_past me_](https://github.com/jupyter-lsp/jupyterlab-lsp/issues/880)) don't even always use the ServiceManager's APIs, even when they make sense (`fetch` and `WebSocket`).

The thing I'd more like to see is a hard look at where the ServiceManager could grow more places to offer multiple ways to do things, with well-defined, easy to consume APIs. Specifically from the jupyterlite perspective, the biggest thing would be to make kernelspecs and kernels composable:
  - perhaps introduce a URI scheme for kernel names
    - the "null" protocol would belong to a first-in-wins `defaultKernelProvider` which lite, as an app, could provide
    - jupyterlite could `addKernelSpecProvider({id: 'lite', getKernelSpecs: async () => ...})`
      - or, if it _is_ the app, it could claim the null one... or both, to account for hard-coded kernel names (happens)
    - and then also `addKernelProvider` to mirror the other side
      - a key thing here might be the ability to mount multiple kernels inside the same emscripten WebWorker, which could make sense in certain cases

jasongrout:
From the dev meeting today:

- Nick Bollweg says that this wouldn't help JupyterLite, since JLite needs its own application anyway
- Darian says that if this doesn't solve JupyterLite's usecase, which normally would just be a clone of JupyterLab, then perhaps it's not worth the extra complexity this PR introduces, and he'll close this PR.

jtpio:
Something that was also discussed elsewhere: maybe we could go all the way and make the underlying services provided by the `ServiceManager` swappable and plugin based.

So the [individual managers](https://github.com/jupyterlab/jupyterlab/blob/f8bdeed9273d86ea89fc5b54e0891acbae32ba8b/packages/services/src/manager.ts#L238-L300) like the kernel and session managers would be provided via regular JupyterLab plugins. JupyterLab would provide default implementations by default just like now. But they could easily be replaced by third-party extensions via the `disabledExtensions` and federated extension systems.

This likely requires significant changes so something to discuss and perhaps consider for JupyterLab 5.

Additional APIs like `addKernelSpecProvider` and others discussed above could also be considered as additions to the API surface. Maybe those could land in a 4.x if they are backward compatible or made optional.

PR #13565, opened 2022-12-09
Improve API of `MarkdownCell` to extend it

fcollonval:
## References

I worked with @stevejpurve on replacing the markdown renderer in JupyterLab to use mystjs. We have realized that the current API is not ideal as it is not possible to avoid rendering with the default renderer as the container widget and some methods are private.

See https://github.com/executablebooks/jupyterlab-mystjs/blob/main/src/MySTMarkdownCell.ts

## Code changes

Makes the `_renderer` widget protected so it can be updated when extending the cell content factory.

Call `_updateRenderedInput` within `renderInput`. So it can be shunted if required.

## User-facing changes

N/A

`MarkdownCell.renderInput` does not take any argument


docs/source/extension/extension_migration.rst
```diff
@@ -36,13 +36,14 @@ bumped their major version (following semver convention). We want to point out p
...
-   ``MarkdownCell.toggleCollapsedSignal`` renamed ``MarkdownCell.headingCollapsedChanged``
-   To support notebook windowing, cell widget children (e.g. the editor or the output area) are not instantiated
+   * ``MarkdownCell.toggleCollapsedSignal`` renamed ``MarkdownCell.headingCollapsedChanged``
+   * To support notebook windowing, cell widget children (e.g. the editor or the output area) are not instantiated
    when the cell is attached to the notebook. You can test for ``isPlaceholder()`` to see if the cell has been
    fully instantiated or wait for the promise ``ready`` to be resolved. Additionally an attribute ``inViewport``
    and a signal ``inViewportChanged`` are available to test if the cell is attached to the DOM.
    If you instantiate standalone cells outside of a notebook, you will probably need to set the constructor option
    ``placeholder`` to ``false`` to ensure direct rendering of the cell.
+   * ``MarkdownCell.renderInput`` does not take any more arguments.
```
krassowski, line 46:
```suggestion
   * ``MarkdownCell.renderInput`` does not take arguments anymore.
```

jtpio:
Just rebased via the GitHub UI to trigger the CI and grab the latest changes.

packages/cells/src/widget.ts
```diff
@@ -2203,6 +2212,8 @@ export namespace MarkdownCell {
      * Show editor for read-only Markdown cells.
      */
     showEditorForReadOnlyMarkdown?: boolean;
+
+    /** */
```
jtpio, line 2216: Looks like this can be removed:

```suggestion
```

docs/source/extension/extension_migration.rst
```diff
@@ -36,13 +36,14 @@ bumped their major version (following semver convention). We want to point out p
    * The ``add-sibling`` script has been removed. Check out :ref:`source_dev_workflow` instead.
    * The ``exitOnUuncaughtException`` util function has been renamed to ``exitOnUncaughtException`` (typo fix).
 - ``@jupyterlab/cells`` from 3.x to 4.x
-   ``MarkdownCell.toggleCollapsedSignal`` renamed ``MarkdownCell.headingCollapsedChanged``
-   To support notebook windowing, cell widget children (e.g. the editor or the output area) are not instantiated
+   * ``MarkdownCell.toggleCollapsedSignal`` renamed ``MarkdownCell.headingCollapsedChanged``
```
krassowski, line 39:
```suggestion
   * ``MarkdownCell.toggleCollapsedSignal`` has been renamed to ``MarkdownCell.headingCollapsedChanged``
```

packages/cells/src/widget.ts
```diff
@@ -1945,9 +1945,9 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
   }
 
   /**
-   * Renderer
+   * Markdown rendered container
    */
-  get renderer(): IRenderMime.IRenderer {
+  get renderer(): Widget {
```
krassowski, line 1950: Why was the type changed to `Widget` rather than `IRenderMime.IRenderer` as it was? It seems that `IRenderer` inherits from `Widget` so at first glance it should be better as-is?

https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/rendermime-interfaces/src/index.ts#L288C20-L301

docs/source/extension/extension_migration.rst
jtpio: Thanks!

packages/cells/src/widget.ts
fcollonval: Reusing the rendermime renderer for the markdown cell is an implementation detail (actually we need to build a rendermime datamodel each time just to use it). When extending the markdown class, the only interest is to provide a Widget to be injected in the input area via:

https://github.com/jupyterlab/jupyterlab/blob/ecec3f6c90f54c4ad41b922e5ab16b3cae321953/packages/cells/src/inputarea.ts#L112

That method already accept a more generic `Widget`.

fcollonval: The public getter was added in 

https://github.com/jupyterlab/jupyterlab/pull/11689/files#diff-352916c4cc928b1298618fffc0424a71ee8816925f3a36c77d8593057e7b6f89

because it is needed to search in the rendered markdown widget at (`GenericSearchProvider` expect a `Widget`):

https://github.com/jupyterlab/jupyterlab/blob/ecec3f6c90f54c4ad41b922e5ab16b3cae321953/packages/cells/src/searchprovider.ts#L579-L581

There are no other usage and I exported the more restrictive `IRenderMime.IRenderer` as it was the type of the variable.

krassowski: https://github.com/jupyterlab/jupyterlab/blob/21820b8762d9f058defe4411d9ad77c7e3d558a9/packages/cells/src/widget.ts#L2166

What I was worried about is that downstream extension authors could override renderer getter with a `Widget` without a `renderModel` method which would break the above. Is my reasoning correct?

fcollonval: Yes you are right

Issue #13545, opened 2022-12-06
Oval-shaped circles in TOC running markers


denilsonsa:
## Description

When running all cells of a notebook while having the TOC open, little circles are displayed to the right of each section title. That's great, but those circles are oval-shaped sometimes, depending on the length of the title. This is specially noticeable if the cells take minutes to run.

## Reproduce

1. Create an IPython notebook inside Jupyter Lab.
2. Make sure to create several Markdown cells with titles (headings). Some of those should be long, others should be short.
3. Open the TOC (Table Of Contents) sidebar. Observe how some of the section titles are wrapped into multiple lines.
4. Restart and run all cells.
5. Pay attention to the little circles next to each section title.

## Expected behavior

All the circles should be uniform and look the same.

## What happens

Since each row from the TOC is an horizontal `flex` box, the circle can get its width adjusted according to the flex model. Look at this screenshot:

![Screenshot of the issue, several circles are distorted into ovals](https://user-images.githubusercontent.com/121676/205956868-16502a96-be4e-405e-8fcb-60e4acf93f93.png)

## Solution

Please add `flex: auto 0 0` to `.toc-entry-holder::after` selector. This is a shorthand for setting `flex-shrink` and `flex-grow` to zero, making sure that element maintains a fixed size.

See how I fixed it directly into the browser dev tools:

![Screenshot of the browser dev tools, showing the suggested fix.](https://user-images.githubusercontent.com/121676/205958210-259997de-165e-4393-b2ad-c9ab599a234d.png)

I can't find where this fix should be applied in this repository source-code. Can someone please apply the fix instead of me?

## Context

- Browser and version: Google Chrome 107
- JupyterLab version: 3.5.0

uenot:
I'd like to claim this! I'm a new contributor, but I believe I've found the location of the fix— will open a PR shortly.

sagar-ghub:
is this issue still open or the changes has been made ?

uenot:
Changes have been made, see #13568

fcollonval:
Closing as solved by #13568

PR #13566, opened 2022-12-09
Fix TOC running marker resizing

uenot:
## References

Addresses issue #13545

This is an update to version 3.5.x.

## Code changes

As described in the issue, added `flex: auto 0 0` to the selector `.toc-entry-holder::after`. This prevents the running indicators in the table of contents from warping when long titles are added.

## User-facing changes

Before:
Screen Shot 2022-12-09 at 4 43 40 PM

After:
Screen Shot 2022-12-09 at 4 40 50 PM

None

### DIFF
--- a/packages/toc/style/base.css
+++ b/packages/toc/style/base.css
@@ -406,6 +406,7 @@
   margin: 2px;
   background: none;
   border: none;
+  flex: auto 0 0;
 }
 
 .toc-entry-holder[data-running='0']::after {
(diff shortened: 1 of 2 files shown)


krassowski:
The usual workflow would be to target latest minor branch in this case 3.6.x. Would you mind re-applying the changes on top of 3.6.x and opening the PR targeting that branch?

krassowski:
Superseded by #13568

Issue #13580, opened 2022-12-13
Attempts to write workspaces to non-writable workspaces dir in shared env


minrk:
## Description

When running in a shared virtualenv where `$PREFIX/etc/jupyter`.

## Reproduce

```
python3 -m venv ./test-env
source ./test-env/bin/activate
pip install jupyterlab
# make $PREFIX/etc/jupyter read-only
chown root ./test-env/etc/jupyter
jupyterlab
```

Upon launching jupyterlab, when it attempts to save the workspace it fails with a permission error:

```
W 2022-12-13 09:45:49.114 LabApp] wrote error: "[Errno 13] Permission denied: '/Users/minrk/env/lab/etc/jupyter/lab'"
[E 2022-12-13 09:45:49.115 LabApp] {
      "Host": "localhost:8888",
      "Accept": "*/*",
      "Referer": "http://localhost:8888/lab?",
      "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Safari/605.1.15"
    }
```

 This can be avoided by setting `JUPYTER_PREFER_ENV_PATH=0`, but I think it may make sense to check for a writable dir when picking the default workspaces dir at startup. I don't think it should generally be assumed that `jupyter_config_dir()` is writable.

I'm not 100% sure if this is a jupyterlab issue (workspaces manager assumes writable path), or a jupyter-core issue (jupyter_config_dir() should be expected to be writable), but my inclination is it's a jupyterlab one, since that's where the assumption of writability is. But I suspect user-settings, etc. also have the same assumption, so a common location for 'writable config dir' probably makes sense, which may belong in core.

## Expected behavior

workspaces dir picks a writable dir, if jupyter_config_dir is not writable.

## Context

- Operating System and version: macOS 13.0.1
- Browser and version: Safari 16.1
- JupyterLab version: 3.5.1

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/Users/minrk/env/lab/bin
	/opt/homebrew/opt/ruby/bin
	/Users/minrk/conda/bin
	/usr/local/opt/ruby/bin
	/Users/minrk/dev/mine/git-stuff/bin
	/Users/minrk/conda/condabin
	/opt/homebrew/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin
	/Users/minrk/.local/bin
[...]

PR #13581, opened 2022-12-13
ensure user-settings/workspaces are writable

minrk:
searches config path for a writable dir instead of returning the first item.

avoids issues where workspaces/user-settings are assumed to be writable, but jupyterlab is running in a shared virtualenv, which comes at the front of the path, but may not be writable.

Last case not handled is if no writable dir is found. Should it make a temporary dir, or fallback on the non-writable first match and let writes fail (current behavior)? This is rather unlikely, since it includes `$HOME/.jupyter`.

Another option would be to _just_ check `jupyter_config_path()[0]` (sometimes env, sometimes home) and `jupyter_config_dir()` (always home or explicit), rather than the full path, which is going to include lots of system paths.

## References

#13580

## Code changes

Instead of picking `jupyter_config_dir()` unconditionally, pick the first writable config dir to store user-settings/workspaces by default.

## User-facing changes

Workspaces/user-settings will pick a writable directory instead of the first match, which may not be writable.

### DIFF
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -136,17 +136,53 @@ def pjoin(*args):
     return osp.abspath(osp.join(*args))
 
 
+def _writable_dir_or_parent(path):
+    """Return whether a path represents a writable directory
+
+    Returns true a non-existent path if it can be created
+    because the nearest parent is writable.
+    """
+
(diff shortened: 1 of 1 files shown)


minrk:
closed in favor of #13589

PR #13584, opened 2022-12-13
Drop typestyle

fcollonval:
## References

https://github.com/jupyter/notebook/pull/6627#issuecomment-1320045745

## Code changes

This removes the usage of `typestyle` (except for the `LabIcon`) in the status bar and in the shortcuts extension to align better with the code base.
It also simplifies a bit the Shortcuts UI styles (but not everything as this is not the primary intend)

## User-facing changes

Small changes in the shortcuts setting editor.

None


packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,418 @@
...
+}
+
+.jp-mod-defunc-Submit {
+  background: var(--jp-layout-color3);
+  background-image: var(--jp-icon-checkmark-white);
+}
+
+.jp-mod-conflict-Submit {
+  background: var(--jp-error-color1);
+  background-image: var(--jp-icon-error-white);
```
krassowski, line 103: There is no such an icon, as there is no `jp-icon-checkmark-white`. It seems that it was not brought in when the extension got merged into the core, so it is not a problem in this PR. Would you mind (in order of preference):
- using a different icon (e.g. `jp-icon-check` or `jp-icon-save` for `jp-icon-checkmark-white`)/adding an icon if needed, or
- removing the icon and just adding text in the appropriate buttons, or
- adding a TODO comment when non-existing icons are referenced?

krassowski: Currently the buttons looks like this:

Screenshot from 2022-12-17 16-59-53

Screenshot from 2022-12-17 17-00-36

Also it appears we don't have an UI screenshots for setting editor; I would not want to ask for too much here - if it is out of scope, we can create a "good first issue" to add them.

fcollonval: Thanks for the review @krassowski 

The latest commit add/restore proper icons for the shortcut input elements and the search input (switch to use `InputGroup` for consistency).
I also switch to use icon for the sort indicator on the table header.

I will leave the UI screenshots out of this one.

packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,394 @@
...
+}
+
+/* Shortcut Title Item Style */
+.jp-Shortcuts-Header {
+  display: flex;
+  cursor: pointer;
+}
+
+.jp-Shortcuts-Header:hover .jp-ShortcutTitleItem-sortButton[fill],
+.jp-Shortcuts-Header:focus .jp-ShortcutTitleItem-sortButton[fill] {
```
krassowski, line 292: This one does not have an effect, need to add `.jp-icon3`

```suggestion
.jp-Shortcuts-Header:hover .jp-ShortcutTitleItem-sortButton .jp-icon3[fill],
.jp-Shortcuts-Header:focus .jp-ShortcutTitleItem-sortButton .jp-icon3[fill] {
```

packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,394 @@
...
+
+.jp-Shortcuts-Symbols td:nth-child(2) {
+  padding-right: 10px;
+}
+
+.jp-Shortcuts-Search {
+  height: 30px;
+}
+
+.jp-Shortcuts-Search input {
```
krassowski, line 343: This is a generic tag selector, so lets shorten the search to avoid performance penalty.

```suggestion
.jp-Shortcuts-Search > input {
```

packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,394 @@
...
+.jp-Shortcuts-Row {
+  padding: 10px;
+  width: 100%;
+  display: table-row;
+  border-bottom: var(--jp-border-width) solid var(--jp-border-color1);
+  vertical-align: middle;
+  background-color: var(--jp-layout-color0);
+}
+
+.jp-Shortcuts-Row:hover #shortcut-keys {
```
krassowski, line 138:
```suggestion
.jp-Shortcuts-Row:hover .jp-Shortcuts-ShortcutKeys {
```

Replace ID with class name

packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,394 @@
...
+  outline: none;
+  border: none;
+}
+
+.jp-Shortcuts-ShortcutContainer {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.jp-Shortcuts-ShortcutContainer:hover #shortcut-keys {
```
krassowski, line 190:
```suggestion
.jp-Shortcuts-ShortcutContainer:hover .jp-Shortcuts-ShortcutKeys {
```

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -376,12 +357,12 @@ export class ShortcutItem extends React.Component<
   getShortCutForDisplayOnly(key: string): JSX.Element[] {
     return this.props.shortcut.keys[key].map(
       (keyBinding: string, index: number) => (
-        <div className={ShortcutKeysContainerStyle} key={index}>
-          <div className={ShortcutKeysStyle} id={'shortcut-keys'}>
+        <div className="jp-Shortcuts-ShortcutKeysContainer" key={index}>
+          <div className="jp-Shortcuts-ShortcutKeys" id={'shortcut-keys'}>
```
krassowski, line 361:
```suggestion
          <div className="jp-Shortcuts-ShortcutKeys">
```



I know this was not introduced in this PR but `shortcut-keys` identifier gets incorrectly re-used for each row of the table breaking HTML validity. Is it used elsewhere, or can we drop it?

krassowski (commented):
Looks great! Just some minor suggestions.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
fcollonval: This extension is full of that kind of usage :sad:

That one at least does not seem to be reuse like that one :scream: 

https://github.com/jupyterlab/jupyterlab/blob/5582852e4af32269e2cd7f9ac01bce696b7713ea/packages/shortcuts-extension/src/components/ShortcutInput.tsx#L413-L414

packages/shortcuts-extension/style/base.css
```diff
@@ -0,0 +1,394 @@
...
+}
+
+.jp-Shortcuts-Header:active .jp-ShortcutTitleItem-sortButton {
+  outline: none;
+}
+
+.jp-Shortcuts-CurrentHeader div {
+  color: var(--jp-ui-font-color0);
+  font-weight: bold;
+}
```
krassowski, line 303:
```suggestion

.jp-Shortcuts-CurrentHeader .jp-icon3[fill] {
  fill: var(--jp-ui-font-color0);
}
```


It looks that internal div to which this was previously referring was removed in 47a1eeaf016b0b656e5225e88a11a415cd5395a6 (replaced by a proper icon) so there is no text in div to bolden/change colour for anymore.

krassowski (approved):
Thanks! Just one final suggestion and good to merge.

packages/shortcuts-extension/style/base.css
fcollonval: Nice catch and thanks a lot for the review

jtpio (approved):
Thanks!

Issue #10239, opened 2021-05-13
Multiple cursors not showing up in RTC


krassowski:
![multi-cursor](https://user-images.githubusercontent.com/5832902/118116796-a2c91500-b3e2-11eb-9e46-40dbcc47e200.gif)

### Problem

Extra CodeMirror cursors set by Ctrl + click do not show up on the collaborator view of the editor, nor does the extra selection. It would be nice to have multi-cursor mode work better with RTC.

dmonad:
Thanks for opening this issue @krassowska. I didn't even know that this was possible in y-codemirror

Currently, the editor binding [y-codemirror](https://github.com/yjs/y-codemirror) only syncs the primary selection. I opened a separate ticket to track the work: https://github.com/yjs/y-codemirror/issues/17 

In terms of priority, we are probably working on other stuff first.

PR #13586, opened 2022-12-13
Remove old editor user caret widget (dead code) and selection style

fcollonval:
## References

Follow-up on the RTC work. The old infrastructure to display selection for various users is not needed any longer.

## Code changes

Remove code related to displaying the user caret (that was not used - it is provided by another CodeMirror extension from Yjs).
Remove selection style that is not used anywhere (was envisaged for that custom caret as can be seen on 3.6.x:

https://github.com/jupyterlab/jupyterlab/blob/288bd2da202bd8eb5e8367bc79f61a2bb20692af/packages/codemirror/src/editor.ts#L1092

## User-facing changes

None

Remove `ISelectionStyle` and `IEditor.selectionStyle`.

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -48,6 +48,9 @@ bumped their major version (following semver convention). We want to point out p
    of ``ICompletableAttributes`` interface by ``ICompletionProvider``. To create a completer provider
    for JupyterLab, users need to implement the interface ``ICompletionProvider`` and then register
    this provider with ``ICompletionProviderManager`` token.
+- ``@jupyterlab/codeeditor`` from 3.x to 4.x
+   * Remove ``ISelectionStyle`` (and therefore ``defaultSelectionStyle`` and ``IEditor.selectionStyle``). This was envisaged
+     for real-time collaboration. But this is not used in the final implementation.
 - ``@jupyterlab/console`` from 3.x to 4.x
    The type of ``IConsoleHistory.sessionContext`` has been updated to ``ISessionContext | null`` instead of ``ISessionContext``.
    This might break the compilation of plugins accessing the ``sessionContext`` from a ``ConsoleHistory``,
--- a/packages/codeeditor/src/editor.ts
+++ b/packages/codeeditor/src/editor.ts
@@ -77,35 +77,6 @@ export namespace CodeEditor {
   }
 
   /**
-   * A selection style.
-   */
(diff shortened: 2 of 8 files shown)


docs/source/extension/extension_migration.rst
```diff
@@ -48,6 +48,9 @@ bumped their major version (following semver convention). We want to point out p
    of ``ICompletableAttributes`` interface by ``ICompletionProvider``. To create a completer provider
    for JupyterLab, users need to implement the interface ``ICompletionProvider`` and then register
    this provider with ``ICompletionProviderManager`` token.
+- ``@jupyterlab/codeeditor`` from 3.x to 4.x
+   * Remove ``ISelectionStyle`` (and therefore ``defaultSelectionStyle`` and ``IEditor.selectionStyle``). This was envisaged
+     for real-time collaboration. But this is not used in the final implementation.
```
krassowski, line 53: Are those being replaced by an equivalent functionality provided in upstream packages? More generally, what is the status of RTC support for multi-selection (https://github.com/jupyterlab/jupyterlab/issues/10239)?

krassowski: And selection in general? Do we have an issue for tracking that selection highlights are not implemented in RTC and therefore collaborators cannot see what others selected?

fcollonval: Each RTC library comes with its own way to exchange non-content information. Yjs is exchanging information about user and cursor position through the awareness. And the transfer of information between CodeMirror editor and Yjs is done through an external library `y-codemirror.next`; for example the line that set the primary cursor in the awareness is:

https://github.com/yjs/y-codemirror.next/blob/318bbf7b684ea64d015f8f40f68e6820461e1d52/src/y-remote-selections.js#L170

The binding for CodeMirror 6 takes the form of additional CodeMirror 6 extensions.

To share selection and multi-cursors, we need to add that information in the awareness.

---

Regarding styling, as the user color is a thing now, it makes no sense to be able to a color for the user selection. That selection color should be the same (probably adding some transparency).

> I'm working on a follow-up of CodeMirror 6 migration to remove the usage of global object in favor of our plugin system. And then I plan to have a PR to improve the collaborative user display (improving the user name display, adding selection, and multi-selections).

krassowski (approved):

hbcarlos (approved):
Thanks @fcollonval!

Issue #13547, opened 2022-12-06
Debugger: pause on exception with more filter


brichet:
Currently the pause on exception (https://github.com/jupyterlab/jupyterlab/pull/11752) allows only the `userUnhandled` filter.
It could be interesting to allow other filters (`raised` or `uncaught`), as discussed in https://github.com/jupyterlab/jupyterlab/pull/13433#pullrequestreview-1185688551.

### Question about current behavior
Currently, when setting up the `pauseOnException`, the request is:
```
{
    "filters": ["raised", "uncaught", "userUnhandled"],
    "exceptionOptions": [
        {
            "path": [{"names": ["Python Exceptions"]}],
            "breakMode": "userUnhandled"
        },
        {
            "path": [{"names": ["Python Exceptions"]}],
            "breakMode": "userUnhandled"
        },
        {
            "path": [{"names": ["Python Exceptions"]}],
            "breakMode": "userUnhandled"
        }
    ]
}
```
According to the corresponding code in [debugpy](https://github.com/microsoft/debugpy/blob/79b7c5634ed5179d193ae563c56ee25dcaed9c30/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L844) (if this is the one), it seems that:
- if `exceptionOptions` are received, then the filters are unused
- if the path of an `exceptionOption` contains only "Python Exceptions", then the corresponding `breakMode` is applied to all kind of exception (`BaseException`).

At the end, the whole request above seems similar to:
```
{
    "filters": ["userUnhandled"]
}
```

I wonder if there is a reason to send the first request instead of the second one.
cc @andrewfulton9

### Proposed Solution

- Add a dropdownlist to select filter.
- Send only the expected filter to `debugpy`, without the exception options as long as there is not UI to select specific option. 

### Additional question

Also, I wonder if there would be a reason to allow multiple filters (like in VSCode). I understand that `raised` includes `uncaught` which includes `userUnhandled`.

brichet:
Closed by https://github.com/jupyterlab/jupyterlab/pull/13601

PR #13601, opened 2022-12-15
Filtering breakpoints on exception in debugger

brichet:
This PR add filters to the breakpoints on exception in debugger.

[simplescreenrecorder-2022-12-15_09.46.32.webm](https://user-images.githubusercontent.com/32258950/207816667-16906c1f-3bae-47eb-8e33-2b0407026e2d.webm)

## References

- https://github.com/jupyterlab/jupyterlab/issues/13547

**NOTE**
As explained in the issue above, this changes does not allow selecting several filter. 
It should not be a big deal to add this feature, but I'm not sure this is relevant.

## Code changes

- Replaces the previous command `debugger:pause-on-exceptions`.
- Simplifies the `setExceptionBreakpoints` request, by providing only the selected filter, instead of duplicated `exceptionOptions`.
- Adds a signal to `DebuggerService` emitting when the filter changes.
- Adds tests.

## User-facing changes

- Replaces the *pause* icon pause by an *exception* icon exceptions.
- Add a menu to select the filter when clicking on the icon.


packages/debugger-extension/src/index.ts
```diff
@@ -719,24 +720,37 @@ const main: JupyterFrontEndPlugin<void> = {
...
+                items: items,
+                current: service.session?.currentExceptionFilter || 'None',
+                editable: false
+              })
+            ).value ?? undefined;
+        }
+        if (!filter) {
+          return;
+        }
+        await service.pauseOnExceptions(filter as string);
```
fcollonval, line 753: The input dialog return the same object as a `Dialog`. So you should prefer checking if the button pressed was accept:

```suggestion
          const result = await InputDialog.getItem({
                title: trans.__('Select a filter for breakpoints on exception'),
                items: items,
                current: service.session?.currentExceptionFilter || 'None',
                editable: false
              });
          filter = result.button.accept ? result.value : null;
        }
        if (!filter) {
          return;
        }
        await service.pauseOnExceptions(filter as string);
```

packages/debugger/src/panels/breakpoints/index.ts
```diff
@@ -32,11 +30,12 @@ export class Breakpoints extends PanelWithToolbar {
...
-      new CommandToolbarButton({
-        commands: commands.registry,
-        id: commands.pause,
-        label: ''
+      'pauseOnException',
+      new PauseOnExceptionsWidget({
+        service: service,
+        commands: commands,
+        icon: exceptionIcon,
+        tooltip: 'Pause on exception filter'
```
fcollonval, line 38: This needs to be translated:

```suggestion
        tooltip: trans.__('Pause on exception filter')
```

packages/debugger/src/panels/breakpoints/pauseonexceptions.tsx
```diff
@@ -0,0 +1,117 @@
...
+      this.node.getBoundingClientRect().left,
+      this.node.getBoundingClientRect().bottom
+    );
+  };
+
+  render(): JSX.Element {
+    return <ToolbarButtonComponent {...this._props} onClick={this.onclick} />;
+  }
+
+  menu: PauseOnExceptionsMenu;
```
fcollonval, line 73: Why do we need this menu to be public? Could it at least be readonly?

packages/debugger/src/panels/breakpoints/pauseonexceptions.tsx
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) Jupyter Development Team.
```
fcollonval, line 1: Could you please add doc string on the public API?

packages/debugger/test/service.spec.ts
```diff
@@ -96,7 +96,7 @@ describe('DebuggerService', () => {
       path: UUID.uuid4()
     });
     await connection.changeKernel({ name: 'python3' });
-    session = new Debugger.Session({ connection });
+    session = new Debugger.Session({ connection, config });
     config = new Debugger.Config();
```
fcollonval, line 100: Should those two lines be inverted?

fcollonval (changes requested):
Thanks @brichet 

I have some comments.

brichet:
Thanks @fcollonval for review and suggestions.

brichet:
> **NOTE**
> As explained in the issue above, this changes does not allow selecting several filter.
> It should not be a big deal to add this feature, but I'm not sure this is relevant.

Thinking again about this, maybe I should update this PR to allow several filters. Even if this would be useless for `Python` and `debugpy`, it can be useful for other languages.

krassowski:
I think that we should be as close to the UI/UX of established debuggers as possible. If there is a large existing user base expecting the UI to follow established convention we should follow it too (unless we are really innovating on new frontiers like notebook UI). If multi-select checkboxes are standard, let's use checkboxes.

brichet:
I opened https://github.com/jupyterlab/jupyterlab/pull/13621 to be able to use the breakpoints on exception from the command palette with multi-select.

packages/debugger/test/service.spec.ts
brichet: Right, done in https://github.com/jupyterlab/jupyterlab/pull/13601/commits/f3074a505a53283f77116b407ada1b1806e2b08d

packages/debugger/src/panels/breakpoints/pauseonexceptions.tsx
brichet: No need, it's private since https://github.com/jupyterlab/jupyterlab/pull/13601/commits/f3074a505a53283f77116b407ada1b1806e2b08d

fcollonval (commented):
Thanks @brichet 

I tested with the following code:

```py
def f(a, b):
    return a / b

f(2, 0)
```

With two exception type selected (_raised_ and _userHandled_). In that case, the code seems to be executed twice. If I check only one exception type, then it runs only once. Do you know if this is expected?

fcollonval:
Peek 2023-01-05 10-14

brichet:
There is the same behaviors in *VSCode*, it must come from *debugpy*.

[simplescreenrecorder-2023-01-05_16.25.00.webm](https://user-images.githubusercontent.com/32258950/210816507-1965d273-5c33-4993-aa99-e96dfe7d5c44.webm)

fcollonval (approved):
Thanks @brichet

PR #13607, opened 2022-12-16
Bump react 18

brichet:
Bump `react` and `react-dom` from `17.0.1` to `18.2.0`.
Bump `@types/react` from `^17.0.0` to `^18.0.26`
Bump `@types/react-dom` from `^17.0.0` to `18.0.9`

## Code changes

- replaces deleted interface `StatelessComponent` with `FunctionComponent`
- uses the new API of `react-dom` for rendering: https://github.com/reactwg/react-18/discussions/5

## User-facing changes

Not known

Not known

---
Edited: Correct new React version


fcollonval (changes requested):
Thanks a lot for starting this @brichet 

Could you add a note about this in docs/source/extension/extension_migration.rst?
In particular could you mention the official post to upgrade to React 18: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

fcollonval:
The shift of text within the editors is soooooooo unexpected :sob:

fcollonval:
But I guess we can just acknowledge it.

fcollonval:
We may need to look for other ways to simulate event in the tests. I cannot figure out why this is happening with React 18 as I don't see mention about changes in the event handling.

Do you think we could also take advantage of 
https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#react-17-jsx-factories

brichet:
> The shift of text within the editors is soooooooo unexpected sob

> But I guess we can just acknowledge it.

Yes, this is related to codemirror styles, probably [here](https://github.com/codemirror/view/blob/3914fb9ba566cb0d343b869b0da1de7695332368/src/theme.ts#L82-L85).
Not sure we need to overwrite it.

fcollonval:
> Yes, this is related to codemirror styles, probably [here](https://github.com/codemirror/view/blob/3914fb9ba566cb0d343b869b0da1de7695332368/src/theme.ts#L82-L85). Not sure we need to overwrite it.

I agree we should not overwrite it as it is acceptable. But very surprising that upgrading React impacts the editor position.

packages/ui-components/src/components/vdom.ts
```diff
@@ -72,22 +78,33 @@ export abstract class ReactWidget extends Widget {
...
+        this._rootDOM = createRoot(this.node);
+      }
       // Split up the array/element cases so type inference chooses the right
       // signature.
       if (Array.isArray(vnode)) {
-        ReactDOM.render(vnode, this.node, resolve);
+        this._rootDOM.render(vnode);
+        // Resolves after the widget has been rendered.
+        // https://github.com/reactwg/react-18/discussions/5#discussioncomment-798304
+        setTimeout(resolve, 0);
```
brichet, line 90: > We may need to look for other ways to simulate event in the tests. I cannot figure out why this is happening with React 18 as I don't see mention about changes in the event handling.

About the JS tests (if this is the ones you're talking about), the error is that `ReactDOM.render` has been replaced in *react-dom 18*, and does not allow callback function anymore.

This [discussion](https://github.com/reactwg/react-18/discussions/5) explains it, but I didn't find a way to implement it for the `ReactWidget`. I tried the [workarounds](https://github.com/reactwg/react-18/discussions/5#discussioncomment-798304) without success.

It results in trying to click on the widget before it has been fully rendered. It can probably be fixed in tests by waiting for the button, but I don't know if there can be consequences elsewhere, as there will be no more promise to wait for.

fcollonval: Thanks for the pointer. The best path is through the `ref` callback. But this is not acceptable for us (we cannot expect developer to set that for all `ReactWidget`). Could we try something more crude like waiting for `this.node` to have children?

brichet:

packages/ui-components/src/components/vdom.ts
brichet: > The best path is through the `ref` callback

This is what is done in the `vdom` package, but it 's not working for `ReactWidget`.

 > Could we try something more crude like waiting for `this.node` to have children?

I'll try this

brichet:

fcollonval:
> FYI @brichet you trigger all snapshots update with `please update  snapshots`

fcollonval:
> You need to fix the conflict - GitHub is blocking `pull_request` jobs when there are conflicts.

brichet:
@fcollonval do you know if the failing `splice_source` tests is expected ? 
Look like it tries to import some `React 18`'s package in a `React 17` installation.

fcollonval:
I'm not really familiar with the splice feature. The documentation states:

```
If you want to test a source extension against the unreleased versions of JupyterLab, you can run the command

::

    jupyter lab --watch --splice-source

This command will splice the local ``packages`` directory into the application directory, allowing you to build source 
extension(s) against the current development sources.  To statically build spliced sources, use 
``jupyter lab build --splice-source``.  Once a spliced build is created, any subsequent calls to `jupyter labextension build` 
will be in splice mode by default.  A spliced build can be forced by calling ``jupyter labextension build --splice-source``. 
Note that :ref:`developing a prebuilt extension <prebuilt_dev_workflow>` against a development version of JupyterLab is 
generally much easier than source package building.
```

So I guess that as the `yarn.lock` in the `staging` folder is not updated with the latest version, it indeed bring in some incorrect React.

> It could be interesting to raise the point at a weekly meeting to remove that API. As source extension are to be deprecated.

packages/apputils/src/dialog.tsx
```diff
@@ -659,7 +677,7 @@ export namespace Dialog {
       title: Header,
       reject: () => void,
       options: Partial<Dialog.IOptions<T>>
-    ): Widget;
+    ): Promise<ReactWidget>;
```
fcollonval, line 680: I am not a fan of those API changes. Could you try using that API for testing: https://reactjs.org/docs/test-utils.html#act

brichet:
I might have to update the JupyterLab staging `package.json` and `yarn.lock` to pass the remaining tests, even if it's not a release.

packages/apputils/src/dialog.tsx
brichet: I reverted it BTW, I forgot to mention it. And I found another way to make the tests pass for the Dialog.

packages/json-extension/src/component.tsx
```diff
@@ -21,6 +21,11 @@ export interface IProps {
    * The application language translator.
    */
   translator?: ITranslator;
+
+  /**
+   * Function called after the component is rendered
+   */
+  callback?: () => void;
```
fcollonval, line 28: Could you use `React.forwardRef` to avoid using something specific that may be cryptic of common React developer: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components

fcollonval (changes requested):

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -703,7 +703,7 @@ export function ToolbarButtonComponent(
  * Adds the toolbar button class to the toolbar widget.
  * @param w Toolbar button widget.
  */
-export function addToolbarButtonClass(w: Widget): Widget {
+export function addToolbarButtonClass(w: ReactWidget): ReactWidget {
```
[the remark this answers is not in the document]
brichet: Nice thanks @fcollonval

packages/apputils/src/dialog.tsx
```diff
@@ -865,14 +878,21 @@ export namespace Dialog {
         body.node.textContent = value;
       } else if (value instanceof Widget) {
         body = value;
+        Styling.styleNode(body.node);
```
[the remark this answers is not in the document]
brichet: For what I understand the styling applies on the element and its children. 
Not sure if you're talking about `Widget` (selected line) or `reactWidget` here, but both can contain elements on which the style must be applied, no ?
For example [this test](https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/packages/apputils/test/dialog.spec.tsx#L384) expects that the style is applyed on the children of a `ReactWidget`.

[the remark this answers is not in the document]
fcollonval: The test is covered as `value instanceof Widget === false` as `value` is a React component. However, it is valid to create a `Dialog` with:

```
new Dialog({
  body: ReactWidget.create(<p>React body</p>)
});
```

as `ReactWidget` is a `Widget`. In that case condition `value instanceof Widget === true`. But the call at line 881 is likely to not affect the widget children node as the react component will not yet be mounted.

packages/apputils/src/dialog.tsx
```diff
@@ -249,7 +257,7 @@ export class Dialog<T> extends Widget {
   /**
    *  A message handler invoked on an `'after-attach'` message.
    */
-  protected onAfterAttach(msg: Message): void {
+  protected async onAfterAttach(msg: Message): Promise<void> {
```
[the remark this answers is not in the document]
brichet: Again, thanks a lot.

packages/apputils/src/dialog.tsx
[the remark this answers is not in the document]
brichet: OK got it, I thought you were saying that the line was useless.
I will add a condition on it.

brichet:
The documentation UI tests should fail, I was not able to run these tests locally, it was looking for a *notebook* directory which was not uploaded.
I'll update the snapshots when the summary is available.

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -223,12 +223,26 @@ export class Toolbar<T extends Widget = Widget> extends Widget {
...
+      (widget as ReactWidget)
+        .renderPromise!.then(() => {
+          widget.addClass(TOOLBAR_ITEM_CLASS);
+        })
+        .catch(() => {
+          console.error(`Error loading toolbar item ${name}`);
+        });
+    } else {
+      widget.addClass(TOOLBAR_ITEM_CLASS);
+    }
```
fcollonval, line 245: I'm surprised by this as the node on which the class is added exists before the React component is mounted in it. Or is React removing the classes on its root?

brichet: Looks like the test still fails, you were right. It was working locally, probably it takes a bit more time to execute locally and the widget is rendered.

Another solution could be to wait for all the buttons to be rendered before attaching the toolbar to the DOM here:
https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/packages/cell-toolbar/src/celltoolbartracker.ts#L112-L117
Again it works locally, and should be more robust as this is triggered every time the selected cell changes. But this adds some code only for the tests...

Or we can also add a timeout in the tests...

fcollonval: There is a helper for the test `await framePromise` to wait for a frame refresh. Sometimes it is enough to make a test more robust.

brichet: Thanks, I'll try it and revert these changes

fcollonval: But it seems that all Jest unit tests are green without it.

brichet: Yes, it's about UI-test on collapsible heading, one does not work because it expect a toolbar in the snapshot which is not ready when the snapshot is captured.
So the `framePromise` will not be an option finally.

brichet: The test which fails switches quickly between the cells, making the buttons (`ReactWidget`) to render/unmount quickly.
It seems to break something, probably because of the new API using `createRoot` in the `vdom` package.
I suspect the widget is detached while the root is creating.

fcollonval (approved):
Thanks a lot @brichet 

> A new alpha release needs to be done right after merging this to fix the splice tests

fcollonval:
bot please update snapshots

brichet:
Kicking the Ci

brichet:
Thanks @fcollonval for fixing it

Issue #13596, opened 2022-12-14
[vega5-extension] The PNG representation added to the output is not scaled according to the `scaleFactor` option


joaopalmeiro:
Hi! 👋 

## Description

When displaying a Vega-Lite chart in JupyterLab or creating a chart with Altair (using the [mimetype renderer](https://altair-viz.github.io/user_guide/display_frontends.html#altair-s-renderer-framework)), the cell output also contains a PNG representation of the chart. Since this representation is created directly from [one of Vega's View API methods](https://vega.github.io/vega/docs/api/view/#view_toImageURL) with only the type argument defined, the `scaleFactor` corresponds to 1, the default value. However, given that it is possible to define a different `scaleFactor` from the [Vega-Embed options](https://github.com/vega/vega-embed#options), if this value is different from 1, it is "ignored" when creating this representation in PNG, as this value is not passed directly when calling the `.toImageURL()` method. The relevant line of code is [this one](https://github.com/jupyterlab/jupyterlab/blob/v3.5.1/packages/vega5-extension/src/index.ts#L136).

## Reproduce

> You can find this example in [this repository folder](https://github.com/joaopalmeiro/python-playground/tree/master/try-vega5-extension).

1. Create a new notebook and `display()` the following chart:

```python
# First cell
from IPython.display import display
```

```python
# Second cell
display(
    {
        "application/vnd.vegalite.v4+json": {
            "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
            "description": "A simple bar chart with embedded data.",
            "data": {
                "values": [
                    {"a": "A", "b": 28},
                    {"a": "B", "b": 55},
                    {"a": "C", "b": 43},
                    {"a": "D", "b": 91},
                    {"a": "E", "b": 81},
                    {"a": "F", "b": 53},
                    {"a": "G", "b": 19},
                    {"a": "H", "b": 87},
                    {"a": "I", "b": 52},
                ]
            },
            "mark": "bar",
            "encoding": {
                "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
                "y": {"field": "b", "type": "quantitative"},
            },
        }
    },
    metadata={
        "application/vnd.vegalite.v4+json": {
            "embed_options": {
                "scaleFactor": 5,
            }
        }
    },
    raw=True,
)
```

3. Click on the Vega-Embed button in the upper right corner of the chart and download it in PNG format.
4. Check the width and height of the generated image.
[...]

PR #13610, opened 2022-12-16
Add the `scaleFactor` value from the embed options when creating the PNG representation for a Vega-based chart

joaopalmeiro:
## References

#13596 

## Code changes

In the `vega5-extension`, when creating the PNG representation using a Vega's View API method for a Vega-based chart, add the `scaleFactor` parameter from the embed options that can be customized by the user:

```diff
- const imageURL = await this._result.view.toImageURL('png');
+ const imageURL = await this._result.view.toImageURL('png', embedOptions.scaleFactor);
```

## User-facing changes

If the user sets a `scaleFactor` different from `1`, the Base64 string representing the chart will be scaled according to this value. This string will represent a PNG image whose width and height are multiplied by the `scaleFactor` value. Thus, the defined `scaleFactor` is consistent with the user-changeable options available in Vega-Embed (for the user to manually download the chart in PNG format, for example) instead of always being `1`.

None, as far as I know.


joaopalmeiro:
Hi! 👋 

I managed to set up a local environment for JupyterLab from the master branch. However, when I save a notebook after running it with one of the snippets below, the PNG representation is not there (I inspected the notebook as a JSON file and nothing). Any idea what might be causing this or is it a problem with my local environment (if you can't reproduce the same issue with one of the snippets below)? Any thoughts, @fcollonval ?

## Snippets

```python
from IPython.display import display

display(
    {
        "application/vnd.vegalite.v4+json": {
            "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
            "description": "A simple bar chart with embedded data.",
            "data": {
                "values": [
                    {"a": "A", "b": 28},
                    {"a": "B", "b": 55},
                    {"a": "C", "b": 43},
                    {"a": "D", "b": 91},
                    {"a": "E", "b": 81},
                    {"a": "F", "b": 53},
                    {"a": "G", "b": 19},
                    {"a": "H", "b": 87},
                    {"a": "I", "b": 52},
                ]
            },
            "mark": "bar",
            "encoding": {
                "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
                "y": {"field": "b", "type": "quantitative"},
            },
        }
    },
    raw=True,
)
```

```python
from IPython.display import display

display(
    {
        "application/vnd.vegalite.v4+json": {
            "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
            "description": "A simple bar chart with embedded data.",
            "data": {
                "values": [
                    {"a": "A", "b": 28},
                    {"a": "B", "b": 55},
                    {"a": "C", "b": 43},
                    {"a": "D", "b": 91},
                    {"a": "E", "b": 81},
                    {"a": "F", "b": 53},
                    {"a": "G", "b": 19},
                    {"a": "H", "b": 87},
                    {"a": "I", "b": 52},
                ]
            },
            "mark": "bar",
            "encoding": {
                "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
                "y": {"field": "b", "type": "quantitative"},
            },
        }
    },
    metadata={
        "application/vnd.vegalite.v4+json": {
            "embed_options": {
                "scaleFactor": 5,
            }
        }
    },
    raw=True,
)
```

fcollonval:
Thanks @joaopalmeiro for the PR.

The behavior you are seeing is expected. The way notebook viewers are working (JupyterLab, classical Notebook,...) should never change the source of truth (aka the notebook JSON file). What is happening is only at runtime that the output model is extended with an additional PNG. But this is not (and cannot) change the JSON output stored in the file.

---

Regarding testing, you can add a new `test` in https://github.com/jupyterlab/jupyterlab/blob/master/galata/test/jupyterlab/notebook-run-vega.test.ts

It will look something like:

```ts
  test('vega scale factor', async ({ page }) => {
    await page.notebook.createNew();
    await page.notebook.setCell(
      0,
      'code',
      `from IPython.display import display

display(
    {
        "application/vnd.vegalite.v4+json": {
            "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
            "description": "A simple bar chart with embedded data.",
            "data": {
                "values": [
                    {"a": "A", "b": 28},
                    {"a": "B", "b": 55},
                    {"a": "C", "b": 43},
                    {"a": "D", "b": 91},
                    {"a": "E", "b": 81},
                    {"a": "F", "b": 53},
                    {"a": "G", "b": 19},
                    {"a": "H", "b": 87},
                    {"a": "I", "b": 52},
                ]
            },
            "mark": "bar",
            "encoding": {
                "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
                "y": {"field": "b", "type": "quantitative"},
            },
        }
    },
    metadata={
        "application/vnd.vegalite.v4+json": {
            "embed_options": {
                "scaleFactor": 5,
            }
        }
    },
    raw=True,
)`
    );

    page.notebook.run();

    expect(await page.screenshot()).toMatchSnapshot(
      'vega-scaled-figure.png'
    );
  });
```

joaopalmeiro:
Thanks for the clarification and the help, @fcollonval! I will add a test as you suggested.

Off-topic: Sorry for asking this question here, but is this behavior you described recent? I have been working with version 3.2.5 mostly for the last few months and was expecting to find the Base64 string in the notebook after saving it.

fcollonval:
Ok I just tested only on Binder. Testing it will be a bit tricky as it requires to open the figure submenu and request exporting it as PNG.

So the test is gonna be something like:

```ts
  test('vega scale factor', async ({ page }) => {
    await page.notebook.createNew();
    await page.notebook.setCell(
      0,
      'code',
      `from IPython.display import display

display(
    {
        "application/vnd.vegalite.v4+json": {
            "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
            "description": "A simple bar chart with embedded data.",
            "data": {
                "values": [
                    {"a": "A", "b": 28},
                    {"a": "B", "b": 55},
                    {"a": "C", "b": 43},
                    {"a": "D", "b": 91},
                    {"a": "E", "b": 81},
                    {"a": "F", "b": 53},
                    {"a": "G", "b": 19},
                    {"a": "H", "b": 87},
                    {"a": "I", "b": 52},
                ]
            },
            "mark": "bar",
            "encoding": {
                "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
                "y": {"field": "b", "type": "quantitative"},
            },
        }
    },
    metadata={
        "application/vnd.vegalite.v4+json": {
            "embed_options": {
                "scaleFactor": 5,
            }
        }
    },
    raw=True,
)`
    );

    await page.notebook.run();

    await page.click('details[title="Click to view actions"] >> summary');
    const downloadPromise = page.waitForEvent('download');
    await page.click('text=Save as PNG');
    
    const download = await downloadPromise;
    await download.path();

    // TODO how to test the downloaded image
  });
```

fcollonval:
I never tested a downloaded file. So I need to figure out a way... .

joaopalmeiro:
I was checking the updated version of the test, and I'm not sure if we are testing the right thing, I'm sorry. Downloading the chart via the button in the upper right corner of Vega-Embed works with a `scaleFactor` other than 1, as the embed options are passed [here](https://github.com/jupyterlab/jupyterlab/blob/master/packages/vega5-extension/src/index.ts#L127), if I'm correct. The problem is in the Base64 string generated to represent the chart as PNG ([here](https://github.com/jupyterlab/jupyterlab/blob/master/packages/vega5-extension/src/index.ts#L137)) because this part uses the Vega's View API directly, and the second argument of the `toImageURL` method, for the `scaleFactor`, is not defined. So, I think the test should be about this string and whether the dimensions are as expected given the size (width and height) of the chart and the `scaleFactor`. What do you think?

fcollonval:
We could indeed do that. I'll provide snippet for going in that direction as it is gonna be a unit test using Jest rather than an integration test.

For info, when exporting the graph with a scale factor of 1 and 5 (through the drop down menu), I got two different PNG:

image

Foreground graph with scale 1 and background with scale 5

fcollonval:
@joaopalmeiro I took the liberty to push a commit to set up test for the package (as it has none before now). And I added a skeleton to test this feature. Let me know if you have trouble completing the skeleton.

joaopalmeiro:
Hi! 👋 

@fcollonval, added two tests for this functionality. Can you check them, please? Also, I've updated one of the MIME types in the README file from `2` to `3` for the example to work and added some development dependencies that seem to be missing (please confirm).

fcollonval (approved):
Thanks @joaopalmeiro 

> CI failures are not related

fcollonval:

joaopalmeiro:
Thanks! 😄 

Should I close the associated issue or is it not necessary?

fcollonval:
@joaopalmeiro yes please

I did not pay attention that the issue was not linked. You can linked an issue to a PR by using special command words in the PR description: see [documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

PR #10348, opened 2021-06-04
tagging deprecated completer APIs

dharmaquark:
## References

This PR adds deprecated tags to address https://github.com/jupyterlab/jupyterlab/issues/10276 , opening this as a draft as we determine whether additional APIs need to be flagged as deprecated for the 3.1 release.

## Code changes

This PR adds deprecation tags to the following APIs:
**Handler**
- IReply
- connector

**Model**
- items

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

Issue #9763, opened 2021-02-09
Completer public API refactor wishlist


krassowski:
Having good code completion is a big deal for many users. Currently extensions such as jupyterlab-lsp and jupyterlab-kite use multiple of private or technically public but not relay designed as such methods to make the experience better possible. Several limitations exist when it comes to adjusting the completer behaviour (e.g. combining completions from multiple sources, retrieving data asynchronously on the fly) and appearance. This issue is a whishlist complied over the last year of jupyterlab-lsp development and aimed as a reference for my plans for contributions to JupyterLab 4.0. Any early feedback and additional wishlist items are warmly welcome.

### General

- [ ] there are two interfaces for completions; the old one should get removed in 4.0; it should get a deprecation notice in 3.1

### Rendering

- [ ] allow to change rendering of documentation in completer (this is kind-of very much needed for the LSP extension right now and here is an open PR: https://github.com/jupyterlab/jupyterlab/pull/9663; as in practice it is not a breaking change, it would be great to see it in 3.1)
- [ ] allow to add a custom CSS class to the completer widget
   - currently any styling changes applied to the completer are not scoped to the completer modified by a specific extension; imagine extension A (e.g. jupyterlab-lsp) modify completer in notebooks and file editor, but not in the code console, but extension B modify the completion in the code console; their style adjustments with clash; it is possible to add custom class to completion items but it is very heavy on the DOM when numpy/pandas or other large library completions with 1000s of items are displayed
- [ ] allow to easily customize item rendering
   - [ ] allow to add prefix and suffix to label (/separate rendering from filtering and sorting). e.g. `func(a=1, b=2)` should be filtered on `func` but `(a=1, b=2)` should be also displayed; it should be possible to style the parameters separately e.g. to make them grey (less prominent). For prefix it might be desirable to  filter on `method` but display `Object.method` and it also requires different styling to align it properly
       - this might entail more than a change to renderer: separation of `label` and `filterText` might be needed; alternatively the additional data could be passed in extra properties of `ICompletionItem` and label documented to act as a both label and `filterText`. I would prefer to have a separate `filterText` getter with a default returning the label though.
       - the renderer class is using heavily encapsulated approach with most of the stuff being private and not very modular; it might be useful to clean up its internals with re-use in mind (e.g. making some of the internals protected rather than private, which means that extensions that inherit get the benefits of inheritance, but without the core guaranteeing anything about the interface in the future)
[...]

PR #13615, opened 2022-12-18
Remove old completer API

krassowski:
## References

partial for #9763

## Code changes

As in "Backwards-incompatible changes" section

## User-facing changes

None

All completer extensions will need to be refactored for JupyterLab 4.0:
- `CompletionHandler.IOptions.connector` needs to be replaced by `CompletionHandler.IOptions.reconciliator`
- `items`, `setItems()`, `options` and `setOptions()` need to be replaced by `completionItems`, `setCompletionItems()` 
- `Renderer.createItemNode` → `Renderer.createCompletionItemNode`

We started signalling changes to these methods in 3.1 (#10348) but not everything was tagged and not everything has a replacement in 3.x so these are in general breaking changes that need to go into a major release.

### DIFF
--- a/packages/completer/src/handler.ts
+++ b/packages/completer/src/handler.ts
@@ -34,20 +34,7 @@ export class CompletionHandler implements IDisposable {
     this.completer = options.completer;
     this.completer.selected.connect(this.onCompletionSelected, this);
     this.completer.visibilityChanged.connect(this.onVisibilityChanged, this);
-    if (options.connector && options.reconciliator) {
-      console.warn(
-        'Both connector and reconciliator were passed; the connector will be ignored'
-      );
-    }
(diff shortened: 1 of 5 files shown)


jtpio:
Should these changes be added to the extension migration guide?

Or later as part of the API refactor in https://github.com/jupyterlab/jupyterlab/issues/9763? (just checking if this is tracked somewhere)

krassowski:
It is already somewhat covered by mention of `ICompletionProvider` (the changes here are downstream of it in `Completer.IRenderer`, `Completer.IModel`); `CompletionHandler` needs a separate subpoint.

I have added an item to #9763 to track this and will include in a follow up PR focusing on completer performance.

Issue #13452, opened 2022-11-20
shift-left clicking event is blocked when no text is selected


Thenorthernsea:
https://github.com/jupyterlab/jupyterlab/blob/26999c5df85dd8a088580b5dafbc5404ef9b4a44/packages/notebook/src/widget.ts#L2140
This caused select multiple widget in ipywidgets to not functioning as expected when try to use shift + click to select in range.

related issue: https://github.com/jupyter-widgets/ipywidgets/issues/3615

jmk89:
In particular, `window.getSelection()` doesn't detect when the list item in the SelectMultiple widget is selected, or has been selected. The call to `window.getSelection()` detects when something else has been **highlighted.** It will also detect a selection if some text **within** the SelectMultiple widget has been highlighted. 

https://github.com/jupyterlab/jupyterlab/blob/6bfef5d478969a6ac59a0ed65405a2df6cdf9109/packages/notebook/src/widget.ts#L2229-L2237

In this case, it seems more useful to consider `window.getSelection()` as something more like `window.getHighlighted()` :stuck_out_tongue_closed_eyes:

fcollonval:
We could probably avoid processing the event if the target node is `select` or `input`?

What do you think @jasongrout ?
@jmk89 or @Thenorthernsea thanks for digging up the code line. Would you be willing to open a PR fixing this?

jmk89:
@fcollonval sure I can get something put together pretty quick. 

How are people testing ipywidgets against the current master branch? Is there a suggestion on how to get a compatible build of ipywidgets to work with the 4.0.0alpha16 dependencies going on?

fcollonval:
Thanks @jmk89 

You can follow the [development installation](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#installing-jupyterlab). Then in your environment, you can install `ipywidgets`. And finally you will need to start JupyterLab with an additional flag:

```
jupyter lab --dev-mode --watch --extensions-in-dev-mode
```

With that you should be able to modify JupyterLab code with ipywidgets installed.

![image](https://user-images.githubusercontent.com/8435071/207261200-78c5042f-33f5-4f2e-8b18-0c7569259a6f.png)

jmk89:
[...]

PR #13616, opened 2022-12-19
Does not prevent default behavior when shift-clicking

jmk89:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13452

## Code changes

If the selected target is an option from a SelectMultiple widget, or is the text within an input widget - don't prevent the default behavior of the browser upon shift-clicking.

## User-facing changes

Left click a single option within a SelectMultiple widget, then shift-click another option to select a range.
Enter some text in an input widget, left click somewhere, then shift-click somewhere else to select multiple characters.

n/a


packages/notebook/src/widget.ts
```diff
@@ -2231,7 +2231,12 @@ export class Notebook extends StaticNotebook {
       // We don't want to prevent the default selection behavior
       // if there is currently text selected in an output.
       const hasSelection = (window.getSelection() ?? '').toString() !== '';
-      if (button === 0 && shiftKey && !hasSelection) {
+      if (
+        button === 0 &&
+        shiftKey &&
+        !hasSelection &&
+        !['input', 'option'].includes(target.localName)
```
fcollonval, line 2238: Would you mind using `tagName` instead of `localName`:

```suggestion
        !['INPUT', 'OPTION'].includes(target.tagName)
```

fcollonval (changes requested):
Thanks @jmk89 I tested it on Binder it works as expected.

I have one suggestion.

packages/notebook/src/widget.ts
jmk89: Easy enough. Why is tagName preferred over localName?

fcollonval (approved):
Thanks @jmk89 

> CI failure is not related

fcollonval:

PR #13621, opened 2022-12-19
Add an input dialog for multiple selection

brichet:
This PR add an input dialog for multiple selection.

## Reference

This is useful for commands with multiple values as  argument, as it will be in https://github.com/jupyterlab/jupyterlab/pull/13601

## Code changes

- Adds a `InputMultipleItemsDialog` class in `apputils/src/inputdialog.ts`.
- Modifies `apputils/src/inputdialog.ts` and `ui-components/style/styling.css` to display a multi select input as a list and not a select input.
- Adds tests on it.

## User-facing changes

None

None

### DIFF
--- a/packages/apputils/src/inputdialog.ts
+++ b/packages/apputils/src/inputdialog.ts
@@ -163,6 +163,40 @@ export namespace InputDialog {
   }
 
   /**
+   * Constructor options for item selection input dialogs
+   */
+  export interface IMultipleItemsOptions extends IOptions {
+    /**
+     * List of choices
+     */
+    items: Array<string>;
+    /**
+     * Default choices
+     */
+    defaults?: string[];
+  }
+
+  /**
+   * Create and show a input dialog for a choice.
+   *
+   * @param options - The dialog setup options.
+   *
+   * @returns A promise that resolves with whether the dialog was accepted
+   */
+  export function getMultipleItems(
+    options: IMultipleItemsOptions
+  ): Promise<Dialog.IResult<string[]>> {
+    return showDialog({
+      ...options,
+      body: new InputMultipleItemsDialog(options),
+      buttons: [
+        Dialog.cancelButton({ label: options.cancelLabel }),
+        Dialog.okButton({ label: options.okLabel })
+      ]
+    });
+  }
+
+  /**
    * Constructor options for text input dialogs
    */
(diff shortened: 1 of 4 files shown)


packages/apputils/test/inputdialog.spec.ts
```diff
@@ -144,6 +144,48 @@ describe('@jupyterlab/apputils', () => {
...
+
+        await acceptDialog();
+
+        const result = await dialog;
+
+        expect(result.button.accept).toBe(true);
+        expect(result.value).toStrictEqual([]);
+      });
+
+      it('should accept option "currents"', async () => {
```
fcollonval, line 172: The test description does not convey the test goal. I guess you mean `defaults`:

```suggestion
      it('should accept option "defaults"', async () => {
```

packages/ui-components/style/styling.css
```diff
@@ -86,3 +88,16 @@ select.jp-mod-styled {
   -webkit-appearance: none;
   -moz-appearance: none;
 }
+
+select.jp-mod-styled:not([multiple]) {
+  height: 32;
```
fcollonval, line 93: I guess the unit is `px`

```suggestion
  height: 32px;
```

fcollonval (changes requested):
Thanks @brichet

packages/apputils/src/inputdialog.ts
```diff
@@ -459,3 +493,63 @@ class InputItemsDialog extends InputDialogBase<string> {
...
+   * @param options Constructor options
+   */
+  constructor(options: InputDialog.IMultipleItemsOptions) {
+    super(options.label);
+
+    let defaults = options.defaults || [];
+
+    this._list = document.createElement('select');
+
+    // Add an hidden option, otherwise the first one is selected by default
```
krassowski, line 513: This will cause issues for screen readers. Can this be avoided by explicitly setting `.selected = false` on first option or using `select.selectedIndex = -1`? I think it should not be needed but if we have to add hidden option, can we add aria-hidden?

brichet: Agree, thanks for suggestions.
I had tried these suggestions without success some days ago, but now I can't reproduce the problem :smile: Don't know what I did...
So I removed it, and I'll fix it if necessary.

fcollonval (approved):

Issue #13362, opened 2022-11-01
Code folding is broken


schmidi314:
## Description

Code folding does not work any more.

## Reproduce

![Peek 2022-10-15 15-42](https://user-images.githubusercontent.com/74381262/199213274-edf9f3af-3386-4c8e-a8d4-7c282fcc28a7.gif)

1. Activate code folding [Notebook->Code Cell Configuration->Code Folding]
2. Try to collapse a function.

## Expected behavior

A triangle should appear that, when clicked, collapses the function bodies.

BTW, in jupyter notebook one could collapse a whole code cell to its first line when starting with a #-comment. That never worked in jupyterlab and I'm still kind of missing it.

## Context

- Operating System and version: Ubuntu Linux 20.04.5
- Browser and version: Firefox 105.0
- JupyterLab version: 4.0.0a30

andrii-i:
This is mentioned as a task in issue #12812

jtpio:
Many thanks @schmidi314 for trying out the 4.0 pre-release and reporting this issue :+1:

Issue #13199, opened 2022-10-07
Debugger: line numbers are lost when running a cell and advancing


jtpio:
## Description

Line numbers are lost when debugging is enabled and new cells are created (using the run toolbar button).

## Reproduce

https://user-images.githubusercontent.com/591645/194505482-1584af7d-7aef-4d2d-849b-3cc479dd8d2c.mp4

## Expected behavior

Line numbers should be preserved.

Behavior looks fine when debugging is **not** enabled:

https://user-images.githubusercontent.com/591645/194505440-c98e9ff8-77f9-471b-be0a-72fc70224d49.mp4

## Context

- Operating System and version: Fedora 36
- Browser and version: Firefox 105
- JupyterLab version: latest `master`: https://github.com/jupyterlab/jupyterlab/commit/596f682721b66d04e250b7f08f2e8537f51d365c

Probably an issue introduced during the CodeMirror 6 migration: https://github.com/jupyterlab/jupyterlab/pull/11638

firai:
It seems like the issue is specifically with the last inserted cell when the line number setting is **off**, until another cell gets inserted or deleted:

![Line numbers](https://user-images.githubusercontent.com/23555/194604435-d00a279c-205e-4bbc-9331-e06587e44b13.gif)

Issue #13162, opened 2022-10-03
JL 4 regression: JSON settings editor not scrollable by scrollbars or mouse


firai:
## Description

The JSON settings editor CM widgets don't show scrollbars when the contents overflow, and they aren't scrollable by mouse wheel. Arrow keys do work though.

![Json Editor](https://user-images.githubusercontent.com/23555/193504075-8a87d04b-2860-4cfc-8c2e-ee87c54de44d.gif)

## Reproduce

1. Open the JSON settings editor through the setting editor.
2. Open any of the settings sections which have long defaults.

## Expected behavior

JSON editor widgets scrollable by scrollbars and by mouse wheel, in addition to scrolling by keyboard.

## Context

- Operating System and version: Windows 11
- Browser and version: Firefox 105.0.1 and Chrome 106.0.5249.91
- JupyterLab version: 65b4d7e (The bug does not appear in 3.4.7. I'm assuming it's caused by the upgrade to CM 6.)

firai:
@meeseeksdev tag "pkg:settingeditor"

gabalafou:
I found this bug as well while trying to edit keyboard shortcuts. Page up / page down keys also work, though not quite correctly; sometimes they over-scroll, sometimes they under-scroll.

Issue #12886, opened 2022-07-30
Regression with CM6: when kernel cannot be started highlighting errors out


krassowski:
## Description

Opening notebooks with unavailable kernels results in repetitive `TypeError`

## Reproduce

1. Build HEAD with CodeMirror6
2. Open an existing notebook with kernelspec pointing to a non-existing kernel, e.g.:
    ```json
    {
     "cells": [
      {
       "cell_type": "code",
       "execution_count": 1,
       "metadata": {},
       "outputs": [],
       "source": [
        "test = 1\n",
        "es = 2\n",
        "b = 2"
       ]
      }
     ],
     "metadata": {
      "kernelspec": {
       "display_name": "test-kernel-no-available",
       "language": "python",
       "name": "test-kernel-no-available"
      },
      "language_info": {
       "codemirror_mode": {
        "name": "ipython",
        "version": 3
       },
       "file_extension": ".py",
       "mimetype": "text/x-python",
       "name": "python",
       "nbconvert_exporter": "python",
       "pygments_lexer": "ipython3",
       "version": "3.8.3"
      }
     },
     "nbformat": 4,
     "nbformat_minor": 4
    }
    ```
3. Open developer console
4. See error:

```typescript
Uncaught (in promise) TypeError: can't access property "name", mode is null
    findBest mode.ts:1442
    ensure mode.ts:1464
    refresh notebooktools.ts:458
    onActiveCellChanged notebooktools.ts:452
    processMessage notebooktools.ts:345
    invokeHandler index.es6.js:444
    sendMessage index.es6.js:180
    _onActiveCellChanged notebooktools.ts:192
    each index.es6.js:1563
    _onActiveCellChanged notebooktools.ts:191
    invokeSlot index.es6.js:461
    emit index.es6.js:416
```

## Expected behavior

- Null check prevents this error
- A tests ensures we get correct highlighting behaviour in absence of original kernel

While the error is raised from:

https://github.com/jupyterlab/jupyterlab/blob/f573c4a7bd55e3eafbd3667716fe55b431021d9a/packages/codemirror/src/mode.ts#L1441-L1443

We really should not be passing a `null` mode here as we can compose mode object using the metadata from the notebook, in particular `kernelspec.language`, `language_info.codemirror_mode`, etc.

## Context

- Operating System and version: Linux Ubuntu
- Browser and version: Firefox Nightly
- JupyterLab version: HEAD

A task for #12812.

PR #13639, opened 2022-12-21
Use tokens to extend CodeMirror editors

### LINKED ISSUE #13206: Pasting in markdown cells pastes twice (regression in 4.0)

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/12861#pullrequestreview-1052150593
Fixes https://github.com/jupyterlab/jupyterlab/issues/13362
Fixes https://github.com/jupyterlab/jupyterlab/issues/12886
Fixes https://github.com/jupyterlab/jupyterlab/issues/9124
Fixes #13206
Fixes https://github.com/jupyterlab/jupyterlab/issues/13162
Fixes https://github.com/jupyterlab/jupyterlab/issues/13199
Fixes https://github.com/jupyterlab/jupyterlab/issues/14011
Fixes https://github.com/jupyterlab/jupyterlab/issues/10486

## Code changes

CodeMirror 6 can conform to the Token based approach of JupyterLab instead of using globals. The new tokens are:
- `@jupyterlab/codemirror:IEditorExtensionRegistry`: Provide the CodeMirror editor extensions - default extensions are provided in a static method and loaded in the plugin providing the token.
- `@jupyterlab/codemirror:IEditorLanguageRegistry`: Provide the languages for the editors - default languages are provided in a static method and loaded in the plugin providing the token. This allows to override the plugin with much less language for small application for example. A single default language for `text/plain` is provided.
- `@jupyterlab/codemirror:IEditorThemeRegistry`: Provide the theme for the editors - default themes are provided in a static method and loaded in the plugin providing the token. This allows to override the plugin with less themes if required. A single default theme based on _jupyter_ theme CSS variable is provided.

`InputArea.defaultContentFactory` and all its extensions (for Cell and Notebook) have been removed. The editor factory requires tokens. Hence we cannot provide a global default. This is similar to the evolution of the sanitizer.

`TextModelFactory.preferredLanguage(path: string)` is now a no-op (return `''` for any path) as the languages are not available globally and as kernel language for text model makes no sense without more context.

Create a new `Language` for `text/plain` that parses nothing. This allows for a better retro-compatibility as well as a default display without syntax highlighting that is confusing for now on master (see for instance the console header).

_Mode_ editor has been renamed to _language_ to align with CodeMirror 6 naming and to clarify the meaning.

Drop configuration not supported in CodeMirror 6 - `electricChars`, `coverGutterNextToScrollbar`, `scrollbarStyle`, `styleSelectedText` and `selectionPointer`.
Add extensions to support `cursorBlinkRate`, `rulers`, `showTrailingSpaces`, `scrollPastEnd`.
`lineWrap` is now a simple boolean and `wordWrapColumn` is removed.

> All configurable listed in https://github.com/jupyterlab/jupyterlab/issues/12812 are handled (= restored, changed or removed).

Improve code highlight style to be closer to 3.x.

The ipython CodeMirror language is for now identical to the Python language (as highlighting the `?` is a minor feature compare to having code folding).

Upgrade all `@codemirror` and `@lezer` packages

Main point missing from https://github.com/jupyterlab/jupyterlab/issues/12812 is LaTeX highlighting in Markdown cell.

## User-facing changes

None

Break API for CodeMirror Mode (compare to 3.x)


packages/codemirror/src/codemirror-ipythongfm.ts
```diff
@@ -1,18 +1,6 @@
...
-Mode.registerModeInfo({
-  name: 'ipythongfm',
-  mime: 'text/x-ipythongfm',
-  load: () => {
-    return import('@codemirror/lang-markdown').then(m =>
-      m.markdown({ codeLanguages: Mode.getModeInfo() as any })
-    );
-  }
-});
-
```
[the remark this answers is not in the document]
krassowski: I would say we should bring LaTeX highlighting back for ipythongfm mode, as this is feature parity with Notebook v6 and Lab < 4.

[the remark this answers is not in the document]
JohanMabille: I agree, but I think it deserves a dedicated PR, syntax highlighting for CM6 is very different from CM5 and requires a complete rewriting.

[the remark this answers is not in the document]
fcollonval: I kept the file as it may be useful when working to restore syntax highlighting for CM6.

fcollonval:
bot please update snapshots

fcollonval:
bot please update snapshots

galata/test/jupyterlab/texteditor.test.ts
```diff
@@ -20,30 +20,20 @@ test.describe('Text Editor Tests', () => {
...
 
-    await page.waitForSelector(`[role="main"] >> text=${DEFAULT_NAME}`);
+    await page.getByRole('main').getByText(DEFAULT_NAME).waitFor();
 
     await page.menu.clickMenuItem('Settings>Settings Editor');
 
-    await page.waitForSelector('text=Text Editor');
-    await page.click('text=Text Editor');
+    await page
+      .getByRole('tab', { name: 'Editor Text Editor' })
```
afshin, line 28:
```suggestion
      .getByRole('tab', { name: 'Editor Text Editor' })
```

Is this value correct?

packages/cells/src/widget.ts
```diff
@@ -1579,10 +1593,6 @@ export abstract class AttachmentsCell<
     const node = this.node;
     node.addEventListener('lm-dragover', this);
     node.addEventListener('lm-drop', this);
-    node.addEventListener('dragenter', this);
-    node.addEventListener('dragover', this);
-    node.addEventListener('drop', this);
-    node.addEventListener('paste', this);
```
afshin, line 1585: So good that you can remove this!

packages/codeeditor/src/editor.ts
```diff
@@ -311,6 +312,9 @@ export namespace CodeEditor {
 
   /**
    * A widget that provides a code editor.
+   *
+   * As of JupyterLab 4.0.0, it is not possible to provide an editor
+   * that is different of CodeMirror 6.
```
afshin, line 317:
```suggestion
   * other than CodeMirror 6.
```

packages/codeeditor/src/widget.ts
```diff
@@ -263,25 +254,20 @@ export namespace CodeEditorWrapper {
      * A code editor factory.
      *
      * #### Notes
-     * The widget needs a factory and a model instead of a `CodeEditor.IEditor`
-     * object because it needs to provide its own node as the host.
+     * The widget needs a factory and a the editor options
```
afshin, line 257:
```suggestion
     * The widget needs a factory and the editor options
```

packages/codemirror/src/extension.ts
```diff
@@ -0,0 +1,877 @@
...
+  setOption(option: string, value: unknown): void {
+    // Don't bother setting the option if it is already the same.
+    if (this._config[option] !== value) {
+      this._config[option] = value;
+      this._configChanged.emit({ [option]: value });
+    }
+  }
+
+  /**
+   * Set a base config option for the editor.
```
afshin, line 170:
```suggestion
   * Set the base config options for the editor.
```

packages/codemirror/src/extension.ts
```diff
@@ -0,0 +1,877 @@
...
+   *
+   * You will need to reconfigure the editor extensions by listening
```
afshin, line 172:
```suggestion
   * #### Notes
   * You will need to reconfigure the editor extensions by listening
```

packages/codemirror/src/extension.ts
```diff
@@ -0,0 +1,877 @@
...
+  }
+
+  /**
+   * Appends extensions to the top-level configuration of the
+   * editor.
+   *
+   * Injected extension cannot be removed.
+   *
+   * ### Notes
+   * You should prefer registering a IEditorExtensionFactory instead
```
afshin, line 263:
```suggestion
   * You should prefer registering an IEditorExtensionFactory instead
```

packages/codemirror/src/extension.ts
```diff
@@ -0,0 +1,877 @@
...
+   * Whether the editor has an extension for the identifier.
+   *
+   * @param key Extension unique identifier
+   * @returns Extension existence
+   */
+  protected has(key: string): boolean {
+    return this._configurableBuilderMap.has(key);
+  }
+
+  protected getEffect<T>(
```
afshin, line 318:
```suggestion
  /**
   * 
   * @param state The editor state
   * @param key  The extension unique identifier
   * @param value The new value to pass to the extension
   * @returns An editor state effect or `null`.
   */
  protected getEffect<T>(
```

packages/codemirror/src/extension.ts
```diff
@@ -0,0 +1,877 @@
...
+   * Get the default editor extensions.
+   *
+   * @returns CodeMirror 6 extension factories
+   */
+  export function getDefaultExtensions(
+    options: {
+      themes?: IEditorThemeRegistry;
+      translator?: ITranslator | null;
+    } = {}
+  ): IEditorExtensionFactory<any>[] {
```
afshin, line 596:
```suggestion
  ): ReadonlyArray<Readonly<IEditorExtensionFactory<any>>> {
```

I'm not sure what the answer to this question is: If you switch to `ReadonlyArray` of `Readonly` objects, are the `Object.freeze` calls still necessary?

krassowski (commented):
This is amazing :rocket:

Highlighting some integration-level issues from local testing:

- I can no longer navigate cells with the cursor (using up and down keys), whereas it works on `4.0.0a34` (and tip of the master branch)
  | 4.0.0a34 | this PR |
  |---|---|
  | navigation-before | navigation-after |
- the CodeMirror UI settings shows validation error for me (I would love to double check this on binder but #14033)
   Screenshot from 2023-02-18 13-09-29
  - it is still an improvement compared to the state of things on master (https://github.com/jupyterlab/jupyterlab/issues/14011; I edited the description of the PR to link this PR to that issue)
- `smartIndent` does not have a label
- the CodeMirror JSON settings shows "missing schema description"; this is a regression compared to 4.0.0a34 but I am not sure if this works on tip of master branch (again would need a check in a clean environment, #14033):
  | 4.0.0a34 | this PR |
  |---|---|
  | Screenshot from 2023-02-18 13-29-48 | Screenshot from 2023-02-18 13-10-09 |

If you believe that some of the above should be tackled as a follow-up I am happy to approve once these were extracted to separate issues (please ping me for review). Ideally we would add galata tests for the things flagged above.

packages/cells/src/widget.ts
```diff
@@ -304,10 +303,10 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
   /**
    * Editor configuration
    */
-  get editorConfig(): Partial<CodeEditor.IConfig> {
+  get editorConfig(): Record<string, any> {
     return this._editorConfig;
   }
-  setEditorConfig(v: Partial<CodeEditor.IConfig>): void {
+  setEditorConfig(v: Record<string, any>): void {
```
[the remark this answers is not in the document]
fcollonval: I think the reason is that it only do a partial set. I renamed it `updateEditorConfig` to clarify that.

packages/codemirror/src/extension.ts
fcollonval: As far as the [documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) goes, it does not modify the generated code. So we should keep the `Object.freeze` (the documentation example snippet is actually about that). But definitely your suggestion is better reflecting the implementation.

fcollonval:
bot please update snapshots

krassowski:
Is this ready to merge?

bot please update playwright snapshots

schmidi314:
I'm very much looking forward to this merge. But I just noticed that there are some strange defaults set in the code cell settings: Indentation unit is 2 spaces and line numbers are switched on. I'm fine with the line numbers (though it might unnecessarily break user experience) but two-spaces indentation for python is weird.

image

I tried this on binder. It might be related to notebook-extension/schema/tracker.json. It seems the settings have been deleted from there?

krassowski:
Good catch, thank you @schmidi314. I would add two observations that can be moved to a separate issue and addressed in a follow up PR:

1. When switching between tab sizing options the "Default:" shown below appears random:
    default
2. The "Default:" often gets lost for some reason but it does not happen for other groups of settings I tested:
    Peek 2023-02-25 23-09

fcollonval:
bot please update snapshots

fcollonval:
Kicking the CI

fcollonval:
@schmidi314 and @krassowski thanks for the comments - as we are reaching API freezing day, I'll merge this as is. And I opened #14097 as follow-up from your comments

PR #13645, opened 2022-12-21
Turn terminal links into anchors using xterm addon

### LINKED ISSUE #5489: Links in terminal output are no longer turned into anchors
**Describe the bug**

Versions < 0.32.1 would autolink urls in the terminal, new versions no longer creates hyperlinks.  I understand that xterm.js is the underlying terminal library, but I'm hoping someone has a sense for why this functionality regressed.  Potentially it needs to be [activated](https://github.com/xtermjs/xterm.js#addons) now?  I'm hoping it's possible to activate from my own extension.  Otherwise is it possible to override the terminal widget?

**Expected behavior**
Links in the terminal should turn into anchors/

**Desktop (please complete the following information):**

- OS: [MacOS High Sierra]
- Browser [Chrome 69.0]
- JupyterLab [0.35.1]

mgcth:
## References

Fixes #5489.

## Code changes

Adding `xterm-addon-web-links` addon to `packages.json` and to `widget.ts`, and loading it below the `FitAddon`.

## User-facing changes

Turn terminal hyperlinks into anchors users can click and be taken to that link in a new tab.

None, as far as I know.


fcollonval (commented):
Thanks a lot @mgcth 

Would you mind adding a test for this feature?
You can add it in `galata/test/jupyterlab/terminal.test.ts` by starting with a copy of that test: 

https://github.com/jupyterlab/jupyterlab/blob/10c16e8bf8f3451d5a16ef8747c7d803fccb00aa/galata/test/jupyterlab/terminal.test.ts#L92

mgcth:
@fcollonval, thanks for pointing me in the right direction.

Would something like this be enough?

test_mouse_hover

A mouse hover shows it's a link. That's implemented in [91184ff](https://github.com/jupyterlab/jupyterlab/pull/13645/commits/91184ff6fa6ca71367ea96b18baf0b5da62e06fb) (snapshot not yet uploaded).

galata/test/jupyterlab/terminal.test.ts
```diff
@@ -103,3 +103,26 @@ test('Terminal should open in Launcher cwd', async ({ page, tmpPath }) => {
...
+  const terminal = page.locator(TERMINAL_SELECTOR);
+  await terminal.waitFor();
+
+  await page.waitForTimeout(1000);
+  await page.keyboard.type('echo https://jupyter.org/');
+  await page.keyboard.press('Enter');
+  await page
+    .locator('[role="region"] canvas')
+    .nth(3)
+    .click({
```
fcollonval, line 121: Do you mean `hover` instead of `click`?

fcollonval:
> @fcollonval, thanks for pointing me in the right direction.
> 
> Would something like this be enough?
> 
> test_mouse_hover
> 
> A mouse hover shows it's a link. That's implemented in [91184ff](https://github.com/jupyterlab/jupyterlab/pull/13645/commits/91184ff6fa6ca71367ea96b18baf0b5da62e06fb) (snapshot not yet uploaded).

You 🚀 

Thanks a lot for adding the test.

The snapshots are very sensitive especially for terminals as their configuration highly depends on the OS. Therefore to get the correct reference snapshot for the CI OS setup, we use a bot that can be triggered by commenting on a PR:

bot please update snapshots

> It will take ~1h to push a new commit with the update snapshots. In this particular case, it may be though to get the hover effect.

fcollonval:
You can see the status of the bot job there: https://github.com/jupyterlab/jupyterlab/actions/workflows/galata-update.yml

galata/test/jupyterlab/terminal.test.ts
mgcth: Fixed in [cea839d](https://github.com/jupyterlab/jupyterlab/pull/13645/commits/cea839d0cd4e6f4170cdeb5c82deaef9a7ebb678).

mgcth:
Looks like https://github.com/jupyterlab/jupyterlab/pull/13645/commits/38b142c7e39c6060bb57657d41f3c82a5abde81c didn't activate the underline. It does when I run the test locally with both `click` and `hover`. If it makes any difference, I'm running the tests locally on a mac.

galata/test/jupyterlab/terminal.test.ts
```diff
@@ -103,3 +103,26 @@ test('Terminal should open in Launcher cwd', async ({ page, tmpPath }) => {
...
+  await page.waitForTimeout(1000);
+  await page.keyboard.type('echo https://jupyter.org/');
+  await page.keyboard.press('Enter');
+  await page
+    .locator('[role="region"] canvas')
+    .nth(3)
+    .hover({
+      position: {
+        x: 100,
+        y: 20
```
krassowski, line 124: Yes, this will be flaky depending on the font which will vary between operating systems. Let's try:

```suggestion
        x: 80,
        y: 33
```

krassowski:
bot please update snapshots

mgcth:
Is there a way to debug this other than trying it on a Linux machine? I tried creating a Ubuntu 20.04 developer environment through VirtualBox on my mac but during the jupyterlab install, `pip install -e .` failed while installing `canvas`.

krassowski:
Well, maybe a screenshot test is not the way to go forward if it is problematic. We can try two things:
- xterm.js allows to copy text. If we copy the text including link, will we see the URL?
- xterm.js supports accessibility mode via `screenReaderMode` option. We should enable it (potentially under an option). This will expose visible the text content in DOM/accessibility tree.

mgcth:
Trying the exact code `codegen` gives in [61b2f4f](https://github.com/jupyterlab/jupyterlab/pull/13645/commits/61b2f4f80dea3c0ec9f5e11f9890a743a831131b) with _more_ vertical centering.

If it doesn't work, I can try the suggestions but would need some help with where to put those tests.

bot please update snapshots

krassowski:
I checked and neither of the suggestions I proposed works:
- the accessibility tree only includes plain text, no links (this may be unexpected)
- `term.selectAll()` followed by `term.getSelection()` copies plain text, no links (this is probably expected)

Since there is no good way to test this, I would propose:
- let's open an issue in xterm (first double checking if there is not any) proposing that links could be included in accessibility tree
- open an issue in JupyterLab to track adding a test using accessibility mode. This test would probably go to [terminal.spec.ts](https://github.com/jupyterlab/jupyterlab/blob/master/packages/terminal/test/terminal.spec.ts).

mgcth:
@fcollonval, looking at the test report, it seems like your trial succeeds (underline is visible). Thanks for the help!

bot please update snapshots

fcollonval (approved):
Thanks @mgcth

fcollonval:

> I realize I forget to backport this.

PR #13103, opened 2022-09-19
Switch back to `display` to hide tabs

fcollonval:
## References

Fixes #13092
Fixes #13039
Fixes #13002

## Code changes

Use `display` style by default to hide tabs instead of `scale`. CSS transform is breaking the application if tabs contain iframes such as help tabs or PDF viewer.

## User-facing changes

On Firefox user will have slower switches
On Chrome, Chromium and Edge, the application should not break when opening help tab or PDF viewer.

> This changes a settings. So Firefox users may switch it back to `scale` if they want.

## Backwards-incompatible changes

N/A

PR #13652, opened 2022-12-23
Expose `contentVisibility` widget hiding mode

krassowski:
## References

- Allows to improve performance of tab switch (https://github.com/jupyterlab/jupyterlab/issues/4292) in Chromium based browsers.
- In Chromium based browsers also reduces time for switching between consecutive tabs, see https://github.com/jupyterlab/lumino/issues/496#issuecomment-1363431726
- Accompanies https://github.com/jupyterlab/lumino/pull/497

## Code changes

- Rewrite defaults back to `display` (this was partially done in https://github.com/jupyterlab/jupyterlab/pull/13103 but only on config level so the change was only effective after config got loaded possibly adding back-forth switching on load increasing load time; as scale is known to be error-prone on Chromium, I suggest we fully revert to `display` as a default)
- Expose `contentVisibility`

## User-facing changes

- new setting which improves performance if selected

Screenshot from 2022-12-23 02-52-38

None


fcollonval:
@krassowski I lost track of this one. I guess you would like to get it in 3.6.0. If I got it correctly this requires a release of Lumino 2.x and 1.x. And bumping lumino dependencies?

krassowski:
Lumino 2.x beta already includes it. Lumino 1.x needs a release to pull https://github.com/jupyterlab/lumino/pull/500 and we need to bump lumino on both master and 3.6.0 (if this is getting in). It would be nice to have, as the gains on Chromium are really noticeable to end user and we would get feedback from those who opt-in to decide whether we want to make it a default in 4.0.

fcollonval:
Release done: https://github.com/jupyterlab/lumino/releases/tag/v2023.1.19

fcollonval:
@krassowski I really would like to see 3.6.0 release. As fixing Lumino 2 beta PR is gonna take some times, what do you think of backporting this one to 3.6.0; not waiting for merging it in master.

krassowski:
I agree.

fcollonval:
Thanks I'll do the backport if this is ok for you.

krassowski:
Thank you for the backport!

Of note, [Firefox 109](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/109) added support for `content-visibility` so now all major browsers support this. I will run a benchmark, but given that 4.0 release is still some time away I would say we should make it a default and see how it works in bet tests.

krassowski:
> Of note, [Firefox 109](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/109) added support for content-visibility so now all major browsers support this.

Well, it was not released but is still behind a flag. Firefox release page was wrong and was updated yesterday to reflect this (https://github.com/mdn/content/pull/24307).

fcollonval (approved):
Thanks

PR #13654, opened 2022-12-23
Remove empty license field in pyproject.toml

jtpio:
## References

Attempt at fixing the failing `check_release` CI check on `master`: https://github.com/jupyterlab/jupyterlab/actions/runs/3764228169/jobs/6398454511

## Code changes

Update `pyproject.toml`.

## User-facing changes

None

None


jtpio:
Maybe the `ignore` field added to `pyproject.toml` in https://github.com/jupyterlab/jupyterlab/pull/13654/commits/5716b985221dfe1e8c02d691d4954735faed09f1 gets ignored because the releaser also specifies an `ignore` via the CLI: 

https://github.com/jupyter-server/jupyter_releaser/blob/a731158065b3a9a1058ef7f12b42535b332d4177/jupyter_releaser/cli.py#L301

jtpio:
cc @blink1073 who might know a fix for this.

blink1073:
Yeah, you can try overriding `pydist-extra-check-cmds` to see if that fixes it.  If so, we can remove that from releaser.

fcollonval:
CLI options override the one in the configuration file (see [documentation](https://github.com/jwodder/check-wheel-contents#configuration-options)).

In 3886e41 I use another option `select` to skip _W004_ to go around the configuration priority. I'll open an PR to remove the CLI option from `pydist-extra-check-cmds` as it better that each project tune the configuration as needed.

jtpio:
`check_release` is now passing with https://github.com/jupyter-server/jupyter_releaser/pull/487 and the latest commits.

fcollonval (approved):
Thanks @jtpio

pyproject.toml
```diff
@@ -114,6 +114,9 @@ test = [
 ]
 dev = ["build", "pre-commit", "pytest-cov", "coverage", "hatch", "bump2version"]
 
+[tool.check-wheel-contents]
+ignore = ["W002", "W004"]
```
jtpio, line 118: Maybe we could add those to the cookiecutter by default now that it has been removed from the default releaser check.

fcollonval: Yes - actually only W002 is needed.

jtpio: Opened https://github.com/jupyterlab/extension-cookiecutter-ts/pull/275

PR #13657, opened 2022-12-23
Fix `replaceSelection`, add unit test

### LINKED ISSUE #13656: `fileeditor:replace-selection` (and possibly friends) broken on 4.0 alpha
## Description

`fileeditor:replace-selection` gets invoked and codemirror function gets called but text in editor does not get updated. Possibly some yjs-codemirror interaction.

## Reproduce

1. Create a shortcut in advanced settings editor like:
```
{
    "shortcuts": [
    {
        "command": "fileeditor:replace-selection",
        "selector": "body",
        "keys": [
                "Ctrl L"
            ],
        "args": {
            "text": "my text"
            
        }
        
    }
]
}
```
3. Open text editor
4. Press Ctrl + L
6. See nothing happens

## Context

- Browser and version: Chrome 108
- JupyterLab version: 4.0.0a31

krassowski:
## References

Fixes #13656

## Code changes

Modifies text by using `model.sharedModel.updateSource` rather than CodeMirror's `state.replaceSelection`.

## User-facing changes

Commands for replacing text which users use for snippets work again in 4.0 alpha

None.

### DIFF
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -660,10 +660,17 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
   /**
    * Replaces the current selection with the given text.
    *
+   * Behaviour for multiple selections is undefined.
+   *
    * @param text The text to be inserted.
    */
   replaceSelection(text: string): void {
-    this.state.replaceSelection(text);
+    const firstSelection = this.getSelections()[0];
+    this.model.sharedModel.updateSource(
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks a lot @krassowski 

I actually had a fix in a (not yet published) follow-up of #13639 that restore handling of paste and drag-and-drop in attachment cells. But your fix is even better as it uses the model and not the view :heart:

PR #10591, opened 2021-07-09
Use browser's built-in sanitizer to escape HTML in completion labels

krassowski:
## References

Closes #9790.

## Code changes

Replace late sanitization with early escaping.

- this has a downside that we need to be careful to always escape before passing to `_createMatchNode`
  - [x] for example`get completionItems()` does not sanitize currently if there is no query
- I could bring back the extra sanitization step as a safguard. think that it is not needed at the moment, but if we want to be on the safe side it would not hurt much (except for a reduced performance).

There is a small and insignificant performance gain for `onUpdateRequest` tested on numpy (`np.<tab>`) averages over 10 repeats:
- escape only (new): 142.62 ms, SD = 17.85
- sanitize only (old): 156.16 ms, SD = 7.47
- both (alternative): 196.21 ms, SD = 20.17

## User-facing changes

Before:

completion-before

After:

completion-after

## Backwards-incompatible changes

None

PR #13663, opened 2022-12-25
Improve completer rendering performance

### LINKED ISSUE #5911: Autocomplete for large DataFrame objects wedges the UI
Autocomplete appears to take time proportional to the number of object properties, and objects with large properties can have very, very long rendering delays that effective wedge the UI.

An example repro using try.jupyter.com follows.
slow-completion

This demo GIF shows the completion delay for a DataFrame with 10,000 columns. The GIF is accelerated; end-to-end time to calculate and render completions is ~30s. Roughly half appears to be calculating completions at the backend. The remaining half is synchronous Javascript which renders the results returned.

I'm surprised about the backend computation time considering that `get_ipython().complete(...)` takes <1s. Perhaps other methods are being called as well.

I speculate that per-property overhead is expected behavior for the completion code, so perhaps fix will require a an upper bound on the number of properties considered and/or results returned.

For easy copy / pasting, the test snippet I used is:

```
%%time
import pandas as pd
import numpy as np
ipy = get_ipython()
rows, cols = 1000, 1000
df = pd.DataFrame(np.random.randn(rows, cols), columns=['col'+str(i) for i in range(cols)])
#_ = ipy.object_inspect('df')
#_ = ipy.complete('df.')
```

### LINKED ISSUE #9790: HTML/XML markup is removed from the completer label
## Description

The XML/HTML markup is not escaped from the label but instead removed by the sanitizer. When working with HTML/XML or other documents or languages that use `<` and `>` as part of the syntax, these should be preserved.

Screenshot from 2021-02-13 16-25-09

## Reproduce

1. In IPython register custom completion (execute the code below):

```python
from IPython import get_ipython


def implies_competer(ipython, event):
    return ['you will see this <span color="red">but not the markup</span>']


ipython = get_ipython()
ipython.set_hook('complete_command', implies_competer, re_key='.*test_span')
```

2. In a new cell paste: `test_span `  (the trailing space is required)
3. Position cursor after the trailing space: `test_span |`
4. Press Tab
5. See that text shows up but markup is removed

Screenshot from 2021-02-13 16-25-19

## Expected behavior

The markup should instead be escaped properly:

Screenshot from 2021-02-13 16-26-07

as it worked in classic notebook:

Screenshot from 2021-02-13 23-24-02

## Context

- JupyterLab version: 3.0.7

I have a workaround ready for jupyterlab-lsp [here](https://github.com/krassowski/jupyterlab-lsp/blob/9dc9f81bfcadfe825868857bb7bafb0717cb6ad8/packages/jupyterlab-lsp/src/features/completion/model.ts#L92-L106):

### LINKED ISSUE #12995: Completer re-invoke does not scroll to active item in Firefox
## Description

> completer caches the completion list node using lumino virtual DOM.

When triggering completer again in the same editor (cell), the completion list will be scrolled down rather, leaving auto-selected completion items out of view or partially out of view (if at the previous invoke user scrolled the list).

When that happens,

```javascript
document.querySelector('.jp-Completer-list').scrollTop
```

returns a positive integer.

## Reproduce

Peek 2022-08-23 14-00

Reproduced with 3.4.5 in two Firefox versions, but I could not reproduce this in Chromium browsers.

## Expected behavior

List gets scrolled up to the top, like:

```javascript
document.querySelector('.jp-Completer-list').scrollTop = 0
```

## Context

- Operating System and version: Ubuntu 21.04
- Browser and version: Firefox 103.0.2 and Firefox Nightly 105.0a1 (2022-08-13)
- JupyterLab version: 3.4.5

### LINKED ISSUE #13047: Completer rendering performance in Chromium
## Description

Drawing completer can be faster. On my high-end device a small lag can be noticed when scrolling because `setGeometry` takes between 15 and 26ms (full task including painting takes between 18 and 25ms, with first and last one taking longer ~31ms for me), which extends far beyond the ideal 16ms for 60 FPS. 

These results do not account for further delay when more than one notebook is open or when using LSP which adds icons and documentation panel to the completer.

On low end devices (simulated with 6x throttling) the delay is on the order of 90-160ms (full task including painting: 150-180ms, first and last ~270ms).

The HoverBox is used to render completer and uses DOM modifications for positioning which invokes layout changes. This leads to poor performance on Chromium browsers (tested with AMD Ryzen 9, 12 cores). GIFs have low framerate so the lag is not as well visible without throttling.

20-150 ms is not much in itself, but the problem is already visible when scrolling, and it does compound with the time it takes to retrieve completions from kernel (and LSP server), so anything that we can shave off from rendering will benefit the UX (with gains more pronounced for users with low-end CPUs).

| Chrome, no throttle | Firefox no throttle |
| --- | --- |
chrome_no_throttle | firefox_no_throttle |

Chrome 6x throttle:

## Reproduce

1. Open a Python notebook in Chrome
2. Run cell with:
    ```python
    d = {
        'candidate:' + f'{i}': i
        for i in range(100)
    }
    ```
3. In a new cell type `d['can` and open completer
4. Scroll up and down
5. See how slow the updates are (fraction of a second sometimes up to a second)

## Expected behavior

Completer has latency as close to zero as possible. Hover box follows the performance guidelines (which are mostly about preventing unnecessary layout operations):
- https://web.dev/animations-guide
   - Avoid properties that trigger layout or paint: ✗ we could switch from `top` and `left` to `transform`
   - Force layer creation: ✗ we could consider using `will-change` on hoverbox as it indeed can move at any time
- https://web.dev/avoid-large-complex-layouts-and-layout-thrashing:
   - Avoid forced synchronous layouts: ✓ we already use `requestAnimationFrame`
   - Avoid layout thrashing: ✗ (see below)
   - Use flexbox over older layout models: ✗ not clear how much benefit there would be, but it might be that it would only help when 1000s of suggestions are presented (like when using numpy global namespace)

#### Avoid layout thrashing

Here we first modify the DOM (`style.maxHeight`) → query the node dimensions (`getBoundingClientRect`) → again modify the DOM (`style.left` and `style.width`) → query `scrollHeight` → modify `style.width` and `scrollTop`→ query node dimensions again (`getBoundingClientRect`) → again modify the DOM (`style.left`):

https://github.com/jupyterlab/jupyterlab/blob/7f87d89dc88ad86e84fb2245c873139ce60dec3c/packages/ui-components/src/hoverbox.ts#L148-L194

If hoverbox could takes the responsibility of computing height and width (which should be possible if we assume specific width and height of items), we could reduce the layout thrashing substantially.

And here we modify the DOM (add class) → query the element → modify the DOM again (remove class):

https://github.com/jupyterlab/jupyterlab/blob/7f87d89dc88ad86e84fb2245c873139ce60dec3c/packages/ui-components/src/hoverbox.ts#L103-L120

This one is harder: there is AFAIK no better way to perform an accurate hit & visibility test than using `elementFromPoint` and indeed we need to hide the completer for that purpose. It requires more exploration.

#13046 changes this logic, but is neutral when it comes to layout issues; it does introduce extra capabilities which means a slightly longer times may be expected:
- 6x throttle: 176.6ms, 142ms, 150.48ms (total task time) - no noticeable impact
- No throttle: 22.68ms, 21.86ms, 24.57ms (total task time) - little impact if any

## Context

- Operating System and version: Linux Ubuntu 22.04
- Browser and version: Chrome 104
- JupyterLab version: 4.0.0a26

<details><summary>Chrome rendering acceleration report</summary>
<pre>
Graphics Feature Status
Canvas: Hardware accelerated
Canvas out-of-process rasterization: Disabled
Direct Rendering Display Compositor: Disabled
Compositing: Hardware accelerated
Multiple Raster Threads: Enabled
OpenGL: Enabled
Rasterization: Hardware accelerated
Raw Draw: Disabled
Video Decode: Software only. Hardware acceleration disabled
Video Encode: Software only. Hardware acceleration disabled
Vulkan: Disabled
WebGL: Hardware accelerated
WebGL2: Hardware accelerated
WebGPU: Disabled
</pre>
</details>

krassowski:
## References

- fixes #13047 on the tokens benchmark:
   - x8.45 speedup in Chrome,
   - x4.79 speedup in Firefox
- fixes #5911
- fixes #9790
   - supersedes #10591
- fixes #12995

## Code changes

Performance:
- when many items are present (more than fits within hard-coded maximum completer height), pre-compute completer box size so that we can add `contain: strict` style drastically reducing time browsers take to layout completer
   - height is set to maximum height (if there are fewer items than max height, we are not in performance critical path)
   - width is estimated using the "widest" item; the widest item is found using a label length heuristic
      -  this heuristic can be replaced in third-party completers by providing optional `IRenderer.itemWidthHeuristic` 
- lazily populate the completer:
   - initially only as many items are rendered as needed to fill the first "page" (up to max height)
   - subsequent items are added in following frames in chunks
      - we dynamically choose the size of chunks so that they fit in the optimal frame rendering time of ~16ms
      - we avoid introducing any layout shift thanks to margin-bottom trick which sets appropriate height for listing scrollbars
- use `z-index` instead of `jp-mod-outofview` class for temporarily hiding hover box item, accelerating hover box positioning resolution by avoiding extra style re-computation
- use much faster browser-native sanitization method over custom sanitizer since we no longer need to allow `<mark>` tags which are now added later

UI-profiler benchmark comparison - 1000 tokens each 50 characters in text editor:

| browser | state | IQM [Q1 - Q3] | mean | min |
|---|---|---|---|--|
| Chrome | before | 461.0 [448.7 – 474.4] ms | 464.2 ms | 431.4 ms |
| Firefox | before | 269.2 [256.8 – 281.5] ms | 272 ms | 238.1 ms |
| Chrome | after | 54.5 [50.8 – 59.3] ms | 56.7 ms | 42.8 ms |
| Firefox | after | 56.2 [53.3 – 60.3] ms | 58.5 ms | 46.2 ms |

A real-word example (`import numpy as np; np.<tab>` in notebook):

| browser | state | IQM [Q1 - Q3] | mean | min |
|---|---|---|---|--|
| Chrome | before | 414.8 [398.6 – 452.9] ms | 426.2 ms | 374.5 ms |
| Firefox | before | 303.5 [283.3 – 338.9] ms | 320.9 ms | 261.2 ms |
| Chrome | after | 252.6 [235.2 – 277.7] ms | 255.6 ms | 212.5 ms |
| Firefox | after | 288.7 [252.3 – 320.5] ms | 302.6 ms | 230.8 ms |

Other bug-fixes/enhancements:
- properly escape completion items containing HTML or HTML-like code
- the default completer `Renderer` constructor signature was changed to follow `IOptions` pattern
- fixes layout of hover box when two panels are open one above the other (see `spaceAbove`/`spaceBelow` changes)
- remove `sanitizer` from field which was added to `IRenderer` interface in #13341: it is an implementation detail of our default `Renderer`, I think there is no need to require third-party renderers to expose their sanitizer
- re-align the completer widget `MAX_HEIGHT` (was 300 px) and `MIN_HEIGHT` (was 20px) with actual values defined in CSS styles (max-height was 240px, min-height was 24px) keeping CSS as dynamic source of truth (with cache); this means that changes made in themes will be properly reflected in positioning of completer hover box.

## User-facing changes

To allow dynamic (below/above cursor) placement of completer with pre-computed height we need to change the out-of-view behaviour of completer hover box. This means that the completer will no longer be hidden when its border first touch the parent (`hidden-inside` mode); instead it will be dynamically resized (using `stick-inside` mode); the alternative  would be to allow it to overflow slightly over the host node (using `hidden-outside` mode).

The default renderer/styles now use CSS `display: grid` over `display: table`; this is needed to prevent flickering with delayed rendering of invisible parts of completer. This results in smaller width of the completer in some circumstances (when pre-computed dimensions are used):

<details>

| before | after |
|---|---|
| Screenshot from 2022-12-27 14-47-14 |Screenshot from 2022-12-27 14-46-46 |

Se how `VisibleDeprecationWarning` has just enough space to fit next to type information (`class`).

</details>

- [x] `IRenderer.sanitizer` will be breaking unless we quickly remove it before 3.6 release
   - https://github.com/jupyterlab/jupyterlab/pull/13700 thanks @fcollonval
- removal of `jp-mod-outofview` class?


krassowski:

packages/completer/test/model.spec.ts
```diff
@@ -140,6 +151,95 @@
...
+        const request = makeState(currentValue);
+        const change = makeState(newValue);
+        const listener = (sender: any, args: Completer.IQueryChange) => {
+          called[args.origin]++;
+        };
+        model.queryChanged.connect(listener);
+        expect(called.editorUpdate).toBe(0);
+        model.original = request;
+        model.cursor = cursor;
+        model.current = change;
```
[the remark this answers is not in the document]
krassowski: I believe this is a false positive (we invoke the getter twice for side effects), and we do that in tests above too.

packages/completer/src/widget.ts
```diff
@@ -244,68 +300,75 @@ export class Completer extends Widget {
...
     }
+    this._geometryLock = false;
+  }
+
+  /**
+   * Get cached dimensions of the completer box.
+   */
+  protected get sizeCache(): Completer.IDimensions | undefined {
+    if (!this._sizeCache) {
+      return this._sizeCache;
```
trungleduc, line 359:
```suggestion
      return;
```

packages/completer/src/widget.ts
```diff
@@ -564,85 +751,149 @@ export class Completer extends Widget {
...
     if (!resolvedItem) {
-      docPanel.setAttribute('style', 'display:none');
+      this._toggleDocPanel(false);
       return;
     }
     docPanel.textContent = '';
-    docPanel.appendChild(this._createLoadingBar());
+    const loadingIndicator = this._renderer.createLoadingDocsIndicator
+      ? this._renderer.createLoadingDocsIndicator()
+      : this._defaultRenderer.createLoadingDocsIndicator();
```
trungleduc, line 817:
```suggestion
    const loadingIndicator =
      this._renderer.createLoadingDocsIndicator?.() ??
      this._defaultRenderer.createLoadingDocsIndicator();
```

packages/completer/src/widget.ts
```diff
@@ -564,85 +751,149 @@ export class Completer extends Widget {
...
-            this._renderer.createDocumentationNode ??
-            Completer.getDefaultRenderer(this.sanitizer)
-              .createDocumentationNode;
-          node = nodeRenderer(activeItem);
-          docPanel!.textContent = '';
-          docPanel!.appendChild(node);
-          docPanel!.setAttribute('style', '');
+          const node = this._renderer.createDocumentationNode
+            ? this._renderer.createDocumentationNode(activeItem)
+            : this._defaultRenderer.createDocumentationNode(activeItem);
```
trungleduc, line 831:
```suggestion
          const node =
            this._renderer.createDocumentationNode?.(activeItem) ??
            this._defaultRenderer.createDocumentationNode(activeItem);
```

packages/completer/src/widget.ts
```diff
@@ -564,85 +751,149 @@ export class Completer extends Widget {
...
+  /**
+   * The minimum height of a completer widget.
+   */
+  private _minHeight: number;
+
+  private _scrollbarWidth: number;
+  private _docPanelWidth: number;
+  private _docPanel: HTMLElement | undefined;
+  private _geometryLock = false;
+  private _geometryCounter: number = 0;
```
trungleduc, line 894: Could you add a docstring for the `_geometryCounter` variable?

trungleduc (commented):
Thanks @krassowski, such great work! I have some minor comments above.

packages/completer/src/model.ts
```diff
@@ -432,30 +457,57 @@ export class CompleterModel implements Completer.IModel {
...
+      return newItem;
+    }
+    return item;
+  }
+
   /**
    * Reset the state of the model.
    */
   private _reset(): void {
+    const hadQuery = this._query;
```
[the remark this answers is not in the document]
krassowski: No, it is needed. See that in line 489 `this._query` is being reset: `this._query = '';`.

packages/completer/src/model.ts
```diff
@@ -432,30 +457,57 @@ export class CompleterModel implements Completer.IModel {
...
     this._current = null;
     this._cursor = null;
     this._completionItems = [];
     this._original = null;
     this._query = '';
+    this._processedItemsCache = null;
     this._subsetMatch = false;
     this._typeMap = {};
     this._orderedTypes = [];
+    if (hadQuery) {
```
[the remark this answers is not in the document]
krassowski: No, this cannot be done like that, see that in line 489 `this._query` is being reset: `this._query = '';` We had to emit after it was resent.

packages/completer/src/widget.ts
krassowski:
```suggestion
  /**
   * Increasing this counter invalidates previous request to save geometry cache in animation callback.
   */
  private _geometryCounter: number = 0;
```

trungleduc (approved):

fcollonval (approved):
Thanks @krassowski

PR #10282, opened 2021-05-22
Copy cell output to clipboard

cameron-toy:
Edit: Changed context menu entry to "Copy Output to Clipboard" and fixed formatting errors.

copy_output3

## References

Cell Output, a way to copy #2155

## Code changes
- Command added to copy an output to the system clipboard.
- Output areas now include a tracker for the most recently focused output. This enables copying a specific output.

## User-facing changes
- "Copy Output" context menu entry for outputs.

## Notes/todo
- Is copying a single output enough?  Would a context menu item in CodeCells to copy all outputs in that output area be useful?
- I use my own clipboard copying code because I couldn't get the Clipboard API in `@jupyterlab/apputils` to work for images.

PR #13670, opened 2022-12-27
Debugger: copy value of variable to clipboard

brichet:
Adds *copy to clipboard* feature for the value of the variables in the debugger's variables panel.
It's available with a new context menu entry.

## References

- https://github.com/jupyterlab/jupyterlab/issues/10139

## Code changes

- Adds a new command
- Changes the tree widget to be able to copy a nested variable
- Adds a test in *galata / documentation / debugger*

## User-facing changes

Adds a context menu entry in the debugger variables panel.

None

### DIFF
--- a/packages/debugger-extension/schema/main.json
+++ b/packages/debugger-extension/schema/main.json
@@ -30,6 +30,10 @@
       {
         "command": "debugger:render-mime-variable",
         "selector": ".jp-DebuggerVariables-body"
+      },
+      {
+        "command": "debugger:copy-to-clipboard",
+        "selector": ".jp-DebuggerVariables-body li:not(.jp-DebuggerVariables-noCopy)"
       }
     ]
   },
--- a/packages/debugger-extension/schema/main.json
+++ b/packages/debugger-extension/schema/main.json
@@ -33,7 +33,7 @@
       },
       {
         "command": "debugger:copy-to-clipboard",
-        "selector": ".jp-DebuggerVariables-body li:not(.jp-DebuggerVariables-emptyValue)"
+        "selector": ".jp-DebuggerVariables-body"
       }
     ]
   },
--- a/packages/debugger-extension/schema/main.json
+++ b/packages/debugger-extension/schema/main.json
@@ -33,7 +33,7 @@
       },
       {
         "command": "debugger:copy-to-clipboard",
-        "selector": ".jp-DebuggerVariables-body li:not(.jp-DebuggerVariables-noCopy)"
+        "selector": ".jp-DebuggerVariables-body li:not(.jp-DebuggerVariables-emptyValue)"
       }
     ]
   },
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -478,6 +478,19 @@ const variables: JupyterFrontEndPlugin<void> = {
         });
       }
     });
+
+    commands.addCommand(CommandIDs.copyToClipboard, {
+      label: 'Copy to Clipboard',
+      caption: trans.__('Copy value to clipboard'),
+      isEnabled: () => !!service.session?.isStarted,
+      isVisible: () => handler.activeWidget instanceof NotebookPanel,
+      execute: async () => {
+        const value = service.model.variables.selectedVariable!.value;
+        if (value) {
+          navigator.clipboard.writeText(value);
+        }
+      }
+    });
   }
 };
 
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -484,19 +484,10 @@ const variables: JupyterFrontEndPlugin<void> = {
       label: trans.__('Copy to Clipboard'),
       caption: trans.__('Copy text representation of the value to clipboard'),
       isEnabled: () => {
-        const test = (node: HTMLElement) => {
-          const treeTest = !(
-            node
-              .closest('.jp-DebuggerVariables-body li')
-              ?.classList.contains('jp-DebuggerVariables-emptyValue') ?? true
-          );
-          const gridTest = !!node.closest(
-            '.jp-DebuggerVariables-body .jp-DebuggerVariables-grid canvas'
-          );
-          return treeTest || gridTest;
-        };
-        const node = app.contextMenuHitTest(test);
-        return !!service.session?.isStarted && !!node;
+        return (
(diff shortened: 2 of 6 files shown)


packages/debugger-extension/src/index.ts
```diff
@@ -478,6 +478,19 @@ const variables: JupyterFrontEndPlugin<void> = {
         });
       }
     });
+
+    commands.addCommand(CommandIDs.copyToClipboard, {
+      label: 'Copy to Clipboard',
```
krassowski, line 483:
```suggestion
      label: trans.__('Copy to Clipboard'),
```

krassowski, line 483:
```suggestion
      label: trans.__('Copy to Clipboard'),
```

packages/debugger-extension/src/index.ts
```diff
@@ -478,6 +478,19 @@ const variables: JupyterFrontEndPlugin<void> = {
...
+      caption: trans.__('Copy value to clipboard'),
+      isEnabled: () => !!service.session?.isStarted,
+      isVisible: () => handler.activeWidget instanceof NotebookPanel,
+      execute: async () => {
+        const value = service.model.variables.selectedVariable!.value;
+        if (value) {
+          await navigator.clipboard.writeText(value);
```
krassowski, line 490: We may prefer to use the `Clipboard` utiility from apputils.

Ideally we would be able to fetch a MIME bundle and copy rich data (if permitted). This way when pasting a dataframe to e.g. Word we would have a choice to paste as a rich table, markdown table or plain text. We had a similar discussion for "Copy Output to Clipboard" (see https://github.com/jupyterlab/jupyterlab/pull/10282#issuecomment-8472397860) but deferred due to browser support limitations (https://github.com/jupyterlab/jupyterlab/pull/10282#issuecomment-853364314). Since it was over a year ago we can re-evaluate if you are interested (it's fine if not - just a suggestion).

packages/debugger-extension/src/index.ts
```diff
@@ -478,6 +478,19 @@ const variables: JupyterFrontEndPlugin<void> = {
         });
       }
     });
+
+    commands.addCommand(CommandIDs.copyToClipboard, {
+      label: 'Copy to Clipboard',
+      caption: trans.__('Copy value to clipboard'),
```
krassowski, line 484: Maybe:

```suggestion
      caption: trans.__('Copy text representation to clipboard'),
```

Not sure if that would be correct.

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -239,9 +240,11 @@ const VariableComponent = (props: IVariableComponentProps): JSX.Element => {
   return (
     <li
       onClick={(e): Promise<void> => onVariableClicked(e)}
-      onMouseDown={() => {
+      onMouseDown={e => {
+        e.stopPropagation();
         onSelection(variable);
       }}
+      className={value ? '' : 'jp-DebuggerVariables-noCopy'}
```
krassowski, line 247: Maybe more genetic name for this class, e.g. `jp-DebuggerVariables-emptyValue` would be better?

packages/debugger-extension/schema/main.json
```diff
@@ -30,6 +30,10 @@
       {
         "command": "debugger:render-mime-variable",
         "selector": ".jp-DebuggerVariables-body"
+      },
+      {
+        "command": "debugger:copy-to-clipboard",
+        "selector": ".jp-DebuggerVariables-body li:not(.jp-DebuggerVariables-noCopy)"
```
krassowski, line 36: Is it better to hide the menu item if there is no value to copy, or to show it greyed out (disabled)? I think it depends on the size of the context menu (if there are many items hide, if few show but disable).

packages/debugger-extension/src/index.ts
brichet: Thanks, I didn't know about this `Clipboard\` utility.

packages/debugger/src/panels/variables/tree.tsx
brichet: I agree. It can also be useful for the rich variable inspector.

packages/debugger-extension/schema/main.json
brichet: I don't think we can disable a menu entries using the plugin settings.
The command itself does not receive information about the DOM element that renders it.
I don't know if there is a way to get such information from the `isEnabled` option of the command.

packages/debugger-extension/src/index.ts
brichet: I changed it to 'Copy text representation of the value to clipboard'

packages/debugger-extension/schema/main.json
krassowski: > The command itself does not receive information about the DOM element that renders it.

True (which is simultaneously a good and a bad thing), but `isEnabled()` can do perform hit testing using `contextMenuHitTest` to find the relevant node.

krassowski: To clarify, I am not insisting that this is the way to go, just brainstorming.

brichet: Nice, I didn't this function, that's helpful, thanks.

> To clarify, I am not insisting that this is the way to go, just brainstorming.

Currently there is only 2 or 3 entries (respectively tree or grid view) so I agree it's better to disable it.
I did it in https://github.com/jupyterlab/jupyterlab/pull/13670/commits/6b40b6f46e1c5e4dea277185c7896ed09bbe7e45.

The way it was done before was not working for grid view, now the entry is always available, no matter the variable (only in grid view).

krassowski: Well, `DataGrid` has a `hitTest` method too. It may require modifying the mouse handler but it should be rather straightforward as we already do this for the debugger.

brichet: Finally there was an simpler way to do it for both tree view and grid view: checking the `selectedVariable.value` to enable the menu entry or not https://github.com/jupyterlab/jupyterlab/pull/13670/commits/3b56b256bb24538e72490b8794a793b5ec9a2705

I also removed the class `jp-DebuggerVariables-emptyValue` which is now useless, and added a test on the disable entry.

krassowski (approved):
Thank you!

Issue #13505, opened 2022-11-29
CI not triggered when the bot pushes updated snapshots


jtpio:
## Description

As noticed in https://github.com/jupyterlab/jupyterlab/pull/13500#issuecomment-1329592227 and other PRs, CI does not seem to be properly triggered when the bot commits new snapshots and pushes them to the branch.

![image](https://user-images.githubusercontent.com/591645/204476755-ca9ff7be-6984-4da7-9dcf-572d00d7b4b0.png)

![image](https://user-images.githubusercontent.com/591645/204357933-394881a9-3520-48c7-9b76-dfbddfdcc100.png)

## Reproduce

This can most likely be reproduced on a PR by commenting with "bot please update snapshots".

## Expected behavior

All the checks should be running again or at least be visible via the GitHub UI.

## Context

https://github.com/jupyterlab/jupyterlab/pull/13500#issuecomment-1329592227

jtpio:
Re-opening since this still seems to be happening, for example on https://github.com/jupyterlab/jupyterlab/pull/13562:

![image](https://user-images.githubusercontent.com/591645/207550090-b90be76e-6cc7-4edd-92a4-ed791ab43b6a.png)

fcollonval:
Thanks for reopening - I tried to add a new trigger on the bot commenting back. But it does not triggered anything neither.

krassowski:
During https://github.com/jupyterlab/team-compass/issues/169 we agreed to enable lint job requirement to prevent this from leading to merges without passing CI checks (as discussed in #14154). This is how it now looks with the lint job requirement enabled:

![Screenshot from 2023-03-09 07-58-43](https://user-images.githubusercontent.com/5832902/223957705-213606aa-63c9-4256-9ee6-62566f0839b4.png)

jtpio:
Wondering if setting up a personal access token for the JupyterLab bot account (https://github.com/jupyterlab-bot) could help trigger the CI when the commits are pushed?

https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens

Probably that would mean the commits would appear as if they were pushed by https://github.com/jupyterlab-bot.

jtpio:
cc @krassowski looks like this can be closed as the new bot now opens a PR against the fork, and merging that PR correctly triggers the CI?

krassowski:
Yes, let's close it

PR #13671, opened 2022-12-27
Contain the tabs within the tabbar (do not use translation transform)

krassowski:
## References

This is a prerequisite for full adoption of `contain: strict` https://github.com/jupyterlab/lumino/issues/504.

Briefly, the current tab was using `transform: translateY()` style exceeding its box which means that we could not contain painting for it (i.e. it would have no effect).

## Code changes

- Re-implement the bottom border using opt-in approach - elements which need bottom border get it in `::after` pseudo-element rather than existing opt-out approach (element which should not have it is resized up and shifted down to cover the common border).
- Remove unnecessary CSS rule (`.lm-TabBar-tab.lm-mod-current:hover`)

## User-facing changes

Removes the jitter of the tab title when switching the active tab.

Old (see how "Launcher" title moves down upon activation):

old-tabbar

New (see no title movement upon activation):

new-tabbar

I _think_ it can be safely backported, but would appreciate a second opinion.


krassowski:
please update snapshots

krassowski:
please update snapshots

jtpio:
Kicking CI (https://github.com/jupyterlab/jupyterlab/issues/13505).

jtpio:
> I _think_ it can be safely backported, but would appreciate a second opinion.

Looks like it should indeed be fine at first glance. Adding to the 3.6 milestone for consideration.

fcollonval (approved):
Thanks @krassowski

fcollonval:

PR #13672, opened 2022-12-28
Enable strict CSS containment for `MainAreaWidget`

krassowski:
## References

Relates to https://github.com/jupyterlab/lumino/issues/504

This is a smaller version of https://github.com/jupyterlab/lumino/pull/506 which I hope can be backported and included in 3.x.

This PR is not needed on 4.x once https://github.com/jupyterlab/lumino/pull/506 is merged and released (but I am making it against 4.x to trigger visual tests and discuss if we want to backport).

## Code changes

Addition of:

```css
.jp-MainAreaWidget {
  contain: strict;
}
```

## User-facing changes

- Notebook outputs producing html elements with `transform: translate*` styles will not be drawn outside of the notebook boundaries.
- Substantial performance boost (reduction in layout cost) on Chromium browsers

As above.

### DIFF
--- a/packages/apputils/style/mainareawidget.css
+++ b/packages/apputils/style/mainareawidget.css
@@ -23,3 +23,7 @@
   white-space: pre-wrap;
   word-wrap: break-word;
 }
+
+.jp-MainAreaWidget {
+  contain: strict;
+}


krassowski:
please update snapshots

krassowski:
please run benchmark

krassowski:
please update snapshots

krassowski:
It looks like the memory job is failing. Is the lack of comment expected though?

fcollonval:
> It looks like the memory job is failing. Is the lack of comment expected though?

The workflow reporter need to be updated as for now it will only report if successful:

https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/.github/workflows/benchmark-report.yml#L25

fcollonval:
@krassowski do you have a PR targeting 3.6.0?

krassowski:
Since https://github.com/jupyterlab/lumino/issues/504 was merged, we don't need this PR on master branch, we can just re-target this on 3.6, does it sound good?

krassowski:
Superseded by #13811 (simply changing base via GitHub UI did not work, we diverged too much)

PR #13682, opened 2022-12-29
Accessibility: role and aria-label

brichet:
This PR adds role, title or aria-label to several elements of the DOM:

- Adds a title to the responsive opener button
- ~~Adds a role to the Notebook cells~~
- Adds a label to the execution indicator
- Adds labels to the checkbox of the filebrowser

## References

- Related to https://github.com/jupyter/notebook/issues/6671

## Code changes

Adds  an option for the label to the the `ProgressCircle`.

## User-facing changes

None

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -965,6 +965,7 @@ export class CodeCell extends Cell<ICodeCellModel> {
     const model = this.model;
     this.maxNumberOutputs = options.maxNumberOutputs;
 
+    this.node.setAttribute('role', 'textbox');
     // Note that modifying the below label warrants one to also modify
     // the same in this._outputLengthHandler. Ideally, this label must
     // have been a constant and used in both places but it is not done
@@ -1784,6 +1785,7 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     this.addClass(MARKDOWN_CELL_CLASS);
     this.model.contentChanged.connect(this.onContentChanged, this);
     const trans = this.translator.load('jupyterlab');
+    this.node.setAttribute('role', 'textbox');
     this.node.setAttribute('aria-label', trans.__('Markdown Cell Content'));
     // Ensure we can resolve attachments:
     this._rendermime = options.rendermime.clone({
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -965,7 +965,6 @@ export class CodeCell extends Cell<ICodeCellModel> {
     const model = this.model;
     this.maxNumberOutputs = options.maxNumberOutputs;
 
-    this.node.setAttribute('role', 'textbox');
     // Note that modifying the below label warrants one to also modify
     // the same in this._outputLengthHandler. Ideally, this label must
     // have been a constant and used in both places but it is not done
@@ -1785,7 +1784,6 @@ export class MarkdownCell extends AttachmentsCell<IMarkdownCellModel> {
     this.addClass(MARKDOWN_CELL_CLASS);
     this.model.contentChanged.connect(this.onContentChanged, this);
     const trans = this.translator.load('jupyterlab');
-    this.node.setAttribute('role', 'textbox');
     this.node.setAttribute('aria-label', trans.__('Markdown Cell Content'));
     // Ensure we can resolve attachments:
     this._rendermime = options.rendermime.clone({
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -2165,8 +2165,6 @@ export namespace DirListing {
       translator?: ITranslator,
       hiddenColumns?: Set<DirListing.ToggleableColumn>
     ): void {
-      translator = translator || nullTranslator;
-
       fileType =
         fileType || DocumentRegistry.getDefaultTextFileType(translator);
       const { icon, iconClass, name } = fileType;
@@ -2256,9 +2254,13 @@ export namespace DirListing {
       );
       checkboxInput?.setAttribute(
         'aria-label',
-        `Select ${
-          fileType.contentType === 'directory' ? 'directory' : 'file'
-        } '${highlightedName}'`
+        trans.__(
+          trans.gettext(
+            "Select %1 '%2'",
+            fileType.contentType === 'directory' ? 'directory' : 'file',
+            highlightedName
+          )
+        )
       );
 
       let modText = '';
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -2260,11 +2260,11 @@ export namespace DirListing {
         let ariaLabel: string;
         if (fileType.contentType === 'directory') {
           ariaLabel = selected
-            ? trans.__('Unselect directory "%1"', highlightedName)
+            ? trans.__('Deselect directory "%1"', highlightedName)
             : trans.__('Select directory "%1"', highlightedName);
         } else {
           ariaLabel = selected
-            ? trans.__('Unselect file "%1"', highlightedName)
+            ? trans.__('Deselect file "%1"', highlightedName)
             : trans.__('Select file "%1"', highlightedName);
         }
         checkbox.setAttribute('aria-label', ariaLabel);
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1982,7 +1982,7 @@ export namespace DirListing {
         );
         checkboxInput?.setAttribute(
           'aria-label',
-          'Select all files and directories'
+          trans.__('Select all files and directories')
         );
         node.appendChild(checkboxWrapper);
       }
@@ -2254,13 +2254,9 @@ export namespace DirListing {
       );
       checkboxInput?.setAttribute(
         'aria-label',
-        trans.__(
-          trans.gettext(
-            "Select %1 '%2'",
-            fileType.contentType === 'directory' ? 'directory' : 'file',
-            highlightedName
-          )
-        )
+        fileType.contentType === 'directory'
+          ? trans.__('Select directory "%1"', highlightedName)
+          : trans.__('Select file "%1"', highlightedName)
       );
 
       let modText = '';
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -812,9 +812,9 @@ export class DirListing extends Widget {
       const trans = this.translator.load('jupyterlab');
       checkAllCheckbox?.setAttribute(
         'aria-label',
-        trans.__(
-          '%1 all files and directories',
-          allSelected || someSelected ? 'Unselect' : 'Select'
+        allSelected || someSelected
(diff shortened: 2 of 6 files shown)


packages/ui-components/src/components/toolbar.tsx
```diff
@@ -966,9 +966,11 @@ class ToolbarPopupOpener extends ToolbarButton {
       icon: ellipsesIcon,
       onClick: () => {
         this.handleClick();
-      }
+      },
+      tooltip: 'More commands'
```
krassowski, line 970: Needs translations (and labels need it too).

packages/filebrowser/src/listing.ts
```diff
@@ -2237,12 +2244,23 @@ export namespace DirListing {
...
 
+      // Adds an aria-label to the checkbox element.
+      const checkboxInput = checkboxWrapper.querySelector(
+        'input[type="checkbox"]'
+      );
+      checkboxInput?.setAttribute(
+        'aria-label',
+        `Select ${
+          fileType.contentType === 'directory' ? 'directory' : 'file'
+        } '${highlightedName}'`
```
krassowski, line 2261: Will need to use translations with gettext formatting for arguments.

packages/ui-components/src/components/toolbar.tsx
brichet: Thanks @krassowski

packages/filebrowser/src/listing.ts
brichet: I don't know how and when to use `gettext`. Is that something like below ?
```
trans.__(trans.gettext('Select %1 \'$2\'', fileType.contentType === 'directory' ? 'directory' : 'file', highlightedName))
```

krassowski: I think that gettext does not need to be explicitly invoked, the following should work according to our documentation:

```ts
trans.__('Select %1 "%2"', fileType.contentType === 'directory' ? trans.__('directory') : trans.__('file'), highlightedName)
```

But I would actually recommend moving the branching logic out of translation string as we would not be able to translate nouns in context.

```ts
fileType.contentType === 'directory' ? trans.__('Select directory "%1"') : ? trans.__('Select file "%1"')
```

This is because "file" or "directory" may be translated to multiple words depending on context (e.g. here in Polish I would use "plik" but elsewehere it would mean "kartoteka"). Joining them together makes context collisions less likely and easier for translators to understand what is happening.

brichet: Thanks a lot for the suggestions and details

jtpio (approved):
Thanks!

cc @krassowski if you want to have another look at it. Feel free to merge if this looks good to you :+1:

packages/cells/src/widget.ts
```diff
@@ -957,6 +957,7 @@ export class CodeCell extends Cell<ICodeCellModel> {
     const model = this.model;
     this.maxNumberOutputs = options.maxNumberOutputs;
 
+    this.node.setAttribute('role', 'textbox');
```
krassowski, line 960: 1. I am not convinced whether this should be added on `jp-Cell` node which includes the execution counter and outputs:
    Screenshot from 2023-01-02 14-18-48
    setting it further down on `jp-InputArea-editor`/`jp-Editor` seems more appropriate to me but I may be wrong here:
     Screenshot from 2023-01-02 14-19-25
2. The issue says *Codemirror: missing aria-label to div with role* but I do see aria-role on CodeMirror div (since https://github.com/codemirror/view/commit/93f45d12293d6373c724534d64bd97cfd959cc95):
    Screenshot from 2023-01-02 14-21-14
    was the accessibility test done with this CodeMirror version?
3. I see this added on code cell and markdown cell. If (1) and (2) are incorrect, then raw cell would need it too.

krassowski (commented):
If we want forward with merging, I would suggest to take the `role=textbox` part off and create issue to discuss what needs to be done first (unless you are confident that this is the way forward).

packages/filebrowser/src/listing.ts
```diff
@@ -1977,6 +1977,13 @@ export namespace DirListing {
         const checkboxWrapper = this.createCheckboxWrapperNode({
           alwaysVisible: true
         });
+        const checkboxInput = checkboxWrapper.querySelector(
+          'input[type="checkbox"]'
+        );
+        checkboxInput?.setAttribute(
+          'aria-label',
+          trans.__('Select all files and directories')
```
krassowski, line 1985: Ideally this would be updated depending on the state ("Select all" or "Deselect all" if all are selected)

packages/filebrowser/src/listing.ts
```diff
@@ -1977,6 +1977,13 @@ export namespace DirListing {
         const checkboxWrapper = this.createCheckboxWrapperNode({
           alwaysVisible: true
         });
+        const checkboxInput = checkboxWrapper.querySelector(
+          'input[type="checkbox"]'
+        );
```
krassowski, line 1982: This actually seems to be breaking things.

```ts
TypeError: Cannot read properties of null (reading 'querySelector')
    at Renderer.updateItemNode (listing.ts:2252:45)
    at listing.ts:817:16
    at Array.forEach (<anonymous>)
    at DirListing.onUpdateRequest (listing.ts:814:11)
    at DirListing.processMessage (index.es6.js:1222:1)
    at invokeHandler (index.es6.js:400:1)
    at sendMessage (index.es6.js:136:1)
    at runMessageLoop (index.es6.js:451:1)
```

checkboxes-buggy

krassowski (changes requested):
Sorry for piece-by-piece review, it's still holiday time for me.

packages/cells/src/widget.ts
brichet: Thanks @krassowski.

1. I agree that the *role* should probably be a different one for the `jp-Cell` node, especially since it contains itself a `textbox`. I'll open an issue to discuss that, the related one is on Notebook repository. 
Nevertheless, a role seems necessary on `jp-cell` node to keep the *aria-label* attribute and respect the WAI-ARIA standard.
2. Yes there is a *role* but no *aria-label*.
3. Right

packages/filebrowser/src/listing.ts
brichet: Updated in https://github.com/jupyterlab/jupyterlab/pull/13682/commits/4b532287f77c9280eeb6d0ff22e69b30813c9dc1

packages/filebrowser/src/listing.ts
brichet: Good catch, thanks. Fixed in https://github.com/jupyterlab/jupyterlab/pull/13682/commits/4b532287f77c9280eeb6d0ff22e69b30813c9dc1

packages/filebrowser/src/listing.ts
```diff
@@ -808,6 +808,15 @@ export class DirListing extends Widget {
...
       checkAllCheckbox.dataset.checked = String(allSelected);
       checkAllCheckbox.dataset.indeterminate = String(someSelected);
+
+      const trans = this.translator.load('jupyterlab');
+      checkAllCheckbox?.setAttribute(
+        'aria-label',
+        trans.__(
+          '%1 all files and directories',
+          allSelected || someSelected ? 'Unselect' : 'Select'
+        )
```
krassowski, line 818: 'Unselect' and 'Select' are missing translation.


I think that when possible it is always preferable to give translators more control. "Select all" may be a single word or subject to declension different than "Deselect all". I believe that variable substitution should be only used for numbers and arbitrary inputs (e.g. file names).

```suggestion
        allSelected || someSelected
            ? trans.__('Deselect all files and directories')
            : trans.__('Select all files and directories')
        )
```

packages/filebrowser/src/listing.ts
```diff
@@ -2237,12 +2246,35 @@ export namespace DirListing {
...
+            ? trans.__(
+                '%1 directory "%2"',
+                selected ? 'Unselect' : 'Select',
+                highlightedName
+              )
+            : trans.__(
+                '%1 file "%2"',
+                selected ? 'Unselect' : 'Select',
+                highlightedName
+              )
```
krassowski, line 2273: Again, 'unselect' and select are not translated (they would be substituted directly). "Deselect" appears more common form by the way.

packages/filebrowser/src/listing.ts
brichet: Thanks @krassowski, I thought the variables were translated using variable substitution.

packages/cells/src/widget.ts
brichet: I opened https://github.com/jupyterlab/jupyterlab/issues/13704 related to discuss this point.

packages/filebrowser/src/listing.ts
```diff
@@ -2237,12 +2245,32 @@ export namespace DirListing {
...
+      if (checkbox) {
+        let ariaLabel: string;
+        if (fileType.contentType === 'directory') {
+          ariaLabel = selected
+            ? trans.__('Unselect directory "%2"', highlightedName)
+            : trans.__('Select directory "%2"', highlightedName);
+        } else {
+          ariaLabel = selected
+            ? trans.__('Unselect file "%2"', highlightedName)
+            : trans.__('Select file "%2"', highlightedName);
```
krassowski, line 2268:
```suggestion
            ? trans.__('Unselect directory "%1"', highlightedName)
            : trans.__('Select directory "%1"', highlightedName);
        } else {
          ariaLabel = selected
            ? trans.__('Unselect file "%1"', highlightedName)
            : trans.__('Select file "%1"', highlightedName);
```

Currently it shows:

Screenshot from 2023-01-04 22-58-11

packages/filebrowser/src/listing.ts
krassowski: Thanks!

packages/filebrowser/src/listing.ts
krassowski: might be worth adding a test to [packages/filebrowser/test/listing.spec.ts](https://github.com/jupyterlab/jupyterlab/blob/master/packages/filebrowser/test/listing.spec.ts)

brichet: Sorry for this leftover, I did it too quickly.

SylvainCorlay (approved):

SylvainCorlay:
I would be in favor of moving forward with this - even with the current role.

- it is almost surely a net improvement over not having one specified at all.
- if the role is not ideal, we can revisit at any point in the future and include the change in a patch release.

brichet:
Thanks @SylvainCorlay 

> I would be in favor of moving forward with this - even with the current role.

The role on cells has been removed in this PR anyway. 
The related discussion is in https://github.com/jupyterlab/jupyterlab/issues/13704.

SylvainCorlay:
> > I would be in favor of moving forward with this - even with the current role.
>
> The role on cells has been removed in this PR anyway.
> The related discussion is in https://github.com/jupyterlab/jupyterlab/issues/13704.

Great, then we should probably move forward with this then!

krassowski (approved):
Thanks!

PR #13685, opened 2022-12-30
Upgrades Xterm to v. 5

### LINKED ISSUE #13649: Upgrade XTerm to the latest
We should update xterm.js to the latest version: https://github.com/xtermjs/xterm.js/releases/tag/5.1.0

It has the benefit of being lighter and using the more efficient webgl renderer.

JasonWeill:
## References

Fixes #13649. Upgrades the Xterm dependency to version 5, modifies the code to make up for the removed `setOptions` function in it.

## Code changes

Updates Terminal package to use the new XtermJS 5.x API.
In [55cc717](https://github.com/jupyterlab/jupyterlab/pull/13685/commits/55cc717657e31e62e578df65cab298b5b100af17), WebGL addon is used if the browser supports it; otherwise fallback to canvas addon (as prior to this PR).
The asynchronous loading has changed as playing with import of file within our package may easily breaks what was done. So I [Frederic] went for a pattern used in more places where we asynchronously import the external library; this induced the need to buffer messages from the session waiting for the terminal to be loaded.

> If neither WebGL nor canvas addon is used, the terminal is rendered as html tags. This is to be avoided for performance reason.

## User-facing changes

None.

None.


jtpio:
Thanks @JasonWeill.

Let's trigger a snapshot update since the UI tests seem to be failing.

jtpio:
bot please update snapshots

fcollonval:
@JasonWeill would you mind rebasing this PR taking into account the new xterm addon added in https://github.com/jupyterlab/jupyterlab/pull/13645

JasonWeill:
@fcollonval Rebased, rebuilt, and updated this branch.

fcollonval:
bot please update snapshots

krassowski (dismissed):
Looks great! It seems that there is a regression related to selecting text in terminal. When you select anything the selection is not visible in the light theme (and barely visible in dark theme, though it might have been this way before?). It might be related to the rename of `ITheme.selection` to `ITheme.selectionBackground` in xterm.js 5.0

krassowski:
I think that we need to change this interface and all usages:

https://github.com/jupyterlab/jupyterlab/blob/d4688fe784c2b7d8cb9af8a153af88c91fd8dcc8/packages/terminal/src/tokens.ts#L163-L170

JasonWeill:
> Looks great! It seems that there is a regression related to selecting text in terminal. When you select anything the selection is not visible in the light theme (and barely visible in dark theme, though it might have been this way before?). It might be related to the rename of `ITheme.selection` to `ITheme.selectionBackground` in xterm.js 5.0

@krassowski I just updated the code and docs; the selections now have high contrast again

fcollonval:

fcollonval:

krassowski:
Looking again at the [changelog for 5.0](https://github.com/xtermjs/xterm.js/releases/tag/5.0.0), I wanted to highlight two things:
- there is now also `ITheme.selectionInactiveBackground`, should we add it too?
- (just for awareness) canvas is no longer the default renderer (and required code is not even installed by default):
> The canvas renderer has moved into an addon, significantly reducing the bundle size of the xterm module. This used to be the default renderer and is now recommended only as a fallback to the webgl addon if that does not work for some reason (eg. no webgl2 support).

fcollonval:
Kicking the CI

fcollonval:

fcollonval:
bot please update snapshots

packages/terminal/src/widget.ts
```diff
@@ -61,66 +68,65 @@ export class Terminal extends Widget implements ITerminal.ITerminal {
...
       });
-    }
+  }
 
-    // Only run this initial connection logic once.
-    this.session.connectionStatusChanged.disconnect(
-      this._initialConnection,
-      this
-    );
+  get ready(): Promise<void> {
```
fcollonval, line 128:
```suggestion
  /**
   * A promise that is fulfilled when the terminal is ready.
   */
  get ready(): Promise<void> {
```

fcollonval (approved):
Thanks @JasonWeill

JasonWeill:
>     * there is now also `ITheme.selectionInactiveBackground`, should we add it too?

This is now added.

>     * (just for awareness) canvas is no longer the default renderer (and required code is not even installed by default):

The logic in the `terminal` package seems to favor the WebGL renderer if it can be used; do we need to do anything else as part of the upgrade?

fcollonval:
> The logic in the `terminal` package seems to favor the WebGL renderer if it can be used; do we need to do anything else as part of the upgrade?

Yes, I did it in my rebase. Starting with xterm.js 5, the default renderer is to produce a real HTML tree. This is really inefficient (for what I saw every character is wrap in a span tag). So we need to load one of the renderer by default. The code I pushed is loading the WebGL renderer by default as it is favor over Canvas. But canvas is still used in case the browser does not support WebGL.

fcollonval:
Thanks @JasonWeill  - the CI failures are not related.

PR #13691, opened 2023-01-02
Remove empty button in the notebook toolbar

### LINKED ISSUE #13680: Removing some leftovers class and settings
There may be some leftovers since https://github.com/jupyterlab/jupyterlab/pull/10730, which introduced a execution indicator, including a kernel status.

### Problem

There is an empty button in the toolbar.

image

This button come from a setting in `panel.json`: https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/notebook-extension/schema/panel.json#L27

This toolbar component refer to an item which is not registered anymore in the `toolbarRegistry`: https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/apputils/src/toolbar/widget.tsx#L177

### Proposed Solution

Delete the plugin settings entry at least, but maybe also remove the unused `KernelStatus` class. I just want to make sure it has no more use.

cc @trungleduc

trungleduc:
## References

Fixes #13680 

## Code changes

- Remove the leftover kernel status button in the notebook toolbar.
- Add deprecation notice for `createKernelStatusItem`


packages/apputils/src/toolbar/widget.tsx
```diff
@@ -108,6 +108,8 @@ export namespace Toolbar {
   /**
    * Create a kernel status indicator item.
    *
+   * @deprecated since v3.5
+   * This is dead code now.
```
jtpio, line 112: Maybe an alternative comment could be to mention the execution status indicator instead?

Since extension authors might still be using this method in their extension.

jtpio: Also should this be deprecated in 3.6 and removed in 4.0? Or just deprecated?

trungleduc: I'm not sure about the plan of removing deprecated functions in this namespace, we have other deprecated items  since 3.2 

https://github.com/jupyterlab/jupyterlab/blob/1d36ce596c48401320e6c33b2a4d17d724db1943/packages/apputils/src/toolbar/widget.tsx#L40-L45

https://github.com/jupyterlab/jupyterlab/blob/1d36ce596c48401320e6c33b2a4d17d724db1943/packages/apputils/src/toolbar/widget.tsx#L61-L66

fcollonval (dismissed):
Thanks @trungleduc 

> I created https://github.com/jupyterlab/jupyterlab/issues/13719 to track the need to clean deprecated code for 4.0.0

packages/apputils/src/toolbar/widget.tsx
jtpio: Ah ok, maybe we can do a pass on the docstrings at some point so they are more informative than "dead code" :+1:

fcollonval:
bot please update snapshots

fcollonval:
bot please update snapshots

fcollonval:

brichet:
Thanks @trungleduc

Issue #13629, opened 2022-12-20
CI Feature: Implement pull request preview environments


waveywaves:
### Problem

No pull request previews on jupyterlab

### Proposed Solution

I would like to support Jupyterlab by implementing [Uffizzi](https://github.com/UffizziCloud/uffizzi) preview environments.
Disclaimer: I work on [Uffizzi](https://github.com/UffizziCloud/uffizzi).

Uffizzi is a Open Source full stack previews engine and our platform is available completely free for Jupyterlab (and all open source projects). This will provide maintainers with preview environments of every PR in the cloud, which enables faster iterations and reduces time to merge. You can see the open source repos which are currently using Uffizzi over [here](https://uffizzi.notion.site/)

Uffizzi is purpose-built for the task of previewing PRs and it integrates with your workflow to deploy preview environments in the background without any manual steps for maintainers or contributors. We can go ahead and create an Initial PoC for you right away if you think there is value in this proposal.

### Additional context

We are working on a PoC for Jupyter Notebook as well, our issue has been accpeted on the repo. You can check the discussion on the issue here https://github.com/jupyter/notebook/issues/6661

jweill-aws:
We currently use Binder for pull request preview environments, but if you feel that Uffizzi would be better, please circulate a pull request so that we can consider using it instead.

PR #13699, opened 2023-01-03
feat: create environment for every PR using Uffizzi

daramayis:
This PR is a part of https://github.com/jupyterlab/jupyterlab/issues/13629 and helps demonstrate Uffizzi integration with **Jupyterlab**. 

This will help create preview environments on all of **Jupyterlab** Pull Requests and help contributors including the maintainers iterate faster on their PRs. 
I have created a PR over https://github.com/daramayis/f-jupyterlab/pull/1#issuecomment-1369667030 to show what it looks like to have a preview environment deployed against your PR.

You can check out the preview over [here](https://app.uffizzi.com/github.com/daramayis/f-jupyterlab/pull/1) as well.

@waveywaves


jasongrout:
Thanks, this indeed looks very fast and nice.

On the service side, https://www.uffizzi.com/pricing indicates that if we started using this in this open source project, we'd have unlimited concurrent previews, etc.? Is that right?

jasongrout:
@daramayis - someone pointed out that when I go to the PR review at https://pr-1-deployment-10033-f-jupyterlab.app.uffizzi.com/lab, I see artifacts from a previous person trying it out. This is concerning for two reasons:

1. The other person trying things out may affect the state of things so that a review is not a true clean fresh copy of the PR
2. It's a vector for someone to be malicious - someone could embed something malicious in the preview that is not visible in the actual PR

I'd love to hear your thoughts about how the PR previews preserve state across multiple people accessing them.

PR #13707, opened 2023-01-03
Faster rendering of the debugger tree

### LINKED ISSUE #13694: Debugger freezes UI on namespace updates when there are many variables
## Description

When debugger is open and has many variables, the UI freezes on user actions while the debugger is updating. It seems that events get intercepted and re-invoked by `react-invokeguardedcallback` while the debugger is refreshing.

Screenshot from 2023-01-02 23-08-36

<details>

Reproduction on production version:

Screenshot from 2023-01-02 23-46-04

</details>

the `handlers.mousedown` blocks is the re-emitted event for switching cell. `performWorkUntilDeadline` is the problematic React code.

When zooming in it appears that `VariableComponent` is at least partially to blame:

Screenshot from 2023-01-02 23-13-09

It **always** renders two icons:
- collapser (`ui-components:caret-down-empty`) which is almost never used (except for nested variables and scopes)
- the loop icon (`ui-components:search`) which is only used on hover

Since the icon system does not cache these icons and still uses typestyle it quickly leads to slow rendering.

Given the code, I suspect the loop implementation to be causing React invalidation since it is nested in a button with complex logic.

https://github.com/jupyterlab/jupyterlab/blob/3d7445fd142fedf65b9ba26df3b4bc649b168054/packages/debugger/src/panels/variables/tree.tsx#L240-L305

## Reproduce

1. activate the debugger
1. create three cells, one with `z = 1` and one with:
    ```python
    for i in range(1000):
        globals()[f'x{i}'] = 'y'
    ```
2. run the cell populating globals, wait until debugger populates
3. run the other code cell with `z = 1` (trigger update of the debugger panel by adding a single new variable)
4. immediately click on the third cell (multiple times for better effect)
   - or perform another action like switching to another tab
6. see JupyterLab freezes of ~7 seconds per 1000 variables with React dev mode, or ~2s per 1000 variables in production
   > Note: the performance is worse when using React dev mode which appears to have been [fixed in newer version of React](https://github.com/facebook/react/pull/21666).
8. disable debugger, see that UI works fine now

## Expected behavior

Debugger does not freeze the JupyterLab UI.

## Context

- JupyterLab version: ~4.0.0a31 master

krassowski:
## References

Fixes #13694, as a side-effect fixes a bug with multiple loop buttons showing when hovering over nested branches.

| Firefox 108 | IQM [Q1 - Q3] | mean | min |
|---|---|---|---|
| before | 4261.1 [3760.6 – 4932.3] ms | 4326.3 ms | 2197 ms |
| after | 906.7 [838.4 – 1005.5] ms | 1240.8 ms | 780 ms |

| Chrome 108 | IQM [Q1 - Q3] | mean | min |
|---|---|---|---|
| before | 4306.2 [4181.5 – 4500] ms | 4609.7 ms | 4046.3 ms |
| after | 1482 [1406.8 – 1555.8] ms | 1685.8 ms | 1306.2 ms |

<details>

UI profiler with default settings for new debugger scenario (https://github.com/jupyterlab/ui-profiler/pull/25):

Screenshot from 2023-01-04 21-20-17

</details>

Before

Screenshot from 2023-01-04 21-33-44

After

Screenshot from 2023-01-04 21-34-14

## Code changes

- [x] only ever render one loop button per entire tree
- [x] do not render the collapser icon when the item cannot be expanded
- [x] rename `VariablesComponent` to `VariablesBranch` (it was too confusing to navigate `VariableComponent` and `VariablesComponent` when one is nested in the other, would likely lead to CSS errors with time if left as-is)

## User-facing changes

| Before | After |
|---|---|
| before | after|

Only in CSS styles


krassowski:
please update snapshots

krassowski:
please update snapshots

packages/debugger/style/variables.css
```diff
@@ -16,56 +15,96 @@
   flex: 1 1 auto;
   min-height: 24px;
   overflow: auto;
+
+  /* For absolute positioning of jp-DebuggerVariables-buttons relative to jp-DebuggerVariables-branch. */
```
krassowski, line 19:
```suggestion
  /* For absolute positioning of jp-DebuggerVariables-buttons. */
```

packages/debugger/style/variables.css
```diff
@@ -16,56 +15,96 @@
...
+  grid-area: name;
+}
+
+.jp-DebuggerVariables-name::after {
+  content: ':';
+  margin-right: 5px;
+}
+
+.jp-DebuggerVariables-detail {
+  /* detail contains value for primitive types of type otherwise */
```
krassowski, line 83:
```suggestion
  /* detail contains value for primitive types or name of the type otherwise */
```

krassowski:

krassowski:

fcollonval:
Kicking the CI

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -106,9 +127,153 @@ export class VariablesBodyTree extends ReactWidget {
...
+    const current = ++stateRefreshLock;
+    if (!data.variable) {
+      // Handle mouse leave.
+      if (current !== stateRefreshLock) {
+        return;
+      }
+      const target = data.target;
+      if (
+        target &&
+        // Note: Element, not SVGElement to permit entering <svg> icon.
```
krassowski, line 186:
```suggestion
        // Note: Element, not HTMLElement to permit entering <svg> icon.
```

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -106,9 +127,153 @@ export class VariablesBodyTree extends ReactWidget {
...
+      ) {
+        // Allow to enter the buttons.
+        return;
+      }
+      setVariable(null);
+    } else {
+      // Handle mouse over
+      if (data.variable === null) {
+        console.log('target', data.target);
+      }
```
krassowski, line 198:
```suggestion
      // Handle mouse over.
```

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -106,9 +127,153 @@ export class VariablesBodyTree extends ReactWidget {
...
+const TreeButtons = (props: ITreeButtonsProps): JSX.Element => {
+  const { commands, service, translator, handleSelectVariable } = props;
+  const trans = (translator ?? nullTranslator).load('jupyterlab');
+
+  const [buttonsTop, setButtonsTop] = useState<number>(0);
+  const [variable, setVariable] = useState<IDebugger.IVariable | null>(null);
+
+  let stateRefreshLock = 0;
+
+  const handleHover = (_: VariablesBodyTree, data: IHoverData) => {
```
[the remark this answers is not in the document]
krassowski: Thanks! I added `useCallback` hook in [4b3385a](https://github.com/jupyterlab/jupyterlab/pull/13707/commits/4b3385a751e9e9dd7888cf558d977a6197a33169).

krassowski:
please update snapshots

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -106,9 +127,150 @@ export class VariablesBodyTree extends ReactWidget {
...
+   */
+  variable: IDebugger.IVariable | null;
+}
+
+interface ITreeButtonsProps {
+  /**
+   * The commands registry.
+   */
+  commands: CommandRegistry;
+  service: IDebugger;
```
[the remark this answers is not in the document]
krassowski: Done in https://github.com/jupyterlab/jupyterlab/pull/13707/commits/db717c544800e540c6ddff1a7e832cacecc498ba.

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -124,6 +289,11 @@ interface IVariablesComponentProps {
    * Callback on variable selection
    */
   handleSelectVariable?: (variable: IDebugger.IVariable) => void;
+  /**
+   * Callback on mouseOver/mouseLeave event.
+   */
+  onHoverChanged?: (data: IHoverData) => void;
+  collapserIcon: JSX.Element;
```
fcollonval, line 296:
```suggestion
  /**
   * Collapser icon component
   */
  collapserIcon: JSX.Element;
```

packages/debugger/src/panels/variables/tree.tsx
```diff
@@ -195,6 +375,21 @@ interface IVariableComponentProps {
    * Callback on selection
    */
   onSelect?: (variable: IDebugger.IVariable) => void;
+  /**
+   * Callback on mouseOver/mouseLeave event.
+   */
+  onHoverChanged?: (data: IHoverData) => void;
+  collapserIcon: JSX.Element;
```
fcollonval, line 382:
```suggestion
  /**
   * Collapser icon component
   */
  collapserIcon: JSX.Element;
```

fcollonval (approved):
Thanks @krassowski 

I added two minor suggestions to have a docstring on all props. But those are not mandatory.

PR #13714, opened 2023-01-04
Pin check release action without pyproject validation

### LINKED ISSUE #13713: Check release is not working on 3.6.x
## Description

CI check release is broken on 3.6.x as it check `pyproject.toml` that is actually not used for packaging.

fcollonval:
## References

Fixes #13713

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -19,13 +19,8 @@ jobs:
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
-      - shell: bash
-        run: |
-          # Install releaser prior to action - otherwise it uses an unwanted version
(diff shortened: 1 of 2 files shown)


fcollonval:
@blink1073 I'm struggling to use an old version of the check release action (that does not include the extra check for python package) or to remove the extra check cmds when calling action v2:

https://github.com/jupyter-server/jupyter_releaser/blob/main/jupyter_releaser/cli.py#L297

What is the best path to go around this?

fcollonval:
> The trouble with the older version (1.0.1) tried in this PR is the unauthenticated calls to GH API in `build-changelog`:

UserWarning: Neither GITHUB_TOKEN nor GITHUB_JWT_TOKEN found: running as unauthenticated

This may result in rate limit error

blink1073:
Setting `pydist-extracheck-cmds` makes sense to me.

PR #13715, opened 2023-01-04
Fix environment activation on Gitpod

jtpio:
## References

On Gitpod it looks like the environment should now be explicitly activated after installing the base dependencies.

This might have been a change in behavior with the mamba 1.0 release.

## Code changes

Gitpod setup.

## User-facing changes

None

None


jtpio:
Getting this one in so it can be used by other contributors and on new PRs.

PR #9282, opened 2020-11-04
Remove ensure-max-old-space script

jasongrout:
## References

This change dropping the workaround for Node 10 was originally done in Dec 2019 in #7677, but then reverted in #8112 (to support node 10 in jlab 2.1).

## Code changes

This was a workaround for node 10 that allowed many users to compile JupyterLab without running out of memory (especially if they had lots of extensions installed). It is no longer needed for Node 12 or higher, where memory management is automatic. Node 10 is now in maintenance mode,  (since May 2019) and node 12 is also almost in maintenance mode (starting Nov 2020), so I think we can drop the Node 10 workaround.

Users still on Node 10 will still be able to manually increase their available memory by setting the environment variable NODE_OPTIONS=--max-old-space-size=4096, for example.

PR #13722, opened 2023-01-05
Upgrades Python to 3.10, Node to 18

### LINKED ISSUE #13720: Upgrade dependencies from NodeJS 14 to Node 18
## Description

NodeJS 14.x reaches end of life on April 30, 2023. Source: https://github.com/nodejs/release#release-schedule

Node 14 is mentioned in the following places:

https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/.gitpod.yml#L27

https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/readthedocs.yml#L7

https://github.com/jupyterlab/jupyterlab/blob/6d208970f0b950a1c75c0c08960c651b32a3cc0b/binder/environment.yml#L9

## Expected behavior

A currently-supported version of Node, such as 16 or 18, should be used. Prefer Node 18 for maximum remaining life.

JasonWeill:
## References

Fixes #13720.

## Code changes

Updates dependencies from Node 14 to Node 18. Node 14 and 16, two long-term support (LTS) releases, are reaching end-of-life (EOL) in 2023. The only LTS release not reaching EOL this year is Node 18, which reaches EOL on 2025-04-30.

Also updates documentation to state that Node 18, not Node 12, is the minimum version.

## User-facing changes

None.

None.


JasonWeill:
Looks like the failures in CI are because the agents are running Node 16, not Node 18

dev_mode/package.json
```diff
@@ -211,7 +211,7 @@
     "yarn-deduplicate": "^5.0.0"
   },
   "engines": {
-    "node": ">=12.0.0"
```
jtpio, line 214: Maybe we could indeed update the Node version used on CI, but leave this one to `>=14.0.0`?

jupyterlab/staging/package.json
```diff
@@ -211,7 +211,7 @@
     "yarn-deduplicate": "^5.0.0"
   },
   "engines": {
-    "node": ">=12.0.0"
```
jtpio, line 214: Normally the files under `jupyterlab/staging` don't need to be updated manually (done during the release process). So we can revert that change.

dev_mode/package.json
jtpio: or maybe we require Node 16 now.

Node 14 is already in maintenance according to https://nodejs.dev/en/about/releases/

image

dev_mode/package.json
```diff
@@ -211,7 +211,7 @@
...
+    "node": ">=18.0.0"
```
jtpio, line 214:
```suggestion
    "node": ">=16.0.0"
```

dev_mode/package.json
JasonWeill: Normally, long-term support (LTS) releases end support on April 30, so one would expect Node 16 to be in LTS until 2024-04-30. However, due to a dependency on a soon-to-be-EOL version of OpenSSL, Node 16 is ending LTS on 2023-09-11, just under 8 months from now. That's why I was recommending that we adopt 18 instead of 16.

jtpio: Right, it would be great to encourage folks to use Node 18 if they want to build JupyterLab.

The previous change happened in https://github.com/jupyterlab/jupyterlab/pull/9282, with a more conservative bump from 10 to 12 instead of going to 14 directly.

Potential concerns could be with repo2docker / Binder which usually have Node in the default environment. Linking to https://github.com/jupyterhub/repo2docker/pull/983 for reference.

jtpio: Although now end users are less likely to need Node on their machine. Before it was used more often because source extensions had to be installed with the `jupyter labextension install` command.

jtpio:
> Looks like the failures in CI are because the agents are running Node 16, not Node 18

Maybe we can override the node version in the `base-setup` action for now? Or undraft https://github.com/jupyterlab/maintainer-tools/pull/174?

jtpio:
Kicking CI with a rebase now that https://github.com/jupyterlab/maintainer-tools/pull/174 has been merged and released.

jupyterlab/staging/package.json
```diff
@@ -211,7 +211,7 @@
...
+    "node": ">=18.0.0"
```
jtpio, line 214:
```suggestion
    "node": ">=12.0.0"
```

jtpio:
Maybe we can also bump the `canvas` dependency.

jtpio:
> Maybe we can also bump the `canvas` dependency.

Doing one more rebase since this was done in https://github.com/jupyterlab/jupyterlab/pull/13783

jtpio:
Making some progress now that https://github.com/jupyterlab/maintainer-tools/pull/174 and https://github.com/jupyterlab/jupyterlab/pull/13783 have been merged.

However it looks like we are facing the following issue with `@types/node`: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924

https://github.com/jupyterlab/jupyterlab/pull/13722/commits/a77634fb9a6f14f9eebceead0913b66321d107b3 drops `node-fetch` because `fetch` is now available in Node 18. It was initially marked as experimental but the warning seems to be gone now according to the latest release notes: https://nodejs.org/en/blog/release/v18.13.0/

jtpio:
For reference linking a PR that updates `repo2docker` (used by Binder) to Node 18: https://github.com/jupyterhub/repo2docker/pull/1234

dev_mode/package.json
jtpio: Thinking about it more, updating to Node 18 directly sounds good, since this will be part of JupyterLab 4. Thanks @JasonWeill :+1:

jtpio:
FYI @JasonWeill I pushed a couple of changes and CI now seems to be passing.

If the changes look ok to you we can mark the PR as ready for review.

jtpio:
> If the changes look ok to you we can mark the PR as ready for review.

Going ahead and undrafting it now, but feel free to comment here if needed!

JasonWeill:
@jtpio Thanks for updating the PR! Apologies for not getting back to you sooner about this.

packages/testing/package.json
```diff
@@ -50,8 +50,8 @@
   },
   "devDependencies": {
     "@types/jest": "^29.2.0",
-    "@types/node": "^14.6.1",
-    "@types/node-fetch": "^2.5.4",
+    "@types/node": "^18.11.18",
+    "@types/node-fetch": "^2.6.2",
```
fcollonval, line 54: I guess we should drop `node-fetch` and its typing here too.

jtpio: It was added back in https://github.com/jupyterlab/jupyterlab/pull/13722/commits/ab8710e8252df3f7d545660538c46173c4831d4c for now

jtpio: Mostly because of the test failure of the previous commit, with the intent of dropping it later on but not necessarily in this PR to avoid blocking it.

fcollonval (approved):
Thanks @JasonWeill and @jtpio

PR #13730, opened 2023-01-06
Wrap kernel message binary buffers in DataView

davidbrochart:
## References

Closes https://github.com/martinRenou/ipycanvas/issues/311.

## Code changes

Kernel message binary buffers used to be deserialized as `DataView`s, prior to https://github.com/jupyterlab/jupyterlab/pull/11841 where they are `ArrayBuffer`s. This change restores the original type.

## User-facing changes

None.

None.


fcollonval (approved):
Thanks @davidbrochart

PR #13506, opened 2022-11-29
Depend on `@jupyter/ydoc` instead of `@jupyter-notebook/ydoc`

jtpio:
## References

Follow-up to https://github.com/jupyter-server/jupyter_ydoc/pull/97

## Code changes

Depend on `jupyter/ydoc` instead of `jupyter-notebook/ydoc` now that the package has been renamed and published.

## User-facing changes

None

## Backwards-incompatible changes

A different package will be in the shared scope, but this should only affect alpha releases of 4.0 and 3.6.

PR #9500, opened 2020-12-22
Update default sharing config to be strictVersion: true.

jasongrout:
## References

See the discussion at https://github.com/jupyterlab/jupyterlab/issues/9496

## Code changes

By default, webpack has strictVersion default to false if singleton is true or if import is false. We want strictVersion to be true always by default.

Setting this as 3.1, as I won't have time to adequately test ramifications of this change before 3.0, and I think it won't be a huge disruptive change anyway.

PR #13735, opened 2023-01-09
Updates jupyter_ydoc

hbcarlos:
There is a small issue with the latest alpha published. The dependency `@jupyter/ydoc`, is a singleton package wich version should match the version used by an extension. The problem is that JupyterLab pins `@jupyter/ydoc` v0.2.0 but is using 0.2.2 and when loading an extension that uses `@jupyter/ydoc`, it fails saying the dependency is not in the shared context.

## Code changes
Pins `@jupyter/ydoc` v0.2.2.

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -16,7 +16,7 @@
     "watch": "webpack --watch"
   },
   "resolutions": {
-    "@jupyter/ydoc": "~0.2.0",
+    "@jupyter/ydoc": "~0.2.2",
     "@jupyterlab/application": "~4.0.0-alpha.17",
     "@jupyterlab/application-extension": "~4.0.0-alpha.17",
     "@jupyterlab/apputils": "~4.0.0-alpha.17",


fcollonval (approved):
Thanks @hbcarlos

jtpio:
> It fails, saying the dependency is not in the shared context.

Does it fail, or is it a warning in the dev tools console?

jtpio:
Normally the shared packages should be quite permissive for now, and not using `strictVersion` as experimented in https://github.com/jupyterlab/jupyterlab/pull/9500.

hbcarlos:
> Does it fail, or is it a warning in the dev tools console?

It fails and doesn't load the extension.

jtpio:
Would you mind posting the logs / error for reference? 

It can also be useful later if someone encounters something similar.

jtpio:
Hmm, maybe it could be related to the rename to `@jupyter/ydoc`? https://github.com/jupyterlab/jupyterlab/pull/13506

PR #13745, opened 2023-01-10
Drop support for Python 3.7

### LINKED ISSUE #13739: Drop support for Python 3.7
### Problem

With https://github.com/jupyterlab/jupyterlab/pull/13718 JupyterLab 4 now depends on a stable release of Jupyter Server 2.

Jupyter Server 2 dropped support for Python 3.7 in https://github.com/jupyter-server/jupyter_server/pull/1035.

### Proposed Solution

We can follow the lead of Jupyter Server 2 and also drop support for Python 3.7 in JupyterLab 4.

The EOL for Python is 3.7 Jun 2023: https://endoflife.date/python

### Additional context

Adding to the `4.0` milestone for consideration.

jtpio:
## References

Fixes #13739

## Code changes

Require Python 3.8+

## User-facing changes

End users will need Python 3.8 to be able to install JupyterLab 4.

N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ name = "jupyterlab"
 description = "JupyterLab computational environment"
 readme = "README.md"
 license = { file = "LICENSE" }
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 authors = [
     { name = "Jupyter Development Team", email = "jupyter@googlegroups.com" },
 ]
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @jtpio 

As a matter of fact, depending on `jupyter-server>=2.0.0` already enforce that requirement. So let's go for it.

Issue #13728, opened 2023-01-06
Cell toolbar gets broken with undo


krassowski:
### Reproduce

1. Create a markdown cell
2. Duplicate it
3. Edit the copy
4. Undo a few times
5. See toolbar is no longer visible and errors show up in console:

```js
index.ts:467 TypeError: Cannot set properties of undefined (setting 'text')
    at m._setValue (jsoneditor.ts:331:19)
    at set source [as source] (jsoneditor.ts:134:10)
    at e.CellMetadataEditorTool._update (notebooktools.ts:672:19)
    at e.CellMetadataEditorTool.onActiveCellChanged (notebooktools.ts:660:12)
    at e.CellMetadataEditorTool.processMessage (notebooktools.ts:348:16)
    at c (index.ts:530:15)
    at Object.e [as sendMessage] (index.ts:220:7)
    at notebooktools.ts:195:7
    at f (iter.ts:227:9)
    at P._onActiveCellChanged (notebooktools.ts:194:5)
```

```js
index.ts:349 TypeError: Cannot read properties of null (reading 'contentChanged')
    at d._removeToolbar (celltoolbartracker.ts:146:11)
    at d._onActiveCellChanged (celltoolbartracker.ts:70:12)
    at u (index.ts:673:12)
    at n.emit (index.ts:594:9)
    at n.emit (index.ts:208:15)
    at set activeCellIndex [as activeCellIndex] (widget.ts:1290:31)
    at oe.onCellInserted (widget.ts:1826:10)
    at oe._insertCell (widget.ts:601:12)
    at widget.ts:506:16
    at f (iter.ts:227:9)
```

![toolbar-breaking](https://user-images.githubusercontent.com/5832902/210979948-3beb8375-fc76-4239-9049-9475568a3e58.gif)

### Context

- Browser: Chrome 108
- JupyterLab version: 3.x

krassowski:
Again, it works fine on 4.x alpha

krassowski:
So this was fixed on master in an unrelated PR by adding a check if `model` is not null:

https://github.com/jupyterlab/jupyterlab/blob/1cda7df0e0005ac51fea4bbb8d68f44f6eb1e216/packages/cell-toolbar/src/celltoolbartracker.ts#L66-L68

But this is on some level wrong because the cell model should never be `null` according to types, but in practice it gets set to null on disposal:

[...]

PR #13763, opened 2023-01-12
Explain why cell model may be missing in cell toolbar

krassowski:
## References

https://github.com/jupyterlab/jupyterlab/issues/13728 (should fix on backport, master is not affected as it has a check already)

## Code changes

Previously we were checking for cell model presence on master and doing nothing on 3.x. This PR intends to document why we the model may be missing (disposed cell) and upon backporting harmonise the behaviour.

## User-facing changes

None for master.

None


fcollonval (approved):
Thanks @krassowski

fcollonval:

PR #13765, opened 2023-01-13
Added config to link ts source maps

3coins:
## Problem

The current extension dev build does not link the typescript source files, this causes the debugger in browser to only show the compiled JS files which makes it harder to find the exact line of code during debug sessions. 

## Solution
For extension builds, added configuration to load source maps from compiled JS files. This correctly links the source maps to the original typescript source code during debugging sessions.

Screen Shot 2023-01-13 at 8 48 40 PM

### DIFF
--- a/builder/package.json
+++ b/builder/package.json
@@ -56,6 +56,7 @@
     "mini-svg-data-uri": "^1.4.4",
     "path-browserify": "^1.0.0",
     "process": "^0.11.10",
+    "source-map-loader": "~1.0.2",
     "style-loader": "~3.3.1",
     "supports-color": "^7.2.0",
     "terser-webpack-plugin": "^5.3.6",
--- a/builder/src/extensionConfig.ts
+++ b/builder/src/extensionConfig.ts
@@ -251,6 +251,16 @@ function generateConfig({
     filename += '?v=[contenthash]';
   }
 
+  const rules: any = [{ test: /\.html$/, type: 'asset/resource' }];
+
+  if (mode === 'development') {
+    rules.push({
+      test: /\.js$/,
+      enforce: 'pre',
+      use: ['source-map-loader']
+    });
+  }
+
   const config = [
     merge(
       baseConfig,
@@ -268,7 +278,7 @@ function generateConfig({
       webpackConfig,
       {
         module: {
-          rules: [{ test: /\.html$/, type: 'asset/resource' }]
+          rules
         }
       }
     )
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -124,6 +124,7 @@
   },
   "devDependencies": {
     "lerna": "^6.2.0",
+    "source-map-loader": "4.0.1",
```
jtpio, line 127: Should this be added to `builder/package.json` instead of the top-level `package.json`?

3coins: Thanks for that suggestion. Let me try that.

jtpio: Right since the changes above are in the `builder` it looks like the dependency should be added to that package.

builder/src/extensionConfig.ts
```diff
@@ -268,7 +268,17 @@ function generateConfig({
       webpackConfig,
       {
         module: {
-          rules: [{ test: /\.html$/, type: 'asset/resource' }]
+          rules:
```
fcollonval, line 271: Would you mind extracting this prior to filename? Something like

```
const rules = [
   // Default rules
   ...
]

if (mode === 'development') {
  rules.push([
     // Specific rules
     ...
  ]);
}

The idea is to ease the understanding of the code especially if the list of default rules increases in the future.

**Review by fcollonval — COMMENTED** [review]
Thanks a lot @3coins 

I have a minor style suggestion.

**Comment by 3coins** [comment]
@jtpio @fcollonval 
Thanks for the suggestions, I have made all updates. I have also submitted an update to the `extension-cookiecutter-ts` 4.0 branch to make the default dev build generate source maps; this will avoid adding `sourceMaps: true` to the `tsconfig`. 

I have a quick question about the dependency on `source-map-loader` package. Since, I had to link the JLab changes locally for the extension package, I wasn't able to get the `source-map-loader` package available in the extension package despite moving the dependency out of "devDependencies" in builder. I am not sure if this is a symptom of the local linking or would be the expected behavior in production for all users; if that's the case then we might end up adding `source-map-loader` as devDependency in the `extension-cookiecutter-ts`. Let me know what you guys think.

File: builder/package.json [hunk 3]
```diff
@@ -70,6 +70,7 @@
     "@types/node": "^14.6.1",
     "@types/supports-color": "^5.3.0",
     "rimraf": "~3.0.0",
+    "source-map-loader": "~1.0.2",
```
fcollonval, line 73: You need to move this in the `dependencies` section because devDeps of devDeps are not pulled. So this package is a bit exotic in that regard. This is the reason for having `style-loader` or `worker-loader` for example in the dependencies.

fcollonval (approved):
Thanks @3coins

fcollonval:

Issue #13719, opened 2023-01-05
Remove deprecated code


fcollonval:
## Description

Prior to release 4.0.0, we should remove the JupyterLab 3 deprecated code.

Deprecation code to remove:
- // TODO

jtpio:
Related PR: https://github.com/jupyterlab/jupyterlab/pull/13767

PR #11417, opened 2021-11-08
[3.2.x] Expose `window.jupyterapp`

jtpio:
## References

Partial backport of https://github.com/jupyterlab/jupyterlab/pull/10936 to the `3.2.x` branch.

## Code changes

Backport `window.jupyterapp` for consistency with the other lab-based applications.

## User-facing changes

User can then access `window.jupyterapp` from the dev tools console after setting `--expose-app-in-browser`.

## Backwards-incompatible changes

None

PR #13767, opened 2023-01-13
Remove deprecated `window.jupyterlab`

jtpio:
## References

Part of https://github.com/jupyterlab/jupyterlab/issues/13719.

:information_source:  `window.jupyterlab` was marked as deprecated in https://github.com/jupyterlab/jupyterlab/pull/10936  and backported to `3.2.x` in https://github.com/jupyterlab/jupyterlab/pull/11417.

## Code changes

Drop support for the deprecated `window.jupyterlab`.

## User-facing changes

Those using `--expose-app-in-browser` they will need to switch to `window.jupyterapp` (for example in the dev tools console) if not done already.

Drop support for `window.jupyterlab` when using `--expose-app-in-browser`.


jtpio:
Adding the `api-change` label for visibility.

jtpio:
Doing a rebase to grab the CI fixes from https://github.com/jupyterlab/jupyterlab/issues/13772.

fcollonval (approved):
Thanks @jtpio 

---

UI failures are not related:
- Either due to the math font to change last minute and I think this is what shift the snapshot by a pixel
- Or the cells start to be executed although the kernel is not yet ready.

PR #13769, opened 2023-01-13
Merge Component registries

brichet:
Since https://github.com/jupyterlab/jupyterlab/pull/13056, there are 2 registries for RJSF components, one for `Field` and one for `Widget`.

This PR merges them into one registry, using a component-describing interface (see https://github.com/jupyterlab/jupyterlab/issues/12733) to specify the kind of component.

## References

Related to https://github.com/jupyterlab/jupyterlab/issues/12733

## Code changes

- Changes `FormComponentRegistry` to be more generic.
- Removes `FormWidgetRegistry`
- Adapts the following packages to use this new pattern:
   - completer-extension
   - lsp-extension
   - metadataform-extension
   - settingseditor
   - shortcuts-extension

## User-facing changes

There should be no user facing changes

This changes the `FormComponentRegistry` API, some deprecated function should be include to backport it (again see https://github.com/jupyterlab/jupyterlab/issues/12733).

cc @krassowski @fcollonval


packages/metadataform-extension/src/index.ts
```diff
@@ -225,27 +221,20 @@ namespace Private {
...
-
-            // If renderer is defined (custom widget has been registered), set it as used widget.
-            if (formField !== undefined)
-              if (!uiSchema[metadataKey]) uiSchema[metadataKey] = {};
-            uiSchema[metadataKey]['ui:field'] = formField;
+              if (component.fieldRenderer) {
+                uiSchema[metadataKey]['ui:field'] = component.fieldRenderer;
+              } else {
+                uiSchema[metadataKey]['ui:widget'] = component.widgetRenderer;
+              }
```
fcollonval, line 236: Should we support (and does it makes sense) to have simultaneously `ui:field` and `ui:widget` defined?

packages/settingeditor/src/SettingsFormEditor.tsx
```diff
@@ -225,6 +226,11 @@ export class SettingsFormEditor extends React.Component<
     const trans = this.props.translator.load('jupyterlab');
     const icon = this.props.isCollapsed ? caretRightIcon : caretDownIcon;
 
+    const renderers: { [id: string]: Field } = {};
+    for (let id in this.props.components) {
+      renderers[id] = this.props.components[id].fieldRenderer!;
```
fcollonval, line 231: As `this.props.components[id]` may be defined but only with a widget renderer, I would be more caution here:

```suggestion
      const renderer = this.props.components[id].fieldRenderer;
      if (renderer) {
        renderers[id] = renderer;
      }
```

At some point it will be interesting to add support for custom widget in settings editor too. But let's keep that out of this PR.

packages/ui-components-extension/src/index.ts
```diff
@@ -33,8 +31,8 @@ const labiconManager: JupyterFrontEndPlugin<ILabIconManager> = {
 /**
  * Sets up the component registry to be used by the FormEditor component.
  */
-const registryPlugin: JupyterFrontEndPlugin<IFormComponentRegistry> = {
-  id: '@jupyterlab/settingeditor-extension:form-registry',
+const componentRegistryPlugin: JupyterFrontEndPlugin<IFormComponentRegistry> = {
+  id: '@jupyterlab/ui-components-extension:form-component-registry',
```
fcollonval, line 35: :+1:

packages/ui-components/src/FormComponentRegistry.tsx
```diff
@@ -4,74 +4,99 @@
...
-    if (this._renderers[id]) {
+  addComponent(id: string, component: IFormComponent): boolean {
+    if (this._components[id]) {
+      console.warn(`A component with id '${id} is already registered.`);
       return false;
     }
-    this._renderers[id] = renderer;
+    if (!component.fieldRenderer && !component.widgetRenderer) {
+      console.warn(
+        "A component to register must define a 'fieldRenderer' or a 'widgetRenderer'."
```
fcollonval, line 27: Let's provide more context in the error message.

```suggestion
        `The component for '${id}' cannot be registered as it does not define 'fieldRenderer' nor 'widgetRenderer'.`
```

packages/ui-components/src/FormComponentRegistry.tsx
```diff
@@ -4,74 +4,99 @@
...
+   */
+  widgetRenderer?: Widget;
+}
+
+/**
+ * The token for the form component registry.
+ */
 export const IFormComponentRegistry = new Token<IFormComponentRegistry>(
-  '@jupyterlab/ui-components:ISettingEditorRegistry'
+  '@jupyterlab/ui-components:IFormComponentRegistry'
```
fcollonval, line 101: Good catch

:+1:

fcollonval (changes requested):
Thanks @brichet 

Would you mind adding a note in the migration guide about this? In particular could you mentioned the plugin ids and tokens renames.

packages/metadataform-extension/src/index.ts
brichet: I may be wrong but I don't think it make sense.
The `ui:widget` is used to replace the default widget for an input.
If a custom field is defined, there is no default widget anymore, as the field should contain one or more widgets.

docs/source/extension/extension_migration.rst
```diff
@@ -166,12 +166,17 @@ bumped their major version (following semver convention). We want to point out p
...
+   * Major version bumped following removal of Blueprint JS dependency. Extensions using proxied
+     components like ``Checkbox``, ``Select`` or ``Intent`` will need to import them explicitly
+     from Blueprint JS library. Extensions using ``Button``, ``Collapse`` or ``InputGroup`` may
+     need to switch to the Blueprint components as the interfaces of those components in JupyterLab
+     do not match those of Blueprint JS.
+   * Remove ``Collapse`` React component.
+   * Change the ``IFormComponentRegistry`` token, from ``@jupyterlab/ui-components:ISettingEditorRegistry`` to
+     ``@jupyterlab/ui-components:IFormComponentRegistry``.
+   * The ``FormComponentRegistry`` registers ``FormComponent`` instead of field renderers.
+     A ``FoemComponant`` defines a ``fieldRenderer`` or a ``widgetRenderer``.
```
fcollonval, line 178:
```suggestion
     The methods have been renamed from ``...renderer``  to ``...component``. For example, ``addRenderer`` is now ``addComponent``.
     A ``FormComponent`` defines a ``fieldRenderer`` or a ``widgetRenderer``.
```

fcollonval: Minor suggestion to improve the doc. And hopefully once corrected, the CI will be green so we can merge.

fcollonval (approved):
Thanks @brichet

Issue #13659, opened 2022-12-23
Both completer and document search event handlers get dispatched on Esc


krassowski:
## Description

`Cannot execute key binding 'Escape': command 'documentsearch:end' is not enabled.` shows up when closing completer with <kbd>Esc</kbd> in file editor

## Reproduce

1. Open text editor with python file
2. Invoke completer
3. Press Esc
4. See error in browser console

## Expected behavior

- completer widget cancels event if handled
- document search selector is more specific selecting only documents with open search view

## Context

- Browser and version: Chrome 108
- JupyterLab version: 4.0.0a31

TigreModerata:
Hi, I tried to work on this, I have a draft solution - do I just make a push my branch and make a pull request? Do I have to assigned the issue first?

krassowski:
Yes, feel welcome to open a pull request!

krassowski:
One part of this issue (search box-side) was solved by https://github.com/jupyterlab/jupyterlab/pull/14421. We also changed a lot about key bindings (capture phase, etc) since this issue was opened so I will close it for now, as I am not sure if this is still a problem.

Issue #13283, opened 2022-10-22
Search broken in files for specific patterns, causes infinite error loop


krassowski:
## Description

Search broken in files for specific patterns, causes infinite error loop

## Reproduce

1. Create a text file with content `a-b`
2. Open search bar; you should see:
   ![Screenshot from 2022-10-23 00-27-12](https://user-images.githubusercontent.com/5832902/197366037-eba36601-bcaf-40a4-8d13-57140388a456.png)
3. Enter `a-b` to into the searchbar
4. See the search bar gets hidden, file becomes unresponsive, and infinite flood of errors in console
    ![Screenshot from 2022-10-23 00-28-37](https://user-images.githubusercontent.com/5832902/197366066-b4a5dac5-4a0c-43b6-a119-c8d54ccfd43a.png)

Errors:

```js
Failed to parse expression a-b SyntaxError: Invalid regular expression: /a\-b/: Invalid escape (at index.js:154:1)
    at new RegExp (<anonymous>)
    at new RegExpCursor (index.js:154:1)
    at FileEditorSearchProvider._findNext (searchprovider.ts:453:18)
    at FileEditorSearchProvider._startQuery (searchprovider.ts:173:30)
    at async SearchDocumentModel._updateSearch (searchmodel.ts:269:9)
```

```js     
Uncaught Error: Objects are not valid as a React child (found: SyntaxError: Invalid regular expression: /a\-b/: Invalid escape). If you meant to render a collection of children, use an array instead.
    at throwOnInvalidObjectType (react-dom.development.js:13231:1)
    at reconcileChildFibers (react-dom.development.js:14133:1)
    at reconcileChildren (react-dom.development.js:16990:1)
    at updateHostComponent (react-dom.development.js:17632:1)
    at beginWork (react-dom.development.js:19080:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at beginWork$1 (react-dom.development.js:23964:1)
    at performUnitOfWork (react-dom.development.js:22779:1)
````

```js
The above error occurred in the <div> component:

    at div
    at SearchOverlay (http://127.0.0.1:8889/static/lab/packages_documentsearch_lib_index_js.1c5a630e18e6f8aace2f.js:1070:9)
````

## Expected behavior

Search works in files.

## Context

[...]

Issue #13282, opened 2022-10-22
Searching in files steals focus, prevents typing the search phrase (regression)


krassowski:
## Description

Search in files steals focus, prevents typing search phrase

## Reproduce

1. Create a new text file
2. Enter "test"
3. Search "t", wait for the highlight to show up on the matched letter
4. Try to focus the search box again, see focus going immediately going back to the match in the file preventing typing

This does not happen in notebooks.

## Expected behavior

Searching in files does not steal focus

## Context

- Browser and version: Chromium
- JupyterLab version: HEAD/4.0.0a30

krassowski:
Additionally, closing the search box does NOT clear the highlighted matches.

krassowski:
This is even worse, it is impossible to type in the file if you close the search box without clearing the query manually (which is difficult due to focus stealing).

PR #13771, opened 2023-01-14
Add dismiss command for the completer

TigreModerata:
## References

Issue #13659

## Code changes

Added dismiss-widget command to the completer package that gets invoked by Escape in file-editor. 
Before, Escape went to document-search:end without there being a search view opened, so the command was not enabled. Now the dismiss command is enabled with the completer.
Dismiss-widget uses dispose() from the completer widget.

## User-facing changes

console warning gone.


packages/fileeditor-extension/src/commands.ts
```diff
@@ -904,11 +906,28 @@ export namespace Commands {
...
       command: CommandIDs.selectCompleter,
       keys: ['Enter'],
       selector: '.jp-FileEditor .jp-mod-completer-active'
     });
+
+    commands.addKeyBinding({
+      command: CommandIDs.dismissCompleter,
+      keys: ['Escape'],
+      selector: '.jp-FileEditor .jp-mod-completer-active'
+    });
```
krassowski, line 930: We will need to repeat this for the two other cases: console and notebook, see https://github.com/search?q=repo%3Ajupyterlab%2Fjupyterlab%20CommandIDs.selectCompleter&type=code

I think that we do not need to call `commands.addKeyBinding` if we are adding shortcuts using JSON schema (`jupyter.lab.shortcuts`).

krassowski (commented):
Thank you for working on this! I left two suggestions.

packages/fileeditor-extension/src/commands.ts
```diff
@@ -904,11 +906,28 @@ export namespace Commands {
       }
     });
 
+    commands.addCommand(CommandIDs.dismissCompleter, {
+      label: trans.__('Dismisses the completion wideget.'),
```
krassowski, line 910:
```suggestion
      label: trans.__('Dismisses the completion widget.'),
```

packages/completer/src/manager.ts
```diff
@@ -151,6 +151,19 @@ export class CompletionProviderManager implements ICompletionProviderManager {
...
+  /**
+   * Dismiss the completer in the widget with provided id.
+   *
+   * @param id - the id of notebook panel, console panel or code editor.
+   */
+  dismiss(id: string): void {
+    const handler = this._panelHandlers.get(id);
+    if (handler) {
+      // handler.completer.dispose();
+      handler.dispose();
```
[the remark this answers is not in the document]
TigreModerata: All this could probably be avoided if the documentsearch:end command could be inhibited in this context... case 27 should handle all ESC cases on the completer anyway. I'll do better!

TigreModerata:
I added the controller:dismiss-* for all 3 cases, notebook, console and text editor. For the first two I left the selector very specific, jp-mod-completer-active, because I did not want it to pass by the completer if no completer is yet invoked (say you press esc on the console line, I want it to just do what it did before, default of console, not go invoke the completer). For the case of the fileeditor, the default was the problem (going to documentsearch:end) so I created a new Esc binding which does nothing, leaving the completer:dismiss-file to touch only active completer widgets. 
At this point the case 27 (Esc) from the keystroke handler can probably be removed, it does not seem needed anymore, but I wanted to double check with you.

packages/fileeditor-extension/src/commands.ts
```diff
@@ -872,6 +876,16 @@ export namespace Commands {
...
+
+    /**
+     * Override Esc that causes documentsearch:end, doing nothing
+     */
+    commands.addCommand(CommandIDs.overrideEsc, {
+      execute: () => {
+        return;
+      },
+      label: trans.__('Override Esc')
+    });
```
krassowski, line 888: Instead of adding this command we probably should modify how `documentsearch:end` behaves, e.g. by modifying selector.

TigreModerata: I'm not sure we want to change the documentsearch:end... Its selector is jp-mod-searchable, which includes the container of the whole file editor content page (as it should be, I think). 
That is, if I remove the overrideEsc, then pressing esc anywhere else on the sheet (when neither completer or search widget is opened) will go back to the documentsearch:end warning. The warning comes up because indeed no search widget is enabled. One could get rid of the warning  :D 
I can remove the completer-active part of the selector, that way Esc will always pass through the completer:dismiss-file, even when no completer has been invoked (but maintain the end command if a search widget is opened).
Override Esc or always use completer (or the Escape binding from some other package), I can't come up with a third option...

TigreModerata:
I tried changing the selector of documentsearch:end to .jp-DocumentSearch-input, but then my search widget was broken in dev-mode: if any letter typed into search is found in the text, focus changes to the match in the text and one can only overwrite the found result. Cannot get the focus back into the find-view after that and even after closing the widget, matches stay highlighted and focus can't be moved. 
I tried rolling back to my previous version, but I had the same behaviour - I think I just hadn't properly tested the searcher before (because I was too focused on the completer)... this makes me suspect it's not the new selector on the documentsearch extension, but I don't see anything else I could have changed. 
If anyone has an idea of what I might be messing up, any info is greatly appreciated!

krassowski:
> If anyone has an idea of what I might be messing up, any info is greatly appreciated!

Was that in the file editor? There are two bugs in the latest version:
- https://github.com/jupyterlab/jupyterlab/issues/13282
- https://github.com/jupyterlab/jupyterlab/issues/13283

which are mostly affecting the file editor. You can safely ignore the file editor for now.

TigreModerata:
I have nothing to add. Every time i update the branch, it generates new errors, different from the previous ones. I don't know how to fix them. 
My last proposal was the "cleanest" i have (meaning it needs the least touching of any other sections). It's my final attempt 😐

TigreModerata:
I'm sure you all have seen this, but in my last push the dismiss command for the completer had been removed. The default key binding to Escape, that was already in the completer, is used. I'm just mentioning it because of the branch name change which makes sense if you use one of the previous commits - consider these had some not so nice workarounds as mentioned in previous comments.

packages/documentsearch-extension/src/index.ts
```diff
@@ -63,6 +64,13 @@ const labShellWidgetListener: JupyterFrontEndPlugin<void> = {
       }
       if (registry.hasProvider(widget)) {
         widget.addClass(SEARCHABLE_CLASS);
+        // Only bind Esc in case the document is currently searchable,
+        // else it will clash with Esc to close completer on fileeditor
+        app.commands.addKeyBinding({
+          command: CommandIDs.end,
+          keys: ['Escape'],
+          selector: '.jp-mod-searchable'
+        });
```
krassowski, line 73: I don't believe it is doing what you think it does. Once the command is added it will remain there forever. Hence this will only "solve" the problem for the very dismissals of completer which happen before first opening of search panel. Otherwise, it is the same as defining it in schema, but prevents user customisation, so we should not do it.

TigreModerata: I guess one could remove the command each time the search widget is closed, each time the texteditor widget loses the "searchable" class... But I guess it's not coherent with the overall style. 
I'm guessing you can use or not use whatever commit of mine that you find useful, right, or do you want me to do a rollback to the version with the dismiss command for the completer?

krassowski:
Thank you for opening PR and perseverance! The issue it was trying to solve was closed because another PR addressed the problem using a different approach (https://github.com/jupyterlab/jupyterlab/issues/13659#issuecomment-1904364175) and a number of changes on key binding handling were since made.

I need to admit that there was an error of judgement on my side when I labelled the original issue as "good first issue" - it clearly was a difficult one - apologies.

Since the relevant issue was resolved and this PR has conflicts/is not ready I will close it for now , but please feel welcome to open another PR if you would like to contribute in the future!

TigreModerata:
Thanks for the message, I had completely forgotten about this (apologies), as I had probably just given up. Doesn't mean it wasn't a good first issue, just that I had not understood it. Thanks for all your feedback!

Issue #13756, opened 2023-01-11
Ideas on how to improve the find experience


mlucool:
### Problem

The find experience does not feel as intutive as other editors (jetbrains, vscode, etc). Below is a list of ideas to consider to improve this.

### Proposed Solution

Below is somewhat ordered from our perceived most to least value.

1. If the search box is open, the first escape should close that, even when in cell edit mode. Today you need to either click the x or push the hotkey then escape. At that point you are no longer focused in the cell you were trying to edit.
2. Merge cell search + notebook search. It would improve the UX to have a single experience on how to find something. Today there are two ways and it's confusing at times. UX experience suggestion:
a. Whether cell vs. notebook searching is done should depend on a toggle icon inside the search textbox (i.e. move the selected cell option from the hidden `...`).
b. If the toggle is disabled, it should act like the current notebook search.
c. If the toggle is enabled, we're in cell editing mode, and some text is selected, then the scope should be limited to that selected text
d. if the toggle is enabled, we're in cell editing mode, and no text is selected, then the scope should be limited to the current cell
e. if the toggle is enabled, we're in command mode, and some cells are selected, then the scope should be limited to those cells (and not be affected by any selections inside those cells)
f. if the toggle is enabled, we're in command mode, and no cells are selected, then the scope should be limited to the current cell
3.  When find is opened it should either:
a. Start with the currently selected text
b. The last text it was closed with
c. Nothing if it was the first time and no text is selected
4. Add whole word matching
5. Add replace with preserve case

### Additional context

I think VSCode does this very well and it would be a good project to use as a guide.

krassowski:
> 1. If the search box is open, the first escape should close that, even when in cell edit mode. Today you need to either click the x or push the hotkey then escape. At that point you are no longer focused in the cell you were trying to edit.

This was previously reported in https://github.com/jupyterlab/jupyterlab/issues/13311 and my fix from  https://github.com/jupyterlab/jupyterlab/pull/13340 was merged into 4.0 branch.

> 2. Merge cell search + notebook search. It would improve the UX to have a single experience on how to find something. Today there are two ways and it's confusing at times.

+1 I agree with points (a)-(f). Previously it was suggested that we add a shortcut to activate CodeMirror-level fine and replace (https://github.com/jupyterlab/jupyterlab/issues/6650#issuecomment-879829848.), but I agree that it is better to have a consistent UI box for all of them.

[...]

PR #13777, opened 2023-01-15
Add whole-word matching option in search bar

krassowski:
## References

Point (4) in #13756.

## Code changes

- adds `wholeWords` property on search model
- start unit testing search model

## User-facing changes

- Added toggle button in search area to enable whole word matching.
- Increase size of the icons in the buttons. The buttons already were 20px by 20px so previously the click boundaries were not matching the visual indicator boundaries. If we want to reduce the buttons in the future we should modify hte entire button

New required members of the search model


krassowski:
please update snapshots

krassowski:

fcollonval:
Kicking the CI

fcollonval (approved):
Thanks @krassowski 

> I restarted the CI :crossed_fingers:

PR #13778, opened 2023-01-16
Add support for replace preserving case

krassowski:
## References

Point (5) in https://github.com/jupyterlab/jupyterlab/issues/13756.

## Code changes

- add `options?: IReplaceOptions` to `replaceAllMatches()` and `replaceCurrentMatch()` of search provider
   - question for review: should those be required instead?
- add optional `readonly replaceOptionsSupport: IReplaceOptionsSupport` to search providers
   - again, should it be required for discoverability downstream?
- expose the replace support options via model
- add `GenericSearchProvider.preserveCase` utility
- add a button to replace field and remove no longer used `jp-DocumentSearch-replace-entry` class styles

## User-facing changes

New button in the replace field:

Screenshot from 2023-01-17 21-41-24

We probably should use a different icon, but thematically similar.

None.


krassowski:

krassowski:

fcollonval:
Kicking CI

packages/documentsearch/src/tokens.ts
```diff
@@ -263,7 +283,11 @@ export interface IBaseSearchProvider extends IDisposable {
    *
    * @returns A promise that resolves with a boolean indicating whether a replace occurred.
    */
-  replaceCurrentMatch(newText: string, loop?: boolean): Promise<boolean>;
+  replaceCurrentMatch(
+    newText: string,
+    loop?: boolean,
+    options?: IReplaceOptions
```
fcollonval, line 289: Do you think `loop` should be part of `options`?

packages/documentsearch/src/searchview.tsx
```diff
@@ -322,10 +344,18 @@ interface ISearchOverlayProps {
...
+   */
+  preserveCase: boolean;
   /**
    * Whether or not the replace entry row is visible
    */
   replaceEntryVisible: boolean;
+  /**
+   * Support for replace options
+   */
+  replaceOptionsSupport: IReplaceOptionsSupport | undefined;
```
fcollonval, line 358: Should we use the more common pattern:

```suggestion
  replaceOptionsSupport?: IReplaceOptionsSupport;
```

Should we use the smaller name `replaceOptions`?

packages/notebook/src/searchprovider.ts
```diff
@@ -110,13 +112,19 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
...
    * Set to true if the widget under search is read-only, false
-   * if it is editable.  Will be used to determine whether to show
+   * if it is editable. Will be used to determine whether to show
    * the replace option.
    */
   get isReadOnly(): boolean {
     return this.widget?.content.model?.readOnly ?? false;
   }
 
+  get replaceOptionsSupport(): IReplaceOptionsSupport {
```
fcollonval, line 122:
```suggestion
  /**
   * Support for options adjusting replacement behavior.
   */
  get replaceOptionsSupport(): IReplaceOptionsSupport {
```

fcollonval (commented):
Thanks @krassowski 

i left so minor suggestions.

packages/documentsearch/src/tokens.ts
krassowski: I thought about it too. I am not sure. Pros:
- `replaceCurrentMatch` and `replaceAllMatches` would have more similar signatures
- `loop` is an option indeed

Cons:
- `loop` is not used in `replaceAllMatches`
- while `loop` is an option, it is not a user-facing option modifying the provider replace logic, but an a control variable in our internal logic

I imagine that in future `IReplaceOptions` could include things like: `replaceInComments?: bool` (which provider may support or not), `replaceUsagesInOtherFiles?: bool` which could be used by say an `LSPSearchProvider`. `loop` does not fit that well here.

packages/documentsearch/test/documentsearchprovider.spec.ts
```diff
@@ -22,6 +22,23 @@ describe('documentsearch/genericsearchprovider', () => {
...
     });
 
+    describe('#preserveCase()', () => {
+      it.each([
+        ['OLD_TEXT_1', 'new_text_1', 'NEW_TEXT_1'],
+        ['OLD_TEXT_1', 'NEW_TEXT_1', 'NEW_TEXT_1'],
+        ['old_text_1', 'new_text_1', 'new_text_1'],
+        ['old_text_1', 'NEW_TEXT_1', 'new_text_1'],
+        ['Old', 'new', 'New'],
+        ['Old', '𐐶ew', '𐐎ew']
```
krassowski, line 32: Maybe add (and implement) more cases like:

```
['OldText', 'new_text', 'NewText'],
['old_text', 'NewText', 'new_text']
```

Do we want this?

fcollonval: Do you know what VS Code is doing?

It seems pretty complex to handle. I would go for moving forward with this proposal. It can always be improved latter.

packages/documentsearch/src/tokens.ts
fcollonval: Yes good point, let's keep `loop` outside of `options`

packages/documentsearch/test/documentsearchprovider.spec.ts
krassowski: Yes, I can open a followup PR, this way we can merge this one which is rather uncontroversial and then focus on the specific extra cases judging if they are worth extra complexity.

krassowski:

fcollonval:

fcollonval:
Kicking the CI

fcollonval (approved):

PR #13781, opened 2023-01-16
Reduces the context of the form used in metadataform

brichet:
The `formContext` passed when using the `metadataform` contains all the `props` of the parent widget.

Only some of the `props` are necessary, the *settings* and the function to update the metadata.

This PR reduce this `formContext` to strictly necessary attributes.

## Code changes

- the content of the `formContext` of the from used in *FormWidget* of metadataform package.
- the function `updateMetadata()` in *MetadataFormWidget* becomes an arrow function to keep the reference to `this`.

## User-facing changes

None

Changes the path to use the function `updateMetadata()` from a custom fields:
~~props.formContext.metadataFormWidget.updateMetadata()~~ &rarr; props.formContext.updateMetadata()

... which is probably not used yet, as the package is pretty new.

cc @fcollonval (good catch BTW)

### DIFF
--- a/packages/metadataform/src/form.tsx
+++ b/packages/metadataform/src/form.tsx
@@ -37,11 +37,15 @@ export class FormWidget extends ReactWidget {
    * @returns - The rendered form
    */
   render(): JSX.Element {
+    const formContext = {
+      settings: this._props.settings,
+      updateMetadata: this._props.metadataFormWidget.updateMetadata
+    };
     return (
       <Form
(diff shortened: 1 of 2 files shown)


fcollonval:
> ... which is probably not used yet, as the package is pretty new.

And as not been released yet in stable version :wink:

fcollonval:
Kicking the CI

fcollonval (approved):
Thanks @brichet

Issue #13725, opened 2023-01-06
Move docprovider to IDrive


fcollonval:

jtpio:
Maybe we can add a small description to the issue to get a better idea of what this is about?

I guess it's related to https://github.com/jupyterlab/jupyterlab/pull/13512?

hbcarlos:
Fixed in #13786

Issue #13724, opened 2023-01-06
Restore ModelDBFactory in the form of SharedModelFactory


fcollonval:
## Problem

Restore the ModelDBFactory API to be modified for shared models.

This is gonna be useful to create a `Drive` that use Yjs document of documents (to share a tree of files).

hbcarlos:
We need to restore https://github.com/jupyterlab/jupyterlab/pull/13213

hbcarlos:
Fixed in #13786

PR #13512, opened 2022-11-30
Experiment RTC and Drives

hbcarlos:
Coming soon!

PR #13786, opened 2023-01-16
Updates jupyter_ydoc, removes the docprovider and uses drives as providers

### LINKED ISSUE #13723: Update @jupyter/ydoc
### Problem

Update to the latest @jupyter/ydoc version

hbcarlos:
## TODO
- [X] Update `@jupyter/ydoc`
- [X] Restore ModelDBFactory in the form of SharedModelFactory
- [X] Move docprovider to IDrive
- [x] Improve `@jupyter/ydoc` API
- [x] Update migration guide.

## References
Fixes #13723 #13724 #13725

Commit ff35356af4a7cc904521e9353aa119b18ec45197 replaces #13512.

## Code changes
* Updates to `@jupyter/ydoc` v0.3.0 making notebooks' metadata and cell's metadata observables.
* Adds back the ModelDBFactory in the form of SharedModelFactory to instantiate shared models from drives.
* Moves the document provider to drives allowing to extract RTC into an extension in a future PR.


fcollonval (changes requested):
@hbcarlos here are some suggestions

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+        break;
+      case 'notebook':
+        sharedModel = new YNotebook();
+        break;
+      case 'directory':
+      default:
+        return;
+    }
+
+    this._onCreate(options, sharedModel);
```
fcollonval, line 230: Let's prepare the future.

```suggestion
//      default: // FIXME we should request a registry for the proper sharedModel
    }

    if (sharedModel) {
      this._onCreate(options, sharedModel);
    }
```

packages/docregistry/src/context.ts
```diff
@@ -58,20 +53,24 @@ export class Context<
...
+
+    // Get shared model from IDrive.sharedModelFactory
+    const collaborative = PageConfig.getOption('collaborative') === 'true';
+    if (collaborative && this._factory.collaborative) {
+      const sharedFactory = this._manager.contents.getSharedModelFactory(
+        this._path
+      );
+      const sharedModel = sharedFactory?.createNew({
+        path: localPath,
+        format: this._factory.fileFormat!,
```
fcollonval, line 65: Let's acknowledge it can be null

```suggestion
        format: this._factory.fileFormat,
```

fcollonval: It can be `null` to support: https://github.com/jupyterlab/jupyterlab/pull/7596 - this means the API in `@jupyter/ydoc` needs to be updated. Make sure to comment what is the purpose of null.

We could also assume that `null` means collaboration is not possible and not calling `sharedFactory`. But this may hit us latter. So let's open the API and not support it for `YDrive`.

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+    if (
+      typeof options?.format !== 'string' ||
+      typeof options?.contentType !== 'string'
+    ) {
+      throw new Error('Format and content type must be provided.');
+    }
+
+    if (options.collaborative !== true) {
+      return;
+    }
```
fcollonval, line 215: Raising an error is too restrictive - it only means YDrive cannot deal with the document, not that JupyterLab (and the content API) cannot.
Plus `options` is not optional and only format may be non-string.

```suggestion
    if (
      typeof options.format !== 'string'
    ) {
      console.warning(`Only defined format are supported; got ${format}.`);
      return;
    }

    if (!options.collaborative) {
      return;
    }
```

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+
+    let sharedModel: YDocument<DocumentChange> | undefined;
+    switch (options?.contentType) {
```
fcollonval, line 218:
```suggestion
    switch (options.contentType) {
```

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  private _user: User.IManager;
+  private _sharedPaths: Set<string>;
+  private _providers: Map<string, WebSocketProvider>;
+}
+
+/**
+ *
+ */
+export class SharedModelFactory implements IFactory {
```
fcollonval, line 197: I think we don't need to export this one.

```suggestion
class SharedModelFactory implements IFactory {
```

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+  };
+
+  private _user: User.IManager;
+  private _sharedPaths: Set<string>;
+  private _providers: Map<string, WebSocketProvider>;
+}
+
+/**
+ *
```
fcollonval, line 195:
```suggestion
 * Yjs sharedModel factory for real-time collaboration.
```

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+        path: options.path,
+        format: options.format,
+        contentType: options.contentType,
+        model: sharedModel,
+        user: this._user
+      });
+
+      const key = `${options.contentType}:${options.format}:${options.path}`;
+      this._providers.set(key, provider);
+      // FIXME
```
fcollonval, line 172: What is to be fixed?

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,234 @@
...
+   *
+   * @returns A promise which resolves when the file is deleted.
+   *
+   * #### Notes
+   * Uses the [Jupyter Notebook API](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/notebook/master/notebook/services/api/api.yaml#!/contents).
+   */
+  async delete(localPath: string): Promise<void> {
+    await super.delete(localPath);
+    // TODO should we remove the path from the sharedPath?
+    this._sharedPaths.delete(localPath);
```
fcollonval, line 58:
```suggestion
    // We are not removing the path from `sharedPaths` as multiple providers of the same file (with different model) may exist.
    // this._sharedPaths.delete(localPath);
```

There is something wrong with `_sharedPaths` because it does not deal with out of bound rename. We should track this somewhere for latter analysis.

packages/docregistry/src/registry.ts
```diff
@@ -1182,7 +1182,11 @@ export namespace DocumentRegistry {
      *
      * @returns A new document model.
      */
-    createNew(languagePreference?: string, collaborationEnabled?: boolean): T;
+    createNew(
+      languagePreference?: string,
+      sharedModel?: ISharedDocument,
+      collaborationEnabled?: boolean
+    ): T;
```
fcollonval, line 1189: Could you please change the signature to use a object rather than a list of parameters? You may call it `DocumentRegistry.IModelOptions` for example.

It will ease future evolution if needed.

```suggestion
    createNew(options?: {
      languagePreference?: string,
      sharedModel?: ISharedDocument,
      collaborationEnabled?: boolean
    }): T;
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -284,10 +290,14 @@ const factory: JupyterFrontEndPlugin<IFileBrowserFactory> = {
...
 
+    const driveName =
+      PageConfig.getOption('collaborative') === 'true' ? 'YDrive' : '';
+
     // Manually restore and load the default file browser.
     const defaultBrowser = createFileBrowser('filebrowser', {
       auto: false,
-      restore: false
+      restore: false,
+      driveName
```
fcollonval, line 300: This plugin is really annoying.

Would you mind splitting it into two plugins?

One with the current _factory_ that provides `{ createFileBrowser, tracker }`. And another one that depends on the _factory_ to provide the `defaultBrowser`. It will ease replacing it when extracting the RTC logic.

packages/docregistry/src/context.ts
```diff
@@ -203,7 +202,6 @@ export class Context<
     this._isDisposed = true;
     this.sessionContext.dispose();
     this._model.dispose();
-    this._provider.dispose();
     this._model.sharedModel.dispose();
```
fcollonval, line 205:
```suggestion
    // Ensure we dispose the `sharedModel` as it may have been generated in the context
    // through the shared model factory.
    this._model.sharedModel.dispose();
```

fcollonval (changes requested):
Thanks @hbcarlos 

I have a couple of comments.

packages/docprovider/src/ydrive.ts
hbcarlos: Nothing.

packages/docregistry/src/context.ts
hbcarlos: > We could also assume that null means collaboration is not possible and not calling sharedFactory

We should because we need the format for the document room. We use the contentType, format, and path as id to identify the room every client connects for a shared document.

If we don't know the format, we can not connect to the corresponding room.

jtpio:
This looks promising, thanks @hbcarlos for working on this!

packages/docregistry/src/context.ts
fcollonval: In our specific implementation, I agree. But here, I only speak from an API point of view. Some developers may be interested into getting the `null` value because it needs that for providing a special shared model.

hbcarlos:
I need to update the migration guide again. Everything else is ready for review.

packages/docprovider/src/ydrive.ts
```diff
@@ -0,0 +1,238 @@
...
+      // Falling back to the contents API if opening the websocket failed
+      //  This may happen if the shared document is not a YDocument.
+      console.error(
+        `Failed to open websocket connection for ${options.path}.\n:${error}`
+      );
+    }
+  };
+
+  private _user: User.IManager;
+  private _sharedPaths: Set<string>;
```
fcollonval, line 193: After additional thoughts, we should drop this. The reason is that it does not support out of band renames and the call to `save` is unlikely to happen as it is done through `Context.save` that deal with collaboration. So I would be in favor to drop this.

```suggestion
```

fcollonval (changes requested):

fcollonval (commented):
Once @jupyter/ydoc@0.3.0 is released we can merge this one.

fcollonval (approved):
Thanks @hbcarlos - merging so we can move forward

PR #13802, opened 2023-01-17
Fix starting search with selected text (if any)

krassowski:
## References

Point (3) in https://github.com/jupyterlab/jupyterlab/issues/13756

## Code changes

Actually use `model.initialQuery` which was dead code, add tests

## User-facing changes

Selecting single-line text will now set it in the search box.

## X-fail

While it would be expected that the first match gets highlighted, the second gets highlighted and receives focus (destructive!). This is unexpected, but most likely unrelated to the changes in this PR - issues with focus management on 4.0 (after refactor for windowed notebooks) were previously documented:
- https://github.com/jupyterlab/jupyterlab/issues/13282
- https://github.com/jupyterlab/jupyterlab/issues/13283

The issues above are IMO blockers for 4.0.

None

### DIFF
--- a/galata/test/jupyterlab/notebook-search.test.ts
+++ b/galata/test/jupyterlab/notebook-search.test.ts
@@ -35,6 +35,30 @@ test.describe('Notebook Search', () => {
     expect(await nbPanel.screenshot()).toMatchSnapshot('search.png');
   });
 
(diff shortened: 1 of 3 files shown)


fcollonval (approved):
Thanks @krassowski

I'm fine moving forward as is and dealing with focus in another PR.

PR #13804, opened 2023-01-18
Generative AI in Jupyter

3coins:
## Summary 

Generative AI (GAI) models can generate novel content rather than simply analyzing or acting on existing data. The generated data is both similar to the training data and a response to the user provided natural language prompt that describes a task or question. Recent generative AI models such as Amazon CodeWhisperer, Codex, Stable Diffusion, and ChatGPT have demonstrated solid results for many tasks involving natural language (content generation, summarization, question answering), images (generation, explanation, in-painting, out-painting, transformation), data (synthetic data generation, transformation, querying, etc.), and code (autocompletion, explanation, debugging, code reviews, etc.).

Jupyter already has an architecture and user interface for code generation through autocompletion. This architecture has been used to integrate AI provided autocompletions into JupyterLab and the Jupyter Notebook (Kite, Tabnine). As generative AI models expand to perform other tasks, a generalization of the Jupyter autocompletion architecture will empower users to use generative AI to perform any task in Jupyter’s applications.

We are proposing a new GAI architecture which is based on an extensible Jupyter server API for registering GAI models and the tasks they perform in Jupyter. This lets third parties integrate their generative AI models into Jupyter and it lets users enable the models and tasks. New GAI plugins are installed using the JupyterLab extension manager UI or a simple `pip`/`conda install`. Once models have been enabled, users working in JupyterLab or the Jupyter Notebook can 

1. Select anything (text, notebook cell, image, file, etc), 
2. Select or type a prompt to perform a task with the selection, and 
3. Insert the AI generated response in a chosen location (replace, insert below, new file, etc.) 

## Terminology (users & developers)

* **Task**: an object that fully describes a task that the user wishes to perform. It includes a human readable name, a prompt template (what to do), the model engine to use (how to do it), and the insertion mode (where to put it)
    * **Insertion mode**: Field within a task that describes how the generated output is supposed to be inserted back into the document. For example, the generated output could replace the input, or it could be inserted above or below the input.
* **Model engine**: a class deriving from `BaseModelEngine` that executes a model
    * Multiple model engines may share the same underlying **model**; a model engine is best thought of as a way to use a GAI model
* **Prompt**: a string of text, sent to a GAI model to generate content; the final string output resulting from a synthesis of prompt variables and the prompt template
    * The prompt is computed in the backend model engine and it may be shown to the user
* **Prompt variables**: variables sent to the Prompt API from the client
* **Prompt body**: a reserved prompt variable that serves as the main body of the prompt, e.g. the contents of a notebook cell or a text selection
    * analogous to the relation between the body of a HTML document and the document itself
* **Prompt template**: the template that builds a prompt from given prompt variables.
* **Prompt output**: the response that a GAI model sends back after a prompt is run successfully
    * In the first version, we expect that text-based models’ prompt output will be in Markdown format, a mixture of code blocks and markdown-flavored text
    * Some GAI models can generate images, video, and audio content.

## Terminology (contributors)

* **Document widget**: a main area widget that contains an editor widget.
    * Examples: `INotebookTracker#currentWidget`, `IEditorTracker#currentWidget`
* **Editor widget**: the widget within a document widget that’s meant to render the content of a document and potentially serve as an editor as well. 
    * Retrieved by `DocumentWidget#content`

### Prompt API

The Prompt API synthesizes a prompt, executes the model, and returns the output for a given task. This will be the core handler that provides our generated recommendations in notebooks.

Request schema:

```
POST /api/gai/prompt
{
  "task_id": string,      // task ID
  "prompt_variables": {   // object of prompt variables. `body` is reserved and must be specified in all requests
    "body": string,       // prompt body as a string. for imgs, this may be a URL or b64 encoding
    [key: string]: string // extra prompt variables specific to a prompt template
  }
}
```

Response schema:

```
200 OK
{
  "output": string, // model output. can be a URL or b64 encoding for images
  "insertion_mode": string // insertion mode. see "terminology" for more details.
}
```

### Task API

The Task API allows users to create, read, update, and delete tasks.

#### ListTasks

The ListTasks API allows users to list all the tasks for registered model engines.

Request schema:

```
GET /api/gai/tasks
```

Response schema:

```
200 OK
{
  "tasks": [
    {
      "id": string,     // task ID
      "name": string,   // sentence-cased human readable name, e.g. "Explain code"
      "engine": string, // e.g. "gpt3"
    },
    ...
  ]
}
```

#### DescribeTask

The DescribeTask API allows you to retrieve additional metadata about a task, such as the prompt template.

Request schema:

```
GET /api/gai/tasks/{id}
```

Response schema:

```
200 OK
{
  "name": string,            // human readable prompt template name, e.g. "Explain code"
  "engine" string,           // model engine name
  "prompt_template": string, // prompt template content, e.g. "Explain this:\n{body}"
  "insertion_mode": string   // insertion mode. see "terminology" for more details
}
```

#### ListEngines

The ListEngines API lists all model engines registered in the backend, i.e. ones that can be used to generate outputs via the the Prompt API.

Request schema:

```
GET /api/gai/engines
```

Response schema:

```
{
  "engines": [
    {
      name: string       // name of the model, e.g. "gpt2"
      input_type: "txt"  // input type. just "txt" for now
      output_type: "txt" // output type. just "txt" for now
    },
    ...
  ]
}
```

### jupyter_gai server extension

The base `jupyter_gai` extension adds handlers for the HTTP APIs. Exposes a `BaseGaiModelEngine` class for other server extensions to extend.

```python
class BaseGaiModelEngine:
  name: str
  input_type: str
  output_type: str
  
  *def* list_default_tasks(*self*) -> List[_DefaultTaskDefinition_]:
    pass
  
  async def execute(prompt: str):
    pass
```

### Registering GAI model engines

GAI extension uses the entry points under the group name `jupyter_gai.model_engine_class` to discover new model engines. Model engine providers can create a new engine by extending the `BaseModelEngine` class and publishing a python package with this module exposed as an entry point. For example, here is the code and setup for  `TestModelEngine`.

```python
# test_model_engine.py
class TestModelEngine(BaseModelEngine):
    name = "test"
    input_type = "txt"
    output_type = "txt"

    def list_default_tasks(self) -> List[DefaultTaskDefinition]:
        return [{
             "id": "test-task",
             "name": "Test Task",
             "prompt_template": "Test Prompt Template",
             "insertion_mode": "test-insertion-mode"   
        }]

    async def execute(self, task: DescribeTaskResponse, prompt_variables: Dict[str, str]):
        return "test output"
```

Expose this model engine as an entry point. Here, we are assuming that the package is named `jupyter_gai_test`.

```python
# pyproject.toml
[project.entry-points."jupyter_gai.model_engine_class"]
test = "jupyter_gai_test:TestModelEngine"
```

### Registering default tasks

Each model engine exposes a method `list_default_tasks()`, which lists the default tasks that it will install into the Task DB. It has a return type of `List[DefaultTaskDefinition]`, which is defined as such:

```python
class DefaultTaskDefinition(TypedDict):
    id: str # engine-scoped task ID. should be human readable, e.g. "explain-code"
    name: str
    prompt_template: str
```

The engine-scoped task ID is combined with the engine name to form a human-readable, unique task ID for a default task. A default task ID must have a format of `f"{engine_name}:{task_id}"` . This constraint is required for the UI code to have a stable identifier for default tasks.

Note that there is no such constraint on custom task IDs; any globally unique string is sufficient. For our implementation, we will likely just use UUIDv4 by default.

#### Explain or codify cell button in cell toolbar

There is a new button in a cell toolbar that enables users to explain the code in a “code” cell or generates python3 code based on text in a “markdown” cell.

*Explain code in code cell toolbar*
gai-explain-code

*Generate code in markdown cell toolbar*
gai-generate-code

#### Generate output from selection...

“Generate output from selection” mode provides a set of tasks for generating output based on text selection. New tasks can be installed through python entry points (see “Registering default tasks” section).

Explain_from_selection

#### Command driven insertion logic

```typescript
export type InsertionContext = {
  widget: Widget;
  request: GaiService.IPromptRequest;
  response: GaiService.IPromptResponse;
};

/**
 * Function that handles the insertion of Prompt API output into the
 * active document. This function expects that a command with an id
 * of `gai:insert-<insertion-mode>` is registered, context is passed
 * through to the command for handling insertion. See `InsertionContext`
 * for more info. 
 * 
 * @param app - Jupyter front end application
 * @param context - Insertion context
 */
export async function insertOutput(
  app: JupyterFrontEnd,
  context: InsertionContext
): Promise<boolean> {
  app.commands.execute(
    `gai:insert-${context.response.insertion_mode}`,
    context as any
  );
  return true;
}
```

Default commands are registered to allow prompt output to be added above, below or inline replacement of the selected text. These commands work for all document widgets with an editor. Extension writers can enable additional insertion modes by registering a command following the pattern `gai:insert-<insertion-mode>`, and handling the insertion logic within the execute callback. For example, here is the code that handles insertion of prompt output in the default inserter.

```typescript
export function buildDefaultInserter(mode: 'above' | 'below' | 'replace') {
  return function insert(context: InsertionContext): boolean {
    const { widget, request, response } = context;
    const editor = getEditor(widget);

    if (!editor) {
      return false;
    }

    switch (mode) {
      case 'above':
        editor.replaceSelection?.(
          `${response.output}\n\n${request.prompt_variables.body}`
        );
        break;
      case 'below':
        editor.replaceSelection?.(
          `${request.prompt_variables.body}${response.output}`
        );
        break;
      case 'replace':
        editor.replaceSelection?.(response.output);
        break;
    }

    return true;
  };
}
```

This handler is registered as execute function in the default commands in the GAI app.

```typescript
commands.addCommand(CommandIDs.insertAbove, {
    execute: buildDefaultInserter('above') as any
});
commands.addCommand(CommandIDs.insertBelow, {
    execute: buildDefaultInserter('below') as any
});
commands.addCommand(CommandIDs.insertReplace, {
    execute: buildDefaultInserter('replace') as any
});
```

**Note**: This PR is a placeholder to showcase the work for GAI, we don't intend to merge this code into the JupyterLab repo with this PR.


psychemedia:
Generative models are not limited to generating code: they also apply to text (e.g. in the case of ChatGPT) and images. The UI should also support generated text content in markdown cells.

Responses presented by ChatGPT may combine text and code elements. A single prompt may result in a response that should be presented in  a notebook as a series of cells, both markdown cells and code cells.

Generative AI systems may or may not use a "conversation history" in formulating a response. If a request to a service is stateless, it may be appropriate to include previous cells, and previous cell outputs, as part of a prompt. How might that be reflected in terms of UI design?

JasonWeill:
@psychemedia Thanks for your questions! When you run the "generate code" task in a notebook, the code in this PR interprets the output of text-and-code large language models (LLM) like ChatGPT by splitting it into multiple cells: code is placed into a code cell, and the non-code Markdown data before and after code is placed into Markdown cells. We do not yet maintain a conversation history, but LLMs like ChatGPT use session history to provide additional context, so this is something that would make the generative AI project more useful.

dlqqq:
@psychemedia 

(BTW, we propose renaming this project from Jupyter GAI => Jupyter AI, [see here](https://github.com/jupyterlab/team-compass/issues/172#issuecomment-1400896621))
> Generative models are not limited to generating code: they also apply to text (e.g. in the case of ChatGPT) and images. The UI should also support generated text content in markdown cells.

We agree, and this is actually one of many reasons why we built Jupyter AI. Generative models and their capabilities are advancing at an astonishingly fast pace; we believe providing a modular framework for allowing third-party modules to register new models and handle output insertion will allow users to take advantage of new generative models much more quickly.

As mentioned in the body of the issue, modules can register **insertion commands (inserters)** via the command registry to handle insertion of output from a model. These inserters receive a reference to the current editor widget (e.g. an instance of the `Notebook` class), meaning that they are free to do anything their API allows them to do. See [this example](https://github.com/jupyterlab/jupyterlab/blob/6b27dc5e3c373dafe34981ffa5643ffba1a79afc/packages/gai/jupyter-gai-dalle-module/src/index.ts) for more details.

> Responses presented by ChatGPT may combine text and code elements. A single prompt may result in a response that should be presented in a notebook as a series of cells, both markdown cells and code cells.

This is already handled by the `above-in-cells` and `below-in-cells` inserters. 😁 

> Generative AI systems may or may not use a "conversation history" in formulating a response. If a request to a service is stateless, it may be appropriate to include previous cells, and previous cell outputs, as part of a prompt. How might that be reflected in terms of UI design?

This is a very insightful question. We're excited to see the community think deeply about the technical dimensions of building a modular AI framework for JupyterLab.

In short, we are not sure yet. We have a few ideas:

- chat user interface in right sidebar exclusively for stateful models like ChatGPT
- session scoped to the current file; opening a new file starts a new session
- global session shared by all files, scoped to the current user

However, we believe that we need to solicit more feedback from the community before we commit to any UX for a 1.0 release. This project is still very much in 0.x, so it will remain safe to make breaking API changes and drastic UX overhauls. We do not claim to have the complete user story right now; we are merely presenting what we have and letting the community weigh in on whether this is a direction we are excited to develop further.

psychemedia:
Hi, and thanks for the response.

Re: scoping the sessions, would it make sense to think of the GAI session in similar terms to the Jupyter kernel manager controls? For example, in the classic notebook UI:

- `Kernel` menu option offers `Interrupt`, `Restart and Clear Output`, `Restart and Run All`, `Reconnect` and `Shutdown`;
- `Cell` menu option includes `Run Cells`, `Run All`, `Run All Above`, etc.

Being able to, "Restart and run selected cell", "Restart and run all above" and "Restart and run all" are powerful patterns when creating/checking a notebook runs reproducibly.

When using eg ChatGPT, I find myself using the following patterns:

- conversational linear: new session, prompt, response, prompt, response, etc;  *in this case, I expect the history to be part of the session;*
- conversational interrupted: new session, prompt1, response1, prompt2, r̶e̶s̶p̶o̶n̶s̶e̶2̶, edit-prompt2, new-response2, etc; *in this case, I intend to have the prior part of the conversation as part of the session, but be able to dummy run the last prompt/response in the chain;*
- stateless: new session, prompt, response, new session, prompt, response, etc.; *give me a fresh start each time.*

I'm also reminded of various Jupyter and Python extensions that try to keep track of execution dependencies when executing cells, such as [data flow kernel](https://github.com/dataflownb/dfkernel) or partition namespaces as I think [jupyter_spaces])https://github.com/davidesarra/jupyter_spaces) attempts to do?

As it might be appropriate to include markdown cells (and generated markdown), code cells and code outputs in context, I could imagine a UI style where cell tags or pins are used to identify the cells you want to include in in the prompt context (_"Run with pinned/selected above"_).

Noting that folk can get into all sorts of mess reproducibility wise running out of order cells in a traditional notebook-for-code-execution context, I wonder at what point a reproducible prompt and generated response starts to fall apart, eg at as responses are generated, and "pre-written" prompts further down the notebook are invoked? Would being able to have prompts that refer to different contexts be appropriate (eg "using just the response to the previous prompt" could be forced by starting a new session and declaring the response to the previous prompt as the context?) Or would that just get horribly confusing?

3coins:
Closing this, as the vote on this proposal has been finalized in https://github.com/jupyterlab/team-compass/issues/172.

Issue #13563, opened 2022-12-08
3.6 branch CI is failing since jupyter-server 2.0 release


krassowski:
> I don't know what to think about the CI failures. They seems unrelated and they happen also on the latest merge on https://github.com/jupyterlab/jupyterlab/pull/13543. But I'm surprised by their failure on the second attempt.

by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/13555#pullrequestreview-1210406165

> I think this might have something to do with Jupyter Server 2.0 release two days ago. I see that the failures are in server-related tests and that jupyter-server-2.0.0 does get installed, so looks pretty solid.
>
> I re-run Linux (usage, 3.10) with naive hope that 2.0.1 that was released 10 hours ago may have fixed it, but it does not.
>
> Maybe some server 2.0 compatibility thing that was not backported to 3.6.x?

by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/13555#issuecomment-1343461554

One of the offenders appears to be a browser check using notebook server:

https://github.com/jupyterlab/jupyterlab/blob/f209231f9f3790645e5b09f61f4a1fa2adec4fb3/scripts/ci_script.sh#L240-L254

whereas on master it passes because... there isn't one:

https://github.com/jupyterlab/jupyterlab/blob/e658b6f28e3501968e7fe8e211677c5378be14d0/scripts/ci_script.sh#L234-L239

<details>

The relevant log is:

```python
[W 09:34:39.685 NotebookApp] Config option `nbserver_extensions` not recognized by `NotebookApp`.  Did you mean one of: `reraise_server_extension_failures, server_extensions`?
[I 09:34:39.698 NotebookApp] Writing notebook server cookie secret to /Users/runner/Library/Jupyter/runtime/notebook_cookie_secret
[I 2022-12-08 09:34:39.881 LabApp] JupyterLab extension loaded from /Users/runner/work/jupyterlab/jupyterlab/jupyterlab
[I 2022-12-08 09:34:39.881 LabApp] JupyterLab application directory is /Users/runner/hostedtoolcache/Python/3.10.8/x64/share/jupyter/lab
[I 2022-12-08 09:34:39.912 BrowserApp] JupyterLab extension loaded from /Users/runner/work/jupyterlab/jupyterlab/jupyterlab
[...]

PR #13812, opened 2023-01-19
Clean examples

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/issues/13563

## Code changes

Clean example 
- deactivate terminals when not present
- reorganize the `index.html`
- remove custom handler that was unused has the default `LabAppServer` handlers were having higher priority.
- Restrain the page.content output to 1000characters to keep log reasonable
- Use CSS selector to find `jupyter-config-data`.

## User-facing changes

None

None


examples/app/templates/index.html
```diff
@@ -7,18 +7,18 @@
...
+  {% set _ = page_config_full.update(baseUrl=base_url, wsUrl=ws_url) %}
 
-    <script id="jupyter-config-data" type="application/json">
-      {{ page_config_full | tojson }}
-    </script>
-    <script src="{{page_config['fullStaticUrl'] | e}}/bundle.js" main="index"></script>
+  <script id="jupyter-config-data" type="application/json">
+    {{ page_config_full | tojson }}
+  </script>
+  <script src="{{page_config['fullStaticUrl'] | e}}/bundle.js" main="index"></script>
```
jtpio, line 19: Curious about the reason for moving the `<script>` tag in `<head>`.

jtpio: Maybe for consistency with the main JupyterLab template?

fcollonval: Yes this is the reason and also because I added a limit on the page content to be reported in the log (to avoid getting a 100Mo+ log). So I think that config could be of particular interest (especially compared to stylesheets ;) ).

fcollonval:

examples/app/templates/index.html
jtpio: :+1:

PR #13815, opened 2023-01-19
Add languages to the command palette

jtpio:
## References

Sometimes it could be difficult to switch back to the default language as one would have to guess where the entry menu is located.

Having the commands in the command palette makes it easier to search for a given language.

## Code changes

- [x] Make it possible to change the display via the command palette
- [x] Make `IMainMenu` an optional dependency so the translation plugin could be reused more broadly

## User-facing changes

languages-command-palette

None


jtpio:
Adding to the 4.0 milestone to reduce maintenance efforts.

But feel free to backport it to 3.6.x if you think it would be useful to have it there.

fcollonval:
Should we update the text to explain a bit more what this item is about?

Something like _UI Language: <language>_

jtpio:
Or "Display language" as in the documentation: https://jupyterlab.readthedocs.io/en/stable/user/language.html

fcollonval:

jtpio:
Kicking CI.

JasonWeill (approved):

fcollonval (approved):

PR #13823, opened 2023-01-19
Improve form renderer registry

fcollonval:
## References

Extract part of https://github.com/jupyterlab/jupyterlab/pull/13639 to make it lighter and more focus on CodeMirror changes.

## Code changes

- Fix bug in settings plugin enable to use definitions in nested object to build the default values.
- Use `pluginId.propertyName` as key for form renderers
- Rename to use _renderer_ instead of _component_ as it is a more common wording in Lumino/Lab code base.
- Refactor common form components

## User-facing changes

Custom setting renderers will not get mixed up if they have the same property names accross different plugins.

Yes
API, plugin and token renamed
Enforce structure for renderer id


fcollonval:
@brichet I'm not sure I made all needed changes for the metadata form. In particular, the new rules of id to fit `<settings.pluginId>.<propertyName>` may need some changes to fit as there is no `settings.pluginId`. The other change is that I tried to reduce the needed `formContext` to provide `settings` - it is still there because of how the LSP and the completer custom renderers are coded. But I introduced a new `defaultFormData` because this is actually the only piece needed for the custom field template. But I don't see how to get the same information from the metadataform fake settings.

I'm wondering if we should expose `reifyDefault` from `settingregistry.ts` that is used to build default values for settings plugins.

brichet:
> In particular, the new rules of id to fit `<settings.pluginId>.<propertyName>` may need some changes to fit as there is no `settings.pluginId`

This will be a mandatory rule for extensions that will use the metadata form with custom components. These extensions should have a `pluginId` and a `propertyName`, so they can provide the correct pattern I think.
And, if I'm not mistaken, the rule is not too restrictive, I can register `aaa.bbb` whatever my plugin name is. I will just have to use the same name in the settings file to retrieve the renderer.

brichet:
> I'm wondering if we should expose `reifyDefault` from `settingregistry.ts` that is used to build default values for settings plugins.

That would probably be the better solution, and that would force us to use the same interface for schema.
Currently the settings editor uses 
https://github.com/jupyterlab/jupyterlab/blob/a5d8e430dba6465cf990cc7cb679e38442dfbb3b/packages/settingregistry/src/tokens.ts#L349
while the metadata form uses
https://github.com/jupyterlab/jupyterlab/blob/a5d8e430dba6465cf990cc7cb679e38442dfbb3b/packages/settingregistry/src/tokens.ts#L620
but these schema come from the RJSF spec, so they should be 'unifiable' (not sure about that word...).

packages/metadataform/src/form.tsx
```diff
@@ -38,27 +33,27 @@ export class FormWidget extends ReactWidget {
    */
   render(): JSX.Element {
     const formContext = {
+      defaultFormData: this._props.settings.default(),
       settings: this._props.settings,
       updateMetadata: this._props.metadataFormWidget.updateMetadata
```
brichet, line 38:
```suggestion
      defaultFormData: this._props.settings.default(),
      updateMetadata: this._props.metadataFormWidget.updateMetadata
```
Probably the *settings* can be removed from the `formContext` now there is the `defaultFormData`

packages/metadataform/src/metadataform.ts
```diff
@@ -416,12 +419,15 @@ export class MetadataFormWidget
...
+      if (hasValue) formData[metadataKey] = workingObject as JSONValue;
     }
 
     this.buildWidget({
       properties: formProperties,
-      settings: new Private.Settings(this._metaInformation),
+      settings: new Private.Settings({
+        metaInformation: this._metaInformation,
+        schema: this._metadataSchema
+      }),
```
brichet, line 430:
```suggestion
      settings: new BaseSettings({
        schema: this._metadataSchema as PartialJSONObject
      }),
```
Perhaps we can do without the `Settings` class, it seems to be used only to provide the functionalities of `BaseSettings`.
If so, the interface `ISettings` should be removed too.

brichet:
For consistency, maybe we could also change the attribute `customComponent` to `customRenderer` in 
- packages/metadataform-extension/schema/metadataforms.json
- settingregistry/src/plugin-schema.json
- packages/metadataform-extension/src/index.ts

brichet (approved):

fcollonval:
Thanks for the review @brichet

PR #13828, opened 2023-01-21
Bump lumino packages to `2.0.0-beta.0`

krassowski:
## References

Needed for https://github.com/jupyterlab/jupyterlab/pull/13652 (and 4.0 release of course).

Corresponding lumino calver tag/changelog: [v2023.1.13-beta.0](https://github.com/jupyterlab/lumino/releases/tag/v2023.1.13-beta.0).

## Code changes

- Bumps `2.0.0-alpha.6` → `2.0.0-beta.0`
- Re-aligns `@lumino/datagrid` version to `2.0.0-beta.0`
  Initially we though we will release next `@lumino/datagrid` as 1.0, but it was changed in https://github.com/jupyterlab/lumino/pull/458 without reflecting the changes on the lab side.
- Update examples to include `Signal.block()` shims

## Blockers:
- [x] https://github.com/jupyterlab/jupyterlab/pull/13831
- [x] disable `contain: strict` on `.jp-NotebookPanel-toolbar` (via `!important`) so that the kernel tooltip shows up again (https://github.com/jupyterlab/jupyterlab/pull/13828/commits/8520e395742a6ead070fe643f19800bfd7356ac7)


krassowski:
please update snapshots

krassowski:
The visual changes are a result of https://github.com/jupyterlab/lumino/pull/506. We need to adjust CSS styles for sidebar, as we did for the tabbar (https://github.com/jupyterlab/jupyterlab/pull/13671). I will include the changes in this PR if that's ok.

fcollonval:
bot please update snapshots

fcollonval:
@krassowski I can use your expertise for the notification UI test failures.

The notification center is not displayed and get hidden using `z-index=-1000` and `visibility=hidden` from the `HoverBox` logic. In Firefox 98, this is not happening (but locally using Chromium 108 I can reproduce the error).

I discovered that the body has `height: 0px` (both in FF and Chromium):

image

But it was already like that in 4.0.0a32 e.g.. Would you have some ideas how to fix this?

---

Another source of multiple errors is the menus:

```
waiting for locator('ul[role="menu"]').locator('text=New')
      locator resolved to 18 elements. Proceeding with the first one.
      locator resolved to hidden <div class="lm-CommandPalette-itemLabel">New Console</div>
```

Hiding the command palette items should certainly be dropped - building it when needed should be better especially for performance.

krassowski:
I can take a look tomorrow, but it sounds like CSS strict containment needs to be relaxed for notification centre. For menu tests, I guess the tests need updating selectors.

fcollonval:
The menu errors are the consequence of that change in Lumino 2: https://github.com/jupyterlab/lumino/commit/ca46a0330c16c60261b5db2bb3aa130745d42d8c

This adds `role="menu"` to the command palette. So as suggested by @krassowski, the way to fix it is to update the selectors.

---

> I can take a look tomorrow, but it sounds like CSS strict containment needs to be relaxed for notification centre.

Thanks

fcollonval:
There is a regression in the menu bar style similar to the sidebar panel:

image

fcollonval:
That changes is gonna be more tricky to fix that I fought. It is the consequence of https://github.com/jupyterlab/lumino/commit/55e87caa277ac3ae3d15b8c5da75e78d54b069ff

As the menu is now placed using CSS transformation, the following trick does not work:

https://github.com/jupyterlab/jupyterlab/blob/7227449db8db478dc5e31bbc1ffdfb741766a35d/packages/application/style/menus.css#L30

image

fcollonval:
The galata snapshots highlight some new bugs related to Markdown cells rendering.

krassowski:
bot please update snapshots

fcollonval:
Thanks a lot @krassowski for working on the fixes

krassowski:
> The galata snapshots highlight some new bugs related to Markdown cells rendering.

I did not get to this one. It appears to be a thing with not un-rendering when switching heading level. I did not reproduce it locally so I wonder if this is a tests thing (taking snapshot too early) but did not have time to investigate. Feel free to pickup if you have time in the meantime.

krassowski:
> The galata snapshots highlight some new bugs related to Markdown cells rendering.

So this is not a flake nor tests issue after all, but a regression from lumino (I think from beta7). It is pretty severe as the cell that gets created (just press Shift + A in command mode to reproduce) cannot be entered.

I pushed a workaround (866be56) which could allow us to get this merged, but this will need more investigation.

krassowski:
bot please update snapshots

fcollonval:
Thanks @krassowski for all the fixes. I analyzed more deeply the error with the markdown cell and I think I get to the bottom of the code to blame in 7079293.

krassowski:
Thank you! We still see failures in `js-cells` and `js-filebrowser`; do you think that these are possibly related to the signal rework in lumino?

fcollonval:
> Thank you! We still see failures in `js-cells` and `js-filebrowser`; do you think that these are possibly related to the signal rework in lumino?

I'll have a look tomorrow

fcollonval (approved):
We are good to go

fcollonval:

Issue #12115, opened 2022-02-24
Document how to install and run notebook (v6) aside JupyterLab (v4)


echarles:
### Problem

https://github.com/jupyterlab/jupyterlab/pull/11894 replaces nbclassic with notebook_shim. The consequence is that the user can not go anymore to the `/tree` url to get the notebook v6 if he wants to.

### Proposed Solution

We need to document and communicate on this

### Additional context

This discussion in https://github.com/jupyterlab/jupyterlab/pull/11894

jtpio:
@echarles if I recall correctly the `nbclassic` docs already cover some of these scenarios? should we link to them in from the JupyterLab documentation?

echarles:
I have quickly checked on https://nbclassic.readthedocs.io/en/latest that information but don't really find it. The changes brought to the nbclassic documentation have been mainly done by @Rrosio and mostly tackle name alignement https://github.com/jupyter/nbclassic/commits/main/docs.

@Rrosio can you point useful information regarding the simultaneous run of nbclassic and jupyterlab?

RRosio:
As @echarles mentions the documentation was last updated some time ago and focuses on the name alignment. The section where this documentation can go is probably, [NbClassic in the Jupyter Ecosystem ](https://nbclassic.readthedocs.io/en/latest/nbclassic.html#nbclassic-in-the-jupyter-ecosystem). I can make a PR updating this documentation today if that sounds okay?

RRosio:
I have opened the [PR #196](https://github.com/jupyter/nbclassic/pull/196) for the nbclassic documentation. I would appreciate any feedback you could provide!

echarles:
https://github.com/jupyter/nbclassic/pull/196 is merged. Closing this issue.

jtpio:
Nice thanks!

Wondering if there should also be a small section in the JupyterLab docs that points to this part of the nbclassic docs?

Another option would be to edit the changelog page and add a small paragraph there saying that Notebook is not installed by default with JupyterLab 4 anymore: https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html

Similar to the top of the `v3.0` changelog with the "User Facing" and "For Developers" sections.

echarles:
Reopening as more work on lab docs is needeD.

RRosio:
[...]

PR #13830, opened 2023-01-21
link nbclassic docs, note Lab 4 drop notebook dependency

RRosio:
## References
Updating docs to as mentioned in https://github.com/jupyterlab/jupyterlab/issues/12115

## Code changes

Adds a few sentences in documentation offering links to NbClassic documentation which informs users about using JupyterLab, Notebook and NbClassic. Also, mentions the drop in Notebook dependency as of Lab 4.


jtpio:
Thanks!

docs/source/getting_started/overview.rst
```diff
@@ -85,4 +85,10 @@ We appreciate feedback on our `GitHub issues page <https://github.com/jupyterlab
 as we evolve towards a stable extension development API.
 
 JupyterLab will eventually replace the classic Jupyter Notebook.
-Throughout this transition, the same notebook document format will be supported by both the classic Notebook and JupyterLab.
+Throughout this transition, the same notebook document format will
```
jtpio, line 88: > JupyterLab will eventually replace the classic Jupyter Notebook.

Wondering if the sentence above is still valid? With the current state of the projects Notebook 7 will be the one replacing the classic notebook 6.

jtpio: But it's true we should indeed still mention JupyterLab 4 will not have the `notebook` dependency.

RRosio: Right, I think that sentence could be changed to better reflect the state of projects today. I have made some updates which I thought might fit better, but I would appreciate feedback on them!

jtpio (approved):
Thanks!

PR #13831, opened 2023-01-21
Rework sidebar styles to allow adoption of css `contain:strict`

krassowski:
## References

In lumino we enabled strict CSS containment for widgets by default (https://github.com/jupyterlab/lumino/pull/506); while it can be disabled (overridden with `!important`) downstream, it should not be usually needed.

In tab bar we used to implement the visual effect of no border between the active item by shifting it beyond the parent element boundary with compositor transforms. This was reworked in https://github.com/jupyterlab/jupyterlab/pull/13671 for the tab tabbar, but sidebar also uses a tabbar with custom styling, so this PR does the same for sidebar.

It is blocking update to latest lumino in #13828.

Just to highlight, this is to prevent this change seen in #13828:

image 

## Code changes

- simplify styling of sidebar:
   - we used to rotate the sidebar (clockwise or anti-clockwise depending whether left sidebar or right sidebar) in compositor and then rotate icons back; this was implemented early on when there were text labels rather than icons and when browsers had no native means of rotating text outside of compositor (=preserving dimensions after rotation); nowadays we can use `writing-mode` which is well supported by browsers which enables us to remove a lot of old, complex code
- `jp-Icon-20` is no longer added for tab bar items with no icon nor label; instead a dedicated placeholder icon is shown; this is
- `jp-Icon-20` is no longer added for tab bar items with label but no icon (which was shifting the text by size of icon which was no there)
- add visual regression test for the text labels

## User-facing changes

- None in core.
- No extra space wasted when displaying a tab with label but without icon in sidecar and other downstream extensions.

None


krassowski:

krassowski:

krassowski:

krassowski:

fcollonval (approved):
Thanks a lot for this one (and the tests :heart:) @krassowski

PR #13832, opened 2023-01-21
Reset execution indicator state when kernel restarts

### LINKED ISSUE #13780: Incorrect kernel status after it dies unexpectedly and is restarted
## Description
If a kernel dies unexpectedly during code execution and is restarted by the user. It will show the kernel status as busy always.

A small snippet to reproduce this - 
```python
import os
os.system(f'kill {os.getpid()}')
```

## Reproduce

1. Open JupyterLab
2. Type the above code snippet in a cell
3. Run the cell 
4. Restart kernel and try running some other code.
5. Kernel status will always show busy now.

## Expected behavior

Kernel status should correspond to the actual status of the kernel.

krassowski:
## References

Fixes #13780.

## Code changes

Add special handling for `restarting` message, which is the only message we get if the kernel is killed non-gracefully. Because abrupt killing of kernel prevents it from responding with idle status we need to listen to `restarting` instead (and clear the set with scheduled cells).

## User-facing changes

Execution indicator works after abrupt kernel death.

None

### DIFF
--- a/packages/notebook/test/executionindicator.spec.tsx
+++ b/packages/notebook/test/executionindicator.spec.tsx
@@ -50,6 +50,8 @@ afterAll(async () => {
   await server.shutdown();
 });
 
+const SESSION_SETUP_TIMEOUT = 30000;
+
 describe('@jupyterlab/notebook', () => {
   let rendermime: IRenderMimeRegistry;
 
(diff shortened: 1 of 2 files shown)


packages/notebook/test/executionindicator.spec.tsx
```diff
@@ -157,6 +165,32 @@ describe('@jupyterlab/notebook', () => {
         await NotebookActions.run(widget, ipySessionContext);
         expect(executed).toEqual(expect.arrayContaining([3, 3, 3, 2, 2, 2, 0]));
       });
+
+      it('should reset to idle when kernel gets abruptly interrupted', async () => {
```
krassowski, line 169:
```suggestion
      it('should reset to idle when kernel gets abruptly killed', async () => {
```

krassowski:
The added test is a bit blunt and while it works perfectly locally it does not pass on CI.

I did not find another way to reproduce the issue with a two-liner, but suggestions for an alternative testing strategy (or how to make the tests pass) are very much welcome!

krassowski:
I was able to reproduce the CI error locally; it was because previously I was only running the added test in isolation; as it was killing kernel we had to restart it. This was fixed in 40acbf271b03f9641b7a79f849a1caf1d8024d03.

For record the traceback when tests were failing was:

<details>

```
  File "tornado/web.py", line 1713, in _execute
    result = await result
  File "jupyter_server/services/sessions/handlers.py", line 190, in delete
    await sm.delete_session(session_id)
  File "jupyter_server/services/sessions/sessionmanager.py", line 523, in delete_session
    await ensure_async(self.kernel_manager.shutdown_kernel(session["kernel"]["id"]))
  File "jupyter_core/utils/__init__.py", line 172, in ensure_async
    result = await obj
  File "jupyter_server/services/kernels/kernelmanager.py", line 426, in _async_shutdown_kernel
    return await self.pinned_superclass._async_shutdown_kernel(
  File "jupyter_client/multikernelmanager.py", line 264, in _async_shutdown_kernel
    if not km.ready.cancelled() and km.ready.exception():
asyncio.exceptions.InvalidStateError: Exception is not set.
````

</details>

trungleduc (approved):

trungleduc:
Thanks @krassowski!

fcollonval (approved):

fcollonval:

PR #13834, opened 2023-01-22
Improves translator API

### LINKED ISSUE #12684: More semantic API for the ITranslator
### Problem
The public API of the ITranslator only has one method, `load`.

To know what language JupyterLab uses, we need to access the setting's plugin and load the translator's settings ('@jupyterlab/translation-extension:plugin'). Probably we should have a more semantic API for this plugin so we can request what language is loaded. Some extensions may use libraries with support for multiple languages, and they may need to listen for changes in JupyterLab to load the corresponding language.

### Proposed Solution
Expose a more semantic API. Maybe a signal `languageChanged` is tricky because after changing the language, Jupyterlab is reloaded, but at least a property that indicates what language is being used.

### Additional context

cc @fcollonval

hbcarlos:
## References
Fixes #12684

## Code changes
Makes the property `languageCode` public.

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/translation/src/base.ts
+++ b/packages/translation/src/base.ts
@@ -13,6 +13,8 @@ class NullTranslator implements ITranslator {
     this._languageBundle = bundle;
   }
 
+  readonly languageCode: string = 'en';
+
   load(domain: string): TranslationBundle {
     return this._languageBundle;
   }
--- a/packages/translation/src/base.ts
+++ b/packages/translation/src/base.ts
@@ -19,10 +19,6 @@ class NullTranslator implements ITranslator {
     return this._languageBundle;
   }
 
-  locale(): string {
-    return 'en';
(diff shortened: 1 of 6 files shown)


packages/translation/src/base.ts
```diff
@@ -13,6 +13,8 @@ class NullTranslator implements ITranslator {
     this._languageBundle = bundle;
   }
 
+  readonly languageCode: string = 'en';
```
krassowski, line 16: Should we remove the existing `locale()` below to avoid confusion?

hbcarlos: Yes, that would be better.

krassowski (approved):
Thank you

fcollonval (approved):

PR #13792, opened 2023-01-16
Updates to the kernel panel of the "running" sidebar

afshin:
This PR changes the behavior and user experience of the running kernels sidebar panel. The UX needs a further review and this PR will need further iterations before we arrive at a happy state.

- Instead of showing "sessions" from `api/sessions` (each "session" might be better described as "kernel connection", i.e., the notebook or console that is connected to a live kernel), the panel shows all of the active **kernels**.
- When a running kernel is clicked on, it opens all of the "sessions" connected to it, e.g., if a kernel has a console and a notebook session associated with it, both are opened when the user clicks on it.
- This fixes a bug where multiple connections (i.e., multiple "sessions" in the API's usage of the term) are listed as individual "kernels" incorrectly. With this PR, only one kernel is listed, with this panel being an accurate reflection of a request to `api/kernels` instead of `api/sessions`
- The `KernelManager` instance that is used internally by the `ServiceManager` to underlie the `SessionManager` has been exposed as `ServiceManager#kernels`
- Fixes a bug in the `KernelManager` that caused it to ignore changed kernels when they were refreshed.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/13779

## User-facing changes
The running kernels section has changed.

#### Before
before

#### After
after

Issue #13779, opened 2023-01-16
JupyterLab Running Panel: display kernels only once per instance - and show list of connected sessions


SylvainCorlay:
The "kernels" section of the "Running" left panel currently shows inconsistent data:

In the following screenshot, we see the state after opening a notebook and then connecting a console to it, which seems correct (only one kernel, two tabs):

<img width="252" alt="Screenshot 2023-01-16 at 09 46 07" src="https://user-images.githubusercontent.com/2397974/212635711-ff269217-cb25-4954-923d-027180fc2843.png">

In the following screenshot, we see the state after opening a console and then connecting a notebook to it, which seems incorrect (same kernel listed twice - killing one kills the other one):

<img width="248" alt="Screenshot 2023-01-16 at 09 48 49" src="https://user-images.githubusercontent.com/2397974/212635929-5478136b-ef31-454e-921a-1e75831cc351.png">

The proposed fix is to list all kernels by name in the kernels pane, and list the sessions:

<img width="248" alt="fixed" src="https://user-images.githubusercontent.com/2397974/212640605-a86c875d-a36c-4a32-ae1a-2e34100ae312.png">

As a second iteration, we could make that list of connected tabs expendable below the kernel name, but getting to that state would already be a proper bug fix.

krassowski:
Could we iterate on the UX introduced in #13792; I find it to be a downgrade in the UX for the most common action the user takes (of closing the kernel they would like to close):

![Screenshot from 2023-01-24 10-35-45](https://user-images.githubusercontent.com/5832902/214270009-b9f843cb-efe7-45f3-9441-a28c88e5e80b.png)

1. In worst case scenario it requires user to hover over every kernel, say 20 Python 3 kernels for 20 open noteboks, to find the kernel they want to close
2. It uses Jupyter icon rather than Python icon (as proposed in the this issue). In fact I think it should use a stack of Notebook/Console icons (if the kernel is used by Notebook, only use Notebook, if it is used by Notebook and Console show both icons overlaid with small offset)
3. Minor: curly braces do not have an obvious meaning to me. Maybe we could overlay the number over the icon in a corner?
4. Minor: I don't think we need to show the number if it is equal to 1.

Thoughts @SylvainCorlay?

afshin:
@krassowski 
I have been working on a follow-on PR to improve this UI based on #13792, because I agree that it isn't a good enough user experience.

[...]

PR #13851, opened 2023-01-24
Improve kernels pane in running sidebar

afshin:
This PR:
- Updates the `notebook:create-new` command to support passing in an optional `kernelId` to create a new notebook that uses an existing kernel instance.
- Adds a context menu to the kernels pane in the running sidebar that allows users to:
  - Create a new code console for the given kernel
  - Create a new notebook for the given kernel
  - Switch to an existing session connection for the given kernel
  - Shut down the given kernel
-  Adds a collapsible list of connected sessions beneath each kernel in the running kernels pane
- Displays the icon in the kernel spec's `resources` instead of displaying the Jupyter icon when possible

## References
- #13779
- #13792 

## User-facing changes

https://user-images.githubusercontent.com/159529/214585918-6dd246d2-7eb6-4fc4-bc24-2c73b01b462e.mov

N/A


afshin:

afshin:

afshin:
bot please update snapshots

afshin:
bot please update snapshots

fcollonval (changes requested):
thanks @afshin 

You gonna have to update the selector in the UI documentation tests to make them succeed before asking for an update of the snapshots.

afshin:
@fcollonval, thanks for the review!

The only place I'm finding a selector in the docs is for `#jp-running-sessions`, which still exists. I think I am not looking in the right place 🫤

jtpio:
Thanks @afshin for working on this!

I just tried the latest changes on this branch it's looking good.

Quick question: should a new console started with "New Console for Notebook" also be listed under the kernel? Similar to when a separate console is started and then its kernel changed to an existing one:

https://user-images.githubusercontent.com/591645/215058318-bb272a14-68a8-40dd-b685-79548c3f1f4e.mp4

jtpio:
> I think it should appear in the list as well.

Same here. In the example above I would expect the new console started from the `Untitled.ipynb` notebook to also be listed under the kernel, with `Untitled.ipynb` as the name and with the console icon.

fcollonval:
> The only place I'm finding a selector in the docs is for `#jp-running-sessions`, which still exists. I think I am not looking in the right place 🫤

To fix some tests with the first PR, I changed some selectors to use the text content (it is considered better for integration tests). There are two places:

https://github.com/jupyterlab/jupyterlab/blob/0250b696b12c1421931cbf6588253d8a2de5bde0/galata/test/documentation/general.test.ts#L144

https://github.com/jupyterlab/jupyterlab/blob/0250b696b12c1421931cbf6588253d8a2de5bde0/galata/test/documentation/general.test.ts#L465

As now the text is `Python 3 (ipykernel)`, it is similar to the text use in the kernel status of the notebook. So the selector should be something like:

```ts
page.locator('#jp-running-sessions >> text="Python 3 (ipykernel)"')
```

> If you want to read more about the selection in playwright, the doc is over [there](https://playwright.dev/docs/other-locators)

afshin:
bot please update snapshots

fcollonval:

fcollonval (approved):
Thanks @afshin

jtpio:
Looks good, thanks!

PR #13866, opened 2023-01-26
Use local paths instead of driveName:path in the shared model

hbcarlos:
JupyterLab prefixes every path with the name of the drive (like `driveName:foo/bar.txt`). When we make a call to the contents manager, this one is responsible for removing the name of the drive from the path, making the request through the corresponding drive, and adding the name of the drive back to each path present in the response.

To be able to rename files in a collaborative session, we added the path of the document to the shared model. At the moment, this path is used by the backend (jupyter_server_ydoc) to track whether the file was renamed or not.

Since the path is used by the backend and it doesn't know about drives, the context has the same responsibility as the contents manager (removing/adding the name of the drive into the path every time it writes/reads the path from the shared model). 

## Code changes
Transforms the path every time the context writes/reads from the shared model.

## User-facing changes
Fixes a bug when restoring a document in collaborative mode

N/A

cc @fcollonval


packages/docregistry/src/context.ts
```diff
@@ -91,7 +91,6 @@ export class Context<
       path: this._path,
       contents: manager.contents
     });
-    this.model.sharedModel.setState('path', this._path);
```
fcollonval, line 94: Where is the state initialized? In the backend?

hbcarlos: Yes, in the backend. And the state is only used in collaborative mode.

fcollonval (approved):
Thanks @hbcarlos

hbcarlos:

PR #13859, opened 2023-01-26
upgrade to MathJax 3

dlqqq:
## References

Closes #7218

## Code changes

- Bumps MathJax version to 3
- Ports existing labextension from [`jupyter-renderers/mathjax3-extension`](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/mathjax3-extension)
- Renames `packages/mathjax2{,-extension}` => `packages/mathjax{,-extension}`.

## User-facing changes

Screen Shot 2023-01-25 at 4 21 53 PM

- Some scarcely noticible font CSS differences inherited from `mathjax3-extension`
- Right-clicking on MathJax-rendered output no longer opens the [MathJax context menu](https://docs.mathjax.org/en/latest/options/menu.html) (see below for screenshot)
  - This is likely an unintentional regression, as the MathJax context menu is meant to be enabled by default. However, this may also be a regression we wish to keep in favor of maintaining UX consistency.
  - MathJax options can be added into the JupyterLab context menu. PR: https://github.com/jupyterlab/jupyter-renderers/pull/282
  - If we do decide to integrate MathJax options into JL context menu, I suggest we do so in a follow-up PR rather than blocking this one.

Screen Shot 2023-01-25 at 4 28 28 PM

## Backwards-incompatible changes

None known.

PR #13877, opened 2023-01-28
upgrade to MathJax 3 [with git history]

### LINKED ISSUE #7218: Update to MathJax 3
MathJax 3 was released a few days ago. Perhaps we should upgrade to it in 2.0.

PR for our external plugin is at https://github.com/jupyterlab/jupyter-renderers/pull/213

dlqqq:
## References

- Closes #7218
- Re-write of #13859

## Code changes

- Bumps MathJax version to 3
- Ports existing labextension from [`jupyter-renderers/mathjax3-extension`](https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/mathjax3-extension) **while preserving git history**.
- Removes `packages/mathjax2{,-extension}` and adds `packages/mathjax-extension`.

## User-facing changes

Screen Shot 2023-01-25 at 4 21 53 PM

- Some scarcely noticible font CSS differences inherited from `mathjax3-extension`
- Right-clicking on MathJax-rendered output no longer opens the [MathJax context menu](https://docs.mathjax.org/en/latest/options/menu.html) (see below for screenshot)
  - MathJax options in JL context menu added here: https://github.com/jupyterlab/jupyter-renderers/pull/282

Screen Shot 2023-01-25 at 4 28 28 PM

## Bugs

- **JL4 regression**: MathJax context menu options (e.g. "Scale 1.5x") no longer work. Recommend this gets addressed in a follow-up PR.

None known.


dlqqq:
@jtpio I will need your help to release `@jupyterlab/mathjax-extension 4.0.0-alpha.18` to hopefully pass CI. Could you also look into whether it's possible to unpublish `@jupyterlab/mathjax`, as that no longer exists? I followed Frederic's advice of just using one package, as this package doesn't export any tokens.  Thanks in advance! 😁

dlqqq:
FYI, I wrote a guide on how to migrate existing packages into JupyterLab. Is there anywhere to publish this so future developers won't have to retrace my steps? See below:

<details>
<summary>Migrating an existing package into JupyterLab</summary>

## Migrating an existing package into JupyterLab

Sometimes it is necessary to migrate an existing package into the JupyterLab repository. Doing so in a way that preserves git history credits our contributors and builds trust with our community.

Let us call the existing package the *source package*, and call its parent repository the *source repository*. Migration involves two phases: isolation and integration.


### Package isolation

This sequence of steps isolates the git history for a source package within a repository, and sets the repository root to the package root.  You can skip this if your source package lives at the repository root. 

1. Install `git-filter-repo`: 

```
mkdir -p ~/.local/bin && cd ~/.local/bin
wget https://raw.githubusercontent.com/newren/git-filter-repo/main/git-filter-repo
chmod +x git-filter-repo
```

Then add `~/.local/bin` to your $PATH and restart your shell session.

1. Clone the source repository.
2. With your working directory set to the source repository root, isolate the source package:

```
`git-filter-repo --subdirectory-filter <source-pkg-path>`
```

For example, if I wish to isolate the source package under `packages/mathjax3-extension`:

```
git-filter-repo --subdirectory-filter packages/mathjax3-extension
```

This copies the contents of `packages/mathjax3-extension` to the repository root while deleting all other files and their git history. In other words, it’s as if `packages/mathjax3-extension` was “always its own repository”. Bear in mind the above command doesn’t rename the repository root, so you will have to explicitly rename it.

### Package integration

1. Install git-subtree if it isn’t already installed by your package manager. There are several ways to install it, but the easiest way is to just add the script to your $PATH.

```
mkdir -p ~/.local/bin && cd ~/.local/bin
wget https://raw.githubusercontent.com/git/git/master/contrib/subtree/git-subtree.sh
mv git-subtree.sh git-subtree
chmod +x git-subtree
```

Then add `~/.local/bin` to your $PATH and restart your shell session.

1. Change your working directory to the JupyterLab repository root, and add the package as a git subtree:

```
git subtree add --prefix packages/<source-repo-name> <source-repo-path> <source-repo-branch>
```

For example, to migrate an existing package at path `../mathjax3-extension` with a main branch named `main` to a new package named `mathjax-extension`:

```
git subtree add --prefix packages/mathjax-extension ../mathjax3-extension main
```



</details>

packages/mathjax-extension/package.json
```diff
@@ -24,30 +32,32 @@
...
     "@jupyterlab/application": "^4.0.0-alpha.18",
-    "@jupyterlab/coreutils": "^6.0.0-alpha.18",
-    "@jupyterlab/mathjax2": "^4.0.0-alpha.18",
-    "@jupyterlab/rendermime": "^4.0.0-alpha.18"
+    "@jupyterlab/rendermime": "^4.0.0-alpha.18",
+    "mathjax-full": "^3.2.2"
   },
   "devDependencies": {
+    "@jupyterlab/builder": "^4.0.0-alpha.18",
+    "npm-run-all": "^4.1.5",
```
krassowski, line 51:
```suggestion
```

packages/mathjax-extension/package.json
```diff
@@ -24,30 +32,32 @@
...
     "rimraf": "~3.0.0",
```
krassowski, line 52:
```suggestion
    "rimraf": "~3.0.0",
    "typedoc": "~0.22.10",
```

packages/mathjax-extension/package.json
```diff
@@ -24,30 +32,32 @@
...
     "lib/**/*.{d.ts,eot,gif,html,jpg,js,js.map,json,png,svg,woff2,ttf}",
     "style/**/*.{css,eot,gif,html,jpg,json,png,svg,woff2,ttf}",
+    "schema/*.json",
     "style/index.js"
   ],
   "scripts": {
     "build": "tsc -b",
     "clean": "rimraf lib && rimraf tsconfig.tsbuildinfo",
-    "docs": "typedoc src",
+    "eslint": "eslint . --ext .ts,.tsx --fix",
```
krassowski, line 41:
```suggestion
    "docs": "typedoc src",
    "eslint": "eslint . --ext .ts,.tsx --fix",
```

packages/mathjax-extension/package.json
krassowski: It appears that we do not use it?

packages/mathjax-extension/package.json
krassowski: We most definitely should preserve the documentation.

krassowski:

One important thing to consider is whether we should depend on [`mathjax-full`](https://www.npmjs.com/package/mathjax-full) or [`mathjax`](https://www.npmjs.com/package/mathjax). I think that mathjax could be a better choice - it is 10 MB lighter (and we can include our own fonts/menu bindings for better UX anyways). I wonder if you looked into this and have any thoughts. We do not need to do this in this PR but that would be a 4.0 item too.

Do you think it is worth sending another PR to include your instructions in the [JupyterLab Contributing guide](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html)? I would exclude installation steps, the suggested wget-from-github installation is not the best practice, so I would leave it to the end user how they install it. I would personally add a mention that one can alternatively do that with `git filter-branch`-foo too if they wish.

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      label: args =>
+        'Mathjax Scale ' + (args['scale'] ? `x${args['scale']}` : 'Reset')
+    });
+  }
+
+  /**
+   * Typeset the math in a node.
+   */
+  typeset(node: HTMLElement): void {
+    this._mathDocument.options.elements = [node];
```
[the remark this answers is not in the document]
fcollonval: This is what was actually done in `mathjax2`.

packages/mathjax2-extension/src/index.ts
```diff
@@ -1,42 +0,0 @@
...
- * The MathJax latexTypesetter plugin.
- */
-const plugin: JupyterFrontEndPlugin<ILatexTypesetter> = {
-  id: '@jupyterlab/mathjax2-extension:plugin',
-  autoStart: true,
-  provides: ILatexTypesetter,
-  activate: () => {
-    const [urlParam, configParam] = ['fullMathjaxUrl', 'mathjaxConfig'];
-    const url = PageConfig.getOption(urlParam);
-    const config = PageConfig.getOption(configParam);
```
fcollonval, line 25: Some changes must be done on the backend side as we are not using this page config options anymore.

fcollonval (changes requested):
Thanks @dlqqq 


I left some comments.
Could you also add some note in the [migration guide](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst) especially regarding the drop of some page configuration parameters.

packages/mathjax-extension/package.json
dlqqq: Removed in latest revision :+1:

packages/mathjax-extension/package.json
dlqqq: Added in latest revision 👍

packages/mathjax2-extension/src/index.ts
dlqqq: Do these page config options live anywhere besides `jupyterlab_server` (JLS)? I'm happy to author the PR to remove these options.

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import { TeXFont } from 'mathjax-full/js/output/chtml/fonts/tex';
+
+import { AllPackages } from 'mathjax-full/js/input/tex/AllPackages';
+
+import { SafeHandler } from 'mathjax-full/js/ui/safe/SafeHandler';
+
+import { HTMLHandler } from 'mathjax-full/js/handlers/html/HTMLHandler';
+
+import { browserAdaptor } from 'mathjax-full/js/adaptors/browserAdaptor';
```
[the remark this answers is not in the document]
dlqqq: It's not clear to me whether this is possible, since there is a top-level statement that requires `mathjax` to be imported:

```
mathjax.handlers.register(SafeHandler(new HTMLHandler(browserAdaptor())));
```

If this can be done lazily when `typeset()` is called, then yes; otherwise no.

dlqqq:
@krassowski @fcollonval 

> Should we import those asynchronously to fasten JupyterLab opening?

> One important thing to consider is whether we should depend on [mathjax-full](https://www.npmjs.com/package/mathjax-full) or [mathjax](https://www.npmjs.com/package/mathjax). I think that mathjax could be a better choice - it is 10 MB lighter (and we can include our own fonts/menu bindings for better UX anyways). I wonder if you looked into this and have any thoughts. We do not need to do this in this PR but that would be a 4.0 item too.

I agree that this probably shouldn't be done in this PR. I lack context for this code as I am not one of the original authors of this extension. It's difficult for me to make judgement calls on whether this extension should do X or do Y and the side effects those would cause. Perhaps it's more productive to have these discussions with @jasongrout or @ian-r-rose?

Of course, I can take ownership of this if nobody else is willing to contribute; I just wanted to call out that there may be others more able to help.

packages/mathjax-extension/src/index.ts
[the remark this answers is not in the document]
krassowski: I think you can call it later, just define a global variable to track of it was already called.

There is a trade-off with importing it only at first typeset call. An alternative would be to import async on application ready signal, this way we still get the saving for first load but without the penalty when first opening a notebook with math.

dlqqq:
Hm, on the other hand, a 34 MB bundle does have a significant impact on initial load performance, and that kind of defeats the original intent of improving JL 4 performance. I'm wary of making any changes to the behavior of this extension without getting reviews from the original authors, but I can make these changes now and get reviews later.

packages/mathjax2-extension/src/index.ts
fcollonval: Indeed there is the set up in `jupyterlab_server` at https://github.com/jupyterlab/jupyterlab_server/blob/fb77c74f022d439249f4529687db1582bfe282b3/jupyterlab_server/handlers.py#L108

Could you also open an issue on [jupyter_server_mathjax](https://github.com/jupyter-server/jupyter_server_mathjax) repository as it won't work with JupyterLab 4?

packages/mathjax-extension/src/index.ts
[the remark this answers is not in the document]
dlqqq: Implemented in latest revision.

packages/mathjax2-extension/src/index.ts
dlqqq: Can these be done after this PR is merged? Or do the page config options need to be removed in JLS before upgrading to MathJax 3 in JL?

dlqqq: @fcollonval

dlqqq:
@fcollonval 

Dynamic import implemented in this commit: https://github.com/jupyterlab/jupyterlab/pull/13877/commits/048c591bd9d60403c9ff5d6dfc1046ffd993a90a

I tested and verified that it pulls assets only when opening a math document for the first time. I didn't find any regressions, but I also encourage others to pull this locally and see if they find anything.

If the changes to remove MathJax page options in JLS can be done afterwards, this PR should be ready to merge pending review.

@krassowski 

> I think that mathjax could be a better choice - it is 10 MB lighter (and we can include our own fonts/menu bindings for better UX anyways).

I looked into this for you. Unfortunately there are two blockers that prevent us from using just the base `mathjax` package:

1. Lack of TypeScript declaration files. `@types/mathjax` does not provide the correct types for the latest version of `mathjax` either.

2. Dependency on internal library functions. It looks like we're not using exclusively public interfaces, but rather using some undocumented private interfaces exclusive to `mathjax-full`. For example, behavior of the lines

```
await import('mathjax-full/js/input/tex/require/RequireConfiguration');
...
mathjax.handlers.register(SafeHandler(new HTMLHandler(browserAdaptor())));
```

don't appear to be public or documented at all.

dlqqq:
Hey team, the latest version bump to 4.0.0-alpha.19 caused a lot of merge conflicts. Unfortunately I was not able to fix them with rebasing, as that didn't work with git-subtree's merge strategy. I was forced to use a merge commit instead. I'd like for us to get this merged sooner rather than later, given how much manual intervention in this branch is involved with every alpha release.

I believe CI is still failing because there is no version of `@jupyterlab/mathjax-extension@4.0.0-alpha.19` published. I'm also unable to update `jupyterlab/staging/yarn.lock` until then. Could somebody with permissions help me here? 🤗

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+} from '@jupyterlab/application';
+
+import { ILatexTypesetter } from '@jupyterlab/rendermime';
+
+import type { MathDocument } from 'mathjax-full/js/core/MathDocument';
+
+enum CommandIDs {
+  copy = 'mathjax:clipboard',
+  scale = 'mathjax:scale'
+}
```
fcollonval, line 20: Let's be consistent with the rest of the code

```suggestion
namespace CommandIDs {
  /**
   * Copy MathJax to clipboard
   */
  export const copy = 'mathjax:clipboard';
  /**
   * Scale MathJax elements.
   */
  export const scale = 'mathjax:scale';
}
```

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+        'Mathjax Scale ' + (args['scale'] ? `x${args['scale']}` : 'Reset')
+    });
+  }
+
+  protected async _ensureInitialized() {
+    if (this._initialized) {
+      return;
+    }
+
+    await import('mathjax-full/js/input/tex/require/RequireConfiguration');
```
fcollonval, line 53: As mentioned by @krassowski, I do think to that we should not import `mathjax-full` but the lighter `mathjax`. I saw that it brings question. But we should spend time trying to use the lighter version.

fcollonval (changes requested):
Thanks @dlqqq 

I left some questions.

There is also an opened question about the consequence for the server settings (in particular the case of [jupyter_server_mathjax](https://github.com/jupyter-server/jupyter_server_mathjax)).

The linter is also unhappy. If you execute eslint, you will see there is promise not handled:

```
/home/runner/work/jupyterlab/jupyterlab/packages/mathjax-extension/src/index.ts
Error:   31:9  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

And a regression is seen in the UI tests: a separator is appearing in the code cell context menu:

image

dlqqq:
@fcollonval This is pretty confusing for new developers: there are actually five different `eslint` scripts:

```
"eslint": "eslint --ext .js,.jsx,.ts,.tsx --cache --fix .",
"eslint:check": "eslint --ext .js,.jsx,.ts,.tsx --cache .",
"eslint:check:typed": "eslint --config .eslintrc.typecheck.js --ext .js,.jsx,.ts,.tsx --cache .",
"eslint:files": "eslint --fix",
"eslint:typed": "eslint --config .eslintrc.typecheck.js --cache --ext .js,.jsx,.ts,.tsx --fix .",
```

`yarn lint` calls `yarn eslint`, `pre-commit` calls `yarn eslint:files`, and the CI calls `yarn eslint` and `yarn eslint:check:typed`! This is very confusing.

Do you have any immediate objections to me opening a separate issue to discuss consolidating these scripts together?

packages/mathjax-extension/src/index.ts
dlqqq: 1. This should have no impact on initial bundle size (and by extension, initial load times), because `mathjax-full` is loaded dynamically.

2. There is a potential, unknown security risk with removing the line

```
mathjax.handlers.register(SafeHandler(new HTMLHandler(browserAdaptor())));
```

which requires `mathjax-full`, as these are internal undocumented library classes. Replacing `mathjax-full` without first consulting somebody more knowledgable on this (e.g. the extension contributors or MathJax contributors) is dangerously hasty and imposes risk to users.

3. Part of the large difference in package sizes in `mathjax` and `mathjax-full` is just the presence of comprehensive declaration files that are entirely missing in `mathjax`. We must be careful to distinguish **package size** and **impact on bundle size**. Declaration files increase package size but have no impact on bundle size. Build optimizations such as dependency deduping complicate this further, as even library code included in a package may have no impact on bundle size.

4. If we discover a way to safely replace `mathjax-full` with `mathjax`, this would not be a breaking change, and hence can be addressed in a follow-up PR.

packages/mathjax-extension/src/index.ts
dlqqq: Fixed in latest revision.

dlqqq:
@fcollonval All of your concerns have been addressed, although I have not switched to using `mathjax` for reasons mentioned above.

dlqqq:
Can we also continue this discussion? https://github.com/jupyterlab/jupyterlab/pull/13877#discussion_r1090338268

packages/mathjax2-extension/src/index.ts
fcollonval: Yes there is no problem doing that in follow-up.

jupyterlab/staging/package.json
```diff
@@ -71,8 +71,7 @@
     "@jupyterlab/markdownviewer": "~4.0.0-alpha.19",
     "@jupyterlab/markdownviewer-extension": "~4.0.0-alpha.19",
     "@jupyterlab/markedparser-extension": "~4.0.0-alpha.19",
-    "@jupyterlab/mathjax2": "~4.0.0-alpha.19",
-    "@jupyterlab/mathjax2-extension": "~4.0.0-alpha.19",
+    "@jupyterlab/mathjax-extension": "~4.0.0-alpha.19",
```
jtpio, line 74: We should be able to revert the change made to `jupyterlab/staging/package.json` since this file is updated during the release process.

jtpio: This will help make this CI check pass: https://github.com/jupyterlab/jupyterlab/actions/runs/4189017035/jobs/7260844694

jtpio:
There seems to be a remaining TypeScript issue to fix: https://github.com/jupyterlab/jupyterlab/actions/runs/4189017045/jobs/7260847385

```
Error: test/rendermime/registry.spec.ts(108,51): error TS2345: Argument of type '{ url: string; config: string; }' is not assignable to parameter of type 'JupyterFrontEnd<IShell, "desktop" | "mobile">'.
```

examples/notebook/src/index.ts
```diff
@@ -16,7 +16,7 @@ import '@jupyterlab/theme-light-extension/style/theme.css';
 import '../index.css';
 
 import { ServiceManager } from '@jupyterlab/services';
-import { MathJaxTypesetter } from '@jupyterlab/mathjax2';
+import { MathJaxTypesetter } from '@jupyterlab/mathjax-extension';
```
jtpio, line 19: Looks like the new `MathJaxTypesetter` now takes a `JupyterFrontEnd` in its constructor, while the code below in this file still uses the previous initialization options.

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  protected _initialized: boolean = false;
+  protected _mathDocument: MathDocument<any, any, any>;
+}
+
+/**
+ * The MathJax extension.
+ */
+const mathJaxPlugin: JupyterFrontEndPlugin<ILatexTypesetter> = {
+  id: '@jupyterlab/mathjax-extension:plugin',
+  requires: [],
```
jtpio, line 132: This line could be removed if the list is empty:

```suggestion
```

jtpio:
@dlqqq just checking if you were able to have a look at the latest comment in this PR?

Otherwise happy to help push this across the finish line.

examples/notebook/src/index.ts
dlqqq: Sorry, but what exactly are these directories under `examples/`? I tried looking for references of this in the README.md and the online documentation, but can't find anything on this.

I need help retrieving a reference to a `JupyterFrontEnd` instance. I'm assuming this is a singleton, so I can't just instantiate a new one here. Usually I only work within the context of an extension, so that's just given to me as the first argument when `JupyterFrontEndPlugin.activate()` is called. It looks like in dev mode, `window.jupyterapp` is also available, but I'm not certain that these example directories are run in dev mode.

I'd be happy to send out a PR later adding a new `examples/README.md` file to help explain this to new developers like myself as well.

packages/mathjax-extension/tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
```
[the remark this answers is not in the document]
dlqqq: Done. I had to keep `skipLibCheck: true` because it looks like the `mathjax-full` types aren't self-consistent.

dlqqq:
@jtpio Sorry for the delay, this week has been very busy for me. 😅 Anyways, I addressed most of your comments, but I do need help with fixing the `examples/` directory since I have no context on its purpose.

krassowski:
> the examples/ directory since I have no context on its purpose.

These are standalone `examples/` to help kick-start work for developers who want to build a third-party app re-using JupyterLab components. We have documentation:

https://jupyterlab.readthedocs.io/en/stable/developer/repo.html#examples-examples

examples/notebook/src/index.ts
krassowski: Here is a simple solution:
1. Change `MathJaxTypesetter` constructor to take `commands: CommandRegistry` and update the initialization as needed
2. In example below pass `commands`

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    app.commands.addCommand(CommandIDs.copy, {
+      execute: async (args: any) => {
+        const md = this._mathDocument;
+        const oJax: any = md.outputJax;
+        await navigator.clipboard.writeText(oJax.math.math);
+      },
+      label: 'MathJax Copy Latex'
+    });
+
+    app.commands.addCommand(CommandIDs.scale, {
```
krassowski, line 42:
```suggestion
  constructor(commands: CommandRegistry) {
    commands.addCommand(CommandIDs.copy, {
      execute: async (args: any) => {
        const md = this._mathDocument;
        const oJax: any = md.outputJax;
        await navigator.clipboard.writeText(oJax.math.math);
      },
      label: 'MathJax Copy Latex'
    });

    commands.addCommand(CommandIDs.scale, {
```

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @module mathjax-extension
+ */
+
+import {
+  JupyterFrontEnd,
+  JupyterFrontEndPlugin
+} from '@jupyterlab/application';
+
+import { ILatexTypesetter } from '@jupyterlab/rendermime';
+
```
krassowski, line 14:
```suggestion

import { CommandRegistry } from '@lumino/commands';

```

packages/mathjax-extension/src/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  protected _mathDocument: MathDocument<any, any, any>;
+}
+
+/**
+ * The MathJax extension.
+ */
+const mathJaxPlugin: JupyterFrontEndPlugin<ILatexTypesetter> = {
+  id: '@jupyterlab/mathjax-extension:plugin',
+  provides: ILatexTypesetter,
+  activate: (app: JupyterFrontEnd) => new MathJaxTypesetter(app),
```
krassowski, line 133:
```suggestion
  activate: (app: JupyterFrontEnd) => new MathJaxTypesetter(app.commands),
```

examples/notebook/src/index.ts
krassowski: Even better, remove the args in constructor and move them to plugin. I don't think the commands should be added in the typesetter itself.

fcollonval (changes requested):
I looked at mathjax 3 documentation. Actually there is a conversion tool for config moving from 2 to 3: https://mathjax.github.io/MathJax-demos-web/convert-configuration/convert-configuration.html

In our case it gives:

image

This points out it is possible to load only a very specific Javasript file that should address the package size concern.

Regarding the safe option, it was implemented later and available through a config; see documentation:

https://docs.mathjax.org/en/v3.2-latest/options/safe.html

examples/notebook/src/index.ts
dlqqq: Fixed in latest revision.

packages/mathjax-extension/src/index.ts
dlqqq: Fixed in latest revision.

packages/mathjax-extension/src/index.ts
dlqqq: Fixed in latest revision.

packages/mathjax-extension/src/index.ts
dlqqq: Fixed in latest revision.

dlqqq:
@krassowski I implemented your suggestion of moving the command activation to the plugin body instead of the constructor body. This is much better, thank you! 🤗

dlqqq:
Doesn't look like any of the visual regression failures are a result of this PR; the same visual regressions are happening in https://github.com/jupyterlab/jupyterlab/pull/14102

dlqqq:
please update snapshots

dlqqq:
bot please update snapshots

dlqqq:
@fcollonval I sent out an email asking to set up a meeting to discuss next steps with this PR. Can you verify you received it? I think that once I fix merge conflicts and CI (again), there are no outstanding issues and that this PR is ready to be merged; I have addressed your concerns about the `mathjax-full` package multiple times.

Also, I strongly discourage merging https://github.com/jupyterlab/jupyterlab/pull/14157. The Git history in this PR is complex, and merging that PR first will create extremely difficult merge conflicts. I advise us to first focus on determining whether this PR is ready to merge, and if not, what the **minimum set of changes** are necessary to get this PR ready to merge.

fcollonval:
Some figures about the current PR (https://github.com/jupyterlab/jupyterlab/pull/13877/commits/fba493786c2f4fff235488a17a042735c6b4edd3) vs current master (https://github.com/jupyterlab/jupyterlab/commit/cbf39798a05dfbf533b5706b6e1dbd1cc65b00f5) - the wheel would be increase by 1Mo (+12Mo):

image

krassowski:
I sympathise with @dlqqq here and agree that we should hold https://github.com/jupyterlab/jupyterlab/pull/14157 and merge this one **if** it can be finalised ASAP. I do not see much to be done here other than migrating to `mathjax` over `mathjax-full`.

@dlqqq you say:

> I have addressed your concerns about the `mathjax-full` package multiple times.

Would you mind summarising the current status of this? It is not clear to me what is blocking us as it appears that we can safely migrate. Did you open any issues upstream with MathJax team to consult on any doubts/discuss upstream enhancements we may need?

fcollonval:
We (Jeremy, Brian, David and I) have a discussion about this one tomorrow at 6PM CET. I happily forward the invitation if you are interested @krassowski.

My main concern is the assets size induced by the current approach. From diving into MathJax 3 documentation, the recommended way to load it with specific subset is through CDN. But that raises the question of how to deal with the server side traitlets used for now.

dlqqq:
Here is a list of network requests being made to fetch MJ3 on the current branch. The assets total roughly 4.7 MB.

Screen Shot 2023-03-13 at 9 10 13 AM

However, I think this figure can be greatly reduced if we're able to remove some of the additional code the old MJ3 extension imported.

```ts
await import('mathjax-full/js/input/tex/require/RequireConfiguration');
    const { mathjax } = await import('mathjax-full/js/mathjax');
    const { CHTML } = await import('mathjax-full/js/output/chtml');
    const { TeX } = await import('mathjax-full/js/input/tex');
    const { TeXFont } = await import('mathjax-full/js/output/chtml/fonts/tex');
    const { AllPackages } = await import(
      'mathjax-full/js/input/tex/AllPackages'
    );
    const { SafeHandler } = await import('mathjax-full/js/ui/safe/SafeHandler');
    const { HTMLHandler } = await import(
      'mathjax-full/js/handlers/html/HTMLHandler'
    );
    const { browserAdaptor } = await import(
      'mathjax-full/js/adaptors/browserAdaptor'
    );
```

@krassowski Let me know if you're able to attend the meeting tomorrow to hear my thoughts on the matter. Otherwise, I can summarize the entirety of the preceding discussion regarding `mathjax` v.s. `mathjax-full` as a comment on this PR.

krassowski:
> I happily forward the invitation if you are interested @krassowski.

Please do :)

fcollonval:
> > I happily forward the invitation if you are interested @krassowski.
> 
> Please do :)

I sent an invitation (using the email address of your JupyterLab council membership).

dlqqq:
Brief summary of our discussion: This PR can be merged pending green CI and resolution of merge conflicts. However, I need to create an issue to investigate strategies to minimize the size of the MathJax assets being loaded client-side and determine if it's possible to migrate from `mathjax-full` to `mathjax` to minimize wheel size. I will be chiefly responsible for this effort. I will also open an issue on the MathJax repo and see if we can solicit their input and guidance on how to simplify the code in the current MathJax extension.

dlqqq:
I've created the issues on the JL and MJ repos. Will fix merge conflicts and get CI green tomorrow.

dlqqq:
CI failures all seem to be related to `yarn install` failing. @jtpio I need an interim release of `@jupyterlab/mathjax-extension` to get CI passing.

```
[LabBuildApp] yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
error Couldn't find any versions for "@jupyterlab/mathjax-extension" that matches "^4.0.0-alpha.21"
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Other CI failures are due to upstream issues, e.g. [server error from pre-commit.ci](https://results.pre-commit.ci/run/github/60374819/1678890613.VmQEH34SQOuxM1bxZpfPZQ) and [failing visual regression tests](https://github.com/jupyterlab/jupyterlab/actions/runs/4424325747/jobs/7758015403) on latest version of master.

jtpio:
> nterim release of `@jupyterlab/mathjax-extension` to get CI passing.

Normally the package should already be published on npm (from the last time): https://www.npmjs.com/package/@jupyterlab/mathjax-extension

dlqqq:
@jtpio The latest version published is `4.0.0-alpha.17`, whereas this branch requires `4.0.0-alpha.21`.

jtpio:
Right, I'll have a look and report here.

jtpio:
@dlqqq I just published a newer version and restarted the CI checks.

fcollonval:
bot please update snapshots

fcollonval:
> Remaining CI failures are not relevant

jmshea:
Can anyone help me understand how to call MathJax methods now? 

*Background:* My Python+Javascript library JupyterQuiz generates interactive quizzes that can include LaTeX, but I need to call the appropriate method to tell MathJax to typeset the LaTeX code, because it is loaded by Javascript. For MathJax 3, the usual way to do this to to pass the element to MathJax.typeset(), but the MathJax global object in JupyterLab 4+ does not have the appropriate methods -- it still looks like the pre-loading version of the MathJax object.

*Reference:* From the MathJax 3 docs: "The global variable MathJax is used to store the configuration for MathJax. Once MathJax is loaded, however, MathJax changes the MathJax variable to contain the various methods needed to control MathJax." 

**It seems that the MathJax variable never gets updated in JupyterLab 4.**

Any help would be greatly appreciated!

krassowski:
@jmshea folks usually don't look at comments on closed PRs so I would suggest that you open a new issue instead, please.

Issue #13870, opened 2023-01-27
JupyterLab 3.6.0rc1 - style changes


fcollonval:
## Description

In JupyterLab 3.6.0rc1, there is two styling changes that I'm not sure feeling comfortable.

![image](https://user-images.githubusercontent.com/8435071/215058145-e79b2002-4b36-4af8-b735-1afa809e5263.png)

The menu is position below the menubar due to the new optimization introduced in Lumino https://github.com/jupyterlab/lumino/pull/432 - this prevent the y position trick to be applied:

https://github.com/jupyterlab/jupyterlab/blob/7227449db8db478dc5e31bbc1ffdfb741766a35d/packages/application/style/menus.css#L30

A submenu get a outline focus. It feels wrong.

> The focus effect is probably the consequence of https://github.com/jupyterlab/jupyterlab/pull/13415

For reference, here is the look in JupyterLab 3.5.3.

![image](https://user-images.githubusercontent.com/8435071/215059530-5ca5d376-53f3-46b0-b7b2-728bab9a8bef.png)

## Reproduce

Install 3.6.0rc1 by executing `python -m pip install jupyterlab==3.6.0rc1`

fperez:
Thx @fcollonval for all your work on these releases! 

I totally agree these visual changes are worrisome. This is what even the main menus look like for me:

![image](https://user-images.githubusercontent.com/57394/215180540-257ce341-9c02-42f6-ab21-c27577374396.png)

Not only is that vertical gap odd, the bold blue outlines on all menus really shouldn't be there. They even persist when a menu is closed:

![image](https://user-images.githubusercontent.com/57394/215181058-7b3628f3-8d46-4df3-877a-77a5fda8fe4a.png)

I think these should be fixed to the behavior of 3.5.x before the 3.6.0 release. B/c of the RTC improvements, 3.6 should really be a very good upgrade for many (we're doing experimental testing for it both at Berkeley and LBL). I really don't think we should have these odd visuals in a widely deployed/advertised release.

krassowski:
CC @gabalafou

krassowski:
The outline thingy on menus is confusing because I cannot reproduce it on master (is it a me problem?), but I do see it on 3.6 branch. I agree that having both outline and background effect is redundant. I am nor sure about the full menu outline effect (this one does not have an existing indicator and may be needed indeed) but that could possibly have a more subtle visual indicator,

krassowski:
[...]

PR #13878, opened 2023-01-28
Restore partial border effect for menu

krassowski:
## References

Fixer partial border styling of menu with menubar on 3.6.x reported in https://github.com/jupyterlab/jupyterlab/issues/13870

cherry-picked from 18c3af7de59f2eb44c45f58b98cb3c4d46afce72

## Code changes

This is a simple as switching from `transform` to `top` (the version for 4.x also requires strict containment relaxation, but we don't have containment on all size- managed widgets in 3.x so it does not apply to the backport)

## User-facing changes

Before vs after:

comparison

None

### DIFF
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -27,7 +27,7 @@
 }
 
 .lm-MenuBar-menu {
-  transform: translateY(calc(-2 * var(--jp-border-width)));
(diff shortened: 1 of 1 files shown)


krassowski:
bot please update snapshots

fcollonval (commented):
Thanks @krassowski

PR #13883, opened 2023-01-29
Fix searching backwards in notebook

### LINKED ISSUE #13880: Search in notebook can enter infinite loop in 4.0
## Description

Searching in notebook can enter infinite loop because of invalid assumptions in code.

## Reproduce

1. Open a notebook and search box
2. Type a phrase which is not in a notebook
3. Press arrow up or shift + enter to find previous match
4. See UI freeze after entering infinite loop

[Screencast from 2023-01-28 20-55-53.webm](https://user-images.githubusercontent.com/5832902/215291104-a7d65e69-59af-4a93-9f1e-f309b2821e22.webm)

The looping logic will only break if it visits the search provider with index equal to the `startIndex`:

https://github.com/jupyterlab/jupyterlab/blob/2ba9f60509821b4db3f76532084dce71be2457f1/packages/notebook/src/searchprovider.ts#L539-L557

which is taken from `activeCellIndex`:

https://github.com/jupyterlab/jupyterlab/blob/2ba9f60509821b4db3f76532084dce71be2457f1/packages/notebook/src/searchprovider.ts#L524

But it never actually is visited under some circumstances:

1. when `activeCellIndex` is 0, as this would be the code sequence:
    ```js
    this._currentProviderIndex = this._currentProviderIndex + (reverse ? -1 : 1); // 0 + 1 = 1
    
    if (loop) { 
      if (this._currentProviderIndex === startIndex) { // 1 != 0
         break;
      }
      
      // new value is (1 + 1) % 1 = 0, hence infinite loop
       this._currentProviderIndex = 
           (this._currentProviderIndex + this._searchProviders.length) % 
           this._searchProviders.length;
    }

   // while condition is also always fulfilled as `0 <= 0 && 0 < 1`
    ```
2. when `activeCellIndex` is -1:  
     https://github.com/jupyterlab/jupyterlab/blob/2ba9f60509821b4db3f76532084dce71be2457f1/packages/notebook/src/widget.ts#L1305-L1310

    ```js
    this._currentProviderIndex = this._currentProviderIndex + (reverse ? -1 : 1); // -1 + 1 = 0
    
    if (loop) { 
      if (this._currentProviderIndex === startIndex) { // 0 != -1
         break;
      }

      // new value is (0 + 1) % 1 = 0, hence infinite loop as per case (1)
       this._currentProviderIndex = 
           (this._currentProviderIndex + this._searchProviders.length) % 
           this._searchProviders.length;
    }

   // while condition is also always fulfilled as `0 <= 0 && 0 < 1`
    ```

## Expected behavior

No infinite loops ever

## Context

- JupyterLab version: 4.0.0a32
- this is distinct from similar https://github.com/jupyterlab/jupyterlab/issues/13283

krassowski:
## References

Fixes #13880

## Code changes

- `highlightPrevious` in cell search provider now has previously missing renderer signal handler block which was already present in `highlightNext`
- the logic in the do-while loop was corrected

## User-facing changes

Search in notebook is usable again.

N/A

### DIFF
--- a/packages/cells/src/searchprovider.ts
+++ b/packages/cells/src/searchprovider.ts
@@ -658,7 +658,9 @@ class MarkdownCellSearchProvider extends CellSearchProvider {
     if (cell.rendered && this.matchesCount > 0) {
       // Unrender the cell if there are matches within the cell
       this._unrenderedByHighligh = true;
+      const waitForRendered = signalToPromise(cell.renderedChanged);
       cell.rendered = false;
+      await waitForRendered;
     }
 
     match = await super.highlightPrevious();
--- a/packages/notebook/src/searchprovider.ts
+++ b/packages/notebook/src/searchprovider.ts
@@ -540,34 +540,18 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
           this._currentProviderIndex + (reverse ? -1 : 1);
 
         if (loop) {
-          // We loop on all cells, not hit found
-          if (this._currentProviderIndex === startIndex) {
-            break;
-          }
-
           this._currentProviderIndex =
             (this._currentProviderIndex + this._searchProviders.length) %
             this._searchProviders.length;
         }
       }
     } while (
-      0 <= this._currentProviderIndex &&
-      this._currentProviderIndex < this._searchProviders.length
+      loop
+        ? // We looped on all cells, no hit found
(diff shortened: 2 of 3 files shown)


fcollonval (approved):
Thanks a lot @krassowski

fcollonval:
> CI failures are not related

PR #13884, opened 2023-01-29
Create a unified editor search provider

### LINKED ISSUE #13282: Searching in files steals focus, prevents typing the search phrase (regression)

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13283 and provides a tentative fix for https://github.com/jupyterlab/jupyterlab/issues/13885

Also fixes #13282

## Code changes

Simplifies searching in file editor by removing `CodeMirrorSearchProvider` and replacing it with abstract `EditorSearchProvider` which expects subclasses to define `editor` and `model` properties. `CellSearchProvider` and `FileEditorSearchProvider` now inherit from the new `EditorSearchProvider` which is essentially identical to the old `CellSearchProvider`.

## User-facing changes

Searching in editor works reliably again.


packages/codemirror/src/searchprovider.ts
```diff
@@ -590,7 +431,8 @@ export class CodeMirrorSearchHighlighter {
                     m.position,
                     m.position + m.text.length
                   )
-                )
+                ),
+                filter: () => false
```
fcollonval, line 435: It seems that this is the way: see for example the [match decorator](https://github.com/codemirror/view/blob/91e0a7d1e2c0fc1f7c57a9a29233128b617f1c1a/src/matchdecorator.ts#L143).

I guess the reason for the filtering is to be able to handle just the part of the document that is in the view port.

packages/fileeditor/src/searchprovider.ts
```diff
@@ -20,15 +21,34 @@ export type FileEditorPanel = MainAreaWidget<FileEditor>;
...
    * Constructor
    * @param widget File editor panel
    */
-  constructor(widget: FileEditorPanel) {
-    super(widget.content.editor as CodeMirrorEditor);
+  constructor(protected widget: FileEditorPanel) {
+    super();
+  }
+
+  isReadOnly = false;
```
fcollonval, line 35: Should we read that information from the editor (in `EditorSearchProvider`):

```suggestion
  get isReadOnly(): boolean {
    return this.editor.getOption('readOnly');
  }
```

packages/fileeditor/src/searchprovider.ts
```diff
@@ -20,15 +21,34 @@ export type FileEditorPanel = MainAreaWidget<FileEditor>;
...
+    return {
+      preserveCase: true
+    };
+  }
+
+  get editor() {
+    return this.widget.content.editor as CodeMirrorEditor;
+  }
+
+  get model(): CodeEditor.IModel {
```
fcollonval, line 50:
```suggestion
  /**
   * Editor content model
   */
  get model(): CodeEditor.IModel {
```

packages/codemirror/src/searchprovider.ts
```diff
@@ -30,521 +30,362 @@
...
 
-  private _parseMatchesFromState(): ISearchMatch[] {
-    // Flatten state map
-    const matches = new Array<ISearchMatch>();
-
-    for (const lineKey in this._matchState) {
-      const lineMatches = this._matchState[lineKey];
-      for (const posKey in lineMatches) {
-        matches.push(lineMatches[posKey]);
+  protected async updateCodeMirror(content: string) {
```
fcollonval, line 358:
```suggestion
  /**
   * Update matches
   */
  protected async updateCodeMirror(content: string) {
```

packages/codemirror/src/searchprovider.ts
```diff
@@ -30,521 +30,362 @@
...
+    if (this.query !== null) {
+      if (this.isActive) {
+        this.cmHandler.matches = await TextSearchEngine.search(
+          this.query,
+          content
+        );
+      } else {
+        this.cmHandler.matches = [];
```
fcollonval, line 366: Should we always fallback to `[]` even when `query` is null? To be sure to clean any search?

packages/codemirror/src/searchprovider.ts
```diff
@@ -30,521 +30,362 @@
...
+  protected get cmHandler(): CodeMirrorSearchHighlighter {
+    if (!this._cmHandler) {
+      this._cmHandler = new CodeMirrorSearchHighlighter(
+        this.editor as CodeMirrorEditor
+      );
+    }
+    return this._cmHandler;
   }
 
+  protected abstract get editor(): CodeEditor.IEditor | null;
```
fcollonval, line 75:
```suggestion
  /**
   * Text editor
   */
  protected abstract get editor(): CodeEditor.IEditor | null;
```

fcollonval (changes requested):
Thanks @krassowski

packages/fileeditor/src/searchprovider.ts
krassowski: Note to self: maybe add a test for this one.

packages/codemirror/src/searchprovider.ts
krassowski: > It seems that this is the way: see for example the [match decorator](https://github.com/codemirror/view/blob/91e0a7d1e2c0fc1f7c57a9a29233128b617f1c1a/src/matchdecorator.ts#L143)..
> 
> I guess the reason for the filtering is to be able to handle just the part of the document that is in the view port.

So this was added to fix https://github.com/jupyterlab/jupyterlab/issues/13885. FYI it only applies to old ranges, not the once that get added; the docs for `filter()` says:

> Filter the ranges already in the set. Only those for which this function returns true are kept.

Probably worth adding a comment.

krassowski:

krassowski:

krassowski:
bot please update snapshots

krassowski:

it seems the update job is timing out

krassowski:

packages/documentsearch/test/searchmodel.spec.ts
```diff
@@ -61,6 +62,18 @@ describe('documentsearch/searchmodel', () => {
       });
     });
 
+    describe('#parsingError', () => {
+      it('should set informative string message on invalid regex', async () => {
+        model.useRegex = true;
+        expect(model.parsingError).toEqual('');
+        model.searchExpression = 'query\\';
+        await signalToPromise(model.stateChanged);
```
fcollonval, line 70: It can happen that the signal is already emitted prior to the connection done in `signalToPromise`. So I would advice the safer:

```suggestion
        const waitForStateChanged = signalToPromise(model.stateChanged);
        model.searchExpression = 'query\\';
        await waitForStateChanged;
```

fcollonval (commented):
A part from a minor suggestions, if the snapshots would like to be updated... it is good to go for me.

krassowski:

krassowski:
I increase timeout on galata update workflow, let's see if this helps.

fcollonval:

krassowski:
bot can I very kindly ask to please update snapshots

fcollonval:
@krassowski I have the impression there is something about this PR breaking the update job. I ran the update job successfully on a recent PR yesterday: https://github.com/jupyterlab/jupyterlab/actions/runs/4172386115/jobs/7223418464

Did you try running the update for your new test locally? Hopefully that may give us some idea.

krassowski:
@fcollonval no, it is not specific to this PR, see that the Galata update job also failed on another PR after this comment https://github.com/jupyterlab/jupyterlab/pull/13964#issuecomment-1429333164

link to the cancelled (timed out) job:
https://github.com/jupyterlab/jupyterlab/actions/runs/4171947506/jobs/7222458570

> I ran the update job successfully on a recent PR yesterday: https://github.com/jupyterlab/jupyterlab/actions/runs/4172386115/jobs/7223418464

The link you posted only shows documentation update. That works fine. The main galata one does not.

> Did you try running the update for your new test locally?

Yes, it works locally.

fcollonval:
Thanks for the additional cases; so it seems the latest job that succeed if from 5 days ago: https://github.com/jupyterlab/jupyterlab/actions/runs/4135509044/jobs/7148065705 - it uses the same Playwright version; so let's rule that out as sources for now.

The jobs for this PR and the one you referenced have the same strange timeouts in the few jobs that got executed:

```
·/home/runner/work/jupyterlab/jupyterlab/galata/test/documentation/customization.test.ts-snapshots/default-terminal-position-single-documentation-linux.png does not match, writing actual.
·××T/home/runner/work/jupyterlab/jupyterlab/galata/test/documentation/customization.test.ts-snapshots/default-menu-bar-documentation-linux.png does not match, writing actual.
·××T/home/runner/work/jupyterlab/jupyterlab/galata/test/documentation/customization.test.ts-snapshots/customized-terminal-position-single-documentation-linux.png does not match, writing actual.
·××T/home/runner/work/jupyterlab/jupyterlab/galata/test/documentation/customization.test.ts-snapshots/customized-menu-bar-documentation-linux.png does not match, writing actual.
·××T
```

So many timeouts are unexpected when updating snapshots. But the worst is that the log is telling us it is updating documentation snapshots :exploding_head: 

I'll look to that locally.

fcollonval:

fcollonval:
@krassowski I gave a try to rebase this PR to fix the snapshots. But I was not able to get a proper branch :confused: I let you do it as I don't want to mess your repo. Sorry to not be able to help you more on this.

krassowski:

krassowski:
Thanks for the help here, it is greatly appreciated! I am not sure if this will do it, if it still fails I will look into robustifying the completer test.

fcollonval (commented):
Thanks @krassowski 

The UI test is due to flakiness but the JS test is legit.

packages/documentsearch/test/searchmodel.spec.ts
```diff
@@ -61,6 +62,19 @@ describe('documentsearch/searchmodel', () => {
       });
     });
 
+    describe('#parsingError', () => {
+      it('should set informative string message on invalid regex', async () => {
+        model.useRegex = true;
+        expect(model.parsingError).toEqual('');
+        const waitForStateChanged = signalToPromise(model.stateChanged);
+        model.searchExpression = 'query\\';
+        await waitForStateChanged;
```
krassowski, line 71: The failure in tests started appearing after 923de1a106ec8b9d93eac09c94f1fb177142e2df. This is because `stateChanged` gets emitted twice. Reverting to earlier version.

```suggestion
        model.searchExpression = 'query\\';
        await signalToPromise(model.stateChanged);
```

fcollonval (approved):
Thanks a lot for the patience Mike, let's merge this.

> CI failure is unrelated.

PR #13886, opened 2023-01-30
Add copy option for directories

kenyaachon:
Requires: https://github.com/jupyter-server/jupyter_server/pull/1190

# Objective
Enable users to copy both files and directories
Currently, if users try copying a directory, an error will thrown backing saying "Can't Copy directories". 
## User-facing changes
- The copy option in the right click menu, was enabled for directories. 
- 
## jupter_server

- The copy directory functionality has been added to both the FileContentsManager and AsyncFileContentsManager
- Users can copy folders up to 100 MB,
	- The actual limit for copying directories before a timeout occurs is 500MB, but without a loading indicator, if a user tries to copy a folder that is near 500MB, jupyterlab  will seem like its frozen, since its not clear whether the copy operation is complete or in progress. 
	- if we can add a loading indicator in the future to jupyterlab, then we would be able to increase the limit of copying folders to 500 MB

## How to test
Create a virtual environment with venv (you can use what ever viritual environment tool you like)
```sh
python3 -m venv jupyter-connected
source jupyter-connected/bin/activate
pip install --upgrade pip
# install the jupyter_server package into the environment
git clone https://github.com/kenyaachon/jupyter_server.git
cd jupyter_server
git checkout copy-folder-functionality
pip install -e ".[test]"
# verify the dev version (i.e 2.1.0.dev0) of the package is in the environment
pip list | grep jupyter
# install the jupyterlab in development mode with latest changes
cd ..
git clone https://github.com/kenyaachon/jupyterlab.git
cd jupyterlab
git checkout develop
pip install -e .
jlpm install
jlpm run build 
jupyter lab --dev-mode
```

## Limitations
- We were not able to find a suitable solution yet for being able to copy folders above 500MB in size without having a timeout issue

- Users, cannot easily know if the folder they are copying is above 100MB before they try and copy
	- We did attempt in adding the folder size to the preview info that users get when hovering over a file or folder. But this ended up making the filebrowser-extension very slow when loading the correct directory. Since it would have to recalculate the size of the folders in the given directory each time a new path was loaded.


fcollonval (commented):
@kenyaachon 

Would you mind pointing out where the copy of directories has been enabled? As far as I can tell, it is still forbidden:

https://github.com/jupyter-server/jupyter_server/blob/b6c1edb0b205f8d53f1a2e81abb997bfc693f144/jupyter_server/services/contents/manager.py#L653

kenyaachon:
> @kenyaachon
> 
> Would you mind pointing out where the copy of directories has been enabled? As far as I can tell, it is still forbidden:
> 
> https://github.com/jupyter-server/jupyter_server/blob/b6c1edb0b205f8d53f1a2e81abb997bfc693f144/jupyter_server/services/contents/manager.py#L653


The copy of directories has been enabled in this file for both the FileContentsManager and AsyncContentsManager
https://github.com/kenyaachon/jupyter_server/blob/d6a8eba35e49b3a085bfa3e2c6221aa3d3ecada7/jupyter_server/services/contents/filemanager.py#L1078

https://github.com/kenyaachon/jupyter_server/blob/d6a8eba35e49b3a085bfa3e2c6221aa3d3ecada7/jupyter_server/services/contents/filemanager.py#L607

Since I noticed, the subclass of the ContentsManager was called here:
https://github.com/kenyaachon/jupyter_server/blob/d6a8eba35e49b3a085bfa3e2c6221aa3d3ecada7/jupyter_server/serverapp.py#L1478

fcollonval:
Thanks @kenyaachon for providing the context

kenyaachon:
@fcollonval The PR for jupyter_server package has been merged.
https://github.com/jupyter-server/jupyter_server/pull/1190

fcollonval:
Thanks for the update @kenyaachon 

The failure are likely not related to this PR as the integration tests are broken following another PR. I'm working on a fix https://github.com/jupyterlab/jupyterlab/pull/14152

fcollonval:

fcollonval:
Kicking the CI

Issue #9929, opened 2021-03-10
Vega-Lite 5


davidanthoff:
Vega-Lite 5 is out :) Would be great if Jupyter Lab could ship with that. And I actually also noticed that for example vega and the other related deps are also quite outdated, so probably would make sense to update them all as well?

CC @domoritz

domoritz:
We made sure most of Vega-Lite 5 is backwards compatible (even if the schema is not backwards compatible for example for selections).

jasongrout:
Thanks for letting us know!

Interestingly, our extension is for vega 4 and vega-lite 4. I don't think we considered the possibility of vega still being version 4, but vega-lite being updated to a new major version.

I just published updated versions of the vega2 and vega3 packages as well. Perhaps an easy path forward is to publish a vega-lite 5 extension over in https://github.com/jupyterlab/jupyter-renderers - then we wouldn't have to wait until jlab 4 (or even jlab 3.1).

jasongrout:
@domoritz - a couple of questions about vega-lite (and vega):

1. I see the activity in the vega repo slowing down. Is vega maturing, so that the major updates will be more in vega-lite and other projects, but vega remains a stable base with a very slow major version cycle? Or am I reading too much into the vega repo activity right now?
2. Are there any changes needed to how we are using vega-lite (a quick glance at the release notes indicates no real changes?): https://github.com/jupyterlab/jupyterlab/blob/82604b43eb5563b0cee5fc69fcf4ee52ef46b15c/packages/vega5-extension/src/index.ts#L118-L124
3. Does vega-lite version 5.0 understand vega-lite 4 specs? Can we use vega-lite version 5 similar to how we are supporting vega-lite 3 and 4 currently (by passing any vl3 or vl4 spec into vega-lite 4). I know we talked about backwards compatibility last time, and I'm sure this will keep coming up...


Ideally, all we would need to do is add one more mimetype to our existing plugin for the new vega-lite 5 specs, update our vega-lite npm dependency to version 5, and feed all the specs we support (vega-lite 3, vega-lite 4, or vega-lite 5) into the vega-lite npm package version 5, and things would work. Please tell me this is the case :).

domoritz:
1. Vega is quite mature at this point and I personally don't see a reason why it would break backward compatibility anytime soon. @jheer can comment on his view on this. 

2. No changes needed.

[...]

PR #11179, opened 2021-09-29
Backport PR #10796 to 3.2.x: Add Galata in JupyterLab

fcollonval:
To be able to backport the PR, I had to bump the major version (4 -> 5) of `vega-lite`. I'm unsure of the impact for the users.

PR #13893, opened 2023-01-30
Switch to VEGALITE5_MIME_TYPE

ChristopherDavisUCI:
Draft of a fix for #9929.  (Warning: the changes in this PR are very naive.)

Even if what I've done here is too naive, I hope this fix is ultimately simpler than for example #4107, because the `vega5-extension` is already part of JupyterLab.  (Note that the 5 in `vega5-extension` refers to the Vega version, not the Vega-Lite version.  Both Vega-Lite v4 and Vega-Lite v5 use Vega v5.)

Essentially all I did was make some changes from `VEGALITE4_MIME_TYPE` to `VEGALITE5_MIME_TYPE`.  The currently released version of Altair uses Vega-Lite v4, but the next release will use Vega-Lite v5.  The only test I have done so far was to install the latest version of Altair `pip install git+git://github.com/altair-viz/altair.git` and then check that a chart can render offline in JupyterLab by following [this portion](https://altair-viz.github.io/user_guide/display_frontends.html#displaying-in-jupyterlab) of the Altair documentation.  (The referenced installation step `jupyter labextension install @jupyterlab/vega5-extension` seems to be out of date and can be ignored.)

I would be very grateful for advice on where other changes should be made, or of what kind of issues might arise from this naive fix.  Thanks for your time!


packages/vega5-extension/test/utils.ts
```diff
@@ -14,23 +14,23 @@ interface ISize {
...
+  mimeType: VEGALITE5_MIME_TYPE,
   resolver: null,
   sanitizer: {
     sanitize: (s: string) => s
   }
 });
 
-export const VEGALITE4_SPEC: JSONObject = {
-  $schema: 'https://vega.github.io/schema/vega-lite/v4.json',
+export const VEGALITE5_SPEC: JSONObject = {
```
krassowski, line 32: The tests are now failing as `VEGALITE4_SPEC` is no longer here (https://github.com/jupyterlab/jupyterlab/actions/runs/4046843519/jobs/6960151323). You could either update the tests in `jupyterlab/packages/vega5-extension/test/renderer.spec.ts` or copy-paste the old spec adding a comment stating that v4 is deprecated (if that makes sense, I am not too familiar with vega extension).

krassowski:
Given that this is including breaking changes (removal of exported member) I am tagging this at 4.0. It would be appreciated if you could also include a brief entry in the [migration documentation](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/extension/extension_migration.rst).

krassowski:
The regression in the visual documentation tests is relevant. The icon associated with the renderer got lost:

| Before | With this PR as of now |
|---|---|
| interface-jupyterlab-expected | interface-jupyterlab-actual |

I think that this is because `documentWidgetFactoryOptions` needs updating too.

ChristopherDavisUCI:
Thank you for catching that and for explaining the difference!  I think the correct icon now shows up in the latest commit.

I will check out the migration documentation; thanks for pointing it out.

fcollonval (commented):
Thanks @ChristopherDavisUCI 

> The CI failures are not related.

fcollonval (approved):

domoritz:
> My understanding is that some charts might look different when rendered with Vega-Lite 5 instead of Vega-Lite 4, but they should at least still render and not raise errors.

Yes, version updates don't guarantee that we don't render the chart differently but it should show the same visualization in spirit. If that's not the case, I would consider that a regression we should fix.

ChristopherDavisUCI:
TLDR; I haven't found any breaking changes as a result of this PR.  In particular, I haven't found any Vega-Lite 4 charts which render differently in JupyterLab as a result of this PR.  If you have an idea of how to produce one, I'm happy to test it out!

I spent some time just now trying to find examples of Vega-Lite 4 charts which render differently because of this PR, but the two examples I tried (one specifying a `domain` and one specifying a `band` instead of `bandPosition`) seem to render the same, independent of whether I use the JupyterLab version of this PR, or the released JupyterLab version.  (I chose these two properties after reading the [Breaking Changes](https://github.com/vega/vega-lite/releases/tag/v5.0.0) section of the Vega-Lite 5 release notes.)

This is the code I used with `domain`:

```
import altair as alt
import pandas as pd

alt.renderers.enable('mimetype')

df = pd.DataFrame({"A": [28]})

alt.Chart(df).mark_bar().encode(
    x = alt.X("A", scale=alt.Scale(domain=[0,150]))
)
```
In both cases, the result rendered like this (notice that the domain ends exactly at `150`).
domain

This is the code I used with `band`:

```
import altair as alt
from vega_datasets import data

alt.renderers.enable('mimetype')

weather = data.seattle_weather.url

alt.Chart(weather).mark_line().encode(
    x=alt.X("date:T", timeUnit="month", band=0.5),
    y=alt.Y("mean(temp_max):Q")
)
```

In both cases, the result rendered like this (notice that the lines start at the beginning of the month, rather than in the middle).

band

My theory is that the vega-embed version is not changing as a result of this PR, and so charts are rendering the same.

If I update to the pre-release version of Altair (what's currently on GitHub) and change from `band` to `bandPosition` (which is only possible in Vega-Lite 5), then the chart renders as expected (with lines starting halfway through the month).
altair5

Just as a reality check, the main point of this PR is that this Vega-Lite 5 code will not work in the current version of JupyterLab.  Here is what I get when I try.
altair5b

domoritz:
If you want an example that renders differently, use [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoDSACCikEzQAmABwBfBSbTnL11GLEOAuk5DIAE4A1migbDhIUHQAnmgADJRiSpgxOHCuTEiBIP5wslBsyjSyZGEgAB7lAGY0cILKrigpaRnosmwIJTq5SnGaILX1jejWLemuAI4MSLKYdGo0pLkODkA), which I think should render two bars on top of each other in z in Vega-Lite 4 and stacked bars in Vega-Lite 5. @kanitw may have some other ideas.

ChristopherDavisUCI:
TLDR; in terms of how Altair behaves in JupyterLab, I haven't found any issues.  From that perspective, I think this PR is able to be merged.  The comments above and below are trying to find examples of charts which render differently as a result of this PR.

Thanks again @domoritz!  I think the issue is I don't know how to force charts in JupyterLab (whether the release version or the version of this PR) to display in the Vega-Lite 4 style.  For the example you suggested, the chart displays like the Vega-Lite 5 style (with the bars stacked instead of layered on top of each other).

This is the code I used:
```
import altair as alt
import pandas as pd

alt.renderers.enable('mimetype')

df = pd.DataFrame({"a": ["A", "A"], "b": [28, 55]})

alt.Chart(df).mark_bar(opacity=0.5).encode(
    x = "a",
    y = "b"
)
```
And this is how the result looks:
vl-bars
I don't know if something like a cached version of vega-embed is getting used... I have tried in different browsers and I've tried clearing the outputs of JupyterLab before opening the notebooks, but it's very possible there is some step I'm missing.

I don't know of an analogue in JupyterLab of the `require("vega-lite@4")` that @domoritz used, although I would have thought that Vega-Lite 4 got used automatically in the released version of JupyterLab (as opposed to this PR which updates to Vega-Lite 5).

domoritz:
Huh, strange. Let me know if you want me to dig into this a bit to see why the chart renders as in Vega-Lite 5 even though we expect that the Vega-Lite 4 library is used.

mattijn:
This is expected behavior and the relevant scenario was discussed in this PR: https://github.com/altair-viz/altair/pull/2530#issuecomment-991792530.
That particular PR was put in place since _"Jupyter notebooks with previous Vega-Lite versions already loaded fail to show Altair charts"_ (https://github.com/altair-viz/altair/issues/2520#issuecomment-981128345), as was discussed long-windedly in that issue.

fcollonval:
Thanks for the discussion. Regarding the vega-lite version shipped with JupyterLab, I earthed that PR: https://github.com/jupyterlab/jupyterlab/pull/11179 in which vega-lite was upgrading from 4 to 5 for JupyterLab 3.2.0.

@ChristopherDavisUCI the version shipped within JupyterLab is a consequence of the following version constrain:

https://github.com/jupyterlab/jupyterlab/blob/0db893c74d9171d9516a3b1c3f626704d578805a/packages/vega5-extension/package.json#L43-L45

And the exact version is the one define in `yarn.lock`:

https://github.com/jupyterlab/jupyterlab/blob/0db893c74d9171d9516a3b1c3f626704d578805a/yarn.lock#L12382-L12383


Thanks all for the tests.

> Merging as CI failures are not related

joelostblom:
We are getting close to releasing v5 of Altair and I wonder if it would it make sense to backport this PR to support the latest 3.x releases in JupyterLab? To avoid confusion from people that install the new version of Altair, but haven't updated their Jupyterlab yet. As I understand the discussion above, this PR shouldn't break/change anything for those using Altair 4; they can still make use of the mimetype renderer as before.

Issue #12223, opened 2022-03-16
Controls to expand cell toolbars that overlap cell contents


jweill-aws:
### Problem

In the cell toolbar (#9733), we make the toolbar invisible if it would overlap the code or rendered markdown in the cell. Add a control to expand the cell toolbar either from the top or from the side when it has been hidden due to overlap.

### Proposed Solution

Design concept from @krassowski : 

![Toolbar collapsed to an expander arrow](https://user-images.githubusercontent.com/5832902/158484181-5db34699-fe65-4e0d-ab4a-3feff4bfc908.png)

Alternatively, roll the cell toolbar up and let the user click a thin expander at the top of the cell.

### Additional context

This is an enhancement to the cell toolbar, whose code was added in PR #12028.

firai:
I think having a single button to expand the cell toolbar would an intuitive way to deal with hiding the cell toolbar. As a minor implementation detail, the cell scroll box might need to be expanded to the right with some whitespace to allow the last contents on the first line to be scrolled to the left of this button, which I understand is the reason why the cell toolbar is hidden in the first place. Alternatively, the cell scrolling box could be expanded with the full width of the full cell toolbar, which would avoid the need to hide the toolbar in the first place.

Issue #13115, opened 2022-09-22
Cell contents overlap toolbar at narrow width


kulsoomzahra:
## [#13059](https://github.com/jupyterlab/jupyterlab/pull/13059) PR fixes :
Preserves the horizontal layout of the cell toolbar at 400% zoom

![jp](https://user-images.githubusercontent.com/46336830/190227485-cf04e68d-9063-49ec-a146-5c0b07141464.png)

## What it doesn't fix:
If the contents of the cell overlap with that of the toolbar, like:

![overlap](https://user-images.githubusercontent.com/46336830/190227106-1539e015-0c83-4343-a699-e1a962ab696a.png)

### Steps to reproduce
- Apply PR #13059.
- Open JupyterLab.
- Open a notebook that has code cells in it.
- Squeeze the width of the browser window until it’s below 760 px (the breakpoint)
- The contents of cell will overlap with the cell toolbar
(Previously, before PR #13059, following the above steps led to the cell toolbar running vertically down the cell.)

### Proposed Solution
A possible solution to this issue might be:
- Giving a solid background to the toolbar with/without borders
- Offsetting/pushing the toolbar up
- Adding more white space between cells 

(similar to what Google Colab does)

firai:
A couple of other proposed solutions to this issue, which I applies for both regular zoom and increased zoom: #12223

krassowski:
Another edge case is the collapsed cells; since I think that it is a bug (cell toolbar gets hidden for non-collapsed cells, but not for collapsed cells) I opened: https://github.com/jupyterlab/jupyterlab/issues/13254

krassowski:
@JasonWeill do you suggest that we should pause with review of #13896 until we release 4.1?

andrii-i:
@krassowski @JasonWeill PR #13896 solves issue #12523 so I would recommend to either close both PR and an issue or merge a PR (then continue work on cell toolbar mobile experience within this issue)

krassowski:
I cannot get into this state, I believe it was fixed.

PR #13896, opened 2023-01-31
Hide the cell toolbar on mobile / screens less than 760px wide to prevent cell obstruction

### LINKED ISSUE #12523: Hide the cell toolbar on mobile
## Description

The cell toolbar is still displayed on mobile devices and looks a bit off.

## Reproduce

Using JupyterLab 3.4.0 with this gist on Binder: https://gist.github.com/jtpio/dc217fb83255f54bf826a3804f5a0d7c

https://user-images.githubusercontent.com/591645/166635951-f9545ec5-cb32-437d-87c0-da2bded14e0a.mp4

## Expected behavior

The toolbar should be hidden when the viewport is getting smaller.

## Context

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/12028

andrii-i:
## References

As demonstrated in #12523, cell toolbar in its current form is obstructive on mobile: when there is not enough screen width cell toolbar gets in the way, people click cell toolbar by accident, sometimes cell toolbar overlays over the content of the cell. 

This PR would solve the problem by hiding the cell toolbar if viewport is less than 760 px wide. As a next step we should add option for people to still use a cell toolbar on mobile, for example a button that would open/close the cell toolbar as described in #12223.
 
- Fixes #12523 

## Code changes

Change `@media only screen and (max-width: 760px)` override for cell toolbar in `base.css` to `display: none`

## User-facing changes

On screens / viewports less than 760 px wide cell toolbar would be hidden. 

Before (courtesy of @jtpio ):
demo

After:
Kapture 2023-01-30 at 17 01 59

None


fcollonval:

fcollonval (commented):
Thanks @andrii-i 

One integration test requires a snapshot update. But the others may be flaky.

andrii-i:
@fcollonval thank you. With updated snapshots all integration UI tests pass. 

Other 2 failing tests ("Linux Tests / Linux (lint, 3.11) (pull_request)" and "Linux JS Tests / JS (js-debugger) (pull_request)") errors are not related to cell toolbar css, I've also seen them being flakey on other PRs. Could you please give these 2 tests a look? I can create issues to track them being flakey if it's they are problematic (looked and there are none at the moment).

krassowski (dismissed):
Thank you for pointing this out @andrii-i, I've opened issues for the failing jobs:
- https://github.com/jupyterlab/jupyterlab/issues/13903
- https://github.com/jupyterlab/jupyterlab/issues/13904

As for the PR:
- I think that the behaviour of moving the toolbar above the cell is a good one and we should keep it; it looks like it was a consensus in https://github.com/jupyterlab/jupyterlab/pull/13059 but I am ok with changing it if there were a new consensus about going in the other direction.
- I agree that once there is not enough space for toolbar at all it should be collapsed into a single button (and for now could be hidden), but I believe these should be two different thresholds (one for moving toolbar above and one for hiding it).
- The use of `media screen` query is wrong as it will not hide the toolbar if two or more notebooks are open side-by-side (even though the space would be taken up by them); we should consider switching to [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries). The only problem with `@container` queries is that these are pretty recent (only available since the last Firefox release) with current support of [76.93%](https://caniuse.com/css-container-queries) but since the 4.0 release is some time away I think we should be ok to include them (but not backport)

andrii-i:
@krassowski Thank you for taking a look at the failing tests and for the comments.

> * I think that the behaviour of moving the toolbar above the cell is a good one and we should keep it; it looks like it was a consensus in https://github.com/jupyterlab/jupyterlab/pull/13059 but I am ok with changing it if there were a new consensus about going in the other direction.
>
> * I agree that once there is not enough space for toolbar at all it should be collapsed into a single button (and for now could be hidden), but I believe these should be two different thresholds (one for moving toolbar above and one for hiding it).

As mentioned in PR description, this PR fixes open 4.0.0 bug #12523  while adding a button should be a follow-up. I would be happy to work on the follow-up within this PR or within #13115 . If #12523 and code that fixes it are not relevant, let's close it.

krassowski:
Sorry if I missed this, were there any changes since my last review? I see a force push which makes reviews more difficult.

andrii-i:
> Sorry if I missed this, were there any changes since my last review? I see a force push which makes reviews more difficult.

No problems, please see my [response](https://github.com/jupyterlab/jupyterlab/pull/13896#issuecomment-1412796775) to your comment. I'm a bit confused if this PR and issue that it solves should be closed if they are not relevant anymore

krassowski:
My concern is that this removes toolbar for windows below 760px without a way to show it, regardless of whether they are on mobile or not (and one could argue that this is not relevant but instead the touchscreen presence is more important).

Also, taking a closer look at the code, it actually is a no-op for most cases because the overlap detection logic implemented in JS already hides the toolbar if we remove the old rules defined within `@media only screen and (max-width: 760px)`.

My other comment is that if we were to implement a width-based heuristic, the width of the notebook would matter more than the width of the screen (hence the suggestion on using container queries).

andrii-i:
These are really good points. Thank you for explaining them for me. Here's an issue to track problematic use of media queries:
- #13911

krassowski:
Thank you for opening the issue.

andrii-i:
I agree with implementation improvement points raised in review by @krassowski . Also that ideally we would want to provide an option to show/hide the cell toolbar on demand which would make this implementation and its improvement not relevant. 

Based on this, I'll see if I can make an option to show/hide the cell toolbar on demand happen in the next 1-2 weeks. If that won't work out, we can come back to this implementation and improve/fix it before merging.

jtpio (approved):
Thanks @andrii-i this looks good!

jtpio:
According to the discussion above it looks like we could proceed with merging this and address the remaining points in follow-up PRs.

andrii-i:
@jtpio merging this now and and addressing the remaining points in follow-up PRs is also a valid strategy. 

Merged master into the branch so it's ready to merge when green. I can't merge myself as I don't have an authorization.

fcollonval (approved):
Thanks @andrii-i

andrii-i:
Thanks @fcollonval

PR #13898, opened 2023-01-31
Introduce disabled states for the cell move up/move down icons

### LINKED ISSUE #13502: Introduce disabled states for the cell move up/move down icons in the toolbar
### Problem

The icons on the cell toolbar to move a cell up or down currently have the following states:
- enabled
- hover

However, the following cases should be visually different to users:
- when a cell is at the top of the notebook, the "move upwards" icon should be disabled.
- when a cell is at the bottom of the notebook, the "move downwards" should be disabled.

Although it can seem obvious that these actions are not permitted in such cases, it is important to provide visual cues to users to avoid the frustration of trying to click on a button that seems enabled, only to notice by error/trial that it doesn't work.

### Proposed Solution
I suggest to add a 3rd state: the disabled state for the icons "move cell below" and "move cell above", as illustrated below:

Add disabled state to icons

### Additional context
This issue is a result from a usability study conducted in October and November 2022 on beginner users of JupyterLab, aiming to test a collaborative workflow. Feel free to ask me for the full report.

HaudinFlorence:
Introduce disabled states for the cell move up/move down icons in the toolbars of the notebook cells.
The Move Up (resp. Move Down) icon should be enabled for the first cell (resp. last cell) when it is the active cell.
Both icons are active for any other active cell.
capture

## References
Fixes #13502

## Code changes

Proposed modified versions of `isEnabled` functions for `moveUp` and `moveDown` commands in `packages/notebook-extension/src/index.ts`


krassowski:
bot please update snapshots

packages/notebook-extension/src/index.ts
```diff
@@ -2512,7 +2518,14 @@ function addCommands(
...
       }
     },
-    isEnabled,
+    isEnabled: args => {
+      const current = getCurrent(tracker, shell, args);
+      if (!current || !current.content.model) {
+        return false;
+      }
+      const length = current.content.model?.cells.length;
+      return current!.content.activeCellIndex < length - 1;
```
krassowski, line 2527:
```suggestion
      return current.content.activeCellIndex < length - 1;
```

packages/notebook-extension/src/index.ts
```diff
@@ -2496,7 +2496,13 @@ function addCommands(
         );
       }
     },
-    isEnabled,
+    isEnabled: args => {
+      const current = getCurrent(tracker, shell, args);
+      if (!current) {
+        return false;
+      }
+      return current!.content.activeCellIndex >= 1;
```
krassowski, line 2504:
```suggestion
      return current.content.activeCellIndex >= 1;
```

packages/notebook-extension/src/index.ts
```diff
@@ -2512,7 +2518,14 @@ function addCommands(
         );
       }
     },
-    isEnabled,
+    isEnabled: args => {
+      const current = getCurrent(tracker, shell, args);
+      if (!current || !current.content.model) {
+        return false;
+      }
+      const length = current.content.model?.cells.length;
```
krassowski, line 2526: Since you are checking for model higher up, you should be able to remove `?` here too I think. This might require you to assign `current.content.model` onto a variable before the check (or may not).

krassowski (commented):
Thanks for working on this!

krassowski:
Regarding the last commit, we already have notebooks in tests (see `galata` directory/documentation on visual tests) and your changes are showing up on visual regression tests, so we should be good without adding another notebook.

bot please update snapshots

HaudinFlorence:
bot please update snapshots

packages/notebook-extension/src/index.ts
HaudinFlorence: Thanks for your review. I have tried to address most of your comments. Not sure to understand what you meant about assigning `current.content.model` onto a variable before the check.

krassowski:
This looks good:

Screenshot from 2023-01-31 19-09-01

This one might be just a flake (focus went to a wrong location and should be reverted):

Screenshot from 2023-01-31 19-08-48

Screenshots updates for running kernels sidebar need to be reverted too. I think that you will get less of flaky tests for these if you merge in (/rebase on) master branch.

krassowski:
This one looks like a bug in the logic (there is a markdown cell above the raw cell), am I right?

Screenshot from 2023-02-01 10-10-19

This oculd be happening because the state of the command is not refreshed. We may need to invoke `notifyCommandChanged` manually on specific actions if that is the case (I have not looked in detail).

HaudinFlorence:
> This one looks like a bug in the logic (there is a markdown cell above the raw cell), am I right?
> 
> Screenshot from 2023-02-01 10-10-19
> 
> This could be happening because the state of the command is not refreshed. We may need to invoke `notifyCommandChanged` manually on specific actions if that is the case (I have not looked in detail).

I tried to have a Markdown cell as first cell followed by a Raw one. It seems to be ok
Screenshot from 2023-02-01 11-38-17

krassowski:
I don't think that the type of the cell matters but - as in the test that is failing - the fact of moving/pasting cells and the buttons not updating their state after another cell is pasted e.g. above

HaudinFlorence:
> I don't think that the type of the cell matters but - as in the test that is failing - the fact of moving/pasting cells and the buttons not updating their state after another cell is pasted e.g. above

At least I tried to play a little with this different moving/ copy paste operations locally and did not observe any not updating behavior, if that's the point.

krassowski:
A reliable reproducer is deleting the first cell (<kbd>D</kbd> <kbd>D</kbd>) and the undoing it (<kbd>Z</kbd>):

undo-reproducer

HaudinFlorence:
> A reliable reproducer is deleting the first cell (D D) and then undoing it (Z):
> 
> undo-reproducer [ undo-reproducer ](https://user-images.githubusercontent.com/5832902/216027314-b9325992-5bd6-46b9-bf05-34c4cb526f8e.gif) [ ](https://user-images.githubusercontent.com/5832902/216027314-b9325992-5bd6-46b9-bf05-34c4cb526f8e.gif)

You're right there seems to be a bug here in the update. It doesn't solve it but if you move back to the first cell and go back to the second one, the icons are ok again.

HaudinFlorence:
> This looks good:
> 
> Screenshot from 2023-01-31 19-09-01
> 
> This one might be just a flake (focus went to a wrong location and should be reverted):
> 
> Screenshot from 2023-01-31 19-08-48
> 
> Screenshots updates for running kernels sidebar need to be reverted too. I think that you will get less of flaky tests for these if Sorry, I am not sure to understand what screenshots you refer to as for running kernels sidebar ?

krassowski:
> Sorry, I am not sure to understand to what screenshots you refer to as for running kernels sidebard.

Prior to re-basing there were spurious screenshot updates due to flaky test for running kernels sidebar. When you rebased you included the latest PR which fixed it:

Screenshot from 2023-02-01 13-48-33

So there is nothing more to do with regard to this one, all is good.

HaudinFlorence:
> > Sorry, I am not sure to understand to what screenshots you refer to as for running kernels sidebard.
> 
> Prior to re-basing there were spurious screenshot updates due to flaky test for running kernels sidebar. When you rebased you included the latest PR which fixed it:
> 
> Screenshot from 2023-02-01 13-48-33
> 
> So there is nothing more to do with regard to this one, all is good.

Ok. Thanks. So just to sum up the screenshot that need to be reverted (replaced by the version of master) is : galata/test/documentation/general.test.ts-snapshots/file-formats-html-display-documentation-linux.png. And the one showing a bug is galata/test/jupyterlab/notebook-edit.test.ts-snapshots/copy-paste-cell-jupyterlab-linux.png.

packages/notebook-extension/src/index.ts
```diff
@@ -2501,7 +2501,13 @@ function addCommands(
         );
       }
     },
-    isEnabled,
+    isEnabled: args => {
+      const current = getCurrent(tracker, shell, args);
+      if (!current) {
+        return false;
+      }
+      return current.content.activeCellIndex >= 1;
```
fcollonval, line 2509: As mentioned by Mike, the status `isEnabled` needs to be updated if the active cell changes to fix the bug seen when restoring the first cell for example.

You can achieve this by triggering `commands.notifyCommandChanged(<command id>)`.

Now the trouble is that this should be triggered for all notebooks opened. Luckily, this is already done for you if you listen to `tracker.activeCellChanged`.

So the final code will look like,

```ts
tracker.activeCellChanged.connect(() => {
  commands.notifyCommandChanged(CommandIDs.moveDown);
  commands.notifyCommandChanged(CommandIDs.moveUp);
})
```

fcollonval (changes requested):
Thanks @HaudinFlorence 

I provided some pointers following the review of Mike. And you have to remove the `test_notebook.ipynb` that got committed by mistake.

packages/notebook-extension/src/index.ts
HaudinFlorence: Thanks for the precisions @fcollonval. I tried to take into account the last comment in commit 
fdb0baa. In my local tests, the tentative fix did not solve the issue when removing first cell and restoring it.

fcollonval: Good point @krassowski . 

The notebook signal documentation states:

https://github.com/jupyterlab/jupyterlab/blob/013b6ab817b929f5eaebaec7b282d0d0f52527c1/packages/notebook/src/widget.ts#L1237-L1243

So for me the current code is erroneous:

https://github.com/jupyterlab/jupyterlab/blob/013b6ab817b929f5eaebaec7b282d0d0f52527c1/packages/notebook/src/widget.ts#L1321-L1327

We should trigger `this._activeCellChanged.emit(cell);` if the new active index changes even if the widget is the same.

What do you think?

HaudinFlorence: @krassowski @krassowski Thanks for your feedbacks. I have tried to modify `widget.ts` to cover the case "the new active index changes" in order to fit with the explaining notes relative to the `activeCellChanged`. This seems to work on the previous simple 2-cell test suggested.

krassowski: @fcollonval the way I read the documentation the current implementation is not erroneous. I think that the second part "or cell at the active index changing" refers to a case when say a code cell is converted to e.g. markdown cell (so the cell at the active index _changes_).

It feels to me like `activeCellChanged` signal is doing what I would expect it to do. We could have another signal, say `cellPositionChanged` to deal with situations like this one.

@HaudinFlorence thanks for verifying that this would fix it. We may want to think a little more on what is the clean way to approach it (while I am proposing adding a new signal it is just brainstorming - we are usually reluctant to extend API without taking more time to think about it as it is difficult to change later).

What do you both think about using `notebook.model.cells.changed` signal instead?

HaudinFlorence: @krassowski Thanks for your feedback. I can try to use `notebook.model.cells.changed`. I will wait for @fcollonval's opinion before trying.

packages/notebook/src/widget.ts
```diff
@@ -1319,12 +1319,13 @@ export class Notebook extends StaticNotebook {
...
     this._activeCellIndex = newValue;
     const cell = this.widgets[newValue] ?? null;
-    if (cell !== this._activeCell) {
+    if (cell !== this._activeCell || newValue != oldValue) {
       // Post an update request.
       this.update();
       this._activeCell = cell;
       this._activeCellChanged.emit(cell);
     }
+
```
fcollonval, line 1328: Could you change this to avoid updating the UI if the active cell widget did not change?

```suggestion
    const cellChanged = cell !== this._activeCell;
    if (cellChanged) {
      // Post an update request.
      this.update();
      this._activeCell = cell;
    }
    
    if (cellChanged || newValue != oldValue) {
      this._activeCellChanged.emit(cell);
    }

```

fcollonval (changes requested):

packages/notebook-extension/src/index.ts
fcollonval: @krassowski I was actually referring to the first clause _the active index changing_ :wink:. The problematic case highlights a problem with that clause as restoring a delete cell is not changing the active cell but is changing the active index. No emitting the signal is such a case feels wrong.

Pinging other devs for input on that signal problem: @jasongrout @afshin @blink1073 @jtpio 

Another edge case (not related to this PR), prior to dropping modelDB, moving a cell was not triggering that signal (as the widget was not changed). But now it does as a move is add + delete. But when the move feature lands in Yjs/Ypy, we will fallback to the previous behavior. So I would rather emitted more often but ensure better consistent (more independent of the implementation).

krassowski: I am happy to move forward here (after making sure we don't emit spurious update as in your suggestion).

packages/notebook/src/widget.ts
krassowski: This screenshot looks fine to me. The problem previously was that both arrows were greyed out and on your screenshot only one is greyed out. Am I missing something?

HaudinFlorence: Yes you are right @krassowski . Sorry for the confusion.

fcollonval:
bot please update snapshots

fcollonval:
bot please update snapshots

fcollonval:
Kicking the CI

fcollonval (approved):

fcollonval:
bot please update snapshots

fcollonval:

fcollonval:
-

packages/notebook-extension/src/index.ts
```diff
@@ -2802,7 +2806,13 @@ function addCommands(
         );
       }
     },
-    isEnabled,
+    isEnabled: args => {
+      const current = getCurrent(tracker, shell, args);
```
fcollonval, line 2810:
```suggestion
      const current = getCurrent(tracker, shell, { ...args, activate: false });
```

packages/notebook-extension/src/index.ts
```diff
@@ -2833,7 +2843,15 @@ function addCommands(
...
```
fcollonval, line 2847:
```suggestion
      const current = getCurrent(tracker, shell, { ...args, activate: false });
```

fcollonval:
bot please update snapshots

PR #13907, opened 2023-02-02
Move RTC packages to an extension

### LINKED ISSUE #13726: Move collaboration features to a separate package
### Problem

Once the implementation of RTC through IDrive is done. We can extract every specific RTC UI elements to a separate extension. So RTC is a opt-in through extension installation. And we are able to keep iterating faster on the UI elements.

The elements that need to be in core for RTC are:
- Document model stored in RTC compatible object (= Yjs document)
- Getting user identity and permission (Jupyter Server Identity & Authorization API)

hbcarlos:
Move RTC packages to an extension

### TODO on a follow-up PR:
- [x] Remove RTC documentation. The documentation of the new extension is not set up yet, we can not link there yet.

## References
Fixes #13726

## Code changes
Removes the packages `collaboration`, `collaboration-extension`, and `docprovider`.
Uses the `--collaborative` flag to check whether the `jupyterlab_rtc` extension is installed.

## User-facing changes
To get Real-time collaboration in JupyterLab users will need to install the extension `jupyterlab-rtc` (the extension name is under discussion).


hbcarlos:
Thanks, @krassowski! That makes sense, I'll implement it.

jupyterlab/labapp.py
```diff
@@ -831,6 +831,26 @@ def initialize_handlers(self):  # noqa
     def initialize(self, argv=None):
         """Subclass because the ExtensionApp.initialize() method does not take arguments"""
         super().initialize()
+        if self.collaborative:
+            try:
+                import jupyterlab_rtc  # noqa
+            except Exception:
```
fcollonval, line 837: let's be explicit

```suggestion
            except ImportError:
```

jupyterlab/labapp.py
```diff
@@ -831,6 +831,26 @@ def initialize_handlers(self):  # noqa
...
+                    "To enable real-time collaboration, you must install the extension jupyterlab_rtc."
+                )
+                self.log.critical("You can install it using pip for example:")
+                self.log.critical("")
+                self.log.critical("\tpython -m pip install jupyterlab_rtc")
+                self.log.critical("")
+                self.log.critical(
+                    "*********************************************************************************"
+                )
+                self.log.critical("")
```
fcollonval, line 852: You should never do that because log messages may be sent to any kind of handlers; some may be for example a log aggregator in which you want one entry per event.

For the same reason you should avoid including any kind of format.

```suggestion
                self.log.critical("""
To enable real-time collaboration, you must install the extension `jupyterlab_rtc`.
You can install it using pip for example:

  python -m pip install jupyterlab_rtc
""")
```

pyproject.toml
```diff
@@ -44,7 +44,6 @@ dependencies = [
     "jupyter_server>=2.0.1,<3",
     "jupyterlab_server>=2.19.0,<3",
     "jupyter_ydoc>=0.3.0,<0.4.0",
```
fcollonval, line 46: This has to be removed too as we bundle the JavaScript package we need but don't use the Python package.

```suggestion
```

packages/docregistry/src/context.ts
```diff
@@ -67,7 +66,7 @@ export class Context<
     this._model = this._factory.createNew({
       languagePreference: lang,
       sharedModel,
-      collaborationEnabled: PageConfig.getOption('collaborative') === 'true'
+      collaborationEnabled: sharedFactory === null ? false : true
```
fcollonval, line 69: I would keep the usage of `PageConfig.getOption('collaborative') === 'true'`. Mainly because providing a shared model factory does not ensure live collaboration is enabled.

fcollonval (changes requested):
Thanks @hbcarlos

I have some comment. Could you also make the change to the documentation as part of this PR?

packages/docregistry/src/context.ts
hbcarlos: I am not sure it is a good idea to keep the `PageConfig.getOption('collaborative') === 'true'`. That means we will have to set that attribute from the extension, which will be global. We could have two drives installed, one collaborative and the other not. Two components can influence whether a document is collaborative:

1- A provider that supports RTC (the YDrive).
2- And the DocumentModel is designed to support RTC.

We know the drive supports collaboration when it returns a shared model for a given path:
```
const sharedFactory = this._manager.contents.getSharedModelFactory(
      this._path
);
```

And we know the DocumentModel supports collaboration by checking the `collaborative` property in its factory, `DocumentModelFactory.collaborative === true`.

To activate RTC for a given path, we check whether both components support RTC and set the property `DocumentModel.collaborative` for each document.

Maybe we should make the IDrive more explicit by adding a `collaborative` property instead of assuming the model is collaborative when it returns a shared model.

fcollonval: The multiple drive cases is a good argument to drop `PageConfig`. But still having a `sharedFactory` is not, we know it is what we will use for `jupyterlab_rtc`. But we should not make assumption on what other may do with it. So I like to go in the direction of your proposal. But instead of setting it at the drive level that is deeply hidden, I would either 
- extend `ISharedFactory` to provide a `collaborative` flag
- move the `collaborative` attribute from `DocumentRegistry.IModel` to `ISharedDocument` (this would be valid as `ISharedFactory` receive all information to create the `ISharedDocument`).

I have a small preference for the second approach. What do you think?

hbcarlos: I like it!

hbcarlos: I changed my mind, lets's go with the first option. It will be easier to maintain.

fcollonval (approved):
Thanks @hbcarlos

packages/docregistry/src/context.ts
fcollonval: We went for the first option because in the case of the second, we were not happy about adding a flag in `@jupyter/ydoc` that would have been specific to the JupyterLab infrastructure.

PR #13909, opened 2023-02-02
Improve Galata

fcollonval:
## References

Extract improvement of Galata from https://github.com/jupyterlab/jupyterlab/pull/13240

## Code changes

The major change in Galata is the switch from runtime injection of the helper to a regular JupyterLab extension. This allows to get the helpers in the same namespace as JupyterLab - hence it gives us access to some non-token based elements like `Dialog`s.

Building on that, this introduces the ability to register listener to dialog and notifications events to make tests more robust; e.g. sometimes a test can be flaky because randomly some dialogs get in the way; with the new listener you could dismissed any random dialogs that may appear.

Adds new fixture `mockUser` to mock the user provided by the identity endpoint.

It renames `window.galataip` to `window.galata` as the later _ip_ is not easy to comprehend (it stands for in-page).

As the in-page helpers are now part of an extension, the server setup needs to evolve to add that extra extension. To ease future evolution of the server setup, a new helper `jupyterlab.galata.configure_jupyter_server` is provided.

## User-facing changes

None

Remove `JupyterLab.notificationManager` trick used up to now to get access to notifications.
`window.galataip` is deprecated in favor of `window.galata`


galata/README.md
```diff
@@ -81,6 +76,8 @@ Then start the server with:
 jupyter lab --config jupyter_server_test_config.py
 ```
 
+> If you need to customize the set up for galata, you can look at the `configure_jupyter_server <https://github.com/jupyterlab/jupyterlab/tree/master/jupyterlab/galata/__init__.py>`\_ definition.
```
fcollonval, line 79:
```suggestion
> If you need to customize the set up for galata, you can look at the [`configure_jupyter_server`](https://github.com/jupyterlab/jupyterlab/tree/master/jupyterlab/galata/__init__.py) definition.
```

fcollonval:
`check-links` is expected to fail due to a new link to a to-be-added file.

jtpio:
Thanks this looks good :+1: 

The lint CI failure looks relevant though:

image

fcollonval:
Thanks for the review @jtpio I fix the linter errors.

jtpio:
Restarting the check release check, not sure it's relevant: 

image

jtpio:
Looks like the check is now passing:

image


Maybe we'll want to get https://github.com/jupyterlab/jupyterlab/pull/13931 in first, to make sure the UI tests checks pass and this PR doesn't introduce regressions.

fcollonval:
Merging to move forward - I keep working on #13931

PR #13910, opened 2023-02-02
Rename `@jupyterlab/completer-extension:tracker` to `@jupyterlab/completer-extension:manager`

### LINKED ISSUE #13669: Completer manager plugin is named `:tracker` but probably should be  `:manager`
I noticed that the completer-extension manager plugin is named as `:tracker`:

https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/completer-extension/src/index.ts#L24

I would think that `:manager` is more appropriate here as it is not a tracker. Is this a copy-paste error, or was this intended? On 3.x branch we did have a `:manager` too (but with a different API).

CC @trungleduc

fcollonval:
## References

Fixes #13669

## Code changes

Rename `@jupyterlab/completer-extension:tracker` to `@jupyterlab/completer-extension:manager`

> I did not update the migration note as this was the name in 3.6.x

## User-facing changes

None

None


krassowski (approved):
Thank you!

Issue #13903, opened 2023-02-01
js-debugger tests are failing on master


krassowski:
```
FAIL  test/session.spec.ts (63.244 s)
```

I see several errors/warnings in logs, not sure which are the source of the problem:

<details>

<summary>An update to FramesComponent inside a test was not wrapped in act</summary>

```
    console.error
      Warning: An update to FramesComponent inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at model (/home/runner/work/jupyterlab/jupyterlab/packages/debugger/src/panels/callstack/body.tsx:41:3)

      53 |   useEffect(() => {
      54 |     const updateFrames = (): void => {
    > 55 |       setSelected(model.frame);
         |       ^
      56 |       setFrames(model.frames);
      57 |     };
      58 |     model.framesChanged.connect(updateFrames);

      at printWarning (../../node_modules/react-dom/cjs/react-dom.development.js:67:30)
      at error (../../node_modules/react-dom/cjs/react-dom.development.js:43:5)
      at warnIfNotCurrentlyActingUpdatesInDEV (../../node_modules/react-dom/cjs/react-dom.development.js:24064:9)
      at dispatchAction (../../node_modules/react-dom/cjs/react-dom.development.js:16135:9)
      at setSelected (src/panels/callstack/body.tsx:55:7)
      at invokeSlot (../../node_modules/@lumino/signaling/src/index.ts:616:12)
      at Object.emit (../../node_modules/@lumino/signaling/src/index.ts:538:9)
      at Signal.emit (../../node_modules/@lumino/signaling/src/index.ts:179:13)
      at CallstackModel.emit [as frames] (src/panels/callstack/model.ts:34:25)
      at DebuggerService._getAllFrames (src/service.ts:781:33)
      at DebuggerService.restoreState (src/service.ts:423:7)
      at Object.<anonymous> (test/debugger.spec.ts:334:7)
```

</details>

<details>

<summary>An update to BreakpointsComponent inside a test was not wrapped in act</summary>

```
    console.error
      Warning: An update to BreakpointsComponent inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
[...]

PR #13915, opened 2023-02-03
Notify command explicitly

### LINKED ISSUE #13684: Debugger invokes `notifyCommandChanged` force-updating commands after tab switch
## Description

The debugger extension plugins for notebook, console and file all invoke a full commands update, e.g.:

https://github.com/jupyterlab/jupyterlab/blob/edc294b06fe6607be80094dd590cb9f766ab83ff/packages/debugger-extension/src/index.ts#L138

resulting in a long task preventing fast consecutive tab switch.

## Reproduce

1. Open two notebooks
2. Record profile
3. Switch tabs
4. See a long task invoking React re-render as below:

Screenshot from 2022-12-29 17-02-45

## Expected behavior

There is no needless updates. Maybe something like;

https://github.com/jupyterlab/jupyterlab/blob/edc294b06fe6607be80094dd590cb9f766ab83ff/packages/toc-extension/src/index.ts#L303-L304

## Context

- JupyterLab version:  4.0.0a31

fcollonval:
## References

Fixes #13684

## Code changes

Call explicitly the commands of the extension and not all commands to avoid useless rerendering.

## User-facing changes

None

None

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -48,6 +48,12 @@ import { Session } from '@jupyterlab/services';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITranslator } from '@jupyterlab/translation';
 
+function notifyCommands(app: JupyterFrontEnd): void {
+  Object.values(Debugger.CommandIDs).forEach(command => {
+    app.commands.notifyCommandChanged(command);
+  });
+}
+
 /**
  * A plugin that provides visual debugging support for consoles.
  */
@@ -74,7 +80,7 @@ const consoles: JupyterFrontEndPlugin<void> = {
       const { sessionContext } = widget;
       await sessionContext.ready;
       await handler.updateContext(widget, sessionContext);
-      app.commands.notifyCommandChanged();
+      notifyCommands(app);
     };
 
     if (labShell) {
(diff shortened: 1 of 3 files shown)


krassowski (approved):
Thank you!

Ideally we would fix the debugger tests (#13903) before merging this one.

jtpio:
> Ideally we would fix the debugger tests (#13903) before merging this one.

Just rebased with the GitHub UI: 

image

To grab the `js-debugger` fix from https://github.com/jupyterlab/jupyterlab/pull/13941

jtpio:
The `js-debugger` check is now passing: 

image

jtpio (approved):
Thanks!

packages/debugger-extension/src/index.ts
```diff
@@ -48,6 +48,12 @@ import { Session } from '@jupyterlab/services';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITranslator } from '@jupyterlab/translation';
 
+function notifyCommands(app: JupyterFrontEnd): void {
```
jtpio, line 51: Ah this seems to be triggering commands regardless whether the plugin is available or activated on the page.

Spotted downstream in Notebook 7, which does not include the command to inspect variables (`@jupyterlab/debugger-extension:variables`):

image

jtpio: Opened https://github.com/jupyterlab/jupyterlab/issues/14023

Issue #13746, opened 2023-01-10
Upgrade json schema version


fcollonval:
### Problem

It could be interesting to bump the JSON Schema version used from draft04 to some more recent flavor (last proposal is [2020-12](https://json-schema.org/)).

The new version needs to be compatible with:

- JS [ajv](https://www.npmjs.com/package/ajv)
- JS [react-json-schema-form](https://www.npmjs.com/package/@rjsf/validator-ajv8)
- Python [jsonschema](https://pypi.org/project/jsonschema/)

The Python package seems to be the one falling behind in term of standard support.

bollwyvl:
> Python `jsonschema`

Things look mostly fine here: [Draft202012Validator](https://python-jsonschema.readthedocs.io/en/latest/api/jsonschema/validators/#jsonschema.validators.Draft202012Validator), with the README suggesting a key missing useful thing (`$vocabulary`) will be supported, presumably at some point within the `jsonschema@4.*` release cycle.

> JS [react-json-schema-form](https://www.npmjs.com/package/@rjsf/validator-ajv8)

Indeed: I've been waiting with baited breath to get a live [rjsf 5](https://github.com/rjsf-team/react-jsonschema-form/issues/3144#issuecomment-1382404859), which splits out the `ajv` dependency, so it can be upgraded. `ajv` is already a _huge_ dep, so should be moved wholesale to 8, if possible, for lab4.

Getting this hoisted as a shared package would be huge, as well as loading it as lazily as possible, as it is, well, _huge_.

> JS [ajv](https://www.npmjs.com/package/ajv)

Of note: even `ajv@8` _still_ gets flagged for `unsafe-eval`, by default, and there's really no way around it for the kinds of dynamic things jupyterlab does (e.g. translation). When addressing this systematically, it likely makes sense to offer a `IJSONSchema` plugin, such that one _could_ inject a "safe" parser... if it were to exist. It would _definitely_ be slower... even the `rust`-based one says _just go use `ajv`_... and WASM is usually not the _best_ place to go without a sound reason.

bollwyvl:
[rjsf 5 is out](https://github.com/rjsf-team/react-jsonschema-form/releases/tag/v5.0.0) :tada:

bollwyvl:
[...]

PR #13924, opened 2023-02-04
Update to ajv 8 and react-json-schema-form 5

bollwyvl:
## References

- starts #13746

## Code changes

- [x] add version 5 of `@rjsf/core`, `@rjsf/utils` and `@rjsf/validator-ajv8` 
- [x] update to `ajv8`
- [x] add explicit `version` property to plugin schema 
- [x] remove submit button with default `ui:options` rather than CSS (the `.rjsf` class is gone).
- [x] applies (and removes) TODOs about replacing home-grown types with RJSF ones 

## User-facing changes

- forms return somewhat more useful error messages

| `master` | #13924 |
|-|-| 
|  | 

- the `ajv` instances wasn't _strictly_ exposed, so this probably doesn't change much
- the `validator` prop on `FormComponent` is required
- the props of custom fields, etc. will change

## Feedback Sought

:bellhop_bell: @brichet @marthacryan @krassowski 

## Future Work
A follow-on PR will refactor the dependencies on `ajv 8` and `@rjsf/validator-ajv8` in a single `@jupyterlab/validator-ajv8`.

This will allow for cleanly replacing the _runtime_ JSON Schema-related dependencies served to the browser by disabling the default provider.

`@jupyterlab/builder` _could_ be changed (in another PR) to use a correct, but less-performant validator that isn't flagged for `unsafe-eval`, or somehow pre-compiles all the schema with `ajv-cli`.


packages/settingregistry/src/plugin-schema.json
```diff
@@ -102,6 +102,10 @@
     "jupyter.lab.transform": {
       "type": "boolean",
       "default": false
+    },
+    "version": {
+      "description": "The effective version of this schema",
```
fcollonval, line 107: For information, what do you expect as `version` specifier? A version number, the schema id URI,... ?

fcollonval (commented):
Thanks a lot for working on this @bollwyvl 

I have one question.

packages/settingregistry/src/plugin-schema.json
bollwyvl: Ah, that may be an error: this file itself has a `version`, which i believe was running afoul of the initial `strict` mode, which i eventually disabled everywhere, as it goes beyond (any of) the metaschemas.

I'll fix it up.

bollwyvl:
Thanks for the speedy review.

The follow-on work is proving quite tricky: moving the form validator out to a plugin/package in #13927 is no problem (aside from the CI complaints about installing not-on-npm packages), but trying to generalize the way settings (and other future consumers) use a number of `ajv`'s features (like merging `default` values in-place) is challenging to think about without having a specific implementation in mind.

Of the [recognized JS implementations](https://json-schema.org/implementations.html#validator-javascript), many appear unmaintained. Of the remaining, few have these capabilities, or the model of a "validator factory."

On the general topic of other implementations: all of this smells somewhat like security theater, as _oh noes! `unsafe-eval`!_ in full JupyterLab is kind of... silly. I'd much rather spend time getting a nice, reusable lumino form for extension authors working for 4, but the architectural stuff seems like it is important to at least consider.

bollwyvl:
Ubuntu's mirrors are having _A Day_, so i wouldn't sweat the [snapshot tests failing](https://github.com/jupyterlab/jupyterlab/actions/runs/4104378230/jobs/7079783049) at the moment. Not sure what the right approach is to get access to more mirrors.

brichet:
Thanks @bollwyvl, that looks good to me.

Do you think we could do without the `IMetadataSchema` described in *settingregistry* ? It seems that its properties already exist in `RJSFSchema`. 
It can also be done in a follow up PR, as I don't know the consequences of changing the `properties` property.

bollwyvl:
Yeah, on this PR was looking to keep stuff similar-ish when i could.

But it indeed needs a further specialization of `RJSFSchema` to capture having properties, and or any special sauce.

bollwyvl:
> `settingregistry`

Actually, to that point: I think we don't _want_ any _RJSF_ (or any _R_ for that matter) imports (even just `type` ones) in places we don't absolutely need them. By having a custom, documented type in the plumbing package, we are a little more able to keep stuff consistent if `rjsf6` or whatever comes along and changes things during the lab 4.x line.

This is part of my motivation for breaking some more of these things up, and providing more Lab-owned APIs than relying on R/JSF: as an extension author, relying on 

```ts
import {SchemaForm} from '@jupyterlab/ui-components'
...
activate: (app) => {
  //...
  const form = new SchemaForm({schema, instance});
  form.submitted.connect(() => app.commands.execute('some-command', form.instance))
}
``` 

...will require caring much less over time about the underlying packages, and getting to UI+validation quickly.

If an author wades into heavy customization, that's up to them! Indeed, I'd _really_ like to be able to wrap a lumino Widget into the registry  (i can _taste_ the datagrid as an `ArrayField`) but that's real far down the line, and I'm unlikely to get to it in the 4.0.0 window.

fcollonval (approved):
Thanks @bollwyvl 

As you mentioned, let's move on with the package upgrade with this PR. And leave further enhancement for follow up tasks.

PR #13927, opened 2023-02-05
Add `IFormValidator`

bollwyvl:
## References

- continues #13746

## Elevator Pitch

This starts the process of making the JupyterLab's JSON Schema validator replaceable. 

## Motivation

The choice of underlying validator library is now inextricably (but correctly) linked to the visual presentation of settings and notebook metadata.

Being able to swap to another, as-of-yet unspecified validator, is ultimately to avoid breakage when JupyterLab is hosted with the `unsafe-eval` CSP set (this can be done, e.g. by portal iframes, over which the application deployer and consumer may have no control). Choosing a validator is out of scope, and discussed in the _Future Work_ session below

And yes, all of this is a little silly when the user controls the application. But here we are.

## Code changes

- [x] add a new package, `@jupyterlab/validator-ajv8-extension`
  - [x] add a plugin `@jupyterlab/validator-ajv8-extension:form-validator`
    - [x] the definition of `IFormValidator` stays in `@jupyterlab/ui-components`
  - [x] consolidates imports of `@rjsf/validator-ajv8` into a single async import 
    - ...except one test

## User-facing changes

- n/a (beyond #13924)

- n/a (beyond #13924)

## Future Work

On a follow-on PR:

- [ ] add a new package, `@jupyterlab/validator`
  - [ ] consolidate all imports of `ajv`
  - [ ] move `ISchemaValidator` here
  - [ ] expand typings to offer concepts like
    - [ ] known schema
    - [ ] extra keywords
  - [ ] make a dependent of `ISettingsRegistry`
- [ ] refactor settings to not rely on the Schema validator to calculate the `composite` from defaults
  - > this is not supported by any implementation i could find
- [ ] add a `ISchemaValidator` provider to `@jupyterlab/validator-ajv8-extension`

Even if _that_ were done,  there would be more work to address the use of `eval` in a number other _known_ places in this codebase and dependencies.

- [translation](https://github.com/jupyterlab/jupyterlab/blob/b7c241483ec8562eb2a836ebc86086d04597aa4a/packages/translation/src/gettext.ts#L558)
- [javascript](https://github.com/jupyterlab/jupyterlab/blob/b7c241483ec8562eb2a836ebc86086d04597aa4a/packages/javascript-extension/src/index.ts#L21)
- [vega](https://github.com/jupyterlab/team-compass/issues/176)
- [pageconfig](https://github.com/jupyterlab/jupyterlab/blob/b7c241483ec8562eb2a836ebc86086d04597aa4a/packages/coreutils/src/pageconfig.ts#L68)

Of those, probably only `pageconfig` is "critical," difficult to replace functionality, while the others could probably be worked around.


bollwyvl:
This PR has bitrotted, unlikely to continue this work at this time.

PR #13931, opened 2023-02-06
Fix UI tests

fcollonval:
## References

Some UI tests are failing recently. This fixes the documentation broken due to a change in the underlying renderered map. And update the two tests waiting for the kernel indicator to be idle before comparing snapshots.

## Code changes

None

## User-facing changes

None

None


bollwyvl:
Look at that alley!

But seriously: why even keep that under screenshot test? It's basically guaranteed to break again. Seems like blanking out that DOM before the screenshot would be a lot more stable.

bollwyvl:
Some other options:
- apply a heavy [gaussian blur](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/blur) to the map elements. 
- ship a set of custom tiles with the tests
  - this would likely be the most robust

fcollonval:
bot please udpate documentation snapshots

fcollonval:

fcollonval:

Issue #13797, opened 2023-01-17
Improving `stdin` line history: forward and reverse search


telamonian:
### Problem

Currently, jlab support line history for `stdin` input boxes (which pop up during eg an `ipdb` debugging session). Said line history can be paged through one-by-one with the up/down arrow keys. However, there's currently no way to search  the history.

### Proposed Solution

Implement forward and reverse history search for `stdin` elements, a la bash/readline

krassowski:
This is a follow up to https://github.com/jupyterlab/jupyterlab/issues/7006

krassowski:
Closing as implemented in https://github.com/jupyterlab/jupyterlab/pull/13937.

PR #13937, opened 2023-02-06
implements #13797: adds reverse/forward history search for stdin inputs

telamonian:
## References

implements #13797: adds reverse/forward history search for stdin inputs

## Code changes

Adds a search implementation to the [`Stdin`](https://github.com/jupyterlab/jupyterlab/blob/affa621731641c4c052bc81ac0ad5c12b3042888/packages/outputarea/src/widget.ts#L909) class

## User-facing changes

When a user enters a search string (which can be a regex pattern) into the input box of a `Stdin` widget (eg what gets spawned when an ipdb session is initiated) and then presses ctrl-r, their command history will be searched in reverse for the first match. The user may then continue to press ctrl-r to find earlier and earlier matches to their search pattern. Similarly, if they are already at an earlier point in their command history they may instead press ctrl-s to search forward for newer and newer matches.

The goal of this implementation is to at least roughly match `reverse-i-search`/`forward-i-search` from bash. I plan to follow up this PR with various UI improvements that will more closely match the bash implementation, but initially I aimed to keep this relatively simple.


telamonian:
Currently the shortcuts for reverse/forward search are c-↑/c-↓ (ie ctrl plus an up or down arrow key). I've updated the stdin placeholder text to reflect this:

image


I'm currently planning on making two followup PRs that will build on this one:

- create an `ouputarea-extension` extension module and move the stdin line history commands into that. This will allow for remappable keyboard shortcuts with proper priority
- improve the UI around stdin line search. For one thing, I'd really like to highlight the matched substring, but I'm not sure how to partially highlight text in an `<input>` element. I'm sure there's some way to do it, or at least to fake it with a user editable `<div>`

packages/outputarea/src/widget.ts
```diff
@@ -936,15 +984,17 @@ export class Stdin extends Widget implements IStdin {
...
+    this._historyPat = '';
     this._parentHeader = options.parent_header;
-    this._value = options.prompt + ' ';
     this._password = options.password;
+    this._trans = (options.translator ?? nullTranslator).load('jupyterlab');
+    this._value = options.prompt + ' ';
+
+    this._input = this.node.getElementsByTagName('input')[0];
+    // make users aware of the line history feature
+    this._input.placeholder = this._trans.__(Stdin._placeholder);
```
krassowski, line 997: unfortunately this would not work - the translation extraction tooling requires the strings to be put within `this._trans.__()` (as it was before).

packages/outputarea/src/widget.ts
```diff
@@ -966,8 +1016,60 @@ export class Stdin extends Widget implements IStdin {
...
+        );
+
+        if (searchHistoryIx !== undefined) {
+          const historyLine = Stdin._historyAt(searchHistoryIx);
+          if (historyLine !== undefined) {
+            if (this._historyIndex === 0) {
+              this._valueCache = input.value;
+            }
+
+            console.log(`historyLine: ${historyLine}`);
```
krassowski, line 1065:
```suggestion
```

packages/outputarea/src/widget.ts
```diff
@@ -1120,7 +1214,7 @@ namespace Private {
       iframe.scrolling = 'auto';
 
       iframe.addEventListener('load', () => {
-        // Workaround needed by Firefox, to properly render svg inside
+        // Workaround needed by Firefox, to properly render svg insidehistory
```
krassowski, line 1217:
```suggestion
        // Workaround needed by Firefox, to properly render svg inside
```

packages/outputarea/src/widget.ts
```diff
@@ -908,14 +908,22 @@ export interface IStdin extends Widget {
  */
 export class Stdin extends Widget implements IStdin {
   private static _history: string[] = [];
+  private static _placeholder = '↑↓ for history. Search history with c-↑/c-↓';
```
krassowski, line 911: See below, we need this in the place where it is used.

krassowski (commented):
Would you mind adding tests?

packages/outputarea/src/widget.ts
telamonian: done

packages/outputarea/src/widget.ts
telamonian: done

packages/outputarea/src/widget.ts
telamonian: done

packages/outputarea/src/widget.ts
telamonian: didn't know that, kind of sucks. Done

fcollonval:
FYI @telamonian I want to highlight that we are freezing the API for 4.0.0 on February 28th. So you can plan appropriately the refactoring you have in mind.

telamonian:
@krassowski I added a script for the galata test that you requested. I've never actually used galata before, so some feeback/info would be helpful:

- is the script correct?
- what's the best way to run a single galata test while you're developing said test?
- how should I create and prep the related screenshot?

krassowski:
> what's the best way to run a single galata test while you're developing said test?

One terminal:

```
jlpm run start
```

another terminal:

```
jlpm run test --project jupyterlab --grep 
```

> how should I create and prep the related screenshot?

It will be auto-created when you run the test. You can try adding it to version control but likely will differ slightly due to font rendering differences between operating systems etc. Instead the recommended way is to push the code and post a comment with: "please update galata snapshots" (see https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests)

> is the script correct?

It looks good to me, but I would try to simplify it slightly - because you are taking a screenshot of entire notebook panel, any change in cell rendering will also update the screenshot. I think that markdown and raw cells in `populateNotebook` do not make any difference - can we remove them?

galata/test/jupyterlab/outputarea-stdin.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { expect, IJupyterLabPageFixture, test } from '@jupyterlab/galata';
+
+const fileName = 'stdin.ipynb';
+
+async function populateNotebook(page: IJupyterLabPageFixture) {
+  await page.notebook.setCell(0, 'raw', 'Just a raw cell');
+  await page.notebook.addCell(
+    'markdown',
+    '## This is **bold** and *italic* [link to jupyter.org!](http://jupyter.org)'
+  );
```
krassowski, line 13:
```suggestion
```

galata/test/jupyterlab/outputarea-stdin.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  await page.notebook.addCell('code', '2 ** 3');
```
krassowski, line 14:
```suggestion
  await page.notebook.addCell('code', 'raise');
```

krassowski:
I updated the `populateNotebook` notebook function as the produced screenshot shows the test did not work the way I would expect:

image

galata/test/jupyterlab/outputarea-stdin.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+test.describe('Stdin for ipdb', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.notebook.createNew(fileName);
+  });
+
+  test('Stdin history search', async ({ page }) => {
+    await populateNotebook(page);
+    await page.notebook.addCell('code', '%debug');
+    await page.notebook.run();
```
krassowski, line 20:
```suggestion
    await page.notebook.runCellByCell();
```

`raise` throws so `%debug` is not run. I think that `runCellByCell` should solve this.

krassowski:

krassowski (approved):
Thank you @telamonian!

PR #13939, opened 2023-02-07
Fixes toolbar button for Restart Kernel and Run All

### LINKED ISSUE #11898: Fast forward button in notebook toolbar has tooltip " and "
The fast-forward button in the notebook toolbar has the wrong tooltip.

## Description

The ⏩ (fast forward) button in the notebook toolbar should have a tooltip that reads, in English, "Restart Kernel and Run All Cells …". This button was introduced in #8024.

Instead, this button has the tooltip " and " (with a space before and after the word) with no additional content. You can verify this by inspecting the element in your browser.

Screen Shot 2022-01-18 at 3 33 30 PM

If you hover over a *disabled* fast-forward button, the tooltip is complete and present.

## Reproduce

Open a notebook using JupyterLab (tip of `master` branch). Hover over the ⏩ button on the toolbar.

## Expected behavior

A complete, comprehensible tooltip appears.

## Context

- Operating System and version: macOS 11.6.2 (Big Sur)
- Browser and version: Firefox 91.4.1esr (64-bit), Chrome 97.0.4692.71 (Official Build) (x86_64)
- JupyterLab version: 4.0.0a19

JasonWeill:
## References

Fixes #11898.

## Code changes

Falls back to default label/caption if all labels/captions of subcommands in a semanticCommand are empty strings

## User-facing changes

Before the change, the ⏩ toolbar button had the caption " and " (with leading/trailing spaces):

before change

After the change, the "Restart the kernel and run all cells" notebook toolbar button has an appropriate caption:

after change

Other toolbar buttons also have captions in sentence case (only the first word is capitalized) with definite articles used ("Interrupt Kernel" becomes "Interrupt the kernel").

None.

### DIFF
--- a/packages/notebook-extension/schema/panel.json
+++ b/packages/notebook-extension/schema/panel.json
@@ -18,7 +18,7 @@
       { "name": "restart", "command": "kernelmenu:restart", "rank": 32 },
       {
         "name": "restart-and-run",
-        "command": "runmenu:restart-and-run-all",
+        "command": "notebook:restart-run-all",
         "rank": 33
       },
       { "name": "cellType", "rank": 40 },
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -154,7 +154,7 @@ bumped their major version (following semver convention). We want to point out p
      mode.
    * Added the property ``sharedModel`` to the interface ``NotebookModel.IOptions``.
    * The method ``NotebookModelFactory.createNew`` receives a parameter ``NotebookModelFactory.IModelOptions``.
-   * The default Notebook toolbar's ``restart-and-run`` button now refers to the command 
+   * The default Notebook toolbar's ``restart-and-run`` button now refers to the command
      ``notebook:restart-run-all`` instead of ``runmenu:restart-and-run-all``.
 - ``@jupyterlab/rendermime`` from 3.x to 4.x
   The markdown parser has been extracted to its own plugin ``@jupyterlab/markedparser-extension:plugin``
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -154,6 +154,8 @@ bumped their major version (following semver convention). We want to point out p
      mode.
    * Added the property ``sharedModel`` to the interface ``NotebookModel.IOptions``.
    * The method ``NotebookModelFactory.createNew`` receives a parameter ``NotebookModelFactory.IModelOptions``.
+   * The default Notebook toolbar's ``restart-and-run`` button now refers to the command 
+     ``notebook:restart-run-all`` instead of ``runmenu:restart-and-run-all``.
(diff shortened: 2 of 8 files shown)


krassowski (commented):
Would you be able to add a unit test to https://github.com/jupyterlab/jupyterlab/blob/5a9896fdf16a1be81c27e185835874ba49f65c5a/packages/application/test/utils.spec.ts#L17 for this one, please?

JasonWeill:
Upon further inspection, it looks like `caption` is always considered to be `""` for the subcommands of a `semanticCommand`, even when I modified those commands to have a `caption` of their own. This seems to be the root cause of this issue.

JasonWeill:

JasonWeill:
This PR now fixes all toolbar buttons to have consistent casing by:

1. Declaring `caption` for all commands used in the notebook toolbar, with sentence casing and definite articles used. For example, if `label` is "Interrupt Kernel", `caption` is "Interrupt the kernel".
2. Using the `notebook`, not the `runmenu`, command for "Restart and run all". The `semanticCommand` logic, now reverted, would have capitalized the semantic command as "Restart the kernel and Run all cells" because each command starts with a capital letter. In the run menu, where the Title Case `label` is used, this is fine.

packages/notebook-extension/schema/panel.json
```diff
@@ -18,7 +18,7 @@
       { "name": "restart", "command": "kernelmenu:restart", "rank": 32 },
       {
         "name": "restart-and-run",
-        "command": "runmenu:restart-and-run-all",
+        "command": "notebook:restart-run-all",
```
fcollonval, line 21: Would you mind adding a small note in the migration guide?

If people have customized the toolbar, this may change their settings.

JasonWeill: Thanks — I updated the migration guide with a note.

fcollonval (approved):
Thanks @JasonWeill

krassowski:
Since this was merged I saw the icon was gone but instead the text was shown:

Screenshot from 2023-02-12 12-54-45

This was because I had customised my buttons in toolbar and the old `runmenu:restart-and-run-all` remains for me rather than getting the now-default.

To fix this manually go into Settings → Notebook Panel → find item with `runmenu:restart-and-run-all` and replace it with `notebook:restart-run-all`.

I think that this will affect everyone who opened settings Settings → Notebook Panel in earlier version as this wrongly triggers an update.

I am not sure if want to take any action here, just flagging this.

JasonWeill:
@krassowski I also saw this and I didn't realize that this was the cause. Is there something that I can do to automate this process on upgrade?

krassowski:
I guess we could consider shipping a migration script, but I think we don't attach application version to the settings which will makes it non trivial. Let's continue the discussion in https://github.com/jupyterlab/jupyterlab/issues/13972

PR #13944, opened 2023-02-07
Split standard input line history per session (notebook)

### LINKED ISSUE #13798: Improving `stdin` line history: option to split line history per notebook
### Problem

Currently, the line history support for `stdin` input boxes (which pop up during eg an ipdb debugging session) is global for a single instance of jupyterlab. This is personally how I prefer line history to work, but I've gotten some feedback that it can be confusing for this history to not be split up for different notebooks.

### Proposed Solution

Add an option that will segregate the `stdin` line history per notebook. This might be a little tricky to do, but I think it can be done relatively easily by making a few modifications to the data structure under which `stdin` line history is currently stored.

telamonian:
## References

Closes #13798

## Code changes

- adds new `inputHistoryScope` setting to notebook (and intialization options of: `Cell`, `OutputArea` and `Stdin`)

## User-facing changes

- new `Input History Scope` setting

NA


telamonian:
The trick to getting this to work turned out to be buried in the kernel msg that's used to request the `Stdin` input widget in the first place. The kernel session id is included as part of that stdin request msg, which is then passed on to the `Stdin` constructor, which can then be used to split up the cmd history when the `splitStdinHistoryBySession` preference is enabled.

So technically this ended up being a "split stdin history by kernel session" feature rather than being "split by notebook", but I actually think it's better this way.

@krassowski @fcollonval Could one of you please take a look at this when you have a sec? Ideally I'd like to get this in before the feature freeze on Wednesday (though I realize that may be a heavy ask).

fcollonval (changes requested):
Thanks @telamonian 

I tested it on Binder. It works as expected and the history is kept shared for console for example.

On the implementation level, we should move the configuration flag in [`INotebookConfig`](https://github.com/jupyterlab/jupyterlab/blob/d11a3e73bdf4c9b5d2266911312f837c84f1f69b/packages/notebook/src/widget.ts#L1036) as this is not related to the editor

> The current code will also not be possible if you rebase on master - so I advice you to first move to notebook config before rebasing.

This means also you need to move the settings as a primary property like `maxNumberOutputs`.

Regarding the history storage, I would prefer using a `Map` rather than a object. And we should try to clean the history if the session gets disposed. But that seems quite complex for this small amount of data; something like a static public method on `Stdin` triggered by `session.disposed`?

packages/notebook-extension/schema/tracker.json
```diff
@@ -734,6 +734,12 @@
           "type": "boolean",
           "default": false,
           "title": "Show Trailing Space"
+        },
+        "splitStdinHistoryBySession": {
+          "type": "boolean",
+          "default": false,
+          "title": "Split Stdin History by Kernel",
+          "description": "When enabled, line history for stdin (eg the ipdb prompt) is kept separately for different kernel sessions."
```
krassowski, line 742: I think we should go for a more future-proof `inputHistoryScope: 'global' | 'session'` (and in future 'environment') to allow extensibility. In fact we could also add 'kernel' or 'language' immediately which would group by kernel language (and could be a sensible default). I will push a commit.

krassowski:
> Regarding the history storage, I would prefer using a `Map` rather than a object. And we should try to clean the history if the session gets disposed. But that seems quite complex for this small amount of data; something like a static public method on `Stdin` triggered by `session.disposed`?

This could be a `WeakMap` using Session object which would avoid adding extra complexity in signals and defer GC to browser.

krassowski:
So doing this with a static method sounds nice (though introduces an additional dependency on `outputarea` in `notebook-extension` - at least in the way I tried) but in fact the session ID is different from both: session (connection) ID and kernel (connection) ID which are easily accessible from `session` object, hence `session.disposed` does not give us information on what we should clean (or I cannot find it).

Using a `WeakMap` where key is session object would be more suitable but the behaviour would be different as there would be no way to achieve persistence when refreshing page; while there is no persistence as-is, it could be more easily implemented in the future if we do not go for `WeakMap` solution.

Given the tiny benefit of cleaning it up I propose we move forward without it.

fcollonval (approved):
Thanks @krassowski for finishing up this one

PR #13959, opened 2023-02-10
Fix verdaccio start up with nodejs 18.14.0

fcollonval:
## References

The CI release test is failing very often recently

## Code changes

Bump verdaccio

## User-facing changes

None

None

### DIFF
--- a/buildutils/package.json
+++ b/buildutils/package.json
@@ -55,7 +55,7 @@
     "semver": "^7.3.2",
     "sort-package-json": "~1.53.1",
     "typescript": "~4.7.3",
-    "verdaccio": "^5.19.1"
+    "verdaccio": "^5.20.1"
   },
   "devDependencies": {
     "@types/fs-extra": "^9.0.1",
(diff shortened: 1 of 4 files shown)


fcollonval:
This does not solve the issue

There is a redundancy between check_release and release_test. The later is more complete as it test the release assets. But the former is closer to the release process as it uses the releaser. So it would be good to add a job in the check_release workflow that test the package as done in the release_test and drop that one.

jtpio:
> So it would be good to add a job in the check_release workflow that test the package as done in the release_test and drop that one.

This sounds good. I think the `release_test` predates the releaser and was kept because it indeed tests more things.

fcollonval:
The issue is due to the removal of `npm bin` in npm 9: https://docs.npmjs.com/cli/v9/using-npm/changelog#-breaking-changes

I don't know why it did not fail for previous nodejs 18 version - I guess the command was silently failing.

fcollonval:
Ok I found the culprit, with NodeJS v18.14.0 npm has been bump from v8 to v9:

image

Ref: https://nodejs.org/en/download/releases/

jtpio (approved):
Thanks!

fcollonval:

Issue #13857, opened 2023-01-25
Cell toolbar sometimes overlays with text (if long enough text is present)


andrii-i:
## Description
When cells size changes, cell toolbar disappears to prevent obscuring / overlaying text. However this behavior is not consistent if long text that reaches cell toolbar is typed or pasted into the cell, cell toolbar will overlay over text: 
![cell_toolbar_overlayed](https://user-images.githubusercontent.com/26686070/214718554-bfaf2c00-5c96-4aa8-8f64-69033d31d782.gif)

## Reproduce

1. Create a new notebook
2. Create a cell
3. Type a string long enough to almost "reach" cell toolbar
4. Resize the cell
5. Note that cell toolbar doesn't disappear and overlays the text

Problem probably has to do with how and when `jp-toolbar-overlap` css class/class name that hides the toolbar is applied:

https://github.com/jupyterlab/jupyterlab/blob/fc7ff81505b130a3c9c5364c85bd25e27160bfda/packages/cell-toolbar/src/celltoolbartracker.ts#L32-L35

## Expected behavior

Toolbar disappears at the point that prevents overlaying text. 

## Context

Operating System and version: MacOS 12.6.1
Browser and version: Firefox 102.6.0esr (64-bit)
JupyterLab version: 3.5.2, head-of-main (WIP 4.0.0)

jtpio:
Not sure this is the exact same issue, but there are some other odd behaviors when collapsing a cell. Here testing with the latest pre-release (`4.0.0a33`):

https://user-images.githubusercontent.com/591645/217554120-1be89006-a14d-467d-b621-578aad8e6989.mp4

JasonWeill:
@jtpio The cell toolbar showing up on collapsed cells was covered by #13254, ostensibly fixed by #13847. We might want to reopen #13254 if this is still an issue using builds from tip of `master`

JasonWeill:
Update: reopened #13254 for cell toolbar issue.

JasonWeill:
When `_updateCellForToolbarOverlap` is called, the toolbar may not yet be visible in the cell. `_addToolbar` calls `insertWidget` on the cell's layout (a `PanelLayout`) but the DOM as shown in the browser may not be updated at the moment that `_updateCellForToolbarOverlap` starts to run.

Issue #13683, opened 2022-12-29
All toolabar buttons get re-rendered on tab switch


krassowski:
## Description

When switching tabs, toolbars get re-rendered via React.

## Reproduce

1. Open four notebooks so that three are visible and third is hidden as so:
   ![Screenshot from 2022-12-29 15-59-18](https://user-images.githubusercontent.com/5832902/209978378-46851419-cc4a-4824-a62a-f0226b3915ca.png)
3. Record with React profiler
4. Switch to between the active and inactive tabs in the top panel
5. See that all toolbar buttons (`CommandToolbarButtonComponent`) were re-rendered

For example:

![Screenshot from 2022-12-29 16-14-30](https://user-images.githubusercontent.com/5832902/209980183-62c89491-724d-419d-a836-81a33ae313ee.png)

While each button renders fast (average ~0.15ms), in the above layout (4 notebooks) there are 64 toolbar buttons giving a total of ~10ms.

With 10 notebooks open like that:

![Screenshot from 2022-12-29 16-11-42](https://user-images.githubusercontent.com/5832902/209979816-a909f7bb-2626-4d6a-885b-ae632fa5e407.png)

There are 160 buttons giving ~24ms delay.

(this is of course worse with more buttons added by extensions)

## Expected behavior

We do not re-render things which did not change/are hidden.

## Context

- JupyterLab version: 4.0.0a31

fcollonval:
This is an intrinsic trouble with the way `ReactWidget` works. I guess the best fix would be to only trigger update if the widget is visible and add a update trigger on visible changed from false to true.

Line to blame: https://github.com/jupyterlab/jupyterlab/blob/3d7445fd142fedf65b9ba26df3b4bc649b168054/packages/ui-components/src/components/vdom.ts#L48

> This is one of the reason for JupyterLab 5, we should get away from `Widget` as they don't have a good enough lifecycle  (too many steps are useless and the update one is not smart enough) and nowadays a standard exists.

krassowski:
The problem here is possibly more self-contained. When looking at the `CommandToolbarButtonComponent`:

https://github.com/jupyterlab/jupyterlab/blob/ab04b50169c1e0c524f975895f657e787a4d9902/packages/ui-components/src/components/toolbar.tsx#L1065-L1083

[...]

PR #13964, opened 2023-02-10
Hides cell toolbar when cell is collapsed

### LINKED ISSUE #13254: Cell toolbar shows up on collapsed cells even when shrunk
## Description

Cell toolbar shows up on collapsed cells even when shrunk.

## Reproduce

Screenshot from 2022-10-16 22-05-06

1. Create a second cell, enter it
2. Collapse the cell
3. See the cell toolbar stays even when it should be disabled due to space constraints (too narrow window)

## Expected behavior

Cell toolbar is either:
- not visible on collapsed cells
- not visible on collapsed cells on narrow screens to avoid visual overflow and inability to click on the expand button

## Context

- JupyterLab version: 3.x and 4.0.0a30

JasonWeill:
## References

Fixes #13254 (second change)

## Code changes

Modifies the code to respond to changes in cells being collapsed or expanded.

## User-facing changes

After collapsing a cell, the cell toolbar should never appear beside the placeholder.

after collapsing

After expanding a collapsed cell, the cell toolbar should appear.

after expanding

Known issue: cell toolbar may overlay content (see #13857). This change partially fixes the overlap checking functionality.

None.


packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -322,13 +365,21 @@ export class CellToolbarTracker implements IDisposable {
...
+    let lineRight = codeMirrorLines[0].getBoundingClientRect().left;
+    const oneChar = document.createElement('span');
+    oneChar.textContent = 'a';
+    codeMirrorLines[0].appendChild(oneChar);
+    const charWidth = oneChar.getBoundingClientRect().width;
+    codeMirrorLines[0].removeChild(oneChar);
+    // Record only the text content's length for each child node
+    codeMirrorLines[0].childNodes.forEach((childNode: ChildNode) => {
+      lineRight += (childNode.textContent?.length || 0) * charWidth;
+    });
```
krassowski, line 382: Forced style calculation alert. We can do this without modifying DOM (and get more accurate results at the same time):

```suggestion
    const range = document.createRange();
    range.selectNodeContents(codeMirrorLines[0].childNodes);
    const sizes = range.getClientRects();
    lineRight += [...sizes].reduce((prev, rect) => prev + rect.width, 0);
```

krassowski:
I tagged this for 4.0 because of changes tailored towards CodeMirror 6. I _think_ that the part reflected in title of the PR (cell collapsing fix) could still be backported to 3.6.

packages/cell-toolbar/src/celltoolbartracker.ts
JasonWeill: Thank you! The proposed line 375 gave me a compilation error for attempting to provide a node list where a single node was expected. I pushed a new change with `setStart` and `setEnd` calls instead.

krassowski: Sorry it should be `range.selectNodeContents(codeMirrorLines[0]);` without `.childNodes`.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -322,13 +365,22 @@ export class CellToolbarTracker implements IDisposable {
...
+
+    let lineRight = codeMirrorLines[0].getBoundingClientRect().left;
+    const range = document.createRange();
+    range.setStart(codeMirrorLines[0].childNodes[0], 0);
+    const nodeLength = codeMirrorLines[0].childNodes.length;
+    const lastNode = codeMirrorLines[0].childNodes[nodeLength - 1];
+    range.setEnd(
+      lastNode,
+      Math.max((lastNode?.textContent?.length || 0) - 1, 0)
+    );
```
krassowski, line 381: Can we try?

```suggestion
    range.selectNodeContents(codeMirrorLines[0]);
```

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -322,13 +365,22 @@ export class CellToolbarTracker implements IDisposable {
...
+    const sizes = range.getClientRects();
+    lineRight += [...sizes].reduce((prev, rect) => prev + rect.width, 0);
```
krassowski, line 383: Am I correct thinking that `lineRight == sizes[sizes.length - 1].right`?

Could we possibly simplify this further by just taking the last child in the line and taking its `right`, or would that fail?

I am suggesting something like:

```
lineRight = codeMirrorLines[0].lastChild.getClientRects().right;
```

JasonWeill: `getBoundingClientRect` exists on an `Element`, but not on a text `Node`, so `ChildNode` doesn't include it as a callable method.

packages/cell-toolbar/src/celltoolbartracker.ts
krassowski: Since the other suggestion was a miss, what do you think about this one?

JasonWeill: This works! Thanks @krassowski

krassowski:
bot please update snapshots

jtpio:
CI kick

fcollonval:

fcollonval:

fcollonval (commented):
@JasonWeill lost of toolbars disappear in the UI test snapshots. I could not see that when testing locally. So I guess it has to do with some asynchronicity of the code. What do you think?

JasonWeill:
@fcollonval Thanks for updating the snapshots! Looking at them:

1. ✅ galata/test/jupyterlab/notebook-edit.test.ts-snapshots/delete-cell-jupyterlab-linux.png — this change is correct; the first line overlaps with the cell toolbar, so the toolbar should be invisible.
2. ✅ galata/test/jupyterlab/notebook-markdown.test.ts-snapshots/do-not-highlight-not-latex-jupyterlab-linux.png — this change is correct; the first line overlaps with the cell toolbar, so the toolbar should be invisible.
3. ❌ galata/test/jupyterlab/notebook-markdown.test.ts-snapshots/do-not-highlight-standalone-dollar-jupyterlab-linux.png — the first line does not overlap the toolbar, so the toolbar should be visible.
4. ❌ galata/test/jupyterlab/notebook-replace.test.ts-snapshots/replace-in-markdown-rendered-cell-jupyterlab-linux.png — the first line does not overlap the toolbar, so the toolbar should be visible.
5. ❌ galata/test/jupyterlab/notebook-run-vega.test.ts-snapshots/run-cells-dark-vega-jupyterlab-linux.png — the first line does not overlap the toolbar, so the toolbar should be visible.
6. ❌ galata/test/jupyterlab/notebook-run-vega.test.ts-snapshots/run-cells-vega-jupyterlab-linux.png — the first line does not overlap the toolbar, so the toolbar should be visible.

JasonWeill:
I can't reproduce the test failures in the pipeline using Firefox on macOS. In my screen shots below, I see the cell toolbar as expected.

galata/test/jupyterlab/notebook-markdown.test.ts-snapshots/do-not-highlight-standalone-dollar-jupyterlab-linux.png

Image

﻿﻿﻿﻿---

galata/test/jupyterlab/notebook-replace.test.ts-snapshots/replace-in-markdown-rendered-cell-jupyterlab-linux.png

Image (1)

---

galata/test/jupyterlab/notebook-run-vega.test.ts-snapshots/run-cells-dark-vega-jupyterlab-linux.png

Image (2)

---

galata/test/jupyterlab/notebook-run-vega.test.ts-snapshots/run-cells-vega-jupyterlab-linux.png

Image (3)

@fcollonval I don't have regular access to a Linux machine; how can I debug these screenshot test failures?

fcollonval:
I tested the following test locally:

```sh
PWDEBUG=1 jlpm test -g 'Do not enter math mode for standalone dollar'
```

It should have a toolbar but it does not. And this is not a question of asynchronicity; waiting never bring the toolbar. But activating another cell and coming back to the cell does.

Introspecting the DOM, the cell toolbar is on the expected cell but it has `display: none` style.
We may need to wait for some sort of readiness in `_codeOverlapsToolbar` before checking to display the toolbar or not.

fcollonval:
Did you reproduce the test manually on Firefox or just changed the browser target of playwright to be Firefox?

fcollonval:
Switching to `firefox` in playwright configuration locally, indeed the toolbar is displayed.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -322,13 +365,17 @@ export class CellToolbarTracker implements IDisposable {
...
     }
-    const lineRight = codeMirrorLines[0].children[0] // First span under first pre
-      .getBoundingClientRect().right;
+
+    let lineRight = codeMirrorLines[0].getBoundingClientRect().left;
+    const range = document.createRange();
+    range.setStart(codeMirrorLines[0].childNodes[0], 0);
+    range.selectNodeContents(codeMirrorLines[0]);
+    const sizes = range.getClientRects();
+    lineRight += [...sizes].reduce((prev, rect) => prev + rect.width, 0);
```
krassowski, line 378: Let's try this:

```suggestion
    range.selectNodeContents(codeMirrorLines[0]);
    lineRight += range.getBoundingClientRect().width;
```

After debugging the failures it seems that Chrome double-counts some nodes in the range when using `getClientRects` approach. The proposed approach works when testing in-browser.

JasonWeill:
@krassowski Thanks for making this new commit! I see three test failures in UI tests, although for two of them, the new "actual" is actually correct. For the third, "Delete cell" (`test/jupyterlab/notebook-edit.test.ts:92`), the cell toolbar appears overlapping line 1 of the Markdown source, whereas it shouldn't appear in this situation.

krassowski:
> For the third, "Delete cell" (test/jupyterlab/notebook-edit.test.ts:92), the cell toolbar appears overlapping line 1 of the Markdown source, whereas it shouldn't appear in this situation.

From a quick look this is not a problem with `lineRight` anymore but an issue with `toolbarLeft`. When cell gets focused the toolbar widget is not yet correctly positioned because the buttons are not actually rendered. This only happens when activating a cell, not when resizing an active cell.

toolbar-issue

It is related to #13683 but distinct.

1) A naive fix would be to add min-width to the buttons on the toolbar:
    ```css
    .jp-cell-toolbar > .jp-Toolbar-item {
      min-width: 30px
    }
    ```
    This is of course a bad solution long-term because we would get overlap in presence of longer toolbar items if an extension adds a longer item.
2) A better fix would be to wait until all toolbar buttons are rendered. This could be implemented by waiting for a specific class name to appear in the toolbar node (to avoid depending on React lifecycle).
3) The ideal fix would be to prevent re-rendering toolbar on changing cells altogether.


Looking at the code, the toolbar is re-rendered on every change of active cell:

https://github.com/jupyterlab/jupyterlab/blob/02113198e1cf453b34c4248ccfc3143c3f0f5e5d/packages/cell-toolbar/src/celltoolbartracker.ts#L107-L125

It appears that the code already tries to wait for the items to be rendered (2), but these are clearly not ready by the time we reach `this._updateCellForToolbarOverlap` in:

https://github.com/jupyterlab/jupyterlab/blob/02113198e1cf453b34c4248ccfc3143c3f0f5e5d/packages/cell-toolbar/src/celltoolbartracker.ts#L151-L188

From a quick check the render promises are added to the `promises` list correctly but they are all already fulfilled.
It seems that for this logic to work we would need to call `update` to regenerate the promises (which works locally, I will push a commit).

krassowski:
Let's see how it works: bot please update galata snapshots. @JasonWeill is there a reason for re-rendering the toolbar when switching cells? Could we avoid it altogether (solution 3 above)? If there is a no strong counterpoint, should we open a new issue to track this as a performance enhancment?

JasonWeill:
> Let's see how it works: bot please update galata snapshots. @JasonWeill is there a reason for re-rendering the toolbar when switching cells? Could we avoid it altogether (solution 3 above)? If there is a no strong counterpoint, should we open a new issue to track this as a performance enhancment?

@krassowski If we use the same toolbar every time, would we lose the ability to disable the move up / move down buttons, as suggested in #13502?

fcollonval:
> @krassowski If we use the same toolbar every time, would we lose the ability to disable the move up / move down buttons, as suggested in #13502?

We will not as the buttons themself will get rerendered if the notify command changed is emitted:

https://github.com/jupyterlab/jupyterlab/blob/c6180cefc57448a92a4761b66069cdd6c6f6637c/packages/ui-components/src/components/toolbar.tsx#L841-L849

fcollonval:
Kicking the CI

krassowski (approved):
Thank you @JasonWeill and thanks for the patience while we were iterating to find a clean and performant solution!

krassowski:
Let's continue discussion on refactoring the cell-toolbar rendering in https://github.com/jupyterlab/jupyterlab/issues/14139#issue-1611970404.

JasonWeill:
Thank you @krassowski @fcollonval for your help!

PR #13974, opened 2023-02-13
Relax memory-leak action

fcollonval:
## References

The action for memory leak is pinned to a patch version. This makes maintenance annoying.

## Code changes

Relax version - the tag for major and minor version is updated on jupyterlab/benchmark repository

## User-facing changes

None

None


fcollonval (commented):
bot please run benchmarks

Issue #13973, opened 2023-02-13
Add back `@types/react`?


fcollonval:
## Description

In the [bump to React v18](https://github.com/jupyterlab/jupyterlab/pull/13607/files#diff-5cda0f66564895b4d43d7c4bd1820b333c65296644f3f9ba39f7601013c480d3), we drop the dependency to `@types/react` in `@jupyterlab/apputils`.

This brings trouble for downstream extension:

```
       $ tsc --sourceMap
      node_modules/@jupyterlab/apputils/lib/dialog.d.ts(4,24): error TS7016: Could not find a declaration file for module 'react'. '/home/runner/work/extension-cookiecutter-ts/extension-cookiecutter-ts/my_extension/node_modules/react/index.js' implicitly has an 'any' type.
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`
      node_modules/@jupyterlab/apputils/lib/kernelstatuses.d.ts(23,15): error TS2503: Cannot find namespace 'JSX'.
      node_modules/@jupyterlab/apputils/lib/runningSessions.d.ts(15,15): error TS2503: Cannot find namespace 'JSX'.
      node_modules/@jupyterlab/codeeditor/lib/lineCol.d.ts(3,19): error TS7016: Could not find a declaration file for module 'react'. '/home/runner/work/extension-cookiecutter-ts/extension-cookiecutter-ts/my_extension/node_modules/react/index.js' implicitly has an 'any' type.
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`
      node_modules/@jupyterlab/ui-components/lib/components/button.d.ts(1,19): error TS7016: Could not find a declaration file for module 'react'. '/home/runner/work/extension-cookiecutter-ts/extension-cookiecutter-ts/my_extension/node_modules/react/index.js' implicitly has an 'any' type.
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`
      node_modules/@jupyterlab/ui-components/lib/components/button.d.ts(21,54): error TS2503: Cannot find namespace 'JSX'.
[...]

PR #4386, opened 2018-04-14
Restore typings to apputils dependencies.

ian-r-rose:
Reverts #4367.

I am unsatisfied with this solution, since different versions of `@types/react` are mutually incompatible, and it makes this a very brittle dependency to include. I see this as the lesser of two evils, though.

PR #13981, opened 2023-02-13
Add back `@types/react` as a dependency of `@jupyterlab/apputils`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13973

Similar to https://github.com/jupyterlab/jupyterlab/pull/4386

## Code changes

Add back `@types/react` as a dependency of `@jupyterlab/apputils`. 

## User-facing changes

None


fcollonval (approved):
Thanks

PR #13985, opened 2023-02-13
Increase galata update timeouts

krassowski:
Galata snapshots update job is timing out after 80 minutes, blocking https://github.com/jupyterlab/jupyterlab/pull/13884#issuecomment-1428620997.

This PR increases the timeouts to 120 minutes.


fcollonval (approved):
Thanks @krassowski

Issue #13972, opened 2023-02-13
Toolbar change due to button definition change


fcollonval:
Since https://github.com/jupyterlab/jupyterlab/issues/13939 was merged I saw the icon was gone but instead the text was shown:

![Screenshot from 2023-02-12 12-54-45](https://user-images.githubusercontent.com/5832902/218312306-83935977-ac16-4125-9faa-e1b2a2e19584.png)

This was because I had customised my buttons in toolbar (well I did not, but when you go to Notebook Panel in settings it saves the default to user settings) and the old `runmenu:restart-and-run-all` remains for me rather than getting the now-default.

To fix this manually go into Settings → Notebook Panel → find item with `runmenu:restart-and-run-all` and replace it with `notebook:restart-run-all`.

I think that this will affect everyone who opened settings Settings → Notebook Panel in earlier version as this wrongly triggers an update.

I am not sure if want to take any action here, just flagging this.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/13939#issuecomment-1427027301_

fcollonval:
I created a specific issue, so we can discuss this for 4.0.0

PR #14000, opened 2023-02-14
Restore using `runmenu:restart-an-run-all`

fcollonval:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13972

## Code changes

Keep the previous command for the notebook toolbar
Lower case captions in semantic commands.

## User-facing changes

Previously modified settings will not need change for JupyterLab v4

None

### DIFF
--- a/packages/application/src/utils.ts
+++ b/packages/application/src/utils.ts
@@ -125,7 +125,12 @@ export function createSemanticCommand(
     attribute: 'label' | 'caption'
   ): string | CommandRegistry.CommandFunc<string> | undefined {
     return () => {
-      const texts = reduceAttribute(attribute);
+      const texts = (reduceAttribute(attribute) as string[]).map(
+        (text, textIndex) =>
+          attribute == 'caption' && textIndex > 0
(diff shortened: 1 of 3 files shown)


packages/application/src/utils.ts
```diff
@@ -125,7 +125,12 @@ export function createSemanticCommand(
     attribute: 'label' | 'caption'
   ): string | CommandRegistry.CommandFunc<string> | undefined {
     return () => {
-      const texts = reduceAttribute(attribute);
+      const texts = (reduceAttribute(attribute) as string[]).map(
+        (text, textIndex) =>
+          attribute == 'caption' && textIndex > 0
+            ? text.toLocaleLowerCase()
```
JasonWeill, line 131: This is not going to be useful in languages like German, where common nouns are in uppercase.

I don't think this is going to be fixable before 4.0.0 is done, so we should have a separate issue for this.

JasonWeill: Filed #14005 about this.

JasonWeill (approved):

PR #14012, opened 2023-02-15
Galata config helper should not set dev_mode

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/extension-cookiecutter-ts/pull/287

## Code changes

The server config helper should not set `dev_mode`

## User-facing changes

Testing with galata starting the server with the helper method should work.

None


jtpio (approved):
Thanks!

PR #14024, opened 2023-02-17
Adds placeholder when TOC is empty

### LINKED ISSUE #13988: TOC is totally blank when no headings exist
## Description

When there are no headings in a notebook, the Table of Contents displays nothing in its main area. As a result, a novice user would have no idea what the widget is, or whether it's working correctly.

image

## Reproduce

1. Open a new Jupyter Notebook in JupyterLab
2. Open the TOC in the left sidebar

## Expected behavior

See explanatory text or icon (with text on hover/click) that the Table of Contents cannot be displayed because there are no headings in the notebook.

## Context

- Operating System and version: macOS 12.6.3
- Browser and version: Firefox 102.7.0esr
- JupyterLab version: 4.0.0a33

JasonWeill:
## References

Fixes #13988.

## Code changes

Adds placeholder messaging when there are no headers in the active Jupyter notebook, or when the active editor is of a non-notebook file (a text file, etc)

## User-facing changes

When no headings exist, there is now a placeholder in the left panel, using the [GNOME empty state design](https://wiki.gnome.org/Design/OS/EmptyStates) as inspiration:

image

Before, the largest portion of the TOC panel was blank.

None.


packages/toc/src/treeview.tsx
```diff
@@ -26,8 +28,18 @@ export class TableOfContentsWidget extends VDomRenderer<TableOfContents.IModel<T
    * includes layout triggered rendering.
    */
   render(): JSX.Element | null {
-    if (!this.model) {
-      return null;
+    if (!this.model || this.model.headings.length === 0) {
+      this.parent?.addClass('jp-TableOfContents-empty');
+      return (
+        <div className="jp-TableOfContents-placeholder">
+          <div className="jp-TableOfContents-placeholderContent">
```
ellisonbg, line 35: Let's make sure to use the JupyterLab theme variables for the type styling: font size, color, etc.

packages/toc/src/panel.ts
```diff
@@ -28,7 +28,12 @@ export class TableOfContentsPanel extends SidePanel {
     this._title = new Private.Header(this._trans.__('Table of Contents'));
     this.header.addWidget(this._title);
 
-    this._treeview = new TableOfContentsWidget();
+    this._treeview = new TableOfContentsWidget({
+      placeholderHeadline: this._trans.__('No Headings'),
+      placeholderText: this._trans.__(
+        'The table of contents shows headings in Jupyter notebooks.'
```
krassowski, line 34:
```suggestion
        'The table of contents shows headings in notebooks and supported files.'
```

Unless it changed in 4.0, In Python files ToC shows the outline (function/class definitions). It should (aspirationally) also show up in markdown files. Maybe even shorter:

`Open a notebook with headings or supported file to see outline`

packages/toc/src/treeview.tsx
```diff
@@ -26,8 +28,18 @@ export class TableOfContentsWidget extends VDomRenderer<TableOfContents.IModel<T
...
+            <h1>{this._placeholderHeadline}</h1>
```
krassowski, line 36:
```suggestion
            <h3>{this._placeholderHeadline}</h3>
```

`<h1>` is out of place here (the hierarchy should be maintained) - it cannot be after `<h2>` in hierarchy (and sidebar title is `<h2>`). It also looks too big and IMO should be in grey and use standard themed sizes and colour (as in the Brian's point about using lab theme variables).

packages/toc/style/base.css
```diff
@@ -20,8 +20,22 @@
...
+  padding: 8px;
+}
+
 .jp-TableOfContents .jp-SidePanel-content {
   overflow-y: auto;
+  display: flex;
+  align-items: flex-start;
+}
+
+.jp-TableOfContents .jp-TableOfContents-empty.jp-SidePanel-content {
```
krassowski, line 37: Can we simplify this selector? (style calculation performance).

```suggestion
.jp-TableOfContents-empty {
```

JasonWeill: This is too simplified; with the change applied, the `align-items: center` rule does not take precedence. With `.jp-TableOfContents-empty.jp-SidePanel-content` as the selector, it does.

krassowski: `important!`?

JasonWeill: I could, but I try to avoid using `!important` as much as possible.

krassowski (commented):
This looks good! I would suggest one more iteration:
- could we style the `<h3>` heading with  `margin-bottom: var(--jp-content-heading-margin-bottom);` or similar to reduce the margin between the heading and the text?
- when the ToC appears after adding a heading the blank indicator flickers (changes position) as seen on this gif:
   jitter-toc
   This is because the `jp-TableOfContents-empty` class gets removed before rendering the contents:
   | before re-renderign starts | during re-render |
   | --- | --- |
   | Screenshot from 2023-02-18 11-18-18 | Screenshot from 2023-02-18 11-18-36 |
   
   we should avoid this spurious re-render both to prevent flicker and prevent performance degradation when working with large documents.


Two questions:
- should it be placed in the middle of the panel or on top? (I am not convinced either way myself, but that would also simplify things with the jitter)
- does it deserve a screenshot test?

packages/toc/src/treeview.tsx
```diff
@@ -15,8 +15,10 @@ export class TableOfContentsWidget extends VDomRenderer<TableOfContents.IModel<T
    *
    * @param options Widget options
    */
-  constructor(options: TableOfContents.IOptions = {}) {
```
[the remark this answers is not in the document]
JasonWeill: Advantages of putting the placeholder in the middle:

* Aligns with GNOME UI guidelines
* Consistent with other existing "blank widget" placeholders in JupyterLab

Advantages of putting the placeholder up top:

* No flicker associated with switching CSS classes from "empty" to "non-empty" TOC
* Closer to the other UI elements, especially on a tall monitor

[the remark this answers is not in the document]
JasonWeill: Here's what the TOC placeholder looks like if we remove the "empty" class, so content always aligns to the top of the container.

We could potentially improve the experience by putting in a large icon or stock image representing "table of contents".

image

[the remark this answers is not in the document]
JasonWeill: … and here's what the placeholder would look like with the TOC icon blown up to 128x128 and modified to have opacity 0.7 (similar to the text color).

image

image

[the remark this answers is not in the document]
krassowski: @JasonWeill thanks for exploring this a bit. I am happy to move forward here by always aligning with the top of the container. I think the TOC icon does not look good here - it was not designed for that size. We could pursue:
- finding an icon which would look better (for now keeping it without an icon), and
- performance improvements
in follow-up issues.

JasonWeill:
@krassowski Updated to top-align the placeholder, per discussion. Updated the screen shot in the PR description.

packages/toc/style/base.css
```diff
@@ -20,8 +20,23 @@
...
+.jp-TableOfContents-placeholder {
+  text-align: center;
+}
+
+.jp-TableOfContents-placeholderContent {
+  color: var(--jp-content-font-color2);
+  padding: 8px;
+}
+
+.jp-TableOfContents-placeholderContent h3 {
```
krassowski, line 32:
```suggestion
.jp-TableOfContents-placeholderContent > h3 {
```

CSS performance

krassowski (approved):
Thank you @JasonWeill!

fcollonval (changes requested):
The  UI test failures are legit. One need to be updated to see the new placeholder and two needs fixes as the execution indicator position is broken:

image

packages/toc/style/base.css
```diff
@@ -20,8 +20,23 @@
...
+  margin-bottom: var(--jp-content-heading-margin-bottom);
+}
+
 .jp-TableOfContents .jp-SidePanel-content {
   overflow-y: auto;
+  display: flex;
+  align-items: flex-start;
```
fcollonval, line 39: This breaks the execution indicator - I guess it should only be applied in the case of the placeholder

JasonWeill: These two lines weren't necessary for the placeholder; I've removed them.

JasonWeill:
@fcollonval Can you please take another look when you get the chance?

fcollonval (approved):
Thanks for the patience @JasonWeill

PR #14026, opened 2023-02-17
Using metadataform for default notebooktools

brichet:
This PR replaces the current implementation of the *common tools* and the *advanced tools* using `metadataform` settings.

It will make these tools more configurable as user or extension settings will be able to add fields in it.
In future, some fields of a tool or a whole tool could also be hidden (these features are not yet supported by `metadataform`).

## Code changes

- **creates** widgets "RJSF compliant" for:
   - cell tags
   - active cell tool
   - metadata editors (cells and notebook)
- removes `_commonTools` and `_advancedTools` from the `NotebookTools` widget.
- removes widgets related to `_commonTools` in the NotebookTools namespace:
   - NotebookTools.KeySelector()
   - NotebookTools.NotebookMetadataNumberTool()
- removes functions creating the  `_commonTools` fields in the NotebookTools namespace:
   - NotebookTools.createSlideShowSelector()
   - NotebookTools.createNBConvertSelector()
   - NotebookTools.createEditableToggle()
   - NotebookTools.createToCBaseNumbering()
- removes `celltags` package as the widget is implemented in `celltags-extension`

## User-facing changes

Changes the appearance of the common tool.
BEFORE: 
  
AFTER: 
 

Removes some previously exported tools, but not used anymore in JupyterLab itself.

## Notes

- the `celltags-extension` content could probably be moved to the `notebook-extension` package.
- the `NotebookTools.NotebookMetadataEditorTool` and `NotebookTools.CellMetadataEditorTool` are not used anymore in JupyterLab (as far as I know) but remains in code currently, they may be useful for extensions...


bollwyvl:
Cool stuff! I'd like to see the whitespace line up closer to the original:
- no indent on the selectors for the clean left vertical alignment 
- slightly more offset between the labels and fields, maybe `0.5em`

Not matching the select style is... fine.

bollwyvl:
Also, the [form validation shots](https://github.com/jupyterlab/jupyterlab/pull/14026/files#diff-fc918c2a4ba73f5dedec9df4a8488ef8975dfe882990894822ab369bc577e399) show how the error bar is too close to the text. Should probably also be more like `0.5em`

bollwyvl:
The grey "selected field background" is also a bit tight on:

brichet:
Thanks @bollwyvl.

I updated the CSS according to your suggestions, it should look better now:
- removes padding left on descriptions and input/select
- uses a darker grey on selected field background
- adds a margin between the error bar and the text
- adds an offset between the title and the input
- better use of the 100% width (the right part of the input was hidden sometime)

The changes can be seen in the updated galata snapshots.

bollwyvl:
Changes looking really clean, and I really appreciate the quick turnaround, and _deeply_ empathize with the challenges of theming `rjsf` stuff.

Some remaining nits (while waiting to try some of this out once binder works):



perhaps one of:
- remove the background altogether
  - the light grey description text will be entirely unreadable against this, i'd wager
- give the highlighted area some more whitespace (that doesn't "jiggle" when you select something)
- full bleed the background to the edges
  - this may not be feasible, as it gets really complicated with nested fields
- use the "fat active blue left border", like notebook cells do
  - this might force the whole thing to have more whitespace, but as long as it's consistent and doesn't jiggle per keystroke, that's fine
    - if it _does_ jiggle, perhaps move the color border to the right of the item so that the typography lines up
      - and, yes, this is entirely unfair to other bidi languages



perhaps some of:
- clobber the huge whitespace to the left of the bullet so that the bullets line up with the field label/description
- maybe reduce the whitespace above the error items
- add a `pattern` to the schema so that we can see both "missing" and "doesn't match" errors, which might reveal _more_ whitespace issues

brichet:
> perhaps one of:
> * remove the background altogether
>   * the light grey description text will be entirely unreadable against this, i'd wager

Another option could be to have something more similar to the current *common tools*, with highlighting on the control (input/select) only, and not on the title and description.

> * clobber the huge whitespace to the left of the bullet so that the bullets line up with the field label/description
> * maybe reduce the whitespace above the error items

I agree

> * add a `pattern` to the schema so that we can see both "missing" and "doesn't match" errors, which might reveal _more_ whitespace issues

Can we have both these errors on the same field ?
We only display the errors for each field, and not the summary for the whole form if this is what you are referring to:
https://github.com/jupyterlab/jupyterlab/blob/fc3e02658c66a18001535651840c8448412b0b73/packages/metadataform/style/base.css#L23-L25

bollwyvl:
> highlighting on the control (input/select) only,

Sure, that works, but would introduce new whitespace issues, perhaps: it would probably look better fully wrapping around the control, but that would then bump up directly against the title above and below. 

It's a tricky balance between typographic harmony, interactivity, and accessibility.

> not the summary for the whole form 

Right, I'm assuming since that display is a list, it _could_ be more than one thing, even for a single field.

brichet:

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const input = event.target as HTMLInputElement;
+    if (event.key === 'Enter') {
+      this.addTag(props, input.value);
+    } else if (event.key === 'Escape') {
+      input.value = '';
+      input.style.width = '49px';
+    } else {
+      const tmp = document.createElement('span');
+      tmp.className = 'add-tag';
+      tmp.innerHTML = input.value;
```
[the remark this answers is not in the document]
bollwyvl: This is probably legit: the schema definitly allows for a tag like `foo><script>alert('pwnd')</script>`.

Also: we've lost the "Cell tags" label.

brichet:
kicking the CI

brichet:

packages/celltags-extension/src/celltag.tsx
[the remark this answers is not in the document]
brichet: Thanks, it's solved.

brichet:
Kicking the Ci

fcollonval (changes requested):
Thanks @brichet 

It seems that there is two regressions in the information displayed:
- The execution count placeholder is not displayed left of the active cell content
- The _Raw NBConvert Format_ is missing

Are those related to asynchronous tasks not yet done?

Could you remove the metadata editors?

Let's get that in by Tuesday :wink:

bollwyvl:
Removing the JSON editors is a bridge too far, I feel. They should
definitely not be visible (or rendered or updated per keystroke) by
default, but given 0 extensions support the new mechanism yet, and rjsf
can't do arbitrary JSON, users need the escape hatch.

This could follow the pattern of settings, and an additional side bar,
which only contained the two editors (cell, notebook) and their button bars.

brichet:
Thanks @fcollonval 

> * The _Raw NBConvert Format_ is missing

The field is not missing, it is only displayed when a raw cell is selected. I should have mention it. Otherwise it was always displayed but unusable for other cells.



> * The execution count placeholder is not displayed left of the active cell content

You're right, it is not updated after the execution, only when the cell is selected. As you said it must be related to the task not done yet.

brichet:
> Removing the JSON editors is a bridge too far, I feel

To clarify, the `MetadataEditorTool` would not be removed (this is the one used also with this PR).
The class not used anymore are:
https://github.com/jupyterlab/jupyterlab/blob/a920f332d0b12d084588f672af3926fc7d31163d/packages/notebook/src/notebooktools.ts#L722
https://github.com/jupyterlab/jupyterlab/blob/a920f332d0b12d084588f672af3926fc7d31163d/packages/notebook/src/notebooktools.ts#L774

brichet:
The failing test is not related to this PR, it failed 2 times about debugger panel for 2 different reasons.

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+        return [...allTags, ...tags];
+      },
+      []
+    );
+    return [...new Set(allTags)].filter(tag => tag !== '');
+  }
+
+  private _emptyAddTag(target: HTMLInputElement) {
+    target.value = '';
+    target.style.width = '49px';
```
andrii-i, line 56: Should either css variable or relative length unit like `em` be used here to make sure width of the element is responsive (to for example work area or viewport width changes)?

Should we opt for class names + css sheets vs inline styling? CC @krassowski

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+      this._emptyAddTag(input);
+    }
+  }
+
+  private _onChange(event: React.ChangeEvent<HTMLInputElement>) {
+    if (!event.target.value) {
+      this._emptyAddTag(event.target);
+    } else {
+      const tmp = document.createElement('span');
+      tmp.className = 'add-tag';
```
andrii-i, line 92: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+      tmp.textContent = event.target.value;
+      // set width to the pixel length of the text
+      document.body.appendChild(tmp);
+      event.target.style.width = tmp.getBoundingClientRect().width + 8 + 'px';
```
andrii-i, line 96: Should either css variable or relative length unit like em be used here to make sure width of the element is responsive (to for example work area or viewport width changes)?

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+        {allTags &&
+          allTags.map((tag: string, i: number) => (
+            <div
+              key={i}
+              className={`tag ${
+                props.formData.includes(tag) ? 'applied-tag' : 'unapplied-tag'
+              }`}
+              onClick={() => this._onTagClick(props, tag)}
+            >
+              <div className="tag-holder">
```
andrii-i, line 145: Should all caps global consts instead of inline CSS class strings be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    marginLeft="5px"
+                    marginRight="-3px"
+                  />
+                )}
+              </div>
+            </div>
+          ))}
+        <div className="tag unapplied-tag">
+          <div
+            className="tag-holder"
```
andrii-i, line 163: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    height="18px"
+                    width="18px"
+                    marginLeft="5px"
+                    marginRight="-3px"
+                  />
+                )}
+              </div>
+            </div>
+          ))}
+        <div className="tag unapplied-tag">
```
andrii-i, line 161: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+          <div
+            className="tag-holder"
+            onMouseDown={(e: React.MouseEvent<HTMLElement>) =>
+              this._onAddTagClick(props, e)
+            }
+          >
+            <input
+              className="add-tag"
```
andrii-i, line 169: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/celltags-extension/src/celltag.tsx
```diff
@@ -0,0 +1,204 @@
...
+              type="text"
+              style={{ width: '49px' }}
```
andrii-i, line 171: Should either css variable or relative length unit like em be used here to make sure width of the element is responsive (to for example work area or viewport width changes)?

Also, should we opt for class names + css sheets vs inline styling? CC @krassowski

packages/notebook-extension/src/tool-widgets/activeCellToolWidget.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * ## Note
+ * This field does not work as other metadata form fields, as it does not update metadata.
+ */
+export class ActiveCellTool extends NotebookTools.Tool {
+  constructor(options: Private.IOptions) {
+    super();
+    const { languages } = options;
+    this._tracker = options.tracker;
+
+    this.addClass('jp-ActiveCellTool');
```
andrii-i, line 44: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/notebook-extension/src/tool-widgets/activeCellToolWidget.tsx
```diff
@@ -0,0 +1,110 @@
...
+    this.layout = new PanelLayout();
+
+    this._inputPrompt = new InputPrompt();
+    (this.layout as PanelLayout).addWidget(this._inputPrompt);
+
+    // First code line container
+    const node = document.createElement('div');
+    node.classList.add('jp-ActiveCell-Content');
```
andrii-i, line 52: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/notebook-extension/src/tool-widgets/activeCellToolWidget.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const container = node.appendChild(document.createElement('div'));
+    const editor = container.appendChild(document.createElement('pre'));
+    container.className = 'jp-Cell-Content';
```
andrii-i, line 55: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/notebook-extension/src/tool-widgets/activeCellToolWidget.tsx
```diff
@@ -0,0 +1,110 @@
...
+    (this._cellModel?.sharedModel as ISharedText).changed.connect(
+      this.refresh,
+      this
+    );
+    this._cellModel?.mimeTypeChanged.connect(this.refresh, this);
+    this.refresh()
+      .then(() => undefined)
+      .catch(() => undefined);
+    return (
+      <div className="cell-tool">
```
andrii-i, line 95: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  render(props: FieldProps): JSX.Element {
+    const notebook = this._tracker.currentWidget;
+    this.editor.source = notebook
+      ? new ObservableJSON({ values: notebook.model?.metadata as JSONObject })
+      : null;
+    this.editor.source?.changed.connect(this._onSourceChanged, this);
+
+    return (
+      <div className="cell-metadata-editor">
```
andrii-i, line 118: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

andrii-i (changes requested):
Thank you, @brichet. Please see some suggestions on the use of consts for css classes instead on inline strings, on necessity of use of css variables or adaptive units, on inline styling

brichet:
Thanks @andrii-i for all these catching up, I'll fix it.

packages/celltags-extension/src/celltag.tsx
brichet: Sure, I'll add a class for empty input.

packages/celltags-extension/src/celltag.tsx
brichet: The `getBoundingClientRect()` method  returns values in pixels.
I'm not sure there is a simple way to convert pixels to em on the flight.

packages/celltags-extension/src/celltag.tsx
brichet: Added a `jp-CellTags-Empty` class and replaced the `width: '49px'` by a `width: '4em'` in css sheets.

packages/celltags-extension/src/celltag.tsx
brichet: Added a `jp-CellTags-Empty` class and replaced the `width: '49px'` by a `width: '4em'` in css sheets.

packages/celltags-extension/src/celltag.tsx
fcollonval: Setting size related styles to support some complex computed layout is ok (this is actually what widget is doing heavily).

In this specific case this allows to adapt to user input and is per nature transient. The only enhancement I can see is maybe to use a CSS property instead of an hard coded `8px` value. Something like:

`calc(${tmp.getBoundingClientRect().width} + var(--jp-add-tag-extra-width))`

packages/celltags-extension/src/celltag.tsx
andrii-i: Thank you @brichet

packages/celltags-extension/src/celltag.tsx
andrii-i: Thank you @brichet

packages/celltags-extension/src/celltag.tsx
andrii-i: I see, thank you both for explaining this. I think @fcollonval's suggestion is a good idea but please feel free to choose what do you think works best

andrii-i (approved):
Thank you @brichet for implementing suggested changes, this PR looks good to me.

(please feel free to deal with [this](https://github.com/jupyterlab/jupyterlab/pull/14026/files#r1123308323) the way you think works best)

packages/celltags-extension/src/celltag.tsx
brichet: Thanks @fcollonval for the suggestion, I updated it.

fcollonval (approved):
Thanks @brichet and thanks for reordering the doc

PR #14028, opened 2023-02-17
Include source in dists

### LINKED ISSUE #7851: Either not include source map, or include source in dist
Currently, building against RC services gives a bunch of webpack build warnings:
```
WARNING in C:/dev/nbdime/node_modules/@jupyterlab/services/lib/builder/index.js
Module Warning (from C:/dev/nbdime/node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../../src/builder/index.ts': Error: Can't resolve '../../src/builder/index.ts' in '[...]\nbdime\node_modules\@jupyterlab\services\lib\builder'
```

vidartf:
## References

Fixes #7851 .

## Code changes

Includes files in ./src in npm package dist, so that the relative paths in the built JS source maps can be resolved. Adds an integrity check for the same, so that it will also be enforced for packages in the future.

## User-facing changes

None (slightly larger dist on disk, but not loaded in browser unless debugging).

None

### DIFF
--- a/builder/package.json
+++ b/builder/package.json
@@ -24,7 +24,8 @@
     "metadata_schema.json",
     "lib/*.d.ts",
     "lib/*.js.map",
-    "lib/*.js"
+    "lib/*.js",
+    "src/**/*.{ts,tsx}"
   ],
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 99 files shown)


vidartf:
Haha, I broke the PR auto-labeler: 
```
Validation Failed. Issues cannot have more than 100 labels:
```

PR #14034, opened 2023-02-18
Update react-json-tree, replace react-highlighter with react-highlight-words

bollwyvl:
## References

- part of #13928

## Code changes

- [x] update `react-json-tree`
- [x] replace unmaintained `react-highlighter` with the (slightly more) maintained `react-highlight-words`
  - [x] removes browser-side duplicates of `escape-string-regexp` and `color-name`
- [x] apply `jp-mod-selected` to highlighted words, use lab vars  
- [x] lazily load json viewer react component
- [x] resolve `yarn.lock`

## User-facing changes

- site should start ever-so-slightly faster 
  - shaves another 67kb (tree) + 18kb (highlither) (gzip, 20kb + 6kb) off the critical path to clearing moon animation
- search results will now match high-contrast themed search results (instead of just hard-coded `yellow`)
  - image
  - image

## Backwards-incompatible changes

- n/a

## Future work

This is still a very unsatisfying experience for structured data editing and viewing.

The current react-based solution is:

- slow
- can't JSON edit
- is not aware of JSON Schema
- can't be deep-linked

A significant overhaul of this should be a part of JupyterLab, including:
- lazy-loaded support for other JSON-compatible files like yaml, toml, json5 (ick)
- JSON schema validation

A customized codemirror-6 based editor with code folding and schema-based linting probably makes the most sense.

Issue #13928, opened 2023-02-05
Minimize duplicate packages delivered to browser


bollwyvl:
### Problem

A current build of `master` reports:

```
WARNING in color-name
  Multiple versions of color-name found:
    1.1.3 /~/color-convert/~/color-name from /~/color-convert/conversions.js
    1.1.4 /~/color-name from /~/color-string/index.js

WARNING in escape-string-regexp
  Multiple versions of escape-string-regexp found:
    1.0.5 /~/escape-string-regexp from /~/react-highlighter/lib/highlighter.js
    4.0.0 /~/sanitize-html/~/escape-string-regexp from /~/sanitize-html/index.js
```

A number of packages are _not_ found through tools as they aren't _technically_ the same packages:
- `lodash`, `lodash-es`, `lodash.*`

### Proposed Solution

For the named ones, probably update some `dependencies` and/or `resolutions`.

For `lodash`: perhaps there are some webpack tricks we can apply. For example, `lodash.mergeWith` (used in a couple places) could be coaxed into being answered with `lodash/mergeWith`, saving 35k (unzipped) on the wire on every page load.

### Additional context

Discovered during #13927

MMesch:
@bollwyvl the triaging agent thinks this issue may be resolved. Given its age I think that could be true although PR #14034 seems to address only parts of the above mentioned duplication. Could you quickly comment on this issue's relevance and what is left to be done?

PR #14035, opened 2023-02-18
Deduplicate lodash, typescript, prettier

bollwyvl:
## References

- continues #13928
- includes #14034

## Code changes

- [x] add `typescript`  and `prettier` to resolutions, saving ~80mb in `node_modules`
  - > of course, this is silly vs the long poles of `canvas` (180mb) and `@stdlib` (196mb)
- [x] use webpack aliases to replace 
  - [ ] `lodash` with `lodash-es`
  - [ ] `lodash.*` with `lodash-es/*` (though this is done one-by-one instead of a fancy script)

## User-facing changes

- with this, and its upstream, the total assets when shipped is under 6.85mb (1.93mb when gzipped)

- ???


bollwyvl:
As this _does_ touch lsp (`lodash.mergeWith`) it may be worth looking at... 

Some more benchmarks to see if this is worth it:

| `master` | this pr |
|-|-|
| image | image

bollwyvl:
As rjsf uses lodash extensively, this is probably a real fail, suggesting this PR might not be worth it.

Deduplicating `typescript` and `prettier` probably _is_ worth it, and indeed, likely upgrading prettier.

PR #14037, opened 2023-02-19
Lazy load `@lumino/datagrid`

### LINKED ISSUE #14036: Unbundle `@lumino/datagrid` from core
### Problem

The lumino datagrid is used by a limited number of core features, but is 123kb (26kb gzip) that blocks the application from starting (e.g. starting the moon animation).

### Proposed Solution

- unbundle in webpack
```js
    optimization: {
      splitChunks: {
        chunks: 'all',
        cacheGroups: {
          jlab_core: {
            test: /[\\/]node_modules[\\/]@(jupyterlab|lumino(?!\/datagrid))[\\/]/,
            name: 'jlab_core'                            // ^^^^^^^^^^^^^^
          }
        }
      }
    },
```  
- `@jupyterlab/csvviewer`
  - rename `widget.ts` to `grid.ts`
  - in the new `widget.ts`, lazily-load `grid.ts` while waiting for `context.ready`
- `@jupyterlab/csvviewer-extension`
  - don't use any exported members from `@lumino/datagrid` (styles, mostly)
- `@jupyterlab/debugger`
  - maybe need to do something
 
### Additional context

There are probably a few other core features and renderers that can also come off the bottom line at application startup, but is the biggest outlier... certainly more impactful than nickle-and-diming for small wins like #14035 and #14034.

bollwyvl:
## References

- fixes #14036

## Code changes

- `dev_mode`
  - [x] unbundle `@lumino/datagrid` from the core distribution in webpack
- `@jupyterlab/csvviewer(-extension)`
  - [x] move loading of default style into grid widget
  - [x] delay loading and instantiation of DataGrid until actually requested (maybe on search, but meh)
- `@jupyterlab/debugger`
  - [x] create new isolation boundary for `Grid` in separate `gridpanel.ts` file
  - [x] delay instantiating grid until actually requested (by clicking on the table button in the variables panel) 

## User-facing changes

- every page load will load and parse 123kb less JS (27kb gzip) until needed
  - there may be a _slight_ delay the first time opening a new CSV or the debugger variable window

image

- the debugger `Grid` was previously not exported
  - it's _not_ hoisted to `index.ts`, so this doesn't really entail much of a change.


bollwyvl:
Looks like i'm hitting rate limits working on multiple PRs. Will take a break until some of them have cleared.

bollwyvl:
The UI fails seem unrelated to this (so far)... not sure if the grid will actually be exposed.

bollwyvl:
Hoo dang, all :heavy_check_mark: ! Haven't seen that in a while.

fcollonval (commented):
Thanks a lot @bollwyvl 

I got some small questions.

fcollonval (approved):
Thanks @bollwyvl

bollwyvl:
What is the baseline for the the benchmark numbers? I've seen them on a couple of recent PRs, but don't really have a referent frame...

fcollonval:
> What is the baseline for the the benchmark numbers? I've seen them on a couple of recent PRs, but don't really have a referent frame...

As we don't control the machine type nor load, the benchmark are run for the two branches (master and PR branches). So they should be interpreted as relative comparison without much thoughts about the absolute figures.

Memory leaks data are run only on the PR branches - the figures are unfortunately also flaky; except for the leaking collection. The tool used is [fuite](https://github.com/nolanlawson/fuite) with Puppeteer.

PR #14038, opened 2023-02-19
Get dev binder up and running again

### LINKED ISSUE #14033: Binder dev builds are broken slowing down review process
Reviewing JupyterLab PRs slowed down as the Binder environment does not work currently: while the binder badge gets added, the installation no longer boots up, e.g. https://mybinder.org/v2/gh/fcollonval/jupyterlab/fix%2F13972-use-back-runmenu-in-toolbar?urlpath=lab or https://mybinder.org/v2/gh/andrii-i/jupyterlab/closing_tabs_w_kb?urlpath=lab results in:

```
Spawn failed: Server at [...] didn't respond in 30 seconds.
```

On gitter @bollwyvl suggested that this might be related to our `start` file:

https://github.com/jupyterlab/jupyterlab/blob/master/binder/start

Edit: Community Forum post: https://discourse.jupyter.org/t/how-to-debug-server-at-didnt-respond-in-30-seconds-error/18042

bollwyvl:
## References

- fixes #14033 (sorta)

## Code changes

- `package.json`
  - [x] add `build:dev:prod:minimize:report`
    - [x] generate `dev_mode/static/webpack-bundle-analyzer.html`
- `binder/start`
  - [x] just remove it, rely on well-known config file locations
- `binder/jupyter_notebook_config.py`
  - [x] rename to `jupyter_config.py`
- `binder/environment.yml`
  - [x] add most of the dependencies from `pyproject.toml[dev,test]` to get better docker caching
- `binder/postBuild`
  - [x] create a `_reports_` folder
  - [x] explicitly call and log `yarn` up-front to reduce chances of deep-in-hatch failures during 200mb solve/download
    - [x] call with `--ignore-optional` and `--ignore-scripts` to avoid re-work about to be called below 
    - [x] copy the `webpack-bundle-analyzer.html` up to `_reports_` 
  - [x] call and log `build:dev:prod:minimize:report`
  - [x] call and log `pip install -e` with `-vv` to reveal potential issues 
  - [x] copy `jupyter_config.py` to well-known location
  - [x] call and log a ton of diagnostic commands up-front
    - > there's a lot of :red_envelope: for things like `jupyterlab_pygments`, widgets, etc.
  - [x] copy `jupyter-lab` on top of `jupyter-notebook` :vomiting_face: 
  - [x] clean up `~/.cache`, `node_modules/@stdlib` to save ~1GB

## User-facing changes

- badges displayed to PR authors and reviewers should work... more... than they do now
  - an optional follow-on could add _another_ badge which deep-linked into the `_reports_` folder

- n/a
  - unless you were using `postBuild` and `start` locally, which fairly clearly nobody was

## Future Work
- figure out a way to get more pre-compiled reports available in binder
  - discussed below: basically they just can't cost much
- collaboration features still doesn't appear to work
  - presumably someone that understands what's missing will be able to debug from a running binder, so needn't block this PR
- make better use of `lerna`, and explore `make -j8` or `doit -n8` to achieve parallelization of tasks, to be realized on developer machines/CI as well
- while I usually like to point at the `gzip` size for things, this is partially a lie, as it requires rather a bit of engineering to actually realize those benefits for end users
  - we should probably consider
    - generating production assets that are pre-compressed
      - this takes `dev_mode/static` down to 8.4mb vs 34mb
    - determining the right way in backends (e.g. `jupyter_server`) to serve these with the correct headers 
      - without blanket turning on the built-in `gzip` feature, which has security implications
    - providing tools for downstreams to do the same


krassowski (commented):
That's amazing! Would it be worth mentioning `_webpack_stats_.html` in [contributing documentation](https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/developer/contributing.rst)?

bollwyvl:
aw, you know me, love some PR creep. 

But seriously: we could even take it out, as it does force a second, even more expensive rebuild. But otherwise... folk just _don't look at these things_, so maybe it's worth the extra minute per build for the :eyeglasses: .

My problem with the report (which is, of course, quite the engineering marvel) is it doesn't do a very good job of showing the inter-relationships between bundles, and, crucially, what matters to the bottom line of loading the site. Maybe we can find some other reporting tool that would run (ideally in parallel) that would do a better job of that, but we _really_ don't want another browser in the mix... heck, if we could skip building galata in binder, and its 100s of megabytes of deps, I definitely would.

But yeah, if we _do_ keep the report, it should probably be mentioned... or perhaps better, added to a default workspace in the binder PR URL which could include all the reports we might want.

krassowski:
I thought about workspace but losing the report upfront comes with a cost and might be distracting.

If this takes about a minute, maybe it should be an extension which generates and opens the report? Could be a new launcher item and included in our dev binder?

bollwyvl:
> comes with a cost and might be distracting.

sure, it could just be a separate link to the same underlying binder e.g. 

```
Thanks for making a pull request to jupyterlab!
To try out this branch on binder, follow one of these links:

| production, minified                         | reports                        |
|----------------------------------------------|--------------------------------|
| [Binder]                                     | [Binder]                       |
| Open the launcher, close to how a user would | Open to reports about app size |
```

> extension which generates and opens the report?

I think the goal of the PR-oriented binder should be to deliver a product as close as what would be shipped to end users, which also provides as much additional, actionable information as possible. 

More extensions and complex startup config are more things to break, as the need for this PR demonstrates. Executing a production build is definitely _not_ a waste of time.

Further, the report generation itself isn't the issue, per se. It's the time it takes for the `build:prod:minimize` _after_ the uninstrumented baseline build that happens deep inside `hatch` without much configurability.  (see: anything related to galata). The "readable" js file names are kinda garbage anyway, and sourcemaps generally work in 2023.

binder/jupyter_config.py
```diff
@@ -43,3 +43,11 @@
 }
 
 c.ServerApp.log_level = logging.DEBUG
+
+c.LabApp.dev_mode = True
+c.LabApp.collaborative = True
```
fcollonval, line 48: The flag is not needed as there is a need for explicitly installing an additional package.

```suggestion
```

> And for now it will fail as while setting `collaborative`, the required test is failing: fix is there - https://github.com/jupyterlab/jupyterlab/pull/14047

fcollonval (commented):
Huge thanks @bollwyvl

binder/environment.yml
```diff
@@ -1,14 +1,45 @@
...
+  - packaging
+  - tornado >=6.2.0
+  - traitlets
+  # collaboration
+  - jupyter_server_fileid >=0.7.0
+  # extra conda stuff
+  - jsonschema-with-format-nongpl
+  # not yet on conda
+  - pip:
+    - jupyterlab-rtc >=0.0.5
```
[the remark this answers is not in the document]
bollwyvl: "about to be released" means "broken today"

binder/jupyter_config.py
bollwyvl: when it's released, we can fix it

binder/environment.yml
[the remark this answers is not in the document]
fcollonval: yes :wink:

binder/jupyter_config.py
fcollonval: Sorry I wasn't clear - that flag can already be removed as the collaboration code has been removed from core. The flag only exists to inform user (familiar with that flag in 3.x) that it should be done by installing another extension now.

bollwyvl: so what's the right play: take out _all_ collaboration-related stuff now, and do all of it on a follow-on PR?

bollwyvl: removed

binder/environment.yml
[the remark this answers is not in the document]
bollwyvl: removed

jtpio (approved):
Looks good on Binder, thanks!

image

bollwyvl:
Thanks all.

Issue #12889, opened 2022-08-01
Markdown cells break notebook trust


divyansshhh:
## Description
When I rsync a notebook (which contains atleast 1 markdown cell) from another host to my host, the notebook is not trusted by default which is expected since the signature wouldn't exist on my machine by default. 

But when I run `Trust Notebook` from the Command Palette, it trusts the notebook temporarily but when closing and reopening the notebook, the trust doesn't not persist and I have to re-trust the notebook. 

Investigating this, I found that this is caused because in the `NotebookActions.trust` function, we set `trusted: true` in all cells but only pop it out of code cells when computing signature. Hence, when computing the signature through the `Trust Notebook` command it adds an additional key to the signature which is not present when we write a notebook to disk and hence trust doesn't persist.

Relevant code refs - 
- `NotebookActions.trust`- https://github.com/jupyterlab/jupyterlab/blob/master/packages/notebook/src/actions.tsx#L2024
- `_check_cells` - https://github.com/jupyter/nbformat/blob/f712d60f13c5b168313222cbf4bee7face98a081/nbformat/sign.py#L499

## Reproduce

1. To simulate the rsync, we can delete the signature of any notebook (with 1 markdown cell) from the `~/.local/share/jupter/nbsignatures.db`.
2. Now, opening the notebook will show as untrusted (expected)
3. Run `Trust Notebook` from the Command Palette.
4. Close notebook
5. Reopen notebook

## Expected behavior
The trust should persist when reopening the notebook

## Context

- Operating System and version: Linux
- Browser and version: Chrome 103
- JupyterLab version: 3.4.3

divyansshhh:
I can raise a PR to fix this but I have some questions -
1. Is there a reason for trusting all cell? If not, I can fix it here.
2. If there is a usecase for trusting all cells, we can fix this in nbformat.

mlucool:
CC @Carreau

jweill-aws:
@davidbrochart has a PR that may be related: https://github.com/jupyterlab/jupyterlab/pull/12756

There is also an issue involving blank cells and notebook trust: https://github.com/jupyterlab/jupyterlab/issues/9765 (thanks @krassowski).

Carreau:
I merged a few hours a go a change in nbformat that will be much more verbose and unhappy if given an invalid notebook. I'll try to make a beta release soon but I want to make sure I do it properly as nbformat apparently is also on npm.

[...]

PR #14039, opened 2023-02-19
Minor trust improvements

krassowski:
## References

- https://github.com/jupyterlab/jupyterlab/issues/14025
- https://github.com/jupyterlab/jupyterlab/issues/12889
- https://github.com/jupyterlab/jupyterlab/issues/9765

## Code changes

Done:
- [x] Empty notebook gets populated with a **trusted** code cell
- [x] Test that serialization does not include `trusted` for non-code cells

In follow-up:
- [ ] Set user-created cells as trusted (but not if pasting untrusted cells)

### DIFF
--- a/packages/notebook/src/model.ts
+++ b/packages/notebook/src/model.ts
@@ -380,7 +380,9 @@ close the notebook without saving it.`,
 
     // Ensure there is at least one cell
     if ((copy.cells?.length ?? 0) === 0) {
-      copy['cells'] = [{ cell_type: 'code', source: '', metadata: {} }];
+      copy['cells'] = [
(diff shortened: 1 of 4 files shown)


jtpio:
Nice thanks @krassowski for starting this :+1: 

Since the current state already includes some fixes, it would also be fine to continue further improvements in follow-up PRs.

krassowski:
> Since the current state already includes some fixes, it would also be fine to continue further improvements in follow-up PRs.

Right, especially since setting user-created cells as trusted might require changes in `@jupyter/ydoc` so that would be waiting and accumulating merge conflicts. I un-derafted it.

jtpio (approved):
Thanks!

Issue #13189, opened 2022-10-05
Compress static assets served by Jupyter Lab


suganya-sk:
### Problem

Jupyterlab's static assets are marked in Chrome's Lighthouse tests for text compression, even when there are no extensions installed. With the prevalent push in the Jupyterlab community to improve performance, this might be a candidate worth looking into.

I have been trying to compress the static assets associated with a few custom extensions I have, which are pretty large in size and I noticed that the static assets served from `static/lab/` are flagged, even without any extensions. I'm wondering if compression might help here.

### Proposed Solution

Consider compressing the static assets served by Jupyterlab.

I dug through the build process of Jupyterlab a bit and noticed that compression is set to false [here](https://github.com/jupyterlab/jupyterlab/blob/e6599b59064b8b30e3ed72e0ed3a0ddc2e3d2e5c/dev_mode/webpack.prod.minimize.config.js#L26). Is there a specific reason this is so or does compression occur anywhere else? Any pointers to relevant discussions is helpful! I have a high-level understanding of the build process; pls redirect if I should be looking anywhere else.

krassowski:
I can reproduce this; it indeed includes assets such as JavaScript or settings JSON. It seems really important as the Lighhouse claims we could save many seconds on page load here, though the question is how much time the extra compression will take.

<details>

![Screenshot from 2022-10-06 17-32-14](https://user-images.githubusercontent.com/5832902/194369491-b5c078db-8590-4410-b58d-aff3e3effee5.png)

</details>

This can be addressed in `jupyter-server` by enabling `compress_response` in tornado request handler. I will check it out locally and open a PR if it works well. I don't know if there is a reason for it not being currently enabled, but we can ask over at [`jupyter-server`](https://github.com/jupyter-server/jupyter_server).

mlucool:
FWIW, we also have thought about https://github.com/jupyter-server/jupyter_server/issues/312#issuecomment-1050010913. This may be better to not compress in Jupyter Server and instead give us a way to "eject it" and put a high perfomrance webserver in front (e.g. nginx, caddy, apache). This server would take care of compression + serving.

Alternatively, we can precompress assets and if a client accepts a compression type server that with the right headers. That is, we gzip/brotli everything at a high level of compression and store the compressed file in memory/on disk. Then if a client accepts this type, we use that, if not we don't compress that assets or maybe just compress on the fly. This can be done for all extensions.

suganya-sk:
[...]

PR #14040, opened 2023-02-19
[PoC] Build and serve brotli-compressed assets

bollwyvl:
## References

- proof-of-concept for #13189
- includes #14038

## Code changes

- `dev_mode`
  - [x] adopt `compression-webpack-plugin`
  - [x] add to `minimize`
- `juptyerlab/labapp`
  - [x] add a configurable for serving pre-compressed static assets
  - [x] add an opt-out flag 
- `jupyterlab/handlers`
  - [x] add brotli-aware file finder

## User-facing changes

- _slashes_ total data on the wire to under 2mb
  - this PR (left) vs #14038 (right)
    - image
  - falls back to uncompressed data if the client doesn't accept that format

- n/a

## Future Work

- rework this so it doesn't break more tests
  - i don't know the relationship between `dev_mode` and `staging` anymore
- figure out how to package these different built assets
  - shipping both the `*.js` _and_ `*.js.br` seems expensive
    - same applies to source maps, which benefit _even more_ from compression, and are presently not included which is detrimental to everyone, as `jupyter lab build` can break lightweight CI envs
  - some a approaches:
    - once-per-install, CLI-based
      - `jupyter lab compress` which pre-generates all of the `.br` assets
        - this could also discover and compress extension assets, which would be **awesome**
      - ... but would need to be optional, as it would require a heavyweight `.whl` dependency for the `brotli` executable
        - or fall back to `gzip`, which would sacrifice 10-20% of the savings (no real data to back this up)
    - split packages
      - `jupyterlab` ships only the `.js` assets
      - `jupyterlab-debug` includes the `.map` assets
      - `jupyterlab-brotli` includes `.js.br` and `.map.br`
- investigate additional server-side delays
  -  looking at the waterfall charts, a _lot_ of time is spent in the "waiting for server" stage
    - this is more pronounced with these drastically smaller assets
    - it's possible this is due to synchronous `@classmethod` lookups on the file finding, inherited from `jupyter_server`, but ultimately `tornado`
      - this could possibly be delegated to an asynchronous pool


mlucool:
Re server side delays:

https://github.com/jupyter-server/jupyter_server/issues/312:
> Create a document for how to configure nginx/apache in front of Jupyter Server. This document should include tips of the right settings to skip the jupyter server for certain assets (e.g. anything in /static).

Of course you can pre-load all the assets in memory (on the order of a few hundred mb), but I think that's the best way for people really tune these for high performance is to allow an external web-server to serve static assets. If you are running/viewing notebooks on the same host, this isn't a big deal. If you are doing it cross-host (which implies a more advanced user), it should be possible to skip any python server (which can block) in favor of a high performance web server.

With nginx/apache/... you don't need to precompress as the right config settings does it on-the-fly. Still I would pre-compress as you can use a much higher level of compression than you would choose for on-the-fly (otherwise the compression time would dominate any savings).

bollwyvl:
Again, I keep hearing stuff about enterprise/managed settings where folks have the resources to co-deploy things. Getting letsencrypt working, installing and running `nginx`, and managing your firewall on a windows laptop is well and out of scope for many users.

On this repo, looking for means to improve and instrument the experience for _any_ JupyterLab user, including contributors, reviewers, and those just _using_ jupyterlab locally on the desktop, with a minimum of extra work, while maintaining the live configurability of the system, is the best we can do. Even locally, serving the compressed files is noticeably faster. 

Getting some of the bottlenecks out of the way on the python side is almost certainly doable, without resorting to running a pile of extra services.

mlucool:
Totally understand your point. Did you try using a single javascript bundle? While untested, I suspect without HTTP2 it may noticeably improve performance issues as HTTP 1.1 has head of line blocking as a design flaw. The assets produced here are tailored towards small files which is what HTTP2 excels at. This would make one file read + pay the waiting penalty once (and avoid any issues potentially from the GIL adding more serialization). This strategy would very likely do worse in settings with HTTP2, but if you want to assume no certs + no advanced web servers; it seems worth trying if you find local setups slow.

See more on HTTP 1.1 optimizations https://stackoverflow.com/a/30864259.

For the avoidance of doubt, I very much agree with pre-compressing assets before shipping python packages and think this idea is great for lab + all extensions. It would help with both HTTP 2 and 1.1.

bollwyvl:
Unlikely anything will come of this: likely needs to be handled at the `jupyter_server` level.

PR #14043, opened 2023-02-20
Fix JupyterLab command on Gitpod

jtpio:
## References

There were some warnings due to the update to Jupyter Server 2.

Also related to https://github.com/jupyterlab/jupyterlab/issues/13726

A follow-up would be to install the `jupyter-collaboration` package when it's published: https://github.com/jupyterlab/jupyter_collaboration

## Code changes

Gitpod config update:

- [x] Update traits used in the `jupyter lab` command
- [x] Remove the `--collaborative` flag

## User-facing changes

None

None


jtpio:
Merging so it can be used for new Gitpod workspaces.

Issue #14042, opened 2023-02-20
Show file size in the file browser


jtpio:
### Problem

Current the JupyterLab File Browser does not show the size of the files:

![image](https://user-images.githubusercontent.com/591645/220053375-480ecdb7-56af-4923-be49-db6b777bd381.png)

While the Classic Notebook does:

![image](https://user-images.githubusercontent.com/591645/220053253-f492c3d4-3a01-4742-8688-a5a7b30b091b.png)

### Proposed Solution

Add an option to show the file size in the file browser, similar to the one for enabling check boxes:

![image](https://user-images.githubusercontent.com/591645/220053797-ba33f1f0-776f-4824-9e9f-d23ef5e227d6.png)

### Additional context

Opening a new issue because searching for a similar issue with the following filters didn't yield any result: https://github.com/jupyterlab/jupyterlab/issues?q=is%3Aopen+is%3Aissue+label%3A%22tag%3AFeature+Parity%22+label%3Apkg%3Afilebrowser

Relevant for https://github.com/jupyter/notebook/issues/6397

PR #14044, opened 2023-02-20
Add support for showing the file size in the File Browser

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14042

## Code changes

Follow the same logic as for the last modified column.

A follow-up would be to make the handling of hidden columns more generic but that might require extra refactoring work and delay delivering this feature before 4.0.

- [x] Make the "File Size" column hidden by default
- [x] Make it possible to show the "File Size" column via the settings and the context menu
- [x] Use `B` instead of `Bytes` for consistency
- [x] Allow sorting on the file size

## User-facing changes

Users will be able to show the file size column (opt-in):

https://user-images.githubusercontent.com/591645/220101734-72768e0f-1146-4c5c-b093-91f4303448a5.mp4

Downstream applications like Notebook 7 can then enable the column by default: https://github.com/jupyter/notebook/issues/6397

None

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1997,7 +1997,7 @@ export namespace DirListing {
       fileSize.classList.add(FILE_SIZE_ID_CLASS);
       narrow.classList.add(NARROW_ID_CLASS);
       narrow.textContent = '...';
-      if (!hiddenColumns?.has?.('is_selected')) {
+      if (!hiddenColumns?.has('is_selected')) {
         const checkboxWrapper = this.createCheckboxWrapperNode({
           alwaysVisible: true
         });
@@ -2008,13 +2008,13 @@ export namespace DirListing {
       node.appendChild(modified);
       node.appendChild(fileSize);
 
-      if (hiddenColumns?.has?.('last_modified')) {
+      if (hiddenColumns?.has('last_modified')) {
         modified.classList.add(MODIFIED_COLUMN_HIDDEN);
       } else {
         modified.classList.remove(MODIFIED_COLUMN_HIDDEN);
       }
 
-      if (hiddenColumns?.has?.('file_size')) {
+      if (hiddenColumns?.has('file_size')) {
         fileSize.classList.add(FILE_SIZE_COLUMN_HIDDEN);
       } else {
         fileSize.classList.remove(FILE_SIZE_COLUMN_HIDDEN);
@@ -2145,7 +2145,7 @@ export namespace DirListing {
       text.className = ITEM_TEXT_CLASS;
       modified.className = ITEM_MODIFIED_CLASS;
       fileSize.className = ITEM_FILE_SIZE_CLASS;
-      if (!hiddenColumns?.has?.('is_selected')) {
+      if (!hiddenColumns?.has('is_selected')) {
         const checkboxWrapper = this.createCheckboxWrapperNode();
         node.appendChild(checkboxWrapper);
       }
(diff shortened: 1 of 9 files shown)


jtpio:
Need to look into some visual regressions:

image

packages/filebrowser/src/listing.ts
```diff
@@ -1987,13 +2006,20 @@ export namespace DirListing {
       node.appendChild(name);
       node.appendChild(narrow);
       node.appendChild(modified);
+      node.appendChild(fileSize);
```
krassowski, line 2009: :thinking: performance-wise the approach taken by checkboxes seems better, maintenance-wise the simpler the better. Any benefit of not rendering these will only slightly increase the maximal bearable size of the directory before it starts lagging and we could save much more by lazy-rendering and using content-visibility CSS - so I think it is fine to proceed as is.

jtpio:
CI is now passing, marking as ready for review.

packages/filebrowser/src/listing.ts
jtpio: Right, while working on this it looked like the file browser could benefit from some refactoring especially in how the hidden columns are handled.

Since I'm not sure we will want to add more columns in the near future I would also say it's fine to leave it as is for now to not delay the merge before the 4.0 feature freeze.

We can open an issue to track further improvements, which could happen during the beta and rc cycles if time permits.

jtpio:
Rebasing so it can be tested in Binder now that https://github.com/jupyterlab/jupyterlab/pull/14038 is merged.

packages/filebrowser/src/listing.ts
```diff
@@ -2106,6 +2165,13 @@ export namespace DirListing {
       } else {
         modified.classList.remove(MODIFIED_COLUMN_HIDDEN);
       }
+
+      if (hiddenColumns?.has?.('file_size')) {
```
fcollonval, line 2169: I don't know why that `?` was added in the previous test. But it should exist if `hiddenColumns` is defined, isn't it?

```suggestion
      if (hiddenColumns?.has('file_size')) {
```

packages/filebrowser/src/listing.ts
```diff
@@ -2196,6 +2263,12 @@ export namespace DirListing {
         modified.classList.remove(MODIFIED_COLUMN_HIDDEN);
       }
 
+      if (hiddenColumns?.has?.('file_size')) {
```
fcollonval, line 2266:
```suggestion
      if (hiddenColumns?.has('file_size')) {
```

fcollonval (approved):
Thanks

packages/filebrowser/src/listing.ts
jtpio: Thanks, yeah not sure why it was like that.

I pushed a commit to remove that on the whole file.

packages/filebrowser/style/base.css
```diff
@@ -149,6 +153,12 @@
   text-align: right;
 }
 
+.jp-DirListing-headerItem.jp-id-filesize {
+  flex: 0 0 70px;
```
krassowski, line 157:
```suggestion
  flex: 0 0 75px;
```

`72px` was the minimum for me to avoid this problem:

Screenshot from 2023-02-21 21-22-16

but I suggest adding 3 more pixels to be on the safe side.

packages/filebrowser/style/base.css
```diff
@@ -278,6 +288,11 @@
   text-align: right;
 }
 
+.jp-DirListing-itemFileSize {
+  flex: 0 0 84px;
```
krassowski, line 292:
```suggestion
  flex: 0 0 90px;
```

(needed when increasing `.jp-DirListing-headerItem.jp-id-filesize`)

packages/filebrowser/style/base.css
krassowski: (happy to meet in the middle)

krassowski (commented):
One minor small style suggestion.

Is it worth adding a galata test with chekboxes and file sizes shown? I am not 100% sure but I think we do not have it under test. It could reuse [this snippet](https://github.com/jupyterlab/jupyterlab/blob/305976a74aba4e2aed176633000a0588a2641a8d/galata/test/jupyterlab/sidebars.test.ts#L45-L53)

jtpio:
> I am not 100% sure but I think we do not have it under test

I am also not sure, because this PR adds a couple of styling rules for the file size column.

> It could reuse [this snippet](https://github.com/jupyterlab/jupyterlab/blob/305976a74aba4e2aed176633000a0588a2641a8d/galata/test/jupyterlab/sidebars.test.ts#L45-L53)

Maybe we can try adding a click on "Show File Size Column" in the same test.

jtpio:
OK CI is now passing.

krassowski (approved):
Thank you!

Issue #10284, opened 2021-05-23
Show a concise summary of keyboard shortcuts when a user hits the "h" key (parity with classic)


telamonian:
### Feature request

recent ref: #10270

I really miss the simple command cheat sheet from "classic":

![image](https://user-images.githubusercontent.com/2263641/119251581-39c96600-bb75-11eb-8d38-f2674add3ebc.png)

The simple clarity of the "Keyboard shortcuts" modal, plus the ease of bringing it up from within the app with a single <kbd>h</kbd> keypress, is a large part of what got me into IPython Notebook in the first place.

We should bring the "Keyboard shortcuts" modal back in lab. It should be more compact and summarized than eg https://github.com/jupyterlab/jupyterlab-shortcutui, something closer in spirit and design to the original modal from classic.

telamonian:
I do think we should keep the "open help modal on <kbd>h</kbd>" behavior from classic, and I also kind of think that maybe <kbd>h</kbd> should not be remappable (at least in notebook command mode). In general I think it's a good idea to ensure that the help screen is always accessible in the expect way.

@afshin Do we have a precedent for a non-remappable shortcut? Or have we avoided that pattern so far?

goanpeca:
This would indeed help @telamonian. I have a question. Would the idea be to have a modal, per widget (if provided)? I can see this modal being useful if one is in the Notebook, but if I was in the file explorer/terminal, I would expect different shortcut helps to appear.

telamonian:
> I would expect different shortcut helps to appear

That is likely the sticking point that's kept this issue in stasis for the last ~4 years. There is, of course, the currently unused "contextual help". So we could just populate that with context-dependent man-ish pages. Currently, though, "contextual help" opens as a main area widget. This is less than ideal since it will switch away from whatever tab you were actually trying to get help with, as well as completely obscuring it.

So yeah, something like:

"contextual help" but as a sidebar or modal + populating said help screen with a concise summary of the notebook keyboard shortcuts when a notebook is focused.

If that's reasonably successful, we can start thinking about how to populate the help screen in other non-notebook contexts.

krassowski:
+1 for having it as a modal, just like the command palette in 3.0 + having an option to place it in the sidebar instead (just like the command palette). The widget could use a flexbox to show one row in sidebar or two rows when in modal.

telamonian:
> one row

You mean column, right? But yeah, that sounds good

fperez:
[...]

Issue #4004, opened 2018-02-25
Help > 'Keyboard Shortcuts'


westurner:
In Jupyter Notebook, I can go to Help > 'Keyboard Shortcuts'. This seems to be missing in JupyterLab.

- [ ] JupyterLab: Add Help > 'Keyboard Shortcuts' 
- [ ] Jupyter Notebook docs: Include a table of (default) keyboard shortcuts (that I can find by searching for 'keyboard' or 'keyboard shortcuts')

I was able to find my Keyboard Shortcuts with:
-  Help > Launch classic notebook |> Help > Keyboard Shortcuts.

saulshanabrook:
Adding a menu item (as suggested here https://github.com/jupyterlab/jupyterlab/issues/2976#issuecomment-331064686) makes sense to me!

breckuh:
Not sure if someone else is working on this, but I just took a first pass.

Working code and a screenshot is here: https://github.com/breckuh/jupyterlab/commit/dfdc6f2220a93c954a8936051f5c7cb3151d6935

saulshanabrook:
@breckuh That looks great! Could you [open a PR](https://help.github.com/articles/creating-a-pull-request/) for that branch so we can continue discussion there?

westurner:
That looks nice. From reading the other linked issues, other
features/criteria mentioned include:

- A thematic organization
  - This would need to be updated for every new command; whereas listing
all bound commands is simpler but maybe not as useful

- Printable
  - Can this list be opened in a new JupyterLab tab, browser tab/window?
And printed?
  - Maybe a toggle between single and multiple columns would be most usable
and printable?

breckuh:
> @breckuh That looks great! Could you [open a PR](https://help.github.com/articles/creating-a-pull-request/) for that branch so we can continue discussion there?

As you mentioned in the other thread #2976 https://github.com/jupyterlab/jupyterlab-shortcutui seems to cover this!

BrenBarn:
Was this ever resolved?  The last comment I see is apparently about an extension that lets you look up shortcuts.

The old Jupyter had a button you could press to see a list of keyboard shortcuts.  If Jupyterlab doesn't have one, that's a regression from a user's perspective.  Having to install an extension or look through raw JSON in "advanced settings" is not reasonable for someone who is just trying to learn keyboard shortcuts.

jasongrout:
[...]

PR #11960, opened 2022-01-29
Display shortcuts

jasongrout:
## References

This takes steps toward addressing https://github.com/jupyterlab/jupyterlab/issues/10284

## Code changes

Determine relevant keybindings for the current active element (taking into account the convention that matching stops when an element matches [data-lm-suppress-shortcuts] or [data-p-suppress-shortcuts]) and display them in the console. This is a step towards showing the user all relevant keybindings for wherever they are.

We still need to display these bindings in a way the user will see.

We also see that some existing core keybindings do not have labels, and others do not have captions. This needs to be fixed as well.

In order to activate this command, add this user keyboard shortcut:

{command: "apputils:display-shortcuts",
 keys: ["Ctrl Shift H"],
 selector: "body"}

PR #14053, opened 2023-02-20
Display shortcuts

HaudinFlorence:
## References

Supersedes PR from https://github.com/jupyterlab/jupyterlab/pull/11960

Fixes https://github.com/jupyterlab/jupyterlab/issues/4004
FIxes https://github.com/jupyterlab/jupyterlab/issues/10284
Screenshot from 2023-02-23 22-11-10
## Code changes

Changes are mainly style changes for the display of the shortcuts, also changes on the level of details/ the choice of informations to be displayed.


packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,148 @@
...
+      targ = targ.parentElement, ++dist
+    ) {
+      if (targ.hasAttribute('data-lm-suppress-shortcuts')) {
+        return -1;
+      }
+      /* <DEPRECATED> */
+      if (targ.hasAttribute('data-p-suppress-shortcuts')) {
+        return -1;
+      }
+      /* </DEPRECATED> */
```
krassowski, line 45: Safe to remove, we dropped `p-` prefixes in Lumino 2.0 the current implementation [here](https://github.com/jupyterlab/lumino/blob/d08b6f71828438dd7bf1ba7f59ea706a95680822/packages/commands/src/index.ts#L1595) does not include it.

```suggestion
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <th style={{ textAlign: 'left' }}>Shortcut</th>
+          <th style={{ textAlign: 'left' }}>Label</th>
+        </tr>
+      </thead>
+      <tbody>{bindingTable}</tbody>
+    </table>
+  );
+
+  return showDialog({
+    title: 'Keyboard Shortcuts',
```
krassowski, line 139:
```suggestion
    title: trans.__('Keyboard Shortcuts'),
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,148 @@
...
+      bindingTable.push(<tr style={{ height: '2em' }}></tr>);
+    }
+  }
+
+  const body = (
+    <table>
+      <thead>
+        <tr>
+          <th style={{ textAlign: 'left' }}>Shortcut</th>
+          <th style={{ textAlign: 'left' }}>Label</th>
```
krassowski, line 131:
```suggestion
          <th style={{ textAlign: 'left' }}>{ trans.__('Shortcut') }</th>
          <th style={{ textAlign: 'left' }}>{ trans.__('Label') }</th>
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,148 @@
...
+  // Display shortcuts by group
+  const bindingTable = [];
+  for (let d = 0; d <= maxDistance; d++) {
+    if (groupedBindings.has(d)) {
+      console.log(`Binding level ${d}:`);
+      bindingTable.push(
+        groupedBindings.get(d)!.map(b => (
+          <tr key={b.command}>
+            <td
+              style={{
```
krassowski, line 104: We should prefer defining styles in CSS files to enable consistent theming and use our CSS variables whenever possible (font size, colour, etc).

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,148 @@
...
+      groupedBindings.set(distance, []);
+    }
+    groupedBindings.get(distance)!.push(binding);
+  }
+
+  // Display shortcuts by group
+  const bindingTable = [];
+  for (let d = 0; d <= maxDistance; d++) {
+    if (groupedBindings.has(d)) {
+      console.log(`Binding level ${d}:`);
```
krassowski, line 99:
```suggestion
```

packages/apputils-extension/src/shortcuts.tsx
krassowski: It feels to me that we should just change signature of `targetDistance` in lumino, expose it publically and reuse here. This can be done in a follow up.

packages/apputils-extension/package.json
```diff
@@ -53,11 +53,13 @@
...
     "@lumino/coreutils": "^2.0.0-beta.0",
     "@lumino/disposable": "^2.0.0-beta.1",
+    "@lumino/domutils": "^2.0.0-beta.0",
     "@lumino/polling": "^2.0.0-beta.0",
     "@lumino/widgets": "^2.0.0-beta.1",
+    "es6-promise": "~4.2.8",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-toastify": "^9.0.8"
+    "react-toastify": "~9.1.1"
```
krassowski, line 62: Spurious changes?

krassowski (commented):
Thank you for picking this up! Some quick feedback.

packages/apputils-extension/src/shortcuts.tsx
HaudinFlorence: @krassowski Thanks for the comments. They should be taken into account in 
a72ed0c.

packages/apputils-extension/package.json
```diff
@@ -53,11 +53,13 @@
     "@lumino/commands": "^2.0.0-beta.1",
     "@lumino/coreutils": "^2.0.0-beta.0",
     "@lumino/disposable": "^2.0.0-beta.1",
+    "@lumino/domutils": "^2.0.0-beta.0",
     "@lumino/polling": "^2.0.0-beta.0",
     "@lumino/widgets": "^2.0.0-beta.1",
+    "es6-promise": "~4.2.8",
```
fcollonval, line 59: Not needed any longer

```suggestion
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+      bindingTable.push(<tr style={{ height: '2em' }}></tr>);
+    }
+  }
+
+  const body = (
+    <table>
+      <thead>
+        <tr>
+          <th style={{ textAlign: 'left' }}>{trans.__('Shortcut')}</th>
+          <th style={{ textAlign: 'left' }}>{trans.__('Label')}</th>
```
fcollonval, line 114: Could you use CSS classes to set the style here too?

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+        groupedBindings.get(d)!.map(b => (
+          <tr className="jp-ContextualShortcut-TableRow" key={b.command}>
+            <td className="jp-ContextualShortcut-TableItem">
+              {b.keys.map(CommandRegistry.formatKeystroke).join(', ')}
+            </td>
+            <td>{commands.label(b.command, b.args)}</td>
+          </tr>
+        ))
+      );
+      bindingTable.push(<tr style={{ height: '2em' }}></tr>);
```
fcollonval, line 105: could you use CSS classes here too?

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+    activeBindings.set(formatted, [distance, kb]);
+  }
+
+  // Group shortcuts by distance
+  let maxDistance = -1;
+  const groupedBindings = new Map<number, CommandRegistry.IKeyBinding[]>();
+  for (let [distance, binding] of activeBindings.values()) {
+    maxDistance = distance > maxDistance ? distance : maxDistance;
```
fcollonval, line 84: To ease understanding:

```suggestion
    maxDistance = Math.max(distance, maxDistance);
```

packages/apputils-extension/style/base.css
```diff
@@ -6,3 +6,26 @@
...
+  border-width: '1px';
+  font-size: '11.7px';
+  font-family: monospace;
+}
+
+.jp-ContextualShortcut-TableRow {
+  vertical-align: 'top';
+  font-size: '13px';
+  font-family: Arial, Helvetica, sans-serif;
+}
```
fcollonval, line 31: Could you move those in a specific file like `shortcuts.css` to better structure the code? Then here you can add `@import './shortcuts.css'`.

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+        return dist;
+      }
+    }
+    return -1;
+  }
+
+  function formatKeybinding(keys: readonly string[]): string {
+    return keys.map(CommandRegistry.formatKeystroke).join(', ');
+  }
+
```
fcollonval, line 52: Since https://github.com/jupyterlab/lumino/pull/433, we should directly use Lumino method for consistency.

```suggestion
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+    string,
+    [number, CommandRegistry.IKeyBinding]
+  >();
+  for (let i = 0; i < commands.keyBindings.length; i++) {
+    const kb = commands.keyBindings[i];
+    let distance = matchDistance(kb.selector, elt);
+    if (distance < 0) {
+      continue;
+    }
+    let formatted = formatKeybinding(kb.keys);
```
fcollonval, line 64:
```suggestion
    let formatted = CommandRegistry.formatKeystroke(kb.keys);
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  // Display shortcuts by group
+  const bindingTable: any = [];
+  for (let d = 0; d <= maxDistance; d++) {
+    if (groupedBindings.has(d)) {
+      bindingTable.push(
+        groupedBindings.get(d)!.map(b => (
+          <tr className="jp-ContextualShortcut-TableRow" key={b.command}>
+            <td className="jp-ContextualShortcut-TableItem">
+              {b.keys.map(CommandRegistry.formatKeystroke).join(', ')}
```
fcollonval, line 99:
```suggestion
              {CommandRegistry.formatKeystroke(b.keys)}
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -191,6 +192,24 @@ const browser: JupyterFrontEndPlugin<void> = {
...
     );
 
+    // Show the current file browser shortcut in its title.
+    const updateBrowserTitle = () => {
+      const binding = find(
+        app.commands.keyBindings,
+        b => b.command === CommandIDs.toggleBrowser
+      );
+      if (binding) {
+        const ks = binding.keys.map(CommandRegistry.formatKeystroke).join(', ');
```
fcollonval, line 202:
```suggestion
        const ks = CommandRegistry.formatKeystroke(binding.keys);
```

yarn.lock
```diff
@@ -6006,6 +6006,11 @@ es6-iterator@^2.0.3:
     es5-ext "^0.10.35"
     es6-symbol "^3.1.1"
 
+es6-promise@~4.2.8:
+  version "4.2.8"
+  resolved "https://registry.yarnpkg.com/es6-promise/-/es6-promise-4.2.8.tgz#4eb21594c972bc40553d276e510539143db53e0a"
+  integrity sha512-HJDGx5daxeIvxdBxvG2cb9g4tEvwIk3i8+nhX0yGrYmZUzbkdg8QbDevheDB8gd0//uPj4c1EQua8Q+MViT0/w==
+
```
fcollonval, line 6013:
```suggestion
```

yarn.lock
```diff
@@ -10324,7 +10329,7 @@ react-paginate@^6.3.2:
   dependencies:
     prop-types "^15.6.1"
 
-react-toastify@^9.0.8:
+react-toastify@~9.1.1:
```
fcollonval, line 10332:
```suggestion
react-toastify@^9.0.8:
```

packages/apputils-extension/src/index.ts
```diff
@@ -606,6 +608,14 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
         }
       }
     });
+
+    commands.addCommand(CommandIDs.displayShortcuts, {
```
fcollonval, line 612: Could you add a shortcut for the new command?

You need to create a new file `packages/apputils-extension/schema/utilityCommands.json` that contains:

```json
{
  "title": "Shortcuts Help",
  "description": "Shortcut help settings.",
  "jupyter.lab.shortcuts": [
    {
      "command": "apputils:display-shortcuts",
      "keys": ["Accel Shift H"],
      "selector": "body"
    }
  ],
  "properties": {},
  "type": "object",
  "additionalProperties": false
}
```

fcollonval (changes requested):
Thanks @HaudinFlorence 

Could you add a integration test for this one?

You can add it in `galata/test/documentation/general.test.ts` around Line 490. The steps would be to open a notebook, trigger the display of the help dialog, check the dialog; something like:

```typescript
  test('Command Palette', async ({ page, tmpPath }) => {
    await page.goto(`tree/${tmpPath}`);

    await page.notebook.createNew();

    await page.keyboard.press('Control+Shift+H');

    await expect(page.locator('.jp-Dialog')).toHaveScreenshot('shortcuts_help.png');
  });

```

> This test will compare screenshots. At the end it is better as we will be able to use the screenshots for documentation purpose.

packages/apputils-extension/src/index.ts
HaudinFlorence: This has been added.

HaudinFlorence: @jtpio Thanks for the suggestion. I will try to add the command to the main menu and to the command palette.

jtpio:
It would indeed be nice to have the shortcuts command available as a menu item and via the command palette.

The logic would be similar to the license plugin with `IMainMenu` and `ICommandPalette` defined as `optional` here: 

https://github.com/jupyterlab/jupyterlab/blob/3ecf75ac8eda5f11e55f318658577bba227ecdf6/packages/help-extension/src/index.tsx#L458

Which are then used here:

https://github.com/jupyterlab/jupyterlab/blob/3ecf75ac8eda5f11e55f318658577bba227ecdf6/packages/help-extension/src/index.tsx#L595-L603

It seems that the shortcuts logic in this PR looks for the `document.activeElement` though, not sure how this will work when the command is access via the menu or the palette.

The main advantage of having the shortcuts available from the menu and palette would be that the [downstream plugin in Notebook 7](https://github.com/jupyter/notebook/blob/0ccb87867d9de4af6feaeccfdf1539af541f6268/packages/help-extension/src/index.tsx#L159-L228) could then be removed to use the one from JupyterLab instead.

packages/apputils-extension/src/index.ts
krassowski: Question: did we discuss the shortcut? In classic notebook it was just <kbd>H</kbd> in command mode.

For context, this is where it is in classic notebook:

Screenshot from 2023-02-23 20-20-01

HaudinFlorence: @krassowski You suggest to use capital H too for this shortcut in Jupyter Lab ? I don't have an opinion on the choice. @fcollonval What do you think about it ?

HaudinFlorence:
> It would indeed be nice to have the shortcuts command available as a menu item and via the command palette.
> 
> The logic would be similar to the license plugin with `IMainMenu` and `ICommandPalette` defined as `optional` here:
> 
> https://github.com/jupyterlab/jupyterlab/blob/3ecf75ac8eda5f11e55f318658577bba227ecdf6/packages/help-extension/src/index.tsx#L458
> 
> Which are then used here:
> 
> https://github.com/jupyterlab/jupyterlab/blob/3ecf75ac8eda5f11e55f318658577bba227ecdf6/packages/help-extension/src/index.tsx#L595-L603
> 
> It seems that the shortcuts logic in this PR looks for the `document.activeElement` though, not sure how this will work when the command is executed via the menu or the palette.
> 
> The main advantage of having the shortcuts available from the menu and palette would be that the [downstream plugin in Notebook 7](https://github.com/jupyter/notebook/blob/0ccb87867d9de4af6feaeccfdf1539af541f6268/packages/help-extension/src/index.tsx#L159-L228) could then be removed to use the one from JupyterLab instead.

@jtpio Thanks for your help. From what I see at first and what I was suspecting, the contextual help seems to always be the same (the one of the notebook it seems) when opened from the help menu or from the command palette.

HaudinFlorence:
bot please update snapshots

packages/apputils-extension/src/index.ts
```diff
@@ -564,8 +568,13 @@ const sessionDialogs: JupyterFrontEndPlugin<ISessionContextDialogs> = {
 const utilityCommands: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlab/apputils-extension:utilityCommands',
   requires: [ITranslator],
+  optional: [IMainMenu, ICommandPalette],
```
fcollonval, line 571: You don't use the menu so `IMainMenu` is unneeded - this is what creates the error as the `activate` function gets as arguments `[app, ...requires, ...optional]`. So in the current code, the variable `palette` received `IMainMenu` and not `ICommandPalette`

```suggestion
  optional: [ICommandPalette],
```

packages/apputils-extension/src/index.ts
```diff
@@ -564,8 +568,13 @@ const sessionDialogs: JupyterFrontEndPlugin<ISessionContextDialogs> = {
...
   autoStart: true,
-  activate: (app: JupyterFrontEnd, translator: ITranslator) => {
+  activate: (
+    app: JupyterFrontEnd,
+    translator: ITranslator,
+    palette: ICommandPalette
```
fcollonval, line 576: As it is optional, it may be `null`.

```suggestion
    palette: ICommandPalette | null
```

packages/apputils-extension/src/index.ts
fcollonval: @krassowski I fear using `H` is to simple to be robust in JupyterLab. And as the shortcuts help can be displayed on any document, it will for example be impossible to use it with the file editor. So we would end up having multiple shortcuts depending on the active document.

But at least the infrastructure does not prevent us to use `H` in Notebook v7.

packages/apputils-extension/src/index.ts
```diff
@@ -606,6 +615,17 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
...
+    commands.addCommand(CommandIDs.displayShortcuts, {
+      label: trans.__('Display Keyboard Shortcuts'),
+      caption: trans.__(
+        'Display relevant keyboard shortcuts for the current active element'
+      ),
+      execute: args => displayShortcuts(app, trans)
+    });
+
+    const category: string = trans.__('Help');
+    palette.addItem({ command: CommandIDs.displayShortcuts, category });
```
fcollonval, line 628: As `palette` is optional:

```suggestion
    if (palette) {
      const category: string = trans.__('Help');
      palette.addItem({ command: CommandIDs.displayShortcuts, category });
    }
```

packages/apputils-extension/style/base.css
```diff
@@ -6,3 +6,33 @@
 
 @import './notification.css';
 @import './splash.css';
+
+.jp-ContextualShortcut-TableItem {
```
fcollonval, line 10: Could you please move the style in a specific file?

yarn.lock
```diff
@@ -10325,7 +10325,7 @@ react-paginate@^6.3.2:
     prop-types "^15.6.1"
 
 react-toastify@^9.0.8:
-  version "9.1.1"
+  version "9.0.8"
```
fcollonval, line 10328: This should be kept as is - the current file is inconsistent.

```suggestion
  version "9.1.1"
```

fcollonval (changes requested):
@HaudinFlorence here are some suggestions that should fix the CI.

packages/apputils-extension/src/index.ts
```diff
@@ -42,6 +42,8 @@
 import { themesPaletteMenuPlugin, themesPlugin } from './themesplugins';
 import { toolbarRegistry } from './toolbarregistryplugin';
 import { workspacesPlugin } from './workspacesplugin';
+import { displayShortcuts } from './shortcuts';
+import { IMainMenu } from '@jupyterlab/mainmenu';
```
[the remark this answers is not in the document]
HaudinFlorence: This is fixed.

jtpio:
> From what I see at first sight (not checked in detail) and as I was suspecting, the contextual help seems to always be the same (the one of the notebook it seems) when opened from the help menu or from the command palette.

Ah ok thanks @HaudinFlorence for looking into this. Then we could just skip adding the command to the main menu and command palette for now. I might check it out later separately to not block this PR.

packages/apputils-extension/style/base.css
HaudinFlorence: Ok. This is done a specific file `packages/apputils-extension/style/contextualhelp.css`has been added and the style details moved to it.

fcollonval:
Kicking the CI

fcollonval:

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { JupyterFrontEnd } from '@jupyterlab/application';
+import { CommandRegistry } from '@lumino/commands';
+import { Selector } from '@lumino/domutils';
+import * as React from 'react';
+import { Dialog, showDialog } from '@jupyterlab/apputils';
+import { TranslationBundle } from '@jupyterlab/translation';
+
+export function displayShortcuts(
+  app: JupyterFrontEnd,
+  trans: TranslationBundle
```
fcollonval, line 15:
```suggestion
  trans: TranslationBundle,
  activeElement?: Element
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+) {
+  const { commands } = app;
+  const elt = document.activeElement;
```
fcollonval, line 18:
```suggestion
  const elt = activeElement ?? document.activeElement;
```

galata/test/documentation/general.test.ts
```diff
@@ -487,6 +487,18 @@ test.describe('General', () => {
...
   });
 
+  test('Keyboard Shortcuts Help', async ({ page, tmpPath }) => {
+    await page.goto(`tree/${tmpPath}`);
+
+    await page.notebook.createNew();
+
+    await page.keyboard.press('Control+Shift+H');
+
+    await expect(page.locator('.jp-Dialog')).toHaveScreenshot(
```
fcollonval, line 497: This is a bit to large and will bring trouble, let's reduce the captured area:

```suggestion
    expect(await page.locator('.jp-Notebook').screenshot()).toMatchSnapshot(
```

fcollonval (changes requested):
Thanks @HaudinFlorence here are some suggestions to improve the case of triggering the command not from the keyboard shortcut.

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+  background-color: white;
+  vertical-align: top;
+  padding-left: 2px;
+  padding-right: 2px;
+  padding-top: 1px;
+  padding-bottom: 1px;
+  color: #888;
+  border-style: solid;
+  border-width: 1px;
+  font-size: 11.7px;
```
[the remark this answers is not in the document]
andrii-i: `--jp-ui-font-size0` and `--jp-ui-font-size1` seem to be good candidates here

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+  font-family: monospace;
+}
+
+.jp-ContextualShortcut-TableRow {
+  vertical-align: top;
+  font-size: 13px;
```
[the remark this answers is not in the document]
andrii-i: `--jp-ui-font-size0` and `--jp-ui-font-size1` seem to be good candidates here

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+  font-family: Arial, Helvetica, sans-serif;
```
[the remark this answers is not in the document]
andrii-i: `  --jp-ui-font-family` seems to be a good candidate here

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+  vertical-align: top;
+  padding-left: 2px;
+  padding-right: 2px;
+  padding-top: 1px;
+  padding-bottom: 1px;
+  color: #888;
+  border-style: solid;
+  border-width: 1px;
+  font-size: 11.7px;
+  font-family: monospace;
```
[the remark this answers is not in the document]
andrii-i: `  --jp-ui-font-family` seems to be a good candidate here

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+.jp-ContextualShortcut-TableItem {
+  margin-left: auto;
+  margin-right: auto;
+  background-color: white;
+  vertical-align: top;
+  padding-left: 2px;
+  padding-right: 2px;
+  padding-top: 1px;
+  padding-bottom: 1px;
+  color: #888;
```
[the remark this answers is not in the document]
andrii-i: Please see `--jp-layout-colorN` family of variables

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+.jp-ContextualShortcut-TableItem {
+  margin-left: auto;
+  margin-right: auto;
+  background-color: white;
```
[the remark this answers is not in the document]
andrii-i: Please see `--jp-layout-colorN` family of variables

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,34 @@
...
+  vertical-align: top;
+  padding-left: 2px;
+  padding-right: 2px;
+  padding-top: 1px;
+  padding-bottom: 1px;
+  color: #888;
+  border-style: solid;
+  border-width: 1px;
```
andrii-i, line 17: Should `--jp-border-width` css variable be used here?

fcollonval (changes requested):
@HaudinFlorence I added some pointers to improve PR.

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,122 @@
...
+          <tr className="jp-ContextualShortcut-TableRow" key={b.command}>
+            <td>{commands.label(b.command, b.args)}</td>
+            <td className="jp-ContextualShortcut-TableItem">
+              {CommandRegistry.formatKeystroke(b.keys)}
+            </td>
+          </tr>
+        ))
+      );
+      bindingTable.push(
+        <tr className="jp-ContextualShortcut-TableLastRow"></tr>
```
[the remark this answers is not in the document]
HaudinFlorence: @andrii-i Thanks for the suggestion. I don't know where exactly I should declare these caps global const (in my case, the css classes I defined were in a .tsx file).  And in the example I saw, let's say we have a `const NAME_CLASS = 'jp-Name'`, then in the constructor of the related class, a `this.addClass(NAME_CLASS). Sorry to ask for this, I don't have experience with that.

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
HaudinFlorence: @andrii-i Thanks for the comment. This has been taken into account.

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
HaudinFlorence: This has been modified.

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
HaudinFlorence: This has been changed.

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
HaudinFlorence: This has been modified.

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,122 @@
...
+  // Display shortcuts by group
+  const bindingTable: any = [];
+  for (let d = 0; d <= maxDistance; d++) {
+    if (groupedBindings.has(d)) {
+      bindingTable.push(
+        groupedBindings.get(d)!.map(b => (
+          <tr className="jp-ContextualShortcut-TableRow" key={b.command}>
+            <td>{commands.label(b.command, b.args)}</td>
+            <td className="jp-ContextualShortcut-TableItem">
+              {CommandRegistry.formatKeystroke(b.keys)}
```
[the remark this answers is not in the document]
HaudinFlorence: thanks for this proposition @fcollonval. The integration, adaptation and use of the function ` formatKeys' is not yet functionnal.

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
andrii-i: Thanks @HaudinFlorence

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
andrii-i: Thanks @HaudinFlorence

packages/apputils-extension/style/contextualhelp.css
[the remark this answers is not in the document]
andrii-i: Thanks @HaudinFlorence

packages/apputils-extension/src/shortcuts.tsx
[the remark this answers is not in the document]
andrii-i: @HaudinFlorence No worries, sorry that I was not clear. By "[all caps](https://en.wikipedia.org/wiki/All_caps)" I meant exactly what you understood, that all letters of the constant name are capital letters. The idea is that instead of using arbitrary strings in different places, you declare all arbitrary constants (css class names, resolutions, endpoints IPs and so on) in one place (below import statements, outside of any function), add descriptive comments to them, and then use these constants in your code. This makes code easier to understand, makes it possible to reuse constants if needed).

You can follow example I linked above and declare it in in the same file you use it in, outside of scope of any function, right below imports. For example you can do it at line 12 of packages/apputils-extension/src/shortcuts.tsx. It's also a good idea to add a comment describing the purpose of this class Something like:
```ts
/**
 * The class name added to last row of each ContextShortcutTable to ***insert purpose here***.
 */
const CONTEXT_SHORTCUT_TABLE_LAST_ROW_CLASS = 'jp-ContextualShortcut-TableLastRow'
```

Then you just use this constant wherever you would use the string before. Line 102 would then be:
```ts
        <tr className="jp-ContextualShortcut-TableLastRow"></tr>
```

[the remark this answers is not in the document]
HaudinFlorence: @andrii-i Thanks a lot for the precisions. Changes have been included to follow your comments.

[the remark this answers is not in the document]
andrii-i: @HaudinFlorence Thank you for taking this and other suggestions into consideration

andrii-i:
> @fcollonval @andrii-i @jtpio @krassowski @jasongrout Here is a update of the screenshot of the shortcuts display. Feel free to share your suggestions 

I think font family and font size here look good. 

Labels and Shortcuts don't align vertically, there is also a large horizontal space between Label and Shortcut which makes it hard to understand which Label belongs to which Shortcut. I think initial version presented in the PR description or table from classic works really well in terms of vertical alignment between label and shortcut, in terms of not having much horizontal space between them. 

I would recommend keeping using css variables and/or adaptive units but making table look closer to initial version or to classic for vertical alignment between label and shortcut, for horizontal space between them. 

Vertical misalignment in current version:
223461134-c7db9d67-5d7c-4cf9-a8b8-2d2fb47c97f4

Initial version:
Screenshot 2023-03-07 at 3 58 00 PM

Classic from #10284:
image

jasongrout:
> @fcollonval @andrii-i @jtpio @krassowski @jasongrout Here is a update of the screenshot of the shortcuts display. Feel free to share your suggestions

Thanks for all your work on this! A suggestion from looking at the screenshot: in the menu and command palette, we show keyboard modifier keys without the `+` and with the appropriate symbols on various platforms. For example, here's a screenshot of the edit menu on a mac:

Screen Shot 2023-03-08 at 08 53 52

In my original PR, I had a function [`formatKeyBinding`](https://github.com/jupyterlab/jupyterlab/pull/11960/files#diff-bff7008237b575f8298c99c905c175a510873cb9a948ebc9c6226a7cfb252bd1R48-R50) that formatted shortcuts consistent with how they were formatted elsewhere in the application. Can we use something like that function to format these shortcuts as well for consistency? That way we have consistency across the application, conforming to norms on a particular platform.

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,155 @@
...
+   * Based on Lumino private function commands.Private.targetDistance
+   * This traverses the DOM path from `elt` to the root
+   * computes the distance from `elt` to the first node which matches
+   * the CSS selector. If no match is found, `-1` is returned.
+   *
+   * It also stops traversal if the `data-lm-suppress-shortcuts` or
+   * `data-p-suppress-shortcuts` attributes are found.
+   */
+
+  function formatKeys(keys: string[]): JSX.Element {
```
jasongrout, line 49: Small note: it looks like this function is now separating the comment above from the function it applies to (the matchDistance function below). Can we move the comment down to be with its corresponding function again?

krassowski:
As for the categories I would suggest staring with re-using the categories as in Keyboard Shortcuts in settings:

Screenshot from 2023-03-08 17-44-02

HaudinFlorence:
@krassowski To echo your comments on the categories in the settings editor (I haven't checked how they are defined), an alternative display can be the following one (for a command that  is `'notebook:copy-cell`, category will what is at the left of the `:`)
Screenshot from 2023-03-09 22-21-56

HaudinFlorence:
> @HaudinFlorence let's try to focus on a first version that works with a scrollable list. If you can figure out a easy way to get the category, we can add it. But otherwise let's split that as follow-up action (same for multi columns and other style enhancements). The primary goal is to get a first version for 4.0.0-beta and iterate on it for the RC.

@fcollonval Ok for the priority definition. I proposed just above a tentative way to give the category of the shortcut ( not yet pushed in this PR).

packages/apputils-extension/src/shortcuts.tsx
HaudinFlorence: @jasongrout Thanks for pointing this issue. This will be corrected in the next commit.

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,53 @@
...
+
+.jp-ContextualShortcut-TableHeader {
+  text-align: left;
+}
+
+.jp-ContextualShortcut-TableLastRow {
+  height: '2em';
+}
+
+.jp-ContextualShortcut-TableItem .jp-ContextualShortcut-Key {
```
krassowski, line 42: Do we need to combine the selector? If not let's drop the `TableItem` part

```suggestion
.jp-ContextualShortcut-Key {
```

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,53 @@
...
+ */
+:root {
+  --jp-private-shortcuts-table-padding-vertical: 9px;
+  --jp-private-shortcuts-table-padding-right: 14px;
+  --jp-private-shortcuts-key-padding-horizontal: 5px;
+  --jp-private-shortcuts-key-padding-vertical: 3px;
+}
+
+tr,
+td {
```
krassowski, line 13: Please use selectors specific to the shortcuts table.

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+:root {
```
krassowski, line 5: Since these are private we can move them to `.jp-ContextualShortcut-Table` (proposed new class) and `jp-ContextualShortcut-Key` respectively.

Please see https://github.com/jupyterlab/jupyterlab/issues/14167 for context.

krassowski (changes requested):
CSS review

packages/apputils-extension/style/contextualhelp.css
HaudinFlorence: @krassowski Thanks for the comments. My knowledge of css is not very high. I tried to remove this but the reason I used directly tr and td is that it was the only way I found to manage to apply the padding that were suggested in the issue from a design point of view.

packages/apputils-extension/style/contextualhelp.css
HaudinFlorence: @krassowski Thanks for the comments. I have changed `--jp-private-shortcuts-table-padding-****`to `--jp-private-shortcuts-row-padding-****` and finally did not need to introduce the new class `.jp-ContextualShortcut-Table`but used already existing one `.jp-ContextualShortcut-TableRow`. I moved the classes out of root section.

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -2,22 +2,15 @@
...
-tr,
-td {
-  padding-bottom: var(--jp-private-shortcuts-table-padding-vertical);
-  padding-top: var(--jp-private-shortcuts-table-padding-vertical);
-  padding-right: var(--jp-private-shortcuts-table-padding-right);
-}
 
 .jp-ContextualShortcut-TableRow {
+  --jp-private-shortcuts-row-padding-vertical: 9px;
+  --jp-private-shortcuts-row-padding-right: 14px;
```
andrii-i, line 8: Should [relative length units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths) be used here so that for example when font size is changed the padding would also change?

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -39,7 +31,12 @@ td {
   height: '2em';
 }
 
-.jp-ContextualShortcut-TableItem .jp-ContextualShortcut-Key {
+.jp-ContextualShortcut-Key {
+  font-family: var(--jp-code-font-family);
+
+  --jp-private-shortcuts-key-padding-horizontal: 5px;
+  --jp-private-shortcuts-key-padding-vertical: 3px;
```
andrii-i, line 38: Should [relative length units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths) be used here so that for example when font size is changed the padding would also change?

andrii-i (commented):
When using Dark theme (Settings > Theme > JupyterLab Dark), background color behind shortcuts doesn't match the background color of a popup window.

Thank you for taking so many comments into account @HaudinFlorence. Overall, I think this looks really clean now. If you need any support in terms of code / css, please let me know, I would be happy to contribute. 

Screenshot 2023-03-13 at 10 04 57 AM
Screenshot 2023-03-13 at 10 12 03 AM

packages/apputils-extension/style/contextualhelp.css
```diff
@@ -0,0 +1,50 @@
...
+  font-size: var(--jp-ui-font-size1);
+  padding-right: var(--jp-private-shortcuts-row-padding-right);
+  font-family: var(--jp-ui-font-family);
+}
+
+.jp-ContextualShortcut-TableItem {
+  margin-left: auto;
+  margin-right: auto;
+  vertical-align: center;
+  background-color: var(--jp-layout-color0);
```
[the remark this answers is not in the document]
HaudinFlorence: @andrii-i Thanks for the comment. Your suggestion has been taken into account.

HaudinFlorence:
Please update galata snapshots

andrii-i:
Kicking CI

packages/apputils-extension/src/index.ts
```diff
@@ -612,6 +620,31 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
...
+          document.activeElement
+        );
+
+        if (!included) {
+          const currentNode =
+            (app.shell.currentWidget as MainAreaWidget)?.content.node ??
+            app.shell.currentWidget?.node;
+          currentNode?.focus();
+        }
+        displayShortcuts(app, trans);
```
krassowski, line 640: Lint is failing here:

```
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator
```

you could either convert this function to `async` and add an `await` or simply return the promise like:

```suggestion
        return displayShortcuts(app, trans);
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * The class name for each button-like symbol representing a key used in a shortcut in the ContextShortcutTable
+ */
+const SHORTCUT_TABLE_HEADER = 'jp-ContextualShortcut-TableHeader';
+
+export function displayShortcuts(
+  app: JupyterFrontEnd,
+  trans: TranslationBundle,
+  activeElement?: Element
+) {
```
krassowski, line 38: Two thoughts:
- narrowing public API: does `displayShortcuts` need an entire `app`? Can it just get `commands` instead?
- making API easily extensible: is it worth wrapping the arguments in an `IOptions` interface so that we can add arguments in 4.1 (if needed) without waiting for 5.0?

krassowski (commented):
Two comments (one highlighting a failing lint test and one requesting a rework to public API).

packages/apputils-extension/src/shortcuts.tsx
HaudinFlorence: Thanks @krassowski. This has been taken into account.

fcollonval (commented):
@HaudinFlorence the code is looking good.

I triggered the bot to update the snapshots. And I have a small suggestion to add code documentation.

andrii-i:
Kicking CI

andrii-i (commented):
I would recommend adding a bit of padding between hotkey name and key combination as currently longest hotkey name is very close to key combination. 

Screenshot 2023-03-28 at 3 39 01 PM


Screenshot 2023-03-28 at 3 58 16 PM

andrii-i (commented):
Thank you for the updates @HaudinFlorence. Updated Keyboard Shortcuts table and help menu item look good to me and work well. Good job! 

Note https://github.com/jupyterlab/jupyterlab/issues/14240#issuecomment-1487782944 interaction for when some extension registers a keyboard shortcut with a long description but this can be dealt with as a part of V2. 

Please also note that currently there are 2 failing tests:
- [Visual Regression](https://github.com/jupyterlab/jupyterlab/actions/runs/4575686146/jobs/8078827909?pr=14053). There is screenshot comparasion fail for newly-added Shortcuts tests: `test/documentation/general.test.ts:582:7 › General › Keyboard Shortcuts Help `. [This](https://github.com/jupyterlab/jupyterlab/pull/13921#issuecomment-1432758906) comment by @fcollonval and one under it describe how to run and debug visual regression tests locally.
- [Linux (usage, 3.11](https://github.com/jupyterlab/jupyterlab/actions/runs/4575686141/jobs/8078829927?pr=14053). There is something going on with duplicate package and then a failure on trying to clone cookiecutter package. Not exactly sure if this is connected to this PR, maybe other reviewers who understand CI better could advise, CC @krassowski @fcollonval  

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/work/jupyterlab/jupyterlab/jupyterlab/upgrade_extension.py", line 201, in <module>
    update_extension(args.path, args.branch, args.no_input is False)
  File "/home/runner/work/jupyterlab/jupyterlab/jupyterlab/upgrade_extension.py", line 80, in update_extension
    cookiecutter(
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/cookiecutter/main.py", line 69, in cookiecutter
    repo_dir, cleanup = determine_repo_dir(
                        ^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/cookiecutter/repository.py", line 106, in determine_repo_dir
    cloned_repo = clone(
                  ^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/cookiecutter/vcs.py", line 95, in clone
    subprocess.check_output(  # nosec
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'clone', 'https://github.com/jupyterlab/extension-cookiecutter-ts']' returned non-zero exit status 128.
```

fcollonval (approved):
Thanks @HaudinFlorence 

Merging this first version

PR #14063, opened 2023-02-21
Fix non-document wide undo stack

### LINKED ISSUE #10791: Allow to disable the document-wide history tracking

### LINKED ISSUE #11556: `experimentalDisableDocumentWideUndoRedo` breaks undo of completer

fcollonval:
## References

Fixes #10791
Fixes #11556
Fixes https://github.com/jupyterlab/jupyterlab/issues/14113
Closes https://github.com/jupyterlab/jupyterlab/pull/11640

Requires https://github.com/jupyter-server/jupyter_ydoc/pull/148

Question: should we rename the settings and activate it by default so the behavior of undo is restored to the old (<=3.0) behavior?

## Code changes

Fix properly the action origin to be tracked in the history track

## User-facing changes

Non-document wide history matches the old undo behavior

None

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -748,7 +748,7 @@
     },
     "documentWideUndoRedo": {
       "title": "Enable undo/redo actions at the notebook document level.",
-      "description": "Enables the undo/redo actions at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell.",
+      "description": "Enables the undo/redo actions at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell. This is deprecated and will be removed in 5.0.0.",
       "type": "boolean",
       "default": false
     },
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -746,9 +746,9 @@
         "showProgress": true
       }
     },
-    "experimentalEnableDocumentWideUndoRedo": {
-      "title": "Enable the undo/redo on the notebook document level.",
-      "description": "Experimental settings that enables the undo/redo at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell.",
+    "documentWideUndoRedo": {
+      "title": "Enable undo/redo actions at the notebook document level.",
+      "description": "Enables the undo/redo actions at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell.",
       "type": "boolean",
       "default": false
     },
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -746,9 +746,9 @@
         "showProgress": true
       }
     },
-    "experimentalDisableDocumentWideUndoRedo": {
-      "title": "Experimental settings to enable the undo/redo on the notebook document level.",
-      "description": "Disable the undo/redo on the notebook document level, so actions independent cells can have their own history. The undo/redo never applies on the outputs, in other words, outputs don't have history. A moved cell completely looses history capability for now.",
+    "experimentalEnableDocumentWideUndoRedo": {
+      "title": "Enable the undo/redo on the notebook document level.",
+      "description": "Experimental settings that enables the undo/redo at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell.",
       "type": "boolean",
       "default": false
     },
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -56,4 +56,4 @@ dependencies:
   # extra conda stuff
   - jsonschema-with-format-nongpl
   - pip:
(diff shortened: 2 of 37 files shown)


hbcarlos:
> should we rename the settings and activate it by default so the behavior of undo is restored to the old (<=3.0) behavior?

In my opinion, we should remove that setting and keep only the old behavior.

hbcarlos:
> > should we rename the settings and activate it by default so the behavior of undo is restored to the old (<=3.0) behavior?

> In my opinion, we should remove that setting and keep only the old behavior.

 
I'm also fine renaming the settings and activating it by default.

hbcarlos (approved):
Thanks, @fcollonval!

krassowski:
It does not seem to work for me on binder the way I would expect to work. When I press undo, changes from any cell get undone, regardless of the setting.

hbcarlos:
> It does not seem to work for me on binder the way I would expect to work. When I press ctrl + z changes from any cell get undone, regardless of the setting.

Thanks, @krassowski. That's right, it doesn't work on Binder. I believe is related to the `jupyter_collaborative` extension because it works locally for me without that extension.

fcollonval:
Thanks for looking at it @hbcarlos - I'll come back to this one next week after the API freeze.

hbcarlos:
Hi @krassowski. It is fixed now. The problem was that jupyter_collaboration did not use the setting and was always using a global undo manager.

packages/notebook-extension/schema/tracker.json
```diff
@@ -746,9 +746,9 @@
         "showProgress": true
       }
     },
-    "experimentalDisableDocumentWideUndoRedo": {
-      "title": "Experimental settings to enable the undo/redo on the notebook document level.",
-      "description": "Disable the undo/redo on the notebook document level, so actions independent cells can have their own history. The undo/redo never applies on the outputs, in other words, outputs don't have history. A moved cell completely looses history capability for now.",
+    "experimentalEnableDocumentWideUndoRedo": {
+      "title": "Enable the undo/redo on the notebook document level.",
+      "description": "Experimental settings that enables the undo/redo at the notebook document level; aka undoing within a cell may undo the latest notebook change that happen in another cell.",
```
krassowski, line 751: Should we drop experimental? I guess this could be done in a separate PR.

```suggestion
      "description": "Experimental settings that enables the undo/redo at the notebook document level; aka undoing within a cell may undo the latest notebook change that happened in another cell.",
```

fcollonval: The undo/redo at document level does not feel user friendly. So I kept the _experimental_. But I guess we could remove it and iterate when we get feedback.

krassowski: If we are considering dropping it we could add "deprecated, will be removed in 5.0" in the description and see if any one complains. If no one does we can drop it.

fcollonval:

Issue #14023, opened 2023-02-16
More granular `notifyCommandChanged` in the debugger extension


jtpio:
# Description

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/13915

This logic seems to be triggering commands regardless whether the plugin is available or activated on the page.

Spotted downstream in Notebook 7, which does not include the command to inspect variables (`@jupyterlab/debugger-extension:variables`):

![image](https://user-images.githubusercontent.com/591645/219375471-e127e33d-544f-46c1-9556-ee11db35d712.png)

# Expected Behavior

Each plugin should probably be responsible for notifying its own set of commands, since some plugins can be deactivated or simply not available in other applications (for example Notebook 7).

# Additional Context
_Originally posted by @jtpio in https://github.com/jupyterlab/jupyterlab/pull/13915#discussion_r1108463072_

bollwyvl:
It would also be _frigging rad_ if it the debugger didn't take the default "hard reload" keyboard shortcut until it was activated.

PR #14066, opened 2023-02-22
Debugger: check `hasCommand` before calling `notifyCommandChanged`

jtpio:
## References

This should fix https://github.com/jupyterlab/jupyterlab/issues/14023

## Code changes

Make sure a command is registered before calling `notifyCommandChanged`.

## User-facing changes

None

None

### DIFF
--- a/packages/debugger-extension/src/index.ts
+++ b/packages/debugger-extension/src/index.ts
@@ -50,7 +50,9 @@ import { ITranslator } from '@jupyterlab/translation';
 
 function notifyCommands(app: JupyterFrontEnd): void {
   Object.values(Debugger.CommandIDs).forEach(command => {
-    app.commands.notifyCommandChanged(command);
+    if (app.commands.hasCommand(command)) {
+      app.commands.notifyCommandChanged(command);
+    }
   });
 }
 


fcollonval (commented):
Thanks @jtpio 

Could we cache the list of commands to notify or make `notifyCommandChanged` call forgiven?

jtpio:
Should we expect the `hasCommand` lookup to be expensive? An alternative could be to surround `notifyCommandChanged` with a try catch.

fcollonval:
> Should we expect the `hasCommand` lookup to be expensive? An alternative could be to surround `notifyCommandChanged` with a try catch.

I don't know - but as they are so many commands in Lab, I guess it is not that cheap. But I'm fine moving forward as is.

krassowski:
> Should we expect the hasCommand lookup to be expensive? An alternative could be to surround notifyCommandChanged with a try catch.

I would say no. Expectations aside, `hasCommand` is currently [implemented as a single `Map.has()` call](https://github.com/jupyterlab/lumino/blob/7438017b3a5a05353ae378f2c0d79799c3ad1533/packages/commands/src/index.ts#L92-L94) which is at least sublinear according to ECMA specification and in practice often implemented as O(1) so we should not worry here.

krassowski (approved):
Thanks!

jtpio:
Thanks both for the extra details and feedback.



> Expectations aside, `hasCommand` is currently [implemented as a single `Map.has()` call](https://github.com/jupyterlab/lumino/blob/7438017b3a5a05353ae378f2c0d79799c3ad1533/packages/commands/src/index.ts#L92-L94) which is at least sublinear according to ECMA specification and in practice often implemented as O(1) so we should not worry here.

Right that was also my understanding, thanks for providing this extra context :+1: Also `hasCommand` is used in another places in the JupyterLab code base.

Issue #9005, opened 2020-09-14
Option to disable the autolink feature


fcollonval:
### Problem

The autolinker can be annoying when the goal is to copy plain text.

### Proposed Solution

It would be interesting to be able to deactivate the autolink feature.

jasongrout:
Would you propose this be a JupyterLab-wide setting?

fcollonval:
I think the easiest would be a trigger option at the global level. Remembering it per notebook or per cell does not sound necessarily if the option can be switched quickly.

jasongrout:
+1 on a global jlab-level setting.

The autolinking logic is called here: https://github.com/jupyterlab/jupyterlab/blob/1b63011f95220b5c1e045109aea34308569d63d5/packages/rendermime/src/renderers.ts#L541

So I think what I would do if I were implementing this is:

- create an option in the renderText.IRenderOptions just below that line for an option controlling whether autolinking is called
- Create a similar option for the render text widget: https://github.com/jupyterlab/jupyterlab/blob/1b63011f95220b5c1e045109aea34308569d63d5/packages/rendermime/src/widgets.ts#L366
- Create a *new* factory at https://github.com/jupyterlab/jupyterlab/blob/d69f588d2d589dd50b18e65d0a1856af31574b7d/packages/rendermime/src/factories.ts#L62 that copies the current factory, but passes in the autolinking false option.
- Use the setting option to decide which text factory to initialize the rendermime registry with at https://github.com/jupyterlab/jupyterlab/blob/d69f588d2d589dd50b18e65d0a1856af31574b7d/packages/rendermime-extension/src/index.ts#L80

This would make the setting take effect on refresh. You could also listen for changes in the setting and replace the text rendering in the global rendermime registry so that setting changes took effect on new documents immediately.

fcollonval:
Thanks a lot @jasongrout for the pointer on how to achieve that feature.

PR #14068, opened 2023-02-22
Option to disable the autolink feature

cccs-nik:
## References

Implements #9005: Option to disable the autolink feature. I was going to create my own issue but found this existing one from 2020. I would like the same feature implemented. I manage a JupyterHub deployment for our organization where our users deal with malicious URLs that we don't want to be clickable when outputted.

## Code changes

An additional setting for the rendermime extension to skip URL replacement in the autolink function used by the text renderer.

## User-facing changes

A new setting for the rendermime extension:
image

The new setting would allow skipping URL replacement like so:
image

I think none.


fcollonval (commented):
Thanks for the contribution @cccs-nik 

As the sanitizer is already part of available information (and this is somehow related), I was wondering if we could not extend `ISanitizer` to provide a new flag `autolink: boolean`. It will be connected to the settings in that plugin:

https://github.com/jupyterlab/jupyterlab/blob/3ecf75ac8eda5f11e55f318658577bba227ecdf6/packages/apputils-extension/src/index.ts#L615

This will make the code change much smaller with no need for signals.

cccs-nik:
I've modified it based on your suggestion. The code changes are indeed smaller. Let me know if this is what you had in mind and if you have any other suggestions.

Merci!

packages/rendermime-interfaces/src/index.ts
```diff
@@ -378,6 +378,11 @@ export namespace IRenderMime {
    * An object that handles html sanitization.
    */
   export interface ISanitizer {
+    /**
+     * @returns Whether to replace URLs.
```
fcollonval, line 382:
```suggestion
     * @returns Whether to replace URLs by HTML anchors.
```

packages/apputils/src/tokens.ts
```diff
@@ -196,6 +196,11 @@ export const ISanitizer = new Token<ISanitizer>(
 );
 
 export interface ISanitizer {
+  /**
+   * @returns Whether to replace URLs.
```
fcollonval, line 200:
```suggestion
   * @returns Whether to replace URLs  by HTML anchors.
```

packages/apputils/src/sanitizer.ts
```diff
@@ -447,6 +447,13 @@ export class Sanitizer implements ISanitizer {
     return sanitize(dirty, { ...this._options, ...(options || {}) });
   }
 
+  /**
+   * @returns Whether to replace URLs.
```
fcollonval, line 451:
```suggestion
   * @returns Whether to replace URLs by HTML anchors.
```

packages/completer/src/widget.ts
```diff
@@ -1199,7 +1199,10 @@ export namespace Completer {
     ): HTMLElement {
       const host = document.createElement('div');
       host.classList.add('jp-RenderedText');
-      const sanitizer = { sanitize: this.sanitizer.sanitize };
+      const sanitizer = {
+        getAutolink: this.sanitizer.getAutolink,
+        sanitize: this.sanitizer.sanitize
+      };
```
fcollonval, line 1205: Let's simplify this:

```suggestion
      const sanitizer = this.sanitizer;
```

packages/rendermime/src/renderers.ts
```diff
@@ -496,9 +496,14 @@ export namespace renderSVG {
...
  * @param content - The text content of a node.
  *
+ * @param replace - If URLs should be replaced.
+ *
  * @returns A list of text nodes and anchor elements.
  */
-function autolink(content: string): Array<HTMLAnchorElement | Text> {
+function autolink(
+  content: string,
+  replace: boolean
```
fcollonval, line 505: Let's shunt this method to avoid unnecessary regex lookup.

```suggestion
  content: string
```

packages/rendermime/src/renderers.ts
```diff
@@ -515,7 +520,7 @@ function autolink(content: string): Array<HTMLAnchorElement | Text> {
   let lastIndex = 0;
 
   let match: RegExpExecArray | null;
-  while (null != (match = webLinkRegex.exec(content))) {
+  while (null != (match = webLinkRegex.exec(content)) && replace) {
```
fcollonval, line 523:
```suggestion
  while (null != (match = webLinkRegex.exec(content))) {
```

fcollonval (changes requested):
Thanks @cccs-nik 

Tiny changes and a request: could you add unit tests there (you can use that text as example):

https://github.com/jupyterlab/jupyterlab/blob/a920f332d0b12d084588f672af3926fc7d31163d/packages/rendermime/test/factories.spec.ts#L121

cccs-nik:
Thanks for the quick reviews @fcollonval. I fixed the above, just need to add unit tests but I'll have to push that to later next week. I'll send a re-request review your way once that's done.

fcollonval:
Thanks @cccs-nik 

I realize that there is an inconsistency in interface inheritance for `ISanitizer` - it is defined in two places when it should not. Would you mind if I push a commit to remove that inconsistency? I would also make the new `getAutolink` optional to avoid api-changes in the mimetype renderer.

cccs-nik:
Yup feel free to push a commit for both of those changes! @fcollonval

fcollonval (approved):
Thanks for the patience @cccs-nik

PR #14074, opened 2023-02-24
Styling fixes for rjsm Settings Editor: remove accordions, rework "Restore to defaults" button, add placeholder

### LINKED ISSUE #14007: Font size of accordion titles in Settings Editor is much larger than font size of other components
## Description

Font size of accordion titles in Settings Editor is much larger than font size of other components

## Reproduce

1. Open Settings Editor
2. Note font size of accordion titles

image

## Expected behavior

Font size is consistent with other jupyterlab components

## Context

Operating System and version: MacOS 12.6.3
Browser and version: Chrome version 109.0.5414.87 (Official Build) (arm64)
JupyterLab version: head-of-master (WIP 4.0.0)

andrii-i:
## References

- Fixes #14007 

## Code changes

Remove accordions from Settings Editor, rework "Restore to defaults" button to be shown as a part of "button bar" area, add placeholder for when no plugin is chosen from pluginlist to rjsm, reuse it for JSON editor as well.

On search, stack of plugins with matching fields would be shown (in line with for example VSCode).

new_settings_editor

## User-facing changes

- No accordions
- When no plugin is chosen from pluginlist, placholder screen with instructions is shown. When plugin is chosen, react json schema for that plugin is shown. When search is used, stack of all plugins with matching fields is shown. 
- Restore to defaults button is now of "confirm" color (blue) and is positioned lower

Current head-of-master:
Screenshot 2023-02-23 at 5 35 54 PM

With this change:
Screenshot 2023-03-30 at 3 17 45 PM
image

None


andrii-i:
Preview with changes to font size but without removing header descriptions:
Screenshot 2023-02-23 at 4 55 54 PM

andrii-i:

andrii-i:
Alternative version in #14073 with descriptions present:
image

andrii-i:
bot please update snapshots

andrii-i:
Kicking CI

andrii-i:
@marthacryan, requesting your review as from author of existing Settings Editor accordion header styling.

@fcollonval, @krassowski, @isabela-pf requesting your review as I thought that you might have thoughts / opinions on this proposed change.

Would also mention this PR during JupyterLab call on Wednesday.

jtpio:

jtpio:
Triggering a new snapshot update since this one does not seem related to the change:

image

krassowski:
> fcollonval, @krassowski, isabela-pf requesting your review as I thought that you might have thoughts / opinions on this proposed change.

I do have an opinion, but not necessarily one that everyone agrees with: In my opinion we should removal all the accordions as these serve no useful purpose when each section takes up full screen. Instead we should only ever show the active section (the same way as JSON Settings Editor does works - one section ever visible at a time) UNLESS in search mode (where we do not show titles. As a user I never found myself in a need of having two sections of settings open simultaneously and was always annoyed by the accordion because it detracts attention. If that was to be the solution we decide to pursue removing description may not be necessary.

I think I brought up this before late in the process of merging the original UI in so I am not sure if this is something that folks do not agree with or just did not have a chance to think about.

andrii-i:
I agree with @krassowski that removing Accordions altogether might be a good idea. The only additional functionality that accordions provide vs settings menu to the left is quick access to "Restore to Defaults" button (if it exists, majority of accordions don't have it). Accordions might play some role that I'm missing. I looked at PR #11079 that adds accordion headers and issue #8148 that PR references as a design spec but they don't mention accordions.

[Binder link](https://mybinder.org/v2/gh/jupyterlab/jupyterlab-demo/75ab1750f37b27db8e135c2c4f2139da6b641609?urlpath=lab/tree/demo) from jupyterlab/jupyterlab master leads to 3.2.9 version that has extension manager with no accordions. Here's how it looks:
Screenshot 2023-02-28 at 10 39 33 AM

andrii-i:

andrii-i:
Fixed css vars for font size of accordion headers (now `--jp-ui-font-size`, was `--jp-content-font-size3`), triggering snapshot update

bollwyvl:
Ah, i misunderstood: one settings form at a time is _great_, provided they: 
- get thoroughly garbage collected when switching
- can be deeply linked
- are still searchable in a meaningful way

marthacryan:
> Accordions might play some role that I'm missing. I looked at PR #11079 that adds accordion headers and issue #8148 that PR references as a design spec but they don't mention accordions.

I have no problem with removing the accordion! Here's my background knowledge about the accordion decision:
* The design doc you referenced #8148 had all of the settings on one page, organized into categories (but not collapsible) 
* When I tried to implement that design, there were very long rendering times because it was too long of a page
* I think we talked through it in the JupyterLab weekly dev call and thought the accordion style would be a good solution to keep the look of having all settings in one page without having to render all of them at once. Some kind of lazy loading would be good too if we wanted to keep that look

krassowski:
> Accordions might play some role that I'm missing.

The reasoning was:
- "we want to search in settings, let's display all and filter them down"
- "oh no, we cannot render so many forms because performance suffers"
- "let's add accordions to only show part of the settings"
- "actually accordions are not great UX"

We still need to show headers (or another indicator) when user searches for a setting using the search box as there may be multiple identically named settings in different plugins.

fcollonval:
I don't have strong opinion on the look and feel here (except maybe the viewport when changing section in the accordion panel does not always jump at the expected position).

> The only additional functionality that accordions provide vs settings menu to the left is quick access to "Restore to Defaults" button (if it exists, majority of accordions don't have it).

I just want to ensure there is no misunderstanding here. The _Restore to Defaults_ button exist on all accordion headers. It is only displayed if the settings for that section is not matching its default values.
If we were to remove the accordion, I guess a good solution would be to _move_ that button in the panel toolbar (similar to the button to open the JSON editor) and use a enable/disable state rather than a visibility state.

fcollonval:
FYI on firefox the header text is very thin:

image

andrii-i:
Currently search in new extension manager / head-of-master is broken, created issue #14292 to track it. 

settings_editor_search

andrii-i:
Bot please update snapshots

andrii-i:
Made some updates and changes, please review if possible. 

- Removed accordions
- When no plugin is chosen from pluginlist, placholder screen with instructions is shown. When plugin is chosen, react json schema for that plugin is shown. When search is used, stack of all plugins with matching fields is shown.
- Restore to defaults button is now of "confirm" color (blue) and is positioned lower

andrii-i:
Kicking CI

andrii-i:
Bot please update Galata snapshots

andrii-i:
Hi @GabrielaVives! Somehow UI doesn't allow me to request a review from UI through reviewers tab but if you could give this PR a look from UIUX standpoint, I would be very grateful. Thank you!

GabrielaVives:
HI @andrii-i , I agree that the accordion here is not useful. There is just one design issue that I see during the neutral state, when nothing is selected, the `Cell Toolbar` item on the left menu is highlighted: 

image

I think nothing should be selected here.

krassowski:
> I think nothing should be selected here.

Or, would it be better if there was no blank state indicator and we always had a plugin selected? That would possibly simplify code also save the user one click?

fcollonval (commented):
@andrii-i do you have time to take into account the latest comments from Mike and Gabriela?

andrii-i:
@fcollonval yes, I will account for the requested changes today

andrii-i:
Hey @fcollonval @krassowski @GabrielaVives I updated based on the suggestion of displaying the selected plugin, preview below. Note difference in user experience vs json settings editor where nothing is selected by default. I think they should be consistent. And I would rather change json settings editor as a part of larger refactor that would eliminate duplication of widgets and state between json settings editor and rjsm settings editor, would eliminate mapping and filtering through array of all plugins on every rerender. What do you think?

Also I don't get this locally but note how on initial render plugin selected in pluginlist on the left and plugin rendered in settingspanel on the right are different. I think it's due to a race condition between pluginlist and settingspanel. 

Kapture 2023-04-09 at 17 48 13

andrii-i:
So here's a version that follows [initial suggestion](https://github.com/jupyterlab/jupyterlab/pull/14074#issuecomment-1494391941)  by @GabrielaVives and has nothing selected by default in rjsm settings editor. This way it's consistent with react json editor and there are no race conditions to deal with. I think this might be a good option to go with for 4.0.0. 

Settings Editor could use both UIUX and software design review so we can come up with a list of improvements for Settings Editor V2 and I would be happy to work on it but it's not realistic to do it for 4.0.0.

rework_settings_editor

andrii-i:
bot please update snapshots

galata/test/jupyterlab/settings.test.ts-snapshots/settings-panel-jupyterlab-linux.png
krassowski, line 1: I feel like we could preserve the old font and add the description below the heading, but this is not a blocker.

krassowski:
I think that either way is fine, the current experience with placeholder looks good to me and if you say this makes the codebase simpler lets keep it that way.

On stylistic note I find the blue "Reset to defaults" button too strong. Semantically, since this is a destructive action (user may have been put a lot of time and effort to customise their settings) I would expect it to be red, not blue. I would also say it should be slightly smaller (not much) and maybe moved to the same line as the title?

None of the stylistic suggestions are blocking - if you don't agree or feel this PR is getting long, we can open a new issue to track this in a follow-up PR.

galata/test/jupyterlab/settings.test.ts-snapshots/settings-panel-jupyterlab-linux.png
andrii-i: I'm very open to design tweaking but heading has the same font and font color, [the only change](https://github.com/jupyterlab/jupyterlab/pull/14074/files/1827b6c11df0ca6a9f88b26ed0229c5cb516a241#diff-d5b826bc4ab275967dfb9d3c7bd01f723d7d9b99142ca1186c4a3816ec65abc7R195-R199) is that now `font-weight: var(--jp-content-heading-font-weight);` is used instead of `font-weight: 300;`. I think use of css variables is preferable as it makes different widgets and UI elements look more similar and streamlined, makes design adaptable to theme change.

On descriptions I'm also very open to showing them if they bring any additional value to people seeing them. As I see it, [currently](https://github.com/jupyterlab/jupyterlab/pull/14074#issuecomment-1444468943) all descriptions all just duplicating the title and are not giving any additional information, they just clutter the UI. What do you think?

andrii-i:
@krassowski Well "Reset to defaults" currently uses `var(--jp-accept-color-normal)` so if you don't think it should convey acceptance, it's a problem 😄. Below are preview of button using grey (`--jp-reject-color-normal`) and red (`--jp-warn-color-normal`). Semantically, I could make an argument for any of the options. I think your argument that this is a destructive action makes sense but red button also seems so I left the button red on head-of-the-branch (is it too bright/too much? Not sure as well).

I also fixed the placeholder functionality so now this PR is ready for review. Let's get an alignment on the "Reset to default" button color and defer all other discussions and improvements to the following PRs (alternatively feel free to propose exact size adjustments for the button if you have them on your mind, I would be happy to accept a suggestion).

Screenshot 2023-04-10 at 5 19 38 PM
Screenshot 2023-04-10 at 5 18 45 PM

andrii-i:
To account for latest changes: bot please update snapshots

andrii-i:
To account for latest changes: bot please update snapshots

andrii-i:
After more consideration and discussion with teammates, I'm sure that red button is a good choice as restoring to defaults is a destructive action that user should be warned about. @fcollonval @krassowski I think this is ready, let's tackle additional comments and design considerations in separate PRs.

Screenshot 2023-04-11 at 2 27 27 PM

andrii-i:
After rebase on master, visual regression test is failing. Bot please update documentation snapshots

andrii-i:
Bot please update snapshots

andrii-i:
Kicking CI

krassowski:
I re-triggered the Visual Regression Tests as some are flaky but this failure might be relevant:

```
Error:   1) [jupyterlab] › test/jupyterlab/settings.test.ts:7:5 › Open the settings editor with a specific search query 
    Error: expect(received).toHaveCount(expected) // deep equality

    Expected: 1
    Received: 0
    Call log:
      - expect.toHaveCount with timeout 5000ms
      - waiting for locator('.jp-SettingsForm')
      -   locator resolved to 0 elements
      -   unexpected value "0"
      - waiting for locator('.jp-SettingsForm')
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"


      18 |   ).toEqual('Command Palette');
      19 |
    > 20 |   await expect(page.locator('.jp-SettingsForm')).toHaveCount(1);
```

fcollonval:
@andrii-i the following integration test failure seems legit:

```
[jupyterlab] › test/jupyterlab/settings.test.ts:7:5 › Open the settings editor with a specific search query: galata/test/jupyterlab/settings.test.ts#L20 
```

andrii-i:
@krassowski @fcollonval It's all green now 🟢

andrii-i:
@fcollonval @krassowski @marthacryan, for future reference, when scenario tested in `[jupyterlab] › test/jupyterlab/settings.test.ts:7:5 › Open the settings editor with a specific search query` occurs? When settings editor is  "opened with a specific search query"?

I see that `settingseditor.tsx` (parent for both `pluginlist.tsx` and `settingspanel.tsx`) passes `initialFilter` prop to `settingspanel.tsx` based on the state of `_filter`property of `pluginlist.tsx`. But at the same time `pluginlist.tsx` sends a signal to `settingspanel.tsx` every time `_filter` is changed. This causes `settingspanel.tsx` to render twice every time filter is updated. I'm either missing something or it's a state duplication and either sending signals or passing `initialFilter` prop to `settingspanel.tsx` can be avoided.

packages/settingeditor/style/base.css
```diff
@@ -320,18 +277,34 @@
   height: 100%;
 }
 
+.jp-SettingsForm .jp-Buttonbar {
```
krassowski, line 280: Let's discuss whether we want to have button bar in https://github.com/jupyterlab/jupyterlab/issues/14403. on CSS-level, I would suggest to replace `jp-Buttonbar` with a more specific `jp-SettingsForm-buttonbar` to avoid introducing a new unscoped class which is only used in a single place (and for performance reasons) - but this can be done in a separate PR

packages/settingeditor/style/base.css
```diff
@@ -320,18 +277,34 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  margin: 8px;
+  gap: 0.5em;
+  max-width: 800px;
```
krassowski, line 286: 800px is a very arbitrary threshold, putting the button in a bit weird place on larger screens:

Screenshot from 2023-04-18 20-30-16

Why not right-align it?

krassowski (approved):
I left suggestions for improvements in #14403; I think that if we want this in for 4.0 we should merge as is and iterate on style details.

andrii-i:
Thank you @krassowski.  I copied your latest per-line suggestions into #14403 so we could continue discussing them.

andrii-i:
`UI Tests / Visual Regression Documentation (3.10.6) (pull_request)` is failing after merging master into this branch, requesting documentation snapshots update: bot please update documentation snapshots

andrii-i:
Kicking CI

andrii-i:
After updating documentation snapshots, Documentation visual regression test doesn't fail anymore but `UI Tests / Visual Regression Tests (pull_request)` fails now, updating galata snapshots: bot please update galata snapshots

andrii-i:
Kicking CI

andrii-i:
Failing `Linux Tests / check_links (pull_request)` does not seem to be related to this PR

fcollonval (approved):
Thanks @andrii-i

andrii-i:
Thank you @fcollonval

Issue #8686, opened 2020-07-13
Display the versions in the extension manager


jtpio:
### Problem

It would be nice if there was a way to know the version of an installed extension via the extension manager (and maybe also the latest for those in "Discover")

![image](https://user-images.githubusercontent.com/591645/87302129-196e5b00-c511-11ea-89be-65775718d0c7.png)

For now this can be achieved by opening a terminal and executing `jupyter labextension list`:

![image](https://user-images.githubusercontent.com/591645/87302474-c3e67e00-c511-11ea-91f9-4dc070f46920.png)

### Proposed Solution

Maybe the version could be displayed next to the name of the extension:

![image](https://user-images.githubusercontent.com/591645/87302593-f3958600-c511-11ea-8d79-944e07d37e08.png)

echarles:
... and `Update` label could be e.g. `Update to v1.2.4`

jtpio:
The version information should be available as part of the response:

![image](https://user-images.githubusercontent.com/591645/221201483-1a0a8814-7199-468e-a665-6b69c8b36a83.png)

Issue #13402, opened 2022-11-09
Show memory usage and load time in extension manager


krassowski:
### Problem

With many extensions installed the performance may slow down. Users have no way to know how much of a performance penalty each extension introduces.

### Proposed Solution

- Measure time for plugin loading and delta in memory usage using `window.performance` functions.
- Discretely show the result in the extension manager (possibly only for plugins exceeding certain load time/size)
- Offer more detail on hover. An additional endpoint could be number of CSS rules contributed.

### Additional context

VSCode shows the activation time:

<details>

![Screenshot from 2022-11-09 17-17-37](https://user-images.githubusercontent.com/5832902/200907690-0a7f1a1e-3f14-4e19-b986-485217f9b9d2.png)

</details>

rutviklhase:
Hey @jweill-aws ! Is anyone working on this? Would love to take this one up!

jweill-aws:
@rutviklhase Nobody's picked this one up yet. Please feel free to submit a pull request if you'd like to implement this!

rutviklhase:
@jweill-aws Awesome! I'll start working on it

krassowski:
> Not necessarily in this PR, but maybe it would also be useful to add these values to the Advanced Plugin Manager, for example in a new column?
> 
> So plugins could be sorted by the time they take to activate, which would help identify potential issues with some third-party extensions.
> 
> <img width="3600" height="2090" alt="image" src="https://github.com/user-attachments/assets/ded01910-7b52-4ac9-9b83-b8903a5c813c" />
> 
> _Originally posted by @jtpio in https://github.com/jupyterlab/jupyterlab/issues/18181#issuecomment-3654103549_

Indeed, we could show it in both the Plugin Manager and the Extension Manager, with the former giving a more granular view and providing frontend-only metrics; the time in Extension Manager could also show server-side loading cost.

PR #14078, opened 2023-02-24
Display the version in the extension manager

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/8686

This help give users better information about their extensions so they don't have to open a new terminal to know the versions.

## Code changes

Update the extension manager widget to display the `installed_version` and `latest_version`.

## User-facing changes

**Before**

image

**After**

image

None


krassowski:
The alternative place to put this would be an "about" button which would allow us to show more version information later. In particular there are often frontend and backend versions (and in case of extensions like lsp multiple frontend versions). Also, the right top-corner could be used for transient timing info as in https://github.com/jupyterlab/jupyterlab/issues/13402#issue-1442541890.

jtpio:
For reference VS Code opens a popup on hover which contains the version and other details:

image

jtpio:

fcollonval (approved):
Thanks @jtpio

jtpio:
Thanks @krassowski and @fcollonval for the reviews.

@krassowski I'll open a follow-up issue to discuss creating an "About" button. It would be fine moving the version to that about dialog later if needed.

jtpio:
> I'll open a follow-up issue to discuss creating an "About" button.

Opened https://github.com/jupyterlab/jupyterlab/issues/14098

PR #14079, opened 2023-02-24
Fix extension manager button background color

jtpio:
## References

The "No, disable" button in the extension manager was not very readable when using the dark theme.

## Code changes

- [x] Add missing CSS class to the button
- [x] Update the background color

## User-facing changes

**Before**

image

**After**

image


jtpio:
bot please update snapshots

andrii-i (approved):
Tested locally, PR fixes the problem. 

On a related note, If we ship dark theme, should we also add a playwright test for dark theme (this can be a separate PR)?

jtpio:
> On a related note, If we ship dark theme by default, should we also add a snapshot verification test for "No, disable" button for dark theme (this can be a separate PR)?

The dark theme is already tested in a few places like this one:

https://github.com/jupyterlab/jupyterlab/blob/27f917cb3b678b4bf3ecf004ba59c16ec6c46aa8/galata/test/jupyterlab/general.test.ts#L27-L31

Technically we could indeed also take reference snapshots with the dark theme. Although this could increase the time for the CI check to run, which is already quite long (~40 minutes).

krassowski:
I tend to agree that it is not necessarily worth running everything in both light and dark mode  because of the time it takes on CI (and since it duplicates number of snapshots we need to review which tends to increase chances of regression slipping through).

I much prefer to add more test cases for complex user interactions rather than themes (which are mostly changing colours and any contrast issues should be caught by the accessibility CI pipeline rather than the main integration tests anyways).

krassowski (approved):
Thanks @jtpio!

PR #13810, opened 2023-01-19
New multi line search

c00kie123:
## References
#13801 

## Code changes

At `packages/documentsearch/src/searchmodel.ts`, I added the `m` flag to  `const flag = caseSensitive ? 'g' : 'gi';`
forming: `const flag = caseSensitive ? 'gm' : 'gim';`. This will enable multi-line search, allowing the regular expression to match across multiple lines

At `packages/fileeditor/src/searchprovider.ts` and `packages/notebook/src/searchprovider.ts`, I removed the line ` return selection?.search(/\r?\n|\r/g) === -1 ? selection : ;''` By removing this line, the function will always return the selection regardless of whether it is a single-line or multi-line selection

At `packages/documentsearch/src/searchview.tsx`, I replaced `<input `with `<textarea`

PR #14085, opened 2023-02-25
Search boxes: switch to `defaultValue`, add a test for typing

### LINKED ISSUE #13881: Search box cursor jumps to end of input after typing a character
## Description

Search box cursor jumps to end of input after typing a character giving bad UX.

## Reproduce

1. Enter `12` in search box
2. Move cursor left so that it is between 1 and 2
3. Press 3
4. See that cursor moved to end
5. Press 4, see that we get 1324 instead of expected 1342

[Screencast from 2023-01-28 22-16-59.webm](https://user-images.githubusercontent.com/5832902/215293651-7e026f68-b2d7-49f2-940d-63e4bd332d2c.webm)

It seems to be a spurious re-render in React.

## Expected behavior

User can modify search text easily

## Context

- JupyterLab version: 4.0.0a32

krassowski:
## References

Closes #13881

While the issue was already fixed by #13810, the solution included in there was generating a warning in development mode of React highlighting that textarea in React should not have children elements and value needs to be implemented using `value` or `defaultValue`.

## Code changes

- switch to `defaultValue` in search input (note that `value` makes React fully manage the input causing #13881 bug)
- added a minimal integration test case (without a screenshot)

## User-facing changes

Typing works well.

None

### DIFF
--- a/galata/test/jupyterlab/notebook-search.test.ts
+++ b/galata/test/jupyterlab/notebook-search.test.ts
@@ -35,6 +35,18 @@ test.describe('Notebook Search', () => {
     expect(await nbPanel.screenshot()).toMatchSnapshot('search.png');
   });
 
+  test('Typing in search box', async ({ page }) => {
+    // Check against React being too eager with controling state of input box
+    await page.keyboard.press('Control+f');
+
+    await page.fill('[placeholder="Find"]', '14');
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @krassowski

PR #14072, opened 2023-02-24
Add output prompt overlay for toggling scroll mode

krassowski:
## References

Closes #5897 by implementing the highly upvoted proposal from https://github.com/jupyterlab/jupyterlab/issues/5897#issuecomment-591884430 (thanks @Linusnie).

## Code changes

- adds optional `promptOverlay` option to `OutputArea.IOptions` (off by default as output area does not manage whether the output is scrollable)
- adds `toggleScrolling` signal to `OutputArea` to connect it in code cells.

## User-facing changes

output-scrolling-overlay

## Backwards-incompatible changes

None, this could possibly be backported.

PR #14087, opened 2023-02-26
Improve position of output scroll toggle overlay

krassowski:
## References

Follow-up to #14072

## Code changes

Add `left` offset when scrolling is enabled to account for `margin-left` from:

https://github.com/jupyterlab/jupyterlab/blob/830931dead61c78dbc5ab12cafd6a8d2e280a75d/packages/cells/style/widget.css#L72-L77

## User-facing changes

Before:

before

After:

after

None

### DIFF
--- a/packages/cells/style/widget.css
+++ b/packages/cells/style/widget.css
@@ -104,6 +104,10 @@
   );
 }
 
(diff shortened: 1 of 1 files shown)


krassowski:

fcollonval (approved):
Thanks @krassowski

Issue #13968, opened 2023-02-12
Match highlight after replace in notebook gets lost


krassowski:
While in #13884 I fix a number of issues in search and add tests, a few non-trivial bugs in the replace logic remain. These are not apparent in the unit tests at the moment because the updates are emitted after a debouncer delay; we should also test with the debouncer disabled so that we can account for the side-effects from delayed updates.

One of the issues is that after replacing the highlight is in a broken state:
- codemirror editor correctly shows the highlight
- search overlay shows `-/2` (after briefly correctly showing `1/2`)

Starting from:

![Screenshot from 2023-02-12 15-27-41](https://user-images.githubusercontent.com/5832902/218320281-2817bc7f-4e23-4eae-88ff-f0dba38c5247.png)

Expected (and briefly seen - greyed out because I needed to set a breakpoint in debugger to capture it in this transitive state):

![Screenshot from 2023-02-12 15-27-56](https://user-images.githubusercontent.com/5832902/218320271-9da9dd7b-9aa3-42fd-bf0d-9752b6176dfb.png)

Current:

![Screenshot from 2023-02-12 15-27-59](https://user-images.githubusercontent.com/5832902/218320258-212882bc-bd39-4e9d-8f28-d596ba18d890.png)

This is because the query gets restarted after replace operation:

![Screenshot from 2023-02-12 15-24-46](https://user-images.githubusercontent.com/5832902/218320058-3acf9f0a-cc84-4a52-b6c4-b070e8d3924d.png)

Notebook listens to changes (`_onSelectionChanged`):

https://github.com/jupyterlab/jupyterlab/blob/edfbea4f9c556875882dfbe71bf77d31e81e5883/packages/notebook/src/searchprovider.ts#L270-L273

which emits `stateChanged` signal (with an incorrect assumption that we should not highlight next item - this is only true if we are not in replace mode!):

https://github.com/jupyterlab/jupyterlab/blob/edfbea4f9c556875882dfbe71bf77d31e81e5883/packages/notebook/src/searchprovider.ts#L579-L592

https://github.com/jupyterlab/jupyterlab/blob/edfbea4f9c556875882dfbe71bf77d31e81e5883/packages/notebook/src/searchprovider.ts#L572-L577

this then leads via debouncer:

https://github.com/jupyterlab/jupyterlab/blob/edfbea4f9c556875882dfbe71bf77d31e81e5883/packages/documentsearch/src/searchmodel.ts#L41

[...]

PR #13861, opened 2023-01-26
Make changes to the tooltips of the toolbar actions.

HaudinFlorence:
## References
Fixes #13503

## Code changes
Add or modify captions of the relevant commands related to the tooltips of the toolbar actions, in `packages/notebook-extension/src/index.ts`.

PR #14090, opened 2023-02-26
Rework search in selection in notebook

### LINKED ISSUE #13756: Ideas on how to improve the find experience

### LINKED ISSUE #14091: Search match and selection of search matches are indistinguishable
## Description

Because current search match highlight is implemented as search selection, it is impossible to understand where the selection is when selecting text with search results. This blocks #14090.

**This also means that current match highlight disappears when cell is not focused**.

## Reproduce

1. In a cell paste:
  ```python
  def solve_lorenz(sigma=10.0, beta=8./3, rho=28.0):
      """Plot a solution to the Lorenz differential equations."""
  ```
2. Search for `lorenz`
3. The first one should be highlighted
4. Now select the text. Both will be highlighted.
5. Can you tell which will be replaced when you use search and replace?

repro-highlight-by-selection-is-wrong

## Expected behavior

Current match highlight is distinct from selection highlight.

## Context

- JupyterLab version: 4.0.0a34 and earlier

### LINKED ISSUE #14105: Cannot create multiple selections/set multiple cursors
## Description

Since migration to CodeMirror6 it is no longer possible. This will also enable us to restore `cursorBlinkRate` as described in #12812:

> cursorBlinkRate: this can be done thanks to another [extension](https://codemirror.net/docs/ref/#view.drawSelection), however this requires some work in the CSS

## Reproduce

1. Open editor
2. Try to ctrl-click to set multiple cursors
3. See still jsut one cursor

## Expected behavior

Multiple cursors are set as in previous versions.

We need to toggle [`allowMultipleSelections`](https://codemirror.net/docs/ref/#state.EditorState^allowMultipleSelections) and enable https://codemirror.net/docs/ref/#view.drawSelection extension; that comes with a performance cost so maybe this could be opt-out.

## Context

- JupyterLab version: 4.0.0a34

krassowski:
## References

- Closes #13756.
- Fixes #14091 (0e0a594ba14065d14688b58e2f1c6454c56e88a0)
- Fixes #14105
- Fixes (partially) #13968

## Code changes

- implements search in editor selection settable with `EditorSearchProvider.setSearchSelection(v)` - only exposed for notebook, but can be easily reused for editor
- reworks existing "search in selected cell(s)" filter with a dynamic "search in selection" filter
- because we need to preserve notebook mode, we cannot focus the editor just to get selection to highlight current match (which also led to other issues as described in [#14091](https://github.com/jupyterlab/jupyterlab/issues/14091).
- because we need the selection to remain unchanged we can no longer use it as implicit state of search engine; for that reason this PR will:
  - always enable CodeMirror6 `drawSelection` addon to show the selection even if the editor is blurred
     - according to CM6 docs this may have a performance impact, but disabling it is a UX regression in several aspects (multiple cursors visibility, selection visibility, selection style) so I propose to keep it always on and only disable if we see it in benchmarks as a real bottleneck
    - this enables us to bring back multiple cursors support
  - do not rely on selection position in `CodeMirrorHighlighter` when editor is not focused and do not set selection until the very moment editor gets focused

#### Search in selected: text when in edit mode, cells when in command mode

demo

selection-in-cells

#### Active filters now have an indicator (an orange dot)

indicator

#### Visibility of search filters is restored when re-opening search box

preserved

N/A


krassowski:

krassowski:

krassowski:

packages/codemirror/src/extension.ts
```diff
@@ -772,6 +772,18 @@ export namespace EditorExtensionRegistry {
           title: trans.__('Tab size')
         }
       }),
+      Object.freeze({
+        name: 'allowMultipleSelections',
+        default: true,
+        factory: () =>
+          createConfigurableExtension((value: boolean) =>
+            EditorState.allowMultipleSelections.of(value)
```
fcollonval, line 780: :heart:

packages/fileeditor/src/searchprovider.ts
```diff
@@ -59,6 +60,15 @@ export class FileEditorSearchProvider
     return this.widget.content.model;
   }
 
+  async startQuery(
+    query: RegExp,
+    filters: IFilters | undefined
+  ): Promise<void> {
+    const promise = await super.startQuery(query, filters);
+    await this.highlightNext(false, true);
+    return promise;
```
fcollonval, line 69: I guess we could simplify this:

```suggestion
    await super.startQuery(query, filters);
    await this.highlightNext(false, true);
```

fcollonval (changes requested):
Thanks @krassowski 

This is a great improvement.

I only have a small suggestions. And I'm surprised by the missing prompt for snapshot: `galata/test/documentation/general.test.ts-snapshots/interface-right-common-documentation-linux.png`

> I did not get the chance to test it on Binder - hard to get it ready.

krassowski:
The missing prompt is also seen in the updated snapshots in https://github.com/jupyterlab/jupyterlab/pull/13861. I am not sure if this is a regression or we need to robustify the test..

fcollonval (approved):
Thanks @krassowski

krassowski:
bot please update snapshots

krassowski:
I will merge this one as soon as green again as merge conflicts accumulate due to extensive snapshots updates. bot please update snapshots

fcollonval:
Thanks for the maintenance to get this one through @krassowski

Alexboiboi:
Thanks @krassowski for the amazing work ;)

Just no sure if this still work in progress or even planed, but when I tested this version it does not behave like many other programs when calling the search with `ctrl+F`. What I mean by that, is, when you have selected a different word in the page than the one currently in the search bar, pressing`ctrl+F` should update the search query.

fcollonval:
Thanks for reaching out @Alexboiboi 

Would you mind looking if there is an issue already opened for the bug you found and if not open a new issue for it?

packages/documentsearch/src/searchview.tsx
```diff
@@ -282,34 +287,32 @@ function SearchIndices(props: ISearchIndexProps) {
 }
 
 interface IFilterToggleProps {
-  enabled: boolean;
-  toggleEnabled: () => void;
+  visible: boolean;
+  toggleVisible: () => void;
+  anyEnabled: boolean;
```
andrii-i, line 292: Function of `anyEnabled` variable was not evident to me from its name. Would using `filterEnabled` or `anyFilterenabled` instead improve clarity?

packages/notebook/src/searchprovider.ts
```diff
@@ -266,43 +346,48 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
...
+      cells.map(async (cell, index) => {
         const cellSearchProvider = createCellSearchProvider(cell);
-        cellSearchProvider.stateChanged.connect(
-          this._onSearchProviderChanged,
-          this
-        );
 
         await cellSearchProvider.setIsActive(
-          !this._filters!.selectedCells ||
+          !this._filters!.selection ||
```
andrii-i, line 363: `_filters` can potentially be `undefined` as [it is not initialized to anything](https://github.com/jupyterlab/jupyterlab/pull/14090/files#diff-40a4a43da20f655025356b02df592dd309f3f67562f6a66df22021d3f2771a7eR792), similar problem to what caused #14419. So using non-null assertion is not safe here. Potential alternatives:
1. Add check instead of non-null assertion at the current line
2. Explicitly initialize `_filters` to `null`, remove union with `undefined`, add check instead of non-null assertion at the current line
3. Initialize `_filters` to something like empty filter and remove union with undefined from its type

packages/notebook/src/searchprovider.ts
```diff
@@ -562,44 +662,142 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
...
+
+    this._filtersChanged.emit();
   }
 
   private _currentProviderIndex: number | null = null;
   private _filters: IFilters | undefined;
-  private _onSelectedCells = false;
+  private _onSelection = false;
+  private _selectedCells: number = 1;
+  private _selectedLines: number = 0;
```
andrii-i, line 795: This is not always correctly updated (please see GIF with testing in the main comment)

andrii-i (commented):
Thank you @krassowski. This fixes #14308 (so there is no random switch from command mode to selection mode when trying to select multiple cells while "Search in N selected cells" is active) and improves performance. 

Note that N in "Search in N selected cells" is not always updated when selecting more than one cell:
N_selected_cell

Please also find a couple of code suggestions.

packages/documentsearch/src/searchview.tsx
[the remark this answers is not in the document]
krassowski: It is not needed because it is in namespace called `IFilterToggleProps` which allows us to use short names as `filters` is implied by context.

packages/documentsearch/src/searchview.tsx
```diff
@@ -282,34 +287,32 @@ function SearchIndices(props: ISearchIndexProps) {
...
           : props.trans.__('Show Search Filters')
       }
     >
-      <ellipsesIcon.react
-        className={className}
-        tag="span"
-        height="20px"
-        width="20px"
-      />
+      <icon.react className={className} tag="span" height="20px" width="20px" />
```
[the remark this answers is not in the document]
krassowski: This should be discussed in an issue with someone who created this bit. I barely switched `ellipsesIcon` to `icon`.

packages/documentsearch/src/searchview.tsx
```diff
@@ -560,8 +558,12 @@ class SearchOverlay extends React.Component<
     const hasFilters = Object.keys(filters).length > 0;
     const filterToggle = hasFilters ? (
```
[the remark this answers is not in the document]
krassowski: Prior to React 18 this was not possible, hence the habit is to return `null` when component is not to be rendered https://github.com/reactwg/react-18/discussions/75

andrii-i:
@krassowski Sorry for the comments, I've put this into my review queue instead of #14387

packages/documentsearch/src/searchview.tsx
```diff
@@ -560,8 +558,12 @@ class SearchOverlay extends React.Component<
...
       <FilterToggle
-        enabled={this.state.filtersOpen}
-        toggleEnabled={() => this._toggleFiltersOpen()}
+        visible={this.props.filtersVisible}
+        anyEnabled={Object.keys(filters).some(name => {
```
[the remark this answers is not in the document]
krassowski: Where `n` = 2 in notebook and 0 in file editor (right now, it should become 1 at some point). Not convinced it will improve anything.

packages/documentsearch/src/searchview.tsx
```diff
@@ -678,11 +680,10 @@ class SearchOverlay extends React.Component<
                 translator={this.translator}
               />
               <div className={SPACER_CLASS}></div>
-              {filterToggle}
             </>
           ) : null}
         </div>
-        {this.state.filtersOpen ? filter : null}
+        {this.props.filtersVisible ? filter : null}
```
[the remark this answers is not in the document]
krassowski: The proposed approach is not more readable. It is implicit. It will also render literal `false` when filters are not visible.

andrii-i:
Sorry about that, not enough rest around 4.0.0 rc0 release

packages/notebook/src/searchprovider.ts
andrii-i: Thank you for answering this and other questions, this helps me understand the codebase better

Issue #13965, opened 2023-02-11
UI shows <code>


domoritz:
## Description

When I install an extension and there is an error, I get this UI

<img width="788" alt="Screenshot 2023-02-10 at 22 37 25" src="https://user-images.githubusercontent.com/589034/218237001-8857c5a7-3060-45aa-9280-b721c84a2101.png">

Notice how the UI shows `<code>` and `</code>`, which it probably shouldn't.

## Context

- Browser and version: Chrome 110
- JupyterLab version: 3.6.1

jasongrout:
Looks like that comes from https://github.com/jupyterlab/jupyterlab/blob/master/packages/extensionmanager/src/dialog.tsx#L23. I think the easiest fix is converting the `<code>` elements to just quotes, so the message is something like `An error occurred installing "mypackage"`

Setting as good first issue. Thanks for the report!

Hamza0601:
Hi, I am new to open source and would like to take this up. Could you tell what steps need to be taken to create a PR fixing this?

jtpio:
Thanks @Hamza0601!

You can follow the instructions in https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html.

Or if you are familiar with the tool use Gitpod to make the contribution in the browser to not have to set up the dev environment locally: https://gitpod.io/#https://github.com/jupyterlab/jupyterlab

And then follow the suggestion above:

> I think the easiest fix is converting the `<code>` elements to just quotes, so the message is something like `An error occurred installing "mypackage"`

When you are ready commit your changes to a new branch and open a new PR on the repo. Thanks for the help!

PR #14095, opened 2023-02-27
 Fix:Issue #13965 resolved. <code> block in the UI must not show now when there is an error installing any extension. 

doesnots:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13965


andrii-i (approved):
This looks good, follows suggestions, and fixes #13965. Thank you, @doesnots.

@doesnots could you please also fill out the template in this PRs header so that people looking at it in the future would be able to quickly see what this PR is about?

andrii-i (changes requested):
There are 2 tests currently failing: 
* [Depreciated har-validator@5.1.5 library warning ](https://github.com/jupyterlab/jupyterlab/actions/runs/4281879501/jobs/7455444442) which does not seem to be related to this change
* [Linter fails](https://github.com/jupyterlab/jupyterlab/actions/runs/4281879501/jobs/7455446292). @doesnots could you please make sure you ran pre-commit hooks on this change? They should run linter and correct all linting problems

packages/extensionmanager/src/dialog.tsx
```diff
@@ -19,9 +19,7 @@ export function reportInstallError(
   translator = translator || nullTranslator;
   const trans = translator.load('jupyterlab');
   const entries = [];
-  entries.push(
-    <p>{trans.__('An error occurred installing <code>%1</code>.', name)}</p>
-  );
+  entries.push(<p>{trans.__(`An error occurred installing "${name}."`)}</p>);
```
jtpio, line 22: Looks like the dot could be moved:

```suggestion
  entries.push(<p>{trans.__(`An error occurred installing "${name}".`)}</p>);
```

jtpio (approved):
Thanks!

Issue #7534, opened 2019-11-18
Switch to lumino


blink1073:
We have continued the Phosphor project code base as a new project, [lumino](https://github.com/jupyterlab/lumino).  We have released all of the `lumino` packages at the same version as they were previously in Phosphor.  The plan is for 2.0 to depend on `lumino`, and resolve any `@phosphor` libraries to `lumino` using `yarn` resolutions.

- [ ] Migrate code to point to `@lumino` packages
- [ ] Add yarn resolution for `@phosphor` -> `@lumino`
- [ ] Update documentation

PR #14109, opened 2023-03-01
Add bottom area to add cells

### LINKED ISSUE #7535: Double-clicking to add cells
In the classic notebook, just below the last cell, there is a div with the class `end_space`. Double-clicking on this div appends a new cell to the notebook. Here is a screenshot of the div (I've added a red border around it to highlight where it is and how big it is:
Screen Shot 2019-11-18 at 9 31 47 AM

Here is the code in the classic notebook that makes it active and append new cells on double-click:

https://github.com/jupyter/notebook/blob/9640e1f94366aa442af279ad77390f6c0a29b8b1/notebook/static/notebook/js/notebook.js#L273-L292

We don't have this in JLab. @bassio just [pointed out on Gitter](https://gitter.im/jupyterlab/jupyterlab?at=5dd2046e50010612b2da18d5) that they use it.

So:

1. Should we add something like this for backwards compatibility and/or as a valuable addition?
2. What should the ux look like? UX? Should it use double-click? I can see something like this being valuable on mobile, for example - double-tapping the extra space adds a new cell. Should the region extend to all of the space below the notebook? Should the region add space below the notebook if none would exist normally?

HaudinFlorence:
Add a footer widget at the end of the notebook that is clickable, in a similar way to what exists in the classic notebook.

add-cells-double-clicking1

## References
Fixes #7535
## Code changes
- a new file `packages/notebooks/src/notebookfooter.ts` introducing the Notebook footer class has been added
- changes have been introduced in `packages/notebooks/src/widget.ts` and in `packages/notebooks/src/windowing.ts`
- css changes have been introduced in `packages/notebooks/style/base.css` to customize the footer widget

### DIFF
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -850,7 +850,7 @@ export namespace NotebookActions {
 
     if (notebook.activeCellIndex === maxCellIndex) {
       const footer = (notebook.layout as NotebookWindowedLayout).footer;
-      footer?.focus();
+      footer?.node.focus();
     }
 
     // Find last non-hidden cell
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -17,6 +17,7 @@ import {
   isRawCellModel,
   MarkdownCell
 } from '@jupyterlab/cells';
+/*import { NotebookWindowedLayout } from './windowing';*/
 import { signalToPromise } from '@jupyterlab/coreutils';
 import * as nbformat from '@jupyterlab/nbformat';
 import { KernelMessage } from '@jupyterlab/services';
@@ -27,6 +28,7 @@ import { JSONExt, JSONObject } from '@lumino/coreutils';
 import { ISignal, Signal } from '@lumino/signaling';
 import * as React from 'react';
 import { Notebook, StaticNotebook } from './widget';
+import { NotebookWindowedLayout } from './windowing';
 
 /**
  * The mimetype used for Jupyter cell data.
@@ -846,15 +848,18 @@ export namespace NotebookActions {
       return;
     }
     let maxCellIndex = notebook.widgets.length - 1;
+
+    if (notebook.activeCellIndex === maxCellIndex) {
+      const footer = (notebook.layout as NotebookWindowedLayout).footer;
+      footer?.focus();
+    }
+
     // Find last non-hidden cell
     while (
       notebook.widgets[maxCellIndex].isHidden ||
       notebook.widgets[maxCellIndex].inputHidden
     ) {
       maxCellIndex -= 1;
-    }
-    if (notebook.activeCellIndex === maxCellIndex) {
-      return;
     }
 
     let possibleNextCellIndex = notebook.activeCellIndex + 1;
@@ -869,7 +874,6 @@ export namespace NotebookActions {
     }
 
     const state = Private.getState(notebook);
-
     notebook.activeCellIndex = possibleNextCellIndex;
     notebook.deselectAll();
     Private.handleState(notebook, state, true);
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -17,7 +17,6 @@ import {
   isRawCellModel,
   MarkdownCell
 } from '@jupyterlab/cells';
-/*import { NotebookWindowedLayout } from './windowing';*/
 import { signalToPromise } from '@jupyterlab/coreutils';
 import * as nbformat from '@jupyterlab/nbformat';
 import { KernelMessage } from '@jupyterlab/services';
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -848,17 +848,17 @@ export namespace NotebookActions {
     }
     let maxCellIndex = notebook.widgets.length - 1;
 
-    if (notebook.activeCellIndex === maxCellIndex) {
-      const footer = (notebook.layout as NotebookWindowedLayout).footer;
-      footer?.node.focus();
-    }
-
     // Find last non-hidden cell
     while (
       notebook.widgets[maxCellIndex].isHidden ||
       notebook.widgets[maxCellIndex].inputHidden
     ) {
       maxCellIndex -= 1;
+    }
+    if (notebook.activeCellIndex === maxCellIndex) {
+      const footer = (notebook.layout as NotebookWindowedLayout).footer;
+      footer?.node.focus();
+      return;
     }
 
     let possibleNextCellIndex = notebook.activeCellIndex + 1;
@@ -873,6 +873,7 @@ export namespace NotebookActions {
     }
 
     const state = Private.getState(notebook);
+
     notebook.activeCellIndex = possibleNextCellIndex;
     notebook.deselectAll();
     Private.handleState(notebook, state, true);
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
 import { Widget } from '@lumino/widgets';
 import { Message } from '@lumino/messaging';
 import { Notebook } from './widget';
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -54,21 +54,8 @@ export class NotebookFooter extends Widget {
    * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
    */
   protected onArrowUp(): void {
-    this.blur();
-    this.notebook.mode = 'command';
-  }
-  /**
-   * Focus the footer and the hidden button
-   */
-  focus(): void {
-    this.node.focus();
-  }
-
-  /**
-   * Blur the footer and the hidden button
-   */
-  blur(): void {
     this.node.blur();
+    this.notebook.mode = 'command';
   }
 
   /*
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -8,6 +8,8 @@ import { Message } from '@lumino/messaging';
 import { Notebook } from './widget';
 import { NotebookActions } from './actions';
 
+const NOTEBOOK_FOOTER_CLASS = 'jp-Notebook-footer';
+
 /**
  * A footer widget added after the last cell of the notebook.
  */
@@ -17,7 +19,11 @@ export class NotebookFooter extends Widget {
    */
   constructor(protected notebook: Notebook) {
     super();
-    this.node.className = 'jp-Notebook-footer';
+    this.addClass(NOTEBOOK_FOOTER_CLASS);
+    this._createHiddenButton();
+    const text = document.createElement('p');
+    text.innerText = 'Click to add a cell.';
+    this.node.appendChild(text);
   }
 
   /**
@@ -26,24 +32,53 @@ export class NotebookFooter extends Widget {
   handleEvent(event: Event): void {
     switch (event.type) {
       case 'click':
-        this.onClick(event);
+        this.onClick(event as MouseEvent);
         break;
+      case 'keydown':
+        if ((event as KeyboardEvent).key === 'ArrowUp') {
+          this.onArrowUp();
+          break;
+        }
     }
   }
 
   /**
-   * On single click, insert a cell below (at the end of the notebook as default behavior).
+   * On single click (mouse event), insert a cell below (at the end of the notebook as default behavior).
    */
   onClick(event: any): void {
     NotebookActions.insertBelow(this.notebook);
   }
 
+  /**
+   * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
+   */
+  onArrowUp(): void {
+    this.blur();
+    this.notebook.mode = 'command';
+  }
+  /**
+   * Focus the footer and the hidden button
+   */
+  focus(): void {
+    this.addClass('focused');
+    this._hiddenButton.focus();
+  }
+
+  /**
+   * Blur the footer and the hidden button
+   */
+  blur(): void {
+    this.removeClass('focused');
+    this._hiddenButton.blur();
+  }
+
   /*
-   * Handle `before-detach` messages for the widget.
+   * Handle `after-detach` messages for the widget.
    */
   protected onAfterAttach(msg: Message): void {
     super.onAfterAttach(msg);
     this.node.addEventListener('click', this);
+    this.node.addEventListener('keydown', this);
   }
 
   /**
@@ -51,6 +86,16 @@ export class NotebookFooter extends Widget {
    */
   protected onBeforeDetach(msg: Message): void {
     this.node.removeEventListener('click', this);
+    this.node.removeEventListener('keydown', this);
     super.onBeforeDetach(msg);
   }
+  /**
+   * Create an hidden button in the notebookfooter widget that enables to add a new cell when it is cliked.
+   */
+  private _createHiddenButton(): void {
+    this._hiddenButton = document.createElement('button');
+    this._hiddenButton.classList.add('hidden');
+    this.node.appendChild(this._hiddenButton);
+  }
+  private _hiddenButton: HTMLButtonElement;
 }
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -1,0 +1,32 @@
+import { Widget } from '@lumino/widgets';
+import { Message } from '@lumino/messaging';
+import { Notebook } from './widget';
+import { NotebookActions } from './actions';
+
+export class NotebookFooter extends Widget {
+  constructor(protected notebook: Notebook) {
+    super();
+    this.node.className = 'jp-Notebook-footer';
+  }
+
+  handleEvent(event: Event): void {
+    switch (event.type) {
+      case 'dblclick':
+        this.onDblClick(event);
+        break;
+    }
+  }
+
+  onDblClick(event: any): void {
+    NotebookActions.insertBelow(this.notebook);
+  }
+
+  protected onAfterAttach(msg: Message): void {
+    super.onAfterAttach(msg);
+    this.node.addEventListener('dblclick', this);
+  }
+  protected onBeforeDetach(msg: Message): void {
+    this.node.removeEventListener('dblclick', this);
+    super.onBeforeDetach(msg);
+  }
+}
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -44,7 +44,6 @@ export class NotebookFooter extends Widget {
    * On single click (mouse event), insert a cell below (at the end of the notebook as default behavior).
    */
   protected onClick(): void {
-    console.log('notebook:', this.notebook);
     if (this.notebook.widgets.length > 0) {
       this.notebook.activeCellIndex = this.notebook.widgets.length - 1;
     }
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -18,12 +18,10 @@ export class NotebookFooter extends Widget {
    * Construct a footer widget.
    */
   constructor(protected notebook: Notebook) {
-    super();
+    super({ node: document.createElement('button') });
+    const trans = notebook.translator.load('jupyterlab');
     this.addClass(NOTEBOOK_FOOTER_CLASS);
-    this._createHiddenButton();
-    const text = document.createElement('p');
-    text.innerText = 'Click to add a cell.';
-    this.node.appendChild(text);
+    this.node.innerText = trans.__('Click to add a cell.');
   }
 
   /**
@@ -32,7 +30,7 @@ export class NotebookFooter extends Widget {
   handleEvent(event: Event): void {
     switch (event.type) {
       case 'click':
-        this.onClick(event as MouseEvent);
+        this.onClick();
         break;
       case 'keydown':
         if ((event as KeyboardEvent).key === 'ArrowUp') {
@@ -45,14 +43,18 @@ export class NotebookFooter extends Widget {
   /**
    * On single click (mouse event), insert a cell below (at the end of the notebook as default behavior).
    */
-  onClick(event: any): void {
+  protected onClick(): void {
+    console.log('notebook:', this.notebook);
+    if (this.notebook.widgets.length > 0) {
+      this.notebook.activeCellIndex = this.notebook.widgets.length - 1;
+    }
     NotebookActions.insertBelow(this.notebook);
   }
 
   /**
    * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
    */
-  onArrowUp(): void {
+  protected onArrowUp(): void {
     this.blur();
     this.notebook.mode = 'command';
   }
@@ -60,16 +62,14 @@ export class NotebookFooter extends Widget {
    * Focus the footer and the hidden button
    */
   focus(): void {
-    this.addClass('focused');
-    this._hiddenButton.focus();
+    this.node.focus();
   }
 
   /**
    * Blur the footer and the hidden button
    */
   blur(): void {
-    this.removeClass('focused');
-    this._hiddenButton.blur();
+    this.node.blur();
   }
 
   /*
@@ -89,13 +89,4 @@ export class NotebookFooter extends Widget {
     this.node.removeEventListener('keydown', this);
     super.onBeforeDetach(msg);
   }
-  /**
-   * Create an hidden button in the notebookfooter widget that enables to add a new cell when it is cliked.
-   */
-  private _createHiddenButton(): void {
-    this._hiddenButton = document.createElement('button');
-    this._hiddenButton.classList.add('hidden');
-    this.node.appendChild(this._hiddenButton);
-  }
-  private _hiddenButton: HTMLButtonElement;
 }
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -11,22 +11,22 @@ export class NotebookFooter extends Widget {
 
   handleEvent(event: Event): void {
     switch (event.type) {
-      case 'dblclick':
-        this.onDblClick(event);
+      case 'click':
+        this.onClick(event);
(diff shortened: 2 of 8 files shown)


GabrielaVives:
Here are my recommendations:

image

These have slightly evolved since the previous prototype but I think it looks neater that way.

andrii-i (changes requested):
Thank you, @HaudinFlorence. This looks and works well.

Please find a couple of suggestions below:
- Lefthand and righthand edges of "Double-click here to add a new cell." area do not horizontally align with respective edges of cell UI elements above it. Should they align? This way UI would looks more streamlined
Screenshot 2023-03-01 at 7 06 02 PM
- "Double-click here to add a new cell." area is only visible on mouse over. While this is much more discoverable than not having any visual indication, this functionality is not evident and needs to be discovered by person using a notebook. This is even more problematic on mobile as interaction area is relatively small and there is no mouse cursor moving randomly around to help in discovery. Should are be visible all the time?
image
- Code comment on argument naming

packages/notebook/src/windowing.ts
```diff
@@ -97,6 +99,30 @@ export class NotebookWindowedLayout extends WindowedLayout {
...
+  get footer(): NotebookFooter | null {
+    return this._footer;
+  }
+
+  /**
+   * Set Notebook's footer
+   *
+   * @param v: Widget | null
+   */
+  set footer(v: NotebookFooter | null) {
```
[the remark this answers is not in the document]
GabrielaVives: Hi @andrii-i , the UI of the rectangle should be aligned with the cells, as I commented above. 

However, I do not think that it should be there permanently. I agree that for a mobile use this is not efficient but for the moment I think we should focus on the web-oriented problem and tackle that later.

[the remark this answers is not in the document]
andrii-i: @GabrielaVives  thank you for the feedback. I'd like to better understand your point of view, could you please elaborate on both points (why UI element should not be visible permanently; why we should focus on web-oriented problems only)?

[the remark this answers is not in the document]
andrii-i: JupyterLab is working well on mobile and are people who use it there on daily basis. Therefore I think we should make sure any design decision / UI element works well on mobile.

andrii-i:
Requesting review from @isabela-pf as from JupyterLab expert on UI/UX and accessibility

andrii-i:
Requesting review from @tonyfast as from jupyterlab mobile user

tonyfast:
i like this feature. it is a good use of the empty space.

when i use it, i want access to it with keyboard navigation. if i navigate passed the last cell, i end up in the button  that lets me add a cell, when i hit <kbd>Enter</kbd> i get a new code cell. currently, we can't add a cell without knowing application keyboard shortcuts or using the mouse. considering this navigation feature could improve the keyboard navigation for assistive devices.

the hit area violates some of expectations of mine. i expect to be able to single click when i see a button looking interface show up, the double click feels forced. in the trello and classic notebook double click examples, the double click happens on a background, not a specific interactive widget. 

my expectations are i can single click on the region to add a code (and maybe there is a button for adding another kind of cell like markdown). after seeing the double click on the background, i'd expect that the empty area below the last cell and new feature to be double clickable.

packages/notebook/src/windowing.ts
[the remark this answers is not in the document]
GabrielaVives: Hi @andrii-i , yes of course! First, let's agree that this feature _could_ be visible permanently if it's encouraged by a specific user need, for the moment with the information I have about the needs from which this solution came from, I advise **not to** for the following reasons:
- in the case of the notebook being small, this button will be visible all the time. This is not a problem for buttons in the toolbar for example as it is specifically meant for this, but the area of the notebook is meant for code, images, graphs, content created by the user. Having a (big) button in this area will create visual noise and bring an unnecessary cognitive load to the user;
- in the case of larger notebooks, this will also happen when the user is at the bottom of it, which probably represents most of the time the user is creating content;
- from benchmarking, we can see that this feature already exists in similar applications, such as Colab or VScode. The button to add a cell appears on hover in both of these. Even though we must question these choices and by no means take decisions based only on a benchmark and "they did it so it must be good", we must take into account that this reflects our user's mental models.

I hope this answered your question, please tell me if you would like to discuss more about this topic!

GabrielaVives:
Thanks a lot @tonyfast for your input, it's really helpful and made me think about a UI that may be more accurate, I'll post it soon!
Regarding the single-click, now that you mention it it seems indeed more natural. I think we should change from double-click to single click if there is a button.
Also, I think there is no need for a sentence prompt (to be tested however), I will sketch some ideas to see how it looks and feels.

HaudinFlorence:
> Here are my recommendations:
> 
> image
> 
> These have slightly evolved since the previous prototype but I think it looks neater that way.

@GabrielaVives Thanks for your comments. They have been taken into account.

GabrielaVives:
@tonyfast Here is the link to a quick prototype illustrating two UI ideas with a single click:
[Prototype A](https://www.figma.com/proto/UKjEjsd5pRi2hMLm8ajqfB/Double-clicking-to-add-cells-%237535?page-id=34%3A6&node-id=34%3A10&viewport=-111%2C349%2C0.39&scaling=min-zoom&starting-point-node-id=34%3A10&show-proto-sidebar=1): whole length button replicating a cell
[Prototype B](https://www.figma.com/proto/UKjEjsd5pRi2hMLm8ajqfB/Double-clicking-to-add-cells-%237535?page-id=34%3A6&node-id=37%3A61&viewport=-111%2C349%2C0.39&scaling=min-zoom&starting-point-node-id=37%3A61&show-proto-sidebar=1): classic button style

HaudinFlorence:
> Thank you, @HaudinFlorence. This looks and works well.
> 
> Please find a couple of suggestions below:
> 
> * Lefthand and righthand edges of "Double-click here to add a new cell." area do not horizontally align with respective edges of cell UI elements above it. Should they align? This way UI would look more streamlined
>   Screenshot 2023-03-01 at 7 06 02 PM
> * "Double-click here to add a new cell." area is only visible on mouse over. While this is much more discoverable than not having any visual indication, this functionality is not evident and needs to be discovered by person using a notebook. This is even more problematic on mobile as interaction area is relatively small and there is no mouse cursor moving randomly around to help in discovery. Should "Double-click here to add a new cell."  area be visible all the time?
>   image
> * Code comment on argument naming

@andrii-i Thanks for your review. Concerning the horizontal alignment, the border of the clickable area are aligned with the one of the code cells (as displayed in the updated presentation gif). For the moment, I haven't changed the visibility of the area (still visible on hover). This point probably needs further discussion.

packages/notebook/src/windowing.ts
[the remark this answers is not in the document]
fcollonval: To support both scenario we could add a conditional style depending if the device support hover or not ([MDN doc](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover)). This is for example what is done for tuning the behavior the file browser check boxes:

https://github.com/jupyterlab/jupyterlab/blob/a51b7d9df2b4e24d39b2333e22f46a96d004bfa6/packages/filebrowser/style/base.css#L243

fcollonval:
> [Prototype A](https://www.figma.com/proto/UKjEjsd5pRi2hMLm8ajqfB/Double-clicking-to-add-cells-%237535?page-id=34%3A6&node-id=34%3A10&viewport=-111%2C349%2C0.39&scaling=min-zoom&starting-point-node-id=34%3A10&show-proto-sidebar=1): whole length button replicating a cell 
> [Prototype B](https://www.figma.com/proto/UKjEjsd5pRi2hMLm8ajqfB/Double-clicking-to-add-cells-%237535?page-id=34%3A6&node-id=37%3A61&viewport=-111%2C349%2C0.39&scaling=min-zoom&starting-point-node-id=37%3A61&show-proto-sidebar=1): classic button style

I like the proposal A better as it highlights the all cell width as interaction area.

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,32 @@
+import { Widget } from '@lumino/widgets';
+import { Message } from '@lumino/messaging';
+import { Notebook } from './widget';
+import { NotebookActions } from './actions';
+
+export class NotebookFooter extends Widget {
```
fcollonval, line 6: Could you add documentation string on this class and its methods.

packages/notebook/src/windowing.ts
```diff
@@ -97,6 +99,30 @@ export class NotebookWindowedLayout extends WindowedLayout {
...
+
+  get footer(): NotebookFooter | null {
+    return this._footer;
+  }
+
+  /**
+   * Set Notebook's footer
+   *
+   * @param footer: Widget | null
+   */
```
fcollonval, line 116:
```suggestion
```

packages/notebook/src/windowing.ts
```diff
@@ -97,6 +99,30 @@ export class NotebookWindowedLayout extends WindowedLayout {
     }
   }
 
+  /**
+   * Returns Notebook's footer
+   *
+   * @return NotebookFooter | null
```
fcollonval, line 105: This is an oversight on my side on the header documentation. But actually getter+setter should only have a single doc string.

```suggestion
   * Notebook widget's footer
```

packages/notebook/src/widget.ts
```diff
@@ -1213,6 +1224,7 @@ export class Notebook extends StaticNotebook {
         this.activeCellIndex = newActiveCellIndex;
       }
     }
+    this.addFooter();
```
fcollonval, line 1227: Could you add in the `constructor` because here it will be called every time a cell is added or removed.

fcollonval (changes requested):
To support the keyboard navigation propose by Tony (focus the placeholder when navigating after the last cell). The action `selectBelow` is the one to modify:

https://github.com/jupyterlab/jupyterlab/blob/a51b7d9df2b4e24d39b2333e22f46a96d004bfa6/packages/notebook/src/actions.tsx#L799

> You will have to modify also `selectAbove` to switch from the footer to the last cell.

I would say the best will be in that case to set the focus on the footer. So you can use CSS style to make it appear. And then you can add an event listener on `keydown` on the footer to trigger adding a cell on <kbd>Enter</kbd>.

packages/notebook/src/widget.ts
HaudinFlorence: This has been changed.

packages/notebook/src/notebookfooter.ts
HaudinFlorence: Sure. Documentation strings have been added.

HaudinFlorence:
> To support the keyboard navigation propose by Tony (focus the placeholder when navigating after the last cell). The action `selectBelow` is the one to modify:
> 
> https://github.com/jupyterlab/jupyterlab/blob/a51b7d9df2b4e24d39b2333e22f46a96d004bfa6/packages/notebook/src/actions.tsx#L799
> 
> > You will have to modify also `selectAbove` to switch from the footer to the last cell.
> 
> I would say the best will be in that case to set the focus on the footer. So you can use CSS style to make it appear. And then you can add an event listener on `keydown` on the footer to trigger adding a cell on Enter.

@fcollonval Thanks for the input. This is not yet implemented. Still need to figure out how to do it.

tonyfast:
@GabrielaVives thanks for making those prototypes so fast. i am in agreement with @fcollonval that prototype A is nearest to my expectation.

>  Also, I think there is no need for a sentence prompt (to be tested however), I will sketch some ideas to see how it looks and feels.

well this area is a basically button so the button `title` can can contain the call to action if its too verbose for the main view.
i think prototype A gives you more flexibility in design. now that i can add cell. i am asking questions can i add a code cell OR a markdown cell. 

if we continue with the keyboard navigation. i feel like i could navigate to the new cell area, then use <kbd>Tab</kbd> to navigate to alternative options, then <kbd>Enter</kbd> would confirm the addition of the cell.

packages/notebook/src/windowing.ts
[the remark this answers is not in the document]
andrii-i: @GabrielaVives Thank you for providing detailed reply. These are really good points, I especially agree that it is important to not have notebook visually cluttered. 

I think @fcollonval's idea might be a good middle ground between absence or visual clutter and usability on mobile. Again, I don't use jupyterlab on mobile much so people who do like @tonyfast could give a better insight if always showing "add cell" button on mobile (vs showing it on touch / mouse over only) would be useful. Just wanted to took both options into consideration.

andrii-i:
Kicking CI

andrii-i:
@HaudinFlorence updated width looks great to me, thank you. 

On visibility, [I think](https://github.com/jupyterlab/jupyterlab/pull/14109#discussion_r1125204404)
> @fcollonval's idea might be a good middle ground between absence or visual clutter and usability on mobile. I don't use jupyterlab on mobile much so people who do like @tonyfast can give a better insight if always showing "add cell" button on mobile (vs showing it on touch / mouse over only) would be useful. Just want to make sure we consider mobile experience.

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A footer widget added after the last cell of the notebook.
+ */
+export class NotebookFooter extends Widget {
+  /**
+   * Construct a footer widget.
+   */
+  constructor(protected notebook: Notebook) {
+    super();
+    this.node.className = 'jp-Notebook-footer';
```
andrii-i, line 20: Should all caps global const instead of inline CSS class string be used here?

Please see `shell.ts` for reference:
https://github.com/jupyterlab/jupyterlab/blob/82e1a74b61a41fbe5d65e8542053b749cdfe6ea0/packages/application/src/shell.ts#L39

packages/notebook/src/windowing.ts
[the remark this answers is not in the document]
tonyfast: on a small mobile screen, the cell toolbar is always persistent so i can always add a cell. i think since there is already interaction for adding a new cell in the toolbar maybe we don't need anything special for very small screens. 

adding more feature to a small screen might be bad, but on a larger tablet this UX might be possible because there is more screen real estate.

andrii-i:
@GabrielaVives thank you for organizing this. Posted some questions about this meeting [in #10284 thread](https://github.com/jupyterlab/jupyterlab/issues/10284#issuecomment-1462361763)

packages/notebook/src/widget.ts
HaudinFlorence: @fcollonval If I am not mistaking it seems that moving `addFooter`to the constructor causes the footer widget to become invisible or just not added after the last cell of the notebook. I don't know why.

packages/notebook/src/widget.ts
```diff
@@ -1194,6 +1195,17 @@ export class Notebook extends StaticNotebook {
...
+    /*this.addFooter();*/
+  }
+
+  /**
+   * Adds a footer to the notebook.
+   */
+  protected addFooter(): void {
+    const trans = this.translator.load('jupyterlab');
+    const info = new NotebookFooter(this);
+    info.node.textContent = trans.__('Click to add a cell.');
```
trungleduc, line 1207: You should set the content of the node inside the class `NotebookFooter`

HaudinFlorence: @trungleduc Thanks for the comment. This has been taken into account.

HaudinFlorence:
The keyboard navigation is now working on this PR.
add-cells-keyboard-navigation2

HaudinFlorence:

andrii-i:
Kicking CI

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+.jp-Notebook-footer {
+  height: 10%;
```
fcollonval, line 7: Could you set a fix height because this make the area height change with the window size?

```suggestion
  height: ...px;
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+  }
+
+  /**
+   * On single click (mouse event), insert a cell below (at the end of the notebook as default behavior).
+   */
+  onClick(event: any): void {
+    NotebookActions.insertBelow(this.notebook);
+  }
```
fcollonval, line 50: Please make this protected and not public.
Don't request the event if you don't need it.
You should set the active index to the latest cell otherwise the cell does not get inserted where it should.

```suggestion
  protected onClick(): void {
    if (this.notebook.cells.length > 0) {
      this.notebook.activeCellIndex = this.notebook.cells.length - 1;
    }
    NotebookActions.insertBelow(this.notebook);
  }
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
+   */
+  onArrowUp(): void {
```
fcollonval, line 55: This one does not need to be public

```suggestion
  protected onArrowUp(): void {
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Construct a footer widget.
+   */
+  constructor(protected notebook: Notebook) {
+    super();
+    this.addClass(NOTEBOOK_FOOTER_CLASS);
+    this._createHiddenButton();
+    const text = document.createElement('p');
+    text.innerText = 'Click to add a cell.';
+    this.node.appendChild(text);
```
fcollonval, line 26: Let's the widget be a button
Let's translate the displayed text

```suggestion
    super({ node: document.createElement('button') });
    const trans = notebook.translator.load('jupyterlab');
    this.addClass(NOTEBOOK_FOOTER_CLASS);
    this.node.innerText = trans.__('Click to add a cell.');
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+  onArrowUp(): void {
+    this.blur();
+    this.notebook.mode = 'command';
+  }
+  /**
+   * Focus the footer and the hidden button
+   */
+  focus(): void {
+    this.addClass('focused');
+    this._hiddenButton.focus();
```
fcollonval, line 64: This can be simplified:

```suggestion
    this.node.focus();
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Blur the footer and the hidden button
+   */
+  blur(): void {
+    this.removeClass('focused');
+    this._hiddenButton.blur();
```
fcollonval, line 72: Mirror focus changes:

```suggestion
    this.node.blur();
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+  /**
+   * Create an hidden button in the notebookfooter widget that enables to add a new cell when it is cliked.
+   */
+  private _createHiddenButton(): void {
+    this._hiddenButton = document.createElement('button');
+    this._hiddenButton.classList.add('hidden');
+    this.node.appendChild(this._hiddenButton);
+  }
+  private _hiddenButton: HTMLButtonElement;
```
fcollonval, line 100: Unneeded

```suggestion
```

packages/notebook/src/widget.ts
```diff
@@ -1196,6 +1197,7 @@ export class Notebook extends StaticNotebook {
     this.node.setAttribute('data-lm-dragscroll', 'true');
     this.activeCellChanged.connect(this._updateSelectedCells, this);
     this.selectionChanged.connect(this._updateSelectedCells, this);
+    this.addFooter();
```
fcollonval, line 1200: Could you add this line after `super(options);`

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,48 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
fcollonval, line 10: This will need some changes as soon as the widget is a button.

```suggestion
  display: flex;
  justify-content: center;
  align-items: center;
```

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,48 @@
...
+  box-shadow: var(--jp-input-box-shadow);
+  background-color: var(--jp-cell-editor-active-background);
+}
+
+.jp-Notebook-footer .hidden {
+  opacity: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+}
```
fcollonval, line 48: Unneeded

```suggestion
```

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,48 @@
...
+  }
+}*/
+
+/* .jp-Notebook-footer:hover {
+  border: var(--jp-border-width) solid var(--jp-cell-editor-border-color);
+  border-radius: 0;
+  color: var(--jp-layout-color3);
+} */
+
+.jp-Notebook-footer.focused {
```
fcollonval, line 37: This should do the trick

```suggestion
.jp-Notebook-footer:focus {
```

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,48 @@
...
+    color: var(--jp-layout-color3);
+  }
+}*/
+
+/* .jp-Notebook-footer:hover {
+  border: var(--jp-border-width) solid var(--jp-cell-editor-border-color);
+  border-radius: 0;
+  color: var(--jp-layout-color3);
+} */
+
```
fcollonval, line 36: Remove as unused.

```suggestion
```

packages/notebook/src/actions.tsx
```diff
@@ -17,6 +17,7 @@ import {
   isRawCellModel,
   MarkdownCell
 } from '@jupyterlab/cells';
+/*import { NotebookWindowedLayout } from './windowing';*/
```
fcollonval, line 20:
```suggestion
```

fcollonval (changes requested):
Thanks @HaudinFlorence 

Here are some suggestions to simplify and improve the code.

There are some glitches on the vertical positioning as seen in the test snapshots. But I saw some also when testing locally. Let's wait for the latest changes and see what is the situation.

fcollonval (changes requested):
Thanks @HaudinFlorence 

Could you please improve the CSS rules of the button to:
- Remove the button background
- Remove the hover effect
- Add `cursor: pointer` CSS rule

HaudinFlorence:
> Thanks @HaudinFlorence
> 
> Could you please improve the CSS rules of the button to:
> 
> * Remove the button background
> * Remove the hover effect
> * Add `cursor: pointer` CSS rule

Thanks for the comments @fcollonval. I am not sure how to remove the hover effect. Other 2 bullet points have been taken into account.

andrii-i:
@HaudinFlorence @fcollonval Testing locally, the only effect happening on button hover is `cursor: pointer`

andrii-i:
Kicking CI

andrii-i (commented):
Note that when you create a new cell with "Click to add a cell." button, focus remains on the "Click to add a cell." and you remain in `COMMAND` mode. So if you press enter, another cell would be created. If you create a new cell by using `Shift + Enter` keyboard shortcut while editing a cell, new cell becomes active and you remain in `EDIT` mode. I think this behavior is consistent as both interactions don't change the current mode. 

Tagging @jasongrout (author of #7535) and @bassio (author of request based on which  #7534 was created) to request their opinion.

fcollonval:
bot please update snapshots

fcollonval (approved):
Thanks @HaudinFlorence 

Merging this first version

I'll create a follow-up issue about the scrolling in the notebook that is not user friendly when adding cells.

fcollonval:
> To maintainer, I approve it but we should still check all tests are passing.

fcollonval:
bot please update snapshots

andrii-i (changes requested):
After testing this with a couple of people including teammates, I get consistent feedback that this change significantly changes the user experience and adds visual clutter. Some people are ok with that but other people strongly don't like it. 

At the same time, option where button is only shown on hover works well for everyone and doesn't clutter the user experience.

andrii-i:
The number of actions to add the new cell from keyboard should be the same with button visible in hover only. 

One way to do this using media query and css would be to use `:hover` and `:focus-visible` to set `opacity: 1` (while otherwise making it 0).

fcollonval:
@andrii-i [837bc8b](https://github.com/jupyterlab/jupyterlab/pull/14109/commits/837bc8b6d81e3f1b7be5a3cd99476a420f4ea59f) adds the opacity change on hover. What is your team thinking of that version.

andrii-i:
@fcollonval this looks great, doesn't make UI more cluttered, and doesn't increase the number of actions required to add a new cell from keyboard
add_cell_on_hover

andrii-i (approved):

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,40 @@
...
+  margin-top: 6px;
+  background: none;
+  cursor: pointer;
+}
+
+.jp-Notebook-footer:focus {
+  border-color: var(--jp-cell-editor-active-border-color);
+}
+
+/* For devices that support hovering, footer until hover */
```
krassowski, line 30:
```suggestion
/* For devices that support hovering, hide footer until hover */
```

krassowski (commented):
I see this jumping when adding a new cell, this is presumably unrelated to this PR but just to windowed notebooks? @fcollonval what do you think?

jumps

packages/notebook/style/notebookfooter.css
```diff
@@ -0,0 +1,40 @@
...
+  );
+  width: calc(
+    100% -
+      (
+        var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) +
+          var(--jp-cell-padding) + var(--jp-cell-padding)
+      )
+  );
+  border: var(--jp-border-width) solid var(--jp-cell-editor-border-color);
+  color: var(--jp-layout-color3);
```
krassowski, line 20: Should we use font variables for font colour?

```suggestion
  color: var(--jp-ui-font-color3);
```

fcollonval:
... I don't get the glitch in the integration test. When doing it manually I'm unable to see it and I'm not able to reproduce it when running the test locally. Let's see what happen on the second attempt.

fcollonval:
Ok that was flaky - merging it now.

PR #14115, opened 2023-03-02
Fix tab trap notebook cells

gabalafou:
## Short summary

This PR is an attempt to solve an issue that makes it difficult to use JupyterLab with a keyboard only.

Before this PR, any code cell editor in an open notebook was a tap trap, also known as a [keyboard trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html) or focus trap. After this PR, it is no longer a keyboard trap.

This PR introduces a noticeable visual change. Whenever a cell has browser focus, it is surrounded by a blue outline as shown in the following two screenshots:

| Light Theme | Dark Theme |
|-|-|
| JupyterLab light theme code cell focus ring | JupyterLab dark theme code cell focus ring |

## Long explanation

A quick reminder: in the JupyterLab codebase, the Notebook widget can be thought of as a collection of Cell widgets. Most Cell widgets contain an Editor widget (technically a CodeEditorWrapper widget). The relationship can be expressed as a CSS selector: `.jp-Cell .jp-Editor`.

This PR makes two significant changes to solve the notebook tab trap problem:

1. Removes tab-inclusive focusability from cell editors.
    - This means that after this PR, an end user can no longer enter cell editor fields using the tab key. They must use either the mouse or the `Enter` keyboard shortcut.
2. Removes tab-inclusive focusability from the notebook node in favor of adding tab-inclusive focusability to the active cell, plus adding tab-**exclusive** focusability to all other cells.
    - Previously, the user could press the tab key to move focus to the notebook node then press it again to move focus to the first code cell editor in the notebook. But with this change, the user bypasses the notebook node and tabs directly to the active cell.

What do I mean by tab-inclusive versus tab-exclusive focusability? Tab-inclusive focusability means setting `tabIndex = 0` (or higher) on a node. This allows the user to move focus to that node using the `Tab` key, in addition to tapping or clicking on it. Tab-exclusive focusability refers to setting `tabIndex = -1`. This allows the application to focus the node programmatically, as in: `document.getElementById('#notebook').focus()`.

Taken together, what's the end result of all of these focusability changes for keyboard-only users? It's that they can move focus to the active cell by pressing the tab key and get into that cell's editor box by then pressing the enter key (they can get out of the editor by pressing the escape key). They can move focus to other cells by pressing the up and down arrow keys. When they tab through the JupyterLab UI, they can easily return to the cell where they left off via tab navigation (because with this PR, the app only sets `tabIndex = 0` on the active cell and none of the other cells).

How did we get here?

The current library behind the code cell editor in the notebook widget is [Code Mirror 6, which has a solution for tab traps](https://codemirror.net/examples/tab/), namely: you press the `Escape` key immediately before pressing `Tab`. However, JupyterLab intercepts the `Escape` key, because it is a keyboard shortcut to change from "edit" to "command" mode in the notebook. Furthermore, when entering command mode, focus is put on the notebook widget's node (the one that contains all of the cell nodes). This means that when the user presses the `Tab` key after hitting `Escape`, they start from the beginning of the notebook, therefore never able to get past the code cell editor using only the keyboard.

My solution, therefore, was to remove tab-focusability from the CodeMirror editor. (Or more specifically, remove it only from the notebook configuration of the CodeMirror editor because CodeMirror is used in other places in JupyterLab.) This means that a user can still enter the cell editor via the mouse or the keyboard (by pressing `Enter` after moving to the cell with up and down arrow keys), but they cannot enter the cell editor by pressing the tab key. That way, the editor cannot trap the tab key.

Problem solved, right?

Well... Not quite. The current dev build of JupyterLab makes it hard to get back, with keyboard only, to editing a notebook if you leave it. This is because of a [change I made in PR #13415](https://github.com/jupyterlab/jupyterlab/pull/13415/files#diff-87e821a396b7aa41c021de1352d73a0f51cab272261d9a73356dadac6f3a6078R70). The intent of that PR was to make sure that every element in the JupyterLab UI has a [visible focus indicator](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html). Prior to that PR, the `MainAreaWidget` class added `tabIndex = 0` (tab-inclusive focusability) to any widget loaded in the main area. So my PR changed it to `tabIndex = -1`, removing the ability to tab-focus the node. My justification for doing that was that every main area widget I could find had some tab-focusable element inside of it, so it seemed there was no need to make the containing element tab-focusable.

But if you remove tab-focusability from cell editors, now you have notebooks without tab-focusable elements, which means it can be hard to get into, or back into the notebook using only the keyboard.

One seemingly easy way to get out of this knot is to make the notebook node tab-focusable again, but I didn't want to move backwards. I didn't want to solve this accessibility violation (focus trap) by re-introducing another (focus not visible).

Instead I started thinking about what would move the app towards better usability. From an accessibility perspective, it's not great to keep focus on the notebook node while a user presses up and down arrows to change the active cell, because if the browser-level focus is not changed then that information is not exposed to assistive tech. In other words, the JupyterLab app knows which cell is active and should be the target of certain commands, but the system does not. The app exposes that information visually via a blue bar, but the system literally cannot see it and neither can many blind or low-vision users. Therefore, it can become difficult for all but sighted users to know which cell is active and would be affected by pressing a keyboard shortcut.

So I decided to make the active cell tab-focusable. In order to conform to [WCAG 2.4.7 - Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html), that means that the active cell needs a visible focus indicator, which I added in this PR.

The shift from notebook-centric to cell-centric focus is a fairly big one, and it creates some issues that I have not solved, such as losing focus when a user scrolls an active cell so far out of the viewport that it gets popped off the DOM by the `WindowedList` class. This PR probably also introduces some bugs that slipped through my testing. But I think it moves the app's accessibility in the right direction, and the issues that it creates are smaller and less severe than the issues that it solves. I believe those issues can be solved in future PRs, but I did not want to indefinitely delay this PR.

## References

- I believe the "tab trap in code cell" issue was first described in #9399. As the issue mentions, this bug is a violation of [WCAG 2.1.2 - No Keyboard Trap](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap).
- (**External**) https://github.com/Quansight-Labs/jupyter-a11y-mgmt/issues/168

## Code changes

Fundamentally, the way the tab trap is removed in this PR is by turning off tab-focusability of Code Mirror 6 editor instances inside the Notebook widget.

Before this PR, when the user was in command mode, focus lived primarily on the Notebook widget node. Now focus shifts primarily among the notebook's cells, as the user moves around the notebook:

- Tab-inclusive focusability (`node.tabIndex = 0`) of Notebook widget removed.
- Tab-exclusive focusability (`node.tabIndex = -1`) added to Notebook cells, but `tabIndex = 0` if cell is active.
- All or nearly all command-mode notebook keyboard shortcuts modified so that they match any focusable, non-input element inside the notebook, rather than the focussed notebook widget node itself. One exception is the enter key, which only works when a cell is focussed (as opposed to something inside the cell).
- Many notebook actions modified so that the active cell is focussed after the action occurs.

## User-facing changes

- Getting into code cells
  - Before: pressing tab key could get you into a cell editor.
  - After: the only way to get into a cell editor is to enter edit mode by either tapping or clicking into the editor or pressing the enter key, when focussed on the cell that contains the editor.
- Focus
  - Before: focus tended to either be inside the cell editor (with focus indicator) or on the Notebook widget node (with no focus indicator).
  - After: notebook cells now receive focus. And when they are focussed using the keyboard, they have a big focus ring around them. It should now be clear when the cell versus the cell editor is focussed—or versus some button in the cell toolbar, or versus some element in the code cell output area.
- Keyboard shortcuts
  - Before: most notebook keyboard shortcuts would not work if some element inside the Notebook (e.g., cell toolbar) had focus.
  - After: most notebook keyboard shortcuts work so long as any non-input element inside the notebook is focussed. In other words, the user can be focused on anything inside a cell and still use the keyboard shortcuts on that cell, just so long as they're not in a text input of some sort, in which key presses should be interpreted as input characters rather than keyboard shortcuts.

None that I know of.

## Known issues

<details>
<summary>Focus is lost when scrolling.</summary>
This is due to the WindowedList class, which pops cell nodes off the DOM when they scroll out of view, which in turn causes those nodes to lose focus.
</details>

<details>
<summary>Scroll-into-view logic does not take into account outline width</summary>

<p>Update: I think this may have been fixed in #14407.</p>

When using J/K to move up and down the document, cells scroll into view but part of the outline remains outside the view. The scroll into view logic does not account for outline width. We need to pad it with a few pixels.
</details>

<details>
<summary>Cannot use escape (or ctrl m) from a debug input prompt</summary>
Typing Exit then hitting Enter does what I want -> so we just need to hook up the escape key to that? What is a debug input prompt, you ask? It's something you can activate, for example, if you run the following line in a code cell: `input()`.
</details>

<details>
<summary>Cell focus ring on mouse click</summary>
The heuristics that Chromium uses for when to apply `:focus-visible` are not working how I want/expect. I was hoping the focus ring would only show when using the keyboard, but in my experience so far, it shows up when I use a keyboard but also when I use a mouse, and I'm not entirely sure why. My suspicion is that it has something to do with the app's rather aggressive management of focus. For example, when a main area widget is loaded, its content node is focussed, then if a user presses a key, any key, a keydown handler calls a Notebook method that attempts to ensure focus is on the right node.
</details>


gabalafou:
I am putting this into draft mode until I add screenshots or videos and until I have had some time to do some additional testing.

packages/notebook/src/actions.tsx
```diff
@@ -1659,12 +1660,15 @@ export namespace NotebookActions {
...
    */
   export function collapseAllHeadings(notebook: Notebook): any {
+    const state = Private.getState(notebook);
     for (const cell of notebook.widgets) {
       if (NotebookActions.getHeadingInfo(cell).isHeading) {
         NotebookActions.setHeadingCollapse(cell, true, notebook);
         NotebookActions.setCellCollapse(cell, true);
       }
     }
+    notebook.activeCellIndex = 0;
```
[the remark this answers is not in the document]
gabalafou: Done: checked that this works if the first cell is hidden.

gabalafou:
I'm taking this out of draft because I think it's ready for review. 

However, before this PR is merged, I will need to:

- [ ] Update the screenshots
- [ ] Check the docs, update if necessary

gabalafou (commented):
Done self-reviewing!

packages/notebook/style/base.css
```diff
@@ -134,6 +134,14 @@
   background: transparent;
 }
 
+.jp-Notebook.jp-mod-commandMode .jp-Cell:focus {
+  outline: none;
```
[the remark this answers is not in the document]
gabalafou: The behavior was too buggy and erratic in Safari. I decided to just always show the outline when the cell is focussed.

fcollonval (commented):
Thanks @gabalafou 

I did not look yet in the details of this PR. This is more a UX question, I think the _focus border_ around cell seems heavy. Would it be better to change the active cell color when the notebook is not focused, like gray?

Pinging @GabrielaVives for comments.

gabalafou:
@fcollonval, I agree. On first instinct, I like the idea of graying the (normally blue) active cell bar when the cell is not focussed.

I'm guessing you meant something like the following screenshots.

Active, focussed:

screenshot

Active, not focussed:

screenshot

For accessibility, we may need to do more than just change the color of the bar (because communicating changes of state via color only is generally bad for colorblindness), which is why I gave the bar a dashed appearance in the second screenshot. To be clear, I don't think that a gray, dashed bar is the right solution (I realize that it makes it impossible to distinguish the two separate bars from each other), but there might be something workable in that direction. I bet @isabela-pf has thought of this before, so I'm inviting her to the discussion too.

fcollonval:
Yes this is what I meant and thanks to highlight color change is not enough.

GabrielaVives:
Hi @gabalafou @fcollonval , I've read through the PR but there are many technical considerations that I did not get. If understand right, from a user perspective a cell can have the following states:
- unfocused
- focused but not in edit mode
- focused and in edit mode ?

The issue being that when a cell is active (focused), there is no visual cue to differentiate it from the unfocused state?

gabalafou:
Hi @GabrielaVives, thank you for looking into this 😃

I think your summary of the three states *from the user perspective* is right. And yes, in these terms, the issue is that these three states are not visually distinguishable from each other (before this PR). With this PR in its current form as of today, the three states that you listed have the following visual differences:

- unfocussed = no black outline around the entire cell  
- focussed but not in edit mode = black outline around the entire cell
- focussed, in edit mode = outline around the edit box (no black outline around the entire cell)

The complicating factor is that there is another visual indicator for cells which is whether the cell is *active*. An active cell could be focussed or unfocussed. This distinction is important because for example, if you run the "run cell" command from anywhere in the UI, then it will execute the cell which is active. 

A cell that is focussed is active, but a cell that is active is not necessarily focussed. But focus state is important to indicate so that a user knows which part of the UI will respond to keyboard events. Does that make sense?

tonyfast:
when this gets merged it will clear a major blocker for the Notebook v7 accessibility. great work y'all.

ref: https://github.com/jupyter/notebook/issues/6948 https://github.com/jupyter/notebook/issues/6800

packages/notebook/style/base.css
```diff
@@ -137,6 +137,10 @@
   background: transparent;
 }
 
+.jp-Notebook.jp-mod-commandMode .jp-Cell:focus {
+  outline: solid var(--jp-inverse-layout-color0) 1px;
```
gabalafou, line 141: Hi @fcollonval, I tried to make the cell:focus outlines a bit less heavy, what do you think?

| Light Theme | Dark Theme |
|-|-|
| JupyterLab light mode | JupyterLab dark mode |
| #&#8203;757575 focus outline color | #&#8203;757575 focus outline color (yes, same as light theme) |
| #fff background | #&#8203;111 background |
| contrast ratio [4.6:1](https://webaim.org/resources/contrastchecker/?fcolor=757575&bcolor=FFFFFF) | contrast ratio [4.09:1](https://webaim.org/resources/contrastchecker/?fcolor=757575&bcolor=111111) |

To make the focus ring less heavy, I decreased the contrast of the outline with the background, and I also rounded the corners. 

In CSS, the changes look like this:

```suggestion
  outline: none;
  border: none;
  border-radius: 2px;
  box-shadow: 0 0 0 1px var(--jp-inverse-layout-color4);
  transition: 0.1s;
```

This idea came out of a short brainstorming session with @trallard and @steff456. We also considered the idea (brought up earlier in this PR's conversation history) of repurposing the current active cell indicator (the blue bar to the left of an active cell), but we were concerned that overloading that indicator could cause confusion (because it's not a common pattern).

If this still feels like too heavy of a UI change, then I think maybe the next best course of action is to go with @tonyfast's idea in the JupyterLab team call this week. We could punt on focus styling for now and keep working on it until we find a solution that both satisfies WCAG and yet isn't too intrusive to the existing UI, so that the rest of this PR can be reviewed for its main intended purpose, which is to fix the keyboard trap issue.

A note about `:focus-visible`... You might be wondering, if you've played with focus styles before, why not just use the `:focus-visible` selector instead of `:focus`. After all, the point of `:focus-visible` is to make a compromise between people who want the extra visual noise of a focus ring versus those who don't. Generally speaking, it's supposed to tell the browser: "hey browser, use these styles when you think the user really needs them (for example, when they can't use a mouse); otherwise don't show them."  So I tried that but I could not get consistent behavior in my browser (Chrome). The behavior I want is that if you load JupyterLab and you only interact with the UI using a mouse, no focus outline should ever appear on any notebook cell. But what I observed is that the focus ring would appear after page load when selecting via mouse. But then after I had clicked around the UI a few times, the focus ring would stop showing up. That level of inconsistency made me think that the only reasonable solution is to use `:focus` to turn the focus ring on all the time and keep things consistent. 

My hunch was that this had something to do with JupyterLab's programmatic focus handling. Sure enough, I was able to reproduce this behavior (Chrome, Firefox, Safari) in a code pen: [.focus() on mousedown coerces :focus-visible](https://codepen.io/gabalafou-the-looper/pen/vYQxMVd).

trallard: Nice! I think This looks definitely more subtle than the previous shades without making it unnoticeable and thus, defeating its purpose. 
I see no reason why this should not be merged, especially since the discussed alternatives (here) would probably cause more confusion among users and a focus ring is somewhat a more standard approach.

On the `focus-visible` note I've seen plenty of arguments against relying on this due to flaky/not consistent behaviours across web browsers.

tonyfast: it would be great to merge this because the keyboard UX is really blocking jupyter from a keyboard only experience. 

when this closes. it is probably worth opening an issue that summarizes the focus styling that has been done so we don't repeat past experiments. having a working focus indicator will be good in getting feedback about how to improve that styling. utlimately, users should be able to customize these preferences because we won't find a one size fits all solution. 

it is probably work including https://www.sarasoueidan.com/blog/focus-indicators/ in references, this document thoroughly explores focus styling across browsers.

tonyfast:
will this pull request address https://github.com/jupyter/notebook/issues/6948 ?

gabalafou:
I put a thumbs down on previous comment to mean that this PR does not address jupyter/notebook#6948.

However, there is an issue on my plate to[ fix all of the tab traps in JupyterLab](https://github.com/Quansight-Labs/jupyter-a11y-mgmt/issues/134). It may also address 6948, since Notebook 7 reuses many of the same JupyterLab components.

g547315:
Hi I was thinking on working on a keyboard shortcut to allow keyboard users to navigate the app easier, currently it takes a lot of tabbing for these users to navigate to the right tab bar. 

Proposed fix:

- Extend show right side bar command to move the tab focus to the first element of the right tab bar when the command is used
- add a keyboard shortcut for the new extended command and display it in the command palette 

Value add this issue:

- Serves as a hot-fix for users when they encounter the tab trapped in notebooks

tonyfast:
we discussed at one point merging this and saving style fixes for later. is that still the plan?

brichet:
This is really nice @gabalafou.

A lot of the test failures are due to the new focus border on cells, and can be fixed by updating the snapshots indeed.
Nevertheless some snapshot match failures seems to be relevant (i.e. 'Collapsible-headings.ReExpand Headers 01' in jupyterlab or 'general.notebook' in documentation).

I have some questions/remarks (I have no clue currently why and where it occurs):
- when switching from a notebook to another, the scroll position in the notebook is recalculated and changed. That may be annoying for peoples who work on several notebooks.
- it seems that the left and right panels icons cannot be reach anymore with tab
- I cannot move forward of the notebook using tab. As soon as I reached a cell, the tab switches between the cell toolbar's items and the 'add new cell' button. To access other menus, I have to move backward (shift + tab) from the first toolbar item. Is this expected ?

For a following PR probably: now we can reach the 'add new cell' button with tab, we should probably add a listener on the 'enter' key. Currently it leads to the cell edition.

krassowski (changes requested):
That's awesome, thank you for working on it. See comments for some minor things to improve/consider.

gabalafou (commented):
Thanks @brichet and @krassowski for your encouraging reviews!

My latest comments and commits should address everything you raised except the following:

- Tabbing and side bar (@brichet). I feel fairly confident that this PR has nothing to do with that. For what it's worth, yesterday I installed from source versus conda, and found the same discrepancy. 
- Schema (@krassowski). I think I will probably need to make another commit removing this field, but first I want to wait for your response to my question (see comments).
- UI snapshots (@brichet). I haven't looked into these yet but I will.
- Notebook shortcuts selector (@krassowski). For now, I used your suggestion to use `:read-write`. But please see my comment about further modifying the selector versus bubbling.

gabalafou:
I updated the screenshots in the PR description

packages/codemirror/src/extension.ts
```diff
@@ -778,6 +778,46 @@ export namespace EditorExtensionRegistry {
...
+        factory: () =>
+          createConditionalExtension(
+            EditorView.contentAttributes.of({
+              tabIndex: '0'
+            }),
+            EditorView.contentAttributes.of({
+              tabIndex: '-1'
+            })
+          ),
+        schema: {
```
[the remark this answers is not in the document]
krassowski: Yes, schema here is used to generate dynamic CodeMirror settings form. If you plan to add a setting in the future simple omitting schema (as done for example for `scrollPastEnd`) should be fine I think.

krassowski:
> So to me, the only manageable way to deal with this situation would be to allow anything in the cell output to handle the event first, then let it bubble up.
>
> In other words, what I really want to do is bind all of these keyboard shortcuts on the notebook node during the bubble phase. That way if the user presses the up arrow key, for example, while on some widget within the notebook, then if it's not handled by that widget, then it will bubble up until the notebook widget handles it.
> 
> But that seemed like too big of a departure from the way things are currently implemented. But I dunno. Should I change where these keyboard shortcuts get handled?

Yes, that's probably what needs to be done. Question is how feasible it is?

Also, as a datapoint, lumino has a [special logic for `data-lm-suppress-shortcuts`](https://github.com/jupyterlab/lumino/blob/af2ae00daeee993a73cbc964c7bf095c724a247c/packages/commands/src/index.ts#L509-L511) which could somewhat help.

gabalafou:
> Also, as a datapoint, lumino has a [special logic for `data-lm-suppress-shortcuts`](https://github.com/jupyterlab/lumino/blob/af2ae00daeee993a73cbc964c7bf095c724a247c/packages/commands/src/index.ts#L509-L511) which could somewhat help.

Yeah, I tried that too. I've tried everything except rewriting the whole thing to use bubble instead of capture phase. 😅

From some notes I took for myself back in January: 😅

> If you're familiar with Lumino, you might be thinking about adding the data-lm-suppress-shortcuts attribute to the output area. That would solve most of the issues. But it would also mean that while the user is focussed on any element inside the output area, none of the notebook keyboard shortcuts will work. For example, if the output area contains a link and the user tabs to that link and then presses the down arrow key, nothing will happen. Which means that the UX will diverge between links in rendered markdown versus links in code cell output. This discrepancy could create confusion for the user. They might think it's a bug: why would keyboard shortcuts like 'J' and 'K' work when focussed on one link in the notebook but not when focussed on another?

brichet:
> The thing about this that is maddening is that Lumino handles the keydown event during the capture phase and if any selector is matched, then it calls event.stopPropagation. So if you don't want to block some interactive output, you have to be very careful with your selector so that it doesn't accidentally match the interactive output.

Do we know if there is a reason for handling the keydown event on capture phase instead of bubble phase ? 
There is no comment on that change in git history, it has been pushed on main in https://github.com/phosphorjs/phosphor/commit/f4503e935343dafa71e7df1219c057c6227b5c28.

I'm tried to remove the flag and it seems to work with a simple javascript code catching the keydown event : if we stop propagation, the cell shortcuts are not called. I haven't tried with ipywidgets.

Thinking about it, I don't really understand why it is on capture phase, it can be very confusing if some shortcuts have the same command.

gabalafou:
> Thinking about it, I don't really understand why it is on capture phase

I was told in the [February 8 Jupyter accessibility call](https://jupyter-accessibility.readthedocs.io/en/latest/community/meeting-minutes/jupyterlab-accessibility-meetings/2023/index.html#february-8-2023) that it has to do with supporting Emacs-style "chord" keyboard shortcuts. For example, to save a file in Emacs, you press <kbd>Ctrl</kbd> and <kbd>X</kbd> together and then you press <kbd>Ctrl</kbd> and <kbd>S</kbd> together.

To be honest, though, now that you've asked the question and I'm thinking a bit more about it, I actually don't understand why processing the keydown event on the capture phase is absolutely essential to supporting chords. It seems to me that we could implement the same logic in the bubbling phase that Lumino does in the capturing phase.

It seems to me that using the capture phase must be more about control—for example, being able to guarantee that the shortcut to open the command palette (Ctrl + Shift + C) works everywhere in the app and no other widget can steal that shortcut away from the app, analogous to the way the browser doesn't allow you to override the new tab (Ctrl + T) shortcut (at least not on my Mac).

As an aside, I find the term "chord" really confusing to describe this kind of keyboard shortcut because in music a chord comes from pressing several notes or keys together, not in sequence. However, I found this same usage in a part of the [VS Code docs](https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-rules) that describes the notation for chord shortcuts: "Chords (two separate keypress actions) are described by separating the two keypresses with a space. For example, `Ctrl+K Ctrl+C`."

@afshin do you have any further insight into this question?

krassowski:
It would be handy to have a binder with the capture phase limitation lifted so we could try various shortcuts and libraries on it.

brichet:
> It would be handy to have a binder with the capture phase limitation lifted so we could try various shortcuts and libraries on it.

It could be nice indeed.
The only way I think of is to embed the lumino packages in a jupyterlab branch, using `yarn link` for the resolution.
Did you about another way?

fcollonval:
@krassowski @gabalafou @tonyfast Nicolas and I discussed about the status of this PR, we think that the code and the feat' are good to go and we should discuss the issue with keyboard shortcut event separately.

So how can we help merging it? And should we backport it - an API has been extended, but I would go for a backport as it improve tremendously the keyboard navigation?

krassowski:
I was trying to build the binder to share it on the call for testing today but it is failing. I will merge main branch into this branch to see if this helps.

gabalafou:
In today's team call, @krassowski mentioned that he's concerned that this PR allows the notebook widget to potentially "steal" keyboard shortcuts from cell outputs in ways that it couldn't before.

Here's a longer explanation of how this PR does that.

Previously, many notebook keyboard shortcuts (like pressing J/K to go up/down) were mapped to a `.jp-Notebook:focus` selector, which meant that the notebook widget node (rather than some node inside of it) had to have focus when the keyboard shortcut was pressed. After this PR, most of the shortcuts (including J/K) are mapped to `.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)`, which will match any non-read-write node inside the notebook when the notebook is in command mode. 

For example, if the user is focussed on a link in the notebook, pressing J/K will move focus to the cell above/below—because an `<a>` tag does not match `:read-write`. But if the user is focussed in, say, a Stdin prompt in the output area of a cell, J/K will be handled by the Stdin prompt—because `<input>` matches `:read-write`, so the negated selector fails. The concern here, therefore, is some extension or widget that gets put into the cell output area that has a focusable element that (1) does not match `:read-write` and (2) defines keyboard shortcuts that conflict with the notebook. Imagine, for example, an interactive data grid that allows the user to move between elements in the grid using arrow keys. In the current iteration of this PR, the app would intercept the up/down arrow keys and use them not to move about the data grid, but to move the **notebook** cell above or below.

krassowski:
Thank you @gabalafou for explaining in more detail. I went ahead and tested this scenario to drive the point home: this PR without further changes to keyboard shortcut event logic breaks certain keyboard interactions in ipydatagrid and likely other widget libraries:

| main branch | this PR | 
|--|--|
| main-branch |this-pr-2 |

gabalafou:
Thanks @krassowski for verifying that the concern is not just theoretical.

Here's some more background on the selector change and a couple ways I see that we could try to move forward.

So, why did I change the selector? 

This PR changes the app so that focus moves from cell to cell, rather than staying on the notebook node, so the old `.jp-Notebook:focus` simply no longer works.

So why not replace it with `.jp-Notebook.jp-mod-commandMode .jp-Cell:focus`? The reason is that I think from a UX perspective most notebook keyboard shortcuts should work even if the user isn't focused directly on the outmost, _.jp-Cell_ node. For example, if the user has focussed on some button in the cell toolbar and presses <kbd>D D</kbd>, that should delete the cell. Another example: a Markdown paragraph with links inside of it. I think if a user has tabbed onto a link in the Markdown cell, then presses the up or down arrows, then it should take them to the cell above or below. 

It's only when some widget _inside the cell_ has bound certain keyboard shortcuts that I think the notebook shortcuts should stop working. This is where the whole conversation around bubbling versus capturing comes in. If the app's keyboard shortcut paradigm were built around bubbling then this problem would be simple: the notebook widget would handle any keydown events that weren't handled (and stopped) by some other widget/node inside of it.

So what are some paths forward?

- Use the selector `.jp-Notebook.jp-mod-commandMode .jp-Cell:focus` and don't worry about notebook keyboard shortcuts working when the user is focussed on some other focusable thing in the notebook. This is basically how the app currently works. A lot of the notebook keyboard shortcuts only work if the notebook node itself has focus and do not work if the user is focussed on something inside of the notebook node, such as a link or a cell toolbar button. 
- Exclude the cell output area by using a selector like `.jp-Notebook.jp-mod-commandMode :not(.jp-Cell-outputArea *)`. The main thing I don't like about this solution is that it sets the stage for what will probably be perceived by end users as inscrutable UX inconsistencies. For example, if we use this selector, then notebook keyboard shortcuts like <kbd>J</kbd> or <kbd>K</kbd> won't work if the user has tabbed onto a link in the output area but they will work if the user has tabbed onto a link in some markdown. I doubt that for most users the reason for that inconsistency will be self-evident.
- Rewrite all of this to use bubbling. But this will require some effort, including, I think, a change to the Lumino _CommandRegistry_ class.

krassowski:
> Exclude the cell output area by using a selector like `.jp-Notebook.jp-mod-commandMode :not(.jp-Cell-outputArea *)`.

Just to add to this point, the inconsistencies described above are worse because cell output often takes more screen space than cell input, thus existing users often focus given cell (especially to perform command mode actions) by clicking on the cell _output_. In my view this solution would severely disrupt workflows of existing users (and as described above would be difficult to explain/understand for all users).

gabalafou:
> the inconsistencies described above are worse 

I'd have to test it, but I think it actually wouldn't be all that different from the current behavior of the app. If you look at the set of selectors that I changed in this PR, every single one of them (except collapse/expand all headings / `Ctrl Shift ArrowLeft/ArrowRight`) required the notebook node to have focus. 

Thus, in the current app, if the user clicks in the output area, so long as they don't click on something focusable in the output area, such as a Stdin prompt, then that click on the output area causes the notebook node (_.jp-Notebook_) to gain focus, and so all of the notebook keyboard shortcuts work.

Similarly, with this PR, if the user clicks in a non-focusable area of the output area, the click should cause the cell (_.jp-Cell_) to have focus, and the notebook keyboard shortcuts should work, right? ... since _.jp-Cell_ is not a descendant of _.jp-Cell-outputArea_, so the output-area-excluding selector should match when the keydown event is dispatched on the cell node.

krassowski:
Absolutely. But if `.jp-Notebook.jp-mod-commandMode :not(.jp-Cell-outputArea *)`  was used, it would no longer be the case which is what I was referring to. Does it make sense?

gabalafou:
Oh okay. I think I misread this line: "In my view this solution would severely disrupt workflows of existing users." I thought by "solution" you were referring to what I'm calling the output-area-excluding selector, and I thought by using the word "disrupt" you meant that applying the output-area-excluding selector would make things worse then the current state of the app.

fcollonval:
Hey sorry for the late contribution - if we were to implement an experimental flag on Lumino to switch the shortcut capture logic. Would it be enough to get this in and release in a test version?

krassowski:
> Hey sorry for the late contribution - if we were to implement an experimental flag on Lumino to switch the shortcut capture logic. Would it be enough to get this in and release in a test version?

After enabling this for JupyterLab and minor adjustments in this PR I think yes, this would be a good way forward.

fcollonval:
Thanks @krassowski

fcollonval:
#15142 is merged; @krassowski could you elaborate on what are the _minor adjustments in this PR_ ?

krassowski:
By minor adjustments I meant enabling the bubbling behaviour by default. I merged main, resolved conflicts and tested locally. I noted some issues:

- :red_circle: we are not testing the new behaviour on CI; I think that if we are to merge this PR `bubbling_keydown` added in  https://github.com/jupyterlab/jupyterlab/pull/15142 should become the default and all tests should pass with it
- :red_circle: the completer does not work - pressing tab just inserts a tab
- :orange_circle: navigating to the footer is no longer possible using arrow keys nor j/k; after tabbing into the footer it is possible to escape though; @gabalafou was this change intended? My concern is that users may have gotten used to being able to press <kbd>down</kbd> or <kbd>j</kbd> on the last cell to create a new cell (this is unrelated to bubbling change)
   no-longer-accessible
- :green_circle: when going up the focus indicator was occluded by multi-selection; I fixed it in https://github.com/jupyterlab/jupyterlab/pull/14115/commits/cb2bbbea9ce7486726f4fd6af896f3fe4839bd23
   Screenshot from 2023-09-27 09-07-38

fcollonval:
Thanks a lot @krassowski for pushing on this one. @brichet would you mind looking at the remaining points listed above by Mike?

brichet:
Please update snapshots

brichet:
Bumping CI

brichet:
I pushed some commits against this PR at https://github.com/brichet/jupyterlab/pull/111

These should fix the errors raised by @krassowski (thanks for the integration by the way), when capturing the keydown events on bubbling.

- the CI run on bubbling with a minor modification in *script/ci_install.sh* (so only for linux and mac, I assumed the integration test run on linux). This modification have to be removed if we decide to keep the event capture at bubbling phase by default.
- it doesn't fix the *create new cell* navigation with arrow keys, since I don't know if this is expected.

I'll do some tests, but I think we can remove the additional `:focus:not(:read-write)` in selectors, now the events are handled on bubbling.

@gabalafou I can either open a PR on your branch or push the commit in the current PR if you give me right, let me know.

krassowski:
PSA for anyone testing this: there is a bug in `bubbling_keydown` logic (it uses `PageConfig.getOption()` which always returns a string which is always truthy) which means that setting `"bubbling_keydown": false` in your page config will not actually disable the new behaviour.

brichet:
> PSA for anyone testing this: there is a bug in `bubbling_keydown` logic (it uses `PageConfig.getOption()` which always returns a string which is always truthy) which means that setting `"bubbling_keydown": false` in your page config will not actually disable the new behaviour.

Thanks @krassowski, I hadn't realized that `PageConfig.getOption()` only returns a string.

brichet:
Switching to bubbling mode brings some issues with selectors.

### Problem

- **Before this PR, events in capture mode**: almost all the shortcuts on notebook were on the selector `jp-Notebook:focus`, because the cells couldn't have the focus, so the focus was 
  - on the Notebook, and the shortcuts were working.
  - on an input, like`jp-InputArea-editor` or any input in the `jp-OutputArea`, and the shortcuts were not working.

- **With this PR, events in capture mode**: all these shortcuts' selectors have been changed to `.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)`, to work when the cell is focused but not an input in the cell. 
  This raises 2 problems:
  - we can not handle some keyboard events on output widgets which are not `:read-write`. These events can be caught by the notebook as a shortcut before it reaches the target widget.
  - this selector does not work if the notebook itself has the focus. Deleting several cells for example should not require having focus on a cell.

- **With this PR, events in bubbling mode**: we shouldn't have to care about focus anymore has the event is captured first by the element with focus, and goes up. So `.jp-Notebook.jp-mod-commandMode` should be enough to catch every shortcut on the notebook (not captured first by a widget). This work well because `codemirror` catches keyboard events on `jp-InputArea-editor`.
But it does not work for input elements in `jp-OutputArea`, because an input does not stop propagation by default, so the event  is caught as a shortcut by the notebook.

### Proposition

For me, the solution here is to have the 2 previous selectors for these shortcuts:
- `jp-Notebook:focus` when no cell has focus, only the notebook itself
- `.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)` when any element in the Notebook has the focus, but is not an input

But *lumino* does not allow commas in selectors.
We should therefore either duplicate the shortcuts in https://github.com/jupyterlab/jupyterlab/blob/main/packages/notebook-extension/schema/tracker.json, or allow commas in lumino.

Any opinion @gabalafou @krassowski @fcollonval ?

krassowski:
> Deleting several cells for example should not require having focus on a cell.

Agreed.

#### Alternative: place focus on a child node of notebook

Maybe we could have another node within the DOM tree, say `.jp-Notebook-content` which would take the focus instead of the notebook mode. In that case `.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)` would just work. I am not saying this is easy solution (it may be hard to refactor things to make it this way) but maybe worth a shot.

This could possibly be the existing `jp-WindowedPanel-window` element but we should probably add another class name to avoid making selectors depend on implementation detail.

#### Alternative: making selection first-class citizen

> this selector does not work if the notebook itself has the focus.

Maybe the focus should not fall back to the notebook node (or its immediate child as proposed above) when a cell is blurred while multi-selection is active? Maybe there should be a new element say, `.jp-Cell-selectionIndicator` which would be positioned in behind or beside the selected cells and which would receive focus instead?

It could for example take place of the existing backdrop (currently implemented as background of the cells) that would be now a standalone element (con: might be difficult to position):

Screenshot from 2023-10-03 09-14-14

Or we could add a new element, a "toolbar for multiple cells" that would allow these interactions using mouse (without the need to use context menu, mobile-device pointer-friendly); clicking on it could open a menu with multi-selection actions such as "delete cells", "duplicate", "cut" etc

image

Or it could be over the cells allowing to drag them easily:

image

Or it could be invisible without taking any role, just a focus-grabber:

Screenshot from 2023-10-03 09-14-14

The point is that if selection is not on a cell or its element it still probably should not be on the notebook element itself.

> Note I am not suggesting to implement the UI of the selection indicator proposed above in this PR, just making a point that if we add such an element it could indeed be handy in the future.

#### Thoughts on lumino solution

No objections towards supporting comma in lumino but FYI this would require also updates to specificity logic not only removal of the check. However, it would probably be better to find a solution which does not resort to that for simplicity sake.

brichet:
**self-contradiction**: an other alternative may also be to always have the focus on a cell (except when there is no cell). 
If the focus on the notebook itself is only available for these commands, maybe we could avoid it.
The user experience would remain unchanged, I think, since clicking outside a cell would allow commands to be executed without visual change (no visual focus on the cell), and tabbing would never fall on the notebook itself.

fcollonval:
about the multi selection case, I'm wondering if we should not only focus the active cell. This thinking comes from the behavior implemented in the listbox with multi selection example: Example 2: Multi-Select Listbox

---

@krassowski @gabalafou would it be useful to organize a specific technical meeting on this to get to a first version?

gabalafou:
> @krassowski @gabalafou would it be useful to organize a specific technical meeting on this to get to a first version?

If we don't have much to discuss in the team meeting tomorrow, I think we could take some time to talk about this PR and then decide if we want to schedule a specific technical call

gabalafou:
For reference, this is the [Example 2: Multi-Select Listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/#ex2_label) that I think Frederic was referring to

tonyfast:
can someone provide a TLDR for this multi select problem? 

y'all are definitely going to want to use the **[`aria-multiselectable`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable) parent.

the rub is that the container for the cells has the `role=feed` which doesn't support selection. multi-select support on assistive technology requires `grid`, `listbox`, `tree`, or `tablist` parents. it would probably be fine the change the cells container to `grid`, and that would require more ARIA on the children.

krassowski:
Implementation question: if we keep focus on the single cell in multi-selection, what happens in `full` windowed mode? Probably needs creating a follow-up issue.

brichet:
Following the discussion about shortcuts and focus in bubbling mode (during jupyterlab weekly https://github.com/jupyterlab/team-compass/issues/205#issuecomment-1758393808):

- With shortcuts selectors = `".jp-Notebook.jp-mod-commandMode"` 
  This is the current state, with https://github.com/brichet/jupyterlab/pull/111/commits/a8c342e6b842ec67b72c7377b85710f0f736da92 and https://github.com/brichet/jupyterlab/pull/111/commits/cf31c6b7b7c0e9475fd0260b104084071cebccdd, the selectors for shortcuts are not anymore on a child of the notebook, but on the notebook in *command mode* or any of its children.
  
  This means the commands are available if the notebook has the focus (which can still happen with a click outside of a cell) or if a cell has the focus, on any of the output element.

  The breaking with that implementation occurs when a cell output contain an input, because the inputs do not stop propagation by default. So typing in an input triggers the command (for example typing `a`, or `b`, or `d+d`)

- With shortcuts selectors = `".jp-Notebook.jp-mod-commandMode :focus:not(:read-write)"`
  The shortcuts works well when the cell or any of its output elements have the focus, but not if the notebook itself has the focus.

The quickest workaround to move forward is to duplicate the shortcut, as explained above (https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1744379784).

gabalafou:
This is great, thanks so much for your help!

I have a couple questions.

> The breaking with that implementation occurs when a cell output contain an input, because the inputs do not stop propagation by default.

So I guess our expectation is that *all other types of interactive cell outputs* must stop propagation of keydown events if they handle them and don't want them to trigger some notebook shortcut, is that correct?

> The shortcuts works well when the cell or any of its output elements have the focus, but not if the notebook itself has the focus.

I can't remember, do we need to support the notebook node itself having focus? If we don't, then we don't need duplicate shortcut selectors.

brichet:
> So I guess our expectation is that _all other types of interactive cell outputs_ must stop propagation of keydown events if they handle them and don't want them to trigger some notebook shortcut, is that correct?

This is correct. This can be seen in one of the failing regression test: *test/jupyterlab/notebook-max-outputs.test.ts > Display input value*

> I can't remember, do we need to support the notebook node itself having focus? If we don't, then we don't need duplicate shortcut selectors.

I wasn't sure about the conclusion. If we want to drop this behavior, we need to add a listener on the notebook node to handle the click event and focus on the active cell. then we can probably revert https://github.com/jupyterlab/jupyterlab/commit/a8c342e6b842ec67b72c7377b85710f0f736da92 and https://github.com/jupyterlab/jupyterlab/commit/cf31c6b7b7c0e9475fd0260b104084071cebccdd.
I can give it a try, and hope it won't break anything else.

krassowski:
Stepping outside of this PR some broader context

> If we want to drop this behavior, we need to add a listener on the notebook node to handle the click event and focus on the active cell. then we can probably revert https://github.com/jupyterlab/jupyterlab/commit/a8c342e6b842ec67b72c7377b85710f0f736da92 and https://github.com/jupyterlab/jupyterlab/commit/cf31c6b7b7c0e9475fd0260b104084071cebccdd.

:+1: I believe that we want to avoid ever having focus no notebook not only because of our discussion yesterday about multiselect and a cell being always present, but also because this is the cause of an annoying scrolling to the top of the notebook:
- https://github.com/jupyterlab/jupyterlab/issues/14106

> `:focus:not(:read-write)`

I have a strong suspicion that even if `:not(:read-write)` part is not currently needed overall, it would likely fix:
- https://github.com/jupyter-widgets/ipywidgets/issues/3836

krassowski:
> we need to add a listener on the notebook node to handle the click event and focus on the active cell

What is causing the notebook to receive focus? What if we set `tabIndex = -1` on the notebook node instead?

brichet:
> What is causing the notebook to receive focus? What if we set `tabIndex = -1` on the notebook node instead?

`tabIndex=-1` prevents the element to get focus with `tab`, but not with mouse click.
This is actually the case for any `MainAreaWidget`: https://github.com/jupyterlab/jupyterlab/blob/85d95861a9a51de210a3fd69e39b84786d1e61b6/packages/apputils/src/mainareawidget.ts#L70

If we simply remove the `tabIndex` property for the notebook node, clicking in the notebook outside of a cell git the focus to the `body` element (closest parent that can have focus). So the keyboard shortcut won't work.
This is why I think we should handle the click on the notebook node, to give the focus to the active cell.

tonyfast:
to get this work merged, i think we should judge the keyboard accessibility merits. the assistive tech and user experience should be broken into future work with constrained scopes.

> 👍 I believe that we want to avoid ever having focus no notebook not only because of our discussion yesterday about multiselect and a cell being always present

agreed, i don't think there is ever a reason to focus on the notebook ever. the notebook is not an interactive widget, it is a composite widget. even, navigating multiples notebooks you would put focus on an interactive element in the notebook. basically, by design, there is always one interactive element - at least the first text input of the first cell (maybe the schema should constrain a minLength == 1) - in an running notebook.

> My general rule is that you should probably ignore ARIA grid unless you are trying to recreate Excel. 
>>  [adrian roselli - ARIA Grid As an Anti-Pattern](https://adrianroselli.com/2020/07/aria-grid-as-an-anti-pattern.html)

since we are in excel territory, the [`grid`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role) pattern is one of the more consistent [`aria-multiselectable`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable) compliant patterns that make sense for the notebook. 

the `grid` turns a composite widget with many tab stops into a single tab stop. the tab pattern for the running, edit mode notebook is different that the reading state. the reading state has continuous tabbing. the grid role would represent a single tab stop, with focus managed by jupyterlab. this bubbling PR is essential to capturing.

it will take work to properly create a the `grid` pattern. that can be handled in future PRs.

more references:
* https://sarahmhigley.com/writing/grids-part1/
* https://sarahmhigley.com/writing/grids-part2/

brichet:
@tonyfast maybe your comment should be duplicated in https://github.com/jupyterlab/jupyterlab/pull/14768

gabalafou:
> This is why I think we should handle the click on the notebook node, to give the focus to the active cell.

While I agree that this is probably good app behavior to implement, in the interest of avoiding scope creep, I'm wondering what problem this would actually be solving. For users using only a keyboard, they won't accidentally click outside a cell. For users using a mouse, it should be relatively easy to click back to the cell they clicked away from. Am I missing something?

brichet:
> > This is why I think we should handle the click on the notebook node, to give the focus to the active cell.
> 
> While I agree that this is probably good app behavior to implement, in the interest of avoiding scope creep, I'm wondering what problem this would actually be solving. For users using only a keyboard, they won't accidentally click outside a cell. For users using a mouse, it should be relatively easy to click back to the cell they clicked away from. Am I missing something?

Indeed the user can click again on the active cell. But it would be a very strange behavior to not be able to perform action after clicking on the notebook (outside of a cell), in my opinion.

brichet:
The focus is now removed from the notebook node itself, and fall back on the active cell (if possible).
This involved some changes in the windowed notebook to keep the focus on an cell attached and displayed when scrolling.

These changes are worth reviewing.

packages/notebook/src/actions.tsx
```diff
@@ -843,7 +843,7 @@ export namespace NotebookActions {
    * This will skip any collapsed cells.
    * The existing selection will be cleared.
    */
-  export function selectAbove(notebook: Notebook): void {
+  export async function selectAbove(notebook: Notebook): Promise<void> {
```
krassowski, line 846: This (and `selecteBelow`) is a braking change in public API. Is it really needed, or can we keep it sync? It would be fine if we were targeting 5.0.

brichet: If we keep it sync but return a `Promise<void>` instead of nothing it's easy to fix, but it also change the public API (even if it is without consequences).

It's about the tests on these actions, but it can probably be fixed in the tests themselves.

packages/notebook/test/actions.spec.ts
```diff
@@ -1073,27 +1074,39 @@ describe('@jupyterlab/notebook', () => {
...
+        const waitForChange = signalToPromise(widget.stateChanged);
+        NotebookActions.selectAbove(widget);
+        // should timeout as there is no change.
+        await waitForChange;
+        expect(widget.activeCellIndex).toBe(0);
       });
 
-      it('should not wrap around to the bottom', async () => {
-        await NotebookActions.selectAbove(widget);
+      it.failing('should not wrap around to the bottom', async () => {
```
brichet, line 1092: Some tests are expected to fail because of timeout while awaiting the `stateChange` signal.
This was the only way I found to test that no changes occur without using async function, but the failing does not take into account the error thrown.

I'm interested in any better solution.

krassowski:
Just a note: we picked up a conflict on merge of https://github.com/jupyterlab/jupyterlab/pull/15286 which will need resolving, along with potential reverting of focus tracking on active cells. I think that other than that it is ready, is this correct?

brichet:
> Just a note: we picked up a conflict on merge of #15286 which will need resolving, along with potential reverting of focus tracking on active cells. I think that other than that it is ready, is this correct?

I also think that this PR is ready, except for the conflict.
I'll fix it in the next few days.

brichet:
@gabalafou @krassowski I think this PR is ready (until next conflict at least :smile:).

If you have time, can you take a look at it ?

krassowski (approved):
I think this is good to go. I do not like that there is a slight jitter when changing cells due to the focus indicator showing with a delay. This can be addressed in a follow-up PR or we could change the CSS so that it is not as visible for example using filled/empty active cell indicator (as it was in Notebook <6) instead.

Still I think there is more value in getting this to users for testing in a pre-release than in perfecting styles in this PR.

tonyfast:
yea it would be great to get a pre-release out of this sooner than later. we could also talk about a test plan for when it does drop. currently, we're massively limited in keyboard testing until this is resolved. so there is a to test that can't be tested now. we'll want to see how this work with widgets and extensions. the more pre release time the better. when we close this it should open a feedback issue.

brichet:
@fcollonval @krassowski 
I don't know whether the benchmark tests should be fixed or not. Otherwise we can merge this PR.

fcollonval (approved):

krassowski:
I wish the top-level comment of this PR contained description for why `bubblingKeydown` was enabled and some of the discussion around this. For reference some relevant comments in the thread are:
- https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1717361835 is puzzled as to why using the "capture" phase is essential for supporting chords; I don't know but https://github.com/jupyterlab-contrib/jupyterlab-vim/issues/138 _might_ be one example for chords being broken with "bubbling" approach
- https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1744379784 explains that with "capture" rather than "bubbling" mode deleting multiple cells when no cell is focused would not be possible (if we kept the current set of selectors used in this PR)
- https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1744466061 highlights that the only reason for enabling "bubbling" may be worked around by forcing focus on intermediary node in the DOM tree (between notebook and its content) - without modification of selectors

In general I feel like the assumption that "capture" mode was needed to support cords misses the point; it was needed to have higher priority over the CodeMirror native shortcuts, as seen in this PR (changes to `indentMoreOrInsertTab` and addition of `completerOrInsertNewLine`) and subsequent issues, recently [Jupyterlab 4.1 dedent broken #15821](https://github.com/jupyterlab/jupyterlab/issues/15821). The issue is that currently we have no mechanism to assert a priority of user-configured lumino shortcuts over the CM6 shortcuts (beside overriding them manually which is not user-configurable).

PR #14123, opened 2023-03-03
Use upstream `Stream` instead of downstream stream

afshin:
This removes a `TODO` by using the upstream `Stream` implementation from Lumino 2 instead of a local private implementation.

## User-facing changes
N/A

N/A


fcollonval (approved):
Thanks @afshin

fcollonval:
I reran the js-services tests to be sure - it is weird as I don't see any tests failing in the log.

PR #14128, opened 2023-03-05
Doc session

### LINKED ISSUE #14031: Damaged notebook (duplicate cells) in RTC-active hosted hub
## Description

We've activated RTC with Lab 3.6.1 again for one of my courses. So far all tests had gone well, including one where I had ~ 50 students join a session simultaneously. 

But today we experienced again notebook damage, I've put a copy of the damaged notebook [here](https://gist.github.com/fperez/4aaa805706fff6fe82141a104285a7d6) in case anyone wants to take a look. Not sure how much that will help.

## Reproduce

We don't know how to reproduce the problem, I'll describe the scenario where it occurred (Facu, please correct any inaccuracies I may report)

1. At 9am, @facusapienza21 logged into a hosted [JupyterHub that runs Lab 3.6.1 with collaboration on](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/stat159/config/common.yaml#L36) to teach a class, using his personal laptop. He opened a notebook, used it, and then moved on to other notebooks and activities.
2. Around noon, he went to his office, and logged into the same server from his desktop computer. When he visited the same notebook as before, it had duplicate cells.

In case it's useful, we also noticed two minor strange problems, but they may be unrelated. Mentioning them in case they ring any bells.
1. Facu saw some cells where he'd edit the Markdown but when rendering them, his edits would not show up on the rendered version. Yet they were correctly written to the file.
2. I logged into his server (admins can log in to one another's servers) and opened a terminal to do some debugging. His "running" left-panel tab did not show this new terminal, despite refreshing it several times and waiting. Only after doing a full browser tab refresh, did he see the terminal I'd opened. At that point, he was able to access it fine.

## Expected behavior

User notebooks are not corrupted by JupyterLab :)

## Context

- Operating System and version: Clients are MacOS, server is JupyterHub with Ubuntu 18.04, full config [here](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/stat159).
- Browser and version: Facu?
- JupyterLab version: 3.6.1

hbcarlos:
Fixes #14031

## References
* https://github.com/jupyterlab/jupyter_collaboration/pull/108
* https://github.com/jupyterlab/jupyter_collaboration/pull/113

## Code changes
Adds a document session to the WebSockerProvider to check whether the document in the client's memory is still valid.

## User-facing changes
If the user reaches a point where the document is invalid, the user will be prompted with an error dialog and forced to reload the page.

Captura de pantalla 2023-03-05 a las 12 47 10


packages/docprovider/src/yprovider.ts
```diff
@@ -110,14 +121,32 @@ export class WebSocketProvider implements IDocumentProvider {
...
+  private _onConnectionClosed = (event: any): void => {
+    if (event.code === 1003) {
+      console.error('Document provider closed:', event.reason);
+
+      showErrorMessage(
+        'Session expired',
+        'The document session expired. You need to reload this browser tab.',
+        [Dialog.okButton({ label: 'Reload' })]
+      )
+        .then(r => window.location.reload())
```
fcollonval, line 133: I oversight this one in `jupyter_collaboration`. But we should be consistent and not reload if the user did not want that (for example that will happen if it clicks outside the dialog):

```suggestion
        .then(r => { 
          if (r.button.accept) {
            window.location.reload();
          }
        })
```

packages/docprovider/src/yprovider.ts
```diff
@@ -110,14 +121,32 @@ export class WebSocketProvider implements IDocumentProvider {
     this._awareness.setLocalStateField('user', user.identity);
   }
 
+  private _onConnectionClosed = (event: any): void => {
+    if (event.code === 1003) {
+      console.error('Document provider closed:', event.reason);
+
+      showErrorMessage(
+        'Session expired',
```
fcollonval, line 129: Could please bring down the `translator` object as new optional constructor option so those strings can be translated.

yarn.lock
```diff
@@ -11304,6 +11304,13 @@ lib0@^0.2.31, lib0@^0.2.42, lib0@^0.2.49:
   dependencies:
     isomorphic.js "^0.2.4"
 
+lib0@^0.2.52:
+  version "0.2.63"
```
fcollonval, line 11308: Could you clean the dependencies tree so that we get only one version of lib0.

fcollonval (changes requested):
Thanks @hbcarlos 

I have some small comments.

packages/docprovider/src/yprovider.ts
```diff
@@ -37,8 +42,9 @@ export class WebSocketProvider implements IDocumentProvider {
     this._contentType = options.contentType;
     this._format = options.format;
     this._serverUrl = options.url;
-    this._ydoc = options.model.ydoc;
+    this._sharedModel = options.model;
     this._awareness = options.model.awareness;
+    this._trans = options.translator ?? nullTranslator.load('jupyterlab');
```
fcollonval, line 47: Small fix:

```suggestion
    this._trans = (options.translator ?? nullTranslator).load('jupyterlab');
```

hbcarlos: Ouch. I committed this change. But it was OK.

`options.translator` is already a TranslationBundle. We do not need to call `options.translator.load('jupyterlab')`. This is done in the activation function of the plugin.

fcollonval: oh yes sorry I forget about that.

fcollonval (approved):
Thanks @hbcarlos 

> Internationalization is expected to fail.

Issue #14083, opened 2023-02-25
 The class name of status bar items includes `undefined`


krassowski:
## Description

The class name of status bar items includes `undefined` which is wrong:

![Screenshot from 2023-02-25 16-30-19](https://user-images.githubusercontent.com/5832902/221368431-ad8a99d1-4fda-4989-969e-74697136ff58.png)

![Screenshot from 2023-02-25 16-37-04](https://user-images.githubusercontent.com/5832902/221368610-a93deab9-51fb-4189-9c50-3d2ba6177164.png)

## Expected behavior

No `undefined` class/correct class name added.

## Context

- JupyterLab version: 4.0.0a30

kushalnl7:
Hi, I am interested in this.

zrottman:
I think I have a fix for this, but I'm brand new to this project and to open source in general--would appreciate some feedback before attempting a PR.

Relevant code seems to be [jupyterlab/packages/statusbar/src/components/group.tsx](https://github.com/jupyterlab/jupyterlab/blob/aad5f3418da140a653481328728bc2490bb76d5f/packages/statusbar/src/components/group.tsx), lines 9-28. Here it is with proposed change:

```diff
export function GroupItem(
  props: GroupItem.IProps & React.HTMLAttributes<HTMLDivElement>
): React.ReactElement<GroupItem.IProps> {
-  const { spacing, children, className, ...rest } = props;
+  const { spacing, children, className = '', ...rest } = props;
  const numChildren = React.Children.count(children);

  return (
    <div className={`jp-StatusBar-GroupItem ${className}`} {...rest}>
      {React.Children.map(children, (child, i) => {
        if (i === 0) {
          return <div style={{ marginRight: `${spacing}px` }}>{child}</div>;
        } else if (i === numChildren - 1) {
          return <div style={{ marginLeft: `${spacing}px` }}>{child}</div>;
        } else {
          return <div style={{ margin: `0px ${spacing}px` }}>{child}</div>;
        }
      })}
    </div>
  );
}
```

As far as I can tell, the issue is that the extracted `props.className` property may be missing, in which case it is `undefined`. 

I am assuming that, if there is no `props.className` to extract, we do not want to append any additional class to the returned `<div>` element, is that right? In which case, my proposed solution is simply to pass a default value (an empty string) to `props.className` in the event there is nothing to extract.

Alternatively:

```diff
export function GroupItem(
[...]

PR #14140, opened 2023-03-06
bug: Add fallback value for undefined className prop in GroupItem component

zrottman:
## References
- Fixes [The class name of status bar items includes undefined #14083](https://github.com/jupyterlab/jupyterlab/issues/14083)

## Code changes
This pull request adds a fallback value to the className prop in the GroupItem component in order to resolve 'undefined' being added as a class. The fallback value, `className || ''`, provides an empty string as a default value for the className prop in case it's undefined or falsy. This ensures that the resulting className string will always be defined, even if the className prop is not provided.

```diff
export function GroupItem(
  props: GroupItem.IProps & React.HTMLAttributes<HTMLDivElement>
): React.ReactElement<GroupItem.IProps> {
  const { spacing, children, className, ...rest } = props;
  const numChildren = React.Children.count(children);

  return (
-    <div className={`jp-StatusBar-GroupItem ${className}`} {...rest}>
+    <div className={`jp-StatusBar-GroupItem ${className || ''}`} {...rest}>
     {React.Children.map(children, (child, i) => {
        if (i === 0) {
          return <div style={{ marginRight: `${spacing}px` }}>{child}</div>;
        } else if (i === numChildren - 1) {
          return <div style={{ marginLeft: `${spacing}px` }}>{child}</div>;
        } else {
          return <div style={{ margin: `0px ${spacing}px` }}>{child}</div>;
        }
      })}
    </div>
  );
}
```

## User-facing changes
None I'm aware of

None I'm aware of

*Note: I am brand new to open source and this is my first commit to a code base. I am trying my best to make this PR correctly, but please forgive me if I've made a mistake. I would be extremely grateful for any feedback, in general or specific to my proposed solution.*

### DIFF
--- a/packages/statusbar/src/components/group.tsx
+++ b/packages/statusbar/src/components/group.tsx
@@ -13,7 +13,7 @@ export function GroupItem(
   const numChildren = React.Children.count(children);
 
   return (
-    <div className={`jp-StatusBar-GroupItem ${className}`} {...rest}>
+    <div className={`jp-StatusBar-GroupItem ${className || ''}`} {...rest}>
       {React.Children.map(children, (child, i) => {
         if (i === 0) {
           return <div style={{ marginRight: `${spacing}px` }}>{child}</div>;


fcollonval (approved):
Thanks @zrottman and congrats for the first PR.

FYI you could have use the [nullish operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) `??`.

PR #14141, opened 2023-03-06
Show first line of cell when collapsed

### LINKED ISSUE #6512: Cell info when collapsed
I've been user of iPython Notebooks and Jupyter Notebooks for a long time, and now I'm trying the JupyterLab environment, and I feel it's really cool.

The collapse of a cell is a very good feature, specially if you write a lot of cells in the same Notebook, as I often do.

My suggestion is **to keep the first line of the cell visible** somehow, all the time or just when you hover over the three big dots of the collapsed cell.

This would permit to keep a info line, just a comment or so, like the name of the class or function (```class Dot:``` or ```def quadratic(x, y):```), or just a comment on the content of the hidden cell (```## Heavy geo calc here```), so the hidden cell content will be known without the need of open it.

Thank you in advance,

marthacryan:
## References
Fixes #6512 - although there are requests in the comments of that issue to add code *folding*, which we might want to make a separate issue for. 

## Code changes
Adds a new field to the Placeholder options for text to be displayed in the placeholder. 

## User-facing changes
Before collapsing:
image

After collapsing: 
image

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -585,7 +585,8 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
       callback: () => {
         this.inputHidden = !this.inputHidden;
       },
-      text: input.model.sharedModel.getSource().split('\n')[0]
+      text: input.model.sharedModel.getSource().split('\n')[0],
+      translator: this.translator
     });
 
     input.model.contentChanged.connect((sender, args) => {
@@ -1068,7 +1069,8 @@ export class CodeCell extends Cell<ICodeCellModel> {
       callback: () => {
         this.outputHidden = !this.outputHidden;
       },
-      text: this.getOutputPlaceholderText()
+      text: this.getOutputPlaceholderText(),
+      translator: this.translator
     });
 
     const layoutWrapper = outputWrapper.layout as PanelLayout;
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -414,6 +414,11 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
       const layout = this._inputWrapper!.layout as PanelLayout;
       if (value) {
         this._input!.parent = null;
+        if (this._inputPlaceholder) {
+          this._inputPlaceholder.text = this.model.sharedModel
+            .getSource()
+            .split('\n')?.[0];
+        }
         layout.addWidget(this._inputPlaceholder!);
       } else {
         this._inputPlaceholder!.parent = null;
@@ -581,14 +586,6 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
         this.inputHidden = !this.inputHidden;
       },
       text: input.model.sharedModel.getSource().split('\n')[0]
-    });
-
-    input.model.contentChanged.connect((sender, args) => {
-      if (this._inputPlaceholder) {
-        this._inputPlaceholder.text = sender.sharedModel
-          .getSource()
-          .split('\n')?.[0];
-      }
     });
 
     if (this.inputHidden) {
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -576,8 +576,20 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
     inputWrapper.addWidget(input);
     (this.layout as PanelLayout).addWidget(inputWrapper);
 
-    this._inputPlaceholder = new InputPlaceholder(() => {
-      this.inputHidden = !this.inputHidden;
+    this._inputPlaceholder = new InputPlaceholder({
+      callback: () => {
+        this.inputHidden = !this.inputHidden;
+      },
+      text: ((input.model.toJSON().source as string) ?? '')?.split('\n')?.[0]
+    });
+
+    input.model.contentChanged.connect((sender, args) => {
+      console.log(this._inputPlaceholder);
+      if (this._inputPlaceholder) {
+        this._inputPlaceholder.text = (sender.toJSON().source as string)?.split(
+          '\n'
+        )?.[0];
+      }
     });
 
     if (this.inputHidden) {
@@ -1040,8 +1052,13 @@ export class CodeCell extends Cell<ICodeCellModel> {
       this.addClass(DIRTY_CLASS);
     }
 
-    this._outputPlaceholder = new OutputPlaceholder(() => {
-      this.outputHidden = !this.outputHidden;
+    this._outputPlaceholder = new OutputPlaceholder({
+      callback: () => {
+        this.outputHidden = !this.outputHidden;
+      },
+      text: ((this.model.toJSON().outputs?.[0].text as string) ?? '').split(
+        '\n'
+      )[0]
     });
 
     const layoutWrapper = outputWrapper.layout as PanelLayout;
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -589,6 +589,14 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
       text: input.model.sharedModel.getSource().split('\n')[0]
     });
 
+    input.model.contentChanged.connect((sender, args) => {
+      if (this._inputPlaceholder && this.inputHidden) {
+        this._inputPlaceholder.text = sender.sharedModel
+          .getSource()
+          .split('\n')?.[0];
+      }
+    });
+
     if (this.inputHidden) {
       input.parent = null;
       (inputWrapper.layout as PanelLayout).addWidget(this._inputPlaceholder!);
@@ -1054,6 +1062,12 @@ export class CodeCell extends Cell<ICodeCellModel> {
         this.outputHidden = !this.outputHidden;
       },
       text: this.getOutputPlaceholderText()
+    });
+
+    this.model.outputs.changed.connect((sender, args) => {
+      if (this._outputPlaceholder && this.outputHidden) {
+        this._outputPlaceholder.text = this.getOutputPlaceholderText();
+      }
     });
 
     const layoutWrapper = outputWrapper.layout as PanelLayout;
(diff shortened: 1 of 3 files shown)


packages/cells/src/widget.ts
```diff
@@ -1040,8 +1051,23 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+      callback: () => {
+        this.outputHidden = !this.outputHidden;
+      },
+      text: (
+        ((this.model.toJSON().outputs as nbformat.IOutput[])[0]
+          ?.text as string) ?? ''
+      ).split('\n')[0]
+    });
+
+    this.model.contentChanged.connect((sender, args) => {
```
fcollonval, line 1064: This is wrong, you should listen for `this.model.outputs.changed`.

Could you also implement the same optimization as the one suggested for the input placeholder (aka only updating the placeholder if it is visible).

fcollonval (changes requested):
Thanks @marthacryan 

Would you mind adding style to differentiate between code cell and raw cell VS markdown cell - using or not monospace font?

Could you add test for all cell types?

packages/cells/src/widget.ts
```diff
@@ -1040,8 +1051,23 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
-    this._outputPlaceholder = new OutputPlaceholder(() => {
-      this.outputHidden = !this.outputHidden;
+    this._outputPlaceholder = new OutputPlaceholder({
+      callback: () => {
+        this.outputHidden = !this.outputHidden;
+      },
+      text: (
+        ((this.model.toJSON().outputs as nbformat.IOutput[])[0]
+          ?.text as string) ?? ''
+      ).split('\n')[0]
```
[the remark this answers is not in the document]
marthacryan: Not sure why but `this.model.outputs.get(0)` is returning undefined? Any idea why that might be happening?

marthacryan:

packages/cells/src/widget.ts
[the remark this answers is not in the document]
marthacryan: Update: `this.model.outputs.get(0)` actually is returning an `OutputModel` object, but `this.model.outputs.get(0)?.data` only has a field called `application/vnd.jupyter.stdout` - no `'text/plain'`.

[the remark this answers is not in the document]
marthacryan: Just pushed with that field but let me know if you want me to update it!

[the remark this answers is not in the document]
fcollonval: This is the problem with outputs; you don't know which mimetype will be available. I guess we should try the 3 following default text type:

https://github.com/jupyterlab/jupyterlab/blob/b69d2c3b91ac670b68f7bbf54b7d330060cd7e9c/packages/rendermime/src/factories.ts#L67-L73

If you could use directly `textRendererFactory.mimeTypes` it would be the best.

packages/cells/src/widget.ts
```diff
@@ -414,6 +414,11 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
       const layout = this._inputWrapper!.layout as PanelLayout;
       if (value) {
         this._input!.parent = null;
+        if (this._inputPlaceholder) {
+          this._inputPlaceholder.text = this.model.sharedModel
```
fcollonval, line 418: I would advice to restore the model change signal because in RTC the source may change due to another user.

packages/cells/src/widget.ts
```diff
@@ -1156,6 +1172,14 @@ export class CodeCell extends Cell<ICodeCellModel> {
         if (this.inputHidden && !this._outputWrapper!.isHidden) {
           this._outputWrapper!.hide();
         }
+        if (this._outputPlaceholder) {
+          this._outputPlaceholder.text =
```
fcollonval, line 1176: Same comment about listening for a change to support RTC.

fcollonval (changes requested):

packages/cells/src/widget.ts
marthacryan: I think I might not understand what you're going for - we can't make edits when the placeholder is shown (in a single-user context) so it would never update the placeholder text (for single users)

fcollonval: Yes this is true. My comment is exactly for the case of multiple users editing the same document. The content may be changed by other users.

packages/cells/style/placeholder.css
```diff
@@ -37,3 +37,8 @@
   box-shadow: 0 0 2px 0 rgba(0, 0, 0, 0.25);
   background-color: var(--jp-layout-color0);
 }
+
+.jp-RawCell .jp-Placeholder span,
+.jp-CodeCell .jp-Placeholder span {
```
krassowski, line 42: Please add a new class to the `span`; if after that you still need to combine selectors (hopefully not) please use descendant combinator.

packages/cells/src/widget.ts
marthacryan: Oh ok so do you think I should update it in the `set outputHidden` function how I am [here](https://github.com/jupyterlab/jupyterlab/pull/14141/files#diff-352916c4cc928b1298618fffc0424a71ee8816925f3a36c77d8593057e7b6f89R1181)?

packages/cells/style/placeholder.css
marthacryan: Done! Thanks

packages/cells/src/widget.ts
marthacryan: BTW added that back in - let me know how it looks!

packages/cells/src/widget.ts
```diff
@@ -1040,8 +1057,17 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+      callback: () => {
+        this.outputHidden = !this.outputHidden;
+      },
+      text: this.getOutputPlaceholderText()
+    });
+
+    this.model.outputs.changed.connect((sender, args) => {
+      if (this._outputPlaceholder && this.outputHidden) {
+        this._outputPlaceholder.text = this.getOutputPlaceholderText();
+      }
```
fcollonval, line 1070: Could you add this in `onOutputChanged` (~line 1370) instead of hooking a lambda here?

packages/cells/src/widget.ts
```diff
@@ -576,8 +582,19 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
...
-    this._inputPlaceholder = new InputPlaceholder(() => {
-      this.inputHidden = !this.inputHidden;
+    this._inputPlaceholder = new InputPlaceholder({
+      callback: () => {
+        this.inputHidden = !this.inputHidden;
+      },
+      text: input.model.sharedModel.getSource().split('\n')[0]
+    });
+
+    input.model.contentChanged.connect((sender, args) => {
```
fcollonval, line 592: Could you use a protected `onContentChanged` method instead of a lambda? And as it is defined in `MarkdownCell` make the necessary `super` call there.

fcollonval (changes requested):
Thanks @marthacryan this indeed what I was thinking. I have two minor suggestions to avoid using lambdas. And some tests are failing because you need to be very careful about the output handling. The string is actually usually an array of strings. Moreover you should distinguish the output depending on its `output_type`. For instance, `stream` output could also be displayed but they have a different structure compared to `execute_result` or `display_data`. You can look at the `examples/notebooks/OutputExamples.ipynb` e.g. or refer to nbformat.

marthacryan:
@fcollonval Thank you that example notebook was super helpful! I added some hard-coded output types, does that look ok?

packages/cells/src/widget.ts
```diff
@@ -1061,6 +1088,45 @@ export class CodeCell extends Cell<ICodeCellModel> {
     this.node.setAttribute('aria-label', ariaLabel);
   }
 
+  getOutputPlaceholderText(): string | undefined {
```
fcollonval, line 1091: This should not be public

```suggestion
  protected getOutputPlaceholderText(): string | undefined {
```

packages/cells/src/widget.ts
```diff
@@ -1061,6 +1088,45 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+    const outputData = this.model.outputs.get(0)?.data;
```
fcollonval, line 1092: The code below is wrong and too complex.

First the `output_type` is at the level of `this.model.outputs.get(0)`. Then depending on the output type you may have or not a `data` attributes.

So I would recommend you use a _switch-case_. In a structure like:

```
// Supported mimetype in order of preference
const supportedMimeTypes = [
  'text/html',
  'image/svg+xml',
  // ...
]
const first_output = this.model.outputs.get(0);

switch (first_output?.output_type) {
  // Treat those cases similarly
  case 'display_data':
  case 'execute_result': {
      const preferredOutput = supportedMimeTypes.find(mt => {
        const data = first_output.data[mt];
        return (Array.isArray(data) ? typeof data[0] : typeof data) === 'string'
      });
      const dataToDisplay = first_output.data[preferredOutput];
      return (Array.isArray(dataToDisplay) ? dataToDisplay : dataToDisplay.split('\n')).find(part => part !== '');
    }
  case 'stream':
    return (Array.isArray(first_output.text) ? first_output.text : first_output.text.split('\n')).find(part => part !== '');
  case 'error':
    return `{first_output.ename} {first_output.evalue}`;
}
```

fcollonval (changes requested):
@marthacryan thanks for the improvements. Do you think you could address the two latest comments by the end of next week? As this changes the API, we should get it before 4.0.0RC

packages/cells/src/widget.ts
marthacryan: Oh ok I can definitely add a switch. Thanks!

marthacryan: Ok update after looking into these changes - I'm not seeing any types of outputs that don't use the `data` field (and this is reflected in the type definition). I'm also not seeing an `ename` or an `evalue`, only the `application/vnd.jupyter.stderr` field. I agree that it should be simplified though and I pushed a change to simplify it.

fcollonval (approved):
Thanks @marthacryan

PR #14144, opened 2023-03-07
Switch to monthly dependabot update

fcollonval:
## References

Lots of maintenance time is spent dealing with dependabot update that are not critical (not security related). As we do not release the package weekly, I propose to switch to monthly update of Python packages and GitHub actions.

## Code changes

N/A

## User-facing changes

N/A

N/A


jtpio (approved):
Sound good, thanks!

PR #14147, opened 2023-03-07
Micro update terminal.rst: added xonsh

anki-code:
Hello! JupyterLab is awesome! 

I added small mention of the python-powered [xonsh shell](https://xon.sh) because with xonsh the shell in JupyterLab becomes supercharged. You can use python in the notebook as well as in the shell with xonsh. It's handy and easy. This is why xonsh mention is very appreciated here.

Thanks!

### DIFF
--- a/docs/source/user/directories.rst
+++ b/docs/source/user/directories.rst
@@ -87,7 +87,7 @@ application environment.
 
 The following configurations may be present in this file:
 
-1. ``terminalsAvailable`` identifies whether a terminal (i.e. ``bash/tsch``
+1. ``terminalsAvailable`` identifies whether a terminal (i.e. ``bash/tsch/xonsh``
    on Mac/Linux OR ``PowerShell`` on Windows) is available to be launched
    via the Launcher. (This configuration was predominantly required for
    Windows prior to PowerShell access being enabled in Jupyter Lab.) The
--- a/docs/source/user/terminal.rst
+++ b/docs/source/user/terminal.rst
@@ -3,7 +3,7 @@
 Terminals
 ==========
 
-JupyterLab terminals provide full support for system shells (bash, tsch,
+JupyterLab terminals provide full support for system shells (bash, tsch, xonsh,
 etc.) on Mac/Linux and PowerShell on Windows. You can run anything in
 your system shell with a terminal, including programs such as vim or
 emacs. The terminals run on the system where the Jupyter server is


fcollonval (changes requested):
Thanks for the suggestion @anki-code 

Unfortunately I'm not in favor of this one as the goal here is to mention common default shell on the various OS. As we do not endorse any specific shells, starting adding one means we should adding all; what will make the documentation overcrowded with low value information.

JasonWeill:
I have not used `xonsh` and I have no opinion about whether it's a good shell or not. I think it is not necessary to mention it in the documentation. Google Trends shows [virtually no search activity for "xonsh shell" compared with "bash shell" and "tcsh shell"](https://trends.google.com/trends/explore?date=now%201-d&geo=US&q=xonsh%20shell,bash%20shell,tcsh%20shell&hl=en). Searching for `bash shell` on Google yields about 505 million results, compared with about 415,000 for `tcsh shell` and about 22,200 for `xonsh shell`.

I would prefer to remove all named shells from the documentation in the interest of brevity, and to avoid pull requests like this in the future.

echarles:
> ... All as in real science ;)

Thx @anki-code for this PR and for the comparison.

I'd love us keeping mention of the shells and see this PR getting merged.

krassowski:
> I think it is not necessary to mention it in the documentation.

I agree with @JasonWeill here. The role of this part of documentation is to explain the core functions of JupyterLab, and not promote external additions (however great these would not be).

> I would prefer to remove all named shells from the documentation in the interest of brevity,

I do not agree with @JasonWeill here, as outlined in https://github.com/jupyterlab/jupyterlab/pull/14165#issuecomment-1464667011. Brevity is not a goal of our documentation - clarity and utility are; naming the most popular shells enhance clarity of the documentation and therefore these should remain.

> and to avoid pull requests like this in the future.

Similarly, we should not focus of maintainer experience in documentation but on user experience; as a maintainer I do not mind having to vote on PRs like these in the future; in my mind it is unlikely that any would be accepted but I would not rule it out.

> May be there is another way how to mention that xonsh as the shell is the bonus for JupyterLab?

@anki-code yes, there is! You are absolutely welcome to (in order of recommendation):
- write a dedicated post on [Jupyter Community Forum](https://discourse.jupyter.org/) in [Show and Tell](https://discourse.jupyter.org/c/general/show-and-tell/45) section
- present on monthly [Jupyter Community Call](https://discourse.jupyter.org/t/jupyter-community-calls/668)
- write a blog post on Medium and ask for it to be featured on [Jupyter Blog](https://blog.jupyter.org/)
   - this one may havea slightly higher threshold for relevance and quality, but I think it would be considered if you wrote it in a way which shows what are advantages very specific to Jupyter workflow

I believe that points above demonstrate that documentation is not the best place for the mention of another shell and since there are more suitable channels available I would vote in favour of closing this PR.

JasonWeill:
Closing this PR per @krassowski (related PR #14165 is also closed).

PR #14149, opened 2023-03-07
Clean up events service types in preparation for release

afshin:
This is a follow-on PR to the events service functionality to use the JSON object types that are used by other `@jupyterlab/services`.

## Code changes
- Makes explicit use of `JSONObject` and `ReadonlyJSONObject` to define `Event.Emission` and `Event.Request` types.

## User-facing changes
N/A

Updates the API from other pre-releases, but no release.


fcollonval (approved):
Thanks @afshin

Issue #14150, opened 2023-03-07
Documentation snapshots are flaky and failing on main branch


krassowski:
A faulty snapshots was merged in #13861 and needs to be reverted:
- `galata/test/documentation/general.test.ts-snapshots/file-formats-html-display-documentation-linux.png`

Also flaky snapshots changed, might cause problems:
- possibly `galata/test/documentation/general.test.ts-snapshots/interface-tabs-documentation-linux.png`
- `galata/test/documentation/general.test.ts-snapshots/interface-tabs-menu-documentation-linux.png`
- `galata/test/documentation/general.test.ts-snapshots/interface-jupyterlab-documentation-linux.png`

Finally, this one was also seen on another PR (#14090) and needs some work: `galata/test/documentation/general.test.ts-snapshots/interface-right-common-documentation-linux.png`

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/13861#issuecomment-1458591137_

PR #14151, opened 2023-03-07
Robustify documentation snapshot tests

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14150

## Code changes

- Add explicit wait for previous file to be opened before opening next file
- ensure the prompt is not hidden
- try focusing on input area editor not code mirror implementation detail
- accept the change to overview snapshot which accidentally brought the focus to notebook as this actually looks better (/closer reflects what the user would see).


krassowski:

galata/test/documentation/general.test.ts
```diff
@@ -625,7 +628,8 @@ test.describe('General', () => {
       '.lm-Menu ul[role="menu"] >> text=New Console for Notebook'
     );
 
-    await page.click('.jp-CodeConsole-input >> .cm-content');
+    await page.click('.jp-CodeConsole-input .jp-InputArea-editor');
+    await page.$('.jp-CodeConsole-input .jp-InputArea-editor.jp-mod-focused');
```
krassowski, line 632: Running out of ideas on this one (and unable to repro locally). It may be not a testing problem but a genuine race condition. If we want a quick fix, the worst case scenario is adding something like `sleep(200)`.

fcollonval:
Thanks @krassowski for looking into this - I realized that https://github.com/jupyterlab/jupyterlab/pull/13861 is actually introducing a bug; user is trapped in a notebook document as soon as one become active. I hope I nailed the reason in https://github.com/jupyterlab/jupyterlab/pull/14152.

galata/test/documentation/general.test.ts
krassowski: Resolved in https://github.com/jupyterlab/jupyterlab/pull/14152.

krassowski:
Closing as superseded by:
- https://github.com/jupyterlab/jupyterlab/pull/14152 and
- https://github.com/jupyterlab/jupyterlab/pull/14456.

PR #13925, opened 2023-02-05
Upgrade to TypeScript 5.0 beta

jasongrout:
## References

This carries forward work done in #13028 to update Typescript, this time to version 5.0 beta, as suggested in #13872.

The Typescript 5 [timeline](https://github.com/microsoft/TypeScript/issues/51362) indicates that 5.0 final is expected on 14 March, so during the bugfix phase of JupyterLab 4.0.

## Code changes

Upgrade Typescript to 5.0 beta.

## User-facing changes

Should be little to none.

## Backwards-incompatible changes

This upgrade seemed rather painless (thanks to the work from @jtpio and others in #13028!), so hopefully there are little if any backwards incompatibilities for us.

PR #14156, opened 2023-03-09
Fix create-package

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/13925

In that PR inquirer was updated to v9. That version is only distributed as ES module resulting in a broken usage in buildutils/src/create-package.ts

## Code changes

Use dynamic import of ES module in CommonJS code.

## User-facing changes

None

None


fcollonval:
I looked if there is a way to simply have a call with `--help`. But it is not supported. So creating a test won't be straightforward.

Issue #14088, opened 2023-02-26
Switching pages in extension manager "Discover" section resets its size


krassowski:
## Description

When switching pages in extension manager "Discover" section, the "Discover" section height gets reset:

![extension-panel-jumps-discover](https://user-images.githubusercontent.com/5832902/221425113-5f165cc8-ad87-44ed-9e36-ea2f5e86e95d.gif)

## Reproduce

As on the gif above

## Expected behavior

The height remains as set by the user.

## Context

- JupyterLab version: 4.0.0a34

FoSuCloud:
When updating data, it is necessary to maintain the height set by the user. Does it need to maintain the height set by the user after the browser is refreshed?

FoSuCloud:
For this bug, I think it is only necessary to maintain the height set by the user when updating the data.
Don't need to keep user set height after browser refresh right

krassowski:
> For this bug, I think it is only necessary to maintain the height set by the user when updating the data.
> Don't need to keep user set height after browser refresh right

Yes.

PR #14171, opened 2023-03-12
Keep extension manager panel sizes when pagination changes

FoSuCloud:
## References

fix https://github.com/jupyterlab/jupyterlab/issues/14088

## Code changes

packages/extensionmanager/src/widget.tsx

at _onDisclaimedChanged function cancel  setRelativeSizes
image

## User-facing changes

Maintain the height set by the user when updating the data

no

### DIFF
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -740,7 +740,6 @@ export class ExtensionsPanel extends SidePanel {
       (this.content as AccordionPanel).collapse(0);
       (this.content as AccordionPanel).expand(1);
       (this.content as AccordionPanel).expand(2);
-      (this.content.layout as AccordionLayout).setRelativeSizes([0, 1, 1]);
     }
   }
 
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -659,7 +659,7 @@ export class ExtensionsPanel extends SidePanel {
       (this.content as AccordionPanel).collapse(2);
     }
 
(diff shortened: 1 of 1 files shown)


FoSuCloud:
@meeseeksdev tag bug

packages/extensionmanager/src/widget.tsx
```diff
@@ -740,7 +740,6 @@ export class ExtensionsPanel extends SidePanel {
       (this.content as AccordionPanel).collapse(0);
       (this.content as AccordionPanel).expand(1);
       (this.content as AccordionPanel).expand(2);
-      (this.content.layout as AccordionLayout).setRelativeSizes([0, 1, 1]);
```
krassowski, line 743: Could we keep this line but update the `_wasDisclaimed` status so that none of these will execute? Something like:

```ts
     (this.content.layout as AccordionLayout).setRelativeSizes([0, 1, 1]);
  }
  this._wasDisclaimed = this.model.isDisclaimed;
```

Also, I would rename this method to `_onStateChanged` because in fact it is run on every state update which is likely why this bug has arisen.

FoSuCloud: Yes, your approach seems to be in line with the original design idea, but I don't understand what the function of line 743

FoSuCloud: Yes, your writing is correct to avoid unnecessary code execution, but I still don’t understand the function of line 743. It seems that the relative size setting only needs to be done during initialization

fcollonval (approved):
Thanks @FoSuCloud

Issue #13676, opened 2022-12-29
`Section` in running sidebar triggers React rerender on any tab switch even if hidden


krassowski:
## Description

On initialization of running sidebar `Section` (`PanelWithToolbar` subclass) a ReactWidget observing `options.manager.runningChanged` signal is created:

https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/running/src/index.tsx#L248-L266

![Screenshot from 2022-12-29 13-14-05](https://user-images.githubusercontent.com/5832902/209956434-f100a4a2-8d32-400a-a17b-bb647f154e6f.png)

There are two problems
- this signal fires too often leading to unncessary re-rendering when tab was switched (rather than closed/removed), which is because no filtering of the `tabsChanged` signal takes place - it is just proxied as-is:
   https://github.com/jupyterlab/jupyterlab/blob/552cdfe9314365d668c029e0863cf2fdbbbee4df/packages/running-extension/src/opentabs.ts#L85
- this signal is received and triggers re-render even if the running sidebar is closed; it should be proxied in the widget and skipped if widget `isHidden`.

## Reproduce

1. Record profile
2. Change tab
3. See that running sidebar was re-rendered

## Expected behavior

Hidden widgets do not re-render unnecessarily.

We could use `props.shouldUpdate` for `UseSignal`.

## Context

- JupyterLab version: 4.0.0.a31

krassowski:
> this signal fires too often leading to unnecessary re-rendering when tab was switched (rather than closed/removed), which is because no filtering of the `tabsChanged` signal takes place - it is just proxied as-is:

This one is difficult because the original signal comes from lumino's `title.changed` which emits without any data from changes to text, icon or `className`; it is reasonable that we listen to this signal because we may want to sync some of the title styling changes (e.g. text, icon, maybe colour in the future) in the "Open tabs" section of the running panel. The problem is that this `changed` fires on tab switch as a consequence of `className` update and we would need to make assumptions on what class name changes mean "tab switched" (and therefore are safe to ignore) and which should possibly need to be propagated.

I can also imagine a feature request of highlighting active tab in the panel so maybe this is not a priority.

PR #14172, opened 2023-03-12
Do not update running list when hidden

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13676

## Code changes

Filter our signal triggering updates if the section widget (or more accurately any of its parents) is hidden.

## User-facing changes

Slightly better performance on consecutive tab switch when running section is hidden.

Comparison using a workspace with 120 tabs open (exaggerating a bit for ease of visualisation - most workspaces are smaller).

### Before/after when running section active

Screenshot from 2023-03-12 13-34-25

### After when running section is hidden

Screenshot from 2023-03-12 13-35-06

While the performance benefit in this PR is modest, having it in will also help future profiling effort as the cleaner the timeline the easier it is to notice the things we can optimise.

None


packages/running/src/index.tsx
```diff
@@ -250,6 +251,90 @@ function List(props: {
...
+      translator?: ITranslator;
+    }
+  ) {
+    super();
+    _options.manager.runningChanged.connect(this._emitUpdate, this);
+  }
+
+  dispose() {
+    super.dispose();
+    this._options.manager.runningChanged.disconnect(this._emitUpdate, this);
```
fcollonval, line 269: The undo action should mirror the action order.

```suggestion
    this._options.manager.runningChanged.disconnect(this._emitUpdate, this);
    super.dispose();
```

packages/running/src/index.tsx
```diff
@@ -250,6 +251,90 @@ function List(props: {
...
+  }
+
+  onBeforeShow() {
```
fcollonval, line 272: Let's keep the default visibility

```suggestion
  protected onBeforeShow(msg: Message): void {
    super.onBeforeShow(msg);
```

packages/running/src/index.tsx
```diff
@@ -250,6 +251,90 @@ function List(props: {
...
+    this._update.emit();
+  }
+
+  render() {
```
fcollonval, line 276:
```suggestion
  render(): JSX.Element {
```

fcollonval (changes requested):
Thanks @krassowski

I have small comments and a general design question:

Should we make this (aka no update if hidden) part of `UseSignal`?

> I have no strong feelings moving forward with the code as is.

krassowski:
> Should we make this (aka no update if hidden) part of `UseSignal`?

That would be great but how could we do that? It seems to me that `UseSignal` does not know about the parent widget so we would need to add an argument passing `this` with reference to lumino widget which seems a bit problematic? We can discuss in an separate issue especially if this requires deeper changes.

fcollonval (approved):
Thanks @krassowski

PR #14173, opened 2023-03-12
Prevent memory leak when clearing outputs via context menu

krassowski:
## References

https://github.com/deshaw/jupyterlab-limit-output/issues/3#issuecomment-1465280675

## Code changes

Remove the reference to pointer event when then context menu closes.

## User-facing changes

Lower memory usage when clearing outputs after opening context menu (or closing file after opening a context menu on it, etc).

None.

### DIFF
--- a/packages/application/src/frontend.ts
+++ b/packages/application/src/frontend.ts
@@ -54,6 +54,12 @@ export abstract class JupyterFrontEnd<
       sortBySelector: false
     });
 
+    this.contextMenu.menu.aboutToClose.connect(() => {
+      // Make sure we do not keep a reference to DOM node that was hovered over
+      // when the context menu was last opened in event's `target` property.
+      this._contextMenuEvent = null;
+    });
+
     // The default restored promise if one does not exist in the options.
     const restored = new Promise<void>(resolve => {
       requestAnimationFrame(() => {
@@ -203,7 +209,7 @@ export abstract class JupyterFrontEnd<
     }
   }
 
-  private _contextMenuEvent: MouseEvent;
+  private _contextMenuEvent: MouseEvent | null = null;
   private _format: U;
   private _formatChanged = new Signal<this, U>(this);
 }


krassowski:
This is problematic and cannot be back-ported because the way we handle command execution - commands are not passed the event that triggered them in the first place _and_ the menu gets [closed before the command gets executed](https://github.com/jupyterlab/lumino/blob/06a87ead9b6b87842e8592fda8a15f27802c0646/packages/widgets/src/menu.ts#L310-L317):

```ts
    // Close the root menu before executing the command.
    this.rootMenu.close();

    // Execute the command for the item.
    let { command, args } = item;
    if (this.commands.isEnabled(command, args)) {
      this.commands.execute(command, args);
    }
```

which means that commands would have no way to access the event.

This PR would need an accompanying change to lumino menu implementation allowing us to pass the event in command args. Something like:

```diff
    // Close the root menu before executing the command.
    this.rootMenu.close();

    // Execute the command for the item.
-    let { command, args } = item;
+    let { command, originalArgs } = item;
+    const args = {...originalArgs, triggerEvent: event};
    if (this.commands.isEnabled(command, args)) {
      this.commands.execute(command, args);
    }
```

krassowski:
Alternatively we would extend Menu public API to emit a signal after the command was executed:

```diff
    // Close the root menu before executing the command.
    this.rootMenu.close();

    // Execute the command for the item.
    let { command, args } = item;
    if (this.commands.isEnabled(command, args)) {
-      this.commands.execute(command, args);
+      this.commands.execute(command, args).then(this.commandExecuted.emit);
    }
```

and then clean up the reference each time we receive `commandExecuted` signal, but this is a bad choice as it would lead to concurrency issues if commands take long to execute.

bollwyvl:
Good find!

While clearing the reference is great (and this should land), the larger question of what to do about events vs JSON serialization:

Perhaps an `IEventRegistry` could host a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)  that kept around references for "a while," which could be accessed externally by an id. Then the `eventId` could be passed into an `execute(args)`.

As discussed in the weekly call, this _doesn't_ help "real" JSON-encoded events from e.g. commandlinker (which shouldn't need them, anyway) or even crazier stuff like  [jyg](https://github.com/deathbeds/jyg).

Issue #13442, opened 2022-11-18
Jupyter Releaser v2 fails to publish the release (`ensure-sha` step)


jtpio:
## Description

Publishing the release fails when using the Releaser v2 workflows with the following:

```
RuntimeError: master current sha c33d6ef73933af0a701d018c24edf89f1651e3df is not equal to expected sha f8dee96054510d238ad7aa0997420c5bb85d8164
```

This is because the versioning scripts create two commits during the release: 

![image](https://user-images.githubusercontent.com/591645/202656472-2a14902b-bf8b-4d25-bb01-41f6bd3dcc09.png)

## Reproduce

Here is an example run for reference: https://github.com/jtpio/jupyter_releaser/actions/runs/3495078160/jobs/5851451020

## Expected behavior

A workaround is to skip the check when triggering the workflow:

![image](https://user-images.githubusercontent.com/591645/202656613-514fe01b-52ac-4f3c-979f-116a6ac7d2c4.png)

A better solution would be to handle this, for example by creating only one release commit. 

## Context

Happened while releasing `4.0.0a31`: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.0a31

jtpio:
Looks like the `check_release` workflow logs the same message, although it doesn't appear to make the run fail. For example: https://github.com/jupyterlab/jupyterlab/actions/runs/4262413995/jobs/7417886905

![image](https://user-images.githubusercontent.com/591645/221196740-3d557c70-d54f-4349-8d78-e327f6e337d8.png)

PR #14176, opened 2023-03-13
Add the releaser workflows to the repo

jtpio:
## References

Fixes https://github.com/jupyterlab/team-compass/issues/165

## Code changes

- [x] Add the Jupyter Releaser workflows to the repo so releases can more easily be made from https://github.com/jupyterlab/jupyterlab/actions
- [x] Skip `ensure-sha` for now https://github.com/jupyterlab/jupyterlab/issues/13442
- [x] Add `jupyterlab-bot` to the repo: https://github.com/jupyterlab-bot
- [x] Add `jupyterlab-releaser` to PyPI: https://pypi.org/user/jupyterlab-releaser
- [x] Add `jupyterlab-releaser-bot` to `npm`: https://www.npmjs.com/~jupyterlab-release-bot
- [x] Update `RELEASE.md` (update instructions and workflow names to releaser v2)
- [x] Add `ADMIN_TOKEN`
- [x] Add `PYPI_TOKEN`
- [x] Add `NPM_TOKEN`

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,55 @@
+name: "Step 2: Publish Release"
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "The target branch"
+        required: false
+      release_url:
+        description: "The URL of the draft GitHub release"
+        required: false
+      steps_to_skip:
+        description: "Comma separated list of steps to skip"
+        required: false
+        default: "ensure-sha"
+
+jobs:
+  publish_release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+
+      - name: Populate Release
+        id: populate-release
+        uses: jupyter-server/jupyter_releaser/.github/actions/populate-release@v2
+        with:
+          token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
+          target: ${{ github.event.inputs.target }}
+          branch: ${{ github.event.inputs.branch }}
+          release_url: ${{ github.event.inputs.release_url }}
+          steps_to_skip: ${{ github.event.inputs.steps_to_skip }}
+
+      - name: Finalize Release
+        id: finalize-release
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+          PYPI_TOKEN_MAP: ${{ secrets.PYPI_TOKEN_MAP }}
+          TWINE_USERNAME: __token__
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,55 @@
...
+      branch:
+        description: "The target branch"
+        required: false
+      release_url:
+        description: "The URL of the draft GitHub release"
+        required: false
+      steps_to_skip:
+        description: "Comma separated list of steps to skip"
+        required: false
+        default: "ensure-sha"
```
jtpio, line 14: Keeping `ensure-sha` as default here for now as I believe https://github.com/jupyterlab/jupyterlab/issues/13442 is still an issue with Releaser v2 (cc @blink1073 for confirmation).

blink1073:
I updated the tokens

.github/workflows/publish-release.yml
blink1073: Yes, that is correct

blink1073 (approved):
LGTM!

jtpio:
Thanks!

jtpio:
> I updated the tokens

@blink1073 curious which token has been set for `ADMIN_GITHUB_TOKEN`? Is it a fine grained token with an expiration time?

I tried starting a new release from the repo but kept getting `401` errors, even after adding myself as "Admin" to the repo directly. Could this be because the bot token is expired? https://github.com/jupyterlab/jupyterlab/actions/workflows/prep-release.yml

Starting the workflow from the fork seems to go past the auth step: https://github.com/jtpio/jupyter_releaser/actions/runs/4794420430/jobs/8527782598

jtpio:
Ah ok thanks for the extra context.

Then the 90 days expiration is unlikely to be the issue since it's been less than 3 months. Maybe it could be something related to how the releaser handles users that are part of the multiple teams which have different rights? Although that too used to be fine before iirc.

blink1073:
` RuntimeError: Could not get user permission level, assuming user was not admin!` means that the admin token itself did not work.  I created the "release token apr 2023" token 11 days ago, my guess is JupyterLab is using the expired "release token jan 2023" token.

jtpio:
Ah ok, it says it was updated March 13 here:

image

But it could indeed be an older one.

Would it be possible to update the token? I don't have access to the bot account.

Also what could be a way for handling expired tokens in the future?

jtpio:
ok thanks!

I'll open a new issue in the team compass to track this since this is something that might affect other repos in the `jupyterlab` org using the releaser bot.

jtpio:
Opened https://github.com/jupyterlab/team-compass/issues/190

PR #14181, opened 2023-03-13
Remove shebang from non-executable script node-version-check.js

### LINKED ISSUE #14131: Non-executable files with shebangs
This "problem" has been uncovered by linters for RPM packages during the packaging of Jupyterlab for Fedora Linux.

The reported problem is that `jupyterlab/staging/yarn.js` and `jupyterlab/node-version-check.js` both have shebang lines `#!/usr/bin/env node` but they are not executable.

Should we remove the shebangs or make the scripts executable?

frenzymadness:
Partially fixes #14131

### DIFF
--- a/jupyterlab/node-version-check.js
+++ b/jupyterlab/node-version-check.js
@@ -1,4 +1,3 @@
-#!/usr/bin/env node
 /*
  * Copyright (c) Jupyter Development Team.
  * Distributed under the terms of the Modified BSD License.


jtpio (approved):
Thanks!

Looks like CI does not complain so we should be able to go ahead and remove it.

fcollonval (approved):

frenzymadness:
> @frenzymadness were you packaging the JupyterLab 4 alpha releases or the stable 3.x versions (this is to know if this one will need to be backported)?

I've packaged the alpha version so from my point of view, there is no need for a backport.

PR #14183, opened 2023-03-13
Make codemirror state and view packages singleton

fcollonval:
## References

Adding a new extension in `jupyter-collaboration` to display the remote users selections/cursors I got hit by:

image

If I tell the builder that the codemirror package are singleton, it fails to find them.

## Code changes

Make codemirror state and view packages part of the shared singletons.

## User-facing changes

None

N/A

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -17,6 +17,8 @@
     "watch": "webpack --watch"
   },
   "resolutions": {
+    "@codemirror/state": "^6.2.0",
+    "@codemirror/view": "^6.7.0",
     "@jupyter/ydoc": "~0.3.4",
     "@jupyterlab/application": "~4.0.0-alpha.20",
     "@jupyterlab/application-extension": "~4.0.0-alpha.20",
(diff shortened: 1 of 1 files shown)


bollwyvl:
New codemirror, same old problems: are there any additional CM packages that be hoisted?

We also historically re-exported some of the actual members so they could be [consumed downstream](https://github.com/MarketSquare/jupyterlab_robotmode/blob/main/src/mode.ts#L791) without relying on webpack tricks: is that worth doing for the new stuff

Additionally: noted elsewhere, a lot of the CM modes are loaded at startup in a manner that blocks the app loading.... the python mode in particular is quite heavy. Perhaps we can we get back to lazy mode loading?

bollwyvl:
Maybe all of the non-mode packages?

```yaml
    "@codemirror/autocomplete": "^6.4.0",
    "@codemirror/commands": "^6.1.0",
    "@codemirror/language": "^6.4.0",
    "@codemirror/legacy-modes": "^6.3.0",
    "@codemirror/search": "^6.2.0",
```

bollwyvl:
And, potentially, the `lezer` stuff:

```yaml
    "@lezer/common": "^1.0.0",
    "@lezer/generator": "^1.0.0",
    "@lezer/highlight": "^1.0.0",
```

fcollonval:
Opening a new issue as follow-up to get these two early.

PR #14195, opened 2023-03-14
Improves page loading time by reducing server requests related to settings

brichet:
The page load is slowed down by several requests to get the settings from the server (detailed in  https://github.com/jupyterlab/jupyterlab_server/issues/380)

This pull request 
- avoids to fetch again the settings containing `jupyter.lab.transform=true`, not loaded when all the settings are fetched at first time.
- avoids to reload the settings from the server when the aim is to play the `transform()` methods.

Without the changes
image

With the changes
image

## References

This PR fixes a part of https://github.com/jupyterlab/jupyterlab_server/issues/380

## Code changes

- Adds a cache for the settings that have been already fetched but not loaded
- Adds an optional parameter to `SettingRegistry.load()` method to force playing again the `transform()` method, and to avoid using `SettingRegistry.reload()` when it's not necessary.

## User-facing changes

None

None


packages/settingregistry/src/settingregistry.ts
```diff
@@ -694,6 +720,8 @@ export class SettingRegistry implements ISettingRegistry {
       [phase in ISettingRegistry.IPlugin.Phase]: ISettingRegistry.IPlugin.Transform;
     };
   } = Object.create(null);
+  private _unloadedPlugins: { [name: string]: ISettingRegistry.IPlugin } =
```
fcollonval, line 723: Would you mind using a Map rather than an object? It will clarify the meaning?

fcollonval (changes requested):
Thanks @brichet

I have minor comments regarding adding documentation. And questions for other plugins with transformation. Did you have a look at those places that use the settings transformation - just to be sure we are not missing something?

- Context menu definition

https://github.com/jupyterlab/jupyterlab/blob/283074b1601a9318ad960d0ecf166c91bc32e48f/packages/application-extension/src/index.tsx#L1343

-  Main menu

https://github.com/jupyterlab/jupyterlab/blob/283074b1601a9318ad960d0ecf166c91bc32e48f/packages/mainmenu-extension/src/index.ts#L904

- Toolbar definition

https://github.com/jupyterlab/jupyterlab/blob/283074b1601a9318ad960d0ecf166c91bc32e48f/packages/apputils/src/toolbar/factory.ts#L215

packages/settingregistry/src/settingregistry.ts
```diff
@@ -284,6 +284,10 @@ export class SettingRegistry implements ISettingRegistry {
 
     // Preload with any available data at instantiation-time.
     if (options.plugins) {
+      options.plugins
```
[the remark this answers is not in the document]
brichet: Done in https://github.com/jupyterlab/jupyterlab/pull/14195/commits/3c03a7eb41349852b93e23e0b52ae479e5630c8a

packages/settingregistry/src/settingregistry.ts
```diff
@@ -359,7 +363,10 @@ export class SettingRegistry implements ISettingRegistry {
    * @returns A promise that resolves with a plugin settings object or rejects
    * if the plugin is not found.
    */
-  async load(plugin: string): Promise<ISettingRegistry.ISettings> {
+  async load(
+    plugin: string,
+    forceTransform: boolean = false
```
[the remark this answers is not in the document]
brichet: Done in https://github.com/jupyterlab/jupyterlab/pull/14195/commits/3c03a7eb41349852b93e23e0b52ae479e5630c8a

packages/settingregistry/src/settingregistry.ts
brichet: Done in https://github.com/jupyterlab/jupyterlab/pull/14195/commits/3c03a7eb41349852b93e23e0b52ae479e5630c8a

brichet:
Thanks @fcollonval for review.

> * Context menu definition

I don't think the settings are reloaded there.

> * Main menu

Same here

> * Toolbar definition

I updated it to avoid reloading the settings.

fcollonval (approved):
Thanks @brichet

packages/apputils/src/toolbar/factory.ts
```diff
@@ -220,9 +220,13 @@ async function setToolbarItems(
         await displayInformation(trans);
       } else {
         if (newItems.length > 0) {
+          // Empty the default values to avoid toolbar settings collisions.
           canonical = null;
-          // This will trigger a settings.changed signal that will update the items
-          await registry.reload(pluginId);
+          const schema = registry.plugins[pluginId]!.schema;
+          schema.properties!.toolbar.default = [];
```
jtpio, line 226: Wondering if this could be the cause of the issue noticed in https://github.com/jupyter/notebook/pull/6838#issuecomment-1500342416 ?

PR #14203, opened 2023-03-15
Suppresses header check inside frontmatter

### LINKED ISSUE #14200: Markdown table of contents doesn't handle YAML frontmatter
## Description

When editing a Markdown document with YAML front-matter, the Table of Contents shows only the last item of frontmatter.

image

## Reproduce

- Open the _Launcher_
- Click on _Markdown File_
- Open the _Table of Contents_ sidebar 
- Add a header
  ```md
  # Hello World
  ```
- See _Hello World_ in the _Table of Contents_
- Add some _YAML front matter_ before the header

  ```md
  ---
  foo: bar
  boo: baz
  ---

  # Hello World
  ```

- See only `boo: baz` in the Table of Contents

> This was found while working on a [JEP](https://github.com/jupyter/enhancement-proposals/pull/101) in JupyterLab.

By adding a newline above the second `---`, the problem _also_ goes away, so presumably it's treating that last item as an `h2` which counterfeits everything below it.

## Expected behavior

- Definitely see all the headers
- _Maybe_ see a collapsible table above/below the rest of the items
  - but definitely _not_ worth adding/faking a YAML parser
  - maybe just a chunk of YAML-formatted code (e.g. if the first line starts with `[^:]+:`

## Context

- Operating System and version: Linux Ubuntu 21.04 
- Browser and version: Chromium 110
- JupyterLab version: 4.0.0a39 (+ #14102)

JasonWeill:
## References

Fixes #14200.

## Code changes

Suppresses Markdown header check inside YAML frontmatter, defined as lines surrounded by `---` on a line by itself.

## User-facing changes

In the Markdown block below, `Summary` and `Motivation` are recognized as headings, and nothing in the frontmatter is.

```
---
title: Adopting a text-based diagram syntax in Jupyter Markdown
authors: (alphabetically) Nick Bollweg (@bollwyvl)
issue-number: 100
pr-number: 101
date-started: 2021-03-14
---

# Summary

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

# Motivation

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

None.

### DIFF
--- a/packages/toc/src/utils/markdown.ts
+++ b/packages/toc/src/utils/markdown.ts
@@ -82,10 +82,13 @@ export function getHeadings(text: string): IMarkdownHeading[] {
       continue;
     }
 
-    // Don't check for Markdown headings if in a YAML frontmatter block
-    if (line === '---') {
+    // Don't check for Markdown headings if in a YAML frontmatter block.
+    // We can only start a frontmatter block on the first line of the file.
+    // At other positions in a markdown file, '---' represents a horizontal rule.
+    if (line === '---' && (isFrontmatterBlock || lineIdx === 0)) {
       isFrontmatterBlock = !isFrontmatterBlock;
     }
+
     if (isFrontmatterBlock) {
       continue;
     }
--- a/packages/toc/src/utils/markdown.ts
+++ b/packages/toc/src/utils/markdown.ts
@@ -65,6 +65,7 @@ export function getHeadings(text: string): IMarkdownHeading[] {
   // Iterate over the lines to get the header level and text for each line:
   const headings = new Array<IMarkdownHeading>();
   let isCodeBlock;
+  let isFrontmatterBlock;
   for (let lineIdx = 0; lineIdx < lines.length; lineIdx++) {
     const line = lines[lineIdx];
 
@@ -73,11 +74,19 @@ export function getHeadings(text: string): IMarkdownHeading[] {
       continue;
     }
 
-    // Don't check for Markdown headings if in a code block:
+    // Don't check for Markdown headings if in a code block
     if (line.startsWith('```')) {
       isCodeBlock = !isCodeBlock;
     }
     if (isCodeBlock) {
+      continue;
+    }
+
+    // Don't check for Markdown headings if in a YAML frontmatter block
+    if (line === '---') {
+      isFrontmatterBlock = !isFrontmatterBlock;
+    }
+    if (isFrontmatterBlock) {
       continue;
     }
 
--- a/packages/toc/src/utils/markdown.ts
+++ b/packages/toc/src/utils/markdown.ts
@@ -65,8 +65,27 @@ export function getHeadings(text: string): IMarkdownHeading[] {
   // Iterate over the lines to get the header level and text for each line:
   const headings = new Array<IMarkdownHeading>();
   let isCodeBlock;
-  let isFrontmatterBlock;
-  for (let lineIdx = 0; lineIdx < lines.length; lineIdx++) {
+  let lineIdx = 0;
+
+  // Don't check for Markdown headings if in a YAML frontmatter block.
+  // We can only start a frontmatter block on the first line of the file.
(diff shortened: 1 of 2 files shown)


bollwyvl:
---
front: matter
---

^^^ that's some frontmatter

Ha, hadn't even looked at the implementation... figured it actually used the markdown parser or something. That's a lot of heuristics. But this is markdown, which seems to be _made_ entirely of heuristics.

The changed seems reasonable, but I imagine some more cases around the .rst-like/`setext` style of [headers](https://daringfireball.net/projects/markdown/syntax#header):


```md
---
front: matter
---

Not-front-matter
===

Really-not-front-matter
---

Not even looking
~~~

> ^^^ this is probably an alternate code block

```

bollwyvl:
Oh, and the (very whitespace-sensitive) "semantic break" that makes a `<hr/`>`:

```

---

```

JasonWeill:
@bollwyvl Since this is *front* matter, should we only do the frontmatter check once, on the very first line of the file? That would prevent the horizontal rule `---` from triggering this logic.

fcollonval:
thanks @JasonWeill 

> @bollwyvl Since this is _front_ matter, should we only do the frontmatter check once, on the very first line of the file? That would prevent the horizontal rule `---` from triggering this logic.

Yes we should only check for the very first line; switch to a logic:

1. Is there a header
2. Skip the header
3. Look for headings after the header

Another reason is that your current implementation will not work with the following case:

```
Title 1
---

Title 2
---
```

I'm pretty sure the current logic will escape the second title.

---

I'm :+1: for reusing the markdown parsing (especially as we are already parsing twice the string, once by CM and once by marked.js). But let's open a follow-up issue as it will require more complex changes.

bollwyvl:
Yep: I guess we're trying to match upstream GFM: here are some weird cases: https://gist.github.com/bollwyvl/84f19e4bcb3a6d5a8d23a26a5bcaf329

Here are some weird [edge cases](https://gist.github.com/bollwyvl/84f19e4bcb3a6d5a8d23a26a5bcaf329).

packages/toc/src/utils/markdown.ts
```diff
@@ -73,14 +74,25 @@ export function getHeadings(text: string): IMarkdownHeading[] {
...
       isCodeBlock = !isCodeBlock;
     }
     if (isCodeBlock) {
       continue;
     }
 
+    // Don't check for Markdown headings if in a YAML frontmatter block.
+    // We can only start a frontmatter block on the first line of the file.
+    // At other positions in a markdown file, '---' represents a horizontal rule.
+    if (line === '---' && (isFrontmatterBlock || lineIdx === 0)) {
```
fcollonval, line 88: Could we move this logic before the for loop line 69? Then you can enter the _regular_ parsing with `lineIdx` greater than 0 (aka after the header).

So the code will look like:

```typescript
// Skip frontmatter header if any

// ...

for (let lineIdx = lineAfterHeader; lineIdx < lines.length; lineIdx++) {
  // parsing for headings
}

fcollonval (changes requested):

fcollonval (approved):
Thanks @JasonWeill

PR #14211, opened 2023-03-16
Adds ids_only argument to the fetch settings request

brichet:
When the app is restored, the settings are requested again to use only the schema ids [here](https://github.com/jupyterlab/jupyterlab/blob/a396badbb6a3c840bdc088b22463fbcb42376bbe/packages/apputils-extension/src/settingsplugin.ts#L41).

This PR adds an optional argument to the `api/settings` request to retrieve only the IDs, and do not look into the schema and settings files.

The server side part of this feature is https://github.com/jupyterlab/jupyterlab_server/pull/385.

NOTE:
- this is backward compatible. If the server doesn't know this new argument, it will be ignored and the list of the settings and schema content will be returned as currently.

## References

Related to https://github.com/jupyterlab/jupyterlab_server/issues/380

## Code changes

Adds a new query option to the parameter of the `SettingsConnector.list()` and to the `service.settings.SettingManager.list()` methods.

## User-facing changes

None

None

### DIFF
--- a/packages/services/src/setting/index.ts
+++ b/packages/services/src/setting/index.ts
@@ -68,12 +68,14 @@ export class SettingManager extends DataConnector<
    *
    * @returns A promise that resolves if successful.
    */
-  async list(): Promise<{ ids: string[]; values: ISettingRegistry.IPlugin[] }> {
+  async list(
+    query?: 'ids'
+  ): Promise<{ ids: string[]; values: ISettingRegistry.IPlugin[] }> {
     const { serverSettings } = this;
     const { baseUrl, appUrl } = serverSettings;
     const { makeRequest, ResponseError } = ServerConnection;
     const base = baseUrl + appUrl;
-    const url = Private.url(base, '');
+    const url = Private.url(base, '', query === 'ids');
     const response = await makeRequest(url, {}, serverSettings);
 
     if (response.status !== 200) {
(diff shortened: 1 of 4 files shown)


packages/services/src/setting/index.ts
```diff
@@ -68,25 +68,34 @@ export class SettingManager extends DataConnector<
    *
    * @returns A promise that resolves if successful.
    */
-  async list(): Promise<{ ids: string[]; values: ISettingRegistry.IPlugin[] }> {
+  async list(
+    query?: 'ids'
+  ): Promise<{ ids: string[]; values: ISettingRegistry.IPlugin[] }> {
```
fcollonval, line 73: Should we make `values` undefined in case only the ids are requested to avoid confusion with an hypothetical empty plugin list?

```suggestion
  ): Promise<{ ids: string[]; values?: ISettingRegistry.IPlugin[] }> {
```

fcollonval (commented):
Thanks @brichet I have one question

packages/services/src/setting/index.ts
brichet: That would bring less confusion, I agree, but since the `SettingsManager` inherits from `DataConnector`, it should be also described in the `DataConnector` and its interface, I think.
That would change the API.

brichet:
I realized just now that the following warning should be removed as it would throw an error.

https://github.com/jupyterlab/jupyterlab/blob/67ba10e28bc72235c86ef6b0b2811ece1c85a85e/packages/apputils-extension/src/settingsplugin.ts#L51-L55

We'll lost a bit of information on the message, but I'm not sure to understand the condition about `jupyter.lab.transform=true`.

fcollonval (approved):
Thanks @brichet

PR #14182, opened 2023-03-13
Fix save as in collaborative mode

hbcarlos:
Fixes #14107

## References
This PR was started in #14130 as a fix for lab v3.6. I am opening a new PR because we must also fix it in lab v4.

## Code changes
After saving the document with a different name, we do not rename the document in memory. Instead, we close it and open the new one.

## User-facing changes
In collaborative mode, we can not rename the document in memory. After a "Save As" action, we need to open the new document and close the old one. By closing the document and opening the new one, we are losing the state of the old document. Most editors do this, and, in my opinion, it is the most predictable behavior.

See https://github.com/jupyterlab/jupyterlab/pull/14130 for more context.

## Screencast

https://user-images.githubusercontent.com/26092748/224733083-bc40840e-cd91-4bf3-8527-cb144b263e0c.mov

PR #14212, opened 2023-03-16
Fix save as without changing the file name

hbcarlos:
Follow-up #14182

I just realized there is an issue. When saving the document with the same name, the Drive emits a `save` event with the same name, so we close the document but do not open the new one.

This fix will prevent closing the document if we are not opening a new one.

https://user-images.githubusercontent.com/26092748/225632450-ec712710-3882-4c3f-b974-f637b02f1705.mov

cc @fcollonval

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -965,6 +965,7 @@ function addCommands(
             args.newValue &&
             args.newValue.path !== context.path
           ) {
+            void docManager.closeFile(context.path);
             void commands.execute(CommandIDs.open, {
               path: args.newValue.path
             });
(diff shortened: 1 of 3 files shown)


fcollonval (changes requested):
Thanks @hbcarlos 

If we do that, could we remove the logical change introduced in https://github.com/jupyterlab/jupyterlab/pull/14182/files#diff-bd91309321cead1ad13aa66f461c81110113c117fababf12ea04e346b0b5afb5

aka switching back to

```ts
if (this.isDisposed || !newPath) {
  return;
}
```

instead of rejecting the promise.

And could you add a jest test that if the path is the same, no `save` event should be triggered.

> I can help if you are not available.

fcollonval (approved):
Thanks @hbcarlos

fcollonval:

PR #14242, opened 2023-03-21
Exclude .yarn cache from the dist

### LINKED ISSUE #14241: jupyterlab `4.0.0a37` package is considerably bigger than jupyterlab `4.0.0a36`
The  4.0.0a37 package on PyPI is 96MB while the a36 is just 18MB

a37

a36

martinRenou:
## References

Fixes #14241

The last release of JupyterLab is 90Mb! https://pypi.org/project/jupyterlab/4.0.0a37/#files

This PR fixes it by removing the `.yarn` cache from the sdist and the wheel

Kudos to @trungleduc for spotting it

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,7 +180,6 @@ exclude = [
     "*.js.map",
     "/jupyterlab/staging/build",
     "/jupyterlab/staging/.yarn",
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

Issue #14252, opened 2023-03-21
Output prompt area is a large button


JasonWeill:
## Description

The entire output prompt area is a button that is clickable to toggle output scrolling. This makes it very difficult to drag and drop cells. This should probably be reverted until a different UI can be designed.

![image](https://user-images.githubusercontent.com/93281816/226767714-efb71462-fd3f-487b-b41a-de6575ab6c66.png)

Found by @ellisonbg during a bug bash.

## Reproduce

1. Run a cell that generates a lot of output
2. Hover over the area to the left of the output
3. Observe that the entire area to the left of the output and to the right of the blue cell bar is a clickable button

## Expected behavior

The blank area should be used as a drag handle, not entirely as a "toggle output scrolling" button

## Context

- Operating System and version: macOS
- Browser and version: Firefox
- JupyterLab version: 4.0.0a37

krassowski:
This was implemented in https://github.com/jupyterlab/jupyterlab/pull/14072 as feature parity with classic notebook and was one of the most requested features in JupyterLab issue tracker ever, with multiple users mentioning this as a blocker for their transition to JupyterLab. Saying that this does be reverted is a bit harsh. We could make it configurable with opt out setting. Also, dragging cella is still possible using the area to left.

krassowski:
Also there were 4 years for someone to make an alternative UI proposal ;)

JasonWeill:
We could use the `resize-ns` cursor to communicate that the button expands or collapses the output:

https://user-images.githubusercontent.com/93281816/227047739-6d68ba86-eaf2-4b32-b036-18849b28917a.mov

JasonWeill:
Or we could use the `zoom-in` and `zoom-out` cursors for a similar effect:

https://user-images.githubusercontent.com/93281816/227047847-e9a64ee0-ab85-4f54-9c24-583b40b1f6a1.mov

jtpio:
Agree with @krassowski this is a very valuable and useful feature to keep by default :+1: Thanks again for implementing it!

And also helpful for Notebook 7 (it works there out of the box).

For dragging cells there is already the input prompt which seems to be working fine.

JasonWeill:
Updated #14266 with the zoom cursors.

PR #14266, opened 2023-03-22
Adds buffer, uses zoom cursors for cell output collapser

JasonWeill:
## References

Partial fix for #14252. 

## Code changes

None (style changes only)

## User-facing changes

The large "button" to collapse cell outputs goes all the way to the left edge of the output, making it difficult to start selecting text at the left edge of the output. This change introduces an 8px buffer where the button is absent.

Before:

image

After:

image

In addition, we now use zoom in/out cursors when mousing over the big button:

https://user-images.githubusercontent.com/93281816/227047847-e9a64ee0-ab85-4f54-9c24-583b40b1f6a1.mov

None.


krassowski:
Would it be possibly better if it was implemented as an extra padding within the output rather than reduction in button width?

For comparison, classic Notebook did not have a buffer zone and had the button nicely align with the edge of the cell (as is currently the case but would be lost if we merge this PR as-is):

Screenshot from 2023-03-25 00-32-58

JasonWeill:
The issue that @ellisonbg showed me affects selecting the output starting at row 1, column 1 (or 0), which is a common use case when using [Jupyter AI](https://github.com/jupyterlab/jupyter-ai) tasks. Moving the mouse to the leftmost point of a line in the output makes the cursor very close both to the point where he'd need to click to start selecting and to the right edge of the "scrollable output" button. Clicking to start the selection and clicking to collapse the output are both possible. Adding about 8px more space would reduce this risk.

JasonWeill:
@krassowski Thoughts about the styling? Based on the feedback I got from Brian above, I'd like to keep about one character-width of blank space between the left edge of the output text and the right edge of the collapser button.

krassowski:
I wonder if this should be expressed in code by using calc with -1ch (not sure of current browser support here). Alternatively, a comment would be nice.
As for UX I'm not convinced (IMO internal padding would work too) and inclined to let this sit two more days to see if there is feedback from anyone else and if not we can merge in time for next beta iteration to get more eyes on it.

fcollonval:
@krassowski I'm trying to understand your suggestion about the padding. You mean using a padding-left on the output itself? If we go that road this means we should add a padding in the input cell to keep the alignment, isn't it? If so, it may feel awkward to increase the space between the input prompt and the editor.

I'm not convinced about the zoom cursor usage. It does not reflect what is happening. We could use the pointer cursor or a custom image (e.g. something like the expand/[compress](https://fontawesome.com/icons/compress?f=classic&s=solid) icons).

krassowski:
| current master | this PR as of March 28 | my proposal 1 |
|---|---|---|
| Screenshot from 2023-03-29 20-25-58 | Screenshot from 2023-03-29 20-27-13 | Screenshot from 2023-03-29 20-28-34 |

proposal 1 is to add functionally equivalent of:

```css
.jp-OutputArea-output {
    /* might need an `!important` or higher specificity selector */
   padding-left: calc(1 / 8 * var(--jp-cell-prompt-width));      /* 8px */
}
```

whereas the current style applied here is:

```css
.jp-RenderedText {
    padding-left: var(--jp-code-padding);      /* 5px */
}
```

and the only which aligns output text with input text:

| current master | this PR as of March 28 | my proposal 1 |
|---|---|---|
| Screenshot from 2023-03-29 20-25-58 | Screenshot from 2023-03-29 20-27-13 | Screenshot from 2023-03-29 20-28-34 |

My proposal 2 is visually equivalent to proposal 1 but would be implemented as:

```css
.jp-OutputArea-output {
    /* might need an `!important` or higher specificity selector */
   padding-left: 1ch  /* 1 character; 7.827px for me */
}
```

which is exactly what you say Brian was suggesting :)

JasonWeill:
@krassowski Made the changes you describe; things look better now! (And there's no longer the problem that my changes were only applicable in the provided light theme.)

krassowski:
bot please update snapshots

fcollonval:
Kicking CI

fcollonval:
Thanks @krassowski for the details explanation and screenshots.

Pinging @GabrielaVives for her thoughts on the usage of zoom cursors.

JasonWeill:
Are we OK to merge this change in as written? I think adding an icon is a good improvement, and I could add an enhancement to put the icon that appears on hover, if we're good to merge the current change.

krassowski:
Ok, looking at screenshot changes (e.g. `galata/test/documentation/general.test.ts-snapshots/file-formats-altair-documentation-linux.png`) I am not sure if maybe we should set 1ch only on `.jp-RenderedText` and not on `.jp-OutputArea-output` which also applies to plots. What do you think?

As for icon on hover: I like this idea, but it feels we should either have icon on hover, or change the cursor not both. Since we have more control over icon than cursor maybe icon is a better way forward (though cursor does have the benefit of instantly telling user when they moved over the boundary where action changes without having to move sight to icon which would show at some distance from cursor).

ellisonbg:
The usability issue with the cursor approach is that 1) there isn't a cursor that means "click to enable or disable scrolling." and 2) cursors are designed to be used on the object itself (for example, you would see a zoom in cursor style when your mouse in on top of an image that you can zoom by clicking). A stateful button gives us a lot more degrees of freedom to design something that better matches the functionality. Granted, that still isn't an easy task, as there isn't a well established icon that means this. I would probably remove the cursor styling from the PR until we can figure that out, but the 1ch padding is an improvement.

krassowski:
> Granted, that still isn't an easy task, as there isn't a well established icon that means this. 

I do like the icons proposed by @GabrielaVives. The other problem is the placement of the icon when the output is long. If the output is 10k pixels long and I am in the cell below I would like the icon to show at the bottom to reassure me of the action I am about to perform. If I am at the top it seems more reasonable to have it at the top. I think there was another comment where you mentioned using fixed positioning which sounds good; an alternative would be to snap it to `y` coordinate of the cursor.

ellisonbg:
This is why we originally implemented this as a context menu item - you can right click anywhere on the cell and perform the action efficiently. But the idea of follow the y coordinate of the cursor is definitely worth exploring.

ellisonbg:
Another minor change that might help the usability of this would be to slightly reduce the severity of the inner shadow. We need some way to show which part of the page scrolls, but the shadow we are using is a lot more than is typically used for this type of situation. We could possibly even move away from the shadow and explore other visual treatments (borders, etc.).

fcollonval:
> Ok, looking at screenshot changes (e.g. `galata/test/documentation/general.test.ts-snapshots/file-formats-altair-documentation-linux.png`) I am not sure if maybe we should set 1ch only on `.jp-RenderedText` and not on `.jp-OutputArea-output` which also applies to plots. What do you think?

I agree; it may be better to only apply this to text.

@JasonWeill what do you think?

andrii-i:
@fcollonval will try implementing the suggested changes and will post screenshots here.

andrii-i:
@GabrielaVives I think using icon you shared is a good option. What is the icon used here? If this is something you created, can we use it?

krassowski:
I would suggest splitting the icons work into a separate PR. The increase of padding for text outputs appears uncontroversial, but icons may need iterations and we would end up rebasing/updating screenshots many times with detriment to readability of the PR discussion.

andrii-i:
Here's how it looks with `padding-left: 1ch` applied to `.jp-RenderedText` only:
Screenshot 2023-04-07 at 11 54 21 AM

With `padding-left: 1ch` applied to `.jp-OutputArea-output`:
Screenshot 2023-04-07 at 11 53 40 AM

andrii-i (approved):

andrii-i:
Please disregard the review above, was a missclick

andrii-i:
Because I can't directly push to this branch, to optimize the process I created #14344. #14344 is a copy of this PR with merge conflicts solved and that I can push to. @krassowski @fcollonval, please give it a look when possible.

JasonWeill:
Closes in favor of #14344.

PR #14269, opened 2023-03-24
Fix color contrast of gutter line numbers in code mirror

### LINKED ISSUE #14254: Low contrast for line numbers' font color
## Description

The font color for line numbers in text / code files has a contrast ratio too low for WCAG requirements in both Light and Dark theme.

Screenshot 2023-03-20 at 3 53 38 PM

This was found during a bug bash by @andrii-i.

andrii-i:
## References

- Fixes #14254

## Code changes

Add css rule targeting cm-gutter line numbers that sets color to `var(--jp-ui-font-color2)` matching gutters background color  `var(--jp-layout-color2)`

## User-facing changes

Line numbers in codemirror gutter are now using `var(--jp-ui-font-color2)` font color matching background color  `var(--jp-layout-color2)`. Numbers are more readable in Jupyter Dark theme, line numbers font color would now change with var change when different themes are used.

Contrast in light theme: 4.44 (was 4.52)
Contrast in dark theme: 4.24 (was 1.91)

Without this change:
Screenshot 2023-03-24 at 1 40 25 AM
Screenshot 2023-03-24 at 1 40 31 AM
Screenshot 2023-03-24 at 1 39 57 AM
Screenshot 2023-03-24 at 1 40 07 AM

With this change:

Screenshot 2023-03-30 at 1 13 46 PM
Screenshot 2023-03-30 at 1 13 54 PM
Screenshot 2023-03-30 at 1 13 11 PM
Screenshot 2023-03-30 at 1 13 24 PM

Note that in for example VSCode, all line numbers also don't have a proper contrast all the time, only line number of the "active" line (line at which text editing caret is positioned) changes the color to have proper contrast (see screenshot below). This is a good approach but it would require deeper changes in codemirror / jupyterlab as currently there is no "active line" state (if there is and I missed it, please let me know).
Screenshot 2023-03-24 at 2 17 44 AM

None


andrii-i:
Bot please update snapshots

andrii-i:
Kicking CI

krassowski:
Do you need WCAG **AAA** (rather than WCAG AA) compliance? Is there a law you are trying to comply with or is it just an attempt to make things _better_? Do you think that the UI would be usable if every component had 7:1 contrast? (I don't).

At numerous occasions I am raising the issue that nothing is free and increasing contrast has costs:
- causing focus issues as element hierarchy is weakend
- reducing the range of colours used causing issues for colourblind and partially colourblind users
- potentially causing migraine issues for certain groups of users

Finally, contrast checkers used in WCAG 2.0 are flawed as described in https://github.com/jupyter/accessibility/issues/65

Given the above, should we possibly work more towards meeting WCAG AA contrast (this is only adjusting contrast in dark theme as that one is bad indeed) and instead focus on implementing an AAA compliance using  [G174 technique](https://www.w3.org/WAI/WCAG21/Techniques/general/G174) that is having a dedicated accessible theme that I believe @steff456 is working on?

Another implementation idea along G174 technique spirit: we could keep focus on 4.5:1 for the default state and add a stronger colour on `:hover` over the lines gutter.

andrii-i:
@krassowski thank you for the extensive comment, will reply on Tuesday of the upcoming week after coming back to work

krassowski:
Another option would be to use [`@media (prefers-contrast: more)`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast). [Examples](https://www.tempertemper.net/blog/using-the-increased-contrast-mode-css-media-query), [can I use](https://caniuse.com/mdn-css_at-rules_media_prefers-contrast)

andrii-i:
@krassowski My primary reasoning for this change is to make line numbers readable as currently they are not. My reasoning for following [WCAG AAA (Enchanced)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html) vs [WCAG AA (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) for text contrast was to make JupyterLab more accessible for a broad range of people (in particular, with low vision and color deficiency).

At the same time https://github.com/jupyter/accessibility/issues/65#issuecomment-1051185398 notes that there is a legal dimension to this too:

> there is a legal dimension of this because of laws such as Section 508 (the US law that governs accessibility and which many institutions are subject to). Section 508 currently refers to WCAG 2.0, but other jurisdictions have moved to WCAG 2.1 or have language about the "latest WCAG version." (such as CA). Because Jupyter is used extensively in educational and research institutions that are subject to these laws, I think it makes sense for us to track the latest official version of the WCAG standard.
>
> I think this is important enough that it should be written into the (yet to be written) charter of the accessibility working group that would eventually be approved by the Jupyter governing body (see https://github.com/jupyter/governance/pull/121 for a description of working groups in the new governance model).

andrii-i:
After looking more into the topic, I can see that there's an [Accessibility project ](https://github.com/orgs/jupyterlab/projects/1#card-62801310). 

@manfromjupyter, @isabela-pf, maybe you could review this PR if possible or just chime in with the comment?

steff456:
Hi @andrii-i! Thanks for opening this PR, right now I'm working in creating accessible themes for Jupyterlab as an extension, the goal is to prioritize [WCAG AAA (Enchanced)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html) and if it is not possible, to at least ensure [WCAG AA (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html). 

I think either way is fine, complying with either WCAG AAA or AA will make the interface better. It is just important to check all the possible states and state what will be the tradoff on colors.

I'm really happy to see this PR up!

andrii-i:
kicking CI

andrii-i:
Bot please update snapshots

andrii-i:
@krassowski changed text number color to use var(--jp-ui-font-color2) font color matching background color var(--jp-layout-color2). Line numbers are now much more readable in dark theme (1.91 --> 4.24) without increasing contrast in light theme (4.52 --> 4.44).

andrii-i:
Kicking CI

fcollonval (approved):
Thanks @andrii-i 

> I'll wait before merging for Mike to have a second look

andrii-i:
Thank you @fcollonval

andrii-i:
Bot please update snapshots

andrii-i:
Kicking CI

andrii-i:
Kicking CI again

packages/codemirror/src/theme.ts
```diff
@@ -54,6 +54,10 @@ export const jupyterEditorTheme = EditorView.theme({
     backgroundColor: 'var(--jp-layout-color2)'
   },
 
+  '.cm-gutter, .cm-lineNumbers': {
```
krassowski, line 57: Question: do we need `.cm-gutter` here? It looks that both classes are on the same element:

Screenshot from 2023-04-01 17-01-09

If that is not needed, then fewer CSS rules is better.

```suggestion
  '.cm-lineNumbers': {
```

krassowski (commented):
Thank you @andrii-i. I opened https://github.com/jupyterlab/jupyterlab/issues/14317 to track further improvement on increasing contrast.

PS. It could be useful to add a snapshot test using dark mode to avoid regressions (could be a follow-up PR).

krassowski:
Also, before we merge `galata/test/documentation/general.test.ts-snapshots/interface-right-common-documentation-linux.png` needs reverting.

packages/codemirror/src/theme.ts
andrii-i: `  '.cm-gutter, .cm-lineNumbers'` selector was already in place so I considered that we might be planning to show line numbers in some other place except gutter. If it's not the case, just ` '.cm-lineNumbers':` should work

andrii-i:
@krassowski thank you for the review! Reverted `interface-right-common-documentation-linux.png` and created #14331 to track adding a snapshot test using dark mode. By the way, could you recommend a good source to learn about adding galata snapshots?

andrii-i:
@krassowski it's all green now 🟢, please give it a look when possible

krassowski (approved):
Thank you!

andrii-i:
Thank you @krassowski

PR #14272, opened 2023-03-24
Update react-json-schema-form packages to 5.4.0

bollwyvl:
## References

- n/a

## Code changes

- updates to latest rjsf

## User-facing changes

- some bugs on deep errors are fixed
- file previews on upload fields

- ???

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -47,7 +47,7 @@ dependencies:
   - python-build
   - ruff
   # demo
-  - jupyter-server-proxy
+  # - jupyter-server-proxy
   - jupyterlab-link-share >=0.2
   - matplotlib-base
   - numpy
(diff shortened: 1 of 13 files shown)


fcollonval (commented):
@bollwyvl after the rebase this PR seems good to go as is. Is it okay for you to move forward?

bollwyvl:
Ha, yeah, forgot about this when binder didn't work. I'll take another look at it... there might be some hard fails with react 18, which is still not _fully_ supported by rjsf. Probably my least favorite issues (as a subset of the larger `nodejs` dependency situation). 

On the main, there's some better support for i10n, which _might_ remove some of the extra complexity we created while waiting for it to happen. It also includes some nice features like, "copy item" for array items, but given how bespoke the two implementations are, i don't know what the level of effort will be to achieve this.

This weekend is kinda booked, will see when i can get to it. If it's too deep in react-16vs17vs18-land, i might not have the patience.

fcollonval:
Closing as you went much further in https://github.com/jupyterlab/jupyterlab/issues/15302 :wink:

Issue #7619, opened 2019-12-11
OS-level *.ipynb open support


flutefreak7:
tldr: Notebooks should be double-click-able / executable the way other files are for users who run jupyterlab locally.

Now that JupyterLab supports opening notebooks based on a URL and supports having multiple workspaces managed with URL's it seems like it should be possible to make a file-handler at the OS level that converts a filepath to a URL and opens it in the locally running jupyter server instance.

I completely understand that Jupyter is designed with client-server separation as a core tenant, but the user experience as a desktop application is pretty frustrating. If a desktop file handler could be developed that would open a notebook in a the current local jupyterlab session (or possibly start a jupyterlab session if needed), that would make it much easier to use notebooks with the rest of the OS. Notebooks could be opened directly from a filebrowser, from an email, or from the command-line. Ideally, we should be able to interact with notebooks the same way we interact with other files, like an Excel Workbook for example.

vidartf:
Ref prior art: [nbopen](https://github.com/takluyver/nbopen)

vidartf:
Also relevant: https://github.com/jupyterlab/jupyterlab_app

vidartf:
If I considered it solved I would close the issue 😉. I was merely cross-referencing other work related to the issue, so that others who look at this issue might have a better context. 

If you know of a solution that works, please consider helping out by creating a PR with the proposed solution 😊

flutefreak7:
I don't have the bandwidth to steward the issue, but I'll ask the first question that occurs to me.

Are these potential solutions mature? Even if not, is there consensus in the community that these efforts are the primary effort to solve this concern? If so then, yeah this issue should be considered closed and point to those solutions either as the "solution" or at least the best path towards solution.

frenzymadness:
My plan is to open a PR proposing a .desktop file which makes it much easier to use lab without opening a terminal window first. It's an improvement but it probably does not solve all the issues.

PR #14277, opened 2023-03-27
Add desktop file and icon for it

frenzymadness:
You don't need to open a terminal window anymore.

I'm not able to build sdist/wheel locally so I'm going to test them from CI.

## References

Related: https://github.com/jupyterlab/jupyterlab/issues/7619

## Code changes

When these files are installed in proper locations, it is much easier to run JupyterLab on systems with GUI.

## User-facing changes

This is how it looks in Gnome.

image

None.

### DIFF
--- a/jupyterlab.desktop
+++ b/jupyterlab.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Name=JupyterLab
+Comment=Run JupyterLab
+Exec=jupyter-lab %f
+Terminal=true
+Type=Application
+Icon=jupyterlab
+StartupNotify=true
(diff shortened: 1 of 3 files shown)


jtpio:
Thanks @frenzymadness.

> so I'm going to test them from CI.

You should be able to find the assets as artifacts here: https://github.com/jupyterlab/jupyterlab/actions/runs/4531700263?pr=14277

frenzymadness:
Thank you. The new files are in sdist archive and wheel and they install in the correct locations so everything seems to be okay.

fcollonval (approved):
Thanks @frenzymadness

PR #14279, opened 2023-03-27
Update extension upgrade script

fcollonval:
## References

Update extension upgrade script for JupyterLab 4.0

## Code changes

- Get more cookiecutter input values
- Add Python deps and extras to pyproject.toml from setup.cfg
- Change default override file answer dependending on the file

## User-facing changes

User can use `python -m jupyterlab.upgrade_extension` script

N/A

### DIFF
--- a/jupyterlab/upgrade_extension.py
+++ b/jupyterlab/upgrade_extension.py
@@ -248,7 +248,6 @@ def update_extension(  # noqa
                     override_pyproject = True
             else:
                 warnings.append(f"skipped _temp_extension/{relpath!s}")
-    print(override_pyproject, (target / "setup.cfg").exists())
     if override_pyproject and (target / "setup.cfg").exists():
         config = configparser.ConfigParser()
         with (target / "setup.cfg").open() as setup_cfg_file:
--- a/jupyterlab/upgrade_extension.py
+++ b/jupyterlab/upgrade_extension.py
@@ -1,12 +1,18 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import configparser
 import json
-import os
-import os.path as osp
+import re
 import shutil
 import subprocess
 import sys
+
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
+import tomli_w
 
 try:
     from importlib.resources import files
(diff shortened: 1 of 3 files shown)


krassowski:
This PR adds a new jupyterlab dependency: [tomli-w](https://github.com/hukkin/tomli-w) package ([PyPI](https://pypi.org/project/tomli-w/)) which does not seem to be used elsewhere in Jupyter stack. Should we move it to `extra_requires` under a new `jupyterlab[dev]` to avoid installing it for every user?

> Context: the alternative would be [tomlkit](https://github.com/sdispater/tomlkit) ([PyPI](https://pypi.org/project/tomlkit/)) and maybe [pytomlpp](https://github.com/bobfang1992/pytomlpp) ([PyPI](https://pypi.org/project/pytomlpp/)). `tomli_w` is smaller than `tomlkit` (6kB vs 36kB wheel), `tomlkit` appears to be more popular and versatile (preserving style). All three are MIT licensed.

jupyterlab/upgrade_extension.py
```diff
@@ -40,40 +77,68 @@ def update_extension(target, branch=DEFAULT_COOKIECUTTER_BRANCH, interactive=Tru
...
+    output_dir = target / "_temp_extension"
+    if output_dir.exists():
         shutil.rmtree(output_dir)
 
     # Build up the cookiecutter args and run the cookiecutter
+    author = data.get("author", "<author_name>")
+    author_email = "<author_email>"
+    if isinstance(author, dict):
+        author = author.get("name", "<author_name>")
+        author_email = author.get("email", author_email)
```
krassowski, line 118:
```suggestion
        author_email = author.get("email", author_email)
        author = author.get("name", "<author_name>")
```

Or even better, switch to explicit `author_name`. Currently we get:

```python
AttributeError: 'str' object has no attribute 'get'
```

(as of now `author` gets overridden with a string).

fcollonval: Good catch

fcollonval:

peytondmurray:
Just tested this out on https://github.com/deshaw/jupyterlab-pyflyby. I'm not sure how relevant these are (whether I selected the right options), but here are some things I saw:

* `.github/workflows/build.yml` - the `Install Dependencies` section is installing `jupyterlab~=3.1`; should this be 4.x? This also happens in the `test_isolated` job.
* `.gitignore` and `.eslintignore` includes testing-related ignores even if I answer "no" to the testing prompt.
* An existing `schema/plugin.json` gets overwritten with a blank template if you answer "yes" to settings.
* Looks like `setup.py` is kept, though I think it can be deleted altogether. `setuptools` is not a dependency of the project, so the import should fail.

Aside from these minor things, this looks really good so far - thanks for this, it is already making the transition easier!

fcollonval:
Thanks for testing @peytondmurray 

>     * `.github/workflows/build.yml` - the `Install Dependencies` section is installing `jupyterlab~=3.1`; should this be 4.x? This also happens in the `test_isolated` job.

Could you check you use this branch (or did you check about week ago)? It looks like the upgrade script is running with the 3.0 branch.
 
>  `.gitignore` and `.eslintignore` includes testing-related ignores even if I answer "no" to the testing prompt.

Good catch I'll push an improvement to the cookiecutter

>  An existing `schema/plugin.json` gets overwritten with a blank template if you answer "yes" to `overwrite "schema/plugin.json"?`. I thought this would update the settings to a new format, but I must be misunderstanding this prompt.

This is expected in the current migration - FYI there has not been any changes to the settings format.

>  Looks like `setup.py` is kept, though I think it can be deleted altogether. `setuptools` is not a dependency of the project, so the import should fail.

It is kept for backward compatibility as some tools are still relying on having `setup.py`. But you can definitely drop it.

>  Inside `handlers.py`, all the `IPythonHandler` class definitions were removed.

You should not mix up handler from `notebook` v6 and prior with handler from jupyter server. So the correct code is to drop all import from `notebook` and use `JupyterHandler` or better `APIHandler`.

Looking at the extension pyflyby, I also strongly recommend protected the endpoint with the decorator `@tornado.web.authenticated`

PR #14287, opened 2023-03-28
Add missing lumino package as singleton

fcollonval:
## References

Seen when updating the versions

## Code changes

None

## User-facing changes

None

None


jtpio:
Isn't there a script (maybe `ensure_repo`) that is supposed to keep the list of `resolutions` in sync automatically?

fcollonval:
The script only checks that core packages with `token.ts` file are in the list:

https://github.com/jupyterlab/jupyterlab/blob/ed0e0eb9251546cce0acef580c7b6349ee9bbad8/buildutils/src/ensure-repo.ts#L473

Checking for third-party dependencies is gonna be tricky as we don't want all of them. We could maybe add an exception for lumino packages at:

https://github.com/jupyterlab/jupyterlab/blob/ed0e0eb9251546cce0acef580c7b6349ee9bbad8/buildutils/src/ensure-repo.ts#L496

jtpio:
Right normally it should be fine to share all `@lumino` packages by default.

PR #14289, opened 2023-03-29
[apputils-extensions] increase splash recover timeout to 20 secs

### LINKED ISSUE #14288: Splash Recovery Modal should be delayed to about 20 seconds
### Proposed Solution

change SPLASH_RECOVER_TIMEOUT from 12000 -> 20000 ms
../jupyterlab/packages/apputils-extension/src/index.ts
 -->

### Additional context
N/A
-->

cdrosser-amz:
## References

Fixes #14288

## Code changes

Changing the delay from 12s to 20s should allow more applications to have a longer initial load time before prompting users to 'Clear Workpace'


Christopherallen:
+1

fcollonval:
Hey @cdrosser-amz I'm not really a big fan of increasing that already large default timeout.

Would you be open to implement an alternative approach allowing customization through settings? The best would be through a server config. So it gets written to the index.html and avoid waiting for a server request (compare to user settings).

Let me know and I can provide pointers to achieve that.

cdrosser-amz:
Hi @fcollonval !
 Thanks for the feedback. That sounds like a great alternative. I'm not entirely familiar with the JL repo so some pointers on this would be greatly appreciated

fcollonval:
@cdrosser-amz 

Ok so the steps would be:

- Add a server settings like that (but using a [Float](https://traitlets.readthedocs.io/en/stable/trait_types.html#numbers) traits):

https://github.com/jupyterlab/jupyterlab/blob/b279092d88de650ea36460689257e1b8e8a418bf/jupyterlab/labapp.py#L576

- Pass the configuration parameter to the `page_config` similarly to `quit_button`:

https://github.com/jupyterlab/jupyterlab/blob/b279092d88de650ea36460689257e1b8e8a418bf/jupyterlab/labapp.py#L701

- Get the parameter in the frontend through `PageConfig.getOption`:

https://github.com/jupyterlab/jupyterlab/blob/b279092d88de650ea36460689257e1b8e8a418bf/packages/mainmenu-extension/src/index.ts#L169

> Additional notes, `PageConfig.getOption` only returns string. So you will need to parse it to get a number. You could also use a integer type for the parameter (as you fill is best for this case).

JasonWeill:
@cdrosser-amz Are you planning to revise this PR to incorporate the feedback above?

JasonWeill:
Retitled #14288 to indicate that this value should be made configurable, rather than changed as a constant.

Issue #14221, opened 2023-03-18
Latest examples do not render outputs


echarles:
I am running the notebook example and after https://github.com/jupyterlab/jupyterlab/commit/2142d76f917032d332645d9fb0cde3660432d522#diff-81f7333ccbe5cab9f436c43df103c8158eea12521445f33a95e58a06940a1e13 the outputs are not rendered anymore  although the  kernel message are sent 

cc @krassowski 

![Untitled6](https://user-images.githubusercontent.com/226720/226104572-68412b64-9326-4b60-80a1-1bb8a2ccc712.gif)

echarles:
cc/ @fcollonval

jtpio:
Also there is no syntax highlighting. Not sure if that was the case before already though, or also related to https://github.com/jupyterlab/jupyterlab/pull/13639.

echarles:
Thx for looking at this @jtpio and confirming the issue. I have tried to roll back a few commits behind, and even a few releases alpha behind. I had issues to make it work and tried to pin the alpha (npm/yarn resolution is not very friendly with alpha handling...), but was not able to nail down the commit that make the example fail.

Having more tests for the examples will help for sure.

akloster:
I ran into this issue as well, and it is a little annoying. I tried rolling back some versions of jupyter, but so far haven't gotten the notebook example to work.

echarles:
> I ran into this issue as well, and it is a little annoying. I tried rolling back some versions of jupyter, but so far haven't gotten the notebook example to work.

Thx @akloster for your inputs. It is now taken in the `Milestone 4.0.0` so hopefully will be solved before the final 4.0.0. The Beta release track has just started https://github.com/jupyterlab/jupyterlab/issues/9647#issuecomment-1484726405.

I have debugged more with the beta version trying to identify the source of the issue without success. It looks to me no developer is looking at this, so pinging @jupyterlab/jupyterlab-council to raise a bit of attention and get feedback on the next steps.

jtpio:
Generally speaking we should address issues added to the `4.0` milestone before the final release.

And move them to `4.0.x` / `4.1` if they are not critical for the release.

echarles:
Example of [7227449db8db478dc5e31bbc1ffdfb741766a35d ](https://github.com/jupyterlab/jupyterlab/tree/7227449db8db478dc5e31bbc1ffdfb741766a35d) on Jan 25 2023 are working fine.

echarles:
[...]

PR #14293, opened 2023-03-30
Fix examples further to CodeMirror API change

echarles:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14221

## Code changes

Update the cell and notebook examples

## User-facing changes

The outputs are now displayed as before for the `cell`, `notebook` and `console` examples..

Untitled6

None

### DIFF
--- a/packages/codemirror-extension/src/services.tsx
+++ b/packages/codemirror-extension/src/services.tsx
@@ -200,7 +200,7 @@ export const extensionPlugin: JupyterFrontEndPlugin<IEditorExtensionRegistry> =
   };
 
 /**
- * CodeMirror binding provider.
+ * CodeMirror shared model binding provider.
  */
 export const bindingPlugin: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlab/codemirror-extension:binding',
--- a/packages/codemirror-extension/src/services.tsx
+++ b/packages/codemirror-extension/src/services.tsx
@@ -200,7 +200,7 @@ export const extensionPlugin: JupyterFrontEndPlugin<IEditorExtensionRegistry> =
   };
 
 /**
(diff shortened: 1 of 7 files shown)


jtpio:
Thanks @echarles for looking into it :+1: 

Looks like it needs a lint check, but looking good otherwise.

echarles:
Thx @jtpio. Just linted and pushed.

packages/codemirror-extension/src/services.tsx
```diff
@@ -200,15 +200,15 @@ export const extensionPlugin: JupyterFrontEndPlugin<IEditorExtensionRegistry> =
   };
 
 /**
- * CodeMirror real-time collaboration binding provider.
+ * CodeMirror binding provider.
```
fcollonval, line 203: Could we keep a reference to what it is bound to CodeMirror?

```suggestion
 * CodeMirror shared model binding provider.
```

fcollonval (changes requested):
Thanks @echarles 

I have one small comments and the releaser test needs to be fixed - the explicit dependency to `@jupyter/ydoc` should be provided in the examples package.json.

> the other test failures are due to flaky errors or are not the consequence of this particular PR.

echarles:
> the explicit dependency to @jupyter/ydoc should be provided in the examples package.json.

thx for the review @fcollonval - I have added the needed deps.

fcollonval:
Almost there @echarles, you need to run `jlpm install` to update the yarn.lock file and this PR will be good to go.

> As you are submitting from an org, I unfortunately cannot help without those changes.

echarles:
@fcollonval Thx for your help and patience :)

fcollonval (approved):
Thanks @echarles 

> CI failures are not related

Issue #14299, opened 2023-03-30
Empty output area is not cleared


3coins:
## Description
After resizing the output area, clearing the output cell doesn't make it go away, and shows an empty box.
![jlab-4-beta-output-cell-bug](https://user-images.githubusercontent.com/289369/228935322-04d4b28f-11bf-42a9-a7d2-105a4659c8d6.gif)

## Reproduce
1. Execute a cell with large output. For example:
```python
for i in range(0, 50):
    print('Hello World!')
```
2. Collapse the output cell by clicking in the output row button on the left.
3. Resize the output cell by moving the resizer at the right bottom.
4. Clear the cell output
5. Should see an empty output cell.

## Expected behavior
The output area should be completely removed.

## Context

- Operating System and version: Mac OS Monterey 12.6.3
- Browser and version: Chrome 111.0.5563.64 & Safari Version 16.3 (17614.4.6.11.6, 17614)
- JupyterLab version: 4.0.0b0

krassowski:
Good catch.I believe this is not specific to 4.0 and being a subjectively minor issue should not block us for 4.0.0. As for expectation, the styling (borders and inset shadow) indicates state which is saved in ipynb file and users might prefer it to stay even after clearing output, but height is transient and should be cleared indeed (well, until we decide to start storing it).

Implementation wise this could be a trivial one line fix by addition of `height` property removal in clear outputs code but that would not work if "clear outputs" was invoked by another user in RTC. Maybe that's good enough - if not we would need to listen to outputs changes and clear `height` on the scroll container each time the number of outputs drops to zero.

krassowski:
This can be fixed in a similar way as https://github.com/jupyterlab/jupyterlab/pull/16124

PR #14304, opened 2023-03-31
Update actions.spec.ts

kamalika0363:


fcollonval:
@doesnots would you mind providing some background for this PR?

fcollonval:
Thanks for reaching out

The file you are modifying is a test. So it will not have any impact on the software itself. Would you mind pointing or opening an issue with the problem you are facing (adding some small videos or screenshots would help a lot).

doesnots:
@fcollonval thanks for pointing that out I guess I mishandled the files. I was trying to solve this issue https://github.com/jupyterlab/jupyterlab/issues/14299#issue-1648146481 
I'll try to get the solution. _

fcollonval:
@doesnots I would suggest you ask for pointers on the issue.

krassowski:
@doesnots what do you think about switching this PR to draft while you look further into a solution?

krassowski:
It has been a year since last comment - I will close this PR, but please reopen if you get to finishing it.

Issue #11012, opened 2021-09-04
Default variable filters no longer work for xeus-python


krassowski:
## Description

The default variable filters do not seem to work anymore in the debugger for xeus on 4.0 / master. It works well on JupyterLab 3.1 demo binder.

## Reproduce

1. Switch to xeus-python kernel
2. Activate the debugger
3. Enter a breakpoint
4. See that `debugpy` shows up on the list of variables

![Screenshot from 2021-09-04 19-12-24](https://user-images.githubusercontent.com/5832902/132104353-e79e0c59-7541-47f0-bc37-963225471684.png)

## Expected behavior

Should work as it used to in 3.1.

## Context

- JupyterLab version: 4.0.a7

jtpio:
Thanks @krassowski for reporting.

For reference the filter is still defined here: 

https://github.com/jupyterlab/jupyterlab/blob/8c693aec308a57cf716c9a876a0a056b429c45b2/packages/debugger-extension/schema/main.json#L93-L115

But maybe it's not set correctly anymore.

Mithil467:
I tried reproducing this issue but it seems to be working fine for me.
![image](https://user-images.githubusercontent.com/41439841/132909978-33f4136a-7213-4dda-946a-18a245284ccc.png)

I am on wsl.
Installed `xeus-python-0.12.5` from pypi.
Dev mode logs says `Run with XEUS 1.0.3`.
Running latest commit https://github.com/jupyterlab/jupyterlab/commit/b89064d77803c4b10fe2fc44d885507d2aa3051a on master.

krassowski:
I cannot reproduce it either using current master/binder - interesting. I will take another look later on.

krassowski:
I was able to reproduce it in the same session after switching tabs a few times between two debugger sessions (one with ipykernel, and one with xeus-python). Clearly my reproduction instructions were incomplete. I will try to narrow it down further later.

PR #14321, opened 2023-04-03
Fix failure when setting debugger variables filter

afshin:
This is a follow-up PR after:
- https://github.com/jupyterlab/jupyterlab/pull/14037

This PR fixes a run-time failure when trying to set variable filters for `xeus-python`:

https://user-images.githubusercontent.com/159529/229621418-8824ea52-0892-41b7-844a-6527c1b8cb0b.mov

---
Additionally, this PR **possibly** fixes https://github.com/jupyterlab/jupyterlab/issues/11012

---
cc: @bollwyvl @krassowski 

## References
- https://github.com/jupyterlab/jupyterlab/issues/11012
- https://github.com/jupyterlab/jupyterlab/pull/14037

## Code changes
Refactors the `VariablesBodyGrid` class to accept `filter`, `scope`, and `model` changes whether or not the grid is visible.

## User-facing changes
Setting a filter no longer generates a console error.

N/A

### DIFF
--- a/packages/debugger/src/panels/variables/grid.ts
+++ b/packages/debugger/src/panels/variables/grid.ts
@@ -3,18 +3,19 @@
 
 import { IThemeManager } from '@jupyterlab/apputils';
 
+import { ITranslator } from '@jupyterlab/translation';
+
 import { CommandRegistry } from '@lumino/commands';
+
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @afshin

PR #14325, opened 2023-04-04
chore: move configurations under `package.json`

SauravMaheshkar:
## References

Similar to https://github.com/jupyter-server/jupyter-resource-usage/pull/188 and https://github.com/jupyter-server/jupyter_server/pull/1246

## Code changes

Changes proposed by this PR can be summarised as follows :-

* Move the content from `.prettierrc` under the `prettier` key in `package.json`. [Source(prettier docs)](https://prettier.io/docs/en/configuration.html)
* Move the content from `.eslintrc.js` under the `eslintConfig` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files)
* Move the content from `.eslintignore` under the `eslintIgnore` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/ignore#using-eslintignore-in-packagejson)
* Move the content from `.stylelintrc` under the `stylelint` key in `package.json`. [Source(stylelint docs)](https://stylelint.io/user-guide/configure/)

Doing so, would lead to a more minimal project structure without any functional change.


krassowski:
What are the implications for packaging? Would this increase package size for npm since package.json has to be included? Also, manually editing such a large package.json may be less pleasant of an experience.
I would propose to wait with merging an bring this up during the team meeting on Wednesday.

SauravMaheshkar:
Gentle ping

PR #14328, opened 2023-04-05
Pass traceback to ServerConnection.ResponseError

a3626a:
In my Jupyter Lab environment, `Error: Unhandled Error` is reported by monitoring tool. I'm investigating each errors and making some handling for them. However `Error: Unhandled Error` does not have enough infomation to be handled properly. On jupyter lab side, I found that `ServerConnection.ResponseError` has a room for server side `traceback`, but it is not well managed. If `trackback` works properly, it can give us enough infomation for errors to be handled.

(It is reported as `Error`, but actually `ServerConnection.ResponseError`)

## References
- https://github.com/jupyter-server/jupyter_server/issues/1247

## Code changes
Pass `trackback` to ServerConnection.ResponseError. I think it is intended behavior of ResponseError at first, but looks like missing somehow.

I wrote the code in a conservative manner to preserve previous behavior as much I can. That is why `data['message'] ? data['message'] : undefined` comes.

## User-facing changes
No

Actually, I found no code that uses `traceback` of the ResponseError constructor. It is backword compatible.

### DIFF
--- a/packages/services/src/serverconnection.ts
+++ b/packages/services/src/serverconnection.ts
@@ -143,13 +143,14 @@ export namespace ServerConnection {
     static async create(response: Response): Promise<ResponseError> {
       try {
         const data = await response.json();
-        if (data['traceback']) {
-          console.error(data['traceback']);
+        const { message, traceback } = data;
+        if (traceback) {
+          console.error(traceback);
         }
         return new ResponseError(
           response,
-          data['message'] ? data['message'] : undefined,
-          data['traceback'] ? data['traceback'] : undefined
+          message ?? ResponseError._defaultMessage(response),
+          traceback ?? ''
         );
       } catch (e) {
         console.debug(e);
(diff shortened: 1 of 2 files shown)


fcollonval (commented):
Thanks @a3626a I took the liberty to push an improvement to preserve default values.

a3626a:
> Thanks @a3626a I took the liberty to push an improvement to preserve default values.

My previous implementation preserves default values. `undefined` function arguments are overriden by default values. I also double checked that this implentation passes your new test.

```
    static async create(response: Response): Promise<ResponseError> {
      try {
        const data = await response.json();
        const { message, traceback } = data;
        if (traceback) {
          console.error(traceback);
        }
        return new ResponseError(
          response,
          message ? message : undefined,
          traceback ? traceback : undefined
        );
      } catch (e) {
        console.debug(e);
        return new ResponseError(response);
      }
    }
```

However I don't care about details of implmentation. I'm 100% satisfied if the traceback is passed anyway. Thank you for your feedback.

fcollonval:
Thanks @a3626a I learnt something - I was not aware that explicitly passed `undefined` was still replaced by the default value.

fcollonval:

PR #14218, opened 2023-03-17
Update more dev dependencies

jtpio:
## References

Follow-up to #14216 

## Code changes

Update more dev dependencies, followed by a lint run:

```shell
jlpm
jlpm run lint
```

- Bump to TypeScript 5.0.3
- Bump to typedoc 0.23.28
- Remove [stylelint-config-prettier](https://github.com/prettier/stylelint-config-prettier#readme) as those rules are deprecated since stylelint v15

## User-facing changes

None

## Backwards-incompatible changes

None

PR #14339, opened 2023-04-07
Upgrade dependencies

fcollonval:
## References

Other attempt to upgrade dependencies vs https://github.com/jupyterlab/jupyterlab/pull/14218
The goal is to see if we can avoid the memory overflow of the older PR

## Code changes

Upgrade typescript to 5.0.4
Upgrade ts-jest to 29.1.0 (compatible with ts 5)
Upgrade yarn 3 to 3.5.0

## User-facing changes

None

None


fcollonval:
CI failure is expected as the updated patch for yarn in staging is required

PR #14346, opened 2023-04-08
Fix "additional properties" settings editor validation error

### LINKED ISSUE #13651: Unable to change some settings after searching
## Description

Settings fail to save as JupyterLab Shell schema defaults does not meet the schema (which is wrong).

## Reproduce

1. Go to 'Settings'
2. Search "panel"
3. Click on 'JupyterLab Shell'

Screenshot from 2022-12-23 00-13-53

## Expected behavior

Settings can be set

## Context
JupyterLab 4.0.0a30

marthacryan:
## References
Fixes #13651. 

## Code changes
Adds a filter for form data in the settings editor to remove fields that aren't defined in the (filtered) schema. 
The validation error was happening because:
* The schema had filtered out some of the properties so that only fields that are relevant to the search show up
* The form data being used was including all values, so the validation was complaining that additional properties weren't allowed. 

None


krassowski (commented):
Thank you for working on the fix! I tested it on binder and it works well. Preferably we would avoid the `any` typing and add a test of some kind.

marthacryan:
@krassowski I agree about avoiding the any type - thoughts on what type it should be? The `formData` field is currently defined as any. Maybe `PartialJSONObject`? If I change this I can also change the `formData` type in the other instances its used. 

Also, we do have a test for "open the settings editor with a specific search query" - do you think there should be a separate test for this? I think it just didn't catch it because this bug didn't affect all of the settings.

krassowski:
`IFormComponentProps` by default takes a `ReadonlyJSONObject`. Maybe it does not have to be read-only, but I think that this is a good practice to have it as immutable. So maybe:

```ts
private _formData: ReadonlyPartialJSONObject | undefined;
```

and then

```ts
private _getFilteredFormData(filteredSchema?: ISettingRegistry.ISchema): ReadonlyJSONObject {
    if (!filteredSchema?.properties) {
      return this._formData as ReadonlyJSONObject;
    }
    const filteredFormData = JSONExt.deepCopy(this._formData as PartialJSONObject);
    for (const field in filteredFormData) {
      if (!filteredSchema.properties[field]) {
        delete filteredFormData[field];
      }
    }
    return filteredFormData as ReadonlyJSONObject;
  }
```


It looks like `this.props.settings.isDefault` should also have a guard against `undefined`.

> Also, we do have a test for "open the settings editor with a specific search query" - do you think there should be a separate test for this? I think it just didn't catch it because this bug didn't affect all of the settings.

In my opinion the best action would be to seed a `test` folder for jest unit tests (as done in all other core packages), testing this very specific case on dummy schema rather than adding an integration test (and hoping that the future schema changes in other plugins will not invalidate this test case). I think that `packages/metadataform/test/metadataFormWidget.spec.ts` could be some inspiration as it also tests a widget based on the `FormComponent` (but here we probably don't need to have notebook nor spawn the server).

andrii-i:
@marthacryan This is 1 of 2 blocking issues for 4.0.0 release, please let me know if you need help, I can find time to help tomorrow on Tuesday Apr 25 PDT

andrii-i:
Context: https://github.com/jupyterlab/jupyterlab/issues/9647#issuecomment-1520925685
> Do we still plan to release 4.0.0 by JupyterCon (May 10-12)? If we do and we need 2 weeks of RC, RC needs to be released by Wed Apr 26, the day after tomorrow. Currently there are 2 [blocking issues]

fcollonval:
@marthacryan I'm working on adding a test to finish this PR.

andrii-i:

marthacryan:
@fcollonval Thanks for doing that! Sorry for the delay on that test.

galata/test/galata/notebook.spec.ts-snapshots/run-cells-galata-linux.png
andrii-i, line 1: Flakey snapshot, note "Click to add cell", needs reverting

galata/test/jupyterlab/notebook-create.test.ts-snapshots/run-cells-jupyterlab-linux.png
andrii-i, line 1: Flakey snapshot, note "Click to add cell", needs reverting

galata/test/jupyterlab/completer.test.ts-snapshots/completer-with-doc-panel-jupyterlab-linux.png
andrii-i, line 1: This also looks like a flakey snapshot. Have not seen this before. CC @krassowski

andrii-i (approved):
This restores search functionality for `pluginlist`, all suggestions are accounted for. Note that search does not appear in `settingspanel` until something in `pluginlist` is clicked but we can tackle this in a separate PR

andrii-i:
Benchmark Tests are failing with `Warning: "cml publish" is deprecated since "cml comment" now supports "inline"`, UI regression test is failing with a snapshot not related to this PR as discussed during todays JupyterLab call

PR #14353, opened 2023-04-10
Ensure that initial text is updated on subsequent searches

### LINKED ISSUE #14306: Search should always update query to the currently selected text (if any)
> Just no sure if this still work in progress or even planed, but when I tested this version it does not behave like many other programs when calling the search with `ctrl+F`. What I mean by that, is, when you have selected a different word in the page than the one currently in the search bar, pressing`ctrl+F` should update the search query. 

_Originally posted by @Alexboiboi in https://github.com/jupyterlab/jupyterlab/issues/14090#issuecomment-1474075981_

Reproducible in JupyterLab 4.0.0-beta.0

krassowski:
## References

Fixes #14306

## Code changes

- split up `initialQuery` in search model that had a circular logic preventing proper functioning into:
  - `initialQuery` getter/setter pair for read/manipulation from widget
  - `suggestedInitialQuery` to expose the initial query suggested by search provider (usually the selected text)
- use `key` trick to enforce update of textarea's `defaultValue` when `initialQuery` changes
  - use effect hook to ensure the text selection in textarea gets updated afterwards
- rename `searchText` widget props to `initialSearchText` to clarify the purpose

## User-facing changes

Invoking search (e.g. with <kbd>Ctrl</kbd> + <kbd>F</kbd>) will update the query to text selected in editor:

Initial-search-input

**Note**: some editors do not update the text if current selection spans multiple lines. This PR does not introduce such a restriction (multi-line selections also update the search text). If any of reviewers believe we should adapt such a restrictions (I don't see a solid argument for it, but happy to be proven wrong) let's open a new issue to discuss.

As previously, previous search text is restored if there is no selection; this behaviour is now covered by integration tests:

previous-search-text-is-restored

### DIFF
--- a/galata/test/jupyterlab/notebook-search.test.ts
+++ b/galata/test/jupyterlab/notebook-search.test.ts
@@ -7,6 +7,13 @@ import * as path from 'path';
 
 const fileName = 'search.ipynb';
 
+function getSelectionRange(textarea: HTMLTextAreaElement) {
+  return {
+    start: textarea.selectionStart,
+    end: textarea.selectionEnd
+  };
+}
+
 test.describe('Notebook Search', () => {
   test.beforeEach(async ({ page, tmpPath }) => {
     await page.contents.uploadFile(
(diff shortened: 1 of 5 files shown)


fcollonval (approved):
Thanks a lot @krassowski

Alexboiboi:
Awsome, thanks a lot @krassowski ;)

PR #14360, opened 2023-04-11
Document plugins and tokens

fcollonval:
## Code changes

Add plugins and tokens description
Upgrade @lumino to 2.1.0 (to support token description)

## User-facing changes

None

Fix a token name


fcollonval:
> CI failure is expected for the labeler and otherwise not related

Issue #14366, opened 2023-04-12
JupyterLab `4.0.0b1` breaks Notebook 7 `TopBar` items


jtpio:
## Description

In https://github.com/jupyter/notebook/pull/6838 we noticed that updating from `4.0.0b0` to `4.0.0b1` breaks some of the `TopBar` toolbar items defined via the settings system (`jupyter.lab.toolbars`):

![image](https://user-images.githubusercontent.com/591645/230625673-7a22f5fe-def0-40f9-886a-dbddc7a15fcf.png)

## Reproduce

See discussion in https://github.com/jupyter/notebook/pull/6838#issuecomment-1500342416.

## Context

As mentioned in https://github.com/jupyter/notebook/pull/6838, it seems related to a settings PR that landed in `4.0.0b1`.

Quoting @brichet:

> Actually, the problem comes from another part of https://github.com/jupyterlab/jupyterlab/pull/14195:

> https://github.com/jupyterlab/jupyterlab/blob/6d9217cc9681879dd822e6aaa5cedad2a648f99c/packages/settingregistry/src/settingregistry.ts#L394-L403

> Everything works fine if I avoid this part for `@jupyterlab/application-extension:top-bar` plugin. And I don't know why it works if the `IDocumentManager` is not a dependency.

cc @fcollonval @brichet since you worked on that PR.

jtpio:
Opening a new issue here in JupyterLab so it can get more attention, since it is currently blocking further updates of Notebook 7.

brichet:
Before the use of cache for settings, the settings were fetched again form server. This fetch is delayed by 100 ms using a `Throttler`:
https://github.com/jupyterlab/jupyterlab/blob/25658adabce206725cafe42161181ba43ae87bcd/packages/apputils-extension/src/settingconnector.ts#L36

Using the cache, the delay does not exist anymore, and the widget factory has not been set yet.
If we artificially add a sleep of 100 ms before the load from cache, everything works fine.
We were lucky it used to work without trouble before, it was just a matter of scheduling.

brichet:
A workaround for Notebook while this is not fixed is to force playing again the transform function after the `widgetFactory` is added. The following fixes the issues for the `4.0.0b1` upgrade, but must be fixed in JupyterLab:

```ts
if (toolbarRegistry) {
  toolbarRegistry.addFactory('TopBar', 'checkpoint', (toolbar) => {
    const widget = new Widget({ node });
[...]

PR #14376, opened 2023-04-13
Add signal when an item factory is added to the toolbar

brichet:
See https://github.com/jupyterlab/jupyterlab/issues/14366 for context.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14366
The issue was first notified in https://github.com/jupyter/notebook/pull/6838

## Code changes

Adds a signal in `ToolbarWidgetRegistry`, which is emitted when an item factory is added to the registry.
This signal is caught by the `toolbarFactory` to render again the widget.

## User-facing changes

None

None

EDIT:
Tested successfully on https://github.com/jupyter/notebook/pull/6838

### DIFF
--- a/packages/apputils/src/toolbar/factory.ts
+++ b/packages/apputils/src/toolbar/factory.ts
@@ -303,6 +303,25 @@ export function createToolbarFactory(
       }
     };
 
+    const updateWidget = (
+      registry: IToolbarWidgetRegistry,
+      itemName: string
+    ) => {
+      const itemIndex = Array.from(items).findIndex(
+        item => item.name === itemName
+      );
+      if (itemIndex >= -1) {
+        toolbar.set(itemIndex, {
+          name: itemName,
+          widget: toolbarRegistry.createWidget(
+            factoryName,
+            widget,
+            items.get(itemIndex)
+          )
+        });
+      }
+    };
+
     const toolbar = new ObservableList<ToolbarRegistry.IToolbarItem>({
       values: Array.from(items).map(item => {
         return {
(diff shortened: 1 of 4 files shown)


packages/apputils/src/toolbar/factory.ts
```diff
@@ -303,6 +303,25 @@ export function createToolbarFactory(
...
     };
 
+    const updateWidget = (
+      registry: IToolbarWidgetRegistry,
+      itemName: string
+    ) => {
+      const itemIndex = Array.from(items).findIndex(
+        item => item.name === itemName
+      );
+      if (itemIndex >= -1) {
```
fcollonval, line 313: I guess it is:

```suggestion
      if (itemIndex >= 0) {
```

fcollonval (changes requested):
Thanks @brichet I think there is a small typo but otherwise it looks good.

packages/apputils/src/toolbar/factory.ts
brichet: Thanks

jtpio:
Thanks @brichet!

Wondering if there could be other situations than just toolbar items that could be affected by this new caching logic of the settings?

brichet:
> Wondering if there could be other situations than just toolbar items that could be affected by this new caching logic of the settings?

This is hard to say, this is really an indirect effect, and there is probably no test on that behavior.
Actually, we were lucky it worked without trouble until now. It is likely that adding dependencies on other plugins with long loading times could have produced the same issue.

Issue #14308, opened 2023-04-01
Search "in selected" sometimes reverts to edit mode preventing cell selection


krassowski:
## Description

When searching "in selection" (https://github.com/jupyterlab/jupyterlab/pull/14090), switching to `command` mode and attempting to select multiple cells the mode switches back to  `edit` mode putting cursor back in the cell and preventing selecting more cells.

This appears somewhat similar to seemingly unrelated issue https://github.com/jupyterlab/jupyterlab/issues/14271 (could be unrelated or could have to do with recent cursors work)

## Reproduce

![switches-to-edit-mode](https://user-images.githubusercontent.com/5832902/229264108-b7c87af5-112a-4d37-a6fb-3f4d8ffcaa97.gif)

## Expected behavior

When <kbd>Shift</kbd> + <kbd>Up</kbd>/<kbd>Shift</kbd> + <kbd>Down</kbd> is pressed the cell selection extends and the command mode remains unchanged.

## Context

- JupyterLab version: 4.0.0b

krassowski:
This looks hard, and originate from CodeMirror focusing the editor back when decorations get updated. Question upstream: https://discuss.codemirror.net/t/how-to-add-decoration-without-focusing-the-editor-on-dispatch/6283

It is still not clear to me why CodeMirror only focuses editor back on some but not all updates to decorations.

fcollonval:
I tagged this as release blocker as the associated PR changes the API; just to be sure we merge it before the final release.

PR #14387, opened 2023-04-16
Improve focus behaviour in search

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14308

Depends on https://github.com/codemirror/view/commit/f49f2eea1ae1b81c8e943b76a57ca787bfd46bf5 included in [@codemirror/view](https://github.com/codemirror/view) 6.9.5, therefore required https://github.com/jupyterlab/jupyterlab/pull/14413

## Code changes

- only create DOM event (i.e. focus) CM extension once, and only add it during editor setup phase
- support arbitrary nesting of highlight markers fixing current highlight not shown in some cases
- preserve current match when switching cells if possible 
- if there is no current match, try to assign one
- do not ever set selection when searching in selection, instead just scroll to the match
- introduces (non-public) `SearchStartAnchor` type to disambiguate the search start behaviour allowing to fix highlighting issues
- introduces `IHighlightAdjacentMatchOptions` interface to specify whether highlighting should scroll/select new match
- do not scroll/select new match when starting search
   - this can be improved in follow-up: we could scroll/select if this is a new query (but not restart of query after document change)

## User-facing changes

Extending selection after changing to command mode now works smoothly (does not revert to edit mode, which was caused by CodeMirror bug, and is faster - which was previously slowed down by needless cleaning of highlights due to both `activeCellChanged` and `selectionCellChanged` handlers being used - now the shared job will only be performed once).

Initial match is set consistently across various scenarios, e.g.:

| Before | After |
|---|--|
| image | image |

Including, in file editor, and when populating search box from selection (regardless of selection direction):

image

N/A


packages/codemirror/style/base.css
```diff
@@ -59,24 +59,27 @@
...
 .cm-searching,
 .cm-searching span {
+  /* `.cm-searching span`: we need to override syntax highlighting */
   background-color: var(--jp-search-unselected-match-background-color);
   color: var(--jp-search-unselected-match-color);
 }
 
-.cm-content .cm-searching::selection,
-.cm-content .cm-searching span::selection {
+.cm-searching::selection {
```
krassowski, line 67:
```suggestion
.cm-searching::selection,
.cm-searching span::selection {
```

krassowski:
Merged master to clean update-links job failure that was addressed earlier.

fcollonval (commented):
Thanks a lot @krassowski for working on this.

fcollonval (approved):

krassowski:
@andrii-i can we merge this one or did you have any specific feedback here?

fcollonval:
Merging now to get it in the RC

andrii-i:
@krassowski No feedback on my end, it was mentioned as needing a review last week but I have not came around to reviewing. Thanks for reviewing @fcollonval.

Issue #14385, opened 2023-04-15
"Click to add a cell" is flaky on CI


krassowski:
> `jupyterlab/galata/test/jupyterlab/notebook-create.test.ts-snapshots/run-cells-jupyterlab-linux.png` needs to be fixed as the expectation is incorrect/"click to add a cell is" problematic here:

![run-cells-expected](https://user-images.githubusercontent.com/5832902/232229420-243b2987-f964-4de7-80f3-22ffe3b45848.png)

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/14344#issuecomment-1509848101_

krassowski:
There are actually two similar test cases one in galata project and one in jupyterlab, both are flaky:

![Screenshot from 2023-04-23 20-44-14](https://user-images.githubusercontent.com/5832902/233861930-a8c4c324-caba-42f6-99f5-4e4bd7eef591.png)

The other one is `galata/test/galata/notebook.spec.ts-snapshots/run-cells-galata-linux.png`

This is leading to many failures and is high burden, sometimes we need to restart the CI 4 times delaying a merge by ~4 hours.

JasonWeill:
Possible mitigation for this: apply masks to the screen shot to avoid failures on just the "click to add a cell" UI component

krassowski:
This is due to a difference in `.jp-WindowedPanel-inner` height.

Good snapshot:

![good](https://user-images.githubusercontent.com/5832902/234998105-87fc70a7-55de-4cbd-b4d3-d519653eadee.png)

Bad snapshot:

![bad](https://user-images.githubusercontent.com/5832902/234998182-e34efce4-a001-4529-82e5-5b0076928852.png)

The height is too short - it does not include output:

![Screenshot from 2023-04-27 22-47-17](https://user-images.githubusercontent.com/5832902/234998380-8e95cafb-3185-4e64-948e-229c4d3301f4.png)

HTML files for ease of inspection:

[good-snapshot.html.zip](https://github.com/jupyterlab/jupyterlab/files/11348226/good-snapshot.html.zip)
[problematic-snapshot.html.zip](https://github.com/jupyterlab/jupyterlab/files/11348227/problematic-snapshot.html.zip)

gabalafou:
Steps to reproduce the "height is too short" bug in latest JupyterLab:

1. Simulate an iPad or other touch device using browser dev tools
2. Create a new notebook
[...]

PR #14344, opened 2023-04-08
Adds buffer, uses zoom cursors for cell output collapser (supersedes #14266)

andrii-i:
## References

- Partial fix for https://github.com/jupyterlab/jupyterlab/issues/14252
- Supersedes #14266 by @JasonWeill 

## Code changes

Cell output collapser styling

## User-facing changes

The large "button" to collapse cell outputs goes all the way to the left edge of the output, making it difficult to start selecting text at the left edge of the output. This change introduces an 8px buffer where the button is absent.

Before:
image

After:
Screenshot 2023-04-07 at 6 35 36 PM
after_change
image

## Backwards-incompatible changes

None

PR #14388, opened 2023-04-16
Fix run-cells snapshot

krassowski:
## References

Incorrect snapshot was merged with https://github.com/jupyterlab/jupyterlab/pull/14344

I am not sure how to fix the underlying problem but it is tracked in https://github.com/jupyterlab/jupyterlab/issues/14385.

## Code changes

None

## User-facing changes

None

None


fcollonval (approved):
Thanks

Issue #14347, opened 2023-04-08
Trusting notebook in RTC mode leads to `File Changed` dialog


krassowski:
## Description

Regardless of fixes in https://github.com/jupyterlab/jupyterlab/pull/14345 the trust status is not set for empty (1 code cell) notebooks on RTC, and an attempt to Trust notebook ends up with `File Changed` dialog.

## Reproduce

![trust-rtc-itneraction](https://user-images.githubusercontent.com/5832902/230744083-35010694-aad3-432a-ad78-1025d9411cf8.gif)

1. Spin up RTC setup on binder
2. Open command palette
3. Choose "Trust Notebook"
4. Click Trust
5. See File Changed dialog

## Expected behavior

- whatever causes the `trusted` metadata to be lost on code cells when RTC is active gets ironed out.
- trusting notebook works the same on RTC as it works locally, without dialog.

## Context

- Browser and version: latest Chromium family
- JupyterLab version: 4.0.0b1

PR #14395, opened 2023-04-17
Upgrade ydoc

hbcarlos:
Includes https://github.com/jupyter-server/jupyter_ydoc/pull/161

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -20,7 +20,7 @@
     "@codemirror/language": "^6.0.0",
     "@codemirror/state": "^6.2.0",
     "@codemirror/view": "^6.7.0",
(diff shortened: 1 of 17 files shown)


fcollonval (commented):
Good catch

Usually when we update a package, we increase the version spec in the various package.json. Would you mind bumping the package.json spec to `"jupyter/ydoc": "^1.0.2"`.

krassowski (approved):
Thank you. More debugging will be needed as creating a notebook on binder still makes it untrusted by default (c-f https://github.com/jupyterlab/jupyterlab/issues/14347)

Visual regression failures are unrelated (you can merge or rebase and should go away).

fcollonval (approved):
Thanks @hbcarlos

Issue #9682, opened 2021-01-27
Alt Text and Marking Elements Decorative (for Accessibility)


manfromjupyter:
## Purpose:
Given WCAG requirement 2.5.3 (A), need to add alt texts and labels to elements that provide the context visual users have that screenreaders don't. In addition to provide a more efficient and less annoying or confusing experience, need to hide (for the screenreader) elements that are purely decorative.

## Notes
It's a big ticket that can absolutely be broken up. Luckily, none of these changes are dependent on this one getting done so can be done asynchronously with the other tickets.
Isabela (@isabela-pf) is going to work a bit on this and then can divide up from there. Aria-labels will need translations so will need to also tie in Gonzalo (@goanpeca) at some point before it's released.

Original from Main Ticket: #9399 (but now fully detailed)

**ALT text in packages**
> Expectation: The programmatic and visual labeling of all user interface components with text or images of text are the same. (WCAG Criteria 2.5.3 (A)).
    Observed: Images found under the packages section appear decorative but provide context of who is it that made the package/extension. Some of those labels such as IBM, Lab, Jupyter all include text within the image that is meant to provide additional information that a blind user wouldn’t have.
    Solution: Images found under the packages section need ALT text (ex. alt=”Created by IBM”) or hidden with aria-hidden=“true” and another field added to these listings that provide information on the maker.

saulshanabrook:
I see you broke up the sidebar issue. If we wanted to find other issues to break up, does anyone have suggestions on best ways to find other things we should add labels too?

manfromjupyter:
@saulshanabrook, so @tonyfast, @isabela-pf, and I was talking about just this and we think this one might be perfect for you guys to split up and work on because it's just adding single-line element attributes and allow you to get comfortable in the system. Aka, would just entail scavenging for elements by searching for classnames or ids most likely and then adding a `<description>`, `aria-label=""`, `alt text=""`, etc or hiding the element to screenreaders using an `aria-hidden="true"`. I'm not sure how to do these things in typescript but bet it'd be super quick to google. Below I included by best guess of how to do this given my own queries. If you or your friends get stuck, simply tag @marthacryan, @telamonian, or @jasongrout and they will be able to tell you the proper way to do this in the code and to format it in a way that follows the rest of the project so we're all following the best practices and formatting used in the Jupyter code base.

## How to split this ticket up
[...]

PR #14400, opened 2023-04-18
Alt text and marking elements decorative 

s596757:
## References

#9682

## Code changes
added aria-hidden to code to prevent decorative images from being picked up by screen reader and also edited code to allow left and right side bars to be read by screen reader 

## User-facing changes

side bar icons are read by screen readers but decorative images are ignored. All decorative imagesa re no longer picked up by screen reader

none


packages/launcher/src/widget.tsx
```diff
@@ -174,8 +174,12 @@ export class Launcher extends VDomRenderer<ILauncher.IModel> {
                 icon={icon}
                 iconClass={classes(iconClass, 'jp-Icon-cover')}
                 stylesheet="launcherSection"
+                title={cat + ' Icon'}
+                aria-hidden={true}
               />
-              <h2 className="jp-Launcher-sectionTitle">{cat}</h2>
+              <h2 className="jp-Launcher-sectionTitle" title={cat + ' Title'}>
```
fcollonval, line 180: What is the reasoning to add title on the icon and on the header? At first glance it does not bring much value.

packages/ui-components/src/icon/labicon.tsx
```diff
@@ -829,11 +829,20 @@ namespace Private {
...
     const titleNodes = svgNode.getElementsByTagName('title');
+    // make the title the aria label0
+
+    /// TODO set better unique id for title
     if (titleNodes.length) {
       titleNodes[0].textContent = title;
+      if (!titleNodes[0].id) {
+        titleNodes[0].id = performance.now().toString();
+      }
+      svgNode.setAttribute('aria-labelledby', titleNodes[0].id);
```
fcollonval, line 840: Looking at https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title, the `title` seems enough to provide an accessible _description_ for a SVG:

image

Which use case would you like to cover by adding the additional `aria-labelledby`?

packages/launcher/src/widget.tsx
s596757: reads out title in screenreader then

packages/ui-components/src/icon/labicon.tsx
```diff
@@ -829,11 +829,20 @@ namespace Private {
   export function setTitleSvg(svgNode: HTMLElement, title: string): void {
```
gabalafou, line 829: I think this whole function is maybe dead code.

When I search the codebase, I can find only one place that it's called, inside the protected `_initSvg` method. And when I search the codebase for "_initSvg", I find exactly one call site, in the `svgElement` getter, and that call site does not pass `title`. So it looks to me that `setTitleSvg` isn't getting called anywhere in the JupyterLab codebase. Maybe there's some extension out there that subclasses `LabIcon` and redefines `_initSvg`, but not in this repo.

gabalafou (changes requested):
It's really nice and exciting to see contributions to improve accessibility, thank you!

I have some doubts about these changes, though.

Before I weigh in, I should let you know that I do not use a screen reader regularly. In other words, I am a sighted developer who uses screen readers for development and testing.

I'm curious to know what operating system + screen reader combination you used to test the changes in this PR. The reason I ask is because I tested this PR using Mac OS + Voice Over, as well as Windows + NVDA. My doubts about this PR are based on my testing with those two screen readers. But perhaps you tested with a different screen reader and got different results.

I will now share the transcripts from my testing. First I will share the transcripts from NVDA (Windows) both before and after this PR. Then I will share the transcripts from Voice Over (Mac). All of the transcripts begin from the top of the region and stop at the level 2 heading labeled "Console." 

Transcript from NVDA before applying this PR:

```
heading level 2 Notebook
clickable Unlabeled graphic To get missing image descriptions, open the context menu
Python 3 (ipykernel)
heading level 2 Console
```

Transcript from NVDA after this PR applied:

```
Notebook Icon
heading level 2 Notebook
clickable Unlabeled graphic To get missing image descriptions, open the context menu
Python 3 (ipykernel)
Console Icon
heading level 2 Console
```

From my testing with NVDA, it seems that the main effect of this PR is to read out the icons. From my understanding, it is not helpful to read out icons for screen reader users (for example, refer to Sarah Soueidan's [Accessible Icon Buttons](https://www.sarasoueidan.com/blog/accessible-icon-buttons/#icon-sitting-next-to-text)).

I generated these transcripts in NVDA by using the "Say all" command (<kbd>NVDA</kbd> + <kbd>↓</kbd>). I also used the [object navigation shortcuts](https://www.nvaccess.org/files/nvda/documentation/userGuide.html#ObjectNavigation), and got the same results.

Now let's look at Voice Over.

Here's the transcript from Voice Over when loading JupyterLab from the current main branch:

```
notebook content, region
image
heading level 2, Notebook
Python 3 (ipykernel), group
To get missing image descriptions, open the context menu, Unlabeled image
Python 3 (ipykernel), group
Python 3 (ipykernel)
image
heading level 2, Console
```

Transcript from Voice Over with this PR applied:

```
notebook content, region
Notebook Icon, group
image
heading level 2, Notebook
Python 3 (ipykernel), group
To get missing image descriptions, open the context menu, Unlabeled image
Python 3 (ipykernel), group
Python 3 (ipykernel)
Console icon, group
image
heading level 2, Console
```

As with NVDA, the main effect seems to be to read the icon, which I don't think is desired. 

My hunch is that we should probably undo the places where the title attribute has been added, but we should keep the place where aria-hidden is added. However, if I'm missing or misunderstanding anything, please let me know. 

If we decide to go down the route I suggest, we will need to modify the LabIcon class so that the aria-hidden attribute gets passed through to the rendered DOM.

Once again, thank you for your contribution!

packages/launcher/style/base.css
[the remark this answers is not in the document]
s596757: I have removed the title attributes for the icons however I am unsure of how exactly you want me to amend the code to have labicon altered to pick up aria-labelledby by the dom and where you want this done. Any advice?

PR #14407, opened 2023-04-20
Improve scrolling for insert cell and run and advance

### LINKED ISSUE #14220: A new cell created below the viewport is not focused immediately
## Description

A new cell created below the bottom of the viewport is not visible immediately after it is created and focused.

cc @andrii-i @ellisonbg 

## Reproduce

1. Create a new notebook.
2. Create cells to fill the entirety of the browser viewport:

image

3. Press SHIFT+RETURN to run the bottommost cell.

Actual behavior (4.0.0a36): the bottommost cell runs and a new cell is created and focused, but is not visible.
image

## Expected behavior

Expected behavior (3.6.1): the bottommost cell runs and a new cell is created, focused, and made visible by scrolling the viewport.
image

## Context

- Operating System and version: macOS 12.6.3 "Monterey"
- Browser and version: 102.8.0esr (64-bit)
- JupyterLab version: 4.0.0a36

fcollonval:
## References

Fixes #14220

## Code changes

Improve scrolling into viewport when adding a cell or run and insert cell

## User-facing changes

Active cell is in viewport

None


krassowski:
This is certainly an improvement compared to 4.0.0beta1. This changes behaviour compared to 3.x:

| This PR | 3.6 |
|---|---|
| current-insert-cell | 3-6-behaviour |

I am not sure which is better for UX. Maybe giving user a choice would be best?

Of note, pasting does not scroll in either 3.6 nor 4.0beta1, nor on this PR:

copy-paste-does-not-work

I _think_ we should scroll on paste too. This PR could also have a scope expanded to cover pasting if this is easy enough, or it could be addressed later - I am not advocating one way or the other, but mentioning it in case if it influences implementation choices in this PR.

JasonWeill:
I agree with @krassowski that we should scroll on paste. The pasted cells should be made visible and editable as soon as they become visible.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -267,32 +267,36 @@ export abstract class WindowedListModel implements WindowedList.IModel {
...
+   * @param margin In 'smart' mode the viewport proportion to add
    * @returns The needed scroll offset
    */
   getOffsetForIndexAndAlignment(
     index: number,
-    align: WindowedList.ScrollToAlign = 'auto'
+    align: WindowedList.ScrollToAlign = 'auto',
+    margin: number = 0.25
   ): number {
+    const margin_ =
```
JasonWeill, line 278: The name `margin_` seems a bit confusing here; should we have a more descriptive name?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -267,32 +267,36 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    *
    * @param index Item index
    * @param align Where to align the item in the viewport
+   * @param margin In 'smart' mode the viewport proportion to add
```
JasonWeill, line 270: Might be good to document what 'smart' mode is, for the benefit of other devs

JasonWeill (approved):
Sorry for the delay in reviewing this. The logic here looks good; just a couple of minor suggestions for improvement.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -267,32 +267,36 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    *
```
fcollonval, line 267:
```suggestion
   *
   * By default, the list will scroll as little as possible to ensure the item is visible. You can control the alignment of the item though by specifying a second alignment parameter. Acceptable values are:
   *
   *   auto (default) - Scroll as little as possible to ensure the item is visible. (If the item is already visible, it won't scroll at all.)
   *   smart - If the item is already visible (including the margin), don't scroll at all. If it is less than one viewport away, scroll so that it becomes visible (including the margin). If it is more than one viewport away, scroll so that it is centered within the list.
   *   center - Center align the item within the list.
   *   end - Align the item to the end of the list
   *   start - Align the item to the beginning of the list
   *
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -302,22 +306,22 @@ export abstract class WindowedListModel implements WindowedList.IModel {
...
-          this._scrollOffset >= minOffset &&
-          this._scrollOffset <= maxOffset
+          this._scrollOffset >= bottomOffset &&
+          this._scrollOffset <= itemMetadata.offset
         ) {
           return this._scrollOffset;
-        } else if (this._scrollOffset < minOffset) {
-          return minOffset;
+        } else if (this._scrollOffset < bottomOffset) {
+          return bottomOffset + margin_ * size;
```
fcollonval, line 322:
```suggestion
          return bottomOffset + boundedMargin * size;
```

fcollonval:
Thanks both for testing and reviewing.

I'll have a look at the paste cell case and will look how to get an experience closer to 3.6.

fcollonval:

andrii-i:
Kicking CI

galata/test/jupyterlab/notebook-create.test.ts-snapshots/run-cells-jupyterlab-linux.png
andrii-i, line 1: Flakey snapshot, note "Click to add cell", needs reverting

krassowski (approved):

andrii-i:
- Created #14442 to facilitate discussion and work around difference with earlier versions that @krassowski [mentioned](https://github.com/jupyterlab/jupyterlab/pull/14407#issuecomment-1518591954) earlier

PR #14408, opened 2023-04-20
Only apply margin on `<select>` instead of using universal CSS selector

### LINKED ISSUE #14236: Universal selector (`*`) should not be used for LSP form
## Description

While I am not a  fan of current LSP UI setting in 4.0 alpha (of course I would prefer something like https://github.com/jupyter-lsp/jupyterlab-lsp/pull/778), there is a critical CSS issue in implementation - this line uses universal selector for no good reason:

https://github.com/jupyterlab/jupyterlab/blob/8230a1a41d48207ae05394607e368b196670cd6a/packages/lsp-extension/style/base.css#L6-L8

This should be removed/replaced before we release 4.0 to avoid affecting style calculation performance.

fcollonval:
## References

Fixes #14236

## Code changes

Apply margin only on the middle item

## User-facing changes

Slight visual change

| Before | After |
| --- | --- |
| image | image |

None

### DIFF
--- a/packages/lsp-extension/style/base.css
+++ b/packages/lsp-extension/style/base.css
@@ -3,6 +3,6 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
-.jp-LSPExtension-FormGroup-content > * {
-  margin-right: 25px;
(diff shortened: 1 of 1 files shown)


krassowski:
CI failure are unrelated merging.

PR #14414, opened 2023-04-20
Fix CI: remove/update broken docs links

krassowski:
Fixes failing `Linux Tests / check_links (pull_request)` job across recent pull requests.

It looks like `jupyterlab.github.io` stopped working. It is probably better to keep deploying docs on RDT only (and https://github.com/jupyterlab/jupyterlab/tree/gh-pages was not updated for three years now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 **[Installation](#installation)** |
-**[Documentation](http://jupyterlab.readthedocs.io)** |
+**[Documentation](https://jupyterlab.readthedocs.io)** |
 **[Contributing](#contributing)** |
 **[License](#license)** |
 **[Team](#team)** |
 **[Getting help](#getting-help)** |
 
-# [JupyterLab](http://jupyterlab.github.io/jupyterlab/)
+# [JupyterLab](https://jupyterlab.readthedocs.io)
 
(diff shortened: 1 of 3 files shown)


krassowski:
Hmm, there are at least two more occurrences:
```
RELEASE.md: http://jupyterlab.github.io/jupyterlab/
README.md: http://jupyterlab.github.io/jupyterlab/
```

we should not have http either way, but why is it not working?

krassowski:
In particular RELEASE.md says:

> ## Updating API Docs
>
> Run `source scripts/docs_push.sh` to update the `gh-pages` branch that backs http://jupyterlab.github.io/jupyterlab/.

Is this script being used during releases? Why is the contents of https://github.com/jupyterlab/jupyterlab/tree/gh-pages 3 years old?

jtpio:
> Is this script being used during releases? Why is the contents of https://github.com/jupyterlab/jupyterlab/tree/gh-pages 3 years old?

It's likely not used during releases. iirc this had to be done manually before.

Maybe we should remove that from RELEASE.md. And re-add it later as part of the CI using `actions/deploy-pages` instead of pushing to `gh-pages` if needed. Or just keep everything under RTD yes.

jtpio (approved):
Thanks!

krassowski:

PR #14415, opened 2023-04-21
Harmonizes form renderer ids

brichet:
This PR is simply to harmonize some ids of `IFormRenderer`, by starting them with the full plugin id.

## References

None

## Code changes

Some form renderers IDs.

## User-facing changes

None

None


fcollonval (approved):
Thanks @brichet 

> CI failure is not related

I'm merging this right away as it change the ids that should be considered as API as they are unique identifiers.

Issue #14372, opened 2023-04-12
On a notebook, no option selectable in long Text Editor Syntax Highlighting submenu


JasonWeill:
## Description

Observed by @Carreau in https://github.com/jupyter/notebook/issues/6825.

When the user has a notebook open, the View > Text Editor Syntax Highlighting menu item is still available. None of the submenu items are active. Some are quite obscure (e.g., Brainfuck, not generally used for serious coding) and some are misspelled (e.g., "CoffeScript" for CoffeeScript). This menu item should be absent for notebook documents and its list of supported languages should be curated or customizable.

![image](https://user-images.githubusercontent.com/93281816/231529845-88fb7664-1951-4b84-a8c4-c1d7e68aa968.png)

## Reproduce

1. Open a notebook
2. Open the View > Text Editor Syntax Highlighting submenu
3. Observe that no items are selectable

## Expected behavior

Submenu contains items that I can select to activate

JasonWeill:
Alternatively, the submenu should be absent entirely when the user is not using a text editor.

Carreau:
> We could remove this menu entirely from JupyterLab

I would not do that, I would have never thought of clicking on the status bar. I think it's better to keep all options on the menus.

PR #14418, opened 2023-04-23
Fix typo in CoffeeScript label

krassowski:
## References

Addresses a typo reported in #14372

## Code changes

```diff
-CoffeScript
+CoffeeScript
```

## User-facing changes

As a bove

N/A

### DIFF
--- a/packages/codemirror/src/language.ts
+++ b/packages/codemirror/src/language.ts
@@ -679,7 +679,7 @@ export namespace EditorLanguageRegistry {
       },
       {
         name: 'CoffeeScript',
-        displayName: trans.__('CoffeScript'),
+        displayName: trans.__('CoffeeScript'),
         alias: ['coffee', 'coffee-script'],
         mime: [
           'application/vnd.coffeescript',


JasonWeill:
Should we continue to offer CoffeeScript as a language for syntax highlighting at all? Its [last release](https://github.com/jashkenas/coffeescript/tags) was a year ago today, and most of the interesting features of CoffeeScript have already been released in JavaScript/ECMAscript or TypeScript.

krassowski:
No strong opinion here, but we would not be saving much by not including it. Even if the language was officially unmaintained, I would give users some more time to migrate away before actively dropping it; we do not have any overhead here as it is all provided by CodeMirror out of the box as a legacy syntax highlighting mode.
Also, it may be easier to just say that we just include everything that CodeMirror offers than discuss each on individual basis.

JasonWeill (approved):

JasonWeill:
Sounds good; we can talk about curating this list in #14420.

jtpio (approved):
Thanks!

Issue #14351, opened 2023-04-10
Documenting the requirements of `OK` extension status


RRosio:
### Problem
In trying to update some JupyterLab extensions to the alpha/beta releases of JupyterLab 4 and Lumino 2, I have run into an issue:  The status of the extension is not `OK` i.e. I would hope to port my extension so that it should be showing as  `@jupyterlab/myextension v1.0.0 enabled OK (python, myextension)` when I run the command `jupyter labextension list`. I instead get 
```
@jupyterlab/myextension v1.0.0 enabled  X

   The following extensions are outdated:
        @jupyterlab/myextension
        
   Consider running "jupyter labextension update --all" to check for updates.
```

### Proposed Solution
- Documenting this issue in the extension migration guide or a solution to this issue for those that may potentially encounter it. 
- More documentation provided about how the status of an extension is determined, both in the docs and as a part of the message delivered to a developer in the terminal.

### Additional context
Steps to I take to update an extension:
1. Update the Jupyterlab and Lumino package dependencies to the most recent versions available on npm, for example `"@jupyterlab/application": "^4.0.0-beta.1",`
2. Update TypeScript and React version.
3. Update target version in TSConfig.
4. Update JupyterLab version in `pyproject.toml` and `setup.py`.
5. Fix compiling errors due to API changes
6. Build and install the extension

When using the cookiecutter to upgrade the extension, via the upgrade script `python -m jupyterlab.upgrade_extension .`, after fixing the compile errors, the extension status does get changed to `OK`. I am not sure what I am missing from my initial attempt to port the extension. Even if maybe developers are encouraged to use the upgrade script, this information might still be valuable for more comprehensively understanding what is required to be done to update extensions.

dlqqq:
@RRosio I did a bit of digging into this issue recently when [upgrading Jupyter Scheduler to JupyterLab 4](https://github.com/jupyter-server/jupyter-scheduler/pull/402) and happened to stumble on this. I agree this should be communicated explicitly in the JL 3->4 migration guide, as the default labextension cookiecutter CI will fail unless its status is `OK`. 

The validation logic is implemented under `jupyterlab/commands.py::_AppHandler#_get_extension_compat()`, which calls `jupyterlab/commands.py::_validate_compatibility()` to validate your extension’s JS dependencies against a mysterious `core_data` object:

```python
def _validate_compatibility(extension, deps, core_data):
    """Validate the compatibility of an extension."""
[...]

PR #14428, opened 2023-04-24
Updating extension status in migration guide

RRosio:
## References
Addresses https://github.com/jupyterlab/jupyterlab/issues/14351

## User-facing changes

Documentation update in Extension Migration Guide

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -88,6 +88,15 @@ is to force packages deduplication using:
 
    jlpm dedupe --strategy highest
 
+Extension Status
+^^^^^^^^^^^^^^^^
+The extension has four possilbe status reportings: red 'X', red 'DSIABLED', green 'OK', and green 'ENABLED'.
+The extension status is determined in the following ways:
+- Red 'X': the extension is reporting compatibility errors
+- Red 'DISABLED': the extension is disabled.
+- Green 'OK': the extension is installed.
+- Green 'ENABLED': the extension is installed and enabled.
+
 API breaking changes
 ^^^^^^^^^^^^^^^^^^^^
 


docs/source/extension/extension_migration.rst
```diff
@@ -88,6 +88,15 @@ is to force packages deduplication using:
 
    jlpm dedupe --strategy highest
 
+Extension Status
+^^^^^^^^^^^^^^^^
+The extension has four possilbe status reportings: red 'X', red 'DSIABLED', green 'OK', and green 'ENABLED'.
```
echarles, line 93:
```suggestion
The extension has four possible status reportings: red 'X', red 'DSIABLED', green 'OK', and green 'ENABLED'.
```

echarles:
The proposed changes do not cover the point raised in the issue: `Documenting this issue in the extension migration guide or a solution to this issue for those that may potentially encounter it.` - Do you intent to document that?

docs/source/extension/extension_migration.rst
krassowski: Also `DSIABLED` → `DISABLED`

docs/source/extension/extension_migration.rst
```diff
@@ -88,6 +88,22 @@ is to force packages deduplication using:
...
+The extension status is determined in the following ways:
+- Red 'X': the extension is reporting compatibility errors
+- Red 'DISABLED': the extension is disabled.
+- Green 'OK': the extension is installed.
+- Green 'ENABLED': the extension is installed and enabled.
+
+When upgrading extensions manually you may encounter the red 'X' status that indicates compatibility errors
+in the dependencies of your extension. It is likely that there are dependencies, in addition those which
+are JupyterLab packages, that are being compared to the dependencies of core JupyterLab and whose versions
+are not coinciding. Some of those dependencies may include the following: yjs, react, react-dom, typescript,
```
fcollonval, line 103: did you encounter an error with typescript? If so could you please provide an error trace as this should never happen for development dependencies (like typescript) - only runtime dependencies may be incompatible.

```suggestion
are not coinciding. Some of those dependencies may include the following: yjs, react, react-dom,
```

fcollonval (changes requested):
Thanks @RRosio 

The code used is at: 

https://github.com/jupyterlab/jupyterlab/blob/7ba18f10e7465789d7ef49c75595b2af87f18f7c/jupyterlab/commands.py#L1523

- Which dependencies are compared?  
  All direct `dependencies` of the extensions are compared to the packages part of the shared webpack namespace defined at https://github.com/jupyterlab/jupyterlab/blob/7ba18f10e7465789d7ef49c75595b2af87f18f7c/dev_mode/package.json#L268 with the version specified at https://github.com/jupyterlab/jupyterlab/blob/7ba18f10e7465789d7ef49c75595b2af87f18f7c/dev_mode/package.json#L19
- Is there a way for a user to list the dependencies which are compared to JupyterLab core?  
  You could add a new script in `conf.py` to generate a documentation file from the `dev_mode/package.json`. See that example building a documentation file from the commands list:
https://github.com/jupyterlab/jupyterlab/blob/7ba18f10e7465789d7ef49c75595b2af87f18f7c/docs/source/conf.py#L195
- Is there an additional place where this documentation may be helpful? Does it already exist?  
  Unfortunately none that I'm aware of.

- I’d like to add a section pointing extension developers to sections of JupyterLab documentation which may be relevant to understanding this process or at least aspects of this process, does anyone have suggestions for what I could link here?

Unfortunately I don't think so.

PR #14441, opened 2023-04-26
Improve header space use, restore description, remove unused styles

### LINKED ISSUE #14403: Should we use space better in the setting editor UI?
### Problem

PR https://github.com/jupyterlab/jupyterlab/pull/14074:
- removes the descriptions for plugins
- moves the button to new line (which is called a button bar even though there is only one, hard-coded button)

this does not appear to be an optimal use of space.

### Proposed Solution

| With PR #14704 in | Proposed |
|---|---|
| image | porposed-changes |

I am also proposing to use a font in between the very strong one used in #14704 and the light one used previously. Finally I would propose slightly reducing the size of the "Restore to Defaults" button (not shown on the mockup).

Large-size proposed:

porposed-changes

### Additional context

Previous font for context:

Screenshot from 2023-04-18 20-09-25

Thoughts?

krassowski:
## References

Closes #14403 

## Code changes

- Rework header
- Reduce CSS selectors complexity
- Move CSS rules together into thematic groups
- Remove unused rules:
  - `.jp-SettingEditor-header` (no such an element)
  - rules for `.jp-SaveSettingsBanner` (no such a banner)
  - `.jp-SettingsPanel .jp-SettingEditor-Switch .jp-switch` (no such a switch)

## User-facing changes

| Before | After |
|--|--|
|Screenshot from 2023-04-26 20-48-02 | Screenshot from 2023-04-26 20-46-52 |

N/A


andrii-i:
Thank you @krassowski. Could we decouple (1) adding plug-in descriptions back into header from (2) moving "Reset to default" button back into header? I think 1st is a good idea now when plugin descriptions were added while 2nd is debatable and we could use more consultations or user feedback do decide on it

krassowski:
> and we could use more consultations or user feedback do decide on it

Equally well we could perform such consultations after this is merged, as after all this only restores the placement as in JupyterLab 3. Unless https://github.com/jupyterlab/jupyterlab/issues/14007 was based on user feedback/evaluation - if so, please share it so we can evaluate.

andrii-i:
> Equally well we could perform such consultations after this is merged, as after all this only restores the placement as in JupyterLab 3. Unless https://github.com/jupyterlab/jupyterlab/issues/14007 was based on user feedback/evaluation - if so, please share it so we can evaluate.

As far as I understand [Jupyter decision making guide](https://jupyter.org/governance/decision_making.html), we seek informal consensus and call a vote if it can't be reached. We [reached](https://jupyter.org/governance/decision_making.html) informal consensus on #14074 before merging it and [agreed](https://github.com/jupyterlab/jupyterlab/pull/14074#discussion_r1170475971) to discuss button bar, max-width threshold, and other questions such as plugin descriptions and pluginlist background color separately. I'm open to discuss all these issues here and try to reach consensus but decoupling them into separate issues might be more efficient as for example we already have consensus on plugin descriptions.

Decision to place button into body vs header was driven by desire to reduce visual clutter in the header and to use "buttonbar" design pattern, a pattern that puts all page-level actions at inside the same 1-layer right-aligned area called  buttonbar that should be positioned at the top of the widget or at the bottom of the dialog. We already implicitly use it in Dialog component and contributors explicitly introduce it, for example in the announcement popup [here](https://github.com/jupyterlab/jupyterlab/pull/12959/files#diff-c41da044f23b3ed4759e16358ece487cc20c11f77d20da38f3adff7e53cf7f96R806). Here's a first link google gave me describing this pattern: [Abode Buttonbar](https://developer.adobe.com/commerce/admin-developer/pattern-library/controls/button-bar/). It is also based on consultations with experts as our team have extensively discussed where buttons should be located and arrived to the decision of using this pattern in [jupyter scheduler ](https://jupyter-scheduler.readthedocs.io/en/latest/users/index.html#job-detail).

I agree that we can have a discussion about it. Is that correct that you think that button should be placed inside the header to save vertical space? Let's discuss it.

We should work out and codify basic JupyterLab design principles that would answer questions like this (where page-level actions should be located) to be able to reference them during discussions like this.

krassowski:
> decoupling them into separate issues might be more efficient

Sorry, I am out of my free time. Please appreciate that I am doing this in my free unpaid time. I did open an issue (https://github.com/jupyterlab/jupyterlab/issues/14403) for discussion. It was open for over a week and both you and @JasonWeill gave a thumbs up to the mockup that I prepared there without reservations. I am not going to put more time here as I simply have to earn bread for family.

> It is also based on consultations with experts

Can we have these experts engage directly with us, please?

> to use "buttonbar" design pattern

Subjectively, I find this as a circular reasoning for the sake of one button.

andrii-i:
I most definitely appreciate your time @krassowski. Would you be open to me modifying your PR / opening a separate issue for the button if you don't have time take care of it? 

> Can we have these experts engage directly with us, please?

This question does not make sense to me. We discussed this before, I define experts as (1) long-time contributors and (2) designers. Under this definition you, me, and other contributors are all experts. You can see who worked on the jupyter scheduler under [contributors](https://github.com/jupyter-server/jupyter-scheduler/graphs/contributors) and I just wrote out arguments that we used to arrive at the decision. 

I presented you my arguments and ask you to do the same. Why button should be placed into the header?

> Subjectively, I find this as a circular reasoning for the sake of one button.

Well, that's why we could really use formalized design principles, to lean on them in the situations like this where both options have arguments both for and against them.

andrii-i:
Bot please update galata snapshots

andrii-i (approved):
This restores design parity with JupyterLab 3, let's continue discussions on button placement and other questions in separate issues

andrii-i:
Bot please update galata snapshots

PR #14444, opened 2023-04-27
Fix custom CSS in Virtual notebook rendering

### LINKED ISSUE #14379: Virtual Notebook rendering "breaks" custom CSS in JupyterLab 4?
## Description

In Jupyter notebooks it is common to inject custom CSS (for example with the `%%html` magic) to style the notebook or tweak a few things, without having to write an extension. This is useful for example when making presentations.

However with JupyterLab 4 it looks like doing so might produce some unwanted behavior and UI glitches when using the notebook.

This is the case even for small notebooks.

## Reproduce

Here are two screencasts to compare `4.0.0b1` against `3.6.3` using the following code snippet:

```
%%html

<style>
    .jp-Notebook-cell { margin-bottom: 400px; }
</style>
```

**4.0.0b1**

Scrolling slowly with the mouse we can clearly notice "UI shifts". It's even more apparent when dragging the scrollbar. 

https://user-images.githubusercontent.com/591645/232089176-5c238e0b-57e7-422b-9845-c3a5a9655da5.mp4

**3.6.3**

Doing the same in 3.6 is smooth:

https://user-images.githubusercontent.com/591645/232089153-5391306f-8d28-49aa-9b59-b2f43f24c08c.mp4

## Expected behavior

The same scroll experience as in 3.6.

## Context

Tested with JupyterLab `4.0.0b1`.

I *suspect* this is related to the virtual notebook implementation, but feel free to correct this and the title of the issue if not.

fcollonval:
## References

Fixes #14379 

## Code changes

Loop over all cells at initial rendering (during CPU idle cycle) to render code cell with html outputs containing styles and scripts that may be contaminant outside their cell (e.g. custom CSS injection). 

## User-facing changes

Custom CSS and JS snippet are loaded at notebook opening whatever their position in the notebook.

None


jtpio:
Thanks for working on this!

docs/source/getting_started/faq.rst
```diff
@@ -16,6 +16,38 @@ General
...
+
+Notebook
+^^^^^^^^
+
+My notebook is displaying cell outputs in `iframe <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe>`__. They are reset when scrolling back and forth.
+
+    Since JupyterLab v4, notebook rendering is optimized to display only the cells needed.
+    This has side effects with iframes.
+
+    The current workaround is to set the settings *Notebook* => *Windowing mode* to ``defer`` or ``none``.
```
jtpio, line 30: This brings the question: what default do we want?

Maybe the default should be `none`, and those who want performance could easily switch to `defer` via this settings?

Thinking that the majority of notebook users are dealing with small notebooks, and would not really benefit much from the performance improvements. And those who are dealing with large notebooks can then check the FAQ or the performance section of the docs to switch to `defer`.

jtpio: I'll open a new issue to discuss this and not block this PR.

jtpio: https://github.com/jupyterlab/jupyterlab/issues/14455

jtpio:
Curious if that also fixes customizing the notebook CSS via a regular JupyterLab extension, as mentioned in https://github.com/jupyterlab/jupyterlab/issues/14379#issuecomment-1513682437?

fcollonval:
> Curious if that also fixes customizing the notebook CSS via a regular JupyterLab extension, as mentioned in [#14379 (comment)](https://github.com/jupyterlab/jupyterlab/issues/14379#issuecomment-1513682437)?

The issue is not due to the way CSS rules are added but due to the rule itself. `margin` + DOM size measurement = :boom: - so it will break virtual rendering. The solution is to use padding.

fcollonval:
This is one of the FAQ addition in this PR

jtpio:
Ah ok thanks :+1:

packages/notebook/src/widget.ts
```diff
@@ -892,6 +900,44 @@ export class StaticNotebook extends WindowedList {
...
+    //   directly evaluate without adding DOM elements (see @jupyterlab/javascript-extension)
+    if (cell instanceof CodeCell) {
+      for (
+        let outputIdx = 0;
+        outputIdx < cell.model.outputs.length;
+        outputIdx++
+      ) {
+        const output = cell.model.outputs.get(outputIdx);
+        const html =
+          (output.data[htmlRendererFactory.mimeTypes[0]] as string) ?? '';
```
krassowski, line 928: What happens if another `mimeType` is added to `htmlRendererFactory`? Should we either;
- hard-code `html/text` to have agreement with what is written in user-facing FAQ, or
- iterate over all `mimeTypes`

packages/notebook/src/widget.ts
```diff
@@ -892,6 +900,44 @@ export class StaticNotebook extends WindowedList {
...
+    //   We don't need to render JavaScript mimetype outputs because they get
+    //   directly evaluate without adding DOM elements (see @jupyterlab/javascript-extension)
+    if (cell instanceof CodeCell) {
+      for (
+        let outputIdx = 0;
+        outputIdx < cell.model.outputs.length;
+        outputIdx++
+      ) {
+        const output = cell.model.outputs.get(outputIdx);
+        const html =
```
krassowski, line 927: Should we also check for `javaScriptRendererFactory` and render it unconditionally? For example, in IPython execute:

```
%%javascript
alert('test')
```

packages/notebook/src/widget.ts
```diff
@@ -892,6 +900,44 @@ export class StaticNotebook extends WindowedList {
...
+          (output.data[htmlRendererFactory.mimeTypes[0]] as string) ?? '';
+        if (
+          html.match(
+            /(<style[^>]*>[^<]*<\/style>|<script[^>]*>.*?<\/script>)/gms
+          )
```
krassowski, line 932: So this will always be a hard problem with false positives and negatives. False positives are things like commented out `<script>`:

```html
<!--- <script> </script> --->
```

And false negatives:

```html
<img src="doesnotexist" onerror="alert(1)">
````

I think that this is fine to keep it as-is for transitional period and fail on edge cases (false positive cost is a cell rendered too early, false negative is some very advanced users getting slightly upset but those will know how to workaround it).

However, in the bigger picture, supporting this pattern goes slightly against the goal of introducing style isolation and JS isolation for outputs (shadow DOM). Here is what I would propose:
- we always render cells with `javaScriptRendererFactory` outputs
- we create a new rendered `cssRendererFactory` (name could be different) and always render cells with it too
- we keep rendering cells with HTML if those meet the regex condition **but** emit a future warning in the console (telling users to switch to js/css outputs for notebook-wide effects, or ignore the warning if they intend to decorate HTML content in this output) and this will enable us by JupyterLab 5 or 6 to roll out shadow DOM isolation for HTML outputs (with JS and CSS outputs being rendered as-is).

I do not feel strongly about the plan above, if you don't immediately agree we can hoist it to a new issue. I only feel strongly that we should include `javaScriptRendererFactory`.

krassowski (commented):

packages/notebook/src/widget.ts
fcollonval: I would rather limit to `html/text`.

packages/notebook/src/widget.ts
fcollonval: We don't need it per the above comment (line 917); - to explicit it here - it already works as the JavaScript will not be injected as a `script` element in the DOM but directly evaluated.

> This is tested in the updated `windowed_notebook.ipynb`.

packages/notebook/src/widget.ts
fcollonval: > So this will always be a hard problem with false positives and negatives.

Yep - there is space to improve this as I'm not checking the _preferred_ output that could be different than `text/html` for instance.

> JavaScript factory

The JavaScript case is already working - see above comment.

> `cssRendererFactory` 

You mean that we should introduce a `%%css` magic or that the cell output with mimetype `text/css` should be rendered?

> Warning

I like the idea; especially as the notebook with contaminent outputs is likely not working as expected on other applications like Colab, VS Code,... .

fcollonval:

I was looking to apply it to the JavaScript factory but in fact it is not needed because we evaluate the code directly. We don't inject the code as `script` element.
But I added a test to check for it :wink:

PR #14459, opened 2023-04-28
Force rerendering if changes happen during an update.

### LINKED ISSUE #14385: "Click to add a cell" is flaky on CI

fcollonval:
## References

Fixes #14385

> Thanks a lot to @gabalafou for the reproducible test and @krassowski to find the root cause.

## Code changes

Force re-rendering the notebook if a change happens during a update.

## User-facing changes

The viewport content and height should be updated when needed.

None

### DIFF
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -1059,9 +1059,9 @@ export class WindowedList<
         this.scrollToItem(...this._scrollToItem).catch(reason => {
           console.log(reason);
         });
-      } else {
-        this.update();
       }
+
+      this.update();
     }
   }
 
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -905,10 +905,17 @@ export class WindowedList<
     if (this.viewModel.windowingActive) {
       // Throttle update request
       if (this._scrollRepaint === null) {
+        this._needsUpdate = false;
         this._scrollRepaint = window.requestAnimationFrame(() => {
           this._scrollRepaint = null;
           this._update();
+          if (this._needsUpdate) {
+            this.update();
+          }
         });
+      } else {
+        // Force re rendering if some changes happen during rendering.
+        this._needsUpdate = true;
       }
     } else {
       this._update();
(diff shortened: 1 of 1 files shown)


gabalafou (commented):
I can confirm that this fixes https://github.com/jupyterlab/jupyterlab/issues/14385.

I think you're probably right that there's a bug here—as in, if an update is requested while a scroll repaint is queued, then we probably need to request another update after the scroll repaint is executed. (In the case of issue 14385, here's [the update request that happens while the scroll repaint is queued](https://github.com/jupyterlab/jupyterlab/blob/acb87d51ed0c489fb5a8f6781059024cf62f255b/packages/notebook/src/actions.tsx#L2222).)

However, I wonder if there's another bug at play. I was trying  to understand why `shift` + `enter` (run cell) was fine whereas `ctrl`/`accel` + `enter` (run cell and advance) was not, and this led me to some lines in the widget resize handler in the `WindowedList` class:

https://github.com/jupyterlab/jupyterlab/blob/acb87d51ed0c489fb5a8f6781059024cf62f255b/packages/ui-components/src/components/windowedlist.ts#L1048-L1058
The reason why the simple "run" commands work correctly is because they do NOT call `WindowedList.scrollToItem`, so when the markdown cell widget is resized, the second block of the above if-else, which calls `WindowedList.update`, is followed. But "run and advance" calls `WindowedList.scrollToItem`, which sets `this._scrollToItem`, so the widget resize handler follows the first block of the if-else, which doesn't request an update. 

I'm not familiar enough with the `WindowedList` class to understand the intent behind this if-else, but it seems a mistake not to queue an update request on when one of the widgets has triggered a resize event and the total size of notebook has changed.

I wonder if the code above should be:

```ts
   // If some sizes changed 
   if (this.viewModel.setWidgetSize(newSizes)) { 
     // Update the list 
     if (this._scrollToItem) { 
       this.scrollToItem(...this._scrollToItem).catch(reason => { 
         console.log(reason); 
       }); 
     }
     this.update(); 
   } 
```

krassowski:
I did not have time to delve further, but just leaving a note to highlight that `Notebook Create › Run cells` is still failing on this PR in `UI Tests / Visual Regression Tests (pull_request)`.

fcollonval:
Thanks @gabalafou I implemented your suggestion and the test works for me locally.

With only that `this.update` forced all the time your reproducible case is solved also. But I kept the original update re-trigger as indeed it seems required.

Let's see what is the CI test result now.

fcollonval:
All tests are green. Merging it to include it in the next rc.

PR #14460, opened 2023-04-28
add native viewing support for jsonl/ndjson

timkpaine:
## References

`jsonlines` / `ndjson` have traction (some examples lazily collected from the web, [shopify](https://shopify.dev/docs/api/usage/bulk-operations/imports#create-a-jsonl-file-and-include-graphql-variables), [duckdb](https://duckdb.org/docs/extensions/json.html)).

Since the differences are very trivial wrt existing `json` support, it isn't really worth it to defer this functionality to e.g. a community extension. 

## Code changes

Register `jsonl` and `ndjson` file types, and corresponding mime types. NOTE: standardization is trailing utilization, so it is messier than one would prefer.
https://github.com/ndjson/ndjson-spec/issues/35
https://github.com/wardi/jsonlines/issues/22

Then tweak the JSON extension to convert `jsonl` to `json` when the mime type is that of `jsonl`. 

## User-facing changes

`.jsonl` and `.ndjson` files will now appear as `json` files and open in the default json opener. 

tmp

None. May effect existing mime render extensions for `jsonl` and `ndjson`, if these exist.

### DIFF
--- a/packages/json-extension/src/index.tsx
+++ b/packages/json-extension/src/index.tsx
@@ -23,6 +23,12 @@ const CSS_CLASS = 'jp-RenderedJSON';
  * The MIME type for JSON.
  */
 export const MIME_TYPE = 'application/json';
+// NOTE: not standardized yet
+export const MIME_TYPES_JSONL = [
+  'text/jsonl',
+  'application/jsonl',
+  'application/json-lines'
+];
 
 /**
  * A renderer for JSON data.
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks @timkpaine I targeted this to 4.0.0 to discuss on Wednesday weekly call to see if we accept this as late addition (we have release RC so no promise). Otherwise it will need to move on 4.1.0.

packages/json-extension/src/index.tsx
```diff
@@ -23,6 +23,12 @@ const CSS_CLASS = 'jp-RenderedJSON';
  * The MIME type for JSON.
  */
 export const MIME_TYPE = 'application/json';
+// NOTE: not standardized yet
```
krassowski, line 26: If we delay merging this PR until 4.1, is there a chance these would be standardized by then, or is it not worth waiting for it?

timkpaine: not worth waiting, i doubt this will be standardized soon

fcollonval:
The consensus at the dev meeting was too push this to 4.1.0

packages/json-extension/src/index.tsx
sp4ce: Hi, I am working on helping to standardize `application/jsonl` MIME_TYPE, but I have no experience in doing so, so I confirm it will take me a bit of time to write the RFC and submit it, so it's not worth waiting.

fcollonval:
I'll rebase this PR on `main` as we branch out 4.0.x to be able to merge 4.1 feat in `main`.

fcollonval (approved):
Thanks @timkpaine

PR #14462, opened 2023-04-30
Update contributing.rst

gabalafou:
Put quotes on docs install command to make it zsh friendly


krassowski (approved):
Thank you @gabalafou!

PR #14465, opened 2023-04-30
Fix documentation formatting

JaSpa:
Use correct RST formatting. Before:

Screenshot from the "Getting Started" documentation page showing botched markup: the text reads "… e.g. C: on Windows or / on Linux" with a backtick after "C:" and before the slash and the initial part in italic.


krassowski (approved):
Thank you @JaSpa!

Looks good on preview:

Screenshot from 2023-05-01 18-16-57

Issue #14329, opened 2023-04-05
4.0.0b0: impor<tab> -> importimport


Carreau:
Completion has this issue:

```
impor<tab><tab><tab><tab><tab><tab>
```
Lead to `importimport`

See https://github.com/jupyter/notebook/issues/6818 (found in notebook v7)

krassowski:
Confirmed, confirmed not present in 3.6 tagging as regression.

davidbrochart:
It looks similar to https://github.com/jupyterlite/jupyterlite/issues/961.

Carreau:
I think this one is triggered when you reach the end of the completer list in jupyterLab in it loops over back to the top, so it might be different.

afshin:
I'm still working on this issue, but I've removed the `Release Blocker` label.

PR #14475, opened 2023-05-02
Fix completer bug with cycling through options

afshin:
There is a regression in the completer that causes it to behave incorrectly when the user cycles all the way through the options and loops back to the top of the list.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14329

## Code changes

- `getTokenAt` was improved to short-circuit if a token has already been discovered as discussed [in this comment](https://github.com/jupyterlab/jupyterlab/pull/14475#discussion_r1183605532) which was the cause of [issue #14329](https://github.com/jupyterlab/jupyterlab/issues/14329)
- all changes in `completer` package are purely stylistic

## User-facing changes

If the selected item in the completer is an exact match for the text that has already been typed, it will not be erroneously inserted into the text.

**Before**

https://user-images.githubusercontent.com/159529/236578645-248ed66c-8175-47c8-a4b2-c1a2cd2fe44b.mov

**After**

https://user-images.githubusercontent.com/159529/236578667-50655ef7-8572-4615-bded-76bd396586b1.mov

N/A


trungleduc:
@afshin i don't have pushing right on this PR

packages/completer/src/model.ts
```diff
@@ -336,6 +336,7 @@ export class CompleterModel implements Completer.IModel {
     }
 
     let { start, end } = cursor;
+    console.log('start', start, 'end', end);
```
trungleduc, line 339:
```suggestion
```

afshin:
Looks like the test failure is a legitimate issue (either with the test or with the logic in this PR).

packages/completer/src/model.ts
```diff
@@ -368,13 +368,13 @@ export class CompleterModel implements Completer.IModel {
...
     for (let item of items) {
       // See if label matches query string
-      // With ICompletionItems, the label may include parameters, so we exclude them from the matcher.
+      // With ICompletionItems, the label may include parameters,
+      // so we exclude them from the matcher.
       // e.g. Given label `foo(b, a, r)` and query `bar`,
       // don't count parameters, `b`, `a`, and `r` as matches.
       const index = item.label.indexOf('(');
-      let prefix = index > -1 ? item.label.substring(0, index) : item.label;
-      prefix = escapeHTML(prefix);
```
krassowski, line 376: Yes, the test failures are legit. We absolutely need to keep escaping HTML here.

krassowski (changes requested):
I would appreciate some hints on why you think the changes as proposed here are needed, for example by populating the `Code changes` section :)

> Btw, I can help to fix this regression in a counter PR as it is likely related to refactoring that either I did or I approved - just let me know if help is needed.

Also, I think we should add a test case for the regression.

packages/codemirror/src/editor.ts
```diff
@@ -522,7 +522,7 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
           if (node.node.firstChild !== null) {
             return true;
           }
-          if (offset >= node.from && offset <= node.to) {
+          if (!token && offset >= node.from && offset <= node.to) {
```
trungleduc, line 525: Without the check `!token`, after the first matched node, we continue to iterate through the tree and then match other nodes. The returned token is from the last matched node.
In the case of `import`, the first matched token is `{value: 'import', offset: 0, type: 'import'}` but the returned value is `{value: '', offset: 6, type: '⚠'}`, which causes the cycling issue.
I suppose that the first matched node is always the correct one so the `!token` check will ignore all other matched nodes. This fix is tested with other keywords (`class`, `def`,...) and it works as expected.

afshin (commented):
bot please update snapshots

fcollonval:
Seems that the bot was not triggered (maybe due to the GitHub troubles).

Bot please update galata snapshots

fcollonval:
Ok this time, the job got triggered: https://github.com/jupyterlab/jupyterlab/actions/runs/4890958620

@krassowski I have a doubt about the failure for search on typing. Do you think the new behavior is legit or not?

krassowski:
I will take a look later, I want to test some edge cases.

fcollonval:
Thanks for the reply Mike. Actually it is not related to this PR, the failure is happening there too e.g.: https://github.com/jupyterlab/jupyterlab/actions/runs/4881704634?pr=14480

fcollonval:
Kicking the CI

packages/completer/src/model.ts
```diff
@@ -368,13 +368,14 @@ export class CompleterModel implements Completer.IModel {
...
-      let match = StringExt.matchSumOfSquares(prefix, query);
+      const text = index > -1 ? item.label.substring(0, index) : item.label;
+      const match =
+        text !== query && StringExt.matchSumOfSquares(escapeHTML(text), query);
```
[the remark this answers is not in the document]
afshin: Thank you!

afshin:

fcollonval (approved):
Thanks @afshin; it works good on Binder

I left this one opened for Mike to review it as the final version could be included any time (no API changes).

trungleduc (approved):

krassowski:
I updated the `Code changes` section as many code changes in the PR are purely stylistic. I will push a test case for the relevant change.

As for whether we should or should not show the item identical to current token: this is something we could explore indeed; as some auto-insertion logic may depend on it I would not rush to change in before 4.0 though.

packages/completer/src/widget.ts
```diff
@@ -522,34 +522,31 @@ export class Completer extends Widget {
...
-      // Clamp to the length of the list.
-      this._activeIndex = Math.min(
-        Math.max(0, this._activeIndex),
-        items.length - 1
-      );
     }
 
     active = items[this._activeIndex] as HTMLElement;
-    active.classList.add(ACTIVE_CLASS);
+    active?.classList.add(ACTIVE_CLASS);
```
krassowski, line 549: Not sure here: is it be more useful for this to lead to an error in console so that users can report it? If `active` ever gets `null`/`undefined` it would indicate a regression in the logic, and would be more difficult to diagnose if this is silenced, right? Not worth halting this PR on this but just food for thought.

krassowski (approved):
Thank you all! Good to go from my point of view, but I will someone else merge in case if there are comments on my changes.

packages/completer/src/widget.ts
afshin: I think this might not actually be necessary and might have been part of the debugging process of this PR. I'm playing around with it and if I can't generate an error, I'll remove the check. It never existed before.

afshin:

Issue #13848, opened 2023-01-24
Side-by-side rendering in Firefox doesn't show input and output; graphics glitches when dragging resize handle


JasonWeill:
## Description

In Firefox, on tip-of-`master`, when I switch to side-by-side rendering, I see a really distorted view, and as I grab the divider and drag it to the right, I see some additional graphic weirdness until I get about halfway across. See video below.

https://user-images.githubusercontent.com/93281816/214181092-5aaa7642-9577-4914-9067-56063bc85d2c.mov

This occurs even in a private browsing window in Firefox, with extensions disabled. I tried unsuccessfully to reproduce this in Chrome. A teammate was unable to repro with JupyterLab 3.5.3.

## Reproduce

1. Open a notebook in Firefox.
2. Select View > Render Side-by-Side.
3. Observe that only the output column is visible.
4. Click and drag the divider between input and output to the right.
5. Observe that the input column rapidly changes from very small to the expected width, until the resize handle is dragged about halfway across the window.

## Expected behavior

Initial appearance shows input and output columns, each roughly half the width of the main area.

## Context

- Operating System and version: macOS 12.6.2 "Monterey"
- Browser and version: Firefox 102.7.0esr
- JupyterLab version: tip-of-`master` (4.0.0a32)

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

andrii-i:
Reproduced in:

Operating System and version: MacOS 12.6.2
Browser and version: Firefox 102.6.0esr (64-bit)
JupyterLab version: tip-of-master (4.0.0a32)

PR #14479, opened 2023-05-02
Fix side-by-side cell input/output rendering in Firefox

afshin:
This PR throttles side-by-side rendering so that the rendering thread is not flooded with every cursor move event and it fixes the broken rendering of side-by-side input/output for cells in Firefox.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/13848

## Code changes
- Uses a `Throttler` to handle quickly occurring cursor move events.
- Sets the `jp-WindowedPanel-window` class to use the full available width.

## User-facing changes
**Before**
Screenshot 2023-05-02 at 22 51 37

**After**
Screenshot 2023-05-02 at 22 50 43

N/A

### DIFF
--- a/packages/cells/src/resizeHandle.ts
+++ b/packages/cells/src/resizeHandle.ts
@@ -3,8 +3,9 @@
  * Distributed under the terms of the Modified BSD License.
  */
 
-import { Widget } from '@lumino/widgets';
 import { Message } from '@lumino/messaging';
+import { Throttler } from '@lumino/polling';
+import { Widget } from '@lumino/widgets';
 
 const RESIZE_HANDLE_CLASS = 'jp-CellResizeHandle';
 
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @afshin

packages/cells/src/resizeHandle.ts
```diff
@@ -93,4 +74,40 @@ export class ResizeHandle extends Widget {
...
+    super.onBeforeDetach(msg);
+  }
+
+  private _resize(event: MouseEvent): void {
+    // Gate the output size ratio between {0.05, 50} as sensible defaults.
+    const { width, x } = this.targetNode.getBoundingClientRect();
+    const position = event.clientX - x;
+    const ratio = width / position - 1;
+    if (0 < ratio) {
+      const normalized = Math.max(Math.min(Math.abs(ratio), 50), 0.05);
```
krassowski, line 102: Seems that this is more than enough:

limits

krassowski (approved):

packages/cells/src/resizeHandle.ts
krassowski: The grid system has a native [`minmax`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax) function for this use case, but I recall it may require some experimentation to get this right. I was just documenting the current behaviour as of the state of this PR. I believe that we can merge this PR and improve later (would be good to open an issue targeting 4.0.x).

krassowski: (and maybe a change is needed in another part altogether)

krassowski: I opened #14485, merging as-is.

PR #14480, opened 2023-05-02
Updates announcements to better conform to RFC atom standard.

### LINKED ISSUE #14358: Announcements from custom ATOM feed are not displayed in front-end
## Description
I'm looking to show custom announcements in Jupyterlab using a custom ATOM feed.

I start a lab instance (3.6.2) using the command - 

`jupyter lab --LabApp.news_url="<foo>"`

where foo is the url to a xml file containing a ATOM feed generated through [feedgen](https://feedgen.kiesow.be/).

However, I do not see the custom announcements show up in the UI - 

image

My notification settings look like this - 

image

Note - I **do not want** to have official Jupyter news, I'm looking to use the Announcements framework to provide custom announcements. I toggled the options related to official Jupyter news to check if they have any effect and they did not.

## Context

- Browser and version: Chrome 102
- JupyterLab version: 3.6.2

andrewfulton9:
## References

Fixes #14358 

## Code changes

Updates `NewsHandler.get` to better reflect the standards for an atom feed entry defined in [RFC 4287](https://datatracker.ietf.org/doc/html/rfc4287#section-4.1.2). It distinguishes between required and optional entry tags, and adds logic to handle missing optional tags. It also allows for an entry to have multiple link tags, selecting the link with attribute `rel="alternate"`.

I also added tests to account for different valid atom entry schemes.

## User-facing changes

It will be easier for users to link custom atom feeds as it is closer to conforming to the atom standard. Users will be less likely to have notifications be silently dropped due to misaligned entry's

None that I am aware of. 

---
[fcollonval] Auto link issue with action keyword

### DIFF
--- a/docs/source/user/announcements.rst
+++ b/docs/source/user/announcements.rst
@@ -37,7 +37,8 @@ using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``.
 
 .. note::
 
-    An entry in a custom feed is required to have the following tags: ``title``, ``id``, and ``updated``. The following tags are also used but not required: ``published``, ``summary``, and at least one ``link`` with attribute ``rel="alternate"``.
+    An entry in a custom feed is required to have the following tags: ``title``, ``id``, and ``updated``.
+    The following tags are also used but not required: ``published``, ``summary``, and at least one ``link``.
 
 Binder
 ------
--- a/docs/source/user/announcements.rst
+++ b/docs/source/user/announcements.rst
@@ -33,7 +33,7 @@ Jupyter news
 ^^^^^^^^^^^^
 
 The news are fetched from a news feed. The URL can be customized to point to any `Atom feed <https://www.rfc-editor.org/rfc/rfc5023>`_
-using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``.
+using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``. An entry in a custom feed is required to have the following tags: title, id, and updated. The following tags are also used but not required: published, summary, and at least one link.
 
 Binder
 ------
--- a/docs/source/user/announcements.rst
+++ b/docs/source/user/announcements.rst
@@ -33,7 +33,11 @@ Jupyter news
 ^^^^^^^^^^^^
 
 The news are fetched from a news feed. The URL can be customized to point to any `Atom feed <https://www.rfc-editor.org/rfc/rfc5023>`_
-using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``. An entry in a custom feed is required to have the following tags: title, id, and updated. The following tags are also used but not required: published, summary, and at least one link.
+using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``.
+
+.. note::
+
+    An entry in a custom feed is required to have the following tags: ``title``, ``id``, and ``updated``. The following tags are also used but not required: ``published``, ``summary``, and at least one ``link`` with attribute ``rel="alternate"``.
 
 Binder
 ------
--- a/jupyterlab/handlers/announcements.py
+++ b/jupyterlab/handlers/announcements.py
@@ -230,34 +230,36 @@ class NewsHandler(APIHandler):
                 tree = ET.fromstring(response.body)  # noqa S314
 
                 def build_entry(node):
-                    def get_xml_text(attr: str, required: bool = False) -> str:
+                    def get_xml_text(attr: str, default: Optional[str] = None) -> str:
                         node_item = node.find(f"atom:{attr}", xml_namespaces)
                         if node_item is not None:
                             return node_item.text
-                        elif required:
+                        elif default is not None:
+                            return default
+                        else:
                             error_m = (
                                 f'atom feed entry does not contain a required attribute: {attr}'
                             )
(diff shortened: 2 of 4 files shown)


jupyterlab/handlers/announcements.py
```diff
@@ -226,32 +230,54 @@ async def get(self):
...
-                        + node.find("atom:summary", xml_namespaces).text,
-                        createdAt=datetime.fromisoformat(
-                            node.find("atom:published", xml_namespaces).text
-                        ).timestamp()
-                        * 1000,
-                        modifiedAt=datetime.fromisoformat(
-                            node.find("atom:updated", xml_namespaces).text
-                        ).timestamp()
-                        * 1000,
+                    def get_xml_text(attr: str, required: bool = False) -> str:
```
fcollonval, line 233: Could we do a slight change to mimic dictionary `get` method: `get(key: str, default: Optional[str] = None) -> str` if default is `None`, raise otherwise return default.

jupyterlab/handlers/announcements.py
```diff
@@ -226,32 +230,54 @@ async def get(self):
...
+                            raise KeyError(error_m)
+                        else:
+                            return ''
+
+                    entry_title = get_xml_text("title", required=True)
+                    entry_id = get_xml_text("id", required=True)
+                    entry_updated = get_xml_text("updated", required=True)
+                    entry_published = get_xml_text("published")
+                    entry_summary = get_xml_text("summary")
+                    entry_link = node.find("atom:link[@rel='alternate']", xml_namespaces).get(
```
fcollonval, line 250: @suganya-sk is it possible for you to set the attribute `rel="alternate"` so that the link will be picked?

@andrewfulton9  I guess we should protect against no `link[@rel='alternate']` match; could you add a test for that case?

jupyterlab/tests/test_announcements.py
fcollonval, line 1: You :rocket: thanks for adding the tests

jupyterlab/handlers/announcements.py
fcollonval: `link` should be optional - could you update the frontend type there:

https://github.com/jupyterlab/jupyterlab/blob/19050c7b816bb4dd412a4bed94aee7a34050e5f4/packages/apputils-extension/src/announcements.ts#L164

> The frontend code is actually ok as it assumes `link` can be optional but better have correct typing :wink:

fcollonval (changes requested):
Thanks a lot @andrewfulton9 

I left a couple of comments but overall it is really good (thanks for the tests :heart:)

fcollonval:
Would you mind also adding a note in the documentation about what is read from an feed entry at:

https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/user/announcements.rst#L36

packages/apputils-extension/src/announcements.ts
```diff
@@ -161,7 +161,9 @@ export const announcements: JupyterFrontEndPlugin<void> = {
         if ((settings?.get('fetchNews').composite ?? 'false') === 'true') {
           try {
             const response = await requestAPI<{
-              news: (Notification.INotification & { link: [string, string] })[];
+              news: (Notification.INotification & {
+                link: [string, string | undefined];
```
fcollonval, line 165:
```suggestion
                link?: [string, string];
```

packages/apputils-extension/src/announcements.ts
```diff
@@ -220,7 +222,9 @@ export const announcements: JupyterFrontEndPlugin<void> = {
         if ((settings?.get('checkForUpdates').composite as boolean) ?? true) {
           const response = await requestAPI<{
             notification:
-              | (Notification.INotification & { link: [string, string] })
+              | (Notification.INotification & {
+                  link: [string, string | undefined];
```
fcollonval, line 226:
```suggestion
                  link?: [string, string];
```

docs/source/user/announcements.rst
```diff
@@ -33,7 +33,7 @@ Jupyter news
 ^^^^^^^^^^^^
 
 The news are fetched from a news feed. The URL can be customized to point to any `Atom feed <https://www.rfc-editor.org/rfc/rfc5023>`_
-using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``.
+using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``. An entry in a custom feed is required to have the following tags: title, id, and updated. The following tags are also used but not required: published, summary, and at least one link.
```
fcollonval, line 36:
```suggestion
using the command line option ``--LabApp.news_url="<URL_TO_FEED_XML_FILE>"``.

.. note::

    An entry in a custom feed is required to have the following tags: ``title``, ``id``, and ``updated``. The following tags are also used but not required: ``published``, ``summary``, and at least one ``link`` with attribute ``rel="alternate"``.
```

andrewfulton9:
> Thanks for the quick follow up.

No problem! I really appreciate the review. Let me know if there is anything else that I can do for this. I'm not sure why the visual regression tests and the linux examples are failing :thinking:

fcollonval:
They are likely flaky - re running them now.

fcollonval (approved):
Thanks @andrewfulton9 

Failures are not related

fcollonval:

andrewfulton9:
Thanks for opening the backport PR and all the help getting this through quickly, @fcollonval!

Issue #14145, opened 2023-03-07
Contextual help is not properly initialized on initial page load


fcollonval:
Testing on Binder, there is an issue at initial page load; the contextual help is not connected to the notebook:

![contextual-help](https://user-images.githubusercontent.com/8435071/222722415-79e630ee-b9da-49bf-90db-faa21544f6d7.gif)

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/14120#pullrequestreview-1323732862_

PR #14484, opened 2023-05-03
Fix initial load of contextual help

afshin:
This PR resolves an issue where the contextual help widget fails to connect to a console or notebook on initial page load.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/14145

## Code changes
The `inspector-extension` plugins for notebook and console contextual help both check whether there is a relevant active widget after the `app.restored` promise resolves.

## User-facing changes
Initial page load of contextual help works now.

N/A


fcollonval (approved):
Thanks @afshin

Issue #14332, opened 2023-04-06
For notebook v7: tree view should be able to sort notebook ahead of other files.


Carreau:
See upstream issue https://github.com/jupyter/notebook/issues/6819

### Problem

Notebook V6 tree view sorts `folders > notebook > other files`, reason being in notebook users are way less likely to access other files. 

As V7 uses JupyterLab Components notebook get sorted as they are in JupyterLab, and thus we lost having notebooks at the top of the tree view. 

### Proposed Solution

Add an option for the file browser to sort notebooks preferentially. Folder are already on top, so it should not be too hard.

jtpio:
Thanks @Carreau for reporting the issue in the JupyterLab repo.

Adding it to the `4.0` milestone. It's not a blocker for JupyterLab but might be for Notebook 7.

krassowski:
Relevant parts of the codebase:

https://github.com/jupyterlab/jupyterlab/blob/a3dd6af90a011854305b83ff5f64a0f474567e4b/packages/filebrowser/src/listing.ts#L325-L329

https://github.com/jupyterlab/jupyterlab/blob/a3dd6af90a011854305b83ff5f64a0f474567e4b/packages/filebrowser/src/listing.ts#L2771-L2807

JasonWeill:
@Carreau The `sort()` method is public, as @krassowski described above, so we should be able to override it in notebook for 7.0.0, if it blocks Notebook 7.0.0.

PR #14497, opened 2023-05-05
FileBrowser: Added the `sortNotebooksFirst` option

tpatel:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14332

## Code changes

Changes to `filebrowser-extension`:

- New `sortNotebooksFirst` setting
- New menu item shown when right-clicking on the filebrowser header

Changes to `filebrowser`:

- Expose a `sortNotebooksFirst` property storing the setting internally as `_sortNotebooksFirst` that propagates changes to the listing
- The listing exposes a `setNotebooksFirstSorting` method storing the setting internally as `_sortNotebooksFirst`. Modifying it triggers a sorting operation.
- Updated the sorting method to take into account the two sorting behaviors.

## User-facing changes

New setting in the filebrowser header setting menu (which opens when right-clicking on the header).

image

The behavior is:

- When disabled, the file browser sorts the directories above all other files (doesn't modify the current behavior)
- When enabled, the file browsers sorts the directories above everything, then the notebooks above other files (matches the Notebook V6 behavior).

None


packages/filebrowser/src/browser.ts
```diff
@@ -170,6 +170,22 @@ export class FileBrowser extends SidePanel {
...
+  get sortNotebooksFirst(): boolean {
+    return this._sortNotebooksFirst;
+  }
+
+  set sortNotebooksFirst(value: boolean) {
+    if (this.listing.setNotebooksFirstSorting) {
+      this.listing.setNotebooksFirstSorting(value);
+      this._sortNotebooksFirst = value;
+    } else {
+      console.warn('Listing does not support toggling column visibility');
```
krassowski, line 185:
```suggestion
      console.warn('Listing does not support sorting notebooks first');
```

krassowski (commented):
Looking good!

packages/filebrowser/src/browser.ts
tpatel: Good catch, thanks!

packages/filebrowser/src/listing.ts
```diff
@@ -936,6 +940,18 @@ export class DirListing extends Widget {
...
+   * This sorts the items again if the internal value is modified.
+   */
+  setNotebooksFirstSorting(isEnabled: boolean) {
+    let previousValue = this._sortNotebooksFirst;
+    this._sortNotebooksFirst = isEnabled;
+    if (this._sortNotebooksFirst !== previousValue) {
+      this.sort(this._sortState);
+    }
+  }
+
```
fcollonval, line 954: Would you mind switching to a getter-setter? This is useful to have a getter when a setter is defined. So that external code can check what is the current value.

FYI: I also switched the `let` to `const` as it is a good practice to preferred `const` when a variable is expected to not be modified.

```suggestion
  /**
   * Update the setting to sort notebooks above files.
   * This sorts the items again if the internal value is modified.
   */
  get notebooksFirstSorting(): boolean {
    return this._sortNotebooksFirst;
  }
  set notebooksFirstSorting(isEnabled: boolean) {
    const previousValue = this._sortNotebooksFirst;
    this._sortNotebooksFirst = isEnabled;
    if (this._sortNotebooksFirst !== previousValue) {
      this.sort(this._sortState);
    }
  }

```

packages/filebrowser/src/browser.ts
```diff
@@ -170,6 +170,22 @@ export class FileBrowser extends SidePanel {
...
 
+  /**
+   * Whether to sort notebooks above other files
+   */
+  get sortNotebooksFirst(): boolean {
+    return this._sortNotebooksFirst;
+  }
+
+  set sortNotebooksFirst(value: boolean) {
+    if (this.listing.setNotebooksFirstSorting) {
```
fcollonval, line 181: What is the reason for testing for `setNotebooksFirstSorting`?

As the `listing` is of type `DirListing` (no interface used), it will always exist, isn't it?

packages/filebrowser/test/listing.spec.ts
```diff
@@ -737,5 +745,153 @@ describe('filebrowser/listing', () => {
         });
       });
     });
+
+    describe('should sort correctly', () => {
```
fcollonval, line 749: :heart:

fcollonval (changes requested):
Thanks @madx 

I have some questions. But generally this looks good, thanks!

packages/filebrowser/src/listing.ts
tpatel: I've used a method as I've copied the pattern from the column visibility. I do prefer the setter/getter though, thanks for the suggestion!

https://github.com/jupyterlab/jupyterlab/blob/d47d50bbeb2db80df09489eb7c9588c29a140d22/packages/filebrowser/src/listing.ts#L925C3-L941

packages/filebrowser/src/browser.ts
tpatel: I couldn't find why the `setColumnVisibility` was being checked. I believe it should always exist but I was a bit confused and went on the safe side. Do you have more info about this?

https://github.com/jupyterlab/jupyterlab/blob/d47d50bbeb2db80df09489eb7c9588c29a140d22/packages/filebrowser/src/browser.ts#L164C7-L171

krassowski: Given how downstream swaps `listing`, maybe it should become an interface? (this is not a discussion for this PR though). I just wanted to highlight that unfold extension goes [pretty far with the modifications](https://github.com/jupyterlab-contrib/jupyterlab-unfold/blob/6909f91e05d5bf378f076438c71264840d5e47d9/src/unfold.ts ) and depending on the version of `DirListing` it would load the guard may actually be useful (well at least if this lands in 4.x rather than 4.0). We also wanted to modify listing in jupyterlab-git but can probably make dedicated APIs for that use case.

jtpio: Probably this condition was indeed added to make it easier for downstream applications to customize the file browser.

Thanks for providing the extra context, agree we can track follow-ups in other issues :+1:

jtpio:
It would be great to get this one in so it makes it to the final 4.0 release, and would directly benefit Notebook 7.

Leaving it open until tomorrow in case @krassowski or @fcollonval have any more feedback.

packages/filebrowser/src/browser.ts
fcollonval: Ok then I'll agree we should keep the safe guard test. But I still would like to switch to a getter/setter pattern. @tpatel, you could replace the safeguard condition by

```ts
if (this.notebooksFirstSorted !== undefined)
```

jtpio: > switch to a getter/setter pattern

Would it be possible to address this in a separate PR? And address this for `setColumnVisibility` as well for consistency?

jtpio (approved):
Thanks!

As discussed during the [JupyterLab weekly meeting today](https://github.com/jupyterlab/team-compass/issues/170#issuecomment-1551816414), it would be great to get this PR in so it can be used in Notebook 7.

This can be seen as a new feature to be included in a 4.1 release. But this can also be considered a "bug fix" for feature parity with the classic notebook.

Also having it opt-in like it is the case in this PR makes it easier to ship in a `4.0.x` release.

packages/filebrowser/src/browser.ts
tpatel: I think it makes sense to address moving both methods to the getter/setter pattern in a separate PR to avoid being in a state with two different patterns 👍

PR #14498, opened 2023-05-05
Fix yarn deduplication and check

fcollonval:
## References

Started from a silent error seen when working on jupyter-collaboration about the `yarn check` call that does not exist on yarn 3
And `yarn-deduplicate` only supporting yarn 1: https://github.com/scinos/yarn-deduplicate#readme

It leads also to discovering we are not checking for deduplication in yarn lock on the CI.

## Code changes

Replace `yarn-deduplicate` by `yarn-berry-deduplicate` command; we don't use stock `yarn dedupe` because it has only one strategy _highest_ that does not look for an optimum compare to `yarn-berry-deduplicate`.

Add check on CI for deduplication

## User-facing changes

No silent failure on `yarn check`

None

---

First commit is expected to fail lumino packages need deduplication


fcollonval:
:sparkles: first commit failed on integrity test as expected.

dev_mode/package.json
```diff
@@ -207,7 +207,7 @@
     "webpack-merge": "^5.8.0",
     "whatwg-fetch": "^3.0.0",
     "worker-loader": "^3.0.2",
-    "yarn-deduplicate": "^5.0.0"
+    "yarn-berry-deduplicate": "^2.0.0"
```
brichet, line 210: Do we need that dependency ?
It is always used with `dlx` command, which seems to download the package in a temporary environment.

brichet (approved):
Thanks @fcollonval.

Look good to me, even if I don't really know how to reproduce the silent failure you mentioned.

Should we update the extension migration to use `yarn-berry-deduplicate`:
https://github.com/jupyterlab/jupyterlab/blob/0e6c0a624d6315b77cc550a806ad8fcc08b72bc4/docs/source/extension/extension_migration.rst?plain=1#L89

dev_mode/package.json
fcollonval: Good point - and indeed then it will be good to update the migration guide - I did not because I was trying to avoid adding a tool for all extensions.

fcollonval:
> Look good to me, even if I don't really know how to reproduce the silent failure you mentioned.

You need to start JupyterLab in debug mode to seen it triggers the function of jupyterlab/commands.py L211.

afshin (approved):
Thanks @fcollonval!

PR #14505, opened 2023-05-07
Fix kernel completion provider assumptions

### LINKED ISSUE #14500: Tab completion not working on 4.0.x with IRKernel
## Description

tab completion is not working with the R kernel, the kernel itself seems to be working fine and pressing shift+tab on a function shows the help doc as expected. its just the tab completion that doesn't work.

also typing `plot(<tab>`, offers to complete `plot` again instead of the function arguments, that could be a clue of whats wrong.

I tried to use the kernelspy extension to see whats happening but I couldn't get it to work, probably due to version constraints unrelated to this.

## Reproduce

1. installed on a new env: `mamba create -n jl4 -c conda-forge/label/jupyterlab_rc jupyterlab`
2. already had IRKernel package installed on system R (works fine with 3.6.x)
3. open a R notebook and type `pl<tab>` 

## Expected behavior

show list of functions that starts with `pl`

krassowski:
## References

Fixes #14500

## Code changes

- Checks if `response.metadata._jupyter_types_experimental` is defined
- Adds a test for `KernelCompleterProvider.fetch`

## User-facing changes

Kernel completion suggestions in IRkernel work again.

None

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -79,6 +79,7 @@ module.exports = {
           '__webpack_public_path__',
           '__webpack_share_scopes__',
           '__webpack_init_sharing__',
+          '_jupyter_types_experimental',
           'allow_stdin',
           'allowed_extensions',
           'allowed_extensions_uris',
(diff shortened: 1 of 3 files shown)


afshin (approved):
Thank you!

PR #14514, opened 2023-05-09
Fix some usages of the translation object

fcollonval:
## Code changes

Fix some improper translation domain name or syntax to get the string extracted.

## User-facing changes

More strings should be translated.

None


afshin (approved):
Thank you!

PR #14519, opened 2023-05-10
Fix drive handling for sessions and terminals

### LINKED ISSUE #14443: Wrong kernel working directory in collaborative mode
## Description

Opening a notebook in collaborative mode doesn't run the kernel in the notebook directory.

## Reproduce

1. `pip install jupyterlab==4.0.0b2`
2. `pip install jupyter_collaboration`
3. `mkdir examples`
4. `jupyter lab`
5. Open or create a notebook in `examples`
6. In the notebook, run:
```py
import os

os.getcwd()
```
7. See that the current working directory is not `./examples`, but the directory where JupyterLab was launched.

## Expected behavior

The kernel current working directory should be the directory of the notebook.

## Context

- Operating System and version: Linux Ubuntu 22.04 
- Browser and version: Google Chrome
- JupyterLab version: 4.0.0b2
- jupyter_collaboration version: 1.0.0a8

fcollonval:
## References

Fixes #14443

## Code changes

Set `sessionContext` has `localPath` (dropping the drive name)
Pass `cwd` without the drive name when starting terminals

## User-facing changes

For browser with non-default drive (like for RTC), the kernels should start in the file folder and the terminals and consoles should open in the current directory.

None

### DIFF
--- a/packages/terminal-extension/src/index.ts
+++ b/packages/terminal-extension/src/index.ts
@@ -353,6 +353,7 @@ function addCommands(
     execute: async args => {
       const name = args['name'] as string;
       const cwd = args['cwd'] as string;
+      const localPath = serviceManager.contents.localPath(cwd);
 
       let session;
       if (name) {
@@ -364,12 +365,15 @@ function addCommands(
         } else {
           // we are restoring a terminal widget but the corresponding terminal was closed
           // let's start a new terminal with the original name
-          session = await serviceManager.terminals.startNew({ name, cwd });
+          session = await serviceManager.terminals.startNew({
+            name,
+            cwd: localPath
+          });
         }
       } else {
         // we are creating a new terminal widget with a new terminal
         // let the server choose the terminal name
-        session = await serviceManager.terminals.startNew({ cwd });
+        session = await serviceManager.terminals.startNew({ cwd: localPath });
       }
 
       const term = new XTerm(session, options, translator);
--- a/packages/console/src/panel.ts
+++ b/packages/console/src/panel.ts
@@ -8,7 +8,7 @@ import {
   SessionContextDialogs
 } from '@jupyterlab/apputils';
 import { IEditorMimeTypeService } from '@jupyterlab/codeeditor';
-import { PathExt, Time, URLExt } from '@jupyterlab/coreutils';
+import { PathExt, Time } from '@jupyterlab/coreutils';
 import {
   IRenderMimeRegistry,
   RenderMimeRegistry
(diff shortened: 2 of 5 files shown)


krassowski:
The console tests appear relevant.

packages/terminal-extension/src/index.ts
```diff
@@ -353,6 +353,7 @@ function addCommands(
     execute: async args => {
       const name = args['name'] as string;
       const cwd = args['cwd'] as string;
+      const localPath = serviceManager.contents.localPath(cwd);
```
hbcarlos, line 356: The UI tests are also relevant.

The problem is that `cwd` can be undefined. For example, when creating a new terminal from the File -> New -> Terminal menu, the `cwd` is undefined.

hbcarlos:
The remaining test failures are not related.

krassowski (approved):
Thank you!

jtpio:
> For browser with non-default drive (like for RTC), the kernels should start in the file folder and the terminals and consoles should open in the current directory.

What does "file folder" means here? What should then be the expected behavior in the case when multiple drives with possibilly the same file names (`example.ipynb` in both the default filebrowser and another drive) are used simultaneously?

With 3.6.x it was possible to get some information about this since the drive name was kept as the session context prefix, and visible in the tooltip:

image

But with JupyterLab 4 it's not the case anymore:

image

Which makes it difficult to know about which kernels are being used, especially when there are files with the same name in two different drives:

image

jtpio:
Raising this question because it looks like this change (released in JupyterLab 4.0) affected JupyterLite downstream: https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/55#issuecomment-1772879905

The drive name used to be passed along as a prefix (for example `FileSystemAccess:`) which means the server was aware of the drive currently in use, which could then be passed to the kernel on startup. This allowed for accessing files from other drives from within a JupyterLite kernel in a seamless way.

Looking more into it, JupyterLab does not even seem to allow this anymore, since the information about the drive being used in never sent to the server, now that the session context path defaults to the local path (without the drive name).

Potentially something to look into as part of https://github.com/jupyterlab/jupyterlab/issues/15143, to allow for such uses case like in JupyterLite.

jtpio:
Looks like https://github.com/jupyterlab/jupyterlab/pull/13866 might also be related.

echarles:
> What does "file folder" means here? What should then be the expected behavior in the case when multiple drives with possibilly the same file names (example.ipynb in both the default filebrowser and another drive) are used simultaneously?

I have similar questions as @jtpio.

Overall, my question is what is the strategy/logic of the drives (registration, usage, naming convention, URI...) across JupyterLab and its extensions.

This a broader question than the issue fixed by this PR, and this may be already defined somewhere (?), just asking...

jtpio:
> This is a broader question than the issue fixed by this PR, and this may be already defined somewhere (?), just asking...

Yes very likely a broader question that would involve more design work on the drive feature. I opened https://github.com/jupyterlab/jupyterlab/issues/15289 to discuss this more in an issue.

Issue #14485, opened 2023-05-03
Side-by-side should have better limits for min/max width


krassowski:
## Description

Notebook side-by-side mode should have better min/max limits for resizing. Once #14479 gets merged it will still allow to cross the execution counter (`[ ]:` in gif below):

## Reproduce

![limits](https://user-images.githubusercontent.com/5832902/235912594-c9f28d08-af21-4026-8963-16f7aa816f23.gif)

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/14479#discussion_r1183607083_

## Expected behavior

We either prevent resize beyond the state in which editor is unusable, or keep allowing it but hide the execution indicator?

## Context

- JupyterLab version: 4.0.0rc0 + #14479

dharmaquark:
I can help take a pass at adding minmax logic if no one else has picked this up!

krassowski:
Please feel welcome to, and feel free to tag me for review once PR is open :)

PR #14529, opened 2023-05-13
Add min width to constrain resizing in side-by-side view

dharmaquark:
## References

https://github.com/jupyterlab/jupyterlab/issues/14485

## Code changes

I've taken an initial pass at setting a non-zero minimum value in the `minmax` css function for the side-by-side resized cell

## User-facing changes

From trial and error, and at multiple zoom levels, this looks to enforce a minimum width for the cell that prevents 'negative' width that prevents the cell from going left past the execution indicator, and dynamically hides the active cell toolbar when the output starts to overlap:

https://github.com/jupyterlab/jupyterlab/assets/14363644/06b8897b-e0e1-455d-94a1-c9f2f3054da5

I believe this is backwards-compatible!


dharmaquark:
@afshin and @krassowski for awareness!

packages/notebook/style/base.css
```diff
@@ -329,8 +329,8 @@
 }
 
 .jp-mod-sideBySide.jp-Notebook .jp-CodeCell.jp-mod-resizedCell {
-  grid-template-columns: minmax(0, 1fr) min-content minmax(
-      0,
+  grid-template-columns: minmax(70px, 1fr) min-content minmax(
```
dharmaquark, line 332: a quick note- I haven't pushed this upstream yet, but I also had to set this 70px for `jp-CodeCell` minMax (otherwise adjacent cells reacting to the min/max don't respect the minimum value set in `for jp-mod-resizedCell`)

dharmaquark:
another note: this PR only introduces a limit on the minimum size, I'm still working through trial and error to set a maximum cell width that doesn't cause overflow into the right cell menu options and/or cause cell output wrapping. After sleeping on it, the next avenue I'm going to dig into is setting `max-content` explicitly, using `minmax` similar to how the css here is setting `min-content`. I'm going to run through some of the syntax [options]( https://developer.mozilla.org/en-US/docs/Web/CSS/minmax#syntax) to try to get a better understanding of what minmax is doing to the columns in this particular notebook case under the hood (especially with respect to cell width versus output width)

dharmaquark:
on further investigation, the correct way to handle the max width case seems to be to hook into the existing logic that hides and shows the toolbar depending on the cell and output width: https://github.com/jupyterlab/jupyterlab/blob/25e52500908e3237006f5f9dc7588ae68b1927e9/packages/cell-toolbar/src/celltoolbartracker.ts#L245

To do that, I added a resize signal to `ResizeHandle` that then triggers a `displayChanged` signal from the cell widget (happy to explore variations on this or alternative approaches if this seems too kludgy!) Will update screen recording in PR description to show current sizing behavior with this changeset

dharmaquark:
I think the UI tests failures could be related to these changes (screenshots not matching? but not 100% sure - was hoping these changes would only affect display of cell toolbar in side-by-side mode). Not 100% sure of how to resolve the check release issue (related to versioning?)

krassowski (approved):
Thank you @dharmaquark!

krassowski:

Issue #6579, opened 2019-06-14
Modal dialogs are not accessible


clapierre:
From web4all hackathon https://github.com/diagram-codesprint/jupyterlab/issues/3
**Describe the bug**
Modal dialogs are not accessible. When a modal dialog is displayed and a screen reader is used, the web page is visible to the screen reader (only the modal dialog should be).

**To Reproduce**

1. Launch Jupyter
2. Select Kernel > Restart Kernel... (to bring a sample modal dialog)

**Expected behavior**
The modal dialog should appear as modal to a screen reader

@sinabahram
Initial fix is to add aria-modal="true" and role="dialog" on the containing element for the modal. Because modals are created/destroyed in the DOM upon usage, this should work fine. We then need to examine accessible ways of closing hte modal. There must be an accessible close button, and the escape key needs to also trigger this close button. Lastly, we need to make sure that focus is returned to where it originally was when the modal is closed. This means that jupyterlab needs to remember the last focussed element, and then upon modal close, this element needs to regain focus e.g. by calling .focus() on the reference to that element.

tonyfast:
this relates to focus issues in #9399 

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

jasongrout:
Perhaps the new dialog HTML element is a natural way to gain accessibility and other desirable behaviors?

* https://www.stefanjudis.com/blog/a-look-at-the-dialog-elements-super-powers/
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

tonyfast:
`dialog` is most likely the best semantics for the container. its probably worth referencing the WAI-ARIA pattern that describes the expectations for a dialog modal with assistive technology. 

* https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

krassowski:
Closing as resolved by #14560.

PR #14560, opened 2023-05-17
Implement dialog element for dialogs with accessibility

t03857785:
## References
#6579 Modal dialogs are not accessible

## Code changes

Change the dialog to use the dialog html element rather than the div element
Add aria title to the dialog
Add aria modal to the dialog to control focus for accessibility tools using aria
Add aria title to buttons and implement where required (this may be expanded upon in the future)

## User-facing changes

Users using accessibility tools will now have dialog labels announced and more meaningful titles for some dialog buttons 
Other users will not see any changes

Dialog buttons now have an optional ariaLabel that can be used as well as label if the label is not suitable for users with accessibility needs

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -1338,8 +1338,7 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
     }
     const buttons = [
       Dialog.cancelButton({
-        label,
-        ariaLabel: trans.__('Cancel Kernel Select')
+        label
       }),
       Dialog.okButton({
         label: trans.__('Select'),
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -1337,8 +1337,14 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
       label = sessionContext.kernelDisplayName;
     }
     const buttons = [
-      Dialog.cancelButton({ label }),
-      Dialog.okButton({ label: trans.__('Select') })
+      Dialog.cancelButton({
+        label,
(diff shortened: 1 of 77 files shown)


packages/apputils/src/sessioncontext.tsx
```diff
@@ -1410,14 +1416,20 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
       throw new Error('No kernel to restart');
     }
 
-    const restartBtn = Dialog.warnButton({ label: 'Restart' });
+    const restartBtn = Dialog.warnButton({
+      label: 'Restart',
```
krassowski, line 1420: It seems it was missing previously, let's add it in:

```suggestion
      label: trans.__('Restart'),
```

t03857785:
This was tested on windows using the built in Narrator

krassowski:
The visual test failures are relevant - tests need to be updated by adding [`exact: true`](https://playwright.dev/docs/api/class-locator#locator-get-by-role-option-exact) to the options of affected `getByRole` calls.

<details>

Example issue:

```
Error: locator.click: Error: strict mode violation: locator('.jp-Dialog').getByRole('button', { name: 'Select' }) resolved to 2 elements:
    1) <button class="jp-Dialog-button jp-mod-reject jp-mod-sty…>…</button> aka getByRole('button', { name: 'Cancel Kernel Select' })
    2) <button class="jp-Dialog-button jp-mod-accept jp-mod-sty…>…</button> aka getByRole('button', { name: 'Select Kernel' })
```

</details>

t03857785:
@krassowski I think all of the required/requested changes have been made I have also on reflection reverted the No Kernel Aria Label to fix tests as it may not be accurate in all situations

krassowski (approved):
Looks good to me, thank you @t03857785.

PR #14566, opened 2023-05-19
Fix react error log in shortcuts dialog

### LINKED ISSUE #14337: Shortcuts dialog logs a React Error in the dev tools console on open
## Description

On the latest commit (a94615522d5d9a199387d2874b894d57df438cf2) opening the Shortcuts dialog gives the following error in the dev tools console:

```
Warning: Each child in a list should have a unique "key" prop.
```

## Reproduce

https://user-images.githubusercontent.com/591645/230618846-3482da4d-f42b-4728-89da-8ed13f2650fa.mp4

## Expected behavior

There shouldn't be any error in the dev tools console.

## Context

This was noticed while updating Notebook 7 to JupyterLab 4.0.0b1 in https://github.com/jupyter/notebook/pull/6838

c3Vu:
## References

Fixes #14337 

## Code changes

Add 'key' props to the react components so that error logs are avoid.

## User-facing changes

None.

 None.

### DIFF
--- a/packages/apputils-extension/src/shortcuts.tsx
+++ b/packages/apputils-extension/src/shortcuts.tsx
@@ -62,19 +62,22 @@ export function displayShortcuts(options: IOptions) {
 
   function formatKeys(keys: string[]): JSX.Element {
     const topContainer: JSX.Element[] = [];
-    for (const key of keys) {
+    keys.forEach((key, index) => {
       const container: JSX.Element[] = [];
       for (const ch of key.split(' ')) {
         container.push(
-          <span className={SHORTCUT_KEY_CLASS}>
+          <span className={SHORTCUT_KEY_CLASS} key={ch}>
             <kbd>{ch}</kbd>
           </span>,
-          <> + </>
+          <React.Fragment key={`${ch}-fragment`}> + </React.Fragment>
         );
       }
       /*topContainer.push(<span>{container.slice(0, -1)}</span>, <>,</>);*/
-      topContainer.push(<span>{container.slice(0, -1)}</span>, <> + </>);
-    }
+      topContainer.push(
+        <span key={index}>{container.slice(0, -1)}</span>,
+        <> + </>
+      );
+    });
     return <span>{topContainer.slice(0, -1)}</span>;
   }
 
@@ -161,7 +164,10 @@ export function displayShortcuts(options: IOptions) {
     if (groupedBindings.has(d)) {
       bindingTable.push(
         groupedBindings.get(d)!.map(b => (
-          <tr className={SHORTCUT_TABLE_ROW_CLASS} key={b.command}>
+          <tr
+            className={SHORTCUT_TABLE_ROW_CLASS}
+            key={`${b.command}-${b.keys.join('-')}`}
+          >
             <td className={SHORTCUT_TABLE_ITEM_CLASS}>{formatLabel(b)}</td>
             <td className={SHORTCUT_TABLE_ITEM_CLASS}>
               {formatKeys([...b.keys])}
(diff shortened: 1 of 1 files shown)


packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -62,19 +62,22 @@ export function displayShortcuts(options: IOptions) {
...
         container.push(
-          <span className={SHORTCUT_KEY_CLASS}>
+          <span className={SHORTCUT_KEY_CLASS} key={ch}>
             <kbd>{ch}</kbd>
           </span>,
-          <> + </>
+          <React.Fragment key={`${ch}-fragment`}> + </React.Fragment>
         );
       }
       /*topContainer.push(<span>{container.slice(0, -1)}</span>, <>,</>);*/
```
krassowski, line 75: I know it was not added in this PR, but let's remove the old comment too (as if it was to be un-commented in the future, it would have the same issue).

```suggestion
```

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -62,19 +62,21 @@ export function displayShortcuts(options: IOptions) {
 
   function formatKeys(keys: string[]): JSX.Element {
     const topContainer: JSX.Element[] = [];
-    for (const key of keys) {
+    keys.forEach((key, index) => {
       const container: JSX.Element[] = [];
       for (const ch of key.split(' ')) {
         container.push(
-          <span className={SHORTCUT_KEY_CLASS}>
+          <span className={SHORTCUT_KEY_CLASS} key={ch}>
```
krassowski, line 69: `ch` is not a unique key because we have keybindings composed of pressing the same key twice in short succession, e.g. delete cell is <kbd>D</kbd> + <kbd>D</kbd>:

Screenshot from 2023-05-20 10-17-47

maybe another `forEach` here?

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -62,19 +62,23 @@ export function displayShortcuts(options: IOptions) {
...
-          <> + </>
+          <React.Fragment key={`fragment-${chIndex}`}> + </React.Fragment>
         );
-      }
-      /*topContainer.push(<span>{container.slice(0, -1)}</span>, <>,</>);*/
-      topContainer.push(<span>{container.slice(0, -1)}</span>, <> + </>);
-    }
+      });
+      topContainer.push(
+        <span key={index} className={`key-${index}`}>
```
krassowski, line 76: Did you mean:

```suggestion
        <span key={`key-${index}`}>
```

(may need linting)

packages/apputils-extension/src/shortcuts.tsx
```diff
@@ -161,15 +163,23 @@ export function displayShortcuts(options: IOptions) {
     if (groupedBindings.has(d)) {
       bindingTable.push(
         groupedBindings.get(d)!.map(b => (
-          <tr className={SHORTCUT_TABLE_ROW_CLASS} key={b.command}>
+          <tr
+            className={SHORTCUT_TABLE_ROW_CLASS}
+            key={`${b.command}-${b.keys.join('-')}`}
```
krassowski, line 168: I got a duplicate key warning when testing it locally with two shortcuts for the same command if both start with a keystroke with modifier, such as `Ctrl X` and `Ctrl Y`. This is because there will be a space which will terminate the key prematurely (both would then have a key `plugin:command-Ctrl`). This should solve the issue:

```suggestion
            key={`${b.command}-${b.keys.join('-').replace(' ', '_')}`}
```

krassowski (approved):
Thank you @c3Vu!

PR #14575, opened 2023-05-22
Add skip links to all regions 9688

s596757:
This PR addresses ticket number 9688

## User-facing changes

I have added the ability to add aria-labels to buttons to allow them to be picked up screenreader
i have also added in three aditional skiplinks
I have enabled the skiplinks so now when you tab in they display in the top centre of the page.

If you navigate page via tab key when narrator/screen-reader is on, the skiplinks cause the reader to skip to different areas of the page. If you navigate with down key, all elements of the page are read aloud.

I am not sure if the skiplinks I have added in are enough or whether you require more. Please can someone advice me on this. Thank you


packages/apputils/src/mainareawidget.ts
```diff
@@ -44,10 +44,10 @@ export class MainAreaWidget<T extends Widget = Widget>
     const trans = (options.translator || nullTranslator).load('jupyterlab');
     const content = (this._content = options.content);
     content.node.setAttribute('role', 'region');
-    content.node.setAttribute('aria-label', trans.__('notebook content'));
+    content.node.setAttribute('aria-label', trans.__('launcher content'));
```
gabalafou, line 47: I don't think this is always the Launcher just as it's not always a notebook. 

I think there may be an issue about this somewhere. The aria-label probably needs to be set programmatically when a new MainAreaWidget is instantiated. 

Perhaps as a stop-gap measure, we could set these labels to:

- `main panel content`
- `main panel toolbar`

Issue #13678, opened 2022-12-29
Switching to markdown preview is slow; the scroll position gets lost


krassowski:
## Description

Switching tabs to markdown viewer is slow. It takes ~11 partially presented frames and `renderMarkdown` gets called twice.

## Reproduce

1. Open JupyterLab README.md with right click: "Open With" → "Markdown Preview"
2. Open a new launcher
3. Switch back and forth
4. See that the markdown tab contents flickers

![Screenshot from 2022-12-29 13-51-24](https://user-images.githubusercontent.com/5832902/209963460-13fc3107-91a1-4d0c-b1b9-2211405c35f8.png)

![Screenshot from 2022-12-29 13-56-06](https://user-images.githubusercontent.com/5832902/209963513-4094b892-9502-48cc-8af8-914d4e04026e.png)

For scroll position not being preserved see:

![scroll-position-md](https://user-images.githubusercontent.com/5832902/209964136-70a0b76a-08ff-44b4-b287-31f88503bbd6.gif)

## Expected behavior

- We should not need to invalidate the rendered markdown when switching back and forth and definitely we should not call `renderMarkdown` twice.
- The scroll position should be preserved.

## Context

- JupyterLab version: 4.0.0a31 and 3.x

jtpio:
Adding to `4.0.x` to have it on the radar. Also this does seem to be the case in both JupyterLab 4 and 3.x.

This is particularly annoying when editing a markdown file with the preview open on the side:

https://user-images.githubusercontent.com/591645/228195244-9b6b7c67-175f-4be0-a3ee-f4950c6cf05f.mp4

xuezhongyu01:
Hi, I'd like to investigate the remaining performance aspect of this issue on the current `main` branch. I see that #14579 already alleviated the DOM-clearing problem and #18959 is addressing scroll synchronization and preservation. I'll focus on checking whether switching tabs still causes flickering or duplicate `renderMarkdown` calls, and avoid overlapping with #18959.

PR #14579, opened 2023-05-23
Avoid clearing the host node while rendering Markdown

### LINKED ISSUE #13734: blink: Open the preview-markdown  and edit-markdown  tab at the same page
## Description
Open the preview-markdown  and edit-markdown  tab at the same page

## Reproduce

1、create a markdown file
2、source code:
```
### a
sdsd

a fdsdsd
sdsdsdsadsadasd

sdsd

sdsd
11
sdsdsdasdsad
sdsdsdsdsd
sd
a

sdsdasdsad

### sddsd
    sdsds
    
    
asdas
```
3、As long as you use code syntax, it will cause flicker.
Use the syntax of the following link.
https://github.github.com/gfm/#example-3

## Context

- Operating System and version: 
- Browser and version: 
- JupyterLab version: 
max os 12.2
Chrome 108
JupyterLab v3.4.7

1、After editing, the code is executed
marked.marked()
2、debug in highlight
```
        setTimeout(() => {
          highlight(token.text, token.lang, function(err, code) {
            if (err) {
              return done(err);
            }
            if (code != null && code !== token.text) {
              token.text = code;
              token.escaped = true;
            }

            pending--;
            if (pending === 0) {
              done();
            }
          });
        }, 0);
```
3、after setTimeout,the page will blink

<pre>
</pre>
</details>

<pre>
</pre>
</details>

<pre>

</pre>
</details>

c3Vu:
## References

Fixes #13734, alleviates #13678 (scroll is not lost when editing with preview side-by-side but is still lost when switching tabs)

## Code changes

Before rendering the markdown, avoid clearing the host node so that the scroll position would not be lost. Markdown rendering took two steps before (1. empty the DOM; 2. render the DOM). And now the step 1 is skipped in markdown rendering. 

## User-facing changes

1. When editing markdown file keeps the content and the scroll position in the last render.
2. Rendering markdown would take a while and the content would be outdated before the render is complete.
3. This change would affect users who use markdownRendererFactory or RenderedMarkdown

https://github.com/jupyterlab/jupyterlab/assets/71494005/8efed9e4-c457-4d01-9aab-5fe954f93f31

### DIFF
--- a/packages/rendermime/src/widgets.ts
+++ b/packages/rendermime/src/widgets.ts
@@ -76,12 +76,12 @@ export abstract class RenderedCommon
    *
    * @param model - The mime model to render.
    *
-   * @param clearExisting - whether to clear existing nodes before render
+   * @param clearExisting - Whether to clear existing nodes before render.
    *
    * @returns A promise which resolves when rendering is complete.
    *
    * #### Notes
-   * If the DOM node for this widget already has content, it is emptied
+   * By default, if the DOM node for this widget already has content, it is emptied
    * before rendering. Subclasses that do not want this behavior
    * (if, for instance, they are using DOM diffing), should override
    * this method and not call `super.renderModel()`.
--- a/packages/rendermime/src/widgets.ts
+++ b/packages/rendermime/src/widgets.ts
@@ -79,15 +79,21 @@ export abstract class RenderedCommon
    * @param clearExisting - whether to clear existing nodes before render
    *
    * @returns A promise which resolves when rendering is complete.
+   *
+   * #### Notes
+   * If the DOM node for this widget already has content, it is emptied
+   * before rendering. Subclasses that do not want this behavior
+   * (if, for instance, they are using DOM diffing), should override
+   * this method and not call `super.renderModel()`.
    */
-  async renderCommonModel(
+  async renderModel(
     model: IRenderMime.IMimeModel,
-    clearExisting = true
+    clearExisting?: boolean
   ): Promise<void> {
     // TODO compare model against old model for early bail?
 
     // Empty any existing content in the node from previous renders
-    if (clearExisting) {
+    if (clearExisting !== false) {
       while (this.node.firstChild) {
         this.node.removeChild(this.node.firstChild);
       }
@@ -104,23 +110,6 @@ export abstract class RenderedCommon
     if (fragment) {
       this.setFragment(fragment as string);
     }
-  }
-
-  /**
-   * Render a mime model.
-   *
-   * @param model - The mime model to render.
-   *
-   * @returns A promise which resolves when rendering is complete.
-   *
-   * #### Notes
-   * If the DOM node for this widget already has content, it is emptied
-   * before rendering. Subclasses that do not want this behavior
-   * (if, for instance, they are using DOM diffing), should override
-   * this method and not call `super.renderModel()`.
-   */
-  async renderModel(model: IRenderMime.IMimeModel): Promise<void> {
-    await this.renderCommonModel(model);
   }
 
   /**
@@ -339,7 +328,7 @@ export class RenderedMarkdown extends RenderedHTMLCommon {
    * @returns A promise which resolves when rendering is complete.
    */
   async renderModel(model: IRenderMime.IMimeModel): Promise<void> {
-    await this.renderCommonModel(model, false);
+    await super.renderModel(model, false);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/rendermime/src/widgets.ts
```diff
@@ -313,6 +313,20 @@ export class RenderedMarkdown extends RenderedHTMLCommon {
...
+    this.toggleClass('jp-mod-trusted', model.trusted);
+
+    // Render the actual content.
+    await this.render(model);
+
+    // Handle the fragment identifier if given.
+    const { fragment } = model.metadata;
+    if (fragment) {
+      this.setFragment(fragment as string);
+    }
```
krassowski, line 327: Thank you, this solution looks good to me! On implementation level, what if we added a second argument called `clear=true` to `RenderedCommon.renderModel` and then in here call `super.renderModel(model, false)` to avoid duplicating code (and incidental drift in the future)?

c3Vu: Thanks for your advice. `RenderedCommon.renderModel` is an implementation of `IRenderMime.IRenderer`, which is not responsible for providing such options. Instead, I add another method with the `clearExising` option in `RenderedCommon`.

krassowski: True, but we can add optional arguments at implementation level without modifying the interface. I am fine with the proposed solution, but let's make `renderCommonModel` protected.

c3Vu: That makes sense. I thought an optional arguments would break the lint rule. I just tried it and it should look cleaner now

packages/rendermime/src/widgets.ts
```diff
@@ -76,20 +76,27 @@ export abstract class RenderedCommon
...
    *
    * #### Notes
-   * If the DOM node for this widget already has content, it is emptied
-   * before rendering. Subclasses that do not want this behavior
-   * (if, for instance, they are using DOM diffing), should override
-   * this method and not call `super.renderModel()`.
+   * By default, if the DOM node for this widget already has content, it
+   * is emptied before rendering. Subclasses that do not want this behavior
+   * (if, for instance, they are using DOM diffing), should override this
+   * method or call `super.renderModel(model, false)`.
```
afshin, line 87:
```suggestion
   * method or call `super.renderModel(model, true)`.
```

packages/rendermime/src/widgets.ts
```diff
@@ -76,20 +76,27 @@ export abstract class RenderedCommon
...
    */
-  async renderModel(model: IRenderMime.IMimeModel): Promise<void> {
+  async renderModel(
+    model: IRenderMime.IMimeModel,
+    clearExisting?: boolean
```
afshin, line 91:
```suggestion
    keepExisting?: boolean
```

packages/rendermime/src/widgets.ts
```diff
@@ -313,6 +320,17 @@ export class RenderedMarkdown extends RenderedHTMLCommon {
...
 
+  /**
+   * Render a mime model.
+   *
+   * @param model - The mime model to render.
+   *
+   * @returns A promise which resolves when rendering is complete.
+   */
+  async renderModel(model: IRenderMime.IMimeModel): Promise<void> {
+    await super.renderModel(model, false);
```
afshin, line 331:
```suggestion
    await super.renderModel(model, true);
```

afshin (approved):
Thanks for this fix, it is great!

I propose flipping the boolean so that if you want to override the default behavior you send in a `true` to keep existing rendered content.

packages/rendermime/src/widgets.ts
```diff
@@ -76,20 +76,27 @@ export abstract class RenderedCommon
    *
    * @param model - The mime model to render.
    *
+   * @param clearExisting - Whether to clear existing nodes before render.
```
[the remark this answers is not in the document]
c3Vu: Thanks for your suggestion, it has been modified.

krassowski (approved):
Thank you @c3Vu!

I tested the latest version locally and noted that while it does fix blinking and the annoying scrolling when editing markdown with preview, it does not solve scrolling when switching tabs (at least not in the browsers I tested) - I updated the PR description to reflect this.

FoSuCloud:
> Thank you @c3Vu!
> 
> I tested the latest version locally and noted that while it does fix blinking and the annoying scrolling when editing markdown with preview, it does not solve scrolling when switching tabs (at least not in the browsers I tested) - I updated the PR description to reflect this.

Can this bug fix be updated to 3.6.x?

FoSuCloud:

fcollonval:

PR #14583, opened 2023-05-23
Use @jupyterlab/shared-models

fcollonval:
This fixes `@jupyterlab/shared-models` not being provided in the shared scope.

### DIFF
--- a/packages/cells/package.json
+++ b/packages/cells/package.json
@@ -42,7 +42,6 @@
     "watch": "tsc -b --watch"
   },
   "dependencies": {
-    "@jupyter/ydoc": "~0.2.4",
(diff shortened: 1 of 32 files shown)


krassowski:
Integrity failures are relevant. Is there an issue that this PR closes? I think I saw a potentially relevant discussion recently but cannot find it.

fcollonval:
> Integrity failures are relevant. Is there an issue that this PR closes? I think I saw a potentially relevant discussion recently but cannot find it.

I thought that https://github.com/jupyterlab/jupyterlab/pull/14229 solved the trouble seen in https://github.com/jupyterlab/extension-examples/pull/224. But actually it did not. This fixes it - I checked locally.

PR #14588, opened 2023-05-24
Highlight booleans using the same style as keywords

### LINKED ISSUE #14584: Booleans True/False no longer highlighted as numbers in 4.0.0
## Description

JupyterLab 4.0.0 seems to highlight the booleans `False` or `True` with the default font color. In the previous version, jupyterlab used a green color for booleans, integers, and floats to indicate numeric values.

Screenshot 2023-05-23 at 12 23 52

## Reproduce

Download JupyterLab 4.0.0 and open any notebook.

## Expected behavior

True/False highlighted in green, as with floats and integers.

## Context

- Operating System and version: Tested on macOS and Linux (system independent) 
- Browser and version: JupyterLab Desktop 3.6.2-1 
- JupyterLab version: 4.0.0

krassowski:
## References

Fixes #14584 

## Code changes

Adds style for `t.bool`.

Maybe a different style would be preferable (we can change it); the choice here is purely based on style of JupyterLab 3.x

## User-facing changes

| Before on 4.0 | After on 4.0 | 3.x |
|---|---|---|
| Screenshot from 2023-05-24 20-16-35 |Screenshot from 2023-05-24 20-16-48 | Screenshot from 2023-05-24 20-16-00 |

None


krassowski:

jtpio (approved):
Thanks!

Issue #14569, opened 2023-05-20
Cell toolbar does not hide on paste, nor reappear if deleting all text


krassowski:
## Description

- If long text is pasted into the cell the cell toolbar does not auto-hide.
- If long text is removed at once the cell toolbar does not re-appear

## Reproduce

![cell-toolbar-remaining-issues](https://github.com/jupyterlab/jupyterlab/assets/5832902/6e95f109-227f-461d-993a-fbc52af5b927)

## Expected behavior

Cell toolbar overlap logic also fires:
- on paste/cut events
- when all text is deleted

## Context

- Browser and version: Edge 114.0.1823.18 (Official build) beta (64-bit)
- JupyterLab version: 4.0.04

PR #14596, opened 2023-05-26
Fix: Toolbar doesn't update on paste.

brijsiyag:
In order to address delays in DOM layout updates, a `requestAnimationFrame` function is used for the code inside `_updateCellForToolbarOverlap`.
Fixes: #14569


krassowski:
Thank you for working on this! There are some linting failures - the promises which were introduced by converting the `_updateCellForToolbarOverlap` to async would need to be handled:

Screenshot from 2023-05-27 11-46-17

An alternative would be to keep it sync and add `setTimeout` with callback inline.

Would `requestAnimationFrame` work here?

brijsiyag:
Ooh, Thanks, I had no idea of requestAnimationFrame. It works pretty well.
I will update the code.

brijsiyag:
@krassowski some tests are failing yet but I don't think these are related to my changes. PTAL !

krassowski:
There is one visual regression test which actually gets fixed by this PR so relevant in a good sense:

krassowski:
I will close and reopen this PR to trigger CI jobs as those got stuck on bot update.

krassowski (approved):
Thank you @brijsiyag!

Issue #9688, opened 2021-01-28
Add Skip Link to All Regions


manfromjupyter:
Requirement: WCAG 2.4.1 (A)

From: #9399 and doesn't require a screenreader to make/test.
> Expectation: The user can skip navigation functions/sidebar and go straight to the content
> - No skip link exists is available to skip the header, navs, etc and just straight to the left panel or notebook, etc editing.
> - Proposed Solution: On first or second tab, have an element that is off screen gain focus and upon gaining focus changes position down 50px or whatever amount (all via CSS) and something that is either a single skip to the notebook panel or a table of contents that lets you jump to sidebar, left panel, notebook panel, whatever.

For now could simply add a single link. Example how: https://css-tricks.com/how-to-create-a-skip-to-content-link/ but ideally if possible, upon tabbing once, it should cause with css because of this focus to move up. THe container itself is probably a `<div>` with `aria-label="Skip to section"` and gets lowered where you then see a list of like 5-7 anchor links that are linked to the ids of each region we put a landmark on from #9622. 
ex.
1. Header
2. Main
3. Primary Sidetray
(May need to add landmark for the Stacked Panel region with id="jp-left-stack")
4. Secondary Sidetray
(May need to add landmark for the Stacked Panel region with id="jp-right-stack")
5. Footer

If they don't have ids need to add one.

Wait until after the landmark tickets get merged so you have landmarks to jump to (https://github.com/jupyterlab/jupyterlab/issues/9491).

0618:
🙋‍♀️ I can try working on the skip link.

manfromjupyter:
@0618 Yes please! If you get stuck or have questions just reach out here and @isabela-pf and I will tag the people that can help (unless you already know who to talk to). Once you get close I think we can make another ticket with just the info for the skip link in there so it's clear to people what's being merged.

isabela-pf:
@gabalafou was experimenting with the skip link and mentioned some more notes we'll want to be aware of when we get here.

- It seems the current skip link doesn't disappear from view when used, a regression from the work in #10126
- Because we are currently lacking visible focus indicators for all regions of the UI, it can be unclear to users what the skip link did/if it worked because the left side bar is open by default. To be clear, the focus works properly. It may be worth having some kind of stopgap for this if possible.
[...]

PR #14597, opened 2023-05-26
Fix skiplink and add placeholder for additional skiplinks

t03857785:
## References
#9688 Add Skip Links to all regions
#14575 also looks to address this issue

## Code changes
Add skip link widget to the top widget
Added an activate widget function to Shell and implemented main other areas can be implemented when underlying widgets take focus on activation which they do not at present
Added an area target data element to the SkipLink and altered _createSkipLink to allow population

## User-facing changes
The current skiplink functionality does not work as required as focus is not changed. It appears there was some change in focus behavior on side panel expansion previously that is no longer present.
I have altered the functionality to focus on the main panel as you would when you closed or opened a tab there

None


t03857785:
This is an fix to get a skiplink working, additional changes would be beneficial and are anticipated in the code but requires a solution that allows focus on widgets, not sure of the best approach there to allow accessibility and not break existing functionality

packages/application/src/shell.ts
```diff
@@ -770,6 +774,31 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
...
+          current.currentTitle.owner.activate();
+        }
+        return;
+      case 'left':
+      case 'right':
+      case 'header':
+      case 'top':
+      case 'menu':
+      case 'bottom':
+        return;
```
krassowski, line 797: Should it raise non implemented error or at least log a warning?

packages/application/src/shell.ts
```diff
@@ -770,6 +774,31 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate Specified Area Widget.
+   */
+  activateWidget(area?: ILabShell.Area): void {
```
krassowski, line 780: Naming suggestion: `activateArea`?

packages/application/src/shell.ts
```diff
@@ -770,6 +774,31 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate Specified Area Widget.
```
krassowski, line 778:
```suggestion
   * Activate specified Area widget.
```

packages/application/src/shell.ts
krassowski: Actually since this is an API change quite close to the core, I will wait with setting milestone here. It may be better to put it in 4.1

packages/application/src/shell.ts
```diff
@@ -2162,18 +2195,16 @@ namespace Private {
...
-      this._shell.expandLeft();
-    }
     private _shell: ILabShell;
 
-    private _createSkipLink(skipLinkText: string): void {
+    private _createSkipLink(skipLinkText: string, area: ILabShell.Area): void {
       const skipLink = document.createElement('a');
       skipLink.href = '#';
       skipLink.tabIndex = 1;
+      skipLink.addEventListener;
```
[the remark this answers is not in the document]
t03857785: removed I had retained this from #14575

packages/application/src/shell.ts
```diff
@@ -770,6 +774,31 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate Specified Area Widget.
+   */
+  activateWidget(area?: ILabShell.Area): void {
+    switch (area ?? 'main') {
+      case 'main':
+        const current = this._currentTabBar();
```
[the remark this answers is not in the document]
t03857785: I cannot see the check warning but have added the braces

packages/application/src/shell.ts
t03857785: Changed to activateArea

packages/application/src/shell.ts
t03857785: I have thrown a not implemented error here, thinking about how to implement these

packages/application/src/shell.ts
t03857785: Changed as suggested

t03857785:
The failing tests seem to be related to line 30 of shell.spec.ts, 
Changing it from
shell = new LabShell({ waitForRestore: true });
to
shell = new LabShell({ waitForRestore: false });
fixes all of the tests but I am not sure if this is the right fix @fcollonval if you have any input that would be helpful

fcollonval (changes requested):
Thanks @t03857785 here are some comments to improve the code and fix the tests

packages/application/src/shell.ts
```diff
@@ -2135,13 +2166,17 @@ namespace Private {
       this.addClass('jp-skiplink');
       this.id = 'jp-skiplink';
       this._shell = shell;
-      this._createSkipLink('Skip to left side bar');
+      this._createSkipLink('Skip to main panel', 'main');
```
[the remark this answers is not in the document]
t03857785: yes would be nice to sort this but lower immediate priority

packages/application/src/shell.ts
```diff
@@ -308,6 +308,7 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     const skipLinkWidget = (this._skipLinkWidget = new Private.SkipLinkWidget(
       this
     ));
+    this.add(skipLinkWidget, 'top', { rank: 0 });
```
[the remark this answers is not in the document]
fcollonval: Oki I look a bit into this. And we are hit by a CSS rule preventing the skip link to show.

In `packages/application/style/skiplink.css` could you add `contain: none !important;` to the `.jp-skiplink-wrapper` rule? (and remove the `add` call)

Then the focus is not on the skip link because the focus is on the main area element (I need to press like 8 time Shift-Tab to get on the Skip link widget using the code on the `main` branch). Do you know how we could improve that?

gabalafou (commented):
@fcollonval had a question that led me to spend a good amount of time this week thinking about the JupyterLab skip link.

It was the question about how to improve the situation where the skip itself gets "skipped" over because the application focuses the main area when it loads. 

It's a thorny problem.

For one thing, I think this behavior has been a part of JupyterLab for quite some time:

- Looks like Steve added the [`dock.activateWidget(widget)`](https://github.com/jupyterlab/jupyterlab/commit/8782ee7877f5d62c5d564a9a2e876555e78fe3f5#diff-ab50b55e5abb5056564fb1392a224c39769836ae96296d781ec2be3c19131f53R407) line 5 years ago.
- And it looks like the [MainAreaWidget has been focussing its content node on activate](https://github.com/jupyterlab/jupyterlab/commit/8782ee7877f5d62c5d564a9a2e876555e78fe3f5#diff-87e821a396b7aa41c021de1352d73a0f51cab272261d9a73356dadac6f3a6078L33) for even longer

Another thing is that a lot of apps have similar behavior, for example Google Docs, Google Colab, and VS Code in the browser (for example, github.dev/jupyterlab/jupyterlab if you previously had a file open). Notably, HackMD does not follow this behavior, neither in split mode (source plus rendered markdown side-by-side) nor in single pane edit mode. 

I'm honestly not sure what the best option is from an accessibility point of view. If our choice is a binary like so:

- Auto focus the main area
- Do not auto focus anywhere

Then to me, it feels less like an accessibility issue per se, and more like we're making a trade-off between the new user experience versus a returning user. For a returning user, it's probably better to drop them directly into the main area, because that's presumably where they left off or where they want to work. For a completely new user, it could be a bit disorienting to have the application "auto skip" them to the main area. At the moment, I don't have any good ideas on how to mitigate that concern, or really, how to resolve the different needs here between a returning versus new user.

If we can't rethink this, if we decide to stick with the status quo and always activate (and thereby focus) the main area, then I am led to wonder if the skip link still provides value for accessibility. 

[Skip links may seem necessary for WCAG conformance](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html), but when I read the specification closely, I'm not actually sure that it's necessary to have a skip link in JupyterLab in order to meet WCAG 2.4.1 for two reasons:

1. The success criterion refers to "multiple pages." Once again, we must keep in mind that WCAG was written with normal websites in mind, not web applications. In the case of a normal website, it's expected that the user might navigate to several different pages on the site, and each of those pages might have a nav bar, side bar, table of contents, etc. that the user needs to skip past in order to get to the main content. This is distinct from JupyterLab, which is a single page application. In JupyterLab, the user is never expected to navigate to a new web page. They're not in a situation where a new page load causes their cursor/focus to constantly return to the top of the page, which as far as I can tell is the situation in which the skip link really becomes critical to website usability for many disabled users.
2. The success criterion can be met in other ways, such as [using ARIA regions](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11.html).

To bring things back to this PR, I think we should think about both the short term and the long term.

In the short term, since with this PR we are very close to fixing the skip link—and there may be some situations in which it still provides value—then I say we go ahead and fix it but perhaps we don't spend any more time on perfecting or extending it.

For the long term, I am going to throw what might be a bit of a curve ball. For the long term, I think we should consider getting rid of the skip link widget and re-using the command palette as a skip link instead. There are a lot of details that would need to be worked out, but I think I can imagine a future in which the command palette is reworked so that it becomes the first item in the tab order: if a user presses the tab key from the browser address bar, it drops them into the command palette, perhaps with the command palette already pre-populated with a (currently non-existent) command to "Go to the main panel" and a list of (also currently non-existent) commands to go to other regions of the UI (maybe also a "Learn more about the command palette" command).

This would also help with command palette discoverability. As of now, you kind of a have to know that it exists in order to find it, but if we made it something that the user can bring up by  pressing the tab key, then it's more likely to be discovered.

But you may wonder how any of this addresses @fcollonval's question that I started with. Honestly, I do not see an easy path to changing the current behavior of the app—the behavior that puts focus on the main area when the app loads. So in my mind, we're stuck with that behavior. I just take it as a given. (Perhaps it's worth raising to the community at large. I don't know.) If it's a given, then we can think instead of ways to improve things without changing that one behavior. Currently, you cannot access the skip link without tab-key-pressing through the interface. But if we roll that functionality into the command palette, then users could effectively access it via a keyboard shortcut. On Mac, for example, a user would be able to press Cmd+Shift+C then type "go" then hit Enter in order to go to the main area. 

I know that was long. Thanks for reading this far! I wanted to make sure to put my thoughts on the subject out in public since I've spent so much time digging into it this week.

gabalafou:
The test failures don't look like they have anything to do with this PR

packages/application/src/shell.ts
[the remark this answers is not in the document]
t03857785: resolved by  commit 383dfdd

gabalafou (approved):
My two cents: if this is safe to merge, we should go ahead and merge it.

Even if the skip link itself is usually skipped by JupyterLab's auto-activation of the main area, this PR puts in the scaffolding to build out skip functionality to other regions of the app, which we will be able to re-use later in other places such as the command palette.

packages/application/style/skiplink.css
```diff
@@ -6,6 +6,10 @@
 
 .jp-skiplink-wrapper {
   overflow: visible;
+
+  /* override strict containment added via Lumino PR
+   [#506](https://github.com/jupyterlab/lumino/pull/506) */
+  contain: none !important;
```
gabalafou, line 12: @krassowski this CSS change needs specific scrutiny and review by you because I'm not sure if this has downstream performance implications that would undo a lot of the performance work you've done.

gabalafou (changes requested):
Marking as request changes until @krassowski has a chance to approve the CSS change. The performance implications may outweigh the benefits provided by this PR.

packages/application/style/skiplink.css
krassowski: Thank you @gabalafou sorry I missed it earlier. Relaxing `contain` locally for leaf-level widgets is fine, but most likely it does should still not be set to `none`. `strict` means `size layout paint style` so one should first replace it with these four and then remove one by one to see which can stay and which to remove. As there are few children in skiplink it would not matter much though.

krassowski: @t03857785 just pinging you directly to ensure that you saw the suggestion above.

t03857785: I am taking a look at the moment to determine which option(s) work as intended

t03857785: I have updated this to the most restrictive contain possible which still has the correct rendering

packages/application/src/shell.ts
```diff
@@ -789,6 +792,34 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate specified Area widget.
+   */
+  activateArea(area: ILabShell.Area = 'main'): void {
```
krassowski, line 798: Given that only `main` is implemented as of now I think we should add `@alpha` (is this the right one @fcollonval?) in documentation string and explain that only `main` is implemented.

Otherwise this PR looks fine to merge to me.

t03857785: @fcollonval @krassowski Shall I add @alpha to this comment now I would really like to get this merged then I can look at adding new elements once there is more to link to

packages/application/src/shell.ts
```diff
@@ -2231,18 +2266,15 @@ namespace Private {
...
-    private _focusLeftSideBar() {
-      this._shell.expandLeft();
-    }
     private _shell: ILabShell;
 
-    private _createSkipLink(skipLinkText: string): void {
+    private _createSkipLink(skipLinkText: string, area: ILabShell.Area): void {
       const skipLink = document.createElement('a');
       skipLink.href = '#';
       skipLink.tabIndex = 1;
```
tonyfast, line 2274:
```suggestion
      skipLink.tabIndex = 0;
```

it is recommended to only use `0` and `-1` for `tabindex`.

tonyfast:
testing on binder on the demo branch. i am not observing the skip link. with the orca screen activated the stops are:

1. moves the virtual cursor and announces the title of the notebook
2. a non-visual link that announces skip to left side bar.
3. returns with focus to the editable version of the the title
4. generate shared link

i would expect to see a visual skip link and i'd expect that the first tab is a skip link (which it isn't).

t03857785:
> testing on binder on the demo branch. i am not observing the skip link. with the orca screen activated the stops are:
> 
> 1. moves the virtual cursor and announces the title of the notebook
> 2. a non-visual link that announces skip to left side bar.
> 3. returns with focus to the editable version of the the title
> 4. generate shared link
> 
> i would expect to see a visual skip link and i'd expect that the first tab is a skip link (which it isn't).

Sorry tony this branch had not been merged into the demo branch hence you have seen these issues that this was designed to fix.
It has been merged into demo now and should provide a link to main, other links can be added in the future as required.

packages/application/src/shell.ts
```diff
@@ -789,6 +792,34 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate specified Area widget.
+   */
```
gabalafou, line 797: I think the following comment will do the trick. @fcollonval, does this look good to you?

```suggestion
  /**
   * Activate specified Area widget.
   * 
   * ### Notes
   * The alpha version of this method only supports activating the "main" area.
   *
   * @alpha 
   * @param area Name of area to activate
   */
```

gabalafou:
Concerning @tonyfast's recent review, I guess the question his testing raises is: is it within the scope of this PR to make sure that the skip link is always the first tab stop on the page? Or is the goal of this PR just to make the skip link visible again and to fix the click action?

I think that fixing the tab stop issue is going to be a bit difficult because I suspect that a lot of people are used to having JupyterLab drop them directly into the notebook. As I mentioned before, there's a design tradeoff here between new and returning users. 

I think whenever JupyterLab is loaded with an open notebook, we should keep the current app behavior and drop the user directly into the notebook, since I believe that's already an established UI pattern of document-editing apps. 

It's not as clear to me what should happen when JupyterLab is loaded without an open document, and it's just the launcher tab. Should the UI drop the user into the launcher tab? Not sure.

t03857785:
> Concerning @tonyfast's recent review, I guess the question his testing raises is: is it within the scope of this PR to make sure that the skip link is always the first tab stop on the page? Or is the goal of this PR just to make the skip link visible again and to fix the click action?
> 
> I think that fixing the tab stop issue is going to be a bit difficult because I suspect that a lot of people are used to having JupyterLab drop them directly into the notebook. As I mentioned before, there's a design tradeoff here between new and returning users.
> 
> I think whenever JupyterLab is loaded with an open notebook, we should keep the current app behavior and drop the user directly into the notebook, since I believe that's already an established UI pattern of document-editing apps.
> 
> It's not as clear to me what should happen when JupyterLab is loaded without an open document, and it's just the launcher tab. Should the UI drop the user into the launcher tab? Not sure.

@gabalafou just for clarity the skiplinks are the last tab on the page so that if you shift-tab they will appear, the goal of this PR is to restore skiplinks and fix the click action so that additional skiplinks can be added in the future

gabalafou:
@t03857785, sorry, when I made that comment I hadn't realized that Tony had tested on a branch that didn't include these PR changes, so I got confused. 

For clarity, my point was that this PR doesn't **totally** fix the skip link because a skip link should be the first (or one of the first) tab stops on the page. But also: I think it's totally okay for this PR to **not** fix that aspect of the skip link because trying to ensure that the skip link is the first tab stop in the app could break existing expectations of how the app behaves (namely, that the app should move your focus to the notebook if you have one open when you load the app).

packages/application/src/shell.ts
```diff
@@ -789,6 +792,40 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     }
   }
 
+  /**
+   * Activate specified Area widget.
```
gabalafou, line 796:
```suggestion
   * Activate widget in specified area.
```

t03857785:
> @t03857785, sorry, when I made that comment I hadn't realized that Tony had tested on a branch that didn't include these PR changes, so I got confused.
> 
> For clarity, my point was that this PR doesn't **totally** fix the skip link because a skip link should be the first (or one of the first) tab stops on the page. But also: I think it's totally okay for this PR to **not** fix that aspect of the skip link because trying to ensure that the skip link is the first tab stop in the app could break existing expectations of how the app behaves (namely, that the app should move your focus to the notebook if you have one open when you load the app).

I think merging this will add useful functionality that can be extended later as required, the restoration of skip link functionality as a first step seems a useful building block even if at present as it was previously it is imperfect.

gabalafou (approved):
Thanks for your contribution and your perseverance getting this through review!

This looks good to me and all of the checks are passing except the Galata visual regression tests. I tried re-running those tests but they still failed. Most (but not all) of the failures are timeouts, not sure about the others. But I don't see anything in the failures that seem linked to this PR, so I'm approving this from my end.

gabalafou:
And if I don't hear any objections, then I will merge this PR in next week on Wednesday.

krassowski:
@gabalafou @fcollonval was there any objection against merging this one, or can I merge?

tonyfast (changes requested):
https://github.com/deathbeds/jupyak/pull/16#pullrequestreview-1763141483 builds a lite site to test this feature (use a private browser). https://jupyak--16.org.readthedocs.build/en/16/_static/work/lite/lab/index.html

it is weird to me that the skip link is the second tab stop on the page. i'm not sure where the first tab is taking me. with orca it announces "main content region launcher content kernel icon image". so for screen reader users, that first tab stop is probably super confusing. i can't figure what element is being activated and why.

krassowski:
> `main content region launcher content kernel icon image`

Are we sure this is coming from this PR? I see that jupyak merges a number of PRs.

tonyfast:
good point. i can't build locally. can you check if the skip link is the first or second tab stop?

krassowski:
> can you check if the skip link is the first or second tab stop?

It seems to me this is the second stop as you indicated, at least if counting `<html>` as the first stop. Is it normal that `<html>` gets focus? Anyways, the key question is if this is an undesirable behaviour that we already have, or if it was introduced in this PR.

tonyfast:
thanks for confirming.  i've actually never encounter `html` getting focus. that is wild. i think this is undesirable behavior. i think the fix is out of scope for relative to OP, but it is in scope for the title of the pull request. nominally, i would guess a pull request fix would fix the tab focus and that the skip link is the first tab stop. 

if we are going to message [improved keyboard navigation](https://github.com/jupyter/accessibility/issues/142) then i think a fix for the skip link tab order is prudent.

gabalafou:
> It seems to me this is the second stop as you indicated, at least if counting `<html>` as the first stop. Is it normal that `<html>` gets focus?

I saw this in my notifications and just dropping in real quick here to mention that this happens when you have Dev Tools open. I didn't know this myself until I started working on Tab stops, and I lost several hours of my life being very confused until I figured out that it was Dev Tools

tonyfast (approved):
approved after reading the recent comment by @gabalafou 

while i don't know if this pr effects notebook. i can confirm that the skip link to go to the main content is activated in both Jupyterlab and Notebook. In Lab, the skip link says <samp>Skip to main content</samp>. In notebook, the skip link says <samp>Skip to Main</samp>. it is good that the call to action is different because they are different applications.

jtpio:
> while i don't know if this pr effects notebook. i can confirm that the skip link to go to the main content is activated in both Jupyterlab and Notebook. In Lab, the skip link says Skip to main content. In notebook, the skip link says Skip to Main. it is good that the call to action is different because they are different applications.

Notebook implements a different shell, and has its own skip link, which was added in https://github.com/jupyter/notebook/pull/6844.

Maybe the changes in this PR will have to be ported to the notebook repo (and maybe adapted)

PR #14600, opened 2023-05-26
Add readonly indicator

### LINKED ISSUE #7980: UX: Clearer Read Only Status

andrewfulton9:
## References
This is a update of #12826, which fixes #7980.

## Code changes

I've reworked the status indicator to match wireframes in #7980. I also added a notification for any opened read-only files, though they may not be as loud as wanted.

## User-facing changes

Adds a status indicator next to debug button/kernal status. Also adds a notification if a read-only permissioned file is opened.

image
Above is a read-only notebook with status indicator and the notification pop-up.

image
Here is the same in dark mode.

image
And this is what the read-only file indicator on a file looks like.

None that I am aware of.


krassowski (changes requested):
Thank you for picking this up! When testing on Binder I noticed that this logic does not seem to work well in RTC mode as it shows that notebook is readonly even though it is perfectly writable:

Screenshot from 2023-05-27 15-12-30

On the other hand, when it is actually needed, this is in readonly source preview that can be opened from debugger, the indicator does not show up:

file-not-marked-as-readonly

andrewfulton9:
@krassowski, can you take another look at this when you get a chance? I changed up where the read-only indicator is defined and implemented to be all in the docregistry extension. This allowed me to keep all the logic regarding file vs notebook handling, and RTC all in the same place instead of duplicating code. To do this I had to switch it to adding the indicator to the toolbar via `Toolbar.addItem` or `Toolbar.insertBefore`after the fact rather than defining it the toolbar schema for the file and notebook editors each respectively. For now, I am just by-passing the indicator when RTC is being used, since the autosaving functionality was causing the files to all appear as read-only. Currently in RTC, the autosave fails semi-silently (shows up in the logs, but not in the jupyterlab UI) when a file is read-only which likely should be its own issue.

andrewfulton9:
Also, should `yarn.lock` be updated in a commit on the PR, or is that done as part of CI?

packages/ui-components/style/toolbar.css
```diff
@@ -101,6 +102,11 @@ button.jp-ToolbarButtonComponent .jp-ToolbarButtonComponent-label {
   font-family: var(--jp-ui-font-family);
 }
 
+.jp-ToolbarLabelComponent {
+  background-color: var(--jp-brand-color1);
+  color: white;
```
krassowski, line 107: Does it look good in both light and dark mode? Can you add screenshots to the top-level comment?

I have not tried it yet as the Binder is flaky.

krassowski (commented):
This looks like it's no on the right track, noting some minor things to address.

packages/ui-components/style/toolbar.css
andrewfulton9: No, I'll give that a try and post the screen shots

packages/docregistry/src/default.ts
```diff
@@ -616,6 +616,25 @@ export class DocumentWidget<
     if (args.name === 'dirty') {
       this._handleDirtyState();
     }
+    if (!this.context.model.dirty) {
+      const path = this.context.contentsModel?.path;
+      if (path && !path.startsWith('RTC:')) {
```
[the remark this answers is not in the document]
andrewfulton9: No, that looks a lot cleaner though. I'll give it a try.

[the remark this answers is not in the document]
andrewfulton9: This has been updated to use `this.context.model.collaborative`. Thanks for pointing that out!

packages/notebook/src/panel.ts
```diff
@@ -266,8 +259,6 @@ export class NotebookPanel extends DocumentWidget<Notebook, INotebookModel> {
     });
   }
 
-  translator: ITranslator;
```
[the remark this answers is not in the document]
andrewfulton9: This has been added back in

packages/ui-components/style/toolbar.css
andrewfulton9: I updated this to `var(--jp-ui-inverse-font-color1)`. Let me know what you think. I added in screenshots to the top-level-comment that you can see.

andrewfulton9:
@krassowski, could you take another look at this when you get a chance?

packages/docregistry/src/default.ts
```diff
@@ -616,6 +616,24 @@ export class DocumentWidget<
...
+            'kernelName',
+            'read-only-indicator',
+            readOnlyIndicator
+          );
+          if (!roi) {
+            this.toolbar.addItem('read-only-indicator', readOnlyIndicator);
+          }
+          Notification.warning(
+            this._trans.__(`%1 is a read-only document.`, this.title.label)
+          );
```
krassowski, line 633: I think that it would be better UX if this notification was only shown on an attempt to save with <kbd>ctrl</kbd> + <kbd>s</kbd>. See this fragment of Isabela's comment https://github.com/jupyterlab/jupyterlab/issues/7980#issuecomment-655703618 (emphasis mine):

>With toast notifications moving forward in 689, I think this could be a good opportunity to remind users a notebook is read-only **if they try to save**.

Further, currently the notification does not show up but just lands in the notification centre (makes the bell in status bar blue and requires two clicks to dismiss). As it would be responding to user action I think it should be shown outright.

Nice to have: it could also include a button for "Save as" command.

packages/docregistry/src/components.tsx
```diff
@@ -0,0 +1,33 @@
...
+  panel: IDocumentWidget,
+  translator?: ITranslator
+): Widget {
+  let trans = (translator ?? nullTranslator).load('jupyterlab');
+  return ReactWidget.create(
+    <div>
+      <span
+        className="jp-ToolbarLabelComponent"
+        title={trans.__(
+          `document is permissioned readonly; "save" is disabled, use "save as..." instead`
```
krassowski, line 26:
```suggestion
          `Document is permissioned readonly; "save" is disabled, use "save as..." instead`
```

krassowski (commented):
My only remaining concern is on the UX level with implementation of the notification. Otherwise it seems good to go.

packages/ui-components/style/toolbar.css
```diff
@@ -101,6 +102,11 @@ button.jp-ToolbarButtonComponent .jp-ToolbarButtonComponent-label {
   font-family: var(--jp-ui-font-family);
 }
 
+.jp-ToolbarLabelComponent {
```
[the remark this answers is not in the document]
andrewfulton9: Sounds good. I added this.

packages/docregistry/src/default.ts
andrewfulton9: I'm not sure how to deal with the notification on Ctrl + s without enabling the save commandID. Personally, I think having it disabled is more clear than enabling it and notifying if the user clicks. I did update to remove the notification, updated the caption that shows up when hovering over the save button and I added a dialog that pops up if the "save all" command is given.

What do you think? Is there a way to keep the save command disabled but still clickable?

krassowski: Thank you. FYI https://github.com/jupyterlab/jupyterlab/pull/14664 changes `isEnabled` for toolbar commands which will make it always enabled even if it appears as disabled in the menu. I think we can proceed with implementation as-is and see if it needs adjusting from user feedback.

andrewfulton9:
@fcollonval, would you mind giving this PR a review when you have a chance? Thank you!

packages/docregistry/src/default.ts
```diff
@@ -616,6 +617,21 @@ export class DocumentWidget<
     if (args.name === 'dirty') {
       this._handleDirtyState();
     }
+    if (!this.context.model.dirty) {
```
krassowski, line 620: Can you remind my what the `dirty` check does here?

packages/ui-components/style/toolbar.css
```diff
@@ -101,6 +102,12 @@ button.jp-ToolbarButtonComponent .jp-ToolbarButtonComponent-label {
...
+  background-color: var(--jp-brand-color1);
+  color: var(--jp-ui-inverse-font-color1);
+  font-size: var(--jp-ui-font-size0);
```
krassowski, line 108: One tiny addition:

```suggestion
  font-size: var(--jp-ui-font-size0);
  user-select: none;
```

To avoid inconsistent look when user accidentally highlights it:

Screenshot from 2023-08-29 21-33-07

krassowski (approved):
Thank you @andrewfulton9.

I opened https://github.com/jupyterlab/jupyterlab/issues/15040 to track follow-up improvements to the logic. I think this PR can be merged as it adds the widget scaffold which can be re-used in the remaining scenarios.

packages/docregistry/src/default.ts
andrewfulton9: If I remember correctly if `this.context.model.dirty === true` then `this.context.contentsModel.writable` could also be true even if it was a read only file or vice-versa (I forget which). By adding this check here, I'm making sure it doesn't try to add a read only label until the context is fully initialized.

PR #14601, opened 2023-05-27
Inherit font family from `.cm-editor`

### LINKED ISSUE #14592: The backslash symbol is replaced by the Korean won(₩‎) symbol.
In JupyterLab 3.6, the backslash symbol appears correctly.
image

In JupyterLab 4.0, the backslash symbol is replaced with the Korean won(₩‎) symbol.
image

Is there a way to make the backslash symbol appear as it should?

I use Korean Windows. I tested it on Python 3.9, 3.11 and it was the same. It doesn't seem to be a python version issue.

krassowski:
## References

Should fix #14592

## Code changes

| Before | After | 3.x (reference)
|--|--|--|
|Screenshot from 2023-05-27 11-07-46 | Screenshot from 2023-05-27 10-56-07 | image |

Note: this PR uses `inherit`(as in 3.x) rather than just hard-coding font family in `.cm-scroller` to `var(--jp-code-font-family)` as `inherit` this is needed for changing fonts from settings to work.

## User-facing changes

Changing font family from settings works again

| Before | After | 
|--|--|
| Screenshot from 2023-05-27 11-05-18 | Screenshot from 2023-05-27 11-00-48 |

None

### DIFF
--- a/packages/codemirror/src/theme.ts
+++ b/packages/codemirror/src/theme.ts
@@ -32,6 +32,11 @@ export const jupyterEditorTheme = EditorView.theme({
     caretColor: 'var(--jp-editor-cursor-color)'
   },
 
+  /* Inherit font family from .cm-editor */
+  '.cm-scroller': {
+    fontFamily: 'inherit'
+  },
+
   '.cm-cursor, .cm-dropCursor': {
     borderLeft:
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks

PR #14605, opened 2023-05-28
Fix flaky search tests

### LINKED ISSUE #14550: UI Tests check is flaky on CI
## Description

UI Tests check have been flaky on CI recently.

## Reproduce

For example: https://github.com/jupyterlab/jupyterlab/actions/runs/4984379184

Does not show any failure in the report: 

image

But in the logs:

image

## Expected behavior

It should pass.

## Context

This adds noise when reviewing  PRs and would be great to fix.

krassowski:
## References

Fixes #14550 (except for the issue with benchmark test which probably deserves its own issue).

## Code changes

Accepts confirmation dialog when it shows up if notebook has not fully loaded and awaits for initial search before entering the cell

## User-facing changes

None

None

### DIFF
--- a/galata/test/jupyterlab/notebook-search.test.ts
+++ b/galata/test/jupyterlab/notebook-search.test.ts
@@ -203,6 +203,11 @@ test.describe('Notebook Search', () => {
 
     await page.click('text=Search Cell Outputs');
 
+    // If the notebook is not fully loaded yet a confirmation dialog will show up
+    if (await page.locator('.jp-Dialog').isVisible()) {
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Nice thanks!

jtpio:
> except for the issue with benchmark test which probably deserves its own issue

Yes it would make sense to track this in an issue, especially if the UI tests are still failing because of it (like in this PR)

krassowski:
Thanks! I opened https://github.com/jupyterlab/jupyterlab/issues/14608.

Issue #14576, opened 2023-05-22
Defer loading of extensions


brichet:
### Problem

Some extensions can be very long to load, and extends the pageload time of the application.
It could be interesting to defer the loading of some extensions after the application is started.

In addition, there may be a confusion between `autoStart` option and `deferredExtensions` attribute:
> Currently the *page_config.json* allows a `deferredExtensions` attribute, containing a list of extensions or plugins. 
> It looks like these deferred extension are not started at all if they are not required. It acts like the `autoStart: false` plugin parameter.

### Proposed Solution

1. **Lumino**
- add a `deferred` option (boolean) to the plugins, to start a plugin only after the main application is started.

2. **JupyterLab**
- use the `deferredExtensions` attribute from *page_config.json* like the deferred option from plugins.
- wait for the application fully restored to activate the deferred extensions.
- *OPTIONAL*: add an `autoStartExtensions` attribute to the *page_config.json*, which act like the `autoStart` option in plugins.

PR #14610, opened 2023-05-29
Activate the deferred plugins after shell restore

brichet:
This PR improves pageload time by deferring the activation of huge extensions.

After the shell restore, activates the plugins that have been deferred from:
- `page_config.json`, currently they are not activated at all if not required by another plugin
- the new plugin option `deferred` coming from https://github.com/jupyterlab/lumino/pull/588

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14576

Related to https://github.com/jupyterlab/lumino/pull/588

## Code changes

The changes are in `JupyterLab` class.

## User-facing changes

Activation of a plugin containing a 5s sleep:

<table>
  <tr>
    <th>CURRENT LOADING</th>
    <th>DEFERRING ACTIVATION</th>
  </tr>
  <tr>
    <td></td>
     <td></td>
  </tr>
</table>

This is currently backward compatible but I used `// @ts-ignore` to avoid error at compilation.
There may be better implementation for it.

### DIFF
--- a/packages/application/src/lab.ts
+++ b/packages/application/src/lab.ts
@@ -44,15 +44,9 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
 
     this.restored = this.shell.restored
       .then(() => {
-        // Backward compatibility
-        try {
-          // @ts-ignore
-          this.activateDeferredPlugins().catch(error => {
-            console.error('Error when activating deferred plugins\n:', error);
-          });
-        } catch (error) {
-          // no-op
-        }
+        this.activateDeferredPlugins().catch(error => {
+          console.error('Error when activating deferred plugins\n:', error);
+        });
 
         if (this._info.deferred) {
           Promise.all(
--- a/packages/application/src/lab.ts
+++ b/packages/application/src/lab.ts
@@ -47,7 +47,7 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
         // Backward compatibility
         try {
           // @ts-ignore
-          this.activateDeferredPlugins();
+          this.activateDeferredPlugins().catch(error => { console.error('Error when activating deferred plugins\n:', error); };
         } catch (error) {
           // no-op
         }
--- a/packages/application/src/lab.ts
+++ b/packages/application/src/lab.ts
@@ -53,9 +53,9 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
         }
 
         if (this._info.deferred) {
-          this._info.deferred.matches.forEach(pluginID =>
+          Promise.all(this._info.deferred.matches.map(pluginID =>
             this.activatePlugin(pluginID)
-          );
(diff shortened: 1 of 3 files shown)


packages/application/src/lab.ts
```diff
@@ -45,6 +42,24 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
...
+          // @ts-ignore
+          this.activateDeferredPlugins();
+        } catch (error) {
+          // no-op
+        }
+
+        if (this._info.deferred) {
+          this._info.deferred.matches.forEach(pluginID =>
+            this.activatePlugin(pluginID)
+          );
```
fcollonval, line 58: Could we wrap this in a promise and display any error?

```suggestion
          Promise.all(this._info.deferred.matches.map(pluginID =>
            this.activatePlugin(pluginID)
          )).catch(error => { console.error('Error when activating customize list of deferred plugins:\n', error); });
```

packages/application/src/lab.ts
```diff
@@ -45,6 +42,24 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
     // Populate application info.
     this._info = { ...JupyterLab.defaultInfo, ...info };
 
+    this.restored = this.shell.restored
+      .then(() => {
+        // Backward compatibility
+        try {
+          // @ts-ignore
+          this.activateDeferredPlugins();
```
fcollonval, line 50:
```suggestion
          this.activateDeferredPlugins().catch(error => { console.error('Error when activating deferred plugins\n:', error); };
```

packages/apputils-extension/src/settingconnector.ts
```diff
@@ -59,8 +59,8 @@ export class SettingConnector extends DataConnector<
     }
 
     return {
-      ids: ids.filter(id => !isDeferred(id) && !isDisabled(id)),
-      values: values.filter(({ id }) => !isDeferred(id) && !isDisabled(id))
```
fcollonval, line 63: Just to be sure, this will load the settings even for deferred extensions avoiding multiple calls (in particular if they contain definition to be gathered by a transform operation?)

fcollonval (changes requested):
Thanks @brichet 

I have some comments and I think we should wait for a new release of lumino with the related change. As the behavior of deferred extensions is gonna changed, I would prefer to have it in a new minor version.

packages/apputils-extension/src/settingconnector.ts
brichet: Deferred extension settings should be loaded in exactly the same way as non-disabled extension settings.
I think the settings for deferred extensions were not loaded at all before.

brichet:
Test failure does not appear to be linked to this PR.

fcollonval (approved):
Thanks @brichet

jtpio:
Kicking another rebase to grab the latest fixes to the UI tests on `main`.

fcollonval:
Thanks @brichet merging this one now.

Issue #10004, opened 2021-03-25
Everything needed for Low Vision (up to 400% zoom) (Mobile/Tablet-Responsive) Support


manfromjupyter:
**Purpose:** Fully support all users with a low vision disability on JupyterLab  
**All Screenshots:** https://imgur.com/a/OStTXXr
**Derivative from:** #9399 

## Todo
All sites needs to have the capability to zoom to 200% for WCAG2.0 compliance (~Tablet support), or 400% zoom for WCAG2.1 compliance (~Mobile support) WITHOUT causing anything but vertical scroll bars to appear. The goal of this ticket is to support 400% zoom. If we're gonna do it, let's do it all the way.
1. All header dropdowns need to no longer truncate off the side of the page (6-7 screenshots on the [imgur link](https://imgur.com/a/OStTXXr))
2. All panels needs to have their heights flexible (ex. 100%) so they can be scrolled (may have to remove a `overflow: hidden` somewhere)
3. All panels with sections with predefined heights (ex. 200px), this needs to be removed and moved to an automatic height. Goal is to remove all vertical scroll bars but the main one that controls the scrolling down of the entire panel section. Some of these sections are collapsible so if having to scroll for too long is a problem, new features need to be added to support collapsing of other panel's sections that don't already have it. (2-3 screenshots on the [imgur link](https://imgur.com/a/OStTXXr))
4. Not labeled on screenshots but very important is that the right sidebar AND footer still appears. If the final design decision for the sidebars is to make them derive from a single source, switch to horizontal tabs, OR collapse on each side to give way for more real estate for the notebook section, that is beyond the scope of this ticket. For this ticket, merely need to make it sticky (on the right side and left side) at every screensize AND make sure the footer is sticky to the bottom (it doesn't take up much space, should be fine). `position: fixed` in CSS will let you affix something in one place so it remains on scroll.
5. Not labeled on screenshots but very important is that the sidebars need to not have elements that are unable to be scrolled to. Example in the screenshots, the bottom 2 icons to opening debugger panel and extensions panel I think is unable to be put into view to be clicked on at 400% zoom. This can be done by either making the height set to a viewport height (vh) of maybe 100% or calc(100vh-120px) or whatever the size of the header is and then just make sure a min and max-width is set so the size does not get too small or too large if they move all of the actions to the opposite side bar or zoom in to 500% despite it not being supported. The alternative and perhaps ideal way to achieve this is to just allow it also to be vertically scrolled down.

[...]

PR #14626, opened 2023-06-01
400% Zoom screen usability enhancements

j264415:
## References
#10004  Full low Vision Support (400% Zoom) (Mobile/Tablet-Responsive Support)

## Code changes

Changed padding and spacing to use min CSS function alongside vh (viewport-height) and vw(viewport-width) to reduce proportional spacing at high zoom and increase useful screen real estate usage.

## User-facing changes

Currently with the 400% zoom at 1280px the help main menu bar does not appear and within the file-browser tab bar only half of one 'dir-listing' shows up. 
The changes I have made now allows for the help dropdown to be shown in the main menu. I found that when trying to check the changes the sidebar 'Extension Manger' disappears from view, so this was also changes so that users that needs this implemented would be able to have access to it without zooming back out. Furthermore, within the file-browser tab, users would now be able to see least 2 notebook directory listed below than before. 

None

@krassowski Please review and comment

### DIFF
--- a/packages/apputils-extension/schema/notification.json
+++ b/packages/apputils-extension/schema/notification.json
@@ -27,7 +27,7 @@
   "properties": {
     "checkForUpdates": {
       "title": "Check for JupyterLab updates",
-      "description": "Whether to check for newer version of JupyterLab or not. It requires `fechNews` to be `true` to be active. If `true`, it will make a request to a website.",
+      "description": "Whether to check for newer version of JupyterLab or not. It requires `fetchNews` to be `true` to be active. If `true`, it will make a request to a website.",
       "type": "boolean",
       "default": true
     },


krassowski:
In the current state this PR would modify the UI on 100% zoom too, leading to some regressions, notably, the menu bar is cut off (has to be fixed) and the spacing in sidebars is smaller (ideally would be fixed to):

| Before | This PR |
|--|--|
| simple-mode-expected | simple-mode-actual |

j264415:
@krassowski Please see changes made from feedback.

j264415:
The before changes meant the the UI was cutting off the menu on 100% zoom. 
These are the changes made from the feedback changes the UI but at the point where it matters. These means that when returned to a 100% zoom menu back will not be cut off anymore and sidebar spacing will be fixed.

j264415:

j264415:

j264415:
please update snapshots

PR #14650, opened 2023-06-07
Update README.md

shellingghost:
Updated README.md file to help new users better understand how to update the path, what the path is, and what it means for them moving forward. This commit was created in the hopes of lowering the barrier to entry so users who are not as computer savvy, but are more research inclined, may have an easier transition into jupyter notebooks.


krassowski:
The proposed additions are rather generic. I mean that these could be equally relevant to a user of Jupyter Notebook and JupyterLab. Maybe it would be better if these were described in the project-wide documentation on https://docs.jupyter.org and then both JupyterLab and Notebook docs could link there?

krassowski:
@shellingghost should we close this PR in favour of updating the project-wide documentation?

JasonWeill:
After discussion today with other JupyterLab contributors, we decided to close this pull request. The README is intended to be short and only include basic information. Providing more detailed information should be done in the main documentation. Thank you for your interest in contributing to JupyterLab.

PR #14660, opened 2023-06-08
Improve typing to avoid bad usage of `context.contentsModel.content`

### LINKED ISSUE #14659: Context.contentsModel is not a copy and provide the content
## Description

The API `context.contentsModel` should not give access to the file content (as stated in the method documentation) but it does. This can give the impression the file content can be modified using it. But this is a very wrong assumption; its sole purpose is to give access to file metadata like is the file writable. Moreover we do not copy the content model what could lead to its corruption.

Xref: https://github.com/jupyter-widgets/ipywidgets/pull/3752#discussion_r1221410162

fcollonval:
## References

Fixes #14659

## Code changes

Remove explicitly `content` from `IContext.contentsModel` to ensure developers don't try to access the file content from that API - it was only written in the documentation that the content will be empty.

## User-facing changes

None

Change `IContext.contentsModel` from `Contents.IModel` to `Omit<Contents.IModel, 'content'>`
Change `IContext.fileChanged` to pass `Omit<Contents.IModel, 'content'>` as argument instead of `Contents.IModel`.

Although this is a breaking change, it clarifies the API and prevents developper of consuming an always `undefined` content.

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -569,7 +569,7 @@ export namespace ToolbarItems {
    */
   export function createSaveButton(
     commands: CommandRegistry,
-    fileChanged: ISignal<any, Contents.IModel>
+    fileChanged: ISignal<any, Omit<Contents.IModel, 'content'>>
   ): Widget {
     return addCommandToolbarButtonClass(
       ReactWidget.create(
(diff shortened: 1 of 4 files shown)


krassowski (approved):
Thank you!

fcollonval:
Thanks @krassowski for the review

fcollonval:

PR #14664, opened 2023-06-09
Fix semantic commands enabled status

### LINKED ISSUE #14658: "Run Code" button become unclickable when a console has been opened
## Description
图片
"Run Code" button become unclickable when a console has been opened.
I can only run the code cell by ctrl+enter or shift+enter.
Close the console and the "Run Code" button could be available again, but I want to click it with a console.
Is it on purposed?

## Reproduce
1. Run some code cells;
2. Right click the ipynb tab, "New Console for Notebook";
3. Return to ipynb tab and sometimes the "Run Code" button is unavailable.

## Expected behavior

The "Run Code" button is available when a console is opened.

## Context

- Operating System and version: Arch Linux
- Browser and version: Firefox 113.0.2
- JupyterLab version: 4.0.1

<pre>
</pre>
</details>

<pre>
</pre>
</details>

<pre>

</pre>
</details>

fcollonval:
## References

Fixes #14658

## Code changes

- Fix `isEnabled` for file editor semantic commands
- Switch notebook toolbar button to use the notebook commands; before by using a semantic command, if the focus switches in a console for example, some actions would still appear enabled
- Add caption to customized toolbar buttons to customized the button tooltip

## User-facing changes

Notebook toolbar buttons status are consistent


fcollonval:

krassowski:
> before by using a semantic command, if the focus switches in a console for example, some actions would still appear enabled

But is this the expected behaviour? Could these be enabled (rather than disabled) but possibly greyed out as "inactive"? My problem with the buttons being disabled is that it is now a two-step action to run a cell in another notebook (first focus the notebook, then click on the button).

two-step

Even though other UI elements e.g. the "add cell" button are always active:

but-adding-cell-works

I think most desktop UI does not require focusing first to click a button. As an example, the window initially in the background on the screencast has the buttons greyed out (inactive) but when I click on "X" it just works (it is not disabled):

window-buttons

krassowski:
Yes, I think this makes sense - subsequently we could apply a different kind of "grey out" style to all toolbar buttons by conditioning on the document not having an "active" class (we don't have a class like that on document widgets - only on tab bar, which is a related UX issue).

fcollonval:

fcollonval:

PR #14667, opened 2023-06-09
Document UX implications of windowed notebooks on search

ericsnekbytes:
Added notes for Jupyter's intelligent find plugin, and an acknowledgement of browser find caveats (same as docs additions for https://github.com/jupyter/notebook/pull/6905).


docs/source/user/interface.rst
```diff
@@ -161,6 +161,17 @@ Toggle Simple Interface mode using the View menu:
 When you leave Simple Interface mode, the original layout of the main
 area is restored.
 
+Searching
+---------
+
+JupyterLab has an intelligent built-in search plugin for finding text within a
+notebook or other document, which uses the `Ctrl-F` shortcut by default.
```
krassowski, line 168:
```suggestion
JupyterLab has a built-in search plugin enabling finding text within partially rendered
notebooks or other documents, which uses the `Ctrl-F` shortcut by default.
```

docs/source/user/interface.rst
```diff
@@ -161,6 +161,17 @@ Toggle Simple Interface mode using the View menu:
...
+
+Your browser's `find` function will give unexpected results because it doesn't have
+access to the full content of a document, but you can still use your browser find
+function from the browser menu if you want, or you can disable the intelligent search
```
krassowski, line 172:
```suggestion
function from the browser menu if you want, or you can disable the built-in document search
```

docs/source/user/interface.rst
```diff
@@ -161,6 +161,17 @@ Toggle Simple Interface mode using the View menu:
...
+shortcut using the advanced settings editor.
```
krassowski, line 173:
```suggestion
shortcut using the advanced settings editor, or disable the windowed notebook rendering to expose the full document to the browser (at expense of performance).
```

krassowski (commented):
This looks like a good addition. I would avoid 'intelligent' as this is not used in the UI and users may have very different expectations on what intelligent means nowadays. I also added a suggestion to mention disabling notebook virtualization as an alternative.

I am actually not sure if suggesting disabling the shortcut is a good idea as this would mean users have no way to search in CSV/tsv files rendered on canvas in datagrid. Ideally we could only disable it for notebook but keep for datagrids.

In general there are ways to improve browsers search experience if we used `content-visibility` with `hidden=until-found` instead of `display: none`, see https://developer.chrome.com/articles/hidden-until-found/ but sadly browser support is not there yet.

docs/source/user/interface.rst
ericsnekbytes: I merged this, but changed the wording to `special` instead of intelligent as I think it better communicates the purpose and function of the plugin. I omitted the partial rendering text as I think it hurts plain-language goals/readability by adding jargon and complexity/verbosity. Related info is still included later down in the text.

docs/source/user/interface.rst
ericsnekbytes: This has been incorporated in a separate commit.

ericsnekbytes:
@krassowski I incorporated the suggestions from the Notebook parity PR here, should be ready for you.

docs/source/user/interface.rst
```diff
@@ -161,6 +161,20 @@ Toggle Simple Interface mode using the View menu:
 When you leave Simple Interface mode, the original layout of the main
 area is restored.
 
+Searching
+---------
+
+JupyterLab has an advanced built-in search plugin for finding text within a
+notebook or other document, which uses the `Ctrl-F` shortcut by default.
```
fcollonval, line 168: In RST file, you need to use double tick to get the text highlighted in mono space and in other places the `+` seems to be used rather than `-` to point to combine keys shortcut:

```suggestion
notebook or other document, which uses the ``Ctrl+F`` (``Cmd+F`` for macOS) shortcut by default.
```

The shortcut is actually Cmd-F for Apple.

docs/source/user/interface.rst
```diff
@@ -161,6 +161,20 @@ Toggle Simple Interface mode using the View menu:
...
+
+Your browser's `find` function will give unexpected results because it doesn't have
```
fcollonval, line 170:
```suggestion
Your browser's ``find`` function will give unexpected results because it doesn't have
```

fcollonval (changes requested):
Thanks @ericsnekbytes 

I have some minor comments

ericsnekbytes:
@fcollonval The docs build looks good, do you have any bandwidth to merge this so I can resolve this one plus the docs parity PR for notebook (makes these areas the same for notebook and lab)?

fcollonval (approved):
Thanks @ericsnekbytes

fcollonval:
@krassowski I leave this one open in case you want to have another look.

ericsnekbytes:
@krassowski Can you merge this?

PR #14683, opened 2023-06-14
Propagate command changed signal in semantic commands.

fcollonval:
## References

Follow-up of #14664

This requires https://github.com/jupyterlab/jupyterlab/pull/15449 that has been extracted from this PR.

## Code changes

- Propagate command changed in semantic command
- ~~[x] Add signal `shell.currentChanged` to `JupyterFrontEnd.IShell`~~
- [x] Remove notifying all commands on LabShell layout modification

## User-facing changes

Small performance gain by avoiding noisy 'many-changed' command signals

None


afshin (approved):
Thank you!

packages/application/src/frontend.ts
```diff
@@ -276,6 +276,14 @@ export namespace JupyterFrontEnd {
      */
     readonly currentWidget: Widget | null;
 
+    /**
+     * A signal emitted when the focused widget in the application shell changes.
+     */
+    readonly currentChanged?: ISignal<
```
jtpio, line 282: Should this new field have the same `### Notes` docstring as above?

jtpio (changes requested):

packages/application/src/frontend.ts
jtpio: The top comment of the PR says:

> Small performance gain by avoiding noisy 'many-changed' command signals

Also is the performance gain of this change to the `JupyterFrontEnd.IShell` worth it? This interface has been quite stable since JupyterLab 1, so we should be careful (even though in this case it's optional and backwards compatible)

krassowski: This signal would be useful in a few other core plugins and in the extensions so I am in favour of adding it.

I agree that it needs better docs, in particular what is the expected behaviour for single widget apps such Notebook.

fcollonval: The doc string has been expanded in 528124e

packages/application/src/frontend.ts
```diff
@@ -276,6 +276,23 @@ export namespace JupyterFrontEnd {
...
 
+    /**
+     * A signal emitted when the focused widget in the application shell changes.
+     *
+     * ### Notes
+     * Shells may not always have a {@link currentWidget} or it may not change.
+     * Therefore implementing this signal is only expected for shell with the ability
+     * to switch between active widgets.
+     *
+     * Although the signal argument type mentionned a focus tracker, the shell
```
afshin, line 287:
```suggestion
     * Although the signal argument type mentioned a focus tracker, the shell
```

jtpio:
As discussed at the lab call today, moving the change that adds the `currentChanged` signal to the shell interface would help merge this PR more easily.

That way the fact that the `currentChanged` signal is added to the interface (as optional) will also be more visible on GitHub and in the changelog, so we can understand better the context and the motivations for adding it.

afshin:
I don't want to obstruct this, allowing the perfect to be the enemy of the good. @krassowski, do you have an opinion on this?

packages/application/src/frontend.ts
krassowski: FWIW noisy signals can cause issues beyond small performance impacts; for example in https://github.com/jupyterlab/jupyterlab/issues/15465#issuecomment-1989046706 I tracked down a browser crash caused by a rendering bottleneck caused by the use of `ActivateRequest` where `labShell.currentChanged` would suffice (but was not used presumably because it was not on `IShell`).

krassowski (approved):
I am in favour of moving forward with this one (and https://github.com/jupyterlab/jupyterlab/pull/15449 for that matter).

krassowski:
However, I would reserve the right to revert this and https://github.com/jupyterlab/jupyterlab/pull/15449 if we find during beta testing that this causes negative impact downstream.

afshin:
@jtpio are you okay with this one, too?

jtpio:
I think I'm neutral on this change.

Practically speaking, https://github.com/jupyterlab/jupyterlab/pull/15449 would have to be merged first so we keep a good PR history and make it easier to track changes.

During the beta cycle, we should indeed look into updating downstreams (extensions and Notebook 7) to double check https://github.com/jupyterlab/jupyterlab/pull/15449 does not cause issues.

packages/notebook-extension/src/index.ts
```diff
@@ -3560,6 +3561,17 @@ function addCommands(
...
     }
   });
+
+  // All commands with isEnabled defined directly or in a semantic commands
+  // To simplify here we added all commands as most of them have isEnabled
+  const skip = [CommandIDs.createNew];
+  const notify = () => {
+    Object.values(CommandIDs)
+      .filter(id => !skip.includes(id))
+      .forEach(id => app.commands.notifyCommandChanged(id));
```
jtpio, line 3571: This seems to be assuming that all commands registered by all the plugins will be available in a lab application.

For example in Notebook the command `notebook:create-output-view` is not registered, leading to an error in the dev tools console: https://github.com/jupyter/notebook/pull/7312#issuecomment-2039087409

jtpio: Shouldn't a plugin take care of notifying the command it registered instead of all commands in the `CommandIDs` namespace?

jtpio: Opened https://github.com/jupyterlab/jupyterlab/issues/16113 so this can be fixed before 4.2.0

PR #14694, opened 2023-06-15
Update branch configuration

fcollonval:
## References

This is a follow-up PR from branching out 4.0.x from main to fix the URLs and configuration.

## Code changes

none

## User-facing changes

None

## Backwards-incompatible changes

None

PR #14698, opened 2023-06-16
Ensure documentation switcher data always points to the latest JSON.

fcollonval:
## References

Follow-up of #14694

## Code changes

The documentation version switcher must always point to the latest version otherwise older versions won't list newer versions.

This breaks the inline URL so the integrity auto fix does not update that specific URL.

## User-facing changes

None

None

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -276,7 +276,11 @@ html_theme_options = {
     "navbar_start": ["navbar-logo", "version-switcher"],
     "footer_start": ["copyright.html"],
     "switcher": {
-        "json_url": "https://jupyterlab.readthedocs.io/en/latest/_static/switcher.json",
+        # Trick to get the documentation version switcher to always points to the latest version without being corrected by the integrity check;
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -276,7 +276,11 @@ def document_plugins_tokens_list(list_path: Path, output_path: Path) -> None:
     "navbar_start": ["navbar-logo", "version-switcher"],
     "footer_start": ["copyright.html"],
     "switcher": {
-        "json_url": "https://jupyterlab.readthedocs.io/en/latest/_static/switcher.json",
+        # Trick to get the documentation version switcher to always points to the latest version without being corrected by the integrity check;
```
jtpio, line 279: What was the integrity check doing with this URL?

jtpio: OK thanks for the extra context :+1:

jtpio (approved):
Thanks!

jtpio:

PR #14705, opened 2023-06-17
Fix HTML lang being set to 'default' resulting in blind users with screenreader reading everything in German

### LINKED ISSUE #14704: Blind users with screenreader read everything in German
- Copy of https://github.com/jupyter/notebook/issues/6924

Language

- Expectation: Language tags are specified on <html> element based on the language the content is being displayed as. If content is in English, needs to be <html lang="en">, etc. (WCAG Criteria 3.1.1 (A) and 3.1.2 (AA))
- Observed:
  - On all pages with my language set to English, lang="default" which reads my content in GERMAN (probably looking at the first 2 letters "de"). Need to have JS that powers this instead change to the proper lang codes.
  - If importing over the Settings Editor to Notebookv7 ever, recommend adding dropdown menu for languages so user only selects from list of languages that are currently supported/installed so bilingual blind users can easily toggle without fear of messing up.

Screenshot 2023-06-16 at 7 50 00 PM

Lang is getting set to `default` here:
https://github.com/jupyterlab/jupyterlab/blob/8fe49b14f82b5902d88b4f3edd2f0df484212f79/packages/translation-extension/src/index.ts#L87-L102

andrii-i:
## References

- Fixes #14704
- Fixes (transitively) https://github.com/jupyter/notebook/issues/6924

## Code changes
Ensure `currentLocale` is not `default` which is not a valid language code before assigning it to `document.documentElement.lang`

## User-facing changes
Language is not getting set to `default` (invalid language code) resulting in all screen readers detecting it as German, screen readers can detect language

Without this PR:
Screenshot 2023-06-16 at 7 50 00 PM

With this PR:
Screenshot 2023-06-16 at 7 50 32 PM

None

### DIFF
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -106,8 +106,7 @@ const langMenu: JupyterFrontEndPlugin<void> = {
             '_',
             '-'
           );
-        }
-        else {
+        } else {
           document.documentElement.lang = 'en-US';
         }
 
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -99,7 +99,14 @@ const langMenu: JupyterFrontEndPlugin<void> = {
       .then(setting => {
         // Read the settings
         loadSetting(setting);
-        document.documentElement.lang = (currentLocale ?? '').replace('_', '-');
+
+        // Ensure currentLocale is not 'default' which is not a valid language code
+        if (currentLocale !== 'default') {
+          document.documentElement.lang = (currentLocale ?? '').replace(
+            '_',
+            '-'
+          );
+        }
 
         // Listen for your plugin setting changes using Signal
         setting.changed.connect(loadSetting);
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -107,6 +107,9 @@ const langMenu: JupyterFrontEndPlugin<void> = {
             '-'
           );
         }
+        else {
+          document.documentElement.lang = 'en-US';
+        }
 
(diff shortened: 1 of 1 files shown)


packages/translation-extension/src/index.ts
```diff
@@ -99,7 +99,14 @@ const langMenu: JupyterFrontEndPlugin<void> = {
...
         loadSetting(setting);
-        document.documentElement.lang = (currentLocale ?? '').replace('_', '-');
+
+        // Ensure currentLocale is not 'default' which is not a valid language code
+        if (currentLocale !== 'default') {
+          document.documentElement.lang = (currentLocale ?? '').replace(
+            '_',
+            '-'
+          );
+        }
```
krassowski, line 109:
```suggestion
        }
        else {
          document.documentElement.lang = 'en-US';
        }
```

Thank you for working on this! Based on the [MDN recommendation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang):

> Note: The default value of lang is unknown, therefore it is recommended to always specify this attribute with the appropriate value.

I think we should set it to `en-US` if it is `default` rather than leaving it hanging.

andrii-i: That's a good point, thank you

krassowski (approved):
Thank you! This needs linting (the lint job is failing) but other than that looks good.

andrii-i:
Thank you @krassowski. 

Re-running failing visual regression test. It's not immediately evident how difference in screenshots relates to current change. At the same time I don't I only see this failure in this PR.

fcollonval:

andrii-i:
Thank you @fcollonval

Issue #14711, opened 2023-06-18
LSP follow-up checklist


krassowski:
This is a running checklist for things noted when porting jupyterlab-lsp extension:

- [x] there does not seem to be a way to add extension conditionally on editor type (see also https://github.com/jupyterlab/jupyterlab/issues/14710).
- [x] we could probably remove `packages/lsp/src/adapters/statusmessage.ts`
- [x] `ILSPDocumentConnectionManager` may need to expose:
  - `initialConfigurations`
  - `updateLogging`
- [x] the plugin providing `ILSPDocumentConnectionManager` should be separate from plugin providing settings/running tab integration
- [ ] `ILanguageServerManager` needs to expose `specs` → maybe we can just use `getMatchingSpecs()` with no args
- [ ] `VirtualDocuments` may need to expose `virtualLines` (currently protected) - or maybe not, maybe we just need a method which gets indices of cells covered by it (the only substantial donwstream usage is in `getBreadcrumbs`).
- [x] `ILSPConnection` needs to expose:
   - `serverCapabilities`
   - `provides()` (or remove it as redundant to `serverCapabilities`)

trungleduc:
Thanks @krassowski for working on the LSP. My lack of experience with the LSP extension has caused many regressions in the integration PR. I will keep following your work and please let me know if I can join the effort on some parts

krassowski:
It is difficult to get from CodeMirror's `EditorView` to `CodeEditor.IEditor` and `Document.IEditor`. The latter used to be synonymous with the  former but due to windowed notebook refactor an accesor wrapper was created as follows in https://github.com/jupyterlab/jupyterlab/pull/12554:

https://github.com/jupyterlab/jupyterlab/blob/42fb46470fb747ec1a8e6ad9a32ed11a0d4991a7/packages/lsp/src/tokens.ts#L179-L228

Worst case scenario, I could workaround this using the fact that the `factory` of `editorExtensionRegistry.addExtension` includes `CodeEditor.IModel`:

```ts
const editor = adapter.editors.find(e => e.model === options.model);
```

but neither `Document.ICodeBlockOptions` nor `Document.IEditor` expose the model.

1. Can we expose the model? If we did, we could remove `value` string which I think may not be needed on the public interface if the model is available. 
2. Is there a better way to get the `Document.IEditor` for an event invoked within CodeMirror extension?

Please note that using `adapter.activeEditor` is only appropriate for responding to user events such as typing or mouse click, because:
[...]

PR #14712, opened 2023-06-18
Expose `serverCapabilities`, `provides`, and  `updateLogging`

krassowski:
## References

Part one for #14711

## Code changes

- `ILSPDocumentConnectionManager` now exposes `updateLogging` along other `update*` methods
  - I do not expose `initialConfigurations` in this PR yet, I want to think about API more
  - this is needed to allow custom connection managers and/or settings forms
- `ILSPConnection` now exposes `serverCapabilities` and `provides`
   - `provides` is just a utility function but, pretty handy; we do not have to make a final call on it (it, as all other APIs is tagged as `@alpha`)
   - this is pretty fundamental—absolutely required to implement most LSP features

## User-facing changes

None

None

### DIFF
--- a/packages/lsp/src/tokens.ts
+++ b/packages/lsp/src/tokens.ts
@@ -412,6 +412,11 @@ export interface ILSPDocumentConnectionManager {
   updateConfiguration(allServerSettings: TLanguageServerConfigurations): void;
 
   /**
+   * Enable or disable the logging of language server communication.
+   */
+  updateLogging(logAllCommunication: boolean, setTrace: lsp.TraceValues): void;
+
+  /**
    * Handles the settings that the language servers accept using
    * `onDidChangeConfiguration` messages, which should be passed under
    * the "serverSettings" keyword in the setting registry.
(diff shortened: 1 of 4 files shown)


packages/lsp/src/tokens.ts
```diff
@@ -851,6 +856,20 @@ export interface ILSPConnection extends ILspConnection, IObservableDisposable {
...
+   * Check if a capability is available in the server capabilities.
+   */
+  provides(capability: keyof lsp.ServerCapabilities): boolean;
+
+  /**
+   * @alpha
+   *
+   * Lists server capabilities.
+   */
+  serverCapabilities: lsp.ServerCapabilities;
```
fcollonval, line 871: Should it be `readonly`?

krassowski: I asked myself the same question. Possibly yes. But none of the APIs which are read-only in implementation are marked as such in the interface. I am not sure why.

fcollonval (approved):
Thanks @krassowski

fcollonval:

PR #14713, opened 2023-06-18
Drop `StatusMessage` which is dead code

krassowski:
## References

Part two for #14711

## Code changes

Removes `StatusMessage` which is not used in core and will not be used downstream in LSP extension since a better alternative (`Notification`) is now available for statuses which were previously (poorly) communicated on status bar.

## User-facing changes

None

This class was not documented and we can say it was not a part of public API.


fcollonval (commented):
Candid question, is the LSP API allowed to be changed within a Lab major version? I forget what was decided.

krassowski:
Most of the public LSP APIs are tagged with alpha/experimental so the answer I would give: yes if tagged as such. But in this case I think this was never a part of the API - it is not exported from `@jupyterlab/lsp`, undocumented, and really not relevant to LSP itself - so I would suggest we remove it before someone starts using it. Unless I am missing something.

fcollonval (approved):
Thanks @krassowski 

I guess it will be better to backport it to 4.0.x to limit the risk of use?

fcollonval:

PR #14721, opened 2023-06-20
Fix kernel message deserialization

### LINKED ISSUE #14720: All comm messages contain empty DataView in buffers
Since https://github.com/jupyterlab/jupyterlab/pull/11841 all Comm messages that are received inside JupyterLab seem to have an empty `DataView` appended to the `buffers`:

Screen Shot 2023-06-20 at 09 59 12

Sadly my minimal example isn't so minimal and relies on our `pyviz_comms` library to create a proxy for the `CommManager`:

```python
import uuid

from pyviz_comms import JupyterCommManager

class BidirectionalComm:

    manager = JupyterCommManager()
    
    _EXEC_MIME_TYPE = 'application/vnd.holoviews_exec.v0+json'
    
    _loaded = False

    def __init__(self, msg_handler=None):
        self._received = []
        self._uuid = uuid.uuid4().hex
        if not self._loaded:
            # Initialize the window.PyViz.comm_manager
            display({'application/javascript': self.manager.js_manager}, raw=True)
            BidirectionalComm._loaded = True
            
        # Initialize the server and client comms
        # server: Sends messages Python -> JS
        # client: Sends messages JS -> Python (and then receives ACK message when server completes processing)
        self._server_comm = self.manager.get_server_comm()
        self._client_comm = self.manager.get_client_comm(
            on_msg=msg_handler
        )
        
    def send(self, msg, metadata=None, buffers=[]):
        """
        Sends message to JS including metadata and bytes buffers
        """
        self._server_comm.send(msg)
    
    def _repr_mimebundle_(self, include=None, exclude=None):
        """
        Basic example of connecting to comm_manager, rendering received output and sending button events to Python. 
        """
        html = f"""
        <script>
        window.PyViz.comm_manager.register_target(
            {self._uuid!r}, {self._server_comm.id!r}, (msg) => {{
              console.log(msg)
              document.getElementById({self._server_comm.id!r}).textContent += msg.content.data
            }}
        )

        var client_comm = window.PyViz.comm_manager.get_client_comm(
            {self._uuid!r}, {self._client_comm.id!r}, (msg) => console.log('Server ACK')
        );
        
        var clicks = 0
        function send_click() {{
          clicks += 1
          client_comm.send({{'clicks': clicks}})
        }}
        </script>
        
        <button onclick="send_click()">Click me!</button>
        <div id="{self._server_comm.id}"></div>
        """
        mime_bundle = {
            self._EXEC_MIME_TYPE: '', # This allows the Jupyter(Lab) renderers to intercept this message
            'text/html': html
        }
        metadata = {self._EXEC_MIME_TYPE: {'id': self._uuid}}
        return mime_bundle, metadata

test = []
comm = BidirectionalComm(msg_handler=test.append)

comm.send('Foo')
```

Ping @davidbrochart since you were the last to touch that code.

davidbrochart:
## References

Closes #14720

## Code changes

Prior to this change, an empty buffer could be inserted.
This PR checks if the buffer is empty before inserting it.

## User-facing changes

None (but fixes a bug).

None.

### DIFF
--- a/packages/services/src/kernel/serialize.ts
+++ b/packages/services/src/kernel/serialize.ts
@@ -74,14 +74,12 @@ namespace Private {
     );
     let buffers = [];
     for (let i = 5; i < offsets.length; i++) {
-      let buffer: DataView;
       if (i == offsets.length - 1) {
-        buffer = new DataView(binMsg.slice(offsets[i]));
+        if (offsets[i] != binMsg.byteLength) {
+          buffers.push(new DataView(binMsg.slice(offsets[i])));
+        }
       } else {
-        buffer = new DataView(binMsg.slice(offsets[i], offsets[i + 1]));
-      }
-      if (buffer.byteLength > 0) {
-        buffers.push(buffer);
+        buffers.push(new DataView(binMsg.slice(offsets[i], offsets[i + 1])));
       }
     }
     msg = {
--- a/packages/services/src/kernel/serialize.ts
+++ b/packages/services/src/kernel/serialize.ts
@@ -74,10 +74,14 @@ namespace Private {
     );
     let buffers = [];
     for (let i = 5; i < offsets.length; i++) {
+      let buffer: DataView;
       if (i == offsets.length - 1) {
-        buffers.push(new DataView(binMsg.slice(offsets[i])));
+        buffer = new DataView(binMsg.slice(offsets[i]));
       } else {
-        buffers.push(new DataView(binMsg.slice(offsets[i], offsets[i + 1])));
+        buffer = new DataView(binMsg.slice(offsets[i], offsets[i + 1]));
+      }
+      if (buffer.byteLength > 0) {
+        buffers.push(buffer);
       }
     }
     msg = {
--- a/packages/services/src/kernel/serialize.ts
+++ b/packages/services/src/kernel/serialize.ts
@@ -73,14 +73,8 @@ namespace Private {
       decoder.decode(binMsg.slice(offsets[4], offsets[5]))
     );
     let buffers = [];
-    for (let i = 5; i < offsets.length; i++) {
-      if (i == offsets.length - 1) {
-        if (offsets[i] != binMsg.byteLength) {
-          buffers.push(new DataView(binMsg.slice(offsets[i])));
-        }
-      } else {
-        buffers.push(new DataView(binMsg.slice(offsets[i], offsets[i + 1])));
-      }
+    for (let i = 5; i < offsets.length - 1; i++) {
+      buffers.push(new DataView(binMsg.slice(offsets[i], offsets[i + 1])));
     }
     msg = {
       channel,


packages/services/src/kernel/serialize.ts
```diff
@@ -74,10 +74,14 @@ namespace Private {
     );
     let buffers = [];
     for (let i = 5; i < offsets.length; i++) {
+      let buffer: DataView;
       if (i == offsets.length - 1) {
-        buffers.push(new DataView(binMsg.slice(offsets[i])));
+        buffer = new DataView(binMsg.slice(offsets[i]));
       } else {
-        buffers.push(new DataView(binMsg.slice(offsets[i], offsets[i + 1])));
+        buffer = new DataView(binMsg.slice(offsets[i], offsets[i + 1]));
```
davidbrochart, line 81: Is it worth not even creating the `DataView`? We could compare the last offset with the message length instead.

packages/services/src/kernel/serialize.ts
```diff
@@ -74,10 +74,14 @@ namespace Private {
...
+      }
+      if (buffer.byteLength > 0) {
```
philippjfr, line 83: Is this safe? Seems like you could in fact push empty arrays (for whatever reason) and those should be deserialized correctly. In Panel I tried a test case where I pushed an empty array and that (correctly) got deserialized as a zero-length DataView (which appeared in addition to the bogus additional `DataView`).

philippjfr: Here's what I mean, this is a message where I've sent an empty array:

Screen Shot 2023-06-20 at 14 32 49

You can see that we get two `DataView` with length 0, one being the real empty array and then the bogus empty `DataView` that is always added.

davidbrochart: Yes I can see a need for inserting empty buffers. Could you try this PR with your Panel example?

philippjfr: Having some trouble and about to be in meetings for a few hours. `jupyter lab build` fails without any indication of what went wrong.

davidbrochart: Maybe run `jlpm run build`?

echarles: ... and `jlpm install` to install the npm deps.

https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#setting-up-a-local-development-environment should contain the ordered list of commands to run, which is documented on https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#installing-jupyterlab

```
git clone https://github.com/<your-github-username>/jupyterlab.git
cd jupyterlab
pip install -e ".[dev,test]"
jlpm install
jlpm run build  # Build the dev mode assets (optional)
```

I haven't tried those simple commands in a fresh env (cc/ @ericsnekbytes)

davidbrochart: Actually, according to [this comment](https://github.com/jupyter-server/jupyter_server/pull/657#issuecomment-1021202444):
> `offsetN == len(binMsg)` (i.e., the last offset is the length of the file, so that the last buffer is `binMsg[offset[N-1], offset[N]]`?

I don't think we need to process the last offset at all. That's what I did in https://github.com/jupyterlab/jupyterlab/pull/14721/commits/ce3c98aa3d6ed5b119d7d25aa6e709c2b3e337b4.
Maybe @jasongrout can confirm?

philippjfr: That was my feeling too, it seems like the simplest fix. In all the messages I've manually stepped through it also would have produced the correct result. For example in my screenshot above where I sent an empty array you can see that the accumulated offsets ends with `..., 547, 547]` i.e. it correctly records the offset for the empty array and then adds an extra one that should not be processed.

jasongrout: > Maybe @jasongrout can confirm?

I think that is right (see also https://github.com/jupyter-server/jupyter_server/pull/657#issuecomment-1021406374). IIRC, the reason we made the last offset the length of the message is so that the decoding would be very simple, and it provides a sanity check on the message size, and lets us possibly include content after the last buffer in the future.

jasongrout (approved):
LGTM. Thanks!

fcollonval (approved):
Thanks @davidbrochart 

> CI failures are not related

fcollonval:

Issue #12318, opened 2021-12-22
Add support for custom.css


yhslai:
## Description

I'm using Jupyterlab Desktop on Windows. I add a `custom.css` in `%AppData%\Roaming\jupyterlab-desktop\custom\custom.css`, but it doesn't take effect at all.

## Reproduce

1. Create `%AppData%\Roaming\jupyterlab-desktop\custom\custom.css`
2. Add styles

## Expected behavior

Styles are applied to Jupyterlab Desktop.

## Context

- Operating System and version: Windows 10
- JupyterLab-Desktop version: 3.2.5-2

jtpio:
Adding to the `4.1` milestone for consideration.

As discussed in https://github.com/jupyterlab/jupyterlab/issues/8457#issuecomment-1598791146, it could be interesting to have support for loading custom.css in JupyterLab 4.1.

For reference this was implemented in Notebook 7 in https://github.com/jupyter/notebook/pull/6841. Likely there is a good chunk of the code that could be moved to JupyterLab directly.

Issue #14319, opened 2023-04-03
Set custom CSS, like you could in a Jupyter notebook


cormullion:
In a Jupyter notebook, you could specify a font by adding some CSS to a file `~/.jupyter/custom/custom.css`. This allowed users to specify a font such as JuliaMono that could correctly handle special characters, and the use of packages such as UnicodePlots.jl.

The default font (Menlo) doesn't have the necessary glyphs, and the replacement glyphs (from Apple Braille) aren't the right width:

![26cdd156e807b76e076c6e79b7fc58e0a785c30e_2_579x500](https://user-images.githubusercontent.com/52289/229461270-f66e8d3e-7c65-43df-89ee-3bc14888f43c.png)

It should be easy to specify the font family using CSS, allowing the display to be rendered correctly, as here in Jupyter notebook.

![3926ddf4c1bd2caecf2eb3b880ad9d2b41fedc71_2_690x311](https://user-images.githubusercontent.com/52289/229461943-265c1624-4b88-4d3b-9eaf-bffc46fdab4b.png)

cormullion:
I didn't try that - I couldn't see where to insert the:

```css
@font-face {
    font-family: JuliaMono-Light;
    src: url("https://cdn.jsdelivr.net/gh/cormullion/juliamono/webfonts/JuliaMono-Light.woff2");
}
```
or similar CSS instructions...

Issue #8457, opened 2020-05-22
Feature suggestion: visually distinguish color cells with a tag


SultanOrazbayev:
Right now, to infer whether a cell has a tag one needs to open the sidebar and go over the cells that have associated tags. It would be great if the cells that have a tag are visually distinct (e.g. different cell background, different cell border size/color).

If there is a way to achieve this already, then kindly let me know.

telamonian:
I like this idea. +1 to the concept of being able to set a color on a particular tag, and then somehow marking tagged cells with that same color

ricwillis98:
I'd like to work on this

Should the color be specified with a text input? A drop down?

I think a different cell border would be good for visually distinguishing them.

jtpio:
For reference, a similar idea was recently shared on Discourse: https://discourse.jupyter.org/t/styling-cells-through-metadata/4978

With classic notebook extensions:

- https://github.com/innovationOUtside/nb_cell_execution_status
- https://github.com/innovationOUtside/nb_extension_tagstyler

ricwillis98:
What should happen when multiple tags with associated colors are added to the cell? Just pick one for now?

SultanOrazbayev:
Thanks, @ricwillis98, this is looking promising! 

For multiple colors, I like telamonian's suggestion in your PR (having multiple 'stripes'/mini-badges). Your initial suggestion also looks good to me, in case of multiple tags, it's possible to introduce stripes to the left side as in the attached image... but any visual distinction would be great.

echarles:
Would be great to reuse the jupyterlab color palette defined in the CSS for more homogenous look and fell.

ricwillis98:
@echarles good suggestion, it's a bit of a work around but you can by specifying the color as you would in css e.g. tag1;var(--jp-brand-color0)

This is how it looks in my current version

![image](https://user-images.githubusercontent.com/62662248/86282734-6b27f480-bba5-11ea-809a-1068bb556a8e.png)

SultanOrazbayev:
The original use case for me is the "parameters" tag used by `papermill`. It's not always obvious whether a notebook has it, and so it's great to be able to see it right away. However, this adds a constraint of what the value of the tag can be (`parameters`). I understand that this will be implemented in a later version, but thought I'd mention it just in case.

Thanks for your work, @ricwillis98!

echarles:
@ricwillis98 Great workaround!

jasongrout:
[...]

PR #14743, opened 2023-06-23
Load custom CSS functionality and documentation

RRosio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/12318
Related to https://github.com/jupyterlab/jupyterlab/issues/8457

## Code changes

The code adds a link to the custom CSS file in the template html page and a handler to either serve the user's local custom CSS file or the placeholder file. There is also a flag added to allow disabling the default behavior that loads custom CSS.

## User-facing changes

Users can apply custom styling. I tested this PR by adding some color to the Toolbar etc.

jupyter_styles_custom_css

None applicable that I can think of.


fcollonval (changes requested):
Thanks for upstreaming this @RRosio 

This is a great start but looking at the classical notebook code, I have a couple of questions.

> I also think this resolves #14319 but not fully #8457 (that one will requires also adding data attributes on cells).

docs/source/user/custom_css.rst
```diff
@@ -0,0 +1,105 @@
...
+  }
+  .lm-MenuBar-content {
+      color: #02484d
+  }
+
+  .lm-TabBar-content, .jp-DirListing-content, .lm-MenuBar-content {
+      font-size: small;
+  }
+
+.. figure:: https://user-images.githubusercontent.com/12378147/248197127-7e825096-3394-438f-9136-be4cc5c5cf01.png
```
krassowski, line 50: It would be great if we could generate these figures using playwright (galata) rather than rely on githubusercontent.com domain. It would also ensure that no browser UI is captured and that we are actually testing the integration works

jupyterlab/labapp.py
```diff
@@ -573,6 +577,14 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
         help="Whether to expose the global app instance to browser via window.jupyterapp",
     )
 
+    custom_css = Bool(
+        True,
```
krassowski, line 581: Should the default for 4.1 be `True` or `False`? I would not ask this question if it was aimed at 5.0. I have no strong opinion, but believe it is important to discuss.

jtpio: Maybe `False` in JupyterLab to be on the conservative side? Users can opt-in easily since there is documentation that explains how to do so.

jupyterlab/labapp.py
```diff
@@ -707,6 +724,8 @@ def initialize_handlers(self):  # noqa
         self.log.info("JupyterLab extension loaded from %s" % HERE)
         self.log.info("JupyterLab application directory is %s" % self.app_dir)
 
+        handlers.append(("/custom/custom.css", CustomCssHandler))
```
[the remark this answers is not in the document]
krassowski: Discussed at jupyterlab call, we could make addition of handlers conditional on whether this feature is enabled to limit impact

fcollonval (commented):
Thanks @RRosio 

To make things easier, I pushed two commits to
- Fix the galata tests
- Use those to produce the documentation
- Fix the Python test
- Switch the default behavior to not load custom.css

To explain how I did the integration test,
1. I'm activating the feature when starting the server only for the documentation test through the command line option.
1. I capture the `custom/custom.css` request to return a mocked content.

fcollonval (approved):

krassowski:
```
    [documentation] › test/documentation/custom_css.test.ts:15:7 › Use custom CSS layout › should apply custom CSS to main page 
    [documentation] › test/documentation/custom_css.test.ts:33:7 › Use custom CSS layout › should apply custom CSS to h1 headings 
```

The failures seem related.

fcollonval:
> Would it be okay to update the branch then?

Now that Mike updated manually the snapshots it should not be needed.

krassowski:
Now that you said it, I decided to merge main anyways just to be sure.

galata/test/documentation/custom_css.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      })
+    );
+    const fileName = 'markdownCells.ipynb';
+    await page.contents.uploadFile(
+      path.resolve(__dirname, `./data/${fileName}`),
+      `${tmpPath}/${fileName}`
+    );
+    await page.goto();
+
+    await page.notebook.openByPath(`${tmpPath}/${fileName}`);
```
krassowski, line 47:
```suggestion
    await page.notebook.openByPath(`${tmpPath}/${fileName}`);
    await page.notebook.activate(fileName);
```

krassowski:
Remaining failures are unrelated, merging - thanks again!

Issue #14668, opened 2023-06-10
indentation problem after multi-line Python function definition


mbakker7:
When I create a multi-line function definition in JupyterLab 4 (4.0.0), the indentation is messed up. So when I do a return, the cursor doesn't go to the right place, but always at about 13 spaces or something. This problems occurs on both Chrome and Safari on a Mac.
The simple example below shows what I mean. I wonder if others see the same problem.

Example:
```
def test(x,
         y): # after giving return here, the cursor doesn't go to the correct 4 spaces indentation
    print(x) # even after manually positioning the cursor in the right place, it goes to the wrong place after every return
    print(y) # still have a problem and had to manually position the cursor
```

The problem goes away when the function definition is on one line. For this case that would make sense, of course, but I ran into this for a function with a long argument list. The problem goes away when you finish the function by manually positioning the cursor back to position 0, even when omitting the `return` at the end of the function. But inside the function (when editing the function for example), the problem remains. Feels like a CodeMirror problem to me. Or is there something in Settings that I can adjust?

krassowski:
I can reproduce this in [CodeMirror 6 playground](https://codemirror.net/try/?c=aW1wb3J0IHtiYXNpY1NldHVwLCBFZGl0b3JWaWV3fSBmcm9tICJjb2RlbWlycm9yIgppbXBvcnQge3B5dGhvbn0gZnJvbSAiQGNvZGVtaXJyb3IvbGFuZy1weXRob24iCgpuZXcgRWRpdG9yVmlldyh7CiAgZG9jOiAiZGVmIHRlc3QoeCxcbiAgICAgICAgIHkpOlxuIiwKICBleHRlbnNpb25zOiBbYmFzaWNTZXR1cCwgcHl0aG9uKCldLAogIHBhcmVudDogZG9jdW1lbnQuYm9keQp9KQo=).

This also affects `if` and other statements:

```python
if (x
    and y):  # after pressing enter the indent will be deeper than expected.
```

I will open an issue on CodeMirror repo.

krassowski:
Upstream issue: https://github.com/codemirror/dev/issues/1182.

If someone has more time to contribute upstream, the code for indentation is here: https://github.com/codemirror/lang-python/blob/main/src/python.ts

mbakker7:
Thanks for submitting the issue in the right place. 
Not something I have any knowledge about, so unfortunately I cannot help.

mbakker7:
[...]

PR #14748, opened 2023-06-24
Fix indentation problem after multi-line Python function definition

FoSuCloud:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14668

## Code changes

Use resolutions to limit the @codemirror/lang-python version
image

Execute "yarn install" and "jlpm build" to update dependencies
image

## User-facing changes

nothing

nothing

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,6 @@
     "watch:packages": "python scripts/watch_packages.py"
   },
   "resolutions": {
-    "@codemirror/lang-python": "^6.1.3",
     "@jupyterlab/services/ws": "8.12.0",
     "@types/react": "^18.0.26",
     "react": "^18.2.0",
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "watch:packages": "python scripts/watch_packages.py"
   },
   "resolutions": {
+    "@codemirror/lang-python": "^6.1.3",
     "@jupyterlab/services/ws": "8.12.0",
     "@types/react": "^18.0.26",
     "react": "^18.2.0",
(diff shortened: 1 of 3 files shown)


FoSuCloud:
@meeseeksdev tag bug

package.json
```diff
@@ -106,6 +106,7 @@
     "watch:packages": "python scripts/watch_packages.py"
   },
   "resolutions": {
+    "@codemirror/lang-python": "^6.1.3",
```
fcollonval, line 109: We should avoid using `resolutions` as much as possible. In this case, could you upgrade the minimal dependency version in [jupyterlab/packages/codemirror/package.json](https://github.com/jupyterlab/jupyterlab/blob/f699d32d19e1687bf27e4c83318f50062adb0c5a/packages/codemirror/package.json#L51) to `^6.1.3`.

```suggestion
```

> For information, if you want to upgrade a third-party dependency in the lock file without changing any package.json, you should use the yarn command [yarn up](https://yarnpkg.com/cli/up).

fcollonval (changes requested):
thanks @FoSuCloud 

Could you update the specification a bit differently (see below comment)?

> GitHub tip: if you use [special keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) like _Fixes_ in front of the issue in the PR description, the issue will be linked with this PR easing documentation and repository management :wink:

package.json
FoSuCloud: ok, i'll fix it tonight

krassowski (approved):
Thank you @FoSuCloud

krassowski:

Issue #14765, opened 2023-06-28
Implement ARIA recommendations from Notebooks for All


gabalafou:
The [Notebooks for All](https://github.com/Iota-School/notebooks-for-all) project did research into ways to improve notebook HTML markup for disabled users. 

@tonyfast put out a [specification to make notebook markup more accessible](https://tonyfast.github.io/tonyfast/xxiii/2023-06-21-v7-aom.html).

This issue proposes modifying the @jupyterlab/notebook package so that the HTML it outputs matches those recommendations.

This will affect the HTML markup in both:

- JupyterLab
- Notebook 7

@afshin and @gabalafou (me) have volunteered to work on a PR for this issue.

tonyfast:
i've placed my recommendations in https://github.com/jupyter/notebook/issues/6956 . it is quite a long tedious issue. 

i do want to disclose that i've removed the feed pattern from the interactive version and we've assigned cells as region landmarks

PR #14768, opened 2023-06-28
Improve Notebook ARIA

gabalafou:
## References

This PR is a place to collaboratively implement the corresponding issue:

- https://github.com/jupyterlab/jupyterlab/issues/14765

## Code changes
- [ ]  https://github.com/jupyterlab/jupyterlab/issues/14804

Work in progress, to be determined.

## User-facing changes

Work in progress, to be determined.

Work in progress, to be determined.

### DIFF
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -679,6 +679,9 @@ export class WindowedList<
       document.createElement('div')
     );
     windowContainer.className = 'jp-WindowedPanel-window';
+    if (options.windowContainerCallback) {
+      options.windowContainerCallback(windowContainer);
+    }
     super({ node });
     super.layout = options.layout ?? new WindowedLayout();
     this._viewModel = options.model;
@@ -1449,6 +1452,11 @@ export namespace WindowedList {
      * Windowed list layout
      */
     layout?: WindowedLayout;
+    /**
+     * Callback to allow subclasses to set up/modify the window container
+     * during instantiation.
+     */
+    windowContainerCallback?: (windowContainer: Element) => void;
   }
 
   /**
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -201,7 +201,12 @@ export class StaticNotebook extends WindowedList {
           (options.notebookConfig?.windowingMode ??
             StaticNotebook.defaultNotebookConfig.windowingMode) === 'full'
       }),
-      layout: new NotebookWindowedLayout()
+      layout: new NotebookWindowedLayout(),
+      windowContainerCallback: (windowContainer: Element) => {
+        windowContainer.setAttribute('role', 'feed');
+        // TODO translation
+        windowContainer.setAttribute('aria-label', 'Cells');
+      }
     });
     this.addClass(NB_CLASS);
     this.cellsArray = cells;


packages/ui-components/src/components/windowedlist.ts
```diff
@@ -679,6 +679,9 @@ export class WindowedList<
       document.createElement('div')
     );
     windowContainer.className = 'jp-WindowedPanel-window';
+    if (options.windowContainerCallback) {
+      options.windowContainerCallback(windowContainer);
```
gabalafou, line 683: Something about this doesn't feel quite right to me but I can't put my finger on it. Many other widgets rely on a separate renderer that can be substituted and/or subclassed. Would it be worth trying to implement that pattern for the `WindowedList` class?

gabalafou: cc @afshin. 

Or some other altogether different approach?

tonyfast: i felt the same way. this is a recent feature that was added to include an "add cell" dialog in the empty space of the notebook footer. i think for our situation we consider the this feature to be part of a `contentinfo` with a `toolbar` inside it. it can probably be refactored, but proper aria might do a better job.

fcollonval: The renderer pattern would be the best pattern indeed.

You should add in `WindowedList.IOptions<T>` a `renderer: WindowedList.IRenderer` with

```ts
interface IRenderer {
  createWindow(): HTMLElement;
}
```

With default

```ts
class Renderer implements IRenderer {
  createWindow(): HTMLElement {
    return document.createElement('div');
  }
}
```

The constructor will become:

```diff
    const renderer = options.renderer ?? new WindowList.Renderer();
    const windowContainer = innerElement.appendChild(
      renderer.createWindow()
    );
    windowContainer.className = 'jp-WindowedPanel-window';
```

---

@tonyfast I'm not sure to understand your comment. The window container should be seen as the view port of the virtual list. The inner and outer elements are there for handling the scrolling behavior - all of those are needed for the windowed rendering. This was not added for handling a footer for the _add cell_ button.

tonyfast: > The window container should be seen as the view port of the virtual list. The inner and outer elements are there for handling the scrolling behavior - all of those are needed for the windowed rendering. This was not added for handling a footer for the _add cell_ button.

the visual and audible behaviors of this component are different. virtual scrolling can be challenging for screen readers users if there are actions following the feed. in a presumably infinite list, they may never reach the controls at the bottom. the "click to add cell" button should be in a footer with a label so that a screen reader can access the landmark independent of scrolling. it would be a better assistive experience to have the Notebook controls in a header and then trailing controls in a footer.

afshin: I created an issue for this and will write a PR in the coming week:

https://github.com/jupyterlab/jupyterlab/issues/14804

gabalafou:
[@tonyfast on the suitability of using role=grid](https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1760303516)

krassowski:
Provisionally tagging this for 4.2, but feel free to change milestone as you see it fit @gabalafou.

krassowski:
@gabalafou what is the status of this PR?

tonyfast:
i was going to make some suggestions, but there needs to be a little more work to complete the screen reader improvements. i'm not sure this is the pattern to recommend at the moment. i didn't realize when i first made the suggestion for the feed pattern that it is not well supported. recent work with notebooks for all and nbconvert a11y has established lists as being a good default, but, we also, allow screen readers to change their navigation preferences. basically, lab's best implementation would be one that is configurable from the list, landmark, table, grid, and treegrid navigation patterns. 

the first implementation should be the lists. screen reader users could navigate between notebook documents using `L` list navigation then they may access navigate cells with `I`. since there could be multiple notebooks on the page we'll need to label each list with the specific notebook name to remove ambiguities,.

if anyone wants to spend a hour or two working with me we could have a good draft of this effort together. let me know if anyone is interested in a productive working session to nail down screen navigation.

gabalafou:
Responding to the review of the [4.3.0 milestone](https://github.com/jupyterlab/jupyterlab/milestone/78) in today's Jupyter Frontends call...

This work has eventually gone stale, and as @tonyfast pointed out, we will need to take a different approach, so I am closing without merging.

PR #14770, opened 2023-06-29
Relax `Event.IManager` interface

trungleduc:
## References

The current `Event.IManager` interface is deducted from the `EventManager ` class. This enforces that any implementation of `Event.IManager` needs to extend `EventManager`. This PR relaxes this constraint.

cc @afshin if this constraint is intended. 

## Code changes

`EventManager` is now implementing `Event.IManager`

## User-facing changes

N/A

N/A since this is an interface relaxation.

### DIFF
--- a/packages/services/src/event/index.ts
+++ b/packages/services/src/event/index.ts
@@ -173,7 +173,7 @@ export namespace Event {
     /**
      * An event stream that emits and yields each new event.
      */
-    stream: Event.Stream;
+    readonly stream: Event.Stream;
     /**
      * Post an event request to be emitted by the event bus.
      */
--- a/packages/services/src/event/index.ts
+++ b/packages/services/src/event/index.ts
@@ -16,7 +16,7 @@ const SERVICE_EVENTS_URL = 'api/events';
 /**
  * The events API service manager.
  */
-export class EventManager implements IDisposable {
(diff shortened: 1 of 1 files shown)


afshin (approved):
Thanks, @trungleduc! 👍

packages/services/src/event/index.ts
```diff
@@ -165,5 +165,18 @@ export namespace Event {
...
-  export interface IManager extends EventManager {}
+  export interface IManager extends IDisposable {
+    /**
+     * The server settings used to make API requests.
+     */
+    readonly serverSettings: ServerConnection.ISettings;
+    /**
+     * An event stream that emits and yields each new event.
+     */
+    stream: Event.Stream;
```
fcollonval, line 176: Would you mind making this one read only?

```suggestion
    readonly stream: Event.Stream;
```

fcollonval (changes requested):
Thanks @trungleduc 

I have a suggestion. But otherwise it looks good.

packages/services/src/event/index.ts
trungleduc: Thanks Fred!

fcollonval (approved):
Thanks @trungleduc 

Backporting it as it is part of a 4.0.x new API.

Issue #13826, opened 2023-01-20
Show in Table of Contents where error occurred


astrofrog:
### Problem

When I run a very long notebook in Jupyter lab it's nice to have the visual representation in the table of contents of which section is being run:

![Screenshot from 2023-01-20 09-58-24](https://user-images.githubusercontent.com/314716/213668219-cf338d7e-adea-4d93-a912-3c41a03e418c.png)

However if I am not looking at the table of contents and an error occurs, all the indicators disappear and I have to scroll through the whole notebook to find where it failed.

### Proposed Solution

If an error occurs, an indicator such as :warning: or :exclamation:  could be shown next to the section where the empty/filled circles were shown before.

krassowski:
Related issues:
- https://github.com/jupyterlab/jupyterlab/issues/9786
- https://github.com/jupyterlab/jupyterlab/issues/5790
- https://github.com/jupyterlab/jupyterlab/issues/8714

fcollonval:
Closing as #14784 brings that feature

Issue #14604, opened 2023-05-27
Table of contents "waiting..." open-circle indicators never turn solid + do not go away after cell finishes running


lukelbd:
## Description

Lately, I have been experiencing a strange + intermittent issue with the table of contents. The open-circle indicators used to indicate sections with cells that are pending execution (and waiting for the completion of other cells) do not turn into filled-circles when they start running (as they are supposed do/as they did previously), and the open-circles never go away after the cell has finished. Might be tricky to reproduce. Here's a screenshot of what this looks like (note that nothing is running in this notebook).

<img width="262" alt="Screenshot 2023-05-27 at 13 35 27" src="https://github.com/jupyterlab/jupyterlab/assets/19657652/2ba2e0e9-3688-4013-85d1-e547bb402d93">

## Reproduce

1. Create a notebook with multiple sections
2. Add time-consuming processes to each section (e.g. `import time; time.sleep(10)`).
3. Open the table of contents pane
4. Run every cell in the notebook

## Expected behavior

Open circles turn solid when the cell is actually running (not waiting), and circles disappear when finished.

## Context

- Operating System and version: Linux CentOS 7.9.2009 
- Browser and version: JupyterLab Desktop 3.6.2-1 
- JupyterLab version: 3.6.3 

<details><summary>Command Line Output</summary>

```sh
[I 2023-05-26 14:57:13.634 ServerApp] Starting buffering for 2dd0cc1e-7740-41e2-ad50-2f227988727f:12f93ea8-a8cb-47f6-8774-9b60631c7bd8
[W 2023-05-26 14:57:13.690 ServerApp] Replacing stale connection: 2dd0cc1e-7740-41e2-ad50-2f227988727f:12f93ea8-a8cb-47f6-8774-9b60631c7bd8
[E 2023-05-26 14:57:13.790 ServerApp] Uncaught exception GET /api/kernels/2dd0cc1e-7740-41e2-ad50-2f227988727f/channels?session_id=12f93ea8-a8cb-47f6-8774-9b60631c7bd8 (::1)
    HTTPServerRequest(protocol='http', host='localhost:3002', method='GET', uri='/api/kernels/2dd0cc1e-7740-41e2-ad50-2f227988727f/channels?session_id=12f93ea8-a8cb-47f6-8774-9b60631c7bd8', version='HTTP/1.1', remote_ip='::1')
    Traceback (most recent call last):
[...]

PR #14784, opened 2023-07-02
Add error indicator in Table of Contents

skyetim:
## References

#13826 

## Code changes

The existing indicator flow is: 
1. when the cell is running, the status changes from idle to running; when it's scheduled, from idle to scheduled;
2. when the execution is complete, no matter the execution is successful or not, status goes back to idle. 

The code changes update the indicator flow as:
1. when the cell is running, the status changes from idle to running; when it's scheduled, from idle to scheduled;
  a. **NEW:** when an error cell is running/scheduled, the status changes from error to running/scheduled.
3. when the execution is complete, 
  a. when the execution is successful, status goes back to idle, 
  b. **NEW:** when the execution raises an error, status changes from running to error;
4. **NEW:** when an error cell's output is cleared, the status changes from error to idle. 

As for the code changes, 
1. in `toc.ts`, a new status is added to indicate error, a new array is added to record errorCells, `onExecuted` and `updateRunningStatus` is updated to handle flow 2b, and `onOutputCleared` is added to handle flow 3. 
2. in `actions.tsx` a new signal is emitted when output is cleared to handle flow 3. 
3. in `toc.css`, a visual is added for error status.

## User-facing changes

Before: 
image

After: 
image

None


skyetim:

packages/notebook/style/toc.css
```diff
@@ -30,7 +30,14 @@
   background-color: var(--jp-inverse-layout-color3);
 }
 
+.jp-tocItem-content[data-running='-1']::after {
+  border-radius: 50%;
+  border: var(--jp-border-width) solid var(--jp-error-color2);
+  background-color: var(--jp-error-color2);
```
krassowski, line 36: This produces a nice UX in general. While the existing implementation is not accessible for screenreaders, we can at least make this addition colourblind-accessible by using a different symbol. For example we could use
```suggestion
  content: '⚠';
  color: var(--jp-error-color2);
```

I have not tested, so it may need some further adjustments. Or we could use 🛑 sign.

skyetim:

packages/notebook/style/toc.css
skyetim: image
I made the color one level darker (`var(--jp-error-color1)`). color2 is slightly too light.

krassowski:
Bot, asking nicely, please update galata snapshots

krassowski:
[Last time](https://github.com/jupyterlab/jupyterlab/actions/runs/5500291388/jobs/10023110900) it failed errored out:

```
  1) [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 

    Internal error: worker process exited unexpectedly (code=null, signal=SIGABRT)

    Retry #1 ───────────────────────────────────────────────────────────────────────────────────────

    Internal error: worker process exited unexpectedly (code=null, signal=SIGABRT)

  1 failed
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 
  318 passed (55.3m)
Error:   1) [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 
    Internal error: worker process exited unexpectedly (code=null, signal=SIGABRT)
Error:   1) [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 

    Retry #1 ───────────────────────────────────────────────────────────────────────────────────────
    Internal error: worker process exited unexpectedly (code=null, signal=SIGABRT)
Notice:   1 failed
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 
  318 passed (55.3m)
```

Suggesting that there may be something wrong with the test.

The new run is here: https://github.com/jupyterlab/jupyterlab/actions/runs/5603502881/jobs/10250215665 but I am of little hope that it would have fixed itself - it was the only test failing.

krassowski:
Kicking CI

packages/notebook/src/toc.ts
```diff
@@ -21,7 +21,11 @@ export enum RunningStatus {
   /**
    * Cell is idle
    */
-  Idle = -1,
+  Idle = -2,
+  /**
+   * Cell execution is unsuccessful
+   */
+  Error = -1,
```
fcollonval, line 28: Could you set this to `-2` and leave `-1` on _Idle_ otherwise this is breaking the API (as it impacts also the DOM).

fcollonval (changes requested):
Thanks @skyetim 

The code looks good but I spotted the following problem:

error_clearing_output

In the screencast, 
1. execute the all notebook
2. Fix the cell with an error
3. execute that cell
4. The execution indicator is briefly shown but then it goes back to error state instead of idle.

---

I realize also that the table of content breaks when changing cell type for the cell containing a heading - this is independent of this feature; just for awareness.

packages/notebook/src/toc.ts
skyetim: If Idle is greater than Error and there are multiple cells under one heading, the heading will show as Idle instead of Error given how `getMaxIsRunning` is coded right now. Logically Idle should also be at the lowest priority. What API is it breaking? If we do have to work around that, we probably need to assign Error as -0.5.

skyetim:
There's another issue independent of this feat and somewhat related to #14604 : when deleting a running cell or error cell, the toc won't get updated.

packages/notebook/src/toc.ts
fcollonval: Ok let's go for `-0.5`

---

> What API is it breaking?

The enum number is directly set into the html data attribute. So strictly speaking if `Idle` value changes, the selector changes. I know this is being picky but I have seen consumer doing convoluted things.

krassowski:
@skyetim just flagging that this needs resolving conflicts

skyetim:

fcollonval:
Thanks for the patience @skyetim 

The last version looks good.

On Windows the emoji is used what ends up having a cut warning sign:

image

but we could tackle that in a follow-up

I took the liberty to push a commit to fix the integration test.

fcollonval:
The Windows issue should be fixed by the commit 509b4d2 that uses the special character [\FE0E](https://codepoints.net/U+FE0E) forcing the preceding unicode to be rendered as text.

fcollonval:

fcollonval:
Kicking the ci

fcollonval (approved):
Thanks @skyetim for the feature and the patience

krassowski:
The failures are relevant:

```
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:26:7 › ToC Running indicator › should display running indicators 
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 
```

krassowski:

krassowski (changes requested):
While the code and UX is fine, the visual tests are too flaky to merge, we need to improve them to be more reproducible:

```
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:26:7 › ToC Running indicator › should display running indicators 
    [jupyterlab] › test/jupyterlab/toc-running.test.ts:39:7 › ToC Running indicator › should display error indicators 
```

One idea would be to use notebook upload to eliminate a source of failures due to just problems with notebook creation. Then we would need to ensure that header selection state is consistent.

skyetim:
Is there a way to actually see the visual test results? I find it hard to reproduce on my local.

krassowski:
1. Hover over a commit with failure, click on Details:
   Screenshot from 2023-10-13 09-24-17
2. Click on Summary at the top of the page:
    Screenshot from 2023-10-13 09-24-34
3. Scroll down, click on `jupyterlab-galata-report`
   Screenshot from 2023-10-13 09-24-47

Unpack the zip, open the HTML with the report in a browser.

krassowski:

krassowski (approved):
Thanks @skyetim, all green now!

skyetim:
Thanks for updating the test!

PR #14785, opened 2023-07-02
Emit changed signal from outut area when an output is updated

### LINKED ISSUE #14781: vega (any?) MIME renderers do not update saved output MIME bundle
## Description

In `main` (but not `3.6.1`, which I happened to have running) MIME renderers appear to no longer be updating MIME bundles-at-rest when [`setData`](https://github.com/jupyterlab/jupyterlab/blob/main/packages/vega5-extension/src/index.ts#L153) is used.

## Reproduce

- make a notebook with just a vega mime output, a la 

```py
spec = {
  '$schema': 'https://vega.github.io/schema/vega-lite/v5.json',
  'data': {'values': [{ 'a': 'A', 'b': 28 }]},
  'mark': 'bar',
  'encoding': {
    'x': { 'field': 'a', 'type': 'nominal', 'axis': { 'labelAngle': 0 } },
    'y': { 'field': 'b', 'type': 'quantitative' }
  }
}

display({'application/vnd.vegalite.v4+json': spec}, raw=True)
```
- run it
- save it
- open with an editor
- do **not** see `image/png`

## Expected behavior

- see `image/png`

## Context

This is under [unit test](https://github.com/jupyterlab/jupyterlab/blob/main/packages/vega5-extension/test/renderer.spec.ts#L50), but not browser-based acceptance test.

A novel renderer on #14102, which is trying to `setData` a rendered SVG and some metadata (version number) is also showing this condition.

bollwyvl:
## References

- fixes #14781

## Code changes

- [x] emit `changed` from `OutputAreaModel` when an output is updated with `setData` (or changes in any other way, presumably)
- [ ] add browser-based acceptance test that verifies changes are updated by e.g. vega

## User-facing changes

- users should see fallback/metadata from mime renderers which use `setData`

- ???

### DIFF
--- a/packages/outputarea/src/model.ts
+++ b/packages/outputarea/src/model.ts
@@ -418,13 +418,23 @@ export class OutputAreaModel implements IOutputAreaModel {
    */
   private _onGenericChange(itemModel: IOutputModel): void {
     let idx: number;
+    let item: IOutputModel | null = null;
     for (idx = 0; idx < this.list.length; idx++) {
-      const item = this.list.get(idx);
+      item = this.list.get(idx);
       if (item === itemModel) {
         break;
       }
     }
-    this._stateChanged.emit(idx);
+    if (item != null) {
+      this._stateChanged.emit(idx);
+      this._changed.emit({
+        type: 'set',
+        newIndex: idx,
+        oldIndex: idx,
+        oldValues: [item],
+        newValues: [item]
+      });
+    }
   }
 
   private _lastStream = '';


galata/test/jupyterlab/notebook-run-vega.test.ts
```diff
@@ -41,16 +60,20 @@ test.describe.serial('Notebook Run Vega', () => {
...
     page,
     tmpPath
   }) => {
-    await page.notebook.openByPath(`${tmpPath}/${fileName}`);
+    const nbPath = `${tmpPath}/${fileName}`;
+    await page.notebook.openByPath(nbPath);
     await page.notebook.activate(fileName);
     await page.theme.setDarkTheme();
 
+    expect(await nbDiskContent(page, nbPath)).not.toContain(PNG_MIME_TYPE);
```
krassowski, line 75: A failure here is surprising. As if changing to dark theme had some effect on the paths?

galata/test/jupyterlab/notebook-run-vega.test.ts
```diff
@@ -59,5 +82,6 @@ test.describe.serial('Notebook Run Vega', () => {
     const nbPanel = await page.notebook.getNotebookInPanel();
 
     expect(await nbPanel.screenshot()).toMatchSnapshot(imageName);
+    expect(await nbDiskContent(page, nbPath)).toContain(PNG_MIME_TYPE);
```
krassowski, line 85: Or does it come from here?

bollwyvl:
well that was clearly wrong...

fcollonval:
Some subtle elements, as the fixture `tmpPath` is set at the file level, all tests will be executed within the same working folder. So as the file was saved, the _initial_ content changed between the 1st and 2nd test.

I pushed a fix to also use the `files` API instead of `fs` as getting the local path is a tough business (the root directory used is a dynamically generated temporary directory).

Hopefully the CI will be green(er :wink:)

bollwyvl:
Thanks for the alley-oop 🏀, there's clearly more about galata i don't know
than otherwise.

Likely this pattern of "check what ends up on disk after doing stuff" could
be applied in other places.

fcollonval (approved):
Thanks @bollwyvl

fcollonval:
> CI failure are not related

fcollonval:

PR #14792, opened 2023-07-04
Deprecates collaborative flag

### LINKED ISSUE #14774: --collaborative flag is not applied when jupyter-collaboration package is installed
The `--collaborative` flag aims to indicate if JupyterLab should run or not with RTC.

When you use use flag without `jupyter-collaboration` package, the server stops with  `To enable real-time collaboration, you must install the extension `jupyter_collaboration You can install it using pip for example: python -m pip install jupyter_collaboration`which is the expected behavior.

When  `jupyter-collaboration` is installed, the RTC mode is enabled even if you start without the `--collaborative` flag. RTC should be enabled only if you give `--collaborative` (see also https://github.com/jupyterlab/jupyterlab/issues/14715#issuecomment-1615172289)

hbcarlos:
Fixes #14774 

## Code changes
Deprecates the collaborative flag.

### DIFF
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -486,7 +486,11 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
     )
     flags["collaborative"] = (
         {"LabApp": {"collaborative": True}},
-        "Whether to enable collaborative mode.",
+        """To enable real-time collaboration, you must install the extension `jupyter_collaboration`.
+        This flag is now deprecated and will be removed in JupyterLab v5.
+        You can install it using pip for example:
+
+            python -m pip install jupyter_collaboration""",
     )
 
     subcommands = {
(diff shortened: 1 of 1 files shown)


jupyterlab/labapp.py
```diff
@@ -486,7 +486,11 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
     )
     flags["collaborative"] = (
         {"LabApp": {"collaborative": True}},
-        "Whether to enable collaborative mode.",
+        """To enable real-time collaboration, you must install the extension `jupyter_collaboration`.
+        This flag is now deprecated and will be removed in JupyterLab v5.
+        You can install it using pip for example:
+
+            python -m pip install jupyter_collaboration""",
```
krassowski, line 493: Would it read better if the order was switched around? Otherwise "it" seems to refer to the flag, not the extension.

```suggestion
        You can install it using pip for example:

            python -m pip install jupyter_collaboration
            
        This flag is now deprecated and will be removed in JupyterLab v5.""",
```

hbcarlos: You are right. I pushed a new commit with the change.

Thanks @krassowski

krassowski (approved):
Thank you!

fcollonval:

PR #14796, opened 2023-07-04
blur footer for any key that triggers a 'select above', not just ArrowUp

smacke:
This just makes sure that 'k' can still work for navigating after the 'Click to add a cell' footer has been focused.

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -513,8 +513,18 @@
     },
     {
       "command": "notebook:move-cursor-up",
+      "keys": ["ArrowUp"],
+      "selector": ".jp-Notebook-footer"
+    },
+    {
+      "command": "notebook:move-cursor-up",
       "keys": ["K"],
       "selector": ".jp-Notebook:focus"
+    },
+    {
+      "command": "notebook:move-cursor-up",
+      "keys": ["K"],
+      "selector": ".jp-Notebook-footer"
     },
     {
       "command": "notebook:move-cursor-heading-above-or-collapse",
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -499,32 +499,22 @@
     {
       "command": "notebook:move-cursor-down",
       "keys": ["ArrowDown"],
-      "selector": ".jp-Notebook:focus"
+      "selector": "[data-jp-traversable]:focus"
     },
     {
       "command": "notebook:move-cursor-down",
       "keys": ["J"],
-      "selector": ".jp-Notebook:focus"
+      "selector": "[data-jp-traversable]:focus"
     },
     {
       "command": "notebook:move-cursor-up",
       "keys": ["ArrowUp"],
-      "selector": ".jp-Notebook:focus"
-    },
-    {
-      "command": "notebook:move-cursor-up",
-      "keys": ["ArrowUp"],
-      "selector": ".jp-Notebook-footer"
+      "selector": "[data-jp-traversable]:focus"
     },
     {
       "command": "notebook:move-cursor-up",
       "keys": ["K"],
-      "selector": ".jp-Notebook:focus"
-    },
-    {
-      "command": "notebook:move-cursor-up",
-      "keys": ["K"],
-      "selector": ".jp-Notebook-footer"
+      "selector": "[data-jp-traversable]:focus"
     },
     {
       "command": "notebook:move-cursor-heading-above-or-collapse",
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -32,13 +32,6 @@ export class NotebookFooter extends Widget {
       case 'click':
         this.onClick();
         break;
-      case 'keydown': {
-        const key = (event as KeyboardEvent).key;
-        if (key === 'ArrowUp' || key === 'k') {
-          this.onArrowUp();
-          break;
-        }
-      }
     }
   }
 
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -50,6 +50,10 @@ export class NotebookFooter extends Widget {
    * blur the footer and switch to command mode.
    */
   public onSelectAbove(): void {
+    if (document.activeElement !== this.node) {
+      // short-circuit if the footer is not currently focused
+      return;
+    }
     this.node.blur();
     this.notebook.mode = 'command';
   }
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -33,8 +33,8 @@ export class NotebookFooter extends Widget {
         this.onClick();
         break;
       case 'keydown': {
-        const keyEvent = event as KeyboardEvent;
-        if (keyEvent.key === 'ArrowUp' || keyEvent.key === 'k') {
+        const key = (event as KeyboardEvent).key;
+        if (key === 'ArrowUp' || key === 'k') {
           this.onArrowUp();
           break;
         }
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -11,6 +11,11 @@ import { NotebookActions } from './actions';
 const NOTEBOOK_FOOTER_CLASS = 'jp-Notebook-footer';
 
 /**
+ * The data attribute added to a widget that can be traversed with up/down arrow and j/k shortcuts.
+ */
+const TRAVERSABLE = 'jpTraversable';
+
+/**
  * A footer widget added after the last cell of the notebook.
  */
 export class NotebookFooter extends Widget {
@@ -19,6 +24,7 @@ export class NotebookFooter extends Widget {
    */
   constructor(protected notebook: Notebook) {
     super({ node: document.createElement('button') });
+    this.node.dataset[TRAVERSABLE] = 'true';
     const trans = notebook.translator.load('jupyterlab');
     this.addClass(NOTEBOOK_FOOTER_CLASS);
     this.node.innerText = trans.__('Click to add a cell.');
--- a/packages/notebook/src/notebookfooter.ts
+++ b/packages/notebook/src/notebookfooter.ts
@@ -45,28 +45,12 @@ export class NotebookFooter extends Widget {
     NotebookActions.insertBelow(this.notebook);
   }
 
-  /**
-   * When an event (such as a keydown keyboard event) occurs that triggers selection of the widget above the footer,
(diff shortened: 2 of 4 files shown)


packages/notebook/src/notebookfooter.ts
```diff
@@ -32,11 +32,13 @@ export class NotebookFooter extends Widget {
...
         break;
-      case 'keydown':
-        if ((event as KeyboardEvent).key === 'ArrowUp') {
+      case 'keydown': {
+        const key = (event as KeyboardEvent).key;
+        if (key === 'ArrowUp' || key === 'k') {
           this.onArrowUp();
           break;
         }
+      }
```
krassowski, line 41: Thank you for highlighting it.

Ideally we would not have any of the keys hard-coded here and instead just check against keys defined for `notebook:move-cursor-up`, as users may want to remap their shortcuts (and in some cases when the notebook component is reused the <kbd>k</kbd> may not event be defined as cycling down). You can get the keys with:

```ts
const binding = app.commands.keyBindings.find(
    b => b.command === 'notebook:move-cursor-up'
);
binding.keys
```

The shortcuts themselves are defined here:

https://github.com/jupyterlab/jupyterlab/blob/c4de22b707eecc17c5d5c181c7b9991242abf9db/packages/notebook-extension/schema/tracker.json#L510-L518

smacke: Thanks for the pointer. I think we may need to do some additional work due to casing (i.e. I'm guessing that on the command it will be registered as `'K'` while on the event itself it will look like `'k'`). If so, would it be OK to check against the lowercase variants of everything?

krassowski: Yes, I think so.

smacke: Hmm just took another look at this today, seems like we don't have easy access to `app` / `JupyterFrontEnd` object from here. We would have to thread it down through `Notebook` to `NotebookFooter`, unless there's some magic bypass attached to `window`. Doing so seems like it would pollute a lot of public interfaces. Maybe we can merge the hack to at least get it working for the default shortcuts, and think about how to do a deeper design / integration with the keybinding system as a followup?

(Feel free to correct anything I misunderstood, my knowledge of the JupyterLab codebase is fairly superficial and based on very hacky extension development :) )

krassowski: Right. Maybe we should do this the other way round and teach `notebook:move-cursor-up` to handle the footer? It invokes `NotebookActions.selectAbove` defined here:

https://github.com/jupyterlab/jupyterlab/blob/42fb46470fb747ec1a8e6ad9a32ed11a0d4991a7/packages/notebook/src/actions.tsx#L846-L852

We could expose `footer` on `Notebook` and if it is focused, handle blurring it in `selectAbove()`?

smacke: Trying this now; hmm is there some upstream build issue? Can't install packages locally, and seems like CI may have the same problem.

krassowski: On CI only `check_release` job is expected to fail due to transitive GitHub API failure, all others jobs are passing elsewhere. Logs show a ts error in build:

```ts
  ../notebook/src/notebookfooter.ts(38,16): error TS2339: Property 'onArrowUp' does not exist on type 'NotebookFooter'.
```

It is still used in the fragment we are commenting on, so just removing it should fix it:

```suggestion
```

packages/notebook/src/notebookfooter.ts
```diff
@@ -51,9 +53,10 @@ export class NotebookFooter extends Widget {
   }
 
   /**
-   * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
+   * When an event (such as a keydown keyboard event) occurs that triggers selection of the widget above the footer,
+   * blur the footer and switch to command mode.
    */
-  protected onArrowUp(): void {
+  public onSelectAbove(): void {
```
krassowski, line 59: I think it should have a guard checking if `this.node` is focused (and if it is not, short-circuiting). This can be done comparing node with `document.activeElement`.

krassowski: Rationale is to avoid spurious switching to command mode as the setter has side effects.

packages/notebook/src/widget.ts
```diff
@@ -1250,6 +1250,9 @@ export namespace StaticNotebook {
  * A notebook widget that supports interactivity.
  */
 export class Notebook extends StaticNotebook {
+
+  footer: NotebookFooter;
```
krassowski, line 1254:
```suggestion
  protected footer: NotebookFooter;
```

packages/notebook/src/notebookfooter.ts
smacke: Oh yikes, thanks for catching that!

packages/notebook/src/widget.ts
smacke: Ah, for this to be accessible from `NotebookActions`, it actually has to be public I think.

packages/notebook/src/notebookfooter.ts
smacke: Finally managed to build jupyterlab locally in order to test e2e, and seems that `notebook:move-cursor-up` is not triggered when the footer is in focus, so unfortunately that approach doesn't work either. I did test out the original hacky approach of just checking for 'k' directly in the footer, and it seemed to work. I think to do this "properly" we would either need to figure out how to make the footer aware of the command registry, as in the original suggestion, or we would need to figure out if there's a way to make sure the notebook actions are triggered when the footer is in focus.

krassowski: So the selector in schema (the tracker.js file linked above) needs to be modified. There is another PR which will change these selectors from Notebook:focus to something more reasonable. Let's wait here for that PR, I will let you know once it's merged.

smacke: Checking back in here -- I took a look at trunk this morning at it seems the selectors are still the same. I'm guessing that the changes to fix that are fairly involved and will take some time to merge. Wanted to get your thoughts on hardcoding the shortcuts for now to take focus off the footer when the user presses `j` and `k` so that we can at least get the usability improvement in for the common case so that users (and selfishly, me in particular :) ) can start to enjoy it.

krassowski: That's fair. Could we special case the selector for `notebook:move-cursor-up` so that it matches either `.jp-Notebook:focus` or `.jp-Notebook-footer`? This is maybe `.jp-Notebook:focus,.jp-Notebook-footer`. I am not 100% sure if this will work, but seems worth trying.

packages/notebook-extension/schema/tracker.json
```diff
@@ -511,11 +511,21 @@
...
+    },
     {
       "command": "notebook:move-cursor-up",
       "keys": ["K"],
       "selector": ".jp-Notebook:focus"
     },
+    {
+      "command": "notebook:move-cursor-up",
+      "keys": ["K"],
+      "selector": ".jp-Notebook-footer"
```
krassowski, line 527: Can we collapse collapse this with the entry above by using a selector like `.jp-Notebook-footer, .jp-Notebook:focus`?

packages/notebook/src/notebookfooter.ts
```diff
@@ -51,11 +46,19 @@ export class NotebookFooter extends Widget {
   }
 
   /**
-   * On arrow up key pressed (keydown keyboard event), blur the footer and switch to command mode.
+   * When an event (such as a keydown keyboard event) occurs that triggers selection of the widget above the footer,
+   * blur the footer and switch to command mode.
+   * Returns whether the footer was blurred
+   * Returns whether the footer was blurred.
```
krassowski, line 52:
```suggestion
```

packages/notebook/src/notebookfooter.ts
smacke: So I wasn't able to use a selector with commas because it looks like it's a lumino selector and not a proper css selector (and lumino disallows commas), but what did work was to add an entirely new entry in the array for `.jp-Notebook-footer`. I did also have to check in the handler if the footer got unblurred, because otherwise it would both unfocus the footer and move up to the second-to-last cell (rather than staying at the last cell, which is what I expect), but overall it seems to be working. Thanks for the suggestion!

The only downside is that, in the keyboard shortcuts editor, there are now two entries for `Select Cell Above`, and both need to be edited if the user wants a working custom shortcut 😅 

Maybe @jasongrout knows how to get "boolean or" behavior in a lumino selector?

packages/notebook-extension/schema/tracker.json
krassowski: Discussion keeps going in https://github.com/jupyterlab/jupyterlab/pull/14796#discussion_r1320825853.

smacke: Unfortunately no (at least not with the ", " syntax); see https://github.com/jupyterlab/jupyterlab/pull/14796#discussion_r1320825853

packages/notebook/src/notebookfooter.ts
krassowski: I see, I guess the restriction in lumino could be lifted but this is a larger refactor due to other code that would need to be updated (e.g. specificity calculation). A solution is to add a class or data attribute which indicates the target; something like `[data-jp-traversable]:focus` and then the notebook gets `data-jp-traversable` attribute.

krassowski: I will push a commit in a minute.

packages/notebook/src/widget.ts
krassowski: Well, it does not need to be there at all since we can use layout directly, see https://github.com/jupyterlab/jupyterlab/pull/14796/commits/209612599eeb32d1759a486aaf3ffbf1b23068ee.

packages/notebook/src/notebookfooter.ts
krassowski: See https://github.com/jupyterlab/jupyterlab/pull/14796/commits/21659f7ef5b3cc0481a8a629a7aaf1dc8c2a0cbc.

krassowski:
@smacke can you confirm whether the latest commits solve the problem for you? This way we do not expand our public API, and we only have one shortcut.

krassowski (approved):
This is good to go from me, but will wait for confirmation from @smacke and potentially any other comments. It would be great to include this PR in next point release.

smacke:
Thanks for stepping in @krassowski! I can confirm that with these changes, everything behaves the way I would expect:
- both "k" and "arrowup" can be used to navigate up from the "add new cell" button
- if I change the keyboard shortcut in keyboard settings, the new setting also works (and there is just one entry in the manifest, unlike with my previous attempt)

krassowski:

PR #14799, opened 2023-07-07
Add global keyboard shortcuts for all tab bar widgets

g547315:
## References
The intended functionality here is to have a keyboard shortcuts for every widget on the tab bar that can be used from anywhere in the application to toggle it open and close

## Code changes

- Added a new commands with the functionality to activate, close and shift tab focus on to the chosen element on tab-bar

- Exposed this command in the command palette

## User-facing changes
New commands to activate:

- Left sidebar widgets with alt + number
- Right sidebar widgets with alt + shift + number

None


g547315:
Please review @gabalafou

g547315:

g547315:

packages/application-extension/schema/commands.json
```diff
@@ -46,6 +46,11 @@
       "command": "application:toggle-presentation-mode",
       "keys": [""],
       "selector": "body"
+    },
+    {
+      "command": "application:focusFirstElementInRightSidebar",
+      "keys": ["Alt Shift Z"],
```
krassowski, line 52: Why this shortcut? Is it standard? Would users know to use it?

packages/application-extension/src/index.tsx
```diff
@@ -327,6 +330,16 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         isToggled: () => !labShell.rightCollapsed,
         isEnabled: () => !labShell.isEmpty('right')
       });
+      //WIP
+      commands.addCommand(CommandIDs.focusFirstElementInRightSidebar, {
+        label: trans.__('Focus Right Sidebar First Element'),
+        execute: () => {
+          let rightSidBarFirstElement = document.getElementById('tab-key-1-1');
```
krassowski, line 337: The element should be retrieved programatically from `labShell` - relying from on element IDs is error-prone because we would not notice that something is wrong if the IDs were changed. IDs would be re acceptable if there is _really_ no other way and there is an accompanying test case.

krassowski (changes requested):

packages/application-extension/schema/commands.json
g547315: This shortcut was chosen due to its proximity to the tab key, secondly for keyboard users there is limited disruption to their hand alignment and positioning when using the shortcut.
This shortcut is not standard as the problem it solves is unique to this application (tab trap in notebooks). It also allows keyboard users to navigate the app more easily as an additional benefit

g547315:

krassowski (changes requested):
I do not believe that we should introduce such an non-standard shortcut (alt shift z) - this risks conflicts with other applications, extensions, and user actions and is not discoverable.

I think that the command proposed in this PR can be merged if made into generic, e.g. `application.focusSidebarElement` taking arguments `side: 'left' | 'right'` and `index: number`. We can then either to define a shortcut for each item in the sidebar, or just let user configure shortcuts that suit them.

For example, Slack uses ctrl + number and the numbers show up upon pressing ctrl:

image

The problem is that user may prefer to use `ctrl + number` for tab bar, not sidebar and then there are two sidebars - so it may require a combos like `alt + number` for left and `alt + shift + number` for right. Ideally user could modify which combo used for left/right sidebar and which for tabbar (which does not need to be implemented in this RP).

As for focusing **any** specific element in the sidebar, this can already be achieved using existing commands like `filebrowser:toggle-main` (ctrl + shift + f) `toc:show-panel` (ctrl + shift + k), `debugger:show-panel` (ctrl + shift + e), `extensionmanager:show-panel` (extension manager)

g547315:
Thanks for the feedback:

I will make the function more generic accepting varied arguments. Each argument combination will correspond to a keyboard shortcut, facilitating focus on the designated element. For instance, `"args":` `side: 'left'`, `index: 0` will correlate to the shortcut (ctrl shift 1), focusing on the first element of the left tab bar. 

Concerning the display of numbers on elements: using ctrl shift can reveal numbers for the left tab bar, while only ctrl can do so for the right tab bar. It might be best left to be address in a subsequent PR.

@krassowski

g547315:
Please review 
Changes  implemented:

- Refactor to generic method 
- Alt + number for left sidebar and alt + shift + number for right sidebar.

Still to be implemented in another PR:
- Numbers show up upon pressing alt or alt + shift: (Lumino Repo)
- User customization to modify which combo used for left/right sidebar and which for tabbar

galata/test/documentation/general.test.ts-snapshots/interface-right-common-documentation-linux.png
krassowski, line 1: This can be reverted. Apologies about the bot adding flaky snapshots from time to time.

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +330,24 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
+          if (index >= widgets.length) {
+             throw Error(`No item with index ${index} in ${args.side} sidebar`);
+          }
+          const dataId = widgets[index].id;
+          const focusElement = document.querySelector(
+            "[data-id='" + dataId + "']"
+          );
+          if (focusElement) {
+            focusElement.focus();
+          }
```
krassowski, line 348:
```suggestion
          }
        }
```

krassowski:
In the latest commit, https://github.com/jupyterlab/jupyterlab/pull/14799/commits/6c3a552bdc2580813ea82861b2bbb6b21f1a82eb I rewritten the logic to first open the sidebar widget and the focus on it. @g547315 what do you think about this approach? If not, please undo my changes.

However, the focus part does not seem to be working well. It appears to that sidebars do not receive focus in the first place. It maybe be different with https://github.com/jupyterlab/lumino/pull/612. Does this PR depend on it?

g547315:
@Krassowski I think the approach is good, I would note as per your previous comment

There are existing commands like filebrowser:toggle-main (ctrl + shift + f) toc:show-panel (ctrl + shift + k), debugger:show-panel (ctrl + shift + e), extensionmanager:show-panel (extension ctrl + shift + x) to activate these tabs so the might be some redundancy but i think this is ok as this will activate and shift tab focus 

Secondly tab focus depended on this [PR](https://github.com/jupyterlab/lumino/pull/583) which was then changed by the PR you mentioned so Yes

g547315:
update galata snapshots

krassowski:
I am not convinced whether an update is needed here. I re-triggered the regression tests run to see if it passes. For future, "please " prefix is needed for bot to pick up the comment.

g547315:
I think this ready for merging or is there still work to be done

krassowski:
For me it seems good but I would like to have an impartial opinion (since I contributed to this one), so waiting for a review (some maintainers are just coming back from holidays so the queue is longer this time of year).

I mentioned this PR (along other PRs with "needs review" label) during the dev meeting yesterday. If you would like to push it forward, help with reviewing other PRs would be appreciated if you feel like it ([_we all love puppies_](https://github.com/jupyterlab/jupyterlab/blob/main/CORPORATE.md)).

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +331,32 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
+        label: trans.__('Open Sidebar Element'),
+        caption: trans.__('By default opens first element of left sidebar'),
+        execute: args => {
+          const index = parseInt(args.index as string, 10);
+          if (args.side != 'left' && args.side != 'right') {
+            throw Error(`Unsupported sidebar: ${args.side}`);
+          }
+          const widgets = Array.from(labShell.widgets(args.side));
+          if (index >= widgets.length) {
+            throw Error(`No item with index ${index} in ${args.side} sidebar`);
```
brichet, line 344:
```suggestion
            return;
```
Maybe it can do nothing silently.

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +331,32 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
+          }
+          const widgetId = widgets[index].id;
+          // activate widget to show the associated content
+          labShell.activateById(widgetId);
```
brichet, line 348: Shouldn't it be useful to toggle the panel instead of just activating it ?
e.g. toggle-filebrowser command

brichet (commented):
Thanks @g547315 and @krassowski, that's really nice.

I have some comments below.

packages/application-extension/src/index.tsx
g547315: Unsure of the indented change here

Not sure what the difference of Activating the widget (simulating an on click) and toggling the widget.
secondly it is unclear to me that all the tab bar widgets can be toggled and if its even possible to access it by just having the widgets id.

The intended functionality here is to have a keyboard shortcut for every widget on the tab bar that can be used from anywhere in the application to activate and more tab focus to that widget

brichet: My idea was to be able to close an open widget.
It might be interesting to be able to close the sidebar using the same keyboard shortcut used to open it.

For example, I'm not sure if there is a shortcut to close the right panel.

g547315: changes made, can you please review

g547315:

g547315:

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +364,38 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         isEnabled: () => !labShell.isEmpty('right')
       });
 
+      commands.addCommand(CommandIDs.activateSidebarWidget, {
+        label: trans.__('Open Sidebar Element'),
```
brichet, line 368:
```suggestion
        label: trans.__('Toggle Sidebar Element'),
```

packages/application-extension/schema/commands.json
```diff
@@ -39,13 +39,193 @@
     },
     {
       "command": "application:toggle-right-area",
-      "keys": [""],
+      "keys": ["Accel Shift B"],
```
brichet, line 42: This shortcut is already used to open the 'running kernels' sidepanel:
https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/running-extension/schema/plugin.json#L56-L58

g547315:

g547315:

g547315:

g547315:
lease update documentation snapshots

g547315:

gabalafou:
I'm a little confused. Is this ready for review? When I tried testing this PR, the command fails because no arg (left/right) is provided to it when you execute the command from the command palette. In other words, it hits [line 373](https://github.com/jupyterlab/jupyterlab/pull/14799/files#diff-88062eb06c9ca4fb9080028ccdb5d0169b6053e050452fc40c33e767ad5ca0c6R373):

```js
throw Error(`Unsupported sidebar: ${args.side}`);
```

I'm also a little bit lost on what the overall goal of the PR is. I think the PR's description might be out of date. My best guess at what this PR aims to achieve, given the current state of the code is:

1. Provide a command to move browser focus to the left sidebar
2. Provide a command to move browser focus to the right sidebar 

I'm guessing that implies that if the sidebar is closed/collapsed/hidden, the command would first have to open/expand/unhide it. I'm also guessing that we want to put focus not on the tab-button itself but on the panel that clicking on the the tab-button opens/closes.

krassowski:
> when you execute the command from the command palette

This command is not intended to be invoked via command palette - it should not be there, good catch. The implementation that I worked on included shortcuts for each possible sidebar item (see https://github.com/jupyterlab/jupyterlab/pull/14799/commits/a2ba278d5660cd214421a869d7e2c61bc1326148) but it must have been removed later. I did not follow and do not understand why.

krassowski:
It looks like the shortcuts were removed by this commit:

image

https://github.com/jupyterlab/jupyterlab/pull/14799/commits/aa571ac7f7c585d1c48180e04967bb9a11131601

Apologies if I missed it but I do not see relevant discussion around it, nor good explanation in the commit message.

g547315:

g547315:

g547315:

g547315:

g547315:
Apologies I have updated the PR's description and re added the commands
@gabalafou

g547315:
> I'm a little confused. Is this ready for review? When I tried testing this PR, the command fails because no arg (left/right) is provided to it when you execute the command from the command palette. In other words, it hits [line 373](https://github.com/jupyterlab/jupyterlab/pull/14799/files#diff-88062eb06c9ca4fb9080028ccdb5d0169b6053e050452fc40c33e767ad5ca0c6R373):
> 
> ```js
> throw Error(`Unsupported sidebar: ${args.side}`);
> ```
> 
> I'm also a little bit lost on what the overall goal of the PR is. I think the PR's description might be out of date. My best guess at what this PR aims to achieve, given the current state of the code is:
> 
> 1. Provide a command to move browser focus to the left sidebar
> 2. Provide a command to move browser focus to the right sidebar
> 
> I'm guessing that implies that if the sidebar is closed/collapsed/hidden, the command would first have to open/expand/unhide it. I'm also guessing that we want to put focus not on the tab-button itself but on the panel that clicking on the the tab-button opens/closes.

Apologies I have updated the PR's description and re added the commands

g547315:
This PR is ready for re-review @gabalafou gabalafou

krassowski:
Kicking tests

g547315:

g547315:

packages/application-extension/src/index.tsx
```diff
@@ -104,6 +104,9 @@ namespace CommandIDs {
 
   export const toggleSideTabBar: string = 'application:toggle-side-tabbar';
 
+  export const activateSidebarWidget: string =
+    'application:activate-sidebar-widget';
```
brichet, line 108:
```suggestion
  export const toggleSidebarWidget: string =
    'application:toggle-sidebar-widget';
```

For a better readability, we should rename the command (it should also be updated in *commands.json*).

packages/application-extension/src/index.tsx
```diff
@@ -201,6 +204,39 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
+    };
+
+    // Sets tab focus on the element
+    function setTabFocus(focusElement: Element | null) {
+      if (focusElement) {
+        if ((focusElement as HTMLElement).focus) {
+          (focusElement as HTMLElement).focus();
+        }
+      }
+    }
```
brichet, line 238:
```suggestion
    function setTabFocus(focusElement: HTMLElement | null) {
      if (focusElement) {
        focusElement.focus();
      }
    }
```
The test can probably be simplified.

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +364,38 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         isEnabled: () => !labShell.isEmpty('right')
       });
 
+      commands.addCommand(CommandIDs.activateSidebarWidget, {
```
brichet, line 367:
```suggestion
      commands.addCommand(CommandIDs.toggleSidebarWidget, {
```

Renaming for readability (see above).

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +364,38 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
+          const index = parseInt(args.index as string, 10);
+          if (args.side != 'left' && args.side != 'right') {
+            throw Error(`Unsupported sidebar: ${args.side}`);
+          }
+          const widgets = Array.from(labShell.widgets(args.side));
+          if (index >= widgets.length) {
+            return;
+          }
+          const widgetId = widgets[index].id;
+          const focusElement = document.querySelector(
```
brichet, line 380:
```suggestion
          const focusElement: HTMLElement | null = document.querySelector(
```

brichet (changes requested):
Thanks @g547315, this looks great.

I have some comments below, mostly for code readability.

g547315:

g547315:

galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
```diff
@@ -31,6 +31,14 @@
       "Ctrl Shift ,"
     ]
   },
+  {
+    "id": "application:activate-sidebar-widget",
```
brichet, line 35: This command id should be updated.

g547315:

g547315:

g547315:

galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
g547315: Updated

brichet:
Bumping the CI

brichet (approved):
Thanks @g547315, it looks good to me.

galata/test/jupyterlab/readonly.test.ts-snapshots/readonly-jupyterlab-linux.png
brichet, line 1: This snapshot should probably be restored

g547315:

galata/test/jupyterlab/readonly.test.ts-snapshots/readonly-jupyterlab-linux.png
g547315: snapshot restored

g547315:

g547315:

g547315:

g547315:
@gabalafou
@krassowski 

Can this PR be merged?, The failing test are not due to the code changes make

galata/test/documentation/general.test.ts-snapshots/jupyterlab-documentation-linux.png
krassowski, line 1: This change appears to be causing a test failure:

>    [documentation] › test/documentation/general.test.ts:19:7 › General › Welcome ──────────────────

Can you revert this snapshot update manually?

brichet: I think @krassowski means you should probably revert it using git (locally) then push your changes, instead of using the galata comment in github.

g547315: Thanks for the clarification @brichet. @krassowski 
I have now updated the snap shot in question manually

brichet:
@krassowski what do you think about merging this PR.
The failing tests are not related.

packages/application-extension/src/index.tsx
```diff
@@ -328,6 +362,38 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         isEnabled: () => !labShell.isEmpty('right')
       });
 
+      commands.addCommand(CommandIDs.toggleSidebarWidget, {
+        label: trans.__('Toggle Sidebar Element'),
+        caption: trans.__('By default opens first element of the sidebar'),
```
krassowski, line 367: This is not the case I think? By default it errors as `args.index` is `undefined`?

```suggestion
```

krassowski: Also, caption may not be the best place to mention it even if it were working.

krassowski: @brichet good to merge once this thread is addressed. Feel free to go either way.

galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
```diff
@@ -31,6 +31,14 @@
       "Ctrl Shift ,"
     ]
   },
+  {
+    "id": "application:toggle-side-tabbar",
+    "label": "Toggle Sidebar Element",
+    "caption": "By default opens first element of the sidebar",
```
brichet, line 37:
```suggestion
```
This one should also be removed, then we should be good to merge.

packages/application-extension/src/index.tsx
g547315: I have removed the caption

galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
g547315: Removed

PR #14800, opened 2023-07-07
Adds rank to ICompletionProvider

### LINKED ISSUE #14790: Should completer providers be able to advertise suggested default rank?
### Problem

When adding a toolbar item or a menu item extensions can define a rank which determines positioning. There are some reserved ranges but the assumption is that everyone will play nice. The manager of `CompleterProvider` also uses ranks but it makes the default ranks up from the list of available providers (for third-party providers). This means that any provider added by extension always has a rank lower than the default providers, and in consequence the model nor renderer from extension is never used (unless user goes to settings and manually changes ranks).

https://github.com/jupyterlab/jupyterlab/blob/e04293eb2ff13b93a0ba8aedbd17e0d591adece7/packages/completer-extension/src/index.ts#L85-L89

### Proposed Solution

Allow extensions to set the preferred rank by adding `defafultRank` or `rank` or to `CompleterProvider` API and use this as default if user has not set a different value.

### Additional context

An alternative (for extensions) is to programatically modify `completer` settings to inject their defaults which is error-prone and deviates from how rank is used in other extendable components (menus, toolbars, context menus).

A short term solution that would unlock release of jupyterlab-lsp would be to put auto-generated extension ranks higher than those of built-in providers.

CC @trungleduc for opinions.

hbcarlos:
## References
fixes #14790

## Code changes
Adds rank to the ICompletionProvider and sets a rank of ~500 to the default providers.

## User-facing changes
The default rank of the providers.

N/A

### DIFF
--- a/packages/completer-extension/schema/manager.json
+++ b/packages/completer-extension/schema/manager.json
@@ -15,8 +15,8 @@
       },
       "additionalProperties": false,
       "default": {
-        "CompletionProvider:context": 550,
-        "CompletionProvider:kernel": 500
+        "CompletionProvider:context": 500,
+        "CompletionProvider:kernel": 550
       }
     },
     "providerTimeout": {
--- a/packages/completer-extension/schema/manager.json
+++ b/packages/completer-extension/schema/manager.json
@@ -15,8 +15,8 @@
       },
       "additionalProperties": false,
       "default": {
-        "CompletionProvider:context": 100,
-        "CompletionProvider:kernel": 200
+        "CompletionProvider:context": 501,
+        "CompletionProvider:kernel": 500
       }
     },
     "providerTimeout": {
--- a/packages/completer-extension/schema/manager.json
+++ b/packages/completer-extension/schema/manager.json
@@ -15,7 +15,7 @@
       },
       "additionalProperties": false,
       "default": {
-        "CompletionProvider:context": 501,
+        "CompletionProvider:context": 550,
         "CompletionProvider:kernel": 500
       }
     },
--- a/packages/completer/src/tokens.ts
+++ b/packages/completer/src/tokens.ts
@@ -50,7 +50,7 @@ export interface ICompletionProvider<
   readonly identifier: string;
 
   /**
-   * Rank use to order completion results from different completion providers.
+   * Rank used to order completion results from different completion providers.
    *
    * #### Note: The default providers (CompletionProvider:context and
    * CompletionProvider:kernel) use a rank of ≈500. If you want to give
--- a/packages/completer/src/tokens.ts
+++ b/packages/completer/src/tokens.ts
@@ -50,6 +50,19 @@ export interface ICompletionProvider<
   readonly identifier: string;
 
   /**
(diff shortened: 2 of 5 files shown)


packages/completer/src/tokens.ts
```diff
@@ -49,6 +49,19 @@ export interface ICompletionProvider<
    */
   readonly identifier: string;
 
+  /**
+   * Rank use to order completion results from different completion providers.
+   *
+   * #### Note: The default providers (CompletionProvider:context and
+   * CompletionProvider:kernel) use a rank of ≈500. If you want to give
+   * priority to your provider, use a rank of 1000 or above.
```
krassowski, line 57: I agree with the approach in general but instead of using rank 500 and 501 maybe we should leave some space between them to allow providers higher than one but lower than the other? For example 500 and 550

packages/completer/src/tokens.ts
```diff
@@ -49,6 +49,19 @@ export interface ICompletionProvider<
...
+   *
+   * The rank is optional for backwards compatibility. If the rank is `undefined`,
+   * it will assign a rank of [1, 499] making the provider available but with a
```
krassowski, line 60: Is this limit implemented?

packages/completer/src/tokens.ts
```diff
@@ -49,6 +49,19 @@ export interface ICompletionProvider<
    */
   readonly identifier: string;
 
+  /**
+   * Rank use to order completion results from different completion providers.
```
krassowski, line 53:
```suggestion
   * Rank used to order completion results from different completion providers.
```

packages/completer/src/tokens.ts
hbcarlos: No, the limit is not implemented, but you need 49 providers to reach a rank of 500.

fcollonval (approved):
Thanks @hbcarlos 

I would wait before merging so @krassowski can have another look at this one.

packages/completer-extension/schema/manager.json
```diff
@@ -15,8 +15,8 @@
       },
       "additionalProperties": false,
       "default": {
-        "CompletionProvider:context": 100,
-        "CompletionProvider:kernel": 200
+        "CompletionProvider:context": 550,
+        "CompletionProvider:kernel": 500
```
krassowski, line 19: Does higher rank = preferred? If so, kernel should have a higher rank, right? This is how it used to be. It may be also why the integration tests are failing.

krassowski (changes requested):
The failing tests is relevant:

```
jupyterlab › test/jupyterlab/completer.test.ts:85:9 › Completer › Notebook › Token completions show up without running the cell when in the same cell
```

packages/completer-extension/schema/manager.json
hbcarlos: Yes, rank = preferred. I pushed a commit to fix it.

krassowski (approved):
I see the relevant test passing now - thank you!

krassowski:

PR #14808, opened 2023-07-09
feat: sync theme with system settings

### LINKED ISSUE #8777: Mirror operating system's dark/light mode setting
### Problem

* Currently, JupyterLab supports toggling between dark and light mode.
* Various operating systems also support this feature, including [macOS], [Windows] and [Ubuntu].
* Many browsers support the CSS query `prefers-color-scheme`, which mirrors the OS theme ([MDN prefer-color-scheme docs][1])

Despite these features, Jupyter Lab does not support automatically toggling between dark/light mode to mirror the OS theme.

A good example of a site that has this support is the WebKit blog,  https://webkit.org/blog/.

[2]:https://stsievert.com/
[1]:https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
[macOS]:https://support.apple.com/en-us/HT208976
[Windows]:https://www.cnet.com/how-to/how-to-enable-dark-mode-in-windows-10/
[Ubuntu]:https://www.howtogeek.com/669755/how-to-enable-dark-mode-on-ubuntu-20.04-lts/

### Proposed Solution

Add an option to for JupyterLab to automatically toggle between light and dark mode depending on the OS theme. Maybe "Settings > Theme > Auto."

### Additional context

Here's a good site describing how to implement `prefers-color-scheme`: https://www.bram.us/2019/05/23/css-color-scheme-queries-dark-mode-css/. They show this demo:

This demo shows  https://webkit.org/blog/ and toggles the OS light/dark mode (see the navigation bar of the web browser window).

skyetim:
## References

Fixes #8777

## Code changes

1. Command palette: three options are added: a toggle to enable sync with system, preferred light and dark theme. 
2. Main menu: Sync with system option is added. I find it confusing to add options to select preferred light and dark theme here so leave these two options out. If users decide to change the theme in main menu, sync with system is also turned off to respect users' choice. 

The main logic changes are in L423 in `ThemeManager`, if `adaptive-theme` is True, ignore the original `theme` and use `preferredLightTheme` or `preferredDarkTheme` instead. 

## User-facing changes

Main Menu: 

image

Command palette: 

image

Effect: 
Old
image

New
image

None


skyetim:
Leave as draft now as I don't know how to toggle system settings in galata and add tests to this.

skyetim:
@krassowski I need some guidance on how to add tests to this change.

packages/apputils-extension/src/themesplugins.ts
```diff
@@ -123,10 +129,58 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
...
+        const theme = args['theme'] as string;
+        if (theme === manager.preferredDarkTheme) {
+          return;
+        }
+        return manager.setPreferredDarkTheme(theme);
+      }
+    });
+
+    commands.addCommand(CommandIDs.toggleAdaptiveTheme, {
+      label: trans.__('Sync with system settings'),
```
krassowski, line 177: Should it use Title Caps to align with other menu items?

packages/apputils-extension/src/themesplugins.ts
```diff
@@ -123,10 +129,58 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
...
+      isToggled: () => manager.isToggledAdaptiveTheme(),
+      execute: () => {
+        manager.toggleAdaptiveTheme();
```
krassowski, line 180: Needs an await or `.catch(console.warn)`

packages/apputils/src/tokens.ts
```diff
@@ -101,11 +101,36 @@ export interface IThemeManager {
...
    */
   readonly themes: ReadonlyArray<string>;
 
+  /**
+   * Get the names of the registered light themes.
+   */
+  readonly lightThemes: ReadonlyArray<string>;
+
+  /**
+   * Get the names of the registered light themes.
```
krassowski, line 130:
```suggestion
   * Get the names of the registered dark themes.
```

packages/apputils-extension/schema/themes.json
```diff
@@ -105,9 +109,27 @@
...
       "type": "string",
       "title": "Selected Theme",
-      "description": "Application-level visual styling theme",
+      "description": "Application-level visual styling theme. Ignored when Adaptive Theme is enabled.",
       "default": "JupyterLab Light"
     },
+    "adaptive-theme": {
+      "type": "boolean",
+      "title": "Adaptive theme",
+      "description": "Sync styling theme with system settings",
```
krassowski, line 118:
```suggestion
      "description": "Synchronise styling theme with system settings",
```

Let's make lives of translators easier.

krassowski (commented):
As for adding a galata test, `defineConfig` with `use: {colorScheme: 'dark'}` should do, see 
 https://playwright.dev/docs/emulation#color-scheme-and-media

skyetim:

fcollonval:
Kicking CI following snapshots update

krassowski (approved):
Thank you @skyetim! This looks good to me, subject to CI passing. I will wait with merging to let others take a second look.

packages/apputils/src/tokens.ts
```diff
@@ -101,11 +101,36 @@ export interface IThemeManager {
...
+
   /**
    * The names of the registered themes.
    */
   readonly themes: ReadonlyArray<string>;
 
+  /**
+   * Get the names of the registered light themes.
+   */
+  readonly lightThemes: ReadonlyArray<string>;
```
jtpio, line 127: Could this new property affect downstreams implementing an `IThemeManager`?

jtpio: For example Voila has its own `IThemeManager`: https://github.com/voila-dashboards/voila/blob/3ac569a4f4df15d094f7cad7c22c9934231023a6/packages/voila/src/plugins/themes/thememanager.ts#L36

fcollonval: I second @jtpio all those new attributes should support being `undefined` otherwise this will break overridden token `IThemeManager`.

packages/apputils-extension/src/themesplugins.ts
```diff
@@ -123,10 +129,58 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
...
       }
     });
 
+    commands.addCommand(CommandIDs.changePreferredLightTheme, {
+      label: args => {
+        const theme = args['theme'] as string;
+        const displayName = manager.getDisplayName(theme);
+        return args['isPalette']
+          ? trans.__('Use Theme: %1', displayName)
+          : displayName;
```
fcollonval, line 146: The label is not specific enough:

```suggestion
        return trans.__('Set %1 as preferred light theme', displayName);
```

packages/apputils-extension/src/themesplugins.ts
```diff
@@ -123,10 +129,58 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
...
+      }
+    });
+
+    commands.addCommand(CommandIDs.changePreferredDarkTheme, {
+      label: args => {
+        const theme = args['theme'] as string;
+        const displayName = manager.getDisplayName(theme);
+        return args['isPalette']
+          ? trans.__('Use Theme: %1', displayName)
+          : displayName;
```
fcollonval, line 164: Same here:

```suggestion
        return trans.__('Set %1 as preferred dark theme', displayName);
```

packages/apputils-extension/src/themesplugins.ts
```diff
@@ -123,10 +129,58 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
...
+        const theme = args['theme'] as string;
+        if (theme === manager.preferredDarkTheme) {
+          return;
+        }
+        return manager.setPreferredDarkTheme(theme);
+      }
+    });
+
+    commands.addCommand(CommandIDs.toggleAdaptiveTheme, {
+      label: trans.__('Synchronize with System Settings'),
```
fcollonval, line 177: I would be more specific here too as this is visible in the palette:

```suggestion
      label: trans.__('Synchronize Color Scheme with System Settings'),
```

fcollonval (changes requested):
Thanks @skyetim this looks great

I agree with @jtpio that the new API should be optional to not break code for 4.0.x. In addition the new command labels need to be changed to be more specific because for now we have duplication in the command palette:

image

packages/apputils-extension/src/themesplugins.ts
krassowski: I would suggest that the loner label is only used if `args['isPalette']` and the current one stays for menu to avoid the menu size exploding.

fcollonval (approved):
Thanks @skyetim

packages/apputils/src/thememanager.ts
```diff
@@ -60,20 +60,77 @@ export class ThemeManager implements IThemeManager {
...
    * A signal fired when the application theme changes.
    */
   get themeChanged(): ISignal<this, IChangedArgs<string, string | null>> {
     return this._themeChanged;
   }
 
+  /**
+   * Test if the preferred OS color scheme is dark
+   */
+  isOSColorSchemeDark(): boolean {
```
jtpio, line 127: Maybe the following would be more coherent since `system` seems to be used elsewhere:

```suggestion
  isSystemColorSchemeDark(): boolean {
```

fcollonval:
Thanks @skyetim for the patience - I'll merge this one as all comments have been taken into account and there were two approvals.

> CI failures are not related.

Issue #13006, opened 2022-08-25
Completer needs to support `isIncomplete` to properly implement LSP specification


krassowski:
The LSP's [`textDocument/completion`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) request may return _incomplete_ response. This is indicated by `isIncomplete` flag.

Incomplete response is used to implement faster completion where server decides to only send few relevant suggestions and then only send more suggestions once user types a character which narrows down what to send next. The client is responsible for sending another request after user types the character (with `TriggerForIncompleteCompletions` kind).

 Supporting `isIncomplete` is needed because:
- it enables faster completions when latency between server is low
- it is a non-optional part of LSP spec
- servers like TexLab and R server depend on it for completions

I attempted to implement this in a draft PR within the LSP extension: https://github.com/jupyter-lsp/jupyterlab-lsp/pull/838, but it looks like changes in JupyterLab completer itself will be needed.

For example, there is no signal emitted on each update of the filter text (there is only a signal about any change of the completer model state).

Also, there is no way to properly update completer in flight; while there is a `setCompletionItems` method on the model, the model relies on complex state of cursor and editor (`original` in snippet below) and update of that is tightly coupled with `CompleterHandler._updateModel` which is private:

https://github.com/jupyterlab/jupyterlab/blob/f7686ac86e6ec10f6fe8febcce0e404811b7e894/packages/completer/src/handler.ts#L386-L406

Updating `original` and `cursor` is needed as model extracts information from them when user changes text:

https://github.com/jupyterlab/jupyterlab/blob/f7686ac86e6ec10f6fe8febcce0e404811b7e894/packages/completer/src/model.ts#L58-L109

Speaking of which, the scope of `Completer.ITextState` seems too wide - completer does not care about `lineHeight` or `charWidth`.

https://github.com/jupyterlab/jupyterlab/blob/f7686ac86e6ec10f6fe8febcce0e404811b7e894/packages/completer/src/widget.ts#L712-L740

I think that it would be best to `isIncomplete` in core (not only because it is part of the specification, and servers rely on it), but also because of the extend of the changes to the public API that we might need to make to support implementing it downstream.

CC @trungleduc

trungleduc:
Thanks @krassowski for the heads up! I'll take a look at it.

PR #14809, opened 2023-07-10
Adds missing parameters to ICompleterProvider API

### LINKED ISSUE #14806: `shouldShowContinuousHint` needs context, fetch needs reason
## Description

In completer `shouldShowContinuousHint` has too little context. It only gets a boolean on completer visibility and ydoc source change. It should get adapter and widget information (i.e. the context object that fetch gets). We need the context to get connection. This is needed to obtain `triggerCharacters`

```ts
const triggerCharacters = connection.serverCapabilities['completionProvider']?.triggerCharacters
```

Also, it would be ideal if fetch also got information on the whether it was triggered as a consequence of `shouldShowContinuousHint` returning true.

https://github.com/jupyterlab/jupyterlab/blob/2161e6ca33cbaf98c4e5980cc97dc294b74c1836/packages/completer/src/tokens.ts#L99-L111

## Reproduce

This is roughly the dance we have to do right now:

```ts
class Provider {
   shouldShowContinuousHint(completerIsVisible, changed) {
     const connection = noWayToGetItWithout(context);  // HERE
     const triggerCharacters = connection.serverCapabilities['completionProvider']?.triggerCharacters;
     return triggerCharacters.includes(changed.inserted); // (simplified)
  }
  async fetch(
    context: ICompletionContext
  ): Promise<CompletionHandler.ICompletionItemsReply<CompletionItem>> {
    const manager = this.options.connectionManager;
    const widget = context.widget as IDocumentWidget;
    const adapter = manager.adapters.get(widget.context.path);
    const virtualDocument = someBlackMagic(adapter, context.editor, request.offset);
    const virtualPosition = moreBlackMagic(adapter, context.editor, request.offset);
    const connection = manager.connections.get(virtualDocument.uri);
    const wasTriggeredByContiniousHint = statefulBadMagicFixMePlease() // HERE
    const lspCompletionReply = await connection.clientRequests[
      'textDocument/completion'
    ].request({
      textDocument: {
        uri: virtualDocument.documentInfo.uri
      },
      position: {
        line: virtualPosition.line,
        character: virtualPosition.ch
      },
      context: {
        triggerKind: wasTriggeredByContiniousHint ?
          CompletionTriggerKind.TriggerCharacter :
          CompletionTriggerKind.Invoked
        triggerCharacter: wasTriggeredByContiniousHint ?
          getLastCharacterAndHopeItWasTheTrigger(editor, request) : // HERE
          undefined
      }
    });
   // do something with reply
  }
}
```

## Expected behavior

- `shouldShowContinuousHint` receives `context`; ideally both get the same set arguments
- `fetch` gets result of `shouldShowContinuousHint` as an optional argument

## Context

- JupyterLab version: 4.0.2

hbcarlos:
## References
Fixes #14806 

## Code changes
see #14806 

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/completer/src/tokens.ts
+++ b/packages/completer/src/tokens.ts
@@ -12,6 +12,15 @@ import { CompletionHandler } from './handler';
 import { Completer } from './widget';
 
 /**
+ * The type of completion request.
+ */
+export enum CompletionTriggerKind {
+  Invoked = 1,
+  TriggerCharacter = 2,
+  TriggerForIncompleteCompletions = 3
+}
+
+/**
  * The context which will be passed to the `fetch` function
  * of a provider.
  */
@@ -66,12 +75,12 @@ export interface ICompletionProvider<
    *
    * @param request - the completion request text and details
    * @param context - additional information about context of completion request
-   * @param shouldShowContinuousHint - Who triggered the request (optional).
+   * @param trigger - Who triggered the request (optional).
    */
   fetch(
     request: CompletionHandler.IRequest,
     context: ICompletionContext,
-    shouldShowContinuousHint?: boolean
+    trigger?: CompletionTriggerKind
   ): Promise<CompletionHandler.ICompletionItemsReply<T>>;
 
   /**
@@ -165,9 +174,11 @@ export interface IProviderReconciliator {
    * the result of this provider will be ignore.
    *
    * @param {CompletionHandler.IRequest} request - The completion request.
+   * @param trigger - Who triggered the request (optional).
    */
   fetch(
-    request: CompletionHandler.IRequest
+    request: CompletionHandler.IRequest,
+    trigger?: CompletionTriggerKind
   ): Promise<CompletionHandler.ICompletionItemsReply | null>;
 
   /**
--- a/packages/completer/src/tokens.ts
+++ b/packages/completer/src/tokens.ts
@@ -113,7 +113,7 @@ export interface ICompletionProvider<
    * completion items should be shown.
    *
    * @param  completerIsVisible - Current visibility status of the
-   *  completer widget0
+   *  completer widget
    * @param  changed - changed text.
    * @param  context - The context of the completer (optional).
    */
--- a/packages/completer/src/tokens.ts
+++ b/packages/completer/src/tokens.ts
@@ -66,10 +66,12 @@ export interface ICompletionProvider<
    *
    * @param request - the completion request text and details
    * @param context - additional information about context of completion request
+   * @param shouldShowContinuousHint - Who triggered the request (optional).
    */
   fetch(
     request: CompletionHandler.IRequest,
-    context: ICompletionContext
+    context: ICompletionContext,
+    shouldShowContinuousHint?: boolean
   ): Promise<CompletionHandler.ICompletionItemsReply<T>>;
 
   /**
@@ -104,10 +106,12 @@ export interface ICompletionProvider<
    * @param  completerIsVisible - Current visibility status of the
    *  completer widget0
    * @param  changed - changed text.
+   * @param  context - The context of the completer (optional).
    */
   shouldShowContinuousHint?(
     completerIsVisible: boolean,
-    changed: SourceChange
+    changed: SourceChange,
+    context?: ICompletionContext
   ): boolean;
 }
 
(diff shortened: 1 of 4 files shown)


packages/completer/src/tokens.ts
```diff
@@ -66,10 +66,12 @@ export interface ICompletionProvider<
    *
    * @param request - the completion request text and details
    * @param context - additional information about context of completion request
+   * @param shouldShowContinuousHint - Who triggered the request (optional).
    */
   fetch(
     request: CompletionHandler.IRequest,
-    context: ICompletionContext
+    context: ICompletionContext,
+    shouldShowContinuousHint?: boolean
```
krassowski, line 74: In the future we will also need a third state, `TriggerForIncompleteCompletions`, see [CompletionTriggerKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionTriggerKind) and https://github.com/jupyterlab/jupyterlab/issues/13006, so instead of adding multiple booleans, I would either use a custom enum, or re-use the type from lsp protocol. I think we should export or redefine such an enum/type from here to avoid completer providers interested in using it from depending on lsprotocol.

```suggestion
    trigger?: CompletionTriggerKind
```

krassowski:
If we are changing the interface we should also implement the logic here. It will be useful beyond LSP implementation (for other completion providers). In the current draft the LSP implementation no longer needs to swap the completion handler (which was a nightmare previously), so hopefully we won't need to do this again.

packages/completer/src/tokens.ts
```diff
@@ -104,10 +115,12 @@ export interface ICompletionProvider<
    * @param  completerIsVisible - Current visibility status of the
    *  completer widget0
```
krassowski, line 116:
```suggestion
   *  completer widget
```

While we are at it..

krassowski (approved):
Thanks! One minor thing: a tests checking that `trigger` gets passed down would be useful here (because `trigger` is optional typing will not save us if someone deletes it accidentally in a refactor).

packages/metapackage/test/completer/handler.spec.ts
```diff
@@ -317,6 +346,88 @@ describe('@jupyterlab/completer', () => {
...
+        editor.model.sharedModel.setSource('bar');
+        editor.setCursorPosition({ line: 0, column: 2 });
+        // This signal is emitted (again) because the cursor position that
+        // a natural user would create need to be recreated here.
+        // (editor.model.value.changed as any).emit({ type: 'set', value: 'bar' }); @todo remove?
+        (editor.model.sharedModel.changed as any).emit([]);
+        expect(model.methods).toEqual(
+          expect.arrayContaining(['handleTextChange'])
+        );
+      });
```
krassowski, line 428: This one looks like accidental copy-paste from an earlier test, do I see it right?
```suggestion
```

packages/metapackage/test/completer/handler.spec.ts
```diff
@@ -317,6 +346,88 @@ describe('@jupyterlab/completer', () => {
...
+        handler.editor = createEditorWidget().editor;
+
+        expect(provider.methods.length).toEqual(0);
+
+        handler.editor.model.sharedModel.setSource('foo.');
+        // @ts-ignore
+        handler._makeRequest(
+          { line: 0, column: 3 },
+          CompletionTriggerKind.Invoked
+        );
```
krassowski, line 399: It ain't end-to-end but I guess good enough for now.

packages/metapackage/test/completer/handler.spec.ts
hbcarlos: Ouch. Yes, it is a leftover from a copy-paste. Thanks @krassowski

packages/metapackage/test/completer/handler.spec.ts
hbcarlos: @krassowski, I improved the tests in my last commit. please take a look at them and let me know.

packages/metapackage/test/completer/handler.spec.ts
```diff
@@ -8,10 +8,15 @@ import {
...
   CompletionHandler,
+  CompletionTriggerKind,
+  ICompletionContext,
+  ICompletionProvider,
   ProviderReconciliator
 } from '@jupyterlab/completer';
+import { Widget } from '@lumino/widgets';
 import { ISharedText, SourceChange, YFile } from '@jupyter/ydoc';
 import { createSessionContext } from '@jupyterlab/apputils/lib/testutils';
+//import { simulate } from 'simulate-event';
```
krassowski, line 19:
```suggestion
```

fcollonval (approved):
Thanks @hbcarlos

fcollonval:
> CI failure are not related

fcollonval:

PR #14812, opened 2023-07-11
Adds this context when calling ICompletionProvider.modelFactory

### LINKED ISSUE #14805: `this` context gets lost in `modelFactory`
## Description

In this fragment:

https://github.com/jupyterlab/jupyterlab/blob/2161e6ca33cbaf98c4e5980cc97dc294b74c1836/packages/completer/src/manager.ts#L216-L222

`modelFactory` method gets assigned to a variable loosing `this` context. This can be fixed by either using `call`:

```
model = await modelFactory.call(this, completerContext); 
```

or binding:

```
const modelFactory = provider ? provider.modelFactory.bind(provider) : null; 
```

and of course this can be worked around downstream by assigning an anonymous function (but we should not need to).

## Reproduce

```ts
class CompletionProvider {
   myObj = 1
   async modelFactory() {
      this.myObj; // will throw
     return null as any
   }
}
```

will throw:

```ts
provider.ts:62 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'myObj')
    at modelFactory (provider.ts:62:27)
    at A.generateHandler
    at A.updateCompleter
````

Note this tests passes:

https://github.com/jupyterlab/jupyterlab/blob/2161e6ca33cbaf98c4e5980cc97dc294b74c1836/packages/metapackage/test/completer/manager.spec.ts#L168-L168

only because the factory is not using `this`:

https://github.com/jupyterlab/jupyterlab/blob/2161e6ca33cbaf98c4e5980cc97dc294b74c1836/packages/metapackage/test/completer/manager.spec.ts#L72-L74

## Expected behavior

Completer model factories can access attributes on their context.

## Context

- JupyterLab version: 4.0.2

hbcarlos:
## References
Fixes #14805 

## Code changes
See #14805 

## User-facing changes
N/A

N/A

### DIFF
--- a/packages/completer/src/manager.ts
+++ b/packages/completer/src/manager.ts
@@ -216,7 +216,7 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     const modelFactory = provider?.modelFactory;
     let model: Completer.IModel;
     if (modelFactory) {
-      model = await modelFactory(completerContext);
+      model = await modelFactory.call(provider, completerContext);
     } else {
       model = new CompleterModel();
     }
--- a/packages/metapackage/test/completer/manager.spec.ts
+++ b/packages/metapackage/test/completer/manager.spec.ts
@@ -70,6 +70,11 @@ class FooCompletionProvider implements ICompletionProvider {
   }
 
   async modelFactory(context: ICompletionContext): Promise<Completer.IModel> {
+    if (!(this instanceof FooCompletionProvider)) {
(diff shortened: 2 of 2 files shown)


krassowski (commented):
Worth modifying tests to test this?

krassowski (approved):
Thank you @hbcarlos!

fcollonval:

PR #14320, opened 2023-04-03
Add more alt text descriptions

s596757:
PR-label: Accessibility
Labels: Enhancement
Have added titles and descriptions to headings, icons and images as per ticket request. Have removed descriptions from within svgs and instead added changes to the addtitlestosvgs function to allow titles to be read @krassowski please can you check and approve so we can move forward
## References
#9682 and ticket from prioratised task list 

## Code changes

Added alt description to images and icons 

## User-facing changes

Icons and images have descriptive labels when you hover over them

PR #14819, opened 2023-07-12
Alt text and marking elements decorative

s596757:
As per comments on PR (https://github.com/jupyterlab/jupyterlab/pull/14320), removed redundant title attribute in code and left aria-hidden.

Changed the getTitleSvg function to include DOMUtils as per recommendation

tested via speak aloud and narrator. Previously was reading out icons descriptions but is no longer doing so in either.
Created new PR as unable to push into previous PR due to base branch changes.

Reference PR: https://github.com/jupyterlab/jupyterlab/pull/14320


isabela-pf (commented):
Thank you so much for working on this and being persistent even when it takes time for me to get reviewing. Overall, I agree with the idea that the launcher icons (next to the text title of a launcher section) should be decorative and note read from a user experience perspective. 

I'm not clear what
> Changed the getTitleSvg function to include DOMUtils as per recommendation
impacts on the user side. Based on comments on past PRs it looks like this was a request to remove unused code? Since it was requested I'm open to it, but I don't really know how to usefully review it. @gabalafou was this your request in the past?

For clarity, I don't have the resources to review this with assistive tech myself at the moment, so know I am reviewing this solely on reading the Files Changed. I also don't have quite enough knowledge of ARIA in this use case and how it interacts with JupyterLab. I'm really reviewing this based on the proposed user interaction of the PR.

I do have a few questions I'd like answered before I approve it:
- Are there any negative consequences of hiding what is also part `h2`? 
- Is the intent for the `TODO` to stay in the JupyterLab codebase, or is that a marker for the author before this PR is to be merged? I am seeing precedent for it elsewhere in the codebase, so if it's intentional I won't block on it.

gabalafou:
For reference, this PR supersedes https://github.com/jupyterlab/jupyterlab/pull/14400

gabalafou (commented):
I believe we said in the call yesterday that we should pause review on this PR. However, I decided to look at it anyway and leave some remarks.

gabalafou (changes requested):
Oops meant to click "request changes"

isabela-pf:
> I believe we said in the call yesterday that we should pause review on this PR. 

@gabalafou then that was my misunderstanding from the accessibility meeting. Sorry about that! I'll check in again at the upcoming meeting and won't mess with this PR until then.

krassowski:
@s596757 are you still interested in finishing this one?

Issue #14250, opened 2023-03-21
CSV files' icon in tab bar is not vertically centered


JasonWeill:
## Description

The CSV "spreadsheet" icon is aligned too close to the top of the tab bar, not centered like the "notebook" icon is.

<img width="473" alt="Screen Shot 2023-03-20 at 3 20 28 PM" src="https://user-images.githubusercontent.com/93281816/226766794-25592c6d-abb5-4941-bc13-04376f1af017.png">

## Context

- Operating System and version: macOS
- Browser and version: Chrome
- JupyterLab version: 4.0.0a37

JasonWeill:
This is a regression; it works in 3.5.x.

I still see the spreadsheet icon top-aligned when I open it for the first time in JupyterLab 4.x, but after I reload, the icon is centered.

There is no extra class associated with the spreadsheet icon upon initial addition, as there is with the notebook icon.

Could this be related to #13671 by @krassowski, which removed the `transformY` rule used to center content vertically inside the tab bar?

krassowski:
The issue appears to be with `mainAreaTab` stylesheet not being applied under some circumstances:

https://github.com/jupyterlab/jupyterlab/blob/899e6bbafa99abc3f4a7636fb1af2b5463fe7cd9/packages/application/src/shell.ts#L1418-L1426

This only happens once on the first time you open a spreadsheet file. The `title.icon` is undefined at the time it is being added to main area. I narrowed it down to https://github.com/jupyterlab/jupyterlab/pull/14037.

There is a related issue - on opening second CSV file the theme is not applied correctly:

![Screenshot from 2023-07-12 18-48-19](https://github.com/jupyterlab/jupyterlab/assets/5832902/2de2b3bd-7496-490d-a5bb-082f972e53d8)

and the following error gets thrown:

```js
widget.ts:316  Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'style')
    at set style [as style] (widget.ts:316:21)
    at index.ts:183:25
    at invokeSlot (index.es6.js:543:1)
    at Object.emit (index.es6.js:501:1)
    at Signal.emit (index.es6.js:100:1)
    at CSVViewerFactory.createNew (default.ts:484:25)
    at DocumentWidgetManager.createWidget (widgetmanager.ts:98:28)
    at DocumentManager._createOrOpenDocument (manager.ts:670:40)
    at DocumentManager.open (manager.ts:435:17)
[...]

PR #14822, opened 2023-07-12
Fix icon and theme/styles for grid in CSV/TSV viewer

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14250

## Code changes

Move spreadsheet icon setter to before `await` to execute it synchronously - otherwise the icon is set too late (after adding widget to main area and hence too late to get proper class applied) due to async code execution.

Also fixes an issue with style being set on content which is not ready and throwing an error by awaiting on initialisation (adds new public `ready` member on `CSVViewer`).

## User-facing changes

| Before | After |
|--|--|
|Screenshot from 2023-07-12 18-54-57 | Screenshot from 2023-07-12 18-55-23 |

Adds a non-optional API member `ready` on `CSVViewer` class, we can make it optional on backport.

### DIFF
--- a/packages/csvviewer-extension/src/index.ts
+++ b/packages/csvviewer-extension/src/index.ts
@@ -161,12 +161,6 @@ function activateCsv(
   let searchProviderInitialized = false;
 
   factory.widgetCreated.connect(async (sender, widget) => {
-    if (searchRegistry && !searchProviderInitialized) {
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @krassowski

krassowski:

PR #14834, opened 2023-07-17
Bump version to 4.1.0a0

### LINKED ISSUE #14833: Check release is broken on main
## Description

Check release is broken on main

The reason that the package on main are targeted 4.1.0 prereleases but the version have not been bumped.

fcollonval:
## References

Fixes #14833

## Code changes

- Bump packages to 4.1.0a0
- Remove setting the NPM tag by providing a command in pyproject.toml as the releaser support dynamically setting the tag
- Fix some versions not aligned in the federated example

## User-facing changes

None

(fix the check release job on main)

None


jtpio:
> Remove setting the NPM tag by providing a command in pyproject.toml as the releaser support dynamically setting the tag

Does this require changing the releaser config in JupyterLab? Or does the releaser handle the `latest` / `next` tag?

fcollonval:
> Does this require changing the releaser config in JupyterLab? Or does the releaser handle the latest / next tag?

I updated the releaser so it supports it natively if the GitHub release pre-release are consistent: see https://github.com/jupyter-server/jupyter_releaser/pull/510

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4, 0, 2, "final", 0
+current_version = 4, 1, 0, "alpha", 0
```
jtpio, line 2: Will that not cause an issue when releasing the first `4.1.0a0` with the releaser?

jtpio: Or we need to start at `4.1.0a1`

PR #14837, opened 2023-07-17
Move opening help resources to a separate plugin

jtpio:
## References

This should help with https://github.com/jupyter/notebook/issues/6963 and https://github.com/jupyter/notebook/pull/6968

## Code changes

Move the command that opens help resources in the `main` area in IFrames or new browser tabs to a new plugin.

That way, downstream applications like Notebook 7 can choose to not include this plugin and provide their own instead.

## User-facing changes

None

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@
 
 # JupyterLab Changelog
 
+## v4.1
+
+### 4.1.0 - Highlights
+
+#### Moved plugins
+
+Some internal JupyterLab plugins have been re-organized to allow for better flexibility for deployments and downstream applications like Notebook 7. This might affect users that disable specific plugins with the `jupyter labextension disable` command or the `disabledExtensions` config option.
+
+- The `help:open` command is not defined in the `@jupyterlab/help-extension:resources` plugin anymore, but has been moved to a new `@jupyterlab/help-extension:open` plugin instead.
+
 ## v4.0
 
 ### 4.0.0 - Highlights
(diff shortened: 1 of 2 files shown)


packages/help-extension/src/index.tsx
```diff
@@ -298,6 +272,51 @@ const resources: JupyterFrontEndPlugin<void> = {
...
+  autoStart: true,
+  requires: [IMainMenu, ITranslator],
+  optional: [ILabShell, ICommandPalette],
+  activate: (
+    app: JupyterFrontEnd,
+    mainMenu: IMainMenu,
+    translator: ITranslator,
+    labShell: ILabShell | null,
+    palette: ICommandPalette | null,
+    restorer: ILayoutRestorer | null
```
fcollonval, line 293: Not requested any more

```suggestion
```

fcollonval (changes requested):
Thanks @jtpio 

I spotted a small error. But otherwise it looks good to go.

However I'm not in favor of backporting it to 4.0.x as the new plugin defines a command from the old `resources` plugin. So if someone disable the `resources`, it may create a conflict by providing its own implementation of that command or simply adding a command he does not want.

packages/help-extension/src/index.tsx
jtpio: Thanks for catching this :+1:

fcollonval (approved):
Thanks @jtpio 

Do you think you could add a note in the CHANGELOG as part of this PR so we don't forget to do it when releasing?

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@
 
 # JupyterLab Changelog
 
+## v4.1
```
jtpio, line 7: Let's see if CI still passes with this new section.

fcollonval:
Thanks @jtpio 

> Remaining CI failures are not related

PR #14841, opened 2023-07-18
Fix CodeMirror theme selection background color

### LINKED ISSUE #14737: Theme problem: Focused selection always ends up with background #d7d4f0
## Description

No matter what theme I chose (even the official dark theme), when I select an area of text inside a notebook or the editor of the lab the selected text always ends up being #d7d4f0. I tried all sorts of css classes i could grasp but I am not able to change the background color of selected text that is focused. Is it only me experiencing this?

## Reproduce

Open a notebook or txt-file in jupyterlab, select some text.

## Expected behavior

Color changes to whatever is set in theme css variable --jp-editor-selected-focused-background

AllanChain:
## References

Fixes #14737

## Code changes

Update the CSS selector for the selection background color.

Previously, the rule from the base CodeMirror6 theme `.cm-focused > .cm-scroller > .cm-selectionLayer .cm-selectionBackground` takes precedence over our `.cm-focused .cm-selectionBackground` rule, so the theme selection color is not taking effect. Changing the selector to be the same as the base theme fixes the issue.

## User-facing changes

|Before|After|
|:-----|:----|
|Before|After|

None.

### DIFF
--- a/packages/codemirror/src/theme.ts
+++ b/packages/codemirror/src/theme.ts
@@ -46,7 +46,7 @@ export const jupyterEditorTheme = EditorView.theme({
     backgroundColor: 'var(--jp-editor-selected-background)'
   },
 
-  '&.cm-focused .cm-selectionBackground': {
+  '&.cm-focused > .cm-scroller > .cm-selectionLayer .cm-selectionBackground': {
     backgroundColor: 'var(--jp-editor-selected-focused-background)'
   },
 


krassowski (approved):
Thank you @AllanChain.

fcollonval (approved):
Thanks @AllanChain

fcollonval:

Issue #14852, opened 2023-07-19
Improve keyboard accessiblity on statusBar


j264415:
### Problem

1. When testing keyboard navigation the **line column**, **kernel status and running session** components were not reachable when tabbing. 

2. Furthermore, these components do not respond to **"Enter"** and **"Spacebar"** . 

### Proposed Solution

To increase keyboard navigation accessibility for users who need it, we suggest adding **"tabIndex=0"** to: **the line/Column, the Kernel status and, the running session** components in the statusBar. 

Additionally, keyboard functionality should be added so that these elements will respond to **"Enter"** and **"Spacebar"**. 

### Additional context

Working on a ticket for this issue.

krassowski:
Closing as resolved by https://github.com/jupyterlab/jupyterlab/pull/14853

PR #14853, opened 2023-07-19
Enable keyboard navigation in the statusbar

j264415:
## References

#14852 Improve keyboard accessibility on statusBar

## Code changes

A **"tabIndex='0'"** was added to the **lineCol.tsx** , **kernalStaus.tsx** and the **runningSessions.tsx** files to allow for the statusBar to be reachable via keyboard navigation. 
StatusBar components such as the Mode command, notification and Notebook title were left out because they provided no action when clicked. 

## User-facing changes

Currently when using keyboard navigation to tab through Jupyter, it was noticed that apart from the **"Simple switch",** the rest of the statusBar elements were not tab able. Component that did not perform actions when clicked, outside of keyboard navigation, were left without a **tabIndex**. 
This means that with the **tabIndex='0'** added to the elements of the **statusBar** elements such as the **running session, kernel status and the line/Column**, these component can now be fully accessed by users who require keyboard navigation. 

None


packages/apputils/src/kernelstatuses.tsx
```diff
@@ -58,8 +71,10 @@ function KernelStatusComponent(
   return (
     <TextItem
```
krassowski, line 72: This is a question: what do you all think about making `TextItem` and `GroupItem` default to adding an `onKeyDown` mirroring `onClick` when `onClick` is provided?

krassowski: Thinking: this would ensure downstream extensions are accessible too, without much overhead.

krassowski:
@j264415 I had reviewed this PR two weeks ago.

packages/apputils/src/kernelstatuses.tsx
j264415: Yeah i think this would be useful as this mighty be a functionality we would have create for other components down the line, so it would be good to add it as a default.

j264415:

j264415:

j264415:

j264415:

j264415:

galata/test/jupyterlab/outputarea-stdin.test.ts-snapshots/stdin-history-search-jupyterlab-linux.png
krassowski, line 1: This change appears unintentional and causes tests to fail. Can you please revert it?

packages/apputils/src/kernelstatuses.tsx
krassowski: Can this be done in this PR?

galata/test/jupyterlab/outputarea-stdin.test.ts-snapshots/stdin-history-search-jupyterlab-linux.png
j264415: Hi @krassowski all done. Reverted it as suggested.

tonyfast:
i am not able to succesfully test this feature on [binder](https://mybinder.org/v2/gh/jupyterlab/lumino/main?urlpath=lab/tree/examples). 

testing this feature is incredibly difficult with the lingering #14115 issue. the only way i could to `Tab` to the status bar was with no notebooks open. with no notebooks open, i could tab to the simple document switcher. once there i could NOT tab to the running session or kernel status. is the binder building as expected for y'all. i am not seeing the `tabindex=0` value that is expected.

packages/apputils/src/kernelstatuses.tsx
j264415: currently working on how to implement this feature.

j264415:
> i am not able to succesfully test this feature on [binder](https://mybinder.org/v2/gh/jupyterlab/lumino/main?urlpath=lab/tree/examples).
> 
> testing this feature is incredibly difficult with the lingering #14115 issue. the only way i could to `Tab` to the status bar was with no notebooks open. with no notebooks open, i could tab to the simple document switcher. once there i could NOT tab to the running session or kernel status. is the binder building as expected for y'all. i am not seeing the `tabindex=0` value that is expected.

Hi @tonyfast this is still a work in progress. Will switch it to draft and put it out on review when ready.

packages/apputils/src/kernelstatuses.tsx
j264415: Been looking at how to do this and cannot find a working mechanism. Any suggestions will be welcome. We cannot find a working solution without generating synthetic mouse clicks, which was rejected as a possible solution previously.

j264415:
Hi @krassowski . I have been looking at how to do this and cannot find a working mechanism. Any suggestions will be welcome. We cannot find a working solution without generating synthetic mouse clicks, which was rejected as a possible solution previously.

m158261:
Hey @krassowski, I've pushed changes that fix this functionality. The onKeyDown now mirrors onClick.

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -42,15 +43,30 @@ export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
...
+        event &&
+        (event.key === 'Enter' || event.key === 'Spacebar' || event.key === ' ')
+      ) {
+        event.preventDefault();
+        event.stopPropagation();
+        if (!item.model.sessionContext) {
+          return;
+        }
+        await sessionDialogs.selectKernel(item.model.sessionContext);
+      } else if (!event!.key) {
```
krassowski, line 57: This bit is hard to follow, could you decouple the key-handling logic (restore `changeKernel` as it was, and define a new function to be passed to `onKeyDown` that will invoke `changeKernel`)? Thank you!

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -49,8 +50,25 @@ export const kernelStatus: JupyterFrontEndPlugin<IKernelStatusModel> = {
...
+      event: KeyboardEvent<HTMLImageElement>
+    ) => {
+      if (
+        event.key === 'Enter' ||
+        event.key === 'Spacebar' ||
+        event.key === ' '
+      ) {
+        event.preventDefault();
+        event.stopPropagation();
+        changeKernel();
```
krassowski, line 63: This should either `return` or `await` the promise - otherwise lint will rightly fail.

```suggestion
        return changeKernel();
```

krassowski (approved):
Thank you @j264415 !

krassowski:
Of note this completes the ability to navigate around the main UI elements with keyboard alone (which is one of the highlights of 4.1 release).

PR #14766, opened 2023-06-28
400% Zoom screen usability enhancements - Updated

j264415:
Opened new PR to fix merge conflict that was causing #14626 test to fail but PR ended up being closed when syncing branch.
This new PR has all the changes of #14626 plus the changes asked for by @krassowski. 
## References

#14626 400% Zoom screen usability enhancements
#10004 Full low Vision Support (400% Zoom) (Mobile/Tablet-Responsive Support)

## Code changes
Changed padding and spacing to use min CSS function alongside vh (viewport-height) and vw(viewport-width) to reduce proportional spacing at high zoom and increase useful screen real estate usage.

## User-facing changes

Currently with the 400% zoom at 1280px the help main menu bar does not appear and within the file-browser tab bar only half of one 'dir-listing' shows up.
The changes I have made now allows for the help dropdown to be shown in the main menu. I found that when trying to check the changes the sidebar 'Extension Manger' disappears from view, so this was also changes so that users that needs this implemented would be able to have access to it without zooming back out. Furthermore, within the file-browser tab, users would now be able to see least 2 notebook directory listed below than before.

**Updated changes with this new PR:** 
The before changes meant the the UI was cutting off the menu on 100% zoom.
These are the changes made from the feedback changes the UI but at the point where it matters. These means that when returned to a 100% zoom menu back will not be cut off anymore and sidebar spacing will be fixed.

## Backwards-incompatible changes

None

PR #14854, opened 2023-07-19
Make status bar accessible at 400% zoom by hiding items with priority of zero (default)

### LINKED ISSUE #14793: Status bar at 400% zoom
### Problem
When using high zoom settings on lower resolution screens the status bar becomes unusable as the text and icons overlap.
You can see this by zooming to 400% in the browser on a screen less than 1280 pixels wide.

### Proposed Solution

In order to make the status bar accessible and usable at these zoom levels (principally low vision users) we suggest making a few adjustments at high zoom
Hide document title as it visible in the active notebook tab title
Hide kernel name (Python 3 - ipykernel) as this can be accessed by looking at the running tab 
Hide processor icon as this can be accessed by looking at the running tab 
Retain line and column number as not available easily elsewhere
Retain simple interface toggle as useful to have immediately available for users that require it 

### Additional context
None

j264415:
## References

- Closes #14793
- Relates to #14766

## Code changes

- new `priority`  argument is added to `IStatusBar.IRankItem`, which defaults to `0`
- when window is narrow (less or equal to 630 pixels, which is a heuristic for 400% zoom) the items with priority of zero get hidden so that the higher priority items remain visible 

## User-facing changes

Currently with 400% zoom at 1280px the status bar elements overlap each other which hinders usability. Unrequired elements have been removed and the remaining elements no longer overlap at 320px apparent width (1280px/400%).

This means that when a window is below 320px apparent width, the kernel title, document name and the processor icons will be hidden from view to allow space for the simple switch and the line-column number to be fully visible and accessible. If future text elements are added these will also be hidden unless they are given the priority-zoom class.

Extensions which add custom status bar items may want to set custom priority greater than one if these need to be visible at narrow screens.


j264415:
Please could this PR be reviewed.

j264415:
Please could this PR be reviewed. @gabalafou

gabalafou (changes requested):
While I think this PR has some great ideas, it also makes me uncomfortable because I feel like it's trying to achieve at the CSS level what should be achieved at the app/JS level.

For example, the StatusBar class provides a `registerStatusItem` method. This method is used by extensions, including built-in extensions, to add items to the status bar, like so:

```ts
// Add the status item to the status bar.
statusBar.registerStatusItem(lineColItem.id, {
  item,
  align: 'right',
  rank: 2,
  isActive: () => !!item.model.editor
});
```

It seems to me that this is the point at which something like your "priority-zoom" property should be passed.

I also question some of the decisions about what is and isn't priority in this PR. To me the notification icon is a better use of space than the simple-mode toggle—both because the toggle takes up a lot of space and also because the same command can be accessed in two other ways, via the menu and via the command palette; also there are other more obvious cues about when the app is in simple mode or the not, so I'm not sure how much value is added by always showing that state in the status bar.

I also question how effective this approach is because, for example, here's a screenshot on my laptop with the UI zoomed to 400% and it shows that the changes in this PR have done nothing to help because my laptop screen at 400% does not trigger the at-media rule:



I agree that the status bar is too cluttered at narrow width and/or high zoom. But I'm not sure this is the way to solve it. Or maybe this is the right direction, maybe the items in the status bar need to indicate some priority-level, and then if the status bar doesn't have enough width for all of the items, it starts to remove the lower-priority ones. But I think that functionality probably needs to implemented at the level of the StatusBar TypeScript class.

packages/statusbar/style/base.css
```diff
@@ -43,6 +43,21 @@
...
 }
 
+@media screen and (max-width: 360px) {
+  .jp-StatusBar-TextItem {
+    display: none;
+  }
+
+  .jp-StatusBar-TextItem.priority-zoom {
+    display: block;
+  }
```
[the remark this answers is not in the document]
j264415: Hi @gabalafou I have just made these changes you suggested.

j264415:
Hi @gabalafou looking at the comment you made above about implementing the changes that we want for the 400% zoom through the app/JS level and not CSS; I have made the relevant code changes required to achieve this. 
Please could you review these changes. 
Thank you

j264415:
please update galata snapshot

j264415:
Hi @gabalafou could you please review the changes I have made from your feedback. 
Thank you.

j264415:

j264415:

j264415:

j264415:

j264415:

j264415:

gabalafou:
I guess I wasn't clear enough in my previous review.

In your previous iteration of this PR, you targeted various elements in the status bar either explicitly (with the `priority-zoom` class) or implicitly (by targeting `.jp-StatusBar-TextItem`, `-Item`, and `-Group-Item`). 

You were basically creating a concept at the CSS level that some items in the status bar could be hidden at some zoom breakpoint while others could remain. And I was trying to say that the concept itself (the concept of whether an item could be hidden at higher zoom) should perhaps be implemented at a deeper level, at the code level. The idea is that the StatusBar class would have a notion that some of the items in the status bar are more important than others and if it starts running out of space, it could start hiding those in order to provide enough space for the more important status bar items. 

Maybe it's possible to re-use the `rank` property. Or maybe we need to introduce a new property like `priorityZoom`, I dunon.

My thinking was that this would allow the StatusBar class more flexibility in deciding when and how to remove items in the status bar. 

But maybe we don't need that level of flexibility. Maybe for now it's okay to use a special CSS class (such as `jp-priority-zoom`) with the following CSS rule:

```css
@media screen and (width <= 320px) {
  /* At high zoom or low width, remove all status bar 
     items that have not been marked as "priority" */
  .jp-StatusBar-Item:not(.jp-priority-zoom) {
    display: none;
  }
}
```

But one thing I would like to insist on is that we explicitly mark all status bar items that should remain visible at zoom. I don't want us to be implicit about it; I don't want to see selectors like `.jp-StatusBar-TextItem` in this PR. After the breakpoint, all status bar items should be hidden except for the ones that have a special CSS class. I'm not sure what that CSS class name should be, but perhaps `jp-priority-zoom` works.

gabalafou:
And you might be thinking something like: but I did move the concept of "priority" to the code level by changing the `isActive` functions. 

I think there are several issues with using the `isActive` function:

1. I'm not sure it's really possible to detect if the window is zoomed in that function, at least not in a robust way
2. I'm not sure that the function is called when the window is resized or zoomed
3. At the code level, I think that the status bar class should be responsible for handling layout and zoom, not each individual status item. In other words, I think that each status item should be allowed to tell the parent class (the status bar class) whether it thinks its important enough to remain on the screen, but the parent class should ultimately make the decision.

j264415:
> And you might be thinking something like: but I did move the concept of "priority" to the code level by changing the `isActive` functions.
> 
> I think there are several issues with using the `isActive` function:
> 
> 1. I'm not sure it's really possible to detect if the window is zoomed, at least not in a robust way
> 2. I'm not sure that the function is called when the window is resized or zoomed
> 3. At the code level, I think that the status bar class should be responsible for handling layout and zoom, not each individual status item. In other words, I think that each status item should be allowed to tell the parent class (that is, the StatusBar class) whether it thinks its important enough to remain on the screen, but the parent class should ultimately make the decision.

Hi @gabalafou I have made changes so that status bar widgets are hidden at the code level not the CSS level.

packages/statusbar/src/statusbar.ts
```diff
@@ -112,6 +123,13 @@ export class StatusBar extends Widget implements IStatusBar {
     super.dispose();
   }
 
+  private WindowZoomedLevel = () => {
+    // Get the current browser zoom level.
+    const browserZoomLevel = Math.round(window.devicePixelRatio * 100);
+    // Check if the zoom level is greater than or equal to 400%.
+    return browserZoomLevel >= 400;
+  };
```
gabalafou, line 131: Hmm, I'm not confident that this works. Have this been tested? Are we sure this works?

gabalafou (commented):
Hi @j264415, thanks for PR update. This is indeed in the direction I was thinking, but I'm not sure we can robustly detect browser zoom level. 

Would you like to schedule some time to pair program on this PR? I think if we were working on this together synchronously, we could hammer it out in less than an hour.

packages/statusbar/src/statusbar.ts
j264415: Yeah this has been tested and seems to work how we want it too.

packages/statusbar/src/statusbar.ts
```diff
@@ -42,16 +42,27 @@ export class StatusBar extends Widget implements IStatusBar {
...
       throw new Error(`Status item ${id} already registered.`);
     }
 
+    if (
+      id === '@jupyterlab/application-extension:mode-switch' ||
+      id === '@jupyterlab/apputils-extension:kernel-status' ||
+      id === '@jupyterlab/codemirror-extension:line-col-status' ||
+      id === '@jupyterlab/notebook-extension:mode-status'
+    ) {
+      statusItem.priorityZoom = 1;
```
fcollonval, line 57: Could you move this to the place the items are defined?

https://github.com/jupyterlab/jupyterlab/blob/5d7c44bdd89db72bf40ca7aeb8ecee92e07f7e06/packages/application-extension/src/index.tsx#L1183

https://github.com/jupyterlab/jupyterlab/blob/5d7c44bdd89db72bf40ca7aeb8ecee92e07f7e06/packages/apputils-extension/src/statusbarplugin.ts#L103

https://github.com/jupyterlab/jupyterlab/blob/5d7c44bdd89db72bf40ca7aeb8ecee92e07f7e06/packages/codemirror-extension/src/index.ts#L49

https://github.com/jupyterlab/jupyterlab/blob/5d7c44bdd89db72bf40ca7aeb8ecee92e07f7e06/packages/notebook-extension/src/index.ts#L431

packages/statusbar/src/statusbar.ts
```diff
@@ -42,16 +42,27 @@ export class StatusBar extends Widget implements IStatusBar {
    * @param statusItem - The item to add to the status bar.
    */
   registerStatusItem(id: string, statusItem: IStatusBar.IItem): IDisposable {
+    console.log(this._statusItems);
+    console.log(statusItem.item.title);
```
fcollonval, line 46: To be removed:

```suggestion
```

packages/statusbar/src/statusbar.ts
```diff
@@ -112,6 +123,13 @@ export class StatusBar extends Widget implements IStatusBar {
     super.dispose();
   }
 
+  private WindowZoomedLevel = () => {
```
fcollonval, line 126: Small wording change to fit with the naming convention

```suggestion
  private _isWindowZoomed = () => {
```

packages/statusbar/src/statusbar.ts
```diff
@@ -129,7 +147,10 @@ export class StatusBar extends Widget implements IStatusBar {
 
   private _refreshItem(id: string) {
     const statusItem = this._statusItems[id];
-    if (statusItem.isActive()) {
+    if (
+      statusItem.isActive() &&
+      !(this.WindowZoomedLevel() && statusItem.priorityZoom === 0)
```
fcollonval, line 152: Let's first test for the value comparison to avoid the function call if we can.

```suggestion
      !(statusItem.priorityZoom === 0 && this._isWindowZoomed())
```

packages/statusbar/src/tokens.ts
```diff
@@ -55,6 +55,11 @@ export namespace IStatusBar {
      */
     rank?: number;
 
+    /**
+     *  Displaying Items based on zoom priority -- higher zoom priority gets prioritised when zoom levels increase
+     */
+    priorityZoom?: number;
```
fcollonval, line 61: Could we call it `priority`? The idea is in the future to use that information to tune the status bar items displayed with the window width to have a better reactive UI.

```suggestion
    priority?: number;
```

fcollonval (changes requested):
Thanks @j264415 

I have some comments.

packages/statusbar/src/statusbar.ts
j264415: Hi @fcollonval just pushed up the changes you asked for.

packages/statusbar/src/statusbar.ts
gabalafou: I still have concerns about this approach. The intent, as I understand it, is to start hiding things in the status bar at 400% zoom. However, on my Mac's screen, at 100% zoom, `window.devicePixelRatio` is 2, so at 200%, the ratio is 4 and causes the status bar to start hiding stuff. So this creates inconsistent behavior between high pixel density screens versus regular ones. 

Another thing I noticed while testing this, not sure if it's related: when the file browser is open, the status bar does not start hiding items at 400% zoom. I actually first have to hide the status bar, *then* the items will hide at 400%. Does this happen for you too? Any idea what's going on?

j264415: Hi @gabalafou I have made a change to the status bar zoom functionality that should work no matter your device pixel ration. Please let me know if there is any other changes that need to be made.

packages/statusbar/src/statusbar.ts
```diff
@@ -112,6 +112,12 @@ export class StatusBar extends Widget implements IStatusBar {
     super.dispose();
   }
 
+  private _isWindowZoomed = () => {
+    // Check is the window width is less than or equal to 630px as this is when the staus bar become unaccessible to the user.
+    // Works across all screen resolution.
+    return window.innerWidth <= 630;
```
krassowski, line 118: Question from @gabalafou on the call: is there a justification?

packages/statusbar/src/statusbar.ts
```diff
@@ -112,6 +112,12 @@ export class StatusBar extends Widget implements IStatusBar {
     super.dispose();
   }
 
+  private _isWindowZoomed = () => {
+    // Check is the window width is less than or equal to 630px as this is when the staus bar become unaccessible to the user.
+    // Works across all screen resolution.
```
gabalafou, line 117:
```suggestion
    // The value for 630px was chosen by trial and error. 
    // When the screen width drops below 630px, there is no
    // longer enough space for all the items in the status bar 
    // (with notebook open), and items become clipped.
```

packages/statusbar/src/statusbar.ts
```diff
@@ -112,6 +112,12 @@ export class StatusBar extends Widget implements IStatusBar {
     super.dispose();
   }
 
+  private _isWindowZoomed = () => {
```
gabalafou, line 115:
```suggestion
  private _isWindowNarrow = () => {
```

packages/statusbar/src/statusbar.ts
```diff
@@ -129,7 +135,10 @@ export class StatusBar extends Widget implements IStatusBar {
 
   private _refreshItem(id: string) {
     const statusItem = this._statusItems[id];
-    if (statusItem.isActive()) {
+    if (
+      statusItem.isActive() &&
+      !(statusItem.priority === 0 && this._isWindowZoomed())
```
gabalafou, line 140:
```suggestion
      !(statusItem.priority === 0 && this._isWindowNarrow())
```

gabalafou (commented):
Thank you for working on this! I added just a few suggestions to reflect the last changes in the comments and method names

j264415:
> Thank you for working on this! I added just a few suggestions to reflect the last changes in the comments and method names

Thank you for reviewing this !

packages/statusbar/src/statusbar.ts
j264415: This number is because this is when the status bar starts to become inaccessible. The comment above has been reviewed by @gabalafou and changed to make the reasoning of this number clearer.

gabalafou:
This looks good to go from my perspective. The only check failing is UI Tests / Visual Regression, which may be unrelated to this PR.

If there are no objections, I will merge it in on Monday.

krassowski:
> This looks good to go from my perspective. 

Can you approve it? It seems there is still a an "X" from you. I often sort by approved when merging things before release.

image

> The only check failing is UI Tests / Visual Regression, which may be unrelated to this PR.

I think they are unrelated, but I kicked it

> If there are no objections, I will merge it in on Monday, December 18.

I might merge it earlier to fit it in `beta.0`.

krassowski (approved):
Thank you @j264415!

krassowski:
@j264415 I updated the top-level comment's Code Changes and Backwards-incompatible changes sections. Please take note for future PRs that we need to keep them aligned with the content of the PR and they can be brief, as long as the key take-away messages are conveyed :)

@j264415 can you please follow with another pull request adding a note to extension migration guide about the new `priority` attribute and what it entails for extension developers (see my note in Backwards-incompatible changes section, feel welcome to expand on this with an example).

j264415:
> @j264415 can you please follow with another pull request adding a note to extension migration guide about the new `priority` attribute and what it entails for extension developers (see my note in Backwards-incompatible changes section, feel welcome to expand on this with an example).

@krassowski All done. This is the PR: https://github.com/jupyterlab/jupyterlab/pull/15556

Issue #14857, opened 2023-07-19
CodeMirror should be configured to use absolute positioning of tooltips


krassowski:
## Description

Tooltips are incorrectly positioned in CodeMirror editor, which affects at least jupyterlab-lsp and jupyterlab-spellchecker extensions.

This is because we use css `contain`. CodeMirror states in the docs:

> By default, tooltips use "fixed" [positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/position), which has the advantage that tooltips don't get cut off by scrollable parent elements. However, CSS rules like contain: layout can break fixed positioning in child nodes, which can be worked about by using "absolute" here.

All we need to do is add an two liner CodeMirror extension: `tooltips({position: 'absolute'})`.

The alternative would be for each extension to add it but it would be wasteful - extensions should not need to worry about whether we use `contain` or not.

## Reproduce

[...]

PR #14858, opened 2023-07-19
Fix CodeMirror tooltip positioning and tooltip background

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14857 and background (see below).

This mostly concerns extensions, core lab does not make use of the built-in CodeMirror tooltips. Since extensions may want to let's fix it. Also, it is a trivial change.

## Code changes

- Add background color override for tooltips as currently we override font color (indirectly) but not background
- Configure tooltip positioning to fix it when used in JupyterLab which uses `contain: strict` for performance reasons

## User-facing changes

| Before | After |
|--|--|
|  Screenshot from 2023-07-19 19-54-26 | Screenshot from 2023-07-19 19-53-59 |

None

### DIFF
--- a/packages/codemirror/src/extension.ts
+++ b/packages/codemirror/src/extension.ts
@@ -26,7 +26,8 @@ import {
   keymap,
   lineNumbers,
   rectangularSelection,
-  scrollPastEnd
+  scrollPastEnd,
+  tooltips
 } from '@codemirror/view';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { JSONExt, ReadonlyJSONObject } from '@lumino/coreutils';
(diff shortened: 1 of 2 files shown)


fcollonval (commented):
Thanks @krassowski 

in https://github.com/jupyterlab/jupyter-collaboration/blob/a2671b2e226ba8163fb878885305f27ad9c77e3a/packages/collaboration/src/cursors.ts#L432 to allow displaying the user name when collaborating, I also added the parent to be `document.body` so that the tooltip is displayed outside of its containing widget (for the collaboration example that is needed if the cursor is in the editor first line).

Should we bring that one in also by default?

krassowski:
Yes, I think that attaching to body is fine. Though I wonder if we should explore defining [`tooltipSpace`](https://codemirror.net/docs/ref/#view.tooltips^config.tooltipSpace)⁠ function (too or instead).

fcollonval:
> Yes, I think that attaching to body is fine. Though I wonder if we should explore defining [`tooltipSpace`](https://codemirror.net/docs/ref/#view.tooltips%5Econfig.tooltipSpace)⁠ function (too or instead).

Thanks for the pointer I oversight that feat'. I guess it would be nice. I'm gonna approved this as we could do a follow-up with that feature. I let you press forward or merge :wink:

fcollonval (approved):

krassowski:

PR #14865, opened 2023-07-20
In defer mode render enough cells to fill the viewport

### LINKED ISSUE #14608: Benchmark UI test fails since switching to `defer` mode
Follow up to #14550 - I see this test to fail frequently:

`test/benchmark/notebook.spec.ts:94:9 › Benchmark › measure large_code_notebook.ipynb`

image

Here the notebook is clearly not fully rendered yet:

| Expected | Actual |
|--|--|
| large-code-notebook-ipynb-expected | large-code-notebook-ipynb-actual |

It seem to stem from the spinner being hidden before the deferred cells are shown. This is expected, but then I am not sure how we want to update the test to account for that. The relevant lines are:

https://github.com/jupyterlab/jupyterlab/blob/3fd8c726a22c636a447133345b1c1c27fc8a5674/galata/test/benchmark/notebook.spec.ts#L111-L131

fcollonval:
even if it takes more than a macro cycle

## References

Fixes #14608

## Code changes

In notebook defer rendering mode, cells are rendered to at least fill the viewport before starting rendering on idle time.

## User-facing changes

Cells should not pop-up when opening a notebook
Notebook opening time may increase

None

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -887,8 +887,19 @@ export class StaticNotebook extends WindowedList {
       cellIdx++;
     }
 
+    // If the notebook is not fully rendered
     if (cellIdx < this.cellsArray.length) {
-      this._scheduleCellRenderOnIdle();
+      // If we are defering the cell rendering and the rendered cells do
+      // not fill the viewport yet
+      if (
+        this.notebookConfig.windowingMode === 'defer' &&
+        this.viewportNode.clientHeight < this.node.clientHeight
+      ) {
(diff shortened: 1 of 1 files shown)


fcollonval (commented):
bot please run benchmark

fcollonval:
Benchmark tests are ok in term of cell rendered - what happen (too) often is that the cell toolbar is not always displayed:

image

So I consider this as good to go.

krassowski (approved):
LGTM

fcollonval:

Issue #13770, opened 2023-01-14
File browser: add setting to show the full path


jtpio:
### Problem

Notebook 7 reuses the same file browser widget as in JupyterLab. However since the file browsers takes the full page in `/tree` there would be plenty of space to show the full folder path:

![image](https://user-images.githubusercontent.com/46796385/123949145-8852f700-d9a2-11eb-80d0-1968fb66dee0.png)

This is less of an issue in JupyterLab because the file browser is in the `left` area and usually does not take that much space.

### Proposed Solution

Add a file browser setting to allow displaying the full path. Maybe make the widget scrollable (horizontally) in case the path is very long. Or provide a setting to set the max number of directories to show.

A more advanced logic would determine when to put an ellipsis automatically based on the current width.

### Additional context

This was originally reported in https://github.com/jupyter/notebook/issues/6404

jtpio:
Digging a little bit into the code, there seems to be some references to a `MAX_DIRECTORIES` here:

https://github.com/jupyterlab/jupyterlab/blob/1cda7df0e0005ac51fea4bbb8d68f44f6eb1e216/packages/filebrowser/src/crumbs.ts#L429-L430

There is also a hardcoded `2` here which might refer to the same max number of directories?

https://github.com/jupyterlab/jupyterlab/blob/1cda7df0e0005ac51fea4bbb8d68f44f6eb1e216/packages/filebrowser/src/crumbs.ts#L374

jtpio:
Adding to `4.0` as it would be nice to have such option that could be used downstream in Notebook 7, and also by regular JupyterLab users.

jtpio:
That would be great, thanks @dharmaquark!

zachwaldron4:
Hello, 
This appears to be an issue still be an issue.  If the feature to allow showing the full path is implemented, I can't seem to find it anywhere in the settings or view menus.

jtpio:
Thanks @zachwaldron4 for reporting here.

Indeed, when the Settings Editor UI does not seem to be showing anything when searching for `showFullPath`:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/594bc041-e91f-40e4-885d-98dcbfb0a26b)

Although the "Advanced Settings Editor" does:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/4f1aad87-d098-40bd-a168-be02c9bf1b44)

[...]

PR #14866, opened 2023-07-20
Add a setting to show full path in breadcrumbs

dharmaquark:
## References

This PR introduces support for full path breadcrumbs. Fixes https://github.com/jupyterlab/jupyterlab/issues/13770

## Code changes

To support full paths while maintaining the default ellipsis behavior, this PR adds a new BreadCrumbs option `fullPath` - if full path is set to true, the path will not be compressed as a user drills into a multi-level file directory.

## User-facing changes

Navigation behavior if `fullPath` is set to true:
https://github.com/jupyterlab/jupyterlab/assets/14363644/cfca9224-6bc1-4357-a3dc-1809fe2f76fe

Navigation behavior if `fullPath` is set to false or not set:
https://github.com/jupyterlab/jupyterlab/assets/14363644/220e760e-8562-4210-b623-f4b528425d7a

This should be backwards compatible, if `fullPath` is not set, the bread crumbs will fall back to the default existing ellipsis behavior.


jtpio:
Thanks @dharmaquark for working on this!

Should there also be a setting to enable / disable showing the full path? So end users can change the value via the Settings Editor.

jtpio:
> would it make sense to still keep the `BreadCrumbs` option as well? or just control this via the settings editor

I think it would still make sense to keep it. Also the handling of the setting would likely be implemented in the `filebrowser-extension` package, which would then set / unset `fullPath`.

krassowski:
@dharmaquark do you need help finishing this PR?

dharmaquark:
> @dharmaquark do you need help finishing this PR?

apologies for the delay- I can work on adding this to the settings this week!

dharmaquark:
> Discussion from lab call: it would be good to merge this in for 4.1. The soft deadline for having a PR ready for review is 1st November, but we can still merge it afterwards.

Ok thanks for the heads up on the deadline! I got delayed this month, but will have time to carve out next week to wrap this up with the settings logic

dharmaquark:
Thank you for the patience here! I am looking forward to finishing this on my end :)

dharmaquark:
added the setting in `filebrowser-extension` (via a setter/getter in browser => crumb classes); I need to do another round of testing (ran into headaches getting jupyterlab building on a new M2, that canvas build issue, didn't seem to be resolved by brew installing the extra packages), going to try to build on my older intel Mac tomorrow AM.

dharmaquark:
ok, got things building on an M1- made a few testing patches (needed to add the new setting to the schema)

packages/filebrowser/src/crumbs.ts
```diff
@@ -309,6 +319,15 @@ export class BreadCrumbs extends Widget {
   private _hasPreferred: boolean;
   private _crumbs: ReadonlyArray<HTMLElement>;
   private _crumbSeps: ReadonlyArray<HTMLElement>;
+  private _fullPath: boolean;
+
+  get fullPath(): boolean {
+    return this._fullPath;
+  }
+
+  set fullPath(value: boolean) {
```
dharmaquark, line 328: I'm wondering here if I should check `options.fullPath`? otherwise I think the default value in `fileBrowserConfig` might always override a true `options.fullPath`? assuming, if we go with adding the toggle command, that wouldn't check the options setting?

krassowski: This is fine as-is, the `options` is mostly for testing, ensuring there is a value to start with testing, and to enable standalone usage (when another application building upon jupyterlab core packages rewrites their own `-extension` packages)

packages/filebrowser/src/crumbs.ts
```diff
@@ -309,6 +319,15 @@ export class BreadCrumbs extends Widget {
   private _hasPreferred: boolean;
   private _crumbs: ReadonlyArray<HTMLElement>;
   private _crumbSeps: ReadonlyArray<HTMLElement>;
+  private _fullPath: boolean;
+
+  get fullPath(): boolean {
```
krassowski, line 324: Can you please move these two methods to above the private/protected methods (around the line 116) and add a documentation strings before the getter?

krassowski (approved):
Thank you @dharmaquark!

PR #14870, opened 2023-07-23
Align the style of standard input to match ipywidgets style

### LINKED ISSUE #14458: The input (stdin) styling does not match that of JupyterLab
We style the input area in a consistent way with the rest of the JupyterLab UI. In the screenshop below, I show the looks of the `Text` widget first, to highlight the difference.

Screenshot 2023-04-28 at 11 28 14

alewilliam789:
## References

Fixes #14458 

## Code changes

Changes HTML/CSS to match the Text widget from ipywidgets and follows same styles.

## User-facing changes

The input widget now matches the Text Widget in the output area.

Input before changes: 
image

Input after changes:
image

No it really only styles the Stdin widget.


fcollonval (changes requested):
Thanks @alewilliam789 

One of the integration test failing is legit:

This is the test failing: https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/galata/test/jupyterlab/outputarea-stdin.test.ts#L11

The reason is the styling that changes (as expected :wink:) - before is left / this PR is right

image

That case makes it unclear the box can be edited. Would you mind improving the styling for that case?

You can find the result of the integration test (including a video of it as artifact `jupyterlab-galata-test-assets` at the bottom of https://github.com/jupyterlab/jupyterlab/actions/runs/5646058234)

alewilliam789:
@fcollonval At this point it looks like the only reason this last UI test is failing is because the snapshot being used has the old styling. After reading the Galata README it looks like new benchmarks are made after an approved PR review and it is requested in the code review correct?

ohrely:
@alewilliam789 Thanks for this contribution!  The failing visual regression test is something you can fix locally by taking the test's output screenshot and using it to replace the reference screenshot.  Alternatively you can add "please update snapshots" to the PR description to trigger a bot to do this for you.

I just rebased this branch, so you'll want to pull the current state of the PR before making further changes.  It looks like there's a lint issue turned up in one of the tests that will need to be addressed before merge.

galata/test/jupyterlab/outputarea-stdin.test.ts-snapshots/stdin-history-search-jupyterlab-linux.png
krassowski, line 1: This is definitely nicer, but the wide indent which almost centers the input on smaller screens seems a bit odd to me.

PR #14882, opened 2023-07-30
Improve upgrade extension script

fcollonval:
## References

Fixes https://github.com/jupyterlab/extension-template/issues/36

## Code changes

- Remove `prepare` script when upgrading an extension
- Add more workflows recommended to override
- Better cleaning of the python package name when using the npm package name

## User-facing changes

None

None

### DIFF
--- a/jupyterlab/upgrade_extension.py
+++ b/jupyterlab/upgrade_extension.py
@@ -31,6 +31,9 @@ RECOMMENDED_TO_OVERRIDE = [
     ".github/workflows/binder-on-pr.yml",
     ".github/workflows/build.yml",
     ".github/workflows/check-release.yml",
+    ".github/workflows/enforce-label.yml",
+    ".github/workflows/prep-release.yml",
+    ".github/workflows/publish-release.yml",
     ".github/workflows/update-integration-tests.yml",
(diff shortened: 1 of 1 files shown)


fcollonval:
@krassowski if you have seen other things that could be improved let me know.

krassowski:
The only thing that got me was `stylelint` configuration being in `package.json`. The recommend to override list still includes `.stylelintrc`, `.prettierrc` and `.eslintrc.js` even though these are now in `package.json`. I needed to copy the keys over manually - I am not sure if this is because of an answer I given or if this is not handled by the script. Somewhat related but on the template side, I needed to add `selector-class-pattern` to allow JupyterLab style class names.

krassowski (approved):
As for the changes that are already in, LGTM. Not sure what you think about the lint config situation (an alternative to improving the upgrade script would be to revert the template to have them in individual files).

fcollonval:
The script does not import the styling config from the template (only the `scripts` and `(dev)dependencies` are processed. But I agree with you, I found myself always switching the config for the linter. So I'll push a commit to improve the upgrade script to handle that.

fcollonval:

PR #14901, opened 2023-08-02
Save and restore sidebar subpanels sizes and expansion states

DenisaCG:
This PR makes possible the full saving and restoration of the sidebars, including the subpanels sizes and expansion states.

## References

It listens to the new signal introduced  in this PR https://github.com/jupyterlab/lumino/pull/614 in `lumino`, namely [`expansionToggled`](https://github.com/jupyterlab/lumino/blob/95f69a3a0caba53bce4a642b1f1e47b7fd7e4006/packages/widgets/src/accordionpanel.ts#L60), such that the saving of the sidebar's layout is triggered when one of the subpanels is collapsed or expanded.

It also listens to the signal [`handleMoved`](https://github.com/jupyterlab/lumino/blob/95f69a3a0caba53bce4a642b1f1e47b7fd7e4006/packages/widgets/src/splitpanel.ts#L115), which makes it possible for the layout of the sidebar to also be saved when the vertical size of any of the subpanels is changed. The horizontal size of the entire sidebar is already being saved, as a consequence of the layout saving of the main dock panel.

## Code changes

To allow for the full saving and restoration of the sidebar layout, the `ISideArea` interface has been extended, such that it includes two new properties `sizes` and `expansionStates` which will store the vertical sizes, and respectively the expansion state of each subpanel.

The listeners for the signals mentioned above are also initialized when a widget is added to the sidebar, in the case of an `AccordionPanel` or `SidePanel`.

When a saving of the sidebar's layout is triggered, the `dehydrate` method now extracts the vertical sizes, as well as the expansion states, while the `rehydrate` method will restore them.

## User-facing changes

The users will now be able to see the sidebars' layout being automatically saved with any changes that might have been made, regarding sizes of subpanels or their expansion states, and respectively restored in the case of refreshing or reopening JupyterLab.

image

### DIFF
--- a/packages/application/src/shell.ts
+++ b/packages/application/src/shell.ts
@@ -7,6 +7,7 @@ import {
   classes,
   DockPanelSvg,
   LabIcon,
+  SidePanel,
   TabBarSvg,
   tabIcon,
   TabPanelSvg
@@ -17,6 +18,7 @@ import { IMessageHandler, Message, MessageLoop } from '@lumino/messaging';
 import { Debouncer } from '@lumino/polling';
 import { ISignal, Signal } from '@lumino/signaling';
 import {
+  AccordionPanel,
   BoxLayout,
   BoxPanel,
   DockLayout,
@@ -276,6 +278,16 @@ export namespace ILabShell {
      * The collection of widgets held by the sidebar.
      */
     readonly widgets: Array<Widget> | null;
+
+    /**
+     * Vertical sizes of the widgets situated in the sidebar.
+     */
+    readonly sizes: Array<number> | null;
+
+    /**
+     * Expansion states of the widgets situated in the sidebar.
+     */
+    readonly expansionStates: Array<boolean>;
   }
 }
 
@@ -1910,6 +1922,20 @@ namespace Private {
     }
 
     /**
+     * Handles a movement to the handles of a widget
+     */
+    private _onHandleMoved(): void {
+      return this._refreshVisibility();
+    }
+
+    /**
+     * Handles changes to the expansion status of a widget
+     */
+    private _onExpansionToggle(sender: AccordionPanel, index: number): void {
+      return this._refreshVisibility();
+    }
+
+    /**
      * Expand the sidebar.
      *
      * #### Notes
@@ -1967,7 +1993,6 @@ namespace Private {
       // Store the parent id in the title dataset
       // in order to dispatch click events to the right widget.
       title.dataset = { id: widget.id };
-
       if (title.icon instanceof LabIcon) {
         // bind an appropriate style to the icon
         title.icon = title.icon.bindprops({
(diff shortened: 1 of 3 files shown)


packages/application/test/layoutrestorer.spec.ts
```diff
@@ -60,13 +60,17 @@ describe('apputils', () => {
             collapsed: true,
             currentWidget: null,
             widgets: null,
-            visible: false
+            visible: false,
+            sizes: null,
+            expansionStates: [true, true, true, true]
```
fcollonval, line 65: Here is the trouble, a sidebar can have multiple panels (called _widgets_ in this object). So `sizes` and `expansionStates` should exist per panel (child of the stack panel).

The panel are identify by a _id_ for example

```json
"left":{
  "collapsed":false,
  "visible":true,
  "current":"filebrowser",
  "widgets":["filebrowser","running-sessions","@jupyterlab/toc:plugin","extensionmanager.main-view"]
},
"right":{
  "collapsed":true,
  "visible":true,
  "widgets":["jp-property-inspector","debugger-sidebar"]
}
```

So I would use a object like

```js
left/rightArea: {
  collapsed: boolean,
  currentWidget: string | null
  widgets: string[] | null,
  widgetStates: {
    [id: string]: {
     sizes: number[] | null,
     expansionStates: boolean[] | null
    }
  }
}
```

Does it make sense?

fcollonval (changes requested):
Thanks @DenisaCG 

The code is good. We are only missing one point: the `sizes` and `expansionStates` must be stored per children of the sidebar stack panel.

packages/application/src/shell.ts
```diff
@@ -1993,11 +2021,35 @@ namespace Private {
       const collapsed = this._sideBar.currentTitle === null;
       const widgets = Array.from(this._stackedPanel.widgets);
       const currentWidget = widgets[this._sideBar.currentIndex];
+      let sizes = new Array<number>();
+      let expansionStates = new Array<boolean>();
+      Object.entries(this).forEach(([key, value]) => {
```
fcollonval, line 2026: I don't think this is doing what you want. You should loop on `this._stackedPanel.widgets` not `this` I guess. And then test if `panel.content` is an instance of `SplitPanel` (you don't need to test for `AccordionPanel` as it inherits from `SplitPanel`).

> Tip: you can directly use the local `widgets` variables.

packages/application/test/layoutrestorer.spec.ts
DenisaCG: Oh okay. I assumed that each panel would require its own sidebar instance so we would have multiple left and right ones, but this makes more sense. Thanks @fcollonval! I will make the changes.

DenisaCG (commented):
Thanks for the review @fcollonval! I restructured the code such that the sizes and expansion states of the subpanels of each sidebar instance is included in `widgetStates`, each also being identified by their widget id. Unfortunately, it seems to be missing something as the layout is not saved and restored properly.

fcollonval:
Thanks a lot @DenisaCG this is working as expected. The integration tests should pass after https://github.com/jupyterlab/jupyterlab/pull/14945 is merged.

Issue #14883, opened 2023-07-30
Should we expose `IShell.activeWidget`, or short-circuit in `IShell.activateById`?


krassowski:
### Problem

Extensions may want to activate a widget if it is not active. This is because activating a widget may have side effects. In case of notebooks the side effect of ensuring focus can lead to problematic behaviour in CodeMirror editor extensions, such as jupyterlab-vim. In https://github.com/jupyterlab-contrib/jupyterlab-vim/pull/85 I workaround it by comparing the widget with `IShell.currentWidget` but it feels like we should be comparing against `IShell.activeWidget`; however that is not exposed as public API.

### Proposed Solution

a) expose `IShell.activeWidget`,
b) short-circuit in `IShell.activateById` if provided id is equal `IShell.activeWidget.id`

Downsides:
- (a) increases API surface
- (b) may break existing code if it is relying on side effects always being invoked

### Additional context

Implementation has `activateById`, `activeWidget`, and `currentWidget`:

<details>

https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/packages/application/src/shell.ts#L499-L505

https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/packages/application/src/shell.ts#L547-L552

https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/packages/application/src/shell.ts#L743-L746

</details>

Public API only exposes `activateById` and `currentWidget`: 

<details>

https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/packages/application/src/frontend.ts#L242-L251

https://github.com/jupyterlab/jupyterlab/blob/e2cdaf1f01457195a7fb73bd25924a543343c78c/packages/application/src/frontend.ts#L270-L277

</details>

fcollonval:
I support you with this proposal.

Issue #14782, opened 2023-07-02
Expose `currentChanged` in `JupyterFrontEnd.IShell`


krassowski:
### Problem

`JupyterFrontEnd.IShell` defines `currentWidget` but not `currentChanged`. However, both `INotebookShell` from `@jupyter-notebook/application` and `ILabShell` do implement `currentChanged` (although with slightly different signature).

This makes it hard to write application independent code as we need to cast `app.shell` to access `currentChanged`.

### Proposed Solution

- Solution 1 (strict): Add `currentChanged` signal with permissive generics to `JupyterFrontEnd.IShell`. This means that all applications would need to implement it. For applications which only ever have one widget, having it as a dummy signal which never emits would be explicitly allowed in documentation.
- Solution 2 (permissive): as above but make it optional
- Solution 3: do nothing.

### Additional context

This is nice to have for 5.0.

krassowski:
`jupyter-ai` is another extension which would benefit from it, as it currently [just does not support selection tracking in notebook](https://github.com/jupyterlab/jupyter-ai/blob/d6d34a59bd7289e30fdb024f1c4cb03f0bed6535/packages/jupyter-ai/src/selection-watcher.ts#L102-L109):

```ts
    if (!(shell instanceof LabShell)) {
      throw 'Shell is not an instance of LabShell. Jupyter AI does not currently support custom shells.';
    }

    this._shell = shell;
    this._shell.currentChanged.connect((sender, args) => {
      this._mainAreaWidget = args.newValue;
    });
```

PR #14920, opened 2023-08-04
Improve LSP API

### LINKED ISSUE #14710: How to get `WidgetLSPAdapter`/`VirtualDocument` from CodeMirror event?
## Description

In original implementation of LSP the features were receiving sufficient context connecting `CellCodeEditor` to the `WidgetAdapter`/`VirtualDocument` via `EditorAdapter`. The version merged into the core does **not** have the concepts of `EditorIntegration` nor `EditorAdapter` thus:
- features have to assume the editor is CodeMirror6 (cannot provide integration for other editors) - this was intentional, though maybe short-sighted
- when an event is triggered in a code cell editor there is no obvious way to associate it with the corresponding `WidgetLSPAdapter` nor `VirtualDocument`; this prevents reimplementation of many LSP features, with the solutions varying from error-prone to duplicating the adapter pattern which is not great maintenance-wise.

I would love to get some opinions on the preferred solution among the few proposed below, or learn about other solutions you may think of.

### Previous

```mermaid
flowchart TD
    B
    B[LSP feature, e.g. highlights] -->|provides EditorIntegration factory| C[WidgetAdapter]
    B -->|provides capabilities information| CON[ConnectionManager]
    C --> |manages| MA[Cell <=> VirtualDocument]
    C -->|for each cell creates| EI[EditorAdapter with one or more EditorIntegration]
    EI -->|registers handlers on| D[CodeMirror cell editor]
    D -->|user event, e.g. moves cursor| E[Event]
    EI --> |knows about WidgetAdapter| H[EventHandler]
    E --> |does not know about WidgetAdapter| H
    H --> X(clears highlight on other cells using adapter)
    H --> Y(adds highlight on current cell)  
```

<details>

```typescript
class WidgetAdapter {
  connect_adapter(
    virtual_document: VirtualDocument,
    connection: LSPConnection,
    features: IFeature[] | null = null
  ): EditorAdapter<any> {
    let adapter = this.create_adapter(virtual_document, connection, features);
    this.adapters.set(virtual_document.id_path, adapter);
    return adapter;
  }
  private create_adapter(
    virtual_document: VirtualDocument,
    connection: LSPConnection,
    features: IFeature[]
  ): EditorAdapter<IVirtualEditor<IEditor>> {
    let adapter_features = new Array<
      IFeatureEditorIntegration<IVirtualEditor<IEditor>>
    >();

    for (let feature of features) {
      let featureEditorIntegrationConstructor =
        feature.editorIntegrationFactory.get(this.virtual_editor.editor_name)!;
      let integration = new featureEditorIntegrationConstructor({
        feature: feature,
        virtual_editor: this.virtual_editor,
        virtual_document: virtual_document,
        connection: connection,
        adapter: this,
      });
      adapter_features.push(integration);
    }

    let adapter = new EditorAdapter(
      this.virtual_editor,
      virtual_document,
      adapter_features
    );
    connection.sendOpenWhenReady(virtual_document.document_info);
    return adapter;
  }
}
```

</details>

### Current

```mermaid
flowchart TD
    C[WidgetAdapter] --> |manages| MA[Cell <=> VirtualDocument]
    C --> |registers on| CON
    B -->|registers handlers on CodeMirror factory| D[CodeMirror cell editor]
    B[LSP feature, e.g. highlights] -->|provides LSP capabilities information| CON[ConnectionManager]
    B --> |does NOT know about WidgetAdapter| H[EventHandler]
    D -->|user event, e.g. moves cursor| E[Event]
    E --> |does NOT know about WidgetAdapter| H
    H --> X(cannot clears highlight on other cells using adapter)
    H --> Y(adds highlight on current cell)
  ```

<details>

https://github.com/jupyterlab/jupyterlab/blob/4a267dfc1e3241ed936da6c60f80d11422ab1e55/packages/lsp/src/connection_manager.ts#L207-L218

</details>

The reason it worked for completer which was the feature demonstrated when merging into core is that completer has its own bindings for file editor and notebook. Doing this for each feature will be prohibitive maintenance-wise and is currently impossible as explained in *Do not register event handlers on CodeMirror level* section.

### "Just query by path, this is the new API"

This does not work: we do not know the path for the current editor.

```mermaid
flowchart TD
    C[WidgetAdapter] --> |manages| MA[Cell <=> VirtualDocument]
    B -->|registers handlers on CodeMirror factory| D[CodeMirror cell editor]
    B[LSP feature, e.g. highlights] -->|provides LSP capabilities information| CON[ConnectionManager]
    B --> |does NOT know about WidgetAdapter| H[EventHandler]
    D -->|user event, e.g. moves cursor| E[Event]
    E --> |does NOT know about WidgetAdapter| H
    C --> |registers on| CON
    CON --> FA
    H --> |Problem: path not known| FA["adapter = connectionManager.adapters.get(path)"]
```

### "Just find the adapter by current widget"

This does not work:
- in multi-pane layout when multiple features are active in background (but only one widget is current)
- when user switches current widget we would get race conditions

```mermaid
flowchart TD
    C[WidgetAdapter] --> |manages| MA[Cell <=> VirtualDocument]
    B -->|registers handlers on CodeMirror factory| D[CodeMirror cell editor]
    B[LSP feature, e.g. highlights] -->|provides LSP capabilities information| CON[ConnectionManager]
    B --> |does NOT know about WidgetAdapter| H[EventHandler]
    D -->|user event, e.g. moves cursor| E[Event]
    E --> |does NOT know about WidgetAdapter| H
    C --> |registers on| CON
    CON --> FA
    H --> FA["adapter = findAdapter(Shell.currentWidget, connectionManager.adapters)"]
    FA --> Ops[clears WRONG editors if user switched widgets in the meantime]
```

### "Do not register event handlers on CodeMirror level"

Indeed, if we register event handlers going top-down: on each cell and each file editor, we do know the path (hooray). The problems are:
- A) each feature needs to know about notebook and editor leading to a lot of boilerplate and maintenance burden
- B) AFAIK there is no API to register CodeMirror handlers on Notebook/Editor level (modify their editor factory) at this point (there is either `injectExtension` which happens after factory was called (so features would need to call it on each cell/keep track of every cell), or `editorExtensionRegistry.addExtension` which is global)

```mermaid
flowchart TD
    B[LSP feature, e.g. highlights] -->|provides LSP capabilities information| CON[ConnectionManager]
    C[WidgetAdapter] --> |manages| MA[Cell <=> VirtualDocument]
    EditorTracker --> EditorAdapter
    NotebookTracker --> EditorAdapter
    EditorAdapter --> CellCreatedEvent
    CellCreatedEvent --> |handler knows about WidgetAdapter| H[EventHandler]
    B -->|"implements (problem A)"| EditorAdapter
    CellCreatedEvent -->|"registers handlers on cell CodeMirror (problem B)"| D[CodeMirror cell editor]
    D -->|user event, e.g. moves cursor| E[Event]
    E --> |event does NOT know about widget| H
    C --> |registers on| CON
    CON --> WA
    H --> |knows about widget| WA["adapter = findAdapter(connectionManager.adapters, widget)"]
```

### "Maintain a (WeakRef)Map from CM editor to WidgetAdapter"

This would work but requires implementation in core: each `WidgetAdapter` should add to the map when they become aware of new editor instance. The question is where would the editor ⇒ adapter mapping reside? 

```mermaid
flowchart TD
    C[WidgetAdapter] --> |manages| MA[Cell <=> VirtualDocument]
    C --> |on cell creation update| MAP["mapping = Map(CodeMirror, Adapter)"]
    B -->|registers handlers on CodeMirror factory| D[CodeMirror cell editor]
    B[LSP feature, e.g. highlights] -->|provides LSP capabilities information| CON[ConnectionManager]
    B --> |does NOT know about WidgetAdapter| H[EventHandler]
    D -->|user event, e.g. moves cursor| E[Event]
    E --> |does NOT know about WidgetAdapter| H
    C --> |registers on| CON
    CON --> FA
    H --> |knows emitter editor| FA["adapter = mapping.get(editor)"]
```

## Context

- JupyterLab version: 4.0.2

### LINKED ISSUE #14783: There is no sufficient access to LSP adapters (`WidgetLSPAdapter`) from widgets
### Problem

This is similar to https://github.com/jupyterlab/jupyterlab/issues/14710 which describes the need to access adapters from editors.

Access to adapter from widget is needed to integrate some features, e.g.:
- executing actions from context menu
- updating diagnostics panel when switching editors

Currently:
- there is no mapping from widget adapter
- there is no signal emitted when new adapter gets added

This leads to suboptimal workaround like below:

```ts
app.shell.currentChanged.connect((shell) => {
    const adapter = [...connectionManager.adapters.values()].find(adapter =>
        adapter.widget == shell.currentWidget
    );

    if (!adapter) {
      this.console.debug('No adapter found');
    } else {
      diagnosticsFeature.switchDiagnosticsPanelSource(adapter);
    }
});
// now the problem is that if user switches widget but adapter activates
// after the `app.shell.currentChanged` signal fired we have no way to know.
```

or

```ts
app.shell.currentChanged.connect((shell) => {
    if (!shell.currentWidget) {
      this.console.debug('No current widet');
      return;
    }

    const context = (shell.currentWidget as IDocumentWidget).context;
    if (!context) {
      this.console.debug('Not a document widget');
      return;
    }
    // the problem here is that renaming file will break this - adapters in connectionManager would need to listen to path changes and update mapping.
    // also, what happens if there are multiple views for the same document? It would not work!
    const adapter = connectionManager.adapters.get(context.path);

    if (!adapter) {
      this.console.debug('No adapter found');
    } else {
      diagnosticsFeature.switchDiagnosticsPanelSource(adapter);
    }
});
```

### Proposed solution

One could argue that `AdapterManager` which was removed among comments about too many _managers_ was actually needed, but if we just want to keep adding things to `connectionManager`, we could add:
- `adapterChanged` signal which emits when widget is changed or new adapter is registered on current widget (and thus current adapter changes)

### Additional context

The old LSP extension had the following [`ILSPAdapterManager` interface](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/ff8b6c44d5a81a7bc051604ea37a5bc4da6ec94a/packages/jupyterlab-lsp/src/tokens.ts#L141):

```ts
export interface ILSPAdapterManager {
  adapterTypeAdded: Signal<
    ILSPAdapterManager,
    IAdapterTypeOptions<IDocumentWidget>
  >;
  adapterChanged: Signal<ILSPAdapterManager, WidgetAdapter<IDocumentWidget>>;
  adapterDisposed: Signal<ILSPAdapterManager, WidgetAdapter<IDocumentWidget>>;
  currentAdapter: WidgetAdapter<IDocumentWidget>;
  isAnyActive: () => boolean;
  registerExtension(extension: LSPExtension): void;
  registerAdapterType(options: IAdapterTypeOptions<IDocumentWidget>): void;
  readonly types: IAdapterTypeOptions<IDocumentWidget>[];
}
```

hbcarlos:
## References
Fixes #14710
Fixes #14783

## Code changes
Improves the LSP API by including CodeMirror extension factories in LSP features. With this new architecture, we can create LSP features with CodeMirror extensions that will be injected into the editor only if the LSP feature is available. In addition,  the `EditorAdapter.ILSPEditorExtensionFactory` enhances the API of `IEditorExtensionFactory` to pass more context to the CM extension by adding the document path and WidgetLSPAdapter.

This PR also changes ILSPDocumentConnectionManager's API to track the lifecycle of the adapters by adding multiple signals.

Yes. There are additions to the public LSP API, but we should treat them as bug fixes and release them on v4.1 because they are necessary for migrating JupyterLab-LSP to Lab v4

<details>
  <summary>Example plugin for adding a new LSP feature</summary>

```ts
const renamePlugin: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/lsp-extension:rename',
  description: 'Provides the language server settings.',
  autoStart: true,
  requires: [ILSPDocumentConnectionManager, ILSPFeatureManager, IWidgetLSPAdapterTracker],
  activate: (app: JupyterFrontEnd, connectionManager: ILSPDocumentConnectionManager, featureManager: ILSPFeatureManager, tracker: IWidgetLSPAdapterTracker) => {
    tracker.adapterAdded.connect((sender, adapter) => {
      console.debug("Adapter added:", adapter.documentPath);
    });
    tracker.adapterUpdated.connect((sender, adapter) => {
      console.debug("Adapter changed:", adapter.documentPath);
    });
    tracker.currentChanged.connect((sender, adapter) => {
      console.debug("Current Adapter changed:", adapter?.documentPath);
    });
  
    const FEATURE_ID = "rename_symbol";
    const extensionFactory: EditorAdapter.ILSPEditorExtensionFactory = {
      name: FEATURE_ID,
      factory: (options) => {
        const { editor, widgetAdapter } = options;
    
        if (!widgetAdapter.virtualDocument) {
          return null;
        }
    
        const connection = connectionManager.connections.get(widgetAdapter.virtualDocument.uri);
        if (!connection || !connection.provides('renameProvider')) {
          return null;
        }
    
        const ext = EditorView.domEventHandlers({ dblclick: (e, view) => {
          const range = editor.getSelection();
          const res = connection.clientRequests['textDocument/rename'].request({
            newName: "test",
            position: { line: range.start.line, character: range.start.column },
            textDocument: { uri: widgetAdapter.virtualDocument!.uri }
          });
      
          res.then(value => {
            console.debug(value);
          }).catch(e => console.error);
        }});
    
        return EditorExtensionRegistry.createConditionalExtension(ext);
      }
    }
  
    featureManager.register({
      id: FEATURE_ID,
      extensionFactory
    });
  }
};
```
</details>


packages/lsp/src/connection_manager.ts
```diff
@@ -42,6 +42,13 @@ export class DocumentConnectionManager
     Private.setLanguageServerManager(options.languageServerManager);
   }
 
+  /**
+   * The path of the active widget adapter.
+   */
+  get currentAdapter(): string | null {
+    return this._currentAdapter;
+  }
```
hbcarlos, line 50: The current adapter contains the path to the active adapter instead of the adapter itself, because when adding a new adapter,  `app.shell.currentChanged` is emitted before the adapter is registered.

packages/lsp/src/connection_manager.ts
```diff
@@ -120,6 +127,48 @@ export class DocumentConnectionManager
...
+  /**
+   * Signal emitted when the current adapter changes.
+   *
+   * Emits null when there is no current adapter.
+   */
+  get currentAdapterChanged(): ISignal<
+    ILSPDocumentConnectionManager,
+    string | null
+  > {
+    return this._currentAdapterChanged;
```
hbcarlos, line 139: We have the same problem as described in the `currentAdapter`.

packages/lsp/src/connection_manager.ts
```diff
@@ -198,6 +247,16 @@ export class DocumentConnectionManager
...
 
+  /**
+   * Update the active adapter.
+   *
+   * @param  path - path to the inner document of the adapter or null.
+   */
+  updateCurrentAdapter(path: string | null): void {
+    this._currentAdapter = path;
+    this._currentAdapterChanged.emit(path);
+  }
```
hbcarlos, line 258: Same problem as before. In addition, here, we are emitting `currentAdapterChanged` before the adapter is registered. When a plugin subscribes to this signal, the plugin needs to check whether the adapter exist or wait for `adapterAdded` signal.

A possible solution could be to debounce emitting `currentAdapterChanged` until the adapter exists.

packages/lsp/src/adapters/adapter.ts
```diff
@@ -592,6 +599,32 @@ export abstract class WidgetLSPAdapter<T extends IDocumentWidget>
...
+  private _onEditorAdded(
+    sender: WidgetLSPAdapter<T>,
+    change: IEditorChangedData
+  ): void {
+    const { editor } = change;
+    const editorAdapter = new EditorAdapter({
+      editorReady: editor.ready,
+      getEditor: editor.getEditor,
+      path: this.documentPath,
+      adapter: this,
```
trungleduc, line 613: Since `EditorAdapter` has also `Adapter` in its name
```suggestion
      widgetAdapter: this,
```

packages/lsp-extension/src/index.ts
```diff
@@ -302,12 +304,36 @@ function addRunningSessionManager(
     )
   });
 }
+
+const currentAdapter: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/lsp-extension:currentAdapter',
+  description: 'Provides the current `WidgetLSPAdapter`.',
```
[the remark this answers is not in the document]
krassowski: More details in https://github.com/jupyterlab/jupyterlab/issues/14783#issuecomment-1669988642

packages/lsp-extension/src/index.ts
```diff
@@ -302,12 +304,36 @@ function addRunningSessionManager(
...
+    app: JupyterFrontEnd<LabShell>,
+    connectionManager: ILSPDocumentConnectionManager
+  ) => {
+    app.shell.currentChanged.connect((_, args) => {
+      let newValue = args.newValue;
+
+      if (!newValue || !(newValue instanceof DocumentWidget)) {
+        console.log('No current widget');
+        return;
+      }
```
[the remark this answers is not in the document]
krassowski: If we had a proper tracker, we would not need to resort to these checks.

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+  export interface IOptions {
+    /**
+     * Promise resolving when the editor is ready.
+     */
+    editorReady(): Promise<CodeEditor.IEditor>;
+
+    /**
+     * Get the code editor to handle.
+     */
+    getEditor(): CodeEditor.IEditor | null;
```
krassowski, line 100:
```suggestion
    /**
     * Promise resolving when the editor is ready.
     */
    editorReady(): Promise<CodeEditor.IEditor>;
```

`getEditor` appears redundant in current implementation

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  constructor(options: EditorAdapter.IOptions) {
+    this._path = options.path;
+    this._editor = options.getEditor;
+    this._widgetAdapter = options.widgetAdapter;
+    this._extensions = options.extensions;
+
+    void options.editorReady().then(() => {
+      this._injectExtensions();
+    });
```
krassowski, line 33:
```suggestion
    this._widgetAdapter = options.widgetAdapter;
    this._extensions = options.extensions;

    void options.editorReady().then((editor) => {
      this._injectExtensions(editor);
    });
```

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+      if (!ext) {
+        return;
+      }
+
+      editor.injectExtension(ext.instance(editor));
+    });
+  }
+
+  private _path: string;
+  private _editor: () => CodeEditor.IEditor | null;
```
krassowski, line 79:
```suggestion
```

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * A path to a source file.
+     */
+    path: string;
+
+    /**
+     * The widget lsp adapter.
+     */
+    widgetAdapter: WidgetLSPAdapter<any>;
```
krassowski, line 110: Can the adapter change? What happens when path gets changed during file rename (note: it is a frequent bug for LSP)? Maybe we should not pass the adapter itself (nor the path) but instead retrieve these dynamically from editor model/widget?

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,145 @@
...
+  IEditorExtensionFactory
+} from '@jupyterlab/codemirror';
+
+import { IDisposable } from '@lumino/disposable';
+
+import { Signal } from '@lumino/signaling';
+import { WidgetLSPAdapter } from './adapter';
+
+/**
+ * A handler for a CodeEditor.IEditor.
```
krassowski, line 17: Docstring needs updating?

krassowski (changes requested):
Two major points:
- I think we should not add the tracking methods to `ILSPDocumentConnectionManager` to preserve separation of concerns and instead have a new tracker token which would be populated by both connection manager, and used by downstream.
- we should be careful with using file paths as source of truth; paths change and this is a major headache.

packages/lsp/src/tokens.ts
```diff
@@ -356,6 +357,11 @@ export interface IDocumentConnectionData {
  * The LSP connection state manager
  */
 export interface ILSPDocumentConnectionManager {
+  /**
+   * The path of the active widget adapter.
+   */
+  currentAdapter: string | null;
```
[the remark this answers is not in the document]
hbcarlos: I left three comments explaining it above. 

> The current adapter contains the path to the active adapter instead of the adapter itself, because when adding a new adapter, `app.shell.currentChanged` is emitted before the adapter is registered.
> *Posted [here](https://github.com/jupyterlab/jupyterlab/pull/14920#discussion_r1284318643)*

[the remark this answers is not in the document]
krassowski: Sorry I missed it. IMO this shows that the logic needs to be rewritten. For such edge cases instead of emitting when `shell.currentChanged` is fired, we should emit when adapter is first registered if it is current widget of the shell. This is I believe it should be responsibility of the tracker to emit signal when adapter actually changed, not when it "could have changed if it was already registered".

packages/lsp-extension/src/index.ts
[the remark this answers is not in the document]
hbcarlos: I'm not sure if we can get rid of that check. I added the tracker, but to know what widget is active, we need to listen for changes in the shell.

See: https://github.com/jupyterlab/jupyterlab/blob/3ca0cd659d1324d7189bbc0da0c2d52d3a5cfc68/packages/lsp/src/adapters/tracker.ts#L36-L52

Do you know a better way to track the active widget? I thought about using the [`FocusTracker`](https://github.com/jupyterlab/lumino/blob/2fe4d6316711231ff3bdb0ea3e4b03baa92f7c5d/packages/widgets/src/focustracker.ts#L24), but the adapters are not widgets. We will need to track the DocumentWidget, and the shell is already doing that.

krassowski:
Merging main to (hopefully) resolve integrity errors.

packages/lsp-extension/src/index.ts
```diff
@@ -302,12 +307,24 @@ function addRunningSessionManager(
     )
   });
 }
+
+const currentAdapter: JupyterFrontEndPlugin<IWidgetLSPAdapterTracker> = {
```
krassowski, line 311:
```suggestion
const adapterTrackerPlugin: JupyterFrontEndPlugin<IWidgetLSPAdapterTracker> = {
```

packages/lsp-extension/src/index.ts
```diff
@@ -302,12 +307,24 @@ function addRunningSessionManager(
...
 /**
  * Export the plugin as default.
  */
 export default [
   plugin,
   featurePlugin,
   settingsPlugin,
-  codeExtractorManagerPlugin
+  codeExtractorManagerPlugin,
+  currentAdapter
```
krassowski, line 329:
```suggestion
  adapterTrackerPlugin
```

buildutils/src/clean-packages.ts
```diff
@@ -13,7 +13,7 @@ exitOnUncaughtException();
 // Get all of the packages.
 const basePath = path.resolve('.');
 const baseConfig = readJSONFile(path.join(basePath, 'package.json'));
-const packageConfig = baseConfig.workspaces.packages;
+const packageConfig = baseConfig.workspaces;
```
krassowski, line 16: Can you clarify this change? I wonder if the failing CI has something to do with it.

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  export interface IFactoryOptions extends IEditorExtensionFactory.IOptions {
+    /**
+     * A path to a source file.
+     */
+    path: string;
+    /**
+     * The code editor.
+     */
+    editor: CodeEditor.IEditor;
```
krassowski, line 124: The last problem with this is that we have something called `Document.IEditor` which is what the adapter and virtual document APIs operates on (for example `transformFromEditorToRoot` and `getIndexAt` methods). `Document.IEditor` replaced `CodeEditor.IEditor` on the LSP APIs when virtual notebook feature was introduced in https://github.com/jupyterlab/jupyterlab/pull/12554.

The lack of mapping from `CodeEditor.IEditor` to `Document.IEditor` prevents the direct use of `CodeEditor.IEditor` on adapter and virtual document methods. To workaround it  I have to iterate `adapter.editors` to find which `Document.IEditor` wraps the `CodeEditor.IEditor` at hand, which is potentially inefficient for long notebooks. Btw, I use "editorAccessor" to refer to `Document.IEditor` as this gets very confusing otherwise.

Conceptually the solution is simple we can: pass `editor: Document.IEditor` instead of `editor: CodeEditor.IEditor` but this may be non-trivial to implement. Alternatively, if there is an oracle with a method `getEditorAccessorForEditor(e: CodeEditor.IEditor): Document.IEditor` we can keep it as-is.

CC @fcollonval for thoughts

buildutils/src/clean-packages.ts
hbcarlos: The attribute "packages" doesn't exist anymore, and when I tried to run `jlpm clean` there was an error. See https://github.com/jupyterlab/jupyterlab/blob/c570641c70899b83e5cbf93e8fad8bb275c4be85/package.json#L6-L7

I don't think the CI failures are related to this change.

packages/lsp/src/connection_manager.ts
```diff
@@ -51,7 +56,7 @@ export class DocumentConnectionManager
   /**
    * Map between the path of the document and its adapter
    */
-  readonly adapters: Map<string, WidgetLSPAdapter<IDocumentWidget>>;
```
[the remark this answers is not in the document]
hbcarlos: That's a good idea, otherwise, it could be confusing to have multiple ways of accessing the adapters.

packages/lsp/src/adapters/editorAdapter.ts
hbcarlos: > Conceptually the solution is simple we can: pass editor: Document.IEditor instead of editor: CodeEditor.IEditor but this may be non-trivial to implement.

I think it is fine, we can pass the `Document.IEditor` instead of the `CodeEditor.IEditor`. See 420b76db8824b64ebe6e73cdcb17560a0e6e1f85

buildutils/src/clean-packages.ts
hbcarlos: Hi @krassowski,

The integrity failures come from the `@jupyterlab/nbconvert-css` package. Some how, when building Jupyterlab from this PR, the `@jupyterlab/nbconvert-css` package generates
* jupyterlab/packages/nbconvert-css/style/e3b0c44298fc1c149afb.eot
* jupyterlab/packages/nbconvert-css/style/e3b0c44298fc1c149afb.ttf
* jupyterlab/packages/nbconvert-css/style/e3b0c44298fc1c149afb.woff
* jupyterlab/packages/nbconvert-css/File style/e3b0c44298fc1c149afb.woff2

The `@jupyterlab/nbconvert-css` package shouldn't generate those files, see the `webpack.conf.js` https://github.com/jupyterlab/jupyterlab/blob/694d3801fefc23d223fd404390d51f6403a5775b/packages/nbconvert-css/webpack.config.js#L31-L51

I reverted the change made to the script clean packages, but it is failing yet.

Do you know what could be the cause?

trungleduc: `null-loader` is deprecated in Webpack 5, don't know why it works up to now.

krassowski: I've tired a few things but could not narrow it down. Really puzzling :question: .

krassowski (commented):
Thanks for the updates!

packages/lsp/src/adapters/tracker.ts
```diff
@@ -0,0 +1,223 @@
...
+ */
+export namespace WidgetLSPAdapterTracker {
+  /**
+   * The instantiation options for the WidgetLSPAdapterTracker.
+   */
+  export interface IOptions {
+    /**
+     * The JupyterLab shell for tracking all widgets.
+     */
+    shell: LabShell;
```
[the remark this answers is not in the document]
hbcarlos: Good catch! Thanks, Michał.

Let's do it on another PR since #14782 is targeting v5.0. I'll add a TODO and document it on this PR.

buildutils/src/clean-packages.ts
fcollonval: Ok I figured out what is happening but does not yet have a fix.

So within this PR, `@jupyterlab/application` was added as dependency of `@jupyterlab/lsp`. This implies that the associated styles are added too and in particular the font awesome styles 

https://github.com/jupyterlab/jupyterlab/blob/d58fcbc3a45677edf427fdd2878e597918655314/packages/application/style/index.js#L7-L8

By dependency propagation (through `@jupyterlab/notebook`), `@jupyterlab/application` are now included when building nbconvert-css. But we were explicitly not importing all application styling:

https://github.com/jupyterlab/jupyterlab/blob/d58fcbc3a45677edf427fdd2878e597918655314/packages/nbconvert-css/raw.js#L6

So due to the new dependency cycle, `@jupyterlab/nbconvert-css` is including fonts that were not included previously. Hence we were never using the rules with `null-loader`.

---

Now I'll look for a fix

fcollonval: In fact (and partly related to the discussion below about https://github.com/jupyterlab/jupyterlab/issues/14782), I guess it makes sense that dependency imported only for their types should not implies that the associated styles should be part of the package. However this will be very complex to implement in the integrity test. So I propose to do it manually for this case (meaning we may be hit again by it at any point in the future).

fcollonval: I can confirm that removing `@jupyterlab/application` from `@jupyterlab/lsp` resolves the issue.

cc @krassowski @hbcarlos what do you think?

krassowski: Maybe we do not need `@jupyterlab/application` here. The tracker could could just accept a shell-like object with an interface including `currentChanged` and `activeWidget`?

krassowski: Or could rely on trackers of individual document widgets passed during registration.

hbcarlos: > So I propose to do it manually for this case (meaning we may be hit again by it at any point in the future).

What do you mean by manually?  Adding `@jupyterlab/application` to the skip CSS here https://github.com/jupyterlab/jupyterlab/blob/97d320a850eaa5729d8a1d706799033ef19f31ac/buildutils/src/ensure-repo.ts#L192

so the integrity script doesn't try to include it?

fcollonval: > so the integrity script doesn't try to include it?

Yes

---

> The tracker could just accept a shell-like object with an interface including currentChanged and activeWidget?

This would work too if it is defined locally within `@jupyterlab/lsp` as we cannot rely on `JupyterFrontEnd` as it is part of `@jupyterlab/application`.

hbcarlos: Sorry @krassowski, I didn't see your comments earlier.

> Maybe we do not need @jupyterlab/application here. The tracker could could just accept a shell-like object with an interface including currentChanged and activeWidget?

Sounds good. I implemented this solution in my last commit.

packages/lsp/src/adapters/tracker.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import { WidgetLSPAdapter } from './adapter';
+import { IShell, IWidgetLSPAdapterTracker } from '../tokens';
+
+/**
+ * A class that keeps track of widget adapter instances.
+ *
+ * @typeparam T - The type of widget being tracked. Defaults to `WidgetLSPAdapter`.
+ */
+export class WidgetLSPAdapterTracker<
```
krassowski, line 16: Unit tests for this class would be nice (and should be easy to write).

packages/lsp/src/adapters/editorAdapter.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import { Signal } from '@lumino/signaling';
+
+import { WidgetLSPAdapter } from './adapter';
+import { Document } from '../tokens';
+
+/**
+ * The CodeEditor.IEditor adapter.
+ */
+export class EditorAdapter implements IDisposable {
```
krassowski, line 19: A single test which invokes this logic would be nice.

krassowski (approved):
Thank you @hbcarlos, I think this is good, we are only missing tests. Ideally we would document this for extension authors, or add an example in `/examples` (which could happen in a follow-up PR).

hbcarlos:
Thanks for the review, @krassowski!

I added tests and documentation.

docs/source/extension/extension_points.rst
```diff
@@ -951,10 +951,70 @@ Here is an example for making requests to the language server.
...
+                textDocument: { uri: widgetAdapter.virtualDocument!.uri }
+              });
+
+              res.then(value => {
+                console.debug(value);
+              }).catch(e => console.error);
+          }});
+
+          // Create a conditional CodeMirror extension.
+          return EditorExtensionRegistry.createConditionalExtension(ext);
```
krassowski, line 1010: I guess `createConfigurableExtension` will be often useful, but for sake of having simple example I think it is better to use immutable vs conditional?

```suggestion
          // Wrap the CodeMirror extension in the extension registry object.
          return EditorExtensionRegistry.createImmutableExtension(ext);
```

packages/lsp-extension/src/index.ts
```diff
@@ -302,12 +307,24 @@ function addRunningSessionManager(
     )
   });
 }
+
+const adapterTrackerPlugin: JupyterFrontEndPlugin<IWidgetLSPAdapterTracker> = {
+  id: '@jupyterlab/lsp-extension:tracker',
+  description: 'Provides the `WidgetLSPAdapter`s tracker.',
+  autoStart: true,
+  provides: IWidgetLSPAdapterTracker,
+  activate: (app: JupyterFrontEnd<LabShell>): IWidgetLSPAdapterTracker => {
```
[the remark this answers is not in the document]
jtpio: This seems related to the use of the `currentChanged` signal here:

https://github.com/jupyterlab/jupyterlab/blob/0b33a5ca9318d909233914adb0cbd90d3bc7a472/packages/lsp/src/adapters/tracker.ts#L28-L45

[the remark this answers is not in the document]
jtpio: Ideally the tracker plugin could also work with shells that do not implement a `currentChanged` signal, or for which the signature is different.

In this case the plugin could check `currentWidget`, which is part of the `JupyterFrontEnd.IShell` interface.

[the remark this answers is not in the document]
krassowski: Yes, I brought it up earlier in https://github.com/jupyterlab/jupyterlab/pull/14920#discussion_r1316507818.

The problem with `IShell` is that it indeed is missing a few things which are really handy (there is a reason why these were added to `LabShell`)
- https://github.com/jupyterlab/jupyterlab/issues/14782
- https://github.com/jupyterlab/jupyterlab/issues/14883

Is the Notebook compatibility problem related to one of the above, or a separate issue due to divergence of shells? Would it be a good idea to make the Notebook shell conform closer it Lab shell even if sometimes the signals/attributes would be no-op for Notebook?

packages/lsp/src/adapters/tracker.ts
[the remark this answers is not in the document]
krassowski: This indeed caused problems for Notebook. See https://github.com/jupyterlab/jupyterlab/pull/14920#discussion_r1350240731.

packages/lsp-extension/src/index.ts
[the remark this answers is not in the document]
jtpio: > Would it be a good idea to make the Notebook shell conform closer it Lab shell even if sometimes the signals/attributes would be no-op for Notebook?

Yes I think we can do that. Currently the signature is as follows: https://github.com/jupyter/notebook/blob/c3ec39e8d7428aac2cda6b0a9fed17643eab5210/packages/application/src/shell.ts#L166

```ts
get currentChanged(): ISignal<NotebookShell, void>
```

While in JupyterLab:

https://github.com/jupyterlab/jupyterlab/blob/0b33a5ca9318d909233914adb0cbd90d3bc7a472/packages/application/src/shell.ts#L545

Which is why it is causing an issue in Notebook 7 since there is no `newValue`. But I guess `INotebookShell` could align its signature with the one from JupyterLab to reduce friction for that particular case. But this wouldn't prevent the LSP plugin to use other `ILabShell` specific APIs if this is what it expects.

Issue #14919, opened 2023-08-04
HTML Viewer can not open external links


yuvipanda:
## Description

When a `.html` file is viewed inside JupyterLab, the iframing seems to set things up in such a way that external links never work, regardless of trust status.

## Reproduce

1. Create a simple html file with the following contents:
   ```html
   <body>
      <a href="https://github.com" target="_blank">GitHub</a>
   </body>
   ```
2. Open this in the HTML viewer, by selecting the file, right clicking and opening in HTML Viewer inside JupyterLab.
3. Try to click the link. Nothing happens
4. Click the "Trust HTML" button. Try to click the link. Nothing happens. I expect it to open in a new tab, or at least open somewhere.

## Context

This is using the latest version of jupyterlab, 4.0.4 at this time.

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/Users/yuvipanda/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/opt/homebrew/opt/openjdk@17/bin
	/Users/yuvipanda/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/opt/homebrew/opt/openjdk@17/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/usr/local/bin
	/usr/bin
	/bin
	/usr/sbin
	/sbin
	/Users/yuvipanda/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/opt/homebrew/opt/openjdk@17/bin
	/Users/yuvipanda/micromamba/condabin
	/Users/yuvipanda/.virtualenvs/infrastructure/bin
	/Applications/Emacs.app/Contents/MacOS/bin-arm64-11
	/Applications/Emacs.app/Contents/MacOS/libexec-arm64-11
	/Users/yuvipanda/code/dotfiles/bin
	/Users/yuvipanda/code/dotfiles/bin
	/Users/yuvipanda/code/dotfiles/bin

sys.path:
	/Users/yuvipanda/.local/share/virtualenvs/infrastructure/bin
	/opt/homebrew/Cellar/python@3.10/3.10.12/Frameworks/Python.framework/Versions/3.10/lib/python310.zip
	/opt/homebrew/Cellar/python@3.10/3.10.12/Frameworks/Python.framework/Versions/3.10/lib/python3.10
[...]

PR #14933, opened 2023-08-08
Allow opening links in new tabs when HTML is trusted

yuvipanda:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14919

## Code changes

When user opts to trust a HTML file in the HTML viewer, allow code there to open
pages in a new tab by adding `allow-popups` to the IFrame's sandbox attribute.

## User-facing changes

Once a user trusts a HTML file, the file can open new windows either via JS or
via `target="_blank"` links.

### DIFF
--- a/packages/htmlviewer/src/widget.tsx
+++ b/packages/htmlviewer/src/widget.tsx
@@ -272,7 +272,7 @@ namespace Private {
   /**
    * Sandbox exceptions for trusted HTML.
    */
-  export const trusted: IFrame.SandboxExceptions[] = ['allow-scripts'];
+  export const trusted: IFrame.SandboxExceptions[] = ['allow-scripts', 'allow-popups'];
 
   /**
    * Namespace for TrustedButton.
--- a/packages/htmlviewer/src/widget.tsx
+++ b/packages/htmlviewer/src/widget.tsx
@@ -36,12 +36,18 @@ const CSS_CLASS = 'jp-HTMLViewer';
 const UNTRUSTED_LINK_STYLE = (options: { warning: string }) => `<style>
 a[target="_blank"],
 area[target="_blank"],
-form[target="_blank"] {
+form[target="_blank"],
+button[formtarget="_blank"],
+input[formtarget="_blank"][type="image"],
+input[formtarget="_blank"][type="submit"] {
   cursor: not-allowed !important;
 }
 a[target="_blank"]:hover::after,
 area[target="_blank"]:hover::after,
-form[target="_blank"]:hover::after {
+form[target="_blank"]:hover::after,
+button[formtarget="_blank"]:hover::after,
+input[formtarget="_blank"][type="image"]:hover::after,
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks @yuvipanda 

I expanded following the discussion at #14919 to dynamically styled the links in the frame when it is untrusted. Let me know what do you think?

add_link_styling


cc @krassowski

fcollonval:
Kicking CI following bot commit

packages/htmlviewer/src/widget.tsx
```diff
@@ -33,6 +33,22 @@ const RENDER_TIMEOUT = 1000;
  */
 const CSS_CLASS = 'jp-HTMLViewer';
 
+const UNTRUSTED_LINK_STYLE = (options: { warning: string }) => `<style>
+a {
+  cursor: not-allowed !important;
+}
+a:hover::after {
```
krassowski, line 40: Normal links work and do not need this. Only popup links are restricted.

```suggestion
a[target="_blank"] {
  cursor: not-allowed !important;
}
a[target="_blank"]:hover::after {
```

krassowski (commented):
Thanks @yuvipanda and thank you @fcollonval - this solution is great!

packages/htmlviewer/src/widget.tsx
fcollonval: Thanks @krassowski
I was looking at the [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) and apparently 4 tags support `target`: `a`, `area`, `base` and `form`. I guess we should cover `a[target="_blank"]`, `area[target="_blank"]` and `form[target="_blank"]` (as the one for `base` is anyway overwritten) as the documentation for `allow-popups` does not specify any tag - only `target="blank_"`?

krassowski: Right, thank you! I also see `formtarget`:

> This value can be overridden by a [formtarget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#formtarget) attribute on a [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), [`<input type="submit">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit), or [`<input type="image">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image) element.

fcollonval:
The latest commit 
- add the case of `formtarget`
- rephrase the warning to _Action disabled as the file is untrusted._ as button can become disabled; good English speaker feel free to suggest a better wording.

krassowski (approved):
Minor wording suggestions on tooltip.

krassowski:
I wonder if "not trusted" would be more consistent. Also, we could set a non-serif font on the banner but it's all minor details.

fcollonval:
CI failure is not related - thanks @krassowski for the review

fcollonval:

Issue #14924, opened 2023-08-07
Wrong yarn.lock in released 4.0.4


fcollonval:
## Description

Since the release of 4.0.4, integration tests are failing; see for examples: 
- https://github.com/jupyterlab/extension-template/actions/runs/5777728638/job/15658220778?pr=39
- https://github.com/jupyterlab-contrib/jupyter-archive/actions/runs/5777706952/job/15658204740?pr=118

The reason is that the galata xpath selector to check for the _Launcher_ tab is not valid any longer as a consequence of https://github.com/jupyterlab/lumino/pull/606 (the selector was using `jp-Activity` class that is now gone).

---

The underlying painful bug is that the yarn.lock got modify to pull the latest version of @lumino instead of keeping the one in the dev yarn.lock. This is the reason the CI is green and we did not capture that bug :cry:

fcollonval:
The upgrade in the staging yarn.lock occurred in the publication commit: https://github.com/jupyterlab/jupyterlab/commit/de28aca9ca8a60b25d28efdf359bb2c69d58c634

We have now quite some discrepancies with the dev yarn.lock.

fcollonval:
Similar to https://github.com/jupyterlab/jupyterlab/issues/11429 but it is gonna be more tricky to solve due to the new yarn version algorithm.

jtpio:
> The upgrade in the staging yarn.lock occurred in the publication commit: [de28aca](https://github.com/jupyterlab/jupyterlab/commit/de28aca9ca8a60b25d28efdf359bb2c69d58c634)

Isn't is the expected behavior that `staging/yarn.lock` should be replaced by `yarn.lock` on release?

If we look at the release commit for `4.0.3`, `staging/yarn.lock` was also updated with newer versions of packages: https://github.com/jupyterlab/jupyterlab/commit/f8226e7e4efd8fadebc1eee7b003fad33615fa5c#diff-285db1844b4ca4d4bd926aba6bd1b9e430857f11840b33a5fed086354d0978c8

fcollonval:
Yes it is expected to be replaced by the one in dev-mode. But since we change the application top package and pin the `@jupyterlab/*` packages (to the one we have just released), we are executing `jlpm install` (and deduplicating the lock file), after copying the file. That call is the culprit as now `jlpm install` will bump any packages of the yarn.lock as long as their version spec is respected. This breaks the control we should have on the released yarn.lock as basically any uncontrolled bump of dependencies may occur.

[...]

PR #14945, opened 2023-08-10
Bump lumino packages

fcollonval:
## References

Follow-up of #14924 to avoid publishing a release with downgraded package

## Code changes

Bump lumino package on dev
Fix galata tab selector (starts deprecating xpath selectors usage as they are not a good practice)

## User-facing changes

None

> Only dev using galata will have their integration tests corrected

None


fcollonval:

PR #14953, opened 2023-08-11
Check backward dep

fcollonval:
## References

Follow-up of #14395 - when executing `jlpm up "@lumino/*"` the backward compatible versions are dropped. As we were not testing if those versions were defined, we could have released a version of `@jupyterlab/rendermime-interfaces` breaking again extension for Lab 3.

## Code changes

Add check for backward deps in integrity check

## User-facing changes

None

None


fcollonval:

PR #14973, opened 2023-08-15
added title to Close tab icon

### LINKED ISSUE #14961: Close Launcher Tab Icon not being read with screen reader
### Problem

We discovered that the close tab icon was not being read out by the screen reader when using it to navigate through Jupyter.

### Proposed Solution

We advise making a quick change, such as giving the close tab icon a "title," so it can be read by the screen reader, in order to make the launcher tab accessible to screen readers.

### Additional context

None

j264415:
## References

Fixes #14961 - Close Launcher Tab Icon not being read with screen reader

## Code changes

The **tabbarsvg.ts** file was modified to include **"title"** by setting it as a **const** so that when a launcher or notebook tab has to be closed, the user can tab to the icon and be notified by the screen reader when they have reached the close tab icon.

## User-facing changes

Currently, the screen reader does not pick up or read the close tab icon when navigating the Launcher section of Jupyter. By adding a title to this element, users who use screen readers will be able to understand what this icon is used for when appropriate.

None

---

[fcollonval]: Edited to link the issue with the PR - FYI @j264415 you can use [special keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to automatically linked a PR with an issue.


packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -46,6 +46,7 @@ export namespace TabBarSvg {
      * @returns A virtual element representing the tab close icon.
      */
     renderCloseIcon(data: TabBar.IRenderData<any>): VirtualElement {
+      const title = 'Close Launcher';
```
krassowski, line 49: This needs translation https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html

fcollonval: @j264415 there is an additional problem with your proposal is that the close icon is generated for all tabs - not only the launcher one. So `Close Launcher` is not appropriate.

You should be able to get the widget name using `data.title.label`. And taking into account the comment from Mike, you can achieve that by applying:

```suggestion
      const trans = (TabBarSvg.translator ?? nullTranslator).load('jupyterlab');
      const title = data.title.label ? trans.__('Close %1', data.title.label) : trans.__('Close tab');
```

j264415: @fcollonval thank you for the comment. I have made the changes now that you have suggested.

fcollonval (approved):
Thanks @j264415

fcollonval:

PR #14979, opened 2023-08-16
Fix translation syntax to extract it by the parser

fcollonval:
## References

Follow-up of #14933

## Code changes

Fix translation syntax to be extracted by the parser

## User-facing changes

The new warning on untrusted html file will be translated

None


fcollonval:

PR #14987, opened 2023-08-17
adding labels to user input 

s596757:
Text Issue #6

In order to provide meaningful labels for user input, as per https://github.com/jupyterlab/jupyterlab/issues/9399,

I have done the following: 

- Search fields have been changed to specify what they are searching e.g. shortcuts.
- added type=text to find/replace
- added aria-labels to user inputs
- added aria labels and type to Line Column Form
- Labelled top nav bar and updated placeholder text

-

### DIFF
--- a/packages/codeeditor/src/lineCol.tsx
+++ b/packages/codeeditor/src/lineCol.tsx
@@ -107,7 +107,7 @@ class LineFormComponent extends React.Component<
             <input
               type="text"
               className="jp-lineFormInput"
-              aria-labelledby='Line Column form'
+              aria-label={this._trans('Line Column form')}
               onChange={this._handleChange}
               onFocus={this._handleFocus}
               onBlur={this._handleBlur}
--- a/packages/codeeditor/src/lineCol.tsx
+++ b/packages/codeeditor/src/lineCol.tsx
@@ -107,6 +107,7 @@ class LineFormComponent extends React.Component<
             <input
               type="text"
               className="jp-lineFormInput"
+              aria-labelledby='Line Column form'
               onChange={this._handleChange}
               onFocus={this._handleFocus}
               onBlur={this._handleBlur}
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -360,7 +360,7 @@ class Header extends ReactWidget {
           )}
         </div>
         <FilterBox
-          placeholder={this.trans.__('Search')}
+          placeholder={this.trans.__('Search Extension Manager')}
           disabled={!this.model.isDisclaimed}
           updateFilter={(fn, query) => {
             this.model.query = query ?? '';
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -360,7 +360,7 @@ class Header extends ReactWidget {
           )}
         </div>
         <FilterBox
-          placeholder={this.trans.__('Search Extensions')}
+          placeholder={this.trans.__('Search extensions')}
           disabled={!this.model.isDisclaimed}
           updateFilter={(fn, query) => {
             this.model.query = query ?? '';
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -360,7 +360,7 @@ class Header extends ReactWidget {
           )}
         </div>
         <FilterBox
-          placeholder={this.trans.__('Search Extension Manager')}
+          placeholder={this.trans.__('Search Extensions')}
           disabled={!this.model.isDisclaimed}
           updateFilter={(fn, query) => {
             this.model.query = query ?? '';
--- a/packages/json-extension/src/component.tsx
+++ b/packages/json-extension/src/component.tsx
@@ -75,7 +75,7 @@ export class Component extends React.Component<IProps, IState> {
         <InputGroup
           className="filter"
           type="text"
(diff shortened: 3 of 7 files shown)


packages/settingeditor/src/pluginlist.tsx
```diff
@@ -451,7 +451,7 @@ export class PluginList extends ReactWidget {
         <FilterBox
           updateFilter={this.setFilter}
           useFuzzyFilter={false}
-          placeholder={trans.__('Search…')}
+          placeholder={trans.__('Search Plugins List…')}
```
fcollonval, line 454: As the feature looks at all settings (see screenshot below), i would use something like:

```suggestion
          placeholder={trans.__('Search Settings…')}
```

image

packages/shortcuts-extension/src/components/TopNav.tsx
```diff
@@ -162,9 +162,10 @@ export class TopNav extends React.Component<ITopNavProps> {
           <Symbols />
           <InputGroup
             className="jp-Shortcuts-Search"
+            aria-labelledby='Search Shortcuts'
```
fcollonval, line 165: Same here I guess

```suggestion
            aria-label={trans.__('Search Shortcuts')}
```

fcollonval (changes requested):
Thanks @s596757 

I made a couple of suggestions to improve this.

s596757:
> Thanks @s596757
> 
> I made a couple of suggestions to improve this.

Changes made, please review again. Thank you :)

krassowski (changes requested):

packages/codeeditor/src/lineCol.tsx
```diff
@@ -107,6 +107,7 @@ class LineFormComponent extends React.Component<
             <input
               type="text"
               className="jp-lineFormInput"
+              aria-label={this._trans('Line Column form')}
```
gabalafou, line 110: I think it would be better to give this text input an id and connect it to the label below, e.g.:

```
const textInputId = DOMUtils.createDomID() + '-line-number-input';
<input id={textInputId} ... />
...
<label for={textInputId}
```

packages/extensionmanager/src/widget.tsx
```diff
@@ -360,7 +360,7 @@ class Header extends ReactWidget {
           )}
         </div>
         <FilterBox
-          placeholder={this.trans.__('Search')}
+          placeholder={this.trans.__('Search Extensions')}
```
gabalafou, line 363:
```suggestion
          placeholder={this.trans.__('Search extensions')}
```

packages/json-extension/src/component.tsx
```diff
@@ -75,7 +75,7 @@ export class Component extends React.Component<IProps, IState> {
         <InputGroup
           className="filter"
           type="text"
-          placeholder={trans.__('Filter…')}
+          placeholder={trans.__('Find')}
```
gabalafou, line 78: Yeah I agree. This is more like "Find" (Ctrl + F) functionality than filter.

packages/lsp-extension/src/renderer.tsx
```diff
@@ -218,6 +218,7 @@ function BuildSettingForm(props: ISettingFormProps): JSX.Element {
                   <div className="jp-inputFieldWrapper jp-FormGroup-contentItem">
                     <input
                       className="form-control"
+                      aria-label={props.trans.__('Server name')}
```
gabalafou, line 221: Here too, I think it would be better to connect this input with the "Server name" above. Will probably need to change the `<h3>` to a `<label>`

packages/settingeditor/src/pluginlist.tsx
```diff
@@ -451,7 +451,7 @@ export class PluginList extends ReactWidget {
         <FilterBox
           updateFilter={this.setFilter}
           useFuzzyFilter={false}
-          placeholder={trans.__('Search…')}
+          placeholder={trans.__('Search Settings…')}
```
gabalafou, line 454:
```suggestion
          placeholder={trans.__('Search settings…')}
```

packages/shortcuts-extension/src/components/TopNav.tsx
```diff
@@ -162,6 +162,7 @@ export class TopNav extends React.Component<ITopNavProps> {
           <Symbols />
           <InputGroup
             className="jp-Shortcuts-Search"
+            aria-label={trans.__('Search Shortcuts')}
```
gabalafou, line 165:
```suggestion
            aria-label={trans.__('Search shortcuts')}
```

Is there a reason you're adding aria-label here, but modifying the placeholder value for "search settings"? Seems inconsistent.

gabalafou (changes requested):
These are good improvements, but they need to be implemented slightly differently in a few places, I think, to avoid duplication of code, as well as translation strings.

Also please update the PR description so that it links to the appropriate issue. I think there's a typo currently.

packages/codeeditor/src/lineCol.tsx
krassowski: This would be `htmlFor={textInputId}` because React.

packages/shortcuts-extension/src/components/TopNav.tsx
krassowski: That would probably be because of my earlier suggestion: https://github.com/jupyterlab/jupyterlab/pull/14987#discussion_r1298422593

krassowski:
Closing in favour of #15222

PR #14989, opened 2023-08-17
Fix more links

### LINKED ISSUE #14955: Make links to documentation under help menu use configuration vars rather than hard coding
### Problem
Under the help menu, the links for Jupyterlab FAQ for instance, are all hardcoded in the compiled Javascript objects.  when running Jupyter in an isolated environment, changing those URLs to point to locally hosted documentation involves lots of "sed" commands against the jupyter/lab/static directory, and the urls keep changing meaning the regexps keep having to be updated.  (Is it /stable/ is it /latest/, etc...) If those URLs were instead in some kind of configuration file, even if as an option, that would make life much easier.

### Proposed Solution

I am not sure where this would fit best.  Could be a configuration setting you can change in the settings editor since you can apply overrides for those, could be in some hidden json file.  Anything to make life easier going forward.  Ideally the same solution would be leveraged by Notebookv7 but figured lab is where to get this discussion started.

fcollonval:
## References

- Fix yarn documentation broken link (https://github.com/jupyterlab/jupyterlab/actions/runs/5881217766/job/15949238800)
- Fixes #14955

## Code changes

Fix links

## User-facing changes

Links work in doc

None

### DIFF
--- a/.github/workflows/linuxtests.yml
+++ b/.github/workflows/linuxtests.yml
@@ -13,7 +13,7 @@ jobs:
     name: Linux
     strategy:
       matrix:
-        group: [integrity, integrity2, integrity3, release_test, docs, usage, usage2, splice_source, python, examples, interop, nonode, linkcheck, lint]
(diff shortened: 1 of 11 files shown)


fcollonval:

Issue #14988, opened 2023-08-17
Remove timeout during settings transformation


brichet:
When Jupyterlab starts up, nothing seems to happen for a second.

![image](https://github.com/jupyterlab/jupyterlab/assets/32258950/466d302b-bff9-43a8-b57f-7ac6d2917b7f)

This delay is due to plugin transform (thanks to @fcollonval for discovering this), which contains a 1s timeout waiting for an available transform function.

https://github.com/jupyterlab/jupyterlab/blob/1441c45888b91840a02a489557743e866329477b/packages/settingregistry/src/settingregistry.ts#L685-L693

with a `timeout` value set to 1000ms:

https://github.com/jupyterlab/jupyterlab/blob/1441c45888b91840a02a489557743e866329477b/packages/settingregistry/src/settingregistry.ts#L46

### Problem

This delay occurs during the preloading of the plugins:

https://github.com/jupyterlab/jupyterlab/blob/1441c45888b91840a02a489557743e866329477b/packages/settingregistry/src/settingregistry.ts#L606

At this stage, none of the transform functions are registered, and it seems that all plugins are waiting for the `settingsRegistry` to be "preloaded" before registering the transform functions.
The consequence is that the deadline is always reached for every plugin containing a transform function, at least for the core plugins.

### Proposed Solution

I think we can do away the timeout concept (and the associated loop). In practice, it is never used except during preloading (where the timeout is reached).

- if the transform function is already registered during preloading, this function will be played as expected.
- if the transform function is not registered during preload, the plugin will remain in the `_unloadedPlugins` map, and the transform function will be played the next time `load()` is called, after registration (see <a href=https://github.com/jupyterlab/jupyterlab/blob/1441c45888b91840a02a489557743e866329477b/packages/settingregistry/src/settingregistry.ts#L394-L403>settingregistry.ts#L394-L403</a>).

### Additional context

Here is the performance chart after applying the proposed solution (no more gap), with no consequence as far as I know:

![image](https://github.com/jupyterlab/jupyterlab/assets/32258950/6d75ce38-3754-4ec0-b144-b3e0f0df7cb7)

cc. @fcollonval, @afshin, @mlucool

fcollonval:
Thanks @brichet for investigating - yes dropping it sounds reasonable.

krassowski:
[...]

PR #14990, opened 2023-08-17
Remove the loop to discover transform function if not registered

brichet:
This PR should improve the pageload time by removing a loop waiting for the transform functions to be registered.

## References

See https://github.com/jupyterlab/jupyterlab/issues/14988 for context.
Fixes https://github.com/jupyterlab/jupyterlab/issues/14988.

## User-facing changes

None

None


jtpio:
Does the git history say in which PR this delay was originally added? Just to double check whether there was a reason to it.

packages/settingregistry/src/settingregistry.ts
```diff
@@ -606,7 +598,7 @@ export class SettingRegistry implements ISettingRegistry {
           await this._load(await this._transform('fetch', plugin));
         } catch (errors) {
           /* Ignore preload timeout errors silently. */
```
krassowski, line 600:
```suggestion
          /* Ignore  silently if no transformers. */
```

krassowski (commented):
I tested this against jupyterlab-lsp which takes advantage of heavy transformations and it works fine.

Do you have any a timing comparison?

brichet:
> I tested this against jupyterlab-lsp which takes advantage of heavy transformations and it works fine.

Nice, thanks for testing

> Do you have any a timing comparison?

There is a comparison in the referenced issue https://github.com/jupyterlab/jupyterlab/issues/14988

krassowski (approved):
Thanks!

packages/settingregistry/src/settingregistry.ts
krassowski: Let's just update this comment before merging.

packages/settingregistry/src/settingregistry.ts
```diff
@@ -678,28 +667,17 @@ export class SettingRegistry implements ISettingRegistry {
...
       return transformed;
+    } else {
+      throw [
+        {
+          instancePath: '',
+          keyword: 'unset',
+          message: `${plugin.id} has no transformers yet.`,
+          schemaPath: ''
+        } as ISchemaValidator.IError
+      ];
```
afshin, line 679:
```suggestion
    }
    // If the plugin has no transformers, throw an error and bail.
    throw [
      {
        instancePath: '',
        keyword: 'unset',
        message: `${plugin.id} has no transformers yet.`,
        schemaPath: ''
      } as ISchemaValidator.IError
    ];
```

afshin (approved):
Cheers!

afshin:
I initially added the delay in order to retry failed attempts. I can't remember now if this is over engineering a solution to a problem we didn't have or a race condition. But this change feels pretty safe right now and it'll be transparent to modify if we need something more nuanced. Thanks for working on it @brichet and thank you for testing it @krassowski

brichet:
@afshin I don't why but it seems the CI applied again the licence header...

afshin:
I modified the `.yaml` file that was failing to ignore `.txt` files and I think because it had been failing the previous times, there were many files it ignored?

brichet:
> I modified the `.yaml` file that was failing to ignore `.txt` files and I think because it had been failing the previous times, there were many files it ignored?

OK, we should probably revert https://github.com/jupyterlab/jupyterlab/pull/14990/commits/749edd6cdce5b499363bedbac6530349908b361f

afshin:
Why should we revert it? Are those files not supposed to have a license?

When I re-based against `upstream/main` those files didn't have headers ... which makes me think maybe that job has been broken for a little while.

brichet:
> Why should we revert it? Are those files not supposed to have a license?
> 
> When I re-based against `upstream/main` those files didn't have headers ... which makes me think maybe that job has been broken for a little while.

Some of them at least already had a licence header, and now it breaks because of shebang not on the top of some file:
https://github.com/jupyterlab/jupyterlab/pull/14990/commits/749edd6cdce5b499363bedbac6530349908b361f#diff-7e1ac2cc0f728c59bfe1b92a6b9b9c4ef5fd6979e552c480cdf8a8a7dfb44317R6

afshin:
Ugh, you're right. Terrible.

krassowski:
There is an issue: https://github.com/jupyterlab/jupyterlab/issues/15067 and a PR attempting to fix it https://github.com/jupyterlab/jupyterlab/pull/15068. This was caused by a merge of https://github.com/jupyterlab/jupyterlab/pull/15054 which annoyingly passed as green (because the action was run from main branch, not from PR so the status check was misleading). Please help with the PR

Issue #13361, opened 2022-11-01
Race condition in kernel restart and cell execution


divyansshhh:
## Description
If kernel is restarted and cell execution commands are given quickly, then some cells tend to be skipped. There appears to be a race condition here.

## Reproduce
This is flaky to reproduce but restarting the kernel and running cells quickly should work. 

## Expected behavior
Cells shouldn't be skipped

andrii-i:
Issue #5554 and #8477 addressed similar issue previously and were solved but it seems that issue is happening again

krassowski:
@divyansshhh would you be able to provide a reproducible example, or describe the circumstances in more detail? I do not see a problem by running cells quickly in sequence after restart. I tested this manually and with an automated test comparing 4 ways to restart and re-run all cells in a notebook, 100 times in a loop and comparing results (a stripped down version was committed in #14607; see details below) but there was no irregularity I could detect with the example notebook used in tests.

<details>

Note: this test runs 30 minutes so it is not something we would want to include on CI with 100 repeats (but 1 repeat is ok).

```ts
test('Restart kernel and execute cells', async ({ page, tmpPath }) => {
    test.setTimeout(60000 * 100);
    await page.notebook.openByPath(`${tmpPath}/${fileName}`);
    await page.notebook.activate(fileName);

    const acceptDialog = async () => {
      const dialogSelector = '.jp-Dialog-content';
      await page.waitForSelector(dialogSelector);
      // Accept option to trust the notebook
      await page.click('.jp-Dialog-button.jp-mod-accept');
    };

    const nbPanel = await page.notebook.getNotebookInPanel();
    const firstCell = await page.notebook.getCell(0);
    // Ignore the cell toolbar as currently it is flaky on markdown cells
    await page.addStyleTag({content: '.jp-cell-toolbar{display: none}'});

    for (let repeat = 0; repeat < 100; repeat++) {
      // 1. Restart and run all using a single command
      await page.menu.clickMenuItem('Kernel>Restart Kernel and Run All Cells…');
      await acceptDialog();
      await page.notebook.waitForRun();
      // Click on first cell to avoid random hover effects due to mouse movement
      await firstCell.click();
      expect(await nbPanel.screenshot()).toMatchSnapshot('restart-and-run.png');

      // 2. Restart manually and run all cells at once
      await page.menu.clickMenuItem('Kernel>Restart Kernel…');
      await acceptDialog();
      await page.notebook.run();
      await firstCell.click();
[...]

PR #14996, opened 2023-08-21
Define cells to run as independent of selection

krassowski:
## References

Addresses #13361

Instead of relying on selection state for running cells (which is how individual run commands are implemented), save the cells to run before asynchronous actions to ensure that selection changes do not influence which cells get run.

## Code changes

- adds `Private.runSelected()` function
- adds public `NotebookActions.runCells()` function
- rewrites `notebook:restart-and-run-to-selected` and `notebook:restart-run-all` commands to use `NotebookActions.runCells`

## User-facing changes

When restarting and re-running up to selected cell, the cell which was selected during invoking the command (rather than the cell selected after restart) will be the one executed in the end.

None I think?


packages/notebook/src/actions.tsx
```diff
@@ -803,13 +838,9 @@ export namespace NotebookActions {
...
 
-    notebook.deselectAll();
-    for (let i = notebook.activeCellIndex; i < notebook.widgets.length; ++i) {
-      notebook.select(notebook.widgets[i]);
-    }
-
-    const promise = Private.runSelected(
+    const promise = Private.runCells(
       notebook,
+      notebook.widgets.slice(notebook.activeCellIndex, undefined),
```
fcollonval, line 843: Can be omitted

```suggestion
      notebook.widgets.slice(notebook.activeCellIndex),
```

packages/notebook/src/actions.tsx
```diff
@@ -803,13 +838,9 @@ export namespace NotebookActions {
 
     const state = Private.getState(notebook);
 
-    notebook.deselectAll();
-    for (let i = notebook.activeCellIndex; i < notebook.widgets.length; ++i) {
-      notebook.select(notebook.widgets[i]);
-    }
-
-    const promise = Private.runSelected(
+    const promise = Private.runCells(
```
fcollonval, line 841: I have a change in behavior for this one. It does not activate the last cell after it is done (it keeps the selected cell activated).

fcollonval (changes requested):
Thanks @krassowski 

The code looks good. However I spotted one regression. When calling run selected cell and all below, the last cell is not activated (the current select cell stays the one active)

packages/notebook/src/actions.tsx
krassowski: Good catch, thank you! Addressed and tested in 202e958e967955d7a3d94a1d666dd7661a77a644.

fcollonval (approved):
Thanks @krassowski

Issue #14637, opened 2023-06-02
Settings get unselected when I select another one


krassowski:
## Description

In CodeMirror settings checking all:
- Code Folding
- Highlight the active line
- Highlight trailing white spaces
- Highlight white spaces
is not possible as one get unselectedd when selecting another. Also switching tabs unselect somes of the checked positions.

## Reproduce

![what-is-going-on](https://github.com/jupyterlab/jupyterlab/assets/5832902/0a2f593c-4145-4d94-babc-0b1dd9ed5395)

Sadly, this belongs with https://twitter.com/volodarik/status/1657755496852475906

## Expected behavior

I can check all checkboxes.

## Context

- Browser and version: Edge 113.0.1774.57
- JupyterLab version: 4.0.1

andrii-i:
Also present in JupyterLab 3.6.3

brijsiyag:
I am able to re-produce. I want to give it a try, assign me if someone is not already working on this. Thanks !!

andrii-i:
Thank you for volunteering @brijsiyag

PR #15009, opened 2023-08-23
fixes codemirror settings

andrewfulton9:
## References

#14637 

## Code changes

The defaultFormData is updated to truly represent the defaults, where before if would take the current baseConfiguration as the default if possible

## User-facing changes

The user can now change two or more codemirror settings consistently and reliable 

https://github.com/jupyterlab/jupyterlab/assets/13774419/a91da208-a8fb-4bf7-b8a0-10a505e77a93

None that I am aware.

### DIFF
--- a/packages/codemirror-extension/src/services.tsx
+++ b/packages/codemirror-extension/src/services.tsx
@@ -146,14 +146,11 @@ export const extensionPlugin: JupyterFrontEndPlugin<IEditorExtensionRegistry> =
               () => registry.settingsSchema,
               []
             ) as any;
-            const editorConfiguration =
-              (props.formContext.settings as ISettingRegistry.ISettings).id ===
-              SETTINGS_ID
-                ? registry.baseConfiguration
-                : registry.defaultConfiguration;
             const defaultFormData: Record<string, any> = {};
             // Only provide customizable options
-            for (const [key, value] of Object.entries(editorConfiguration)) {
+            for (const [key, value] of Object.entries(
+              registry.defaultConfiguration
+            )) {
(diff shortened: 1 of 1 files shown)


krassowski (commented):
Thank you!

Do you have time to add a simple test here? It think it could be either done by moving the `fieldRenderer` implementation to `codemirror` package and unit-testing with jest, or adding a simple playwirght test to [galata/test/jupyterlab/settings.test.ts](https://github.com/jupyterlab/jupyterlab/blob/ac8f9e2f47b271785127757541a6fd8aa10c4d31/galata/test/jupyterlab/settings.test.ts) (I would not even add a snapshot, just ensure that the settings stay on after save by counting selector matches). Edit: the alternative would be opening an issue to track adding a test.

galata/test/jupyterlab/settings.test.ts
```diff
@@ -198,3 +198,35 @@ test.describe('change font-size', () => {
...
+  await page.evaluate(async () => {
+    await window.jupyterapp.commands.execute('settingeditor:open', {
+      query: 'CodeMirror'
+    });
+  });
+
+  expect(
+    await page.locator('.jp-PluginList .jp-FilterBox input').inputValue()
+  ).toEqual('CodeMirror');
+
```
krassowski, line 214:
```suggestion
  expect(
    await page.locator('.jp-PluginList .jp-FilterBox input').inputValue()
  ).toEqual('CodeMirror');

```

this is tested separately with 'Command Palette', let's simplify.

krassowski (approved):
Thank you @andrewfulton9!

krassowski:

PR #15015, opened 2023-08-24
Pass completion context to `shouldShowContinuousHint()`

### LINKED ISSUE #15014: Completion context is not ever passed to `shouldShowContinuousHint`
## Description

This was meant to be added in #14809 but slipped - the actual implementation did not thread it down and tests for it were not added.

krassowski:
## References

Fixes #15014

## Code changes

- Passes `ICompletionContext` on the level of reconciliator to properly implement the public API:
  https://github.com/jupyterlab/jupyterlab/blob/ac8f9e2f47b271785127757541a6fd8aa10c4d31/packages/completer/src/tokens.ts#L129-L138
    Note: it is only optional above to avoid breaking changes in API, but really it should always be available.
- Fixes a typo in related test
- Removes spurious type cast in handler's  `onTextChanged()`

## User-facing changes

Extensions (LSP) will be able to access context of when determining whether to automatically show completer after user typing text, which is critical for LSP.

None

### DIFF
--- a/packages/completer/src/handler.ts
+++ b/packages/completer/src/handler.ts
@@ -296,9 +296,8 @@ export class CompletionHandler implements IDisposable {
     }
     if (
       this._autoCompletion &&
-      (this._reconciliator as IProviderReconciliator)
-        .shouldShowContinuousHint &&
-      (this._reconciliator as IProviderReconciliator).shouldShowContinuousHint(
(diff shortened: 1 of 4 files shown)


jtpio (approved):
Thanks!

jtpio:

PR #15024, opened 2023-08-25
Do not shrink completer's documentation panel

### LINKED ISSUE #15023: When completer documentation panel is enabled, the completer width jumps
## Description

When moving between completion items with/without the documentation panel, the completer width jumps. This is a small CSS flex layout bug, one line fix on the way.

## Reproduce

compelter-resizes-when-doc-panel

## Expected behavior

The left side stays fixed.

## Context

- JupyterLab version: 4.0.05

krassowski:
## References

Fixes #15023

## Code changes

Adds `flex-shrink: 0` on documentation panel.

## User-facing changes

| Before | After |
|---|---|
| compelter-resizes-when-doc-panel | after |

None


krassowski:
Question: is the current 400px width a good default, or should we shrink it? Here are some options (of course we can go for something in between):

| 400px (current) | 350px | 300 px |
|--|--|--|
| Screenshot from 2023-08-26 00-28-17 | Screenshot from 2023-08-26 00-30-52 | Screenshot from 2023-08-26 00-31-21 |

krassowski:

jtpio:
> Question: is the current 400px width a good default, or should we shrink it? Here are some options (of course we can go for something in between):

If it was already `400px` before, maybe we can leave it like this for now and then discuss a new default separately?

jtpio (approved):
Thanks!

PR #15025, opened 2023-08-26
Expose `processedItemsCache` as a protected member

krassowski:
## References

Follow-up to #13663

## Code changes

Exposes `processedItemsCache` as a protected member of the default `CompleterModel` to allow sub-classes to re-use it. This will allow the completer model in the LSP extension to benefit from caching.

Because `processedItemsCache` needs to be nullfied among others in private `_reset()` method (and cache invalidation is hard, especially in sub-classes because the parent class may change without it being visible in subclass code) it is preferable to expose this property rather than re-implement it downstream. An alternative solution would be exposing protected signals for cache invalidation but I don't think this is worth the overhead.

## User-facing changes

None

None

### DIFF
--- a/packages/completer/src/model.ts
+++ b/packages/completer/src/model.ts
@@ -114,7 +114,7 @@ export class CompleterModel implements Completer.IModel {
     const ending = original.text.substring(end);
     query = query.substring(0, query.lastIndexOf(ending));
     this._query = query;
-    this._processedItemsCache = null;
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

jtpio:
@krassowski looks like this could / should be backported to `4.0.x`?

jtpio:

PR #15039, opened 2023-08-29
Fix Binder link in `README.md`

jtpio:
## References

The previous link was pointing to an outdated version of JupyterLab (`3.2.9`):

image

## Code changes

`README.md` edit.

## User-facing changes

Point to the latest commit of https://github.com/jupyterlab/jupyterlab-demo:

image

None


krassowski (approved):
I guess it made sense to pin hashes not only for better Binder caching (well, we do not change the demo repo often) but also for convenience when going back between branches. But it does make sense to have the main branch point to HEAD and only switch the pins when creating branches as needed. No strong opinion on this one, but for now it is an improvement, thanks!

PR #15042, opened 2023-08-29
Commit JS and Python packages in one commit on release

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13442

## Code changes

Instead of creating 2 commits on release, 1 for bumping the JS packages, 1 for the Python package, commit everything in a single commit.

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -11,7 +11,6 @@ on:
       steps_to_skip:
         description: "Comma separated list of steps to skip"
         required: false
-        default: "ensure-sha"
 
 jobs:
   publish_release:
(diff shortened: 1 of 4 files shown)


jtpio:
cc @blink1073 just checking if there was a reason for creating two commits in the first place? Or would it be fine to go with just one, as this would fix https://github.com/jupyterlab/jupyterlab/issues/13442?

blink1073:
I can't think of a good reason to prefer two commits, I think a single commit makes sense.

jtpio:
cc @jupyterlab/release for awareness

blink1073 (approved):
LGTM!

fcollonval (approved):
Thanks @jtpio 

---

is there a specific reason for holding back a backport to 4.0.x?

jtpio:
> is there a specific reason for holding back a backport to 4.0.x?

Right it should be possible to backport it.

But maybe it could be worth trying it out on a `4.1.x` alpha first.

krassowski:
Was it all fine with alpha releases @fcollonval?

fcollonval:
Yes it went well

For reference 0ffbf860155e484c093e104efbef8efe5f262708

Issue #6573, opened 2019-06-14
Screen reader with keyboard shortcuts


clapierre:
@ivanov commented on May 15
Using NVDA - we can confirm that

Ctrl-Shift-] and Ctrl-Shift-[ work for switching tab navigation, but no announcement of the tab name is made - so the user has no idea what tab they switched to.

Ctrl-B works to toggle the left pane, but again, does not announce.

Ctrl-Shift-B to toggle the single document mode works, but again does not announce.

Ctrl-Shift-C to focus the command palette works, but again does not announce.

Tab completion behavior - work, not announcing when initially invoked, and improperly repeating the text already typed, and not announcing the options themselves as they are navigated through.

Ctrl-S works to save, but does not announce the status. It should use an aria-live region.

Additionally, after discussing with James and Neil - you might want to enable that behavior to be customizable, to allow screen reader users to opt in and out of this behavior.

Ctrl-Shift-S brings up a non-real-dialog, but Tab behavior does allow navigation between those buttons.

Ctrl-F and Ctrl-G work - but no announcement of navigation. The default browser behavior announces the number of the result, at least.

Ctrl-Shift-L works - but again, no announcement (feedback) about a new panel having been created.

tonyfast:
these descriptions relate the lack of status messages referenced in #9399

PR #15048, opened 2023-08-30
Added aria-live to commands to improve screen reader accessibility. 

m158261:
This task implements an aria-live region into jupyterlab to improve screen reader accessibility when running commands either via the command palette or keyboard shortcut (i.e. Ctrl + B - close/open left sidebar).

Here is the aria-live documentation...

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

**Expected Functionality**

This functionality provides customisable text to be announced by a screen reader whenever a command is run either via the command palette or via a keyboard shortcut. This will provide an audible cue to screen reader users when a command has successfully been run. 

## References

https://github.com/jupyterlab/jupyterlab/issues/9399
https://github.com/jupyterlab/jupyterlab/issues/6573

Pull request in Lumino that implements these changes. 

https://github.com/jupyterlab/lumino/pull/627

## Code changes

The changes to application-extension get the aria-live region by it's ID then append text to the div by setting the inner HTML. This text will be read out by a screen reader and provides an audible cue when a command is run. 

## User-facing changes

Users that are using a screen reader will now get a prompt when they use a keyboard shortcut or access a command via the command palette.

@krassowski 
@gabalafou 
@isabela-pf


packages/application-extension/src/index.tsx
```diff
@@ -203,6 +203,7 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
 
     commands.addCommand(CommandIDs.close, {
       label: () => trans.__('Close Tab'),
+      commandDialog: () => trans.__('Close Tab'),
```
krassowski, line 206: 1. Why would we want to have a new property for this? I think we should reuse `label` and `caption`. This is be related to which shows and discusses how we can customize/caption for different uses https://github.com/jupyterlab/lumino/issues/570 
2. If a new property is really needed (for now I am not convinced), let's think of a new which explains the intent better. the `command dialog` is very unclear for me. Why add `command`? What is `dialog`?

krassowski: I would defer to @gabalafou here who I believe has more experience with screen readers.

m158261: Hey @gabalafou, do you have any suggestions on how to proceed with this? Thanks in advance.

isabela-pf:
Hello and thanks for the tag! Based on reading this, I'm going to try and summarize what I think you are saying the user experience should be to make sure I understand.

1. User is in JupyterLab and using a screen reader.
2. User activates the command palette via menu item, sidebar option (if configured for that), or via keyboard shortcut.
3. User focus is changed to the filter/search input box at the top of the command palette.
4. The screen reader announces this new focus by reading the default text filling that input box ("Search" if I recall correctly). 
5. User has been informed of where they are in the interface and what actions they can take from there.

Does this sound like what you intend? Or are there details I am missing?

m158261:
> Hello and thanks for the tag! Based on reading this, I'm going to try and summarize what I think you are saying the user experience should be to make sure I understand.
> 
> 1. User is in JupyterLab and using a screen reader.
> 2. User activates the command palette via menu item, sidebar option (if configured for that), or via keyboard shortcut.
> 3. User focus is changed to the filter/search input box at the top of the command palette.
> 4. The screen reader announces this new focus by reading the default text filling that input box ("Search" if I recall correctly).
> 5. User has been informed of where they are in the interface and what actions they can take from there.
> 
> Does this sound like what you intend? Or are there details I am missing?

Hi Isabela, thanks for commenting. That's almost the intended functionality. 

1. User is in JupyterLab using a screen reader.
2. User activates the command palette via menu item, sidebar option (if configured for that), or via keyboard shortcut.
3. User focus is changed to the filter/search input box at the top of the command palette.
4. The screen reader announces this new focus by reading the default text filling that input box ("Search" if I recall correctly).
5. The User activates a command. 
6. The screen reader informs the user that the command has been executed.

The alternative is...

1. User is in JupyterLab using a screen reader.
2. User activates a command using a keyboard shortcut i.e. opening the sidebar
3. The screen reader announces the command has been executed.

isabela-pf:
That mostly makes sense to me experience-wise. To be clear, I currently have no way of validating that this PR makes these experience changes. For anyone reviewing this that can validate it, that's what I would personally be looking for.

The only question I still have is about the meaning of
> the command has been executed.

Because we're talking about the command palette, I think you mean that there is feedback when (after opening the command palette) a command from the command palette has been executed. So there is no change in telling users once they've entered the command palette, only a change when a user selects a command to execute.

I did think we were just talking about opening the command palette at all, so that's my misunderstanding. Is the idea here that when a user selects a command, they do not know when that command has completed execution? So this feedback could come immediately or even minutes or hours later depending on what command is run?

packages/application-extension/src/index.tsx
tonyfast: if can you share  some rendered html from your local demo i might be able to help you. i'd need to see the structure of the dom elements you are trying to get announcements from.

m158261: Hey @tonyfast, thanks for helping out. I have attached a screenshot of the label div with the aria-live region. I am setting the innerHTML of this div when the open/close left sidebar command (ctrl + B) is run. As the text is changing in the aria-live region I would expect the screen reader (MS Narrator) to announce the text but it's not. It announces the keyboard shortcut instead. 

Any feedback or suggestions would be much appreciated. 

command dialog

m158261:
Hey @tonyfast, I have updated my implementation by moving the live region to the tab bar. This now appends text to the region when a command is run which is announced by the screen reader. Please let my know your thoughts and what the future direction of aria-live in jupyterlab should be. 

Thanks in advance!

tonyfast:
are these changes merged into https://github.com/t03857785/jupyterlab/tree/demo?

m158261:
> are these changes merged into https://github.com/t03857785/jupyterlab/tree/demo?

Not at the moment. I'll let you know when they are.

tonyfast:
thanks. since there is a lumino dependency i need to test both on that branch built together.

m158261:
Hey @krassowski, is there a way to get jupyterlab and lumino branches linked on binder? This is a two part PR with changes in both repos. I think it would be easier to test if I could get it running in binder.

tonyfast:
@m158261 this has been an on going challenge for years. we've had some duct tape solutions that worked temporarily. we really need a solution to this cause it has been holding a11y PRs for a bit. right now the best place to test these combinations in on https://github.com/t03857785/jupyterlab/tree/demo if they can get merged into there.

tonyfast:
on my demo build i can't get the `#tabbar-aria-live` region to change. it looks like this

```
<div aria-live="polite" role="region" id="tabbar-aria-live" class="lm-TabBar-ariaLive">Close Tab</div>
```

for aria live to work correctly we need to use an appending log. maybe i missed something in the lumino pr. i have an example reference implementation here https://github.com/deathbeds/nbconvert-a11y/blob/b27d98ba325a35c7d7469b87bcdb0ab1e07cf3ed/nbconvert_a11y/templates/a11y/static/settings.js#L93 .

if an element removes content from an aria live region before it is announced it won't be announced. i think we might want to consider an appending log so messages don't get lost. from this example, i have no been able to hear aria live changes when using the command palette or keyboard commands.

@m158261 are there specific command palette or keyboard actions i should try?

m158261:
> on my demo build i can't get the `#tabbar-aria-live` region to change. it looks like this
> 
> ```
> <div aria-live="polite" role="region" id="tabbar-aria-live" class="lm-TabBar-ariaLive">Close Tab</div>
> ```
> 
> for aria live to work correctly we need to use an appending log. maybe i missed something in the lumino pr. i have an example reference implementation here https://github.com/deathbeds/nbconvert-a11y/blob/b27d98ba325a35c7d7469b87bcdb0ab1e07cf3ed/nbconvert_a11y/templates/a11y/static/settings.js#L93 .
> 
> if an element removes content from an aria live region before it is announced it won't be announced. i think we might want to consider an appending log so messages don't get lost. from this example, i have no been able to hear aria live changes when using the command palette or keyboard commands.
> 
> @m158261 are there specific command palette or keyboard actions i should try?

Hey @tonyfast, thanks for the feedback. I have had to change the implementation as appending the live region to the tabbar didn't work. There are multiple tabbars rendered which means multiple live regions with the same ID. 

I've moved the region into JupyterLab and appended it to the dock panel. This means the Lumino PR is no longer necessary to test this. I have also appended the text to be announced rather than setting the innerHTML so it keeps the record of actions. 

I have tested this using Narrator. Opening/closing left and right sidebars are good commands to use. 

Do you think I should add a timestamp like in your implementation to make it more like a log? 

Thanks again for the input with this. I really appreciate the help.

m158261:
This works when running locally but not in Binder. The live region is present and text is being appended when commands are run but Narrator isn't announcing anything.

tonyfast:
> Hey @tonyfast, thanks for the feedback. I have had to change the implementation as appending the live region to the tabbar didn't work. There are multiple tabbars rendered which means multiple live regions with the same ID.

be aware that there are cases where you will have to have multiple live regions. the `nbconvert-a11y` implementation i shared with you has them in a few places. in my example, i'm using `dialog` elements that when active make all content outside the dialog `inert`. when aria live elements are `inert` they are NOT announced to the screen reader. in the case of dialogs, i need to carry around multiple activity logs so that aria live is announced. each new `dialog` i open has an activity log at the bottom so announcements can make it to the screen reader. the changes propagate to all of the activity logs on the page when i announce something. we'll need to be aware of these conditions if we use proper `dialog` elements .

> 
> I've moved the region into JupyterLab and appended it to the dock panel. This means the Lumino PR is no longer necessary to test this. I have also appended the text to be announced rather than setting the innerHTML so it keeps the record of actions.

this makes sense. i think there might be future cases where users or developers might want to explore the activity log. 

i'll rebuild a lite demo without the lumino pr. https://readthedocs.org/projects/jupyak/builds/22759565/

> 
> Do you think I should add a timestamp like in your implementation to make it more like a log?

i don't know if it needs to be there, but i thought this feature sounded like a user facing log so i borrowed design principles from system log. i think there is a future when some visits the activity log to undo past changes at specific times. i see time as a door into more visual features, audibly i don't feel like the time in important for the aria live announcement.

packages/application-extension/src/index.tsx
```diff
@@ -299,10 +317,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
       commands.addCommand(CommandIDs.toggleLeftArea, {
         label: trans.__('Show Left Sidebar'),
         execute: () => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           if (labShell.leftCollapsed) {
             labShell.expandLeft();
+            ariaLiveRegion!.append(trans.__('Show Left Sidebar'));
           } else {
             labShell.collapseLeft();
+            ariaLiveRegion!.append(trans.__('Closing Left Sidebar'));
```
tonyfast, line 326:
```suggestion
            ariaLiveRegion!.append(trans.__('Left Sidebar Closed'));
```

packages/application-extension/src/index.tsx
```diff
@@ -315,10 +336,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
       commands.addCommand(CommandIDs.toggleRightArea, {
         label: trans.__('Show Right Sidebar'),
         execute: () => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           if (labShell.rightCollapsed) {
             labShell.expandRight();
+            ariaLiveRegion!.append(trans.__('Opening Right Sidebar'));
```
tonyfast, line 342:
```suggestion
            ariaLiveRegion!.append(trans.__('Right Sidebar Closed'));
```

tonyfast (changes requested):
thanks for suggesting that test to run on my side. that really helped.

this work is so exciting to me for reasons:

* it works with Orca! it feels great. super :sunglasses:  stuff.
* i'm really excited we're thinking about `aria-live` it will add some must to the assistive experience and make it feel like a native screen reader application
* we can use https://github.com/deathbeds/jupyak/pull/16 to incrementally update your changes into a working jupyter lite ui that i can test with a screen reader. we've wanted this experience for years and it has arrived and feels good.

i tested open and closing the sidebar with <kbd>Ctrl+B</kbd> and i can confirm that the i am hearing the `aria-live` announcements. the announcements come in an active tense, but they are announced after the action completes. i made two suggestions that i think apply generally to these log messages. if a log happens before an event it should use the active phrasing you've applied and if the after happens after the action then use a past tense.

i can't confirm this at the moment. however, i bet if we run the IBM Equal Access checker will raise [Sensory Characteristics Needs Review](https://www.ibm.com/able/requirements/requirements/?version=v7_2#1_3_3). this is because of the use of left and right. this intention of the sensory characteristic is:

> Instructions provided for understanding and operating content do not rely solely on sensory characteristics of components such as shape, size, visual location, orientation, or sound.

_there should be a way to test actual content in one of these `jupyak` lite builds. i'll think about that, but in the mean time its hand done._

m158261:
> thanks for suggesting that test to run on my side. that really helped.
> 
> this work is so exciting to me for reasons:
> 
> * it works with Orca! it feels great. super 😎  stuff.
> * i'm really excited we're thinking about `aria-live` it will add some must to the assistive experience and make it feel like a native screen reader application
> * we can use [a11y prs deathbeds/jupyak#16](https://github.com/deathbeds/jupyak/pull/16) to incrementally update your changes into a working jupyter lite ui that i can test with a screen reader. we've wanted this experience for years and it has arrived and feels good.
> 
That's excellent. Really glad it's working. 
> i tested open and closing the sidebar with Ctrl+B and i can confirm that the i am hearing the `aria-live` announcements. the announcements come in an active tense, but they are announced after the action completes. i made two suggestions that i think apply generally to these log messages. if a log happens before an event it should use the active phrasing you've applied and if the after happens after the action then use a past tense.
> 
Ok, I'll review the announcements and make sure they're in the correct tense. 
> i can't confirm this at the moment. however, i bet if we run the IBM Equal Access checker will raise [Sensory Characteristics Needs Review](https://www.ibm.com/able/requirements/requirements/?version=v7_2#1_3_3). this is because of the use of left and right. this intention of the sensory characteristic is:
> 
> > Instructions provided for understanding and operating content do not rely solely on sensory characteristics of components such as shape, size, visual location, orientation, or sound.
> 
That's a good point. I'll see what I can do. I used left/right to describe sidebars as that's what they're named in the command palette. 
> _there should be a way to test actual content in one of these `jupyak` lite builds. i'll think about that, but in the mean time its hand done._

tonyfast:
> That's a good point. I'll see what I can do. I used left/right to describe sidebars as that's what they're named in the command palette.

yea i figured that would be a bigger issue with the design system. i don't know what is better phrasing either. "primary" "secondary" doesnt make sense because its really subjective where icons in sidebars are.

m158261:
Hey @krassowski, I'm happy for this to be re-reviewed. Please let me know what you think. 

Thank you @tonyfast for your input with this.

tonyfast:
This is some great work @m158261   Thanks for braving new territory. it is necessary to start experimenting with aria live because of all the state changes that can happen in the application. I hope this feature can be highlighted in the release notes as a step towards more assistive experiences; it is a significant step beyond wcag. We will have to be cautious about regressions in this feature since it can’t be tested automatically. I hope we can document it rigorously to avoid that.

packages/application-extension/src/index.tsx
```diff
@@ -334,10 +358,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
             ? trans.__('Show Right Activity Bar')
             : trans.__('Show Left Activity Bar'),
         execute: args => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           if (args.side === 'right') {
             labShell.toggleSideTabBarVisibility('right');
+            ariaLiveRegion!.append(trans.__('Show Right Activity Bar'));
```
krassowski, line 364: Should it be "showing" rather than "show"?

packages/application-extension/src/index.tsx
```diff
@@ -353,7 +380,9 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
       commands.addCommand(CommandIDs.togglePresentationMode, {
         label: () => trans.__('Presentation Mode'),
         execute: () => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           labShell.presentationMode = !labShell.presentationMode;
+          ariaLiveRegion!.append(trans.__('Presentation Mode'));
```
krassowski, line 385: Should it be "entering presentation mode"? Similar in a few other places.

packages/application-extension/src/index.tsx
```diff
@@ -315,10 +336,13 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
       commands.addCommand(CommandIDs.toggleRightArea, {
         label: trans.__('Show Right Sidebar'),
         execute: () => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           if (labShell.rightCollapsed) {
             labShell.expandRight();
+            ariaLiveRegion!.append(trans.__('Right Sidebar Closed'));
```
krassowski, line 342: I do not follow why it says "Right Sidebar Closed". It was closed but it is getting opened, right? I just don't feel that the content of these hints is exactly right.

packages/application-extension/src/index.tsx
```diff
@@ -385,24 +416,28 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
         execute: () => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           // Turn off presentation mode
           if (labShell.presentationMode) {
             commands
               .execute(CommandIDs.togglePresentationMode)
               .catch(reason => {
                 console.error('Failed to undo presentation mode.', reason);
               });
+            ariaLiveRegion!.append(trans.__('Turn off presentation mode'));
```
krassowski, line 440: Since it happens after the action was performed, should it be more like "Presentation mode turned off"?

Should it be in `.then()` to only show when action succeeded?

Should there be an alternative next to `console.error` in `catch` when it failed?

packages/application-extension/src/index.tsx
```diff
@@ -412,6 +447,7 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
             commands.execute(CommandIDs.toggleHeader).catch(reason => {
               console.error('Failed to display title header.', reason);
             });
+            ariaLiveRegion!.append(trans.__('Display top header'));
```
krassowski, line 450: Again, I think it should be in `.then()`, and the `console.error` could be replaced by a method that writes to both console and the aria live region. Similarly in other places.

packages/application/style/core.css
```diff
@@ -90,3 +90,8 @@ body {
   display: flex;
   align-items: center;
 }
+
+.jp-ContentPanel-ariaLive {
+  width: 0;
+  overflow: hidden;
```
krassowski, line 96: I do not follow this one. Was the intent to use `visibility: hidden`?

packages/application-extension/src/index.tsx
```diff
@@ -208,9 +208,11 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
         return !!widget && widget.title.closable;
       },
       execute: () => {
+        let ariaLiveRegion = document.getElementById('commands-aria-live');
```
krassowski, line 211: Can you please create a `const ARIA_LIVE_ID = 'commands-aria-live'` at the top of the file and use it instead?

packages/application-extension/src/index.tsx
```diff
@@ -372,9 +401,11 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
           return mode === 'single-document' || mode === 'multiple-document';
         },
         execute: args => {
+          let ariaLiveRegion = document.getElementById('commands-aria-live');
           const mode = args['mode'] as string;
           if (mode === 'single-document' || mode === 'multiple-document') {
             labShell.mode = mode;
+            ariaLiveRegion!.append(trans.__(`${mode} mode`));
```
krassowski, line 408: We cannot use JS templated strings with translations, it needs to use `%` and pass the variable in the second argument, see: https://jupyterlab.readthedocs.io/en/4.0.x/extension/internationalization.html

krassowski (changes requested):
My review echoes what @tonyfast has pointed previously:

> the announcements come in an active tense, but they are announced after the action completes

I think that every string added needs to be reviewed to ensure that it accurately reflects that an action has been completed (or not).

packages/application-extension/src/index.tsx
m158261: Changes made.

packages/application-extension/src/index.tsx
m158261: Changes made.

packages/application/style/core.css
m158261: An aria-live region must be visible on the DOM for the screen reader to pick up changes to it. This code makes the div visually hidden while visible to the screen reader.

[Relevant Stack Overflow Page](https://stackoverflow.com/questions/50747587/announce-aria-live-text-change-despite-div-being-hidden)

m158261:
Hey @krassowski, thank you for the feedback. I have gone through the announcements and made sure they reflect the tense of the action. Most of the announcements come after the action has completed. 

I have addressed the other comments too. Please let me know what you think. 

Many thanks!

packages/application/style/core.css
krassowski: Ok, in this case, can you also add `height: 0;` and a comment to explain the intent?

tonyfast: here is a recommended pattern for visually hidden elements https://www.tpgi.com/the-anatomy-of-visually-hidden/

cf https://www.a11yproject.com/posts/how-to-hide-content/

tonyfast: might be worth having `sr-only` somewhere in lab or lumino.

krassowski: Thanks @tonyfast, I agree we should follow what you linked (with a reference in a comment).

> might be worth having `sr-only` somewhere in lab or lumino

+1. I would suggest to name it `jp-mod-sr-only` to avoid conflicts with any third party styles if they have it slightly different, but for the purposes of getting this PR in I would suggest that having a more generic class could be postponed (as this would become a public API we need to think more where to put it, document it etc).

m158261: Thanks @tonyfast for that documentation. I've added it to the PR description. 

Shall I leave the name as `.jp-ContentPanel-arialive.` for now and address `sr-only` in a future PR?

m158261:

JasonWeill:
This code change would be considered after #16145 is implemented. Thank you for your contribution!

PR #15052, opened 2023-09-01
`HoverBox`: adjust `right` when moving node to fit within window

### LINKED ISSUE #14845: Bug in calculation in `HoverBox.setGeometry`
## Description

I am using [`HoverBox.setGeometry`](https://github.com/jupyterlab/jupyterlab/blob/3e45cb9e3038b8e40274ac7cb4d03c1be9439768/packages/ui-components/src/hoverbox.ts#L131) to position a widget in an extension, and want to pass `stick-inside` for all options in [`outOfViewDisplay`](https://github.com/jupyterlab/jupyterlab/blob/3e45cb9e3038b8e40274ac7cb4d03c1be9439768/packages/ui-components/src/hoverbox.ts#L109). However, it does not seem to function as expected for the `right` option. After lot of debugging, I've found an issue in the calculation.

## Reproducer & Bug

https://github.com/jupyterlab/jupyterlab/blob/3e45cb9e3038b8e40274ac7cb4d03c1be9439768/packages/ui-components/src/hoverbox.ts#L226-L237

Consider the above code, and a case when `anchor.left` is located on the very right of the screen. At line `228`, we update the `node`'s left to `anchor`'s left (assume `offsetHorizontal` to be zero for simplicity). Suppose `node` has some width, then virtually, at this point the right edge of the `node` is out of window, and the condition at line `234` will be true. Here, `left` is correctly adjusted so that the right edge of `node` touches  right edge of window. 

**The bug lies here in re-calculation of width** - 
https://github.com/jupyterlab/jupyterlab/blob/3e45cb9e3038b8e40274ac7cb4d03c1be9439768/packages/ui-components/src/hoverbox.ts#L270

Since, we reduced `right - window.innerWidth` from `left`, the new `right` now is actually just `window.innerWidth`. Because the original `right` is out of the window, `width` calculated here is actually greater than original.

`width` is re-used in below code, which finally leads to wrong re-setting of `left` ,and node is placed wrongly horizontally - 

https://github.com/jupyterlab/jupyterlab/blob/3e45cb9e3038b8e40274ac7cb4d03c1be9439768/packages/ui-components/src/hoverbox.ts#L374-L378

Cases when bug can be noticed - When `anchor.left` is less than node's `width` away from right edge of the window.

## Proposed Solution

One solution could be to adjust `right` correctly if `left` was adjusted before re-calculation of width as explained above.

## Context

- Operating System and version: Linux RHEL8
- Browser and version: Chrome 114.0.5735.199 
- JupyterLab version: 4.0.0

krassowski:
## References

Fixes #14845

## Code changes

- starts unit tests for hover box, previously only tested via completer code
- implements solution proposed in #14845

## User-facing changes

None (because we only use `right = 'stick-inside'` in JupyterLab core), unless in extensions where hover box positioning will behave better for anchors placed by the right edge.

None

### DIFF
--- a/packages/ui-components/src/hoverbox.ts
+++ b/packages/ui-components/src/hoverbox.ts
@@ -230,9 +230,10 @@ export namespace HoverBox {
     let rect = node.getBoundingClientRect();
 
     // Move left to fit in the window.
-    const right = rect.right;
+    let right = rect.right;
     if (right > window.innerWidth) {
       left -= right - window.innerWidth;
+      right = window.innerWidth;
       node.style.left = `${Math.ceil(left)}px`;
     }
 
--- a/packages/ui-components/test/hoverbox.spec.ts
+++ b/packages/ui-components/test/hoverbox.spec.ts
@@ -1,0 +1,222 @@
+import { HoverBox } from '@jupyterlab/ui-components';
+
+function createDomRect(options: {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+}): DOMRect {
+  return {
+    ...options,
+    bottom: options.x + options.height,
+    top: options.y,
+    left: options.x,
+    right: options.x + options.width,
+    toJSON: () => 'DummyDOMRect'
+  };
+}
+
+function createPointAnchor(options: { x: number; y: number }): DOMRect {
+  return createDomRect({ ...options, width: 0, height: 0 });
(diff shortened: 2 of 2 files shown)


jtpio (approved):
Thanks!

jtpio:

Issue #9646, opened 2021-01-20
Access IPython command history via key-bindings in Jupyter Notebooks 


rsaim:
### Problem

Similar to IPython, can we support accessing command history via Up/Down (or other) keys in Jupyter notebooks? This would be same as history-search-* feature commonly used in shells and in IPython.

I am aware that we can use %history feature to copy-paste snippets. However, Searching and then copy-pasting code is too slow and annoying. The feature will make the development which involve both IPython and Jupyter much faster. 

### Proposed Solution

Support accessing IPython history via Up/Down arrows in Jupyter Notebooks.

### Additional context
We can access the cell history of Jupyter Notebooks in IPython with simple Up/Down arrow key bindings. The feature is to do the opposite - access IPython command history in Jupyter via some key bindings.

This is not about accessing shell history in Ipython or Jupyter.

echarles:
> Support accessing IPython history via Up/Down arrows in Jupyter Notebooks.

What would be the behavior in case of the notebook cells with those Up/Down arows. For now, those keystrokes allows the user to navigate within the cell. If the cursor reached the top or bottom of the cell, it jumps to the neighbour cell.

I guess we would need to bind the something else than "Shift-Up/Down" which selects content, or "Cmd-Up/Down" which forces to jump directly to the top or bottom of the cell.

Any proposal for such a new keystroke to support the history?

mlucool:
`Alt Up/Down`?

echarles:
I have started working on this feature in this branch https://github.com/datalayer-externals/jupyterlab/tree/feat/notebookExecutionHistory

An history already exists for the console, but the requirement for that are very specific to the console (1 editor, relying on the edge movement to trigger the history), so it is better to not reuse/depend on that implementation 

The current work implements a history at the notebook package level. On a user perspective, I wonder if the history should be shared across all cells, or if it should be tracked cell-by-cell.

If we want to track at cell level, it may make more sense to implement the history at the codeeditor level, but in that case we may come in conflict soon with the work being done with CodeMirror6, and for sure, with the current implementation for the console.

It may sound safer to leave the history at the notebook level, but we need to decide if we want a shared history, or a cell-per-cell history.

krassowski:
Closing as implemented in #15062 and released in JupyterLab 4.1.0.

Issue #8776, opened 2020-08-06
Switch to using gifs in Jupyterlab documentation


marthacryan:
The [current documentation](https://jupyterlab.readthedocs.io/en/latest/) uses many 10-15 second youtube videos ([code console documentation](https://jupyterlab.readthedocs.io/en/latest/user/code_console.html) for example) that could be easily replaced with gifs. Using gifs would also make it easier to add a PR for new documentation because with youtube videos, you'd need to have access to the youtube channel to add videos into the documentation.

lgfunderburk:
I'd like to have this assigned please.

blink1073:
Done, thanks @lgfunderburk!

lgfunderburk:
Thank you!

jtpio:
In today's Jupyter Community Call someone mentioned the idea of automatically generating these videos for the docs, using a tool like [playwright](https://playwright.dev/) which can render videos of the tests suites.

We could then use a tool like `playwright codegen` to record a script based on user actions: https://github.com/microsoft/playwright-python#record-and-generate-code

Not sure whether videos via `<video>` elements are natively supported in the docs, otherwise it would require a second step to convert them to gifs.

krassowski:
1. I am a proponent of using auto-generated screencasts as it would also help to:
- provide documentation translation (if we should decide to do that)
- ensure that no outdated interface/settings are shown on the videos (see https://github.com/jupyterlab/jupyterlab/issues/9804#issuecomment-779958528)

2. Auto-playing gifs can be troublesome if they do not have play/pause button. I am using GIFs in readme of one project to document some of the settings and recently as user pointed out that they are moving a little bit too fast for them (https://github.com/krassowski/jupyterlab-lsp/issues/548#issuecomment-786008719). A gif with play/pause button would be ideal.

3. Auto-playing indicates one of two things: a) the browser downloads a lot of gif which is heavy and not nice for some users, or b) the gifs are cropped/low resolution (not nice for other users). I think that gif that play on clicking would be preferrable 

3. Can we make the GIFs accessible? Can we provide them with plain text description of what is happening (see also discussion linked in point 2)? In ideal scenario one could have one-to-one correspondence between actions on screen and the description (which could be auto-generated when using [RobotFramework-JupyterLibrary](https://robotframework-jupyterlibrary.readthedocs.io/en/latest/); not sure if galata can do this too; my biggest problem with galta is that I have not seen it yet).

jtpio:
[...]

Issue #6823, opened 2019-07-13
Replace all YouTube embeds with auto-playing GIFS


van-8:
- [ ] Make list of all necessary gifs
- [ ] Export, optimize, compress gifs
- [ ] Upload gifs to documentation

van-8:
Assign to myself? I don't have permissions.

ian-r-rose:
Is there a reason you want to replace with auto-playing Gifs? We had a long discussion about this some time ago here: https://github.com/jupyterlab/jupyterlab/pull/3689 ,and decided upon YouTube.

I'm open to switching to a different format (I still think mp4s might be a good choice), but we should probably re-litigate the discussions in the previous issue.

van-8:
Thanks for highlighting the previous discussion, Ian! Didn't see it. After reviewing the thread, I'd make the following observations from the perspective of a total newbie to both coding and JupyterLab who used the documentation to learn just a few weeks ago:

1. It breaks the flow of learning to click the YouTube play button to access the visual example. I've found other tech docs using gifs as a smooth learning transition from "text" into "visual example" as your eyes move down the page.
2. If the visual example is less than 10 seconds, a gif feels more appropriate. My expectations for a video is at least 30 seconds or more. But maybe that's just longtime Reddit lurker behavior...
3. If file size or hosting is the issue, then I defer to more experienced judgement.
4. As a "total newbie" to Jupyter, I've found myself preferring long thorough tutorials. I start from zero, so want to learn all the Basics in one swoop. I've found [Corey Schaffer's outdated tutorial](https://www.youtube.com/watch?v=HW29067qVWk&t=1282s) the best one on Youtube and watched it twice. As I'm slowly transitioning to "Familiar User," the docs are great for searching specific actions or advanced features or troubleshooting.

valenmgama:
I was about to open a new issue when I found this one. (Note #8776 seems to be the same issue.)

Answering your question: @ian-r-rose 
> Is there a reason you want to replace with auto-playing Gifs? We had a long discussion about this some time ago here: #3689 ,and decided upon YouTube.
> 
> I'm open to switching to a different format (I still think mp4s might be a good choice), but we should probably re-litigate the discussions in the previous issue.

I think an important thing to consider is that when opening the **JupyterLab Reference** from within a Jupyter Lab instance itself the YouTube videos won't load and you are left with a black screen. At least that is what I am getting running JupyterLab on Google Chrome. I am attaching a picture.

[...]

PR #15062, opened 2023-09-01
Notebook execution history

andrewfulton9:
## References

#9646

follows up on @echarles work [here](https://github.com/datalayer-externals/jupyterlab/tree/feat/notebookExecutionHistory)

## Code changes

- Adds keybindings "Alt ArrowUp" and "Alt ArrowDown" to notebook to iterate through kernel history in an activated cell
- Adds history code file to notebook for managing the kernel history iteration file
- Adds code to notebook widget file to add history functionality

## User-facing changes

This PR gives the user the ability to use Alt + up or down arrows to iterate through the kernel history. 

https://github.com/jupyterlab/jupyterlab/assets/13774419/e1532299-857d-4009-8b61-674152e06701

None that I am aware


packages/notebook/src/widget.ts
```diff
@@ -1033,6 +1036,11 @@ export namespace StaticNotebook {
      * The application language translator.
      */
     translator?: ITranslator;
+
+    /**
+     * The kernel history retrieval object
+     */
+    history?: INotebookHistory;
```
krassowski, line 1043: Maybe worth being more specific as `inputHistory`? The other kinds that quickly come to mind are "editHistory" (undo stack), "checkpointHistory". I think this should be reflected in the name of attributes and interface.

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,375 @@
...
+    this._retrieveHistory(cell);
+  }
+
+  /**
+   * retrieve the history from the kernel
+   *
+   * @param cell - The string to use when filtering the data.
+   */
+  private async _retrieveHistory(cell?: Cell): Promise<void> {
+    return this._kernel?.requestHistory(Private.initialRequest).then(v => {
```
krassowski, line 313: Can be rewritten using await

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,375 @@
...
+   * @param activeCell - The currently selected Cell in the notebook.
+   *
+   * @returns A Promise for console command text or `undefined` if unavailable.
+   */
+  back(activeCell: Cell): Promise<string> {
+    this.checkSession(activeCell);
+    --this._cursor;
+    this._cursor = Math.max(0, this._cursor);
+    const content = this._filtered[this._cursor];
+    console.log('this._cursor: ', this._cursor);
```
krassowski, line 180:
```suggestion
```

packages/notebook-extension/schema/tracker.json
```diff
@@ -366,6 +366,16 @@
     ]
   },
   "jupyter.lab.shortcuts": [
+    {
+      "command": "notebook:access-last-history",
+      "keys": ["Alt ArrowUp"],
```
krassowski, line 371: Let's move these down to all other commands with `.jp-Notebook.jp-mod-editMode` selector?

packages/notebook/src/widget.ts
```diff
@@ -1274,6 +1282,32 @@ export class Notebook extends StaticNotebook {
     this.addFooter();
   }
 
+  /**
+   * Access previous item in ipython kernel for current kernel
+   */
+  accessLastHistory(): void {
```
krassowski, line 1288: I am worried about this being added on `Notebook` class - it extends the public API by two non-optional methods. What if instead the implementation was moved to respective `NotebookActions` functions (which currently just call these methods) and `history` getter were made public as an optional member on `Notebook`? This way is also easier to preserver backward compatibility. In this scenario we would review `INotebookHistory` to make it as narrow as possible.

packages/notebook/src/widget.ts
```diff
@@ -1274,6 +1282,32 @@ export class Notebook extends StaticNotebook {
...
+    const activeCell = this.activeCell;
+    if (activeCell) {
+      if (this._history) {
+        const lastHistory = this._history.back(activeCell);
+        this._history.updateEditor(activeCell, lastHistory);
```
krassowski, line 1293: I would suggest implementing this as async rather than passing a promise.

```suggestion
        const lastHistory = await this._history.back(activeCell);
        this._history.updateEditor(activeCell, lastHistory);
```

packages/notebook/src/actions.tsx
```diff
@@ -2076,6 +2076,24 @@ export namespace NotebookActions {
       }
     });
   }
+
+  /**
+   * Access last notebook history.
+   *
+   * @param notebook - The target notebook widget.
+   */
+  export function accessLastHistory(notebook: Notebook): any {
```
krassowski, line 2085:
```suggestion
  export function accessLastHistory(notebook: Notebook): void {
```

same below. But I really it feels like this should return `Promise<void>` - also see comments on implementation in `Notebook` class.

krassowski (commented):
I left some early comments on API and naming. Since this is a new feature we will need tests and documentation.

andrewfulton9:
Thanks for the review @krassowski! I'll work on addressing all these comments and update ASAP

packages/notebook/src/widget.ts
andrewfulton9: Does it make sense to not have them be promises or async at all. As far as I can tell all the calls within the function are synchronous so maybe it makes sense to just return a string. with no await.

packages/notebook/src/actions.tsx
andrewfulton9: I'm not sure I understand. What's the advantage of making these Promises?

krassowski: The advantage is that when a third-party code calls `NotebookActions.accessPreviousHistory()` (or respective command) they will be able to await it and only then perform another action. This is useful for downstream automation and performance testing. If the promise is dangling at this level we have no way to know when the action is done. Returning the promise rather than loosing it also makes it easier to unit test it.

andrewfulton9: I probably should have been more clear this question I think depends on the above question that I asked in a different thread:

> Does it make sense to not have them be promises or async at all? As far as I can tell all the calls within the function are synchronous so maybe it makes sense to just return a string. with no await.

If I make the underlying back and forward calls regular synchronous functions, then there shouldn't be any dangling promise, right?

krassowski: Yes, but should they really by synchronous? On on hand requesting history is always async, on the other I see that the current implementation just requests it after each cell execution. This may be sub-optimal (because we generate overhead and wire traffic on each cell execution) and if the response is delayed may lead to subpar UX (being presented with outdated history). My thinking is that it should be async all along. This is, we would invalidate history on every execution but only request it when user asks for it. Then navigating within history can be from cache to improve response time.

andrewfulton9: Oh I see. That makes a lot of sense. I made a push that should address that. Let me know what you think. I am still working on adding tests and docs though.

andrewfulton9:
@krassowski, I've added a test for this. Let me know if you think it's sufficient, or I should add something.

Do you have a recommendation for where to add a mention of this feature for the user documentation? I was thinking docs/source/user/notebook.rst  or docs/source/user/documents_kernels.rst. I'm also seeing that a lot of actions have youtube videos along with the description. Is there any documentation on how to add a video or is that something that has to be done by someone with rights to the Jupyter youtube channel?

krassowski:
If I remember correctly we wanted to move away from YouTube videos for user-facing documentation (https://github.com/jupyterlab/jupyterlab/issues/8776 and https://github.com/jupyterlab/jupyterlab/issues/6823) and are not adding new ones. Either notebook or kernels section of user guide sounds good to me; a single paragraph would do fine, but if you think a video is useful, it would ideally be generated using playwright. This is however a bit difficult because there is no API to start/stop recording at specific time in the test case.

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  /**
+   * handle a cell execution
+   */
+  private async _onExecuted(
+    _: unknown,
+    args: { notebook: StaticNotebook; cell: Cell }
+  ): Promise<void> {
+    const cell = args['cell'];
+    await this._retrieveHistory(cell).catch();
```
krassowski, line 292: This is still fetching 500 entries of history on each cell execution. I just tried this locally and for my current history it is extra 180kB on the websocket wire for each cell; execute a notebook with 100 cells and you got 18MB noise on websocket. This is compared to execution and reply messages being barely 300-600 bytes. 

I would suggest we only fetch history once when it is first requested by the user.

docs/source/user/notebook.rst
```diff
@@ -151,6 +151,10 @@ and select “New Console for Notebook”:
      <iframe src="https://www.youtube-nocookie.com/embed/eq1l7DBngQQ?rel=0&amp;showinfo=0" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
   </div>
 
+.. _kernel_history:
+
+You can iterate through the kernel history in a document cell using ``Alt Up-Arrow`` and ``Alt Down-Arrow``.
```
krassowski, line 156: Can this feature be disabled? Should it be opt-in? I am specifically concerned about sending a lot of history which is possibly originating from root_dir other than current; some changes I suggest in other comments (only requesting on first user action triggering `notebook:access-previous-history-entry`/`notebook:access-next-history-entry`, reducing _n_) would alleviate the issue but it still feels not a great thing to do without warning users about it.

krassowski:
Kicking CI after bot update.

packages/notebook/src/history.ts
andrewfulton9: I have it fetching history on each reset so that newly run cells are included in the history without manually appending them. Also, fetching it each time allows a user to access the history from different kernels as well, though they would have to scroll through the history of the current kernel first. What if I implemented something where it only requested a few at a time, requesting more as the user scrolls farther back through the history.

I'm also now wondering if accessing different kernel histories could be considered a security risk with multiple users on jupyter_server? Would users be able to access the kernel history for other users as it is now?

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  /**
+   * retrieve the history from the kernel
+   *
+   * @param cell - The string to use when filtering the data.
+   */
+  private async _retrieveHistory(cell?: Cell): Promise<void> {
+    return await this._kernel
+      ?.requestHistory(Private.initialRequest)
```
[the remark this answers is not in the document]
andrewfulton9: Any suggestion on how this should be handled? It seems easy enough to catch this, but I'm wondering how it should be handled after the catch. Would a notification.warning with autoClose be sufficient?

docs/source/user/notebook.rst
andrewfulton9: Yeah, I think that could be a good idea. Could you point me to an example of another feature that is opt-in that I could use as an example?

packages/notebook/src/history.ts
krassowski: It is fine to fetch say 10-20 items at a time and then maybe more if there is no overlap with what was fetched before. All I am arguing is that it should never happen on cell execution, but only when user does invoke `notebook:access-previous-history-entry`/`notebook:access-next-history-entry`. We could use cell execution as a heuristic for how many items to fetch; in pseudo-code:

```ts
class {
  onCellExecuted() {
     this._minToFetch += 1;
  }
  
  previousEntry() {
    this._ensureHistorySynced();
    this._setPrevious()
  }
  
  private _ensureHistorySynced() {
    let historyReconcliliated = false;
    let initialToFetch = this._minToFetch + 5;
    let historyChunk = this._fetch(initialToFetch);
    const historyChunks = [];
    while (!historyReconcliliated) {
      if (historyChunk overlaps this._historyCache) {
        historyChunks.push(historyChunk[overlapStart:]);
        historyReconcliliated = true;
      } else {
        // this assumes no execution in the meantime
        historyChunks.push(historyChunk);
        historyChunk = this._fetch(10);
      }
    }
    this._historyCache.push(...reversed(historyChunks));
    this._minToFetch = 0;
  }
  private _minToFetch = 0;
}
```

Though this will be tricky to get right. Do we really want to have history from concurrently executing kernels though?

Maybe using `session`, `start` and `stop` arguments of, `history_request` could help? In particular if we restrict ourselves to single `session`, the problem of cross-session history leakage would be eliminated (but then the history would be more limited).

packages/notebook/src/history.ts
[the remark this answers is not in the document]
krassowski: I would say that this is even fine to have as `console.warn` - this is a bit pathological case of kernel dying; it is just a bad practice to have uncaught exceptions because these sometimes cause trouble to users with extensions using catch-all handlers (and make user bug reports noisy).

docs/source/user/notebook.rst
krassowski: In notebook context, side-by-side rendering, in completer there is documentation panel (though to be honest these are not very well documented).

packages/notebook/src/history.ts
andrewfulton9: Yeah I was on the fence on whether or not to include multiple kernel histories or just the history of the notebook kernel, I ultimately decided to include them all since that's how the console history works. I did mostly implement the single notebook kernel history though, and that's actually why cell history is collected on execution. I couldn't figure out another way to get the kernel id number without collecting from the first history item immediately after a cell in the notebook was run, though if we go that way it should probably only call history after the first execution. So it seems to me the options are:

1) include whole history, and only collect on access commands or
2) include only history from current kernel, getting kernel number from first execution of the session

and then with either of those options it can either:

a) grab history in intervals or
b) grab n number (what we currently have)

What are your thoughts?

andrewfulton9: @rsaim, do you have an opinion on this?

andrewfulton9:
@krassowski, I added in the batch logic for pulling history to this PR. It still includes the whole ipython history. I also made it so that scrolling through the history appears linear, even if history items are added above the cursor. There is an efficiency tradeoff of doing it like this, in that while accessing near history items it's considerably cheaper for network traffic, scrolling deep will also be considerable more expensive.

packages/notebook/src/widget.ts
```diff
@@ -1189,6 +1197,7 @@ export namespace StaticNotebook {
      * - 'none': Attach all cells to the viewport
      */
     windowingMode: 'defer' | 'full' | 'none';
+    accessKernelHistory: boolean;
```
fcollonval, line 1200: Could you add a doc string for this interface?

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,419 @@
...
+  export interface IOptions {
+    /**
+     * The client session used by the foreign handler.
+     */
+    sessionContext: ISessionContext;
+
+    /**
+     * The application language translator.
+     */
+    translator: ITranslator;
```
fcollonval, line 408: The constructor supports optional translator but not this interface:

```suggestion
    translator?: ITranslator;
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,419 @@
...
+    this._cursor = Math.min(this._filtered.length - 1, this._cursor);
+    const content = this._filtered[this._cursor];
+    return content;
+  }
+
+  /**
+   * Get the next item in the console history.
+   *
+   * @param activeCell - The currently selected Cell in the notebook.
+   * @param update - the promise returned from back or forward
```
fcollonval, line 217: Let's drop _promise_ as it is the content

```suggestion
   * @param update - the result from back or forward
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,419 @@
...
+  /**
+   * Reset the history navigation state, i.e., start a new history session.
+   */
+  reset(): void;
+
+  /**
+   * Get the next item in the console history.
+   *
+   * @param activeCell - The currently selected Cell in the notebook.
+   * @param update - the promise returned from back or forward
```
fcollonval, line 69:
```suggestion
   * @param update - the result from back or forward
```

packages/notebook/src/widget.ts
```diff
@@ -989,6 +991,7 @@ export class StaticNotebook extends WindowedList {
   private _idleCallBack: number | null;
   private _mimetype: string;
   private _mimetypeService: IEditorMimeTypeService;
+  readonly _history: INotebookHistory | undefined;
```
fcollonval, line 994: I would recommend making this one protected

```suggestion
  protected readonly _history: INotebookHistory | undefined;
```

fcollonval (commented):
Thanks @andrewfulton9  (and thanks @krassowski for the ping in the call).

i have a couple of minor suggestions.

packages/notebook/src/widget.ts
fcollonval: And it should be optional as it is changing the API for `set notebookConfig`. This probably requires to modify that setter to use the default value if it is not provided.

```suggestion
    accessKernelHistory?: boolean;
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,419 @@
...
+ */
+export class NotebookHistory implements INotebookHistory {
+  /**
+   * Construct a new console history object.
+   */
+  constructor(options: NotebookHistory.IOptions) {
+    this.sessionContext = options.sessionContext;
+    this.translator = options.translator || nullTranslator;
+    void this._handleKernel();
+    this.sessionContext.kernelChanged.connect(this._handleKernel, this);
```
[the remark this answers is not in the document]
andrewfulton9: good point, I think you're right about that

packages/notebook/src/widget.ts
krassowski: We need a way to access history:

```
      ../notebook/src/actions.tsx(2100,20): error TS2445: Property '_history' is protected and only accessible within class 'StaticNotebook' and its subclasses.
```

Maybe it needs to be just `readolny` (public) but then it should probably not have `_` prefix. This is similar to the story with `footer`. Thoughts?

andrewfulton9: Right, it used to be private and notebook had its own back/forward methods that called history from the `Notebook` class which were in turn called from `accessPreviousHistory` and `accessNextHistory` in [actions](https://github.com/jupyterlab/jupyterlab/blob/a37354a42b6d63f64dcb0e2a16216897b5919348/packages/notebook/src/actions.tsx#L2092). Now those just directly use `notebook._history`. So it seems we should either revert the code to keep `_history` private or just make it readonly and lose the `_` prefix. I would lean towards the second option.

krassowski: > I would lean towards the second option.

Let's do that to unblock this pull request. We will still have some time to change things before 4.1 release if we decide it was not the right call.

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+    this._hasSession = false;
+    this._placeholder = '';
+    this.toRequest = this.requestBatchSize;
+  }
+
+  /**
+   * Fetches a subsequent batch of history. Updates the filtered history and cursor to correct place in history,
+   * accounting for potentially new history items above it.
+   */
+  async fetchBatch() {
```
krassowski, line 254: private/protected?

Would it be worth adding a jest unit test for this logic (indirectly, by testing `back` which is the public API for it)?

andrewfulton9: Good call. I changed it to private, and added a test as well

krassowski:
bot please update snapshots

packages/notebook/test/history.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        ipySessionContext.shutdown()
+      ]);
+    });
+
+    describe('#history', () => {
+      it('should emit when Markdown and code cells are run', async () => {
+        for (let i = 0; i < 12; i++) {
+          let cell = widget.widgets[i] as CodeCell;
+          let source = 'print(this is input ${i})';
+          cell.model.sharedModel.setSource(source);
```
krassowski, line 86: This test is failing here, it looks like the cells need to be explicitly added to the notebook.

```
@jupyterlab/notebook › NotebookActions › #history › should emit when Markdown and code cells are run

TypeError: Cannot read properties of undefined (reading 'model')

    at Object.<anonymous> (test/history.spec.ts:86:16)
````

File: galata/test/documentation/general.test.ts-snapshots/jupyterlab-documentation-linux.png
Comment (krassowski) [inline galata/test/documentation/general.test.ts-snapshots/jupyterlab-documentation-linux.png:1]: This one appears unrelated and can be reverted to reduce noise.

File: packages/notebook/src/widget.ts (hunk 1 above)
Comment (andrewfulton9) [reply packages/notebook/src/widget.ts:1049]: Maybe `kernelHistory` or `executedInputHistory`?

Comment (krassowski) [reply packages/notebook/src/widget.ts:1049]: `kernelHistory` sounds good to me

File: packages/notebook/src/history.ts [hunk 19]
```diff
@@ -0,0 +1,420 @@
...
+
+  /**
+   * Translator to be used for warnings
+   */
+  readonly translator: ITranslator;
+
+  /**
+   * The number of history items to request.
+   */
+  private toRequest: number;
```
krassowski, line 103: Nitpick: elsewhere we use underscore prefix for private properties (though not strictly required).

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+
+  /**
+   * Kernel session number for filtering
+   */
+  get kernelSession(): string {
+    return this._kernelSession;
+  }
+
+  /**
+   * Get whether the console history manager is disposed.
```
krassowski, line 149:
```suggestion
   * Get whether the notebook history manager is disposed.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   */
+  get isDisposed(): boolean {
+    return this._isDisposed;
+  }
+
+  /**
+   * Dispose of the resources held by the console history manager.
```
krassowski, line 156:
```suggestion
   * Dispose of the resources held by the notebook history manager.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+      this.editor = activeCell.editor;
+      this._placeholder = this._editor?.model.sharedModel.getSource() || '';
+      // Filter the history with the placeholder string.
+      this.setFilter(this._placeholder);
+      this._cursor = this._filtered.length - 1;
+    }
+  }
+
+  /**
+   * Get the previous item in the console history.
```
krassowski, line 182:
```suggestion
   * Get the previous item in the notebook history.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   *
+   * @param activeCell - The currently selected Cell in the notebook.
+   *
+   * @returns A Promise for console command text or `undefined` if unavailable.
```
krassowski, line 186:
```suggestion
   * @returns A Promise resolving to the historical cell content text.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+    if (this._cursor < 0) {
+      await this.fetchBatch();
+    }
+    this._cursor = Math.max(0, this._cursor);
+    const content = this._filtered[this._cursor];
+    return content;
+  }
+
+  /**
+   * Get the next item in the console history.
```
krassowski, line 200:
```suggestion
   * Get the next item in the notebook history.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   *
+   * @param activeCell - The currently selected Cell in the notebook.
+   *
+   * @returns A Promise for console command text or `undefined` if unavailable.
```
krassowski, line 204:
```suggestion
   * @returns A Promise resolving to the historical cell content text.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   */
+  async back(activeCell: Cell): Promise<string> {
+    await this.checkSession(activeCell);
+    --this._cursor;
+    if (this._cursor < 0) {
+      await this.fetchBatch();
+    }
+    this._cursor = Math.max(0, this._cursor);
+    const content = this._filtered[this._cursor];
+    return content;
```
krassowski, line 196: The typing suggests that it will never return undefined as it will stop at the oldest entry. But what happens if there is no history yet? In a very fresh installation, `this._filtered[0]` would resolve to `undefined`, right? Will the downstream error out due to incorrect typing? I would suggest updating the return type to `Promise<string | undefined>` and making any changes that may be needed in the calling functions accordingly.

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   * @param activeCell - The currently selected Cell in the notebook.
+   *
+   * @returns A Promise for console command text or `undefined` if unavailable.
+   */
+  async forward(activeCell: Cell): Promise<string> {
+    await this.checkSession(activeCell);
+    ++this._cursor;
+    this._cursor = Math.min(this._filtered.length - 1, this._cursor);
+    const content = this._filtered[this._cursor];
+    return content;
```
krassowski, line 211: Again, what happens if history is empty?

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  /**
+   * Get the next item in the console history.
```
krassowski, line 215:
```suggestion
   * Update the editor of the cell with provided text content.
```

packages/notebook/src/history.ts
```diff
@@ -0,0 +1,420 @@
...
+   *
+   * @param activeCell - The currently selected Cell in the notebook.
+   * @param update - the result from back or forward
```
krassowski, line 218: While `update` is technically correct, it would be more intuitive for me if it was named `content`

docs/source/user/notebook.rst
```diff
@@ -151,6 +151,10 @@ and select “New Console for Notebook”:
      <iframe src="https://www.youtube-nocookie.com/embed/eq1l7DBngQQ?rel=0&amp;showinfo=0" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
   </div>
 
+.. _kernel_history:
+
+You can iterate through the kernel history in a document cell using ``Alt Up-Arrow`` and ``Alt Down-Arrow``. To use this feature, enable it in the notebook settings.
```
krassowski, line 156:
```suggestion
You can iterate through the kernel history in a document cell using ``Alt Up-Arrow`` and ``Alt Down-Arrow``. To use this feature, enable kernel history access in the notebook settings.
```

packages/notebook-extension/schema/tracker.json
```diff
@@ -801,6 +811,12 @@
       "description": "'defer': Improve loading time - Wait for idle CPU cycles to attach out of viewport cells - 'full': Best performance with side effects - Attach to the DOM only cells in viewport - 'none': Worst performance without side effects - Attach all cells to the viewport",
       "enum": ["defer", "full", "none"],
       "default": "defer"
+    },
+    "accessKernelHistory": {
+      "title": "Enable kernel history access.",
```
krassowski, line 816: Simpler?

```suggestion
      "title": "Kernel history access",
```

krassowski (commented):
This is almost there, just minor changes, mostly updating documentation strings + one issue with translation and narrowing the public API.

packages/notebook/src/history.ts
andrewfulton9: The source of the active cell is added to `this._filtered` in `setFilter` and it isn't possible to access `_filtered` before `setFilter` is run so it shouldn't be possible. I can update it just in case though.

krassowski: In that case no need to change code, but adding a comment would be helpful.

krassowski: Or ideally a unit test case

packages/notebook/src/widget.ts
```diff
@@ -234,6 +235,7 @@ export class StaticNotebook extends WindowedList {
     this._updateNotebookConfig();
     this._mimetypeService = options.mimeTypeService;
     this.renderingLayout = options.notebookConfig?.renderingLayout;
+    this.kernelHistory = options.kernelHistory;
```
krassowski, line 238: Currently this PR the integration test fails because the factory is still passing `history` rather than `kernelHistory`, see the other two comments.

krassowski (approved):
Thank you @andrewfulton9!

PR #15065, opened 2023-09-02
Completer: switch renderer & model on update, update on providers change

### LINKED ISSUE #15017: Completer model and renderer are not updated on completer update
## Description

In LSP extension I observe that our custom model is only attached in newly created files and notebooks but not in the files and notebooks that were already open on load. Switching tabs back and forth does not fix it - it remains like this until file is closed and re-opend.

## Reproduce

This is because the completer handler update sequence mismatches the generation sequence. Compare:

https://github.com/jupyterlab/jupyterlab/blob/ac8f9e2f47b271785127757541a6fd8aa10c4d31/packages/completer/src/manager.ts#L207-L234

with

https://github.com/jupyterlab/jupyterlab/blob/ac8f9e2f47b271785127757541a6fd8aa10c4d31/packages/completer/src/manager.ts#L116-L138

in the update sequence the `modelFactor` and provider `renderer` is not considered. However, the updated providers will be injected via `generateReconciliator()` which leads to inconsistent state (completions from a provider requiring a specific renderer will be rendered with the default one, similarly with model).

## Expected behavior

Two things:
- when a new provider is added, the manager should update the current completer (possibly unless the completer is currently displayed)
- the renderer and model should be updated in `updateCompleter()` routine

## Context

- JupyterLab version: 4.0.05

krassowski:
## References

Fixes #15017

## Code changes

- updates model and renderer
- updates both handler and completer attributes to ensure they are up-to-date if providers change
- replaces tests testing private `_generateHandler()` with tests for public `updateCompleter()` expanding them to cover the actual update of completer attributes when providers change
- suppresses spurious warning in *should not register a provider twice* which is now part of the test itself

## User-facing changes

Completer renderer and model will correctly reflect the providers hierarchy.

None? The `Completer` class gets `renderer` method but this is implementation detail which will not break the existing behaviour if missing on runtime.


packages/completer/src/manager.ts
```diff
@@ -119,17 +123,41 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     const { widget, editor } = newCompleterContext;
     const id = widget.id;
     const handler = this._panelHandlers.get(id);
+
+    const firstProvider = [...this._activeProviders][0];
```
trungleduc, line 127: Why do we need the spread operator here?

packages/completer/src/manager.ts
```diff
@@ -119,17 +123,41 @@ export class CompletionProviderManager implements ICompletionProviderManager {
...
       // Create a new handler.
-      const handler = await this.generateHandler(newCompleterContext);
+      const handler = await this.generateHandler(
+        newCompleterContext,
+        model,
+        renderer
+      );
       this._panelHandlers.set(widget.id, handler);
       widget.disposed.connect(old => {
         this.disposeHandler(old.id, handler);
```
trungleduc, line 151: The `old.id` also needs to be removed from the `_mostRecentContext` map.

krassowski: Thank you, this was addressed in https://github.com/jupyterlab/jupyterlab/pull/15065/commits/ecde7776852bc77764d1239ea17e82e6d4797983.

trungleduc (approved):

trungleduc:
Thanks @krassowski!

krassowski:

PR #15066, opened 2023-09-02
Fix `_updateConstraints` missing `protected`/`private` classifier.

krassowski:
## References

Follow up to #13663 

## Code changes

It was meant to be private - simple typo in a previous PR. To avoid breakages in case if anyone has adopted it in a subclass I am marking it as `protected`. If anyone was using it as a public method this was an error - it should not be used like that.

## User-facing changes

None

See code changes.

### DIFF
--- a/packages/completer/src/widget.ts
+++ b/packages/completer/src/widget.ts
@@ -72,7 +72,7 @@ export class Completer extends Widget {
   /**
    * Cache style constraints from CSS.
    */
-  _updateConstraints() {
+  protected _updateConstraints() {
     const tempNode = document.createElement('div');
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

Maybe it would also be fine to declare it as `private`, depending on when this method was added? If it was introduced in 4.0, maybe downstreams have not had the chance to depend on it yet.

krassowski:
I am on the fence here, but since this is used in another protected method, `onUpdateRequest` I think that by extension it makes sense to allow access to `_updateConstraints` as protected.

krassowski:
It was added in 4.0.

krassowski:

PR #15068, opened 2023-09-02
License header job fix: ignore text files, update format

krassowski:
## References

Closes #15067

## Code changes

Ignores text files in galata snapshots (`txt` is no longer automatically ignored since https://github.com/apache/skywalking-eyes/pull/141).

## User-facing changes

None

## Backwards-incompatible changes

None

Issue #15067, opened 2023-09-02
`Fix License Headers` job on CI is failing


krassowski:
## Description

Since #15054 the `Fix License Headers` job is failing

## Reproduce

See https://github.com/jupyterlab/jupyterlab/actions/runs/6060071826/job/16443852957

## Expected behavior

It does not fail

## Context

None

PR #15072, opened 2023-09-04
Pin skywalking-eyes to patched version detecting license headers after `|`

krassowski:
## References

This should help with  https://github.com/jupyterlab/jupyterlab/issues/15067 and #15068

Note: the check is expected to keep failing; it will be fixed in #15068.

## Code changes

Pin to include fix from https://github.com/apache/skywalking-eyes/pull/168 as recommended in https://github.com/apache/skywalking/issues/11302#issuecomment-1705088577

## User-facing changes

None

None


krassowski:
Moving forward with this to fix CI.

PR #15080, opened 2023-09-05
Fix text wrapping in the search box

### LINKED ISSUE #15049: Search box wraps text but it should grow
## Description

The search box wraps long text, which is undesirable because it is very confusing when multi-line search is used.

## Reproduce

Search for:
```
this_is_a_very_very_very_very_long_variable
this_is_a_new_line
```

repro

## Expected behavior

1. The search input grows up until reaching the full width of the container
2. Once full width is reached, it overflows
3. It never wraps

## Context

- JupyterLab version: 4.0.5

eliaslma:
## References

A step towards #15049 

The search box wraps long text, which is undesirable because it is very confusing when multi-line search is used.

264590626-fc8ad7ee-e237-4e4d-8852-ed74c9fb1815

## Code changes

Added CSS property to the input that handles white spaces following [MDN Web Docs.](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)

## User-facing changes

[Screencast from 05-09-2023 18:11:27.webm](https://github.com/jupyterlab/jupyterlab/assets/70176310/8096f2b8-ef8f-42b3-9881-c0a6dbee33bb)

## Context

Browser: Firefox
JupyterLab version: 4.1.0a1


krassowski:
Thank you for opening a PR! This helps, and should be included, though is still not ideal because:
- the overlay scrollbar makes it hard to click on a specific letter (instead, if clicking on the lower half it scrolls over there) - this is browser dependent behaviour
- the longer text is not visible even if the box could be grown to accommodate the longer text.

Would you be open to also implementing the logic changes for growing the width of the textarea suggested in https://github.com/jupyterlab/jupyterlab/issues/15049#issuecomment-1700630860?

krassowski (approved):
Lets get this in as an incremental improvement

krassowski:

Issue #14947, opened 2023-08-10
Flaky UI tests


jtpio:
## Description

The `Visual Regression Tests` and `Visual Regression Documentation` CI checks are often flaky, which makes the CI checks appear as failing.

## Reproduce

Example run for https://github.com/jupyterlab/jupyterlab/pull/14944: https://github.com/jupyterlab/jupyterlab/actions/runs/5817952284/job/15773457526?pr=14944

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/e59cfdf5-0345-4786-a953-2dd5f14c79e4)

## Expected behavior

The check should not fail that often, so maintainers can review and merge PRs with more confidence.

## Context

Flaky tests lead to maintenance burden and CI fatigue on maintainers. One has to download the Playwright repo and carefully check the errors to assess whether the failure is related to the change.

Also other maintainers might not feel confident merging PRs if one of the checks is failing (which is understandable).

jtpio:
> Example run for #14944: https://github.com/jupyterlab/jupyterlab/actions/runs/5817952284/job/15773457526?pr=14944

In this particular run the failure is caused by a (unrelated) snapshot mismatch:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/113518a7-e4f7-4127-bdae-4b25f0714b5e)

jtpio:
Here is another example of a failing UI test on the latest commit on `main` at the moment: 7b1effc02bb11e237f98bdf3c888862bc183e15d

[galata-debugger-failure.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/10e9f16e-3543-4a8f-9479-e7f8f70574ae)

In this case it looks like the `add` function is not properly added to the notebook, which seems to indicate this is a real issue (or race condition) in the galata tests.

jtpio:
> In this case it looks like the `add` function is not properly added to the notebook, which seems to indicate this is a real issue (or race condition) in the galata tests.

Looks like this code does not do what it is supposed to do:

https://github.com/jupyterlab/jupyterlab/blob/7b1effc02bb11e237f98bdf3c888862bc183e15d/galata/test/jupyterlab/debugger.test.ts#L320-L324

krassowski:
Interestingly, for a split-second the notebook area has no cells at all.

investigating in https://github.com/krassowski/jupyterlab/pull/66:
- no uncaught errors nor relevant messages in console
[...]

PR #15087, opened 2023-09-07
Enable Playwright trace

jtpio:
## References

The `trace` option was added as a comment in https://github.com/jupyterlab/jupyterlab/pull/13909, but https://github.com/jupyterlab/jupyterlab/pull/13909 does not mention why it was left as a comment.

Playwright docs: https://playwright.dev/docs/trace-viewer

## Code changes

Enable `trace` so it's easier to debug failing UI tests on CI.

## User-facing changes

None

None

### DIFF
--- a/galata/src/playwright-config.ts
+++ b/galata/src/playwright-config.ts
@@ -20,7 +20,7 @@ module.exports = {
     viewport: { width: 1024, height: 768 },
 
     // Artifacts
-    // trace: 'on-first-retry',
+    trace: 'on-first-retry',
     video: 'retain-on-failure'
   }
 } as PlaywrightTestConfig;


jtpio:
Since the UI tests are failing on this PR (https://github.com/jupyterlab/jupyterlab/issues/14947) it's a good occasion to show what it looks like when opening the Playwright report:


https://github.com/jupyterlab/jupyterlab/assets/591645/f0a384c4-9d6f-4240-8a6d-74d1257cd61b

jtpio:
One downside is that it might increase the size of the artifacts when many tests are failing:

image

krassowski:
30 MB is not that bad, I think it is worth it.

I was also considering whether to enable console log forwarding as in: https://github.com/krassowski/jupyterlab/pull/66/commits/8e5cb06e6d3508b443b29cd2ab5b0823f91cf03c, which has a downside of making the progress a bit more noisy but then we probably should take care not to spam with logs anyways so maybe this is a feature not an issue - thoughts?

fcollonval:
> 30 MB is not that bad, I think it is worth it.

:+1:

> I was also considering whether to enable console log forwarding as in: [krassowski@8e5cb06](https://github.com/krassowski/jupyterlab/commit/8e5cb06e6d3508b443b29cd2ab5b0823f91cf03c), which has a downside of making the progress a bit more noisy but then we probably should take care not to spam with logs anyways so maybe this is a feature not an issue - thoughts?

We could add a configurable fixture for that to get the console to be forwarded only on retry on the CI for example. But that could be changed locally for debug.

fcollonval (approved):
Thanks @jtpio

jtpio:
> 30 MB is not that bad, I think it is worth it.

Right for this run it's fine. But I think it will be proportional to the number of failing tests.

jtpio:

Issue #15016, opened 2023-08-24
Completion providers list is not refreshed dynamically


krassowski:
## Description

The list of applicable completion providers is materialized at the time when `updateCompleter` is called. If a provider implements `isApplicable` method dynamically, e.g. it returns true if an LSP adapter is found (or if connection to LSP server is open), it may initially return false while the setup is ongoing. Problematically, this will not be updated until user swtiches back and forth between another editor of the same type triggering `updateCompleter()` code path.

## Reproduce

Relevant code:

https://github.com/jupyterlab/jupyterlab/blob/ac8f9e2f47b271785127757541a6fd8aa10c4d31/packages/completer/src/manager.ts#L176-L188

## Expected behavior

We could move the check to `ProviderReconciliator` which could call it each time (e.g. in a getter). My assumption is that `isApplicable` method is always (or should be) cheap.

Any other ideas?

## Context

- JupyterLab version: 4.0.5

fcollonval:
:+1: on the proposed solution

PR #15094, opened 2023-09-07
Check applicable provider in ProviderReconciliator

trungleduc:
## References
Closes https://github.com/jupyterlab/jupyterlab/issues/15016

## Code changes

- The list of applicable providers is checked on every `fetch` call instead of on the `ProviderReconciliator` creation time.

## User-facing changes

N/A

- `IProviderReconciliator` interface is changed, `shouldShowContinuousHint` returns `Promise<boolean>` instead of `boolean`

### DIFF
--- a/packages/completer/src/reconciliator.ts
+++ b/packages/completer/src/reconciliator.ts
@@ -24,6 +24,11 @@ export class ProviderReconciliator implements IProviderReconciliator {
     this._timeout = options.timeout;
   }
 
+  /**
+   * Check for the providers which are applicable with the current context
+   *
+   * @return  List of applicable providers
+   */
   async applicableProviders(): Promise<Array<ICompletionProvider>> {
     const isApplicablePromises = this._providers.map(p =>
       p.isApplicable(this._context)
@@ -39,7 +44,7 @@ export class ProviderReconciliator implements IProviderReconciliator {
    *
    * @param {CompletionHandler.IRequest} request - The completion request.
    */
-  public async fetch(
+  async fetch(
     request: CompletionHandler.IRequest,
     trigger?: CompletionTriggerKind
   ): Promise<CompletionHandler.ICompletionItemsReply | null> {
--- a/packages/completer/src/reconciliator.ts
+++ b/packages/completer/src/reconciliator.ts
@@ -24,6 +24,14 @@ export class ProviderReconciliator implements IProviderReconciliator {
     this._timeout = options.timeout;
   }
 
+  async applicableProviders(): Promise<Array<ICompletionProvider>> {
+    const isApplicablePromises = this._providers.map(p =>
+      p.isApplicable(this._context)
+    );
+    const applicableProviders = await Promise.all(isApplicablePromises);
+    return this._providers.filter((_, idx) => applicableProviders[idx]);
+  }
+
   /**
    * Fetch response from multiple providers, If a provider can not return
    * the response for a completer request before timeout,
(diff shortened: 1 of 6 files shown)


packages/metapackage/test/completer/handler.spec.ts
```diff
@@ -47,7 +47,7 @@ class TestCompleterModel extends CompleterModel {
 class TestCompletionHandler extends CompletionHandler {
   methods: string[] = [];
 
-  onTextChanged(str: ISharedText, changed: SourceChange): void {
+  async onTextChanged(str: ISharedText, changed: SourceChange): Promise<void> {
```
trungleduc, line 50: @krassowski do you know why the completer tests are in the `metapackage` and not in the `completer` package?

krassowski: These were moved there in #12495 by @fcollonval as far as I understand to remove a dependency cycle, but I don't know the details and whether this is still needed.

packages/completer/src/tokens.ts
```diff
@@ -206,5 +206,5 @@ export interface IProviderReconciliator {
   shouldShowContinuousHint(
     completerIsVisible: boolean,
     changed: SourceChange
-  ): boolean;
+  ): Promise<boolean>;
```
krassowski, line 209: I think that with little effort we could have a union of a promise or boolean for better backward compatibility. I find it unlikely that downstreams modify reconciliator since it was just recently released and is not the documented as an extension point, but if we are to improve extension ecosystem experience it probably starts with such little decisions.

krassowski (commented):
Thank you, looks good overall. I left two API questions.

packages/completer/src/tokens.ts
trungleduc: Could you give me a hint about the implementation?

packages/completer/src/reconciliator.ts
```diff
@@ -24,21 +24,35 @@ export class ProviderReconciliator implements IProviderReconciliator {
     this._timeout = options.timeout;
   }
 
+  /**
+   * Check for the providers which are applicable with the current context
+   *
+   * @return  List of applicable providers
+   */
+  async applicableProviders(): Promise<Array<ICompletionProvider>> {
```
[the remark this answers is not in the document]
trungleduc: Thanks, updated in https://github.com/jupyterlab/jupyterlab/pull/15094/commits/d32a77fe8f1e0efeb81371d9d4ca0c84449e72fa

packages/completer/src/tokens.ts
krassowski: I was thinking that it would be just changing the return type to union of promise and a boolean, without changed to implementation, but if that's not that simple then maybe it's not worth it after all...

trungleduc: Yep, I think it's not that simple, and returning `boolean | Promise<boolean>` would confuse the downstream usage.

krassowski: To be fair I think this would not be used downstream much, but then this argument is also why I don't think we need to change it :) Marking as resolved, thank you for engaging in brainstorming this one!

krassowski (approved):
Thank you @trungleduc!

packages/metapackage/test/completer/handler.spec.ts
fcollonval: Yes lerna does not distinguish between dep and dev dep when building the packages dep graph.

And for tests, they were chicken/egg issue; like notebook depends on the completer but the completer tests depends on notebook (unit tests should have used mocked documents for interface - but better integration tests than nothing). Hence the need to move the tests to an upper package.

fcollonval (approved):
Thanks @trungleduc

krassowski:

Issue #15091, opened 2023-09-07
Small styling issue Property Inspector Dark mode


Rmarieta:
## Description

When opening the drop down menus in the notebook Property Inspector in the sidebar, I observe that in dark mode, the color of the dropdown options stay white as in the picture below : 
![image](https://github.com/jupyterlab/jupyterlab/assets/33026272/2bbb35d6-d175-4680-8fe3-fa1d2648a7d8)

This requires a very minor fix by adding some CSS to the `property-inspector` package so I can make a PR to quickly fix it.

## Reproduce

Nothing too fancy here : 
1. Start a JupyterLab server (I tried with JupyterLab 4.0.1 and 4.0.5 to check it wasn't fixed)
2. Open a notebook
3. Go to Property Inspector on the right sidebar
4. Open either of 'Editable' or 'Slide Type' dropdowns

The color of the dropdown worked correctly on JupyterLab 3.6

## Expected behavior

The color of the dropdown options to be like this : 
![image](https://github.com/jupyterlab/jupyterlab/assets/33026272/c15f1a0b-17f6-4b5d-9a3d-a7364996e752)

## Context

- Operating System and version: Windows 11
- Browser and version: Chrome 116
- JupyterLab version: 4.0.1 or 4.0.5

krassowski:
Closing as fixed by #15098

PR #15098, opened 2023-09-08
Fix background-color in `<select>` elements in dark theme

Rmarieta:
## References

Issue #15091

## Code changes

Small CSS fix

### DIFF
--- a/packages/property-inspector/style/base.css
+++ b/packages/property-inspector/style/base.css
@@ -13,10 +13,6 @@
   flex-grow: 1;
 }
 
-.jp-PropertyInspector-content option {
-  background-color: var(--jp-layout-color1);
-}
-
 .jp-PropertyInspector-placeholder {
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


krassowski (changes requested):
Thank you for opening a PR @Rmarieta! I can confirm that this fixes the issue in the property inspector. However, it looks that the bug affects not only the property inspector, but also other places where the form editor is used, for example settings:

image

Would it be better to instead add:

```css
.jp-inputFieldWrapper > select > option {
  background-color: var(--jp-layout-color1);
}
```

to a file where other `.jp-inputFieldWrapper` styles are?

krassowski:
I pushed a commit generalising the fix to all `<select>`s under form wrappers. I also narrowed down related selectors. I did not add a visual test because playwright does not support capturing `<select>` dropdowns in open state as of now (https://github.com/microsoft/playwright/issues/12229, which is understandable if you know how browsers implement them).

krassowski (approved):

krassowski:

PR #15101, opened 2023-09-09
Fix return type of `getMimeTypeByLanguage()`

### LINKED ISSUE #14627: Code cell input coloring with ijavascript Kernel not working
## Description

I have been trying to migrate my previous extension to JupyterLab 4.0.1 from JupyterLab 3.x, and I observe that the code cell input coloring is gone when using the JavaScript kernel (`ijavascript`) as compared to using the default Python one.

image

In addition to this, I also observe an error in my browser developer tools console when I switch to that kernel (`JavaScript (Node.js)`) from the `Select Kernel` window and press `Select` : 
```
TypeError: Cannot read properties of undefined (reading 'widget')
    at isReady (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1054549)
    at jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:872986
    at b (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:873080)
    at P.onKernelChanged (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1053706)
    at m (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1646566)
    at Object.l [as emit] (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1646242)
    at a.emit (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1644079)
    at b._onKernelChanged (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:132174)
    at m (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1646566)
    at Object.l [as emit] (jlab_core.b473ae48d19e9025bb00.js?v=b473ae48d19e9025bb00:1:1646242)
```

This `reading 'widget'` error also occurs if I switch from two default JupyterLab kernels, for example if I change Kernel to another notebook's kernel with the option `Use a Kernel From Preferred Session`.

I am aware that this could potentially be a problem with the JavaScript Kernel in itself, but since it was working fine in JupyterLab 3, it might still be worth reporting this issue to have a look at what is going wrong.

## Reproduce

1. Build the JupyterLab environment and install the javascript kernel : 
```
conda create -n tmp --override-channels --strict-channel-priority -c conda-forge -c nodefaults jupyterlab=4
conda activate tmp
npm install -g ijavascript
ijsinstall
```
2. Create a notebook and add a code cell with JavaScript code such as : 
```
const func = () => {
    console.log("Hello");
}
func()
``` 
3. Swith to the JavaScript Kernel
4. Wait and observe that the code cell input does not pick up the right colors and that the error mentioned in the Description is printed to the console. 

## Expected behavior

I would expect the code to display JavaScript code with colors following the JavaScript syntax but it's displayed in black and white instead.

## Context

- Browser and version: Chrome, version 112.0.5615.139
- JupyterLab version: 4.0.1

### LINKED ISSUE #15100: Incorrect typing of `getMimeTypeByLanguage()`
## Description

Differently to typing, `getMimeTypeByLanguage` can return a list of strings rather than a string. This is the case when asked about `javascript`. This is because of the type cast: `(mode.mime as string)` as seen below:

https://github.com/jupyterlab/jupyterlab/blob/b656a4fba951dabaa66062c928d5416df63673e1/packages/codemirror/src/mimetype.ts#L20-L32

Screenshot from 2023-09-09 02-14-29

This has lead to a bug in jupyterlab-lsp pre-releases because then `editor.model.mimeType = mode;` became invalid.

## Expected behavior

Typings are correct

## Context

- Browser and version: Chrome
- JupyterLab version: 4.0.5

krassowski:
## References

- Fixes #15100 
- Fixes #14627

## Code changes

`getMimeTypeByLanguage` was incorrectly annotated as always returning string even though it could return a list of strings.

This PR remove the problematic type cast and always returns a string, docuemtning and testing this behaviour.
## User-facing changes

| Before | After |
|--------|--------|
| Screenshot from 2023-09-09 11-16-26 | Screenshot from 2023-09-09 11-16-35 |

None.

### DIFF
--- a/examples/cell/src/index.ts
+++ b/examples/cell/src/index.ts
@@ -139,7 +139,11 @@ function main(): void {
   sessionContext.kernelChanged.connect(() => {
     void sessionContext.session?.kernel?.info.then(info => {
       const lang = info.language_info;
-      const mimeType = mimeService.getMimeTypeByLanguage(lang);
+      const mimeTypes = mimeService.getMimeTypeByLanguage(lang);
+      const mimeType =
+        Array.isArray(mimeTypes) && mimeTypes.length !== 0
+          ? mimeTypes[0]
+          : mimeTypes;
       cellWidget.model.mimeType = mimeType;
     });
   });
--- a/packages/console/src/widget.ts
+++ b/packages/console/src/widget.ts
@@ -730,7 +730,11 @@ export class CodeConsole extends Widget {
     }
     this._banner!.model.sharedModel.setSource(info.banner);
     const lang = info.language_info as nbformat.ILanguageInfoMetadata;
-    this._mimetype = this._mimeTypeService.getMimeTypeByLanguage(lang);
+    const mimetypes = this._mimeTypeService.getMimeTypeByLanguage(lang);
+    this._mimetype =
+      Array.isArray(mimetypes) && mimetypes.length !== 0
+        ? mimetypes[0]
+        : mimetypes;
     if (this.promptCell) {
       this.promptCell.model.mimeType = this._mimetype;
     }
--- a/packages/console/src/widget.ts
+++ b/packages/console/src/widget.ts
@@ -730,11 +730,7 @@ export class CodeConsole extends Widget {
     }
     this._banner!.model.sharedModel.setSource(info.banner);
     const lang = info.language_info as nbformat.ILanguageInfoMetadata;
-    const mimetypes = this._mimeTypeService.getMimeTypeByLanguage(lang);
-    this._mimetype =
-      Array.isArray(mimetypes) && mimetypes.length !== 0
-        ? mimetypes[0]
-        : mimetypes;
+    this._mimetype = this._mimeTypeService.getMimeTypeByLanguage(lang);
     if (this.promptCell) {
       this.promptCell.model.mimeType = this._mimetype;
     }
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -748,7 +748,11 @@ export class StaticNotebook extends WindowedList {
     if (!info) {
       return;
     }
-    this._mimetype = this._mimetypeService.getMimeTypeByLanguage(info);
+    const mimetypes = this._mimetypeService.getMimeTypeByLanguage(info);
+    this._mimetype =
+      Array.isArray(mimetypes) && mimetypes.length !== 0
+        ? mimetypes[0]
+        : mimetypes;
     for (const widget of this.widgets) {
       if (widget.model.type === 'code') {
         widget.model.mimeType = this._mimetype;
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -748,11 +748,7 @@ export class StaticNotebook extends WindowedList {
     if (!info) {
       return;
     }
-    const mimetypes = this._mimetypeService.getMimeTypeByLanguage(info);
-    this._mimetype =
-      Array.isArray(mimetypes) && mimetypes.length !== 0
-        ? mimetypes[0]
-        : mimetypes;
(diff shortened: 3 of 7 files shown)


examples/cell/src/index.ts
```diff
@@ -139,7 +139,11 @@ function main(): void {
   sessionContext.kernelChanged.connect(() => {
     void sessionContext.session?.kernel?.info.then(info => {
       const lang = info.language_info;
-      const mimeType = mimeService.getMimeTypeByLanguage(lang);
+      const mimeTypes = mimeService.getMimeTypeByLanguage(lang);
+      const mimeType =
+        Array.isArray(mimeTypes) && mimeTypes.length !== 0
+          ? mimeTypes[0]
+          : mimeTypes;
```
fcollonval, line 146: Do we need a better fallback if `mimeTypes === []`?

The current code results in:

mimeTypes | mimeType 
--- | ---
 'type1' |  'type1'
['type1', ...] | 'type1'
[] | []

But `CodeEditor.Model.mimeType` should be a string. So should we switch to 

```suggestion
        Array.isArray(mimeTypes)
          ? (mimeTypes[0] ?? IEditorMimeTypeService.defaultMimeType)
          : mimeTypes;
```

packages/console/src/widget.ts
```diff
@@ -730,7 +730,11 @@ export class CodeConsole extends Widget {
     }
     this._banner!.model.sharedModel.setSource(info.banner);
     const lang = info.language_info as nbformat.ILanguageInfoMetadata;
-    this._mimetype = this._mimeTypeService.getMimeTypeByLanguage(lang);
+    const mimetypes = this._mimeTypeService.getMimeTypeByLanguage(lang);
+    this._mimetype =
+      Array.isArray(mimetypes) && mimetypes.length !== 0
```
fcollonval, line 735: Same question

packages/notebook/src/widget.ts
```diff
@@ -748,7 +748,11 @@ export class StaticNotebook extends WindowedList {
     if (!info) {
       return;
     }
-    this._mimetype = this._mimetypeService.getMimeTypeByLanguage(info);
+    const mimetypes = this._mimetypeService.getMimeTypeByLanguage(info);
+    this._mimetype =
+      Array.isArray(mimetypes) && mimetypes.length !== 0
```
fcollonval, line 753: Here too

fcollonval (commented):
Thanks @krassowski 

I have one question regarding the fallback value in case of empty array.

examples/cell/src/index.ts
krassowski: What if instead we ensured that `getMimeTypeByLanguage()` never returns an empty array?

fcollonval: Yes it will ease usage avoiding testing for each call

krassowski:
I'm having second thoughts here because the sister function, `getMimeTypeByFilePath` just returns a string:

https://github.com/jupyterlab/jupyterlab/blob/9be8052d846290e22a0e51a5d3e3dec43ba9efdc/packages/codemirror/src/mimetype.ts#L40-L52

And plausibly the downstream could just use `IEditorLanguageRegistry` to get a full list of mime types. This would also also prevent breaking changes.

fcollonval (approved):
Thanks @krassowski 

Indeed it makes sense to align with `getMimeTypeByFilePath`.

krassowski:

PR #15105, opened 2023-09-10
Remove `unusedDocuments`, fix culling of foreign documents

krassowski:
## References

Follow up to https://github.com/jupyterlab/jupyterlab/pull/12534
Developed in response to a failing test caught on https://github.com/jupyter-lsp/jupyterlab-lsp/pull/978

## Code changes

- The unused protected property `unusedDocuments` is removed (it was never written to since #12534, it was functioning properly with different logic in lsp extension)
- The logic to cull the unused foreign documents is re-implemented in `closeExpiredDocuments()` without need to keep state on the `VirtualDocument` instance

## User-facing changes

When user creates a markdown or raw cell and then deletes it (without other cells of this type present), the connection to LSP server is not held indefinitely but culled after six further document changes.

Removal of the protected `unusedDocuments` method is notable but we are not advertising LSP as stable yet, and it would be useless in the current state anyways so it should be fine.

### DIFF
--- a/packages/lsp/src/virtual/document.ts
+++ b/packages/lsp/src/virtual/document.ts
@@ -806,14 +806,12 @@ export class VirtualDocument implements IDisposable {
     const usedDocuments = new Set<VirtualDocument>();
     for (const line of this.sourceLines.values()) {
       for (const block of line.foreignDocumentsMap.values()) {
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @krassowski

krassowski:

Issue #13343, opened 2022-10-31
Scrollbar jumps back in a virtualized notebook on scrolling


suganya-sk:
## Description

In a virtualized notebook, on scrolling, the scrollbar jumps back and shrinks as more content is rendered. This is a bit jarring visually and the original scrollbar is not indicative of the actual size of the notebook.

## Reproduce

In a notebook running from Jupyterlab >= 4.0.0a29, 

1. Open a notebook with about a 100 cells - the scrollbar at this point does not reflect the actual size of the notebook (100 cells)
3. Start scrolling down to view cells below the original viewport
4. The scrollbar jumps back and changes in size as more DOM elements are added.

## Expected behavior

Scroll bar moves smoothly during scrolling

## Context

- Browser and version:  Chrome 106, Firefox 106
- JupyterLab version: 4.0.0a29

fcollonval:
Some background on the root cause:

Virtual rendering requires the estimation of content not rendered. This is a very hard problem for notebook as we have no good way to estimate their sizes. The height estimator is very crude for now (some heuristic for rendered markdown cells and cell outputs would definitely be beneficial) as it only take into account the number of lines in the cell sources to get an estimation of the cell height.

> Any solution that would try to get the real content in the DOM to estimate the height of out of view cells will dramatically reduce the benefit of the windowing.

suganya-sk:
Thanks for the quick reply here.

I agree guessing dimensions of content that has not been rendered is a difficult problem to solve and we wouldn't want to do something that foregoes all the benefit virtualization has brought in.

This isn't much of an improvement but can we consider adding the something like this to èstimateWidgetSize` - it already estimates size of a cell by accessing the number of input cells it has. Could we consider adding similar logic for output for code cells and attachments for markdown/raw cells?


```js
const thisCellModel = this.cells[index].model.sharedModel;
const cellType = thisCellModel.cell_type;
let outputHeight = 0;
if  (cell_type == "code") {
    // this would work for text output but is a vague estimate for other kinds
    // can be refined further by differentiating IExecuteResult, IDisplayData and IStream
    const outputLines = thiscellModel
                            .getOutputs()
                            .split('\n').length;
    outputHeight = (
        // we'll have to define DEFAULT_OUTPUT_LINE_HEIGHT
        NotebookViewModel.DEFAULT_OUTPUT_LINE_HEIGHT * outputLines +
        NotebookViewModel.DEFAULT_CELL_MARGIN
    );
[...]

PR #15036, opened 2023-08-29
Add a renderer for windowed lists

afshin:
[WIP]

cc: @gabalafou 

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/14804

## User-facing changes
N/A

## Backwards-incompatible changes
N/A

Issue #15008, opened 2023-08-23
Debugger tests are failing in Visual Regression Documentation job


krassowski:
As seen in https://github.com/jupyterlab/jupyterlab/pull/15007, [here](https://github.com/jupyterlab/jupyterlab/actions/runs/5954002808/job/16149558904?pr=15007)

```
Notice:   1 failed
    [documentation] › test/documentation/debugger.test.ts:98:7 › Debugger › Stop on breakpoint ─────
  4 flaky
    [documentation] › test/documentation/debugger.test.ts:53:7 › Debugger › Set breakpoint ─────────
    [documentation] › test/documentation/debugger.test.ts:218:7 › Debugger › Variables panel ───────
    [documentation] › test/documentation/debugger.test.ts:245:7 › Debugger › Call Stack panel ──────
    [documentation] › test/documentation/debugger.test.ts:276:7 › Debugger › Breakpoints panel ────
```

With a recurring problem around a line not being available in gutters map:

```
    TypeError: Cannot read properties of undefined (reading 'click')

       at src/helpers/notebook.ts:886

      884 |       return false;
      885 |     }
    > 886 |     await gutters[lineNumber].click();
          |                               ^
      887 |     return true;
      888 |   }
      889 |
```

krassowski:
Related to but distinct from #14947

krassowski:
I have not see this one in a while, will close it.

PR #15109, opened 2023-09-12
Scrollbar UI for virtually rendered notebook

afshin:
This PR updates the `WindowedList` class that the notebook widget inherits from to have customizable renderers (which allows for adding `aria` labels to make windowed lists more accessible and allows for customization of the UI).

The renderer in turn is used to add a scrollbar component to replace the native scrollbar that jumps when virtual notebook rendering is enabled.

The scrollbar (and its toolbar button) only appear when "Windowing mode" is set to "full".

## References
- supersedes https://github.com/jupyterlab/jupyterlab/pull/15036
- fixes https://github.com/jupyterlab/jupyterlab/issues/13343

## Code changes
- Updates the `WindowedList` API to allow for custom renderers of windowed list subcomponents
- Adds a virtual scrollbar UI to replace the native scrollbar when virtual rendering is turned on
- Updates the command toolbar button to respect `isEnabled`, `isVisible`, `isToggled`.

## User-facing changes
Adds an optional new scrollbar control for virtually rendered notebooks.

**before**

https://github.com/jupyterlab/jupyterlab/assets/159529/1a46ee2e-785c-44c0-89f6-fe3ffba3e036

**after**

https://github.com/jupyterlab/jupyterlab/assets/159529/f83a575f-cfe5-4395-9221-a961e40fb397

N/A


packages/ui-components/src/components/windowedlist.ts
```diff
@@ -821,6 +854,12 @@ export class WindowedList<
     return this._isScrolling.promise;
   }
 
+  dispose(): void {
+    this._scrollbar.dispose();
+    this.layout.dispose();
```
fcollonval, line 859: Unneeded as it is already taking care in `Widget`: https://github.com/jupyterlab/lumino/blob/a817cb2d05ac264403ad930fb9c038746c5393ff/packages/widgets/src/widget.ts#L74

```suggestion
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -832,6 +871,21 @@ export class WindowedList<
       this._applyNoWindowingStyles();
     }
     this.viewModel.height = this.node.getBoundingClientRect().height;
+
+    if (this._scrollbar.isDisposed) {
+      const { renderer, viewModel } = this;
+      this._scrollbar = new WindowedList.Scrollbar(viewModel, renderer);
```
fcollonval, line 877: If you add a scrollbar here, you will have to listen for its signal - like in the constructor, isn't it?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -917,6 +980,13 @@ export class WindowedList<
    * The default implementation of this handler is a no-op.
    */
   protected onUpdateRequest(msg: Message): void {
+    if (this.scrollbar) {
+      // Hide the native scrollbar.
+      this.node.style.width = 'calc(100% + 25px)';
+      this._scrollbar.update();
+    } else {
+      this.node.style.width = 'calc(100%)';
```
fcollonval, line 988: let's do the math :stuck_out_tongue_winking_eye: 

```suggestion
      this.node.style.width = '100%';
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1260,10 +1331,118 @@ export class WindowedLayout extends PanelLayout {
...
+          content.appendChild(item);
+        }
+      }
+    }
+
+    protected content: HTMLUListElement;
+    protected model: IModel<T>;
+    protected renderer: IRenderer<T>;
+
+    private _evtPointerDown(event: PointerEvent): void {
```
fcollonval, line 1422: What should happen if the user move the pointer when keeping it pressed? Should we listen for pointer motion and/or pointer up too?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1439,16 +1620,52 @@ export namespace WindowedList {
...
+  export interface IRenderer<T = any> {
+    /**
+     * Create the outer, root element of the windowed list.
+     */
+    createOuter(): HTMLElement;
+
+    /**
+     * Create the virtual scrollbar element.
+     */
+    createScrollbar(): HTMLUListElement;
```
fcollonval, line 1658: I would argue it should be an ordered list:

```suggestion
    createScrollbar(): HTMLOListElement;
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -841,6 +895,11 @@ export class WindowedList<
     if (this._viewModel.windowingActive) {
       this._removeListeners();
     }
+
+    if (!this._scrollbar.isDisposed) {
+      this._scrollbar.dispose();
```
fcollonval, line 900: Could you elaborate the need to dispose the scrollbar on before-detach message?

fcollonval (commented):
I have not been able to test it locally - the scrollbar has no item. But it is probably related to my first comment: you need to listen for changes to the itemsList (or to the model state change for name = 'list' | 'count') - and take a decision on whether it implies rerendering all scrollbar items or something smarter.

Otherwise here are some comments and questions.

packages/ui-components/src/components/windowedlist.ts
afshin: You're right. I'm rewriting this bit to try to move the scrollbar into layout after all. I think it's causing too many side-effects.

packages/ui-components/src/components/windowedlist.ts
afshin: I've abandoned the standalone widget implementation and moved back to implementing it within `WindowedList`.

packages/ui-components/src/components/windowedlist.ts
afshin: I am treating this like simple clicks for the initial implementation. I wonder if that is sufficient.

packages/ui-components/src/components/windowedlist.ts
afshin: This was because I was trying to maintain the `Widget` lifecycle myself instead of either:
1. relying on `Layout` to do the work
2. containing the scrollbar within the `WindowedList` class
I tried the first, but it was too messy, the second is really the way to go.

afshin:
I tried three approaches to this functionality:

1. Creating a standalone scrollbar widget whose lifecycle is manually kept in sync with its semantic parent, the windowed list. This turns out to have too many edge cases because Lumino can pull the rug out from under the widget and cause even `dispose()` calls to throw.
2. Adding the scrollbar widget to the `WindowedLayout` and managing its lifecycle this way. The trouble with this is that in *every* place where the `.widgets` are iterated over, you'd have to special-case if it's `instanceof WindowedList.Scrollbar` or some other mechanism to treat it differently than list contents, so it seemed about as bad as option 1.
3. Add the scrollbar markup and content and event handling into the `WindowedList` class. The main downside that made me avoid this solution was having to add an additional tier in the DOM tree to hold the outer container of the list as a sibling of the scrollbar. But this seems much less painful than the above.

---

So the new implementation is almost there, but it has two show stopper bugs, currently.

1. (This bug existed in all three version, I am not sure I understand the cause yet.) The notebook tab has become a black hole. I can switch from one notebook to another. But if I try to switch to a non-notebook tab, it immediately reverts back:

https://github.com/jupyterlab/jupyterlab/assets/159529/14a887a2-19b1-4d9c-a5ea-d8867471cd60

2. Something I have changed, probably in the CSS but I'm not 100% sure yet, has called the `scrollToItem()` method in the windowed list (a method I have not modified), to change the dimensions of the notebook in a way that makes it unusable:


https://github.com/jupyterlab/jupyterlab/assets/159529/cc154e38-cb2f-463d-8a68-203b92392ff9

packages/notebook/src/searchprovider.ts
```diff
@@ -721,7 +721,7 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
       // change, and if selection is getting extended, we do not want to clear
       // highlights just to re-apply them shortly after, which has side effects
       // impacting the functionality and performance.
-      this._delayedActiveCellChangeHandler = setTimeout(() => {
+      this._delayedActiveCellChangeHandler = window.setTimeout(() => {
```
afshin, line 724: This is unrelated to this PR but it was causing the `watch` compiler to fail so I updated it.

packages/notebook/src/widget.ts
```diff
@@ -206,7 +206,9 @@ export class StaticNotebook extends WindowedList {
           (options.notebookConfig?.windowingMode ??
             StaticNotebook.defaultNotebookConfig.windowingMode) === 'full'
       }),
-      layout: new NotebookWindowedLayout()
+      layout: new NotebookWindowedLayout(),
+      renderer: StaticNotebook.defaultRenderer,
+      scrollbar: true
```
afshin, line 211: We will want this behind a user setting, it's set to `true` for now for testers to see and use.

packages/notebook/src/widget.ts
```diff
@@ -1084,6 +1086,33 @@ export namespace StaticNotebook {
...
+    }
+
+    createViewport() {
+      const el = document.createElement('div');
+      el.setAttribute('role', 'feed');
+      el.setAttribute('aria-label', 'Cells');
+      return el;
+    }
+
+    createScrollbarItem(index: number, item: ICellModel): HTMLLIElement {
```
afshin, line 1104: Given that this receives the cell model, the list item it generates can be more interesting than just a number at some point in the future.

afshin (commented):
bot please update snapshots

afshin (commented):
bot please update snapshots

afshin:
bot please update snapshots

afshin (commented):
bot please update snapshots

afshin:
bot please update snapshots

afshin:
bot please update snapshots

afshin:

afshin:
bot please update snapshots

afshin:

afshin:
bot please update snapshots

krassowski:
This looks nice! A first pass with some high-level UX issues below.

1. Something odd happening on resize, fixes itself on click:
    resize-odd
2. In browsers which use overlay scrollbars (Firefox by default, Chrome on specific OSes, JupyterLab with Theme → Theme scrollbars in dark theme) the minimap is hard to click as scrollbar occludes it:
    overlay-scrollbars-new
3. Styling nitpick in Firefox: should there be a margin/padding between the minimap and the cell edge?
    image
4. When I saw the numbers I wanted to click on them. I intuitively expected the active cell to change after click to the target one but it does not. I see that this could be problematic when user has a cell selection that they want to preserve. This is probably a good topic for follow up discussion on what the ideal UX would be here:
    should-cliclking

krassowski:
> This is probably a good topic for follow up discussion on what the ideal UX would be here

(maybe it could flash highlight the cell without making it active?)

galata/src/helpers/notebook.ts
```diff
@@ -883,7 +883,7 @@ export class NotebookHelper {
     if (gutters.length < lineNumber) {
       return false;
     }
-    await gutters[lineNumber - 1].click();
+    await gutters[lineNumber].click();
```
krassowski: I guess it was an attempt at tackling https://github.com/jupyterlab/jupyterlab/issues/15008?

afshin:
bot please update snapshots

galata/src/helpers/notebook.ts
afshin: Yes, but it was incorrect and I have reverted it.

krassowski: I think I have fallen for thinking this would be a fix before too. Just wanted to share the link to the issue tracking it for reference :)

GabrielaVives:
> 3. Styling nitpick in Firefox: should there be a margin/padding between the minimap and the cell edge?
>    image

Yes I agree we could add a margin or a padding of 2px to the cell, it should be enough.


> 4. When I saw the numbers I wanted to click on them. I intuitively expected the active cell to change after click to the target one but it does not. I see that this could be problematic when user has a cell selection that they want to preserve. This is probably a good topic for follow up discussion on what the ideal UX would be here:
>    should-cliclking
>          
>            [
>              
>            
>                should-cliclking
>              ](https://user-images.githubusercontent.com/5832902/274029563-e535199e-2f6e-401a-aa06-0987d6e93e33.gif)
>            
>            
>              
>                
>                  
>                
>                
>                  
>                  
>                
>              
>              [
>                
>                  
>                
>              ](https://user-images.githubusercontent.com/5832902/274029563-e535199e-2f6e-401a-aa06-0987d6e93e33.gif)

We have been investigating what users expect for this behaviour. I think the main problem in this version is that there is only one state for the numbered boxes. If by default the active cell is always represented in a different state in the virtual scrollbar, you could immediately identify it and understand that clicking on another box would change the active cell. Users would not be confused and wouldn't click it if they want to preserve a cell selection. 
However, the intuitive behaviour seems to be changing the active cell but also navigating to it (here it is not obvious but in the cas of a long notebook I think it would be).

afshin:
@krassowski:
> In browsers which use overlay scrollbars (Firefox by default, Chrome on specific OSes, JupyterLab with Theme → Theme scrollbars in dark theme) the minimap is hard to click as scrollbar occludes it

I have been searching for solutions to this, but it seems that in modern macOS, it's not possible to style scrollbars even with the webkit pseudo-selectors. The only cross-platform solution I can think of is to pad the right-hand side and in all browser/OS combinations that allow it, to set `overflow: scroll` and in macOS with auto-hidden scrollbars at least there is suitable whitespace for the scrollbar to occlude.

I don't like this solution but I don't have a better idea.

krassowski:
[CSS Scrollbars Styling Module Level 1](https://drafts.csswg.org/css-scrollbars-1/) support is slowly getting there [in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=891944) so at least `scrollbar-width` should be customizable cross-platform soon, but I am not sure if they will keep allowing the OS to override it.

Could adjusting [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) help here?

krassowski:
[Intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/PkEsMirl2zE?pli=1) for CSS scrollbars in Chromium was just posted, targeting v121 so using `scrollbar-width` soon appears even more realistic possibility.

mlucool:
Taking a look at the binder can we:
1. Indicate the current active cell
2. Clicking on a cell number in the new scroll UI should make that cell active (aside from scrolling it into view)
3. As you scroll using the a mouse (eg. the wheel over the notebook), it feels like there should be some interaction with the scrollbar UI. I'm not sure what it is, but maybe something like if the active cell moves from visible on the scrollbar to out of view, we also scroll new UI to keep the active cell visible?

GabrielaVives:
> 3. As you scroll using the a mouse (eg. the wheel over the notebook), it feels like there should be some interaction with the scrollbar UI. I'm not sure what it is, but maybe something like if the active cell moves from visible on the scrollbar to out of view, we also scroll new UI to keep the active cell visible?

If the scrollbar moves when the user scrolls the notebook, it could be confusing as the scrollbar UI also has its own native scrollbar. For example if the user is working on a cell and scrolls back and forth to the previous output, if the virtual scrollbar also moves back and forth it could be unexpected and visually distracting. When users scroll in an area they expect this area to move (+ the native scrollbar), and probably not another area that has its own native scrollbar. 

When an active cell is in the visible notebook area, we could have visual indications like these:
- text at the top of the scrollbar giving information on how many cells/total cells are visible
- the active cell is highlighted in the virtual scrollbar (TBD what would be the exact UI)


image


If the user scrolls in the notebook and the active cell goes out of the visible area, clicking on the active cell in the scrollbar would immediately bring it back to view. This could be very helpful when exploring a notebook looking for information, you want to be able to easily (and quickly) get back to your active cell to continue your work. 

image


If the user then selects another cell in the notebook, the scrollbar will interact and scroll to the bring the new active cell representation in view.


image

afshin:
bot please update snapshots

afshin:
bot please update snapshots

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -883,6 +944,7 @@ export class WindowedList<
     this.viewModel.height =
       msg.height >= 0 ? msg.height : this.node.getBoundingClientRect().height;
     super.onResize(msg);
+    this._updater.invoke();
```
krassowski, line 947: #15357 also invokes update but only if height has changed (as otherwise there is nothing to update I believe) but also adds a test and can be backported as it is self-contained. Would you be interested in reviewing it? I can add the throttler patterns as in this PR over in #15357.

afshin: Good call, thank you @krassowski!

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,53 @@
...
+}
+
+ol.jp-WindowedPanel-scrollbar-content {
+  background-color: var(--jp-layout-color1);
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+}
+
+ol.jp-WindowedPanel-scrollbar-content li {
```
krassowski, line 47:
```suggestion
ol.jp-WindowedPanel-scrollbar-content > li {
```

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,53 @@
...
+  border-bottom: 1px solid var(--jp-layout-color3);
+  padding: 2px;
+  text-align: right;
+}
+
+ol.jp-WindowedPanel-scrollbar-content li:hover {
```
krassowski, line 53:
```suggestion
ol.jp-WindowedPanel-scrollbar-content > li:hover {
```

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,53 @@
...
   left: 0;
   right: 0;
   overflow: visible;
 }
+
+.jp-WindowedPanel-scrollbar {
+  display: none;
+}
+
+.jp-WindowedPanel.jp-mod-virtual-scrollbar .jp-WindowedPanel-scrollbar {
```
krassowski, line 28:
```suggestion
.jp-WindowedPanel.jp-mod-virtual-scrollbar > .jp-WindowedPanel-scrollbar {
```

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,53 @@
  */
 
 .jp-WindowedPanel-outer {
+  height: 100%;
   position: relative;
-  overflow-y: auto;
+  overflow: auto;
+  transform: none;
```
krassowski, line 10:
```suggestion
  transform: none;
}

.jp-WindowedPanel.jp-mod-virtual-scrollbar > .jp-WindowedPanel-outer {
  margin-left: 25px;
}
```

could this solve the issue of lacking margin:

Before (current):
Screenshot from 2023-11-08 11-47-33

Proposed:
Screenshot from 2023-11-08 11-52-12

krassowski (commented):
A few CSS suggestions (reducing CSS selector matching cost and improving alignment).

afshin (commented):
bot please update snapshots

afshin:
bot please update snapshots

afshin:
bot please update snapshots

afshin:
I am not sure why this has happened, but at least one of the visual regression tests is because of the existence of one additional file in the file browser. I would have thought refreshing the snapshots would capture this but they didn't. I will try that again.

Screenshot 2023-11-22 at 16 48 16

afshin:
bot please update snapshots

krassowski:
> I am not sure why this has happened, but at least one of the visual regression tests is because of the existence of one additional file in the file browser. I would have thought refreshing the snapshots would capture this but they didn't. I will try that again.

This is due to GitHub Actions runners now using concurrency and our tests not having been adopted yet. This is tracked in https://github.com/jupyterlab/jupyterlab/issues/15372 and there are two PR which tackle some of the issues:
- https://github.com/jupyterlab/jupyterlab/pull/15375 (draft)
- https://github.com/jupyterlab/jupyterlab/pull/15355 (waiting for review, only addresses one of the problems)

afshin:
bot please update snapshots

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,58 @@
...
+ol.jp-WindowedPanel-scrollbar-content > li {
+  border-bottom: 1px solid var(--jp-layout-color3);
+  padding: 2px;
+  text-align: left;
+}
+
+ol.jp-WindowedPanel-scrollbar-content > li:hover {
+  cursor: pointer;
+  background-color: var(--jp-layout-color2);
+}
```
fcollonval, line 61: It is uncommon to use HTML tag for styling. Should we stick to class based styling only?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -733,29 +734,48 @@ export class WindowedList<
    *
    * @param options Constructor options
    */
-  constructor(options: WindowedList.IOptions<T>) {
-    // TODO probably needs to be able to customize outer HTML tag (could be ul / ol / table, ...)
+  constructor(options: WindowedList.IOptions<U, T>) {
```
fcollonval, line 737: Could you avoid switching the template `U, T` to ease understanding (because U becomes T what is confusing).

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1182,18 +1270,56 @@ export class WindowedList<
...
+        item.dataset.index = `${index}`;
+        content.appendChild(item);
+      }
+    }
+  }
+
+  /**
+   * Handle `pointerdown` events on the virtual scrollbar.
+   */
+  private _evtPointerDown(event: PointerEvent): void {
```
fcollonval, line 1297: Should it be on PointerUp instead?

If the user press than move while pressing, what is expected to happen?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1182,18 +1270,56 @@ export class WindowedList<
...
+    }
+
+    const list = viewModel.itemsList;
+    if (list?.length) {
+      for (let index = 0; index < list.length; index += 1) {
+        const item = renderer.createScrollbarItem(index, list.get(index), this);
+        item.dataset.index = `${index}`;
+        content.appendChild(item);
+      }
+    }
```
fcollonval, line 1291: `itemsList` is optional because a `WindowedList` can be used for a fixed-size list.

Therefore this should become:

```suggestion
    const count = viewModel.itemsList?.length ?? viewModel.widgetCount;
    for (let index = 0; index < count; index += 1) {
      const item = renderer.createScrollbarItem(index, list?.get(index), this);
      item.dataset.index = `${index}`;
      content.appendChild(item);
    }
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1420,6 +1587,7 @@ export namespace WindowedList {
      * Items list to be rendered
      */
     itemsList: {
+      get(index: number): T;
```
fcollonval, line 1590: Should be optional to be backward compatible:

```suggestion
      get?: (index: number) => T;
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1524,16 +1692,56 @@ export namespace WindowedList {
...
+     * Create the virtual scrollbar element.
+     */
+    createScrollbar(): HTMLOListElement;
+
+    /**
+     * Create an individual item rendered in the scrollbar.
+     */
+    createScrollbarItem(
+      index: number,
+      item: T,
```
fcollonval, line 1737: If the `itemsList` is not provided, `item` cannot be defined.

```suggestion
      item?: T,
```

packages/notebook/src/widget.ts
```diff
@@ -1284,10 +1292,42 @@ export class Notebook extends StaticNotebook {
...
+        createScrollbarItem(
+          index: number,
+          model: ICellModel,
+          notebook: Notebook
+        ): HTMLLIElement {
+          const li = document.createElement('li');
+          li.appendChild(document.createTextNode(`${index + 1}`));
+          if (notebook.selectedCells.some(cell => model === cell.model)) {
+            li.className = 'jp-mod-selected';
+          }
```
fcollonval, line 1321: Should we distinguish between selected and active?

packages/notebook-extension/src/index.ts
```diff
@@ -3439,6 +3450,24 @@ function addCommands(
       }
     }
   });
+  commands.addCommand(CommandIDs.virtualScrollbar, {
+    label: trans.__('Virtual Scrollbar'),
+    caption: trans.__('Virtual Scrollbar (enabled with windowing mode: full)'),
```
fcollonval, line 3455:
```suggestion
    caption: trans.__('Toggle virtual scrollbar (enabled with windowing mode: full)'),
```

fcollonval (changes requested):
Thanks @afshin 

I got a couple of questions. And a macro one, should we tag some new API as experimental as the feature will likely be refined when it gets tested by more people?

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -963,7 +963,39 @@ export class CommandToolbarButton extends ReactWidget {
...
   constructor(private props: CommandToolbarButtonComponent.IProps) {
     super();
+    const { commands, id, args } = props;
     addCommandToolbarButtonClass(this);
+    this.setCommandAttributes(commands, id, args);
+    commands.commandChanged.connect((_, change) => {
+      if (change.id === props.id) {
+        this.setCommandAttributes(commands, id, args);
+      }
+    }, this);
```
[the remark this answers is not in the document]
afshin: I can't isolate where the problem is, but that `shouldUpdate` seems to not be respected in all cases.

If I remove the signal connection I added and use the state of things before this PR, this is the behavior of changing the setting.

**before this PR**

https://github.com/jupyterlab/jupyterlab/assets/159529/d4fb79a5-336d-47ac-8975-0ee1551b6e24

**after this PR**

https://github.com/jupyterlab/jupyterlab/assets/159529/78f20de7-63e7-4c44-9b79-94c482e9c997

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1524,16 +1692,56 @@ export namespace WindowedList {
...
+  export interface IRenderer<T = any> {
+    /**
+     * Create the outer, root element of the windowed list.
+     */
+    createOuter(): HTMLElement;
+
+    /**
+     * Create the virtual scrollbar element.
+     */
+    createScrollbar(): HTMLOListElement;
```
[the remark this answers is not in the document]
afshin: I'm not sure about this because it is inherently an ordered list (going into the windowed list) ... and both an `<ol>` and its children `<li>` elements are very flexible in the content they support. I would propose keeping this signature to give consumers of it some idea and structure for what the underlying code is doing.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1524,16 +1692,56 @@ export namespace WindowedList {
...
+
+    /**
+     * Create an individual item rendered in the scrollbar.
+     */
+    createScrollbarItem(
+      index: number,
+      item: T,
+      list: WindowedList
+    ): HTMLLIElement;
```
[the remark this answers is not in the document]
afshin: Same thought here ... I'm not too strongly committed to leaving it as is, but mildly so.

packages/ui-components/style/toolbar.css
```diff
@@ -66,6 +66,19 @@
...
+
+button.jp-ToolbarButtonComponent.lm-mod-hidden {
+  display: none;
+}
+
+button.jp-ToolbarButtonComponent > span {
+  padding: 0;
+  flex: 0 0 auto;
+}
+
```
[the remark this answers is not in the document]
afshin: You are correct. These are artifacts of an earlier text button and it is superfluous. I'll remove it.

packages/ui-components/style/windowedlist.css
afshin: I think this is related to your suggestion above. If the API contract is for an `<li>` element, we could leave this as is. So I suppose: how strongly do you feel about the above `<ol>` and `<li>` conversation?

packages/notebook/src/widget.ts
afshin: I think that is a good idea and we should do it. We should probably add a lot of semantic information to this scrollbar over time and have it approach being a document mini-map. Do you think it is necessary for a minimum viable feature?

packages/ui-components/src/components/windowedlist.ts
afshin: `createScrollbarItem` is only called if the list exists and is populated.
```ts
const list = viewModel.itemsList;
if (list?.length) {
  for (let index = 0; index < list.length; index += 1) {
    const item = renderer.createScrollbarItem(index, list.get(index), this);
    item.dataset.index = `${index}`;
    content.appendChild(item);
  }
}
```

packages/ui-components/src/components/windowedlist.ts
afshin: You're right. I was unhappy about this at the time. I'll look at this again.

packages/ui-components/src/components/windowedlist.ts
afshin: Maybe `item: T | undefined`

afshin:
> you mentioned some extension manger failures on the call but I do not see any. Can you point me to it? PS. I see the linter failing.

@krassowski, thanks for the lint hint! I pushed that change.

In my branch, if I run `jlpm build:test` this is what happens:

Screenshot 2023-11-30 at 00 32 47

krassowski:
Sorry for not replying earlier, I was planning to open a PR fixing the pluginmanager issue this evening. I think a different solution would be preferred than what you have in [6cd186b](https://github.com/jupyterlab/jupyterlab/pull/15109/commits/6cd186bf5212ab14ca42d3a141c028d20793d12d), this is simply removing `test:build` from `package.json`; that was an omission in my PR.

afshin:
bot please update snapshots

afshin:
bot please update snapshots

krassowski:
https://github.com/jupyterlab/jupyterlab/pull/15496 has a temporary fix for the labeller.

packages/notebook/src/widget.ts
afshin: Added the CSS modifier and an initial style for `selected` and also for `active`.

afshin:
bot please update snapshots

afshin:
bot please update snapshots

packages/ui-components/style/windowedlist.css
krassowski: A different point: if we were to migrate to a custom webcomponent in the future (see the recent toolbar PR) not having the `ol` and `li` in style files would be somewhat easier to make this change. I somewhat prefer to exclusively use class names if possible.

Performance-wise there are arguments both ways (e.g. `li:hover` will initially match on a lot of things but will be quickly invalidated via parent, but then `.jp-WindowedPanel-scrollbar-content > li` avoids adding a class name so may be tiny bit faster to populate).

galata/test/jupyterlab/sidebars.test.ts-snapshots/opened-sidebar-jp-running-sessions-jupyterlab-linux.png
krassowski, line 1: Look identical, can you manually revert it? (to reduce conflicts and repo size inflation)

galata/test/jupyterlab/readonly.test.ts-snapshots/readonly-jupyterlab-linux.png
krassowski, line 1: Look identical, can you manually revert it? (to reduce conflicts and repo size inflation)

galata/test/jupyterlab/notebook-search.test.ts-snapshots/search-within-outputs-jupyterlab-linux.png
krassowski, line 1: I do not understand why this is happening?

galata/test/jupyterlab/notebook-search.test.ts-snapshots/search-in-two-selected-cells-jupyterlab-linux.png
krassowski, line 1: Look identical, 4 bites difference ?

galata/test/jupyterlab/debugger.test.ts-snapshots/image-debug-session-global-variables-jupyterlab-linux.png
krassowski, line 1: Looks lke a fluke, needs reverting?

krassowski (commented):
Comments on snapshots

galata/src/helpers/notebook.ts
```diff
@@ -891,7 +891,7 @@ export class NotebookHelper {
     if (gutters.length < lineNumber) {
       return false;
     }
-    await gutters[lineNumber].click();
+    await gutters[lineNumber - 1].click();
```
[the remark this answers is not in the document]
afshin: I thought I removed this, hm, I'll remove it now.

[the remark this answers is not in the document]
afshin: I removed this. It was an ineffective fix. I thought I had removed it before but it may have accidentally been reintroduced in a rebase.

galata/test/documentation/general.test.ts-snapshots/running-layout-documentation-linux.png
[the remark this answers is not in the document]
afshin: I think this might be an artifact of how I rebased (by grabbing what was in the `main` branch) because I didn't know how else to resolve a merge conflict when rebasing.

afshin:
bot please update snapshots

krassowski:
Taking it over for snapshot management :building_construction:

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,58 @@
  */
 
 .jp-WindowedPanel-outer {
+  height: 100%;
```
krassowski, line 7: This breaks position of notebook footer

| Before | After |
|--------|--------|
| image | image |

packages/ui-components/style/windowedlist.css
```diff
@@ -4,17 +4,58 @@
...
   position: relative;
-  overflow-y: auto;
+  overflow: auto;
+  transform: none;
```
krassowski, line 10: Why is transform needed here?

packages/ui-components/style/windowedlist.css
krassowski: It looks like the solution is to move `jp-Notebook-footer` inside `jp-WindowedPanel-outer`

packages/notebook/src/widget.ts
```diff
@@ -193,16 +193,19 @@ export class StaticNotebook extends WindowedList {
    */
   constructor(options: StaticNotebook.IOptions) {
     const cells = new Array<Cell>();
+    const windowingActive =
+      (options.notebookConfig?.windowingMode ??
+        StaticNotebook.defaultNotebookConfig.windowingMode) === 'full';
```
krassowski, line 198: When windowing gets switched of the scrollbar stays on, but the button hides. It seems like this needs to be a callback or a signal?

krassowski:
This is kind of hard to manage due to faults in GitHub UI. After every force-push I have to click 7 times to get the most recent discussion show up:

1-minute-long-srcolling-session

I also get two emails, one for the pull request and one for the commit with "fix plugin manager test @krassowski" message.

As a partial alleviation, can I kindly suggest to:
- squash all the `Update Playwright Snapshots` commits as these do not provide any thing useful (in fact I would say lets revert _all_ snapshots and correct the CSS styles so that there are no changes at all; something I started working in the evening but did not manage to get in time)
- (minor) remove usernames from commit messages
- remove all comments requesting bot snapshot updates, and stop issuing those. I think we do not need any single snapshot to change in this PR, it just makes the review harder.

krassowski:
I see how this could be merged soon if the style changes are limited to when the scrollbar is indeed active. It would ensure no breakages downstream. This could be confirmed by snapshots having no changes.

packages/ui-components/style/windowedlist.css
afshin: I removed the tag names from the CSS and updated the `IRenderer` interface to not require specific tags.

packages/ui-components/style/windowedlist.css
afshin: Thank you! Fixed.

packages/ui-components/style/windowedlist.css
afshin: It appears it isn't. I made a lot of iterations here and this is an artifact of that. Thanks!

PR #12806, opened 2022-07-16
#12717 Add a new section: automation of local dev environments

markgreene74:
## References

This PR is related to #12717.

## Code changes

With this PR a new sub-section "Using automation to set up a local development environment" is added to "Setting up a local development environment" in `contributing.rst`.

The sub-section contains a small description of some of the advantages and link to a working example of Vagrantfile and additional documentation.

## User-facing changes

Changes to the "Contribute" documentation.

## Backwards-incompatible changes

N/A, no changes to JupyterLab public APIs.

PR #15111, opened 2023-09-12
Create dev environment with docker

trungleduc:
## References

Similar to the `gitpod` setup, but for people who prefer to work offline.
To start a JupyterLab dev container on a UNIX system with docker installed:
```bash
bash docker/start.sh
````
This command will build the docker image if it does not exist, then start the container with JupyterLab running in watch mode. The port `8888` is exposed and the current JupyterLab repo is mounted into the container.

Other available commands:

- `bash docker/start.sh dev`: same as calling `bash docker/start.sh`
- `bash docker/start.sh stop`: stop the running container
- `bash docker/start.sh clean`: remove the docker image
- `bash docker/start.sh build`: rebuild the docker image
- `bash docker/start.sh shell`: log into the container's shell with the JupyterLab environment activated. It's useful to run the tests or install dependencies. 

##  How it works

To avoid polluting the `node_module` directory with packages coming from the host machine while still preserving the hot-reload capability, the js dependencies in the container are actually stored in the `/home/$DEV_USER/jupyterlab_cache/node_modules/` while the working directory is mounted to `/home/$DEV_USER/jupyterlab`. On container startup,  the `node_modules` at `jupyterlab_cache` is `rsync-ed` back to the host.

Since the image and container name contain the hash of the current JupyterLab directory, two clones of JupyterLab will generate different images.

## How to add dependencies 

Due to the above setup, the easiest way to add new dependencies is to log into the container's shell, install the dependencies to update the `package.json` and `yarn.lock` files, and then rebuild the docker image.

```bash
bash docker/start.sh shell
# In the container shell
jlpm add ...
exit
# back to host shell
bash docker/start.sh build
```

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -1,12 +1,8 @@
 name: Docker Tests
 
 on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
-  release:
-    types: [published]
+  schedule:
+    - cron: '0 0 * * *'
 
 jobs:
   docker:
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -1,0 +1,32 @@
+name: Docker Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  release:
+    types: [published]
+
+jobs:
+  docker:
+    name: Docker Dev Container
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the image
+        run: |
+          bash ./docker/start.sh build
+          docker image ls 2>&1 | grep -ie "jupyterlab_dev" -
+
+      - name: Start the dev container
+        run: |
+          bash ./docker/start.sh dev-detach
+
+      - name: Wait for JupyterLab
+        uses: ifaxity/wait-on-action@v1
+        with:
+          resource: http-get://localhost:8888/
+          timeout: 360000
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,3 +1,6 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
 FROM mambaorg/micromamba:latest
 
 # Create new user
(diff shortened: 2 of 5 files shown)


trungleduc:
cc @jtpio for awareness

jtpio:
Thanks @trungleduc.

Maybe this will also be useful to enable better support for dev containers and GitHub Codespaces. And also streamline the setup between Gitpod and Codespaces: https://github.com/jupyterlab/jupyterlab/issues/13049

jtpio:
Also should there be a new section in the contributing guide to mention how to use this setup? https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html

trungleduc:
> Also should there be a new section in the contributing guide to mention how to use this setup? https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html

Done

jtpio:
Similar to the discussion about Vagrant (https://github.com/jupyterlab/jupyterlab/pull/12806 and https://github.com/jupyterlab/team-compass/issues/152#issuecomment-1191170828), one of the main challenges with this setup will be for maintainers to be able to maintain this Docker configuration, since Docker is yet another tool and not everyone might be familiar with it.

But since Docker is available on GitHub Actions by default maybe there could be a way to put it under tests? To make sure the config does not lag behind or become out of date.

trungleduc:
@jtpio I added a workflow for building and starting the dev container. But I am not sure about the activation rate, running this test on every commit wastes a lot of resources.

krassowski:
We could run it on PR approval only, as we do with benchmarks.

jtpio:
Otherwise as a cron job?

But I guess we would first need to know if someone from the JupyterLab maintainers will be willing to maintain this new dev setup.

trungleduc:
> Otherwise as a cron job?
> 
> But I guess we would first need to know if someone from the JupyterLab maintainers will be willing to maintain this new dev setup.

I can maintain this script but I am also open to move it outside of the JupyterLab repo

krassowski:
I am ok with having this in if it helps at least one contributor - if it falls unmaintained, removing it will be easy (4 files to delete plus documentation fragment to cut).

fcollonval (approved):
Thanks @trungleduc

trungleduc:
> Otherwise as a cron job?
> 
> But I guess we would first need to know if someone from the JupyterLab maintainers will be willing to maintain this new dev setup.

 a cron job sounds good! I'll update the action script.

krassowski (approved):
Thanks!

PR #15132, opened 2023-09-17
Fix completer width inflation and jitter

### LINKED ISSUE #15131: Completer width keeps increasing when only one item is left and panel is open
## Description

When completer documentation panel is open, as user types the completer width keeps increasing in leaps of size similar to documentation panel itself

## Reproduce

1. Enable "Show the documentation panel":
   image
2. In IPython kernel type `l`
3. Trigger completer with <kbd>tab</kbd>
4. Slowly type `icense`:
   width-goes-wrong

## Expected behavior

The completer width stays the same once it was displayed (or possibly reduces are fewer items are displayed)

## Context

- JupyterLab version: 4.0.6

krassowski:
## References

- fixes #15131
- moves the documentation loading indicator from the documentation panel to over the item, which prevents a visual jitter when the documentation panel shows for a split second and then closes because there is no documentation

## Code changes

- always pre-compute the completer box, use heuristics for height too
- always attach the documentation panel
- extend the internal size cache interface to use it as source of truth for documentation box size

### 1. Typing no longer increases width of completer box

| Before | After |
|--------|--------|
| width-goes-wrong | no-w-jitter |

#### 2. Typing no longer causes jitter

| Before (but after fixing point 1) | After |
|--------|--------|
| after-2 | no-w-jitter |

This was occurring when less than a full page of items was rendered as for that case we were not pre-computing the box size; now we are always pre-computing.

#### 3. No jitter due to panel being show to reveal loading indicator 

| Before | After |
|--------|--------|
| jittter04 | no-jitter |

Before the completer documentation panel was showing for a split second.

#### 4. New positioning of loading indicator

| Before | After |
|--------|--------|
| oldIndicator | newIndicator  | 

In practice I found that the loading indicator is almost never shown for longer than half a second, which really creates an annoyance due to flashing panel; ideally we would pre-fetch the documentation for +/-2 items from the active one to avoid the loading indicator in most scenarios altogether (which is what LSP extension does).

I am still not satisfied with the loading indicator itself, but this is a CSS-only job now. One idea would be to animate the full background of active item instead of a thin progress bar. Alternatively, we could have an animated spinner or dots at the right edge.

None.


krassowski:

trungleduc (approved):

trungleduc:
Thanks @krassowski, it looks neat!

fcollonval (approved):
Thanks @krassowski

PR #15133, opened 2023-09-18
Add a valid email in update extension

brichet:
This PR fixes the *usage* test, which currently fails while upgrading the mock extension.

## References

Probably since https://github.com/jupyterlab/extension-template/pull/43, which requires a valid email pattern.

Error: https://github.com/jupyterlab/jupyterlab/actions/runs/6221683949/job/16884129806

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/jupyterlab/upgrade_extension.py
+++ b/jupyterlab/upgrade_extension.py
@@ -116,7 +116,7 @@ def update_extension(  # noqa
 
     # Build up the template answers and run the template engine
     author = data.get("author", "<author_name>")
-    author_email = "author_email@jupyter.com"
+    author_email = ""
     if isinstance(author, dict):
         author_name = author.get("name", "<author_name>")
         author_email = author.get("email", author_email)


jupyterlab/upgrade_extension.py
```diff
@@ -116,7 +116,7 @@ def update_extension(  # noqa
 
     # Build up the template answers and run the template engine
     author = data.get("author", "<author_name>")
-    author_email = "<author_email>"
+    author_email = "author_email@jupyter.com"
```
fcollonval, line 119: Could we use empty instead of a dummy placeholder? That case is also available in the extension template.

```suggestion
    author_email = ""
```

fcollonval (changes requested):
thanks @brichet 

I have a minor suggestion.

fcollonval (approved):
Thanks @brichet

fcollonval:

PR #9622, opened 2021-01-15
Add aria roles and labels

marthacryan:
## References
Fixes the first part of #9491 to add basic roles and aria labels to various panels in JupyterLab. 

## Code changes
See issue for specific areas that are changed. The code only adds attributes to nodes of widgets. 
Todo:
- [x] Add translation

## User-facing changes
For non screenreaders, there are no user-facing changes. This gives more information that can be used for accessibility purposes. 

## Backwards-incompatible changes
None

Issue #13045, opened 2022-09-02
ARIA labels on main area widget are incorrect (always referring to a notebook)


krassowski:
## Description

https://github.com/jupyterlab/jupyterlab/blob/b57ec962a7b8cda65bb89a72223f195564f8387a/packages/apputils/src/mainareawidget.ts#L46-L50

says "notebook content" when using a file editor or opening a CSV file, etc, but it should not.

We could:
- change it to "main area content" and "main area toolbar"
- in each of the `MainAreaWidget` subclasses set a proper text
  - we could take the string from a getter so subclasses have less work (getter so that we can translate)
  - if we want to be more strict we could use an abstract getter and force all subclasses of `MainAreaWidget` to provide an aria-label since 4.0

thevyomonaut:
I would like to work on this, please! I am a first timer, so some help would be appreciated.

scmmmh:
I've been looking into this and I have a question, which may be caused by my misunderstanding of the architecture. Since the MainAreaWidget is generic, shouldn't the aria-label value be taken from the Widget that is used to parameterise the MainAreaWidget?

krassowski:
@scmmmh you are right, sorry about inaccurate suggestion. Indeed for the most part we do not subclass `MainAreaWidget` (`DocumentWidget` aside). Using a property from the parameter Widget seems like a good idea!

Maybe it is the Widget's responsibility to set the `aria-label` and `rule` on its node, and not a responsibility of `MainAreaWidget`, so we could remove `MainAreaWidget` from the equation if we make changes in lumino's Widget instead?

scmmmh:
@krassowski I agree, the Widget is probably the better place to add that support. I'll look into that and then this issue can be used to track updating the MainAreaWidget when the lumino Widget has been updated.

m158261:
There is a test that covers the incorrect behavior described in this issue. This test will need to be updated when this issue is worked on. The test can be found in this merged PR. https://github.com/jupyterlab/jupyterlab/pull/15137

dnlzrgz:
Hi everyone,

I’ve been looking into this issue and noticed that there are currently no open PR addressing it. I’ve checked the previous PR and reviewed the test mentioned, but I wanted to confirm if this issue is still relevant before I create a pull request of my own.

krassowski:
Yes, this issue is still relevant - feel welcome to open a PR!

PR #15137, opened 2023-09-19
Added tests to check for aria labels and roles for accessibility

m158261:
## References

Related PR - https://github.com/jupyterlab/jupyterlab/pull/9622

## Code changes

Added tests to check aria labels and roles are present for users with screen readers. There a mostly around sidebar widgets. Some of the tests are run in Galata due to restrictions with rendering extension packages. 

## User-facing changes

None

None


packages/extensionmanager-extension/src/index.ts
```diff
@@ -57,6 +57,7 @@ const plugin: JupyterFrontEndPlugin<void> = {
       v.id = 'extensionmanager.main-view';
       v.title.icon = extensionIcon;
       v.title.caption = trans.__('Extension Manager');
+      v.node.setAttribute('data-testid', 'extension-manager-section');
```
krassowski, line 60: What is `data-testid`?

m158261: It's used by getByTestId in Galata to get the extension manager section. I tried using page.locator() with the id 'extensionmanager.main-view' but that returned multiple elements.

krassowski: In the commit history I see that you tried `getByTitle` but I do not see you using the ID locator directly (instead of using `getByTitle`. I think this is not needed - can you give using just ID another try please?

m158261: Hey @krassowski, I've tried just using the ID locator and that can't find the element. If I try and click on the extension manager it returns two elements. I've tried first() and last() but those time out too. I think it's because there are two elements that have an id or data-id of extensionmanager.main-view. Adding the data-testid was the solution that I found that worked.

krassowski: Do you see two elements with `extensionmanager.main-view` ID in the page? That would be a pretty serious bug. My guess is that there was some mistake in the selector, for example unescaped dot `.`. Did you try `#extensionmanager\\.main-view`?

m158261: The extension manager section has an ID of extensionmanager.main-view. The extension manager li has a data-id of extensionmanager.main-view. 

I'll try it with an escaped character. Thanks, I didn't think of that.

packages/application/test/shell.spec.ts
```diff
@@ -517,4 +517,61 @@ describe('LabShell', () => {
       expect(widget.isVisible).toBe(false);
     });
   });
+
+  describe('#ariaLabelsAndRoles', () => {
```
krassowski, line 521: I would name the group `accessibility` in case if we want to add more tests which do not fit neatly to specific methods or attributes in the future.

packages/apputils/test/mainareawidget.spec.ts
```diff
@@ -23,6 +23,25 @@ describe('@jupyterlab/apputils', () => {
...
+          'notebook actions'
+        );
+        expect(widget.toolbar.node.getAttribute('role')).toEqual('navigation');
+      });
+
+      it('content should have an aria-label of notebook content and a role of region', () => {
+        const content = new Widget();
+        const widget = new MainAreaWidget({ content });
+        expect(widget.content.node.getAttribute('aria-label')).toEqual(
+          'notebook content'
```
krassowski, line 41: This test demonstrated incorrect behaviour (not every main area widget is a notebook, for example a text editor is not a notebook, csv viewer is not a notebook etc), see https://github.com/jupyterlab/jupyterlab/issues/13045. I am not sure what the default for a main area widget should be but my expectation is that it would not be "notebook". Also, is the toolbar really `navigation`? Shouldn't it be a [toolbar role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role)?

m158261: I can change the navigation role to toolbar. As the notebook content label still has an outstanding issue, I could either remove the test or leave it to be updated when that issue is worked on?

packages/application/test/shell.spec.ts
m158261: Changes made.

packages/extensionmanager-extension/src/index.ts
m158261: The test now passes after adding an escaped character.

packages/apputils/test/mainareawidget.spec.ts
krassowski: Let's keep the test, but:
- mark it with [`test.failing`](https://jestjs.io/docs/api#testfailingname-fn-timeout) as xfail for now
- add a comment on the test pointing to issue
- once this PR is merged, comment on the issue that there is now a test.

m158261: Changes made.

krassowski (approved):
Thank you  @m158261!

PR #15142, opened 2023-09-21
Page_config attribute to handle keydown event at bubbling phase

brichet:
This PR adds a new attribute to the `page_config.json` file, to allow handling the keydown event at bubbling phase instead of capture phase.

--------------------------

**page_config.json**
```json
{
  "bubbling_keydown": true
}
```

See the [documentation](https://jupyterlab.readthedocs.io/en/latest/user/directories.html#labconfig-directories) to set it up.

## References

See https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1727686379 for context.

Draft PR until https://github.com/jupyterlab/lumino/pull/635 is merged and released.
 
## Code changes

When launching the application (*dev_mode/index.js*), the value of `bubbling_keydown` (boolean) is send to the lumino `Application`.
If the `bubbling_keydown` is `false` or `undefined`, the application starts as usual.

## User-facing changes

None

None

cc. @fcollonval @krassowski @gabalafou


brichet:
> Ok I did not think about using the pageconfig. It indeed works without the need for a dynamic settings change in Lumino. But is it convenient enough for testing?

Yeah, I did not think about a dynamic flag :smile: 
This PR is probably useless since the dynamic flag is working. Except if we want long term testing, than it can be useful to avoid setting up the flag each time.

fcollonval (approved):
Thanks @brichet

fcollonval:
I'm moving forward with this to be able to get it in an alpha release

fcollonval:
> CI failures are due to flakiness

jtpio:
Is this a new setting that end users should be able to enable, or is it mostly for helping with https://github.com/jupyterlab/jupyterlab/pull/14115? Should there be a mention somewhere in the JupyterLab documentation?

krassowski:
I think we don't know the answer yet. If we find it to work well I would not expose it to users. If we find any issues then it might be worth letting user decide. I would suggest opening an issue to come back to it before releasing 4.1.

jtpio:
Opened https://github.com/jupyterlab/jupyterlab/issues/15158

PR #15122, opened 2023-09-14
Backport PR #15117 on branch 3.6.x (Install playwright browser in jupyterlab.browser_check)

krassowski:
Backport PR https://github.com/jupyterlab/jupyterlab/pull/15117 on branch 3.6.x (Install playwright browser in jupyterlab.browser_check)

PR #15154, opened 2023-09-25
Remove pre-commit job

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/15122

## Code changes

Remove pre-commit job as done in https://github.com/jupyterlab/jupyterlab/pull/13562 for 4.x

That step is not reliable and it is duplicating the job using the pre-commit app.


krassowski (approved):
Thank you!

PR #15156, opened 2023-09-26
Debugger sidebar variables tree expansion states restoration

### LINKED ISSUE #14488: Debugger: Retain variable expansion
### Problem

When using the debugger is common to jump up/down in the stack or take a step forward. Lab forgets about what you expanded previously so it leads to a lot of extra clicks. The gif shows both of these below.

debugger variables

### Proposed Solution

If the user choose to expand them, keep them expanded.

### Additional context

Tried in lab@4.0.0b2 and lab@3.6.2. 

As an example, Chromium's debugger preserves this.

DenisaCG:
## References
Save the expansion states, set by the user, of the variables in the tree view of the debugger sidebar and restore them along with the layout. 

I'm working to a variant to not change the API.

## Code changes

The state is stored to be restored when switching from cell-by-cell run and debug step-by-step.
It is also restored when switching between tabs.

## User-facing changes

Variable tree view expansion is stored


krassowski:
Thanks for starting this @DenisaCG. I am moving this to 4.2 as we are past feature freeze on 4.1 milestone and this one appears to still be a work in progress. Looking forward to reviewing this one once ready!

DenisaCG:
Thanks for all the work @fcollonval! Everything seems to be working great. 

The expansion tree is stored and restored properly when jumping up and down in the stack or even when closing and opening the `Variables` subpanel or the entire `Debugger` panel itself. 

The only drawback I noticed was that once we refresh the page, the expansion tree is not saved, but everything else from the debugging process is restored as previously left (breakpoints, callstack) . Is this something that would be easy to improve on?

krassowski:
It looks like this accumulated a few conflicts, possibly due to https://github.com/jupyterlab/jupyterlab/pull/15375.

krassowski:
@DenisaCG and @fco just checking back here as we are nearing 4.2 release - do you have time to resolve the merge conflicts and fix the failing lint check?

krassowski:
I moved this to 4.4 as it appears there has not been any recent movement on this, but if you do get to finish it before Monday which is the feature freeze for 4.3 please do let me know and I will change the milestone back!

krassowski:
This was not updated in a few years, I will close it for now as it requires extensive rebase to work with the current code. Please feel welcome to open a fresh PR if you would like to add this enhancement.

PR #10346, opened 2021-06-04
Feature: select all cells with tags in toc

NPetz:
## References

closes [#8566](https://github.com/jupyterlab/jupyterlab/issues/8566)

## Code changes

- create a new method that selects all the cells that have all of the current selected tags and activates the first of those cells

## User-facing changes

- creates a new button the near the 'Clear Filters' button 

Screen Shot 2021-06-05 at 0 57 55

Issue #8566, opened 2020-06-13
I'm not seeing - Select All Cells with this Tag - option


LukaPitamic:
## Description

Option `Select All Cells with this Tag` is missing since extension was included by default into JL. 
Some of our workflows are dependent on executing parts of the notebook which was easily achieved with this option and then executing selected cells only. 

## Reproduce

1. open cell tags UI (left side panel)
2. there is no option: Select All Cells with this Tag

## Context

- Operating System and version: MacOS
- Browser and version: Brave 1.2.80
- JupyterLab version: 2.1.3

andswitch:
Imho this is an important feature for large notebooks. In that situation I want to create reusable code cells. Using tags, I can group dependent code cells. However, I can't efficiently run a group without this option!

NPetz:
Hello, I'd like to try working on this one as my first contribution to OSS if nobody else is working on it and if it is needed.

As far as UI goes, would a mashup of the extension this feature comes from with the current UI be ok?
Something like in the rightmost screenshot.
<div style="display:flex;">
<img width="248" alt="current" src="https://user-images.githubusercontent.com/38439788/120341835-303cbc80-c332-11eb-86b3-cf6dab4ec04e.png">
<img width="226" alt="extension" src="https://user-images.githubusercontent.com/38439788/120341821-2ca93580-c332-11eb-8b50-5405d0d6ec09.png">
<img width="226" alt="future" src="https://user-images.githubusercontent.com/38439788/120342524-d5579500-c332-11eb-8661-29f039d21aef.png">
</div>

LukaPitamic:
@NPetz WTG!!! It looks more than perfect to me

krassowski:
Did this go into the TOC package? I wonder if it would be better to have it in cell tools. What is the rationale for having it in TOC rather than in cell tools?

NPetz:
@krassowski 

I think it makes more in sense in the TOC because the cell tools (notebook tools?) is used to _edit_ cells, whereas the TOC is used to _filter/select_ cells in the entire notebook.

What do you think?

LukaPitamic:
@NPetz (IMH) I agree with @krassowski - same tool for making tags in the first place and selecting cells with tags makes perfect sense. However the effort being put into adding feature is worth more. This was a feature I was missing for a long time.

PR #15157, opened 2023-09-26
Add filtering notebook cells with tags

### LINKED ISSUE #8298: Add Data Attribute to Parameter Cells
## Description
We should add a CSS data attribute to a cell that has the papermill "parameters" [tag](https://papermill.readthedocs.io/en/latest/usage-parameterize.html#lab).  Right now you'd have to dig into the model to get this information yourself.  This information is useful because you might want to style the parameter cell differently.  In fact, we may want to do that ourselves at some point.  

This can be accomplished at the cell level, by listening for changes to cell metadata and applying or removing a CSS attribute based on the presence of that tag.

### LINKED ISSUE #12542: Add filtering notebook cells by tags
### Problem

In https://github.com/jupyterlab/jupyterlab/pull/12374, the filtering of cells by tags was removed from the table of contents. That feature must be restored.

### Proposed Solution

We won't go the same road of adding content in the table of content - that panel will and should focus only on outline. Moreover displaying part the document in small font is bad for performance (multiplication of lots of DOM elements) and for accessibility (it can be very challenging to see clearly small colored fonts).

So the solution envisioned is to add a specific tags filtering tool in the notebook tools sidebar. Filtering by tags will trigger collapsing the non-selected cells (inputs and outputs). And a button to undo that collapsing for all cells will be set-up.

Additionally a data attribute for each tags will be added as recommended in https://github.com/jupyterlab/jupyterlab/pull/8410#issuecomment-631030585. This will allow further investigation and third-party extensions to implement ideas like https://github.com/jupyterlab/jupyterlab/pull/8630.

> We would recommend using the heading collapsing feature to group cells rather than using tags filtering.

HaudinFlorence:
This PR aims at adding a way to filter (basically to collapse) cells of a notebook that have specific tags. It provides a dialog listing all the tags of all the cells. The user can choose and check tags in the interface. All the tagged cells will then be collapsed respectively.

filter_cell_collapse_with_tags2

Final styling
Screenshot from 2023-10-29 10-00-05

## References
This PR 
fixes https://github.com/jupyterlab/jupyterlab/issues/12542
fixes https://github.com/jupyterlab/jupyterlab/issues/8298

## Code changes
In `packages/notebook-extensions/src`:
- add a jupyterFrontEndPlugin with a command called `filterCells`
- add a new file `celltaglist.tsx`
- add button in the common tools associated with `filterCells` command: when clicking a dialog open displaying the list of available tags for all the cells


krassowski:
Is there an issue discussing design of this feature? I know that tag filtering was previously (3.x) available in Table of Contents. From the UX point of view it appears of high important to indicate that the tag filters that are active. Is the intent to move it back to ToC or keep it on the toolbar?

HaudinFlorence:
> Is there an issue discussing design of this feature? I know that tag filtering was previously (3.x) available in Table of Contents. From the UX point of view it appears of high important to indicate that the tag filters that are active. Is the intent to move it back to ToC or keep it on the toolbar?

I am not sure there is an issue discussing this design apart from what is said in issue https://github.com/jupyterlab/jupyterlab/issues/12542. For the moment the idea is too keep this filtering tool accessible from the toolbar. I would be happy to further discuss UX point. Can you please elaborate what you mean by its is of high importance to indicate what tags filters are active. Thanks.

HaudinFlorence:
> In past, when the feature of collapsing cells or collapsing headers was introduced users were very distressed when they triggered it accidentally or triggered it intentionally and forgot that it was active because this lead to impression of:
> 
> * cells being lost from notebook
> * unintentional sharing of a notebook with sensitive information because author thought it is no there
> 
> therefore whenever a feature temporarily hiding cells is active it is important to offer an indication of that.
> 
> On placement of the feature, we previously had it in the ToC, please see mockups/screenshots from [#8566 (comment)](https://github.com/jupyterlab/jupyterlab/issues/8566#issuecomment-852182930):
> 
> > current extension future
> 
> and #10346:
> 
> > image
> 
> At the time I suggested that maybe a good placement would be in the cell tools sidebar where the cell tags are added in the first place([#8566 (comment)](https://github.com/jupyterlab/jupyterlab/issues/8566#issuecomment-856147539)), but I do not have a strong opinion on it.

@krassowski Thanks for the precision. For the moment the tool is very basic but it is worth to keep this in mind for the future.

krassowski:
> For the moment the tool is very basic but it is worth to keep this in mind for the future.

I think that the UI/UX here should be addressed before merging because:
- historically there were comments about too many items in the toolbar, so adding a new one by default should have a broad consensus (on the other hand adding the same feature to a sidebar would be a no-brainier)
- it has a large impact on UX and a minimal solution could be as simple as using the `filterDotIcon` (if it was to stay in the toolbar)

HaudinFlorence:
@krassowski Thanks for the comment. I will either use the filterdot icon option in a first time or move the button to the common tools panel in a future commit. I am first addressing the addition of buttons to validate the tag selection or to clear all filters as it was in the previous filtering tool of jupyterlab 3 version.

HaudinFlorence:
You can see here a gif with demonstration of clicking on a button in the common tool, opening a dialog and having the possibility to check some filters

HaudinFlorence:
@krassowski I hope with this last commit we approach something close to what you preconised in terms of UX.

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
bot please update snapshots

packages/statusbar/src/components/hover.tsx
```diff
@@ -162,7 +162,7 @@ export class Popup extends Widget {
       offset: {
         horizontal: aligned
       },
-      privilege: 'forceAbove',
+      privilege: 'below',
```
krassowski, line 165: Was this change intended? What is its purpose?

HaudinFlorence: @krassowski This is a left-over of an attempt of using a popup instead of a dialog. I will remove it.

packages/notebook-extension/src/celltaglist.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { ReactWidget, VDomModel, VDomRenderer } from '@jupyterlab/apputils';
+import { CommandRegistry } from '@lumino/commands';
+
+interface IPropsFilterButton {
+  onClick: () => void;
+  sectionLabel: string;
+  buttonLabel: string;
+}
+
+export function FilterButtonComponent(props: IPropsFilterButton) {
```
krassowski, line 18: Should we reuse `Button` from `ui-components` instead to ensure consistent styling?

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/ui-components/src/components/button.tsx#L30C17-L30C23

packages/notebook-extension/src/celltaglist.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+export class FilterButtonWidget extends ReactWidget {
+  private _commands: CommandRegistry;
+  private _commandID: string;
+  private _buttonLabel: string;
+  private _sectionLabel: string;
+  constructor(
+    commands: CommandRegistry,
+    commandID: string,
+    sectionLabel: string,
```
krassowski, line 37:
```suggestion
```

`sectionLabel` appears to be set to `''` in the only usage I see, should it be removed?

packages/notebook-extension/src/celltaglist.tsx
```diff
@@ -0,0 +1,165 @@
...
+    <div>
+      <p> {props.sectionLabel}</p>
+      <button type="button" onClick={props.onClick}>
+        {props.buttonLabel}
+      </button>
+    </div>
+  );
+}
+
+export class FilterButtonWidget extends ReactWidget {
```
krassowski, line 29: I am not quite sure if we need a new dedicated button widget class for each button?

It seems to do mostly the same thing as [`CommandToolbarButton`](https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/ui-components/src/components/toolbar.tsx#L869C14-L869C34).

But `CommandToolbarButton` takes care of many accessibility, reactivity and usability aspects already.

On possibility would be moving most of the `CommandToolbarButton` code into a new `CommandButton` in `ui-components` and make the only distinction of `CommandToolbarButton` inherit from it and apply the toolbar classes.

packages/notebook-extension/src/celltaglist.tsx
krassowski: Alternatively, see the next comment about reusing `CommandToolbarButton` or splitting out a `CommandButton` out of it.

krassowski (commented):
Only commenting on the button implementation for now.

packages/notebook-extension/src/celltaglist.tsx
HaudinFlorence: @krassowski I can use one of the both type of already existing buttons. Do you think that one option is better ?

krassowski: Depends.
- a) `CommandToolbarButton` is a more complete implementation as it received a lot of care to ensure it shows right tooltip, reacts when command changes and can be interacted with by using keyboard as per accessibility requirements, but you would probably need to put some more time in to extract the proposed `CommandButton` out of it
- b) `Button` does not do any of that, just adds a style class, but you should be able to use it easily without any extra work.

(a) seems preferred to me long term, but short term this PR can be merged with (b).

Also, `CommandToolbarButton` is getting refactored in https://github.com/jupyterlab/jupyterlab/pull/15021 so I would hold off any work around it for now.

packages/notebook-extension/src/celltaglist.tsx
HaudinFlorence: @krassowski If we decide to split out a `commandButton`out of `CommandToolbarButton`I guess this should be done in another PR right ?

krassowski: Yes, another PR would be preferable

HaudinFlorence:
@krassowski I have replaced the previous custom react button by a `Button` from the `ui-components`. Sorry if it is obvious, but I did not find how to add text inside the button.

packages/notebook-extension/src/celltaglist.tsx
```diff
@@ -0,0 +1,138 @@
...
+    super();
+    this._commands = commands;
+    this._commandID = commandID;
+  }
+  onClick = () => this._commands.execute(this._commandID);
+  render() {
+    return (
+      <>
+        <p> Filter cells with tags</p>
+        <Button onClick={this.onClick} />
```
krassowski, line 27:
```suggestion
        <p>{this._trans.__('Filter cells with tags')}</p>
        <Button onClick={this.onClick}>{this._commands.label(this._commandID)}</Button>
```

where `_trans` needs to be threaded in the constructor separately.

krassowski: This is what I think should work because the Button accepts the [`children` props](https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children)

HaudinFlorence: @krassowski Ok thanks. The `FilterButtonWidget` has been updated taken this into account.

packages/notebook-extension/src/celltaglist.tsx
```diff
@@ -0,0 +1,145 @@
...
+    this._commands = commands;
+    this._commandID = commandID;
+    this._trans = translator.load('jupyterlab');
+  }
+  onClick = () => this._commands.execute(this._commandID);
+  render() {
+    return (
+      <>
+        <p>{this._trans.__(this._commands.label(this._commandID))}</p>
+        <Button onClick={this.onClick}>{'Open the filtering tool'}</Button>
```
krassowski, line 35: It should be the other way round. Translation should be applied in the second line as it has a string literal. Translation has no effect on strings in variables/returned by functions. This is because extraction is based on static analysis  please see https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html

```suggestion
        <p>{this._commands.label(this._commandID)}</p>
        <Button onClick={this.onClick}>{this._trans.__('Open the filtering tool')}</Button>
```

krassowski (dismissed):
This is quite close to something we can merge and iterate later on. Just three things:
- A conflict with the main branch needs resolving
- Snapshots need updating to reflect button changes
- Translations need correcting, see inline comment

krassowski:
Oh, sorry one more thing: are there any tests for this feature/plugin?

HaudinFlorence:
> Oh, sorry one more thing: are there any tests for this feature/plugin?

@krassowski That was planned. I started to implement one previously. But is is not yet completed and will take me some additionnal time to make it work.

HaudinFlorence:
bot please update snapshots

fcollonval:
bot please update snapshots

galata/test/documentation/plugin_manager.test.ts-snapshots/plugin-manager-plugin-required-warning-documentation-linux.png
fcollonval, line 1: @krassowski do you think we should make the plugin list scrollable to ensure the last line _Please disable..._ is always visible?

fcollonval:

galata/test/documentation/plugin_manager.test.ts-snapshots/plugin-manager-plugin-required-warning-documentation-linux.png
krassowski: Yes, good catch, thank you!

fcollonval: Done in [1dcb2be](https://github.com/jupyterlab/jupyterlab/pull/15157/commits/1dcb2beb63c81d09983f880a00c89e82d3b5af40)

fcollonval:

packages/notebook-extension/src/index.tsx
```diff
@@ -1060,6 +1065,87 @@ const activeCellTool: JupyterFrontEndPlugin<void> = {
...
+            model.filters = new Set<string>();
+          }
+        }
+      },
+      isEnabled: args => (args.toolbar ? true : isEnabled())
+    });
+
+    if (palette) {
+      palette.addItem({
+        category: 'Notebook Operations',
```
krassowski, line 1142: Translation?

packages/notebook-extension/src/index.tsx
```diff
@@ -1060,6 +1065,87 @@ const activeCellTool: JupyterFrontEndPlugin<void> = {
   }
 };
 
+export const filterCellsPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/notebook-extension:filter-cells',
+  description: 'Filter cells.',
```
krassowski, line 1070: Maybe rename the plugin to `:cell-filter` instead?

galata/test/documentation/export_notebook.test.ts-snapshots/exporting-slide-type-documentation-linux.png
krassowski, line 1: The alignment does not look great and the button does not look like it is a button.

Maybe using an icon (we already have a filter icon) and moving it could help but I would be fine with polishing it in a subsequent PR. For this one though let's at least consider if we can add a border or icon to make it clear it is a button prior to hovering.

packages/cells/src/widget.ts
```diff
@@ -683,11 +684,23 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
...
     }
   }
 
+  private _setTagsAttribute() {
+    const tags: string[] = this.model.getMetadata('tags') ?? [];
+    if (tags.length) {
+      this.node.dataset['tags'] = tags.sort().join(',');
+    } else {
+      delete this.node.dataset['tags'];
+    }
```
krassowski, line 701: @bollwyvl highlighted in a gitter discussion that this is not scoped. Indeed most of JupyterLab data attributes use `jp-` prefix, eg. `jp-theme-name`/`jpThemeName`, jp-code-runner/jpCodeRunner etc.

packages/notebook-extension/src/index.tsx
```diff
@@ -1060,6 +1065,87 @@ const activeCellTool: JupyterFrontEndPlugin<void> = {
...
+            label: trans.__('Filter Cell(s)'),
+            actions: ['select']
+          });
+          const clearButton = Dialog.warnButton({
+            label: trans.__('Clear Filter(s)'),
+            actions: ['clear']
+          });
+          const view = new CellFiltersView({ model, translator });
+          const result = await showDialog({
+            title: trans.__('Select cell filters:'),
```
krassowski, line 1124:
```suggestion
            title: trans.__('Select cell filters'),
```

packages/notebook-extension/src/index.tsx
```diff
@@ -1060,6 +1065,87 @@ const activeCellTool: JupyterFrontEndPlugin<void> = {
...
+            model = new CellFiltersModel({ notebook: current.content });
+            notebookModelMap.set(current.content, model);
+          }
+
+          const selectButton = Dialog.okButton({
+            label: trans.__('Filter Cell(s)'),
+            actions: ['select']
+          });
+          const clearButton = Dialog.warnButton({
+            label: trans.__('Clear Filter(s)'),
```
krassowski, line 1119: How hard is it to know if plural/singular should be used? This would help with translations because not every language can pluralise by adding a short suffix.

packages/notebook-extension/style/base.css
```diff
@@ -0,0 +1,11 @@
...
+| Copyright (c) Jupyter Development Team.
+| Distributed under the terms of the Modified BSD License.
+|----------------------------------------------------------------------------*/
+
+.jp-cell-filters-types,
+.jp-cell-filters-tags {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
```
krassowski, line 11: There is no value for user if double click selects the word "code"/"markdown"/"raw" in the label:

```suggestion
}

.jp-cell-filters-types {
  user-select: none;
}
```

packages/notebook-extension/src/index.tsx
```diff
@@ -1060,6 +1065,87 @@ const activeCellTool: JupyterFrontEndPlugin<void> = {
...
+            actions: ['clear']
+          });
+          const view = new CellFiltersView({ model, translator });
+          const result = await showDialog({
+            title: trans.__('Select cell filters:'),
+            body: view,
+            buttons: [Dialog.cancelButton(), selectButton, clearButton]
```
krassowski, line 1126: `clearButton` (destructive action) is the default; should it be changed so that cancel is the default? I can see the case for either, feel free to ignore if this was a deliberate decision.

krassowski (commented):
Some quick minor comments. Is it worth adding a single screenshot of the dialog to tests?

bollwyvl:
My concerns about `,{tag-name},` and `jp-cell-tags` have been addressed, thanks.

This is a good direction for a feature, but this is... very hard to do right. 

However, I'll pretty much always come down on the side of "don't add more dialogs that could preserve the interactive computing experience". 

If it's a settings-backed thing, things are more complicated, and I guess would get its own sidebar entirely. But as there is, by design, no generally-agreed upon tags for _anything_, I don't see how this would work as a global.

Alternately, assuming the state of filters is stored in notebook-level metadata, aside from probably needing to be supported by at least `nbconvert`, then a convention seen in other places would be for the directly-manipulable "universe" of tags to include a visibility indicator:

```
Cell  v

  Tags
  (existing tag 1 ☑)   (existing tag 2 ☐) ...

  ( [new tag] ⊞)

Notebook  v

  Enable Content Hiding ☑
     Hide Cells Parts
         [☐ code cells]  [☐ code cell source]  [☐ code cell output]   [☐ markdown cells]  [☐ raw cells]
      Hide Cells With Tag  
         (existing tag 1 👁 )    (existing tag 2 ☒)
```

Of course, someone will come along and "need" a _Hide **cells outputs** in cells **not** tagged with `my-good-stuff`_, but eventually you can't please everyone without a full query language like full CSS queries...

```css
#id-cell-id-123.jp-Cell:not([data-jp-cell-tags="my-good-stuff"]) .jp-OutputArea { display: none; }
```

... or, even _less_ humane, XPath, which will do things like 

```
//[contains(., "Traceback")]
```

which CSS (rightly) just _won't_.

The above raises another point: some libraries emit CSS on import (_blech_), so the above might "fail" for _other_ exciting reasons. So there might be another option of:
- "hide", roughly: `visibility: hidden; max-height:0; padding; 0; margin: 0;`
- "remove", roughly: `display: none`.
- "collapsed":  maybe not possible in pure CSS, haven't thought about it much

I went down this road a bit in `jupyterlab-fonts`, and decided I couldn't really do it justice, leaving it for another day.

Given either storage approach, whenever cells/parts are currently being hidden in this way, there should likely be some visual indicator (probably next to the _Cell Type_ toolbar dropdown) which gets highlighted (maybe with a sub-badge how many) when hiding is active, which opens the (non-dialog) tool when clicked.

The selection-buildign thing is.... a whole other thing, and would _definitely_ benefit from all of the direct manipulation stuff mentioned above: how many cells are selected, visual indicators, etc.

packages/notebook-extension/src/index.tsx
HaudinFlorence: I have chosen `:cell-filters` and `cellFilters` everywhere it is relevant.

packages/notebook-extension/src/index.tsx
HaudinFlorence: Sorry but how do you define or know that the `clearButton` is the default one ?

galata/test/documentation/export_notebook.test.ts-snapshots/exporting-slide-type-documentation-linux.png
HaudinFlorence: @krassowski I have made some css changes on the button. Here is what it now looks like

Screenshot from 2023-11-13 13-04-38

packages/notebook-extension/src/index.tsx
HaudinFlorence: @HaudinFlorence I have add for testing a counter for the number of active filters in the `CellFiltersComponent`. I guess that then using this counter it would be possible to set something like what is made for the commands linked to the buttons of notebook toolbars something similar to:
```ts
commands.addCommand(CommandIDs.runAndAdvance, {
    label: trans.__('Run Selected Cells'),
    caption: trans.__('Run the selected cells'),
    label: args => {
      const current = getCurrent(tracker, shell, args);
      return trans._n(
        'Run Selected Cell',
        'Run Selected Cells',
        current?.content.selectedCells.length ?? 1
      );
    },
    caption: args => {
      const current = getCurrent(tracker, shell, args);
      return trans._n(
        'Run this cell and advance',
        'Run these %1 cells and advance',
        current?.content.selectedCells.length ?? 1
      );
    },
    ```

**Comment by fcollonval** [comment]
Thanks a lot @bollwyvl for all the inputs. Regarding displaying the information, this PR is actually not hiding the cells but collapsing them for the reason you mentioned.

I propose to get this first version out. And then follow-up PRs could bring those information in the document for nbconvert and/or to switch to hide rather than collapsing the cells (behind a setting?).

File: packages/notebook-extension/src/index.tsx (hunk 13 above)
Comment (HaudinFlorence) [reply packages/notebook-extension/src/index.tsx]: I have put the cancel button as the last element of the list of buttons of the dialog.

Comment (krassowski) [reply packages/notebook-extension/src/index.tsx]: [`Dialog.IOptions` has `defaultButton`](https://jupyterlab.readthedocs.io/en/latest/api/interfaces/apputils.Dialog.IOptions.html#defaultButton) option for this purpose so maybe you could restore the order as it was before?

**Comment by krassowski** [comment]
> However, I'll pretty much always come down on the side of "don't add more dialogs that could preserve the interactive computing experience".

I agree (which is why I was pointing to placing it all in the ToC/property inspector sidebar). If we want to merge this as-is maybe we should tag dialog-related exported classes/functions as experimental to make it explicit to downstreams that it is subject to change in future?

More broadly some dedicate design work could help to find a good solution.

File: galata/test/jupyterlab/notebook-tags.test.ts [hunk 14]
```diff
@@ -0,0 +1,185 @@
...
+    await page.getByText('Python 3 (ipykernel) | Idle').waitFor();
+  });
+
+  test('should set data-jp-tags', async ({ page }) => {
+    const codesCells = page.locator('.jp-CodeCell');
+
+    await expect.soft(codesCells).toHaveCount(3);
+    await expect
+      .soft(codesCells.first())
+      .toHaveAttribute('data-jp-tags', 'banana');
```
fcollonval, line 27:
```suggestion
      .toHaveAttribute('data-jp-tags', ',banana,');
```

HaudinFlorence:
> Just to make it easy for other potential reviewers, the current state is that it still needs work around the integration tests as these are failing:
> 
> ```
>     [jupyterlab] › test/jupyterlab/notebook-tags.test.ts:58:7 › Notebook tags › should filter on tag 
>     [jupyterlab] › test/jupyterlab/notebook-tags.test.ts:94:7 › Notebook tags › should filter on cell type 
>     [jupyterlab] › test/jupyterlab/notebook-tags.test.ts:124:7 › Notebook tags › should filter on cell type and tag 
>     [jupyterlab] › test/jupyterlab/notebook-tags.test.ts:162:7 › Notebook tags › should clear all filters 
> ```

Ok thanks Mike @krassowski I can try to have a look at these failing tests.

HaudinFlorence:
> > However, I'll pretty much always come down on the side of "don't add more dialogs that could preserve the interactive computing experience".
> 
> I agree (which is why I was pointing to placing it all in the ToC/property inspector sidebar). If we want to merge this as-is maybe we should tag dialog-related exported classes/functions as experimental to make it explicit to downstreams that it is subject to change in future?
> 
> More broadly some dedicate design work could help to find a good solution.

@krassowski @fcollonval  Could you please elaborate more about what would be in your opinion the best solution without using a dialog? Having a button in the sidebar and open a popup instead of the dialog?

fcollonval:
Using a pop up approach will be easier because you won't have to listen to document changes events.

HaudinFlorence:
Hello. Back on this PR, I am wondering if something could be done and if maybe there is too much work for this PR to be accepted/this is not the right design? I guess that at least rebasing is the first step

krassowski:
> I guess that at least rebasing is the first step

FYI I will cut another pre-release today or tomorrow so versions will move again.

> there is too much work for this PR to be accepted

I had added this PR to the punchlist for 4.2 (https://github.com/jupyterlab/jupyterlab/issues/15801) a some time ago and no one raised an objection so I think that this feature is welcome.

> not the right design

I think adding this in a modal dialog for now is fine, although a popup would be preferable. I think that it should **not** be added to the notebook toolbar, especially not if we are still figuring out how this should behave. Having this as a section in the sidebar, or just as a button which opens popup or modal dialog works for the first iteration.

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
Hello. The PR has been rebased and all tests are now 🟢 except the Linux integrity3 one.
@krassowski @fcollonval Do you think the PR is ready for merging ?
cc: @afshin

krassowski:
Can you demo and discuss this on the Lab call today? We are past feature freeze for 4.2.0 by three weeks now and today was aspirational date for an RC release.

krassowski:
> except the Linux integrity3 one

Passed after restart. It looked odd, like a stuck runner.

HaudinFlorence:
> Can you demo and discuss this on the Lab call today? We are past feature freeze for 4.2.0 by three weeks now and today was aspirational date for an RC release.

Yes sure I can make a short demo at the today's Lab call.

krassowski:
Thank you for demoing this on the lab call today @HaudinFlorence. What do you think is the best course of action for shipping this feature to users who need it, and for ensuring it works the way users would expect it to work?

HaudinFlorence:
> Thank you for demoing this on the lab call today @HaudinFlorence. What do you think is the best course of action for shipping this feature to users who need it, and for ensuring it works the way users would expect it to work?

I don't really what is the best plan. The least effort would be to start with this PR and merge it, and then iterate from this, maybe replacing the dialog by the pop up. The two other options proposed if I understood correctly, could be to:
- make an extension
- move the cell collapser ui to the table of contents. 
At this stage, really I don't know what is the best option from the users' perspective.

krassowski:
> At this stage, really I don't know what is the best option from the users' perspective and user' experience.

In my experience it is easier to iterate on a feature by prototyping it in an extension first. This way you might be able to get user feedback early on, regardless of which JupyterLab version they are using (and it seems that this PR would only land in 4.3 given that we are past the feature freeze date, and that there were objections raised on the call). Are there any missing extension points that would make it hard for you to implementing this feature as an extension? If there are, then this would be an argument for merging this PR.

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
Bot please update snapshots

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
@krassowski I have introduced a minor change in the naming of the buttons in the dialog, to partially address comments about the term "filter" that is maybe not the appropriate to describe the actions on cells:
- `Filter Cells` => `Collapse Cells`
- `Clear Filters` => `Clear`
Screenshot from 2024-04-30 19-00-42

HaudinFlorence:
There are still 3 failing tests and 4 flaky ones.

jtpio:
Thanks @HaudinFlorence.

The failures don't seem to be related, but I just restarted the checks to make sure.

ericsnekbytes:
Binder wasn't working for me, I'll try a local build of this tomorrow.

ericsnekbytes:
Sorry I can't be more help, I've been trying to get this to build for a few hours without success so I'm going to leave it there (and return to finishing docs work). I think there's a more general problem for me just trying to get the normal build process to work on MacOs (I have apple silicon).

ericsnekbytes:
Okay, I was able to run this (found a workaround for my [issue](https://github.com/jupyterlab/jupyterlab/issues/16311)). It looks like a good first implementation for cell filtering, though I would suggest in the future adding an "expand" option alongside collapse. I also seem to have had an issue, where my cells were not collapsing (hopefully I'm not making an obvious mistake here):

foo

krassowski:
@ericsnekbytes what was your workaround? was this using Python 3.11?

HaudinFlorence:
> Okay, I was able to run this (found a workaround for my [issue](https://github.com/jupyterlab/jupyterlab/issues/16311)). It looks like a good first implementation for cell filtering, though I would suggest in the future adding an "expand" option alongside collapse. I also seem to have had an issue, where my cells were not collapsing (hopefully I'm not making an obvious mistake here):
> 
> ![foo](https://private-user-images.githubusercontent.com/104786633/329314200-7baa40e0-0c2c-46c6-b55b-dca19f5815b9.gif?

@ericsnekbytes Thanks for testing and for your feedback. Can you please share your notebook or maybe make some similar tests with this notebook `galata/test/jupyterlab/notebooks/tags.ipynb`? Thanks

HaudinFlorence:
@ericsnekbytes I tried to reproduce your notebook. I am also not sure that I am not missing important, but I don't get why if `bleu` is already tagged in the dialog, from a previous selection in the dialog, the other cells are not already collapsed. I am joining a gif to illustrate that point
/home/fhaudin/Videos/Screencasts/Screencast from 11-05-2024 16:10:01.webm/home/fhaudin/Videos/Screencasts/Screencast from 11-05-2024 16:10:01.webm

ericsnekbytes:
@krassowski Yes, I just downgraded to Python 3.10.

@HaudinFlorence I will try to reproduce it with the notebook you linked (and will attach if so).

ericsnekbytes:
@HaudinFlorence I'm experiencing another error when using the tags.ipynb notebook, shown here:


https://github.com/jupyterlab/jupyterlab/assets/104786633/4942d5f4-31f8-446d-ab4d-ef066a723157

HaudinFlorence:
> @HaudinFlorence I'm experiencing another error when using the tags.ipynb notebook, shown here:
> 
>  out.mp4
@ericsnekbytes Thanks for testing the notebook. You comment rises some interesting point. The behavior you reported in your gif is indeed the one expected. When you open the dialog, the form should initially be empty if no "filters" or collapsing actions has been applied previously : the check boxes are not showing the tags of the active cell but the filters that are active on the whole notebook. Maybe we could add in a future PR such a feature, showing the tags of the active cell too.

afshin:
Thank you @HaudinFlorence and @ericsnekbytes! So the initial implementation is to launch a dialog without any presets. And a follow-up user experience enhancement PR is to preset the dialog values to include the tags of the current cell (if any). How does this sound?

afshin:
I think the CI failures are upstream.

Screenshot 2024-05-16 at 15 05 43

HaudinFlorence:
bot please update snapshots

afshin:
bot please update snapshots

krassowski:
It looks like there are conflicts:

image

FYI, force-pushing so many commits makes it really hard to see the most recent discussion due how GitHub works

HaudinFlorence:
> It looks like there are conflicts:
> 
> image
> 
> FYI, force-pushing so many commits makes it really hard to see the most recent discussion due how GitHub works

@krassowski I totally agree that force pushing makes it difficult to see most recent elements of the discussion. If you have any suggestion to make it better and more convenient, please let me know. I will try to change this.

krassowski:
Well, I do not know of any mitigation other than using merge commits for resolving conflicts rather than using rebase + force-push (and I completely understand that sometimes force-push is more convenient).

HaudinFlorence:
@krassowski The PR is now up to date

ericsnekbytes:
@HaudinFlorence There's definitely a mismatch in expectations for me using this tool.

I was expecting to select tags in the popup dialog, then hit the "collapse cells" button to collapse cells that had that tag applied.

From what I understand now, it looks like the tool intends for you to select tags, and when applied, it will hide cells that do not have that exact combination of tags. I was not expecting the dialog to have any tags already highlighted: I expected them to be blank, then allow me to pick tags for a one-time filtering operation.

Some suggestions:

- The text "Collapse Cells" on the dialog button maybe add to the confusion about how the tool works. Maybe use "Apply Filter" instead?

- Both behaviors above are useful (and others too):
  - "Hide cells that have these tags"
  - "Show only cells that have these tags"
  - "Show only cells that have these tags and no others"
  - "Hide cells that have these tags and no others"
  - etc. (More below)

- So, all of these are different valid behaviors for a cell filtering tool (I roughly lump them into affirmative/negative buckets, with different matching criteria). Some checkboxes or options of some kind to control which type of filtering you want would be great. Maybe "Filter Mode:", with ["Hide Cells", "Show Cells"] options, and "Match Type:", with ["Cell Has Only These Tags", "Cell Has One or More", "Cell Has All or More"] options.

HaudinFlorence:
@ericsnekbytes Thanks your detailled inputs.

So if I understand correctly, you would have expected to select the tags directly in the dialog and not using the current `add tags` button in the common tools section of the right panel as it is now? Is is correct ? 
About the various suggestions:

- The text on the button was `Filter cells `but I recently changed the text since after some feedback it appears that using filtering may be confusing.
- About the various mode and options, yes maybe it would be worth to have a more complete tool, maybe in a follow up PR
- This also makes sense to have an expand all buttons.

HaudinFlorence:
> Just checking back @ericsnekbytes, any additional thoughts on [#15157 (comment)](https://github.com/jupyterlab/jupyterlab/pull/15157#issuecomment-2136978435)?

I am about to force push after rebase my local working branch. Sorry in advance for the huge list of commits...

HaudinFlorence:
> > Just checking back @ericsnekbytes, any additional thoughts on [#15157 (comment)](https://github.com/jupyterlab/jupyterlab/pull/15157#issuecomment-2136978435)?
> 
> @krassowski I am about to force push after rebasing my local working branch. Sorry in advance for the huge list of commits... I forgot about doing it another way. Please let me know if you want me to change something on this PR.

I am finally using git merge to avoid the aforementionned issues

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
@krassowski All tests are green except some documentation ui tests

ericsnekbytes:
@HaudinFlorence @krassowski I'm not sure what this part below means, can you clarify?

> ...you would have expected to select the tags directly in the dialog and not using the current add tags button in the common tools section of the right panel as it is now? Is is correct ?

I thought the "add tags button" in common tools was for adding a tag to a cell, and that the dialog was for selecting tags that make up the filter criteria. Does the "add tags button" do something else? Thanks.

HaudinFlorence:
> I thought the "add tags button" in common tools was for adding a tag to a cell, and that the dialog was for selecting tags that make up the filter criteria. Does the "add tags button" do something else? Thanks.

@ericsnekbytes That's totally correct. The "add tags button" doesn't do anything else, it has the same functionality as the one it has currently : adding various tags to various cells.

ericsnekbytes:
@HaudinFlorence Just to clarify this:

> ...you would have expected to select the tags directly in the dialog and not using the current add tags button in the common tools section of the right panel as it is now? Is is correct ?

I think you meant to imply that the initial dialog state should reflect the tags on the current cell (the checked tags in the common tools area). Is that right? I would indeed not expect that if so (would probably expect a button to populate from the current cell instead, or a checkbox to use tags from current cell).

HaudinFlorence:
> @HaudinFlorence Just to clarify this:
> 
> > ...you would have expected to select the tags directly in the dialog and not using the current add tags button in the common tools section of the right panel as it is now? Is is correct ?
> 
> I think you meant to imply that the initial dialog state should reflect the tags on the current cell (the checked tags in the common tools area). Is that right? I would indeed not expect that if so (would probably expect a button to populate from the current cell instead, or a checkbox to use tags from current cell).

No. I did not mean that. The dialog is initially free from any quotation at first opening (this is different if filters have already been applied in previous selections using the dialog). Sorry if it was unclear. On your side you initially said that you
were expecting to select tags in the popup dialog, then hit the "collapse cells" button to collapse cells that had that tag applied.
So your initial expectation is not fitting with the way the tool is currently working. Hoping this clarifies the tool behavior.

Considering these explanations and later additions of more options as you suggested, do you think this feature may be ok as it is as a first version?

ericsnekbytes:
@HaudinFlorence Yes we should add it in, this version is a good start for cell filtering. I'm merely trying to offer feedback in the hopes that it is helpful. Would you mind perhaps either briefly documenting how it works, perhaps in `docs/source/user/interface.rst` (I can help with this if needed) and/or maybe adding a short line of text at top of the dialog to explain how it works? Maybe something like:

- "Select tags below. Cells that do not have those tags will be collapsed".

Thanks for your work on this.

HaudinFlorence:
> @HaudinFlorence Yes we should add it in, this version is a good start for cell filtering. I'm merely trying to offer feedback in the hopes that it is helpful. Would you mind perhaps either briefly documenting how it works, perhaps in `docs/source/user/interface.rst` (I can help with this if needed) and/or maybe adding a short line of text at top of the dialog to explain how it works? Maybe something like:
> 
> * "Select tags below. Cells that do not have those tags will be collapsed".
> 
> Thanks for your work on this.

@ericsnekbytes Thanks for your reply. That's a good idea, I will try to add infos in`docs/source/user/interface.rst` and I also thought that some label on buttons or text in the dialog may help. I will propose something going in that direction

HaudinFlorence:
@ericsnekbytes @krassowski I added some explaining text in the dialog and some text in the documentation too (in `interface.rst`). Please tell me if looks ok to you. 

Apart from that, I observed kind of a bug, the filtering state reported in the dialog is not correct if the user make these operations:
- open a notebook
- apply some filtering
- save the notebook
- reopen it
- open the dialog. 

I would have expected to have the filters previously selected still checked. It is like the notebook does not have any filter applied and pushing `Clear` button does not have any effect to uncollapse cells.

HaudinFlorence:
Maybe we don't want to save the notebook with filters applied and respective cells collapses.

HaudinFlorence:
> Maybe we don't want to save the notebook with filters applied and respective cells collapses.

@krassowski @ericsnekbytes Do you have any expected behavior in mind?

brichet:
Thanks @HaudinFlorence for working on this.

My expectation of a filtering tool is to hide the other cells than the filtered ones, but here it collapses the non filtered only once.
The filter seems to be 'active' (the tags are still selected when opening again the popup) even if it is not relative to the current state of the notebook:
- removing a filtered tag from a cell is not filtered again
- opening a collapsed cell does not initialize the popup

Additionally, applying again the filter has no effect on the notebook, the filter need to be cleared and play again.

Peek 2024-08-02 11-40

My opinion to clarify the intention:
- restore the popup if a change related to the filter occurred on the notebook (one the filtered tag added or removed form a cell, one cell manually deployed or collapsed...)
- rename the `Clear filter` button to `Unfold all cells`

GabrielaVives:
Hello
I have dived into this issue with Florence in the last few days. I mainly tackled the UX and UI issues, not the technical ones about the issues when saving the notebook, manually expanding/collapsing cells, etc.

Here are a few insights I found:
- There is a confusion between filtering and selecting cells to collapse. From my understanding, the action performed by users is collapsing cells, which is not exactly the same as filtering them. Therefore, I think re-writing the labels and CTA buttons could improve the usability of this feature
- There is overall a lack of information about how many cells are displayed vs how many are collapsed. I found that there is not a lot of feedback to inform users what they have done and what is the state of the interface
- Manually opening a collapsed cell should not open the dialogue, but just update the number of collapsed cells displayed in the right panel

I worked on a very low-fidelity prototype to illustrate the first steps of improvement we could implement. All feedback is highly welcomed, there may be some use cases I missed. The prototype illustrates the following:

- Information provided on the right panel about how many cells are collapsed in the notebook
- The change from a "filter" action to a "selection" action
- Selecting cells in batches thanks to drop down per metadata category


https://github.com/user-attachments/assets/dd971a16-88c1-44b1-aad5-f9a5de590fcc



I will continue to cover the cases where users expand/collapse cells, and how this would interfere with the information displayed in the dialogue and in the side panel.

HaudinFlorence:
@GabrielaVives Thanks for the work and the prototype proposed. We would be happy to have some feedback on it. Feel free to share your thoughts on it
@afshin @fcollonval @ericsnekbytes @brichet 

> Thanks @HaudinFlorence for working on this.
> 
> My expectation of a filtering tool is to hide the other cells than the filtered ones, but here it collapses the non filtered only once. The filter seems to be 'active' (the tags are still selected when opening again the popup) even if it is not relative to the current state of the notebook:
> 
> * removing a filtered tag from a cell is not filtered again
> * opening a collapsed cell does not initialize the popup
> 
> Additionally, applying again the filter has no effect on the notebook, the filter need to be cleared and play again.
> 
> Peek 2024-08-02 11-40 [ Peek 2024-08-02 11-40 ](https://private-user-images.githubusercontent.com/32258950/354550606-28f9f3f6-a7eb-4299-8369-4334f2b962e1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ0MjAyMTMsIm5iZiI6MTcyNDQxOTkxMywicGF0aCI6Ii8zMjI1ODk1MC8zNTQ1NTA2MDYtMjhmOWYzZjYtYTdlYi00Mjk5LTgzNjktNDMzNGYyYjk2MmUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODIzVDEzMzE1M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZlNzBiMTAyNmI5YjA5YjZmNmM4NzViMzAyMzk4YmUxZDA1NzI4NjBmMmMwN2U1MzQyZTdkZGExYzY1YTZlNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.F9-neiFSD9hFChB80sdx1_R50Bl9XeMX65OzoGIe-uE) [ ](https://private-user-images.githubusercontent.com/32258950/354550606-28f9f3f6-a7eb-4299-8369-4334f2b962e1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ0MjAyMTMsIm5iZiI6MTcyNDQxOTkxMywicGF0aCI6Ii8zMjI1ODk1MC8zNTQ1NTA2MDYtMjhmOWYzZjYtYTdlYi00Mjk5LTgzNjktNDMzNGYyYjk2MmUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODIzVDEzMzE1M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZlNzBiMTAyNmI5YjA5YjZmNmM4NzViMzAyMzk4YmUxZDA1NzI4NjBmMmMwN2U1MzQyZTdkZGExYzY1YTZlNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.F9-neiFSD9hFChB80sdx1_R50Bl9XeMX65OzoGIe-uE)
> 
> My opinion to clarify the intention:
> 
> * restore the popup if a change related to the filter occurred on the notebook (one the filtered tag added or removed form a cell, one cell manually deployed or collapsed...)
> * rename the `Clear filter` button to `Unfold all cells`

@brichet Thank you very much for your review. This indeed highlights some need for improvements.

HaudinFlorence:
> Thanks @HaudinFlorence for working on this.
> 
> My expectation of a filtering tool is to hide the other cells than the filtered ones, but here it collapses the non filtered only once. The filter seems to be 'active' (the tags are still selected when opening again the popup) even if it is not relative to the current state of the notebook:
> 
> * removing a filtered tag from a cell is not filtered again
> * opening a collapsed cell does not initialize the popup
> 
> Additionally, applying again the filter has no effect on the notebook, the filter need to be cleared and play again.
> 
> Peek 2024-08-02 11-40 [ Peek 2024-08-02 11-40 ](https://private-user-images.githubusercontent.com/32258950/354550606-28f9f3f6-a7eb-4299-8369-4334f2b962e1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ0MjAyMTMsIm5iZiI6MTcyNDQxOTkxMywicGF0aCI6Ii8zMjI1ODk1MC8zNTQ1NTA2MDYtMjhmOWYzZjYtYTdlYi00Mjk5LTgzNjktNDMzNGYyYjk2MmUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODIzVDEzMzE1M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZlNzBiMTAyNmI5YjA5YjZmNmM4NzViMzAyMzk4YmUxZDA1NzI4NjBmMmMwN2U1MzQyZTdkZGExYzY1YTZlNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.F9-neiFSD9hFChB80sdx1_R50Bl9XeMX65OzoGIe-uE) [ ](https://private-user-images.githubusercontent.com/32258950/354550606-28f9f3f6-a7eb-4299-8369-4334f2b962e1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ0MjAyMTMsIm5iZiI6MTcyNDQxOTkxMywicGF0aCI6Ii8zMjI1ODk1MC8zNTQ1NTA2MDYtMjhmOWYzZjYtYTdlYi00Mjk5LTgzNjktNDMzNGYyYjk2MmUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODIzVDEzMzE1M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTZlNzBiMTAyNmI5YjA5YjZmNmM4NzViMzAyMzk4YmUxZDA1NzI4NjBmMmMwN2U1MzQyZTdkZGExYzY1YTZlNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.F9-neiFSD9hFChB80sdx1_R50Bl9XeMX65OzoGIe-uE)
> 
> My opinion to clarify the intention:
> 
> * restore the popup if a change related to the filter occurred on the notebook (one the filtered tag added or removed form a cell, one cell manually deployed or collapsed...)
> * rename the `Clear filter` button to `Unfold all cells`

@brichet Thank you very much Nicolas for your review. This indeed highlights some need for improvements.

HaudinFlorence:
> Hello I have dived into this issue with Florence in the last few days. I mainly tackled the UX and UI issues, not the technical ones about the issues when saving the notebook, manually expanding/collapsing cells, etc.
> 
> Here are a few insights I found:
> 
> * There is a confusion between filtering and selecting cells to collapse. From my understanding, the action performed by users is collapsing cells, which is not exactly the same as filtering them. Therefore, I think re-writing the labels and CTA buttons could improve the usability of this feature
> * There is overall a lack of information about how many cells are displayed vs how many are collapsed. I found that there is not a lot of feedback to inform users what they have done and what is the state of the interface
> * Manually opening a collapsed cell should not open the dialogue, but just update the number of collapsed cells displayed in the right panel
> 
> I worked on a very low-fidelity prototype to illustrate the first steps of improvement we could implement. All feedback is highly welcomed, there may be some use cases I missed. The prototype illustrates the following:
> 
> * Information provided on the right panel about how many cells are collapsed in the notebook
> * The change from a "filter" action to a "selection" action
> * Selecting cells in batches thanks to drop down per metadata category
> 
>  Screen.Recording.2024-08-23.at.11.39.04.mov 
> I will continue to cover the cases where users expand/collapse cells, and how this would interfere with the information displayed in the dialogue and in the side panel.

Thanks Gabriela @GabrielaVives for this work and for the prototype. It would be very helpful to have some feedback on it. Please feel free to add you comments. 
cc @afshin @fcollonval @ericsnekbytes @brichet

HaudinFlorence:
bot please udpate galata references

HaudinFlorence:
bot please update snapshots

krassowski:
@HaudinFlorence is this something that you wanted to merge in 4.6 or should we bump it to 4.7 given that the last time comment indicates need for some more feedback? There are also some conflicts but I think these should not be that hard to resolve.

HaudinFlorence:
@krassowski Thanks for your comment. I can try to merge with main. I think the PR should be bumped to 4.7. If we still want this feature to be in, there is still some work to be done.

PR #15160, opened 2023-09-26
Experimental inline completer

krassowski:
## References

Relates to #14267

## Code changes

- Adds a new `IInlineCompletionProvider` for extensions to provide inline completions (for example from various LLMs)
- Extends public `ICompletionManager` API with new methods:
  - `.registerInlineProvider(provider: IInlineCompletionProvider)`
  - `.invokeInline(id: string)`
  - `.cycleInline(id: string, direction: 'next' | 'previous')`
  - `.acceptInline(id: string)`
- Implements an example provider of inline completions using kernel history request

## User-facing changes

image

To check.

## To be done

- inline completion widget (hover box)
  - [ ] reduce space taken by the widget (UI/UX input welcome)
  - [ ] ensure it is placed above the line so that it never obstructs neither the 
  - [ ] add a setting to enable/disable it
  - [ ] decide if it should be opt-in or opt-out
- settings
  - [ ] setting to enable/disable inline completions globally (default on)
  - [ ] setting to enable/disable specific providers
    - [ ] make the history provider off by default (so no providers are there and effectively the feature is off)
- tests:
  - [ ] unit
  - [ ] integration
- [ ] document the new API
- [ ] rethink what shortcuts to use to allow implementing accepting partial suggestions in the future
- [ ] loading indicator when more than one provider
- [ ] implement streaming support
- [ ] improve history provider

### DIFF
--- a/packages/completer-extension/src/index.ts
+++ b/packages/completer-extension/src/index.ts
@@ -21,10 +21,14 @@ import {
 } from '@jupyterlab/completer';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import {
+  caretLeftIcon,
+  caretRightIcon,
+  checkIcon,
   IFormRenderer,
   IFormRendererRegistry
 } from '@jupyterlab/ui-components';
 import type { FieldProps } from '@rjsf/utils';
+import { CommandRegistry } from '@lumino/commands';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 
 import { renderAvailableProviders } from './renderer';
@@ -34,6 +38,7 @@ const COMPLETION_MANAGER_PLUGIN = '@jupyterlab/completer-extension:manager';
 namespace CommandIDs {
   export const nextInline = 'inline-completer:next';
   export const previousInline = 'inline-completer:previous';
+  export const acceptInline = 'inline-completer:accept';
 }
 
 const defaultProviders: JupyterFrontEndPlugin<void> = {
@@ -80,24 +85,46 @@ const inlineCompleter: JupyterFrontEndPlugin<IInlineCompleterFactory> = {
     app: JupyterFrontEnd,
     translator: ITranslator | null
   ): IInlineCompleterFactory => {
-    //const trans = (translator || nullTranslator).load('jupyterlab');
+    const trans = (translator || nullTranslator).load('jupyterlab');
     return {
       factory: options => {
         const inlineCompleter = new InlineCompleter(options);
+        const describeShortcut = (commandID: string): string => {
+          const binding = app.commands.keyBindings.find(
+            binding => binding.command === commandID
+          );
+          const keys = binding
+            ? CommandRegistry.formatKeystroke(binding.keys)
+            : '';
+          return keys ? ` (${keys})` : '';
+        };
         inlineCompleter.toolbar.addItem(
-          'previous',
+          'previous-inline-completion',
           new CommandToolbarButton({
             commands: app.commands,
-            id: CommandIDs.previousInline
-            //icon: ,
-            //label: trans.__('Previous (shortcut)')
+            id: CommandIDs.previousInline,
+            icon: caretLeftIcon,
+            label:
+              trans.__('Previous') + describeShortcut(CommandIDs.previousInline)
           })
         );
         inlineCompleter.toolbar.addItem(
-          'next',
+          'next-inline-completion',
           new CommandToolbarButton({
             commands: app.commands,
-            id: CommandIDs.nextInline
+            icon: caretRightIcon,
+            id: CommandIDs.nextInline,
+            label: trans.__('Next') + describeShortcut(CommandIDs.nextInline)
+          })
+        );
+        inlineCompleter.toolbar.addItem(
+          'accept-inline-completion',
+          new CommandToolbarButton({
+            commands: app.commands,
+            icon: checkIcon,
+            id: CommandIDs.acceptInline,
+            label:
+              trans.__('Accept') + describeShortcut(CommandIDs.acceptInline)
           })
         );
         return inlineCompleter;
@@ -122,14 +149,22 @@ const inlineCompleterCommands: JupyterFrontEndPlugin<void> = {
       execute: () => {
         completionManager.cycleInline(app.shell.currentWidget!.id!, 'next');
       },
-      label: () => trans.__('Next'),
+      label: trans.__('Next Inline Completion'),
       isEnabled: () => !!app.shell.currentWidget
     });
     app.commands.addCommand(CommandIDs.previousInline, {
       execute: () => {
         completionManager.cycleInline(app.shell.currentWidget!.id!, 'previous');
       },
-      label: () => trans.__('Previous'),
+      label: trans.__('Previous Inline Completion'),
+      isEnabled: () => !!app.shell.currentWidget
+    });
+    app.commands.addCommand(CommandIDs.acceptInline, {
+      execute: () => {
+        // TODO
+        completionManager.cycleInline(app.shell.currentWidget!.id!, 'previous');
+      },
+      label: trans.__('Accept Inline Completion'),
       isEnabled: () => !!app.shell.currentWidget
     });
   }
--- a/packages/completer-extension/src/index.ts
+++ b/packages/completer-extension/src/index.ts
@@ -12,6 +12,7 @@ import {
 import {
   CompletionProviderManager,
   ContextCompleterProvider,
+  HistoryInlineCompletionProvider,
   ICompletionProviderManager,
   KernelCompleterProvider
 } from '@jupyterlab/completer';
@@ -21,12 +22,13 @@ import {
   IFormRendererRegistry
 } from '@jupyterlab/ui-components';
 import type { FieldProps } from '@rjsf/utils';
+import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 
 import { renderAvailableProviders } from './renderer';
 
 const COMPLETION_MANAGER_PLUGIN = '@jupyterlab/completer-extension:manager';
 
-const defaultProvider: JupyterFrontEndPlugin<void> = {
+const defaultProviders: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlab/completer-extension:base-service',
   description: 'Adds context and kernel completion providers.',
   requires: [ICompletionProviderManager],
@@ -37,6 +39,26 @@ const defaultProvider: JupyterFrontEndPlugin<void> = {
   ): void => {
     completionManager.registerProvider(new ContextCompleterProvider());
     completionManager.registerProvider(new KernelCompleterProvider());
+  }
+};
+
+const inlineHistoryProvider: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/completer-extension:inline-history',
+  description:
+    'Adds inline completion provider suggesting code from execution history.',
+  requires: [ICompletionProviderManager],
+  optional: [ITranslator],
+  autoStart: true,
+  activate: (
+    app: JupyterFrontEnd,
+    completionManager: ICompletionProviderManager,
+    translator: ITranslator | null
+  ): void => {
+    completionManager.registerInlineProvider(
+      new HistoryInlineCompletionProvider({
+        translator: translator ?? nullTranslator
+      })
+    );
   }
 };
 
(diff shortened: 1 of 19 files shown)


fcollonval:
> better default icon/icon for the history provider

What about https://pictogrammers.com/library/mdi/icon/history/

packages/completer-extension/src/index.ts
```diff
@@ -40,23 +55,140 @@ const defaultProvider: JupyterFrontEndPlugin<void> = {
...
+            label: trans.__('Accept')
+          })
+        );
+        return inlineCompleter;
+      }
+    };
+  }
+};
+
+const inlineCompleterCommands: JupyterFrontEndPlugin<void> = {
```
fcollonval, line 128: What is the reason to split this out of the previous one?

I see a risk of inconsistency if the `@jupyterlab/completer-extension:inline-completer-commands` is disabled or replace by another set of command ids.

packages/completer/src/default/inlinehistoryprovider.ts
```diff
@@ -0,0 +1,88 @@
...
+  TranslationBundle
+} from '@jupyterlab/translation';
+
+/**
+ * An example inline completion provider using history to populate suggestions.
+ */
+export class HistoryInlineCompletionProvider
+  implements IInlineCompletionProvider
+{
+  identifier = 'history';
```
fcollonval, line 22: Should it be static? read only?

```suggestion
  static readonly identifier = 'history';
```

packages/completer/src/ghost.ts
```diff
@@ -0,0 +1,324 @@
...
+   * Clear all ghost texts from the editor.
+   */
+  clearGhosts(view: EditorView) {
+    const effects: StateEffect<unknown>[] = [Private.removeMark.of(null)];
+    view.dispatch({ effects });
+  }
+}
+
+class GhostTextWidget extends WidgetType {
+  constructor(readonly content: string) {
```
fcollonval, line 64: I learnt something :wink:

packages/completer/src/ghost.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/**
+ * Spacers are used to reduce height jitter in the transition between multi-line inline suggestions.
+ * In particular, when user removes a letter they will often get a new suggestion in split-second,
+ * but without spacer they would see the editor collapse in height and then elongate again.
+ */
+class TransientSpacerWidget extends GhostTextWidget {
+  // no-op
+}
```
fcollonval, line 96: Is this needed as it does not bring any feature?

packages/completer/src/handler.ts
```diff
@@ -376,6 +417,64 @@ export class CompletionHandler implements IDisposable {
...
+    if (!model) {
+      return;
+    }
+    model.cursor = position;
+
+    const current = ++this._fetchingInline;
+    const promises = this._reconciliator.fetchInline(request, trigger);
+
+    let i = 0;
+    for (const promise of promises) {
```
fcollonval, line 451: following on the linter comment

I guess it will be better to do something like

```ts
await Promise.all(promises.map(async p => {
  const result = await p;
  /* ... */
}));

File: packages/completer/src/inline.ts [hunk 6]
```diff
@@ -0,0 +1,551 @@
...
+    /**
+     * A signal emitted when filter text is updated.
+     * Emits a mapping from old to new index for items after filtering.
+     */
+    readonly filterTextChanged: ISignal<IModel, IndexMap>;
+
+    /**
+     * Original placement of cursor
+     */
+    cursor: CodeEditor.IPosition;
```
fcollonval, line 409: Should this be readonly?

```suggestion
    readonly cursor: CodeEditor.IPosition;
```

packages/completer/src/manager.ts
```diff
@@ -197,6 +223,46 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     }
   }
 
+  /**
```
fcollonval, line 226: I don't see any issues with the new API. Only a question to keep a trace, do we want arguments list like the current proposal or an arguments object?

I'm mainly asking especially for the `cycleInline` in light of the mistake I did when refactoring the search.

packages/completer/src/tokens.ts
```diff
@@ -138,6 +138,131 @@ export interface ICompletionProvider<
   ): boolean;
 }
 
+/**
+ * Describes how an inline completion provider was triggered.
+ * @alpha
```
fcollonval, line 143: should we use @alpha or @experimental to align with the new API ts doc?

packages/completer/src/tokens.ts
```diff
@@ -138,6 +138,131 @@ export interface ICompletionProvider<
...
+> {
+  fetch(
+    request: CompletionHandler.IRequest,
+    context: IInlineCompletionContext
+  ): Promise<IInlineCompletionList<T>>;
+
+  /**
+   * Name of the provider to be displayed in the UI.
+   */
+  name: string;
```
fcollonval, line 237: I guess it is set at instantiation:

```suggestion
  readonly name: string;
```

packages/completer/src/tokens.ts
```diff
@@ -138,6 +138,131 @@ export interface ICompletionProvider<
...
+
+  /**
+   * Unique identifier, cannot change on runtime.
+   */
+  identifier: string;
```
fcollonval, line 242: To enforce the doc mention?
```suggestion
  readonly identifier: string;
```

packages/completer/src/tokens.ts
```diff
@@ -138,6 +138,131 @@ export interface ICompletionProvider<
...
+    done: boolean;
+    response: T;
+  };
+}
+
+/**
+ * Inline completer factory
+ */
+export interface IInlineCompleterFactory {
+  factory(options: Omit<InlineCompleter.IOptions, 'trans'>): InlineCompleter;
```
fcollonval, line 255: Could we invert the interface from filtering to extension?

Because if some extensions provide their own implementation, they will have to use that `Omit` typing. So better creating a base one and extending it for `InlineCompleter.IOptions`.

packages/completer/src/handler.ts
```diff
@@ -517,6 +622,13 @@ export namespace CompletionHandler {
     items: Array<T>;
   }
 
+  export interface IInlineItem extends IInlineCompletionItem {
+    /**
+     * The source provider.
+     */
+    provider: Omit<IInlineCompletionProvider, 'fetch'>;
```
fcollonval, line 629: Would you be able easily to avoid using `Omit`?

fcollonval (commented):
Thanks @krassowski 

Here are my questions/suggestions on the API.

packages/completer-extension/src/index.ts
krassowski: I am not happy with this either.

The reason is dependency chain of `inline-completer-commands` → `ICompletionProviderManager` → `IInlineCompleterFactory`. If we merge the factory with commands we have a cycle.

On the positive side, if downstream wants to override the hoverbox widget, they can just replace `IInlineCompleterFactory` and the commands on the toolbar would still be populated without them needing to cop-ypaste it so maybe it is a good thing?

packages/completer/src/tokens.ts
krassowski: Done in c4577e18ebbe2f46f0f3da32507661e61e895290.

packages/completer/src/handler.ts
krassowski: I had addressed the linter comment in d93acdb22b9742b1f409d45bf8adf7296303f5ce. I feel that keeping hanging promise here makes it clear that we terminate the promise chain here and the rest happens in callbacks.

packages/completer/src/manager.ts
krassowski: Good point, there is a place in the `IModel` where it seems more relevant (but then it is further away from the extension point). I do not see it for `cycleInline` specifically, can you elaborate? The future possibilities for extension that I see could be `direction: 'first' | 'last'`.

krassowski (commented):
Thank you for taking a look @fcollonval, hugely appreciated!

packages/completer-extension/src/index.ts
krassowski: The dependency chain is now removed in 00a7f2b76fde148d3f045643632a61ddb82cbf89, the commands are now in `inline-completer` and factory is now in `inline-completer-factory`. This means that the commands are now guaranteed to be present if the inline completer is enabled, while still preserving the option to change the factory.

krassowski:
Two API questions came up when testing this with an LLM implementation of a provider:
- a) where to place debouncer/timeout: in providers or in core?
- b) should we expose `language` and cell type in context? (this can be in follow-up)

Details:
- a) implementing the debouncer setting in providers is problematic because when the provider's debouncer delay is longer than the timeout configured in core, then the provider will time out. We could solve it by either:
   - having the debouncer setting in core, or
   - allowing providers to manage their own timeout
   
  the argument for having per-provider debouncer is that users may want to set different thresholds for different providers, e.g. the history provider always knows the answer in an instant and is free, an local provider may be free but slower/making increasing the fan noise, and remote API providers may have $$ costs associated with queries. There is somewhat weaker argument for having a provider-specific timeout.
- b) a different model may need to be queried for different cell types and language; cell type and language is easy to get, unless there are transclusions in which case we need to expose CodeMirror's [EditorState.languageDataAt](https://codemirror.net/docs/ref/#state.EditorState.languageDataAt). transclusions, have effect in both code (e.g. `%%typesscript`) and markdown cells :
    ~~~markdown
   this is **markdown**, inline completer should suggest human language

    ```python
    print("this is code")   # inline completer should suggest code 
    ```
    ~~~

Wzixiao:
Hello @krassowski ,

After integrating my code with yours, I'm pleased to find that they are quite compatible. However, I have a couple of questions I'd like to discuss with you:

I noticed in the completer-extension package of your code, there are several commands defined. Specifically, the CommandIDs.invokeInline seems to execute the accept function. Is this intentional?

Would it be possible for us to add a configuration for auto-completion? I observed that it always auto-completes, but sometimes it's not what I intended. Could we consider turning off this feature and manually execute invoke instead?

Thank you for your assistance and understanding.

krassowski:
Thanks for spotting it @Wzixiao, no it was not intentional and was fixed in cf932de201c76290ac1fadf42c9574aa5c5246d8.

krassowski:

packages/completer/src/tokens.ts
fcollonval: If I recall correctly, this is exactly what is mentioned in [tsdoc](https://tsdoc.org/pages/tags/experimental/):

> Same semantics as @\beta, but used by tools that don't support an @\alpha release stage.

Some tools have _experimental_ and not _alpha_/_beta_, other the other way around (like [API extractor](https://api-extractor.com/pages/tsdoc)). I think _experimental_ was even the only one supported by tsdoc in the past.

dongs0104:
Hello @krassowski In this case, I don't get additional information like type of documentation from the inline completer like I do from the completer widget, is this intended behavior?
The current LLM model is more generative when you give it a lot of context, so it would be nice to get more information.
Can i Parse using `context` on it?

packages/completer/src/tokens.ts
krassowski: Would you prefer me to change the existing tag comments to:

```
    * @alpha 
    * @experimental 
```

or are they good as they are?

krassowski:
If there are no further comments, it would be helpful to merge and cut a pre-release to allow wider testing (setting up a binder with [krassowski/jupyterlab-transformers-completer](https://github.com/krassowski/jupyterlab-transformers-completer) is proving to be difficult due to Binder runners restrictions).

packages/completer/src/manager.ts
fcollonval: I was not seeing a specific case for `cycleInline`. It was more to be sure to think a second times about that; as you did :wink:

packages/completer/src/tokens.ts
fcollonval: Let's keep it like this as `alpha` / `beta` is clearer than `experimental` in term of API stability expectation.

packages/lsp/src/connection.ts
```diff
@@ -120,6 +120,7 @@ class ServerRequestHandler<
 export const Provider: { [key: string]: keyof lsp.ServerCapabilities } = {
   TEXT_DOCUMENT_SYNC: 'textDocumentSync',
   COMPLETION: 'completionProvider',
+  //INLINE_COMPLETION: 'inlineCompletionProvider',
```
fcollonval, line 123: Should we remove this dead case?

```suggestion
```

fcollonval (changes requested):
Hey @krassowski 

I only have one comment about the icon for the settings. It seems they are not using `jp-icon` classes and therefore may not be styled properly depending on the theme. For example in dark mode:

image

packages/lsp/src/connection.ts
krassowski: Yes, this will be added in a future PR once the LSP API stabilises.

fcollonval (approved):
Thanks @krassowski

PR #14842, opened 2023-07-18
Add page option `copyAbsolutePath` enabling to copy absolute path

pauky:
## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/9485
Relates jupyterlab_server https://github.com/jupyterlab/jupyterlab_server/pull/405

## Code changes
add page option for copy absolute path, which can replace copy path that is relative.

## User-facing changes
None

## Backwards-incompatible changes
Obtain absolute paths instead of relative paths when copying paths in filebrowser.

PR #15168, opened 2023-09-28
Copy absolute path with first slash in the file browser

pauky:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/14842
There is no first slash when coping a absolute path.

## Code changes

Change `PathExt.json` to `posix.json`.

## User-facing changes

None.

None.

---

Edit to specify how it relates to 14842

### DIFF
--- a/packages/filebrowser-extension/src/index.ts
+++ b/packages/filebrowser-extension/src/index.ts
@@ -65,7 +65,6 @@ import {
 import { find, map } from '@lumino/algorithm';
 import { CommandRegistry } from '@lumino/commands';
 import { ContextMenu } from '@lumino/widgets';
-import { posix } from 'path';
 
 const FILE_BROWSER_FACTORY = 'FileBrowser';
 const FILE_BROWSER_PLUGIN_ID = '@jupyterlab/filebrowser-extension:browser';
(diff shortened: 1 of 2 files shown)


packages/filebrowser-extension/src/index.ts
```diff
@@ -65,6 +65,7 @@ import {
 import { find, map } from '@lumino/algorithm';
 import { CommandRegistry } from '@lumino/commands';
 import { ContextMenu } from '@lumino/widgets';
+import { posix } from 'path';
```
krassowski, line 68: This will not work as the code is running in the browser, not in a Node runtime.

yangrw1: Function `PathExt.join` was wrapped by it, and it can be execed in browser.

krassowski: No, it cannot be used in browser because browsers do not implement Node runtime. You would need to add a dependency on `path-browserify`, which is what the `@jupyterlab/coreutils` package which implements `PathExt.join` does:

https://github.com/jupyterlab/jupyterlab/blob/3c3c65ae03a758c620d8dd19678cb891d6bf0f20/packages/coreutils/package.json#L47

However, ideally this implementation detail and dependency would not be propagated to filebrowser-extension, and either a new argument would be added to one of `PathExt` methods, or a new method created altogether and used in here.

fcollonval (approved):
thanks @pauky

Issue #15169, opened 2023-09-28
"Help > Show Keyboard Shortcuts" gives `Cannot read properties of undefined (reading 'node')`


jtpio:
## Description

On the file browser page In Notebook 7, opening the list of keyboard shortcuts with `Help > Show Keyboard Shortcuts` does not work and gives the following error in the dev tools console:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'node')
```

## Reproduce

Open Notebook 7, for example on Binder: https://gist.github.com/jtpio/d368ab89cee5123ecee60683115e15f3

https://github.com/jupyterlab/jupyterlab/assets/591645/50edfb22-80e7-4451-8b45-175134899ba5

The full error is the following:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/ff751bae-330d-4e57-a526-3b5c8843111d)

## Expected behavior

It should open the keyboard shortcuts dialog.

## Context

Also noticed in https://github.com/jupyterlite/jupyterlite/pull/1170

jtpio:
Looks like this might be because of the following line, which assumes `app.shell.currentWidget` is a `MainAreaWidget` and tries to access its content:

https://github.com/jupyterlab/jupyterlab/blob/1692817b23ce35a31d02889c4af83df732c00a17/packages/apputils-extension/src/index.ts#L655

PR #15170, opened 2023-09-28
Fix Show Keyboard Shortcuts command

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15169

## Code changes

Handle the case where `app.shell.currentWidget` is not a `MainAreaWidget`. Like the file browser widget in Notebook 7 for example.

## User-facing changes

This should allow Notebook 7 users to open the dialog on the file browser page.

None

### DIFF
--- a/packages/apputils-extension/src/index.ts
+++ b/packages/apputils-extension/src/index.ts
@@ -650,8 +650,7 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
         const included = currentWidget?.node.contains(document.activeElement);
 
         if (!included && currentWidget instanceof MainAreaWidget) {
-          const currentNode =
-            currentWidget.content.node ?? currentWidget?.node;
+          const currentNode = currentWidget.content.node ?? currentWidget?.node;
           currentNode?.focus();
         }
         const options = { commands, trans };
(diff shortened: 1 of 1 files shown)


packages/apputils-extension/src/index.ts
```diff
@@ -646,14 +646,12 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
...
-        );
+        const currentWidget = app.shell.currentWidget;
+        const included = currentWidget?.node.contains(document.activeElement);
 
-        if (!included) {
+        if (!included && currentWidget instanceof MainAreaWidget) {
           const currentNode =
-            (app.shell.currentWidget as MainAreaWidget)?.content.node ??
-            app.shell.currentWidget?.node;
+            currentWidget.content.node ?? app.shell.currentWidget?.node;
```
krassowski, line 654: Since the `currentWidget` is now declared (improving clarity, thank you!), can we use in this line too?

```suggestion
            currentWidget.content.node ?? currentWidget?.node;
```

jtpio: Good catch, thanks!

krassowski (approved):
Thanks @jtpio!

krassowski:

jtpio:
Thanks for the review!

PR #15175, opened 2023-09-28
Restore syntax highlighting for mimetypes with more than one identifier

### LINKED ISSUE #15172: .ipynb files have no syntax highlighting in texteditor
## Description

Whenever I open .ipynb files in the jupyterlab texteditor, no syntax highlighting is available.
This is also true after changing the filetype selector in the bottom left from 'none' to 'JSON'.

## Reproduce

1. Rightclick any .ipynb in jupyterlab and select open with -> editor
2. (optional) Click on 'none' in the bottom left and choose JSON

## Expected behavior

I expected to see JSON syntax highlighting when editing .ipynb files.

## Context

- Operating System and version: Archlinux
- Browser and version: any
- JupyterLab version: 4.0.6

jans-code:
## References

Fixes #15172

## Code changes

Checking if 'spec.mime' is an array, and if it is using the first string for 'widget.content.model.mimeType', else business as usual.

## User-facing changes

n/a

n/a

---

Edit to link the issue.


packages/fileeditor-extension/src/commands.ts
```diff
@@ -567,7 +567,11 @@ export namespace Commands {
         if (name && widget) {
           const spec = languages.findByName(name);
           if (spec) {
-            widget.content.model.mimeType = spec.mime as string;
+            if (Array.isArray(spec.mime)) {
+              widget.content.model.mimeType = spec.mime[0] as string;
+            } else {
+              widget.content.model.mimeType = spec.mime as string;
```
krassowski, line 573: Is the casting here needed? (genuinely asking)

```suggestion
              widget.content.model.mimeType = spec.mime;
```

If it is not an array it should be a `string` as per:

https://github.com/jupyterlab/jupyterlab/blob/bd46c76046671ae375b3ddd207ca441841e57123/packages/codemirror/src/token.ts#L270

packages/fileeditor-extension/src/commands.ts
```diff
@@ -567,7 +567,11 @@ export namespace Commands {
         if (name && widget) {
           const spec = languages.findByName(name);
           if (spec) {
-            widget.content.model.mimeType = spec.mime as string;
+            if (Array.isArray(spec.mime)) {
+              widget.content.model.mimeType = spec.mime[0] as string;
```
krassowski, line 571: I think having an empty array would be unlikely but if it happens we could fallback to `text/plain`. What do you think?

packages/fileeditor-extension/src/commands.ts
jans-code: while this might be unnecessary, vscode does show a TS2322 error without the casting.
i would not want that distractive red error line in there :)

packages/fileeditor-extension/src/commands.ts
jans-code: seems totally reasonable.

packages/fileeditor-extension/src/commands.ts
krassowski: Right, I think that in typescript marking it as `const` ahead would make the compiler infer types correctly:

```ts
const mime = spec.mime;
if (Array.isArray(mime)) {
   // `mime` is now inferred to be an array
} else {
    // `mime` is now inferred to be a string
}
```

see [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgJIBUIGczIN4CwAUMqcgLbDkQBcyOUoA5sgD71iMhMDaAugG5iAX2IIA9iBzJgyALz5hyOFjSYcQomMnTK1eTIB0eiMWAwAFAEEoUOAE9DwLDbv2LJgJSf8xMsgkpXCgIADcIOAAbdHsAB1oOLl4+AxNNJQhIrBRCEjJA6RDwqJj4ugZmVKoIdKA)

packages/fileeditor-extension/src/commands.ts
```diff
@@ -567,7 +567,15 @@ export namespace Commands {
         if (name && widget) {
           const spec = languages.findByName(name);
           if (spec) {
-            widget.content.model.mimeType = spec.mime as string;
+            if (Array.isArray(spec.mime)) {
+              if (spec.mime.length == 0) {
+                widget.content.model.mimeType = 'text/plain';
+              } else {
+                widget.content.model.mimeType = spec.mime[0] as string;
+              }
```
fcollonval, line 575: We can be more condense for the fallback. And there is a variable to avoid hard coding `text/plain`.

```suggestion
              widget.content.model.mimeType = spec.mime[0] as string ?? IEditorMimeTypeService.defaultMimeType;
```

fcollonval (changes requested):
Thanks @jans-code 

I have a minor improvement but otherwise the logic is good.

packages/fileeditor-extension/src/commands.ts
jans-code: Much nicer indeed!

packages/fileeditor-extension/src/commands.ts
jans-code: i see, but we only want strings here for mimeType so i guess leaving the definition at 'mimeType: string;' feels like the right thing to do?

krassowski: Just checking if you saw my response here @jans-code

jans-code: now i have :)

the if case does work without casting the string (also without a const declaration beforehand), but the else still shows the ts2322.

so this would be as tidy as i can get it:

```typescript
if (Array.isArray(spec.mime)) {
              widget.content.model.mimeType = spec.mime[0] ?? IEditorMimeTypeService.defaultMimeType;
            } else {
              widget.content.model.mimeType = spec.mime as string;
            }
```

krassowski (approved):
Thank you @jans-code!

fcollonval:

PR #15182, opened 2023-09-29
Remove hard coded text-plain in multiple places

fcollonval:
## References

Follow-up of https://github.com/jupyterlab/jupyterlab/pull/15175

I searched for hard-coded places of `text/plain` and remove those not making sense (reviewers please don't hesitate to challenge that statement)

## Code changes

Remove hard-coded `text/plain` by the default mime type constant.

## User-facing changes

None

None

### DIFF
--- a/packages/codemirror/src/language.ts
+++ b/packages/codemirror/src/language.ts
@@ -27,7 +27,7 @@ export class EditorLanguageRegistry implements IEditorLanguageRegistry {
     // Add default language text/plain -> No expressions to parse
     this.addLanguage({
       name: 'none',
-      mime: 'text/plain',
+      mime: IEditorMimeTypeService.defaultMimeType,
       support: new LanguageSupport(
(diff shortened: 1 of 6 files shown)


packages/codemirror/src/language.ts
```diff
@@ -27,7 +27,7 @@ export class EditorLanguageRegistry implements IEditorLanguageRegistry {
     // Add default language text/plain -> No expressions to parse
     this.addLanguage({
       name: 'none',
-      mime: 'text/plain',
+      mime: IEditorMimeTypeService.defaultMimeType,
```
krassowski, line 30: I am not sure about this one. It could have unexpected consequences if someone changes `defaultMimeType` downstream.

fcollonval, line 30:
```suggestion
      mime: 'text/plain',
```

krassowski (approved):
Thanks @fcollonval!

PR #15184, opened 2023-09-30
Fix clicking in the TOC does not scroll

parmentelat:
workaround for cases where headingToElement map is not properly filled

## References

fixes #14591  
intended to supersede #14884

## Code changes

as suggested by @fcollonval in this post
https://github.com/jupyterlab/jupyterlab/pull/14884#pullrequestreview-1568943032

the `onActiveHeadingChanged` function (notebook viewer)
uses a fallback strategy in pathological cases where `headingToElement` map is not properly filled

## User-facing changes

no visual change; clicking in a TOC entry should more reliably scroll to the corresponding cell

## Backwards-incompatible changes

none

Issue #14591, opened 2023-05-24
Table of Contents: selecting heading only scrolls to position the first time


minerharry:
## Description

When you select a heading in the table of contents, it automatically starts editing the cell. This is not preferred behavior for me, but I can live with it. More problematic, however, is that once the cell is in edit mode, clicking the header again in the ToC will not scroll to the cell. This happens even if you have currently selected/are editing another code or markdown cell, and can be true for multiple headers at the same time. Effectively, you could have an entire table of contents and none of the headers work!

## Reproduce

In the example notebook below: scroll up so that the heading h1 is offscreen. Click h1 in the table of contents, you get scrolled down to it and the cell enters edit mode. Scroll up and optionally select one of the filler cells; clicking the heading again in the table of contents does nothing. Only fixed when the markdown cell is run again.

## Expected behavior

When you click a heading cell in the table of contents, you should always be scrolled to it. In my opinion, it also shouldn't automatically start editing markdown cells, but that's of a secondary concern.

## Context

example notebook contents:
```json
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": 3,
   "id": "1733768e-510c-4fef-b5b9-b086fd969d44",
   "metadata": {},
   "outputs": [],
   "source": [
    "filler = 4"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": 4,
   "id": "9d81f6ec-cbd8-49e7-908f-f315b26de2d3",
   "metadata": {},
   "outputs": [
    {
     "data": {
      "text/plain": [
       "4"
      ]
     },
     "execution_count": 4,
     "metadata": {},
     "output_type": "execute_result"
    }
   ],
   "source": [
    "filler"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": 5,
   "id": "262057a5-f605-4e23-bf26-f0035189fd8c",
   "metadata": {},
   "outputs": [
    {
     "data": {
      "text/plain": [
       "4"
      ]
     },
     "execution_count": 5,
     "metadata": {},
     "output_type": "execute_result"
    }
   ],
   "source": [
    "filler"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": 6,
   "id": "2f2d07fc-3bb6-4464-8dda-453d74d5766c",
   "metadata": {},
   "outputs": [
    {
     "data": {
      "text/plain": [
       "4"
      ]
     },
     "execution_count": 6,
     "metadata": {},
     "output_type": "execute_result"
[...]

PR #15195, opened 2023-10-01
new ui test: clicking the TOC should scroll the notebook panel

parmentelat:
## References

as suggested by @krassowski in
https://github.com/jupyterlab/jupyterlab/pull/15184
here's a ui test to check that clicking a header in the TOC does scroll to ensure that heading becomes visible

this PR should help reproduce issue #14591

## Code changes

only a new ui test is added:
- opening a reasonably long notebook
- clicking on the last header
- show result in the nb viewport having scrolled to near the bottom of the notebook

**note** that the expected output (the snapshot) needs more work because
- my laptop is a mac so I could produce only the `darwin` version of the snapshot, my understanding is that a `linux` version is desirable too
- more importantly, as of this writing, the jupyterlab code breaks this test, so the snapshot in this PR is just an indication of what it should become

## User-facing changes

n/a

n/a


parmentelat:

parmentelat:
the snapshot now correctly describes the expected behaviour (seeing the bottom of the notebook)
it is still named in `darwin` though, I'm not sure if that matters or not

also I have been able to run this test on top of #15184, and can confirm it does fix the issue - this is how I produced the correct snapshot

parmentelat:
after the automatic snapshot update, we have:
- the linux snapshot that is the actual result of the test, so it's broken
- the darwin snapshot, which I created manually using the fix, which shows the expected outcome (the bottom of the notebook)
-

fcollonval:
Thanks a lot @parmentelat 

The snapshots are always os scoped as you realized. This is the reason the CI is generating and using only the _linux_ variant - basically the _darwin_ one will have to be removed before merging.

Would you mind merging this into #15184? This will ease the generation of the correct snapshot and the validation of the PR actually fixing the code.

parmentelat:
ok @fcollonval 
I added the test in #15184
closing this one then

PR #15204, opened 2023-10-02
Restore removed API

fcollonval:
## References

Follow-up of #14796

## Code changes

Restore public API; for a no-op.

## User-facing changes

None

Remove incompatible change of https://github.com/jupyterlab/jupyterlab/pull/14796


fcollonval:
Hey @krassowski would you mind having a quick look at this one because in https://github.com/jupyterlab/jupyterlab/pull/14796 the protected API `onArrowUp` was removed. This restores it as a no-op to be strict about no breaking API (although the probability is low for that specific case).

fcollonval:

PR #15206, opened 2023-10-02
haticecan37:14670-the-tutorial-documentation-needs-to-be-updated

haticecan37:


fcollonval:
Hey @haticecan37 would you mind providing some background for your changes? Moreover we do not support JupyterLab 1.x any longer and all PRs should target the default `main` branch (except if a bug is only for a specific version).

krassowski:
Closing since this appears to have been created by accident.

Issue #15167, opened 2023-09-27
Extension discovery is not working when an HTTP proxy is used


dolevf:
## Description
Extension discovery is not working when an HTTP proxy is used, a timeout error is seen instead of the extensions list.

## Context
JupyerLab 4.0.5

## Reproduce
**Assumes an environment with no internet connectivity**
1. Set environment variable with a proxy (`http_proxy`, `https_proxy`) with an egress proxy 
2. Run Jupyter Lab
3. Under extension discovery, a timeout error will be returned

## Expected behavior
Extension discovery can successfully extensions when a proxy is defined in the environment variable

## Actual behaviour
We observed the following error:
```
  File "/opt/conda/lib/python3.11/site-packages/tornado/web.py", line 1786, in _execute
    result = await result
             ^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/handlers/extension_manager_handler.py", line 37, in get
    extensions, last_page = await self.manager.list_extensions(query, page, per_page)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/extensions/manager.py", line 368, in list_extensions
    await self.refresh(query, page, per_page)
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/extensions/manager.py", line 402, in refresh
    await self._update_extensions_list(query, page, per_page)
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/extensions/manager.py", line 575, in _update_extensions_list
    extensions, last_page = await self.list_packages(query, page, per_page)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/extensions/pypi.py", line 201, in list_packages
    matches = await self.__get_all_extensions()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/jupyterlab/extensions/pypi.py", line 255, in __get_all_extensions
    self.__all_packages_cache = await self.__throttleRequest(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #15219, opened 2023-10-03
Add support for HTTP proxy in extension discovery

dolevf:
## References

Resolves https://github.com/jupyterlab/jupyterlab/issues/15167

## Code changes

Adds Python dependency on `httpx` (https://github.com/encode/httpx).

1. Checks if an `HTTP_PROXY` variable is set in the environment
2. Uses tornado `CurlAsyncHTTPClient` adapter which supports using HTTP/S proxies
3. Overrides XMLRPC Transport to set the proxies


jupyterlab/extensions/pypi.py
```diff
@@ -13,14 +14,17 @@
...
 from pathlib import Path
 from subprocess import CalledProcessError, run
 from tarfile import TarFile
 from typing import Any, Callable, Dict, List, Optional, Tuple
+from urllib.parse import urlparse
 from zipfile import ZipFile
 
 import tornado
 from async_lru import alru_cache
+from tornado.curl_httpclient import CurlAsyncHTTPClient
```
krassowski, line 27: Hmm it depends on `pycurl` and a specific build and version of libcurl is also required as per [docs](https://www.tornadoweb.org/en/stable/httpclient.html#module-tornado.httpclient).

Not sure if we can just import it conditionally (if proxy is enabled, otherwise use `AsyncHTTPClient`) or if we prefer to add it as a dependency anyways. What others think?

fcollonval: That dependency is likely to be annoying (looking at you Windows...). I would prefer adding a new dependency like [aiohttp](https://github.com/aio-libs/aiohttp) or [httpx](https://www.python-httpx.org/) (I checked they support proxy).

@dolevf would you mind changing this PR for one of the above ?

bollwyvl: ... and https://github.com/curl/curl/discussions/12026, so not ideal timing.

bollwyvl: Before considering fresh dependencies (and their dependencies), which _inevitably_ bring fresh headaches, perhaps a route would be to consider hoisting to a configurable, defaulting to the current client, and making the `libcurl`-based one (which _is_ used on e.g. `nbviewer`). Someone would then be able to relatively easily overload this behavior in an extension, which would make it feasible to enable-on-install with the `{sys.prefix}/etc/jupyter/jupyter_server_config.d/` deployment approach.

e.g. 

```yaml
{
  "PyPIExtensionManager": {
    "pypi_client_cls": "jupyterlab.extensions.pypi.clients.TornadoPyPI",
    # vs                jupyterlab.extensions.pypi.clients.TornadoCurlPyPI
    # vs                jupyterlab_httpx.pypi.HttpxPyPI
  }
}
```

davidbrochart: I would be in favor of using `httpx` which is very lightweight and somewhat more modern than `aiohttp`, which also comes with a server.

dolevf: Hi,

I've changed it to `httpx`. I've never worked with Jupyterlab code base before, so I couldn't find a great way to test it end to end other than ensuring it works locally against a squid proxy.

jupyterlab/extensions/pypi.py
```diff
@@ -109,12 +148,11 @@ async def get_latest_version(self, pkg: str) -> Optional[str]:
...
-            response = await http_client.fetch(
-                self.base_url + f"/{pkg}/json",
-                headers={"Content-Type": "application/json"},
-            )
-            data = json.loads(response.body).get("info")
+            async with httpx.AsyncClient(proxies=proxies) as httpx_client:
+                response = await httpx_client.get(
+                    self.base_url + f"/{pkg}/json", headers={"Content-Type": "application/json"}
+                )
+                data = json.loads(response.text).get("info")
```
fcollonval, line 155: Same comment

```suggestion
                data = json.loads(response.content).get("info")
```

jupyterlab/extensions/pypi.py
```diff
@@ -332,9 +370,10 @@ async def install(self, name: str, version: Optional[str] = None) -> ActionResul
             try:
                 download_url: str = pkg_action.get("download_info", {}).get("url")
                 if download_url is not None:
-                    response = await self._http_client.fetch(download_url)
+                    async with httpx.AsyncClient(proxies=proxies) as httpx_client:
+                        response = await httpx_client.get(download_url, proxies=proxies)
                     if download_url.endswith(".whl"):
-                        with ZipFile(io.BytesIO(response.body)) as wheel:
+                        with ZipFile(io.BytesIO(response.text)) as wheel:
```
fcollonval, line 376:
```suggestion
                        with ZipFile(io.BytesIO(response.content)) as wheel:
```

jupyterlab/extensions/pypi.py
```diff
@@ -344,7 +383,7 @@ async def install(self, name: str, version: Optional[str] = None) -> ActionResul
                                 if jlab_metadata is not None:
                                     break
                     elif download_url.endswith("tar.gz"):
-                        with TarFile(io.BytesIO(response.body)) as sdist:
+                        with TarFile(io.BytesIO(response.text)) as sdist:
```
fcollonval, line 386:
```suggestion
                        with TarFile(io.BytesIO(response.content)) as sdist:
```

jupyterlab/extensions/pypi.py
```diff
@@ -31,30 +35,60 @@
...
+    def make_connection(self, host):
+        connection = http.client.HTTPConnection(*self.proxy)
+        connection.set_tunnel(host, headers=self.proxy_headers)
+        self._connection = host, connection
+        return connection
+
+
+xmlrpc_transport_override = None
+
+http_proxy_url = environ.get('http_proxy') or environ.get('HTTP_PROXY')
```
fcollonval, line 52: Let's drop lower case as it should not be used and let's support HTTPS_PROXY and ALL_PROXY

```suggestion
all_proxy_url = environ.get("ALL_PROXY")
http_proxy_url = environ.get("HTTP_PROXY") or all_proxy_url
https_proxy_url = environ.get("HTTPS_PROXY") or all_proxy_url or http_proxy_url
```

jupyterlab/extensions/pypi.py
```diff
@@ -31,30 +35,60 @@
...
+proxies = {}
+
+if http_proxy_url:
+    http_proxy = urlparse(http_proxy_url)
+    proxy_host, _, proxy_port = http_proxy.netloc.partition(":")
+
+    proxies = {
+        "http://": f"http://{proxy_host}:{proxy_port}",
+        "https://": f"http://{proxy_host}:{proxy_port}",
```
fcollonval, line 61: What is the reason to recreate the URL?

jupyterlab/extensions/pypi.py
```diff
@@ -31,30 +35,60 @@
...
+        connection = http.client.HTTPConnection(*self.proxy)
+        connection.set_tunnel(host, headers=self.proxy_headers)
+        self._connection = host, connection
+        return connection
+
+
+xmlrpc_transport_override = None
+
+http_proxy_url = environ.get('http_proxy') or environ.get('HTTP_PROXY')
+proxies = {}
```
fcollonval, line 53: Let's use the default httpx value:

```suggestion
proxies = None
```

jupyterlab/extensions/pypi.py
fcollonval, line 61:
```suggestion
        "http://": http_proxy_url,
        "https://": https_proxy_url,
```

fcollonval (changes requested):
Thanks @dolevf

The code looks good. I have some suggestions and a question.

jupyterlab/extensions/pypi.py
dolevf: Committed your change to remove the duplication

fcollonval (commented):
@dolevf  could you please confirm that the latest version solves the issue in your environment?

fcollonval:
CI failures are not related

dolevf:
> @dolevf could you please confirm that the latest version solves the issue in your environment?

Hi, which version are you referring to?

fcollonval:
> Hi, which version are you referring to?

I meant the latest version of this PR. Just to a trace that it works as it is not tested.

dolevf:
Hi,

I just checked with the recent changes, worked as expected.

fcollonval (approved):
Thanks @dolevf 

CI failures are not related

dolevf:
Can we expect this to be part of the next release 4.0.8?

Issue #15163, opened 2023-09-27
Error message related to ExpansionStates when refreshing a notebook


HaudinFlorence:
## Description
When working in dev mode and watch mode, with necessity to frequently update the view of the notebook, it has appeared that when making a change and refreshing, it may happen that operation takes time or not succeed. The following error message appears in the console:

![Screenshot from 2023-09-27 10-42-16](https://github.com/jupyterlab/jupyterlab/assets/99649086/e431b141-7079-454b-aa34-37694ff38d31)

After clicking on `Clear workspace` and refreshing another time, normally the notebook correctly restarts without error message.

## Reproduce

1. Open a notebook in dev mode and watch mode
2. Make some change in the code and wait that compilation is finished
3. Refresh the notebook
4. See that it freezes, or lead to a dialog proposing to clean the workspace. In the console, the error message is:
```
shell.ts:2072 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'expansionStates')
    at shell.ts:2072:40
    at Array.forEach (<anonymous>)
    at shell.ts:2071:31
    at Array.forEach (<anonymous>)
    at SideBarHandler.rehydrate (shell.ts:2068:36)
    at LabShell.restoreLayout (shell.ts:1205:25)
```

krassowski:
Thank you for opening the issue. Appears related to #14901 CC @DenisaCG.

DenisaCG:
Thanks for opening the issue @HaudinFlorence and @krassowski for pointing us towards the PR it is related to. I will investigate the behavior to figure out what seems to be the problem.

fcollonval:
i think the fix is to replace that line:

https://github.com/jupyterlab/jupyterlab/blob/1692817b23ce35a31d02889c4af83df732c00a17/packages/application/src/shell.ts#L2070

by

```ts
const state = data.widgetStates[w.id] ?? {}; 
```

You should be able to try it by removing the widgetStates content in `$HOME/.jupyter/lab/workspaces/default-####.jupyterlab-workspace`

> The file has a hash after default that varies

PR #15225, opened 2023-10-05
Fix expanionStates error 

DenisaCG:
This should fix the error coming from the `expansionStates` encountered when working in dev mode on jupyterlab.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15163

## Code changes

When rehydrating the sidebar and extracting the `widgetStates`, if the user didn't modify the expansion states or sizes of any subpannel, so they are still in their original state, the variable should be empty.

```
const state = data.widgetStates[w.id] ?? {};
```

## User-facing changes

This issue was encountered by those working in dev mode, so now that will be fixed and there should be no more errors, even when frequently refreshing the notebook.

None.

cc: @fcollonval

### DIFF
--- a/galata/test/jupyterlab/workspace.test.ts
+++ b/galata/test/jupyterlab/workspace.test.ts
@@ -266,18 +266,37 @@ test.describe('Workspace in doc mode', () => {
         left: {
           collapsed: false,
           visible: true,
-          current: 'filebrowser',
+          current: 'running-sessions',
           widgets: [
             'filebrowser',
             'running-sessions',
             '@jupyterlab/toc:plugin',
             'extensionmanager.main-view'
-          ]
+          ],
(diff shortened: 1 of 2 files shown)


fcollonval:
@HaudinFlorence would you mind reviewing this PR?

HaudinFlorence:
@DenisaCG Thanks for this PR. I have tried it and the error message previously mentioned in issue #15163 no longer appears.

krassowski:
Thank you! Would it be easy to add a unit test for this one?

krassowski (approved):
Thank you @DenisaCG 

I am slightly worried that we needed to restart visual regression tests 3 times on this PR, but maybe it we were just out of luck :man_shrugging:. Anyways, it is passing now :rocket:.

PR #14726, opened 2023-06-21
Improve examples

fcollonval:
## References

Fixes #14695
Fixes https://github.com/jupyterlab/jupyterlab/issues/10377

## Code changes

It improves the examples:
- notebook; restore code highlighting, markdown parser and toolbar
- code cell: code highlighting
- console: code highlighting

It adds snapshot test on examples

There is two minor API changes:
- markedparser-extension: expose parser factory for easier consumption (new API)
- notebook: execution indicator - makes an argument optional (it was actually already supported by the factory was enforcing it).

--> I think it will be good to backport this to 4.0.x despite the additional API to ease porting integration of Jupyter as mentioned in #14695.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #15230, opened 2023-10-06
Fix MathJax font override, avoid double initialization

bollwyvl:
## References

- maybe roll back some ignored issues on https://github.com/jupyterlab/jupyterlab/pull/14726
- downstream
  - noted on https://github.com/jupyterlite/jupyterlite/issues/1192
  - started on https://github.com/jupyterlite/jupyterlite/pull/1193 by @jtpio 

## Code changes

- [x] make the overload of `TeXFont.defaultFonts` static 
- [x] move inner contents of `protected _ensureInitialized` to a `Private` namespace
  - [x] use a `PromiseDelegate` to avoid race initilization conditions
  - [x] load all used named MathJax imports in parallel
- [x] remove example `404` ignores  
- [x] use `font-display: swap` to reduce FLOUC
  - > tried some `preload`, etc. but the `no-cache` setting makes that... worse, and while loading `Zero` might be a win, it's hard to anticipate the rest...

## User-facing changes

- `.woff` fonts should no longer be downloaded twice from the default URL (`{window.location}/js/output/chtml/`)
- multiple calls to `typeset` should not be able to enter the async init code twice  

- n/a


packages/mathjax-extension/src/index.ts
```diff
@@ -157,3 +108,77 @@ const mathJaxPlugin: JupyterFrontEndPlugin<ILatexTypesetter> = {
...
+        import('mathjax-full/js/adaptors/browserAdaptor'),
+        import('mathjax-full/js/a11y/assistive-mml')
+      ]);
+
+      mathjax.handlers.register(
+        AssistiveMmlHandler(SafeHandler(new HTMLHandler(browserAdaptor())))
+      );
+
+      class EmptyFont extends TeXFont {
+        protected static defaultFonts = {} as any;
```
bollwyvl, line 152: this was the actual important thing

bollwyvl:
Presumably the example test passed: certainly not seeing the issue locally any longer. Not sure what the story is on the other UI tests.

jtpio:
Nice, thanks!

I switched the label from `enhancement` to `bug` as this looks more like an issue with the current logic.

Also this can likely be backported to `4.0.x`.

examples/example.spec.ts
```diff
@@ -22,15 +22,7 @@ test('should load the example', async ({ page }) => {
...
-    if (
-      msg.type() === 'error' &&
-      // Ignore 404 for MathJax fonts
-      msg
-        .location()
-        .url.match(
-          /js\/output\/chtml\/fonts\/tex-woff-v2\/MathJax_(Zero|Math-Italic|Main-Regular)\.woff$/
-        ) === null
-    ) {
+    if (msg.type() === 'error') {
```
jtpio, line 25: Yeah these ignored error were suggesting something was not correct.

jtpio (approved):
Thanks!

jtpio:

bollwyvl:
I'll take a look at the manual backport when i get a chance.

PR #15241, opened 2023-10-09
Improve docker script

trungleduc:
## References

- Improve Dockerfile to accelerate the build process
- Allow starting the dev container at different ports, the following command will start JupyterLab at port `4567` instead of the default one.
```bash
bash ./docker/start.sh dev 4567 
```


krassowski (approved):
Thank you @trungleduc

Issue #15144, opened 2023-09-21
Virtual document updates may lead to a memory leak


krassowski:
## Description

Editing large notebook leads to substantial allocation from `appendCodeBlock` on each keypress, even if jupyterlab-lsp is not installed:

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/8bfd7858-e0b2-45a6-8e24-94c1825eb318)

## Reproduce

1. Open a large notebook with lots of code and 1000+ cells
2. Wait until it is fully loaded
3. Start memory profiler
4. Start typing
5. See increasing memory allocation

## Expected behavior

- The LSP virtual document is not listening to updates if there is no LSP feature registered or no server connected
- The LSP virtual document is not being re-created on each keystroke, but updated based on transaction

## Context

- Browser and version: Chrome/Electron
- JupyterLab version: 4.0.6

trungleduc:
Hi @krassowski, the first item can be addressed by #15250, I took a look at the second point but It's much more complicated. Basically, we need to rewrite the `VirtualDocument` to support updating by patch. So I put it on hold and started working on other low-hanging fruits

PR #15250, opened 2023-10-10
Update virtual documents conditionally

trungleduc:
## References

Partially addresses https://github.com/jupyterlab/jupyterlab/issues/15144

> The LSP virtual document is not listening to updates if there is no LSP feature registered or no server connected

## Code changes

- The virtual document is connected to the `contentChanged` signal if three conditions are satisfied: 
  -  The LSP feature is enabled.
  -  There is a connected server for the language of the document.
  - The LSP features list is not empty.

### DIFF
--- a/packages/lsp/src/adapters/adapter.ts
+++ b/packages/lsp/src/adapters/adapter.ts
@@ -97,6 +97,14 @@ export abstract class WidgetLSPAdapter<
     this._editorToAdapter = new WeakMap();
     this.editorAdded.connect(this._onEditorAdded, this);
     this.editorRemoved.connect(this._onEditorRemoved, this);
+    this._connectionManager.languageServerManager.sessionsChanged.connect(
+      this._onLspSessionOrFeatureChanged,
+      this
+    );
+    this.options.featureManager.featureRegistered.connect(
+      this._onLspSessionOrFeatureChanged,
+      this
+    );
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/lsp/src/adapters/adapter.ts
```diff
@@ -718,4 +733,41 @@ export abstract class WidgetLSPAdapter<
...
+    if (!this._virtualDocument) {
+      return;
+    }
+
+    const languages = [
+      this._virtualDocument.language,
+      ...[...this._virtualDocument.foreignDocuments.values()].map(
+        it => it.language
+      )
+    ];
```
krassowski, line 763: I am afraid that this is a chicken and egg problem. Downstream we do not know all the languages until we updated the document as this is when the extraction happens.

krassowski (commented):
- The LSP feature is enabled.
:+1: 
- The LSP features list is not empty.
:+1: 
- There is a connected server for the language of the document.
  This one is problematic when transclusions/nested documents are at play. We do not know the languages until we parsed (updated) the document. Is my reasoning correct?

Maybe for now it would be good enough to use the first two as good enough heuristic?

trungleduc:
Failed UI tests do not seem to be related.

packages/lsp/test/adapters.spec.ts
```diff
@@ -212,6 +214,130 @@ describe('@jupyterlab/lsp', () => {
       tracker.dispose();
     });
   });
+
+  describe('WidgetLSPAdapter1', () => {
```
krassowski, line 218:
```suggestion
  describe('WidgetLSPAdapter', () => {
```

krassowski (approved):
Thank you @trungleduc!

packages/lsp/test/adapters.spec.ts
trungleduc: Thanks!

krassowski:
Visual test failure unrelated, merging.

Issue #15198, opened 2023-10-02
Support to override the serializer used in kernel.


DonJayamanne:
### Problem

I'm one of the maintainers of the Jupyter extension for VS Code.

**Context:** 

* Not really a problem, but more of a feature request.
* The Jupyter extension in VS Code makes use of the @jupyterlab/services packages and a few others extensively.
Thanks for these packages, I have mentioned to @jasongrout in the past how great and flexible they are
* Today we use these pacakges even in the cases where we start kernels locally.
* We start the kernels, use zmq to communicate and then create a fake socket that is passed via the `ServerConnection.ISettings`, brilliant design (thanks)
* However we have slightly modified version of the `default.js` file at our end, we override the serializer.

**Problem:** 

As the messages are already serialized/deseirliazed, we do not need these serializers.
Today we tweek the code slightly to ensure nothing gets serialized/deserialized in the KernelConnection class (default.ts).

I was hoping there would be a way to pass the serializer into the constructor of `KernelConnection` that would allow us (Jupyter extension in VSC) to bypass the built in serializer/deserializer, just as we can pass in a custom WebSocket class today.

This would make it possible to use the Jupyterlab package without any changes at our end.
Basically I would like to remove these little hacks/work arounds if possible.

Once again, thanks for publishing these great packages and documenting them.

### Proposed Solution

* Add an optional property to `ServerConnection.ISettings` that would contain the serializer as follows:
```typescript
export interface ISerializer { 
    /**
     * Serialize a kernel message for transport.
     */
    serialize(
      msg: KernelMessage.IMessage,
      protocol: string = ''
    ): string | ArrayBuffer ;
    /**
     * Deserialize and return the unpacked message.
     */
    deserialize(
      data: ArrayBuffer,
      protocol: string = ''
    ): KernelMessage.IMessage;
}

export interface ISettings {
.....
    serializer?: ISerializer;
....
}

// The `ServerConnection.makeSettings` would then return create the default serializer in there.
```

### Additional context

* Applies to Jupyter extension in VS Code.

krassowski:
Please feel welcome to submit a pull request, I can help with review.

DonJayamanne:
@krassowski 
Here's the draft PR for review/discussion https://github.com/jupyterlab/jupyterlab/pull/15254

echarles:
@DonJayamanne Out of curiosity, why are you passing a custom serializer? (any link to the vscode python extension which does that).

[...]

PR #15254, opened 2023-10-11
Custom kernel message serializer

DonJayamanne:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15198

## Code changes

Exposes the message serializer in the ISettings, allowing one to customize how messages are serialized/deserialized based on the server (settings).

## User-facing changes

N/A

N/A


krassowski (approved):
Thanks @DonJayamanne, this looks good to me!

DonJayamanne:
@krassowski 
Should I mark this as ready for review, or is there something else we need?
I do not think we need tests for this change, and the docs are generated from the ts code.

krassowski:
Addding a unit test ensuring that the serializer can actually be changed can help catch regression if this function gets rewritten (all it takes to break it is to move `...options` one line up) so it could benefit you on the future. Other than that no follow up tasks here.

fcollonval (approved):
Thanks @DonJayamanne 

I agree that it will be nice to test providing custom de/serializers in a test.

Issue #14871, opened 2023-07-24
Find/replace highlighting keeps coming back up after find prompt closed


James-Huffman:
## Description

Highlighting for find/replace in text editor comes back unexpectedly after closing find interface and typing additional text.
![Screenshot 2023-07-21 130935](https://github.com/jupyter/notebook/assets/1227014/3d46c8ef-b616-47c8-994b-888984895572)
![Screenshot 2023-07-21 131004](https://github.com/jupyter/notebook/assets/1227014/4f2e96d3-313f-4bab-ab5b-71202dd4dc15)

I originally found this bug in notebook, but it appears that they implement this search feature from jupyterlab.

## Reproduce

1. Open a text document in JupyterLab or jupyter notebook.
2. Ctrl-f or use Edit -> Find... to open the find prompt.
3. Type string you want to find
4. Press enter to highlight matching cases in your document
5. Press esc to close the find interface
6. Type any new character - the find highlighting will reappear despite the closed interface.

This seems to only affect the text editor, not find/replace behavior within jupyter notebook cells.

Highlighting cannot be resolved by clearing the find query before closing the find/replace interface. Only way to make this go away is to either reload the page or search for a non-matching query.

## Expected behavior

Text should remain unhighlighted until the find/replace interface is re-opened. 

## Context

- Operating System and version: Windows 10 Enterprise
- Browser and version: Chrome 114.0.5735.199

- Jupyter Notebook version: 7.0.0
- JupyterLab version 4.0.3

<details><summary>Troubleshoot Output</summary>
<pre>
[troubleshoot.txt](https://github.com/jupyter/notebook/files/12131167/troubleshoot.txt)
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
[command_line.txt](https://github.com/jupyter/notebook/files/12131168/command_line.txt)
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
No output.
</pre>
</details>

JasonWeill:
Do you also see this behavior in JupyterLab 3 or Notebook 6? I know that @krassowski made a lot of improvements to search and replace in the newer versions.

James-Huffman:
Notebook 6 (6.5.4) implements a different find/replace interface, and this bug is not present.

JupyterLab3 (3.6.5) does not have persistent highlighting after using find/replace, but it also does not have cell by cell searching when running notebooks.

krassowski:
[...]

PR #15259, opened 2023-10-11
a test case to illustrate issue #14871

parmentelat:
## References

issue #14871

## Code changes

one more test case, with related input and outputs

**notes** 
- the test right now breaks, because of #14871 precisely
- also at first the outputs are made on darwin - will get rid of that once the correct linux outputs are produced

## User-facing changes

none

none


parmentelat:
also I had to run `git commit --no-verify`, the prettier stage was failing, even though `jlpm run prettier` seemed happy
please let me know how to deal with that

krassowski:
`jlpm run prettier` is only prettier, there is also `eslint` which might have caught that `getSelectionRange` is unused and needs to be used or removed. In general `jlpm run lint` from root of the repo should work.

parmentelat:
I must have misunderstood, I had the impression it was the prettier stage that was wrong..
anyway, I have removed the unused function and it is now all good wrt committing

parmentelat:
trying to improve the situation, here's a few thoughts to share

###  level1

the first-level bug has to do with the search dialog having lasting effects
and indeed, it will be an obvious improvement 
if giving up on the search dialog (using either Escape or the cross icon) 
brings us back in a state where nothing is being searched

### level2

however, playing with the search a little more, I came to realize that, 
when the search dialog is still on, it feels like the job in charge of outlining stuff
sometimes has a very intrusive behaviour, notably putting some cells in edit mode while they were not,
which is imho wrong in itself, and worse still it often leads to arbitrary cells becoming in focus (in the viewport)
as a result this intrusion is very often devastatingly confusing for the user,
because it makes one lose one's train of thought by scrolling up or down with no apparent sense 


my own experience has been to face level2, so I tried to get rid off the issue by turning off the search dialog,
but it was not enough because of level1

So I guess what I'm saying is, I might eventually be able to fix stage1 as part of this PR, I'll try at least
but even so, there will remain some agreement to reach on how the search outlining business
is expected to behave, and imho the expected behaviour would be:
- no change of editing mode
- no scrolling whatsoever

but I guess that's just me
and apologies for getting a little off-track as compared to the initial target of this PR

krassowski:
Thanks again for your work here @parmentelat, closing this in favour of https://github.com/jupyterlab/jupyterlab/pull/15443 which includes your commits from this PR. Pleas feel welcome to review.

Issue #15126, opened 2023-09-15
Problems extending virtual LSP documents downstream


krassowski:
When LSP package was migrated to this repository a decision was made that:
- transclusions allowing magics such as `%%R` or `%%javascript` (nested or foreign documents) will not to be migrated, but the downstream extension will be able to add them back
- overrides will not be migrated and downstream will need to reimplement them

Some crucial code enabling the above was removed, and some methods were made private preventing downstream from adding it back. Further, the original class was not really designed for subclassing so there are some modifications needed anyways.

Things that require work to honour the agreement with downstream:
- as of now `private chooseForeignDocument()` no longer recycles connections for standalone transclusions; it not doing so lead to <s>dozens hours of debugging lost</s> infinite loops and performance degradation downstream because half of the code responsible for it was removed and the other half was not; see https://github.com/jupyter-lsp/jupyterlab-lsp/issues/959. There might be a better solution to the underlying problem of "how do I avoid creating a new foreign document on each notebook change"
- `openForeign()` uses hard-coded `VirtualDocument` but should use `this.constructor` to allow sub-classing [here](https://github.com/jupyterlab/jupyterlab/blob/1d989bad87e89307a336d2f671c9cc98ec37f5e9/packages/lsp/src/virtual/document.ts#L995-L1006).
- mapping between positions from transclusions to top-level document became difficult because:
   - the `virtualLines` mapping needed to find corresponding editor is protected
   - `transformVirtualToRoot` from old adapter was removed (which was a good decision), but we have not got new method for it; it would fit nicely on `VirtualDocument` class
- shifts and `skipInspect` were removed from `appendCodeBlock` and friends; again it is embarrassing to admit how many hours I lost on it because some of the code was removed, but other was not so it was again difficult to see the problem.

I believe some of the problems could be addressed by designing a much better APIs rather than restoring things to how they were, so for now I am just trying to document some of the downstream requirements.

PR #15262, opened 2023-10-11
Fix connection loop issue with standalone foreign document in LSP

trungleduc:
## References

Partially addresses #15126
Improvement in `VirtualDocument`: 
- `chooseForeignDocument` now reuses unused standalone virtual document
- `openForeign` creates foreign documents from the current class instead of `VirtualDocument`.
- New method `transformVirtualToRoot`

### DIFF
--- a/packages/lsp/test/document.spec.ts
+++ b/packages/lsp/test/document.spec.ts
@@ -225,16 +225,16 @@ describe('@jupyterlab/lsp', () => {
         expect(foreignDocumentsMap.size).toEqual(1);
       });
     });
-    describe('#chooseForeignDocument', () => {
+    describe('#_chooseForeignDocument', () => {
       it('should select the foreign document for markdown cell', () => {
-        const md: VirtualDocument = document['chooseForeignDocument'](
+        const md: VirtualDocument = document['_chooseForeignDocument'](
           markdownCellExtractor
         );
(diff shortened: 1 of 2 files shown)


packages/lsp/test/document.spec.ts
```diff
@@ -225,18 +233,41 @@ describe('@jupyterlab/lsp', () => {
...
         );
         expect(md.uri).toBe('test.ipynb.python-markdown.md');
       });
       it('should select the foreign document for raw cell', () => {
         const md: VirtualDocument =
-          document['chooseForeignDocument'](rawCellExtractor);
+          document['_chooseForeignDocument'](rawCellExtractor);
         expect(md.uri).toBe('test.ipynb.python-text.txt');
       });
+      it('should use unused virtual document if availble', () => {
```
krassowski, line 248:
```suggestion
      it('should use unused virtual document if available', () => {
```

krassowski (approved):
Thank you @trungleduc, thanks for patience here.

JasonWeill:
@krassowski @trungleduc Thank you for fixing this! Would it be possible to backport this fix to the 4.0.x branch, so that it can go out in a patch release before 4.1.0 is generally available?

JasonWeill:

PR #15265, opened 2023-10-12
Added alt descriptions to a few icon and images

j264415:
This PR is to replace the #14320 addressing the requested changes from @isabela-pf and @gabalafou feedback from that PR. 
## References

#14320 Add more alt text descriptions

## Code changes

These are the additional changes made :

- Made changes to **packages/apputils/src/dialog.tsx** **line 390** by changing the **kernelIconURL** alt tag to `` per @isabela-pf request. 

- Removed the title from the arguments that is being passed into `_initSvg` that were asked for from @gabalafou in the **packages/ui-components/src/icon/labicon.tsx** **_line 552 and 825_**.

This PR also retains the changes from #14320:

- Added alt and title description to images and icons

## User-facing changes

Icons and images have descriptive labels when you hover over them these will also be picked up by screenreaders.

None


j264415:

packages/launcher/src/widget.tsx
```diff
@@ -174,7 +174,8 @@ export class Launcher extends VDomRenderer<ILauncher.IModel> {
                 icon={icon}
                 iconClass={classes(iconClass, 'jp-Icon-cover')}
                 stylesheet="launcherSection"
-              />
+                aria-hidden="true"
```
tonyfast, line 177: what is being hidden from assistive tech?

packages/help-extension/src/index.tsx
```diff
@@ -405,7 +405,9 @@ const resources: JupyterFrontEndPlugin<void> = {
             isEnabled,
             execute: () => {
               // Create the header of the about dialog
-              const headerLogo = <img src={kernelIconUrl} />;
+              const headerLogo = (
+                <img src={kernelIconUrl} alt={trans.__('Project Jupyter')} />
```
krassowski, line 409:
```suggestion
                <img src={kernelIconUrl} alt={trans.__('Kernel Icon')} />
```

image

packages/ui-components/src/icon/labicon.tsx
```diff
@@ -549,10 +549,6 @@ export class LabIcon implements LabIcon.ILabIcon, VirtualElement.IRenderer {
       if (uuid) {
         svgElement.dataset.iconId = uuid;
       }
-
-      if (title) {
-        Private.setTitleSvg(svgElement, title);
-      }
```
krassowski, line 555: Why is this being removed? CC @gabalafou

krassowski (changes requested):

packages/help-extension/src/index.tsx
gabalafou: Can someone explain the rationale for providing alt text for this icon? To me, the icon appears to be purely decorative, so I think we should explicitly set the alt text to an empty string, `alt=""`. What do people think?

packages/launcher/src/widget.tsx
gabalafou: I believe this is each of the section icons in the launcher, like so:

1. (icon: orange square containing orange bookmark) Notebook
2. (icon: blue square containing `>_`) Console
3. (icon: black square containing `$_`) Other

packages/launcher/src/widget.tsx
```diff
@@ -298,7 +299,11 @@ function Card(
       <div className="jp-LauncherCard-icon">
         {kernel ? (
           item.kernelIconUrl ? (
-            <img src={item.kernelIconUrl} className="jp-Launcher-kernelIcon" />
+            <img
+              src={item.kernelIconUrl}
+              className="jp-Launcher-kernelIcon"
+              alt={trans.__('Kernel Icon')}
```
gabalafou, line 305: Same: what is the rationale for not setting `alt=""`?

packages/ui-components/src/icon/labicon.tsx
gabalafou: Oh, hmm. I think I messed up here. I thought it was safe to remove `setTitleSvg` because it is private and only called on this line and there's only one place in the codebase where `_initSvg` is called and it does not pass `title` (line 375 above):

```
      this._svgElement = this._initSvg({ uuid: this._uuid });
```

But I guess because `_initSvg` is a protected method, it could be called by some other code outside the repo that subclasses `LabIcon`, so I guess the `setTitleSvg` function is not safe to remove, sorry!

packages/launcher/src/widget.tsx
j264415: The reason alt is not an empty string is because we believe it is not a decorative element and adds value to the end users.

packages/ui-components/src/icon/labicon.tsx
j264415: I have put it back into the file.

krassowski: No sorry I was not clear in my question (I think it is likely fine to remove it, but possibly we should start with a deprecation); the intent/gist of my question is to understand why we do not want to have an option to set a title on the svg icon using this code path because top level comment in this PR does not explain.

krassowski: Ok, sorry that we got back and forth here. The bottom line: if we are removing them, it needs a good rationale and explanation in the top-level comment; otherwise, lets restore it and we can merge this PR as is..

packages/apputils/src/dialog.tsx
```diff
@@ -875,6 +875,7 @@ export namespace Dialog {
                   iconClass="jp-Icon"
                   className="jp-ToolbarButtonComponent-icon"
                   tag="span"
+                  title={trans.__('Close toolbar')}
```
krassowski, line 878: Since this is used in popus like:

image

It should probably read:

```suggestion
                  title={trans.__('Close popup')}
```

(not toolbar)

tonyfast: i might suggest `dialog` so that this relates to an element with a definition and expectation.

krassowski (approved):
Thank you @j264415, this looks good from my side.

packages/launcher/src/widget.tsx
gabalafou: In the accessibility call today, we discussed a scenario in which the text description that a kernel provides to describe itself might make it unclear in the launcher that it is a kernel. Imagine a kernel called Coconut that labels itself as just "Coconut." In that case, it might be helpful to a screen reader to announce the launcher card as "Kernel Icon coconut".

We also discussed some alternative in the call. What if the alt text were something like "Create new notebook using kernel", so then the full accessible text for a Python 3 ipykernal notebook launcher card would read like so: "Create new notebook using kernel Python 3 (ipykernel)". That seems a bit verbose to me but clear.

My chief concern in asking these questions is to avoid a common pitfall, which is: in our desire to make things better for disabled users, we actually make them worse. I want to try to avoid that.

packages/launcher/src/widget.tsx
```diff
@@ -174,7 +174,8 @@ export class Launcher extends VDomRenderer<ILauncher.IModel> {
                 icon={icon}
                 iconClass={classes(iconClass, 'jp-Icon-cover')}
                 stylesheet="launcherSection"
-              />
+                aria-hidden="true"
+              />{' '}
```
gabalafou, line 178: Was this space intentionally added?

```suggestion
              />
```

packages/apputils/src/dialog.tsx
```diff
@@ -875,6 +875,7 @@ export namespace Dialog {
                   iconClass="jp-Icon"
                   className="jp-ToolbarButtonComponent-icon"
                   tag="span"
+                  title={trans.__('Close dialog')}
```
gabalafou, line 878: Testing this last night, I realized that this "x" button already had an accessible name: "Cancel". See line 870 above. I don't understand the rationale for adding an additional title here because we end up with the following HTML tree structure:

```
button title=Cancel
  span title=Close dialog
    svg
```

j264415: So would you recommend I take out the additional title on line 878.

j264415: Hi, just tested this with the screen reader, to see if the `title="close dialog" `was needed and it wasn't needed after all as it was already picking up the `title="cancel" `; so I have take it out.

packages/launcher/src/widget.tsx
j264415: Hi @gabalafou your comment makes sense and I completely understand where you are coming from. Thanks for the feedback. I have made the change that you have recommended for the alt text being "Create new notebook using kernel".

packages/launcher/src/widget.tsx
```diff
@@ -298,7 +299,11 @@ function Card(
       <div className="jp-LauncherCard-icon">
         {kernel ? (
           item.kernelIconUrl ? (
-            <img src={item.kernelIconUrl} className="jp-Launcher-kernelIcon" />
+            <img
+              src={item.kernelIconUrl}
+              className="jp-Launcher-kernelIcon"
+              alt={trans.__('Create new notebook using kernel')}
```
fcollonval, line 305: Let's use the title here as the card may be use for creating notebook or console.

```suggestion
              alt={title}
```

fcollonval (commented):
Thanks @j264415 

I have a minor comment. After that this is good to go.

j264415:
> Thanks @j264415
> 
> I have a minor comment. After that this is good to go.

Hi @fcollonval all done!

fcollonval (approved):
Thanks

PR #15271, opened 2023-10-16
Keyboard navigation to top menuBar test

e218736:
## Code changes
added tests for keyboard navigation on top bar menu

## User-facing changes
None

None

### DIFF
--- a/galata/test/jupyterlab/menus.test.ts
+++ b/galata/test/jupyterlab/menus.test.ts
@@ -176,3 +176,35 @@ function reduceItem(
   }
   return null;
 }
+
+test.describe('Top menu keyboard navigation', () => {
+  test('navigate to open launcher with keyboard', async ({ page }) => {
+    await page.goto();
+    await page.keyboard.press('Shift+Tab');
+    await page.keyboard.press('Shift+Tab');
+    await page.keyboard.press('Shift+Tab');
+    await page.keyboard.press('Enter');
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('Enter');
+    await expect(page.locator('#tab-key-2-1')).toBeFocused();
+  });
+
+  test('navigate to close launcher with keyboard', async ({ page }) => {
+    await page.goto();
+    for (let i = 0; i < 4; i++) {
+      await page.keyboard.press('Shift+Tab');
+    }
+    await page.keyboard.press('Enter');
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('Enter');
+    for (let i = 0; i < 3; i++) {
+      await page.keyboard.press('Shift+Tab');
+    }
+    await page.keyboard.press('Enter');
+    for (let i = 0; i < 4; i++) {
+      await page.keyboard.press('ArrowDown');
+    }
+    await page.keyboard.press('Enter');
+    await expect(page.locator('#tab-key-2-0')).toBeFocused();
+  });
+});


tonyfast:
i'm unclear about what you are asking?

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,180 @@ function reduceItem(
   }
   return null;
 }
+
+test.describe('Top menu keyboard navigation', () => {
```
gabalafou, line 180: Should we tag this as an accessibility test?

```suggestion
test.describe('Top menu keyboard navigation @a11y', () => {
```

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,180 @@ function reduceItem(
...
+  test.use({ autoGoto: false });
+  test('navigate to open file with keyboard', async ({ page }) => {
+    await page.goto();
+    for (let i = 0; i < 3; i++) {
+      await page.keyboard.press('Shift+Tab');
+    }
+    await page.keyboard.press('Enter');
+
+    expect(await page.menu.isOpen('File')).toBeTruthy();
```
gabalafou, line 189: This test feels a bit brittle to me. We can make the test less brittle by not counting the number of tab presses needed, instead pressing tab continually until we land on the "File" menu bar item. Note that I did not test the following code suggestion:

```suggestion
    const fileMenu = page.getByRole('menuitem', { name: 'File' });
    while (true) {
      await page.keyboard.press('Shift+Tab');
      if (fileMenu.evaluate(el => el === document.activeElement)) {
        break;
      }
    }
    await page.keyboard.press('Enter');

    expect(await page.menu.isOpen('File')).toBeTruthy();
```

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,180 @@ function reduceItem(
   }
   return null;
 }
+
+test.describe('Top menu keyboard navigation', () => {
+  test.use({ autoGoto: false });
+  test('navigate to open file with keyboard', async ({ page }) => {
+    await page.goto();
+    for (let i = 0; i < 3; i++) {
+      await page.keyboard.press('Shift+Tab');
```
gabalafou, line 185: Why are we using shift+tab rather than tab?

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,180 @@ function reduceItem(
...
+    }
+    await page.keyboard.press('Enter');
+
+    expect(await page.menu.isOpen('File')).toBeTruthy();
+  });
+
+  // test("navigate to open edit with keyboard", async ({ page }) => {
```
gabalafou, line 192: Heads up, I cannot approve this PR while it has commented-out code

gabalafou (changes requested):
Really makes me happy to see more tests, thank you. I left some inline questions, suggestions, comments :)

gabalafou:
I just clicked the Approve and Run button to unblock the GitHub checks from running.

For those who are following this mystery, it appears that GitHub Actions have the following default security rule: 

> Require approval for first-time contributors who are new to GitHub. 
> Only first-time contributors who recently created a GitHub account will require approval to run workflows.

It's unclear to me if we can turn that off, but @krassowski recommended that we keep this rule active since we have limited CI credits, and I agree with him.

e218736:
Thanks, I will look to add changes/more tests once my account is old enough to run work flows and check passing tests

krassowski:
It been two weeks, I would think it should be old enough already? We can keep pressing the button or you can also open a PR against your own fork in which CI would not have any restrictions.

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,184 @@ function reduceItem(
   }
   return null;
 }
+
+test.describe('Top menu keyboard navigation @a11y', () => {
+  test.use({ autoGoto: false });
+  test('navigate to open file with keyboard', async ({ page }) => {
+    await page.goto();
+    const fileMenu = page.getByRole('menuitem', { name: 'File' });
+    while (true) {
```
gabalafou, line 185:
```suggestion
    // eslint-disable-next-line no-constant-condition
    while (true) {
```

galata/test/jupyterlab/menus.test.ts
```diff
@@ -176,3 +176,184 @@ function reduceItem(
   }
   return null;
 }
+
+test.describe('Top menu keyboard navigation @a11y', () => {
+  test.use({ autoGoto: false });
+  test('navigate to open file with keyboard', async ({ page }) => {
```
gabalafou, line 182:
```suggestion
  test('open file menu with keyboard', async ({ page }) => {
```

e218736:
I have opened a PR on my own fork and will take this one out of draft when ready for tests to be reviewed, thanks

galata/test/jupyterlab/menus.test.ts
e218736: Thanks, have amended to tab

Issue #12929, opened 2022-08-10
Open/limit FileDialog.getExistingDirectory from a subdir of root


suganya-sk:
### Problem
The current `FileDialog.getExistingDirectory` is great for getting user input as a directory. While this is very useful, it opens, by default, from the root dir of the app. 

I could not find an option to modify this in the docs - https://jupyterlab.readthedocs.io/en/stable/api/interfaces/filebrowser.filedialog.idirectoryoptions.html

### Proposed Solution

* Add an option to accept the path of the subdir from which `FileDialog.getExistingDirectory` should open.
* This could also mean that it must not be possible to navigate to a dir higher than the path provided, in this dialog.

krassowski:
> - Add an option to accept the path of the subdir from which FileDialog.getExistingDirectory should open.
> - This could also mean that it must not be possible to navigate to a dir higher than the path provided, in this dialog.

This sounds like two separate options to me. I would be happy to help you with adding the first one (it could be called `defaultPath`) - it would be easier to implement.

I _think_ that you would need to add `defaultPath` to `IDirectoryOptions` interface, pass it on down to `OpenDialog` constructor and further to `createFilteredFileBrowser` and then add `model.cd(defaultPath)` after:

https://github.com/jupyterlab/jupyterlab/blob/b57ec962a7b8cda65bb89a72223f195564f8387a/packages/filebrowser/src/opendialog.ts#L242-L249

Restricting traversal would also require modifying the `root` instead; it can be done by adding a new argument in `FileBrowserModel` (`driveName` already works somewhat like that) and possibly even more extensive changes; it is not clear to me if this is worth it (it could make the code more difficult to maintain).

Would you like to make a pull request for it?

suganya-sk:
Thanks for your comment!

I'll create a PR for the first part - an option to open the file dialog from a specific dir.

For the second one, I'll add validation to the path selected in my implementation, that should work.

Throwing ideas here - I dont know much about the internal workings of Jupyter here, but do you think the restriction of traversal can be achieved purely though UI? In the dialog box, currently, the way to navigate to other dirs is by clicking on the various parts of the path displayed on top. If this were to either - 

- only display the path from which the dialog box is opened (OR)
- display the whole path but the higher dirs are disabled so they cannot be clicked 

would that help what we want to achieve?

suganya-sk:
[...]

PR #15278, opened 2023-10-18
Add defaultPath option to set the default directory for file dialog

mmichilot:
## References

Adds the `defaultPath` option referenced in #12929 

## Code changes

Adds a new optional `defaultPath` option to `IDirectoryOptions`. This  option is then passed to `createFilteredFileBrowser` where it uses `defaultPath` to change the directory of the model. The resulting file dialog opens to the given path instead of root.

Also added two new tests to verify `defaultPath` functionality. 
Within a folder in root, a new file and folder are added as shown:
```
.
├── Untitled Folder/ <---- defaultPath
│   ├── Untitled File
│   └── Untitled Folder
├── Untitled File
└── Untitled Notebook
```
The tests check that the parent directory of the selected file/folder matches the given `defaultPath`.

## User-facing changes

N/A

N/A


mmichilot:
There is a lint error regarding floating promises. Since `model.cd()` returns `Promise<void>`, would it be safe to mark the call as intentionally not awaited using the void operator?

PR #15282, opened 2023-10-19
Add defaultPath option to set the default directory for file dialog

mmichilot:
Apologies for the duplicate pull request, decided to move changes into a separate branch
## References

Adds the defaultPath option referenced in https://github.com/jupyterlab/jupyterlab/issues/12929

## Code changes

Adds a new optional `defaultPath` option to `IDirectoryOptions`. This option is then passed to `createFilteredFileBrowser` where it uses `defaultPath` to change the directory of the model. The resulting file dialog opens to the given path instead of root.

Also added two new tests to verify `defaultPath` functionality.
Within a folder in root, a new file and folder are added as shown:

```
.
├── Untitled Folder/ <---- defaultPath
│   ├── Untitled File
│   └── Untitled Folder
├── Untitled File
└── Untitled Notebook
```

The tests check that the parent directory of the selected file/folder matches the given `defaultPath`.
## User-facing changes

None

None

## Additional Question(s)
There is a lint error regarding floating promises. Since `model.cd()` returns `Promise<void>` and doesn't throw an error, would it be safe to mark the call as intentionally not awaited using the void operator?


packages/filebrowser/src/opendialog.ts
```diff
@@ -246,6 +259,11 @@ namespace Private {
       refreshInterval: options.refreshInterval,
       filterDirectories
     });
+
+    if (defaultPath) {
+      model.cd(defaultPath);
```
krassowski, line 264: Ideally this would get an `await` to ensure that the path change has enough time to propagate in the model (especially if implementation of `cd` changes in the future). For that to happen the `createFilteredFileBrowser` function will need to be changed to `async` and respective changes propagated up to where it is used, for example instead of:

```
layout.addWidget(this._browser);
```

we will want to call widget on value the promise resolves to. Similarly in `getOpenFiles` (which fortunately is already async) we should only initialise `Dialog` once `OpenDialog` is ready. We could add a public `OpenDialog.ready: Promise<void>` to know if it is ready.

fcollonval (approved):
Thanks @mmichilot

fcollonval:
CI failures are not related

packages/filebrowser/src/opendialog.ts
```diff
@@ -125,57 +140,68 @@ class OpenDialog
...
+        // Set Widget content
+        this.layout = layout;
 
-    // Set Widget content
-    this.layout = layout;
+        this._ready.resolve();
+      })
+      .catch(() => {
+        console.error('Error while creating file browser in open dialog');
+        this._ready.reject(void 0);
```
mmichilot, line 203: Wasn't sure how to approach the error handling. Would it be best to also pass an error message during the rejection?

packages/filebrowser/src/opendialog.ts
```diff
@@ -125,57 +140,68 @@ class OpenDialog
...
-    layout.addWidget(this._browser);
+        // Set Widget content
+        this.layout = layout;
 
-    // Set Widget content
-    this.layout = layout;
+        this._ready.resolve();
+      })
+      .catch(() => {
+        console.error('Error while creating file browser in open dialog');
```
krassowski, line 202:
```suggestion
      .catch(reason => {
        console.error('Error while creating file browser in open dialog', reason);
```

packages/filebrowser/src/opendialog.ts
krassowski: I do not have an opinion here. I think it is fine to merge as-is and only expand if someone asks for it. However, please look at a suggestion above to expose the failure reason in console for easier debugging.

PR #15286, opened 2023-10-21
Keep active cell rendered in the `full` windowed mode

### LINKED ISSUE #15073: Cannot interact with/in a cell when scrolled out of view (windowed notebook)
## Description

With `full` windowed notebook mode when I edit a cell and scroll up to see variable name, and then try to continue typing, the focus is no longer on the cell and no text gets typed.

## Reproduce

1. Create long notebook
2. Start typing in a cell
3. Scroll it away from view (with some margin)
4. Continue typing
5. See nothing happens
6. Scroll back
7. See no text got typed and the cell still does not accept input

## Expected behavior

When we start typing, the notebook should scroll active cell into view and allow typing.

## Context

- Chrome
- JupyterLab version: 4.0.5

krassowski:
## References

Fixes #15073

## Code changes

- Adds `activeCell` to `NotebookWindowedLayout`.
- Instead of using `display: none` uses `opacity` + `height` to "hide" the cell without detaching its event handlers
- Ensure that typing in the cell scrolls to the active cell (by updating `ensureFocus` which is invoked on `keydown` so that it checks if the cell is in the viewport, not only if it was previously rendered)

### Before

before

### After

after

I presume none because `Notebook.layout` is typed as `WindowedLayout` rather than `NotebookWindowedLayout` which is an implementation detail, but let me know if you consider it to be public.

### DIFF
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -179,10 +179,11 @@ export class NotebookWindowedLayout extends WindowedLayout {
     // Status may change in onBeforeAttach
     const wasPlaceholder = (widget as Cell).isPlaceholder();
     // Initialized sub-widgets or attached them for CodeCell
-    if (this.parent!.isAttached) {
+    // Because this reattaches all sub-widget to the DOM which leads
+    // to a loss of focus, we do not call it for soft-hidden cells.
+    if (this.parent!.isAttached && !this._isSoftHidden(widget)) {
       MessageLoop.sendMessage(widget, Widget.Msg.BeforeAttach);
     }
-
     if (!wasPlaceholder && widget.node.parentElement) {
       if (this._isSoftHidden(widget)) {
         // Restore visibility for active, or previously active cell
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -184,17 +184,20 @@ export class NotebookWindowedLayout extends WindowedLayout {
     if (this.parent!.isAttached && !this._isSoftHidden(widget)) {
       MessageLoop.sendMessage(widget, Widget.Msg.BeforeAttach);
     }
-    if (!wasPlaceholder && widget.node.parentElement) {
-      if (this._isSoftHidden(widget)) {
-        // Restore visibility for active, or previously active cell
-        this._toggleSoftVisibility(widget, true);
-      } else if (widget instanceof CodeCell) {
-        // We don't remove code cells to preserve outputs internal state
-        widget.node.style.display = '';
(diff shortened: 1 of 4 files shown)


packages/notebook/src/windowing.ts
```diff
@@ -235,6 +251,17 @@ export class NotebookWindowedLayout extends WindowedLayout {
       !widget.node.classList.contains(DROP_SOURCE_CLASS) &&
       widget !== this._willBeRemoved
     ) {
+      // Note: `index` is relative to the displayed cells, not all cells,
+      // hence we compare with the widget itself.
+      if (widget === this.activeCell) {
```
fcollonval, line 256: The condition lines 248-252 must be changed because this will only apply to `CodeCell` currently.

fcollonval (changes requested):
Thanks @krassowski 

It works well but only for code cell but I think we want to apply it for any cells (I guess it is not worth the trouble for markdown cell to distinguish if they are rendered or not).

brichet:
Thanks @krassowski.

Concerning the PR https://github.com/jupyterlab/jupyterlab/pull/14115, some changes are about setting the focus on cell node instead of notebook node. This includes a mechanism to keep the focus (and the tabindex=0) on a cell when the active cell is detached from windowing: https://github.com/jupyterlab/jupyterlab/commit/67e12ae268f95364a12ced5b0fa589e00c3ef71c#diff-5d4be2d463ce0219eead775eeb3bbc80160d57d64673559fc0e928640637ff6aR2103-R2109

If I understand correctly, we won't have to move the focus anymore since the active cell will not be detached, and will keep the focus and its attributes. Is that right ?
I don't know which PR will be merged first, but we have to bear in mind that the other will have to be updated.

krassowski:
> If I understand correctly, we won't have to move the focus anymore since the active cell won't be detached, and will keep the focus and its attributes. Is that right ?

Yes with this PR in, focus management will be simpler as the active cell will never be detached from windowing. The difference in the approach in this PR (compared to tracking focus) is that we ensure that all aspects of state of active cell will be preserved.

packages/notebook/src/windowing.ts
krassowski: Thanks! I added a test case, and I tried moving the lines as suggested in da7e71201c569d5a8b985aa5018096db76f2c956, but moving that condition down meant that we never detach markdown/raw cells (as caught by tests) so instead I moved out the added condition from lines 256+ one level up in 15a3569684e0560dffe2c94851fddd137d247bd6.

fcollonval (approved):
Thanks @krassowski

PR #15288, opened 2023-10-22
Fix overreactive scrolling to next cell after `Shift + Enter`

### LINKED ISSUE #14878: Switch off the new overreactive interface jump to end of output behaviour?
Since updated jupyter-lab recently the view sometimes suddenly jumps to the bottom of the output of the next cell. I am not sure what exaclty that triggers, but it might be intended as a new 'feature'. I find this behaviour rather disorienting and would prefer to scroll manually. Is there an option to switch these new behaviour off?

I just merged some cells and all of a sudden the view jumps three to five cells below? Maybe it is a bug?

Version 4.0.3

krassowski:
## References

Fixes #14878

## Code changes

- improves logic of `getOffsetForIndexAndAlignment` for `smart` and `auto` modes

## User-facing changes

When an item larger than the viewport is scrolled to:
- instead of scrolling to its bottom, stay in the current position if the long item was already non-marginally visible
- if it is over one viewport away, instead of centring on it, scroll to its edge

> Note: while centering a small cell in the viewport is good UX, it does not work with very long cells, as then the user would be shown a seemingly random section of the cell, which at times would be (depending on relative sizes) the middle of the output, the middle of the editor, or a mix of both

### Before

before

### After

after

None


fcollonval:
Thanks a lot @krassowski for working on this.

Using you test notebook with full windowing I hit a case I think we want to avoid:

Peek 2023-10-24 10-42

What do you think about changing the `auto` case as following:

if the item size is smaller than the viewport height - margin => keep the logic in master; i.e. don't scroll if it is in the viewport - scroll to display bottom + margin if it is not displayed - else scroll to on top - margin

if the item size is larger than the viewport height - margin => don't scroll if it is already covering the majority of the height (unsure what criteria to use); scroll to bottom + margin if the item is above the viewport; scroll to top - margin if the item is below the viewport.

fcollonval:
Let's me know if I can help - I have time this week and the next one.

krassowski:
Thank you for checking it out! Lets look at it visually, considering only the case of advancing to the next cell upon execution of active cell with<kbd>Ctrl</kbd> + <kbd>Shift</kbd> or navigation down to next cell using <kbd>Arrow Down</kbd>.

### The problem with this PR

I belive that what you see is the case when the the cell is "marginally visible" (i.e. interpreted as not visible enough to matter).

| Before action | Proposed by @fcollonval | This PR right now | 
| --- | --- | --- |
| image | image | image |  

I agree that the behaviour that you propose is a better UX that what this PR behaves like in this specific case (when the cells is considered to be marginally visible by the logic in this PR). I believe that we can implement this change (alignment to top when navigating down) while keeping the remaining logic changes in this PR.

Another improvement that we could have is changing the definition of "margin" from percentage-based to pixel-based, equal to about the size of one empty code cell (but it could happen independently).

| Before action | Proposed by @krassowski | Proposed by @fcollonval |
| --- | --- | --- |
| image | image | image |

### Other trade-offs with the solution proposed by @fcollonval

> if the item size is smaller than the viewport height - margin => keep the logic in master; i.e. don't scroll if it is in the viewport - scroll to display bottom + margin if it is not displayed - else scroll to on top - margin

The question is what behaviour we want for partially visible cells which are smaller than the viewport.

You suggest that upon action the new cell should be aligned with the bottom. I suggest that it should not scroll to minimise distractions:

| Before action | Proposed by @fcollonval | This PR right now |
| --- | --- | --- |
|  image | image | image |

> if the item size is larger than the viewport height - margin => don't scroll if it is already covering the majority of the height); [...] scroll to top - margin if the item is below the viewport.

In case of a long cell you suggest that upon execution of a cell above with Ctrl + Shift or navigation with Arrow Down it should be:
- left as-is if it covers majority (say threshold 50%) of the viewport
- is aligned with the top if it does not

Here the long cell below the active one covers <50% of the viewport:

| Before action | Proposed by @fcollonval | This PR right now |
| --- | --- | ---|
| image | image | image |

To summarise, I agree that:

> if the item size is larger than the viewport height - margin => [... ] scroll to bottom + margin if the item is above the viewport; scroll to top - margin if the item is below the viewport.

is the preferred behaviour, but the remaining suggestions may lead to more aggressive scrolling which may or may not be desirable. What do you think? Should we go for the compromise that I am proposing, or make it a user preference, or have a vote on Wednesday/call a user survey?

krassowski:
Interestingly, both behaviours were present in JupyterLab 3, depending on the user action.

For execute-and-advanced (<kbd>Shift</kbd> + <kbd>Enter</kbd>) JuptyterLab 3.x behaved as proposed by myself (preventing scrolling at all cost):

| Long cell | Short cell |
|--|--|
| long-cell-below | short-cell-below gif |

For navigation with <kbd>Arrow Down</kbd>, JuptyterLab 3.x behaved as proposed by @fcollonval (scrolling to the edge):

| Long cell | Short cell |
|--|--|
| long-cell-arrow | short-below-arrow |

fcollonval:
Thanks a lot for the deep analysis @krassowski 

Due to the heterogeneous behavior in 3.x, I would say it is worth discussing it tomorrow.

I'm more in favor of homogeneous behavior that you are proposing.

> One element that may be relevant is also the size of the input area versus the cell height. But this will increase the complexity of something already hard especially as scrolling is currently meant to be unaware of the item content.

krassowski:
Summarising the discussion from the meeting:
- the conservative course of action would be to closely  re-implement behaviour resembling that of JupyterLab 3.x
- there is some appeal in having unified behaviour as it would reduce the code complexity
- the dual behaviour has a benefit in enabling navigation to top/bottom edge when navigating with arrow keys
- users may prefer one or the other, so maybe it should be a setting
- especially users with accessibility needs may have [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) enabled

For now I will:
- fix the margin to ensure that if first line of code cell is visible, no scrolling appears
- ensure that navigation down results in top alignment
- switch the cell navigation to use smart edge alignment

krassowski:
Here is an edge case: if outputs arrive late (e.g. visualisation with a few seconds delay), the next cell may be hidden. This is because when the scroll logic was executed, it was visible. I think that would be best addressed in a follow-up PR, just noting it down for now.

edge-case

galata/test/jupyterlab/notebook-scroll.test.ts
```diff
@@ -52,3 +53,77 @@ test.describe('Notebook Scroll', () => {
...
+test.describe('Notebook scroll on execution', () => {
+  test.beforeEach(async ({ page, tmpPath }) => {
+    await page.contents.uploadFile(
+      path.resolve(__dirname, `./notebooks/${longOutputsNb}`),
+      `${tmpPath}/${longOutputsNb}`
+    );
+
+    await page.notebook.openByPath(`${tmpPath}/${longOutputsNb}`);
+    await page.notebook.activate(longOutputsNb);
+  });
```
krassowski, line 66:
```suggestion
  });

  test.afterEach(async ({ page, tmpPath }) => {
    await page.contents.deleteDirectory(tmpPath);
  });
```

fcollonval:
> I think that would be best addressed in a follow-up PR, just noting it down for now.

Yes especially because it is though to determine what is the correct behavior for that case.

fcollonval (approved):
Huge thanks @krassowski

krassowski:

krassowski:

Issue #15290, opened 2023-10-23
Creating files from custom drives creates them in the default file browser


jtpio:
## Description

The current file browser context menu entries defined here (and elsewhere as well):

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/filebrowser-extension/schema/browser.json#L129-L138

Currently use `.jp-DirListing-content` as the `selector`.

This works fine for the default file browser. But as a side-effect, custom drives reusing the same structure by instantiating the widget with `browser.createFileBrowser()` (`browser` being a `IFileBrowserFactory`) inherit these menu items.

These menu items refer to commands that are specific to the default file browser, such as creating a new file or folder:

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/filebrowser-extension/src/index.ts#L1150-L1160

These operate on the default file browser.

So when a user interacts with a custom drive, for example [jupyterlab-filesystem-access](https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access), they create new files and notebooks but these end up in the default file browser instead of the custom drive:

https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/assets/591645/2f64fd92-c346-45b6-8d41-5ffe6ffc2326

## Reproduce

Example of a custom drive where this is an issue: https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access

Which can be tested here (Chrome only for now): https://jupyterlab-filesystem-access.readthedocs.io/en/latest/lite/lab/

## Expected behavior

The default context menu should not be visible in other drives reusing the same file browser widget structure.

A solution could be narrow the selector down so it's specific to the default file browser.

## Context

- Operating System and version: Any
- Browser and version: Any
- JupyterLab version: 4.0.7

Related issue: https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/issues/58

jtpio:
This observation also applies to other selectors such as `.jp-DirListing-header` and `.jp-DirListing-item`.

jtpio:
It seems to be possible to access the drive name here:

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/filebrowser-extension/src/index.ts#L1150-L1160

Using:

```ts
tracker.currentWidget.model.driveName
```

Issue #15289, opened 2023-10-23
Accessing files of a custom drive from a kernel


jtpio:
### Problem

As noticed and discussed in https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091, JupyterLab 4 does not store the drive name as a prefix on the session path anymore.

In JupyterLab 3, this drive name prefix used to be passed along when starting a new session with `/api/session`. That way the server could technically use this information and pass it to the kernel to allow accessing files from the kernel in a seamless way.

Also opening a notebook from a custom drive and executing the following cell would not list the content from the custom drive, which can be quite confusing to most users:

```py
import os
os.listdir()
```

### Proposed Solution

Drives are convenient ways to add more sources of contents to JupyterLab via frontend plugins. They also work well in other lab-based applications such as JupyterLite: https://github.com/jupyterlite/jupyterlite

Find a way to integrate the drive feature of JupyterLab so it plays well with other components of the Jupyter stacks.

### Additional context

Related discussions and issues:

- https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091
- https://github.com/jupyterlite/jupyterlite/pull/1227
- https://github.com/jupyterlab/jupyterlab/issues/15143

echarles:
> Also opening a notebook from a custom drive and executing the following cell would not list the content from the custom drive, which can be quite confusing to most users:

Let's imagine a `S3Drive` is defined on the frontend and its content is listed by the server content API. This leads me already to a first question: the relationship between the frontend drives and the content API.

The user double clicks on a listed notebook, and JupyterLab displays it and of course starts a python kernel. In you example, the òs.listdir()` python command run by the user in a S3 notebook cell will return the files and folders of the kernel, not the S3 files and folders... Second question: the relationship between the frontend drives and the kernels "drives" (mount point?).

What command would be supposed to list S3? In terms of user-experience, this is very confusing. I think we need to be clear on the boundaries between content API and kernel.

Being clear does not mean that we can not find a way to leverage the Frontend Drives in the Content API (enriching the API with an additional paramater?) and the Kernel (with mount points per drive??, the responsibility to mount them relying on the server/kernellauncher ?).

davidbrochart:
> Let's imagine a `S3Drive` is defined on the frontend and its content is listed by the server content API. This leads me already to a first question: the relationship between the frontend drives and the content API.

[...]

PR #15291, opened 2023-10-23
Fix creating files in custom drives, fix `ContentsManagerMock`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15290
Fixes https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/issues/58
Fixes https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/issues/22
Fixes https://github.com/jupyterlab-contrib/jupyterlab-browser-storage/issues/1

## Code changes

The file browser widget should pass along its drive name when using the contents manager for manipulating files.

- [x] Fix creating new files from the context menu
- [x] Check other file manipulation operations such as delete, rename...
- [x] Fix creating notebooks
- [x] Fix `ContentsManagerMock` to have basic support for drives

## User-facing changes

**Before**

https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/assets/591645/2f64fd92-c346-45b6-8d41-5ffe6ffc2326

**After**

https://github.com/jupyterlab/jupyterlab/assets/591645/cfcf4962-416a-43df-8849-468f320fa043

None

### DIFF
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -255,7 +255,8 @@ export class FileBrowser extends SidePanel {
   ): Promise<Contents.IModel> {
     // normalize the path if the file is created from a custom drive
     if (options.path) {
-      options.path = this._toDrivePath(this.model.driveName, options.path);
+      const localPath = this._manager.services.contents.localPath(options.path);
+      options.path = this._toDrivePath(this.model.driveName, localPath);
     }
     try {
       const model = await this._manager.newUntitled(options);
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -253,9 +253,9 @@ export class FileBrowser extends SidePanel {
   private async _createNew(
     options: Contents.ICreateOptions
   ): Promise<Contents.IModel> {
-    // normalize the path if create a file from a custom drive
+    // normalize the path if the file is created from a custom drive
     if (options.path) {
-      options.path = this._toGlobalPath(this.model.driveName, options.path);
+      options.path = this._toDrivePath(this.model.driveName, options.path);
     }
     try {
       const model = await this._manager.newUntitled(options);
@@ -409,14 +409,15 @@ export class FileBrowser extends SidePanel {
   }
 
   /**
-   * Given a drive name and a local path
+   * Given a drive name and a local path, return the full
+   * drive path which includes the drive name and the local path.
    *
    * @param driveName: the name of the drive
    * @param localPath: the local path on the drive.
    *
-   * @returns the fully qualified path.
+   * @returns the full drive path
    */
-  private _toGlobalPath(driveName: string, localPath: string): string {
+  private _toDrivePath(driveName: string, localPath: string): string {
     if (driveName === '') {
       return localPath;
     } else {
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -2,6 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { showErrorMessage } from '@jupyterlab/apputils';
+import { PathExt } from '@jupyterlab/coreutils';
 import { IDocumentManager } from '@jupyterlab/docmanager';
 import { Contents, ServerConnection } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
@@ -252,6 +253,10 @@ export class FileBrowser extends SidePanel {
   private async _createNew(
     options: Contents.ICreateOptions
   ): Promise<Contents.IModel> {
+    // normalize the path if create a file from a custom drive
+    if (options.path) {
+      options.path = this._toGlobalPath(this.model.driveName, options.path);
+    }
     try {
       const model = await this._manager.newUntitled(options);
       await this.listing.selectItemByName(model.name, true);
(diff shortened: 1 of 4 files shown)


packages/filebrowser/src/browser.ts
```diff
@@ -394,6 +399,22 @@ export class FileBrowser extends SidePanel {
...
 
+  /**
+   * Given a drive name and a local path
+   *
+   * @param driveName: the name of the drive
+   * @param localPath: the local path on the drive.
+   *
+   * @returns the fully qualified path.
+   */
+  private _toGlobalPath(driveName: string, localPath: string): string {
```
echarles, line 410: The docstring says `@returns the fully qualified path.` and the method is called `toGlobalPath`.

I wonder how that combination of driveName:localPath should be called?

jtpio: Ah this method was copied from here for now:

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/services/src/contents/index.ts#L974-L990

Maybe this should be made reusable. Although we probably want to avoid adding a new method to the interface to avoid breaking changes if that new method is not optional.

echarles: I would say let's name it as it should (toDriveURI, or whatever) and make it public as optional for V4, making it mandatory for V5?

jtpio: Having a second look at it, we could also just use a different method here since it would only be used by the browser widget. This would likely help keep things simpler to have this patch in a `4.0.x` release.

Also the one in the services package takes a `Contents.IDrive` as a first a parameter instead of a `string`.

But agree we should try to improve that, so normalizing the drive paths becomes accessible to other plugins. Happy to open an issue about this.

echarles: Don't want to block this PR so OK to address only the BrowserWidget. 

We already have https://github.com/jupyterlab/jupyterlab/issues/15289 that is looking more toward kernels, we may rename it to have a broader discussion, including the drive paths nomenclature.

packages/filebrowser/src/browser.ts
```diff
@@ -252,6 +253,10 @@ export class FileBrowser extends SidePanel {
   private async _createNew(
     options: Contents.ICreateOptions
   ): Promise<Contents.IModel> {
+    // normalize the path if the file is created from a custom drive
```
krassowski, line 256: Is there an easy way to add a unit test for this?

jtpio: Hoping so yes.

Gonna first rebase to grab the UI tests fixes from https://github.com/jupyterlab/jupyterlab/pull/15524, to make sure this change does not break other things.

krassowski: Just following up on the question about a unit test here - do you think it could be added in this PR?

jtpio: So the UI tests are now passing after https://github.com/jupyterlab/jupyterlab/pull/15524.

Hopefully it should be possible to add a simple unit test yes. Maybe would need to double check if the current testutils mock correctly handle drives too.

krassowski: Did you plan to add a unit test in this PR or would you prefer it to be merged without a test?

fcollonval:
We should check if this proposal work with the collaboration package.

jtpio:
We seem to have `jupyter-collaboration` available on the dev mode binder:

https://github.com/jupyterlab/jupyterlab/blob/1a19d634695d5d7a429ef4786fa271f40bfc5e72/binder/environment.yml#L63-L65

So maybe we can test it on the Binder for this PR directly: https://mybinder.org/v2/gh/jtpio/jupyterlab/fix-create-files-custom-drives?urlpath=lab

jtpio:
I'll try to see if we can get this merged because this is also affecting https://github.com/jupyterlab-contrib/jupyterlab-browser-storage.

trungleduc (approved):

jtpio:
Looks like the `ContentsManagerMock` does not really handle drives correctly at the moment:

https://github.com/jupyterlab/jupyterlab/blob/f4c011e4844c4cfe1c8ded5f517aa736c0d7d190/packages/services/src/testutils.ts#L306-L499

jtpio:
> Looks like the `ContentsManagerMock` does not really handle drives correctly at the moment:

Pushed a couple of commits so this `ContentsManagerMock` can have basic support for drives.

jtpio:
> I'll try to see if we can get this merged because this is also affecting https://github.com/jupyterlab-contrib/jupyterlab-browser-storage.

Trying locally with this extension seems to be working fine:

[create-new-notebook-drive.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/c4fd38d6-e508-4731-b1ad-e7e60c889ab4)

packages/notebook-extension/src/index.ts
```diff
@@ -1957,7 +1962,9 @@ function activateNotebookHandler(
     caption: trans.__('Create a new notebook'),
     icon: args => (args['isPalette'] ? undefined : notebookIcon),
     execute: args => {
-      const cwd = (args['cwd'] as string) || (defaultBrowser?.model.path ?? '');
+      const currentBrowser =
+        filebrowserFactory?.tracker.currentWidget ?? defaultBrowser;
```
krassowski, line 1966: If this pattern is needed across the board to support custom drives, we should probably document it for extension authors. I would suggest either pushing a small docs update or opening an issue to track doing it later.

krassowski (approved):
Thanks! One suggestion on documenting the pattern used to grab the browser model.

packages/notebook-extension/src/index.ts
jtpio: Yes, just opened https://github.com/jupyterlab/jupyterlab/issues/15939.

Also there are still a few fixes needed for drives in general, such as implementing support for checkpoints in the `ContentsManagerMock`, which could be done as part of https://github.com/jupyterlab/jupyterlab/issues/15930 (and other linked issues).

krassowski:

krassowski:

jtpio:
Maybe it's also fine to have this in 4.1.x and not backport it to 4.0.x, if it can ease maintenance.

PR #15295, opened 2023-10-24
Bump yarn to 3.6.4

fcollonval:
## References

Maintenance to bump to the latest Yarn v3+

I stayed away from the new [v4](https://yarnpkg.com/blog/release/4.0). We may try it after JupyterLab 4.1

## Code changes

Bump vendored yarn to 3.6.4

## User-facing changes

None

None


bollwyvl:
> I stayed away from the new [v4](https://yarnpkg.com/blog/release/4.0). We may try it after JupyterLab 4.1

While the performance benefits of 4 are tempting, we'd need some canary testing of a number of complex downstreams to assess whether this buys anything valuable... the transition from 1 to 3 still hasn't demonstrated a lot of value to me, as we upgraded in an entirely-breaking way (CLI, lockfile, etc), but didn't take many of its new opinions which _might_ have contained those benefits (zero install).

More broadly: still in favor of removing the `jlpm` blob (and `jupyter lab(extension) build`) entirely from the user-facing package which provides increasingly less value to users as the maintainer community makes less use of `jupyter lab build`.

It _would_ be worth it to keep it in a super-cut-down `jupyterlab-builder` PyPI package (_not_ `hatch-jupyterlab-builder`), suitable for use in a `build-system`, which would allow for potentially supporting multiple target versions of `jupyterlab`, which would act as a more resilient bridge.

krassowski:
Something else to note here - with immutable builds the downstream extensions will need to refresh lock files after updating JupyterLab (even if they don't bump any npm dependencies). Not a huge pain but I saw reports of "this lockfile is different for me locally" because an earlier point release of yarn was shipped in JupyterLab RC phase.

krassowski:
For anyone interested: https://github.com/yarnpkg/berry/compare/%40yarnpkg/cli/3.5.0...%40yarnpkg/cli/3.6.4 there are a few performance improvements and a few bugfixes.

krassowski (approved):

jtpio:
> More broadly: still in favor of removing the `jlpm` blob entirely from the user-facing package

Curious: removing the `jlpm` blob just from the user facing package, or entirely?

If it now becomes tedious for extension authors relying on `jlpm` to update their `yarn.lock` on minor versions, I'm wondering if we wouldn't be better off dropping `jlpm` entirely, and let extension authors choose whatever they want (`npm`, `pnpm` or something else).

Happy to open an issue to discuss that more broadly if there is interest.

bollwyvl:
> update their `yarn.lock` on minor versions

They have to do this anyway if rebuilding against a new version?

At any rate, getting it (and all code related to `nodejs`) out of the `jupyterlab 5.x` distribution altogether would be magnificent for 5.x, as 99% of users don't need it anymore.... i thought there was an issue, but can't find it.

krassowski:
> They have to do this anyway if rebuilding against a new version?

The point is that this breaks CI even if not rebuilding nor updating versions (unless you pinned your jupyterlab to 4.0), see https://github.com/jupyter/notebook/pull/7170 as an example.

bollwyvl:
So, in the near term, perhaps we roll back this change. I still haven't heard the compelling reason to use `.4`, and _breaks `notebook`_ is a strong mark against.

> dropping `jlpm` entirely

I would maintain that having _a_ working, (apparently-not-very) tested system _available_ is, and will continue to be, crucial.

I blinked, so may have missed the new js package manager that Fixes All The Problems. Provided that _didn't_ happen, and the opinion of the week rules, then _we_ would inherit supporting _all_ of the problems with _all_ the systems on _all_ the operating systems for _all_ of the users that happen to type `jupyter labextension install`. 

Thus far, we have at least _partially_ mitigated this by going through the python-declared `entry_points` route with a white-label installer that relies on:

- the existence of `nodejs` on `$PATH`
- a populated `node_modules` 
- something that respects `package.json#/resolutions`
- a stable lockfile (but guess not, right?)

... _none_ of which are guaranteed by any of the other systems.

A potential solution, even in 4.1, is to offer an expansion of the schema to eject from all the [`jlpm`](https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/federated_labextensions.py#L282) defaults, which might look something like, via dumb-old-env-vars:

```json
{
  "jupyterlab": {
    "builder": {
      "scripts": {
        "install": ["${JLAB_NODEJS}", "${JLAB_YARN}"],
        "build": ["${JLAB_NODEJS}", "${JLAB_BUILDER_JS}"],
        "watch": ["${JLAB_NODEJS}", "${JLAB_YARN}", "watch"],
      },
      "needs": {
        "install": ["${JLAB_NODEJS}", "${JLAB_YARN}"],
        "build": ["${JLAB_NODEJS}", "${JLAB_YARN}", "${JLAB_BUILDER_JS}"],
        "watch": ["${JLAB_NODEJS}", "${JLAB_YARN}", "${JLAB_BUILDER_JS}"],
      }
    }
  }
}
```

If someone has a toolchain that accepts the same arguments as `@jupyterlab/builder` (`--core-path`, etc), and generates all the correct `package.json` metadata in the right places after the fact, without rebundling an incompatible version of `@lumino/widgets`, go to town!

PR #15296, opened 2023-10-24
Fix rulers position with gutter width

### LINKED ISSUE #15292: Ruler lines are at wrong places in a text editor
## Description

The vertical ruler lines are at incorrect columns, at least in the case when line numbers are shown. 

For example,
- If line numbers have two digits, the ruler lines are 2 to 3 characters to the left of the place where they are supposed to be.
- If line numbers have three digits, the ruler lines are 3 to 4 characters to the left.

## Reproduce

Just set the one or more rulers and open an text file. 

## Context

- Operating System and version: Server: Debian 11, Ubuntu 22.04, Client: Windows 11
- Browser and version: Edge 118.0.2088.57, Chrome 118.0.5993.89, Firefox 118.0.2.
- JupyterLab version: 4.0.7

## Screenshot 

image

User preference.
```
{
    "scrollPastEnd": true,
    "toolbar": [],
    "editorConfig": {
        "cursorBlinkRate": 530,
        "lineWrap": "on",
        "rulers": [
            10,
            20,
            80,
            96
        ],
        "fontFamily": null,
        "fontSize": null,
        "insertSpaces": true,
        "lineHeight": null,
        "readOnly": false,
        "showTrailingSpace": true,
        "wordWrapColumn": 80
    }
}
```

fcollonval:
## References

Fixes #15292

## Code changes

Take into account the gutters width when positioning the rulers
Peek 2023-10-24 16-28

## User-facing changes

rulers are correctly positionned even when gutters are defined

None


krassowski:
Is it worth adding a visual test with editor populated like on https://codemirror.net/5/demo/rulers.html?

image

krassowski (approved):
Thank you @fcollonval tested locally, works well also when there is a debugger gutter.

> Test failures unrelated. Maybe we should just disable cell toolbar by default in tests.

krassowski:

PR #15297, opened 2023-10-24
Fix autobrackets and other default CM extension

### LINKED ISSUE #14559: Deleting brackets with auto-close-brackets enabled only deletes one of the brackets (jupyterlab 4.0)
## Description

As in the title, deleting brackets with auto-close-brackets enabled in jupyterlab 4.0, only deletes one of the brackets. In previous verions deleting brackets/parenthesis/braces deleted both. 

## Reproduce

Deleting brackets/braces/parethesis in jupyterlab 4.0.

## Expected behavior

Deleting one of the brackets with auto-close-brackets enabled should delete both as in previous versions of jupyterlab.

## Context

- Operating System and version: Linux mint 21.1
- Browser and version: firefox 113
- JupyterLab version: 4.0.0

fcollonval:
## References

Fixes #14559

looking at the [basic setup](https://github.com/codemirror/basic-setup/blob/b3be7cd30496ee578005bd11b1fa6a8b21fcbece/src/codemirror.ts#L69) from CM, this adds also:
- [highlightSpecialCharacters](https://codemirror.net/docs/ref/#view.highlightSpecialChars)

There is one I'm unsure about:
- [highlightSelectionMatches](https://codemirror.net/docs/ref/#search.highlightSelectionMatches)

I'm not confident about that one as it won't work for notebook (only for file editor). And I guess the LSP extension as something better for it.

cc @krassowski

## Code changes

Add CM extensions: `closeBracketsKeymap`, `highlightSpecialChars`

## User-facing changes

- Backspace between two brackets deletes both if auto closing option is on
- [Special characters](https://github.com/codemirror/view/blob/e6a55a6515521a84e709f02b4eb32e57aebbbfba/src/special-chars.ts) will now be highlighted with red dots, e.g. `­؜​‎‏  `

Before
Screenshot from 2023-10-26 21-51-01

After
Screenshot from 2023-10-26 21-51-08

None

### DIFF
--- a/packages/codemirror/src/extension.ts
+++ b/packages/codemirror/src/extension.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { closeBrackets } from '@codemirror/autocomplete';
+import { closeBrackets, closeBracketsKeymap } from '@codemirror/autocomplete';
 import { defaultKeymap, indentLess } from '@codemirror/commands';
 import {
   bracketMatching,
@@ -13,6 +13,7 @@ import {
   Compartment,
   EditorState,
   Extension,
+  Prec,
   StateEffect
 } from '@codemirror/state';
 import {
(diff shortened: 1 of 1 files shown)


krassowski:
I agree that inclusion of `highlightSpecialCharacters` is a good call because it might help catch some security/correctness issues and likely has little overhead. I do not have a strong opinion on `highlightSelectionMatches` but I would worry that searching for selection could have slight performance penalty in large documents - I guess it could be exposed as disabled by default.

krassowski (approved):
Tested locally, fixes the issue.

fcollonval:

Issue #15046, opened 2023-08-30
Read-only status notification toast should show up on `ctrl + s`


krassowski:
## Description

https://github.com/jupyterlab/jupyterlab/pull/14600 implements a read-only indicator and a toast notification when user attempts to save the file with "save all" (see https://github.com/jupyterlab/jupyterlab/issues/7980 for design discussion). Notification was not implemented for "save" command as that command is disabled when file is read-only.

## Reproduce

- Open a readonly file in non-RTC mode (create one with chmod 444 file.py)
- Press `ctrl+s`
- See no notification

## Expected behavior

The notification is shown anyways. We could:
- (a) add a new command or just listen to keystrokes and synchronise the keyboard shortcuts but this may be unreliable.
- (b) add a new argument `isKeybinding` to command's `args`, which would allow to conditionally enable the save command when used from shortcut and then show the notification. We already do that for `toolbar` (#14664) and `isPalette`
- (c) have a bespoke new argument in commands, a callback to execute when the command is disabled, let's call it `notifyDisabled()`

(b) and (c) require changes in Lumino. This is somewhat related to the discussion in https://github.com/jupyterlab/lumino/issues/570.

## Context

- JupyterLab version: 4.1 main

JasonWeill:
Other apps with real-time collaboration and autosaving handle CTRL+S/CMD+S differently:

* Quip does nothing when the user presses the key combination
* OneNote does nothing, although there is an option to manually force a sync

andrewfulton9:
From my understanding, changes in lumino would revolve around [this function](https://github.com/jupyterlab/lumino/blob/a817cb2d05ac264403ad930fb9c038746c5393ff/packages/commands/src/index.ts#L614)

If we go with something like options B above, would the idea be to:
- a) make `isKeybinding` return a boolean and allow the `!this.isEnabled(command, args)` conditional to be bypassed when `isKeybinding` is True in `_executeKeyBinding`
- b) make `isKeybinding` a callable that would be called at the beginning of `_executeKeyBinding` to allow custom code execution
- c) something else I'm not seeing

[...]

Issue #644, opened 2016-08-11
No keyboard way to get input focus in the console


fperez:
If the input area doesn't have focus, there seems to be no keyboard-based way to get it currently.

This happens on console first start (thus requiring a click to start typing) and if for example an `input()` call is made, which will move focus to the stdin area. After entering text, the new input area won't be focused and there's no way to re-focus it with the keyboard.

Suggested fixes:
- after stdin usage and on start, input area should have focus by default, thus not requiring further user action.
- `Enter` should focus it, allowing users to regain focus in other situations that may arise. This is how the notebook works, so it's the most natural thing for users to try.

sccolbert:
Replacing "bug" with "enhancement".

fperez:
no prob, it was borderline, but agreed :)

jasongrout:
On start, the console now focuses the input. After stdin usage and after pressing enter, the cursor is still not focused on the input cell.

jasongrout:
Interestingly, now if you just do *one* `input()`, the input is not focused, but if you do *two* `input(), input()`, the input box *is* focused.

ClaytonPassmore:
@jasongrout I was with you when that weird "one doesn't work but two does" behavior happened but I can't reproduce. Everything is working fine for me on master. Focus goes to the input box(es) as expected. Obviously focus still does not return to the console, though.

ClaytonPassmore:
@blink1073 This is fixed, right? Can we close this issue?

blink1073:
Not quite.  It is better, but we still have some work to do to make this seamless.

PR #15317, opened 2023-10-27
Read only status notification on ctrl+s

andrewfulton9:
## References

adds feature described in #15046 

depends on [lumino #644](https://github.com/jupyterlab/lumino/pull/644) to function

## Code changes

Uses contextual information passed from lumino commands `_executeKeyBinding` via `args` to modify execution if the save is called from a keybinding to raise a notification notifying a user that the file is read-only and can't be saved

## User-facing changes

If a user trys to save  read-only document with the keybinding "ctr+s", a notification will briefly pop up notifying them that the document is read-only and cannot be saved.

https://github.com/jupyterlab/jupyterlab/assets/13774419/da5ae300-ee82-4e96-9f79-cf3da18a3b46

None that I am aware


krassowski:
Kicking CI which failed due to `npm` outage: https://status.npmjs.org/incidents/zdznxkrp22py

packages/docmanager-extension/src/index.tsx
```diff
@@ -857,8 +866,20 @@ function addCommands(
...
-          if (context.model.readOnly) {
+          if (!context.contentsModel?.writable) {
+            let type;
+            if (args._luminoEvent) {
+              type = (args._luminoEvent as ReadonlyPartialJSONObject).type;
+            }
+            if (type === 'keybinding') {
+              return Notification.warning(
+                trans.__(
+                  `%1 is permissioned as read-only. Use "save as..." instead.`,
```
krassowski, line 877: The other notification says: `%1 is permissioned as readonly. Use "save as..." instead.` (readonly vs read-only). Would it make sense to move creation of the warning to a function to ensure that only one version of the string is used?

krassowski (commented):
Could you look at how hard adding an integration test would be here? I does not need to include a snapshot as there is a dedicated `page.notifications` galata helper which could be used, as in:

https://github.com/jupyterlab/jupyterlab/blob/c6e7a14416ebc6c1b7c66ba64423f635a0d3c0a9/galata/test/jupyterlab/announcements.test.ts#L15-L26

andrewfulton9:
> Could you look at how hard adding an integration test would be here? I does not need to include a snapshot as there is a dedicated `page.notifications` galata helper which could be used, as in:
> 
> https://github.com/jupyterlab/jupyterlab/blob/c6e7a14416ebc6c1b7c66ba64423f635a0d3c0a9/galata/test/jupyterlab/announcements.test.ts#L15-L26

@krassowski, I'm a little lost trying to figure out how to make a test notebook read only. Do you have any suggestions on how I should do this?

krassowski:
Good question. I think you that setting immutable attribute could work on CI but this would annoy anyone running tests locally as it would require a manual/OS-specific step before running the test. Probably the best option would be adding a mock that forces the "writable" attribute to false. You can follow the existing mocks such as `freezeContentLastModified`:

https://github.com/jupyterlab/jupyterlab/blob/453adc4fb7d57b60967ce12cbe65a4cc1295de2e/galata/src/galata.ts#L573-L625

galata/src/galata.ts
```diff
@@ -624,6 +624,59 @@ export namespace galata {
...
+              if (!page.isClosed() && !isClosed) {
+                return route.fulfill({
+                  status: response.status(),
+                  body: await response.text()
+                });
+              }
+              break;
+            }
+            const data = await response.json();
+            // Modify the last_modified values to be set one day before now.
```
krassowski, line 661:
```suggestion
```

galata/test/jupyterlab/readonly.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    await page.notebook.createNew('notebook.ipynb');
+
+    // have to open and close notebook for notification to show up
+    await page.notebook.close();
+    await page.notebook.open('notebook.ipynb');
+
+    await page.keyboard.press('Control+s');
+
+    const imageName = 'readonly.png';
+    await page.waitForSelector('.Toastify__toast');
```
krassowski, line 20: Nitpick: a `jp-` class would be preferred here as toastify is implementation detail, but this is ok too.

galata/test/jupyterlab/readonly.test.ts-snapshots/readonly-jupyterlab-linux.png
krassowski, line 1: Should it just show the base name? Or is showing the full path valuable here?

packages/docmanager-extension/src/index.tsx
```diff
@@ -857,13 +876,14 @@ function addCommands(
...
-              buttons: [Dialog.okButton()]
-            });
+          if (!context.contentsModel?.writable) {
+            if (args._luminoEvent) {
+              let type = (args._luminoEvent as ReadonlyPartialJSONObject).type;
+              if (type === 'keybinding') {
+                readonlyNotification(context.path);
+                return;
+              }
+            }
```
krassowski, line 886: Should it fall back to dialog otherwise? In practice it is unlikely to happen (only if `isWritable()` returned `true` but between the check and execution the document changed to read-only) but maybe worth ensuring that it does not enter some odd state?

krassowski (commented):
The test looks good, thank you! Two suggestions/questions on:
- showing just file name not path
- throwing an error if command gets executed when it should not be.

galata/test/jupyterlab/readonly.test.ts-snapshots/readonly-jupyterlab-linux.png
andrewfulton9: I see it as being a tradeoff between full path being more explicit vs just the file name potentially being more tidy. Personally I would lean towards the more explicit option. You probably have a better understanding of JLab users though, so I'm happy to defer to your opinion. It's also worth noting that the same function and input are being used for the notification for the save-all and save commands though. I could see the full path being better for the save-all notification even if we update the save notification to just use the base name.

packages/docmanager-extension/src/index.tsx
andrewfulton9: Good point. I added that showDialog back in and fixed up the logic a bit.

krassowski (approved):
Thanks @andrewfulton9

PR #15318, opened 2023-10-28
Fix scrolling when dragging files in the file browser

### LINKED ISSUE #6959: Moving files to another directory
**Describe the bug**
I cannot move a file from the parent directory to a children directory when the list of files is very long.

- Bug: moving mouse with the dragged file to the top edge (so that I will see the directory I want to drop the file in) does not cause scrolling of the files list view for me.
- Related: I could not find a way to move a file via other means (except for doing this in a console or outside of JuptyterLab); Is there a UI feature I am missing or do we lack an interface for moving files (except for drag-and-drop in filebrowser)?
- Somehow related: having filtering of files by gitignore would alleviate the issue for me; This is a common optional functionality in many IDEs - has anyone considered this here before? Most IDEs I know implements this in an extension, but not sure if it would belong to https://github.com/jupyterlab/jupyterlab-git or a new extension should be created.

**To Reproduce**

1. Open JupyterLab in a folder with more files and directories than it the file browser list 
2. Try to drag a file from the very bottom to the folder at the very top
3. Fail to do so

**Expected behavior**
The file list scrolls so that I can drop the file in the folder.

**Screenshots**
Will try to provide a gif once my gif recorder starts working again

**Desktop:**

- OS: Ubuntu 19.04
- Browser: Chrome 76
- JupyterLab: 1.0.3

**Additional context**

I *think* that it worked in previous versions, but I may be mistaken.

krassowski:
## References

Fixes #6959

## Code changes

Adds `data-lm-dragscroll` which is required by lumino to enable scrolling on edges.

## User-facing changes

| Before | After |
|--|--|
| before | after |

None


fcollonval (approved):
Thanks @krassowski

krassowski:

PR #15331, opened 2023-10-30
Fix update button in extension manager

### LINKED ISSUE #15316: Extension manager "Update to" button does not update extensions
## Description

Clicking the "Update to #.#.#" button in the extension manager does not properly update the extension. 

Screenshot 2023-10-27 at 12 41 09 PM

**Note:** I've tested this on multiple extensions. In this description I'll be using the `jupyterlab-atp-extension` but it seems arbitrary. 

## Reproduce

1. Create a new conda test environment per [guidelines](https://jupyterlab.readthedocs.io/en/latest/getting_started/issue.html)

```bash
conda create -n jlab-test --override-channels --strict-channel-priority -c conda-forge -c nodefaults jupyterlab
```

2. Activate the conda environment

```bash
conda activate jlab-test
```

3. Before starting jupyterlab, install an extension via pip at a version earlier than the extension's latest version on pypi. In my example, I used `jupyerlab-atp-extension==1.0.2` ([pypi link](https://pypi.org/project/jupyterlab_atp_extension/#history)):

```bash
pip install jupyterlab-atp-extension==1.0.2
```

4. Start jupyterlab:

```bash
jupyter lab
```

5. In Jupyterlab, open the Extension Manager via the sidebar:

Screenshot 2023-10-27 at 3 07 04 PM

6. Confirm that the extension is installed and that the extension manager is aware of a more recent version by observing the "Update to #.#.#" button:

Screenshot 2023-10-27 at 12 41 09 PM

7. Click the "Update to #.#.#" button. Accept the Dialog that appears, then refresh as per instructions on the dialog. Notice that the extension has not been upgraded after refresh.

https://github.com/jupyterlab/jupyterlab/assets/111072326/5db1a235-eaa1-47ac-93f2-b2e3ea672a14

## Investigation

When the "Update to" button is clicked, a request is made to server to update: `POST http://localhost:8888/lab/api/extensions?<id>`. The body of the request is:

```json
{
    "cmd": "install",
    "extension_name": "jupyterlab-atp-extension"
}
```

The request is successful and returns 200. However, it seems that the request does not have the desired effect, and the extension stays at the same version. 

I think the issue is that the request's body should include a `"extension_version": "..."`. In [the logic for the handler](https://github.com/jupyterlab/jupyterlab/blob/79a4e38213969a60f74b998330b034fad1598ffc/jupyterlab/handlers/extension_manager_handler.py#L99), it looks for an `"extension_version"` property in the body. If that property exists, the `PyPIExtensionManager` uses it in the underlying `pip install` command it makes (see [here](https://github.com/jupyterlab/jupyterlab/blob/79a4e38213969a60f74b998330b034fad1598ffc/jupyterlab/extensions/pypi.py#L290)). Note how in the jupyterlab logs, we see this command instead, which doesn't include any version:

```bash
[D 2023-10-27 15:19:35.432 LabApp] Executing '/Users/nbowditch/miniconda3/envs/jlab-test/bin/python3.12 -m pip install --no-input --quiet --progress-bar off jupyterlab-atp-extension'
``` 

If I run that pip command in my terminal, it won't upgrade the package since there is no `--upgrade` flag or version specified. But if I change the pip command to include the version (which would be executed if there was an `extension_version` property):

```bash
/Users/nbowditch/miniconda3/envs/jlab-test/bin/python3.12 -m pip install --no-input --quiet --progress-bar off jupyterlab-atp-extension==1.0.3
```

The upgrade occurs as desired. 

Tracing the call from the front-end:

- [Extension manager widget --> ListEntry Button onClick](https://github.com/jupyterlab/jupyterlab/blob/79a4e38213969a60f74b998330b034fad1598ffc/packages/extensionmanager/src/widget.tsx#L132)
- [Extension manager widget --> ListView onAction method](https://github.com/jupyterlab/jupyterlab/blob/79a4e38213969a60f74b998330b034fad1598ffc/packages/extensionmanager/src/widget.tsx#L502).
- [ListView model --> install method](https://github.com/jupyterlab/jupyterlab/blob/main/packages/extensionmanager/src/model.ts#L347)
- [ListView model --> performAction](https://github.com/jupyterlab/jupyterlab/blob/main/packages/extensionmanager/src/model.ts#L484) <-- This is what builds the request

It seems that for both "install" and "update", the logic uses an "install" action. Perhaps there should be an entirely separate "upgrade" action, since the underlying manager (in this case pypi extension manager) might need separate logic paths. However, I think a simple/quick way of fixing this would be to allow an optional `version` parameter in some of these methods. That way the logic can continue to use the same "install" action, just opt to include a "version" (which would be used for updates).

## Expected behavior

The call made to Jupyterlab's server by the Extension Manager should properly upgrade the extension to the appropriate version.

## Context

- Operating System and version: macOS Ventura 13.4
- Browser and version: Chrome Version 118.0.5993.88 (Official Build) (arm64)
- JupyterLab version: 4.0.7

<pre>
[jupyter-troubleshoot.log](https://github.com/jupyterlab/jupyterlab/files/13192885/jupyter-troubleshoot.log)
</pre>
</details>

<pre>
[jupyter-lab-debug.log](https://github.com/jupyterlab/jupyterlab/files/13192887/jupyter-lab-debug.log)
</pre>
</details>

<pre>
 Didn't seem relevant
</pre>
</details>

nbowditch-einblick:
## References

Fixes #15316 Extension manager "Update to" button does not update extensions 

## Code changes

- Added missing `extension_version` parameter to the backend request when the "Update to ###" button is clicked in the extension manager.
   - This was done by adding options with a `useVersion` parameter to the `install` action on the front-end model. The request uses this value to set the `extension_version` property to the request body.
   - For the update button, we use `entry.latest_version`
   - Generic front-end action functions (eg `performAction` and `onAction`) now have an optional `actionOptions` parameter where additional parameters/specifications can be defined

## User-facing changes

None

None


packages/extensionmanager/src/model.ts
```diff
@@ -176,6 +176,16 @@ const EXTENSION_API_PATH = 'lab/api/extensions';
...
 export type Action = 'install' | 'uninstall' | 'enable' | 'disable';
 
+/**
+ * Additional options for an action.
+ */
+export interface IActionOptions {
+  /**
+   * If true, install the latest version of the entry.
+   */
+  useLatestVersion?: boolean;
```
fcollonval, line 186: As the backend support a specific version, let's add that as an option:

```suggestion
  /**
   * Install the version of the entry.
   *
   * #### Note
   * This is ignored by all actions except install.
   */
  useVersion?: string;
```

packages/extensionmanager/src/widget.tsx
```diff
@@ -129,7 +129,11 @@ function ListEntry(props: ListEntry.IProperties): React.ReactElement<any> {
                     <>
                       {ListModel.entryHasUpdate(entry) && (
                         <Button
-                          onClick={() => props.performAction!('install', entry)}
+                          onClick={() =>
+                            props.performAction!('install', entry, {
+                              useLatestVersion: true
```
fcollonval, line 134: This will become:

```suggestion
                              useVersion: entry.latest_version
```

packages/extensionmanager/src/widget.tsx
```diff
@@ -318,7 +326,11 @@ namespace ListView {
      *
      * Not provided if actions are not allowed.
      */
-    performAction?: (action: Action, entry: IEntry) => void;
+    performAction?: (
+      action: Action,
+      entry: IEntry,
+      actionOptions: IActionOptions
```
fcollonval, line 332: let's make this one optional:

```suggestion
      actionOptions?: IActionOptions
```

packages/extensionmanager/src/model.ts
```diff
@@ -343,14 +353,20 @@ export class ListModel extends VDomModel {
...
-   */
-  async install(entry: IEntry): Promise<void> {
-    await this.performAction('install', entry).then(data => {
-      if (data.status !== 'ok') {
-        reportInstallError(entry.name, data.message, this.translator);
+   * @param useLatestVersion Whether to use the latest version to install.
+   */
+  async install(
+    entry: IEntry,
+    useLatestVersion: boolean = false
```
fcollonval, line 360: We prefer to use object that are more flexible to extend an API in a non-breaking way.

```suggestion
    options: { useVersion?: string }
```

> e.g. in the future we may want to add an option `force` to the install action.

packages/extensionmanager/src/model.ts
```diff
@@ -343,14 +353,20 @@ export class ListModel extends VDomModel {
...
+  ): Promise<void> {
+    await this.performAction('install', entry, { useLatestVersion }).then(
```
fcollonval, line 362:
```suggestion
    await this.performAction('install', entry, options).then(
```

packages/extensionmanager/src/model.ts
```diff
@@ -480,19 +496,27 @@ export class ListModel extends VDomModel {
...
+    actionOptions: IActionOptions = {}
   ): Promise<IActionReply> {
+    const bodyJson: Record<string, string> = {
+      cmd: action,
+      extension_name: entry.name
+    };
+
+    if (action === 'install' && actionOptions.useLatestVersion) {
+      bodyJson['extension_version'] = entry.latest_version;
+    }
```
fcollonval, line 513:
```suggestion
    if (actionOptions.useVersion) {
      bodyJson['extension_version'] = actionOptions.useVersion;
    }
```

packages/extensionmanager/src/widget.tsx
```diff
@@ -498,11 +510,16 @@ class InstalledList extends ReactWidget {
...
-  onAction(action: Action, entry: IEntry): Promise<void> {
+  onAction(
+    action: Action,
+    entry: IEntry,
+    actionOptions: IActionOptions = {}
+  ): Promise<void> {
     switch (action) {
       case 'install':
-        return this.model.install(entry);
+        return this.model.install(entry, actionOptions.useLatestVersion);
```
fcollonval, line 522:
```suggestion
        return this.model.install(entry, actionOptions);
```

packages/extensionmanager/src/widget.tsx
```diff
@@ -538,11 +555,16 @@ class SearchResult extends ReactWidget {
...
```
fcollonval, line 567:
```suggestion
        return this.model.install(entry, actionOptions);
```

fcollonval (changes requested):
Thanks a lot for working on this @nbowditch-einblick 

I would like to propose a slightly different approach to pass directly the version as this is supported by the backend and may provide more flexibility.

packages/extensionmanager/src/model.ts
```diff
@@ -343,9 +356,13 @@ export class ListModel extends VDomModel {
    * Install an extension.
    *
    * @param entry An entry indicating which extension to install.
+   * @param options Additional options for the action.
    */
-  async install(entry: IEntry): Promise<void> {
-    await this.performAction('install', entry).then(data => {
+  async install(
+    entry: IEntry,
+    options: { useVersion?: string } = {}
```
nbowditch-einblick, line 363: @fcollonval - Thanks for the review! I also gave a default value to this parameter here, but let me know if you'd rather it not be optional.

fcollonval (commented):
Thanks @nbowditch-einblick 

Would you be willing to add a test for this?

You can add it in that file: https://github.com/jupyterlab/jupyterlab/blob/b835dd9acc2a00e2ee12450b9bc2b988ad789bd3/galata/test/documentation/extension_manager.test.ts

Starting by duplicating that test:

https://github.com/jupyterlab/jupyterlab/blob/b835dd9acc2a00e2ee12450b9bc2b988ad789bd3/galata/test/documentation/extension_manager.test.ts#L54

To fake an available update; you will need to set a newer version for one of the two packages in https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/documentation/data/extensions.json

The test will look like:

```ts
   const waitRequest = page.waitForRequest(request => {
     if(request.method() !== 'POST' || !galata.Routes.extensions.test(request.url()) {
       return false;
     }
     const data = request.postDataJSON();
     return data.cmd === 'install' && data.extension_name === '...' && data.extension_version === '...';
   });
   await page.getByRole('button', { name: 'Update to <the fake version you set>'}).click();
   await waitRequest

File: galata/test/documentation/extension_manager.test.ts [hunk 10]
```diff
@@ -102,6 +102,28 @@ test.describe('Extension Manager', () => {
...
+      const data = request.postDataJSON();
+      return (
+        data.cmd === 'install' &&
+        data.extension_name === '@jupyterlab/geojson-extension' &&
+        data.extension_version === '3.2.1'
+      );
+    });
+    await page.getByRole('button', { name: 'Update to 3.2.1' }).click();
+    await waitRequest;
+  });
```
fcollonval, line 126: :rocket:

krassowski (approved):
Looks good to me. I will wait with merging to allow @fcollonval to take another look.

> `test/jupyterlab/kernel.test.ts:108:9 › Kernel › Console › Should not ask kernel when creating console from launcher` failure is unrelated and is getting addressed in #15355

nbowditch-einblick:
@krassowski - thanks for taking a look! please let me know if I can do anything to update the branch, I wasn't sure if you guys prefer me doing a merge or rebase, or just taking care of it yourselves before merging

fcollonval (approved):
Thanks @nbowditch-einblick !

Let's merge

krassowski:

PR #15344, opened 2023-11-01
Add error field and animation in inline completion

### LINKED ISSUE #15343: Add request error animation in inline completion
### Problem

There are times during inline completion when network or local requests encounter errors.

Without stream enabled, such events largely go unnoticed as if nothing happened, yet users might be actively invoking the completion function.

Moreover, during a streaming session, the network could fail or the server might suddenly go offline.

Therefore, I suggest we inform the user whenever an error occurs to indicate that code completion cannot proceed due to certain issues.

### Proposed Solution

I believe adding an ```isError``` field in ```IInlineCompletionItem``` would be a great option. When ```isError == true``` it would be updated by the ```reconciliator```, and the ```ghost``` would initiate an error animation.

### Possible issues for discussion?

If the error animation is really processed in the form of a field, then in theory, the data will not be continued (stream)?

### Additional context

Maybe this is the animation I expected?

https://github.com/jupyterlab/jupyterlab/assets/90966139/462869dc-992b-41c0-93a2-822ef399ce27

Wzixiao:
## References

Fixes #15343

## Code changes

<details>

<summary>packages/completer/src/ghost.ts</summary>

```typescript
const ERROR_INDICATOR_CLASS = 'jp-GhostText-errorIndicator';

export interface IGhostText {
  /**
   * An error occurred in the request.
   */
  isError?: boolean;
}

class GhostTextWidget{
 /**
  * Delete error animation.
  */
  private _removeErrorAnimation(dom: HTMLElement): void;
  
 /**
  * Mount the error animation dom and delete stream animation ghost dom.
  */
  private _mountErrorAnimation(dom: HTMLElement) : void;
  
  private _updateDOM(dom: HTMLElement) {
    if (this.options.isError) {
      this._mountErrorAnimation(dom);

      setTimeout(() => {
        this._removeErrorAnimation(dom);
      }, 2000);
      return;
    }
    ...
  }
}
```

</details>

<details>

<summary>packages/completer/src/inline.ts</summary>

```typescript
export class InlineCompleter extends Widget {
  private _setText(item: CompletionHandler.IInlineItem) {
   ... 
    this._ghostManager.placeGhost(view, {
      from: editor.getOffsetAt(model.cursor),
      content: text,
      providerId: item.provider.identifier,
      addedPart: item.lastStreamed,
      streaming: item.streaming,
      onPointerOver: this._onPointerOverGhost.bind(this),
      onPointerLeave: this._onPointerLeaveGhost.bind(this),
      // new added
      isError: item.isError
    });
  }
}
```

</details>

<details>

<summary>packages/completer/src/reconciliator.ts</summary>

```typescript
export class ProviderReconciliator implements IProviderReconciliator {
  private async _stream(
    item: CompletionHandler.IInlineItem,
    provider: IInlineCompletionProvider
  ) {
    ...
    for await (const reply of provider.stream(token)) {
      const updated = reply.response;
      const addition = updated.insertText.substring(item.insertText.length);
      item.insertText = updated.insertText;
      item.lastStreamed = addition;
      // new added
      item.isError = reply.response.isError;
      streamed.emit(CompletionHandler.StraemEvent.update);
    }
    ...
  }
}
```

</details>

<details>

<summary>packages/completer/src/tokens.ts</summary>

```typescript
export interface IInlineCompletionItem extends IInlineCompletionItemLSP {
  token?: string;
  /**
   * Whether generation of `insertText` is still ongoing. If your provider supports streaming,
   * you can set this to true, which will result in the provider's `stream()` method being called
   * with `token` which has to be set for incomplete completions.
   */
  isIncomplete?: boolean;

  /**
   * This field is marked when an error occurs during a stream or fetch request.
   */
  isError?: boolean;
}
```

</details>

<details>

<summary>packages/completer/style/base.css</summary>

```css
.jp-GhostText-errorIndicator::after {
  animation: jp-GhostText-error 2s infinite;
  animation-delay: 400ms;
  content: ' ';
  background: rgba(255, 0, 0, 0.5);
}

@keyframes jp-GhostText-error {
  0% {
    opacity: 0.5;
  }

  100% {
    opacity: 0;
  }
}
```

</details>

## User-facing changes

https://github.com/jupyterlab/jupyterlab/assets/90966139/462869dc-992b-41c0-93a2-822ef399ce27


packages/completer/style/base.css
```diff
@@ -280,6 +280,23 @@
   }
 }
 
+.jp-GhostText-errorIndicator::after {
+  animation: jp-GhostText-error 2s infinite;
+  animation-delay: 400ms;
+  content: ' ';
```
krassowski, line 286: Just the colour change feels a bit subtle. It is possibly fine if providers will also show the details some place else. Accessibility-wise we should not relay solely on colour. An alternative to only changing colour would be adding a non-empty content, for example ⚠ though i'm not sure if this would look nice.

packages/completer/src/ghost.ts
```diff
@@ -139,7 +144,43 @@ class GhostTextWidget extends WidgetType {
...
+    dom.appendChild(streamingIndicator);
+  }
+
   private _updateDOM(dom: HTMLElement) {
+    if (this.options.isError) {
+      this._mountErrorAnimation(dom);
+
+      setTimeout(() => {
+        this._removeErrorAnimation(dom);
+      }, 2000);
```
krassowski, line 180: Should it be removed in 2s *or on the next update* whichever is sooner?

packages/completer/src/ghost.ts
```diff
@@ -139,7 +144,43 @@ class GhostTextWidget extends WidgetType {
...
+    elementsToRemove.forEach(element => {
+      element.remove();
+    });
+  }
+
+  /**
+   * Mount the error animation dom and delete all displayed ghost text dom.
+   */
+  private _mountErrorAnimation(dom: HTMLElement) {
+    const streamingIndicator = document.createElement('span');
```
krassowski, line 159: Can you renamed this to `errorIndicator` for clarity please?

packages/completer/src/tokens.ts
```diff
@@ -212,6 +212,11 @@ export interface IInlineCompletionItem extends IInlineCompletionItemLSP {
    * with `token` which has to be set for incomplete completions.
    */
   isIncomplete?: boolean;
+
+  /**
+   * This field is marked when an error occurs during a stream or fetch request.
+   */
+  isError?: boolean;
```
krassowski, line 219: To consider:
- should it be a boolean or maybe an object so that we can add "message" and "type" in the future?
- should it be a field at all, or should we rely on throwing errors in stream/fetch methods and catching them in reconciliator to populate ghost text instead?

packages/completer/src/reconciliator.ts
```diff
@@ -127,6 +127,7 @@ export class ProviderReconciliator implements IProviderReconciliator {
       // Stream an update.
       item.insertText = updated.insertText;
       item.lastStreamed = addition;
+      item.isError = reply.response.isError;
```
krassowski, line 130: Does it also need to be set in `fetchInline`?

krassowski (commented):
Thank you for starting this @Wzixiao. Please see a few suggestions/questions below.

krassowski:
@Wzixiao are you interested in finishing this PR? Our merge window for 4.1 will close soon.

packages/completer/src/tokens.ts
krassowski: In LSP spec error is two levels higher, in the response message which contains the `result` (and optionally `error` which is a `ResponseError` object). https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#responseMessage

But the error should be defined per item. Maybe it is fine to have `error?: CompletionError` field here.

krassowski:
How to test:

1. Checkout this PR
2. Modify the history provider by adding `, error: {message: 'Completion error!'}` in line 102:
   https://github.com/jupyterlab/jupyterlab/blob/a5f9207a28244b5d407281244ff310700be402e1/packages/completer/src/default/inlinehistoryprovider.ts#L101-L103
3. Rebuild, start JupyterLab in dev mode
4. Enable history provider
5. Type something in a cell which is a prefix of previously run command in your history

completion-error-indicator

krassowski (approved):
This is good to go from my side - thank you @Wzixiao for starting this! I will wait a while before merging in case if anyone else wants to review the latest version (but preferably further enhancements would be tracked in separate issues and addressed in a follow-up PRs).

fcollonval (approved):
thanks @Wzixiao and @krassowski

Issue #15303, opened 2023-10-25
Accessibility - Keyboard Navigation using a Screen Reader


m158261:
These issues have been discovered while trying to navigate in Jupyterlab using keyboard navigation with a screen reader active. The screen reader used is Microsoft Narrator. 

### Problem

**Issue no. 1 - New Launcher Icon label not descriptive**

 - When navigating to the new launcher button the screen reader announces "graphic". This should be solved using aria-labels. 

**Issue no. 2 - Unhelpful description when opening new launcher**

 - When opening a new launcher the screen reader announces "launcher close launcher". This is not descriptive of the action being done. 

**Issue no. 3 - Arrow key navigation between launcher tabs not as expected**

 - When screen reader is active, pressing right arrow keys steps through the launcher title (each individual letter) rather than moving to the next launcher tab. 

**Issue no. 4 - Cannot expand toggle arrows in Terminal/Debugger/Extension Manager sidebar widgets**

 - When navigating in sidebar widgets the toggle arrows are not expandable using keyboard keys.

**Issue no. 5 - Unable to open menu items using enter/arrow keys**

 - Menu items in the top menu bar are not expandable using enter/space/arrow keys. 

**Issue no. 6 - Unable to select all cells using Ctrl + A in notebook**

 - When navigating cells in a notebook, Ctrl + A does not select all cells. It selects all elements in the document. 

**Issue no. 7 - Move cell up/down in notebook not working with keyboard commands**

 - This functionality is flaky when using a screen reader. It is difficult to focus on the cell to move it up/down. 

**Issue no. 8 - Navigation to cell buttons difficult**

 - It is difficult to access the buttons in the right of a cell using keyboard navigation. It is not intuitive, it can be currently accessed via arrow keys but you have to exit the cell, tab to switch kernel then use arrow keys. 

**Issue no. 9 - Pressing A in notebook does not insert a new cell above**

 - Pressing A does not insert a new cell when using the screen reader. The screen reader announces "unavailable".

**Issue no. 10 - Pressing B in notebook does not insert a new cell below**

 - Pressing B does not insert a new cell when using the screen reader. The screen reader announces "unavailable".

**Issue no. 11 - Pressing C/V does not copy/paste a cell in a notebook**

 - Pressing C/V does not copy and paste a cell when using the screen reader. The screen reader announces "unavailable" for V and "no next" for C.

**Issue no. 12 - Pressing D+D in notebook does not delete a cell**

[...]

PR #15347, opened 2023-11-02
Change close launcher div to button for screen reader accessibility 

e218736:
## References

#15303 - Issue no. 3 - Arrow key navigation between launcher tabs not as expected

depends on [Lumino #655](https://github.com/jupyterlab/lumino/pull/655)

## Code changes

Changed close Launcher div element to a button element so that tab will close on enter

## User-facing changes

User should be able to navigate through tabs and close tab icons using  right arrow key without going through each individual letter and icon 

- With a screen reader active, navigate to first launcher tab using the Tab key
- Navigate to open a new launcher with caps lock + right arrow key and press enter (screen reader should not read out individual letters)
- Navigate to open a new launcher with caps lock + right arrow key (screen reader should read 'close launcher
button' rather than 'graphic')
- Press enter and the selected launcher tab should close

None


packages/apputils/src/mainareawidget.ts
```diff
@@ -48,6 +48,7 @@ export class MainAreaWidget<T extends Widget = Widget>
     const toolbar = (this._toolbar = options.toolbar || new ReactiveToolbar());
     toolbar.node.setAttribute('role', 'toolbar');
     toolbar.node.setAttribute('aria-label', trans.__('notebook actions'));
+    toolbar.node.setAttribute('aria-hidden', 'true');
```
gabalafou, line 51: I think I understand the desire to hide this "toolbar" node because when the content node is the launcher or a normal file or any other number of non-notebook content, then this toolbar node is most often empty, and it looks like this:

literally a 740px wide by 3px tall box with a bottom border

However, it becomes super important when a notebook is loaded. Then it becomes the node into which the notebook toolbar is loaded, and it looks like this:

a typical application toolbar with icon buttons for commands like save, add new cell, cut, copy, paste, etc.

Therefore, we can't put aria-hidden on this node toolbar because then it gets hidden from screen readers when  you load a notebook, which means you can't navigate to it using caps lock plus arrow keys, which breaks the expected UX.

packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -61,7 +61,13 @@ export namespace TabBarSvg {
 
       return hpass(
         'div',
-        { className, title },
+        {
+          className,
+          title,
+          'aria-label': 'Close Launcher',
+          'aria-hidden': 'false',
```
gabalafou, line 68: Do we need to explicitly set `aria-hidden` to `false`?

packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -61,7 +61,13 @@ export namespace TabBarSvg {
...
+          role: 'button'
```
gabalafou, line 69: I wonder if we should convert this div to a button rather than add the button role.

gabalafou (commented):
It's great to see issue #15303 being addressed!

I left some review comments in the PR. Could you also add a manual test plan to the PR description?

packages/apputils/src/mainareawidget.ts
e218736: Yes, makes sense I will amend. Is using caps lock + arrow keys standard to navigate between elements for all assistive screen readers? 
I wasn't aware of this functionality and if so would mean I can remove the some of the changes you have commented on in another PR (https://github.com/jupyterlab/lumino/pull/655) since it will work fine with caps lock + arrow keys

fcollonval (changes requested):
Thanks @e218736 

This breaks the styling:

image


You must add the following CSS rules in packages/ui-components/style/tabbar.css

```css
.lm-TabBar-tabCloseIcon {
  border: none;
  background: none;
}
```

e218736:
@fcollonval 

CSS changes have been made in lumino [#655](https://github.com/jupyterlab/lumino/pull/655)

You can check out the linked changes by launching this binder [here](https://mybinder.org/v2/gist/e218736/0f1378df2d41d700f15b88041f952f54/HEAD?urlpath=lab)

But let me know if the styling would be best in the file mentioned, thanks

krassowski:
https://github.com/deathbeds/jupyak/pull/16

gabalafou:
That [Jupyak build](https://jupyak--16.org.readthedocs.build/en/16/_static/work/lite/lab/) works great for testing this PR together with Lumino PR #655, thanks for that!

I have two concerns with this PR:

1. Merging is delicate because of the coordination with the Lumino PR, but it doesn't **have** to be tied to the Lumino PR. We could duplicate the CSS here so that both PRs can be merged independently.
2. Changing the close icon from a div to a button makes it a tab stop. Normally this would be a good thing if the button were wired up to handle <kbd>Enter</kbd> and <kbd>Space</kbd> keys. But this button is not unfortunately wired up to handle Enter or Space. I don't think it should be a tab stop until the standard button keyboard shortcuts work. 

I will push commits to this PR that address both of the above issues.

gabalafou:
So, after giving this some thought, I think the best way to move forward with this PR is to just add role=button to the div that serves as the close tab icon. This avoids messing up the styles, and gives the icon the proper semantics that it needs for screen readers and other assistive tech to use the icon as a button, while not making the icon a tab stop, which would be bad because then keyboard users would be able to press the tab key to land on the button but then not be able to actually do anything, because as I mentioned the Enter and Space keys are not wired up.

Side note: while testing this I realized that we have a problem in JupyterLab where the DOM nodes are not added to the tree in the logical order. At first I thought something was broken because I couldn't get to the tab bar by pressing left and up arrows from the notebook node, but then I realized that's because the app put the tab bar node in the DOM after the notebook node, even though it displays it visually above the notebook node.

gabalafou:
@e218736 when you get a chance could you weigh in on these changes and my comments? If it looks good to you, I will approve this PR and merge it in (after checking in on the failing UI test)

packages/ui-components/src/icon/widgets/tabbarsvg.ts
```diff
@@ -61,7 +61,13 @@ export namespace TabBarSvg {
 
       return hpass(
         'div',
-        { className, title },
+        {
+          className,
+          title,
+          // TODO once Lumino PR #655 is incorporated into JupyterLab, make this
+          // element a button rather and a div with role=button
```
gabalafou, line 68:
```suggestion
          // element a button rather than a div[role=button]
```

e218736:
> @e218736 when you get a chance could you weigh in on these changes and my comments? If it looks good to you, I will approve this PR and merge it in (after checking in on the failing UI test)

Hi @gabalafou Have checked changes on my side, all looks good to me. Thank you.

gabalafou:
Since this depends on a Lumino PR (https://github.com/jupyterlab/lumino/pull/655) which has not been merged in, I'm bumping this to a future milestone.

Issue #15330, opened 2023-10-30
Windowed notebook does not update window on panel resize


krassowski:
## Description

With `full` windowed notebook, the window is not updated on panel resize.

## Reproduce

1. Follow steps on the screencast
   ![windowing-does-not-update](https://github.com/jupyterlab/jupyterlab/assets/5832902/ae2769cf-0679-4e5b-9402-35c8477f5795)
2. See that cells revealed by resize are shown in `defer` mode
3. See that cells revealed by resize are not show in `full` model

## Expected behavior

Cells revealed on resize are shown when we resize the notebook

## Context

- Browser and version: Chrome
- JupyterLab version: 4.0.7

PR #15355, opened 2023-11-05
Faster galata/UI tests

krassowski:
## References

When developing tests against https://github.com/jupyterlab/jupyterlab/issues/15330 and other issues I faced a problem where this code path would freeze the test in an infinite loop.

From an experiment on my fork it appears that this little change could cut the runtime for galata tests by 50%. From about 44 minutes down to about 22 minutes.

## Code changes

Use playwright `getAttribute` instead of evaluation to get class name.

## User-facing changes

None

None

### DIFF
--- a/galata/src/helpers/activity.ts
+++ b/galata/src/helpers/activity.ts
@@ -51,13 +51,11 @@ export class ActivityHelper {
       return activeTab === name;
     } else {
       const tab = await this.getTab(name);
-      return (
-        (tab &&
-          (await tab.evaluate((tab: Element) =>
-            tab.classList.contains('jp-mod-current')
-          ))) ??
-        false
-      );
+      if (!tab) {
+        return false;
+      }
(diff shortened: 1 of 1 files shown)


fcollonval:
Some comments about the execution time huge various you have encountered. I realized that lately by times to times the integration tests where execution with 2 workers instead of 1. This is nice but introduce discrepancies in documentation as in that case we end up with 2 `test-documentation-*` folders instead of 1. I pushed some improvement in #15375 :crossed_fingers:

krassowski:
Thanks! Should we merge these branches? It seems that we picked on different things I guess it would be good to see how it works together (and whether it does result in fully green tests on CI). Feel free to merge this branch (or specific commits) into your PR and then close this one.

fcollonval (approved):
Thanks @krassowski

fcollonval:

PR #15356, opened 2023-11-05
Fix scrolling past long outputs in presence of un-rendered headings

### LINKED ISSUE #15352: Notebook window jumps when scrolling past long outputs
### Reproduce

1. Use Firefox or Chrome latest or stable (I reproduced across multiple versions)
2. Go to JupyterLab settings, change `Windowing mode` to `full`
3. Create a notebook with ~200 empty cells
4. Convert a cell into markdown and type  `# test` creating a heading
5. In a code cell below enter `print('\n'.join(str(i) for i in range(100)))` and execute
6. Perform the motion as on the screencast:
    no-cel-active-all-fine
7. See that all is good
8. Enter the markdown cell with a heading
9. Try to repeat the motion as in point (6)
10. See annoying, disorienting jump which makes me disable the `full` windowing mode:
   in the first attempt it took me up to the selected cell and then back down; I experienced a very distressing circumstance when I was stuck and unable to navigate past a cell by scrolling up at all but it is not as reliably reproducible
   jump-back
   in take two it just jumped over the output of the cell preventing me from reading the lines in between
   jump-back-take-2

I though that ToC is to blame because it does trigger this behaviour because when you click on a cell heading it enters the cell into markdown mode (which is super annoying too and I believe someone else reported it but cannot find the issue).

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/14106#issuecomment-1749112997_

### Context

JupyterLab 4.0.8

krassowski:
## References

Fixes #15352

## Code changes

- do not scroll to newly attached un-rendered markdown cell with a heading
- exposes `options` in `setCursorPosition` API (already present - but not previously functional - in `CodeMirrorEditor` implementation)

## User-facing changes

Scrolling past long cell outputs in full windowing mode works ok even if there are unrendered markdown cells.

None

### DIFF
--- a/galata/src/helpers/notebook.ts
+++ b/galata/src/helpers/notebook.ts
@@ -787,7 +787,10 @@ export class NotebookHelper {
   }
 
   /**
-   * Whether the cell is in editing mode or not
+   * Whether the cell is in editing mode or not.
+   *
+   * This method is not suitable for checking if a cell is unrendered
+   * as it will return false when the cell is not active (not focused).
    *
    * @param cellIndex Cell index
    * @returns Editing mode
--- a/galata/test/jupyterlab/notebook-scroll.test.ts
+++ b/galata/test/jupyterlab/notebook-scroll.test.ts
@@ -65,6 +65,10 @@ test.describe('Notebook scroll over long outputs', () => {
     await page.notebook.activate(outputAndHeading);
   });
 
+  test.afterEach(async ({ page, tmpPath }) => {
+    await page.contents.deleteDirectory(tmpPath);
+  });
+
   test('should scroll smoothly without snapping to headings', async ({
     page
   }) => {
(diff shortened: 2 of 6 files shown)


krassowski:
The failing `Kernel › Console › Should not ask kernel when creating console from launcher` test is getting fixed in #15355.

fcollonval (approved):
Thanks @krassowski 

I think we could target 4.0.x although there is an API change as it is backward compatible. What do you think?

krassowski:
> I think we could target 4.0.x although there is an API change as it is backward compatible. What do you think?

I agree. It seems unlikely this change could impact downstreams and if it does I can revert it, but I think the benefit of having this fixed outweights the minimal risk due to extension of the interface.

krassowski:

PR #15357, opened 2023-11-05
Update notebook window on resize if height changes

### LINKED ISSUE #15330: Windowed notebook does not update window on panel resize

krassowski:
## References

Fixes #15330

## Code changes

- invokes `WindowedList.update()` if the height changed during resize
- adds a test case

## User-facing changes

Resizing notebooks in `full` windowed mode will show all the cells that should be visible.

None

### DIFF
--- a/galata/test/jupyterlab/notebooks/20_empty_cells.ipynb
+++ b/galata/test/jupyterlab/notebooks/20_empty_cells.ipynb
@@ -1,0 +1,185 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "27ad8e86-a071-4008-9473-21336849d03d",
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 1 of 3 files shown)


afshin (approved):
Thanks for this change! I think you can either do the rate limiting here or I can resolve it when I rebase in the virtual scrollbar UI PR. Cheers, @krassowski!

krassowski:
Rate-limiting wise I went for a throttler rather than a debouncer (after testing both) because debouncer would not reveal the cells as user resizes the panels (and user may want to just to "peek" one cell and resize back which should be a single smooth action):

| Debouncer | Throttler |
|--------|--------|
| debouncer| throttler |

krassowski:

PR #15371, opened 2023-11-08
Set fallback for material css properties

fcollonval:
## References

The use of `--md-` CSS properties make some style sheets (in particular the theme ones) useless when loaded directly from CDN.

## Code changes

Add hard coded values fallback for all material color CSS properties

## User-facing changes

None

None


krassowski:
Mixed feelings. Would it help if instead we vendored a subset of `--md` rules in a separate file (that would be available on CDN but not necessarily loaded in lab)?

fcollonval:
> Would it help if instead we vendored a subset of --md rules in a separate file (that would be available on CDN but not necessarily loaded in lab)?

What would be the advantage of that approach? It will be even worst as it will add yet another place were `--md-` CSS properties may be defined from.

Fore reference the MD colors are hard coded (and vendored in `@jupyterlab/apputils`): https://github.com/jupyterlab/jupyterlab/blob/main/packages/apputils/style/materialcolors.css.

If a user wants to override them to change the JupyterLab core theme, the changes will still be applied - this only adds a fallback in case there are not defined anywhere making `variables.css` self contained (what I considered a good practice).

packages/debugger/style/editor.css
Lokedout89, line 1: # GitHub CLI api
# https://cli.github.com/manual/gh_api

gh api \
  --method POST \
    -H "Accept: application/vnd.github+json" \
      -H "X-GitHub-Api-Version: 2022-11-28" \
        /repos/OWNER/REPO/issues \
          -f title='Found a bug' \
           -f body='I'm having a problem with this.' \
            -f "assignees[]=octocat" \
             -F milestone=1 \
              -f "labels[]=bug"

krassowski (commented):
Performance-wise it is fine in FF and Chrome based on micro-benchmarks :heavy_check_mark:

It looks like some grey colours could have been mis-pasted?

packages/theme-dark-extension/style/variables.css
```diff
@@ -166,9 +166,9 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-content-font-color1: rgba(255, 255, 255, 1);
   --jp-content-font-color2: rgba(255, 255, 255, 0.7);
   --jp-content-font-color3: rgba(255, 255, 255, 0.5);
-  --jp-content-link-color: var(--md-blue-300);
-  --jp-content-link-visited-color: var(--md-purple-300);
-  --jp-content-link-hover-color: var(--md-blue-400);
+  --jp-content-link-color: var(--md-blue-300, #64b5f6);
+  --jp-content-link-visited-color: var(--md-purple-300, #ba68c8);
+  --jp-content-link-hover-color: var(--md-blue-400, #42a5f5)
```
fcollonval, line 171:
```suggestion
  --jp-content-link-hover-color: var(--md-blue-400, #42a5f5);
```

krassowski (approved):

Issue #15372, opened 2023-11-08
Docs screenshots tests often fail (due to concurrency?)


krassowski:
This started to happen more frequently recently. It appears to happen when two tests run concurrently (which depends on the runner assigned my GitHub) in which case both runs tests may create the temporary directory for themselves:

Actual

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/27326048-1d1f-486f-b7ce-5b0fa723a8ea)

Expected

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/3daaae17-8437-4330-90b6-1b678deb137a)

Actual

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/b5fe94e2-ca56-4590-9e71-22cea62183eb)

Expected

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/471453aa-a6a4-4144-a539-36e2bc2329e4)

```
Notice:   4 failed
    [documentation] › test/documentation/customization.test.ts:190:7 › Customized › should use customized layout 
    [documentation] › test/documentation/general.test.ts:231:7 › General › Open tabs ───────────────
    [documentation] › test/documentation/internationalization.test.ts:14:7 › Internationalization › Menu 
    [documentation] › test/documentation/internationalization.test.ts:43:7 › Internationalization › UI in Chinese 
  4 flaky
    [documentation] › test/documentation/customization.test.ts:14:7 › Default › should use default layout 
    [documentation] › test/documentation/debugger.test.ts:39:7 › Debugger › Activate ───────────────
    [documentation] › test/documentation/debugger.test.ts:124:7 › Debugger › Breakpoints on exception 
    [documentation] › test/documentation/general.test.ts:19:7 › General › Welcome ──────────────────
```

We should rewrite these snapshot tests to:
- [ ] clone demo repo into the test sub-directory or filter out temporary test directories
- [ ] maybe not use tree command in terminal but instead something with fixed output

PR #15375, opened 2023-11-09
Some galata improvements

fcollonval:
## References

Improve galata

Fixes https://github.com/jupyterlab/jupyterlab/issues/15372

## Code changes

- Extract better debugger helpers from https://github.com/jupyterlab/jupyterlab/pull/15156
- Better and faster `setCell`
- Hide the `test-*` folders in filebrowser for documentation snapshots. We start to see discrepancy because some times the test are executed with 2 workers instead of 1.
- Add new API to get locator instead of elementHandle

## User-facing changes

None

None


fcollonval:
Mike if you come by, I'm looking at this one right now.

krassowski:
Thanks for the heads up!

brichet:
@fcollonval what is the status of this PR ? 
Do you want to provide a copy of all the helpers returning handle, but returning a locator ? It would be great if this were the case, I can probably help on it if necessary.

fcollonval:
> @fcollonval what is the status of this PR ? 
> Do you want to provide a copy of all the helpers returning handle, but returning a locator ? It would be great if this were the case, I can probably help on it if necessary.

Hey @brichet yes the main idea is to move to locators. Unfortunately lots of tests are still failing. If you have some cycles to look at the remaining failures your help will be much appreciated.

brichet:
@fcollonval, I partially restored the previous logic to set cell mode.
Indeed double clicking set the edit mode for any cell type, but it can also select some text.
I think some tests were failing because of it.
Let me know if you think it is important to change this logic.

brichet:

brichet:

brichet:
@fcollonval I think this PR is ready now. The remaining failures should not be related to this PR.
It removes all the use of element handles in the and `waitForSelector` (also deprecated) in the tests.
It adds new API in helpers to use locators instead of element handles, but keep the existing API.

We should add a deprecated comment on the previous API, to be able to remove it on the next major release of Galata.

krassowski:
Sorry about the conflicts! I see that this PR adds many `*Locator` methods; there are two kinds of these:
- (a) `getCell(index: number): ElementHandle` now gets `getCellLocator(index: number): Locator`
- (b) `waitForTransition(element: ElementHandle<Element> | string): Promise<void>` now also gets `waitForLocatorTransition(element: Locator | string): Promise<void>` sibling

Both make tests a bit harder to read do to increased verbosity.  My thinking is that (a) is fine but (b) is not necessary. Instead we could have:

```js
waitForTransition(element: ElementHandle<Element> | Locator | string): Promise<void>
```

and deprecate passing `ElementHandle<Element>`. This would also reduce the delta of this PR. Thoughts?

brichet:
Thanks for the comment @krassowski 

> My thinking is that (a) is fine but (b) is not necessary.

I agree, but I didn't find a way to share some code in the functions, while avoiding the use of `ElementHandle` if a `Locator` is provided. I'll try again, otherwise I'll duplicate the code.


> This would also reduce the delta of this PR.

Yes, I think 4 functions are impacted.

brichet:
> Both make tests a bit harder to read do to increased verbosity. My thinking is that (a) is fine but (b) is not necessary. Instead we could have:
> 
> ```js
> waitForTransition(element: ElementHandle<Element> | Locator | string): Promise<void>
> ```

I updated it in https://github.com/jupyterlab/jupyterlab/pull/15375/files#diff-280c4ca697fcec0a3945f7919e10481049453ecbe9637fbfc3ce328c37f52dceR211, using `try catch` because I didn't found a way to discriminate `ElementHandle` from `Locator`, these are not classes but interfaces.
Let me know if you think of another implementation.

brichet:
@fcollonval I rebased and opened the PR.

Let me know if you wanted to add some features in it.

krassowski:
I think we should rebase and merge this one soon for it to avoid gathering more conflicts

brichet:
> I think we should rebase and merge this one soon for it to avoid gathering more conflicts

I agree, I'll rebase it soon

brichet:
@krassowski @fcollonval I rebased this PR, if you want to take a look

galata/src/helpers/notebook.ts
```diff
@@ -599,25 +637,40 @@ export class NotebookHelper {
...
 
-    const isRenderedMarkdown = await cellEditor.evaluate(editor =>
-      editor.classList.contains('lm-mod-hidden')
-    );
+    const isRenderedMarkdown = (
+      await Utils.getLocatorClassList(cellEditor)
+    ).includes('lm-mod-hidden');
+    // const isRenderedMarkdown = await cellEditor.evaluate(editor =>
+    //   editor.classList.contains('lm-mod-hidden')
+    // );
```
krassowski, line 671: What is the story here?
```suggestion
```

krassowski (approved):
Thank you! From my point of view good to merge after cleaning up the commented code.

galata/src/helpers/notebook.ts
brichet: Good catch, thanks

brichet:
Thanks @krassowski for review.

Failed test is not related.

PR #15380, opened 2023-11-10
Don't show default value for objects in Settings Editor

### LINKED ISSUE #15323: CodeMirror settings – suppress "Default: [object Object]" below section when settings are changed
### Problem

When any of the settings within a CodeMirror settings section in the Settings Editor (i.e., in the _CodeMirror_, _Notebook_, _Text Editor_ and _Code Console_ pages), the user's settings object no longer matches the default settings object, and "Default: [object Object]" appears below the CodeMirror settings section. However, the message does not seem to be meaningful or helpful for the user, so it just adds visual noise to the settings pages, which are already quite dense.

image

### Proposed Solution

Don't display the default when the setting schema type is `object`.

firai:
## References

Fixes #15323 

## Code changes

Don't show default for `objects` in form UI component, the result of which will always be "Object". I'm not entirely sure where else this component is used, but I don't think the "Default: [object Object]" message would be useful in any case.

## User-facing changes

After changing a setting that is part of an object, the message "Default: [object Object]" no longer appears below the settings section for the object.

None as far as I know.

### DIFF
--- a/packages/ui-components/src/components/form.tsx
+++ b/packages/ui-components/src/components/form.tsx
@@ -575,14 +575,18 @@ const CustomTemplateFactory = (options: FormComponent.ILabCustomizerProps) =>
                 {schema.description}
               </div>
             )}
-            {isModified && defaultValue !== undefined && schema.type !== 'object' && (
-              <div className="jp-FormGroup-default">
-                {trans.__(
-                  'Default: %1',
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you @firai!

krassowski:

PR #15177, opened 2023-09-29
Update to `@jupyter/ydoc` 1.1.1

jtpio:
## References

Update to `@jupyter/ydoc` 1.1.1.

To grab the fix from https://github.com/jupyter-server/jupyter_ydoc/pull/186 that landed in `1.1.1`: https://github.com/jupyter-server/jupyter_ydoc/releases/tag/v1.1.1

For reference the changelog for `1.1.0`: https://github.com/jupyter-server/jupyter_ydoc/releases/tag/v1.1.0

## Code changes

- [x] Dependency update.
- [x] Fix incorrect test expectation

## User-facing changes

Should be none for regular JupyterLab / Jupyter Server users.

## Backwards-incompatible changes

None

PR #15383, opened 2023-11-10
Update @jupyter/ydoc in dev_mode

fcollonval:
## References

Follow-up of #15177

## Code changes

Update pinned version in dev_mode

## User-facing changes

None

None


fcollonval:
> This does not need to be backported as it is part of #15382 (which allow me to see this oversight).

PR #15387, opened 2023-11-10
Do not update filebrowser breadcrumbs when not needed

### LINKED ISSUE #15365: Jitter in the file browser when refreshed
## Description
I am noticing a jitter in the file browser on refreshing it. Please see the attached gif.
jitter2

## Reproduce
Click on the refresh button in the filebrowser.
This is more easily reproducible when there is a large number of items in a directory.

## Context
I'm using  -  jupyterlab==4.0.6

krassowski:
This slightly improves performance by eliminating spurious DOM updates, and prevents jitter on idle update when the breadcrumbs height is modified in extensions.

## References

Fixes #15365 (the JupyterLab side of the issue)

## Code changes

Split state out the private arguments, check if state changed before proceeding with the update.

## User-facing changes

Users of third party extensions will not see jitter when the breadcrumbs widget updates without change in the content.

None

### DIFF
--- a/packages/filebrowser/test/crumbs.spec.ts
+++ b/packages/filebrowser/test/crumbs.spec.ts
@@ -223,6 +223,59 @@ describe('filebrowser/model', () => {
         expect(items.length).toBe(3);
         model.dispose();
       });
+
+      it('should trigger DOM updates if state changes', async () => {
+        const model = new FileBrowserModel({ manager });
+        crumbs = new LogCrumbs({ model });
+        let modifications = 0;
+        const observer = new MutationObserver(() => modifications++);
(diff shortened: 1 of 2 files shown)


packages/filebrowser/test/crumbs.spec.ts
```diff
@@ -223,6 +223,59 @@ describe('filebrowser/model', () => {
         expect(items.length).toBe(3);
         model.dispose();
       });
+
```
fcollonval, line 226: :heart_eyes:

fcollonval (approved):
Thanks @krassowski

fcollonval:

PR #15393, opened 2023-11-12
Fix completer cycling documentation formatting

krassowski:
## References

Before

image

After

image


fcollonval (approved):
Thanks @krassowski

PR #15405, opened 2023-11-14
Maintenance bump lumino

fcollonval:
## User-facing changes

Keyboard event are capture are the bubbling phase not the capturing phase

None


krassowski (approved):
Thanks! Re-running the visual tests before merging to make sure the failures are just flaky tests rather than regressions from new lumino versions.

Issue #15403, opened 2023-11-14
Visited link are invisible in light theme


fcollonval:
## Description

The color `--jp-content-link-visited-color` used for the light theme makes it invisible

![image](https://github.com/jupyterlab/jupyterlab/assets/8435071/8457838c-a650-407b-91e4-0fd5fa242363)

## Reproduce
1. Go to the extension manager
2. Click on a link to open a extension home page
3. Come back to jupyterlab

## Expected behavior

The text is visible also in light theme

## Context

Main branch
Tested on Firefox and Chrome

That color was introduced by https://github.com/jupyterlab/jupyterlab/pull/15181

---

For reference this is how it looks with the dark theme:

![image](https://github.com/jupyterlab/jupyterlab/assets/8435071/3e82dc5c-5c78-4115-9317-d7cc4cbdbf3b)

fcollonval:
@m158261 by any chance do you have time to solve this issue?

m158261:
Hey @fcollonval. Yeah, I'll try and get that fixed today.

m158261:
Hi @fcollonval. I've raised a PR. https://github.com/jupyterlab/jupyterlab/pull/15406

PR #15406, opened 2023-11-14
Updated light theme visited link colour to make text visible

m158261:
## References

This fixes https://github.com/jupyterlab/jupyterlab/issues/15403

## Code changes

The visited link CSS variable in light theme has been changed to make the text visible. 

## User-facing changes

Link after and before visit - Light Theme
new visited link

Link after and before visit - Dark Theme
new visited link dark v2

None


krassowski:
Thank you @m158261 can you add screenshots in dark mode as well?

krassowski:
So this is what I see in dark/light theme. I am not convinced weather the light mode contrast is enough and whether the dark mode contrast is not too much. Top not visited, bottom visited:

| Light | Dark |
|--|--|
| Screenshot from 2023-11-18 14-22-03 | Screenshot from 2023-11-18 14-21-59 |

m158261:
Hi @krassowski, I have changed the visited link colour in light and dark themes for better contrast. I have used colours similar to the Google Chrome light/dark themes. Please let me know what you think. 

| Light | Dark |
|--------|----------|
| new visited link | new visited link dark |

| Google Light | Google Dark |
|--------|----------|
| google visited links light |  google visited links |

packages/theme-dark-extension/style/variables.css
```diff
@@ -167,7 +167,7 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-content-font-color2: rgba(255, 255, 255, 0.7);
   --jp-content-font-color3: rgba(255, 255, 255, 0.5);
   --jp-content-link-color: var(--md-blue-300);
-  --jp-content-link-visited-color: rgba(255, 255, 255, 0.3);
+  --jp-content-link-visited-color: var(--md-purple-400);
```
krassowski, line 170: This fails WCAG AA. What about:

```suggestion
  --jp-content-link-visited-color: var(--md-purple-300);
```

krassowski (commented):
That's definitely the right direction, but I think the dark theme would have too little contrast (3.34:1) so I propose to change it (upping it to 4.52:1).

m158261:
> That's definitely the right direction, but I think the dark theme would have too little contrast (3.34:1) so I propose to change it (upping it to 4.52:1).

Looks good to me. I've updated the screenshot in the description.

m158261:
I've had a play about with different colours, we could use #d497de which passes AA and AAA with a contrast ratio of 7.11. What do you think? 

d497de

krassowski (approved):
LGTM, thanks!

PR #15413, opened 2023-11-15
Scroll to the active cell when typing (in edit mode)

brichet:
## References

Follow up https://github.com/jupyterlab/jupyterlab/pull/14115

https://github.com/jupyterlab/jupyterlab/pull/14115 removed the scroll to the active cell on keydown event. 
When the notebook is in edit mode and the active cell is not visible, using the keyboard should scroll to the active cell.

This PR restores this behavior, and removes some complexity included in the PR above.

## Code changes

- scroll to the active cell on keydown event on the notebook in edit mode
- `selectAbove()` and `selectBelow()` notebook actions are not asynchronous anymore, which simplifies some tests.

> This results in reverting asynchronous methods introduced in #15413

## User-facing changes

Scroll to the active cell when typing, as previously.

None

cc @gabalafou @krassowski who worked on the related PR

### DIFF
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -896,34 +896,25 @@ export namespace NotebookActions {
       return;
     }
 
-    // Scroll first to the active widget in case it is not attached,
-    // in windowed notebook.
-    notebook
-      .scrollToItem(notebook.activeCellIndex)
-      .then(() => {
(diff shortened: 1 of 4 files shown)


brichet:
Test failures are not related

fcollonval (approved):
Thanks @brichet

fcollonval:
Merging as the test will pass (the last commit is only adding a code comment)

PR #15418, opened 2023-11-16
Link JupyterLab and Lumino PRs in same Binder instance

m158261:
## References

https://discourse.jupyter.org/t/linking-jupyterlab-and-lumino-in-binder/22333/2
https://github.com/jupyter/accessibility/pull/20

## Code changes

Added code to link JupyterLab and Lumino PRs to make testing easier for issues that are spread across different repositories. 

## User-facing changes

Community members that are testing PRs will be able to configure binder to use multiple repositories. 

None


krassowski:
Currently it is not clear how a developer would make use the scripts added in this PR or where these scripts would run/deposit artefacts/create a runable binder.

Among a number of alternative solutions discussed during accessibility call one was dynamically creating a gist by using an action that runs on comments. This would work like follows:
1. User comments `bot please link with krassowski/lumino fix/test branch`
2. A gist from this template: https://gist.github.com/krassowski/7d3172a1c343045cdf1bd2a662688bd2 is generated where `git clone` commands are substituted with appropriate URLs
3. The gist is posted to the bot account using gist API: https://docs.github.com/en/free-pro-team@latest/rest/gists/gists?apiVersion=2022-11-28#create-a-gist
4. The comment is posted on this PR with a link to the binder to be spawned from said gist

You can test it out with https://mybinder.org/v2/gist/krassowski/7d3172a1c343045cdf1bd2a662688bd2/HEAD?urlpath=lab which links latest lumino `main` branch with latest `jupyterlab` branch. It took ~20 minutes to build. The gist probably can be simplified by removing some non-essential dependencies but it is unlikely to speed up the build as most of the time is in yarn fetching node modules and in ts compilation.

jtpio:
As an alternative to this PR @bollwyvl recently shared this new project: https://github.com/deathbeds/jupyak

PR #15419, opened 2023-11-16
Restart ci through bot after updating the snapshots

fcollonval:
## References

Use the new ability of jupyterlab-bot to trigger CI following snapshot update

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/galata-update.yml
+++ b/.github/workflows/galata-update.yml
@@ -14,7 +14,6 @@ jobs:
     if: ${{ github.event.issue.pull_request && (contains(github.event.comment.body, 'please update galata snapshots') || contains(github.event.comment.body, 'please update snapshots')) }}
     timeout-minutes: 80
     runs-on: ubuntu-22.04
-
     steps:
       - name: React to the triggering comment
         run: |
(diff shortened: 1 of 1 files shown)


krassowski:
Could we instead configure a `PAT` for jupyterlab-bot or similar and use it instead of `GITHUB_TOKEN` in the update action?
This should work as per documentation and multiple other sources, e.g. https://stackoverflow.com/questions/64076798/how-to-trigger-a-github-actions-workflow-on-push-to-another-branch
Should it possibly be a separate bot without release rights? CC @blink1073

blink1073:
It doesn't have to be a bot, you could store a PAT on the repo with the correct perms, and add a guard for who made the comment, making sure they have the right permissions.

jtpio:
@krassowski is that something you think should be fixed as part of this PR? Or could this PR already be merged as an incremental improvement? And we look into using a `PAT` in the next iteration?

Thinking it could already be useful to have the CI restarted automatically when updating snapshots.

krassowski:
Let me open a new PR trying the PAT route.

krassowski:
Apologies for the delay. It seems PAT is not the only way to achieve this, and possibly not the most secure one. This write up looks interesting: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs. The other appealing options are SSH deploy keys or GitHub App tokens (and we already have jupyterlab-probot). Thoughts?

blink1073:
That doc is out of date: I think of the options, a fine-grained token on the org level scoped to the repo is the best best.

Issue #15402, opened 2023-11-14
Send the cell model to the cell header and footer


brichet:
### Problem

The cell header and footer do not receive the cell model and are not designed to do so.

https://github.com/jupyterlab/jupyterlab/blob/388fbcccc6139fd45657f8d97811bc614f827416/packages/cells/src/widget.ts#L565
https://github.com/jupyterlab/jupyterlab/blob/388fbcccc6139fd45657f8d97811bc614f827416/packages/cells/src/widget.ts#L823

In addition, their height is fixed at *0px*.

https://github.com/jupyterlab/jupyterlab/blob/4e8012e73e5ae8f6e5d8df92a48bf5e63e42f776/packages/cells/style/headerfooter.css#L13

Therefore, it seems there is no way to implement header and footer adapted to the cell model (type, content...).

### Proposed Solution

Modify the `Cell.IContentFactory` interface, by adding a *CellModel* argument to `createCellHeader()` and `createCellFooter()`, even if the default implementation do not use it.

JasonWeill:
In Jupyter Scheduler, https://github.com/jupyter-server/jupyter-scheduler/issues/449 concerned `height: 0` being set, and was fixed by unsetting it. We used to set the height of a control to 0, even though we wanted it to be visible, because browsers at the time worked better with that rule in place. This doesn't seem to be the case anymore, because of browser rules engine changes, so I would encourage devs to remove `height: 0` rules that are not necessary.

brichet:
After some tests using a custom cell header (created in a custom cell content factory), I don't think that this issue is relevant.

Concerning the cell model, I didn't realized that it can be retrieved after attaching the header, using the `parent` property which return to the cell widget (including its model).

Concerning the height to `0px`, this rule is only applied on the default header, with the class `CELL_HEADER_CLASS`:
https://github.com/jupyterlab/jupyterlab/blob/d59aad9a25f18b1549a483339511621012fc0eaa/packages/cells/src/headerfooter.ts#L23-L34
I suppose that any custom cell header should simply not include this class.

AFAIK, the only thing to fix to use the cell header is the cell toolbar, whose top property is defined from the top of the cell, including the header. It should probably be set from the top of the cell input, to avoid overlapping with the header.

fcollonval:
Although using the `parent` attribute is an interesting work around, I think the API will be better by passing the model when constructing the header widget rather than building its content when it gets attached.

Issue #15305, opened 2023-10-26
Should we improve cell toolbar logic or disable cell toolbar in tests?


krassowski:
## Description

A bottleneck in the PR review are often the visual test failures. In addition to a few existing test cases which are flaky, a very common problem is the cell toolbar randomly showing up when it should not or randomly hiding when it should be visible for a split second. It is not a problem in normal usage when it is mostly unnoticeable, but it leads to many spurious failed tests, wasting hundreds of hours in CI time and in the time of maintainers, for example:

| Actual | Expected | Diff |
|--|--|--|
| ![reexpand-headers-03a-actual](https://github.com/jupyterlab/jupyterlab/assets/5832902/5f095d94-ed4d-45f6-8a93-ac1555faf1b7) | ![reexpand-headers-03a-expected](https://github.com/jupyterlab/jupyterlab/assets/5832902/69f27c42-64e3-4fc7-a574-962020e6f2c7) | ![reexpand-headers-03a-diff](https://github.com/jupyterlab/jupyterlab/assets/5832902/490d7f84-d360-4f92-bb83-72a99cae1635) |

| Actual | Expected | Diff |
|--|--|--|
| ![notebook-panel-1-actual](https://github.com/jupyterlab/jupyterlab/assets/5832902/da3dc8ef-e9a5-44ea-abe1-6aacc5410b3e) | ![notebook-panel-1-expected](https://github.com/jupyterlab/jupyterlab/assets/5832902/df5b733d-6a93-4222-aeee-46a207ec0324) | ![notebook-panel-1-diff](https://github.com/jupyterlab/jupyterlab/assets/5832902/603dd81b-86e1-4942-9c57-d6f28dea1d1b) |

## Reproduce

Screenshots taken from https://github.com/jupyterlab/jupyterlab/actions/runs/6640451043/job/18051878571?pr=15296 but every other PR has this issue. If you seen green status it likely means that one of the maintainers manually restarted the job after checking the failures manually (so they did that instead of improving the codebase or merging a PR).

JasonWeill:
This may be related to known bugs #13115 and #13857 regarding the cell toolbar's display. In addition, there's #12223, an enhancement request to make the cell toolbar visible all the time, perhaps in collapsed form.

krassowski:
Also, there is https://github.com/jupyterlab/jupyterlab/pull/15314 which might help with things.

krassowski:
[...]

PR #15422, opened 2023-11-17
Move the cell toolbar widget from the cell to the cell input

### LINKED ISSUE #13166: Side-by-Side rendering: cell toolbar disappears
## Description

Using Side-by-Side rendering, the cell toolbar might disappear in some situations:

https://user-images.githubusercontent.com/591645/193530736-40eccc9e-d36e-4608-b4d0-b666d0ce4a6c.mp4

## Reproduce

1. Create a new notebook
2. Add a new cell
3. Switch to `Side-by-side` rendering (`View > Render Side-by-Side`)
4. type `import this`, execute
5. The cell toolbar is gone for that cell

## Expected behavior

The cell toolbar should still be available when executing the stes described above.

## Context

- Operating System and version: Fedora 36
- Browser and version: Chrome 106

### LINKED ISSUE #15617: Cell toolbar position depends on cell header
## Description

I'm working on JupyterLab extension that will add widgets in cell header. I observe that when widgets are added to the cell header, then cell toolbar is moving up. It is expected to have cell toolbar always in the same position without dependency on cell header content.

#### Screenshot with header content
image

#### Screenshot without header content
image

## Reproduce

I attach code snippet that will create extension that will put two `div` elements to the cell header.

```ts
export class POCContentFactory extends NotebookPanel.ContentFactory {
  constructor(options: Cell.ContentFactory.IOptions) {
    super(options);
  }

  createCellHeader(): ICellHeader {
    return new POCCellHeader();
  }
}

export class POCCellHeader extends Widget implements ICellHeader {
  constructor() {
    super();
    this.layout = new PanelLayout();
    if (this.layout instanceof PanelLayout) {
      const t = document.createElement('div');
      t.textContent = 'Ola koduje';
      const a = new Widget({ node: t });
      this.layout.addWidget(a);
      const t2 = document.createElement('div');
      t2.textContent = 'Ola koduje';
      const b = new Widget({ node: t2 });
      this.layout.addWidget(b);
    }
  }
}

/* content factory extension for custom cell header */
const contentFactory: JupyterFrontEndPlugin<NotebookPanel.IContentFactory> = {
  id: 'piece-of-code-contend-factory-id',
  provides: NotebookPanel.IContentFactory,
  requires: [IEditorServices],
  autoStart: true,
  activate: (app: JupyterFrontEnd, editorServices: IEditorServices) => {
    const editorFactory = editorServices.factoryService.newInlineEditor;
    return new POCContentFactory({ editorFactory });
  }
};

```

## Expected behavior

Cell toolbar buttons position shouldn't depend on cell header.

I would be happy to provide PR with fix but I need assistance, where to look and where might be the issue. Maybe it is CSS issue?

brichet:
This PR changes the parent of the cell-toolbar, from the `Cell` widget to the `InputArea` widget.

## References

Fixes #15617
Fixes #13166 

As explained in https://github.com/jupyterlab/jupyterlab/issues/15402#issuecomment-1816055434 (end of comment), the cell toolbar is currently attached to the `Cell` widget. Its position is relative to the `Cell` dom, including the header.

Currently the cell header is not used, and has a height set to `0px`.

If the header get content (and height), the cell-toolbar position does not remain in the cell input.

## Code changes

Insert the `Toolbar` widget in the `Cell.inputArea` instead of the `Cell` itself, and modify the CSS accordingly.

## User-facing changes

Should not.

As it changes the parent of the toolbar, it may break some extensions that rely on this assumption.
i.e. probably breaks https://github.com/jupyterlab-contrib/jlab-enhanced-cell-toolbar (yet not compatible with jupyterlab 4)

cc. @fcollonval


fcollonval (commented):
The code looks good. But there is definitely side effect in tests.

fcollonval:
Nicolas do you know if this will fix https://github.com/jupyterlab/jupyterlab/issues/13166?

brichet:
> Nicolas do you know if this will fix #13166?

I didn't know this feature.
This fixes the problem, the toolbar is located in the input area. But if the input area is too short, the toolbar will disappear.

image

galata/test/jupyterlab/notebook-markdown.test.ts-snapshots/do-not-highlight-not-latex-jupyterlab-linux.png
krassowski, line 1: This looks unexpected, I think the cell toolbar should not show up here

galata/test/jupyterlab/notebook-run.test.ts-snapshots/notebook-panel-1-jupyterlab-linux.png
krassowski, line 1: Nor here.

galata/test/documentation/overview.test.ts-snapshots/interface-jupyterlab-documentation-linux.png
krassowski, line 1: Nor here. It looks the logic for hiding toolbar no longer works with this change?

brichet: Thanks @krassowski for reviewing it.
Actually I don't know when the toolbar is supposed to hide, as soon as there is an active cell.
I did some tries with a released version of Jupyterlab and the toolbar never hides, even if the focus is not on the notebook panel.
Do you know how I can test toolbar hiding?

krassowski: > Actually I don't know when the toolbar is supposed to hide, as soon as there is an active cell.

The toolbar should hide if there is an overlap with the text in the cell, but indeed it does not always do that immediately after reloading (at least on `main`), which I believe is a regression, and needs fixing:

should-hide

krassowski: Of note this is a major source of test flakiness https://github.com/jupyterlab/jupyterlab/issues/15305

brichet: Thanks for the explication, this PR is actually breaking this logic.
I'll look at it and let you know.

brichet:
I restored the logic to hide the toolbar, and reverted the modified snapshots.

galata/test/documentation/overview.test.ts-snapshots/interface-jupyterlab-documentation-linux.png
brichet: @krassowski the logic is fixed now

krassowski:
Thanks for the ping @brichet. This looks good to me, can you just add an entry in https://github.com/jupyterlab/jupyterlab/blob/main/docs/source/extension/extension_migration.rst on this change and its implications?

brichet:
This one should be ready now @krassowski.
Thanks again for the review.

krassowski (approved):
Thanks @brichet!

Issue #15437, opened 2023-11-23
Improve sidebar accessibility test coverage  


e218736:
### Problem

Tests for side bar tabs do not exist for keyboard navigation. It is important that tests exist to test the functionality of sidebar items using tabbing and arrow keys to ensure accessibility for keyboard users.  

### Proposed Solution

Add tests which follow a users navigation to the left and right side bar and verify that they can be opened 

### Additional context

Working on a ticket for this issue.

krassowski:
Closing as resolved by https://github.com/jupyterlab/jupyterlab/pull/15438.

PR #15438, opened 2023-11-23
Keyboard navigation to sidebar tests

e218736:
## References
#15437 Improve sidebar accessibility test coverage.

#15303 Issues noting keyboard navigation not as expected.

## Code changes

Adding these tests will validate that the side bar tabs and accordion panel items within tabs can be opened using only keyboard navigation (Tabbing/Arrow keys/Enter - not keyboard shortcuts) to prevent regression.

## User-facing changes
none

none

### DIFF
--- a/galata/test/jupyterlab/sidebars.test.ts
+++ b/galata/test/jupyterlab/sidebars.test.ts
@@ -226,7 +226,7 @@ test.describe('Sidebar keyboard navigation @a11y', () => {
     }
     await page.keyboard.press('Enter');
 
-    expect(await page.sidebar.isTabOpen('jp-property-inspector')).toEqual(true);
+    expect(await page.sidebar.isTabOpen('jp-running-sessions')).toEqual(true);
   });
 
   test('Open table of contents tab', async ({ page }) => {
@@ -305,7 +305,7 @@ test('Close then open File Browser tab', async ({ page }) => {
   }
   await page.keyboard.press('Enter');
 
-  expect(await page.sidebar.isTabOpen('filebrowser')).toEqual(false);
+  expect(await page.sidebar.isTabOpen('filebrowser')).toEqual(false); //doesn't work, you can't close it on enter!
 
   // eslint-disable-next-line no-constant-condition
   while (true) {
--- a/galata/test/jupyterlab/sidebars.test.ts
+++ b/galata/test/jupyterlab/sidebars.test.ts
@@ -203,12 +203,12 @@ test.describe('Sidebar keyboard navigation @a11y', () => {
   test('Open Running Terminals and Kernels tab', async ({ page }) => {
     await page.goto();
 
-    const fileBrowserTab = page.getByRole('tab', {
-      name: 'File Browser (Ctrl+Shift+F)'
-    });
-    const TerminalsAndKernelsTab = page.getByRole('tab', {
-      name: 'Running Terminals and Kernels'
-    });
+    // const fileBrowserTab = page.getByRole('tab', {
+    //   name: 'File Browser (Ctrl+Shift+F)'
+    // });
+    // const TerminalsAndKernelsTab = page.getByRole('tab', {
+    //   name: 'Running Terminals and Kernels'
+    // });
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
@@ -221,7 +221,6 @@ test.describe('Sidebar keyboard navigation @a11y', () => {
         break;
       }
     }
-    await page.keyboard.press('Enter');
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
@@ -244,18 +243,20 @@ test.describe('Sidebar keyboard navigation @a11y', () => {
   test('Open table of contents tab', async ({ page }) => {
     await page.goto();
 
-    const fileBrowserTab = document.getElementById('tab-key-1-6');
-    const tableOFContentsTab = document.getElementById('tab-key-1-4');
+    // const fileBrowserTab = document.getElementById('tab-key-1-6');
+    // const tableOFContentsTab = document.getElementById('tab-key-1-4');
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
       await page.keyboard.press('Tab');
-      let fileBrowserIsFocused = document.activeElement === fileBrowserTab;
-      if (fileBrowserIsFocused) {
+      let fileBrowserIsFocused = await page.evaluate(
+        () => document.activeElement
+      );
+      fileBrowserIsFocused?.getAttribute;
+      if (fileBrowserIsFocused?.getAttribute('data-id') === 'filebrowser') {
         break;
       }
     }
-    await page.keyboard.press('Enter');
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
--- a/galata/test/jupyterlab/sidebars.test.ts
+++ b/galata/test/jupyterlab/sidebars.test.ts
@@ -207,10 +207,9 @@ test.describe('Sidebar keyboard navigation @a11y', () => {
     while (true) {
       await page.keyboard.press('Tab');
       let fileBrowserIsFocused = await page.evaluate(
-        () => document.activeElement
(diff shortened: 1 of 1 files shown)


e218736:

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -194,3 +201,157 @@ test.describe('Sidebars', () => {
...
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
+        await page.keyboard.press('Tab');
+        let IsFocused = await page.evaluate(
+          () => document.activeElement?.getAttribute('data-id')
+        );
+        if (IsFocused === leftSidebarIds[0]) {
+          break;
+        }
+      }
```
brichet, line 245: Since this pattern can be found several time in the tests, it would be nice to have a generic function to navigate using keyboard as helper, maybe in [activity helper](https://github.com/jupyterlab/jupyterlab/blob/main/galata/src/helpers/activity.ts).
That way it will be available for futur navigation tests.

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -194,3 +201,157 @@ test.describe('Sidebars', () => {
...
+
+      await page.keyboard.press('Enter');
+
+      const isExpanded = await page.evaluate(
+        () => document.activeElement?.getAttribute('aria-expanded')
+      );
+
+      expect(isExpanded).toBeTruthy();
```
brichet, line 354: We should expect `isExpanded` to be false, the accordion is opened by default, and close when pressing `entrer` the first time.
`isExpanded` here is a string containing 'false', the test pass because it is not a empty string.

For more robustness we could check the initial state, press enter, check that the state changed, and do it again (to ensure opening and closing works well on enter).

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -194,3 +201,157 @@ test.describe('Sidebars', () => {
...
+];
+const propertyElementAriaLabels: string[] = [
+  'Warning Section',
+  'Installed Section',
+  'Discover Section'
+];
+
+const sessionsAndDebuggerAriaLabels: string[] =
+  runningSessionsElementAriaLabels.concat(debuggerElementAriaLabels);
+
```
brichet, line 226: We could use a structure like 
```ts
elementAriaLabels = {
  'jp-running-sessions': [
    'Open Tabs Section',
    '...'
  ],
  'jp-debugger-sidebar': [
    '...',
  ],
  'extensionmanager.main-view': [
    '...',
  ]
}
```
and have only one test function, looping first on tab name and then on section name.
And maybe something similar can be done for left and right sidebars.

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -194,3 +201,157 @@ test.describe('Sidebars', () => {
...
+test.describe('Sidebar keyboard navigation @a11y', () => {
+  leftSidebarIds.forEach(leftSidebarId => {
+    test(`Open Sidebar tab ${leftSidebarId} via keyboard navigation`, async ({
+      page
+    }) => {
+      await page.goto();
```
brichet, line 232:
```suggestion
```
This is done by default, it just reload the page, but is not necessary.

brichet: Same comment for the other tests.

brichet (changes requested):
Thanks @e218736 for working on this, it is really important to add tests on accessibility.

I have some comments below.

galata/test/jupyterlab/sidebars.test.ts
e218736: Hi @brichet, thanks for your comments. I've implemented most however, I'm struggling to get the the left/right side bar test passing with use of a helper function. Would you be able to advise me on where I'm going wrong in implementing this?

Thanks

brichet: @e218736 you should try to run the tests locally, it would be easier to debug.
You need 2 terminals in your develop environment, and in the directory `galata`.
In the first one, you run `jlpm start` to start jupyterlab for the tests.
In the second one you run 
- `jlpm test test/jupyterlab/sidebars.test.ts` to test the whole file
- or `jlpm test test/jupyterlab/sidebars.test.ts -g 'via keyboard navigation'` (for example) to test only some specific test (using regex).

In this specific case, you are not sending the correct `dataId`.
`sideBarTabName` is the `data-id` (a string). So `sideBarTabName[0]` is only the first letter of the string.

To make it more generic, you can also add a parameter `attributeName` to the function, to be able to use it also with the "*open accordion panel*" tests.

e218736: Hi @brichet,

Apologies but I'm unable to run tests locally. 

Thank you for the comments, I've amended if you are able to review again please.

brichet (commented):
Thanks @e218736 for updating the PR.
I have a few minor comments below.

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -194,3 +201,83 @@ test.describe('Sidebars', () => {
...
+
+      for (let dataId of keyValueArray) {
+        await page.goto();
+        const nextFocused = await page.activity.keyToSidebar(
+          keyValueArray[0],
+          'Tab',
+          'data-id'
+        );
+
+        const newParentElement = nextFocused as unknown as Locator;
```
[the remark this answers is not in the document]
e218736: Thank you for your comments again, have amended your suggestions :)

brichet (approved):

krassowski:
Re-tagged this to 4.2.0 because it adds a new method to galata. Let's merge this as soon as we switch `main` to track 4.2.0 (which I propose to do next week/after 4.1.2 release).

galata/src/helpers/activity.ts
```diff
@@ -59,6 +59,29 @@ export class ActivityHelper {
...
 
+  /**
+   * Keyboard navigation to element using navigation key
+   *
+   * @param attributeValue value of selected widget attribute
+   * @param key navigation key to press
+   * @param attribute name of attribute selector
+   */
+
+  async keyToSidebar(
```
krassowski, line 70: Having a second look at this I could not understand what this helper does from reading the docstring alone; I needed to read the code to understand and have further request on it as this would become a part of public API and a headache to change in the future if needed:
- can you clarify the docstring that it will press the key until it reaches an element
- the name needs changing: the docstring mentions generic  "element" but function name includes "sidebar"; if it means to be generic, it should be named as such
- the key can be customized; but in fact is expected to be either a `Tab`, `ArrowDown`, `ArrowUp` or possibly a few others; maybe it would be worth replacing `string` with `'Tab' | 'ArrowDown' | 'ArrowUp'`?
- the `attribute` and `attributeValue` are too specific to make this helper useful in the general case; I would suggest to replace these two with a single `selector: string` argument and use `document.activeElement.matches(selector)` instead.
- there is an empty new line between doscting and function, it needs to be deleted

krassowski (changes requested):
A few small requests on the public API before we merge this.

galata/src/helpers/activity.ts
e218736: Thanks for the suggestions, have updated

krassowski (approved):
LGTM, thanks @e218736

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -205,3 +211,74 @@ test.describe('Sidebars', () => {
     expect(tableOfContentsElementRole).toEqual('region');
   });
 });
+
+const elementAriaLabels = {
+  'jp-running-sessions': [
+    'Open Tabs Section',
+    'Kernels Section',
+    'Language servers Section',
+    'Terminals Section'
```
krassowski, line 220: I guess we will need these to reflect new additions:

```suggestion
    'Language servers Section',
    'Recently Closed Section',
    'Workspaces Section',
    'Terminals Section'
```

PR #15443, opened 2023-11-27
Fix search coming back in notebook and editor

krassowski:
## References

Builds on the test proposed in #15259 (thanks @parmentelat)
Fixes https://github.com/jupyterlab/jupyterlab/issues/14871

## Code changes

Track the search on/off state and only announce state changes/act on them if the search is on.

## User-facing changes

Search does not come back in notebook and file editor

None


krassowski:
`jupyter lab --dev-mode` (do you see the orange dev mode indicator at the top)? 

If you still see it in the dev mode, can you record a screencast?

parmentelat:
just to confirm, I have been able to test this manually with larger notebooks and so far I have not seen the search dialog reappearing unintentionally, so it's all good as far as I am concerned; at the very least it is **way better** than before :)

thanks a lot @krassowski for fixing this, it is a huge burden off heavy users who use Search all the time

ps1. I tried to analyse the outcome of CI that has many failures, it does not seem to relate to these changes but I could not say for sure

ps2. just as a comment, not quite related to the topic at hand, but
I think I remember having seen issues where people were not at ease with the search feature switching cells in edit mode
I think I understand the need to do that in cases where the search pattern would be present in the cell source but not in the rendered version;  however in the vast majority of cases, it's not the case, so after all these manual tests, it's my feeling that this issue is maybe next thing that could be improved :)

krassowski:
> ps1. I tried to analyse the outcome of CI that has many failures, it does not seem to relate to these changes but I could not say for sure

No worries, I will take care of these. Briefly there are two issues: (a) we need to update snapshots as new PRs were merged since and the UI looks slightly different (b) parallel execution of tests on CI is breaking some test cases which is being worked on elsewhere.

> I think I remember having seen issues where people were not at ease with the search feature switching cells in edit mode
I think I understand the need to do that in cases where the search pattern would be present in the cell source but not in the rendered version; however in the vast majority of cases, it's not the case, so after all these manual tests, it's my feeling that this issue is maybe next thing that could be improved :)

I agree from the UX point of view. Having it not touch render state and showing a tooltip "3 more matches in source code" could be a good solution IMO. If you cannot find an issue about this, please create a new one.

galata/test/jupyterlab/notebook-search-highlight.test.ts-snapshots/notebook-search-highlight-1-jupyterlab-darwin.png
krassowski, line 1: Comment to self: crop the screenshot to content to avoid flaky status indicator

galata/test/jupyterlab/notebook-search-highlight.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+test.beforeEach(async ({ page, tmpPath }) => {
+  await page.notebook.openByPath(`${tmpPath}/${TEST_FILENAME}`);
+  await page.notebook.activate(TEST_FILENAME);
+});
+
+test('Open and close Search dialog, then add new code cell', async ({
+  page
+}) => {
+  const imageName1 = 'notebook-search-highlight-1.png';
```
krassowski, line 27: Note  to self: rename.

krassowski:

galata/test/jupyterlab/notebook-search-highlight.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  await page.evaluate(async () =>
+    window.jupyterapp.commands.execute('notebook:insert-cell-below')
+  );
+
+  // wait an arbitrary amount of extra time
+  // and expect the outlining to be still gone
+  // but because of #14871, text is highlighted again
+  setTimeout(async () => {
+    expect(await notebook.screenshot()).toMatchSnapshot(imageNameAfter);
+  }, 1000);
```
krassowski, line 60:
```suggestion
  await new Promise(resolve => setTimeout(resolve, 1000));
  expect(await notebook.screenshot()).toMatchSnapshot(imageNameAfter);
```

krassowski:
bot please update galata snapshots again

parmentelat:
@krassowski 
thanks for youor invitation to review this PR; I can't say that I feel exactly relevant, but FWIW I do not see anything sticking out in the code changes

in any case I'd like to thank you a million times again for taking care of this nagging issue, I am eager to see this released so I can stop writing junk in the search dialog only to keep the issue from happening ;-)

fcollonval (approved):
Thanks a lot at @krassowski

krassowski:

PR #15448, opened 2023-11-28
Support server-side execution

### LINKED ISSUE #2833: Reconnect to running session: keeping output
There's been a lot of requests in the past to:
* Reconnect to a long-running notebook and keep receiving new output.
* Reconnect to a notebook running on the server side and recover all output when reconnecting.

I'd like to resurface this issue. To the best of my knowledge, these features have been planned at least since 2015 (earliest reference I found: [link](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jupyter/8hyGKoBY6O0/RyEDDyOZAQAJ)). It hasn't been implemented yet because it directly conflicts with the architecture of the stack and would require a big refactor.

I don't know if this is easier in the JupyterLab context or not. But now that this project is in very active development, it might be a good opportunity to revisit this and try to get it in now at the foundations.

I found to open issues on the topic in the old notebooks repo:
* jupyter/notebook#1150
* jupyter/notebook#641

### LINKED ISSUE #12867: Support new kernels REST API
### Problem

Jupyverse just got support for a [new code execution mode](https://github.com/jupyter-server/jupyverse/pull/191) through a POST request, where the front-end doesn't have to deal with the kernel protocol (this is handled in the back-end). It's still experimental and limited (e.g. widgets are currently not supported), but it could be interesting to start supporting that mode in JupyterLab.

### Proposed Solution

Support code execution using the new POST request instead of using the kernel protocol over WebSocket when JupyterLab is launched with e.g. `--post_execute_request` and RTC is enabled.

### Additional context

https://github.com/jupyter-server/jupyter_server/issues/900.

davidbrochart:
## References

Closes #2833.
Closes #12867.

Needs:
- ~~https://github.com/jupyterlab/jupyter-collaboration/pull/217~~
- ~~https://github.com/jupyter-server/jupyter_ydoc/pull/198~~
- ~~https://github.com/jupyter-server/jupyter_ydoc/pull/201~~
- `jupyter-collaboration >=2.0`
- `jupyverse >=0.4.0`

Currently, the only Jupyter server that supports this execution mode is [Jupyverse](https://github.com/jupyter-server/jupyverse).

## Code changes

A new `server-side-execution` flag allows to execute a notebook server-side (false by default). In this execution mode, the frontend doesn't use the kernel protocol over WebSocket, but interacts with the notebook only through its shared model. Thus, this mode needs [jupyter-collaboration](https://github.com/jupyterlab/jupyter-collaboration).

## User-facing changes

The user can close their JupyterLab browser window, reopen it later and recover the notebook state, including (live) cell outputs.

None.

### DIFF
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -499,6 +499,10 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
         {"LabApp": {"custom_css": True}},
         "Load custom CSS in template html files. Default is False",
     )
+    flags["server-side-execution"] = (
+        {"LabApp": {"server-side-execution": True}},
+        """Whether to execute notebooks in the server using the REST API, not using the kernel protocol over WebSocket. The frontend only interacts with the notebook through its shared model. This execution mode requires jupyter-collaboration.""",
+    )
 
     subcommands = {
         "build": (LabBuildApp, LabBuildApp.description.splitlines()[0]),
@@ -601,6 +605,12 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
             python -m pip install jupyter_collaboration
 
         This flag is now deprecated and will be removed in JupyterLab v5.""",
+    )
+
+    server_side_execution = Bool(
+        False,
+        config=True,
+        help="""Whether to execute notebooks in the server using the REST API, not using the kernel protocol over WebSocket. The frontend only interacts with the notebook through its shared model. This execution mode requires jupyter-collaboration.""",
     )
 
     news_url = Unicode(
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -18,9 +18,9 @@ import {
   MarkdownCell
 } from '@jupyterlab/cells';
 import { Notification } from '@jupyterlab/apputils';
-import { signalToPromise } from '@jupyterlab/coreutils';
+import { PageConfig, signalToPromise, URLExt } from '@jupyterlab/coreutils';
 import * as nbformat from '@jupyterlab/nbformat';
-import { KernelMessage } from '@jupyterlab/services';
+import { KernelMessage, ServerConnection } from '@jupyterlab/services';
 import { ISharedAttachmentsCell } from '@jupyter/ydoc';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { every, findIndex } from '@lumino/algorithm';
@@ -2528,36 +2528,60 @@ namespace Private {
 
           let ran = false;
           try {
-            const reply = await CodeCell.execute(
-              cell as CodeCell,
-              sessionContext,
-              {
-                deletedCells,
-                recordTiming: notebook.notebookConfig.recordTiming
+            if (PageConfig.getOption('serverSideExecution') === 'true') {
+              const kernelId = sessionContext.session?.kernel?.id;
+              const settings = ServerConnection.makeSettings();
+              const apiURL = URLExt.join(
+                settings.baseUrl,
+                `api/kernels/${kernelId}/execute`
+              );
+              const cellId = cell.model.sharedModel.getId();
+              const documentId = `json:notebook:${notebook.model?.sharedModel.getState(
+                'file_id'
+              )}`;
+              const body = `{"cell_id":"${cellId}","document_id":"${documentId}"}`;
+              const init = {
+                method: 'POST',
+                body
+              };
+              try {
+                await ServerConnection.makeRequest(apiURL, init, settings);
+              } catch (error) {
+                throw new ServerConnection.NetworkError(error);
               }
-            );
-            deletedCells.splice(0, deletedCells.length);
+              ran = true;
+            } else {
+              const reply = await CodeCell.execute(
+                cell as CodeCell,
+                sessionContext,
+                {
+                  deletedCells,
+                  recordTiming: notebook.notebookConfig.recordTiming
+                }
+              );
+              deletedCells.splice(0, deletedCells.length);
 
-            ran = (() => {
-              if (cell.isDisposed) {
-                return false;
-              }
-
-              if (!reply) {
-                return true;
-              }
-              if (reply.content.status === 'ok') {
-                const content = reply.content;
-
-                if (content.payload && content.payload.length) {
-                  handlePayload(content, notebook, cell);
+              ran = (() => {
+                if (cell.isDisposed) {
+                  return false;
                 }
 
-                return true;
-              } else {
-                throw new KernelError(reply.content);
-              }
-            })();
+                if (!reply) {
+                  return true;
+                }
+                if (reply.content.status === 'ok') {
+                  const content = reply.content;
+
+                  if (content.payload && content.payload.length) {
+                    handlePayload(content, notebook, cell);
+                  }
+
+                  return true;
+                } else {
+                  throw new KernelError(reply.content);
+                }
+              })();
+            }
           } catch (reason) {
             if (cell.isDisposed || reason.message.startsWith('Canceled')) {
               ran = false;


matthewwiese:
This is really great David! So cool to follow your progress on this. 🙂

davidbrochart:
Here is a demo:

https://github.com/jupyterlab/jupyterlab/assets/4711805/c65b5253-da77-4203-9223-b9097e6688c6

krassowski:
> server-side-execution

While I see why you named it "server-side-execution", I wonder if naming it "server-side-composition" could potentially avoid confusion with the execution which happens in kernels? It feels like the fact that execution is triggered from server (well, really proxied) is an implementation detail and what users care about is where the execution results are composed into the document.

davidbrochart:
I agree that "server-side-execution" is not great, because kernels always execute code in the server. The main difference here is that:
- execution is triggered by a higher-level protocol (an HTTP POST instead of the frontend talking to the kernel through the kernel protocol over WebSocket), and the backend is responsible for talking to the kernel through the kernel protocol over ZMQ.
- cell outputs are not populated by the frontend, but by the backend, and the frontend sees the output changes reflected in their shared model.

But I'm not sure about "server-side-composition" either. Maybe "server-side-outputs"?

jtpio:
> but interacts with the notebook only through its shared model. Thus, this mode needs [jupyter-collaboration](https://github.com/jupyterlab/jupyter-collaboration).

Curious if this could be implemented in `jupyter-collaboration` directly then? Just trying to see if this implicit dependency could create issues in the future since `jupyterlab` and `jupyter-collaboration` evolve at a different pace.

davidbrochart:
The execution command is done through an HTTP POST instead of using the kernel protocol, so there are changes to be done in JupyterLab:
https://github.com/jupyterlab/jupyterlab/blob/21cc49a32669a998d2f73f5b0ed3f4ea9793a40e/packages/notebook/src/actions.tsx#L2480-L2500
But I'm not sure it's the right place.

davidbrochart:
Here is a way to try it out:

```bash
mkdir jupyterlab
cd jupyterlab
git clone https://github.com/davidbrochart/jupyterlab.git
git clone https://github.com/davidbrochart/jupyverse.git
pixi init
pixi add nodejs pip "python<3.12"
pixi shell

cd jupyterlab
git checkout server-side-execution
pip install -e .cd ..

cd jupyverse
git checkout ywidget-subdoc
pip install -e jupyverse_api
pip install -e plugins/noauth
pip install -e plugins/contents
pip install -e plugins/frontend
pip install -e plugins/lab
pip install -e plugins/jupyterlab
pip install -e plugins/kernels
pip install -e plugins/yjs
pip install -e plugins/nbconvert
pip install -e plugins/terminals
pip install -e .
cd ..

pip install ypywidgets==0.6.3
pip install jupyter-collaboration

mkdir tmp
cd tmp
jupyverse --set jupyterlab.dev_mode=true --set jupyterlab.server_side_execution=true --set kernels.require_yjs=true
```

Open a browser window at http://127.0.0.1:8000, it should open JupyterLab.
Create a notebook with the following cell and run it:
```py
from time import sleep

for i in range(1000000):
    print(i)
    sleep(2)
```
Close your browser window and open a new one at http://127.0.0.1:8000. Open the notebook, you should see the live outputs.

jtpio:
Also still curious if this could be implemented in `jupyter-collaboration` directly (https://github.com/jupyterlab/jupyterlab/pull/15448#issuecomment-1833511805)? If it's not, maybe we can provide an extension point to do it as an alternative?

jtpio:
> Where would this extension point be?

Currently there is no such extension point, as the notebook execution logic seems to be living in a `Private` namespace which makes it difficult to customize indeed. But ideally we could imagine having a plugin that would allow for using a different execution mode (WebSocket or HTTP POST), that would be made available to the notebook actions.

davidbrochart:
Yes I was looking for something like that, but eventually took a shortcut. What I like in the current solution is that it's simple and it also doesn't drop the connection to the kernel through the kernel protocol over WebSocket, which means that the kernel busy indicator still works, and widgets should also still work (although they won't be recovered from the notebook state).
I see this PR as an incremental step towards completely getting rid of the kernel protocol over WebSocket in the frontend, which will require more work.

davidbrochart:
I released Jupyverse v0.4.0, so testing this PR is now simpler:
```bash
mkdir jupyterlab
cd jupyterlab
git clone https://github.com/davidbrochart/jupyterlab.git
pixi init
pixi add nodejs pip "python<3.12"
pixi shell

cd jupyterlab
git checkout server-side-execution
pip install -e .

pip install "jupyverse[jupyterlab,noauth]"
pip install "ypywidgets>=0.6.4"
pip install "jupyter-collaboration>=2.0.0"

mkdir tmp
cd tmp
jupyverse --set jupyterlab.dev_mode=true --set jupyterlab.server_side_execution=true --set kernels.require_yjs=true
```
I changed the [top comment](https://github.com/jupyterlab/jupyterlab/pull/15448#issue-2014190659) and updated what is needed for this feature to be enabled (mainly, `jupyverse` and `jupyter-collaboration`).
Since this PR externalizes almost everything in the backend, except for the HTTP POST to execute a cell, which is under a flag, I don't see any reason to delay merging, what do you think?

krassowski:
> I don't see any reason to delay merging, what do you think?

Main still tracks 4.1, so my suggestion would be to merge as soon as this is reviewed and 4.2 branch out (unless there is a consensus to make an exception and add it in 4.1 but it is _really late_ for this).

Wh1isper:
I prefer to name this feature as `server-side-save/saving`,  

Previously we performed the save operation in the browser, now we perform the save operation on the server side and the browser just receives the push from the server side (via collaboration).

It's very pleased to see `Jupyverse` offering more features, which will motivate us to evolve towards `Jupyverse`!

davidbrochart:
Now that JupyterLab v4.1.0 is out, should we go ahead with this PR?

echarles:
I see needed definitions and clarifications on the proposed features. Looking at the changes, only cellid and docid are sent, which is fine for server-side model. There should be a way to continue supporting code execution via REST with client-side model.

davidbrochart:
Yes, in order to execute a cell of a notebook, the notebook document ID and the notebook cell ID are needed.
Support for client-side execution is still there, since the new server-side execution mode is under a flag.

jtpio:
> But ideally we could imagine having a plugin that would allow for using a different execution mode (WebSocket or HTTP POST), that would be made available to the notebook actions.

FYI @fcollonval just opened a PR to allow for something like this: https://github.com/jupyterlab/jupyterlab/pull/15830. Thanks Fred!

EduardDurech:
Does anything still need to be done to merge this? I can give some time as this is critical and still the top 👍 issue after many years

In the meantime until the next release, can we just build locally and run this to get the functionality?

davidbrochart:
I'm sorry this is taking so long, I'd love it to be released in JupyterLab 4.2.
But yes building JupyterLab with this PR is all that is needed to try it out, since Jupyverse has been released with the feature.

jtpio:
https://github.com/jupyterlab/jupyterlab/pull/15830 is likely the way forward with this, with `jupyter-collaboration` providing its own `ICellExecutor`.

davidbrochart:
Sure, #15830 is the right solution long-term, but maybe we could live with the shortcut provided by this PR in the meantime?

jtpio:
#15830 is almost ready and also planned for 4.2.0. So I'm not sure there is a need to merge this PR in the meantime.

Once https://github.com/jupyterlab/jupyterlab/pull/15830 is merged and available in a 4.2 pre-release, the logic in this PR can be implemented in `jupyter-collaboration` directly instead.

krassowski:
@davidbrochart FYI https://github.com/jupyterlab/jupyterlab/pull/15830 is now merged.

davidbrochart:
Thanks, I'll update this PR.

krassowski:
@davidbrochart is the plan to add a new implementation of `ICellExecutor` as a new plugin in this PR or to add it in `jupyter-collaboration` for time being and possibly bring the non-RTC bits of jupyter-collaboration into the core in the next release, say 4.3 (we are already past feature freeze on 4.2)?

davidbrochart:
You're right, I think a new implementation of `ICellExecutor` as a new plugin should be done in jupyter-collaboration, as this execution mode doesn't make sense in non-RTC mode.

davidbrochart:
Closing in favor of https://github.com/jupyterlab/jupyter-collaboration/pull/279.

PR #15440, opened 2023-11-25
Fix Shift + L not working in stdin

krassowski:
## References

Fixes #15429

## Code changes

Adds selector that should have been added before.

## User-facing changes

When input box is focused, shift + l types "L" rather than toggling line numbers

## Backwards-incompatible changes

None

Issue #15326, opened 2023-10-29
Don't pre-populate keyboard shortcut user settings at first run


firai:
### Problem

JupyterLab currently copies the entire set of default keyboard shortcut settings into the user's settings when JupyterLab is run the first time in an environment. However, since user's settings are stored in `$HOME` by default rather than per-installation, the keyboard shortcut settings don't get updated when the defaults change in newer JupyterLab releases, even if the newer JupyterLab is installed "fresh".

An example of changing default keyboard shortcuts in core is #14421. Extensions also implement changes to defaults in keyboard shortcuts, such as https://github.com/jupyterlab-contrib/jupyterlab-vim/pull/107 and https://github.com/jupyterlab-contrib/jupyterlab-vim/pull/110.

While changes are generally limited to minor releases rather than patch releases, and there may be some expectation that user settings may need to be reset after such an "upgrade", the current setup increases the risk of conflict (https://github.com/jupyterlab-contrib/jupyterlab-vim/issues/124). The only way to fix this after the fact is to either manually remove/merge the offending settings, or reset the entire keyboard shortcut settings and recopy customizations truly desired by the user.

### Proposed Solution

Don't pre-populate keyboard shortcut user settings with the default settings at first run. Only write a rule into the user's settings when the user requests customization. Although rules customized by the user can still cause conflicts, the potential for such conflicts is reduced.

### Additional context

The problem this proposal is trying to solve is related to #8335, but this proposal would provide an immediate mitigation ahead of implementing #8335.

PR #15451, opened 2023-11-28
Backport PR #15440 on branch 4.0.x (Fix Shift + L not working in stdin)

fcollonval:
Backport PR #15440

---

The users may need to reset the keyboard shortcut fo the fix to be applied - related issue https://github.com/jupyterlab/jupyterlab/issues/15326


krassowski (approved):
Thanks for the backport! I agree, while users who changed the shortcut will still be affected the benefits outweigh the cons

PR #15452, opened 2023-11-28
Add a facultative driveName option to the BreadCrumbs and FileBrowser IOptions interfaces.

HaudinFlorence:
In the idea of working with more than one filebrowser, it would be of interest to display the name of the drive in the breadcrumbs.
To do this, we propose here to add a facultative driveName option to the breadCrumbs.IOptions and to the FileBrowser.IOptions interfaces and to
modify the createCrumbSeparators method to be able to display the drive name inside the breadCrumbs. 
Screenshot from 2023-11-28 18-53-36

## Code changes
- Add a facultative FileName option in FileBrowser.IOptions interface in `packages/filebrowser/src/browser.ts`
- Add a facultative FileName option in BreadCrumbs.IOptions interface in `packages/filecrowser/src/crumbs.ts`
- Modify the createCrumbsSeparators in Private namespace in `packages/filecrowser/src/crumbs.ts` to support the possibility to display the driveName in the breadcrumbs if there is a driveName parameter given for the filebrowser.

The changes should not be backwards-incompatible since the option introduced is optional.


HaudinFlorence:
bot please update snapshots

packages/filebrowser/src/browser.ts
```diff
@@ -72,7 +72,11 @@ export class FileBrowser extends SidePanel {
     this.mainPanel.addClass(FILE_BROWSER_PANEL_CLASS);
     this.mainPanel.title.label = this._trans.__('File Browser');
 
-    this.crumbs = new BreadCrumbs({ model, translator });
+    this.crumbs = new BreadCrumbs({
+      model,
+      translator,
+      driveName: 'mydrive'
```
krassowski, line 78: I assume this is still a draft?

HaudinFlorence: @krassowski It was not a draft but I just converted it into one.

HaudinFlorence: I will close this PR. This solution is not the most suited one to display a drive name in a drive filebrowser.

PR #15463, opened 2023-11-30
Fix file dialog contents loading with `defaultPath`, and model disposal

### LINKED ISSUE #15453: File dialog loads slowly when using `defaultPath`
## Description
When using `FileDialog.getOpenFiles` with the `defaultPath` option, it takes a while (~10 seconds) to load the file list. However, when it's manually refreshed, the file list loads immediately.
OpenDialogBug

When `defaultPath` is not given, the file list is loaded immediately, however this limits the file dialog to starting from the root folder.

## Reproduce

Create a file dialog with `FileDialog.getOpenFiles` and pass a `defaultPath` option.

## Expected behavior
I would've expected the file list the to load once the dialog opened, especially since manually refreshing works immediately.

## Context

- Browser and version:  Firefox 119
- JupyterLab version: 4.1.0a3

### LINKED ISSUE #15458: File dialog doesn't terminate its file browser model when closed
## Description
It appears that the `FilterFileBrowserModel` created by a `OpenDialog` is not terminated when the file dialog is closed. I'm not sure if this is intended behavior, but this results in extraneous content requests. The image shows two file browsers polling the root directory. Requests underlined in red are created by the default file browser. Requests underlined in green are created by the file browser created by a file dialog. Although the file dialog is closed by 13:07:00, the requests still persist indicating that its file browser model is still polling.
image
Possibly related to #15453?

## Reproduce
1. Create a file dialog using either `FileDialog.getOpenFiles` or `FileDialog.getExistingDirectory`
2. Close the dialog
3. Observe content requests still persisting

## Expected behavior
I expect the requests to terminate once the dialog is closed.

## Context

- JupyterLab version: v4.1.0a3

mmichilot:
## References
Fixes #15458
Fixes #15453

### File browser model still active after closing file dialog
Whenever a file dialog was closed, the file browser model was still active and polling, resulting in extraneous content requests. Although benign, the requests could accumulate harming responsiveness. The fix involves overriding the `dispose()` method to dispose the model before the file dialog is disposed.
```
// Dispose of file browser model when dialog closes
this.dispose = () => {
  this._browser.model.dispose();
  super.dispose();
};
```

### File dialog not displaying contents on launch when using `defaultPath`
As referenced in #15453, there was seemingly a delay whenever a file dialog was opened with `defaultPath`. The issue was that 
the model changed directory before it was passed to `FileBrowser`. The fix involves moving when the directory is changed till after `FileBrowser` is instantiated. When the model changes directory, it appropriately signals to `FileBrowser` to update its listings.

### DIFF
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -249,6 +249,17 @@ export class FileBrowser extends SidePanel {
     return this.listing.paste();
   }
 
+  /**
+   * Dispose of the resources held by the browser.
+   */
+  dispose(): void {
+    if (this.isDisposed) {
+      return;
+    }
+    this.model.dispose();
+    super.dispose();
+  }
+
   private async _createNew(
     options: Contents.ICreateOptions
   ): Promise<Contents.IModel> {
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -249,17 +249,6 @@ export class FileBrowser extends SidePanel {
     return this.listing.paste();
   }
 
-  /**
-   * Dispose of the resources held by the browser.
-   */
-  dispose(): void {
-    if (this.isDisposed) {
-      return;
-    }
-    this.model.dispose();
-    super.dispose();
-  }
-
   private async _createNew(
     options: Contents.ICreateOptions
   ): Promise<Contents.IModel> {
--- a/packages/filebrowser/src/opendialog.ts
+++ b/packages/filebrowser/src/opendialog.ts
@@ -196,12 +196,6 @@ class OpenDialog
         // Set Widget content
         this.layout = layout;
 
-        // Dispose of file browser model when dialog closes
-        this.dispose = () => {
-          this._browser.model.dispose();
-          super.dispose();
-        };
-
         this._ready.resolve();
       })
       .catch(reason => {
--- a/packages/filebrowser/src/opendialog.ts
+++ b/packages/filebrowser/src/opendialog.ts
@@ -291,15 +291,15 @@ namespace Private {
       filterDirectories
     });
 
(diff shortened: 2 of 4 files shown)


krassowski (commented):
It it feasible to add unit-test for these (in [openfiledialog.spec.ts](https://github.com/jupyterlab/jupyterlab/blob/main/packages/filebrowser/test/openfiledialog.spec.ts))?

mmichilot:
I apparently misplaced a `defaultPath` test with the `#items()` tests, so the test has been moved to the `FileDialog.getOpenFiles()` tests.

mmichilot:
> Maybe a timer could be added to the existing `defaultPath` tests to check whether the listing is available within 10 seconds, but otherwise I don't think it's necessary.

Yeah, I'm not sure if its possible to test this. The existing tests already had a 1 second timeout waiting for the listing to populate.

packages/filebrowser/src/browser.ts
```diff
@@ -249,6 +249,17 @@ export class FileBrowser extends SidePanel {
...
   }
 
+  /**
+   * Dispose of the resources held by the browser.
+   */
+  dispose(): void {
+    if (this.isDisposed) {
+      return;
+    }
+    this.model.dispose();
```
krassowski, line 259: I see why you moved it here, though I am worried it could backfire because model is not created by the browser but given to it in options. I think we should use previous implementation as this may lead to disposal of model which is still used elsewhere.

mmichilot:
With the dispose in OpenDialog, I'm still not sure how to develop a unit test since OpenDialog is not exposed.

krassowski (approved):
Ok, I agree adding tests for this particular code path is hard and we should ideally refactor the implementation to make it testable but that would be a separate PR.

Thanks @mmichilot!

Issue #15447, opened 2023-11-27
All `FormComponent` fields have an error indicator when `customValidate` function is passed


mmichilot:
## Description
When using `FormComponent` with a `customValidate` function, all fields have an error indicator
![rsz_screenshot_2023-11-27_120900](https://github.com/jupyterlab/jupyterlab/assets/22762462/38656174-c5cc-4a25-832c-f5e71cbe270a)

## Reproduce
Create a custom form using `FormComponent`, then pass a `customValidate` function to it. I found the issue to be with the following code in `CustomTemplateFactory`  function, which is used by `FormComponent`.
```
 {!hasCustomField &&
  (rawErrors ? (
    // Shows a red indicator for fields that have validation errors
    <div className="jp-modifiedIndicator jp-errorIndicator" />
  ) : (
    // Only show the modified indicator if there are no errors
    isModified && <div className="jp-modifiedIndicator" />
  ))}
```

When RJSF validates a form, if there are no errors, then `rawErrors` is either `undefined` with no `customValidate` function or an empty array with a `customValidate` function. Since an empty array is still defined, the `div` indicating an error is rendered. 

## Expected behavior
Upon removing the `customValidate` function, the form behaves as normal, with the error indicator only appearing when there is a validation error.
![rsz_screenshot_2023-11-27_124428](https://github.com/jupyterlab/jupyterlab/assets/22762462/b7fff34a-9106-4b81-8f74-7e18194cd61c)

## Context

- JupyterLab version: 4.1.0a3
I have also observed this error in version 4.0.9

## Solution
Add an additional check to check the length of the `rawErrors` array. I'm willing to create a pull request for this, if needed.

krassowski:
> I'm willing to create a pull request for this, if needed.

A PR would be appreciated!

PR #15464, opened 2023-11-30
Fix `FormComponent` showing error indicators in all fields when using a `customValidate` function

mmichilot:
## References
Fixes: #15447

## Code changes
Also check the length of `rawErrors` since `rawErrors` will be an empty array rather than `undefined` when a `customValidate` function is passed to `FormComponent`.


mmichilot:
This PR should be ready, is there anything else that I should do?

krassowski (approved):
Indeed, this looks good thanks @mmichilot!

krassowski:

PR #15479, opened 2023-12-02
Workaround/input box focus

### LINKED ISSUE #12322: ipdb focus issues causes some characters to be put into cell above
## Description

If you push enter to fast in ipdb, this puts the character into the input above. While this may result in some extra whitespace, it could delete code if you had any text selected.

## Reproduce
```
import ipdb
def inf_loop():
    i = 0
    while True:
        print(i)
        i += 1
ipdb.runcall(inf_loop)
```
Push n then hold enter.
spaces

## Expected behavior

The input cell is not changed.

## Context
- JupyterLab version: 3.3.0

krassowski:
## References

Fixes #12322

## Code changes

- immediately after submission the input box will:
   - become soft-invisible (opacity + height) to keep it hidden but focused
   - reject any key events (but will not be disabled to keep focus)
- the input box will be detached after 0.5s delay

> In the first commit I was trying to use `:has()` but gave up on this due to lack of FF support (and the support behind flag failing on this use case, reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1867893).

## User-facing changes

Users will be able to quickly send replies without the input box loosing focus and leading to the code editor being overwritten.

This is 99% solution - if the delay is over 0.5s it will fail.

None


jtpio (approved):
Thanks!

krassowski:

Issue #15293, opened 2023-10-23
Quickly jump to an open/recently closed notebook


mlucool:
### Problem

I often have many notebooks open and I mean to go back to one I was working on. I can't find it easily.

### Proposed Solution

Add search + recently closed tabs to open tabs. This should be accessible via hotkey.

I think the Chrome search tabs (ctrl+shift+A) is an ideal experience. Mostly because it's a keyboard shortcut where I can see in order the names of the last accessed tabs (notebooks in our case) then type to filter. I can click to jump to a tab or click x to close it. It even has recent tabs. Most of this is there in JupyterLab's open tabs already.

![image](https://github.com/jupyterlab/jupyterlab/assets/1813603/7638d034-09fc-479d-931b-539b545c21d0)

krassowski:
Linking to a recent discussion where we bounced some ideas on better tab navigation, and discussed limitations of the current "Open Tabs" implementation https://github.com/jupyterlab/jupyterlab/issues/15245#issuecomment-1773744541.

Having a popup/modal with tab list that can be filtered could solve the problem of user being required to navigate to the "Running" tab in the sidebar and of the difficulty finding the tab of interest.

mlucool:
> Having a popup/modal with tab list that can be filtered could solve the problem of user being required to navigate to the "Running" tab in the sidebar and of the difficulty finding the tab of interest.

+1

krassowski:
Some additional thoughts:
- the keyboard shortcut is a good UX but probably should not be the only approach
- placing a caret button on the tab bar (like in Chrome) may not work as well because there can be many tab bars
- to make it discoverable we could add:
    - (a) a "modal" button in "Running" sidebar,
    - (b) a new menu entry "Search open and recent tabs" in Tabs menu
- maybe there should be a "Recently closed" section in the "Running" sidebar
- a way to access the recently closed files is implemented in [jupyterlab-contrib/jupyterlab-recents](https://github.com/jupyterlab-contrib/jupyterlab-recents); we could create a service/token with history of recent files (and an action to clear it) in JupyterLab so that extensions can access this information and then extensions can expose it in UI as they wish (e.g. in menu as in jupyterlab-recents, or in a modal as proposed here)
- if we are storing the history of opened files, there will be privacy implications for different use ceases (some administrators will want it always on, some will want to disable it) and we should design an API to cater to these needs (basically this needs to be a separate plugin)
[...]

Issue #11613, opened 2021-12-03
Open recently closed notebook


mlucool:
### Problem

I sometimes close a notebook only to want to come back to it quickly (oops forgot to do X). Since I use a lot of folders, they are buried somewhere that takes a lot of navigating. It should be possible to reopen a notebook quickly. 

### Proposed Solution

Similar to browser's recently closed feature, we should add a recently closed feature to jupyter ([example](https://www.businessinsider.com/how-to-restore-google-chrome-tabs)). I don't think we need to implement a full history but the last N from the session (I think 10 is a good default).

davidbrochart:
Actually your workspace is restored when you relaunch JupyterLab, and it reopens the notebooks you had opened. Isn't it the behavior you are asking for?
Or maybe you are referring to the case where you close a notebook tab (not the whole JupyterLab window), and want to reopen it? In that case, if you didn't shut down the associated kernel, you can double click on the notebook in the Kernels list:

![image](https://user-images.githubusercontent.com/4711805/144682329-0e06f3dc-f5c0-490f-8e7d-094288a8ea00.png)

krassowski:
+1 on this. I think "Reopen Closed Tab" could be in the context menu of the main area tab, because:
- this is where "Close Tab", "Close All Other Tabs" and "Close Tabs to Right" are
- this is where it is in web browsers

Also, somehow related extension: https://github.com/jupyterlab-contrib/jupyterlab-recents/

Issue #5924, opened 2019-01-31
Any thoughts on an "Open recent" menu?


BoPeng:
I am used to the "Open recent" menu on so many applications (MS Office, VS Code, Atom, Keynote, mac vim, Adobe series of tools, etc) and just noticed that JLab does not have one. Does it make sense to add one?

jasongrout:
I think it can make a lot of sense. Should it be stored as an attribute of the workspace? Keep in mind that if I launch jlab in a different directory, all the paths are different, so the list becomes useless in that case.

BoPeng:
I think most applications store full path and displays only filename. MS Word shows full paths when there are ambiguities.

jasongrout:
My point is that those full path names are relative to the server root, so they wouldn't mean anything if the server was started in a different directory (and hence had a different root).

I do think it would be great to have this when starting a server always from the same directory.

BoPeng:
You are right, and it can be irritating to select a file and only find it missing. So the bottom line is that only files that exist with respect to the server root should be displayed.

jasongrout:
Yes, so perhaps before displaying the list, the extension providing this menu needs to query the server to make sure the file exists. This also would handle the case where a file was deleted, etc.

dhirschfeld:
Could you not also store the server root and then clear the list if it's different?

Or patch the paths with the difference between the saved/new roots!?

timhoffm:
There's a plugin available: https://github.com/tslaton/jupyterlab-recents

PR #15483, opened 2023-12-04
Reopen recently opened/closed files and modal navigation

krassowski:
## References

- Closes https://github.com/jupyterlab/jupyterlab/issues/5924
- Closes https://github.com/jupyterlab/jupyterlab/issues/15293
- Closes https://github.com/jupyterlab/jupyterlab/issues/11613
- Upstream discussion https://github.com/jupyterlab-contrib/jupyterlab-recents/issues/10

Punchlist
- [x] galata tests
- [x] add an icon [↗] to sidebar

## Code changes

Compared to `jupyterlab-recents` extension:

<details>

- remove polling for validation of paths; this was generating many background requests and was deemed undesirable overhead if we were to increase the number of requests by a factor of 10 (even though these were lightweight GET requests)
- do not send a get request to fetch the model/content type of the document for the that the user switch to - we already have it available at hand now thanks to integration directly into docmanager (this is not public/not emitted by the signal which is why the extension had to have this sub-optimal implementation)
- path validation is now conducted when opening the sub-menu, and when clicking on the file, limiting any overhead to the users of the feature
- when an item is discovered to be invalid after submenu was opened it will not be removed but greyed out to avoid jitter and user accidentally clicking a wrong item
- split recents into "recently opened" and "recently closed"
- use `IFileBrowserCommands` to explicitly depend on the availability of the filebrowser to open directories (and to fallback to document manager command if filebrowser does not provide such token

Because rather than adding a new package the extension gets integrated on the lower level by adding separate plugins to relevant existing packages (document manager, main menu) merging git history did no make much sense; instead the autorship was preserved by using `Co-authored-by` feature.

</details>

### Open Recent submenu

image

### Recently closed in the sidebar

image

Also seen is the [↗] button to open modal

### Reopen modal

reopen-widget-2

None.

> File browsers are encouraged to implement `IFileBrowserCommands` which was previously a no-op; if they don't then the recents menu will not show the directories.


krassowski:
Any suggestions on alternative shortcuts we could use?

> A part of me wishes that we had two sets of shortcuts: one for JupyterLab Desktop and one for when running in a standard browser.

krassowski:
For now switching to <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>A</kbd> as it is not used by major browsers/operating systems as per https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts

krassowski:
bot please update snapshots

krassowski:

packages/docmanager-extension/schema/plugin.json
```diff
@@ -149,6 +149,13 @@
       "title": "Rename Untitled File On First Save",
       "description": "Whether to prompt to rename untitled file on first manual save.",
       "default": true
+    },
+    "maximalRecents": {
```
jtpio, line 153: Another suggestion for the name of this setting:

```suggestion
    "maxNumberRecents": {
```

fcollonval (changes requested):
Thanks a lot for this @krassowski 

I have some minor suggestions.

packages/docmanager/src/recents.ts
```diff
@@ -0,0 +1,309 @@
...
+    );
+    return invalidPathsOrNulls.filter(x => typeof x === 'string') as string[];
+  }
+  /**
+   * Save the recent items to the state.
+   */
+  protected saveRecents(): void {
+    clearTimeout(this._saveRoutine);
+    // Save _recents 500 ms after the last time saveRecents has been called
+    this._saveRoutine = setTimeout(async () => {
```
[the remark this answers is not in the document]
krassowski: Good catch, done in 63c3e30fa79d0d327267dd164cb47fb8ea2f811d

packages/docmanager/src/recents.ts
```diff
@@ -0,0 +1,296 @@
...
+import { IStateDB } from '@jupyterlab/statedb';
+import { IDisposable } from '@lumino/disposable';
+import { Debouncer } from '@lumino/polling';
+import { ISignal, Signal } from '@lumino/signaling';
+import { IRecentsManager, RecentDocument } from './tokens';
+
+/**
+ * Manager for recently opened and closed documents.
+ */
+export class RecentsManager implements IRecentsManager, IDisposable {
```
fcollonval, line 17: I guess this is not needed any longer, isn't it?

```suggestion
export class RecentsManager implements IRecentsManager {
```

fcollonval (approved):
Thanks @krassowski 

I have an optional suggestion.

packages/docmanager/src/recents.ts
```diff
@@ -0,0 +1,296 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { PageConfig } from '@jupyterlab/coreutils';
+import { Contents, ServerConnection } from '@jupyterlab/services';
+import { IStateDB } from '@jupyterlab/statedb';
+import { IDisposable } from '@lumino/disposable';
```
krassowski, line 9:
```suggestion
```

krassowski:
Snapshots will need updating after merging. I think one of the docs snapshots needs some work in general.

krassowski:
but please update documentation snapshots

Issue #15273, opened 2023-10-17
[Debugger] float variable equal to math.inf displayed as NaN


pierre-haessig:
## Description

In JupyterLab Debugger, when debugging a code with a `float` variable having infinity value (i.e. from `math.inf`, `numpy.inf` or `float('inf')`), the variable is displayed as `NaN` in the Variables panel, see screenshot:

![screenshot JLab_debugger_inf_NaN](https://github.com/jupyterlab/jupyterlab/assets/1010404/f8319040-dba0-40d5-a71b-b5754a714983)

## Reproduce

Create new notebook in JupyterLab using a Python kernel ("classical" kernel, not xeus-python which I didn't tested.

Add cell with code:

```python
import math
a = math.inf
b = -a
print(a)
```

Run the cell and enable JupyterLab debugger → see the Variables panel.

## Expected behavior

Variables having a +inf or -inf value should be displayed as such.

## Context

- Operating System and version: Linux Ubuntu 22.04
- Browser and version: tested with
  - Firefox 118.0.2 (64 bits)
  - Chromium Version 118.0.5993.70 (Build officiel) snap (64 bits)
- JupyterLab version: 4.0.6

andrewfulton9:
I was able to reproduce this as well.

pierre-haessig:
is it easy to fix? I have no experience with JLab's codebase. Is it a problem in Javascript?

ashna1jain:
Hey can someone assign this issue to me? I have fixed this.

PR #15487, opened 2023-12-04
Fix display of float variables with value of infinity in the debugger tree

ashna1jain:
## References

This PR fixes [#15273](https://github.com/jupyterlab/jupyterlab/issues/15273)

## Code changes

Return value without checking isNaN when the type is float and value is "inf" or "-inf"

## User-facing changes
Now user will be able to see inf instead of NaN

Screenshot 2023-12-04 at 11 20 13 PM

None


ashna1jain:
@meeseeksdev tag [foo](pkg:debugger)[pkg:debugger](https://github.com/jupyterlab/jupyterlab/labels/pkg%3Adebugger) [bug](https://github.com/jupyterlab/jupyterlab/labels/bug)

krassowski (approved):
Thank you @ashna1jain!

PR #15504, opened 2023-12-08
Clean up eslint scripts in the translation packages

jtpio:
## References

The `translation` and `translation-extension` packages still had some `eslint` and `eslint:check` scripts in their `package.json`.

Linting is handled at the top-level, so we can remove these scripts.

## Code changes

Script cleanup in:

- `translation`
- `translation-extension`

## User-facing changes

None

None


krassowski (approved):
Makes sense, thank you @jtpio!

PR #15505, opened 2023-12-08
Disable Scan mode notification for screen reader users

j264415:
## References

15303- [Accessibility - Keyboard Navigation using a Screen Reader](https://github.com/jupyterlab/jupyterlab/issues/15303#top) 

## Code changes

New notification message added in the `jupyterlab/packages/apputils-extension/src/announcements.ts` file .
Also added `aria-live="assertive"` and `role="alert" ` to ensure the screen-reader announces the text first thing as you load Jupyterlab.

## User-facing changes
Notification popups to let user know which command to use to turn off scan mode when using a screen reader to ensure effective interactions with cells and files in notebook (check issue referenced above). 

None

### DIFF
--- a/packages/apputils-extension/src/announcements.ts
+++ b/packages/apputils-extension/src/announcements.ts
@@ -153,6 +153,16 @@ export const announcements: JupyterFrontEndPlugin<void> = {
             ]
           }
         );
+        const alertId = Notification.emit(
+          trans.__(
+            'For best user experience when using screenreader, turn off scan mode. Use the command: "Modifier Key + Spacebar." For ORCA use: "Insert+A"'
+          ),
+          'default',
+          {
+            autoClose: false
+          }
+        );
+        return alertId;
       } else {
         await fetchNews();
       }
--- a/packages/apputils-extension/src/notificationplugin.tsx
+++ b/packages/apputils-extension/src/notificationplugin.tsx
@@ -871,7 +871,7 @@ namespace Private {
         : message;
     return (
       <>
-        <div className="jp-toast-message">
+        <div className="jp-toast-message" role="alert" aria-live="assertive">
           {shortenMessage.split('\n').map((part, index) => (
             <React.Fragment key={`part-${index}`}>
               {index > 0 ? <br /> : null}


packages/apputils-extension/src/announcements.ts
```diff
@@ -153,6 +153,16 @@ export const announcements: JupyterFrontEndPlugin<void> = {
...
+        const alertId = Notification.emit(
+          trans.__(
+            'For best user experience when using screenreader, turn off scan mode. Use the command: "Modifier Key + Spacebar." For ORCA use: "Insert+A"'
+          ),
+          'default',
+          {
+            autoClose: false
+          }
+        );
+        return alertId;
```
krassowski, line 165: I do not think that we should display this message to non-screenreader users, they would be (rightly) annoyed. I do not see logic making this conditional on screen reader use, so I assume this was intended to show to everyone or this PR is still work in progress? If it is the latter, I would suggest switching it to draft status to help reviewers prioritise the PRs which are ready.

krassowski (changes requested):

packages/apputils-extension/src/announcements.ts
j264415: Hi @krassowski I have made some changes to the problem you mentioned above. 

- Firstly, the changes I have made should only notify the screen reader users that they should disable scan mode to make Jupyterlab screen reader friendly. 
- Aria live is set to` role: "assertive"` because that seemed to be the only way the screen reader would pick it up. 
- Furthermore, I have also placed the aria live append in the `announcement.ts` file as the scan mode information we are trying to convey is an announcement in itself.

j264415:
Hey @krassowski, do you mind reviewing this PR as i have made some changes from the comment you made.

packages/application/src/shell.ts
```diff
@@ -412,6 +412,13 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     hboxPanel.addWidget(vsplitPanel);
     hboxPanel.addWidget(rightHandler.sideBar);
 
+    const scanModeOffRegion = document.createElement('div');
+    scanModeOffRegion.setAttribute('aria-live', 'assertive');
+    scanModeOffRegion.setAttribute('role', 'region');
+    scanModeOffRegion.setAttribute('id', 'screen-reader-aria-live');
+    scanModeOffRegion.className = 'jp-ScanModepanel-ariaLive';
+    hboxPanel.node.appendChild(scanModeOffRegion);
```
krassowski, line 420: This is appending a child to a managed layout outside of the supported `addWidget` method; it might have side-effects, and the node might be lost on refresh. Does it need to be added to hboxPanel? Can it be added directly to body?

packages/apputils-extension/src/announcements.ts
```diff
@@ -153,6 +153,14 @@ export const announcements: JupyterFrontEndPlugin<void> = {
             ]
           }
         );
+        let scanModeOffRegion = document.getElementById(
+          'screen-reader-aria-live'
+        );
+        scanModeOffRegion!.append(
```
krassowski, line 159: Why is it only shown when `mustFetchNews` is set to none? Is it the right place to add this message?

packages/application/src/shell.ts
j264415: We have followed how skip links is rendered (as a widget) and implemented scan mode off that same way.

packages/apputils-extension/src/announcements.ts
j264415: We have placed it with notifications but moved it out of the `mustFetchNews === 'none' `

packages/application/src/shell.ts
```diff
@@ -2285,6 +2296,64 @@ namespace Private {
     }
   }
 
+  export class scanModeOffWidget extends Widget {
```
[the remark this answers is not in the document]
j264415: This has been fixed from the last commit .

Issue #15089, opened 2023-09-07
Broken benchmark check on CI


jtpio:
## Description

The "Benchmark Tests" CI check fails consistently when PRs are approved:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/8ded0d8c-72f4-45a0-b320-f5e238f38e77)

## Reproduce

Example run: https://github.com/jupyterlab/jupyterlab/actions/runs/6043794173/job/16401409899?pr=15042

## Expected behavior

The benchmark check should pass most of the time.

## Context

Do we still need to run the "Benchmark Tests" on PR approval? This triggers a 1h long CI job which adds CI fatigue on maintainers who might not be sure whether this check is important or not.

Also it looks like maintainers are now ignoring this check when merging PRs (since it's consistently failing), so this raises the question about it is really useful.

jtpio:
Looks like this may be fixable if the only issue is about publishing the report: 

```
Warning: "cml publish" is deprecated since "cml comment" now supports "![inline](./asset.png)"
```

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/590e8162-68e9-4c37-bcdc-7fe1deeae5f6)

krassowski:
We could just use GitHub actions instead of cml, as in https://github.com/jupyterlab/benchmarks/pull/144. That supports markdown so if we generate a PNG we can bease64 encode it.

PR #15523, opened 2023-12-12
Manually trigger benchmark tests instead of running them on `pull_request_review`

jtpio:
## References

Hoping to fix https://github.com/jupyterlab/jupyterlab/issues/15089

And progressively get CI to pass again consistently.

The benchmarks checks have been failing for many months on CI.

These jobs are run on PR approval, and may take a very long time to complete. Which means that in theory a maintainer would have to wait again for a new job to complete (and succeed) before hitting the merge button. Overall this adds a lot of maintenance fatigue and makes it difficult to merge PRs. To the point that it seems that it's just been ignored for the past months?

## Code changes

Run the benchmark on `schedule` instead of `pull_request_review` to get these runs out of the review workflow.

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,13 +1,13 @@
 name: Benchmark Tests
 
 on:
-  schedule:
-    - cron: '12 1 * * 0'
+  issue_comment:
+    types: [created, edited]
 
 jobs:
   test:
     name: Execute benchmark tests
-    if: github.event_name == 'schedule' || (github.event_name == 'pull_request_review' && (github.event.review.state == 'approved' || contains(github.event.review.body, 'please run benchmark')))
+    if: ${{ github.event.issue.pull_request && (contains(github.event.comment.body, 'please run benchmarks') }}
 
     runs-on: ubuntu-22.04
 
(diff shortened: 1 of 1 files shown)


krassowski:
**> To the point that it seems that it's just been ignored for the past months?

More like only looked on manually for changes which could introduce memory leaks.

Though this does not fix the underlying issue.

jtpio:
Should we trigger these benchmarks when the special "please run benchmarks" sentence is used then? (like for updating Galata snapshots).

This would already help avoid running these workflows by default, and only when they are needed.

krassowski:
Maybe depending on label would be better? Currently benchmarks cover mostly the `pkg:notebook`, maybe a few select others?

krassowski (approved):

jtpio:
> Maybe depending on label would be better? Currently benchmarks cover mostly the `pkg:notebook`, maybe a few select others?

Yes I guess that can make sense. Although since these benchmarks take a while a run, maybe a maintainer would be interested in running them only at different points in time, not necessarily for each commit?

jtpio:

Merging this PR as an incremental improvement towards #15089. We can look into fixing the benchmark workflow and add more triggers as part of #15089.

Issue #15503, opened 2023-12-08
Revert the `jlpm` update for JupyterLab 4.1?


jtpio:
https://github.com/jupyterlab/jupyterlab/pull/15295 updated the `yarn`  used by `jlpm` to version 3.6.4.

https://github.com/jupyterlab/jupyterlab/pull/15295 also includes some discussions on how this affects extension repos and other repos using the `jlpm` command.

For example the release of JupyterLab `4.1.0a4` required updating the `yarn.lock` in the Notebook repo, as the Notebook CI started to fail shortly after the release: https://github.com/jupyter/notebook/pull/7170

This is expected to be the same for most extension repos using `jlpm`.

The question is: should we revert this change in JuyterLab 4.1 to avoid disrupting third-party extensions workflows too much?

Technically the fix is not too difficult and was documented in https://github.com/jupyter/notebook/pull/7170. But it's still an annoyance to extension authors and if we can avoid it it may be better.

cc @krassowski @fcollonval @bollwyvl since you commented in https://github.com/jupyter/notebook/pull/7170.

bollwyvl:
Maybe there is a way to put something in `@jupyterlab/builder` that would transparently heal these kinds of things?

jtpio:
> Maybe there is a way to put something in `@jupyterlab/builder` that would transparently heal these kinds of things?

Doing some automatic repairs would be nice, but not sure to see how this would prevent CI builds from failing? (as illustrated in the previous discussion)

Also `@jupyterlab/builder` might not necessarily be "in sync" with the `jupyterlab` Python package (which provides `jlpm`). For example an extension can depend on `@jupyterlab/builder@^4.0.9` with `4.0.9` in its lock, while the latest `jupyterlab` stable release will at some point 4.1.

krassowski:
Maybe we could remove pnp plugin from `jlpm` using https://github.com/yarnpkg/berry/issues/2384#issuecomment-772204038 which would remove the patch and hash issue altogether? Since all we can do is manipulate `jlpm` in 4.1, we could make it inject a preinstall script which removes the hash... but this assumes that downstream did not want to use it which is not something we would now.

ericsnekbytes:
During triage we decided to revert this change.

Triage Notes:

- We can print messages in the install script to inform extension devs
- Option: Revert the last yarn update...or ship both versions

PR #15526, opened 2023-12-13
Revert "Bump yarn to 3.6.4"

jtpio:
Reverts jupyterlab/jupyterlab#15295

Fixes https://github.com/jupyterlab/jupyterlab/issues/15503 as per https://github.com/jupyterlab/jupyterlab/issues/15503#issuecomment-1854333331

The main motivation for reverting is that this update does not bring enough advantages to extension authors to be worth it (no real speed gains).


krassowski (approved):
Thanks you @jtpio!

PR #15528, opened 2023-12-14
Ignore `ipython` in dependabot updates

jtpio:
## References

As noticed in https://github.com/jupyterlab/jupyterlab/pull/15476 dependabot sometimes bumps `ipython`, and the `ipython` version number is visible on the snapshots. This can cause a snapshot mismatch and make the UI test fail, requiring updating the snapshots.

image

image

As suggested by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/15476#discussion_r1421403465, one solution can be to ignore updating `ipython`.

## Code changes

Ignore `ipython` updates in `dependabot.yml`.

## User-facing changes

None

None

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,5 @@ updates:
       # Check for updates to GitHub Actions every month
       # Align with pre-commit configuration .pre-commit-config.yaml
       interval: "monthly"
+    ignore:
+    - dependency-name: "ipython"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -13,3 +13,7 @@ updates:
       # Check for updates to GitHub Actions every month
       # Align with pre-commit configuration .pre-commit-config.yaml
       interval: "monthly"
+    ignore:
+    # Prevent the UI tests from failing, as the version number of ipython is visible
+    # in some of the Galata snapshots
+    - dependency-name: "ipython"
```
krassowski, line 19:
```suggestion
    - dependency-name: "ipython"
    - update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```

This way we will still get updates on major versions.

fcollonval (approved):
Thanks @jtpio

Issue #15511, opened 2023-12-09
Should button `jp-ToolbarButtonComponent` styles be restored for 4.x?


krassowski:
Some extensions, such as `jupyterlab-favourites`, add buttons relying in upstream styles. These buttons are unstyled in 4.1.0a4 due to changes in #15021

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/21f1b860-28b1-47fe-b9b0-96bb85f13b11)

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/f75ae0a8-786c-4184-be75-5494d8ea6e82)

I think we should restore the old styles, even if we do not use them any more, for backward compatibility.

ericsnekbytes:
@brichet Would you like to have a look and create a PR?

Other notes:

- We can add information to extension migration notes
- We can consider adding a red border to all deprecated items in dev mode

brichet:
> @brichet Would you like to have a look and create a PR?

Sure, I'll do it.

> * We can add information to extension migration notes

Can you elaborate ?

jtpio:
> We can add information to extension migration notes

That would be to let extension authors know the styles are being deprecated. The extension migration guide would be a good place to do that.

brichet:
This button looks good on my side, without any change, because it get the class `.jp-Button`, and get [these rules](https://github.com/jupyterlab/jupyterlab/blob/7d3160d50d511343f262e7c7d287bb2df3bec99d/packages/ui-components/style/base.css#L51-L93) applied:

![image](https://github.com/jupyterlab/jupyterlab/assets/32258950/8967311f-3914-40b3-8370-2ce86f2447f4)

I can still restore the `.jp-ToolbarButtonComponent` rules, but I wonder why the button doesn't receive this class in your case @krassowski.

krassowski:
It looks good on your side because you are running v3.2.1 with my https://github.com/jupyterlab-contrib/jupyterlab-favorites/pull/21 which added `.jp-Button` ;) Earlier versions fail as on the screenshot. I would not have time to patch all extensions like this though, nor do we even know about existence of many internal extensions.

PR #15530, opened 2023-12-14
Restore some CSS rules for backward compatibility

brichet:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15511

Restores some CSS rules that has been removed in https://github.com/jupyterlab/jupyterlab/pull/15021.
These rules concerns `button` elements in toolbar, and are deprecated.

## Code changes

Add some deprecated CSS rules.

## User-facing changes

Restore an expected behavior for extensions.

None

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -75,6 +75,12 @@ See https://github.com/jupyterlab/team-compass/issues/143 for more context on th
       ...
     ].join('|');
 
+- Some CSS rules for ``button`` with the class ``.jp-ToolbarButtonComponent`` has been kept for backward compatibility.
+
+  These rules are now **deprecated** and will be removed in Jupyterlab 5.
+  The ``button`` elements in toolbars must be updated to ``jp-button``, from
+  `jupyter-ui-toolkit <https://github.com/jupyterlab-contrib/jupyter-ui-toolkit>`_.
+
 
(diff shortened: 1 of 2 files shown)


docs/source/extension/extension_migration.rst
```diff
@@ -75,6 +75,12 @@ See https://github.com/jupyterlab/team-compass/issues/143 for more context on th
       ...
     ].join('|');
 
+- Some CSS rules for ``button`` with the class ``.jp-ToolbarButtonComponent`` has been kept for backward compatibility.
+
+  These rules are now **deprecated** and will be removed in Jupyterlab 5.
```
jtpio, line 80: Wondering if this removal would require a major release of JupyterLab (in that case JupyterLab 5)?

Or just a major bump of the JS package where these styles are defined (in that `@jupyterlab/ui-components`).

jtpio: At some point in the past, there were some discussions about versioning JS packages a bit more independently, without requiring a major lab release.

jtpio: Probably it's fine to leave it as is for now (mentioning JupyterLab 5). We can also discuss that separately. This just came to mind while looking at this addition to the extension migration guide.

jtpio (approved):
Thanks!

Kicking the UI tests CI check to make sure this does not introduce any visual change, otherwise looks good!

PR #15533, opened 2023-12-14
Add virtual scrollbar component to windowed lists.

### LINKED ISSUE #13343: Scrollbar jumps back in a virtualized notebook on scrolling

afshin:
This PR updates the `WindowedList` class that the notebook widget inherits from to have customizable renderers (which allows for adding `aria` labels to make windowed lists more accessible and allows for customization of the UI).

The renderer in turn is used to add a scrollbar component to replace the native scrollbar that jumps when virtual notebook rendering is enabled.

The scrollbar (and its toolbar button) only appear when "Windowing mode" is set to "full".

## References
- supersedes #15109 
- supersedes #15036
- fixes #13343

## Code changes
- Updates the `WindowedList` API to allow for custom renderers of windowed list subcomponents
- Adds a virtual scrollbar UI to replace the native scrollbar when virtual rendering is turned on
- Updates the command toolbar button to respect `isEnabled`, `isVisible`, `isToggled`.

## User-facing changes
Adds an optional visual scrollbar component for "full" windowing mode notebooks.

https://github.com/jupyterlab/jupyterlab/assets/159529/be8f9288-ed89-432b-ba4d-7e977f7ad3ab

`.jp-WindowedPanel-window` is now called `.jp-WindowedPanel-viewport`


afshin:
@krassowski I wasn't sure how to do the git shenanigans you described, so I've made a new branch and a clean PR. I hope this is okay!

afshin:
cc: @fcollonval @GabrielaVives

krassowski (changes requested):
Thank you! There are two blocking regression left:
- a) scroll past end is no longer working
- b) the logic for enabling the scrollbar seems flowed; it auto-enables in our visual tests when windowing mode is set to full, making them fail because native scrolling does not work; I would expect it to only show up when explicitly enabled by the user

The remaining comments are the bugs in the new feature rather than regressions so I do not want to block on them as we can iterate in beta/RC cycle.

### Snapshots

Can you please revert all snapshot updates by running:

```
git checkout 98d1e06796906ce622559146ed127075a7a46f71 galata/test/*/*.ts-snapshots/
```

and committing changes?

This way:
- we do not inflate the repository size unnecessarily
- we know the snapshot changes are due to random encoding differences (which I believe is the case) and not some tiny shifts in layout
- review becomes easier as I do not have to analyse one-pixel or smaller differences on the snapshots each time

I wold suggest not calling the bot for snapshot updates for now.

 I did this on a sub-branch and while on this branch there are 19+1 failures, my branch has 18+0 failures which suggests that the snapshot updates so far are not relevant. Instead they are all due to the faulty logic for toggling the scrollbar.

### UX

I really miss an indication on where my viewport is on the scrollbar and will probably not use this this feature until it shows the contour of the visible cells but this can be addressed in the future.

packages/notebook/src/widget.ts
```diff
@@ -193,16 +193,19 @@ export class StaticNotebook extends WindowedList {
...
           StaticNotebook.defaultNotebookConfig.overscanCount,
-        windowingActive:
-          (options.notebookConfig?.windowingMode ??
-            StaticNotebook.defaultNotebookConfig.windowingMode) === 'full'
+        windowingActive
       }),
-      layout: new NotebookWindowedLayout()
+      layout: new NotebookWindowedLayout(),
+      renderer: options.renderer ?? WindowedList.defaultRenderer,
+      scrollbar: windowingActive
```
krassowski, line 208: I think we should make this a setting instead of forcing users of windowed mode to use this (or toggle of each time).

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1182,18 +1278,58 @@ export class WindowedList<
     }
   }
 
-  protected _viewModel: T;
-  private _innerElement: HTMLDivElement;
+  /**
+   * Render virtual scrollbar.
+   */
+  private _renderScrollbar(): void {
```
krassowski, line 1284: Note for future: there does not have a guard preventing regeneration when nothing changed, this means it is recreated on every cursor movement in resize.

krassowski:
I experimented a bit to address issues that I consider blocking highlighted in the review above: https://github.com/krassowski/jupyterlab/pull/80 - feel welcome to cherry-pick (for some reason I cannot open a PR against your branch). There are some improvements that could be made like only creating the resize observer if needed.

I think that a way forward would be to disable this feature by default (rather than enable it always when windowed mode is on) and then add a setting (which would only take effect in windowed mode is on). Once we have the setting, we could flip it on by default once we feel this is ready. Having a setting would also make it easy to disable it for specific UI tests.

packages/ui-components/src/components/windowedlist.ts
afshin: I started to tackle this problem by creating a scroll state variable that holds the list of `selected` items and the index of the `active` item. But the trouble is that part of the "state" of the scrollbar is out of our hands because the renderer that generates each item in the scrollbar might change arbitrarily, for example it could include a timestamp in it or something else that is nondeterministic.

So I think the only way to approach this would be to use a virtual DOM (whether React or vdom, etc.) and let that decision be made by the tree-diffing algorithm.

krassowski (approved):
Thank you @afshin, this looks good now! I pushed a few commits to fix the galata tests and update the extension migration guide.

I plan to merge this PR tomorrow on 20th Dec at 13:00 UTC (unless changes are requested before this time) and follow with immediate release of 4.1.0.beta0.

jtpio:
Congrats @afshin on getting this one merged, and thanks @krassowski for the review!

PR #15534, opened 2023-12-14
Contrast colour change

j264415:
## References
This implements the accent color changes (--jp-brand-color0 ) in this [PR](https://github.com/jupyterlab/jupyterlab/pull/9335#issuecomment-761288181) so it passes AAA normal text contrast accessibility 

## User-facing changes

**Before:**
Highlight Colour before

**After:**
Contrast Colour AAA
Colour palette AAA constrast


krassowski:
Why? We already did this once (https://github.com/jupyterlab/jupyterlab/pull/9335#issuecomment-761288181).

krassowski:
The current colours pass both AA and AAA checks:

image

A pull request like this should really come with a reference and an explanation.

j264415:
> The current colours pass both AA and AAA checks for large text and AA for normal text:
> 
> image
> 
> A pull request like this should really come with a reference and an explanation.

@krassowski Thanks for the quick review please see the added PR description. This is intended to fix the failing test for AAA normal text contrast

krassowski:
But _why_ do you want to make is pass WCAG 2 AAA here?

> Note 2: It is not recommended that Level AAA conformance be required as a general policy for entire sites because it is not possible to satisfy all Level AAA Success Criteria for some content.

https://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance.html#uc-levels-head

The current colours have APCA (draft WCAG 3) contrast of 76.9 which is a very good result.

If needs be I think it is fine to make contrast stronger, but this will often **require** changes to the UI design. If we just swapped everything for harder contrast the UI would become very distractive  and difficult to use.

As always a link to some discussion of the WCAG 2 contrast criteria being flawed:
- https://blog.datawrapper.de/color-contrast-check-data-vis-wcag-apca/
- https://www.bounteous.com/insights/2019/03/22/orange-you-accessible-mini-case-study-color-ratio
- https://github.com/w3c/wcag/issues/695

Issue #15497, opened 2023-12-06
Screen reader doesn't read all keyboard shortcut punctuation 


e218736:
### Problem

When navigating using a screen reader (Windows Narrator) some keyboard shortcuts that include punctuation are not read out correctly. 

For example, within keyboard shortcuts 
Activate Previous Tab Bar	<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>,</kbd> 

Screen reader annouces 'Activate Previous Tab Bar control plus shift plus' and ignores the comma

### Proposed Solution

Adding aria-labels and aria-keyshortcuts properties to announce specific key combinations properly

JasonWeill:
@e218736 Thank you for opening this! In which context do you encounter this issue? Navigating the pull-down menus, command palette, focusing toolbar buttons, or somewhere else?

e218736:
Hi @JasonWeill, 
whilst navigating through the top menu bar e.g. within Tabs, 3 of 5 keyboard shortcuts are not read out correctly.
Also when opening keyboard shortcuts from the help menu any shortcuts with punctuation are announced incorrectly.
Currently working on a fix for this with aria-labels and will check whether command palette & toolbar buttons should be included.

PR #15539, opened 2023-12-18
Keyboard shortcut screen reader

e218736:
## References
Issue 15497 - [Screen reader doesn't read all keyboard shortcut punctuation](https://github.com/jupyterlab/jupyterlab/issues/15497)

## Code changes
- Function added to render aria labels for punctuation key symbols within the keyboard shortcut dialog module
- Test added to assert punctuation keys include aria-label

## User-facing changes
none visible. 
For screen reader users, keyboard shortcuts that include punctuation will be announced by default when focused.

### DIFF
--- a/packages/shortcuts-extension/test/shortcuts.spec.ts
+++ b/packages/shortcuts-extension/test/shortcuts.spec.ts
@@ -165,3 +165,26 @@ describe('@jupyterlab/shortcut-extension', () => {
     });
   });
 });
+
+describe('shorcuts list', () => {
+  it('should add punctuation aria-label', () => {
+    const keyboardShortcuts = document.getElementsByClassName(
+      'jp-contextualShortcut-key'
+    );
+    for (let i = 0; i < keyboardShortcuts.length; i++) {
+      const keyboardLabelAria = keyboardShortcuts[i].getAttribute('aria-label');
+      const keyboardLabelText = keyboardShortcuts[i].innerHTML;
+      const punctuation = ",}{.'-";
+      if (punctuation.includes(keyboardLabelText)) {
+        expect(keyboardLabelAria).toEqual([
+          'Comma',
+          'Closing bracket',
+          'Opening bracket',
+          'Full stop',
+          'Single quote',
+          'Hyphen-minus'
+        ]);
+      }
+    }
+  });
+});
--- a/packages/shortcuts-extension/test/shortcuts.spec.ts
+++ b/packages/shortcuts-extension/test/shortcuts.spec.ts
@@ -7,6 +7,7 @@ import { IDataConnector } from '@jupyterlab/statedb';
 import { CommandRegistry } from '@lumino/commands';
 import { Platform } from '@lumino/domutils';
 import pluginSchema from '../schema/shortcuts.json';
+import expect from 'expect';
 
 describe('@jupyterlab/shortcut-extension', () => {
   const pluginId = 'test-plugin:settings';
--- a/packages/apputils-extension/src/shortcuts.tsx
+++ b/packages/apputils-extension/src/shortcuts.tsx
@@ -48,6 +48,20 @@ export function displayShortcuts(options: IOptions) {
   const { commands, trans, activeElement } = options;
   const elt = activeElement ?? document.activeElement;
 
+  function keyAriaLabel(ch: string) {
+    const keyToText: { [key: string]: string } = {
+      ']': 'Closing bracket',
+      '[': 'Opening bracket',
+      ',': 'Comma',
+      '.': 'Full stop',
+      "'": 'Single quote',
+      '-': 'Hyphen-minus'
+    };
+    if (Object.keys(keyToText).includes(ch)) {
+      return keyToText[ch];
+    }
+  }
+
   /**
    * Find the distance from the target node to the first matching node.
    *
@@ -66,7 +80,11 @@ export function displayShortcuts(options: IOptions) {
       const container: JSX.Element[] = [];
       key.split(' ').forEach((ch, chIndex) => {
         container.push(
-          <span className={SHORTCUT_KEY_CLASS} key={`ch-${chIndex}`}>
+          <span
+            className={SHORTCUT_KEY_CLASS}
+            key={`ch-${chIndex}`}
+            aria-label={keyAriaLabel(ch)}
+          >
             <kbd>{ch}</kbd>
           </span>,
           <React.Fragment key={`fragment-${chIndex}`}> + </React.Fragment>


packages/shortcuts-extension/src/components/ShortcutInput.tsx
```diff
@@ -424,6 +429,7 @@ export class ShortcutInput extends React.Component<
             : 'jp-mod-hidden'
         }
         onBlur={event => this.handleBlur(event)}
+        aria-label="keyboard shortcut Input Box"
```
krassowski, line 432: Needs translation

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -40,7 +51,7 @@ enum ShortCutLocation {
   Left,
   Right
 }
-
+let currentNode = 0;
```
krassowski, line 54: We do not want to keep state on global level as it will often lead to errors when components are reused. This will need to be handled in the react component.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -208,6 +218,26 @@ export class ShortcutItem extends React.Component<
     }, '');
   };
 
+  punctuationToText = (value: string): string => {
+    return value.split(' ').reduce((result, key) => {
+      if (key === ']') {
+        return (result + ' Closing bracket').trim();
```
krassowski, line 224: Strings in this function need to be translated

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -460,6 +502,165 @@ export class ShortcutItem extends React.Component<
     );
   }
 
+  //navHandler = (event: React.KeyboardEvent): void => {}
+
+  // Handle key down for row navigation
+
```
krassowski, line 508:
```suggestion
  /**
   * Handle key down for row navigation
   */
```

The jsdoc format helps with documentation

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -460,6 +502,165 @@ export class ShortcutItem extends React.Component<
     );
   }
 
+  //navHandler = (event: React.KeyboardEvent): void => {}
```
krassowski, line 505: To remove?

packages/shortcuts-extension/src/components/ShortcutList.tsx
```diff
@@ -36,10 +46,17 @@ export class ShortcutList extends React.Component<IShortcutListProps> {
         }}
         id="shortcutListContainer"
       >
-        <div className="jp-Shortcuts-ShortcutList">
+        <div
+          className="jp-Shortcuts-ShortcutList"
+          id="Shortcuts-ShortcutList"
```
krassowski, line 51: Please do not add an identifier here. There can be multiple shortcut lists in principle.

```suggestion
```

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -460,6 +502,165 @@ export class ShortcutItem extends React.Component<
...
+
+  // Handle key down for row navigation
+
+  handleRowKeyDown = (event: React.KeyboardEvent): void => {
+    const shortcutList = document.getElementById('Shortcuts-ShortcutList');
```
krassowski, line 510: This will not behave as intended if there is more than one shortcut list. I would strongly suggest rewriting the implementation so that the keydown event which should navigate is propagated up to `ShortcutList` component which would then perform focus switching.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -460,6 +502,165 @@ export class ShortcutItem extends React.Component<
...
+      if (nextFocused) {
+        focusable[focusedIndex]?.setAttribute('tabindex', '-1');
+        nextFocused?.setAttribute('tabindex', '0');
+        (nextFocused as HTMLButtonElement).focus();
+      }
+    }
+  }
+
+  handleEvent(event: React.KeyboardEvent): void {
+    let tabOrder = document.getElementById('tab-key-1-1') as HTMLLIElement;
```
krassowski, line 620: It is very unclear what this identifier refers to.

krassowski (changes requested):
Thank you for working on this! I left some suggestions by the code.

e218736:
Hi @krassowski,

I have implemented the changes you've requested, thanks.

For the potential of multiple shortcut lists the hard coded Id has been changed to an optional unique identifier, this seems to be the simplest way to enable selection and implement the vertical keyboard navigation within the key handler. Please let me know if these changes are better

packages/shortcuts-extension/test/shortcuts.spec.ts
```diff
@@ -313,3 +314,26 @@ describe('@jupyterlab/shortcut-extension', () => {
     });
   });
 });
+
+describe('shorcuts list @A11y', () => {
```
gabalafou, line 318:
```suggestion
describe('shortcuts list @A11y', () => {
```

gabalafou (changes requested):
Good catch on how the keyboard shortcuts with punctuation marks are not read out by a screen reader.

I've left some review comments.

My two big concerns with this PR are about using the tablist and tab roles, as well as the up and down keys not working when the user is focused on one of the buttons.

packages/shortcuts-extension/src/components/ShortcutList.tsx
```diff
@@ -23,10 +25,91 @@ export interface IShortcutListProps {
...
+      let currentNode = focusable.indexOf(activeElement);
+      let activeNode = focusable[currentNode] as HTMLElement;
+      let nextNode = focusable[currentNode + 1] as HTMLElement;
+      let previousNode = focusable[currentNode - 1] as HTMLElement;
+
+      if (event.key === 'ArrowDown') {
+        let nxtNode = nextNode;
+        if (nxtNode) {
+          nxtNode.setAttribute('tabindex', '0');
+          activeNode.setAttribute('tabindex', '-1');
```
gabalafou, line 79: This line throws an error when the function is called by ShortcutItem. That's because the active element in that case is not one of the direct children of the ShortcutList element, so `activeNode` ends up undefined.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -253,12 +285,14 @@ export class ShortcutItem extends React.Component<
   getResetShortCutLink(): JSX.Element {
```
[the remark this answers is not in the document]
krassowski: Worth checking if it is safe to rename this public method now. If this PR was targeting 5.0 then sure, otherwise we need to be careful and follow a deprecation cycle (i.e. rename method but keep the old one as a deprecated alias).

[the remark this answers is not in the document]
gabalafou: We should just follow the safe route: keep the old one and deprecate it.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -406,15 +447,17 @@ export class ShortcutItem extends React.Component<
   getAddLink(): JSX.Element {
     const trans = this.props.external.translator.load('jupyterlab');
     return (
-      <a
+      <button
         className={!this.state.displayNewInput ? 'jp-Shortcuts-Plus' : ''}
         onClick={() => {
           this.toggleInputNew(), this.props.clearConflicts();
         }}
         id="add-link"
```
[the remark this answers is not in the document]
gabalafou: Oh, good point. Maybe it's not worth changing the id then.

packages/shortcuts-extension/src/components/ShortcutList.tsx
e218736: Hi @gabalafou,

Just working through these comments and after removing things like arrow nav wrapping, extra key down handler etc. I'm unable to replicate this error

packages/shortcuts-extension/test/shortcuts.spec.ts
```diff
@@ -313,3 +314,26 @@ describe('@jupyterlab/shortcut-extension', () => {
...
+  it('should add punctuation aria-label', () => {
+    const keyboardShortcuts = document.getElementsByClassName(
+      'jp-contextualShortcut-key'
+    );
+    for (let i = 0; i < keyboardShortcuts.length; i++) {
+      const keyboardLabelAria = keyboardShortcuts[i].getAttribute('aria-label');
+      const keyboardLabelText = keyboardShortcuts[i].innerHTML;
+      const punctuation = ",}{.'-";
+      if (punctuation.includes(keyboardLabelText)) {
```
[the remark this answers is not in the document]
e218736: Hi @gabalafou, 
Thanks, I understand that the current test isn't working. However, with your approach I am unsure that I would be able to retrieve the HTML element needed to assert that the aria-label attribute is correct. Are you able to provide direction if this is possible?

galata/test/jupyterlab/settings.test.ts
```diff
@@ -248,3 +248,126 @@ test('Opening Keyboard Shortcuts settings does not mangle user shortcuts', async
...
+    const shortcutRows = page.locator('.jp-Shortcuts-Row');
+
+    for (let i = 0; i < (await shortcutRows.count()) - 1; i++) {
+      const shortcutListButton = shortcutRows.nth(i).locator('button').first();
+
+      await shortcutListButton.focus();
+
+      await page.keyboard.press('Escape');
+
+      expect(shortcutRows.nth(i)).toBeFocused();
```
krassowski, line 370: > Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator

packages/shortcuts-extension/src/components/ShortcutItem.tsx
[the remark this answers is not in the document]
krassowski: I will be opening a new PR which reworks the shortcuts UI fixing many bugs in it. I removed the ID in that (which was a bit more work).

krassowski:
FYI PR https://github.com/jupyterlab/jupyterlab/pull/16043 clears up the implementation of settings UI; there should not be many conflicts with this PR (I waited with starting that work hoping that this will be ready and merged by then, but I could not wait any more), but some things to note:
- there will be no more "left" and "right" shortcut as more than two shortcuts will be now supported
- `#or` ID got deleted and only a class name is used now; and styles which were using `display: none`/`display: block` were replaced with `opacity` (to prevent row jitter when hovering over shortcuts)

I can take care of rebasing this PR once that one is merged (or vice versa).

packages/settingeditor/test/settingsformeditor.spec.tsx
```diff
@@ -114,10 +123,16 @@ describe('@jupyterlab/settingeditor', () => {
...
+      expect(form.props.formData).toEqual({
+        keyA: 'A',
+        keyC: 'C',
+        keyComma: ','
+      });
       expect(Object.keys(form.props.schema.properties)).toEqual(
         Object.keys(form.props.formData)
       );
+      //check AriaLabels are
+      expect(ariaLabels).toEqual(form.props.formData.ariaLabel);
```
krassowski, line 135: I do not follow this test. It appears that you are testing the generic settings editor, but you did not make any changes to source code in `settingeditor/src`.

If you wanted to test `shortcuts-extension` components I would suggest going the same route as I went in https://github.com/jupyterlab/jupyterlab/pull/16043/files#diff-ef23219df9ce6e7419379a5189c9252aac220bdf696d407f45d99a0c33ab453e

krassowski: This is creating a new e.g. `shortcuts-extension/test/components/ShortcutInput.spec.ts` file mirroring the shortcuts-extension/test/components/ShortcutUI.spec.ts` file from my PR. Of note it might be best deferred until that PR is merged as it simplifies testing individual components by reducing the external bundle content.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -26,6 +36,8 @@ export interface IShortcutItemProps {
   clearConflicts: Function;
   contextMenu: Function;
   external: IShortcutUIexternal;
+  tabIndex: number;
+  handleRowKeyDown: Function;
```
krassowski, line 40: `Function` will be disallowed going forward, see https://github.com/jupyterlab/jupyterlab/issues/16038#issuecomment-2015398728.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -469,7 +605,13 @@ export class ShortcutItem extends React.Component<
     } else {
       return (
         <div
+          title={this.props.shortcut.commandName}
```
krassowski, line 608:
```suggestion
          title={this.props.shortcut.label}
```


`commandName` is something like "my-package:my-command". Label is the thing which is user-readable.

packages/shortcuts-extension/src/components/ShortcutList.tsx
```diff
@@ -23,10 +25,86 @@ export interface IShortcutListProps {
   height: number;
   contextMenu: Function;
   external: IShortcutUIexternal;
+  id?: string;
```
krassowski, line 28: I would consider renaming this to "index" to avoid confusion with DOM element's `id` property.

krassowski: Oh, no this is actually used as an ID. Please disregard.

packages/shortcuts-extension/test/shortcuts.spec.ts
```diff
@@ -309,7 +309,154 @@ describe('@jupyterlab/shortcut-extension', () => {
...
+        type: 'object'
+      },
+      version: 'test'
+    };
+
+    const connector: IDataConnector<
+      ISettingRegistry.IPlugin,
+      string,
+      string,
+      string
```
krassowski, line 346:
```suggestion
      string
```

The last two are not needed, see changes in https://github.com/jupyterlab/jupyterlab/pull/16043

packages/shortcuts-extension/src/components/ShortcutItem.tsx
e218736: Thanks for letting us know, to be able to refactor is dependent on your PR being merged. As mentioned previously, without being able to work on this past 28/03/24 would you be happy to take care of the refactor once merged?

galata/test/jupyterlab/settings.test.ts
```diff
@@ -248,3 +248,126 @@ test('Opening Keyboard Shortcuts settings does not mangle user shortcuts', async
...
+  });
+
+  test('Should focus shortcuts container first row using tab key', async ({
+    page
+  }) => {
+    const shorcutRow = page.locator('.jp-Shortcuts-Row').first();
+    await expect(shorcutRow).toBeFocused();
+  });
+
+  test('Should retain tab order by focusing property inspector using tab key', async ({
```
krassowski, line 285: Why is there a test for property inspector?

packages/shortcuts-extension/src/components/ShortcutList.tsx
```diff
@@ -36,10 +114,11 @@ export class ShortcutList extends React.Component<IShortcutListProps> {
         }}
         id="shortcutListContainer"
       >
-        <div className="jp-Shortcuts-ShortcutList">
-          {this.props.shortcuts.map((shortcut: ShortcutObject) => {
+        <div className="jp-Shortcuts-ShortcutList" id={this.props.id}>
+          {this.props.shortcuts.map((shortcut: ShortcutObject, index) => {
             return (
               <ShortcutItem
+                tabIndex={index === 0 ? 0 : -1}
```
krassowski, line 121: Why is only first shortcut focusable?

packages/shortcuts-extension/src/components/ShortcutUI.tsx
```diff
@@ -530,6 +531,7 @@ export class ShortcutUI extends React.Component<
           height={this.props.height}
           contextMenu={this.contextMenu}
           external={this.props.external}
+          id={DOMUtils.createDomID() + '-shortcut-list'}
```
krassowski, line 534: In React if you want to get an element we would use refs instead.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -358,9 +386,14 @@ export class ShortcutItem extends React.Component<
     return this.props.shortcut.keys[key].map(
       (keyBinding: string, index: number) => (
         <div className="jp-Shortcuts-ShortcutKeysContainer" key={index}>
-          <div className="jp-Shortcuts-ShortcutKeys">
+          <button
```
krassowski, line 389: Why is this a button? This element corresponds to individual key in a sequence, it has no button role.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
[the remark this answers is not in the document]
krassowski: Note: we bumped package version for this shortcuts-extension, safe to rename.

packages/shortcuts-extension/src/components/ShortcutInput.tsx
```diff
@@ -162,6 +162,10 @@ export class ShortcutInput extends React.Component<
         keys: keys,
         currentChain: currentChain
       });
+    } else if (event.key === 'Escape') {
+      const focusedElement = document.activeElement;
+      const parentRow = focusedElement?.closest('.jp-Shortcuts-Row');
+      (parentRow as HTMLDivElement).focus();
```
krassowski, line 168: This will error out if active element is outside of the of table. It should not usually happen but any casting with `as` is unsafe because it is an implicit non-null assertion.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -225,9 +255,13 @@ export class ShortcutItem extends React.Component<
...
       >
-        <div className="jp-Shortcuts-ShortcutKeys">
+        <button
+          className="jp-Shortcuts-ShortcutKeys"
+          aria-label={this.punctuationToText(keyboardKey)}
+          onKeyDown={this.handleKeyDown}
+        >
           {this.toSymbols(keyboardKey)}
-        </div>
+        </button>
```
krassowski, line 264: Why are individual keys `button`s?

galata/test/jupyterlab/settings.test.ts
krassowski: Ah it is meant to tab down to a row with property inspector.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -315,6 +351,106 @@ export class ShortcutItem extends React.Component<
...
+        focusedElement?.setAttribute('tabindex', '-1');
+        const parentRow = focusedElement?.closest('.jp-Shortcuts-Row');
+        (parentRow as HTMLDivElement).focus();
+      } else if (
+        event.key === 'ArrowDown' &&
+        evTarget.className !== 'jp-Shortcuts-Reset'
+      ) {
+        if (rowBelow !== null) {
+          const rowBelowShortcuts = rowBelow!.querySelectorAll('button');
+          const rowBelowFocusable: Element[] = [...rowBelowShortcuts];
```
krassowski, line 398: This line does not make much sense. It appears to be an attempt to workaround TypeScript guards but also adds runtime overhead.

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -315,6 +351,106 @@ export class ShortcutItem extends React.Component<
     );
   }
 
+  // handle key down function to navigate within each row
+  handleKeyDown(event: React.KeyboardEvent): void {
```
krassowski, line 355: This entire function needs to be refactored:
- `tabindex` manipulations should be explained
- typing should be fixed
- repetitive code should be minimalised

krassowski:
I'm done fixing merge conflicts and correcting some of the issues. There are more issues. Is there anyone interested in finishing this off? If not we probably need to move it to the next milestone.

krassowski:
Alternatively, it could be split up in two parts: punctuation and navigation. The punctuation part seems to be ready so we can merge it for 4.2.

gabalafou:
Update: I'm currently working on splitting this PR into two parts:

1. For [4.3.0 milestone](https://github.com/jupyterlab/jupyterlab/milestone/78) - the mapping of punctuation symbols to names
2. For future milestone - Keyboard navigation of the shortcut editor UI

gabalafou:
Update: I tried pulling out the punctuation-to-text part of this PR because that seemed nearly ready to ship. But then, as I worked on it, I realized that this is kind of reinventing the wheel. Screen readers already allow you to step into a word or a sentence to read it character by character. It does mean that the screen reader user has to take the extra step of stepping into the keyboard shortcut, but I would imagine that such users are used to having to do this when consulting keyboard shortcuts. (This would be useful to verify with blind or low-vision folks who rely on screen readers.)

One benefit of relying on the screen reader software for this feature is that we do not have to add translations for any of the punctuation symbols. And we have less code to maintain. 

In short, this punctuation-to-text feature seems to me more like a nice-to-have, and as such I'm not going to try to rush it for the 4.3.0 milestone.

PR #15548, opened 2023-12-20
Link to paths starting at Unix-style root in the tracebacks

### LINKED ISSUE #15513: Tracebacks do not link to files in `/`
## Description

https://github.com/jupyterlab/jupyterlab/pull/13390 introduced support for opening files from tracebacks. It works well for files in the user home (e.g. `~`) but it fails to link to files in Unix-like file systems this is any path starting with `/`

## Reproduce

Create a test file in `/tmp/test`:

```
mkdir -p /tmp/test && echo 'def div(a, b): return a / b' > /tmp/test/div.py
```

Run jupyterlab from some place else

```
PYTHONPATH=$PYTHONPATH:/tmp/test jupyter lab
```

Or even in the `/tmp` to make sure this is not due to prefix mismatch:

```
PYTHONPATH=$PYTHONPATH:/tmp/test jupyter lab --ServerApp.root_dir /tmp/
```

Create a notebook with a code cell `import div; div.div(1, 0)`

Run and see no link:

image

## Expected behavior

`/tmp/test/div.py:1` should be linkified

## Context

- JupyterLab version: 4.0.0a4

krassowski:
## References

Fixes #15513

## Code changes

- Adds an optional `allowRoot` argument `isLocal` methods/functions.
- Sets this argument to `true` for `handlePathAnchor` used exclusively in traceback renderer

## User-facing changes

Paths to files in Unix-style root in  traceback are linkified:

| Before | After |
|--|--|
| image | image |

None

### DIFF
--- a/packages/coreutils/src/url.ts
+++ b/packages/coreutils/src/url.ts
@@ -141,16 +141,18 @@ export namespace URLExt {
   /**
    * Test whether the url is a local url.
    *
+   * @param allowRoot - Whether the paths starting at Unix-style filesystem root (`/`) are permitted.
+   *
    * #### Notes
    * This function returns `false` for any fully qualified url, including
    * `data:`, `file:`, and `//` protocol URLs.
    */
-  export function isLocal(url: string): boolean {
+  export function isLocal(url: string, allowRoot: boolean = false): boolean {
     const { protocol } = parse(url);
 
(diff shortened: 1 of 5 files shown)


jtpio (approved):
Thanks!

Issue #15509, opened 2023-12-09
Inconsistent toolbar button positions when rendered in notebook


firezym:
# Description
The Debugger Button is "jumping around" among different notebooks in the Notebook Panel Toolbar epecially after refreshing, as shown in below gif demo:

# Reproduce
![gif-pics](https://github.com/jupyterlab/jupyterlab/assets/3654421/e768abf9-1cdd-4e3d-8d67-5fffc25e3a64)

# Expected behavior
User should see a stable button order on the Notebook Toolbar.

# Context
os: wsl2 linux system, using jupyterlab 4.0.9

jupyter_client 8.6.0
jupyter_core 5.5.0
jupyter-events 0.9.0
jupyter-lsp 2.2.1
jupyter_server 2.12.1
jupyter_server_terminals 0.4.4
jupyterlab 4.0.9
jupyterlab_celltagsclasses 0.5.1
jupyterlab_gridwidth 0.2.4
jupyterlab_pygments 0.3.0
jupyterlab_server 2.25.2

krassowski:
Thank you for reporting this! Can you try jupyterlab pre-release [4.1.0a4](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.1.0a4) in which the logic was refactored?

firezym:
I conda created a new env for 4.1.0a4, the previous "jump around" issue reduced a lot. But there is `still a small chance of dis-organized order of icons after refreshing`, shown at the end of the following gif ( I cut out about 60s of refreshing back and forth more than 20 times in the middle of the screen recording).

Another issue is shown at the the begining of the gif demo, which is very frequently observed: `the debugger button turn gray and disabled often after refreshing and stay gray until next refresh`.

![gif-pics](https://github.com/jupyterlab/jupyterlab/assets/3654421/58e43656-ad5e-4230-b1c1-64c6a885efdb)

krassowski:
The refactor of toolbars happened in https://github.com/jupyterlab/jupyterlab/pull/15021 and I can confirm that it introduced a regression with toolbar items being hidden or placed randomly on first load. It feels that we should defer the position shuffling and collapsing until the layout has settled. CC @brichet

An issue which _sounds_ similar (but is largely unrelated in the codebase) was brought up by @gabalafou in context of the menu hamburger implementation (see https://github.com/jupyterlab/jupyterlab/issues/13148#issuecomment-1858374638).

[...]

PR #15553, opened 2023-12-21
Fixes the order of items in notebook toolbar

brichet:
https://github.com/jupyterlab/jupyterlab/pull/15021 changes some logic in the reactive toolbar (the one used as Notebook toolbar), to compute the items position when resizing the toolbar.

This change included a side effect when displaying a notebook loaded at startup but not displayed (in an other tab of the main area): the order of items in toolbar was reversed.

As far as I understand, for hidden notebook, the items are not added in the toolbar at startup because the toolbar was not rendered. These items were added to the responsive toolbar, and it uses the resize event to add them when the Notebook is displayed. This was not working with the new logic to order the items.

This PR also fixes the relative insert in the reactive toolbar (see https://github.com/jupyterlab/jupyterlab/issues/15509#issuecomment-1866479457 for context). This fixes the debugger item position in the toolbar when the notebook is not displayed.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15509

## Code changes

Changes in the reactive toolbar:
- add a relative insert method
- do not force items index to 0 if it does not have a known position

## User-facing changes

Restore the correct order of items in Notebook toolbar.

None

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -572,7 +572,7 @@ export class ReactiveToolbar extends Toolbar<Widget> {
           opener.hide();
         }
         if (twiceCall) {
-          this._onResize();
+          void this._onResize();
         }
       })
       .catch(msg => {
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -259,7 +259,7 @@ export class Toolbar<T extends Widget = Widget> extends Widget {
    * The item can be removed from the toolbar by setting its parent to `null`.
    */
   insertAfter(at: string, name: string, widget: T): boolean {
-    return this._insertRelative(at, 1, name, widget);
+    return this.insertRelative(at, 1, name, widget);
   }
 
   /**
@@ -279,10 +279,10 @@ export class Toolbar<T extends Widget = Widget> extends Widget {
    * The item can be removed from the toolbar by setting its parent to `null`.
    */
   insertBefore(at: string, name: string, widget: T): boolean {
-    return this._insertRelative(at, 0, name, widget);
+    return this.insertRelative(at, 0, name, widget);
   }
 
-  private _insertRelative(
+  protected insertRelative(
     at: string,
     offset: number,
     name: string,
@@ -416,6 +416,25 @@ export class ReactiveToolbar extends Toolbar<Widget> {
   }
 
   /**
+   * insert an item relatively to an other item.
+   */
+  protected insertRelative(
+    at: string,
+    offset: number,
+    name: string,
+    widget: Widget
+  ): boolean {
+    const targetPosition = this._widgetPositions.get(at);
+    const position = (targetPosition ?? 0) + offset;
+    this._widgetPositions.forEach((value, key) => {
+      if (key !== TOOLBAR_OPENER_NAME && value >= position) {
+        this._widgetPositions.set(key, value + 1);
+      }
+    });
+    return this.insertItem(position, name, widget);
+  }
+
+  /**
    * Insert an item into the toolbar at the specified index.
    *
    * @param index - The index at which to insert the item.
@@ -444,8 +463,11 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     }
 
     // Save the widgets position when a new widget is inserted.
-    if (this._widgetPositions.get(name) === undefined) {
-      this._saveWidgetPosition();
+    if (
+      name !== TOOLBAR_OPENER_NAME &&
+      this._widgetPositions.get(name) === undefined
+    ) {
+      this._widgetPositions.set(name, index);
     }
     return status;
   }
@@ -474,7 +496,7 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     }
   }
 
-  private _onResize() {
+  private async _onResize() {
     if (!(this.parent && this.parent.isAttached)) {
       return;
     }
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @brichet for working on this :+1: 

Maybe this is also be related to this issue in Notebook 7: https://github.com/jupyter/notebook/issues/6553. Although it looked like it was not an issue anymore recently (and was closed).

brichet:
> Maybe this is also be related to this issue in Notebook 7: [jupyter/notebook#6553](https://github.com/jupyter/notebook/issues/6553). Although it looked like it was not an issue anymore recently (and was closed).

This PR mostly fixes the items order in the toolbar.
The logic of the reactive toolbar was changed a bit in https://github.com/jupyterlab/jupyterlab/pull/15021, maybe it fixed the issue you linked (if it is fixed).

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -415,6 +415,25 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     return super.insertAfter(at, name, widget);
   }
 
+  /**
+   * insert an item relatively to an other item.
```
krassowski, line 419:
```suggestion
   * Insert an item relatively to an other item.
```

packages/ui-components/src/components/toolbar.tsx
```diff
@@ -279,10 +279,10 @@ export class Toolbar<T extends Widget = Widget> extends Widget {
    * The item can be removed from the toolbar by setting its parent to `null`.
    */
   insertBefore(at: string, name: string, widget: T): boolean {
-    return this._insertRelative(at, 0, name, widget);
+    return this.insertRelative(at, 0, name, widget);
   }
 
-  private _insertRelative(
+  protected insertRelative(
```
krassowski, line 285:
```suggestion
  /**
   * Insert an item relatively to an other item.
   */
  protected insertRelative(
```

krassowski (commented):
Thank you so much! The order seems to be fixed, but sometimes after reloading page with multiple notebooks I still see all items collapsed into the collapser in the background notebooks:

stil-happening

After resizing they move into correct positions.

I guess the toolbar width used in ReactiveToolbar may have zero width initially for hidden widgets? This was on Chromium browser with hidden mode set to "display" (default).

krassowski:
Maybe invoking the `onResize` logic in `onAfterAttach` or `onAfterShow` of `ReactiveToolbar` is the way to solve it?

brichet:
Thanks @krassowski for review and suggestion.

> The order seems to be fixed, but sometimes after reloading page with multiple notebooks I still see all items collapsed into the collapser in the background notebooks

Surprisingly, I observe the opposite. The background notebooks get all the items in the toolbar on after show, without the ellipses one.
What I understand is:
- the toolbar of background notebooks is not rendered, and have a size of 0px.
- all the items are added to the responsive toolbar (since the toolbar is not rendered), but all the items are not rendered too, so they have a size of 0px too.
- when displaying the notebook, the toolbar get a size, but the items have still a known size of 0px (not yet rendered), so they are all added in the toolbar, and the ellipses items seems not necessary.
- if we compute again the toolbar it works, because the toolbar and the items have already been rendered, and have a known size.

> Maybe invoking the `onResize` logic in `onAfterAttach` or `onAfterShow` of `ReactiveToolbar` is the way to solve it?

Unfortunately, this is not enough, see explication above.

The 2 ways I found to fix it are:

1. adding every item in the toolbar to retrieve its real size, and remove it if necessary. It looks a bit weird because we can see every items addition when displaying the notebook.

Peek 2023-12-27 09-02

2. computing the toolbar twice after show. This is probably not the most efficient, but it works well on my side, because it first adds all the items in the toolbar (and retrieves their size), and then use the correct size of the toolbar and items.

Peek 2023-12-27 09-16

I kept the second solution, let me know if it fixes it for you too.

brichet:
Test failures may be relevant, since it is the same on Linux and Mac, but I have no idea how to fix it.

krassowski:
> Test failures may be relevant, since it is the same on Linux and Mac, but I have no idea how to fix it.

No `usage2` failures are due to an upstream change (https://github.com/jupyterlab/jupyterlab/issues/15570).

krassowski (commented):
This works locally for me too. It would be great if we could avoid the initial jitter, but that is a minor problem we can handle in a future PR.

Just two suggestions:
- can you add a comment in code documenting why we need `twiceCall`?
- would it be better to name it `callTwice`?

jtpio:
Just tried the latest state in a clean environment, and also seeing a little delay when switching notebooks. But the items don't seem to be collapsing anymore after reloading the page multiple times.

[jupyterlab-notebook-toolbar.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/4ab11133-cc95-4635-899c-bee1b741a905)

brichet:
> Just two suggestions:
> 
> * can you add a comment in code documenting why we need `twiceCall`?
> * would it be better to name it `callTwice`?

Done

brichet:
> It would be great if we could avoid the initial jitter, but that is a minor problem we can handle in a future PR.

> Just tried the latest state in a clean environment, and also seeing a little delay when switching notebooks.

Thanks for testing. The delay should be visible only the first time the notebook is displayed. Not sure if this can be improved, other than by forcing a rendering before the notebook is visible (if possible).

krassowski:
Not sure why CI got stuck:
image

I will close/reopen to see if it helps

krassowski (approved):
Thank you @brichet, looks good to me!

I will wait a little bit with merging in case if anyone else wants to take a look (accommodating for slower pace in the festive season).

PR #15554, opened 2023-12-21
Added aria-live announcement when a notebook cell is taking a long time to process. 

m158261:
## References

 [Accessibility Issues Needing Addressing for WCAG 2.1 compliance (As of Version 2.2.6)](https://github.com/jupyterlab/jupyterlab/issues/9399) - Extra Credit Issue Area # 1

Long waiting times or screens without text should inform the users with screen readers what the system is doing so they don’t think think it crashed or computer black screened (so they don’t turn restart their computer).

## Code changes

I have added an aria-live region that appends to the main content panel as that is high up in the DOM and never hidden. When the user runs a notebook cell the kernel status changes to busy and a 10 second timer is set. If this timer expires, text is append to the live region which will be announced by a screen reader. If the timer does not reach the end it is cancelled when the kernel status changes back to idle. 

## User-facing changes

No visual changes. Users with screen readers will get an announcement that the kernel status is busy if a notebook cell takes more than 10 seconds to complete. 

None

### DIFF
--- a/packages/application/src/status.ts
+++ b/packages/application/src/status.ts
@@ -73,13 +73,22 @@ export class LabStatus implements ILabStatus {
    */
   setBusy(): IDisposable {
     const oldBusy = this.isBusy;
+    const ariaLiveRegion = document.getElementById('notify-user-aria-live');
+    let timeout: any;
     this._busyCount++;
     if (this.isBusy !== oldBusy) {
       this._busySignal.emit(this.isBusy);
+      timeout = setTimeout(() => {
+        ariaLiveRegion!.append(
+          'A notebook cell is processing. Kernel status is busy.'
+        );
+      }, 10000);
     }
     return new DisposableDelegate(() => {
       const oldBusy = this.isBusy;
       this._busyCount--;
+      clearTimeout(timeout);
+      ariaLiveRegion!.append('Processing complete. Kernel status is idle.');
       if (this.isBusy !== oldBusy) {
         this._busySignal.emit(this.isBusy);
       }
--- a/packages/application/src/shell.ts
+++ b/packages/application/src/shell.ts
@@ -412,6 +412,13 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     hboxPanel.addWidget(vsplitPanel);
     hboxPanel.addWidget(rightHandler.sideBar);
 
+    const ariaLiveRegion = document.createElement('div');
+    ariaLiveRegion.setAttribute('aria-live', 'polite');
+    ariaLiveRegion.setAttribute('role', 'region');
+    ariaLiveRegion.setAttribute('id', 'notify-user-aria-live');
+    ariaLiveRegion.className = 'jp-ContentPanel-ariaLive';
+    hboxPanel.node.appendChild(ariaLiveRegion);
+
     rootLayout.direction = 'top-to-bottom';
     rootLayout.spacing = 0; // TODO make this configurable?
     // Use relative sizing to set the width of the side panels.
--- a/packages/application/style/core.css
+++ b/packages/application/style/core.css
@@ -90,3 +90,8 @@ body {
   display: flex;
   align-items: center;
 }
+
+.jp-ContentPanel-ariaLive {
+  width: 0;
+  overflow: hidden;
+}


packages/application/src/status.ts
```diff
@@ -73,13 +73,22 @@ export class LabStatus implements ILabStatus {
...
   setBusy(): IDisposable {
     const oldBusy = this.isBusy;
+    const ariaLiveRegion = document.getElementById('commands-aria-live');
+    let timeout: any;
     this._busyCount++;
     if (this.isBusy !== oldBusy) {
       this._busySignal.emit(this.isBusy);
+      timeout = setTimeout(() => {
+        ariaLiveRegion!.append(
+          'A notebook cell is processing. Kernel status is busy.'
```
krassowski, line 83: This needs translation

packages/application/src/status.ts
```diff
@@ -73,13 +73,22 @@ export class LabStatus implements ILabStatus {
...
+        );
+      }, 10000);
     }
     return new DisposableDelegate(() => {
       const oldBusy = this.isBusy;
       this._busyCount--;
+      clearTimeout(timeout);
+      ariaLiveRegion!.append('Processing complete. Kernel status is idle.');
```
krassowski, line 91: Needs translation?

m158261:

packages/application/src/status.ts
m158261: Changes made.

packages/application/src/status.ts
m158261: Changes made.

packages/application/src/status.ts
```diff
@@ -71,15 +72,28 @@ export class LabStatus implements ILabStatus {
...
+    const ariaLiveRegion = document.getElementById('commands-aria-live');
+    let timeout: any;
     this._busyCount++;
     if (this.isBusy !== oldBusy) {
       this._busySignal.emit(this.isBusy);
+      timeout = setTimeout(() => {
+        ariaLiveRegion!.append(
+          trans.__('A notebook cell is processing. Kernel status is busy.')
+        );
+      }, 10000);
```
gabalafou, line 88: Concern raised by @tonyfast, this time should be configurable

packages/application/src/status.ts
```diff
@@ -71,15 +72,28 @@ export class LabStatus implements ILabStatus {
    *
    * @returns A disposable used to clear the busy state for the caller.
    */
-  setBusy(): IDisposable {
+  setBusy(translator?: ITranslator): IDisposable {
```
krassowski, line 75: this is not passed anywhere

packages/application/src/status.ts
```diff
@@ -71,15 +72,28 @@ export class LabStatus implements ILabStatus {
...
+    translator = translator || nullTranslator;
+    const trans = translator.load('jupyterlab');
     const oldBusy = this.isBusy;
+    const ariaLiveRegion = document.getElementById('commands-aria-live');
+    let timeout: any;
```
krassowski, line 80:
```suggestion
    let timeout: number;
```

Or `ReturnType<typeof setInterval>`

packages/application/src/status.ts
```diff
@@ -71,15 +72,28 @@ export class LabStatus implements ILabStatus {
...
     this._busyCount++;
     if (this.isBusy !== oldBusy) {
       this._busySignal.emit(this.isBusy);
+      timeout = setTimeout(() => {
+        ariaLiveRegion!.append(
+          trans.__('A notebook cell is processing. Kernel status is busy.')
+        );
```
krassowski, line 87: Do we need to clear the live region to avoid memory leak?

tonyfast: yes. an activity log or aria log will need to be cleared. over the course of an interactive computing session the activity will fill up, especially after hours of work. memory leaks are inevitable in these cases.

assertive aria live notifications will be announced immediately and may be cleared sooner while polite aria live messages linger until current speaking buffer clears. this AT buffer could be reading an entire page or post meaning that polite aria messages may take many minutes to be announced. 
unfortunately, there is NO way to know when to clear the aria live buffer because the browser can't know when messages have reached assistive technology. for polite aria messages the buffer might need to persist for minutes maybe up tens of minutes; it is possible an aria live message will become stale if enough time and activity have elapsed. we can't make too many assumptions about time because for the varied ways disabled people experience crip time.

tonyfast:
i dont have a good perspective of the jupyterlab issues anymore so i couldnt find issues about long running cells. maybe one of yall know better than me. 

im wondering if this feature could be considered for ALL users, maybe it is a follow up issue. long running cells impose a cognitive inaccessibility that can distract different neurotypes from their task at hand. it is easy to forget the actions done in the past while a long running cell is executed. exposing visual and/or audible feedback for ALL users might be a direction this work could take us. is the long running cell the right feature to tether the aria live activity log to the toast notification system. when a long running cell completes maybe we hear a message that cells "cell xx in notebook yy completed after zz seconds. click here to return to the cell". considering this feature might help with ["ensuring processes do not rely on memory"](https://www.w3.org/WAI/WCAG2/supplemental/objectives/o6-memory/)

PR #15386, opened 2023-11-10
Improve scrolling to heading

fcollonval:
## References

Fixes #15228
Fixes #15215 (@krassowski - would you mind confirm this I'm unsure)
Fixes #14684
Fixes #14591

## Code changes

- Emit `onActiveHeadingChanged` in toc model even when the value is not changing
- Add a new attribute `scrollToTop` to the notebook ToC factory that is true by default but allows to configure the heading navigation scroll behavior
- Switch to command mode when in edit mode and jumping to a heading in a md cell

## User-facing changes

By default, jumping to a heading from the toc will not make the markdown cell turn in edit mode and will scroll it to the top of the view port.

## Backwards-incompatible changes

None

PR #15565, opened 2023-12-26
Backport PR #15386: Improve scrolling to heading

krassowski:
Backport PR #15386: Improve scrolling to heading


krassowski:

krassowski:
`Check internationalization changes / check_i18n` is failing because we are adding a setting with user-facing setting string; we normally do not do this in backports to patch releases. We can override this rule (which we did in the past) if there is an agreement that adding the setting is more useful than the inconvenience of users with non-English interfaces for the time when the translations are not updated). Thoughts?

`usage2` failures are unrelated and tracked in https://github.com/jupyterlab/jupyterlab/issues/15570, visual tests failures are unrelated (a fix needs backporting).

packages/notebook/src/toc.ts
```diff
@@ -521,24 +532,37 @@ export class NotebookToCFactory extends TableOfContentsFactory<NotebookPanel> {
...
+            );
           }
         };
 
         const cell = heading.cellRef;
         const cells = widget.content.widgets;
         const idx = cells.indexOf(cell);
+        // Switch to command mode to avoid entering Markdown cell in edit mode
+        // if the document was in edit mode
+        if (cell.model.type == 'markdown' && widget.content.mode != 'command') {
```
trungleduc, line 562: no need `===`?

krassowski: I don't think it is needed here; the only case these would return different results would be if `cell.model.type` were `new String('markdown')` but it cannot be that since it is typed to be literal rather than object string.

Issue #15567, opened 2023-12-27
Show the notification center when the status bar plugin is disabled


jtpio:
### Problem

Currently clicking on the `View > Show Notifications` menu item does not open the notification center if the status bar plugin is disabled:

[jupyterlab-statusbar-disabled.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/1dacf844-d99d-4ccd-860c-523d0a4aa806)

### Proposed Solution

Ideally the notification center would still open in the bottom right corner of the page, even if the status bar is disabled, just like when it is available (by default):

[jupyterlab-statusbar-enabled.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/258a3386-66fd-4982-a1c5-7a8e27bea6f4)

### Additional context

Being able to show the notification center even if the status bar is not available would be useful in Notebook 7: https://github.com/jupyter/notebook/issues/7195

jtpio:
Quickly looking at the plugin, it could be because the `anchor` for the notification center popup is only added to the DOM id the status bar is available:

https://github.com/jupyterlab/jupyterlab/blob/e075957cf54814353ab5f7b6562584503acb7a5d/packages/apputils-extension/src/notificationplugin.tsx#L626-L632

PR #15568, opened 2023-12-27
Show the notification center if the status bar is disabled

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15567

[jupyterlab-notification-center-no-statusbar.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/7fc7b2e4-28dd-467d-b5bb-8395048a8030)

## Code changes

If the status bar is not available, still position the `notificationStatus` widget on the page, so it can be used as the anchor for the popup.

## User-facing changes

Notebook 7 users will be able to open the notification center via the same `View > Show Notifications` menu entry as in JupyterLab: https://github.com/jupyter/notebook/issues/7195

None

### DIFF
--- a/packages/apputils-extension/src/notificationplugin.tsx
+++ b/packages/apputils-extension/src/notificationplugin.tsx
@@ -40,6 +40,7 @@ import {
   ReadonlyJSONObject,
   ReadonlyJSONValue
 } from '@lumino/coreutils';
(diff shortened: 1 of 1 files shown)


jtpio:
Labeling this PR as `bug`, as this fixes a behavior that might be confusing to users who have disabled the status bar.

PR #15571, opened 2023-12-27
Accept/invoke inline completions with `Tab`

### LINKED ISSUE #15549: Support for accepting/triggering inline completions with `Tab` key
## Description

Currently inline completions can be:
- accepted with <kbd>Alt</kbd> + <kbd>End</kbd>
- invoked manually with <kbd>Alt</kbd> + <kbd>\\</kbd>

Some users prefer to use <kbd>Tab</kbd> for either or both of these actions. While all of actions have configurable shortcuts, <kbd>Tab</kbd> is special and does not work even if configured because it is intercepted by notebook handlers.

## Reproduce

1. Enable history inline completion provider
2. Change shortcut to Tab in settings
3. Try to invoke or accept with Tab
4. See that neither works

## Expected behavior

It should be possible to use Tab here

## Context

- JupyterLab version: 4.1.0b0

krassowski:
## References

- fixes #15549
- follow up to #15160

## Code changes

- add event listener to intercept the <kbd>Tab</kbd> press and if needed dispatch appropriate action
  - correct the selector for command accepting of inline completion so that it is only active when there is a suggestion (by adding `jp-mod-inline-completer-active` class) - this is required to allow configuring both `invoke` and `accept` actions to respond to <kbd>Tab</kbd>
- make shortcut labels on the inline completer widget reflect the current shortcut rather than initial one
  - change `CommandToolbarButtonComponent` to accept `label` function, matching the commands API

## User-facing changes

-  <kbd>Tab</kbd>  works for `accept` and `invoke` of inline completer
- It is now possible to set both `accept` and `invoke` to  <kbd>Tab</kbd> 
   image
- The shortcuts now appear reliably in the inline completer widget; previously these were randomly missing despite being enabled
- The shortcuts in the inline completer widget now reflect the current shortcut without need to reload JupyterLab
  image

None

### DIFF
--- a/galata/test/jupyterlab/inline-completer.test.ts
+++ b/galata/test/jupyterlab/inline-completer.test.ts
@@ -7,6 +7,7 @@ const fileName = 'notebook.ipynb';
 const COMPLETER_SELECTOR = '.jp-InlineCompleter';
 const GHOST_SELECTOR = '.jp-GhostText';
 const PLUGIN_ID = '@jupyterlab/completer-extension:inline-completer';
+const SHORTCUTS_ID = '@jupyterlab/shortcuts-extension:shortcuts';
 
 const SHARED_SETTINGS = {
   providers: {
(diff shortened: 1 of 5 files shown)


jtpio (approved):
Thanks!

Issue #11352, opened 2021-10-25
Jupyter File open in multiple browsers -> Warning + Copy


stromal:
## Problem
Whenever you have a notebook opened by 2 individual (opening Jupyter lab in multiple browser windows).

ERROR MESSAGE: 

`"folder/filename.ipynb" has changed on disk since the last time it was opened or saved. Do you want to overwrite the file on disk with the version open here, or load the version on disk (revert)?`

![2021 10 25 delete changs 2](https://user-images.githubusercontent.com/19979901/138732852-105cae85-2ab3-49d4-8863-851bc84ac48b.png)

## Solution
A.) Warning message that it is open elsewhere (IP address or it can on this machine or by someone else).
B.) create a copied version of the project so you don't lose code.

## WHY
Because like this one or the other opening want to overwrite the file and if it has not git backup than it results in to code loss.

Issue #7743, opened 2020-01-06
Time conflict and the "File Changed" dialog appear every time an attempt is made to save a Notebook


mpaunovi:
## Description
When saving a Notebook file (6.83 MB), the PUT request (api/contents/{someNotebook}) takes about 1.5 seconds. This causes a time conflict and displays the "File Changed" popup since it does not match the allowed 0.5 seconds difference between disk time and client time.

Screenshot:
![image](https://user-images.githubusercontent.com/43001714/71818777-2a35a400-308a-11ea-86b5-2bd0dd6b2082.png)

Note: The same notebook is not open on another tab or browser, and no one changes that notebook except the current user.

## Reproduce

This issue happens occasionally. It's not completely clear what causes this behavior, but once it happens it will happen every time with every attempt to save the notebook file.

## Context

- Operating System and version: macOS Mojave Version 10.14.6
- Browser and version: Chrome Version 79.0.3945.88
- JupyterLab version: 1.0.4
- Notebook version: 6.0.0

<details><summary>Server log snippet</summary>
<pre>
[I 14:18:29.230 NotebookApp] 200 GET /jupyter/jupyter-{{user_id}}/api/sessions?sessionId(IP address) 2.71ms
[I 14:18:35.744 NotebookApp] Saving file at /Some_Notebook_File.ipynb
[W 14:18:35.800 NotebookApp] Notebook Some_Notebook_File.ipynb is not trusted
[I 14:18:37.178 NotebookApp] 200 PUT /jupyter/jupyter-{{user_id}}/api/contents/Some_Notebook_File.ipynb?1************ (IP address) 1511.96ms
[I 14:18:37.234 NotebookApp] 200 GET /jupyter/jupyter-{{user_id}}/api/contents/?content=1&1************ (IP address) 53.37ms
[I 14:18:37.277 NotebookApp] 200 GET /jupyter/jupyter-{{user_id}}/api/terminals?1************(IP address) 1.88ms
[I 14:18:37.725 NotebookApp] 200 GET /jupyter/jupyter-{{user_id}}/api/contents/Some_Notebook_File.ipynb/checkpoints?1************ (IP address) 3.08ms
[I 14:18:38.316 NotebookApp] 201 POST /jupyter/jupyter-{{user_id}}/api/contents/Some_Notebook_File.ipynb/checkpoints?1************ (IP address) 590.44ms
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
[...]

PR #15037, opened 2023-08-29
bug fix: timestamp changed but content did not change

ctappy:
## References

Fixes #7743
Fixes #11352 

## Code changes

fixes timestamp changes but the content did not change. An issue with fuse filesystems, timing issue #7743, multiple users opening the same file #11352, and file touched but not edited issues.

## User-facing changes

N/A

## Backwards-incompatible changes

N/A

PR #15577, opened 2023-12-29
Fix spurious "File Changed" dialogs using `hash` from `jupyter-server` v2.11.1+

Wh1isper:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/7743
Fixes https://github.com/jupyterlab/jupyterlab/issues/11352

References PR: https://github.com/jupyterlab/jupyterlab/pull/15037

Upstream:
- https://github.com/jupyter-server/jupyter_server/pull/1363
- https://github.com/jupyter-server/jupyter_server/pull/1367

## Code changes

Since jupyter server may provide hash in model, we compare hash for the file changes first.

**Forgive me for being a TypeScript and front-end novice, more advice would be appreciated, thanks!**

## User-facing changes

N/A

N/A


packages/docregistry/src/context.ts
```diff
@@ -780,13 +827,13 @@ or load the version on disk (revert)?`,
       label: this._trans.__('Overwrite'),
       actions: ['overwrite']
     });
-    this._timeConflictModalIsOpen = true;
+    this._ConflictModalIsOpen = true;
```
krassowski, line 830:
```suggestion
    this._conflictModalIsOpen = true;
```

packages/docregistry/src/context.ts
```diff
@@ -780,13 +827,13 @@ or load the version on disk (revert)?`,
...
     return showDialog({
       title: this._trans.__('File Changed'),
       body,
       buttons: [Dialog.cancelButton(), revertBtn, overwriteBtn]
     }).then(result => {
-      this._timeConflictModalIsOpen = false;
+      this._ConflictModalIsOpen = false;
```
krassowski, line 836:
```suggestion
      this._conflictModalIsOpen = false;
```

packages/docregistry/src/context.ts
```diff
@@ -917,7 +964,7 @@ or load the version on disk (revert)?`,
   private _disposed = new Signal<this, void>(this);
   private _dialogs: ISessionContext.IDialogs;
   private _lastModifiedCheckMargin = 500;
-  private _timeConflictModalIsOpen = false;
+  private _ConflictModalIsOpen = false;
```
krassowski, line 967:
```suggestion
  private _conflictModalIsOpen = false;
```

packages/docregistry/src/context.ts
```diff
@@ -746,22 +801,14 @@ export class Context<
...
     options: Partial<Contents.IModel>
   ): Promise<Contents.IModel> {
-    const tDisk = new Date(model.last_modified);
-    console.warn(
-      `Last saving performed ${tClient} ` +
-        `while the current file seems to have been saved ` +
-        `${tDisk}`
-    );
-    if (this._timeConflictModalIsOpen) {
+    if (this._ConflictModalIsOpen) {
```
krassowski, line 811:
```suggestion
    if (this._conflictModalIsOpen) {
```

krassowski (changes requested):
Thank you for this PR! It looks good except for one condition which I am not sure about.

Would you consider also adding a test for this? I think it could be added along:

https://github.com/jupyterlab/jupyterlab/blob/6b498b46e98dd796e6b3022363907ba97de5bce8/packages/docregistry/test/context.spec.ts#L85-L108

packages/docregistry/src/context.ts
```diff
@@ -476,11 +476,19 @@ export class Context<
...
     };
     const mod = this._contentsModel?.last_modified ?? null;
+    const hash = this._contentsModel?.hash ?? null;
     this._contentsModel = newModel;
-    if (!mod || newModel.last_modified !== mod) {
+    if (
+      !mod ||
+      newModel.last_modified !== mod ||
+      !hash ||
+      newModel.hash !== hash
```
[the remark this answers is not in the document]
Wh1isper: Yes, we should assume that it is possible that `hash` does not exist(is `null`). This is not the same as `mod`.

Wh1isper:
I will add a test on this soon, thanks!

packages/docregistry/test/context.spec.ts
Wh1isper, line 1: I'm not sure such a test is complete or enough. Feel free to give more feedback!

Wh1isper, line 1: I'm not sure such a test is complete or enough. Feel free to give more feedback!

packages/docregistry/test/context.spec.ts
```diff
@@ -85,10 +85,15 @@ describe('docregistry/context', () => {
...
         const path = context.path;
+        const hash = context.contentsModel?.hash ?? null;
         let called = false;
         context.fileChanged.connect((sender, args) => {
           expect(sender).toBe(context);
           expect(args.path).toBe(path);
+          if (hash) {
+            expect(args.hash).not.toBeNull();
+            expect(args.hash).not.toBe(hash);
+          }
```
krassowski, line 96: I think that rather than using a conditional expect a new dedicated test would be desirable. In particular such a test should verify that the `context.fileChanged` is not emitted. It will need some work to mock the server response here I believe.

Wh1isper:
I changed some logic about emit `fileChanged` and added `hash` and `hash_algorithm` in mock.

I have tested following situations manually: 
- when `hash` is provided, `fileChanged` is not emitted.
- when `hash` is not provided, `fileChanged` is emitted because of `timeStamp` changed.

But I don't know how to reorganise the test cases to support both scenarios.

Feel free to give me more feed back, thanks!

krassowski (approved):
Thank you @Wh1isper!

I pushed a few more tests and tiny adjustments in the conditionals.

Issue #15547, opened 2023-12-20
Notebook is permissioned as read-only shows up incorrectly on RTC


krassowski:
## Description

https://github.com/jupyterlab/jupyterlab/pull/15317 introduced a helpful pop-up which shows on ctrl + s in read-only notebooks. Because RTC advertises its model as read-only, it leads to a spurious notification on every ctrl + S 

## Reproduce

1. Open a notebook with RTC mode on, you can use https://mybinder.org/v2/gist/krassowski/847732eb6c16e168fb4c0b16d9706142/HEAD?urlpath=lab
2. Change something
3. Press Ctrl + S
4. See the spurious notification:
![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/bfa56c09-a1fd-4d3f-a233-b389b15df619)

## Expected behavior

No notification in this case

## Context

CC @andrewfulton9

JasonWeill:
@andrewfulton9 Can you please take a look at this when using real-time collaboration (RTC)?

andrewfulton9:
Yeah, I'll take a look asap

PR #15597, opened 2024-01-02
Fix ctrl+s notification incorrectly showing up in RTC mode

andrewfulton9:
## References

#15547 

## Code changes

Adds a check to verify that a document is not rtc mode along with the check on if a document is not writable for raising a notification that a document is read-only.

## User-facing changes

The user will no longer see an improper notification that a file is read-only when they use the keybinding "ctrl+s" when in rtc mode


krassowski (approved):
Thank you @andrewfulton9!

Issue #11416, opened 2021-11-08
Galata: upload content via `serviceManager.contents`?


jtpio:
### Problem

At the moment Galata uploads contents by making requests to the Jupyter Server:

https://github.com/jupyterlab/jupyterlab/blob/114159269a76c4a8d4c1e725e882d7c8a7791869/galata/src/contents.ts#L186-L189

With `_fetch` defined here:

https://github.com/jupyterlab/jupyterlab/blob/114159269a76c4a8d4c1e725e882d7c8a7791869/galata/src/contents.ts#L387-L404

### Proposed Solution

Maybe this could be an additional method instead of replacing the existing behavior.

Ideally it would be nice to have a helper function wrapping `window.galataip.app.serviceManager.contents` to upload content via the `ServiceManager`.

### Additional context

This came up when adding UI tests to JupyterLite: https://github.com/jupyterlite/jupyterlite/pull/414

JupyterLite doesn't have an application server. Instead it relies on providing a custom [`ServerConnection.ISettings`](https://github.com/jupyterlab/jupyterlab/blob/114159269a76c4a8d4c1e725e882d7c8a7791869/packages/services/src/serverconnection.ts#L53-L107) to manage calls to the contents API.

krassowski:
+1 in addition, filebrowser model has a better implementation which takes care of large files (and shows progress on status bar), but:
- it uploads relative to the current directory, which makes it useless when using via `factory.defaultBrowser.model.upload` (`factory: IFileBrowserFactory`)
- it is not very convenient to use as we have to depend on/access the fielbrowser.

And of course, we should ideally only have one implementation of upload if possible.

PR #15607, opened 2024-01-05
Restore previous Galata `page.filebrowser.refresh()` timeout logic

jtpio:
## References

As noticed in https://github.com/jupyterlite/jupyterlite/pull/1263#issuecomment-1878410761, https://github.com/jupyterlab/jupyterlab/pull/15021 seems to have changed the logic of the `page.filebrowser.refresh()` Galata helper, more than just updating the selector.

## Code changes

Restore the previous timeout logic, to allow for using this helper in lab-based applications using a different contents API (like in JupyterLite storing files in the browser).

## User-facing changes

None

None

### DIFF
--- a/galata/src/helpers/filebrowser.ts
+++ b/galata/src/helpers/filebrowser.ts
@@ -216,7 +216,9 @@ export class FileBrowserHelper {
         '.jp-ToolbarButtonComponent[data-command="filebrowser:refresh"]'
       );
 
-    // wait for network response or timeout
+    // Use Promise.race to manage the network response timeout
+    // This is useful for lab-based applications not using the Jupyter Server Contents API.
+    // such as JupyterLite, to avoid having the waitForAPIResponse call fail.
     await Promise.race([
       page.waitForTimeout(2000),
       this.contents.waitForAPIResponse(async () => {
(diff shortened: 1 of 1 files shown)


jtpio:
cc @brichet as author of https://github.com/jupyterlab/jupyterlab/pull/15021

galata/src/helpers/filebrowser.ts
```diff
@@ -217,12 +217,12 @@ export class FileBrowserHelper {
       );
 
     // wait for network response or timeout
-    await this.contents.waitForAPIResponse(
-      async () => {
+    await Promise.race([
```
krassowski, line 220: Since we cannot easily test jupyterlite as downstream, is it worth adding a comment explaining why this approach is preferred?

jtpio: Sounds good :+1:

jtpio:
Also this might also be related to https://github.com/jupyterlab/jupyterlab/issues/11416, as to how Galata makes use of the contents API.

brichet:
> cc @brichet as author of #15021

Sorry for that, I misunderstood the use of the `promise.race` in that case.
:+1: for the comment

brichet (approved):
Thanks @jtpio, LGTM

Issue #14354, opened 2023-04-10
Styled input checkbox focus state is not visible on Ubuntu/Firefox


krassowski:
## Description

For checkboxes with `jp-mod-styled` class the focus state is not visible. This might be Operating System specific.

## Reproduce

1. Open search box in a notebook
2. Open filters
3. Tab to a checkbox
4. See no focus indication

On Ubuntu/Firefox

![Screenshot from 2023-04-10 15-28-33](https://user-images.githubusercontent.com/5832902/230921695-c89dbd91-8e27-4319-9a1e-3f18f0cd05b3.png)

The one on right is focused.

The styles are defined here:

https://github.com/jupyterlab/jupyterlab/blob/89154f525578e3c79d9f688be1e67605f6d2d443/packages/ui-components/style/styling.css#L45-L48

but it has no effect on styles for me in Firefox on Ubuntu.

Manually adding `outline: 3px solid red;` to `input.jp-mod-styled:focus` shows an indication properly:

![Screenshot from 2023-04-10 15-30-40](https://user-images.githubusercontent.com/5832902/230921881-ceba7417-88d0-4029-a99a-e7f7071fb8c5.png)

Of note, the styles are effective on Ubuntu/Chromium but are the indicator is rather faint:

![Screenshot from 2023-04-10 15-36-15](https://user-images.githubusercontent.com/5832902/230923107-7a814346-0064-41d4-8b7b-3298e346c57f.png)

## Expected behavior

Some kind of indicator is visible for checkboxes across browsers

## Context

- Operating System and version: Ubuntu
- Browser and version: Firefox Nightly 108
- JupyterLab version: 4.0.0b1

JasonWeill:
Also happens on Firefox ESR 102.9 on macOS with 4.0.0b1.

alden-ilao:
Hello! This is my first time taking a swing at open source contributions and would love some feedback if possible!

fcollonval:
Thanks @alden-ilao for working on this. @krassowski (thanks for the review), did #15612 fix this fully?

krassowski:
Yes, thanks for catching this!

PR #15612, opened 2024-01-06
Fixes focus indicator on input checkbox for Firefox

alden-ilao:
## References

#14354 

## Code changes
* updated packages/apputils/style/dialog.css:124
* packages/ui-components/style/styling.css:49

- outline displayed on focus for checkboxes now visible on all platforms
- styling for checkbox:focus now sourced by single file instead of two (packages/ui-components/style/styling.css)

## User-facing changes
The user will now see outlines when hovering checkboxes on all platforms
Before (Firefox): 
image
Before (Chrome):
image
After (Firefox):
image
After (Chrome):
image


krassowski:
Thank you! Can you confirm that it still works in the dialog; for example you can try clicking on the kernel name to show change kernel dialog, this is how it looks before:

image

Based on the CI results it looks this will be reflected in the snapshots, let's get them updated:

alden-ilao:
Thanks for the quick reply! I've been meaning to ask where you are able to get the checkboxes in the dialogs

here's what i got:
image
image

packages/apputils/style/dialog.css
```diff
@@ -121,11 +121,6 @@ button.jp-Dialog-close-button {
   padding-right: 5px;
 }
 
-.jp-Dialog-checkbox > input:focus-visible {
-  outline: 1px solid var(--jp-input-active-border-color);
-  outline-offset: 1px;
-}
-
```
alden-ilao, line 128: (Specifically for checkboxes in dialogs)
for context, this styling is applied by `packages/apputils/src/dialog.tsx` by using the label class as its initial selector.

image

But further down the execution create footer calls `Styling.style(footer.node);` which adds `jp-mod-styled` to the checkbox element. Deleting this styling in`packages/apputils/style/dialog.css` reduces code duplication.
image

krassowski:

packages/ui-components/style/styling.css
```diff
@@ -46,6 +46,11 @@ input.jp-mod-styled:focus {
   box-shadow: inset 0 0 4px var(--md-blue-300);
 }
 
+input[type='checkbox'].jp-mod-styled:focus {
```
krassowski, line 49:
```suggestion
input[type='checkbox'].jp-mod-styled:focus-visible {
```

krassowski: It feels like we should only show the indicator when browser believes that the focus is applicable.

krassowski (approved):
Thank you @alden-ilao, this looks good to me.

Sorry for back-and-forth with updating the snapshots - when preserving `:focus-visible` approach updates to snapshots were not needed.

krassowski:

alden-ilao:
@krassowski  Apologies for the late reply, I did not have access computer the past couple days, thank you for reviewing and making the changes!

PR #15626, opened 2024-01-10
Update lifecycle info for JupyterLab major versions

### LINKED ISSUE #12397: Add version support timelines to documentation
### Problem

As discussed at the JupyterLab security meetings on March 29 and April 12, 2022, JupyterLab does not have published timelines for when a version ends support for bug fixes or for security fixes. 

### Proposed Solution

Provide a matrix in the documentation, hosted on the Jupyter web site, that indicates a date after which users should not expect bug fixes or security fixes. The latter should be considered the end of life (EOL) date for that version.

### Additional context

Node.js's releases document includes a table and a visualization of several recent releases, as well as a brief text summary of what the terms mean. It does an excellent job of communicating expectations concisely. https://nodejs.org/en/about/releases/

@Carreau has been working on a similar approach for Scientific Python: https://scientific-python.org/ 

#8577 concerns support for Python versions, each of which also has an "end of life"

In the past, we have discussed an "N−1" (N minus 1) strategy for support: when Notebook 8 is released, for example, Notebook 7 moves into maintenance mode and Notebook 6 reaches EOL. This is less effective at communicating urgency to upgrade from a previous version because we have no regular cadence for releasing major versions of JupyterLab.

JasonWeill:
## References

Fixes #12397.

## User-facing changes

Per @jasongrout 's [comment](https://github.com/jupyterlab/jupyterlab/issues/12397#issuecomment-1414393697) in #12397, we will support a major version of JupyterLab for 1 year after its successor's first release. In the case of JupyterLab 3, JupyterLab 4.0.0 was released on May 15, 2023, so JupyterLab 3 will no longer be maintained after May 15, 2024.


JasonWeill:
@jtpio Sounds good — I just changed this PR to be a draft. I've emailed the council and, later, I opened the team-compass issue linked above.

docs/source/developer/contributing.rst
```diff
@@ -86,9 +86,10 @@ breaking changes. Consider documenting your maintenance plans to users in these
 You may also wish to consider pinning the major version of JupyterLab when developing
 extensions (in your package metadata).
 
-We maintain the **two most recently released major versions of JupyterLab**,
-JupyterLab v3 and JupyterLab v4. JupyterLab v1 and v2 are no longer maintained.
-All JupyterLab v2 users are strongly advised to upgrade as soon as possible.
+We maintain a major version of JupyterLab for **one year after its successor's first release**.
+JupyterLab v4 was released on May 15, 2023, so JupyterLab v3 will be maintained
+until May 15, 2024. JupyterLab v1 and v2 are no longer maintained.
+All JupyterLab v2 and v3 users are strongly advised to upgrade as soon as possible.
```
krassowski, line 92: Thank you! Is it worth also linking to the lifecycle document for details?

Issue #15569, opened 2023-12-27
Benchmark tests triggered on comment


brichet:
## Description

The benchmark tests are triggered every time a comment is added in a PR.

This may be related to https://github.com/jupyterlab/jupyterlab/pull/15523.

### Example: 

<img src=https://github.com/jupyterlab/jupyterlab/assets/32258950/e2e3dfba-789f-4eee-bf4c-b0e5a94181d7>

https://github.com/jupyterlab/jupyterlab/actions/runs/7337248858

## Reproduce

Write a comment in a PR.

## Expected behavior

Do not trigger the tests for each comment.

JasonWeill:
I have also been seeing this today during triage: every comment I add triggers a workflow, which causes me to get email with the results of the flow.

PR #15630, opened 2024-01-12
Fix syntax error in benchmark workflow

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15569

Follow up to https://github.com/jupyterlab/jupyterlab/pull/15523

## Code changes

remove spurious `(`

## User-facing changes

None

None


jtpio (approved):
Thanks!

Issue #15595, opened 2024-01-02
Clicking away from rendered markdown cell unrenders it


krassowski:
## Description

When an active cell is a markdown cell and user clicks on a different cell, the markdown cell unrenders causing content jump thus distraction and annoyance.

## Reproduce

![clicking-away-unrenders-header](https://github.com/jupyterlab/jupyterlab/assets/5832902/3b2fa13d-68df-4e3f-a1dc-7a058ab1fcea)

Interestingly, this does not happen when:
- changing the active cell with arrow keys
- clicking outside of the notebook or on its toolbar

## Expected behavior

Clicking on another cell does not unrender the previous cell.

## Context

- Browser and version: Chrome
- JupyterLab version: 4.1.0b0

Likely a regression in 4.1.0beta but Binder does not start for me at the moment so I cannot confirm.

krassowski:
Confirmed I cannot see this in 4.0.9 nor 3.6.x.

JasonWeill:
Is this an issue in 4.0.10 as well?

fcollonval:
It may be a consequence of https://github.com/jupyterlab/jupyterlab/pull/15408

My guess, the focus change switches the notebook to edit mode before the active cell index is updated.

afonit:
> Is this an issue in 4.0.10 as well?

I do not see this behavior on 4.0.10

![example_001](https://github.com/jupyterlab/jupyterlab/assets/1798497/8667498c-9755-4595-80e7-8a7880396b5c)

![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/b9372917-008d-42d8-ac44-2cef0839aea3)

PR #15634, opened 2024-01-13
Fix markdown getting un-rendered when focusing on a different cell

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15595

## Code changes

- Add regression unit test
- Move change of active index ahead of mode enforcement as it assumes the active index is already set

## User-facing changes

| Before | After |
|--|--|
| before | after |

None


jtpio (approved):
Thanks!

krassowski:
Thanks! Merging - the visual docs failure is unrelated (https://github.com/jupyterlab/jupyterlab/issues/15635).

Issue #15636, opened 2024-01-13
Opening "Keyboard Shortcuts" UI breaks custom shortcuts, hundreds "Skipping this shortcut" warnings in console


ManPython:
## Description
This is what have in console: - check Browser Output

## Reproduce

On url: /tree?
Also: /notebooks/test/Untitled.ipynb

## Context

https://github.com/jupyterlab/jupyterlab/issues/15535#issuecomment-1890625204

- Operating System and version: W11
- Browser and version: FF120
- JupyterLab version: 
```
jupyter --version
Selected Jupyter core packages...
IPython          : 8.20.0
ipykernel        : 6.28.0
ipywidgets       : 8.1.1
jupyter_client   : 8.6.0
jupyter_core     : 5.7.1
jupyter_server   : 2.12.4
jupyterlab       : 4.0.10
nbclient         : 0.9.0
nbconvert        : 7.14.1
nbformat         : 5.9.2
notebook         : 7.0.6
qtconsole        : 5.5.1
traitlets        : 5.14.1
```

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
```

Skipping this shortcut because there are no actionable keys on this platform 
Object { command: "application:activate-next-tab", keys: (1) […], winKeys: [], macKeys: [], linuxKeys: [], selector: "body", args: {} }
5649.bundle.js:1:14978
Skipping this shortcut because there are no actionable keys on this platform 
Object { command: "kernelmenu:interrupt", keys: (2) […], winKeys: [], macKeys: [], linuxKeys: [], selector: "[data-jp-kernel-user]:focus", args: {} }
5649.bundle.js:1:14978
Skipping this shortcut because there are no actionable keys on this platform 
Object { command: "kernelmenu:restart", keys: (2) […], winKeys: [], macKeys: [], linuxKeys: [], selector: "[data-jp-kernel-user]:focus", args: {} }
5649.bundle.js:1:14978
Skipping this shortcut because there are no actionable keys on this platform 
[...]

Issue #15599, opened 2024-01-03
Kernel-related shortcuts (e.g. restart, interrupt) do not work anymore


krassowski:
## Description

Shortcuts such as <kbd>I</kbd> <kbd>I</kbd> (interrupt) or <kbd>0</kbd> <kbd>0</kbd> do not work in 4.1.0b0. Manually selecting the command from menu or command palette works. My custom shortcut (<kbd>ctrl</kbd> + <kbd>R</kbd>) for "restart and run up to selected cell" does not work anymore either.

This is likely  an issue with the selector (`[data-jp-kernel-user]:focus`) due to changes introduced in https://github.com/jupyterlab/jupyterlab/pull/14115 (CC @gabalafou @brichet).

## Reproduce

![shortcuts-brokendoes-not-work](https://github.com/jupyterlab/jupyterlab/assets/5832902/5a7c8335-b1a1-46ce-93c3-368bc873f503)

## Expected behavior

- Default shortcuts for kernel actions work
  - possibly adding a space before `:focus` would be sufficient to fix this
- Shortcuts configured by users are ported over automatically or there should be a user-facing entry in the changelog documenting how/which shortcuts will need to be reconfigured (e.g. by updating selectors)

## Context

- JupyterLab version: 4.1.0b0

krassowski:
Also, it is worth checking if there are any other selectors with `:focus` touching the notebook node selected by data attributes. This will potentially affect extensions too; for known selectors we could indeed swap them in settings transparently but for shortcuts added in extensions it would be harder; so two other things we should do:
- [ ] check the extensions in `jupyterlab` and `jupyterlab-contrib` for `:focus` containing selectors in schema
- [ ] ensure there is an entry in extension migration guide explaining the changes which extension authors may be needed to make

brichet:
> * possibly adding a space before `:focus` would be sufficient to fix this

This would make some characters unusable in the cell input (like "i" in this case).
The selector `"[data-jp-kernel-user] :focus:not(:read-write)"` should avoid it.

brichet:
> * check the extensions in `jupyterlab`

Looking for `:focus` selector in jupyterlab, I found that https://github.com/jupyterlab/jupyterlab/blob/main/examples/notebook/src/commands.ts has not been updated with the correct selectors.

For example: 
[...]

Issue #12294, opened 2022-03-29
Setting editor has incorrect defaults


mlucool:
## Description

The settings editor's "System Defaults" has the incorrect default types for some keys. Instead of arrays, it defaults to objects.

## Reproduce

1. Open the JSON settings editor
2. Copy/paste a shortcut from "Systems Defaults" to "User Preferences"
3. Save

You'll see type errors on `winKeys`, `macKeys`, and `linuxKeys`:
![image](https://user-images.githubusercontent.com/1813603/160674537-6f4311df-693d-443e-a016-2314dc2d293c.png)

## Expected behavior

Copy/paste from system defaults to user preferences just works.

## Context
Lab@3.3.0

jweill-aws:
Triage notes: @krassowski can provide more context for where this occurs in code

krassowski:
The `winKeys`, `macKeys` and `linuxKeys` are handled by lumino; these are described in JupyterLab by:

https://github.com/jupyterlab/jupyterlab/blob/83b2ae958f86b8a6925b9e0c36adb7b897bff62b/packages/shortcuts-extension/schema/shortcuts.json#L26-L37

and:

https://github.com/jupyterlab/jupyterlab/blob/83b2ae958f86b8a6925b9e0c36adb7b897bff62b/packages/settingregistry/src/plugin-schema.json#L228-L250

The bug must be somewhere in the normalization code or JSON schema population, which I am looking at right now as an offshoot from fixing a 4.1 release blocker (https://github.com/jupyterlab/jupyterlab/issues/15636#issuecomment-1890991475).

The population of the defaults happens in:

https://github.com/jupyterlab/jupyterlab/blob/83b2ae958f86b8a6925b9e0c36adb7b897bff62b/packages/settingregistry/src/settingregistry.ts#L1438-L1496

krassowski:
Note to self: rather than having an empty array on the defaults side we want to avoid populating these keys at all because an empty array means that the keybinding is disabled on the specific OS.

Issue #12056, opened 2022-02-15
Opening keyboard shortcuts UI result in "destruction" of shortcut settings


krassowski:
## Description

Opening the newly incorporated keyboard shortcuts settings UI results in dumping of JSON for all possible shortcuts, making it impossible to navigate manually and distinguish between default and user-defined settings.

This is a notable regression as carefully crafted JSON settings of advanced users will be destroyed  (along with any comments they may have written) when they open keyboard shortcuts UI, without touching a single button.

## Reproduce

1. Open settings UI
2. Click Keyboard Shortcuts
3. Wait 2-3 seconds to see that "Restore to defaults" shows up automatically
4. Open JSON Settings Editor (Advanced Settings Editor), see that all default shortcuts were dumped as a big JSON blob:
    ![Screenshot from 2022-02-15 21-10-10](https://user-images.githubusercontent.com/5832902/154149812-a17c282b-c9fe-4875-8b9a-bdabbeba0f18.png)
5. Close and reopen the shortcuts UI, see that all shortcuts are incorrectly shown as "Custom" making it impossible to understand which shortcuts were actually modified.
    ![Screenshot from 2022-02-15 21-08-19](https://user-images.githubusercontent.com/5832902/154149918-97240d71-f76b-4c94-b92e-d89b36af238f.png)

## Expected behavior

1. Only shortcuts actually modified by user are added to the right panel
2. No change in shortcuts JSON is performed without user interaction

## Context

- Browser and version: Chrome
- JupyterLab version: 3.3

fcollonval:
I looked to this one. The trouble gonna arise for all settings that are merging user settings with the default one. We probably needs to extend the API of the `IFormComponentRegistry` to be able to add a transformer from the raw form data to the one that need to be saved. This would reflect the transformation capability for loading the settings.

fcollonval:
This is definitely a blocker for 3.3.0

pinging for opinions @krassowski @jtpio @jasongrout

jtpio:
> to extend the API of the IFormComponentRegistry to be able to add a transformer

Probably if `IFormComponentRegistry` is new in 3.3 (added in https://github.com/jupyterlab/jupyterlab/pull/11079 it seems), then it should be fine to change its API during the beta phase.

krassowski:
Reopening as it is still happening in 4.0.10 and 4.1.0 beta; in 4.1.0 beta it even prevents user from changing any shortcuts at due to #15636!

PR #15639, opened 2024-01-14
Fix kernel shortcuts, add migration, fix defaults population 

krassowski:
## References

- fixes https://github.com/jupyterlab/jupyterlab/issues/15599
- fixes https://github.com/jupyterlab/jupyterlab/issues/12056
- fixes https://github.com/jupyterlab/jupyterlab/issues/15636
- fixes https://github.com/jupyterlab/jupyterlab/issues/12294

## Code changes

- update selectors following change of focused HTML node in command mode from notebook node to cell in a PR that fixed tab traps ([#14115](https://github.com/jupyterlab/jupyterlab/pull/14115))
- migrate user-defined and extension-defined selectors to the new ones transparently
- fix `winKeys`, `linuxKeys` and `macKeys` incorrectly popping up in defaults and leading to corruption of shortcuts, which fixes configuring shortcuts and user shortcuts being incorrectly populated
- fix representation of defaults in the user panel
- add a test ensuring that merely opening the keyboard shortcut settings does not mangle the user-defined settings JSON blob

#### JSON setting editor defaults now show up correctly

Shortcuts (optional linux/mac/win keys no longer incorrectly shown as objects)

| Before | After |
|--|--|
|Screenshot from 2024-01-14 19-35-21 | Screenshot from 2024-01-14 19-36-01 |

Context menu

| Before | After |
|--|--|
|Screenshot from 2024-01-14 19-30-39 | image |

HTML sanitizer

| Before | After |
|--|--|
| Screenshot from 2024-01-14 19-33-01 | Screenshot from 2024-01-14 19-32-53 |

None

### DIFF
--- a/packages/settingregistry/src/settingregistry.ts
+++ b/packages/settingregistry/src/settingregistry.ts
@@ -1524,6 +1524,11 @@ namespace Private {
   }
 
   /**
+   * Selectors which were previously warned about.
+   */
+  const selectorsAlreadyWarnedAbout = new Set<string>();
+
+  /**
    * Upgrade shortcuts to ensure no breaking changes between minor versions.
    */
   export function upgradeShortcuts(
@@ -1556,11 +1561,15 @@ namespace Private {
       const oldSelector = shortcut.selector;
       let newSelector = oldSelector;
       for (const change of changes) {
-        if (oldSelector.includes(change.old)) {
+        if (
+          oldSelector.includes(change.old) &&
+          !selectorsAlreadyWarnedAbout.has(oldSelector)
+        ) {
           newSelector = oldSelector.replace(change.old, change.new);
           selectorDeprecationWarnings.add(
-            `"${change.old}" was replaced with "${change.new}" in ${change.versionDeprecated}`
+            `"${change.old}" was replaced with "${change.new}" in ${change.versionDeprecated} (present in "${oldSelector}")`
           );
+          selectorsAlreadyWarnedAbout.add(oldSelector);
         }
       }
       shortcut.selector = newSelector;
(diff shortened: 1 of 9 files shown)


packages/settingregistry/src/settingregistry.ts
```diff
@@ -1478,21 +1498,81 @@ namespace Private {
...
+      }
+      shortcut.selector = newSelector;
+      return shortcut;
+    });
+    if (selectorDeprecationWarnings.size > 0) {
+      console.warn(
+        'Deprecated shortcut selectors: ' +
+          [...selectorDeprecationWarnings].concat('\n') +
+          '\n\nThe selectors will be substituted transparently this time, but need to be updated at source before next major release.'
+      );
```
brichet, line 1574: It should be nice to have a way to warn only once, when all the plugins are loaded.
This is not for this PR, but having a way to concatenate all messages from plugin loading should be nice, to avoid duplication.
For example, if we restore only one selector from `plugin.json`, we receive 16 warnings in the console.

brichet (commented):
Thanks @krassowski, I only have some comments below, otherwise it looks good to me.

packages/settingregistry/src/settingregistry.ts
```diff
@@ -1457,14 +1468,23 @@ namespace Private {
...
-          definitions
+          definitions,
+          schema.required instanceof Array &&
+            schema.required?.includes(property as any)
         );
       }
 
       return result;
     } else if (schema.type === 'array') {
+      const defaultUndefined = typeof schema.default !== 'undefined';
```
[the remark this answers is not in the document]
krassowski: Absolutely, good catch - thanks!

packages/settingregistry/src/settingregistry.ts
krassowski: Indeed. We could use some global variable in `Private` to de-duplicate; let me take a look.

brichet (approved):

brichet:
Thanks @krassowski

Issue #15660, opened 2024-01-18
Mermaid.js state diagrams do not always render in Windows


afonit:
I have had some mermaid diagrams work, but many don't.

I setup a new environment to test the beta:
```bash
conda create -n lab_b1
conda activate lab_b1
conda install -c conda-forge/label/jupyterlab_beta -c conda-forge jupyterlab
jupyterlab 
```

I was working through some examples on https://mermaid.js.org/ and noticed a lot of them don't work in jupyterlab.

## Example 1

On this page:
https://mermaid.js.org/syntax/stateDiagram.html

This is what that site shows:
![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/ecc153eb-6c8f-4cb3-8d5b-e3cb9abb6bce)

Here is the code:
```
---
title: Simple sample
---
stateDiagram-v2
    [*] --> Still
    Still --> [*]

    Still --> Moving
    Moving --> Still
    Moving --> Crash
    Crash --> [*]
```

Here is what jupyterlab renders:
![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/f3210b1c-89cf-4cf4-8fe0-5ac6e673c8f3)

## Example 2

On this page:
https://mermaid.js.org/syntax/classDiagram.html
I see this:
![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/383fd85f-8fcf-482d-b346-57448c9d7dfe)

Here is the code:
```
---
title: Animal example
---
classDiagram
    note "From Duck till Zebra"
    Animal <|-- Duck
    note for Duck "can fly\ncan swim\ncan dive\ncan help in debugging"
    Animal <|-- Fish
    Animal <|-- Zebra
    Animal : +int age
    Animal : +String gender
    Animal: +isMammal()
    Animal: +mate()
    class Duck{
        +String beakColor
        +swim()
        +quack()
    }
    class Fish{
        -int sizeInFeet
        -canEat()
    }
    class Zebra{
        +bool is_wild
        +run()
    }
```

This is what jupyterlab renders:
![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/d14e3633-a7ed-4684-8eee-23d940a17db4)

## Example 3

On this page:
https://mermaid.js.org/syntax/sequenceDiagram.html
I see this:
![image](https://github.com/jupyterlab/jupyterlab/assets/1798497/01deeffd-1b06-49e8-af3e-0365bef984ce)

Here is the code:
```
[...]

PR #15661, opened 2024-01-18
Ensure void elements have closing slash in mermaid svg

bollwyvl:
## References

- starts #15660

## Code changes

- [x] transform `<br>` tags into closed `<br/>` when generating the `img` tag containing mermaid SVG
  - > the regular expression will actually find [all void elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element), and preserves attributes
- [x] test
  - [x] add a test/notebook with examples of all (but one) of the upstream documentation examples
  - [x] add screenshots

## User-facing changes

- users will be able to use expressive more typographic features in mermaid diagrams

## Security/Compatibility considerations
- some implementations (e.g. GitHub) will _not even parse_ diagrams with any void elements (or probably _any_ fancy HTML) _except_ `hr` and `br`
  - currently, in this PR, as in the [mermaid playground](http://mermaid.live), one can do really interesting/outrageous things, for example:
    - the `style="border-color: red"` attribute can be used on an `hr` to change its color
    - `<input type="checkbox" />` and `radio` can be used to do weird things
  - in the playground, these are actually clickable, etc. but sadly not feasible with the (otherwise robust and portable) `` approach
    - though text labels _may_ contain further `img` tags
- no implementation render `<script>` tags
- `onclick="javascript:alert('pwnd')"` attributes _do_ appear to pass through on the playground, and in `4.1.0b1` **if** a user goes through _right-click, open in new tab_ and then interacts with the diagram

- if we decide to do more aggressive sanitizing, this could break some diagrams/examples in the wild


krassowski:
Upstream issue: https://github.com/mermaid-js/mermaid/issues/1766

What effect the suggested `flowchart.htmlLabels=false` have on compatibility?

bollwyvl:
Yeah, there's that one (not sure how breaking). I need to take a look at
various other control mechanisms. There's also a sandbox mode, which is
probably worth investigating, though it does target an iframe with html
srcdoc instead of SVG, so does more weird stuff to entities in
foreignobjects. Ending up with an img tag is still the most portable output.

bollwyvl:
> htmlLabels=false

This does not solve the problem for `\n` in node labels.

> `sandbox`

This resulted in strictly _worse_ results e.g. `&gt;br&lt;` as these are "magic" entities, not defined in the SVG spec.

The `<iframe srcdoc>` approach _is_ very attractive, as, for example, HTML (and therefore markdown) images can do icky things like:

```md
This is totally valid markdown:

<img src="/api/shutdown" alt="shutdown"/>
``` 

And pretty much all frontends will render it, including (at least in preview) GitHub

> This is totally valid markdown:
> 
> shutdown

But `marked` communicates over strings, and filters out `iframe`s... we probably _don't_ want to change that posture, today, as they can very much be used for deceptive things, not _just_ their "white hat" security techniques).

krassowski:
> But `marked` communicates over strings, and filters out `iframe`s... we probably don't want to change that posture, today, as they can very much be used for deceptive things, not just their "white hat" security techniques).

Could we use the same splicing approach as we do with MathJax? See:

https://github.com/jupyterlab/jupyterlab/blob/e38e0b8d59b8cba1d559040236a142931c17c42b/packages/rendermime/src/renderers.ts#L323-L330

krassowski:
Kicking this PR to check if the labeller for mermaid now works (https://github.com/jupyterlab/jupyterlab/pull/15717), also FYI once you merge/rebase the tests for mermaid should now run on CI.

bollwyvl:
Thanks. Alas, it might be a tick before I can get back to this...

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts-snapshots/mermaid-diagram-dark-09-requirement-jupyterlab-linux.png
krassowski, line 1: This looks amiss? In general I am not a fan of that many snapshots but I am ok with these being here as long as they are isolated to the diagrams alone (no other UI).

packages/mermaid/src/manager.ts
```diff
@@ -81,7 +81,9 @@ export class MermaidManager implements IMermaidManager {
     const el = document.createElement('div');
     document.body.appendChild(el);
     try {
-      const { svg } = await _mermaid.render(id, text, el);
+      let { svg } = await _mermaid.render(id, text, el);
+      svg = svg.replace(Private.RE_VOID_ELEMENT, Private.replaceVoidElement);
+
```
krassowski, line 86: I would love to see this unit tested. For example input `'<br/>'` (which I understand we do not expect from the current version of marked but I guess user or future versions could introduce it) results in `'<br / >'` which strictly speaking is invalid XHTML. Basically:

```js
new DOMParser().parseFromString('<br>'.replace(RE_VOID_ELEMENT, replaceVoidElement), 'image/svg+xml')
```

works, but:

```js
new DOMParser().parseFromString('<br/>'.replace(RE_VOID_ELEMENT, replaceVoidElement), 'image/svg+xml')
```

fails.

krassowski (changes requested):

bollwyvl:
Ok, all of the screenshots are now working (and less sensitive to scroll skew). Will merge upstream, and see if anything pops out, but no further work planned.

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts-snapshots/mermaid-diagram-dark-09-requirement-jupyterlab-linux.png
bollwyvl: Of _all_ the things a screenshot is good at, _diagram renders right_ seems like a legit use case vs all the "whole screen" shots that are sensitive to individual pixel skew. 

At any rate, these are now as isolated as they can be... the "screenshot element" function doesn't handle bigger-than-viewport stuff very well, and there are two examples diagrams that are larger than the viewport.

packages/mermaid/src/manager.ts
bollwyvl: Sure, will add some coverage. 

I'm semi-disappointed we _still_ don't have any [property-based testing](https://github.com/dubzzz/fast-check) or overarching [frontend coverage](https://github.com/istanbuljs/istanbuljs) that combines `jest` and `puppeteer` outputs, but certainly won't be wading into the stack any more than i have to on this PR.

packages/mermaid/test/clean.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+describe('@jupyterlab/mermaid', () => {
+  for (const element of VOID_ELEMENTS) {
+    it(`should clean void element ${element}`, () => {
+      const variants = makeElementVariants(element);
+      const raw = [...SVG_XHTML_HEADER, ...variants, ...SVG_XHTML_FOOTER].join(
+        '\n'
+      );
+
+      const clean = MermaidManager.cleanMermaidSvg(raw);
+      const parsed = parser.parseFromString(clean, SVG_MIME);
```
bollwyvl, line 25: This would now error if the cleaned string is not valid XML... it's harder to verify that the browser would go into the "fallback" rendering mode (the most likely failure case).

Because of that, it's probably _not_ worth roundtripping inside the implementation, which is already pretty heavy.

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts-snapshots/mermaid-diagram-dark-09-requirement-jupyterlab-linux.png
krassowski: I pushed a fix for this using workaround from https://github.com/microsoft/playwright/issues/13486/#issuecomment-1112012053 (an upstream fix is apparently not planned).

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  'sankey',
+  'xy'
+];
+
+/**
+ * Workaround for playwright not handling screenshots
+ * for elements larger than viewport, derived from:
+ * https://github.com/microsoft/playwright/issues/13486#issuecomment-1112012053
+ */
+async function resizePageAndScreenshot(locator: Locator) {
```
bollwyvl, line 33: Perhaps this is something that should be hoisted up outside of a deeply-nested `utils`?

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts-snapshots/mermaid-diagram-dark-09-requirement-jupyterlab-linux.png
bollwyvl: Thanks for the fix! Pretty weird its a non-priority upstream, as this hardly seems to that unique. As commented on the new file, perhaps this belongs someplace more prominent/easier to find?

bollwyvl:
hooray all :green_circle:

galata/test/jupyterlab/notebook-mermaid-diagrams.test.ts
krassowski: Yes, we could add it as a galata helper in the future if it turns out useful elsewhere too. For now it's easier to do it here because this way we don't need to worry about it becoming a public API yet ;)

krassowski (approved):
Thank you @bollwyvl!

bollwyvl:
Thanks. Alas, the wheels turn: since this was merged, https://github.com/mermaid-js/mermaid/releases/tag/v10.8.0 :crying_cat_face: 

I don't see any hard blockers/features that would _require_ going out, but will take a look, especially to kick the tires on the screenshot approach.

PR #15648, opened 2024-01-16
User-facing changelog for 4.1

krassowski:
## References

Closes #15576

For review: https://jupyterlab--15648.org.readthedocs.build/en/15648/getting_started/changelog.html#v4-1

- [ ] After approval but before merging the links to images need substituting `krassowski/jupyterlab/changelog-4.1` → `jupyterlab/jupyterlab/main`

## Code changes

None

## User-facing changes

Changelog for upcoming 4.1 has high-level overview of the highlights

## Backwards-incompatible changes

None

PR #15669, opened 2024-01-20
Mention jupyter-ai pre-release in inline completer changelog section

krassowski:
## References

Follow up to #15648 

## Code changes

None

## User-facing changes

Mention the jupyter-ai pre-release which supports inline completer to get more user testing.

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,12 @@ The default keyboard shortcuts are displayed in the small widget shown when hove
 
 To enable the inline suggestions based on the kernel history, go to Settings → Inline Completer → History provider → check the "enabled" checkbox.
 
+In addition to the built-in history suggestions,
+the [`jupyter-ai`](https://github.com/jupyterlab/jupyter-ai) extension will provide
+suggestions from supported models.
+The `jupyter-ai` integration with inline completer can be tested starting with the
+[v2.10.0beta0](https://github.com/jupyterlab/jupyter-ai/releases/tag/v2.10.0beta0) pre-release.
+
 The Inline Completer API is still considered experimental and may be subject to changes, please share feedback!
 
 ### Keyboard navigation improvements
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

fcollonval (approved):

CHANGELOG.md
```diff
@@ -62,6 +62,12 @@ The default keyboard shortcuts are displayed in the small widget shown when hove
 
 To enable the inline suggestions based on the kernel history, go to Settings → Inline Completer → History provider → check the "enabled" checkbox.
```
dhirschfeld, line 63: I can't see `Inline Completer` - am I missing something?

image

krassowski: I cannot see code completer section on your screenshot either. Can you choose and reopenn the settings panel? There may be a glitch on intital load .

dhirschfeld: Hmm, settings now looks like:

image

dhirschfeld: Maybe I updated without closing down my JupyterLab tab? :thinking:

Unless you get flooded with other questions about missing settings , it's probably just user error - sorry for the noise!

krassowski: Sorry, I do not follow - your second screenshot looks correct to me, isn't it?

krassowski: Or is it about jupyter-ai not showing up in there? Which version do you have installed?

dhirschfeld: Yes, the 2nd screenshot looks correct so I'm not sure why it didn't look like that before. I was wondering if I hadn't closed down my JupyterLab tab when installing 4.1.0rc0 and so initially wasn't seeing the updated UI.

After taking a look this morning, after a reboot, it looks fine so I think it's likely user error and not an issue with JupyterLab.

dhirschfeld: I have `jupyter-ai-core` 2.9.1 installed:

image

SO, IIUC I need to wait for 2.10 to be able to use it for completions.

Issue #15672, opened 2024-01-21
Locked plugin migration logic should handle config levels/permission errors better


krassowski:
## Description

Some users are getting startup error in JupyterLab 4.1.0b2 and earlier:

```
Traceback (most recent call last):
  File "/home/jj/.local/bin/jupyter-lab", line 5, in <module>
    from jupyterlab.labapp import main
  File "/home/jj/.local/lib/python3.11/site-packages/jupyterlab/labapp.py", line 101, in <module>
    app_version = get_app_version()
                  ^^^^^^^^^^^^^^^^^
  File "/home/jj/.local/lib/python3.11/site-packages/jupyterlab/commands.py", line 599, in get_app_version
    handler = _AppHandler(app_options)
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jj/.local/lib/python3.11/site-packages/jupyterlab/commands.py", line 642, in __init__
    self._maybe_mirror_disabled_in_locked()
  File "/home/jj/.local/lib/python3.11/site-packages/jupyterlab/commands.py", line 1141, in _maybe_mirror_disabled_in_locked
    write_page_config(page_config, level=level)
  File "/home/jj/.local/lib/python3.11/site-packages/jupyterlab_server/config.py", line 213, in write_page_config
    cm.set("page_config", page_config)  # type:ignore[no-untyped-call]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jj/.local/lib/python3.11/site-packages/jupyter_server/services/config/manager.py", line 35, in set
    return self.write_config_manager.set(section_name, data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jj/.local/lib/python3.11/site-packages/jupyter_server/config_manager.py", line 112, in set
    self.ensure_config_dir_exists()
  File "/home/jj/.local/lib/python3.11/site-packages/jupyter_server/config_manager.py", line 66, in ensure_config_dir_exists
    os.makedirs(self.config_dir, 0o755)
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 225, in makedirs
PermissionError: [Errno 13] Permission denied: '/usr/etc/jupyter
```

Originally reported on https://discourse.jupyter.org/t/jupyterlab-4-1-beta-call-for-testing/23529/4

## Reproduce

Not sure.

[...]

PR #15678, opened 2024-01-22
Migrate to `user` level or to none (as fallback)

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15672

## Code changes

- catch exceptions in migration
- fallback to `user` level
- and then fallback to no migration at all

## User-facing changes

JupyterLab will not error out on startup for users who installed jupyterlab outside of a virtual environment using `sudo` (or in a similar setup where `sys.prefix` is read-only).

None


jtpio (approved):
Thanks!

Issue #15611, opened 2024-01-06
Should the "Click to add a cell" button be excluded from tab navigation?


krassowski:
## Description

JupyterLab 4.1 includes improved tab-navigation (https://github.com/jupyterlab/jupyterlab/pull/14115). However, I find one aspect particularly annoying: whenever I press <kbd>Tab</kbd> in command mode (sometimes accidental, because I did not realise I was not in the edit mode) the notebook scrolls to the very end because the focus jumps to "Click to add a cell" which causes distraction and annoyance.
- when I am in the middle of the notebook I almost never want to add a cell at the end of the notebook
- it feels like the "Click to add a cell" button should only be navigable with up/down as all cells are because in essence it is a cell placeholder

## Reproduce

Either:
1. Enter command mode in a cell somewhere in the beginning or a middle of a notebook
2. Press <kbd>Tab</kbd>
3. See you are in a cell toolbar (ok)
4. Press <kbd>Tab</kbd> again
5. See the notebook scrolled away to the bottom
6. If you worked on two notebooks side-by-side be annoyed that you cannot switch to the other notebook with keyboard only without loosing the cell you were working out of view; even in a single notebook be annoyed that you lost the context and if you pressed Tab one more time no number of  <kbd>Shift</kbd>  + <kbd>Tab</kbd> will land you back at the cell you were working on

![with-cell-toolbar](https://github.com/jupyterlab/jupyterlab/assets/5832902/6afda545-b2e4-4501-8c39-a0d51181b751)

Even worse:
1. Disable cell toolbar with `jupyter labextension disable @jupyterlab/cell-toolbar-extension`
2. Repeat the instructions above but it happens on first Tab making it even worse

![without-cell-toolbar](https://github.com/jupyterlab/jupyterlab/assets/5832902/8f00f21d-80fd-4802-8488-b122637f029f)

## Expected behavior

The "add cell" button is excluded from tab navigation and instead navigable with up/down arrow keys and j/k keys (as it is).

## Context

- JupyterLab version: 4.1.0b0

PR #15679, opened 2024-01-22
Remove the add cell button from the tabbable item

brichet:
This PR set the "add cell" button (in the notebook footer) as not reachable using tab key.

The button should now be reached using arrow keys only, when navigating between the cells of the notebook.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15611

## Code changes

Add attribute `tabindex=-1` to the "add cell" button.

## User-facing changes

The button is not reached anymore using tab key, to avoid scrolling to the bottom of the notebook unintentionally.

None


brichet:
Failure is not related, it seems to fail on other PRs too.

krassowski (approved):
Thank you @brichet!

PR #15578, opened 2023-12-29
Backport PR #15524: Fix visual tests

krassowski:
Backport PR #15524 on branch 4.0.x (Fix visual tests, <s>allow to disable termianal cursor blinking</s>)

PR #15691, opened 2024-01-24
Backport PR #15262 on branch 4.0.x (Fix connection loop issue with standalone foreign document in LSP)

meeseeksmachine:
Backport PR #15262: Fix connection loop issue with standalone foreign document in LSP


krassowski:
@JasonWeill CI failures are unrelated, it looks like something in the runner's environment changed leading failures on `hatchling` level, see https://github.com/jupyterlab/jupyterlab/issues/15692; we can restart once this is fixed.

krassowski:
Restarting CI after new hatchling release

JasonWeill:

krassowski:
@JasonWeill there should be no snapshot changes for this PR. The updates for snapshots are basically flaky runs showing up. There is a PR https://github.com/jupyterlab/jupyterlab/pull/15578 backporting fixes for this but it still needs more work.

JasonWeill:
@krassowski Reverted; sorry about the confusion

JasonWeill:
@krassowski Should I wait until #15578 is merged before I merge this change?

krassowski:
No, but I would like to test it with `jupyterlab-lsp` extension to ensure that users would not be affected by this change; it is likely that some mitigation may be needed.

krassowski (approved):
Thanks. It looks like this does not break jupyterlab-lsp further than it was broken by earlier backports. It might break source installs but I will deal with it on my side.

Issue #15670, opened 2024-01-20
"Notebook scrolls to heading" test fails subtly


krassowski:
Since recent fixes in ToC this test: "Table of Contents scrolling to heading › Notebook scrolls to heading" fails by positioning the heading slightly off:

| Actual | Expected |
|--|--|
| ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/0f60c0ce-4d3e-47f1-8717-d996215a022b) |  ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/e5e7d827-01a8-4e24-be72-a6864e92e5d6) |

krassowski:
This appears unrelated to recent security fixes to ToC implementation and indeed was already failing before these were merged as seen in https://github.com/jupyterlab/jupyterlab/pull/15375.

krassowski:
It is related to them after all. For context, this snapshot was updated in https://github.com/jupyterlab/jupyterlab/pull/15386 which changed the behaviour from aligning heading to bottom to aligning to the top (i.e. what users were expecting). However, the scrolling logic was not using the per-heading scrolling at the time (due to a bug in heading retrieval) and instead using cell scrolling fallback logic, see:

https://github.com/jupyterlab/jupyterlab/blob/85445cc40913ec6a32cab100bc1132306b5075c5/packages/notebook/src/toc.ts#L593-L616

Once a fix for https://github.com/advisories/GHSA-4m77-cmpx-vjc4 was merged, it also fixed the bug which was leading the execution into the fallback branch.

The cell scrolling logic aligns to the cell so it gave us a nice padding above the cell. But there may be multiple headings in a cell.

PR #15695, opened 2024-01-25
Fix shortcut UI failing on filtering when empty command is given

### LINKED ISSUE #15687: Keyboard shortcuts settings page breaks on search
## Description
Searching anything in the keyboard shortcuts settings page doesn't work. Here's a gif - 

settings

## Reproduce

Go to the keyboard shortcuts settings page and search something.

krassowski:
## References

Fixes #15687

## Code changes

- fix typing in `ShortcutUI.tsx` (use proper types instead of `any`)
- correct a name of private method to follow underscore convention
- fix the issue by guarding against incorrect access to second item of one-element array created by `split()` on an empty string

## User-facing changes

1. Filtering shortcuts works in Shortcuts UI even when an extension provides incorrect but technically allowed shortcut with empty command (see jupyterlab-vim).
2. `(Command label missing)` placeholder is shown when a label for a command cannot be retrieved
    - [ ] I will also open a PR against jupyterlab-vim to add a dummy command so that user does not see it as first thing in the shortcuts table

image

None

### DIFF
--- a/packages/shortcuts-extension/src/components/ShortcutUI.tsx
+++ b/packages/shortcuts-extension/src/components/ShortcutUI.tsx
@@ -7,7 +7,11 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 
 import { ArrayExt, StringExt } from '@lumino/algorithm';
 
+import { CommandRegistry } from '@lumino/commands';
+
 import { ReadonlyJSONArray } from '@lumino/coreutils';
+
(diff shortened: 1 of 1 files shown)


krassowski:
The single visual regression failure is unrelated (tracked in https://github.com/jupyterlab/jupyterlab/issues/15670)

jtpio (approved):
Thanks!

jtpio:

PR #15702, opened 2024-01-26
Fix scrolling on execution and switching notebook mode

### LINKED ISSUE #15693: Erratic scroll jumps when performing actions in notebook
## Description

Hi, I've experienced some erratic jumps when doing some actions on a notebook (eg. re-running cells). Here's a GIF - 

scroll_jump

FTR, the windowing mode is set to the default which is `defer`.

The jumps probably feel more erratic when the height of the cell output changes too drastically between re-runs or when there are too many cells.

## Reproduce

The snippet in the gif is - 

```
import matplotlib.pyplot as plt
from IPython.display import display

for _ in range(4):
    axes = plt.axes()
    axes.plot([1, 2, 3], [4, 5, 6])
    plt.close()
    display(axes.get_figure())
```

## Context

- JupyterLab version: v4.1.0b2

krassowski:
## References

Fixes #15693, solving two regressions introduced in 4.1 cycle:
- fixes erratic scrolling after Shift+Enter in non-windowing modes (defer, none),
- fixes too much scrolling after Ctrl+Enter in both windowing and non-windowing modes (Chromium-based browsers, Firefox not affected).

## Code changes

- corrects the node used for scrolling and padding computation - fixing a regression due to change in DOM structure introduced by https://github.com/jupyterlab/jupyterlab/pull/15533 (see the use of `_outerElement`)
- fixes behaviour of setting command mode which was causing a spurious scroll before running cells due to `focus` triggering a browser scroll - regression introduced by https://github.com/jupyterlab/jupyterlab/pull/14115 (see `preventScrollOnFocus`)
- the logic for scroll offset computation is now unified between windowing modes; when windowing is disabled the actual dimensions of the notebook and cell are used (see `precomputed` argument).
- repeats scrolling if notebook area resizes within 100ms of scroll request in non-windowing mode; this ensures that if a very brief change in dimensions due to output of the previously active cell hiding and re-appearing (in Shift + Enter scenario) shifts the newly active cell out of the viewport, that cell will be scrolled to again. This was previously used (with 50ms) in windowed mode which is why Shift + Enter worked there.

<details>

The last point might be better solved by changing when scroll gets called when a cell is run. Currently the sequence is:
- schedule cell execution (clears output)
- switch to next cell
- scroll if needed
- return a promise which resolves when cell(s) executed

This does not work great when executing a cell which has long output in the following scenario:
- schedule cell execution
- switch to next cell
- scroll if needed: **decides no scrolling is needed**
- execution completes, output is populated
- the newly added output shifts the next cell out of view (**bad**)

Therefore, it might be tempting to change this sequence to:
- schedule cell execution
- switch to next cell
- await execution to complete (and output to be populated)
- scroll to the output

But this would lead to surprising results if execution takes longer than ~100ms because a user could manually scroll away and be surprised that once the cell(s) completed execution, they are snapped back to the cell after the last executed cell.

Thus we would also need to cancel the pending scroll to next cell if user has already scrolled the view.

</details>

## User-facing changes

**First**, spurious scrolling is eliminated if the following cell (for shift + enter) or current cell (for ctrl + enter) is significantly visible (see examples below)

### Shift + Enter: short cell → long cell

| 4.1.0b2 (`defer` mode) | This PR (`defer` mode) | 3.6.x (reference) |
|--|--|--|
shift-enter-short-long-defer-mode-4 1 | shift-enter-short-long-defer-mode-PR | shift-enter-short-long-3 6 |

Note: this already worked in `full` mode since 4.0.9+ thanks to https://github.com/jupyterlab/jupyterlab/pull/15288 (see details).

<details>

| 4.1.0b2 (`full` mode) | This PR (`full` mode) | 4.0.10 (`full` mode) |
|--|--|--|
| shift-enter-short-long-full-mode-4 1 | shift-enter-short-long-full-mode-PR | shift-enter-short-long-full-mode-4 0 |

</details>

### Ctrl + Enter: long cell (partially out of view), Chrome 

| 4.1.0b2 (`defer` mode) | This PR (`defer` mode) | 3.6.x (reference) |
|--|--|--|
| ctrl-enter-long-defer-mode-4 1-chrome | ctrl-enter-long-defer-mode-PR-chrome | ctrl-enter-long-3 6-chrome |

Note: this was also broken in `full`, and this PR also fixes it in the full mode (see details).

<details>

| 4.1.0b2 (`full` mode) | This PR (`full` mode) |
|--|--|
| ctrl-enter-long-full-mode-4 1-chrome |  ctrl-enter-long-full-mode-PR-chrome |

</details>

**Second**, when execution with Shift + Enter invokes scrolling to next cell which was out of viewport, that cell will be positioned so that its top is in the middle of the viewport.

None


packages/notebook/src/actions.tsx
```diff
@@ -652,7 +659,10 @@ export namespace NotebookActions {
       notebook.activeCellIndex++;
     }
 
-    void Private.handleRunState(notebook, state, true);
+    // If a cell is outside of viewport and scrolling is needed,
+    // prefer to align the cell to the top viewport edge,
+    // rather than to the bottom (so that editor is visible)
+    void Private.handleRunState(notebook, state, true, 'start');
```
krassowski, line 665: This is leading to the `should scroll as little as possible on markdown rendering` test failing, which highlights trade-offs of different approaches, compare:

image

with alternative:

image

and with:

image

krassowski: It seems to me that we would want to restore the JupyterLab 3 behaviour here, maybe a variation of it (say "top edge in the 1/4 of the viewport from the top").

krassowski: So the problem with the earlier implementation (before https://github.com/jupyterlab/jupyterlab/pull/15288) was not that `auto` alignment differed from Lab 3.6, just that "center" was implemented as "align the middles of the cell and the viewport". This obviously led to very annoying behaviour for cells longer than viewport as seemingly random part of the output was shown:

image

I will push a commit to fix that behaviour, which should simplify the code too.

krassowski:
I now see that the fault of cell scrolling on switching to command mode affects many more actions than just running cells; for example pressing <kbd>Esc</kbd> in edit mode for a cell partially out of view would scroll, opening context menu over such a cell would scroll, etc. I will try to switch scroll on focus behaviour off by default.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1125,25 +1182,33 @@
...
-      );
-    }
-    for (const widget of this.layout.widgets) {
-      this._resizeObserver.observe(widget.node);
-      widget.disposed.connect(
-        () => this._resizeObserver?.unobserve(widget.node)
+    if (this.viewModel.windowingActive) {
+      if (!this._itemsResizeObserver) {
+        this._itemsResizeObserver = new ResizeObserver(
+          this._onItemResize.bind(this)
```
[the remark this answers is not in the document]
krassowski: False positive from testing code, resolved in 0870ffee94d957a129e65d0e6a8137fcd700e460

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1125,25 +1182,33 @@
...
+      for (const widget of this.layout.widgets) {
+        this._itemsResizeObserver.observe(widget.node);
+        widget.disposed.connect(
+          () => this._itemsResizeObserver?.unobserve(widget.node)
+        );
+      }
+      this._outerElement.addEventListener('scroll', this, passiveIfSupported);
+
+      this._scrollbarResizeObserver = new ResizeObserver(
+        this._adjustDimensionsForScrollbar.bind(this)
```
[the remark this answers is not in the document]
krassowski: False positive from testing code, resolved in 0870ffee94d957a129e65d0e6a8137fcd700e460

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1125,25 +1182,33 @@
...
       );
+      this._scrollbarResizeObserver.observe(this._outerElement);
+      this._scrollbarResizeObserver.observe(this._scrollbarElement);
+    } else {
+      if (!this._areaResizeObserver) {
+        this._areaResizeObserver = new ResizeObserver(
+          this._onAreaResize.bind(this)
```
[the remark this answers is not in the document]
krassowski: False positive from testing code, resolved in 0870ffee94d957a129e65d0e6a8137fcd700e460

krassowski:
Note for reviewer: failures of `Visual Regression Tests` are unrelated:
- `Notebook Run › Run Notebook and capture cell outputs` is flaky (toolbar blinking)
- `Notebook Toolbar › Paste cell` and `Table of Contents scrolling to heading › Notebook scrolls to heading` will be fixed by #15703 (also needs review)

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -292,13 +293,20 @@ export abstract class WindowedListModel implements WindowedList.IModel {
...
-   *   end - Align the item to the end of the list
-   *   start - Align the item to the beginning of the list
+   *   auto - Automatically align with the top or bottom minimising the amount scrolled,
+   *          If `alignPreference` is given, follow such preferred alignment.
+   *          If item is smaller than the viewport and fully visible, do not scroll at all.
+   *   smart - If the item is significantly visible, don't scroll at all (regardless of whether it fits in the viewport).
+   *           If the item is less than one viewport away, scroll so that it becomes fully visible (following the `auto` heuristics).
+   *           If the item is more than one viewport away, scroll so that it is centered within the viewport (`center` if smaller than viewport, `top-center` otherwise).
+   *   center - Align the middle of the item with the middle of the viewport (it only works well for items smaller than the viewport).
+   *   top-center - Align the top of the item with the middle of the viewport (works well for items larger than the viewport).
```
gabalafou, line 307: Nice

packages/notebook/src/actions.tsx
```diff
@@ -549,7 +549,10 @@ export namespace NotebookActions {
       translator
     );
 
-    void Private.handleRunState(notebook, state, false);
+    // While we do not want to scroll if cell is visible,
+    // we still need to invoke the scrolling logic
+    // in case if cell is outside of viewport.
```
gabalafou, line 554: How does one execute the run command on a cell outside of the viewport?

packages/notebook/src/actions.tsx
```diff
@@ -652,7 +659,10 @@ export namespace NotebookActions {
       notebook.activeCellIndex++;
     }
 
-    void Private.handleRunState(notebook, state, true);
+    // If a cell is outside of viewport and scrolling is needed,
+    // prefer to align the cell to the top viewport edge,
+    // rather than to the bottom (so that editor is visible)
+    void Private.handleRunState(notebook, state, true, 'center');
```
gabalafou, line 665: Why 'center' and not 'top-center'? I thought the idea was to keep the cell that we just left in view, which we cannot guarantee if we scroll the center of the next cell into the center of the viewport.

gabalafou (commented):
I'm not done reviewing, but I thought I would go ahead and leave the review comments I have so far. 

I'm not sure I feel confident about my ability to review the changes to the windowed list classes.

packages/notebook/src/actions.tsx
krassowski: Many possibilities, one is the ToC:

image

galata/playwright.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
     },
     {
       name: 'jupyterlab',
-      testMatch: 'test/jupyterlab/**',
+      testMatch: 'test/jupyterlab/**/*.test.ts',
```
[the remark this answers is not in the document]
krassowski: Yes, this is the same regex that we use for documentation project which also has a `utils.ts`, see lines 10-14 above.

packages/notebook/src/actions.tsx
krassowski: Great question! This is because it only applies to `auto` strategy and `handleRunState` uses the `smart` strategy, which will:
- (a) not scroll if significantly visible, otherwise
- (b) scroll using `auto` heuristics if less than one viewport away
        this point is important because if an item is less than one viewport and not significantly visible it implies that it fits in the viewport hence we can apply `center` rather than `top-center`.
- (c) choose between `top-center` and `center` depending on item size.

So in (b) we can use `center` over `top-center`; but should we? For small cells it is nicer, and in a rare case of item size being equal to viewport we loose past context (future context would not be visible either way), so maybe we could choose between `center` and `top-center` depending on item size, but I did not want to make the logic any more complicated; in fact I was considering removing this fallback to `auto` heuristic here but was also cautious not to modify the logic too much in case if downstream already depends on `smart` falling back to `auto` in this circumstance. In either case, yes we _could_ also use `top-center` here. What do you think?

This also highlights that the comment above the line you commented on is outdated because I had not updated it since the previous solution was in place. I will update it now.

krassowski: The comment was updated in 5b02962df509818e7cc5ca0b8c6515a4c74f03fa

gabalafou: I can't help but wondering if the added code complexity outweighs the UX benefit. Seems to me that if we don't introduce the `alignPreference` parameter then the consequence is that in the scenario labeled (b), in which the cell fits within the viewport and is less than one viewport away, then the cell will be scrolled to in such a way that either its top edge will align with the top edge of the scrollable container or its bottom edge with the bottom edge. I'm probably missing something but I don't see how that's significantly worse than scrolling the cell's top edge (or center) to the center of the scroll-container.

gabalafou (changes requested):
I can see that a lot of thought, care, and hard work has gone into this PR, so I hesitate to ask this question, but I was wondering if it would make sense to split up this PR. I may be misunderstanding the code, but it seems that only some of the code changes in this PR strictly address fixing the regressions that caused the bug reported in issue #15693. The other code changes seem like enhancements. Would it make sense to separate the two?

Marking request changes for my question about the test code.

galata/test/jupyterlab/notebook-scroll.test.ts
```diff
@@ -13,12 +14,12 @@ test.use({
     ...galata.DEFAULT_SETTINGS,
     '@jupyterlab/notebook-extension:tracker': {
       ...galata.DEFAULT_SETTINGS['@jupyterlab/notebook-extension:tracker'],
-      windowingMode: 'full'
+      windowingMode: 'none'
```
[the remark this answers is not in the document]
krassowski: Yup, that's a mistake when copy-pasting when updating this file to maintain the files identical beyond mode change. Thanks!

> Is there a better way to parameterize this set of tests or does the mocking require two nearly copy-pasted files?

There are two options:
- modify settings during on test runtime  - which will extend the runtime for tests and require adding a lot of code which would depend on settings editor layout - so not good
- modify projects with custom settings (preferred) - but this one is a larger modification to our CI because we split jobs by existing projects (jupyterlab/galata/documentation) - so out of scope for this PR

galata/test/jupyterlab/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  return intersection / n.height;
+}
+/**
+ * Scroll so that the given cell is below the viewport, revealing only a given `ratio` of the cell.
+ */
+export async function scrollToAboveCell(
+  page: IJupyterLabPageFixture,
+  notebook: ElementHandle,
+  cell: ElementHandle,
+  ratio: number
```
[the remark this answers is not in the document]
krassowski: Yeah, naming here is not perfect, but `scrollToCell` is too generic. The intent here was to indicate that this follows the "bottom" alignment strategy; I will update it.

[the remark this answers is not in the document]
krassowski: Yes, naming here is not perfect, but `scrollToCell` is too generic and wrong (because this helper will for example not align to top or middle). In fact it does not scroll cell into view when ratio is zero - it scrolls it **below the viewport**. I will still update the name.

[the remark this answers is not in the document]
krassowski: Updated to `positionCellPartiallyBelowViewport`

galata/test/jupyterlab/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+): Promise<void> {
+  const scroller = (await notebook.$(
+    OUTER_SELECTOR
+  )) as ElementHandle<HTMLElement>;
+  const notebookBbox = await scroller.boundingBox();
+  const cellBBox = await cell.boundingBox();
+  await page.mouse.move(notebookBbox.x, notebookBbox.y);
+  const scrollOffset =
+    cellBBox.y - notebookBbox.y - notebookBbox.height + cellBBox.height * ratio;
+  await page.mouse.wheel(0, scrollOffset);
```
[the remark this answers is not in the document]
krassowski: Thanks for suggestion. I would prefer to keeping it as-is for consistency with offset naming in the implementation file.

galata/test/jupyterlab/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  page: IJupyterLabPageFixture,
+  notebook: ElementHandle,
+  cell: ElementHandle,
+  ratio: number
+): Promise<void> {
+  const scroller = (await notebook.$(
+    OUTER_SELECTOR
+  )) as ElementHandle<HTMLElement>;
+  const notebookBbox = await scroller.boundingBox();
+  const cellBBox = await cell.boundingBox();
```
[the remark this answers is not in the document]
krassowski: 39ce21d1927e71b37dad7be1891b2fbdb103634f

packages/notebook/src/actions.tsx
krassowski: I think that we should also have a function that does not do mode changing nor scrolling (and some extension developers have specifically asked about it), but that might not even use `handleRunState` at all and is out of scope for this PR. I removed the `scroll` argument in 39ce21d1927e71b37dad7be1891b2fbdb103634f.

galata/test/jupyterlab/notebook-scroll.test.ts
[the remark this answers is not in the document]
gabalafou: There's one other possibly non-trivial difference between these two files when I diff'd them, are you aware of it?

galata/test/jupyterlab/notebook-scroll.test.ts
```diff
@@ -131,15 +124,53 @@ test.describe('Notebook scroll on execution', () => {
     // The third cell should be positioned at the bottom, revealing between 10 to 20% of its content.
     await expect(thirdCellLocator).toBeInViewport({ ratio: 0.1 });
     await expect(thirdCellLocator).not.toBeInViewport({ ratio: 0.2 });
+    await expect(thirdCellLocator).not.toBeInViewport({ ratio: 0.2 });
```
gabalafou, line 127: This line is duplicated

gabalafou (changes requested):
Okay thanks for the clarification. 

Apart from some reservations I have about increasing the complexity of the code with the `alignPreference` parameter, this code looks good to me!

However, I still think there's something amiss with those test files. Could you diff them to make sure?

galata/test/jupyterlab/notebook-scroll.test.ts
krassowski: Great catch, thanks! It now looks good for me:

```diff
20c20
<       windowingMode: 'full'
---
>       windowingMode: 'none'
25c25
< test.describe('Notebook scroll on navigation (with windowing)', () => {
---
> test.describe('Notebook scroll on navigation (no windowing)', () => {
65c65
< test.describe('Notebook scroll on execution (with windowing)', () => {
---
> test.describe('Notebook scroll on execution (no windowing)', () => {
172c172
< test.describe('Notebook scroll over long outputs (with windowing)', () => {
---
> test.describe('Notebook scroll over long outputs (no windowing)', () => {
```

krassowski:
@gabalafou does it look good to you now?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1283,17 +1367,30 @@ export class WindowedList<
...
   }
 
+  /**
+   * Scroll to the item which was most recently requested.
+   *
+   * This is meant to be used when immediately after scrolling to the item
+   * a resize happened (e.g. output of a previous cell got populated after
+   * Shift+Enter). This is ensured by scrolling to `this._scrollToItem`
+   * which is cleared after a short timeout once the scrolling settles
+   * (see `this._resetScrollToItem()`).
```
gabalafou, line 1383: This comment would be easier for me to follow if it were presented as a list of sequential events.

```suggestion
   * This method ensures that the app scrolls to the item even if a resize event
   * occurs shortly after the scroll. Consider the following sequence of events:
   *
   * 1. User is at the nth cell, presses Shift+Enter (run current cell and
   *    advance to next)
   * 2. App scrolls to the next (n+1) cell
   * 3. The nth cell finishes running and renders the output, pushing the
   *    (n+1) cell down out of view
   * 4. This triggers the resize observer, which calls this method and scrolls
   *    the (n+1) cell back into view 
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1256,9 +1333,16 @@ export class WindowedList<
   }
 
   /**
-   * Handle viewport content (e.g. widgets) resize.
+   * Handle viewport area resize.
+   */
+  private _onAreaResize(_entries: ResizeObserverEntry[]): void {
+    this._scrollBackToItemOnResize();
```
gabalafou, line 1339: Just checking my own understanding here. 

Is this line more for code symmetry and to cover edge cases than to cover a commonly occurring scenario? I mean, it's generally unlikely that an area resize will happen within 100 ms of the app scrolling to some item, right?

packages/notebook/src/actions.tsx
gabalafou, line 554: I generally lean toward more comments than fewer, but in this case I think it would better to remove the comment or move it to the `handleRunState` function definition.

```suggestion
```

packages/notebook/src/actions.tsx
```diff
@@ -587,7 +590,11 @@ export namespace NotebookActions {
       sessionDialogs,
       translator
     );
-    void Private.handleRunState(notebook, state, false);
+
+    // While we do not want to scroll if cell is visible,
+    // we still need to invoke the scrolling logic
+    // in case if cell is outside of viewport.
```
gabalafou, line 596:
```suggestion
```

packages/notebook/src/toc.ts
```diff
@@ -610,7 +610,8 @@ export class NotebookToCFactory extends TableOfContentsFactory<NotebookPanel> {
             console.debug('scrolling to heading: using fallback strategy');
             await widget.content.scrollToItem(
               widget.content.activeCellIndex,
-              this.scrollToTop ? 'start' : undefined
+              this.scrollToTop ? 'start' : undefined,
+              0
```
gabalafou, line 614: Hmm, I never asked about this change (margin=0) and I'm not sure I understand why it's here

packages/testing/src/jest-shim.ts
```diff
@@ -57,6 +57,34 @@ window.focus = () => {
...
+  }
+  unobserve(_target: any) {
+    // no-op
+  }
+  disconnect() {
+    // no-op
+  }
+}
+
+window.ResizeObserver = ResizeObserverMock;
```
gabalafou, line 86: Also checking my understanding here. If the ResizeObserver is basically just a no-op mock, that means that none of the tests are hitting code paths that would lead to _scrollBackToItemOnResize() being called, correct?

gabalafou (approved):
Okay, I'm done reviewing and I approve with one reservation, which is that I'm still not sure whether the UX benefit of [`alignPreference` is worth the added complexity](https://github.com/jupyterlab/jupyterlab/pull/15702#discussion_r1472282664) (not sure if you saw my comment asking about that). But this is not blocking for me. I also left some comments in this last round of review, but these are also not blocking.

Great job unifying the windowing and no-windowing UX!

packages/notebook/src/actions.tsx
gabalafou: By "function that does not do mode changing nor scrolling", do you mean, for example, being able to run a cell but stay in edit mode and do not scroll?

packages/notebook/src/toc.ts
krassowski: Good question. First, this is the fallback strategy which is unlikely to be used. Before this PR it was broken in defer mode because of wrong node being used for scrolling further down the line. Currently both defer and full mode will work, invoking a logic which considers the margin. Now, the default visibility margin for scrolling on action is 25%. But when user uses ToC they often want to find the heading and if it was partially revealed and the clicked either way it means that they expect actions anyways.

packages/ui-components/src/components/windowedlist.ts
krassowski: No, it is actually quite common when user executes a cell which produces output and advances to the next cell. See:

image

packages/testing/src/jest-shim.ts
krassowski: > that means that none of the tests are hitting code paths

More precisely none of the _unit_ tests get to call `_scrollBackToItemOnResize()`, but the integration tests do. In fact if one was to remove it the visual tests would fail.

packages/ui-components/src/components/windowedlist.ts
krassowski: Thanks! I incorporated your suggestion but also kept the implementation explanation.

krassowski:
Thank you for the review @gabalafou!

> with one reservation, which is that I'm still not sure whether the UX benefit of `alignPreference` is worth the added complexity (not sure if you saw my comment asking about that).

I just want to acknowledge that I saw your comment and take responsibility of dealing with the added complexity arising from having this parameter if it ever becomes a problem (and I believe this is worth the - small - UX benefit).

packages/ui-components/src/components/windowedlist.ts
gabalafou: Ah, for some reason I thought the area resize observer was observing the outer (scrollable) element (for e.g. window resize). Thanks for the clarification.

PR #15716, opened 2024-01-30
Fix `overrides.json` not working for shortcuts

krassowski:
## References

Fixes #15713

## Code changes

- [x] add test ensuring that the default settings (such as from `overrides.json`) are respected - it should fail in the <s>first</s> second commit (it seems this test group was not run on CI, second commit fixes it)
- [x] fix the overrides getting lost

## User-facing changes

Shortcuts modifications defined in `overrides.json` now work.

## Backwards-incompatible changes

None

PR #15717, opened 2024-01-30
Run Mermaid tests on CI

krassowski:
## References

#15716 highlighted that not all packages are tested on CI; while it handles the missing `shortcuts-extension` group (which will need to be backported) I also found that `mermaid` was missing (by using [this search](https://github.com/search?q=repo%3Ajupyterlab%2Fjupyterlab+%22%5C%22test%5C%22%3A+%5C%22jest%5C%22%22&type=code)) which is what this PR is fixing. Mermaid was only added in 4.1 so no need for backporting.

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -276,6 +276,14 @@ pkg:mathjax:
     - packages/mathjax-extension/**/*
     - packages/mathjax-extension/*
 
+pkg:mermaid:
+- changed-files:
+  - any-glob-to-any-file:
+    - packages/mermaid/**/*
+    - packages/mermaid/*
+    - packages/mermaid-extension/**/*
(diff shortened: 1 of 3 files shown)


jtpio (approved):
Thanks for catching this!

jtpio:
Maybe we should we also add the `pkg:mermaid` GitHub label?

krassowski:
Thanks! I added the mermaid label to config in [848ada1](https://github.com/jupyterlab/jupyterlab/pull/15717/commits/848ada1b909229f9eb5a06b8a8cb05b23a45b507), we should see it applied once this one is merged.

jtpio:
Thanks!

Issue #15705, opened 2024-01-28
Running Terminals and Kernels sidepanel scrollbar issue


jans-code:
## Description

In the 'Running Terminals and Kernels' sidepanel is some issue with an always visible scrollbar.
It seems the entries in the list always get a scrollbar because of `.jp-RunningSessions-sectionContainer` has set
`overflow: auto` and for me it always shows a scrollbar next to the entries. This probably originates from the Shutdown-Button margin being too big.

![image](https://github.com/jupyterlab/jupyterlab/assets/112409139/9003ca54-de3f-489e-971f-29dff677ed03)

This leads to having two scrollbars when there is actually scrolling needed because `.jp-RunningSessions-section` overflow additionally kicks in.

![image](https://github.com/jupyterlab/jupyterlab/assets/112409139/7fc900b5-ff23-4053-aca0-8cb20892ac14)

## Reproduce

Just opening some Launchers makes problem visible for me.

## Expected behavior

Just one scrollbar when needed.

## Context

- Operating System and version: Archlinux
- Browser and version: Chromium
- JupyterLab version: 4.1.0rc0

krassowski:
This is a regression in 4.1.0 (thank you for testing!) due to the toolbar button sizing changing - the buttons are still 24px high, but now also include 4px margin; this in turn makes the (incorrect) double overflow kick in.

PR #15718, opened 2024-01-31
Fix kernels sidepanel scrollbar

FoSuCloud:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15705

## Code changes

cancel `.jp-RunningSessions-sectionContainer` css `overflow:auto`
Because `.jp-RunningSessions-section` also has `overflow:auto`
截屏2024-01-31 09 12 44

## User-facing changes

Before the change, there would be two scroll bars when scrolling, and it was difficult to scroll to the end.
截屏2024-01-31 09 24 50

After the change, only one scrollbar exists
截屏2024-01-31 09 20 08

nothing


FoSuCloud:
@meeseeksdev tag bug

krassowski (approved):
Thank you @FoSuCloud! I can confirm that this fixes the issue.

krassowski:
Merging - the failing `Paste Cell` test is unrelated (tracked in https://github.com/jupyterlab/jupyterlab/issues/15719).

Issue #15698, opened 2024-01-25
Visual Regression Documentation tests action uses Node.js 16, which is deprecated


JasonWeill:
## Description

In the run details for visual regression tests, I see an error:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/setup-python@v4, actions/cache@v3, actions/setup-node@v3, ifaxity/wait-on-action@v1, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

## Expected behavior

We need to upgrade these actions to Node 20, which is still supported until late 2024, with maintenance continuing until Q2 2026. NodeJS [ended support on September 11, 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol/).

krassowski:
I the above is referring to `4.0.x` branch; on `main` there is only one of these (`ifaxity/wait-on-action@v1`):

> The following actions uses node12 which is deprecated and will be forced to run on node16: ifaxity/wait-on-action@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Of note there might be even older versions on 3.6.x branch.

krassowski:
Depends on https://github.com/iFaxity/wait-on-action/issues/12

krassowski:
A few remain:

> actions/setup-python@v4, actions/cache@v3, actions/setup-node@v3

but these are not defined in https://github.com/jupyterlab/jupyterlab/blob/main/.github/workflows/galata.yml but from dependencies too.

krassowski:
Will be fixed by https://github.com/jupyterlab/maintainer-tools/pull/235

Issue #13650, opened 2022-12-22
Galata tests: use `webServer` in the Playwright config to start JupyterLab


jtpio:
### Problem

Currently we need to start JupyterLab separately to run the Galata tests:

https://github.com/jupyterlab/jupyterlab/blob/10c16e8bf8f3451d5a16ef8747c7d803fccb00aa/.github/workflows/galata.yml#L30-L33

And wait for it to be ready:

https://github.com/jupyterlab/jupyterlab/blob/10c16e8bf8f3451d5a16ef8747c7d803fccb00aa/.github/workflows/galata.yml#L42-L46

### Proposed Solution

Playwright added support for starting web servers via the `playwright.config.js` in 1.24:

https://github.com/microsoft/playwright/releases/tag/v1.24.0

```ts
// playwright.config.ts
import type { PlaywrightTestConfig } from '@playwright/test';
const config: PlaywrightTestConfig = {
  webServer: [
    {
      command: 'npm run start',
      port: 3000,
      timeout: 120 * 1000,
      reuseExistingServer: !process.env.CI,
    },
    {
      command: 'npm run backend',
      port: 3333,
      timeout: 120 * 1000,
      reuseExistingServer: !process.env.CI,
    }
  ],
  use: {
    baseURL: 'http://localhost:3000/',
  },
};
export default config;
```

We could leverage this to streamline how the JupyterLab application is started for the UI tests. Playwright handles waiting for the application to be ready (`port` and `url` are configurable).

### Additional context

Docs: https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

PR #15721, opened 2024-01-31
Replace `ifaxity/wait-on-action` wrapper with direct `wait-on` use

krassowski:
## References

Partial fix for #15698

## Code changes

Removes CI dependency on third-party action which is not keeping up with the deprecations

## User-facing changes

None

None


jtpio (approved):
Thanks!

jtpio:
Probably we won't need this action anymore after switching to `webServer`: https://github.com/jupyterlab/jupyterlab/issues/13650 (in case someone would like to take a look!)

jtpio:

krassowski:
Thanks! Good idea on using `webServer`, though I guess we will still need this for docker test.

PR #15724, opened 2024-01-31
Update maintainers affiliation

fcollonval:
## References

Update some affiliation on the README

@marthacryan I updated yours - is it ok?
@krassowski Does your affiliation need to be changed?

## Code changes

None

## User-facing changes

none

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ JupyterLab's current maintainers are listed in alphabetical order, with affiliat
 - Brian Granger, AWS (co-creator, strategy, vision, management, UI/UX design,
   architecture).
 - Jason Grout, Databricks (co-creator, vision, general development).
-- Michał Krassowski, University of Oxford (general development, extensions).
+- Michał Krassowski, Quansight (general development, extensions).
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @fcollonval!

Curious if we should continue maintaining this list in the JupyterLab repo? Or just link to https://jupyterlab-team-compass.readthedocs.io/en/latest/team.html, which already contains the current list of council members?

fcollonval:
> Curious if we should continue maintaining this list in the JupyterLab repo? Or just link to https://jupyterlab-team-compass.readthedocs.io/en/latest/team.html, which already contains the current list of council members?

Less maintenance is always good. It raises three questions:
- is the council a superset of the maintainers?
- is it ok to link the coming new frontend council?
- what about the "emeritus"? - I guess that list could be published on the team-compass too.

fcollonval:
I opened an issue on the team-compass. So we can move forward with this while pushing forward the discussion.

jtpio (approved):
Thanks!

Maybe can wait a bit for @marthacryan to check if it looks good too?

krassowski:
Just checking back @marthacryan?

Issue #15671, opened 2024-01-21
Cell toolbar gets stuck when using collapse cell feature


krassowski:
## Description

Cell toolbar does not hide when switching away after using collapse cell feature

## Reproduce

1. Click on cell collapser
2. Switch to another cell
3. Uncollapse the collapsed cell
4. Switch to another cell
5. See the cell toolbar remains stuck

(Ignore notebook name on the gif, happens with both modes)

![toolbar-gets-stuck](https://github.com/jupyterlab/jupyterlab/assets/5832902/895b4b03-8ff5-430b-b45d-9f0565c483b8)

## Expected behavior

Cell toolbar gets hidden

## Context

- Browser and version: tested on Chromium but probably all
- JupyterLab version: 4.0.9, 4.1.0b2 and all in between; works well in 4.0.8 and 3.6.x

Likely introduced by https://github.com/jupyterlab/jupyterlab/pull/15314 ?

No errors in the console.

FoSuCloud:
I want to try to solve this problem

PR #15732, opened 2024-02-02
Fix cell toolbar getting stuck when using collapse cell

FoSuCloud:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15671

## Code changes

The reason for the bug is that after the folded cell is clicked, the `_onActiveCellChanged` function is triggered, but because the`inputHidden` attribute of the folded cell activeCell is true, the `_previousActiveCell` variable is not updated.
After that, clicking other cells cannot destroy the previous active cell toolbar

截屏2024-02-02 08 25 32

## User-facing changes

Before the change

https://github.com/jupyterlab/jupyterlab/assets/49218295/f34f61e8-0049-4ef4-9d97-bcec72023e18

After the change

https://github.com/jupyterlab/jupyterlab/assets/49218295/5da68ee0-2d7b-4cff-bd98-4237cb0f2b07

nothing

### DIFF
--- a/packages/cell-toolbar/src/celltoolbartracker.ts
+++ b/packages/cell-toolbar/src/celltoolbartracker.ts
@@ -137,7 +137,7 @@ export class CellToolbarTracker implements IDisposable {
     }
 
     const activeCell = notebook.activeCell;
-    // change previous active cell
+    // Change previously active cell.
     this._previousActiveCell = activeCell;
     if (activeCell === null || activeCell.inputHidden) {
       return;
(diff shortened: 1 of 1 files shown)


FoSuCloud:
@meeseeksdev tag bug

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -137,14 +137,15 @@ export class CellToolbarTracker implements IDisposable {
     }
 
     const activeCell = notebook.activeCell;
+    // change previous active cell
```
krassowski, line 140:
```suggestion
    // Change previously active cell.
```

krassowski (approved):
Thank you @FoSuCloud!

Minor suggestion on the style of the comment.

krassowski:

PR #15733, opened 2024-02-02
Upgrade to `mermaid` 11.x, `marked` 15.x

bollwyvl:
## References

- continues #15661

## Code changes

- [x] `*/package.json`
  - [x] `dependencies`
    - [x] update to [`mermaid ^11.4.1`](https://github.com/mermaid-js/mermaid/releases/tag/v11.4.1)
    - [x] update to [`marked ^15.0.3`](https://github.com/markedjs/marked/releases/tag/v15.0.3)
  - [x] `resolutions`
    - [x] `marked` (as above)
    - [x] `typefest ^4.3.0`
- [x] `tsconfigbase.json`
  - [x]  `lib` 
    - [x] add the `ES2020.BigInt` 
      - used by the `type-fest` depended on by... something this upgrades. seems better than wholesale `ES2020` or `skipLibCheck` 
  - [x] `ES2020.String` for `String.matchAll`
    - used to detect renderers (currently only `elk`)
- [x] `galata`
  - [x] add snapshot test for _block_ diagram
  - [x] add snapshot test for _kanban_ diagram
  - [x] add snapshot test for _flowchart-elk_ diagram
- [x] `packages/mermaid`
  - [x] update for new `@mermaid-js/elk-layout` package
  - [x] sniff for several flavors of `defaultRenderer: elk`, load lazily

## User-facing changes

- users can now 
  - create _block_ and _kanban_ diagrams
    - more are enabled, but have the `-beta` name, so not testing
  - use some new symbols in _sequence_ diagrams
  - use additional characters in _gantt_ diagrams
  - see better memory/performance due to better use of  `dompurify` after heavy diagram editing over a long period of time
  - get denser _git_ diagrams with parallel commits in unrelated commit streams 
  - create many more custom shapes in flowcharts

- at least 
  - `\n` newlines are broken in notes on diagrams
  - some new text wrap issues on long lines of requirements diagrams (wasn't under test, as they weren't great to begin with)

## Future Work (not on this PR)
- now that `@mermaid-js/layout-elk` is a wholly separate feature and package we could also separate it out
  - add a `IMermaidManager` shim around `Mermaid.registerLayoutLoaders`
  - move the `@mermaid-js/layout-elk` dependency to a new package, e.g. `@jupyterlab/mermaid-elk`
    - use the new method to register the renderer


bollwyvl:
Huzzah: no fails against screenshots (this is good, as upstream _also_ does screenshot-based testing).

Boo: the new _block_ diagram generates custom entities e.g. `&nbsp;` that aren't implicitly well-defined in XHTML. 

We may need to revisit running through the full XML parser/serializer, as hinted at in most recently-added test for void entities.

Bottom line: I don't think we would need to get this out in 4.1.0, and it will be possible to update in the future now that we have a baseline of what we _know_ we support.

bollwyvl:
Also, of note, I do see a lot of this when working on diagrams interactively:

```

Private.exceptionHandler @ index.es6.js:308
invokeSlot @ index.es6.js:558
emit @ index.es6.js:513
emit @ index.es6.js:112
reset @ inline.ts:673
handleSelectionChange @ inline.ts:741
onSelectionsChanged @ handler.ts:257
invokeSlot @ index.es6.js:555
emit @ index.es6.js:513
emit @ index.es6.js:112
set @ observablemap.ts:211
_onCursorActivity @ editor.ts:640
_evtFocus @ editor.ts:719
handleEvent @ editor.ts:700
(anonymous) @ index.js:2745
ignore @ index.js:6198
updateSelection @ index.js:2723
update @ index.js:6630
EditorView._dispatch @ index.js:6481
dispatch @ index.js:6547
setCursorPosition @ editor.ts:406
showEditor @ widget.ts:2257
_handleRendered @ widget.ts:2298
set rendered @ widget.ts:2076
setMode @ widget.ts:1448
set mode @ widget.ts:1411
execute @ index.ts:3080
execute @ index.es6.js:364
_executeKeyBinding @ index.es6.js:526
processKeydownEvent @ index.es6.js:462
evtKeydown @ index.es6.js:2111
handleEvent @ index.es6.js:2067

```

bollwyvl:
Boy howdy, not proud of this solution. We'll see what the tests think.

bollwyvl:
@meeseeksdev tag maintenance

krassowski:
@bollwyvl is this something you would like to include in 4.2.0? It looks like it picked up a small merge conflict.

krassowski:
I tried to resolve the merge conflict, I hope it's ok :)

bollwyvl:
Yeah, and looks like we're a couple behind now: https://github.com/mermaid-js/mermaid/releases

Of note, 10.9.0 added katex (buh)... math in diagrams is good, but while katex is not _as_ big as mathjax, it's still not nothing... much like their markdown rendering, really wish more of these things were pluggable.

I had kind been holding out for some of the Big Refactor stuff coming down the pipe, as it will normalize a lot of the internals so there are fewer inconsistencies between different diagrams, and make a LS more feasible, but keeping up with the upstream is probably a good play, anyway.

bollwyvl:
Re: above: looks like `marked` will be in the next release: https://github.com/mermaid-js/mermaid/pull/5636 . Small wins. :sweat_smile:

krassowski:
@bollwyvl is this ready for review now? GitHub UI still shows this as a draft.

krassowski:
@bollwyvl just checking back on this one

bollwyvl:
This happened:

https://github.com/mermaid-js/mermaid/releases/tag/v11.0.0

Will start looking into the changes for the 4.4 timeline.

bollwyvl:
Quick update: it took a couple upstream PRs, but the next release of the `11.x` line should be ready for another look at integrating here. As a recap, this will:
- add a dependency on katex :crying_cat_face: 
- remove the dependency on _another_ markdown parser, as it uses `marked`

The `elkjs` renderer will also become optional, but it is my feeling we should keep shipping it as, even with some of its limitations and drawbacks (mostly size), it provides the best output for a lot of kinds of diagrams.

bollwyvl:
https://github.com/mermaid-js/mermaid/releases/tag/mermaid%4011.2.0

bollwyvl:
I've got some more things working for (actually-exists-on-`main`) docs issues, but it's getting unwieldy to keep working on this already-too-complex branch. Will take a break here (no real hurry) and try to get some build/docs stuff updated which would make this easier.

bollwyvl:
In digging through our screenshot diffs, did uncover a regression:

https://github.com/mermaid-js/mermaid/issues/6126

packages/mermaid/src/manager.ts
```diff
@@ -82,6 +84,11 @@ export class MermaidManager implements IMermaidManager {
   async renderSvg(text: string): Promise<IMermaidManager.IRenderInfo> {
     const _mermaid = await this.getMermaid();
 
+    // could trigger a false-positive for diagrams about real cervidae, but...
+    if (text.includes('elk')) {
```
krassowski, line 88: :rofl: also cities of [Ełk](https://en.wikipedia.org/wiki/E%C5%82k) and [Elk](https://en.wikipedia.org/wiki/Elk_City,_Oklahoma) if wrongly capitalised. And any word containing `elk`? 

- What is the cost of a false positive here?
- Is there a delimiter that could be included in the query?

tsconfigbase.json
```diff
@@ -7,7 +7,7 @@
     "esModuleInterop": true,
     "incremental": true,
     "jsx": "react",
-    "lib": ["DOM", "DOM.Iterable", "ES2018", "ES2020.Intl"],
+    "lib": ["DOM", "DOM.Iterable", "ES2020", "ES2020.Intl"],
```
krassowski, line 10: Probably worth surfacing to PR description. Or opening an issue about changing `target` to `ES2020`.

krassowski:
Also, it looks like lockfile picked up a conflict - probably from merge of https://github.com/jupyterlab/jupyterlab/pull/17070

tsconfigbase.json
bollwyvl: yeah, it's this business:

```
> @jupyterlab/metapackage@4.4.0-alpha.0 build
> tsc -b

../../node_modules/type-fest/source/typed-array.d.ts:16:4 - error TS2583: Cannot find name 'BigInt64Array'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2020' or later.

16  | BigInt64Array
      ~~~~~~~~~~~~~

../../node_modules/type-fest/source/typed-array.d.ts:17:4 - error TS2583: Cannot find name 'BigUint64Array'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2020' or later.

17  | BigUint64Array;
      ~~~~~~~~~~~~~~

```

bollwyvl: yerp. isolating the change to just `mermaid/{package,tsconfig}.json` doesn't work, either.

I don't think having access to _types_ from the modern specs is going to hurt the generated runtime code, as the `target` is still the same and will complain if there is something it can't downlevel.

packages/mermaid/src/manager.ts
bollwyvl: So, `IMermaidManager` is already a token: since this behavior is split up now, we could add a `registerLayoutLoader` which would isolate this further, and allow folks to disable it. Will do an API sketch _after_ it's tested.

The biggest win here would be if there existed an accessible renderer that generated useful HTML representations... any flowchart-kinda thing can be represented as two linked tables, a mindmap as a bulleted list, etc. Way out of scope for this, but might be germaine to your non-human interpretations as well.

As per usual, this would _not_ be portable to other platforms (e.g. Git{Hu,La}b), but they don't (and might never) support elk. Nothing we can do about that, but we _could_ ensure jupyter implementations worked consistently.

tsconfigbase.json
bollwyvl: Yep, this worked.

packages/mermaid/src/manager.ts
```diff
@@ -328,12 +333,47 @@
...
+    const tmpMermain = (_mermaid = (await import('mermaid')).default);
     initMermaid();
+    _mermaid = tmpMermain;
     _loading.resolve(_mermaid);
     return _mermaid;
   }
 
+  /** Detect and load a renderer configured via `%init` or YAML front matter. */
+  export async function ensureRenderer(text: string): Promise<void> {
+    const match = RE_DEFAULT_RENDERER.exec(text);
```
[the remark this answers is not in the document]
bollwyvl: bah, that's probably why i did the string include in the first place

packages/mermaid/src/manager.ts
bollwyvl: (also, error sniffing does not work, it doesn't complain at all and just falls back to dagre)

tsconfigbase.json
bollwyvl: added to the description

packages/mermaid/src/manager.ts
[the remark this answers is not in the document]
bollwyvl: Welp, have something local with more `\b`... perhaps we should consider adding some more best-practice tools so it's not whack-a-mole on CI, e.g. https://ota-meshi.github.io/eslint-plugin-regexp/

bollwyvl:
CodeQL ate my comment (or something) but wanted to re-note:

https://ota-meshi.github.io/eslint-plugin-regexp/

(I used this locally to iterate a bit faster after the second fail)

Unfortunately (as with the `tsdoc` stuff) it would take a while to adopt:

> ✖ 62 problems (27 errors, 35 warnings)

bollwyvl:
Ok, all back to :heavy_check_mark: .

I could _now_ more confidently do the refactor [mentioned above](https://github.com/jupyterlab/jupyterlab/pull/15733#discussion_r1882125483):
- add `IMermaidManager.addLayout`
- add `packages/mermaid-elk(-extension)` which used it, and was independently disable/removable

This could also be in a follow-on, if we're thinking this could go into an alpha sooner rather than later.

krassowski (approved):
Thank you @bollwyvl!

krassowski:
FYI adjusting PR title slightly just to get easier time when running spellcheck on changelog which gets autogenerated from PR titles

bollwyvl:
(and thank you @krassowski for all your time in reviewing this over the year :heart_eyes_cat: )

Issue #15704, opened 2024-01-27
Code Completion Partially Hidden by Scrollbars


phil-zxx:
## Description

When triggering code completion in JupyterLab (after hitting the `TAB` key), the output is sometimes partially covered by the scrollbars. See **left-hand-side** for an example of when this happens. For comparison, the **right-hand-side** shows the correct behaviour.
![image](https://github.com/jupyterlab/jupyterlab/assets/32247590/20e239ce-bbb1-43ce-9d4d-42d1b1cc703d)

This can happen at various browser zoom levels:
![image](https://github.com/jupyterlab/jupyterlab/assets/32247590/e1c80ed2-3c57-43a0-acf7-8f6b17b251d8)

This seems to be a recent issue (last ~6 months), as I have been using JupyterLab for years without ever encountering this issue. So I believe this is a **bug**.

## Reproduce

1. Open Jupyterlab
2. Create a new notebook
3. Type anything into the first cell (like `T` or `Ti`) and then hit `TAB` to trigger the code completion
4. Part of the code completion may be covered by scrollbars

## Expected behavior

Code completion should **_not_** be covered by scrollbars (unless you have more than 10 suggestions, of course).

## Context

- OS: Windows 10 & 11 (I tested this on both)
- Browser: Chrome 120.0
- JupyterLab version: 4.0.11
- Python version: 3.11.2
- Installed Packages: See `pip list` below
- Screens: 2 Screens side-by-side (4K, 3840×2160 each)

<details><summary>pip list</summary>
<pre>
>>pip list
Package                  Version
------------------------ ------------
aiofiles                 22.1.0
aiosqlite                0.18.0
anyio                    3.6.2
argon2-cffi              21.3.0
argon2-cffi-bindings     21.2.0
arrow                    1.2.3
asttokens                2.2.1
async-lru                2.0.2
attrs                    22.2.0
Babel                    2.11.0
backcall                 0.2.0
beautifulsoup4           4.11.2
bleach                   6.0.0
certifi                  2022.12.7
cffi                     1.15.1
charset-normalizer       3.0.1
colorama                 0.4.6
comm                     0.1.2
contourpy                1.1.0
cycler                   0.11.0
debugpy                  1.6.6
decorator                5.1.1
defusedxml               0.7.1
executing                1.2.0
fastjsonschema           2.16.2
fonttools                4.40.0
fqdn                     1.5.1
idna                     3.4
ipykernel                6.21.2
ipython                  8.10.0
[...]

PR #15736, opened 2024-02-03
Fix scrollbars partially blocking code completion

phil-zxx:
This pull request addresses issue https://github.com/jupyterlab/jupyterlab/issues/15704 (Code Completion Partially Hidden by Scrollbars).

### DIFF
--- a/packages/completer/style/base.css
+++ b/packages/completer/style/base.css
@@ -45,6 +45,7 @@
 }
 
 .jp-Completer-list {
+  overflow-x: hidden;
   margin: 0;
   padding: 0;
   list-style-type: none;


packages/completer/style/base.css
```diff
@@ -45,6 +45,7 @@
 }
 
 .jp-Completer-list {
+  overflow-x: hidden;
```
krassowski, line 48: This does not have any effect because there is a second `overflow-x` in line 53 (old line 52), which is what the failing lint job is warning about:

```
packages/completer/style/base.css
 53:3  ✖  Unexpected duplicate "overflow-x"  declaration-block-no-duplicate-properties
```

Can you instead change the line 52/53?

krassowski (changes requested):
Thanks for starting this @phil-zxx!

packages/completer/style/base.css
phil-zxx: Very good point. Apologies for this oversight.

krassowski (approved):
Thank you @phil-zxx.

krassowski:

PR #15751, opened 2024-02-06
Fix outputs preservation on splitting cells

### LINKED ISSUE #15748: Cell outputs lost after splitting a cell
## Description
Cell outputs are lost after a cell is split

## Reproduce
Write a multiline cell and split the cell using the context menu option

## Expected behavior
I would expect the cell outputs to be retained similar to lab3

## Context

- JupyterLab v4.1.0rc1

krassowski:
## References

Fixes #15748

## Code changes

- [x] add unit test for outputs being preserved in the last cell (should fail in the first commit)
- [ ] fix the split action to preserve the outputs

## User-facing changes

| Before | After |
|--|--|
| before | after |

None


fcollonval (approved):
Thanks @krassowski

krassowski:

PR #15753, opened 2024-02-06
Fix completer auto-invoking on non-source changes

### LINKED ISSUE #15746: Inline completer is invoked automatically when enabling output scrolling
Untitled

Inline completer is invoked automatically when enabling output scrolling. I think it should be in the way where inline completer is invoked only when editing code.

krassowski:
## References

Fixes #15746

## Code changes

- [x] add a test ensuring non-source changes do not trigger completer (should fail in the first commit)
- [x] filter the `changed` signal down to source changes before calling `onTextChanged`

## User-facing changes

| Before | After |
|--|--|
| before | after gif |


jtpio (approved):
Thanks!

Issue #15461, opened 2023-11-30
Unable to run Jupyter Lab after installing it on a read-only filesystem


zimmerrol:
<!-- Welcome! Thank you for contributing. These HTML comments will not render in the issue.

Before creating a new issue:
* Search for relevant issues
* Follow the issue reporting guidelines:
https://jupyterlab.readthedocs.io/en/latest/getting_started/issue.html
-->

## Description

I installed jupyter lab in a conda environment (of a singularity container). After installation, the location to which jupyter is installed becomes read-only. When I now try to start jupyter lab, I run into an error in the config manager of jupyter: ` Read-only file system: '/opt/conda/etc/jupyter/labconfig/page_config.json'`. I then saw that setting the `JUPYTER_CONFIG_DIR` to a writeable directory, might fix the issue. However, this does not change anything. Everything works fine when using jupyter notebook instead of jupyter lab.

## Reproduce

1. Install jupyter lab in a conda environment (of a singularity container). After installation, the location to which jupyter is installed becomes read-only. 
2. When I now try to start jupyter lab, I run into an error in the config manager of jupyter.

<!--Describe step-by-step instructions to reproduce the behavior-->


<!--Describe how you diagnosed the issue. See the guidelines at
 https://jupyterlab.readthedocs.io/en/latest/getting_started/issue.html -->

## Expected behavior

<!--Describe what you expected to happen-->
The server is supposed to start without crashing.

## Context

<!--Complete the following for context, and add any other relevant context-->

- Operating System and version: Debian GNU/Linux 11 <!-- e.g. Linux Ubuntu 21.04 -->
- Browser and version: n.A. <!-- e.g. Chrome 92 -->
- JupyterLab version:  4.1.0a3 <!-- e.g. 3.1.7 -->

<!--The more content you provide, the more we can help!-->

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
        /opt/conda/bin
        /usr/local/bin                                                                                                                                                                                                                                             
        /usr/sbin
        /usr/bin
        /sbin
        /bin
        /home/rzimmermann/bin

sys.path:
        /opt/conda/bin
        /opt/conda/lib/python311.zip
        /opt/conda/lib/python3.11
        /opt/conda/lib/python3.11/lib-dynload
        /opt/conda/lib/python3.11/site-packages

sys.executable:
        /opt/conda/bin/python

sys.version:
[...]

PR #15756, opened 2024-02-06
Catch `OSError` in addition to `PermissionError` when `sys_prefix` is read-only

dhml:
## Code changes

Catch `OSError` in addition to `PermissionError` when `sys_prefix` is read-only in `_AppHandler.__init__()`. (#15461)


krassowski (approved):
Thank you @dhml!

PR #15762, opened 2024-02-07
Fix migration of command selectors for shortcuts

krassowski:
## References

Fixes https://github.com/jupyterlab-contrib/jupyterlab-vim/issues/134 (but jupyterlab-vim should still migrate to new selectors)

## Code changes

Decouples the check for warning being already emitted from the actual selector migration.

## User-facing changes

All shortcuts that needed migration will work in JupyterLab 4.1.x

None


jtpio (approved):
Thanks!

PR #15767, opened 2024-02-07
Update `jupyter-collaboration` for the Binder environment

jtpio:
## References

Trying to update to the latest stable release of `jupyter-collaboration`, as this may be useful for https://github.com/jupyterlab/team-compass/issues/180

## Code changes

Update Binder dependencies.

## User-facing changes

Folks can use the latest `jupyter-collaboration` when using the dev Binder.

None

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -55,11 +55,5 @@ dependencies:
   - xeus-python
   # extra conda stuff
   - jsonschema-with-format-nongpl
-  # jupyter-collaboration 1.2.0
-  - jupyter_ydoc >=1.0.1,<2.0.0
-  - ypy-websocket >=0.12.1,<0.13.0
+  - jupyter-collaboration >= 2.0.1
   - jupyter_events >=0.7.0
-  - jupyter_server_fileid >=0.7.0,<1
-  - pip:
-    # can't be installed with `conda` in binder, due to `jupyterlab-link-share``
-    - jupyter-collaboration ==1.2.0


binder/environment.yml
```diff
@@ -55,11 +55,5 @@ dependencies:
   - xeus-python
   # extra conda stuff
   - jsonschema-with-format-nongpl
-  # jupyter-collaboration 1.2.0
-  - jupyter_ydoc >=1.0.1,<2.0.0
-  - ypy-websocket >=0.12.1,<0.13.0
```
jtpio, line 60: cc @davidbrochart looks like we could remove these dependencies and just depend on `jupyter-collaboration` directly?

jtpio:
Ah looks like Binder is having issues to build JupyterLab:

image

jtpio:
Looks like the Binder build failure might be a bit random. This time it was able to build and RTC seems to be enabled as expected:

image

binder/environment.yml
davidbrochart: Indeed, that's how it should be.

jtpio:
`jupyter-collaboration` includes a command to generate a share link:

image

davidbrochart:
I just tried this PR's binder, and it fails with a simple notebook cell like:
```py
from time import sleep

for i in range(1000):
    print(i)
    sleep(1)
```
Outputs stop being displayed after a while. This should be fixed by https://github.com/jupyterlab/jupyter-collaboration/pull/232, but it's not released yet. There are breaking changes (e.g. non-private methods are renamed), but I was hesitant to make a v3.0.0 release, since it's essentially bug fixes. What do you think?

jtpio:
> There are breaking changes (e.g. non-private methods are renamed), but I was hesitant to make a v3.0.0 release, since it's essentially bug fixes. What do you think?

It probably depends if we expect users to depend on that package and potentially extend it? In other words if these non-private methods are expected to be part of the API surface.

An alternative would be to check whether these breaking changes are necessary, and if the methods could be named as before while keeping the functionality the same?

davidbrochart:
Although these methods didn't get a leading `_`, I don't really expect users to use them, but who knows. Also, I didn't write all of them, so I don't know what was the intention of the author.
I don't think keeping the same names is a good idea, as they were confusing.
I would go with a patch release, if that is fine.

jtpio:
>  I would go with a patch release, if that is fine.

OK then maybe we document this decision in an issue on https://github.com/jupyterlab/jupyter-collaboration since this is more related to the `jupyter-collaboration` rather than this PR? That way it will be easier to understand the motivations for a patch release later.
Or by editing the changelog when the GitHub draft release is created by the releaser, so it ends up in the changelog and is more visible to users.

binder/environment.yml
krassowski: Binder resolved `ypy-websocket` to 0.8.2 - the previous range required a newer version (`>=0.12.1,<0.13.0`) - I guess this is ok because it works, but is this intended? 

image

krassowski (approved):
Thanks!

binder/environment.yml
davidbrochart: Still curious why `ypy-websocket` is installed?

jtpio: Is there a mamba command to check which package may depend on `ypy-websocket`?

jtpio: Wondering if this could be because of some of the pre-installed packages in the base Binder environment.

davidbrochart: > Is there a mamba command to check which package may depend on `ypy-websocket`?

I wanted to check, but the binder for this PR is not available anymore?

jtpio: We could use the one from `main` now: https://mybinder.org/v2/gh/jupyterlab/jupyterlab/main

(although it will have to build first)

jtpio: > Is there a mamba command to check which package may depend on `ypy-websocket`?

This one probably:

```
mamba repoquery whoneeds ypy-websocket
```

davidbrochart: I couldn't launch a binder, but at least from `main` no ypy package has been installed after installing jupyter-collaboration.

PR #15774, opened 2024-02-08
Fix typing in editable elements inside of open shadow DOM

### LINKED ISSUE #15757: Typing into into fields that are wrapped with shadow DOM is intercepted by commands
## Description

Typing most characters into fields wrapped in shadow DOM, such as widgets generated by panel, is impossible because command shortcuts intercept the keystrokes.

## Reproduce

See https://github.com/holoviz/panel/issues/6314, but briefly:

```python
import panel as pn
pn.extension()
text_input = pn.widgets.TextInput(name='Text Input', placeholder='Enter a string starting with "A" or "B" here...')
text_input
```

and try to type <kbd>A</kbd>, <kbd>C</kbd>, <kbd>X</kbd>, etc.

Dependency-free reproducer:

```python
from IPython.display import HTML
HTML("""
<div id='shadow-test'></div>
<script>
{
const shadowHost = document.querySelector('#shadow-test');
const shadowRoot = shadowHost.attachShadow({mode: 'open'});
shadowRoot.innerHTML = '<input/>';
}
</script>
""")
```

Note: iframes work as they also isolate keystrokes.

<details>

```python
from IPython.display import HTML
HTML("<iframe srcdoc='<input>'></iframe>")
```

```python
from IPython.display import HTML
HTML("<iframe srcdoc='<div><template shadowrootmode=\"open\"><input></template></div>' />")
```

</details>

## Expected behavior

Typing works as in 4.0 and earlier.

## Context

- JupyterLab version: 4.1.0

krassowski:
## References

Fixes #15757

## Code changes

- adds an integration test ensuring that typing works in open shadow DOM (should fail in the first commit)
- adds a new `jp-mod-readWrite` class that is added to the notebook node on focus change when the active element (or active element within active shadow DOM) is editable
- updates selectors for command-mode commands accordingly
- updates migration script and migration documentation to reflect the above
- documents the `lm-suppress-shortcuts` data attribute
- `hasActiveEditableElement` utility gets added to `DOMUtils`; this function checks the current `activeElement` (and all elements within active shadow DOM) against `:read-write` selector, but in the future could be extended for additional heuristics such as checking if the active element has a key event handler.

## User-facing changes

| Before | After |
|--|--|
| [before.webm](https://github.com/jupyterlab/jupyterlab/assets/5832902/ac7668a2-1531-4e44-b619-77478f3db358) | [after.webm](https://github.com/jupyterlab/jupyterlab/assets/5832902/c0fe312c-44e7-436e-be90-adac0b52e933) |

None

### DIFF
--- a/docs/source/extension/notebook.rst
+++ b/docs/source/extension/notebook.rst
@@ -190,12 +190,14 @@ enabling quick access to cell- and notebook-specific actions.
 These shortcuts are only active when the notebook is in command mode
 and the active element is non-editable,
 as signalled by absence of ``.jp-mod-readWrite`` class on the notebook node.
-This class is set if active element is editable as ascertained by matching
+This class is set if the active element is editable as ascertained by matching
 to the ``:read-write`` pseudo-selector, and accounts for any elements nested
 in the open shadow DOM, but not for the closed shadow DOM nor non-editable
 elements with custom key event handlers (such as
 ``<div contenteditable="false" onkeydown="alert()" tabindex="0"></div>``).
-If your output widget uses closed shadow DOM or non-editable elements with custom
+If your output widget (for example created with ``IPython.display.HTML``,
+or created by your MIME renderer on cell output in a notebook or console)
+uses closed shadow DOM or non-editable elements with custom
 key event handlers, you may wish to set ``lm-suppress-shortcuts`` data attribute
 on the host element to prevent side-effects from the command-mode actions, e.g:
 
--- a/docs/source/extension/notebook.rst
+++ b/docs/source/extension/notebook.rst
@@ -175,6 +175,41 @@ The ipywidgets widget manager is an example of an extension that adds a
 notebook-specific renderer, since rendering a widget depends on
 notebook-specific widget state.
 
+Keyboard interaction model
+""""""""""""""""""""""""""
+
+Multiple elements can receive focus in the Notebook:
+- the main toolbar,
+- cells,
+- cell components (editor, toolbar, outputs).
+
+When the focus is outside of the cell input editor,
+the Notebook switches to so-called "command" mode.
+In the command mode additional keyboard shortcuts are accessible to the user,
+enabling quick access to cell- and notebook-specific actions.
+These shortcuts are only active when the notebook is in command mode
+and the active element is non-editable,
+as signalled by absence of ``.jp-mod-readWrite`` class on the notebook node.
+This class is set if active element is editable as ascertained by matching
+to the ``:read-write`` pseudo-selector, and accounts for any elements nested
+in the open shadow DOM, but not for the closed shadow DOM nor non-editable
+elements with custom key event handlers (such as
+``<div contenteditable="false" onkeydown="alert()" tabindex="0"></div>``).
+If your output widget uses closed shadow DOM or non-editable elements with custom
+key event handlers, you may wish to set ``lm-suppress-shortcuts`` data attribute
+on the host element to prevent side-effects from the command-mode actions, e.g:
+
+.. code:: html
+
+   <div
+     contenteditable="false"
+     onkeydown="alert()"
+     tabindex="1"
+     data-lm-suppress-shortcuts="true"
+   >
+     Click on me and press "A" with and without "lm-suppress-shortcuts"
+   </div>
+
 .. _extend-notebook-plugin:
 
 How to extend the Notebook plugin
--- a/examples/notebook/src/commands.ts
+++ b/examples/notebook/src/commands.ts
@@ -396,17 +396,17 @@ export const setupCommands = (
       command: COMMAND_IDS.findPrevious
     },
     {
-      selector: '.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)',
+      selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus',
       keys: ['I', 'I'],
       command: COMMAND_IDS.interrupt
     },
     {
-      selector: '.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)',
+      selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus',
       keys: ['0', '0'],
       command: COMMAND_IDS.restart
     },
     {
-      selector: '.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)',
+      selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus',
       keys: ['Enter'],
       command: COMMAND_IDS.editMode
     },
@@ -416,7 +416,7 @@ export const setupCommands = (
       command: COMMAND_IDS.commandMode
     },
     {
-      selector: 'jp-Notebook.jp-mod-commandMode :focus:not(:read-write)',
+      selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus',
       keys: ['Shift M'],
       command: COMMAND_IDS.merge
     },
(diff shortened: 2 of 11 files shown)


examples/notebook/src/commands.ts
```diff
@@ -396,17 +396,17 @@ export const setupCommands = (
       command: COMMAND_IDS.findPrevious
     },
     {
-      selector: '.jp-Notebook.jp-mod-commandMode :focus:not(:read-write)',
+      selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus',
```
krassowski, line 399: Question: should we keep the extra `not(:read-write)`? No functional difference, but if user downgrades to 4.1.0 for some reason this would be compatible.

packages/mainmenu-extension/schema/plugin.json
```diff
@@ -270,22 +270,22 @@
     {
       "command": "kernelmenu:interrupt",
       "keys": ["I", "I"],
-      "selector": "[data-jp-kernel-user] :focus:not(:read-write)"
+      "selector": "[data-jp-kernel-user]:not(.jp-mod-readWrite) :focus:not(:read-write)"
```
brichet, line 273: Is there a reason to keep the `:not(read-write)` here ?

packages/notebook/src/widget.ts
```diff
@@ -2771,10 +2779,26 @@ export class Notebook extends StaticNotebook {
...
+   */
+  private _updateReadWrite(): void {
+    const inReadWrite = DOMUtils.hasActiveEditableElement(this.node);
+    const hasReadWriteClass = this.node.classList.contains(READ_WRITE_CLASS);
+    if (inReadWrite && !hasReadWriteClass) {
+      this.node.classList.add(READ_WRITE_CLASS);
+    } else if (!inReadWrite && hasReadWriteClass) {
+      this.node.classList.remove(READ_WRITE_CLASS);
+    }
+  }
```
brichet, line 2793: It can be simplified with:
```ts
this.node.classList.toggle(READ_WRITE_CLASS, inReadWrite);
```

brichet (commented):
Thanks @krassowski, this looks good to me.

examples/notebook/src/commands.ts
brichet: Not sure, since both versions are documented in the extension migration guide.

packages/mainmenu-extension/schema/plugin.json
krassowski: Yes, I added it back because extensions may have  elements which are kernel runners but would not have time to implement `.jp-mod-readWrite`. Basically out of abundance of caution to not break anything in the wild.

packages/apputils/src/domutils.ts
```diff
@@ -50,4 +50,22 @@ export namespace DOMUtils {
...
+  /**
+   * Check whether the active element descendant from given parent is editable.
+   * When checking active elements it includes elements in the open shadow DOM.
+   */
+  export function hasActiveEditableElement(
+    parent: Node | DocumentFragment,
+    root: ShadowRoot | Document = document
+  ): boolean {
+    const element = root.activeElement;
+    return !!(
```
gabalafou, line 63: For JavaScript purist points:

```suggestion
    return Boolean(
```

packages/console/src/widget.ts
```diff
@@ -846,6 +879,20 @@ export class CodeConsole extends Widget {
...
+   * Update the console node with class indicating read-write state.
+   */
+  private _updateReadWrite(): void {
+    // TODO: de-duplicate with code in console/src/widget.ts
+    const inReadWrite = DOMUtils.hasActiveEditableElement(this.node);
+    const hasReadWriteClass = this.node.classList.contains(READ_WRITE_CLASS);
+    if (inReadWrite && !hasReadWriteClass) {
+      this.node.classList.add(READ_WRITE_CLASS);
+    } else if (!inReadWrite && hasReadWriteClass) {
+      this.node.classList.remove(READ_WRITE_CLASS);
```
gabalafou, line 892: Why can't you do 

```suggestion
    this.node.classList.toggle(READ_WRITE_CLASS, inReadWrite);
```

like you do in the notebook code?

packages/console/src/widget.ts
```diff
@@ -70,6 +70,13 @@ const CONTENT_CLASS = 'jp-CodeConsole-content';
  */
 const INPUT_CLASS = 'jp-CodeConsole-input';
 
+/**
+ * The class name added to the notebook when the focused notebook element takes keyboard input.
```
gabalafou, line 74:
```suggestion
 * The class name added to the notebook when an element within it is focused and takes keyboard input, such as <input> and <div contenteditable> 
```

gabalafou (commented):
I'm relieved not to see the solution with extended ("/deep/") selectors.

This is a clever and elegant solution that builds on available methods without reinventing the wheel.

docs/source/extension/notebook.rst
```diff
@@ -175,6 +175,41 @@ The ipywidgets widget manager is an example of an extension that adds a
...
+.. code:: html
+
+   <div
+     contenteditable="false"
+     onkeydown="alert()"
+     tabindex="1"
+     data-lm-suppress-shortcuts="true"
+   >
+     Click on me and press "A" with and without "lm-suppress-shortcuts"
+   </div>
```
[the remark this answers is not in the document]
brichet: It also works in a Notebook cell with `IPython.display.HTML()`, perhaps we can specify it.

docs/source/extension/notebook.rst
```diff
@@ -175,6 +175,41 @@ The ipywidgets widget manager is an example of an extension that adds a
...
+enabling quick access to cell- and notebook-specific actions.
+These shortcuts are only active when the notebook is in command mode
+and the active element is non-editable,
+as signalled by absence of ``.jp-mod-readWrite`` class on the notebook node.
+This class is set if active element is editable as ascertained by matching
+to the ``:read-write`` pseudo-selector, and accounts for any elements nested
+in the open shadow DOM, but not for the closed shadow DOM nor non-editable
+elements with custom key event handlers (such as
+``<div contenteditable="false" onkeydown="alert()" tabindex="0"></div>``).
+If your output widget uses closed shadow DOM or non-editable elements with custom
```
krassowski, line 198:
```suggestion
If your output widget (for example created with ``IPython.display.HTML``,
or created by your MIME renderer on cell output in a notebook or console)
uses closed shadow DOM or non-editable elements with custom
```

packages/notebook/src/widget.ts
```diff
@@ -92,6 +93,13 @@ const OTHER_SELECTED_CLASS = 'jp-mod-multiSelected';
  */
 const UNCONFINED_CLASS = 'jp-mod-unconfined';
 
+/**
+ * The class name added to the notebook when the focused notebook element takes keyboard input.
```
krassowski, line 97:
```suggestion
 * The class name added to the notebook when an element within it is focused
 * and takes keyboard input, such as focused <input> or <div contenteditable>.
```

packages/apputils/src/domutils.ts
gabalafou: I was being facetious when I said "purist." I think it essentially comes down to a stylistic choice. But a lot of programmers just don't know about it. :)

packages/console/src/widget.ts
```diff
@@ -846,6 +880,16 @@ export class CodeConsole extends Widget {
...
   }
 
+  /**
+   * Update the console node with class indicating read-write state.
+   */
+  private _updateReadWrite(): void {
+    // TODO: de-duplicate with code in notebook/src/widget.ts
+    const inReadWrite = DOMUtils.hasActiveEditableElement(this.node);
+    this.node.classList.toggle(READ_WRITE_CLASS, inReadWrite);
+    }
```
krassowski, line 890:
```suggestion
```

krassowski:
Is this PR still needed if we merge https://github.com/jupyterlab/jupyterlab/pull/15790? Yes, because while it also fixes the typing part it does not block commands bound to shortcuts which do not change input, e.g. Ctrl + A:

still-needed

brichet (approved):

PR #15782, opened 2024-02-09
Restore notebook scrolling on dragging a cell to the viewport edge

### LINKED ISSUE #15777: Can't drag and drop cells out of viewport
## Description
In lab3, it was possible to drag and drop a cell out of the initial viewport because the viewport would scroll. This doesn't happen in lab4. Here's a gif showing the differences in behavior - 
Drap-n-drop

## Reproduce
shown in gif

## Expected behavior
I expect the behavior in lab4 to be the same as lab3

## Context

- JupyterLab v4.1.0rc1

krassowski:
## References

Fixes #15777

## Code changes

- [x] add integration test for scrolling on dragging cell to the top and bottom edges of notebook viewport (warranted because it regressed twice in the 4.x development)
   - note there is a copy for windowing and no-windowing; I do wish to reconciliate these duplicated test files but this requires a change in galata config so it will be a 4.2.0 PR; this also why the `dragCellTo` goes to private `utils` rather than to galata itself (for now!)
- [x] move the `data-lm-dragscroll` attribute from the main notebook widget node to the outer node which became the scroller in 4.1.0

## User-facing changes

| Before | After |
|--|--|
| before | after |

None

### DIFF
--- a/galata/test/jupyterlab/notebook-scroll-no-windowing.test.ts
+++ b/galata/test/jupyterlab/notebook-scroll-no-windowing.test.ts
@@ -5,6 +5,7 @@ import { expect, galata, test } from '@jupyterlab/galata';
 import * as path from 'path';
 
 import {
+  dragCellTo,
   notebookViewportRatio,
   positionCellPartiallyBelowViewport
 } from './utils';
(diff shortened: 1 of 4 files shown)


krassowski:
@afshin would you have a minute to take a look? (this steams form the virtual scrollbar PR).

fcollonval (approved):
Thanks @krassowski

PR #15783, opened 2024-02-09
Fix undo/redo in console, fix undo/redo enabled state in file editor

### LINKED ISSUE #8590: Code mirror editor could expose the history of undo/redo events
### Problem
Currently, code mirror editor provides `undo()`, `redo()` and `clearHistory()` methods in order to manipulate undo/redo commands.
It would be helpful if it could also expose the history of those events, so that could be used when adding related features in jupyterlab, such as #2778 

### Additional context
This concern came up when adding new commands to file editor context menu (see  #8425), where currently there isn'y a way to consult if undo/redo events are available in order to properly enable/disable those commands from the menu.
Therefore the buttons for those commands are enabled as default.

The same behaviour was noticed in notebook cells undo/redo commands, these are always enabled even when there is no event history for them, even though it is a different implementation.

### LINKED ISSUE #15773: Console input has unexpected undo behavior
## Description
Pressing Ctrl+Z for undo in a console input block has unexpected undo behavior. Here's a gif - 
weird-undo-bug

## Reproduce

Type something in the console input box and use Ctrl+Z for undo.

## Context

- JupyterLab version: 4.1.0rc1

krassowski:
## References

Fixes #15773, closes #8590

## Code changes

- make the console a `jp-undoer` in order to fix the undo/redo commands being dispatched to native CodeMirror rather than though yjs bindings which resulted in garbled code on undo
  - adds two new commands: `console:undo` and `console:redo` which get registered with edit menu
  - the commands are also added to context menu
- fix undo/redo commands of file editor being always enabled

## User-facing changes

1. Undo works correctly in console (not much to show)

2. Undo/redo menu entries are no longer permanently greyed out in console when these actions are available

| Before | After |
|--|--|
| image | image |

4. New context menu entry in console when over the cell:

| Before | After |
|--|--|
| image | Untitled |

5. In file editor the undo actions in edit menu and in context menu are no longer permanently enabled and instead are greyed out when undo/redo is not available

None


packages/console-extension/schema/tracker.json
```diff
@@ -68,6 +68,16 @@
         "command": "console:restart-kernel",
         "selector": ".jp-CodeConsole",
         "rank": 30
+      },
+      {
+        "command": "console:undo",
+        "selector": ".jp-CodeConsole-promptCell",
+        "rank": 1
```
jtpio, line 75: More for consistency: should we put these to the top of the list so they are ordered by rank?

jtpio (approved):
Thanks!

Left a minor comment, looks good otherwise :+1:

PR #15784, opened 2024-02-10
Add Edge, remove IE 11 from browser support docs

### LINKED ISSUE #8030: Supporting Edge officially?
Now that Edge uses the same underlying engine as Chrome, I think it may have enough features to support it officially (which we don't do now: https://jupyterlab.readthedocs.io/en/stable/getting_started/installation.html#supported-browsers)

JasonWeill:
Fixes #8030. Edge is now generally considered a modern, first-tier browser, with similar functionality to Chrome (and it uses the same engine). Removes mention of IE 11 and a very old version of Edge.


krassowski (approved):
:100:, thank you!

PR #15788, opened 2024-02-12
Catch errors attempting to access `document.cookie`

minrk:
behave as if cookie is empty if cookie cannot be read for sandbox reasons.

Attempting to access `document.cookie` can raise SecurityError if page is served with `Content-Security-Policy: sandbox`. Because this is accessed unconditionally, ~any page with JupyterLab js served with CSP sandbox crashes when attempting to build URLs instead of behaving as if xsrf token is unknown.

## References

[CSP: sandbox](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox)

## Code changes

If accessing `document.cookie` throws an error, behave as if cookie is empty, rather than error.

## User-facing changes

Loading JupyterLab javascript should no longer crash when loaded in a page with CSP: Sandbox. XSRF token will be unavailable, but is not required.

None

### DIFF
--- a/packages/services/src/serverconnection.ts
+++ b/packages/services/src/serverconnection.ts
@@ -306,7 +306,7 @@ namespace Private {
       authenticated = true;
       request.headers.append('Authorization', `token ${settings.token}`);
     }
-    if (typeof document !== 'undefined' && document?.cookie) {
+    if (typeof document !== 'undefined') {
       const xsrfToken = getCookie('_xsrf');
       if (xsrfToken !== undefined) {
         authenticated = true;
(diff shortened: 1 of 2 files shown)


packages/services/src/serverconnection.ts
```diff
@@ -334,7 +334,14 @@ namespace Private {
    */
   function getCookie(name: string): string | undefined {
     // From http://www.tornadoweb.org/en/stable/guide/security.html
-    const matches = document.cookie.match('\\b' + name + '=([^;]*)\\b');
+    let cookie = '';
+    try {
+      cookie = document.cookie;
```
minrk, line 339: it would also be a reasonable spec to let this raise here and catch where getCookie is called. It is called in exactly one place, to get the xsrf token.

krassowski (approved):
Thank you @minrk, it looks good to me.

I opened https://github.com/jupyterlab/jupyterlab/issues/15792 to track adding an end-to-end test as it does not seem easy to unit test in jest (nor would we cover all conditions with a unit test here).

Issue #15799, opened 2024-02-14
"Try JupyterLab on Binder" link in documentation outdated


nluetts:
## Reproduce

1. Go to https://jupyterlab.readthedocs.io/en/latest/ (or stable, same issue)
2. Click on "Try JupyterLab on Binder" (points at https://mybinder.org/v2/gh/jupyterlab/jupyterlab-demo/3818244?urlpath=lab/tree/demo)
3. On mybinder.org, wait for build to "finish"

> Error loading jupyterlab/jupyterlab-demo/3818244!
> See logs below for details.

(I paste the build log below)

## Expected behavior

The binder image builds correctly and one can try JupyterLab.

## Context

I think nothing else is relevant here, it seems that the link is just outdated.
If you go to 

https://github.com/jupyterlab/jupyterlab-demo

and click the badge "launch binder", which points at the latest commit of the master branch

https://mybinder.org/v2/gh/jupyterlab/jupyterlab-demo/master?urlpath=lab

the image launches correctly.

To make a long story short, the link in the docs should be updated.

Here is the build log of the currently linked binder:

```
Waiting for build to start...
Picked Git content provider.
Cloning into '/tmp/repo2dockerc72g6m6i'...
HEAD is now at 3818244 Install jupyter-offlinenotebook manually
Building conda environment for python=3.7
Using CondaBuildPack builder
Step 1/63 : FROM docker.io/library/buildpack-deps:jammy
 ---> 97e9d72362a2
Step 2/63 : ENV DEBIAN_FRONTEND=noninteractive
 ---> Using cache
 ---> bca4faa40313
Step 3/63 : RUN apt-get -qq update &&     apt-get -qq install --yes --no-install-recommends locales > /dev/null &&     apt-get -qq purge &&     apt-get -qq clean &&     rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> 8e085a29b3b0
Step 4/63 : RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen &&     locale-gen
 ---> Using cache
 ---> 48acb7ce1af0
Step 5/63 : ENV LC_ALL=en_US.UTF-8     LANG=en_US.UTF-8     LANGUAGE=en_US.UTF-8
 ---> Using cache
 ---> 3c1409b332fa
Step 6/63 : ENV SHELL=/bin/bash
 ---> Using cache
 ---> e8f65fa7294b
Step 7/63 : ARG NB_USER
[...]

PR #15800, opened 2024-02-14
Fix outdated link to mybinder.org on index page of documentation

nluetts:
This PR is in response to #15799

The current link "Try JupyterLab on Binder" on `index.rst` points to the commit 

https://github.com/jupyterlab/jupyterlab-demo/commit/3818244680a477cdead8182216f5c87e685e0bee

which is not producing a valid binder image anymore.

Here, the link is changed to point to HEAD, like in `README.md`, which yields a functioning binder session.


JasonWeill:
@nluetts Thank you for opening this pull request! It took me a few tries, but I was able to verify that the new link works properly.

JasonWeill (approved):

Issue #14678, opened 2023-06-13
File browser items in `focus` state draw a rectangle around the `itemText` node, which disappears when hovering on any menu


ibdafna:
## Description
File browser items in `focus` state draw a rectangle around the `itemText` node, which disappears when hovering on any menu

## Reproduce

Select any item in the file browser and then hover over the top menu, or right click on the currently selected file. See attached gif.

![jlab4x-selection-bug](https://github.com/jupyterlab/jupyterlab/assets/24281433/8d43a687-01fa-4c99-a6b1-757dac71bd24)

## Expected behavior

Two behaviors I'd expect:

1.  I understand the reasoning behind responding to the `focus` state in the file browser (#13577), but perhaps there is a better way of indicating focus than drawing this rectangle around the object.
2. The logic _appears_ inconsistent when hovering on menus. Does hover indicate focus? I would argue it does not unless the menu item is clicked on. If hovering does indicate `focus`, then hovering over other files in the browser, which are currently not selected or clicked on, should follow the same logic as hovering over menu items.

## Context

- Operating System and version: MacOS
- Browser and version: Chrome/Edge 114.0.5735.106 
- JupyterLab version: 4.0.2

ColinConwell:
Any updates on this design feature? Can the behavior be reverted easily to jupyter 3.6.3 style behavior?

krassowski:
> perhaps there is a better way of indicating focus than drawing this rectangle around the object

GNOME file browser uses a subtle border colour change to differentiate focused from selected. The slight challenge here is that we also have checkboxes (hidden by default) so I guess this is why it is not an outline over the entire item for only over the file name. I do agree that this is suboptimal and a bit distracting. Suggestions welcome!

ColinConwell:
Hmmm. I think at least my aesthetic issue could be resolved effectively with a css override that removes the outline properties from jupyter/packages/style/base.css? Or maybe I'm not quite getting the subtleties here?

.jp-DirListing-itemText:focus {
  outline-width: 2px;
  outline-color: var(--jp-inverse-layout-color1);
  outline-style: solid;
  outline-offset: 1px;
}

.jp-DirListing-item.jp-mod-selected .jp-DirListing-itemText:focus {
  outline-color: var(--jp-layout-color1);
}

ColinConwell:
I got rid of this by running a manual override in my web environment. Would there be a way to inject this when loading jupyterlab?

function overrideStyles() {
[...]

PR #15802, opened 2024-02-14
Only show focus ring around file name when needed

krassowski:
## References

Alleviates #14678

## Code changes

```diff
- :focus
+ :focus-visible
```

[:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) is widely available across all modern browsers.

## User-facing changes

| Before | After |
|--|--|
| focus-always | no-focus-ring-on-click |

None

### DIFF
--- a/packages/filebrowser/style/base.css
+++ b/packages/filebrowser/style/base.css
@@ -276,14 +276,14 @@
   user-select: none;
 }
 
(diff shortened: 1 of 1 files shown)


krassowski:
Superseded by https://github.com/jupyterlab/jupyterlab/pull/15860.

PR #15803, opened 2024-02-15
Fix highlight sequencing when replacing text in code cells

### LINKED ISSUE #15657: When using find and replace, the next string is selected, but the previous is replaced
## Description
When I use find and replace, as one string is replaced, the next occurrence is highlighted.
However, if I press replace again, the previous occurrence is replaced, not the one highlighted.

I've confirmed this issue with a clean install and in incognito mode on chrome.

## Reproduce

Open "Find and replace".
Type the string to be found.
Type the replacement string.
Use the up and down buttons to select the first occurrence you want to replace.
Press replace. The string is replaced as expected and the highlight moves to the next occurence.
Press replace again. The highlight moves again, but the replaced occurrence is the one **before** the first replacement.
This will happen every time I press replace.

image

## Expected behavior

The highlighted string is replaced.

## Context
- Operating System and version: Linux Ubuntu 20.04
- Browser and version: Chrome 120
- JupyterLab version: 4.0.7

JasonWeill:
## References

Fixes #15657.

## Code changes

After replacing a match in the middle of a code cell, with matches before and after the cursor position, the next highlight (if any remain) is selected.

Adds a notebook search unit test, which indirectly invokes the search provider, that starts in the middle of a cell.

## User-facing changes

Start a code cell with:

```
29, 29, 29, 29, 29, 29
```

Put the cursor before the 2 in the third `29`. Search for `29` and replace with `86`. Replace one instance. You should see:

```
29, 29, 86, 29, 29, 29
```

… with the fourth item (the third 29) selected. Replace another instance. You should see:

```
29, 29, 86, 86, 29, 29
```

Repeat until all items are replaced. No item should be selected.

```
86, 86, 86, 86, 86, 86
```

None.

### DIFF
--- a/packages/codemirror/src/searchprovider.ts
+++ b/packages/codemirror/src/searchprovider.ts
@@ -321,7 +321,7 @@ export abstract class EditorSearchProvider<
    * calling `replaceCurrentMatch()` attribute `this.currentIndex` is null.
    * It is necesary to let the caller handle highlighting because this
    * method is used in composition pattern (search engine of notebook cells)
-   * and highligthing on the composer (notebook) level needs to switch to next
+   * and highlighting on the composer (notebook) level needs to switch to next
    * engine (cell) with matches.
    *
    * @param newText The replacement text.
--- a/packages/codemirror/src/searchprovider.ts
+++ b/packages/codemirror/src/searchprovider.ts
@@ -348,10 +348,24 @@ export abstract class EditorSearchProvider<
         this.currentIndex = null;
       } else {
         this.cmHandler.matches.splice(this.currentIndex, 1);
-        this.currentIndex =
-          this.currentIndex < this.cmHandler.matches.length
-            ? Math.max(this.currentIndex - 1, 0)
-            : null;
+        const cmMatchesRemaining = this.cmHandler.matches.length;
+
+        // If there are no remaining matches, make nothing selected.
+        if (cmMatchesRemaining === 0) {
+          this.currentIndex = null;
+        }
+        else {
+        // Move to the next match, wrapping around if necessary.
+        if (loop) {
+            this.currentIndex = (this.currentIndex) % cmMatchesRemaining;
+          }
+          else {
+            // End at the end of the CodeMirror matches list; do not loop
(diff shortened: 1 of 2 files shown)


krassowski (commented):
Thank you for working on it! I think we need more nuanced approach, taking into account that the codemirror search provider is used as one of the engines for notebook search provider, as described in the docstring:

> The caller of this method is expected to call `highlightNext` if after calling `replaceCurrentMatch()` attribute `this.currentIndex` is null.
> It is necessary to let the caller handle highlighting because this method is used in composition pattern (search engine of notebook cells) and highlighting on the composer (notebook) level needs to switch to next engine (cell) with matches.

So in place where this PR adds looping on `currentIndex` I think it should set it to `null` if called from compositor (notebook). This is because given the current implementation leads to something like:

wrong-order-cross-cells

replacing numbers 12**6**345 whereas I would expect 123456.

But I do recognise that it might not be obvious to know whether the engine is used standalone or from notebook. Maybe the loop argument should be set to a special value by the notebook when it calls the cell?

I am not necessarily requesting changes in this PR because it does improve situation and the wrong order is a pre-existing issue, however the sequencing in notebook is still broken even with this fix - more work is needed. If you do not plan to work on it more in this PR, let's open an issue to track the remaining problem with sequencing.

JasonWeill:
@krassowski Thanks! I had seen the comment, but didn't connect that to the reason why the `loop` variable wasn't being used. I can correct that and test again. I appreciate your feedback!

JasonWeill:
@krassowski I simplified the logic in `codemirror/…/searchprovider.ts` to return `null` at the end of the current code cell, but I'm still seeing the same behavior as what you observed above. Should we merge this in and have a separate issue to further improve the highlight sequencing logic?

pr-after-update

krassowski:
Kicking the CI because for some reason it did not run:

image

krassowski (approved):
Yes, thank you @JasonWeill!

I opened https://github.com/jupyterlab/jupyterlab/issues/15809 to track the notebook sequencing issue.

krassowski:

PR #15805, opened 2024-02-15
Fix codemirror highlight for Python builtin

### LINKED ISSUE #14589: Syntax highlighting for Python built-in functions does not work with JupyterLab 4.0.0
## Description

Syntax highlighting for Python built-in functions does not work with JupyterLab 4.0.0. It worked in previous versions as seen here:

image

@krassowski noted in https://github.com/jupyterlab/jupyterlab/issues/14584#issuecomment-1561783918 that this was likely caused by the migration from CodeMirror5 to CodeMirror6 and that a custom solution is necessary since upstream is not interested in supporting built-in functions as per https://github.com/lezer-parser/python/issues/9 and https://github.com/lezer-parser/python/pull/13.

## Reproduce

Open any notebook with JupyterLab 4.0.0 and type any [Python built-in functions](https://docs.python.org/3/library/functions.html), such as print(), len(), list(), or int(), into a code block.

## Expected behavior

Python built-in functions are highlighted in some way similar to previous JupyterLab versions.

## Context

- Operating System and version: Windows 11 22H2
- Browser and version: Microsoft Edge 113.0.1774.35
- JupyterLab version: 4.0.0

AllanChain:
## References

- Fixes #14589

## Code changes

Added a CodeMirror `ViewPlugin` to highlight Python built-in based on https://github.com/codemirror/language/blob/3c9ad8ffee32e158512d252fee9457a0e6469e2e/src/highlight.ts#L155

## User-facing changes

Built-in functions get highlighted:

|Before|After|
|:----:|:---:|
|before|after|

None


AllanChain:

galata/test/jupyterlab/metadataform.test.ts-snapshots/metadata-required-filled-jupyterlab-linux.png
AllanChain, line 1: I don't know why this snapshot got updated. It seems unrelated to my changes.

AllanChain:
@meeseeksdev tag bug

galata/test/jupyterlab/metadataform.test.ts-snapshots/metadata-required-filled-jupyterlab-linux.png
AllanChain: Probably due to the flakiness of the tests. Manually reverted and let's see if that's OK.

packages/codemirror/src/theme.ts
```diff
@@ -156,7 +157,8 @@ export const jupyterHighlightStyle = HighlightStyle.define([
  */
 export const jupyterTheme: Extension = [
   jupyterEditorTheme,
-  syntaxHighlighting(jupyterHighlightStyle)
+  syntaxHighlighting(jupyterHighlightStyle),
+  ViewPlugin.fromClass(PythonBuiltin, { decorations: v => v.decorations })
```
fcollonval, line 161: This is not the best place for loading the extension as ideally we should load it as a sidecar extension when using the Python language parser.

To achieve that, you can modify at 

https://github.com/jupyterlab/jupyterlab/blob/e40ed2ecd299e70c95dea1cd4ef233dfd42cee92/packages/codemirror/src/language.ts#L474-L477

```patch
-    return m.python(); 
+    return new LanguageSupport(m.pythonLanguage, [
+      ViewPlugin.fromClass(PythonBuiltin, { decorations: v => v.decorations })
+    ])
```

You will need to apply the same trick at https://github.com/jupyterlab/jupyterlab/blob/e40ed2ecd299e70c95dea1cd4ef233dfd42cee92/packages/codemirror/src/language.ts#L488

packages/codemirror/style/base.css
```diff
@@ -151,3 +151,7 @@
   opacity: 1;
   transition-delay: 0s;
 }
+
+.cm-builtin {
+  color: var(--jp-mirror-editor-builtin-color);
+}
```
fcollonval, line 157: This rule will be better set in the Jupyter codemirror theme at https://github.com/jupyterlab/jupyterlab/blob/e40ed2ecd299e70c95dea1cd4ef233dfd42cee92/packages/codemirror/src/theme.ts#L15

AllanChain:
The UI tests look flaky, but I think all snapshots are correctly updated. I'm marking it ready for review.

packages/codemirror/src/pythonBuiltin.ts
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { pythonLanguage } from '@codemirror/lang-python';
```
fcollonval, line 6: This is a bit annoying as we try to only load language parser dynamically.

One way to achieve this would be to pass it at construction time using 

`ViewPlugin.define(view => new PythonBuiltin(view, pythonLanguage), { decorations: v => v.decorations })`

But first you should move the extension creation. Then this suggestion will be easier to implement.

fcollonval (changes requested):
Thanks a lot @AllanChain 

This is awesome. I only have minor comments to load the new extension and style rule in more adequate places. I also have an performance suggestion.

packages/codemirror/src/pythonBuiltin.ts
AllanChain: The import from `@codemirror/lang-python` is no longer needed after the extension is moved :+1:. It's just used to detect the current language

packages/codemirror/src/language.ts
```diff
@@ -473,7 +474,11 @@ export namespace EditorLanguageRegistry {
         filename: /^(BUCK|BUILD)$/,
         async load() {
           const m = await import('@codemirror/lang-python');
-          return m.python();
+          const pythonSupport = m.python();
+          return new LanguageSupport(m.pythonLanguage, [
+            pythonSupport.support,
```
fcollonval, line 479: Calling `m.python()` should not be need as basically it creates the `LanguageSupport` and we don't care about completion (because JupyterLab brings its own mechanism).

```py
export function python() {
  return new LanguageSupport(pythonLanguage, [
    pythonLanguage.data.of({autocomplete: localCompletionSource}),
    pythonLanguage.data.of({autocomplete: globalCompletion}),
  ])
}
```

Ref https://github.com/codemirror/lang-python/blob/ce4804b2163180989a95d368ffdfa68d48f8d991/src/python.ts#L75

```suggestion
          return new LanguageSupport(m.pythonLanguage, [
```

fcollonval (changes requested):
Thanks for the quick update @AllanChain 

I have another optimization. And a request for testing a edge case:

How is the editor rendering markdown such as:

<pre>
# Title

```py
print("hello")
```

Some built in python like print
</pre>

If the last _print_ is highlighted, the test `if (!pythonLanguage.isActiveAt(view.state, node.from)) return;` cannot be avoided - I guess.

AllanChain:
> How is the editor rendering markdown such as

That's a good point. Currently, it doesn't work inside Markdown:

image

Nested languages are handled differently in CodeMirror and I need to read more about its documentation to figure out how to make it work inside Markdown.

krassowski:
@fcollonval do you think this is this good to go now?

packages/codemirror-extension/src/services.tsx
```diff
@@ -62,17 +63,22 @@ export const languagePlugin: JupyterFrontEndPlugin<IEditorLanguageRegistry> = {
...
+        const mdlang = m.markdown({
           base: m.markdownLanguage,
           codeLanguages: (info: string) => languages.findBest(info) as any,
           extensions: [
             parseMathIPython(StreamLanguage.define(tex.stexMath).parser)
           ]
         });
+        return new LanguageSupport(
+          mdlang.language,
+          pythonBuiltin(py.pythonLanguage)
```
fcollonval, line 80: Nice, you are only missing to add the default support from markdown:

```suggestion
          [mdlang.support, pythonBuiltin(py.pythonLanguage)]
```

And this is my final comment for this PR.

fcollonval (changes requested):
Final suggestion from me @AllanChain :wink:

AllanChain:
Thank @fcollonval for the review. The code has been well improved with your help!

jtpio:
Thanks @AllanChain!

cc @fcollonval if you would like to take a final look, thanks!

fcollonval (approved):
Thanks a lot @AllanChain 

cc @krassowski sorry for the late review.

jtpio:

jtpio:

AllanChain:
> There seem to be a conflict, please backport manually. Here are approximate instructions:

Hi, @jtpio! I'm not familiar with the workflow, but may I ask if I need to backport it to `4.0.x` manually, or it's the maintainers' job?

krassowski:
@AllanChain feel welcome to open a backport PR following the instructions above

PR #15825, opened 2024-02-19
Adds title to close button of search and replace box

### LINKED ISSUE #15815: Search and replace close button has no tooltip
## Description

The Search and Replace pane's close button has no tooltip on hover.

image

## Expected behavior

An informative tooltip, such as "Close search and replace", appears on hover.

## Context

- JupyterLab version: 4.1.0

This is adapted from #6650, sub-issue 2, item 4, originally opened by @tgeorgeux.

JasonWeill:
## References

Fixes #15815.

## Code changes

Adds tooltip (title) to the close button of the search and replace box.

## User-facing changes

The tooltip "Close Search and Replace" now appears when the user hovers over or focuses the close button on the find and replace panel.

image

None.

### DIFF
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -729,7 +729,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
             className={BUTTON_WRAPPER_CLASS}
             onClick={() => this._onClose()}
             tabIndex={0}
-            title={trans.__('Close Search and Replace')}
+            title={trans.__('Close Search Box')}
           >
             <closeIcon.react
               className="jp-icon-hover"
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -729,6 +729,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
             className={BUTTON_WRAPPER_CLASS}
             onClick={() => this._onClose()}
             tabIndex={0}
+            title={trans.__('Close Search and Replace')}
           >
             <closeIcon.react
               className="jp-icon-hover"


packages/documentsearch/src/searchview.tsx
```diff
@@ -729,6 +729,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
             className={BUTTON_WRAPPER_CLASS}
             onClick={() => this._onClose()}
             tabIndex={0}
+            title={trans.__('Close Search and Replace')}
```
krassowski, line 732:
```suggestion
            title={trans.__('Close Search Box')}
```

For me this sounds ambiguous ("Close Search and (then) Replace" vs Close "Search and Replace"). Even if I am silly here, it would still be an ellipsis for "Close Search and Replace **Box**". This is different from having a "Search and Replace" entry in menu where it is a set of action verbs for what user does intend to do.

I feel strongly about this but not a deal breaker - if you disagree I will still approve it later in the week.

JasonWeill: Although "search box" is used in the code, is this convention used in the docs? I'm concerned about how well this would be translated or understood. My second choice would be "pane", but that has a specific meaning in JupyterLab's UI layout that doesn't apply to the search/replace overlay.

krassowski: > Although "search box" is used in the code, is this convention used in the docs?

The docs for search box are still to be written so this is not the best question to ask I guess ;)

I believe search box is the most widely recognised term but https://en.wikipedia.org/wiki/Search_box offers some alternatives: search field, search bar.  Some data points from ecosystem:
- VSCode: `Close (Escape)`
- PyChar: `Close Escape`
- Firefox: `Close find bar` (uses a full-width bar, does not have replace function)
- Chrome: `Close find bar` (but uses a box - huh? no replace function)
- Gnome notepad: `Close search` (even when replace field is expanded too)

JasonWeill: I asked some other colleagues, and found consensus that "search box" is generally well understood, so let's go with it. We can always change it later, if we really need to. Thanks!

krassowski (approved):

PR #15830, opened 2024-02-20
Add token `ICellExecutor` providing an entry to customize cell execution

fcollonval:
## References

Follow-up of the discussion https://github.com/jupyterlab/team-compass/issues/229#issuecomment-1933852360
This will allow experimentation like https://github.com/jupyterlab/jupyterlab/pull/15448 to happen outside of core. It will also open the box for handling non-standard cell type execution for example (e.g. [SQL cell](https://github.com/QuantStack/jupyter-sql-cell/pull/9/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80)).

## Code changes

Provide the `Private.runCell` generic function as a plugin to allow customization.

## User-facing changes

None

None


fcollonval:
Failure are not related.

cc @echarles for awareness

almereyda:
As there is no documentation associated with this PR that shows how this faces to the user, I'd like to know if it brings native capabilities for what people are doing with the [SoS - Script of Script](https://vatlab.github.io/sos-docs/index.html) kernel nowadays?

packages/notebook/src/tokens.ts
```diff
@@ -130,3 +137,62 @@ export interface INotebookTracker extends IWidgetTracker<NotebookPanel> {
...
+  /**
+   * Notebook to which the cell belongs
+   */
+  notebook: Notebook;
+  /**
+   * A signal that emits whenever a cell completes execution.
+   */
+  executed: Signal<
+    any,
+    {
```
krassowski, line 155: 1. Should `executed` and `executionScheduled` signals have an exported interface?
2. Should we make the full signal public or narrow it down to `emit` method?

krassowski: To clarify I am thinking something like `SignalEmitter = Pick<Signal, 'emit'>` (possibly not the right syntax but I hope you see what I mean)

krassowski: Another take: it seems that `Notebook` is mostly passed so that these signal can be emitted (`handlePayload` only needs notebook model really). Could we proxy emitting the signals and add the `notebook` outside to reduce how much we pass down in options?

I guess there is no getting around passing `Cell` object though.

fcollonval: Thanks for the review @krassowski 

---

### Signal comments

The two signals `executed` and `executionScheduled` are [publicly available](https://github.com/jupyterlab/jupyterlab/blob/a0b9c86c93bbf66b070fb483ca2bc23beec671a6/packages/notebook/src/actions.tsx#L81) as `ISignal` from `NotebookActions`. So I don't see why we should drop `connect` and `disconnect`. Then we obviously need `emit`. So that left us only with the attribute `sender` that is exposed in `Signal` but not `ISignal` (that is a dummy placeholder in this case). So I would not bother creating a specific interface.

Another way to look at it, if those actions would have been implemented as Object method (never got why they are functions when most of the code is Object oriented), overriding the behavior would have meant inheriting from or duplicate `Notebook` with direct access to the signals (not the interface).

---

I'm fine passing the notebook model rather than the widget (it actually makes more sense). I think it is still important to pass the notebook document to get the context of execution. For example, one may require access to the notebook metadata (for the kernel information or custom info). Another scenario would be the reactive notebook case where cells depending on the current cell being executed should be queued for execution too.

packages/notebook-extension/src/cell-executor.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { JupyterFrontEndPlugin } from '@jupyterlab/application';
+import { ICellExecutor } from '@jupyterlab/notebook';
+import { runCell } from '@jupyterlab/notebook';
+
+/**
+ * Notebook cell executor plugin.
+ */
+export const cellExecutor: JupyterFrontEndPlugin<ICellExecutor> = {
+  id: '@jupyterlab/notebook-extension:cell-executor',
+  description: 'Provides the notebook cell executor.',
```
krassowski, line 13: What about console cells? These get executed as well and 90% of the code is shared with notebook. Since this is named `ICellExecutor`, should we make it handle either notebook _or_ console (and then if in future another execution context comes along we can easily add it).

fcollonval: I thought about it. But the two are quite different (presumably because they are separated). So should I try to reconciliate them or should the cell executor has two methods: `runNotebookCell` and `runConsoleCell`?

For reference, the console code is there:
https://github.com/jupyterlab/jupyterlab/blob/946f18cc1178155a3b979adf058918ab63fb9a60/packages/console/src/widget.ts#L724

krassowski: >  So should I try to reconciliate them or should the cell executor has two methods: `runNotebookCell` and `runConsoleCell`?

I am not sure which would be better; I brought it up on the lab call yesterday but there were no strong opinions. 

My thoughts:
- The disadvantage of reconciliating them into one method is that implementers might need to do a few conditional logic blocks depending on the execution context
- The disadvantage of two methods is that implementers would need to provide both, or we would need to have a default implementation as a fallback if we decided to make them optional (but then an interface with all methods optional risks that the implementer will fail to provide any if they make a typo and will be confused why nothing is working).

Thinking more about it, maybe there should be two separate plugins:
- `@jupyterlab/notebook-extension:cell-executor` providing `INotebookCellExecutor`
- `@jupyterlab/console-extension:cell-executor` providing `IConsoleCellExecutor`

krassowski: Separate plugins sounds better in the sense that additional execution context means just a new plugin, and that there is no need to have any fallbacks/optional methods, and no conditional logic.

The downside is that adding new execution context does require implementers to add a new plugin, but realistically this is something which does not happen frequently so I think this is a good trade-off.

fcollonval: Having two plugins seems a good path. Thanks for bringing this PR for discussion at the call.

krassowski: So I think the minimum required change for this PR would be renaming `ICellExecutor` to `INotebookCellExecutor`.

fcollonval:
Ok I pushed this one forward; :crossed_fingers: for the CI.

For the reviewer I have a final wording question, should it be `I...CellExector.runCell` (current version) or `I...CellExecutor.run`?

krassowski:
> should it be I...CellExector.runCell (current version) or I...CellExecutor.run?

I like how `runCell` how gives us freedom to expand this interface to include something like `runAllCells` in the future if needed (not saying that this is a good idea, just saying that having more specific names might be more future-proof).

krassowski:
Kicking CI after bot updates

packages/console/src/tokens.ts
```diff
@@ -19,3 +20,44 @@ export const IConsoleTracker = new Token<IConsoleTracker>(
  * A class that tracks console widgets.
  */
 export interface IConsoleTracker extends IWidgetTracker<ConsolePanel> {}
+
+export interface IExecutionOptions {
```
krassowski, line 24:
```suggestion
/**
* Execution options for console cell executor.
*/
export interface IExecutionOptions {
```

packages/notebook/src/tokens.ts
```diff
@@ -130,3 +138,62 @@ export interface INotebookTracker extends IWidgetTracker<NotebookPanel> {
    */
   readonly selectionChanged: ISignal<this, void>;
 }
+
+export interface IExecutionOptions {
```
krassowski, line 142: So if an extension wants to implement both `INotebookCellExecutor` and `IConsoleCellExecutor` this will be a slightly annoyance for explicit typing because one cannot import to `IExecutionOptions` objects without renaming them in the import. Some options include renaming it to `INotebookCellExecutionOptions`/`IConsoleCellExecutionOptions` or creating a namespace for each `INotebookCellExecutor` and `IConsoleCellExecutor` and just calling it `IOptions`. What do you think?

```suggestion
/**
* Execution options for notebook cell executor.
*/
export interface IExecutionOptions {
```

krassowski (commented):
Just one suggestion on renaming `IExecutionOptions` or nesting them in namespaces.

fcollonval:
Thanks @krassowski I went for the namespacing as it is a common pattern in core and will be a good fit in case new API are added to the executors.

I used `IRunCellOptions` instead of `IOptions` as the later is usually used for constructor. Moreover in case of a API extension the generic `IOptions` for `runCell` may bring confusion.

krassowski (approved):
Thank you @fcollonval.

I would personally prefix the namespaces with `I` (which is already common in the codebase) to take advantage of declaration merging, but I know it was called confusing by some new contributors so I am equally happy to go with the current version.

fcollonval:
> I would personally prefix the namespaces with `I` (which is already common in the codebase) to take advantage of declaration merging, but I know it was called confusing by some new contributors so I am equally happy to go with the current version.

done

krassowski:
The visual test failure is relevant. Maybe I should mock the descriptions, but for now updating it.

fcollonval:
Kicking the CI

galata/test/documentation/workspaces.test.ts-snapshots/workspaces-sidebar-documentation-linux.png
krassowski, line 1: Flaky snapshot update (a kernel from another test was captured). I feel like we need to mock the kernels out for some tests.

fcollonval: Thanks for the help

PR #15834, opened 2024-02-21
Uses the browser window's selection as the default search query

### LINKED ISSUE #14352: Initial query for search should be set based on output selection and rendered markdown selection too
## Description

Currently the initial search query in notebook is set based on selection in cell editor, but not based on selection of:
- rendered markdown text, or
- selection of outputs

This is not consistent as depending on where the selected text is, it the initial query in the search dialog will be pre-populated or not.

## Reproduce

1. Open notebook with a cell with output/markdown cell
2. Select text in output/rendered markdown cell
3. Press <kbd>Ctrl</kbd> + <kbd>F</kbd>
4. See search field was not pre-populated

Issue examplified on rendered Markdown:

markdown

Initial query being properly set for code cell selection but not for output:

initial query not set for output selection

## Expected behavior

Initial query is set for the active selection, regardless of the element type.

## Context

- JupyterLab version: 3.6.1 and 4.0.0b1

JasonWeill:
## References

Fixes #14352.

## Code changes

Changes the initial search query to whatever text is selected, if any, in the browser window.

## User-facing changes

When a user selects any text in the browser window, including in a cell input, in a cell output, in an error resulting from running a cell, in rendered Markdown, or in any other UI control such as the property inspector, then activates the search box with <kbd>CTRL</kbd>+<kbd>F</kbd> or <kbd>CMD</kbd>+<kbd>F</kbd> or the menu option, the default search text is the browser's selection. If no text is selected, the Find box is blank initially.

None.

### DIFF
--- a/packages/notebook/src/searchprovider.ts
+++ b/packages/notebook/src/searchprovider.ts
@@ -9,10 +9,7 @@ import {
   ICellModel,
   MarkdownCell
 } from '@jupyterlab/cells';
-import {
-  CodeMirrorEditor,
-  IHighlightAdjacentMatchOptions
-} from '@jupyterlab/codemirror';
+import { IHighlightAdjacentMatchOptions } from '@jupyterlab/codemirror';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import { IChangedArgs } from '@jupyterlab/coreutils';
 import {
@@ -289,16 +286,8 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
    * @returns Initial value used to populate the search box.
    */
   getInitialQuery(): string {
-    const activeCell = this.widget.content.activeCell;
-    const editor = activeCell?.editor as CodeMirrorEditor | undefined;
-    if (!editor) {
-      return '';
-    }
-    const selection = editor.state.sliceDoc(
-      editor.state.selection.main.from,
-      editor.state.selection.main.to
-    );
-    return selection;
+    // Get whatever is selected in the browser window.
+    return window.getSelection()?.toString() || '';
   }
 
   /**
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -670,7 +670,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
               description={
                 description + (name == 'selection' ? selectionKeyHint : '')
               }
-              isEnabled={isEnabled}
+              isEnabled={!showReplace || filter.supportReplace}
               onToggle={async () => {
                 await this.props.onFilterChanged(
                   name,
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -670,7 +670,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
               description={
                 description + (name == 'selection' ? selectionKeyHint : '')
               }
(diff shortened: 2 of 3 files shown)


packages/notebook/src/searchprovider.ts
```diff
@@ -286,16 +283,8 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
...
-    if (!editor) {
-      return '';
-    }
-    const selection = editor.state.sliceDoc(
-      editor.state.selection.main.from,
-      editor.state.selection.main.to
-    );
-    return selection;
+    // Get whatever is selected in the browser window.
+    return window.getSelection()?.toString() || '';
```
krassowski, line 287: This will require a lint pass and maybe some non-null checks but I tested it works in principle.

```suggestion
    const activeElement = document.activeElement;
    const currentWindow = activeElement instanceof HTMLIFrameElement ? activeElement.contentWindow : window;
    return currentWindow.getSelection()?.toString() || '';
```

krassowski (commented):
This is very nice solution, and covers 95% of use cases. The 5% which it does not cover is outputs wrapped with iframes (we also use iframes to render some untrusted content like some svg images which may contain text):

does-not-work-for-iframes

We could solve it for most iframes which have the same origin by checking if `document.activeElement` is an iframe and if it is then using its `contentWindow` (see the inline suggestion). What do you think @JasonWeill?

Finally, we should consider adding expanding the existing integration test with new cases. The current test is defined here:

https://github.com/jupyterlab/jupyterlab/blob/d582282a2fe9fd81804a3c47a3b4fe69f226fe0f/galata/test/jupyterlab/notebook-search.test.ts#L104-L152

packages/notebook/src/searchprovider.ts
krassowski: @JasonWeill just checking back on your thoughts on this in case if you missed the suggestion

krassowski: Just following up on this one, what do you think?

JasonWeill: @krassowski Thanks for your patience! I tried creating an iframe, highlighting the contents, and pressing CMD+F, but in that case, I ran the browser's "Find" function — the keyboard shortcut wasn't captured within the iframe. I can test by using the Edit > Find option in the menus.

image

JasonWeill: When I made your suggested change, in both Firefox ESR and Chrome, selecting `test` in the iframe, then pressing CMD+F, caused the browser's native find dialog to appear, not JupyterLab's. In each browser, when I chose Edit > Find from the pulldown menus, the selected `test` text was not prepopulated. Looking at the debugger, the active element at runtime is the HTML `body`, not the iframe. (This could be because selecting the command from the menus changed the active element.)

krassowski: That would be issue https://github.com/jupyterlab/jupyterlab/issues/15934

krassowski: I am saying this because I saw the same behaviour when testing https://github.com/jupyterlab/jupyterlab/pull/15857 locally yesterday.

However, I can imagine that the selector for Ctrl + F keybinding might not be reaching into the iframe if you still see this after rebasing.

JasonWeill: I'm still seeing the same behavior after rebasing against upstream `main`.

krassowski (approved):
Ok, let's track the iframe thing in a new issue.

PR #15838, opened 2024-02-21
Fix code sample in identity API example

krassowski:
## References

`app.services` does not exist, it is `app.serviceManager` [as seen in `collaboration-extension`](https://github.com/jupyterlab/jupyter-collaboration/blob/6bb0298165586bce6c1134cd0b29825503097243/packages/collaboration-extension/src/collaboration.ts#L69).

## Code changes

None

## User-facing changes

None

None


fcollonval (approved):
Thanks @krassowski

Issue #15758, opened 2024-02-07
Inconsistency in the toolbar


djangoliv:
## Description

Since jupyterlab 4.1.0, I noticed inconsistencies when adding buttons (with panel.toolbar.insertItem, panel.toolbar.insertAfter and panel.toolbar.inserBefore) in the toolbar.

## Reproduce

Create an extension adding a button with panel.toolbar.insertItem or panel.toolbar.insertAfter in the toolbar and toggle from full screen to half screen, and you will see differences in the order of the buttons.

A more simple example is to install https://github.com/logilab/jupyterlab-friendly-traceback/ which add a button with insert Item here: https://github.com/logilab/jupyterlab-friendly-traceback/blob/f62c03af0355437269559da07ffb8773eb3c7eda/src/index.ts#L68

Before (without jupyterlab-friendly-traceback) :
![before](https://github.com/jupyterlab/jupyterlab/assets/13416971/61b65700-6681-4e89-814e-0ad1d98fcfcb)

After (with jupyterlab-friendly-traceback):
![after](https://github.com/jupyterlab/jupyterlab/assets/13416971/08494037-9c93-46a8-bc29-469630292131)

## Expected behavior

Stability even after adding button with extension.

## Context

Jupyterlab 4.1.0 on firefox 122.0 (64 bits) on debian bookworm with i3.

Thanks and regards

JasonWeill:
This looks similar to #15509, which was reported as fixed in 4.1.x, but which seems to be recurring

mbektas:
issue probably caused by this [commit](https://github.com/jupyterlab/jupyterlab/commit/7084699339bde3c10567f80601538428e9c5da76#diff-00e4c3dd1d53c87918423484191102bdfda9a86263e410a858ec62b1669bfbe1R431) which defaults to position zero when reference item is not found. It used to return false if not found. Some extensions use this logic to try inserts. for example debugger button may not be available right away and insert is tried with other reference etc. cc. @brichet

brichet:
Thanks @djangoliv for opening this issue.
For reference the same issue was already mentioned here https://github.com/jupyterlab/jupyterlab/issues/15676

@mbektas thanks for the link, the issue probably comes from that commit.
I think the order of the items is not properly saved when an item is inserted in the toolbar.
It does not compute the position of the following items, which result in several item with the same position.

Issue #15676, opened 2024-01-22
Resizing notebook panel affects the order of icons in the toolbar


divyansshhh:
## Description
Resizing the notebook panel affects the order of icons in the toolbar. Here's an example - 

![resize](https://github.com/jupyterlab/jupyterlab/assets/30082866/f1894494-b364-42fe-b5be-01ccc3a4f940)

## Reproduce

Follow the steps in the above gif.

## Expected behavior

The kernel should remain right aligned.

## Context

JupyterLab version: 4.1.0b2

krassowski:
I can reproduce this when having buttons contributed by extensions (with native buttons only it seems less common occurrence):

![reprod](https://github.com/jupyterlab/jupyterlab/assets/5832902/6e5f45e1-a201-4b8c-8a53-9ad31d59f6c4)

CC @brichet

JasonWeill:
We saw this previously in #15509, which was closed, but this seems to be a regression now.

PR #15843, opened 2024-02-21
Fix saving of item positions in reactive toolbar

brichet:
This PR fixes the inconsistent item order in reactive toolbar.

Follow up https://github.com/jupyterlab/jupyterlab/pull/15553 and https://github.com/jupyterlab/jupyterlab/pull/15021

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15676
Fixes https://github.com/jupyterlab/jupyterlab/issues/15758

## Code changes

- Fix the position of items in the toolbar

  When an item is inserted in the reactive toolbar, its position in the toolbar is saved in a Map.
This allows items to be restores to their correct positions when moving between the `ToolbarPopup` and the `ReactiveToolbar` 
(when resizing). But the position of previously inserted items was not updated after an insertion, resulting in duplicate 
positions in the map (and inconsistent behavior).
 
  This PR compute the position of all items every time a new item is inserted in the toolbar.

- Add tests on notebook toolbar

  This PR also adds a function in the sidebar helper of galata, to change the width of a sidebar. this function has been moved from *documentation -> utils.ts` to the sidebar helper.

## User-facing changes

Fixes the inconsistent position of items.

None

## TODO

- [x] add tests on reactive toolbar

### DIFF
--- a/galata/src/helpers/sidebar.ts
+++ b/galata/src/helpers/sidebar.ts
@@ -270,12 +270,10 @@ export class SidebarHelper {
   /**
    * Set the sidebar width
    *
-   * @param page Page object
    * @param width Sidebar width in pixels
    * @param side Which sidebar to set: 'left' or 'right'
    */
   async setWidth(
-    page: Page,
     width = 251,
     side: galata.SidebarPosition = 'left'
   ): Promise<boolean> {
@@ -283,7 +281,7 @@ export class SidebarHelper {
       return false;
     }
 
-    const handles = page.locator(
+    const handles = this.page.locator(
       '#jp-main-split-panel > .lm-SplitPanel-handle:not(.lm-mod-hidden)'
     );
     const splitHandle =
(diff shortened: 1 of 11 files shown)


packages/ui-components/src/components/toolbar.tsx
```diff
@@ -376,7 +376,7 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     this.popupOpener.hide();
     this._resizer = new Throttler(async (callTwice = false) => {
       await this._onResize(callTwice);
-    }, 300);
+    }, 500);
```
krassowski, line 379: Is this increase required?

galata/src/helpers/sidebar.ts
```diff
@@ -267,6 +267,45 @@ export class SidebarHelper {
     });
   }
 
+  /**
+   * Set the sidebar width
+   *
+   * @param width Sidebar width in pixels
+   * @param side Which sidebar to set: 'left' or 'right'
+   */
+  async setWidth(
```
krassowski, line 276: Fine to add it in 4.2.0, but let's remove it on backport to 4.1.x to avoid expanding API surface in patch releases.

krassowski (approved):
Looks good to me. We can reconsider the testing approach in the future, a possibly cleaner alternative would  be creating a new example or using unit tests with `testing-library`.

Note: on backbort the `setWidth` method would ideally be reverted to a helper function (possibly duplicated in `/documentation` and `/jupyterlab`).

packages/ui-components/src/components/toolbar.tsx
brichet: Not really, but as it looks pretty smooth with a longer delay, I thought it best to avoid excessive calculations.
I have no objection to revert it.

krassowski: That's fine then, if we get any user feedback on it we can adjust it

krassowski:

PR #15844, opened 2024-02-21
Long items should not wrap

mdietz94:
If they wrap, it causes the text to overlap with the text on the next line, and can happen for instance if tab completing directories that contain a hyphen.

### DIFF
--- a/packages/completer/style/base.css
+++ b/packages/completer/style/base.css
@@ -64,7 +64,7 @@
   display: grid;
   grid-template-columns: min-content 1fr min-content;
   position: relative;
-  whitespace: nowrap;
+  white-space: nowrap;
 }
 
 .jp-Completer-item .jp-Completer-match {
--- a/packages/completer/style/base.css
+++ b/packages/completer/style/base.css
@@ -64,7 +64,6 @@
   display: grid;
   grid-template-columns: min-content 1fr min-content;
   position: relative;
-  white-space: nowrap;
 }
 
 .jp-Completer-item .jp-Completer-match {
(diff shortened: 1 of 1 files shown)


mdietz94:
Attached are pictures from before (with wrapping) and after my fix. I could not find any open issues that were related to this problem (though I could create one if you need).

mdietz94:
image

mdietz94:
The bottom picture is before and the top picture was after my fix. As you can see post fix you can actually tell what is being completed. Sorry for all the comments I was having trouble uploading the images.

Thanks!

packages/completer/style/base.css
```diff
@@ -64,6 +64,7 @@
   display: grid;
   grid-template-columns: min-content 1fr min-content;
   position: relative;
+  white-space: nowrap;
```
krassowski, line 67: Question to self: should it go to `.jp-Completer-item` or `.jp-Completer-match`?

For context, `jupyterlab-lsp` has more complex support for elision of completions and it sets it on `.jp-Completer-match` ([here](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/main/packages/completion-theme/style/index.css#L38)). I need to check if there was a reason for this in LSP, or if it is a meaningless difference.

krassowski: @mdietz94 would it work for you if this rule was moved to `jp-Completer-item .jp-Completer-match` below?

mdietz94: Yeah this still works, moved it in the PR

krassowski (approved):
Thank you @mdietz94!

mdietz94:
Thank you for the review!

krassowski:

Issue #15294, opened 2023-10-23
The use of space in kernel/tab hierarchy should be improved


krassowski:
### Problem

I often use the "Kernels" panel of the "Running Terminals and Kernels" sidebar to re-open notebooks/consoles for kernels which are running in the background after closing the tab. This used to be rather straightforward in JupyterLab 3.x but got quite less convenient in JupyterLab 4.0:

| JupyterLab <4 | JupyterLab 4.0 |
|---|---|
|![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/8cb91289-078d-4c80-a5c0-79787eb6d63f) | ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/e2feb88f-0982-48a8-82d9-bd4d5c1ec545) |

In both cases I closed the tab with `Unititled6.ipynb`; previously I could just click on it again in the "Kernels" section but now I need to:
- close the remaining two panels (and then)
- scroll back and forth to find it amid the repetitive "Python 3 (ipykernel)" grouping.

### Proposed Solution

Steps:
1. It could help if the kernels hierarchy auto-collapsed when a kernel has only one child. However, currently, when the hierarchy gets manually collapsed it is of little use
2. It could be handy to show both the kernel name and the notebook name in such cases
3. The repetitive part "Python 3 (ipykernel)" could be moved up to a new hierarchy level, so that each kernel group nests 1 or more kernels, and each kernel nests one or more tabs (main area widgets) attached to that kernel

| Step 1 | Step 2 | Step 3* |
|--|--|--|
| ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/72ae253b-394e-48bf-b0af-493a4b64a594) | ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/9b9b1ba5-069f-4402-8ec8-e05a8b3cfe03) | ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/47aa9f88-8076-41d4-8cd4-bda01e6924b8) |

\* Step 3 shows a slightly different but more representative situation where there are kernels of two different types, and where there are two notebooks attached to the same kernel.

### Additional context

- https://github.com/jupyterlab/jupyterlab/pull/13851
- https://github.com/jupyterlab/jupyterlab/issues/13779

mlucool:
I liked opening the old notebooks with a single click but now I still think I would still need to expand then click?

[...]

PR #15845, opened 2024-02-21
Re-organise the kernel tree in sidebar

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15294

## Code changes

- Adds `KernelSpecItem` which wraps `RunningKernel` items.
- `RunningKernel` no longer includes kernelspecs
- CSS is cleaned up
- [x] resize down the collapse/expand buttons to match visually with list/tree toggle
- [x] add a setting to remember preference list/tree
- [x] fix behaviour of buttons overflow when sidebar is resized down
- [x] update snapshots in tests

## User-facing changes

Early draft - to allow for more  interactive feedback

collapse-expand-list-view

None


krassowski:
Now that there are more buttons a visual glitch appeared:
- "Kernels" sometimes get ellipsed to "Kern..." when the sidebar is resized down
- more importantly the "Shut Down All" buttons gets hidden because it moves to new line in overflow

I experimented with hiding the expand/collapse buttons which is now possible using CSS only with container queries:

```css
.jp-RunningSessions.jp-SidePanel {
  /* Allow defining container queries when embedded in sidebar (which already uses strict size containment) */
  container-type: size;
  container-name: jp-running-sessions-side-panel;
}

@container jp-running-sessions-side-panel (max-width: 250px) {
  .jp-RunningSessions-collapseButton {
    /* hide if sidebar too narrow */
    display: none;
  }
}
```


But given that we only need to hide it on very small distance for the default usage with "Kernels" title I feel like this is an overkill and might be more confusing than helpful ("why is it hiding when I resize it down - there _seems_ to be enough space").

Instead I am shrinking the margins of these two collapse/expand buttons slightly which will bring them together (which makes sense to me as it will visually grouping them); and to prevent the "Shut Down All" button from dropping I am setting `min-width: max-content;` on the toolbar:

| Problem | Hiding collapse/expand buttons | Final Solution |
|--|--|--|
| the-problem | experiment | min-width |

krassowski:
This is ready for review. I am not convinced whether there is enough benefit to warrant the expand/collapse buttons but I will let @ericsnekbytes who suggested these let experiment with the functional version and provide feedback.

ericsnekbytes:
Conceptually, I think expand/collapse are good ideas for these kind of multi-notebook/console/shared-kernel workflows. Some things to note during my testing:

- Expand and collapse all are only valuable when the expand/collapse state are remembered per item in the tree. Right now, it seems expanding one item also expands all its children, which defeats the purpose of collapsing all to reduce noise and to allow users to individually drill down only into items that they care to expand/manage at that point.
- Remembering which sections are expanded/collapsed would be an essential part of navigating and managing open kernels/notebooks/sessions in this UI and with these multi-kernel/shared-kernel workflows...
- Right clicking the notebook tab/name and starting "new console for notebook" does not show the new console in the UI here (it seems to only work when right clicking the kernel entry and making a new console from this UI)

The UI and experience look good, and successfully communicate which kernels share state 👍  Expand/collapse all will not provide much value without individual and persistent expand/collapse state for individual tree entries.

Thanks for working on this!

fcollonval (changes requested):
Thanks @krassowski 

A couple of comments/questions:

- When opening a console linked to a notebook from the context menu on the notebook tab, the console is not listed. Is it expected?

image

- For expanding/collapsing all, in the toc with use a unique button with different icons than the one used here. I propose to be homogeneous by using one or the other (no preference)

- Similarly for switching between list and tree, in the debugger for switching between table and tree, the two buttons are always displayed and the selected mode has emphasis. Being homogeneous would be better (again no preference for one or the other approach)

- Should we display the kernel id differently - like using lighter color ?

krassowski:
Thanks @fcollonval

> When opening a console linked to a notebook from the context menu on the notebook tab, the console is not listed. Is it expected?

No, this is not expected, good catch. It looks this is pre-existing issue unrelated to this PR - I opened https://github.com/jupyterlab/jupyterlab/issues/15960 to track it. My guess is that consoles created using `notebook:create-console` are not getting added to the tracker.

> For expanding/collapsing all, in the toc with use a unique button with different icons than the one used here

I think the icons from the ToC might be better here indeed. I switched icons in https://github.com/jupyterlab/jupyterlab/pull/15845/commits/f4d259725555db876aad079c82f0ed299d1964bf and made it into a unique button in https://github.com/jupyterlab/jupyterlab/pull/15845/commits/8a65c32bf6788f90c1311d7518431de195e8e124 

> Similarly for switching between list and tree, in the debugger for switching between table and tree, the two buttons are always displayed and the selected mode has emphasis

I would prefer to have a single icon here. If it makes sense to propagate this to the debugger is not clear to me but if homogeneity is the guiding principle then I would go that way. Happy to hear counterpoints though.

> Should we display the kernel id differently - like using lighter color ?

Done in https://github.com/jupyterlab/jupyterlab/pull/15845/commits/f4d259725555db876aad079c82f0ed299d1964bf

packages/running-extension/src/kernels.tsx
```diff
@@ -224,20 +290,26 @@ namespace Private {
...
+              '%1 and %2 more',
+              children[0].label(),
+              children.length - 1
+            );
+      const kernelIdPrefix = kernel.id.split('-')[0];
+      return (
+        <>
+          {summary} <span className={KERNEL_LABEL_ID}>({kernelIdPrefix})</span>
+        </>
+      );
```
fcollonval, line 312: This breaks `labelTitle` (see line 318 ```const title = [`${this.label()}: ${id}`];```):

image

fcollonval (changes requested):
Thanks @krassowski 

I spotted a minor bug and I have a question about the table view; should we display the kernel id in a second column (a bit like last modified in the file browser)? Otherwise the kernels section does not bring lots of information compare to the open tabs.
> That can be a follow-up

packages/running-extension/src/kernels.tsx
krassowski: Thanks, fixed in 5f10b30.

krassowski:
> should we display the kernel id in a second column (a bit like last modified in the file browser)?

I do like this idea. I opened https://github.com/jupyterlab/jupyterlab/issues/15991 to track it.

fcollonval (approved):
Thanks @krassowski

PR #15849, opened 2024-02-21
Reconfigure the 4.0.x branch (4.1.x is new stable)

krassowski:
4.0.x should no longer be labelled as stable. Sibling to https://github.com/jupyterlab/jupyterlab/pull/15848


fcollonval (approved):
Thanks @krassowski

PR #15850, opened 2024-02-21
Add `4.0.x` and `4.1.x` to the documentation switcher

jtpio:
## References

Some versions were missing from the switcher on the documentation page: https://jupyterlab.readthedocs.io/en/latest/

image

## Code changes

Documentation config update.

## User-facing changes

Make more versions available via the switcher in the docs.

None

### DIFF
--- a/docs/source/_static/switcher.json
+++ b/docs/source/_static/switcher.json
@@ -25,43 +25,8 @@
     "url": "https://jupyterlab.readthedocs.io/en/3.6.x/"
   },
   {
-    "name": "3.5.x",
-    "version": "3.5.x",
-    "url": "https://jupyterlab.readthedocs.io/en/3.5.x/"
-  },
-  {
-    "name": "3.4.x",
-    "version": "3.4.x",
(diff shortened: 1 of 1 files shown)


jtpio:
> Should we keep all 3.x entries or just the latest one?

Good question. Or maybe just the currently supported versions, to reflect https://github.com/jupyterlab/jupyterlab/pull/15626 and https://github.com/jupyterlab/team-compass/issues/231?

Users can still modify the URL to get to a specific version if they want.

krassowski:
+1 to trimming 3.x to 3.6.x only. I think both 4.0.x and 4.1.x should stay for now.

I would still keep 2.x even if it is not supported, but do not see a value in keeping 1.x indeed.

jtpio:
Done in https://github.com/jupyterlab/jupyterlab/pull/15850/commits/32c61d6d11eda52a8fdfb93c801a4fb2f4d854f4

krassowski (approved):

PR #15857, opened 2024-02-22
Disable next/previous match buttons when no text is in Find box

### LINKED ISSUE #15813: Find and replace next/previous buttons are active even when no text is in the find textbox
Originally reported by @tgeorgeux in #6650, sub-issue 1.

The "next match" and "previous match" buttons in the search and replace UI control are always active, even when no text is in the Find textbox:

image

These buttons are active, and they can be pressed, when nothing is entered in the search text box. They also cause JavaScript errors like the following when pressed, although they do not scroll the page, as reported in #6650-1. (The error below was thrown in code on the JupyterLab tip of `main`, in Lab 4.1.)

```
Uncaught (in promise) TypeError: searchEngine is undefined
    _stepNext packages_notebook_lib_index_js.ca88e975802794059d81.js:6115
    highlightNext searchprovider.ts:327
    highlightNext searchmodel.ts:273
    onHighlightNext searchview.tsx:917
    onHighlightNext searchview.tsx:723
    onClick searchview.tsx:322
    React 23
    renderDOM vdom.ts:82
    renderDOM vdom.ts:79
    onUpdateRequest vdom.ts:51
    processMessage index.es6.js:1206
    invokeHandler index.es6.js:412
    sendMessage index.es6.js:166
    onAfterAttach vdom.ts:59
    processMessage index.es6.js:1240
    invokeHandler index.es6.js:412
    sendMessage index.es6.js:166
    attach index.es6.js:1663
    getSearchWidget index.ts:308
    execute index.ts:329
    execute index.es6.js:364
    _executeKeyBinding index.es6.js:526
    processKeydownEvent index.es6.js:462
    evtKeydown index.es6.js:2111
    handleEvent index.es6.js:2067
    addEventListeners index.es6.js:2098
    node_modules jlab_core.3a6c2825271f06388f85.js:4616
    promise callback*start index.es6.js:2021
    main index.out.js:701
    bootstrap bootstrap.js:96
    js bootstrap.js:97
    Webpack 5
3 packages_notebook_lib_index_js.ca88e975802794059d81.js:6115:1
```

JasonWeill:
## References

Fixes #15813.

## Code changes

Disables the next and previous match buttons when there is no text in the Find box.

## User-facing changes

When no text is in the Find box, the next and previous match buttons are disabled, do nothing when clicked, and appear with 60% opacity, the same level used for disabled commands in the command palette:

Screenshot 2024-02-22 at 9 43 02 AM

When there is text in the Find box, the next and previous match buttons are enabled and work the same way as before:

Screenshot 2024-02-22 at 9 43 16 AM

None.

### DIFF
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -39,9 +39,6 @@ const INPUT_BUTTON_CLASS_ON = 'jp-DocumentSearch-input-button-on';
 const INDEX_COUNTER_CLASS = 'jp-DocumentSearch-index-counter';
 const UP_DOWN_BUTTON_WRAPPER_CLASS = 'jp-DocumentSearch-up-down-wrapper';
 const UP_DOWN_BUTTON_CLASS = 'jp-DocumentSearch-up-down-button';
-const DISABLED_UP_DOWN_BUTTON_CLASS =
-  'jp-DocumentSearch-disabled-up-down-button';
-
 const FILTER_BUTTON_CLASS = 'jp-DocumentSearch-filter-button';
 const FILTER_BUTTON_ENABLED_CLASS = 'jp-DocumentSearch-filter-button-enabled';
 const REGEX_ERROR_CLASS = 'jp-DocumentSearch-regex-error';
@@ -56,8 +53,6 @@ const TOGGLE_WRAPPER = 'jp-DocumentSearch-toggle-wrapper';
 const TOGGLE_PLACEHOLDER = 'jp-DocumentSearch-toggle-placeholder';
 const BUTTON_CONTENT_CLASS = 'jp-DocumentSearch-button-content';
 const BUTTON_WRAPPER_CLASS = 'jp-DocumentSearch-button-wrapper';
-const DISABLED_BUTTON_WRAPPER_CLASS =
-  'jp-DocumentSearch-disabled-button-wrapper';
 const SPACER_CLASS = 'jp-DocumentSearch-spacer';
 
 interface ISearchInputProps {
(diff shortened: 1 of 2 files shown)


packages/documentsearch/src/searchview.tsx
```diff
@@ -304,30 +310,62 @@ function UpDownButtons(props: IUpDownProps) {
...
+    <button
+      className={DISABLED_BUTTON_WRAPPER_CLASS}
+      title={downButtonTitle}
+      disabled={true}
+    >
+      <caretDownEmptyThinIcon.react
+        className={classes(DISABLED_UP_DOWN_BUTTON_CLASS, BUTTON_CONTENT_CLASS)}
+        tag="span"
+      />
+    </button>
```
krassowski, line 362: This comes down to a stylistic choice but I would advocate for keeping it DRY and not adding an extra class name but using the `[disabled]` attribute selector with the existing class name - unless there is a strong reason for the implementation you suggest?

```suggestion
  const downButton = <button
      className={BUTTON_WRAPPER_CLASS}
      onClick={() => props.isEnabled ? props.onHighlightNext() : null}
      tabIndex={0}
      disabled={props.isEnabled}
      title={downButtonTitle}
    >
      <caretDownEmptyThinIcon.react
        className={classes(UP_DOWN_BUTTON_CLASS, BUTTON_CONTENT_CLASS)}
        tag="span"
      />
    </button>
```

JasonWeill: I had tried the `[disabled]` selector on `button` elements, but I found that in cases where the Find box was empty, the rule did not trigger as expected. There are other rules on the respective classes that caused UI elements to look active (e.g., color changes on hover) even when they were disabled and did nothing when clicked. This fixed the functionality problem, but not the appearance problem.

krassowski: It sounds like an issue with the selectors you tried. I can take a look if you would like to give it another go.

JasonWeill:
@krassowski With the most recent code pushed to this branch, the buttons now appear dimmed (less opacity) when the Find box is empty, but they still show a different background color on hover. I can pick this back up tomorrow.

JasonWeill:
The CSS rules are now updated to suppress hover and active effects on disabled buttons.

packages/documentsearch/style/base.css
```diff
@@ -34,11 +34,15 @@
   outline: 0;
 }
 
-.jp-DocumentSearch-overlay button:hover {
+.jp-DocumentSearch-overlay button.jp-DocumentSearch-button-wrapper:disabled {
+  opacity: 0.6;
```
krassowski, line 38: When I hover over the debugger button when disabled it shows not allowed cursor. When I hover over the disable search buttons they show hand cursor. Should we align to the debugger button style?

```suggestion
  opacity: 0.6;
  cursor: not-allowed;
```

cursor-style

packages/documentsearch/style/base.css
```diff
@@ -155,11 +159,11 @@
   margin: 1px 5px 2px;
 }
 
-.jp-DocumentSearch-up-down-button:hover {
+button:not(:disabled) .jp-DocumentSearch-up-down-button:hover {
```
krassowski, line 162:
```suggestion
button:not(:disabled) > .jp-DocumentSearch-up-down-button:hover {
```

packages/documentsearch/style/base.css
```diff
@@ -155,11 +159,11 @@
...
   background-color: var(--jp-layout-color2);
 }
 
-.jp-DocumentSearch-up-down-button:active {
+button:not(:disabled) .jp-DocumentSearch-up-down-button:active {
```
krassowski, line 166:
```suggestion
button:not(:disabled) > .jp-DocumentSearch-up-down-button:active {
```

packages/documentsearch/style/base.css
```diff
@@ -34,11 +34,15 @@
   outline: 0;
 }
 
-.jp-DocumentSearch-overlay button:hover {
+.jp-DocumentSearch-overlay button.jp-DocumentSearch-button-wrapper:disabled {
+  opacity: 0.6;
+}
+
+.jp-DocumentSearch-overlay button:not(:disabled):hover {
```
krassowski, line 41: Performance: having a generic element with three pseudo-elements on the right hand side of selector is not great. I don't feel like fiddling with it, but one way to make it work and eliminate performance impact would be making all the buttons within `.jp-DocumentSearch-overlay` use a common class name (in addition to the specific class names they already have). Another solution would be switching to `jp-button` element from `ui-compontents` which is only slightly better than `button` because if user has 1000s native `button`s in their notebook our styles would not cause any performance penalty (of course until they start using `jp-button`).

krassowski (commented):
Looks good in general, some tiny suggestions to consider.

packages/documentsearch/style/base.css
krassowski: Also debugger button uses 0.4 opacity (I do not have a preference here).

JasonWeill: I don't see the pointer cursor on the prev/next buttons in Firefox ESR, and the `cursor: not-allowed` rule doesn't seem to change the cursor to the 🚫 cursor as your animation shows.

packages/documentsearch/style/base.css
```diff
@@ -34,11 +34,16 @@
   outline: 0;
 }
 
-.jp-DocumentSearch-overlay button:hover {
+.jp-DocumentSearch-overlay button.jp-DocumentSearch-button-wrapper:disabled {
```
krassowski, line 37:
```suggestion
.jp-DocumentSearch-button-wrapper:disabled > .jp-DocumentSearch-button-content {
```

krassowski: > I don't see the pointer cursor on the prev/next buttons in Firefox ESR, and the `cursor: not-allowed` rule doesn't seem to change the cursor to the 🚫 cursor as your animation shows.

@JasonWeill what it we change the selector to the one proposed above? The hand cursor gets set on `.jp-DocumentSearch-button-content` so I think we should be overriding it on the same level.

krassowski (approved):
Tested locally works well:

works-ok

Thank you @JasonWeill!

PR #15862, opened 2024-02-23
Support suggestions

davidbrochart:
This adds suggestion support in RTC, similarly to how Google Docs works. A menu allows to choose the editing mode of a shared document:
- Editing: edit document directly. This is the current state of RTC for everyone collaborating on a document.
- Suggesting: edits become suggestions. This is the new mode added by this PR.

When entering suggesting mode, the frontend asks for a fork of the shared document. The forked document is a new document which is kept in sync with the root document, i.e. all updates to the root document are applied to the forked document. However the changes to the forked document are not applied to the root document yet. A user in suggesting mode makes changes to the forked document, not the root document. Only when these changes are accepted do they get merged in the root document. A new room is created in the backend for the forked document. The user in suggesting mode disconnects their shared document from the root room, and connects it to the forked room through its room ID.
Forked documents are advertised on the root document, in the `ystate` root Map type as a new entry `fork_{forkId}`. This way, a user editing the root document can be notified of suggestions. They can then connect their shared document to a forked room through its room ID.
Currently a user either sees the root document or a forked document, but not a diff of the forked document and the root document. Google Docs shows a unified view where all suggestions are shown with a different color. This would require the frontend to connect to the root document and all its forks at the same time, interpret the changes and show them on the same view. This is not currently possible as there can be only one document provider at a time.

This PR needs:
- https://github.com/jupyter-server/jupyter_ydoc/pull/218
- https://github.com/jupyterlab/jupyter-collaboration/pull/239

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -22,7 +22,7 @@
     "@codemirror/view": "^6.9.6",
     "@jupyter/react-components": "~0.13.3",
     "@jupyter/web-components": "~0.13.3",
-    "@jupyter/ydoc": "~1.1.1",
+    "@jupyter/ydoc": "file:.yalc/@jupyter/ydoc",
     "@jupyterlab/application": "~4.2.0-alpha.1",
     "@jupyterlab/application-extension": "~4.2.0-alpha.1",
     "@jupyterlab/apputils": "~4.3.0-alpha.1",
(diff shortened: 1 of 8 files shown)


davidbrochart:
Closing as there is currently nothing to do in this repository.
I opened https://github.com/jupyterlab/jupyterlab/issues/16008.

PR #15865, opened 2024-02-23
Check if editor exists before checking if it is disposed

### LINKED ISSUE #15864: Race condition in `_injectExtensions` when notebook model content changes immediately after being set
## Description

This assumes suggests that editor will always exist, https://github.com/jupyterlab/jupyterlab/blob/2d681271b159ac533df1ddb576ddcb256c44e59c/packages/lsp/src/adapters/editorAdapter.ts#L29-L31

so we can use non-null assertion in:

https://github.com/jupyterlab/jupyterlab/blob/2d681271b159ac533df1ddb576ddcb256c44e59c/packages/lsp/src/adapters/editorAdapter.ts#L54-L55

right? Nope, it does not work when someone has a test code like:

```js
notebook.model = new NotebookModel();
notebook.model.fromJSON(testNotebook);
```

But it can be fixed with changing the order

```js
const model = new NotebookModel();
model.fromJSON(testNotebook);
notebook.model = model;
```

The problem is that between the time the editor fired `ready()` the default cell has already been substituted.

## Context

- JupyterLab version: 4.1.2

krassowski:
## References

Fixes #15864

## Code changes

Check non-null rather than asserting because sometimes things move fast.

## User-facing changes

None

None

### DIFF
--- a/packages/lsp/src/adapters/editorAdapter.ts
+++ b/packages/lsp/src/adapters/editorAdapter.ts
@@ -51,8 +51,8 @@ export class EditorAdapter implements IDisposable {
    * Setup the editor.
    */
   private _injectExtensions(editor: Document.IEditor): void {
-    const codeEditor = editor.getEditor()!;
-    if (codeEditor.isDisposed) {
+    const codeEditor = editor.getEditor();
+    if (!codeEditor || codeEditor.isDisposed) {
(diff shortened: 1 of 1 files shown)


afshin (approved):
Thank you!

PR #15867, opened 2024-02-23
Adds description for disabled search filter

### LINKED ISSUE #13167: "Search Cell Outputs" Find/Replace is disabled when Replace is expanded, without an indication why
While using the Find/Replace feature I came across the following bug: the "Search Cell Outputs" can't be toggled while the Replace submenu is active. In principle, this is reasonable, as you can't replace the output of a cell. However, if the toggle was active before using the submenu (as is the default) the Find still searches and returns cell output. Thus you have to deactivate the Replace submenu in order to toggle the option off, and then resume trying to Find and Replace cell inputs.

I propose the toggle should be overridden and set to not search cell output while the Replace submenu is active, and return to the last toggled value after the submenu is deactivated.

I found this old issue which coincidentally includes a GIF depicting the described behaviour. Note how the "Search Cell Outputs" toggle turns grey once the Replace submenu is activated.

> Hi @LuckyMD this was implemented in #10067 and is included in JupyterLab 3.1 which is now in release candidate phase (please help testing it if you would like to try out this feature). Example:

test2

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/6650#issuecomment-879741501_

JasonWeill:
## References

Fixes #13167.

## Code changes

Adds a "disabled description" field to search filters, to provide an alternate description when the filter is disabled in "find and replace" mode.

## User-facing changes

When the user is in "find and replace" mode and opens the filter options in the search box, the "search cell outputs" filter is disabled. Mousing over it now displays an alternate title:

Screenshot 2024-02-23 at 2 16 16 PM

None.

### DIFF
--- a/packages/notebook/src/searchprovider.ts
+++ b/packages/notebook/src/searchprovider.ts
@@ -236,6 +236,7 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
       output: {
         title: trans.__('Search Cell Outputs'),
         description: trans.__('Search in the cell outputs.'),
+        disabledDescription: trans.__('Search in the cell outputs. (Not available in find and replace mode.)'),
         default: false,
         supportReplace: false
       },
--- a/packages/notebook/src/searchprovider.ts
+++ b/packages/notebook/src/searchprovider.ts
@@ -237,7 +237,7 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
         title: trans.__('Search Cell Outputs'),
         description: trans.__('Search in the cell outputs.'),
         disabledDescription: trans.__(
-          'Search in the cell outputs. (Not available in find and replace mode.)'
+          'Search in the cell outputs (not available when replace options are shown).'
         ),
         default: false,
         supportReplace: false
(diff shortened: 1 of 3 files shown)


packages/notebook/src/searchprovider.ts
```diff
@@ -236,6 +236,9 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
       output: {
         title: trans.__('Search Cell Outputs'),
         description: trans.__('Search in the cell outputs.'),
+        disabledDescription: trans.__(
+          'Search in the cell outputs. (Not available in find and replace mode.)'
```
krassowski, line 240: What do you think about:

```suggestion
          'Search in the cell outputs (not available in replace mode).'
```

I was also thinking about being more suggestive of action user needs to take, like "not available when the replace bar is expanded".

jtpio:
Restarting the UI tests, not sure if the failures are relevant or just flaky.

jtpio:
> Restarting the UI tests, not sure if the failures are relevant or just flaky.

Out of curiosity I restarted the UI tests 4 times, until they passed:

image

So it looks like there are still improvements to be done to the UI tests flakiness in general to reduce maintenance load. Maybe https://github.com/jupyterlab/jupyterlab/pull/15375 will help with that.

jtpio (approved):
Thanks!

PR #15875, opened 2024-02-26
Copy edits in dialog

JasonWeill:
## User-facing changes

Copy edits for a dialog box that appears if a user opts to search all outputs.

### DIFF
--- a/packages/notebook/src/searchprovider.ts
+++ b/packages/notebook/src/searchprovider.ts
@@ -521,11 +521,11 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
       const reply = await showDialog({
         title: trans.__('Confirmation'),
         body: trans.__(
-          'Searching outputs is expensive and requires to first rendered all outputs. Are you sure you want to search in the cell outputs?'
+          'Searching outputs requires you to run all cells and render their outputs. Are you sure you want to search in the cell outputs?'
         ),
         buttons: [
           Dialog.cancelButton({ label: trans.__('Cancel') }),
-          Dialog.okButton({ label: trans.__('Ok') })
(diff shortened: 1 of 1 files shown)


packages/notebook/src/searchprovider.ts
```diff
@@ -521,11 +521,11 @@ export class NotebookSearchProvider extends SearchProvider<NotebookPanel> {
       const reply = await showDialog({
         title: trans.__('Confirmation'),
         body: trans.__(
-          'Searching outputs is expensive and requires to first rendered all outputs. Are you sure you want to search in the cell outputs?'
+          'Searching outputs requires you to run all cells and render their outputs. Are you sure you want to search in the cell outputs?'
         ),
         buttons: [
           Dialog.cancelButton({ label: trans.__('Cancel') }),
-          Dialog.okButton({ label: trans.__('Ok') })
+          Dialog.okButton({ label: trans.__('OK') })
```
krassowski, line 528: All buttons in JupyterLab use "Ok" rather than "OK" so I would keep it as is for consistency (and specifying it here is not actually required as it gets pre-populated in [here](https://github.com/jupyterlab/jupyterlab/blob/d582282a2fe9fd81804a3c47a3b4fe69f226fe0f/packages/apputils/src/dialog.tsx#L770); if you wish to change to OK I would suggest doing it globally in a separate PR.

What about using a more specific label instead? (not sure about capitalization feel free to adjust)

```suggestion
          Dialog.okButton({ label: trans.__('Search Outputs') })
```

JasonWeill: If "Ok" is standard, I'll leave it at that. Thanks for the feedback!

krassowski (approved):
Thanks

Issue #15809, opened 2024-02-16
Sequencing of highlights is incorrect in notebook when replacing


krassowski:
## Description

Sequencing of highlights is incorrect in notebook when replacing but good when searching

## Reproduce

![wrong-order-cross-cells](https://github.com/jupyterlab/jupyterlab/assets/5832902/079c03b5-7c9f-4ba7-8796-0537407efc4e)

replacing numbers 12**6**345 whereas I would expect 123456, as is the case for search.

## Expected behavior

Should replace 123456 in order, the same order as search iteration uses:

![search-ok](https://github.com/jupyterlab/jupyterlab/assets/5832902/8766f29b-5331-4251-97fa-fe925293c6c7)

## Context

- JupyterLab version: >4.0.0,<=4.1.1

JasonWeill:
The file `searchprovider.ts` in the `cells` package forces a loop through the current cell if `currentProviderIndex === -1`, and `-1` is the default value for this attribute:

https://github.com/jupyterlab/jupyterlab/blob/2af428c23caae5e9fd0fcf7f87046845bcae283a/packages/cells/src/searchprovider.ts#L137-L145

It seems like when we're going through the first cell with matches, `currentProviderIndex` should not be `-1`, so that the `cells` `searchprovider` can iterate through all cells.

JasonWeill:
This is fixed by the current version of #15881.

PR #15881, opened 2024-02-27
Update highlight when replace string matches search expression

### LINKED ISSUE #15572: When searching for 'foo' and replacing with 'foo bar', replacing does not skip to the next occurrence of 'foo'
## Description

When I search for a string X, then replace it with a string Y that includes X as a substring, clicking "Replace" on the first occurrence causes the first occurrence to be selected again.

repeat-replace

## Reproduce

1. Enter `foo bar foo bar foo bar` in a cell.
2. Search for `foo` and choose to replace with `foo baz`.
3. Click "Replace" on the first occurrence of `foo`.
4. The first occurrence of `foo`, which is now followed by `baz`, remains selected.

## Expected behavior

The second occurrence of `foo` is selected.

## Context

- Operating System and version: macOS 13.6.3 "Ventura"
- Browser and version: Firefox 115.5.0esr (64-bit)
- JupyterLab version: 4.0.3

JasonWeill:
## References

Fixes #15572, #13968, and #15809.

## Code changes

Updates search providers so that, if the replacement string also matches the search query, the highlight and selection move to the next search result.

## User-facing changes

After replacing a string with a superset of the string, such as `foo` with `food`, the next search result highlights. The just-replaced string does not remain highlighted.

None.

### DIFF
--- a/galata/test/jupyterlab/notebook-replace.test.ts
+++ b/galata/test/jupyterlab/notebook-replace.test.ts
@@ -145,4 +145,41 @@ test.describe('Notebook Search and Replace', () => {
     await page.click('button:has-text("Replace")');
     await page.locator('text=-/-').waitFor();
   });
+
+  test('Replace with a string containing the query string', async ({
+    page
+  }) => {
+    // Create a small test notebook
+    await page.notebook.createNew();
+    await page.notebook.setCell(0, 'code', 'test\ntest');
+    await page.notebook.addCell('code', 'test\ntest');
+
+    await page.keyboard.press('Control+f');
+    await page.fill('[placeholder="Find"]', 'test');
+
+    await page.click('button[title="Show Replace"]');
+    await page.fill('[placeholder="Replace"]', 'tester');
+
+    // TODO: Next Match press count should be one less
+    // (the -/4 state should not be necessary).
+    await page.locator('text=-/4').waitFor();
+    await page.click('button[title^="Next Match"]', {
+      clickCount: 3
+    });
+
+    await page.locator('text=1/4').waitFor();
+    await page.click('button:has-text("Replace")');
+
+    await page.locator('text=2/4').waitFor();
+    await page.click('button:has-text("Replace")');
+
+    // At this point we should be in the second cell
+    await page.locator('text=3/4').waitFor();
+    await page.click('button:has-text("Replace")');
+
+    await page.locator('text=4/4').waitFor();
+
+    await page.click('button:has-text("Replace")');
+    await page.locator('text=1/4').waitFor();
+  });
 });
--- a/packages/cells/src/searchprovider.ts
+++ b/packages/cells/src/searchprovider.ts
@@ -131,11 +131,21 @@ class CodeCellSearchProvider extends CellSearchProvider {
     loop?: boolean,
     options?: IHighlightAdjacentMatchOptions
   ): Promise<ISearchMatch | undefined> {
-    if (this.matchesCount === 0 || !this.isActive) {
+    // Need to record whether all matches in the current cell have been processed,
+    // and to treat it here as a sign to move to the next cell.
+    if (
+      this.matchesCount === 0 ||
+      (options &&
+        options.from &&
+        options.from === 'previous-match' &&
+        this.currentIndex !== null &&
+        this.currentIndex + 1 >= this.cmHandler.matches.length) ||
+      !this.isActive
+    ) {
       this.currentIndex = null;
     } else {
       if (this.currentProviderIndex === -1) {
-        const match = await super.highlightNext(true, options);
+        const match = await super.highlightNext(loop, options);
         if (match) {
           this.currentIndex = this.cmHandler.currentIndex;
           return match;
--- a/packages/cells/src/searchprovider.ts
+++ b/packages/cells/src/searchprovider.ts
@@ -131,13 +131,12 @@ class CodeCellSearchProvider extends CellSearchProvider {
     loop?: boolean,
     options?: IHighlightAdjacentMatchOptions
   ): Promise<ISearchMatch | undefined> {
-    // Need to record whether all matches in the current cell have been processed,
-    // and to treat it here as a sign to move to the next cell.
+    // If we're scanning from the previous match, test whether we're
+    // at the end of the matches list.
+    const from = options?.from ?? '';
     if (
       this.matchesCount === 0 ||
-      (options &&
-        options.from &&
-        options.from === 'previous-match' &&
+      (from === 'previous-match' &&
         this.currentIndex !== null &&
         this.currentIndex + 1 >= this.cmHandler.matches.length) ||
       !this.isActive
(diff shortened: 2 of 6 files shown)


JasonWeill:
The way that we keep track of matches needs to be rewritten as a revision to this pull request. Especially because both find and replace support regular expressions, the number of matches may go up or down by any number as a result of a single "replace" operation. The "replace current" function needs to:

1. Replace the old text with the new text
2. Perform a search for the old text starting at a position of the previous match, plus the length of the new text
3. If there is a match between the start point and the end of a cell, highlight it, and set this.currentIndex. Otherwise, make this.currentIndex be null.

For reference, see CodeMirror’s search code’s “replaceNext'” function: https://github.com/codemirror/search/blob/0d8af3e4cc6e9e9bc2350952698015ff639ae857/src/search.ts#L465-L490

For the sake of the search dialog's result count, we might need to stop the search and start another search after the replacement is done. @krassowski please stop me if this change is getting far beyond the scope you'd expect for a fix of this nature. 😄

JasonWeill:
@krassowski This WIP change forcibly highlights the next match if the replacement text also matches the query. This is not ideal; replacing `bar` with `barbell` (one match in the replacement text) will move the highlight to the following `bar`, but replacing `bar` with `barbar` (two matches) will move the highlight to the second half of the replacement text. Not sure whether this is expected; what do you think?

Also, I ran into a problem where, in the first cell with matches, I'm looping through the matches after a single step with no matches at all. I think this is related to #15809, specifically a quirk with the logic of how `CodeCellSearchProvider` works. See https://github.com/jupyterlab/jupyterlab/issues/15809#issuecomment-1972274000.

krassowski:
> For the sake of the search dialog's result count, we might need to stop the search and start another search after the replacement is done.

I agree that search results count may change during the replace operation (or even during notebook loading); this does not mean that we should resort to restarting the search. I see these as two different things:
- stopping/restarting search is allowed to loose current index
- updating number or matches should not change anything other than the counter and non-active highlights

Some editors/browsers do not update search results on modification at all, until user presses the "next" button.

> replacing `bar` with `barbar` (two matches) will move the highlight to the second half of the replacement text. Not sure whether this is expected; what do you think?

Based on a quick check of the most popular editors, I believe users would be surprised by such behaviour and instead expect the highlight to move to the next original match; here are the editors I tried:

- VSCode: [bar] bar - (replace) → barbar [bar] - (replace) → [bar]bar barbar
- PyCharm: [bar] bar - (replace) → barbar [bar] - (replace) → barbar barbar (shows hint "press F3 to start from top")
- LibreOffice Writer: [bar] bar - (replace) → barbar [bar] - (replace) → [bar]bar barbar
- Google Docs: [bar] bar - (replace) → barbar [bar] - (replace) → [bar]bar barbar

I used `[]` to indicate active highlight.

krassowski:
> The way that we keep track of matches needs to be rewritten as a revision to this pull request

I am thinking if we could just cleverly adjust `currentIndex` (and `matches`) by the amount of matches inserted to the left.

For example, starting with:

```
# currentIndex = 1
bar [bar] bar
```

if we replace active `bar` with `foo` we get:

```
# currentIndex = 1
bar foo [bar]
```


but if we instead replace with `barbarbar` we get:

```
# currentIndex = 4 (this is 1 + 3)
bar barbarbar [bar]
```

It looks like some logic from `onSharedModelChanged`/`updateCodeMirror` could be resued to populate new `matches`, so the only part where a new logic is needed is the custom `currentIndex` handling. Does it make sense?

JasonWeill:
@krassowski Thanks for your patience! I'm ready to take another look at this change. 

I had looked into searching the replaced text for matches, but what if the replacement creates an additional match beyond the replacement text?

For example, start with `barr bar`, replace `bar` with `barbarba`.

```
# currentIndex = 0
[bar]r bar
```

Replace first instance, and we expect the status to be:

```
# currentIndex = 3
barbarbar [bar]
```

… but if we compare `barbarba` for matches of `bar`, the original search query, there are only 2 matches, not 3. The third match includes a trailing `r`.

JasonWeill:
Updated the PR to correctly update the highlight to, after replacing a target string with a replacement that contains the target string, move forward to the next highlight after the replacement, if possible. Please try it out.

krassowski (changes requested):
I tested it locally. The replacement behaviour looks good but there are two edge cases which are regressions:
1. When you press replace on the last match you see `-/x` rather than get the highlight back at the first match (`1/x`); you need to press replace again for it to cycle back to the front in the same cell
2. When you do press replace again instead of moving to the matches in the next cell, it moves you to the first match in the current cell

does-not-go-to-next-cell


The lint check failures are relevant:

```
/home/runner/work/jupyterlab/jupyterlab/packages/codemirror/src/searchprovider.ts
Error:   364:9   error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
Error:   379:13  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

And `js-notebook` test failure looks relevant too:

```
  ● @jupyterlab/notebook › NotebookSearchProvider › #replaceCurrentMatch() › should replace with a longer text and highlight next

    Expected: 0
    Received: null
```

krassowski:
For avoidance of doubt here is the expected behaviour for looping and transition between cells as captured in released versions/main:

across-cell
within-cell

JasonWeill:
From a little debugging, it looks like the new code in `codemirror`'s search provider is getting executed *after* the code in `notebook`'s search provider checks whether the current match is `null`. This results in the "-/N" search result status.

JasonWeill:
The most recent code uses a promise for the codemirror updates.

krassowski:
While there is no `-/N` anymore, is still do not see a progression between cell boundaries with the latest commit:

no-progression-across-cell-boundaries

Maybe this PR should include a playwright test for this?

krassowski:
In the scenario when search string is not a substring of the replacement string this works well now though:

progresion-works-for-non-overlapping

I am not sure what to make of this.

JasonWeill:
@krassowski Thanks! Can work on improving test coverage for this case.

krassowski:
Noting that this does not modify the public API and can be merged in RC and probably even backported to 4.1.x

JasonWeill:
The most recent code changes should fix the highlight order when replacing source text with target text that contains the source text.


https://github.com/jupyterlab/jupyterlab/assets/93281816/e6980235-9c35-483d-bfed-93d26a8d5cbc

JasonWeill:
@krassowski If you have time, could you please look at this revised PR? Thanks!

krassowski:
It still does not work in one tricky case; let's not block this PR on it - I opened https://github.com/jupyterlab/jupyterlab/issues/16242 to track it.

krassowski (approved):
Thank you @JasonWeill, let's merge and iterate!

jtpio:

PR #15888, opened 2024-02-29
Remove `child_process` dependency

jtpio:
## References

We should be able to remove that package.

image

https://www.npmjs.com/package/child_process

## Code changes

Clean up dependencies.

## User-facing changes

None

None


krassowski (approved):
Great catch, thank you!

PR #15889, opened 2024-02-29
Use toolkit search input for all search inputs: filebrowser, extension manager, debugger kernel source

Mehak261124:
**Problem Statement:**
Use toolkit search input for all search inputs: filebrowser, extension manager, debugger kernel source.

**Issue:**  
The UI in JupyterLab is displaying differently after integrating the Search Input component from the Jupyter Toolkit in various extensions like the file browser, extension manager, and debugger kernel source.

**Background:**  
I utilised React components from the Jupyter Toolkit, which are wrappers over web components made using the Fast Foundation's Open UI system. These components are based on a template slot mechanism, as shown in their Storybook documentation. I modified the LabIcon React component to accept a slot as a prop and set the slot attribute to the LabIcon's container.

**Specific Problem:** 

1. In the File Explorer and Search Extension, the UI appears inconsistent. The Search Extension's UI occupies the full height, unlike the File Explorer, which is contained within a ReactWidget class. A screenshot is provided for reference:

2. The shadow DOM inspection reveals __forwardedref="[object Object]", which might be causing the issue.
WhatsApp Image 2024-02-29 at 3 01 21 PM

**Question:**  
Should I wrap the Search Extension in a ReactWidget to resolve this CSS issue? What would be the best approach to fix this UI inconsistency? and I am trying to debug the forward Ref issue

**Screenshots:** 
1.Search extension before
Before1

2. File explorer search before
Before2

4. Search extension after
After1

5.File explorer search after
AFTER2


fcollonval:
Thanks @Mehak261124 for working on this.

> The UI in JupyterLab is displaying differently after integrating the Search Input component from the Jupyter Toolkit in various extensions like the file browser, extension manager, and debugger kernel source.

Some visual discrepancies are expected. But I agree with you the current discrepancy are too large. 

> Should I wrap the Search Extension in a ReactWidget to resolve this CSS issue? What would be the best approach to fix this UI inconsistency? and I am trying to debug the forward Ref issue

Don't worry about the `forwardRef`, it is definitely ugly but not related to trouble you are seeing here.

---

I use locally your branch to see how we can address the styling issue best. First of, a small tweak needs to be done in the search component itself - I opened a [fix](https://github.com/jupyterlab-contrib/jupyter-ui-toolkit/pull/90) and will make a new release when the CI passes.

Secondly we need to modify a bit more the `LabIcon` - see comment in code.

Finally we should preserve the CSS class on the component to apply custom styles.

In particular:

```css
/* Style the search component dimension in packages/ui-components/style */
.jp-FilterBox {
  --design-unit: 3;
  --density: 0;
}

.jp-extensionmanager-header .jp-FilterBox {
  /* This CSS rule needs to be added in  packages/extensionmanager/style/base.css */
  width: calc(100% - 20px);
}
```

The background color change in the settings editor is an open question for me as having a white background increase contrast (especially for the placeholder text). So it may be an improvement for accessibility.

packages/ui-components/src/components/search.tsx
```diff
@@ -185,18 +184,16 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
...
-      onChange={handleChange}
+    <Search
+      ref={props.inputRef}
       value={filter}
-    />
+      onChange={handleChange}
+      placeholder={props.placeholder}
+      disabled={props.disabled}
+    >
+        <searchIcon.react slot='end' />
```
fcollonval, line 195: We should not wrap the svg within another DOM element:

```suggestion
        <searchIcon.react slot='end' tag={null}/>
```

packages/ui-components/src/icon/labicon.tsx
```diff
@@ -458,6 +458,7 @@ export class LabIcon implements LabIcon.ILabIcon, VirtualElement.IRenderer {
           label,
           title,
           tag = 'div',
+          slot,
```
fcollonval, line 461: To apply the `slot` attribute in case `tag` is `null` (this use case), we need to change the logic below with the following patch

```patch
@@ -487,10 +487,21 @@ export class LabIcon implements LabIcon.ILabIcon, VirtualElement.IRenderer {
           // bail if failing silently
           return <></>;
         }
+        const svgProps = { ...this.svgReactAttrs };
+        if (!tag) {
+          Object.assign(svgProps, {
+            className:
+              className || styleProps
+                ? classes(className, LabIconStyle.styleClass(styleProps))
+                : undefined,
+            title: title,
+            slot: slot
+          });
+        }
 
         const svgComponent = (
           <svg
-            {...this.svgReactAttrs}
+            {...svgProps}
             dangerouslySetInnerHTML={{ __html: this.svgInnerHTML }}
             ref={ref}
           />
```

fcollonval (changes requested):

Mehak261124:
@fcollonval  Thank you so much for your detailed and insightful review of my PR. I really appreciate the helpful feedback and will work on incorporating your suggestions to improve it further.

Mehak261124:
@fcollonval Will there be  the need of "slot" attribute in the container element of svg ? As we are providing the slot attribute to the svg also in case the tag is null so we should only keep the slot attribute to the container if there is a use case of providing the svg with the container in the slot ( which I think there might be ) .

fcollonval:
> Will there be the need of "slot" attribute in the container element of svg ? As we are providing the slot attribute to the svg also in case the tag is null so we should only keep the slot attribute to the container if there is a use case of providing the svg with the container in the slot ( which I think there might be ) .

Yes you are correct

---

I released [v0.15.3](https://github.com/jupyterlab-contrib/jupyter-ui-toolkit/releases/tag/v0.15.3) of the toolkit with the search style fix. You should be able to upgrade the toolkit by executing `jlpm up "@jupyter/*components"`

fcollonval:
> The integrity test is failing on the CI. You should be able to fix it by running `jlpm integrity` and commit the changes.

Mehak261124:
Stiil some test cases are failing . I will try to correct the issue by tomorrow.

Mehak261124:
Hi @fcollonval  In above PR, I noticed some issues, during the debugging process, I noticed that the UI is still lacking. In an effort to ensure all test cases pass in the current PR and user experience standard is maintained, I will need to fix small css rules, that are in jupyter toolkit.

I have attached the video showcasing the issue, i noticed in the UI, and possible css rule that might be causing the bad UI in search component (https://drive.google.com/file/d/1ClGyOdUTX9rkyt4qqIfdu9QmfU7lt3tc/view?usp=sharing)

I've successfully set up the toolkit environment and made the necessary modifications. However, I'm keen on understanding the CSS design principles employed here. Are there any specific rules or conventions regarding margin and padding that I should adhere to?
Additionally, as part of the changes, I will need to update few test cases in the toolkit. I noticed that the PNGs attached to the previous PR (https://github.com/jupyterlab-contrib/jupyter-ui-toolkit/pull/90/files#diff-069c43cb14aeea39c087a0fa82d22021b16263385a23a1ceceb4b65f040bf76f) were generated. Could you please provide guidance on how these PNGs were created?

Mehak261124:
Screenshot 2024-03-08 at 5 30 45 PM

### Above is the issue with the padding inline end (I am highlighted it with the background color of yellow) in max width search component . 
<br>
<br>
Screenshot 2024-03-08 at 5 23 16 PM   

### Above image is without the padding inline end (I am highlighted it with the background color of yellow) in max width search component .

fcollonval:
Hey @Mehak261124 

Yes the padding was the source of the issue. I corrected it in the toolkit in https://github.com/jupyterlab-contrib/jupyter-ui-toolkit/pull/90 and released a new version 0.15.3. You can upgrade the toolkit in JupyterLab by executing `jlpm up "@jupyter/*-components" in the root folder.

---

Regarding your screenshot, the _space_ on the right matches the clear button space that will appear as soon as the user type something (in your video that space is bigger due to the wrong padding - but it is not the case any longer in the latest version).

---

> I've successfully set up the toolkit environment and made the necessary modifications. However, I'm keen on understanding the CSS design principles employed here. Are there any specific rules or conventions regarding margin and padding that I should adhere to?

In the toolkit, we are using the rules defined by the Fast team. There are already quite advanced because they support changing the density of an UI; see their [demo website](https://www.fast.design/)

image

---

> Could you please provide guidance on how these PNGs were created?

When opening a PR on the toolkit, one of the CI job is _Toolkit Visual Test_. It runs the integration tests. And if some failed (some are flaky so it is almost always the case unfortunately), it generates the newer reference snapshots.
The test results and the new references are available as artifact _
jupyter-ui-test _ of the job; see [that example](https://github.com/jupyterlab-contrib/jupyter-ui-toolkit/actions/runs/8112885739) - you will need to scroll to the bottom to see them.

The artifact is Zip file with the following structure:
image

The `test-assets` are the test results and the `src` are all the new references.

---

Thanks for reading all of this long answer :wink:

fcollonval (commented):
@Mehak261124 if you want I can push a commit to reduce the friction linked to the base packages bump.

Mehak261124:
@fcollonval thanks for offering, but had messed up a merge commit in between, so would have been a time consuming task to correct above PR.

I have made the required changes and reopened another PR, with updated head branch
https://github.com/jupyterlab/jupyterlab/pull/16046

sorry for the inconvenience caused

Issue #15821, opened 2024-02-19
Jupyterlab 4.1 dedent broken


harper357:
## Description

With JupyterLab <4.1, Shift+Tab would open object introspection, tab-completion if I was in a function or if I highlight part of the line it would just dedent it. However after upgrading to 4.1, it now will dedent the current line withought highlighting, so now it will open object introspection and dedent everytime I press Shift+Tab and there is no clear way to change this short cut.

I tried reverting to Jupyterlab 4.0.12 and this fixes it, so I think it might be a bug, but no one else has reported it so maybe it is just something with my setup?

I asked this on the Discourse, but have not received any reply.

## Reproduce

In a code cell, indent a line. Now hit `Shift+Tab` to bring up the object introspection. The line gets dedented and object introspection comes up. 

I created a new python3.11 virtualenv, installed Jupyterlab 4.1.1 and did the above steps.

## Expected behavior

Dedents should only happen if part of the line is hightlighted like in Jupyterlab < 4.1

## Context
macOS 12.3 (M1)
Chrome 121.0.6167.184 
JupyterLab version: 4.1.1

krassowski:
This is quite likely another side-effect of https://github.com/jupyterlab/jupyterlab/pull/14115; that pull request included mitigation for tab completion and indentation but not for dedentation.

Issue #15897, opened 2024-03-02
Allow extensions commands to suppress CodeMirror6 default shortcuts


krassowski:
### Problem

As summarised in https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1974835451, with changes from listening to keydown events in capture phase to listening to them in bubbling phase (JupyterLab 4.1) we lost ability to `preventDefault()` the default CodeMirror shortcuts behaviour. The current solution is override respective CM commands manually and check if it should possibly be prevented (e.g. based on the class of a closest DOM node), but it does not work if user disables such command.

### Proposed Solution

Either:
1. Restore listening to events in the capture phase for _all_ shortcuts, or
2. Make it a property of lumino's IKeybinding so some shortcuts can be invoked already in capture phase and others only in bubbling phase, or
3. Allow extensions to require a plugin that would allow them to mark CM keybindings as ignored under specific circumstances, or
4. Modify CM6 keybinding implementation such that it invokes lumino's keyboard processor before proceeding with execution of CM6 command; basically have a pre-hook for all key interactions

### Additional context

- https://github.com/jupyterlab/jupyterlab/issues/15821
- https://github.com/jupyterlab-contrib/jupyterlab-vim/issues/138

krassowski:
> but it does not work if user disables such a command or changes the shortcut.

Moreover, it makes extending jupyterlab difficult because one needs to hard-code these exceptions in core codemirror package to get the custom behaviour to work.

gabalafou:
Can you give a specific example? If I open JupyterLab 4.4 and install the Vim extension and then go into notebook code cell, I can use Vim shortcuts to insert, append, cut, paste - but I'm don't use the Vim extension regularly, so I do not understand where the points of conflict between Vim and CodeMirror are in JupyterLab >= 4.1.

daeyun:
I'd also appreciate an update on this. I'm stuck on an older version too. Certain default CodeMirror shortcuts, like Ctrl+K, can't be reassigned. Thanks for looking into this!

gabalafou:
I see. So the problem is essentially the [table of Vim (Jupyter Edit mode) bindings](https://github.com/jupyterlab-contrib/jupyterlab-vim?tab=readme-ov-file#vim-jupyter-edit-mode-bindings).

gabalafou:
[...]

PR #15898, opened 2024-03-02
Fix spurious dedent when opening inspector tooltip

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15821

Related to https://github.com/jupyterlab/jupyterlab/issues/15897

## Code changes

Checks if editor node is within a node matching tooltip opener before allowing dedent (`indentLess`) command to proceed.

## User-facing changes

| Before (tooltip + dedent) | After (only tooltip, as in 4.0 and 3.x) |
|--|--|
| before | after |

Without change: dedent works when any text fragment is selected:

dedent-with-selection-works

None

### DIFF
--- a/packages/codemirror/src/commands.ts
+++ b/packages/codemirror/src/commands.ts
@@ -4,6 +4,7 @@
  */
 
 import {
+  indentLess,
   indentMore,
   insertNewlineAndIndent,
   insertTab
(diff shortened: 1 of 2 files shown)


schmidi314:
Thanks @krassowski ! I tested your PR via binder and it seems to fix the problem for me.

krassowski:
CC @brichet in case if you have thoughts as this is following the approach that you used in https://github.com/jupyterlab/jupyterlab/pull/14115

brichet (approved):
Thanks @krassowski, LGTM

krassowski:

PR #15905, opened 2024-03-04
Print out incompatibility errors for "outdated" extensions on `--verbose`

krassowski:
## References

Relates to https://github.com/jupyterlab/jupyterlab/issues/14351

## Code changes

<s>Always</s> When `--verbose` flag is given, print out the conflicting dependencies, even if extension is judged to be outdated.

## User-facing changes

Before:

```
JupyterLab v4.1.2
/opt/hostedtoolcache/Python/3.11.8/x64/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        jupyterlab-conda-store v2024.1.1 enabled  X (python, jupyterlab-conda-store)


   The following extensions are outdated:
        jupyterlab-conda-store
        
   Consider checking if an update is available for these packages.
```

After (with verbose):

```
JupyterLab v4.1.2
/opt/hostedtoolcache/Python/3.11.8/x64/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        jupyterlab-conda-store v2024.1.1 enabled  X (python, jupyterlab-conda-store)


   The following extensions might be outdated or specify dependencies incorrectly:
        jupyterlab-conda-store
        
   Consider checking if an update is available for these packages.
   See below for the details.


"jupyterlab-conda-store@2024.1.1" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLabExtension        Package
>=18.2.0 <19.0.0>=17.0.1 <18.0.0 react
>=18.2.0 <19.0.0>=17.0.1 <18.0.0 react-dom
```

Note: initially this PR considered printing all of the errors by default; however, that had a disadvantage for users: `labextension list` would be more verbose when genuinely outdated extensions are installed; still we wanted the advantage for developers (`labextension list` shows relevant information when trying to upgrade an extension) so we ended up using `--verbose` flag.

None


krassowski:
An alternative might be to have a "verbose" flag and change:

```
Consider checking if an update is available for these packages.
```

to:

```
Consider checking if an update is available for these packages (use --verbose to see the full list of incompatible versions)."
```

or similar. Thoughts?

krassowski:
CC @RRosio @dlqqq for thoughts as you run into it yourself too

jtpio:
> * Disadvantage for users: `labextension list` is more verbose when genuinely outdated extensions are installed
> 
> * Advantage for developers `labextension list` shows relevant information when trying to upgrade an extension

Maybe we should make it nicer for users? Probably there are more users using the `jupyter labextension list` command than developers?

So the suggestion above sounds like good a good tradeoff:

> An alternative might be to have a "verbose" flag and change:

krassowski:
@RRosio do you feel comfortable taking a a look at the code and approving this PR as well, so that we can merge it?

jupyterlab/commands.py
```diff
@@ -2409,32 +2412,37 @@ def _format_compatibility_errors(name, version, errors):
...
             outdated.append(name)
-        else:
-            others.append(name)
 
     if outdated:
         logger.warning(
             "\n        ".join(
                 [
-                    "\n   The following extensions are outdated:",
+                    "\n   The following extensions might be outdated or specify dependencies incorrectly:",
```
dlqqq, line 2429:
```suggestion
                    "\n   The following extensions may have outdated dependencies that are incompatible with the current version of jupyterlab:",
```

jupyterlab/commands.py
```diff
@@ -2409,32 +2412,37 @@ def _format_compatibility_errors(name, version, errors):
...
                     *outdated,
-                    "\n   Consider checking if an update is available for these packages.\n",
+                    "\n   As a user, consider checking if an update is available for these packages.\n"
```
dlqqq, line 2431:
```suggestion
                    "\n   If you are a user, check if an update is available for these packages.\n"
```

dlqqq (changes requested):
@krassowski Thank you for working on this! I just got a chance to review.

I've left some minor suggestions for user-facing strings below. I'm not very opinionated on this, so please don't let them block this PR. 🤗

jupyterlab/commands.py
krassowski: I think that the OR is important here. The confusion before steamed from developers seeing a message about outdated extension which was in fact not outdated itself. Let's do something in between:

```suggestion
                    "\n   The following extensions may be outdated or specify dependencies that are incompatible with the current version of jupyterlab:",
```

krassowski:
>  I'm not very opinionated on this, so please don't let them block this PR. 

I tried to accommodate the suggestions but changed one of them slightly. Merging on the basis of the above sentence (and approval from Rosio), but if you do have any further suggestions, we can address these in a follow-up.

Issue #15912, opened 2024-03-05
`Ctrl + S` triggle browser's "save page", not save notebook in `4.1.3`


lf-shaw:
similar #306，tested with `Edge` and `firefox`.

it's ok in `4.1.2`.

Deeperand:
I use Jupyter Lab 4.1.3 on macOS 13.6.3 with Chrome 122.0.6261.94, also found this problem. Besides, other key-combination looks also meet the similar problem. e.g. Option-w will print ∑ instead of close current Jupyter tab.

mikewilkerson-solas:
This regression also affects the ability to use `CMD-SHIFT-c` (MacOS, Chrome/Brave/etc) to open the command menu (it _does_ open the menu, but a split second later the Chrome dev tools window pops up

krassowski:
Thank you all! I opened https://github.com/jupyterlab/jupyterlab/pull/15913 and added workarounds for `Ctrl + S` and `CMD-SHIFT-c` (I would appreciate if you could confirm that it works by testing out the PR [on Binder](https://mybinder.org/v2/gh/krassowski/jupyterlab/fix-ctrl-s?urlpath=lab) - please be patient with Binder when it builds, it might need reloading at some time). If you see any other problematic command please let me know.

> other key-combination looks also meet the similar problem. e.g. Option-w will print ∑ instead of close current Jupyter tab

This is an interesting case, so previously users complained than certain key combinations should type a very specific character but did something else. In JupyterLab 4.1.3 PR https://github.com/jupyterlab/jupyterlab/pull/15790 ensures that user input is never overruled, but this is turning out to be problematic too. I think we need to implement https://github.com/jupyterlab/lumino/issues/688 soon and allow users to tune it on per-command binding basis.

krassowski:
Before we solve it with fine-grained per-shortcut option, we could add a setting for toggling between "prevent default action by default" (JupyterLab 4.1.2 style) and "do not prevent default action by default" (JupyterLab 4.1.3 style) so that everyone would be happy.

krassowski:
Long-term solution PR in https://github.com/jupyterlab/jupyterlab/pull/15927

PR #15913, opened 2024-03-05
Fix saving shortcut not preventing default

krassowski:
## References

- Hot fix for #15912 
- Clean solution is proposed in https://github.com/jupyterlab/lumino/issues/688 but we should ship a patch ASAP while that is being discussed and developed

## Code changes

Immediately process keydown event for ctrl + s and on command + shift + c.

## User-facing changes

Saving with ctrl + s will no longer show spurious browser dialog

None


packages/application/src/lab.ts
```diff
@@ -206,10 +206,18 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
...
+    // TODO: generalise with https://github.com/jupyterlab/lumino/issues/688
     if (
+      // Navigation shortcuts which do not result in user input
+      // (except for Tab which is handled specially).
       ['Tab', 'ArrowDown', 'ArrowUp', 'ArrowRight', 'ArrowLeft'].includes(
         event.key
-      )
+      ) ||
+      // Saving shortcut which competes with the default browser action
+      (event.ctrlKey && event.key.toLowerCase() == 's') ||
```
azurelim, line 217: this not yet handle mac save file cmd+s that also got pop up browser save

krassowski: Thank you! This is now done in https://github.com/jupyterlab/jupyterlab/pull/15913/commits/ba68439adbf4cc9ad27660cad9fd6e3e5b346135.

packages/application/src/lab.ts
```diff
@@ -206,10 +206,19 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
...
       ['Tab', 'ArrowDown', 'ArrowUp', 'ArrowRight', 'ArrowLeft'].includes(
         event.key
-      )
+      ) ||
+      // Saving shortcut which competes with the default browser action;
+      // metaKey is `Command` on Mac.
+      ((event.ctrlKey || event.metaKey) && event.key.toLowerCase() == 's') ||
+      // Command + Shift + C on Mac should only open the Command Palette
+      // (rather than also opening the Dev Tools)
+      (event.metaKey && event.shiftKey && event.key.toLowerCase() == 'c')
```
jtpio, line 221: Does this condition also handle on non-mac devices?

jtpio: I guess this would also fix https://github.com/jupyterlab/jupyterlab/issues/15936

krassowski: Yes, though honestly I'm also considering just reverting this handler logic altogether until we get the the improved handling in lumino.

jtpio: Right, but I guess we'll still want this as a hot fix for now?

krassowski: No, I mean closing this one and merging revert PR https://github.com/jupyterlab/jupyterlab/pull/15938. This is mostly because I think we should not be breaking shortcuts contributed by extensions for which we do not have ability to provide a set of workaround like in this PR.

jtpio: Ah ok got it :+1:

Issue #15786, opened 2024-02-11
Theme color is not applied to jp-button in 4.1.0


AllanChain:
## Description

After upgrading from v4.0.x to v4.1.0 (maybe related to #15021), the button style was changed. The color of the button is hard-coded as `var(--neutral-foreground-rest)`, and it takes precedence over the rule `.jp-Toolbar { var(--jp-ui-font-color1) }`.

As a result, for example in solarized dark theme, the button is in bright white instead of gray (like the debugger icon):

![image](https://github.com/jupyterlab/jupyterlab/assets/36528777/f3c448b0-c6ef-4ce8-bfcf-39e77d76bc51)

## Reproduce

1. Install JupyterLab v4.1.0
2. Install a JupyterLab theme, for example
    ```shell
	pip install jupyterlab_theme_solarized_dark
	```
3. Open a notebook, `Python 3 (ipykernel)` is bright white

## Expected behavior

The color of `Python 3 (ipykernel)` is gray

## Context

- Operating System and version: Linux EndeavourOS (Arch-based distro) 
- Browser and version: Firefox 122 
- JupyterLab version: 4.1.0 

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/home/ac/temp/.venv/bin
	/home/ac/.asdf/shims
	/home/ac/.asdf/bin
	/opt/android-sdk/tools/bin
	/home/ac/.local/share/pnpm
	/usr/local/sbin
	/usr/local/bin
	/usr/bin
	/home/ac/.local/bin
	/opt/android-sdk/platform-tools
	/opt/android-sdk/tools
	/usr/lib/jvm/default/bin
	/usr/bin/site_perl
	/usr/bin/vendor_perl
	/usr/bin/core_perl

sys.path:
	/home/ac/temp/.venv/bin
	/usr/lib/python311.zip
	/usr/lib/python3.11
	/usr/lib/python3.11/lib-dynload
	/home/ac/temp/.venv/lib/python3.11/site-packages

sys.executable:
	/home/ac/temp/.venv/bin/python

sys.version:
	3.11.7 (main, Jan 29 2024, 16:03:57) [GCC 13.2.1 20230801]

platform.platform():
	Linux-6.7.4-arch1-1-x86_64-with-glibc2.39

which -a jupyter:
	/home/ac/temp/.venv/bin/jupyter
	/usr/bin/jupyter
[...]

PR #15915, opened 2024-03-05
Fix Theme color is not applied to jp-button

FoSuCloud:
## References

## Code changes

截屏2024-03-05 21 29 52

## User-facing changes

The color of `Python 3 (ipykernel)` use `var(--neutral-foreground-rest)`

The color of `Python 3 (ipykernel)`  use `var(--jp-ui-font-color1)`


FoSuCloud:
@meeseeksdev tag bug

PR #15916, opened 2024-03-05
Use docker on CI to speed it up

### LINKED ISSUE #13600: Improve CI jobs speed
### Problem

The CI jobs are taking lots of time and their number is easily triggering queuing when multiple contributors are active simultaneously.

### Proposed Solution

Ideas to speed the CI
- Filter jobs with the modified paths (see [GitHub syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore))
- Set up a single Jupyter server for Jest unit tests to remove the current limitation of running test sequentially (see [globalSetup](https://jestjs.io/docs/configuration#globalsetup-string) documentation)
- Build less, run more  
  The current job are all installing the dependencies and most of them then install JupyterLab. Caching the dependencies packages helps but only when the cache is used and valid - it often happen on a new PR that a new upstream Python dependency is downloaded due to a newer version being available but as the job fails the cache is not updated.  
  We could either move to docker stack (see https://github.com/docker/build-push-action/issues/225)
  But as this is not a technology we are really using, we could also simply share a large artifact built by a single initial job

fcollonval:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13600
Closes #11682

## Code changes

Building on top of the Docker file we are maintaining for dev (via devcontainer or directly), this switches the CI stack to first build a test docker image and feed it to ~~all~~ most of the tests.

It modifies slightly `jupyterlab.browser_check` to only install a single web browser based on `JLAB_BROWSER_TYPE` (that variable was already defined in `browser-test.js`).

> The speed up seen is substantial as creating the docker image take ~8min. But most job execution using it take 3 to 4 min less. It also allows for sharding the integration tests to speed up the overall execution.

Remaining:
- [ ] Access data outside the docker to store as artifacts (like for integration or release tests) --> needs to use volume
- [ ] Should we speed up the build by caching outside of the Docker image the yarn cache (and potentially the pip cache). This will fasten the build when the yarn.lock changes.
- [ ] Should we push a docker image per PR? The issue is that the currently test image is 2.5Go. So if we go that path, I think the only reasonable image would be a minimal python image with JupyterLab installed (assets non-minimized) to allow user testing but not all CI tests.

More optimization:
- [x] Cache Playwright chromium flavor.

## User-facing changes

None

None


fcollonval:
First test is encouraging, building the docker image took 6minutes then the 3 tests passing took 3min less (-20% to -35%) than the current job.

fcollonval:
Sharding of playwright tests is working well. But of course lots of tests are failing likely because the OS config has changed (we are running in the docker). A good side-effect of this PR will be the ability for dev to update the snapshots locally - meaning I'll keep a workflow for it; but we should advice knowledgeable dev to prefer updating just what is needed locally.

fcollonval:
Just to keep a trace; at the call it was requested to merge the integration tests reports (as this PR splits them). It is supported by Playwright: https://playwright.dev/docs/test-sharding#merging-reports-from-multiple-shards

krassowski:
Closing as sharding was implemented in https://github.com/jupyterlab/jupyterlab/pull/18427 with subsequent PRs further improving CI speed:
- https://github.com/jupyterlab/jupyterlab/pull/18454
- https://github.com/jupyterlab/jupyterlab/pull/18621
and work-in-progress for fixing all fonts in:
- https://github.com/jupyterlab/jupyterlab/pull/18535
- https://github.com/jupyterlab/jupyterlab/pull/18542
- https://github.com/jupyterlab/jupyterlab/pull/18541
- https://github.com/jupyterlab/jupyterlab/pull/18568

PR #15924, opened 2024-03-06
Fix Theme color is not applied to SwitchKernel ToolbarButton

FoSuCloud:
## References

## Code changes

截屏2024-03-06 11 40 56

## User-facing changes

The color of `Python 3 (ipykernel)` use `var(--neutral-foreground-rest)`

The color of `Python 3 (ipykernel)`  use `var(--jp-ui-font-color1)`


FoSuCloud:
@meeseeksdev tag bug

FoSuCloud:
It seems that ui test fail is not caused by this commit.

krassowski (approved):
Thank you @FoSuCloud!

krassowski:

Issue #15744, opened 2024-02-05
Unable to use multiple characters with international keyboards on MacOS in 4.1.0


miccoli:
## Description

Unable to use `[` `]` in editor with Italian keyboard on macOS.

## Reproduce

The Italian keyboard does not have `[` `]` keys, which are accessed via the option `⌥` key. Apparently the new keyboard shortcuts `⌥[` and `⌥]` mask the ability to properly input these characters.

1. Open a
    - notebook, or
    - console, or
    - text/markdown/python file          
3. Type `a = [1]`
4. The `[` `]` do not appear.

Please note that typing `[` at the start of a line results in the expected behaviour. Terminals work as expected.
This bug surfaced after the installation of v. 4.1.0: v. 4.0.12 works as expected.

## Context

- Operating System and version: macOS Sonoma 14.2.1 
- Browser and version: Chrome 121.0.6167.139 and Safari 17.2.1
- JupyterLab version: 4.1.0

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/Users/miccoli/.pyenv/versions/jupyter_312/bin
	/opt/homebrew/Cellar/pyenv/2.3.35/libexec
	/opt/homebrew/Cellar/pyenv/2.3.35/plugins/python-build/bin
	/opt/homebrew/Cellar/pyenv-virtualenv/1.2.1/shims
	/Users/miccoli/.pyenv/shims
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/usr/local/bin
	/System/Cryptexes/App/usr/bin
	/usr/bin
	/bin
	/usr/sbin
	/sbin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
	/Library/Apple/usr/bin
	/Library/TeX/texbin

sys.path:
	/Users/miccoli/.pyenv/versions/3.12.0/envs/jupyter_312/bin
	/Users/miccoli/.pyenv/versions/3.12.0/lib/python312.zip
	/Users/miccoli/.pyenv/versions/3.12.0/lib/python3.12
	/Users/miccoli/.pyenv/versions/3.12.0/lib/python3.12/lib-dynload
	/Users/miccoli/.pyenv/versions/jupyter_312/lib/python3.12/site-packages

[...]

Issue #4301, opened 2018-04-02
Feature request: Togle browser shortcut reactions 


bohdankornienko:
Having console kernel is awesome and let it use console editior such as vim, emacs, nano, etc. 
But these tools has shortcunts suchas Ctr+W which will close the broser tab instead. 

It would be great to have either option in settings or button in UI which allowed to toggle shortcuts catching by browser itself. 

It seems Google Docs has sort of feature like this.

Carreau:
I believe some of this shortcuts can't be disabled for security reason. Google has the ability to go around the protection in Chrome, but I don't think we can, at least no without writing a Browser extension that would need to be installed by users.

bohdankornienko:
I see. Looks like the user has to find its own workaround for now.

It would be great to have this feature anyways though. If it will be possible at some point.

bohdankornienko:
Could it be a hint on how one could implement this shortcut dealing mode:
https://askubuntu.com/questions/162700/how-can-one-disable-shortcuts-of-chrome/170909#170909
?

krassowski:
> I believe some of this shortcuts can't be disabled for security reason

I don't think this is true as of 2026. I think this is a reasonable request.

PR #15927, opened 2024-03-06
Improve the keybinding input guard using proposed lumino features

krassowski:
## References

- In https://github.com/jupyterlab/jupyterlab/issues/15744 we learned that users of specific layouts of international keyboards on Mac struggled to type certain characters due to new shortcuts introduced in JupyterLab 4.1.0.
- In https://github.com/jupyterlab/jupyterlab/pull/15790 (JupyterLab 4.1.3) we added a logic which ensures that shortcuts are not executed if user input would be produced by any particular keystroke; this worked...
- But in https://github.com/jupyterlab/jupyterlab/issues/15912 we learned that this had more side effects than intended; namely some default browser actions (like saving) were no longer prevented due to change in when the `processKeydownEvent` is called; this lead to two conclusions:
   - we need to have a way to prevent default action for some keybindings but not others
   - the number of actions which need preventing default is larger than expected and should not be hard-coded
- In https://github.com/jupyterlab/lumino/issues/688#issuecomment-1980585902 I noted that the current logic will fail for cords; to solve this we need to hold the execution internally and always call `processKeydownEvent()`
- https://github.com/jupyterlab/lumino/pull/689 implements a way to selectively disable `preventDefault` and to hold execution until a promise resolves (and to cancel execution if it resolves to false).

This PR requires a new lumino release and may need to wait for 4.2.0; a short-term solution (hotfix) for JupyterLab 4.1.4 is available in https://github.com/jupyterlab/jupyterlab/pull/15913.

The CI on this PR is expected to fail until https://github.com/jupyterlab/lumino/pull/689 is merged, released and we bump versions here.

Fixes https://github.com/jupyterlab/jupyterlab/issues/15744

Also relates to https://github.com/jupyterlab/jupyterlab/issues/4301

## Code changes

Use new `holdKeyBindingExecution()` to improve the logic from https://github.com/jupyterlab/jupyterlab/pull/15790 and mark the problematic shortcuts with `preventDefault: false`.

## User-facing changes

- when a keybinding marked with `"preventDefault": false` conflicts with default browser action such as typing a character, it will not be executed, and it the default browser action will be executed. 
- all other keybindings will prevent the default browser action as they used to in JupyterLab <=4.1.2

None


krassowski:
This still requires https://github.com/jupyterlab/lumino/pull/689 but I am marking this as ready for review as it can be tested by linking local installation of lumino with https://github.com/jupyterlab/lumino/pull/689 applied.

krassowski:
How to test it:

1. Without the patch add the following shortcut in JSON Setting Editor → Keyboard Shortcuts:
    ```json
    {
        "shortcuts": [
            {
                "command": "runmenu:run",
                "selector": "[data-jp-code-runner]",
                "keys": [
                    "Shift ["
                ]
            }
        ]
    }
    ```
2. In notebook press <kbd>Shift</kbd> + <kbd>[</kbd>
3. See that cell gets run (and `{` does not get typed)
4. Apply the patch by checking out the https://github.com/jupyterlab/lumino/pull/689 lumino branch
5. Build lumino with `yarn run build`
6. In JupyterLab repo with this branch run `jlpm link path/to/lumino/packages/commands`
7. Build JupyterLab with `jlpm run build`
8. Change the setting to (remember to press save icon):
    ```json
    {
        "shortcuts": [
            {
                "command": "runmenu:run",
                "selector": "[data-jp-code-runner]",
                "keys": [
                    "Shift ["
                ],
                "preventDefault": false
            }
        ]
    }
    ```
9. In notebook press <kbd>Shift</kbd> + <kbd>[</kbd> again
10. See that `{` gets typed
11. Change `preventDefault` to `true` (and save)
12. In notebook press <kbd>Shift</kbd> + <kbd>[</kbd> again
13. See that cell gets run.

krassowski:
demo

krassowski:
The failing `UI Tests / Visual Regression Tests` appear potentially relvant.

fcollonval (commented):
Thanks @krassowski  it looks good to me.

I guess we want to merge first #16078 to ensure the tests are passing for this one too.

krassowski:
The failures of:

```
    [jupyterlab] › test/jupyterlab/notebook-search.test.ts:276:7 › Notebook Search › Search in multiple selected cells 
```

are persistent across three re-runs - I will investigate.

krassowski:
No that test does not appear relevant:

| Actual | Expected |
|--|--|
| image | image |

There is 1 frame when the toolbar is not yet there, and one frame when it is half-populated.

However, I wonder if this introduces some more flakiness to the test runs because when we assumed that a command would be executed instantaneously after pressing keys, now it can take a few milliseconds more.

We can improve the logic by attaching `keyup` which should short-circuit; this may save a few milliseconds in the case if `onbeforeinput` would not be called.

fcollonval (approved):
Thanks @krassowski

PR #15935, opened 2024-03-07
Fixes filter for Chinese, other non-ASCII filenames

### LINKED ISSUE #14340: Filtering files by name doesn't work for Chinese file names.
hi,
When there are Chinese file names in my file browser list, I am unable to use the file name filtering function. And I noticed that it cannot be filtered if it starts with Chinese characters only, but it can be filtered and searched if it starts with letters, numbers, underscores, and then adds Chinese characters.

My context is as follows:

step1:
docker images jupyter/base-notebook
REPOSITORY              TAG       IMAGE ID       CREATED      SIZE
jupyter/base-notebook   latest    91d725edce1e   4 days ago   976MB

step2:
Start JupyterLab in a clean environment.
docker run -d -p 10000:8888 --name=jupyter-search-test -e RESTARTABLE=yes -v /etc/localtime:/etc/localtime -v /etc/timezone:/etc/timezone -e RESTARTABLE=yes -e JUPYTER_ENABLE_LAB=yes jupyter/base-notebook

I recorded the problem. Please take a look at the video, it's about 1 minute and 30 seconds long.

中文文件名：
测试1.ipynb
测试2第二个.ipynb
2测试.ipynb

When I typed "测试", "2测试.ipynb" can be filtered out, but "测试1.ipynb" and "测试2第二个.ipynb" cannot be filtered out.

thku！

https://user-images.githubusercontent.com/1202440/230634460-4f5037be-02ad-4fb0-9d32-a7b5396231b5.mp4

JasonWeill:
## References

Fixes #14340. 

## Code changes

Updates `fuzzySearch` function to not search based on word boundaries, which are ASCII-centric, and instead filters based on Unicode letters (including for Asian scripts), numbers, and diacritical marks, in any combination, as segments. Adds unit tests.

## User-facing changes

Filters based on Chinese characters now match at the beginning or middle of a filename, even if no ASCII characters occur before them. 

image

### DIFF
--- a/packages/ui-components/test/search.spec.tsx
+++ b/packages/ui-components/test/search.spec.tsx
@@ -9,38 +9,38 @@ describe('@jupyterlab/ui-components', () => {
     // tested in the lumino package.
     it('should find an ASCII match in text', () => {
       const result = fuzzySearch('food', 'foo');
-      expect(result).not.toBeNull;
+      expect(result).not.toBeNull();
     });
 
     it('should find an ASCII match in text with a space', () => {
       const result = fuzzySearch('fast food', 'foo');
-      expect(result).not.toBeNull;
+      expect(result).not.toBeNull();
     });
 
     it('should fail to find an ASCII match in text', () => {
       const result = fuzzySearch('fast food', 'bar');
-      expect(result).toBeNull;
+      expect(result).toBeNull();
     });
 
     it('should find a French match in text', () => {
       const result = fuzzySearch('Liberté, égalité, fraternité', 'erté');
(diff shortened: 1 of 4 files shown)


packages/ui-components/test/search.spec.tsx
krassowski, line 1: Unit tests :heart:

krassowski (changes requested):

krassowski (approved):
Thanks @JasonWeill!

krassowski:

PR #15946, opened 2024-03-07
Implement Workspaces GUI

krassowski:
## References

Closes https://github.com/jupyterlab/jupyterlab/issues/6944

## Code changes

- adds dedicated `workspaces` and `workspaces-extension` packages
- moves existing workspace logic for "save", "save as" and "open" (and commands) from `apputils-extension:workspaces` (which retains only the mime extension for opening `.jupyter-workspace` files on click)
- adds `workspaces-extension:sidebar` plugin which populates the `running` sidebar with the workspaces as per mockups in #6944
- adds `workspaces-extension:menu` plugin which populates a "Workspaces..." submenu in the main "File" menu; the existing "Save Workspace" and "Save Workspace As..." menu entries are moved to this new sub-menu.
- adds `workspaces-extension:commands` plugin which implements all the new commands and adds them to the command palette
- improves dialog and file dialog APIs to enable implementation of dialogs from mockups in #6944

<details>

<summary>Other (less relevant) code changes</summary>

- simpliies `saveAs` logic by using `InputDialog.getText()` (which gives us pre-selection and placeholder)
- removes duplicated licence banner in filebrowser/style/base.css
- allows to set `defaultButton` on input dialogs; this is just passed to dialog which supports the `defaultButton` already

</details>

## User-facing changes

workspaces

None


packages/workspaces-extension/src/commands.ts
```diff
@@ -0,0 +1,585 @@
...
+    state: IStateDB,
+    translator: ITranslator,
+    paths: JupyterFrontEnd.IPaths,
+    router: IRouter | null,
+    palette: ICommandPalette | null
+  ): IWorkspaceCommands => {
+    const trans = translator.load('jupyterlab');
+
+    const namingHintLabel = trans.__(
+      'Naming the workspace will create a unique URL. Please name your workspace without special characters.'
```
krassowski, line 105: `jupyterlab-server` explicitly allows spaces and does not list any restrictions on manager level (it does use [`slugify`](https://github.com/jupyterlab/jupyterlab_server/blob/91a081672f05d6611b7d7a9f832695e863403c1d/jupyterlab_server/workspaces_handler.py#L66-L95) to remove invalid characters when creating file name, but it has no bearing on workspace resolution from URL).

There are restrictions on the routing/handler level though:

```
MASTER_URL_PATTERN = (
    r"/(?P<mode>{}|doc)(?P<workspace>/workspaces/[a-zA-Z0-9\-\_]+)?(?P<tree>/tree/.*)?"
)
```

[`MASTER_URL_PATTERN`](https://github.com/jupyterlab/jupyterlab_server/blob/91a081672f05d6611b7d7a9f832695e863403c1d/jupyterlab_server/handlers.py#L34) only allows `[a-zA-Z0-9\-\_]+`.

Interestingly there is no such a limitation on the Workspaces Manager API:

```
workspace_api_path = ujoin(extension_app.workspaces_api_url, "(?P<space_name>.+)")
        handlers.append((workspace_api_path, WorkspacesHandler, workspaces_config))
```

which takes basically any combination of characters (`.+`) - defined [here](
https://github.com/jupyterlab/jupyterlab_server/blob/91a081672f05d6611b7d7a9f832695e863403c1d/jupyterlab_server/handlers.py#L266).

To be on the safe side I will update the instructions and add validation for `[a-zA-Z0-9\-\_]+`

JasonWeill:
@krassowski Thanks for your work so far on this! Please let me know when you'd like some feedback on it.

krassowski:
Note to reviewers: failures of `npm dependencies failed to install` are expected because this PR adds a new package which is not on npm yet. These will get resolved once we merge and follow with a new pre-release.

JasonWeill:
@krassowski Yes! I was unavailable for part of last week, but can take a look at this as soon as I can.

JasonWeill:
@krassowski Thank you so much for circulating this PR! Some early notes from using it, before digging into the code:

The menu should use ellipses consistently. [Apple's Human Interface Guidelines for menus](https://developer.apple.com/design/human-interface-guidelines/menus) suggest, "Append an ellipsis to a menu item’s label when people need to provide additional information before the action can complete." A cascading menu, like File > Workspaces, should not have an ellipsis (as it does now), but all of its submenu items require additional information in a modal dialog, so they should all have ellipses.

In the running pane’s “Workspaces” section“, when I hover over the trash icon to the right of each workspace, the tooltip only says ”Delete“. For accessibility and usability, it should say ”Delete %1“, with the workspace name substituted in. (”Delete workspace ‘%1’“ could also work.)

The confirmation for deleting a workspace doesn’t put the name of the workspace in quotes or otherwise emphasize it, so users may miss it. This will also avoid confusion in case the workspace’s name is a word.

The first sentence of the confirmation dialog for deleting a workspace reads, “Deleting workspace %1 will also delete its URL.” I’m not sure what this means. The URL itself is not deleted, but visiting it will no longer take the user to workspace %1.  Consider alternate wording, such as, “If you delete workspace ‘%1’, URLs that refer to this workspace will go to a different workspace.”

After I click a link to go to the auto-w workspace, when I open the running widget in the left pane, then I expand workspaces, language servers, kernels, and open tabs, in that order, I see “language servers” is very tall and empty, and both “kernels” and “workspaces” have content in them but are very small:

running-expanded

The “Delete All” dialog needs work. Its title is in Title Case and ends in a question mark (Delete All?) whereas the dialog title for deleting a single workspace is in sentence case and has no punctuation (Delete workspace). Other dialogs, like those for shutting down all kernels or opening from an arbitrary path, use Title Case, and some end in a question mark. In addition, the “Delete All” dialog does not convey that deleting all workspaces is irreversible (cannot be undone), whereas the "Delete workspace" dialog does. I think that it’s important to convey that something cannot be undone as prominently as possible.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
...
+There are two top level keys requires, `data`, and `metadata`.
+
+The `metadata` must be a mapping with an `id`
+key that has the same value as the ID of the workspace. This should also be the relative URL path to access the workspace,
+like `/lab/workspaces/foo`. Additionally, `metadata` may contain `created` and `last_modified` fields with date and time creation and most recent modification, respectively. The date and time are encoded using ISO 8601 format.
+
+The `data` key maps to the initial state of the ``IStateDB``. Many plugins look in the State DB for the configuration.
+Also any plugins that register with the ``ILayoutRestorer`` will look up all keys in the State DB
+that start with the `namespace` of their tracker before the first ``:``. The values of these keys should have a `data`
+attribute that maps.
```
JasonWeill, line 84: This is not clear. We should use "state database" as the shorthand for `IStateDB` (even though it is longer in characters, it is easier to read). Although this refers to "many plugins", it's not clear what a user would find important about this. The final sentence also trails off; it is not clear what is mapped to what.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
...
+    command: 'docmanager:open',
+    args: widget => ({
+      path: widget.context.path,
+      factory: Private.factoryNameProperty.get(widget)
+    }),
+    name: widget =>
+      `${widget.context.path}:${Private.factoryNameProperty.get(widget)}`
+  });
+
+Note the part of the data key after the first ``:`` (``package.json:JSON``) is dropped and is irrelevant.
```
JasonWeill, line 115: This needs to be clarified or rewritten. The term "irrelevant" is very concerning in user documentation; clearly, the part after the first `:` is relevant to _something_ — why else would it be there?

galata/test/jupyterlab/menus.test.ts
```diff
@@ -7,6 +7,7 @@ import type { ISettingRegistry } from '@jupyterlab/settingregistry';
 const menuPaths = [
   'File',
   'File>New',
+  'File>Workspaces…',
```
JasonWeill, line 10: This should be `File>Workspaces` once corrected (see earlier)

packages/apputils/src/inputdialog.ts
```diff
@@ -347,22 +397,49 @@ class InputNumberDialog extends InputDialogBase<number> {
 }
 
 /**
- * Widget body for input text dialog
+ * Base widget body for input text/password/email dialog
  */
-class InputTextDialog extends InputDialogBase<string> {
+class InputDialogTextualBase extends InputDialogBase<string> {
```
JasonWeill, line 402: The phrase `TextualBase` seems unusual here. Could we refer to this as `InputDialogTextBase` or `InputTextDialogBase`? The latter is more similar to the class name used before, and `Text` is used much more than `Textual` in front-end code, both ours, Lumino's, and other packages'.

packages/filebrowser/style/base.css
```diff
@@ -104,6 +99,14 @@
   display: none;
 }
 
+.jp-Open-Dialog > .jp-FileBrowser {
+  min-height: 200px;
```
JasonWeill, line 103: Do we have a suitable variable for minimum dialog height?

JasonWeill (commented):
I didn't add annotations to specific code lines regarding the UI questions I left in a separate comments. Thanks for your contribution! Happy to review again or discuss further.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
+.. Copyright (c) Jupyter Development Team.
+.. Distributed under the terms of the Modified BSD License.
+
+.. _workspaces:
+
+Workspaces
+==========
+
+A JupyterLab Workspace defines the layout and state of the user interface such as the position of files, notebooks, sidebars, and open/closed state of the panels.
```
krassowski, line 9: Comment from @JasonWeill which I mistakenly deleted while preparing my reply to it :facepalm: 

> We should indicate that a workspace name may only contain ASCII letters a-z, A-Z, and digits 0-9 (regex "word" characters) and hyphens. [#14340](https://github.com/jupyterlab/jupyterlab/issues/14340) was a bug related to Chinese text handling because we incorrectly relied on "word" characters being used, exclusively, globally, in filenames.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
...
+  Saved workspace: <workspaces-directory>/labworkspacesfoo-54d5.jupyterlab-workspace
+
+The ``export`` functionality is as friendly as possible: if a workspace does not
+exist, it will still generate an empty workspace for export.
+
+The ``import`` functionality validates the structure of the workspace file and
+validates the ``id`` field in the workspace ``metadata`` to make sure its URL is
+compatible with either the ``workspaces_url`` configuration or the ``page_url``
+configuration to verify that it is a correctly named workspace or it is the
+default workspace.
```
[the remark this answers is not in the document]
krassowski: Since this is pre-existing docs, I opened #16020 to track this.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
...
+---------------------
+
+A workspace file in a JSON file with a specific spec.
+
+
+There are two top level keys requires, `data`, and `metadata`.
+
+The `metadata` must be a mapping with an `id`
+key that has the same value as the ID of the workspace. This should also be the relative URL path to access the workspace,
+like `/lab/workspaces/foo`. Additionally, `metadata` may contain `created` and `last_modified` fields with date and time creation and most recent modification, respectively. The date and time are encoded using ISO 8601 format.
```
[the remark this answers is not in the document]
krassowski: Done in 8134d1f5f1e8759fb6d60104376e7e45b591d1df.

galata/test/jupyterlab/menus.test.ts
krassowski: Done in 2872d004058e0751e026ced4651d14bbe07fb5b9.

docs/source/user/workspaces.rst
krassowski: Done in 6459bc05f21581966736ee13d754b50dc9904bf6.

docs/source/user/workspaces.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Managing Workspaces (GUI)
+-------------------------
+
+A number of commands is available to manage workspaces from the main menu, sidebar, and command palette:
+
+- `create-new`, `clone`, `rename`, `reset`, and `delete` act on the workspaces stored by on the server in :ref:`the dedicated location <workspaces-directory>`.
+- `save`, `save as`, `import`, and `export` can load and store the workspace to/from the file system (contained within the Jupyter root directory); `save` will save the workspace to the most recently saved file.
+
+In the sidebar the current workspace is indicated with check mark (✓). A different workspace can be opened by clicking on the corresponding sidebar item. Opening context menu (right click) over the workspace item in the sidebar will present actions available for management of that workspace:
```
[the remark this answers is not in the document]
krassowski: Clicking on another workspace does not open it in a new browser tab or window. It opens it in the same tab. Which browser did you see the behaviour of opening in a new tab or window? For now I will leave that fragment out:

```suggestion
In the sidebar, in the "Running Terminals and Kernels" panel, under "Workspaces", the current workspace has a check mark (✓). Clicking on another workspace will open. Opening the context menu (right click) over the workspace item in the sidebar will present actions available for management of that workspace:
```

krassowski:
Thank you for the review @JasonWeill!

> The menu should use ellipses consistently. [Apple's Human Interface Guidelines for menus](https://developer.apple.com/design/human-interface-guidelines/menus) suggest, "Append an ellipsis to a menu item’s label when people need to provide additional information before the action can complete." A cascading menu, like File > Workspaces, should not have an ellipsis (as it does now), but all of its submenu items require additional information in a modal dialog, so they should all have ellipses.

Done in 2872d004058e0751e026ced4651d14bbe07fb5b9. What I was trying to do was being consistent with "Save and Export Notebook As..." just above which does have an ellipsis. Do you think that should be changed too (if so, let's open an issue to track it)?

> In the running pane’s “Workspaces” section“, when I hover over the trash icon to the right of each workspace, the tooltip only says ”Delete“. For accessibility and usability, it should say ”Delete %1“, with the workspace name substituted in. (”Delete workspace ‘%1’“ could also work.)

Done in 19f3ad82e01a5d4a2a43932129abd63f31d5eaeb.

> The confirmation for deleting a workspace doesn’t put the name of the workspace in quotes or otherwise emphasize it, so users may miss it. This will also avoid confusion in case the workspace’s name is a word.

Done in f15ed801bf1502442ef39dabda9efb8e572fccfd.

> The first sentence of the confirmation dialog for deleting a workspace reads, “Deleting workspace %1 will also delete its URL.” I’m not sure what this means. The URL itself is not deleted, but visiting it will no longer take the user to workspace %1. Consider alternate wording, such as, “If you delete workspace ‘%1’, URLs that refer to this workspace will go to a different workspace.”

This wording was taken from mockups in #6944. While I agree that "deleting a URL" sounds odd, I do not think that "URLs that refer to this workspace will go to a different workspace" is accurate (it will not - it will either show an error dialog that workspace does not exist and load the default workspace, or create a new one; I have no idea why it sometimes does one or the other). I suggest to discuss this in #6944 and we can adjust it after this PR is merged (as it staying open has a high cost due to merge conflicts).

> After I click a link to go to the auto-w workspace, when I open the running widget in the left pane, then I expand workspaces, language servers, kernels, and open tabs, in that order, I see “language servers” is very tall and empty, and both “kernels” and “workspaces” have content in them but are very small:

You can resize them by dragging. Changes to the logic for managing sizes of the panels is out of scope for this PR, but if you want to suggest a better behaviour I would advise to open a new issue.

> The “Delete All” dialog needs work. Its title is in Title Case and ends in a question mark (Delete All?) whereas the dialog title for deleting a single workspace is in sentence case and has no punctuation (Delete workspace).

This comes from the `running` package code:

https://github.com/jupyterlab/jupyterlab/blob/a1166c15297bbae0948609879055af00940947dc/packages/running/src/index.tsx#L356-L378

> Other dialogs, like those for shutting down all kernels or opening from an arbitrary path, use Title Case, and some end in a question mark.

Yes, because they all generated by `running` package code. I would suggest that changes to that are handled in separate follow up PR as these might lead to broader API changes, best discussed separately.

> In addition, the “Delete All” dialog does not convey that deleting all workspaces is irreversible (cannot be undone), whereas the "Delete workspace" dialog does. I think that it’s important to convey that something cannot be undone as prominently as possible.

Done in c257f6d09191bc292271b13e0671569d1b742ada.

JasonWeill (approved):
Thanks for opening related issues! I think we're good to proceed now.

krassowski:
bot please update snapshots

krassowski:
Note to self on docs test needs work before merging:

```
    [documentation] › test/documentation/workspaces.test.ts:35:7 › Workspaces sidebar › Workspaces context menu 
```

krassowski:
I'm planning to merge and cut a new pre-release early on Monday.

Issue #14237, opened 2023-03-20
File size column in File Browser is not visible at default width


JasonWeill:
## Description

﻿﻿﻿At the default width, the file browser is not wide enough to show the file size:

![image](https://user-images.githubusercontent.com/93281816/226480793-e779302f-d1ba-478f-8fc2-5840b7fd275f.png)

Expanding the left panel makes the file size column visible:

![image](https://user-images.githubusercontent.com/93281816/226480846-da0c97e4-7cba-473c-8611-0162e9403c0c.png)

## Reproduce

1. Launch JupyterLab
2. Make File Browser visible

## Expected behavior

All columns are visible, or if they are not visible, there is a visible indicator that more columns are present but can't be shown (e.g., a `»` button)

## Context

- Operating System and version: Windows 10 Enterprise 21H2
- Browser and version: Firefox 102.9.0esr
- JupyterLab version: 4.0.0a37

krassowski:
I would love to have the columns take up less space. There is some padding CSS tirckery we could do to alleviate it but otherwise we need to have some extra JS logic to either resize columns when a new column is added (whether just to fit or to show overflow indicator as proposed above) or to resize the sidebar itself.

Tagging for 4.0.x as this does not appear blocking for 4.0 release but would ideally be addressed early as it is a severe annoyance deteriorating UX.

JasonWeill:
Related:
* #8497 (no UI to signal existence of invisible columns)
* #3875 (resize columns of file browser)

JasonWeill:
Still an issue as of 4.2.0 alpha; the "File Size" column must be enabled by right-clicking on a column header, but at the default left panel width, it is not visible.

JasonWeill:
PR #15948, which fixes #15947, makes file size visible at the default width, although showing it makes the "Name" column effectively unreadable, and "File Size" does not fit in the tiny column header. Shortening it to "Size" would probably fit, at least in English:

![image](https://github.com/jupyterlab/jupyterlab/assets/93281816/b40f3961-e2b0-45cb-94fa-73d2de7d8b62)

PR #15948, opened 2024-03-08
Much smaller "Last Modified" column, date

### LINKED ISSUE #15947: Shorter "Last Modified" display in narrow columns
### Problem

The "Last Modified" column takes up a lot of horizontal space in a narrow container. In English, a relative date like "11 months ago" is likely longer than the filename it describes:

long-date

The absolute date appears on hover, in a tooltip.

### Proposed Solution

We use an [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) object to make human-readable dates in an internationalizable format. When changing that to use the "short" format, units like "months" shorten, but "days" does not, and "ago" remains, even though we can be reasonably certain that almost all files will have been created in the past.

short-date

As a more extreme alternative, we could reduce this to "#u", with a number and a very short unit. This pattern is already used by mobile apps and mobile web sites, and may require additional manual internationalization. We can also shorten "Last Modified" to "Mod", to allow for the column to be shortened.

image

Note that Asian languages already have single characters for most time units, 

In the rare case that a file's time stamp in the future, we can put a `+` before the number; the normal expectation is that a date is in the past.

### Additional context

See also #14237, #3875, and #8497 for other issues about columns in the filebrowser.

JasonWeill:
## References

Fixes #15947.

## Code changes

Shortens dates to extremely short units, as short as one character (`days` becomes `d`) and no longer than three characters (`min`, `mon`). All numbers are assumed to be negative; if a file's timestamp is in the future, a `+` is prepended.

## User-facing changes

Before the change, the relative timestamp was so long that the "last modified" minimum width was wider than the file name container's:

long-date

Now, the last modified column is shortened to "Modified", and is narrower:

Screenshot 2024-03-12 at 5 36 10 PM

One of three dates is displayed based on how wide the "Modified" column is:

3-step


andrii-i:
In my subjective opinion, middleground between this PR and version proposed in #15947 would be a good compromise between saving space and maintaining readability: "Modified" for the column name and "7 mo." for value format (with a space and a dot).

packages/coreutils/src/time.ts
```diff
@@ -25,17 +44,17 @@ export namespace Time {
    * @returns A formatted date.
    */
   export function formatHuman(value: string | Date): string {
-    const lang = document.documentElement.lang || 'en';
-    const formatter = new Intl.RelativeTimeFormat(lang, { numeric: 'auto' });
```
jtpio, line 29: Are we not losing support for other display languages by dropping this line?

krassowski:
I like the idea of having a shorter form _when needed_. I do not like forcing the shorter form and "Mod" header by default. Internationalisation-wise it would be great if we could off load this to self-contained package which supports such short format with translations.

jtpio:
> having a shorter form _when needed_.

Or at least configurable. For example Notebook 7 would likely keep the longer version by default.

krassowski:
Of note, this filebrowser is also used in Jupyter Notebook where it takes full-width of the page. It can be also used by user widgets and in other places where the short form does not make sense.

On the other hand, the file selection dialog like `FileDialog.getOpenFiles` should probably always use the short form.

packages/coreutils/src/time.ts
JasonWeill: We are, although we could use localizable strings for the ultra-short displays. I could imagine "4d" (4 days) becoming "4j" (4 jours) in French, for example, or "4日間" in Japanese. I'm not sure what's involved with translating them.

JasonWeill:
The same i18n-friendly library we use also supports a super-compact format called `narrow`, although it still includes `ago` or `in`:

```
const rtf1 = new Intl.RelativeTimeFormat('en', { style: 'narrow' });
console.log(rtf1.format(3, 'quarter'));
console.log(rtf1.format(-1, 'day'));
```

causes log messages:

```
> "in 3q"
> "1d ago"
```

JasonWeill:
With the most recent change, the name and modified columns flex in width, and based on the container width, one of three dates is displayed.

Notably, the [`csstree/stylelint-validator`](https://github.com/csstree/stylelint-validator) module we use flags container queries as invalid CSS, even though all modern browsers support them, and even though we're using the newest version of this module.

3-step

krassowski:
This UX is nice and who doesn't love the container queries (it's awesome that we finally can use them!) but rendering the three variants will have three times the cost. The file browser already has performance issues when the list of files is long - I would strongly suggest to attempt to reimplement the conditional rendering on JS rather than CSS level.

> Notably, the [csstree/stylelint-validator](https://github.com/csstree/stylelint-validator) module we use flags container queries as invalid CSS, even though all modern browsers support them, and even though we're using the newest version of this module

If you do stick to CSS queries you can add ignore comments disabling the validator around the problematic rules.

JasonWeill:

JasonWeill:
please update documentation screenshots

JasonWeill:

JasonWeill:
please update snapshots

JasonWeill:
Updated with added test cases, ready for review. This also fixes #15979, filed by @FoSuCloud .

JasonWeill:
The newly created screen shots are failing because the files are only seconds old. I wonder if there's a way to create a file with an older time stamp.

andrii-i:
I had a similar problem with timestamps in Jupyter Scheduler E2E tests and opted for using Playwright's Route object to intercept an API response and overwrite a timestamp to some constant value (you can use different timestamps to check different formatting options too):
* Implementation: https://github.com/jupyter-server/jupyter-scheduler/blob/main/ui-tests/helpers/SchedulerHelper.ts#L219
* Docs: https://playwright.dev/docs/api/class-route#route-fulfill

(Another option I tried was to [mask ](https://playwright.dev/docs/api/class-locator#locator-screenshot-option-mask) the timestamp but it was partly defeating the purpose of the test, checking that timestamp is displayed correctly).

JasonWeill:
Multiple UI tests are now failing with a timeout: `Test timeout of 60000ms exceeded while setting up "page".` This is after I tried to use the `page` element of the Galata context to call `await galata.Mock.freezeContentLastModified(page);` in an attempt to make all of the newly-created files one day old, to avoid problems with the screen shots.

JasonWeill:
Marking this as a possible fix for #14237, although the column header for "File Size" does not totally fit at the default width.

image

packages/filebrowser/style/base.css
```diff
@@ -296,9 +296,65 @@
   outline: 0;
 }
 
+/* This file uses container queries, which are supported by all the
+ * browsers that JupyterLab works with, but which stylelint cannot
+ * validate.
```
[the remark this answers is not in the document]
JasonWeill: Opened a sub-PR for this: https://github.com/JasonWeill/jupyterlab/pull/82

packages/coreutils/src/time.ts
```diff
@@ -17,16 +17,25 @@ const UNITS: { name: Intl.RelativeTimeFormatUnit; milliseconds: number }[] = [
  * The namespace for date functions.
  */
 export namespace Time {
+  // Intl.RelativeTimeFormatStyle contains these, but we can't compile with it yet
```
krassowski, line 20: Can you expand on this comment?

packages/filebrowser/src/listing.ts
```diff
@@ -2660,13 +2660,24 @@ export namespace DirListing {
...
+                new Date(model.last_modified),
+                style as Time.HumanStyle
+              ) +
+              '</div>'
+          )
+          .join('');
         modTitle = Time.format(new Date(model.last_modified));
       }
-      modified.textContent = modText;
+      modified.innerHTML = modHTML;
```
krassowski, line 2680: A solution in-between the re-write to an implementation that only ever renders one format, and this implementation which manually assembles the HTML code and forces it on the node (which will force-update the DOM each time resulting in bad performance), we could replace this fragment to use [`VirtualDOM.render`](https://lumino.readthedocs.io/en/latest/api/functions/virtualdom.VirtualDOM.render.html). This would mean that the DOM nodes are not touched unless there was a change.

galata/test/galata/notebook.spec.ts-snapshots/example-run-galata-linux.png
krassowski, line 1: Is this a spurious update?

krassowski (changes requested):
The visual test failures are relevant:

```
    [jupyterlab] › test/jupyterlab/cells.test.ts:25:7 › Cells › should collapse a cell ─────────────
    [jupyterlab] › test/jupyterlab/notebook-run-mermaid.test.ts:41:7 › Notebook Run Mermaid › Run notebook with Mermaid cell in default theme 
    [jupyterlab] › test/jupyterlab/notebook-run-mermaid.test.ts:62:7 › Notebook Run Mermaid › Run notebook with Mermaid cell in dark theme 
    [jupyterlab] › test/jupyterlab/notebook-run-vega.test.ts:41:7 › Notebook Run Vega › Run notebook with Vega cell in default theme 
    [jupyterlab] › test/jupyterlab/notebook-run-vega.test.ts:62:7 › Notebook Run Vega › Run notebook with Vega cell in dark theme 
    [jupyterlab] › test/jupyterlab/notebook-run.test.ts:33:7 › Notebook Run › Run Notebook and capture cell outputs 
    [jupyterlab] › test/jupyterlab/notebook-scroll-no-windowing.test.ts:282:7 › Notebook scroll over long outputs (no windowing) › should scroll smoothly without snapping to headings 
    [jupyterlab] › test/jupyterlab/output-scrolling.test.ts:54:7 › Output Scrolling › Switching with prompt overlay 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:36:9 › Sidebars › Open Sidebar tab filebrowser ─
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:36:9 › Sidebars › Open Sidebar tab jp-property-inspector 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:36:9 › Sidebars › Open Sidebar tab jp-running-sessions 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:36:9 › Sidebars › Open Sidebar tab table-of-contents 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:36:9 › Sidebars › Open Sidebar tab extensionmanager.main-view 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:51:7 › Sidebars › File Browser has no unused rules 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:86:7 › Sidebars › Left light tabbar (with text) 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:94:7 › Sidebars › Right dark tabbar (with text) 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:102:7 › Sidebars › Move File Browser to right ──
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:107:7 › Sidebars › Open File Browser on right ──
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:113:7 › Sidebars › Open Sidebar on right ───────
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:118:7 › Sidebars › Close Sidebar on right ──────
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:124:7 › Sidebars › Capture File Browser on right 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:132:7 › Sidebars › Move Debugger to left ───────
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:139:7 › Sidebars › Check Running Session button on sidebar has correct aria label and role 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:152:7 › Sidebars › Check Extension Manager button on sidebar has correct aria label and role 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:169:7 › Sidebars › Check File Browser button on sidebar has correct aria label and role 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:181:7 › Sidebars › Check Debugger button on sidebar has correct aria label and role 
    [jupyterlab] › test/jupyterlab/sidebars.test.ts:193:7 › Sidebars › Check Table of Contents button on sidebar has correct aria label and role 
```

packages/coreutils/src/time.ts
JasonWeill: I had intended to use the type `Intl.RelativeTimeFormatStyle` to represent the different relative time format styles. This is part of the standard JavaScript library, but is not available in VS Code, and may not compile using the TypeScript version we use.

packages/filebrowser/style/base.css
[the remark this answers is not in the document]
JasonWeill: This is blocked because of an unmet upstream dependency; see #16062.

packages/coreutils/src/time.ts
krassowski: What I meant is _which_ version of typescript is needed. From a quick check `typescript` `4.1.2` added  `es2020.intl` and I see `Intl.RelativeTimeFormatStyle` in https://github.com/microsoft/TypeScript/blob/release-4.1/src/lib/es2020.intl.d.ts whereas we use `5.1.6`, so I think we should be able to compile with it after adding `es2020.intl` to `compilerOptions` here:

https://github.com/jupyterlab/jupyterlab/blob/96c28b6107ffa7ff923819a3bec8b9be3b250925/packages/coreutils/tsconfig.json#L3

the same way as it is done for example here:

https://github.com/jupyterlab/jupyterlab/blob/96c28b6107ffa7ff923819a3bec8b9be3b250925/packages/ui-components/tsconfig.json#L3-L4

krassowski:
Visual regression tests are still failing. It looks like `examples` snapshots need updating too:

image

I can see that the differences are indeed relevant:

| Actual | Expected |
|--|--|
| example-actual | example-expected |

You can download the artifacts from https://github.com/jupyterlab/jupyterlab/actions/runs/8547285304?pr=15948 (at the bottom of the page, you need to be signed in).

JasonWeill:
@krassowski Thanks! Will take a look as soon as I can.

JasonWeill:
Some of the visual regression tests now show a border around the currently active cell, which I now see on initial load on my local computer. I don't see anything in my CSS changes that would cause this, though …

image

galata/test/documentation/custom_css.test.ts-snapshots/custom-css-notebook-markdown-documentation-linux.png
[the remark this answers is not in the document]
krassowski: Actually should not change at all.

packages/filebrowser/src/listing.ts
```diff
@@ -2736,14 +2747,17 @@ export namespace DirListing {
     /**
      * Create a node for a header item.
      */
-    protected createHeaderItemNode(label: string): HTMLElement {
+    protected createHeaderItemNode(label: string, title?: string): HTMLElement {
```
krassowski, line 2750: The `title` attribute gets added, but I don't see it being used. Did you mean to add title on "Modified"?

galata/test/jupyterlab/sidebars.test.ts
```diff
@@ -30,6 +27,10 @@ async function mockLabelOnFirstTab(tabbar: Locator, text: string) {
     }, text);
 }
 
+test.use({
+  tmpPath: 'test-sidebars'
+});
```
krassowski, line 32: It looks like none of the sidebar tests passes anymore. Could it be related to the changes here (removal of `mockState: true` and addition of `tmpPath: 'test-sidebars'`)?

packages/filebrowser/src/listing.ts
JasonWeill: I created this new parameter when I shortened the column header all the way to "Mod", but that was very unpopular, so I have reverted this change.

galata/test/jupyterlab/sidebars.test.ts
JasonWeill: Sidebar tests now pass, although other failures persist.

JasonWeill:
Sidebar tests are now passing, although the blue outline on the first active cell persists, causing numerous other test failures. Doesn't seem like something that this change should cause …

krassowski:
Yes it is perplexing indeed. Are you able to isolate whether this is a side-effect of the CSS changes or JS changes? I think we should understand _why_ this happens, but once we know I am not opposed to updating these snapshots (unless the "why" points to an issue though I don't expect it)

packages/coreutils/src/time.ts
JasonWeill: The type `RelativeTimeFormatStyle` is part of the `Intl` namespace defined here, in `es2020.intl.d.ts`:

https://github.com/microsoft/TypeScript/blob/066773b99ba1867903182e3b1e3899c7b202b217/src/lib/es2020.intl.d.ts#L68-L73

I modified `tsconfig.json` in `tsconfig.json` in each of `packages/coreutils`, `packages/filebrowser`, and `packages/lsp-extension`, to include:

```
  "compilerOptions": {
    "lib": ["DOM", "DOM.Iterable", "ES2020.Intl"],
```

and `jlpm build` still fails with an error:

```
    ✖  @jupyterlab/lsp-extension:build:all
       > @jupyterlab/lsp-extension@4.2.0-beta.0 build
       > tsc -b
       
       ../coreutils/lib/time.d.ts(12,62): error TS2694: Namespace 'Intl' has no exported member 'RelativeTimeFormatStyle'.
       npm ERR! Lifecycle script `build` failed with error: 
       npm ERR! Error: command failed 
       npm ERR!   in workspace: @jupyterlab/lsp-extension@4.2.0-beta.0 
       npm ERR!   at location: /Users/jweill/git/jupyterlab/packages/lsp-extension 
```

The problem persists after running `jlpm install`, then `jlpm build` again.

krassowski: Yes, I see it. So this is because other packages such as `translation`, `statusbar`, etc also need it because they also depend on coreutils (you can debug this by compiling `lsp-extension` with `-v` flag). The simple solution would be to set it in tsconfigbase:

https://github.com/jupyterlab/jupyterlab/blob/6152e3a5f94d728ccd2727d6d8413f12ddafb669/tsconfigbase.json#L3

But this might be more disruptive so I think we do not have to do this.

At some point we should update our target to ES2020.

For now I would suggest to make the comment more specific/easier to discover when searching for ES2018/2020:

```suggestion
  // Intl.RelativeTimeFormatStyle contains but it requires `ES2020.Intl` (as of writing we use `ES2018` target)
```

so this is easier to find when we will be making the transition.

JasonWeill:
The blue border is coming from the rule `.jp-Notebook.jp-mod-commandMode .jp-Cell.jp-mod-active:focus-visible` in `base.css` in `notebook`:

https://github.com/jupyterlab/jupyterlab/blob/5cfcae170275c4d7f5a60ba8df66d58b7b6df9a1/packages/notebook/style/base.css#L143-L149

It looks like this rule was introduced in #14115 by @fcollonval . With my code change, I see the focus indicator only on the very first cell of a notebook, not on any other cells.

JasonWeill:
I'm seeing the same "border on first cell" behavior in the `main` branch as well.

JasonWeill:
In a separate environment, I just installed JupyterLab 4.2.0b1. I see the blue focus indicator around all cells that I select. I believe that the changes I'm seeing in this change are likely consistent with the changes present in 4.2.0b1. May need to update screenshots to reflect this.

krassowski:
> In a separate environment, I just installed JupyterLab 4.2.0b1. I see the blue focus indicator around all cells that I select.  I believe that the changes I'm seeing in this change are likely consistent with the changes present in 4.2.0b1. May need to update screenshots to reflect this.

There has not been a change with respect to focus behaviour recently. On the other hand, the file browser a has a bespoke logic for manipulating focus. Could it be that changes in the DOM in this PR affected that logic in some way?

| Before | After |
|--|--|
| <video src="https://github.com/jupyterlab/jupyterlab/assets/5832902/a7249a87-c70c-4a6e-a5be-de8e62661ac6"> | <video src="https://github.com/jupyterlab/jupyterlab/assets/5832902/469e8eaf-142d-4315-8609-df9fb0c8a5c1"> |


> Arhg, I very much dislike how GitHub hides all comments behind a collapser after rebases.

krassowski:
Referring specifically to logic added in the PR https://github.com/jupyterlab/jupyterlab/pull/13577. I would suspect that the logic in:

https://github.com/jupyterlab/jupyterlab/blob/5cfcae170275c4d7f5a60ba8df66d58b7b6df9a1/packages/filebrowser/src/listing.ts#L2013-L2026

gets called at a different time now, and no longer steals focus from the notebook cells, or steals it at a different time . The change in focus stealing order means that the browser's built-in heuristic for `:focus-visible` now kicks in and shows outline around cells, whereas it was not showing previously.

I do not like the `requestAnimationFrame` call in that place. It makes things a bit non-deterministic. The thing that does worry me is that this suggests that this PR may have enough performance impact to actually shift the execution by a full animation frame (33ms) on a very small filebrowser listing (just four items). However, I do not see that in profiler:

image

Yes, it takes 3 times as long as it took, but the task is only 12ms so should be fine.

It would be perfect if the VirtualDOM did not update the DOM for elements with `display: none` but it I think even React does not do that (although we do that for some Widgets).

Thinking of how to implement this without rendering the column 3 times on every update, one solution would be adding a resize observer, storing down the column width, and re-rendering if width crossed a threshold.

For the time being I think I'm fine with just updating the snapshots.

> To add to the confusion, the jupyter-ui-toolkit is also doing some focus manipulation when adding the toolbar.

JasonWeill:
@krassowski The most recent commit removes the custom CSS rules and triple rendering. It attaches logic to the `onResize` function that attempts to rerender every item's listing. I have verified that the functionality is working as it was before; is this amount of rerendering a good performance move for us?

packages/filebrowser/src/listing.ts
```diff
@@ -2663,7 +2669,18 @@ export namespace DirListing {
       let modText = '';
       let modTitle = '';
       if (model.last_modified) {
-        modText = Time.formatHuman(new Date(model.last_modified));
+        // Render the date in one of multiple formats, depending on the container's size
+        const modifiedWidth = modified.getBoundingClientRect().width;
```
krassowski, line 2673: Thank you for addressing my suggestion! This line has to be moved out of here. This is because right now the rendering sequence leads to layout trashing and opening a larger folder takes 3 seconds in rendering time alone:

image

Two ways to do that (possibly more):
- create a private `this._modifiedWidth` and record it in `onResize()`, or
- pass it down as an argument to `updateItemNode()`

packages/filebrowser/src/listing.ts
```diff
@@ -941,6 +945,8 @@ export class DirListing extends Widget {
     const { width } =
       msg.width === -1 ? this.node.getBoundingClientRect() : msg;
     this.toggleClass('jp-DirListing-narrow', width < 250);
+    // Rerender item nodes, so that their modified dates update.
+    this.updateNodes(this._sortedItems, this._items);
```
krassowski, line 949: I would only re-render the last modified column instead. I would also bail if `this._modifiedWidth` has not crossed the boundary since the last time. Once this only does what it should I would consider putting it behind a throttler.

packages/filebrowser/src/listing.ts
JasonWeill: Recent changes only rerender the modified date elements, and no other elements of the file listings, when a resize event causes the date style to change.

packages/filebrowser/src/listing.ts
```diff
@@ -806,6 +809,91 @@ export class DirListing extends Widget {
     }
   }
 
+  // Update the modified column's size
+  updateModifiedSize(node: HTMLElement) {
```
krassowski, line 813: Let's minimize API surface so that we can change the signature easily in the future.

```suggestion
  /*
   * Update the modified column's size
   */
  private _updateModifiedSize(node: HTMLElement) {
```

packages/filebrowser/src/listing.ts
```diff
@@ -806,6 +809,91 @@ export class DirListing extends Widget {
...
+    const modified = DOMUtils.findElement(node, ITEM_MODIFIED_CLASS);
+    this._modifiedWidth = modified.getBoundingClientRect().width;
```
krassowski, line 815: Test failure rightly highlights that modified can be `null`.

```suggestion
    this._modifiedWidth = modified ? modified.getBoundingClientRect().width : 200;
```

(or some other default)

packages/filebrowser/src/listing.ts
```diff
@@ -227,6 +227,9 @@ export class DirListing extends Widget {
     this._manager = this._model.manager;
     this._renderer = options.renderer || DirListing.defaultRenderer;
 
+    // Get the width of the "modified" column
+    this.updateModifiedSize(this.node);
```
krassowski, line 231:
```suggestion
    this._updateModifiedSize(this.node);
```

krassowski:
Looking good! It looks like it the API could be narrowed a bit and null checks needs addressing.

packages/filebrowser/src/listing.ts
krassowski: `protected` would be fine with me too.

packages/filebrowser/src/listing.ts
```diff
@@ -941,6 +976,13 @@ export class DirListing extends Widget {
     const { width } =
       msg.width === -1 ? this.node.getBoundingClientRect() : msg;
     this.toggleClass('jp-DirListing-narrow', width < 250);
+
+    // Rerender item nodes' modified dates, if the modified style has changed.
+    const oldModifiedStyle = this._modifiedStyle;
+    this._updateModifiedSize(this.node);
```
krassowski, line 982: There is potential for layout trashing on 250px boundary here:
- we have read in line 977 (`this.node.getBoundingClientRect()`)
- write in line 978 (toggle class)
- read in `_updateModifiedSize()` via `getBoundingClientRect` on modified column
- write in `_updateModifiedSize()`

I am not sure if we can improve it much though, so no action requested.

krassowski (commented):
Tested the latest changes locally and it looks and feels great! Thank you for your persistence here. I left small suggestions regarding the public API so that we do not break extensions by this PR, and a tiny performance improvement (parsing date only once, this is like 0.25ms improvement on Firefox so barely any).

As for performance, resizing in a folder with 300 files is smooth now since the DOM update only happens on boundaries. It is not perfect with such a number of items because it takes a few animation frames rather than one (<33ms) for me on a good laptop.

<details>

On Chrome (~95ms - three animation frames) the cost appears to be mostly in formatting:

image

On Firefox (~170ms - five frames) the cost is also largely in formatting, and in DOM writes:
  
  image
  
  - `set Node.textContent`: 50ms 
  - `format()`: 45 ms
  - `formatHuman()`: 37ms
  - `Element.getBoundingClientRect()`: 26ms
  - `set HTMLElemet.title`: 8ms
  - `findElement()`: 4ms
  - `Date` constructor: 0.5ms
  
</details>

krassowski:
I guess one way to make it smoother would be first updating the date for visible files/directories, and deferring the update of remaining files for next animation frame. This can be done in a follow up PR and I think we would not need to change API for that (we would just call `updateItemModified` on a subset of items in batches).

packages/filebrowser/src/listing.ts
```diff
@@ -806,6 +809,100 @@ export class DirListing extends Widget {
...
+            item.last_modified,
+            this._modifiedStyle
+          );
+        } else {
+          // Render the date in one of multiple formats, depending on the container's size
+          modified.textContent = Time.formatHuman(
+            item.last_modified,
+            this._modifiedStyle
+          );
+          modified.title = Time.format(new Date(item.last_modified));
```
krassowski, line 842: I think this can be simplified; the fallback to the `defaultRenderer` is the pattern we use frequently when introducing new optional methods:

```suggestion
          DirListing.defaultRenderer.updateItemModified(
            modified,
            item.last_modified,
            this._modifiedStyle
          );
```

krassowski (commented):
One final suggestion + screenshots need updating as merge of https://github.com/jupyterlab/jupyterlab/pull/16046 introduced some conflicts. I would aim to merge this today and cut another beta once this is in.

JasonWeill:
Thanks for catching this! Reverted here and cherry-picked in the #15881 branch.

krassowski:
bot please update snapshots

galata/test/jupyterlab/output-scrolling.test.ts-snapshots/prompt-overlay-hover-normal-jupyterlab-linux.png
krassowski, line 1: Odd

JasonWeill: I don't think this change is intended, but the new screenshot certainly looks better 😄

krassowski (approved):
Thank you @JasonWeill. Good to merge once all green.

krassowski:
The only failure:

```
[jupyterlab] › test/jupyterlab/sidebars.test.ts:42:9 › Sidebars › Open Sidebar tab jp-running-sessions 
```

should be fixed by merge of https://github.com/jupyterlab/jupyterlab/pull/16188 (review welcome!)

packages/filebrowser/src/listing.ts
```diff
@@ -2297,7 +2368,7 @@ export namespace DirListing {
       const trans = translator.load('jupyterlab');
       const name = this.createHeaderItemNode(trans.__('Name'));
       const narrow = document.createElement('div');
-      const modified = this.createHeaderItemNode(trans.__('Last Modified'));
+      const modified = this.createHeaderItemNode(trans.__('Modified'));
```
jtpio, line 2371: Would it be possible to keep the original `Last Modified` here?

Since the file browser is also used in Notebook, and there is plenty of space to display it.

jtpio: Noticed in https://github.com/jupyter/notebook/pull/7333

PR #15949, opened 2024-03-08
Case insensitive match for completion items filtering

christophediprima:
Missing a letter is allowed by matchSumOfSquares so it makes more sense to make filtering case insensitive.

### DIFF
--- a/packages/completer/src/model.ts
+++ b/packages/completer/src/model.ts
@@ -390,7 +390,7 @@ export class CompleterModel implements Completer.IModel {
         index > -1
           ? originalItem.label.substring(0, index)
           : originalItem.label;
-      const match = StringExt.matchSumOfSquares(escapeHTML(text), query);
+      const match = StringExt.matchSumOfSquares(escapeHTML(text.toLowerCase()), query.toLowerCase());
       // Filter non-matching items.
       if (match) {
         // Highlight label text if there's a match


krassowski (changes requested):
Thank you for the contribution @christophediprima.

It certainly does make sense to allow this as a setting. jupyterlab-lsp has a setting for it [here](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/71c8e0c53b6f3b609dc9ebab21fc53a45d7d8e50/packages/jupyterlab-lsp/schema/completion.json#L50-L55). As whatever the default should be - opinions will vary.

Can you try to add the setting and unit tests for completion with this case sensitive and insensitive matchin on and off?

christophediprima:
Sure but I am a bit busy right now. I will have a look when I come back from holidays! :)

krassowski:
@christophediprima did you have a chance to take another look? It seems to have missed to 4.2 merge window, I will switch this PR to draft.

packages/completer/src/model.ts
```diff
@@ -390,7 +390,7 @@ export class CompleterModel implements Completer.IModel {
         index > -1
           ? originalItem.label.substring(0, index)
           : originalItem.label;
-      const match = StringExt.matchSumOfSquares(escapeHTML(text), query);
+      const match = StringExt.matchSumOfSquares(escapeHTML(text.toLowerCase()), query.toLowerCase());
```
chanhee-park, line 393: I think using `toLocaleLowerCase()` instead of `toLowerCase()` might be a better call here. It tends to handle language-specific rules more gracefully, which could save us some headaches down the line. Plus, it aligns better with user expectations, especially in multilingual environments. Cheers!

chanhee-park (changes requested):

krassowski:
@christophediprima are you still interested in finishing this one? There is about 2 weeks left to complete it before 4.3 otherwise we will need to bump it to next release or close out to signal that anyone else can open a fresh PR if they would like to contribute this option.

PR #15957, opened 2024-03-10
Fix Theme color is not applied to Toolbar Button

FoSuCloud:
## References

## Code changes

截屏2024-03-10 22 05 35

## User-facing changes

The color of `.jp-Toolbar .jp-ToolbarButtonComponent` use `var(--neutral-foreground-rest)`

If there is no higher weight, the color of `.jp-Toolbar .jp-ToolbarButtonComponent` use `var(--jp-ui-font-color1)`

### DIFF
--- a/galata/test/jupyterlab/toolbars.test.ts
+++ b/galata/test/jupyterlab/toolbars.test.ts
@@ -37,23 +37,47 @@ toolbars.forEach(([plugin, parameter]) => {
     expect(missingCommands).toEqual([]);
   });
 });
+test.describe('Toolbar Button', () => {
+  test.beforeEach(async ({ page, tmpPath }) => {
+    await page.notebook.createNew();
+  });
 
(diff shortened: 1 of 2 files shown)


FoSuCloud:
@meeseeksdev tag bug

krassowski:
@AllanChain would you be happy to review this PR?

AllanChain:
Yes, I'll review this in the next couple of days.

FoSuCloud:
> I've checked this with custom theme enabled, and it's working well.
> 
> Just a small note: The button color for the default dark theme is also changed (from pure white to gray, which is how it looks like in JupyterLab 3.x), but there seems no UI test covering this, or the color change is so small that it is within tolerance.

Yes we are missing this ui test case. I will complete it later

AllanChain (approved):

krassowski:

Issue #15931, opened 2024-03-06
Can't open a notebook from a drive using the running panel


jtpio:
## Description

Opening a notebook from a custom drive using the running panel does not work.

[bug-open-notebook-from-drive.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/2382e4f6-6aa2-4400-ab6b-55cd09c2340e)

## Reproduce

1. Install a custom drive, for example https://github.com/jupyterlab/jupyterlab-github/
2. Open a notebook from the drive
3. Close it
4. Go to the running panel
5. Click on the notebook to re-open it

For example on Binder: https://mybinder.org/v2/gh/jupyterlab/jupyterlab-github/main?urlpath=lab

## Expected behavior

It should re-open the notebook.

## Context

- Operating System and version: All
- Browser and version: All
- JupyterLab version: 4.2.0a0

PR #15958, opened 2024-03-10
Fix opening active notebook from kernel - custom drive 

itsmevichu:
Fixes [#15931](https://github.com/jupyterlab/jupyterlab/issues/15931)


itsmevichu:
I've noticed that localPath is intentionally used for the session creation. Is there a specific reason for this?

krassowski:
The test failure appears related, if this change is indeed desired we should update the the tests too:

```

  ● docregistry/context › Context › #constructor() › should set the session path with local path

    expect(received).toEqual(expected) // deep equality

    Expected: "373e1985-29d0-487c-9153-91ae151c002b.txt"
    Received: "TestDrive:373e1985-29d0-487c-9153-91ae151c002b.txt"

      64 |         });
      65 |
    > 66 |         expect(context.sessionContext.path).toEqual(localPath);
         |                                             ^
      67 |       });
      68 |     });
      69 |

      at Object.<anonymous> (test/context.spec.ts:66:45)


Error: expect(received).toEqual(expected) // deep equality

Expected: "373e1985-29d0-487c-9153-91ae151c002b.txt"
Received: "TestDrive:373e1985-29d0-487c-9153-91ae151c002b.txt"
```

krassowski:
@jtpio this looks reasonable to me but wanted to check with you as the author of https://github.com/jupyterlab/jupyterlab/issues/15931 and someone who recently worked with the drive API - is this the right solution?

jtpio:
The changes does look reasonable indeed.

As originally noticed in https://github.com/jupyterlite/jupyterlite/pull/1227 there were some changes in JupyterLab 4 when it comes to the handling of these paths. Which led to this comment: https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091

Looks like it was needed for the collaboration extension, although not sure why exactly. It could be because the collaboration drive replaces the default file browser.

jtpio:
Trying from the Binder for this PR (which has `jupyter-collaboration` installed), it does seem to be fixing the UX issue mentioned https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091 about the drive prefix being missing when hovering on the kernel in the running tab. However "New Console for Notebook" seems to be starting a different kernel, likely because the paths are now different:

image

itsmevichu:
@jtpio, I've tried opening the console via 'New Console for Notebook,' where it prompts to select a kernel. Upon choosing the notebook's kernel, the console opens with that selection as expected.

https://github.com/jupyterlab/jupyterlab/assets/78495319/93ac8422-6d2b-40be-971c-07d22fffc766

jtpio:
@itsmevichu normally the expected behavior is the same kernel being reused for both the notebook and the console:

[jupyterlab-new-console-for-notebook.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/a462946c-2873-4125-ba2a-55a47ed85641)

Zsailer:
@itsmevichu how's it going here? 

It looks like this PR will solve [this issue](https://github.com/jupyterlab/jupyter-collaboration/issues/242) on jupyter-collaboration. 🚀  

I was planning to work on PR but saw you had started one. Great work! I'm happy to help test and review going forward. 

I also see the regression @jtpio mentioned when opening a console from a previous session when testing your PR. Let us know if you have questions here to push this PR through. Thanks!

Zsailer:
>  However "New Console for Notebook" seems to be starting a different kernel, likely because the paths are now different:

@jtpio I think this is may be caused by jupyter-collaboration's replacement of the file browser, because I only see this behavior when jupyter-collaboration is installed (I tested this PR locally with + without JC).

jtpio:
Thanks @Zsailer for looking into this!

> this is may be caused by jupyter-collaboration's replacement of the file browser, because I only see this behavior when jupyter-collaboration is installed (I tested this PR locally with + without JC).

OK so if this fixes the default install (without RTC), maybe we can look into fixing this in `jupyter-collaboration` directly?

krassowski:
@Zsailer following up here, what do you think we should do here?

jtpio:
Maybe a rebase / update of the branch to make sure CI still passes?

The last comments suggest there might be something to fix in `jupyter-collaboration`. Which means we could merge this PR as it fixes the case when RTC is not enabled, and look into fixing the issue separately?

gjmooney:
I just tried this PR locally and it resolves this issue in `jupyter-drives` QuantStack/jupyter-drives#46

jtpio:
Thanks for testing!

It's been a while, and in the meantime it's possible that the issue in `jupyter-collaboration` has been fixed by the use of content providers.

I just tried on Binder with the merge commit: https://mybinder.org/v2/gh/itsmevichu/jupyterlab/open-notebook-from-kernel-custom-drive?urlpath=lab

And it seems to be fixing the issue 👍 

https://github.com/user-attachments/assets/40bd50ca-14b5-4464-ba4a-06e58f90aef2




To be sure, maybe we could wait before backporting to 4.4.x, so the fix could be tested with a JupyterLab 4.5 alpha release first?

Issue #15808, opened 2024-02-16
Flaky js-terminal test `getRasterizedGlyph`


krassowski:
## Description

The js-terminal test fails from time to time with:

```
  ● terminal/index › Terminal › #session › should set the title when ready

    TypeError: Cannot read properties of undefined (reading 'getRasterizedGlyph')

      at l.cell [as _drawChars] (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/BaseRenderLayer.ts:374:50)
      at handleGridChanged (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/TextRenderLayer.ts:246:10)
      at l._drawBackground [as _forEachCell] (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/TextRenderLayer.ts:162:11)
      at l.cell [as _drawForeground] (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/TextRenderLayer.ts:243:43)
      at lastRow [as handleGridChanged] (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/TextRenderLayer.ts:258:36)
      at _.end [as renderRows] (../../node_modules/xterm-addon-canvas/lib/webpack:/CanvasAddon/src/CanvasRenderer.ts:139:34)
      at u.this [as _renderRows] (../../node_modules/xterm/lib/webpack:/xterm/src/browser/services/RenderService.ts:158:9)
      at t.RenderDebouncer.this [as _renderCallback] (../../node_modules/xterm/lib/webpack:/xterm/src/browser/services/RenderService.ts:70:5)
      at t.RenderDebouncer.end [as _innerRefresh] (../../node_modules/xterm/lib/webpack:/xterm/src/browser/RenderDebouncer.ts:73:33)
      at _innerRefresh (../../node_modules/xterm/lib/webpack:/xterm/src/browser/RenderDebouncer.ts:55:11)
      at invokeTheCallbackFunction (../../node_modules/jsdom/lib/jsdom/living/generated/Function.js:19:26)
      at runAnimationFrameCallbacks (../../node_modules/jsdom/lib/jsdom/browser/Window.js:603:13)
      at Timeout._onTimeout (../../node_modules/jsdom/lib/jsdom/browser/Window.js:581:11)
```

[...]

PR #15962, opened 2024-03-11
Upgrade xterm.js from 5.1.0 to 5.5.0

krassowski:
## References

- fixes https://github.com/jupyterlab/jupyterlab/issues/15808
- migrates to new packages under new `@xterm` namespace which is important to make sure we get security updates as the old packages (with no namespace) are no longer maintained as these were susceptible to typo-squatting attacks

## Code changes

None

## User-facing changes

- https://github.com/xtermjs/xterm.js/releases/tag/5.2.0:
  - Support for the overline attribute SGR 53/55
  - 15 bug fixes
- https://github.com/xtermjs/xterm.js/releases/tag/5.3.0:
  - The default DOM-based renderer is significantly faster (not used in JupyterLab)
  - Smooth scroll improvements (not enabled it JupyterLab)
  - `minimumContrastRatio` improvements (not enabled in JupyterLab)
  - 12 bug fixes
- https://github.com/xtermjs/xterm.js/releases/tag/5.4.0:
   - 22 bug fixes
- https://github.com/xtermjs/xterm.js/releases/tag/5.5.0:
   - fix for DOM renderer blink animation
   - improve detection of node and browser - this was a blocker for us

None


krassowski:
Blocked on:
- https://github.com/xtermjs/xterm.js/issues/4995

depending on response we might decide to stick to 5.3.x in the meantime

krassowski:
The patch that we need was merged but it is targeting 5.5.0 which is due by May 03, 2024 (https://github.com/xtermjs/xterm.js/milestone/79) so it might not make it for our 4.2.0 release.

krassowski:
On the other hand, we could upgrade to 5.5.0 beta now, and delay our planned 4.2.0 release by two days to get the final 5.5.0 in.

krassowski:
So thanks to an early 5.5.0 release we can include it before beta1 :tada:

fcollonval (approved):
Thanks @krassowski

Issue #15258, opened 2023-10-11
Full windowed mode for notebooks enabled by default


krassowski:
### Problem

Full windowing mode brings performance enhancements for users with long notebooks. This mode being disabled by default means that users do not benefit from it, and means that it may not be fully tested/may regress over time.

### Proposed Solution

Enable the mode by default.

### Additional context

It was disabled in the run-up to 4.0 release by https://github.com/jupyterlab/jupyterlab/pull/14490 because of a number of regressions, as noted in the discussion in https://github.com/jupyterlab/jupyterlab/issues/14455

Work is ongoing on an additional UI element which will enhance the user awareness and navigation in the `full` mode:
- https://github.com/jupyterlab/jupyterlab/pull/15533

### Blockers

There is a number of issues which we need to address before we enable `full` windowing mode, most related to scrolling:
- https://github.com/jupyterlab/jupyterlab/issues/13343
- https://github.com/jupyterlab/jupyterlab/issues/14092
- https://github.com/jupyterlab/jupyterlab/issues/14878
- https://github.com/jupyterlab/jupyterlab/issues/14881
- https://github.com/jupyterlab/jupyterlab/issues/15073
- https://github.com/jupyterlab/jupyterlab/issues/15330
- https://github.com/jupyterlab/jupyterlab/issues/15352
- https://github.com/jupyterlab/jupyterlab/issues/15407
- https://github.com/jupyterlab/jupyterlab/issues/15415
- https://github.com/jupyterlab/jupyterlab/issues/15465
- https://github.com/jupyterlab/jupyterlab/issues/15594
- https://github.com/jupyter/notebook/issues/7231

### Nice to have

- https://github.com/jupyterlab/jupyterlab/issues/13172
- https://github.com/jupyterlab/jupyterlab/issues/16007
- https://github.com/jupyterlab/jupyterlab/issues/14728

Related meta-issues:
- https://github.com/jupyterlab/jupyterlab/issues/14106

fcollonval:
@krassowski I have time to work on those issues. Could you please list the one you are currently working on to avoid overlap?

krassowski:
I self-assigned to #14881 #14878 and #15073. It is a pity GitHub paused onboarding to [task lists](https://docs.github.com/en/issues/managing-your-tasks-with-tasklists/creating-a-tasklist) which we used in nbdime.

fcollonval:
[...]

Issue #15801, opened 2024-02-14
4.2 Release Plan


krassowski:
## Current Status

`4.1.0` was released two weeks ago. A first [`4.1.1`](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.1.1) patch version was released on Monday.

This is an issue to track the features for the `4.2` release.

## Features

There are currently 85 items in the `4.2.0` milestone: https://github.com/jupyterlab/jupyterlab/milestone/76.

We can document major planned features for coordination. PR ready by the feature freeze date will be reviewed, but please aim to share things you plan to contribute to 4.2.0 ahead (by commenting in this issue, or editing this list if you have access) to allow time for discussion.

**Feel free to edit this list or comment below to add more items**:
- [ ] https://github.com/jupyterlab/jupyterlab/pull/15448 (implemented in `jupyter-collaboration` extension)
- [ ] https://github.com/jupyterlab/jupyterlab/pull/15157 (deferred to 4.3)
- [x] https://github.com/jupyterlab/jupyterlab/pull/15483
- [x] https://github.com/jupyterlab/jupyterlab/pull/15946
- [x] https://github.com/jupyterlab/jupyterlab/pull/15964
- [x] https://github.com/jupyterlab/jupyterlab/pull/15962
- [x] https://github.com/jupyterlab/jupyterlab/pull/15845
- [x] https://github.com/jupyterlab/jupyterlab/pull/15623
- [ ] https://github.com/jupyterlab/jupyterlab/pull/15155 (deferred)

## Timeline (provisional)

- feature freeze: 27 March (beta0 release; no additional pull requests adding new features would be considered past this date)
- first release candidate: 17 April (rc0 release if no release blockers identified)
- final release: <s>1 May</s> 3 May  (at a minimum one week after RC)

krassowski:
The first alpha of 4.2.0 is out (https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0a0, https://pypi.org/project/jupyterlab/4.2.0a0/). As a remainder, by the timeline above there are three more weeks to prepare pull requests that would go into 4.2.0 release.

jtpio:
For reference the first Notebook `7.2.0a0` is out: https://github.com/jupyter/notebook/releases/tag/v7.2.0a0

krassowski:
The first beta for 4.2.0 is out:
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0b0
[...]

PR #15964, opened 2024-03-11
Enable `full` notebook windowing mode by default

krassowski:
## References

- Closes https://github.com/jupyterlab/jupyterlab/issues/15258
- https://github.com/jupyterlab/jupyterlab/issues/15801

## Code changes

Enables `full` notebook windowing mode by default to gather wider user feedback during JupyterLab 4.2.0 pre-release testing. If we do not get far enough with tackling issues from https://github.com/jupyterlab/jupyterlab/issues/15258 before 4.2.0 we can revert this PR.

## User-facing changes

JupyterLab works faster when long notebooks are open.

None


jtpio:
Linking to https://github.com/jupyter/notebook/issues/7231 as something we will likely want to fix if this becomes the default.

Unless Notebook still defaults to `defer` to avoid issues originally spotted in https://github.com/jupyter/notebook/pull/6855 (not sure these still apply though)

krassowski:
Thanks, I added https://github.com/jupyter/notebook/issues/7231 to the list in https://github.com/jupyterlab/jupyterlab/issues/15258.

Looking at visual regression CI failures, something to consider is whether we want to pin some tests to use `defer` mode, or whether we want more tests to run in both `defer` and `full` mode. I think both solutions may be desirable, depending on specific test.

krassowski:
```
[galata] › test/galata/notebook.spec.ts:22:7 › Notebook Tests › Create Markdown cell ───────────
```

| Before | After |
|---|--|
| image | image |


This one was taking snapshots of the entire main area; I think we should not do that.

krassowski:
bot please update snapshots

krassowski:
@gabalafou I hope the trimming of snapshots in this PR will make you happy :)

fcollonval (approved):
Thanks @krassowski

Issue #13838, opened 2023-01-23
Remove `node-fetch` from the `testing` package


jtpio:
### Problem

`node-fetch` is currently still used in the `testing` package: https://github.com/jupyterlab/jupyterlab/blob/aa3e1e07ffbe3db40c7ee71662114513b5492360/packages/testing/package.json#L47

As a follow-up to https://github.com/jupyterlab/jupyterlab/pull/13722, we should look into removing it now that Node 18 supports `fetch` by default.

### Proposed Solution

Simplify the testing setup by removing `node-fetch` if possible.

### Additional context

https://github.com/jupyterlab/jupyterlab/pull/13722

PR #15969, opened 2024-03-12
Remove `node-fetch` dependency

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/13838

Looking into cleaning up test dependencies.

## Code changes

- Remove `node-fetch` dependency from `@jupyterlab/testing`
- Recommend Node.js 20 for development

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/linuxjs-tests.yml
+++ b/.github/workflows/linuxjs-tests.yml
@@ -71,8 +71,6 @@ jobs:
 
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
-        with:
-          node_version: '20.x'
 
       - name: Install dependencies
         env:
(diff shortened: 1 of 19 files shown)


jtpio:
Might still need https://github.com/jsdom/jsdom/issues/1724

jtpio:
OK CI now seems to be passing. As mentioned above, some tests were relying on the default behavior of the previous version of `node-fetch` (v2) that was adding a status text by default. This PR updates the tests to not include the `statusText` checks anymore.

> Recommend Node.js 20 for development

A follow-up could be to bump Node 20 for as the minimal supported version, in other places in the code base.

krassowski (changes requested):

I think the required version should be also bumped in: https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#installing-node-js-and-jlpm before we merge this one to avoid bad first-contributor experience.

PR #16000, opened 2024-03-14
Remove `whatwg-fetch` polyfill

jtpio:
## References

Web browsers have had support for `fetch` for quite a few years now: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility

So we should be able to perform this cleanup on the whole code base.

## Code changes

Remove `whatwg-fetch` polyfill

## User-facing changes

None

None


krassowski (approved):
Thank you @jtpio!

Issue #15594, opened 2024-01-02
Page up/page down partially scrolls notebook back


krassowski:
## Description

Pressing Pg up/pg down in a cell scrolls notebook in unexpected way.

## Reproduce

1. Switch windowing mode to full
2. Open a notebook with more cells than fit in the viewport
3. Press pg down in the cell once
4. See that it navigates to top/bottom of the cell (ok)
5. Press pg down in the cell again
6. See that it scrolls notebook (ok)
7. Press pg down in the cell again and again
8. See that notebook scrolls in chaotic way jumping back to the active cell (not ok)

![unexpected-scroll](https://github.com/jupyterlab/jupyterlab/assets/5832902/bd5e9145-cd83-4131-a8ff-237a927e10fa)

## Expected behavior

Scrolling in full windowing mode behaves predictably.

## Context

- Browser: Chrome
- JupyterLab version: 4.1.0b0

JasonWeill:
Because windowed notebooks are not the default in 4.1.0, this bug, while noticeable and annoying, should not block a 4.1.0 release.

krassowski:
This is interesting, so `keydown` is invoked in both `full` and `none`/`defer` mode; the difference comes down to `activeCell.inViewport` property evaluating to `false` in `full` mode (correctly) and to `true` in `none`/`defer` modes (incorrectly?).

https://github.com/jupyterlab/jupyterlab/blob/4df55ca002fdd83773aaacc7f390b2051a3d7e0f/packages/notebook/src/widget.ts#L2212-L2217

`inViewport` is just a boolean flag:

https://github.com/jupyterlab/jupyterlab/blob/4df55ca002fdd83773aaacc7f390b2051a3d7e0f/packages/cells/src/widget.ts#L240-L251

which gets set when windowing is active:

https://github.com/jupyterlab/jupyterlab/blob/4df55ca002fdd83773aaacc7f390b2051a3d7e0f/packages/notebook/src/windowing.ts#L238

It might be that we do not need to call `focus` if the cell is out of viewport, but it needs more investigation. The condition above was recently modified in https://github.com/jupyterlab/jupyterlab/pull/15286.

krassowski:
[...]

PR #16006, opened 2024-03-17
Fix scrolling on editor interactions when active cell is out of view in windowed mode

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15594

## Code changes

- [x] Adds test cases for scrolling (should fail in the first commit)
- [x] Fixes scrolling on editor interactions, taking advantage of new `EditorView.scrollHandler`:
   - old logic focused editors that were out of view in windowed mode upon any keypress; it was also used to ensure scrolling, but it had a bad side effect of scrolling on non-input and non-selection key presses, such as <kbd>Alt</kbd>; now the condition permitting this code path when active cell is out of view was removed in favour of dedicated `scrollRequested` logic (see below)
   - when `Cell` is out of viewport (in windowed mode) and the editor requests scrolling it does not attempt to scroll as this is futile (the editor does not know by how much to scroll the virtualized container); instead it emits a new `scrollRequested` signal,
   - the windowed notebook receives the `scrollRequested` signal, scrolls the cell into the view and then scrolls the editor as requested.

## User-facing changes

Interacting with out-of view cells in windowed mode is more predictable and reflects the interaction model seen in non-windowed mode.

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -205,8 +205,9 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
     // For cells disable searching with CodeMirror search panel.
     this._editorConfig = { searchWithCM: false, ...options.editorConfig };
     this._editorExtensions = options.editorExtensions ?? [];
+    this._editorExtensions.push(this._scrollHandlerExtension);
     this._placeholder = true;
-    this._inViewport = false;
+    this._inViewport = null;
     this.placeholder = options.placeholder ?? true;
 
     model.metadataChanged.connect(this.onMetadataChanged, this);
@@ -239,9 +240,14 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
 
   /**
    * Whether the cell is in viewport or not.
+   *
+   * #### Notes
+   * This property is managed by the windowed container which holds the cell.
+   * When a cell is not in a windowed container, it always returns `false`,
+   * but this may change in the future major version.
    */
   get inViewport(): boolean {
-    return this._inViewport;
+    return this._inViewport ?? false;
   }
   set inViewport(v: boolean) {
     if (this._inViewport !== v) {
@@ -553,6 +559,13 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
   }
 
   /**
+   * Signal emitted when cell editor requests scrolling.
+   */
+  get editorScrollRequested(): ISignal<Cell, Cell.IEditorScrollRequest> {
+    return this._editorScrollRequested;
+  }
+
+  /**
    * Create children widgets.
    */
   protected initializeDOM(): void {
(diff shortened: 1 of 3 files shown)


packages/cells/src/widget.ts
```diff
@@ -552,6 +558,13 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
     }
   }
 
+  /**
+   * Signal emitted when cell editor requests scrolling.
+   */
+  get editorScrollRequested(): ISignal<Cell, Cell.IEditorScrollRequest> {
```
krassowski, line 564: Based on the investigation in https://github.com/jupyterlab/jupyterlab/issues/16007#issuecomment-2003496842 it appears likely that a similar signal will be needed for stdin box to fix a related issue with cell not being scrolled to when user types into the stdin box. Maybe it is better to rename this signal to a generic `scrollRequested`?

krassowski: Change made in 438e829

brichet:
Thanks @krassowski for making this PR.

I probably missed some use case, but I wonder if we need such complexity.
When the cell is 'hidden' because it's outside of the viewport, it is not really hidden but its height is set to `0px`. However the editor keep a height, so the scroll request should work even if the cell is not in the viewport.

It seems to work with:
- the signal only sending a boolean (whether the cell is in the viewport or not), and that boolean used to call (or not) `scrollToItem` from the Notebook widget. No need for the function `scrollWithinCell` in the signal.
- `_scrollHandlerExtension` always returns `true` to scroll in the editor anyway.

Again I may have missed some use case where it wouldn't work as expected.

krassowski:
> However the editor keep a height, so the scroll request should work even if the cell is not in the viewport.

No, it will not work. CodeMirror will attempt to scroll the container in which it is embedded (see the code of [`scrollRectIntoView`](https://github.com/codemirror/view/blob/0c538f74a544014d26ff0dabe082ee427c6fad44/src/dom.ts#L116C17-L116C35) function in CM6 - it attempts to scroll each parent container to make the editor visible). It cannot scroll that container correctly because that container does not have a defined height in general in the windowing situation.

> the signal only sending a boolean (whether the cell is in the viewport or not), and that boolean used to call (or not) `scrollToItem` from the Notebook widget. No need for the function `scrollWithinCell` in the signal.

So the thing I think you are missing is the fact that a cell may be longer than the viewport. In that case after we scrolled to a cell, we may have scrolled to a beginning of a cell. But if the user's cursor was at the end of the cell, the user will be confused.

Similarly, if the cell is longer than the height of the viewport but user types in the output (e.g. after running `input()` in a long code cell), we do not want to scroll just to the cell but to the cell output - which is tracked in https://github.com/jupyterlab/jupyterlab/issues/16007 but the design of `scrollWithinCell()` in this PR will enable it.

On some level, passing the exact location within cell to avoid scrolling twice would be ideal, but is not achievable with CodeMirror API as of today, and would be way more complex on the level of calculating the appropriate scroll delta.

> `_scrollHandlerExtension` always returns true to scroll in the editor anyway.

I do not understand this point.

krassowski:
Here are two screen recording showing when the logic for `scrollWithinCell` kicks in:

1. Pressing arrow key moves cursor: (a) if the cursor is in the top part of the cell no second scroll is made (b) if the cursor is in the bottom part immediately after scrolling to cell `scrollWithinCell` activates scrolling the cursor into view

why-needed

2. With page down/up it is even more obvious: it first scrolls to cell and then within cell.

page-down-up

brichet:
> So the thing I think you are missing is the fact that a cell may be longer than the viewport. In that case after we scrolled to a cell, we may have scrolled to a beginning of a cell. But if the user's cursor was at the end of the cell, the user will be confused.

Thanks for the clarification and records, this was indeed the case I was missing.

brichet (approved):
Looks good to me, thanks @krassowski

PR #16011, opened 2024-03-18
Fix `ServerApp.token` deprecation warnings

jtpio:
## References

This should fix the following warnings, noticed on some CI runs:

```
jupyterlab/tests/test_jupyterlab.py::test_load_extension
  /opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/traitlets/traitlets.py:1897: DeprecationWarning: ServerApp.token config is deprecated in jupyter-server 2.0. Use IdentityProvider.token
    return t.cast(Sentinel, self._get_trait_default_generator(names[0])(self))
```

## Code changes

Use `IdentityProvider` to get and set the token, as suggested in the warning.

For reference the warning was added in https://github.com/jupyter-server/jupyter_server/pull/825

And JupyterLab depends on Jupyter Server 2 as the minimum version: https://github.com/jupyterlab/jupyterlab/blob/9eea31bc38c7118cfe725e2f60a808f0a3e4c05e/pyproject.toml#L44

## User-facing changes

None

None

### DIFF
--- a/binder/jupyter_config.py
+++ b/binder/jupyter_config.py
@@ -9,7 +9,7 @@ common = [
     "--debug",
     "--port={port}",
     "--ServerApp.ip=127.0.0.1",
-    '--ServerApp.token=""',
+    '--IdentityProvider.token=""',
     # Disable dns rebinding protection here, since our 'Host' header
     # is not going to be localhost when coming from hub.mybinder.org
     "--ServerApp.allow_remote_access=True",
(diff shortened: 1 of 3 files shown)


krassowski (approved):
Thank you @jtpio!

> Note to self: https://github.com/jupyter-server/jupyter_server/pull/825 was released in 2.0.

PR #9898, opened 2021-03-04
Add websocket token auth in case of different ws domain

darcsoel:
Related issues:
https://github.com/jupyterlab/jupyterlab/issues/9070
https://github.com/jupyterlab/jupyterlab/pull/9080

As user I want to have possibility to use ws on another domain, this PR added one more check to appendToken method.

Issue #9070, opened 2020-09-24
Terminal token authentication broken


languy:
## Description
When configuring jupyter server with token authentication, the terminal fails to connect.
It appears that the latest service/terminal code ignores the `token` field passed in `ServerConnection.ISettings` when establishing the websocket connection.

The url should look like: `wss://__my_hostname__/notebook/terminals/websocket/1?token=__my_token__`
but is missing the `?token=__my_token__` part.

I believe the faulty code is in [service/terminal/default.ts](https://github.com/jupyterlab/jupyterlab/blob/2bade97edadea865b294eb49c8d693d4261ca404/packages/services/src/terminal/default.ts#L261). The url is constructed without the token parameter (if specified), right before establishing the websocket connection.

The package `@jupyterlab/services@4.2.0` used to work and properly form the url with the token in the query parameters, but I believe the latest refactoring from `TerminalSession` to `TerminalManager` broke it.

## Reproduce
1. Configure jupyter server to use token authentication
2. Open new terminal using `@jupyterlab/services@5.2.4` and `@jupyterlab/services@2.2.3`.

An error appears in the console:
```
WebSocket connection to 'wss://__hostname__/notebook/terminals/websocket/1' failed: HTTP Authentication failed; no valid credentials available
```
Notice the token missing in the url parameters.

## Expected behavior
Terminal connects to jupyter server and a prompt appears.

## Context

- Operating System and version: Windows 10
- Browser and version: MS Edge 85.0.564.51 (Chromium version)
- JupyterLab version: @jupyterlab/services@5.2.4

languy:
I'm submitting a patch for this: #9080 .

jcleyba:
@languy 
The code introduced in your PR is almost identical to the one removed in this commit: 

https://github.com/jupyterlab/jupyterlab/commit/e58660c11a6eed62fcb788b823dc9b36886bf21c#diff-326d12d41a9e5bac8f0f5572beb988f3

I don't understand why this was removed from all of the websocket connection handlers (kernel/default.ts for example) and this is causing problems to connect to websockets using the services library. 
Updating to 5.2.4 caused this websocket failure for me throwing a 403 error.

jasongrout:
For reference, the code was removed in https://github.com/jupyterlab/jupyterlab/pull/8835, fixing https://github.com/jupyterlab/jupyterlab/issues/8813

#8813 claims it is a security issue to have the token in the url. Can you comment on that?

[...]

PR #16022, opened 2024-03-19
Allow setting `ServerConnection.appendToken` via `PageConfig`

minrk:
allows explicit override in cases where default may not be desirable.

## References

#9070, #9898 and other related issues where the default logic to appendToken have been incorrect in some situations.

## Code changes

`ServerConnection.appendToken` will be derived from `PageConfig.appendToken` if defined.

It's unclear from the name alone that this is an option only affecting websockets. I considered calling the config `wsAppendToken`, but I thought it was probably better for the key to match.

## User-facing changes

Deployments may override `ServerConnection.appendToken`, if they wish.

None


krassowski (approved):
LGTM, thank you!

PR #16032, opened 2024-03-20
Preventing changing cell type when input is pending to avoid kernel deadlock

### LINKED ISSUE #15981: Accidentally converting a %debug cell to a raw cell blocks the kernel permanently
## Description
If a cell is in %debug mode, accidently pressing "R" can convert it into a raw cell which leaves the kernel in an unrecoverable state. 

## Reproduce
1. In cell 1, run the following: 
```python
def fn():
    raise ValueError()

fn()
```
2. In cell 2, run - 
```
%debug
```
3. Now convert the cell into a raw cell.

## Expected behavior
Converting such a cell to a raw cell shouldn't be allowed or should throw a warning. 

## Context
- JupyterLab version: 4.1.2

krassowski:
## References

Fixes #15981

## Code changes

- adds a `pendingInput` flag on `OutputArea`
- checks this flag before changing cell type
- if the flag is set, aborts and displays the same message as displayed in execution guard (adjusted to refer to cell type change rather than execution)

## User-facing changes

Accidentally typing `m` or `r` or digits from `1` to `6` in command mode on a cell with pending input does not cause a kernel deadlock anymore.

input-guard

The dialog:

Screenshot from 2024-03-20 13-15-34

None

### DIFF
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -420,7 +420,7 @@ export class SessionContext implements ISessionContext {
   }
 
   /**
-   * A flag indicating if the session has ending input, proxied from the kernel.
+   * A flag indicating if the session has pending input, proxied from the kernel.
    */
   get pendingInput(): boolean {
     return this._pendingInput;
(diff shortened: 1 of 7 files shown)


jtpio (approved):
Thanks!

PR #16009, opened 2024-03-18
Test on Python 3.12 on CI

jtpio:
## References

Python 3.12 was released in October 2023: https://www.python.org/downloads/release/python-3120/

## Code changes

- Update the Linux Python tests to test on the latest Python stable release.
- Add trove classifier

## User-facing changes

None

## Backwards-incompatible changes

None

PR #16041, opened 2024-03-22
Update mock package dependencies

jtpio:
## References

The mock packages used for testing were still depending on `jupyterlab~=3.0`:

https://github.com/jupyterlab/jupyterlab/blob/66ed61006257a2a316d3d10c54e6aebfac9947cc/jupyterlab/tests/mock_packages/test_no_hyphens/pyproject.toml#L5

As noticed in https://github.com/jupyterlab/jupyterlab/pull/16009 this was causing issues and pulling older dependencies for the `usage2` check on CI.

## Code changes

Update the build dependency for the test packages under `jupyterlab/tests/mock_packages`.

## User-facing changes

None

None


jtpio:
This should be enough for now for https://github.com/jupyterlab/jupyterlab/pull/16009 to pass.

As a follow-up we should check how we could better maintain these mock packages so they don't lag behind (maybe via a `buildutils` script). And update from `jupyter_packaging` to `hatch`.

krassowski (approved):
Thank you @jtpio!

Issue #16038, opened 2024-03-21
Forbid `any` for new code (unless absolutely needed)


krassowski:
Some packages had errors due to bad typing (notably shortcuts UI).

We currently have the `no-explicit-any` rule disabled:

https://github.com/jupyterlab/jupyterlab/blob/b318b57d3e5e15ade52c493d264c7818ffcd9381/.eslintrc.js#L63

What do folks think about turning it on?

Similarly, casting from `T | undefined` to `T`  with `as T` and using non-null assertions (!) has caused a lot of problems, especially  after CodeMirror 6 migration. I would like to turn these of in non-test files, but I know it's a bigger ask.

I do NOT propose banning it altogether (if warranted, an ignore comment can be added).

krassowski:
Another rule to consider is switching `@typescript-eslint/ban-types` from warning to error. I will squash a number of these in shortcuts package in an upcoming PR but there will be more.

```
  30:16  warning  Don't use `Function` as a type. The `Function` type accepts any function-like value.
It provides no type safety when calling the function, which can be a common source of bugs.
It also accepts things like class declarations, which will throw at runtime as they will not be called with `new`.
If you are expecting the function to accept certain arguments, you should explicitly define the function shape  @typescript-eslint/ban-types
```

jtpio:
> What do folks think about turning it on?

Sounds good. Hopefully it's not too much effort.

Maybe it will also help with https://github.com/jupyterlab/jupyterlab/issues/9392.

krassowski:
Yes, of note compare `main`:

```
✖ 56 problems (0 errors, 56 warnings)
```

with  https://github.com/jupyterlab/jupyterlab/pull/16043:

```
✖ 31 problems (0 errors, 31 warnings)
```

That's half down!

krassowski:
> Similarly, casting from `T | undefined` to `T` with `as T`

This can guarded and auto"fixed" (by rewriting to null assertions) via [`@typescript-eslint/non-nullable-type-assertion-style`](https://typescript-eslint.io/rules/non-nullable-type-assertion-style/) rule.

krassowski:
The autofixer for this rule generates more trouble than it is worth though because it also replaces things like `(e: Element) as HTMLElemet` with `(e: Element)!`. In these cases the case is partially warranted but should still be decoupled from non-null so it should be something like `(e: Element) as HTMLElemet | undefined`.

krassowski:
[...]

PR #14015, opened 2023-02-15
Add commands without shortcuts

fcollonval:
## References

Fix #13705

> as of now, it does not work as expected because a selector cannot be defined in the UI.

## Code changes

Add the commands without shortcuts defined

## User-facing changes

Can add a shortcut on commands without it

## Backwards-incompatible changes

None

Issue #13705, opened 2023-01-03
Semantic commands are missing in shortcut editor (regression)


krassowski:
## Description

Semantic commands (#11386, e.g. "Restart and run all cells", `runmenu:restart-and-run-all`) are missing in Setting Editor's Shortcuts section:

This might be related to #11898.

## Reproduce

![Screenshot from 2023-01-03 19-27-27](https://user-images.githubusercontent.com/5832902/210427560-136c3fec-a171-4ad0-891a-22d301208f11.png)

## Expected behavior

1. Shortcuts are searched by all: label, caption and command ID
2. Semantic commands have proper labels and captions

## Context

- JupyterLab version: 3.x and 4.0.0alpha

fcollonval:
Do you see the bug in 3.x too? If so, the bug may not be related to _Semantic_ command as they were not backported (due to the induced api change).

krassowski:
I see it on both 3.5 and master. Maybe was introduced by #10469? But then I don't know why it is also seen on 4.0.

JasonWeill:
I can also repro with `master`

fcollonval:
Looking into this, actually this concerns all commands that do not have an existing shortcuts. The shortcut editors does not get the list of commands but the `shortcuts` settings.

Fun fact the list of all commands is built to set the description of the JSON schema that is displayed in the JSON Settings editor.

fcollonval:
Adding the list of commands without shortcuts is [_easy_](https://github.com/jupyterlab/jupyterlab/pull/14015). But the trouble is that they don't have selectors. So the UI must be extended to allow for a selector to be defined; may be we could define a default selector (function of the command category (?)).

krassowski:
> may be we could define a default selector

or just go with "body" and advise users to change it as needed?

JasonWeill:
Per discussion at triage, this affects all commands that don't yet have shortcuts, not just semantic commands.

brichet:
We can think of several ways to set a new shortcut to a command (without default):
- display all the commands in the list (if requested only), allow creating a shortcut on any of them and allow setting the associated selector/args for the command. This is the current state of https://github.com/jupyterlab/jupyterlab/pull/18163
[...]

PR #16043, opened 2024-03-22
Refactor shortcuts UI to fix most of the known issues

### LINKED ISSUE #15038: Allow a way to disable default keybindings
### Problem

I am trying to get jupyterlab-vim into a usable state. And a major pain point is the "Enter Cell", I believe from a vim standpoint it should be `I` and not `Enter` as a shortcut. I'm able to bind `I` but there is the `I, I` binding conflict, which forces me to wait a full second before I can enter a cell which essentially renders `jupyterlab` useless to me. In the settings I remapped `interrupt-kernel` to something else but it keeps adding back `I, I` as a secondary key binding. I tried the "disabled" attribute but I don't want to disable `interrupt-kernel` just the **default** keybinding

### Proposed Solution

Do not insert back the default keybinding once a user has provided a custom binding for it

### LINKED ISSUE #15759: Shortcuts with same command but different arguments amiss in shortcuts UI
## Description

> Are the 20 different bindings all for one action intended? Or a bug? (If not a bug: why has this behaviour been chosen??). Ok, I see from the json settings and the PR discussion that the different bindings are for selecting different objects in the sidebar — but I’ll note that this isn’t reflected at all in the GUI settings editor. 

_Originally posted by @defjaf in https://github.com/jupyterlab/jupyterlab/issues/15744#issuecomment-1930741245_

## Reproduce

1. Open "Settings Editor"
2. Choose "Keyboard shortcuts"
3. See that all 20 keybindings for "Show Left Activity Bar" are shown as a single shortcut, even though each of these has a different set of arguments
4. See that you cannot modify these shortcuts (edits are not saved)

image

## Expected behavior

1. Shortcuts with different arguments are displayed separately
2. Shortcuts with different arguments can be modified independently
3. The arguments are displayed or even modifiable from shortcuts UI

## Context

- JupyterLab version: 4.1.0 but also affects other shortcuts with arguments in 4.0.x and possibly older

### LINKED ISSUE #15937: Delete shortcuts through the settings UI
### Problem

It would be helpful for users if they can use the settings editor to delete shortcuts too. Currently it only allows modifying but the original shortcut remains.

### Additional context
Jupyterlab v4.1.2

### LINKED ISSUE #16037: Memory leak in shortcuts UI: new commands created for every row
## Description

The current implementation of shortcuts UI creates a new command for every row and every action:

https://github.com/jupyterlab/jupyterlab/blob/b318b57d3e5e15ade52c493d264c7818ffcd9381/packages/shortcuts-extension/src/components/ShortcutItem.tsx#L121-L190

On the upside it only happens when actually using context menu.

## Expected behavior

Only one command is registered; schema is used for the menus population.

## Context

- JupyterLab version: 4.1.5

### LINKED ISSUE #16039: Sorting shortcuts in language other than English does not work
## Description

We should not mix up sort keys and translated labels, compase:

https://github.com/jupyterlab/jupyterlab/blob/b318b57d3e5e15ade52c493d264c7818ffcd9381/packages/shortcuts-extension/src/components/ShortcutTitleItem.tsx#L24

https://github.com/jupyterlab/jupyterlab/blob/b318b57d3e5e15ade52c493d264c7818ffcd9381/packages/shortcuts-extension/src/components/ShortcutInput.tsx#L64-L76

## Reproduce

error

## Expected behavior

Works

## Context

- JupyterLab version: 4.1.5

### LINKED ISSUE #16042: The logic for replacing conflicting shortcuts removes all keybindings from the other shortcut
## Description

If user wants to add a new keybinding for command A, say "Ctrl + J", but command B already uses "Ctrl + J" they will be shown an information about the conflict:

Screenshot from 2024-03-22 09-43-32

There are two problems:
- if "Ctrl + J" is the default shortcut, it will not be overwritten when clicking "Overwrite"
- if command B had multiple keybindings assigned, say "Ctrl + J" and "Ctrl +Shift + J" (and both are non-default), both will be removed

This is ok-ish:

https://github.com/jupyterlab/jupyterlab/blob/f00eef9bd9048990ab1c135b48920600d5d09f86/packages/shortcuts-extension/src/components/ShortcutInput.tsx#L136-L143

This is bad, see that `shortcutId` is not being used:

https://github.com/jupyterlab/jupyterlab/blob/f00eef9bd9048990ab1c135b48920600d5d09f86/packages/shortcuts-extension/src/components/ShortcutUI.tsx#L392-L399

Related to https://github.com/jupyterlab/jupyterlab/issues/15937.

## Expected behavior

Only the relevant keybinding is removed, and it is removed reliably regardless if it is default or not

## Context

- JupyterLab version: 4.1.5

krassowski:
## References

- fixes #16042
- fixes #16039
- fixes #16037
- fixes #15937
- fixes #15759
- fixes #15038
- addresses https://github.com/jupyterlab/jupyterlab/issues/16038#issuecomment-2015398728

## Code changes

Previously:
- shortcut targets and keybindings were used interchangeably, leading to errors in code
- shortcuts UI would aggregate keybindings by command ID and selector (ignoring arguments) assigning them to a poorly typed `keys` attribute of `ShortcutObject` (which was mutable data carrier object going against all good practices).
- the logic and presentation were mixed up badly in the react components code, including creation of temporary commands just to open a context menu

This PR solves numerous code-smells:
- defines a read-only `IShortcutTarget` as the thing by which the shortcuts are aggregated (which includes the command, selector and the arguments) and defines an `IKeybinding`, one or more of which can be assigned to `IShortcutTarget` (which should prevent any mixups in the future)
- organizational:
  - moves the logic for shortcut aggregation and conflict discovery out to a separate `ShortcutRegistry` class to enable unit-testing
  - hoists any remaining logic to the top-level `ShortcutUI` class now implementing `IShortcutUI` interface, so that there is no multiplicity of implementation and so that the logic can be reviewed and tested easily
- adds proper typings for the functions passed down in props to react components instead of error-prone `Function`
- implements schema-based context menu, allowing to delete the special logic which was creating commands for each row (sic!)
- removes reliance on most `id` attributes (which was invalid because there can be only one element with given `id` on the page)
- replaces the logic which was adding a row to the table on the `ShortcutList` level and then sorting the listing just to show the validation error ("overwrite"?) by rendering the second row within `ShortcutItem` when needed
- replaces comparisons of translated values to non-translated hard-coded strings (fixing the sorting issue and fixing unreported issue of reset button not showing in other languages)
- adds unit tests for `ShortcutRegistry` and the core logic of `IShortcutUI`

## User-facing changes

1. It is now possible to remove/delete the default shortcuts from the UI
   removeing-default
2. Sorting by columns works in all languages
3. More than two keybindings can be added and edited in the UI (previously UI was hard-coded to handle only "left" and "right" fields)
4. Overwriting a conflicting keybinding will now only overwrite that keybinding rather than all keybindings for given shortcut
5. When moving mouse over the shortcuts the content no longer jumps around (thanks to use of `opacity` over `display` for elements shown on hover, and adjusted `min-width`)

I think none because the react components were not exported from index. On the CSS side I tried to kept the changes to minimum required so themes should be good to.


krassowski:
Currently commands with all keybindings disabled do not render at all. Making commands without existing shortcuts show up is tracked in https://github.com/jupyterlab/jupyterlab/issues/13705 and https://github.com/jupyterlab/jupyterlab/pull/14015 so I will not be tackling it in this PR.

packages/shortcuts-extension/src/components/ShortcutUI.tsx
```diff
@@ -195,313 +189,268 @@ function matchItems(items: ShortcutRegistry, query: string): ISearchResult[] {
...
+  };
+
+  /**
+   * Replace the given keybinding with a new keybinding as defined by given keys.
+   *
+   * If keybinding is not given a new keybinding will be created.
+   * If keys are empty (or include a single empty string) the keybinding will be deleted.
+   * If keybinding is a default keybinding and the provided keys differ from default, it will be disabled.
+   */
+  private async _setKeybinding(
```
krassowski, line 349: Thinking whether this should be moved to `ShortcutRegistry` to make it even easier to unit-test. An argument against is that `ShortcutRegistry` is a function of command registry for now, whereas this function depends on `SettingsRegistry`. Because it is private method, it could be extracted out in a follow-up PR withouch changing the API surface.

krassowski: Maybe it is better placed here because it makes it easy to understand the logic of the public functions when it is all in a single file. In either case do not intend to make more changes in this PR.

fcollonval (changes requested):
Thanks a lot @krassowski 

This is some huge piece of work. I only scan through the code quickly but did not see anything outstanding.

When testing, I hit the following issue: if you press an existing shortcut when editing the shortcut, the shortcut is actually executed. The most pathological case is setting a shortcut matching the command close tab or activate next/previous tab. I guess we should deactivate the shortcut handling when editing a shortcut.

Regarding the huge code change, I was wondering if it makes sense to bump the major version of that package. I know it will be a departure of what we currently have. But at the same time it is the goal we are targeting for future major version of JupyterLab - aka only bumping major version of JS packages that need it.

fcollonval (approved):
Huge thanks for this one @krassowski

jtpio:
> I was wondering if it makes sense to bump the major version of that package. I know it will be a departure of what we currently have. But at the same time it is the goal we are targeting for future major version of JupyterLab - aka only bumping major version of JS packages that need it.

Yes it's fine to bump individual packages if needed.

Just to double check, the proposal here is to bump `@jupyterlab/shortcuts-extension` to version `5.0.0-alpha.2`?

PR #15855, opened 2024-02-22
400% zoom low vision

g547315:
## References
#14626
#14766

#10004 This PR covers Full low Vision Support (400% Zoom) (Mobile/Tablet-Responsive Support).
It resolves most of the failing screenshots that where present in #14766. The approach used was to isolate each code change in the previous PR to see what/ If the code changes are breaking screenshots.

Isolated Code changes and their effects on Screenshots 

- [padding changes to menu bar & padding changes to side panel](https://github.com/g547315/jupyterlab/pull/3)
- [margin changes to .jp-FileBrowser-toolbar.jp-Toolbar](https://github.com/jupyterlab/jupyterlab/commit/1a0b5918e31dd2b2f2564ca4a05a441261fc21b8)
- [padding chnages to .jp-DirListing-headerItem](https://github.com/g547315/jupyterlab/pull/1/files?file-filters%5B%5D=.png&show-viewed-files=true)
- [margin changes for .jp-BreadCrumbs](https://github.com/g547315/jupyterlab/pull/2/files?file-filters%5B%5D=.css&file-filters%5B%5D=.png&show-viewed-files=true)

@g547315

## Code changes
Changed padding and spacing to use min CSS function alongside vh (viewport-height) and vw(viewport-width) to reduce proportional spacing at high zoom and increase useful screen real estate usage.

## User-facing changes
Currently with the 400% zoom at 1280px the help main menu bar does not appear and within the file-browser tab bar only half of one 'dir-listing' shows up.

The changes I have made now allows for the help dropdown to be shown in the main menu. I found that when trying to check the changes the sidebar 'Extension Manger' disappears from view, so this was also changes so that users that needs this implemented would be able to have access to it without zooming back out.
Furthermore, within the file-browser tab, users would now be able to see least 2 notebook directory listed below than before.

Issue #13148, opened 2022-09-28
Collapse main menu options in a hamburger menu in small screens and high zoom


steff456:
### Problem

Currently, when Jupyterlab is opened in a small screen or the user increases the zoom the main menu options start to hide, making  the right most menus not usable,

![image](https://user-images.githubusercontent.com/20992645/192878361-8eef20ae-eba1-4cf8-8dd2-a3a91f81c7aa.png)

⬆️ `Help` menu is not visible and `Settings` menu is cropped.

### Proposed Solution

The proposed solution is to add a hamburger menu in the right side to store the menus that are not visible in such way that the user will be able to use them.

### Additional context

I think that this change will probably have to be done in Lumino, but any feedback in the proposed solution will be great!

fcollonval:
I'm wondering if instead of displaying an hamburger, we should display an ellipsis that display the hidden entries. Something like this (taken from VS Code):

![image](https://user-images.githubusercontent.com/8435071/192982050-11efb541-49ab-4ba6-9183-5a4cbf85ce5b.png)

This comes from the following considerations:

- If the width is slightly too small to fit the menu. We will end up with a large unused empty white space.
- A single hamburger icon will feel weird in simple mode and in notebook v7:

| simple mode | notebook v7 |
| --- | --- |
| ![](https://raw.githubusercontent.com/jupyterlab/jupyterlab/master/galata/test/jupyterlab/general.test.ts-snapshots/simple-mode-jupyterlab-linux.png) | ![](https://raw.githubusercontent.com/jupyter/notebook/main/ui-tests/test/general.spec.ts-snapshots/notebook-chromium-linux.png) |

krassowski:
Reopening as it does not seem to work yet, maybe some follow up action is needed.

krassowski:
> I spent a few hours tonight digging into the overflow menu ("hamburger"). It's actually [turned on by default in Lumino](https://github.com/jupyterlab/lumino/blob/7f4bd513004d2aea3e96623210d015a834a4fca7/packages/widgets/src/menubar.ts#L53) so there's nothing to do to turn it on in the current build of JupyterLab.
> 
> So then I did some debugging. I discovered that the overflow implementation caches the size of each item in the menu bar. But it does this when the app loads and when these items have width 0. So the logic that shunts menu items into the overflow if the menu bar is too big for the available screen width never kicks in because it thinks the whole menu bar has zero width. 
> 
[...]

PR #16044, opened 2024-03-23
Decrease top menu padding for high zoom / narrow screen

gabalafou:
### References

PR #15855.

This is the first of a series of PRs breaking up #15855 into individual CSS changes, with the hope that this will make it easier to tackle the issues that we've been having with Playwright Galata snapshot testing.

### Code changes

Updates the padding rule for the `.lm-MenuBar-item` CSS selector so that menu bar items will all fit within narrow or zoomed-in screens.

### User-facing changes

See above.

### Backwards-incompatible changes

n/a

### How to test

Load JupyterLab. Drag window wide, drag window narrow. Observe how it affects padding in the menu bar. At 480px, padding should start to decrease. At 320px, the top menu bar items (File, Edit, View, Run, Kernel, Tabs, Settings, Help) will be scrunched together but should all be visible, as the following screenshot shows:

Alternatively, set your browser window to 1280px, then zoom in to 400%. Again, the menu items will be tightly spaced but should fit within the width of your browser window.

### DIFF
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -31,7 +31,12 @@
 }
 
 .lm-MenuBar-item {
-  padding: 0 8px;
+  /* TODO: remove ignore directive when fixed: https://github.com/csstree/csstree/issues/164 */
+  /* stylelint-disable-next-line csstree/validator */
+  padding: 0 min(calc(4vw - 11.2px), 8px); /* The use of min(), calc() and vw units
+    has the effect that as the screen drops below 480px, horizontal padding will
+    steadily decrease so that the standard menu items fit in a 320px wide screen
+    (equivalent to zooming to 400% from 1280px) */
   border-left: var(--jp-border-width) solid transparent;
   border-right: var(--jp-border-width) solid transparent;
   border-top: var(--jp-border-width) solid transparent;


packages/application/style/menus.css
```diff
@@ -31,7 +31,12 @@
 }
 
 .lm-MenuBar-item {
-  padding: 0 8px;
+  /* TODO: remove ignore directive when fixed: https://github.com/csstree/csstree/issues/164 */
+  /* stylelint-disable-next-line csstree/validator */
+  padding: 0 min(calc(4vw - 11.2px), 8px); /* The use of min(), calc() and vw units
+    has the effect that as the screen drops below 480px, horizontal padding will
```
gabalafou, line 37: Note: 480px was chosen through trial and error. At around 420px with 8px padding between menu items, the menu bar starts to be too wide for the screen width. 480px gives a little wiggle room.

This gave two equations to solve (units in pixels):

```
(480px)a + b = 8
(320px)a + b = 1.5
```

Because at 320px, the padding needs to be set to 1.5px for the items to fit. Solving the equations gives:

```
a ~= 0.04 (equivalent to 4vw)
b ~= -11.2px
```

krassowski (changes requested):
> I hesitated to review this PR because I am moderately against this change, but since no one else had left a review, here we go:

Among all the changes in https://github.com/jupyterlab/jupyterlab/pull/15855, this one appears to me to be the most controversial, because it fixes one accessibility issue (visibility of options at 400% zoom) by introducing another accessibility issue (too small [tap targets](https://web.dev/articles/accessible-tap-targets#:~:text=A%20minimum%20recommended%20touch%20target,target%20size%20up%20to%2048px.)).

I think this issue would be better solved by making use of the hamburger menu. While it is a tricky thing to get right, there was already a lot of effort put into it:
- https://github.com/jupyterlab/jupyterlab/issues/13148
- https://github.com/jupyterlab/lumino/pull/489

krassowski:
Of note WCAG 2.2 requires:
- [for AA](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) 24 CSS pixels
- [for AAA](https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced.html) 44 CSS pixels

This PR makes all the menu items except for "Kernel" and "Settings" smaller than 44 CSS pixels in width:
   violates-wcag

However, if there are more voices in favour of this PR (please comment) I would not want to block this getting merged.

krassowski:
Another thought is that it does not really solve the problem when:
- extensions add menu entries; for example in a deployment which adds "Services" in between "Settings" and "Help", then "Help" will still be inaccessible
- any language other than English in which translated strings take up more space than the English ones (which is many languages)

gabalafou:
Responding to the review of [4.3.0 milestones] in today's Jupyter Frontends call...

While I think making the UI more dense at higher zoom is a desirable goal, there are too many issues with the bit of CSS in this PR, as @krassowski pointed out. 

It is not feasible for me to address this anytime soon, so I'm going to go ahead and close this.

Issue #16057, opened 2024-03-25
Examples CI job is failing


krassowski:
The `examples` job is failing:

```
log >> JupyterLab extension middle is activated!
log >> CommandRegistry
error >> Plugin '@jupyterlab/apputils-extension:workspaces' failed to activate.
error >> TypeError: No provider for: @jupyterlab/workspaces:IWorkspaceCommands.
    at JupyterLab.resolveRequiredService (http://localhost:8888/foo/static/__main__/jlab_core.bundle.js:4526:19)
    at http://localhost:8888/foo/static/__main__/jlab_core.bundle.js:4439:56
    at Array.map (<anonymous>)
    at JupyterLab.activatePlugin (http://localhost:8888/foo/static/__main__/jlab_core.bundle.js:4439:42)
    at http://localhost:8888/foo/static/__main__/jlab_core.bundle.js:4608:25
    at Array.map (<anonymous>)
    at JupyterLab.start (http://localhost:8888/foo/static/__main__/jlab_core.bundle.js:4607:35)
    at main (http://localhost:8888/foo/static/__main__/build_index_js.bundle.js:544:7)
warning >> Language pack 'en_US' not installed!
debug >> Starting application in workspace: "default"
debug >> Example started!
F

  1) example.spec.ts:12:5 › should load the example ────────────────────────────────────────────────

    Error: expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 2

      58 |   }
      59 |
    > 60 |   expect(errorLogs).toEqual(0);
```

which 

because the workspaces package was not added to `package.json` in examples.

PR #16058, opened 2024-03-25
Add missing `workspaces-extension` package to examples, add migration note

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16057

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/15946

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -19,6 +19,14 @@ API updates
 - ``CodeEditor.getCoordinateForPosition`` return type was corrected to clarify that it can return
   ``null``; previously ``null`` could be returned despite the return type indicating it would always
   return a non-null ``ICoordinate`` value.
+- The ``@jupyterlab/apputils-extension:workspaces`` plugin  now depends on the new
+  ``@jupyterlab/workspaces-extension`` package which defines the workspace-related commands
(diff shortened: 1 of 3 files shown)


packages/apputils-extension/src/workspacesplugin.ts
```diff
@@ -29,20 +31,36 @@ export const workspacesPlugin: JupyterFrontEndPlugin<void> = {
...
+  optional: [IRouter, IWorkspaceCommands],
   activate: (
     app: JupyterFrontEnd,
     state: IStateDB,
-    commands: IWorkspaceCommands,
-    translator: ITranslator
+    translator: ITranslator,
+    paths: JupyterFrontEnd.IPaths,
+    router: IRouter | null,
+    commands: IWorkspaceCommands | null
```
jtpio, line 42: Maybe we can take the opportunity to name this parameter `workspaceCommands` so it's more explicit? To avoid some potential shadowing if the plugin code ends up making use of the default commands (for example via `const { commands } = app` at some point?

krassowski: `const { commands } = app` would give you type error, but yes, I changed it for clarity in 7304c31d4b459a64b5326c6f14af219c95aca5a8.

jtpio (approved):
Thanks!

PR #16068, opened 2024-03-28
Fix scrolling to output area inputs on caret movement

### LINKED ISSUE #16007: In windowed notebook active cell is not scrolled into view on switching cell type/typing in output area
## Description

When working on https://github.com/jupyterlab/jupyterlab/pull/16006 I noticed that in `full` windowed mode the active cell that is out of viewport will not be scrolled back into the viewport upon:
- (a) cell actions in command mode
- (b) typing in output stdin box

## Reproduce

1. Open a notebook, create ~100 cells
2. Switch windowing mode to `full` in settings

Scenario (a):
3. Enter command mode by pressing Esc
4. Scroll to the bottom of the notebook
5. Press "m" to switch cell to Markdown
6. See that notebook did not scroll back to the active cell despite it changing cell type

Scenario (b):
3. In first cell type `input()` and execute, this will show a stdin input box
4. With input box focused scroll to the bottom of the notebook
5. Type "a"
6. See that notebook did not scroll back to the active cell despite the stdin receiving "a" key

## Expected behavior

Notebook scrolls back to active cell on actions which require user attention to guide user in what is happening, as in non-windowed mode.

## Context

- JupyterLab version: 4.1.5

krassowski:
## References

- Fixes #16007
- Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16006

## Code changes

Cells will emit `scrollRequested` when there is a caret movement induced by keyboard event (`keydown`) in any of the inputs/textareas/etc of the cell's output area.

## User-facing changes

In windowed notebook typing in a character, or moving cursor with arrow keys, Home/End, Pg Down/up causes scrolling to the active input element, the same way as in non-windowed notebook.

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -1085,7 +1085,8 @@ export class CodeCell extends Cell<ICodeCellModel> {
       promptOverlay: true,
       inputHistoryScope: options.inputHistoryScope
     }));
-    this._observeCaretMovementInOuput(output);
+    output.node.addEventListener('keydown', this._detectCaretMovementInOuput);
+
     output.addClass(CELL_OUTPUT_AREA_CLASS);
     output.toggleScrolling.connect(() => {
       this.outputsScrolled = !this.outputsScrolled;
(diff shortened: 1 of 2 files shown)


packages/cells/src/widget.ts
```diff
@@ -1098,6 +1101,70 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+  /**
+   * Observe the movement of the caret in the output area.
+   *
+   * Emits scroll request if the caret moved.
+   */
+  private _observeCaretMovementInOuput(output: OutputArea) {
+    let lastOnCaretMovedHandler: () => void;
+    let lastTarget: HTMLElement | null;
+
+    output.node.addEventListener('keydown', (e: KeyboardEvent) => {
```
fcollonval, line 1113: Should this listener be removed when disposing the cell?

packages/cells/src/widget.ts
```diff
@@ -1098,6 +1101,70 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+        );
+      }
+
+      const onCaretMoved = () => {
+        this._scrollRequested.emit({
+          scrollWithinCell: ({ scroller }) => {
+            ElementExt.scrollIntoViewIfNeeded(scroller, target);
+          },
+          defaultPrevented
+        });
```
fcollonval, line 1146: Shouldn't the default scrollRequest be emitted when this gets disconnected?

fcollonval (commented):
I got some questions about this one @krassowski

packages/cells/src/widget.ts
krassowski: I do not understand, can you clarify?

fcollonval: Nevermind I read the code too quickly and I had the impression that the callback `scrollWithinCell` was stored in the notebook. But it is not.

Sorry for the noise.

packages/cells/src/widget.ts
krassowski: Done in e458d5586af80179995a3e4e4e386317f52e4846.

fcollonval (approved):
Thanks @krassowski

PR #16069, opened 2024-03-28
Make visual regression tests faster

### LINKED ISSUE #16067: Typing in cells takes a lot of time on visual regression tests
Reducing the delay for typing can speed up the tests, see:

we could cut 15s out of 23s here:

[c84eecd8bbcdbc0a1855eb89dae3fc12e93823d7.webm](https://github.com/jupyterlab/jupyterlab/assets/5832902/14f9757d-9b35-46fc-9ca4-f8e1f00c85a9)

and 27s out of 36s here:

[ab416e61e63473e7f0dc001735452e47d425fd07.webm](https://github.com/jupyterlab/jupyterlab/assets/5832902/b5e1cc81-723c-44f0-b253-d3efa592d26c)

and so on.

krassowski:
## References

Attempts to fix #16067

## Code changes

- [x] Upload notebooks/files when creating them for each test takes longer than 250ms
- [x] Remove some explicit waits
- [x] Use `beforeAll` over `beforeEach` where appropriate

maybe the setup slowness is because of dev mode not compressing assets? unlikely but worth checking.

## User-facing changes

None

None

### DIFF
--- a/galata/src/helpers/notebook.ts
+++ b/galata/src/helpers/notebook.ts
@@ -1255,14 +1255,37 @@ export class NotebookHelper {
 
     await this._setCellMode(cell, 'Edit');
     await cell.getByRole('textbox').press('Control+A');
-    await cell
-      .getByRole('textbox')
-      .type(source, { delay: cellType === 'code' ? 100 : 0 });
+    await cell.getByRole('textbox').type(source, { delay: 0 });
     await this._setCellMode(cell, 'Command');
 
-    // give CodeMirror time to style properly
     if (cellType === 'code') {
-      await this.page.waitForTimeout(500);
+      // Wait until the CodeMirror highlighting is stable
+      // over 10 consecutive animation frames.
+      await cell.evaluate((cell: HTMLElement) => {
+        let _resolve: () => void;
+        const promise = new Promise<void>(resolve => {
+          _resolve = resolve;
+        });
+        let framesWithoutChange = 0;
+        let content = cell.querySelector('.cm-content')!.innerHTML;
+        const waitUntilNextFrame = () => {
+          window.requestAnimationFrame(() => {
+            const newContent = cell.querySelector('.cm-content')!.innerHTML;
+            if (content === newContent) {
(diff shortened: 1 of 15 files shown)


krassowski:
Comparing the test step itself (excluding setup/teardown) with runs in #16070 and after merge of #16043

| PR | Visual regression test | Documentation test |
|--|--|--|
| #16043 | 30m 39s | 5m 29s |
| #16070 | 31m 48s | 5m 27s |
| this one | 27m 37s | 4m 40s |

To be honest, I hoped for more savings, but it's still an improvement.

galata/src/helpers/notebook.ts
```diff
@@ -1255,14 +1255,37 @@ export class NotebookHelper {
...
-      .type(source, { delay: cellType === 'code' ? 100 : 0 });
+    await cell.getByRole('textbox').type(source, { delay: 0 });
     await this._setCellMode(cell, 'Command');
 
-    // give CodeMirror time to style properly
     if (cellType === 'code') {
-      await this.page.waitForTimeout(500);
+      // Wait until the CodeMirror highlighting is stable
+      // over 10 consecutive animation frames.
+      await cell.evaluate((cell: HTMLElement) => {
```
fcollonval, line 1264: Do you think `toHaveScreenshot` will work in this case - trying to simplify the code if possible?

Ref: https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-screenshot-2

fcollonval (approved):
Thanks @krassowski I have a suggestion but you can merge without it.

galata/src/helpers/notebook.ts
krassowski: I think the problem with most playwright's "is it stable" logic is that these only look at 2 consecutive frames:

> toHaveScreenshot(): This function will wait until two consecutive locator screenshots yield the same result, and then compare the last screenshot with the expectation.

but CodeMirror takes rather long pauses when rendering so I am not actually sure if even 10 animation frames is enough for some cases.

PR #16072, opened 2024-03-29
Tackle flaky readonly notification and workspaces context menu snapshots

krassowski:
## References

Addresses some cases from https://github.com/jupyterlab/jupyterlab/issues/14947

## Code changes

- Replace explicit wait with an a wait for toast animation to complete
- Mock sessions late in the test to avoid kernels showing up in the workspaces snapshot

## User-facing changes

None

None


jtpio (approved):
Thanks!

Issue #16073, opened 2024-03-29
Shortcuts sort does not use locale-specific sort order


krassowski:
We should also be sorting using a sort algorithm suitable to the language in question.

_Originally posted by @JasonWeill in https://github.com/jupyterlab/jupyterlab/issues/16039#issuecomment-2020901578_

In particular:

https://github.com/jupyterlab/jupyterlab/blob/5af2cc9346a49bc1354448a547413937ab754b56/packages/shortcuts-extension/src/components/ShortcutUI.tsx#L426-L450

does not use [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) for string values.

krassowski:
As bonus point, a test for sorting could be added in:

https://github.com/jupyterlab/jupyterlab/blob/5af2cc9346a49bc1354448a547413937ab754b56/packages/shortcuts-extension/test/components/ShortcutUI.spec.ts#L47

but it would not be required of a first-time contributor.

UntitledError-09:
@krassowski, I'd like to pick up this as my first issue. Can you assign this issue to me?

krassowski:
Fixed by https://github.com/jupyterlab/jupyterlab/pull/16077

PR #16077, opened 2024-04-01
Use locale-based sorting in the table of shortcuts

UntitledError-09:
## References

* Fixes issue #16073 for locale-specific sorting of keyboard shortcuts using any criteria. 

## Code changes

* Fix uses `String.prototype.localeCompare` to compare the `IShortcutTarget`s  in the `Array.sort` comparator. 
* As before, still compares either the `IShortcutTarget[sortCriteria]` value returned from `getValue` function, or else uses `IShortcutTarget.label` by default for the comparison.

## User-facing changes

* None.

* None.

### DIFF
--- a/packages/shortcuts-extension/src/components/ShortcutUI.tsx
+++ b/packages/shortcuts-extension/src/components/ShortcutUI.tsx
@@ -438,14 +438,13 @@ export class ShortcutUI
     shortcuts.sort((a: IShortcutTarget, b: IShortcutTarget) => {
       const compareA: string = getValue(a);
       const compareB: string = getValue(b);
-      if (compareA < compareB) {
-        return -1;
-      } else if (compareA > compareB) {
-        return 1;
+      const compareResult = compareA.localeCompare(compareB);
+      if (compareResult) {
+        return compareResult;
       } else {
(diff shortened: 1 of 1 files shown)


UntitledError-09:
Hello, 

Regarding the 3 failing checks:
1. I am unsure about how to add a label, and if I can at all since I do not see an option to change the labels.
2. 2 UI Tests failing checks seem to be related to the *Galata* Project, which does not seem to be related to what I was working on.

Please let me know if I missed something.

krassowski:
Thank you!

> I am unsure about how to add a label, and if I can at all since I do not see an option to change the labels.

I've added the label. In the future you can ask the `meeseeksdev` bot, see: https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#tagging-issues-with-labels

> 2 UI Tests failing checks seem to be related to the Galata Project, which does not seem to be related to what I was working on.

Correct, these tests are randomly failing and there is a n ongoing work to fix them (https://github.com/jupyterlab/jupyterlab/issues/14947). I will restart them before we merge a couple of times to ensure there are no regressions.

krassowski (approved):
Thank you @UntitledError-09!

Issue #15525, opened 2023-12-13
"Comment" shortcut not working since version 4.0.x


diyoyo:
## Description

The shortcut keys don't work anymore after upgrading from 3.6.5 to 4.0. I tested 4.0.7 and 4.0.9

## Reproduce

- MacBook Pro Keyboard, French Layout (AZERTY)
- Open Notebook .ipynb
- Select part of code in code cell
- Hit "fn" + "Ctrl^" + "+"
- Nothing is commented out.

## Expected behavior

Code should be commented 

## Context

Client:
- Operating System and version: MacOS latest
- Browser and version: Brave  latest (with and without shield)

Server:
- Debian bookwork (RaspberryPi OS lite, latest)
- Docker latest
- Jupyter Minimal Notebook image : quay.io/jupyter/minimal-notebook
- JupyterLab version: 4.0.7 and 4.0.9
- Jupyter settings mounted as external mount for persistence reasons. (maybe a hint here: incompatibility between versions?)

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

diyoyo:
Following up on the investigation:

- Using "Cmd" + "Shift" + "/" working **when using Firefox**
- But still doesn't work using Brave.

diyoyo:
Ok @JasonWeill , here I continue the conversation. So the bug is still present in Chromium based browsers, after upgrading to 4.1.3.

I'm going to open a feature request to have at least the following workaround: a context menu item that can comment the current selection.

Issue #14548, opened 2023-05-15
CTRL+D delete string shortcut removed in jupyter 4.0.0


mahashiah:
Deleting a line (string) has become an impossible task on my keyboard since developers removed the Ctrl+D shortcut in version 4.0. I must commend the developers for making it exceptionally challenging to utilize Jupyter on a 60% mechanical keyboard lacking a dedicated delete key. It seems to be an erroneous assumption that every keyboard possesses such a key, which is evidently not the case. Moreover, you have imposed the compulsory assignment of delete as a vital key while assigning Ctrl+D to a new feature that is essentially futile. It would have been far more practical to allocate Ctrl+F for word searches, as you did for Ctrl+D. How can I delete a row in Jupyter 4.0 when the Ctrl+D shortcut is invariably linked to an unhelpful search function? To be frank, I had a favorable opinion of Jupyter, but now I find it necessary to downgrade and revert to a previous version, abandoning 4.0 entirely.

JasonWeill:
@mahashiah Thank you for your suggestion! We apologize for the regression and we welcome contributions to reinstate it.

krassowski:
A few clarifications to dissect the issue:
- there was no search keybinding for <kbd>Ctrl</kbd> + <kbd>D</kbd> added in by JupyterLab team in JupyterLab 4.0. The search widget keybindings are defined here:
       https://github.com/jupyterlab/jupyterlab/blob/94a27fb987afddaf75d65265d825f24774247598/packages/documentsearch-extension/schema/plugin.json#L35-L56
- the change is a result of migration from CodeMirror5 to CodeMirror6 which changed the default keybindings. The new keybinding is <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>K</kbd> (defined in [`codemirror/commands` package, here](https://github.com/codemirror/commands/blob/f17b5b889801959cdc01435ba2b5f23c68ecc841/src/commands.ts#L895)) but it conflicts with the JupyterLab shortcut to open Table of Contents sidebar
  - CodeMirror does define [`Mod-d` for select next occurrence]( https://github.com/codemirror/search/blob/3c1346e213a9635f56f92b23066ea0638319de25/src/search.ts#L583) command (which we do not need in most editors as described later)

How we can address this?
- the default keymap of CodeMirror in JupyterLab can be amended to reduce friction for users migrating from 3.x; this likely should happen either;
[...]

Issue #13128, opened 2022-09-26
How to comment multiple lines on German (QWERTZ) keyboard


qebrahym:
Hello everyone, 
i have tried and tested almost every possible method and solution from internet and suggested by different people here but still got no success. 

Can someone pls help me out **"How can i COMMENT OUT MULTIPLE LINES in JUPYTER NOTEBOOK" on Mac** with German Keyboard layout. 

Your +ve and kind response will be highly appreciated. 
Thanks

firai:
Have you tried `Cmd + ß` (from https://github.com/jupyterlab/jupyterlab/issues/4778#issuecomment-944970341)?

qebrahym:
> Have you tried `Cmd + ß` (from [#4778 (comment)](https://github.com/jupyterlab/jupyterlab/issues/4778#issuecomment-944970341))?

Yes I tried, but it doesn’t work.

jweill-aws:
The key combination using a US English keyboard is <kbd>CMD+/</kbd> but it looks like <kbd>/</kbd> (forward slash) is a shifted character on a QWERTZ keyboard.

AlexanderMelde:
As a general suggestion, the <kbd>/</kbd> key on the numpad usually works, but I'm not sure if that helps you in this particular case.

qebrahym:
> The key combination using a US English keyboard is CMD+/ but it looks like / (forward slash) is a shifted character on a QWERTZ keyboard.

Yes indeed that‘s the issue. the Slash / button on QWERTY Keyboard doesn‘t work.

qebrahym:
> As a general suggestion, the / key on the numpad usually works, but I'm not sure if that helps you in this particular case.

Actually i want to use Slash / button on QWERTY Keyboard to comment out multiple lines. Because the Keyboard Layout changes and it get difficult to type quickly on US keyboard as all the buttons are different and more importantly it affects the typing speed and impact the workflow.

MatusGasparik:
This is also my pain point with jupyterlab. I still haven't figured out how to comment out stuff in the notebook cells, other than manually. There is the multi-line cursor (with Alt-key), but I'd rather not use mouse. No even the `jupyterlab-vim` extension is of any help (unless it is possible to add custom plugins, like `vim-commentary`, which I don't know if its possible...).

Any working solutions will be very appreciated!

v18nguye:
Let's try with 
`Strg  + divide button`

[...]

Issue #4778, opened 2018-06-26
Comment code block command


cheTesta:
It appears to me that, althought previously available in Jupyter Notebook, the command to "**comment code**" (also toggled by keyboard shortcut `Crtl + /`) is currently not listed in the Commands vertical pane.
Command is still available to use, normally.

Personal opinion but, it would be better to have it shown among other possible commands.

Also, since the shortcut is not handy on non US keyboard layouts, it would be cool to have it listed **and editable** in Keyboard Shortcut menù inside Advanded Settings Editor.

saulshanabrook:
This seems like a sensible request.

It also seems like other text editor commands, like indent, aren't available in the command prompt for the notebook.

adriantre:
I am having a non-US keyboard layout, which makes it impossible to toggle block comment.

TheIdealis:
@adriantre Just in case you have a number pad and didn't consider it, like I did: Until the feature is added, you can also use the division on the number pad.

Asutu:
Also struggled a bit with this, and thought this might be useful to others.

In a MacBook Pro with a Portuguese keyboard layout (that uses `Cmd + Shift + 7` to toggle comments on Sublime Text 3, for Python), by trial and error figured that `Ctrl + -` works in JupyterLab.

jasongrout:
Thanks for posting your findings here!

dschick:
would be great to change the shortcut.
see also this issue: https://github.com/jupyterlab/jupyterlab-shortcutui/issues/34

realmariano:
I have a latam keyboard (no number section since it is a notebook), couldn't find any way to toggle comments yet. 
The <kbd>CTRL</kbd>+<kbd>-</kbd> mentioned before does not work in my case (I am using firefox, so when I try it it juts decrease the zoom).
It would be great to be able to set a specific short cut for this.

flaxdev:
it would be extremely useful (for non-US) to have the possibility to change the shortcut keys for commenting/uncommenting code lines

neodelphis:
all the same down here. On a french mac keyboard with no numeric keypad I'm unable to use any of the comment code block command listed previously!

jasongrout:
Cross-posting from #7579:

FYI, those specific keys are set here:

https://github.com/jupyterlab/jupyterlab/blob/4fe4dcfe5c9dc329bed2dcf2602f569ddef8a8a0/packages/codemirror/src/factory.ts#L24-L25

[...]

PR #16078, opened 2024-04-01
Upgrade lumino to v2024.3.25, improve CodeMirror shortcuts handling

krassowski:
## References

- Fixes regressions caused by https://github.com/jupyterlab/lumino/pull/662 due to our code calling `preventDefault` to prevent default CodeMirror actions - instead we filter these out in the first place.
- Adds commands allowing to customise shortcuts for:
  - delete line (default <kbd>Ctrl</kbd> + <kbd>D</kbd> or <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>K</kbd>) and extend selection (default <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>D</kbd>):
     - this is a change in behaviour because:
        - issue [#14548](https://github.com/jupyterlab/jupyterlab/issues/14548) asking for Ctrl + D to be used for delete line (like in JupyterLab 3.x) had 17 :+1:
        - issue [#15432](https://github.com/jupyterlab/jupyterlab/issues/15432) asking for Ctrl + D to be used for extending selection had only one reporter
     - fixes https://github.com/jupyterlab/jupyterlab/issues/16096; the new behaviour is to delete the line but the shortcut can now be disabled from settings if a user prefers it to always open the ToC
     - fixes https://github.com/jupyterlab/jupyterlab/issues/14548
  -  toggle comment (default <kbd>Ctrl</kbd> + <kbd>/</kbd>; this can be used to comment multiple lines too) and toggle block comment (default <kbd>Alt</kbd> + <kbd>A</kbd>; note this does not work in Python because CodeMirror does not define block comment syntax for Python)
     - fixes https://github.com/jupyterlab/jupyterlab/issues/4778  (which has 42 :+1:)
     - closes https://github.com/jupyterlab/jupyterlab/issues/15525
     - closes https://github.com/jupyterlab/jupyterlab/issues/13128
    

## Code changes

None

## User-facing changes

image

None


krassowski:
Yup, the same tests are failing here:

```
Notice:   4 failed
    [galata] › test/galata/notebook.spec.ts:22:7 › Notebook Tests › Create Markdown cell ───────────
    [galata] › test/galata/notebook.spec.ts:40:7 › Notebook Tests › Create Raw cell ────────────────
    [galata] › test/galata/notebook.spec.ts:59:7 › Notebook Tests › Create Code cell ───────────────
    [jupyterlab] › test/jupyterlab/cells.test.ts:41:7 › Run Cells With Keyboard › Run code cell with Ctrl + Enter 
```

@fcollonval do you think this could be related to https://github.com/jupyterlab/lumino/pull/662?

krassowski:
So this indeed is caused by https://github.com/jupyterlab/lumino/pull/662.

1. We use `preventDefault` in JupyterLab to modify CodeMirror behaviour here:

https://github.com/jupyterlab/jupyterlab/blob/41be86d5142a793ab9cb8b9eb99f92a33c8dc441/packages/codemirror/src/commands.ts#L77-L85

2. Even if we disable that login (which leads to inserting a new line when it should run a cell), CodeMirror still runs `newlineAndIndent` on Ctrl + Enter, which prevents default.

krassowski:
I think we could filter the `defaultKeymap`, here:

https://github.com/jupyterlab/jupyterlab/blob/41be86d5142a793ab9cb8b9eb99f92a33c8dc441/packages/codemirror/src/extension.ts#L701-L714

krassowski:
This also means that Ctrl+Shift+K no longer opens Table Of Contents when in edit mode (as it not only deletes the current line; previously it did both: https://github.com/jupyterlab/jupyterlab/issues/16096).

Given the issues:
- https://github.com/jupyterlab/jupyterlab/issues/16096
- https://github.com/jupyterlab/jupyterlab/issues/14548 (this one has 17 upvotes)

I propose that we disable these default keybindings and add proper commands for them.

packages/codemirror/src/extension.ts
```diff
@@ -798,16 +810,15 @@ export namespace EditorExtensionRegistry {
       }),
       Object.freeze({
         name: 'extendSelection',
-        default: true,
+        default: false,
```
krassowski, line 813:
```suggestion
        default: true,
```

krassowski:
It looks like the galata tests are failing because `Escape` does not leave the edit mode. I think there may be a similar CM issue with a similar solution. Also I need to change the expected shortcut for extending selection.

galata/test/jupyterlab/file-edit.test.ts
fcollonval, line 1: :heart:

fcollonval (approved):
Thanks @krassowski

Issue #16101, opened 2024-04-03
An extension disabled at the sys-prefix level can't be enabled at the user level


divyansshhh:
## Description
If an extension is enabled at the sys-prefix level, it can't be enabled at the user level. 

## Reproduce

Disable an extension at the sys-prefix level and try to enable it at the user level. 

## Expected behavior
Users should be able to opt into using extensions that have been disabled at the sys-prefix level

## Context

- JupyterLab version: v4.1.5

krassowski:
Reopening as #16102 did not actually fix this.

brichet:
Duplicate of https://github.com/jupyterlab/jupyterlab/issues/15574

brichet:
Thanks for working on this issue.
I still can't enable an extension at `user` level if it has been disabled at `sys_prefix` level.

For example with the following commands:
```bash
jupyter labextension disable --level=sys_prefix @jupyterlab/notebook-extension
jupyter labextension unlock --level=sys_prefix @jupyterlab/notebook-extension
jupyter labextension enable --level=user @jupyterlab/notebook-extension
```

The respective `page_config.json` files are correctly populated, but the extension is still not enabled:
- **sys_prefix level:**
```bash
$ cat ~/mambaforge/envs/jupyterlab/etc/jupyter/labconfig/page_config.json 
{
  "disabledExtensions": {
    "@jupyterlab/notebook-extension": true
  },
  "lockedExtensions": {
    "@jupyterlab/notebook-extension": false
  }
}
```
- **user level:**
```bash
$ cat ~/.jupyter/labconfig/page_config.json 
{
  "disabledExtensions": {
    "@jupyterlab/notebook-extension": false
  }
}
```

Did I missed something ?

brichet:
IIUC the `page_config.json` files are loaded with *jupyterlab_server* in [config.py -> get_static_page_config](https://github.com/jupyterlab/jupyterlab_server/blob/1e0367f5daf66c8c4a9fdccd66320b3006830770/jupyterlab_server/config.py#L76).
It uses the `ConfigManager` from *jupyter_server*, which read the config files from ['back to front' of the list](https://github.com/jupyter-server/jupyter_server/blob/af1342df16963b952374b9cb110db2121423f24f/jupyter_server/services/config/manager.py#L27).
This means that the `sys_prefix` one is the latest to be read, and overwrite the `user` one.

[...]

PR #16102, opened 2024-04-03
Fix extension toggling at different level

divyansshhh:
## References
Fixes jupyterlab/jupyterlab#16101

## Code changes
Changed the _AppHandler.toggle_extension method to look into all levels to decide whether an extension is enabled or not, if it's not, we change the page_config at the specified level


krassowski:
> If an extension is [disabled] at the sys-prefix level, it can't be enabled at the user level.

I think this was intentional prior to having extension locks. This guaranteed that admin could disable extension globally and user would not be able to change it.

Now that we have plugin locks this change makes sense to me.

This looks like a step in the right direction, but we would need some tests to:
- ensure that plugin cannot be toggled if locked at the higher level
- ensure that `jupyter labextension list` reflects the expected status

Those can be added in the "usage" tests group.

divyansshhh:
@krassowski I was trying to do the requested changes for this PR but on running the `disable` command for any extension, I get the following message and nothing changes in my `~/.jupyter/labconfig/page.json` file.

```
╰─⠠⠵ jupyter-labextension disable bqplot --level user
Starting with JupyterLab 4.1 individual plugins can be re-enabled in the user interface. While all plugins which were previously disabled have been locked, you need to explicitly lock any newly disabled plugins by using `jupyter labextension lock` command.
```

Is there something I'm missing?

divyansshhh:
FYI, this also needs jupyterlab/jupyterlab_server#448

divyansshhh:
@krassowski can you re-run the checks here now that jupyterlab_server 2.27.0 is released (I don't have the perms to)

divyansshhh:
@krassowski the failures seem unrelated to this PR, can you take a look and let me know if this is good to merge?

krassowski (approved):
Thank you @divyansshhh!

krassowski:

PR #16111, opened 2024-04-04
Store the real position of the item in reactive toolbar

brichet:
When an item is inserted in a reactive toolbar, its position is stored. This storage is used when the item moves between the main toolbar and the popup toolbar, to position them correctly.

For example if the expected position is 8 but there are only 2 widgets, it will be added at position 3. The stored position was 8 instead of 3, which cause errors on later computation.

## References

Should fix https://github.com/jupyter/notebook/issues/7311.

## Code changes

Store the real insertion index.

## User-facing changes

Avoid flickering items in reactive toolbar when some items have not been inserted at the expected position.

None

cc. @jtpio

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -452,11 +452,23 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     if (widget instanceof ToolbarPopupOpener) {
       status = super.insertItem(index, name, widget);
     } else {
-      index = Math.max(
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @brichet!

Wondering if a unit test would be relevant to cover this use case too?

brichet:
> Thanks @brichet!
> 
> Wondering if a unit test would be relevant to cover this use case too?

Right, it could be nice to have it.
I'll have a look.

krassowski (approved):
Thank you @brichet!

Issue #16113, opened 2024-04-05
Do not call `notifyCommandChanged` for commands registered by other plugins


jtpio:
## Description

As noticed in https://github.com/jupyterlab/jupyterlab/pull/14683, the logic added in https://github.com/jupyterlab/jupyterlab/pull/14683 calls `notifyCommandChanged` for commands not registered by the plugin itself.

This leads to errors logged in the dev tools console for lab-based apps using a subset of the plugins (for example Notebook 7), as noticed in https://github.com/jupyter/notebook/pull/7312#issuecomment-2039087409

## Reproduce

See https://github.com/jupyter/notebook/pull/7312#issuecomment-2039087409

![image](https://github.com/jupyter/notebook/assets/591645/fb1c6ed4-2c18-4fe3-9538-fa53aa4aea3c)

## Expected behavior

A plugin should only notify the commands it registered.

## Context

Originally reported in https://github.com/jupyterlab/jupyterlab/pull/14683#pullrequestreview-1982123117

jtpio:
Noticed another one when testing Notebook 7.2.0b0:

![image](https://github.com/jupyterlab/jupyterlab/assets/591645/90887781-da83-4723-af80-d3cfea8e752d)

krassowski:
@jtpio should we close this one with #16273, or is more work needed?

jtpio:
It seems to be limiting the number of errors in the dev tools console in Notebook 7, but not sure all cases are covered.

krassowski:
Per discussion two years ago it appears to have been addressed. If not let's reopen.

PR #16138, opened 2024-04-08
Do not notify the `notebook:create-output-view` command

jtpio:
## References

Partial fix for https://github.com/jupyterlab/jupyterlab/issues/16113

## Code changes

Just skip `notebook:create-output-view` for now.

The comment above mentions that all commands are added for simplicity:

https://github.com/jupyterlab/jupyterlab/blob/6152e3a5f94d728ccd2727d6d8413f12ddafb669/packages/notebook-extension/src/index.ts#L3566

But this should likely be reconsidered. It looks like only the plugins that register the commands should then call `app.commands.notifyCommandChanged()` for these particular commands.

## User-facing changes

None for most users. Should fix one error in the dev tools console when running Notebook 7: https://github.com/jupyter/notebook/pull/7312

None

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -3564,7 +3564,7 @@ function addCommands(
 
   // All commands with isEnabled defined directly or in a semantic commands
   // To simplify here we added all commands as most of them have isEnabled
-  const skip = [CommandIDs.createNew];
+  const skip = [CommandIDs.createNew, CommandIDs.createOutputView];
   const notify = () => {
     Object.values(CommandIDs)
       .filter(id => !skip.includes(id))


packages/notebook-extension/src/index.ts
```diff
@@ -3564,7 +3564,7 @@ function addCommands(
 
   // All commands with isEnabled defined directly or in a semantic commands
   // To simplify here we added all commands as most of them have isEnabled
-  const skip = [CommandIDs.createNew];
+  const skip = [CommandIDs.createNew, CommandIDs.createOutputView];
   const notify = () => {
     Object.values(CommandIDs)
       .filter(id => !skip.includes(id))
```
krassowski, line 3570: Could we call `app.commands.hasCommand(id)` here? Something like:

```suggestion
      .filter(id => !skip.includes(id) && app.commands.hasCommand(id))
```

Ref: https://lumino.readthedocs.io/en/latest/api/classes/commands.CommandRegistry-1.html#hasCommand

Would it be a full fix?

krassowski: @jtpio what do you think? I would like to release a new beta soon, but I do not want to do this without unblocking you on Notebook with this.

jtpio: Hmm I think the full fix would be to only notify the commands added by the plugin. But that would require keeping track of the list of commands and iterate through them in this loop to call `notifyCommanChanged`. Which is maybe why this code chooses to notify all of the them instead.

krassowski: But wouldn't we still want to notify commands added by other plugins (if these plugins are active)?

krassowski (approved):
Ok, let's move forward with this one!

Issue #15329, opened 2023-10-30
Split `ServiceManager` into multiple plugins


jtpio:
### Problem

Currently a JupyterLab application needs to instantiate a single `ServiceManager` before startup. This service manager is then accessible via the `JupyterFronEnd.IShell` interface. Core plugins and extensions can then access it via `app.serviceManager`.

A `ServiceManager` allows accessing server components and features such as contents, kernels, sessions, nbconvert:

https://github.com/jupyterlab/jupyterlab/blob/897418f0f25ffbbee3aa06a62770a53246823f85/packages/application/src/frontend.ts#L108-L111

However it is currently not possible to easily and partially replace some of its underlying services such as the kernel or session managers. Custom applications need to instantiate their own `ServiceManager` if they want to tweak the behavior of some of the internals.

### Proposed Solution

Probably the main reason for having the `ServiceManager` be a "special" component of a JupyterLab application is likely based on the assumption that a `JupyterFronEnd` should be served via a fully-featured Jupyter Server (or even a JupyterLab Server).

However over the past years we have seen more use cases that are starting to challenge this assumption. For example distributions with in-browser kernels like JupyterLite, or plugin-based Jupyter Server alternatives like Jupyverse, which use or enable only a subset of what a traditional Jupyter Server offers.

In these situations a plugin-based `ServiceManager` would allow for more easily disable or replace some services via normal extensions. This would mean providing each service of the `ServiceManager` as a token, for example `IKernelManager` or `ISessionManager`.

One challenge could be to keep backward compatibility with all existing extensions that access the contents, session, kernel managers via `app.serviceManager`. This could likely be achieved and would avoid existing extensions to have to require new tokens.

### Additional context

For reference there was a previous attempt at doing something similar: https://github.com/jupyterlab/jupyterlab/pull/13558. This PR was focused on replacing the whole `ServiceManager` via a plugin, but was still treated as a special case.

This would allow for some use cases.

**Allow extending the kernel / service / session manager using an extension**

Which could help fix: https://github.com/jupyterlab/jupyterlab/issues/12504

**Custom lab distributions**

Custom lab distributions like JupyterLite that currently need to redefine the whole `ServiceManager` to allow for custom kernel and contents managers. Splitting the `ServiceManager` into multiple plugins would allow for such applications to just be regular JupyterLab extensions instead.

**Remote content and kernels**

Users and administrators could more easily combine extensions, aggregate content from remote locations and use custom kernel managers as well.

[...]

PR #16141, opened 2024-04-09
Make the default drive in the Contents Service configurable

### LINKED ISSUE #1609: Create New Notebook-dialog: file name area cropped
The file name text is cropped both at top and bottom of text area:

screenshot from 2017-02-05 13-24-23

Using latest jupyterlab from conda-forge (v0.15.0), on ubuntu (16.04)

### LINKED ISSUE #16099: Make the default filebrowser a real drive, and make it configurable
cc. @davidbrochart @ellisonbg @Zsailer @SylvainCorlay @jtpio 

### Problem

Currently, the default file browser does not use a real drive (not [an IDrive](https://github.com/jupyterlab/jupyterlab/blob/41be86d5142a793ab9cb8b9eb99f92a33c8dc441/packages/services/src/contents/index.ts#L479)). Instead, it makes direct calls to the content manager using the REST APIs.

This makes it difficult to overwrite the behavior of the default filebrowser:
- JupyterLite has to implement routers to reroute the REST API calls
- jupyter-collaboration used to also reroute the REST API calls, but it was apparently difficult to maintain. Instead, they now implement a custom drive prefixed with `RTC:` and overwrite the default filebrowser to use it, this has its [load of issues](https://github.com/jupyterlab/jupyter-collaboration/issues/202). 
- jupyterlab-unfold is incompatible with jupyter-collaboration, because they both overwrite the default filebrowser.

### Proposed Solution

I would like to suggest that JupyterLab switches to making use of a real drive for the default filebrowser. 

- To keep the current behavior intact, this drive should be prefix-less. This means introducing the notion of prefix-less drives.
- There is already a [drive implementation making use of REST APIs](https://github.com/jupyterlab/jupyterlab/blob/41be86d5142a793ab9cb8b9eb99f92a33c8dc441/packages/services/src/contents/index.ts#L1072), JupyterLab could use it for the default file browser. (Saying such implementation is already available, making the feature "easy" to implement)
- This drive should be in its own plugin, making it easy for external extensions to overwrite it. This would also loosen the relationship between the default file browser widget and the default drive implementation, fixing incompatibility between jupyterlab-unfold and jupyter-collaboration.

Zsailer:
## References

Fixes #16099.

## Code changes

This PR makes two changes:
* _always_ creates a drive named "contents" to connect the server REST API
* defaults to making this drive the `defaultDrive`. The default drive doesn't need a prefix when a file path is given.
* adds a method, `setDefaultDrive` to update the default drive. 

Adds a public API to set/change the defaultDrive in the contents service. 

Leaving this in draft state for now to discuss.

## User-facing changes

There are no user-facing changes by default. 

This allows extensions like `jupyter-collaboration` to set the default drive to something other than the REST API. In the RTC case, it uses the ydoc APIs. 

None.

@martinRenou @SylvainCorlay @davidbrochart @dlqqq @jzhang20133 @jtpio

### DIFF
--- a/packages/services/src/contents/index.ts
+++ b/packages/services/src/contents/index.ts
@@ -295,6 +295,8 @@ export namespace Contents {
      */
     readonly serverSettings: ServerConnection.ISettings;
 
+    setDefaultDrive(driveName: string): void;
+
     /**
      * Add an `IDrive` to the manager.
      */
@@ -637,8 +639,30 @@ export class ContentsManager implements Contents.IManager {
   constructor(options: ContentsManager.IOptions = {}) {
     const serverSettings = (this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings());
-    this._defaultDrive = options.defaultDrive ?? new Drive({ serverSettings });
-    this._defaultDrive.fileChanged.connect(this._onFileChanged, this);
+    // Create a drive for the Jupyter Server REST API.
+    let contentsDrive = new Drive({ serverSettings, name: 'contents' });
+    this._additionalDrives.set('contents', contentsDrive);
+    // If a different drive as given in options, also add it.
+    let defaultDriveName = 'contents';
+    if (options.defaultDrive) {
+      defaultDriveName = options.defaultDrive.name;
+      this._additionalDrives.set(
+        options.defaultDrive.name,
+        options.defaultDrive
+      );
+    }
+    this.setDefaultDrive(defaultDriveName);
+  }
+
+  setDefaultDrive(driveName: string): void {
+    let drive = this._additionalDrives.get(driveName);
+    if (drive) {
+      // Remove old signal.
+      this._defaultDrive.fileChanged.disconnect(this._onFileChanged);
+      // Set the new Drive.
+      this._defaultDrive = drive;
+      this._defaultDrive.fileChanged.connect(this._onFileChanged, this);
+    }
   }
 
   /**
--- a/packages/services/src/contents/index.ts
+++ b/packages/services/src/contents/index.ts
@@ -669,7 +669,9 @@ export class ContentsManager implements Contents.IManager {
     let drive = this._additionalDrives.get(driveName);
     if (drive) {
       // Remove the signal from the previous default drive.
-      this._defaultDrive.fileChanged.disconnect(this._onFileChanged);
+      if (this._defaultDrive) {
+        this._defaultDrive.fileChanged.disconnect(this._onFileChanged);
+      }
       // Set the new Drive.
       this._defaultDrive = drive;
       this._defaultDrive.fileChanged.connect(this._onFileChanged, this);
--- a/packages/services/src/contents/index.ts
+++ b/packages/services/src/contents/index.ts
@@ -295,6 +295,9 @@ export namespace Contents {
      */
     readonly serverSettings: ServerConnection.ISettings;
 
+    /**
+     * Set the default `IDrive` in the manager.
+     */
     setDefaultDrive(driveName: string): void;
 
     /**
@@ -642,8 +645,9 @@ export class ContentsManager implements Contents.IManager {
     // Create a drive for the Jupyter Server REST API.
     let contentsDrive = new Drive({ serverSettings, name: 'contents' });
     this._additionalDrives.set('contents', contentsDrive);
-    // If a different drive as given in options, also add it.
+    // If a different drive is given in options, also add it here.
     let defaultDriveName = 'contents';
+    // Set the default drive based on the given options.
     if (options.defaultDrive) {
       defaultDriveName = options.defaultDrive.name;
       this._additionalDrives.set(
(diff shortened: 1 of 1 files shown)


krassowski:
I guess this is intended to fix #16099 (missing 9).

Zsailer:
oops! haha yes, thank you @krassowski. Updated.

packages/services/src/contents/index.ts
```diff
@@ -637,8 +639,30 @@ export class ContentsManager implements Contents.IManager {
...
+      );
+    }
+    this.setDefaultDrive(defaultDriveName);
+  }
+
+  setDefaultDrive(driveName: string): void {
+    let drive = this._additionalDrives.get(driveName);
+    if (drive) {
+      // Remove old signal.
+      this._defaultDrive.fileChanged.disconnect(this._onFileChanged);
```
krassowski, line 661:
```suggestion
      if (this._defaultDrive)
        this._defaultDrive.fileChanged.disconnect(this._onFileChanged);
      }
```

Zsailer: :+1:

jtpio:
Thanks @Zsailer for looking into this :+1: 

> Adds a public API to set/change the defaultDrive in the contents service.

It looks like being able to set a different default drive would be similar to being able to set other services (for example the kernel manager) provided by the `ServiceManager` by default as well (https://github.com/jupyterlab/jupyterlab/issues/15329).

If we start adding new public methods to replace parts of a `ServiceManager`, there may be a chance the API surface area will increase in the future, if we want to allow for swapping other services?

Raising the question now so we can get a better idea of how this would play with https://github.com/jupyterlab/jupyterlab/issues/15329 in the long run.

Zsailer:
> It looks like being able to set a different default drive would be similar to being able to set other services (for example the kernel manager) provided by the ServiceManager by default as well (https://github.com/jupyterlab/jupyterlab/issues/15329).

I think this particular case is slightly different. This doesn't replace the e.g. Contents Manager, simply changing a property on the Contents Manager, so it's much less disruptive than swapping out a whole service.

It _does_ increase the API surface a small amount. However, regardless of what we do to improve the pluggability of the ServiceManager proposed in #15329, swapping out default Drives underneath the Contents API seems like something we'd want to allow either way.

fcollonval:
I think allowing to change the default drive at runtime is a bad idea because there are no control over when it changes and it can change at any moment by multiple extensions.

As the default drive is already a top options of the `ServiceManager`:

https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/src/manager.ts#L223

I second the idea of @jtpio that a better approach would be to make the service manager more extensible with a clear definition of the default plugins used.

Something like:

Resolve plugins for the service manager -> Create the service manager -> Create the app -> load the other plugins.

packages/services/src/contents/index.ts
```diff
@@ -637,8 +642,40 @@ export class ContentsManager implements Contents.IManager {
   constructor(options: ContentsManager.IOptions = {}) {
     const serverSettings = (this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings());
-    this._defaultDrive = options.defaultDrive ?? new Drive({ serverSettings });
-    this._defaultDrive.fileChanged.connect(this._onFileChanged, this);
+    // Create a drive for the Jupyter Server REST API.
+    let contentsDrive = new Drive({ serverSettings, name: 'contents' });
```
fcollonval, line 646: The default drive name is `Default`

https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/src/contents/index.ts#L1079

So although it is not a good name if this PR is merged, I would not change it for backward compatibility.

Zsailer: :+1:

Zsailer:
> I think allowing to change the default drive at runtime is a bad idea because there are no control over when it changes and it can be changed at any moment by multiple extensions.

Fair enough. 

> Resolve plugins for the service manager -> Create the service manager -> Create the app -> load the other plugins.

Thanks @fcollonval. We discussed this in the JupyterLab call today and reached a similar conclusion; however, the specifics for how to achieve this was a bit muddy. I'll think about this more and see if I can come up with a draft PR. 

@afshin previously attempted something similar in #13558. I'll dive into this a bit more and see what I can learn. 

Thank you, all, for reviewing so quickly. I'll post an update here once I can get more time to think about this.

jtpio:
> Basically it involves extracting the plugin resolver logic in its own class (because why should such great feat be hard linked to the `Application`)

@fcollonval should we discuss this in an issue on the Lumino repo? Or maybe there is one already?

fcollonval:
> @fcollonval should we discuss this in an issue on the Lumino repo? Or maybe there is one already?

Issue opened => https://github.com/jupyterlab/lumino/issues/697

jtpio:
> Issue opened => [jupyterlab/lumino#697](https://github.com/jupyterlab/lumino/issues/697)

Thanks!

> I was not able to connect to the call

For reference there is the link to the recording in the meeting notes of the call, in case that can be useful: https://github.com/jupyterlab/frontends-team-compass/issues/229#issuecomment-2048074876

jtpio:
> @Zsailer Thank you for opening this issue! I noticed that [jupyterlab/lumino#703](https://github.com/jupyterlab/lumino/pull/703), which closed [jupyterlab/lumino#697](https://github.com/jupyterlab/lumino/issues/697), was merged some time ago. Is it possible to rebase and reconsider this PR for inclusion in JupyterLab?

Looks like this PR could likely be closed in favor of https://github.com/jupyterlab/jupyterlab/pull/16556, which would allow customizing the default drive via an extension.

krassowski:
@Zsailer can we close this PR in favour of:
- https://github.com/jupyterlab/jupyterlab/pull/17092
- https://github.com/jupyterlab/jupyterlab/pull/16794

or would you like it to still be considered in the future?2

Zsailer:
yep, let's close :+1:

PR #16154, opened 2024-04-12
Add information about configuration Windows 11 for building an extension

AnastasiaSliusar:
The issue:
- If a user has Windows 11 and tries to make a build for an extension it is obligation to configure Windows 11. Otherwise JupyterLab will use an old build even though there will be a new one.

## Code changes

Adding a note to the Extension tutorial about necessity of configuring "For Developer" mode on Windows 11.


docs/source/extension/extension_tutorial.rst
```diff
@@ -201,7 +201,7 @@ JupyterLab:
    On Windows, symbolic links can be activated on Windows 10 or above for Python version 3.8 or higher
    by activating the 'Developer Mode'. That may not be allowed by your administrators.
    See `Activate Developer Mode on Windows <https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development>`__
-   for instructions.
+   for instructions. Making this and runing ``jupyter labextension develop --overwrite`` is necessary for Windows 11, so that JupyterLab will use an updated build.
```
krassowski, line 204: I recently noticed that this information is not exposed properly too - thank you for proposing to emphasise it more!

> and runing ``jupyter labextension develop --overwrite`` is necessary for Windows 11, so that JupyterLab will use an updated build

This is in general true regardless of whether Windows 11 is used or any other OS. I would suggest not to include this sentence.

Instead, I would suggest that we make three different changes to address the underlying issue:
- let's mention the need to enable symbolic links in [Developing a prebuilt extension](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#developing-a-prebuilt-extension) section; this could be a single sentence and a link to this note (you will need to create an anchor).
- lets reword this box to say that "links _need_ to be activated" rather than "links _can_ be activated"
- let's consider if we should change it from `note::` to `important::` admonition. Maybe it should even have a custom title, like "Important for Windows users".

AnastasiaSliusar: @krassowski sure! Thank you for suggestion.

AnastasiaSliusar: Dear @krassowski I have updated this PR according to your suggestions. Please, take a look. Thank you:)

docs/source/extension/extension_dev.rst
```diff
@@ -548,7 +548,7 @@ While authoring a prebuilt extension, you can use the ``labextension develop`` c
 
    jupyter labextension develop . --overwrite
 
-Then rebuilding your extension and refreshing JupyterLab in the browser should pick up changes in your prebuilt extension source code.
+Then rebuilding your extension and refreshing JupyterLab in the browser should pick up changes in your prebuilt extension source code. If Windows is used, please, take a look how to configure Windows :ref:`important-for-windows-users`
```
krassowski, line 551: Three things here:
1. Splitting to new line to make editing easier (it will render the same)
2. Using active voice to align with the Jupyterlab documentation [writing style](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#writing-style).
3. Explicitly mentioning that the `develop` command may not work without additional configuration.

```suggestion
Then rebuilding your extension and refreshing JupyterLab in the browser should pick up changes in your prebuilt extension source code.
If using Windows, you may need to configure your operating system for the ``develop`` command described above to work, please see the note: :ref:`important-for-windows-users`
```

krassowski (approved):
Looks good to me, thank you @AnastasiaSliusar! Just a small suggestion on rewording the added line, see comment in code.

PR #16156, opened 2024-04-12
Fix accepting completion with `Enter`

### LINKED ISSUE #16150: `Enter` key does not accept a completion in 4.2.0b1
## Description

> I invoke the completion with tab as usual which works fine, also tabbing through works as expected. But the only way to select something is by clicking with the mouse. With the keyboard I can just type or press escape to close the completer.
>
> Originally reported by @spookster on https://discourse.jupyter.org/t/jupyterlab-4-2-beta-is-available-for-testing/25097/2

## Reproduce

enter-does-not-work

## Expected behavior

- Enter accepts the completion
- We have an integration test case for this

## Context

It works well in 4.2.0b0 so it looks like this was introduced as a side-effect of lumino upgrade in https://github.com/jupyterlab/jupyterlab/pull/16078 and was not picked up by the tests.

krassowski:
## References

Fixes #16150

## Code changes

- [x] Adds a test for accepting completion with <kbd>Enter</kbd>
- [x] Fixes the issue by removing the default Enter handler and returning `false` from our custom handler.

## User-facing changes

Accepting completion with Enter works

None

### DIFF
--- a/galata/test/jupyterlab/completer.test.ts
+++ b/galata/test/jupyterlab/completer.test.ts
@@ -13,7 +13,9 @@ test.describe('Completer', () => {
       await page.notebook.createNew(fileName);
     });
 
-    test('Open completer on notebook', async ({ page }) => {
+    test('Open completer on notebook and accept suggestion', async ({
+      page
+    }) => {
       await page.notebook.setCell(
         0,
         'code',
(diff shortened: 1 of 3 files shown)


fcollonval (approved):
Thanks @krassowski

PR #16157, opened 2024-04-13
Enable "Search Cell Outputs" in Find/Replace, disable "Replace" button when an output is highlighted

### LINKED ISSUE #15445: search in output
## Description

Find in output doesn't seem to work in jupyter lab even though was fixed years ago.

Screenshot from 2023-11-27 14-49-15

Related: https://stackoverflow.com/questions/66461218/jupyter-lab-search-find-in-outputs-shortcut

## Reproduce

Click anywhere in the notebook then Ctrl+F.

## Context

- Operating System and version: Ubuntu 20.04
- Browser and version: Firefox 120.0
- JupyterLab version: 4.0.8

### LINKED ISSUE #15966: Enable "Search Cell Outputs" in Find/Replace, disable "Replace" button when an output is highlighted
This is a follow up to #13167.

When a user searches for something, and has the "Replace" option enabled, we now disable the "Search Cell Outputs" feature. This is because outputs cannot be modified directly, but this may miss a user expectation: users expect that they can still search cell outputs, but only replace writable content.

Re-enable "Search Cell Outputs" all the time, even when the "Replace" options are showing. After searching, the "Replace" button should be disabled if the currently highlighted match is in a cell output. The "Replace All" button should replace all matches except for those in cell outputs.

Tanmay-Deshmukh:
Fixes #15966 
Closes #15445

Changed searchprovider.ts in notebook package to allow for searching cell outputs while replace is active. Passed information on currently highlighted string and if it is an output to replace overlay to disable "Replace" button when highlighting an output.

Replace is disabled but not grayed out when highlighting an output due to what I believe to be a styling issue.


JasonWeill:
@Tanmay-Deshmukh Thank you for opening this PR! I checked out your code. I think the "styling issue" you noted in your description should be fixed as part of this change. I appreciate that I can still search in cell outputs when the replace options are visible, but the "Replace" button should be visibly disabled (different color, different cursor, tooltips, etc) when I'm currently looking at a cell output result. In addition, I don't see a highlight to indicate which result is actively selected when I'm iterating through results in cell outputs. The "Replace All" button should be disabled if the only results are inside cell outputs, since none of these can be replaced.

krassowski:
(I shortened the PR title as we also use the to auto-generate changelog)

krassowski:
@Tanmay-Deshmukh are you interested in finishing up this PR? It is ok to say no, I just need to make a quick update on this one for release planning.

krassowski:
Based on the review from @JasonWeill which I interpret to mean "changes need to be made" I am switching this PR to draft. Please undraft when the changes are addressed :)

krassowski:
Given lack of response, I am closing this in in favour of https://github.com/jupyterlab/jupyterlab/pull/16260 which builds upon this PR.

PR #16160, opened 2024-04-14
Clarify the LSP documentation

krassowski:
## References

As seen in https://github.com/jupyterlab/jupyterlab/discussions/15145 the LSP documentation introduced in https://github.com/jupyterlab/jupyterlab/pull/12899 is causing a lot of confusion to the users.

The language servers listed are also confusing because it suggests that other servers are not supported by JupyterLab, whereas in fact many more servers can be used. Further this list drifts away from the source of truth in `jupyter-lsp` repository so I do not think it should stay here unless anyone is interested in actively monitoring and updating it.

## Code changes

- Reword the warning in LSP section to avoid confusion
- Add another warning in LSP section because users will not read the first warning
- Remove the list of servers as it causes confusion
- Make note that the settings page will look different when `jupyterlab-lsp`  is installed

## User-facing changes

As above.

None

### DIFF
--- a/docs/source/user/lsp.rst
+++ b/docs/source/user/lsp.rst
@@ -4,95 +4,27 @@
 .. _lsp:
 
 Language Server Protocol support
-=====================================
+================================
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

krassowski:

Issue #16025, opened 2024-03-19
Use ellipses consistently in menu items and commands


JasonWeill:
This is a spin-off from #15946.

[Apple's Human Interface Guidelines for menus](https://developer.apple.com/design/human-interface-guidelines/menus) suggest, "Append an ellipsis to a menu item’s label when people need to provide additional information before the action can complete." A cascading menu, like File > Save and Export Notebook As, should not have an ellipsis, as it does now.

Audit commands in pull-down menus; for each one that pops up a dialog box to accept additional information, ensure that there is an ellipsis in the menu item. For each menu item that takes immediate action, there should be no ellipsis. Cascading menus should never have an ellipsis, but submenu items that require additional information should have one.

Update the JupyterLab documentation to mention this style improvement.

BiniyamDS:
I am beginner to open source so could I take this issue

krassowski:
Please do!

Rob-P-Smith:
Working on this now.

PR #16168, opened 2024-04-15
Update documentation on menu item labelling and ellipsis use

hanlaman:
Use ellipses consistently in menu items and commands #16025

Updated the JupyterLab documentation to mention Apple's Human Interface Guidelines for menus' style improvements.


docs/source/index.rst
```diff
@@ -51,3 +51,24 @@ See the sections below (and the top-level links above) for more information abou
    extension/extension_dev
    developer/contributing
    privacy_policies
+
+.. .. only:: html
+
+   <div class="alert alert-warning" role="alert">
+      <strong>Note:</strong> The following section provides guidelines for menu item labeling based on Apple's Human Interface Guidelines.
+   </div>
```
krassowski, line 59: I am not sure if this should be an alert. It currently does not render at all:

image

If it should be a visible admonition, let's implement it using the native admonitions syntax, for example with `.. note:`

https://github.com/jupyterlab/jupyterlab/blob/73a947de96941d7d254f397d7f2cbb0df7f69958/docs/source/user/toc.rst?plain=1#L24-L31

which render like this:

image

See https://jupyterlab.readthedocs.io/en/latest/user/toc.html

krassowski (changes requested):

I would suggest to move the added text from index to either:
- [the main contributing guide page](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html), or
- [the design patterns page](https://jupyterlab.readthedocs.io/en/latest/developer/patterns.html#command-names)

Can you update the title of the PR to reflect the content (what does this PR change) rather than which files it edits?

Finally, I left a comment on the alert added. It does not render in the preview for me, I am not sure if we want to keep it but if we do let's use the native admonition.

krassowski:
@hanlaman thank you again for picking this one up! Just checking if you saw my comments/suggestions above?

krassowski:
@hanlaman if you are no longer interested, should we close this PR to signal that anyone else can open a new one addressing the change requests above?

Issue #15103, opened 2023-09-10
Flaky `Open Sidebar tab jp-running-sessions` test


krassowski:
## Description

I saw this a couple of times, for example [here](https://github.com/jupyterlab/jupyterlab/actions/runs/6130672040/job/16648118565?pr=15101). It appears that one of the tests which runs in the group before it is not shutting its kernel down properly, leading to a dangling kernel.

```
Error:   1) [jupyterlab] › test/jupyterlab/sidebars.test.ts:31:9 › Sidebars › Open Sidebar tab jp-running-sessions 
    Error: Screenshot comparison failed:

      395 pixels (ratio 0.01 of all image pixels) are different.

    Expected: /home/runner/work/jupyterlab/jupyterlab/galata/test-results/test-jupyterlab-sidebars-Sidebars-Open-Sidebar-tab-jp-running-sessions-jupyterlab/opened-sidebar-jp-running-sessions-expected.png
    Received: /home/runner/work/jupyterlab/jupyterlab/galata/test-results/test-jupyterlab-sidebars-Sidebars-Open-Sidebar-tab-jp-running-sessions-jupyterlab/opened-sidebar-jp-running-sessions-actual.png
        Diff: /home/runner/work/jupyterlab/jupyterlab/galata/test-results/test-jupyterlab-sidebars-Sidebars-Open-Sidebar-tab-jp-running-sessions-jupyterlab/opened-sidebar-jp-running-sessions-diff.png

      36 |       const position = await page.sidebar.getTabPosition(sidebarId);
      37 |       const sidebar = await page.sidebar.getContentPanel(position);
    > 38 |       expect(await sidebar.screenshot()).toMatchSnapshot(
         |                                          ^
      39 |         imageName.toLowerCase()
      40 |       );
      41 |     });

        at /home/runner/work/jupyterlab/jupyterlab/galata/test/jupyterlab/sidebars.test.ts:38:42
```

| Actual | Expected |
|--------|--------|
| ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/82417132-b2ee-4056-acac-ee5de16e83d4) | ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/ce7e1f40-9cea-47ef-9c75-519b964edb2e) |

krassowski:
[...]

PR #16188, opened 2024-04-18
Fix error in kernels sidebar when switching kernels, remove unused prop

krassowski:
## References

Addresses some issues noted in https://github.com/jupyterlab/jupyterlab/issues/15103 and failing check on main, but does not solve it yet.

## Code changes

- Fixes incorrect access to first element in the `children` list causing an uncaught error
- Reverts incorrectly updated snapshot which was in a state indicating the error fixed above
- Removes unused private props as these were confusing when trying to navigate code

## User-facing changes

When switching kernels, the kernels section will not get stuck.

None


fcollonval (approved):
Thanks @krassowski

PR #16189, opened 2024-04-18
Enable `noUncheckedIndexedAccess`

krassowski:
## References

A number of errors in the codebase such as https://github.com/jupyterlab/jupyterlab/issues/15103#issuecomment-2063421265 could be prevented by enabling `noUncheckedIndexedAccess`.

Trying to evaluate the benefit/cost ratio in this PR

## Code changes

TBD

## User-facing changes

None

None

### DIFF
--- a/packages/coreutils/src/markdowncodeblocks.ts
+++ b/packages/coreutils/src/markdowncodeblocks.ts
@@ -57,8 +57,7 @@ export namespace MarkdownCodeBlocks {
     const lines = text.split('\n');
     const codeBlocks: MarkdownCodeBlock[] = [];
     let currentBlock = null;
-    for (let lineIndex = 0; lineIndex < lines.length; lineIndex++) {
-      const line = lines[lineIndex];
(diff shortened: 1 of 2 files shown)


krassowski:
There is a few annoyances here:

```ts
val: string
const [key, value] = val.split('=');
```

both `key` and `value`are treated as `string | undefined`  although `key` is in fact `string`.

We cannot use `parts.length > 1` for a check, but need to do `parts[0] !== undefined`:

```ts
export function join(...parts: string[]): string {
    if (parts[0] === undefined) {
      throw Error('join requires at least one string part')
    }
    parts[0]
}
```

Changing typing to:

```ts
export function join(...parts: [string, ...string[]]): string {
    parts[0]
}
```

works but requires calling with first ad rest like:

```ts
// cannot do this: join(...url.split('/').map(encodeURIComponent));
const [first, ...rest] = url.split('/').map(encodeURIComponent);
join(first as string, ...rest);
```

again because it cannot infer that `first` will be always `string`

PR #16198, opened 2024-04-19
Fix updating search value on input

### LINKED ISSUE #16193: Search/filter boxes do not update when typing
## Description

Testing locally after merge of https://github.com/jupyterlab/jupyterlab/pull/16046 - the search/filter boxes only take effect after `onblur`.

From a quick look there is `<Search onChange=` and previously there was `<InputGroup onChange=`

My guess is that this the `Search` component does not call the callback on every keystroke compared to the `<input` that we used previously?

> Unlike the [input](https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event) event, the change event is not necessarily fired for each alteration to an element's value.

CC @fcollonval @Mehak261124 

## Reproduce

search-boxes-take-too-long

## Expected behavior

Search filter boxes take effect as user types as in 3.x and 4.0 and 4.1

## Context

- JupyterLab version: 4.2.0b2

fcollonval:
## References

Fixes #16193

## Code changes

React to `onInput` event on `Search` react component

## User-facing changes

Restore reaction on user input.

None

### DIFF
--- a/packages/ui-components/src/components/search.tsx
+++ b/packages/ui-components/src/components/search.tsx
@@ -191,6 +191,7 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
       ref={props.inputRef}
       value={filter}
       onChange={handleChange}
+      onInput={handleChange}
       placeholder={props.placeholder}
       disabled={props.disabled}
     >


krassowski (approved):
Thank you @fcollonval!

Issue #14396, opened 2023-04-17
Create an accessibility section in JupyterLab documentation


steff456:
### Problem

- We don't have any documentation that is accessibility-focused in the JupyterLab documentation Given that the community is including accessibility-focused improvements, it is becoming increasingly necessary that we provide a dedicated focus on adding one page for user accessibility documentation for JupyterLab.

### Proposed Solution

- Create an accessibility section in JupyterLab documentation containing default accessibility settings and customizations inspired by https://code.visualstudio.com/docs/editor/accessibility

steff456:
> This is a single page, right?

Yes! It is a single page

> Perhaps that should be added as a note at the top of the page? Not sure how to message it.

We can add it like the dates that are in the accessibility documentation :)

krassowski:
I see that accessibility section is in there: https://jupyterlab.readthedocs.io/en/latest/getting_started/accessibility.html

It looks a bit outdated, showing 2022 date and JupyterLab 3.4.4. Are there plans to update it going forward?

gabalafou:
The accessibility statement and this issue are separate concerns.

This issue was about creating a help section for a disabled end-user audience that explains various accessibility-related tips and tricks for JupyterLab.

The accessibility statement is more of a formal document about the overall state of JupyterLab conformance to WCAG. 

Unfortunately, the help section work had to be abandoned.

I re-read the accessibility statement just now, and to my eyes, apart from the reference to the older version of JupyterLab, I'm not sure much needs to be changed. I created a pull request to make those changes: #16206.

PR #15001, opened 2023-08-22
[WIP] Add user accessibility documentation

dannyjohnson2050:
Created on: August 22, 2023

Last updated on: August 24, 2023

## References

https://github.com/jupyterlab/jupyterlab/issues/14396

## Code changes

I have implemented the user accessibility documentation by creating comprehensive sections that outline JupyterLab's accessibility features, customization options, known issues, and further resources. These changes involve adding detailed explanations of keyboard navigation, tab trapping, color themes, text scaling, focus indicators, and ARIA roles and labels. Additionally, I have addressed accessibility-related findings in the "Advanced Settings," including font size options and keyboard shortcuts.

These changes fulfill the project's need to provide users with a clear and structured guide to enhancing accessibility within JupyterLab. By documenting these features, users will have the knowledge and tools to navigate, customize, and interact with JupyterLab in an accessible manner.

## User-facing changes

The user accessibility documentation introduces JupyterLab users to a comprehensive guide on enhancing accessibility. Users will now have access to information on keyboard shortcuts, customization options, and accessibility-related issues. This documentation will empower users to navigate JupyterLab efficiently using keyboard input, customize the interface to match their accessibility requirements, and address common accessibility challenges. This contribution enhances the overall accessibility experience for JupyterLab users, making it more inclusive and user-friendly.

PR #16206, opened 2024-04-19
Update Accessibility Statement

gabalafou:
To my eyes, this statement is still current for JupyterLab 4, so I made a few tweaks so that it doesn't seem to only be addressing version 3. 

I also made a few edits and moved the link to the CZI grant roadmap to the "Links" section.

The background/motivation for this PR is a comment Mike wrote on April 19, 2024: 

> I see that accessibility section is in there: https://jupyterlab.readthedocs.io/en/latest/getting_started/accessibility.html
> It looks a bit outdated, showing 2022 date and JupyterLab 3.4.4. Are there plans to update it going forward?
> 
> -- https://github.com/jupyterlab/jupyterlab/issues/14396#issuecomment-2066601226:

### DIFF
--- a/docs/source/getting_started/accessibility.rst
+++ b/docs/source/getting_started/accessibility.rst
@@ -10,6 +10,25 @@ JupyterLab Accessibility Statement
 
     This statement was last updated on the 22nd of July 2024.
 
+Jupyter Notebook versus JupyterLab
+----------------------------------
+
+For accessibility purposes, it is recommended to use `Jupyter Notebook
+<https://jupyter-notebook.readthedocs.io/en/stable/>` over JupyterLab.
+
+A comparison of the two apps will help to explain this recommendation. Jupyter
+Notebook and JupyterLab are both web apps for authoring computational notebooks.
+But Jupyter Notebook has a document-centric user interface whereas JupyterLab
+provides multiple panels and tabs designed to work with several notebooks and
+extensions in the same interface. Jupyter Notebook is more like Google Docs, in
+which each document opens in a separate browser tab. JupyterLab is more like VS
+Code for the Web, in which the app provides its own, in-app tabs so that
+multiple documents can be opened without having to open more browser tabs.
+Because Jupyter Notebook has a simplified UI, it poses fewer accessibility
+challenges than JupyterLab (in particular, it's easier to zoom in). But because
+it shares the same codebase, it also benefits from all of the accessibility work
+done in JupyterLab. This is why Jupyter Notebook is recommended over JupyterLab.
+
 Jupyter audits
 --------------
 
--- a/docs/source/getting_started/accessibility.rst
+++ b/docs/source/getting_started/accessibility.rst
@@ -6,6 +6,10 @@
 JupyterLab Accessibility Statement
 ==================================
 
+.. note::
+
+    This statement was last updated on the 19th of April 2024.
+
 Jupyter audits
 --------------
 
@@ -22,16 +26,12 @@ Accessibility Statement for JupyterLab
 
 Edited from the `W3C accessibility statement generator <https://www.w3.org/WAI/planning/statements/generator/#create>`_
 
-This is an accessibility statement for JupyterLab 3.4.4 from Jupyter accessibility contributors. We acknowledge there are more recent releases of JupyterLab that have not yet been evaluated.
+This is an accessibility statement for JupyterLab from Jupyter accessibility contributors.
 
 The current state of JupyterLab
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Jupyter accessibility statements are living documents. This statement was first created on 16 May 2022 using the `W3C Accessibility Statement Generator Tool <https://www.w3.org/WAI/planning/statements/>`_ with additions and edits from the Jupyter accessibility contributors community.
-
-.. note::
-
-    This statement was last updated on the 09 of August 2022.
+Jupyter accessibility statements are living documents. This statement was first created for JupyterLab 3.4.4 on 16 May 2022 using the `W3C Accessibility Statement Generator Tool <https://www.w3.org/WAI/planning/statements/>`_ with additions and edits from the Jupyter accessibility contributors community.
 
 Conformance status
 """"""""""""""""""
@@ -116,7 +116,7 @@ Assessment approach
 Jupyter accessibility contributors assessed the accessibility of JupyterLab by the following approaches:
 
 * Self-evaluation
-* Regular automated testing to monitor for regressions (can be found at `the jupyter-a11y-testing repository <https://github.com/Quansight-Labs/jupyter-a11y-testing>`_).
+* Automated testing (can be found at `the jupyter-a11y-testing repository <https://github.com/Quansight-Labs/jupyter-a11y-testing>`_).
 * User feedback
 
 Evaluation report
@@ -139,9 +139,6 @@ Jupyter accessibility contributors take the following measures to ensure accessi
 * Employ formal accessibility quality assurance methods.
 * Document changes, approaches, and improvements to the above methods and to JupyterLab itself.
 
-.. note::
-
-    For more information on current efforts to improve JupyterLab's accessibility, visit the `CZI - JupyterLab accessibility grant roadmap <https://jupyter-a11y.netlify.app/roadmap/intro.html>`_.
 
 Feedback and Formal complaints
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


gabalafou:
I'm not really sure this document belongs under "Get Started" - should we move it somewhere else while keeping the old URL alive?

krassowski:
> I'm not really sure this document belongs under "Get Started" - should we move it somewhere else while keeping the old URL alive?

Well I think that it is fine; it includes things which a potential user or operator should know before installing/deploying JupyterLab. I think we should have a separate section under "User Guide" which talks about how a user/operator can enable accessibility options such as the High Contrast Theme and suggests specific customizations that may be preferred for users with accessibility needs (maybe changing some shortcuts, maybe increasing font size by default, maybe adding some buttons, all depending on what a user needs).

krassowski:
> JupyterLab is not compatible with
> 
> Browsers (mobile and desktop):
>    Internet Explorer
>     Edge

JupyterLab **is** compatible with Edge 79 and newer. It is not compatible with Edge <79. This needs to be updated, Edge is like 13% of the market share and as of now it is wrongly saying they would need to switch browsers to use JupyterLab.

krassowski:
Is it still accurate that JupyterLab is not compatible with any of the screen readers after the substantial number of PRs improving specific areas? What does it mean to be compatible, is it like 100% elements work as user would expect (where expectations will vary user-to-user) or that most things can be done with a screen reader?

krassowski:
> JupyterLab is nonconforming with WCAG 2.1 level AA. 

From PoV of maintaining this we should have a list of issues remaining to meet AA. I know this is partially related to automated testing and partially to issue tracking. Maybe we should have two separate labels in in issue tracker, one for AA and one for AAA. Of top of my head I don't know of major areas where AA is not meat, but certainly there are many smaller issues remaining.

tonyfast:
i dont think notebook 7 was an idea when this accessibility statement was posted. i think that the accessibility statement should be update to recommend notebook 7 as our _most accessible_ authoring system for assistive technology users. for assistive technology, it is the least confusing situation to work in.

i think it would be worth noting improvements to low vision zoom modes. low vision/magnifier users will have a better experience  notebook 7 over jupyterlab. low vision mode makes more sense notebook cause of the application doesnt take up real estate on the margins.

> Of top of my head I don't know of major areas where AA is not met, but certainly there are many smaller issues remain.

there are some forms that are definitely going to fail wcag AA at this moment, but otherwise several issues were handled that we had before.

> From PoV of maintaining this we should have a list of issues remaining to meet AA. I know this is partially related to automated testing and partially to issue tracking. Maybe we should have two separate labels in in issue tracker, one for AA and one for AAA. 

it can be hard to separate these AA/AAA at times, especially with wcag 2.0/2.1. wcag 2.2 disambiguated some of these confusions, but our targets are not modern standards. a lot of the contributions we've recieved for accessibility have been towards compliance of section 508 which requires satisfying wcag 2.0 AA. i'd suggest:

* we mention lab is NOT section 508 compliant to wcag 2.0 AA. this would indicate to institutions that lab is not compliant with federal standards for access to information technology.
* we should triage past audits and see where we stand. there are multiple checklists that should be verified about the state they are in. after this effort we'd be in a better place to track the issue progress.
* we use funds to pay for a formal audit of jupyter notebook and actively work to redress the inaccessibility.

isabela-pf (approved):

krassowski (changes requested):
I realise that I commented but have not used the GitHub review feature before - this is because it does not allow to use code reviews on things outside of the diff.
I would appreciate if the points raised in the discussion were addressed, in particular the Edge browser issue, and otherwise just an acknowledgement for the other points raised (seeing Isabela's review I suspect it was possibly discussed on accessibility meeting - sorry I could not make it this week, I'm off sick, but even if things are discussed in meetings it helps future selves a lot of those are briefly summarised on the PRs).

krassowski:
WDYT @gabalafou?

krassowski:
@gabalafou would it be ok if I pushed small updates to reflect the browser support? 

> drop the 2.1 constraint to 2.0. we mention lab is NOT section 508 compliant to wcag 2.0 AA. this would indicate to institutions that lab is not compliant with federal standards for access to information technology

What is the industry practice to saying that a product is compliant? Surely, there are always bugs in each complex product, 
 for example I doubt that all applications shipped by Windows are compliant with accessibility standards (but most users don't use them!), is there a point that we will ever fill confident to say that lab is compliant or would we need someone else to say that the core functionality is complaint?

My thinking is pragmatic. If 90% of lab functionality is compliant but 10% is not, but this 10% is optional and is not available in closed-source alternatives, it looks like this statement would encourage institutional users to use the closed-source alternative, even though OSS is not worse in the core experience. Would having some per-component statements like
- Core notebook authoring UI - compliant to X since version X.X.X, with limitations A, B, C
- Text file edition - compliant to X since version X.X.X
- CSV edition - not compliant (uses canvas without assistant)
- Terminal - compliant in HTML mode (needs configuration), not compliant by default
- Advanced Settings Editor - not compliant

be potentially more useful?

tonyfast:
> would it be ok if I pushed small updates to reflect the browser support?

an accessibility statement should include all the user agents used for testing; this is the browser + any assistive technologies (eg screen reader, braille display). as an example, [vscode](https://github.com/microsoft/vscode/wiki/Accessibility-Guidelines#screen-readers) shares the screen readers used for testing. jupyter has been tested with smattering of nvda, jaws, orca, narrator, and voiceover. we are not good about documenting that, but should be.

> What is the industry practice to saying that a product is compliant? 

industry practice would include experts in accessibility in the decision making process. using legal expert. in the US, we could be validate section 508 (wcag 2.0) compliance by paying for compliance audits from [trusted testers](https://www.dhs.gov/trusted-tester). in the EU, the [european accessibility act](https://www.levelaccess.com/compliance-overview/european-accessibility-act-eaa/) will push for wcag 2.2 compliance by 2025. 

> Surely, there are always bugs in each complex product, for example I doubt that all applications shipped by Windows are compliant with accessibility standards (but most users don't use them!), is there a point that we will ever fill confident to say that lab is compliant or would we need someone else to say that the core functionality is complaint?

the difference between bugs and accessibility violations are that bugs are considered in the design and the violations are features excluded from the design; neglecting to add features or consider others access needs is different than a bug. yes, there can be a point that we say lab is compliant ( realistically it would make the most sense to make notebook compliant first cause it has reduced complexity). right now, i don't think we can begin to say that jupyter is compliant until we meet the needs in #9399 and https://github.com/jupyter/notebook/issues/6800 .

> My thinking is pragmatic. If 90% of lab functionality is compliant but 10% is not, but this 10% is optional and is not available in closed-source alternatives, it looks like this statement would encourage institutional users to use the closed-source alternative, even though OSS is not worse in the core experience. Would having some per-component statements like

pragmatism makes sense when everyone shares the same reality, but disabled people experience [spacetime](https://thesophian.com/moving-together-toward-a-theory-of-crip-spacetime-margaret-price-speaks-during-disability-visibility-week/) differently than the majority of the world so reality might not be the best reference. regardless, i have not seen any specific statements from other products regarding the granularity of their accessibility. i have seen highlights of specific accessibility features like [vscode](https://vscode-docs.readthedocs.io/en/latest/editor/accessibility/). there might legal considerations in statements like these so i think we'd need to defer to experts there.

gabalafou:
Hi folks, I pushed two commits to try to incorporate some of the discussion on this PR. 

@krassowski, I wonder if part of what you're getting at is the lack of a user guide docs page like [VS Code Accessibility](https://code.visualstudio.com/docs/editor/accessibility). As you may recall, we started working on this, but the work had to be abandoned (see #15001).

I do share your sentiment about it feeling a little too black-or-white to say that JupyterLab does not support any screen readers. But these kinds of accessibility statements are formal documents and I'm not quite sure I know how to incorporate your thoughts into this document. At any rate, it's more than I want to tackle in this PR.

krassowski (approved):
Thank you @gabalafou!

> But these kinds of accessibility statements are formal documents

Out of curiosity do you have links to equivalent statements (formal documents) for other editors/IDEs/similar?

gabalafou:
That's a good question. I do not. I spent about 10-15 minutes searching but couldn't turn up accessibility statements for other editors (Google Docs, VS Code, Jetbrains, and a few others). I came across [Atlassian's accessibility page](https://www.atlassian.com/accessibility), which is interesting because it includes a [list of audits (VPATs)](https://www.atlassian.com/accessibility#vpat) for their products but not an accessibility statement per se.

Just FYI, the document in this PR was at least scaffolded using the W3 [generator tool](https://www.w3.org/WAI/planning/statements/generator/#create).

PR #16208, opened 2024-04-20
Clear output area when empty

### LINKED ISSUE #14299: Empty output area is not cleared

NexVeridian:
## References
closes #14299

## Code changes
Clear output area when empty in Callback on output changes, see [this](https://github.com/jupyterlab/jupyterlab/issues/14299#issuecomment-1491137352)

## User-facing changes
For before see #14299

After:

https://github.com/jupyterlab/jupyterlab/assets/46305192/d97241f8-0cae-4103-83b3-209a8dd5e9e3

None

### DIFF
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -1603,6 +1603,11 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
     // This is to hide/show icon on single line output.
     this.updatePromptOverlayIcon();
+
+    // Clear output area when empty
+    if (this.model.outputs.length === 0) {
+      this.outputArea.node.style.height = '';
+    }
   }
 
   /**
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -1593,6 +1593,7 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
   }
 
+  private _lastOutputHeight = '';
   /**
    * Callback on output changes
    */
@@ -1605,8 +1606,11 @@ export class CodeCell extends Cell<ICodeCellModel> {
     this.updatePromptOverlayIcon();
 
     // Clear output area when empty
+    this._lastOutputHeight = this.outputArea.node.style.height;
     if (this.model.outputs.length === 0) {
       this.outputArea.node.style.height = '';
+    } else if (this.model.outputs.length > 0) {
+      this.outputArea.node.style.height = this._lastOutputHeight;
     }
   }
 
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -1593,7 +1593,6 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
   }
 
-  private _lastOutputHeight = '';
   /**
    * Callback on output changes
    */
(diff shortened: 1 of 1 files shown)


packages/cells/src/widget.ts
```diff
@@ -1603,6 +1603,11 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
     // This is to hide/show icon on single line output.
     this.updatePromptOverlayIcon();
+
+    // Clear output area when empty
+    if (this.model.outputs.length === 0) {
+      this.outputArea.node.style.height = '';
+    }
```
krassowski, line 1610: Thank you for picking this up @NexVeridian, it looks good! Here is a way to improve it before we merge it: users expect that when they re-run a single cell it will retain the height they set on the output. However, as cell gets re-run it gets back to 0 outputs for a brief moment so the height would be lost due to this conditional.

Could you store the value of `this.outputArea.node.style.height` in a private variable before clearing it and then if the number of outputs is greater than 0 restore it?

packages/cells/src/widget.ts
```diff
@@ -1603,6 +1604,14 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
     // This is to hide/show icon on single line output.
     this.updatePromptOverlayIcon();
+
+    // Clear output area when empty
+    this._lastOutputHeight = this.outputArea.node.style.height;
+    if (this.model.outputs.length === 0) {
+      this.outputArea.node.style.height = '';
```
krassowski, line 1611: I think the assignment needs to be inside the conditional as otherwise we would overwrite the cached value with `''` set after it was stored.

```suggestion
    if (this.model.outputs.length === 0) {
      this._lastOutputHeight = this.outputArea.node.style.height;
      this.outputArea.node.style.height = '';
```

packages/cells/src/widget.ts
```diff
@@ -1593,6 +1593,7 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
   }
 
+  private _lastOutputHeight = '';
```
krassowski, line 1596: Can you move this to around line 1667 where all other private variables are defined?

packages/cells/src/widget.ts
```diff
@@ -1603,6 +1603,14 @@ export class CodeCell extends Cell<ICodeCellModel> {
     }
     // This is to hide/show icon on single line output.
     this.updatePromptOverlayIcon();
+
+    // Clear output area when empty
+    if (this.model.outputs.length === 0) {
+      this._lastOutputHeight = this.outputArea.node.style.height;
```
krassowski, line 1609: I tested it locally and it was still not preserving the height when re-running a cell because this signal is emitted more than once. Also, it would force the previous height even if it was changed. I will push a commit to fix it.

krassowski: https://github.com/jupyterlab/jupyterlab/pull/16208/commits/50eda95fbb27ac52a72ca5b8feb054f1de97704a

krassowski (approved):
Thank you @NexVeridian!

Issue #16203, opened 2024-04-19
Unable to lock any extension


EnzoPB:
## Description
When using the `jupyter labextension lock extension:plugin` command, the concerned extension is not locked and can still be enabled/disabled in the plugin manager.

## Reproduce
- Disable an extension: `jupyter labextension disable @jupyterlab/application-extension:status` (this is an example, the behavior is the same for every extention)
- Lock this extension: `jupyter labextension lock @jupyterlab/application-extension:status`
- Restart Jupyter
- Open the advanced plugin manager in Lab, and search for the extension: it is disabled, but not locked and you can enable it
![image](https://github.com/jupyterlab/jupyterlab/assets/25568144/1f443c90-8a6d-457e-98ed-8084f40da89d)

## Expected behavior
The extension should be locked: an lock icon should be next to the checkbox, and the checkbox should be disabled.

## Context

- Operating System and version: Ubuntu 22.04.4 LTS (also tested with Windows 10)
- Browser and version: Firefox 123
- JupyterLab version: 4.1.6
- Tested with Python 3.12.2 & 3.10.12

The content of the page_config.json file looks correct after disabling & locking the extension:
```
$ cat <jupyter config dir>/labconfig/page_config.json 
{
  "disabledExtensions": {
    "@jupyterlab/application-extension:status": true
  },
  "lockedExtensions": {
    "@jupyterlab/application-extension:status": true
  }
}
```

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/home/enzo/jupyter_venv/bin
	/home/enzo/.nvm/versions/node/v21.7.2/bin
	/home/enzo/.local/bin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin
	/usr/games
	/usr/local/games
	/snap/bin
	/snap/bin
	/home/enzo/.local/share/JetBrains/Toolbox/scripts

sys.path:
	/home/enzo/jupyter_venv/bin
	/usr/lib/python310.zip
	/usr/lib/python3.10
	/usr/lib/python3.10/lib-dynload
	/home/enzo/jupyter_venv/lib/python3.10/site-packages

sys.executable:
	/home/enzo/jupyter_venv/bin/python3

sys.version:
	3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]

platform.platform():
	Linux-6.5.0-27-generic-x86_64-with-glibc2.35

[...]

PR #16213, opened 2024-04-20
Fix for unable to lock any extension

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/16203

This issue arises because the page_config lacks context regarding the locked plugins.


krassowski:
Looks good, but I think we need to add a test here. In particular something directly hitting the `get()` endpoint of PluginHandler API:

https://github.com/jupyterlab/jupyterlab/blob/f3867b04b57fb1f6f401c6ee36378204099de0ac/jupyterlab/handlers/plugin_manager_handler.py#L21

It would look similar to:

https://github.com/jupyterlab/jupyterlab/blob/f3867b04b57fb1f6f401c6ee36378204099de0ac/jupyterlab/tests/test_custom_css_handler.py#L21-L37

but using `lock_extension` and `unlock_extension` to setup the state, similarly to how disable/enable are used in:

https://github.com/jupyterlab/jupyterlab/blob/f3867b04b57fb1f6f401c6ee36378204099de0ac/jupyterlab/tests/test_jupyterlab.py#L553-L571

itsmevichu:
@krassowski I have added the tests, please have a look.

anxietymonger:
Any plan to include this patch in recent releases?

jupyterlab/labapp.py
```diff
@@ -729,6 +730,8 @@ def initialize_handlers(self):  # noqa
 
         # Set config for Jupyterlab
         page_config = self.serverapp.web_app.settings.setdefault("page_config_data", {})
+        page_config.update(get_static_page_config(logger=self.log, level="sys_prefix"))
```
krassowski, line 733: I believe this should also show plugins disabled at `user` and `system` level, right? 

```suggestion
        page_config.update(get_static_page_config(logger=self.log, level="all"))
```

krassowski (approved):
Thank you @itsmevichu!

PR #16214, opened 2024-04-20
Fix logic for changing keybindings in shortcut editor

### LINKED ISSUE #16211: Changing a second shortcut changes the first one in shortcuts editor
After `pip install -U --pre jupyterlab` on a Mac, adding a second shortcut does not work:

https://github.com/jupyterlab/jupyterlab/assets/44469195/06516d55-18ea-481c-aa3b-9a1ed89b579e

this was first reported in https://github.com/jupyterlab/jupyterlab/issues/16186#issuecomment-2067649537 and confirmed by @krassowski :

> I can reproduce this, it is a bug. I think it deserves its own issue as it is not Mac specific.

krassowski:
## References

Fixes #16211

## Code changes

- [x] Adds a test case
- [x] Adds missing check for keybinding keys equality with iterated shortcut
- [x] Fix shortcut/keybinding data being set at wrong level

## User-facing changes

Repalcing shortcuts in settings work when a mix of default and non-default shortcuts is used, and when multiple keybindings are used.

None

### DIFF
--- a/packages/shortcuts-extension/src/components/ShortcutItem.tsx
+++ b/packages/shortcuts-extension/src/components/ShortcutItem.tsx
@@ -219,12 +219,7 @@ export class ShortcutItem extends React.Component<
 
   getShortCutForDisplayOnly(binding: IKeybinding): JSX.Element[] {
     return binding.keys.map((keyboardKey: string, index: number) => (
-      <div
-        className="jp-Shortcuts-ShortcutKeysContainer"
-        key={index}
-        data-keybinding={index}
-        data-shortcut={this.props.shortcut.id}
-      >
+      <div className="jp-Shortcuts-ShortcutKeysContainer" key={index}>
         <div className="jp-Shortcuts-ShortcutKeys">
           {this.toSymbols(keyboardKey)}
         </div>
@@ -248,6 +243,8 @@ export class ShortcutItem extends React.Component<
       <div
         className="jp-Shortcuts-ShortcutContainer"
         key={this.props.shortcut.id + '_' + index}
+        data-keybinding={index}
+        data-shortcut={this.props.shortcut.id}
         onClick={() => this.toggleInputReplaceMethod(index)}
       >
         {this.isLocationBeingEdited(index)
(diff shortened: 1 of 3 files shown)


kolibril13:
Hi Mike,
thanks a lot for taking a look at this!
I've just tried this at https://mybinder.org/v2/gh/krassowski/jupyterlab/fix-changing-second-default-shortcut?urlpath=lab
In firefox, everything works fine, but safari looks quite broken.
There is a big magnifying glass, text is overlapping, and the second shortcut is not assigned correctly.

https://github.com/jupyterlab/jupyterlab/assets/44469195/8e957edb-6540-49ee-b2d4-35e644dd1443

krassowski:
> fine, but safari looks quite broken

From the recording, it appears that in Safari you end up clicking on the "Default" text because the accept button is moved to under the "Default". It appears that latest Safari switched to a new (and apparently not quite ready) [inline layout implementation](https://webkit.org/blog/15063/webkit-features-in-safari-17-4/#css) - is this the Safari version you are using?

> There is a big magnifying glass, text is overlapping, and the second shortcut is not assigned correctly.

Can you open a dedicated issue for magnifying glass? I think replacing the search component with jupyter-ui-toolkit one would solve the issue. Can you also check if this is happening in released versions already (in that issue).

kolibril13:
> It appears that latest Safari switched to a new (and apparently not quite ready) [inline layout implementation](https://webkit.org/blog/15063/webkit-features-in-safari-17-4/#css) - is this the Safari version you are using?
> 
> >

Yep, I'm using that new version -> Safari 17.4.1.


> Can you open a dedicated issue for magnifying glass? I think replacing the search component with jupyter-ui-toolkit one would solve the issue. Can you also check if this is happening in released versions already (in that issue).

I'll do that! :)

JasonWeill:
In Firefox 124.0.2 on macOS, with this code change, I see the expected behavior.

In Safari 17.4.1 on macOS, I see the UI component for a new shortcut overlap some other elements. After I click ✔️ , the new shortcut is _not_ saved.

image

krassowski:
> In Safari 17.4.1 on macOS, I see the UI component for a new shortcut overlap some other elements. After I click ✔️ , the new shortcut is _not_ saved.
> 
> image

This is a separate issue tracked in https://github.com/jupyterlab/jupyterlab/issues/16233

jtpio (approved):
Thanks!

Issue #16217, opened 2024-04-20
The label of toggle sidebar element command should be dynamic


krassowski:
## Description

Currently in 4.2.0b3 the `application:toggle-sidebar-widget` command uses a fixed label which is always "Toggle Sidebar Element".  This makes it difficult to identify which which sidebar element will be toggled.

## Reproduce

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/74286be3-460b-424b-b529-47d70b65d891)

## Expected behavior

Maybe should show something like: "Toggle 2nd Element in Right Sidebar" instead?

## Context

- JupyterLab version: 4.2.0b2

krassowski:
The relevant code is in:

https://github.com/jupyterlab/jupyterlab/blob/2dd8f4ebfdf604f61dd3cc1e4f6da14547e89833/packages/application-extension/src/index.tsx#L373-L379

The `label` can be a function taking args the same way as `execute` does.

The strings will need to be created using the translation formatting as documented in https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html#internationalization-and-localization

PR #16218, opened 2024-04-21
Dynamic label for toggle sidebar element in ketboard shortcuts

itsmevichu:
Fix for https://github.com/jupyterlab/jupyterlab/issues/16217

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -377,8 +377,14 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
           args.index === undefined
             ? trans.__('Toggle Sidebar Element')
             : args.side === 'right'
-            ? trans.__('Toggle Element %1 in Right Sidebar', args.index)
-            : trans.__('Toggle Element %1 in Left Sidebar', args.index),
+            ? trans.__(
+                'Toggle Element %1 in Right Sidebar',
+                parseInt(args.index as string, 10) + 1
+              )
+            : trans.__(
+                'Toggle Element %1 in Left Sidebar',
+                parseInt(args.index as string, 10) + 1
+              ),
         execute: args => {
           const index = parseInt(args.index as string, 10);
           if (args.side != 'left' && args.side != 'right') {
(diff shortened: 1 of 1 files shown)


packages/application-extension/src/index.tsx
```diff
@@ -371,7 +371,10 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
       });
 
       commands.addCommand(CommandIDs.toggleSidebarWidget, {
-        label: trans.__('Toggle Sidebar Element'),
+        label: args =>
+          args.side == 'right'
+            ? trans.__('Toggle Element %1 in %2 Sidebar', args.index, 'Right')
+            : trans.__('Toggle Element %1 in %2 Sidebar', args.index, 'Left'),
```
krassowski, line 377: We need to embed "Right' (and 'Left') in the text as otherwise these would not get translated (argument are not translated).

```suggestion
            ? trans.__('Toggle Element %1 in Right Sidebar', args.index)
            : trans.__('Toggle Element %1 in Left Sidebar', args.index),
```

krassowski (commented):
Thanks for starting this! Two suggestions:
- the call to `trans.__` needs a slight adjustment as arguments do not get translated
- the existing label should be kept as a fallback when args are not given because a command label can be also queried without arguments.

packages/application-extension/src/index.tsx
itsmevichu: I have adapted the review changes.

packages/application-extension/src/index.tsx
```diff
@@ -371,7 +371,14 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
       commands.addCommand(CommandIDs.toggleSidebarWidget, {
-        label: trans.__('Toggle Sidebar Element'),
+        label: args =>
+          args === undefined ||
+          args.side === undefined ||
+          args.index === undefined
+            ? trans.__('Toggle Sidebar Element')
+            : args.side === 'right'
+            ? trans.__('Toggle Element %1 in Right Sidebar', args.index)
+            : trans.__('Toggle Element %1 in Left Sidebar', args.index),
```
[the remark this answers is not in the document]
itsmevichu: Adapted. 👍

krassowski (approved):
Thank you @itsmevichu!

PR #16219, opened 2024-04-22
Forbid explicit `any`, migrate to `unknown` or proper typing

krassowski:
# References

https://github.com/jupyterlab/jupyterlab/issues/16038

## Code changes

Forbids any, except in variadics.

The intent is to:
- keep `any` in core APIs to avoid disruption (just add `eslint-disable` comments as needed)
- use `unknown` in leaf APIs and in function bodies where we can be fairly certain it will not cause disruption

TODO:
- [x] this PR and this rule should not apply to test files to reduce the diff and because we worry less about typing in tests

## User-facing changes

None

TBD


packages/services/examples/browser/src/index.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
-(window as any).__webpack_public_path__ = URLExt.join(
+(window as unknown).__webpack_public_path__ = URLExt.join(
```
krassowski, line 9: this will need reverting

krassowski:
Closing in favour of two-staged approach:
- https://github.com/jupyterlab/jupyterlab/pull/18847 (merged)
- https://github.com/jupyterlab/jupyterlab/issues/18854 (to be done)

Issue #16220, opened 2024-04-22
Translation wrapper is missing around debugger filter box


krassowski:
## Description

The placeholder is not translated:

https://github.com/jupyterlab/jupyterlab/blob/8819549495cd11470b4db7464d1d59c53cfce787/packages/debugger/src/panels/kernelSources/filter.tsx#L27

See: https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html

fcollonval:
Closed by https://github.com/jupyterlab/jupyterlab/pull/16224

PR #16224, opened 2024-04-22
Add missing translation wrapper in debugger filter box

itsmevichu:
Fix for https://github.com/jupyterlab/jupyterlab/issues/16220


packages/debugger/src/panels/kernelSources/filter.tsx
```diff
@@ -8,23 +8,30 @@ import React from 'react';
...
   model: IDebugger.Model.IKernelSources;
+
+  /**
+   * The application language translator
+   */
+  translator?: ITranslator;
 }
 
 const FilterBox = (props: IFilterBoxProps) => {
+  const trans = (props.translator ?? nullTranslator).load('jupyterlab');
```
krassowski, line 26: Tiny optimization suggestion: to avoid loading the translator again you could instead pass the translation bundle (`trans`) in the props.

krassowski (commented):
Please see a small suggestion on passing the `trans` instead of the translator object - this should also reduce the changeset.

Looks good otherwise:

Screenshot from 2024-04-25 11-57-35

packages/debugger/src/panels/kernelSources/filter.tsx
itsmevichu: Adapted the comments.

krassowski (approved):
Thank you @itsmevichu!

packages/debugger/src/panels/kernelSources/filter.tsx
```diff
@@ -8,12 +8,18 @@ import React from 'react';
...
+import { TranslationBundle } from '@jupyterlab/translation';
 
 /**
  * The class name added to the filebrowser crumbs node.
  */
 export interface IFilterBoxProps {
   model: IDebugger.Model.IKernelSources;
+
+  /**
+   * The application language translator
```
krassowski, line 20:
```suggestion
   * The language bundle.
```

Issue #16060, opened 2024-03-25
Dirty dot goes away after moving a modified cell


JasonWeill:
## Description

The dirty dot, which appears to the left of a cell that has been modified since it was last executed, disappears when a cell is moved.

## Reproduce

First, create three code cells, and run all of them.

![image](https://github.com/jupyterlab/jupyterlab/assets/93281816/0d9a2cf0-ca29-400c-b138-0fd280c4ec49)

Now, select the second cell, and modify it. Do not run it.

![image](https://github.com/jupyterlab/jupyterlab/assets/93281816/79ae4f3c-5c22-44cc-b2f7-45a9985b25bf)

Use drag and drop to move the second cell below the third cell. Note that the dirty dot and orange color disappear.

![image](https://github.com/jupyterlab/jupyterlab/assets/93281816/51b54db9-2ea5-4258-abfb-c43c159b2081)

## Expected behavior

The second cell, modified, still has the orange color and dirty dot to indicate that it has been modified since it was last run.

## Context

- Operating System and version: macOS 13.6.4 "Ventura"
- Browser and version: Firefox ESR 123.0.1
- JupyterLab version: 4.1.5

fcollonval:
To provide some background, the dirty indicator is a in-memory state comparing the last executed snippet with the current cell content. When moving a cell, we actually delete the old cell and insert a new one (limitation of YJs the [move feature is WIP](https://github.com/yjs/yjs/pull/357) - stuck?). So the in-memory state is lost, hence the current behavior.

kguptas:
Hello - I'm a student trying to contribute. I was looking into this and trying to recreate the issue; however, I don't see a dirty dot when editing even before moving. Is this the expected behavior?

krassowski:
@kguptas no, its not. What version of JupyterLab do you have installed?

kguptas:
@krassowski 4.2.0b0 after running jupyter lab --version. I tried again and it seem to work. For some reason initially it wasn't working for me. Which file should I look to to begin debugging this?

krassowski:
My hunch is that this method may require modification to preserve the dirty state:

https://github.com/jupyterlab/jupyterlab/blob/9aca629a98a1518edcfbca886b0ba7d78e49261a/packages/notebook/src/widget.ts#L378-L422

For reference the dirty state of a cell is defined here:

[...]

PR #16225, opened 2024-04-22
Fix dirty dots not preserved when moving multiple cells

Alanhou1222:
## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/16060
Closes https://github.com/jupyterlab/jupyterlab/pull/16161 (includes)

## Code changes
In packages/cells/src/model.ts, make the setDirty() function into a set function set isDirty().
In packages/notebook/src/widget.ts, move logics that set new cells dirty into the for loop of the moveCell function and make the code more clear.

## User-facing changes
dirty dots are kept even when moving multiple cells
image
image

None


Alanhou1222:
@krassowski Not sure if it's expected, but moved cells won't remember their content before being moved. Therefore, even if users revert the changes on a moved cell, the cell is still considered dirty.

krassowski:
> but moved cells won't remember their content before being moved

Indeed, undo does not work after moving a cell - good catch! This is not expected, we should open a separate issue to track it and then fix it.

This PR looks good to me on the implementation side (the failing Python tests are unrelated); ideally we would also include unit tests in `packages/notebook/test/widget.spec.ts` for the dirty dot being preserved. Would you be interested in adding such a unit test (whether in this PR or in another)?

Alanhou1222:
> > but moved cells won't remember their content before being moved
> 
> Indeed, undo does not work after moving a cell - good catch! This is not expected, we should open a separate issue to track it and then fix it.
> 
> This PR looks good to me on the implementation side (the failing Python tests are unrelated); ideally we would also include unit tests in `packages/notebook/test/widget.spec.ts` for the dirty dot being preserved. Would you be interested in adding such a unit test (whether in this PR or in another)?

Sure, I'll write some unit tests.

Alanhou1222:
@krassowski I added two tests, one verifies dirty state is preserved after moving one code cell and another verifies dirty states are preserved after moving multiple cells. Let me know if you think there are other situations that should be tested.

packages/notebook/src/widget.ts
```diff
@@ -418,6 +424,17 @@ export class StaticNotebook extends WindowedList<NotebookViewModel> {
...
+
+      if (from > to) {
+        if (this.widgets[to + i].model.type === 'code') {
+          (this.widgets[to + i].model as CodeCellModel).isDirty = dirtyState[i];
+        }
+      } else {
+        if (this.widgets[to + i - n + 1].model.type === 'code') {
+          (this.widgets[to + i - n + 1].model as CodeCellModel).isDirty =
+            dirtyState[i];
+        }
```
krassowski, line 436: I wonder why is it that `rendered` and `headingCollapsed` do not need `- n + 1` but `isDirty` needs it? It seems a bit suspicious to me that the implementation here differs from the one for markdown cell properties. I suspect that one of those is wrong.

krassowski (approved):
Tested locally, works great. Thank you @Alanhou1222!

PR #16230, opened 2024-04-23
Use the application name instead of hard-coded "JupyterLab" in dialogs/commands

sawickid:
## References

https://github.com/jupyter/notebook/issues/7188 
"File > Shutdown in classic notebook ask if I want to shutdow JupyterLab"

## Code changes

Edits shutdown dialogue body to ask if the user wants to shutdown <app.name>, instead of exclusively 'JupyterLab' 

User will be prompted to confirm shutdown of Jupyter Notebook, instead of JupyterLab, when using Jupyter Notebook


packages/mainmenu-extension/src/index.ts
```diff
@@ -375,7 +375,9 @@ function createFileMenu(
     execute: () => {
       return showDialog({
         title: trans.__('Shutdown confirmation'),
-        body: trans.__('Please confirm you want to shut down JupyterLab.'),
+        body: trans.__(
+          'Please confirm you want to shut down ' + app.name + '.'
+        ),
```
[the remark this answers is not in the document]
krassowski: +:1+ thank you for reviewing this!

For reference: https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html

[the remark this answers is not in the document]
sawickid: Thank you for the feedback! Making the changes now.

packages/mainmenu-extension/src/index.ts
```diff
@@ -369,13 +369,13 @@ function createFileMenu(
 
   commands.addCommand(CommandIDs.shutdown, {
     label: trans.__('Shut Down'),
-    caption: trans.__('Shut down JupyterLab'),
+    caption: trans.__('Shut down "%1"', app.name),
```
Carreau, line 372: I think you took the examples a bit too literally. 

```
caption: trans.__('Shut down "%1"', app.name),
```

Will render as `Shut down "JupyterLab"` / `Shut down "Notebook"`; only the `%1` will be replaced. I think we want without the quotes in the various places. But a core dev can confirm.

Otherwise this seem good to me,

krassowski: I agree with the above.

krassowski:
```suggestion
    caption: trans.__('Shut down %1', app.name),
```

packages/mainmenu-extension/src/index.ts
```diff
@@ -369,13 +369,13 @@ function createFileMenu(
...
     isVisible: () => menu.quitEntry,
     isEnabled: () => menu.quitEntry,
     execute: () => {
       return showDialog({
         title: trans.__('Shutdown confirmation'),
-        body: trans.__('Please confirm you want to shut down JupyterLab.'),
+        body: trans.__('Please confirm you want to shut down "%1".', app.name),
```
krassowski, line 378:
```suggestion
        body: trans.__('Please confirm you want to shut down %1.', app.name),
```

packages/mainmenu-extension/src/index.ts
```diff
@@ -412,7 +412,8 @@ function createFileMenu(
                 );
                 const p2 = document.createElement('p');
                 p2.textContent = trans.__(
-                  'To use JupyterLab again, you will need to relaunch it.'
+                  'To use "%1" again, you will need to relaunch it.',
```
krassowski, line 415:
```suggestion
                  'To use %1 again, you will need to relaunch it.',
```

packages/mainmenu-extension/src/index.ts
```diff
@@ -437,7 +438,7 @@ function createFileMenu(
 
   commands.addCommand(CommandIDs.logout, {
     label: trans.__('Log Out'),
-    caption: trans.__('Log out of JupyterLab'),
+    caption: trans.__('Log out of "%1"', app.name),
```
krassowski, line 441:
```suggestion
    caption: trans.__('Log out of %1', app.name),
```

krassowski (approved):
Thank you @sawickid, kooks good to me.

Issue #13993, opened 2023-02-13
strange styling of "Jupyter news" notification


dlqqq:
## Description

The "official Jupyter news" notification styling is strange; link feels like it should belong in the notification body rather than as a separate action button, and the "Please read privacy policy" sentence is preceded by a line break that causes the notification to occupy more vertical space than necessary. Finally, there is a noticeable lack of margin between the notification body and the action buttons below.

<img width="442" alt="Screen Shot 2023-02-13 at 2 41 45 PM" src="https://user-images.githubusercontent.com/44106031/218597113-ca4e9f57-95f7-4902-9e78-f49e24c0c8db.png">

## Reproduce

1. Open JupyterLab and click on the bottom right notifications icon.

## Expected behavior

1. "Please read the [privacy policy](google.com)." should not be preceded by a newline, and should include a direct link to the privacy policy as shown in the rendered markdown.
2. Remove the "Open privacy policy" action button.
3. Add more margin between notification body and its action buttons. 

## Context

- Operating System and version: OS X Monterey
- Browser and version: Chrome 109.0.5414.119 (Official Build) (x86_64)
- JupyterLab version: [4.0.0a33 binder](https://gist.github.com/jtpio/4e9a3d435d7c3950dc4150c10abbb1d7)

JasonWeill:
This only happens when clicking on the notification UI in the bottom-right corner; on initial load, only the inner notification is visible.

fcollonval:
FYI Links cannot be part of the notification text as for security reasons, it was decided notification will be displayed as plain text. Link styled button are common in UI framework, it is used here to highlight that the button will take you to an external URL.

This was implemented in https://github.com/jupyterlab/jupyterlab/pull/13510

JasonWeill:
I also find the "1 notification" styling to be unusual; the outer container has no outer border, while the notification inside the container has a drop shadow.

JasonWeill:
My comment above is already reflected in #13992.

JasonWeill:
There is now a bit of vertical space between the message and the action buttons:

<img width="334" alt="Screenshot 2024-04-24 at 3 58 11 PM" src="https://github.com/jupyterlab/jupyterlab/assets/93281816/ff0d428c-417b-456e-ac1c-d71094826913">

JasonWeill:
Fixed by #16236.

PR #16236, opened 2024-04-24
Update notification box styling, content

### LINKED ISSUE #13992: strange styling in "Notifications" popup
## Description

When clicking the "Notifications" bell on the bottom right, the resultant popup lacks a border and shadow to distinguish it from the background. This hinders usability of the popup as it is not immediately obvious where the popup ends.

Screen Shot 2023-02-13 at 2 41 45 PM

## Reproduce

1. Open JupyterLab and click on the bottom right notifications icon.

## Expected behavior

The notifications popup should have a border and shadow to distinguish it from the background.

## Context

- Operating System and version: OS X Monterey
- Browser and version: Chrome 109.0.5414.119 (Official Build) (x86_64)
- JupyterLab version: [4.0.0a33 binder](https://gist.github.com/jtpio/4e9a3d435d7c3950dc4150c10abbb1d7)

JasonWeill:
## References

Fixes #13992 and #13993.

## Code changes

Updates the content of the "attempting to save read-only file" notification to match Jupyter styling with copy edits. Updates the CSS to put a small border around a notification display and to modify the padding for the header.

Adds another newline to the Jupyter news notification that appears on first startup, providing additional separation between the question and a clarifying request.

## User-facing changes

Before the change:

Screenshot 2024-04-24 at 3 41 17 PM

Screenshot 2024-04-24 at 3 58 11 PM

After the change:

Screenshot 2024-04-25 at 10 05 45 AM

Screenshot 2024-04-25 at 10 05 32 AM

None.


packages/docregistry/src/components.tsx
```diff
@@ -23,7 +23,7 @@ export function createReadonlyLabel(
       <span
         className="jp-ToolbarLabelComponent"
         title={trans.__(
-          `Document is permissioned read-only; "save" is disabled, use "save as..." instead`
+          `Document is read-only. "Save" is disabled; use "Save as…" instead`
```
chanhee-park, line 26: Why use semicolons? Wouldn’t the punctuation be neater? 
```
`Document is read-only. "Save" is disabled. Use "Save as…" instead`
```

packages/apputils-extension/src/announcements.ts
```diff
@@ -100,7 +100,7 @@ export const announcements: JupyterFrontEndPlugin<void> = {
       if (mustFetchNews === 'none') {
         const notificationId = Notification.emit(
           trans.__(
-            'Would you like to receive official Jupyter news?\nPlease read the privacy policy.'
+            'Would you like to receive official Jupyter news?\n\nPlease read the privacy policy.'
```
krassowski, line 103: Some thoughts:
- there are numerous complaints about this popup. This change will make it even larger.
- do we need the "Please read the privacy policy" line at all? There is a link to privacy policy below anyways.
- "receive" is correct but for me it has a connotation of something more heavy weight (e.g. receive (e)mail). I feel like "get notified of" might fit here better?

```suggestion
            'Would you like to get notified official Jupyter news?'
```

krassowski:
The added border looks great!

packages/docregistry/src/components.tsx
JasonWeill: I'm not sold that this is necessarily neater. The semicolon links two closely related statements, so I think it is still appropriate to keep it in this notification.

krassowski (dismissed):
Thank you @JasonWeill! Good to merge once green.

krassowski (approved):
Should be good now :)

PR #16237, opened 2024-04-24
Harmonize the use of ellipses in command and menu items

Rob-P-Smith:
Audited all menu-bar options for required ellipses or removal of ellipses.

As mentioned in the commit, a few items required ellipses at certain times while not requiring them at other times. Those dual nature elements were not adjusted.

## References

#16025 

## Code changes

/packages/notebook-extension/schema/export.json line 16 removed ellipses from label
/packages/notebook-extension/index.ts line 2420 added ellipses to label
/packages/apputils-extension/index.ts line 653 added ellipses to label

## User-facing changes

Adjusted ellipses to conform with the guidance posted in the issue.
Added ellipses to Show Keyboard Shortcuts under help menu.
Added ellipses to Close and Shut Down Notebook under file menu.
Removed ellipses on Save and Export Notebook As under file menu.

Updated areas:
three
two
One

None


JasonWeill:
Several of the test cases now fail, because the menus are different. You can comment with `please update snapshots` when you're ready to regenerate both galata and documentation screenshots.

.gitignore
```diff
@@ -148,3 +148,4 @@ packages/lsp/src/_*
 # generated reports
 webpack-bundle-analyzer.html
 examples/**/test-results/
+docker/start.sh
```
JasonWeill, line 151: This file isn't relevant to this change. Can you please remove this line from `.gitignore`, and remove the empty `docker/start.sh` file from your branch?

packages/apputils-extension/src/index.ts
```diff
@@ -650,7 +650,7 @@ const utilityCommands: JupyterFrontEndPlugin<void> = {
     });
 
     commands.addCommand(CommandIDs.displayShortcuts, {
-      label: trans.__('Show Keyboard Shortcuts'),
+      label: trans.__('Show Keyboard Shortcuts...'),
```
JasonWeill, line 653: Please use the ellipsis character …, which looks very similar to three periods, but which is not the same.

* Ellipsis: `…`
* Three periods: `...`

JasonWeill (commented):
"Save Current Workspace…" should not have an ellipsis, because if the workspace already has a name, this command does not require user attention.

If the current workspace has not yet been saved, the command will call "Save Current Workspace As…", which has an ellipsis, and should keep it.

.gitignore
Rob-P-Smith: I will attempt to do so, it was not intended to be in the commit.

packages/apputils-extension/src/index.ts
Rob-P-Smith: Thanks, i can redo the changes, they're small to fix this.

Rob-P-Smith:
Thank you for your patience Jason, trying not to make life harder on others while I learn.

I think I've satisfied your feedback.

krassowski:
The failure in:

```
    [jupyterlab] › test/jupyterlab/kernel.test.ts:27:9 › Kernel › Notebook › Should remember kernel auto start for notebook 
```

appears relevant, the test will need updating.

packages/notebook-extension/src/index.ts
```diff
@@ -2417,7 +2417,7 @@ function addCommands(
     isEnabled
   });
   commands.addCommand(CommandIDs.closeAndShutdown, {
-    label: trans.__('Close and Shut Down Notebook'),
+    label: trans.__('Close and Shut Down Notebook…'),
```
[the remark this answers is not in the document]
krassowski: Ah because it shows a confirmation dialog.

krassowski (approved):
Thank you!

PR #16246, opened 2024-04-26
Allow to select the path in debugger source component

### LINKED ISSUE #16235: Debugger: Unable to Copy/Paste Paths
### Problem
It's hard to copy paths from the debugger.

Things tried:
1. Selecting the text from here is not possible 
image

2. Once you open a file, the right click->show in file browser does not work

### Proposed Solution

Make option 1 above work. 2 may be possible in some cases but not all. In cases where it is, it would be nice if that worked too.

krassowski:
## References

Fixes #16235

## Code changes

1. Make the source path user-selectable
2. Do not open the read-only editor on click as it steals focus and removes selection; instead make it open on <kbd>Ctrl</kbd> + click

## User-facing changes

Clicking on the source path now allows selection and thus copying the file

image

<kbd>Ctrl</kbd> + click will open the main area widget:

image

Note: user can still click on the button next to the path to open the source without holding <kbd>Ctrl</kbd>:

image

None


jtpio:
Adding to `4.3.0` since it's tagged as https://github.com/jupyterlab/jupyterlab/labels/enhancement

jtpio (approved):
Thanks, diff looks good!

Waiting for the CI to be done. Maybe some snapshots will need to be updated?

krassowski:
Looks like its passing after restart :) Let's hold of with merging here until we release 4.2.0 (today!)

PR #16251, opened 2024-04-26
Consider higher levels when toggling plugin

### LINKED ISSUE #16101: An extension disabled at the sys-prefix level can't be enabled at the user level

divyansshhh:
Fixes #16101

### DIFF
--- a/jupyterlab/commands.py
+++ b/jupyterlab/commands.py
@@ -1135,24 +1135,29 @@ class _AppHandler:
                 self.logger.info("Extension locked at a higher level, cannot toggle status")
                 return False
 
-        page_config = get_static_page_config(
+        complete_page_config = get_static_page_config(
+            app_settings_dir=app_settings_dir, logger=self.logger, level="all"
+        )
+
+        level_page_config = get_static_page_config(
             app_settings_dir=app_settings_dir, logger=self.logger, level=level
         )
 
-        disabled = page_config.get("disabledExtensions", {})
+        disabled = complete_page_config.get("disabledExtensions", {})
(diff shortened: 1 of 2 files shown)


divyansshhh:
This was the initial commit made for jupyterlab/jupyterlab#16102 ([commit link](https://github.com/jupyterlab/jupyterlab/pull/16102/commits/85ec8135e34bb64ac34ca1e751ff47b3ed4d1b0c)).

I think I missed redoing this change while I was adding the _is_locked checks. 

@krassowski can we release another patch with this since this is a blocking change for us? Apologies for the inconvenience.

krassowski:
It looks like having a test sequence in the usage test would help here. Do you have time to add it?

divyansshhh:
> I can wait a bit with releasing 4.2.0rc0 if that works for you

4.2 has a couple of changes we want to take time and test so would want to do thorough testing at our end before we upgrade. A 4.1 patch would be quicker to test. 

> We try not to release too frequently unless there are critical or security bugfixes because users get annoyed by frequent notifications about updates being available.

Yeah I understand but it'd be helpful to release another patch today.

divyansshhh:
> It looks like having a test sequence in the usage test would help here. Do you have time to add it?

Yes I'm adding this

krassowski (approved):
Thank you @divyansshhh!

krassowski:

PR #16252, opened 2024-04-26
Add an option to continue indentation, upgrade `@codemirror/commands`

### LINKED ISSUE #16196: Issue with Python indented structures in JupyterLab
## Description

Consider the following snippet -
```python
def fn():
    print("hi")
    for i in range(10):
        print("hello")
```
If the cursor is at the end of the last line and enter is pressed twice, the user is sent outside the function definition which is unintuitive. Attaching a gif demonstrating this - 
enter-indentation

This happens even when defining functions in a class. 

## Reproduce

<see gif>

## Expected behavior
I would expect the behavior to be the same as lab3/vscode. 

vscode-enter-indentation

## Context

- JupyterLab version: 4.1.5

krassowski:
## References

Closes #16196

## Code changes

- [x] update `@codemirror/commands`
- [x] add an option to continue indentation
- [ ] add a test

## User-facing changes

New continue indentation option:

image

TBD, as the current behaviour of `insertNewlineKeepIndent` turned out to be different from what I thought it would do, see https://github.com/codemirror/dev/issues/1370#issuecomment-2079392856.

Depending on whether upstream (CodeMirror) reconsiders, we may need to do something different, for example always run `insertNewlineAndIndent` while comparing the indentation before and after and if it decreased at the end of the block we would re-add the indentation. Or maybe we need to vendor a modified copy of `insertNewlineAndIndent` which does not dedent at the end of the block (not sure if this is feasible). Or maybe we need to modify the Python language implementation only in which case this PR would be of little use in the current state.

None

### DIFF
--- a/packages/codemirror-extension/package.json
+++ b/packages/codemirror-extension/package.json
@@ -37,7 +37,7 @@
     "watch": "tsc -b --watch"
   },
   "dependencies": {
-    "@codemirror/commands": "^6.3.3",
+    "@codemirror/commands": "^6.5.0",
     "@codemirror/lang-markdown": "^6.2.4",
     "@codemirror/language": "^6.10.1",
     "@codemirror/legacy-modes": "^6.3.3",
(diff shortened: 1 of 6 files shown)


krassowski:
Closing in favour of https://github.com/jupyterlab/jupyterlab/pull/16306

PR #16260, opened 2024-04-28
fixing search and replace with cell outputs

### LINKED ISSUE #15966: Enable "Search Cell Outputs" in Find/Replace, disable "Replace" button when an output is highlighted

bruno-neira:
## References

Builds from #16157
Resolves #15966 

## Code changes

- updated highlighting logic in genericsearchprovider.ts to support highlighting in cell outputs
- updated searchmodel.ts to support disabling replace button when current match is in cell output

## User-facing changes

- highlights search matches even if matches are in cell outputs
- disables the replace button when the current match is in the cell output.

### DIFF
--- a/packages/documentsearch/src/providers/genericsearchprovider.ts
+++ b/packages/documentsearch/src/providers/genericsearchprovider.ts
@@ -11,7 +11,7 @@ import { SearchProvider } from '../searchprovider';
 import { ITranslator } from '@jupyterlab/translation';
 
 export const FOUND_CLASSES = ['cm-string', 'cm-overlay', 'cm-searching'];
-const SELECTED_CLASSES = ['CodeMirror-selectedtext'];
+const SELECTED_CLASSES = ['CodeMirror-selectedtext', 'jp-current-match'];
 
 /**
  * HTML search engine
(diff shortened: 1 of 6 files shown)


BrunoNeira-cmu:
Currently draft PR because I'm working on disabling the Replace All button when all matches in are in cell outputs.

krassowski:
@BrunoNeira-cmu thank you for working on it! Since it has been a while I just wanted to check if you are still interested in finishing this PR?

krassowski:
@BrunoNeira-cmu would you like any help to finish this PR? We could still merge it in 4.3 merge window if we can get it finalised in next two weeks.

krassowski:
Thank you for submitting this PR - given inactivity here and a newer PR https://github.com/jupyterlab/jupyterlab/pull/18051 I will close this one.

Issue #16182, opened 2024-04-17
Increase/Decrease Text Editor Font Size not working


matthewgasbarro:
## Description

Increase/Decrease Text Editor Font Size is not working and is a no-op

## Reproduce

1. Open a Text Editor in either JupyterLab 4 or Notebook 7
2. Type in some text
3. Click on Settings -> Increase/Decrease Text Editor Font Size
4. See that no font resizing occurs

Browser console will show following JavaScript error
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'fontSize')
    at execute (commands.js:141:77)
    at c.execute (index.es6.js:364:33)
    at p.triggerActiveItem (index.es6.js:6365:27)
    at p._evtMouseUp (index.es6.js:6707:14)
    at p.handleEvent (index.es6.js:6510:22)
```

![chrome_PtBVZmiGKp](https://github.com/jupyterlab/jupyterlab/assets/11775352/ce9fba98-bce1-498c-9973-b544ff6a6355)

## Expected behavior

Text should resize accordingly

## Context

- Operating System and version: Windows 10
- Browser and version: Chrome Version 123
- JupyterLab version: 4.1.5
- Notebook version: 7.1.2

krassowski:
This seems to be coming from this line:

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/packages/fileeditor-extension/src/commands.ts#L261-L264

It was not caught by type checking because config is typed as `any`:

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/packages/fileeditor-extension/src/commands.ts#L147

which is another argument for forbidding `any` (https://github.com/jupyterlab/jupyterlab/issues/16038).

And we will need to add a test case for these commands.

janengcodes:
Hi, we have been trying to solve this issue. When trying to reproduce the error, we were getting uncaught error saying that fontSize was undefined. However, after adding console.log statement in the code sections referencing fontSize (in commands.ts) we were not able to see any of our console.logs and were wondering if its possible that fontSize is being referenced somewhere else, which is causing the error to be thrown. When trying to trace the error in the console, it was unable to point us to a file. We tried catching the error as well in commands.ts but it had no effect. Are there any other suggestions of where this uncaught error may be originating from?

krassowski:
[...]

PR #16261, opened 2024-04-28
Fix changing font size in text editor

FoSuCloud:
## References

Fixes: https://github.com/jupyterlab/jupyterlab/issues/16182
## Code changes

packages/fileeditor-extension/src/commands.ts
截屏2024-04-28 21 47 53

## User-facing changes

no

no


galata/test/jupyterlab/texteditor.test.ts
```diff
@@ -103,4 +103,42 @@ ut elit.`
     const tabHandle = await page.activity.getPanelLocator(DEFAULT_NAME);
     expect(await tabHandle?.screenshot()).toMatchSnapshot(imageName);
   });
+
+  test.describe('Changing a text editor font-size', () => {
```
fcollonval, line 107: Thanks a lot for adding a test for this :heart:

fcollonval (approved):
Thanks @FoSuCloud 

I spot a small glitch that the gutter line number is not updated right away when changing the font size; example when increasing the font size

image

CodeMirror does update the font size of the gutter if I switch back and forth to fileeditor. I guess forcing an update after update the font size may smooth the user experience.

FoSuCloud:
> Thanks @FoSuCloud
> 
> I spotted a small glitch that the gutter line number is not updated right away when changing the font size; example when increasing the font size
> 
> image
> 
> CodeMirror does update the font size of the gutter if I switch back and forth to the file editor. I guess forcing an update after updating the font size may smooth the user experience.

Thanks for knowing the problem, I will force update the number of lines after updating the font size

FoSuCloud:
It seems that the "should Decrease Content Font Size" and "should Increase Content Font Size" test cases failing have nothing to do with the pr here

krassowski:

PR #16262, opened 2024-04-29
Set aria-label to title if no label for

### LINKED ISSUE #16263: Icon toolbar button has no accessibility name
## Description

Toolbar buttons without label have no accessibility name because the title is set in the container web-component and not in the element with role button within the element.

## Reproduce

Look at the accessibility tree for a toolbar icon button in a web browser - here the file browser refresh button in Firefox for Lab 4.1.2

image

## Expected behavior

The element with role button is named using the title when the label is unset.

## Context

- Operating System and version: Debian
- Browser and version: Firefox 115
- JupyterLab version: 4.1.2

fcollonval:
## References

Fixes #16263 

## Code changes

Set explicitly the aria-label from the title when no label is defined on a toolbar button

## User-facing changes

| Before (4.1.2) | After |
| --- | --- |
|image |image |

None

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -843,6 +843,9 @@ export function ToolbarButtonComponent(
     }
   };
 
+  const title = getTooltip();
+  const disabled = props.enabled === false;
+
   return (
     <Button
       appearance="stealth"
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

jtpio:

PR #16265, opened 2024-04-29
Add checkbox to skip showing the kernel restart dialog

### LINKED ISSUE #6420: Add command to allow kernel-restart without confirmation dialog
When working with notebooks in a development-context, perhaps coding underlying APIs and libraries, it is often desirable to completely restart the running kernel, so that all changed code is loaded and the context is cleared.

Currently when doing that, a confirmation dialog has to be accepted each time.  In development mode it would be useful if such a dialog could be bypassed, i.e. kernel-restart with one click or keystroke.

### LINKED ISSUE #8021: Warning message when restarting Kernel
Select 'Kernel'
Choose either:
- Restart Kernel ...
- Restart Kernel and Clear Outputs ...
- Restart Kernel and Run All Cells ...

Result:
Capture

Expected result:
Do not show the warning message.

I find that restarting the Kernel (and rerunning all code), is a frequent action when working with a notebook. After iterating back and forth, you want to restart with a clean slate once in a while.

As it's a frequent and common action, it believe this does not really require a warning message.
Furthermore, restarting the Kernel is not a destructive action. All code is preserved, and can be recreated by a simple rerun.
There is also the 'checkpoints' feature, if you want to revisit older code and/or results.

Can this be disabled in settings? (i've not found any option so far)
How do others feel about disabling this message by default?

NexVeridian:
## References

closes #8021 and closes #6420

## Code changes

Add checkbox to skip showing the kernel restart dialog see https://github.com/jupyterlab/jupyterlab/issues/8021#issuecomment-1926819764
## User-facing changes

https://github.com/jupyterlab/jupyterlab/assets/46305192/f4d52968-4219-460b-b973-574c876c9cf0

none

## Thing that might need to be added
https://github.com/jupyterlab/jupyterlab/issues/6420#issuecomment-1238451259
>
    a) [x] Do not ask again in this session: restart dialogs silenced until user reloads JupyterLab
    b) [x] Do not ask again for this notebook: restart dialogs silenced for this specific notebook until user reloads JupyterLab
    c) [x] Do not show this dialog ever again: disables the dialog altogether, user needs to go to settings to re-enable it

this pr does option b, hiding the dialogs is per file, and is reset every session

to convert to option c, the checkbox value would need to be persisted and added to settings like `Automatically Start Preferred Kernel` is, I don't think I'm going to able to implement that though, this is might good enough for now


fcollonval (commented):
Thanks a lot @NexVeridian 

Would you feel comfortable implementing option (c) - never display the dialog through a setting instead of per session that cannot be undone?

To achieve that you will need to extend `ISessionContext.IDialogsOptions` to accept the `ISettingRegistry` token. And then you'll be able to do something like there:

https://github.com/jupyterlab/jupyterlab/blob/82051a9d0c6dc93677b0b66560aea48d71aaa043/packages/docmanager-extension/src/index.tsx#L905-L942

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1421,9 +1420,12 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
     }
 
     // Skip the dialog and restart the kernel
-    const skipKernelRestartDialog =
-      sessionContext.kernelPreference?.skipKernelRestartDialog ?? false;
-    if (skipKernelRestartDialog) {
+    const kernelPluginId = '@jupyterlab/notebook-extension:kernel-status';
```
NexVeridian, line 1423: probably the plugin id wrong

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1324,6 +1322,7 @@ export namespace SessionContext {
 export class SessionContextDialogs implements ISessionContext.IDialogs {
   constructor(options: ISessionContext.IDialogsOptions = {}) {
     this._translator = options.translator ?? nullTranslator;
+    this._settingRegistry = options.settingRegistry;
```
NexVeridian, line 1325: I'm not sure where to get this from, in this it gets pass though from somewhere, 
https://github.com/jupyterlab/jupyterlab/blob/82051a9d0c6dc93677b0b66560aea48d71aaa043/packages/docmanager-extension/src/index.tsx#L929-L930

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1428,20 +1443,40 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
...
+          'If checked, no confirmation to restart the kernel will be asked in the future.'
+        )
+      }
     });
 
     if (kernel.isDisposed) {
       return false;
     }
     if (result.button.accept) {
+      if (result.isChecked === skipKernelRestartDialogSetting) {
```
fcollonval, line 1459: You need to test if `isChecked` is a boolean because, if the user click on cancel or outside the dialog the checkbox value will be undetermined.

```suggestion
      if (typeof result.isChecked === 'boolean' && result.isChecked === skipKernelRestartDialogSetting) {
```

fcollonval (changes requested):
Very nice @NexVeridian you are on the right path.

The dialogs plugin is defined there:

https://github.com/jupyterlab/jupyterlab/blob/59acb332b53095c36452e10a553eac574592842b/packages/apputils-extension/src/index.ts#L574-L585

So you will have to add `ISettingRegistry` as additional `optional` dependency. Then you'll be able to pass it to `SessionContextDialogs`.

This also answer the question about the settings plugin. It should be `@jupyterlab/apputils-extension:sessionDialogs`. That implies you need to create in `packages/apputils-extension/schema` directory a new file named `sessionDialogs.json` (naming is important). That contains

```json
{
  "title": "Kernel dialogs",
  "description": "Kernel dialogs settings.",
  "properties": {
    "skipKernelRestartDialog": {
      "title": "...", # I let you add a short description here
      "description": "...", # I let you add a long description here
      "type": "boolean",
      "default": false
    },
  },
  "type": "object"
}
```

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1319,6 +1322,7 @@ export namespace SessionContext {
 export class SessionContextDialogs implements ISessionContext.IDialogs {
   constructor(options: ISessionContext.IDialogsOptions = {}) {
     this._translator = options.translator ?? nullTranslator;
+    this._settingRegistry = options.settingRegistry || null;
```
NexVeridian, line 1325: using the setting page works now, but checkbox dialog doesn't

`this._settingRegistry = options.settingRegistry || null;` is undefined, when the setting is unset

so I think this need to be nullSettingRegistry which I can't find

krassowski:
@fcollonval just for awareness, from a quick straw poll on https://github.com/jupyterlab/jupyterlab/issues/6420 it seems that option (b) may be preferred by a number of users. I think the consensus is that we should allow for both (b) and (c), but maybe the checkbox should be connected to (b)?

NexVeridian:
Everything should be working now, by doing this
> @fcollonval just for awareness, from a quick straw poll on #6420 it seems that option (b) may be preferred by a number of users. I think the consensus is that we should allow for both (b) and (c), but maybe the checkbox should be connected to (b)?


Or if you want to checkbox to be linked to c, then you could revert the last commit and fix this
> using the setting page works now, but checkbox dialog linked to the setting doesn't
> 
> `this._settingRegistry = options.settingRegistry || null;` is undefined, when the setting is unset
> 
> so I think this need to be nullSettingRegistry which I can't find

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1428,20 +1452,34 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
       buttons: [
         Dialog.cancelButton({ ariaLabel: trans.__('Cancel Kernel Restart') }),
         restartBtn
-      ]
+      ],
+      checkbox: {
+        label: trans.__('Do not ask me again.'),
+        caption: trans.__(
+          'If checked, no confirmation to restart the kernel will be asked in the future.'
```
krassowski, line 1459:
```suggestion
          'If checked, the kernel will restart without confirmation prompt in the future; you can change this back in the settings.'
```

packages/apputils/src/sessioncontext.tsx
```diff
@@ -326,6 +332,8 @@ export namespace ISessionContext {
      * Application translator object
      */
     translator?: ITranslator;
+
+    settingRegistry?: ISettingRegistry | null;
```
krassowski, line 336:
```suggestion
    /**
     * Optional setting registry used to access restart dialog preference.
     */
    settingRegistry?: ISettingRegistry | null;
```

packages/apputils-extension/schema/sessionDialogs.json
```diff
@@ -0,0 +1,14 @@
+{
+  "title": "Kernel dialogs",
+  "description": "Kernel dialogs settings.",
+  "additionalProperties": false,
+  "properties": {
+    "skipKernelRestartDialog": {
+      "title": "Skip kernel restart Dialog",
+      "description": "Whether the kernel restart confirmation dialog is shown when restarting the kernel.",
```
krassowski, line 8:
```suggestion
      "description": "Whether the kernel restart confirmation dialog is skipped when restarting the kernel.",
```

krassowski (commented):
The code here is looking good - just minor suggestions from me.

krassowski:
@fcollonval would you like to have another look, or is it good to merge?

krassowski:
@fcollonval just checking back on this

fcollonval (approved):
Thanks @NexVeridian this is good for me too.

Issue #8374, opened 2020-05-05
dismiss the connection error pop up permanently


VersBersh:
When the browser loses a connection to jupyter server a window pops up with the message "a connection to jupyter lab server could not be established...". If you dismiss the window and continue to edit a cell the window just pops up again a couple seconds later. This is rather annoying if you're working over a vpn connection that drops out regularly and want to continue working while it reconnects. It would be nice if after clicking dismiss the window doesn't show again until it has reconnected.

Jupyterlab: 2.1.2
Browsers: Firefox, Chrome
Replication: start jupyterlab, open a new notebook, and then kill the server.

possibly to do with this commit: 6ba1eb4dfd2044872a30be24ee18e2717e49f033

jasongrout:
Thanks. I agree that this is annoying - the dialog is too disruptive. I'd prefer a warning in the status bar or notebook toolbar, perhaps even glaring red or something to catch attention, over a modal dialog.

VersBersh:
https://github.com/VersBersh/jupyterlab/tree/connection-lost-window

on this branch I removed the dialogue, and added styling in the status bar (i.e. make the kernel status bold, red, and with loading dots when it disconnects), which is working okay for me atm.

blink1073:
Nice!  Can you please open a PR and include a screenshot?

vadimkantorov:
Here're couple screenshots:
![image](https://user-images.githubusercontent.com/1041752/101490339-7f48fc80-3962-11eb-9275-896c1b6ae66d.png)
![image](https://user-images.githubusercontent.com/1041752/101490379-8a9c2800-3962-11eb-8ac3-48b26bcd095a.png)

And they keep popping up and accumulate in the background, I had three modals to close when I switched to the tab.

Loss of connection is a very frequent mode, given that the server is often tunneled through ssh/vpn and they go down at Wi-Fi breakage or laptop resume from hibernation, and often need manual reconnect.

ceteri:
This is consistently one of the reasons why JL has become almost unusable in our business use cases. In the very least, JL must provide a setting to disable this pop up.

alexlenail:
Still unresolved?

morningmoni:
+1 on disabling this

krassowski:
An alternative to reworking this dialog into a statusbar element (which would not work for Jupyter Notebook app as it has no status bar) would be adding a little checkbox to the dialog, like:

```
Server Connection Error

(description)

[...]

PR #16269, opened 2024-05-01
Add option to dismiss "Server Connection Error" dialog for the duration of the session

RRosio:
## References

Related to https://github.com/jupyterlab/jupyterlab/issues/8374

## Code changes

This PR uses the `showDialog` function to display an error message with a checkbox giving the user the option to dismiss the popup dialog for the remainder of their tab session.

## User-facing changes

The user will now see a clickable checkbox that allows them to dismiss the error dialog informing them their connection to the server was lost.

Screenshot 2024-05-01 at 12 10 47 AM

Not applicable

### DIFF
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -24,7 +24,7 @@ export const ConnectionLost: IConnectionLost = async function (
   );
   const buttons = [Dialog.okButton({ label: trans.__('Dismiss') })];
 
-  if (!getDisplayConnection()) {
+  if (!Private.displayConnectionLost) {
     return;
   }
 
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -50,7 +50,7 @@ export const ConnectionLost: IConnectionLost = async function (
   })
     .then(result => {
       if (result.isChecked) {
-        disableConnectionLostDialog();
+        Private.displayConnectionLost = false;
       }
       return;
     })
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -42,7 +42,7 @@ export const ConnectionLost: IConnectionLost = async function (
         title: title,
         body: networkMsg,
         checkbox: {
-          label: trans.__('Do not show this message again.'),
+          label: trans.__('Do not show this message again in this session.'),
           caption: trans.__(
             'If checked, you will not see a dialog informing you about an issue with server connection in the future.'
           )
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -7,7 +7,7 @@ import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { IConnectionLost } from './tokens';
 
 let displayConnectionLost = true;
-let connectionLostCachePromise: Promise<Dialog.IResult<string>> | undefined;
+let serverConnectionLost: Promise<void | Dialog.IResult<string>> | undefined;
 
 /**
  * A default connection lost handler, which brings up an error dialog.
@@ -31,44 +31,34 @@ export const ConnectionLost: IConnectionLost = async function (
     return;
   }
 
-  if (connectionLostCachePromise) {
+  if (serverConnectionLost) {
     // Wait for the pre-existing promise to complete
-    await connectionLostCachePromise;
+    await serverConnectionLost;
     return;
   }
 
-  try {
-    const errorDialogPromise: Promise<Dialog.IResult<string>> = showDialog({
-      title: title,
-      body: networkMsg,
-      checkbox: {
-        label: trans.__('Do not show this message again in this session.'),
-        caption: trans.__(
-          'If checked, you will not see a dialog informing you about an issue with server connection in the future.'
-        )
-      },
-      buttons: buttons
+  serverConnectionLost = showDialog({
+    title: title,
+    body: networkMsg,
+    checkbox: {
+      label: trans.__('Do not show this message again in this session.'),
+      caption: trans.__(
+        'If checked, you will not see a dialog informing you about an issue with server connection in the future.'
+      )
+    },
+    buttons: buttons
+  })
+    .then(result => {
+      if (result.isChecked) {
+        displayConnectionLost = false;
+      }
+      return;
     })
-      .then(result => {
-        if (result.button.accept && !result.isChecked) {
-          connectionLostCachePromise = undefined;
-        }
-
-        if (result.isChecked) {
-          displayConnectionLost = false;
-        }
-        return result as Dialog.IResult<string>;
-      })
-      .finally(() => {
-        connectionLostCachePromise = undefined;
-      });
-
-    connectionLostCachePromise = errorDialogPromise;
-
-    // Wait for the dialog promise to resolve
-    await errorDialogPromise;
-    return;
-  } catch (error) {
-    console.error('An error occurred while showing the dialog: ', error);
-  }
+    .catch(error => {
+      console.error('An error occurred while showing the dialog: ', error);
+    })
+    .finally(() => {
+      serverConnectionLost = undefined;
+    });
+  return;
 };
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -6,26 +6,6 @@ import { ServerConnection, ServiceManager } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { IConnectionLost } from './tokens';
 
-export function getDisplayConnection(): boolean {
-  return Private.displayConnectionLost;
-}
-
-export function disableConnectionLostDialog(): void {
-  Private.displayConnectionLost = false;
-}
-
-export function getServerConnectionLost():
-  | Promise<void | Dialog.IResult<string>>
-  | undefined {
-  return Private.serverConnectionLost;
-}
-
-export function setServerConnectionLost(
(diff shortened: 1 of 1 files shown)


packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,45 @@ export const ConnectionLost: IConnectionLost = async function (
...
+        }
+        return;
+      });
 
-  return showErrorMessage(title, { message: networkMsg });
+      connectionLostCachePromise[key] = errorDialogPromise;
+      await errorDialogPromise;
+      return;
+    } catch (error) {
+      console.error('An error occurred while showing the dialog:', error);
```
JasonWeill, line 66:
```suggestion
      console.error('An error occurred while showing the dialog: ', error);
```

Add space before error

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,45 @@ export const ConnectionLost: IConnectionLost = async function (
       'JupyterLab will continue trying to reconnect.\n' +
       'Check your network connection or Jupyter server configuration.\n'
   );
+  const buttons = [Dialog.okButton({ label: trans.__('Dismiss') })];
+
+  if (!displayConnectionLost) {
+    return;
+  }
+
+  const key = title + '----' + networkMsg;
```
JasonWeill, line 34:
```suggestion
  const key = 'server-connection-error';
```

Localized strings aren't typically used for internal record-keeping like this, due to the admittedly small risk that two strings in English could translate to the same string in some other language. I think it's fine to use a non-translated string in this case.

JasonWeill (commented):
Thanks for opening this! I added a couple of comments. Also, when I checked out and ran your code, I found that if I stop the server, then see the dialog, then click somewhere on the page not in the dialog's body, the dialog goes away, and I don't see it again, even if I don't have the box checked. Is this something you could fix in this PR?

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,45 @@ export const ConnectionLost: IConnectionLost = async function (
...
+
+  if (promise) {
+    return promise;
+  } else {
+    try {
+      const errorDialogPromise: any = showDialog({
+        title: title,
+        body: networkMsg,
+        checkbox: {
+          label: trans.__('Do not show this message again.'),
```
krassowski, line 45: I would be afraid to check this box thinking that I will never see the error again and that I won't find a way to re-enable it.

```suggestion
          label: trans.__('Do not show this message again in this session.'),
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,45 @@ export const ConnectionLost: IConnectionLost = async function (
...
       'Check your network connection or Jupyter server configuration.\n'
   );
+  const buttons = [Dialog.okButton({ label: trans.__('Dismiss') })];
+
+  if (!displayConnectionLost) {
+    return;
+  }
+
+  const key = 'server-connection-error';
+  const promise = connectionLostCachePromise[key];
```
krassowski, line 35: After the recent changes, I do not see a reason for this to be a mapping if the key is constant.

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,45 @@ export const ConnectionLost: IConnectionLost = async function (
...
+
+  if (promise) {
+    return promise;
+  } else {
+    try {
+      const errorDialogPromise: any = showDialog({
```
krassowski, line 41: New code should not use `any` except for tests. `any` disables type checking and in past lead to numerous user-facing bugs. This will be enforced with https://github.com/jupyterlab/jupyterlab/pull/16219

packages/application/src/connectionlost.ts
```diff
@@ -1,11 +1,14 @@
...
 // Distributed under the terms of the Modified BSD License.
 
-import { showErrorMessage } from '@jupyterlab/apputils';
+import { Dialog, showDialog } from '@jupyterlab/apputils';
 import { ServerConnection, ServiceManager } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { IConnectionLost } from './tokens';
 
+let displayConnectionLost = true;
+const connectionLostCachePromise: { [key: string]: Promise<any> } = {};
```
krassowski, line 10: New code should not use `any` except for tests. Any disables type checking and in past lead to numerous user-facing bugs. This will be enforced with https://github.com/jupyterlab/jupyterlab/pull/16219

krassowski (changes requested):

krassowski:
The `check_release` failure is unrelated, tracking it here: https://github.com/jupyterlab/jupyterlab/issues/16448

krassowski:
@RRosio is it ready for review?

packages/application/src/connectionlost.ts
```diff
@@ -1,11 +1,14 @@
...
 // Distributed under the terms of the Modified BSD License.
 
-import { showErrorMessage } from '@jupyterlab/apputils';
+import { Dialog, showDialog } from '@jupyterlab/apputils';
 import { ServerConnection, ServiceManager } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { IConnectionLost } from './tokens';
 
+let displayConnectionLost = true;
+let serverConnectionLost: Promise<void | Dialog.IResult<string>> | undefined;
```
krassowski, line 10: This is just stylistic, but I would suggest to wrap this in a `Private` namespace and move to the end of the file. See how this is done on in the clipboard package for example:

https://github.com/jupyterlab/jupyterlab/blob/ae123cb82015540804964f76ba4e56f21799ea27/packages/apputils/src/clipboard.ts#L99-L107

This will makes it clear these variables are private, and also allow you to write a structured docstring for each of them :)

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +25,40 @@ export const ConnectionLost: IConnectionLost = async function (
...
+  }
 
-  return showErrorMessage(title, { message: networkMsg });
+  serverConnectionLost = showDialog({
+    title: title,
+    body: networkMsg,
+    checkbox: {
+      label: trans.__('Do not show this message again in this session.'),
+      caption: trans.__(
+        'If checked, you will not see a dialog informing you about an issue with server connection in the future.'
```
krassowski, line 46:
```suggestion
        'If checked, you will not see a dialog informing you about an issue with server connection in this session.'
```

krassowski (commented):
Thank you, this is looking good to me! Just minor stylistic suggestions.

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
...
   );
+  const buttons = [Dialog.okButton({ label: trans.__('Dismiss') })];
 
-  return showErrorMessage(title, { message: networkMsg });
+  if (!Private.displayConnectionLost) {
+    return;
+  }
+
+  let connectionDialog: Promise<void | Dialog.IResult<string>> | undefined =
+    Private.serverConnectionLost;
```
krassowski, line 32:
```suggestion
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
...
+
+  if (connectionDialog) {
```
krassowski, line 34:
```suggestion
  if (Private.serverConnectionLost) {
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
...
+    // Wait for the pre-existing promise to complete
+    await connectionDialog;
```
krassowski, line 36:
```suggestion
    await Private.serverConnectionLost;
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
...
+    .catch(error => {
+      console.error('An error occurred while showing the dialog: ', error);
+    })
+    .finally(() => {
+      Private.serverConnectionLost = undefined;
+    });
+
+  Private.serverConnectionLost = dialog;
+
+  return;
```
krassowski, line 66:
```suggestion
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
       'JupyterLab will continue trying to reconnect.\n' +
       'Check your network connection or Jupyter server configuration.\n'
   );
+  const buttons = [Dialog.okButton({ label: trans.__('Dismiss') })];
```
krassowski, line 25: I will move it down as this is where it will be used

```suggestion
```

packages/application/src/connectionlost.ts
```diff
@@ -22,6 +22,62 @@ export const ConnectionLost: IConnectionLost = async function (
...
+  const dialog = showDialog({
+    title: title,
+    body: networkMsg,
+    checkbox: {
+      label: trans.__('Do not show this message again in this session.'),
+      caption: trans.__(
+        'If checked, you will not see a dialog informing you about an issue with server connection in this session.'
+      )
+    },
+    buttons: buttons
```
krassowski, line 49:
```suggestion
    buttons: [Dialog.okButton({ label: trans.__('Dismiss') })]
```

krassowski (approved):
Thank you so much @RRosio!

> I will make a few stylistic changes (some of which are subjective) on things that only became apparent when looking at the latest revision.

PR #16273, opened 2024-05-01
Check the command is registered before calling `notifyCommandChanged()`

jtpio:
## References

https://github.com/jupyterlab/jupyterlab/issues/16113.

## Code changes

Implement the suggestion from @krassowski in https://github.com/jupyterlab/jupyterlab/pull/16138#discussion_r1555892452.

Still not sure it's the "right" fix, but maybe it can already help limit the amount of errors seen in the dev tools console when using Notebook 7.

## User-facing changes

This should help fix the following errors showing up in the dev tools console when using Notebook 7:

image

image

None


krassowski (approved):
Thank you!

PR #16288, opened 2024-05-06
coreutils: add required TS intl libraries

### LINKED ISSUE #16283: coreutils and post-ES2018 Intl API features
## Description

Since the removal of the `moment` library in #13469, the Time functions in coreutils now depend on the ECMA-402 internationalization APIs, namely DateTimeFormat and RelativeTimeFormat. As spotted recently when changes were made to time.ts in #15948, this starts to run into ES2018 compatibility issues.

In fact, RelativeTimeFormat doesn't exist at all under ES2018, nor do the `dateStyle` and `timeStyle` options passed to DateTimeFormat. The reason those features currently don't cause build errors is due to `"types": ["node"]` in the coreutils tsconfig, which as a side-effect pulls in the es2020.intl library definitions via the triple-slash directives at the top of index.d.ts in @types/node, suppressing the errors that would appear otherwise.

In addition, DateTimeFormat's `dateStyle` and `timeStyle` options were actually only added in the ES2021 standard, and there is a current proposal for these to be moved to the es2021.intl library accordingly. If and when this happens, those properties will no longer be defined in es2020.intl, and the coreutils build will subsequently break.

## Expected behavior

If the decision is that JupyterLab should continue to use these post-ES2018 features, then the coreutils tsconfig should probably explicitly add the es2020.intl library to make this intention clear (and potentially es2021.intl to pre-empt the above), and/or implement a polyfill for users running on true ES2018 environments.

Renegade334:
## References

Resolves #16283.

## Code changes

Explicitly adds the es2020.intl library, required for RelativeTimeFormat and its ancillary types, to tsconfigbase.

Also pre-emptively adds es2021.intl to the coreutils tsconfig, which will be required to avoid build errors in time.ts if/when the DateTimeFormat option properties `dateStyle` and `timeStyle` are moved there from es2020.intl.

## User-facing changes

Nil. (It's no longer necessary to explicitly define a `HumanStyle` type, but as it's an exported namespace type, it's left as-is.)

Nil.

### DIFF
--- a/packages/coreutils/src/time.ts
+++ b/packages/coreutils/src/time.ts
@@ -17,7 +17,7 @@ const UNITS: { name: Intl.RelativeTimeFormatUnit; milliseconds: number }[] = [
  * The namespace for date functions.
  */
 export namespace Time {
-  export type HumanStyle = Intl.ResolvedRelativeTimeFormatOptions["style"];
+  export type HumanStyle = Intl.ResolvedRelativeTimeFormatOptions['style'];
 
   /**
    * Convert a timestring to a human readable string (e.g. 'two minutes ago').
--- a/packages/coreutils/src/time.ts
+++ b/packages/coreutils/src/time.ts
@@ -17,9 +17,7 @@ const UNITS: { name: Intl.RelativeTimeFormatUnit; milliseconds: number }[] = [
  * The namespace for date functions.
  */
 export namespace Time {
-  // Intl.RelativeTimeFormatStyle contains these, but it requires `ES2020.Intl`.
-  // We currently compile to an `ES2018` target.
-  export type HumanStyle = 'long' | 'short' | 'narrow';
+  export type HumanStyle = Intl.ResolvedRelativeTimeFormatOptions["style"];
 
   /**
    * Convert a timestring to a human readable string (e.g. 'two minutes ago').
--- a/packages/coreutils/tsconfig.json
+++ b/packages/coreutils/tsconfig.json
@@ -4,7 +4,8 @@
     "outDir": "lib",
     "rootDir": "src",
     "module": "commonjs",
-    "types": ["node"]
+    "types": ["node"],
+    "lib": ["DOM", "ES2018", "ES2020.Intl", "ES2021.Intl"]
   },
   "include": ["src/*"],
   "references": []
--- a/tsconfigbase.json
+++ b/tsconfigbase.json
@@ -7,6 +7,7 @@
     "esModuleInterop": true,
     "incremental": true,
     "jsx": "react",
+    "lib": ["DOM", "DOM.Iterable", "ES2018", "ES2020.Intl"],
     "module": "esnext",
     "moduleResolution": "node",
     "noEmitOnError": true,


packages/coreutils/src/time.ts
```diff
@@ -17,9 +17,7 @@ const UNITS: { name: Intl.RelativeTimeFormatUnit; milliseconds: number }[] = [
  * The namespace for date functions.
  */
 export namespace Time {
-  // Intl.RelativeTimeFormatStyle contains these, but it requires `ES2020.Intl`.
-  // We currently compile to an `ES2018` target.
-  export type HumanStyle = 'long' | 'short' | 'narrow';
+  export type HumanStyle = Intl.ResolvedRelativeTimeFormatOptions["style"];
```
Renegade334, line 20: This could target `Intl.RelativeTimeFormatStyle` as originally proposed. The only downside to this is that there is a proposal to rename the RelativeTimeFormat option types to bring them in line with the Intl namespace convention (which would be `Intl.RelativeTimeFormatOptionsStyle`), so this type name may be removed or deprecated in a future TypeScript release.

Renegade334, line 20: The `HumanStyle` type is no longer strictly necessary, but as it's exported, removing it would technically be a breaking change.

krassowski: It looks like it does not compile:

```
  ../coreutils/lib/time.d.ts(5,28): error TS2724: 'Intl' has no exported member named 'ResolvedRelativeTimeFormatOptions'. Did you mean 'ResolvedDateTimeFormatOptions'?
```

Renegade334: What was missing from the PR (mea culpa) was the default library set in tsconfigbase. This seems like an acceptable approach, given that Jupyter explicitly relies on features defined in those libraries.

packages/coreutils/src/time.ts
```diff
@@ -17,9 +17,7 @@ const UNITS: { name: Intl.RelativeTimeFormatUnit; milliseconds: number }[] = [
  * The namespace for date functions.
  */
 export namespace Time {
-  // Intl.RelativeTimeFormatStyle contains these, but it requires `ES2020.Intl`.
-  // We currently compile to an `ES2018` target.
-  export type HumanStyle = 'long' | 'short' | 'narrow';
+  export type HumanStyle = Intl.ResolvedRelativeTimeFormatOptions['style'];
```
JasonWeill, line 20: As the original author of the code you're replacing, I'm OK with deleting `HumanStyle` entirely, replacing references to `Time.HumanStyle` with `Intl.RelativeTimeFormatStyle`.

JasonWeill (commented):
Thanks for opening this PR!

packages/coreutils/src/time.ts
Renegade334: Although recent, the commit that introduced the `format` parameter did make it into coreutils 6.2.0, and removing this exported type would _technically_ be a breaking change to the coreutils library. It becomes a redundant alias, for sure, but equally it doesn't do any harm to keep it.

As an aside, where the type is used elsewhere in jupyterlab, `Time.HumanStyle` probably gives clearer context than `Intl.Blah`, although YMMV.

JasonWeill: I'm OK to keep `Time.HumanStyle` as well; I like the idea of avoiding a breaking change if we can help it. Thanks again for working on this!

fcollonval (approved):

packages/coreutils/src/time.ts
jhgoebbert: Is this issue related to this change?
https://github.com/jupyter-server/jupyter-resource-usage/issues/237

PR #16291, opened 2024-05-06
Wait for the cell toolbar items to be rendered the first time before looking for overlap

brichet:
When the cell toolbar is attached the first time, the items may not be rendered when we try to see if it overlap the cell content.
This make the toolbar to be displayed when it should not (see https://github.com/jupyterlab/jupyterlab/issues/15305#issuecomment-2036613861)

Peek 2024-05-06 16-36

## References

Related to https://github.com/jupyterlab/jupyterlab/issues/15305, and more specifically should fix https://github.com/jupyterlab/jupyterlab/issues/15305#issuecomment-2036613861

## Code changes

- hide the toolbar by default when it is added to the cell, to avoid flickering
- wait for all the items to be rendered, using `requestIdleCallback()` instead of `requestAnimationFrame()` before looking for overlap (this should wait for changes in the cell and in the toolbar to be effective).

## User-facing changes

Avoid the toolbar to be displayed when it should not.

None


JasonWeill (approved):
I see a slight (~1 frame) period after focusing a cell before the toolbar appears, but this should be OK in day-to-day use. Good work!

krassowski:
> I see a slight (~1 frame) period after focusing a cell before the toolbar appears

The delay is slightly distracting annoying when clicking on the editor. That said, when just adding cells, I think the delay is less annoying than the previous behaviour of rendering one button at a time (captured with x6 CPU throttling):

| Before | After |
|--|---|
| before | after |

krassowski:
Maybe there should be an opacity-based transition so that they appear smoothly?

PR #16293, opened 2024-05-06
Search and replace with substring in markdown and raw cells

### LINKED ISSUE #16292: In markdown cells, replacing a string with text that contains the string fails
## Description

This is a follow up to #16242, itself a follow up to #15572.

In a markdown cell, replacing a string with a replacement string that contains the search string at least once (for example, `tom` replaced with `tommy` or `tomtom`) does not cause the highlight to go to the next cell.

Kapture 2024-05-06 at 09 58 18

## Reproduce

1. Create two markdown cells, each containing the word `tom` twice.
2. Search for `tom` and replace it with a string that contains `tom` at least once.
3. Replace the first two instances in the first cell.
4. The highlight **does not** move to the second cell.

## Expected behavior

The first match in the second cell is highlighted.

JasonWeill:
Fixes #16292.

Adapts the codemirror searchprovider changes from #15881 to also apply to the MarkdownCellSearchProvider. Updates unit tests to test all pairs of code, markdown, and raw cells.

### DIFF
--- a/galata/test/jupyterlab/notebook-replace.test.ts
+++ b/galata/test/jupyterlab/notebook-replace.test.ts
@@ -111,151 +111,157 @@ test.describe('Notebook Search and Replace', () => {
     expect(await nbPanel!.screenshot()).toMatchSnapshot('replace-all.png');
   });
 
-  test('Replace step-by-step across cell boundaries', async ({ page }) => {
-    // Create a small test notebook
-    await page.notebook.createNew();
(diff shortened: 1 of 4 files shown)


srdas:
@JasonWeill - I installed, built, and tested the new version from your branch `search-replace-substring-markdown` and the find and replace is now working across all types of cells (code, markdown, raw).

PR #16298, opened 2024-05-07
Allow to invoke inline completer anywhere (not only at the end of line)

### LINKED ISSUE #16295: InlineCompleter only runs when  cursor is at the end of a line
## Description

The inline completer will only be invoked when run from the end of a non-empty line. This is caused by the following chunk of code:

https://github.com/jupyterlab/jupyterlab/blob/f14c89145d697974ecbd02872643b530381d90b6/packages/completer/src/handler.ts#L462-L465

Based on the inline comment, that chunk of code should only run when `trigger` is `Automatic`, not when it's `Invoke`

## Reproduce

1. Enable an inline completer
2. Invoke the inline completer on a notebook cell or file when the cursor is not at the end of a line
3. Note that nothing happens
4. Compare to running at end of line

## Expected behavior

The inline completer should run when the command is executed, no matter where the cursor is

## Solution

Add a check for `trigger === Automatic` in the if statement

ajbozarth:
## References

Fixes #16295

## Code changes

The inline completer prevents automatic running of the completer when not at the end of a line of code, but it is also applying this when invoked manually, which was unintended. This addresses that bug.

## User-facing changes

Allows a user to manually invoke the inline completer anywhere in their code

### DIFF
--- a/packages/completer/src/handler.ts
+++ b/packages/completer/src/handler.ts
@@ -459,7 +459,10 @@ export class CompletionHandler implements IDisposable {
     }
 
     const line = editor.getLine(position.line);
-    if (typeof line === 'undefined' || position.column < line.length) {
+    if (
+      trigger === InlineCompletionTriggerKind.Automatic &&
+      (typeof line === 'undefined' || position.column < line.length)
+    ) {
       // only auto-trigger on end of line
       return;
     }


krassowski (approved):
Thank you @ajbozarth!

ajbozarth:

ajbozarth:
@krassowski I noticed 4.2.1 went out last week and this wasn't in it. I believe this was missed when creating the 4.2.x branch prior to moving main to 3.0.x, how should we address this?

fcollonval:
Hey @ajbozarth I see the commit on the branch 4.2.x: https://github.com/jupyterlab/jupyterlab/commits/4.2.x/

But indeed it is not listing in the changelog (releaser bug?). Could you check it is fixed in 4.2.1?

ajbozarth:
Ah sorry, I only checked the changelog. It is there, thanks :)

fcollonval:
Thanks for quickly checking, we need to update the changelog I guess.

Issue #16299, opened 2024-05-07
Icon classes (like `jp-icon0`) do not work in launcher tabs for kernels


krassowski:
## Description

> @krassowski Your SVG does not work for me.
> 
> The `jp-LauncherCard-icon` is embedded as `img` and seems to prevent `jp-icon0` from working.

_Originally posted by @benz0li in https://github.com/betatim/vscode-binder/issues/61#issuecomment-2097708178_
            

## Reproduce

https://github.com/jupyterlab/jupyterlab/blob/f14c89145d697974ecbd02872643b530381d90b6/packages/launcher/src/widget.tsx#L299-L306

## Expected behavior

We should allow for theme-responsive icons in launcher

We could:
- embed SVG directly (needs sanitization!)
- explore https://github.com/w3c/csswg-drafts/issues/8634 or CSS linked parameters approach

## Context

- JupyterLab version: 4.2.0

krassowski:
Another solution would be having two different sets of icon URLs, one for dark theme and one for light theme.

krassowski:
After researching a few solutions, it looks like the best short-term solution is indeed setting the preferred-color like this:

```
<div style="color-scheme: dark">
  <img />
</div>
```

based on the current JupyterLab theme.

PR #16301, opened 2024-05-07
Advertise the color scheme based on the theme

krassowski:
## References

Closes https://github.com/jupyterlab/jupyterlab/issues/16299

## Code changes

Set [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) style on `<body>` to `dark` if the current theme is dark, and to `light` if the current theme is light.

## User-facing changes

- Scrollbars and other UI elements (not styled by JupyterLab) in dark mode will use the system dark theme if available
- SVG icons for kernels which include [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) directive will correctly render with high contrast in dark theme

With "Theme Scrollbars" option **disabled** (default):

| Before | After |
|--------|--------|
|Screenshot from 2024-05-07 14-45-26 | Screenshot from 2024-05-07 14-45-00 |

None

### DIFF
--- a/packages/apputils-extension/src/themesplugins.ts
+++ b/packages/apputils-extension/src/themesplugins.ts
@@ -86,6 +86,9 @@ export const themesPlugin: JupyterFrontEndPlugin<IThemeManager> = {
         manager.isLight(currentTheme)
       );
       document.body.dataset.jpThemeName = currentTheme;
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

fcollonval (approved):

PR #16302, opened 2024-05-07
Allow RegExp in galata helper when looking for a tab

brichet:
Using the galata helper function `activity -> getTabLocator` [doesn't look for the exact string name](https://github.com/jupyterlab/jupyterlab/blob/f14c89145d697974ecbd02872643b530381d90b6/galata/src/helpers/activity.ts#L109), and can fail due to duplicate string.

For example it is not possible to use it with the tab `Settings`, it lead to 2 locators, (1) the expected panel's tab and (2) the settings named `Settings Editor Form UI` (because the plugin list entry also have a ['tab' role](https://github.com/jupyterlab/jupyterlab/blob/f14c89145d697974ecbd02872643b530381d90b6/packages/settingeditor/src/pluginlist.tsx#L410))

Allowing also `RegExp` can help specifying the expected match string.

## References

None

## Code changes

Allow `string` or `RegExp` (instead of only `string`) in the activity function that look for a tab name.

## User-facing changes

None

None


krassowski (approved):
Thank you @brichet!

krassowski:
We will need to branch out 4.2.x from the previous commit (and update configs). I can do that later today or tomorrow.

brichet:
Sorry for this, I'll be more careful next time.

brichet:
> We will need to branch out 4.2.x from the previous commit (and update configs). I can do that later today or tomorrow.

Out of curiosity, what are the configs to update ? I think of adding a `4.2.x` milestone. Is there something else ?
BTW thanks for fixing it already @krassowski.

Issue #16297, opened 2024-05-06
Resetting a shortcut does not restore and enable the default if it was modified


RRosio:
## Description
I have tried updating the shortcuts, whether I add a new shortcut in addition to an existing one or I modify the single provided shortcut, there are times where the shortcut disappears entirely and if I visit the user-settings `shortcuts.jupyterlab-settings`  config file I can see it is disabled.

https://github.com/jupyterlab/jupyterlab/assets/12378147/5cd1fd28-c1da-4e1e-b422-1f0cebf8ff1a

## Reproduce

1. Go to 'Settings'
2. Select 'Settings Editor'
3. Select the 'Keyboard Shortcuts' from the left menu.
4. Modify a shortcut and accept changes 
5. Reset the shortcut (steps 4-5 may need to be repeated multiple times)
6. At some point, the shortcut will disappear from the list displayed.
7. See the  'user-settings' `shortcuts.jupyterlab-settings` and see the shortcut now has the `"disabled": true` property

## Expected behavior

I would expect that the shortcut, when reset, goes back to it's initial values.  

## Context

- Operating System and version: M1 macOS 3.6.1 (22G313)
- Browser and version: 123.0.6312.87 (Official Build) (arm64)
- JupyterLab version: 4.2.0

krassowski:
The "disabled" entry is added when you modify the default shortcut (it is a two step process - "disable" the default shortcut, add a new shortcut). The issue is that when you click "Reset", it removes the new shortcut but does not enable the default shortcutback. I will rename the issue to reflect this.

PR #16304, opened 2024-05-07
Resetting a shortcut does not restore and enable the default if it was modified

itsmevichu:
Fix for https://github.com/jupyterlab/jupyterlab/issues/16297

### DIFF
--- a/packages/shortcuts-extension/src/components/ShortcutUI.tsx
+++ b/packages/shortcuts-extension/src/components/ShortcutUI.tsx
@@ -384,12 +384,12 @@ export class ShortcutUI
           });
         }
         found = true;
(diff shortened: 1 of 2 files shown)


krassowski:
Thanks for starting this. It looks like the tests are failing, and the failures are indeed relevant. May I suggest starting this PR with a test case for the issue, it could possibly go into:

https://github.com/jupyterlab/jupyterlab/blob/393d80f2553ad8d1debd62c14817378d5f22ff9c/packages/shortcuts-extension/test/components/ShortcutUI.spec.ts#L272

itsmevichu:
@krassowski I have added the test cases. And the visual regression test failures are irrespective of my changes I guess.

krassowski (approved):
Thank you @itsmevichu!

PR #16305, opened 2024-05-08
Fix typo in workspace format documentation (4.1.x)

jerrison:
Fixing typo.

### DIFF
--- a/docs/source/user/urls.rst
+++ b/docs/source/user/urls.rst
@@ -207,7 +207,7 @@ default workspace.
 Workspace File Format
 ---------------------
 
-A workspace file in a JSON file with a specific spec.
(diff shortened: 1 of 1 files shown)


docs/source/user/urls.rst
```diff
@@ -207,7 +207,7 @@ default workspace.
 Workspace File Format
 ---------------------
 
-A workspace file in a JSON file with a specific spec.
+A workspace file is a JSON file with a specific spec.
```
krassowski, line 210: Of note this is fixed on 4.2.x and rewritten to: "A workspace file is a JSON file that contains one object with two required top-level keys, data, and metadata."

krassowski (approved):
Thanks!

PR #16306, opened 2024-05-08
Upgrade CodeMirror dependencies

krassowski:
## References

Picks up https://github.com/codemirror/lang-python/commit/0468622ddfee9c2a0011544c8f3128749b714168 which closes https://github.com/jupyterlab/jupyterlab/issues/16196

Upgrades other CodeMirror dependencies as these typically need to be bumped in unison to avoid out-of-sync issues.

## Code changes

- `@codemirror/autocomplete`: 6.15.0 → 6.16.0
- `@codemirror/commands`: 6.3.3 → 6.5.0
  > - Fix an issue where deleteLine sometimes leaves the cursor on the wrong line.
- `@codemirror/lang-markdown`: 6.2.4 → 6.2.5
  > - Disable folding for list nodes (since it will shadow the folding on the first list item).
- `@codemirror/legacy-modes`: 6.3.3 → 6.4.0
  > - Only match Solr operator words when they are upper-case.
  > - Fix an infinite loop when tokenizing heredoc strings in the Crystal mode.
  > - Add the old Pug mode.
- `@codemirror/lang-html`: 6.4.8 → 6.4.9
  > - Fix a bug in `autoCloseTags` that made tags not close when typing > after an attribute.
- `@codemirror/lang-python`: 6.1.4 → 6.1.6
  > - Improve the way indentation for the current body is preserved when inenting new lines.
  > - Properly indent `else:` when attached to a `for` or `while` statement.
- `@codemirror/lang-sql`: 6.6.1 → 6.6.4
  > - Make statement folding leave the entire first line visible.
  > - Fix a bug where Postgres-style dollar-quoted strings were enabled for all dialects, and the `doubleDollarQuotedStrings` options was ignored.
  > - Properly support tags in PostgreSQL 4073 quoted strings.
- `@codemirror/view`: 6.26.0 → 6.26.3
  > - Fix an issue where dispatching an update to an editor before it measured itself for the first time could cause the scroll position to incorrectly move.
  > - Fix a crash when multiple tooltips with arrows are shown.
  > - Improve behavior of scrollPastEnd in a scaled editor.
  > - When available, use Selection.getComposedRanges on Safari to find the selection inside a shadow DOM.
  > - Remove the workaround that avoided inappropriate styling on composed text after a decoration again, since it breaks the stock Android virtual keyboard.
  > - Fix the editor getting stuck in composition when Safari fails to fire a compositionend event for a dead key composition.
  > - Fix an issue where, with IME systems that kept the cursor at the start of the composed text, the editor misidentified the target node and disrupted composition.
  > - Fix a bug where in a line-wrapped editor, with some content, the initial scroll position would be off from the top of the document.
- `@lezer/markdown`: 1.2.0 → 1.3.0:
  > - GFM autolinks will no longer include the closing bracket of a surrounding link or image.

## User-facing changes

- Python auto-indentation after `else`: in `for` and `while` will be improved
- Python indentation will be retained in indented blocks
- Pug templates (previously known as Jade) will be syntax highlighted, as in JupyterLab 3.x

image

None


fcollonval (approved):
Thanks @krassowski

PR #10676, opened 2021-07-22
added fullscreen command

DerThorsten:
Added a command to use jupyterlab in fullscreen mode
## References
#8710

## Code changes
Added Fullscreen Mode command to `applications-extension`.
Added getter / setter of fullScreenMode to  `LabShell`

## User-facing changes
Users can toggle fullscreen  via :
* (Ctrl + Shift + F11)
* Fullscreen Mode Command
* Fullscreen Mode Command via View Menue

## Backwards-incompatible changes

None

PR #16308, opened 2024-05-08
Add fullscreen mode menu entry and shortcut

### LINKED ISSUE #8710: Add full screen option to JupyterLab
### Problem

I often want to make my JupyterLab interface full screen, without any browser tabs or other window chrome, so I can concentrate on the problem at hand.

Maximizing a chrome browser in macOS almost does what I want, but only if I uncheck "Always show toolbar in Full Screen". I can't figure out a way to do this in Firefox, and I can't figure out a way to do this in Windows either.

### Proposed Solution

Add a command to make the JupyterLab root application DOM element full screen, using the [requestFullScreen](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen) api.

Add this as a menu item, perhaps under the View menu next to the Single-Document Mode?

dhysdrn:
## References:
Fixes #8710 - Add fullscreen option to JupyterLab
Adding the base fullscreen mode functionality in preparation for fullscreen mode expansion to allow individual widgets to enter fullscreen mode for presentation purposes per the requests in PR #10676.

## Code Changes:

- Add new `application:toggle-fullscreen-mode` command
- Set default shortcut to <kbd>F11</kbd>; other shortcuts work in Chrome but not in Firefox
- Add menu entry in View → Appearance to allow toggling the fullscreen mode
- Set `data-jp-debugger-stopped-threads` attribute on the `body` when debugger has stopped threads and adjust the shortcut selectors in debugger accordingly to solve conflict with the F11 shortcut

## User-Facing Changes:
Provides a menu option to enter fullscreen mode and uses <kbd>F11</kbd> as the shortcut key to toggle into and out of fullscreen mode.
no-fullscreen-menu
fullscreen-menu

If entering fullscreen using the F11 shortcut or button in the view > Fullscreen Mode menu, users can exit with the view > Fullscreen Mode menu option, ESC key, or F11 key, and JupyterLab maintains fullscreen synchronization of the menu item, check box, and CSS styling.
fullscreen-menu-checked

## Backwards-incompatible Changes:
None


Rob-P-Smith:
@meeseeksdev tag enhancement

packages/application-extension/src/index.tsx
```diff
@@ -445,6 +436,23 @@
...
 
+      commands.addCommand(CommandIDs.toggleFullscreenMode, {
+        label: () => trans.__('Fullscreen Mode'),
+        execute: () => {
+          labShell.fullscreenMode = !labShell.fullscreenMode;
+
+          if(labShell.fullscreenMode){
+            document.documentElement.requestFullscreen();
+            document.onfullscreenchange = ()=> {
+              !document.fullscreenElement ? labShell.fullscreenMode = false : null;
```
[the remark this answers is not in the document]
Rob-P-Smith: Same as comment above. This does the work, not sure why the test shows no effect.

Rob-P-Smith:
@krassowski We've adjusted the code to align more with the existing codebase and cleaned up the uncaught error issue that previously failed "jplm run eslint:typed".

If this is acceptable, we'd like to close this and work on full-screening individual elements as a separate issue next, e.g. terminal/notebook/widgets etc as requested in the previous PR that attempted to solve issue #8710 here:
https://github.com/jupyterlab/jupyterlab/pull/10676

bot please update snapshots

krassowski:
Just as FYI, you do not need to write down the code changes by file and line range in the PR description (these these quickly get outdated). What I would recommend is to focus on listing major API changes if there are any, e.g. that `fullScreenMode` was added to `LabShell`.

packages/application-extension/src/index.tsx
```diff
@@ -371,20 +374,7 @@ const mainCommands: JupyterFrontEndPlugin<void> = {
...
-            : args.side === 'right'
-            ? trans.__(
-                'Toggle Element %1 in Right Sidebar',
-                parseInt(args.index as string, 10) + 1
-              )
-            : trans.__(
-                'Toggle Element %1 in Left Sidebar',
-                parseInt(args.index as string, 10) + 1
-              ),
+        label: trans.__('Toggle Sidebar Element'),
```
krassowski, line 377: The previous code here needs to be restored. It is critical to rendering shortcuts properly in the settings, otherwise users do not know which toggle they are modifying.

galata/test/documentation/debugger.test.ts-snapshots/debugger-stop-on-raised-exception-documentation-linux.png
krassowski, line 1: This is unrelated, should be reverted.

galata/test/jupyterlab/notebook-run.test.ts-snapshots/notebook-panel-1-jupyterlab-linux.png
krassowski, line 1: This is unrelated, should be reverted.

packages/application/src/shell.ts
```diff
@@ -1303,6 +1314,26 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     return layout;
   }
 
+  /**
+   * Toggle the fullscreen mode based on user inputs
+   */
+  setFullscreenMode(): void {
```
krassowski, line 1320: Why do we need this method? Could it be handled in the setter of `fullscreenMode` property instead to minimize the API surface?

packages/application/src/shell.ts
```diff
@@ -609,6 +609,17 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     this.toggleClass('jp-mod-presentationMode', value);
   }
 
+  /**
+   * Whether JupyterLab is in fullscreen mode with the
+   * `jp-mod-fullscreenMode` CSS class.
+   */
+  get fullscreenMode(): boolean {
+    return this.hasClass('jp-mod-fullscreenMode');
```
krassowski, line 617: I don't think we should store state which can be modified by the JS API by using a CSS class - this can lead to inconsistent state if an extension or JS in outputs of user executed code manipulate the fullscreen state (for example rise extension for presenting slides could do that).

I would prefer if the getter returned the state derived from the browser API like `fullscreenElement`. That said, I am not even sure if we need a getter/setter here - if we are not setting the CSS class, then we could get away without extending the public API of `LabShell` at all, which will make merging this PR much easier.

krassowski (changes requested):
I do not have a strong opinion on this, but I think it is fine to start with.
Before this can be merged here is a few

packages/application/src/shell.ts
dhysdrn: We made the changes and removed the getter/setter and directly queried the fullscreenElement value, we also reverted toggle sidebar widget to its original code. Thank you! (:

dhysdrn:
@krassowski it looks like we're failing some tests and unsure what we need to do to fix it, could you help us with this?

krassowski:
Some of the failing tests are just flaky, I can restart these. One screenshot test appears to need updating:

```
[jupyterlab] › test/jupyterlab/menus.test.ts:30:9 › General Tests › Open menu item View>Appearance
```

bot please update galata snapshots (we may need to revert some spurious updates afterwards)

packages/application/src/shell.ts
```diff
@@ -1303,6 +1303,24 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     return layout;
   }
 
+  /**
+   * Toggle the fullscreen mode based on user inputs
+   */
+  setFullscreenMode(): void {
```
krassowski, line 1309:
```suggestion
  toggleFullscreenMode(): void {
```

galata/test/jupyterlab/notebook-run.test.ts-snapshots/notebook-panel-1-jupyterlab-linux.png
krassowski, line 1: Needs reverting

galata/test/jupyterlab/notebook-edit.test.ts-snapshots/split-cell-jupyterlab-linux.png
krassowski, line 1: Needs reverting

galata/test/jupyterlab/menus.test.ts-snapshots/opened-menu-view-appearance-jupyterlab-linux.png
krassowski, line 1: Good!

packages/application/src/shell.ts
dhysdrn: do you want us to make this change?

dhysdrn:

dhysdrn:
bot please update snapshots

dhysdrn:
@krassowski i think we made all the right changes, could you help us with this?

packages/application/src/shell.ts
```diff
@@ -1303,6 +1303,24 @@ export class LabShell extends Widget implements JupyterFrontEnd.IShell {
     return layout;
   }
 
+  /**
+   * Toggle the fullscreen mode based on user inputs
+   */
+  toggleFullscreenMode(): void {
```
krassowski, line 1309: Can I ask for one more thing - to move this method into a function in `packages/application-extension/src/index.tsx`?

The rationale is a follows: once we add something to `LabShell` class , any attempt to change it will be potentially breaking for an extension. Since there is further expansion of the fullscreen capabilities requested (to cover making individual elements take the full screen) enhancements, I think it is likely that the implementation may change.

dhysdrn: @krassowski we’ve removed the method and added it to `packages/application-extension/src/index.tsx`, let us know if it looks good

dhysdrn:
@krassowski Hey, just checking in 👋🏻 We would love to close this out and get it merged. Do let us know if we’re good or if we need anymore changes (:

packages/application-extension/schema/commands.json
```diff
@@ -47,6 +47,11 @@
       "keys": [""],
       "selector": "body"
     },
+    {
+      "command": "application:toggle-fullscreen-mode",
+      "keys": ["Alt F"],
```
krassowski, line 52: `Alt + F` is used by browsers to:
- open File menu (Firefox)
- open three dots menu (Chrome/Edge)

It may or may not be ok to override these. We can discuss this on JupyterLab call on Wednesday. Would it be also ok to just use F11 which is typically default for full screen? This way the contribution of this PR is:
- it adds a menu entry
- it allows user to change the shortcut
- it does not force a no-standard shortcut which is normally used for something else

krassowski (commented):
Testing:

- GNOME Web (webkit): works ok
- Chrome: works ok
- Firefox: works when clicking on the menu, does not work with `Alt + F` - it throws:

```
Request for fullscreen was denied because Element.requestFullscreen() was not called from inside a short running user-generated event handler. [index.tsx:458:37](webpack://jupyterlab/packages/application-extension/src/index.tsx)
Failed to enter fullscreen mode. TypeError: Fullscreen request denied
    execute index.tsx:458
    execute index.es6.js:365
    _executeKeyBinding index.es6.js:569
    processKeydownEvent index.es6.js:473
    evtKeydown lab.ts:217
```

In both Chrome and Firefox, entering full screen manually with F11 and then invoking the `application:toggle-fullscreen-mode` command forces the user to escape the full screen twice. This is odd.

krassowski (approved):
Thank you @dhysdrn! Tested in Chrome, Firefox, GNOME Web (Safari proxy) - at least with the default shortcuts it behaves well.

krassowski:
There is a failure because `F11` is also used for the debugger step in with the same selector (`body`):

```
    [jupyterlab] › test/jupyterlab/shortcuts.test.ts:29:5 › Shortcuts must be unique ───────────────
```


https://github.com/jupyterlab/jupyterlab/blob/fdd45e49e2ad059e4843a4dab915fb1a5cd59902/packages/debugger-extension/schema/main.json#L50-L74

PR #16312, opened 2024-05-09
Add Signal to proxy all Completer selected Signals

### LINKED ISSUE #16294: No consistent way to execute a callback function upon acceptance of a Completer item
### Problem

When accepting a item in the InlineCompleter you can connect a callback function to the `inline-completer:accept` command, but in the case of the Completer, the various equivalent commands `completer:select-notebook`, `completer:select-file`, and `completer:select-console` are not always executed when accepting an item.

The `Completer` class has a private Signal `_selected` that is triggered in the code in four places:
1. The `selectActive()` function which is connected to the various commands listed above
2. When the Completer only has one item it is automatically accepted
3. When the Completer is run with a query and only has one item it is automatically accepted
4. When the user selects the item by click on it with their mouse

This means that there is no way to connect a callback function to cases 2-4 since the command is not executed and the signal triggered is private.

### Proposed Solution

The most straight forward solution I see is finding a way to surface the private `_selected` Signal, this is already possible to do where the Completer class is accessible via the `selected()` function, but there is no current access of a Completer class to extensions. All calls to Completer instances are done via the private `_panelHandlers` Map in the `CompletionProviderManager` which is not surfaced.

If we create a accessor function to `_panelHandlers`, possibly a lookup method that with a widget id arg, to the `CompletionProviderManager` then any extension can get the Completer for a given widget and add a connection to the `selected` signal.

### Additional context

This is important to my use case because our Completer extension wants to track the metrics when users accept completion items.

This could be related to #15026, which would also solve this problem, but with a more complicated solution for this use case, trying the callback into the actual implementation of the insertText

ajbozarth:
## References

Fixes #16294

## Code changes

`Completer` currently emits a signal (`_selected`) when a completion is selected. This PR creates a new signal in `CompletionProviderManager`, which proxies the `Completer` signal from every `CompletionHandler` that is generated.

This will allow extensions to know when a selection is made by any Completer. Currently there are edge cases (as listed in  #16294) where a selection can be made without calling a command that can be connected to.

### DIFF
--- a/packages/completer/src/manager.ts
+++ b/packages/completer/src/manager.ts
@@ -10,6 +10,7 @@ import { CompletionHandler } from './handler';
 import { CompleterModel } from './model';
 import { InlineCompleter } from './inline';
 import {
+  ICompleterSelection,
   ICompletionContext,
   ICompletionProvider,
   ICompletionProviderManager,
@@ -36,7 +37,10 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     this._activeProvidersChanged = new Signal<ICompletionProviderManager, void>(
       this
     );
-    this._selected = new Signal<ICompletionProviderManager, string>(this);
+    this._selected = new Signal<
+      ICompletionProviderManager,
+      ICompleterSelection
+    >(this);
     this._inlineCompleterFactory = null;
   }
 
@@ -50,7 +54,7 @@ export class CompletionProviderManager implements ICompletionProviderManager {
   /**
    * Signal emitted when a selection is made from a completer menu.
    */
-  get selected(): ISignal<ICompletionProviderManager, string> {
+  get selected(): ISignal<ICompletionProviderManager, ICompleterSelection> {
     return this._selected;
   }
 
@@ -180,8 +184,8 @@ export class CompletionProviderManager implements ICompletionProviderManager {
       // Create a new handler.
       const handler = await this._generateHandler(newCompleterContext, options);
       this._panelHandlers.set(widget.id, handler);
-      handler.completer.selected.connect((sender, value) =>
-        this._selected.emit(value)
+      handler.completer.selected.connect((completer, text) =>
+        this._selected.emit({ text })
       );
       widget.disposed.connect(old => {
         this.disposeHandler(old.id, handler);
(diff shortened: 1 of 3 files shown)


packages/completer/src/manager.ts
```diff
@@ -172,6 +180,9 @@ export class CompletionProviderManager implements ICompletionProviderManager {
       // Create a new handler.
       const handler = await this._generateHandler(newCompleterContext, options);
       this._panelHandlers.set(widget.id, handler);
+      handler.completer.selected.connect((sender, value) =>
+        this._selected.emit(value)
```
krassowski, line 184: No objections to your approach in principle. The only thing that I would change is that the emit type ideally would not be a string, but an an object following an interface (which could contain the string value) so that we could later add more properties without making a breaking change. Maybe:  `Required<Pick<ICompletionItem, 'insertText'>>`.

ajbozarth: that's a great idea, I'll update it

ajbozarth: pushed the update, feel free to check it, I decided to call the interface `ICompleterSelection` since it is a selection coming from a completer. I also added the actual `Completer` object to the interface, but decided that if we ever need that in the future it can easily be added.

packages/completer/src/tokens.ts
```diff
@@ -378,6 +383,13 @@ export interface ICompletionProviderManager {
   inlineProviders?: IInlineCompletionProviderInfo[];
 }
 
+export interface ICompleterSelection {
+  /**
+   * The text selected by the completer.
+   */
+  text: string;
```
krassowski, line 390: I was thinking about using `insertText` to align with the other interfaces. Not a strong preference though.

```suggestion
  insertText: string;
```

packages/completer/src/manager.ts
```diff
@@ -46,6 +51,13 @@ export class CompletionProviderManager implements ICompletionProviderManager {
     return this._activeProvidersChanged;
   }
 
+  /**
+   * Signal emitted when a selection is made from a completer menu.
+   */
+  get selected(): ISignal<ICompletionProviderManager, ICompleterSelection> {
```
krassowski, line 57: Would it be easy to add a simple unit test for this signal being emitted in:

https://github.com/jupyterlab/jupyterlab/blob/7ae2d436fc410b0cff51042a3350ba71f54f4445/packages/metapackage/test/completer/manager.spec.ts#L105

?

packages/completer/src/tokens.ts
ajbozarth: No preference from me, I'll make this change

packages/completer/src/manager.ts
ajbozarth: ok, so after writing my comment saying no it started bothering me that I gave up half way and decided to just push through with the copying code from the surrounding text files and managed to write a 25 line test. It's pushed feel free to review

krassowski (approved):
Thank you @ajbozarth!

PR #16331, opened 2024-05-14
Remove Ctrl+Shift+Q shortcut for "Close and Shut Down [Notebook]"

### LINKED ISSUE #16330: CTRL+SHIFT+Q shortcut for "Close and Shut Down Notebook" does not work
On macOS Ventura 13.6.6 (22G630), with Firefox and Chrome, <kbd>CTRL+SHIFT+Q</kbd> appears in the File menu of Jupyter Notebook as a shortcut for "Close and Shut Down Notebook", but pressing this shortcut does not appear to do anything. In JupyterLab, this shortcut is bound to the same command, although the command does not appear in the file menu.

On Windows 10, with Firefox, pressing <kbd>CTRL+SHIFT+Q</kbd> immediately closes the browser and does not shut down Jupyter Notebook or JupyterLab.

Originally reported by @jph00 in https://github.com/jupyter/notebook/issues/7041.

JasonWeill:
## References

Fixes #16330.

## Code changes

Removes the <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>Q</kbd> keybinding for "Close and Shut Down [Notebook]", which either did not work or caused the user's entire browser to quit, particularly on Windows, where <kbd>CTRL</kbd> is a common accelerator key for browser commands in our supported browsers.

## User-facing changes

Users will no longer be able to trigger a command with <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>Q</kbd>, even in browsers and on OSes where this previously worked. 

No public API changes.


JasonWeill:
please update snapshots

krassowski:
I think we need to finish https://github.com/jupyterlab/jupyterlab/pull/14015 so that users can add it the shortcut back from the UI.

krassowski:
This is no longer blocked as https://github.com/jupyterlab/jupyterlab/pull/18163 allows users to add shortcuts from UI relatively easily. It does require rebasing due to a number of conflicts.

krassowski:
> Users will no longer be able to trigger a command with CTRL+SHIFT+Q, even in browsers and on OSes where this previously worked.

On Linux/Edge I see that it works correctly:

image

The original report was saying that this does not work.

image

I am not convinced that removing the shortcut for everyone is the right call. I will bump it to 4.7 and add "needs discussion" label to chat about this on a future developers call.

PR #16341, opened 2024-05-17
Shut down unused kernels

### LINKED ISSUE #13916: Add button closing dangling kernels
> The "dangling" kernel is actually a real live kernel session that still exists but had no visibility before.

It was visible before too (and is on 3.x and 2.x). This is the where users refer to when they run low on memory (killing old kernels from closed notebooks that were previously closed). Maybe we even should have a button for "Kill all kernels which do not have any notebook/console".

On this use-case we could have something like this:

Screenshot from 2023-01-24 19-53-59

> Note: the numbers in circles on the graphic are irrelevant; those were from an exploration suggesting how to show the number of documents/widgets connected to the kernel.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/13779#issuecomment-1402523525_

NexVeridian:
## References
closes #13916

## Code changes

Adds a button to Shut Down Unused kernels

## User-facing changes
image


krassowski:
- It looks like there are conflicts which need to be resolved, these seem to prevent CI from running
- can you add a screenshot to the top-level comment?
- "only show button in kernels section" - shall we allow `IRunningSessions.IManager` to contribute arbitrary buttons to the section toolbars?

packages/running-extension/src/kernels.tsx
```diff
@@ -41,6 +44,9 @@ export async function addKernelRunningSessionManager(
   const { runningChanged, RunningKernel } = Private;
   const throttler = new Throttler(() => runningChanged.emit(undefined), 100);
   const trans = translator.load('jupyterlab');
+  const shutdownUnusedLabel = trans.__('Shut Down Unused');
+  const runningKernels = await KernelAPI.listRunning();
```
NexVeridian, line 48: when I moved everything to this file, runningKernels is not refresh the same as the other file
this file seems to be on page reload, while the other seems to be on kernel update

also the kernels section got moved to the end of the list for some reason

krassowski: `CommandToolbarButton` may be easier to use use here, although it will require defining a command. It would be easier because the commands have `isVisible() => boolean` and `isEnabled()` functions that the `CommandToolbarButton` will respect. If you need more hints on this one let me know.

> also the kernels section got moved to the end of the list for some reason

maybe do to the await in this line (`await KernelAPI.listRunning();`)?

packages/running-extension/src/kernels.tsx
```diff
@@ -49,6 +53,47 @@ export async function addKernelRunningSessionManager(
   // Wait until the relevant services are ready.
   await Promise.all([kernels.ready, kernelspecs.ready, sessions.ready]);
 
+  commands.addCommand(SHUTDOWN_UNUSED_BUTTON_CLASS, {
+    label: shutdownUnusedLabel,
+    execute: async () => {
+      const runningKernels = await KernelAPI.listRunning();
+      // const enabled = runningKernels.length > 0;
```
NexVeridian, line 60: How would I set the isVisible()

packages/running-extension/src/kernels.tsx
```diff
@@ -49,6 +58,59 @@ export async function addKernelRunningSessionManager(
...
+
+      if (confirmed.button.accept) {
+        for (const kernel of unusedKernels) {
+          await KernelAPI.shutdownKernel(kernel.id);
+        }
+        Promise.all([kernels.refreshRunning(), sessions.refreshRunning()]);
+      }
+    },
+    isEnabled: () => {
+      shutdownUnusedThrottler.invoke();
```
krassowski, line 109: Do we want to have it here, or should this be invoked only after user shut downs the kernels, and in response to `kernels.runningChanged`?

I am worried that it is either not awaited as now so it leads to inconsistent state, or that if it were awaited it would lead to delayed rendering.

packages/running-extension/src/kernels.tsx
```diff
@@ -49,6 +58,59 @@ export async function addKernelRunningSessionManager(
...
+    // Identifies unused kernels
+    const runningKernels = await KernelAPI.listRunning();
+    return runningKernels.filter(
+      kernel => kernel.connections !== undefined && kernel.connections < 1
+    );
+  }
+
+  async function checkShutdownUnusedEnabled() {
+    shutdownUnusedEnabled = (await getUnusedKernels()).length > 0;
+    commands.notifyCommandChanged(SHUTDOWN_UNUSED_BUTTON_CLASS);
```
krassowski, line 71: This needs to check whether the status changed before notifying, otherwise it would invoke throttler in an infinite loop (although throttled to once every 5 seconds) because `isEnabled` invokes it and it is invoked when the command changes. 

```suggestion
    const wasEnabled = shutdownUnusedEnabled;
    shutdownUnusedEnabled = (await getUnusedKernels()).length > 0;
    if (wasEnabled !== shutdownUnusedEnabled) {
      commands.notifyCommandChanged(SHUTDOWN_UNUSED_BUTTON_CLASS);
    }
```

packages/running-extension/src/kernels.tsx
```diff
@@ -49,6 +58,59 @@ export async function addKernelRunningSessionManager(
   // Wait until the relevant services are ready.
   await Promise.all([kernels.ready, kernelspecs.ready, sessions.ready]);
 
+  async function getUnusedKernels() {
+    // Identifies unused kernels
+    const runningKernels = await KernelAPI.listRunning();
+    return runningKernels.filter(
+      kernel => kernel.connections !== undefined && kernel.connections < 1
+    );
```
krassowski, line 66: Is this information possibly available from `kernels` service already?

krassowski (commented):
This is looking better - leaving some more feedback on this iteration.

packages/running-extension/src/kernels.tsx
NexVeridian: It is, though when I changed it to Array.from(kernels.running()), the code stopped working
every kernel would always get shutdown, regardless of connection count

I'm not sure why, If I print both I don't see a difference?

I reverted it back to await KernelAPI.listRunning() which works correctly

await KernelAPI.listRunning()
```
[
  {
    "id": "5349a541-b573-4c75-aaf5-efd484103e5f",
    "name": "python3",
    "last_activity": "2024-06-04T01:37:31.710898Z",
    "execution_state": "idle",
    "connections": 0
  },
  {
    "id": "1686c8f7-dc10-4122-938c-0ae73c5cc3f6",
    "name": "python3",
    "last_activity": "2024-06-04T01:37:39.079313Z",
    "execution_state": "idle",
    "connections": 1
  }
]
```

Array.from(kernels.running())
```
[
  {
    "id": "5349a541-b573-4c75-aaf5-efd484103e5f",
    "name": "python3",
    "last_activity": "2024-06-04T01:37:31.710898Z",
    "execution_state": "idle",
    "connections": 0
  },
  {
    "id": "1686c8f7-dc10-4122-938c-0ae73c5cc3f6",
    "name": "python3",
    "last_activity": "2024-06-04T01:37:39.079313Z",
    "execution_state": "idle",
    "connections": 1
  }
]
```

packages/running-extension/src/kernels.tsx
```diff
@@ -87,7 +157,14 @@ export async function addKernelRunningSessionManager(
     shutdownAllLabel: trans.__('Shut Down All'),
     shutdownAllConfirmationText: trans.__(
       'Are you sure you want to permanently shut down all running kernels?'
-    )
+    ),
+    toolbarButtons: [
+      new CommandToolbarButton({
+        commands,
+        id: SHUTDOWN_UNUSED_BUTTON_CLASS,
```
krassowski, line 164: From a quick glance the logic looks all good now. The last thing I would suggest is using an icon instead of text for this button. Let's first choose and add an icon to the command. I think some kind of material design icon for cleaning could do?

```suggestion
        icon: someIcon,
        id: SHUTDOWN_UNUSED_BUTTON_CLASS,
```

NexVeridian: image

I added one, I don't see very many options
I manually added it to `/jupyterlab/packages/ui-components/src/icon/iconimports.ts` which needs to be fixed since it's supposed to be auto-generated

Should the text for `Shut Down All` and `Shut Down Unused` be removed, and just have icons instead, since there's not that much room

packages/running-extension/src/kernels.tsx
```diff
@@ -41,14 +45,81 @@ export async function addKernelRunningSessionManager(
...
 
   // Wait until the relevant services are ready.
   await Promise.all([kernels.ready, kernelspecs.ready, sessions.ready]);
 
+  async function getUnusedKernels() {
+    // Identifies unused kernels
+    const runningKernels = await KernelAPI.listRunning();
+    return runningKernels.filter(
+      kernel => kernel.connections !== undefined && kernel.connections < 1
+    );
```
fcollonval, line 70: Using the following code works for me:

```suggestion
  function getUnusedKernels() {
    // Identifies unused kernels
    return Array.from(kernels.running()).filter(
      kernel => (kernel.connections ?? 1) < 1
    );
```

packages/running-extension/src/kernels.tsx
```diff
@@ -41,14 +45,81 @@ export async function addKernelRunningSessionManager(
...
+      const unusedKernels = await getUnusedKernels();
+
+      if (unusedKernels.length === 0) {
+        void showDialog({
+          title: trans.__('No Unused Kernels'),
+          body: trans.__('There are no unused kernels to shut down.'),
+          buttons: [Dialog.okButton()]
+        });
+        return;
+      }
```
fcollonval, line 93: I would prefer we bail early without displaying a dialog here as the commands is disabled when no unused kernel are found.

```suggestion
      const unusedKernels = getUnusedKernels();

      if (unusedKernels.length === 0) {
        return;
      }
```

packages/running-extension/src/kernels.tsx
```diff
@@ -41,14 +45,81 @@ export async function addKernelRunningSessionManager(
   const { runningChanged, RunningKernel } = Private;
   const throttler = new Throttler(() => runningChanged.emit(undefined), 100);
   const trans = translator.load('jupyterlab');
+  const shutdownUnusedLabel = trans.__('Shut Down Unused');
+  let shutdownUnusedEnabled = false;
+  const shutdownUnusedThrottler = new Throttler(
+    checkShutdownUnusedEnabled,
+    5000
```
fcollonval, line 52: Let's not be more aggressive than the kernel list poll:

```suggestion
    10000
```

fcollonval (dismissed):
Thanks @NexVeridian 

A couple of improvements but it looks quite good.

Questions for every body
- should we positioned the additional toolbar button on the left of the shutdown all button - especially as we are gonna use an icon for this new feature.
- Should we display more than the kernel name in the confirmation dialog? I think yes as it may confused the user.

packages/running-extension/src/kernels.tsx
```diff
@@ -41,14 +45,81 @@ export async function addKernelRunningSessionManager(
...
+        return;
+      }
+
+      const confirmed = await showDialog({
+        title: shutdownUnusedLabel,
+        body: trans.__(
+          `Are you sure you want to shut down the following unused kernels?\n\n${unusedKernels
+            .map(kernel => kernel.name)
+            .join('\n')}`
+        ),
```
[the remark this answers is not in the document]
NexVeridian: > Should we display more than the kernel name in the confirmation dialog? I think yes as it may confused the user.

you could add `kernel.id` here, which would be like:
python3 (7e3e5c2d)

the file name nice would probably be nicer but that's not in `kernel`, so `kernel.id` would be easier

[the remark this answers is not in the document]
fcollonval: Adding the id is already a good step as the user will see it in the tree view.

krassowski:
@NexVeridian it looks like this picked up a few conflicts, can you resolve them easily?

packages/running-extension/src/kernels.tsx
```diff
@@ -41,14 +45,76 @@ export async function addKernelRunningSessionManager(
...
+  async function checkShutdownUnusedEnabled() {
+    const wasEnabled = shutdownUnusedEnabled;
+    shutdownUnusedEnabled = (await getUnusedKernels()).length > 0;
+    if (wasEnabled !== shutdownUnusedEnabled) {
+      commands.notifyCommandChanged(SHUTDOWN_UNUSED_BUTTON_CLASS);
+    }
+  }
+
+  commands.addCommand(SHUTDOWN_UNUSED_BUTTON_CLASS, {
+    caption: args => (args.toolbar ? '' : shutdownUnusedLabel),
```
krassowski, line 81: To address:

```
    [documentation] › test/documentation/commands.test.ts:7:5 › All commands must have a default label 
```

What was the motivation to change it to caption in the first place?

```suggestion
  commands.addCommand(SHUTDOWN_UNUSED_BUTTON_CLASS, {
    label: args => (args.toolbar ? '' : shutdownUnusedLabel),
    caption: args => (args.toolbar ? '' : shutdownUnusedLabel),
```

NexVeridian: I though we wanted just a icon and no text?
so caption removes the text and keeps showing `Shut Down Unused` on hover

# caption
cap

# label
lab

krassowski: Yes, we want to hide the label on the button level but we always need to keep the label on command level. I will look into this and push some commits to get it over the finish line.

krassowski: When changing it to `label` locally it works as expected:

image

This is because `args.toolbar` checks whether it is in a toolbar and hides the label there accordingly.

packages/running/src/index.tsx
```diff
@@ -559,6 +560,12 @@ class Section extends PanelWithToolbar {
         ? 'tree'
         : 'list';
     this.addWidget(this._listWidget);
+
+    if (this._manager.toolbarButtons) {
+      this._manager.toolbarButtons.forEach(button =>
+        this.toolbar.addItem(button.id, button)
```
krassowski, line 566: `button.id` is empty for command toolbar buttons. We may need to change that or, the API.

krassowski: Solved in https://github.com/jupyterlab/jupyterlab/pull/16341/commits/88ba6852382b197656b04ef753bbb404d049c33f

krassowski:
bot please update snapshots

krassowski:
please update snapshots

krassowski (approved):
This looks good to me know, thank you @NexVeridian!

krassowski:
Merging to release for testing in beta, further comments are welcome and I will happily work to accommodate requests in the beta phase :)

Issue #16319, opened 2024-05-11
Shift + tab does not open inspector tooltip on 4.2.0


krassowski:
## Description

Shift + tab does not open inspector tooltip on 4.2.0

as reported on https://stackoverflow.com/questions/78464382/shift-tab-isnt-showing-docstring-anymore

## Reproduce

1. Type `int`
2. Press Shift + Tab

## Expected behavior

- There is an integration test case for this shortcut
- The shortcut works

## Context

- Works in 4.1.0
- Works in 4.2.0b0
- Does not work in 4.2.0b1 - it could be related to https://github.com/jupyterlab/jupyterlab/pull/16078 or another PR merged in 4.2.0b1 (https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0b1)

krassowski:
So this looks to be another side-effect of https://github.com/jupyterlab/lumino/pull/662 but in this case it seems to be because `indentMoreOrInsertTab` returns `true` on Shift + Tab:

![Screenshot from 2024-05-18 19-17-37](https://github.com/jupyterlab/jupyterlab/assets/5832902/bc794b2e-3811-457c-962f-6bb8ac48156b)

PR #16343, opened 2024-05-18
Fix `Shift + Tab` contextual help inspector tooltip regression

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16319

## Code changes

- [x] adds an integration test case
- [x] fixes the issue by not preventing default (i.e. not returning `true`) in the `dedentIfNotLaunchingTooltip` handler

## User-facing changes

Shift + Tab works again

None


fcollonval (approved):
Thanks @krassowski

Issue #16367, opened 2024-05-23
Console error when launching some kernels


afshin:
The debugger extension tries to read the `msg_type` header of a type cast iopub message header, but because it can be `null` sometimes, this creates a runtime error in the console. I ran into this circumstance trying to launch a xeus-python raw kernel:

https://github.com/jupyterlab/jupyterlab/assets/159529/721659d5-76ed-46cc-9d25-4cbafe19b0fe

## Description
The debugger extension listens to all iopub messages of a kernel

## Reproduce

1. Install xeus-python
2. Launch JupyterLab in your browser and open the developer console
3. Click the "raw" kernel version in the launcher to create a new notebook
4. See error in console

## Expected behavior
No error in console

krassowski:
> because it can be `null` sometimes

When can the header be `null` as in the specs? If it is in the spec we should also correct the typing.

afshin:
> > because it can be `null` sometimes
> 
> When can the header be `null` as in the specs? If it is in the spec we should also correct the typing.

To be honest, I am not sure whether this is a bug in xeus-python or actually allowed by the spec. This is why I didn't actually make the typecasting any more or less permissive.

Instead I just changed the order in which things are checked because strictly speaking, we should not inspect any messages in the debugger extension unless it is running.

JohanMabille:
> When can the header be `null` as in the specs? If it is in the spec we should also correct the typing.

The "welcome" message sent by the kernel on iopub has no parent header (and I think it's the only one). It was introduced in this [JEP](https://github.com/jupyter/enhancement-proposals/pull/65), and has been implemented in the xeus kernels, but probably not in ipykernel yet (probably because we haven't released a version of the protocol including this feature).

PR #16368, opened 2024-05-23
Fix runtime console error in debugger extension

afshin:
The debugger extension tries to read the `msg_type` header of a type cast iopub message header, but because it can be `null` in some cases, this creates a runtime error in the console.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/16367

## Code changes
With this PR, the debugger extension only inspects iopub message headers if the debugger service is running. This means that early iopub messages without headers are never intercepted and the console error is avoided.

## User-facing changes
N/A

N/A


krassowski (approved):
Thank you!

krassowski:

Issue #16340, opened 2024-05-17
"preferred" in "Always start the preferred kernel" is not documented


krassowski:
## Description

- the code name uses "default" but the user-facing strings say "preferred"
- it is not explained anywhere what "preferred" means:
   https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/notebook-extension/schema/tracker.json#L696-L700
    https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/apputils/src/sessioncontext.tsx#L1357-L1363
- the preference seems to refer to "default" designation in `IKernelSearch.options.preference` which is `ISessionContext.IKernelPreference`
    https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/apputils/src/sessioncontext.tsx#L1504-L1557

- the `kernelPreference` seems to be an in-memory variable:
   https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/apputils/src/sessioncontext.tsx#L348
    https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/apputils/src/sessioncontext.tsx#L463-L476

- only `autoStartDefaultKernel` but not the id of the default/preferred kernel gets stored in settings
   https://github.com/jupyterlab/jupyterlab/blob/44b82c8c55880a02c088ebd9576da807dc471b48/packages/notebook-extension/src/index.ts#L1655-L1693

## Expected behavior

- The label in kernel selection dialog is more useful, explaining what "preference" means; maybe it should not say "preferred kernel" but "selected kernel". Maybe it should not say "Always" but "for the duration of this session" (in my understanding the setting is persisted forever, but the "preference" is not)
- The setting description should explain the behaviour in more detail

## Context

- JupyterLab version: 4.2.0

ericdatakelly:
"selected kernel" makes a lot of sense. And clarifying that it's for the duration of the session would be very helpful.

Gecervantes01:
I'd like to take on this issue!

ctcjab:
> Maybe it should not say "Always" but "for the duration of this session" (in my understanding the setting is persisted forever, but the "preference" is not)

[...]

PR #16378, opened 2024-05-26
Updated wording for "Change Kernel" checkbox

Gecervantes01:
## References
Issue #16340 

## Code changes
Made the requesting wording changes to better describe the function of the checkbox within the "Change Kernel" setting.

## User-facing changes
When changing the kernel, user will now see the updated text next to the checkbox. Rather than using the word "Always", it now says "During" implying the selection only lasts for the duration of the session. Also changed "preferred" to "selected" to better match the wording in the rest of the popup.

For added clarity, the tooltip explicitly says, "Remember my choice for the duration of this session and start the selected kernel".

Before:
before

After:
after

N/A


Gecervantes01:
@meeseeksmachine tag documentation

PR #16381, opened 2024-05-27
Updated wording in "Change Kernel" settings

Gecervantes01:
## References
Issue #16340 

## Code changes
Made the requesting wording changes to better describe the function of the checkbox within the "Change Kernel" setting.

## User-facing changes
When changing the kernel, user will now see the updated text next to the checkbox. Rather than using the word "Always", it now says "During" implying the selection only lasts for the duration of the session. Also changed "preferred" to "selected" to better match the wording in the rest of the popup.

For added clarity, the tooltip explicitly says, "Remember my choice for the duration of this session and start the selected kernel".

Before:
before

After:
after

N/A

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -694,8 +694,8 @@
       "default": "code"
     },
     "autoStartDefaultKernel": {
-      "title": "Automatically Start Preferred Kernel",
-      "description": "Whether to automatically start the preferred kernel.",
+      "title": "Automatically Start Selected Kernel",
+      "description": "Whether to automatically start the Selected kernel. Kernel selection is made within the toolbar in the 'Kernel' window under the option 'Change Kernel...'.",
       "type": "boolean",
       "default": false
     },
--- a/packages/apputils/src/sessioncontext.tsx
+++ b/packages/apputils/src/sessioncontext.tsx
@@ -1392,9 +1392,9 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
       buttons,
       checkbox: hasCheckbox
         ? {
-            label: trans.__('Always start the preferred kernel'),
+            label: trans.__('During this session start the selected kernel'),
             caption: trans.__(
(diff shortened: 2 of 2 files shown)


Gecervantes01:
@meeseeksdev tag documentation

packages/notebook-extension/schema/tracker.json
```diff
@@ -694,8 +694,8 @@
       "default": "code"
     },
     "autoStartDefaultKernel": {
-      "title": "Automatically Start Preferred Kernel",
-      "description": "Whether to automatically start the preferred kernel.",
+      "title": "Automatically Start Selected Kernel",
+      "description": "Whether to automatically start the Selected kernel.",
```
krassowski, line 698: When user is in the Settings window, the "Selected kernel" is not clear because they do not have a way to make a selection in Settings. We need to describe when the selection happens.

krassowski (commented):

Gecervantes01:
@meeseeksdev tag enhancement

packages/notebook-extension/schema/tracker.json
Gecervantes01: @krassowski I went ahead and updated the description, making it clear where the user can make a kernel selection.

krassowski:
@Gecervantes01 this PR seems to include unrelated changes related to full screen feature. If we want to include it 4.3 we will need to review the clean version soon - do you have time to clean it by next week?

krassowski:
I cleaned the commit history.

krassowski (commented):
I am not sure of the wording here, especially for the setting.

This is because https://github.com/jupyterlab/jupyterlab/pull/16475 adjusted the logic in the selector and clarified that that _"preferred" means "using the programming language I used last time"_. I am not sure how accurate this is in the context of `autoStartDefaultKernel` and the checkbox.

hxrshxz:
@Gecervantes01 are you still working on this? if not please tell because i would like to work on this

krassowski:
Closing since there was no follow-up and it looks like we need to iterate on the wording more

Issue #16380, opened 2024-05-27
The width and top margin of the `Click to add a cell` element should be fixed


HaudinFlorence:
## Description

This issue is related to the width and top margin of the `Click to add a cell area` element
![Screenshot from 2024-05-27 15-07-16 png](https://github.com/jupyterlab/jupyterlab/assets/99649086/18479ee4-af7c-4064-9b50-9a9160878b41)

This feature that was in the classic notebook and not yet in jupyterlab has been introduced last year in PR https://github.com/jupyterlab/jupyterlab/pull/14109 and initially the dimensions were supposed to be as shown in the following gif extracted from the cited PR:
![initial_behavior](https://github.com/jupyterlab/jupyterlab/assets/99649086/2de59f58-4710-40bc-8c7e-3e4eb849cf9c)
These styling seems to have been broken at some point.

## Reproduce

1. Open a new jupyter lab
2. Put you mouse below the last cell: the `Click to add a cell` element should appear
3. Observe its dimensions: it is wider that the cell input and the gap between the last cell and this element is smaller than the gap between 2 cell inputs.

## Expected behavior

The width of the `Click to add a cell area` should be the same as the cell input ones and the space between this element and the last cell should also be the same as the space between the different cells' input.

## Context

- Operating System and version: 
- Browser and version: 
- JupyterLab version: 

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
Paste the output from your browser Javascript console here, if applicable.

</pre>
</details>

krassowski:
Are there any visual tests with this button?

MMesch:
The triaging agent has identified this issue as likely resolved. I just tried and it seems indeed fixed.

<img width="1285" height="121" alt="Image" src="https://github.com/user-attachments/assets/44487ea6-ac8a-4612-8826-e68a620d2409" />

I'm not sure what the policy for adding visual tests is, whether it is worth here or not and how much work it would be. @krassowski @HaudinFlorence could you either close this as completed or leave a comment making it very explicit that visual test is required to complete this?

PR #16383, opened 2024-05-28
Fix width and margins of the notebook footer.

HaudinFlorence:
This PR fixes the width and the margins of the notebook footer.

## References

This PR should fix issue #16380 
## Code changes

The proposes changes only concerns the style of the notebook footer and are in `packages/notebook/style/notebookfooter.css`


HaudinFlorence:
bot please update snapshots

galata/test/documentation/general.test.ts-snapshots/command-palette-documentation-linux.png
krassowski, line 1: These snapshot changes appear unrelated and are causing the tests to fail I think:

```
Notice:   3 failed
    [documentation] › test/documentation/general.test.ts:618:7 › General › Command Palette ─────────
    [documentation] › test/documentation/internationalization.test.ts:14:7 › Internationalization › Menu 
    [documentation] › test/documentation/internationalization.test.ts:28:7 › Internationalization › Confirm language 
```

krassowski (commented):
Thank you for working on a fix! It looks good in the (now default) full windowing mode:

Screenshot from 2024-05-28 20-55-37

but in the `none` mode it seems that the spacing is too much:

Screenshot from 2024-05-28 20-55-27

packages/notebook/style/notebookfooter.css
```diff
@@ -5,20 +5,20 @@
...
     100% -
       (
-        var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) +
-          var(--jp-cell-padding) + var(--jp-cell-padding)
+        var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
+          var(--jp-cell-padding) + 2 * var(--jp-notebook-padding)
       )
   );
-  border: var(--jp-border-width) solid var(--jp-cell-editor-border-color);
+  border: 0.8px solid var(--jp-cell-editor-border-color);
```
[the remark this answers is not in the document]
HaudinFlorence: @krassowski I can revert this change.

galata/test/documentation/general.test.ts-snapshots/command-palette-documentation-linux.png
HaudinFlorence: Do you mean that I should take back the snapshots from the main branch ?

krassowski: Yes, I think that after that these tests should pass. If not, then we need to investigate (it looks like maybe different runners get different variants of Chinese fonts?)

krassowski:
@HaudinFlorence can you test the changes I pushed?

HaudinFlorence:
> @HaudinFlorence can you test the changes I pushed?

@krassowski Thanks for working on this PR.  In the none mode, the width and left margin are ok but the footer remains at it initial position when adding new cells:
file:///home/fhaudin/Videos/Screencasts/Screencast%20from%2005-06-2024%2017:26:31.webm

Note also that in both modes, the top margin of the footer is slightly too large

krassowski:
Thank you for this recording! I saw this happen earlier on CI (https://github.com/jupyterlab/jupyterlab/issues/14385); I now realise what is happening - the button is positioned below the spacer `jp-WindowedPanel-inner` element, which has either:
- managed height when `full` windowing mode is active
- unset height when `defer` or `none` is used

I did not see the issue you pointer out in testing because I was reloading JupyterLab or opening a new notebook after changing settings. I think that you did not reload JupyterLab after the change and this revealed a bug in that the previously set `height` of `jp-WindowedPanel-inner` remained set after switching to `none` but it was no longer managed so it stuck on an incorrect value.

This should be fixed by aafff6e1409aff45e585a1a0f10dcf14b61c8bec.

> Note also that in both modes, the top margin of the footer is slightly too large

They seem fine to me. If there was any margin added explicitly I would accommodate your suggestion by reducing it, but since there is no explicit `margin-top` anymore I think maybe we can leave it as is?

krassowski (approved):

krassowski:

krassowski:
FYI I opened two follow-up issues for edge cases:
- https://github.com/jupyterlab/jupyterlab/issues/16491
- https://github.com/jupyterlab/jupyterlab/issues/16492

PR #16387, opened 2024-05-28
Use lazy loading for HTML/PDF files, and help tabs

### LINKED ISSUE #15334: Defer loading of iframes to accelerate JupyterLab start
### Problem

JupyterLab opens HTML/PDF/help documents using `iframe`s. If those are open in the background like:

image

they are still loaded immediately when JupyterLab opens.

### Proposed Solution

Using [lazy](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#lazy) loading of iframes can cut the loading time for the active tab as the browser will defer loading/populating the iframe.

Benchmarking would be beneficial to establish how much it can help with loading time.

### Additional context

This was supported by Chrome for a long time but is only rolling out in Firefox in v120 (upcoming version).
https://caniuse.com/loading-lazy-attr

hnben:
## References 

Closes #15334 

## Code changes

Added a line of code to set the lazy loading attribute on iFrames in widgets for rending HTML and help documents. Modified the RenderedPDF class to include lazy loading attribute on iFrame

HTML 
HTML

PDF
PDF

Help
HELP

Inspecting elements in the browser's dev tools indicate the attributes should be set correctly.

## User-facing changes

N/A
N/A


hnben:
@meeseeksdev tag enhancement

packages/help-extension/src/index.tsx
```diff
@@ -232,6 +232,8 @@ const open: JupyterFrontEndPlugin<void> = {
       content.addClass(HELP_CLASS);
       content.title.label = text;
       content.id = `${namespace}-${++counter}`;
+      //sets loading to lazy
+      content.node.querySelector('iframe')?.setAttribute('loading', 'lazy');
```
krassowski, line 236: It would be good to move this implementation detail into `IFrame` implementation, this is (a) add an option for `loading` (b) pass this option into the constructor here in line 229 and in the other file where it is used:

https://github.com/jupyterlab/jupyterlab/blob/b2ff3fbbe60e1d08d23f6a4738d4f2d76f175152/packages/ui-components/src/components/iframe.ts#L9-L13

krassowski (commented):
Thank you for starting this!

One suggestion on where to put the implementation for 2/3 files.

packages/help-extension/src/index.tsx
hnben: Hi @krassowski if you're able to check out the changes I made. I added your suggestion of having an option for loading in the constructor of iFrame and updated the html and help rendering widgets to have the loading option set to lazy. Thanks!

Lois-Lanctot:
Hi @krassowski we added the changes you requested, is there anything else we should be doing? Also, we're a little concerned about one of the Linux tests failing, is that something we should address? Thank you!

krassowski:
I merged in the `main` branch which should fix the failing test. The code looks good but I am yet to test it locally - thank you for your patience.

PR #16390, opened 2024-05-29
Add the toolbar again when updating the title of PanelWithToolbar

brichet:
This PR restore the toolbar of a `PanelWithToolbar` when the title is updated.

When any update occurs on a Lumino `AccordionPanel`, the title node is built from scratch https://github.com/jupyterlab/lumino/blob/63a0e8d581795342a1d416c8a47d344b3f5f4743/packages/widgets/src/accordionlayout.ts#L80

But the toolbar of a panel with toolbar is only added when the panel is attached the first time https://github.com/jupyterlab/jupyterlab/blob/b2ff3fbbe60e1d08d23f6a4738d4f2d76f175152/packages/ui-components/src/components/accordiontoolbar.ts#L84

After this PR, the toolbar node is attached to the new title node after update.

## References

None

## Code changes

- adds an `updateTitle()` function in `AccordionToolbarLayout` to restore the toolbar after any update in the title.
- add a test on it.

## User-facing changes

None

None

### DIFF
--- a/packages/ui-components/src/components/accordiontoolbar.ts
+++ b/packages/ui-components/src/components/accordiontoolbar.ts
@@ -64,8 +64,7 @@ class AccordionToolbarLayout extends AccordionLayout {
   }
 
   /**
-   * Function is called when updating the title. Updating the title build a new Title
-   * widget, we need to attach again the toolbar node to it.
+   * Called when a title attribute has changed, to attach again the toolbar node.
    */
   updateTitle(index: number, widget: Widget): void {
     super.updateTitle(index, widget);
--- a/packages/ui-components/src/components/accordiontoolbar.ts
+++ b/packages/ui-components/src/components/accordiontoolbar.ts
@@ -64,6 +64,15 @@ class AccordionToolbarLayout extends AccordionLayout {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/accordiontoolbar.ts
```diff
@@ -63,6 +63,15 @@ class AccordionToolbarLayout extends AccordionLayout {
     }
   }
 
+  /**
+   * Function is called when updating the title. Updating the title build a new Title
+   * widget, we need to attach again the toolbar node to it.
```
krassowski, line 68: The "Updating the title build a new Title widget" part of the comment is not clear to me, it looks like it might be an editing error?

krassowski (commented):
Thanks! Just one tiny request.

packages/ui-components/src/components/accordiontoolbar.ts
brichet: Thanks for you review @krassowski.
I agree that the comment was unclear, I simplified it.

krassowski (approved):
Thank you!

krassowski:
CI failure is unrelated (https://github.com/jupyterlab/jupyterlab/issues/16404)

krassowski:

Issue #16332, opened 2024-05-15
Encoding more information in the virtual scroll bar


mlucool:
### Problem

The virtual scroll bar is nice, but now that its a custom scrollbar, it seems like we can use it to encode more information. 
### Proposed Solution

For example, maybe code and markdown cells should look slightly different. We could also imagine encoding something about which are the big code cells with some sort of very slim sparkline. I'm not sure of any exact idea - but it seems navigation could be greatly enhanced by hinting at what is in the cell. Other ideas including showing which cells have an error etc.

![image](https://github.com/jupyterlab/jupyterlab/assets/1813603/0bb3f515-97b3-4de7-9adc-5519198baa6c)
Above cell 3 is markdown (gray) and cell 2 has more than 2x the code than cell 4.

krassowski:
Related to:
- https://github.com/jupyterlab/jupyterlab/issues/3738
- https://github.com/bollwyvl/ipython-notebook-navigator

We do not want to render a mini version of the cell in the scrollbar as this would take away the performance benefits of virtualization.

- We already have the size information about cells and outputs, we could resize the scrollbar cells accordingly.
- I think showing the execution count number by the code cells could be very useful, because right now there is a disconnect between cell index and the execution number which users see in the notebook. I think the cell index is not needed if we have a rich representation of the cells, but it could still be shown on hover.
- For code we could store information about number of lines in the editor, and:
  - use 1-3 variants like:
     - ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/7db973b7-abdd-4084-8380-37c1322b3bb9) for an editor with a lot of text/code,
     - ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/bad43288-9de0-48f0-bbf2-fc66c14dedf1) for an editor with medium amount of text, and
     - ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/f451ad05-3738-4792-8386-564c202b93a2) for an editor with little text.
  - or we could even generate the number of lines (and lengths) dynamically to to match the text content; this will still be way faster to render than syntax highlighted code
- Users will be interested to find cells with specific outputs, cells which are executing, and cells which failed to execute; we could show:
  - ![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/fe64fe5f-2d44-4c40-9aaf-68ee67bf1045) a spinner in output section of cells which are executing
[...]

PR #16392, opened 2024-05-29
Implement viewport tracking and reactive rendering in scrollbar

krassowski:
## References

- a step towards https://github.com/jupyterlab/jupyterlab/issues/16332

## Code changes

- `WindowedList.IRenderer.createScrollbarItem` can now return either `HTMLElement` (as before) or `IScrollbarItem` defined as:
  ```ts
  interface IScrollbarItem extends IDisposable {
    /**
     * Render the scrollbar item as an HTML element.
     */
    render: (props: { index: number }) => HTMLElement;
  
    /**
     * Unique item key used for caching.
     */
    key: string;
  }
  ```
- The entire scrollbar is no longer re-rendered on each update of the notebook; this enables adding more reactive information and updating only the cells that have changed; instead `IScrollbarItem.render` is called to modify the node reactively
- a new viewport indicator element is added which shows which cells are visible in the viewport (styling subject to change); this element is independent of cells to avoid scrolling causing re-render of items

## User-facing changes

viewport-indicator

None

### DIFF
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -1284,7 +1284,7 @@ export class WindowedList<
         const scrollbarItems = this._renderScrollbar();
         const first = scrollbarItems[firstVisibleIndex];
         const last = scrollbarItems[lastVisibleIndex];
-        this._viewportIndicator.style.top = first.offsetTop + 'px';
+        this._viewportIndicator.style.top = (first.offsetTop - 1) + 'px';
         this._viewportIndicator.style.height =
           last.offsetTop - first.offsetTop + last.offsetHeight + 'px';
       }
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -1284,7 +1284,7 @@ export class WindowedList<
         const scrollbarItems = this._renderScrollbar();
         const first = scrollbarItems[firstVisibleIndex];
         const last = scrollbarItems[lastVisibleIndex];
-        this._viewportIndicator.style.top = (first.offsetTop - 1) + 'px';
+        this._viewportIndicator.style.top = first.offsetTop - 1 + 'px';
         this._viewportIndicator.style.height =
           last.offsetTop - first.offsetTop + last.offsetHeight + 'px';
       }
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -792,6 +792,13 @@ export class WindowedList<
     const scrollbarElement = node.appendChild(document.createElement('div'));
     scrollbarElement.classList.add('jp-WindowedPanel-scrollbar');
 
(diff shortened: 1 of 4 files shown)


srdas:
@krassowski - This is really nice! As a parallel, I use the mini view in VSCode to traverse quickly to where my edits to the source file are, and I am not sure if this is a useful way to use the viewport in the notebook. If so, adding color to modified cells may be useful. It would definitely be nice to add this and get some feature parity with other IDEs.

krassowski:
Thank you for the suggestion! I implemented it in https://github.com/jupyterlab/jupyterlab/pull/16432.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1271,7 +1277,18 @@ export class WindowedList<
...
     if (newWindowIndex !== null) {
-      const [startIndex, stopIndex] = newWindowIndex;
+      const [startIndex, stopIndex, firstVisibleIndex, lastVisibleIndex] =
+        newWindowIndex;
+
+      if (this.scrollbar) {
+        const scrollbarItems = this._renderScrollbar();
+        const first = scrollbarItems[firstVisibleIndex];
+        const last = scrollbarItems[lastVisibleIndex];
+        this._viewportIndicator.style.top = first.offsetTop + 'px';
```
brichet, line 1287:
```suggestion
        this._viewportIndicator.style.top = (first.offsetTop - 1) + 'px';
```

packages/ui-components/style/windowedlist.css
```diff
@@ -38,19 +40,33 @@
...
+  position: relative;
+  z-index: 3;
 }
 
 .jp-WindowedPanel-scrollbar-content > .jp-WindowedPanel-scrollbar-item {
   border-bottom: 1px solid var(--jp-layout-color3);
+  background-color: rgb(from var(--jp-layout-color1) r g b / 50%);
   padding: 2px;
   text-align: left;
+  margin: 2px;
```
brichet, line 56:
```suggestion
  margin: 0 2px;
```

brichet (commented):
Thanks @krassowski, this looks great.

There is currently a gap between a selected item and the top separator, but not the bottom separator. And the viewport also seems to be a bit too low



With changes proposed below, it seems more regular:

brichet:
Sorry I split the review for a better readability.

For a long notebook, the virtual scrollbar does not fit in the page, and the scrollbar viewport disappear.



Do you think that the virtual scrollbar could be scrolled to have the scrollbar viewport always displayed ?

krassowski:
>Do you think the virtual scrollbar could be scrolled, to follow the scrollbar viewport ?

Yes, I had the same thought and completely agree that this could be useful, but since this is an additional behaviour change (as of today the virtual scrollbar does not scroll when you move the viewport outside of it) I would defer it to a separate PR. I opened https://github.com/jupyterlab/jupyterlab/issues/16518 to track this.

brichet (approved):
Thanks @krassowski, looks good to me.

PR #16398, opened 2024-05-30
Support collaborative stdin output

davidbrochart:
## References

In https://github.com/jupyterlab/jupyter-collaboration/pull/307 and [jupyter-server-nbmodel](https://github.com/datalayer/jupyter-server-nbmodel), @fcollonval and @echarles are working on adding support for input requests in server-side execution. While the proposed solution is an improvement over the [initial PR](https://github.com/jupyterlab/jupyter-collaboration/pull/279), it is not based on CRDTs and thus suffers from not being recoverable: if a user closes their browser window after executing a cell with an input request, and reopens it, the input widget is not displayed and the notebook is deadlocked. Also, the input widget is only shown to the user who executed the cell.
This PR goes all the way with a CRDT-based input widget, making it recoverable and collaborative, as shown below:

https://github.com/jupyterlab/jupyterlab/assets/4711805/b0951037-968a-42b9-829e-fee368f8e7bc

This currently works with Jupyverse (through [this PR](https://github.com/jupyter-server/jupyverse/pull/415)), but the changes here shouldn't have any effect when not running in server-side execution mode.
When implemented in [jupyter-server](https://github.com/jupyter-server/jupyter_server), this PR will be needed: https://github.com/jupyter-server/jupyter_ydoc/pull/233.

## Code changes

The cell shared model outputs are watched and if an output of type `stdin` is added, a CodeMirror input widget is shown. When the input is entered, the stdin output's `submitted` field is set to `true`.
The backend watches the stdin output, and when the `submitted` field is set to `true`, it replaces the `stdin` output with a `stream` output.

## User-facing changes

The user can use inputs in server-side execution. Inputs are collaborative and recoverable.

None.

### DIFF
--- a/packages/cells/package.json
+++ b/packages/cells/package.json
@@ -69,7 +69,8 @@
     "@lumino/signaling": "^2.1.2",
     "@lumino/virtualdom": "^2.0.1",
     "@lumino/widgets": "^2.3.2",
-    "react": "^18.2.0"
+    "react": "^18.2.0",
+    "yjs": "^13.5.40"
   },
   "devDependencies": {
     "@jupyterlab/testing": "^4.3.0-alpha.0",
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -3,9 +3,9 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
-import { Extension } from '@codemirror/state';
+import { Extension, StateCommand } from '@codemirror/state';
 
-import { EditorView } from '@codemirror/view';
+import { EditorView, KeyBinding, keymap } from '@codemirror/view';
 
 import { ElementExt } from '@lumino/domutils';
 
@@ -20,6 +20,10 @@ import { CodeEditor, CodeEditorWrapper } from '@jupyterlab/codeeditor';
 import { DirListing } from '@jupyterlab/filebrowser';
 
 import * as nbformat from '@jupyterlab/nbformat';
+
+import * as Y from 'yjs';
+
+import { ybinding } from '@jupyterlab/codemirror';
 
 import {
   IOutputPrompt,
@@ -39,7 +43,7 @@ import {
 
 import { Kernel, KernelMessage } from '@jupyterlab/services';
 
-import { IMapChange } from '@jupyter/ydoc';
+import { IMapChange, YCodeCell } from '@jupyter/ydoc';
 
 import { TableOfContentsUtils } from '@jupyterlab/toc';
 
@@ -182,6 +186,14 @@ const RENDER_TIMEOUT = 1000;
  * The mime type for a rich contents drag object.
  */
 const CONTENTS_MIME_RICH = 'application/x-jupyter-icontentsrich';
+
+const OUTPUT_AREA_ITEM_CLASS = 'jp-OutputArea-child';
+const OUTPUT_AREA_STDIN_ITEM_CLASS = 'jp-OutputArea-stdin-item';
+const OUTPUT_AREA_PROMPT_CLASS = 'jp-OutputArea-prompt';
+const OUTPUT_AREA_OUTPUT_CLASS = 'jp-OutputArea-output';
+const STDIN_CLASS = 'jp-Stdin';
+const STDIN_PROMPT_CLASS = 'jp-Stdin-prompt';
+const STDIN_INPUT_CLASS = 'jp-Stdin-input';
 
 /** ****************************************************************************
  * Cell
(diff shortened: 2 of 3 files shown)


fcollonval:
Thanks @davidbrochart for sharing this solution.

I have multiple comments at multiple levels.

- I'm against merging this PR right now because it requires lots of hypothesis that are not met by the standard installation: e.g. not compatible with Jupyter Server, jupyter-collaboration must be activated, requires document/kernel binding  
  It will be better to use the extensibility of JupyterLab to add this in an extension. This seems possible to implement using a widget extension like in the [jupyterlab-execute-time](https://github.com/deshaw/jupyterlab-execute-time/blob/c020582313868db92f1b0bdbae71aeb305b29fc1/src/index.ts#L68) extension.
- On the PR itself, the usefulness of an collaborative non-password input is debatable. But password should never be shared between clients like it is in this PR. Worst is that it is possible to retrieve the password without introspecting the websocket message but by using a simple copy-paste of the editor content (for reference copy is blocked on `<input type="password" />` element)
- This adds a [new type of output](https://github.com/jupyter-server/jupyter_ydoc/pull/233). This is not part of nbformat and we should avoid as much as possible to introduce discrepancy between the ydoc and nbformat. Therefore, I would strongly suggest opening the subject for discussion as JEP.
```
        {
            "output_type": "stdin",
            "submitted": bool,
            "password": bool
            "prompt": str,
            "value": Text
        }
```
- Finally and more fundamentally, is this the way we want to handle stdin in collaborative context? There is a real issue on how to inform all users about the need to provide an input to the kernel. But do we really need the candidate input to be collaborative? What about the issue with the password, leaking them in the document is very bad - even if it is removed when serializing the document on disk, they will be stored in the CRDT db.

Back to the basis, the need for the user to provide an input value is the consequence of a kernel being in a lock state waiting for an input (due to an execution request). That state may be reached independently of a document (although it is good to inform the user on the context leading to that state). Therefore a pure document based solution will only create more edge cases where the kernel state and the document are not properly synchronized.
With the work of [jupyter-server-nbmodel](https://github.com/datalayer/jupyter-server-nbmodel) (execution controlled by the backend) and in light of the discussion at the jupyter server meeting, a more robust path is to get the information of a pending input (and more generally of the state) from the backend using the kernel state machine proposal of @Zsailer (including it as part of the KernelManager). Then to resolve the issue of linking the state context with the document, we could ask the kernel (as metadata are sent with an execution request) or we could move forward with @krassowski's idea proposed in https://github.com/jupyter-server/jupyter_ydoc/pull/227

The cons of that proposal:
- Input won't be collaborative - each users will have the opportunity to enter an input but the input won't be shared
- Having a robust state machine outside the kernel will never be 100% reliable as multiple process can connect to a kernel _directly_. (this PR does not solve that case either).
The pros:
- We can handle connection to a kernel waiting for an input; without the need for RTC - immediate advantage for all users. This can happen if you refresh your browser while a Stdin prompt is displayed.
- We get a more robust kernel state
- We don't touch the document format

echarles:
Could we start with the user experience? If you were to ask me what I am expecting from such a feature, as a data analyst working on a collaborative sessions with teammates:

- I want to see the code changes the other users are typing in the cells.
- I want to see the exact status of the kernels: Available for executing code, Processing code from User_X, Waiting for input triggered by User_X...
- If the kernel is Processing code, I am interested to know which cell and exact code it is about. I am interested to see the outputs that execution is generating.
- If the kernel is Waiting for input, I am interested to know which cell and code has triggered that input. I don't want to see what the other users are typing for various reasons. First reason is that I have not requested the input, so I should not be able to provide any input in a free form. Second reason is privacy: this goes for the passwords, but also for any other free inputs. Notebooks can be used to collect any type of data, so I don't want to have a view on those data. Once the data is collected, and the notebook code is willing to display it, that is fine for me to see it.

This is just my view and I don't pretend nor think that the user experience should be like that, therefor I call for deeper thoughts with much more feedback to grasp what an agreed user experience should be.

Once we know what we have to offer, we could discuss how to implement that, impacting nformat or not. My understanding is that any change on the nbformat would go via a JEP.

davidbrochart:
I think there is a misunderstanding here. There is absolutely no change required in nbformat, hence no JEP.
As far as saying that only the user who requested the cell execution should see the input box, I think that it doesn't make a lot of sense in a collaborative setup. I would reformulate to: the collaborators execute a cell, thus they should all see the same result, be it an output or an input.
Giving a special role to the user who executes a cell because they are the first to execute is pretty arbitrary anyway, since another user could re-execute the cell later. A clearer model for that would be using permissions: a user could have read-only access to a document, and thus not be able to run cells at all or type into an input box.

echarles:
I am not with you on your previous comment. Let's wait on other users perspectives.

davidbrochart:
Happy to hear your thoughts meanwhile. For instance, how would you handle the case where a user who is supposed to submit in input, doesn't? Other users couldn't work on the notebook anymore, because it would be deadlocked.

echarles:
> Other users couldn't work on the notebook anymore, because it would be deadlocked.

That exact question/case/issue is also applicable to the long/everlasting running code. This is why I have written in https://github.com/jupyterlab/jupyterlab/pull/16398#issuecomment-2143728081 that I'd like to know which cell is runing what and/or asking input from whom. My experience come of dead(lock) kernel is coming from some bad acting code rather that from input request... The "state" of the kernel, as well who can do what on that state. There is work atm being done by @Zsailer and @fcollonval

krassowski:
> We can think of other ways to hide the password in the frontend than using [-webkit-text-security](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security), that was just an initial idea.

If the password is in the shared model, then it is not safe regardless of how the frontend obfuscates it.

> What do you think about hashing the password, so that it is never stored in clear form?

I don't think this helps much. Syncing just the number of characters? Maybe.

> But in the case of server-side execution, the kernel is owned by the server, so I don't see a scenario where the kernel state and the document are not properly synchronized

If ydoc is syncing via one websocket and the server requests are handled via an independent REST API there is no guarantee that things will be in sync. We could make use of some state hash and locks to fix this but not sure how well it would work.

krassowski:
The current input box offers many more functions (search in history, navigate between lines in history). It might be simpler to achieve feature parity if we can have the input box be a simple `input` element rather than CodeMirror instance; also CM6 has some milliseconds delay for initialization, and a quick consecutive input from users is an edge cases relevant to use of `input()` when debugging - if the input is not ready, this causes focus jumping and destroys UX.

How hard would be connecting an `input` element directly without using CM?

davidbrochart:
> How hard would be connecting an `input` element directly without using CM?

It wouldn't be hard, but the input wouldn't be collaborative.

fcollonval:
> I don't think this deserves a JEP. Jupyter shared models are not part of any standard, and we already store extra information that is not present in the notebook format (e.g. the state).

Let take a slightly futuristic point of view in which RTC is highly robust and there are tooling like timeline to navigate the document history. Deployed JupyterLab will start to gain of not serving the notebook from files but directly from database much like Google Docs and co. So the source of truth for the document will be the CRDT model. So this feature will add a non-standard output type. Hence we should work hard to not introduce discrepancy compare to nbformat.

> What do you think about hashing the password, so that it is never stored in clear form?

I second Mike on this.

> But in the case of server-side execution, the kernel is owned by the server, so I don't see a scenario where the kernel state and the document are not properly synchronized. Am I missing something?

Cases for not proper sync are to look for in degraded situation like:
A user with a pending input (no RTC), shutdown the application (server included). When it reopens it, the cell will display an input prompt although the kernel will be fresh and not waiting for any.

---

> It wouldn't be hard, but the input wouldn't be collaborative.

That is exactly one of the question we need to settle on; does the input itself should be collaborative.

> [!NOTE]
> I think the input can be collaborative. It is a question of synchronization of keyboard event with a Y.Text, isn't it?

davidbrochart:
It turns out that this can be implemented as a JupyterLab extension, thanks @fcollonval for pointing to [jupyterlab-execute-time](https://github.com/deshaw/jupyterlab-execute-time), it was very useful.
I created [jupyterlab-collaborative-input](https://github.com/davidbrochart/jupyterlab-collaborative-input).

PR #16409, opened 2024-05-31
Reactive toolbar computation, again...

brichet:
There is still an issue with the reactive toolbar, which can occur when an item is inserted during `_onResize()` execution.

https://github.com/jupyter/notebook/issues/7370#issuecomment-2128791005

This PR uses a static view of the items in the layout, to avoid some changes in the items list during computation.

A way to fix it is to make a copy of the existing items before starting computation, and use only that copy instead of the widget in the layout (that may change).

## References

Should fix https://github.com/jupyter/notebook/issues/7370.

## Code changes

Copy the list of widgets in the toolbar layout to avoid issue if the widgets list change during computation.

## User-facing changes

None

None

cc. @jtpio

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -640,14 +640,14 @@ export class ReactiveToolbar extends Toolbar<Widget> {
     openerWidth: number
   ) {
     const opener = this.popupOpener;
-    const layout = this.layout as ToolbarLayout;
-    const toIndex = layout.widgets.length - 1;
+    const widgets = [...(this.layout as ToolbarLayout).widgets];
+    const toIndex = widgets.length - 1;
 
     const widgetsToRemove = [];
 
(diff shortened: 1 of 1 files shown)


brichet:
An other way to fix it (and also https://github.com/jupyterlab/jupyterlab/pull/16335, and maybe future issues) could be to change [`insertItem()`](https://github.com/jupyterlab/jupyterlab/blob/21ffea9749637633b4a4d82b3f277b55f80afb3d/packages/ui-components/src/components/toolbar.tsx#L450) to an asynchronous function, to properly await any current computation before inserting the item.
But that would change the API.

jtpio:
Thanks @brichet!

> But that would change the API.

Right, if we can avoid this that would be great. Maybe we can go ahead with the approach in this PR and reconsider later.

jtpio (approved):
Thanks!

jtpio:

PR #16430, opened 2024-06-04
Fix some flaky ui tests 

brichet:
This PR may fix some flaky UI tests.

- replace some of the deprecated `Locator.type()` with `Locator.fill()`, to try to fix some tests writing in the cell input.
For example `test/jupyterlab/notebook-max-outputs.test.ts:37:5 › Don't limit cell outputs if input is requested` often fails due to indentation error in cell input.

image

In the screenshot above, the input should [not be indented](https://github.com/jupyterlab/jupyterlab/blob/228bd651ec924a2c5c8c1443bb5188c904e18ef3/galata/test/jupyterlab/notebook-max-outputs.test.ts#L47)

- set `'docmanager:open'` as `async`, according to VSCode warning "*This may be converted to an async function.ts(80006)*". 
  Not sure to understand why, but it seems to fix (locally) `test/jupyterlab/settings.test.ts:129:7 › change font-size › should Increase Content Font Size` and `test/jupyterlab/settings.test.ts:129:7 › change font-size › should Decrease Content Font Size`, which also often fails withe the following error:

image

## References

related to https://github.com/jupyterlab/jupyterlab/issues/14947

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/galata/test/jupyterlab/texteditor.test.ts
+++ b/galata/test/jupyterlab/texteditor.test.ts
@@ -57,7 +57,6 @@ test.describe('Text Editor Tests', () => {
 
     await page.locator(`[role="main"] >> text=${DEFAULT_NAME}`).waitFor();
 
-    await page.pause();
     await page
       .locator('.jp-FileEditorCodeWrapper .cm-content')
       .pressSequentially(
--- a/galata/test/jupyterlab/texteditor.test.ts
+++ b/galata/test/jupyterlab/texteditor.test.ts
@@ -31,9 +31,9 @@ test.describe('Text Editor Tests', () => {
 
     // Add two rulers
     await page.locator('#root').getByRole('button', { name: 'Add' }).click();
-    await page.locator('input[id="root_rulers_0"]').type('50');
+    await page.locator('input[id="root_rulers_0"]').fill('50');
     await page.locator('#root').getByRole('button', { name: 'Add' }).click();
-    await page.locator('input[id="root_rulers_1"]').type('75');
+    await page.locator('input[id="root_rulers_1"]').fill('75');
 
     await page.activity.activateTab(DEFAULT_NAME);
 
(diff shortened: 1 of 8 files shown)


galata/test/jupyterlab/texteditor.test.ts
```diff
@@ -57,10 +57,12 @@ test.describe('Text Editor Tests', () => {
 
     await page.locator(`[role="main"] >> text=${DEFAULT_NAME}`).waitFor();
 
-    await page.type(
-      '.cm-content',
-      'Not active\nActive line with >>selected text<<\nNot active'
-    );
+    await page.pause();
```
krassowski, line 60:
```suggestion
```

galata/test/jupyterlab/texteditor.test.ts
```diff
@@ -57,10 +57,12 @@ test.describe('Text Editor Tests', () => {
...
+    await page
+      .locator('.jp-FileEditorCodeWrapper .cm-content')
+      .pressSequentially(
+        'Not active\nActive line with >>selected text<<\nNot active'
+      );
```
krassowski, line 65: I would expect `fill` to work here. Does it not?

galata/test/jupyterlab/texteditor.test.ts
brichet: Good catch, thanks

galata/test/jupyterlab/texteditor.test.ts
brichet: It should, yes

brichet: I updated it

krassowski (approved):
Thank you so much for working on it @brichet!

jtpio:
Nice, thanks @brichet!

jtpio:

PR #16431, opened 2024-06-04
Define `ICodeCellModel.executionState`, deprecate `setPrompt()`

krassowski:
## References

- required for https://github.com/jupyterlab/jupyterlab/issues/16332
- discussed in https://github.com/jupyter-server/jupyter_ydoc/issues/169
- in future may be refactored to:
   - use `sharedModel` as source of truth if https://github.com/jupyter-server/jupyter_ydoc/pull/197 is merged, or
   - be computed based on `pending_requests` if https://github.com/jupyter-server/jupyter_ydoc/pull/227 is merged

## Code changes

- adds `ICodeCellModel.executionState: 'idle' | 'running'`. In future we may want to add more states such as `'errored'` or `'scheduled'`.
- deprecates the use of `CodeCell.setPrompt(value: string)` method which was circumventing the model

## User-facing changes

None

None

### DIFF
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -154,6 +154,11 @@ export interface ICodeCellModel extends ICellModel {
   executionCount: nbformat.ExecutionCount;
 
   /**
+   * The code cell's state.
+   */
+  executionState: 'running' | 'idle';
+
+  /**
    * The cell outputs.
    */
   readonly outputs: IOutputAreaModel;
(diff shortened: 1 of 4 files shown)


packages/cells/src/model.ts
```diff
@@ -153,6 +153,11 @@ export interface ICodeCellModel extends ICellModel {
    */
   executionCount: nbformat.ExecutionCount;
 
+  /**
+   * The code cell's state.
+   */
+  executionState: 'running' | 'idle';
```
fcollonval, line 159: Thanks a lot @krassowski for proposing this.

Should there be additional state:
- `pending-input` state
- `queued` state (sent for execution but not running yet)

krassowski: In the future I would love to have `queued` state. I am not sure about `pending-input` but maybe. However, I do not want to hold this PR on reaching a consensus on additional future values we might include; I think it is fair to say that we may want to include more states and hence it should not be a boolean but an enumeration of possible values.

Issue #3738, opened 2018-01-29
"minimap" of notebook structure for navigation


gfrogat:
it would be nice to have some sort of a "minimap" preview of the notebook structure, that can be used to navigate large notebooks or large cells.

this idea was already mentioned once in #3013 but it seems it wasn't discussed any further.

I think this feature could be very helpful for users, but it's still not completely obvious to me what to show when in the "minimap".
Here are some options that came to my mind:

### cell preview
preview the cells of the notebook, where the cells in the preview are scaled to match the size in the real notebook (so no not just showing a "pearl chain" where all cells are the same size).
In this view one could color-code the cells based on their type (e.g. code, markdown, ...) and cell tags (mentioned in issue #2277). It would be even more interesting to automatically tag and highlight cells with certain contents such as function definitions or module imports.

### code preview for very big cells
for very large cells one could switch to a standard code minimap that simply displays the structure of the code in this cell.

### code preview of all code cells
display a code minimap that contains the code of all notebook cells where different cell regions are visually hinted

### thumbnail preview of all figures / graphics
a slideshow of all graphics in the notebook that can be used for navigating to the respective part of the notebook

### table of contents preview
this functionality is related to #723. One could even use the "section information" to further structure the the other minimap types mentioned above.

I would be willing to work on this and would like to get some thoughts on the idea/concept.
I'll add some sketches later this week.

jasongrout:
(Also, just to set expectations, a performant minimap of code can be incredibly complicated to implement)

gfrogat:
yes definitely, I'm aware of that 🙂 

I would have started with the normal cell minimap anyway. I just wrote down all my ideas so there's some basis for discussion.

bollwyvl:
Hey folks, this is wonderful to see it come up again. I've wanted for as long as I can remember. 

Back In The Day, one of my first notebook extensions was [ipython-notebook-navigator](ipython-notebook-navigator)... 
![](https://github.com/bollwyvl/ipython-notebook-navigator/raw/master/screenshot.png)

wow, those install instructions are brutal! `custom.js` hacking! Feel free to grab anything you find there, and please don't judge 😊!

Since then, it has also show up here:
- https://github.com/jupyter/notebook/issues/594

[...]

Issue #16315, opened 2024-05-10
4.3 Release Plan


jtpio:
## Current Status

`4.2.0` was released a few days ago: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.0

This is an issue to track the features for the `4.3` release.

## Features & notable enhancements

- Standalone features
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16432
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16308
- File browser
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16646
  - [x] https://github.com/jupyterlab/jupyterlab/pull/10596
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16598
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16446
- Server-side execution/real time collaboration enhancements
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16498
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16651
- Improved kernel and server interactions
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16475
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16269
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16265
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16228
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16341
- Noticeable polish to the UX:
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16566
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16572
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16615
  - [x] https://github.com/jupyterlab/jupyterlab/pull/16637

To be done:
- [ ] https://github.com/jupyterlab/jupyterlab/issues/16528

## Timeline (provisional)

- Final 9th September
- RC 26th August
- Beta <s>5th</s>11th August

<details>
<summary>Postpotned RPs</summary>
- [ ] https://github.com/jupyterlab/jupyterlab/pull/15157
- [ ] https://github.com/jupyterlab/jupyterlab/pull/15155
</details>

fcollonval:
I would like to have those ones considered:
- https://github.com/jupyterlab/lumino/pull/715
- https://github.com/jupyterlab/lumino/pull/705

krassowski:
First beta version was released today:

[...]

PR #16432, opened 2024-06-04
Notebook minimap in the virtual scrollbar

krassowski:
## References

- closes https://github.com/jupyterlab/jupyterlab/issues/16332
- closes https://github.com/jupyterlab/jupyterlab/issues/3738

For now this PR includes two PRs on which it depends:
- https://github.com/jupyterlab/jupyterlab/pull/16392
- https://github.com/jupyterlab/jupyterlab/pull/16431

## Code changes

Creates a private (non-exported) `ScrollbarItem` class which implements reactive scrollbar item allowing for high-performance rendering of minimap even in very large notebooks.

## User-facing changes

The virtual scrollbar now exposes information on whether the cell:
- is running/scheduled (`[*]`), idle (`[ ]`) or was already run (e.g. `[1]`)
- was modified since it was last executed ("dirty") - orange background - thank you @srdas for the suggestion
- has output contains an error (traceback) - red background
- is markdown/code (`[X]` for code, nothing for markdown)
- roughly what text content does it contain

wip

running-and-modified

## Plan

Depending on feedback:

- [ ] make the display of source code configurable?
   - [ ] allow to configure maximum number of lines or height shown?
- [ ] make the entire thing configurable, allowing fallback to previous simpler implementation?

Ideas for future improvements:
- option for viewport to be always visible
- option to show headings as a line with bold?

None

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -1321,7 +1321,7 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
       this._notebook.selectionChanged.connect(this._updateSelection);
       if (this._model.type === 'code') {
         const model = this._model as ICodeCellModel;
-        model.outputs.changed.connect(this._updateOutput);
+        model.outputs.changed.connect(this._updatePrompt);
         model.stateChanged.connect(this._updateState);
       }
     }
@@ -1338,7 +1338,7 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
 
     this._updateActive();
     this._updateSelection();
-    this._updateOutput();
+    this._updatePrompt();
     this._updateDirty();
     return this._element;
   };
@@ -1372,7 +1372,7 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
     if (this._model.type === 'code') {
       const model = this._model as ICodeCellModel;
       if (model.outputs) {
-        model.outputs.changed.disconnect(this._updateOutput);
+        model.outputs.changed.disconnect(this._updatePrompt);
         model.stateChanged.disconnect(this._updateState);
       }
     }
@@ -1413,11 +1413,7 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
     }
   }
 
-  private _updateOutput = () => {
-    this._updatePrompt();
-  };
-
-  private _updatePrompt() {
+  private _updatePrompt = () => {
     if (this._model.type !== 'code') {
       return;
     }
@@ -1448,7 +1444,7 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
     if (this._element!.dataset.output !== state) {
       this._element!.dataset.output = state;
     }
-  }
+  };
 
   private _createElement() {
     const li = document.createElement('li');
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -85,6 +85,11 @@ const ACTIVE_CLASS = 'jp-mod-active';
 const SELECTED_CLASS = 'jp-mod-selected';
 
 /**
+ * The class name added to the cell when dirty.
+ */
+const DIRTY_CLASS = 'jp-mod-dirty';
+
+/**
  * The class name added to an active cell when there are other selected cells.
  */
 const OTHER_SELECTED_CLASS = 'jp-mod-multiSelected';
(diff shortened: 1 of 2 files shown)


ellisonbg:
Ohh, very cool! Some minor feedback:

* I think the rendering of the prompt/cell number using `[1]` in the minimap cells works well. However, we might want to try to display the prompt at a slightly smaller size. It is a bit hard to tell just from a screenshot, but it looks like the same size as the actual prompt. Maybe one or two stops smaller?
* I realize the value of putting the icon button right above the UX. However the kernel name and kernel status indicator have been in that position for a long time (since the beginning). Moving them to the right could confuse users who have grown used to their location (those items are all related to the kernel and its status). Could you try putting it to the right of the cell type selector as a variation? Another downside of its current location is that it would be the first icon to disappear (into the `...` button) when the toolbar gets narrow.

If it isn't too much trouble, could you generate some alternate screenshots to help us explore and weight the different visual options.

krassowski:
Thank you for leaving feedback @ellisonbg!

> I think the rendering of the prompt/cell number using [1] in the minimap cells works well. However, we might want to try to display the prompt at a slightly smaller size. It is a bit hard to tell just from a screenshot, but it looks like the same size as the actual prompt. Maybe one or two stops smaller?

I agree that it should be smaller than the actual prompt. There is of course a limit as to how small we can make it to keep it legible. Currently the prompt in the notebook and in the minimap are both 13px and in the the minimap they are 80% of that which works out to 10.4px. Visually they may seem more similar because the former are grey and the latter are black (left - minimap as of now, right - cell in a notebook)

image

> I realize the value of putting the icon button right above the UX. However the kernel name and kernel status indicator have been in that position for a long time (since the beginning). Moving them to the right could confuse users who have grown used to their location (those items are all related to the kernel and its status). Could you try putting it to the right of the cell type selector as a variation? Another downside of its current location is that it would be the first icon to disappear (into the ... button) when the toolbar gets narrow.

To clarify, this PR is **not** adding the hamburger icon button to the toolbar. For context, that button (and the initial virtual scrollbar implementation) was added 6 months ago in https://github.com/jupyterlab/jupyterlab/pull/15533 by @afshin after previous discussion in https://github.com/jupyterlab/jupyterlab/pull/15109 (and discussing this on multiple JupyterLab calls). Discussing it in a dedicated issue rather than here would give it more prominence and allow users who only monitor the issues by screening the titles to participate so I opened https://github.com/jupyterlab/jupyterlab/issues/16478 for it. Also, if we were to make this change here, this PR would grow to touch dozens of snapshot files and toolbar tests, so I would really prefer not to do that.

andrewfulton9:
I tested this and it works well for me (Ubuntu 20.04.6 LTS, firefox 114.0.2). Scrolling works well, execution indicator, ect. all work smoothly and as expected on the `examples/OutputExamples.ipynb` notebook.

brichet:
Thanks @krassowski for adding this feature, it's really nice.

In my opinion, we should make it configurable, with this display as default.

> * [ ]  make the display of source code configurable?
>   
>   * [ ]  allow to configure maximum number of lines or height shown?

We could have an option to fully render the cells, which would remove the `max-height` of `.jp-scrollbarItem-source` and the limitation to 10000 characters.

> * [ ] make the entire thing configurable, allowing fallback to previous simpler implementation?

In the case of a simple render (rectangle with cell index), would it be easy (and necessary) to fallback to a simple `HTMLElement`, as it was before https://github.com/jupyterlab/jupyterlab/pull/16392, to avoid any computation ?

packages/notebook/src/widget.ts
```diff
@@ -1350,12 +1369,101 @@ class ScrollbarItem implements WindowedList.IRenderer.IScrollbarItem {
...
+        this._element.classList.add(DIRTY_CLASS);
+      } else {
+        this._element.classList.remove(DIRTY_CLASS);
+      }
+    }
   }
 
+  private _updateOutput = () => {
+    this._updatePrompt();
+  };
```
brichet, line 1418: Can't we use an arrow function for `_updatePrompt()` instead of using this intermediate `_updateOutput` ?

krassowski:
> Is this expected that there is no new line in markdown cells ?

I guess you did not expect it, so probably not exactly.

The current behaviour corresponds to `white-space-collapse: collapse`. If we added `white-space: pre` as in code cells then it would not accurately represent long lines in markdown. So neither is great. However, we can use `white-space: pre-line` (a.k.a `white-space-collapse: preserve-breaks`) which seems like a good option to me.

Top - markdown code, bottom - rendered:

| default (current) | white-space: pre | preserve-breaks |
|--|--|--|
| image | image | image |

packages/notebook/src/widget.ts
krassowski: Done in [5a77107](https://github.com/jupyterlab/jupyterlab/pull/16432/commits/5a77107e46de8662ed7c94a175f0df7e8d873478)

krassowski:
> In my opinion, we should make it configurable, with this display as default.

I agree. My original intent was to add settings in this PR but I realised that we will probably want the minimap to be an extension point and therefore we should not be adding settings to the Notebook package because then it would be a breaking change to remove them (which would be needed to move then to a new plugin providing the default implementation so that extensions can provide their own settings schema), so let's settle on the extension API for minimap first and implement settings second. I opened https://github.com/jupyterlab/jupyterlab/issues/16528 to track this.

krassowski:
> In the case of a simple rendering (rectangle with cell index), would it be easy (and necessary) to fallback to a simple `HTMLElement`, as it was before https://github.com/jupyterlab/jupyterlab/pull/16392, to avoid any computation ?

This could be implemented as a separate plugin (disabled by default) which could be swapped by the user if they wanted the other implementation once we implement an extension point for the minimap (#16528); that said I am not sure how much value there would be in it.

brichet (approved):
Thanks for the changes and details @krassowski, this looks good.

Should we add https://github.com/jupyterlab/jupyterlab/issues/16528 in [4.3 release plan](https://github.com/jupyterlab/jupyterlab/issues/16315) ?

krassowski:
Thank you for the review!

> Should we add https://github.com/jupyterlab/jupyterlab/issues/16528 in https://github.com/jupyterlab/jupyterlab/issues/16315 ?

Done!

Issue #16371, opened 2024-05-24
JupyterLab 4. Table of Contents shows HTML tags in headers


FredDuj:
## Description

With JupyterLab 4.2.0, when using markdown with html tags, the Table of Content display the full markdown code, including the html tags, instead of removing them (or displaying the markdown in the ToC with the correct format). This is with Chrome Version 123.0.6312.123.

## Reproduce

For example:
`## <span style='background :darkviolet' ><span style='color:White'> Initialization  >`

![ToC](https://github.com/jupyterlab/jupyterlab/assets/113023746/48563a66-2d51-46cd-b14d-069d56241b99)

## Expected behavior

The ToC should display either only teh text of teh markdown (without) teh HTML tags or the formatted text (using HTML tags)

## Context

- Operating System and version: <Windows 10Enterprise>
- Browser and version: <Chrome Version 123.0.6312.123>
- JupyterLab version: <4.2.0>

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/17583

PR #16444, opened 2024-06-05
Fix TOC shows HTML tags in headers 

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/16371

### DIFF
--- a/packages/toc/src/utils/markdown.ts
+++ b/packages/toc/src/utils/markdown.ts
@@ -253,24 +253,13 @@ function parseHeading(line: string, nextLine?: string): IHeader | null {
   let match = line.match(/^([#]{1,6}) (.*)/);
   if (match) {
     return {
-      text: cleanTitle(match[2]),
+      text: removeHTMLTags(cleanTitle(match[2])),
       level: match[1].length,
       raw: line,
       skip: skipHeading.test(match[0])
     };
   }
-  // Case: Markdown heading (alternative style)
-  if (nextLine) {
-    match = nextLine.match(/^ {0,3}([=]{2,}|[-]{2,})\s*$/);
-    if (match) {
-      return {
-        text: cleanTitle(line),
-        level: match[1][0] === '=' ? 1 : 2,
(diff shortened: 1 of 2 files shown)


packages/toc/src/utils/markdown.ts
```diff
@@ -253,34 +253,36 @@ function parseHeading(line: string, nextLine?: string): IHeader | null {
...
   if (match) {
     return {
-      text: cleanTitle(match[2]),
+      text: removeHTMLTags(cleanTitle(match[2])),
       level: match[1].length,
       raw: line,
       skip: skipHeading.test(match[0])
     };
   }
+
```
itsmevichu, line 262: I have moved this condition to handle the below scenario:

<h1>Hello World</h1>
===

The <h1> tag should be addressed first and then the ===.

itsmevichu:
@krassowski Pease help me address the codeQL failures.

krassowski:
What if the `parseHeading` function followed the same principle as `getHeadingId`, this is rendering the heading using a proper markdown parser and sanitizer to a temporary HTML element (instead of using regular expressions)?

krassowski:
Also CC @fcollonval for opinion on my comment above.

krassowski:
Maybe we can reconfigure sanitizer to not drop whatever we need? I think we should use sanitizer as otherwise we can get dangerous execution.

JasonWeill:
#16716 is an issue related to a lack of superscript character support using `<sup>` tags. Stripping all HTML tags would improve the user experience somewhat, but there are still cases in which an HTML tag provides useful presentational improvement.

afonit:
Adding that yes, it would be nice to support the TOC rendering anything that can be rendered in the markdown cell.  Frequent cases I run into are coloring text, bolding and math.

Left is the TOC from jupyterlab, the right is the rendered cell in jupyterlab

image

Here is the code:
```
# Main Header
## Secondary **Header**
## <span style="color:orange">Other Header</span>
## Yet Another Header $x^2$
```

Issue #14125, opened 2023-03-04
Large empty space in file browser toolbar


krassowski:
## Description

![Screenshot from 2023-03-04 11-25-16](https://user-images.githubusercontent.com/5832902/222897269-d0104780-4770-40c0-b99d-764353552f7d.png)

## Reproduce

1. Open file browser
2. Resize the sidebar to dimensions as on screenshot

## Expected behavior

Space is utilised efficiently.

## Context

- JupyterLab version: 4.0.0a

JasonWeill:
On tip of `master`, I see the search box appear to the right of the "reload" button when the file browser is expanded to 473px wide:

![image](https://user-images.githubusercontent.com/93281816/223518123-6e9075de-fa0f-475b-a9f2-452cb5fe06ee.png)

At 455 px or narrower, I see it on a separate line.

JasonWeill:
@krassowski What is the expected behavior? Do you want the search box to get progressively smaller beyond its current min. width? Should it eventually collapse to an icon, and then when the icon is clicked, it expands to allow text input?

GabrielaVives:
When the icon is clicked, I would expect the whole panel to expand to allow space for the text box to reach its min. width. Was that what you were thinking about @JasonWeill ?

JasonWeill:
@GabrielaVives I think that expanding the panel may be a visually jarring change. Instead, what about reducing the search/filter box to a very small size:

![image](https://user-images.githubusercontent.com/93281816/227366376-053d2580-1a47-4f4d-b559-b21cc4df833d.png)

… then, when the user clicks on it, expand it and make the other buttons invisible, and add a collapser that makes the other buttons visible again:

![image](https://user-images.githubusercontent.com/93281816/227367285-0fe75bb6-2258-417d-aa27-0d4dcf60bde0.png)

krassowski:
This seems like a good idea @JasonWeill but with a minor adjustment: because the filter stays active, collapsing it back would introduce hidden state and require us to add an indicator. So instead of ">" collapse button on the left I would suggest `x` on the right which would clear the filter and collapse the filter field.

JasonWeill:
We may need both, if collapsing the filter keeps it active. (We would then have to modify the collapsed filter control to indicate that a filter is active.

amzon-ex:
Just a minor suggestion: you could also check if the search box is empty and is focused. If empty and not in focus > collapse; in all other cases, keep it expanded

[...]

PR #16446, opened 2024-06-06
Filebrowser filter below breadcrumbs

JasonWeill:
## References

Partial fix for #14125. 

## Code changes

Takes the filename searcher out of the toolbar in the file browser, moves it below the breadcrumbs. The new "filter toolbar" is _not_ yet driven by a toolbar registry, and is not customizable.

## User-facing changes

Takes the filename searcher out of the toolbar in the file browser, moves it below the breadcrumbs, behind a "filter" button. Makes the buttons slightly narrower so that they all fit in the left sidebar at default width.

Before:

Screenshot 2024-06-18 at 4 39 54 PM
<br />
Screenshot 2024-06-18 at 4 40 05 PM

After:

Screenshot 2024-06-18 at 4 42 16 PM
<br />
Screenshot 2024-06-18 at 4 42 57 PM

<br />

Screenshot 2024-06-18 at 4 24 37 PM

The file searcher is no longer in the top toolbar.


krassowski:
This is a really nice idea! It would be great if the filter box could be hidden when not in use; I am not sure how this could work with this proposal. With the design that you proposed earlier in https://github.com/jupyterlab/jupyterlab/issues/14125#issuecomment-1481936819 it would seems easier to me to implement "show only when needed" behaviour as the user could just click on a filter icon (maybe it should not be a magnifying glass but a funnel btw).

> My bigger pet peeve with this filter box is that it was meant to solve an issue that when you type to select to a file there is no feedback of the current state, but because of a misunderstanding the PR aiming to fix that instead introduced a filter box which has a different behaviour because it does not implement a proper search nor does it work for quickly finding a file to open because then each time you need to clear the filter box (differently from "type to select" behaviour which clears on a timeout). And it takes up a lot of precious real estate in the file browser. I do not know of IDEs or file browsers which have such a filter box and personally would like to disable it altogether.

andrii-i:
> This is a really nice idea! It would be great if the filter box could be hidden when not in use; I am not sure how this could work with this proposal. With the design that you proposed earlier in https://github.com/jupyterlab/jupyterlab/issues/14125#issuecomment-1481936819 it would seems easier to me to implement "show only when needed" behaviour as the user could just click on a filter icon (maybe it should not be a magnifying glass but a funnel btw).

Here's how macos does it. I think it's a good reference for usability:
* Off state (note absence of the border around the icon):
Screenshot 2024-06-06 at 7 13 25 AM


* Focused and active state when nothing is typed (note "close search" button on the left ">>")
Screenshot 2024-06-06 at 7 14 58 AM


* When something is typed into the input field (note both "close search" button on the left ">>" and clear input button on the right "X")
Screenshot 2024-06-06 at 7 15 55 AM

JasonWeill:
This change remains a WIP, primarily because it doesn't clear the filter when the user clicks a "pressed" filter button to hide the control. I'll be out of office next week, June 10–14.

ellisonbg:
I like this exploration :-) Maybe more space around the element to match the spacing of other items in the file explorer?

packages/filebrowser/src/browser.ts
```diff
@@ -68,6 +85,20 @@ export class FileBrowser extends SidePanel {
...
+      new ToolbarButton({
+        // Make icon vary based on state of filter?
+        icon: filterIcon,
+        onClick: () => {
+          this.toggleFileFilter();
+        },
+        tooltip: this._trans.__('Filter files by name'),
+        pressed: this._showFileFilter
+      })
+    );
```
krassowski, line 100: By hard-coding the button in the toolbar, this line prevents user- and administrator-configuration of the layout. Instead it would be great if a command could be used. Please note that if you use a command, the state returned by `isEnabled`, `isVisible` and `isToggled` will be automatically reflected and it will also help extension authors in case if they want to toggle it programatically or via a shortcut. Of note, all other buttons in this toolbar are already added via schema see:

https://github.com/jupyterlab/jupyterlab/blob/ae123cb82015540804964f76ba4e56f21799ea27/packages/filebrowser-extension/schema/widget.json#L4-L15

https://github.com/jupyterlab/jupyterlab/blob/ae123cb82015540804964f76ba4e56f21799ea27/packages/launcher-extension/schema/plugin.json#L24-L32

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +236,51 @@ export class FileBrowser extends SidePanel {
...
+            }
+          });
+        }
+      });
+
+      // Return a filter that doesn't exclude anything.
+      this.model.setFilter(value => {
+        return {};
+      });
+      this.model.refresh();
```
krassowski, line 273: Missing promise handling:

```suggestion
      this.model.refresh().catch(console.warn);
```

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +236,51 @@ export class FileBrowser extends SidePanel {
...
+    }
+    this._showFileFilter = value;
+
+    // Update widget visibility
+    const folderToolbarNode = this.mainPanel.node.getElementsByClassName(
+      FOLDER_TOOLBAR_CLASS
+    )[0] as HTMLDivElement;
+    folderToolbarNode.style['display'] = this.showFileFilter ? 'block' : 'none';
```
krassowski, line 281: Can you operate on `this.folderToolbar.node` instead? This should reduce the complexity and touching the DOM directly which is more error prone.

packages/filebrowser/src/browser.ts
```diff
@@ -376,6 +477,18 @@ export class FileBrowser extends SidePanel {
...
+  toggleFileFilter(): void {
+    const oldValue = this.showFileFilter;
+    this.showFileFilter = !oldValue;
+    // Toggle button state for the filter button
+    (this.toolbar.layout as PanelLayout).widgets.forEach(widget => {
+      const button = widget as ToolbarButton;
+      if (button.node.getAttribute('data-jp-item-name') === 'toggleFilter') {
+        button.pressed = !oldValue;
+      }
+    });
```
krassowski, line 489: Instead of directly modifying the DOM, can you use the command approach discussed earlier? I would suggest using `isToggled` with custom styling if needed. It adds `lm-mod-toggled` as implemented here:

https://github.com/jupyterlab/jupyterlab/blob/ae123cb82015540804964f76ba4e56f21799ea27/packages/ui-components/src/components/toolbar.tsx#L1076-L1081

packages/filebrowser/src/browser.ts
```diff
@@ -76,6 +107,30 @@ export class FileBrowser extends SidePanel {
...
+        });
+      },
+      useFuzzyFilter: true,
+      placeholder: this._trans.__('Filter files by name'),
+      forceRefresh: true,
+      showIcon: false
+    });
+    searcher.addClass(FILTERBOX_CLASS);
+
+    this.folderToolbar = new Toolbar();
```
krassowski, line 127: Why is it named `folderToolbar`? Do we need it to be a toolbar, or can we just add this widget directly?

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +236,51 @@ export class FileBrowser extends SidePanel {
     }
   }
 
+  /**
+   * Whether to show a text box to filter files by name.
+   */
+  get showFileFilter(): boolean {
```
krassowski, line 242: Note for the next revision: we should probably think about the API surface; at first glance it appears that having both `showFileFilter` and `toggleFileFilter` as public may not be necessary. I think that once you implement the suggestion of using command with `isToggled` for reflecting the pressed status the need for `toggleFileFilter` will disappear.

krassowski (changes requested):
Thank you for working on it @JasonWeill!

packages/filebrowser/src/browser.ts
krassowski: Maybe `isToggled` in context of a button should in fact cause `pressed` state to be changed? Let me know your thoughts, we can always modify `packages/ui-components/src/components/toolbar.tsx` if needed.

jtpio:
> The new "folder toolbar" is not yet driven by a toolbar registry, and is not customizable.

We should check how this would affect Notebook 7, which uses the settings to show or hide the file name searcher.

packages/filebrowser/src/browser.ts
JasonWeill: As @jtpio suggests below, this could be made into a toolbar of its own, with a toolbar registry, although it isn't yet. Should it be?

JasonWeill:
> We should check how this would affect Notebook 7, which uses the settings to show or hide the file name searcher.

@jtpio I tried the changes in this branch with Notebook 7, and I didn't see any difference. I see filter options in the settings editor, but none of them cause the file browser to show a filter box, with or without my changes from this PR applied.

packages/filebrowser/src/browser.ts
krassowski: I guess it could be a toolbar if we want to add other buttons to configure the behaviour of the filter in the future. My first point was more about the "folder" part - it is not obvious why it is called "folder"

JasonWeill: I was trying to convey that it applies in the context of a folder (directory) and that it appears below the folder breadcrumbs in the UI. Open to other suggestions.

krassowski:
Test failures are relevant (`js-filebrowser` group)

JasonWeill:
please update snapshots

JasonWeill:
@krassowski Looks like all the test failures were related to screen shots; I just updated the screen shots for test and documentation.

krassowski:
No, the CI just did not run because the bot does not have rights to trigger the CI run. You need to close and reopen the PR. The CI failure I highlighted is unrelated to screenshots.

galata/test/jupyterlab/sidebars.test.ts-snapshots/opened-sidebar-jp-running-sessions-jupyterlab-linux.png
krassowski, line 1: Flaky, needs reverting

galata/test/documentation/general.test.ts-snapshots/command-palette-documentation-linux.png
krassowski, line 1: Font change, might be flaky and might need reverting

galata/test/documentation/general.test.ts-snapshots/running-layout-documentation-linux.png
krassowski, line 1: Flaky needs reverting

galata/test/documentation/internationalization.test.ts-snapshots/language-change-documentation-linux.png
krassowski, line 1: Might need a revert

krassowski (changes requested):
These tests need fixing (updating snapshots for examples and fixing unit tests):

Screenshot from 2024-06-27 20-22-52
Screenshot from 2024-06-27 20-22-37

Additionally some snapshots look like need reverting which will probably help with the failing visual regression tests

JasonWeill:
Got it — sorry for the confusion! It looks like the open dialog is using a filter that, for example, should constrain a file listing to only `*.workspace` files:

Screenshot 2024-06-27 at 4 43 07 PM

… but with this code change applied, the constraint is no longer applied, and all files are displayed:

Screenshot 2024-06-27 at 4 40 46 PM

JasonWeill:
The creation of a search box, not just a search box factory, with this change, seems to be breaking the open dialog with a filter already applied. The newly-created search box, even though it isn't visible in the "import workspaces" dialog, is causing the import file list not to be filtered at all, presumably because the search box is empty.

JasonWeill:
Changing the `forceRefresh` parameter in the search box to be `false` continues to allow the search box to filter files, but it also reinstates the expected functionality of the import dialog's filter.

packages/filebrowser/src/browser.ts
krassowski: While it does filter a single folder by default, with https://github.com/jupyterlab-contrib/jupyterlab-unfold it filters all (expanded) folders. It depends what should be in this toolbar. Since right now it only contains a filter, I would imagine `filterToolbar` to be a fitting name.

Thinking into the future I was imaging that it could also have buttons to toggle whether to filter in a single folder vs all folders, maybe whether to show hidden files, or whether to hide files in `.gitignore`. Maybe buttons are not a best solution but a dropdown with a set of options?

JasonWeill:
@krassowski @jtpio Could you please take another look at this when you get the chance? The js-filebrowser tests are now passing, and I don't think that the current pipeline failures are caused by this change.

galata/test/documentation/internationalization.test.ts-snapshots/language-settings-documentation-linux.png
krassowski, line 1: It looks like there is an issue with the Chinese fonts here.

krassowski (commented):
All tests should pass. I restarted the two failing jobs and the MacOS one passed but the visual regression test is still failing with:

> Notice:   1 failed
>    [documentation] › test/documentation/internationalization.test.ts:14:7 › Internationalization › Menu 

I think we have an issue with different runners or different fonts being used in snapshot generation vs testing runs. For now you could download the snapshot generated in the test run and commit it manually.

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +220,49 @@ export class FileBrowser extends SidePanel {
...
+          widget.node.getAttribute('data-jp-item-name') === 'fileNameSearcher'
+        ) {
+          // Find the search element's "clear" button and click it
+          widget.node.children[0].shadowRoot?.childNodes.forEach(childNode => {
+            if (childNode.nodeType === Node.ELEMENT_NODE) {
+              const clearButtonElements = (
+                childNode as HTMLElement
+              ).getElementsByClassName('clear-button');
+              if (clearButtonElements.length > 0) {
+                (clearButtonElements[0] as HTMLButtonElement).click();
```
krassowski, line 246: This looks pretty involved. @fcollonval is there a simpler way of cleaning the input?

galata/test/documentation/internationalization.test.ts-snapshots/language-settings-documentation-linux.png
krassowski: I pushed da04bd5c3abb88ffbfd9e54b52a42252876c7cf5

packages/filebrowser/src/browser.ts
```diff
@@ -6,8 +6,13 @@ import { PathExt } from '@jupyterlab/coreutils';
...
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
-import { SidePanel } from '@jupyterlab/ui-components';
-import { Panel } from '@lumino/widgets';
+import {
+  FilenameSearcher,
+  IScore,
+  SidePanel,
+  Toolbar
+} from '@jupyterlab/ui-components';
+import { Panel, PanelLayout } from '@lumino/widgets';
```
fcollonval, line 15:
```suggestion
import { Panel } from '@lumino/widgets';
import { createRef } from 'react';
```

packages/filebrowser/src/browser.ts
```diff
@@ -76,6 +91,30 @@ export class FileBrowser extends SidePanel {
...
+        query?: string
+      ) => {
+        this.model.setFilter(value => {
+          return filterFn(value.name.toLowerCase());
+        });
+      },
+      useFuzzyFilter: true,
+      placeholder: this._trans.__('Filter files by name'),
+      forceRefresh: false,
+      showIcon: false
```
fcollonval, line 107:
```suggestion
      showIcon: false,
      inputRef: this._fileFilterRef
```

packages/filebrowser/src/browser.ts
```diff
@@ -76,6 +91,30 @@ export class FileBrowser extends SidePanel {
...
+    });
+    searcher.addClass(FILTERBOX_CLASS);
+
+    this.filterToolbar = new Toolbar();
+    this.filterToolbar.addClass(FILTER_TOOLBAR_CLASS);
+    this.filterToolbar.addItem('fileNameSearcher', searcher);
+    this.filterToolbar.node.style['display'] = this.showFileFilter
+      ? 'block'
+      : 'none';
```
fcollonval, line 116:
```suggestion
    this.filterToolbar.setHidden(!this.showFileFilter);
```

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +220,49 @@ export class FileBrowser extends SidePanel {
...
+              const clearButtonElements = (
+                childNode as HTMLElement
+              ).getElementsByClassName('clear-button');
+              if (clearButtonElements.length > 0) {
+                (clearButtonElements[0] as HTMLButtonElement).click();
+              }
+            }
+          });
+        }
+      });
```
fcollonval, line 251:
```suggestion
      if (this._fileFilterRef.current) {
        this._fileFilterRef.current.value = '';
      }
```

packages/filebrowser/src/browser.ts
```diff
@@ -180,6 +220,49 @@ export class FileBrowser extends SidePanel {
...
+      this.model.setFilter(value => {
+        return {};
+      });
+      this.model.refresh().catch(console.warn);
+    }
+    this._showFileFilter = value;
+
+    // Update widget visibility
+    const filterToolbarNode = this.filterToolbar.node;
+    filterToolbarNode.style['display'] = this.showFileFilter ? 'block' : 'none';
```
fcollonval, line 263:
```suggestion
    this.filterToolbar.setHidden(!this.showFileFilter);
    if (this.showFileFilter) {
      this._fileFilterRef.current?.focus();
    }
```

packages/filebrowser/src/browser.ts
```diff
@@ -426,18 +509,20 @@ export class FileBrowser extends SidePanel {
...
 
+  protected filterToolbar: Toolbar;
   protected listing: DirListing;
   protected crumbs: BreadCrumbs;
   protected mainPanel: Panel;
 
-  private _manager: IDocumentManager;
   private _directoryPending: Promise<Contents.IModel> | null = null;
   private _filePending: Promise<Contents.IModel> | null = null;
+  private _manager: IDocumentManager;
```
fcollonval, line 519:
```suggestion
  private _fileFilterRef = createRef<HTMLInputElement>();
  private _manager: IDocumentManager;
```

packages/ui-components/src/components/search.tsx
```diff
@@ -185,6 +190,9 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
     );
   };
 
+  // Show the icon by default, or if the caller specifically requests it
+  const showSearchIcon = props.showIcon === undefined || props.showIcon;
```
fcollonval, line 194:
```suggestion
  const showSearchIcon = props.showIcon ?? true;
```

packages/ui-components/src/components/search.tsx
```diff
@@ -211,6 +219,7 @@ export const FilenameSearcher = (props: IFilterBoxProps): ReactWidget => {
       placeholder={props.placeholder}
       forceRefresh={props.forceRefresh}
       caseSensitive={props.caseSensitive}
+      showIcon={props.showIcon}
```
fcollonval, line 222: This should be simplified with as we are missing passing some props:

```ts
    return ReactWidget.create(<FilterBox {...props} />);
```

fcollonval (changes requested):
All my suggestions are mainly to answer Mike's question.

The search element behave like an input[type="search"] element. So the simplest thing to do is to set the `value`.

The suggestion use a ref to the search element. Therefore I use that to improve the code to focus on the search element when it is displayed from a click on the filter button.

The suggestion also use `Widget` API to deal with visibility rather than using the CSS.

> For info, a better behavior would be to drop the DOM nodes altogether rather than hiding things. But this is unfortunately a good practice we don't use much due to widgets technology.

krassowski:
FYI for the future, if you go to the "Files changed" tab you can commit multiple suggestion at once.

krassowski (approved):
Thank you @JasonWeill and thanks @fcollonval for suggestions!

fcollonval (approved):

Issue #15629, opened 2024-01-12
Default file browser creates new files and folders in a custom drive if installed


jtpio:
## Description

When a custom drive is installed, the default file browser might try to create files and directories in the installed drive, instead of the default file browser.

## Reproduce

- Install the latest JupyterLab version: `pip install jupyterlab==4.0.10`
- Install a custom drive, for example [jupyterlab-github](https://github.com/jupyterlab/jupyterlab-github/): `pip install jupyterlab-github`
- Click on the GitHub drive input area
- Go back to the default file browser, click on the icon to create a new directory

[jupyterlab-drive-focus-issue.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/0df44b69-0d4b-45d5-b138-3ba5aeb87ce1)

## Expected behavior

After switching drives, it should be possible to create a new directory in the default filebrowser.

## Context

- Operating System and version: Linux Ubuntu 23.04
- Browser and version: Firefox 118 & Chromium 120
- JupyterLab version: 4.0.10

This was initially noticed in https://github.com/jupyterlite/jupyterlite/pull/1263 and https://github.com/jupyter/notebook./issues/7210.

Thanks @brichet for reproducing the issue in JupyterLab 4 as well.

jtpio:
Create a new folder uses the following command, which check for the `currentWidget` in the file browser tracker:

https://github.com/jupyterlab/jupyterlab/blob/07b93390cf74fbf14d3a1dac3f662fdb4664afed/packages/filebrowser-extension/src/index.ts#L1143-L1147

jtpio:
Digging a bit more, this is also an issue in JupyterLab 3:

[jupyterlab-3-focus-issue.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/e6c4e2d2-71a7-4385-82eb-4c42904942df)

But it's still odd this started to surface by default in Notebook 7, with the recent pre-releases based on JupyterLab 4.1: https://github.com/jupyter/notebook/issues/7210. In Notebook 7 this is an issue only in `7.1.0a0` and higher, not in `7.0.x`.

jtpio:
This is also reproducible with another drive: https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access

[jupyterlab-filesystem-access-issue.webm](https://github.com/jupyterlab/jupyterlab/assets/591645/ccc36861-2ed8-4e67-b905-997adc9a8f90)

[...]

PR #16452, opened 2024-06-07
Handles a focus on the toolbar button on mousedown

brichet:
This PR fixes https://github.com/jupyterlab/jupyterlab/issues/15629, but can bring some unexpected changes.

The issue mentioned above comes from https://github.com/jupyterlab/jupyterlab/pull/12281, when the filebrowser panel became a `SidePanel` and its toolbar customizable.

The command to create a new directory get the current browser widget using the tracker. The tracker is updated when an `onClick` event is triggered on an element of the browser.
But this tracker is not yet updated when clicking on a button because the command is triggered [`onMousedown`](https://github.com/jupyterlab/jupyterlab/blob/963cbd6e76fd3a7b4be640662405e62a68268376/packages/ui-components/src/components/toolbar.tsx#L817).

To avoid this behavior, this PR change the focus to a toolbar element `onmousedown`, in `capture` mode to ensure it is triggered before the command.

Not sure about the consequences of this change yet.

## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15629

## Code changes

Focus the toolbar element on mouse down.

## User-facing changes

None

None

cc. @jtpio

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -312,8 +312,8 @@ export class Toolbar<T extends Widget = Widget> extends Widget {
    */
   handleEvent(event: Event): void {
     switch (event.type) {
-      case 'click':
-        this.handleClick(event);
+      case 'mousedown':
+        this.handleMousedown(event);
         break;
       default:
         break;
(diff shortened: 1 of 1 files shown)


fcollonval:
An alternative approach would be to test if switching to `actualOnClick` would work to solve this:

https://github.com/jupyterlab/jupyterlab/blob/6143d29918d0d3686646e22716293f288c659667/packages/ui-components/src/components/toolbar.tsx#L805

brichet:
> An alternative approach would be to test if switching to `actualOnClick` would work to solve this:

Thanks @fcollonval, this seems to work too.

Do you think that we should remove this property ? Or at least set it to `true` by default to keep the API ?

Otherwise we could add an option in [settingregistry.IToolbarItem](https://github.com/jupyterlab/jupyterlab/blob/6143d29918d0d3686646e22716293f288c659667/packages/settingregistry/src/tokens.ts#L726), but that adds some complexity that may be useless...

brichet:
I opened https://github.com/jupyterlab/jupyterlab/pull/16470 as an alternative, according to the suggestion of @fcollonval.

brichet:
Closed in favor of https://github.com/jupyterlab/jupyterlab/pull/16470

Issue #15902, opened 2024-03-03
ToC heading error in markdown comment


choies1:
## Description

Jupyter labs ToC recognizes the heading in the comment of markdown. ToC2 in the old Jupyter notebook didn't recognize the heading (#abc) inside markdown comments, which is inconvenient.

## Reproduce

The following example shows an error in Jupyterlab's ToC when markdown has a heading inside a comment.

[Firefox 123.0(64bit)]

![image](https://github.com/jupyterlab/jupyterlab/assets/10408785/ec5f892b-2893-4daa-948d-37e9e8140127)

## Expected behavior

![image](https://github.com/jupyterlab/jupyterlab/assets/10408785/50e47cca-98cc-4129-b8e2-ee8760f51dc9)

## Context

- Operating System and version: Windows 11
- Browser and version: Firefox 123.0(64bit), Chrome 122.0.6261.95 (64bit)
- JupyterLab version: 
 ```
jupyter --version
Selected Jupyter core packages...
IPython          : 8.15.0
ipykernel        : 6.25.0
ipywidgets       : 8.0.4
jupyter_client   : 7.4.9
jupyter_core     : 5.3.0
jupyter_server   : 2.10.0
jupyterlab       : 4.1.2
nbclient         : 0.5.13
nbconvert        : 6.5.4
nbformat         : 5.9.2
notebook         : 7.1.0
qtconsole        : 5.4.2
traitlets        : 5.7.1
```

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/17583

PR #16456, opened 2024-06-08
Fix markdown comment rendered as TOC Heading

itsmevichu:
Fix for - https://github.com/jupyterlab/jupyterlab/issues/15902

### DIFF
--- a/packages/toc/src/utils/markdown.ts
+++ b/packages/toc/src/utils/markdown.ts
@@ -74,9 +74,6 @@ export function getHeadings(text: string): IMarkdownHeading[] {
 
   // Iterate over the lines to get the header level and text for each line:
   const headings = new Array<IMarkdownHeading>();
-  let isCodeBlock;
-  let openingFence = 0;
-  let fenceType;
   let lineIdx = 0;
 
   // Don't check for Markdown headings if in a YAML frontmatter block.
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks @itsmevichu 

I'm more and more convinced that we should stop increasing our imperfect parsing rules to use a more robust approach based on the tokenizer of the markdown renderer.

For reference, the tokens can be obtained using marked js lexer: https://marked.js.org/using_pro#lexer

Now about the performance, we have two approaches, the renderer caches the tokens to be consumed by the Table of content. Or we parse again to only get the tokens in the table of contents.

The advantage of the first approach is a perfect fit between the markdown renderer and the table of content (but not the text editor highlighter that use CodeMirror parser). The disadvantage is that we will increase the API surface of the markdown parser with a tokenizer API that is certainly not standardized across markdown parsers.

@itsmevichu @krassowski thoughts on that idea?

fcollonval:
> Of not we are more and more parsing Markdown again in the table of contents but only what we assumed to be headings.

krassowski:
I think we could have a reasonable performance without increasing the API surface with the following approach:
- populate ToC asynchronously cell by cell
- for markdown cells render them to HTML in full; if they are rendered in notebook maybe just reuse the rendered content, if not do that async or in a worker
- parse headings out of generated HTML
- do not parse the markdown cell again until its content or type changed

itsmevichu:
> I think we could have a reasonable performance without increasing the API surface with the following approach:
> 
> * populate ToC asynchronously cell by cell
> * for markdown cells render them to HTML in full; if they are rendered in notebook maybe just reuse the rendered content, if not do that async or in a worker
> * parse headings out of generated HTML
> * do not parse the markdown cell again until its content or type changed

I completely agree with your approach; it should ensure reasonable performance without expanding the API surface.

PR #16465, opened 2024-06-09
Add a notice for Windows users to activate symbolic links in contributing section

Darshan808:
## Issue

When setting up the JupyterLab codebase on a Windows machine, I encountered a long list of errors after running pip install -e ".[dev,test]". The documentation did not provide a solution for this issue. After consulting with a contributor, I discovered that the errors were due to the need for symbolic links to be activated on Windows 10 or above for Python version 3.8 or higher by enabling ‘Developer Mode’.

This important detail was mentioned in the extension section of the documentation but not in the contributing section. As a result, new contributors using Windows may run into this issue without knowing how to resolve it.

## Solution

To help other contributors avoid this issue, I have added a notice to the contributing section of the documentation, instructing Windows users to activate ‘Developer Mode’ to enable symbolic links ( as mentioned in the extension section of the documentation ). This should provide a smoother setup experience for Windows users and prevent potential setup errors.

## Benefits:

1. Helps new contributors using Windows avoid setup issues.
2. Provides clear and necessary instructions for enabling symbolic links on Windows at the Project setup section itself.
2. Improves the overall documentation and setup process for the JupyterLab project.

## User-facing changes

The docs of Project Installation for at Contributing section has been updated

Screenshot 2024-06-09 234242

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -431,16 +431,6 @@ includes a ``Vagrantfile``, the bootstrap files and additional documentation.
 Installing JupyterLab
 ---------------------
 
-Important for Windows users
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. important::
-
-   On Windows, symbolic links need to be activated on Windows 10 or above for Python version 3.8 or higher
-   by activating the 'Developer Mode'. That may not be allowed by your administrators.
-   See `Activate Developer Mode on Windows <https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development>`__
-   for instructions.
-
 Fork the JupyterLab
 `repository <https://github.com/jupyterlab/jupyterlab>`__.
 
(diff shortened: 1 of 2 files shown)


docs/source/developer/contributing.rst
```diff
@@ -431,6 +431,16 @@ includes a ``Vagrantfile``, the bootstrap files and additional documentation.
 Installing JupyterLab
 ---------------------
 
+Important for Windows users
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. important::
+
+   On Windows, symbolic links need to be activated on Windows 10 or above for Python version 3.8 or higher
```
krassowski, line 439: Thanks for adding this section! Since it duplicates the existing section in extension tutorial:

https://github.com/jupyterlab/jupyterlab/blob/9746d1599ec2056f9a3c3f928214623a00e4e550/docs/source/extension/extension_tutorial.rst?plain=1#L206

I am slightly concerned that with time someone will edit this or that notice and they will diverge. Can you add a comment here and there to remind anyone who might be editing this section to also update the other instance?

See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#comments for comment syntax in RST

docs/source/developer/contributing.rst
```diff
@@ -431,6 +431,16 @@ includes a ``Vagrantfile``, the bootstrap files and additional documentation.
 Installing JupyterLab
 ---------------------
 
+Important for Windows users
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
krassowski, line 435: With this heading added in this place the outline looks a bit odd:

image

I mean that a user does not see the section with actual install instructions. What do you think about:
1) removing the heading
2) moving the `.. important::` box down to line 467 (previously 457)
   https://github.com/jupyterlab/jupyterlab/blob/9746d1599ec2056f9a3c3f928214623a00e4e550/docs/source/developer/contributing.rst#L457
3) making the "Notes" a section so that it is shown in the outline? We could name it "Frequent issues"  or "Troubleshooting installation".

Darshan808: I've implemented the suggested changes and committed them. Could you please take a moment to review and verify the modifications?

krassowski (approved):
Looks good to me, thank you @Darshan808!

Issue #16466, opened 2024-06-10
Bug: 'make html' command raises error while building docs


Darshan808:
## Description

When running the make html command to build the documentation, an error occurs. Specifically, the build process fails with a FileNotFoundError. This error arises because the api directory is missing in the docs/source directory.
![Screenshot (125)](https://github.com/jupyterlab/jupyterlab/assets/127468609/9baa653a-e03d-463e-b4f1-bb4d8f09ee48)

## Reproduce

1. cd into docs
2. Run command 'make html'

## Expected behavior

The make html command should successfully build the documentation without any errors and all the HTML build files shoul be in build/html.

## Context

- Operating System and version: Windows 11
- JupyterLab version: latest clone

PR #16467, opened 2024-06-10
Ensure api Directory is Present to Build Documentation

Darshan808:
## References

This PR addresses the issue #16466 

## Issue

When running make html, the build process fails with the following error:
`FileNotFoundError: [WinError 3] The system cannot find the path specified: '...\\jupyterlab\\docs\\source\\api'`

## Code changes

This PR addresses an issue where the documentation build process fails if the api directory is not present. Specifically, running make html results in a FileNotFoundError because Sphinx expects the docs/source/api directory to exist.

Since the api directory is under .gitignore, it is not present in a newly cloned repo. To solve this, I added a few lines of code to create the api directory if it is not present during the documentation build process. This ensures that the api directory is always present, so Sphinx doesn't raise any errors.

### After modifying the make.bat file, docs build successfully in windows:

Screenshot 2024-06-10 074142

## Additional Context:

1. This change ensures that new contributors or users can build the documentation without encountering this error.
2. This PR helps improve the developer experience by preventing a common pitfall during the setup process.

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -11,10 +11,16 @@ if "%SPHINXBUILD%" == "" (
 	set SPHINXBUILD=python -msphinx
 )
 set SOURCEDIR=source
+set APIDIR=%SOURCEDIR%\api
 set BUILDDIR=build
 set SPHINXPROJ=JupyterLab
 
 if "%1" == "" goto help
+
+if not exist "%APIDIR%" (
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you @Darshan808!

krassowski:

PR #16470, opened 2024-06-10
Trigger the toolbar button action on click instead of on mousedown

brichet:
This PR is an alternative of https://github.com/jupyterlab/jupyterlab/pull/16452, according to the suggestion of @fcollonval.

## References

Partial fix for https://github.com/jupyterlab/jupyterlab/issues/15629

## Code changes

Trigger the toolbar button action on click instead of on mousedown by default.

**EDIT**: to avoid the issue with the Notebook panel, mentioned [here](https://github.com/jupyterlab/jupyterlab/pull/16470#issuecomment-2160025585), the main area widgets still use the mousedown event by default.
To handle it, a new option is added to the toolbar and the buttons (`noFocusOnClick`), specifying whether the buttons should get the focus on click.
To be more specific, we could think of adding this option to the toolbar schema, and propagate it to the buttons created from the `ToolbarFactory`.

## User-facing changes

None

AFAIU triggering the action `onmousedown` was to avoid giving the focus to the button on most browser ([except on Safari](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus))... not really sure why. Anyway, with the current change, the focus seems to be properly given to the parent widget on Firefox and Chrome.


brichet:
I'm a bit puzzled with the behavior of this PR against https://github.com/jupyterlab/jupyterlab/pull/16452.

The failing test in this PR is expected because triggering the action `onclick` move the focus to the button. Therefore the function [`getState()`](https://github.com/jupyterlab/jupyterlab/blob/6c87a239faf592618dad68c020a3cb6f83a00cf6/packages/notebook/src/actions.tsx#L2309) does not set the notebook widget as active (`wasFocused`), and the [`activate`](https://github.com/jupyterlab/jupyterlab/blob/6c87a239faf592618dad68c020a3cb6f83a00cf6/packages/notebook/src/actions.tsx#L2330) function is not triggered.

But I don't understand why this does not append with https://github.com/jupyterlab/jupyterlab/pull/16452, which manually set the focus to the toolbar...

And the problem is that we expect 2 different behaviors for the Notebook widget and the Filebrowser. 
In the case of the Notebook, we want the focus to stay on the Notebook to handle `getState()` after click a toolbar button, and in the case of the Filebrowser we expect a child to get the focus (namely the toolbar or the button) to trigger the `focusTracker`.

brichet:
Thinking again about this, the special case is only for Notebook, which expects the focus to stay on the Notebook widget when clicking on the button.

Maybe a way to fix it could be to add an optional `arg` to the command arguments, informing on how the command has been triggered (button click, button keyboard, command palette...). Each command could use or not this arg for specific use case.

In the case of the Notebook, we can probably assume that clicking on the button or using the command palette should give the focus to the Notebook, but not using the keyboard to add a cell from the button, for example.

brichet:
@fcollonval @jtpio, any opinion on this implementation ?
It may be a breaking change for some extensions that use 'no-focus' behavior.

> To be more specific, we could think of adding this option to the toolbar schema, and propagate it to the buttons created from the `ToolbarFactory`.

Would it be worthwhile to modify the schema ?

brichet:
As said [here](https://github.com/jupyterlab/jupyterlab/issues/15629#issuecomment-2164795154) this is a workaround.
A better implementation would be to always get the focus on the widget when clicking on a toolbar button, and find a better way to handle the special case of the notebook, which need to know the focus status before the button was clicked.

jtpio:
Thanks @brichet for working on this!

>  It may be a breaking change for some extensions that use 'no-focus' behavior.

Looking at the changes, there does not seem to be introducing breaking changes per se. Not sure we consider the use of `no-focus` as part of the API surface?

Even if it's a workaround, it would still be worth getting this in, as the issue seems to always be happening when a drive is installed.

cc @fcollonval for input, otherwise I would be in favor of merging this PR :+1:

jtpio:
I just tried this PR in a clean environment and it does seem to be fixing the issue when creating a new folder from the filebrowser toolbar:

[jupyterlab-github-filebrowser.webm](https://github.com/user-attachments/assets/176e18f4-6b43-45e0-b78c-5c2b31f0a226)

However it looks like the issue is still here when trying to create a new notebook from the launcher. This seems to be the case after clicking on the input area of the `jupyterlab-github` extension, which may still be keeping the focus?

brichet:
Thanks for testing it @jtpio.

> However it looks like the issue is still here when trying to create a new notebook from the launcher. This seems to be the case after clicking on the input area of the `jupyterlab-github` extension, which may still be keeping the focus?

This is another issue (maybe) but yes, displaying a side panel does not change the current widget in the related `WidgetTracker`.
AFAIU switching the panel does not set the focus to it, so it does not trigger the focus tracker, and therefore the change is not caught by the widget tracker.

jtpio:
OK ,so maybe we still need to keep https://github.com/jupyterlab/jupyterlab/issues/15629 open to track the remaining issue.

jtpio (approved):
Thanks @brichet!

This fixes one of the issues reported in https://github.com/jupyterlab/jupyterlab/issues/15629 so it's already an improvement.

I updated the top comment to not close https://github.com/jupyterlab/jupyterlab/issues/15629 after merging this PR to keep track of the remaining focus issue when switching side panels.

PR #16471, opened 2024-06-10
Use correct hub restart URL

mahendrapaipuri:
## Code changes

Seems like `hub` extension is never taking the `restartUrl` for the current single user server. This PR corrects it by using the `restartUrl` of current single user server when config parameters are found. If not, `restartUrl` will be set to spawn page.

## User-facing changes

Users will be able to restart their servers instead of redirecting to spawn page

NA

@krassowski Does it make sense? I think it was a regression introduced in [this PR](https://github.com/advisories/GHSA-44cc-43rp-5947).

### DIFF
--- a/.github/workflows/linuxjs-tests.yml
+++ b/.github/workflows/linuxjs-tests.yml
@@ -35,6 +35,7 @@ jobs:
             js-documentsearch,
             js-filebrowser,
             js-fileeditor,
+            js-hub-extension,
             js-imageviewer,
             js-inspector,
             js-logconsole,
--- a/packages/hub-extension/babel.config.js
+++ b/packages/hub-extension/babel.config.js
@@ -1,0 +1,6 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+module.exports = require('@jupyterlab/testing/lib/babel-config');
--- a/packages/hub-extension/jest.config.js
+++ b/packages/hub-extension/jest.config.js
@@ -1,0 +1,7 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+const func = require('@jupyterlab/testing/lib/jest-config');
+module.exports = func(__dirname);
--- a/packages/hub-extension/package.json
+++ b/packages/hub-extension/package.json
@@ -30,7 +30,12 @@
   ],
   "scripts": {
     "build": "tsc",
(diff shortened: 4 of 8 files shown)


krassowski:
Makes sense, thank you for catching this! I think we should add a test for this (and for the issue that was patched by https://github.com/advisories/GHSA-44cc-43rp-5947) - are you happy to add a test here or would you like help in setting it up (I see that hub-extension does not have a tests folder yet so you may need to use `jlpm create:test` or copy the directory structure from another package and update `package.json` accordingly)

mahendrapaipuri:
Cheers @krassowski ! Sure, I will it a go and try to add a test.

mahendrapaipuri:
@krassowski Added a basic unit test for `hub-extension` with my limited `jest` skills. Let me know what do you think!

mahendrapaipuri:
@krassowski Ahhh, by mistake I synced the JupyterLab's `main` directly into the source branch of PR instead of forked `main`. Is it ok or should I clean it up and rebase?

krassowski:
It's ok, we do squash merges anyways.

krassowski:
Can you add `js-hub-extension` line to `.github/workflows/linuxjs-tests.yml` so that these tests are run on CI?

mahendrapaipuri:
@krassowski Done! Thought test discovery is automatic.

krassowski (approved):
Thank you @mahendrapaipuri!

krassowski:

guyq1997:
This still does not work for me. Even with the 4.2.4 Version.
The restartURL is still /hub/spawn.

mahendrapaipuri:
@guyq1997 If your single user server is running, `/hub/spawn/<user>` should automatically redirect to that running server at `/user/<user>/lab`. I can verify it on our deployment using JupyterLab 4.2.4

guyq1997:
@mahendrapaipuri 
Thanks a lot for your response! What if the Server is stopped by the Idle Culler, and the dialog tells user to start it again? The current server will not be opened. This is also important. Because users always want to get back to the server, where they were.

mahendrapaipuri:
> What if the Server is stopped by the Idle Culler, and the dialog tells user to start it again?

If you are using default server, hitting the URL `/hub/spawn/<user>` should trigger a new spawn which should eventually spawn a new single user server. However, if you have named servers enabled and a spawner form (to configure your single user servers), it might be tricky to start a new spawn directly from single user server as we need a `_xsrf` token and current spawner config to submit spawn data which single user server might not have access to.

Issue #15255, opened 2023-10-11
Open notebook with existing kernel


davidbrochart:
### Problem

JupyterLab has the ability to connect to external kernels, i.e. kernels that it didn't launch itself but that were started by another application. Unfortunately I didn't find a way to run a notebook with an external kernel:

- Right-clicking on the kernel in the left tab looks like this:

![image](https://github.com/jupyterlab/jupyterlab/assets/4711805/51a50558-f944-402c-a7e4-331c536cd79b)

We can only create a new notebook for the kernel.

- Right-clicking on the notebook in the file explorer and choosing "Open With" will launch a new kernel:

![image](https://github.com/jupyterlab/jupyterlab/assets/4711805/b459fd20-9075-44f5-bb6d-d8767562d4de)

- Opening an existing notebook and trying to select a kernel for it doesn't allow to pick an existing kernel (it will only launch a new kernel):

![image](https://github.com/jupyterlab/jupyterlab/assets/4711805/ae2e292c-baab-4933-b148-d04ca0d5bd29)

### Proposed Solution

I don't know what would be the best solution, but having a way to open a notebook with an existing kernel should be possible.
Currently, my workaround is to create a new notebook for this kernel and copy-paste all cells, not ideal.

cc @afshin

davidbrochart:
Actually there is a way to do this, but it involves creating e.g. a console for the kernel first, so that a session is created with this kernel:
- right-click on the kernel in the left tab, and select "New Console for Kernel".
- in the file browser, double-click on the notebook.
- in the "Select Kernel" menu, under "Use Kernel from Other Session" at the bottom, select "Console 1".

PR #16475, opened 2024-06-10
Update notebook kernel selector

afshin:
The "change kernel" dialog is complex, considering that it is primarily just a `<select>` dropdown. Since the release of the running kernels tab, though, the notion of what we mean when we say "kernel" has become somewhat muddled. The "change kernel" dialog actually shows you kernel specs against which a new kernel instance can be created or running **sessions** in JupyterLab. However, the running kernels sidebar shows you a list of running **kernels** and uses any session that has more information about a kernel instance to update the UI, but not to group the kernels.

This PR updates the "change kernel" UI:
- It slightly changes the language in the `<optgroup>` sections of the dropdown to clarify that "preferred" means "using the programming language I used last time".
- It lists every running kernel as a choice – even if there is no JupyterLab session associated with it.

## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/15255
cf. https://github.com/jupyterlab/jupyterlab/issues/16340

## Code changes
- This PR updates the change kernel dialog to include all kernel sessions as well as all kernelspecs.
- It adds an optional kernel manager to session contexts, in future versions, this should be required.
- It exposes a new function `SessionContextDialogs.kernelOptions()` that returns grouped kernelspecs and running kernel options that can be used to build a UI (*e.g.*, a `<select>` tag or React component).
- It adds a test for checking whether external kernels that JupyterLab's REST endpoint are reflected in the kernel options presented to the user.

#### before
before

#### after
after

N/A

**NB** The `kernelManager` option for `SessionContext` is currently optional. In the next major version of this package, we should make it required.

The `SessionContext` instances created by JupyterLab core have all been updated to pass in a kernel manager.

If there is no kernel manager available (i.e. if there is a plugin that creates `SessionContext` instances that do not have a kernel manager), the dropdown will harvest all known kernels from running sessions, which is roughly the old behavior.

## To test this yourself

I have updated the tests to include this use case, but I can show how to test this pretty easily in your own environment, it doesn't require installing anything exotic if all we want is to see these UI changes and don't care about whether there is a real underlying kernel to work with.

Create a `kernel.json` file with contents like this in an otherwise empty directory (I call mine `external_kernels`):

```json
{
    "transport": "tcp",
    "ip": "127.0.0.1",
    "control_port": 63607,
    "shell_port": 49458,
    "stdin_port": 54582,
    "iopub_port": 50266,
    "hb_port": 59840,
    "signature_scheme": "hmac-sha256",
    "key": "883965661293433ea9c07538beeb8451",
    "kernel_name": "mock-kernel"
}
```

Start JupyterLab with `ServerApp.allow_external_kernels` set to `True` and `ServerApp.external_connection_dir` set to the directory you created above. The command I used also included `--dev-mode` and `--watch` and a blank identity provider because I was in development mode as well as `extensions-in-dev-mode` to allow me to see `jupyterlab_launchpad`:

```
jupyter lab \
--dev-mode --extensions-in-dev-mode --watch \
--no-browser \
--IdentityProvider.token="" \
--ServerApp.external_connection_dir=./external_kernels \
--ServerApp.allow_external_kernels=True
```


krassowski:
> The `SessionContext` class now requires a `kernelManager` in its `options`. Before this PR, it only required a `sessionManager` and a `specsManager`. After this PR, all `SessionContext` instantiations will need this argument.

My first thought was whether this will break https://github.com/nebari-dev/jupyterlab-new-launcher (to be renamed to `jupyter-launchpad`) but from looking at code only it should not. Still it would be good to test it.

SylvainCorlay:
Thanks! I am really happy to see this come to completion!

packages/apputils/src/sessioncontext.tsx
```diff
@@ -337,6 +343,11 @@ export class SessionContext implements ISessionContext {
    * Construct a new session context.
    */
   constructor(options: SessionContext.IOptions) {
+    this.kernelManager =
+      options.kernelManager ??
+      new KernelManager({
```
afshin, line 348: We should remove this and make `options.kernelManager` required for the next major version of `@jupyterlab/apputils`

afshin:
bot please update snapshots

packages/apputils/src/sessioncontext.tsx
```diff
@@ -196,6 +197,11 @@ export interface ISessionContext extends IObservableDisposable {
    */
   readonly prevKernelName: string;
 
+  /**
+   * The kernel manager
+   */
+  readonly kernelManager: Kernel.IManager;
```
jtpio, line 203: Should this be made optional?

It looks like by adding this to the interface, downstream code that "implements ISessionContext" will now have to provide a `kernelManager`, which may be considered as breaking.

jtpio: Although there is likely not many downstreams implementing their own `ISessionContext`, so another option could also be to document this change in https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html for visibility.

afshin: I did a web search and couldn't find any public instance of `implements ISessionContext` outside of JupyterLab, but you're totally right about the actual API contract. I've updated the code.

jtpio:
@afshin maybe it could be useful to post a small snippet on how to start an external kernel and have it show up in the JupyterLab UI, so it's easier to test? Which likely involves setting `--ServerApp.allow_external_kernels` and `--ServerApp.external_connection_dir`.

> My first thought was whether this will break https://github.com/nebari-dev/jupyterlab-new-launcher (to be renamed to `jupyter-launchpad`) but from looking at code only it should not. Still it would be good to test it.

Trying from this branch with `jupyterlab-new-launcher` installed, it looks like the external kernel does not show up in the dialog:

image

Not sure it's an issue with this PR, or that `jupyterlab-new-launcher` should take these changes into account to show the kernel in the dialog.

krassowski:
Thank you for testing it!

> Not sure it's an issue with this PR, or that `jupyterlab-new-launcher` should take these changes into account to show the kernel in the dialog.

I think extensions will need to adopt to benefit from this change, but it is ok - the important thing is that it does not break the API in any way.

> @afshin maybe it could be useful to post a small snippet on how to start an external kernel and have it show up in the JupyterLab UI, so it's easier to test? Which likely involves setting `--ServerApp.allow_external_kernels` and `--ServerApp.external_connection_dir`.

Seconded, but even better would be also having an integration test here. This is a relatively rare use case right now so I would be afraid to touch the code in fear of breaking it if there is no test for it.

afshin:
I installed `jupyterlab_launchpad` on my branch to play around with it. I confirm what @jtpio notes above. The change kernel dialog is iterating through `/api/sessions`, finding none, and only showing the available kernelspecs (which is exactly how JupyterLab currently behaves).

The change kernel dialog in this PR iterates over every available item in the `/api/kernels` endpoint and uses any corresponding information from the `/api/sessions` endpoint as decoration for labeling each kernel with more contextual information. So `jupyterlab_launchpad` could pretty easily switch to doing this as well.

### This branch displaying change kernel dialog of `jupyterlab_launchpad`

This branch displaying change kernel dialog of `jupyterlab_launchpad`

### This branch displaying core change kernel dialog

This branch displaying core change kernel dialog

### To test this yourself

I am not sure how to update the tests to include this use case, but I can show how to test this pretty easily, it doesn't require installing anything exotic if all we want is to see these UI changes and don't care about whether there is a real underlying kernel to work with.

Create a `kernel.json` file with contents like this in an otherwise empty directory (I call mine `external_kernels`):

```json
{
    "transport": "tcp",
    "ip": "127.0.0.1",
    "control_port": 63607,
    "shell_port": 49458,
    "stdin_port": 54582,
    "iopub_port": 50266,
    "hb_port": 59840,
    "signature_scheme": "hmac-sha256",
    "key": "883965661293433ea9c07538beeb8451",
    "kernel_name": "mock-kernel"
}
```

Start JupyterLab with `ServerApp.allow_external_kernels` set to `True` and `ServerApp.external_connection_dir` set to the directory you created above. The command I used also included `--dev-mode` and `--watch` and a blank identity provider because I was in development mode as well as `extensions-in-dev-mode` to allow me to see `jupyterlab_launchpad`:

```
jupyter lab --dev-mode --extensions-in-dev-mode --watch \
--no-browser \
--IdentityProvider.token="" \
--ServerApp.external_connection_dir=./external_kernels \
--ServerApp.allow_external_kernels=True
```

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1484,21 +1510,241 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
...
+   *    - `No Kernel`
+   *  - (Connect to Existing Kernel)
+   *    - { all running kernels in alphabetical order  }
+   *
+   * If `options.preference.id` is set and a kernel exists with that id, its
+   * corresponding `<option>` is selected. Otherwise if
+   * `options.preference.language` is set, the first kernelspec that matches it
+   * is selected.
+   */
+  export function populateKernelSelect(
```
fcollonval, line 1546: As this aims to be public, it would be more useful to return the list of options (probably grouped, something like `{ [group: string]: KernelOption[] }`). Like that if someone wants to define its own flavor of the dialog, it can reuse this and we don't expose a function with an interface too close to the UI implementation details.

fcollonval (changes requested):

packages/apputils/src/sessioncontext.tsx
afshin: Fair enough. I think because a `<select>` tag is a known quantity it's not so bad. But I agree with your design principle here. The only reason I exposed it was for testing, but even for this reason, it's nicer to taste data validity than to use DOM methods to query values. I'll update this today.

afshin:
@krassowski the new `SessionContextDialogs.kernelOptions` might be relevant to your interests.

packages/apputils/src/sessioncontext.tsx
```diff
@@ -1484,21 +1510,290 @@ export class SessionContextDialogs implements ISessionContext.IDialogs {
...
+    selected?: boolean;
+    text: string;
+    title?: string;
+    value: string;
+  };
+
+  type KernelOptionGroup = {
+    label: string;
+    options: Array<KernelOption>;
+  };
```
krassowski, line 1529: Since `KernelOptions` is returned by an exported function, should it be exported too, and by extension, should the other two be exported and documented too?

Should these by types or interfaces? Types can get annoying in TS when working with multiple versions of dependencies.

afshin: I'm happy to change them to interfaces.

I don't have strong feelings regarding whether they should be public. I think maybe keeping them private is slightly preferable because these aren't really types we expect others to create and a smaller public footprint is nicer.

Do you think they ought to be public interfaces?

afshin: I moved everything into one public `interface`.

fcollonval (approved):
Thanks @afshin

krassowski (approved):
I was unable to test this locally with `external_connection_dir` approach - nothing showed up for me. That said, I think the code looks fine and I do not want to block this PR from being included in alpha for testing so I will approve and merge.

PR #16482, opened 2024-06-12
Fix check link CI failure in README (time zone converter site)

afshin:
The `thetimezoneconverter.com` website is unresponsive and the README link to it causes the CI job that checks links to fail. This PR removes the link from the README.

## References
N/A

## Code changes
N/A

## User-facing changes
N/A

N/A


krassowski:
It looks like this is responsive again. Should we revert this PR?

Issue #16364, opened 2024-05-23
Inline completer should not resize the cell editor aggressively when streaming 


krassowski:
## Description

When inline completions are streaming sometimes the editor height changes aggressively. This is not accessible especially if user prefer reduced motion.

## Reproduce

1. Use https://github.com/krassowski/jupyterlab-transformers-completer
2. Change settings to always stream
3. Start typing and invoke inline completer
4. See how the cell height changes drastically when streaming and when switching between alternative suggestions

## Expected behavior

Ideas to explore:
- adding a configurable maximum height for the ghost text, e.g. 3-4 lines of code
   - as a variation on this, expanding the following lines of the ghost text only when user stops typing, after a configurable timeout - if they started reading the suggestion they will be interested to see more lines - this allows to reduce the max height to 2-3 lines without detriment to users ability to read the entire thing before accepting
- if there are multiple suggestions of variable height keeping reserved space for the suggestion with the most lines so that when user cycles between suggestions the cell does not resize
- keeping some constant reserved space, e.g. minimum editor height when inline completer is enabled
- keeping the empty space for a while after last suggestion was invalidated so that the editor does not bounce when a new suggestion comes in (this is already implemented but the timeout is very short - maybe we should increase it or make it configurable)

## Context

- JupyterLab version: 4.2.0

krassowski:
On particular bug is in how the line spacer is not working correctly; it is meant to preserve the space below and then animate collapse slowly to avoid jitter, but when it gets activated it instead increases the white area.

![incorrect-styling-of-spacer](https://github.com/jupyterlab/jupyterlab/assets/5832902/7eeab851-075c-445b-b528-1c14010d9dd9)

I narrowed it down to `max-width: 0` in:

https://github.com/jupyterlab/jupyterlab/blob/5e4cae5d0b3b4f080caf7afffbfb381bdf80cc5b/packages/completer/style/base.css#L216-L222

However, that was added here for a reason: it was meant to ensure that the placeholder does not expand the scrollbar; without it this is what happens:

![scrollbar2](https://github.com/jupyterlab/jupyterlab/assets/5832902/1f770aa0-27c1-4eec-896a-f2f7f922a84f)

PR #16493, opened 2024-06-16
Fix the lines placeholder taking up too much space

krassowski:
## References

Fixes issue in https://github.com/jupyterlab/jupyterlab/issues/16364#issuecomment-2171826119

## Code changes

- [x] Adds a regression test
- [x] Fixes the issue by replacing `max-width: 0` on `.jp-GhostText-lineSpacer` with `text-wrap: none` to ensure the right number of lines is shown.

Longer explanation of the fix:

Previously `max-width: 0` was used to prevent the placeholder from expanding in width (and thus showing a spurious scrollbar when cell is narrow enough). However, it also meant that sometimes more lines were rendered because the width restriction was causing new line breaks to be added.

> This was a problem due to the necessity of using `display: inline-block` to animate font size reduction effectively.

Now, instead of using `max-width: 0` we use `text-wrap: none` which:
- prevents the horizontal scrollbar from showing when not needed and shows it when needed
- populates the vertical space to preserve the height

## User-facing changes

When user types a character that cancels the suggestion (e.g. `x` on gif below) the editor height now remains the same; as before, if they then press backspace the previous suggestion re-appears (as demonstrated on the gif).

| Before | After |
|--|--|
| before| after |

None


krassowski:
A few less important edge case:
- pasting multiple lines extends the height but if it the pasted height is less than the spacer we could just reduce the spacer height to limit jitter
- similarly tping a letter extends the scrollbar even if `text-wrap: none` is used

krassowski:
There are more things that can be improved, but I am marking this as ready for review because I think this PR as of now addresses the most severe issue and can be easily backported; I will open additional PRs with more improvements.

jtpio (approved):
Thanks!

krassowski:

PR #16498, opened 2024-06-18
Improve stream output handling

davidbrochart:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16515.
Needs https://github.com/jupyter-server/jupyter_ydoc/pull/241.

In https://github.com/jupyter-server/jupyter_ydoc/pull/201, code cell outputs changed from an untyped `Y.Array` to a `Y.Array` where each item is a `Y.Map`. For stream outputs, the `text` entry of this `Y.Map` becomes a `Y.Text`. This is to handle incoming stream data in an incremental and efficient way, by just modifying the `text`. Previously, the whole output was discarded and a new one was created, identical to the previous one except for the `text` that had the new data, leading to catastrophic memory consumption.

## Code changes

This PR takes advantage of this new schema, by doing the equivalent change in the `OutputModel`, where stream outputs have a `text` entry of type `ObservableString`.

## User-facing changes

Better performances, especially in collaborative mode.

None.

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -22,7 +22,7 @@
     "@codemirror/view": "^6.9.6",
     "@jupyter/react-components": "^0.16.6",
     "@jupyter/web-components": "^0.16.6",
-    "@jupyter/ydoc": "^2.0.1",
+    "@jupyter/ydoc": "^3.0.0-a2",
     "@jupyterlab/application": "~4.3.0-alpha.2",
     "@jupyterlab/application-extension": "~4.3.0-alpha.2",
     "@jupyterlab/apputils": "~4.4.0-alpha.2",
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -13,11 +13,7 @@ import { IChangedArgs } from '@jupyterlab/coreutils';
 
 import * as nbformat from '@jupyterlab/nbformat';
 
-import {
-  IObservableString,
-  ObservableString,
-  ObservableValue
-} from '@jupyterlab/observables';
+import { IObservableString, ObservableValue } from '@jupyterlab/observables';
 
 import { IOutputAreaModel, OutputAreaModel } from '@jupyterlab/outputarea';
 
@@ -765,9 +761,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
         case 'add': {
           for (const output of event.newValues) {
             if (output.type === 'stream') {
-              (
-                output.observableData.get('text') as unknown as ObservableString
-              ).changed.connect(
+              output.observableText!.changed.connect(
                 (
                   sender: IObservableString,
                   textEvent: IObservableString.IChangedArgs
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -13,7 +13,11 @@ import { IChangedArgs } from '@jupyterlab/coreutils';
 
 import * as nbformat from '@jupyterlab/nbformat';
 
-import { ObservableValue } from '@jupyterlab/observables';
+import {
+  IObservableString,
+  ObservableString,
+  ObservableValue
+} from '@jupyterlab/observables';
 
 import { IOutputAreaModel, OutputAreaModel } from '@jupyterlab/outputarea';
 
@@ -759,6 +763,32 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
     globalModelDBMutex(() => {
       switch (event.type) {
         case 'add': {
+          for (const output of event.newValues) {
+            if (output.type === 'stream') {
+              (
+                output.observableData.get('text') as unknown as ObservableString
+              ).changed.connect(
+                (
+                  sender: IObservableString,
+                  textEvent: IObservableString.IChangedArgs
+                ) => {
+                  const codeCell = this.sharedModel as YCodeCell;
+                  if (textEvent.type === 'remove') {
+                    codeCell.removeStreamOutput(
+                      event.newIndex,
+                      textEvent.start
+                    );
+                  } else {
+                    codeCell.appendStreamOutput(
+                      event.newIndex,
+                      textEvent.value
+                    );
+                  }
+                },
+                this
+              );
+            }
+          }
           const outputs = event.newValues.map(output => output.toJSON());
           codeCell.updateOutputs(event.newIndex, event.newIndex, outputs);
           break;
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -761,7 +761,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
         case 'add': {
           for (const output of event.newValues) {
             if (output.type === 'stream') {
-              output.observableText!.changed.connect(
+              output.streamText!.changed.connect(
                 (
                   sender: IObservableString,
                   textEvent: IObservableString.IChangedArgs
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -790,7 +790,12 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
             }
           }
           const outputs = event.newValues.map(output => output.toJSON());
-          codeCell.updateOutputs(event.newIndex, event.newIndex, outputs);
+          codeCell.updateOutputs(
+            event.newIndex,
+            event.newIndex,
+            outputs,
+            'modeldb'
+          );
           break;
         }
         case 'set': {
@@ -798,12 +803,18 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
           codeCell.updateOutputs(
             event.oldIndex,
             event.oldIndex + newValues.length,
-            newValues
+            newValues,
+            'modeldb'
           );
           break;
         }
         case 'remove':
-          codeCell.updateOutputs(event.oldIndex, event.oldValues.length);
+          codeCell.updateOutputs(
+            event.oldIndex,
+            event.oldValues.length,
+            [],
+            'modeldb'
+          );
           break;
         default:
           throw new Error(`Invalid event type: ${event.type}`);
@@ -820,8 +831,23 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
   ): void {
     if (change.outputsChange) {
       globalModelDBMutex(() => {
-        this.outputs.clear();
-        slot.getOutputs().forEach(output => this._outputs.add(output));
+        let retain = 0;
+        for (const outputsChange of change.outputsChange!) {
+          if ('retain' in outputsChange) {
+            retain += outputsChange.retain!;
+          }
+          if ('delete' in outputsChange) {
+            for (let i = 0; i < outputsChange.delete!; i++) {
+              this._outputs.remove(retain);
+            }
+          }
+          if ('insert' in outputsChange) {
+            // Inserting an output always results in appending it.
+            for (const output of outputsChange.insert!) {
+              this._outputs.add(output.toJSON());
+            }
+          }
+        }
       });
     }
 
--- a/packages/cells/src/model.ts
+++ b/packages/cells/src/model.ts
@@ -794,7 +794,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
             event.newIndex,
             event.newIndex,
             outputs,
-            'modeldb'
+            'silent-change'
           );
           break;
         }
@@ -804,7 +804,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
             event.oldIndex,
             event.oldIndex + newValues.length,
             newValues,
-            'modeldb'
+            'silent-change'
           );
           break;
         }
@@ -813,7 +813,7 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
             event.oldIndex,
             event.oldValues.length,
             [],
-            'modeldb'
+            'silent-change'
           );
           break;
         default:
--- a/packages/cells/src/widget.ts
+++ b/packages/cells/src/widget.ts
@@ -1390,26 +1390,30 @@ export class CodeCell extends Cell<ICodeCellModel> {
     // to changes in metadata until we have fully committed our changes.
     // Otherwise setting one key can trigger a write to the other key to
     // maintain the synced consistency.
-    this.model.sharedModel.transact(() => {
-      super.saveCollapseState();
+    this.model.sharedModel.transact(
+      () => {
+        super.saveCollapseState();
 
-      const collapsed = this.model.getMetadata('collapsed');
+        const collapsed = this.model.getMetadata('collapsed');
 
-      if (
-        (this.outputHidden && collapsed === true) ||
-        (!this.outputHidden && collapsed === undefined)
-      ) {
-        return;
-      }
+        if (
+          (this.outputHidden && collapsed === true) ||
+          (!this.outputHidden && collapsed === undefined)
+        ) {
+          return;
+        }
 
-      // Do not set jupyter.outputs_hidden since it is redundant. See
-      // and https://github.com/jupyter/nbformat/issues/137
-      if (this.outputHidden) {
-        this.model.setMetadata('collapsed', true);
-      } else {
-        this.model.deleteMetadata('collapsed');
-      }
-    }, false);
+        // Do not set jupyter.outputs_hidden since it is redundant. See
+        // and https://github.com/jupyter/nbformat/issues/137
+        if (this.outputHidden) {
+          this.model.setMetadata('collapsed', true);
+        } else {
+          this.model.deleteMetadata('collapsed');
+        }
+      },
+      false,
+      'modeldb'
+    );
   }
 
   /**
@@ -1722,9 +1726,13 @@ export namespace CodeCell {
     const model = cell.model;
     const code = model.sharedModel.getSource();
     if (!code.trim() || !sessionContext.session?.kernel) {
-      model.sharedModel.transact(() => {
-        model.clearExecution();
-      }, false);
+      model.sharedModel.transact(
+        () => {
+          model.clearExecution();
+        },
+        false,
+        'modeldb'
+      );
       return;
     }
     const cellId = { cellId: model.sharedModel.getId() };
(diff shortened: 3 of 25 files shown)


packages/outputarea/src/model.ts
```diff
@@ -309,24 +309,21 @@ export class OutputAreaModel implements IOutputAreaModel {
...
+      // This creates a text list change event.
       const index = this.length - 1;
-      const prev = this.list.get(index);
-      this.list.set(index, item);
-      prev.dispose();
+      const prev = this.list.get(index) as IOutputModel;
+      const textList = prev.observableData.get(
+        'text'
+      ) as unknown as IObservableList<string>;
+      const newText = Private.removeOverwrittenChars(value.text as string);
```
davidbrochart, line 325: Here `removeOverwrittenChars` is only called on the new text, not on the whole concatenated text as previously. This means that a backspace in a stream output data cannot affect the previous data. For instance:
```python
from time import sleep

print("abc", end="")
sleep(0.5)
print("\n")
```
Will not not result in the output `ab`, but in `abc`.

krassowski:
> Better performances, especially in collaborative mode.

Do you have any benchmarks to give an idea when the benefits start to materialize and how much of a speed up one could expect?

If the change in behaviour for overwriting characters is a requirement rather than a work-in-progress state it would need to be a really substantial benefit to consider breaking the existing behaviour.

davidbrochart:
> Do you have any benchmarks to give an idea when the benefits start to materialize and how much of a speed up one could expect?

It's rather on the memory side that things will improve. Currently, memory grows exponentially with streamed data, which is very bad. With this PR, it will grow linearly. We have seen cases where the server eventually crashes for big outputs.

> If the change in behaviour for overwriting characters is a requirement rather than a work-in-progress state it would need to be a really substantial benefit to consider breaking the existing behaviour.

I can probably restore the original behavior, maybe by changing the list of text into a text.

packages/outputarea/src/model.ts
```diff
@@ -484,38 +488,26 @@ namespace Private {
...
+        if (
+          curText.text.length > 0 &&
+          curText.text[curText.text.length - 1] != '\n'
+        ) {
+          curText.remove(curText.text.length - 1, curText.text.length);
+        }
+      } else {
+        curText.insert(curText.text.length, character);
+      }
+    }
```
davidbrochart, line 511: This loop will create a text change event for every character in the new text (and accordingly, a refresh of the output in the UI), which is far from optimal.
It is possible to optimize it, but this would probably not be needed if we were directly using Yjs, as we would do all these changes in a single transaction. Unfortunately, ModelDB doesn't seem to support this feature.

davidbrochart, line 511: This loop will create a text change event for every character in the new text (and accordingly, a refresh of the output in the UI), which is far from optimal.
It is possible to optimize it, but this would probably not be needed if we were directly using Yjs, as we would do all these changes in a single transaction. Unfortunately, ModelDB doesn't seem to support this feature.

davidbrochart:
> If the change in behaviour for overwriting characters is a requirement rather than a work-in-progress state it would need to be a really substantial benefit to consider breaking the existing behaviour.

I restored the initial behavior (and actually [fixed it](https://github.com/jupyterlab/jupyterlab/issues/16515)).

packages/outputarea/src/model.ts
davidbrochart: I did some optimizations in https://github.com/jupyterlab/jupyterlab/pull/16498/commits/d4219c5ad20e6181bd63c2a9a889ca737d2f6e0e.

davidbrochart:
The way shared output changes are handled currently is not optimal: any change in the Y model results in removing all previous outputs from the modelDB and adding all new outputs. This is because we were only supporting the initialization from a file, but other cases are appearing, for instance in server-side execution an output can be added. In this case we don't want to clear all outputs, but just handle the added output.
That's what https://github.com/jupyterlab/jupyterlab/pull/16498/commits/971d11a676c4b7b2587de70d127ebe368890a097 changes.

krassowski:
Thanks for working on it! It looks that a number of tests are failing due to this change. Once the tests pass I can help with resolving the merge conflicts to get this over the finish line. Also, having a test for the new concatenate function would be nice.

packages/outputarea/src/model.ts
```diff
@@ -255,13 +264,21 @@ export class OutputAreaModel implements IOutputAreaModel {
     return this._add(output);
   }
 
+  /**
+   * Remove an output at a given index.
+   */
+  remove(index: number): void {
+    this.list.get(index).dispose();
+    this.list.remove(index);
```
fcollonval, line 272: `remove` returns the item a more efficient pattern is:

```suggestion
    this.list.remove(index)?.dispose();
```

packages/outputarea/src/model.ts
```diff
@@ -484,38 +503,70 @@ namespace Private {
...
-  }
-
   /*
-   * Remove characters overridden by backspaces and carriage returns
-   */
-  export function removeOverwrittenChars(text: string): string {
-    return fixCarriageReturn(fixBackspace(text));
+   * Concatenate a string to an observable string, handling backspaces.
+   */
+  export function concatenateText(
```
fcollonval, line 509: This sounds like an improvement for backspaces but you drop the support for carriage return `\r`, isn't it?

packages/cells/src/widget.ts
```diff
@@ -1390,26 +1390,30 @@ export class CodeCell extends Cell<ICodeCellModel> {
...
+        // Do not set jupyter.outputs_hidden since it is redundant. See
+        // and https://github.com/jupyter/nbformat/issues/137
+        if (this.outputHidden) {
+          this.model.setMetadata('collapsed', true);
+        } else {
+          this.model.deleteMetadata('collapsed');
+        }
+      },
+      false,
+      'modeldb'
```
fcollonval, line 1415: I'm not a big fan of the word `modeldb` as it is not technically correct. As it is a trick to have a different behavior in the ydoc model (aka not emitting changes), something like `silent-change` sounds more appropriate for more.

dev_mode/package.json
```diff
@@ -22,7 +22,7 @@
     "@codemirror/view": "^6.9.6",
     "@jupyter/react-components": "^0.16.3",
     "@jupyter/web-components": "^0.16.3",
-    "@jupyter/ydoc": "^2.0.1",
+    "@jupyter/ydoc": "^3.0.0-a1",
```
fcollonval, line 25: Small note, if we start using alpha here we should create an issue in jupyterlab that forces us to wait for stable 3.0.0 prior to 4.3.0 release.

fcollonval (changes requested):
Thanks @davidbrochart

I get some questions and suggestions.

packages/outputarea/src/model.ts
davidbrochart: Thanks, done in https://github.com/jupyterlab/jupyterlab/pull/16498/commits/0bd367e40639658d4d1dc29ed41b3d5cb64a3a9b.

packages/outputarea/src/model.ts
davidbrochart: I rewrote it in https://github.com/jupyterlab/jupyterlab/pull/16498/commits/0bd367e40639658d4d1dc29ed41b3d5cb64a3a9b.

packages/cells/src/widget.ts
davidbrochart: Indeed, we'll need to change it in jupyter-ydoc too (see [here](https://github.com/jupyter-server/jupyter_ydoc/blob/023afaf3a6f177be52a97e9c731e5d7aed44820a/javascript/src/ycell.ts#L663)).

davidbrochart:
The [logic to handle new streams](https://github.com/jupyterlab/jupyterlab/blob/0bd367e40639658d4d1dc29ed41b3d5cb64a3a9b/packages/outputarea/src/model.ts#L508) is not very efficient, as it processes every character and does potentially quite a lot of string concatenations. Again, I think we could defer all of that to Yjs and it would do some optimizations if it's done in a single transaction.

packages/cells/src/widget.ts
davidbrochart: jupyter-ydoc v3.0.0a2 has `silent-change` instead of `modeldb`.

packages/rendermime/src/outputmodel.ts
```diff
@@ -82,6 +91,16 @@ export class OutputModel implements IOutputModel {
     const { data, metadata, trusted } = Private.getBundleOptions(options);
     this._data = new ObservableJSON({ values: data as JSONObject });
     this._rawData = data;
+    const newData: { [key: string]: any } = {};
```
krassowski, line 94: Can you replace `any` here with a proper typing, or worst case scenario an `unknown`?

packages/rendermime/src/outputmodel.ts
```diff
@@ -177,11 +200,21 @@ export class OutputModel implements IOutputModel {
     for (const key in this._raw) {
       output[key] = Private.extract(this._raw, key);
     }
+    const data: { [key: string]: any } = {};
```
krassowski, line 203: New code should not have `any` (except for tests).

packages/rendermime/src/outputmodel.ts
davidbrochart: If I replace `any` with `unknown` I get this error:
```
src/outputmodel.ts:103:39 - error TS2322: Type '{ [key: string]: unknown; }' is not assignable to type 'JSONObject'.
  'string' index signatures are incompatible.
    Type 'unknown' is not assignable to type 'JSONValue'.

104     this._data = new ObservableJSON({ values: newData });
                                          ~~~~~~
```

krassowski: That's why `unknown` is useful! Then, I guess it should be:

```suggestion
    const newData: { [key: string]: JSONValue } = {};
```

or `typeof JSONObject` or similar. The use of `any` has lead to countless bugs in JupyterLab which could have been avoided if we used proper typing (and often not because the author who introduced `any` made a mistake, but someone who came after them).

davidbrochart: But the problem is that for the `'text'` key we want to store an `ObservableString` value, which cannot be assigned to a `JSONValue` or a `PartialJSONValue`.

krassowski: So in that case `_data` is not a valid `ObservableJSON` and the public API in `observableData` will return an invalid value. It seems that you are introducing `observableData` but only ever taking the `text` out of it. Would it make sense to store the `ObservableString` separately and expose it as `observableText` getter instead?

davidbrochart: > Would it make sense to store the `ObservableString` separately and expose it as `observableText` getter instead?

Good point, I made that change in https://github.com/jupyterlab/jupyterlab/pull/16498/commits/87e04593ca4d2b1c0f028064ba09ba67f9764093.

packages/rendermime/src/outputmodel.ts
```diff
@@ -35,6 +40,11 @@ export interface IOutputModel extends IRenderMime.IMimeModel {
    */
   readonly executionCount: nbformat.ExecutionCount;
 
+  /**
+   * The observable data.
```
krassowski, line 44:
```suggestion
   * The observable data, present when the output `type` is set to `"stream"`.
```

packages/outputarea/src/model.ts
```diff
@@ -309,44 +325,43 @@ export class OutputAreaModel implements IOutputAreaModel {
...
+      const curText = prev.observableText!;
+      const newText = value.text as string;
+      Private.addText(curText, newText);
       return this.length;
     }
 
+    let newText = '';
     if (nbformat.isStream(value)) {
-      value.text = Private.removeOverwrittenChars(value.text as string);
+      newText = value.text as string;
```
krassowski, line 345: > I am fine with keeping it as-is and tackling later.

This is a pre-existing issue, but it looks like there should be no cast to `string` here; `isStream` is a type guard and correctly identifies that `value.text` exists and is of `string | string[]` type as per:

https://github.com/jupyterlab/jupyterlab/blob/6f202ff2d0ddf42414b09a27af1bbcb7954b51e7/packages/nbformat/src/index.ts#L458-L473

https://github.com/jupyterlab/jupyterlab/blob/6f202ff2d0ddf42414b09a27af1bbcb7954b51e7/packages/nbformat/src/index.ts#L65

```suggestion
      newText = Array.isArray(value.text) ? value.text.join('\n') : value.text;
```

Of note, `nbformat` source of truth schema also says that stream can be an array of strings https://github.com/jupyter/nbformat/blob/ba2c6f53afbeb18f6fad0c9b9f67def34e757cb4/nbformat/v4/nbformat.v4.schema.json#L360-L363

packages/rendermime/src/outputmodel.ts
```diff
@@ -35,6 +40,11 @@ export interface IOutputModel extends IRenderMime.IMimeModel {
...
+   */
+  readonly observableText?: IObservableString;
```
krassowski, line 46: Should we maybe name it `streamedText` or `streamText` instead to avoid calling it by type name and highlight that it is only relevant to streams?

krassowski (commented):
Some final questions, thanks so much again!

davidbrochart:
Thanks for the review @krassowski, I tried to take your review into account in https://github.com/jupyterlab/jupyterlab/pull/16498/commits/24cc98d28ab653e41d202e41ede484ecda124387.

packages/rendermime/src/outputmodel.ts
```diff
@@ -80,8 +90,10 @@ export class OutputModel implements IOutputModel {
    */
   constructor(options: IOutputModel.IOptions) {
     const { data, metadata, trusted } = Private.getBundleOptions(options);
-    this._data = new ObservableJSON({ values: data as JSONObject });
     this._rawData = data;
+    if (options.value !== undefined && nbformat.isStream(options.value)) {
+      this._text = new ObservableString(options.value['text'] as string);
```
krassowski, line 95: And here:

```suggestion
      this._text = new ObservableString(typeof options.value.text === 'string' ? options.value.text : options.value.text.join(''));
```

krassowski (commented):
Thanks! If we are including the fix for the `string[]` issue let's make sure to change it in all places:

davidbrochart:
Thanks @krassowski, I rebased on `main` and tried to add you as a co-author, but couldn't find your email address.

krassowski:
@fcollonval just checking back here?

krassowski:
Since it has been almost two weeks, I plan to merge tomorrow if there are no objections/new reviews :)

dev_mode/package.json
krassowski: I opened https://github.com/jupyterlab/jupyterlab/issues/16649

packages/cells/src/model.ts
```diff
@@ -790,8 +825,23 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
...
+          }
+          if ('delete' in outputsChange) {
+            for (let i = 0; i < outputsChange.delete!; i++) {
+              this._outputs.remove(retain);
+            }
+          }
+          if ('insert' in outputsChange) {
+            // Inserting an output always results in appending it.
+            for (const output of outputsChange.insert!) {
+              this._outputs.add(output.toJSON());
```
krassowski, line 841: This throws an error when used with `jupyter-server-nbmodel`:

```
TypeError: e.toJSON is not a function
    at jlab_core.abd83dd7cfacd6da4d00.js?v=abd83dd7cfacd6da4d00:1:238328
    at 1168.8e5df3139aa020a611cb.js?v=8e5df3139aa020a611cb:1:551
    at q._onSharedModelChanged (jlab_core.abd83dd7cfacd6da4d00.js?v=abd83dd7cfacd6da4d00:1:238119)
    at m (jlab_core.abd83dd7cfacd6da4d00.js?v=abd83dd7cfacd6da4d00:1:1838868)
    at Object.l [as emit] (jlab_core.abd83dd7cfacd6da4d00.js?v=abd83dd7cfacd6da4d00:1:1838544)
    at a.emit (jlab_core.abd83dd7cfacd6da4d00.js?v=abd83dd7cfacd6da4d00:1:1836381)
    at Array._modelObserver (1168.8e5df3139aa020a611cb.js?v=8e5df3139aa020a611cb:1:7524)
    at Object.r (383.db345dbeef5ef774e50c.js?v=db345dbeef5ef774e50c:1:936)
    at ji (383.db345dbeef5ef774e50c.js?v=db345dbeef5ef774e50c:1:43499)
    at Array.<anonymous> (383.db345dbeef5ef774e50c.js?v=db345dbeef5ef774e50c:1:52559)
```

It was working with JupyterLab 4.2. I am not sure which package is breaking the API contract yet.

krassowski:
Heads-up, it looks like this PR introduced two regressions; these are not visible in JupyterLab without extensions but are showstoppers for collaboration and server-side execution. I opened issues to track these:
- https://github.com/jupyterlab/jupyterlab/issues/16670
- https://github.com/datalayer/jupyter-server-nbmodel/issues/26

packages/cells/src/model.ts
krassowski: This is still happening on `main` after https://github.com/jupyterlab/jupyterlab/pull/16692 was merged

davidbrochart: I see the latest change there was one month ago, and https://github.com/jupyterlab/jupyterlab/pull/16498 was merged two weeks ago, so isn't that expected?

krassowski: I think it is not expected in the sense that this looks like a breaking API requirement change. This is previously it accepted JSON and now it requires an object with `toJSON()` method? I am trying to understand this really. I tested locally that a simple change of `'toJSON' in output ? output.toJSON() : output` makes it work.

davidbrochart: Previously we had:
```js
    if (change.outputsChange) {
      globalModelDBMutex(() => {
        this.outputs.clear();
        slot.getOutputs().forEach(output => this._outputs.add(output));
      });
    }
```
So any change to the outputs resulted in clearing them and adding the updated outputs again, which was very inefficient.
Now we have:
```js
    if (change.outputsChange) {
      globalModelDBMutex(() => {
        let retain = 0;
        for (const outputsChange of change.outputsChange!) {
          if ('retain' in outputsChange) {
            retain += outputsChange.retain!;
          }
          if ('delete' in outputsChange) {
            for (let i = 0; i < outputsChange.delete!; i++) {
              this._outputs.remove(retain);
            }
          }
          if ('insert' in outputsChange) {
            // Inserting an output always results in appending it.
            for (const output of outputsChange.insert!) {
              this._outputs.add(output.toJSON());
            }
          }
        }
      });
    }
```
So we really interpret the changes and do things in a finer grained manner.
`this._outputs.add` accepts a JSON, just that before we were converting all the shared outputs to JSON in `slot.getOutputs()` and now we do it only for the inserted outputs.

PR #16450, opened 2024-06-06
Align extension migration docs with the latest extension template

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16414

There is another place in the docs where the `copier` dependencies are mentioned: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#upgrading-extension-using-the-upgrade-script

## Code changes

Extension migration docs update.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #16503, opened 2024-06-19
Backport PR #16450 on branch 4.2.x (Align extension migration docs with the latest extension template)

krassowski:
Backport PR https://github.com/jupyterlab/jupyterlab/pull/16450 on branch 4.2.x (Align extension migration docs with the latest extension template)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,7 +125,7 @@ upgrade-extension = [
     "pyyaml-include<3.0",
     "copier>=9,<10",
     "jinja2-time<0.3",
-    "pydantic<2.0",
+    "pydantic<3.0",
     "tomli-w<2.0"
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,8 +122,8 @@ dev = [
     "ruff==0.3.5",
 ]
 upgrade-extension = [
-    "pyyaml-include<2.0",
-    "copier>=8,<10",
+    "pyyaml-include<3.0",
+    "copier>=9,<10",
     "jinja2-time<0.3",
     "pydantic<2.0",
     "tomli-w<2.0"
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -189,7 +189,7 @@ Or with ``conda``:
 
 .. code:: bash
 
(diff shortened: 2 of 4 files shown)


krassowski:
So the failure here is interesting case of CI pulling both the released and dev version. We may want to merge this and do an immediate patch release to fix CI:

```
ERROR: Cannot install jupyterlab and jupyterlab[upgrade-extension]==4.2.2 because these package versions have conflicting dependencies.

The conflict is caused by:
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
    jupyterlab[upgrade-extension] 4.2.2 depends on pydantic<2.0; extra == "upgrade-extension"
    copier 9.2.0 depends on pydantic>=2.4.2
    jupyterlab[upgrade-extension] 4.2.2 depends on pydantic<2.0; extra == "upgrade-extension"
    copier 9.1.1 depends on pydantic>=2.4.2
    jupyterlab[upgrade-extension] 4.2.2 depends on pydantic<2.0; extra == "upgrade-extension"
    copier 9.1.0 depends on pydantic>=2.4.2
    jupyterlab[upgrade-extension] 4.2.2 depends on pydantic<2.0; extra == "upgrade-extension"
    copier 9.0.1 depends on pydantic>=2.4.2
```

pyproject.toml
```diff
@@ -122,8 +122,8 @@ dev = [
     "ruff==0.3.5",
 ]
 upgrade-extension = [
-    "pyyaml-include<2.0",
-    "copier>=8,<10",
+    "pyyaml-include<3.0",
+    "copier>=9,<10",
     "jinja2-time<0.3",
     "pydantic<2.0",
     "tomli-w<2.0"
```
fcollonval, line 129: Should removing those two be enough or just changing the pydantic constraint? There are no failure on main due to `pydantic<3.0`.

```suggestion
```

krassowski: I think it is safer to keep a pin for dependencies which do know that may break things when they release a new version; let's just bump pydantic constraint to match with the `main`:

```suggestion
    "pydantic<3.0",
    "tomli-w<2.0"
```

Issue #16500, opened 2024-06-19
Configure is not called for inline completion providers if no documents are open


krassowski:
## Description

`IInlineCompletionProvider.configure()` is not called for providers if no notebook is open. This means that initial initialization does not take place after settings are loaded.

## Reproduce

1. Launch https://github.com/krassowski/jupyterlab-transformers-completer
2. See that the model download does not automatically
3. Open a notebook
4. Refresh browser tab
5. See that models are now downloading automatically

https://github.com/jupyterlab/jupyterlab/blob/ae123cb82015540804964f76ba4e56f21799ea27/packages/completer/src/manager.ts#L267-L284

## Expected behavior

Should immediately call `configure()` as needed. In jupyterlab-transformers-completer it would manifest by the download starting with no notebooks/files open:

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/a834459f-39ef-42fc-9fee-d660d97051be)

## Context

- JupyterLab version: 4.2.2

PR #16508, opened 2024-06-19
Fix inline completer `configure` calls not being propagated correctly

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/16500

## Code changes

- [x] Adds a test case for the `IInlineCompleterActions.configure()` invoking the `configure()` of `IInlineCompletionProvider`s
- [x] Fixes the issue by moving the logic up one level

## User-facing changes

Inline completer works more reliably, especially if JupyterLab is loaded with empty workspace

None


jtpio (approved):
Thanks!

jtpio:

PR #16520, opened 2024-06-24
Improve handling of shared output changes

davidbrochart:
## Code changes

The way shared output changes are handled currently is not optimal: any change in the Y will results in removing all previous outputs from the modelDB model and adding all new outputs. This is because we were only supporting the initialization from a file, but other cases are appearing, for instance in server-side execution an output can be added. In this case we don't want to clear all outputs, but just handle the added output.

## User-facing changes

None

None.


davidbrochart:
Closing, I'll make it part of #16498.

Issue #15871, opened 2024-02-25
Settings editor does not load tabs of some plugins on first open


krassowski:
## Description

When "Settings" Editor is opened before the loading/transformation of settings completed, the tabs for packages which were not ready are not shown. Closing and re-opening the "Settings" Editor after the settings for remaining packages got loaded helps.

This especially affects `Code Completion` and `Inline Completer` settings (both using `jupyter.lab.transform`), but it was also reported for extensions which do not use `jupyter.lab.transform` but are just loaded later, e.g.:
- https://github.com/deshaw/jupyterlab-execute-time/issues/98

Another report for this issue was in: https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1053

## Reproduce

1. Open "Settings Editor"
2. Take a note of the list of "SETTINGS"
3. Refresh page (so that JupyterLab loads with "Settings Editor" already open)
4. Compare the list of "SETTINGS" now with the one from step (2)
5. See that a number of tabs is now missing

| On first open | After reopening |
|--|--|
|  ![Screenshot from 2024-02-25 17-04-28](https://github.com/jupyterlab/jupyterlab/assets/5832902/427d810e-ea8a-4ba1-9b4d-67f67bd6a79e) | ![Screenshot from 2024-02-25 17-04-08](https://github.com/jupyterlab/jupyterlab/assets/5832902/fc347988-e710-4974-bbe5-7906ba7c3a51) |

## Expected behavior

Settings for all plugins are are shown regardless of when they get loaded/activated/transformed.

There are two options:
- (a) the "Settings" widget shows with whatever is ready and additional tabs show up as new plugins finish loading or as transformation of settings gets completed
- (b) the "Settings" widget only opens once all plugins are ready

I think we should go for (a) because (b) might lead to the "Settings" widget not showing up in a reasonable time at all.

## Context

- JupyterLab version: 4.1.2

PR #16523, opened 2024-06-25
Fix settings editor missing plugins with transform step or registered late

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/15871

## Code changes

- removes bad code duplication between `PluginList` and `SettingsEditor` which lead to loading the settings for each plugin twice (this leads to a minimal performance improvement because the fetched plugins were already cached before)
- adds a dedicated `PluginList.Model` model class which is used by both `PluginList` and `SettingsEditor`
- deprecates `PluginList.sortPlugins()` - this function was only used to implement the aforementioned duplicated logic; an equivalent function is now one of the private model methods
- deprecates `toSkip` option previously passed to the `PluginList` constructor which should now be supplied to the `PluginList.Model` instead
- deprecates `PluginList.registry` attribute - the plugin list should not provide the settings registry as it is not its responsibility; private `_registry` name is to be used going forward

## User-facing changes

- When the plugin gets added after the Settings Editor was opened this plugin is now visible on the list
- When settings schema gets populated after the Settings Editor was opened (due to the use of `transform`), the settings are now correctly shown in the editor

None

### DIFF
--- a/packages/settingeditor/src/pluginlist.tsx
+++ b/packages/settingeditor/src/pluginlist.tsx
@@ -55,9 +55,13 @@ export class PluginList extends ReactWidget {
     this.addClass('jp-PluginList');
     this._confirm = options.confirm;
     this._model = options.model ?? new PluginList.Model(options);
-    this._model.ready.then(() => {
-      this.update();
-    });
+    this._model.ready
+      .then(() => {
+        this.update();
+      })
+      .catch(reason => {
+        console.error(`Failed to load the plugin list model:\n${reason}`);
+      });
     this._model.changed.connect(() => {
       this.update();
     });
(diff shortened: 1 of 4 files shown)


packages/settingeditor/src/pluginlist.tsx
```diff
@@ -50,13 +50,21 @@ export class PluginList extends ReactWidget {
...
-    this.registry.pluginChanged.connect(() => {
+    this._model = options.model ?? new PluginList.Model(options);
+    this._model.ready
+      .then(() => {
+        this.update();
+      })
+      .catch(reason => {
+        console.error(`Failed to load the plugin list model:\n${reason}`);
+      });
+    this._model.changed.connect(() => {
```
fcollonval, line 65: Should we connect this signal only after `this._model.ready` to ensure we don't trigger update when the model is not yet ready?

fcollonval (changes requested):
Thanks @krassowski 

i have a minor suggestion and I spotted a regression. In the settings editor, the plugin list does not update when its status changes from all default values to some custom values:

bug_default_not_default

fcollonval (approved):
Thanks @krassowski

PR #16531, opened 2024-06-26
Enhance readability by adjusting sizes of the headings

Mochatr:


fcollonval:
Thanks @Mochatr could you please provide more background on this PR by fulfilling the template?

krassowski:
@Mochatr if you are no longer interested in pursuing this, should we close it? Since you provided no info, I can only suspect that you are trying to address a problem with styling by changing the document structure. I would prefer if we instead adjusted styles to make the lower level headings more prominent. Would you be happy to rework this PR or open a new one following such an approach?

krassowski:
Closing due lack of response and a merge conflict. If you would like to revive this work please reopen this PR or open a new one - thank you!

PR #16507, opened 2024-06-19
Add customisation options to prevent inline completer resizing aggressively

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/16364
- Depends on https://github.com/jupyterlab/jupyterlab/pull/16493

## Code changes

Adds new completion inline settings :
- `minLines`
- `maxLines`
- `reserveSpaceForLongest`
- `spacerRemovalDelay`

## User-facing changes

None by default, but:
- `maxLines` adds a collapser for preview of remaining lines
  collapser
- `minLines` adds additional empty lines
- `reserveSpaceForLongest`: adds additional empty lines to match the space of the longest completion

## Backwards-incompatible changes

None

PR #16552, opened 2024-07-03
Backport PR #16507 on branch 4.2.x (Add customisation options to prevent inline completer resizing aggressively)

krassowski:
Backport PR https://github.com/jupyterlab/jupyterlab/pull/16507 on branch 4.2.x (Add customisation options to prevent inline completer resizing aggressively)


krassowski:
`Check internationalization changes / check_i18n (pull_request)` fails because this PR backports new translatable strings for the new options; this may render some UI without translated strings but I think it is ok given that aggressive resizing being addressed by the newly added option is a blocker for using the inline completer for many users.

PR #16556, opened 2024-07-07
Service manager as plugin

fcollonval:
## References

Closes #16141


fcollonval:
@krassowski I would like to check that _locking_ the new service manager plugin would prevent third-party extensions to override the core default provider. Is it possible with the current lock feature? Does this needs specific tests? How would you suggest to test that?

krassowski:
>  Is it possible with the current lock feature? 

I think it should be possible, but may require some custom logic. I am not up to speed on this PR yet.

> How would you suggest to test that?

There are tests for locking in the `usage` tests group. I think you could write an extension that swaps a service plugin and as a side effect hits a server endpoint to print something out. Then you could test if this "something" is printed out with/without lock on.

krassowski:
@fcollonval do you think this could still be ready in time for 4.3.0 feature freeze?

fcollonval:
> @fcollonval do you think this could still be ready in time for 4.3.0 feature freeze?

I don't think so because I get some pressure to change the approach for something more composable to avoid extension compatibilities.

krassowski:
Is this related to https://github.com/jupyterlab/lumino/pull/705?

fcollonval:
> Is this related to [jupyterlab/lumino#705](https://github.com/jupyterlab/lumino/pull/705)?

Not really - they are different parts of the same goal _Make the `Application` fully customizable_.

packages/application-extension/src/services.ts
```diff
@@ -0,0 +1,34 @@
...
+  IDefaultDrive,
+  IServerSettings,
+  IServiceManager,
+  ServerConnection,
+  ServiceManager
+} from '@jupyterlab/services';
+import type { IPlugin } from '@lumino/coreutils';
+
+export const services: IPlugin<null, ServiceManager.IManager> = {
+  id: '@jupyterlab/application-extension:services',
```
jtpio, line 16: Wondering if this plugin should be named `@jupyterlab/application-extension:service-manager` so it matches a bit more what it provides (`IServiceManager`)?

jtpio:
@fcollonval curious about the reasons for closing this PR?

It looks like would still be useful and relevant. Happy to revive it if you don't have time to work on it anymore.

jtpio:
> provide registries for the services to allow greater flexibility without reducing extensions compatibility.

Right, in the end we seem to always come back to the fact that the `ServiceManager` is expected to be fully initialized and ready on plugin activation, and due to the fact that it was designed around the expectation of a full Jupyter Server in the backend, makes it difficult to customize its underlying services easily.

Having registries could help with this, and avoid having to introduce breaking changes or changes in behavior. But the idea would likely have to be applied to all underlying services, such as the kernels manager or even the nbconvert handler, and would mean introducing many new APIs.

Thinking more about this, maybe it could be useful to have two phases for registering plugins, splitting more `ServiceManager` functionalities in more plugins, and still registering all plugins on the same app:
- one when the service manager is initialized and its plugins are resolved. The plugins would get `null` for their `app` parameter
- the other ones being registered as normal, just like the current implementation
 
If I recall correctly, this was discussed during this weekly meeting (recording is on YouTube): https://github.com/jupyterlab/frontends-team-compass/issues/229#issuecomment-2048074876. But that might require adding a way to "discover" these plugins that should be activated before the other "normal" plugins.

Issue #16065, opened 2024-03-27
Consider adding additional shortcut for accepting inline completions


krassowski:
### Problem

Currently inline completions are accepted using <kbd>Alt</kbd> + <kbd>End</kbd>.

Problems:
- this shortcut is not very convenient to use (<kbd>Alt</kbd> and <kbd>End</kbd> are far away on most keyboards)
- other applications use a different shortcut

Design constraints:
-  using just <kbd>End</kbd> or just <kbd>→</kbd> is not desirable as these would interfere with the native cursor movement behaviour
- ideally there would be a natural variant of the shortcut for partial acceptance (https://github.com/jupyterlab/jupyterlab/issues/15531); for example if the "accept all" is <kbd>Ctrl</kbd>+<kbd>→</kbd>, we could use <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>→</kbd> for word-by-word acceptance (there are also line-by-line and character-by-character options)
- the shortcut should be intuitive to the users coming from other applications:
   - fish uses:
      - <kbd>End</kbd> or <kbd>→</kbd> for full acceptance
      - <kbd>Alt</kbd>+<kbd>→</kbd> or <kbd>Alt</kbd>+<kbd>F</kbd> for word-by-word
   - IPython uses:
       - <kbd>End</kbd> or <kbd>Ctrl</kbd>+<kbd>F</kbd> for full acceptance
       - <kbd>Alt</kbd>+<kbd>F</kbd> for word-by-word
       - <kbd>Ctrl</kbd>+<kbd>→</kbd> for token-by-token (word-by-word but with a better implementation)
       - <kbd>Alt</kbd>+<kbd>→</kbd> for character-by-character
    - Copilot uses:
       - <kbd>Tab</kbd> for full acceptance
       - <kbd>Ctrl</kbd>+<kbd>→</kbd> for word-by-word
- using <kbd>Tab</kbd> might be surprising because we use it to invoke completer
- the shortcut should ideally not interfere with any browser or OS shortcuts

### Proposed Solution

Add a default shortcut other than <kbd>Alt</kbd> + <kbd>End</kbd> for accepting the inline suggestion.

### Additional context

We can keep <kbd>Alt</kbd> + <kbd>End</kbd> to avoid breaking the experience of users that got used to it (we can have multiple shortcuts here).

JasonWeill:
[...]

PR #16565, opened 2024-07-09
Add `Tab` as an alternative shortcut for accepting inline completion

andrewfulton9:
## References
#16065 

## Code changes
Changes inline completion shortcut from "Alt End" to "Tab" and completer invocations from "Tab" to "Ctrl ArrowDown". I am open to other ideas for the shortcuts, especially the completer invocations. I considered "Alt Tab" and "Ctrl Tab" in place, but these both conflict with system and browser shortcuts, at least on linux.

## User-facing changes
New shortcuts for inline completer

None that I am aware.


krassowski:
I think that changing the default shortcut for tab completer would be a breaking change. Do we need to do that?

andrewfulton9:
> I think that changing the default shortcut for tab completer would be a breaking change. Do we need to do that?

@krassowski, If I don't change that then both the inline completer and the dropdown completer use Tab, and at least in my experiments the inline completer takes precedent making the dropdown completer unusable. I could have missed something though so I am open to ideas.

krassowski:
>  If I don't change that then both the inline completer and the dropdown completer use Tab, and at least in my experiments the inline completer takes precedent making the dropdown completer unusable

This should be fixed by https://github.com/jupyterlab/jupyterlab/pull/16613. I tested it using tab. Let's test again here after we have that one merged.

krassowski:
Since https://github.com/jupyterlab/jupyterlab/pull/16613 is in, @andrewfulton9 what do you think about making another attempt, but instead of replacing the shortcut for inline completion acceptance, adding a new one and ensuring that it gets shown in the hover tooltip? (along reverting the change to tab completer)?

andrewfulton9:
> instead of replacing the shortcut for inline completion acceptance, adding a new one and ensuring that it gets shown in the hover tooltip?

@krassowski,  Can you help me understand what you mean here? Do you mean that "Accept Inline Completion" should stay as "Alt End" and there should be a new shortcut for inline completion acceptance that should change to "Tab"? I've gone ahead and reverted the change to tab completer so that should be good to go. I think everything here is working as expected with changing "Alt End" to "Tab" for the inline-completer with just changing the shortcut.

krassowski:
> Do you mean that "Accept Inline Completion" should stay as "Alt End" and there should be a new shortcut for inline completion acceptance that should change to "Tab"

Yes, because changing the default shortcut is a change which breaks experience for existing users. Adding a new shortcut is not a breaking change. We can have multiple shortcuts for the same action, but only one of them can be displayed in the tooltip - I think we would want "Tab" displayed.

andrewfulton9:
> only one of them can be displayed in the tooltip - I think we would want "Tab" displayed.

Do you mean for UX? It looks like the tooltip can show two shortcuts for a command if we want both there.

image

krassowski:
Yes this is what I mean. I think one _blessed_ shortcut would be better, but happy to hear other opinions.

krassowski:
I updated the title to reflect the current state. If this is to be merged in 4.3.0 we would need to get this ready early next week. The argument for merging it into 4.3.0 (late in the beta cycle) is that it will allow accepting inline completions on Mac (https://github.com/jupyterlab/jupyterlab/issues/16662) so it is a bug fix as much as an enhancement.

andrewfulton9:
> I updated the title to reflect the current state. If this is to be merged in 4.3.0 we would need to get this ready early next week. The argument for merging it into 4.3.0 (late in the beta cycle) is that it will allow accepting inline completions on Mac (#16662) so it is a bug fix as much as an enhancement.

Sounds good. Everything is working locally for me. I just need to fix the UI tests. @krassowski, If you have time, can you give it a try and see if it is working as expected for you?

andrewfulton9:

krassowski (approved):
Thank you @andrewfulton9! Tested that:
- accepting with `Tab` works
- accepting with `Alt + End` still works
- `Tab` navigation is not prevented when cell is blurred with `Esc`

The failing test is unrelated, tracked in https://github.com/jupyterlab/jupyterlab/issues/16722.

PR #16566, opened 2024-07-09
Standardize Property Inspector placeholder according to TOC

### LINKED ISSUE #14235: Placeholder for property inspector should be consistent with TOC
## Description

Follow up to #13988. See also #14234.

The new TOC placeholder has a headline and explanation in gray text near the top of the pane. The property inspector displays a placeholder in black text, centered vertically, with no headline.

image

## Reproduce

1. Create a new Markdown file.
2. Open property inspector.

## Expected behavior

Property inspector placeholder uses similar styling to TOC placeholder, including alignment and sizing of text.

joaopalmeiro:
Hi! 👋 

## References

Closes #14235 

## Code changes

The Property Inspector placeholder styling was adapted based on the [Table of Contents placeholder](https://github.com/jupyterlab/jupyterlab/blob/c4490110093280be3a2859749fb6b5675df3c0f2/packages/toc/style/base.css#L23-L30) as well as the [HTML elements](https://github.com/jupyterlab/jupyterlab/blob/c4490110093280be3a2859749fb6b5675df3c0f2/packages/toc/src/treeview.tsx#L30-L40) that make up it to take advantage of `<p>` styles and for consistency. 

Let me know your feedback, please!

## User-facing changes

The Property Inspector placeholder is now visually and element-wise similar to the Table of Contents placeholder:

Screenshot 2024-07-09 at 18 25 33

None.

### DIFF
--- a/packages/property-inspector/src/index.ts
+++ b/packages/property-inspector/src/index.ts
@@ -186,8 +186,10 @@ export class SideBarPropertyInspectorProvider extends PropertyInspectorProvider 
     } else {
       const node = document.createElement('div');
       const content = document.createElement('div');
-      content.textContent = this._trans.__('No properties to inspect.');
+      const placeholderText = document.createElement('p');
+      placeholderText.textContent = this._trans.__('No properties to inspect.');
       content.className = 'jp-PropertyInspector-placeholderContent';
+      content.appendChild(placeholderText);
       node.appendChild(content);
       this._placeholder = new Widget({ node });
       this._placeholder.addClass('jp-PropertyInspector-placeholder');
--- a/packages/property-inspector/style/base.css
+++ b/packages/property-inspector/style/base.css
@@ -14,11 +14,10 @@
 }
 
 .jp-PropertyInspector-placeholder {
-  display: flex;
-  align-items: center;
-  justify-content: center;
+  text-align: center;
 }
 
 .jp-PropertyInspector-placeholderContent {
+  color: var(--jp-content-font-color2);
   padding: 8px;
 }


krassowski:
I would go ahead and change the text to something more useful and following the convention as in TOC, e.g.:

> **No properties**
> Property inspector allows to view and edit properties of a selected notebook.

_Unless_ you would like this PR to be backported to 4.2.x as-is and then make a new PR with the updated text (to avoid breaking experience of users who use language packs we do not change translatable strings in the main parts of the UI in backports).

joaopalmeiro:
I can implement this change too, no problem, @krassowski!

Personally, I have no need for this PR to be backported to JupyterLab 4.2 (for context, we are working with JupyterLab 4.1 and the plan is to upgrade to 4.3 when possible).

Let me know if I can make the change you suggested or if you would prefer to proceed another way.

Btw, after changing a translatable string, do I need to open a PR on the https://github.com/jupyterlab/language-packs side?

krassowski:
> Let me know if I can make the change you suggested or if you would prefer to proceed another way.

Sure, let's target 4.3 with this PR and add the text changes here.

No, no need to open PR on language packs, it is all automated.

joaopalmeiro:
@krassowski, the PR is ready for a new review!

I changed the text you suggested implementing a bit to, from my perspective, make it more consistent with the Table of Contents one, that is, the header in _Title Case_ and the text starting with "The":

Screenshot 2024-07-10 at 11 46 23

Let me know your thoughts, please!

fcollonval (approved):
Thanks @joaopalmeiro and @krassowski

PR #16228, opened 2024-04-23
Show kernel ID for notebooks and consoles in list view

gderocher:
## References
Fixes #15991 

## Code changes
Added name() function to return the name of children (previously done by label). 
Reimplemented label function to display kernel ID through a ReactNode. 

## User-facing changes
Kernel ID displays for all notebooks and consoles in both list and tree view. 
Planning on updating so the ID only shows in list view. 

After Changes: 
image
image

PR #16571, opened 2024-07-10
added name and changed and reimplemented label

### LINKED ISSUE #15991: Display kernel ID in the list view of kernels in the sidebar
> about the table view; should we display the kernel id in a second column (a bit like last modified in the file browser)? Otherwise the kernels section does not bring lots of information compare to the open tabs.

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/15845#pullrequestreview-1933776385_
            

Quick first mockup (probably should be aligned more to the right and maybe does not need the parentheses, but just to give the idea what this is about):

mockup

rritik772:
## References

Fixes #15991

## Code changes

Reimplemented the way name and kernel-id shows to section and list view by changing return type for label and added new var name which return the name.
Implemented label to show kernel-id and name through ReactNode

## User-facing changes
section view
hello

list view
hello

### DIFF
--- a/packages/running-extension/src/kernels.tsx
+++ b/packages/running-extension/src/kernels.tsx
@@ -239,7 +239,7 @@ namespace Private {
   type DocumentWidgetWithKernelItem = Omit<
     IRunningSessions.IRunningItem,
     'label'
-  > & { label(): string };
+  > & { label(): ReactNode; name(): string };
 
   export class RunningKernel implements IRunningSessions.IRunningItem {
     constructor(options: RunningKernel.IOptions) {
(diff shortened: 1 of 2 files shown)


krassowski:
Thanks @rritik772, are there any changes compared to #16228? Can you list them and add `Co-authored-by:` with @gderocher's details (https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)?

krassowski:
Closing due to lack of response and because this is practically identical to https://github.com/jupyterlab/jupyterlab/pull/16228.

PR #16572, opened 2024-07-10
Standardize Contextual Help placeholder according to TOC

### LINKED ISSUE #14234: Placeholder for contextual help should be consistent with TOC
## Description

Follow up to #13988.

The new TOC placeholder has a headline and explanation in gray text near the top of the pane. The contextual help section displays a placeholder in a large, italicized, left-aligned font:

image

## Reproduce

1. Create a new blank notebook.
2. Open contextual help.

## Expected behavior

Contextual help placeholder uses similar styling to TOC placeholder, including alignment and sizing of text.

joaopalmeiro:
Hi! 👋 

## References

Closes #14234

## Code changes

The Contextual Help placeholder styles have been adjusted to match those of the Table of Contents and the Property Inspector. Additionally, the class name has been updated to include the "placeholder" keyword instead of "default" for consistency.

Let me know if I should update the placeholder string or not, please.

## User-facing changes

The Contextual Help placeholder is now visually similar to the Table of Contents and Property Inspector placeholders:

Screenshot 2024-07-10 at 19 24 02

None.

### DIFF
--- a/packages/inspector/style/base.css
+++ b/packages/inspector/style/base.css
@@ -42,3 +42,7 @@
   text-align: center;
   color: var(--jp-content-font-color2);
 }
+
+.jp-Inspector-placeholderContent > h3 {
+  margin-bottom: var(--jp-content-heading-margin-bottom);
+}
--- a/packages/inspector/style/base.css
+++ b/packages/inspector/style/base.css
@@ -38,11 +38,7 @@
   white-space: pre-wrap;
 }
 
-.jp-Inspector-default-content {
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  font-size: xx-large;
-  font-style: italic;
-  color: darkgray;
+.jp-Inspector-placeholderContent {
+  text-align: center;
+  color: var(--jp-content-font-color2);
 }
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -21,6 +21,8 @@ anymore. The side effects for extensions are:
   prepend the class ``.jp-ThemedContainer`` to your rule; e.g.
   ``.jp-Inspector-content pre`` becomes ``.jp-ThemedContainer .jp-Inspector-content pre``
 
+The ``jp-Inspector-default-content`` class was renamed to ``jp-Inspector-placeholderContent``.
+The name of this contextual help class is now consistent with the equivalent table of contents and property inspector classes.
 
 JupyterLab 4.1 to 4.2
 ---------------------
--- a/packages/inspector/src/inspector.ts
+++ b/packages/inspector/src/inspector.ts
@@ -23,7 +23,7 @@ const CONTENT_CLASS = 'jp-Inspector-content';
 /**
  * The class name added to default inspector content.
  */
-const DEFAULT_CONTENT_CLASS = 'jp-Inspector-default-content';
+const DEFAULT_CONTENT_CLASS = 'jp-Inspector-placeholderContent';
 
 /**
  * A panel which contains a set of inspectors.
--- a/packages/inspector/src/inspector.ts
+++ b/packages/inspector/src/inspector.ts
@@ -51,7 +51,7 @@ export class InspectorPanel
         'No Documentation'
       )}</h3>`;
       const placeholderText = `<p>${this._trans.__(
-        'Select a code fragment (e.g. function or object) to request information about it from the kernel attached to the editor.'
+        'Move the cursor to a code fragment (e.g. function or object) to request information about it from the kernel attached to the editor.'
       )}</p>`;
 
       this._content = InspectorPanel._generateContentWidget(
--- a/packages/inspector/src/inspector.ts
+++ b/packages/inspector/src/inspector.ts
@@ -47,10 +47,15 @@ export class InspectorPanel
         `<p>${options.initialContent}</p>`
       );
     } else {
+      const placeholderHeadline = `<h3>${this._trans.__(
+        'No Documentation'
+      )}</h3>`;
+      const placeholderText = `<p>${this._trans.__(
+        'Select a code fragment (e.g. function or object) to request information about it from the kernel attached to the editor.'
+      )}</p>`;
+
       this._content = InspectorPanel._generateContentWidget(
-        '<p>' +
-          this._trans.__('Click on a function to see documentation.') +
-          '</p>'
+        `${placeholderHeadline}${placeholderText}`
       );
     }
 


packages/inspector/style/base.css
```diff
@@ -38,11 +38,7 @@
...
 }
 
-.jp-Inspector-default-content {
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  font-size: xx-large;
-  font-style: italic;
-  color: darkgray;
+.jp-Inspector-placeholderContent {
```
krassowski, line 41: Can you mention the class name change in the [Extension Migration Guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#css-styling)?

krassowski:
> Let me know if I should update the placeholder string or not, please.

Not strong opinion on this one. @JasonWeill?

JasonWeill:
How about a heading in Title Case (level to match that used in TOC) and longer explanatory text in sentence case?

> ## No Function Selected
> 
> Click on a function in a notebook or source code file to see documentation about it.

krassowski:
My problem with the current use of the word _function_ is that it is often inaccurate as it will also work on objects and other things. I do not have a better proposal though (_object_ would be less intuitive). But if we are using more of the available space maybe we can list a few examples?

_Notebook or source file_ is ok-ish, though on source file it only works when they have a kernel attached. Maybe "an editor with a kernel attached"? Then notebook is always easiest to understand and covers 90% of use case.

JasonWeill:
How about: "In an editor with a kernel attached, click on a function or object to see documentation about it."

joaopalmeiro:
For the title, I thought of two suggestions:

- Following the idea of ​​the Table of Contents one ("No Headings"): "No Documentation" or "No Help Available"
- Something more generic like "No Code Selected"

As for the description, the [description of Python's built-in `help()` function](https://docs.python.org/3/library/functions.html#help) talks about _any other kind of object_. Based on your ideas, why not change @JasonWeill's suggestion from _a function or object_ to _a class, function, or any other kind of object_?

packages/inspector/style/base.css
joaopalmeiro: Added! What do you think?

Screenshot 2024-07-11 at 08 10 35

krassowski:
More thoughts:
- it is also used by non-Python kernels, for example IRKernel; while most languages have a concept of function (or equivalent), not all languages will have a "class" or "module"; I think all languages have "variables"
- in protocol this is called `inspect_request` and it operates on "code" and cursor position
- the returned result can include much more than just docstrings (signatures, file location, even contents of a data frame) so "documentation" is an ok word to use but not exhaustive
- the panel already includes the name "contextual help" on the tab bar
- given the above maybe we can avoid repeating "help" word and instead opt in to highlight that this works on "code" or performs "inspection" (while trying to avoid confusion with property inspector).

I kind of like "No Code Selected" or "No Function Selected" as it indirectly tells user what action is needed to get something to show up, but it can be confusing when some code is selected but the notebook is just not focused. Even worse, if the kernel does not have anything to return (or does not implement `inspect_request`) it will be straight up frustrating as in the example below (ipykernel does not know about my `test()` function because I did not execute the code yet):

image

While I slightly worry that using "No Documentation" or "No Help Available" may make the user think that something is wrong or missing, it appears indeed a bit better in the sense it will not frustrate user when things do not work.

> ### No Documentation
> 
> Select a code fragment (e.g. function or object) to request information about it from the kernel attached to the editor.

The above:
- is generic enough that it should be valid for any kernel while not focusing on a single type of an object (function/variable)
- includes a call to action
- explains that a kernel request will be made - this will also hint to the user that kernel may not be able to reply soon enough if it is too busy executing code
- indirectly explains that if user selects code and there is no result, it is likely to do with kernel not having the `inspect_request` handling or not being able to reply for this particular code fragment
- avoids using "see" which is frustrating for users who cannot see (accessibility)
- avoids using "click" for users who cannot operate a mouse (accessibility)

<details>

While ToC/Property Inspector are docked in the sidebar, the Contextual Help is a detachable panel (and in Notebook 7 the plan is to have it show up at the bottom as in Notebook 6) so it will have more space for a bit longer heading/description - I think it is ok to use more than two/three words (in contrast to ToC/Property Inspector where space is at premium).


Here is how GNOME Text Editor handles empty state:

image

</details>

joaopalmeiro:
Amazing, thank you so much for sharing your thought process, @krassowski!

Based on your suggestion, and given that we have space, I'm missing some information that would allow a user to know more about whether the kernel they are using implements this functionality. 

I tried searching for documentation on the [IPython kernel](https://github.com/ipython/ipykernel/blob/d4a8703593cbe5d82bc955d68543dc623f5b4510/ipykernel/kernelbase.py#L744), [IRkernel](https://github.com/IRkernel/IRkernel/blob/124f2347f15cfadeabc6738868e05e6087f4c456/R/kernel.r#L247) ,and [Deno kernel](https://github.com/denoland/deno/blob/de30f7c1884d8392dd956dce0d0627d5f94562e3/cli/tools/jupyter/server.rs#L364) websites/repos, but I only found information going straight to the code. However, there's some documentation in [jupyter_client](https://jupyter-client.readthedocs.io/en/stable/messaging.html#introspection). 

So, assuming I'm getting the messaging right, would a sentence make sense to help the user learn more about whether the kernel in use has this feature or not? Something like: "Check if your kernel addresses [introspection messaging](https://jupyter-client.readthedocs.io/en/latest/messaging.html#introspection) if necessary."

krassowski:
I agree there is no single place to go to check this and the documentation of kernels often does not highlight this. What I think is often the case is that kernels either implement everything that was in older versions of the protocol or just the most basic functionality.

> Something like (I'm not sure about the right wording): "Check if your kernel addresses [introspection messaging](https://jupyter-client.readthedocs.io/en/latest/messaging.html#introspection) if necessary."

I do agree that a message like this could be useful. That said, embedding links to documentation in UI could bite us in the future if they change, and it will also be annoying to translate properly. I do not have a strong opinion on whether this should be added or not nor on the exact wording.

joaopalmeiro:
> That said, embedding links to documentation in UI could bite us in the future if they change, and it will also be annoying to translate properly. I do not have a strong opinion on whether this should be added or not nor on the exact wording.

That's a good point. Since your suggestion is already an improvement to what is currently available, I would say to not include my suggestion and keep it simpler. What do you think?

krassowski:
Yeah, I think we can merge it with either version and then iterate, especially if we receive any user feedback.

joaopalmeiro:
> Yeah, I think we can merge it with either version and then iterate, especially if we receive any user feedback.

I will update these changes, then. Thanks!

JasonWeill:
I'm personally not a fan of "No Documentation", as it implies a problem with JupyterLab (it cannot find or load the documentation) instead of with the user (you have not clicked anything that could be documented). "Nothing Selected" or "Nothing is Selected" might suggest the latter.

krassowski:
I agree, this is a tradeoff that I explained in https://github.com/jupyterlab/jupyterlab/pull/16572#issuecomment-2222319458 ("Nothing is Selected" might bring frustration in the example explained above).

joaopalmeiro:
I implemented the changes as per Miguel's suggestion. Let me know what you think, please!

Personally, I don't have strong opinions about the title because they both seem to make sense. However, after playing around with this feature a bit, I think that `Select` is not entirely clear. I think `Move the cursor to` would be more concrete about the action expected by the user and, therefore, the notebook/editor would be active to trigger Contextual Help (if I'm correct). What do you think?

krassowski:
>  I think that Select is not entirely clear. I think Move the cursor to would be more concrete about the action expected by the user and, therefore, the notebook/editor would be active to trigger Contextual Help (if I'm correct). What do you think?

I agree.

joaopalmeiro:
Feedback implemented, @krassowski:

Screenshot 2024-07-12 at 11 52 51

krassowski (approved):
Thank you @joaopalmeiro, LGTM! I will wait a bit until Jason has time to take another look, otherwise I will merge on Monday evening.

krassowski:
Merging, we can always iterate during the alpha/beta cycle.

PR #16598, opened 2024-07-17
Add setting to enable single-click navigation in file browser

RRosio:
## References

Adding a setting in file-browser extension to allow single-click navigation as per https://github.com/jupyterlab/jupyterlab/issues/14071.

## Code changes

- [x] add command to toggle single-click navigation
- [x] update logic to allow single-click navigation

## User-facing changes

Users can access the setting option by going to the Settings Editor -> File Browser and select the option `Navigate files and directories with single click` 

N/A


packages/filebrowser-extension/schema/browser.json
```diff
@@ -28,6 +28,15 @@
...
         ]
+      },
+      {
+        "id": "jp-mainmenu-view",
+        "items": [
+          {
+            "command": "filebrowser:toggle-single-click",
+            "rank": 9.99
+          }
+        ]
```
krassowski, line 39: Do we really want this in the main menu? I would say a setting would be enough.

krassowski:
I pushed a commit after running `jlpm lint` locally. Maybe you need to re-install the pre-commit hooks? What does `jlpm --version` return?

krassowski:
User-facing documentation would be a nice enhancement but would not block this PR

packages/filebrowser/src/listing.ts
```diff
@@ -1025,6 +1025,18 @@ export class DirListing extends Widget {
...
+   * This enables opening files/directories with a single click.
+   */
+  setAllowSingleClickNavigation(isEnabled: boolean) {
+    this._allowSingleClick = isEnabled;
+  }
+
+  getAllowSingleClickNavigation() {
+    return this._allowSingleClick;
+  }
+
```
krassowski, line 1039: I think we do not need it; let's keep the public API to minimum here.

```suggestion
```

packages/filebrowser/src/listing.ts
```diff
@@ -1138,6 +1150,10 @@ export class DirListing extends Widget {
       document.addEventListener('mouseup', this, true);
       document.addEventListener('mousemove', this, true);
     }
+
+    if (this.getAllowSingleClickNavigation()) {
```
krassowski, line 1154: Allows to reduce the API surface, and removes one function call :)

```suggestion
    if (this._allowSingleClick) {
```

krassowski (commented):
Some minor suggestions on namig

krassowski (approved):
Thank you @RRosio! Tested locally, works great. Selecting multiple files without opening them is still possible with <kbd>Ctrl</kbd> + :computer_mouse: click.

krassowski:
A great follow up here would be adding a test because otherwise this feature may break with any future refactoring.

jtpio:
> A great follow up here would be adding a test because otherwise this feature may break with any future refactoring.

Opened https://github.com/jupyterlab/jupyterlab/issues/16640 to track this.

jtpio (approved):
Thanks @RRosio @krassowski !

Issue #16358, opened 2024-05-22
Inline completer and tab completer are sometimes invoked simultaneously


krassowski:
## Description

It is possible to trigger both the inline completer and tab completer at the same time:

![Screenshot from 2024-05-22 12-38-32](https://github.com/jupyterlab/jupyterlab/assets/5832902/ba2249c2-b9be-49fb-912b-1a3d19493f30)

## Reproduce

1. Go "Settings" → "Code Completion" → check "Enable autocompletion'.
2. Go to "Settings" → "Inline Completer" → "History provider" → check "Enabled"
3. Try to complete a common prefix of history and kernel variables
4. See both show up

## Expected behavior

- one of the completers takes precedence
- (maybe) the precedence is user-configurable

## Context

- JupyterLab version: 4.2.0

PR #16613, opened 2024-07-25
Suppress the inline/tab completer if the other is already active

krassowski:
## References

Closes https://github.com/jupyterlab/jupyterlab/issues/16358

## Code changes

- When inline completion binding is set to <kbd>Tab</kbd> do not prevent event handling (=do not suppress the tab completer) if there is no inline completion binding; this is done by:
   - enhancing `isEnabled` check on the  `inline-completer:invoke` command
   - checking for `isEnabled()` in the special logic for handling <kbd>Tab</kbd> binding
- suppress invocation of inline completer if tab completer is open and `suppressIfTabCompleterActive` setting is on; this is implemented in two places:
   - mainly in handling of the inline completer responses, suppressing them if the tab completer box is detected
   - in the special logic for handling <kbd>Tab</kbd> binding (it needs to short-circuit to avoid preventing the default, so that the Tab toggling between suggestions is not prevented from an open tab completer)
- suppress invocation of tab completer if inline completions are shown and `suppressIfInlineCompleterActive` setting is on
- fixes `isEnabled` state propagation so that it is reflected on buttons correctly by using `app.notifyCommandChanged`

## User-facing changes

Adds two settings:
- `suppressIfTabCompleterActive` in the inline completer
- `suppressIfInlineCompleterActive` in the tab completer

Both are enabled by default. If users prefers to see completions from inline and tab complete at once they can untick the boxes.

By default only one will ever be active. Here is a recording with both tab-completer set to autocomplete and inline completer:

both-inline-and-tab

- First it shows that automatic tab completer is suppressed when inline completion suggestions are found.
- Then it shows that tab completer still works after dismissing inline completions by moving the cursor back and forth
- Finally it shows that automatic tab completions work when inline completer does not find any matches

Of course users may prefer it to work differently but that's why we have settings for it :)

None


jtpio (approved):
Thanks!

PR #16615, opened 2024-07-25
Copy and style edits for upgrade news notification

### LINKED ISSUE #16608: Improve upgrade notification styling
### Problem

The upgrade popup has a few usability problems:

1. The button labels are so long that they flow to a second line
2. The button labels are left-justified, not centered like other buttons in Jupyter are
3. The "Open changelog" button label is styled like a link, including an underline, even though it works like a button
4. The tooltip for "Do not check for updates" starts with "If pressed, you will not be prompted", which is verbose.

image

### Proposed Solution

1. Shorten "Do not check for updates" to "Ignore" and shorten "Open changelog" to "Changelog"
2. Center-align button label text
3. Style "[Open] changelog" as text, not as a link
4. Shorten the tooltips to be commands, not conditional statements: for example, "Do not prompt me if a new JupyterLab version is available."

JasonWeill:
## References

Fixes #16608.

## User-facing changes

Before the change, the title of the primary button is "Open changelog" and it is styled like a link. The title of the secondary button is "Do not check for updates".

image

This patch changes the changelog button to be styled like a primary (accented) action button, with an ellipsis to suggest that it opens in a new window/tab. The primary button's label is "Changelog…" and the secondary button's is "Disable update notifications".

Screenshot 2024-07-25 at 3 01 12 PM

None.


packages/apputils-extension/src/announcements.ts
```diff
@@ -240,9 +240,9 @@ export const announcements: JupyterFrontEndPlugin<void> = {
               let notificationId: string;
               options.actions = [
                 {
-                  label: trans.__('Do not check for updates'),
+                  label: trans.__('Disable update notifications'),
```
fcollonval, line 243: For the news the text used is:

```
                    label: trans.__('Hide'),
                    caption: trans.__('Never show this notification again.'),
```

Should we use a similar wording?

packages/apputils-extension/src/announcements.ts
```diff
@@ -266,7 +266,9 @@ export const announcements: JupyterFrontEndPlugin<void> = {
                   callback: () => {
                     window.open(link[1], '_blank', 'noreferrer');
                   },
-                  displayType: 'link'
+                  // Because the link to the changelog is the primary option,
+                  // display it in an accent color.
+                  displayType: 'accent'
```
fcollonval, line 271: FYI: The news pop-up mention there is no primary color used. One reason I did not use it is to reduce visual noise for the pop-up.

jupyterlab/handlers/announcements.py
```diff
@@ -110,7 +110,7 @@ async def __call__(self) -> Awaitable[Tuple[str, Tuple[str, str]]]:
                 trans = translator.load("jupyterlab")
                 return (
                     trans.__(f"A newer version ({last_version}) of JupyterLab is available."),
-                    (trans.__("Open changelog"), f"{JUPYTERLAB_RELEASE_URL}{last_version}"),
+                    (trans.__("Changelog…"), f"{JUPYTERLAB_RELEASE_URL}{last_version}"),
```
fcollonval, line 113: It is recommended to use action verb for button label.

If we are to use to use the ellipsis, for homogeneity we should also align the button `Open privacy policy`.

fcollonval (commented):
I have some comments about this one.

packages/apputils-extension/src/announcements.ts
JasonWeill: I had proposed "Ignore" in https://github.com/jupyterlab/jupyterlab/issues/16608#issuecomment-2247142106, but @krassowski was concerned that a one-word label like "Hide" or "Ignore" could lead users to believe that it was a one-time dismissal, without the connotation of "never show this notification again". I personally prefer shorter button labels, though.

krassowski: What about `Ignore all` or `Ignore all updates`?

jupyterlab/handlers/announcements.py
krassowski: `View changes` or `View details`?

JasonWeill: Thoughts about using `More info…` or `Read more…` instead? The word "Changelog" is commonly known among developers, but not all Jupyter users are developers.

krassowski: > Thoughts about using `More info…` or `Read more…` instead

`Learn more...` is another option with a verb (not that I think we have to stick to this rule here)

krassowski: I do not like that "info" is used in "More info".

krassowski: Another point: we should not be prioritising for brevity in English because in localised versions with language packs the equivalents may be much longer.

JasonWeill:
image

"Ignore all" and "Read more…" are readable and compact (in English)

krassowski:
Looks good. Since you updated top-level comment - forgot to push?

JasonWeill:
Pushed most recent button labels.

packages/apputils-extension/src/announcements.ts
fcollonval: _Ignore all updates_ could be a good candidate. _Ignore all_ sounds too generic and may raise questions about what will be ignored.

krassowski: @JasonWeill shall we settle on "Ignore all updates" as per above suggestion?

JasonWeill:
Updated with a new look, per feedback:

image

jupyterlab/tests/test_announcements.py
```diff
@@ -62,7 +62,7 @@ async def test_CheckForUpdateHandler_get_pypi_success(mock_client, labserverapp,
     payload = json.loads(response.body)
     assert payload["notification"]["message"] == message
     assert payload["notification"]["link"] == [
-        "Open changelog",
+        "Changelog",
```
krassowski, line 65: Tests are failing, it looks like this needs to be aligned with the others:

```suggestion
        "Read more…",
```

krassowski (approved):
Thanks @JasonWeill!

Issue #16616, opened 2024-07-25
Style guide for dialog boxes and notifications


JasonWeill:
### Problem

As noted in https://github.com/jupyterlab/jupyterlab/issues/16608#issuecomment-2251103352, we do not have a documented style guide for dialog boxes or notifications.

### Proposed Solution

Add style guides for dialog boxes and notifications to our documentation, which already includes [UI naming conventions](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#user-interface-naming-conventions). It might be worthwhile to split writing style into a separate page for prose writers (contrasted with code writers).

Jupyter products typically use [Apple's alert style](https://developer.apple.com/design/human-interface-guidelines/alerts) for dialogs and notifications: the primary button is accented and on the right, buttons other than OK are typically labeled with a verb, and "Yes" and "No" are typically discouraged in favor of "Do action" / "Don't do action", although these are not consistently applied.

### Additional context

Additional visual style guides:
* [GNOME dialogs](https://developer.gnome.org/hig/patterns/feedback/dialogs.html)
* [Windows dialogs](https://learn.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs)

SarthakPaandey:
I want to work on this task. May you assign this issue to me @JasonWeill.

krassowski:
https://github.com/jupyterlab/jupyterlab/pull/16622 is adding some initial docs here, still worth expanding

PR #16622, opened 2024-07-29
Standardize the button role and label for single, dismissal buttons in dialogs

### LINKED ISSUE #16578: Standardize the button type and label for Dismiss/Close buttons in dialogs
Hi! 👋 

### Problem

Currently, similar buttons for closing dialogs have different button types and labels:

Screenshot 2024-07-12 at 13 45 52
Screenshot 2024-07-12 at 13 45 59
Screenshot 2024-07-12 at 13 49 42
Screenshot 2024-07-12 at 14 01 39
Screenshot 2024-07-12 at 14 14 36
image

These examples present two button types and two labels:

- Button types: `Dialog.cancelButton()` and `Dialog.okButton()`
- Labels: `Dismiss` and `Close`

Assuming there is no historical or functional reason for such differences, I think it would be beneficial to standardize the button type and label for the button to close dialogs.

### Proposed Solution

Personally, I don't have a strong opinion about the button type and label to follow. I would say a combination of `Dialog.cancelButton()` and `Close` (I feel like _Dismiss_ is more typical of JupyterLab, but _Close_ is simpler) works.

Nevertheless, a possible solution would be to go to the following files and change the button type and/or label:

- https://github.com/jupyterlab/jupyterlab/blob/804398451328f39a3705cdc2ecf0593651ebf218/packages/filebrowser/src/listing.ts#L1488
- https://github.com/jupyterlab/jupyterlab/blob/804398451328f39a3705cdc2ecf0593651ebf218/packages/apputils-extension/src/shortcuts.tsx#L197
- https://github.com/search?q=repo%3Ajupyterlab%2Fjupyterlab+trans.__%28%27Dismiss%27%29&type=code

### Additional context

- Browser and version: Chrome 126.0.6478.127
- JupyterLab version: 4.3.0a2
- Python version: 3.10.13

joaopalmeiro:
Hi! 👋 

## References

Closes #16578

## Code changes

The button type/role and the respective label have been adapted where necessary to standardize single, dismissal buttons in dialogs. Documentation about this pattern has also been added.

## User-facing changes

Screenshot 2024-07-29 at 15 28 32

Screenshot 2024-07-29 at 16 08 22

Screenshot 2024-07-29 at 16 08 37

Screenshot 2024-07-29 at 16 15 09

Screenshot 2024-07-29 at 16 25 57

None.

### DIFF
--- a/docs/source/developer/patterns.rst
+++ b/docs/source/developer/patterns.rst
@@ -181,5 +181,5 @@ Dialogs
 Buttons in dialogs with a single, dismissal button (e.g., *About JupyterLab*)
 should have the following attributes:
 
--  Button role: Cancel (``Dialog.cancelButton()``)
+-  Button variant: Cancel (``Dialog.cancelButton()``)
 -  Label: ``Close``
--- a/docs/source/developer/patterns.rst
+++ b/docs/source/developer/patterns.rst
@@ -174,3 +174,12 @@ Command Names
 Commands used in the application command registry should be formatted as
 follows: ``package-name:verb-noun``. They are typically grouped into a
 ``CommandIDs`` namespace in the extension that is not exported.
+
+Dialogs
+-------
+
+Buttons in dialogs with a single, dismissal button (e.g., *About JupyterLab*)
+should have the following attributes:
+
+-  Button role: Cancel (``Dialog.cancelButton()``)
+-  Label: ``Close``
--- a/packages/application/src/connectionlost.ts
+++ b/packages/application/src/connectionlost.ts
@@ -42,7 +42,7 @@ export const ConnectionLost: IConnectionLost = async function (
         'If checked, you will not see a dialog informing you about an issue with server connection in this session.'
       )
     },
-    buttons: [Dialog.okButton({ label: trans.__('Dismiss') })]
+    buttons: [Dialog.cancelButton({ label: trans.__('Close') })]
   })
     .then(result => {
       if (result.isChecked) {
--- a/packages/apputils/src/dialog.tsx
+++ b/packages/apputils/src/dialog.tsx
@@ -44,7 +44,7 @@ export function showErrorMessage(
   buttons?: ReadonlyArray<Dialog.IButton>
 ): Promise<void> {
   const trans = Dialog.translator.load('jupyterlab');
-  buttons = buttons ?? [Dialog.okButton({ label: trans.__('Dismiss') })];
+  buttons = buttons ?? [Dialog.cancelButton({ label: trans.__('Close') })];
   console.warn('Showing error:', error);
 
   // Cache promises to prevent multiple copies of identical dialogs showing
--- a/packages/help-extension/src/index.tsx
+++ b/packages/help-extension/src/index.tsx
@@ -149,9 +149,8 @@ const about: JupyterFrontEndPlugin<void> = {
           title,
           body,
           buttons: [
-            Dialog.createButton({
-              label: trans.__('Dismiss'),
-              className: 'jp-About-button jp-mod-reject jp-mod-styled'
+            Dialog.cancelButton({
+              label: trans.__('Close')
             })
           ]
         });
@@ -422,9 +421,8 @@ const resources: JupyterFrontEndPlugin<void> = {
                 title,
                 body,
                 buttons: [
-                  Dialog.createButton({
-                    label: trans.__('Dismiss'),
-                    className: 'jp-About-button jp-mod-reject jp-mod-styled'
+                  Dialog.cancelButton({
+                    label: trans.__('Close')
                   })
                 ]
               });


docs/source/developer/patterns.rst
```diff
@@ -174,3 +174,12 @@ Command Names
 Commands used in the application command registry should be formatted as
 follows: ``package-name:verb-noun``. They are typically grouped into a
 ``CommandIDs`` namespace in the extension that is not exported.
+
+Dialogs
```
joaopalmeiro, line 178: @JasonWeill, is something like this what you had in mind?

Reference: https://github.com/jupyterlab/jupyterlab/issues/16578#issuecomment-2231336738

JasonWeill: Yes, this is a good start! I also opened #16616 to break out the style guide into an issue of its own.

joaopalmeiro: Should I add something to this PR or is it good for now, @JasonWeill?

packages/application/src/connectionlost.ts
```diff
@@ -42,7 +42,7 @@ export const ConnectionLost: IConnectionLost = async function (
         'If checked, you will not see a dialog informing you about an issue with server connection in this session.'
       )
     },
-    buttons: [Dialog.okButton({ label: trans.__('Dismiss') })]
+    buttons: [Dialog.cancelButton({ label: trans.__('Close') })]
```
krassowski, line 45: I'm not convinced whether this is a cancel action. If user checks the checkbox, clicking the button would cause an action to occur.

joaopalmeiro: That's a very good point! To be honest, it didn't seem problematic given that closing seemed appropriate whether the checkbox was selected or not. I felt that after selecting the checkbox, the `Close` button would be to close that dialog and ignore it during that session.

I'm seeing three options: keep this button as it was before this suggestion, since it's different from the others, keep this button as it is now, or add a new button to this dialog like the kernel selection one or the workspace-related ones:

image

image

I'm not having any idea to combine the current checkbox with two buttons, though. Maybe adapt the one to reset the workspace and have just two buttons, with the buttons being "Close" and "Resume"?

image

Thoughts?

fcollonval: Just to highlight that the checkbox state can be used independently of the clicked button type.

docs/source/developer/patterns.rst
```diff
@@ -174,3 +174,12 @@ Command Names
...
+-------
+
+Buttons in dialogs with a single, dismissal button (e.g., *About JupyterLab*)
+should have the following attributes:
+
+-  Button role: Cancel (``Dialog.cancelButton()``)
```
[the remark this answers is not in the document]
joaopalmeiro: Good point! I followed [Apple's nomenclature](https://developer.apple.com/design/human-interface-guidelines/buttons), but I can also imagine using [button](https://vercel.com/geist/button) [type](https://m2.material.io/components/buttons), [button appearance](https://atlassian.design/components/button/examples#appearance) or [button variant](https://carbondesignsystem.com/components/button/usage/). What do you think?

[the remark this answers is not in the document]
fcollonval: I would go for _variant_ as it is more than appearance and avoid type to not confuse it with variable typing.

packages/application/src/connectionlost.ts
joaopalmeiro: Currently, to also be consistent with the other buttons, the label is `Close`. Do you think it works or should I switch to `Dismiss`, @fcollonval?

krassowski:
@JasonWeill would you be able to help with review of this PR? I recall that you had some opinions on button labels.

packages/application/src/connectionlost.ts
JasonWeill: I'm under the impression that if I click a Cancel button, whether it's labeled `Cancel` or `Close`, the dialog should disappear and nothing should happen. The checkbox should only result in an action if I click a button whose label suggests an action. I'm OK with opening another issue concerning the checkbox/Cancel combination.

In addition, in the "Choose Workspace to Delete", I would expect the primary button to be a red `Delete` button, not a blue `Choose` button, because this action is destructive and irreversible.

joaopalmeiro: Thanks for the feedback, @JasonWeill!

So, for this MR, should I change the blue `Choose` button to a red `Delete` button in the "Choose Workspace to Delete" dialog?

JasonWeill: @joaopalmeiro If the button would result in an immediate deletion of the workspace, a red `Delete` button seems like the right choice.

joaopalmeiro: Regarding your suggestion for the "Choose Workspace to Delete" dialog, @JasonWeill, I had to try it and given the current workflow (video below), I think it is as expected, as there is still a confirmation dialog before the final action. Furthermore, it seems to me that changing this workflow is outside the scope of this MR. What do you think?

https://github.com/user-attachments/assets/d3e3db0b-f499-4ef9-9fcc-ff5047e9fd78

---

If you agree, the last open topic in this thread is about the `Close` vs. `Dismiss` label because of dialogs with a checkbox. Thoughts?

JasonWeill: @joaopalmeiro Thanks for clarifying regarding the "Choose" dialog. I agree, let's leave that as it is. I agree that "Close" seems logical for the dialog with a checkbox; "Dismiss", to me, suggests that the checkbox will have no action when I click the action button.

joaopalmeiro: Thanks, @JasonWeill! 

If everyone agrees on the `Close` vs `Dismiss` discussion, I think this MR is ready for a final review.

krassowski (approved):
Looks good to me, thanks @joaopalmeiro and everyone for the discussion!

jtpio:
Looks like it would indeed be great to include this PR for `4.3.0`, even though it's currently labelled as `enhancement`.

Issue #16623, opened 2024-07-29
Error rendering .ipynb file with array value in a "text/plain" cell output data field


box-turtle:
## Description

I am re-using the notebook component outside of JupyterLab. While trying to load an .ipynb file with the following structure, I receive an unhandled exception and the page does not load. The relevant part of the structure is that the output data contains a string[] and not a string.

```
  {
   "cell_type": "code",
   "execution_count": 6,
   "metadata": {},
   "outputs": [
    {
     "data": {
      "text/plain": [
       "Success: 0 rows affected"
      ]
     }
      ...
    }
```

## Reproduce

load an .ipynb file with an array instead of a string in the ouput.data["text/plain"] field

Debugging in the browser led me to [the code here](https://github.com/jupyterlab/jupyterlab/blob/main/packages/notebook/src/windowing.ts#L64):

```
  estimateWidgetSize = (index: number): number => {
    const model = this.cells[index].model;
    const height = this.cellsEstimatedHeight.get(model.id);
    if (typeof height === 'number') {
      return height;
    }

    const nLines = model.sharedModel.getSource().split('\n').length;
    let outputsLines = 0;
    if (model instanceof CodeCellModel) {
      for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
        const output = model.outputs.get(outputIdx);
        outputsLines = ((output.data['text/plain'] as string) ?? '').split(        // <--- error occurs here
          '\n'
        ).length;
      }
    }
    return (
      NotebookViewModel.DEFAULT_EDITOR_LINE_HEIGHT * (nLines + outputsLines) +
      NotebookViewModel.DEFAULT_CELL_MARGIN
    );
  };
```

I believe the issue is that simply casting an array to string `output.data['text/plain'] as string` results in undefined rather than a string, so the call to `split` fails. I think we need to check whether the value is a string or string[] and handle accordingly.

## Expected behavior

the notebook file renders

## Context

- Operating System and version: Mac Sonomo 14.5
- Browser and version: Chrome Version 126.0.6478.185
- JupyterLab version: 4.2.3

krassowski:
I saw a related error in review of #16498 too, this is why the code should never use `as` casts unless in type guards or tests. A PR  here (with a fix, or unit test, or both) would be very welcome!

JasonWeill:
[...]

PR #16628, opened 2024-07-30
Fix Error Rendering .ipynb with Array in 'text/plain' Output

itsmevichu:
Fix for - https://github.com/jupyterlab/jupyterlab/issues/16623

### DIFF
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -61,9 +61,12 @@ export class NotebookViewModel extends WindowedListModel {
     if (model instanceof CodeCellModel) {
       for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
         const output = model.outputs.get(outputIdx);
-        outputsLines = ((output.data['text/plain'] as string) ?? '').split(
-          '\n'
-        ).length;
+        const data = output.data['text/plain'];
+        if (typeof data === 'string') {
+          outputsLines += data.split('\n').length;
+        } else if (Array.isArray(data)) {
+          outputsLines += data.join('\n').split('\n').length;
+        }
       }
     }
     return (
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -65,7 +65,7 @@ export class NotebookViewModel extends WindowedListModel {
         if (typeof data === 'string') {
           outputsLines += data.split('\n').length;
         } else if (Array.isArray(data)) {
-          outputsLines += data.join('\n').split('\n').length;
+          outputsLines += data.join('').split('\n').length;
         }
       }
     }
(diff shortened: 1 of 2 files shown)


packages/notebook/src/windowing.ts
```diff
@@ -61,9 +61,12 @@ export class NotebookViewModel extends WindowedListModel {
...
       for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
         const output = model.outputs.get(outputIdx);
-        outputsLines = ((output.data['text/plain'] as string) ?? '').split(
-          '\n'
-        ).length;
+        const data = output.data['text/plain'];
+        if (typeof data === 'string') {
+          outputsLines += data.split('\n').length;
+        } else if (Array.isArray(data)) {
+          outputsLines += data.join('\n').split('\n').length;
```
krassowski, line 68: I need to double check this but I think the new lines should be included in the array.

```suggestion
          outputsLines += data.join('').split('\n').length;
```

packages/notebook/test/modelfactory.spec.ts
```diff
@@ -79,4 +85,72 @@ describe('@jupyterlab/notebook', () => {
...
+      const height = notebook.estimateWidgetSize(0);
+      expect(height).toBe(56); // (1 source_line + 1 output_line) * 17 (line_height) + 22 (cell_margin)
+    });
+
+    test('should calculate height based on number of lines in source and output (string[] output)', () => {
+      const outputObj = [
+        {
+          output_type: 'execute_result',
+          data: {
+            'text/plain': ['Output line 1', 'Output line 2', 'Output line 3'],
```
krassowski, line 129:
```suggestion
            'text/plain': ['Output line 1\n', 'Output line 2\n', 'Output line 3\n'],
```

image

packages/notebook/test/modelfactory.spec.ts
```diff
@@ -79,4 +85,72 @@ describe('@jupyterlab/notebook', () => {
...
+          },
+          execution_count: 4,
+          metadata: {}
+        }
+      ];
+      const sharedModel = createStandaloneCell({
+        cell_type: 'code',
+        execution_count: 1,
+        outputs: outputObj,
+        source: ['for i in range(3):\n', '    print(f"Line {i+1}")'],
```
krassowski, line 140: Just to align with output:

```suggestion
        source: ['for i in range(3):\n', '    print(f"Output line {i+1}")'],
```

packages/notebook/test/modelfactory.spec.ts
krassowski: This is the format that I see in ipykernel and IRkernel

krassowski (commented):
It does look like the array elements already contain `\n` to me

packages/notebook/test/modelfactory.spec.ts
itsmevichu: I agree with you, sorry for the misinterpretation. Will adapt the change.

krassowski (approved):
LGTM, thanks @itsmevichu!

PR #16629, opened 2024-07-31
Standardize dialogs to shut down all kernels

### LINKED ISSUE #16624: Standardize dialogs to shut down all kernels
Hi! 👋 

### Problem

The dialogs to confirm or cancel the action of shutting down all kernels are different depending on whether this action is triggered from the sidebar or the command palette:

Screenshot 2024-07-29 at 17 59 32

Screenshot 2024-07-29 at 17 59 42

The body text and the cancel button label are not standardized.

### Proposed Solution

Adopt the body text and the cancel button label of the dialog when the action of shutting down all kernels is triggered from the sidebar: https://github.com/jupyterlab/jupyterlab/blob/78cd6b60e5eb43b1278d825290ac81fd378faed6/packages/running-extension/src/kernels.tsx#L88-L97

... in the equivalent dialog when the same action is triggered from the command palette: https://github.com/jupyterlab/jupyterlab/blob/78cd6b60e5eb43b1278d825290ac81fd378faed6/packages/mainmenu-extension/src/index.ts#L535-L555

### Additional context

- Browser and version: Chrome 127.0.6533.72
- JupyterLab version: 4.3.0a2
- Python version: 3.10.13

joaopalmeiro:
Hi! 👋 

## References

Closes #16624

## Code changes

The content of the command palette dialog was standardized according to the sidebar. Additionally, a new method was introduced in the kernel manager to obtain the number of running kernels and the `shutdownAllConfirmationText` prop was updated to also receive a _dynamic_ function.

## User-facing changes

The dialog content is now consistent regardless of whether the user triggers the shut-down of all kernels from the sidebar or the command palette. Additionally, the number of kernels that will be shut down is now included in the body text.

https://github.com/user-attachments/assets/428f20ba-14b3-4fd4-a9ed-cbc7d01e4fab

Yes, given that some translatable strings have changed.

### DIFF
--- a/packages/mainmenu-extension/src/index.ts
+++ b/packages/mainmenu-extension/src/index.ts
@@ -535,13 +535,15 @@ function createKernelMenu(
   commands.addCommand(CommandIDs.shutdownAllKernels, {
     label: trans.__('Shut Down All Kernels…'),
     isEnabled: () => {
-      return !app.serviceManager.sessions.running().next().done;
+      return !app.serviceManager.kernels.running().next().done;
     },
     execute: () => {
       return showDialog({
         title: trans.__('Shut Down All?'),
-        body: trans.__(
-          'Are you sure you want to permanently shut down all running kernels?'
+        body: trans._n(
+          'Are you sure you want to permanently shut down the running kernel?',
+          'Are you sure you want to permanently shut down the %1 running kernels?',
+          app.serviceManager.kernels.runningCount()
         ),
         buttons: [
           Dialog.cancelButton(),
@@ -549,7 +551,7 @@ function createKernelMenu(
         ]
       }).then(result => {
         if (result.button.accept) {
-          return app.serviceManager.sessions.shutdownAll();
+          return app.serviceManager.kernels.shutdownAll();
         }
       });
     }
--- a/packages/mainmenu-extension/src/index.ts
+++ b/packages/mainmenu-extension/src/index.ts
@@ -540,9 +540,11 @@ function createKernelMenu(
     execute: () => {
       return showDialog({
         title: trans.__('Shut Down All?'),
-        body: trans.__('Shut down all kernels?'),
+        body: trans.__(
+          'Are you sure you want to permanently shut down all running kernels?'
+        ),
         buttons: [
-          Dialog.cancelButton({ label: trans.__('Dismiss') }),
+          Dialog.cancelButton({ label: trans.__('Cancel') }),
           Dialog.warnButton({ label: trans.__('Shut Down All') })
         ]
       }).then(result => {
--- a/packages/mainmenu-extension/src/index.ts
+++ b/packages/mainmenu-extension/src/index.ts
@@ -535,7 +535,7 @@ function createKernelMenu(
   commands.addCommand(CommandIDs.shutdownAllKernels, {
     label: trans.__('Shut Down All Kernels…'),
     isEnabled: () => {
-      return !app.serviceManager.kernels.running().next().done;
+      return !app.serviceManager.sessions.running().next().done;
     },
     execute: () => {
       return showDialog({
@@ -551,7 +551,7 @@ function createKernelMenu(
         ]
       }).then(result => {
         if (result.button.accept) {
-          return app.serviceManager.kernels.shutdownAll();
+          return app.serviceManager.sessions.shutdownAll();
         }
       });
     }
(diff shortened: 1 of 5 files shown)


joaopalmeiro:
Hi, @krassowski!

Based on your idea in #16624, I implemented a proposal to include the number of kernels that will be turned off in the respective dialogs. What do you think of the approach? I needed to change some things, like adding a method in the kernel manager. I'm not sure if I addressed the problem correctly.

Also, do you think it's worth including? Should we extend this approach to other sections, such as open tabs, language servers, etc.? 

Let me know your thoughts, please. All opinions from other people are also welcome!

krassowski (commented):
I'm on the fence as to whether we should add `runningCount` or not (because services are very close to the "core API"), but I think that if we are adding it, then it should likely be better made a readonly getter

packages/mainmenu-extension/src/index.ts
```diff
@@ -535,19 +535,23 @@ function createKernelMenu(
   commands.addCommand(CommandIDs.shutdownAllKernels, {
     label: trans.__('Shut Down All Kernels…'),
     isEnabled: () => {
-      return !app.serviceManager.sessions.running().next().done;
+      return !app.serviceManager.kernels.running().next().done;
```
krassowski, line 538: Is this change needed?

```suggestion
      return !app.serviceManager.sessions.running().next().done;
```

packages/mainmenu-extension/src/index.ts
```diff
@@ -535,19 +535,23 @@ function createKernelMenu(
...
+        ),
         buttons: [
-          Dialog.cancelButton({ label: trans.__('Dismiss') }),
+          Dialog.cancelButton(),
           Dialog.warnButton({ label: trans.__('Shut Down All') })
         ]
       }).then(result => {
         if (result.button.accept) {
-          return app.serviceManager.sessions.shutdownAll();
+          return app.serviceManager.kernels.shutdownAll();
```
krassowski, line 554: I think we want to retain the old behaviour of calling on `sessions.shutdownAll()` here. I am not quite sure of implications if we were to change it so I would not touch it. @afshin may have an opinion as he's worked on the adjacent code recently.

```suggestion
          return app.serviceManager.sessions.shutdownAll();
```

packages/mainmenu-extension/src/index.ts
joaopalmeiro: Not necessarily. I changed them for consistency with what is done in the sidebar. Should I keep it as it was?

packages/mainmenu-extension/src/index.ts
afshin: I think @krassowski is right and we should keep `sessions.shutdownAll()`. Sessions are owned by the JupyterLab server, but kernels may have external origins.

Relatedly, this conversation highlights an issue that the label for this command is `Shut Down All Kernels…` and this ambiguity is unfortunate. We don't have an end-user concept of "session" well-defined, but we should think about the language of this command.

joaopalmeiro: Thanks for the feedback, @afshin!

Given this feedback, I have two questions: 

- In the sidebar, the `shutdownAll` action is triggered for the `kernels` (see snippet below, please). Should we change it to `sessions` or both?
- In the case of the body text for the command palette dialog, should the kernel count be a session count instead? Should I implement `runningCount` in the session manager?

Thanks!

https://github.com/jupyterlab/jupyterlab/blob/653c22621f996249862093d0bb0f23b1279b0f52/packages/running-extension/src/kernels.tsx#L88

afshin: I'm not really sure what to do here. The kernels panel very clearly shows actual kernels, including kernels that do not have an actual "session" associated with them. So that's clear at least.

But as far as the shut down links for them and the running count, I think we should think about the user experience a bit and make some decisions on whether we need to solve this ambiguity or leave things as they are. And if we do need to solve them, the answer might be different language or it might be different behavior. I'm a little unsure, personally.

joaopalmeiro: Thanks for sharing your perspective, @afshin!

So, maybe this PR should just be to standardize the body text of the two dialogs without implementing `runningCount`?

krassowski (approved):
Thank you @joaopalmeiro!

joaopalmeiro:
@krassowski, is this change planned to be released in 4.3.0 or do you not know yet?

Issue #16320, opened 2024-05-12
Rewrite "pending input" popup to be clearer


mr-aryanwasule:
run code problem _the cell not executed due to pending input _ **the cell has not been executed to avoid kernal deadlock as there is another pending input! Submit your pending input and try again_**
   #this problem faced...

krassowski:
This means that one of your cells has an input box which you need to submit before you can continue. For example in Python the input boxes come from executing `input()` or `getpass()` functions (but specific kernels can sent these boxes any time when user input is required):

![input-box](https://github.com/jupyterlab/jupyterlab/assets/5832902/8acee943-a396-4c0b-b5a8-244969d093b5)

JasonWeill:
@mr-aryanwasule Can you please provide information about the OS name and version and the JupyterLab version you're running? Also, if possible, attaching a notebook or a screen shot of the problem would help us to diagnose this issue further.

krassowski:
(To clarify, further details are not needed here - the dialog just needs to make the instructions easier to understand; currently this guards against kernel deadlock but users have no idea what a kernel deadlock is because, well the guard works well against it; the instructions need to specify how to recover, i.e. type something in the input box and press `Enter`)

krassowski:
Flagging this as higher priority because it was reported a few more times, see:
- https://discourse.jupyter.org/t/input-function-not-working/2290/14
- https://discourse.jupyter.org/t/cell-not-executed-due-to-pending-input/24036

PR #16632, opened 2024-07-31
Update pending input message for clarification

RRosio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16320

## Code changes

N/A

## User-facing changes

Users will see a new message indicating they must type something and press enter when there is pending input.

N/A


krassowski (approved):
Thank you @RRosio!

PR #16637, opened 2024-08-01
Option to disable cell toolbar 

### LINKED ISSUE #16609: Add option not to show cell toolbar, without disabling plugin
### Problem

As noted by @krassowski on the 2024-07-24 frontends call, the primary way to disable the cell toolbar from appearing is to disable the `@jupyterlab/cell-toolbar` plugin entirely, which is not available to non-admin users on most JupyterLab installs.

### Proposed Solution

Add a setting, accessible to users, to not show the cell toolbar, even if the plugin remains installed.

### Additional context

The cell toolbar already has an options section in the settings editor.

JasonWeill:
## References

Fixes #16609.

## Code changes

Adds an option to disable the cell toolbar without disabling the cell toolbar plugin.

## User-facing changes

When disabling the setting to show the cell toolbar, the cell toolbar shouldn't show in the active notebook cell.

When re-enabling the setting, or by default, the cell toolbar should show in the active notebook cell, if there is enough room to show it.

toggle-cell-toolbar

None.

### DIFF
--- a/packages/cell-toolbar-extension/src/index.ts
+++ b/packages/cell-toolbar-extension/src/index.ts
@@ -41,25 +41,22 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
      * @param setting Extension settings
      */
     function loadSetting(setting: ISettingRegistry.ISettings | null): void {
-      // Read the settings and convert to the correct type
-      const oldShowCellToolbar = showCellToolbar;
+      // Read the setting and convert to the correct type
       showCellToolbar =
         setting === null
           ? true
           : (setting.get('showToolbar').composite as boolean);
 
-      // If this has changed, re-render the extension
-      if (oldShowCellToolbar !== showCellToolbar) {
-        if (removeWidgetExtension && !removeWidgetExtension.isDisposed) {
-          removeWidgetExtension.dispose();
-        }
+      // Re-render the extension, with new buttons and new visibility
+      if (removeWidgetExtension && !removeWidgetExtension.isDisposed) {
+        removeWidgetExtension.dispose();
+      }
 
-        if (showCellToolbar) {
-          removeWidgetExtension = app.docRegistry.addWidgetExtension(
-            'CellToolbar',
-            new CellBarExtension(app.commands, toolbarItems)
-          );
-        }
+      if (showCellToolbar) {
+        removeWidgetExtension = app.docRegistry.addWidgetExtension(
+          'Notebook',
+          new CellBarExtension(app.commands, toolbarItems)
+        );
       }
     }
 
@@ -94,7 +91,7 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
     }
 
     removeWidgetExtension = app.docRegistry.addWidgetExtension(
-      'CellToolbar',
+      'Notebook',
       new CellBarExtension(app.commands, toolbarItems)
     );
   },
--- a/packages/cell-toolbar-extension/src/index.ts
+++ b/packages/cell-toolbar-extension/src/index.ts
@@ -18,7 +18,6 @@ import {
   IToolbarWidgetRegistry
 } from '@jupyterlab/apputils';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
-import { IDisposable } from '@lumino/disposable';
 
 const PLUGIN_ID = '@jupyterlab/cell-toolbar-extension:plugin';
 
@@ -33,7 +32,6 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
     translator: ITranslator | null
   ) => {
     let showCellToolbar: boolean | null;
-    let removeWidgetExtension: IDisposable | null;
 
     /**
      * Load the settings for this extension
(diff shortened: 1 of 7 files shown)


JasonWeill:
@fcollonval I was looking through search results on GitHub, and I found a few examples of developers calling `app.docRegistry.addWidgetFactory`, but I couldn't find any examples of developers capturing the return value of `addWidgetFactory`, which is-a `IDisposable`, and then trying to dispose it. Are there any examples of extensions or plugins that can update or disable/enable a toolbar on demand?

JasonWeill:
Disregard the comment above; I have rewritten this change to not remove and re-add the widget.

fcollonval:
> @fcollonval I was looking through search results on GitHub, and I found a few examples of developers calling `app.docRegistry.addWidgetExtension`, but I couldn't find any examples of developers capturing the return value of `addWidgetExtension`, which is-a `IDisposable`, and then trying to dispose it. Are there any examples of extensions or plugins that can update or disable/enable a toolbar on demand?

None that I'm aware of.

Note: the disposable should not be accessed across extensions. It follows the same principle as the one of a command: only the extension adding an item can handle it's disposal. Otherwise it will be the chaos depending on the set of extensions installed.
As a reminder, the way of preventing an plugin to be loaded by another one is by specifying disabled plugins not playing with JS.

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -509,11 +525,20 @@ export class CellBarExtension implements DocumentRegistry.WidgetExtension {
...
   createNew(panel: NotebookPanel): IDisposable {
-    return new CellToolbarTracker(panel, undefined, this._toolbarFactory);
+    return (this._tracker = new CellToolbarTracker(
+      panel,
+      undefined,
+      this._toolbarFactory
+    ));
+  }
+
+  set visible(value: boolean) {
```
fcollonval, line 535: Could you add a getter for this?

It is a good practice to define a getter when defining a setter.

fcollonval (changes requested):

packages/cell-toolbar/src/celltoolbartracker.ts
JasonWeill: Getter added.

packages/cell-toolbar-extension/src/index.ts
```diff
@@ -29,7 +31,24 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
     toolbarRegistry: IToolbarWidgetRegistry | null,
     translator: ITranslator | null
   ) => {
-    const toolbarItems =
+    let showCellToolbar: boolean | null;
+
```
krassowski, line 35: This appears to be used only once, inside of the `loadSetting` function body:

```suggestion
```

packages/cell-toolbar-extension/src/index.ts
```diff
@@ -29,7 +31,24 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
...
+    /**
+     * Load the settings for this extension
+     *
+     * @param setting Extension settings
+     */
+    function loadSetting(setting: ISettingRegistry.ISettings | null): void {
+      // Read the setting and convert to the correct type
+      showCellToolbar =
```
krassowski, line 43:
```suggestion
      const showCellToolbar: boolean | null =
```

packages/cell-toolbar-extension/src/index.ts
```diff
@@ -29,7 +31,24 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
...
+        setting === null
+          ? true
+          : (setting.get('showToolbar').composite as boolean);
+
+      extension.visible = showCellToolbar;
+    }
+
+    let toolbarItems =
```
krassowski, line 51: I do not see this change

```suggestion
    const toolbarItems =
```

packages/cell-toolbar-extension/src/index.ts
```diff
@@ -40,10 +59,21 @@ const cellToolbar: JupyterFrontEndPlugin<void> = {
...
+    // Wait for the application to be restored and
+    // for the settings for this plugin to be loaded
+    if (settingRegistry !== null) {
+      void settingRegistry.load(PLUGIN_ID).then(setting => {
+        // Read the settings
+        loadSetting(setting);
+
+        // Listen for your plugin setting changes using Signal
+        setting.changed.connect(loadSetting);
+      });
```
krassowski, line 73: There is a tradeoff here (loading immediately if enabled, vs potentially flashing when disabled), so I will not insist on this, but in case if you did not consider it previously I wanted to suggest:

```suggestion
      const settings = await settingRegistry.load(PLUGIN_ID);

      // Read the settings
      loadSetting(setting);

      // Listen for your plugin setting changes using Signal
      setting.changed.connect(loadSetting);
```

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -152,6 +153,15 @@ export class CellToolbarTracker implements IDisposable {
     return this._isDisposed;
   }
 
+  get visible(): boolean {
```
krassowski, line 156: - 1) Should it be `visible` or `enabled`? As a future user of this API I would expect `visible` to tell me something about the visbility when the toolbar is hidden due to the overlap, but it would not
- 2) Worth adding a docstring?

packages/cell-toolbar/src/celltoolbartracker.ts
```diff
@@ -509,11 +525,24 @@ export class CellBarExtension implements DocumentRegistry.WidgetExtension {
...
+      undefined,
+      this._toolbarFactory
+    ));
+  }
+
+  get visible(): boolean {
+    return this._tracker.visible;
+  }
+
+  set visible(value: boolean) {
```
krassowski, line 539: Same as above, should it be `enabled` and have a docstring?

krassowski (changes requested):
Thank you @JasonWeill! Some comments below.

krassowski (approved):
LGTM, thank you @JasonWeill!

krassowski:
As a follow-up we should update the docs:

https://github.com/jupyterlab/jupyterlab/blob/59d74ad4359696dc0fe5b2e67a0bf84699377f46/docs/source/user/notebook.rst?plain=1#L160-L174

Issue #16553, opened 2024-07-04
Galata helper to create a notebook for a specified kernel. 


pauldruce:
### Problem
I want to leverage the Playwright helpers that Galata provides to interact with notebooks. However, we want to start a notebook with a different kernel specified from the normal. 

I would like the method page.notebook.createNew to have an optional field to specify the name of the kernel to use.
Alternatively, I'd like to be able to switch the kernel a notebook is using more easily. 

### Proposed Solution

My proposed solution would be to update the createNew function to have a second optional argument called `options` which is a json structure which contains a field called `kernel` which specifies the name of the kernel you want to open.

I've got a rough working prototype with the code changes at the end of this issue.
Which I've tested with a custom kernel locally - not sure of the best way to update the tests in the repo to validate this new behaviour.

Some guidance on whether this is an appropriate change would be appreciated. 

For file: `galata/src/helpers/notebook.ts`
```ts
  /**
   * Create a new notebook
   *
   * @param name Name of the notebook (optional)
   * @param options Options for creating the notebook (optional)
   * @param options.kernel Name of the kernel to use for the notebook (optional)
   * @returns Name of the created notebook or null if it failed
   */
  async createNew(name?: string, options?: { kernel?: string }): Promise<string | null> {
    await this.menu.clickMenuItem('File>New>Notebook');

    const page = this.page;
    await page.locator('.jp-Dialog').waitFor();

    /************ New code *******************/
    if (options?.kernel) {
      await page.getByRole('dialog').getByRole('combobox').selectOption(`{"name":"${options.kernel}"}`);
    }
   /*****************************************/

    await page.click('.jp-Dialog .jp-mod-accept');

    const activeTab = this.activity.getTabLocator();
    if (!(await activeTab.count())) {
      return null;
    }

    const label = activeTab.locator('div.lm-TabBar-tabLabel');
    if (!(await label.count())) {
      return null;
    }

    const assignedName = await label.textContent();

    if (!name) {
      return assignedName;
    }

    const currentDir = await this.filebrowser.getCurrentDirectory();
    await this.contents.renameFile(
      `${currentDir}/${assignedName}`,
      `${currentDir}/${name}`
    );
    const renamedTab = this.activity.getTabLocator(name);

    return (await renamedTab.count()) ? name : null;
  }
```

krassowski:
[...]

PR #16639, opened 2024-08-02
Add kernel selection to Galata notebook creation

pauldruce:
## References

Change to address #16553 
Summary of the issue is that the Galata test tools can only open notebooks that use the Python 3 kernel. To make this test tool useful for the author of other kernels, I propose we support the ability to select which kernel to use. 

## Code changes

The proposed change is to add an optional second parameter to the method createNew in the Galata notebook fixture. 
This parameter is a json structure with an optional field called 'kernel' which is the to choose which kernel the notebook should use. 

I should note that the value you have to pass as the "kernel" field is the kernel.id, not the user visible name. This is due to how the Playwright selectOption function works. This value is easily found either via the HTML or via the kernel definition. 

## User-facing changes

Should be no user facing changes - test tool change only. 

Should be backwards compatible - new arguments are optional. No existing tests needed updating.

### DIFF
--- a/galata/src/helpers/notebook.ts
+++ b/galata/src/helpers/notebook.ts
@@ -1430,7 +1430,7 @@ export class NotebookHelper {
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
 
-    if (options && option.kernel !== undefined) {
+    if (options && options.kernel !== undefined) {
       if (options.kernel === null) {
         await page
           .getByRole('dialog')
--- a/galata/src/helpers/notebook.ts
+++ b/galata/src/helpers/notebook.ts
@@ -1414,16 +1414,36 @@ export class NotebookHelper {
   }
 
   /**
-   * Create a new notebook
+   * Creates a new notebook.
    *
-   * @param name Name of the notebook
-   * @returns Name of the created notebook or null if it failed
+   * @param name - The name of the notebook. If provided, the notebook will be renamed to this name.
+   * @param options - Parameters for creating the notebook.
+   * @param options.kernel - The kernel to use for the notebook.
+   * @returns A Promise that resolves to the name of the created notebook, or `null` if the notebook creation failed.
    */
-  async createNew(name?: string): Promise<string | null> {
+  async createNew(
+    name?: string,
+    options?: { kernel?: string }
+  ): Promise<string | null> {
     await this.menu.clickMenuItem('File>New>Notebook');
 
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
+
+    if (options?.kernel) {
+      if (options.kernel === 'null') {
+        await page
+          .getByRole('dialog')
+          .getByRole('combobox')
+          .selectOption('null');
+      } else {
+        await page
+          .getByRole('dialog')
+          .getByRole('combobox')
+          .selectOption(`{"name":"${options.kernel}"}`);
+      }
+    }
+
     await page.click('.jp-Dialog .jp-mod-accept');
 
     const activeTab = this.activity.getTabLocator();
--- a/galata/src/helpers/notebook.ts
+++ b/galata/src/helpers/notebook.ts
@@ -1430,7 +1430,7 @@ export class NotebookHelper {
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
 
-    if (options?.kernel) {
+    if (options && option.kernel !== undefined) {
       if (options.kernel === null) {
         await page
           .getByRole('dialog')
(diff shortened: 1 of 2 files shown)


galata/src/helpers/notebook.ts
```diff
@@ -1414,16 +1414,36 @@ export class NotebookHelper {
...
+    name?: string,
+    options?: { kernel?: string }
+  ): Promise<string | null> {
     await this.menu.clickMenuItem('File>New>Notebook');
 
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
+
+    if (options?.kernel) {
+      if (options.kernel === 'null') {
```
krassowski, line 1434: Would it make sense to make it an actual `null` value here?

```suggestion
    options?: { kernel?: string | null }
  ): Promise<string | null> {
    await this.menu.clickMenuItem('File>New>Notebook');

    const page = this.page;
    await page.locator('.jp-Dialog').waitFor();

    if (options?.kernel) {
      if (options.kernel === null) {
```

galata/test/galata/notebook.spec.ts
```diff
@@ -18,6 +18,32 @@ test.describe('Notebook Tests', () => {
...
+    const toolbar = page.getByRole('toolbar', { name: 'notebook actions' });
+    await expect(toolbar.getByText('Python 3 (ipykernel)')).toBeVisible();
+  });
+
+  test('Create New Notebook with kernel - no kernel', async ({
+    page,
+    tmpPath
+  }) => {
+    const fileName = 'create_no_kernel_test.ipynb';
+    await page.notebook.createNew(fileName, { kernel: 'null' });
```
krassowski, line 38: Then we would also adjust the test case here:

```suggestion
    await page.notebook.createNew(fileName, { kernel: null });
```

galata/src/helpers/notebook.ts
pauldruce: I wouldn't have expected this to work, as `null` is Falsy, so we wouldn't end up inside the first if block. 

There would be no difference between setting it to null and not setting it at all. As I'm trying to make this backwards compatible, I would rather set it to something like "No Kernel".

krassowski: `null` and `undefined` are different values in JavaScript and differentiated when using triple equals (`===`).

pauldruce: Sure, but the `if (options?kernel)` check would fail for null, right?

krassowski: Double checked and this is backwards compatible with `null`. Let's go with using `null` as it is being used like this in other public APIs.

pauldruce: I don't think we would be able to set the No Kernel scenario with code in its current state - I think we need to modify the first if statement I proposed. 
We would need to explicitly check that options.kernel is not undefined now. But that should then work

galata/src/helpers/notebook.ts
```diff
@@ -1414,16 +1414,36 @@ export class NotebookHelper {
...
+  async createNew(
+    name?: string,
+    options?: { kernel?: string | null }
+  ): Promise<string | null> {
     await this.menu.clickMenuItem('File>New>Notebook');
 
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
+
+    if (options?.kernel) {
```
krassowski, line 1433:
```suggestion
    if (options?.kernel !== undefined) {
```

Like this, right?

pauldruce: I'm not sure, I'd be confident in this:
```suggestion
    if (options && option.kernel !== undefined) {
```

Not sure if this is exactly equivalent to what you've provided. 

From the Mozilla page on optional chaining:
> If the object accessed or function called using this operator is [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) or [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null), the expression short circuits and evaluates to [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) instead of throwing an error.

So I think just checking the result of the optional chaining message will not suit our needs.

galata/src/helpers/notebook.ts
```diff
@@ -1414,16 +1414,36 @@ export class NotebookHelper {
...
+  async createNew(
+    name?: string,
+    options?: { kernel?: string | null }
+  ): Promise<string | null> {
     await this.menu.clickMenuItem('File>New>Notebook');
 
     const page = this.page;
     await page.locator('.jp-Dialog').waitFor();
+
+    if (options && option.kernel !== undefined) {
```
krassowski, line 1433: Typo

```suggestion
    if (options && options.kernel !== undefined) {
```

krassowski (approved):
Thank you @pauldruce!

PR #16643, opened 2024-08-05
Update the copyright year in About JupyterLab and LICENSE

### LINKED ISSUE #16642: Outdated copyright year in the About JupyterLab dialog
Hi! 👋 

## Description

The copyright year shown in the _About JupyterLab_ dialog is out of date, if I'm correct. It shows 2023 instead of 2024:

Screenshot 2024-08-05 at 10 31 35

## Reproduce

1. Go to _Help_ > _About JupyterLab_ in the top menu.
2. Check if the second year in the copyright is 2023 instead of 2024.

## Expected behavior

The copyright should show the current year (2024). To this end, 2023 should be updated to 2024: https://github.com/jupyterlab/jupyterlab/blob/653c22621f996249862093d0bb0f23b1279b0f52/packages/help-extension/src/index.tsx#L136-L140

## Context

- Browser and version: Chrome 127.0.6533.89
- JupyterLab version: 4.3.0a2
- Python version: 3.10.13

joaopalmeiro:
Hi! 👋 

## References

Closes #16642 

## Code changes

Change `2023` to `2024`, as well as the format of the translatable string.

## User-facing changes

The current year (2024) is now shown in the _About JupyterLab_ dialog:

Screenshot 2024-08-05 at 11 54 43

Yes, given that a translatable string has changed.

### DIFF
--- a/packages/help-extension/src/index.tsx
+++ b/packages/help-extension/src/index.tsx
@@ -135,11 +135,7 @@ const about: JupyterFrontEndPlugin<void> = {
         );
         const copyright = (
           <span className="jp-About-copyright">
-            {trans.__(
-              '© %1-%2 Project Jupyter Contributors',
-              2015,
-              new Date().getUTCFullYear()
-            )}
+            {trans.__('© %1-%2 Project Jupyter Contributors', 2015, 2024)}
           </span>
         );
         const body = (
--- a/packages/help-extension/src/index.tsx
+++ b/packages/help-extension/src/index.tsx
@@ -135,7 +135,7 @@ const about: JupyterFrontEndPlugin<void> = {
         );
         const copyright = (
           <span className="jp-About-copyright">
-            {trans.__('© 2015-2023 Project Jupyter Contributors')}
+            {trans.__('© 2015-2024 Project Jupyter Contributors')}
           </span>
         );
         const body = (
--- a/packages/help-extension/src/index.tsx
+++ b/packages/help-extension/src/index.tsx
@@ -135,7 +135,11 @@ const about: JupyterFrontEndPlugin<void> = {
         );
         const copyright = (
           <span className="jp-About-copyright">
-            {trans.__('© 2015-2024 Project Jupyter Contributors')}
+            {trans.__(
+              '© %1-%2 Project Jupyter Contributors',
+              2015,
+              new Date().getUTCFullYear()
+            )}
           </span>
         );
         const body = (
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2022 Project Jupyter Contributors
+Copyright (c) 2015-2024 Project Jupyter Contributors
 All rights reserved.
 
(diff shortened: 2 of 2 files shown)


JasonWeill (commented):
Can this be replaced with a variable based on the current year, so that we don't have to revise this every year (including about 5 months from now)?

joaopalmeiro:
> Can this be replaced with a variable based on the current year, so that we don't have to revise this every year (including about 5 months from now)?

Yes, sure! Are there any considerations I should have because this string is a translatable string?

JasonWeill:
I would recommend using a translatable string with two substitutions for the starting and ending year. Example:

https://github.com/jupyterlab/jupyterlab/blob/653c22621f996249862093d0bb0f23b1279b0f52/packages/application/src/utils.ts#L253

joaopalmeiro:
> I would recommend using a translatable string with two substitutions for the starting and ending year. Example:
> 
> https://github.com/jupyterlab/jupyterlab/blob/653c22621f996249862093d0bb0f23b1279b0f52/packages/application/src/utils.ts#L253

Got it! [UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCFullYear)?

JasonWeill:
That should be fine, thank you.

I'm personally of the opinion that copyright years are not necessary in code controlled by source code management, but my opinion does not reflect that of the JupyterLab Council or my employer.

packages/help-extension/src/index.tsx
```diff
@@ -135,7 +135,11 @@ const about: JupyterFrontEndPlugin<void> = {
         );
         const copyright = (
           <span className="jp-About-copyright">
-            {trans.__('© 2015-2023 Project Jupyter Contributors')}
+            {trans.__(
+              '© %1-%2 Project Jupyter Contributors',
+              2015,
+              new Date().getUTCFullYear()
```
krassowski, line 141: What if user's browser has a mis-configured calendar? This could even end up being `© 2015-1982` or so. I think dynamic year is fine as long as this is generated on release time and fed down here. That said, even if we keep it as hard-coded in this PR, the change to use substitution in the translation string is a welcome one as it reduces the burden on translators

joaopalmeiro: Thanks for sharing your perspective, @krassowski! Is the following snippet your suggestion?

```tsx
{trans.__(
  '© %1-%2 Project Jupyter Contributors',
  2015,
  2024
)}
```

If yes, let me know if I should update the PR with this change, please.

krassowski: I think for now that is a safe bet, yes

joaopalmeiro:
Btw, I've now noticed that the year is also outdated in the license file: https://github.com/jupyterlab/jupyterlab/blob/main/LICENSE. Should I update it too in this PR, @JasonWeill, @krassowski?

krassowski:
Thanks for spotting it, yes it makes sense too!

krassowski (approved):
Thanks!

Issue #16239, opened 2024-04-25
The "Modified" header moves sort caret icon to next line if sidebar is narrow


krassowski:
## Description

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/5799f9c2-b811-4803-bd3d-efac99820e4c)

## Reproduce

1. Sort by modified
2. Resize the sidebar in 4.2.0b3

## Expected behavior

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/95b473c8-93ea-4ff9-9de3-efb01b00ff51)

## Context

- JupyterLab version: 4.2.0b3

krassowski:
![issue](https://github.com/jupyterlab/jupyterlab/assets/5832902/e2aa358f-87ae-4264-8f27-930b102604a0)

With https://github.com/jupyterlab/jupyterlab/pull/16207:

![v2](https://github.com/jupyterlab/jupyterlab/assets/5832902/965008f0-8781-4ccb-b80d-fdc848a90641)

There are two options:
- have a third state, the earlier proposed "Mod." for the most narrow option (and delete the ... "too narrow" indicator logic)
- change the "too narrow" indicator logic to kick in eariler

CC @JasonWeill

krassowski:
This also looks bad when there is an extra column.

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/167f27e8-6c3a-43de-be29-19303d2ee4a1)

I think we need a more generic logic on minimum width of each column and to allow each column to change size separately.

Issue #3875, opened 2018-02-15
Resize columns of file browser


mocquin:
In the file browser, say I have long filenames and that I'm not interested in the "last modified" information of my files.

It seems like I can't resize the "columns" of the file browser, in order for instance to display all my long filenames and hide the last-modified-information.

hoishing:
+1 for this feature request, it does help saving the real estate,  especially on small screen

westurner:
- [ ] Where in the source is this [widget, ui control]? [URL]
- [ ] Does the existing widget support resizing a column?
  - [ ] If so, can we just add that attribute for that column of the ~datagrid?
  - [ ] If not, what widget do we need in order to be able to resize a column?
    - [ ] There's like a CSV viewer that's maybe built-in now? Does *that* have resizeable columns either yet? It may be easy to ENH the UBY usability of the {CSV Viewer, File browser, and [...]} with this upgrade that might as well test this consistently if it's not something that there's already a test for in the widget library
      - [ ] Which other components could *reuse* new widget column resizing column?
        - [ ] CSV viewer
        - [ ] File browser {last modified, }
      - [ ] Is there already a test for column width modification in the widget library?
      - [ ] Is there already functionality to store e.g. the "last modified" column width setting at least for the File browser column(s)?

stefansimik:
I wanted to report the same issue. 
It is very uncomfortable, that Last modified column takes so much space and content is often half-empty.

It should be resize-able, because most of the time, people are not ACTIVELY interested in watching LastModified column and
many would prefer to use precious screen space by open notebooks.

PR #16646, opened 2024-08-06
Implement resizing for the file browser

krassowski:
## References

- Closes https://github.com/jupyterlab/jupyterlab/issues/3875
- Fixes https://github.com/jupyterlab/jupyterlab/issues/16239

## Code changes

- after initial load the fixed-width layout is used in favour of the flexbox which is now only used to load initial widths and for backward compatibility
- the filename icon and label are now grouped together into a single node which allows to set a column size on the file name column
- `FileBrowser` and `DirListing` widgets now accept optional `state?: IStateDB` option which allows the listing to restore the size of columns; this follows the pattern already used in the `FileBrowserModel`
- `DirListing.IRenderer` gets a new optional `updateItemSize?(...): void` method which is a fast path to refresh the widths in the listing without updating other elements
- The default implementation of `DirListing.IRenderer.updateItemModified()` will now check if there is any work to be done to avoid costly `format()` calls
- All hidden columns are now detached from the DOM when hidden, which saves time on rendering (which happens often when resizing); previously checkboxes were detached but "Last Modified" and "File Size" were just hidden using CSS

## User-facing changes

- The file browser render and refreshes faster
- Columns can be resized
resize-demo

None


krassowski:
Because I had to rework (=improve) the caret positioning, the caret is no longer inflating the height of the column headers. This does not change the height as seen when checkboxes are enabled (because the checkbox in the header is inflating the height too) but it reduces the height otherwise:

| Before | Now |
|---|--|
| image | image |

I personally don't mind having a little bit more space for files, but if there are strong opinions I am happy to reinstate the extra pixel or two - just let me know.

packages/filebrowser/src/listing.ts
```diff
@@ -2722,19 +3077,23 @@ export namespace DirListing {
         node.removeChild(checkboxWrapper);
       } else if (showFileCheckboxes && !checkboxWrapper) {
         const checkboxWrapper = this.createCheckboxWrapperNode();
-        node.insertBefore(checkboxWrapper, iconContainer);
+        nameColumn.insertAdjacentElement('beforebegin', checkboxWrapper);
```
afshin, line 3080: Switching to `insertAdjacentElement` is nice.

packages/filebrowser/src/listing.ts
```diff
@@ -2888,6 +3278,34 @@ export namespace DirListing {
       return dragImage;
     }
 
+    /**
+     * Factories for individual parts of the item.
+     */
+    protected _itemFactories = {
```
afshin, line 3284:
```suggestion
    protected itemFactories = {
```

(convention for `protected` accessors)

afshin (approved):
Thank you, @krassowski! I made one minor comment.

PR #16647, opened 2024-08-06
Do not block shift-click mouse up handler on active cell

### LINKED ISSUE #16633: JupyterLab suppresses left mouse up while holding shift.
## Description

When left clicking in an interactive plot, JupyterLab suppresses the mouse up event while `shift` is held. This only applies to left clicks, right clicks are correctly forwarded to the plot even if `shift` is held. This only happens in JupyterLab, I ran the interactive plot outside of JupyterLab and left mouse up events were properly forwarded to the plot.

https://github.com/user-attachments/assets/6a39d729-16cd-4cca-8d27-b1ec848f2263

## Reproduce

Display an interactive plot and `shift` + `left click` in the plot.

## Expected behavior

Left mouse up events should be forwarded regardless if the `shift` key state.

## Context

- Operating System and version: MacOS
- Browser and version: Brave
- JupyterLab version: 4.2.4

The plot element has `data-lm-suppress-shortcuts=true` and `data-jp-suppress-context-menu` set.

EdsterG:
## References

Fixes #16633

## Code changes

Don't prevent default behavior when the current cell is the active cell.

## User-facing changes

`shift+click` in the same cell no longer swallows the `mouseup` event.

N/A

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -2672,14 +2672,10 @@ export class Notebook extends StaticNotebook {
       // We don't want to prevent the default selection behavior
       // if there is currently text selected in an output.
       const hasSelection = (window.getSelection() ?? '').toString() !== '';
-      // We don't want to block the shift-click mouse up handler
-      // when the current cell is the active cell.
-      const isActiveCell = index == this.activeCellIndex;
       if (
         button === 0 &&
         shiftKey &&
         !hasSelection &&
-        !isActiveCell &&
         !['INPUT', 'OPTION'].includes(target.tagName)
       ) {
         // Prevent browser selecting text in prompt or output
@@ -2695,6 +2691,13 @@ export class Notebook extends StaticNotebook {
         }
         // Enter selecting mode
         this._mouseMode = 'select';
+
+        // We don't want to block the shift-click mouse up handler
+        // when the current cell is (and remains) the active cell.
+        this._selectData = {
+          startedOnActiveCell: index == this.activeCellIndex,
+          startingCellIndex: this.activeCellIndex
+        };
         document.addEventListener('mouseup', this, true);
         document.addEventListener('mousemove', this, true);
       } else if (button === 0 && !shiftKey) {
@@ -2740,8 +2743,21 @@ export class Notebook extends StaticNotebook {
    * Handle the `'mouseup'` event on the document.
    */
   private _evtDocumentMouseup(event: MouseEvent): void {
-    event.preventDefault();
-    event.stopPropagation();
+    const [, index] = this._findEventTargetAndCell(event);
+
+    let shouldPreventDefault = true;
+    if (this._mouseMode === 'select' && this._selectData) {
+      // User did not move the mouse over to a difference cell, so there was no selection
+      const { startedOnActiveCell, startingCellIndex } = this._selectData;
+      if (startedOnActiveCell && index === startingCellIndex) {
+        shouldPreventDefault = false;
+      }
+      this._selectData = null;
+    }
+    if (shouldPreventDefault) {
+      event.preventDefault();
+      event.stopPropagation();
+    }
 
     // Remove the event listeners we put on the document
     document.removeEventListener('mousemove', this, true);
@@ -2749,8 +2765,6 @@ export class Notebook extends StaticNotebook {
 
     if (this._mouseMode === 'couldDrag') {
       // We didn't end up dragging if we are here, so treat it as a click event.
-
-      const [, index] = this._findEventTargetAndCell(event);
 
       this.deselectAll();
       this.activeCellIndex = index;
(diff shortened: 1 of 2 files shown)


krassowski:
Thank you, this looks good. I think we would also want to add a test case, somewhere here:

https://github.com/jupyterlab/jupyterlab/blob/653c22621f996249862093d0bb0f23b1279b0f52/packages/notebook/test/widget.spec.ts#L1233-L1273

EdsterG:
Thank you for the support.

EdsterG:
I'm not sure why the tests are failing. After looking at this some more I think the test I added doesn't actually cover the new behavior. For the test case added in this PR, `selected` will be empty even without this PR.

What really needs to be tested is whether the widget is a registered mouse up event listener or whether the the mouse mode is null: `expect(widget._mouseMode.toBe(null))`. Also, the new test case needs to be moved to the top because shift clicking the first cell after expanding the selection is valid, and will still be listening to mouseup events.

EdsterG:
> It looks like the parens are off

Oops, thanks! I'm currently not setup for JS dev, so no linters :(

EdsterG:
Looks like the test linter doesn't like `_mouseMode` because it's private. What's the suggested work around here? Or an alternative to test the same thing.

packages/notebook/test/widget.spec.ts
```diff
@@ -1252,6 +1258,12 @@ describe('@jupyter/notebook', () => {
           simulate(widget.widgets[2].node, 'mousedown', { shiftKey: true });
           expect(widget.activeCellIndex).toBe(2);
           expect(selected(widget)).toEqual([2, 3]);
+
+          // shift click deselect
+          simulate(widget.widgets[3].node, 'mousedown', { shiftKey: true });
+          expect(widget.activeCellIndex).toBe(3);
+          expect(selected(widget)).toEqual([]);
+          expect(widget._mouseMode).toBe('select');
```
krassowski, line 1266:
```suggestion
          // test that selecting mode was entered; in selecting mode we listen
          // to the `mouseup` event to stop selecting when the mouse button gets
          // released; this event gets default prevented as handled by the notebook.
          const mouseUpEvent = new MouseEvent('mouseup');
          widget.widgets[3].node.dispatchEvent(mouseUpEvent);
          expect(mouseUpEvent.defaultPrevented).toBe(true);
```

packages/notebook/test/widget.spec.ts
```diff
@@ -1233,6 +1233,12 @@ describe('@jupyter/notebook', () => {
         it('should extend selection if invoked with shift', () => {
           widget.activeCellIndex = 3;
 
+          // shift click no-op
+          simulate(widget.widgets[3].node, 'mousedown', { shiftKey: true });
+          expect(widget.activeCellIndex).toBe(3);
+          expect(selected(widget)).toEqual([]);
+          expect(widget._mouseMode).toBe(null);
```
krassowski, line 1240:
```suggestion
          // test that selecting mode was NOT entered; in selecting mode we listen
          // to the `mouseup` event to stop selecting when the mouse button gets
          // released; this event gets default prevented as handled by the notebook.
          const mouseUpEvent = new MouseEvent('mouseup', { shiftKey: true });
          widget.widgets[3].node.dispatchEvent(mouseUpEvent);
          expect(mouseUpEvent.defaultPrevented).toBe(false);
```

packages/notebook/src/widget.ts
```diff
@@ -2672,10 +2672,14 @@ export class Notebook extends StaticNotebook {
...
       // if there is currently text selected in an output.
       const hasSelection = (window.getSelection() ?? '').toString() !== '';
+      // We don't want to block the shift-click mouse up handler
+      // when the current cell is the active cell.
+      const isActiveCell = index == this.activeCellIndex;
       if (
         button === 0 &&
         shiftKey &&
         !hasSelection &&
+        !isActiveCell &&
```
krassowski, line 2682: Giving it a bit more thought I think this has to be more subtle. The idea is that user can press Shift, press mouse down and start dragging cursor to select cells. Instead of short-circuiting here I think we should store the `activeCellIndex` if starts shift + mousedown on active cell and skip preventing default on `_evtDocumentMouseup()` if the cell index has not changed.

krassowski: I will push a commit

packages/notebook/test/widget.spec.ts
```diff
@@ -1233,6 +1233,20 @@ describe('@jupyter/notebook', () => {
         it('should extend selection if invoked with shift', () => {
           widget.activeCellIndex = 3;
 
+          // shift click no-op
+          simulate(widget.widgets[3].node, 'mousedown', { shiftKey: true });
+          expect(widget.activeCellIndex).toBe(3);
+          expect(selected(widget)).toEqual([]);
+          // test that selecting mode was NOT entered; in selecting mode we listen
+          // to the `mouseup` event to stop selecting when the mouse button gets
```
krassowski, line 1241: Adjust comment

```suggestion
          // test that selecting mode handler does not prevent default
          // if no cells were selected; in the selecting mode we listen
          // to the `mouseup` event to stop selecting when the mouse button gets
```

EdsterG:
Yay, thanks for pushing the changes. LGTM.

krassowski:

Issue #16059, opened 2024-03-25
Kernel state incorrect after a page refresh


mlucool:
## Description

When the page is refreshed, a kernel shows up as free, even when it is busy. Ideally, the queued cells (*) would also be restored. Maybe this is fixed by/related to the bigger work being done in https://github.com/jupyterlab/jupyterlab/pull/15448?

## Reproduce

Given a notebook with the following:
```
from time import sleep

with open("test.txt", "w") as f:
    for x in range(60):
        sleep(1)
        print(x)
        f.write(f"{x}\n")
```
Refresh the page and note that both the kernel status is white (free) and no cell is marked as queued.

## Expected behavior

The kernel status is black (running) and the first cell is queued.

## Context
lab@4.1.2

krassowski:
I can reproduce this across JupyterLab versions on Binder with 3.6.7, 4.0.12, 4.1.5 and 4.2.0a2.

When using kernel-spy I see that frontend sends `kernel_info_request` upon reopening and gets a reply from the kernel, however the kernel does not send the status notification (with `busy` `execution_state`) on IOPub for currently running execution requests.

In a related issue [jasongrout](https://github.com/jasongrout), @jasongrout commented that:

> We do the kernel info request for several reasons:
> 
> 1. It triggers a kernel status message, so that we can get into a known kernel state.
> 2. It gives us information about the kernel (like the help menu links).

While the protocol says:

> Changed in version 5.0: Busy and idle messages should be sent before/after handling every request, not just execution. [source](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-status)

which indeed happens:

![Screenshot from 2024-03-26 10-10-07](https://github.com/jupyterlab/jupyterlab/assets/5832902/8b33fc87-606b-43f5-943d-85310aaea1af)

It does _not_ specify that after receiving `kernel_info_request` the status for pending executions should be sent, but it appears to be a reasonable solution.

krassowski:
Another solution would be for the JupyterLab to persist the unresolved futures in browser memory/state database. The problem with that approach (over "let's make sure kernels send the current status on when being reconnected") is that if we store the futures but these finished already on the backend we would not know about these and show false positive "busy" status. Solving it would require having a way to ask kernel whether a specific execution request is still running.

krassowski:
There are a few related issues:
- https://github.com/jupyterlab/jupyterlab/issues/10174
[...]

Issue #11551, opened 2021-11-25
Execution state is not synced in RTC


krassowski:
## Description

The kernel busy/progress indicator does not sync state on RTC

## Reproduce

1. Open notebook in RTC
2. `from time import sleep`
3. Run a few cells with `sleep(2)`
4. See the indicator does not sync (nor is the * shown in the cell prompt).

The following shows up from time to time in logs, possibly as a consequence:

```
Error: <path> attribute d: Expected number, "…A 104 104 1 0 0 NaN NaN z".
```

which comes from:

https://github.com/jupyterlab/jupyterlab/blob/1bf5fdff171f8a9e91755821efd376d5742778ba/packages/statusbar/src/components/progressCircle.tsx#L18

![bus-state](https://user-images.githubusercontent.com/5832902/143438703-3f75aa35-2278-4328-8586-1aef7974a230.gif)

## Expected behavior

Progress indicator shows progress / busy state.

## Context

- Browser and version: Chrome
- JupyterLab version: 4.0.0a15

krassowski:
This was fixed with JupyterLab 4.3 https://github.com/jupyterlab/jupyterlab/pull/16651, tested just now with JupyterLab 4.5.0 and jupyter-collaboration v4.1.2.

Partial duplicate of https://github.com/jupyterlab/jupyter-collaboration/issues/227

PR #16651, opened 2024-08-07
Store/retrieve execution state from the shared model

krassowski:
## References

- Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16431
- Fixes a number of issues around cell running indicator `[*]`; before this PR
   - it was not shown to collaborators when RTC extension was installed as reported in https://github.com/jupyterlab/jupyterlab/issues/11551
   - it was not shown at all when using server-side execution; sibling PR:
       - https://github.com/datalayer/jupyter-server-nbmodel/pull/25
    - it was lost when refreshing the page even with server-side execution

In principle fixes https://github.com/jupyterlab/jupyterlab/issues/16059 but does not solve the issue of input boxes on refresh yet.

Depends on:
- https://github.com/jupyter-server/jupyter_ydoc/pull/259
- https://github.com/jupyter-server/jupyter_ydoc/pull/197

## Code changes

Stores/retrieves `executionState` from the shared model instead of storing it locally.

## User-facing changes

Execution state indicator is shown:
- for collaborators in RTC setup
- when using server-side execution, including after refresh

None

### DIFF
--- a/dev_mode/package.json
+++ b/dev_mode/package.json
@@ -22,7 +22,7 @@
     "@codemirror/view": "^6.9.6",
     "@jupyter/react-components": "^0.16.6",
     "@jupyter/web-components": "^0.16.6",
-    "@jupyter/ydoc": "^3.0.0-a2",
+    "@jupyter/ydoc": "^3.0.0-a3",
     "@jupyterlab/application": "~4.3.0-alpha.2",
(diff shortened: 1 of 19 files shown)


afshin (approved):
Thank you, @krassowski!

PR #16656, opened 2024-08-08
Convert `@jupyterlab/completer` type colors to theme CSS variables

### LINKED ISSUE #16589: Convert `@jupyterlab/completer` type colors to theme CSS variables
Hi! 👋 

### Problem

The type annotation colors for the [code completer](https://github.com/jupyterlab/jupyterlab/tree/main/packages/completer) are _hard-coded_ in a stylesheet: https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/completer/style/base.css#L139-L181

Screenshot 2024-07-15 at 18 38 58

Screenshot 2024-07-15 at 18 39 07

Therefore, these colors are not customizable from a JupyterLab theme similar to CodeMirror/syntax highlighting colors, for example: https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/theme-light-extension/style/variables.css#L334-L359

### Proposed Solution

Convert [_hard-coded_ colors](https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/completer/style/base.css#L139-L181) to CSS variables defined in JupyterLab themes. To facilitate their association with indexes, the names of these variables could include their known types:

https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/completer/src/model.ts#L587

https://github.com/jupyterlab/jupyterlab/blob/7bae8e3238967a0a776028150b49a4903759bc90/packages/completer/src/widget.ts#L1290-L1311

### Additional context

- Browser and version: Chrome 126.0.6478.127
- JupyterLab version: 4.3.0a2
- Python version: 3.10.13

joaopalmeiro:
Hi! 👋 

## References

Closes #16589

## Code changes

The `@jupyterlab/completer` type background colors are now set by theme CSS variables.

## User-facing changes

Visually, none. For theme creators, it is now possible to customize the `@jupyterlab/completer` color palette along with other theme CSS variables.

None.

### DIFF
--- a/packages/completer/style/base.css
+++ b/packages/completer/style/base.css
@@ -137,47 +137,47 @@
 }
 
 .jp-Completer-type[data-color-index='0'] {
-  background: var(--jp-completer-type-background0);
+  background: var(--jp-completer-type-background0, transparent);
 }
 
 .jp-Completer-type[data-color-index='1'] {
-  background: var(--jp-completer-type-background1);
+  background: var(--jp-completer-type-background1, #1f77b4);
 }
 
 .jp-Completer-type[data-color-index='2'] {
-  background: var(--jp-completer-type-background2);
+  background: var(--jp-completer-type-background2, #ff7f0e);
 }
 
 .jp-Completer-type[data-color-index='3'] {
-  background: var(--jp-completer-type-background3);
+  background: var(--jp-completer-type-background3, #2ca02c);
 }
 
 .jp-Completer-type[data-color-index='4'] {
-  background: var(--jp-completer-type-background4);
+  background: var(--jp-completer-type-background4, #d62728);
 }
 
 .jp-Completer-type[data-color-index='5'] {
-  background: var(--jp-completer-type-background5);
+  background: var(--jp-completer-type-background5, #9467bd);
 }
 
 .jp-Completer-type[data-color-index='6'] {
-  background: var(--jp-completer-type-background6);
+  background: var(--jp-completer-type-background6, #8c564b);
 }
 
 .jp-Completer-type[data-color-index='7'] {
-  background: var(--jp-completer-type-background7);
+  background: var(--jp-completer-type-background7, #e377c2);
 }
 
 .jp-Completer-type[data-color-index='8'] {
-  background: var(--jp-completer-type-background8);
+  background: var(--jp-completer-type-background8, #7f7f7f);
 }
 
 .jp-Completer-type[data-color-index='9'] {
-  background: var(--jp-completer-type-background9);
+  background: var(--jp-completer-type-background9, #bcbd22);
 }
 
 .jp-Completer-type[data-color-index='10'] {
-  background: var(--jp-completer-type-background10);
+  background: var(--jp-completer-type-background10, #17becf);
 }
 
 .jp-Completer-loading-bar-container {
--- a/packages/completer/style/base.css
+++ b/packages/completer/style/base.css
@@ -137,47 +137,47 @@
 }
 
 .jp-Completer-type[data-color-index='0'] {
-  background: transparent;
+  background: var(--jp-completer-type-background0);
 }
 
 .jp-Completer-type[data-color-index='1'] {
-  background: #1f77b4;
+  background: var(--jp-completer-type-background1);
 }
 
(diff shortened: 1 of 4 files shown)


packages/completer/style/base.css
```diff
@@ -137,47 +137,47 @@
 }
 
 .jp-Completer-type[data-color-index='0'] {
-  background: transparent;
+  background: var(--jp-completer-type-background0);
 }
 
 .jp-Completer-type[data-color-index='1'] {
-  background: #1f77b4;
+  background: var(--jp-completer-type-background1);
```
krassowski, line 144: What happens in third party themes which may not have the variables yet? I think this may be breaking so I would add fallback:

```suggestion
  background: var(--jp-completer-type-background1, #1f77b4);
```

joaopalmeiro: Done, @krassowski!

packages/theme-dark-high-contrast-extension/style/variables.css
```diff
@@ -437,4 +437,18 @@ all of MD as it is not optimized for dense, information rich UIs.
...
+  --jp-completer-type-background1: #1f77b4;
+  --jp-completer-type-background2: #ff7f0e;
+  --jp-completer-type-background3: #2ca02c;
+  --jp-completer-type-background4: #d62728;
+  --jp-completer-type-background5: #9467bd;
+  --jp-completer-type-background6: #8c564b;
+  --jp-completer-type-background7: #e377c2;
+  --jp-completer-type-background8: #7f7f7f;
+  --jp-completer-type-background9: #bcbd22;
+  --jp-completer-type-background10: #17becf;
```
krassowski, line 453: What I would ultimately like to do is to stop putting things into `:root` when they do not need to be there. This is because performance suffers doubly:
- from browser having too many variables declared in a scope (this is a small cumulative effect)
- from them being set on global scope when a local one would do (this is a problem especially in Chrome, see https://issuetracker.google.com/issues/40676722)

Does this one work?

```suggestion
}

/* Completer specific styles */
  
.jp-Completer {
  --jp-completer-type-background0: transparent;
  --jp-completer-type-background1: #1f77b4;
  --jp-completer-type-background2: #ff7f0e;
  --jp-completer-type-background3: #2ca02c;
  --jp-completer-type-background4: #d62728;
  --jp-completer-type-background5: #9467bd;
  --jp-completer-type-background6: #8c564b;
  --jp-completer-type-background7: #e377c2;
  --jp-completer-type-background8: #7f7f7f;
  --jp-completer-type-background9: #bcbd22;
  --jp-completer-type-background10: #17becf;
```

I personally do not see that much of an advantage in having to set a variable vs having to write a three-line CSS, My reluctance is because these variables are used only once in the entire CSS codebase but have a non-obvious cost which only becomes apparent when you work with large notebooks and get into the weeds of browser performance profiling. I spent a lot of time on it with https://github.com/jupyterlab/ui-profiler and more low-level tools and saw the `:root` group of variables show up quite often as a potential culprit of UI freezes.

joaopalmeiro: Very interesting, thank you very much for sharing these resources! Proposed changes implemented.

To be honest, in this case, I was prioritizing consistency given how the other variables are organized. Having everything scoped out in `:root` is quite useful for collaborating with multiple people from different backgrounds because, from my perspective, it makes the CSS look like just a configuration file. Anyway, I agree with you.

As a follow-up to this PR, I would like to suggest opening an issue to discuss the reorganization of the variable files for the various themes, as well as adding a note to the [documentation](https://jupyterlab.readthedocs.io/en/stable/developer/css.html#css-variables) about this organization (there is actually already a point about this, but I think it can be extended): _critical_ variables must be scoped in `:root` , the rest in their respective parent classes (with one example). What do you think?

krassowski: Good call, let's continue the discussion in https://github.com/jupyterlab/jupyterlab/issues/16660

krassowski (approved):
Thanks @joaopalmeiro!

PR #16684, opened 2024-08-16
Update to Playwright 1.46.1

jtpio:
## References

Update to Playwright 1.46.1:

- https://github.com/microsoft/playwright/releases/tag/v1.46.0
- https://github.com/microsoft/playwright/releases/tag/v1.46.1

To check if the breaking change mentioned in https://github.com/microsoft/playwright/releases/tag/v1.46.0 may be relevant to the JupyterLab code base.

## Code changes

Update Playwright.

## User-facing changes

None

None

### DIFF
--- a/galata/package.json
+++ b/galata/package.json
@@ -54,7 +54,7 @@
     "@jupyterlab/services": "^7.3.0-beta.0",
     "@jupyterlab/settingregistry": "^4.3.0-beta.0",
     "@lumino/coreutils": "^2.1.2",
-    "@playwright/test": "^1.45.0",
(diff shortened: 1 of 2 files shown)


krassowski:
Even if it does not break tests in JupyterLab it might affect other galata users (in extensions). Should we add a short reference to this possibly-breaking change in the Extension Migration Guide for 4.3.0?

jtpio:
Thanks Mike for the suggestion, that makes sense.

Just pushed https://github.com/jupyterlab/jupyterlab/pull/16684/commits/f798595df3a8edfe362df515d57185ffd38da022 to mention it.

Issue #16669, opened 2024-08-11
Uncaught error in filebrowser sidebar


krassowski:
## Description

When switching to filebrowser sidebar from another sidebar widget:

```
TypeError: Cannot read properties of undefined (reading 'querySelector')
    at Object.findElement (domutils.ts:33:19)
    at Renderer.updateItemSize (listing.ts:3194:27)
    at listing.ts:924:30
    at Array.forEach (<anonymous>)
    at DirListing.updateNodes (listing.ts:921:11)
    at DirListing._updateColumnSizes (listing.ts:1145:14)
    at DirListing.onResize (listing.ts:1071:10)
    at DirListing.processMessage (index.es6.js:1202:1)
    at invokeHandler (index.es6.js:412:1)
    at Object.sendMessage (index.es6.js:166:1)
```

## Reproduce

1. Switch to another sidebar element
2. Switch back to filebrowser

## Expected behavior

No error

## Context

- JupyterLab version: 4.3.0b0

krassowski:
I think this line needs a check on whether the node already exists:

https://github.com/jupyterlab/jupyterlab/blob/fdd45e49e2ad059e4843a4dab915fb1a5cd59902/packages/filebrowser/src/listing.ts#L924

It should be in a new nested `if` and also return.

krassowski:
I cannot reproduce this in released version. Maybe this was just an interim problem with the built assets.

JasonWeill:
To any contributor who wants to pick this issue up: please try reproducing locally using tip-of-`main` before working on a fix.

PR #16693, opened 2024-08-19
Add a guard on uninitialized nodes when resizing, remove log

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16669

## Code changes

- Adds a guard against undefined node in case if array length changes when the resize request if fired.
- Removes spurious `console.log` which slipped through review

## User-facing changes

None

None


gabalafou (approved):
This looks reasonable to me

PR #16701, opened 2024-08-20
Custom env variables

AnastasiaSliusar:
## Code changes

This PR includes functionality of setuping custom env variables 

## User-facing changes
- Kernel select dialog was changed. Now there is opportunity to add custom env variables for a kernel that has not be run yet
- A user can call context menu on Launcher by clicking on a kernel icon and then setup custom env variables for a kernel
dialog_kernel_select 
dialog_custom_env_laucher

## How to use
Should be use together with https://github.com/AnastasiaSliusar/jupyter_server/tree/custom-env-variables


AnastasiaSliusar:

AnastasiaSliusar:

AnastasiaSliusar:
please update snapshots

krassowski:
Related to https://github.com/jupyter/enhancement-proposals/pull/87 ?

AnastasiaSliusar:
> Related to [jupyter/enhancement-proposals#87](https://github.com/jupyter/enhancement-proposals/pull/87) ?

Dear @krassowski no, this feature is independent. It also touches the topic of running a kernel with some customization. But the difference is that a user can setup custom env variables for a kernel and run it without changes kernel spec files. And https://github.com/jupyter/enhancement-proposals/pull/87 is about allowing to add changes to kernel spec files and then use them when a kernel is run.

Issue #16636, opened 2024-08-01
Heuristic for cell height should consider preferred text output types


krassowski:
## Description

Currently the heuristic for cell height only checks for `text/plain`

https://github.com/jupyterlab/jupyterlab/blob/67bfc528d7e3e9a5c62fa692592669d49651eb28/packages/notebook/src/windowing.ts#L52-L71

but most of long outputs would be generated by tracebacks or stdout streams with `application/vnd.jupyter.stderr` or `application/vnd.jupyter.stdout` key.

## Reproduce

Maybe we should do something similar to:

https://github.com/jupyterlab/jupyterlab/blob/967c3d537db12c81fa0f0827e6db97c26316a054/packages/cells/src/widget.ts#L1247-L1268

## Context

- JupyterLab version: 4.3.0a2

Noted in https://github.com/jupyterlab/jupyterlab/issues/16623#issuecomment-2263592079

RINO-GAELICO:
Hi @krassowski , I would like to work on this issue.

Did you mean something like this?
```
if (model instanceof CodeCellModel) { 
  
  const supportedOutputTypes = [ 
    'text/plain', 
    'application/vnd.jupyter.stderr', 
    'application/vnd.jupyter.stdout', 
  ];

  
  for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
    const output = model.outputs.get(outputIdx);
    
    const preferredOutput = supportedOutputTypes.find(mt => {
      const data = output.data[mt];
      return (Array.isArray(data) ? typeof data[0] : typeof data) === 'string';
    });
    
    const dataToDisplay = output.data[preferredOutput ?? ''];

    if (typeof dataToDisplay === 'string') {
      outputsLines += dataToDisplay.split('\n').length;
    } else if (Array.isArray(dataToDisplay)) {
      outputsLines += dataToDisplay.join('').split('\n').length;
    }
  }
}
```


Please let me know your thoughts

hxrshxz:
kindly have a look at the PR

alok844937-design:
I don't see anyone actively working on this right now. 
I'm planning to investigate the cell/output height calculation logic and propose a heuristic that accounts for preferred text output types. 
I'll update this issue with findings or a draft PR once I Validate the approach locally. Any guidance appreciated!

alok844937-design:
[...]

PR #16706, opened 2024-08-21
Considered preferred text output styles for heuristics

doshi-kevin:
## References

 #16636 

## Code changes

1. Expanded Supported Output Types
Change:
The list of supported output types has been expanded to include additional MIME types that are commonly used in Jupyter notebooks, specifically:
application/vnd.jupyter.stderr
application/vnd.jupyter.stdout
text/html
image/svg+xml
application/pdf
text/markdown
text
Purpose:
This change ensures that the height estimation considers outputs from standard error and standard output streams, which may contain significant text that contributes to the overall height of the cell. By including these types, the estimation becomes more comprehensive and accurate.
2. Dynamic Output Type Selection
Change:
The code now dynamically selects the preferred output type from the expanded list of supported types using the find method. It checks if the data associated with each output type is a string, allowing for a fallback to the first valid type found.
Purpose:
This approach allows the function to prioritize output types that are likely to contain meaningful text data. By finding the first output type that yields a string, the code ensures that it captures the most relevant output for height estimation, rather than defaulting to a less informative type.
3. Improved Line Counting Logic
Change:
The logic for counting lines in the output data has been refined. The code now checks if the data to display is defined and then counts non-empty lines using a filter.
Purpose:
This enhancement ensures that only meaningful lines are counted, which directly impacts the accuracy of the height estimation. By filtering out empty lines, the function avoids inflating the estimated height with irrelevant data.
4. Total Height Calculation
Change:
The total height is calculated by summing the number of lines in the source code (nLines) and the calculated output lines (outputsLines). This total is then stored in the cellsEstimatedHeight map for future reference.
Purpose:
This final calculation provides a complete and accurate estimation of the widget size. By combining both the source code and relevant output lines, the function delivers a more realistic height for the cell, which is essential for proper rendering in the UI.
How These Changes Address the Issue
The primary issue was that the original height estimation heuristic only considered text/plain outputs, which did not account for other significant output types that could affect the visual representation of the cell. By expanding the supported output types and dynamically selecting the most relevant type, the revised code ensures that all meaningful output is considered in the height estimation.
Benefits of the Changes
Enhanced Accuracy: The new logic captures a broader range of output types, leading to a more accurate estimation of cell height.
Dynamic Adaptability: The function can now adapt to different types of outputs, making it more robust and flexible for various use cases.
Improved User Experience: With better height estimations, the UI can render cells more effectively, enhancing the overall user experience in Jupyter notebooks.
In summary, these changes significantly improve the functionality of the estimateWidgetSize method, ensuring that it provides a more accurate representation of cell heights based on a wider array of output types.

### DIFF
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -55,24 +55,41 @@ export class NotebookViewModel extends WindowedListModel {
     if (typeof height === 'number') {
       return height;
     }
-
+  
     const nLines = model.sharedModel.getSource().split('\n').length;
     let outputsLines = 0;
     if (model instanceof CodeCellModel) {
       for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
         const output = model.outputs.get(outputIdx);
-        const data = output.data['text/plain'];
-        if (typeof data === 'string') {
-          outputsLines += data.split('\n').length;
-        } else if (Array.isArray(data)) {
-          outputsLines += data.join('').split('\n').length;
+        const supportedOutputTypes = [
+          'text/html',
+          'image/svg+xml',
+          'application/pdf',
+          'text/markdown',
+          'text/plain',
+          'application/vnd.jupyter.stderr',
+          'application/vnd.jupyter.stdout',
+          'text'
+        ];
+  
+        const preferredOutput = supportedOutputTypes.find(mt => {
+          const data = output.data[mt];
+          return (Array.isArray(data) ? typeof data[0] : typeof data) === 'string';
+        });
+  
+        const dataToDisplay = output.data[preferredOutput ?? ''];
+        if (dataToDisplay !== undefined) {
+          outputsLines += (Array.isArray(dataToDisplay)
+            ? dataToDisplay
+            : (dataToDisplay as string)?.split('\n')
+          )?.filter(part => part !== '').length || 0;
         }
       }
     }
-    return (
-      NotebookViewModel.DEFAULT_EDITOR_LINE_HEIGHT * (nLines + outputsLines) +
-      NotebookViewModel.DEFAULT_CELL_MARGIN
-    );
+  
+    const totalLines = nLines + outputsLines;
+    this.cellsEstimatedHeight.set(model.id, totalLines);
+    return totalLines;
   };
 
   /**


packages/notebook/src/windowing.ts
```diff
@@ -61,18 +61,35 @@ export class NotebookViewModel extends WindowedListModel {
...
       }
     }
-    return (
-      NotebookViewModel.DEFAULT_EDITOR_LINE_HEIGHT * (nLines + outputsLines) +
-      NotebookViewModel.DEFAULT_CELL_MARGIN
-    );
+
+    const totalLines = nLines + outputsLines;
+    this.cellsEstimatedHeight.set(model.id, totalLines);
+    return totalLines;
```
krassowski, line 92: This function ought to return an estimated size in pixels, not the number of lines - can you fix it?

packages/notebook/src/windowing.ts
```diff
@@ -61,18 +61,35 @@ export class NotebookViewModel extends WindowedListModel {
...
         }
       }
     }
-    return (
-      NotebookViewModel.DEFAULT_EDITOR_LINE_HEIGHT * (nLines + outputsLines) +
-      NotebookViewModel.DEFAULT_CELL_MARGIN
-    );
+
+    const totalLines = nLines + outputsLines;
+    this.cellsEstimatedHeight.set(model.id, totalLines);
```
krassowski, line 91: This function should not set the estimated height as it is not its responsibility.

krassowski (commented):
Thanks for starting this! Leaving some initial feedback inline.

packages/notebook/src/windowing.ts
doshi-kevin: sure thing, I will make the necessary changes and send another pull request. I am just starting open-source and enjoying this, It would be great to be a contributor to Jupyterlabs.

krassowski: Usually you do not need to open a new pull request, just push changes to this branch and this PR will get updated :)

doshi-kevin: Hey @krassowski , sorry I didn't know about that, I have made the mentioned changes here #16708. Could you please check it and tell me corrections if any.

doshi-kevin: Hey @krassowski any updates on this?
I made a new pull request, could you please check it.

packages/notebook/src/windowing.ts
```diff
@@ -52,27 +52,58 @@ export class NotebookViewModel extends WindowedListModel {
   estimateWidgetSize = (index: number): number => {
     const model = this.cells[index].model;
     const height = this.cellsEstimatedHeight.get(model.id);
+
+
```
krassowski, line 56:
```suggestion

```

Let's avoid using double new line in side of functions.

packages/notebook/src/windowing.ts
```diff
@@ -52,27 +52,58 @@ export class NotebookViewModel extends WindowedListModel {
...
     if (typeof height === 'number') {
       return height;
     }
 
+
```
krassowski, line 61:
```suggestion
```

packages/notebook/src/windowing.ts
```diff
@@ -52,27 +52,58 @@ export class NotebookViewModel extends WindowedListModel {
...
         }
       }
     }
-    return (
-      NotebookViewModel.DEFAULT_EDITOR_LINE_HEIGHT * (nLines + outputsLines) +
-      NotebookViewModel.DEFAULT_CELL_MARGIN
-    );
+    const totalLines = nLines + outputsLines;
+
+
```
krassowski, line 102:
```suggestion
```

krassowski (commented):
Thank you! Can you add a test case for this as well, following the existing test in:

https://github.com/jupyterlab/jupyterlab/blob/91f1c7c53c78920920a627267dcbae7e9eafed49/packages/notebook/test/modelfactory.spec.ts#L96-L122

As for the code changes description, it would be ok (or even preferred) to write one-two lines summary (for a PR of this ### size). The intent here is to have a high-level overview and call out anything that may need attention (like changes to public API) - it should allow to get a quick understanding of the PR without looking at the source code.

packages/notebook/src/windowing.ts
```diff
@@ -52,27 +52,58 @@ export class NotebookViewModel extends WindowedListModel {
...
     if (model instanceof CodeCellModel) {
+      const supportedOutputTypes = [
+        'text/html',
+        'image/svg+xml',
+        'application/pdf',
+        'text/markdown',
+        'text/plain',
+        'application/vnd.jupyter.stderr',
+        'application/vnd.jupyter.stdout',
+        'text'
```
krassowski, line 74:
```suggestion
        'text/markdown',
        'text/plain',
        'application/vnd.jupyter.stderr',
        'application/vnd.jupyter.stdout',
        'text'
```

Maybe we should stick to only these as estimates from html/xml code would be likely unreliable?

packages/notebook/src/windowing.ts
```diff
@@ -52,27 +52,58 @@ export class NotebookViewModel extends WindowedListModel {
...
     if (typeof height === 'number') {
       return height;
     }
 
+
     const nLines = model.sharedModel.getSource().split('\n').length;
     let outputsLines = 0;
+
     if (model instanceof CodeCellModel) {
+      const supportedOutputTypes = [
```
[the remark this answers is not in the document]
krassowski: Or maybe not - if we decide to use a more tailored approach as suggested below, then we would not want to sync it

packages/notebook/src/windowing.ts
doshi-kevin: @krassowski  on it

doshi-kevin:
Hey @krassowski I have added the test case and removed extra spaces. Kindly check for the same .

packages/notebook/src/windowing.ts
```diff
@@ -52,29 +52,91 @@ export class NotebookViewModel extends WindowedListModel {
...
+    const estimatedHeight = totalLines * lineHeight;
+
+    return estimatedHeight;
   };
 
+  // Add the test case below
+}
+
+// Test case for NotebookViewModel
+describe('NotebookViewModel', () => {
```
krassowski, line 107: The test case needs to go to the file I referenced earlier

doshi-kevin:
@krassowski  Hey, sorry for inconviniences, I have added that peice of code to modelfactory files and removed extra spaces and edited the code as well.

packages/notebook/test/modelfactory.spec.ts
```diff
@@ -152,7 +152,47 @@ describe('@jupyterlab/notebook', () => {
       const height = notebook.estimateWidgetSize(0);
       expect(height).toBe(124); // (2 source_line + 4 output_line) * 17 (line_height) + 22 (cell_margin)
     });
-  });
+
+    test('should calculate height based on number of lines in source and multiple outputs', () => {
```
krassowski, line 156: This test does not actually test the new code added.

Are you using an LLM model? Which one?

doshi-kevin: Yeah for the text model I used the perplexity ai llm.

doshi-kevin: I understood the previous test code and thought that according to the given code added, the test generated by the llm might actually be correct.
Could please tell me where might there be an error so that I can resolve it.

krassowski (changes requested):
Sorry for taking longer to get back here, busy times. `js-notebook` and `lint, 3.12` jobs are failing on CI which indicates that there are problems with code style and the tests not passing - can you fix these?

doshi-kevin:
sure. on it !

doshi-kevin:
Hey @krassowski , I have made some linting changes and changes to the test file. Apart from it, I dont understand why (if it still gives error). I will try my best to understand the issue though...

doshi-kevin:
@krassowski any updates ?

doshi-kevin:
Hey @krassowski , sorry to disturb again, but any updates on this ?

JasonWeill:
@doshi-kevin Your tests are still failing; you can see details inline here: https://github.com/jupyterlab/jupyterlab/pull/16706/files#diff-781a703e5af305446726c378a30e53a445113682c768180c073bc042837e0b14

krassowski:
Closing due to lack of response.

Issue #16388, opened 2024-05-28
Add `crossorigin` `anonymous` to images in extension manager


krassowski:
Using cross origin restrictions on server breaks fetching images from remote URLs, for for example in the Extension Manager sidebar the icons of extensions fetched from GitHub will not display correctly. We should be able to fix this by adding `crossorigin="anonymous"` to the image elements (I am not sure if this already works with all browsers).

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/16382#issuecomment-2134556325_

PR #16535, opened 2024-06-29
Add `crossorigin="anonymous"` to GitHub images

krassowski:
## References

Fixes #16388

## Code changes

```diff
+ crossOrigin="anonymous"
```

## User-facing changes

Images in the extension manager display properly when using restricted cross origin policy (e.g. to allow extensions to use `SharedArrayBuffer`).

## Backwards-incompatible changes

None

PR #16711, opened 2024-08-22
Add `crossorigin="anonymous"` to GitHub images

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16388

Similar to #16535 but changed the order

## Code changes

```diff
+ crossOrigin="anonymous"
```

## User-facing changes

Images in the extension manager display properly when using restricted cross origin policy (e.g. to allow extensions to use `SharedArrayBuffer`).

None

### DIFF
--- a/packages/extensionmanager/src/widget.tsx
+++ b/packages/extensionmanager/src/widget.tsx
@@ -62,6 +62,7 @@ function ListEntry(props: ListEntry.IProperties): React.ReactElement<any> {
         {githubUser ? (
           <img
             src={`https://github.com/${githubUser}.png?size=${BADGE_QUERY_SIZE}`}
+            crossOrigin="anonymous"
             style={{ width: '32px', height: '32px' }}
           />
         ) : (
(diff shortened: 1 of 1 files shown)


JasonWeill:
Some images don't display, and I see the following errors in the Firefox (129.0.1) dev console, with this change:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://github.com/aolney.png?size=32. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 302.

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://github.com/chapyter.png?size=32. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 302.

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://github.com/CLIMB-TRE.png?size=32. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 302.

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://github.com/natemacfadden.png?size=32. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 302.
```

krassowski:
Yes, it's still the same as before, which can be seen in the snapshot tests and which is why it is a draft :) I guess that adding a header on our server side might fix it but potentially break other things. I think this will need to wait for 4.4 so we can test it properly.

Issue #15138, opened 2023-09-19
Option to disable input placeholder text


JasonWeill:
### Problem

#13797 added a placeholder text string for `input()` calls. Developers would like the option to disable this.

### Proposed Solution

Add an option to disable the placeholder on inputs.

### Additional context

See also https://github.com/jupyter/notebook/issues/7063.

gianoddone:
Thanks for checking it out.  I have used notebook to make some interactive math games and enjoy the simplicity of input().  I guess there are other ways that afford more control with widgets, but I got stuck looking for a setting to disable the placeholder--which is impressive by the way--just not what I needed in this case.

Not sure I am qualified to contribute just yet, but certainly appreciate the work you all are doing on this project!

gigoiy:
I can contribute by adding this feature! I'll get working on this soon and open a PR.

JasonWeill:
@gigoiy Thank you! Please leave a comment on this issue, or chat with us on Discourse or Gitter, if you need help. You can find our community resources here: https://jupyter.org/community

krassowski:
For reference for anyone wishing to contribute, here is a related pull request which disabled placeholder for password prompts: https://github.com/jupyterlab/jupyterlab/pull/16128

nitesh2920:
Is this still open? I would love to work on it. @krassowski @JasonWeill

JasonWeill:
@nitesh2920 Thank you for your interest! It looks like #16713 would fix this. I just pinged @krassowski to ask about merging this PR in for Lab 4.4.0, after 4.3.0 is released.

PR #15321, opened 2023-10-29
Added basic implementation to toggle the placeholder text 

navdeepgill14:
## References

Closes #15138

## Code changes

1. Introduced a global state variable isPlaceholderVisible in widget.ts to manage the visibility of the placeholder text.
2. Implemented a function toggleGlobalPlaceholderVisibility in widget.ts to toggle the visibility state.
3. Added conditional rendering logic in the cell widget rendering method in widget.ts to show/hide the placeholder based on the isPlaceholderVisible state.

## User-facing changes

1. Users will see a toggle button in the notebook toolbar. Clicking this button will show or hide the placeholder text in the notebook cells.
2. This provides users with the flexibility to choose whether they want to see the placeholder text or not.

## Backwards-incompatible changes

None. The changes introduced are compatible with the existing JupyterLab public APIs and do not break any existing functionality.

PR #16713, opened 2024-08-23
Add option to disable input placeholder text

maitreya2954:
## References

Fixes: https://github.com/jupyterlab/jupyterlab/issues/15138

This pull request is a follow up on https://github.com/jupyterlab/jupyterlab/pull/15321. 

## Code changes

Adds `showInputPlaceholder` setting to notebook defaulted to `True`. Initialize this option in OutputArea, StdIn.

## User-facing changes

Adds new `Show input placeholder` option in Settings.

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -701,7 +701,7 @@
     },
     "showInputPlaceholder": {
       "title": "Show input placeholder",
-      "description": "Show placeholder text for standard input fields.",
+      "description": "Show placeholder text for standard input fields. (Needs reload)",
       "type": "boolean",
       "default": true
     },
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -701,7 +701,7 @@
     },
     "showInputPlaceholder": {
       "title": "Show input placeholder",
-      "description": "Show placeholder text for standard input fields. (Needs reload)",
+      "description": "Show placeholder text for standard input fields (requires reload)",
       "type": "boolean",
       "default": true
     },
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -699,6 +699,12 @@
       "type": "boolean",
       "default": false
     },
+    "showInputPlaceholder": {
+      "title": "Show input placeholder",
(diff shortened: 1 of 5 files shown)


maitreya2954:
@krassowski I followed the https://github.com/jupyterlab/jupyterlab/pull/13944 thread for this PR too and made these changes. But changes to the setting only does apply to new cells created and old cells still would show old behavior unless the entire notebook is reloaded. Is this a desirable behavior?

krassowski:
I think this is ok. You could add a short mention in the settings description that it requires reloading JupyterLab.

JasonWeill:
@krassowski Is this OK to merge in, after 4.3.0 is released?

packages/notebook-extension/schema/tracker.json
```diff
@@ -699,6 +699,12 @@
       "type": "boolean",
       "default": false
     },
+    "showInputPlaceholder": {
+      "title": "Show input placeholder",
+      "description": "Show placeholder text for standard input fields. (Needs reload)",
```
krassowski, line 704: @JasonWeill what do you think would be the best wording here?

```suggestion
      "description": "Show placeholder text for standard input fields (requires reloading JupyterLab to take effect)",
```

krassowski:
> @krassowski Is this OK to merge in, after 4.3.0 is released?

I have no requests for changes and think that is is mostly ok, although the current punctuation in the description feels a bit odd to me.

packages/notebook-extension/schema/tracker.json
JasonWeill: I think that (needs reload) or (requires reload) is both informative and concise. The longer phrase would be better for screen reader users, though.

maitreya2954: @krassowski Should I change the description?

JasonWeill, line 704:
```suggestion
      "description": "Show placeholder text for standard input fields (requires reload)",
```

@maitreya2954 How about this revision to the text? It removes the period after the first sentence and shortens the note in parentheses/brackets. Thanks for your patience; once you accept this change, this pull request should be OK to merge into `main` because we released version 4.3.0 last week.

krassowski (approved):
Thank you @maitreya2954 and thanks @JasonWeill for your help!

PR #16721, opened 2024-08-27
Fix overwriting shortcuts

### LINKED ISSUE #16612: "Shortcut already in use" warning remains even after pressing "Cancel" when overriding shortcuts
## Description

The "Shortcut already in use" warning remains even when "Cancel" is pressed on the UI. Here's a gif - 

shortcut-already-exists

## Reproduce
Reproducer is in the above GIF

## Expected behavior

The warning should disappear once the Cancel button is pressed.

## Context

- JupyterLab version - 4.2.3

### LINKED ISSUE #16803: The logic for disabling shortcuts removes valid shortcuts
## Description

The logic for disabling shortcuts can unexpectedly disable all shortcuts with the same keybinding and selector as the disabled shortcut. This is because the implicit logic for filtering shortcuts does not actually look at the command ID - it will just disable all consecutive shortcuts where `keys` and `selector` match.

This means that "override" logic in shortcuts UI editor does not work well for overriding default shortcuts as the new overriding shortcut gets disabled too.

## Reproduce

In advanced JSON editor set:

```json
{"shortcuts": [
        {
            "command": "notebook:merge-cells",
            "selector": ".jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus",
            "disabled": true,
            "keys": [
                "Shift M"
            ]
        },
        {
            "command": "notebook:merge-cell-below",
            "selector": ".jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus",
            "keys": [
                "Shift M"
            ]
        }
    ]
}
```

And refresh JupyterLab. See that `notebook:merge-cell-below` gets filtered out by the logic in:

https://github.com/jupyterlab/jupyterlab/blob/058d05f720447737274dbec9377af7f6c6c3e7fa/packages/settingregistry/src/settingregistry.ts#L1158-L1231

```
settingregistry.ts:1189 Skipping this shortcut because it collides with another shortcut. 
{command: 'notebook:merge-cell-below', selector: '.jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus', keys: Array(1)}
```

Now set:

```json
{"shortcuts": [
        {
            "command": "notebook:merge-cell-below",
            "selector": ".jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus",
            "keys": [
                "Shift M"
            ]
        },
        {
            "command": "notebook:merge-cells",
            "selector": ".jp-Notebook.jp-mod-commandMode:not(.jp-mod-readWrite) :focus",
            "disabled": true,
            "keys": [
                "Shift M"
            ]
        }
    ]
}
```

and see that the warning is now flipped (but the behaviour is close to intended, this is `notebook:merge-cell-below` now works on Shift + M.

## Expected behavior

Disabling shortcut X disables shortcut X, not shortcut Y which happens to have the same keys set.

## Context

- JupyterLab version: 4.3.0b1

### LINKED ISSUE #16809: Deleting a keyboard shortcut in the Settings Editor removes other user-set shortcut settings
## Description

When I used the Settings Editor to delete a keyboard shortcut, *JupyterLab* rewrote the entire "shortcuts.jupyterlab-settings" file to contain only boilerplate comments and the one deleted shortcut, thereby removing any shortcuts and comments previously contained in the file.

## Reproduce

1. Optionally, back up the "shortcuts.jupyterlab-settings" file.

2. Within *JupyterLab*, open the Settings Editor (menu item "Settings > Settings Editor"). In the "Settings" tab, within its left pane, click "Keyboard Shortcuts".

3. In the "Settings" tab, at the upper right, click "JSON Settings Editor". In the "Advanced Settings Editor" tab, within its left pane, click "Keyboard Shortcuts". Ensure that the "User Preferences" pane shows non-trivial valid content, such as:

    ``` json
    {
        "shortcuts": [
            {
                "command": "mainmenu:open-first",
                "keys": ["Alt F10"],
                "selector": "body"
            }
        ]
    }
    ```

4. In the "Settings" tab, within its main pane, in the "Shortcut" column, right-click a shortcut and select the "Delete Keybinding" menu item.

5. In the "Advanced Settings Editor" tab, observe that the JSON object in the "User Preferences" pane now contains the comments shown in the "System Defaults" pane followed by only the deleted shortcut:

    ``` json
    {
        // Keyboard Shortcuts
        // ...
        // List of keyboard shortcuts:
        "shortcuts": [
            {
                "command": "application:toggle-sidebar-widget",
                "selector": "body",
                "disabled": true,
                "keys": [
                    "Alt 1"
                ]
            }
        ]
    }
    ```

Moreover, one may return to the "Settings" tab, delete a different shortcut, and find that the user preferences now show only the newly deleted shortcut.

## Expected behavior

I expect that deleting a shortcut in the "Settings" tab would modify the user preferences no more than necessary, leaving prior shortcuts and comments in place.

## Context

- Operating System and version: Windows 11 21H2
- Browser and version: Firefox 130.0, Chrome 129.0.6668.59
- JupyterLab version: 4.2.5

krassowski:
## References

- [x] Adds a test (should fail in first commit)
- [x] Fixes #16612
- [x] Fixes #16803
- [x] Fixes #16809 

## Code changes

- Previously the `conflicts` state was in both parent and child component; this PR moves all management of the `conflicts` state to the parent component which fixes the behaviour of the conflicts banner on clicking the "Cancel"/"Submit" button and on blur.
- 

## User-facing changes

- Clicking "Cancel" removes the overrides notification
- Clicking "Overwrite" removes the overrides notification sooner than before  preventing unintended consequences of accidentally clicking the button twice or more
- Adjust the button to use jupyter style
- Clicking "Overwrite" will no longer drop other keybindings, and the overwritten keybindings will be shown correctly for default overrides

| Before | After |
|--|--|
| before | after |

Note:
- how before clicking on any button was not dismissing the warning
- how the shortcuts table was unexpectedly resized when the warning was shown.
- that before poorly styled (poor contrast  = non accessible) the buttons were.

None


krassowski:

galata/test/jupyterlab/settings.test.ts
```diff
@@ -250,3 +250,42 @@ test('Opening Keyboard Shortcuts settings does not mangle user shortcuts', async
...
+  await page.evaluate(async () => {
+    await window.jupyterapp.commands.execute('settingeditor:open', {
+      query: 'Keyboard Shortcuts'
+    });
+  });
+
+  const shortcutsForm = page.locator('.jp-Shortcuts-ShortcutUI');
+  const filterInput = shortcutsForm.locator('jp-search.jp-FilterBox');
+  await filterInput.click();
+  await filterInput.pressSequentially('merge cell below');
```
fcollonval, line 269: To answer your question, playwright can pierce through open ShadowDOM to reach the input. So it is indeed more convoluted but the following code works:

```suggestion
  await filterInput.locator('input').fill('merge cell below');
```

fcollonval (commented):
Thanks @krassowski 

It works as expected. But there seems to be some another cases in which the new shortcut candidate is discarded but the warning remain:


https://github.com/user-attachments/assets/900d6282-7b8a-4ed3-931d-9083549c360c

I guess this is related to the fact that you are now capturing explicitly the click on cancel and not letting it propagate to consider clicking anywhere other than overwrite to induce a cancel action.

krassowski:
So the problem is that the callback after `setState` in:

https://github.com/jupyterlab/jupyterlab/blob/058d05f720447737274dbec9377af7f6c6c3e7fa/packages/shortcuts-extension/src/components/ShortcutInput.tsx#L334-L352

does not get called thus  `this._updateConflictDisplay();` is not invoked on blur. I verified that this handler does get called and that the `if` clause is entered correctly and that the state before calling `setState` is different.

krassowski:
Removing `this.props.toggleInput()` call "fixes" it (but then the input does not get hidden). The problem is that the state management is really tightly coupled in these components as they are all written in imperative style with multiple callbacks between child and parent.

JasonWeill:
When I expand the "Add" UI, I notice a very slight (1–2 pixel) shrink of the row. Other rows move up. (macOS, Firefox)

row-shrink

krassowski:
> When I expand the "Add" UI, I notice a very slight (1–2 pixel) shrink of the row. Other rows move up. (macOS, Firefox)
> 
> row-shrink
> 

Is this related to this PR?

JasonWeill:
When I am editing the one and only shortcut in a row, while the mouse cursor is over that row, "Add" disappears, but "or" still appears, even though there is nothing after it.

or-appears

krassowski:
Can you open these as separate issues? I know that there are many ways to improve the shortcuts UI.

krassowski:
(unless you can point that any of these is a regression compared to main branch or the released version)

JasonWeill:
@fcollonval Are you satisfied that the use cases you identified as failing earlier are now resolved? I opened separate issues for my items above. If we're all satisfied that the related bugs are fixed, I'm OK to merge.

jtpio (approved):
Thanks!

Tried locally and it seems to be working well :+1: 

I restarted the failed CI check, but we should be able to get this one in after they finish, so it's part of the next beta release.

Issue #16730, opened 2024-08-29
Update snapshot workflows sometimes do not run


krassowski:
> Note update-snapshot runs: workflow was not successful in updating snapshots, all recent runs just say “This job was skipped”. Nothing critical but worth looking at some point in order not to update snapshots manually which takes longer https://github.com/jupyterlab/jupyter-ai/actions/workflows/update-snapshots.yml

Also saw this one recently here: https://github.com/jupyterlab/jupyterlab/pull/16565#issuecomment-2311092393 - where Andrew was a "Member" so it should have passed. I see that @michaelchia is a collaborator so It should have passed too. On the other hand just yesterday it worked well for me here: https://github.com/jupyterlab/jupyterlab/pull/16721#issuecomment-2314825928.

For Andrew's PR I think it is because it was open before we updated workflows on July 16th due to https://github.com/jupyterlab/extension-cookiecutter-ts/security/advisories/GHSA-rc7c-v5qg-g3vw. 

Here? No idea. It would be good to create a dummy workflow and print out values for `github.event.issue.author_association`, `github.event.issue.pull_request`, and `github.event.comment.body` just to make sure. Or maybe we should move the association guards from `if` section which causes skips into the body of the workflow and then print out a useful warning (i.e. useful for debugging).

As per docs the enums we have are fine: https://docs.github.com/en/webhooks/webhook-events-and-payloads#issue_comment

![image](https://github.com/user-attachments/assets/0a09eea0-96ed-4b2a-9210-0a921f9f8d6b)

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyter-ai/issues/951#issuecomment-2316783945_

krassowski:
[This](https://github.com/manics/binder/blob/ff9b4fba9e8c1a22fd83622d9cbb8af2b434d488/doc/howto/gh-actions-badges.md?plain=1#L37) was helpful:

> you can access the association of the person who opened the pull request with the variable `github.event.issue.author_association` and the person commenting on the pull request with `github.event.comment.author_association`.

Issue #13142, opened 2022-09-28
UI design for an empty notebook


hbcarlos:
### Problem
As presented in #13141, adding a new cell every time we have an empty notebook conflict with RTC.

This issue is to discuss a UI design to improve the UX on empty notebooks. 

### Proposed Solution
Add a button to create a new cell when the notebook is empty.

### Additional context

cc @GabrielaVives

GabrielaVives:
I suggest we create a standard design for empty notebooks. It is important for the user experience to for the design to have to components:
- a clear UI informing the user that it is an empty notebook that is being displayed - for example, a clear and short message.
- a CTA to give the user a way to easily move forward - for example, buttons near message prompting the user to add a cell

Here is an example in the following prototype: 
https://www.figma.com/proto/tPuLOg7FLmvVVK4q7EWlNc/Empty-notebook-design?node-id=311%3A3031&scaling=min-zoom&page-id=311%3A3030&starting-point-node-id=311%3A3031

and as a screenshot:

Neutral state
![Screenshot 2022-09-28 at 14 17 03](https://user-images.githubusercontent.com/105775162/192776466-aad1c42e-8f20-4280-991e-4ee4b864ce74.png)

When hovering over a button
![Screenshot 2022-09-28 at 14 17 57](https://user-images.githubusercontent.com/105775162/192776589-a2084bf4-75ab-47cb-a0a4-9a6af6129962.png)

davidbrochart:
Note that in Observable notebooks, creating a new notebook starts with an empty cell, and deleting it just clears its content.
This is the behavior @krassowski described in https://github.com/jupyterlab/jupyterlab/pull/13141#issuecomment-1260785653, and I also prefer that compared to introducing a new UI.

firai:
I would also advocate against moving away from the current behavior to having to click a button in a fresh notebook to add a cell, as the current behavior allows the user to immediately start typing once a new document is started, which is similar to the experience in word processors and text editors. VSCode, Google Colab and Deepnote all add an empty cell to new notebooks and focus into the empty cell by default as well. Like Observable, deleting the last cell in Deepnote, which is another platform that supports RTC, just clears the content of the last cell.

SylvainCorlay:
[...]

PR #16732, opened 2024-08-30
Allow empty notebook

### LINKED ISSUE #16718: Don't show a cell when notebook is empty
## Description

JupyterLab shows an empty cell even when there is no cell in the `.ipynb` file. This is just a UI convenience, as the file is updated with the new cell only when the cell is changed.
In RTC this becomes a problem. For instance, [Jupyter Shared Drive](https://github.com/davidbrochart/jupyter-shared-drive) uses WebRTC to sync clients. When a client opens a shared notebook, it is initially empty, even if another client inserted a cell, because synchronization takes time. Thus JupyterLab inserts this "fake cell", then synchronization happens over the network, and the notebook appears with an extra cell.

https://github.com/user-attachments/assets/483bc70a-3d61-44d3-b05e-516fe620a808

## Reproduce

1. Create an empty notebook by saving the following to a file `Untitled.ipynb`:
```json
{
 "cells": [
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3 (ipykernel)",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.12.5"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 5
}
```
2. Launch `jupyter lab Untitled.ipynb`.
3. See that JupyterLab shows an empty cell:

image

4. Quit JupyterLab, and see that there are no cell in `Untitled.ipynb`:
```
{
 "cells": [
 ],
 ...
}
```

## Expected behavior

I think there is no reason to have this behavior any more, because we now have the "Click to add a cell" placeholder:

image

It currently shows on mouse hover, but we could have it always visible until a cell is created.

davidbrochart:
## References

Closes #16718.

## Code changes

No cell is created when there is no cell in the notebook.

## User-facing changes

When a notebook has no cell, or when a user deletes the only cell in a notebook, no cell is shown. Instead, the placeholder "Click to add a cell" is shown.

https://github.com/user-attachments/assets/f90eead6-1423-4d8f-bde0-a4eda49849be

None.

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -686,9 +686,15 @@
         "lineWrap": false
       }
     },
+    "allowEmptyNotebook": {
+      "title": "Allow Empty Notebook",
+      "description": "Whether not to show a cell when the notebook is empty.",
+      "type": "boolean",
+      "default": false
+    },
     "defaultCell": {
       "title": "Default cell type",
-      "description": "The default type (markdown, code, or raw) for new cells",
+      "description": "The default type (markdown, code, or raw) for new cells.",
       "type": "string",
       "enum": ["code", "markdown", "raw"],
       "default": "code"
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -1796,6 +1796,7 @@ function activateNotebookHandler(
 
   const registry = app.docRegistry;
   const modelFactory = new NotebookModelFactory({
+    allowEmptyNotebook: factory.notebookConfig.allowEmptyNotebook,
     disableDocumentWideUndoRedo:
       factory.notebookConfig.disableDocumentWideUndoRedo,
     collaborative: true
@@ -1870,6 +1871,8 @@ function activateNotebookHandler(
       showHiddenCellsButton: settings.get('showHiddenCellsButton')
         .composite as boolean,
       scrollPastEnd: settings.get('scrollPastEnd').composite as boolean,
+      allowEmptyNotebook: settings.get('allowEmptyNotebook')
+        .composite as boolean,
       defaultCell: settings.get('defaultCell').composite as nbformat.CellType,
       recordTiming: settings.get('recordTiming').composite as boolean,
       overscanCount: settings.get('overscanCount').composite as number,
@@ -1920,6 +1923,9 @@ function activateNotebookHandler(
     modelFactory.disableDocumentWideUndoRedo = !settings.get(
       'documentWideUndoRedo'
     ).composite as boolean;
+
+    modelFactory.allowEmptyNotebook = settings.get('allowEmptyNotebook')
+      .composite as boolean;
 
     updateTracker({
       editorConfig: factory.editorConfig,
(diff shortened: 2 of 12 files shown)


davidbrochart:

davidbrochart:
> bot please update galata snapshots

It looks like I cannot talk to the bot :smile:

krassowski:
We need to investigate a bit more. So far it looks like it runs on "OWNER" and "COLLABORATOR" but not on "MEMBER" despite having "MEMBER" in the code https://github.com/jupyterlab/jupyterlab/issues/16730

https://github.com/jupyterlab/jupyterlab/blob/0b12f664f24da10a6ac04c8a6b7bfde309ed3363/.github/workflows/galata-update.yml#L15-L22

krassowski:

krassowski:
Or maybe it does not run if your `main` is out of date?

krassowski:
Though it's named `master` not `main`.

davidbrochart:
I don't understand, what is named `master`?

davidbrochart:
I renamed `master` to `main`, but it still doesn't work.

jtpio:
Thanks @davidbrochart for exploring this.

As a follow-up to the prior discussion in https://github.com/jupyterlab/jupyterlab/issues/13142, it might a change to be discussed more extensively UX-wise.

Checking with other notebook implementations such as VS Code, Colab, Observable (and likely others), all seem to be adding that extra cell, at least to their UI. This is indeed convenient because users can start typing directly and don't need to click on an extra button.

So maybe there could instead be a way for configuring this behavior, so the `jupyter-shared-drive` extension could choose to not have this default empty cell? This would help to not disrupt the workflow of those not using RTC.

davidbrochart:
> So maybe there could instead be a way for configuring this behavior, so the `jupyter-shared-drive` extension could choose to not have this default empty cell? This would help not disrupt the workflow of those not using RTC.

I think that if we make this behavior configurable:
- it should be off by default (there *is* a cell when the notebook is empty),
- if an extension turns it on (there is no cell when the notebook is empty), another extension cannot turn it off,
- it cannot be set by the user (in the settings).

I'm imagining a token that would provide two functions, `allow_empty_notebook()` and `empty_notebook_allowed()`.
What do you think?

davidbrochart:
Well, in the end I just added a new setting called `allowEmptyNotebook`, which defaults to `false`, i.e. a cell is added when the notebook is empty.

davidbrochart:
Should we include this PR to 4.3.0?

krassowski:
> Should we include this PR to 4.3.0?

Maybe worth discussing on frontends call? I do not have a strong feeling one way or another here.

krassowski:
Would an alternative be to label the default cell as "do not add if there is one already" on jupyter_ydoc level?

krassowski:
@davidbrochart is this still something you would like to merge?

krassowski:
@davidbrochart in case if you missed the previous ping. It looks fine to merge if it is still needed for RTC?

davidbrochart:
Thanks for the ping, I missed that one.
Running JupyterLab with `jupyter-webrtc-provider` installed, it doesn't actually create a cell when creating a new notebook or loading an empty notebook. So I think this PR is not needed anymore.

Issue #16698, opened 2024-08-20
Document runtime packages and runtime dependencies from security perspective


krassowski:
### Problem

Automated scanning alerts often incorrectly flag development/test dependencies of JupyterLab as using vulnerable versions. This is usually incorrect because these are used in tooling and not serving applications - JupyterLab is not using any of the Node.js packages at runtime. We receive many reports (on top of our automated scanning false positives).

Reporters currently do not have a good way to assess if an alert about a dependency is a false positive or not.

### Proposed Solution

Provide a list of packages included in the default runtime bundle as part of security documentation.

### Additional context

Currently such a list (or the best proxy for it) is accessible under `Settings` menu → `Licences`:

![image](https://github.com/user-attachments/assets/1632827e-5f01-46a5-b04c-368ee12a8f5c)

If a package is on this list and has a vulnerability, then it **might** be also a vulnerability in JupyterLab. If a package is not on this list it is either lacking a license (unlikely!) or is not included in the final bundle.

Other ways to confirm that a package is not included in the bundle:
- `jlpm why packagename` can be used to determine the dependency ancestors for a given dependency. If it is only included in test packages (jest), linters and monorepo tools (lerna) then it is likely not a runtime dependency.
- distribution tarballs can be downloaded from PyPI and grepped for package names (can have false positives for package names like "braces")/code (usually by string literals as most of the code gets minimized away)

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

nightcityblade:
I noticed there is already an open PR that appears to address this issue: https://github.com/jupyterlab/jupyterlab/pull/18242. I’ll step back to avoid duplicating work.

PR #16740, opened 2024-09-03
Require newer handlebars version (^4.7.7) to silence false positive CVE notifications

mjpsyapse:
Updates handlebars to 4.7.7.  This resolves [CVE-2021-23369](https://nvd.nist.gov/vuln/detail/CVE-2021-23369)

## References

See https://nvd.nist.gov/vuln/detail/CVE-2021-23369

## Code changes

No code changes, only an update to handlebars

## User-facing changes

None


colin-eberhardt (approved):

jupyterlab/staging/yarn.lock
```diff
@@ -6370,7 +6370,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"handlebars@npm:4.7.7, handlebars@npm:^4.5.3":
+"handlebars@npm:4.7.7":
   version: 4.7.7
```
krassowski, line 6374: The old version is still in the lockfile.

yarn.lock
```diff
@@ -12537,7 +12538,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"handlebars@npm:4.7.7, handlebars@npm:^4.5.3, handlebars@npm:^4.7.7":
+"handlebars@npm:4.7.7, handlebars@npm:^4.7.7":
   version: 4.7.7
```
krassowski, line 12542: Old version still there.

jupyterlab/staging/yarn.lock
mjpsyapse: Ah thanks. Looks like verdaccio is messing with me.  I will update that dep, too.

mjpsyapse: It turns out it's not as easy as i'd hoped. we could go down the rabbit hole of updating deps one by one, but there is also the option of doing a wholesale upgrade of all the deps. I will keep this pr focused on handlebars, but make another PR to upgrade all the deps.

krassowski: It's ok, let's just rename the PR to reflect that and de-duplicate the lockfile (e.g. `jlpm dlx yarn-berry-deduplicate`) to avoid having two versions

mjpsyapse:
was just about to resume this work today.

yarn.lock
```diff
@@ -12556,7 +12556,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"handlebars@npm:^4.7.8":
+"handlebars@npm:^4.7.7, handlebars@npm:^4.7.8":
```
mjpsyapse, line 12559: Did you manually set this? I'm not familiar with and don't have the guts to edit the lock file.

krassowski: No, I just run `jlpm dlx yarn-berry-deduplicate`. But you could see the changelog the integrity job was asking for by looking at the logs:

image

package.json
```diff
@@ -116,6 +116,7 @@
     "eslint-plugin-jest": "~27.6.0",
     "eslint-plugin-prettier": "~5.0.1",
     "eslint-plugin-react": "~7.33.2",
+    "handlebars": "^4.7.8",
```
krassowski, line 119: Actually, I think we should not need this on top-level:

```suggestion
```

krassowski:
Just to clarify: I believe that there was no handlebars older than 4.7.7 in the lock file before this PR, so this PR only bumps the required versions, correct? If so we should rename it to reflect it?

mjpsyapse:
> Just to clarify: I believe that there was no handlebars older than 4.7.7 in the lock file before this PR, so this PR only bumps the required versions, correct? If so we should rename it to reflect it?

Good point. I apologize for making blind changes.  For context, I'm seeing this in the docker image (`quay.io/jupyter/datascience-notebook:notebook-7.2.2`) when i grep for handlebars:
```
/opt/conda/share/jupyter/lab/static/package.json:207:    "handlebars": "^4.5.3",
```
So it may already be resolved here, I just need to figure out how to get this change into the docker image.

krassowski:
`package.json` tells you the minimum version that the package can work with. `yarn.lock` tells you the installed version. `package.json` is copied over to inform build system and extensions. Also `handlebars` is not really a runtime dependency, unless you are rebuilding lab/notebook using (deprecated) source extensions see https://github.com/jupyterlab/jupyterlab/issues/16698

mjpsyapse:
You are correct. Thank you again. We use twistlock to scan for vulnerabliities which is apparently giving a false positive for handlebars in this case.

mjpsyapse:
I'm not sure how you want to proceed, but if you want me to re-open and for me to do additional work, let me know.

krassowski:
I mean we can merge PRs which bump versions even if not strictly required - if this helps folks to have a cleaner board of security issues then why not, it's surely an improvement! I would just likely not spend my time backporting this to older versions.

jtpio (approved):
Thanks!

jupyterlab/staging/package.json
```diff
@@ -204,7 +204,7 @@
     "duplicate-package-checker-webpack-plugin": "^3.0.0",
     "fs-extra": "^10.1.0",
     "glob": "~7.1.6",
-    "handlebars": "^4.5.3",
+    "handlebars": "^4.7.8",
```
jtpio, line 207: Ah, just noticed we usually don't need to update the `jupyterlab/staging/package.json` file manually, as this is done during the release process.

So we should be able to undo this change and the one in `jupyterlab/staging/yarn.lock` too.

Issue #15143, opened 2023-09-21
`Contents.IDrive` wishlist


jtpio:
### Problem

Currently JupyterLab provides a way for third-party extensions to easily add extra source of contents.

### Proposed Solution

This is more of a meta-issue to track potential improvements and bug fixes to be done in JupyterLab to make the UX of working with drives more pleasant. Some existing issues:

- https://github.com/jupyterlab/jupyterlab/issues/14697
- https://github.com/jupyterlab/jupyterlab/issues/7755

The documentation could also be improved to provide more guidance and visibility on the drive feature to extension authors: 

- https://github.com/jupyterlab/extension-examples/issues/215

In downstream applications drives can allow for additional content via a third-party extension:

- JupyterLite: https://github.com/jupyterlite/jupyterlite/issues/1041
- Notebook 7: https://github.com/jupyter/notebook/issues/6665
- Access remote contents: https://github.com/davidbrochart/jupyterlab-remote-contents

### Additional context

Example drives:

- https://github.com/jupyterlab/jupyterlab_collaboration
- https://github.com/jupyterlab/jupyterlab-github/
- https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access

jtpio:
Feel free to edit the top comment or comment in this issue to provide more links to existing issues or external references.

trungleduc:
An accordion panel can be useful for multiple drives since users can easily drag and drop files between drives.

![files](https://github.com/jupyterlab/jupyterlab/assets/4451292/c32012ca-f20f-4631-96e9-cd9038680d07)

timkpaine:
Worth considering prior art too https://github.com/jpmorganchase/jupyter-fs

PR #16744, opened 2024-09-04
Add content providers

### LINKED ISSUE #16717: Allow a drive to support multiple document providers
### Problem

Currently, file contents can be loaded at the drive level. For instance, the default drive allows loading files using the Jupyter Server REST API. Jupyter Collaboration swaps the default drive with its own custom drive, so that it can use another protocol than the REST API (the Y protocol over WebSocket). But this is not ideal: what if one wants to use multiple protocols within the same drive? For instance, open this notebook in collaborative mode, and that huge binary file in non-collaborative mode. For the later, we could imagine a special protocol that would not load the whole file in memory, but only a part of it.

### Proposed Solution

It seems that another abstraction would be useful, maybe `docprovider`? One would register a `docprovider` for the REST API (the default), for the RTC API, etc. Each provider would advertise the type of files it is able to handle (maybe with a "rank" saying how well it is suited for this type of files), and the drive would then choose the best `docprovider` for opening a file. If several doc providers are equally able to open a file, the UI would ask the user which one they wish to use.

### Additional context

This would fix the issue with the `RTC:` prefix of Jupyter Collaboration's custom drive, since it wouldn't need a custom drive anymore. Instead, it would register its custom `docprovider`, that could be used in the default drive.

This idea was discussed with @SylvainCorlay, @afshin, @andrii-i and @3coins.

davidbrochart:
## References

- Closes #16717
- Together with https://github.com/jupyterlab/jupyter-collaboration/pull/345 will enable fixing highly requested https://github.com/jupyterlab/jupyter-collaboration/issues/202

## Code changes

This PR adds the `ContentProviderRegistry` and `RestContentProvider` classes. The default `Drive` now doesn't directly access files and directories using the Jupyter REST API , but through its content provider registry, which has a REST content provider by default. Extensions like `jupyter-collaboration` can register new content providers for the drive, for instance an RTC content provider.
Each content provider exposes the type of files it is able to handle through a regular expression matching the file extension, and a "score" indicating how well the content provider is suited for this type of files. The convention for the score is to use `1` when a content provider is able to open the file, and a greater number if it is "particularly good" at opening the file. The chosen content provider is the one which has the greatest score.
This change has the side effect of removing the `RTC:` prefix in `jupyter-collaboration`, since it won't need a custom drive anymore, but it will instead register an RTC content provider on the default drive.

## User-facing changes

None.

None.


davidbrochart:
I opened https://github.com/jupyterlab/jupyter-collaboration/pull/345 which is a POC that uses this PR to enable RTC on the default JupyterLab, by registering a content provider for the RTC procotol. Since no custom drive is created, files opened in RTC don't have the `RTC:` prefix.

davidbrochart:
[Failures](https://github.com/jupyterlab/jupyterlab/actions/runs/10738595322/job/29782508372?pr=16744) seem to be unrelated?

krassowski:
Given that we are still in beta for 4.3, I went ahead and proposed this to be included in 4.3.0. The rationale is that this enables fixing one of the most annoying bugs in jupyter-collaboration and server-side execution. For example, see the user engagement in https://github.com/jupyterlab/jupyter-collaboration/issues/202.

I think we should discuss this on the JupyterLab call on Wednesday. For now pinging @fcollonval and @Zsailer who earlier opened two PRs with the same goal of dropping the `RTC:` prefix:
- https://github.com/jupyterlab/jupyterlab/pull/16556
- https://github.com/jupyterlab/jupyterlab/pull/16141

to get their thoughts or at least sentiment on this PR.

packages/services/src/contents/index.ts
```diff
@@ -1080,6 +1099,28 @@ export class Drive implements Contents.IDrive {
...
+
+  static getContentProviderRegistry(): IContentProviderRegistry {
+    if (Drive._contentProviderRegistry === undefined) {
+      const contentProviderRegistry = new ContentProviderRegistry();
+      const restContentProvider = new RestContentProvider();
+      contentProviderRegistry.register(restContentProvider);
+      Drive._contentProviderRegistry = contentProviderRegistry;
+    }
+    return Drive._contentProviderRegistry;
+  }
```
krassowski, line 1116: Shall we move it to the non-exported `Private` namespace down in the file to match the existing code and ensure this is actually private?

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
     return extension;
   }
 }
+
+class ContentProviderRegistry implements IContentProviderRegistry {
+  constructor() {
+    this._providers = [];
+    this._drive = undefined;
+  }
+
```
krassowski, line 1584:
```suggestion
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+            score = extension.score;
+          }
+        }
+      }
+    }
+    return bestProvider;
+  }
+
+  private _providers: IContentProvider[];
+  private _drive: Contents.IDrive | undefined;
```
krassowski, line 1623:
```suggestion
  private _providers: IContentProvider[] = [];
  private _drive: Contents.IDrive | undefined = undefined;
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+  private _apiEndpoint: string;
+  private _drive: Contents.IDrive;
+}
+
+/**
+ * the content provider registry interface.
+ */
+export interface IContentProviderRegistry {
+  /**
+   * the drive of the content provider registry.
```
krassowski, line 1757:
```suggestion
   * The drive of the content provider registry.
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+    const baseUrl = this.drive.serverSettings.baseUrl;
+    return URLExt.join(baseUrl, this._apiEndpoint, ...parts);
+  }
+
+  private _apiEndpoint: string;
+  private _drive: Contents.IDrive;
+}
+
+/**
+ * the content provider registry interface.
```
krassowski, line 1753:
```suggestion
 * Registry of content providers.
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+  }
+
+  register(provider: IContentProvider): void {
+    if (this._drive !== undefined) {
+      provider.drive = this._drive;
+    }
+    this._providers.push(provider);
+  }
+
+  getProvider(filePath: string): IContentProvider | undefined {
```
krassowski, line 1603: What is the use case for matching providers by file extension?

In addition to file extension we _could_ use some of the metadata from `Contents.IFetchOptions`.

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+    let ext = filePath.split('.').pop();
+    if (ext == null) ext = '';
```
krassowski, line 1605:
```suggestion
    const ext = filePath.split('.').pop() ?? '';
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+    this._apiEndpoint = SERVICE_DRIVE_URL;
+  }
+
+  get drive(): Contents.IDrive {
+    return this._drive;
+  }
+
+  set drive(drive: Contents.IDrive) {
+    this._drive = drive;
+  }
```
krassowski, line 1640: Neither getter nor setter have side effects, this can be simplified down to:

```suggestion
  drive: Contents.IDrive;
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+   * Get a REST url for a file given a path.
+   */
+  private _getUrl(...args: string[]): string {
+    const parts = args.map(path => URLExt.encodeParts(path));
+    const baseUrl = this.drive.serverSettings.baseUrl;
+    return URLExt.join(baseUrl, this._apiEndpoint, ...parts);
+  }
+
+  private _apiEndpoint: string;
+  private _drive: Contents.IDrive;
```
krassowski, line 1749:
```suggestion
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1575,259 @@ namespace Private {
...
+   * A signal emitted when a file operation takes place.
+   */
+  readonly fileChanged: ISignal<this, Contents.IChangedArgs>;
+}
+
+/**
+ * The content provider extensions interface, consisting of the regural expression matching the file
+ * extension and the score indicating how good the provider is at handling this type of content.
+ */
+export type ContentProviderExtensions = { re: string; score: number }[];
```
krassowski, line 1833: I would suggest to define `IContentProviderExtension` interface and use it like `IContentProviderExtension[]`. Using interfaces is better for extension authors and for debugging in TS over types.

packages/services/src/contents/index.ts
```diff
@@ -254,6 +261,17 @@ export namespace Contents {
...
+    /**
+     * Register a SharedDocumentFactory.
+     *
+     * @param type Document type
+     * @param factory Document factory
+     */
+    registerDocumentFactory(
+      type: Contents.ContentType,
+      factory: SharedDocumentFactory
+    ): void;
```
krassowski, line 274: Adding a note that this is just exposing a function that was already implemented to simplify the code on the jupyterlab-collaboration side. Technically this is a breaking change, so maybe it should have an `?` added for now?

packages/services/src/contents/index.ts
```diff
@@ -477,6 +495,12 @@ export namespace Contents {
    * in the contents manager.
    */
   export interface IDrive extends IDisposable {
+    /**
+     * The content provider registry, consisting of all the
+     * content providers that this drive can use to access files.
+     */
+    readonly contentProviderRegistry: IContentProviderRegistry;
```
krassowski, line 502: Technically this adds a new required attribute, so has a potential to break existing extensions. Can we make it optional for now?

```suggestion
    readonly contentProviderRegistry?: IContentProviderRegistry;
```

That said, the risk is low because most extensions would just inherit from `Drive`.

krassowski (commented):
Thank you for working on it @davidbrochart! Some feedback on the code. I think it might be worth adding a simple test to ensure that swapping the provider works.

packages/services/src/contents/index.ts
davidbrochart: Wouldn't making `_contentProviderRegistry` private enough? `getContentProviderRegistry` must be public, so that extensions can register new content providers.
```ts
private static _contentProviderRegistry: IContentProviderRegistry | undefined = undefined;
```

krassowski: There is no `private static` in TypeScript, that's why we use `Private` non-exported namespaces :)

krassowski: Sorry, I was confused on `private static` availability. I guess that works too :+1:

packages/services/src/contents/index.ts
davidbrochart: A provider advertises the type of files it is able to handle through a list of extensions. For instance, an extension could provide access only for HDF5 files.
Yes `Contents.IFetchOptions` could be useful too.

davidbrochart:
> Thank you for working on it @davidbrochart! Some feedback on the code. I think it might be worth adding a simple test to ensure that swapping the provider works.

Thanks for the review @krassowski, I'll add a test.

Working on this, I was thinking that the default `Drive` should have been a token, that way it could have accepted a token for the content provider registry. Instead the registry is provided as a static attribute of `Drive`, so that JupyterLab extensions can register new content providers for it. But this is not very nice, because all other custom drives must implement this static attribute. But maybe the 2 PRs in [this comment](https://github.com/jupyterlab/jupyterlab/pull/16744#issuecomment-2340311286) will make that possible in the future?

davidbrochart:
I re-added the optional `sharedModelFactory` attribute on the `IDrive`, for backwards compatibility. If the drive provides one, it will be used. Otherwise, the one from the content provider is used, if any.

davidbrochart:
> It would also be worth taking the time to document the different between drives and contents providers

Where would that go? I don't see a section for drives in the JupyterLab documentation.

jtpio:
> Where would that go? I don't see a section for drives in the JupyterLab documentation.

That's probably because there isn't any at the moment. Tracked in https://github.com/jupyterlab/jupyterlab/issues/15143 via https://github.com/jupyterlab/extension-examples/issues/215 (and other linked issues).

Maybe a new page in https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html would work?

davidbrochart:
I'm a bit reluctant about adding some documentation for content providers. They are not extension points in the sense of a token. The first use-case is the removal of the `RTC:` prefix, which can only happen if we use content providers in the default `Drive`, which is not provided as a token. The result of this is that the content provider registry is provided as a static member of the `Drive` class, which is the only way an extension can register a content provider before the `Drive` is created. That works for this particular case, but do we want to document this for extension authors? I don't think so. I would be much more comfortable if we could provide a token for the content provider registry, which means the default drive must also be provided as a token.

packages/services/src/contents/index.ts
```diff
@@ -1080,6 +1108,29 @@ export class Drive implements Contents.IDrive {
...
+    }
+    return Drive._contentProviderRegistry;
+  }
+
+  get contentProviderRegistry(): IContentProviderRegistry {
+    const contentProviderRegistry = Drive.getContentProviderRegistry();
+    if (contentProviderRegistry.drive === undefined) {
+      contentProviderRegistry.drive = this;
+    }
+    return contentProviderRegistry;
```
krassowski, line 1133: Second thoughts: _should_ the registry be per-drive rather than static? Why?

I'm thinking that if we have a specialized network drive, for the use case of RTC it would not work right?

So instead of `Drive.getContentProviderRegistry()` the collaboration plugin would call `contentsManager.defaultDrive.getContentProviderRegistry()`?

We would need to expose the `defaultDrive` (currently a private `_defaultDrive`) but this is much smaller cost.

davidbrochart: But isn't it too late if we register a content provider once the drive is instantiated? At that time, it could already have been used to access a file system, right?
Also, how would the plugin require the contents manager?

krassowski: > But isn't it too late if we register a content provider once the drive is instantiated

Good point, I think we should work to ensure the RTC plugin gets loaded early. That said if you are calling `Drive.getContentProviderRegistry()` in RTC extension, it would have exactly the same timing so I don't think this is an argument against this approach.

> Also, how would the plugin require the contents manager?

`app.serviceManager.contents` is the `ContentsManager` class used by the application. Does that work for you?

davidbrochart: > Good point, I think we should work to ensure the RTC plugin gets loaded early.

That's why there should be relationships between dependencies, and a drive should require a content provider registry, ensuring the former is created after the latter. My understanding was that JupyterLab only "starts" after all plugins have been started, meaning that by the time the RTC plugin has started, it has registered its content provider in the default `Drive`, which is instantiated later. But maybe that's a false expectation?

> `app.serviceManager.contents` is the `ContentsManager` class used by the application.

Yes I could probably use that to access the drive, if we decide to register content providers on the drive instance rather than on the drive class. The downside is that they would not be registered on other instances, but I'm having a hard time understanding what this means.

krassowski: >  The downside is that they would not be registered on other instances, but I'm having a hard time understanding what this means.

I think this is an upside in the sense that if another extensions was using adding a drive (e.g. a network drive, similar to what RTC did, but with files actually outside of jupyter-server control) then installing collaboration would not would not break it. If we hijack all drives by registering collaboration on class level then existing drives which provide content from outside of jupyter-server would break.

davidbrochart: Makes sense, since drives and content providers are orthogonal. I'll make that change.

packages/services/src/contents/index.ts
```diff
@@ -1080,6 +1115,26 @@ export class Drive implements Contents.IDrive {
     this._apiEndpoint = options.apiEndpoint ?? SERVICE_DRIVE_URL;
     this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings();
+    const contentProviderRegistry = new ContentProviderRegistry();
+    const restContentProvider = new RestContentProvider();
```
krassowski, line 1119: If I am not missing something, it looks like the `Drive` no longer needs `_apiEndpoint` private member and we can instead pass it down to `RestContentProvider`:

```suggestion
    this.serverSettings =
      options.serverSettings ?? ServerConnection.makeSettings();
    const contentProviderRegistry = new ContentProviderRegistry();
    const restContentProvider = new RestContentProvider({
      apiEndpoint: options.apiEndpoint ?? SERVICE_DRIVE_URL
    });
```

davidbrochart: hmm `_apiEndpoint` is still used in `_getUrl()` in a bunch of places?

krassowski: Indeed. In that case I still think we should forward the configured API endpoint to `RestContentProvider`, right?

```suggestion
    this._apiEndpoint = options.apiEndpoint ?? SERVICE_DRIVE_URL;
    this.serverSettings =
      options.serverSettings ?? ServerConnection.makeSettings();
    const contentProviderRegistry = new ContentProviderRegistry();
    const restContentProvider = new RestContentProvider({
	  apiEndpoint: this._apiEndpoint
	});
```

davidbrochart: Good point. Either that or access the endpoint from the content provider through `this.drive.apiEndpoint` (if we make it public), or even `this.drive.getUrl()`. This is the reason why I set the content provider factory's `drive`, and it passes it on to each content provider. But I don't have a strong opinion, we could have content providers not keep a reference to their drive, and pass everything when we instantiate them. What do you think?

krassowski: I would prefer passing everything down during instantiation - this leads to smaller API surface and more predictable state.

davidbrochart: See https://github.com/jupyterlab/jupyterlab/pull/16744/commits/d228eb1af630891a9dccd40ae9d615e97f0ffea8 for a version where the REST content provider reference its drive to get the configuration.

krassowski: I think it would be better not to do that the way as in d228eb1. A a couple of reasons:
- it forces us to extend public API of drive by making `getUrl` public (maybe not the worst thing)
- it forces you to case `this.drive` which is `IDrive` initially to a `Drive` in an unsafe manner (this is if an extension author wanted to reuse the default `ContentProviderRegistry` implementation - although it is currently not exported - they could see a runtime error if they used it with a different `IDrive` implementation).
- it requires that the content provider registry sets the `drive` on each provider making making the logic flow more complex than it needs to be. This is because to register a content provider you the extension now already has to have access to the drive, so it could just pass it to content provider constructor.

What do you think about either:

```ts
new RestContentProvider({
   drive: this
});
```

or passing the `apiEndpoint` as proposed earlier?

This would allow to remove `drive` property from `IContentProviderRegistry` and from implementation.

krassowski: Then the `RestContentProvider` could require that drive is either of:
- `Drive` type or
- `IDrive & {getUrl: (...args[string]) => string}` interface (which we could name "IRestDrive")

davidbrochart: Yes, I removed the drive reference in content providers in https://github.com/jupyterlab/jupyterlab/pull/16744/commits/0430081c60d5e8a1e13aea69a45fe2560a2f7235. The configuration is passed when instantiated.

packages/services/src/contents/index.ts
```diff
@@ -1505,6 +1511,7 @@ export class Drive implements Contents.IDrive {
   private _apiEndpoint: string;
   private _isDisposed = false;
   private _fileChanged = new Signal<this, Contents.IChangedArgs>(this);
+  contentProviderRegistry: IContentProviderRegistry;
```
krassowski, line 1514: What do you think about narrowing the API a bit more by:

```suggestion
  /**
  * Content provider registry.
   * @experimental
   */
  readonly contentProviderRegistry: IContentProviderRegistry = new ContentProviderRegistry();
```

and moving it up to be with all the other public members?

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1578,233 @@ namespace Private {
...
+namespace RestContentProvider {
+  export interface IOptions {
+    apiEndpoint: string;
+    serverSettings: ServerConnection.ISettings;
+    fileChanged: ISignal<Contents.IDrive, Contents.IChangedArgs>;
+  }
+}
+
+/**
+ * Registry of content providers.
```
krassowski, line 1731: Should we add tag the APIs as experimental for now?

```suggestion
 * Registry of content providers.
 * @experimental
```

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1578,233 @@ namespace Private {
...
+  /**
+   * Get a content provider for the given file path.
+   *
+   * @param filePath - The file path for which to get a content provider.
+   */
+  getProvider(filePath: string): IContentProvider | undefined;
+}
+
+/**
+ * The content provider interface.
```
krassowski, line 1750:
```suggestion
 * The content provider interface.
 * @experimental
```

packages/services/src/contents/index.ts
```diff
@@ -1080,6 +1115,14 @@ export class Drive implements Contents.IDrive {
...
     this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings();
+    const contentProviderRegistry = new ContentProviderRegistry();
+    const restContentProvider = new RestContentProvider({
+      apiEndpoint: this._apiEndpoint,
+      serverSettings: this.serverSettings,
+      fileChanged: this.fileChanged
+    });
+    contentProviderRegistry.register(restContentProvider);
+    this.contentProviderRegistry = contentProviderRegistry;
```
krassowski, line 1125: See the other suggestion on moving instantiation of `ContentProviderRegistry` to the member definition (and making it readonly).

```suggestion
    const restContentProvider = new RestContentProvider({
      apiEndpoint: this._apiEndpoint,
      serverSettings: this.serverSettings,
      fileChanged: this.fileChanged
    });
    this.contentProviderRegistry.register(restContentProvider);
```

krassowski (commented):
Thank you, looks great now! A few final suggestions on documentation and I think we can merge and release

davidbrochart:
Thanks again for the review @krassowski.
I guess marking the API as experimental means we don't write official documentation yet? I think it would be wise, until we make sure it works fine with jupyter-collaboration. Then we could start advertising it.

jtpio:
> > since it won't need a custom drive anymore, but it will instead register an RTC content provider on the default drive.
> 
> It would also be worth taking the time to document the different between drives and contents providers, and what extension authors should implement. They might be asking some questions such as:
> 
> - Why there are now two ways for providing additional content?
> - Should the preferred way be using a contents provider?
> - What is the purpose of drives now?
> - Are drives going to be deprecated?
> 
> And also check the list of drive related issues if anything in there might be relevant to this discussion: https://github.com/jupyterlab/jupyterlab/issues/15143

Can you please provide answers to these questions from the comment above? It's still not clear what the long term relation between drives and contents providers is.

packages/services/src/contents/index.ts
```diff
@@ -1080,8 +1115,21 @@ export class Drive implements Contents.IDrive {
     this._apiEndpoint = options.apiEndpoint ?? SERVICE_DRIVE_URL;
     this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings();
+    const restContentProvider = new RestContentProvider({
+      apiEndpoint: this._apiEndpoint,
+      serverSettings: this.serverSettings,
+      fileChanged: this.fileChanged
+    });
+    this.contentProviderRegistry.register(restContentProvider);
```
fcollonval, line 1123: Instead of registering the rest API content provider as any other providers, should we set a default provider and use it as fallback to avoid filepath analysis when no other providers are defined?

For reference we are doing something like that in the document registry: https://github.com/jupyterlab/jupyterlab/blob/main/packages/docregistry/src/registry.ts#L41

davidbrochart:
> * Why there are now two ways for providing additional content?

Content providers add flexibility, as you can use multiple protocols within the same drive. Previously, you had to use a new drive for a new protocol.

> * Should the preferred way be using a contents provider?

Not necessarily, only if you need the flexibility mentioned above. Otherwise, the preferred way is still using a custom drive.

> * What is the purpose of drives now?

They have the same purpose as before, and they are still needed e.g. if the data lives in a separate source.

> * Are drives going to be deprecated?

No, content providers need a drive.

bollwyvl:
A thought from the `jupyter_server` call today:

> As a server extension author, I'd like to advertise my REST API as a ContentsProvider

For a use case of a python-only serverextension that wishes to advertise it can handle these kinds of things, it might be possible to ship an `overrides.d/my-extension.json`. 

Consider something that allows for exposing the attachment tree of a PDF:

```yaml
{
  "@jupyterlab/contents-extension": {                                # or wherever
    "contentProviders": {
      "pdf-folder-contents": {
        "title": "PDF Folder",                                       # to be show in _Open with_
        "description": "View/edit PDF attachements",
        "icon": "data:image/svg+xml;utf8,...",                       # or local URL
        "documentation": "https://pdf-folder-contents.rtfd.io/",     # or local URL
        "patterns": {
          "(?<path>^/.*?)(?<name>.*?\\.pdf)/?(?<rest>[^/].*$)": {    # or maybe rfc6570
            "redirect": "/api/pdf-folder-contents/{path}/{name}/{{rest}",
            "methods": ["GET", "PUT", "POST"],
            "rank": 10,
            # ... and maybe a custom `icon`, `documentation`, etc.
          }
        }
      }
    }
  }
}
```

The serverextension implementation would then be (testably) on the hook for answering the [`api/contents` contract](https://github.com/jupyter-server/jupyter_server/blob/v2.14.2/jupyter_server/services/api/api.yaml#L82), but _wouldn't_ have to worry about any further typescript concerns.

davidbrochart:
> Also did you try with drives other than collaboration to make sure they don't break?

I tested [jupyterlab-remote-contents](https://github.com/davidbrochart/jupyterlab-remote-contents) and everything went fine.

fcollonval:
@echarles and I discuss about this and taken into account the [comment](https://github.com/jupyterlab/jupyterlab/pull/16744#issuecomment-2346731271) of Nick, we are seeing one potential issue and a need to expand the API covered by the content provider.

- The potential issue: should the file viewer be aware of the content provider? For example if the content provider is handling or not transferring slice of a document model, how can the viewer leverage or not the feature?

- Currently the content provider only covers `get` and `save`. But it would be more appropriate to cover all methods of `IDrive` (this is highlighted by the suggestion of Nick with the supported actions):

https://github.com/jupyterlab/jupyterlab/blob/f73aa5a7c963cad10267e1e4ff807c306d6fc516/packages/services/src/contents/index.ts#L497

And it raises a question should all methods be provided by all content providers or should there be a default provider for not-implemented methods?

> For example overriding checkpoint handling for RTC would make sense - even rename and deletion if the storage is not a real filesystem but a virtual one for example.

jtpio:
Given the number of comments, the discussion points still to be resolved before merging this PR, and the fact that it is marked as `enhancement`, how about postponing to the `4.4.0` milestone, to not delay the `4.3.0` release too much?

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
+}
+
+/**
+ * The content provider extension interface.
+ */
+export interface IContentProviderExtension {
+  /**
+   * The regular expression matching the file extension
+   */
+  re: string;
```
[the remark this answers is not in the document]
davidbrochart: Actually I'm wondering if the job of identifying the file type is not already done by the time the content provider is called. We receive `IFetchOptions` that have a `ContentType` and a `FileFormat`, which should identify the MIME type? So maybe the content provider should only advertise the MIME types it is able to handle. And when users choose "Open with ...", they explicitly choose a MIME type, which means they can choose a particular content provider, not just relying on the one which has the best "score".

davidbrochart:
> * The potential issue: should the file viewer be aware of the content provider? For example if the content provider is handling or not transferring slice of a document model, how can the viewer leverage or not the feature?

Yes that is something I was thinking about. Say we have a bigger-than-memory file that can be handled by a content provider, which can access chunks of the file. The content provider doesn't know which chunks to access, since it's just called with the file path and no extra information regarding e.g. which parts of the file. But isn't it a limitation of the drives themselves?

krassowski:
> Given the number of comments, the discussion points still to be resolved before merging this PR, and the fact that it is marked as `enhancement`, how about postponing to the 4.4.0 milestone, to not delay the 4.3.0 release too much?

The 4.3.0 is blocked on other issues as of now, and really this is a bug-fix (in the sense that major enhancements in 4.3.0 are around server-side execution and RTC; I've got feedback from many places that the RTC prefix issue is blocking adoption of these features). That said, it seems fair to postpone it if this remains the last blocker standing. In my view the question would then be whether this PR can be ready for final review today or tomorrow (as I expect we would want to make the final decision on other blockers on Wednesday meeting).

jtpio:
Catching up on the latest comments in this PR, it looks like the main issue this PR is trying to solve is the handling of the `RTC:` prefix in the `jupyter-collaboration` extension, similar to what https://github.com/jupyterlab/jupyterlab/pull/16141 tried to achieve previously. Does that mean we give up trying to fix the handling of drives (including the handling the prefixes), and other issues listed in https://github.com/jupyterlab/jupyterlab/issues/15143?

It looks like https://github.com/jupyterlab/jupyterlab/pull/16141 or similar would also fix that issue in `jupyter-collaboration`, and remove the need for custom content provider for now, since `jupyter-collaboration` would be able to provide its custom default drive and do anything it wants with its content?

Overall being able to register additional content providers sounds like a good idea and would allow for more flexibility :+1:
 
However being able to do it via a token (for example `IContentProvider`) would likely make extension code easier to reason about, and avoid patterns like the one used in the current state of https://github.com/jupyterlab/jupyter-collaboration/pull/345, which may look a bit awkward at first:

```ts
Drive.getContentProviderRegistry().register(rtcContentProvider);
```

Before settling on adding this new API on a core service, it would be worth investigating again making the `ServiceManager` more easily customizable, as recently explored in https://github.com/jupyterlab/jupyterlab/pull/16556. Although as mentioned in https://github.com/jupyterlab/jupyterlab/pull/16556#issuecomment-2352329904 and previous discussions, the strong requirements on avoiding breaking changes and the initial design of the `ServiceManager` add some constraints that may require adding two phases of plugin registration.

krassowski:
> Does that mean we give up trying to fix the handling of drives (including the handling the prefixes), and other issues listed in https://github.com/jupyterlab/jupyterlab/issues/15143?

I cannot speak for authors of the PRs that attempted to fix them, but it appears that there has not been a recent progress on it, but the work has not been abandoned. I think these are just more difficult issues to solve and may require a major release.

> and avoid patterns like the one used in the current state of https://github.com/jupyterlab/jupyter-collaboration/pull/345, which may look a bit awkward at first:
>
> ```js
> Drive.getContentProviderRegistry().register(rtcContentProvider);
> ```

This is no longer the case - based on review it was changed to:

```js
app.services.defaultDrive.getContentProviderRegistry()
```

as the registry is on per-drive basis. I don't think there should be a token for the content provider registry as this is property of the drive, and if a drive has a token (the default one does not) then it could be interacted via requesting such token.

> Before settling on adding this new API on a core service, it would be worth investigating again making the ServiceManager more easily customizable, as recently explored in https://github.com/jupyterlab/jupyterlab/pull/16556. Although as mentioned in https://github.com/jupyterlab/jupyterlab/pull/16556#issuecomment-2352329904 and previous discussions, the strong requirements on avoiding breaking changes

I just worry that this will not happen in a long time. To accommodate the RTC mode JupyterLab 4 introduced many breaking changes, but existing users cannot benefit from them yet, while the issues introduced by the transition to RTC-enabling architecture remain. What makes this PR very appealing as-is is that it does solve the issue at hand without introducing any breaking changes, and without requiring to get a consensus on the entire plugin architecture (which can still happen independently).

krassowski:
@davidbrochart do you have time to address the @fcollonval's comments today? If not, I may be able to help

davidbrochart:
@krassowski That would be great if you could help, thanks!
One thing I was no sure about is whether we should use MIME types or handle raw file names (see [this comment](https://github.com/jupyterlab/jupyterlab/pull/16744#discussion_r1760673763)).

krassowski:
Sorry, got sick - will not be able to push today nor attend the frontends meeting. Still think it would be great if we could include it.

krassowski:
Sorry, got sick - will not be able to push today nor attend the frontends meeting. Still think it would be great if we could include it.

bollwyvl:
> a custom `IContentsManager`

This doesn't sound like it would compose for multiple extensions.

jtpio:
> This doesn't sound like it would compose for multiple extensions.

Maybe not as of now, but it would already give more flexibility to extensions that want to manage contents differently. Then we could take some time to define more extension points (via plugins?), for example to add more providers as proposed in this PR.

The point was that I'm not sure if we should "rush" this PR into 4.3, or give it more time so it can be considered for 4.4 with more cases in mind.

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
+ * Registry of content providers.
+ * @experimental
+ */
+export interface IContentProviderRegistry {
+  /**
+   * Add a content provider into the registry.
+   *
+   * @param provider - The content provider to register.
+   */
+  register(provider: IContentProvider): void;
```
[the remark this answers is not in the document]
krassowski: No strong opinion, we `register`:
- routes in Router
- widgets in PropertyInspector
- themes in ThemeManager
- LSP features (but that's on me!)
- message hooks
- status bar items

But yes, we always use `IDisposable` (or at least should always use it).

packages/services/src/contents/index.ts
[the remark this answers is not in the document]
krassowski: 1. Agree to rename `re` → `pattern`
2. Not sure if we should special-case `extensions[]`, it would complicate the code - need to think about it a bit more
3. Agree that we should consider what happens if user chooses a different viewer.  Strictly users choose a viewer (widget factory) rather than anything related to the API or MIME type (strictly the MIME type stays the same). So we go from `WidgetFactory → ModelFactory → {contentType, fileFormat}`.

- `ContentType` is just `file`/`notebook`/`directory`
- `FileFormat` is `'json' | 'text' | 'base64'`

I'm thinking that when I have my special CSV viewer that can access small chunks of 1TB CSV file, I actually don't care if the extension is `.csv`, `.tsv` or different; all I care about is if user chooses my smart CSV viewer, they get rerouted to the content provider which will stream the contents in chunks. We could extend [`IModelFactory`](https://github.com/jupyterlab/jupyterlab/blob/058d05f720447737274dbec9377af7f6c6c3e7fa/packages/docregistry/src/registry.ts#L1173) so that it can pass some kind of a token saying "Use content provider X" (if available).

I imagine that all smart viewers to work with existing constraints would effectively return `json` with description of the file (say how many rows, how many columns), content hash (to allow logic detecting changes to trigger), and instructions for the widget how to request individual data chunks.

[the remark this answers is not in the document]
krassowski: Maybe we should remove pattern matching altogether and explicitly support two use cases:
- a) register a new content provider for use with all files (without having control over the widgets)
- b) register a new content provider for use with a widget which request it (in an extension having control over that widget)

We may want to allow paramterizing the `Contents.IModel` returned by the content provider for type safety.

(a) is the RTC use case.

packages/services/src/contents/index.ts
```diff
@@ -1080,8 +1115,21 @@ export class Drive implements Contents.IDrive {
     this._apiEndpoint = options.apiEndpoint ?? SERVICE_DRIVE_URL;
     this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings();
+    const restContentProvider = new RestContentProvider({
+      apiEndpoint: this._apiEndpoint,
+      serverSettings: this.serverSettings,
+      fileChanged: this.fileChanged
```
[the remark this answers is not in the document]
krassowski: RTC needs to emit `fileChanged` when the file is saved from server-sid. It can currently do this because it is swapping the drive. Once it is just provisioning the content via content provider it would no longer be able to emit `fileChanged`.

I agree that the implementation should be improved - instead of passing down the signal to emit from provider, we should allow content providers to emit `fileChanged` and proxy that in the drive.

packages/services/src/contents/index.ts
```diff
@@ -254,6 +261,17 @@ export namespace Contents {
...
+    /**
+     * Register a SharedDocumentFactory.
+     *
+     * @param type Document type
+     * @param factory Document factory
+     */
+    registerDocumentFactory?(
+      type: Contents.ContentType,
+      factory: SharedDocumentFactory
+    ): void;
```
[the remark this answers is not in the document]
krassowski: AFAIK this was always overridden by providing custom typing on collaboration extension side, and this just formalizes it on the lab side but I will let @davidbrochart confirm.

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
   }
 }
+
+class ContentProviderRegistry implements IContentProviderRegistry {
+  register(provider: IContentProvider): void {
+    this._providers.push(provider);
+  }
+
+  getProvider(filePath: string): IContentProvider | undefined {
+    const ext = filePath.split('.').pop() ?? '';
```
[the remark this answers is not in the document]
krassowski: Can you elaborate on which parts you think we should reuse?

packages/services/src/contents/index.ts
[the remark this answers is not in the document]
krassowski: Done in df6a62ff02ae4ed314b98200cec34a532e13d124

packages/services/src/contents/index.ts
```diff
@@ -686,7 +717,11 @@ export class ContentsManager implements Contents.IManager {
    */
   getSharedModelFactory(path: string): Contents.ISharedFactory | null {
     const [drive] = this._driveForPath(path);
-    return drive?.sharedModelFactory ?? null;
+    if (drive?.sharedModelFactory) {
+      return drive!.sharedModelFactory;
+    }
+    const provider = drive?.contentProviderRegistry?.getProvider(path);
+    return provider?.sharedModelFactory ?? null;
```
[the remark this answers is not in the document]
krassowski: Done in 7712bf083cfc6cad23e19199a569ab42e605b32f

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
     return extension;
   }
 }
+
+class ContentProviderRegistry implements IContentProviderRegistry {
+  register(provider: IContentProvider): void {
+    this._providers.push(provider);
+  }
+
+  getProvider(filePath: string): IContentProvider | undefined {
```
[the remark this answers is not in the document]
krassowski: Done in ad6ae159c2535c05a274f19dc17463d8080f9c36. This require to define a default provider explicitly, otherwise the list could be empty if one was not registered.

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
+    let bestProvider: IContentProvider | undefined = undefined;
+    for (let provider of this._providers) {
+      for (let extension of provider.extensions) {
+        const re = new RegExp(extension.re);
+        if (ext.match(re) !== null) {
+          if (extension.score > score) {
+            bestProvider = provider;
+            score = extension.score;
+          }
+        }
```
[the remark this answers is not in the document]
krassowski: Applied in 8e6b6df9e5ad487b8014dfd76d457fea49cfad70.

packages/services/src/contents/index.ts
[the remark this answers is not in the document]
krassowski: Made it return `IDisposable` in 0a2734e9cb699002a75e366c9b6e96d1b9b3f371

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
+   * @param provider - The content provider to register.
+   */
+  register(provider: IContentProvider): void;
+
+  /**
+   * Get a content provider for the given file path.
+   *
+   * @param filePath - The file path for which to get a content provider.
+   */
+  getProvider(filePath: string): IContentProvider | undefined;
```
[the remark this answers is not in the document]
krassowski: ad6ae159c2535c05a274f19dc17463d8080f9c36

packages/services/src/contents/index.ts
```diff
@@ -686,7 +717,11 @@ export class ContentsManager implements Contents.IManager {
    */
   getSharedModelFactory(path: string): Contents.ISharedFactory | null {
     const [drive] = this._driveForPath(path);
-    return drive?.sharedModelFactory ?? null;
+    if (drive?.sharedModelFactory) {
+      return drive!.sharedModelFactory;
```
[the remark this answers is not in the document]
krassowski: Fixed in 7712bf083cfc6cad23e19199a569ab42e605b32f

packages/services/src/contents/index.ts
```diff
@@ -1571,3 +1582,235 @@ namespace Private {
...
+export interface IContentProviderExtension {
+  /**
+   * The regular expression matching the file extension
+   */
+  re: string;
+
+  /**
+   * The score indicating how good the provider is at handling thie type of content.
+   */
+  score: number;
```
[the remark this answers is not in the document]
krassowski: Done in 8e6b6df9e5ad487b8014dfd76d457fea49cfad70

packages/services/src/contents/index.ts
krassowski: Done in ad6ae159c2535c05a274f19dc17463d8080f9c36

krassowski (commented):
Addressed most parts of the review by @fcollonval

packages/services/src/contents/index.ts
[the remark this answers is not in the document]
krassowski: The use case of "Open with" was also independently brought up by @bollwyvl.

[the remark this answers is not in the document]
krassowski: Examples: open a file inside a PDF (because PDFs are actually a folder).

> Personally I received research data this way.

andreytaboola:
Any update on this fix, I see its stuck now.  This is highly requested one I will appreciate pushing it to closest release

krassowski:
@andreytaboola do you mean removing RTC prefix, or content providers?

For context, this PR was discussed at the Jupyter Frontends meeting along https://github.com/jupyterlab/jupyterlab/pull/16794 last week. Either of them would enable removing the RTC prefix. The discussion among attendees focused on the https://github.com/jupyterlab/jupyterlab/pull/16794 not allowing for composition while this PR supporting composition and having a potential to enable better support for custom widgets (but it is not there yet!). That said, the attendance was low presumably due to PyData Paris taking place so we were not able to reach even an informal quorum.

krassowski:
I wanted to extract one review comment to the main discussion for visibility:

> I agree that we should consider what happens if user chooses a different viewer.  Strictly users choose a viewer (widget factory) rather than anything related to the API or MIME type (strictly the MIME type stays the same). So we go from `WidgetFactory → ModelFactory → {contentType, fileFormat}`.
> 
> - `ContentType` is just `file`/`notebook`/`directory`
> - `FileFormat` is `'json' | 'text' | 'base64'`
> 
> I'm thinking that when I have my special CSV viewer that can access small chunks of 1TB CSV file, I actually don't care if the extension is `.csv`, `.tsv` or different; all I care about is if user chooses my smart CSV viewer, they get rerouted to the content provider which will stream the contents in chunks. We could extend [`IModelFactory`](https://github.com/jupyterlab/jupyterlab/blob/058d05f720447737274dbec9377af7f6c6c3e7fa/packages/docregistry/src/registry.ts#L1173) so that it can pass some kind of a token saying "Use content provider X" (if available).
> 
> I imagine that all smart viewers to work with existing constraints would effectively return `json` with description of the file (say how many rows, how many columns), content hash (to allow logic detecting changes to trigger), and instructions for the widget how to request individual data chunks.

Currently the use case discussed above is not supported in this PR because the content providers are chosen based on extension alone. I think the final review of the API of this PR should focus on whether will we be able to extend the API for choosing content provider to support the use case described above.

I think merging this PR or https://github.com/jupyterlab/jupyterlab/pull/16794 is not mutually exclusive and ultimately we would probably merge a form of both. For now it feels like merging this PR into 4.3.0 would be much less risk than merging #16794 in but I am happy to be proven otherwise.

andreytaboola:
> @andreytaboola do you mean removing RTC prefix, or content providers?
> 
> For context, this PR was discussed at the Jupyter Frontends meeting along #16794 last week. Either of them would enable removing the RTC prefix. The discussion among attendees focused on the #16794 not allowing for composition while this PR supporting composition and having a potential to enable better support for custom widgets (but it is not there yet!). That said, the attendance was low presumably due to PyData Paris taking place so we were not able to reach even an informal quorum.

Mainly I am interested in removing RTC prefix, thank you

krassowski:
@jtpio I agree with most what you wrote, just not not sure about this

> All existing third-party extensions won't be impacted because the new tokens don't exist at the moment

If I am not mistaken, In some very specific use cases like adding an extension for telemetry, the telemetry extension could now be disabled by installing another extension, say "block-telemetry-at-service-manager-level" if https://github.com/jupyterlab/jupyterlab/pull/16794 was merged. Previously if the deployment locked status of a telemetry plugin, and required a browser which is locked out of modifying scripts (say in a controlled research environment) then disabling the telemetry was pretty hard, and definitely required a traceable user interaction.

The scenario above is hypothetical and while a real concern for some deployments - rather niche. The point is, allowing to swap the server manager changes some assumptions that other extensions could have had about the environment they are running in.

Further, the service manager bits include guards for CSFR. If we are allowing to swap these, then the deployment managers have to either audit all extensions, or block all extensions, because a fault in extension can lead to new vulnerabilities _in core APIs_ (of course any new extension extends the surface area for vulnerabilities). Nothing here would prevent us from merging https://github.com/jupyterlab/jupyterlab/pull/16794 in the future, we just need to think through these scenarios and communicate it well in advance to the stakeholders.

Now, for this PR I still believe that we can narrow down the API (remove all the extension matching stuff) to just bare  minimum to solve RTC prefix and then build on this in future iterations. What do you think?

krassowski:
@jtpio @davidbrochart @fcollonval what do you think about:
- removing the extension matching code; the content provider registered by RTC extension would match all files, and
- defining the precedence of content providers registered without matching options as the lowest (so in the RTC case it's content provider would have the least options) - this way we can add other ways to match/select content providers without deciding the details in this PR, and
- documenting this as experimental, and
- merging into 4.3

jtpio:
> The point is, allowing to swap the server manager changes some assumptions that other extensions could have had about the environment they are running in.

I don't see how https://github.com/jupyterlab/jupyterlab/pull/16794 would change the way plugins are currently installed and managed. Any malicious could easily patch the current `ServiceManager` (it's just JavaScript in the end) and do whatever it wants. Also the extension manager still shows a disclaimer, which is relevant to all third-party plugins:

image

But an administrator could then lock the new service manager plugins by default. Mentioning that in the docs would definitely be a good idea :+1:

packages/services/src/contents/index.ts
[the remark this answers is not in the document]
fcollonval: The extraction of the extension. 

We already have discrepancies with code mirror. We should use the same helper to ensure consistency.

JasonWeill:
What are people thinking about merging either this change or #16794 into `main` after 4.3.0 is cut (for consideration in 4.4.0)? The `RTC:` prefix is related to some user-facing bugs that are very noticeable with the collaboration extension installed.

krassowski:
I would suggest to follow the plan https://github.com/jupyterlab/jupyterlab/pull/16744#issuecomment-2385967726 (with last point being "merge into 4.4').

krassowski:
## Use cases for content providers

### RTC

Content Provider needed to:
- provide a custom `sharedModelFactory` for all files
   - except for binary files?
- override `ContentsManager` methods
   - `get(path, options)` - to provide content from the `sharedModel` rather than from the jupyter-server Contents API
   - `save(path, options)` - to prevent sending `save` to jupyter-server Contents API as saving is handled by RTC, but this is strictly no longer needed since https://github.com/jupyterlab/jupyterlab/pull/16900
- be able to emit `fileChanged` to update file if collaborator changes it

### Streaming CSV widget

Use case: right click on a large CSV in file browser, choose "Open with" → "Large CSV Viewer", allowing to view files larger than the browser/jupyter-server can handle at once.

Real world applications: genome browsers, geospatial browsers

Content Provider needed to:
- provide a custom `sharedModelFactory` for files opened with this widget to allow fetching content selectively
- probably override `ContentsManager.get()` so that the full content is never fetched as it would freeze the browser

Entry point:
- passing a custom content provider in `DocumentRegistry.IWidgetFactory`

### Jupytext

Jupytext does not change the frontend contents API, but has its own server-side ContentsManager subclass. Here I explore what would be needed if someone wanted to make a jupytext a frontend-only extension, and if content providers mechanism could help.

Use case:
- right click on a text file with `.md` extension, choose "Open with" → "Jupytext Notebook"
- synchronise notebook to/from multiple formats

Requirements:
- override contents methods (by routing them to a custom content manager on backend), e.g.:
  - `get()` so that it returns something that quacks like a Notebook
  - `save()` to save paired notebook representations at once
  - `new_untitled()` to allow creating notebooks with `ext`
  - `rename_file()` to allow renaming all paired notebooks
- should compose well with RTC

Entry point:
- passing a custom content provider in `DocumentRegistry.IWidgetFactory`

## Limitations specific to this PR (as of today)

### Only `get` and `save` methods in Content Provider

Only `get` and `save` methods are mirrored in content providers (reflecting RTC use case at the time when this PR was written) - but drives mirror all the `ContentsManager` methods

Maybe the content provider should have the same API as `IDrive` excluding:
   - drive prefix
   - content provider registry

### No "Open with" support

This PR does not allow to choose the provider based on widget factory (yet).

### Redundancy of path/extension routing logic

As pointed by @fcollonval, `DocumentRegistry` already implements a dynamically dispatch depending on path/extension. We should not duplicate that logic or if we do we should harmonize the two implementation.

## Open questions around Content Providers

### Composition

Content Providers idea does not help with composition. This is why implementing jupytext using content providers may not be a good idea if simultaneous RTC support is desired; instead `jupytext` might need a way to transform content before it is loaded into RTC/when RTC saves.

### Moving files between drives

The content provider would need to be reset. This can be addressed in a follow-up.

### One content provider per drive

Content provider will usually work in context of a single drive (if we implement streaming in jupyter-server extension and add a content provider to interface with it, it will not provision content from Google Drive), but `IWidgetFactory` then needs to be only shown in "Open with" for drives which support required content provider.

How can we support multiple drives for "Open with"? Should we register content providers on drives, or should content provider be able to provision content on multiple drives?

IMO it should be one content provider registered per each drive; maybe multiple content providers could be associated with the same token, like in:


```ts
defaultDrive.contentProviders.register({
  id: 'streaming-csv',  // or a token object instead of ID
  provider: new StreamingContentProvider(defaultDrive),
})
googleDrive.contentProviders.register({
  id: 'streaming-csv',  // or a token object instead of ID
  provider: : new StreamingGoogleDriveProvider(googleDrive),
})
// then when registering "streaming CSV" widget, in `DocumentRegistry.IWidgetFactory`:
{
  contentProviderId: 'streaming-csv'
}
```

### Provision of directories

Another potential difference between `ContentProvider` and `Drive` is that drive influences the listing in the file browser, but it is not clear if `ContentProvider` should do that too. In fact, when registering with `DocumentRegistry.IWidgetFactory` it would not have impact on the file browser at all.

Technical note: file browser calls `ContentsManager`/`Drive` directly, but for opening files the operations are routed through `DocumentRegistry.IContext` object where we could swap the content provider for the `DocumentRegistry.IWidgetFactory` scenario.

### Competition between multiple extensions and delayed initialization

@fcollonval pointed out in @Zsailer's PR [#16141](https://github.com/jupyterlab/jupyterlab/pull/16141) which aimed to simply allow extensions to swap the `defaultDrive` that such a change could lead to bad side effects if any extension can swap it on runtime. While we could make sure that this change is only allowed once by setting a simple counter flag, it would still mean that on initial load a different drive is used than after loading the RTC plugin.

https://github.com/jupyterlab/jupyterlab/pull/16794 by @jtpio (building on @fcollonval and @afshin work) explores a way to addresses this by splitting the extension loading into two steps.

**A new possibility opens up with** `DocumentRegistry.IWidgetFactory`: we could have RTC register "Open with" handlers for notebooks and for files which would include the appropriate content provider. This also has the benefit of making sure that binary (e.g. PNG files) would not be open collaboratively until these are explicitly supported. RTC could simply register these "editors"/"viewers" as default for Notebook/Text File types. Of course the caveat is that the default "editors"/"viewers" could still be accessed (unless we hide them) potentially confusing users (so we probably want a way to hide them?).

echarles:
> https://github.com/jupyterlab/jupyterlab/pull/16794 by @jtpio (building on @fcollonval and @afshin work) explores a way to addresses this by splitting the extension loading into two steps.

Can you please add the security concerns I raised regarding this and that have not been answered so far. Thank you.

krassowski:
ar@echarles those are security concerns regarding https://github.com/jupyterlab/jupyterlab/pull/16794, not this PR, right? Above I did not mean to make a comprehensive comparison between these two approaches, just to explain how this PR can present an alternative to #16794 (and possibly one that we have not thought about before). Or do you have similar concerns about content providers?

echarles:
@krassowski My security concerns are indeed on https://github.com/jupyterlab/jupyterlab/pull/16794.

While reading your previous post, I was willing to make sure those were not forgotten in the discussion.

The last time I looked more in depth at this PR, I found it quite OK and was in favor to work with this approach. This does not prevent to make it strong in terms of security and features, e.g. regarding the file types, the editors.... But my understanding is that the options are there to make this PR strong enough today and stronger is further iterations.

jtpio:
> Can you please add the security concerns I raised regarding this and that have not been answered so far. Thank you.

They were answered in this comment: https://github.com/jupyterlab/jupyterlab/pull/16794#issuecomment-2362876132 (which includes a link to a previous answer to the same question).

echarles:
> They were answered in this comment: https://github.com/jupyterlab/jupyterlab/pull/16794#issuecomment-2362876132 (which includes a link to a previous answer to the same question).

I saw that and remember finding them not valid enough. I will try to take that over asap.

jtpio:
> I saw that and remember finding them not valid enough. I will try to take that over asap.

I'm more than happy to discuss that more. Maybe we should continue on the other PR or the related issue to not derail the conversation in the present PR. You are also welcome to join the weekly Jupyter Frontends meeting if you want to have a more in-depth conversation about this.

krassowski:
On the call yesterday @afshin brought up that instead of passing a new option (say `contentProviderId`) we could use fully qualified URIs to specify content provider as a kind of a protocol. The default content provider is the REST API, and with "default" drive it could be represented as:

```
rest://default:Untitled.ipynb
```

or


```
rest://Untitled.ipynb
```

When RTC is used, we could have it as:

```
rtc://Untitled.ipynb
```

If someone created an RTC adaptor for google drive or S3-hosted files these could then register an RTC content provider on respective drives and be represented like:

```
rtc://GoogleDrive:Untitled.ipynb
rtc://S3:Untitled.ipynb
```

Similarly for `streaming` providers. In this case we might consider if "content provider" is the bast naming, or if we should call it a "content protocol" or similar.

I like the conceptual clarity that this brings. I do not like that we are not able to easily enforce type validation with this approach though (the old `path` arguments becomes `uri` now but both would be typed as just `string`).

afshin:
Thanks for writing this comment @krassowski! I think that the introduction new providers and also the fact that we are moving into a multi-drive world makes some unified way of understanding file paths and what to do with them more pressing. But I do recognize it's a very big change.

krassowski:
> I do not like that we are not able to easily enforce type validation with this approach though 

Template literal types in newer TypeScript versions allow something like:

```ts
// contentProvider, drive, path
type FullyQualifiedPath = `${string}://${string}:/${string}`;
```

Now, adapting this is harder as we would need to replace `path: string` in public interfaces with incompatible `path: FullyQualifiedPath`; realistically, we would allow either just `string` path or `FullyQualifiedPath` which would get collapse to just `string`, defeating the point of having any typing at all.

krassowski:
Looking at the implementation, I am tempted to use the protocol prefix idea as this simplifies the API surface.

However, technically this would be a breaking change as so far there was no restriction on using using `://` as a path fragment. If someone has a drive with file paths that include `://` then these would no longer open.

On practical level this also seems complex to implement given all the implicit assumptions about paths.

krassowski:
I opened a new PR to replace this one: https://github.com/jupyterlab/jupyterlab/pull/17092

krassowski:
Closing as superseded by #17092

Issue #16660, opened 2024-08-09
Amend best practices for CSS variables definitions in light of performance lessons learned


krassowski:
> What I would ultimately like to do is to stop putting things into `:root` when they do not need to be there. This is because performance suffers doubly:
> - from browser having too many variables declared in a scope (this is a small cumulative effect)
> - from them being set on global scope when a local one would do (this is a problem especially in Chrome, see https://issuetracker.google.com/issues/40676722)
> 
> [...] these variables are used only once in the entire CSS codebase but have a non-obvious cost which only becomes apparent when you work with large notebooks and get into the weeds of browser performance profiling. I spent a lot of time on it with https://github.com/jupyterlab/ui-profiler and more low-level tools and saw the `:root` group of variables show up quite often as a potential culprit of UI freezes.
> 
> _Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/16656#discussion_r1711064127_

> [... ] having everything scoped out in `:root` is quite useful for collaborating with multiple people from different backgrounds because, from my perspective, it makes the CSS look like just a configuration file. Anyway, I agree with you.
>
>[... ] an issue to discuss the reorganization of the variable files for the various themes. The `RTC user specific colors` and `Vega extension styles` variable sets can also be refactored, for example, in my perspective. 
>
>I would like also to suggest adding a note to the [documentation](https://jupyterlab.readthedocs.io/en/stable/developer/css.html#css-variables) about this organization (there is actually already a point about this, but I think it can be extended): _critical_ variables must be scoped in `:root`, the rest in their respective parent classes (with one example). 
> 
> _Originally posted by @joaopalmeiro in https://github.com/jupyterlab/jupyterlab/pull/16656#discussion_r1711266674_

Yes, that makes sense. To clarify, I am not sure about moving anything out of the `:root` right now as that could break existing themes due to specificity, but for new additions which are tightly scoped to a specific component (like in this case) it seems that there is a risk of performance degradation if we keep adding things to `:root` but no historic constraints, so these are maybe better properly scoped.

Anyways, collecting some benchmarks to make sure we know at which points it becomes a problem, and updating the docs with suggestion to add things into scoped blocks as you mention for things like the completer, `RTC user specific colors` and `Vega extension styles` which you mention is a good idea.

Issue #15933, opened 2024-03-06
Support for additional File Browser columns


bollwyvl:
### Problem

The columns shown in the File Browser can't be trivially extended, or configured very much.

### Proposed Solution

- add an extension point for registering new file browser columns
- use the extension point for existing columns
- make columns re-orderable (in the x-axis)
- allow more columns to participate in the filter box
- expose additional columns in context menu
- make columns draggable to adjust the x-axis rank
  - store in
    - workspace?
    - settings? 

### Additional context

- add `FileBrowser.addColumn(options: IColumnOptions)` where:

```ts

export interface IColumnOptions {
   key: string;
   headerLabel: () => string;
   headerTitle: () => string;
   headerAriaLabel: (model: Contents.IModel) => string;
   cellTitle: (model: Contents.IModel) => string;
   cellAriaLabel: (model: Contents.IModel) => string;
   headerElement: () => HTMLElement;
   cellElement: (model: Contents.IModel) => HTMLElement;
   // overloadable in settings
   rank: number;
   visible: boolean;
   // optional
   updateRequested?: ISignal; // if not handled by IModel
   compare?: (a: IModel, b: IModel) => number;  // implies sortability
   search?: (query: string, a: IModel) => number;  // implies searchability
}
```
- rebuild the file 
- add `filebrowser:*-column` commands that take `args: {key}`

krassowski:
Also:

https://github.com/jupyterlab/jupyterlab/blob/56c7d1b9a4421b512722171a65aa909da5500f71/packages/ui-components/src/components/table.tsx#L41-L48

bollwyvl:
> [table.tsx](https://github.com/jupyterlab/jupyterlab/blob/56c7d1b9a4421b512722171a65aa909da5500f71/packages/ui-components/src/components/table.tsx#L41-L48)

I think it was tried to replace the file tree with `react` at some point, and at ~100s of listings, which is not entirely uncommon for a folder listing, it fell down. The newer `ui-components` table implementation _might_ be better (e.g. windowed rendering, which carries its own weaknesses), but this might be one of the cases where a general purpose table implementation is not going to work.

[...]

PR #16751, opened 2024-09-06
Modifies filebrowser to use CSS grid

### LINKED ISSUE #16712: File browser columns are not consistently positioned
## Description

In the screen shot below, note that filenames which end in ellipses (…) have inconsistently positioned horizontal starting points.

image

## Expected behavior

All columns should have a consistent starting point on the X axis.

## Context

- Browser and version: Firefox 129.0.1
- JupyterLab version: 4.3.0b0

JasonWeill:
## References

Fixes #16712.

## Code changes

Migrates the filebrowser to use the CSS grid instead of flexbox to position columns. Resizing columns and sorting columns should still work.

Screenshots may need to be updated.

## User-facing changes

As was the case in older versions of JupyterLab, filenames, icons, modified datestamps, and file sizes, are lined up in a grid.

Should be none.

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1164,14 +1164,22 @@ export class DirListing extends Widget {
       columnWidths[colName] = newSize;
     }
 
-    // Update the first visible column's element's parent's grid columns.
-    visibleColumns[0].element.parentElement!.style.gridTemplateColumns =
-      columnOrder
-        .map(col => {
-          return `[${col}] ${columnWidths[col]}`;
-        })
-        .join(' ');
+    // Update the header's template columns.
+    const newGridTemplateColumns = columnOrder
+      .map(col => {
+        return `[${col}] ${columnWidths[col]}`;
+      })
+      .join(' ');
+    const headerElement = visibleColumns[0].element.parentElement!;
+    headerElement.style.gridTemplateColumns = newGridTemplateColumns;
 
+    // Update the body's template columns.
+    // Find the jp-DirListing-content element adjacent to the element above.
+    (
+      headerElement.parentElement?.getElementsByClassName(
+        'jp-DirListing-content'
+      )[0] as HTMLElement
+    ).style.gridTemplateColumns = newGridTemplateColumns;
     this._updateModifiedStyleAndSize();
 
     // Refresh sizes on the per item widths
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -2779,7 +2779,7 @@ export namespace DirListing {
       id: 'name' as const,
       className: NAME_ID_CLASS,
       itemClassName: ITEM_NAME_COLUMN_CLASS,
-      minWidth: 60,
+      minWidth: 75,
       resizable: true,
       sortable: true,
       caretSide: 'right'
@@ -2788,7 +2788,7 @@ export namespace DirListing {
       id: 'last_modified' as const,
       className: MODIFIED_ID_CLASS,
       itemClassName: ITEM_MODIFIED_CLASS,
-      minWidth: 60,
+      minWidth: 75,
       resizable: true,
       sortable: true,
       caretSide: 'left'
@@ -2797,7 +2797,7 @@ export namespace DirListing {
       id: 'file_size' as const,
       className: FILE_SIZE_ID_CLASS,
       itemClassName: ITEM_FILE_SIZE_CLASS,
-      minWidth: 60,
+      minWidth: 75,
       resizable: true,
       sortable: true,
       caretSide: 'left'
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1136,17 +1136,42 @@ export class DirListing extends Widget {
       }
     }
 
-    // Write to DOM
+    // Compose the grid columns.
+    const columnOrder = [
+      'jp-filebrowser-checkbox',
+      'jp-filebrowser-name',
+      'jp-filebrowser-resize-name-modified',
+      'jp-filebrowser-modified',
+      'jp-filebrowser-resize-modified-size',
+      'jp-filebrowser-filesize'
+    ];
+    let columnWidths: Record<string, string> = {
+      'jp-filebrowser-checkbox': '24px',
+      'jp-filebrowser-name': 'auto',
+      'jp-filebrowser-resize-name-modified': '5px',
+      'jp-filebrowser-modified': 'auto',
+      'jp-filebrowser-resize-modified-size': '5px',
+      'jp-filebrowser-filesize': 'auto'
+    };
+
     for (const column of visibleColumns) {
       const size = this._columnSizes[column.id];
-      const newSize = size === null ? '' : size + 'px';
-      const varName =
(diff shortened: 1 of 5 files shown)


packages/filebrowser/src/listing.ts
```diff
@@ -1138,7 +1138,13 @@ export class DirListing extends Widget {
...
     for (const column of visibleColumns) {
       const size = this._columnSizes[column.id];
-      column.element.style.width = size === null ? '' : size + 'px';
+      const newSize = size === null ? '' : size + 'px';
+      const varName =
+        column.className.replace(RegExp(/^jp-id-/), '--jp-filebrowser-') +
+        '-column-width';
+
+      // Set at the root.
+      document.documentElement.style.setProperty(varName, newSize);
```
krassowski, line 1147: This will not work in general - there may be more than one file browser, e.g. when using dialog which include a file selector.

Also, we should not add new variables to the `:root` anyway due to performance reasons https://github.com/jupyterlab/jupyterlab/issues/16660

packages/filebrowser/style/base.css
```diff
@@ -127,17 +132,20 @@
...
 
 .jp-DirListing-header {
-  flex: 0 0 auto;
-  display: flex;
-  flex-direction: row;
-  align-items: center;
+  display: grid;
+
+  /* TODO: Deduplicate with content grid */
+  grid-template-columns:
```
krassowski, line 142: While `grid-template-columns` may be *the* solution for now, it would be much nicer if we could use `grid-auto-columns` to allow for column reordering (together with setting size explicitly on elements). This or we need to generate `grid-template-columns` in JS.

packages/filebrowser/src/listing.ts
krassowski: Maybe we should set `grid-template-columns` directly via JS on the browser instance: on the header and on the content. This would resolve the TODO around de-duplication, remove the CSS variables, and prevent resizing columns of all filebrowsers at once.

packages/filebrowser/style/base.css
```diff
@@ -127,17 +132,20 @@
...
+    [checkbox] auto
```
krassowski, line 143: I wonder if we should prefix the names of areas like `jp-filebrowser-checkbox` to avoid conflicts.

krassowski (changes requested):
The hover looks funny:

image

Resizing is now broken:
- column headers do not downsize properly
- column may be placed outside of browser

bad-ux

huh

Resizing column in one filebrowser resizes them all

resizes-multiple

I had an almost working grid implementation locally, I will try to open a counter PR on Monday.

krassowski:
Here is the counter PR:
- https://github.com/jupyterlab/jupyterlab/pull/16765

I believe the approach in #16765 is better because it allows to:
- add additional columns in extensions (this PR would have forced extensions to redefine `grid-template-columns` so each time two extensions were trying to add a column, it would result in a conflict) https://github.com/jupyterlab/jupyterlab/issues/15933
- rearrange column order without modifying CSS (this PR would again force setting of `grid-template-columns`, which is 
not a show stopper but not ideal)

packages/filebrowser/style/base.css
```diff
@@ -223,22 +228,35 @@
...
+    [modified] minmax(var(--jp-filebrowser-modified-column-width), auto)
+    [resize-modified-size] 5px
+    [file-size] minmax(var(--jp-filebrowser-filesize-column-width), auto);
+}
+
+.jp-DirListing-content ul {
+  /* Use all the same columns as in the top-level grid */
+  display: grid;
+  grid-template-columns: subgrid;
+  grid-column: 1 / -1;
```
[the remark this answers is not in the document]
JasonWeill: Renaming this to `ul.jp-DirListing-content`, which matches the element in question, produces undesirable visuals. I deleted this rule completely, and there's no change. Good catch!

JasonWeill:
@krassowski I tried modifying this code locally to set the style on the first visible column's parent's element (i.e., the header) to have a particular set of columns, but the problem you observed in the "import workspace" dialog persists: changing the width of columns inside the import dialog causes the file browser dialog's columns to change as well.

I'm going to be out of office for the next week (will be physically around, but not checking email or updating PRs much). I can review updates for sure starting the week of September 23. Thanks for your help on this and your own PR!

JasonWeill:
@krassowski Thanks for your patience! I updated this PR to update column widths for each column in the header grid and the contents grid. I do notice one open bug that using the ⬇️ key to move through the grid may scroll the entire grid. I'm out of office on Friday, September 27, and will be back on Monday. I'd appreciate if you could take a look at this revised PR when you get a chance!

JasonWeill:
The filebrowser contents grid should no longer scroll when you use the arrow keys to move through it.

packages/filebrowser/style/base.css
```diff
@@ -127,17 +127,20 @@
...
+  display: grid;
+
+  /* TODO: Deduplicate with content grid */
+  grid-template-columns:
+    [jp-filebrowser-checkbox] 24px
+    [jp-filebrowser-name] auto
+    [jp-filebrowser-resize-name-modified] 5px
+    [jp-filebrowser-modified] auto
+    [jp-filebrowser-resize-modified-size] 5px
+    [jp-filebrowser-filesize] auto;
```
krassowski, line 143: Can we remove this from CSS and only define in JS?

packages/filebrowser/src/listing.ts
```diff
@@ -1142,11 +1136,54 @@ export class DirListing extends Widget {
       }
     }
 
-    // Write to DOM
```
krassowski, line 1145: Let's keep this comment :) It is important to separate reading and writing to DOM steps for performance reasons.

packages/filebrowser/src/listing.ts
```diff
@@ -1142,11 +1136,54 @@ export class DirListing extends Widget {
...
+        'jp-filebrowser-'
+      );
+      columnWidths[colName] = newSize;
     }
+
+    // Update the header's template columns.
+    const newGridTemplateColumns = columnOrder
+      .map(col => {
+        // Allow the columns containing file information to grow.
+        const newWidth = col.match(/^jp-filebrowser-(name|modified|filesize)$/)
```
krassowski, line 1171: We should move it to specification of columns, rather than hard-code names here. The common request is to allow adding columns in extensions, but hard-coding logic based on column name makes it harder.

krassowski:
Apart from the general request to make sure this does not hard-code any logic (so that extensions can add new columns smoothly), the high-level decision between merging this vs https://github.com/jupyterlab/jupyterlab/pull/16765 comes down to:
- do we want think it is good API to require extension authors to also add a CSS rule when adding a new column to the filebrowser (so that they set the `grid-area`); would they also need to add a CSS rule for resize handle? To me this does not seem extensible, hence I lean towards my https://github.com/jupyterlab/jupyterlab/pull/16765
- do we think that extension authors should be prohibited from adding a dummy row which spans the entire grid? If so this PR is better, as https://github.com/jupyterlab/jupyterlab/pull/16765 allows more flexibility; but I think that we may want to have this ability (of adding a dummy row spanning all columns), e.g. to show messages as we do in shortcuts table:
   image
- is one or the better for performance with large directories? We need a benchmark to settle this one.

PR #16764, opened 2024-09-09
Fix compatibility with server-side execution extensions

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/347
- and probably https://github.com/datalayer/jupyter-server-nbmodel/issues/26 but ideally it would also make a change as in https://github.com/jupyter-server/jupyverse/pull/435.

CC @davidbrochart

## Code changes

The data type assumed by the implementation changed from an plain dictionary to Map in:
- https://github.com/jupyterlab/jupyterlab/pull/16498

However, this already was a public API so I we should keep this compatible.

```diff
- this._outputs.add(output.toJSON());
+ this._outputs.add('toJSON' in output ? output.toJSON() : output);
```

## User-facing changes

Serve-side execution and related features work again.

None

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -37,10 +37,11 @@ Shared model
 ~~~~~~~~~~~~
 
 The outputs set on the shared cell model are now expected to be wrapped
-in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
-Further, the ``"text"`` entry can now be specified as an ``Array<string>``
+in the ``Y.Map`` objects rather than provided as plain objects
+(or pycrdt ``Map`` objects rather than dictionaries when set on the backend).
+Further, the ``"text"`` entry must now be specified as an ``Array<string>``
 object for outputs of ``"stream"`` type, allowing for better performance.
-The use of dictionaries is deprecated and will stop working in a future version.
+The use of plain objects is deprecated and will stop working in a future version.
 For reference, see PRs:
 
 - https://github.com/jupyterlab/jupyterlab/pull/16498
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -38,6 +38,8 @@ Shared model
 
 The outputs set on the shared cell model are now expected to be wrapped
 in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
+Further, the ``"text"`` entry can now be specified as an ``Array<string>``
+object for outputs of ``"stream"`` type, allowing for better performance.
 The use of dictionaries is deprecated and will stop working in a future version.
 For reference, see PRs:
 
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -32,6 +32,19 @@ breaking change when defining fixture values that are array of objects.
 
 See the `Playwright 1.46.0 release notes <https://github.com/microsoft/playwright/releases/tag/v1.46.0>`_ for more information.
 
+
+Shared model
+~~~~~~~~~~~~
+
+The outputs set on the shared cell model are now expected to be wrapped
+in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
+The use of dictionaries is deprecated and will stop working in a future version.
+For reference, see PRs:
+
+- https://github.com/jupyterlab/jupyterlab/pull/16498
+- https://github.com/jupyter-server/jupyter_ydoc/pull/241
+
+
 JupyterLab 4.1 to 4.2
 ---------------------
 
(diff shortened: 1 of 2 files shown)


docs/source/extension/extension_migration.rst
```diff
@@ -32,6 +32,19 @@ breaking change when defining fixture values that are array of objects.
...
+Shared model
+~~~~~~~~~~~~
+
+The outputs set on the shared cell model are now expected to be wrapped
+in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
+The use of dictionaries is deprecated and will stop working in a future version.
+For reference, see PRs:
+
+- https://github.com/jupyterlab/jupyterlab/pull/16498
+- https://github.com/jupyter-server/jupyter_ydoc/pull/241
```
krassowski, line 45: @davidbrochart is this correct? Or should we make this an implementation detail and fix this on `jupyter_ydoc` level by modifying the setter of `outputs` to do the wrapping for the API user? It may be hard because I think cells are just `YMap` for now, but I guess for sake of better typing we might want to have a subclass with custom setters - is this possible?

For reference I also opened https://github.com/jupyter-server/jupyter_ydoc/issues/269 to track typing enhancements.

davidbrochart: This is correct, but not complete. The reason each output is now a `Y.Map<any>` is that if the output is of type `"stream"`, its `"text"` entry is of type `Y.Array<string>`. This is also why it's difficult to type an output more precisely than `Y.Map<any>`: the type is dynamic, as it depends on the output's type which is only known at run-time. I agree that we could at least type it as a union of two types, one corresponding to a `"stream"` output type and the other to other types of outputs.

docs/source/extension/extension_migration.rst
```diff
@@ -32,6 +32,19 @@ breaking change when defining fixture values that are array of objects.
 
 See the `Playwright 1.46.0 release notes <https://github.com/microsoft/playwright/releases/tag/v1.46.0>`_ for more information.
 
+
+Shared model
+~~~~~~~~~~~~
+
+The outputs set on the shared cell model are now expected to be wrapped
+in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
```
krassowski, line 40:
```suggestion
in the pycrdt ``Map`` objects rather than provided as plain dictionaries.
Further, the `"text"` entry can now be specified as an `Array<string>`
object for outputs of `"stream"` type, allowing for better performance.
```

docs/source/extension/extension_migration.rst
```diff
@@ -32,6 +32,21 @@ breaking change when defining fixture values that are array of objects.
...
```
davidbrochart, line 40: Should we mention `pycrdt` here? It seems that we are talking from the frontend point of view, so maybe we should mention `Y.Map`?
```suggestion
in the ``Y.Map`` objects rather than provided as plain objects.
```

docs/source/extension/extension_migration.rst
```diff
@@ -32,6 +32,21 @@ breaking change when defining fixture values that are array of objects.
...
+Further, the ``"text"`` entry can now be specified as an ``Array<string>``
```
davidbrochart, line 41:
```suggestion
Further, the ``"text"`` entry must now be specified as an ``Array<string>``
```

docs/source/extension/extension_migration.rst
```diff
@@ -32,6 +32,21 @@ breaking change when defining fixture values that are array of objects.
...
+object for outputs of ``"stream"`` type, allowing for better performance.
+The use of dictionaries is deprecated and will stop working in a future version.
```
davidbrochart, line 43:
```suggestion
The use of plain objects is deprecated and will stop working in a future version.
```

packages/cells/src/model.ts
```diff
@@ -854,7 +854,10 @@ export class CodeCellModel extends CellModel implements ICodeCellModel {
           if ('insert' in outputsChange) {
             // Inserting an output always results in appending it.
             for (const output of outputsChange.insert!) {
-              this._outputs.add(output.toJSON());
+              // For compatibility with older ydoc where a dictionary,
```
davidbrochart, line 857:
```suggestion
              // For compatibility with older ydoc where a plain object,
```

docs/source/extension/extension_migration.rst
krassowski: Hmm... I think both. But yeas, it makes sense to use `Y.Map` notation as this should be clear that it refers to yjs/pycrdt stack.

krassowski:
```suggestion
in the ``Y.Map`` objects rather than provided as plain objects
(or pycrdt ``Map`` objects rather than dictionaries when set on the backend).
```

My point is that there is usage of documents similar to server-side execution in the wild.

davidbrochart (approved):

krassowski:
Thanks for the review!

Issue #16659, opened 2024-08-08
Only enlarge file name column when expanding file browser side panel


paulsonak:
## Description

In JupyterLab 4.1.8 and below, when you resize the file browser, only the Name column width changes. But in 4.2.X the Name and Last Modifiied/Modified columns both expand equally. This results in difficulty reading long file names since the Modified column takes up ~half the available space.

### 4.1.8 behavior:
<img width="784" alt="Screenshot 2024-08-08 at 3 58 18 PM" src="https://github.com/user-attachments/assets/903255dc-5769-4980-9701-10ddfb18fa1e">

### 4.2.X behavior:
<img width="784" alt="Screenshot 2024-08-08 at 3 59 29 PM" src="https://github.com/user-attachments/assets/1f463310-fbe3-4d5e-9c34-0f27b76179c3">

## Reproduce
1. Expand the file browser on left side
2. Click on the divider between workspace and browser and drag right
3. Note the size of the name column and the modified column both expand

I noticed this because I use the feature all the time to double check which file I need to read into my notebook.

## Expected behavior

Previously, only the name column expanded when dragging outward. I would like this behavior to be restored, or optionally add the ability for column size adjustment inside the file browser.

## Context

- Operating System and version: MacOS arm64 Sonoma
- Browser and version: Chrome Version 127.0.6533.100 (Official Build) (arm64)
- JupyterLab version: 4.2.X

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
        /Users/apaulson/atomsci-venv/bin
        /Users/apaulson/atomsci-venv/bin
        /Users/apaulson/.sdkman/candidates/grails/current/bin
        /Library/Frameworks/Python.framework/Versions/3.9/bin
        /opt/homebrew/bin
        /opt/homebrew/sbin
        /usr/local/bin
        /System/Cryptexes/App/usr/bin
        /usr/bin
        /bin
        /usr/sbin
        /sbin
        /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin
        /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin
        /var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
        /opt/X11/bin
        /Applications/VMware Fusion.app/Contents/Public
        /Users/apaulson/.orbstack/bin

sys.path:
        /Users/apaulson/atomsci-venv/bin
[...]

PR #16765, opened 2024-09-09
Use `inline-grid` for browser column positioning

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/16712
- Fixes https://github.com/jupyterlab/jupyterlab/issues/16659 - it does not implement the request literally, but when the file browser is resized it modifies the size of columns by the `grow` factor; we can adjust it further, but for now I would propose:
  - name: 3
  - last modified: 1
  - file size: 0.5
- Partially supersedes https://github.com/jupyterlab/jupyterlab/pull/16751 (we may still want to reuse the bit which removes the initial positioning using flexbox)

## Code changes

- Adds `grow` attribute to column description
- Uses `inline-grid` layout for listing

## User-facing changes

- Better positioning of items in the file browser on narrow sizes
- Resizing the sidebar causes growth of the name column, three times smaller growth of the last modified column and 6 times smaller growth of the file size column

This (as much as any change to file browser unfortunately) might break jupyterlab-unfold.


JasonWeill:
@krassowski Thank you so much for reviewing #16751 and publishing this PR! I'm happy to close my PR or rewrite it if we merge this one in.

When I had all three of the major columns (Name, Modified, Size) displayed in the file browser, I was surprised to see File Size change in size when I dragged the resizer between Name and Modified. I only expected Name to change in size. In macOS's Finder and Windows Explorer, when dragging the slider between two columns, only the column to the left of the slider changes in size.

JasonWeill:
Also, while your column headers look better than in #16751, I think they would look better with a little more padding on the left or right.

image

image

jtpio:
It looks like https://github.com/jupyterlab/jupyterlab/pull/16751 was going to help with https://github.com/jupyter/notebook/pull/7423 to set default sizes for the file browser columns in Notebook, as the defaults could have been defined in CSS directly.

Does this PR provide a way to set default sizes for the filebrowser columns from downstream distribution and / or extensions?

krassowski:
> When I had all three of the major columns (Name, Modified, Size) displayed in the file browser, I was surprised to see File Size change in size when I dragged the resizer between Name and Modified. I only expected Name to change in size. In macOS's Finder and Windows Explorer, when dragging the slider between two columns, only the column to the left of the slider changes in size.

Just for clarity, did you mean that you expected Name **and** Modified to change in size? Otherwise the last column would no longer touch the edge of the filebrowser and create blank space, right?

> Also, while your column headers look better than in #16751, I think they would look better with a little more padding on the left or right.
> 
> image
>
> image

This is pretty hard to improve on. I was annoyed by it too and will have another look.

krassowski:
This PR uses separate CSS inline grid for each row and sets width on each item (each cell in each row for each column).

Using a single grid for the entire listing, similar to proposed in https://github.com/jupyterlab/jupyterlab/pull/16751 might be better performance-wise, as we would not need to update item sizes for each cell, but just once for the entire grid. This has implications for the renderer so I would say that we should settle this before releasing.

krassowski:
I do not see a significant performance improvement in Chrome by switching to a single grid. The patch I tried on this PR:

<details>

```diff
diff --git a/packages/filebrowser/src/listing.ts b/packages/filebrowser/src/listing.ts
index e13b357810..1c34bdafec 100644
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1187,7 +1187,9 @@ export class DirListing extends Widget {
     if (this.isVisible) {
       const items = this._items;
       if (items.length !== 0) {
-        this.updateNodes(this._sortedItems, this._items, true);
+        const content = DOMUtils.findElement(this.node, CONTENT_CLASS);
+        const template = visibleColumns.map(column => `[${column.id}] ${Math.round(this._columnSizes[column.id] ?? 0)}px\n`).join('');
+        content.style.gridTemplateColumns = template;
       }
     }
 
@@ -2990,12 +2992,8 @@ export namespace DirListing {
         }
         const createElement = this.itemFactories[column.id];
         const element = createElement();
+        element.style.gridColumn = column.id;
         node.appendChild(element);
-
-        if (columnsSizes) {
-          const size = columnsSizes[column.id];
-          element.style.width = size + 'px';
-        }
       }
 
       return node;
@@ -3134,7 +3132,8 @@ export namespace DirListing {
       if (checkboxWrapper && !showFileCheckboxes) {
         node.removeChild(checkboxWrapper);
       } else if (showFileCheckboxes && !checkboxWrapper) {
-        const checkboxWrapper = this.createCheckboxWrapperNode();
+        const checkboxWrapper = this.itemFactories.is_selected();
+        checkboxWrapper.style.gridColumn = 'is_selected';
         nameColumn.insertAdjacentElement('beforebegin', checkboxWrapper);
       }
 
@@ -3143,6 +3142,7 @@ export namespace DirListing {
         node.removeChild(modified);
       } else if (showModified && !modified) {
         modified = this.itemFactories.last_modified();
+        modified.style.gridColumn = 'last_modified';
         nameColumn.insertAdjacentElement('afterend', modified);
       }
 
@@ -3151,6 +3151,7 @@ export namespace DirListing {
         node.removeChild(fileSize);
       } else if (showFileSize && !fileSize) {
         fileSize = this.itemFactories.file_size();
+        fileSize.style.gridColumn = 'file_size';
         (modified ?? nameColumn).insertAdjacentElement('afterend', fileSize);
       }
 
@@ -3247,19 +3248,6 @@ export namespace DirListing {
       modifiedStyle?: Time.HumanStyle,
       columnsSizes?: Record<DirListing.IColumn['id'], number | null>
     ): void {
-      if (columnsSizes) {
-        for (const column of columns) {
-          const element = DOMUtils.findElement(node, column.itemClassName);
-          if (!element) {
-            continue;
-          }
-          const sizeSpec = columnsSizes[column.id];
-          const newWidth = sizeSpec === null ? '' : sizeSpec + 'px';
-          if (newWidth !== element.style.width) {
-            element.style.width = newWidth;
-          }
-        }
-      }
       let modified = DOMUtils.findElement(node, ITEM_MODIFIED_CLASS) as
         | HTMLElement
         | undefined;
diff --git a/packages/filebrowser/style/base.css b/packages/filebrowser/style/base.css
index a91e639ee7..289387ce04 100644
--- a/packages/filebrowser/style/base.css
+++ b/packages/filebrowser/style/base.css
@@ -238,6 +238,7 @@
   overflow: auto;
   background-color: var(--jp-layout-color1);
   overflow-x: hidden;
+  display: grid;
 }
 
 .jp-DirListing-content mark {
@@ -264,11 +265,11 @@
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
-  display: inline-grid;
-  grid-auto-columns: max-content;
   width: 100%;
   box-sizing: border-box;
-  grid-auto-flow: column;
+  display: grid;
+  grid-template-columns: subgrid;
+  grid-column: 1 / -1;
 }
 
 .jp-DirListing-checkboxWrapper {
```

</details>

krassowski:
Also tried adding `height: 25px;` and `content-visibility: auto;` on `.jp-DirListing-item` with `inline-grid`. This enables browser-native lazy rendering, but it still is not smooth enough when resizing columns when in `node_modules` of JupyterLab.


So we have:
- single `grid` with subgrids which _should_ work fast as browsers _should_ be able to layout it quickly (maybe browsers do not have fast paths for subgrids yet?)
- multiple `inline-grid`s (one for each row) which allows to easily enable lazy rendering which _should_ make rendering faster (but it still is not as fast as I would expect when resizing)
- maybe we should try to use the old good `table` layout instead? This might be faster.

Now, for extensibility `inline-grid` may have an edge in the sense that renderer can add a pseudo row below and above with custom sizing easily.

JasonWeill:
I was reading through this 2020 post by @smhigley about when to use a grid and when to use a table: https://sarahmhigley.com/writing/grids-part1/

Sarah writes:

> If the primary purpose is for the user to consume and understand tabular information, then it leans towards being a table. This is true even if it contains many links or buttons.
> …
> If the primary purpose is to provide an interface for the user to edit, manipulate, or otherwise interact with the content it presents, then it leans towards being a grid. Even if most of the cells are static content.

Currently, the file browser is not used to edit, manipulate, or interact with the contents of a directory; it exists as a place to open a file (so that it opens in the main work area) or to spawn a context menu for an item. Semantically, by the above recommendations, I think a table would make sense. I'm not sure whether it would be better for the application's performance, though.

krassowski:
The reason why we do not get sufficient speed boost when using `content-visibility: auto` with `inline-grid` in a folder with 1000s of files might be because of the overhead of `content-visibility` in Chrome's implementation as per https://stackoverflow.com/questions/66661497/why-isnt-content-visibilityauto-working-in-this-simple-example.

JasonWeill:
As I'm testing out your and my PRs, #16659 by @paulsonak seems to stand out. When I enlarge the left panel showing the file browser, I, too, want to expand **only** the name column. I don't expect the other columns to enlarge in that situation.

JasonWeill:
I noticed that when I sort the file browser by "Name", the triangle icon appears on the right:

image

… but for "Modified" and "File Size", it's on the left:

image

image

In addition, the "up" arrow for file size shows files in _descending_, not _ascending_, order of file size:

image

smhigley:
I think I'd probably lean towards recommending grid semantics here, for a couple reasons:
- the primary purpose is interactive, and all items in the table/grid are interactive
- there can be quite a lot of rows, and making it a table would add a ton of tab stops

Table also isn't a bad semantic choice though!

Neither should meaningfully affect performance. The ARIA roles aren't related to the CSS `grid` and `inline-grid` display values, despite the name being the same. It looks like the ARIA grid vs. table choice might be a bit unrelated to this specific PR 😅

(I'll pop back out now, just barged in b/c I saw I was tagged, and I do quite appreciate jupyter notebook)

krassowski:
> I noticed that when I sort the file browser by "Name", the triangle icon appears on the right: [...] … but for "Modified" and "File Size", it's on the left:

This has always been the case (try it on a released version) and reflects that the "name" column is left-aligned whereas "modified" and "file size" are right-aligned, so it shows up where there is space. I think this is good thing!

> In addition, the "up" arrow for file size shows files in descending, not ascending, order of file size:

Good catch, but again this is unrelated to this PR :)

> It looks like the ARIA grid vs. table choice might be a bit unrelated to this specific PR 

I agree here and thank you for your advice!

JasonWeill:
Opened #16779 regarding the sort reversal. PR #16780 fixes it.

krassowski:
@JasonWeill any changes requested here?

jtpio:
According to the UI tests report, it looks like the "Modified" is not right aligned in the column header anymore?

[file-browser-modified-alignment-difference.webm](https://github.com/user-attachments/assets/7e65c4aa-1dde-48a4-99df-ec6dd306b9cf)

JasonWeill:
I've been revising #16751, but if you can rebase this to the tip of `main`, I'm OK to consider this for inclusion in 4.3.0 to fix the release blocker issue.

jtpio:
Just merged `main` here. Curious to see if the misaligned "Modified" header mentioned in https://github.com/jupyterlab/jupyterlab/pull/16765#issuecomment-2357703302 will be fixed by https://github.com/jupyterlab/jupyterlab/pull/16816.

jtpio:
Also, is the consensus to now focus on this PR instead of https://github.com/jupyterlab/jupyterlab/pull/16751? If so, should https://github.com/jupyterlab/jupyterlab/pull/16751 then be closed to avoid confusion on what should be reviewed?

jtpio:
Merged `main` one more time to fix the conflict with https://github.com/jupyterlab/jupyterlab/pull/16804 which got merged in the meantime.

jtpio:
After getting the latest updates from other PRs, it looks like the "Modified" header is still misaligned:

[jupyterlab-filebrowser-modified-header.webm](https://github.com/user-attachments/assets/a88c4fc4-d5f7-48d2-84c7-9df47575339a)

I guess we still want to have the header be right-aligned, unless the new alignment is intentional?

krassowski:
It IS right-aligned; the screnshot does not do it justice have a look at this one:

image

and this one:

image

This is related to padding/space reserved for the gutter rather than alignment. Maybe it can be improved.

jtpio:
Ah ok, yeah the screenshots suggested had changed significantly, but it looks fine when resizing the file browser:

[resize-filebrowser.webm](https://github.com/user-attachments/assets/a03fa6af-7409-4fdc-ad09-ce86c59ca919)

I guess we can then update the reference snapshots.

jtpio:
bot please update snapshots

jtpio:
Maybe we'll want to iterate on the alignment in a follow-up PR.

For example in this snapshot comparison the one of the right looks a bit more "off" since "Modified" does not align with "yesterday". The previous version did not fully align, but it may look a bit less off.

krassowski:
It definitely looked better before, I will look into this.

jtpio:
> It definitely looked better before, I will look into this.

Thanks! I guess it would also be fine to fix it in a separate PR too.

jtpio (approved):
Thanks!

The check-links failure is tracked separately.

PR #16770, opened 2024-09-10
Update to enable block comments in python and other languages that do…

### LINKED ISSUE #16212: Make the block comment work in Python (and other languages without block comments)
### Problem

- CodeMirror does not implement block comment for Python language which means executing block comment shortcut is no-op in Python and users think it is broken
- Python does not define block comment syntax, although `"""` is _de facto_ standard substitute as blessed by BDFL
- we want to keep separate block and line comments shortcuts for users using languages which do have as separate block comments

### Proposed Solution

Either or both:
- use `pythonLanguage.data.of({commentTokens: ...})` to add `"""` as block comment for Python - this only helps for Python
- replace uses of `toggleBlockComment` with a composite command which tries to run `toggleBlockComment` first and `toggleComment` second if the former did not succeed (or check if language data have `commentTokens.block` defined)

### Additional context

See discussion in upstream issue https://github.com/codemirror/dev/issues/1372 - upstream will not include that command nor the `"""` syntax for Python so it is up to JupyterLab to implement it.

adityamali:
## References
Fixes #16212 :  Implements toggleComment if toggleBlock comment is unavailable for the language

## Code changes
In Codemirror-extension replace uses of toggleBlockComment with a composite command which tries to run toggleBlockComment first and toggleComment second if the former did not succeed (or check if language data have commentTokens.block defined).

## User-facing changes
Better block comment implementation for languages that do not support block comments exclusively.


JasonWeill:
@adityamali Thank you for opening this PR! Sorry for the delay in responding. I noticed that some of the unit tests are failing, and they look related to your change. Could you please take a look? Thanks again!

krassowski:
Thank you for working on it @adityamali! Would you be willing to also add a test case following the existing test case for JS?

https://github.com/jupyterlab/jupyterlab/blob/ff101050c627ecebdbac1c73638debf58d0f8961/galata/test/jupyterlab/file-edit.test.ts#L39-L53

PR #16771, opened 2024-09-10
Remove Python 2 from documentation

JasonWeill:
macOS has not included Python 2.7 since macOS 12.3. The current OS version is 14.6, and major version 15 is currently in beta. macOS 12, also known as "Monterey", will likely reach its end of maintenance after macOS 15 "Sequoia" is released, probably in Q4 2024.

This change removes mentions of Python 2 from JupyterLab's documentation. It is no longer necessary to run `pip3` or `python3` to guarantee that Python 3 is being run. The presence of macOS-specific instructions led one colleague to presume that the JupyterLab installation instructions are Mac-specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ If you use [conda](https://docs.conda.io/en/latest/), [mamba](https://mamba.read
   ```shell
   pip install jupyterlab
   ```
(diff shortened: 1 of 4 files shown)


JasonWeill:
"examples" test failure appears to be unrelated to this change.

jtpio:
> "examples" test failure appears to be unrelated to this change.

This check seems to be failing on other PRs as well. Opened https://github.com/jupyterlab/jupyterlab/issues/16772 to track this.

jtpio (approved):
Thanks!

PR #16776, opened 2024-09-11
Update `Show Virtual Scrollbar` -> `Show Minimap`

jtpio:
## References

The user facing changelog for `4.3.0` mentions the word minimap:

https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html#minimap

However this feature is still reference as "Virtual Scrollbar" in the UI. For someone looking at the changelog and for this feature it could be confusing. Also the work "Minimap" might be more user-friendly than "Virtual Scrollbar"?

## Code changes

Update the user facing "Virtual Scrollbar" to "Minimap".

## User-facing changes

image

image

None

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -3536,10 +3536,8 @@ function addCommands(
   });
 
   commands.addCommand(CommandIDs.virtualScrollbar, {
-    label: trans.__('Show Virtual Scrollbar'),
-    caption: trans.__(
-      'Show virtual scrollbar (enabled with windowing mode: full)'
-    ),
+    label: trans.__('Show Minimap'),
+    caption: trans.__('Show Minimap (enabled with windowing mode: full)'),
     execute: args => {
       const current = getCurrent(tracker, shell, args);
 
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -3537,7 +3537,9 @@ function addCommands(
 
   commands.addCommand(CommandIDs.virtualScrollbar, {
     label: trans.__('Show Minimap'),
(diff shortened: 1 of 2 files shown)


jtpio:
Opening as a draft to see what other people thing about this.

No strong opinion on my side. And if it's too late w.r.t the localization process for 4.3, we could also consider this change for a future release.

jtpio:
Marking as ready if folks think that this change should be part of the `4.3.0` release, to match the user facing changelog.

Otherwise happy to postpone to `4.4.0` and convert it back to draft.

packages/notebook-extension/src/index.ts
```diff
@@ -3536,10 +3536,8 @@ function addCommands(
   });
 
   commands.addCommand(CommandIDs.virtualScrollbar, {
-    label: trans.__('Show Virtual Scrollbar'),
-    caption: trans.__(
-      'Show virtual scrollbar (enabled with windowing mode: full)'
-    ),
+    label: trans.__('Show Minimap'),
+    caption: trans.__('Show Minimap (enabled with windowing mode: full)'),
```
krassowski, line 3540: I wonder if it makes sense to mention both minimap and virtual scrollbar in caption for continuity? Otherwise I think it is a great idea!

jtpio: Sound good. So keeping `Show Minimap` for the `label`, and something like `Show Minimap (virtual scrollbar, enabled with windowing mode: full)` for the `caption`?

krassowski: Sounds good to me!

krassowski (approved):
Thanks!

krassowski:
I think this should go into 4.3.0

PR #16783, opened 2024-09-16
Plugin based service manager

jtpio:
## References

Reviving https://github.com/jupyterlab/jupyterlab/pull/16556

https://github.com/jupyterlab/jupyterlab/issues/15329

## Code changes

TBD

## User-facing changes

None

## Backwards-incompatible changes

None

PR #16794, opened 2024-09-19
Allow customizing the `ServiceManager` with plugins

### LINKED ISSUE #12504: Allow extending the kernel / service / session manager using an extension
### Problem

It's impossible to extend the KernelManager / ServiceManager / SessionManager without forking jupyterlab. They are all created by the actual JupyterLab instance (which is created [here](https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/staging/index.js#L177-L189)) and so there is no easy way to extend those services at the moment

### Proposed Solution

Allow injecting the application itself as an extension or allow injecting the services above as extensions

Am I overlooking something?

### LINKED ISSUE #15329: Split `ServiceManager` into multiple plugins

jtpio:
## References

Fixes #15329 
Fixes #12504
Replaces https://github.com/jupyterlab/jupyterlab/pull/16556
Replaces https://github.com/jupyterlab/jupyterlab/pull/16783
Closes https://github.com/jupyterlab/jupyterlab/pull/16141
Depends on https://github.com/jupyterlab/jupyterlab/pull/16804

More references:
- Based on https://github.com/jupyterlab/lumino/pull/703
- Inspired by https://github.com/jupyterlab/jupyterlab/pull/13558
- Alternative to #16744 
- Follows a similar approach as in JupyterLite, which registers "server" plugins separately on a different Lumino application (the plugin registry was not usable outside an application before https://github.com/jupyterlab/lumino/pull/703): https://github.com/jupyterlite/jupyterlite/blob/eb0bcb543703213ea883dc1da948ce07494a3148/packages/server-extension/src/index.ts 

## Motivation

See https://github.com/jupyterlab/jupyterlab/issues/15329#issue-1967825301

## Code changes

- [x] Update to the latest `@lumino` packages
- [x] Resolve the service manager service first, then start the lab application
- [x] Try with `IDefaultDrive` and `IContentsManager`
- [x] Move all the underlying `ServiceManager` services to their own plugins
- [x] Create a more explicit `ServiceManagerPlugin` plugin type
- [x] ~Use a separate `PluginRegistry` for service manager plugins? Or the same one as regular plugins?~
  - ~A different one clearly separates the two phases of plugin activation~
  - [x] The same one would allow regular plugins to consume service manager plugins, for example requiring `IKernelManager` instead of `app.serviceManager.kernels`. 
- [x] Handle `IConnectionStatus`
- [x] Add a section in the extension dev documentation about "Service Manager Plugins"
- [x] ~Add to the user / developer changelog for visibility?~ Postponing for now
- [x] Create a new `@jupyterlab/services-extension` package to host the service manager plugins
- [x] Mock package test (usage)

This can be tested using this example extension, which disables `@jupyterlab/application-extension:default-drive` and replaces with its own implementation: https://github.com/jtpio/jupyterlab-custom-service-manager-plugins

**Example providing a custom default drive**

image

**Example providing a custom `ContentsManager` that logs each `get` call**

```ts
import {
  Contents,
  ContentsManager,
  IContentsManager,
  ServiceManagerPlugin
} from '@jupyterlab/services';

class CustomContents extends ContentsManager {
  async get(
    path: string,
    options?: Contents.IFetchOptions
  ): Promise<Contents.IModel> {
    console.log('CustomContents.get', path);
    return super.get(path, options);
  }
}

const contentsPlugin: ServiceManagerPlugin<Contents.IManager> = {
  id: 'myextension:contents-drive',
  description: 'A JupyterLab extension providing a custom contents manager',
  autoStart: true,
  provides: IContentsManager,
  activate: (_: null): Contents.IManager => {
    console.log('Using a CUSTOM contents manager plugin');
    return new CustomContents();
  }
};

export default [
  contentsPlugin,
];
```

image

## User-facing changes

None

None


echarles:
Hi @jtpio I will take time to review in more details with a focus on security and user information. I have share previously concerns about multiple extensions willing to take ownership of those services, and the "last wins" approach is just not acceptable. The content registry approach sounded like a good middle-ground to answer the flexibility needs. what are you missing with the content registry to cover the jupyterlite needs (as far as I understand, this aims to answer jupyterlite case, maybe others.?)

jtpio:
Thanks @echarles for looking into this.

> security and user information

As discussed in https://github.com/jupyterlab/jupyterlab/pull/13558, and more precisely in https://github.com/jupyterlab/jupyterlab/pull/13558#issuecomment-1344090448, it's all JavaScript in the end, and nothing prevents an extension from replacing the whole service manager (or just part of it) *today*.

> The content registry approach sounded like a good middle-ground to answer the flexibility needs. what are you missing with the content registry to cover the jupyterlite needs (as far as I understand, this aims to answer jupyterlite case?)

It's not only about contents, but also kernels, in-browser or even hybrid (in-browser + real kernels in the server, managed by a single kernel manager), settings (being able to store them in the browser storage), terminals (better integration with https://github.com/jupyterlite/terminal), ...

Overall this PR is just an incremental improvement to be able to customize the `ServiceManager` more easily, which is for now a monolith. And it's backwards compatible.

fcollonval:
> Use a separate PluginRegistry for service manager plugins? Or the same one as regular plugins?
>
>   1.  A different one clearly separates the two phases of plugin activation

I don't understand why considering this. It will make things very hard to reason about and will require new configurations. 

Example of very hard to reason thing: a plugin that depends on plugin from the app registry and from the servicemanager registry will be impossible but why forbidding it.

>    2. The same one would allow regular plugins to consume service manager plugins, for example requiring IKernelManager instead of app.serviceManager.kernels.

It will also avoid to think about what is available by whom and how to specify a plugin. The only impact will be whether the app will be defined or not.

jtpio:
> I don't understand why considering this. It will make things very hard to reason about and will require new configurations.

Yes, I reverted this already and working on some other edits now to simplify the extension and plugin loading in the main index.js

jtpio:
> It will also avoid to think about what is available by whom and how to specify a plugin. The only impact will be whether the app will be defined or not.

Yes exactly. I'm adding a couple more plugins to https://github.com/jtpio/jupyterlab-custom-service-manager-plugins to illustrate more use cases.

jtpio:
Also there seems to be some CSS changes which are likely related to the Lumino update?

[latest-lumino-css-changes.webm](https://github.com/user-attachments/assets/c8dc86d5-587d-4854-808f-5a6376a3cbe8)

It could be worth attempting the Lumino update separately to fix the CSS, or resume https://github.com/jupyterlab/jupyterlab/pull/16674.

jtpio:
Rebasing on top of https://github.com/jupyterlab/jupyterlab/pull/16804 to get the Lumino and CI fixes from https://github.com/jupyterlab/jupyterlab/pull/16804.

docs/source/extension/extension_dev.rst
```diff
@@ -214,6 +214,79 @@ See the `JupyterLab Light Theme <https://github.com/jupyterlab/jupyterlab/tree/m
...
+as they would have to build a new JupyterLab application from scratch.
+
+Starting from JupyterLab 4.3.0, the Service Manager is itself a plugin which can be provided by a third-party extension using the ``IServiceManager`` token.
+Its underlying services (such as the kernel manager and the contents manager) are also now available as plugins,
+They too can be provided by third-party extensions via the following tokens:
+
+* ``IConnectionStatus``: The connection status service.
+* ``IContentsManager``: The contents manager service, responsible for managing files and directories.
+* ``IDefaultDrive``: The default drive service, responsible for providing the default drive in which the contents manager operates.
+* ``IServerSettings``: The server settings service, definiting a set of default server settings.
```
afshin, line 238:
```suggestion
* ``IServerSettings``: The server settings service, defining a set of default server settings.
```

docs/source/extension/extension_dev.rst
```diff
@@ -214,6 +214,79 @@ See the `JupyterLab Light Theme <https://github.com/jupyterlab/jupyterlab/tree/m
...
+* ``IEventManager``: The event manager service, for managing Jupyter Server events.
```
afshin, line 239:
```suggestion
* ``IEventManager``: The event manager service for emitting events that are broadcast by an event bus managed by Jupyter Server.
```

packages/application-extension/src/index.tsx
```diff
@@ -1322,11 +1323,11 @@ const plugins: JupyterFrontEndPlugin<any>[] = [
   paths,
   propertyInspector,
   jupyterLogo,
-  topbar
+  topbar,
+  // TODO: move to a new `@jupyterlab/services-extension` package?
```
afshin, line 1327: 👍

packages/application/src/connectionstatus.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Application connection status
+ */
+export class ConnectionStatus implements IConnectionStatus {
+  /**
+   * Whether the application is connected to the server or not.
+   *
+   * #### Notes
+   *
+   * Every periodic network polling should be paused while this is set
```
afshin, line 15:
```suggestion
   * All periodic network polling should be paused while this is `true`.
```

jtpio:
Thanks @afshin for the review! I'll look into fixing the few conflicts to we can have CI running.

jtpio:
> @jtpio do you still plan to finish this before 4.4.0? Depending on what are your plans on this PR, we could move to beta earlier or later, as I think this is the last big thing that was planned for 4.4.

Yes I was planning to get back to this next week. Maybe it could be released later during the beta cycle if we want a first beta soon, or we wait until next week for `4.4.0b0`.

echarles:
As shared previously a few time, I am not if favor to release this in the current state as being too risky.

afshin:
Eric, please consider consulting with the security subproject to examine this change.

I am willing to be the counterparty to answer any questions they may have, because I am confident this changes *nothing* in our current security posture and I am happy to have the debate with the team who has focused on security.

I am :+1: on this functionality.

jtpio:
The `splice_source` failure is related to the new `services-extension` package: https://github.com/jupyterlab/jupyterlab/actions/runs/12747104668/job/35528527011?pr=16794. Which should fix itself after a new pre-release.

For reference this was also the case when the workspaces UI was added: https://github.com/jupyterlab/jupyterlab/pull/15946

packages/services/src/connectionstatus.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { IConnectionStatus } from './tokens';
+
+/**
+ * Application connection status
+ */
+export class ConnectionStatus implements IConnectionStatus {
```
jtpio, line 9: In the previous version of this PR, `ConnectionStatus` and `IConnectionStatus` were part of the `@jupyterlab/application` package, and its related plugin part of `@jupyterlab/application-extension`.

With the current state of this PR, it is part of `@jupyterlab/services`. Although this may be more related to an "application", having it in `@jupyterlab/services` and its related plugin in `@jupyterlab/services-extension` has two advantages:

- no dependency on `@jupyterlab/application` from `@jupyterlab/services-extension`
- all `ServiceManagerPlugin` plugin definitions are in `@jupyterlab/services-extension`

afshin: What if `@jupyterlab/application` imports it from services and re exports it and deprecates it (optionally)? Since it's a static const it shouldn't matter where you get it from I think?

jtpio: I meant in the previous attempt, for example in https://github.com/jupyterlab/jupyterlab/pull/16556, which was closed by not merged.

Currently in JupyterLab there is no `ConnectionStatus` / `IConnectionStatus`, so not sure we need to re-export? So it was mostly a comment on where the new `ConnectionStatus` class and `IConnectionStatus` interface and token should live.

krassowski:
This one is reliably failing (I restarted it once to check if it was just flaky):

```
Linux Tests / Linux (splice_source, 3.13) (pull_request)
```

jtpio:
> This one is reliably failing (I restarted it once to check if it was just flaky):

Yes I think it's related to the new `services-extension` package? https://github.com/jupyterlab/jupyterlab/pull/16794#issuecomment-2587166344

Carreau:
I took a note to bring this up at the next security meeting; as any configurability this increases the complexity of understanding the attack surface of any application, but I don't this this opens any new attack surface; there is not more user input that gets parsed or trigger new code execution and it's all just new extensions points, so I'm not sure I see where the security concerns are.

echarles:
This PR allows to silently replace the service manager responsible for the core feature of Jupyter (content and computing access). A hacker will easily create an good looking extension with the official API that fools a user and leaks content and result. That is to me a major attack surface increase, ruining previous historical efforts (notary...) to make Jupyter more secure.

Carreau:
> This PR allows to silently replace the service manager responsible for the core feature of Jupyter (content and computing access). A hacker will easily create an good looking extension with the official API that fools a user and leaks content and result. That is to me a major attack surface increase, ruining previous historical efforts (notary...) to make Jupyter more secure.

But I'm confused how is that different from a hacker publishing a `jupiter-lab` (with a i) package that is a fork of `jupyter-lab`, look exactly the same and exfiltrate data ?

Any extension can already make arbitrary request to the backend, execute code kernel side and poison everything anyway. "it's less code to work for an attacker" is generally not a new vulnerability.

echarles:
Today vulnerabilities and ways to hack Jupyter should not be an argument to make the situation even worse and dangerous. Once shipped to the developers, that potential new API will be used and there will be no way to revert. The ServiceManager is a crucial and sensible part of the stack ensuring security and trust to the server with authentication and authorization. It sounds very bad to open even more ways to fool the end-user, who finally will not know with which server he is communicating.

krassowski:
I agree that all of these are true:
- this PR does not change the security posture as of today,
- this PR (as any feature) increases attack surface
- we should not limit ourselves to what is possible today

If we want to create a **secure extension experience** in the future we would want to create a new extension manager which allows extension only limitted public API access, using a permission-based model (same as when you install and app on your smartphone or a browser extension) which prompts for permission to use certain APIs or at least lists which APIs it will have access too. Example Chrome extension:

image

This is not implemented as of today, but it would be _easy_ to implement with webworkers which prevent access to DOM and allow to expose a very limited API to the extension. These extensions would be fundamentally limited in what they can do, but majority of extensions could be build with this model (plus iframe sandbox isolation and tight sanitizer control) - but of course it would be harder to build these extensions.


What does this PR mean for that possibility? I think not much - it would be possible to always prompt user before first loading an extension which asks for "Swap Service Manager plugins" permission. I am happy to be proven wrong if I missed something fundamental here.

echarles:
> this PR does not change the security posture as of today,
Releasing this is a no-way back (you can not remove such a feature) - in that sense I consider this is a change in the security posture as any decision we will make in the fyture to make the ecosystem more secure will have to deal with that.

> this PR (as any feature) increases attack surface
I don't agree that any feature increase attack surface in such a significant way as this PR does.

> it would be possible to always prompt user before first loading an extension which asks for "Swap Service Manager plugins" permission. 

I am happy to read this. That is the kind of discussion I have tried to engage with and I see that is just beginning. A subsequent dicsussion is what if multiple extensions compete for the service manager? I would rather put flat the design and think towards a registry of service managers rather than just the `silent hard replace last-one wins` approach of this PR.

jtpio:
Eric, at the risk of repeating the same arguments again (https://github.com/jupyterlab/jupyterlab/pull/16794#issuecomment-2362876132, https://github.com/jupyterlab/jupyterlab/pull/13558#issuecomment-1344090448, https://github.com/jupyterlab/jupyterlab/pull/13558#issuecomment-1346500780), this PR does not anything regarding security. Writing them down here in case you missed them:

**This has always been possible, in all versions of JupyterLab**:

```ts

const insecure: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/application-extension:insecure',
  autoStart: true,
  activate: (app) => {
    console.log('replacing the application service manager');
    ((app as unknown) as any).serviceManager = new ServiceManager();
  }
};
```

Currently *any* extension can run arbitrary JavaScript code in the main thread. It's all JavaScript in the end. This is why the extension shows a disclaimer:

image

To avoid hijacking the conversation here, I would prefer to have this discussion separately since it's more about the JupyterLab extension system in general than what this PR provides.

The points brought by @krassowski related to a permission system sound like an interesting way to explore, but would likely require more design and thoughts.

docs/source/extension/extension_dev.rst
```diff
@@ -214,6 +214,79 @@ See the `JupyterLab Light Theme <https://github.com/jupyterlab/jupyterlab/tree/m
 
 See the `TypeScript extension template <https://github.com/jupyterlab/extension-template>`__ (choosing ``theme`` as ``kind`` ) for a quick start to developing a theme plugin.
 
+Service Manager Plugins
+^^^^^^^^^^^^^^^^^^^^^^^
+
+.. note::
+   Added in: JupyterLab 4.4.0
```
krassowski, line 221: Should we use https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded ?

krassowski (commented):
@jtpio what is your primary motivation for this PR? Is this JupyterLite or something else?

If it is JupyterLite, could we restrict the service manager plugins to be only installable via source rather than prebuilt extensions? This would mean that only advanced users, and application developers are ever exposed to such extensions as those would not show up in Extension Manager. I wonder if this would be a good compromise between the concerns raised by @echarles and the immediate need. But of course, if this does not help the immediate use case you had in mind then maybe it is not a good idea.

jtpio:
> what is your primary motivation for this PR? Is this JupyterLite or something else?

The related issues (https://github.com/jupyterlab/jupyterlab/issues/15329#issue-1967825301) mentions some use cases:

> Probably the main reason for having the `ServiceManager` be a "special" component of a JupyterLab application is likely based on the assumption that a `JupyterFronEnd` should be served via a fully-featured Jupyter Server (or even a JupyterLab Server).
> However over the past years we have seen more use cases that are starting to challenge this assumption. For example distributions with in-browser kernels like JupyterLite, or plugin-based Jupyter Server alternatives like Jupyverse, which use or enable only a subset of what a traditional Jupyter Server offers.
> In these situations a plugin-based `ServiceManager` would allow for more easily disable or replace some services via normal extensions. This would mean providing each service of the `ServiceManager` as a token, for example `IKernelManager` or `ISessionManager`.

There is also https://github.com/jupyterlab/jupyterlab/issues/12504.

Also: allow for more flexibility, like with the existing extension system, be able to easily build other lab-based apps that are not backed by a full Jupyter Server...

> could we restrict the service manager plugins to be only installable via source rather than prebuilt extensions?

No, that would defeat the purpose of this PR. And installing via source is now deprecated.

Carreau:
> I am happy to read this. That is the kind of discussion I have tried to engage with and I see that is just beginning. A subsequent discussion is what if multiple extensions compete for the service manager?

From a security perspective then we seem to all agree – this does not add security concerns it just extends the attack surface and what we can reason about.

I agree that having sandboxed extensions with permissions (I think in security context it's called a [capability](https://en.wikipedia.org/wiki/Capability-based_security), would be great, but I'm not involved enough in JupyterLab to see how involved that would be. This would also likeley requires having capabilities on the server side in many cases.

docs/source/extension/extension_dev.rst
jtpio: yes good point 👍

echarles:
> From a security perspective then we seem to all agree 

Sorry, I am not part of the "all". Are you planning to ask position from the Security team or are you speaking in its name?

Carreau:
Security team meets on Tuesday. I won't be at next tuesday meeting but will leave an agenda item.

Carreau:
Personally and that's my own opinion: Yes this adds surface area; IMHO adding an API that can't be removed should not be part of the discussion from a security POV, it's mixing actual security question with maintenance question. 

Does this adds a vulnerability: I don't think so. I may be wrong of course. 

Does it makes attacker life easier if there is a vulnerability, likely yes. But isn't that true for any new API ? or any extension point ?

Can it help confuse user ? I'm not sure, maybe ?

It seems to me that it has been shown it is already possible to do what this PR allows, and as you pointed out at some point it would be great to isolate critical extension, and nobody I think disagree with that.

Maybe actually this API will help statically detect extensions that replace the content manager and have extra warnings instead of monkey patching ?

As always I may be missing the forest for the tree.

trungleduc (approved):

echarles:
@rpwagner Thx for the feedback. Is there any minutes or documents where we can read the analysis and reasoning backing that feedback?

rpwagner:
@echarles here is [our page for meeting notes](https://hackmd.io/gcA_DFLLTzaSuxSyyRy5UA). Unfortunately, we were not good about tracking the details of the discussion. I will say that the primary reason we agreed this shouldn't be blocked for security concerns is that it does not introduce a new way for a malicious actor to introduce a compromise. 

It seems that JupyterLab is unique by enabling and promoting ways for users to install packages that run in-browser on their local system. Because of that, it's important that any packages published by JupyterLab and other Jupyter frontend groups are trustworthy (i.e., everything that gets built from the [packages directory](https://github.com/jupyterlab/jupyterlab/tree/main/packages)). Outside of that, it is very challenging. Any hosted service can serve whatever packages they choose to, and malicious actor can publish packages that override important features.

If you have an example of this PR could be used as a new exploit when it is published by JupyterLab as part of release, without the use of an external package, then it would be a concern.

echarles:
Thank you for the link where unfortunately does not contain much details. Based on your previous comment and the previous discussions, I am still not ok with this PR and does not feel we will be able to find consensus with the actors, therefor I am asking to start a vote.

This will give the needed visibility to all JupyterLab developers and will allow to picky back to that decision in case a user is suffering in the future from the taken decision.

rpwagner:
A vote by the Frontends Council is a reasonable approach. We (the Security Subproject) have provided our input, but this decision should be made by the Subproject responsible for JupyterLab.

dlqqq:
@echarles Thank you for your polite honesty! I'll work with the rest of the Security Council to explore ways we can make our discussions more well-documented.

I'd like to share my thought process here in the interest of transparency & collaboration. To start, I believe "risk" is best measured as a combination of two quantities:

1. **Risk impact**: What are the worst-case scenarios?
2. **Risk probability**: How easy is it for a malicious actor to manifest a worst-case scenario?

I will use this rubric to evaluate this PR:

- The impact (worst-case scenarios) remains severe. A malicious extension could mine crypto on your devices, clear your hard drive on init, or override the `ContentsManager` from the backend (via `traitlets`) to exfiltrate all user data to a bad actor. These are all possible today, without this PR.

- The probability (ease of exploit) remains the same. A bad actor can use our extension cookiecutter and fill it with evil instead of good. A user can then install this malicious extension by accident via `pip install`. This PR does not make it any easier to create a worst-case scenario.

**Therefore, based on my current understanding, I do not think that merging this PR increases our security risk.**

dlqqq:
I agree with @echarles that this PR does increase the risk surface (i.e. number of potential code locations to exploit). However, I argue that risk surface is a poor measure of risk. In my opinion, a vulnerable plugin system with 10 plugin points is just as vulnerable as a system with 100 plugin points (given that the extra 90 don't change the set of worst-case scenarios). 

In the Jupyter Security Working Group call this morning, we spent the latter half of the call exploring ways to make the plugin system more secure in a future major version. I think that this would be the best way to address the security risk posed by JupyterLab's plugin system.

jtpio:
> give the needed visibility to all JupyterLab developers

A good first step would be to come to the weekly [Jupyter Frontends meetings](https://github.com/jupyterlab/frontends-team-compass?tab=readme-ov-file#weekly-jupyter-frontends-team-meeting) to discuss this.

krassowski:
We were discussing this during the weekly call last week. I know @afshin and @jtpio were also keen to resolve the situation. I wanted to follow up to clarify:
- @echarles did you intend to start a formal vote in the council repo?
- Is your objection to merging this PR, or to releasing it in a final version without more testing?
- Has there been any new discussion, or presentation of security arguments by @echarles that may have been private (because of nature of arguments)?

echarles:
> I know @afshin and @jtpio were also keen to resolve the situation. I wanted to follow up to clarify:

I have a call planned with @jtpio next week on that.

> @echarles did you intend to start a formal vote in the council repo?

I intend the promoters of this PR to open a vote.

> Is your objection to merging this PR, or to releasing it in a final version without more testing?

Yes, objection from me.

> Has there been any new discussion, or presentation of security arguments by @echarles that may have been private (because of nature of arguments)?

Not from my side.

krassowski:
For public record, a vote has been triggered on 29th Jan with voting happening in https://github.com/jupyterlab/council/discussions/28. Per my understanding of voting timeline discussion, the vote will end tomorrow (after 7 days), unless there is a motion and a second to extend it (https://github.com/jupyterlab/frontends-team-compass/issues/154). As per consensus in https://github.com/jupyterlab/frontends-team-compass/issues/156 the voting happens anonymously with summary results to be published publicly for transparency.

echarles:
@krassowski As shared on that private vote, this vote is not legit. It does not respect the jupyter governance (1)

Also, only one concern, security, is  stated in the vote opening (the second concern, usability, is not stated)

Historically, I have seen votes being held publicly, but it looks like this is not documented in the decision making.

(1) https://jupyter.org/governance/decision_making.html

When a vote is called on GitHub, the sponsor should summarize the decision and paste a checklist of the voting members to use in voting in the top entry (description) of the issue/PR.

jtpio:
> When a vote is called on GitHub, the sponsor should summarize the decision and paste a checklist of the voting members to use in voting in the top entry (description) of the issue/PR.

It looks like this is under the "Optional aspects of decision making" section of the decision-making guide?

image

echarles:
Well it is still recommended `...the following are recommended guidelines rather than mandatory procedures.`

As posted on https://github.com/jupyterlab/frontends-team-compass/issues/156 

```
Currently, the Jupyter Steering Council votes by listing each member of the SC by name on an issue and associating their name to their vote in an asynchronous voting process that takes 4 weeks.

The JupyterLab Council follows the same mechanics except with a shorter voting period.
```

I find this change of way to work very tough process and not transparent, against the openess standard that this Jupyter community is supposed to promote.

echarles:
See also https://github.com/jupyterlab/frontends-team-compass/issues/156#issuecomment-2634659297

Voting by obscurity is again not aligned with the openess and transparency standards supposed to live here.

echarles:
I thought a bit more to the vote opened by @afshin and the anonymous voting approach does not allow to enforce 2/3 vote.

@afshin Please close the private and anonymous vote you have opened. Once close, I will open a public facing with a list of voters to respect the governance and align with former votes done in JupyterLab.

afshin:
The [`jupyterlab/council`](https://github.com/jupyterlab/council/) repo is only accessible by the members of the council. Therefore even if you don't see people's names, you can still count the votes and be confident that all the votes come from members of the relevant voting body. It is a legitimate vote, irrespective of your textual interpretation, @echarles.

Calling a vote is any member's right. It is also valid to check if anyone seconds the motion to call a vote. No one else seems to. I asked in the other thread if anybody wants to second your motion to put this to a vote. No one else does.

I understand you disagree with the consensus that has emerged and you wish to stop it. Please stop calling into question the legitimacy of our decision-making procedures. A secret ballot is one of the hallmarks of democracies around the world. So is the ability to accept the outcome of decisions despite disagreement.

As far as I can see, we have consensus on this issue. We do not have unanimity, but I would argue we do have consensus.

Eric, please come to today's (5 Feb 2025) Jupyter Frontends and Accessibility call if you wish to make some final case that might persuade someone. This PR and conversation has gone on for long enough and needs to be resolved.

We cannot have one lone objection halt movement for months at a time.

---

**NB:** According to the [decision-making guide](https://jupyter.org/governance/decision_making.html):

> **Voter participation and quorum.** All members of a council are required to participate in at least 2/3 of formal votes of that council per calendar year (councils can decide how to account for the specifics of this in low-activity projects, etc.).

This subproject decided to hold twice-yearly check-ins requiring all members to confirm that they are indeed still active and still interested in participating in decision-making, which is plainly allowed by the rule above.

echarles:
> Eric, please come to today's (5 Feb 2025) Jupyter Frontends and Accessibility call if you wish to make some final case that might persuade someone. This PR and conversation has gone on for long enough and needs to be resolved.

@afshin I'd rather get explanation and confirmation from the executive council member, which you are part of. Can you invite me to the next one. In the meantime, I expect this PR to not be merged whatever result would come from the vote you have opened. To avoid adding more confusion, I will not open another vote for now.

afshin:
We have EC office hours tomorrow, it is open to the public. Please [consult the calendar](https://discourse.jupyter.org/t/jupyter-community-calendar/2485) and feel free to join if you wish.

We will still hold our Frontends and Accessibility call today and try to resolve this. I encourage you to come.

SylvainCorlay (approved):

afshin:
The office hours exist for members of the community to come ask questions from members of the EC. If the question requires following up or further decisions to be made, then normal procedure will follow.

From my point of view, this is a case of a decision that should be made within the Jupyter Frontends Council by the members of **this** subproject. I imagine if you come to office hours, you might make the case that this particular technical decision on whether to merge a specific feature-related PR in the JupyterLab code base should be made by the EC. I don't agree with this interpretation, but you're obviously free to make that case tomorrow.

Nonetheless, I again strongly encourage you to come today to the appropriate meeting to discuss this and to make one final case after all this discussion. I don't believe EC office hours is the correct venue to talk about a JupyterLab PR.

afshin:
@echarles, you stated:

> In the meantime, I expect this PR to not be merged whatever result would come from the vote you have opened.

After months of discussion:

- Not a single other member of the Jupyter Frontends Council has objected to merging this PR and a few have already approved it.
- Not a single other member of the Jupyter Frontends Council has seconded your motion to put this PR to a vote.
- The Jupyter Security Subproject has spent an entire meeting discussing whether your security claims have merit and they have decided this feature does not pose an objectionable risk.

You wish to appeal to a body (the Jupyter Executive Council) that does not adjudicate technical decisions made by subprojects. That is absolutely your right, but I don't think included in that right is the ability to yet again stall motion on a feature simply because you alone object to it.

echarles:
The discussion has moved from technical to governance, the governance aspect is the topic I am willing to share with the executive council. I hope enough people will be there tomorrow, and in all case, I will request a follow-up so the case is shared with all the members of the executive council.

echarles:
I can not join today call, and also diccussed 2 days ago during a call with @jtpio. I would like to say again tomorrow topic will be the invalidity of the vote, so I expect the JupyterLab council @jupyterlab/jupyterlab-council to continue discussing but not merge this PR until the governance point is treated by the Jupyter Exeuctive council.

SylvainCorlay:
> so I expect the JupyterLab council @jupyterlab/jupyterlab-council to continue discussing but not merge this PR until the governance point is treated by the Jupyter Executive council.

I don't think this is a fair expectation at this point.

fcollonval:
Hey all,

I will second @echarles to go for a _public_ vote. I think it is important for heated subject that the decision should be clear and public to reflect the open-minded and transparent decision process of this community.
As @afshin pointed out, the discussion with the poll opened in the council repository was not meant to be a formal vote. It would be good to reboot the vote via an issue on the team compass repository with a ping to the council member asking to vote.

Zsailer:
Cross referencing a reply I made on the other thread, realizing I should have posted here: https://github.com/jupyterlab/frontends-team-compass/issues/273#issuecomment-2638108366

> > The reason we need this PR is because I made a design decision, which I later came to regret, back in 2019 when I initially [jupyterlab/jupyterlab#5845](https://github.com/jupyterlab/jupyterlab/pull/5845).
> > In this initial implementation, the API required all Jupyter frontends to ship with a serviceManager instead of making service managers plugins.
> 
> I have one anecdotal piece of evidence where the old ServiceManager design decision blocked me (sorry [@afshin](https://github.com/afshin) 😉).
> 
> I tried to create a new (RTC-enabled) Notebook-like application that gave JupyterLab user's a link to share to a single document in their server for (real-time) collaborative editing only (no kernel execution allowed). Invitees would get access to _only_ that document, in a single document Notebookv7 view to collaborate on the document's content. The requirements were:
> 
> 1. the UI was a stripped down Notebook v7, just used for editing the content of the notebook.
> 2. should only talk to the host's contents and y-doc websocket APIs.
> 3. No access to the host's kernels, kernelspecs, terminals, sessions, etc.
> 4. Still include many of the plugins, like TOC, ipywidgets, vega, jupyter-collaboration etc.
> 
> In the idea case, I would have just liked to use Notebook v7 with a configurable ServiceManager class to **replace** the parts I needed (note that I do need a "replacement" here, not just an API for addition). Instead, the hard-coded ServiceManager posed a problem for me; it hardcodes all Jupyter Server APIs and expects _all_ of the services to working. It also **automatically polls the kernelspecs APIs (any sessions?)** on a regular interval. While we can use server auth to prevent access to these APIs, the client constantly dumps errors due to this polling behavior. On the other hand, there was no way to stop this unwanted behavior without creating a new app from scratch. Ultimately, this is what I did, and it led to a lot of extra duplicate logic.
> 
> Under the proposed PR, I would get what I needed much easier. There may be different approaches we could take, but this PR seems reasonable to me. Just making the current servicemanager "expandable" using providers doesn't feel quite enough and I haven't seen a strong argument on why we shouldn't other than perceived risk of API scope creep.
> 
> Just sharing one, personal use-case outside of just JupyterLite.

Zsailer:
Maybe one improvement that would help address some of @echarles usability concerns is to consider some way to dump out the plugin dependency graph, so if a user reports some unexpected behavior because on conflicting plugins, we can more easily trace the issue?

echarles:
> cc/ Executive Council (as discussed, I will join tomorrow 06 Feb call).

A valid vote being now opened, I will thus not join today Executive council to discuss this case cc/ Executive Council @afshin @Ruv7 @ellisonbg @fperez @jasongrout @Zsailer

packages/application/src/lab.ts
```diff
@@ -126,6 +122,8 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
 
   /**
    * A list of all errors encountered when registering plugins.
+   *
+   * @deprecated This is unused and may be remove in a future version.
```
krassowski, line 126:
```suggestion
   * @deprecated This is unused and may be removed in a future version.
```

packages/application/src/lab.ts
```diff
@@ -169,6 +167,8 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
    * Register plugins from a plugin module.
    *
    * @param mod - The plugin module to register.
+   *
+   * @deprecated This is unused and may be remove in a future version.
```
krassowski, line 171:
```suggestion
   * @deprecated This is unused and may be removed in a future version.
```

packages/application/src/lab.ts
```diff
@@ -192,6 +192,8 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
    * Register the plugins from multiple plugin modules.
    *
    * @param mods - The plugin modules to register.
+   *
+   * @deprecated This is unused and may be remove in a future version.
```
krassowski, line 196:
```suggestion
   * @deprecated This is unused and may be removed in a future version.
```

packages/application/src/lab.ts
```diff
@@ -31,16 +35,8 @@ export class JupyterLab extends JupyterFrontEnd<ILabShell> {
...
-    const info = Object.keys(JupyterLab.defaultInfo).reduce((acc, val) => {
-      if (val in options) {
-        (acc as any)[val] = JSON.parse(JSON.stringify((options as any)[val]));
-      }
-      return acc;
-    }, {} as Partial<JupyterLab.IInfo>);
-
     // Populate application info.
-    this._info = { ...JupyterLab.defaultInfo, ...info };
+    this._info = new JupyterLab.Info(options);
```
krassowski, line 39: Note to self: this will prevent writing to some attributes, but it's ok because it is private

krassowski (approved):
I noted a tiny typo, otherwise good!

jtpio:

Issue #16795, opened 2024-09-19
Same `id` can be used for different headings


HaudinFlorence:
## Description

The same `id` can be used for differents headers if they have the same text content because the `id` is built from the text of the headings, with spaces replaced by `-`.
This possibility of duplicated `id`s was observed when working on this PR in nbconvert: https://github.com/jupyter/nbconvert/pull/2178. For building a table of contents with links to some DOM elements, it is important to have unique identifiers.

## Reproduce

1. Create a notebook containing markdowns with headings of the same level and containing the same text, like the following one for instance
![Screenshot from 2024-09-19 16-19-14](https://github.com/user-attachments/assets/37d116df-385e-4177-8f1f-57ade89a3a44)

3. If you inspect the `Section 1` elements with the dev tools, you can see that their `id`, generated from their text content, is exactly the same (`Section-1`)

![Screenshot from 2024-09-19 16-17-22](https://github.com/user-attachments/assets/7ffef030-48a2-4132-9de4-a2dfa38ec329)

![Screenshot from 2024-09-19 16-17-52](https://github.com/user-attachments/assets/af289f09-4d1b-4e25-a4e8-68dc842d4554)

## Expected behavior

In case there are duplicated headings, a suffix could be added to the `id`.

## Context

- JupyterLab version: 4

onursercanyilmaz:
Hi @HaudinFlorence , if it is available can you assign me this? I handled in renderers.tsx like

```
function getRandomLetters(length: number): string {
    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
    let result = '';
    for (let i = 0; i < length; i++) {
        const randomIndex = Math.floor(Math.random() * characters.length);
        result += characters[randomIndex];
    }
    return result;
}

async function hashString(str: string): Promise<string> {
    // Simple hash function using SHA-1
    const encoder = new TextEncoder();
    const date = new Date();
    const randomLetters = getRandomLetters(5); // Generate 5 random letters
    const text = str + randomLetters + date.toString();
    
    const data = encoder.encode(text);
    const buffer = await crypto.subtle.digest('SHA-1', data);
    return Array.from(new Uint8Array(buffer))
        .map(b => ('00' + b.toString(16)).slice(-2)) // Ensures each byte is represented by 2 characters
        .join('');
}

```

onursercanyilmaz:
[...]

PR #16800, opened 2024-09-21
Fix issue 16795 - Suffix added to header ids

onursercanyilmaz:
## References
 #16795 

## Code changes

Changes in renderers.tsx, as shown below.

```
  export function headerAnchors(node: HTMLElement): void {
    const headerNames = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
    for (const headerType of headerNames) {
      const headers = node.getElementsByTagName(headerType);
      for (let i = 0; i < headers.length; i++) {
        const header = headers[i];
        const parentElement = header.closest(
          '.lm-Widget.jp-Cell.jp-MarkdownCell.jp-Notebook-cell.jp-mod-rendered'
        );
        const parentIndex =
          parentElement?.getAttribute('data-windowed-list-index') ?? '0';
        header.id = renderMarkdown.createHeaderId(header, i, parentIndex);
        const anchor = document.createElement('a');
        anchor.target = '_self';
        anchor.textContent = '¶';
        anchor.href = '#' + header.id;
        anchor.classList.add('jp-InternalAnchorLink');
        header.appendChild(anchor);
      }
    }
  }
  
   export function createHeaderId(header: Element, index: number, parentIndex: string): string {
    return (`${header.textContent} ${index} ${parentIndex}` ?? '').replace(/ /g, '-');
  }
```

If a cell includes more than 1 header it adds 'i' (index) to id and after that adds the parent div's index info as header-{innerIndex}-{parentIndex}. Issue is about parentIndex problem but there was another problem on inner headers that solved too.

image

But still there was an issue that for example - if user adds #Part 1 as h1 and ##Part 1 as h2 - it creates same id.
So i added headerType parameter too

```
  export function createHeaderId(
    header: Element,
    index: number,
    parentIndex: string,
    headerType: string,
  ): string {
    return (`${header.textContent} ${index} ${parentIndex} ${headerType}` ?? '').replace(
      / /g,
      '-'
    );
  }
  
   export function headerAnchors(node: HTMLElement): void {
    const headerNames = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
    for (const headerType of headerNames) {
      const headers = node.getElementsByTagName(headerType);
      for (let i = 0; i < headers.length; i++) {
        const header = headers[i];
        const parentElement = header.closest(
          '.lm-Widget.jp-Cell.jp-MarkdownCell.jp-Notebook-cell.jp-mod-rendered'
        );
        const parentIndex =
          parentElement?.getAttribute('data-windowed-list-index') ?? '0';
        header.id = renderMarkdown.createHeaderId(header, i, parentIndex, headerType);
        const anchor = document.createElement('a');
        anchor.target = '_self';
        anchor.textContent = '¶';
        anchor.href = '#' + header.id;
        anchor.classList.add('jp-InternalAnchorLink');
        header.appendChild(anchor);
      }
    }
  }
```
  
image
image

But i think it seems not good. So I decided to encode suffix part
```
 function encodeSuffix(
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = `${index} ${parentIndex} ${headerType}`;

    // Use TextEncoder to convert the string to a Uint8Array
    const encoder = new TextEncoder();
    const data = encoder.encode(suffix);

    // Convert to Base64
    let binaryString = '';
    for (let i = 0; i < data.length; i++) {
      binaryString += String.fromCharCode(data[i]);
    }

    return btoa(binaryString);
  }

  export function createHeaderId(
    header: Element,
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = encodeSuffix(index, parentIndex, headerType);
    return (`${header.textContent} ${suffix}` ?? '').replace(/ /g, '-');
  }

```

image

## User-facing changes
- Users will now see the newly generated id in the URL when they click on header items, as the id attributes are assigned to the <a> tags as href values.

image


jtpio:
Thanks @onursercanyilmaz for working on this!

The diff shows many changed files, which are likely related to the new `4.3.0b2` that was released in the meantime. Would you be able to rebase and fix the conflicts, which should reduce the number of files changed (even though it's a bit tedious!). Thanks!

HaudinFlorence:
Thanks for your PR @onursercanyilmaz . I am currently testing it and successfully checked that the produced ids were unique on an example notebook. Could you please explain in more details the choice made for the determination of `ParentElement` and `ParentIndex`? What are the a priori expected values for `ParentIndex`? If you could please give an example please? Also, could you please elaborate more about the necessity or the strength of encoding the suffix? Thanks.

onursercanyilmaz:
Thank you for your comments @HaudinFlorence. 
Here’s the detailed explanation:

### **Cause:**
In JupyterLab, when assigning IDs to markdown headers, each cell (node) operates in isolation, meaning only the headers within that cell are considered. Since there is no global ID system, the following issues arise:

1. When the same or different cells use the same header text, the same IDs are repeated. For example:
    - If a **h1** tag is created in Cell 1 with the text "Header," its ID will be `Header`.
    - If the same "Header" is created in Cell 2, this header will also be assigned the ID `Header`.

2. Additionally, when emojis or special characters are used in headers, they are included in the IDs. For example, if a header text is "🚀 Start," it will be assigned the ID `🚀-Start`. This situation reduces the readability and consistency of the IDs.

### **Development:**
Using a **counter** to create unique IDs based on header text might seem logical, but since each cell operates independently, this method fails. Specifically:

- The **node** resets to zero in each cell. For example:
    - If there are two **h1** headers in Cell 1, they might be assigned IDs `Header` and `Header-1`.
    - However, if the same headers are used in Cell 2, the count starts again from zero, resulting in the same IDs: `Header` and `Header-1`.

- In this case, the **counter** method can only apply within that cell, meaning it cannot ensure global uniqueness across cells. As cells are isolated, the **counter** solution proves inadequate for establishing unique IDs.

**Example:**
Cell 1:
```markdown
# Header
# Header
```
IDs: `Header`, `Header-1`

Cell 2:
```markdown
# Header
# Header
```
IDs: again `Header`, `Header-1`, leading to conflicts.

### **Result:**
To resolve this issue, I got a unique **parentId** (or parentIndex) for each cell to ensure the headers have unique IDs. With this method:

1. Each header text had its cell's **parentId** value appended (Each cell's div tag has a data-windowed-list-index attribute value = parentIndex), allowing for differentiation even if the header text was the same.
   
2. Also encoded the header text to make it unique and began using it in Base64 format. This way, headers were made uniquely identifiable.

**Example:**
In Cell 1 with two headers:
```markdown
# Header
# Header
```
These headers might have IDs like `0-Header` and `0-Header-1` (where **0** represents the parentIndex).

In Cell 2, again using the same headers:
```markdown
# Header
# Header
```
These headers would then be assigned IDs `1-Header` and `1-Header-1`, since the parentIndex this time is **1**.

This method ensures that headers in each cell remain unique in large and complex notebooks, thus preventing ID conflicts.


Using a global counter can lead to performance issues and complexity, as it requires tracking header occurrences across all cells, making it less efficient in managing unique IDs within a large notebook.

onursercanyilmaz:
new pr is #16845

Issue #16812, opened 2024-09-25
Cannot open notebooks with stream in outputs in JupyterLab 4.3.0b0-1


krassowski:
## Description

A critical performance regression in JupyterLab 4.3.0b0 was identified in testing: notebooks with large amount of stream data in outputs no longer open in reasonable time.

While users are reporting that notebooks do not open at all, it appears that the time opening a notebook is now a quadratic or exponential function of stream data content, which may indicate that some previously functional notebooks will not open before the heat death of the universe.

This was most likely:
- introduced in https://github.com/jupyterlab/jupyterlab/pull/16498, and
- not fixed by https://github.com/jupyterlab/jupyterlab/pull/16692

CC @davidbrochart

## Reproduce

1. Download the attached notebook [50_char_500_lines.json](https://github.com/user-attachments/files/17131015/50_char_500_lines.json) file
2. Rename it to `50_char_500_lines.ipynb` - it has only one code cell and only one output
3. Open it in JupyterLab 4.3
4. See that it takes 10-15 seconds to open (depending on the browser)

The single output of the cell looks like this:

```js
{
     "name": "stdout",
     "output_type": "stream",
     "text": [
         "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n",
         /* of different lengths, repated many times */
         "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n"
     ]
}
```

Timings in Chrome:
- 50 characters repeated in 100 lines took 0.3s to load
- 50 characters repeated in 200 lines took 1.4s to load
- 50 characters repeated in 500 lines took 8s to load
- 50 characters repeated in 1000 lines took 30s to load

Timings in Firefox:
- 50 characters repeated in 500 lines took 15s to load
- 50 characters repeated in 1000 lines took 93s to load

I have not tested Safari but I would expect even worse timings.

## Expected behavior

Notebooks opens as fast as in JupyterLab 4.2 (or faster) where even the test notebook with 1000 lines takes **less than one second** to open.

## Context

- Browser: Chrome and Firefox
- JupyterLab version: 4.3.0b0 and 4.3.0b1

Here is the relevant flame graph:

![image](https://github.com/user-attachments/assets/306cf581-16f6-4aa7-b680-9a75ad34cd06)

krassowski:
[...]

Issue #16772, opened 2024-09-11
Failing examples check on CI


jtpio:
The `examples` check seems to be failing consistently on CI. For example:
- https://github.com/jupyterlab/jupyterlab/actions/runs/10801167759/job/29961047173?pr=16770
- https://github.com/jupyterlab/jupyterlab/actions/runs/10801370571/job/29961399584?pr=16771

```
Running 1 test using 1 worker
Navigating to page: file:///home/runner/.local/share/jupyter/runtime/jpserver-6198-open.html
[I 2024-09-10 22:11:25.705 ServerApp] Kernel started: fe8f9dba-02a0-45dc-b4b7-eb7570217e60
debug >> Starting WebSocket: ws://localhost:8888/foo/api/kernels/fe8f9dba-02a0-45dc-b4b7-eb7570217e60
[I 2024-09-10 22:11:26.194 ServerApp] Connecting to kernel fe8f9dba-02a0-45dc-b4b7-eb7570217e60.
debug >> Test complete!
[I 2024-09-10 22:11:26.306 ServerApp] Starting buffering for fe8f9dba-02a0-45dc-b4b7-eb7570217e60:58ca4dc7-03a1-4f9c-9ff3-e9f5e8c8455a
·
  1 passed (2.5s)
[I 2024-09-10 22:11:26.384 ServerApp] Test Complete
[I 2024-09-10 22:11:26.384 ServerApp] Exiting normally
[I 2024-09-10 22:11:26.384 ServerApp] Stopping server...
Traceback (most recent call last):
  File "/home/runner/work/jupyterlab/jupyterlab/examples/test_examples.py", line 71, in <module>
    main()
  File "/home/runner/work/jupyterlab/jupyterlab/examples/test_examples.py", line 65, in main
    raise AssertionError(msg)
AssertionError: The following examples failed:
-/home/runner/work/jupyterlab/jupyterlab/examples/app
-/home/runner/work/jupyterlab/jupyterlab/examples/federated
-/home/runner/work/jupyterlab/jupyterlab/examples/filebrowser
[IPKernelApp] WARNING | Parent appears to have exited, shutting down.
Error: Process completed with exit code 1.
```

![image](https://github.com/user-attachments/assets/7b94d187-abb6-4cb9-bf0e-337fa63c250c)

jtpio:
> I think we will fix this with #16765 or #16751 (whichever we decide to go with).

[...]

PR #16814, opened 2024-09-26
Improve notebook output stream processing

davidbrochart:
## References

See #16812.

## Code changes

Take a shortcut when notebook stream outputs don't have `\r` or `\b` characters.

## User-facing changes

Better performances when opening notebooks with lots of text outputs.

None.

### DIFF
--- a/packages/outputarea/src/model.ts
+++ b/packages/outputarea/src/model.ts
@@ -526,6 +526,9 @@ namespace Private {
     if (text === undefined) {
       text = '';
     }
+    if (!(newText.includes('\b') || newText.includes('\r'))) {
+      return text + newText;
+    }
     let idx0 = text.length;
     for (let idx1 = 0; idx1 < newText.length; idx1++) {
       const newChar = newText[idx1];


packages/outputarea/src/model.ts
```diff
@@ -526,6 +526,9 @@ namespace Private {
     if (text === undefined) {
       text = '';
     }
+    if (!(newText.includes('\b') || newText.includes('\r'))) {
+      return text + newText;
+    }
```
krassowski, line 531: That's a good fast-track. Further down I guess we can also make it jump to index of `\b` or `\r` rather than process character by character. I can work on that.

packages/outputarea/src/model.ts
```diff
@@ -526,6 +526,9 @@ namespace Private {
     if (text === undefined) {
       text = '';
     }
+    if (!(newText.includes('\b') || newText.includes('\r'))) {
```
krassowski, line 529: Should it also consider `'\n'`? There is some special-casing for it in the logic?

davidbrochart: No, `\n` is not a special character unless there is `\b` or `\r` in the text.

davidbrochart:
I wouldn't flag it as a bug, but as an enhancement.

packages/outputarea/src/model.ts
davidbrochart: Yes, that would be one of the possible further optimizations.

davidbrochart:
Test [Linux (examples, 3.12)](https://github.com/jupyterlab/jupyterlab/actions/runs/11048186761/job/30691000848?pr=16814) seems to also fail on main, and [Visual Regression Documentation](https://github.com/jupyterlab/jupyterlab/actions/runs/11048186755/job/30690983273?pr=16814) seems to be unrelated?

krassowski:
Yup, tracked in https://github.com/jupyterlab/jupyterlab/issues/16772

krassowski (approved):
Thanks @davidbrochart!

PR #16815, opened 2024-09-26
Improve performance of concatenating stream strings

krassowski:
## References

- A part of the fixes for https://github.com/jupyterlab/jupyterlab/issues/16812
- Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16814

## Code changes

- [x] Adds a performance unit test ensuring that the logic is not using quadratic or worse complexity
  - should fail on CI in the first commit
- [ ] Changes the logic to iterate over the text in chunks rather than character-by-character

## User-facing changes

Opening notebooks with streams which include `\n` is faster

None


davidbrochart:
Awesome, impressive speedup!

davidbrochart (approved):

PR #16824, opened 2024-10-01
Underscore visibility issue in filenames with 'Ubuntu' font

### LINKED ISSUE #16820: Invisible underscores in File Browser
## Description
Since I have updated my Ubuntu to 24.04 LTS version there is a glitch appeared in Jupyter Lab file browser - all underscores in file/folder names are visually replaced with spaces. When you put the cursor over the folder/file name, the popped up window contains the correct name (with underscores). The screenshot is attached below.

This issue is relevant to Google Chrome Browser, Firefox works fine.

## Reproduce

1. Open JupyterLab in Google Chrome.
2. Open File Browser (Folder icon in the left menu or Ctrl+Shift+F).
3. Find or create a folder/file with underscores in the name (e.g. "my_folder").
4. All underscores in the name are visually replaced with spaces (e.g. "my folder"). However if you put the cursor over the folder, the popped up hint displays the correct name (with underscores).

Screenshot from 2024-09-26 16-39-54

## Expected behavior

The folder/file name is displayed with underscores (e.g. "my_folder" instead of "my folder")

## Context

- Operating System and version:  Ubuntu 24.04.1 LTS
- Browser and version:  Google Chrome Version 128.0.6613.137
- JupyterLab version:   4.2.5

Darshan808:
## References

#16820 

### Problem:
When using the 'Ubuntu' font on the default 100% zoom level in JupyterLab, underscores in filenames were hidden. This issue was observed in the directory listing where filenames with underscores were not fully visible.

### Solution:
I explored two solutions to resolve this:
1. Adding padding to the `.jp-Dirlisting-itemtext` element.
2. Adjusting the line-height of the `.jp-Dirlisting-itemtext` class.

I chose to adjust the line-height to `1.2` because it solves the problem without affecting the external layout or adding unnecessary space between elements, as padding would. This keeps the fix more contained and focused on typography.

### Testing:
Tested at 100% zoom with the 'Ubuntu' font, and the underscores are now visible without layout issues.

### Before
Screenshot from 2024-10-01 22-28-29

###After
Screenshot from 2024-10-01 22-28-44

### DIFF
--- a/packages/filebrowser/style/base.css
+++ b/packages/filebrowser/style/base.css
@@ -316,6 +316,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   user-select: none;
+  line-height: 1.2;
 }
 
 .jp-DirListing-item:has(.jp-DirListing-itemText:focus-visible) {


JasonWeill:
@Darshan808 I can ask about this at our next regular meeting. Thanks for opening this!

JasonWeill:
I see no perceptible difference with this change, compared with 4.2.5, on macOS and Firefox.

Darshan808:
@JasonWeill 
The issue with hidden underscores occurs specifically with the 'Ubuntu' font, the default in Ubuntu OS. So far, I haven't encountered this problem with any other fonts I've tested.
The reason you may not have seen any visible changes is likely due to how different fonts handle line-height. For example, setting line-height: 1.2 has a more noticeable effect with the 'Ubuntu' font compared to others like 'Poppins'.
Also this change provides consistent spacing without negatively affecting readability or design. It's a minimal adjustment that prevents issues with the 'Ubuntu' font while maintaining a clean look across different fonts.

Darshan808:
> @Darshan808 I can ask about this at our next regular meeting. Thanks for opening this!
@JasonWeill Could you please update me on the outcome of the meeting regarding this?

JasonWeill:
@Darshan808 Jupyter projects are not part of Hacktoberfest this year, but we still welcome your contributions. Thanks again!

srdas:
I tried to reproduce the error using `4.2.5 (dev), 4.3.0a2 (app)` on Linux Mint (a derivative of Ubuntu). I am unable to reproduce it after pulling the branch and removing line 319 : `  line-height: 1.2;` in `/packages/filebrowser/style/base.css` 
image
I changed the line height to different values, and the underscore remains. Also tested changing the browser resolution from 100% with no impact on the display. 

This appears to be a niche issue that is not originating from Jupyter Lab but a specific version of Ubuntu, and may be automatically handled in new releases of Jupyter Lab. 

I would recommend leaving it untouched for now, given that changing the line height affects the display on all systems that are unaffected, which is mostly all. This is already suggested here: https://github.com/jupyterlab/jupyterlab/issues/16820#issuecomment-2382910779
Visually, increasing the line height to 1.2 leaves a little bit more spacing between items and (IMHO) impairs the visual look and feel in addition to losing some valuable real estate in the file browser.

krassowski:
I was able to reproduce the original issue on the latest Ubuntu. I do not recall what browser it was with and if only certain zoom sizes were affected. I recall it only started in the latest Ubuntu release so Mint may not be affected yet

krassowski:
This is not JupyterLab specific though see for example https://github.com/microsoft/vscode/issues/227583

krassowski:
@Darshan808 it looks like this PR was opened against 4.2.x branch rather than master. Can you change the target branch?

Darshan808:
> @Darshan808 it looks like this PR was opened against 4.2.x branch rather than master. Can you change the target branch?
Oh, that was a silly mistake! I've changed the target branch now—done!

JasonWeill:
@Darshan808 @krassowski Issue #16820 was noted as occurring only in 4.2.x and not in the 4.3.0 beta, so I think the merge into the `4.2.x` branch was intended. See @Darshan808 's comment here: https://github.com/jupyterlab/jupyterlab/issues/16820#issuecomment-2382910779

Darshan808:
Oh, I overlooked that. Yes, the PR was intended to be created in the 4.2.x branch.

krassowski:
I will close it as we are not releasing new versions on 4.2.x anymore - thank you for the work here though!

Issue #14349, opened 2023-04-09
Galata error when running individual tests due to in-page helper


krassowski:
When running a single test which renames a file, e.g. by creating and renaming it like:

```
await page.notebook.createNew(fileName);
```

I consistently see a timeout with:

```
Error: page.evaluate: Target closed

   at src/contents.ts:294

  292 |       // Rename through REST API does not propagate to opened widgets
  293 |       // => Use galata in-page if page is available
> 294 |       return await this.page.evaluate(
      |                              ^
  295 |         async ({ pluginId, oldName, newName }) => {
  296 |           const docManager = (await window.galata.getPlugin(
  297 |             pluginId

    at ContentsHelper.renameFile (jupyterlab/galata/src/contents.ts:294:30)
    at NotebookHelper.createNew (jupyterlab/galata/src/helpers/notebook.ts:1207:25)
````

This appears related to the in-page galata helper having `_app` as undefined. Strangely `_app` is set in getter:

https://github.com/jupyterlab/jupyterlab/blob/89154f525578e3c79d9f688be1e67605f6d2d443/galata/extension/src/global.ts#L742-L747

but this getter may not have been called sine the test was initialized; `ContentsHelper.renameFile` indeed calls `window.galata.getPlugin` which accesses `_app` and not `.app`.

Invoking the getter, e.g. "fixes" the problem:

```diff
window.galata = new GalataInpage();
+ window.galata.app;
```

Now, I wonder if we could just assign app in the extension:

https://github.com/jupyterlab/jupyterlab/blob/89154f525578e3c79d9f688be1e67605f6d2d443/galata/extension/src/index.ts#L24-L39

PR #16831, opened 2024-10-03
Conditional rendering of 'Add' and 'or' buttons during shortcut editing

Darshan808:
### Summary:

This pull request addresses two key UI issues encountered while editing shortcuts within a row. The incorrect display behavior of the "Add" button and "or" text has been rectified, enhancing the overall user experience during shortcut management.

### Detailed Changes:

1. **Issue with 'or' appearing unnecessarily:**
   - When editing a shortcut, the 'or' text was appearing after last shortcut even though there was nothing after that.
   - **Solution:** I introduced a condition that ensures the 'or' text does not appear for the last shortcut in the row when a shortcut is being edited. This guarantees that 'or' is only shown when there are subsequent shortcuts.

2. **'Add' button appearing during edits:**
   - Another issue arose where the "Add" button would still appear while editing other shortcuts (except the first one). This was due to an incorrect condition that only checked whether the first shortcut was being edited, without considering other cases.
   - **Solution:** I modified the condition to check whether any shortcut is being edited, not just the first one. This ensures the "Add" button remains visible during shortcut editing, except for cases where editing is in progress on the first shortcut.

### Testing:

- Manual testing was performed by editing multiple rows of shortcuts to verify that:
  - 'or' only appears when there are additional shortcuts.
  - The "Add" button disappears whenever any shortcut is being edited.


Darshan808:
@JasonWeill Could this PR be labeled 'hacktoberfest-accepted' ? I would really appreciate it.

JasonWeill:
@Darshan808 Done!

packages/shortcuts-extension/src/components/ShortcutItem.tsx
```diff
@@ -250,7 +250,9 @@ export class ShortcutItem extends React.Component<
         {this.isLocationBeingEdited(index)
           ? this.getShortCutAsInput(binding, index)
           : this.getShortCutForDisplayOnly(binding)}
-        {this.getOrDiplayIfNeeded(index < nonEmptyBindings.length - 1)}
+        {!(index===this._nonEmptyBindings.length-1 &&
```
krassowski, line 253: Thanks! Just leaving a note that the linter job is failing on CI with:

```
+ jlpm run prettier:check
Checking formatting...
[warn] packages/shortcuts-extension/src/components/ShortcutItem.tsx
[warn] Code style issues found in the above file. Run Prettier to fix.
+ echo 'Please run `jlpm run prettier` locally and push changes'
+ exit 1
Please run `jlpm run prettier` locally and push changes
```

I think this is because of missing spaces around `===` and `-` but there may be more small reformatting needed

Darshan808: @krassowski Yes, the issue was related to code formatting, and I resolved it.

krassowski:
@Darshan808 would you be open to adding a visual test case for this behaviour? Let me know if you need help with this.

You could follow the test I added recently:

https://github.com/jupyterlab/jupyterlab/blob/53376b3996b6680aaefec31858411b33b48d640f/galata/test/jupyterlab/settings.test.ts#L254-L290

which generated https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/jupyterlab/settings.test.ts-snapshots/settings-shortcuts-conflict-jupyterlab-linux.png

Also see https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#visual-regression-and-ui-tests

Darshan808:
@krassowski 
I'd like to add a visual test case, but I'm encountering a small issue when running individual tests. I researched the problem and found a related issue: https://github.com/jupyterlab/jupyterlab/issues/14349 . Has this been resolved, or could my issue be caused by something else ? I'm still looking into it to find the cause, but any help would be appreciated if you've encountered this before.
Screenshot from 2024-10-08 12-16-27

krassowski:
Does it happen when running an existing test? Can you access the JupyterLab at port 8888 before/during running the test?
By default playwright runs in headless mode but you can set PWDEBUG=1 to see the browser window to debug it

Darshan808:
@krassowski 
Yes it is happening when i run `jlpm playwright test test/jupyterlab/settings.test.ts`
I ran test in `--headed` mode, to see what's going on browser!
Here is the result.

https://github.com/user-attachments/assets/ad5e2abc-c368-403d-b7e2-d2c465940597

1st test result:

Screenshot from 2024-10-08 19-40-28

2nd test result:

Screenshot from 2024-10-08 19-41-28

Darshan808:
@krassowski 
The issue continues even after creating a fresh project and adding tests as per the documentation https://github.com/jupyterlab/jupyterlab/blob/main/galata/README.md . When renaming a notebook, the rename dialog box closes instantly, and the result shows a cross icon on 'return await this.page.evaluate'. Is this problem specific to my device, or could I be doing something wrong?

Darshan808:
@krassowski 

After hours of debugging,  I think I found something meaningful.

During testing, when a custom file is to be created using createNew method, 
e.g: `await page.notebook.createNew(ipynbFileName);`
It invokes renameFile method to rename a file: 
`  await this.contents.renameFile(
      '${currentDir}/${assignedName}',
      '${currentDir}/${name}'
    ); 
`
 In renameFile function, we have
 `
           const docManager = (await window.galata.getPlugin(
            pluginId
          )) as IDocumentManager;
`

#### Inside this getPlugin method, we have code that reads plugin from `app.pluginRegistry._plugins.get(pluginId)` where app is our jupyterFrontEnd

image

#### But the properties of app shows that there is no pluginRegistry data member.

Screenshot from 2024-10-08 21-40-39

#### So, if we replace `app.pluginRegistry._plugins` by `app._plugins`, the issue will be solved

image

#### But two of the tests were still failing due to mismatch of snapshot, the snapshots are identical but not exactly same. What can we do about this ?

krassowski:
Thank you for looking into this!

You should not worry about minor screenshot mismatches when running locally as your system would have different fonts than the virtual machine on CI. The important thing is that these pass on CI. Once you have a test working locally you can just commit the code and get the snapshot from CI by asking the bot to update snapshots.

Darshan808:
@krassowski 
Added the required visual test case.

krassowski:
Thank you! It looks like it needs linting :)

Darshan808:
> Thank you! It looks like it needs linting :)

Fixed it!
I'll make sure to check the linting before every commit from next time.

krassowski:
You can install pre-commit as it is configured on this repo :)

galata/test/jupyterlab/settings.test.ts
```diff
@@ -288,3 +288,49 @@ test('Keyboard Shortcuts: overwriting a shortcut can be cancelled', async ({
...
+    .locator('.jp-Shortcuts-Or-Forced')
+    .count();
+
+  expect(orCount).toBe(shortcutKeysCount);
+  expect(orForcedCount).toBe(shortcutKeysCount - 1);
+
+  const shortcutKey = firstRow.locator('.jp-Shortcuts-ShortcutKeys').first();
+  await shortcutKey.click();
+
+  // After clicking, assert that 'jp-Shortcuts-Or' count equals 'jp-Shortcuts-Or-Forced'
```
krassowski, line 324: Without looking at the code it is not clear to me why this should be the case. Ideally integration tests should be easy to understand and describe behaviour in a way independent of implementation details. The fact that the number of elements with these classes is equal seems disconnected from the behaviour being fixed here.

I think I would lean more towards a test case capturing a small screenshot of the relevant area as this is easy to validate.

Darshan808: I focused on testing the exact behavior that was coded since it was my first time writing a proper visual test, so I approached it with more of a programmer mindset. I get what you're saying though—I'll aim to keep it simple and easily understandable.

Since this PR is about hiding the 'or' button and the 'Add' button during shortcut editing, how about I take a snapshot during the test and match it?"

Darshan808:
@krassowski 
How does this sound? Or would it be better to add a custom shortcut and validate the UI, as that might provide better validation?
Screenshot from 2024-10-10 21-04-31

Darshan808:
@krassowski 
When the shortcut is clicked to edit, it expands and takes about 0.5ms. To handle this, I added a short delay before capturing the screenshot. What are your thoughts on this approach?

krassowski:
> When the shortcut is clicked to edit, it expands and takes about 0.5ms. To handle this, I added a short delay before capturing the screenshot. What are your thoughts on this approach?

Yes, it's 0.5s:

https://github.com/jupyterlab/jupyterlab/blob/2db50b95e612e0adb315341886ba76eb01e7ab17/packages/shortcuts-extension/style/base.css#L34

I think this is fine. Playwright has auto-waiting checking for stability but the heuristic of two consecutive frames being the same sometimes matches before the animation kicked off so I think this is a reasonable approach here.

On a tangent note, I am not sure if this animation should take half a second. In my experience a good rule of thumb for inconsequential animation is for them to be shorter than 0.5s. Maybe 300ms? Thoughts?

Darshan808:

Darshan808:
Yes, I think 300ms will work fine.

krassowski:

galata/test/jupyterlab/settings.test.ts
```diff
@@ -288,3 +288,30 @@ test('Keyboard Shortcuts: overwriting a shortcut can be cancelled', async ({
...
+  await filterInput.locator('input').fill('merge cell below');
+
+  const shortcutsContainer = page.locator(
+    '.jp-Shortcuts-ShortcutListContainer'
+  );
+  const firstRow = shortcutsContainer.locator('.jp-Shortcuts-Row').first();
+
+  const shortcutKey = firstRow.locator('.jp-Shortcuts-ShortcutKeys').first();
+  await shortcutKey.click();
+  await page.waitForTimeout(300);
```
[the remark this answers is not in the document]
krassowski: I think both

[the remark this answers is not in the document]
Darshan808: Oh, there was a small misunderstanding. I'll fix it immediately.

jtpio (approved):
Thanks!

Issue #16185, opened 2024-04-18
Incorrect disposal of LSP connection due to access to undefined `serverRequests`


krassowski:
## Description

Sometimes when switching kernels I see an uncaught error:

```ts
connection.ts:271 Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at LSPConnection.dispose (connection.ts:271:12)
    at DocumentConnectionManager.unregisterDocument (connection_manager.ts:428:20)
    at NotebookAdapter.disconnect (adapter.ts:304:30)
    at NotebookAdapter.reloadConnection (adapter.ts:397:10)
    at NotebookAdapter.onKernelChanged (notebooklspadapter.ts:209:12)
```

This is raised from:

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/packages/lsp/src/connection.ts#L267-L273

and happens because `serverRequests` have no initializer in constructor nor in class definition: 

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/packages/lsp/src/connection.ts#L229-L232

These are only initialized after server initialization in `afterInitialized`:
 
https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/packages/lsp/src/connection.ts#L481-L483

## Expected behavior

- Disposal of connections which have not yet initialized does not raise an error.
- There is a unit test case for this scenario

## Context

- JupyterLab version: 4.1.5

AnanthaKrishnan63:
Is this issue still open? If so, I would like to open a PR with my solution to it.

jtpio:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/16834, thanks!

PR #16834, opened 2024-10-04
Fix incorrect access to uninitialised `serverRequests`

kaushalaaditya99:
# References

This is a potential solution to issue #16185. It modifies the connection.ts to check whether the LSP connection is initialized before disposal and it creates a test case to examine this case.
There was one other pull request from hanlaman that implemented the same solution for checking whether the LSP connection was initialized before running the dispose. The difference between mine and the other user's changes is that I included a unit test as well, as requested in the description.

# Code Changes:
Added an if statement to the problematic code that checks if the serverrequest has been initialized before trying to dispose it.

For the test case, I added an import statement for LSPConnections and a describe test case that creates a new LSPconnection and tries to dispose it before it is properly initialized. It then checks isDisposed to see if it worked out

### DIFF
--- a/packages/lsp/src/connection.ts
+++ b/packages/lsp/src/connection.ts
@@ -268,7 +268,8 @@ export class LSPConnection extends LspWsConnection implements ILSPConnection {
     if (this.isDisposed) {
       return;
     }
-    if (this.serverRequests) {    // `serverRequests` may be undefined if dispose is called during initialization sequence
+    if (this.serverRequests) {
+      // `serverRequests` may be undefined if dispose is called during initialization sequence
       Object.values(this.serverRequests).forEach(request =>
         request.clearHandler()
       );
--- a/packages/lsp/src/connection.ts
+++ b/packages/lsp/src/connection.ts
@@ -268,7 +268,7 @@ export class LSPConnection extends LspWsConnection implements ILSPConnection {
     if (this.isDisposed) {
       return;
     }
-    if (this.serverRequests) { /*if statement to check if serverrequest is empty*/
+    if (this.serverRequests) {    // `serverRequests` may be undefined if dispose is called during initialization sequence
       Object.values(this.serverRequests).forEach(request =>
         request.clearHandler()
       );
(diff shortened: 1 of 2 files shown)


packages/lsp/src/connection.ts
```diff
@@ -268,9 +268,11 @@ export class LSPConnection extends LspWsConnection implements ILSPConnection {
     if (this.isDisposed) {
       return;
     }
-    Object.values(this.serverRequests).forEach(request =>
-      request.clearHandler()
-    );
+    if (this.serverRequests) { /*if statement to check if serverrequest is empty*/
```
krassowski, line 271: Looks good! I would adjust the comment, something like:

```suggestion
    if (this.serverRequests) {    // `serverRequests` may be undefined if dispose is called during initialization sequence
```
To avoid this:

image

:)

packages/lsp/test/connection_manager.spec.ts
```diff
@@ -99,5 +100,12 @@ describe('@jupyterlab/lsp', () => {
         expect(manager.disconnectDocumentSignals).toHaveBeenCalled();
       });
     });
+    describe('#onUninitializedLSPConnectionDisposed', () => {
```
krassowski, line 103: We are describing the behaviour of the `dispose()` method here:

```suggestion
    describe('#dispose()', () => {
```

krassowski (commented):
Thank you @kaushalaaditya99 - please see small suggestions :)

packages/lsp/test/connection_manager.spec.ts
```diff
@@ -99,5 +100,12 @@ describe('@jupyterlab/lsp', () => {
         expect(manager.disconnectDocumentSignals).toHaveBeenCalled();
       });
     });
+    describe('#dispose()', () => {
+      it('should dispose of an uninitialized connection without errors', () => {
+        const connection = new LSPConnection();
```
krassowski, line 105: The test is failing because `LSPConnection` constructor requires `options` argument.

jtpio: Fixed with the latest commits.

jtpio (approved):
Thanks!

PR #16836, opened 2024-10-07
Fix docs snapshot

jtpio:
## References

Even though we pin on Python 3.10.6 here:

https://github.com/jupyterlab/jupyterlab/blob/84572d6cea7eb73f2cfb2a0cc0cec4387b3462be/.github/workflows/galata.yml#L95-L96

The UI tests documentation check seems to now be failing consistently recently, as noticed on recent PRs.

Probably related to a Python build change on GitHub Actions.

## Code changes

| Actual | Expected |
| --------- | -------------- |
|  image | image |

**Diff**

[python-diff.webm](https://github.com/user-attachments/assets/baaadadf-1793-4339-89a7-53ae663b6b81)

## User-facing changes

None

None


krassowski (approved):
Thank you @jtpio!

If this was in the tests I would suggest to use masking like in:

https://github.com/jupyterlab/jupyterlab/blob/84572d6cea7eb73f2cfb2a0cc0cec4387b3462be/examples/example.spec.ts#L53-L55

but since this is for docs I think it is worth updating it once in a while instead to avoid users being confused by the pink rectangles on snapshots.

krassowski:

jtpio:
Yeah hopefully these updates don't happen too often.

jtpio:
Of note, the latest Playwright `1.48.0` release now seems to allow for Websocket routing: https://github.com/microsoft/playwright/releases/tag/v1.48.0

> New methods [page.routeWebSocket()](https://playwright.dev/docs/api/class-page#page-route-web-socket) and [browserContext.routeWebSocket()](https://playwright.dev/docs/api/class-browsercontext#browser-context-route-web-socket) allow to intercept, modify and mock WebSocket connections initiated in the page.

Maybe this could be used to intercept the kernel info reply in this example and override the banner?

PR #16838, opened 2024-10-07
Fix UI test snapshot

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16765

## Code changes

Fix a snapshot that got updated in https://github.com/jupyterlab/jupyterlab/pull/16765.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #16839, opened 2024-10-07
Fix alignment of headers in file browser

krassowski:
## References

Follow up to https://github.com/jupyterlab/jupyterlab/pull/16765 addressing https://github.com/jupyterlab/jupyterlab/pull/16765#issuecomment-2394007950

## Code changes

When computing the width of the column headers account for the resize handle and extra padding around rows:
- deducing the width of the resize handle (proportionally)
- adding the padding to the two outermost resizable columns

## User-facing changes

Text in left-centered columns now aligns more nicely

image

None


krassowski:
bot please update snapshots

jtpio:
It looks like the workflow to update snapshots produces snapshots that then need to be reverted, like in https://github.com/jupyterlab/jupyterlab/pull/16765 and https://github.com/jupyterlab/jupyterlab/pull/16838.

jtpio (approved):
Thanks!

Issue #16837, opened 2024-10-07
Failing `check-links` on CI


jtpio:
The `check-links` check has been failing on CI for the past few days with the following error:

![image](https://github.com/user-attachments/assets/80e6cfeb-6c15-448f-8ab6-72a82d3d287c)

```
FFFFF                                                                    [ 80%]
=================================== FAILURES ===================================
_ /home/runner/work/jupyterlab/jupyterlab/docs/source/user/extensions.rst: https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+4 _
https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+4: 429: Too Many Requests
_ /home/runner/work/jupyterlab/jupyterlab/docs/source/user/extensions.rst: https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions _
https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions: 429: Too Many Requests
_ /home/runner/work/jupyterlab/jupyterlab/docs/source/user/extensions.rst: https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Mime+Renderers _
https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Mime+Renderers: 429: Too Many Requests
_ /home/runner/work/jupyterlab/jupyterlab/docs/source/user/extensions.rst: https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Prebuilt _
https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Prebuilt: 429: Too Many Requests
_ /home/runner/work/jupyterlab/jupyterlab/docs/source/user/extensions.rst: https://pypi.org/search/?c=Framework+%3A%3A+Jupyter+%3A%3A+JupyterLab+%3A%3A+Extensions+%3A%3A+Prebuilt _
[...]

PR #16843, opened 2024-10-08
Ignore `pypi.org` links for now

jtpio:
## References

https://github.com/jupyterlab/jupyterlab/issues/16837

Maybe we can leave https://github.com/jupyterlab/jupyterlab/issues/16837 open for now, or create a new issue to improve the `check-links` check with the suggestions mentioned in https://github.com/jupyterlab/jupyterlab/issues/16837.

The main motivation is get a full green CI check on PRs.

## Code changes

Ignore `pypi.org` links for now, to help make the `check-links` check pass.

## User-facing changes

None

None


krassowski (approved):
Thanks @jtpio!

Shall we also open an issue to track improving the check links job (or keep the existing one open but rename it).

jtpio:
Thanks for the review!

Yes I think we can just rename the issue since there is already some discussion there with good points.

Issue #12402, opened 2022-04-14
Anchors Are Not Working for Japanese heading


ruoyu0088:
## Description

Links with Japanese anchor cannot jump to the heading, for example ` [日本語](test.ipynb#日本語)`.

## Reproduce

Create a notebook `test.ipynb` with two heading cell:

```
## 日本語
```

```
## English
```

Create a markdown file with the following links:

```
* [English](test.ipynb#English)
* [日本語](test.ipynb#日本語)
```

Open the Markdown preview page, and click the two links. the "English" link works, but the "日本語" link doesn't work.

## Context

- Operating System and version: Windows 10
- Browser and version: Chrome  100.0.4896.75
- JupyterLab version: 3.3.3

jweill-aws:
I can reproduce this in JupyterLab 4. The ¶ links beside each heading generate working anchor links, such as `#日本語`, but using Markdown syntax does not generate working intrapage links for Japanese anchors.

jweill-aws:
An anchor link to a header with a mixture of ASCII and Japanese text, like `#Heading-日本語`, also does not work in Markdown syntax.

jweill-aws:
The characters `日本語` are rendered in the Markdown cell output as `%E6%97%A5%E6%9C%AC%E8%AA%9E`, the escaped bytes of the UTF-8 representation of these characters.

jweill-aws:
From the [URL Standard defined by WhatWG](https://url.spec.whatwg.org/#concept-url-fragment): "A [URL](https://url.spec.whatwg.org/#concept-url)’s fragment is either null or an [ASCII string](https://infra.spec.whatwg.org/#ascii-string) that can be used for further processing on the resource the [URL](https://url.spec.whatwg.org/#concept-url)’s other components identify. It is initially null."

To comply with this standard, we may have to define fragments as ASCII text for all heading links, even if those heading links contain non-ASCII characters.

jweill-aws:
This line in `rendermime` generates a header by replacing all spaces with hyphens. This is incomplete. The function needs to:
1. Replace all _non-ASCII characters_ with either nothing or an ASCII character such as a hyphen
2. Generate unique IDs for all headers, even if the logic in (1) above reduces headers to identical ASCII strings.

https://github.com/jupyterlab/jupyterlab/blob/7d23ead2a51c44e261243f916e46dea46ae1b837/packages/rendermime/src/renderers.ts#L835

ruoyu0088:
[...]

PR #16845, opened 2024-10-08
suffix encoding added for header ids

### LINKED ISSUE #1932: UUIDs for header ids to avoid the duplicate id problem
This is a followup to the discussion began in #1736 around using UUIDs as part of the automatic id generation scheme for headers. 

Source of the problem Right now, if you have a notebook and you have 

```markdown
### my header 

### my header
```

in the same notebook, both headers will have the same id (specifically `#my-header`). Aside from violating html standards for id uniqueness, this makes it impossible to link to the second instance of the header even though it assigned an anchor link. 

This problem is dramatically worse for jupyterlab due to the possibility of multiple notebooks being in the same window at the same time (not only because it is more likely that header text is shared between notebooks(in general), but because of the factorial growth of opportunities to conflict that arise with each additional notebook injecting ids into the namespace). 

One proposed solution to this problem is to instead build ids that contain UUIDs for identifying (at least) the notebook from the id originates.

The main problem I see with this approach is that this eliminates the ease of use for _writing_ links to different parts of the document. Not only does no one want to have to type `[link to header](#link-to-header-<header_uuid>) in order to link, what's worse is that there's no way to write the link without rendering the cell and looking at the actual anchor that was generated. 

But there are conceptual issues as well: for example, if you ever copy and paste a cell with a UUID already being present — do you create a new one and therefore break any links that also are copied and pasted?

There are a lot of issues embedded in these topics but I've tried to distill them to specific use cases for ids and where challenges might arise:

- [ ] Linking to headers headers in the same notebook 
    - [ ] Multiple headers have the same text, ids conflict
    - [ ] Manually specified ids that conflict with autogenerated ids
    - [ ] Addition of a new cell with a header with the same text "stealing" the targets of the postceding header (that it originally pointed to)
    - [ ] Cross notebook id conflicts, however the DOM resolves the ordering of the objects, the first id will be the referred item.
- [ ] Cross notebook linking 
- [ ] Copying and pasting cells
    - [ ] Copying and pasting cells between notebooks
- [ ] Multiple views on the same document (e.g., a live nbconvert html preview)
- [ ] Deterministic, robust link resolution not relying on content persistence
      E.g., Header has id **1** , changing the text of an header, render (generates a new id **2**), breaks links pointing to id **1** , changing the header text back to original, render (generates a new id **3**), does **1**==**3**?

onursercanyilmaz:
## References
Fixes #1932  #16795  #12402

## Code changes

Changes in renderers.tsx, as shown below.

```
  export function headerAnchors(node: HTMLElement): void {
    const headerNames = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
    for (const headerType of headerNames) {
      const headers = node.getElementsByTagName(headerType);
      for (let i = 0; i < headers.length; i++) {
        const header = headers[i];
        const parentElement = header.closest(
          '.lm-Widget.jp-Cell.jp-MarkdownCell.jp-Notebook-cell.jp-mod-rendered'
        );
        const parentIndex =
          parentElement?.getAttribute('data-windowed-list-index') ?? '0';
        header.id = renderMarkdown.createHeaderId(header, i, parentIndex);
        const anchor = document.createElement('a');
        anchor.target = '_self';
        anchor.textContent = '¶';
        anchor.href = '#' + header.id;
        anchor.classList.add('jp-InternalAnchorLink');
        header.appendChild(anchor);
      }
    }
  }
  
   export function createHeaderId(header: Element, index: number, parentIndex: string): string {
    return (`${header.textContent} ${index} ${parentIndex}` ?? '').replace(/ /g, '-');
  }
```

If a cell includes more than 1 header it adds 'i' (index) to id and after that adds the parent div's index info as header-{innerIndex}-{parentIndex}. Issue is about parentIndex problem but there was another problem on inner headers that solved too.

image

But still there was an issue that for example - if user adds #Part 1 as h1 and ##Part 1 as h2 - it creates same id.
So i added headerType parameter too

```
  export function createHeaderId(
    header: Element,
    index: number,
    parentIndex: string,
    headerType: string,
  ): string {
    return (`${header.textContent} ${index} ${parentIndex} ${headerType}` ?? '').replace(
      / /g,
      '-'
    );
  }
  
   export function headerAnchors(node: HTMLElement): void {
    const headerNames = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
    for (const headerType of headerNames) {
      const headers = node.getElementsByTagName(headerType);
      for (let i = 0; i < headers.length; i++) {
        const header = headers[i];
        const parentElement = header.closest(
          '.lm-Widget.jp-Cell.jp-MarkdownCell.jp-Notebook-cell.jp-mod-rendered'
        );
        const parentIndex =
          parentElement?.getAttribute('data-windowed-list-index') ?? '0';
        header.id = renderMarkdown.createHeaderId(header, i, parentIndex, headerType);
        const anchor = document.createElement('a');
        anchor.target = '_self';
        anchor.textContent = '¶';
        anchor.href = '#' + header.id;
        anchor.classList.add('jp-InternalAnchorLink');
        header.appendChild(anchor);
      }
    }
  }
```
  
image
image

But i think it seems not good. So I decided to encode suffix part
```
 function encodeSuffix(
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = `${index} ${parentIndex} ${headerType}`;

    // Use TextEncoder to convert the string to a Uint8Array
    const encoder = new TextEncoder();
    const data = encoder.encode(suffix);

    // Convert to Base64
    let binaryString = '';
    for (let i = 0; i < data.length; i++) {
      binaryString += String.fromCharCode(data[i]);
    }

    return btoa(binaryString);
  }

  export function createHeaderId(
    header: Element,
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = encodeSuffix(index, parentIndex, headerType);
    return (`${header.textContent} ${suffix}` ?? '').replace(/ /g, '-');
  }

```

image

- But furthermore if user types any emoji or non-english chars in headers, it will affect the id and the url
So we can add headerText as parameter in encodeSuffix function
```

  function encodeSuffix(
    headerText: string | null,
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = `${headerText} ${index} ${parentIndex} ${headerType}`;

    // Use TextEncoder to convert the string to a Uint8Array
    const encoder = new TextEncoder();
    const data = encoder.encode(suffix);

    // Convert to Base64
    let binaryString = '';
    for (let i = 0; i < data.length; i++) {
      binaryString += String.fromCharCode(data[i]);
    }

    return btoa(binaryString);
  }

  export function createHeaderId(
    header: Element,
    index: number,
    parentIndex: string,
    headerType: string
  ): string {
    const suffix = encodeSuffix(
      header.textContent,
      index,
      parentIndex,
      headerType
    );
    return (`${suffix}` ?? '').replace(/ /g, '-');
  }
}

```

## User-facing changes
- Users will now see the newly generated id in the URL when they click on header items, as the id attributes are assigned to the <a> tags as href values.

image

image

image

- [x] Linking to headers in the same notebook
- [x] Multiple headers have the same text, ids conflict solved by adding parent div index info
- [x] encodeSuffix solves non-english char problem
- [x] Adding headers in same or different cells with different id information
- [x] No any non-english char in id or link -> #12402
- [x] Not created a uuid because it will create different id info for each render. But giving static info to convert it base64 is better to linking headers 

Here’s the detailed explanation:

### **Cause:**
In JupyterLab, when assigning IDs to markdown headers, each cell (node) operates in isolation, meaning only the headers within that cell are considered. Since there is no global ID system, the following issues arise:

1. When the same or different cells use the same header text, the same IDs are repeated. For example:
    - If a **h1** tag is created in Cell 1 with the text "Header," its ID will be `Header`.
    - If the same "Header" is created in Cell 2, this header will also be assigned the ID `Header`.

2. Additionally, when emojis or special characters are used in headers, they are included in the IDs. For example, if a header text is "🚀 Start," it will be assigned the ID `🚀-Start`. This situation reduces the readability and consistency of the IDs.

### **Development:**
Using a **counter** to create unique IDs based on header text might seem logical, but since each cell operates independently, this method fails. Specifically:

- The **node** resets to zero in each cell. For example:
    - If there are two **h1** headers in Cell 1, they might be assigned IDs `Header` and `Header-1`.
    - However, if the same headers are used in Cell 2, the count starts again from zero, resulting in the same IDs: `Header` and `Header-1`.

- In this case, the **counter** method can only apply within that cell, meaning it cannot ensure global uniqueness across cells. As cells are isolated, the **counter** solution proves inadequate for establishing unique IDs.

**Example:**
Cell 1:
```markdown
# Header
# Header
```
IDs: `Header`, `Header-1`

Cell 2:
```markdown
# Header
# Header
```
IDs: again `Header`, `Header-1`, leading to conflicts.

### **Result:**
To resolve this issue, I got a unique **parentId** (or parentIndex) for each cell to ensure the headers have unique IDs. With this method:

1. Each header text had its cell's **parentId** value appended (Each cell's div tag has a data-windowed-list-index attribute value = parentIndex), allowing for differentiation even if the header text was the same.
   
2. Also encoded the header text to make it unique and began using it in Base64 format. This way, headers were made uniquely identifiable.

**Example:**
In Cell 1 with two headers:
```markdown
# Header
# Header
```
These headers might have IDs like `0-Header` and `0-Header-1` (where **0** represents the parentIndex).

In Cell 2, again using the same headers:
```markdown
# Header
# Header
```
These headers would then be assigned IDs `1-Header` and `1-Header-1`, since the parentIndex this time is **1**.

This method ensures that headers in each cell remain unique in large and complex notebooks, thus preventing ID conflicts.

Using a global counter can lead to performance issues and complexity, as it requires tracking header occurrences across all cells, making it less efficient in managing unique IDs within a large notebook.

### DIFF
--- a/packages/rendermime/src/renderers.ts
+++ b/packages/rendermime/src/renderers.ts
@@ -403,14 +403,46 @@ export namespace renderMarkdown {
     translator?: ITranslator;
   }
 
+  function encodeSuffix(
+    headerText: string | null,
+    index: number,
+    parentIndex: string,
+    headerType: string
+  ): string {
+    const suffix = `${headerText} ${index} ${parentIndex} ${headerType}`;
+
+    // Use TextEncoder to convert the string to a Uint8Array
+    const encoder = new TextEncoder();
+    const data = encoder.encode(suffix);
+
+    // Convert to Base64
+    let binaryString = '';
+    for (let i = 0; i < data.length; i++) {
+      binaryString += String.fromCharCode(data[i]);
+    }
+
+    return btoa(binaryString);
+  }
+
   /**
    * Create a normalized id for a header element.
    *
    * @param header Header element
    * @returns Normalized id
    */
-  export function createHeaderId(header: Element): string {
-    return (header.textContent ?? '').replace(/ /g, '-');
+  export function createHeaderId(
+    header: Element,
+    index: number,
+    parentIndex: string,
+    headerType: string
+  ): string {
+    const suffix = encodeSuffix(
+      header.textContent,
+      index,
+      parentIndex,
+      headerType
+    );
+    return `${suffix}`.replace(/ /g, '-');
   }
 }
 
@@ -1162,7 +1194,17 @@ namespace Private {
       const headers = node.getElementsByTagName(headerType);
       for (let i = 0; i < headers.length; i++) {
         const header = headers[i];
-        header.id = renderMarkdown.createHeaderId(header);
+        const parentElement = header.closest(
+          '.lm-Widget.jp-Cell.jp-MarkdownCell.jp-Notebook-cell.jp-mod-rendered'
+        );
+        const parentIndex =
+          parentElement?.getAttribute('data-windowed-list-index') ?? '0';
+        header.id = renderMarkdown.createHeaderId(
+          header,
+          i,
+          parentIndex,
+          headerType
+        );
         const anchor = document.createElement('a');
         anchor.target = '_self';
         anchor.textContent = '¶';


HaudinFlorence:
@onursercanyilmaz Thanks for the details explanation you gave on your previous PR and reported here.
I just tested a simple notebook and console. log the information collected to build the suffix before encoding it. The information are correct when I start a new jupyterlab and open the notebook. The only issue I observed is when modifying the notebook by changing headers, saving it:, the `ParentIndex` and the `index` seem then to be zero in any case. It seems that jupyterlab needs to be refresh to get back the correct result. Maybe this is normal and expected as a behavior. But that the reason why I was not able to get the correct `ParentIndex` and `index` previously.

HaudinFlorence:
There is also this case that could (very rarely) happen: opening 2 notebooks with similar contents or the same headings contents for the first cell for example. In that situation it is possible to have in the DOM 2 elements having the same id as shown on the following example:
Screenshot from 2024-10-09 18-37-10
Screenshot from 2024-10-09 18-37-18
This should be prevented if possible with introducing an identifier related to the notebook.

krassowski (changes requested):
@onursercanyilmaz thank you for working on this!

The problem with changing the format of the anchors is that users have embedded them as cross-links in their notebooks, so any change is breaking in a very bad way - basically links between existing notebooks will stop working. While I do like the approach of using based64 encoding, I believe that we should find a way to make this change backward compatible, for example by keeping the old header ID format AND add the new one on top of that (which would require rendering two HTML elements for each heading).

Regarding the issues that @HaudinFlorence brought up: when two notebooks are open side-by-side JupyterLab should be able to resolve the ambiguity because the notebook name is also embedded in the URL. There is dedicated routing code dealing with hash IDs - I don't think this should be addressed in this PR. I also don't think we should embed the notebook name in the ID as that can change frequently outside of Jupyter.

krassowski:
Closing since this was in draft in for over 1.5 year and as of now uses a backward-incompatible approach. Thank you for exploring this!

Issue #11547, opened 2021-11-25
Status bar is overpopulated with information already on the screen


SylvainCorlay:
It seems that a lot of the information currently shown in the status bar is not really necessary since it is already elsewhere on the page.

<img width="898" alt="Screenshot 2021-11-25 at 11 22 17" src="https://user-images.githubusercontent.com/2397974/143424127-faeb7986-827e-47b2-ad5c-bd01f24b8628.png">

We could easily remove:
 - kernel status,
 - notebook name,
 - the button that opens the left panel for open tabs, kernels, and terminal.

Then we could start using that real estate for other information.

krassowski:
As a general rule of thumb, I would advocate making it user-configurable rather than removing things altogether. Also, some things are useful in multi-panel mode but not in Simple mode - those could be shown/hidden depending on the mode.

Quick comments:
-  +1 for shortening or removing the kernel name (maybe kernelspaecs could provide a short version of the name?)
- kernel status as what is after pipe `|` like `Disconnected`, `Inactive`, etc is not shown anywhere else; personally I would prefer to:
   - keep it, or
   - show it somewhere else prominently, and/or
   - only show it when different from `Idle` or `Busy`
- `Mode:` prefix could be removed/replaced by an icon
- notebook name is very useful when you have multiple notebooks open. I would highly prefer to keep it when in multi-panel mode. +1 for hiding it when in `Simple` mode
- sorry about the LSP verbosity; we are open to suggestions for a shorter text but I could not find a good one.
- "the button that opens the left panel for open tabs, kernels, and terminal." - what does it do in the Simple mode? If it exits simple mode or does nothing, it should be hidden in the Simple mode.

SylvainCorlay:
> "the button that opens the left panel for open tabs, kernels, and terminal." - what does it do in the Simple mode? If it exits simple mode or does nothing, it should be hidden in the Simple mode.

It is there in simple mode, which still has the side panel, so it behaves exactly the same as in the normal mode.

I think that we should remove this button altogether. Another issue with it is that it does not have a toggling behavior (you can't use it hide the panel).

> kernel status as what is after pipe | like Disconnected, Inactive, etc is not shown anywhere else; p

It is there also less evident in the kernel status icon in the upper right corner of notebooks if you hover on it.

> notebook name is very useful when you have multiple notebooks open. I would highly prefer to keep it when in multi-panel mode. +1 for hiding it when in Simple mode

[...]

PR #16846, opened 2024-10-09
Hide the terminals part of the running sessions status bar item by default

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/11547

As discussed in https://github.com/jupyterlab/jupyterlab/issues/11547, this status bar item is probably not very useful in practice. And hiding it by default would leave more space for other items.

## Code changes

- [x] Add a setting to allow for hiding the "Running Sessions" status bar item
- [x] Hide the terminal part of the status bar item if there is no terminal running
- [x] Update UI tests snapshots

## User-facing changes

- The status bar item does not show the terminal icon and number if there is not terminal currently running
- The item can be hidden entirely from the status bar via the Settings Editor

https://github.com/user-attachments/assets/ae904b8f-c7e4-4b48-a9d6-35bf0fcb4d2a

None


SylvainCorlay:
Awesome!

jtpio:
Following @krassowski's comment in https://github.com/jupyterlab/jupyterlab/issues/11547#issuecomment-2464664801 :

> What about auto-hiding the terminals part when there are no running terminals? This would reduce the terminals + kernels item by half for many users.

This PR was updated to:

- auto hide the terminals part of the status item if there is not terminal currently running
- keep the setting, but enabled by default to not change the current behavior

And the UI tests snapshots updated accordingly.

Marking it as ready so it can be considered for 4.4.0.

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -139,39 +141,70 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
...
-    });
+    if (settingRegistry) {
+      let disposable: IDisposable;
+      const onSettingsUpdated = (
+        settings: ISettingRegistry.ISettings
+      ): void => {
+        const showStatusBarItem = settings.get('showStatusBarItem')
+          .composite as boolean;
+        if (showStatusBarItem) {
+          disposable = registerItem();
```
krassowski, line 196: Wouldn't it register the item multiple times once we add more settings? I think it needs a guard to check if the disposable is defined and not disposed

krassowski (commented):
Some minor improvement suggestions

packages/apputils/src/runningSessions.tsx
```diff
@@ -45,10 +45,14 @@ function RunningSessionsComponent(
...
-        <TextItem source={props.terminals} />
-        <terminalIcon.react verticalAlign="middle" stylesheet="statusBar" />
-      </GroupItem>
+      {props.terminals > 0 ? (
+        <GroupItem spacing={HALF_SPACING}>
+          <TextItem source={props.terminals} />
+          <terminalIcon.react verticalAlign="middle" stylesheet="statusBar" />
+        </GroupItem>
+      ) : (
+        <div></div>
```
krassowski, line 54: Can we return `null` or otherwise at least an empty fragment (`<></>`) here? See https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

```suggestion
        null
```

jtpio: Looks like the use of `GroupItem` prevents this:

```
Type 'Element | null' is not assignable to type 'Element | (Element & string)'.
  Type 'null' is not assignable to type 'Element | (Element & string)'.
```

packages/apputils-extension/src/statusbarplugin.ts
jtpio: It looks like that would indeed be the case. Maybe we can add indeed add such guard to anticipate the case where new settings would be added 👍

jtpio: Or maybe we can just dispose the item, and recreate it if `showStatusBarItem` is `true`.

packages/apputils/src/runningSessions.tsx
krassowski: Thanks for checking, I've opened https://github.com/jupyterlab/jupyterlab/issues/17212

krassowski (approved):
Thank you @jtpio!

jtpio:

Issue #12425, opened 2022-04-17
Add explanatory texts for settings in the visual settings editor


joelostblom:
### Problem

It is currently not possible to know what the effect of selecting some of the options in the visual settings editor. For example, I had no idea what "line wrap: bounded" did until I found issue https://github.com/jupyterlab/jupyterlab/pull/4936.

![image](https://user-images.githubusercontent.com/4560057/163723466-a3ec0535-f79f-484c-a2f3-bdaab060f8ec.png)

### Proposed Solution

Add explanatory text,either on hover for each option, or on the side of the dropdown for the selected option. In this case, what is in the PR comment would be great:

-    off: lines never wrap,
-    on: lines wrap at viewport border,
-    wordWrapColumn: lines wrap at column specified by wordWrapColumn parameter,
-    bounded: lines wrap at minimum between viewport width and wordWrapColumn parameter.

jweill-aws:
This is possible in our current schema, but it is not implemented for every option. To add this enhancement, we'd need to modify the schema to have better documentation.

![image](https://user-images.githubusercontent.com/93281816/164519529-43169712-6df5-492d-8725-e7b29fea4b46.png)

jweill-aws:
Affected code in here:

https://github.com/jupyterlab/jupyterlab/blob/7d6749ce4f467f8d374bb3930fcdadadee073fb5/packages/console-extension/schema/tracker.json#L131-L135

JasonWeill:
@kiansahafi Thank you for offering to help! It looks like fixing this issue will require creating an optional additional key to describe values such as `"off"` and `"on"` in `enum` settings, and it will require modifying the settings editor to use those descriptions in the UI. Please feel free to add comments or open a draft pull request (PR) if you have any questions. Once you're ready to propose a change, you can open a PR marked ready for review.

kiansahafi:
@JasonWeill  thanks for the helps, i have been trying to build the project but could not build it through the docker file because of [this issue](https://github.com/jupyterlab/jupyterlab/issues/14651), is there an alternative way to build the project or i should just wait until the issue is resolves?

krassowski:
@kiansahafi how exactly were you trying to use docker? That issue should have been closed long ago (I just did it now) because the scripts that were mentioned were rewritten, and the new docker image is tested regularly.

The up to date docker instructions are here: https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#setup-using-docker

[...]

PR #16848, opened 2024-10-09
Add more descriptive labels for fetching Jupyter news options

Adam-D-Lewis:
## References
Similar to https://github.com/jupyterlab/jupyterlab/issues/12425

## Code changes
adds more descriptive labels for fetching Jupyter news options

## User-facing changes
Users will see the new descriptive labels

None


jtpio (commented):
Thanks!

This does seem to be improving the understanding of these options for those using JupyterLab in English:

image

When using a different display language (for example French), it looks like some of the options are already translated, which is nice:

image

Except for `Ask (None)`. Maybe it should just be `Ask`?

Also the description mentions `true`, which is not surfaced with the Settings editor UI. So wondering if the description should now also mention these new `title` values, for example:

*Whether to fetch news from Jupyter news feed. If Always (`true`), it will make a request to a website.*

PR #16749, opened 2024-09-05
Change "File Size" Column name to "Size" in file browser

ChairBorn:
## References

This PR fixes #16734 and changes the Name of the File Browser "File Size" column to "Size".

## Code changes

Line 2811 of [packages/filebrowser/src/listing.ts] changed to "Size"

## User-facing changes

Now users will see the column name as "Size"

Screenshot 2024-09-05 110521

There are other references to "File Size", for instance: 

line 1302 of [packages/filebrowser-extension/src/index.ts] - overlay Menu in file browser
line 947 /test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
line 70 /test/jupyterlab/sidebars.test.ts

Let me know if you want me to change the above mentioned instances as well.

PR #16849, opened 2024-10-09
Change "File Size" column name to "Size" at small widths

### LINKED ISSUE #16734: Rename "File Size" column in file browser to "Size"
## Description

It is possible to read file sizes when the file size column in the filebrowser is very narrow, so much so that the word "Size" isn't visible in the column header:

image

Shorten this column name to "Size" — there is no ambiguity here. 

For comparison, "Size" is used as a column header in macOS's Finder:

image

## Context

- Operating System and version: macOS Sonoma 14.6.1
- Browser and version: Firefox ESR 129.0.1
- JupyterLab version: 4.3.0beta1

JasonWeill:
This is a copy of #16749 by @ChairBorn, with the latter two commits backed out and rebased to upstream `main`. Creating to run the test pipelines independently.

Fixes #16734.
Closes #16749

### DIFF
--- a/packages/filebrowser/style/base.css
+++ b/packages/filebrowser/style/base.css
@@ -198,7 +198,7 @@
  * to display either a small or large header for the last-modified column.
  */
 /* stylelint-disable */
-@container (max-width: 100px) {
+@container (max-width: 300px) {
   /* stylelint-enable */
   .jp-DirListing-headerItem.jp-id-modified
     > .jp-DirListing-headerItemText-small,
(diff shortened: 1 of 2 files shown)


jtpio:
Looks like the UI tests failures are relevant, and related to the CSS changes in this PR?

[filebrowser-diff.webm](https://github.com/user-attachments/assets/80a6bdb6-a2ce-454f-b492-ec7d38cc4da1)

ChairBorn:
on it, i will double check the tests results before bothering you next time

packages/filebrowser/style/base.css
```diff
@@ -196,42 +197,38 @@
...
   }
 }
 
-.jp-DirListing-headerItem.jp-id-filesize {
-  text-align: right;
-}
-
-.jp-DirListing-headerItem.jp-mod-selected {
-  font-weight: 600;
-}
```
krassowski, line 233: This should be preserved

jtpio:
Hmm, is this change caught by the examples UI tests intended?

| Actual | Expected | 
| --------- | --------------|
| example-actual | example-expected |

packages/filebrowser/style/base.css
```diff
@@ -196,38 +197,38 @@
  * Use container queries (not yet supported by our version of stylelint)
  * to display either a small or large header for the last-modified column.
  */
-/* stylelint-disable */
```
jtpio, line 199: Looks like we also need to keep this comment:

image

jtpio:
Testing this PR in a clean environment, it looks like there is a glitch where both "Modified" and "Last Modified" appear to be displayed at a specific column width:

[jupyterlab-modified-column-issue.webm](https://github.com/user-attachments/assets/989175b0-da53-4d52-b86b-3482e9b2a4b6)

Which does not seem to be the case with `4.3.0b3`.

jtpio (approved):
Thanks @ChairBorn and @JasonWeill!

I pushed a couple more fixes and the UI tests are now passing.

Issue #15441, opened 2023-11-25
 Providers can break completions when `isApplicable()` raises


krassowski:
## Description

If completion provider's `isApplicable()` rejects or never resolves due to an error in provider code, `await Promise.all(isApplicablePromises);` never resolves and completer silently. This lead to https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1006

## Reproduce

https://github.com/jupyterlab/jupyterlab/blob/3d525a2c24524fe1431c3227e62057c9137940ed/packages/completer/src/reconciliator.ts#L42-L48

## Expected behavior

- `Promise.allSettled` and maybe timeout (e.g. via `Promise.race`) are used
- error is exposed in console

## Context

JupyterLab 4.0.9

alewilliam789:
I replaced the section with Promise.allSettled() and console.logged any promises that rejected. The PR number is #16014

Kunalpal216:
Is this issue open ? I would like to work on this.

hrishi-1337:
@krassowski Please assign this issue to me and take a look at my PR. Thanks!

PR #16855, opened 2024-10-11
Added timeout handling

hrishi-1337:
## References

#15441 

## Code changes

Added a 5-second timeout mechanism using race to mark providers as rejected and added a .catch() block to log errors. Used allSettled to process all promises, and filter applicable providers based on their fulfillment status.

## User-facing changes

N/A

N/A


hrishi-1337:
1. Had to remove `allowSeparatedGroups: false` to clear eslint. The key is set to false by default in eslint-sort-imports so it shouldn't impact linting.
2. No changes were made to `.pre-commit-config.yaml`, wasn't able to fix the diff to reflect this.

hrishi-1337:
Hi @krassowski, do you know a way to add the check annotations to prevent this unrelated failure? I see these happening in other PRs as well, but I can't seem to find a way to fix it.

PR #16863, opened 2024-10-15
Fix triggering completer on the beginning of the lines

### LINKED ISSUE #16847: Code completer (inline/auto) only triggers if a character is already present in the line
## Description

Code completer (inline/auto) only triggers if a character is already present in the line

## Reproduce

1. Enable auto-completion 
Screenshot from 2024-10-09 10-28-59
2. Follow the screencast:
completer-only-triggers-on-second-char

and

1. Enable inline-completion: 
Screenshot from 2024-10-09 10-28-49
2. Follow the screencast:
inline-completer-does-not-triggers-on-empty-line

## Expected behavior

- Auto-completion triggers after first character
- Inline completion can be triggered on empty line

## Context

- Operating System and version: Ubuntu
- Browser and version: Chrome 129
- JupyterLab version: 4.3.0b3

andrewfulton9:
## References

Closes #16847

## Code changes

Removes logic in completer that removes `COMPLETER_ENABLED_CLASS` if `!line`. Also updates logic to check if line exists before checking line slice logic. These fix the issue by making sure that the cell has the appropriate class to allow the completer enable command ID to function.

## User-facing changes

User will be able to activate the completer on an empty line. Functionally, a user should not notice any changes unless they are using jupyter-ai since the history completer will return an empty list at this point.

None that I am aware


krassowski (changes requested):
So far this looks _mostly_ good: increasing/decreasing indent works, Shift + Tab for inspector tooltip works, completion works, autocompletion now works reliably, inline completion works.

#### Problem

One notable change in the behaviour is that it is now possible to invoke completer using <kbd>Tab</kbd> when at the begging of the line; I think this may lead to bad UX if users are accustomed to using <kbd>Tab</kbd> for indentation. Say we have:

image

and I want to press <kbd>Tab</kbd> to increase indent and type `return i`.

| Previously I would get | but with this PR as of today I get |
|--|--|
| image | image |

I think this is a blocker as this would disrupt the UX of existing users significantly.

#### Proposed solution

I think we should change the selector for <kbd>Tab</kbd> shortcut so that it does not get active at the beginning of the lines. There are three shortcuts (for the console, notebook and file editor), here is one:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/packages/notebook-extension/schema/completer.json#L4-L10

We could follow a similar idea as in:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/packages/tooltip-extension/schema/notebooks.json#L10-L14

which is implemented in:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/packages/codeeditor/src/widget.ts#L145-L155

but instead of `.jp-mod-in-leading-whitespace` we would add a new `.jp-mod-at-line-beginning` (or similar). This would need to be documented in:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/docs/source/developer/css.rst?plain=1#L236-L244

#### Other requests

We should add a test for the issue being addressed. I think an integration test would be right here, basically something like:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/galata/test/jupyterlab/inline-completer.test.ts#L145-L152

but after removing contents of the cell. For that we will need to make the history provider return _something_ when the prefix is empty. Currently the logic short-circuits if the prefix is empty:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/packages/completer/src/default/inlinehistoryprovider.ts#L73-L90

As @andrewfulton9 mentioned to me this could be the most frequent item in recent history. In that case instead of sending `hist_access_type` of "search" type we could use "tail" and manually count items which appear most commonly.

krassowski:
@andrewfulton9 do you need help finishing up this one?

andrewfulton9:
> @andrewfulton9 do you need help finishing up this one?

@krassowski, yeah I do. I think I've encountered a race condition when I implimeneted the above suggestions, that causes the ghost text to not clear when I hold down the backspace key. Below is a video showing it happening:

https://github.com/user-attachments/assets/0bf45820-0710-4ce6-9221-06b0d5289a08


The branch to reproduce it is [here](https://github.com/andrewfulton9/jupyterlab/tree/inline_completer_bud_debug_snapshot_update). I think it has something to do with the code [here](https://github.com/andrewfulton9/jupyterlab/blob/360590a0102ecc114fdbca1dd27ee7eae6372679/packages/completer/src/handler.ts#L485-L528), but its hard to tell because it only occurs when the code is run asynchronously. I haven't been able to reproduce it when using the browser debugger.

krassowski:
Yes, so when you press <kbd>Backspace</kbd> on `print` it fetches a completion `t("hello")` say with ID 1:

https://github.com/jupyterlab/jupyterlab/blob/360590a0102ecc114fdbca1dd27ee7eae6372679/packages/completer/src/handler.ts#L485

When you quickly press Backspace again new request is sent and internal counter increases to ID 2; the internal counter is checked against the replies to avoid out of sync replies showing up:

https://github.com/jupyterlab/jupyterlab/blob/360590a0102ecc114fdbca1dd27ee7eae6372679/packages/completer/src/handler.ts#L498-L500

The problem is that when removing the last letter, say `p` a special case earlier on kicks in:

https://github.com/jupyterlab/jupyterlab/blob/360590a0102ecc114fdbca1dd27ee7eae6372679/packages/completer/src/handler.ts#L467-L475

and a new request is never sent for automatic completion hence the request counter does not increase and the invalidation logic does not trigger when a late replay comes in.

I think this can be fixed with increasing the counter (`this._fetchingInline`) before short-circuiting in line 474.

packages/completer/src/handler.ts
```diff
@@ -306,12 +300,16 @@ export class CompletionHandler implements IDisposable {
       return;
     }
 
-    // If the part of the line before the cursor is white space, return.
-    if (line.slice(0, position.column).match(/^\s*$/)) {
-      this._enabled = false;
-      model.reset(true);
-      host.classList.remove(COMPLETER_ENABLED_CLASS);
-      return;
+    // If no line or cursor has not characters before
```
krassowski, line 303: Highlighting grammar issue here

krassowski:
It looks like the last remaining thing here is adding a test to ensure that the inline completion can in fact be triggered (manually) at the beginning of a line?

packages/completer/src/handler.ts
andrewfulton9: I updated the language to be more clear here.

krassowski:
The `js-metapackage` test failure is related:

```
 FAIL  test/completer/handler.spec.ts (6.277 s)
  ● @jupyterlab/completer › CompletionHandler › #autoCompletion › should use TriggerCharacter for typed text

    expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: 2

      389 |       it('should use TriggerCharacter for typed text', async () => {
      390 |         // this test depends on the previous one ('should use Invoked for invoke()').
    > 391 |         expect(provider.triggers.length).toEqual(1);
          |                                          ^
      392 |
      393 |         handler.editor!.model.sharedModel.updateSource(4, 4, 'a');
      394 |         await new Promise(process.nextTick);

      at Object.<anonymous> (test/completer/handler.spec.ts:391:42)
```

I think this is because there is a trigger on empty line now, so the test will need to be adjusted and a comment explaining it added.

andrewfulton9:
@krassowski, in terms of the UI regression tests, I think one of the failures needs to have its snapshots updated to account for the completer shortcuts now being valid on and empty line, but there is also a failure for the debugger breaking on exceptions that does not appear to be related to this PR. Should i run update snapshots here, or try to only do it for the completer documentation locally?

krassowski:
> but there is also a failure for the debugger breaking on exceptions that does not appear to be related to this PR

This one is unfortunately a bit flaky. I will try to fix it.

> Should i run update snapshots here, or try to only do it for the completer documentation locally?

I updated the snapshot now, here is what I did for reference:
- downloaded `jupyterlab-documentation-test-assets.zip`
- compared `shortcuts-help-expected.png` with `shortcuts-help-actual.png` to understand the change
- replaced `shortcuts-help-documentation-linux.png` with `shortcuts-help-expected.png`
- run `shortcuts-help-expected.png` though trimage compressor (asking bot to update snapshots will compress the assets automatically)

When it's just one snapshot I tend to do it manually. When it's multiple snapshots I tend to ask the bot and then revert false positives/flaky updates using `git checkout`.

krassowski (approved):
Thank you @andrewfulton9!

krassowski:

postwave:
@krassowski  Hello Mike, What's the motive behind preventing completer's auto-trigger at the beginning of the line?
https://github.com/jupyterlab/jupyterlab/blob/360590a0102ecc114fdbca1dd27ee7eae6372679/packages/completer/src/handler.ts#L467-L475
A typical scenario for jupyter-ai completer is typing "# write me a quick-sort function",  press enter and  wait for ai completion in next blank line. 
For vscode copilot, completion would trigger in such situation (regardless if it's backspace or enter causing an empty line)

PR #16866, opened 2024-10-15
Fix bracket highlighting for Python string literal concatenation

### LINKED ISSUE #16865: Bracket highlighting not working for multiple-line Python string literal concatenation
## Description
When you have strings on multiple lines with 1 f-string not on the last line, bracket highlighting doesn't work although it's valid syntax. 
It only works properly if the f-string is on the last line.

https://github.com/user-attachments/assets/2e24db2f-5637-4af5-bcb9-38dfb243de58

I reported this issue upstream at https://github.com/codemirror/dev/issues/1454 and found upgrading @lezer/python dependency resolved this.

## Reproduce
1. Type any pair of brackets `()` `[]` `{}`
2. Type strings on multiple lines with 1 f-string not on the last line

## Expected behavior
Bracket highlighting works properly

## Context
- Browser and version: JupyterLab Desktop 4.2.5-1 (latest)
- JupyterLab version: 4.2.5 (latest)

Nagidrop:
## References

Fixes #16865 

## Code changes

Update `@lezer/python` and `@lezer/highlight` resolutions to latest versions.

## User-facing changes

Bracket highlighting now works properly for Python string literal concatenation when there are f-strings.

Before | After |
--- | --- |
before |  after |

None


yarn.lock
```diff
@@ -5124,7 +5124,7 @@ __metadata:
 
 "@lezer/highlight@npm:^1.0.0, @lezer/highlight@npm:^1.1.3, @lezer/highlight@npm:^1.2.0":
   version: 1.2.0
-  resolution: "@lezer/highlight@npm:1.2.0"
+  resolution: "@lezer/highlight@npm:1.2.1"
   dependencies:
     "@lezer/common": ^1.0.0
   checksum: 5b9dfe741f95db13f6124cb9556a43011cb8041ecf490be98d44a86b04d926a66e912bcd3a766f6a3d79e064410f1a2f60ab240b50b645a12c56987bf4870086
```
krassowski, line 5130: We also need to update the checksums.

Nagidrop: Is it done by using `yarn install --update-checksums`? I currently don't have `yarn` locally, so I need to install it and perform the checksum update right?

krassowski: JupyterLab comes with a bundled version of `yarn` under `jlpm` alias. The syntax you mention is for `yarn` v1 so it would not work here (`jlpm` points to yarn v3). Instead you can run:

```
YARN_CHECKSUM_BEHAVIOR=update jlpm
```

from an environment where jupyterlab is installed (and commit the changes).

Nagidrop: I was struggling to run the command, but later found that `set YARN_CHECKSUM_BEHAVIOR=update` then `jlpm` works on Windows. I have pushed a new commit. Should I also squash the 2 commits into 1?

krassowski: No, it's fine - we use squash merging.

krassowski:
I tested this change on binder and it indeed fixes the problem. There is a noticeable change in how `\n` are now highlighted differently. I think this is fine. I will ask the bot to update snapshots:

bot please update galata snapshots (it will commit back to this branch if it succeeds)

One snapshots in `examples` (for `notebook`) will need to be updated manually. You can get the update snapshot by downloading the `examples 20374` artifact from https://github.com/jupyterlab/jupyterlab/actions/runs/11347673017?pr=16866 and unpacking it.

Nagidrop:
> I tested this change on binder and it indeed fixes the problem. There is a noticeable change in how `\n` are now highlighted differently. I think this is fine. I will ask the bot to update snapshots:
> 
> bot please update galata snapshots (it will commit back to this branch if it succeeds)
> 
> One snapshots in `examples` (for `notebook`) will need to be updated manually. You can get the update snapshot by downloading the `examples 20374` artifact from https://github.com/jupyterlab/jupyterlab/actions/runs/11347673017?pr=16866 and unpacking it.

Coincidentally, I also checked it on Binder when you replied and updated the User-facing Changes section accordingly. 

Just to make sure, I just need to replace the `examples/notebook/example.spec.ts-snapshots/example-linux.png` using `example-actual.png` right?

krassowski:
Thank you! It looks like the bot did not update the remaining tests. I am working on a PR to fix it.

krassowski:

Nagidrop:
> Thank you! It looks like the bot did not update the remaining snapshots. I am working on a PR to fix it.

Thank you for your kind assistance along the way :D All checks seem good now. Do I need to do anything else?

krassowski:
I will just restart the CI to ensure it passes by closing and re-opening (the bot commits do not trigger it correctly) :)

Nagidrop:
Some tests didn't pass again. They seem to be from the CI process but if they are caused by my PR, please tell me and I'll fix it :D

krassowski (approved):
This looks good to me, thank you @Nagidrop!

I will propose to include this in 4.3.0 milestone as this is a minor change in dependencies and fixes a user-facing bug in the editor.

jtpio:
> I will propose to include this in 4.3.0 milestone as this is a minor change in dependencies and fixes a user-facing bug in the editor.

Sounds good :+1:

Issue #14623, opened 2023-05-31
Better handling of the file browser configuration


jtpio:
## Description

Currently the file browser sets its own default config here:

https://github.com/jupyterlab/jupyterlab/blob/d98ef15bfa916309c82d105a5d442d915d5dd1df/packages/filebrowser-extension/src/index.ts#L219-L247

Since this happens at application startup, it can override other configuration set in different plugins.

## Reproduce

In a separate plugin (for example in a custom third-party extension), use `browser.showFileCheckboxes = true` to show the file browser checkboxes by default on application startup.

However because of the current logic with the code snippet above the default configuration will be applied again when the plugin activates.

In the following screencast we see how the checkboxes first appear and then quickly disappear:

https://github.com/jupyterlab/jupyterlab/assets/591645/0c7dc958-2fba-4441-81ef-86db77f26002

## Expected behavior

Other plugins should be allowed to use public setters like `browser.showFileCheckboxes = true` to customize the behavior of the file browser, when there is no user setting available.

If there is a value in the settings, then it takes precedence.

## Context

This is an issue in Notebook 7 which has to use the settings to store its default values for the file browser (checkboxes, file size column, sort notebooks first). For example: https://github.com/jupyter/notebook/pull/6892

However since these settings are also shared with JupyterLab, they also apply in the JupyterLab interface.

The idea would be for Notebook 7 to be able to customize with the following without having the configuration being overridden by the default lab plugin:

```ts
browser.showFileCheckboxes = true
browser.sortNotebookFirst = true
browser.showFileSizeColumn = true
```

PR #16870, opened 2024-10-15
Fix the handling of the file browser configuration

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14623

## Code changes

Move the handling of the file browser settings to a separate plugin, so it's easier to change the settings and not have them collide in downstream applications such as Notebook 7.

## User-facing changes

None for JupyterLab users.

For Notebook users, using https://github.com/jupyter/notebook/pull/7481, the file browser will show different defaults than JupyterLab (checkboxes, file size column, single click navigation), without having to modify the settings and have these defaults be reflected in JupyterLab too:

[filebrowser-settings-notebook-lab.webm](https://github.com/user-attachments/assets/001df8eb-7c08-404d-bbee-22d414b0e34d)

None


jtpio:
Adding to `4.3.0` as this is actually a bug fix (and hoping to finish the PR before the final release).

jtpio:
hmm the UI test failures seem to be consistent but are not expected, need to check what is happening there:

example-diff

jtpio:
On a local install from `main`, loading JupyterLab would set the following size for the file browser columns:

image

Switching to this branch:

image

So moving the logic to a separate plugin seems to introduce a small delay that has some effect on the final columns widths, which also seems to be reflected in the UI tests.

Maybe it's then still fine to update the reference snapshots, as the differences are quite minor in the end.

jtpio:
bot please update snapshots

jtpio:
bot please update snapshots

jtpio:
Looks like CI is now passing :heavy_check_mark: 

If we are fine with the minor updates to the snapshots, this should be ready to go.

packages/filebrowser-extension/src/index.ts
```diff
@@ -160,7 +160,7 @@ const namespace = 'filebrowser';
  */
 const browser: JupyterFrontEndPlugin<IFileBrowserCommands> = {
   id: FILE_BROWSER_PLUGIN_ID,
-  description: 'Set up the default file browser (commands, settings,...).',
+  description: 'Set up the default file browser commands',
```
krassowski, line 163:
```suggestion
  description: 'Set up the default file browser commands and state restoration',
```

krassowski: Since it also handles the `restorer` and `preferredPath`.

PR #16871, opened 2024-10-16
Upgrade markdown-to-jsx dependency

### LINKED ISSUE #16864: CVE-2024-21535 in markdown-to-jsx < 7.4.0
It seems that there is a vulnerable version of markdown-to-jsx in jupyterlab:

https://github.com/jupyterlab/jupyterlab/blob/c321c27b50431fc0414b20a3b61375af7ccb6c2e/jupyterlab/staging/yarn.lock#L7407

This fix is easy but it'd be better to update to the latest version IMHO: https://github.com/quantizor/markdown-to-jsx/commit/8eb74da825c0d8d2e9508d73c672bcae36ba555a

JasonWeill:
## References

Fixes #16864.

## Code changes

Upgrades `rjsf` packages to version `~5.16.1`, the first version that requires at least `markdown-to-jsx` version 7.4.0, to avoid a dependency on an older version of `markdown-to-jsx` that has a known vulnerability.

## User-facing changes

None.

None.

### DIFF
--- a/packages/celltags-extension/package.json
+++ b/packages/celltags-extension/package.json
@@ -45,7 +45,7 @@
     "@jupyterlab/translation": "^4.3.0-beta.3",
     "@jupyterlab/ui-components": "^4.3.0-beta.3",
     "@lumino/algorithm": "^2.0.2",
-    "@rjsf/utils": "^5.13.4",
+    "@rjsf/utils": "~5.16.1",
     "react": "^18.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


jtpio:
Thanks @JasonWeill.

It looks like the update was handled by `dependabot` here: https://github.com/jupyterlab/jupyterlab/pull/16868. And its CI is passing, so we can likely go ahead and its PR.

If you think it's still worth updating `rjsf` then we can also do that here and try to have it finished for `4.3.0`.

JasonWeill:
Closing in favor of #16868.

PR #16878, opened 2024-10-18
Ignore `https://code.visualstudio.com` in `check-links`

jtpio:
## References

This might be temporary until we resolve issue https://github.com/jupyterlab/jupyterlab/issues/16837.

This check has been failing consistently for the past few days:

image

Not sure if it's because https://code.visualstudio.com detects it's being accessed in automated manner (403)?

## Code changes

Add `https://code.visualstudio.com` to the ignore list `check-links`.

## User-facing changes

None

None


krassowski:
Edited top-level comment so that it does not auto-close https://github.com/jupyterlab/jupyterlab/issues/16837.

Issue #16867, opened 2024-10-15
Clicking the status circle should go to the current executing cell


stuaxo:
When I'm quickly scrolling up and down it's sometimes hard to tell where the current cell is.

The status circle shows how far execution has gone, it would be nice if clicking it went to the currently executing cell.

![Image](https://github.com/user-attachments/assets/3db6f03d-5ebb-464d-810f-3637f70e7631)

Darshan808:
Working on this!

Issue #8714, opened 2020-07-24
Scroll to exception when notebook is running in "restart and re-run" mode


krassowski:
### Problem

* I'm always frustrated when the execution of my notebook stops early in "restart and re-run" mode because the GUI does not give me any indication of this - I often switch to another task "B" when my notebooks execute so an unnoticed early termination is a waste of time:
  - I usually check back after the amount of time I expect the notebook to finish
  - if I do not know how long a notebook will execute, I look at the kernel icon (if it is a full black circle or not)
    - having two-monitor setup it is easy to do without detreacting my attention from the task B
    - if I see the white empty circle I note that I can switch back once I finish task B
       - it's a huge disappointment if I go back and see that the notebook only executed in 3/4 - this means that I will have to wait 3/4 + 1/4 time again;
       - if I knew that there was an error I would have switched earlier (before finishing task B) and would recover some of this time!
* I would like it if an indication of the error was shown in the GUI because this would allow me to quickly get back, fix a typo and re-run again.

### Proposed Solution

I would like the GUI to indicate an early termination of the execution flow:
- It can be done via automated scroll to the exception/last executed cell - this is where (most likely) the error in the code would be.
  - however, the automated scroll could disrupt the user if they were just editing a cell _while the notebook was executing_
- It could be a flashing red button on the toolbar (e.g. close to the re-run button), which when pressed would scroll down to the last executed cell (so the cell with the exception!)
  - if there was no exception, it would be useful anyway to allow the user to scroll to the most-recently executed cells; this would be the missing piece for #6818 (as mentioned by @ian-r-rose it would be useful to expose the scroll-to-cell function)
  - if it were to be flashing it could have to have a timeout after which it stops flashing
  - it should stop flashing after any other cell in this notebook gets executed

### Additional context

As a workaround, I have implemented this earlier in my [`jupyter-helpers`](https://github.com/krassowski/jupyter-helpers) for IPyhon kernel back two years ago. I used automated scrolling but because of the disruption to the user workflow, now I think that having a flashing button would be a better solution. I think that this might be of interest for almost every user and that it could belong to the main JupyterLab project - but obviously I am happy to do a test run creating a third-party extension first.

aiqc:
pasting duplicate comment here.
---
### Problem
When running a long notebook, it is a pain to figure out which cell failed. You have to scroll around until you found it. Especially if you used the "run all above" from the bottom of the notebook, or "run all cells" when viewing the top of the notebook.

---

[...]

PR #16880, opened 2024-10-18
Add feature to scroll to currently executing cell on progress circle click

Darshan808:
### Description
This PR introduces a feature where clicking the progress circle in the status bar will automatically scroll to the currently executing cell if present. This feature improves usability when users are working on large notebook, making it easier to locate the cell that is currently running.

### Changes
- Added a click handler on the progress circle that checks if a cell is in the "running" state.
- If a running cell is detected, the notebook will scroll to that cell using the `scrollToCell` method.

### Additional Notes:
- This behavior only works when the notebook is in the "busy" execution state i.e when some cell is being executed. 

Please review and let me know if further improvements are needed or a test needs to be added.

### DIFF
--- a/packages/notebook/src/executionindicator.tsx
+++ b/packages/notebook/src/executionindicator.tsx
@@ -26,6 +26,7 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { Widget } from '@lumino/widgets';
 import { JSONObject } from '@lumino/coreutils';
 import { IChangedArgs } from '@jupyterlab/coreutils';
+import { isCodeCellModel } from '@jupyterlab/cells';
 
 /**
  * A react functional component for rendering execution indicator.
@@ -69,6 +70,18 @@ export function ExecutionIndicatorComponent(
       width={16}
       height={24}
       label={trans.__('Kernel status')}
+      onClick={async () => {
+        if (props.nb && state.executionStatus === 'busy') {
+          for (let i = 0; i < props.nb.widgets.length; i++) {
+            const cell = props.nb.widgets[i];
+            if (isCodeCellModel(cell.model) && cell.model.executionState === 'running') {
+              await props.nb.scrollToCell(cell);
+              break;
+            }
+          }
+        }
+      }
+    }     
     />
   );
   const titleFactory = (translatedStatus: string) =>
@@ -178,6 +191,10 @@ namespace ExecutionIndicatorComponent {
      * The application language translator.
      */
     translator?: ITranslator;
+    /**
+     * The current notebook.
+     */
+    nb?: Notebook;
   }
 }
 
@@ -209,15 +226,16 @@ export class ExecutionIndicator extends VDomRenderer<ExecutionIndicator.Model> {
             displayOption={this.model.displayOption}
             state={undefined}
             translator={this.translator}
+            nb={undefined}
           />
         );
       }
-
       return (
         <ExecutionIndicatorComponent
           displayOption={this.model.displayOption}
           state={this.model.executionState(nb)}
           translator={this.translator}
+          nb={nb}
         />
       );
     }
--- a/packages/statusbar/src/components/progressCircle.tsx
+++ b/packages/statusbar/src/components/progressCircle.tsx
@@ -25,6 +25,10 @@ export namespace ProgressCircle {
      * Element height
      */
     height?: number;
+    /**
+     * Click handler
+     */
+    onClick?: () => void;
   }
 }
 
@@ -54,6 +58,7 @@ export function ProgressCircle(props: ProgressCircle.IProps): JSX.Element {
       aria-valuemin={0}
       aria-valuemax={100}
       aria-valuenow={props.progress}
+      onClick={props.onClick}
     >
       <svg viewBox="0 0 250 250">
         <circle


packages/notebook/src/executionindicator.tsx
```diff
@@ -178,6 +193,10 @@ namespace ExecutionIndicatorComponent {
      * The application language translator.
      */
     translator?: ITranslator;
+    /**
+     * The current notebook.
+     */
+    nb?: Notebook;
```
krassowski, line 199:
```suggestion
    notebook?: Notebook;
```

Let's avoid unnecessary abbreviations in public interfaces :)

krassowski (commented):
UX thoughts:
- how will user know that clicking here will cause scroll action? The label just says "kernel status"
- how can we add more actions in the future? If we make click on the indicator one thing we cannot make it another thing (except for alt + click etc combinations, but these won't work on touch screens)

I would like to explore if we can instead put a button in the tooltip with execution details which shows up on hover. This could solve both points above (if it has a "jump" icon it would be more discoverable, and in the future we could have more buttons). In that case the tooltip shown on hover might need to be more persistent.

I am not 100% sure if what I am proposing is the ideal solution but I believe it is worth exploring/waiting for some more feedback from other folks.

Darshan808:
@krassowski 
Thank you for the feedback! I completely agree that making the feature more user-friendly and flexible for future enhancements. My initial implementation was based on the description in issue #16867 but I'm more than happy to improve it further.

Would you suggest that I start exploring the idea of adding a "jump" button on hover, or would it be better to pause on this feature for now and gather more feedback before proceeding?

JasonWeill:
I just tried this code, and found an error:

1. Create a notebook with 100 cells in it
2. Scroll way down to the bottom, and write code (see below) to sleep for 100 seconds. Run it.
3. Scroll up to the top, write a cell that says `print("foo")`, run it. This cell is queued.
4. Scroll to where neither running cell is visible.
5. Click the progress circle.
6. Expected: the 100-second sleep cell is focused. Actual: the `print` cell is focused.


```python
import time

time.sleep(100)
```

jtpio:
Nice, thanks for working on this!

Wondering if this feature could then be expanded (not necessarily in this PR) to also address:

- https://github.com/jupyter/notebook/issues/4323
- https://github.com/jupyterlab/jupyterlab/issues/8714

Darshan808:
Just got back to working on this PR after a while. Made some updates (and edited the PR description above). Would love feedback!

packages/notebook/src/executionindicator.tsx
```diff
@@ -620,7 +665,22 @@ export namespace ExecutionIndicator {
       loadSettings
         .then(settings => {
           const updateSettings = (newSettings: ISettingRegistry.ISettings) => {
-            toolbarItem.model.updateRenderOption(getSettingValue(newSettings));
+            const displayOptions = getSettingValue(newSettings);
+
+            // Auto-enable recordTiming if jump button is enabled
+            if (displayOptions.showJumpToCell) {
+              const recordTiming = newSettings.get('recordTiming')
```
krassowski, line 672: Why is this needed?

krassowski: This is why:

https://github.com/jupyterlab/jupyterlab/blob/889d09360e8840c14900d640408cf76a0bf85893/packages/notebook/src/actions.tsx#L1883-L1892

packages/notebook-extension/schema/tracker.json
```diff
@@ -671,6 +671,12 @@
           "type": "boolean",
           "title": "Show execution progress.",
           "default": true
+        },
+        "showJumpToRecentExecutionButton": {
+          "type": "boolean",
+          "title": "Show jump to most recently executed/executing cell button",
+          "description": "If enabled, displays a button that scrolls to the currently executing cell or, if none, the most recently executed cell.",
```
krassowski:
```suggestion
          "description": "If enabled, displays a button that scrolls to the currently executing cell or, if none, the most recently executed cell. The button shows when hovering over the kernel execution indicator. Enabling this option automatically enables recording of execution timing metadata.",
```

packages/notebook/src/executionindicator.tsx
```diff
@@ -87,10 +112,18 @@ export function ExecutionIndicatorComponent(
       >
         <span> {titleFactory(kernelStatuses[status])} </span>
         {popup}
+        {showJumpToCell && hasExecutedCell && (
+          <Button
+            className="jp-Notebook-ExecutionIndicator-jump-btn"
```
krassowski:
```suggestion
            className="jp-Notebook-ExecutionIndicator-jumpButton"
```

galata/test/jupyterlab/notebook-scroll.test.ts
```diff
@@ -415,3 +415,66 @@ test.describe('Notebook scroll beyond a cell with long output (with windowing)',
...
+  test('should show jump button after first execution and scroll to executing cells', async ({
+    page
+  }) => {
+    await page.notebook.setCell(0, 'code', 'from time import sleep\nsleep(2)');
+
+    // Button doesn't exist before execution
+    const indicator = page.locator('.jp-Notebook-ExecutionIndicator');
+    await indicator.hover();
+    await expect(
+      page.locator('.jp-Notebook-ExecutionIndicator-jump-btn')
```
krassowski:
```suggestion
      page.locator('.jp-Notebook-ExecutionIndicator-jumpButton')
```

galata/test/jupyterlab/notebook-scroll.test.ts
```diff
@@ -415,3 +415,66 @@ test.describe('Notebook scroll beyond a cell with long output (with windowing)',
...
+
+    // Run first cell
+    void page.notebook.runCell(0, false);
+
+    await page.notebook.addCell('code', '1');
+    const runPromise = page.notebook.runCell(3, false);
+
+    // Hover and verify button exists
+    await indicator.hover();
+    const jumpButton = page.locator('.jp-Notebook-ExecutionIndicator-jump-btn');
```
krassowski:
```suggestion
    const jumpButton = page.locator('.jp-Notebook-ExecutionIndicator-jumpButton');
```

Darshan808:
> I am also not sure if we should use "jump" or "scroll" in UI.

<strike>I think "scroll" makes more sense here!</strike>
But the scroll is so fast that it’s barely visible, so I guess using "jump" is fine.

packages/notebook/style/executionindicator.css
```diff
@@ -63,3 +66,17 @@
...
 }
+
+.jp-Notebook-ExecutionIndicator-jumpButton {
+  margin-top: 4px;
+  margin-bottom: 3px;
+  width: 100%;
+  font-size: var(--jp-ui-font-size1);
+  border: 1px solid var(--jp-border-color2);
+  color: var(--jp-ui-font-color0);
+  transition: all 0.1s ease;
```
Yahiewi, line 77:
```suggestion
```

This causes a slight delay in the disappearance of "Jump to cell" when the indicator loses hover:

https://github.com/user-attachments/assets/dbbcdcc1-62cd-41ab-84a6-f4066b04827e

Without transition:

https://github.com/user-attachments/assets/11917127-ecc9-4e79-aab8-2cb371a21085

PR #16881, opened 2024-10-19
Do not skip cells with hidden input when navigating with "k" "j" "<Up>" and "<Down>"

### LINKED ISSUE #11162: Configuration option: Don't skip collapsed cells when selecting
### Problem

#6356 changes the default jupyterlab behavior to skip collapsed cells when scrolling with the arrows or `j` + `k`. Would it be possible to add a configuration option to switch back to the old behavior?

I personally dislike this new behavior. I often know which code is in a collapsed cell and need to re-execute it, but prefer to keep it collapsed because it is unusually long. For example: I have a >100 line function definition, and subsequent cells only ~5-10 lines long that call the function with different arguments.

Previously, to re-execute a collapsed cell block, I would simply navigate to it with the arrows or `j` + `k`. Now, I have to select the collapsed cell with the mouse pointer, or `Shift + Enter` through adjacent cells until I reach the cell I want to execute (which is not always practical/possible).

### Proposed Solution

Add a configuration option to include collapsed cells when scrolling with the arrows or `j` + `k`.

lukelbd:
## References

Issue #11162

## Description

This fixes the issue where cells with hidden (i.e. "collapsed") code are not accessible from the cell navigation keyboard shortcuts "k" "j" "\<Up>" and "\<Down>".

This issue makes running or navigating to collapsed cells impossible without either 1) clicking the cell (which also automatically expands the cell) or 2) using the shift + enter shortcut (which requires running preceding cells). Sometimes the user may already know the contents of a collapsed cell by its output or any preceding markdown headers, but may wish to keep the cell collapsed for the sake of readability or to prevent large multi-line cells from dominating the screen.

## Code changes

This removes the `!possibleNextCell.inputHidden` condition used by the `selectAbove()` and `selectBelow()` functions to automatically skip cells with hidden input code.

## User-facing changes

Cells with hidden (i.e. "collapsed") input code are no longer skipped when navigating with the keyboard shortcuts "j" "k" "\<Up>" and "\<Down>" (or any other custom user shortcuts mapped to these actions).

See above (changes cell navigation behavior).

### DIFF
--- a/packages/notebook/src/actions.tsx
+++ b/packages/notebook/src/actions.tsx
@@ -921,7 +921,7 @@ export namespace NotebookActions {
     // find first non hidden cell above current cell
     while (possibleNextCellIndex >= 0) {
       const possibleNextCell = notebook.widgets[possibleNextCellIndex];
-      if (!possibleNextCell.inputHidden && !possibleNextCell.isHidden) {
+      if (!possibleNextCell.isHidden) {
         break;
       }
       possibleNextCellIndex -= 1;
@@ -969,7 +969,7 @@ export namespace NotebookActions {
     // find first non hidden cell below current cell
     while (possibleNextCellIndex < maxCellIndex) {
       let possibleNextCell = notebook.widgets[possibleNextCellIndex];
-      if (!possibleNextCell.inputHidden && !possibleNextCell.isHidden) {
+      if (!possibleNextCell.isHidden) {
         break;
       }
       possibleNextCellIndex += 1;


krassowski:
> This issue makes running or navigating to collapsed cells impossible without either 1) clicking the cell (which also automatically expands the cell) or 2) using the shift + enter shortcut (which requires running preceding cells).

Some thoughts here:
- should it be possible to press Arrow Left to expand the collapsed cells?
- should we have two variants of <Up>,<Down>,k,j? Like `Alt` + `Up` expands, just `Up` does not (or vice versa)?

As for the PR in the current state it looks like the failing `js-notebook` test is related and would need to be updated:

```
  ● @jupyterlab/notebook › NotebookActions › #selectAbove() › should skip collapsed cells in edit mode

    Expected: 0
    Received: 2

      1239 |         widget.widgets[3].inputHidden = false;
      1240 |         NotebookActions.selectAbove(widget);
    > 1241 |         expect(widget.activeCellIndex).toBe(0);
           |                                        ^
      1242 |       });
      1243 |     });
      1244 |

      at Object.<anonymous> (test/actions.spec.ts:1241:40)
```

krassowski:
@lukelbd are you interested in finishing up this PR or should we close it for now?

Issue #16892, opened 2024-10-24
Reduce default autosave timeout


JasonWeill:
### Problem

The autosave timeout, introduced in #5645, is 120 seconds by default:

https://github.com/jupyterlab/jupyterlab/blob/090fe7383422b59588f0337fe0b5ee0d74899204/packages/docmanager-extension/schema/plugin.json#L119-L124

Users today expect autosave to work more quickly than every 2 minutes. A connection can easily become lost within 2 minutes, raising the risk of data loss.

### Proposed Solution

Reduce the autosave timeout to something much shorter by default, such as 5 seconds. Update all defaults, to match the new default.

### Additional context

Prior JupyterLab issue: #5619.

Maduflavins:
@JasonWeill can I take up this task?

JasonWeill:
@Maduflavins You can open a pull request if you want to, but I'd like to discuss this with our other contributors at triage next week before we merge any code in to address this.

Maduflavins:
okay thank you

Maduflavins:
Hi @JasonWeill I have opened a PR at https://github.com/jupyterlab/jupyterlab/pull/16893 you can have a look 
Thank you.

Maduflavins:
@JasonWeill, I tried including the labels, but it looks like I do not have permission to do so. Can someone help me includeciated with this issue in the PR? 
Thank you.

Maduflavins:
@JasonWeill  the regression ui test failed does this have anything to do with the changes i made. i do not see a link myself though and is there something i can do to resolve it meanwhile i saw your comment on the PR do you want me to make that change in this same PR?

Maduflavins:
@JasonWeill did you get the chance to look at what @echarles said about the PR, I agree with his concern also I will be making changes to raise the default to 45secs against the 60secs he suggested

JasonWeill:
@Maduflavins Thanks for opening this pull request! We plan to review this and other newly-filed issues on Tuesday at 09:00 PDT (16:00 UTC) as usual. You can find information about joining this triage meeting on the [Jupyter community calendar](https://discourse.jupyter.org/t/jupyter-community-calendar/2485), if you're interested and available.

JasonWeill:
Per comments on #16893, including from @echarles , we need to determine what the impact is of saving a file more frequently than 2 minutes (120 seconds) after the most recent edit. Would decreasing the autosave timeout from 120s to 5s, for example, impose excessive load on the server?

[...]

PR #16893, opened 2024-10-25
reduced the time for the autosaveInterval from 120sec to 5sec

Maduflavins:
## issue number 
#16892 

## Code changes
changed the autosaveInterval default time  from 120secs to 5ecs and updated all defaults to match the new default.

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -262,7 +262,7 @@ const docManagerPlugin: JupyterFrontEndPlugin<void> = {
       const autosaveInterval = settings.get('autosaveInterval').composite as
         | number
         | null;
-      docManager.autosaveInterval = autosaveInterval || 120;
+      docManager.autosaveInterval = autosaveInterval || 5;
 
       // Handle last modified timestamp check margin
       const lastModifiedCheckMargin = settings.get('lastModifiedCheckMargin')
--- a/packages/docmanager/src/manager.ts
+++ b/packages/docmanager/src/manager.ts
@@ -128,7 +128,7 @@ export class DocumentManager implements IDocumentManager {
         if (!handler) {
           return;
         }
-        handler.saveInterval = value || 120;
+        handler.saveInterval = value || 5;
       });
       this._stateChanged.emit({
         name: 'autosaveInterval',
@@ -713,7 +713,7 @@ export class DocumentManager implements IDocumentManager {
   private _widgetManager: DocumentWidgetManager;
   private _isDisposed = false;
   private _autosave = true;
-  private _autosaveInterval = 120;
+  private _autosaveInterval = 5;
   private _lastModifiedCheckMargin = 500;
   private _renameUntitledFileOnSave = true;
   private _when: Promise<void>;
--- a/packages/docmanager-extension/schema/plugin.json
+++ b/packages/docmanager-extension/schema/plugin.json
@@ -120,7 +120,7 @@
       "type": "number",
       "title": "Autosave Interval",
       "description": "Length of save interval in seconds",
-      "default": 120
+      "default": 5
     },
     "confirmClosingDocument": {
       "type": "boolean",


packages/docmanager-extension/src/index.tsx
```diff
@@ -262,7 +262,7 @@ const docManagerPlugin: JupyterFrontEndPlugin<void> = {
       const autosaveInterval = settings.get('autosaveInterval').composite as
         | number
         | null;
-      docManager.autosaveInterval = autosaveInterval || 120;
```
JasonWeill, line 265: For ease of changing this again in the future, it would be better to make this a symbolic constant (e.g., `DEFAULT_AUTOSAVE_INTERVAL`).

Maduflavins: done @JasonWeill

packages/docmanager/src/manager.ts
```diff
@@ -128,7 +128,7 @@ export class DocumentManager implements IDocumentManager {
         if (!handler) {
           return;
         }
-        handler.saveInterval = value || 120;
```
JasonWeill, line 131: This file does not use the symbolic constant you created earlier

Maduflavins: @JasonWeill  I have added the symbolic constant to the constants file and imported the constant from the file. I believe this should fix the issue.

echarles:
Is 5s not too low with a risk of putting down the network and server with too many requests if many users are constantly typing?

Maduflavins:
> Is 5s not too low with a risk of putting down the network and server with too many requests if many users are constantly typing?
you raised a valid point, what time would you suggest?

echarles:
> you raised a valid point, what time would you suggest?

Difficult to say without real benchmarks. TBH I find 1 minute a good tradeoff, but that is something that I would prefer seeing confirmed in real, as getting more feedbacks from the community.

Maduflavins:
> > you raised a valid point, what time would you suggest?
> 
> Difficult to say without real benchmarks. TBH I find 1 minute a good tradeoff, but that is something that I would prefer seeing confirmed in real, as getting more feedbacks from the community.

okay, let's see what the rest of the team suggest

JasonWeill:
@echarles A five-second timeout would produce, at the very most, 12 "save" command executions per minute per user. Realistically, because it only triggers 5 seconds after the most recent edit, we would see a lot fewer than this. What metrics would we need to collect to determine the impact of saving a file on server performance?

Maduflavins:
> > you raised a valid point, what time would you suggest?
> 
> Difficult to say without real benchmarks. TBH I find 1 minute a good tradeoff, but that is something that I would prefer seeing confirmed in real, as getting more feedbacks from the community.

@echarles did you take a look at  @JasonWeill comment?

echarles:
> What metrics would we need to collect to determine the impact of saving a file on server performance?

I would take the biggest notebook you have ever work with, go to a place with a 2 bars wifi and see what it gives.

krassowski:
Working with notebooks on remote deployments I often see long delays for save operation (2-5 seconds) because my notebooks can include large visualisations and often are 20-50 MB in size (panel/holoviz dumping all their assets in the outputs are one culprit, SVG publication-quality visualisations are another). Even though I have ultra fast broadband at home the remote Jupyter Hubs often have limited capability to receive data (and upload is usually slower than download).

For example, in the UK the average upload speed is 18,4Mbps (2.3 MB/s). This means that a notebook with 50 MB takes 21s to upload for save (disregarding overheads). I think it is reasonable to change the save interval to once a minute or once every 30 seconds but I agree that once every 5 seconds is not a reasonable value for software which is often deployed on remote cloud or in HPC behind multiple hops of ssh.

I would however be open to having a different default for localhost users. This is if frontend sees that the URL is localhost it could default to 5s.

echarles:
> I would however be open to having a different default for localhost users. This is if frontend sees that the URL is localhost it could default to 5s.

Running the server on localhost will not always imply that the content is served from localhost, see https://github.com/jupyterlab/jupyterlab/pull/16744

packages/docmanager-extension/src/index.tsx
```diff
@@ -54,7 +54,7 @@ import { ISignal, Signal } from '@lumino/signaling';
 import { Widget } from '@lumino/widgets';
 import * as React from 'react';
 import { recentsManagerPlugin } from './recents';
-
+import { DEFAULT_AUTOSAVE_INTERVAL } from '../../../notebook/constants';
```
krassowski, line 57: Import like will not work. It should import from `@jupyterlab/notebook/lib/constants` instead, although `lib` imports are best avoided.

krassowski (changes requested):
I think the consensus is that 5 seconds is not the ideal default. Also, the tests are failing due to invalid relative import. Both things should be easy to fix :)

packages/docmanager-extension/src/index.tsx
Maduflavins: okay noted will change the 5 seconds to 10 seconds then

krassowski:
Given the findings in https://github.com/jupyterlab/jupyterlab/issues/17017 I think this could be only considered if also making sure that notebooks that have not changed do not trigger auto-save.

PR #16896, opened 2024-10-28
Export user preference settings to a json file

Darshan808:
## Overview

This pull request implements the functionality to export user preference settings from the JupyterLab Settings Editor to an `overrides.json` file. 

## Changes Made

- Introduced a new command `settingeditor:export` to handle the export action.
- Added a toolbar button labeled "Export Settings" with an appropriate icon.
- Implemented the logic to compile user-modified settings and write them to the `overrides.json` file.

## Notes

This is not a complete PR. I want to discuss some things:
- I wasn't clear on what we should exactly do with the file. Currently, I have implemented logic to download it. If the file needs to be saved in the **Application Settings Directory**, we might need to write server-side logic as well.
- Can I get some guidance on this?

## Snapshot

Screenshot from 2024-10-28 23-38-01

I'm also open to writing tests and updating docs after the feature is finalized.

### DIFF
--- a/galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
+++ b/galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json
@@ -2274,6 +2274,12 @@
     ]
   },
   {
+    "id": "settingeditor:export",
+    "label": "Export User Settings",
+    "caption": "",
+    "shortcuts": []
+  },
+  {
     "id": "settingsmenu:open",
     "label": "Open Settings Menu",
     "caption": "",
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -315,3 +315,17 @@ test('Keyboard Shortcuts: validate "Or" button behavior when editing shortcuts',
     'settings-shortcuts-edit.png'
   );
 });
+
+test('Settings Export: Clicking the export button triggers a download', async ({
+  page
+}) => {
+  await page.evaluate(async () => {
+    await window.jupyterapp.commands.execute('settingeditor:open');
+  });
+  const downloadPromise = page.waitForEvent('download', { timeout: 5000 });
+
+  await page.getByText('Export Settings').click();
+  const download = await downloadPromise;
+
+  expect(download).toBeDefined();
+});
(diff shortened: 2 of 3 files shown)


Darshan808:
@jtpio @krassowski 
I’d appreciate it if you could review this. Any guidance or feedback would be very helpful!

jtpio:
Thanks @Darshan808 for opening this PR, it looks promising!

Also cc @bollwyvl who was commenting about this in previous issues, and may be interested.

bollwyvl:
Yeah, there are a few places an overrides can go:
- part of a `overrides.d` for an extension author
- into a binder
- into a jupyterlite

The other side of this would be _Import Settings_ to accept an uploaded `overrides.json` and apply it locally. That one in specific might need a "preview changes" panel which would be more involved (e.g. a set of checkboxes to allow skipping a specific extension). But this is a fine first start.

Darshan808:
Are there any other places needing documentation for this? And are tests necessary for this ?

jtpio:
Looks good @Darshan808, and CI is also passing, thanks!

Maybe the test could check the downloaded `overrides.json` contains some settings, after making changes to some settings?

krassowski:
Maybe we should use Download icon instead of Save icon in the toolbar, just to avoid user thinking that they need to click on the save icon to save settings?

Darshan808:
> Maybe we should use Download icon instead of Save icon in the toolbar, just to avoid user thinking that they need to click on the save icon to save settings?

You're right.  I've updated the icon accordingly.

Darshan808:
> @Darshan808 I've sent an invite to the triage team so that you will be able to add the labels on PRs (and issues) :)

Thank you so much for the invite! I'm super excited to contribute more and help out with the triage team!

Darshan808:
@krassowski 
Is https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#labels-used-by-triagers the only resource I need, or are there other resources I should check out too?

krassowski:
Yes, that's good for issues. I think we don't have a guide on how to apply labels to PRs, but basically "feature" for big new features, "enhancement" for small improvements, "bug" for bug fixes, "maintenance" for things without user impact, "documentation" for docs updates (all others are handled by the auto-labeller).

jtpio (approved):
Thanks!

Just tried with the latest changes and it works well :+1: 

Looks like we can get it in so it will be available in the first `4.4.0` pre-release. A follow-up could be look into *importing* settings from an `overrides.json`, but this can be tracked in a separate issue.

PR #16900, opened 2024-10-29
Abort saving if a file cannot be saved

JasonWeill:
## References

Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/364.

## Code changes

Aborts saving if the `canSave` property of the context is `false`. This is `false` when the context's model is in collaborative mode.

## User-facing changes

When the user presses <kbd>CTRL</kbd>+<kbd>S</kbd> or <kbd>⌘</kbd>+<kbd>S</kbd> with Jupyter Collaboration installed, the save command does nothing, and does not cause an error dialog to appear. Without Jupyter Collaboration installed, the command works as it did before.

None.

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -599,6 +599,7 @@ export class Context<
       if (!this.canSave) {
         // File cannot be saved. The "save" command is disabled in the UI,
         // but if the user tries to save anyway, act as though it succeeded.
+        this._saveState.emit('completed');
         return Promise.resolve();
       }
 
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -596,9 +596,10 @@ export class Context<
 
     try {
       await this._manager.ready;
-      if (!this.canSave) {
-        // File cannot be saved. The "save" command is disabled in the UI,
-        // but if the user tries to save anyway, act as though it succeeded.
+      if (this._model.collaborative) {
+        // Files cannot be saved in collaborative mode. The "save" command
+        // is disabled in the UI, but if the user tries to save anyway, act
+        // as though it succeeded.
         this._saveState.emit('completed');
         return Promise.resolve();
       }
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -597,8 +597,9 @@ export class Context<
     try {
       await this._manager.ready;
       if (!this.canSave) {
-        // File cannot be saved.
-        return Promise.reject();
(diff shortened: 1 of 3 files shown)


packages/docregistry/src/context.ts
```diff
@@ -596,6 +596,11 @@ export class Context<
 
     try {
       await this._manager.ready;
+      if (!this.canSave) {
+        // File cannot be saved.
+        return Promise.reject();
+      }
```
srdas, line 602: Just for fun, I asked Claude to review this and it reverted with this suggestion: 

```
if (!this.canSave) {
    // File cannot be saved.
    return Promise.reject(new Error('File cannot be saved')); // Add error message
}       

Feel free to ignore Claude! Else it all LGTM.

```

srdas (approved):

jtpio:
Thanks @JasonWeill for looking into this :+1: 

The test failures appear to be related though:

image

packages/docregistry/src/context.ts
```diff
@@ -596,6 +596,13 @@ export class Context<
 
     try {
       await this._manager.ready;
+      if (!this.canSave) {
+        // File cannot be saved. The "save" command is disabled in the UI,
+        // but if the user tries to save anyway, act as though it succeeded.
+        this._saveState.emit('completed');
+        return Promise.resolve();
+      }
```
krassowski, line 604: It would be good to consider what effect it has on the logic for updating the model hash (which is needed to prevent spurious "File Changed" dialogs). This condition prevents updating `contentHash` as it is performed in `_maybeSave` method and this condition prevents ever reaching it. I _hope_ this would not be a problem for RTC because it now uses `fileChanged` signal to propagate new `contentHash` since:
- https://github.com/jupyterlab/jupyterlab/pull/16695
- https://github.com/jupyterlab/jupyter-collaboration/pull/337

I also _think_ that it would not make the situation worse for other drives either as in that case they would not update the hash as they could not save the content.

I think that all we need to do is document the side-effects when saving and when saving is skipped, especially if we are going to emit `saveState` even if the save did not happen.

krassowski:
I think the tests failing might be related to `canSave` being `false` when `_contentsModel` is `null`:

https://github.com/jupyterlab/jupyterlab/blob/8a08b6dc652453becd935999cdfa7556131bb170/packages/docregistry/src/context.ts#L225-L230

JasonWeill:
@krassowski Tests are passing now! Could you please take another look when you get the chance?

JasonWeill:

JasonWeill:

PR #16903, opened 2024-10-30
Maintain autosave timers while disconnected

### LINKED ISSUE #16798: fix for report of auto-save not working?
hello - did the fix proposed in https://discourse.jupyter.org/t/auto-save-does-not-always-work/24756 ever get submitted (and merged into) JupyterLab?

I have gotten a similar report from a user on one of our Jupyter instances (JupyterLab 4.2.4, JupyterHub 5.1.0), though we have not been able to reproduce the issue.

holzman:
## References
 
Fixes #16798.

## Code changes

The autosave feature works by starting a timer.  When the timer expires, it clears itself; then, if the session is connected, it attempts a save (which then restarts the timer for the next save).  However, if the session is disconnected, the timer never gets restarted - this code just calls `_setTimer()` in that case.

### DIFF
--- a/packages/docmanager/test/savehandler.spec.ts
+++ b/packages/docmanager/test/savehandler.spec.ts
@@ -49,6 +49,25 @@ describe('docregistry/savehandler', () => {
       });
     });
 
+    describe('#setTimer()', () => {
+      it('should reset autosave timer when disconnected', async () => {
+        jest.useFakeTimers();
+
+        jest
+          .spyOn(handler as any, '_isConnectedCallback')
+          .mockReturnValue(false);
+        jest.spyOn(handler as any, '_setTimer');
+        jest.spyOn(handler as any, '_save');
+
+        handler.saveInterval = 120;
+        handler.start();
+        jest.advanceTimersByTime(120000); // in ms
+        expect((handler as any)._setTimer).toHaveBeenCalledTimes(2);
+
+        jest.useRealTimers();
+      });
+    });
+
     describe('#saveInterval()', () => {
       it('should be the save interval of the handler', () => {
         expect(handler.saveInterval).toBe(120);
--- a/packages/docmanager/test/savehandler.spec.ts
+++ b/packages/docmanager/test/savehandler.spec.ts
@@ -49,25 +49,6 @@ describe('docregistry/savehandler', () => {
       });
     });
 
-    describe('#setTimer()', () => {
-      it('should reset autosave timer when disconnected', async () => {
-        jest.useFakeTimers();
-
-        jest
-          .spyOn(handler as any, '_isConnectedCallback')
-          .mockReturnValue(false);
-        jest.spyOn(handler as any, '_setTimer');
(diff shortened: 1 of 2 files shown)


holzman:
Added a test.

JasonWeill:
@krassowski All tests are now passing. OK to merge and backport to 4.3.x?

kellyrowland:
This should get backported further back than 4.3.x, no?

JasonWeill:
We typically only backport changes further upon request. @kellyrowland How far back should we backport this fix once it's been merged into main?

kellyrowland:
Ideally as far back as the version where the bug was introduced, but I would settle for 4.2.x 🙂

packages/docmanager/test/savehandler.spec.ts
```diff
@@ -49,6 +49,25 @@ describe('docregistry/savehandler', () => {
...
+        jest
+          .spyOn(handler as any, '_isConnectedCallback')
+          .mockReturnValue(false);
+        jest.spyOn(handler as any, '_setTimer');
+        jest.spyOn(handler as any, '_save');
+
+        handler.saveInterval = 120;
+        handler.start();
+        jest.advanceTimersByTime(120000); // in ms
+        expect((handler as any)._setTimer).toHaveBeenCalledTimes(2);
```
krassowski, line 65: Ideally tests should interface with the public API rather than private implementation details, as the implementation may change more rapidly and the test will need modifying (or worse would get deleted). That said I do not want to hold this off so if anyone else is happy to approve as-is I will not block it.

holzman: Good point - I've revised the test.

holzman:
BTW, while poking around the docmanager tests, I found one that wasn't really testing anything and fixed it in #16933.

krassowski (approved):
Thank you @holzman!

kellyrowland:
Is there an ETA for when these backports will get a release? We'd really like to get the fix out to our users.

JasonWeill:
@jupyterlab/release Per comment above, is now an appropriate time to cut new 4.2.x and 4.3.x releases to include backports of this fix?

krassowski:
Yes 4.3.x is on the menu today. Not sure if 4.2.x is ready to go I will let @JasonWeill make the call.

krassowski:
As of today it looks like this would be the only user-facing change on 4.2.6: https://github.com/jupyterlab/jupyterlab/compare/v4.2.5...4.2.x.

JasonWeill:
@kellyrowland Version 4.2.6 is now available, which includes this change: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.2.6

kellyrowland:
Thank you!

JasonWeill:
[v4.3.1](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.1) is now also released, and that also includes this fix. Thank you @krassowski !

PR #16907, opened 2024-10-31
Update the form schema if it changed

brichet:
This PR allows updating the settings form when the schema has been programmatically updated.

Peek 2024-10-31 12-00

## Code changes

Save the schema in the context of the form, and update the displayed item if it changes.

## User-facing changes

Update the form in live.

Should not be.


jtpio (approved):
Thanks!

PR #16921, opened 2024-11-05
Reuse serverSettings when reopen an existing terminal

### LINKED ISSUE #16920: Don't create new server settings when listing running terminals
When you reopen a closed but not shutdown terminal widget, currently the call to `TerminalAPI.listRunning()` at:
https://github.com/jupyterlab/jupyterlab/blob/f8cdbba3b8ac64c80f2c7701f35db4c568716f3f/packages/terminal-extension/src/index.ts#L363
does not pass in any server settings so a new one is created:
https://github.com/jupyterlab/jupyterlab/blob/f8cdbba3b8ac64c80f2c7701f35db4c568716f3f/packages/services/src/terminal/restapi.ts#L70-L71

It is not necessary to create new server settings as we can reuse the existing one, as occurs in the only other use of terminal `listRunning`:
https://github.com/jupyterlab/jupyterlab/blob/f8cdbba3b8ac64c80f2c7701f35db4c568716f3f/packages/services/src/terminal/manager.ts#L217

I found this whilst working on the JupyterLite terminal extension (jupyterlite/terminal#12) as use of new server settings means we end up trying to use a real websocket rather than the mocked equivalent that JupyterLite supports.

I am not sure if this should be classified as a bug or enhancement. It is a one-line change, I will submit a PR shortly.

ianthomas23:
Fixes #16920.

Single line change to reuse existing server settings when requesting the list of running terminals. There is no user-facing change here, it just avoids unnecessarily creating a new object.

### DIFF
--- a/packages/terminal-extension/src/index.ts
+++ b/packages/terminal-extension/src/index.ts
@@ -360,7 +360,7 @@ function addCommands(
 
       let session;
       if (name) {
-        const models = await TerminalAPI.listRunning();
+        const models = await TerminalAPI.listRunning(serviceManager.serverSettings);
         if (models.map(d => d.name).includes(name)) {
           // we are restoring a terminal widget and the corresponding terminal exists
           // let's connect to it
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

jtpio:

PR #16933, opened 2024-11-06
Fix `docmanager/savehandler` "continue to save" test

holzman:
There were two problems with the previous test so that it didn't really test what was intended.

When the signal was triggered and the test() method was called, the promise was resolved regardless of test()'s return value. The subsequent change to the context.model inside test() then did not trigger anything.  We switch to find() which only resolves the promise when the return value is true.

Additionally, the initial call to the triggered function would always return true due to the ordering of the increment and check of the call counter; this has also been fixed.


holzman:
I think this code hasn't been testing what was intended since it was updated to use `testEmission` in a9f0ff16bd340e6d2b8fbd0e19c3368acc8cf9e9

krassowski (approved):
Thank you @holzman!

Issue #16935, opened 2024-11-06
Improve styling on cell being dragged


JasonWeill:
### Problem

This is what it looks like when the user has just picked up a cell, and is dragging the mouse to a dropoff point:

![image](https://github.com/user-attachments/assets/db84d6c5-051e-41d1-9085-62139337fd40)

This mini-cell could be improved in ways such as: 

* The top of the box has effectively no padding, whereas the bottom has plenty of padding.
* The prompt (`[ ]:`) is less greyed out than the text snippet is.
* There is no syntax highlighting on the inner text.

### Proposed Solution

These could be separated into separate issues.

* Use equal padding on the top and bottom of the dragged cell preview.
* In the dragged cell, have the prompt and the text snippet have similar contrast with their respective backgrounds.
* Apply syntax highlighting to the innter text.

### Additional context

* macOS 14.7 "Sonoma"
* Firefox 131.0.2 (aarch64)
* JupyterLab 4.3.0

JasonWeill:
This may be an issue with the Lumino dragdrop widget. The animation in https://github.com/jupyterlab/lumino/issues/347, from August 2022, suggests that there used to be more even margins in the preview, but with the same uneven contrast problem.

JasonWeill:
When multiple cells are being dragged, the "background" cell has a higher-contrast border than the "foreground" cell does.

<img width="1461" alt="Screenshot 2024-11-06 at 3 55 56 PM" src="https://github.com/user-attachments/assets/6361a2c3-9474-4c75-8a02-0198412266d7">

JasonWeill:
Looks like the uneven padding was due to a switch from `flex` to `block` for the display type of the drag image, introduced in #10127 by @goanpeca. (This is a port of  #8968 by @saulshanabrook, originally for Lab 2.2.x.) Reverting the drag image's display style to `flex` fixes it; the rest of this element and its subelements assume that the display type is `flex` anyway.

PR #16936, opened 2024-11-07
Improve drag image styling

JasonWeill:
## References

Partial fix for #16935.

## Code changes

Back out a layout change in the drag image, introduced for performance, that caused a regression in visual appearance.

## User-facing changes

Before the change, the drag image (the image that represented the selected cells) was not centered inside the container:

image

After the change, the drag image has equal padding on all four sides:

Screenshot 2024-11-06 at 4 48 42 PM

None.

### DIFF
--- a/packages/documentsearch/src/searchview.tsx
+++ b/packages/documentsearch/src/searchview.tsx
@@ -670,7 +670,7 @@ class SearchOverlay extends React.Component<ISearchOverlayProps> {
               description={
                 description + (name == 'selection' ? selectionKeyHint : '')
               }
-              isEnabled={isEnabled}
+              isEnabled={!showReplace || filter.supportReplace}
               onToggle={async () => {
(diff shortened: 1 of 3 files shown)


jtpio (approved):
Thanks!

jtpio:
The `check-links` failure appears to be unrelated and should hopefully fix itself at some point:

image

jtpio:
Ah I was wondering why we are still seeing this issue in JupyterLab 4.3.2:

[juyterlab-drag-image-issue-432.webm](https://github.com/user-attachments/assets/9b0f42b0-ab8b-4a15-97ee-a5a47405033c)

But it's probably because it was not backported, maybe because it was set as `enhancement`. Although in reality it looks more like a bug fix, so we should be able to backport the change.

jtpio:

Issue #16925, opened 2024-11-05
`Find and Replace` / `Use Regular Expressions` / `Replace All`: reference groups are NOT handled correctly


RRosio:
Reporting this issue upstream: https://github.com/jupyter/notebook/issues/7503

PR #16940, opened 2024-11-07
Fix Regex Functionality for Find and Replace / Replace All

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/16925

### DIFF
--- a/packages/cells/src/searchprovider.ts
+++ b/packages/cells/src/searchprovider.ts
@@ -256,11 +256,14 @@ class CodeCellSearchProvider extends CellSearchProvider {
    * @param newText The replacement text.
    * @returns Whether a replace occurred.
    */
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Thank you @itsmevichu!

krassowski:

packages/cells/src/searchprovider.ts
```diff
@@ -256,11 +256,14 @@ class CodeCellSearchProvider extends CellSearchProvider {
    * @param newText The replacement text.
    * @returns Whether a replace occurred.
    */
-  async replaceAllMatches(newText: string): Promise<boolean> {
+  async replaceAllMatches(
+    newText: string,
+    options?: IReplaceOptions
```
kuraga, line 261: @krassowski , should we change the above docstring?

PR #16945, opened 2024-11-10
Enable Scroll for Overflowing Menus on Small Screens

### LINKED ISSUE #16944: Not Scrolling menu, i can't shutting down jupyter lab
## Description

i can not shut down my jupyter lab after updating **4.3.0** version. my screen size is "**1366x768**".  i can not scroll it 
Image

## Reproduce

use any **1366x768** monitor

## Expected behavior

Scrolling this section
Image

## Context

- Operating System and version: Windows 11 23H2
- Browser and version: Google Chrome 130.0.6723.117
- JupyterLab version: 4.3.0

## Troubleshooting
i have tried different browsers. but the problem is the same.

Darshan808:
## References
Closes #16944

### Code Changes
Added scroll behavior to menus to handle overflow.

### User-Facing Changes
Users on small screens can now scroll within menus to access options and buttons at the bottom.

https://github.com/user-attachments/assets/7c9e9f1b-f30a-4ec9-8be5-ce53271687e6

### DIFF
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -29,6 +29,13 @@
 
 .lm-MenuBar-menu.jp-ThemedContainer {
   top: calc(-2 * var(--jp-border-width));
+  overflow: scroll;
+  scrollbar-width: none;
+  -ms-overflow-style: none;
+}
+
+.lm-MenuBar-menu.jp-ThemedContainer::-webkit-scrollbar {
+  display: none;
 }
 
 .lm-MenuBar-item {
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -39,8 +39,8 @@
         farthest-side at 50% 0,
         color-mix(
           in hsl,
-          var(--jp-layout-color0) 30%,
-          var(--jp-inverse-layout-color1) 50%
+          var(--jp-layout-color0) 50%,
+          var(--jp-inverse-layout-color0) 30%
         ),
         rgba(0, 0, 0, 0)
       )
@@ -49,8 +49,8 @@
         farthest-side at 50% 100%,
         color-mix(
           in hsl,
-          var(--jp-layout-color0) 30%,
-          var(--jp-inverse-layout-color1) 50%
+          var(--jp-layout-color0) 50%,
+          var(--jp-inverse-layout-color0) 30%
         ),
         rgba(0, 0, 0, 0)
       )
@@ -195,6 +195,7 @@
   position: relative;
   top: 4px;
   border-top: var(--jp-border-width) solid var(--jp-layout-color2);
+  mix-blend-mode: multiply;
 }
 
 /* gray out icon/caret for disabled menu items */
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -29,9 +29,40 @@
 
 .lm-MenuBar-menu.jp-ThemedContainer {
   top: calc(-2 * var(--jp-border-width));
-  overflow: scroll;
   scrollbar-width: none;
   -ms-overflow-style: none;
+  overflow: auto;
+  background:
+    linear-gradient(var(--jp-layout-color0) 30%, rgba(0, 0, 0, 0)) center top,
+    linear-gradient(rgba(0, 0, 0, 0), var(--jp-layout-color0) 70%) center bottom,
+    radial-gradient(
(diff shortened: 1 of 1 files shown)


Darshan808:
Visual Regression Tests are passing locally, but couldn't figure out why they are failing on CI.

krassowski:
I think this is a good idea.

I also agree with disabling the scrollbars, though it would be best to add some kind of alternative overflow indicator (https://css-tricks.com/thats-just-how-i-scroll/ has good explanation, but solutions there require JS). We would want something simple both visually and when it comes to amount of code needed to maintain it. https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/ is a CSS-only solution, what do you think about exploring integrating it in?

Darshan808:
@krassowski 
I'm having an issue with setting up a shadow effect for a scroll indicator. I’m using CSS variables to make the shadows adapt to different themes, as shown below:

```css
radial-gradient(
  farthest-side at 50% 0,
  rgba(255, 255, 255, 0.4), // not adaptive to themes
  // rgba(var(--jp-inverse-layout-color0), 0.4) // adaptive but doesn't work
  rgba(0, 0, 0, 0)
) center top,
```

To make this truly adaptable across themes, I need to replace `rgba(255, 255, 255, 0.4)` with something like `rgba(var(--jp-inverse-layout-color0), 0.4)`. However, the CSS color variables in JupyterLab are either set directly as keywords (like `white`) or in hex format (like `#FFFFFF`) rather than as RGB values (like `255, 255, 255`). Because of this, I’m unable to adjust the alpha (opacity) value directly within the `rgba()` function for theme compatibility.

I’ve spent some time trying to find a workaround but haven’t been successful yet. If you have any suggestions on how to manage this with the current variable setup, I would really appreciate your input.

krassowski:
It sounds like you are looking for [`color-mix`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix). MDN has a section about [adding transparency](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix#adding_transparency) using `color-mix`. As per https://caniuse.com/?search=color-mix all browsers we support implement `color-mix` already so we should be fine using it.

Darshan808:
> It sounds like you are looking for [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix)

This is exactly what I needed! Thank you so much.

I've added the overflow indicator shadows. Please take a look and let me know if you have any feedback or suggestions regarding the height or color of shadows
UI

krassowski:
I wonder if using a dark shadow colour in the dark theme would make it more visually pleasing.

It looks well on https://nextui.org/docs/components/scroll-shadow

krassowski:
We can fix it by adding `mix-blend-mode` on `.lm-Menu-item[data-type='separator'] > div::after`

| mix-blend-mode: multiply | mix-blend-mode: darken |
|--|--| 
| image | image |

I think I slightly prefer `multiply`.

krassowski:
I tried to see if we can use black shadow in dark mode but it contrast is poor because of how dark the menu background is itself. Maybe we can tune the values to make it work but in principle I do not oppose sticking to the light shadow for now while we explore alternative solutions.

Darshan808:
@krassowski 
Thank you for the feedback! I couldn’t review it sooner as I was traveling a long distance. I adjusted the shadow color, and this seems to be the best fit for now.

| Light mode                       | Dark mode                       |
|----------------------------------|---------------------------------|
| Light mode | Dark mode|


> It looks well on https://nextui.org/docs/components/scroll-shadow

I tried this other approach as well, but it looks like it requires JavaScript to modify the menu classes based on which side is overflowing.

For now, I think we can stick with this solution, and if a better option comes up in the future, we can revisit it.

jtpio:
Should this backported to `4.3.x` even if it's labelled as `enhancement`?

krassowski:
IMO it fixes a bug - users are unable to access vital options on small screens (https://github.com/jupyterlab/jupyterlab/issues/16944). It even partially fixes an accessibility issue related to ability to use JupyterLab on high zoom (see https://github.com/jupyterlab/jupyterlab/issues/10004 and https://github.com/jupyterlab/jupyterlab/issues/13148).

I will re-label this as a bug-fix.

krassowski:

Issue #16326, opened 2024-05-14
Scroll Jumps and jarring effect with full windowing mode


singharpit94:
I am using jupyterlab 4.2.0 and observed that the full windowing mode which is enabled by default causes the native scrollbar jumping and causing jarring effects, Attached a gif for more clarity about the issue.
![windowing-bug](https://github.com/jupyterlab/jupyterlab/assets/7591673/2e966c41-93cf-482b-9be9-49b6b3631de5)

Here is the code I used to create tabular tables in all the cells.

```
import datetime as dt
import numpy as np
import pandas as pd
import panel as pn

np.random.seed(7)
pn.extension('tabulator')

df = pd.DataFrame({
    'int': [1, 2, 3],
    'float': [3.14, 6.28, 9.42],
    'str': ['A', 'B', 'C'],
    'bool': [True, False, True],
    'date': [dt.date(2019, 1, 1), dt.date(2020, 1, 1), dt.date(2020, 1, 10)],
    'datetime': [dt.datetime(2019, 1, 1, 10), dt.datetime(2020, 1, 1, 12), dt.datetime(2020, 1, 10, 13)]
}, index=[1, 2, 3])

df_widget = pn.widgets.Tabulator(df, buttons={'Print': "<i class='fa fa-print'></i>"})
df_widget
```

krassowski:
I can reproduce this with the tabulator panel code. I see some jarring with the windowing mode disabled too but it is of different kind. When windowing is enabled I see that the panel tabulator re-renders each time the cell enters the view.

The panel uses `setInterval` trying to detect if the dependencies are available every 25ms for 200 times (5 seconds total). It appears that if a cell was run when out of view, and user scrolls the notebook so that the cell is in the rendering range (but not necessarily visible) within 5 seconds, panel will populate the output and move the scrollbar slightly, but this may not be the major culprit.

The major issue might instead be that the outputs are re-rendered when cell output visibility changes by panel tabulator; this means that it competes with JupyterLab in its ability to optimize out the widgets (outputs) which are not visible to user.  This cannot be reproduced with simple JS rendering in outputs. The re-rendering can be also seen when switching tabs (independent of windowing mode):

![switching-tag-windowing-off](https://github.com/jupyterlab/jupyterlab/assets/5832902/69e956ee-e980-47ff-b3d5-a5b71951689b)

It calls `redrawTable`, which to make things worse trashes the layout:

[...]

PR #16950, opened 2024-11-11
Fix total size estimation in full windowing mode to reduce scrollbar jitter

krassowski:
## References

Alleviates issues caused by the rendered outputs or cells changing height:
- https://github.com/jupyterlab/jupyterlab/issues/16326
- https://github.com/jupyter-book/jupyterlab-myst/issues/243

## Code changes

1) Previously if height of a cell in the middle of the notebook changed then the sizes and offsets of all subsequent cells were discarded. After change the offset delta is instead computed and subsequent sizes and offsets are adjusted rather than being discarded.

2) Previously the size estimates were only stored after the viewport window was moved away from initial null position; this constraint was now reduced to also store sizes when windowing is active but window size has not been computed yet.

3) Previously if new cell was added  (or a cell was removed/type of a cell was changed) the height estimates were discarded for the purpose of scrollbar size calculation; with this PR this is no longer the case (3341517)

## User-facing changes

| Before (4.3.0) | After (this PR) |
|--|--|
| before | after |

Before this change the total scrollbar length was reset after changing height of any cell other than the last. This was causing very jittrery scroll as the scrollbar height was changing as the notebook was scrolled.

| Before (4.3.0) | After (this PR) |
|--|--|
| before | after |

None

### DIFF
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -504,7 +504,10 @@ export abstract class WindowedListModel implements WindowedList.IModel {
   resetAfterIndex(index: number): void {
     const oldValue = this._measuredAllUntilIndex;
     this._measuredAllUntilIndex = Math.min(index, this._measuredAllUntilIndex);
-    // Heal the offsets
+    // Because `resetAfterIndex` is always called after an operation modifying
+    // the list of widget sizers, we need to ensure that offsets are correct;
+    // e.g. removing a cell would make the offsets of all following cells too high.
+    // The simplest way to "heal" the offsets is to recompute them all.
     for (const [i, sizer] of this._widgetSizers.entries()) {
       if (i === 0) {
         continue;
@@ -583,7 +586,7 @@ export abstract class WindowedListModel implements WindowedList.IModel {
             allPreviousMeasured = false;
           }
         }
-        if (hadSizer && offsetDelta != 0) {
+        if (hadSizer && offsetDelta !== 0) {
           // Adjust offsets for all items where it is needed
           sizer.offset += offsetDelta;
         }
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -504,7 +504,7 @@ export abstract class WindowedListModel implements WindowedList.IModel {
   resetAfterIndex(index: number): void {
     const oldValue = this._measuredAllUntilIndex;
     this._measuredAllUntilIndex = Math.min(index, this._measuredAllUntilIndex);
-    // heal the offsets
+    // Heal the offsets
     for (const [i, sizer] of this._widgetSizers.entries()) {
       if (i === 0) {
         continue;
@@ -576,7 +576,7 @@ export abstract class WindowedListModel implements WindowedList.IModel {
         // If all items so far have actual size measurements...
         if (allPreviousMeasured) {
           if (sizer.measured) {
-            // and this item has a size measurement, we the can say that
+            // and this item has a size measurement, we can say that
             // all items until now have measurements:
             measuredAllItemsUntil = index;
           } else {
--- a/packages/ui-components/src/components/windowedlist.ts
+++ b/packages/ui-components/src/components/windowedlist.ts
@@ -481,10 +481,10 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    * @returns The viewport top position and its height
    */
   getSpan(startIndex: number, stopIndex: number): [number, number] {
-    const startSize = this._getItemMetadata(startIndex);
-    const top = startSize.offset;
+    const startSizer = this._getItemMetadata(startIndex);
+    const top = startSizer.offset;
     const stopSize = this._getItemMetadata(stopIndex);
-    const height = stopSize.offset - startSize.offset + stopSize.size;
+    const height = stopSize.offset - startSizer.offset + stopSize.size;
     return [top, height];
   }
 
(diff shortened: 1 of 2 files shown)


krassowski:
This test failure appears related to this PR:

```
    [jupyterlab] › test/jupyterlab/windowed-notebook.test.ts:186:7 › should scroll back to the active markdown cell on typing 
```

I will switch it do draft, although I anticipate that this only requires fixing the test assumptions.

krassowski:
It seems that this code:

https://github.com/jupyterlab/jupyterlab/blob/55213d8bca9d133f89e4b6fa989f372b7c3da9a9/packages/ui-components/src/components/windowedlist.ts#L553-L560

does not do what it was intended to do. No unit tests to catch this :disappointed:

This is because `new Array(10).map()` does not actually populate the array. It's a common gotcha in JS: https://stackoverflow.com/questions/5501581/why-does-the-map-method-apparently-not-work-on-arrays-created-via-new-arrayc

This leads to `null` being inserted into the `widgetSizers` array.

andrewfulton9:
I was able to test and it worked for me. The logic makes sense and is written cleanly.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -514,24 +527,76 @@ export abstract class WindowedListModel implements WindowedList.IModel {
...
+            sizer.size = measuredSize;
+            minIndex = Math.min(minIndex, index);
+          }
+          // Always set the flag in case the size estimator provides perfect result
+          sizer.measured = true;
+        }
+        // If all items so far have actual size measurements...
+        if (allPreviousMeasured) {
+          if (sizer.measured) {
+            // and this item has a size measurement, we the can say that
```
krassowski, line 579:
```suggestion
            // and this item has a size measurement, we can say that
```

packages/ui-components/test/windowedlist.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+
+import { WindowedListModel } from '@jupyterlab/ui-components';
+import { Widget } from '@lumino/widgets';
+import { ObservableList } from '@jupyterlab/observables';
+
+const FIXED_ESTIMATED_SIZE = 100;
+
+class TestModel extends WindowedListModel {
+  estimateWidgetSize = (_index: number): number => {
+    return 100;
```
krassowski, line 12:
```suggestion
    return FIXED_ESTIMATED_SIZE;
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -497,9 +502,17 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    * @param index
    */
   resetAfterIndex(index: number): void {
-    const oldValue = this._lastMeasuredIndex;
-    this._lastMeasuredIndex = Math.min(index, this._lastMeasuredIndex);
-    if (this._lastMeasuredIndex !== oldValue) {
+    const oldValue = this._measuredAllUntilIndex;
+    this._measuredAllUntilIndex = Math.min(index, this._measuredAllUntilIndex);
+    // heal the offsets
```
krassowski, line 507:
```suggestion
    // Heal the offsets
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -481,10 +486,10 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    * @returns The viewport top position and its height
    */
   getSpan(startIndex: number, stopIndex: number): [number, number] {
-    const startSize = this._getItemMetadata(startIndex);
-    const top = startSize.offset;
+    const startSizer = this._getItemMetadata(startIndex);
```
JasonWeill, line 489: Why is there now `startSizer` (ending with `r`) and `stopSize` (no `r`)? I see `sizer` in the changes below. If `sizer` should be used from now on, should `stopSize` be `stopSizer`?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -497,9 +502,17 @@ export abstract class WindowedListModel implements WindowedList.IModel {
    * @param index
    */
   resetAfterIndex(index: number): void {
-    const oldValue = this._lastMeasuredIndex;
-    this._lastMeasuredIndex = Math.min(index, this._lastMeasuredIndex);
-    if (this._lastMeasuredIndex !== oldValue) {
+    const oldValue = this._measuredAllUntilIndex;
+    this._measuredAllUntilIndex = Math.min(index, this._measuredAllUntilIndex);
+    // Heal the offsets
```
JasonWeill, line 507: What does "heal" mean in this context?

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -514,24 +527,76 @@ export abstract class WindowedListModel implements WindowedList.IModel {
...
+            // and this item has a size measurement, we can say that
+            // all items until now have measurements:
+            measuredAllItemsUntil = index;
+          } else {
+            allPreviousMeasured = false;
+          }
         }
-        // Always set the flag in case the size estimator provides perfect result
-        this._widgetSizers[key].measured = true;
+        if (hadSizer && offsetDelta != 0) {
```
JasonWeill, line 586:
```suggestion
        if (hadSizer && offsetDelta !== 0) {
```

JasonWeill (commented):
Generally looking good! Just a few comments about the code.

packages/ui-components/src/components/windowedlist.ts
krassowski:
```suggestion
    // Because `resetAfterIndex` is always called after an operation modifying
    // the list of widget sizers, we need to ensure that offsets are correct;
    // e.g. removing a cell would make the offsets of all following cells too high.
    // The simplest way to "heal" the offsets is to recompute them all.
```

packages/ui-components/src/components/windowedlist.ts
krassowski: Yes, it should all be size**r** because these are taken from `widgetSizers` array. I was annoyed with `startSize.size` as this is of course confusing (and why would "size" have an offset). But sizers are objects which obviously have a size and can have a position (and since we are in 1D space it is an offset).

I missed the `stopSize`, I will push a commit.

krassowski: Resolved in 2cfa77e.

krassowski:

Issue #16951, opened 2024-11-11
Completer reconciliator prefix removal is incorrect in second and following lines


krassowski:
## Description

The prefix extraction in:

https://github.com/jupyterlab/jupyterlab/blob/017ae3ffe5f687185cbf7b931440adfd7924fca7/packages/completer/src/reconciliator.ts#L229

does not account for previous lines and therefore extracts incorrect fragments.

## Reproduce

1. Install ipython 8.27.0
2. Disable jedi: `%config IPCompleter.use_jedi=False`
3. Type two lines into a single cell:
```python
from_keys_1 = 0
dict.fr
```
4. Press tab and enter
5. See the result is `dict..fromkeys` 

![repro](https://github.com/user-attachments/assets/5946b157-4a64-4198-ace7-bff62fa48b45)

## Expected behavior

The result is `dict.fromkeys`

## Context

- JupyterLab version: 4.3.0

PR #16953, opened 2024-11-11
Fix prefix removal when reconciling completions from multiple sources

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16951

## Code changes

Accounts for the offset of the line for which completion was requested (where the cursor is).

## User-facing changes

| Before | After |
|--|--|
| before | after |

None


andrewfulton9 (approved):
I was able to verify the fix. Looks good to me.

fcollonval (approved):
Thanks @krassowski

Issue #16958, opened 2024-11-14
Can't drag and drop files within the file browser in 4.3.0


tgourdel:
## Description

Before 4.3.0, we use to be able to drag and drop files into folders within the file browser. It seems it's not possible in 4.3.0.
<details><summary>Browser Console when attempting to drag a file</summary>
<pre>
jlab_core.30ac71af4f…c71af4fabafb542d8:1 Uncaught TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
    at t.createDragImage (jlab_core.30ac71af4f…abafb542d8:1:798327)
    at ce._startDrag (jlab_core.30ac71af4f…abafb542d8:1:786589)
    at ce._evtMousemove (jlab_core.30ac71af4f…abafb542d8:1:781297)
    at ce.handleEvent (jlab_core.30ac71af4f…abafb542d8:1:770417)
</pre>
</details>

## Reproduce

Try to move a file into a folder in the file browser.

## Expected behavior

Previous behavior is to being able to drag and drop files in other folders.

## Context

- Operating System and version: MacOS Sonoma 14.6.1
- Browser and version: Version 1.71.121 Chromium: 130.0.6723.91 (Official Build) (arm64)](https://brave.com/latest/)
- JupyterLab version: 4.3.0

tgourdel:
Correct, I always hide the `Modified` column and haven't thought about it. I should have added a screenshot. Thanks for looking into it.

krassowski:
There is a PR waiting for review here: https://github.com/jupyterlab/jupyterlab/pull/16962 - feel welcome to review. 

Otherwise I think @JasonWeill mentioned some time ago that they had a report of a similar issue and wanted to look into this.

PR #16962, opened 2024-11-14
Fix moving files when `Last Modified` column is hidden

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16958

## Code changes

1. Check if the `Last Modified` column is present in drag image before attempting to remove it
2. Remove the other columns as well (when new columns like checkbox and file size were added no one remembered to add the code for their removal here so we ended up with ugly drag images)

## User-facing changes

1. Users are able to drag-and-drop files in the file browser listing when "Last Modified" column is hidden
2. The drag image is less obtrusive for cases where "File Size" and/or "Checkbox" column are visible

| Before | After |
|--|--|
| image | image |

Note: there is no change in padding due to this PR. It is purely due to not including the checkbox in the drag image. This can be confirmed by looking at Before/After when the other columns are hidden:

<details>

| Before | After |
|--|--|
| image | image |

</details>

Adjustments to padding of the drag image are outside of the scope of this PR as I would like to fix a regression in 4.3.0.

None

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -3358,9 +3358,23 @@ export namespace DirListing {
       fileType?: DocumentRegistry.IFileType
     ): HTMLElement {
       const dragImage = node.cloneNode(true) as HTMLElement;
-      const modified = DOMUtils.findElement(dragImage, ITEM_MODIFIED_CLASS);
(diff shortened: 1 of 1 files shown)


JasonWeill (approved):
Change looks good! Tested locally, working as expected.

jtpio:

PR #16963, opened 2024-11-14
Support kernel subshells

### LINKED ISSUE #16882: Support kernel subshells (JEP91)
[Jupyter Enhancement Proposal 91](https://github.com/ipython/ipykernel/pull/1249) was recently accepted, introducing the idea of kernel subshells which are separate threads of execution running in the same process and hence share the same namespace. An example of their use would be when running a long calculation one could create a new subshell to check on progress in that calculation. IPyKernel `main` branch contains a minimal implementation of kernel subshells (ipython/ipykernel#1249) that is backward compatible and will be included in the upcoming IPyKernel 7 release. Hopefully there will be an alpha release available soon (ipython/ipykernel#1274) for testing purposes.

I propose to make kernel subshell functionality available in JupyterLab so that developers and end users can experiment with it. The idea is that the context menu for a running kernel notebook or console will include an option to create a new subshell of that kernel in a new console. There is a primitive video of this at https://github.com/jupyter/enhancement-proposals/pull/91#issuecomment-1924380771 (although the surrounding discussion is somewhat out of date).

I intend to add a new boolean flag (`--LabApp.kernel_subshells` perhaps) which is `False` by default so that devs/users will have to explicitly opt into the new functionality, and if `False` there will be no change to existing UI and behaviour. Kernels self-advertise their support for subshells by including `supported_features: ['kernel subshells']` in their `kernel_info_reply` messages ([JEP92](https://github.com/jupyter/enhancement-proposals/pull/92)), and again without that there will be no change to existing UI and behaviour for a particular kernel.

`IKernelConnection` will need a new optional string attribute `subshellId`, this will be `null` (or `undefined`) for the main subshell. For a subshell this is in the reply to a `create_subshell_request` message. The `subshellId` is an optional parameter of e.g. `execute_request` messages to indicate which subshell the code should be run in.

I am starting work on this straight away. If anyone needs any clarification then please don't hesitate to ask.

ianthomas23:
This exposes the kernel subshells functionality that will be available in ipykernel >= 7 and other kernels subsequently. To try this out you will need to `pip install --upgrade --pre ipykernel==7.0.0a0` and use the new config flag `--kernel-subshells` such as in `jupyter lab --dev-mode --kernel-subshells`.

I'm creating this as a draft PR as I expect it will need further discussion and/or changes.

## References

Closes #16882.

## Code changes

1. `KernelConnection` has a new `subshellId` property that is either `null` for a main shell (called a parent subshell in [JEP91)](https://github.com/jupyter/enhancement-proposals/pull/91) or a unique string for a subshell (called a child subshell in the JEP).
2. Kernel messages include the `subshell_id`, falling back to `null` if not set.
3. New boolean flag `--kernel-subshells` or `--LabApp.kernel_subshells` that is `false` by default so that users have to explicitly opt-in if they want to use subshells. When disabled, there are no visual or behavioural changes to jupyterlab.
4. Reads the new [supportedFeatures](https://github.com/jupyter/enhancement-proposals/pull/92) from the kernel info reply message to determine if a particular kernel supports subshells.
5. New control messages (request and reply) to create and delete a subshell.
6. New `createSubshellConsole` command.
7. Creation of the subshell occurs in `createConsole` and sets the `subshellId` of the `KernelConnection` there. I do not know if this is the best solution, but others seem more invasive such as having to modify the whole `KernelConnection` creation machinery to support subshells.
8. I have (temporarily) modified the galata CI run to install the ipykernel 7 pre-release to prove that subshells work here.

## User-facing changes

Creating a subshell is here only available by right clicking in an ipykernel notebook, which creates a new subshell running in a new console. When that console is closed, the subshell is deleted in the kernel. This is all best demonstrated by the following video:

https://github.com/user-attachments/assets/e7b398ac-2b65-4ff5-bb7b-df5c8b65f7ec

Video walkthrough:
1. Open new ipykernel notebook.
2. Run cell to set some variable `a`.
3. Run `%subshell` magic, note the `subshell id` is `None` as this is a main shell (parent subshell).
4. RIght-click in the notebook, click `New Subshell Console for Notebook`.
5. New console is opened. 
6. Check the value of `a` in the console to show that this shares the same kernel namespace as the notebook above.
9. Run `%subshell`. Note that this has the same `pid` and `main thread id` as the main shell, but a different (string) `subshell id` and `thread id`. The `subshell list` now has one entry as a single subshell exists. The thread count has increased from 15 to 17 for this kernel; one for the actual subshell thread and one for an extra thread that handles closing the subshell via `anyio` (ipykernel implementation detail).
10. Rerun `%subshell` in the main shell notebook. Note the change in `thread count` and `subshell list`.
11. Close the subshell panel.
12. Rerun `%subshell` in the main shell notebook. Note that the subshell has been deleted as the `thread count` is back to 15 and `subshell list` is empty.

There could be other menu items to create a subshell console, but there is an implicit assumption that only one "create console" is available for a particular file type so we can't have both (existing) "create console" and (new) "create subshell console".

None.


krassowski:
High-level question: why is this opt-in?

packages/services/src/kernel/default.ts
```diff
@@ -251,16 +270,32 @@ export class KernelConnection implements Kernel.IKernelConnection {
...
+      this._pendingMessages = [];
+      this._clearSocket();
+
+      // Clear Lumino signals
+      Signal.clearData(this);
+    };
+
+    if (this._subshellId !== null) {
+      const future = this.requestDeleteSubshell(
+        { subshell_id: this._subshellId },
```
krassowski, line 289: You will need to add `subshell_id` to this list:

https://github.com/jupyterlab/jupyterlab/blob/b30baa1f275de1515bee4d1926866c1bb292b32e/.eslintrc.js#L76-L83

ianthomas23:
> High-level question: why is this opt-in (I mean `LabApp.kernel_subshells`)?

For two reasons:
1. I thought it more likely this would be acceptable if the interface was unchanged by default.
2. There are other downstream changes that will be needed when people have had a change to try it out, such as how to deal with execution count. The full list is in the Future Work section of https://github.com/ipython/ipykernel/pull/1249#issue-2350727126.

Personally I would be happy to emphasise in the docs that this is experimental and have it always enabled. This would avoid having to create the `kernel_subshells` config flag at all.

packages/services/src/kernel/default.ts
```diff
@@ -769,6 +811,72 @@ export class KernelConnection implements Kernel.IKernelConnection {
...
+    ) as Kernel.IControlFuture<
+      KernelMessage.ICreateSubshellRequestMsg,
+      KernelMessage.ICreateSubshellReplyMsg
+    >;
+  }
+
+  /**
+   * Send a `delete_subshell_request` message.
+   *
+   * @hidden
```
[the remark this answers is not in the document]
ianthomas23: No rationale, its just a copy, paste, don't-know-what-I-am-doing error.

I've referenced the JEP for the new messages as they are not yet listed at https://jupyter-client.readthedocs.io/en/stable/messaging.html.

packages/services/src/kernel/default.ts
ianthomas23: Thanks.

jtpio:
Another option for rolling this out could also be to make it opt-in for 4.4.0, and opt-out for 4.5.0?

ianthomas23:
I've added the final message type ("list_subshells") so this is now ready for review.

ianthomas23:
Linting failure is 3 of these:

```bash
/home/runner/work/jupyterlab/jupyterlab/packages/services/src/kernel/messages.ts
  1249:12  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead  @typescript-eslint/ban-types
```
We get away with using `content: {}` in some other files, but not here.

jtpio:
There seems to be another relevant failure from the `js-services` check:

```
Error: test/kernel/ikernel.spec.ts(1607,34): error TS2551: Property 'requestListSubshell' does not exist on type 'IKernelConnection'. Did you mean 'requestDeleteSubshell'?
```

ianthomas23:
> There seems to be another relevant failure from the `js-services` check:
> 
> ```
> Error: test/kernel/ikernel.spec.ts(1607,34): error TS2551: Property 'requestListSubshell' does not exist on type 'IKernelConnection'. Did you mean 'requestDeleteSubshell'?
> ```

Fixed in 95432f5.

jtpio:
> We get away with using `content: {}` in some other files, but not here.

Probably because it's used as a value?

Maybe switch to something like the following will help?

```ts
content: Record<string, unknown>
```

packages/services/src/kernel/default.ts
```diff
@@ -225,6 +240,10 @@ export class KernelConnection implements Kernel.IKernelConnection {
     return this._specPromise;
   }
 
+  supportsSubshells(): boolean {
```
jtpio, line 243: Maybe we can add a quick docstring here for consistency with the rest of the file?

packages/services/src/kernel/kernel.ts
```diff
@@ -537,6 +594,13 @@ export interface IKernelConnection extends IObservableDisposable {
       'clientId' | 'username' | 'handleComms'
     >
   ): IKernelConnection;
+
+  supportsSubshells(): boolean;
```
jtpio, line 598: Similar to the comment above: a small docstring here would be useful for consistency.

packages/services/src/kernel/kernel.ts
```diff
@@ -578,6 +642,8 @@ export namespace IKernelConnection {
      * The unique identifier for the kernel client.
      */
     clientId?: string;
+
+    subshellId?: string;
```
jtpio, line 646: re: docstring

.github/workflows/galata.yml
```diff
@@ -37,6 +37,10 @@ jobs:
           # Build dev-mode
           jlpm run build
 
+      - name: Install ipykernel pre-release that supports subshells (TEMPORARY)
+        run: |
+          pip install --upgrade --pre ipykernel==7.0.0a0
```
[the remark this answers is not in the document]
ianthomas23: I'll change to a lower bound. There will be no further API changes related to subshells in `ipykernel` 7 as that is all determined by the JEP.

ianthomas23:
Docstrings added in 7b00164.

.github/workflows/galata.yml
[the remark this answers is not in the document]
ianthomas23: Done in f27c3d4.

packages/console-extension/src/index.ts
```diff
@@ -405,6 +411,17 @@ async function activateConsole(
       void tracker.save(panel);
     });
 
+    panel.sessionContext.ready.then(async () => {
```
jtpio, line 414: There seems to be one more lint failure related to this line.

ianthomas23:
The remaining linting failure is
```
/home/runner/work/jupyterlab/jupyterlab/packages/console-extension/src/index.ts
Error:   414:5  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```
I don't know which strategy is preferred here. If the `sessionContext` isn't `ready` then the console won't be functional.

jtpio:
> I don't know which strategy is preferred here. If the `sessionContext` isn't `ready` then the console won't be functional.

Probably adding a `.catch` and log the error message would be enough for now.

jtpio:
Thanks @ianthomas23 for the fixes!

Looks like CI is almost passing, with 2 remaining failures in the `js-services` tests:

image

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,82 @@ test.describe('Kernel', () => {
...
+        .fill('%subshell');
+      await page.menu.clickMenuItem('Run>Run Cell (forced)');
+
+      const output2 = subshellConsole
+        .locator('.jp-OutputArea-output')
+        .locator('pre');
+      const text2 = (await output2.innerText()).split('\n');
+      expect(text2[0]).toMatch(/^subshell id:/);
+      const subshellId = text2[0].split(':')[1].trim();
+      expect(text2[5]).toEqual(`subshell list: ['${subshellId}']`);
```
jtpio, line 168: Looks like this assertion fails on CI:

image

ianthomas23:
CI is now passing. I've added some extra `await`s around the `requestCreateSubshell` call to avoid some timing issues. This is probably not ideal but I cannot think of a better approach.

jtpio:
> CI is now passing. I've added some extra `await`s around the `requestCreateSubshell` call to avoid some timing issues.

I guess it would probably fine for now, but it would still be useful to understand why they are needed as they may indicate a potential issue somewhere?

Also wondering if this may make CI a bit more flaky over time.

ianthomas23:
> > CI is now passing. I've added some extra `await`s around the `requestCreateSubshell` call to avoid some timing issues.
> 
> I guess it would probably fine for now, but it would still be useful to understand why they are needed as they may indicate a potential issue somewhere?

The relevant code is here:
https://github.com/ianthomas23/jupyterlab/blob/9a14ba4b4d7dcdfeed1a31b3cbca9891ab3bc31d/packages/console-extension/src/index.ts#L418-L420

A newly created `KernelConnection` does not know if it supports subshells until it receives a `KernelInfoReply` message. This is the first real world use of the optional features JEP (https://github.com/jupyter/enhancement-proposals/blob/master/92-jupyter-optional-features/jupyter-optional-features.md). So if you want to create a new subshell, you have to `await` the `kernel.info` first to be sure you have accurate information. In my human use of this functionality is has never been a problem, but in one of the CI runs here, when things are presumably happening very quickly requesting a subshell failed because it didn't know if they were supported at that time.

jtpio:
> Personally I would be happy to emphasise in the docs that this is experimental and have it always enabled. This would avoid having to create the `kernel_subshells` config flag at all.

Right, I guess it can make sense to have a dedicated flag to turn this feature off on the JupyterLab side. But in this case the feature is also enabled based on the `supported_features` from the kernel, which may already be enough?

Otherwise we can keep the `kernel_subshells` flag for now in the first `4.4` pre-releases, and track its removal before a first 4.4 beta or rc is released?

ianthomas23:
OK, let's keep the `kernel_subshells` flag for the moment, and make a decision about it before the next pre-release.

jtpio (approved):
Thanks!

This looks like a good first step for enabling the use of subshells in JupyterLab :+1: 

It would be great to get this released early in the 4.4 release cycle to have plenty of time to test and iterate on issues and API changes if needed.

cc @krassowski who may want to have another as an early reviewer of this PR

packages/console-extension/src/index.ts
```diff
@@ -405,6 +411,26 @@ async function activateConsole(
...
+          future.onReply = (
+            msg: KernelMessage.ICreateSubshellReplyMsg
+          ): void => {
+            const subshellId = msg.content.subshell_id;
+            panel.sessionContext.session!.kernel!.subshellId = subshellId;
+          };
+          await future.done;
+        })
+        .catch(reason => {
+          console.error('Failed to initialize SessionContext.', reason);
```
krassowski, line 430: I this the right messaging here? What if `sessionContext.ready` resolves ok but `.then()` throws (e.g. due to non-null assertions?). It would seem to me that the responsibility of this coude would be to let user know that subshells won't be available, not make a (possibly incorrect) statement about session context, right?

```suggestion
          console.error('Failed to initialize subshells.', reason);
```

packages/services/src/kernel/default.ts
```diff
@@ -225,6 +240,13 @@ export class KernelConnection implements Kernel.IKernelConnection {
     return this._specPromise;
   }
 
+  /**
+   * Check if this kernel supports JEP 91 kernel subshells.
+   */
+  supportsSubshells(): boolean {
```
krassowski, line 246: Why is this a method rather than a getter?

```suggestion
  get supportsSubshells(): boolean {
```

packages/services/src/kernel/kernel.ts
```diff
@@ -537,6 +594,16 @@ export interface IKernelConnection extends IObservableDisposable {
       'clientId' | 'username' | 'handleComms'
     >
   ): IKernelConnection;
+
+  /**
+   * Check if this kernel supports JEP 91 kernel subshells.
+   */
+  supportsSubshells(): boolean;
```
krassowski, line 601:
```suggestion
  get supportsSubshells(): boolean;
```

krassowski (changes requested):
Thank you @ianthomas23, this is looking good. See some small suggestions on the code/logic and making tests easier to maintain :)

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,83 @@ test.describe('Kernel', () => {
         'Python 3 (ipykernel)'
       );
     });
+
+    test('Should support opening subshell in separate code console ', async ({
```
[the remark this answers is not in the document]
ianthomas23: Done

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,83 @@ test.describe('Kernel', () => {
...
+      const notebook = page.locator('.jp-Notebook');
+      await notebook.waitFor();
+      await notebook
+        .getByLabel('Code Cell Content')
+        .getByRole('textbox')
+        .click();
+      await notebook
+        .getByLabel('Code Cell Content')
+        .getByRole('textbox')
+        .fill('%subshell');
```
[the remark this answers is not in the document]
ianthomas23: Thanks, I am not familiar with the details of the test helpers. Done.

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,83 @@ test.describe('Kernel', () => {
...
+      await page.menu.clickMenuItem('Run>Run Selected Cell');
```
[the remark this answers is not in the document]
ianthomas23: Done

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,83 @@ test.describe('Kernel', () => {
...
+
+      // Rerun %subshell in notebook now that subshell exists
+      await notebook
+        .getByLabel('Code Cell Content')
+        .getByRole('textbox')
+        .nth(0)
+        .click();
+      await page.menu.clickMenuItem('Run>Run Selected Cell');
+
+      const output3 = notebook.locator('.jp-OutputArea-output').locator('pre');
```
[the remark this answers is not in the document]
ianthomas23: Yes, I have added some explanatory comments.

packages/console-extension/src/index.ts
```diff
@@ -405,6 +411,26 @@ async function activateConsole(
       void tracker.save(panel);
     });
 
+    if (options.subshell) {
+      panel.sessionContext.ready
+        .then(async () => {
+          // Ensure kernel has received kernel_info.
+          await panel.sessionContext.session!.kernel!.info;
+          const future =
+            panel.sessionContext.session!.kernel!.requestCreateSubshell({});
```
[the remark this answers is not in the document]
ianthomas23: > I suppose the subshells would only ever work on the initial session and initial kernel and not on subsequent ones?

Yes. If the session or kernel change and they are valid then the console will still be a functioning console, but not to the initial subshell.

I have not considered what to do outside of that. If the kernel changes so it is no longer a subshell it could be argued that the console should no longer exist, but closing seems like a very bad thing to do to a user.

packages/services/src/kernel/default.ts
ianthomas23: No reason. It could be either, I picked one, I didn't know what is preferred in this project. I've changed it to a getter as requested.

packages/services/src/kernel/kernel.ts
ianthomas23: Done

packages/console-extension/src/index.ts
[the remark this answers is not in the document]
krassowski: > but closing seems like a very bad thing to do to a user.

I agree.

> If the kernel changes so it is no longer a subshell

But should it? If user changes kernel, should the console become a subshell console on the new kernel?

Like, many workflows involve "Restart and run all" or "Restart and run up to selected cell". Do we want to require user to use `New Subshell Console for Notebook` command each time after restarting?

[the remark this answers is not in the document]
ianthomas23: I think both options are reasonable and it is not obvious to me which is more appropriate. We are in a catch 22 situation with this subshell console. We need to make subshells available in some way to the small number of interested users so they can try it out and discover how they want to interact with them, but we need that information now before we know what it is.

This is different from the remainder of this PR which is just exposing subshells in `KernelConnection`. That enables us to use them for some concrete, easy to understand use cases such as implementing `ipywidgets` and the debugger using subshells if available, eliminating a number of problems caused by blocking without the end user needing to even know that subshells are being used.

Maybe I have approached this PR incorrectly, trying to combine both when they are separate units of functionality. The core subshells in `KernelConnection` doesn't need to be behind a `--kernel-subshells` flag, it should always be present. The subshell console could be separate, either a separate PR behind its own flag, or maybe even in a separate frontend extension although that will invariably require more maintenance in the long run.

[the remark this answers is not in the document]
krassowski: No, I think this PR is fine. I was just a suggesting a tiny modification, wrapping these lines with:

```ts
panel.sessionContext.kernelChanged.connect(() => {
   // do the registration of subshell with future 
})
```

> The core subshells in `KernelConnection` doesn't need to be behind a `--kernel-subshells` flag, it should always be present

I agree.

> The subshell console could be separate, either a separate PR behind its own flag

No, I think it is good to have it here as it gives a concrete path to testing this and allows for the integration test.

In any case, while I think this is what users would prefer, if it is not clear for you I call it a draw and let's move forward merging this PR once all the other bits are resolved :)

galata/test/jupyterlab/kernel.test.ts
```diff
@@ -104,6 +104,72 @@ test.describe('Kernel', () => {
...
+      await output1.click({ button: 'right' });
+      await page
+        .getByRole('menuitem', { name: 'New Subshell Console for Notebook' })
+        .click();
+
+      // Run %subshell in console
+      const subshellConsole = page.locator('.jp-CodeConsole');
+      await subshellConsole.waitFor();
+      await page.notebook.setCell(0, 'code', '%subshell');
+      await page.notebook.runCell(0);
```
krassowski, line 143: I think these cannot be run using the `notebook` helper, and this is why the tests started failing. Notice how my suggestion was only on lines 121-122 and not on these lines.

krassowski: @ianthomas23 do you need help fixing the CI?

ianthomas23: @krassowski Thanks for the offer. I'm back working today after quite a long break and I plan to return to this over the next day or so. Let me see if can fix it myself, otherwise I will ping you if I need help.

packages/console-extension/src/index.ts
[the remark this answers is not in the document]
ianthomas23: I've wrapped using `sessionContext.kernelChanged` in dfd4fca, and added a `sessionContext.isDisposed` check as we don't want to create a subshell then.

packages/console-extension/src/index.ts
```diff
@@ -405,6 +411,26 @@ async function activateConsole(
       void tracker.save(panel);
     });
 
+    if (options.subshell) {
+      panel.sessionContext.ready
+        .then(async () => {
+          // Ensure kernel has received kernel_info.
+          await panel.sessionContext.session!.kernel!.info;
```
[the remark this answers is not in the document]
ianthomas23: Non-null checks added in 8cad8c53.

galata/test/jupyterlab/kernel.test.ts
ianthomas23: The relevant CI changes are reverted in ab02459.

ianthomas23:
As discussed in comment https://github.com/jupyterlab/jupyterlab/pull/16963#discussion_r1889037890 the subshell functionality in now always available for kernels that support it, and the `--kernel-subshells` flag is changed to `--subshell-console` to enable the subshell console. Hence the core functionality from the JEP is always available but the more experimental subshell console is opt-in.

I have also ensured that the `subshell_id` is always included in all `shell` messages rather than just `execute_request` (it was going to be in a later PR but may as well be included now), and added explicit tests for the three subshell messages (`create`, `delete` and `list`) without the subshell console being required.

ianthomas23:
I think the CI failures are unrelated as similar timeouts occur in other PRs.

krassowski:
Correct, `[jupyterlab] › test/jupyterlab/windowed-notebook.test.ts:84:5 › should not update height when hiding` is a flaky test.

krassowski (approved):
Thank you @ianthomas23!

I personally disagree with making `subshell_console` an opt-in option (maybe I just do not understand the motivation for it), but we can discuss this in a follow-up issue.

jtpio:
Thanks!

pplonski:
This functionality can be used in many Jupyter extensions that provide information about current kernel status. For example, it can be use in variable explorer extension to display current variables in the notebook. Thank you very much for this work :1st_place_medal: :heart:

Issue #16957, opened 2024-11-13
Logging to stderr is significantly slower in Jupyterlab vs IPython


divyansshhh:
## Description

Logging content to stderr is quite slow inside jupyterlab. Consider the following snippet - 

```python
import logging

logger = logging.getLogget()

for i in range(1000):
    logger.error(f"{i} XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
```

This takes only a few milliseconds to execute in an IPython session but takes several seconds to execute in JupyterLab. FWIW, the execution time in the `nbclassic` UI is similar to IPython.

## Reproduce

Run the above snippet in lab and IPython (or even nbclassic)

## Expected behavior

The time taken in JupyterLab should be similar to IPython (or even nbclassic).

## Context

- JupyterLab version: 4.3.0b3

krassowski:
I can reproduce this in all recent versions of JupyterLab. I increased the range to 2000 and used a stopwatch to measure the real time elapsed.
| JupyterLab |  Stopwatch time |
|--|--|
| 3.6.7 |18s  |
| 4.0.12 (RTC enabled) | 21s |
| 4.1.0 |  |
| 4.1.7 | 39s |
| 4.2.5 | 40s |
| 4.3.0b3 | 16s |
| 4.3.0 (RTC enabled) | 18s |
| `main` branch | 17s |
| IPython 8.26 | 1s |

 I also tried to measure the time with `%timeit` magic and the execution status circle timer:

![image](https://github.com/user-attachments/assets/43e10e3a-d9e6-4926-af3e-f9be2236982b)

but both were showing times much lower than the stream took to show up for the user.

The results above highlight that:
- there was a performance regression between 4.0 and 4.1, based on the profiling (to follow in the next comment) it might be related to "open files from errors" feature added in 4.1:
  - https://github.com/jupyterlab/jupyterlab/pull/13390
- there was an improvement in performance between 4.2 and 4.3, this is likely due to:
  - https://github.com/jupyterlab/jupyterlab/pull/16814 and
  - https://github.com/jupyterlab/jupyterlab/pull/16815

krassowski:
Across the number of lines rendered the behaviour fits quadratic dependency:

<img src="https://github.com/user-attachments/assets/d965061f-ebef-49d4-a7d6-50c45de4d409" align="right">

|   lines |   seconds |   quadratic_fit |
|--------:|----------:|----------------:|
[...]

PR #16975, opened 2024-11-19
Fix regression in standard error rendering performance

krassowski:
## References

https://github.com/jupyterlab/jupyterlab/issues/16957

## Code changes

Adds caching for results on the `linkify` method which is the most expensive step when streaming large errors and standard outputs. The second most expensive step is sanitization (parsing of HTML) but that cannot be cached without compromising security.

### Faster stdout

| Iterations | Before | After |
|--|--|--|
| 1000 | 4s | 3s |
| 2000 | 18s | 8s |
| 3000 | 42s | 22s |
| 4000 | 1min 23s | 41s |

```python
import sys

for i in range(4000):
    print(f"example.com {i} XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX www.", end='')
    sys.stdout.flush()
```

### Faster stderr

| Iterations | Before | After |
|--|--|--|
| 1000 | 2s | 1s |
| 2000 | 7s | 3s |
| 3000 | 16s | 7s |
| 4000 | 30s | 13s |

```python
import logging

logger = logging.getLogger()

for i in range(4000):
```

I had to increase `--ServerApp.iopub_msg_rate_limit` to `10000` as the stream was faster than `jupyter-server` was allowing for ("IOPub message rate exceeded").

None


krassowski:
Failing test `test/jupyterlab/notebook-search.test.ts:511:7 › Notebook Search › Search new outputs`might be related. The remaining tests are unrelated and tracked in https://github.com/jupyterlab/jupyterlab/issues/16982

packages/rendermime/src/renderers.ts
```diff
@@ -801,16 +819,57 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
+        } else {
+          linkedNodes = [...fromCache, ...newAdditions];
+        }
+      } else {
+        linkedNodes = autolink(preTextContent, autoLinkOptions);
+      }
+      cacheStore.set(host, {
+        preTextContent,
+        linkedNodes
+      });
```
krassowski, line 869: Here is the problem: when search highlights are applied, they modify the nodes in the DOM by calling [`Text.splitText`](https://developer.mozilla.org/en-US/docs/Web/API/Text/splitText):

https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/packages/documentsearch/src/providers/genericsearchprovider.ts#L317

since the same nodes are in cache, the spliced version ends up inserted next time. If search string covers the entire text then it seems as if nothing was added for subsequent renders where cache is hit.

If we add `.cloneNode()` for each of `linkedNodes` then the performance drops to the same level as before this patch.

krassowski: We can fix this by cloning the node in search highlights code instead, but this would impact the performance of highlighting matches.

Ideally, we would only clone cache entries on Node's modification (well, prior to it). This is not easily possible.

krassowski: Maybe instead of storing the DOM elements in the cache we could serialize the recipes for those; this would be easy because we have only text and anchors to deal with...

krassowski: Looking at it again it seems that there is still a significant performance gain even when calling `.cloneNode()` each time. I will update the benchmaks later.

krassowski: Verified both memory usage and performance, neither is significantly affected

For new lines scenario (standard output) where nodes can be collapsed into a single node the overhead of calling `.cloneNode(true)` is negligible:

image

For the scenario with many nodes (standard output example) the overhead adds up to ~1s over when running:

```python
import sys

for i in range(1000):
    print(f"example.com {i} XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX www.", end='')
    sys.stdout.flush()
```

which took 7 seconds with `.deepClone(true)` as compared to 14 seconds before this PR.

image

Note the timings are different than in the top-level comment as I am testing on a different hardware configuration today.

andrii-i (approved):
Great work @krassowski. Let's update the branch so that it would have CI fix from #16989 so that CI is green before merging.

packages/rendermime/src/renderers.ts
```diff
@@ -801,16 +819,61 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
+      );
+      if (cache) {
+        const { cachedNodes, addedText } = cache;
+        const fromCache = cachedNodes;
+        const newAdditions = autolink(addedText, autoLinkOptions);
+        const lastInCache = fromCache[fromCache.length - 1];
+        const firstNewNode = newAdditions[0];
+
+        if (lastInCache instanceof Text && firstNewNode instanceof Text) {
+          const joiningNode = lastInCache;
```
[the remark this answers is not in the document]
krassowski: Yes, self-documenting code principle - instead of a comment which would say "we use the last node in cache as a joining node". The idea is that the ts compiler will optimize this out anyways so assignments are cheap.

packages/rendermime/src/renderers.ts
```diff
@@ -854,6 +914,67 @@ export namespace renderText {
...
+    // continue
+  } else if (lastCachedNode instanceof Text) {
+    // Remove the Text node to re-analyse this text.
+    // This is required when we cached `aaa www.one.com bbb www.`
+    // and the incoming addition is `two.com`. We can still
+    // use text node `aaa ` and anchor node `www.one.com`, but
+    // we need to pass `bbb www.` + `two.com` through linkify again.
+    cachedNodes = cachedNodes.slice(0, -1);
+    addedText = lastCachedNode.textContent + addedText;
+    // continue
```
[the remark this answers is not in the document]
krassowski: Oh, the `// continue` here is just a comment to indicate that we are not short-circuiting (like the `else` clause is). I see why it might be confusing. I will reword it for clarity.

packages/rendermime/src/renderers.ts
```diff
@@ -801,16 +819,61 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
-        : [document.createTextNode(content)];
+    let linkedNodes: (HTMLAnchorElement | Text)[];
+    if (sanitizer.getAutolink?.() ?? true) {
+      const cache = getApplicableLinkCache(
+        cacheStore.get(host),
+        preTextContent
+      );
+      if (cache) {
+        const { cachedNodes, addedText } = cache;
+        const fromCache = cachedNodes;
```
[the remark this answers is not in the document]
krassowski: The right syntax for renaming when deconstructing would be:

```suggestion
        const { cachedNodes: fromCache, addedText } = cache;
```

again, it comes down to the matter of preference because it should be optimized out to the same code on compiler level. I tend to avoid that syntax in this codebase because it is not super-obvious to folks less familiar with TS.

packages/rendermime/src/renderers.ts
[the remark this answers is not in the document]
krassowski: For context, the linter does not allow empty blocks like `if (something) {}` so we need to explain why the block is empty. Typically this is done with `// no-op` comment but more explanation is usually better than just "no-op":

```ts
if (something) {
  // no-op
}
```

packages/rendermime/src/renderers.ts
```diff
@@ -801,16 +819,60 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
-          })
-        : [document.createTextNode(content)];
+    let linkedNodes: (HTMLAnchorElement | Text)[];
+    if (sanitizer.getAutolink?.() ?? true) {
+      const cache = getApplicableLinkCache(
+        cacheStore.get(host),
+        preTextContent
+      );
+      if (cache) {
+        const { cachedNodes: fromCache, addedText } = cache;
```
Carreau, line 846: ah ! +1, not used to destructuring in ts.

packages/rendermime/test/factories.spec.ts
```diff
@@ -570,6 +570,162 @@ describe('rendermime/factories', () => {
...
+          const after = w.node.innerHTML;
+          const linkAfter = w.node.querySelector('a');
+
+          // The contents of the node should be updated with the new line.
+          expect(before).not.toEqual(after);
+          expect(after).toContain('line of text');
+
+          // If cached links were used, the anchor nodes will be reused.
+          expect(cachedLink).not.toBe(null);
+          expect(cachedLink).toBe(w.node.childNodes[0].childNodes[0]);
```
krassowski, line 654: So the problem is that with the fix in the last commit we cannot use identity for testing if cache was used. We would need to somehow mark the nodes as coming from cache for the testing purposes...

krassowski: Fixed in 1aaebbd1d5de5d176dd7d3b58e6958cd76450cab.

krassowski:
All green, merging!

krassowski:

PR #16979, opened 2024-11-20
Restore viewport `min-height` when not windowing

brichet:
This PR restore the `min-height` property of the windowed viewport DOM element when not windowing.

The `min-height` is null if jupyterlab starts without windowing.
When the windowing is set to `full`, this property is set to ensure the minimal size of the viewport.

But this property is not restored to null when switching back to 'no windowing'. This can lead to wrong display with extensions that reorganize the cells, like [jupyterlab-deck](https://github.com/deathbeds/jupyterlab-deck).

## User-facing changes

Should not in lab

None


krassowski (approved):
Thank you @brichet! Test failures appear unrelated, I also saw them on other PRs and opened https://github.com/jupyterlab/jupyterlab/issues/16982.

krassowski:

Issue #16982, opened 2024-11-21
`Debugger Variables Copy` visual regression tests do not pass with Python 3.12


krassowski:
```
Notice:   3 failed
    [jupyterlab] › test/jupyterlab/debugger.test.ts:192:7 › Debugger Variables › Copy to globals should work only for local variables 
    [jupyterlab] › test/jupyterlab/debugger.test.ts:252:7 › Debugger Variables › Copy to globals not available from kernel 
    [jupyterlab] › test/jupyterlab/debugger.test.ts:289:7 › Debugger Variables › Copy to clipboard ─
```

krassowski:
This is evidenced by these tests failing in unrelated PRs:
- https://github.com/jupyterlab/jupyterlab/pull/16975
- https://github.com/jupyterlab/jupyterlab/pull/16979

These tests were passing on Wed, 20 Nov 2024 09:34:33 GMT and failing on Wed, 20 Nov 2024 13:46:10 GMT

krassowski:
Of note `maintainer-tools` [0.26.0](https://github.com/jupyterlab/maintainer-tools/releases/tag/v0.26.0) was released on Wed, 20 Nov 2024 12:48 GMT. It included the bump of the standard Python version to 3.12.

krassowski:
I updated the titled to reflect that the tests are now passing on `main` but we still need to figure out what component in Python 3.12 causes them to fail and report to appropriate upstream.

jtpio:
Thanks for hot-fixing this! Yeah it would be great to know more about what is happening with 3.12.

PR #16988, opened 2024-11-22
Fix FileBrowserModel.onFileChanged

davidbrochart:
## References

Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/406.

## Code changes

When an operation is made on a file, we emit a signal if the file is in the current directory, which may trigger a refresh of the directory content. For instance, if a file is created, it immediately appears in the file browser, instead of after the next poll.
To check if the file is in the current directory, we compare the current file browser path with the path of the file. Both may have a drive prefix (e.g. `RTC:`). To get the directory of the file, we use `PathExt.dirname()` which seems to strip the drive prefix. But the current file browser path has the drive prefix, so the outcome is that we always think that the file is not in the current directory.
This PR removes the drive prefix of the current file browser path when making this comparison.

## User-facing changes

Operations such as creating a file or uploading a file look immediate in the file browser.

None.

### DIFF
--- a/packages/filebrowser/src/model.ts
+++ b/packages/filebrowser/src/model.ts
@@ -606,7 +606,7 @@ export class FileBrowserModel implements IDisposable {
     sender: Contents.IManager,
     change: Contents.IChangedArgs
   ): void {
-    const path = this._model.path;
+    const path = this._model.path.slice(this.driveName.length + 1);
     const { sessions } = this.manager.services;
     const { oldValue, newValue } = change;
     const value =


JasonWeill:
Are the UI tests relevant?

Also, is there an existing test suite that might be improved by testing this logic?

davidbrochart:
UI tests fail on `main` too, it's hard to see if any failure is caused by this PR.

krassowski:
The failed tests are unrelated and tracked in https://github.com/jupyterlab/jupyterlab/issues/16982.

This change looks good on the surface, but I wonder about false positives in two drives have directories with the same name. Could we possibly have an argument to `PathExt.dirname` saying whether to strip the drive prefix or not?

Also, I second the above comment about adding a unit test. It looks like these could be added here:

https://github.com/jupyterlab/jupyterlab/blob/b72b865d0af26a396268e826522b5680d215cd99/packages/filebrowser/test/model.spec.ts#L141-L180

krassowski (approved):
That's even better! Thank you @davidbrochart!

CI failures are unrelated. There is a PR waiting for review to fix one of them (https://github.com/jupyterlab/jupyterlab/pull/16989) and another is tracked in https://github.com/jupyter-server/jupyter_releaser/issues/595

claytonparnell:

krassowski:

PR #16989, opened 2024-11-24
Pin Python version for visual regression testing to 3.11

krassowski:
## References

Workarounds https://github.com/jupyterlab/jupyterlab/issues/16982 which was caused by a change in new `maintainer-tools` release.

Of note the test is failing in cleanup step, not in the actual test step so pinning the Python version for now seems warranted.

Tagging it for backporting two versions back as the change in `maintainer-tools` will affect all branches.

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/galata.yml
+++ b/.github/workflows/galata.yml
@@ -23,6 +23,8 @@ jobs:
 
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+        with:
+          python_version: "3.11"
 
       - name: Set up browser cache
         uses: actions/cache@v4


andrii-i (approved):
This looks good and fixes the CI problem (example of the faulty run: https://github.com/jupyterlab/jupyterlab/actions/runs/12038662055/job/33564819633?pr=17000). 

Alternatively Python version could be pinned to a range `>=3.8,<3.12` instead of 3.11 only but I don't see strong arguments why it would be better taking in account this is only for CI.

Thank you for working on this @krassowski.

krassowski:
This CI action only runs one Python version so it only accepts a concrete specifier

PR #16994, opened 2024-11-25
Add Settings Import Feature from a JSON File

Darshan808:
## Description 

The feature to export user-defined settings to an `overrides.json` file was introduced in https://github.com/jupyterlab/jupyterlab/pull/16896. This PR builds upon that functionality by adding the ability to import settings from a JSON file.

## Feature Overview

With this feature, users can:
- Select a JSON file.
-  Choose the settings that should be applied.
- Import the settings specified into the application.

### Demo Video

https://github.com/user-attachments/assets/1deb4cf8-7a5d-492f-b73a-27ee6546ee10

## Work in Progress

This is an initial implementation, and some areas need further discussion and refinement. Feedback and suggestions are highly appreciated. 

I am open to writing tests once the feature is finalized.

Thank you for your time and feedback!


Darshan808:
@jtpio @krassowski 
Could you please take a moment to review this? Your feedback would be greatly appreciated.

krassowski:
This deserves more visibility in the changelog, bumping label to `feature`.

packages/settingeditor-extension/src/index.ts
```diff
@@ -384,6 +398,87 @@ function activateJSON(
...
+          };
+
+          const settingsKeys = Object.keys(importedSettings);
+          const content = new ImportSettingsWidget(
+            settingsKeys,
+            applySettings,
+            translator
+          );
+          const widget = new MainAreaWidget<ImportSettingsWidget>({ content });
+          widget.title.label = trans.__('Imported Settings');
```
krassowski, line 461: I think this would be better in UI:

```suggestion
          widget.title.label = trans.__('Import Settings');
```

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const handleCheckboxChange = (key: string, isChecked: boolean) => {
+    const updatedStates = { ...checkedStates, [key]: isChecked };
+    setCheckedStates(updatedStates);
+  };
+
+  return (
+    <div className="jp-SettingsImport-container">
+      <div className="jp-SettingsImport-header">
+        <span className="jp-SettingsImport-title">
+          {props.trans.__('The following settings will be affected')}
```
krassowski, line 34: It might be better to use call-to-action here?

```suggestion
          {props.trans.__('Select settings sections to import')}
```

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Constructs a new ImportSettingsWidget.
+   *
+   * @param importedSettings - The settings to display.
+   * @param handleImport - A callback for handling imports.
+   */
+  constructor(
+    importedSettings: string[],
+    handleImport: (settings: string[]) => void,
+    translator: ITranslator
```
krassowski, line 77: Nitpick: using the `IOptions` pattern would be nice here for consistency. Not critical as it is not exported publicly, only locally.

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { ITranslator, TranslationBundle } from '@jupyterlab/translation';
+import { ReactWidget } from '@jupyterlab/ui-components';
+import React, { useState } from 'react';
+
+const Settings = (props: {
```
krassowski, line 10: Maybe:

```suggestion
const SettingsImport = (props: {
```

just to get less confusion down the line when searching for "Settings" across the entire codebase

krassowski (commented):
Thank you for working on this! Just skimming the code right now, it looks good - leaving some minor suggestions.

I wonder if we should shorten the labels to just "Export"/"Import" to avoid overflowing the toolbar on smaller screens.

Note to self: check how the "Import" button behaves on smaller screens and whether it could be better placed elsewhere.

Darshan808:
@krassowski 
Thank you for the feedback. I've made the suggested changes. For the UI, I was exploring and came across this approach: having a button that, when clicked, presents options for importing and exporting. Something like this:
settings

What do you think? I’d appreciate your feedback

Darshan808:
> I am getting a console error while the settings are being applied image
> 
> I believe JupyterLab refreshes itself each time a setting value is changed, which is causing multiple refreshes. I'm unsure if the approach I implemented is the right one.

Also, could you help me on this?

krassowski:
I think this would not be as intuitive as having two separate buttons with short labels. Also the up/down icon looks a lot like data transfer icon. Finally, adding a dropdown on toolbar might introduce some accessibility concerns.

krassowski:
I can look into that error later in the week, if you do not hear from me by Saturday please feel free to ping me to remind me

Darshan808:
The error does not occur when I add a simpler version like this:

```
{
  "@jupyterlab/apputils-extension:themes": {
    "theme": "JupyterLab Dark"
  }
}
```

I encountered an issue while adding the following JSON settings, which I obtained by exporting settings after enabling the dark theme:
```
{
  "@jupyterlab/apputils-extension:themes": {
    "adaptive-theme": false,
    "overrides": {
      "code-font-family": null,
      "code-font-size": null,
      "content-font-family": null,
      "content-font-size1": null,
      "ui-font-family": null,
      "ui-font-size1": null
    },
    "preferred-dark-theme": "JupyterLab Dark",
    "preferred-light-theme": "JupyterLab Light",
    "theme": "JupyterLab Dark",
    "theme-scrollbars": false
  }
}
```

Darshan808:
I've added a test for the feature; any feedback or suggestions are welcome!

Darshan808:
While exploring the registry interface, I discovered that the `upload` method is a faster and more efficient way to handle bulk settings updates compared to using `set` for individual key-value changes.
I’ve updated the function to leverage `upload`, which has resolved the console errors and significantly improved the import speed.
I’d love to hear your thoughts and feedback on this!

Darshan808:
> I can look into that error later in the week, if you do not hear from me by Saturday please feel free to ping me to remind me

Pinging you to remind about the follow-up!

Darshan808:
> If you do not hear from me by Saturday please feel free to ping me to remind me

@krassowski 
Pinging you to remind about the follow-up!

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { ITranslator } from '@jupyterlab/translation';
+import { ReactWidget } from '@jupyterlab/ui-components';
+import React, { useState } from 'react';
```
jtpio, line 8: According to the CI failure, we may need to explicitly add the `react` dependency to the `package.json`: https://github.com/jupyterlab/jupyterlab/actions/runs/12197606507/job/34027611113?pr=16994

Darshan808: Thanks for pointing that out! I just added it.

Darshan808:
I think the failing tests now are unrelated to this PR.

Darshan808:
I think the failing tests now are unrelated to this PR.

jtpio:
> I think the failing tests now are unrelated to this PR.

Indeed they seem to be passing after a restart.

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  return (
+    <div className="jp-SettingsImport-container">
+      <div className="jp-SettingsImport-header">
+        <span className="jp-SettingsImport-title">
+          {trans.__('Select settings sections to import')}
+        </span>
+        <button
+          className="jp-mod-styled jp-mod-reject jp-ArrayOperationsButton"
```
krassowski, line 48: 1. We should not use `jp-mod-reject` but `jp-mod-accept` here as it is a constructive action. This means the button would be blue by default.
2. We should not add `jp-ArrayOperationsButton` this is a very specific class meant for the settings form with arrays.

```suggestion
          className="jp-mod-styled jp-mod-accept"
```

FYI in future we will probably be moving towards `import { Button } from '@jupyter/react-components';` but I think it is fine to use a plain button here for now.

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <div className="jp-SettingsImport-list">
+        {props.importedSettings.map(key => (
+          <div key={key} className="jp-SettingsImport-item">
+            <span className="jp-SettingsImport-itemKey">{key}</span>
+            <input
+              type="checkbox"
+              checked={checkedStates[key]}
+              onChange={e => handleCheckboxChange(key, e.target.checked)}
+              className="jp-SettingsImport-checkbox"
+            />
```
krassowski, line 67: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#providing_a_bigger_hit_area_for_your_checkboxes

```suggestion
            <label>
              <span className="jp-SettingsImport-itemKey">{key}</span>
              <input
                type="checkbox"
                checked={checkedStates[key]}
                onChange={e => handleCheckboxChange(key, e.target.checked)}
                className="jp-SettingsImport-checkbox"
              />
            </label>
```

packages/settingeditor-extension/src/importSettingsWidget.tsx
krassowski: It would be nice if this button had `cursor: pointer` too.

packages/settingeditor-extension/src/index.ts
```diff
@@ -384,6 +397,76 @@ function activateJSON(
...
+                  `Invalid settings for plugin ${pluginId}. Skipping.`
+                );
+              }
+            }
+          };
+
+          const settingsKeys = Object.keys(importedSettings);
+          const content = new ImportSettingsWidget({
+            importedSettings: settingsKeys,
+            handleImport: applySettings,
```
krassowski, line 445: There is no feedback after clicking this button right now, is it correct? I think we should:
- show a [dialog](https://jupyterlab.readthedocs.io/en/stable/extension/ui_helpers.html#generic-dialog) saying "Settings imported" confirming how many settings were imported
- close the `ImportSettingsWidget`

packages/settingeditor-extension/src/index.ts
```diff
@@ -384,6 +397,76 @@ function activateJSON(
...
+          widget.title.icon = fileUploadIcon;
+          app.shell.add(widget, 'main');
+          app.shell.activateById(widget.id);
+        } catch (error) {
+          console.error('Failed to import settings:', error);
+          alert(
+            trans.__(
+              'Error importing settings. Please ensure the file is valid.'
+            )
+          );
```
krassowski, line 459: Let's use [`showErrorMessage()`](https://jupyterlab.readthedocs.io/en/latest/api/functions/apputils.showErrorMessage.html) here (also see [message dialogs](https://jupyterlab.readthedocs.io/en/stable/extension/ui_helpers.html#message-dialogs)).

packages/settingeditor/style/base.css
```diff
@@ -285,3 +285,47 @@
...
+  align-items: center;
+}
+
+.jp-SettingsImport-title {
+  font-size: 1rem;
+  font-weight: bold;
+  padding: 10px 15px;
+}
+
+.jp-SettingsImport-importButton:hover {
```
krassowski, line 306: I do not see an element with `jp-SettingsImport-importButton` class.

Darshan808:
```
{
    "@jupyterlab/apputils-extension:themes": {
      "adaptive-theme": false,
      "theme": "JupyterLab Dark",
      "theme-scrollbars": false
    },
  "@jupyterlab/apputils-extension:sanitizer": {
      "autolink": true,
      "allowNamedProperties": true
    },
  }
  ```
  In settings import like these, should we say `2 settings categories successfully imported.` or `imported 5 settings across 2 categories successfully.` or just `imported settings across 2 categories successfully.` ? 
 Or can we write something better in dialog box's body?

**Comment by Darshan808** [comment]
since we aren't showing individual settings in UI, I think `imported settings across 2 categories successfully.` would be better.

File: packages/settingeditor-extension/src/index.ts [hunk 11]
```diff
@@ -384,6 +400,78 @@ function activateJSON(
...
+                console.warn(
+                  `Invalid settings for plugin ${pluginId}. Skipping.`
+                );
+              }
+            }
+            app.shell.currentWidget?.close();
+            if (settingsEntries.length) {
+              await showDialog({
+                title: 'Settings Imported',
+                body: `imported settings across ${settingsEntries.length} categories successfully."`,
```
krassowski, line 446:
```suggestion
                body: `Imported settings across ${settingsEntries.length} categories successfully.`,
```

image

packages/settingeditor-extension/src/index.ts
```diff
@@ -384,6 +400,78 @@ function activateJSON(
...
+
+          const applySettings = async (settings: string[]) => {
+            // Apply settings to the registry
+            const settingsEntries = Object.entries(importedSettings);
+            for (const [pluginId, pluginSettings] of settingsEntries) {
+              if (
+                typeof pluginSettings === 'object' &&
+                !Array.isArray(pluginSettings)
+              ) {
+                await registry.upload(pluginId, JSON.stringify(pluginSettings));
```
krassowski, line 435: This fails badly if the `pluginId` is unknown:

```
Uncaught (in promise): Schema not found: /home/jovyan/dev_mode/schemas/@jupyterlab/this-plugin-name-does-not-exist/plugin.json
```

To test it I changed the plugin name in `overrides.json`. I would suggest adding a granular `try-catch` in addition to the one you have around the entire block. Of note, your outer `try-catch` does not catch this error because it is raised in a callback to `onClick` function after the `try` block was executed (i.e. the widget was created).

krassowski: By "badly" I mean without informing the user about the error, and importing only the part of the settings which were before the problematic plugin entry.

I think we can expect this to happen as users will be importing settings from installations which included extensions that they no longer have installed.

Darshan808: I thought of resolving this by tracking the plugins whose settings encountered errors during import and then displaying a dialog with the message: "Failed to upload settings for the following IDs:"

krassowski: Yes that would be ideal

Darshan808:
The standard text insertion into the dialog box body wasn't providing the desired formatting, so I opted to create a React element for the dialog box content. Could you review this approach?

Screenshot from 2024-12-13 21-57-18

packages/settingeditor-extension/src/importSettingsWidget.tsx
```diff
@@ -0,0 +1,158 @@
...
+        successMessage={this.successMessage}
+        failureMessage={this.failureMessage}
+        failedSettings={this.failedSettings}
+      />
+    );
+  }
+
+  private successMessage: string;
+  private failureMessage?: string;
+  private failedSettings?: string[];
```
krassowski, line 157: This mostly works, but I would suggest to either:
- a) do away with the functional `ImportSettingsDialogBody` component, or
- b) make this be a bit more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) (see code suggestion below).

I would personally go for (b) as functional components allow to use hooks etc if you wanted to expand the body functionality in the future.

```suggestion
  constructor(private _props: ImportSettings.IDialogBodyBodyOptions) {
    super();
  }

  render(): JSX.Element {
    return <ImportSettingsDialogBody {...this._props} />;
  }
```

packages/settingeditor-extension/src/index.ts
```diff
@@ -384,6 +403,111 @@ function activateJSON(
...
+            const settingsEntries = Object.entries(importedSettings);
+            for (const [pluginId, pluginSettings] of settingsEntries) {
+              if (
+                typeof pluginSettings === 'object' &&
+                !Array.isArray(pluginSettings)
+              ) {
+                try {
+                  await registry.upload(
+                    pluginId,
+                    JSON.stringify(pluginSettings)
```
krassowski, line 440: Imported settings are hard to edit in JSON Settings Editor:

image

When editing settings from the form, 4 spaces of indentation and comments are added. I think this PR does not need to go as far as adding the comments, but adding four spaces indentation is easy enough with:

```suggestion
                    JSON.stringify(pluginSettings, null, 4)
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#using_the_space_parameter

The second argument can be either `null` or `undefined`. The third one would ideally be a constant with a descriptive name, like in:

https://github.com/jupyterlab/jupyterlab/blob/4324f186d289715e1b7c634ee6d577bd67e853ed/packages/settingeditor/src/SettingsFormEditor.tsx#L166

Darshan808:
I've also identified

Darshan808:
I came across another issue!
Even after importing and applying settings, the values in the settings editor remain unchanged. For example, if the imported settings specify a dark theme, the editor still shows a light theme until refreshed.
Screenshot from 2024-12-16 19-42-58

Looking into it

Darshan808:
The same happens when saving settings via the JSON settings editor. Should this be addressed in a separate PR?

krassowski:
> The same happens when saving settings via the JSON settings editor. Should this be addressed in a separate PR?


Yes, as we would probably want to backport the fix (if it is relatively self-contained). Can you open an issue describing it briefly?

Darshan808:
The failing test case appears to be unrelated!

> Even after importing and applying settings, the values in the settings editor remain unchanged.

This issue will likely be resolved by https://github.com/jupyterlab/jupyterlab/pull/17091

krassowski (approved):
Thank you @Darshan808!

PR #16999, opened 2024-11-26
Fix handling of carriage return in output streams

### LINKED ISSUE #16995: carriage return (\r) in print fails when there is a delay between subsequent print calls
## Description

When using carriage return (\r) at the end of a print command so that the next print will overwrite what has been printed fails to return the cursor to that start of the line when there is a delay between print calls.
When the delay is greater than 0.2 seconds (using time.sleep) the cursor simply remains at the end of the line and following prints are appended.
When the delay is less than 0.2 seconds but more than 0.02 seconds the cursor sometimes returns to the start and sometimes does not.
When the delay is less than 0.02 seconds the cursor consistently returns to the start of the line and expected behaviour occurs.

As an example consider the snippet:
```
from time import sleep
for i in range(10):
    print(f"{i}", end="\r")
    sleep(x)
```

The expected output after running is quickly counting up to 9, with a single 9 remeining in the output when the cell has finished.
This only occurs for x < 0.02
For 0.2 > x > 0.02 some but not all number will be appended, for example for x = 0.1 I get 8369 as final output.
For x > 0.2 all numpbers will simply be appended and the output is 0123456789.

This is new behaviour in the last 6 months and has "broken" several progress trackers of mine.
The behaviour does not occur in iPython run directly in the command line.
image

## Reproduce

Run the following code in a cell varying x from about 0.02 to 0.2.
```
from time import sleep
for i in range(10):
    print(f"{i}", end="\r")
    sleep(x)
```

Final output will vary from 9 for x < 0.02 to 0123456789 for x > 0.2 with varying behaviour in between.

## Expected behavior

The final output is 9 regardless of x.
## Context

- Operating System and version:  Windows 11 23H2
- Browser and version:  Firefox 23H2
- JupyterLab version:  4.3.1

<pre>
</pre>
</details>

<pre>
</pre>
</details>

<pre>

</pre>
</details>

davidbrochart:
## References

Fixes #16995.

## Code changes

We need to keep track of the cursor position.

## User-facing changes

Carriage return is properly handled even for fast print statements.

None.

### DIFF
--- a/packages/outputarea/src/model.ts
+++ b/packages/outputarea/src/model.ts
@@ -366,7 +366,10 @@ export class OutputAreaModel implements IOutputAreaModel {
       if (typeof value.text !== 'string') {
         value.text = value.text.join('');
       }
-      const {text, index} = Private.processText(this._streamIndex, value.text);
+      const { text, index } = Private.processText(
+        this._streamIndex,
+        value.text
+      );
       this._streamIndex = index;
       value.text = text;
     }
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you @davidbrochart, tested on Binder confirming it fixes the issue!

Failing tests are unrelated, tracked in https://github.com/jupyterlab/jupyterlab/issues/16982 and https://github.com/jupyter-server/jupyter_releaser/issues/595

krassowski:

PR #17000, opened 2024-11-26
Fix disappearing cells (heal offsets after updating estimated sizes)

### LINKED ISSUE #16978: Disappearing cells regression in JupyterLab 4.3.1
## Description

This was reported in Jupyter Notebook - https://github.com/jupyter/notebook/issues/7523. I narrowed it down based on logs to two issues:
- notebook rendering crashes due to uncaught bad offsets regression
- scroll offset is not correctly set after some operations (to be determined)

## Reproduce

TBD. So far I understand that this affects notebooks:
- with cells which have very long outputs (this is considerably longer larger than viewport height)
- at least as many cells to fit the viewport a few times
- with `full` windowing mode enabled

## Expected behavior

- Windowing never breaks render even if the algorithms encounters an edge case/race condition
- scroll offsets are always correctly populated

## Context

- Browser and version: All browsers are affected
- JupyterLab version: 4.3.1

krassowski:
## References

Fixes #16978

## Code changes

- Do not set `_measuredAllUntilIndex` as a side effect in `_getItemMetadata` - this is wrong as we are not actually measuring sizes there, just estimating
- Update offsets in `_getItemMetadata` after changing values in `_widgetSizers` by populating estimates sizes (this is a side effect which I think should not in this getter, but this PR is a hotfix and it is not feasible to write unit tests for these scenarios)

## User-facing changes

Cells no longer disappear after copy-paste/dragging.

None

### DIFF
--- a/galata/test/jupyterlab/windowed-notebook.test.ts
+++ b/galata/test/jupyterlab/windowed-notebook.test.ts
@@ -406,12 +406,12 @@ test('should display cells below on scrolling after inserting a cell on top', as
   // https://github.com/jupyterlab/jupyterlab/issues/16978
   await page.notebook.openByPath(`${tmpPath}/${fileName}`);
 
-  const h = await page.notebook.getNotebookInPanelLocator();
-  const firstCell = h!.locator('.jp-Cell[data-windowed-list-index="1"]');
-  const lastCell = h!.locator('.jp-Cell[data-windowed-list-index="18"]');
+  const notebook = await page.notebook.getNotebookInPanelLocator()!;
+  const firstCell = notebook.locator('.jp-Cell[data-windowed-list-index="1"]');
+  const lastCell = notebook.locator('.jp-Cell[data-windowed-list-index="18"]');
   await firstCell.waitFor();
 
-  const bbox = await h!.boundingBox();
+  const bbox = await notebook.boundingBox();
   await page.mouse.move(bbox!.x, bbox!.y);
 
   // Needs to be two separate mouse wheel events.
--- a/galata/test/jupyterlab/windowed-notebook.test.ts
+++ b/galata/test/jupyterlab/windowed-notebook.test.ts
@@ -398,6 +398,55 @@ test('should remove all cells including hidden outputs artifacts', async ({
   expect(found).toEqual(false);
 });
 
+test('should display cells below on scrolling after inserting a cell on top', async ({
+  page,
+  tmpPath
+}) => {
+  // Regression test against "disappearing cells" issue:
+  // https://github.com/jupyterlab/jupyterlab/issues/16978
+  await page.notebook.openByPath(`${tmpPath}/${fileName}`);
+
+  const h = await page.notebook.getNotebookInPanelLocator();
+  const firstCell = h!.locator('.jp-Cell[data-windowed-list-index="1"]');
+  const lastCell = h!.locator('.jp-Cell[data-windowed-list-index="18"]');
+  await firstCell.waitFor();
+
+  const bbox = await h!.boundingBox();
+  await page.mouse.move(bbox!.x, bbox!.y);
+
+  // Needs to be two separate mouse wheel events.
+  await page.mouse.wheel(0, 3000);
+  await page.mouse.wheel(0, 3000);
+
+  // Scroll down to reveal last cell to ensure these all items have been measured...
+  await Promise.all([
+    firstCell.waitFor({ state: 'hidden' }),
+    lastCell.waitFor()
+  ]);
+
+  await page.mouse.wheel(0, -3000);
+  await page.mouse.wheel(0, -3000);
+
+  // ...then scroll back up and select first cell.
+  await Promise.all([
+    lastCell.waitFor({ state: 'hidden' }),
+    firstCell.waitFor()
+  ]);
+  await page.notebook.selectCells(0);
+
+  // Insert cell below.
+  await page.keyboard.press('b');
+  await page.mouse.wheel(0, 3000);
+  await page.mouse.wheel(0, 3000);
+
+  // Scroll down again.
+  await Promise.all([
+    firstCell.waitFor({ state: 'hidden' }),
+    lastCell.waitFor()
+  ]);
+  await expect(lastCell).toBeVisible();
+});
+
 test('should center on next cell after rendering markdown cell and advancing', async ({
   page,
   tmpPath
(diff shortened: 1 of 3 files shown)


krassowski:
Tests failures are related, but surprising:

```js
      TypeError: Cannot read properties of null (reading 'length')
  
        60 |     let outputsLines = 0;
        61 |     if (model instanceof CodeCellModel) {
      > 62 |       for (let outputIdx = 0; outputIdx < model.outputs.length; outputIdx++) {
           |                                                         ^
        63 |         const output = model.outputs.get(outputIdx);
        64 |         const data = output.data['text/plain'];
        65 |         if (typeof data === 'string') {
  
        at NotebookViewModel.length [as estimateWidgetSize] (src/windowing.ts:62:57)
        at NotebookViewModel.estimateWidgetSize [as _getItemMetadata] (../ui-components/src/components/windowedlist.ts:656:18)
        at NotebookViewModel._getItemMetadata [as _findNearestItemExponentialSearch] (../ui-components/src/components/windowedlist.ts:749:12)
        at NotebookViewModel._findNearestItemExponentialSearch [as _findNearestItem] (../ui-components/src/components/windowedlist.ts:709:19)
        at NotebookViewModel._findNearestItem [as _getStartIndexForOffset] (../ui-components/src/components/windowedlist.ts:787:17)
        at NotebookViewModel._getStartIndexForOffset [as _getRangeToRender] (../ui-components/src/components/windowedlist.ts:769:29)
        at NotebookViewModel._getRangeToRender [as getRangeToRender] (../ui-components/src/components/windowedlist.ts:464:29)
        at Notebook.getRangeToRender [as _update] (../ui-components/src/components/windowedlist.ts:1354:43)
        at _update (../ui-components/src/components/windowedlist.ts:1227:16)
        at invokeTheCallbackFunction (../../node_modules/jsdom/lib/jsdom/living/generated/Function.js:19:26)
        at runAnimationFrameCallbacks (../../node_modules/jsdom/lib/jsdom/browser/Window.js:603:13)
        at Timeout.<anonymous> (../../node_modules/jsdom/lib/jsdom/browser/Window.js:581:11)
```

krassowski:
Huh, so this is because of invalid assignment in `dispose()` of `CodeCellModel`:

https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/packages/cells/src/model.ts#L709

The consequence of this PR is that logic estimating widget size is actually invoked each time rather than caching it and pretending it was measured (which was causing problems before). This unveiled some cases where the async rendering logic can bump into disposed widgets. Fixing the estimation to avoid touching disposed objects...

krassowski:
Thank you Rosio. This confirms that the part where an error is thrown an rendering is interrupted is fixed, but suggests that for very long cells which use estimated heights the rendering range may be incorrect.

krassowski:
I am unable to reproduce this reliably locally. It seems to be a Heisenbug.

I think I caught it once and then the debugging was hinting at something unrelated to this PR, namely the `scrollOffset` being set incorrectly (thus the rendering range was not updated correctly). I suspect that this might be an unrelated issue.

Ideally we would be able to write a galata test reproducing the problematic behaviour.

krassowski:
I added test as a first commit in 7f1c3fd33560cbfcf930ba39193d690960e54722 and it is failing as expected:

image

I pushed the fixes back now and the CI should now pass again.

krassowski:
All green!

While I appreciate that there is an edge case where it is broken for cells with very long outputs) as found by @RRosio, the added test demonstrates that this PR constitutes as improvement (and I believe tackles the majority use case) and user testing indicates that this PR fixes the issue as per https://github.com/jupyterlab/jupyterlab/issues/16978#issuecomment-2509796088 and https://github.com/jupyterlab/jupyterlab/issues/16978#issuecomment-2509802897. I would recommend merging this in and releasing as a hot fix. Does anyone have capacity to review?

galata/test/jupyterlab/windowed-notebook.test.ts
```diff
@@ -398,6 +398,55 @@ test('should remove all cells including hidden outputs artifacts', async ({
...
 
+test('should display cells below on scrolling after inserting a cell on top', async ({
+  page,
+  tmpPath
+}) => {
+  // Regression test against "disappearing cells" issue:
+  // https://github.com/jupyterlab/jupyterlab/issues/16978
+  await page.notebook.openByPath(`${tmpPath}/${fileName}`);
+
+  const h = await page.notebook.getNotebookInPanelLocator();
```
andrii-i, line 409: Nitpick for your consideration, not a hard ask: could this variable have a more descriptive name or abbreviation? For example, maybe more letters of the word that `h` represents? Options I can come up from top of my mind are all pretty verbose: `panelNotebook`, `panelNbLocator`, `panelNb`

andrii-i (approved):
Thank you for working on this @krassowski. This is a clear improvement that fixes disappearing cells issue (also confirmed by users). Let's track "very long cells height estimation" issue [idenitified](https://github.com/jupyterlab/jupyterlab/pull/17000#issuecomment-2504498174) by @RRosio in a separate issue. 

Please also find nitpick suggestion for your consideration (not a hard ask) below.

galata/test/jupyterlab/windowed-notebook.test.ts
krassowski: That's a good point. I just used whatever was used in the test case above, but I do prefer a longer name. It looks that other names in this file are `notebook` is used 2 times and `h` is used 8 times to denote the result of this method.

I think `notebook` is just fine but it was also my PR which defied this convention.

galata/test/jupyterlab/windowed-notebook.test.ts
```diff
@@ -398,6 +398,55 @@ test('should remove all cells including hidden outputs artifacts', async ({
...
+  const firstCell = h!.locator('.jp-Cell[data-windowed-list-index="1"]');
+  const lastCell = h!.locator('.jp-Cell[data-windowed-list-index="18"]');
```
krassowski, line 411:
```suggestion
  const h = await page.notebook.getNotebookInPanelLocator();
  const firstCell = h!.locator('.jp-Cell[data-windowed-list-index="1"]');
  const lastCell = h!.locator('.jp-Cell[data-windowed-list-index="18"]');
```

galata/test/jupyterlab/windowed-notebook.test.ts
krassowski: Addressed in e6822d10820ff4b1ea67460f5904bd0321a97148

krassowski:
> Let's track "very long cells height estimation" issue [...]  a separate issue.

I opened https://github.com/jupyterlab/jupyterlab/issues/17023 to track it

krassowski:

Issue #16928, opened 2024-11-06
Missed first keystroke on Ctrl+F when switching between notebook tabs


miloaissatu:
## Description

The first keystroke is missed when typing in an existing/previously opened search box after switching focus to it via keyboard shortcut Ctrl+F.

Examples:
1. When switching between notebook tabs and using Ctrl+F to bring up or switch focus to the search bo
[jupyter_troubleshoot.txt](https://github.com/user-attachments/files/17639963/jupyter_troubleshoot.txt)
x, the first keystroke is often missed (easy to reproduce).
2. Within the same notebook tab where an existing searchbox is open, switching focus between cells and search box via Ctrl+F shortcut, sometimes the first keystroke is missed. (inconsistent, harder to reproduce)

## Reproduce

1. Have multiple notebook tabs open
2. Use Ctrl+F to bring up search box and start typing. This should work fine with no missed keystrokes
3. Repeat step 2 for each open notebook tab to open the search box for the first time
4. Once all notebook tabs have a search box open, switch tabs and use Ctrl+F to switch focus to the search box, start typing and the first keystroke should be missed

![missed_keystroke](https://github.com/user-attachments/assets/261d87ae-7e2e-4d43-b183-176cee579d64)

In the gif above I'm typing 'test' every time I switch focus to the search box via keyboard shortcut Ctrl+F. You can see that after the first pass of opening the search box on each tab, the subsequent tab cycles show 'est' in the search box because the first 't' keystroke was missed.

## Expected behavior

First keystroke is not missed when switching focus to existing search box using Ctrl+F keyboard shortcut.

## Context

- Image: quay.io/jupyter/base-notebook:2024-11-04
- Operating System and version: Image runs Ubuntu 24.04
- Browser and version: Reproducible in Firefox 132.0.1, Edge 130.0.2849.68, Chrome 130.0.6723.92
- JupyterLab version: 4.2.5, but also reproducible on 4.2.1, 4.0.5, 4.3.0 etc

<details><summary>Troubleshoot Output</summary>
[jupyter_troubleshoot.txt](https://github.com/user-attachments/files/17640014/jupyter_troubleshoot.txt)

</details>

<details><summary>Command Line Output</summary>
[jupyter_debug.txt](https://github.com/user-attachments/files/17639954/jupyter_debug.txt)
</details>

<details><summary>Browser Output</summary>

<pre>
Nothing printed on browser JavaScript console
</pre>
</details>

[...]

PR #17005, opened 2024-11-27
Fixing missed first keystroke on Ctrl+F 

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/16928

### DIFF
--- a/galata/test/jupyterlab/notebook-search.test.ts
+++ b/galata/test/jupyterlab/notebook-search.test.ts
@@ -67,6 +67,16 @@ test.describe('Notebook Search', () => {
     await expect(page.locator('[placeholder="Find"]')).toHaveValue('1234');
   });
 
+  test('Consecutive searches in the search box', async ({ page }) => {
+    await page.keyboard.press('Control+f');
+
+    await page.fill('[placeholder="Find"]', 'jupyter');
+    await page.keyboard.press('Control+f');
+    await page.locator('[placeholder="Find"]').pressSequentially('jupyter');
+
+    await expect(page.locator('[placeholder="Find"]')).toHaveValue('jupyter');
+  });
+
   test('RegExp parsing failure', async ({ page }) => {
     await page.keyboard.press('Control+f');
 
--- a/packages/documentsearch/src/searchmodel.ts
+++ b/packages/documentsearch/src/searchmodel.ts
@@ -104,13 +104,8 @@ export class SearchDocumentModel
     return this._initialQuery;
   }
   set initialQuery(v: string) {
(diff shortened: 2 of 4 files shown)


itsmevichu:
The issue can be summarized as follows:

1. Press Ctrl+F and search for a term, such as "test."
2. Press Ctrl+F again and type "test." Only "est" is entered.
3. Repeat the previous step, and again only "est" is entered. After one more repetition, "test" appears.

The underlying cause:
- The issue is caused by the useEffect hook. When you first enter "test" and press Ctrl+F, the initial value becomes "test," triggering the useEffect. This causes the first letter, "t," to be selected, and typing the remaining "est" replaces the "t."
- When you press Ctrl+F again and type "test," the initial will now become "est," which triggers useEffect again, causing the "t" to be skipped, so only "est" appears.
- In the third attempt, pressing Ctrl+F and typing "test" doesn’t trigger useEffect because the initial value is still "est" (no change). So, the full "test" appears.
- This cycle keeps happening

https://github.com/jupyterlab/jupyterlab/blob/e0db5da7884f1fdcd8d2c0a79f8b0c66aa8b862e/packages/documentsearch/src/searchview.tsx#L99-L135

@krassowski Any idea on the above issue? I don’t understand the purpose of useEffect here—everything seems to work fine without it.

krassowski:
I recall it is needed to allow populating the search box with selected text, see how these related tests are now failing:

```
    [jupyterlab] › test/jupyterlab/file-search.test.ts:65:5 › Populate search box with selected text 
    [jupyterlab] › test/jupyterlab/notebook-search.test.ts:104:7 › Notebook Search › Populate search box with selected text 
```

krassowski:
Maybe `initialValue` should be ignored after initial render?

itsmevichu:
@krassowski The issue arises when the search input has no initialValue, causing the previously searched term ('searchExpression') to be set as the initialValue. Since both the component re-render and useEffect depend on initialValue, pressing Ctrl+F on an already selected search input and typing 't' updates the initialValue, triggers a re-render, and resets the defaultValue in the input box, overwriting 't' with initialValue 'test'. To address this, I have made changes  to avoid frequent updates to initialValue during consecutive Ctrl+F actions; and also handled the pre-population without the use of initialValue.

krassowski:
Note: visual regression failures and unit test failures are unrelated (https://github.com/jupyterlab/jupyterlab/issues/17012), it's due ot a very recent release of `httpx`.

krassowski (approved):
Thank you @itsmevichu!

krassowski:

Issue #16887, opened 2024-10-23
Reference API documentation no longer shows up in search results in docs


krassowski:
## Description

Previously the sphinx search results included the reference API links. It no longer works as of today.

## Reproduce

https://jupyterlab.readthedocs.io/en/latest/search.html?q=WindowedListModel#

![image](https://github.com/user-attachments/assets/fb96244e-07f0-472b-aed4-dcafaf582f03)

## Expected behavior

Link to https://jupyterlab.readthedocs.io/en/latest/api/classes/ui_components.WindowedListModel.html shows up

## Context

It looks like this might be related to recent changes on RTD?

I don't recall how this was implemented but I think @fcollonval worked on it and might be able to give a hint

JasonWeill:
ReadTheDocs recently made some updates to the way it publishes docs. Through no action of our own, this migration might be related to the bugs we're seeing. We need to figure out how to reindex the docs properly.

bollwyvl:
The good news: on `latest`, RTD-provided, runtime-injected search [addon](https://github.com/readthedocs/addons) (vs the previous `pip` install and `conf.py`-rewriting one), _can_ be accessed. Either by:
- pressing <kbd>/</kbd> and typing (nice)
- clicking the flyout in the bottom right corner, 
    - clicking the search box (which is actually a button: a11y fail)

... one gets the expected `typedoc` HTML search hit.

| light | dark |
|-|-|
| ![image](https://github.com/user-attachments/assets/c6b34369-b8b2-4dc8-9e1c-9a33f5ed2245) | ![image](https://github.com/user-attachments/assets/b5a394d8-07d5-4054-8e68-a28ea306b68a)


The bad news: `pydata-sphinx-theme` is [not using it](https://github.com/pydata/pydata-sphinx-theme/issues/1977), so it falls back to the workmanlike built-in search index/js which one gets via a local `make`. That issue has some links to some workarounds. Note further that the PST theme is leaking into the search box, especially notable in the dark theme.

Meanwhile: On JupyterLite, we had to blanket disable all the hotkeys, as pressing <kbd>/</kbd> or <kbd>d</kbd> when in a `contenteditable` element (e.g. by CodeMirror6), making the Jupyter UIs... very awkward.

I'll see if there's something lightweight in `conf.py` (no `jquery`, please) we can hide behind a `if os.environ.get("READTHEDOCS")`on #17006 or a follow-on.

PR #17006, opened 2024-11-27
update to typedoc 0.26, fix missing module pages

### LINKED ISSUE #16768: Update to typedoc 0.26.*
### Problem

`typedoc 0.26` would likely provide a better documentation reading experience, as it treats all comments as markdown more consistently.

It also changes out its markdown parser from an old, known-ReDoS-susceptible version `marked` to `markdown-it`. This adds a new set of dependencies, but hosting two version of `marked` isn't free, either.

### Proposed Solution

- [x] Update to `typedoc 0.26`
- [x] accept new dependencies
- [x] handle existing broken links, probably with a plugin
  - e.g. the link to _notebook_ on
    - https://jupyterlab.readthedocs.io/en/latest/api/modules/outputarea.html point to
    - https://jupyterlab.readthedocs.io/en/latest/api/notebook  (which doesn't exist)
  - this will likely require a plugin, or pointing to a concrete file (e.g. `../notebook/src/index.ts`)
- [ ] start treating docs warnings as errors with either `--treatWarningsAsErrors` or  `--treatValidationWarningsAsErrors` (if our namespace/interface puns can't be worked around)

### Additional context

Noted on #15733.

bollwyvl:
## References

- fixes #16768
- ultimately motivated by unblocking https://github.com/jupyterlab/jupyterlab/pull/15733
  - > we want _one_ version of `marked`, but `typedoc <0.26` breaks without a very old one
  - > `>=0.26` use `markdown-it`

## Code changes

- [x] bump `typedoc` in `package.json`
- [x] relock/dedupe `yarn.lock`
- [x] fix _some_ of the existing broken links in post
  - > as seen on the first body link to `cells` on the [`notebook` API](https://jupyterlab.readthedocs.io/en/latest/api/modules/notebook.html) page
  - > e.g. `docs/_build/api/modules/cells.html` -> `docs/_build/api/cells/index.html`
- [x] fix _new_ broken links, introduced by `typedoc` copying things to `media`
- [x] pre-build `typedoc` in ReadTheDocs (avoid deep, hard-to-debug errors in `node`-in-`sphinx`)

## User-facing changes

- some markdown rendering should be more accurate
- some broken links at the top level of `modules/*.html` won't be broken 

- n/a

## Future Work

- fix the thousands of `typedoc` warnings
  - adopt the `tsdoc.json` convention
  - add eslint plugin for `tsdoc`


cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

cokyman (approved):

.readthedocs.yml
```diff
@@ -5,6 +5,12 @@ build:
   tools:
     python: "3.10"
     nodejs: "20"
+  jobs:
+    post_install:
+      # avoid the lengthy `typedoc` build inside `sphinx` for cleaner logs
```
krassowski, line 10: Will we need to reinstate it for https://github.com/jupyterlab/jupyterlab/issues/16887?

bollwyvl: I don't believe so. 

Like those screenshots point out, all the `.html` is already being indexed in elasticsearch by the RTD addon, so if we opt to do special things if `READTHEDOCS`, that would be taken care of without looking at a single file... just has to be in `$READTHEDOCS_OUTPUT/html` at the end of the day.

If we want to do indexing by overloading `feed` so that PRs and local builds are right, that _would_ happen inside the sphinx lifecycle, but doing so against already-generated `.html` is fine.

If we skip both, and use something like `lunr(.js|-py)` (buh), it would also work to already have the files... and potentially even the pre-generated index.

krassowski (approved):
Thank you!

krassowski:
Huh, GitHub appears stuck checking mergeability for a few minutes now, at least for me:

image

bollwyvl:
Early in this reboot, I tried typedoc 0.27.0, and found it lacking... as a follow-on, will try that again, but this would unblock the `mermaid` work.

bollwyvl:
Thanks! I don't think 0.27 (#17070) is ready for primetime due to changing 50% of the links, but I do like the page content better. Will keep tinkering over there (and dig into mermaid in anger).

PR #17013, opened 2024-11-29
Bump httpx v0.28.0

davidbrochart:
With [httpx v0.28.0](https://github.com/encode/httpx/pull/3419), the deprecated `proxies` argument has been removed. But we still use it in the `pypi` extension manager:
```
[W 2024-11-29 09:22:50.941 LabApp] Failed to instantiate the extension manager pypi. Falling back to read-only manager.
      File "/home/david/.local/share/mamba/envs/tmp-jupyter-collaboration/lib/python3.13/site-packages/jupyterlab/labapp.py", line 837, in initialize_handlers
        ext_manager = manager_factory(app_options, listings_config, self)
      File "/home/david/.local/share/mamba/envs/tmp-jupyter-collaboration/lib/python3.13/site-packages/jupyterlab/extensions/__init__.py", line 46, in get_pypi_manager
        return PyPIExtensionManager(app_options, ext_options, parent)
      File "/home/david/.local/share/mamba/envs/tmp-jupyter-collaboration/lib/python3.13/site-packages/jupyterlab/extensions/pypi.py", line 134, in __init__
        self._httpx_client = httpx.AsyncClient(proxies=proxies)
                             ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    TypeError: AsyncClient.__init__() got an unexpected keyword argument 'proxies'
```

## References

See https://github.com/encode/httpx/pull/2879.

## Code changes

Use `mounts` instead of `proxies`.

## User-facing changes

The `pypi` extension manager works fine.

None.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
 ]
 dependencies = [
     "async_lru>=1.0.0",
-    "httpx>=0.25.0",
+    "httpx>=0.28.0",
     "importlib-metadata>=4.8.3;python_version<\"3.10\"",
     "importlib-resources>=1.4;python_version<\"3.9\"",
     "ipykernel>=6.5.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
 ]
 dependencies = [
     "async_lru>=1.0.0",
-    "httpx>=0.28.0",
+    "httpx~=0.28.0",
     "importlib-metadata>=4.8.3;python_version<\"3.10\"",
     "importlib-resources>=1.4;python_version<\"3.9\"",
     "ipykernel>=6.5.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -36,7 +36,7 @@ classifiers = [
 ]
 dependencies = [
     "async_lru>=1.0.0",
-    "httpx>=0.25.0",
+    "httpx>=0.28.0",
```
krassowski, line 39: This was pretty bad breakage for users as extension manager stopped working on new installs.
Do we want to use `~=0.28.0` to avoid this in the future?

davidbrochart: We should have had a deprecation warning though.
If we add an upper bound, we will have to manually upgrade `httpx`, but I don't have a strong opinion.

krassowski: I think this will ease maintenance burden, and we can always unpin it later.

krassowski:

pyproject.toml
```diff
@@ -36,7 +36,7 @@ classifiers = [
 ]
 dependencies = [
     "async_lru>=1.0.0",
-    "httpx>=0.25.0",
+    "httpx~=0.28.0",
```
bollwyvl, line 39: As a potentially-security-related package, having a pin of `>=0.28.0,<0.29.0` seems like we're asking for a rushed future release.

PR #17022, opened 2024-12-02
Improve performance of rendering stdout/stderr

### LINKED ISSUE #16957: Logging to stderr is significantly slower in Jupyterlab vs IPython

krassowski:
## References

- another fix for performance regression reported in #16957
- follow-up to https://github.com/jupyterlab/jupyterlab/pull/16975

## Code changes

- [x] Adds a test for a fast path discussed in https://github.com/jupyterlab/jupyterlab/issues/16957#issuecomment-2507653383
   - I initially tried to implement test timing the performance but in that case it does not work as jsdom does not use a complied implementation for sanitization of `textContent` (it's not using a real browse) and also the ANSI parsing has quite some overhead when a match is detected so the fast test case even on `main` was faster than the test case for slow path
- [x] Implements the fast path when logging without ASCII escapes

## User-facing changes

Rendering the stdout and stderr will be faster

| Browser | `4.3.0` | #16975 | #16975 + this PR |
|--|--|--|--|
| Chrome 131.0 | 67s | 25s | 19s |
| Firefox 133.0 | 136s | 35s | 21s |

<details>
<summary>Test case</summary>

```python
import logging

logger = logging.getLogger()

for i in range(4000):
```

</details>

None

### DIFF
--- a/packages/rendermime/src/renderers.ts
+++ b/packages/rendermime/src/renderers.ts
@@ -797,6 +797,19 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
 }
 
 /**
+ * Sanitize HTML out using native browser sanitizer.
+ *
+ * Compared to the `ISanitizer.sanitize` this does not allow to selectively
+ * allow to keep certain tags but escapes everything; on the other hand
+ * it is much faster as it uses platform-optimized code.
+ */
+function nativeSanitize(source: string): string {
+  const el = document.createElement('span');
+  el.textContent = source;
+  return el.innerHTML;
+}
+
+/**
  * Render the textual representation into a host node.
  *
  * Implements the shared logic for `renderText` and `renderError`.
@@ -808,10 +821,16 @@ function renderTextual(
   // Unpack the options.
   const { host, sanitizer, source } = options;
 
-  // Create the HTML content.
-  const content = sanitizer.sanitize(Private.ansiSpan(source), {
-    allowedTags: ['span']
-  });
+  const ansiPrefixRe = /\x1b/; // eslint-disable-line no-control-regex
+  const hasAnsiPrefix = ansiPrefixRe.test(source);
+
+  // Create the HTML content:
+  // If no ANSI codes are present use a fast path for escaping.
+  const content = hasAnsiPrefix
+    ? sanitizer.sanitize(Private.ansiSpan(source), {
+        allowedTags: ['span']
+      })
+    : nativeSanitize(source);
 
   // Set the sanitized content for the host node.
   const pre = document.createElement('pre');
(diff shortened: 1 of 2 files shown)


packages/rendermime/src/renderers.ts
```diff
@@ -808,10 +821,16 @@ function renderTextual(
   // Unpack the options.
   const { host, sanitizer, source } = options;
 
-  // Create the HTML content.
-  const content = sanitizer.sanitize(Private.ansiSpan(source), {
-    allowedTags: ['span']
-  });
+  const ansiPrefixRe = /\x1b/; // eslint-disable-line no-control-regex
```
krassowski, line 824: We need to disable [`no-control-regex`](https://eslint.org/docs/latest/rules/no-control-regex) here, the same way we already disable it in `ansiSpan()`

https://github.com/jupyterlab/jupyterlab/blob/c9700a2802dabfb44109b4327473b6a593e7cc97/packages/rendermime/src/renderers.ts#L1560-L1561

because this invisible control character is exactly what we want to detect.

packages/rendermime/src/renderers.ts
```diff
@@ -796,6 +796,19 @@ export function renderText(options: renderText.IRenderOptions): Promise<void> {
...
+ *
+ * Compared to the `ISanitizer.sanitize` this does not allow to selectively
+ * allow to keep certain tags but escapes everything; on the other hand
+ * it is much faster as it uses platform-optimized code.
+ */
+function nativeSanitize(source: string): string {
+  const el = document.createElement('span');
+  el.textContent = source;
+  return el.innerHTML;
+}
```
krassowski, line 810: This is the recommended method of text sanitization as per [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-6-populate-the-dom-using-safe-javascript-functions-or-properties)

We also already use it in other performance-sensitive places like in the completer:

https://github.com/jupyterlab/jupyterlab/blob/c9700a2802dabfb44109b4327473b6a593e7cc97/packages/completer/src/model.ts#L10-L17

krassowski (commented):
Just leaving some helpful notes for the reviewers.

jtpio (approved):
Thanks @krassowski for the fix and leaving helpful comments!

jtpio:

PR #17024, opened 2024-12-02
Fixes debugger UI-test with python>=3.12

brichet:
## References

Should fix https://github.com/jupyterlab/jupyterlab/issues/16982

Reverts https://github.com/jupyterlab/jupyterlab/pull/16989

## Code changes

Ends the execution of the cell during debugger tests.

I don't know why it used to pass and why it was falling with python 3.12, it must be a race condition.

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/galata.yml
+++ b/.github/workflows/galata.yml
@@ -23,8 +23,6 @@ jobs:
 
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
-        with:
-          python_version: "3.11"
 
       - name: Set up browser cache
         uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


krassowski:
Does it indicate that it is impossible to close jupyterlab on Python 3.12 when debugger is at a breakpoint?

This would sound like a bug to me.

brichet:
On my side I'm able to stop jupyterlab when the debugger is at a breakpoint.

I'm a bit confuse with this issue, since it seems to be related to galata, which does not end the test if the cell is not executed..
The tests reached a timeout, stuck on [this](https://github.com/jupyterlab/jupyterlab/blob/c9700a2802dabfb44109b4327473b6a593e7cc97/galata/src/helpers/notebook.ts#L457). It was expected that the cell was not executed, since it was on a breakpoint.

That said, I'm not able to tell you the link between Python 3.12 and this galata issue.

krassowski (approved):
Fair enough, thank you!

PR #17031, opened 2024-12-03
Use Zulip for instant messaging

jtpio:
## References

Opening as a draft for now, mostly for visibility and awareness, so it can be discussed.

Related to the discussions in https://github.com/jupyter/governance/issues/182.

## Code changes

None

## User-facing changes

Point users to the Jupyter Zulip instance instead of Gitter / Matrix.

None

### DIFF
--- a/CORPORATE.md
+++ b/CORPORATE.md
@@ -67,8 +67,8 @@ accelerates the work of everyone.
 **Use public communications channels whenever possible.** JupyterLab is
 developed openly on GitHub. We have weekly video meetings on Wednesday at 9:00
 am PT on Zoom (https://calpoly.zoom.us/my/jupyter). Finally, we have a public
-chat room on gitter (https://gitter.im/jupyterlab/jupyterlab). As much as
-possible, all work and communications should take place on these public
+chat room on Zulip (https://jupyter.zulipchat.com/#narrow/channel/469762-jupyterlab).
+As much as possible, all work and communications should take place on these public
 channels. We welcome you to join these channels and introduce yourself.
 We are willing to have private conversations once in a while, but we
 try to minimize their frequency and always summarize them in our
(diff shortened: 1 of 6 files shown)


andrii-i (approved):
To me as a JupyterLab contributor and council member, move from Gitter to Zulip makes sense.

As a Jupyter Media Strategy workgroup member I also wanted to clarify that Gitter/Zulip/JupyterLab preferred chat channel discussion is out of out scope for JMS and therefore the change does not need JMS approval.

bollwyvl:
Zulip seems to be going well, and at least thus far hasn't attracted any spam bots.

If we're really going for it, having a Lab-based client as an extension that supported its model (which is rather more complex than a "dumb" chatroom, but not _quite_ as complex as the [XMPP-choose-your-own-adventure](https://xmpp.org/extensions/)) would be quite nice. First party provides [js](https://github.com/zulip/zulip-js) (but not [ts](https://github.com/zulip/zulip-js/issues/74) :crying_cat_face:) and [py](https://github.com/zulip/python-zulip-api/) (with `mypy` :heart_eyes_cat:) bindings. 

A low-risk demo might be an `ipylab` widget which used the python bindings to drive a minimal UI.

As a user, I'm _already_ using rather a few zulip instances, so supporting (and combining) several identities would be grand. And, as an actually-open-source project, we could actually _recommend_ using this for the organizational case, as organizations could host their own for "free" (ain't nothing free, especially sysadmins, and you do have to pay for mobile pushes or get _n_ proprietary vendor accounts).

jtpio:
OK, it's been two weeks and there were no objections here and in https://github.com/jupyterlab/frontends-team-compass/issues/263, so let's do the switch!

jtpio:

PR #16852, opened 2024-10-10
Run CI on Python 3.9 and 3.13 (drop 3.8 from testing matrix)

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16850
Fixes https://github.com/jupyterlab/jupyterlab/issues/13946

## Code changes

- [x] Update the minimum supported version of `setuptools` to `41.1.0`

## User-facing changes

None

## Backwards-incompatible changes

None

PR #17036, opened 2024-12-04
Drop Python 3.8

jtpio:
## References

Python 3.8 is now EOL: https://devguide.python.org/versions/#unsupported-versions

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16852

## User-facing changes

- [x] Require Python 3.9 or above to run JupyterLab.
- [x] Add the 3.13 trove classifier

None

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -31,7 +31,6 @@ from collections import ChainMap
 from functools import partial
 from pathlib import Path
 from subprocess import check_call
-from typing import List
 
 HERE = Path(__file__).parent.resolve()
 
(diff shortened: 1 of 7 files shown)


krassowski:
I think it makes sense to bump target version for `ruff` too, here:

https://github.com/jupyterlab/jupyterlab/blob/761d34f3325c4e93b1474fa9ca631cffd82a8081/pyproject.toml#L250-L252

And run it to transform old syntax to new syntax where possible (probably removing a number of `typing` imports).

krassowski (approved):
Thank you @jtpio!

Issue #17037, opened 2024-12-04
Wrong file name order


Nriver:
## Description

File order is wrong. Numbers are not placed in correct order.

## Reproduce

![ksnip_20241204-164821](https://github.com/user-attachments/assets/55c00fca-cc66-449e-955e-bf8dff0b091b)

## Expected behavior

like this

![ksnip_20241204-164745](https://github.com/user-attachments/assets/a275d2fc-7ab9-41d9-b34c-192363a7c836)

krassowski:
Is this the same as https://github.com/jupyterlab/jupyterlab/issues/14392?

Nriver:
No, I thought it was the same. But it's not. That issue is the difference of `ASCII` order and natural sort order.

PR #17038, opened 2024-12-04
Fix filebrowser name order

Nriver:
## References

#17037

## Code changes

Fix wrong usage of `localeCompare` function. 

## User-facing changes

Before:
ksnip_20241204-164821

After:
ksnip_20241204-164745

No

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -3603,7 +3603,7 @@ namespace Private {
       // Sort by name
       copy.sort(
         compare((a: Contents.IModel, b: Contents.IModel) => {
-          return b.name.localeCompare(a.name);
+          return b.name.localeCompare(a.name, navigator.language || navigator.userLanguage, {numeric: true, sensitivity: 'base'});
         })
       );
     }
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -3568,7 +3568,7 @@ namespace Private {
      * Compare two items by their name using `translator.languageCode`, with fallback to `navigator.language`.
      */
     function compareByName(a: Contents.IModel, b: Contents.IModel) {
-      const languageCode = translator.languageCode.replace('_', '-');
+      const languageCode = (translator.languageCode ?? navigator.language).replace('_', '-');
       try {
         return a.name.localeCompare(b.name, languageCode, {
           numeric: true,
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -3533,10 +3533,16 @@ namespace Private {
   export function sort(
     items: Iterable<Contents.IModel>,
     state: DirListing.ISortState,
-    sortNotebooksFirst: boolean = false
+    sortNotebooksFirst: boolean = false,
+    translator?: ITranslator
   ): Contents.IModel[] {
     const copy = Array.from(items);
     const reverse = state.direction === 'descending' ? 1 : -1;
+    // Determine language code from the translator or fallback to navigator.language
+    translator = translator || nullTranslator;
+    const languageCode = translator
+      ? translator.languageCode
+      : navigator.language;
 
     /**
      * Compares two items and returns whether they should have a fixed priority.
@@ -3578,7 +3584,10 @@ namespace Private {
         }
 
         // Default sorting is alphabetical ascending
-        return a.name.localeCompare(b.name);
+        return a.name.localeCompare(b.name, languageCode, {
+          numeric: true,
+          sensitivity: 'base'
+        });
       };
     }
 
@@ -3603,7 +3612,10 @@ namespace Private {
       // Sort by name
       copy.sort(
         compare((a: Contents.IModel, b: Contents.IModel) => {
-          return b.name.localeCompare(a.name, navigator.language || navigator.userLanguage, {numeric: true, sensitivity: 'base'});
+          return b.name.localeCompare(a.name, languageCode, {
+            numeric: true,
+            sensitivity: 'base'
+          });
         })
       );
     }
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -342,7 +342,8 @@ export class DirListing extends Widget {
     this._sortedItems = Private.sort(
       this.model.items(),
       state,
-      this._sortNotebooksFirst
+      this._sortNotebooksFirst,
+      this.translator
     );
     this._sortState = state;
     this.update();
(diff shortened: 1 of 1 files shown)


Nriver:
I noticed there are several instances where `localeCompare` is used incorrectly. I'm not sure what specific features on the page rely on these implementations.

Nriver:
Related doc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

A quick test code snippet:
```javascript
const a = '10';
const b = '2';

// this returns -1, means '10' will placed before '2', which is the wrong order.
console.log(a.localeCompare(b));

// this returns 1, means '2' will placed before '10', this is the right order.
console.log(a.localeCompare(b, 'en', {numeric: true, sensitivity: 'base'}));

// automatically get user locale
console.log(a.localeCompare(b, navigator.language || navigator.userLanguage, {numeric: true, sensitivity: 'base'}));
```

packages/filebrowser/src/listing.ts
```diff
@@ -3603,7 +3603,7 @@ namespace Private {
       // Sort by name
       copy.sort(
         compare((a: Contents.IModel, b: Contents.IModel) => {
-          return b.name.localeCompare(a.name);
+          return b.name.localeCompare(a.name, navigator.language || navigator.userLanguage, {numeric: true, sensitivity: 'base'});
```
krassowski, line 3606: I think that instead of `navigator.language || navigator.userLanguage` we should first try to use the language of the UI. You would need to pass `ITranslator` instance and access its [`languageCode`](https://jupyterlab.readthedocs.io/en/latest/api/interfaces/rendermime_interfaces.IRenderMime.ITranslator.html#languageCode) parameter. I am just not 100% sure if the language codes have the same BCP 47 format.

Nriver: The embarrassing part is that I’m not sure how to run the project from source. I tried starting with `npm install`, but it failed.

To make the change, I had to do some reverse engineering: I modified the compiled JavaScript code, then traced it back to the source.

krassowski: We use `yarn v3` rather than `npm`. Invoking just `jlpm` (a pinned version of `yarn` berry) should install the dependencies for you. See the full instructions here: https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#installing-jupyterlab

krassowski: The build on CI is failing with:

```
../filebrowser/src/listing.ts(3606,79): error TS2551: Property 'userLanguage' does not exist on type 'Navigator'. Did you mean 'language'?
```

Nriver:
The `userLanguage` section was used for compatibility with older browsers, so it can be safely removed now that we are targeting modern browsers.

Thanks for the information! I've been able to get the project running : )

Now, I’m facing a new issue. When I try to retrieve the `languageCode` from the `ITranslator`, the instance always defaults to `nullTranslator`. As a result, the `languageCode` is fixed to 'en' and it will ignore the value of `navigator.language`.

Here’s the relevant code:

```javascript
// Determine language code from the translator or fallback to navigator.language
const languageCode = translator
  ? translator.languageCode
  : navigator.language;
```

packages/filebrowser/src/listing.ts
Nriver: The `userLanguage` section was used for compatibility with older browsers, so it can be safely removed now that we are targeting modern browsers.

krassowski:
I see that you added `translator` as an argument to the `sort` function in the `Private` namespace - good! Two notes:
- it be required, you can remove `?` (and thus the `nullTranslator`)
- you need to pass `translator` in `DirListing.sort` in as an argument here:

https://github.com/jupyterlab/jupyterlab/blob/7ff8d16493f1c8b70cd76658aee80e9453c396a0/packages/filebrowser/src/listing.ts#L341-L346

It looks like it is available as `this.translator`:

https://github.com/jupyterlab/jupyterlab/blob/7ff8d16493f1c8b70cd76658aee80e9453c396a0/packages/filebrowser/src/listing.ts#L219-L230

Nriver:
Thanks for the advice! I was able to retrieve the `languageCode` from the `translator` as you suggested.

Unfortunately, as you suspected, the format differs slightly, and `localeCompare` doesn't recognize it. `languageCode` gives `zh_CN` and `navigator.language` gives `zh-CN`

ksnip_20241205-174130

Quick test:
```javascript
const a = '10';
const b = '2';

console.log(a.localeCompare(b, 'zh-CN', {numeric: true, sensitivity: 'base'}));
// this one will error
console.log(a.localeCompare(b, 'zh_CN', {numeric: true, sensitivity: 'base'}));
```

So, what should we do? Simply use `navigator.language` or do a string replace from `_` to `-`?

krassowski:
I think we can replace `_` with `-`. The underlying [`python-babel`](https://github.com/python-babel/babel) package uses CLDR identifiers [which are based on BCP 47](https://cldr.unicode.org/index/cldr-spec/picking-the-right-language-code) identifiers that are expected by browsers. For extra caution, could you wrap the `localeCompare` call with `translator`-derived code in try-except and in except fallback to `navigator.language` (and log a warning message)

Nriver:
Did a little refactor work, it's working now : )

packages/filebrowser/src/listing.ts
```diff
@@ -3603,7 +3626,7 @@ namespace Private {
       // Sort by name
       copy.sort(
         compare((a: Contents.IModel, b: Contents.IModel) => {
-          return b.name.localeCompare(a.name);
+          return compareByName(b, a);
```
krassowski, line 3629:
```suggestion
          return compareByName(b, a, translator);
```

packages/filebrowser/src/listing.ts
```diff
@@ -3562,6 +3564,27 @@ namespace Private {
       return 0;
     }
 
+    /**
+     * Compare two items by their name using `translator.languageCode`, with fallback to `navigator.language`.
+     */
+    function compareByName(a: Contents.IModel, b: Contents.IModel) {
```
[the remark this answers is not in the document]
krassowski: Ah, no I am wrong. It is in closure!

packages/filebrowser/src/listing.ts
```diff
@@ -3562,6 +3564,27 @@ namespace Private {
...
+      const languageCode = translator.languageCode.replace('_', '-');
```
krassowski, line 3571: It looks like sometimes `translator.languageCode` will be undefined. It should not be but it looks like the tests are failing because of that:

```
Parsed an error from text content: undefined TypeError: undefined is not an object (evaluating 'translator.languageCode.replace')
```

Maybe this would help?

```suggestion
      const languageCode = (translator.languageCode ?? navigator.language).replace('_', '-');
```

packages/filebrowser/src/listing.ts
```diff
@@ -3562,6 +3564,27 @@ namespace Private {
       return 0;
     }
 
+    /**
+     * Compare two items by their name using `translator.languageCode`, with fallback to `navigator.language`.
+     */
+    function compareByName(a: Contents.IModel, b: Contents.IModel) {
+      const languageCode = (translator.languageCode ?? navigator.language).replace('_', '-');
```
krassowski, line 3571: Linting:

```suggestion
      const languageCode = (
        translator.languageCode ?? navigator.language
      ).replace('_', '-');
```

krassowski (approved):
That's a huge improvement IMO!

| Before | After |
|---|--|
| image | Screenshot from 2024-12-06 17-41-21 |

krassowski:

rraadd88:
Hi, could there be an option to choose the sorting method, like in 
rstudio (`rstudioapi::writePreference("sort_file_names_naturally", TRUE)`) and vs-code (`explorer.sortOrder`)? 
I find this as one of the limitations of jupyterlab, which I'd guess should be possible to fix.

Issue #17035, opened 2024-12-03
Reconsider hard/exact `httpx` dependency


bollwyvl:
### Problem

`4.3.2` shipped with a very tight pin on `httpx ~=0.28.0` (#17013) due to a small (zero?) window for changing its API. For security-related packages, such pins can make it difficult for users (or their administrators) to keep their environments free from (perceived) vulnerabilities through normal means.

### Proposed Solution

Some alternatives:
- revisit whether something from the `jupyter_server` stack (such as `tornado.httpclient`) can fulfill the "requirement" from #15219
- failing that, make `httpx` an optional dependency, e.g. `jupyterlab[extension-manager-httpx]`
  - and, if possible, loosen the constraint and perform version sniffing to allow for a wider window of (tested) versions, as these may incur _other_ overhead (see below)

### Additional context

`httpx` also hauls along a number of _other_ novel dependencies (vs `jupyter_server`):

```
+h11
+h2
+hpack
+httpcore
+hyperframe
```

jtpio:
Quickly searching the code base for `httpx`, it seems to be used in the PyPI extension manager only:

https://github.com/jupyterlab/jupyterlab/blob/7ff8d16493f1c8b70cd76658aee80e9453c396a0/jupyterlab/extensions/pypi.py#L68-L69

dlqqq:
The narrow version range is failing CI on Jupyter AI, because `pip install` is installing several versions of several packages to find a version that supports the version range of `httpx` required by JupyterLab: https://github.com/jupyterlab/jupyter-ai/issues/1138

dlqqq:
In this thread, I see three main proposals:

1. Loosen the `httpx` version range.
2. Make the requests with `tornado.httpclient`, and drop `httpx` entirely.
3. Migrate `httpx` to an optional dependency group.

No. 1 is OK but it still depends on `httpx`, which hasn't had a stable major release yet. Plus it brings in a lot of other dependencies, as you mentioned. 2) is also OK, but I haven't verified that `tornado.httpclient` supports the existing use-cases while producing readable code. 3) raises some usability concerns for me. With 3), it seems as though users would have to install this optional dependency group to use the extension manager. IMO, this sort of defeats the point of the extension manager being a UI for people less familiar with using `pip` on the command line.

I suggest we also explore a 4th option: We can replace `httpx` with `requests`, and enjoy the ability to use a mature HTTP client library without dealing with breaking changes.

[...]

PR #17041, opened 2024-12-05
restore bottom httpx version window

bollwyvl:
## References

- put back `httpx >=0.25.0`
- fixes https://github.com/jupyterlab/jupyterlab/issues/17035

## Code changes

- sniff `httpx.__version__`, use appropriate arguments

## User-facing changes

- users will be easier able to install `jupyterlab 4.3.3` while the libraries they are interested in adjust to the `httpx` API breakage

- n/a

## Future Work

- looks like `https_proxy_url` isn't used. 
  - seems bad.
- the proxy feature has no tests
  - this could be done with `proxy-py` (BSD-3-Clause, no deps, pure python), but its probably more important to get this out so that people (that probably dont' even care about proxies) can actually _install_ the software next to their other code
- as raised when this was _originally_ added, the `fetch` should be isolated within a configurable
  - `jupyterlab_server` already hauls in `requests` which can handle proxies, but is blocking, but could be wrapped in a `run_in_executor`
- going forward, we should likely never use `pkg ~x.y.z` in `[project.dependencies]` (test deps are what they are)

### DIFF
--- a/jupyterlab/extensions/pypi.py
+++ b/jupyterlab/extensions/pypi.py
@@ -21,6 +21,7 @@ from subprocess import CalledProcessError, run
 from tarfile import TarFile
 from typing import Any, Callable, Optional
 from urllib.parse import urlparse
+from packaging.version import Version
 from zipfile import ZipFile
 
 import httpx
@@ -49,8 +50,6 @@ class ProxiedTransport(xmlrpc.client.Transport):
         return connection
 
 
-xmlrpc_transport_override = None
-
 all_proxy_url = environ.get("ALL_PROXY")
 # For historical reasons, we also support the lowercase environment variables.
 # Info: https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/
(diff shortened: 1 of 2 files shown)


jupyterlab/extensions/pypi.py
```diff
@@ -49,24 +50,32 @@ def make_connection(self, host):
...
+xmlrpc_transport_override = None
 
 if http_proxy_url:
     http_proxy = urlparse(http_proxy_url)
     proxy_host, _, proxy_port = http_proxy.netloc.partition(":")
 
     proxies = {
         "http://": httpx.HTTPTransport(proxy=http_proxy_url),
-        "https://": httpx.HTTPTransport(proxy=https_proxy_url),
+        "https://": httpx.HTTPTransport(proxy=http_proxy_url),
```
krassowski, line 74: I do not follow the reason for removal of `s` here:

```suggestion
        "https://": httpx.HTTPTransport(proxy=https_proxy_url),
```

krassowski (approved):
Thank you @bollwyvl!

krassowski:

krassowski:

PR #17047, opened 2024-12-05
Reset resizeData after column adjustment to allow file dragging

### LINKED ISSUE #17042: File browser column resize prevents drag and drop
## Description

When first loaded the files in the side browser can be click and dragged and dropped in a folder etc. 

But after once adjusting the left-right split of the name/size columns, the whole side browser if clicked and dragged continues to adjust the split rather than clicking and dragging a file. 

1. View file browser (ctrl+b)
2. Validate one can click-and-drag a file
3. Then click to adjust the name/size column width 
4. Then try and click+drag a file
5. Column will automatically re-adjust width and no file can be dragged.

## Expected behavior

Column adjustment should only happen when the column header is selected

## Context

- Operating System and version: Mac Sequoia 15.1.1
- Chrome Version 131.0.6778.109 (Official Build) (arm64) & Safari Version 18.1.1 (20619.2.8.11.12)
- JupyterLab version: 4.3.1_1 and 4.3.2_1

Darshan808:
## References
Fixes #17042 

## Description
This PR resolves a bug where the side browser in the UI fails to allow file dragging after adjusting the column split between name/size. Once the split was adjusted, subsequent drag events incorrectly adjusted the column split instead of enabling file drag-and-drop.

## Root Cause
The `resizeData` object was not set to `null` after the `mouseUp` event. As a result, the drag logic interpreted any further drag actions as column resizing since `resizeData` had a non-null value.

## Fix
- Added a reset step to set `resizeData = null` after the `mouseUp` event if `resizeData` was set during the resize operation.

## Behavior After Fix
- Column adjustment now only occurs when the column header is dragged.
- File dragging and dropping functionality works as intended after resizing columns.

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -1477,6 +1477,7 @@ export class DirListing extends Widget {
     // Remove the resize listeners if necessary.
     if (this._resizeData) {
       this._resizeData.overrides.dispose();
+      this._resizeData = null;
       document.removeEventListener('mousemove', this, true);
       document.removeEventListener('mouseup', this, true);
       return;


krassowski (approved):
Thank you @Darshan808!

krassowski:

Issue #15795, opened 2024-02-13
Unable to scroll away from the active cell


Yura52:
## Software

```
Client: macOS=12.5 & Firefox=122.0.1
Server: Ubuntu 18.04.6 LTS

jupyterlab=4.1.1
python=3.10.12
```

## The bug

- I have a notebook with four cells and I am executing them with Shift+Enter.
- When the last (the fourth) one is executed, a new one (the fifth) is created and becomes "active", which is expected.
- Then, using my touchpad, I want scroll to previous cells to check their code/output.
- However, when I start scrolling, JupyterLab suddenly returns my screen back to the active cell (the fifth one). My current workaround is to very quickly scroll and explicitly click any other cell (i.e. make a different cell active), which is inconvenient.

## More information

Not sure what debug information I can add to make it easier to debug this issue, but I am ready to do that.

krassowski:
Does it also happen if you wait for say 1 second before scrolling? Does it also happen in JupyterLab 4.0.12?

Yura52:
> Does it also happen if you wait for say 1 second before scrolling?

Yes

> Does it also happen in JupyterLab 4.0.12?

After quick testing, I cannot reproduce the issue with JupyterLab 4.0.12

Yura52:
UPD: the reported issue happens on JupyterLab 4.1.0 (the 4.1.1 version is not available to me for some reason)

krassowski:
Thank you. Three more questions:
- Does it also happen when scrolling using a mouse?
- In Settings Editor → Notebook → Windowing mode do you have "defer" or "full"?
- Can you make a short recording?

Yura52:
> Does it also happen when scrolling using a mouse?

Unfortunately, I don't have a mouse to test this.

> "defer" or "full"?

"defer". With "full" or "none", it seems that I cannot reproduce the issue. I should mention that the overall reproducibility is not 100% even with "defer" (though it happens often enough to feel like a non-negligible issue), so I don't know how representative my experience is.

> Can you make a short recording?

Ah, I cannot share the actual notebook, and I cannot reproduce the issue in a different one >_<

krassowski:
Do you have jupyterlab-lsp extension installed?

Yura52:
No, I don't have the jupyterlab-lsp extension.

guidov:
[...]

PR #17064, opened 2024-12-09
Bump `@codemirror` packages

jtpio:
## References

- https://github.com/jupyterlab/jupyterlab/issues/15795
- https://github.com/codemirror/dev/issues/1490

## Code changes

Update to https://github.com/codemirror/view/releases/tag/6.35.3

## User-facing changes

Hopefully help with https://github.com/jupyterlab/jupyterlab/issues/15795


krassowski:
The test failures are related:

```
    Cannot find module '@marijn/find-cluster-break' from '../../node_modules/@codemirror/state/dist/index.cjs'
```

krassowski:
Otherwise on Binder it looks good, I was not able to reproduce https://github.com/jupyterlab/jupyterlab/issues/15795 so far on Binder.

yarn.lock
```diff
@@ -5398,6 +5401,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@marijn/find-cluster-break@npm:^1.0.0":
+  version: 1.0.1
+  resolution: "@marijn/find-cluster-break@npm:1.0.1"
```
krassowski, line 5406: I think bumping resolution to 1.0.2 should help as per https://github.com/marijnh/find-cluster-break/issues/1

jtpio: nice, thanks, let's try that.

krassowski:
Checklist after reviewing changelogs:

- `@codemirror/autocomplete` 6.16.0 → 6.18.3:
    - mostly irrelevant as we implement our own completer
    - [ ] Verify that "Add Alt-Backtick as additional binding on macOS, where IME can take over Ctrl-Space." does not cause issues
- `@codemirror/commands` 6.5.0 → 6.7.1
    - [ ] Verify "Bind Shift-Enter to the same command as Enter in the default keymap, so that it doesn't do nothing when on an EditContext-supporting browser." does not cause issues
    - [ ] Verify "The default keymap now binds Ctrl-m (Shift-Alt-m on macOS) to `toggleTabFocusMode`." does not cause issues
    - Noteworthy changes:
       - "Change `toggleBlockCommentByLine` to not affect lines with the selection end right at their start."
       - "Fix a bug in the undo history that would cause it to incorrectly track inverted effects when adding multiple edits to a single history event."
       - "The new `toggleTabFocusMode` and `temporarilySetTabFocusMode` commands provide control over the view's tab-focus mode.
- `@codemirror/lang-css` 6.2.1 → 6.3.1: just completion improvements
- `@codemirror/lang-sql` 6.6.4 → 6.8.0: completion improvements and minor bug fixes
- `@codemirror/lang-markdown` 6.2.5 → 6.3.1 adds `htmlTagLanguage` option, not relevant
- `@codemirror/language` 6.10.1 → 6.10.6: a couple of nice bug fixes
- `@codemirror/legacy-modes` 6.4.0 → 6.4.2: improvements in Common Lisp, Dart, Stylus and Groovy modes
- `@codemirror/search` 6.5.6 → 6.5.8
  - [ ] would be good to do some manual testing for the search function as we rely on `@codemirror/search`:
    - "Fix a bug that put the selection in the wrong place after running `replaceNext` with a regexp query that could match strings of different length."
    - "Fix an issue where `findNext` and `findPrevious` would do nothing when the only match in the document was partially selected."
    - Fix an infinite loop in `SearchCursor` when the normalizer function deletes characters.
- `@codemirror/state` 6.4.1 → 6.5.0: nothing major
- `@codemirror/view` 6.26.3 → 6.35.3: notworthy:
  - "Avoid spurious focus events by not updating the DOM selection when the editor is unfocused but focusable." (this is the one for  https://github.com/jupyterlab/jupyterlab/issues/15795)
   - [ ] "Fix an issue where a tooltip with wrapped content could, in some circumstances, fail to find a stable position due to a cyclic dependency between its width and its position." - yeah, this was super annoying in LSP; there are other tooltip changes, let's check it still works with JupyterLab positioning
   - [ ] "The editor will now, when available (which is only on Chrome for the foreseeable future) use the [`EditContext`](https://developer.mozilla.org/en-US/docs/Web/API/EditContext) API to capture text input." - this was followed by a dozen of bug fixes, looks like a big change
   - [ ] "Fix the default cursor color for dark themes, which was way too dark." - let's check cursor in different themes
   - [ ] "Disable hiding of selection inside focused widgets" - same; for reference:
      image
   - other goodies:
     - "Fix an issue on Chrome where typing at the end of the document would insert a character after the cursor."
     - "Avoid an issue where Chrome would incorrectly scroll the window when deleting lines in the editor."
     - "Fix an issue where `coordsAtPos` would return null when querying a position in a block widget."
     - "Fix a bug where the editor could draw way too big a viewport when not managing its own scrollbar."
     - "Make it easier to move the pointer over a hover tooltip with an arrow by not closing the tooltip when the pointer is moving over the gap for the arrow."

krassowski (approved):
Thank @jtpio, I did some more manual testing and so far all good. I could not test the MacOS-specific keybindings. I think we can release this as alpha and ask community to help.

jtpio:
Nice, thanks @krassowski for doing some testing :+1: 

Sounds good, maybe we can open an issue to track the list above, and ask users to try the new alpha?

krassowski:
Done: https://github.com/jupyterlab/jupyterlab/issues/17080

Issue #17060, opened 2024-12-09
FIM (fill in the middle) support for inline completer on typing


krassowski:
### Problem

_Fill-in-the-middle_ allows to add code in the middle of the code line. This is currently supported when manually invoking the inline completer, but it is not possible for user to get inline completion suggestions automatically when typing in the middle of the line (these are only returned at the end of the line); this is codified in:

https://github.com/jupyterlab/jupyterlab/blob/bb596b37d44cacb7aaec2c09aa5d6c8b43a3a2a8/packages/completer/src/handler.ts#L467-L473

### Proposed Solution

Add an opt-in option allowing for inline completion to be auto-triggered when anywhere in the line.

Of note, the manual invoke is allowed anywhere since https://github.com/jupyterlab/jupyterlab/pull/16298 included `4.1.x`.

### Additional context

Quoting the Codeium [blog post](https://codeium.com/blog/inline-fim-code-suggestions) (emphasis mine):

> This scenario where there is text on both sides of the cursor (on the same line) and we want autocomplete suggestions is a special case of the fill-in-the-middle task called in-line fill-in-the-middle, or in-line FIM for short. **And it is not easy to get in-line FIM to work well enough to be helpful rather than distracting.**
> 
> [...]
> So what is likely going on with these AI code assistants? In the original post, we showed how TabNine didn’t have generic FIM capabilities in the first place. Amazon CodeWhisperer on face value does, but they must have made some significant fundamental errors in building FIM since the suggestions are quite poor, error-prone, and often not referencing post-cursor context.
> 
> But what about Github Copilot? They have reasonable FIM suggestions for filling in lines and using future context, so why is in-line FIM not enabled? Honestly, we are not sure. Copilot has normal FIM suggestions and it seems like there shouldn’t be anything that is stopping it from exposing in-line FIM suggestions. Our guess is that they tried to enable in-line FIM, but the performance wasn’t good enough, maybe from a lack of in-line FIM-like training examples or some other training knobs. All we know is that for now, Copilot has not enabled in-line FIM in production, and likely that means there is something up with the underlying model, which they are unwilling to retrain, or their FIM handling capabilities are inherently worse.

This appears like a good enough rationale for making it opt-in in JupyterLab rather than opt-out, especially because compared to GitHub Copilot or Codeium, JupyterLab does not have a set of curated supported models for which pre-processing and context provision is fine-tuned to return good enough in-line FIM to be more useful than distracting.

---

[...]

Issue #17011, opened 2024-11-28
Accepted inline completion trigger inline completion at the wrong position


fcollonval:
## Description

Accepting an inline completion triggers indirectly an inline completion due to `textChanged` signal.

## Reproduce

_custom provider_

https://github.com/user-attachments/assets/6fa7342c-b716-490c-8814-f234090df29c

_jupyter-ai with OpenAI_

https://github.com/user-attachments/assets/c6aca399-7853-4f2f-a56b-b124418d6f22

In jupyter-ai, I was not able to reproduce the bug -> more on that later.

Reduce sequence flows:

- Accepting the inline completion, trigger a source changed event from Yjs

https://github.com/jupyterlab/jupyterlab/blob/48908a8b07128b651faab72434f8d442b99bc91c/packages/completer/src/inline.ts#L141-L148

- This trigger `onTextChanged`

https://github.com/jupyterlab/jupyterlab/blob/8937ac414ae9999899a3588dc41917fc016a8c78/packages/completer/src/handler.ts#L329

- That triggers a `_makeInlineRequest`

https://github.com/jupyterlab/jupyterlab/blob/8937ac414ae9999899a3588dc41917fc016a8c78/packages/completer/src/handler.ts#L362-L367

But the cursor position at line 364 in the above snippet is prior to the inline text as the cursor position is changed line 147 in the first snippet after changing the editor source.

> The demo with jupyter-ai is not plagued with this bug because inline completion is only allowed at line ending:

https://github.com/fcollonval/jupyterlab/blob/8937ac414ae9999899a3588dc41917fc016a8c78/packages/completer/src/handler.ts#L473

But actually the inline request is also started.

## Expected behavior

Accepting inline completion do not trigger inline completion (or at least not at the cursor position prior to the first inline completion).

## Context

JupyterLab 4.3.1

jupyter_ai                2.28.2
jupyter_ai_magics         2.28.2
langchain                 0.2.17
langchain-community       0.2.19
langchain-core            0.2.43
langchain-openai          0.1.25
langchain-text-splitters  0.2.4
openai                    1.55.2

fcollonval:
@krassowski I'm not deeply familiar with that code. It seems a fix would be to not trigger `onTextChanged` in `CompletionHandler._onSharedModelChanged` if `CompletionHandler.inlineCompleter` has an attribute saying it is fulfilling a completion.


[...]

PR #17067, opened 2024-12-10
Add option for automatic Fill-in-the-Middle inline completion

Darshan808:
### References
#17060 

This PR introduces a new settings option that allows users to enable "fill-in-the-middle" functionality for inline completion providers. Users can navigate to the settings-editor->Inline completer and check the "Fill in Middle" checkbox for each provider. When enabled, the provider will automatically trigger inline completions even when typing in the middle of a line.

#### Changes:
1. Added a new setting for each inline completion provider:
   - "Fill in Middle": A checkbox that users can toggle to enable or disable automatic completions in the middle of a line.
2. Updated the inline completion logic to respect the new setting:
   - If "Fill in Middle" is enabled, inline completions are automatically triggered, even in the middle of a line.
   - If disabled, the default behavior remains (completions are triggered only at the end of a line).

#### Screenshots

image

image

- Note: I’m happy to write tests once the feature is finalized. Please share your feedback on the changes or let me know if anything needs to be approached differently.
---


packages/completer-extension/src/index.ts
```diff
@@ -316,6 +317,14 @@ const inlineCompleter: JupyterFrontEndPlugin<void> = {
                       provider.name
                     ),
                     type: 'boolean'
+                  },
+                  fillInMiddle: {
+                    title: trans.__('Fill in middle'),
+                    description: trans.__(
+                      'Whether to show completions in the middle of the code line from %1 provider.',
+                      provider.name
+                    ),
```
krassowski, line 326: I like this. It is a better choice over placing this as an option for the entire inline completer (like [these](https://github.com/jupyterlab/jupyterlab/blob/b057cc8d2b12bfaa84470c5ce536c8dd2b8277dd/packages/completer-extension/schema/inline-completer.json#L43-L102)) because:
- some providers will support fill-in-the-middle and some won't
- providers can set a default for themselves (declare support for fill-in-the-middle)

It looks like this is a generic option for both `auto` and `manual` trigger. This can be problematic for two reasons:
- manual trigger is already allowed in the middle of the line; this PR as-is would effectively disable it by default, breaking some existing use cases
- there may be providers which have some support for fill-in-the-middle, but which is not great; in these cases users may want to be able to trigger it manually in the middle of the line but may not want it to show up automatically.

What do you think about the following:

```suggestion
                  autoFillInMiddle: {
                    title: trans.__('Fill in middle on typing'),
                    description: trans.__(
                      'Whether to show completions in the middle of the code line from %1 provider on typing.',
                      provider.name
                    ),
```

packages/completer/src/reconciliator.ts
```diff
@@ -50,15 +50,19 @@ export class ProviderReconciliator implements IProviderReconciliator {
 
   fetchInline(
     request: CompletionHandler.IRequest,
-    trigger: InlineCompletionTriggerKind
+    trigger: InlineCompletionTriggerKind,
+    isMiddleOfLine: boolean
```
krassowski, line 54: To match the public API definition:

```suggestion
    isMiddleOfLine?: boolean
```

Darshan808:
A bug is seen with this FIM feature where the `Tab` key needs to be pressed twice to complete both end-of-line and middle completions. Investigating this issue!

#### Suggestion
image

#### Single tab
Screenshot from 2024-12-10 16-08-51

#### Next tab
image

krassowski (commented):
The tests could go into:

https://github.com/jupyterlab/jupyterlab/blob/6e1174e202b6bf5a7b725d59b5ddafc5f0afbed2/packages/completer/test/inline.spec.ts#L21

and/or:

https://github.com/jupyterlab/jupyterlab/blob/6e1174e202b6bf5a7b725d59b5ddafc5f0afbed2/galata/test/jupyterlab/inline-completer.test.ts#L21

It would make most sense to adjust the history provider so that it produces useful results with FIM. The relevant code is here:

https://github.com/jupyterlab/jupyterlab/blob/6e1174e202b6bf5a7b725d59b5ddafc5f0afbed2/packages/completer/src/default/inlinehistoryprovider.ts#L115-L143

krassowski:
> A bug is seen with this FIM feature where the `Tab` key needs to be pressed twice to complete both end-of-line and middle completions. Investigating this issue!

This sounds like https://github.com/jupyterlab/jupyterlab/issues/17011 - it can be tackled in a separate PR, but ideally before we merge this one.

packages/completer/src/reconciliator.ts
```diff
@@ -51,15 +51,19 @@ export class ProviderReconciliator implements IProviderReconciliator {
...
   ): Promise<InlineResult>[] {
     let promises: Promise<
       IInlineCompletionList<CompletionHandler.IInlineItem>
     >[] = [];
     const current = ++this._inlineFetching;
     for (const provider of this._inlineProviders) {
       const settings = this._inlineProvidersSettings[provider.identifier];
-      if (isMiddleOfLine && !settings.fillInMiddle) {
+      if (
+        trigger !== InlineCompletionTriggerKind.Invoke &&
```
Darshan808, line 63: > manual trigger is already allowed in the middle of the line; this PR as-is would effectively disable it by default, breaking some existing use cases

I tried to solve this by not skipping FIM when completer is manually invoked

galata/test/jupyterlab/inline-completer.test.ts
```diff
@@ -251,6 +251,47 @@ test.describe('Inline Completer', () => {
...
+      await page.keyboard.type('g');
+      await expect(ghostText).toHaveText(/gestion.*/); //sug(gestion)n
+      await page.keyboard.press('Escape');
+      await page.keyboard.press('ArrowRight');
+      await page.keyboard.press('Backspace');
+      await page.keyboard.type('q'); //sugq|
+      await page.keyboard.press('ArrowLeft');
+      await page.keyboard.press('Backspace');
+      await page.keyboard.type('g');
+      expect(ghostText).toBeHidden(); //Hidden on sug|q
```
krassowski, line 292: Highlighting linter error:

> Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator


```suggestion
      wait expect(ghostText).toBeHidden(); //Hidden on sug|q
```

galata/test/jupyterlab/inline-completer.test.ts
```diff
@@ -251,6 +251,47 @@ test.describe('Inline Completer', () => {
...
+      const cellEditor = (await page.notebook.getCellInputLocator(2))!;
+      await cellEditor.click();
+      await page.keyboard.press('u');
+
+      // Ghost text shows up
+      const ghostText = cellEditor.locator(GHOST_SELECTOR);
+      await ghostText.waitFor();
+
+      await page.keyboard.type('n');
+      expect(ghostText).toBeHidden(); //Hidden on sun|
```
krassowski, line 281:
```suggestion
      await expect(ghostText).toBeHidden(); //Hidden on sun|
```

galata/test/jupyterlab/inline-completer.test.ts
```diff
@@ -251,6 +251,47 @@ test.describe('Inline Completer', () => {
...
+      await page.evaluate(async () => {
+        await window.jupyterapp.commands.execute('settingeditor:open', {
+          query: 'Inline Completer'
+        });
+      });
+      await page
+        .locator(
+          '#jp-SettingsEditor-\\@jupyterlab\\/completer-extension\\:inline-completer_providers_\\@jupyterlab\\/inline-completer\\:history_autoFillInMiddle'
+        )
+        .click();
```
krassowski, line 267: This locator should use user-facing or accessibility-oriented features. Like select by having text matching "Fill in middle on typing". This means that if implementation of the settings editor changes (e.g. we update `rjsf`) then the test will still work.

galata/test/jupyterlab/inline-completer.test.ts
```diff
@@ -251,6 +251,47 @@ test.describe('Inline Completer', () => {
...
+      await page.sidebar.close();
+
+      await page.locator('#tab-key-2-1').click();
```
krassowski, line 270: Not sure what it does? Can you find a selector that does not depend on implementation details?

packages/completer/src/default/inlinehistoryprovider.ts
```diff
@@ -72,6 +72,11 @@ export class HistoryInlineCompletionProvider
 
     const multiLinePrefix = request.text.slice(0, request.offset);
     const linePrefix = multiLinePrefix.split('\n').slice(-1)[0];
+    const multiLineSuffix = request.text.slice(
+      request.offset,
+      request.text.length
+    );
```
krassowski, line 78: I think `indexEnd` can be omitted here:


```suggestion
    const multiLineSuffix = request.text.slice(request.offset);
```

packages/completer/src/default/inlinehistoryprovider.ts
```diff
@@ -72,6 +72,11 @@ export class HistoryInlineCompletionProvider
...
+    const suffix = multiLineSuffix.split(' ')[0];
```
krassowski, line 79: It is not obvious why it splits on a whitespace and not on new line.

krassowski: If there is suffix, should we remove it from suggestion?

https://github.com/jupyterlab/jupyterlab/blob/6e1174e202b6bf5a7b725d59b5ddafc5f0afbed2/packages/completer/src/default/inlinehistoryprovider.ts#L131-L140

packages/completer/src/default/inlinehistoryprovider.ts
```diff
@@ -117,7 +122,7 @@ export class HistoryInlineCompletionProvider
         output: false,
         raw: true,
         hist_access_type: 'search',
-        pattern: linePrefix + '*',
+        pattern: linePrefix + '*' + suffix + '*',
```
krassowski, line 125: If there is not suffix it becomes `linePrefix + '**'. I am not sure if this will behave as expected, I would only add the second `*` if suffix is indeed there.

krassowski (commented):
Great work! It looks like the test is failing. I left some comments on the code.

Darshan808:
> It is not obvious why it splits on a whitespace and not on new line.

`impo|rt numpy as np`

My approach focuses only on the word immediately after the cursor (rt in this case).

Would it be more beneficial to match just the immediate word after the cursor, or is it better to include the entire line?

Darshan808:
> If there is suffix, should we remove it from suggestion?

If we use the following pattern:
`pattern: linePrefix + '*' + suffix + '*',`

In cases like `pri|("H` the suggestion could be  `nt("Hello World")`  which results `pri|_nt("Hello World")_("H` making it challenging to remove just the suffix since everything after the suffix is being considered as well.

However, if we simplify the pattern to:
`pattern: linePrefix + '*' + suffix`
It is possible

Which approach do you think we should adopt or do you have any other thoughts?

Darshan808:
> This sounds like https://github.com/jupyterlab/jupyterlab/issues/17011 - it can be tackled in a separate PR, but ideally before we merge this one.

Yes, it appears to be the same issue. fcollonval  mentioned  he will be working on a fix. Should I leave it for now, or is there still room for me to work on it?

krassowski:
> `impo|rt numpy as np`
> Would it be more beneficial to match just the immediate word after the cursor, or is it better to include the entire line?

But for `impo|rt numpy as np` there are not many useful FIM suggestions? Unless I miss something.

When thinking about fill in the middle the use cases that come to mind are:
- I have a function definition, say `def add_numbers(a, b):` but after typing it out I realised I was using a different naming patter previously; my history includes `def add_together(a, b):` and `def power(a, n):`; I go back with the cursor, delete `add_numbers` so that my cursor is on `def |(a, b)` and invoke inline completer; it should now suggest `add_together`; here including the entire line helps it choose the best suggestion. Maybe it should be able to match based on `def *(a,` but then sort by the best match?
- I have a variable assignment `x = 3.14`; my history remembers the variables I assigned before (say `y = 2`, `z = 3`). I realised that I wanted to name my variable `y` instead of `x`; I go back, remove `x` and trigger when line looks `| = 3.14`. In this case I expect it to suggest all variable assignments so matching only up to the first word after the whitespace. However, if I had `pi = 3.14` in history, I would expect it to suggest `pi` with the highest priority.

So I think the typical use case is:
- search for `linePrefix + '*' + nextWordSuffix` where `nextWordSuffix` is like  `(a,` or ` =`
- sort by how good is the prefix match

> However, if we simplify the pattern to:

Sounds good. I think we should aim to support minimum viable use case right now and make sure that this gives user nice results, rather than try to provider more suggestions which later require manual editing after accepting them.

krassowski:
> In cases like `pri|("H` the suggestion could be  `nt("Hello World")`  which results `pri|_nt("Hello World")_("H` making it challenging to remove just the suffix since everything after the suffix is being considered as well.

I think it should only suggest `nt` in this case.

Darshan808:
After extensive testing and reviewing various examples and advice from above, I believe the best approach is to search for `pattern: linePrefix + '*' + suffix + '*'`, where the suffix represents the entire line the cursor is on. Then, remove the suffix and everything following it to generate suggestions.
This method provides useful results.

Darshan808:
The newline character ('\n') at the end of each suggestion was somewhat inconvenient, so I removed it from all types of suggestions. However, the test snapshot still includes it. If this change is approved, we can update the snapshot to align with the revised code, which no longer includes the newline at the end.

krassowski:
Yes, removing the trailing new line and updating the snapshot makes sense to me.

krassowski:

krassowski (approved):
Thank you @Darshan808!

PR #17070, opened 2024-12-10
Update to `typedoc` 0.27.4

bollwyvl:
## References

- for #16768
- continues (and includes) #17006

## Code changes

- [x] bump `typedoc` and `typedoc-plugin-mdn-links` 
- [x] deduplicate `yarn.lock`
- [x] add more redirects
  - [x] moves the fixing from #17006 closer to the `typedoc` invocation, so it is picked up for `check_links`
- [x] ignore built typedoc HTML, JS, CSS in linters

## User-facing changes

- :crying_cat_face: package namespaces `-` in the name now use that instead of `_`
  - while probably _better_... this means old links will break unless we backfill all the pages with redirects
  - in the meantime, redirect from the old URLs to the new URLs
    - the cost is very low to maintain redirects, but could be dropped in a future `4.x`

## Screens

| +/- | note | screen (0.27 vs 0.26, sorry) |
|-|-|-|
| :tada: | GitHub-flavored admonitions are handled better in 0.27 | image |
| :shrug: | Module listings are collapsible, but not as compact on wide screens |  image
| :shrug:  | namespaces no longer have their source | image

- n/a

## Future Work

Not on this PR:

- check the _actual_ HTML we generate and serve
- consider supporting GitHub-flavored admonitions in user Markdown if they are adopted by more platforms
  - though apparently they are a bit of an a11y/i18n fail


bollwyvl:
Ah, I just realized that the `check_links` action is checking _source_ files not _built_ HTML... while useful for visitors to GitHub, this doesn't really help the _published_ much, which is arguably more important, and why we have broken links on the site today. I guess, but my recommendation would be to:
- link check all the `README.md`
- don't bother with anything in `docs/source`
- link check all of the built pages in `docs/_build`
... issue for another day, I guess, as `check_links` won't even do that today, and `PYTEST_ADDOPTS` won't do it either. Sigh.

Anyhow: have moved all the typedoc URL maintenance into a dedicated `scripts`, and added more comments. This gets called from `jlpm docs` immediately after typedoc, so should appease `check_links`, work during `make`, etc.

krassowski (approved):
Thank you!

Issue #16801, opened 2024-09-22
In dark theme the "Add" button in keyboard shortcuts UI is not readable


krassowski:
## Description

In dark theme the "Add" button in keyboard shortcuts UI is not readable

## Reproduce

1. Change theme to "JupyterLab Dark"
2. Open Settings
3. Open "Keyboard Shortcuts"
4. Hover over a row
5. See

![image](https://github.com/user-attachments/assets/4449a329-c853-4f93-8b97-a0d22aac79c1)

## Context

- Browser and version: Chrome
- JupyterLab version: 4.3.0b1

krassowski:
The existing shortcut itself when edited has poor contrast:

![Uploading image.png…]()

Gannu456:
Can i work upon this, if it is still open?

krassowski:
Yes, it is still open, please do open a PR :)

omeizan:
Hi, I'm not sure if this has been worked on yet, can I work on this?

krassowski:
Yes, you can. No one opened a PR so far.

Madhu3456:
Hi, after trying some permutations of colors I came up with this, can someone verify?

![Screenshot from 2024-12-10 21-28-46](https://github.com/user-attachments/assets/0f12ac9c-f9be-4d50-83fa-1cd7b4a41106)


---On hover
![Screenshot from 2024-12-10 21-28-49](https://github.com/user-attachments/assets/9b0465bf-6e79-4ec9-9d89-12d0a67646aa)

krassowski:
Is the top one the normal state? If so the contrast looks too weak to me. We aim for WCAG AA contrast you can calculate it https://webaim.org/resources/contrastchecker/

Madhu3456:
Yes, the top one is the normal state and I feel it's weak as well. I will use the resource.

Madhu3456:
To add a color(--md-blue-700) that is brighter than "--jp-brand-color0"(--md-blue-900) should I change the exiting variable names from 0 to 1, 1 to 2 and so on and mark the new color as --jp-brand-color0 ? If so I'll also have to make changes in all the places these are used?

https://github.com/jupyterlab/jupyterlab/blob/2607c1093539850b5b618352dc94893a8b662e48/packages/theme-light-extension/style/variables.css#L219

krassowski:
[...]

PR #17074, opened 2024-12-11
Fix contrast issues for keyboard shortcut 'Add' button 

bsul24:
## References
[#16801 ](https://github.com/jupyterlab/jupyterlab/issues/16801)

## Code changes

The following changes were made in packages/shortcuts-extension/style/base.css to improve the accessibility and contrast of elements in the dark theme:

- Updated .jp-Shortcuts-Plus background and border-color.
- Adjusted hover and active states background of .jp-Shortcuts-Plus.
- Modified the .jp-mod-selected-InputText background color.

These updates ensure better visibility and compliance with WCAG AA contrast standards.

## User-facing changes

Users will notice:

- Improved visibility for the "Add" button in the keyboard shortcuts UI and selected shortcuts that already exist while in dark theme.

### Before
image
image

### After
Normal
image
Hover
image
image

None

### DIFF
--- a/packages/shortcuts-extension/style/base.css
+++ b/packages/shortcuts-extension/style/base.css
@@ -68,7 +68,7 @@
 }
 
 .jp-mod-selected-InputText {
-  background-color: var(--jp-brand-color3);
+  background-color: #004080;
   overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


krassowski:
This looks very nice in the dark theme (thank you!). How does it look in the light theme?

We would ideally use CSS variables to limit color palette and make it easy to develop third-party themes.

I would suggest (in order of preference):
- find best matching CSS variables among the existing ones, ideally with appropriate semantic meaning (e.g. `--jp-accept-color-normal`) that match your intended colors in dark theme as close as possible,
- if these variables do not work great in light mode, let's define new variable like `--jp-shortcuts-accept-color-normal` scoped to the shortcuts component and add an override in the dark theme only.

HeheAnanya:
Hi! Is the issue still open?

PR #17078, opened 2024-12-11
Copy edits in `CONTRIBUTING.md`

JasonWeill:
Read the Docs is now updated automatically, so we don't need to link to the source file as well as to the docs. Corrects capitalization of JupyterLab.


krassowski (approved):
Good catch, thank you @JasonWeill!

PR #17082, opened 2024-12-12
Fix/17011-ai-completion-skip-completion

### LINKED ISSUE #17011: Accepted inline completion trigger inline completion at the wrong position

fcollonval:
## References

Fixes #17011

## Code changes

Apply inline completion suggestion and cursor position update in the same codemirror transaction to avoid document change listener react with a wrong editor position.

This is therefore done using codemirror mechanism and not the shared model.

> I checked that the changes are still applied as single change for the undo manager.

## User-facing changes

Continuous inline completion works as expected.

None

### DIFF
--- a/packages/completer/src/handler.ts
+++ b/packages/completer/src/handler.ts
@@ -2,13 +2,6 @@
 // Distributed under the terms of the Modified BSD License.
 
 import {
-  CodeEditor,
-  COMPLETER_ACTIVE_CLASS,
-  COMPLETER_ENABLED_CLASS,
-  COMPLETER_LINE_BEGINNING_CLASS
-} from '@jupyterlab/codeeditor';
-import { Text } from '@jupyterlab/coreutils';
-import {
   CellChange,
   FileChange,
   ISharedBaseCell,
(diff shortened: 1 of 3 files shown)


krassowski:
The failing test appears to be relevant:

```
  ● @jupyterlab/completer › CompletionHandler › #onCompletionSelected() › should be undoable and redoable

    - Expected  - 1
    + Received  + 1

      eggs
    - foo # comment
    + foobar # comment
      baz

      324 |         // Undo the completion, check its value and cursor position.
      325 |         editor.undo();
    > 326 |         expect(editor.model.sharedModel.getSource()).toBe(text);
          |                                                      ^
      327 |         expect(editor.getCursorPosition()).toEqual({
      328 |           line,
      329 |           column: column + 3

      at Object.<anonymous> (test/completer/handler.spec.ts:326:54)
```

fcollonval:

The issue was an improper configuration of the undoManager in the tests. I took the chance of this PR to update that for all tests not only the one failing earlier - let me know if I should leave alone the other files.

> I added a test for undoing the inline completion in the latest commit.

krassowski (approved):
Thank you @fcollonval!

PR #17085, opened 2024-12-13
Update to Playwright 1.49

jtpio:
## References

https://github.com/microsoft/playwright/releases/tag/v1.49.0

Some accessibility features that may be useful.

## Code changes

- [x] Update `@playwright/test` to 1.49

## User-facing changes

None

None

### DIFF
--- a/galata/package.json
+++ b/galata/package.json
@@ -54,7 +54,7 @@
     "@jupyterlab/services": "^7.4.0-alpha.2",
     "@jupyterlab/settingregistry": "^4.4.0-alpha.2",
     "@lumino/coreutils": "^2.2.0",
-    "@playwright/test": "^1.48.0",
+    "@playwright/test": "^1.49.1",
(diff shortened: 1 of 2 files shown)


jtpio:
UI tests failures appear to be consistent after several re-runs.

krassowski:
Interestingly these two are tests which were flaky previously. Maybe they now just consistently fail. I think these two tests need more work, regardless of playwright upgrade. I am slightly more concerned about the number of flaky tests in general.

jtpio:
CI seems to now be passing after a rebase.

Technically it's likely ready to go, but maybe we want to address the flaky tests first in case this increases the frequency of failures? Or we go with this and see if this is the case in practice, and try to fix the flaky tests soon after.

krassowski (approved):
Thanks!

jtpio:

Issue #17087, opened 2024-12-13
Add linter rules for regular expressions


bollwyvl:
### Problem

Recently on #15733, we spent a bit of time going around and around with CodeQL about some bad regular expressions.

It's... pretty opaque what it's doing, as it's implemented in a couple layers of their bespoke language, and the current workflow basically means push, wait for a docker image, scan all 1k filees, find the job run in the cast of hundreds, dig through the findings, and then hope that the message is helpful.

### Proposed Solution

These kinds of errors are fiddly, and having per-keystroke feedback would improve the experience and help us ship better code.

One approach would be to add eslint rules such as [eslint-plugin-regexp](https://github.com/ota-meshi/eslint-plugin-regexp) ([docs](https://ota-meshi.github.io/eslint-plugin-regexp/)) which will: 
- get picked up by an editor (given an appropriate plugin) 
- get run as documented in local workflows (`jlpm lint`)

This won't _exactly_ match what CodeQL, but does run locally, and in my case was able to help me fix my issue (specifically by adding some extra `\b` word boundaries) and encouraging use of `String.matchAll` instead of the peril-fraught `while(x = re.exec(txt))` (_don't forget the `/g` flag!_). 

### Additional context

The rough plan would be:

- on an initial PR
  - just add the plugin with its recommended settings
  - observe the errors on CI
    - spoliers: `✖ 62 problems (27 errors, 35 warnings)`
  - add explicit ignores for those cases (either by comment, or in the config)
- fix errors however it makes sense (rule by rule)

Some disadvantages:
- new [dependency and _its_ dependencies](https://bundlephobia.com/package/eslint-plugin-regexp@2.7.0)
- eslint is already pretty slow, especially on a cold run, this will no doubt make it slower

Some alternatives:
  - do nothing, i guess
  - encourage use of a codeql IDE plugin, but this would send your local code to somewhere

### Future work

It has been mentioned in some calls, but downstreams such as the copier, notebook, jupyterlite, etc. could benefit from a published `eslint-config-jupyterlab` which could include our set of plugins and a sensible, repo-agnostic configuration (no paths). This might simplify using local tools more effectively, which can be a better local experience than push-and-pray. This would also make an easier box to tick to get code in extensions outside of the core repo/org into form for publishing and maintaining.

krassowski:
[...]

PR #17089, opened 2024-12-13
Run eslint-plugin-regexp

bollwyvl:
## References

- for #17087
- includes #17088

## Code changes

- runs `(prettier && eslint) || (prettier && eslint)`

## User-facing changes

- n/a

- n/a


bollwyvl:
@krassowski of note, there is the precursor #17088 (which i just naively merged up to `main`) which doesn't actually run the fixer.

PR #17092, opened 2024-12-16
Add Content Provider API

### LINKED ISSUE #16717: Allow a drive to support multiple document providers

krassowski:
## References

- Closes #16717
- Builds on and supersedes #16744
- Together with https://github.com/jupyterlab/jupyter-collaboration/pull/418 will enable fixing highly requested https://github.com/jupyterlab/jupyter-collaboration/issues/202

## Code changes

As in #16744
> - adds the `ContentProviderRegistry` and `RestContentProvider` classes. The default `Drive` now doesn't directly access files and directories using the Jupyter REST API , but through its content provider registry, which has a REST content provider by default. Extensions like `jupyter-collaboration` can register new content providers for the drive, for instance an RTC content provider.
> - has the side effect of removing the `RTC:` prefix in `jupyter-collaboration`, since it won't need a custom drive anymore, but it will instead register an RTC content provider on the default drive.

Differently to #16744:
- does not add another pattern matching logic
- allows to associate a content provider with widget factory, as if one wants to use a super set of Contents API, they will need to implement a widget which makes use of these anyways (think the "Streaming CSV viewer")
- exposes `FileEditorFactory` widget factory (the same way as `NotebookWidgetFactory` was already exposed for 4 years since PR [#8066](https://github.com/jupyterlab/jupyterlab/pull/8066)), to allow extensions to register custom factories based of these.

<details>

<summary>Rationale and overview of Content Providers Idea</summary>

### RTC

Content Provider needed to:
- provide a custom `sharedModelFactory` for all files
   - except for binary files?
- override `ContentsManager` methods
   - `get(path, options)` - to provide content from the `sharedModel` rather than from the jupyter-server Contents API
   - `save(path, options)` - to prevent sending `save` to jupyter-server Contents API as saving is handled by RTC, but this is strictly no longer needed since https://github.com/jupyterlab/jupyterlab/pull/16900
- be able to emit `fileChanged` to update file if collaborator changes it

### Streaming CSV widget

Use case: right click on a large CSV in file browser, choose "Open with" → "Large CSV Viewer", allowing to view files larger than the browser/jupyter-server can handle at once.

Real world applications: [genome browsers](https://en.wikipedia.org/wiki/Genome_browser), geospatial browsers

Content Provider needed to:
- provide a custom `sharedModelFactory` for files opened with this widget to allow fetching content selectively
- probably override `ContentsManager.get()` so that the full content is never fetched as it would freeze the browser

Entry point:
- passing a custom content provider in `DocumentRegistry.IWidgetFactory`

### Jupytext

Jupytext does not change the frontend contents API, but has its own server-side ContentsManager subclass. Here I explore what would be needed if someone wanted to make a jupytext a frontend-only extension, and if content providers mechanism could help.

Use case:
- right click on a text file with `.md` extension, choose "Open with" → "Jupytext Notebook"
- synchronise notebook to/from multiple formats

Requirements:
- override contents methods (by routing them to a custom content manager on backend), e.g.:
  - `get()` so that it returns something that quacks like a Notebook
  - `save()` to save paired notebook representations at once
  - `new_untitled()` to allow creating notebooks with `ext`
  - `rename_file()` to allow renaming all paired notebooks
- should compose well with RTC

Entry point:
- passing a custom content provider in `DocumentRegistry.IWidgetFactory`

### Only `get` and `save` methods in Content Provider

Only `get` and `save` methods are mirrored in content providers (reflecting RTC use case at the time when this PR was written) - but drives mirror all the `ContentsManager` methods

Maybe the content provider should have the same API as `IDrive` excluding:
   - drive prefix
   - content provider registry

### No "Open with" support

This PR does not allow to choose the provider based on widget factory (yet).

### Redundancy of path/extension routing logic

As pointed by @fcollonval, `DocumentRegistry` already implements a dynamically dispatch depending on path/extension. We should not duplicate that logic or if we do we should harmonize the two implementation.

### Composition

Content Providers idea does not help with composition. This is why implementing jupytext using content providers may not be a good idea if simultaneous RTC support is desired; instead `jupytext` might need a way to transform content before it is loaded into RTC/when RTC saves.

### Moving files between drives

The content provider would need to be reset. This can be addressed in a follow-up.

### One content provider per drive

Content provider will usually work in context of a single drive (if we implement streaming in jupyter-server extension and add a content provider to interface with it, it will not provision content from Google Drive), but `IWidgetFactory` then needs to be only shown in "Open with" for drives which support required content provider.

How can we support multiple drives for "Open with"? Should we register content providers on drives, or should content provider be able to provision content on multiple drives?

IMO it should be one content provider registered per each drive; maybe multiple content providers could be associated with the same token, like in:

```ts
defaultDrive.contentProviders.register({
  id: 'streaming-csv',  // or a token object instead of ID
  provider: new StreamingContentProvider(defaultDrive),
})
googleDrive.contentProviders.register({
  id: 'streaming-csv',  // or a token object instead of ID
  provider: : new StreamingGoogleDriveProvider(googleDrive),
})
// then when registering "streaming CSV" widget, in `DocumentRegistry.IWidgetFactory`:
{
  contentProviderId: 'streaming-csv'
}
```

### Provision of directories

Another potential difference between `ContentProvider` and `Drive` is that drive influences the listing in the file browser, but it is not clear if `ContentProvider` should do that too. In fact, when registering with `DocumentRegistry.IWidgetFactory` it would not have impact on the file browser at all.

Technical note: file browser calls `ContentsManager`/`Drive` directly, but for opening files the operations are routed through `DocumentRegistry.IContext` object where we could swap the content provider for the `DocumentRegistry.IWidgetFactory` scenario.

### Competition between multiple extensions and delayed initialization

@fcollonval pointed out in @Zsailer's PR [#16141](https://github.com/jupyterlab/jupyterlab/pull/16141) which aimed to simply allow extensions to swap the `defaultDrive` that such a change could lead to bad side effects if any extension can swap it on runtime. While we could make sure that this change is only allowed once by setting a simple counter flag, it would still mean that on initial load a different drive is used than after loading the RTC plugin.

https://github.com/jupyterlab/jupyterlab/pull/16794 by @jtpio (building on @fcollonval and @afshin work) explores a way to addresses this by splitting the extension loading into two steps.

**A new possibility opens up with** `DocumentRegistry.IWidgetFactory`: we could have RTC register "Open with" handlers for notebooks and for files which would include the appropriate content provider. This also has the benefit of making sure that binary (e.g. PNG files) would not be open collaboratively until these are explicitly supported. RTC could simply register these "editors"/"viewers" as default for Notebook/Text File types. Of course the caveat is that the default "editors"/"viewers" could still be accessed (unless we hide them) potentially confusing users (so we probably want a way to hide them?).

Another issue is that jupytext already has custom "editors"/"viewers" accessible with "Open with" for text files which render them as if those were notebooks. But maybe that's a good thing as the conflict between RTC and jupytext would no longer be implicit but explicit?

</details>

## User-facing changes

None in this PR, but with https://github.com/jupyterlab/jupyter-collaboration/pull/418 the `RTC:` prefix will go away.

None


davidbrochart (commented):
It's unclear to me how this PR deals with multiple content providers: does the user have a choice to use one or the other to open a file? Is there a content provider that is used by default?

krassowski:
There is no "global" default content provider, but the default `IDrive` implementation (`Drive`) uses the REST content provider as default. Each drive can have a different default provider, and any number of content providers.

The content providers are not directly exposed in the UI, but if there is a widget factory associated with the provider, the user will be able to choose it via "Open with" or in settings by setting a default viewer.

I added relevant documentation:

| | | 
|--|--|
| image | image |

davidbrochart:
Thanks.
But we can imagine a content provider that is able to open only a given type of file, for instance HDF5. In my original PR, there was a notion of file types associated with a content provider, along with a rank indicating how good the content provider is to open a file type. This seems to have disappeared in this PR, so I'm wondering how this is managed?

davidbrochart:
Awesome! In my original PR the file type part was not well integrated with the existing logic in JupyterLab, and with this PR it is :+1:

docs/source/extension/extension_points.rst
```diff
@@ -1135,3 +1135,182 @@ Occasionally, LSP extensions include a CodeMirror extension to modify the code e
...
+      });
+    }
+
+    get contentProviderId(): string {
+      return 'my-custom-provider';
+    }
+  }
+
+  const customContentProviderPlugin: JupyterFrontEndPlugin<void> = {
+    id: 'my-extension:custom-content-provider',
```
krassowski, line 1276:
```suggestion
  const widgetFactoryPlugin: JupyterFrontEndPlugin<void> = {
    id: 'my-extension:custom-widget-factory',
```

docs/source/extension/extension_points.rst
```diff
@@ -1135,3 +1135,182 @@ Occasionally, LSP extensions include a CodeMirror extension to modify the code e
...
+    ): ExampleDocWidget {
+
+      return new ExampleDocWidget({
+        context,
+        content: new ExamplePanel(context)
+      });
+    }
+
+    get contentProviderId(): string {
+      return 'my-custom-provider';
```
krassowski, line 1271:
```suggestion
      // Instructs the document registry to use the custom provider
      // for context of widgets created with `ExampleWidgetFactory`.
      return 'my-custom-provider';
```

packages/fileeditor-extension/src/index.ts
```diff
@@ -96,14 +98,66 @@ const plugin: JupyterFrontEndPlugin<IEditorTracker> = {
...
-    IToolbarWidgetRegistry,
     ITranslator,
     IFormRendererRegistry
   ],
   provides: IEditorTracker,
   autoStart: true
 };
 
+/**
+ * The editor tracker extension.
```
krassowski, line 109:
```suggestion
 * The widget factory extension.
```

packages/fileeditor/src/tokens.ts
```diff
@@ -21,3 +21,11 @@ export const IEditorTracker = new Token<IEditorTracker>(
   Use this if you want to be able to iterate over and interact with file editors
   created by the application.`
 );
+
+/**
+ * The notebook widget factory token.
```
krassowski, line 26:
```suggestion
 * The editor widget factory token.
```

krassowski (commented):
Small documentation improvements

packages/services/src/contents/index.ts
```diff
@@ -1576,3 +1625,236 @@ namespace Private {
     return extension;
   }
 }
+
+class ContentProviderRegistry implements IContentProviderRegistry {
```
jtpio, line 1629: Maybe we can add some docstrings to this class and methods, for consistency with the rest of the code base?

packages/services/src/contents/index.ts
```diff
@@ -1576,3 +1625,236 @@ namespace Private {
...
+   *
+   * @param identifier - The identifier of the provider; it can be reused between drives.
+   * @param provider - The content provider to register.
+   */
+  register(identifier: string, provider: IContentProvider): IDisposable;
+
+  /**
+   * Get a content provider matching provided identifier.
+   *
+   * If not identifier is provided, return the default provider.
```
jtpio, line 1821:
```suggestion
   * If no identifier is provided, return the default provider.
```

packages/services/src/contents/index.ts
```diff
@@ -1576,3 +1625,236 @@ namespace Private {
...
+  }
+}
+
+/**
+ * Registry of content providers.
+ * @experimental
+ */
+export interface IContentProviderRegistry {
+  /**
+   * Add a content provider into the registry.
```
jtpio, line 1811:
```suggestion
   * Add a content provider to the registry.
```

jtpio (commented):
Thanks @krassowski for rebooting this PR!

Just had a quick look more focused on the docs and extension developer facing parts, with a few minor suggestions.

I'll try to have a closer look again this week, as well as https://github.com/jupyterlab/jupyter-collaboration/pull/418 to get a better idea of what it looks like downstream and in extensions.

packages/services/src/contents/index.ts
```diff
@@ -1576,3 +1625,236 @@ namespace Private {
...
+     */
+    defaultProvider: IContentProvider;
+  }
+}
+
+/**
+ * A content provider using the Jupyter REST API.
+ */
+export class RestContentProvider implements IContentProvider {
+  identifier: 'rest';
```
jtpio, line 1703: Is this used anywhere?

krassowski: Good catch - nope, it is not.

packages/services/src/contents/index.ts
```diff
@@ -1576,3 +1625,236 @@ namespace Private {
...
+
```
krassowski, line 1704:
```suggestion
```

packages/services/src/contents/index.ts
krassowski: Resolved by https://github.com/jupyterlab/jupyterlab/pull/17092/commits/ae36421391c8302c38c0dde44e057220b44da2fa.

packages/services/src/contents/index.ts
krassowski: Done in 6f337f0eadb2ff82b100686ea413350bef2ada29.

docs/source/extension/extension_points.rst
```diff
@@ -1135,3 +1135,184 @@ Occasionally, LSP extensions include a CodeMirror extension to modify the code e
       });
     }
   };
+
+
+File System
```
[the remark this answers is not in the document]
davidbrochart: "Content Provisioning"?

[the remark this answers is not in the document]
krassowski: Done in f70fa14

jtpio:
Thanks @krassowski! I'll have another look this week.

docs/source/extension/extension_points.rst
```diff
@@ -1135,3 +1135,181 @@ Occasionally, LSP extensions include a CodeMirror extension to modify the code e
...
+    }
+
+    fromString(data: string): void {
+      const chunks = JSON.parse(data) as IMyContentChunk[];
+      return this.fromJSON(chunks);
+    }
+  }
+
+
+For a complete example of a widget factory (although not using a content provider), see the `documents extension example <https://github.com/jupyterlab/extension-examples/tree/main/documents>`__.
```
jtpio, line 1315: Linking to https://github.com/jupyterlab/extension-examples/issues/215 as potentially related.

At some point it will be great to have an example about contents in the `extension-examples` repository. Not sure yet if there should be two distinct examples for both drives and content providers, or only one so it's easier to explain the difference between the two APIs and how they are supposed to be used together.

krassowski: > if there should be two distinct examples for both drives and content providers, or only one so it's easier to explain the difference between the two APIs and how they are supposed to be used together

I think there is a big advantage of having drives and content providers in a single example, but on the other hand we also need a widget factory so that gets a pretty big. Maybe in addition to one such a large example there should be a small `IDrive` example.

jtpio: Sounds good, we can track that and discuss more in the `extension-examples` repo.

jtpio (approved):
Thanks!

For reference I tried these changes locally together with https://github.com/jupyterlab/jupyter-collaboration/pull/418#issuecomment-2565443398 and it seemed to work fine :+1: 

It would be interesting to look into more use cases for this new API for extensions. Maybe a good next step could be to add a new example to the `extension-examples` repo, or update an existing extension to use the new API.

Given this, we could get this in early so it's available in the next alpha, and can be tested more easily downstream.

packages/docregistry/src/default.ts
```diff
@@ -487,6 +488,25 @@ export abstract class ABCWidgetFactory<
...
   }
 
+  /**
+   * Identifier of the content provider required for the widget (if any).
+   *
+   * Throws Error if the content provider was already set.
+   *
+   * @experimental
+   */
+  get contentProviderId(): string | undefined {
```
jtpio, line 498: For reference , this seems to be causing the following issue downstream in Notebook: https://github.com/jupyter/notebook/pull/7559#issuecomment-2573550718

```
Error: ../../node_modules/@jupyterlab/docregistry/lib/default.d.ts(267,9): error TS2380: The return type of a 'get' accessor must be assignable to its 'set' accessor type
  Type 'undefined' is not assignable to type 'string'.
```

Issue #17068, opened 2024-12-10
Search highlight is not applied for matches on Python builtin keywords


krassowski:
## Description

Similar to https://github.com/jupyterlab/jupyterlab/pull/17065 but down to higher specificity of

https://github.com/jupyterlab/jupyterlab/blob/6e1174e202b6bf5a7b725d59b5ddafc5f0afbed2/packages/codemirror/src/theme.ts#L86-L88

added in https://github.com/jupyterlab/jupyterlab/pull/15805

## Reproduce

1. Type `from x import y`
2. Change theme to light/dark/high contrast dark
3. Search for `"m"
4. See poor contrast due to green not being replaced by search highlight color

## Expected behavior

Syntax highlighting is overwritten correctly

## Context

- JupyterLab version: 4.3.2

krassowski:
The relevant CSS is in:

https://github.com/jupyterlab/jupyterlab/blob/2607c1093539850b5b618352dc94893a8b662e48/packages/codemirror/style/base.css#L70-L97

hxrshxz:
Could you please provide a screenshot for this? It seems like modifying the code doesn't have any effect on my side   ` .cm-searching, .cm-searching span { background-color: var(--jp-search-unselected-match-background-color) !important; color: var(--jp-search-unselected-match-color) !important; } .cm-searching::selection, .cm-searching span::selection { background-color: var(--jp-search-unselected-match-background-color) !important; color: var(--jp-search-unselected-match-color) !important; } .jp-current-match > .cm-searching, .jp-current-match > .cm-searching span, .cm-searching > .jp-current-match, .cm-searching > .jp-current-match span { background-color: var(--jp-search-selected-match-background-color) !important; color: var(--jp-search-selected-match-color) !important; } .jp-current-match > .cm-searching::selection, .jp-current-match > .cm-searching span::selection, .cm-searching > .jp-current-match::selection, .cm-searching > .jp-current-match span::selection { background-color: var(--jp-search-selected-match-background-color) !important; color: var(--jp-search-selected-match-color) !important; }`

hxrshxz:
> 4. See poor contrast due to green not being replaced by search highlight color
Could you please provide a screenshot for the contrast issue? It seems like modifying the code doesn't have any effect on my side .
Edited Code -
![Screenshot 2025-01-10 134149](https://github.com/user-attachments/assets/c93772a0-0e74-40a5-9618-4ea8a7d53ead)

[...]

PR #17107, opened 2024-12-19
changed the selection color theme addressing the issue #17068

haarit19058:
## References

This pull request addresses issue #17068, which involves updating the text selection theme in JupyterLab.

## Code changes

The changes involve modifications to three CSS files that target the text selection theme. These changes improve the visual appearance and consistency of text selection across different JupyterLab components.

/packages/theme-**themetype **/ style / variable.css

## User-facing changes

This is an enchancement  and there are no major functional changes, but the text selection theme has been updated for better visual consistency. Users will notice a more cohesive appearance when selecting text in the editor and other JupyterLab components.

There are no backwards-incompatible changes to JupyterLab's public APIs.

### DIFF
--- a/packages/theme-dark-extension/style/variables.css
+++ b/packages/theme-dark-extension/style/variables.css
@@ -388,12 +388,10 @@ all of MD as it is not optimized for dense, information rich UIs.
   --jp-search-toggle-off-opacity: 0.6;
   --jp-search-toggle-hover-opacity: 0.8;
   --jp-search-toggle-on-opacity: 1;
-  --jp-search-selected-match-background-color: rgb(255, 225, 0);
-  --jp-search-selected-match-color: black;
-  --jp-search-unselected-match-background-color: var(
-    --jp-inverse-layout-color0
-  );
-  --jp-search-unselected-match-color: var(--jp-ui-inverse-font-color0);
+  --jp-search-selected-match-background-color: #0f5ae6;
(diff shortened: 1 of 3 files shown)


haarit19058:
hello i am a beginner and this is going to be my first ever pull request in a large organization so i would be very greateful to receive you invaluable feedback and suggestions. Thank You !!!

krassowski:
This does not solve https://github.com/jupyterlab/jupyterlab/issues/17068. That bug is caused by selectors specificity. What you are doing is changing previously defined variables to have different values. This does not impact the sectors specificity of selectors.

I do not understand _why_ you are trying to change the existing values of the CSS variables.

> The changes involve modifications to three CSS files that target the text selection theme. These changes improve the visual appearance and consistency of text selection across different JupyterLab components.

1. Can you provide before and after screenshots detailing how this improves the appearance?
2. Can you measure the contrast using colour contrast checkers to ensure that the changed values move us closer to WCAG AA contrast requirements?

haarit19058:
Good morning, Sir,

Apologies for the delayed response; I got caught up with my academics. I
checked the contrast ratio, which is 5.44, and it doesn’t seem satisfactory.

I have another question: upon inspecting, I noticed that the selected text
"m " in from is actually black, with a yellow background. This means the
selected text is not green. I’ve attached a screenshot for your reference.

Could you please clarify this?

Issue #16856, opened 2024-10-13
HTML Sanitizer problem/crash


martenrichter:
It may be a problem with Plotly, and I should file an issue there.
However, I am observing a crash in the HTML sanitizer. It says `this.buffer.charCodeAt` is not a function.
The cause is relatively easy to find. Plotly generates HTML lines, which are an array of strings. The HTML sanitizer only allows strings (with the charCodeAt). This issue is probably relatively rare, as it only occurs when I have not installed Plotly in my development environment, and so the HTML code is usually not rendered.
You can reproduce it by creating a notebook with Plotly plots, saving it, and opening it in a clean environment without Plotly.

Nothing really important.

martenrichter:
And the notebook was generated with jupyter lab and the error is seen in jupyter lite (but I did not check if it is also visible in jupyter lab).

JasonWeill:
@martenrichter Thanks for opening this issue. I'm not seeing the code `this.buffer.charCodeAt` in JupyterLab's or Plotly's repos, although it's possible that this code is generated in JavaScript after compiling TypeScript. Are you able to see, using your browser's debugger, where this error originates? If it originates in Plotly, this is probably better to address in the Plotly repo.

martenrichter:
The `this.buffer.charCodeAt` is in the HTML parser package, which is called by Jupyter Lab.
I am currently away from my dev environment, but it can be traced back to where the content with HTML mime types was put into the sanitizer. (I can look this up later and give you the precise line. I think it just must be checked for arrays, and the arrays should be merged into one string.)
The question is if it is legal for the notebooks to put in an array of strings for the HTML mime type. This is what Plotly does. The notebook crashes only if it is opened in an environment without Plotly since, with Plotly, the Plotly mime would be rendered.

martenrichter:
Here is the call stack:
``` 
    at e.parse (Tokenizer.js:745:33)
    at e.write (Tokenizer.js:159:14)
    at e.write (Parser.js:459:28)
    at m (index.js:604:10)
    at oe.sanitize (sanitizer.js:931:16)
    at get headings (widget.js:972:102)
    at be.getHeadings (toc.js:176:83)
    at be.refresh (model.js:130:44)
    at i (factory.js:45:19)
    at d (index.es6.js:555:18)
    ```
I think it is this line
[...]

PR #17114, opened 2024-12-20
Fix sanitizer call in ToC if html data is array of strings

martenrichter:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16856

## Code changes

If the HTML data is an array of strings (e.g. as plotly generates it), the data is joined).

## User-facing changes

No.

No.

### DIFF
--- a/packages/cells/test/widget.spec.ts
+++ b/packages/cells/test/widget.spec.ts
@@ -546,6 +546,27 @@ describe('cells/widget', () => {
       });
     });
 
+    describe('#headings', () => {
+      it('get headings with string[]', () => {
+        const modelheadings = new CodeCellModel();
+        const widget = new CodeCell({
+          contentFactory: NBTestUtils.createCodeCellFactory(),
+          model: modelheadings,
+          rendermime
+        });
+        widget.initializeState();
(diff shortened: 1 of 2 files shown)


krassowski:
Thank you! It looks like this would benefit from being covered by tests, but unfortunately the `headings` getter on `CodeCell` does not have tests yet. I think these could be added in somewhere in:

https://github.com/jupyterlab/jupyterlab/blob/5bc76f03be9fa5341ed117f7bed3dc776ab30b95/packages/cells/test/widget.spec.ts#L530-L559

martenrichter:
Ok, test is added.

packages/cells/test/widget.spec.ts
```diff
@@ -546,6 +546,27 @@ describe('cells/widget', () => {
       });
     });
 
+    describe('#headings', () => {
+      it('get headings with string[]', () => {
```
krassowski, line 550:
```suggestion
      it('should not throw when outputs contain an array of strings', () => {
```

krassowski (approved):
Thank you @martenrichter!

PR #17120, opened 2024-12-23
Add pixi files to various ignore files

jtpio:
## References

Allow contributors to use [pixi](https://pixi.sh/) as an alternative to other package managers, without leaving the repo in a dirty state.

Contributors who want to use `pixi` as an alternative to `conda` / `mamba` for creating an environment can currently add the following to `.git/info/exclude` in their local checkout of the repo:

```
pixi.toml
pixi.lock
```

As a follow-up to this, we could also consider adding `pixi` to the contributing documentation (to be discussed, so it does not make the docs more complex).

## User-facing changes

None

None


.gitattributes
```diff
@@ -1 +1,3 @@
 jupyterlab/staging/yarn.js binary
+# SCM syntax highlighting
+pixi.lock linguist-language=YAML linguist-generated=true
```
krassowski, line 3: This would only matter if we did add a lockfile to repo, right?

krassowski (approved):
Thanks!

.gitattributes
jtpio: Right, this is more for convenience for now, as running `pixi init` would otherwise modify the `.gitattributes` file to add that line automatically.

krassowski: Makes sense then!

PR #17121, opened 2024-12-23
Respect query argument in `settingeditor:open` when settings editor is already open

### LINKED ISSUE #16479: The `query` argument in `settingeditor:open` does not update query when Settings Editor is already open
## Description

Extensions can open the settings editor on a given setting by passing a `query` argument. It works great when the Settings Editor is closed, but if it is already open, the command switches to the Settings Editor tab but ignores the `query` argument.

## Reproduce

This is because of the short-circuit logic in:

https://github.com/jupyterlab/jupyterlab/blob/130089eddb330a44f7f9d496b1e64eddcb497d6b/packages/settingeditor-extension/src/index.ts#L115-L143

which only calls activate but does not update `query` from `args.query`.

## Expected behavior

The query is respected on second and consecutive executions on `settingeditor:open` command.

## Context

- JupyterLab version: 4.2.2

andrewfulton9:
## References

Closes #16479 

## Code changes

Adds the ability for the function editor to update the PluginList Filter and calls on that functionality when openUI is called.

## User-facing changes

Now if the settings editor is already open, when the `settingeditor:open` command is used with a query argument, the query argument will be respected.

This adds an `updateQuery` parameter to `SettingsEditor` for updated the filter.

### DIFF
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -5,7 +5,6 @@ import { IJupyterLabPageFixture, test } from '@jupyterlab/galata';
 import { expect, Locator } from '@playwright/test';
 import fs from 'fs';
 import path from 'path';
-import { LogConsoleHelper } from '../../lib/helpers';
 
 test('Open the settings editor with a specific search query', async ({
   page
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -34,7 +34,7 @@ test('Open the settings editor with a specific search query', async ({
   expect(await settingsPanel.screenshot()).toMatchSnapshot(
     'settings-panel.png'
   );
-  
+
   await page.evaluate(async () => {
     await window.jupyterapp.commands.execute('settingeditor:open', {
       query: 'CodeMirror'
@@ -42,7 +42,6 @@ test('Open the settings editor with a specific search query', async ({
   });
 
   await expect(page.locator('.jp-PluginList-entry')).toHaveCount(1);
-
 });
 
 test.describe('change font-size', () => {
--- a/galata/test/jupyterlab/settings.test.ts
+++ b/galata/test/jupyterlab/settings.test.ts
@@ -31,9 +31,11 @@ test('Open the settings editor with a specific search query', async ({
 
   const settingsPanel = page.locator('.jp-SettingsPanel');
 
-  expect(await settingsPanel.screenshot()).toMatchSnapshot(
+  expect.soft(await settingsPanel.screenshot()).toMatchSnapshot(
     'settings-panel.png'
   );
+  // Test that new query takes effect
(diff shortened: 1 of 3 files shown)


galata/test/jupyterlab/settings.test.ts
```diff
@@ -34,6 +34,14 @@ test('Open the settings editor with a specific search query', async ({
   expect(await settingsPanel.screenshot()).toMatchSnapshot(
```
krassowski, line 34: When adding to existing test, it is a good practice to use `expect.soft` to allow the test to proceed even if there is minor change in visuals:

```suggestion
  expect.soft(await settingsPanel.screenshot()).toMatchSnapshot(
```

krassowski (commented):
Thanks! Small suggestion on code style, should be easy to apply from the GitHub UI :)

galata/test/jupyterlab/settings.test.ts
```diff
@@ -31,9 +31,19 @@ test('Open the settings editor with a specific search query', async ({
...
+  // Test that new query takes effect
+  await expect(page.locator('.jp-PluginList-entry')).toHaveCount(0);
+
+  await page.evaluate(async () => {
+    await window.jupyterapp.commands.execute('settingeditor:open', {
+      query: 'CodeMirror'
+    });
+  });
+
+  await expect(page.locator('.jp-PluginList-entry')).toHaveCount(1);
```
krassowski, line 46: I think we need to change it to actually check that the entry is CodeMirror, not that it has count 1, because even before calling `execute('settingeditor:open', {query: 'CodeMirror'});` there is one entry, so this test would already pass on `main`

andrewfulton9:
@krassowski, thanks for the review and pushing up your suggestions. I'll update the test today and get that pushed up.

galata/test/jupyterlab/settings.test.ts
```diff
@@ -31,9 +31,25 @@ test('Open the settings editor with a specific search query', async ({
...
+    });
+  });
+  // wait for command to be executed
+  await page.waitForTimeout(1000);
+  const listEntry = await page
+    .locator('.jp-PluginList-entry-label-text')
+    .first();
+  const textContent = await listEntry.textContent();
+
+  expect(textContent).toBe('CodeMirror');
```
krassowski, line 52: Let's avoid  explicit timeouts as these are easy to add but accumulate quickly and slow down the test suite. Can you try:

```suggestion
  // wait for command to be executed
  const fistListEntry = page
    .locator('.jp-PluginList-entry-label-text')
    .first();

  await expect(fistListEntry).toHaveText('CodeMirror');
```

`toHaveText` is an auto-waiting assertion and by default waits for up to 5 seconds.

andrewfulton9: Awesome! I tried a couple of things to avoid having the `waitForTimeout` call, but none of them worked. I just tried this out locally and it passes.

krassowski (approved):
Thank you @andrewfulton9!

PR #17122, opened 2024-12-24
Fix exposing `ISignal` instead of `Signal`

jtpio:
## References

Looks like `RestorableSplitPanel` was exposing a `Signal` instead of `ISignal`.

Not a big deal since it's privately defined withing the shell. But we can likely still update it so it exposes an `ISignal` instead, to follow the rest of the code base.

## Code changes

Expose an `ISignal` instead of `Signal` from `RestorableSplitPanel`.

## User-facing changes

None

None, since it's private.


fcollonval (approved):
Thanks @jtpio

PR #17130, opened 2025-01-01
Fix title for overscan count option

krassowski:
## References

It looks like the French preposition "de" found it's way into the codebase in https://github.com/jupyterlab/jupyterlab/pull/12554 and no one noticed for three years ;)

## Code changes

```diff
-      "title": "Number of cells to render outside de the viewport",
+      "title": "Number of cells to render outside the viewport",
```

## User-facing changes

As above

None


jtpio (approved):
Nice catch, thanks!

PR #17132, opened 2025-01-01
Stabilize cell height when scrolling

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/16326
- Should also fix https://github.com/jupyter-book/jupyterlab-myst/issues/243

## Code changes

Adds `cellHeightStabilization` and `assumeHeightStabilityAfter` options:
- `cellHeightStabilization`: Whether to pin height of the cells when scrolling to stabilize viewport (prevent jitter), and for how long since the last scroll event (in milliseconds).
- `assumeHeightStabilityAfter`: If cell height stabilization is enabled, how many milliseconds should pass since the last height measurement for the stabilization algorithm to be allowed to pin the cell height

To be done:
- [ ] minimise public API additions (`widgetSizes` exposure/`_getCellIndex` method)
- [ ] confirm the default choices: 
   - [ ] should `assumeHeightStabilityAfter` be 5 seconds favouring some widgets rendering for up to seconds and less frequent output "unroll" animation, or should it be 1 second favouring more stability when scrolling but less frequent "unroll" animations?
  - [ ] what should `cellHeightStabilization` be? Ideally it is as large as to catch consecutive scroll events (i.e. >= interval between mouse scroll wheel steps, or mouse movements when using scrollbar), and as small as to minimise the time before the outputs are "unrolled" after the scroll stopped.

## User-facing changes

Scrolling is no longer jittery.

| Before | After |
|--|--|
| TBD | TBD |

None

### DIFF
--- a/packages/notebook-extension/schema/tracker.json
+++ b/packages/notebook-extension/schema/tracker.json
@@ -831,6 +831,20 @@
       "enum": ["defer", "full", "none"],
       "default": "full"
     },
+    "cellHeightStabilization": {
+      "title": "Cell height stabilization",
+      "description": "Whether to pin height of the cells when scrolling to stabilize viewport (prevent jitter), and for how long since the last scroll event (in milliseconds).",
+      "type": "number",
+      "default": 750,
+      "minimum": 0
+    },
+    "assumeHeightStabilityAfter": {
+      "title": "Height stabilization sensitivity",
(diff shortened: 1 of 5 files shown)


krassowski:
Closing for now, I will reopen rebased if needed.

PR #17141, opened 2025-01-06
Fix setter for `contentProviderId`

jtpio:
## References

This should help fix the TypeScript build issue noticed in https://github.com/jupyter/notebook/pull/7559#issuecomment-2573550718

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17092

## Code changes

Update setter for the docregistry `contentProviderId`.

## User-facing changes

None

None

### DIFF
--- a/packages/docregistry/src/default.ts
+++ b/packages/docregistry/src/default.ts
@@ -498,7 +498,7 @@ export abstract class ABCWidgetFactory<
   get contentProviderId(): string | undefined {
     return this._contentProviderId;
   }
-  set contentProviderId(value: string) {
+  set contentProviderId(value: string | undefined) {
     if (this._contentProviderId && value !== this._contentProviderId) {
       throw Error(
         `Cannot change content provider on factory with an existing provider: ${this._contentProviderId}`


jtpio:
> Additionally it would be worth investigating why this was not caught by CI. Maybe it's because the `tsconfig.json` used in Notebook is a bit more strict.

Or maybe it's because the `tsconfig` used for test is different, and enables `skipLibCheck`:

https://github.com/jupyterlab/jupyterlab/blob/5243a851c470688893adfbd1b833ea6b2f985c03/tsconfigbase.test.json#L17

krassowski:
I think this is `eslint`'s job to catch this. There is a rule for this now:
- https://github.com/typescript-eslint/typescript-eslint/issues/7236
- https://typescript-eslint.io/rules/related-getter-setter-pairs/

Yet, this is interesting:

> Prior to TypeScript 4.3, the types had to be identical. From TypeScript 4.3 to 5.0, the `get` type had to be a subtype of the `set` type. As of TypeScript 5.1, the types may be completely unrelated as long as there is an explicit type annotation.

krassowski:
Given https://github.com/jupyter/notebook/pull/7562 I suspect that you noticed the TypeScript condition too :)

jtpio:
It seems https://github.com/jupyter/notebook/pull/7559 now builds again after updating to the same TypeScript 5.1.x version in https://github.com/jupyter/notebook/pull/7562. The `yarn.lock` had references to older versions of `typescript` (via `lerna` 6), and updating + deduplicating seems to be helping.

So maybe it's just an issue with older versions of TypeScript. Still it's probably worth having this change for consistency.

krassowski:
I think there is a good argument to merge this just because extensions or third-party downstreams may be using older typescript versions.

jtpio:
Thanks @krassowski, looks like we commented at the same time :)



> I think there is a good argument to merge this just because extensions or third-party downstreams may be using older typescript versions.

Yes sounds good :+1:

krassowski (approved):
Thank you @jtpio!

PR #17143, opened 2025-01-06
Remove `skipLibCheck` from the test `tsconfig`

jtpio:
## References

Checking if we still need this `skipLibCheck` set to `true` for the test `tsconfig`.

## Code changes

Remove `skipLibCheck` from `tsconfigbase.test.json`.

## User-facing changes

None

None

### DIFF
--- a/tsconfigbase.test.json
+++ b/tsconfigbase.test.json
@@ -13,7 +13,6 @@
     "jsx": "react",
     "resolveJsonModule": true,
     "esModuleInterop": true,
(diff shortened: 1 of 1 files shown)


jtpio:
So the tests are still passing after removing that option. But it doesn't seem to have any effect on the `docregistry` tests noticed in https://github.com/jupyterlab/jupyterlab/pull/17141.

But we can probably still get this in and investigate separately is this may be an issue with `jest` instead.

PR #17156, opened 2025-01-10
Fix emission of `lastCell` from notebook run actions

pawel99k:
## References

None

## Code changes

Minor bugfix resolving issue in actions with accessing last notebook's cell. This PR e.g. enables NotebookActions.selectionExecuted to properly return the information about last cell.

## User-facing changes

None

### DIFF
--- a/packages/notebook/test/actions.spec.ts
+++ b/packages/notebook/test/actions.spec.ts
@@ -1118,7 +1118,8 @@ describe('@jupyterlab/notebook', () => {
         let notebook: Notebook | null = null;
         let lastCell: Cell | null = null;
         NotebookActions.selectionExecuted.connect(async (_, args) => {
-          notebook, (lastCell = args);
+          notebook = args[0];
+          lastCell = args[1];
           emitted += 1;
         });
         const result = await NotebookActions.runCells(
--- a/packages/notebook/test/actions.spec.ts
+++ b/packages/notebook/test/actions.spec.ts
@@ -1111,6 +1111,24 @@ describe('@jupyterlab/notebook', () => {
         expect(result).toBe(true);
         expect(widget.isSelected(widget.widgets[2])).toBe(true);
       });
+
+      it('should run the selected cells', async () => {
+        let emitted = 0;
+        const next = widget.widgets[2];
+        NotebookActions.selectionExecuted.connect(async (_, args) => {
+          const { notebook, lastCell } = args;
+          expect(notebook).not.toBe(null);
+          expect(lastCell).not.toBe(null);
+          emitted += 1;
+        });
+        const result = await NotebookActions.runCells(
+          widget,
+          [widget.widgets[1], widget.widgets[2]],
+          sessionContext
+        );
+        expect(result).toBe(true);
+        expect(emitted).toBe(2);
+      });
     });
 
     describe('#runAll()', () => {
--- a/packages/notebook/test/actions.spec.ts
+++ b/packages/notebook/test/actions.spec.ts
@@ -1129,8 +1129,8 @@ describe('@jupyterlab/notebook', () => {
         );
         expect(result).toBe(true);
         expect(emitted).toBe(1);
-        expect(notebook).not.toBe(null);
-        expect(lastCell).not.toBe(null);
+        expect(notebook).toBeInstanceOf(Notebook);
+        expect(lastCell).toBeInstanceOf(Cell);
       });
     });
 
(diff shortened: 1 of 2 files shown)


krassowski:
Good catch, thank you! Would you be able to add a test for this in:

https://github.com/jupyterlab/jupyterlab/blob/1f6f64bbd56bc9363315114a4fe0f642252ef464/packages/notebook/test/actions.spec.ts#L1086-L1114

To check what gets emitted in the `NotebookActions.selectionExecuted` you could follow a pattern similar to:

https://github.com/jupyterlab/jupyterlab/blob/1f6f64bbd56bc9363315114a4fe0f642252ef464/packages/notebook/test/actions.spec.ts#L721-L734

pawel99k:
Hi @krassowski, is it something you had in mind? I don't really know TS actually

packages/notebook/test/actions.spec.ts
```diff
@@ -1111,6 +1111,24 @@
         expect(result).toBe(true);
         expect(widget.isSelected(widget.widgets[2])).toBe(true);
       });
+
+      it('should run the selected cells', async () => {
+        let emitted = 0;
+        const next = widget.widgets[2];
```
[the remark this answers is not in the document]
krassowski: It looks like this is unused?

```suggestion
```

packages/notebook/test/actions.spec.ts
```diff
@@ -1111,6 +1111,24 @@ describe('@jupyterlab/notebook', () => {
...
+        NotebookActions.selectionExecuted.connect(async (_, args) => {
+          const { notebook, lastCell } = args;
+          expect(notebook).not.toBe(null);
+          expect(lastCell).not.toBe(null);
```
krassowski, line 1121: This works, but it would be a better practice to define `notebook` and `lastCell` one level higher and just assign it here from `args`, and the move the `expect` to the main body of the test.

packages/notebook/test/actions.spec.ts
```diff
@@ -1111,6 +1112,26 @@ describe('@jupyterlab/notebook', () => {
...
         expect(widget.isSelected(widget.widgets[2])).toBe(true);
       });
+
+      it('should run the selected cells', async () => {
+        let emitted = 0;
+        let notebook: Notebook | null = null;
+        let lastCell: Cell | null = null;
+        NotebookActions.selectionExecuted.connect(async (_, args) => {
+          notebook = args[0];
+          lastCell = args[1];
```
krassowski, line 1122: This currently fails on CI:

```
Error: test/actions.spec.ts(1121,22): error TS7053: Element implicitly has an 'any' type because expression of type '0' can't be used to index type '{ notebook: Notebook; lastCell: Cell<ICellModel>; }'.
  Property '0' does not exist on type '{ notebook: Notebook; lastCell: Cell<ICellModel>; }'.
```

```suggestion
          notebook = args.notebook;
          lastCell = args.lastCell;
```

packages/notebook/test/actions.spec.ts
```diff
@@ -1111,6 +1112,26 @@ describe('@jupyterlab/notebook', () => {
...
+          lastCell = args.lastCell;
+          emitted += 1;
+        });
+        const result = await NotebookActions.runCells(
+          widget,
+          [widget.widgets[1], widget.widgets[2]],
+          sessionContext
+        );
+        expect(result).toBe(true);
+        expect(emitted).toBe(2);
```
krassowski, line 1131: I think this indeed should be one, because only one selection gets executed:

```suggestion
        expect(emitted).toBe(1);
```

packages/notebook/test/actions.spec.ts
```diff
@@ -1111,6 +1112,26 @@ describe('@jupyterlab/notebook', () => {
...
+        });
+        const result = await NotebookActions.runCells(
+          widget,
+          [widget.widgets[1], widget.widgets[2]],
+          sessionContext
+        );
+        expect(result).toBe(true);
+        expect(emitted).toBe(1);
+        expect(notebook).not.toBe(null);
+        expect(lastCell).not.toBe(null);
```
krassowski, line 1133: I am always uneasy with `not.toBe(null)` because `undefined` technically is not a `null`. See https://github.com/adobe/brackets/issues/5479

```suggestion
        expect(notebook).toBeInstanceOf(Notebook);
        expect(lastCell).toBeInstanceOf(Cell);
```

krassowski (approved):
Thank you @pawel99k!

krassowski:

Issue #15102, opened 2023-09-10
File Editor does not scroll on undo/redo


krassowski:
## Description

In a longer text files editor no longer scrolls on undo/redo. This can lead to unintentional code/prose loss because user may mistakenly undo too much of text typed without being made aware that a change happened in given editor. As previously discussed for a similar issue in notebook (now fixed) this is very frustrating for developers.

## Reproduce

1. Open Text Editor
2. Insert text twice longer than screen
3. Edit at beginning
4. Scroll to end
5. Undo
6. See it does not scroll

## Expected behavior

It scroll so that the edited section is in the middle

## Context

- JupyterLab version: 4.0.5

krassowski:
This is an upstream issue:
- https://github.com/yjs/y-codemirror.next/issues/31

On our side the undo manager comes from yjs and the binding are defined in (and proceeding code):

https://github.com/jupyterlab/jupyterlab/blob/d9d578ccb0e9385fb332dc9b23357a00648e66fe/packages/codemirror/src/extensions/ybinding.ts#L261-L295

Note that we are vendoring the upstream code (with licence and attribution) since at least https://github.com/jupyterlab/jupyterlab/pull/13639. I wonder if we should instead depend on upstream as much as possible to pickup any fixes and to make it easier to contribute back.

Note: simple solution of "scroll on text change" would not work in RTC setting. We need a more subtle solution that distinguishes undo/redo from other edits.

YqunYang:
This issue significantly affects the experience of using Jupyter Lab. Because of this, I find it quite challenging to write code in Jupyter Lab, as it often leads to various problems. I sincerely hope this can catch your attention, and I truly appreciate your effort in addressing it. Thank you so much!

krassowski:
Thanks for the comment @YqunYang this motivated me to spend an evening trying to solve it - I opened https://github.com/jupyterlab/jupyterlab/pull/17158.

YqunYang:
Thank you so much for your effort! I’m truly grateful that my feedback helped inspire this fix. Your contribution means a lot—thank you! @krassowski

zhangfeiran:
Thanks for your effort! Now the behavior of cursor moving on undo/redo is mostly as expcted, but there are still two aspects that may cause inconvenience:
- when redo after undo, most editors will not jump cursor back to before undo, e.g. 
[...]

PR #17158, opened 2025-01-11
Fix scrolling and selection restoration on undo/redo

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/15102

## Code changes

- vendors `yUndoManager` view plugin from `y-codemirror.next` which was previously wrongly deemed as not necessary
   - the argument in the comment original mentioned that JupyterLab handles keybindings itself, but that plugin does not in fact add keybindings, although it enables kyebindings undo/redo commands upstream
- while I would prefer to just depend on `y-codemirror.next`, this is not currently easy because:
  - the `ybinding` consists of vendored code which has inter-dependencies with `yUndoManager`
  - upstream does not export types correctly which prevents compilation with TypeScript, see https://github.com/yjs/y-codemirror.next/issues/27

## User-facing changes

- Undo/redo now correctly scrolls to the position where change was made
- Selection prior to undo/redo is now correctly restored

None

### DIFF
--- a/packages/codemirror/src/extensions/ybinding.ts
+++ b/packages/codemirror/src/extensions/ybinding.ts
@@ -15,6 +15,11 @@ import {
   SelectionRange
 } from '@codemirror/state';
 import { EditorView, ViewPlugin, ViewUpdate } from '@codemirror/view';
+import {
+  yUndoManager,
+  YUndoManagerConfig,
+  yUndoManagerFacet
+} from './yundomanager';
 import type {
   RelativePosition,
   Text,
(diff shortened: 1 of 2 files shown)


dlqqq:
> upstream does not export types correctly which prevents compilation with TypeScript, see https://github.com/yjs/y-codemirror.next/issues/27

The linked issue is now closed. One of the maintainers commented that the package already provides the `types` field.

krassowski:
Yes, but we are still bound by:

> the `ybinding` consists of vendored code which has inter-dependencies with `yUndoManager`

In other words, we need to go all in for vendored or bundled. I think we can do that in the minor version bump, but for this PR (which I hope we can backport) I would like to keep the changes minimal.

krassowski:
@fcollonval do you think we could merge and backport this one (as the committer of `ybinding.ts` file)?

fcollonval (approved):
Thanks Mike

LGTM

krassowski:

PR #17162, opened 2025-01-12
Heuristic for cell height now considers preferred text output types

hxrshxz:
## References
#16636 
## Code changes

The function estimateWidgetSize was updated to improve how cell height is calculated. earlier it only considered text/plain outputs, but now it supports multiple output types that are -
application/vnd.jupyter.stderr (for errors/tracebacks)
application/vnd.jupyter.stdout (for printed outputs)
Other types like text/html, text/markdown, and image/svg+xml.

None


hxrshxz:
@meeseekdev tag bug

hxrshxz:
@meeseeksdev tag bug

krassowski:
Can you link existing PRs that target this issue and compare your PR with those?

PR #17167, opened 2025-01-14
Fix read-only cells becoming editable on settings change

### LINKED ISSUE #15514: Read-only cells become editable when changing notebook settings programatically
## Description

Read-only cells become editable when changing notebook settings programatically. This approach is used to change `autoStartDefaultKernel` in core JupyterLab by clicking "Always start the preferred kernel" checkbox. Changing settings manually does not lead to this problem.

This was originally reported for [jupyterlab-execute-time](https://github.com/deshaw/jupyterlab-execute-time/issues/107) which changes `recordTiming` programatically.

## Reproduce

1. Open a notebook, select a cell
2. Open property inspector, switch cell to "Read-Only"
3. Click on kernel selector
4. Check "Always start the preferred kernel" in the dialog
5. Accept
6. See that the cell can be edited despite the metadata still forbidding it

breaks-read-only

## Expected behavior

Cells stay read-only when notebook settings are changed programatically.

## Context

- JupyterLab version: 4.0.9 and 4.1.0a4

<details>

<summary>Stack trace</summary>

image

</details>

Darshan808:
## References

Fixes #15514 

### Problem
Read-only cells became editable when modifying notebook settings. This occurred because keys like `readOnly` are only stored in the `config` but not in the cell's `editorConfig`. During config updates, these keys were returned by `getChanged` (as it returns modified keys and keys not present in new config but present in old config) and subsequently removed from the config (when new config is assigned to config), causing the `readOnly` property to be lost.

### Solution
Modified the implementation of `setOptions` to ensure that customized keys present in the config but missing from the cell's `editorConfig` are preserved in the config, preventing the loss of the `readOnly` property.

### DIFF
--- a/packages/codemirror/src/extension.ts
+++ b/packages/codemirror/src/extension.ts
@@ -185,6 +185,7 @@ export class ExtensionsHandler
    * to `IExtensionsHandler.configChanged`.
    */
   setBaseOptions(options: Record<string, any>): void {
+    //Change values of baseConfig
     const changed = this._getChangedOptions(options, this._baseConfig);
     if (changed.length > 0) {
       this._baseConfig = options;
@@ -197,6 +198,12 @@ export class ExtensionsHandler
             return agg;
           }, {})
         );
+      }
+    }
+    //Chang values of config keys if present in options
+    for (const key of Object.keys(options)) {
+      if (key in this._config && this._config[key] != options[key]) {
+        this.setOption(key, options[key]);
       }
     }
   }
--- a/packages/codemirror/src/extension.ts
+++ b/packages/codemirror/src/extension.ts
@@ -185,7 +185,6 @@ export class ExtensionsHandler
    * to `IExtensionsHandler.configChanged`.
    */
   setBaseOptions(options: Record<string, any>): void {
-    console.log("In setBase Options!");
     const changed = this._getChangedOptions(options, this._baseConfig);
     if (changed.length > 0) {
       this._baseConfig = options;
@@ -810,7 +809,7 @@ export namespace EditorExtensionRegistry {
         default: [],
         factory: () =>
           createConfigurableExtension((value: number[]) =>
-            value.length > 0 ? rulers(value) : []
+            value?.length > 0 ? rulers(value) : []
           ),
         schema: {
           type: 'array',
--- a/packages/codemirror/src/extension.ts
+++ b/packages/codemirror/src/extension.ts
@@ -185,6 +185,7 @@ export class ExtensionsHandler
    * to `IExtensionsHandler.configChanged`.
    */
   setBaseOptions(options: Record<string, any>): void {
+    console.log("In setBase Options!");
     const changed = this._getChangedOptions(options, this._baseConfig);
     if (changed.length > 0) {
       this._baseConfig = options;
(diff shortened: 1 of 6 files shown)


Darshan808:
But `setOptions` is implemented so that the whole config can be changed at once. In this implementation, if we have `readOnly` set to `true` and we wish to remove it by not providing it in `setOptions`, it will still consider it as a customized key and preserve it.

I think it would be better if we listen to `changedConfig` and update the cell's `editorConfig` to include the changed settings like `readOnly`.

krassowski:
On the issue I suggested replacing `setOptions` with `setBaseOptions` in `updateEditorConfig`. Could differentiating behaviour at the level of caller make it easier?

krassowski:
> or we could just not emit the keys which are undefined and let them retain their existing values

This sounds reasonable to me, I would try it and see if tests pass (or how many break). We can make wider adjustments across the codebase if needed.

Darshan808:
The errors caused by undefined values being emitted during the use of setBaseOptions occurred because the editor attempted to update its configuration using all changed values, including those that were undefined. To resolve this, I added an additional filter that ensures the editor only updates its configuration with defined values. This effectively resolves the issue without any errors appearing in the console.

krassowski:
It looks like two tests in `js-notebook` group are failing:

```
  ● @jupyterlab/notebook › NotebookActions › #toggleAllLineNumbers() › should toggle line numbers on all cells

    Expected: true
    Received: false

      1663 |           const lineNumbers =
      1664 |             widget.widgets[i].editor!.getOption('lineNumbers');
    > 1665 |           expect(lineNumbers).toBe(!state);
           |                               ^
      1666 |         }
      1667 |       });
      1668 |

      at Object.<anonymous> (test/actions.spec.ts:1665:31)

  ● @jupyterlab/notebook › NotebookActions › #toggleAllLineNumbers() › should be based on the state of the active cell

    Expected: true
    Received: false

      1676 |           const lineNumbers =
      1677 |             widget.widgets[i].editor!.getOption('lineNumbers');
    > 1678 |           expect(lineNumbers).toBe(!state);
           |                               ^
      1679 |         }
      1680 |       });
      1681 |

      at Object.<anonymous> (test/actions.spec.ts:1678:31)
```

krassowski (commented):
I tested locally and can confirm that this fixes the issue, thank you :rocket:

Can you add a test for this issue, ideally using `jest`  but if not possible then with `galata`?

If `galata` test is required I would suggest adding one without a snapshot, just checking if cell has `jp-mod-readOnly` class after changing a setting. This could go to: https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/jupyterlab/settings.test.ts

packages/codemirror/src/extension.ts
```diff
@@ -199,6 +200,12 @@ export class ExtensionsHandler
         );
       }
     }
+    //Chang values of config keys if present in options
```
krassowski, line 203:
```suggestion
    // Change values of config keys if present in options
```

packages/codemirror/src/extension.ts
```diff
@@ -810,7 +816,7 @@ export namespace EditorExtensionRegistry {
         default: [],
         factory: () =>
           createConfigurableExtension((value: number[]) =>
-            value.length > 0 ? rulers(value) : []
+            value?.length > 0 ? rulers(value) : []
```
krassowski, line 819: Is the addition of `?` necessary here?

packages/codemirror/src/extension.ts
```diff
@@ -185,6 +185,7 @@ export class ExtensionsHandler
    * to `IExtensionsHandler.configChanged`.
    */
   setBaseOptions(options: Record<string, any>): void {
+    //Change values of baseConfig
```
krassowski, line 188:
```suggestion
    // Change values of baseConfig
```

krassowski (approved):
Thank you @Darshan808!

Because it adds a new method to the public API I am tagging it for 4.4.0.

Issue #16491, opened 2024-06-16
"Click to add cell" is not expanded to the full cell width on small screens


krassowski:
## Description

![image](https://github.com/jupyterlab/jupyterlab/assets/5832902/c901ab0f-34d7-4ad9-b90b-316214b6de0f)

## Reproduce

1. Open notebook
2. Resize the browser window to small enough width
3. Hover over notebook footer
4. See add cell button not cover the full width available

## Expected behavior

Add cell button expands to cover the width available

## Context

- Browser and version: Chrome
- JupyterLab version: 4.2.2

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16383

krassowski:
This issue should not be closed until visual regression test for it exists as this and edge case that is unlikely to be caught in the usual testing workflow.

HaudinFlorence:
@krassowski Thanks for opening this issue and putting this corner case into evidence.

krassowski:
>  Is this related to a media screen query ?

This would be my guess too.

MUFFANUJ:
Just had a look and seems like, it involves a simpler calculation for margin and width just need to exclude the ```--jp-cell-prompt-width``` as it does not contribute when the screen size is smaller.
i.e, 

<img width="668" height="95" alt="Image" src="https://github.com/user-attachments/assets/df4fb757-cafb-461b-b682-863f3440574c" />

there are on different levels ( aligned column wise ). Opening a PR soon :)

PR #17170, opened 2025-01-15
Click to Add cell expands to full width now on smaller screens

hxrshxz:
## References
#16491 
## Code changes
added media query which will align the.jp-Notebook-footer to the correct place for screens with a size smaller than 760px. Margin-left is explicitly set to 28px to override the 87px Margin-left from the border-box styling of the.lm-widget. For maintaining the same width alignment with the cells above, the calculated width is subtracted by 57.38px. At the smaller screen sizes, this ensures that the footer perfectly matches the cell width.

## User-facing changes
**Before :**
image

**After :** 
image

NONE

### DIFF
--- a/packages/notebook/style/notebookfooter.css
+++ b/packages/notebook/style/notebookfooter.css
@@ -37,3 +37,17 @@
     opacity: 1;
   }
 }
+
+@media (max-width: 760px) {
+  .jp-Notebook-footer {
+    margin-left: 28px;
+    width: calc(
+      100% -
+        (
+          var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
+            var(--jp-cell-padding) + 2 * var(--jp-notebook-padding) - 57.38px
+        )
+    );
+  }
+}
+
--- a/packages/notebook/style/notebookfooter.css
+++ b/packages/notebook/style/notebookfooter.css
@@ -45,9 +45,8 @@
       100% -
         (
           var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
-            var(--jp-cell-padding) + 2 * var(--jp-notebook-padding) - 57.38px
+            var(--jp-cell-padding) + 2 * var(--jp-notebook-padding) - 57.9px
         )
     );
   }
 }
-


hxrshxz:
meeseekdev tag bug

hxrshxz:
can u please update snapshots

packages/notebook/style/notebookfooter.css
```diff
@@ -37,3 +37,16 @@
...
 }
+
+@media (width <= 760px) {
+  .jp-Notebook-footer {
+    margin-left: 28px;
+    width: calc(
+      100% -
+        (
+          var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
+            var(--jp-cell-padding) + 2 * var(--jp-notebook-padding) - 57.9px
```
krassowski, line 48: It looks like this calculation is derived from lines 12-18 - which is fine! Can you add a new private variable (say `--jp-private-notebook-footer-width-spacer`) and use it in both places to avoid duplication which may lead to future errors?

Do you know where the value of `57.9px` comes from?

krassowski (commented):
I think there are no screenshots to update here, but we should add a screenshot test.

hxrshxz:
> I think there are no screenshots to update here, but we should add a screenshot test.

sure that would be interestnig Let me review the Playwright documentation and resources to learn how to implement it, and then we can work on setting it up together.

packages/notebook/style/notebookfooter.css
hxrshxz: After some trial and error, I arrived at 57.9px, but in the latest commit after some more adjustments i added dynamic width that now perfectly aligns with the cell width.

hxrshxz:
@krassowski Should we add more breakpoints or are the current ones enough?

galata/test/jupyterlab/notebook-footer.test.ts
```diff
@@ -0,0 +1,39 @@
...
+            // Dynamically set the viewport for the test
+            await page.setViewportSize({ width, height: 800 });
+
+            // Go to the JupyterLab login page
+            await page.goto('http://127.0.0.1:8888/login?next=%2Flab%2Fworkspaces%2Fauto-L%2Ftree%2FUntitled3.ipynb');
+
+            // Login with the token
+            await page.getByLabel('Password or token:').click();
+            await page.getByLabel('Password or token:').fill('df3ae5e770971bd20578df895ae0d727cdaa765c79591730');
+            await page.getByRole('button', { name: 'Log in', exact: true }).click();
```
krassowski, line 22: This should not be needed when running the test via [galata](https://github.com/jupyterlab/jupyterlab/tree/main/galata) (a thin wrapper around playwright) :)

galata/test/jupyterlab/notebook-footer.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+            // Navigate to the notebook page
+            await page.goto('http://127.0.0.1:8888/lab/workspaces/auto-L/tree/Untitled3.ipynb');
```
krassowski, line 25: In galata you would instead create a new notebook for testing using helper:

```js
 await page.notebook.createNew(fileName);
```

galata/test/jupyterlab/notebook-footer.test.ts
```diff
@@ -0,0 +1,39 @@
...
+            // Wait for the footer element to be available
+            const footer = page.locator('.jp-Notebook-footer');
+            await footer.waitFor();
+
+            // Take a screenshot of the footer for this width
+            const screenshotPath = `footer-responsive-${width}px.png`;
+            await footer.screenshot({ path: screenshotPath });
+
+            // Optionally, verify the screenshot against a baseline
+            await expect(footer).toHaveScreenshot(screenshotPath);
```
krassowski, line 36: We use `toMatchSnapshot` pattern, see example in 

https://github.com/jupyterlab/jupyterlab/blob/87cb79fe29c623f1e2f3d00e8d0d51c656ef12d4/galata/test/jupyterlab/notebook-create.test.ts#L66-L69

galata/test/jupyterlab/notebook-footer.test.ts
```diff
@@ -0,0 +1,39 @@
...
+            // Login with the token
+            await page.getByLabel('Password or token:').click();
+            await page.getByLabel('Password or token:').fill('df3ae5e770971bd20578df895ae0d727cdaa765c79591730');
+            await page.getByRole('button', { name: 'Log in', exact: true }).click();
+
+            // Navigate to the notebook page
+            await page.goto('http://127.0.0.1:8888/lab/workspaces/auto-L/tree/Untitled3.ipynb');
+
+            // Wait for the footer element to be available
+            const footer = page.locator('.jp-Notebook-footer');
```
krassowski, line 28: We will need to force a hover state for the screenshot to be useful, right?

krassowski (commented):
Good start on the test, I left some suggestions

galata/test/jupyterlab/notebook-footer.test.ts
hxrshxz: yes, definitely ! I was actually using a bg color instead and removed it when pushing the changes

hxrshxz:
suggestions noted ! thank you , i will upate you with the changes soon

hxrshxz:
I was able to set up a public server of JL but Whenenver if i try to use the Galata wrapper this happens->

Uploading Screen Recording 2025-01-19 173607.mp4…

hxrshxz:
I actually figured out what was wrong but now i am facing another error i will update you soon with the changes

hxrshxz:
@krassowski could you please review this when you get time

hxrshxz:
@krassowski  coudl you please have a look and let me know if any other changes need to be made?

galata/test/jupyterlab/notebook-footer.test.ts
```diff
@@ -0,0 +1,28 @@
...
+      await page.notebook.createNew(fileName)
+      // Wait for the footer element to be available
+      const footer = page.locator('.jp-Notebook-footer');
+      await footer.waitFor();
+
+      await footer.hover();
+
+      // Capture a screenshot of the open menu and verify responsiveness
+      const imageName = `footer-viewport-${width}px.png`;
+      const footerScreenshot = await footer.screenshot();
```
krassowski, line 24: I think this screenshot should cover more than just the footer so that we can see if it aligns with the cell bounarieboundaries.

galata/test/jupyterlab/notebook-footer.test.ts-snapshots/footer-viewport-400px-jupyterlab-win32.png
krassowski, line 1: The windows snapshots need to be removed. I can ask bot to generate snapshots but I think it makes sense to rework them a bit, see the other comments

packages/notebook/style/notebookfooter.css
```diff
@@ -2,20 +2,35 @@
...
+:root {
+  --jp-private-notebook-footer-width-spacer: calc(
+    100% -
+      (
+        var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
+          var(--jp-cell-padding) + 2 * var(--jp-notebook-padding)
+      )
   );
-  width: calc(
+  --dynamic-extra-width: clamp(58px, 7.87vw, 59.9px);
```
krassowski, line 13: This would need `--jp-private` prefix too. What do these values mean? I think we would need to find a definition which is derived from other variables, not based on just guessing the values.

packages/notebook/style/notebookfooter.css
```diff
@@ -2,20 +2,35 @@
...
+ :root {
+  --jp-private-notebook-footer-width-spacer: calc(
     100% -
       (
         var(--jp-cell-prompt-width) + var(--jp-cell-collapser-width) + 2 *
           var(--jp-cell-padding) + 2 * var(--jp-notebook-padding)
       )
   );
+  --dynamic-extra-width: clamp(58px, 7.87vw, 59.9px);
+}
```
krassowski, line 24: This looks to be an exact duplicate of the code above.

krassowski (changes requested):

PR #17180, opened 2025-01-17
Remove `--subshell-console` flag

### LINKED ISSUE #17146: Maybe remove `--subshell-console` boolean config flag
We now have support for kernel subshells in JupyterLab prerelease 4.4.0a2 (issue #16882, PR #16963). This can be tried out if you also have `ipykernel` prerelease 7.0.0a0 installed using `jupyter lab --subshell-console`. This issue is to discuss whether we should remove the config flag and have it always available, and ideally we make a decision before the full 4.4.0 release.

Core support for subshells is always available, regardless of the config flag, in `KernelConnection` for any kernel supporting [JEP91](https://github.com/jupyter/enhancement-proposals/pull/91). This will allow us to use subshells in `ipywidgets`, the debugger, etc, to give a better experience without the user having to know that subshells are being used.  The `--subshell-console` config flag is used to expose in the UI the option to open a subshell console via the context menu of a kernel notebook, and this will be of interest to a relatively small number of users who want to explicitly experiment with subshells.

My original motivations for the flag forcing users to opt-in were twofold:
1. I was being conservative, not knowing if UI changes in Lab require and long and detailed acceptance process.
2. The subshells implemented in `ipykernel` are a minimum viable implementation, missing some features as outlined in the Future Work section of ipython/ipykernel#1249 such as correct handling of execution count and history. If the subshell console is exposed to all users by default then some will try it without knowing what it is and this may cause confusion for those users and extra issues opened.

Personally I do not have a strong opinion on keeping or removing the config flag. I evidently did not need to be so conservative so item 1 is no longer relevant, and item 2 may not be significant enough to justify the addition of the flag and its future maintenance burden.

ianthomas23:
This PR removes the `--subshell-console` config flag added in #16963.

Closes #17146, which is still open to debate.

### DIFF
--- a/galata/jupyter_server_test_config.py
+++ b/galata/jupyter_server_test_config.py
@@ -5,7 +5,6 @@ from jupyterlab.galata import configure_jupyter_server
 
 configure_jupyter_server(c)
 c.LabApp.dev_mode = True
-c.LabApp.subshell_console = True
 
 # Uncomment to set server log level to debug level
 # c.ServerApp.log_level = "DEBUG"
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -497,10 +497,6 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
         {"LabApp": {"custom_css": True}},
         "Load custom CSS in template html files. Default is False",
     )
-    flags["subshell-console"] = (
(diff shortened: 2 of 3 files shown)


jtpio (approved):
Thanks!

krassowski (approved):
:+1:

PR #17184, opened 2025-01-17
Fix: Add scrollbar to File menu section in JupyterLab 4.3.x

### LINKED ISSUE #17083: File menu drop down truncated with no scrollbar in 4.3.x
With JupyterLab 4.3, if the File menu item drop-down is longer than the height of the browser, the drop-down appears truncated just before the lower edge of the browser with no visible scrollbar or way to scroll within the drop-down:

File menu drop-down with Lab 4.3

Expected behavior, seen with JupyterLab 4.2.6:

File menu drop-down with Lab 4.2.6

Tested in Chrome with Jupyter 4.2.6 and 4.3.

JigyasuRajput:
## References

- Fixes #17083 

## Code changes

- Adjusted the CSS to ensure the scrollbar is visible and functional, providing a smoother user experience in JupyterLab 4.3.x

## User-facing changes

- The File menu now includes a scrollbar when its content exceeds the visible area, allowing users to scroll through all available options.
- This resolves the issue where the menu was truncated without a scrollbar, forcing users to rely on touchpad gestures or mouse wheel scroll.

Before - 
before

After - 
after

None

### DIFF
--- a/packages/application/style/menus.css
+++ b/packages/application/style/menus.css
@@ -29,8 +29,6 @@
 
 .lm-MenuBar-menu.jp-ThemedContainer {
   top: calc(-2 * var(--jp-border-width));
-  scrollbar-width: none;
-  -ms-overflow-style: none;
   overflow: auto;
   background:
     linear-gradient(var(--jp-layout-color0) 30%, rgba(0, 0, 0, 0)) center top,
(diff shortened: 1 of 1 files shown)


brichet:
Removing the scrollbar visibility was intentional in https://github.com/jupyterlab/jupyterlab/pull/16945.

Not sure that we want it back, or maybe with a settings if this is expected by the user.

Issue #15088, opened 2023-09-07
Run UI tests on Firefox


jtpio:
### Problem

Currently UI tests only test on Chromium.

However some issues sometimes happen only on Firefox. So testing on Firefox would help avoid potential regressions.

### Proposed Solution

UI tests are currently only tested on `chromium`: 

https://github.com/jupyterlab/jupyterlab/blob/f43ceaccb308155106be48942736302db41d431c/.github/workflows/galata.yml#L48-L49

We could set up the test matrix to also test on `firefox`.

### Additional context

This would unfortunately increase the CI load.

jtpio:
Re-opening as https://github.com/jupyterlab/jupyterlab/pull/17965 added UI testing on Firefox, but many tests are still skipped.

For example: 

https://github.com/jupyterlab/jupyterlab/blob/f3f75f777147955ea2bee59d848e466f301c6bfe/galata/test/jupyterlab/completer.test.ts#L59

We should look into enabling some of them back, if possible.

PR #17190, opened 2025-01-18
firefox browser added for UI tests

hxrshxz:
## References
#15088

## User-facing changes
Added firefox browser to Visual Regression Tests

None


hxrshxz:
@meeseeksdev tag bug

hxrshxz:
@krassowski please tell if any other changes need to be made

krassowski (changes requested):
@hxrshxz thank you for your interest in improving the testing breadth :pray:

I am concerned about it having a potential to add burden to already stretched review crew if we were to include a second set of snapshots for Firefox. If we were to add testing on Firefox I suggest we should disable snapshot comparison for Firefox.

As of now, this does not work:

image

image

Issue #16917, opened 2024-11-05
Remove "Do not ask me again" checkbox from rename on first save dialog


JasonWeill:
### Problem

The "rename untitled on first save" dialog, which I'll call the "first save dialog" for short, contains four controls: a text box, a "Do not ask me again." check box, a Cancel button, and a Rename button:

![image](https://github.com/user-attachments/assets/8d0c111a-2852-471e-b27e-a3413c3017f7)

There are two problems with this dialog:

1. If the user clicks "Do not ask me again" and then cancels the dialog, no renaming takes place, and the user will be asked to rename files again.
2. It is not clear from the checkbox's label alone what "Do not ask me again" means. The label, on hover, has a caption: "If checked, you will not be asked to rename future untitled files when saving them."

### Proposed Solution

Have three buttons in this dialog, and no checkbox:

1. Cancel — Do nothing, then ask again when the user next tries to save
2. Rename — Rename this file, then ask about the next untitled file we save (the default)
3. Rename and Don't Ask Again — Rename this file, but do not ask about renaming other files in the future.

If the user chooses the third option above, the name in the text box is used, but the first save dialog will never be shown again. If a user chooses to save an untitled file, the default name will be retained.

### Additional context

"Don't Ask Again" would be a shorter option for the new button, although it does not convey an action to be taken. "Rename and Don't Ask Again" is unusually long for a button in English, and may be even longer in other languages.

krassowski:
> .If the user clicks "Do not ask me again" and then cancels the dialog, no renaming takes place, and the user will be asked to rename files again.

I think the problem is broader wider. When I press Ctrl + S for the first time I am presented with this dialog and it might appear as if "Cancel" could also cancel saving so users (e.g. me) developed an un-inutitive habit of pressing Ctrl + S again after closing the dialog.

krassowski:
I feel like three buttons might look overwhelming a bit in this small dialog. I think we should not use Cancel here.

What about:

```
[ ] Do not ask for rename on first save [ Save as Untitled21.ipynb ] [ Rename ]
```

krassowski:
Or maybe:

```
[ ] Do not ask for rename on first save [ Save as Untitled21.ipynb ] [ Rename and Save ]
```

PR #17193, opened 2025-01-19
Dialog box removed

hxrshxz:
## References
#16917 

## Code changes

Removed the "Rename Untitled on First Save" dialog entirely.

## User-facing changes
Before 
image

After 
image

Many popular IDEs and modern applications take a straightforward approach to saving, making it easier for users to focus on their work without getting interrupted by unnecessary dialog boxes which may be distracting and may affect UI
 If users don’t have to interpret unclear dialog
 options or make extra decisions, it simplifies their experience.

None

### DIFF
--- a/packages/apputils/style/dialog.css
+++ b/packages/apputils/style/dialog.css
@@ -114,16 +114,17 @@ button.jp-Dialog-close-button {
 .jp-Dialog-footer {
   display: flex;
   flex-direction: row;
-  justify-content: flex-end;
+  justify-content: flex-start;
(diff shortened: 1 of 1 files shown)


krassowski (changes requested):
This PR does not implements the proposals discussed in #16917 and instead hides all checkboxes in all dialogs. It cannot be merged in this state.

krassowski:
Should we close this in favour of https://github.com/jupyterlab/jupyterlab/pull/17217?

Issue #17164, opened 2025-01-13
Cannot remove cell metadata


plstonge:
## Description

Once the Cell metadata is modified with new attributes, it becomes impossible to remove any added attribute.

## Reproduce

1. In a notebook, create a Markdown cell
2. While this cell is selected, open the Property Inspector
3. Expand the Advanced Tools pane
4. Under Cell metadata, between `{}`, enter something like `"lang": "en"`
5. Click the check mark button.
6. Delete everything between `{}`. Click again the check mark button
7. Collapse the Advanced Tools pane
8. Expand the Advanced Tools pane

The attribute `"lang": "en"` is back.

9. Expand the Common Tools pane

The Cell metadata is now filled with the defaults of the Common Tools.

10. Delete everything again. Click again the check mark button
11. Collapse and Expand the Advanced Tools pane

All deleted attributes are back.

## Expected behavior

When valid metadata are saved through the Advanced Tools, nothing should overwrite what is saved by the check mark button. In the above examples, when all metadata are deleted, it should remain deleted.

## Context

- Operating System and version: Windows 11 Pro, version 24H2
- Browser and version: Firefox 134.0
- JupyterLab version: 4.3.4

andrii-i:
I can reliably reproduce this in JupyterLab 4.3.4.

PR #17194, opened 2025-01-19
Fix for issue preventing cell metadata removal

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/17164

The root cause was that `setMetadata` method of the `ISharedBaseCell` only overrides the partial metadata (existing keys), rather than replacing the entire metadata.

I have just provided a workaround to ensure the complete removal and replacement of all metadata. However, a more prominent solution would be overriding the `setMetadata` method to ensure it completely replaces the metadata.

### DIFF
--- a/packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
+++ b/packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
@@ -61,6 +61,12 @@ export class CellMetadataField extends NotebookTools.MetadataEditorTool {
 
   private _onSourceChanged() {
     if (this.editor.source) {
+      const metadataKeys = Object.keys(
+        this._tracker.activeCell?.model.sharedModel.metadata ?? {}
+      );
+      for (const key of metadataKeys) {
+        this._tracker.activeCell?.model.sharedModel.deleteMetadata(key);
+      }
       this._tracker.activeCell?.model.sharedModel.setMetadata(
         this.editor.source.toJSON()
       );
--- a/packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
+++ b/packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
@@ -60,16 +60,17 @@ export class CellMetadataField extends NotebookTools.MetadataEditorTool {
   }
 
   private _onSourceChanged() {
-    if (this.editor.source) {
-      const metadataKeys = Object.keys(
-        this._tracker.activeCell?.model.sharedModel.metadata ?? {}
-      );
-      for (const key of metadataKeys) {
-        this._tracker.activeCell?.model.sharedModel.deleteMetadata(key);
-      }
-      this._tracker.activeCell?.model.sharedModel.setMetadata(
-        this.editor.source.toJSON()
-      );
+    const activeCell = this._tracker.activeCell?.model.sharedModel;
+    if (activeCell && this.editor.source) {
+      const metadataKeys = Object.keys(activeCell.metadata ?? {});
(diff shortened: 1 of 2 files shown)


packages/notebook-extension/src/tool-widgets/metadataEditorFields.tsx
```diff
@@ -61,6 +61,12 @@ export class CellMetadataField extends NotebookTools.MetadataEditorTool {
 
   private _onSourceChanged() {
     if (this.editor.source) {
+      const metadataKeys = Object.keys(
+        this._tracker.activeCell?.model.sharedModel.metadata ?? {}
+      );
+      for (const key of metadataKeys) {
+        this._tracker.activeCell?.model.sharedModel.deleteMetadata(key);
+      }
```
krassowski, line 69: Just by looking at the code, I have a feeling that lines 67-72 may need to be wrapped in a single `transaction` as for the user this should be undoable as a single change.

For reference see:

https://github.com/jupyterlab/jupyterlab/blob/07bef086048a3ca995db347537bf25726ddc2168/packages/notebook/src/actions.tsx#L223-L226

krassowski (commented):
Thank you! This makes sense as a quick fix. I think it would be worth to add a comment in code explaining why `deleteMetadata` is required.

How difficult would be adding a minimal galata test for this (possibly without a screenshot, just trying to remove a tag and making sure it was removed based on selector counts)? Maybe something could be added in or around:

https://github.com/jupyterlab/jupyterlab/blob/07bef086048a3ca995db347537bf25726ddc2168/galata/test/jupyterlab/metadataform.test.ts#L466

krassowski:
Maybe a jest test testing `CellMetadataField` would be better. We just need to initialize a cell (which we already do in unit tests), render it with `react-test-renderer` as in:

https://github.com/jupyterlab/jupyterlab/blob/19a5a06b0f8138d3ba4c4902b23df0bc5352f370/packages/settingeditor/test/settingsformeditor.spec.tsx#L6

edit the cell contents and see if it the changes are as expected.

itsmevichu:
Yeah sure! I will add the testcases..

itsmevichu:
I have added the galata testcase for the changes mentioned above. However, I’m uncertain about adding unit tests, as the code changes are part of the extension package.

krassowski:
> I have added the Galata testcase for the above changes.

Thank you!

> I’m uncertain about adding unit tests, as the code changes are part of the extension package.

We can have unit tests in the extension packages, including for React code, as we do for `shortcuts-extension`:

https://github.com/jupyterlab/jupyterlab/blob/ad4e2d979d37e90ef91530a8fe4e147078668ccf/packages/shortcuts-extension/test/components/ShortcutUI.spec.ts#L75-L96

Unit tests are significantly faster to run and often cheaper to maintain, but since you already written an integration test, that is absolutely fine! If you need help adding a test suite to an extension package in the future just let me know.

I see some failures on CI, not sure if these are related - but I do not see them on the `main` branch:

```
Notice:   3 failed
    [jupyterlab] › test/jupyterlab/outputarea-stdin.test.ts:110:7 › Stdin for ipdb › Subsequent execution in short succession 
    [jupyterlab] › test/jupyterlab/workspace.test.ts:159:7 › Workspace › should clone the default workspace 
    [jupyterlab] › test/jupyterlab/workspace.test.ts:195:7 › Workspace › should clone the current workspace 
```

The first one is a known flaky test, the later two - first time I see them.

I restarted the job, let's see if it passes now.

krassowski (approved):
Thank you @itsmevichu!

krassowski:

PR #17195, opened 2025-01-20
Fix "running" prompt state with server-side execution

krassowski:
## References

- Follow up to https://github.com/jupyterlab/jupyterlab/pull/16651

## Code changes

- [x] Adds a test demonstrating an issue where the `execution_count` being reset incorrectly overrode the execution state, resulting in no `[*]` prompt in the interface but instead only `[ ]`
- [ ] Fixes the issue by not changing `execution_count` when provided with the value of `null`

## User-facing changes

`[*]` shows up correctly when using server-side exeuction

None


jtpio (approved):
Thanks!

jtpio:

PR #17197, opened 2025-01-20
Speed up large and streamed text outputs with incremental autolinking, preserve selection

### LINKED ISSUE #16957: Logging to stderr is significantly slower in Jupyterlab vs IPython

krassowski:
## References

Fixes #16957

## Code changes

Makes the auto-linking logic execute asynchronously, at the speed matching the approximate refresh rate of standard monitors. Appends new nodes in the textual streams without removing the existing ones minimising layout computation cost.

1. **Spreads the work across frames**
   - Rendering happens in small steps inside a `requestAnimationFrame` loop (a browser callback that runs about once per screen refresh). Each frame does at most ~10 ms of auto-linking and then continues on the next frame, so one output cannot freeze the page.
   - Each step processes a block of text of about 8 KB (extended to the next space), rather than one word or line at a time, which lowers the per-step overhead.
   - When several large outputs render at once, they take turns: only one output is worked on per frame, and the loop cycles through them so each keeps making visible progress and none stays blank.
   - A safety timer (named a "watchdog" in the code) still shows the plain text if the browser stops sending animation frames, for example in a background tab or under heavy load. It only acts when frames have genuinely stopped, not when an output is simply waiting for its turn.
2. **Prioritises what the user can see**
   - Outputs scrolled off screen are found with an `IntersectionObserver`. We use it instead of measuring visibility ourselves, because measuring directly would force the browser to recompute layout at a bad moment.
   - Off-screen outputs keep rendering, but quietly, during the browser's idle time (`requestIdleCallback`, with a timer-based fallback for Safari). They are then ready when scrolled to, without slowing down the outputs on screen.
   - As soon as an off-screen output scrolls back into view, it returns to normal per-frame rendering. The first paint of any output is never delayed by visibility.
3. **Updates the DOM without rebuilding it** (append-only, no trashing)
   - Each output keeps a live `<pre>` element in two parts:
     - a "committed" part at the start: text that has already been linked and is never touched again;
     - a "tail" at the end: the not-yet-linked text, shown as plain text immediately (with terminal colours kept), split into small pieces.
   - Every frame moves a little of the tail into the committed part after linking it, so the amount of page update per frame depends only on how much was processed that frame, not on the total output size.
  For streamed output, the next chunk reuses the page content from the previous render instead of rebuilding it. Only the small region at the end that could still change is redone, for example a link split across two chunks (chunk one ends with `www.`, chunk two starts with `example.com`). This replaces the earlier approach that compared and swapped nodes.
4. **Adds a logic to retain selection when updating text**. This is required because when we are replacing `www.example.com` in partially rendered text output `aaa www.example.com ccc` we are also implicitly destroying the selection. This also leads to an improvement to status quo where selection was being destroyed on each stream chunk being added.

#### Notable edge cases

- Search highlighting keeps working. Because the committed part is never rewritten, highlights added to it stay in place across frames. If another feature changes the tail (for example search splitting or wrapping the text there), only the tail is rebuilt so committed highlights are kept, and a full rebuild is used only as a last resort.
- Text selection is preserved. When text is replaced during linking, the selection is remembered as character positions and restored afterwards. This also fixes an existing problem where the selection was lost every time a new chunk of streamed text arrived.
- File-path links in error outputs (such as File "example.py", line 12) keep their line number. Error outputs link both URLs and file paths, and a file path can contain spaces, so steps there split on line breaks instead of on spaces.
- For very large outputs the slow part is the browser re-laying-out the shifted text on each frame, so the total time depends mostly on how many frames are used. In that case the per-frame budget is allowed to grow (up to 200 ms) so the work finishes in fewer frames, while small outputs keep the responsive 10 ms budget. The budget only grows from this output's own previous frame, so several outputs rendering together do not push each other's budgets up.

#### Code organisation

- The live-page logic, node merging, and selection helpers are now contained in the `LiveText` class in `rendermime/src/livetext.ts`. `renderers.ts` keeps the frame scheduling and the auto-linking.
- A benchmark and analysis scripts were added under `galata/test/benchmark` to measure both stdout and stderr, streaming versus re-opening a notebook, and scaling with output size and number of outputs.

### It's faster

The output content is displayed when it gets returned from the kernel as soon as possible - no more choking the rendering pipeline due to too much output streamed at once.

https://github.com/user-attachments/assets/a8122a4e-4ea2-4ca4-ba35-53d5990b0611

As an example, when profiling the first paint of full (non-linkified) text occurred in under 4 seconds (allowing me to read the last line), whereas on 4.4.0a2 I had to wait 29 seconds to get it (but it was already with all links displayed).

### The links show up with a delay

The links are displayed as soon as these are processed. This means that users may now see non-clickable links, but as seen on the recording above, even though initially the text is not "linkified" the last links shows up faster than before! This is not guaranteed - it will depend on the size of the output.

> Note: the recording is from an earlier iteration; now links only show up with delay on much larger outputs; keeping that older recording as it demonstrates the idea well; see benchmarks below for when it kicks in nowadays.

### Selection is fun again

Selection of outputs that is being streamed is not destroyed when new output arrives

https://github.com/user-attachments/assets/97741eb8-c6f0-4ba7-b11d-61253a617311

### There is an opt-out

I can't think of a way this could introduce a regression, but still just in case if someone was depending on some quirk of the previous implementation, this PR includes a setting to disable incremental rendering.

image

Question: should we use "auto-link" or "autolink"?

### Table 1 Streaming on cell execution

- incremental = this PR
- synchronous = before/with incremental auto linking turned off

During interactive execution both: full text, and all links show up much faster than before. Note that in synchronous case all-text = all-links.

#### Streaming stdout (web links)

<table>
  <caption>1a - Streaming · stdout (web links)</caption>
  <thead>
    <tr><th>pipeline</th><th>time to</th><th>IQM</th><th>Q1-Q3</th><th>longest frame</th><th>median frame</th></tr>
  </thead>
  <tbody>
    <tr><th rowspan="3" scope="rowgroup">incremental</th><td>first text</td><td>85</td><td>79-91</td><td rowspan="3">50</td><td rowspan="3">17</td></tr>
    <tr><td>all text</td><td>448</td><td>444-454</td></tr>
    <tr><td>all links</td><td>448</td><td>444-454</td></tr>
    <tr><th rowspan="3" scope="rowgroup">synchronous</th><td>first text</td><td>91</td><td>89-93</td><td rowspan="3">117</td><td rowspan="3">100</td></tr>
    <tr><td>all text</td><td>1,819</td><td>1,814-1,830</td></tr>
    <tr><td>all links</td><td>1,819</td><td>1,814-1,830</td></tr>
  </tbody>
</table>

#### Streaming stderr (web + path links)

<table>
  <caption>1b - Streaming · stderr (web + path links)</caption>
  <thead>
    <tr><th>pipeline</th><th>time to</th><th>IQM</th><th>Q1-Q3</th><th>longest frame</th><th>median frame</th></tr>
  </thead>
  <tbody>
    <tr><th rowspan="3" scope="rowgroup">incremental</th><td>first text</td><td>134</td><td>108-169</td><td rowspan="3">55</td><td rowspan="3">22</td></tr>
    <tr><td>all text</td><td>452</td><td>414-476</td></tr>
    <tr><td>all links</td><td>452</td><td>414-476</td></tr>
    <tr><th rowspan="3" scope="rowgroup">synchronous</th><td>first text</td><td>284</td><td>246-318</td><td rowspan="3">183</td><td rowspan="3">117</td></tr>
    <tr><td>all text</td><td>12,283</td><td>12,132-12,409</td></tr>
    <tr><td>all links</td><td>12,283</td><td>12,132-12,409</td></tr>
  </tbody>
</table>

#### Bulk rendering stdout (web links)

<table>
  <caption>2a - Re-open · stdout (web links)</caption>
  <thead>
    <tr><th>pipeline</th><th>time to</th><th>IQM</th><th>Q1-Q3</th><th>longest frame</th><th>median frame</th></tr>
  </thead>
  <tbody>
    <tr><th rowspan="3" scope="rowgroup">incremental</th><td>first text</td><td>310</td><td>299-323</td><td rowspan="3">72</td><td rowspan="3">17</td></tr>
    <tr><td>all text</td><td>310</td><td>299-323</td></tr>
    <tr><td>all links</td><td>310</td><td>299-323</td></tr>
    <tr><th rowspan="3" scope="rowgroup">synchronous</th><td>first text</td><td>298</td><td>296-303</td><td rowspan="3">90</td><td rowspan="3">28</td></tr>
    <tr><td>all text</td><td>298</td><td>296-303</td></tr>
    <tr><td>all links</td><td>298</td><td>296-303</td></tr>
  </tbody>
</table>

#### Bulk rendering on re-open · stderr (web + path links)

<table>
  <caption>2b - Re-open · stderr (web + path links)</caption>
  <thead>
    <tr><th>pipeline</th><th>time to</th><th>IQM</th><th>Q1-Q3</th><th>longest frame</th><th>median frame</th></tr>
  </thead>
  <tbody>
    <tr><th rowspan="3" scope="rowgroup">incremental</th><td>first text</td><td>348</td><td>343-360</td><td rowspan="3">95</td><td rowspan="3">22</td></tr>
    <tr><td>all text</td><td>348</td><td>343-360</td></tr>
    <tr><td>all links</td><td>348</td><td>343-360</td></tr>
    <tr><th rowspan="3" scope="rowgroup">synchronous</th><td>first text</td><td>339</td><td>334-344</td><td rowspan="3">95</td><td rowspan="3">33</td></tr>
    <tr><td>all text</td><td>339</td><td>334-344</td></tr>
    <tr><td>all links</td><td>339</td><td>334-344</td></tr>
  </tbody>
</table>

Note that for synchronous scenario all values are identical on reopening here as expected.

#### Scaling with output size - streaming stderr

image

(1600 blocks synchronous pipeline did not run until completion as it exceeded 5 minutes threshold)

#### Scaling with output size - re-opening a notebook with stdout

image

#### Scaling with number of outputs - re-opening a notebook with stderr

image

#### Linkification only starts to lag behind text when re-opening a notebook with very large output

image

## Future work

Potential improvements:
- We could try to infer the refresh rate of the monitor using intervals between `requestAnimationFrame` and use it instead of a hard-coded time limit; this might improve the experience for users with high-refresh rate monitors (bot only if they are also using high-end CPU/GPU combination that can process enough `linkify` passes in such shorter duration of the frame to avoid frame dropping)
- We could avoid the layout cost by imposing size restriction on the output if no new text was streamed (but it is only being auto-linked). This is because the links should be rendered using the same font (but only different colour).
   - I attempted it but in initial attempts I did not see significant gains
- Instead of constantly calling `requestAnimationFrame` (and cancelling it) we could call it once and update arguments if new chunk is streamed; this is because there is an overhead associated (approx 0.6s for each 10s so not terrible, but noticeable); instead we could pass state via a `WeakMap` or similar.

### Already handled in this PR (were previously listed here)

- Re-layout cost: with the append-only update, committed text is never laid out again. Together with `contain: style layout` this gives a very good performance; I previously explored also fixing the size but it is not necessary nor desirable
- Frame timing on capable machines: handled by letting the per-frame budget grow, without having to detect the exact monitor refresh rate.

### To consider in future PRs

- **Parsing repeats over the whole output on every chunk**: Each streamed chunk re-cleans and re-parses all the text received so far, not only the new part. Now that updating the page is cheap per chunk, this parsing is the main remaining cost for long streams. Parsing only the new text (while carrying terminal-colour state from the previous chunk) would fix it.
- **High-refresh-rate monitors**: We could measure the time between frames and stop assuming 60 frames per second, but this only helps where the CPU and GPU can also fit enough linking work into the shorter frame. I did attempt that but it did not seem worth it at the time.
- **Animation-frame scheduling overhead**: The loop requests and cancels a frame on every step (a small but measurable cost). Requesting a frame once and passing state through a `WeakMap` would remove it (at a cost of making it harder to reason about).
- **Progress bars that rewrite the same line** (using `\r`, like tqdm): These still redo work on each update. This is acceptable for now because we do not expect outputs which are 10s of MBs in size to be constrained to a single line (which would be somewhat pathological, even for random bytes to not include a single `\n`), but the impact on when it does happen could be limited with more work.
- **One `IntersectionObserver` is created per output**: A single shared observer would be lighter for notebooks with thousands of outputs (but potentially make the code more complex to reason about).
- **Using document fragments for one-pass insertion**: This could reduce the number of DOM insertions (but make it larger); in interim benchmarks the gains were negligible, and I don't want to complicate this PR even further. It is a simple change which can be considered separately, see below

<details>
<summary>document fragment diff for future</summary>

```diff
       const merged = mergeNodes(preNodes, linkedSlice);
-      for (const child of Array.from(merged.childNodes)) {
-        state.pre.insertBefore(child, insertBeforeNode);
+      // Move the merged nodes in one batch: a single insertion queues one
+      // mutation record - instead of one per node - for any observer watching
+      // the output (e.g. search highlighting).
+      const fragment = document.createDocumentFragment();
+      while (merged.firstChild) {
+        fragment.appendChild(merged.firstChild);
       }
+      state.pre.insertBefore(fragment, insertBeforeNode);
     });
```

</details>

None.

## AI Usage

- The initial code was hand-written in early 2025
- In 2026 revival I:
  - used Opus 4.8 to fix tests and a few minor regressions
  - iterated with Fable 5 (Max) to create benchmarks, further improve performance and refactor the code
  - used Opus 4.8 to assist with final checks
  - used Copilot for self-review

## Backports

Some of the improvements are self-contained and could be backported:
- [x] using `.includes()` over constructing a regex for fast-path sanitization heuristic https://github.com/jupyterlab/jupyterlab/pull/17262
- [x] maybe the entire `replaceChangedNodes` logic? - no not doing so - attempted in https://github.com/jupyterlab/jupyterlab/pull/17261 and it did not seem worth it


krassowski:
There is some limitations of the `requestAnimationFrame` approach:
- we need to be very careful not to read any DOM properties (which we still do when restoring the selection, but this is only the case if user has a selection), because we are writing to the DOM at the end to avoid layout trashing and in consequence frame dropping.
- it does not run when JupyterLab is in a background tabs; this might be a good thing (preserving battery for laptops?)
- I observed that there may be a number of frames for which `rAF` is not invoked when running in heavily resource-constrained conditions (e.g. when I ran out of memory and OS started swapping). This appears to be a choice of the browser (Chrome). While it is good for delayed `linkify` calls, it is bad if the first paint (the plain text) is also delayed because of it.

packages/rendermime/src/widgets.ts
```diff
@@ -100,7 +100,7 @@ export abstract class RenderedCommon
     // TODO compare model against old model for early bail?
 
     // Empty any existing content in the node from previous renders
-    if (!keepExisting) {
+    if (!keepExisting && !this.keepExisting) {
       while (this.node.firstChild) {
```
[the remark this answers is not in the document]
krassowski: Addressed in f606d6c25cee140ff69a4c81cdf38424ad075a04

MUFFANUJ:
This seems very helpful, while testing the same things without this PR was a lot problematic, though by simple combinations i was able to achieve the same results even when the setting is enabled ( the vanishing of the output and the code completely from the editor ) - without this there were several problem with selection, running cells interaction etc. with this PR this seem to fix many below are the visuals of testing this on Safari, i havent spend time on looking at code yet, but interactions wise this still might needs some updates.

With incrementalAutolink Disabled:

https://github.com/user-attachments/assets/b57a71b8-b86e-4134-8eef-955fa29bcce6

With incrementalAutolink Enabled:

https://github.com/user-attachments/assets/aac4afa1-52d5-4e74-9c36-b3999d0ee890

Also, the loading and response time in safari seemed very off i dont think its because of my setup and its not the goal of this PR but just wanted to mention it maybe it we needs to work on that later in future.

krassowski:
I am not sure how to reproduce the issues you are referencing from the screencasts, can you write down reproduction steps?

> vanishing of the output and the code completely from the editor

Do you mean ~1:36 on the second recording? It is quite curious, are you saying that this is there both before and after (thus likely unrelated to this PR)?

It looks like there is flashing of content when you open menu? Can you record a performance profile when this happens to see where the browser is spending time?

> the loading and response time in safari seemed very off i dont think its because of my setup and its not the goal of this PR but just wanted to mention it maybe it we needs to work on that later in future

Yes, I am aware that the performance on Safari is worse in general.

MUFFANUJ:
> I am not sure how to reproduce the issues you are referencing from the screencasts, can you write down reproduction steps?

Run the cell -> keeping shifting between the tabs and open context menu inbetween keeping changing between tabs as in the video -> see the vanishing/flickering of content.



> Do you mean ~1:36 on the second recording? It is quite curious, are you saying that this is there both before and after (thus likely unrelated to this PR)?

Yes 



> Can you record a performance profile when this happens to see where the browser is spending time?

Sure!

MUFFANUJ:
So interesting after [ecf9562](https://github.com/jupyterlab/jupyterlab/pull/17197/commits/ecf9562c386fa7426d482ca00fff7c6afc2dd5ef), I'm not able to reproduce the issue I listed above. I went back to [f6a2557](https://github.com/jupyterlab/jupyterlab/pull/17197/commits/f6a255734984d79f262f4396a55f9baf6e0bd074), and I'm able to reproduce it quite easily.

packages/rendermime/src/renderers.ts
```diff
@@ -821,7 +709,9 @@ function* alignedNodes<T extends Node, U extends Node>(
 export async function renderText(
   options: renderText.IRenderOptions
 ): Promise<void> {
-  renderTextual(options, {
+  // Text has no post-render step, so we do not await completion here (rendering
+  // continues incrementally across animation frames).
+  void renderTextual(options, {
     checkWeb: true,
```
MUFFANUJ, line 725: I think this changes the public completion semantics of `renderText()`/`RenderedText.renderModel()`.

`renderTextual()` now returns a promise that resolves when the incremental animation-frame render is actually finished, but `renderText()` discards that promise with `void`. Because `renderText()` is `async`, callers awaiting it get an already-resolved promise even though the `<pre>` may not have been inserted into the DOM yet.

That seems risky because `RenderedCommon.renderModel()` documents that its promise resolves when rendering is complete, and callers may legitimately do:

```ts
await renderer.renderModel(model);
const text = renderer.node.textContent;
```
With this PR, that can observe empty or partial DOM for text outputs.
Could we return the `renderTextual()` promise here instead?
```ts
return renderTextual(options, {
  checkWeb: true,
  checkPaths: false
});
```
Callers that intentionally do not need completion can still fire-and-forget with void renderText(...), but the exported renderer API would keep its existing “await means rendered” contract.

packages/rendermime/src/livetext.ts
```diff
@@ -0,0 +1,1024 @@
...
+  root: Node,
+  textOffset: number | null,
+  offset: number = 0
+) {
+  if (textOffset !== null) {
+    for (const node of [...root.childNodes]) {
+      // As much as possible avoid calling `textContent` here as it is slower
+      const nodeEnd =
+        node instanceof Text
+          ? node.nodeValue!.length
```
MUFFANUJ: `findTextSelectionNode` should not measure element positions using only an element’s first child. Some rendered nodes can contain multiple text-bearing descendants, so offsets that fall inside later descendants may be treated as if they are outside the element and restored to the wrong place after a DOM update. Could this use a consistent text-node traversal, or at least the element’s full `textContent` length, and also could add a regression test for restoring a selection inside a multi-child element across a streamed update.

on binder:

https://github.com/user-attachments/assets/a1bae117-9311-43c8-9be6-2373f5ea8db7

packages/rendermime/src/renderers.ts
krassowski: Addressed in adbf03c

packages/rendermime/src/livetext.ts
krassowski: Added tests in d21c902 with naive solution and added optimized solution in 029d6ab. Thanks!

MUFFANUJ (approved):
Thanks, LGTM!

PR #17199, opened 2025-01-20
No of items in directories in file size column

### LINKED ISSUE #16743: Display size of directory/folder in the "Size" column of file browser
### Problem

Currently file browser "File size" column (to be renamed to "Size" as per #16734) does not show anything for directories. This makes it a bit useless especially when trying find a directory with many files quickly

### Proposed Solution

Show number of items in the directory in that column

### Additional context

Some applications show "X items", some show nothing, some compute the size of all files and show  "X MB" (but that one is hard)

Rishab87:
## References
Fixes #16743 

## Code changes
Added no of items in size column for directories

## User-facing changes
User can now see how many items are their one level down in a directory in file browser

Screenshot From 2025-01-20 22-57-35

None

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -979,7 +979,8 @@ export class DirListing extends Widget {
         this._hiddenColumns,
         this.selection[item.path],
         this._modifiedStyle,
-        this._columnSizes
+        this._columnSizes,
+        this._model.manager.services.contents
       );
       if (
         this.selection[item.path] &&
@@ -2710,7 +2711,8 @@ export namespace DirListing {
       hiddenColumns?: Set<DirListing.ToggleableColumn>,
       selected?: boolean,
       modifiedStyle?: Time.HumanStyle,
-      columnsSizes?: Record<IColumn['id'], number | null>
+      columnsSizes?: Record<IColumn['id'], number | null>,
+      contentsManager?: Contents.IManager
     ): void;
 
     /**
@@ -3148,7 +3150,8 @@ export namespace DirListing {
       hiddenColumns?: Set<DirListing.ToggleableColumn>,
       selected?: boolean,
       modifiedStyle?: Time.HumanStyle,
-      columnsSizes?: Record<DirListing.IColumn['id'], number | null>
+      columnsSizes?: Record<DirListing.IColumn['id'], number | null>,
+      contentsManager?: Contents.IManager
     ): void {
       if (selected) {
         node.classList.add(SELECTED_CLASS);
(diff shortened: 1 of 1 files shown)


packages/filebrowser/src/listing.ts
```diff
@@ -3220,7 +3223,26 @@
...
+                if (contents.content && fileSize) {
+                  const count = contents.content.length;
+                  fileSize.textContent = `${count} items`;
+                  const currentHover = node.title;
+                  node.title = currentHover + trans.__('\nItems: %1', count);
+                }
+              })
+              .catch(error => {
+                console.error('Error getting directory contents:', error);
+                fileSize ? (fileSize.textContent = trans.__('Unknown')) : '';
```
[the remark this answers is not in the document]
jtpio: Looks like this CodeQL warning is relevant.

jtpio:
Thanks @Rishab87 for working on this 👍 

Displaying the number of items is definitely an improvement compared to displaying nothing.

Out of curiosity, what are some popular applications that also display the number of items in directories? (just for reference)

Rishab87:
@jtpio , nautilus (files app of gnome) does this

image

packages/filebrowser/src/listing.ts
```diff
@@ -3220,7 +3223,28 @@ export namespace DirListing {
...
+                }
+              })
+              .catch(error => {
+                if (fileSize) {
+                  fileSize.textContent = '-';
+                }
+                console.error('Error getting directory contents:', error);
+              });
+          }
+        }
```
krassowski, line 3247: This is a nice proof of concept!

I think this would better be taken from the current listing information the same way as `last_modified` and `size`. I think jupyter-server does not include that information when getting folder contents, so we would need to first make a change in `jupyter-server` to optionally include the number of items for directories.

I think so because:
- if user enters a directory with thousands of folders, this many requests will be sent, blocking jupyter-server
- changing lots of elements in large file browser one at a time (as requests are resolved) will cause many layouts potentially freezing JupyterLab when many large notebooks are open
- if user changes directories while the requests are being resolve the node content might be overwritten (this could be fixed by tracking request count)

Rishab87: I completely agree with you, no of items is not being returned by server that's why I went ahead with this condition but this can cause performance issues and race conditions, so should I try to modify jupyter server first? and then make changes here accordingly?

krassowski: Yes, I think opening an issue on jupyter-server repository (if there is not one already) and a draft PR over there would be good next steps :)

Rishab87: I can't see any issue related to this their, creating a new one, thank you for your suggestion

Issue #17140, opened 2025-01-06
Clicking close All on open tabs will always leave one, not all of them


Summer-Yun:
- JupyterLab version: 4.3.1
Steady emergence

step 1
Open multiple notebook files
![image](https://github.com/user-attachments/assets/79ff1538-38a6-48c5-939b-49e9822b3e4e)

step 2
Click Close All
![image](https://github.com/user-attachments/assets/4565f97f-e5c7-4d91-8c3a-dc0340044f77)

step 3
Not completely closed
![image](https://github.com/user-attachments/assets/9d4877bd-f045-43a6-8ead-fe91da3457d9)

Summer-Yun:
> I will also mention that I see the expected behavior (all notebook file tabs close, and a launcher tab remains/appears) when selecting `File`->`Close All Tabs`.
> 
> Added note from the Triage meeting: It would be best to try to reproduce this issue on the latest 4.3.x version of JupyterLab, which is currently 4.3.4.


Well, thank you very much. I know that the function of close All is to keep a launcher tab, which is normal, but it is not normal if other files are left. I hope to fix this experience problem as soon as possible. Moreover, I have reproduced this problem on the latest version 4.3.4.

krassowski:
The relevant code is:

https://github.com/jupyterlab/jupyterlab/blob/c7426e494067765bb314c5ff3bd7c052913c5674/packages/running-extension/src/opentabs.ts#L78-L82

I do not see anything wrong in the code.

I was not able to reproduce this on 4.4.0a2 but I did not try with older versions.

Lewic1201:
> The relevant code is:
> 
> [jupyterlab/packages/running-extension/src/opentabs.ts](https://github.com/jupyterlab/jupyterlab/blob/c7426e494067765bb314c5ff3bd7c052913c5674/packages/running-extension/src/opentabs.ts#L78-L82)
> 
> Lines 78 to 82 in [c7426e4](/jupyterlab/jupyterlab/commit/c7426e494067765bb314c5ff3bd7c052913c5674)
> 
>  shutdownAll: () => { 
>    for (const widget of labShell.widgets('main')) { 
>      widget.close(); 
>    } 
>  }, 
> I do not see anything wrong in the code.
> 
> I was not able to reproduce this on 4.4.0a2 but I did not try with older versions.
我也发现了这个问题，得点两次才能完全关闭。

Lewic1201:
> The relevant code is:
> 
[...]

PR #17203, opened 2025-01-21
Fix for inconsistent tab closure in "Close All Tabs" operation

itsmevichu:
Fixes https://github.com/jupyterlab/jupyterlab/issues/17140

The issue occurred because the `for..of` loop directly consumed the iterator returned by `labShell.widgets('main')`. As widgets were closed during iteration, the iterator's state was modified, causing it to skip widgets.

More specifically, when the `launcher tab` is closed, it is immediately removed from `labShell.widgets('main')`. This modification to the iterator’s state causes the next widget (the first opened notebook) to be skipped, leaving it unclosed.

The solution is to convert the iterator into an array before iterating over it to avoid this problem.


krassowski (approved):
Thank you @itsmevichu!

krassowski:

PR #17204, opened 2025-01-21
Fixed the search functionality for Cell outputs

hxrshxz:
## References
#15445 

## Code changes
Updated the search functionality to include outputs in JupyterLab cells.
## User-facing changes

Users can now search for text within cell outputs directly (This is now the default behavior)

None


krassowski:
It looks like there is an existing PR https://github.com/jupyterlab/jupyterlab/pull/16260 which tackles https://github.com/jupyterlab/jupyterlab/issues/15966? I understand that just toggling the flag is not desirable as this disables "replace" option by default.

hxrshxz:
I actually missed the existing PR my apologies before opening this one

Issue #16174, opened 2024-04-17
Update to TypeScript 5.4


jtpio:
The last update was for TypeScript 5.1, done in https://github.com/jupyterlab/jupyterlab/pull/14638.

It would be great to look into updating to the latest version to not lag behind for too long.

In https://github.com/jupyterlab/jupyterlab/pull/14638#issuecomment-1803584472 @fcollonval said:

> but bumping to 5.2 or later will have some effects as we will need to change the ts configuration.

Looks like this refers to the "module and moduleResolution Must Match Under Recent Node.js settings" section mentioned in https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#breaking-changes-and-correctness-fixes ?

jtpio:
> Looks like this refers to the "module and moduleResolution Must Match Under Recent Node.js settings" section mentioned in https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#breaking-changes-and-correctness-fixes ?

Looks like we'll want to keep `esnext` here: 

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/tsconfigbase.json#L10

But that will conflict with the `node` module resolution specified here: 

https://github.com/jupyterlab/jupyterlab/blob/2ceabd8c7b98e7b2c305be406ade799f387a90eb/tsconfigbase.json#L11

From https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#breaking-changes-and-correctness-fixes :

> So for example --module esnext --moduleResolution node16 will be rejected — but you may be better off just using --module nodenext alone, or --module esnext --moduleResolution bundler.

jtpio:
Linking to https://github.com/jupyterlab/jupyterlab/issues/12518 and https://github.com/jupyterlab/jupyterlab/pull/12519 as potentially relevant too.

jtpio:
TypeScript 5.5 is now available: https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/

jtpio:
TypeScript 5.6 is now available: https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/

Adding this issue to the `4.4.0` milestone, as it would be nice to not fall behind too much.

jtpio:
TypeScript 5.7 was also released in the meantime: https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/

[...]

PR #17207, opened 2025-01-22
Update to TypeScript 5.2

jtpio:
## References

Small step towards #16174, trying to catch up on the latest TypeScript releases.

## Code changes

- [x] Update to `typescript@~5.2.2`

## User-facing changes

None

None

### DIFF
--- a/buildutils/tsconfig.json
+++ b/buildutils/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "lib",
     "rootDir": "src",
-    "module": "commonjs",
+    "module": "Node16",
     "moduleResolution": "node16"
   },
   "include": ["src/*"],
--- a/builder/package.json
+++ b/builder/package.json
@@ -71,7 +71,7 @@
     "@types/node": "^20.11.27",
     "@types/supports-color": "^5.3.0",
     "rimraf": "~5.0.5",
-    "typescript": "~5.1.6"
+    "typescript": "~5.2.2"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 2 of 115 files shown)


jtpio:
Out of curiosity I tried to bump to TypeScript 5.7 directly but it looked like it would require a bit more work to adapt the code base.

So the motivation with this PR would be to go with incremental updates and see how far we could go for the 4.4.0 release.

buildutils/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "lib",
     "rootDir": "src",
-    "module": "commonjs",
+    "module": "Node16",
```
jtpio, line 6: This is the main change needed for the build to pass when updating from `5.1.x` to `5.2.x`.

jtpio: Some information about `node16`: https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext

jtpio:
Marking as ready as an incremental update.

buildutils/tsconfig.json
jtpio: Not sure we need to mention this change in the extension migration guide, since this applies to the `@jupyterlab/buildutils` package which is likely not heavily used by extension developers.

jtpio: Done in https://github.com/jupyterlab/jupyterlab/pull/17207/commits/52d3215688e5fd29acc1bc86a460ba07bd8f2a3c

jtpio: Yeah it doesn't hurt to leave a note about it.

krassowski (approved):
Thank you @jtpio!

Issue #17176, opened 2025-01-16
Browser tab's name doesn't get updated if Lab is opened in simple mode


krassowski:
## Description

If JupyterLab is opened using `/doc` URL (i.e. in simple mode) the tab name does not update when switching between documents. It looks like this is more reliably reproducible if there is a workspace redirection happening.

This does not happen if JupyterLab is opened in normal mode with `/lab` and then switched to simple mode.

## Reproduce

1. Create an `A.ipynb` and `B.ipynb`
2. Open http://127.0.0.1:8888/doc/tree/A.ipynb
3. Switch between notebooks
4. See browser title does **not** update

https://github.com/user-attachments/assets/777474c1-c416-4ccd-ab72-eedab11b4e95

## Expected behavior

It works the same way as it works:

In normal mode:

6. Open http://127.0.0.1:8888/lab/tree/A.ipynb
7. Switch between notebooks
8. See browser title does update

In simple mode if lab was loaded with normal mode:

9. Open http://127.0.0.1:8888/lab/tree/A.ipynb
10. Switch mode to simple
11. Switch between notebooks
12. See browser title does update

https://github.com/user-attachments/assets/6398ca82-aea3-4cea-9c40-203b2d0e8a15

## Context

- Browser and version: Chrome
- JupyterLab version: 4.3.4, 4.4.0a2

PR #17210, opened 2025-01-22
Fix browser tab name updates when started in single-document mode

Darshan808:
## References #17176 

In doc mode, `isDeferred` is set to true, causing `deferredMainArea` to be used for providing the tab title. However, `deferredMainArea` is always null and is never updated.

#### Solution
Instead of relying on `deferredMainArea`, the issue was resolved by invoking `_dehydrateMainArea()` every time.

#### Observations
There is a comment in the code stating:
> "Save the cached main area layout if restoration is deferred."

This suggests that `deferredMainArea` might have been intended to be updated elsewhere but isn't. While this solution resolves the issue, I am seeking feedback to verify whether `deferredMainArea` is indeed supposed to be updated elsewhere and if the current approach could have any unintended side effects.


krassowski:
Looking at changes in https://github.com/jupyterlab/jupyterlab/commit/1a86b74dc8410ba14b114a1dd74d2fa8afde825c I think `restoreDeferred` call may have been meant to do this? Maybe instead we should do:

```diff
      // If layout restoration has been deferred, restore layout now.
      if (this._layoutRestorer.isDeferred) {
        this._layoutRestorer
          .restoreDeferred()
          .then(mainArea => {
            if (mainArea) {
              const { currentWidget, dock } = mainArea;
              if (dock) {
                this._dockPanel.restoreLayout(dock);
              }
              if (currentWidget) {
                this.activateById(currentWidget.id);
              }
+            // we would need to get hold of `dehydrated` here
+            dehydrated.main = mainArea;
            }
          })
          .catch(reason => {
            console.error('Failed to restore the deferred layout.');
            console.error(reason);
          });
      }
```

In any case, I think it would be great to have a test case for it. I am not sure how to best test it but would presume there is a signal which should be emitted but is not?

CC @afshin who worked on the simple/single-document mode and was the person to last touch the relevant code

Darshan808:
Added a test for this. I placed it under general tests as I wasn’t sure of a better location. Open to feedback and suggestions about test implementation!

Darshan808:
The test passes locally, but CI error says `locator('.cm-editor:not(.jp-mod-readOnly)') resolved to 2 elements:`

Darshan808:
```
+            // we would need to get hold of `dehydrated` here
+            dehydrated.main = mainArea;
```

The above suggested code snippet is inside an if block that executes only when JupyterLab is opened in normal (multiple-document) mode. However, the issue occurs specifically when JupyterLab is opened in single-document mode, meaning the code in this scope doesn’t run in the problematic scenario.

I haven’t completely understood the purpose of deferred trackers or why they are deferred when JupyterLab is directly opened in single-document mode. However, I noticed that if we invoke restoreDeferred here:
```
    if (mode === 'single-document') {
      if (this._layoutRestorer.isDeferred) {
        this._layoutRestorer.restoreDeferred()
      }
```
https://github.com/jupyterlab/jupyterlab/blob/c8a53da7b114f58a915a68182aa4172d6aa30262/packages/application/src/shell.ts#L626-L644
 all trackers are restored, this.isDeferred returns false, and the issue is resolved.

I would appreciate your feedback on whether invoking restoreDeferred here is appropriate or if there’s a better approach to handle deferred trackers or deferredMainArea for single-document mode.

galata/test/jupyterlab/general.test.ts
```diff
@@ -53,4 +53,22 @@ test.describe('General Tests', () => {
...
+    baseURL
+  }) => {
+    // Switch to single-document mode
+    const currentUrl = page.url();
+    const docUrl = currentUrl.replace('/lab/', '/doc/');
+    await page.goto(docUrl);
+
+    // Create a new text file and wait for it to lodd
+    await page.menu.clickMenuItem('File>New>Text File');
+    await page.locator('.cm-editor:not(.jp-mod-readOnly)').waitFor();
```
krassowski, line 68: Would something like this work?

```suggestion
    await page.getByLabel('untitled.txt').locator('.cm-editor').waitFor();
```


Just guessing based on the error message

krassowski:
@afshin is there a good write-up or explanation about the deferred restoration somewhere? Any chance you could chime in here?

I believe the underlying issue also leads to other bugs like tracker signals not being properly emitted so I think this is a bug we should fix if we want the doc mode to no longer feel buggy in general.

krassowski:
> 4. The ?reset parameter is added to the URL so that the state database will be reset. All layout restoration data is lost. The tab title is not correct. This is because there is no deferred main area information, it has been reset.

I think the problem is that even when the notebook is changed later the title is still not updated. This is:

5. Open a new notebook using file browser, see the title is not changed.

In other words, it looks like when restoration fail because the layout was reset the signals for widget tracking do not get connected at all.

brichet:
For the context, IIRC https://github.com/jupyterlab/jupyterlab/pull/13037 has 3 main goals, when Jupyterlab **is opened in `doc` mode**:
- preventing the main area to be restored, to quickly load the expected document.
  *As far as I can tell, the current PR does not break this feature.*
- being able to restore all the main area widgets when switching back to normal mode
  *The current PR doesn't break either this feature.*
- preventing the restorer to overwrite the main area layout in the the `StateDB`. Otherwise, the next restart in `lab` would contain only the single document previously opened in `doc` mode.
  *The current PR does break this feature because it save the main area layout in doc mode*

> My suggestion would be to update the tab title in a better way rather than remove deferred restoration.

Agree :+1: 
Actually, I wonder why the update of the tab title is triggered on the `StateDB` changes and not directly on the shell layout changes: 
- the `StateDB` is not updated only for the layout, so the title is updated more often then necessary
- disabling [`'@jupyterlab/application-extension:layout'`](https://github.com/jupyterlab/jupyterlab/blob/9749b07d54b4ac7622e617f1afdce770849c6f92/packages/application-extension/src/index.tsx#L852) would make the title update triggered only on other `StateDB` update

I think that we should update the title on `labShell.layoutModified` signal directly, using the `ILabShell.ILayout` instead of the `StateDB` content. We could add it to the [shell plugin](https://github.com/jupyterlab/jupyterlab/blob/9749b07d54b4ac7622e617f1afdce770849c6f92/packages/application-extension/src/index.tsx#L1102) for example, or create a new plugin for the title update.

brichet:
Thanks @Darshan808 for updating the PR, but I think that the issue is the same as previously if we still rely on the data in the `StateDB`.
It is probably a bit more tricky to fix, because we need to rely on the labshell layout only, I think. This means performing an equivalent of `restorer._dehydrateMain` on change.

brichet:
> This means performing an equivalent of `LayoutRestorer._dehydrateMainArea()` on change.

To clarify this comment, I think that we need to be able to retrieve (1) the current document path and (2) the number of documents opened, using only the `ILabShell.ILayout`.

Darshan808:
Thanks @brichet, had to go through the code again to understand some stuff!  

I can think of two different solutions here:  

1) Since we only need the current widget info to provide the tab title, we could modify `dehydrated.main.current` to store the current widget when restoration is deferred. This way, we don’t fully `dehydrateMainArea` but retain the `current` attribute just for providing the tab title.  

```typescript
 dehydrated.main = this.isDeferred
      ? this._deferredMainArea
      : this._dehydrateMainArea(layout.mainArea);
    // Add this block
    if (this.isDeferred) {
      const currentWidget = layout.mainArea?.currentWidget;
      if (currentWidget) {
        const widgetName = Private.nameProperty.get(currentWidget);
        dehydrated.main = widgetName ? { current: widgetName } : {};
      }
    }
```

2) As you mentioned, we could explore further and try to fetch the current document path and the number of open documents through a different approach.  

What do you think about (1) ? Though, I was sometimes running into issues with it. It might be due to the `layoutModified` signal firing before the database was updated.

brichet:
Thanks @Darshan808 for sharing it.

I did some testing with you first proposition. It may work with fewer drawbacks, but it needs some change to work properly I think.

1. it still relies on the `StateDB` values, so we should revert https://github.com/jupyterlab/jupyterlab/pull/17210/commits/c3a02a1e1794c842e001bfdb98d7cc7ea0798930 and update the title on `StateDB` changes, to avoid side effects.

2. it fully rewrites the `dehydrated.main` value, which should be avoided to keep the state. I would change the last line to update only the `dehydrated.main.current` value.
  ```ts
   dehydrated.main = {...dehydrated.main, current: widgetName || undefined };
   ```

That way, only the `current` value is updated in the `StateDB`. The only drawback is that it may change the current widget. Let's say you opened `A.ipynb` and `B.ipynb`, and A is the current.
- If you start in doc mode with `C.ipynb`
  - :+1: it will restore if you switch to normal mode, because it use the `_deferredMainArea` to restore
  - :+1:  it will restore if you close jupyterlab and start again in normal mode, because the current widget is not in the dock, so it fallback on the current index in the dock
- If you start in doc mode with `B.ipynb`
  - :+1: it will restore if you switch to normal mode, because it use the `_deferredMainArea` to restore
  - :-1: if you close jupyterlab and start again in normal mode, `B.ipynb` will be the current widget instead of `A.ipynb`
 
I don't know what others think, but it sound reasonable to me, with less changes. 
cc @krassowski @afshin

brichet:
> In the second point here, what exactly does the current index represent

This is the index of the current widget in the `tab-area`.
The main area can contain several `tab-area` (using `split-area`). Each tab area has its own current widget (the `currentIndex`), and there is a main `current` to select the current widget in the main area.

packages/apputils-extension/src/index.ts
brichet, line 1: Thanks @Darshan808 for updating this PR.

Are the changes made to this file intentional?
They seems to revert some recent changes https://github.com/jupyterlab/jupyterlab/pull/17256.

Darshan808: No, those changes weren’t intentional — looks like they got in by mistake. Thanks for pointing that out!

galata/test/jupyterlab/general.test.ts
```diff
@@ -53,4 +53,22 @@ test.describe('General Tests', () => {
     await page.reload();
     expect(await page.theme.getTheme()).toEqual('JupyterLab Dark');
   });
+
+  test('Browser tab name updates to file name in /doc mode', async ({
+    page,
+    baseURL
```
brichet, line 59:
```suggestion
    page
```

galata/test/jupyterlab/general.test.ts
```diff
@@ -53,4 +53,22 @@ test.describe('General Tests', () => {
...
+  }) => {
+    // Switch to single-document mode
+    const currentUrl = page.url();
+    const docUrl = currentUrl.replace('/lab/', '/doc/');
+    await page.goto(docUrl);
+
+    // Create a new text file and wait for it to lodd
```
brichet, line 66:
```suggestion
    // Create a new text file and wait for it to load
```

brichet (approved):
Thanks @Darshan808, this looks good.

I added two minor comments.

afshin:
Thank you @Darshan808 and thank you @brichet and @krassowski for reviewing!

PR #17160, opened 2025-01-12
Ensure search highlight is applied to Python builtin keywords

hxrshxz:
## References

Fixes #17068 

## Code changes

updated the [jupyterlab/packages/codemirror/style/base.css](https://github.com/jupyterlab/jupyterlab/blob/2607c1093539850b5b618352dc94893a8b662e48/packages/codemirror/style/base.css#L70-L97) file : now the syntax is highlighted correctly

## User-facing changes

**Before** - 
Screenshot 2025-01-12 111825

**After** -
Screenshot 2025-01-12 110137

## Backwards-incompatible changes

None

PR #17214, opened 2025-01-23
Fix disabling Fuzzy Filtering in the File Browser

### LINKED ISSUE #17206: Disabling fuzzy filtering in file browser does not work
## Description

Disabling fuzzy filtering in the file browser does not work.

## Reproduce

1. Create files named "test" and "tst"
2. Filter for "tst"
3. See both are shown (ok)
4. Open Settings Editor, go to File Browser
5. Unclick "Filter on file name with a fuzzy search"
6. Filter again
7. See both are still shown

Image

The value is hardcoded here: https://github.com/jupyterlab/jupyterlab/blob/1f6f64bbd56bc9363315114a4fe0f642252ef464/packages/filebrowser/src/browser.ts#L106

## Expected behavior

Only tst shows up.

## Context

- Browser and version: Chrome
- JupyterLab version: 4.3.x and 4.4.0a2

Darshan808:
### Fixes #17206 

This PR addresses an issue where disabling fuzzy filtering in the file browser was not functioning due to the fuzzy filtering value being hardcoded to true. The implementation now uses signals to detect changes in the fuzzy filtering setting and updates the FilenameSearcher element accordingly.

A test case has been added to verify the behavior.

Please review the code changes and the current implementation, and provide feedback for any further improvements.


Darshan808:
The test feels too long; I'll simplify it to work with just a single text file.

galata/test/jupyterlab/settings.test.ts
```diff
@@ -422,3 +422,57 @@ test('Settings Import: Importing a JSON file applies the correct settings', asyn
...
+    .locator('label:has-text("Filter on file name with a fuzzy search")')
+    .click();
+
+  // Clear the filter input and apply a new filter for "test"
+  const searchBox = page.locator('input[placeholder="Filter files by name"]');
+  await searchBox.click();
+  const currentValue = await searchBox.inputValue();
+  for (let i = 0; i < currentValue.length; i++) {
+    await page.keyboard.press('Backspace');
+  }
```
krassowski, line 473: Could `selectText()` with a single backspace as in https://github.com/jupyterlab/jupyterlab/pull/17160/files be more efficient?

krassowski (commented):
+1 to switching to files over notebooks, files are faster to open.

I think that the overall test flow (creating two files) is fine.

packages/filebrowser/src/model.ts
```diff
@@ -786,6 +787,27 @@ export class FilterFileBrowserModel extends TogglableHiddenFileBrowserModel {
...
+  }
+  set useFuzzyFilter(value: boolean) {
+    this._useFuzzyFilter = value;
+    this._filterFileBrowserSettingsChanged.emit({ useFuzzyFilter: value });
+  }
+
+  /**
+   * Signal for settings changed
+   */
+  get filterFileBrowserSettingsChanged(): ISignal<
```
krassowski, line 804: I would suggest using a shorter name, we already know this pertains to file browser:

```suggestion
  get filterSettingsChanged(): ISignal<
```

packages/ui-components/src/components/search.tsx
```diff
@@ -215,6 +215,26 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
 /**
  * A widget which hosts a input textbox to filter on file names.
  */
-export const FilenameSearcher = (props: IFilterBoxProps): ReactWidget => {
-  return ReactWidget.create(<FilterBox {...props} />);
-};
+export class FilenameSearcher extends ReactWidget {
+  private filterBoxProps: IFilterBoxProps;
```
krassowski, line 219: Let's follow convention of prefixing private properties with `_` and placing them at the bottom

```suggestion
  private _filterBoxProps: IFilterBoxProps;
```

packages/filebrowser/src/model.ts
```diff
@@ -786,6 +787,27 @@ export class FilterFileBrowserModel extends TogglableHiddenFileBrowserModel {
...
 
+  /**
+   * Whether to apply fuzzy filter.
+   */
+  get useFuzzyFilter(): boolean {
+    return this._useFuzzyFilter;
+  }
+  set useFuzzyFilter(value: boolean) {
+    this._useFuzzyFilter = value;
+    this._filterFileBrowserSettingsChanged.emit({ useFuzzyFilter: value });
```
krassowski, line 798: Do we need to add this for `filterDirectories` setter too?

krassowski: Also, since this is emitting a signal indicating change, let's short-circuit if `this._useFuzzyFilter === value` already to avoid spurious re-renders :)

Darshan808:
I'm seeing merge conflicts on this PR. Could you guide me on the best way to resolve them?

krassowski:
This one looks simple as just another test was added in that file from the read-only fix PR. I would use the Resolve conflicts button on GitHub and made sure to remove the conflict markers and correct alignment. Alternatively you could move your test and place it above another test, which should help git figure out how to merge it cleanly.

packages/filebrowser/src/browser.ts
```diff
@@ -103,12 +104,20 @@ export class FileBrowser extends SidePanel {
...
       inputRef: this._fileFilterRef
     });
+    this.model.filterSettingsChanged.connect((value, args) => {
+      for (const key in args) {
+        searcher.updateProp(
+          key as keyof IFilterBoxProps,
+          args[key as keyof IFilterBoxProps]
+        );
+      }
+    }, this);
```
krassowski, line 120: Can you instead try to pass `filterSettingsChanged` signal down to `FilenameSearcherWidget` (as an optional prop) and when it fires:
1. update a private copy of props inside `FilenameSearcherWidget`
2. call `update()` to trigger re-render

This will allow us to remove the `updateProp` public method, limiting the public API surface (and in turn reducing the maintenance cost long-term).

packages/ui-components/src/components/search.tsx
```diff
@@ -215,6 +215,26 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
...
+  /**
+   * Update a specific prop.
+   */
+  updateProp<K extends keyof IFilterBoxProps>(
+    key: K,
+    value: IFilterBoxProps[K]
+  ): void {
+    this._filterBoxProps[key] = value;
+    this.update();
+  }
```
krassowski, line 233: This does not need to be public if we pass the signal down to this class.

packages/filebrowser/src/browser.ts
Darshan808: Yes, this sounds good. But when trying to import `FileBrowserModel` from `@jupyterlab/filebrowser` into `search.tsx` of `@jupyterlab/ui-components`, a dependency cycle error occurs, causing the build to fail.
  cyclePath: [
    '@jupyterlab/application-top',
    '@jupyterlab/application',
    '@jupyterlab/apputils',
    '@jupyterlab/statusbar',
    '@jupyterlab/ui-components',
    '@jupyterlab/filebrowser',
    '@jupyterlab/apputils'
  ]

Is there a recommended approach to resolve this issue?

krassowski: You do not need `FileBrowserModel` here. You should define an expected `ISignal` and expected interface of arguments in `search.tsx`

krassowski: Like `ISignal<unknown, IFilterBoxProps>` as emitter can be any

packages/ui-components/src/components/search.tsx
```diff
@@ -140,11 +149,11 @@ export const updateFilterFunction = (
       value = value.toLocaleLowerCase();
     }
     const i = item.indexOf(value);
-    if (i === -1) {
+    if (i != 0) {
```
krassowski, line 152: I do not follow this change?

packages/ui-components/src/components/search.tsx
```diff
@@ -215,6 +224,43 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
...
+class FilenameSearcherWidget extends ReactWidget {
+  constructor(props: IFilterBoxProps) {
+    super();
+    props?.filterSettingsChanged?.connect((_, args) => {
+      for (const key in args) {
+        this._updateProp(
+          key as keyof IFilterBoxProps,
+          args[key as keyof IFilterBoxProps]
+        );
+      }
```
krassowski, line 236: This will call `update()` each time in each `_updateProp` call. Can extract update to happen after the for loop?

krassowski (commented):
Looking good!

Darshan808:
The previous logic matched substrings at any index and all except the first letter of the filename were made bold. This has been corrected.

packages/ui-components/src/components/search.tsx
Darshan808: If fuzzy search is disabled, filenames are shown only if they start with the searched value. Isn't this expected?

krassowski:
> The previous logic matched substrings at any index

This is expected I think. This is what I would expect:

| Fuzzy? | Search term | `canonical correlation analysis.ipynb` match? | `principal component analysis.ipynb` match? |
|--| --|--|--|
| Yes | pca | N | Y |
| Yes | cca | Y | N |
| No | analysis | Y | Y |
| No | factorization | N | N |

> all except the first letter of the filename were made bold. 

This sounds like a bug indeed.

packages/ui-components/src/components/search.tsx
```diff
@@ -215,6 +224,35 @@ export const FilterBox = (props: IFilterBoxProps): JSX.Element => {
...
+
+  private _filterBoxProps: IFilterBoxProps;
+}
+
+/**
+ * Function which returns a widget that hosts an input textbox to filter on file names.
+ */
+export const FilenameSearcher = (
+  props: IFilterBoxProps
+): FilenameSearcherWidget => {
```
krassowski, line 256: While annotating it as returning `FilenameSearcherWidget` is better, especially in new code, here I would be conservative and keep it as it was if possible (which I think is possible) just not to break downstream users when backporting this fix.

```suggestion
export const FilenameSearcher = (props: IFilterBoxProps): ReactWidget => {
```

krassowski (approved):
Thank you @Darshan808!

krassowski:

PR #17217, opened 2025-01-24
Improve UX for "Rename on First Save" Dialog

### LINKED ISSUE #16917: Remove "Do not ask me again" checkbox from rename on first save dialog

MUFFANUJ:
## References
fixes #16917 

## User-facing changes
Reworked the dialog to improve clarity and resolve unintuitive behaviors:

Replaced the "Do not ask me again" checkbox with  Do not ask for rename on first save.
Updated options:
- Save as Untitled: Saves the file with the default untitled name.
- Rename and Save: Prompts for a name, renames the file, and saves it..
- Removed ambiguity around "Cancel" to prevent unintended Ctrl + S repetition by users.

 ### Before 
WhatsApp Image 2025-01-24 at 17 50 17

### After
WhatsApp Image 2025-01-24 at 18 07 47


packages/docmanager-extension/src/index.tsx
```diff
@@ -916,12 +916,13 @@ function addCommands(
           ) {
             const result = await InputDialog.getText({
               title: trans.__('Rename file'),
-              okLabel: trans.__('Rename'),
+              okLabel: trans.__('Rename and Save'),
               placeholder: trans.__('File name'),
+              cancelLabel: trans.__(`Save as ${oldName}`),
```
krassowski, line 921: We will need to pass the `oldName` as separate argument for it to be translatable, see https://jupyterlab.readthedocs.io/en/stable/extension/internationalization.html#internationalization-and-localization

krassowski (commented):
Thanks for working on it!

This looks very clear to me but also a bit too long for smaller screens.

Could you explore how feasible it would be to move the checkbox to above buttons (for this dialog only)?

MUFFANUJ:
Thanks for the feedback,
I’ve made the changes specifically for dialogs where the string length exceeds 35 words, as these can look cramped on smaller screens. Moving the checkbox above the buttons should resolve this issue in these cases. I’ve also checked other instances where the checkbox label exceeds 35 words, and I believe this layout could help avoid any awkward display on smaller screens.

Looking at this, it seems like this approach should work for the current use cases, but I’ll make further adjustments if needed to ensure a more seamless experience. If you'd prefer a more specific approach rather than this general solution based on the checkbox label length, please let me know.

I’ve attached an updated demo for better understanding.

https://github.com/user-attachments/assets/6729fcbc-5203-4828-a21d-ffe78f7f77dc

packages/docmanager-extension/src/index.tsx
MUFFANUJ: Thank you for letting me know about this. I have fixed it by updating the translation function to pass the oldName as a separate argument, as suggested. Everything should be working fine now.

krassowski:
Edited the top-level comment for the "fixes" linking keyword to apply properly

packages/docmanager-extension/src/index.tsx
```diff
@@ -916,12 +916,13 @@ function addCommands(
           ) {
             const result = await InputDialog.getText({
               title: trans.__('Rename file'),
-              okLabel: trans.__('Rename'),
+              okLabel: trans.__('Rename and Save'),
               placeholder: trans.__('File name'),
+              cancelLabel: trans.__('Save as %1', oldName),
```
krassowski, line 921: Not sure about this, but maybe we do not need to show the extension on the label?

joaopalmeiro:
Hi! 👋 

Very happy to see improvements to this dialog! Quick suggestion based on the discussion on #17321:

To better balance the dialog, I think I would either make the input element take up the entire available width or put the input checkbox and buttons on separate rows.

What do you think?

MUFFANUJ:
hey @joaopalmeiro, I was thinking the same way to keep buttons together in single line 
-> Before 
Screenshot 2025-03-03 at 5 30 57 PM
Screenshot 2025-03-03 at 5 48 22 PM

-> After 
Screenshot 2025-03-03 at 5 50 53 PM

Screenshot 2025-03-03 at 5 46 19 PM


Does this look good or need some other modification ?

anujmst:
@krassowski any suggestions

MUFFANUJ:
Hey @krassowski any suggestion ?

krassowski:
The screenshots look good to me

MUFFANUJ:
@krassowski please review the changes

packages/apputils/style/dialog.css
```diff
@@ -17,6 +17,7 @@
   width: 100%;
   height: 100%;
   background: var(--jp-dialog-background);
+  container-type: inline-size;
```
krassowski, line 20:
```suggestion
  /* stylelint-disable */
  container-type: inline-size;
  /* stylelint-enable */
```

packages/apputils/style/dialog.css
```diff
@@ -154,3 +155,11 @@ button.jp-Dialog-close-button {
 .jp-Dialog-button.jp-mod-styled:not(:last-child) {
   margin-right: 12px;
 }
+
+@container (max-width: 560px) {
```
krassowski, line 159:
```suggestion
/* stylelint-disable */
@container (max-width: 560px) {
  /* stylelint-enable */
```

krassowski (commented):
It looks current version of `stylelint` does not know about `container-type` (nor `@container`), lets disable it on these lines

krassowski (approved):
I think this is a good incremental improvement, thank you @MUFFANUJ.

PR #14491, opened 2023-05-04
Don't listen for cell resizing when notebooks is hiding

fcollonval:
## References

Fixes #14461 and fixes #14489

## Code changes

Stop listening for cell resizing if a notebook is hidden.

## User-facing changes

The viewport is not updated when the widget is hidden.

## Backwards-incompatible changes

None

PR #17221, opened 2025-01-26
Improving reliability of the the galata tests

Darshan808:
## References #14947

The two tests `Debugger › Breakpoints on exception` and `windowed-notebook.test.ts› should not update height when hiding` aren't reliable keeps failing on CI. This PR aims to increase their reliability.

#### 1.  `Debugger › Breakpoints on exception`
This test verifies that the code execution pauses when some kind of exception is raised. The test compares a snapshot of the cell with the expected output. However, the red debug box that appears on breakpoints occasionally takes some time to load, causing the snapshots to mismatch and the CI to fail.

To resolve this, a small time delay has been added to ensure the red debug box is fully loaded before the snapshot is captured.

**Before:**
Screenshot from 2025-01-26 15-02-50

**After:**
Screenshot from 2025-01-26 15-02-38

#### 2.  `windowed-notebook.test.ts› should not update height when hiding`

This test is designed to verify that the height of a large notebook does not change when it is hidden (e.g., by opening a new tab). It compares the notebook's height before and after hiding it.

However, during local testing, I observed that the notebook's height continues to increase even after it is hidden. This suggests that the notebook's rendering process is still ongoing in the background. The test currently passes because the height comparison occurs immediately after switching tabs, before the rendering completes. If a delay were introduced before the height verification, the test would consistently fail due to the height continuing to increase.

I am still investigating this issue and working to identify a solution. Any insights or suggestions would be greatly appreciated!

### DIFF
--- a/galata/test/documentation/debugger.test.ts
+++ b/galata/test/documentation/debugger.test.ts
@@ -160,6 +160,8 @@ test.describe('Debugger', () => {
 
     // Wait to be stopped on the breakpoint
     await page.debugger.waitForCallStack();
+    // Wait for the red debug indicator box to appear
+    await page.waitForTimeout(50);
     expect(
       await page.screenshot({
         clip: { y: 110, x: 300, width: 300, height: 80 }
@@ -182,6 +184,8 @@ test.describe('Debugger', () => {
 
     // Wait to be stopped on the breakpoint
     await page.debugger.waitForCallStack();
+    // Wait for the red debug indicator box to appear
+    await page.waitForTimeout(50);
     expect(
       await page.screenshot({
         clip: { y: 110, x: 300, width: 300, height: 80 }
--- a/galata/test/documentation/debugger.test.ts
+++ b/galata/test/documentation/debugger.test.ts
@@ -161,7 +161,11 @@ test.describe('Debugger', () => {
     // Wait to be stopped on the breakpoint
     await page.debugger.waitForCallStack();
     // Wait for the red debug indicator box to appear
-    await page.waitForTimeout(50);
+    const firstCell = (await page.notebook.getCellLocator(0))!;
+    await firstCell.locator('.jp-DebuggerEditor-highlight').waitFor({
+      state: 'visible',
+      timeout: 1000
+    });
     expect(
       await page.screenshot({
         clip: { y: 110, x: 300, width: 300, height: 80 }
@@ -185,14 +189,21 @@ test.describe('Debugger', () => {
     // Wait to be stopped on the breakpoint
     await page.debugger.waitForCallStack();
     // Wait for the red debug indicator box to appear
-    await page.waitForTimeout(50);
+    await firstCell.locator('.jp-DebuggerEditor-highlight').waitFor({
+      state: 'visible',
+      timeout: 1000
+    });
     expect(
       await page.screenshot({
         clip: { y: 110, x: 300, width: 300, height: 80 }
       })
     ).toMatchSnapshot('debugger_stop_on_raised_exception.png');
     await page.click('jp-button[title^=Continue]');
+    await page.debugger.waitForCallStack();
     await page.click('jp-button[title^=Continue]');
+    await page.debugger.waitForCallStack();
+    await page.click('jp-button[title^=Continue]');
+    await page.notebook.waitForRun(0);
   });
 
   test('Debugger sidebar', async ({ page, tmpPath }) => {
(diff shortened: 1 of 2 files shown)


Darshan808:
> However, during local testing, I observed that the notebook's height continues to increase even after it is hidden.

I said this because when i modify the test to:
```
  await page.menu.clickMenuItem('File>New Launcher'); \\Hides the notebook
  do {
    previousHeight = initialHeight;
    await page.waitForTimeout(200);

    initialHeight = await getInnerHeight(notebook!);
    console.log(`Counter: ${counter}, initialHeight: ${initialHeight}`) \\ Monitor height
  } while (counter++ < 3);
```
I get an output
```
Counter: 0, initialHeight: 2566
Counter: 1, initialHeight: 4313 \\ Increase in height is observed
Counter: 2, initialHeight: 4313 \\ MaxHeight
Counter: 3, initialHeight: 4313

```

Darshan808:
The test is failing in this PR, which ironically aims to fix the issue! 😄
Got some hints about the failure, investigating the root cause.

galata/test/documentation/debugger.test.ts
```diff
@@ -160,6 +160,8 @@ test.describe('Debugger', () => {
 
     // Wait to be stopped on the breakpoint
     await page.debugger.waitForCallStack();
+    // Wait for the red debug indicator box to appear
+    await page.waitForTimeout(50);
```
krassowski, line 164: In general adding or increasing `page.waitForTimeout` will not help us much, except it will be less flaky. Instead, we need to find out what selector we want to await and add an await for that selector.

Darshan808:
Three clicks on the `Continue` button were required to fully exit the cell, as both exception options (`userUnhandled` and `raised`) were selected. This caused the debug box to appear twice for the `2/0` line inside the `except` block (once for `raised` and again for `userUnhandled`)

I’m not sure if this is expected behavior or a bug. Could you please confirm?

krassowski:
In that case I would think this is expected? CC @brichet who developed this feature/test in case if I am wrong.

@Darshan808 maybe you could add a comment in test code explaining why three clicks are required?

brichet:
> I’m not sure if this is expected behavior or a bug. Could you please confirm?

I don't remember this behavior precisely, but it seems normal that there are 3 pauses with the 2 options selected. I probably missed the third click when including this test.

brichet:
> However, during local testing, I observed that the notebook's height continues to increase even after it is hidden.

AFAIU the test is designed to make sure that the windowed inner height will not change when hidden **after** the first initialization. I think it's normal for the height to increase during the initialization phase, even if it's hidden.

galata/test/jupyterlab/windowed-notebook.test.ts
```diff
@@ -91,7 +91,7 @@ test('should not update height when hiding', async ({ page, tmpPath }) => {
   // Wait to ensure the rendering logic is stable.
   do {
     previousHeight = initialHeight;
-    await page.waitForTimeout(200);
+    await page.waitForTimeout(300);
```
brichet, line 94: Increasing the timeout here should not be an issue, it's only to ensure that the notebook is initialized.

brichet (approved):
Thanks @Darshan808, this looks good to me.

Darshan808:
> AFAIU the test is designed to make sure that the windowed inner height will not change when hidden after the first initialization. 
Thanks for the reply.
But the test is flaky. It might be because after the notebook rendering stabilizes, hiding it by opening a new launcher takes around 160ms on my device. During this time, the notebook height might slightly increase, causing a difference between the stored stable height and the retrieved height, leading to test failures. What do you think about this?

Darshan808:
> AFAIU the test is designed to make sure that the windowed inner height will not change when hidden after the first initialization. 
Thanks for the reply.
But the test is flaky. It might be because after the notebook rendering stabilizes, hiding it by opening a new launcher takes around 160ms on my device. During this time, the notebook height might slightly increase, causing a difference between the stored stable height and the retrieved height, leading to test failures. What do you think about this?

galata/test/jupyterlab/windowed-notebook.test.ts
Darshan808: The test was passing when I did this, but I later realized that increasing the timeout was causing the notebook to render fully on initialization, which resulted in its height remaining the same before and after hiding.

Darshan808:
> The test currently passes because the height comparison occurs immediately after switching tabs, before the rendering completes. If a delay were introduced before the height verification, the test would consistently fail due to the height continuing to increase.

Adding more cells during testing still shows the same behavior.
image

krassowski:
Should we merge as-is if it improves `Breakpoints on exception` test case already? There seems to be a bigger issue with the logic that `should not update height when hiding` is testing.

I think this is a bug in code, not in the test. For reference it was added in https://github.com/jupyterlab/jupyterlab/pull/14491 and since quite a few changes were made to the windowed mode.

Darshan808:
This sounds good. We can track it in a separate PR.

brichet:
Thanks @Darshan808

PR #17230, opened 2025-01-27
Fix Inconsistent Behavior for Toolbar Button

### LINKED ISSUE #17229: Inconsistent State Toggle Behavior Between Mouse Click and Keyboard Events
## Description
When interacting with the `...` button on the toolbar (which appears when the screen width is smaller) using a mouse click, the state toggles correctly. However, when activating the button using the Spacebar or Enter key, the behavior is inconsistent, and the state does not update as it does with a mouse click.

## Reproduce

1. Reduce the screen width until the ... button appears on the toolbar.
2. Interact with the button using:
        - Mouse Click: The state updates correctly.
        - Keyboard Events (Spacebar/Enter): The state does not update as expected.

## Expected behavior
The button should behave consistently, regardless of whether it is activated using a mouse click, the Spacebar, or the Enter key.

## Context

- Operating System and version: macOS
- Browser and version:  Chrome 
- JupyterLab version: 4.0.0rc0 and earlier

Have a look at https://github.com/jupyterlab/jupyterlab/pull/17222#issuecomment-2615613179 for a better understanding.

MUFFANUJ:
## References
fixes #17229 

This fix ensures the ... button on the toolbar behaves consistently when activated using a mouse click, the Spacebar, or the Enter key. Previously, keyboard interactions did not toggle the state as expected.

## Code changes
Added event.preventDefault() in the handleKeyDown function to prevent the default browser behavior for Spacebar and Enter key events.

```tsx
const handleKeyDown = (event: React.KeyboardEvent) => {
  const { key } = event;
  if (key === 'Enter' || key === ' ') {
    event.preventDefault();
    props.onClick?.();
  }
};
```

### Demo after changes

https://github.com/user-attachments/assets/40bcdfea-7eb6-4396-9144-e477ce21163e

### DIFF
--- a/packages/ui-components/src/components/toolbar.tsx
+++ b/packages/ui-components/src/components/toolbar.tsx
@@ -842,6 +842,7 @@ export function ToolbarButtonComponent(
   const handleKeyDown = (event: React.KeyboardEvent) => {
     const { key } = event;
     if (key === 'Enter' || key === ' ') {
+      event.preventDefault();
       props.onClick?.();
     }
   };


MUFFANUJ:
@meeseeksdev tag bug

krassowski:
An alternative would be to call `event.preventDefault();` in `onClick` handlers. This would allow toolbar item owners to decide if they want it to happen. Would this be possible to only add it for the toggle button?

krassowski:
Ok, let's rephrase this:
- this PR adds `event.preventDefault()` in `handleKeyDown` of `ToolbarButtonComponent` which means it changes behaviour of all toolbar buttons
- but the issue only affects one button - the one which toggles the extra items
- adding `event.preventDefault()` to all buttons call might break some extensions; I have no time to test all extensions
- I note that `ToolbarButtonComponent.handleKeyDown` calls `props.onClick`
- instead of applying this change to _all_ buttons, I propose the props of the affected button to include `onClick` which will call `event.preventDefault()`

Does it make sense now?

PR #17239, opened 2025-01-29
Cache item state to improve filebrowser's performance 

### LINKED ISSUE #16767: Changing selected file in file browser is slow
## Description

It can take a couple of seconds to select a file down there without mouse by using <kbd>Down</kbd> in folders with hunderds of files

## Reproduce

1. Open JupyterLab node_modules director
2. Select a folder in the the filebrowser
3. Press arrow down for 3 seconds
4. See the delay in the render state

image

Zoomed in:

image

## Expected behavior

No delay.

Basically we should only refresh two items (the one unselected + the one selected), not the entire listing.

In case of multi-selection we should re-render the changed items only as well.

## Context

- JupyterLab version: 4.3.0b1

Rishab87:
## References
Fixes #16767 

## Code changes
I've added a cache to check for last state of the node if their are no changes it doesn't rerender, used weakmap so its collected by garbage collector when references of particular node is removed

## User-facing changes
Smooth UI of filebrowser

https://github.com/user-attachments/assets/fe5e9198-43ae-411e-87b5-707274bbc8f8

None

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -2864,7 +2864,6 @@ export namespace DirListing {
    * The default implementation of an `IRenderer`.
    */
   export class Renderer implements IRenderer {
-    private lastRenderedState = new WeakMap<HTMLElement, string>();
     /**
      * Create the DOM node for a dir listing.
      */
@@ -3160,11 +3159,14 @@ export namespace DirListing {
       translator = translator || nullTranslator;
       const trans = translator.load('jupyterlab');
 
-      const prevState = this.lastRenderedState.get(node);
+      const prevState = this._lastRenderedState.get(node);
       const newState = JSON.stringify({
         name: model.name,
         selected,
         lastModified: model.last_modified,
+        modifiedStyle,
+        hiddenColumns,
+        columnsSizes,
         fileSize: model.size
       });
 
@@ -3179,7 +3181,7 @@ export namespace DirListing {
       if (checkbox) checkbox.checked = selected ?? false;
 
       if (prevState === newState) return;
-      this.lastRenderedState.set(node, newState);
+      this._lastRenderedState.set(node, newState);
 
       const iconContainer = DOMUtils.findElement(node, ITEM_ICON_CLASS);
       const text = DOMUtils.findElement(node, ITEM_TEXT_CLASS);
@@ -3484,6 +3486,8 @@ export namespace DirListing {
       HTMLElement,
       { date: string; style: Time.HumanStyle }
     >();
+
+    private _lastRenderedState = new WeakMap<HTMLElement, string>();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


krassowski (commented):
Great idea! Later we could also add some debouncer/throttler so that it is also smooth when user is just keeping arrow down/up pressed.

packages/filebrowser/src/listing.ts
```diff
@@ -3153,13 +3154,33 @@ export namespace DirListing {
...
       const { icon, iconClass, name } = fileType;
       translator = translator || nullTranslator;
       const trans = translator.load('jupyterlab');
 
+      const prevState = this.lastRenderedState.get(node);
+      const newState = JSON.stringify({
+        name: model.name,
+        selected,
+        lastModified: model.last_modified,
+        fileSize: model.size
```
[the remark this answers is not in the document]
Rishab87: yes, I've addded that too

packages/filebrowser/src/listing.ts
```diff
@@ -3153,13 +3154,33 @@ export namespace DirListing {
...
       }
-
       fileType =
         fileType || DocumentRegistry.getDefaultTextFileType(translator);
       const { icon, iconClass, name } = fileType;
       translator = translator || nullTranslator;
       const trans = translator.load('jupyterlab');
 
+      const prevState = this.lastRenderedState.get(node);
+      const newState = JSON.stringify({
```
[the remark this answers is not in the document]
Rishab87: I tried this and saw a decrease in performance when using JSONExt.deepEqual (measured the performance with chrome dev tool, also I tried adding a debouncer on updateItemNode but on adding a debouncer few elements were getting skipped when holding down the arrow key which felt laggy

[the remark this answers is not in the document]
Rishab87: I also managed to add a debouncer but a lot of test cases are failing after that I'll try to resolve them and push the changes

Rishab87:
@krassowski added a debouncer too and it works fine, its not laggy!

Rishab87:
@krassowski just a follow up, anything else I need to change here

packages/filebrowser/src/listing.ts
```diff
@@ -970,17 +971,19 @@ export class DirListing extends Widget {
...
+          item,
+          ft,
+          this.translator,
+          this._hiddenColumns,
+          this.selection[item.path],
+          this._modifiedStyle,
+          this._columnSizes
+        );
+      }, 200);
+      debounceUpdateItemNodes.invoke();
```
krassowski, line 986: I think this would not work. The debouncer needs to be defined once and invoked multiple times to actually debounce, otherwise it is just delaying execution.

I also think that the debounce period may need to be shorter but I am not sure yet. I think this depends on OS/keyboard-set `AutoRepeatRate`.

krassowski: Also, the invoke should take the arguments rather than act as a closure. What about implementing debouncing in a separate PR?

Rishab87: yeah sure I'll keep these points in mind and implement debounce in a seperate PR so should I revert these changes related to debounce for now?

krassowski: I think so, I find the changes in this PR without debouncing worth releasing for testing as-is

Rishab87: I've reverted the changes so my next steps should be to create an issue for debouncer with a draft PR right?

krassowski (approved):
Thank you @Rishab87!

PR #17246, opened 2025-01-30
try to build a real wheel in RTD

bollwyvl:


krassowski:
> don't know if anyone was looking at any of the reporting (bundle size, etc)

Can't look at these things if binder does not boot up ;) I did not look at these in the past three months.

bollwyvl:
Huzzah:

https://jupyterlab--17246.org.readthedocs.build/en/17246/_static/dist/environment.yml

To avoid more churn, will take a look at a gist that does the thing... first step is to get to something declarative (no `postBuild`)

PR #17253, opened 2025-02-03
Add commands to change the console prompt position to the palette

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/13837

## Code changes

Add the commands to change the prompt cell position to the command palette.

## User-facing changes

image

None


jtpio:
> The icons do not look great in the command palette when selected. It seems that most other commands hide icons for palette, should we also do it in this case?

We can indeed hide them. This would also make it more consistent with other commands.

krassowski (approved):
LGTM, thank you!

jtpio:
Thanks for the review!

PR #17254, opened 2025-02-03
Update to Playwright 1.50

jtpio:
## References

Mechanical update to the latest stable version.

New features: https://github.com/microsoft/playwright/releases/tag/v1.50.0

## Code changes

Update to Playwright 1.50.1.

## User-facing changes

None

None


galata/test/jupyterlab/general.test.ts-snapshots/simple-mode-jupyterlab-linux.png
jtpio, line 1: Looks like this snapshot had to be updated for the CI to pass (subtle small change to the Jupyter top left logo)

krassowski (approved):
Thank you @jtpio!

krassowski:
Surfacing breaking changes (which I think are acceptable/should not cause issues):

```
- expect(locator).toBeEditable() and locator.isEditable() now throw if the target element is not <input>, <select>, or a number of other editable elements.
- Option testConfig.updateSnapshots now updates all snapshots when set to all, rather than only the failed/changed snapshots. Use the new enum changed to keep the old functionality of only updating the changed snapshots.
```

jtpio:
> Surfacing breaking changes (which I think are acceptable/should not cause issues):
> 
> ```
> - expect(locator).toBeEditable() and locator.isEditable() now throw if the target element is not <input>, <select>, or a number of other editable elements.
> - Option testConfig.updateSnapshots now updates all snapshots when set to all, rather than only the failed/changed snapshots. Use the new enum changed to keep the old functionality of only updating the changed snapshots.
> ```

Right, maybe we could leave a quick note about the playwright updates in the extension dev when crafting the user facing changelog for 4.4.0.

Opened https://github.com/jupyterlab/jupyterlab/issues/17260 to track this.

PR #17255, opened 2025-02-03
Update to TypeScript 5.4

jtpio:
## References

Building on top of https://github.com/jupyterlab/jupyterlab/pull/17207.

Incremental step towards https://github.com/jupyterlab/jupyterlab/issues/16174.

Keeping separate from https://github.com/jupyterlab/jupyterlab/pull/17207 so it's easier to find issues if there is any with these kind of updates.

## Code changes

- [x] Update to `typescript@~5.4.5`
- [x] Update extension migration guide to mention TypeScript 5.4

The bump from 5.2 to 5.4 does not seem to require any other change.

## User-facing changes

None

None

### DIFF
--- a/builder/package.json
+++ b/builder/package.json
@@ -71,7 +71,7 @@
     "@types/node": "^20.11.27",
     "@types/supports-color": "^5.3.0",
     "rimraf": "~5.0.5",
-    "typescript": "~5.2.2"
+    "typescript": "~5.4.5"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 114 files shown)


krassowski (approved):
Thank you @jtpio. Based on TypeScript changelogs this should be a safe change with potential to speed up the build time.

> I was timing it yesterday and it seems that it does reduce the build time once cache is set up. I do not see these gain on CI though.

jtpio:
Nice, thanks for testing!

Also looking forward to updating to more recent releases, which may help speed up the build time even more: https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#support-for-v8-compile-caching-in-node.js

PR #17256, opened 2025-02-03
Add (opt-in) workspace selector, show workspace name in the title

### LINKED ISSUE #16303: Current workspace indicator for non-default workspace
### Problem

JupyterLab 4.2.0 includes UI for workspaces (https://github.com/jupyterlab/jupyterlab/pull/15946), which now can be switched from the sidebar:

image

While it indicates the current workspace with a little checkmark, this is often not visible because the sidebar may be hidden or showing a different panel. Often when switching the workspace the sidebar will go away because the layout of the target workspace may be different.

### Proposed Solution

Show the workspace indicator when the workspace is different from default (and different from an auto-generated workspace?)

For example, one of the earlier designs (https://github.com/jupyterlab/jupyterlab/issues/6944#issuecomment-649992115) included a workspace switcher in the top bar, next to the menu:

indicator

It could be just an indicator, or also a switcher:

indicator

Alternatively, it could go on the status bar.

### Additional context

Workspace name is often shown on the window title bar, which we already do.

Some applications also show the workspace name in the file browser.

Darshan808:
### Description:

**Workspace Display & Toggle:**
The topbar has been enhanced to display the current workspace and allow users to toggle between available workspaces seamlessly.

**Tab Title Update:**
The tab title now includes the current workspace as a suffix, making it easier to identify the workspace context.

**Snapshots**
Screenshot from 2025-02-04 12-30-18

image

Any feedback or improvement is appreciated.


Darshan808:
Should we display the toggle element only when the workspace differs from 'default' or an auto-generated one, or should we show it whenever the user has created a custom workspace, regardless of the active workspace?

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +48,41 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+    const openWorkspace = (workspace: string) => {
+      app.commands.execute(commands.open, { workspace: workspace });
+    };
+    const widget = new WorkspaceSelectorWidget({
+      currentWorkspace: resolver.name,
+      identifiers: model.identifiers,
+      openWorkspace: openWorkspace,
+      model: model
+    });
+    app.shell.add(widget, 'top', { rank: 1000 });
```
jtpio, line 74: Wondering if it should be placed in the top bar by default, or made opt-in? Via a setting or a toggle in the running panel.

Not sure workspaces is a feature used by many users, and a new dropdown in the top area may take significant space and add distraction?

krassowski: So the thinking is that many users find that their tabs "are gone" if they open JupyterLab in a new window. This is because they were routed to another workspace. Exposing non-default workspace would allow them to recover their workspace easily.  In a sense this would be an important distraction. However, I agree that if we are showing it as opt-out it should occupy much less space.

Maybe we should add the it as opt-in for 4.4 and decide on whether to make it opt-out for next release later?

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => document.removeEventListener('mousedown', handleClickOutside);
+  }, []);
+
+  return (
+    <div className="jp-WorkspaceSelector" ref={dropdownRef}>
+      <button
+        className="jp-WorkspaceSelectorHeader"
+        onClick={() => setIsExpanded(!isExpanded)}
+        aria-expanded={isExpanded}
```
krassowski, line 55: Thank you for including `aria-expanded` here!

It looks like we are implementing "Editable Combobox With List Autocomplete" https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

```suggestion
        aria-expanded={isExpanded}
        aria-controls={dropdownId}
```

where `const dropdownId = useId()` (https://react.dev/reference/react/useId).

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+              type="text"
+              className="jp-WorkspaceSelectorInput"
+              placeholder="Search"
+              value={searchQuery}
+              onChange={e => setSearchQuery(e.target.value)}
+              autoFocus
+            />
+          </div>
+
+          <ul className="jp-WorkspaceSelectorList">
```
krassowski, line 83:
```suggestion
          <ul className="jp-WorkspaceSelector-list" role="listbox" aria-label={trans.__("Workspace")}>
```

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <div className="jp-WorkspaceSelectorSearchIcon">
+              <LabIcon.resolveReact icon={searchIcon} />
+            </div>
+            <input
+              type="text"
+              className="jp-WorkspaceSelectorInput"
+              placeholder="Search"
+              value={searchQuery}
+              onChange={e => setSearchQuery(e.target.value)}
+              autoFocus
```
krassowski, line 79: (this is my best effort on getting the accessibility roles right based on https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/)

```suggestion
              autoFocus
              aria-autocomplete="list"
              role="combobox"
```

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+      {isExpanded && (
+        <div className="jp-WorkspaceSelectorDropdown">
+          <div className="jp-WorkspaceSelectorSearch">
+            <div className="jp-WorkspaceSelectorSearchIcon">
+              <LabIcon.resolveReact icon={searchIcon} />
+            </div>
+            <input
+              type="text"
+              className="jp-WorkspaceSelectorInput"
+              placeholder="Search"
```
krassowski, line 76: Needs translation. Maybe "Search workspace"?

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+              onChange={e => setSearchQuery(e.target.value)}
+              autoFocus
+            />
+          </div>
+
+          <ul className="jp-WorkspaceSelectorList">
+            {filteredIdentifiers.map(identifier => (
+              <li
+                key={identifier}
+                className="jp-WorkspaceSelectorItem"
```
krassowski, line 87:
```suggestion
                className="jp-WorkspaceSelector-item"
                role="option"
```

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+    super();
+    this.id = 'jp-workspace-top-indicator';
+    this.addClass('jp-react-widget');
+    this.identifiers = props.identifiers;
+    this.openWorkspace = props.openWorkspace;
+    this.currentWorkspace = props.currentWorkspace;
+    props.model?.refreshed.connect(() => {
+      this.identifiers = props.model?.identifiers!;
+      this.update();
+    });
```
krassowski, line 127: Let's not introduced any non-null assertions unless in test. Here is appears unnecessary at a first glance.

```suggestion
    props.model.refreshed.connect(() => {
      this.identifiers = props.model.identifiers;
      this.update();
    });
```

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+ */
+export class WorkspaceSelectorWidget extends ReactWidget {
+  private currentWorkspace: string;
+  private identifiers: string[];
+  private openWorkspace: (workspace: string) => void;
+
+  constructor(props: IWorkspaceSelectorProps) {
+    super();
+    this.id = 'jp-workspace-top-indicator';
+    this.addClass('jp-react-widget');
```
krassowski, line 120: This appears unused

```suggestion
```

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+};
+
+/**
+ * A Widget for Workspace Selector at topbar
+ */
+export class WorkspaceSelectorWidget extends ReactWidget {
+  private currentWorkspace: string;
+  private identifiers: string[];
+  private openWorkspace: (workspace: string) => void;
```
krassowski, line 115: Let's add `_` prefix to private attributes and move it to the end of the class please :)

packages/workspaces-extension/src/index.ts
jtpio: > Maybe we should add the it as opt-in for 4.4 and decide on whether to make it opt-out for next release later?

Sounds good.

It's often difficult to decide what should be enabled by default when we lack metrics and usage data. My personal feeling was that workspaces are used by a handful of users only, but this may be a wrong statement.

Darshan808:
Indicator is now opt-in (default: false) and will display whenever toggled on.

packages/workspaces-extension/style/selector.css
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.jp-WorkspaceSelector-item:hover {
+  background: var(--jp-layout-color2);
+}
+
+.jp-mod-active {
+  background: var(--jp-brand-color2);
+  color: var(--jp-ui-inverse-font-color1);
+}
```
krassowski, line 84: This needs to be more specific as it affects notebook:

Screenshot from 2025-02-05 21-55-48

Darshan808:
The test is failing due to a snapshot mismatch after adding a new item under View > Appearance.

Darshan808:
So many snapshot mismatches after updating the snapshots! 🤔

galata/test/documentation/general.test.ts-snapshots/file-formats-open-with-documentation-linux.png
krassowski, line 1: This one looks like an incorrect update, we should revert it

Darshan808:
bot please update snapshots

Darshan808:
bot please update snapshots

Darshan808:
The tests are consistently failing due to snapshot mismatches (different each time), despite multiple attempts to resolve it. I don't see anything in this PR that could be causing the issue. Should we just give it another try ?

Darshan808:
bot please update snapshots

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +50,76 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+      identifiers: model.identifiers,
+      openWorkspace: openWorkspace,
+      model: model,
+      translator: translator
+    });
+    const isToggled = await registry.get(
+      WORKSPACE_INDICATOR_PLUGIN_ID,
+      'toggled'
+    );
+    workspaceSelector.setHidden(!isToggled.composite as boolean);
```
krassowski, line 96: This will not update the state if user changes the setting, right?

It should be something as in:


https://github.com/jupyterlab/jupyterlab/blob/5138e10ce54f136b92fb36b9f863653b37ebdb83/packages/shortcuts-extension/src/index.ts#L319-L325

or 

https://github.com/jupyterlab/jupyterlab/blob/5138e10ce54f136b92fb36b9f863653b37ebdb83/packages/apputils-extension/src/palette.ts#L104-L131

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +50,76 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+    app.shell.add(workspaceSelector, 'top', { rank: 1000 });
+    app.commands.addCommand(WORKSPACE_INDICATOR_COMMAND_ID, {
+      label: trans.__('Show Workspace Indicator'),
+      isToggled: () => workspaceSelector.isVisible,
+      execute: () => {
+        workspaceSelector.setHidden(workspaceSelector.isVisible);
+        if (registry) {
```
krassowski, line 103: `registry` is always defined as per requirements, so this if is not needed, do I see it right?

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +50,76 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+      WORKSPACE_INDICATOR_PLUGIN_ID,
+      'toggled'
+    );
+    workspaceSelector.setHidden(!isToggled.composite as boolean);
+    app.shell.add(workspaceSelector, 'top', { rank: 1000 });
+    app.commands.addCommand(WORKSPACE_INDICATOR_COMMAND_ID, {
+      label: trans.__('Show Workspace Indicator'),
+      isToggled: () => workspaceSelector.isVisible,
+      execute: () => {
+        workspaceSelector.setHidden(workspaceSelector.isVisible);
```
krassowski, line 102: Once you add `settings.changed` signal listener this would not be needed

galata/test/jupyterlab/workspace.test.ts
```diff
@@ -228,6 +228,28 @@ test.describe('Workspace', () => {
     await expect(page.locator(`[role="main"] >> text=${mdFile}`)).toBeVisible();
     await expect(page.locator(`[role="main"] >> text=${nbFile}`)).toBeVisible();
   });
+
+  test('should display workspace indicator and switch workspaces', async ({
+    page
+  }) => {
+    await page.goto();
+    const workspaceSelector = page.locator('.jp-WorkspaceSelector-header');
+    //Not visible by default
```
krassowski, line 237: It's a bit easier to read with a space and the rest of codebase uses such style:

```suggestion
    // Not visible by default
```

galata/test/jupyterlab/workspace.test.ts
```diff
@@ -228,6 +228,28 @@ test.describe('Workspace', () => {
...
+    expect(await workspaceSelector.isVisible()).toBe(false);
+
+    //Open custom workspace
```
krassowski, line 240:
```suggestion
    // Open custom workspace
```

Darshan808:
bot please update snapshots

krassowski:
It looks like a few more snapshots need reverting as the tests are failing on:

```
    [documentation] › test/documentation/general.test.ts:615:7 › General › Command Palette ─────────
    [documentation] › test/documentation/internationalization.test.ts:14:7 › Internationalization › Menu 
    [documentation] › test/documentation/internationalization.test.ts:28:7 › Internationalization › Confirm language 
```

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +50,84 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+      await app.commands.execute(commands.open, { workspace: workspace });
+    };
+    const workspaceSelector = new WorkspaceSelectorWidget({
+      currentWorkspace: resolver.name,
+      identifiers: model.identifiers,
+      openWorkspace: openWorkspace,
+      model: model,
+      translator: translator
+    });
+    app.shell.add(workspaceSelector, 'top', { rank: 1000 });
```
jtpio, line 90: Wondering if this workspace selector item should be specified as a `Topbar` toolbar item?

https://github.com/jupyterlab/jupyterlab/blob/1f1a645d840d28ab7a85b751cf1df7c370a00254/packages/application-extension/schema/top-bar.json#L4-L12

This could let users specify where they would like to the selector to be displayed by changing the `rank`?

krassowski:
The CI is failing because jupyterlab-demo repository was updated in https://github.com/jupyterlab/jupyterlab-demo/pull/148. Restarting CI will not fix it - we need to update the two affected documentation snapshots, ideally in a new PR. FYI, you can restart individual jobs too.

Darshan808:
Got it.
But would the new PR be created from a branch without any changes, and then we trigger the GitHub Actions bot to update the snapshots in that PR?

krassowski:
That's one option. Another is to just download the artifacts, replace them locally, compress and open a PR with all the changes. I can take care of it this time as I already downloaded the artifacts to take a peek.

Darshan808:
> FYI, you can restart individual jobs too.
This would be super helpful to me, but I'm not sure how to do it.

krassowski:
I see options to do that when hovering over individual jobs:

image

And there is "Rerun failed"

image

from https://github.com/jupyterlab/jupyterlab/actions/runs/13283748228/job/37087680489?pr=17256

Do you see these too or is this limited by permissions?

Darshan808:
I don't see any of those. Looks like it's related to permissions. Only way to re-trigger tests for me is to close and reopen PR.

krassowski:
> The tab title now includes the current workspace as a suffix, making it easier to identify the workspace context.

It works when a notebook/file is active:

image

It does not work in the launcher. I think it could say "{workspace} - JupyterLab" in that case or "JupyterLab - {workspace}"?

image

It has "{workspace} - {workspace}" when Settings Editor, Licences or another non-document tab is active:

| Settings | Licences | Contextual |
|--|--|--|
| image | image | image |

packages/workspaces-extension/src/top_indicator.tsx
```diff
@@ -0,0 +1,156 @@
...
+        aria-expanded={isExpanded}
+        aria-controls={dropdownId}
+      >
+        <span className="jp-WorkspaceSelector-current">
+          {currentWorkspace.length > 12
+            ? `${currentWorkspace.slice(0, 12)}...`
+            : currentWorkspace}
+        </span>
+        <span className="jp-WorkspaceSelector-caret">
+          {isExpanded ? '▲' : '▼'}
```
krassowski, line 72: Can we use the SVG carets (I would suggest `caretUpIcon` and `caretDownIcon` but we also have `caretDownEmptyIcon` and `caretDownEmptyThinIcon` variants) for:
- visual consistency across browsers?
- guarantee that it will render even if system font does not support ▲ and ▼ characters?

packages/workspaces-extension/src/index.ts
```diff
@@ -46,13 +50,89 @@ const workspacesMenu: JupyterFrontEndPlugin<void> = {
...
+      openWorkspace: openWorkspace,
+      model: model,
+      translator: translator
+    });
+
+    toolbarRegistry.addFactory('TopBar', 'workspaceIndicator', () => {
+      return workspaceSelector;
+    });
+
+    const loadSettings = registry.load(WORKSPACE_INDICATOR_PLUGIN_ID);
```
krassowski, line 97: I feel it does not look great when we keep adding new setting sections which have only one setting:

image

One solution would be to rename the section to "Workspaces" (rather than "Workspace Indicator") and have it under `plugin.json`. Then we could add more workspace settings in this section in the future. The downside is that if someone disables the `@jupyterlab/workspaces-extension:indicator`, then the setting would be still shown but meaningless.

Another solution would be to remove this setting and make this plugin programatically modify `TopBar` settings programatically. I think this is the best approach. @jtpio I know you worked on a few `TopBar` plugins before, is there a reason not to do it this way?

image

jtpio: Or maybe the toolbar entry for the workspace indicator should be set `"disabled": true` by default? So there wouldn't be a need for a dedicated setting?

PR #17258, opened 2025-02-04
Use the same `isPalette` arg as for other commands

jtpio:
## References

Minor follow-up to https://github.com/jupyterlab/jupyterlab/pull/17253

Other commands use the `isPalette` arg, for example here:

https://github.com/jupyterlab/jupyterlab/blob/68075b17a11258481f5aba68f473788aeaa88650/packages/console-extension/src/index.ts#L685

## Code changes

Make the use of commands in the palette consistent with other commands.

## User-facing changes

None

None


krassowski (approved):
Good catch, thank you!

PR #17261, opened 2025-02-05
Reuse DOM and keep selection on streaming stdout/stderr

krassowski:
## References

This is a subset of changes extracted from https://github.com/jupyterlab/jupyterlab/pull/17197.

### DIFF
--- a/packages/rendermime/src/renderers.ts
+++ b/packages/rendermime/src/renderers.ts
@@ -821,8 +821,8 @@ function renderTextual(
   // Unpack the options.
   const { host, sanitizer, source } = options;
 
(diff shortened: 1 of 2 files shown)


krassowski:
Both in local testing and on CI this change does not improve but deteriorate performance for the case of streaming line-by-line. While these changes are essential to make https://github.com/jupyterlab/jupyterlab/pull/17197 work, it appears that they are not worth it without managing the rendering process with animation frames.

krassowski:
Closing for now, we may revisit this later depending on the discussion around https://github.com/jupyterlab/jupyterlab/pull/17197

Issue #14508, opened 2023-05-08
Extension manager in 4.0: missing `approved` extensions implementation


krassowski:
## Description

In JupyterLab 3 and earlier the extension manager had an indicator of extensions from official Project Jupyter organisations, e.g.:

<img src="https://user-images.githubusercontent.com/5832902/236788160-daf2dc82-0fae-4742-8654-13d743cfe125.png" width=300>

This was protecting users from typo-squatting when they were installing packages from the extension manager. When we migrated from npm backend to PyPI this indicator became dysfunctional; while there is a metadata field `approved`
https://github.com/jupyterlab/jupyterlab/blob/022ef77d799ac5c3d254c18c67537b29c812aa99/jupyterlab/extensions/manager.py#L63
https://github.com/jupyterlab/jupyterlab/blob/022ef77d799ac5c3d254c18c67537b29c812aa99/jupyterlab/extensions/manager.py#L85

And there is code on the frontend side to show the logo:

https://github.com/jupyterlab/jupyterlab/blob/022ef77d799ac5c3d254c18c67537b29c812aa99/packages/extensionmanager/src/widget.tsx#L108-L118

the problem is that the `approved` status is never set by the default PyPI manager.

The reason for it was at least in part technical: when the Extension Manager was refactored PyPI did not have a concept of an organization (while npm did). This April PyPI announced that organizations are now a thing [Introducing PyPI Organizations](https://blog.pypi.org/posts/2023-04-23-introducing-pypi-organizations/). Jupyter secured an organization named jupyter: https://pypi.org/org/jupyter/, thanks @blink1073 (https://github.com/jupyter/security/issues/61)

Moreover, I find the current implementation a bit confusing: it shows "This extension is approved by your security team" and a Jupyter logo. I think the better way forward would be to have:
- a list of PyPI organization (and users) that are "approved" defined on the backend and adminstrator-customizable
- an icon depending on organization (e.g. the Jupyter icon for hard-coded "jupyter" organization, and factory icon for any other organization)

This would be similar to existing `allowed_extensions_uris` option:

https://github.com/jupyterlab/jupyterlab/blob/022ef77d799ac5c3d254c18c67537b29c812aa99/jupyterlab/extensions/manager.py#L124

but rather than being on per-extension basis, it would be on per-organization basis.

This is an API change which may be harder to implement later.

fcollonval:
I'm afraid that using multiple icons will raise confusion.

[...]

PR #17266, opened 2025-02-05
Fix order of extensions in PyPI Extension Manager

### LINKED ISSUE #17235: PyPI Extension Manager should not show results alphabetically in Discovery
## Description

The discovery section of extensions should not be used for political statements by extension authors gaming the system:

Image

Not that the white-blue-red flag is deliberate as it comes from `https://github.com/RussiaZ` account - see https://en.wikipedia.org/wiki/Z_(military_symbol). The one with blue-yellow flag at least is a real extension.

Even worse, someone could use it to show more offensive content.

## Reproduce

Open extension manager

## Expected behavior

- Ideally we would have some popularity or quality sort, but we don't (there was one on npm but not on PyPI)
- Instead let's show extensions from Project Jupyter PyPI org first, then from jupyterlab-contrib and then random.

## Context

- JupyterLab version: 4.x

Princekumarofficial:
## References

Closes #17235
This PR improves the sorting logic for PyPI extensions in the Extension Manager, ensuring that:

Project Jupyter extensions (from github.com/jupyter/ or jupyter.org) appear first.
JupyterLab-Contrib extensions (from github.com/jupyterlab-contrib/) appear next.
All other extensions appear afterward in a default priority order.

## Code changes

Made changes in jupyterlab/extensions/pypi.py

            priority = 3  # Default priority for other packages
            urls_to_check = [
                str(source_url).lower() if source_url else "",
                str(homepage_url).lower() if homepage_url else "",
                str(best_guess_home_url).lower() if best_guess_home_url else "",
            ]

            for url in urls_to_check:
                if any(org in url for org in ["github.com/jupyter/", "jupyter.org"]):
                    priority = 1
                    break
                elif "github.com/jupyterlab-contrib/" in url:
                    priority = 2
                    break

            all_matches.append((priority, extension))

        sorted_matches = sorted(all_matches, key=lambda x: (x[0], x[1].name))

        # Apply pagination
        start_idx = (page - 1) * per_page
        end_idx = start_idx + per_page
        page_matches = sorted_matches[start_idx:end_idx]

        for _, extension in page_matches:
            extensions[extension.name] = extension

        total_pages = math.ceil(len(sorted_matches) / per_page)

        return extensions, total_pages
        
And in packages/extensionmanager/src/model.ts file to show models in the order Project Jupyter PyPI org first, then from jupyterlab-contrib and then random.

## User-facing changes

- Improved Extension Discovery: Users will now see official and community-backed extensions at the top.
- Better Organization: Makes it easier to find high-quality, trusted extensions.

image

None. This change only affects the sorting order and does not modify existing API behavior.


jupyterlab/extensions/pypi.py
```diff
@@ -319,7 +313,37 @@ async def list_packages(
...
             )
 
-        return extensions, math.ceil((counter + 1) / per_page)
+            # Determine organization priority
+            priority = 3  # Default priority for other packages
+            urls_to_check = [
+                str(source_url).lower() if source_url else "",
+                str(homepage_url).lower() if homepage_url else "",
+                str(best_guess_home_url).lower() if best_guess_home_url else "",
+            ]
```
krassowski, line 322: The problem here is that anyone can push an extension to PyPI and put such URL - they do not verify these details be default.

jupyterlab/extensions/pypi.py
```diff
@@ -319,7 +313,37 @@ async def list_packages(
...
+
+            for url in urls_to_check:
+                if any(org in url for org in ["github.com/jupyter/", "jupyter.org"]):
```
krassowski, line 325: `github.com/jupyterlab/` should have the same priority as `github.com/jupyter/`

jupyterlab/extensions/pypi.py
krassowski: I think we can still move forward with this PR as this will be an improvement over the current status quo. We will need to make it more robust soon after though.

Darshan808:
Welcome @Princekumarofficial 
Thank you for making a Pull Request to JupyterLab. Great work! The CI is currently failing due to a snapshot mismatch caused by some of the changes you've made. I'll take care of fixing it later. In the meantime, please review the feedback above.

Darshan808:
To test this, we could add a few more extensions with different URLs and assert whether they render in the correct order. The test can be added to: https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/documentation/extension_manager.test.ts

jupyterlab/extensions/pypi.py
Princekumarofficial: @krassowski Thanks for the review! I'd be happy to work on making it more robust in a follow-up PR

Princekumarofficial:
@Darshan808 thanks for assisting, i have updated the tests

Princekumarofficial:
@krassowski , I have made the changes

Darshan808:
bot please update snapshots

Darshan808:
> To test this, we could add a few more extensions with different URLs and assert whether they render in the correct order. The test can be added to: https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/documentation/extension_manager.test.ts

By this I mean to add a test that verifies whether extensions list is sorted as expected in UI or not.

<details>
<summary>Here is a test you can add:</summary>

```
  test('Custom sorted extension list', async ({ page }) => {
    await page.goto();
    await openExtensionSidebar(page);
    
    // Get all homepage URLs of installed extensions
    const homepageLinks = await page
      .locator('.jp-extensionmanager-installedlist .jp-extensionmanager-entry-name a')
      .evaluateAll((anchors: HTMLAnchorElement[]) => 
        anchors.map(a => a.href)
      );

      const desiredOrder = ['jupyterlab', 'jupyter', 'jupyter-contrib'];

      function getPriority(url: string) {
        // extract the organization 
        const regex = /^https:\/\/github\.com\/([^\/]+)\//;
        const match = url.match(regex);
        if (match) {
          const org = match[1];
          const index = desiredOrder.indexOf(org);
          // return priority
          return index !== -1 ? index : desiredOrder.length;
        }
        // If URL doesn't match the regex, assign it the default rank
        return desiredOrder.length;
      }
      
      // Map each URL to its priority
      const priorities = homepageLinks.map(url => getPriority(url));
      const sortedPriorities = [...priorities].sort((a, b) => a - b);

      // Priorities should be sorted
      expect(priorities).toEqual(sortedPriorities)
  });
});
```
</details>

krassowski:
> We should also add a way to move the archived package at the end of the list.

:+1: (or filter out?) This comes down to whatever the API will be, currently we cannot get the "archived" status via JSON API yet but this is actively being worked on (https://github.com/pypi/warehouse/issues/16844#issuecomment-2599288858)

dlqqq:
@krassowski Thanks for bringing this to my attention. We do have an existing issue for this, but I haven't been able to work on it. I don't know how to remove a package from PyPI.

If you're interested, I can assign this issue to you and add your PyPI account to Jupyter AI packages. I've moved this issue to the "v3.0.0" milestone for now so it doesn't get forgotten.

krassowski:
>  I don't know how to remove a package from PyPI.

I think archival would be sufficient for now, it is generally recommended over removal - see https://blog.pypi.org/posts/2025-01-30-archival/

Princekumarofficial:
@krassowski i have down-ranked extensions with URL matching either https://github.com/jupyterlab/extension-examples or https://github.com/jupyterlab/jupyterlab_apod
will also implement attestation checks in an upcoming PR

galata/test/documentation/extension_manager.test.ts-snapshots/extensions-allowed-list-documentation-linux.png
krassowski, line 1: These snapshots are wrong but it is a pre-existing issue. I opened https://github.com/jupyterlab/jupyterlab/issues/17310.

jupyterlab/extensions/pypi.py
```diff
@@ -319,7 +313,47 @@ async def list_packages(
...
             )
 
-        return extensions, math.ceil((counter + 1) / per_page)
+            # Determine organization priority
+            priority = 3  # Default priority for other packages
+            urls_to_check = [
+                str(source_url).lower() if source_url else "",
+                str(homepage_url).lower() if homepage_url else "",
+                str(best_guess_home_url).lower() if best_guess_home_url else "",
+            ]
```
krassowski, line 322: This would seem slightly easier to read for me and avoid putting empty strings in the URLs list

```suggestion
            urls_to_check = [
                str(source_url).lower()
                for url in [source_url, homepage_url, best_guess_home_url]
                if url
            ]
```

packages/extensionmanager/src/model.ts
```diff
@@ -652,11 +652,49 @@ namespace Private {
    * A comparator function that sorts allowedExtensions orgs to the top.
    */
   export function comparator(a: IEntry, b: IEntry): number {
-    if (a.name === b.name) {
-      return 0;
-    } else {
-      return a.name > b.name ? 1 : -1;
+    // Define priority organizations
+    const priorityOrgs = ['jupyterlab', 'jupyter', 'jupyterlab-contrib'];
```
krassowski, line 656: This only matters for installed extensions, because the "Discover" section would be sorted by pagination, right? Otherwise we could just remove the call to `Private.comparator` in `search()` method?

I think I would lean towards reducing the duplication of the logic and removing sorting from the frontend for results from pagination. In case of search the results should be sorted by similarity score either way. In case of installed extensions, maybe these should be sorted but not necessarily by org.

krassowski: Also, if we remove the changes on frontend it can be fully tested using `pytest`

krassowski (approved):
Thank you @Princekumarofficial!

packages/extensionmanager/src/model.ts
Princekumarofficial: It is there because we have to show the official extensions first if anyone searches for any query, and this is also useful if there are extensions from both official and unofficial orgs, by default it is lexicographical order. what are your thoughts, should we remove it @krassowski ?

fcollonval:
> can you fix the URL for https://pypi.org/project/jupyterlab-rtc/ (not sure what it is) as it now shows up and points to https://github.com/jupyterlab/jupyterlab-rtc although that page does not exist and the PyPI also points to datalayer org repo which is not public so I cannot open an issue over there.

@krassowski actually I don't know what to do about this. jupyterlab-rtc was taken by @echarles; I don't see the project in datalayer org any longer.
The fact that Carlos and I are also listed as maintainer is due to the work on jupyter-collaboration that we [first named jupyterlab-rtc](https://github.com/jupyterlab/jupyter-collaboration/tree/df7a73c2bb1da1f2b239dc5c818788d62b9b284b) asking Eric to transfer us the right. But that triggered a naming discussion and we never published the code under `jupyterlab-rtc`.

I don't really see how to fix that except by adding a filter to check the link exists as anybody can still trick the mechanism (or simply to avoid highlighting dead project) - that fact revives that the only reliable solution for this is to have a proxy server that would do the filtering/data expanding (e.g. adding download count) from pypi. I tested that some year ago but never finished it. But it should not be too complex to do using github actions and publishing a JSON using github pages to be consumed by JupyterLab.

krassowski:
> except by adding a filter to check the link exists as anybody can still trick the mechanism

Well, no because now with trusted publishers and attestations we can check the provenance via API - explored some of it in https://github.com/jupyterlab/jupyterlab/issues/17235 and tracking in https://github.com/jupyterlab/jupyterlab/issues/14508 now (this is only a recent addition to PyPI REST API).

Also, there is going to be a way to check if a project is archived soon, so if @echarles could archive it that would probably make sense (and if it is ever needed it could be un-archived).

> the only reliable solution for this is to have a proxy server that would do the filtering/data expanding (e.g. adding download count) from pypi

Right, we discussed it before and it is a viable solution.

Issue #665, opened 2016-08-15
Provide nightly builds


akosyakov:
It would be nice to have ability to build against the latest master without having a local copy.

blink1073:
I'd suggest we *not* use Travis CI to do this:

cf https://github.com/jupyter/jupyterlab/issues/407

jtpio:
@akosyakov dist files are now generated on CI and available as artifacts: 

![image](https://user-images.githubusercontent.com/591645/146038765-88e4966a-7385-4bac-8eff-637e1959b44f.png)

These can be downloaded locally and installed with `pip install /path/to/the/wheel`.

Maybe that's enough to cover the nightly build case? Maybe it would also be possible to query these artifacts from the latest commit on `master` via the GitHub API?

akosyakov:
Honestly i don't remember the context of this issue from 5 years ago :)

PR #17278, opened 2025-02-10
Add item about nightly releases to the FAQ

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/665

Since releases still take some time, even though they are automated, we probably don't want to start publishing nightly releases as this may likely increase the maintenance burden.

Instead we can simply document how folks can grab and install artifacts built by CI.

## Code changes

Documentation changes.

## User-facing changes

For those interested, provide a way to install unreleased versions of JupyterLab.

None

### DIFF
--- a/docs/source/getting_started/faq.rst
+++ b/docs/source/getting_started/faq.rst
@@ -79,3 +79,20 @@ Development
    For more discussion and potential alternative solutions, please see issues
    `#4623 <https://github.com/jupyterlab/jupyterlab/issues/4623>`__ and
    `#5789 <https://github.com/jupyterlab/jupyterlab/issues/5789>`__.
+
+
+Nightly releases
+----------------
+
+The JupyterLab project does not publish nightly releases to PyPI.
+
+However JupyterLab is built on CI for every commit on the ``main`` branch, and generates the wheel and source distributions as GitHub Action artifacts.
+These artifacts can be downloaded and installed locally.
+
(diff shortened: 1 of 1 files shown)


docs/source/getting_started/faq.rst
```diff
@@ -79,3 +79,24 @@ Development
...
+However JupyterLab is built on CI for every commit on the ``main`` branch, and generates the wheel and source distributions as GitHub Action artifacts.
+These artifacts can be downloaded and installed locally.
+
+To download the JupyterLab wheels from the latest commits on ``main``:
+
+- Go to `Check Release GitHub Action page <https://github.com/jupyterlab/jupyterlab/actions/workflows/check-release.yml?query=branch%3Amain+is%3Asuccess>`__
+- Click on one of the workflow runs
+- Under the "Artifacts" section, click on ``jupyterlab-releaser-dist-<build-number>`` to download the archive
+- Locally, extract the archive
+- Install with ``python -m ./jupyterlab-x.y.z.whl``
```
krassowski, line 98: Should it be

```suggestion
- Install with ``python -m pip install ./jupyterlab-x.y.z.whl``
```

jtpio: Yes good catch, thanks!

krassowski (approved):
Thank you!

Issue #2047, opened 2017-04-17
Long hold on buttons causes Character Accents Popup instead of Key Repeat inside terminal, editor


stephenholtz:
I'm not sure if this is a bug or an an enhancement request, but all of the various windows within JupyterLab on OSX inherit the browser's `ApplePressAndHoldEnabled` value so that holding down keys causes a popup instead of repeats of that character. This results in a real finger workout particularly when in vim mode in the editor, or using the terminal (IMO). For example when in vim mode in the editor, holding down any of the navigation keys `j k l m` does not cause repeated movement as it normally does e.g. in a desktop terminal application. It can be changed globally by running a command in the terminal:

`defaults write -g ApplePressAndHoldEnabled -bool false` 
then logging out and back in 

Maybe this should be entered in a troubleshooting section, or someone who knows more about setting these preferences on a per-app basis.

jasongrout:
This is a general OS X issue, right? IIRC, I had to set this in order to get vi or even just generic key repeat to work in the OS X terminal as well.

+1 to adding it to a FAQ or troubleshooting section somewhere in the docs.

stephenholtz:
It does seem to be a general issue. However in both the OSX terminal and iTerm this setting is application specific, so conceivably JupyterLab could override the setting if people wanted that.

If you point me to the place to add this to the documentation I will do so.

jasongrout:
> both the OSX terminal and iTerm this setting is application specific

Ah, I didn't realize it could be toggled per app (helpful link showing syntax: https://gist.github.com/lsd/1e1826907ab7e49c536a).

> so conceivably JupyterLab could override the setting if people wanted that.

Since the JupyterLab "application" is really using the web browser, it seems that someone would have to toggle it for the browser they are using.

> If you point me to the place to add this to the documentation I will do so.

The docs are very incomplete at this stage. We don't have a good place for it right now, I think. Can we leave this issue open, tagged with the docs label, to add it when we organize the user-level documentation for JupyterLab?

It would be really helpful if you could phrase a paragraph or so of markdown text we could copy and paste into those docs.

stephenholtz:
## OS X Specific Issues:
### Holding down button does not produce repeated key press events.

Recent version of OS X change the default behavior for holding down buttons: instead of giving a repeated key press, a Character Accents Popup occurs. For example, when in vim mode in the editor, holding down any of the navigation keys `h j k l` does not cause repeated movement as it normally does in a desktop terminal application.
[...]

PR #17279, opened 2025-02-10
Add FAQ item about repeated key press events on OS X

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/2047

Based on this comment: https://github.com/jupyterlab/jupyterlab/issues/2047#issuecomment-294586270

## Code changes

Documentation changes.

## User-facing changes

This makes it easier to JupyterLab to find which command to run to disable the "Character Accents Popup" on Macos.

None


docs/source/getting_started/faq.rst
```diff
@@ -16,6 +16,27 @@ General
    JupyterLab? <https://jupyterlab.readthedocs.io>`__
 -  :ref:`How can you report a bug or issue? <issue>`
 
+OS X Specific Issues
+--------------------
+
+Holding down buttons does not produce repeated key press events
```
krassowski, line 22: Given that this was raised 8 years ago and the issue had no new comments since, do we know if this still holds today? Regardless, maybe this should be lower down in FAQ if there was not that many questions about it?

jtpio: Yes it's still the case. Maybe there are not that many issues because it is mostly noticeable when using the vim key bindings to move the cursor with `h j k l`. For example it's documented here in the VSCode vim plugin: https://github.com/VSCodeVim/Vim?tab=readme-ov-file#mac

jtpio: Otherwise yes we can likely move it down in the FAQ.

krassowski: Ok, let's push it down and get it in :)

krassowski (approved):
Thank you @jtpio!

PR #16872, opened 2024-10-16
Fix job updating snapshots when triggered by comment on first-time contributor PR

krassowski:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16730

This should help with updating snapshots in https://github.com/jupyterlab/jupyterlab/pull/16866

## Code changes

Allows the galata snapshot update job to run if the comment author is owner/member/collaborator (in addition to allowing to update snapshots if the PR author is owner/member/collaborator).

## User-facing changes

None

## Backwards-incompatible changes

None

PR #17289, opened 2025-02-13
Fix missing checks for author comment association for docs

krassowski:
## References

- Noticed in https://github.com/jupyterlab/jupyterlab/pull/17244#issuecomment-2656172713
- Follow-up to https://github.com/jupyterlab/jupyterlab/pull/16872

## Code changes

Fixes unitended difference in conditions between:

https://github.com/jupyterlab/jupyterlab/blob/362e9570ecc44a02ff0b3d75f371610438b3052b/.github/workflows/galata-update.yml#L12-L25

and

https://github.com/jupyterlab/jupyterlab/blob/362e9570ecc44a02ff0b3d75f371610438b3052b/.github/workflows/galata-update.yml#L103-L113

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/galata-update.yml
+++ b/.github/workflows/galata-update.yml
@@ -106,7 +106,10 @@ jobs:
       (
         github.event.issue.author_association == 'OWNER' ||
         github.event.issue.author_association == 'COLLABORATOR' ||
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

Issue #17080, opened 2024-12-12
Seeking feedback: new shortcuts added in editor in JupyterLab 4.4.0


krassowski:
In JupyterLab 4.4.0a1 we are upgrading CodeMirror 6 packages (https://github.com/jupyterlab/jupyterlab/pull/17064) which adds a few shortcuts to the editor.

This issue is to gather feedback from community if we should keep the two new shortcuts inherited from CodeMirror or remove them. These are:
-  Ctrl-m (Shift-Alt-m on macOS) which toggles the behaviour of the <kbd>Tab</kbd> key to allow it to move focus; previously:
    - in notebook cells <kbd>Tab</kbd> in edit mode was always indenting (inserting four spaces in Python) or invoking the completer; Jupyter users could press <kbd>Esc</kbd> to enter command mode in which Tab <kbd>Tab</kbd> was moving focus to next widget
     - in file editor there was no  <kbd>Esc</kbd> equivalent so I there might be seen as a plausible accessibility improvement if documented.
- Alt-Backtick as additional binding on macOS, where IME can take over Ctrl-Space

Please let us know if these shortcuts introduce unintended behaviour, or conflict with typing on any international keyboard, or with shortcuts introduced by any popular JupyterLab or browser extension.

hxrshxz:
Maybe we can have a Toggle Option, a configuration option (maybe in the settings) to allow users to toggle these new shortcut behaviors, so that those who prefer the old behavior or face conflicts with their specific setups can go back to the previous one.also i think many users may have become used to the behavior where Tab in edit mode triggers the code completer. Changing this behavior could cause confusion, especially if users are unaware of this new shortcut. Documentation for this will be necessary i believe

krassowski:
> also i think many users may have become used to the behavior where Tab in edit mode triggers the code completer. Changing this behavior could cause confusion, especially if users are unaware of this new shortcut

I mean Tab will still trigger the code completer and indentation, unless user presses Ctrl-m or Shift-Alt-m before pressing Tab. But yes, it will indeed be confusing if they press it accidentally as there is no visual indicator that the Tab mode has switched and they will most likely think that the Tab functionality is broken.

Strictly, it seems that the completer would still work but instead of adding indentation on empty lines the focus would move. Maybe this is an argument for disabling this shortcut as the behaviour is quite hard to understand for end user.

PR #17291, opened 2025-02-13
Disable new `ctrl` + `m` toggle focus binding, enable configuring it via Keyboard Shortcuts

krassowski:
## References

Addresses first point in #17080. I found this Ctrl + M can lead to user confusion as this is not documented anywhere and has no visual indicator; in notebook switching between edit/command mode has equivalent effect. Instead, this PR allows user to opt-in and choose a suitable key-binding by binding a new `codemirror:toggle-tab-focus-mode` command.

## Code changes

Adds new `codemirror:toggle-tab-focus-mode` command.

## User-facing changes

- Ctrl + M no longer switches tab focus mode in CodeMirror (this was never released, it was added in 4.4.0a as inherited change from upstream CodeMirror)
- Users are able to set a preferred shortcut for toggling tab focus mode (no shortcut is set by default yet as we do not know what would be a good one)

None

### DIFF
--- a/packages/codemirror-extension/src/commands.ts
+++ b/packages/codemirror-extension/src/commands.ts
@@ -108,7 +108,7 @@ export const commandsPlugin: JupyterFrontEndPlugin<void> = {
     app.commands.addCommand(CommandIDs.toggleTabFocusMode, {
       label: trans.__('Toggle Tab Focus Mode'),
       caption: trans.__(
-        'Toggles behaviour of Tab key between inserting indentation and moving to next focusable element'
+        'Toggles behavior of Tab key between inserting indentation and moving to next focusable element'
       ),
       execute: () => {
         const view = findEditorView();
--- a/packages/codemirror-extension/src/commands.ts
+++ b/packages/codemirror-extension/src/commands.ts
@@ -6,7 +6,8 @@
 import {
   deleteLine,
   toggleBlockComment,
-  toggleComment
+  toggleComment,
+  toggleTabFocusMode
 } from '@codemirror/commands';
 import { EditorView } from '@codemirror/view';
 import { selectNextOccurrence } from '@codemirror/search';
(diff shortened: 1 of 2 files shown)


packages/codemirror-extension/src/commands.ts
```diff
@@ -103,6 +105,21 @@ export const commandsPlugin: JupyterFrontEndPlugin<void> = {
       isEnabled
     });
 
+    app.commands.addCommand(CommandIDs.toggleTabFocusMode, {
+      label: trans.__('Toggle Tab Focus Mode'),
+      caption: trans.__(
+        'Toggles behaviour of Tab key between inserting indentation and moving to next focusable element'
```
jtpio, line 111: Looks like most (but not all) of the code base follows the other spelling, so maybe better to keep it consistent (also w.r.t localization)?

```suggestion
        'Toggles behavior of Tab key between inserting indentation and moving to next focusable element'
```

packages/codemirror/src/extension.ts
```diff
@@ -734,7 +734,13 @@ export namespace EditorExtensionRegistry {
             //   want to run a cell action (switch to command mode) on Esc
             // - Disable default Enter handler because it prevents us from
             //   accepting a completer suggestion with Enter.
+            // - Disable Ctrl-m (Shift-Alt-m on Mac) which toggles tab focus mode;
+            //   JupyterLab binds `Esc` to an equivalent behaviour (switching
```
jtpio, line 738: Similar as above:

```suggestion
            //   JupyterLab binds `Esc` to an equivalent behavior (switching
```

jtpio (approved):
Thanks!

Testing on MacOS and it seems to be behaving as expected 👍 


https://github.com/user-attachments/assets/04a7134b-d1a1-44ca-b3eb-1a702feb5725

Issue #17293, opened 2025-02-14
Remove remaining `jupyter-packaging` references


krassowski:
> When searching the code base of `jupyter-packaging`, there seems to be a few remaining references:
> 
> https://github.com/jupyterlab/jupyterlab/blob/b93e83488cb610d0803ade14f2cf091369100a90/jupyterlab/tests/test_extensions.py#L243
> 
> https://github.com/jupyterlab/jupyterlab/blob/b93e83488cb610d0803ade14f2cf091369100a90/jupyterlab/tests/test_extensions.py#L251
> 
> https://github.com/jupyterlab/jupyterlab/blob/b93e83488cb610d0803ade14f2cf091369100a90/scripts/release_prep.sh#L17
> 
> 
> _Originally posted by @jtpio in https://github.com/jupyterlab/jupyterlab/pull/17219#pullrequestreview-2617729909_
            
Follow-up to #17219

PR #17219, opened 2025-01-25
Updated mock packages to use hatch-jupyter-builder

Rishab87:
## References
Fixes #17171 

## Code changes
Migrated mock packages which were using jupyter_packaging to use hatch_jupyter_builder

PR #17294, opened 2025-02-14
Clean up references to `jupyter-packaging`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/17293

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17219

## Code changes

These requirements on `jupyter-packaging` should not be needed anymore.

## User-facing changes

None

None


scripts/release_prep.sh
jtpio, line 1: This `release_prep.sh` file is only referenced in the "Manual Release" steps in `RELEASE.md`.

Even though we haven't done a manual release for a long time, maybe it doesn't hurt to keep it around for now.

krassowski (approved):
Thank you @jtpio!

Issue #16996, opened 2024-11-26
Add run-all-below button to Notebook Panel, becomes gray when starting a new notebook


firezym:
## Description

Add the following settings to Notebook Panel as below:
Settings Editor -> Notebook Panel -> Json Settings Editor
```json
{
    "toolbar": [
        {
            "name": "restart-and-run",
            "disabled": true
        },
        {
            "name": "run-all-below",
            "args": {},
            "command": "notebook:run-all-below",
            "icon": "ui-components:fast-forward",
            "disabled": false,
            "rank": 36
        }
    ]
}
```

## Reproduce and Expected behavior

![image](https://github.com/user-attachments/assets/6d5ab54f-166a-4917-8ffb-c3974871f56e)
Refresh the whole page with F5 before testing the following 3 scenarios:
1. open a new notebook from File Browser, most of the time, the newly added `notebook:run-all-below` icon is gray and not valid to click
2. if u switch to another notebook and switch back, the icon will become valid and clickable
3. if u use web browser's F12 dev mode, most of the time, even if u open new notebook from File Browser, the icon is valid and clickable

## Context

- Operating System and version: Windows 11 Pro, 23H2, 22631.4460
- Browser and version: Brave [Version 1.73.91 Chromium: 131.0.6778.85 (Official Build) (64-bit)]
- JupyterLab version: 
jupyter_client             8.6.3
jupyter_core               5.7.2
jupyter-events             0.10.0
jupyter-lsp                2.2.5
jupyter_server             2.14.2
jupyter_server_terminals   0.5.3
jupyterlab                 4.3.1
jupyterlab_celltagsclasses 0.5.2
jupyterlab_gridwidth       0.4.3
jupyterlab_nasadaily       1.0.0
jupyterlab_notechat        1.1.2
jupyterlab_pygments        0.3.0
jupyterlab_server          2.27.3

krassowski:
It looks like this one has some extra logic for determining whether it is enabled which probably does not work well with the toolbar:

https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/packages/notebook-extension/src/index.ts#L2376-L2384

We need to add something like:

https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/packages/notebook-extension/src/index.ts#L2263

Issue #16473, opened 2024-06-10
Verdaccio randomly times out on CI


krassowski:
For example https://github.com/jupyterlab/jupyterlab/actions/runs/9451027141/job/26031020797?pr=16465

```
[test f3094b2] [ci skip] bump version
 127 files changed, 2367 insertions(+), 2367 deletions(-)
+ node buildutils/lib/local-repository.js start
Starting verdaccio on port 4873 in /tmp/verdaccio
Timed out!
LOG:
........................................................................................................................npm warn exec The following package was not found and will be installed: verdaccio@5.31.1
npm warn deprecated glob@6.0.4: Glob versions prior to v9 are no longer supported
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.4.5: Rimraf versions prior to v4 are no longer supported
```

krassowski:
Another instance https://github.com/jupyterlab/jupyterlab/actions/runs/9539482409/job/26289874812

krassowski:
It is happening more and more often, now dozens of times in a row. Looking at it closer:


```
[test 815c68a] [ci skip] bump version
 131 files changed, 2455 insertions(+), 2455 deletions(-)
+ node buildutils/lib/local-repository.js start
Starting verdaccio on port 4873 in /tmp/verdaccio
Timed out!
LOG:
........................................................................................................................npm warn exec The following package was not found and will be installed: verdaccio@6.2.1
```

Seems to come from our code here:

https://github.com/jupyterlab/jupyterlab/blob/09411fab41ac68d9b3619ad487940bedb1597035/buildutils/src/local-repository.ts#L100-L122

krassowski:
So it looks like we are conflating start up time of verdaccio with install time because of the use of `npx`.

krassowski:
Closing as fixed by https://github.com/jupyterlab/jupyterlab/pull/18082

PR #17298, opened 2025-02-17
Updated enabling logic for run-all-below button on Notebook Panel

rsaditya01:
## References
Contains changes to resolve [#issue16996](https://github.com/jupyterlab/jupyterlab/issues/16996)

## Code changes

* Addresses the issue of greyed Run-All-Below button on the Panel when a new notebook is opened
* Changes the logic for Run-All-Below button on Notebook Panel
* Runs the selected cell and all the cells below it
* Removes logic for - 
        1. Check for Single Selected cell 
             Reason - Command states that it is for 'Run Selected Cell and All Below'. Option should be there to run this for multiple cells(Single cell check restricts user actions)
        2. Check for Last Cell in the Notebook
             Reason - This is a redundant check. If this is not there, the command runs only for last cell, which is acceptable
         

## User-facing changes

Run-All-Below button will be enabled whenever the notebook is open and the cell is selected.
 
Before Change

image

After Change

image

## Testing

Local Testing Successful

Steps Performed:
1. jlpm run build - Successful
2. jlpm run build:core - Successful
3. jupyter lab build - Successful
4. jupyter lab --dev-mode - Changes getting reflected and the button is enabled to run cells

N/A


astitv-sh (commented):
This section ([Visual Regression and UI Tests](https://jupyterlab.readthedocs.io/en/latest/developer/contributing.html#id9)) of the JupyterLab contributing docs is helpful in understanding why the visual regression tests are failing and how to fix them.

rsaditya01:

andrii-i:

andrii-i:
Kicking CI

michaelnchin:
`release_test` CI failed due to Verdaccio timeout: https://github.com/jupyterlab/jupyterlab/actions/runs/13377892902/job/37361000380?pr=17298

This looks related to the intermittent issue noted in https://github.com/jupyterlab/jupyterlab/issues/16473.

krassowski:
Thanks for linking to https://github.com/jupyterlab/jupyterlab/issues/16473, it would be amazing if someone could spend some cycles on it. I restarted the failing job manually.

galata/test/jupyterlab/notebook-create.test.ts-snapshots/opened-menu-run-jupyterlab-linux.png
krassowski, line 1: This snapshot brings up an interesting point - by the same logic, should "Run All Above Selected Cell" be enabled too?

packages/notebook-extension/src/index.ts
```diff
@@ -2401,15 +2401,7 @@ function addCommands(
...
     },
-    isEnabled: () => {
-      // Can't run below if there are multiple cells selected,
-      // or if we are at the bottom of the notebook.
-      return (
-        isEnabledAndSingleSelected() &&
-        tracker.currentWidget!.content.activeCellIndex !==
-          tracker.currentWidget!.content.widgets.length - 1
-      );
-    }
```
krassowski, line 2412: I see why removing the condition on it being the last cell could be desired - in particular for case of a notebook with a single cell this is indeed annoying.

I do not understand why the guard against having this command enabled when multiple cells are selected is being removed _in this PR_ (I do not personally think it has to stay in general). I am afraid that this will lead to confusing behaviour: depending on which way selection was created, some of the selected cells will not be executed. Have a look at implementation:

https://github.com/jupyterlab/jupyterlab/blob/5ba4755e5a64eae5bb1ce807dd6bc06b6e908273/packages/notebook/src/actions.tsx#L852-L871

If selection is created going from bottom to the top, the active cell will be the top cell and all cells will be executed; if selection gets created from top to bottom, the cells below (and inclusive of) the bottom cell will be executed but the other selected cells won't.

Here is an idea: could we instead add a condition making the command enabled if notebook has only one cell? I think it would solve the original issue without introducing regressions?

krassowski (commented):
Welcome @rsaditya01 :wave:

packages/notebook-extension/src/index.ts
rsaditya01: Thanks Michał, for the detailed review. 

While looking deeply got to understand that the issue majorly focused on issue when the Notebook has only one cell. While performing the local testing I found that your concerns regarding regressions were correct. There was an issue while selecting cells from top to bottom, which concludes that it is better to have check for selection of only one cell. 
I have added a commit to update the code with an additional condition that the Run-All-Below command should be accessible when the notebook has only one cell.

galata/test/jupyterlab/notebook-edit.test.ts-snapshots/split-cell-jupyterlab-linux.png
krassowski, line 1: This snapshot update is from a flaky test and needs to be manually reverted as it causes the test to fail more often than not (we really need to find a way to stabilise this one). You can run:

```
git checkout main galata/test/jupyterlab/notebook-edit.test.ts-snapshots/split-cell-jupyterlab-linux.png
git commit
```

to revert to the version from `main` branch.

krassowski (approved):
Thank you @rsaditya01!

krassowski:

PR #17322, opened 2025-02-21
Fix handling of a `null` banner in the code console

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/13837.

## Code changes

Fix the use of the non-null assertion operator in the code console widget, now that there is an option to hide the banner:

https://github.com/jupyterlab/jupyterlab/blob/e64f81f24594d72f0ad409a1daf2d85ac5778cb3/packages/console/src/widget.ts#L818-L823

This non-null assertion was causing the following error the be logged in the dev tools console:

https://github.com/user-attachments/assets/5d958c0e-ac84-4660-8d46-8602c42f8ff2

Noticed downstream in https://github.com/jupyterlite/jupyterlite/pull/1573.

## User-facing changes

Likely none.

None

### DIFF
--- a/packages/console/src/widget.ts
+++ b/packages/console/src/widget.ts
@@ -815,12 +815,16 @@ export class CodeConsole extends Widget {
    */
   private _handleInfo(info: KernelMessage.IInfoReplyMsg['content']): void {
     if (info.status !== 'ok') {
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you @jtpio!

Issue #15807, opened 2024-02-16
`simulate-event` is no longer maintained, let's migrate to `user-event`


krassowski:
## Description

[`simulate-event`](https://github.com/blakeembrey/simulate-event) was one of the early libraries that enabled simulating user events. However the last commit was in 2016  and in fact contributed by one of JupyterLab maintainers

`simulate-event` does not support many new event types and we end up manually dispatching those:

https://github.com/jupyterlab/jupyterlab/blob/1a98deafa7dc41607106b55936fbdefe9741faaa/packages/testing/src/common.ts#L25-L45

or doing even more aggressive mocks which are very cumbersome and expensive to write especially when jsdom does not ship these events yet (for example see https://github.com/jupyterlab/jupyterlab/pull/15794).

I propose we migrate to [`testing-library/user-event`](https://github.com/testing-library/user-event) which follows the trial set by @fcollonval in jupyterlab-git and proposed for lumino (https://github.com/jupyterlab/lumino/issues/610#issuecomment-1792131525).

Any conter-proposals?

## Context

- JupyterLab version: 4.1

PR #17324, opened 2025-02-21
Prevent duplicate shortcut addition. Fixes #16829

### LINKED ISSUE #16829: Can create the same shortcutfor the same command twice
I can create the same shortcut, <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>]</kbd> (in that order) for the same command twice. Sometimes, it appears twice; at other times, I see an overwrite non-modal dialog, and when I accept, I see one shortcut with a "Reset" option that works.

duplicate-shortcut

_Originally posted by @JasonWeill in https://github.com/jupyterlab/jupyterlab/issues/16721#issuecomment-2389744804_

kjayan:
## References

Fixes #16829.  The block of code that is removed as part of this PR was already having a comment indicating if it should be deprecated. The removal of the same leads to solving the issue that is reported.  Given below is the working post the fix.

https://github.com/user-attachments/assets/c02611d0-50fd-480c-8dba-c7221ba2b16e

This code was added as part [of this commit](https://github.com/jupyterlab/jupyterlab/commit/5af2cc9346a49bc1354448a547413937ab754b56)

I would like to understand if  removal of this code block will lead to re-opening any of the fixed issues or failure of any user flows. Can someone who has worked on this help review the same? 

## Code changes

Removed the block of code to make sure the edit flow of updating an existing extension will detect the duplicate and show the respective warning message

## User-facing changes

https://github.com/user-attachments/assets/c02611d0-50fd-480c-8dba-c7221ba2b16e

NA


krassowski:
Chiming in given the comment in https://github.com/jupyterlab/jupyterlab/issues/16829#issuecomment-2676298049 asking for review (I would usually wait until the PR is marked as ready for review with my feedback):

> This code was added as part [of this commit](https://github.com/jupyterlab/jupyterlab/commit/5af2cc9346a49bc1354448a547413937ab754b56)

Not really

<details>
the logic was added way before (probably 5 years ago if you consider that the 3 years ago git blame refers to refers to incorporation of the extension in this repo, not for its creation), my commit just added the comment and refactored to match a new API that solved numerous bugs in shortcut editor:

image

Before it was:

image

and had zero tests.

</details>

I think my comment does not refer to the logic being deprecated, but that this clause short-circuits presentation of the genuine conflicts which should be exposed - in some way - to the user.

It seems to me that if deleting this block of code helps, it means that `this._isReplacingExistingKeybinding` is set incorrectly. This logic should not be entered when adding a new shortcut, but only in editing if I recall correctly.

I do not see a strong argument for keeping that block of code. I think it might be fine to remove, but I would like this PR to include a test so that if it turns we need to restore the block, we at least will be able to verify that we are doing it in a way which does not regress the issue being addressed.

The would ideally be added in:

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/packages/shortcuts-extension/test/components/ShortcutUI.spec.ts#L46

I think it would make sense to test `ShortcutInput.handleInput` (which calls `checkShortcutAvailability` that you are modifying).

Alternatively an integration test could be added around:

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/galata/test/jupyterlab/settings.test.ts#L242-L331

but if possible we should avoid it as these run much slower and it is easy to write a test which turns out to be flaky.

krassowski:
Re https://github.com/jupyterlab/jupyterlab/pull/17324/commits/9b89c2f7e147a9e1451d6ec6ce9e8138f84933be - comments are welcome in tests (but should be added with `//`)

kjayan:
Hi @krassowski 

Thanks for the detailed review. Based on your inputs, I have worked on the test cases. 
1. There was no test case that was written for ShortcutInput file, hence I added a basic test case and created the file. Wanted to confirm if this is what can be a good path forward. If yes, I can take it forward with more test cases.
2. I have added test cases under AddNewKeyBinding and ReplaceKeyBinding in ShortcutUI.spec.ts 
3. I have added a new set of flows describe('ShortcutInput', () => {}) in ShortcutUI.spec.ts

It would be great to get additional inputs here.

packages/shortcuts-extension/test/components/ShortcutInput.spec.ts
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import React from 'react';
+import { ShortcutInput } from '../../src/components/ShortcutInput';
+import { IShortcutTarget } from '../../src/types';
+import { nullTranslator } from '@jupyterlab/translation';
+import '@testing-library/jest-dom';
```
krassowski, line 10: Is this needed? I think we would want to start using `testing-library` as discussed in https://github.com/jupyterlab/jupyterlab/issues/15807 but it seems to me that this test file does not take advantage of it as I do not see custom matchers this particular package provides? I might be wrong.

```suggestion
```

krassowski:
Thank you, I think this is great. The tests are failing right now due to an import from a non-installed library.

packages/shortcuts-extension/test/components/ShortcutInput.spec.ts
krassowski: @kjayan this is still failing:

```
FAIL  test/components/ShortcutInput.spec.ts
  ● Test suite failed to run

    Cannot find module '@testing-library/jest-dom' from 'test/components/ShortcutInput.spec.ts'

       8 | import { IShortcutTarget } from '../../src/types';
       9 | import { nullTranslator } from '@jupyterlab/translation';
    > 10 | import '@testing-library/jest-dom';
         | ^
      11 | import { PromiseDelegate } from '@lumino/coreutils';
      12 | import { createRoot } from 'react-dom/client';
      13 |

      at Resolver._throwModNotFoundError (../../node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (test/components/ShortcutInput.spec.ts:10:1)
```

krassowski:
Just following up here @kjayan, do you need help in finishing this PR?

kjayan:
> Re [9b89c2f](https://github.com/jupyterlab/jupyterlab/commit/9b89c2f7e147a9e1451d6ec6ce9e8138f84933be) - comments are welcome in tests (but should be added with `//`)

Thanks for pointing this out. I have addressed this.

kjayan:
> Just following up here @kjayan, do you need help in finishing this PR?

Thanks for the follow up. I got caught up with some project work and couldn't complete it. I have made the changes and marked it ready for review.

krassowski:
@kjayan thank you for adding the tests. I reverted the removal of code from `ShortcutInput.tsx` in https://github.com/krassowski/jupyterlab/pull/142/commits/53902ff95f393424e7fd23077a2dea4cf1e30537 but [all the shortcut tests passed](https://github.com/krassowski/jupyterlab/actions/runs/15558543859/job/43804940614?pr=142). I think the added tests do not capture the regression being addressed.

krassowski:
@kjayan are you interested in finishing this one up?

MUFFANUJ:
This has picked up conflicts! @kjayan Another ping to verify whether you want to work on this and finish it up, or let someone else take over and finish it.

PR #17325, opened 2025-02-21
Provide `ITranslatorConnector` via a separate plugin

jtpio:
## References

It turns out that a `ITranslatorConnector` token was already defined as part of the `@jupyterlab/translation` tokens here: 

https://github.com/jupyterlab/jupyterlab/blob/c0bb677556bb7fffab501df43188271c2a7cdd72/packages/translation/src/tokens.ts#L23-L29

However until now it was not exposed as a service via a plugin, since the `TranslationManager` currently instantiates its own connector here: 

https://github.com/jupyterlab/jupyterlab/blob/c0bb677556bb7fffab501df43188271c2a7cdd72/packages/translation/src/manager.ts#L18

This PR extracts the translation connector so it can be provided via its own plugin. It looks like it was meant to be exposed via a plugin but never ended up being the case?

This will allow downstream extensions and applications to more easily provide a different way for fetching the different bundles if they want to, for example if they are served from a different source or via a different endpoint. With this PR, they can simply replace the connector by providing a plugin exposing a `ITranslatorConnector`, instead of having to replace the `@jupyterlab/translation:translator` plugin entirely to provide a custom `TranslationManager`.

## Code changes

To stay backwards compatible, the `connector` is passed as an optional parameter when instantiating the `TranslationManager`.

## User-facing changes

None

None

### DIFF
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -17,8 +17,10 @@ import { IMainMenu } from '@jupyterlab/mainmenu';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import {
   ITranslator,
+  ITranslatorConnector,
   requestTranslationsAPI,
-  TranslationManager
+  TranslationManager,
+  TranslatorConnector
 } from '@jupyterlab/translation';
 
 /**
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Makes sense, thank you @jtpio!

PR #17327, opened 2025-02-21
Provide `ISettingsConnector` via a separate plugin

jtpio:
## References

Similar to https://github.com/jupyterlab/jupyterlab/pull/17325 but for the settings.

## Code changes

Provide `ISettingsConnector` via a separate plugin, so it can more easily be swapped from an extension.

One use case is to allow for extension to store settings in the browser instead of the server.

Although this should also now be possible after #16794 by providing a custom `Setting.IManager` and replace the following plugin:

https://github.com/jupyterlab/jupyterlab/blob/ad4e2d979d37e90ef91530a8fe4e147078668ccf/packages/services-extension/src/index.ts#L192-L207

This PR enables a similar possibility, except that downstream extensions could then provide a regular `JupyterFrontendPlugin` instead of a `ServiceManagerPlugin` (considered more advanced).

As a follow-up, we could also consider deprecating the use `app.serviceManager.settings`.

## User-facing changes

None

None

### DIFF
--- a/packages/apputils-extension/src/settingsplugin.ts
+++ b/packages/apputils-extension/src/settingsplugin.ts
@@ -8,18 +8,40 @@ import {
   JupyterFrontEndPlugin
 } from '@jupyterlab/application';
 import { PageConfig } from '@jupyterlab/coreutils';
-import { ISettingRegistry, SettingRegistry } from '@jupyterlab/settingregistry';
+import {
+  ISettingRegistry,
+  ISettingsConnector,
+  SettingRegistry
+} from '@jupyterlab/settingregistry';
 import { SettingConnector } from './settingconnector';
+
+/**
+ * The default settings connector.
+ */
+export const settingsConnector: JupyterFrontEndPlugin<ISettingsConnector> = {
+  id: '@jupyterlab/apputils-extension:settings-connector',
+  autoStart: true,
+  provides: ISettingsConnector,
+  activate: (app: JupyterFrontEnd) =>
+    new SettingConnector(app.serviceManager.settings)
+};
 
 /**
  * The default setting registry provider.
  */
 export const settingsPlugin: JupyterFrontEndPlugin<ISettingRegistry> = {
   id: '@jupyterlab/apputils-extension:settings',
+  autoStart: true,
+  provides: ISettingRegistry,
+  optional: [ISettingsConnector],
   description: 'Provides the setting registry.',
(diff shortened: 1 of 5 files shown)


packages/settingregistry/src/tokens.ts
```diff
@@ -27,6 +35,12 @@ export const ISettingRegistry = new Token<ISettingRegistry>(
   See "schemaDir" for more information.`
 );
 
+/**
+ * The settings connector interface.
+ */
+export interface ISettingsConnector
```
krassowski, line 41: Should it be singular or plural to match the name of the class?

jtpio: Good catch, thanks!

packages/settingregistry/src/tokens.ts
```diff
@@ -17,6 +17,14 @@ import { ISignal } from '@lumino/signaling';
 import { ISchemaValidator } from './settingregistry';
 import type { RJSFSchema, UiSchema } from '@rjsf/utils';
 
+/**
+ * A service to connect to the server endpoint.
```
krassowski, line 21: Did you mean

```suggestion
 * A service to connect to the settings endpoint.
```

or was it deliberate?

packages/apputils-extension/src/settingsplugin.ts
```diff
@@ -8,18 +8,43 @@ import {
...
-import { ISettingRegistry, SettingRegistry } from '@jupyterlab/settingregistry';
+import {
+  ISettingConnector,
+  ISettingRegistry,
+  SettingRegistry
+} from '@jupyterlab/settingregistry';
 import { SettingConnector } from './settingconnector';
 
+/**
+ * Provide the settings connector as a separate plugin to allow for alternative
```
krassowski, line 19:
```suggestion
 * Provides the settings connector as a separate plugin to allow for alternative
```

krassowski (approved):
Thank you @jtpio!

packages/settingregistry/src/tokens.ts
jtpio: Right, `settings` makes more sense.

brichet:
Maybe in a follow up PR we could export the default [`SettingConnector`](https://github.com/jupyterlab/jupyterlab/blob/main/packages/apputils-extension/src/settingconnector.ts#L20), e.g. moving it to `apputils` package. 
That way, third party extensions providing their own `SettingConnector` could extend it.

PR #17328, opened 2025-02-22
Normalize translator plugin ids

jtpio:
## References

Rename the `@jupyterlab/translation-extension` plugins to they follow the same naming convention as other JupyterLab plugins.

## Code changes

- Rename `@jupyterlab/translation:translator` to `@jupyterlab/translation-extension:translator`
- Rename `@jupyterlab/translation:translator-connector` to `@jupyterlab/translation-extension:translator-connector`

## User-facing changes

None

None

Mention the change in the extension migration just in case folks were relying on that plugin or replacing it.

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -97,6 +97,12 @@ the updated rendering logic.
 This change improves both the rendering performance and the maintainability of extensions
 using the ``GroupItem`` component.
 
+Plugins
+-------
+
+- ``@jupyterlab/translation-extension`` from 4.3 to 4.4
+   * The ``@jupyterlab/translation:translator`` plugin id has been renamed to ``@jupyterlab/translation-extension:translator``
+
 JupyterLab 4.2 to 4.3
 ---------------------
 
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -34,7 +34,7 @@ const PLUGIN_ID = '@jupyterlab/translation-extension:plugin';
  * source or endpoint.
  */
 const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
-  id: '@jupyterlab/translation:translator-connector',
+  id: '@jupyterlab/translation-extension:translator-connector',
   description: 'Provides the application translation connector.',
   autoStart: true,
   requires: [JupyterFrontEnd.IPaths],
@@ -50,7 +50,7 @@ const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
  * The main translator plugin.
  */
 const translator: JupyterFrontEndPlugin<ITranslator> = {
-  id: '@jupyterlab/translation:translator',
+  id: '@jupyterlab/translation-extension:translator',
   description: 'Provides the application translation object.',
   autoStart: true,
   requires: [JupyterFrontEnd.IPaths, ISettingRegistry],


docs/source/extension/extension_migration.rst
```diff
@@ -97,6 +97,12 @@ the updated rendering logic.
 This change improves both the rendering performance and the maintainability of extensions
 using the ``GroupItem`` component.
 
+Plugins
+-------
+
+- ``@jupyterlab/translation-extension`` from 4.3 to 4.4
+   * The ``@jupyterlab/translation:translator`` plugin id has been renamed to ``@jupyterlab/translation-extension:translator``
```
jtpio, line 104: Only list this plugin, as `@jupyterlab/translation-extension:translator-connector` was added in https://github.com/jupyterlab/jupyterlab/pull/17325 for `4.4`.

packages/translation-extension/src/index.ts
```diff
@@ -34,7 +34,7 @@ const PLUGIN_ID = '@jupyterlab/translation-extension:plugin';
  * source or endpoint.
  */
 const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
-  id: '@jupyterlab/translation:translator-connector',
+  id: '@jupyterlab/translation-extension:translator-connector',
```
krassowski, line 37: :+1:

packages/translation-extension/src/index.ts
```diff
@@ -50,7 +50,7 @@ const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
  * The main translator plugin.
  */
 const translator: JupyterFrontEndPlugin<ITranslator> = {
-  id: '@jupyterlab/translation:translator',
+  id: '@jupyterlab/translation-extension:translator',
```
krassowski, line 53: This might break `rise` as it depends on it [here](https://github.com/jupyterlab-contrib/rise/blob/c912db7c8f28df999724e791ed2918d5291ae608/app/index.js#L135-L137).

I see why we should change it. I am not convinced if should change this in 4.4 yet.

jtpio: Right, this would mostly affect downstream lab based applications such as `rise`, but also `notebook`.

I would say it's probably fine since these can easily be updated? Also in the past there were a few other similar changes that were about renaming plugins, or moving some functionalities to separate plugins, in the early days of RetroLab / Notebook 7, which was released in minor releases.

krassowski: I think the severity here is that `ITranslator` is a required plugin for many many plugins, and without it the plugins will fail to activate, so it is not the same as renaming something in a debugger or another leaf plugin.

krassowski: To give a few random examples:

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/packages/apputils-extension/src/themesplugins.ts#L54

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/packages/statusbar-extension/src/index.ts#L26

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/packages/imageviewer-extension/src/index.ts#L76

jtpio: For sure, but here it's the id of the plugin providing the token which is being updated, not the name of the token, which stays unchanged:

https://github.com/jupyterlab/jupyterlab/blob/7041277fe5d5a547f6d8795c61565e8424895819/packages/translation/src/tokens.ts#L70-L73

krassowski: Hmm.. but `rise` would still fail, right? I think that `require().default` points to array of plugins, and the code in `rise` checks plugin IDs?

image

So there will be no provider for `ITranslator` so these extensions which require it will fail to load.

jtpio: It would indeed fail for `rise` next time someone it updates to use `"@jupyterlab/translation-extension": "^4.4.0"` [here](https://github.com/jupyterlab-contrib/rise/blob/c912db7c8f28df999724e791ed2918d5291ae608/app/package.json#L37). So as part of the change that would update to newer `@jupyterlab` dependencies, `rise` the contributor will also need to update this line to:

```ts
['@jupyterlab/translation-extension:translator'].includes(id)
```

We can open an issue on rise repo to not forget about, even though it will be quite visible when doing the `@jupyterlab` update as many plugins will indeed fail to activate.

krassowski (approved):

PR #17329, opened 2025-02-22
Use `ITranslationConnector` to fetch the translations

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17325

## Code changes

Fix the use of a raw API request to fetch the translation.

Instead, use the translator connector exposed via a plugin in https://github.com/jupyterlab/jupyterlab/pull/17325.

## User-facing changes

None

None

### DIFF
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -152,6 +152,7 @@ const langMenu: JupyterFrontEndPlugin<void> = {
         // Get list of available locales
         translatorConnector
           .fetch({ language: '' })
+          // TODO: fix usage of any
           .then((data: any) => {
             for (const locale in data['data']) {
               const value = data['data'][locale];
--- a/packages/translation-extension/src/index.ts
+++ b/packages/translation-extension/src/index.ts
@@ -18,7 +18,6 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import {
   ITranslator,
   ITranslatorConnector,
-  requestTranslationsAPI,
   TranslationManager,
   TranslatorConnector
 } from '@jupyterlab/translation';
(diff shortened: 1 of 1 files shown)


packages/translation-extension/src/index.ts
```diff
@@ -149,10 +149,11 @@ const langMenu: JupyterFrontEndPlugin<void> = {
 
         let command: string;
 
-        const serverSettings = app.serviceManager.serverSettings;
         // Get list of available locales
-        requestTranslationsAPI<any>('', '', {}, serverSettings)
-          .then(data => {
+        translatorConnector
+          .fetch({ language: '' })
+          // TODO: fix usage of any
```
jtpio, line 155: In its current state, the translation extension does not seem to be using proper types in the code base.

Added a `TODO` here so we can track improving that in an issue and come back to it later.

krassowski: It means `fetch` return type is incorrect. That means we are exposing an invalid contract in newly fixed `ITranslatorConnector`, so yeah this should be addressed before we release 4.4.

Maybe you could cherry-pick from my https://github.com/jupyterlab/jupyterlab/pull/16219 - it is rather large so pasting snapshot of relevant section:

image

And possibly also

image

If you think it should be done in a separate PR that's fine too.

jtpio: > Maybe you could cherry-pick from my #16219 - it is rather large so pasting snapshot of relevant section:

Ah nice thanks for the pointer 👍 We should also try to get such PR across the finish line as it would be quite helpful in general.



> If you think it should be done in a separate PR that's fine too.

Yes happy to track separately (as a release blocker) and be assigned to it. This can be addressed later during the beta or rc cycles.

krassowski (approved):
Thank you @jtpio!

packages/translation-extension/src/index.ts
krassowski: >  We should also try to get such PR across the finish line as it would be quite helpful in general.

Yes, probably early in the 4.5 cycle.

PR #17336, opened 2025-02-23
Clarify documentation on bot privileges for updating snapshots

### LINKED ISSUE #17174: Clarify documentation on bot privileges for updating snapshots
The current documentation states that the bot can handle updating snapshots, but it does not clarify that this feature is restricted to a subset of users with higher privileges due to security concerns. This caused some confusion as I initially expected the bot to handle the task without requiring elevated privileges.

Image

###Expected Changes:

Clearly specify in the documentation that updating snapshots via the bot is restricted to certain users with higher privileges.

We had to restrict it to a subset of users with higher privileges to mitigate security risks. The documentation needs to be updated.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/17173#issuecomment-2593760999_

Princekumarofficial:
This PR updates the documentation to explicitly state that the bot's ability to update snapshots is restricted to a subset of users with higher privileges due to security concerns. This change addresses confusion where users might assume the bot can handle snapshot updates for all users.

## References

Closes #17174 

## Code changes

No code changes, only documentation updates.

### Documentation changes

Updated the relevant documentation section to clarify that only users with elevated privileges can use the bot to update snapshots.

## User-facing changes

Clearer documentation to prevent misunderstandings regarding bot permissions.

None

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -692,7 +692,8 @@ Main reasons for UI test failures are:
    - ``please update snapshots``: Combine the two previous comments effects.
 
     The bot will react with +1 emoji to indicate that the run started and then comment
-    back once it concluded.
+    back once it concluded. This feature is restricted to a subset of users with higher
+    privileges due to security concerns.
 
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you @Princekumarofficial!

krassowski:

PR #17338, opened 2025-02-24
Enforce plugin name convention and rename noncompliant plugins

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17328

This will help avoid PRs like https://github.com/jupyterlab/jupyterlab/pull/17328 in the future, so plugin ids follow the same convention:

```
@jupyterlab/<name>-extension:<type>
```

Technically plugin ids can be set to anything, but it's nice to have a convention so it's easier to consume them downstream (in custom apps or via `disabledExtensions`).

## Code changes

- [x] Ensure plugin ids follow the same naming convention
- [x] Mention renamed plugin ids in the extension migration guide
- [x] Sort the list of plugins generated by the UI tests
- [x] Add mapping of renamed plugins to help with the migration of user config (`disabledExtensions`, `deferredExtensions`)

## User-facing changes

None

None


jtpio:
Example failure when some plugins don't follow the convention:

```
1) [documentation] › test/documentation/plugins.test.ts:76:5 › Plugins must be named using the same convention 
    Error: expect(received).toEqual(expected) // deep equality
    - Expected  - 1
    + Received  + 5
    - Array []
    + Array [
    +   "@jupyterlab/lsp:ILSPCodeExtractorsManager",
    +   "@jupyterlab/workspaces:commands",
    +   "@jupyterlab/application:mimedocument",
    + ]
```

image

galata/test/documentation/plugins.test.ts
```diff
@@ -6,22 +6,28 @@ import { Token } from '@lumino/coreutils';
 import { expect } from '@playwright/test';
 import * as fs from 'fs-extra';
 
+// As the test is ran in the documentation environments, some plugins and tokens are not from core.
```
afshin, line 9:
```suggestion
// As the test is run in the documentation environment, some plugins and tokens are not from core.
```

afshin (approved):

krassowski:
As suggested by @ericsnekbytes during the frontends call it would be great if we could map the old plugin names to new ones (and emit a warning about migration)

jtpio:
> As suggested by @ericsnekbytes during the frontends call it would be great if we could map the old plugin names to new ones (and emit a warning about migration)

Yes that sounds like a good idea. Will have a look.

packages/workspaces-extension/src/commands.ts
```diff
@@ -75,7 +75,7 @@ export const WORKSPACE_ITEM_CLASS = 'jp-mod-workspace';
  * The workspace commands
  */
 export const commandsPlugin: JupyterFrontEndPlugin<IWorkspaceCommands> = {
-  id: '@jupyterlab/workspaces:commands',
+  id: '@jupyterlab/workspaces-extension:commands',
```
jtpio, line 78: This will actually "fix" the documentation inconsistency in the extension migration guide: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#api-updates

image

jtpio:
Pushed a change to handle the case or renamed plugins.

For example in case a user had `@jupyterlab/workspaces:commands` as part of `disabledExtensions`, it will automatically be renamed to `@jupyterlab/workspaces-extension:commands` and the corresponding plugin will still be disabled.

Not sure how long we want to keep that mapping of renamed plugins, since it adds a bit more complexity to the JupyterLab `index.js`. Maybe a couple of minor releases will be enough time for users to update their config to use the new names.

Issue #16952, opened 2024-11-11
SVG rendering bug with duplicate IDs


paddymul:
## Description

When SVGs are rendered with duplicate IDs dynamically that point at patterns, the patterns show/hide on scroll.  This ID collision is global (you will see weird behavior based on what other notebook tabs are open).  This behavior only occurs dynamically, not in statically rendered notebooks.  I have observed this in both Firefox and Chrome

https://github.com/user-attachments/assets/ebdc5c74-05e6-488b-a045-189a63470565

## Reproduce
Here is a statically rendered gist of the same ntoebook.  When statically rendered this behavior isn't present.
To reproduce, clear all output, then save and execute cells one by one.
https://gist.github.com/paddymul/8ecf0251f74a5ab1813971fc46b612a9

1. Open this notebook https://gist.github.com/paddymul/8ecf0251f74a5ab1813971fc46b612a9
2. Clear all output and save
3. execute each cell, one by one and scroll

Here is a log of all of the things I tried
https://github.com/recharts/recharts/issues/5230
## Expected behavior

I am not sure what should be expected.  I don't really expect jupyter to do anything to fix this.  I would like a record of this surprising behaviour so that if someone else stumbles onto it, they have some hope of understanding the problem.  I will fix this with dynamically generated IDs in my SVGs

I have tried this sith and without shared definitions of Patterns.  I would think that with only a single definition of the pattern in the do for an ID, there wouldn't be ID collisions, but that doesn't seem to be the case.  Addtionally, it doesn't seem possible to define patterns with class names.  I haven't figured out how to define these types of patterns in CSS.

## Context

- Operating System and version: OS X (13)
- Browser and version:  Firefox (132) and Chrome (130) current
- JupyterLab version: jupyterlab                    4.2.5

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
jupyter troubleshoot
$PATH:
	/Users/paddy/.nvm/versions/node/v20.18.0/bin
	/Users/paddy/anaconda3/envs/buckaroo-dev-9/bin
	/Users/paddy/anaconda3/condabin
	/Users/paddy/.poetry/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	/usr/local/bin
	/System/Cryptexes/App/usr/bin
	/usr/bin
[...]

Issue #13172, opened 2022-10-03
Dealing with iframe and windowing


fcollonval:
I use iframes a lot in my educational notebooks, so would definitely like to have a fix for this.

Two thoughts, fiddling with a notebook, and applying styles to the `.jp-Cell-outputWrapper`:

```css
z-index: -1;
position: absolute;
```
The Cocalc trick you mentioned.

```css
height:0;
overflow: auto;
```
This seems to hide the frame as well and, on Chrome at least, preserves the scroll state.

_Originally posted by @Pugio in https://github.com/jupyterlab/jupyterlab/pull/12554#discussion_r984950194_

williamstein:
This was just discussed in the JupyterLab call for a moment during triage. For what it is worth, I changed how I deal with iframes and windowing in CoCalc to use html that is at the top level of the DOM, and i use a trick involving an svg clip-region.  Contact me if you want to discuss this...

krassowski:
So this is interesting: the issue on Chromium tracker is still open (https://issues.chromium.org/issues/41368291), but the scroll position is now correctly preserved for me in Chrome 133.

Issue #17331, opened 2025-02-22
Allow renderers to declare if they support detachment in windowing


krassowski:
## Problem

Currently, `full` windowing leads to a number of issues:

- alternative Markdown renderers (`jupyterlab-myst`) may resize the cell on display, causing scroll jumps https://github.com/jupyter-book/jupyterlab-myst/issues/243
- cells with outputs that contains:
  - `iframes` loose state when re-rendered after scrolling (https://github.com/jupyterlab/jupyterlab/issues/13172)
  - `<defs>` do not apply if out of viewport (https://github.com/jupyterlab/jupyterlab/issues/16952)
  - logic that resizes the cell on display causes jitter (https://github.com/jupyterlab/jupyterlab/issues/16326)
  
Most if not all of these issues can be solved by not using `display: none` but "soft-hiding" cells with `opacity: 0; height: 0`; we already do that for active cell to allow interaction when it is outside of the viewport.

Moreover, we use `display: none` not because it is optimal, but because removing cells with HTML output from the DOM would actually lead to even more issues, e.g. CSS styles from outputs of a cell outside of viewport would not longer apply after scrolling.

### Proposed Solution

Maybe we should allow renderers to decide what the windowing logic can do with a cell.

In general there are three levels with decreasing performance benefits but also decreasing side-effect risks:

1. `detach-from-dom` - removing from DOM - currently we never do that as it would cause styles contributed by a cell with `<style>` tag in output to disappear depending on scroll position; however, if cell does not include outputs or includes only plaintext renderers, this is safe to detach
2. `remove-from-layout` - `display: none` and will trigger resizing and other known issues
2. `hide` means `opacity:0; height: 0; padding: 0`  (similar to `visibility: hidden`) and has no known issues as of today.

So maybe we should add an optional `windowingBehavior` attribute to renderers and allow these to tell what is safe to do with a given rendered widget. For example:

```ts
Renderer {
   windowingBehavior: "detach-from-dom" | "remove-from-layout" | "hide"
}
```

This would then be mirrored in `MarkdownCell` based on Markdown render in use, and aggregated across all outputs in `CodeCell`, choosing the behaviour which is compatible with all outputs (this is "hide" in worst case).

### Additional context

[...]

PR #17339, opened 2025-02-24
Improve scrolling edge cases

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/16952
- Fixes https://github.com/jupyter-book/jupyterlab-myst/issues/243
- Mostly fixes https://github.com/jupyterlab/jupyterlab/issues/16326
- Could be used to fix https://github.com/jupyterlab/jupyterlab/issues/13172 but I am unable to repro this anymore so I did not add `iframe` in the list of conditions

## Code changes

- [x] stop `IntersectionObserver`s on cells in a notebook when scrolling (resume one second after scrolling stopped)
  - [ ] <s>only throttle on per-target basis</s>
- [x] do not flip classes and `tabIndex` on cells to avoid spurious layout invalidation
- [x] use `transform: translateY(${top}px)` instead of `top: ${top}px` to reduce layout cost as this can be handled by composer
- [x] soft-hide cells which include `<defs>`
- [x] soft-hide cells which resize frequently
  - [ ] make the logic for detection which cells resize frequently more robust
  - [ ] OR use the approach where it is the render's responsibility to disclose the need for soft-hiding as explored in https://github.com/jupyterlab/jupyterlab/issues/17331
  - [ ] OR detect if the cell/cell output has a resize observer to decide on soft-hiding
  - [x] OR pause resize observers in cells during scrolling?
- [x] disable native [scroll anchoring](https://drafts.csswg.org/css-scroll-anchoring/)
- [x] <s>preserve `scrollTop` if it gets modified when cells get attached/detached</s> this trashes layout, negatively impacting scrolling performance

## User-facing changes

Scrolling is much more pleasant.

None


packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1161,10 +1161,21 @@ export class WindowedList<
...
 
+      if (this._viewport.dataset.isScrolling != 'true') {
+        this._viewport.dataset.isScrolling = 'true';
+      }
+
+      if (this._timerToClearScollStatus) {
+        window.clearTimeout(this._timerToClearScollStatus);
+      }
+      this._timerToClearScollStatus = window.setTimeout(() => {
+        this._viewport.dataset.isScrolling = 'false';
```
krassowski, line 1172: [`scrollend`](https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollend_event) event is a thing but not in Safari (https://bugs.webkit.org/show_bug.cgi?id=201556)

packages/notebook/src/windowing.ts
```diff
@@ -10,10 +10,107 @@ import {
...
+    root &&
+    root instanceof ShadowRoot &&
+    isInScrollingNotebook(root.host)
+  );
+}
+
+/**
+ * Subclass IntersectionObserver to allow suspending callbacks when notebook is scrolling.
+ */
+window.IntersectionObserver = class extends window.IntersectionObserver {
```
brichet, line 42: Out of curiosity, where is this observer attached to the cell ? I can't find any use of `IntersectionObserver` in the codebase of Jupyterlab, except for the `logconsole`...

krassowski: It is not in the JupyterLab codebase, but outputs can contain aribtrary this party widgets. The issue #16326 highlights the case of panel tabulator which has both resize and intersection observer and these fire on scrolling causing resizing of the viewport.

brichet: Thanks for the answer.
I was testing this PR with `jupyterlab_myst` and it seems that it fixes https://github.com/jupyter-book/jupyterlab-myst/issues/243, and I was thinking it was because of this `IntersectionObserver`, this is why I was confused.
I just realised that the change related to the `myst` issue is probably 
```ts
const requiresSoftHiding = widget.node.querySelector('defs,.myst');
```

krassowski: Yes, as documented by the comment:

```js
    // - elements containing `.myst` class as jupyterlab-myst
    //   re-renders cause height jitter and scrolling issues;
    //   in future this could be addressed by the proposal
    //   to allow renderers to specify hiding mode, see
    //   https://github.com/jupyterlab/jupyterlab/issues/17331
```

I do not like special-casing `myst` here, but given the prominence of that extension (and it being a part of an official Jupyter subproject) I feel this is warranted.

packages/ui-components/src/components/windowedlist.ts
brichet: We could add a comment about that, to fix it when it is available in Safari.

brichet:
Most of the test failures seems to be a little offset in rendered text.
In my opinion it looks fine, but maybe I missed some more important changes.

krassowski:
> Most of the test failures seems to be a little offset in rendered text.

I believe this comes from the use of `transform: translateY(${top}px)` over `top: ${top}px`. I do not see a way to avoid this, so we will need to update snapshots.

krassowski:
I reverted the `scrollTop` pin as this leads to worse performance. I do not see a way to fix this in general, other than rewriting the windowing mode altogether (https://github.com/jupyterlab/jupyterlab/issues/16327).

krassowski:
bot please update snapshots

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1161,10 +1161,21 @@ export class WindowedList<
...
       this._scrollUpdateWasRequested = false;
 
+      if (this._viewport.dataset.isScrolling != 'true') {
+        this._viewport.dataset.isScrolling = 'true';
+      }
+
+      if (this._timerToClearScollStatus) {
+        window.clearTimeout(this._timerToClearScollStatus);
+      }
+      this._timerToClearScollStatus = window.setTimeout(() => {
```
afshin:
```suggestion
      this._timerToClearScrollStatus = window.setTimeout(() => {
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1161,10 +1161,21 @@ export class WindowedList<
...
+        this._viewport.dataset.isScrolling = 'false';
+      }, 500);
       this.update();
       // }
     }
   }
+  private _timerToClearScollStatus: number | null = null;
```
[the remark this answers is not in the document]
krassowski: This has already been fixed in https://github.com/jupyterlab/jupyterlab/pull/17339/commits/0a38f1789dfb0f3fb94981112160917f61b35027, GitHub just keeps showing the outdated code because I made my comment on the old version.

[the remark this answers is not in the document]
afshin: Oh sorry for the noise, thank you!

afshin (approved):
This feels much nicer, thank you!

krassowski:
>  use `transform: translateY(${top}px)` instead of top: `${top}px` to reduce layout cost as this can be handled by composer

I think not all the snapshot changes are as intended. It looks like there is some extra scrolling occurring when entering a cell. I gather that the `scrollToItem` `smart` logic inputs are incorrect. I am not sure if this is down to `transform` or lack of explicit `minHeight`. I will revert `transform` part to see if tests pass without it.

https://github.com/jupyterlab/jupyterlab/blob/bde76310b2d7023ef1625a63840d8108df99c7da/packages/ui-components/src/components/windowedlist.ts#L389-L416

packages/ui-components/style/windowedlist.css
```diff
@@ -7,6 +7,7 @@
   height: 100%;
   position: relative;
   overflow: auto;
+  overflow-anchor: none;
```
krassowski, line 10:
```suggestion
```

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1161,10 +1161,22 @@ export class WindowedList<
...
+        this._viewport.dataset.isScrolling = 'true';
+      }
+
+      if (this._timerToClearScrollStatus) {
+        window.clearTimeout(this._timerToClearScrollStatus);
+      }
+      // TODO: use `scrollend` event instead once supported by Safari
+      this._timerToClearScrollStatus = window.setTimeout(() => {
+        this._viewport.dataset.isScrolling = 'false';
+      }, 500);
```
krassowski, line 1174: Re failing tests: given that all other suspected pieces of code were eliminated, this looks like a plausible source of flakiness. I will try implementing `scrollend` support which should help as we are testing on chromium.

The tricky bit is that we do not know if `scrollend` is supported. We may need to allow-list it for Chromium and Firefox? But then testing for browsers is also tricky.

krassowski:

krassowski:
After removing `minHeight`, the failing test is:

```
  1) [jupyterlab] › test/jupyterlab/windowed-notebook.test.ts:48:5 › should update window height on resize 

    Error: expect(received).toBeLessThan(expected)

    Expected: < NaN
    Received:   NaN
```

I introduced it two years ago in https://github.com/jupyterlab/jupyterlab/pull/15357 to together with a fix for https://github.com/jupyterlab/jupyterlab/issues/15330.

Testing for `minHeight` was only testing if side-effect of that bug got fixed. I can confirm that removing `minHeight` does not reintroduce the bug (which makes sense as to fix it I only needed to invoke `_update()` on resize). It seems that we no longer need to set `minHeight`, but we need to rework this test.

packages/ui-components/src/components/windowedlist.ts
```diff
@@ -1161,6 +1166,17 @@ export class WindowedList<
...
 
+      if (this._viewport.dataset.isScrolling != 'true') {
+        this._viewport.dataset.isScrolling = 'true';
+      }
+
+      if (this._timerToClearScrollStatus) {
+        window.clearTimeout(this._timerToClearScrollStatus);
+      }
+      // TODO: remove once `scrollend` event is supported by Safari
+      this._timerToClearScrollStatus = window.setTimeout(() => {
```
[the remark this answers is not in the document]
krassowski: It is working, it is just that the snapshots are taken immediately thus flaky. We would need to add ~1s delay before taking each snapshot, which would make the testing suite slow. This is why using `scrollend` event helps.

galata/test/jupyterlab/windowed-notebook.test.ts
```diff
@@ -56,29 +48,28 @@ test('should update window height on resize', async ({ page, tmpPath }) => {
...
   await page.evaluate(async () => {
     const widget = await window.jupyterapp.commands.execute('launcher:create');
     window.jupyterapp.shell.add(widget, 'main', { mode: 'split-bottom' });
   });
-  // Measure height after splitting the dock area
-  const heightAfterSplit = await getWindowHeight(notebook!);
 
-  expect(heightAfterSplit).toBeLessThan(fullHeight);
+  // The cell should no longer be visible
+  await expect.soft(cell).not.toBeVisible();
```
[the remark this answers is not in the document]
krassowski: > I wonder if this test is relevant, as the cell is hidden by the bottom widget.

Good point.

> Maybe we should test that the cell input is not attached ?

Code cells are always attached, sometimes they have `display:none` or `opacity:0` ("soft hiding").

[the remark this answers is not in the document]
krassowski: Hmm, it looks like my test is fine as per playwright docs:

https://playwright.dev/docs/actionability#visible

> Element is considered visible when it has non-empty bounding box and does not have visibility:hidden computed style.
> 
> Note that according to this definition:
> - Elements of zero size **are not** considered visible.
> - Elements with display:none **are not** considered visible.
> - Elements with opacity:0 **are** considered visible.

This issue confirms that whether another element overlaps it has no bearing on its visibility: 
https://github.com/microsoft/playwright/issues/9923

paddymul:
Thanks. I'll take a look soon!

PR #17348, opened 2025-02-27
Add `IKernelAPIClient` and `ISessionAPIClient` as options for `Kernel.IManager` and `Session.IManager`

jtpio:
## References

As a follow-up to https://github.com/jupyterlab/jupyterlab/pull/16794, extensions can now provide a different `Kernel.IManager` or `Session.IManager` via a plugin. This is great, as it allows experimenting with different custom implementations, including the use case of in-browser kernels.

Although extensions can implement and provide their own `Kernel.IManager`, `Session.IManager`, `Kernel.IKernelConnection`, `Session.ISessionConnection`, it is still a lot of logic to implement, such as handling kernel messages or message hooks. Some of the logic from the concrete `KernelManager` and `SessionManager` implementations often needs to be duplicated since it is currently privately defined. 

The current implementations of `KernelManager`, `SessionManager`, `KernelConnection` and `SessionConnection` talk to the respective Jupyter Sever API endpoints, as illustrated on this diagram: https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/README.md#session-context. Because of this, extensions may need to provide implementations for these 4 concrete classes.

In the case of in-browser kernels, the custom kernel and session managers don't necessarely need to rewrite a `KernelManager` and `SessionManager` from scratch. They mostly need to be able to swap the part that makes requests to the Jupyter Server API. 

In the current state of https://github.com/jupyterlite/jupyterlite/pull/1590, the custom `KernelManager`, `SessionManager`, `KernelConnection` and `SessionConnection` are derived from the concrete implementations from `@jupyterlab/services`, with a few methods overridden, to avoid duplicating too much code from the base implementations. However this also requires accessing private attributes and methods to be able to reuse the rest of the upstream implementation.

With this PR, the proposal is to add more granularity to what can be configured when instantiating a `KernelManager` or `SessionManager`. In this case, it extracts the functions that make requests to the Jupyter Server REST API, so a different implementation could be provided when instantiating a new `KernelManager` and `SessionManager`.

This should make the use case of downstream extensions like JupyterLite easier to handle, as they could then define their plugin in a much more minimal way, for example:

```ts
const kernelManagerPlugin: ServiceManagerPlugin<Kernel.IManager> = {
  id: 'example:kernel-manager',
  description: 'Custom kernel manager plugin.',
  autoStart: true,
  provides: IKernelManager,
  optional: [IServerSettings],
  activate: (_: null, serverSettings?: ServerConnection.ISettings): Kernel.IManager => {
    class CustomKernelAPIClient implements Kernel.IKernelAPIClient {
      // implement the functions
    }
    const kernelAPIClient = new CustomKernelAPIClient();
    return new KernelManager({ kernelAPIClient, serverSettings });
  },
};
```

## Code changes

- [x] Add `Kernel.IKernelAPIClient` and `Session.ISessionAPIClient`
- [x] Allow `Kernel.IManager` and `Kernel.IKernelConnection` to take an optional `Kernel.IKernelClient`
- [x] Allow `Session.IManager` and `Session.ISessionConnection` to take an optional `Session.IKernelClient`

The names are currently set to `KernelAPIClient` / `SessionAPIClient` to not conflict with the already publicly exposed `KernelAPI` / `SessionAPI`.

## User-facing changes

None

None.

The current `KernelAPI` and `SessionAPI` are preserved.

If we want to be cautious, we could also mark the new options as experimental for now.

### DIFF
--- a/packages/services/src/kernel/restapi.ts
+++ b/packages/services/src/kernel/restapi.ts
@@ -250,9 +250,14 @@ export class KernelAPIClient {
    * @param options - The options used to create the client.
    */
   constructor(options: { serverSettings?: ServerConnection.ISettings } = {}) {
-    this._serverSettings =
+    this.serverSettings =
       options.serverSettings ?? ServerConnection.makeSettings();
   }
+
+  /**
+   * The server settings for the client.
+   */
+  readonly serverSettings: ServerConnection.ISettings;
 
   /**
    * List the running kernels.
@@ -265,7 +270,7 @@ export class KernelAPIClient {
    * The promise is fulfilled on a valid response and rejected otherwise.
    */
   async listRunning(): Promise<IModel[]> {
-    return listRunning(this._serverSettings);
+    return listRunning(this.serverSettings);
   }
 
   /**
@@ -282,7 +287,7 @@ export class KernelAPIClient {
    * If the kernel does not exist on the server, the promise is resolved with `undefined`.
    */
   async getKernelModel(id: string): Promise<IModel | undefined> {
-    return getKernelModel(id, this._serverSettings);
+    return getKernelModel(id, this.serverSettings);
   }
 
   /**
@@ -298,7 +303,7 @@ export class KernelAPIClient {
    * The promise is fulfilled on a valid response and rejected otherwise.
    */
   async startNew(options: IKernelOptions = {}): Promise<IModel> {
-    return startNew(options, this._serverSettings);
+    return startNew(options, this.serverSettings);
   }
 
   /**
@@ -314,7 +319,7 @@ export class KernelAPIClient {
    * The promise is fulfilled on a valid response and rejected otherwise.
    */
   async restartKernel(id: string): Promise<void> {
-    return restartKernel(id, this._serverSettings);
+    return restartKernel(id, this.serverSettings);
   }
 
   /**
@@ -330,7 +335,7 @@ export class KernelAPIClient {
    * The promise is fulfilled on a valid response and rejected otherwise.
    */
   async interruptKernel(id: string): Promise<void> {
-    return interruptKernel(id, this._serverSettings);
+    return interruptKernel(id, this.serverSettings);
   }
 
   /**
@@ -346,8 +351,6 @@ export class KernelAPIClient {
    * The promise is fulfilled on a valid response and rejected otherwise.
    */
   async shutdownKernel(id: string): Promise<void> {
-    return shutdownKernel(id, this._serverSettings);
+    return shutdownKernel(id, this.serverSettings);
   }
-
-  private _serverSettings: ServerConnection.ISettings;
 }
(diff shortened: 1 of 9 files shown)


jtpio:
Opening this PR early and leaving it as a draft for now, to gather feedback and hear what other people think.

Also adding to the `4.4.0` milestone since it's a follow-up to https://github.com/jupyterlab/jupyterlab/pull/16794, so we have it on the radar. Happy to discuss here or during the next Jupyter Frontends meeting if that's a change we want to consider for that milestone.

jtpio:
Cool, I wasn't sure if that was a change we would be ok to have in 4.4.0. But if other folks are fine with it, happy to have another look today to make sure it's in a state where it can be reviewed and merged.

jtpio:
Did another pass it and CI seems to be passing.

> If we want to be cautious, we could also mark the new options as experimental for now.

Currently the new options are added as optional already, so for now they are not marked as experimental. We could always mark them as such later before 4.4.0b1, if there is an argument in favor.

packages/services/src/kernel/restapi.ts
```diff
@@ -235,3 +235,122 @@ export async function getKernelModel(
...
+   *
+   * @returns A promise that resolves with the kernel model.
+   *
+   * #### Notes
+   * Uses the [Jupyter Server API](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter-server/jupyter_server/main/jupyter_server/services/api/api.yaml#!/kernels) and validates the response model.
+   *
+   * The promise is fulfilled on a valid response and rejected otherwise.
+   * If the kernel does not exist on the server, the promise is resolved with `undefined`.
+   */
+  async getKernelModel(id: string): Promise<IModel | undefined> {
```
krassowski, line 289: Not an opinioon yet, just a question, have you thought whether these names should be like `getModel` and `restart` over `getKernelModel` and `restartKernel`?

jtpio: Good point.

There is also `startNew` which does not have `kernel` in its name.

I guess it would make sense to drop the `kernel` part since the methods are already under the kernel client.

jtpio: For now these were mostly mappings to the existing functions already accessible from `KernelAPI`. There could also be a point of having the same function/method names when using `KernelAPI` and `KernelClientAPI`.

jtpio: Happy to push a new commit with renamed methods, so we can decided which way would be preferable.

krassowski (approved):
Thank you @jtpio!

jtpio:
To have a consistent approach for the services making use of an internal `restapi.ts` module, we could actually also follow the same pattern for the terminals and kernelspecs:

- https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/src/terminal/restapi.ts
- https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/src/kernelspec/restapi.ts

I'll open a follow-up PR to address this.

PR #17350, opened 2025-02-28
Add widget ID arg to semantic command invocations

### LINKED ISSUE #16664: Restart Kernel and Run All Cells is applied on different notebooks if tabs switch in between
## Description

If you have multiple notebooks open and click `Restart Kernel and Run All Cells` for one notebook and switch to another notebook then the kernel restart and run all cells are applied to different notebooks. This must be caused by kernel restart being async and run all cells executing as separated event and recalculating the active tab / notebook. This issue is probably not limited to restart and rerun all, but more combined events that act on active tab.

## Reproduce

It is more easily reproduced if network is slow or kernel restart takes longer.

jlab restart rerun issue

## Expected behavior

Action should be applied to the original notebook / tab

## Context

Tested with JupyterLab version: 4.2.4

afshin:
This PR changes the invocation of semantic commands to include an `args` object that has a key `SemanticCommand.WIDGET` which contains the widget ID of the widget that was current when the command was *initially* invoked.

This fixes cases where command execution modifies the wrong widget because it is slow and/or multiple commands are chained together (e.g., restart and then run all) and the current widget has been changed by user action.

## References
Fixes #16664 

## Code changes
Adds an `args` object to commands executed as part of a semantic command invocation. Since there were no `args` before, there is no risk of clobbering the old `args`. The `args` key that stores the widget ID is `SemanticCommand.WIDGET`.

## User-facing changes
In consoles and notebooks, if a command is invoked and the user rapidly switches to another notebook (or console), the command still executes on the correct console/notebook.

N/A


afshin:
The visual regression failure appears to be an updated style in the completer output and unrelated to this PR.

afshin:
bot please update snapshots

krassowski:
Hmm, we lost syntax highlighting in Signature line? That would probably be down to a change in IPython.

| Actual | Expected |
|---|--|
| completer-with-doc-panel-actual | completer-with-doc-panel-expected |

packages/apputils/src/semanticCommand.ts
```diff
@@ -37,6 +37,11 @@ export class SemanticCommand {
    */
   static readonly DEFAULT_RANK = 500;
 
+  /**
+   * The `args` key for a semantic command's current widget ID.
+   */
+  static readonly WIDGET = 'semantic-widget';
```
krassowski, line 43: This is a nitpick, but for "special" args it looks like we usually follow camelCase:
- `_luminoEvent` (https://github.com/jupyterlab/lumino/pull/644)
- `isPalette` https://github.com/jupyterlab/jupyterlab/pull/17258

```suggestion
  static readonly WIDGET = 'semanticWidget';
```

krassowski (approved):
Thank you @afshin! I left a small suggestion on the name of the `args` key.

PR #17352, opened 2025-02-28
Add a setting to disable the context menu

### LINKED ISSUE #14682: Option to disable context menu per user, without admin permissions
### Problem

As discussed in #13553 the context menu lacks copy and paste of selected text. For some users, access to copy and paste is more important and used more often than any feature in the custom context menu. 

### Proposed Solution

An option to completely turn off the context menu, returning right click to the browser's context menu only and bringing back copy and paste of selected text via right clicking.

afshin:
This PR adds a user setting to disable the context menu across JupyterLab at runtime without disabling any extensions.

Like other user settings, it can be configured by the user or in [`overrides.json`](https://jupyterlab.readthedocs.io/en/latest/user/directories.html#overridesjson).

However, because it is a setting that is part of the context menu plugin, it is skipped in the visual settings editor:

https://github.com/jupyterlab/jupyterlab/blob/a09f4b1025a75cccd10e440e038393041a2424a3/packages/settingeditor-extension/src/index.ts#L153

This PR adds a toggleable command to the command palette to allow users to enable/disable the context menu. And of course, it can also be modified in the JSON settings editor.

## References
Fixes #14682 

## Code changes
This PR adds a boolean setting (called `disabled`) to the context menu plugin (`@jupyterlab/application-extension:context-menu`) to check whether to suppress the context menu should be disabled.

If a user disables the context menu, the `data-jp-suppress-context-menu` attribute is added to the document `<body>` element, which allows the native context menu to come through instead of the JupyterLab context menu.

## User-facing changes
- There is a new setting that can be modified in the JSON settings editor to disable the JupyterLab context menu.
- There is a command in the command palette to toggle the context menu `disabled` setting.

https://github.com/user-attachments/assets/bbf7155a-1734-429f-b95c-31b5cc76f762

N/A


packages/application-extension/schema/context-menu-suppressor.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "Context Menu Suppression",
+  "description": "Context Menu Suppression Settings.",
+  "jupyter.lab.menus": {},
+  "jupyter.lab.shortcuts": [],
+  "properties": {
+    "disabled": {
+      "title": "Whether the context menu is suppressed",
```
krassowski, line 8: I guess we should clarify if we suppress browser or application (jupyterlab/notebook) menu?

```suggestion
      "title": "Whether the application context menu is suppressed",
```

krassowski (commented):
I think that's a great option to have. Could we avoid adding more one-setting sections in the Settings Editor?

galata/test/jupyterlab/completer.test.ts-snapshots/completer-with-doc-panel-jupyterlab-linux.png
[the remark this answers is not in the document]
afshin: You're right. I just wanted to confirm that this PR is ✅  ... although I wondered if the merging would work because I took these snapshots from that branch but I don't know how git handles this. It should definitely be standalone 👍

packages/application-extension/schema/context-menu-suppressor.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "Context Menu Suppression",
+  "description": "Context Menu Suppression Settings.",
+  "jupyter.lab.menus": {},
+  "jupyter.lab.shortcuts": [],
+  "properties": {
+    "disabled": {
```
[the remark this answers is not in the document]
afshin: Possibly, I debated this too. The `suppress` verb was already in the code, but not in the UI, so maybe you're right about this.

afshin:
bot please update snapshots

packages/application-extension/src/index.tsx
```diff
@@ -1502,6 +1518,16 @@ namespace Private {
         }
       }
     });
+
+    // Handle disabled status.
+    setDisabled(settings);
+    commands.addCommand(CommandIDs.toggleContextMenu, {
+      label: trans.__('Enable Context (right-click) Menu'),
```
krassowski, line 1525: I'm not sure if this should be in the main Settings menu - is this something that users would toggle frequently enough to justify it?

I would say it should not, but if it does need to be in there, I would suggest to make the label shorter, also because depending on a device it will not be a right-click, see https://github.com/jupyterlab/jupyterlab/issues/14682#issuecomment-2351836843.

In any case, I think it could be added to the command palette.

```suggestion
      label: trans.__('Enable Context Menu'),
      caption: trans.__('Enable context (right-click) menu'),
```

afshin: "Context menu" is a term of art that menu users might not know, that's why I added `right-click`, but I do agree it's clunky and I don't have strong feelings about it.

The reason I added it to the `Settings` menu is because it's not at all discoverable as a setting otherwise, because the context menu plugin is specifically omitted in the visual settings editor.

But actually maybe that's fine because this PR can add the feature and another follow-on PR can make it more discoverable

afshin: I will remove the command from the `Settings` menu and maybe add it to the command palette.

I will also remove the `right-click` language and another PR can deal with verbiage and discoverability if we find that there's a problem.

krassowski: > because the context menu plugin is specifically omitted in the visual settings editor

I think this is the problem we should address. I think it should not be omitted. I think we should discuss it separately, but my memory is that:
- setting editor UI was introduced as a huge list rendered all at once with accordion panels
- it was slow, so some plugins with lot's of settings were excluded which improved the performance
- it was then rewritten to only show one plugin at a time
- the plugins which were previously omitted for performance reasons were not re-enabled

So maybe these should be excluded from search as when searching "c" too much would render, but I do not think we should hide them.

krassowski: I opened https://github.com/jupyterlab/jupyterlab/issues/17362 to track it.

afshin:
bot please update snapshots

krassowski (approved):
Thank you @afshin!

Issue #13116, opened 2022-09-22
Request for Accuracy: Rename "Delete" to "Move to Trash"


nicobasile:
## Description

In the Jupyterhub UI, the user has the option to "Delete" the file the have selected.
However, the actual behavior of "Delete" is to move the file to $HOME/.local/share/files/Trash

This caused some problems for my team when we wondered why storage space wasn't being cleared after "Delete"ing a file. Renaming this option to "Move to Trash" would be more accurate and mimic what MacOS/Windows names this identical feature.

![Screen Shot 2022-09-22 at 4 38 54 PM](https://user-images.githubusercontent.com/53441157/191869414-215edc2c-e967-4157-8dce-eb4d87b642e3.png)

## Expected behaviour
We expected "Delete" to delete and clear the storage space that the file once occupied.

## Actual behaviour
"Delete" moves the file to a different location without actually deleting it.

krassowski:
Thank you for opening this issue. This is a good observation, but it needs a more involved fix:
- JupyterLab uses [`/api/contents/delete`](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#delete--api-contents-path) endpoint of the contents API offered by jupyter-server, so on this level the UI is correct (it shows what the application requests).
- jupyter-server has a [configuration option](https://jupyter-server.readthedocs.io/en/latest/other/full-config.html?highlight=delete_to_trash#options) `FileContentsManager.delete_to_trash` and `AsyncFileContentsManager.delete_to_trash` which defaults to `True` for 6 years now (https://github.com/jupyter-server/jupyter_server/commit/ad3ab9b8c7c107629714db0de4dd2a40350eb4e8); if this behaviour is not expected for you, you can change the configuration
- to make the UI accurate, we would need to know what is the server setting; simply switching from "delete" to "move to trash" would lead to even worse consequence if administrator of the jupyter-server configured it to permanently delete files and users believed it would only be moved to trash, so while the current label is often wrong, it is on the safe side
   - alternatively, we could consider making jupyter-server `/api/contents/delete` request include optional `moveToTrash` argument and control this from the frontend, but it is not clear to me if that would not cause some trouble for deployments which assume that users are not permitted to make this kind of change

jesuino:
Hello,

[...]

PR #17359, opened 2025-03-03
Rename "Delete" to "Move to Trash"

jesuino:
## References

This is a fix for https://github.com/jupyterlab/jupyterlab/issues/13116

## Code changes

* Exposed the configuration using the `PageConfig` object
* Read the config `delete_to_trash` from PageConfig when constructing the context menu
* Change the menu label accordingly

## User-facing changes

Users will see the menu `Move to Trash` instead of `Delete` when the configuration `delete_to_trash` is true (Which is the default value)

image

No changes to APIs.

----

Please let me know if I should change anything else. Thanks!

### DIFF
--- a/jupyterlab/labapp.py
+++ b/jupyterlab/labapp.py
@@ -736,6 +736,8 @@ class LabApp(NotebookConfigShimMixin, LabServerApp):
         page_config["exposeAppInBrowser"] = self.expose_app_in_browser
         page_config["quitButton"] = self.serverapp.quit_button
         page_config["allow_hidden_files"] = self.serverapp.contents_manager.allow_hidden
+        if hasattr(self.serverapp.contents_manager, "delete_to_trash"):
+            page_config["delete_to_trash"] = self.serverapp.contents_manager.delete_to_trash
 
         # Client-side code assumes notebookVersion is a JSON-encoded string
         page_config["notebookVersion"] = json.dumps(jpserver_version_info)
--- a/packages/filebrowser-extension/src/index.ts
+++ b/packages/filebrowser-extension/src/index.ts
@@ -917,6 +917,7 @@ function addCommands(
   const trans = translator.load('jupyterlab');
   const { docRegistry: registry, commands } = app;
   const { tracker } = factory;
+  const deleteToTrash = PageConfig.getOption('delete_to_trash') === 'true';
 
   commands.addCommand(CommandIDs.del, {
     execute: () => {
@@ -927,7 +928,7 @@ function addCommands(
       }
     },
     icon: closeIcon.bindprops({ stylesheet: 'menuItem' }),
-    label: trans.__('Delete'),
+    label: deleteToTrash ? trans.__('Move to Trash') : trans.__('Delete'),
     mnemonic: 0
   });
 


jiridanek:
Hi, great work! I still found one inconsistncy there, though.

When I click the new Move to Trash action

image

i then get a modal window that talks about permanent deletion (and not about moving to trash)

image

jesuino:
Hello @jiridanek Thanks for your comment. I included change to these messages as well. 

I wonder if moving the flag down is the way to go, I could read `PageConfig` on `listing.ts` and modify the text from there. Let's wait for reviewers feedback.

I am checking how to update the documentation screenshots to fix the failing tests, any suggestion is welcome!

jesuino:
please update snapshots

packages/filebrowser/src/listing.ts
```diff
@@ -425,30 +425,32 @@ export class DirListing extends Widget {
...
         ? this._trans.__(
-            'Are you sure you want to permanently delete: %1?',
+            `Are you sure you want to ${actionMessage}: %1?`,
             items[0].name
           )
         : this._trans._n(
-            'Are you sure you want to permanently delete the %1 selected item?',
-            'Are you sure you want to permanently delete the %1 selected items?',
+            `Are you sure you want to ${actionMessage} the %1 selected item?`,
+            `Are you sure you want to ${actionMessage} the %1 selected items?`,
```
krassowski, line 445: These will not work correctly as translations strings are extracted via static analysis, otherwise we would have done this with "item"/"items" too ;) This needs to be split up into separate `trans.__()` calls for each case.

krassowski (commented):
Half jokingly, on topic of translations, I wonder if at some point we should add British English language pack and replace the harsh "trash" with much neater "bin" ;)

krassowski:
bot please update snapshots

krassowski:

krassowski:
I suspect there is something wrong with the documentation update job, neither of these updates snapshots:

- https://github.com/jupyterlab/jupyterlab/actions/runs/13813319366/job/38639995458
- https://github.com/jupyterlab/jupyterlab/actions/runs/13810673445/job/38631334968

krassowski:
Some notes from frontents call - we may want to:
- have this per-drive (or per contents manager)
- have a way to force delete
- have a way to override label on deployments (i.e. page config is good because it can be overridden in principle)
- discuss it tomorrow on jupyter-server call

No blockers on the PR identified however.

packages/filebrowser/src/listing.ts
```diff
@@ -425,30 +425,47 @@ export class DirListing extends Widget {
    *
    * @returns A promise that resolves when the operation is complete.
    */
-  async delete(): Promise<void> {
+  async delete(deleteToTrash: boolean = false): Promise<void> {
```
krassowski, line 428: Instead of adding a new argument here, could we duplicate the `const deleteToTrash = PageConfig.getOption('delete_to_trash') === 'true';` here?

krassowski: Did it in [8ad22be](https://github.com/jupyterlab/jupyterlab/pull/17359/commits/8ad22be99de659b90fabe814aaf030d80c3eb9cc).

krassowski (approved):
Thank you @jesuino!

packages/filebrowser/src/listing.ts
jiridanek: > Duplicate page config check to reduce public API changes
that makes sense, I wondered what's the motivation, but I completely forget that plugins may call this

krassowski: So motivation is two-fold:
- if a downstream subclassed this class and in a child method added an argument then technically it would be a breaking change
- if we want to add more arguments in the future we need to think about carefully as removing them is hard

Basically given the feedback on lab call it seemed likely that we will iterate on the delete behaviour in the future so I want to keep this PR as lightweight on the API choices as possible (and merge it ASAP as we are coming to a freeze on 4.4.0).

jesuino: I think calling `PageConfig` from `listings.ts` looks better, thanks @krassowski

PR #17361, opened 2025-03-04
Move the licenses plugin to `apputils-extension`

jtpio:
## References

Currently the licenses plugin is solely defined in the `help-extension`:

https://github.com/jupyterlab/jupyterlab/blob/a09f4b1025a75cccd10e440e038393041a2424a3/packages/help-extension/src/index.tsx#L474-L647

Its viewer and model are defined in a module as part of the `help-extension` package here: https://github.com/jupyterlab/jupyterlab/blob/a09f4b1025a75cccd10e440e038393041a2424a3/packages/help-extension/src/licenses.tsx#L25

This makes it difficult for an extension to swap some parts of the licenses plugin, for example to be able to fetch the licenses from another location.

## Code changes

- [x] Move the licenses plugin to the `apputils-extension`
- [x] Provide a `ILicensesClient` so an extension could more easily swap the part that makes the requests to fetch the licenses.
- [x] Add to the migration guide

## User-facing changes

None

This ends up renaming `@jupyterlab/help-extension:licenses` to `@jupyterlab/apputils-extension:licenses-plugin`, which can be documented in the migration guide, and also handled by the mapping introduced in https://github.com/jupyterlab/jupyterlab/pull/17338.


docs/source/extension/extension_migration.rst
```diff
@@ -102,6 +102,9 @@ Plugins
 
 - ``@jupyterlab/application-extension`` from 4.3 to 4.4
    * The ``@jupyterlab/application:mimedocument`` plugin id has been renamed to ``@jupyterlab/application-extension:mimedocument``
+- ``@jupyterlab/help-extension`` from 4.3 to 4.4
+   * The ``@jupyterlab/help-extension:licences`` plugin id has been moved to the ``@jupyterlab/apputils-extension`` extension,
+     and is now named ``@jupyterlab/apputils-extension:licenses-plugin``
```
krassowski, line 107:
```suggestion
     and is now split between ``@jupyterlab/apputils-extension:licenses-plugin`` and ``@jupyterlab/apputils-extension:licenses-tracker``
```

packages/apputils-extension/src/licensesplugin.ts
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * The command IDs used by the licenses plugin.
+ */
+namespace CommandIDs {
+  export const licenses = 'apputils:licenses';
+
+  export const licenseReport = 'apputils:license-report';
+
+  export const refreshLicenses = 'apputils:licenses-refresh';
```
krassowski, line 37: Renaming of these commands **is** a mildly breaking change. Custom shortcuts for these will stop working and extensions which used these commands will no longer work. Now, I do not see a very good reason for users to have shortcuts for these or for extensions to use them so it is a moot point, but if we wanted to be pedantic about version semantics we would probably want to still add back the old `help:` prefixed commands and emit deprecation warning in console or similar.

packages/apputils-extension/src/licensesplugin.ts
```diff
@@ -0,0 +1,241 @@
...
+};
+
+/**
+ * A plugin to add a licenses reporting tools.
+ */
+export const licensesPlugin: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/apputils-extension:licenses-plugin',
+  description: 'Adds licenses reporting tools.',
+  requires: [ITranslator],
+  optional: [ILicensesTracker, IMainMenu, ICommandPalette],
```
krassowski, line 167: It looks to me `licensesTracker` should be a required one, as this plugin does do anything without it?

krassowski (commented):
>  Provide a `ILicensesTracker` so an extension could more easily swap the part that instantiates new licences widgets.

Why would an extension want to do that? Usually I think the trackers are useful to react to things being added, but I also do not see a good use case here. Probably I am missing something.

krassowski:
I guess it is good to loop-in @bollwyvl who wrote a large the licences plugin and might be using these commands?

jtpio:
> Why would an extension want to do that? Usually I think the trackers are useful to react to things being added, but I also do not see a good use case here. Probably I am missing something.

One use case is in JupyterLite: https://github.com/jupyterlite/jupyterlite/pull/1590. Without such change, it's not possible to reuse most of the upstream JupyterLab plugins, and leads to copy pasting everything (including the model) just to override the network part.

Exposing the tracker was the approach which led to less changes compared to the current logic and implementation. but maybe we could look into the `DataConnector` approach instead, similar to the settings and translations plugins.

packages/apputils-extension/src/licensesplugin.ts
jtpio: Yes, keeping the previous `help:` commands sounds good, happy to add them back.

jtpio:
> Exposing the tracker was the approach which led to less changes compared to the current logic and implementation. but maybe we could look into the `DataConnector` approach instead, similar to the settings and translations plugins.

Since the `Licenses` model is currently not exported and part of the API surface, maybe it could be changed more deeply, since this is where the requests to the server are made:

https://github.com/jupyterlab/jupyterlab/blob/5d82254da33b330b195350f98fbf3b0222a6c4b5/packages/help-extension/src/licenses.tsx#L286

packages/apputils-extension/src/licensesplugin.ts
jtpio: This raises the question of whether we consider the command ids and their args to be part of the public API.

Since we have the list of commands defined in https://github.com/jupyterlab/jupyterlab/blob/main/galata/test/documentation/commands.test.ts-snapshots/commandsList-documentation-linux.json, we could indeed better keep track of commands that have been renamed or changed.

Also in its current state, this PR does not modify `commandsList-documentation-linux.json` while it should, since it changed the commands. Maybe the UI tests should fail when the list of commands changes and that file has not been updated.

krassowski: > Maybe the UI tests should fail when the list of commands changes and that file has not been updated.

It would make sense to me, otherwise all these things get updated in an unrelated PR making the diff a bit confusing.

krassowski:
The `DataConnector` approach sounds right here as otherwise you will end up duplicating the rest of the plugin code (tracker etc) anyways.

bollwyvl:
Hm, not sure I see the carrot here _aside_ from the network access piece, which could be done without changing the API by passing in the new `ServerManager` plugin

About the only time I've used this (other than interactively) is for automating the validation of an as-installed extension and version under test... but it really is the only way to know about certain components without speaking raw webpack.

> fetch the licenses from another location.

It uses `PageConfig`, as defined in [`LabConfig`](https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L286) so swapping it would likely _also_ require a server extension, at which point they could change the value. 

The only thing I'd really see really wanting to get in here is beefing up the data model to be full SPDX, and supporting the browser _and_ server and kernel environments in a single, standards-based document.

> consider the command ids and their args to be part of the public API.

yes, if _anything_ should be stable over time, it should be commands, to the extent possible. In this case, if that meant leaving a wrapper command under `help` which just called the new command under `apputils`, that's great, again, if possible, with some console-observable warning that it was deprecated, with a schedule for when.

jtpio:
To give more context, these two methods are the ones triggering the requests to the server:

https://github.com/jupyterlab/jupyterlab/blob/bc3ed565762a4ba5c7fd8d650c1f17a8d48c265a/packages/help-extension/src/licenses.tsx#L304-L335

As long as there is a way to change that downstream to fit the JupyterLite use case, that would be enough for now.

bollwyvl:
Would that not Just Work if it was using the `IServiceManager.makeRequest`?

jtpio:
> Would that not Just Work if it was using the `IServiceManager.makeRequest`?

Hmm that one does not seem to exist on `IServiceManager`.

There is `ServerConnection.makeRequest()` but it's coming from a namespace:

https://github.com/jupyterlab/jupyterlab/blob/bc3ed565762a4ba5c7fd8d650c1f17a8d48c265a/packages/services/src/serverconnection.ts#L144-L150

packages/apputils-extension/src/licensesplugin.ts
jtpio: Then we can track that in an issue. And maybe this could also apply to the list of tokens and plugins generated for the docs, to make sure the list is updated when new tokens or plugins are added.

jtpio:
The UI tests seemed to be failing for a while, but they are now passing. Undrafting.

packages/apputils-extension/src/licensesplugin.ts
jtpio: Opened https://github.com/jupyterlab/jupyterlab/issues/17393

krassowski (approved):
Thank you @jtpio!

PR #17363, opened 2025-03-04
If subshells are supported by the kernel, send comm messages to subshells

martinRenou:
## References

This builds on top of the work done in both ipykernel and JupyterLab for subshells support:
- https://github.com/ipython/ipykernel/pull/1249
- https://github.com/jupyterlab/jupyterlab/pull/16963

## Code changes

The code changes are only on `Comm` message handling. We're now sending all comm messages on a subshell (one subshell per-comm). There is no public API change.

Subshells are only used if the kernel supports it.

## User-facing changes

This improves a lot the ipywidgets usage, allowing to execute callbacks even if a blocking execute request is being processed.

### BEFORE THIS PR:

https://github.com/user-attachments/assets/55d1bdb2-36e9-4ebe-bb83-e13194a1da98

### AFTER THIS PR:

https://github.com/user-attachments/assets/393850e6-8fd8-48c7-9448-e4ed23728cc2

None


packages/services/src/kernel/comm.ts
```diff
@@ -212,6 +214,31 @@ export class CommHandler extends DisposableDelegate implements Kernel.IComm {
...
   }
 
+  dispose(): void {
+    super.dispose();
+
+    this._maybeCloseSubshell();
+  }
+
+  private async _maybeStartSubshell() {
+    if (this._kernel.supportsSubshells) {
```
ianthomas23, line 224: There should be an `await this._kernel.info` before this, as we can only tell if a kernel supports subshells after we've received the kernel info which contains the appropriate information.

krassowski:
Exciting! I am not familiar with comms very much, is using subshells by default (if ipykernel 7 or another kernel supporting them is used) compatible with all comms consumers, or will there be any changes required for these to work?

martinRenou:
> Exciting! I am not familiar with comms very much, is using subshells by default (if ipykernel 7 or another kernel supporting them is used) compatible with all comms consumers, or will there be any changes required for these to work?

It will continue working as before and use the "main shell" if sub-shells are not supported, since I'm checking `if (this._kernel.supportsSubshells)`. So for now only ipykernel 7.0.0a0 would see a change in behavior.

krassowski:
Yes, I understand that. I do not understand if the other side of the comm, like ipywidgets required any modifications, or if it "just" works with subshells.

martinRenou:
> Yes, I understand that. I do not understand if the other side of the comm, like ipywidgets required any modifications, or if it "just" works with subshells.

Gotcha, yes, it just works without any change in ipywidgets! The subshell is created from the front-end, and the kernel will dispatch comm messages to the right subshells automatically. ipywidgets auto-magically executes the comm message callbacks in the proper shell.

martinRenou:
Putting this on draft for now. We should battle test since. Creating one subshell per-comm may not be the smartest move since I've seen users create thousands of widgets.

martinRenou:
Wondering if we should create a subshell per comm-target instead. Meaning, one subshell for all ipywidgets.

martinRenou:
I make that functionality disabled by default, and allowing to enable it through settings:


https://github.com/user-attachments/assets/91f22423-f2a0-49a2-9fe7-5b9c61a9d25e

packages/application-extension/schema/kernel-subshells.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "Kernel Subshells",
```
krassowski, line 2: I think we need to consolidate kernel settings in a single place. There is also "Kernel dialogs" which only has one setting.

krassowski: Also "Running Terminals and Kernels" with just one setting.

martinRenou: Should I merge them all in some "Terminals and Kernels" settings?

martinRenou: I've moved the subshells settings closer to the existing kernel settings

krassowski: > That would be a breaking change, wouldn't it be? Since user settings would stop working.

No, adding settings to existing sections is fine. Also, that particular one was not even released. I think this should block release but more necessarily this PR.

trungleduc:
> Wondering if we should create a subshell per comm-target basis instead. Meaning, one subshell for all ipywidgets.

👍 this sounds good! is there any downside of this approach?

martinRenou:
> > Wondering if we should create a subshell per comm-target basis instead. Meaning, one subshell for all ipywidgets.
> 
> 👍 this sounds good! is there any downside of this approach?

Widgets would be blocking each other. It would be a better situation than today but still not perfect.

martinRenou:
@krassowski I merged the settings together. Please tell me what you think!

@ianthomas23 @trungleduc I added the option to use one subshell per-comm or per-comm-target

Screenshot From 2025-03-11 10-54-41

ianthomas23:
I think this needs at least a minimal test like the one I outlined above to check that a subshell is created and deleted correctly depending on the settings.

martinRenou:
Will look into adding a test 👍🏽 

> is the subshell closed when I clear the cell output containing the widgets

It should be closed when disposing the comm in case it's one subshell per-comm

martinRenou:
Added some tests for each case!

martinRenou:
Some tests are flaky right now. Looks unrelated to the PR.

krassowski:
> I added the option to use one subshell per-comm or per-comm-target

Would you like to make one of these the default over 'disabled'? Or is there a reason not to do that?
If there is a reason not to, would it be desirable to change the default just for the duration of the beta period?

> Some tests are flaky right now. Looks unrelated to the PR.

Yes, it looks like there was some issue on GH Actions earlier today, see https://github.com/jupyterlab/jupyterlab/issues/17381

martinRenou:
I guess defaulting to one subshell per comm target makes sense. It should improve the current state, without putting too much load on the subshell implementation which is still early.

Will do!

packages/services/src/kernel/comm.ts
```diff
@@ -212,6 +257,61 @@ export class CommHandler extends DisposableDelegate implements Kernel.IComm {
...
+
+    if (this._commsOverSubshells === CommsOverSubshells.PerComm) {
+      // Create subshell
+      const replyMsg = await this._kernel.requestCreateSubshell({}).done;
+      this._subshellId = replyMsg.content.subshell_id;
+      this._subshellStarted.resolve();
+      return;
+    }
+
+    // One shell per comm-target
```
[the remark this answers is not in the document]
ianthomas23: I suppose we could store some sort of `pendingRequests` information and not repeat a request if it is already pending but instead wait for that request to be resolved.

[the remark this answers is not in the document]
fleming79: I'm very excited to see this new feature emerging!

I gained some experience related to this topic by writing a draft [PR](https://github.com/jupyter-widgets/ipywidgets/pull/3922) for ipywidgets. The motivation was around enabling the usage of Ipywidgets outside of Notebooks and Consoles for [a modified version of Ipylab ](https://github.com/fleming79/ipylab). 

One issue I noticed in writing that PR was related to how comm messages are managed by `KernelConnection`:

*Extract from KernelConnection in Jupyterlab*
```
  /**
   * Handle comm messages
   *
   * #### Notes
   * The comm message protocol currently has implicit assumptions that only
   * one kernel connection is handling comm messages. This option allows a
   * kernel connection to opt out of handling comms.
   *
   * See https://github.com/jupyter/jupyter_client/issues/263
   */
  readonly handleComms: boolean;
```

Although it says it is an 'opt out' option, what it really means is that the life of the comms is tied to the `KernelConnection` of whichever connection is made first, generally a notebook or console session.  In the PR mentioned above, comms is re-establishing when the original `SessionConnection` is closed.

I think you can keep this new feature simple but powerful,  I offer the following suggestions:
1. Modify KernelConnction so comms is de-coupled from `KernelConnection` (but still respect `handleComms===false`). The life of comms is tied to the life of the `Kernel` rather than the `KernelConnection`.
2.  Provide a separate subshell per-comm-target per-kernel (rather than per `KernelConnection`). Both the comm-target and kernel ID can be known in advance to ensure the correct number of subshells are created.
3. Drop: *one subshell per comm*.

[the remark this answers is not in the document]
ianthomas23: > I suppose we could store some sort of `pendingRequests` information and not repeat a request if it is already pending but instead wait for that request to be resolved.

On second thoughts we could just change `_commTargetSubShellsId` to store `Promise`s of subshell IDs, and create a `Promise` and add it to the collection before the first request subshell call, resolving on its reply. Then other uses of the same target could just await the Promise.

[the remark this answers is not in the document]
martinRenou: Thanks for your comment @fleming79 !

> Drop: one subshell per comm

Other opinions on this? I could see the benefit of allowing one subshell per-comm, even though it may get out of hand quickly.

> Provide a separate subshell per-comm-target per-kernel (rather than per KernelConnection). Both the comm-target and kernel ID can be known in advance to ensure the correct number of subshells are created.

Since the subshell ID per-target is kept as a static property of `CommHandler` now, that shouldnt be a problem I guess?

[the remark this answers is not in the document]
martinRenou: > create a Promise and add it to the collection before the first request subshell call

Good catch @ianthomas23  I've updated the PR as you suggested

krassowski:
Picked up a conflict in `packages/services/src/kernel/default.ts`

fleming79:
Are there any thread safe issues that need consideration (kernel side)?

Would one subshell per kernel be adequate for most use cases?

packages/services/src/kernel/comm.ts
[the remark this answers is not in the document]
fleming79: In ipywidgets both `layout` and `style` are also widgets. Meaning possibly 3 subshells for a single widget.

fleming79:
From an `asyncio` perspective I think each `subshell` will have its own event loop. 

For example.

``` python
# in notebook

import asyncio
loop = asyncio.get_running_loop()
id(loop)
```

``` python
# in subshell console for the notebook kernel

async def test():
    print("hello")

await = loop.create_task(test())
# Raises RuntimeError: attached to a different loop
```

packages/services/src/kernel/comm.ts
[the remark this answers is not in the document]
martinRenou: Yep 👍🏽 Meaning one should only use one subshell per-comm when using small number of widgets. From our testing it seems less than a hundred of widgets works well.

martinRenou:
> Picked up a conflict in packages/services/src/kernel/default.ts

Rebased and fixed! Thanks

> Would one subshell per kernel be adequate for most use cases?

We're close to this when using one subshell per comm target (there aren't many libraries out there using comms). And I made that the default in the PR.

ianthomas23:
Using one subshell per comm target (2 subshells overall) here are 400 linked sliders:

https://github.com/user-attachments/assets/2897a728-e42f-4e55-8b24-d632df2565c9

500 linked sliders doesn't work for me with or without subshells so that is some unrelated limit.

Although one subshell per comm may not be a good solution long term, I would like to keep it available now as it is a good way to stress the subshell implementation to find problems.

packages/services/src/kernel/default.ts
```diff
@@ -82,6 +84,29 @@ export class KernelConnection implements Kernel.IKernelConnection {
...
 
+  /**
+   * Whether comm messages should be sent to kernel subshells, if the
+   * kernel supports it.
+   *
+   * #### Notes
+   * Sending comm messages over subshells allows processing comms whilst
+   * processing execute-request on the "main shell". This prevents blocking
+   * comm processing.
+   * If enabled, we'll create one subshell per-comm, this may lead to issue
```
ianthomas23, line 95: This text can be updated to also mention one subshell per comm target, and that that is the default.

packages/services/src/kernel/kernel.ts
```diff
@@ -108,6 +108,19 @@ export interface IKernelConnection extends IObservableDisposable {
...
 
+  /**
+   * Whether comm messages should be sent to kernel subshells, if the
+   * kernel supports it.
+   *
+   * #### Notes
+   * Sending comm messages over subshells allows processing comms whilst
+   * processing execute-request on the "main shell". This prevents blocking
+   * comm processing.
+   * If enabled, we'll create one subshell per-comm, this may lead to issue
```
ianthomas23, line 119: Same here.

packages/services/src/kernel/kernel.ts
```diff
@@ -646,6 +659,19 @@ export namespace IKernelConnection {
...
 
+    /**
+     * Whether comm messages should be sent to kernel subshells, if the
+     * kernel supports it.
+     *
+     * #### Notes
+     * Sending comm messages over subshells allows processing comms whilst
+     * processing execute-request on the "main shell". This prevents blocking
+     * comm processing.
+     * If enabled, we'll create one subshell per-comm, this may lead to issue
```
ianthomas23, line 670: Same here.

packages/services/src/kernel/kernel.ts
```diff
@@ -699,6 +725,19 @@ export interface IManager extends IBaseManager {
...
 
+  /**
+   * Whether comm messages should be sent to kernel subshells, if the
+   * kernel supports it.
+   *
+   * #### Notes
+   * Sending comm messages over subshells allows processing comms whilst
+   * processing execute-request on the "main shell". This prevents blocking
+   * comm processing.
+   * If enabled, we'll create one subshell per-comm, this may lead to issue
```
ianthomas23, line 736: Same here.

ianthomas23 (approved):
The use of subshells looks good to me. Approving with just a nit about some comments.

Thanks @martinRenou!

trungleduc:
I think it's useful to have a short explanation about each option of the comm subshell setting. So that users know the implications of their choice.

packages/apputils-extension/schema/sessions-settings.json
```diff
@@ -0,0 +1,25 @@
...
+  "title": "Terminals and Kernels sessions settings",
+  "description": "Terminals and Kernels sessions settings",
+  "additionalProperties": false,
+  "properties": {
+    "showStatusBarItem": {
+      "type": "boolean",
+      "title": "Show the status bar item",
+      "description": "Whether to show the running terminals and kernels item in the status bar",
+      "default": true
+    },
```
krassowski, line 11: Maybe this should be just "Kernel settings" and later we can also fold this orphan in there:

image


My reasoning is:
- the terminals are not shown in status bar unless any is open
- we can rework the behaviour of `showStatusBarItem` to only affect kernels
- we could add another `showStatusBarItem` for terminal specifically, as we already have "Terminal" settings section:

image

Basically I would like to decouple these so that we do not have "terminals and kernels" section in settings just because the "running sessions" encompasses both in code.

CC @afshin @jtpio?

krassowski (commented):
Can you add a paragraph about user implications in user-facing changelog under existing subshells section?

packages/apputils-extension/schema/sessions-settings.json
krassowski: > I would like to decouple these so that we do not have "terminals and kernels" section in settings just because the "running sessions" encompasses both in code.

To make it clear, I do not suggest making this change in this PR. I am only asking if this is the direction we should take (so in this PR we would just rename this settings section to "Kernels", and possibly the file to `kernel.json`).

krassowski:
It picked up a conflict in `kernel/manager.ts` from https://github.com/jupyterlab/jupyterlab/pull/17395.

If we want to include this PR in 4.4.0 we need to finalize it this week and ideally release another beta before Wednesday. Do you all think this is feasible?

martinRenou:
Since the latest review comments are about docstrings, it should be feasible. I will put some time on this tomorrow morning.

martinRenou:
Done!

packages/apputils-extension/schema/sessions-settings.json
krassowski: I opened https://github.com/jupyterlab/jupyterlab/issues/17406 to track this.

CHANGELOG.md
```diff
@@ -76,6 +76,13 @@ A new console panel will open allowing access to variables from the main session
 
 <img alt="The subshell console" src="https://raw.githubusercontent.com/jupyterlab/jupyterlab/main/docs/source/getting_started/changelog_assets/4.4-subshells.png" class="jp-screenshot">
 
+Alternatively, JupyterLab allows automatically creating subshells for comm messages if kernels support this feature.
+This has the by-product of allowing to execute `ipywidgets` callbacks concurrently without requiring a user action.
+This behavior can be changed from the settings, the user can either:
```
krassowski, line 81:
```suggestion
JupyterLab can automatically create subshells for comm messages if kernels support this feature.
This enables concurrent execution of `ipywidgets` callbacks without requiring a user action.
This behavior can be changed from the settings, with options to:
```

packages/apputils-extension/src/subshell-settings.ts
```diff
@@ -0,0 +1,51 @@
...
+} from '@jupyterlab/application';
+import { CommsOverSubshells } from '@jupyterlab/services';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
+
+/*
+ * A plugin providing running terminals and sessions settings
+ */
+export const subshellsSettings: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/apputils-extension:sessions-settings',
+  description: 'Sessions and terminals settings.',
```
krassowski, line 18:
```suggestion
 * A plugin loading kernel subshell settings.
 */
export const subshellsSettings: JupyterFrontEndPlugin<void> = {
  id: '@jupyterlab/apputils-extension:subshell-settings',
  description: 'Kernel subshell settings.',
```

packages/services/src/kernel/kernel.ts
```diff
@@ -707,6 +737,21 @@ export interface IManager extends IBaseManager {
...
+   * #### Notes
+   * Sending comm messages over subshells allows processing comms whilst
+   * processing execute-request on the "main shell". This prevents blocking
+   * comm processing.
+   * Options are:
+   * - disabled: not using subshells
+   * - one subshell per comm-target (default)
+   * - one subshell per comm (can lead to issues if creating many comms)
+   */
+  commsOverSubshells: CommsOverSubshells;
```
krassowski, line 753: Can this be optional?


```suggestion
  commsOverSubshells?: CommsOverSubshells;
```

packages/services/src/kernel/kernel.ts
```diff
@@ -896,6 +941,21 @@ export interface IComm extends IDisposable {
...
+  /**
+   * Promise that resolves when the subshell started, if any
+   */
+  get subshellStarted(): Promise<void>;
+
+  /**
+   * Whether comms are running on subshell or not.
+   */
+  commsOverSubshells: CommsOverSubshells;
+
```
krassowski, line 958: This is a non-optional expansion of the interface; should these be optional?

packages/services/src/kernel/manager.ts
```diff
@@ -157,6 +160,18 @@ export class KernelManager extends BaseManager implements Kernel.IManager {
     return this._models.size;
   }
 
+  get commsOverSubshells(): CommsOverSubshells {
```
krassowski, line 163:
```suggestion
  /**
   * Whether comms are running on subshell or not.
   */
  get commsOverSubshells(): CommsOverSubshells {
```

krassowski (commented):
I wonder if we could make it less disrupting for any downstreams which re-implement interfaces by making the additions optional.

packages/services/src/kernel/kernel.ts
```diff
@@ -896,6 +941,21 @@ export interface IComm extends IDisposable {
...
 
+  /**
+   * The subshell ID, null if not using a subshell.
+   */
+  subshellId: string | null;
+
+  /**
+   * Promise that resolves when the subshell started, if any
+   */
+  get subshellStarted(): Promise<void>;
```
krassowski, line 952: @martinRenou to highlight, `subshellStarted` and `subshellId` are also non-optional additions.

martinRenou: yeah I'm annoyed to have added this. But this is needed for testing

krassowski: Could casting in tests to `as CommHandler` make the tests pass while narrowing the required API?

martinRenou: What do you mean by narrowing the required API?

martinRenou: Gotcha, I misunderstood the casting to `as CommHandler` in tests. Let me try to do it.

martinRenou: Done!

krassowski (approved):
Thank you @martinRenou!

Also thank you @ianthomas23 and @fleming79 for reviews!

fleming79:
I tried Version 4.4.0b2 with my custom version of [ipylab](https://github.com/fleming79/ipylab) and found using subshells for comms breaks things. I haven't tracked down the exact reason but would like to warn that the default setting could also break things for other users.

When switching to 'Disabled' it works correctly.

fleming79:
I had a look inside Ipykernel and wonder if maybe it might be worth considering running execute requests in a task with queue.
This enables widget comms to pass whilst awaiting in a cell.

image

### Ipykernel patch

As a proof of concept. Just overwrite the method in ipykernel.

``` python 
# ipykernel.ipykernel.py  ~ line 356

    async def execute_request(self, stream, ident, parent):
        """Override for cell output - cell reconciliation."""
        parent_header = extract_header(parent)
        self._associate_new_top_level_threads_with(parent_header)
        import asyncio
        if not hasattr(self, '_exec_queue'):
            self._exec_queue = asyncio.Queue()
            self._executor_task = None
        await self._exec_queue.put(super().execute_request(stream, ident, parent))
        if not self._executor_task or self._executor_task.done():
            async def executor_task():
                while True:
                    coro = await self._exec_queue.get()
                    await coro
            self._executor_task = asyncio.create_task(executor_task())

```

### Notebook

``` python
import asyncio
from ipywidgets import IntSlider, HTML
import time

slider = IntSlider()
value = HTML()
display(slider)
display(value)

async def run_till_100():
    for i in range(100):
        await asyncio.sleep(0.1)
        time.sleep(0.1)
        value.value = str(slider.value)
        if slider.value == 100:
            return 'Done'

print(await run_till_100())

```

ianthomas23:
@fleming79 Maybe I am misunderstanding this, but if I try your example locally it works if subshells are enabled and fails if subshells are disabled:

https://github.com/user-attachments/assets/b7821dbc-a2a9-48c8-bfcc-f4c1273dcd1f

This is using ipykernel 7.0.0a1, ipywidgets 8.1.5 and jupyterlab 4.4.0b2 on macOS. The only problem I see is that the execute count is not updated, which is a known problem.

fleming79:
Apologies @ianthomas23, the example I'd provided was flawed. I've re-written the example using a fresh environment with:

- pip install jupyterlab ipywidgets
- ipykernel: git main branch with patch applied

The kernel patch could be used instead of subshells for comms.

Patch from here; https://github.com/jupyterlab/jupyterlab/pull/17363#issuecomment-2739355243

### Notebook

```
import asyncio
from ipywidgets import IntSlider, HTML, __version__
import time

slider = IntSlider()
value = HTML()
print('ipywidgets', __version__)
display(slider)
display(value)

async def run_till_100():
    while True:
        time.sleep(0.1) # not a good idea as it blocks the main thread
        await asyncio.sleep(0.1)
        value.value = str(slider.value)
        if slider.value == 100:
            return 'Done'
print(await run_till_100())

```


image

Can someone advise what tool to perform video recording of screen with Firefox?

fleming79:
Because each subshell provides a different thread, it also has a different asyncio event loop.  

Here is an example illustrating the problem of using subshells for comms.

I'd recommend the default setting for 'Kernel Comms over subshells' is changed to 'Disabled' to avoid disturbing current users who use asyncio.


``` python
#  Cell 1

from ipywidgets import Button, HBox, HTML
import time
import asyncio

button_set = Button(description = "Set")
button_start = Button(description = "Start")
result = HTML()
event = asyncio.Event()

button_set.on_click(lambda _: event.set())
button_start.on_click(lambda _: asyncio.create_task(my_function()))

async def my_function():
    event.clear()
    result.value = "Waiting for event. Click 'set'"
    try:
        await event.wait()
        result.value = f"{time.ctime()} okay"
    except Exception as e:
        result.value = f"{time.ctime()} {e}"
display(result)
display(HBox([button_start, button_set]))

```


``` python 
# Cell 2
await my_function()
```


### With subshell

image

image

### Without subshell (using patch)

image

image

ianthomas23:
OK, so your latest example doesn't work with `ipykernel` 7.0.0a1 either with or without use of subshells. I assume it used to work in the past. Ipykernel has changes in two areas, using `anyio` instead of `tornado`/`asyncio`, and subshells. It looks like your example is affected by both changes, and the former are greater changes that the latter. If you think that `ipykernel` needs something like your patch then you should probably report it there to get the attention of the `anyio` experts.

As you say, your example doesn't work with subshells because `asyncio` doesn't work across threads. It can be changed to work with subshells by using `threading.Event` rather than `asyncio.Event` and ensuring that `my_function` is run in the main thread not a subshell:
```python
# Cell 1
from ipywidgets import Button, HBox, HTML
import time
import asyncio
from threading import Event

button_set = Button(description = "Set")
button_start = Button(description = "Start")
result = HTML()

start_event = Event()
set_event = Event()

button_set.on_click(lambda _: set_event.set())
button_start.on_click(lambda _: start_event.set())

async def my_function():
    start_event.wait()
    start_event.clear()
    set_event.clear()
    result.value = "Waiting for event. Click 'set'"
    try:
        set_event.wait()
        result.value = f"{time.ctime()} okay"
    except Exception as e:
        result.value = f"{time.ctime()} {e!r}"

display(result)
display(HBox([button_start, button_set]))
```
and cell 2 unchanged. This works for me using `ipykernel 7.0.0a1` without your patch.

fleming79:
@ianthomas23 - Thank you for taking the time to trial the suggestion. I'm sorry to hear the suggested patch didn't work for you straight up. It does work, but you need to pause for a few seconds after restarting the kernel. 

Here is a screenshot of the patched kernel (7.0.0.a1) with Jupyterlab after running the example.

image

Note: the patch was only a proof of concept, not a working fix (for example stdio output incorrectly always ends up at the same cell).

The second example I posted above was to illustrate problems existing users could encounter. In the example it was due to unexpectedly having two separate event loops where previously there was only one. The point being that enabling subshell for comms by default could break things for a lot of users.

Another issue I noticed was that debug breakpoints weren't working where previously they were (using a launch request in vscode).


To summarize my observations/suggestions:
1. Kernel subshells for widget comms  could break things for user code involving widget comms that is not designed to be thread safe. For asyncio based code it introduces one or more event loops depending on where the call originated and the subshell setting in Jupyterlab. 
2. An alternate to using subshells for widget comms could be to modify the kernel to run execute requests in a queue (modifying ipykernel). The benefit of not using threads for widget comms being that existing non thread safe code will continue to work without alteration.

ianthomas23:
@fleming79 To clarify, I didn't use your patch. My sentence "OK, so your latest example doesn't work with ipykernel 7.0.0a1 either with or without use of subshells" should be taken literally.

Ipykernel 7 will be a major release, it is not backward compatible for everything which is why it is a major release. I personally expect all sync code to work unmodified, but probably any semi-complex async code may be broken and need to be changed. I expect other projects will pin to ipykernel<7 initially until they have made the necessary changes in their code.

For your specific suggestions about ipykernel, please please please report them to ipykernel.

Personally I think the default use of subshells in this merged PR should remain as it is, i.e. one subshell per comm target. The inability of ipywidgets to support updates in the presence of a blocking task has been a bug for years, and is fixed by this PR. Yes there will be some use cases using some async workarounds that will be broken by this, but in many cases those workarounds will not be required any more. Users and projects affected by this but not pinning ipykernel<7 can change the setting to disable use of subshells for comms in the usual manner.

PR #17364, opened 2025-03-04
Sort the list of plugins and tokens in the documentation

jtpio:
## References

Sort the tokens and plugins generated by the corresponding Galata test.

This should make it easier to browser the files for the list of tokens and plugins, and on the documentation:

- https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html#core-plugins
- https://jupyterlab.readthedocs.io/en/latest/extension/extension_points.html#core-tokens

And also have cleaner diffs when adding / removing new plugins.

## Code changes

Update playwright test that generates the tokens and plugins lists.

Using the insertion order of object properties as mentioned in step 3 of the spec here: https://tc39.es/ecma262/#sec-ordinaryownpropertykeys

## User-facing changes

Documentation changes.

None


krassowski (approved):
Picked up a conflict from https://github.com/jupyterlab/jupyterlab/pull/17352 but otherwise looks good!

PR #17367, opened 2025-03-05
Require `JupyterLab.IInfo` for the plugin manager plugin

jtpio:
## References

Currentl `@jupyterlab/pluginmanager-extension:plugin` was expecting the `app` it gets in the `activate` function to be a `JupyterLab` instance:

https://github.com/jupyterlab/jupyterlab/blob/d470c501f50ad7075413cd89967a1a8a332b9a2f/packages/pluginmanager-extension/src/index.ts#L54

This may not always be the case, for example if the plugin is used in another lab-based application.

Instead we can require `JupyterLab.IInfo`, which provides the information needed for the plugin. That way another lab-based application could provide `JupyterLab.IInfo` via a plugin, instead of being itself a `JupyterLab`.

## Code changes

- Require `JupyterLab.IInfo`

## User-facing changes

None

None

### DIFF
--- a/packages/pluginmanager-extension/src/index.ts
+++ b/packages/pluginmanager-extension/src/index.ts
@@ -8,6 +8,7 @@
  */
 import {
   ILayoutRestorer,
+  JupyterFrontEnd,
   JupyterFrontEndPlugin,
   JupyterLab
 } from '@jupyterlab/application';
@@ -47,11 +48,12 @@ const pluginmanager: JupyterFrontEndPlugin<IPluginManager> = {
   id: PLUGIN_ID,
   description: 'Enable or disable individual plugins.',
   autoStart: true,
-  requires: [],
+  requires: [JupyterLab.IInfo],
   optional: [ITranslator, ICommandPalette, ILayoutRestorer],
   provides: IPluginManager,
   activate: (
-    app: JupyterLab,
+    app: JupyterFrontEnd,
+    appInfo: JupyterLab.IInfo,
     translator: ITranslator | null,
     palette: ICommandPalette | null,
     restorer: ILayoutRestorer | null
@@ -77,7 +79,7 @@ const pluginmanager: JupyterFrontEndPlugin<IPluginManager> = {
       const model = new PluginListModel({
         ...args,
         pluginData: {
-          availablePlugins: app.info.availablePlugins
+          availablePlugins: appInfo.availablePlugins
         },
         serverSettings: app.serviceManager.serverSettings,
         extraLockedPlugins: [
--- a/packages/application/src/lab.ts
+++ b/packages/application/src/lab.ts
@@ -325,7 +325,7 @@ export namespace JupyterLab {
   }
 
   /**
-   * The layout restorer token.
+   * The application info token.
    */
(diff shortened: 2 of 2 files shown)


packages/pluginmanager-extension/src/index.ts
```diff
@@ -47,11 +48,12 @@ const pluginmanager: JupyterFrontEndPlugin<IPluginManager> = {
...
   description: 'Enable or disable individual plugins.',
   autoStart: true,
-  requires: [],
+  requires: [JupyterLab.IInfo],
   optional: [ITranslator, ICommandPalette, ILayoutRestorer],
   provides: IPluginManager,
   activate: (
-    app: JupyterLab,
+    app: JupyterFrontEnd,
+    appInfo: JupyterLab.IInfo,
```
krassowski, line 56: I guess the added value is in being able to override the available plugin list?

I do not follow the use case here (not very well).

jtpio: Yes, since most plugins expect a `JupyterFrontEnd` and not a `JupyterLab` as the first argument of the `activate` function.

There is some information about the motivation in the description above:

> Instead we can require JupyterLab.IInfo, which provides the information needed for the plugin. That way another lab-based application could provide JupyterLab.IInfo via a plugin, instead of being itself a JupyterLab.

krassowski: If a lab-based application is not a `JupyterLab`, then why should it provide `JupyterLab.IInfo`? Shouldn't it be like `JupyterFrontEnd.IInfo` token? I guess I am not suggesting a change needs to be made this, I just mean that this is not much of an improvement.

I suspect that you mean something like Notebook 7 should not need to inherit from `JupyterLab` class. Technically it does not need to inherit as long as it has `.info`, but having this as an implicit required attribute is indeed not great. Having a required token is a little better because the plugin will fail a little bit earlier, but it does not solve the "I am not a JupyterLab, I just quack like one" issue, nor does it guarantee correctness downstream as ultimately if downstream is expected to provide `JupyterLab.IInfo` but they are not a `JupyterLab`, they will need to provide some dummy values and might fail when attempting to do so.

If you think we should move forward with this token as-is, I am happy to approve and merge, but also curious if you thought of other solutions.

jtpio: This PR tries to slightly improve the current situation and make the "requirement" on that `IInfo` object a bit less dependent on the `app` being provided to the `activate` function.

Indeed, in the case of Notebook 7 (or other lab-based applications), this information could be provided as a separate plugin instead of having it as a mandatory property on the `app`. Although it's true it can be weird to have to define a `JupyterLab.IInfo` if the app is not a `JupyterLab` app itself.

This makes it sightly easier for downstreams, but agreed that in the case of JupyterLab it does not change the situation that much.

krassowski (approved):
Approving it as per discussion, though we should find a better way to get the plugin list to the plugin manager in the future - thank you @jtpio!

jtpio:
Cool, thanks @krassowski for the review 👍 

Agreed we can think of ways to improve this in the future.

PR #17371, opened 2025-03-06
Fix rendering URLs as linked in errors

### LINKED ISSUE #17225: `stderr` in cell outputs do not auto-link URLs like `stdin`
## Description

URLs are automatically linked in the output area of a cell if it is written `stdout` but not if it is written to `stderr`. Here is what it looks like:

Image

This rendering is implemented in the `@jupyterlab/rendermime` package for the MIME types `application/vnd.jupyter.stderr` and `application/vnd.jupyter.stdout`:

https://github.com/jupyterlab/jupyterlab/blob/main/packages/rendermime/src/factories.ts#L64-L82

## Reproduce
Create a notebook with a Python kernel and run the following code to see the difference between `stdout` and `stderr`.

```python
import sys
print("https://example.com/ written to stderr", file=sys.stderr)
print("https://example.com/ written to stdout", file=sys.stdout)
```

## Expected behavior

Unless there is some consideration I am missing (perhaps security-related), I would expect the same link handling behavior in both `stdout` and `stderr`.

## Context

- Operating System and version:  macOS
- Browser and version: Chrome 131
- JupyterLab version: 4.4.0a2

afshin:
This PR fixes a bug in error output rendering that prevented URLs from being rendered as HTML links. This bug was introduced when IPython-style file linking was implemented for error output.

Here is how URL linking in error output looked in JupyterLab 4.0.x:

jupyterlab-v4 0 9

In JupyterLab 4.3.5, some files are linked, but a nearby URL can cause a file linking to fail. Additionally, URLs are not linked:

jupyterlab-v4 3 5 unmarked

This PR fixes both of those cases:

jupyterlab-fix-17225

## References

Fixes #17225 

## Code changes
- Changes a broad selector (all `a` tags) to a more narrow selector `a[data-path]` to only do IPython-style file linking for relevant links so that URLs can be linked properly.
- Adds a unit test to verify that both URL and file links render properly.
- Changes some private functions that returns promises to `async` to simplify them and to reduce the number of immediately-resolving promises.

## User-facing changes

**before** (errors highlighted in magenta)

jupyterlab-v4 3 5 marked

**after**

jupyterlab-fix-17225

n/a


packages/rendermime/src/renderers.ts
```diff
@@ -1288,12 +1269,13 @@ namespace Private {
...
     linkHandler: IRenderMime.ILinkHandler | null
-  ): Promise<void> {
-    // Handle anchor elements.
-    const anchors = node.getElementsByTagName('a');
-    for (let i = 0; i < anchors.length; i++) {
-      await handlePathAnchor(anchors[i], resolver, linkHandler);
-    }
+  ): Promise<unknown> {
+    const anchors: HTMLAnchorElement[] = Array.from(
+      node.querySelectorAll('a[data-path]')
```
afshin, line 1274: This selector makes sure that only `<a>` nodes that have a `data-path` attribute added to them are processed in `handlePaths`. This was causing errors to render without URLs linked properly.

krassowski (approved):
Thank you @afshin!

krassowski:

Issue #17375, opened 2025-03-08
%pip magic doesn't work when installed using "uv tool install juypterlab"


simonw:
## Description

If you install `jupyterlab` like this:
```bash
uv tool install jupyterlab
```
And then run it like this:
```bash
jupyter-lab
```
If you try running this in a cell:
```
%pip install llm
```
You get an error that looks like this:
```
.../.local/share/uv/tools/jupyterlab/bin/python: No module named pip
```
## Expected behavior

It should work.

I [found a workaround](https://til.simonwillison.net/jupyter/jupyterlab-uv-tool-install) - running this in a cell fixes the problem:
```python
import subprocess, sys
subprocess.check_call([sys.executable, "-m", "ensurepip"])
```

simonw:
I've seen this problem in my own projects (which call `pip` under the hood to install extra plugins). The fix is to add `pip` as a concrete dependency for the project.

simonw:
My PR has a propsed fix. Here's how to test it:

You can see the bug by running `jupyter-lab` from a [recent commit](https://github.com/jupyterlab/jupyterlab/tree/d470c501f50ad7075413cd89967a1a8a332b9a2f) to main like this:
```bash
uvx --from 'https://github.com/jupyterlab/jupyterlab/archive/d470c501f50ad7075413cd89967a1a8a332b9a2f.zip' jupyter-lab
```
Try running `%pip install llm` in a cell and it will fail with the error.

You can test my fix by running `jupyter-lab` from my branch [with the commit](https://github.com/jupyterlab/jupyterlab/tree/bd2ce80995b39423b58ae37bece3d706e1d09845) in like this using `uvx`:
```bash
uvx --from 'https://github.com/jupyterlab/jupyterlab/archive/bd2ce80995b39423b58ae37bece3d706e1d09845.zip' jupyter-lab
```
Now if you run `%pip install llm` it will work correctly.

jtpio:
Thanks @simonw for the suggestion.

Since the `%pip` magic is not strictly related to JupyterLab itself, but specific to the kernel being used (in this case `ipykernel`), not sure the `pip` dependency should be added to JupyterLab.

Could an alternative `uv` command like the following one work, so make sure `pip` gets install too?

```
uv tool install jupyterlab --with pip
```

jtpio:
[...]

PR #17376, opened 2025-03-08
Add pip as a dependency

simonw:
## References

- #17375

## Code changes

I added `"pip"` to the list of dependencies in `pyproject.toml`.

## User-facing changes

This will fix a bug where `%pip install X` doesn't work in Jupyter Lab installations that were created using `uv tool install`.

None.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
     "jupyterlab_server>=2.27.1,<3",
     "notebook_shim>=0.2",
     "packaging",
+    "pip",
     "setuptools>=41.1.0",
     "tomli>=1.2.2;python_version<\"3.11\"",
     "tornado>=6.2.0",


simonw:
You can see the bug by running `jupyter-lab` from a [recent commit](https://github.com/jupyterlab/jupyterlab/tree/d470c501f50ad7075413cd89967a1a8a332b9a2f) to main like this:
```bash
uvx --from 'https://github.com/jupyterlab/jupyterlab/archive/d470c501f50ad7075413cd89967a1a8a332b9a2f.zip' jupyter-lab
```
Try running `%pip install llm` in a cell and it will fail with the error.

You can test my fix by running `jupyter-lab` from my branch [with the commit](https://github.com/jupyterlab/jupyterlab/tree/bd2ce80995b39423b58ae37bece3d706e1d09845) in like this using `uvx`:
```bash
uvx --from 'https://github.com/jupyterlab/jupyterlab/archive/bd2ce80995b39423b58ae37bece3d706e1d09845.zip' jupyter-lab
```
Now if you run `%pip install llm` it will work correctly.

krassowski:
I think we would not want to merge this because - as mentioned in https://github.com/jupyterlab/jupyterlab/issues/17375#issuecomment-2709885386 - JupyterLab does not implement `%pip` magic; instead it is implemented in IPython which is a dependency of IPykernel and only that is a dependency of JupyterLab. If a dependency was to be added it should be added at an appropriate level, this is probably in IPython.

Still, it is not clear to me if IPython should depend on `pip` to ensure `%pip` magic works - after all, it also ships with `%conda` and there I think it is obvious that conda should not be a dependency of IPython.

Ok to close?

simonw:
Closing this PR because it sounds like it's an ipython issue (I still think that it should get fixed somewhere though.)

PR #17378, opened 2025-03-09
Support for keyboard layouts.

### LINKED ISSUE #7579: Support non-US keyboard layouts
There are a number of issues with using JupyterLab with non-US keyboards. To summarize, we have two keyboard shortcut systems:

* Lumino (phosphor) keyboard shortcut system, used throughout JupyterLab. This shortcut system allows for non-US keyboards, but we need to have relevant keyboard mappings. We've talked about writing some UX to make it easier to make a keymap (basically capturing the key character and usb code and associating them, perhaps combining maps from different browsers)
* codemirror keyboard shortcut system for text editors and code cells. Should we expose this in some way that integrates with the Lumino shortcut system? Or should we expose this as a separate mechanism?

Designing a UX for capturing new keyboard mappings would be a great new extension, for example.

## Related issues

#5164

### LINKED ISSUE #17355: Setting without a default doesn't save in form editor
## Description

I was making a new extension, and introducing a new schema with settings. I noticed that when I used the form editor for it, the setting never got saved (although the UI happily claims that it was saved, no network request was made, and no settings change signal was emitted). If I used the JSON editor, I was able to save it.

## Reproduce

1. Create a new schema with only one setting, without giving it a default value.
2. Open the (form-based) settings editor to that extension, and change the value.
3. No network request or change signal is emitted.

## Expected behavior

The setting is saved.

## Suspected cause

I think the issue happens here:

https://github.com/jupyterlab/jupyterlab/blob/1fb48d636f90928c3b4d883d80030f9bc6da6f73/packages/settingeditor/src/SettingsFormEditor.tsx#L156-L165

The call `this.props.settings.isDefault(this._formData)` intends to check if all the new values are default values. However, `isDefault` checks if **either**: - the value is undefined or empty object/array
- OR if the default is undefined

If any of them are true, is will completely ignore that key:

https://github.com/jupyterlab/jupyterlab/blob/1fb48d636f90928c3b4d883d80030f9bc6da6f73/packages/settingregistry/src/settingregistry.ts#L736-L745

This means that it will fall back to its default claim and return true. I would argue that semantically, if no default is defined, then the value cannot be the default. Or alternatively, with this logic, the function will *always* return true, no matter what the value is.

Probably `defaultValue === undefined` should be dropped from this check?

CC @fcollonval

vidartf:
## References

Fixes #17355.

Resolves #7579.

## Code changes

Adds a new plugin in the shortcuts-extension package. Relies on the WIP lumino branch in this PR https://github.com/jupyterlab/lumino/pull/291/ .

## User-facing changes

- A new section is available in the settings editor for picking the keyboard layout.
- Users can pick from pre-defined keymaps in the settings editor.
- Users can capture their own keyboard mapping by a separate input widget in the settings dialog.
- Users can now define shortcuts based on non-US keyboards. I.e. if I add a shortcut for my `æ` key, it will only trigger if I push a button that produces that character (e.g. I have a Norwegian keyboard and press the `code=Quote` button, or if I have a Danish keyboard and press the `code=Semicolon` button). Non-QWERTY can also be configured (AZERTY, Dvorak, etc).

Not sure if it qualifies as backwards-incompatible, but the shortcut-extension package will require a dependency on an as-yet unreleased feature in the Lumino "keyboard" package.

### DIFF
--- a/packages/settingeditor/src/SettingsFormEditor.tsx
+++ b/packages/settingeditor/src/SettingsFormEditor.tsx
@@ -18,7 +18,7 @@ import {
 import { Debouncer } from '@lumino/polling';
 import { IChangeEvent } from '@rjsf/core';
 import validatorAjv8 from '@rjsf/validator-ajv8';
-import { Field, UiSchema } from '@rjsf/utils';
+import type { Field, UiSchema, Widget } from '@rjsf/utils';
 import { JSONSchema7 } from 'json-schema';
 import { Button } from '@jupyterlab/ui-components';
 
@@ -45,6 +45,11 @@ export namespace SettingsFormEditor {
      * Dictionary used for custom field renderers in the form.
      */
     renderers: { [id: string]: { [property: string]: Field } };
+
+    /**
+     * Dictionary used for custom widget renderers in the form.
+     */
+    widgetRenderers?: { [id: string]: { [property: string]: Widget } };
 
     /**
      * Translator object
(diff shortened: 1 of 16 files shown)


vidartf:
Note: currently I've also added support for this [experimental Web API](https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/getLayoutMap) for demonstration purposes, but the feature does not rely on that. It only makes things more convenient in browsers that support it.

PR #17395, opened 2025-03-14
Add `IKernelSpecAPICLient` and `ITerminalAPIClient`, fix definitions of interfaces

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17348.

To have a consistent approach for the services making requests to the rest API via their internal `restapi.ts` module.

## Code changes

- Add `IKernelSpecAPIClient` and `ITerminalAPIClient`
- Allow the respective managers to use a custom instance
- Make a proper definition of the interfaces, instead of extending from the concrete class (similar to https://github.com/jupyterlab/jupyterlab/pull/17397)

## User-facing changes

None

None

### DIFF
--- a/packages/services/src/terminal/manager.ts
+++ b/packages/services/src/terminal/manager.ts
@@ -170,12 +170,13 @@ export class TerminalManager extends BaseManager implements Terminal.IManager {
    * options.
    */
   async startNew(
-    options?: Terminal.ITerminal.IOptions
+    options: Terminal.ITerminal.IOptions = {}
   ): Promise<Terminal.ITerminalConnection> {
-    const model = await this._terminalAPIClient.startNew(
-      options?.name,
-      options?.cwd
-    );
+    const { name, cwd } = options;
+    const model = await this._terminalAPIClient.startNew({
+      name,
+      cwd
+    });
     await this.refreshRunning();
     return this.connectTo({ model });
   }
--- a/packages/services/src/terminal/restapi.ts
+++ b/packages/services/src/terminal/restapi.ts
@@ -3,7 +3,7 @@
 
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 import { ServerConnection } from '../serverconnection';
-import { ITerminalAPIClient } from './terminal';
+import { ITerminal, ITerminalAPIClient } from './terminal';
 
 /**
  * The url for the terminal service.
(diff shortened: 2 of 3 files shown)


packages/services/src/kernel/default.ts
```diff
@@ -235,11 +239,9 @@ export class KernelConnection implements Kernel.IKernelConnection {
...
       return this._specPromise;
     }
-    this._specPromise = KernelSpecAPI.getSpecs(this.serverSettings).then(
-      specs => {
-        return specs.kernelspecs[this._name];
-      }
-    );
+    this._specPromise = this._kernelSpecAPIClient.get().then(specs => {
+      return specs.kernelspecs[this._name];
+    });
```
krassowski, line 244: Should this one be replaced too?

https://github.com/jupyterlab/jupyterlab/blob/9749b07d54b4ac7622e617f1afdce770849c6f92/packages/services/src/kernelspec/manager.ts#L119-L120

Otherwise, we would could get discordant results, right?

packages/services/src/terminal/manager.ts
```diff
@@ -296,6 +295,11 @@ export namespace TerminalManager {
      * When the manager stops polling the API. Defaults to `when-hidden`.
      */
     standby?: Poll.Standby | (() => boolean | Poll.Standby);
+
+    /**
+     * The Terminal API Client
```
krassowski, line 300: nit:

```suggestion
     * The Terminal API client.
```

packages/services/src/kernel/default.ts
jtpio: Should be fixed in https://github.com/jupyterlab/jupyterlab/pull/17395/commits/c34963e7d0b953801e523b371bb5a265bd1b5ee4

packages/services/src/terminal/restapi.ts
```diff
@@ -120,6 +121,71 @@ export async function shutdownTerminal(
...
+  /**
+   * Start a new terminal session.
+   *
+   * @param name - The name of the target terminal.
+   *
+   * @param cwd - The path in which the terminal will start.
+   *
+   * @returns A promise that resolves with the session model.
+   */
+  async startNew(name?: string, cwd?: string): Promise<IModel> {
```
krassowski, line 161: This is not a strong opinion, but should we use `options?: ITerminalOptions` pattern here? It looks like the caller already does use `ITerminal.IOptions` (we would probably need to make that one inherit from a new interface).

This way all `startNew` methods would follow the `options` API, and adding/removing things is easier.

jtpio: Agree that switching to `options?: ITerminalOptions` brings a couple of advantages 👍 

For kernels, it looks like the manager's `startNew` method currently takes an `IKernelOptions`: 

https://github.com/jupyterlab/jupyterlab/blob/9749b07d54b4ac7622e617f1afdce770849c6f92/packages/services/src/kernel/manager.ts#L179

Defined here:

https://github.com/jupyterlab/jupyterlab/blob/9749b07d54b4ac7622e617f1afdce770849c6f92/packages/services/src/kernel/restapi.ts#L117

So maybe we can follow the same approach have both `startNew` methods take a `ITerminal.IOptions`.

krassowski (approved):
Thank you @jtpio!

PR #17397, opened 2025-03-14
Fix definition of the `ILicensesClient` interface

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17361.

## Code changes

Properly define `ILicensesClient`.

Otherwise, downstreams have to also implement `private` properties, which defeats the purpose of defining the interface in the first place.

Mentioned in https://learn.microsoft.com/en-us/archive/msdn-magazine/2015/january/typescript-understanding-typescript#typescript-inherits-differently:

> TypeScript requires that you include private members in the interface to be inherited from the class that the interface extends, instead of being reimplemented in the derived class.

## User-facing changes

None

None.

This PR updates a public interface which was added in the previous beta (`4.4.0b1`), so not part of the API surface yet.

### DIFF
--- a/packages/apputils/src/licenses.tsx
+++ b/packages/apputils/src/licenses.tsx
@@ -324,7 +324,7 @@ export namespace Licenses {
     }
 
     /**
-     * fetch the license bundles from the server.
+     * Fetch the license bundles from the server.
      */
     async getBundles(): Promise<ILicenseResponse> {
       const response = await ServerConnection.makeRequest(
--- a/packages/apputils/src/licenses.tsx
+++ b/packages/apputils/src/licenses.tsx
@@ -309,14 +309,22 @@ export namespace Licenses {
     }
 
     /**
-     * Get the link to download the licenses in a given format.
+     * Download the licenses in the requested format.
      */
-    async getDownloadLink(options: IDownloadOptions): Promise<string> {
-      return `${this._licensesUrl}?format=${options.format}&download=1`;
+    async download(options: IDownloadOptions): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/apputils/src/tokens.ts
```diff
@@ -83,7 +83,17 @@ export const ILicensesClient = new Token<ILicensesClient>(
 /**
  * An interface for the license client.
  */
-export interface ILicensesClient extends Licenses.LicensesClient {}
+export interface ILicensesClient {
+  /**
+   * fetch the license bundles from the server.
```
krassowski, line 88:
```suggestion
   * Fetch the license bundles from the server.
```

packages/apputils/src/licenses.tsx
```diff
@@ -309,10 +309,18 @@ export namespace Licenses {
     }
 
     /**
-     * Get the link to download the licenses in a given format.
+     * Download the licenses in the requested format.
      */
-    async getDownloadLink(options: IDownloadOptions): Promise<string> {
-      return `${this._licensesUrl}?format=${options.format}&download=1`;
+    async download(options: IDownloadOptions): Promise<void> {
```
jtpio, line 314: Taking the opportunity to make this `download` method more generic, to give more flexibility to downstreams that may want to override it.

krassowski (approved):
Thank you @jtpio, this feels more future proof

packages/apputils/src/licenses.tsx
jtpio: For now using the same approach and creating the `<a>` element is probably the simplest.

But it gives the possibility to create a `Blob` and set a custom file name.

PR #17405, opened 2025-03-18
Fix handling of the CSV viewer theme

jtpio:
## References

Noticed when working with CSV files with the latest JupyterLab 4.3.6, this appears to be an issue with the latest commit on `main` too.

When using the dark theme, opening a CSV file would display it with its light theme.

## Code changes

Fix handling of the current theme for the CSV viewer.

## User-facing changes

**Before**

https://github.com/user-attachments/assets/f8ff74ca-1b6e-4d86-9c97-e98643d0d8a8

**After**

https://github.com/user-attachments/assets/9488be27-7df2-4cb5-89a6-31e19710f585

None


krassowski:
Maybe we should add a visual regression snapshot for the dark theme of CSV editor? It is not a first time it regressed, previously I fixed it in https://github.com/jupyterlab/jupyterlab/pull/14822

krassowski (approved):
Thank you @jtpio!

PR #17412, opened 2025-03-19
When an entire cell is extracted from lsp virtual document do not add new separator lines after it

### LINKED ISSUE #15870: Do not add an empty cell when an entire Markdown cell gets extracted into a virtual document
## Description

JupyterLab 4 added support for Markdown documents by using extractors mechanisms; it seems that markdown cells are erroneously included as empty cells in the resulting Python document.

It seems that the logic in `TextForeignCodeExtractor` and in the `extractForeignCode()` or `appendCodeBlock()` need to be modified so that when an entire cell gets extracted no cell would be added (rather than adding an empty code cell, padded by two empty lines on each side).

## Reproduce

See jupyterlab-lsp issues:
- https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1054
- https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1040

## Expected behavior

When an entire cell is extracted do not add new lines before and after it.

This could be solved by special casing the logic in:

https://github.com/jupyterlab/jupyterlab/blob/2d681271b159ac533df1ddb576ddcb256c44e59c/packages/lsp/src/virtual/document.ts#L614-L673

For example if `lines` is just one empty line we could skip adding `this.blankLinesBetweenCells`. This need to be conditional on `foreignDocumentsMap` being non-empty so that it does not lead to false negatives (e.g. undetected empty code cells),

## Context 

- JupyterLab version: 4.1.2

andrii-i:
## References

Fixes #15870

## Code changes

When an entire cell is extracted from lsp virtual document do not add new separator lines after it.

## User-facing changes

User


krassowski:
@andrii-i is it still a draft?

krassowski:
@andrii-i just checking if you need any help to push this through the finish line?

andrii-i:
Thank you for pinging me on this @krassowski. I've added a follow-up question to your suggestion on next steps, could you please give it a look? That would really help and I'll try to wrap it up quickly after clarifying that https://github.com/jupyterlab/jupyterlab/issues/15870#issuecomment-2863475508

Issue #17411, opened 2025-03-18
Log Console "Copy Output to Clipboard" does not work as expected


jtpio:
## Description

The "Copy Output to Clipboard" does not consistently copy the output when used in the log console:

![Image](https://github.com/user-attachments/assets/a16f322c-00d6-4cbc-a181-302bfb160c2c)

## Reproduce

Right click on a error in the log console:

https://github.com/user-attachments/assets/428095db-8398-4dd7-a971-d06a8fa06476

Actually this seems to be "working" when the selected cell in the corresponding notebook has an output:

https://github.com/user-attachments/assets/d60bded6-2ee2-44e7-8044-87a3684a16ae

## Expected behavior

It should copy the cell output content just like in the notebook, event if the active cell does not have any output.

## Context

This command comes from the notebook extension defined here, which checks for an active cell:

https://github.com/jupyterlab/jupyterlab/blob/e1487fdb0444683193a2d9b078b37dca5bf80e51/packages/notebook-extension/src/index.ts#L1593

Which likely explains why the output is properly copied when the notebook cell is first selected, but not when it's an empty cell.

jtpio:
Thinking more about it, it's not clear if "Copy Output to Clipboard" was actually meant to be used in the log console, or if it was just a side effect of the selector being too generic:

https://github.com/jupyterlab/jupyterlab/blob/e1487fdb0444683193a2d9b078b37dca5bf80e51/packages/notebook-extension/src/index.ts#L1615-L1619

In other words, maybe the selector for the context menu item should be more specific to the notebook, since the tracker is it checking in the command is a `INotebookTracker`.

PR #17413, opened 2025-03-19
More specific selector for "Copy Output to Clipboard"

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/17411

Since the related command (`notebook:copy-to-clipboard`) is added in `@jupyterlab/notebook-extension` and expects an `INotebookTracker` here:

https://github.com/jupyterlab/jupyterlab/blob/e1487fdb0444683193a2d9b078b37dca5bf80e51/packages/notebook-extension/src/index.ts#L1558

This has effect to disable the "Copy Output to Clipboard" context menu entry for the log console.

As noticed in https://github.com/jupyterlab/jupyterlab/issues/17411#issuecomment-2734323606, this context menu entry for the log console may not have been intentional.

## Code changes

More specific selector for "Copy Output to Clipboard".

## User-facing changes

https://github.com/user-attachments/assets/02a81687-4045-412c-9c71-b39749cf8f39

None

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -1614,7 +1614,7 @@ function activateCopyOutput(
 
   app.contextMenu.addItem({
     command: CommandIDs.copyToClipboard,
-    selector: '.jp-OutputArea-child',
+    selector: '.jp-Notebook .jp-OutputArea-child',
     rank: 0
   });
 }


jtpio:
If we want to have a "Copy Output to Clipboard" functionality for the log console, we could look into adding a separate command (and selector) in the `logconsole-extension`. But this could be done separately.

krassowski:
For context menu specifically it would be sufficient to use `app.contextMenuHitTest` to find the output. Then this command could be moved from notebook to `outputarea`, though there is no `outputarea-extension`. 

For use in command palette the tracker is still useful.

jtpio:

PR #17414, opened 2025-03-19
Fix types in translation package, remove usages of `any`

### LINKED ISSUE #17333: The `fetch` return type in `ITranslatorConnector` is incorrect
>  It means `fetch` return type is incorrect. That means we are exposing an invalid contract in newly fixed `ITranslatorConnector`, so yeah this should be addressed before we release 4.4.
> 
> Maybe you could cherry-pick from my https://github.com/jupyterlab/jupyterlab/pull/16219 - it is rather large so pasting snapshot of relevant section:
> 
> image
> 
> And possibly also
> 
> image
> 
> _Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/17329#discussion_r1966565821_

> Yes happy to track separately (as a release blocker) and be assigned to it. This can be addressed later during the beta or rc cycles.
>
> _Originally posted by @jtpio in https://github.com/jupyterlab/jupyterlab/pull/17329#discussion_r1966576370_

krassowski:
## References

- Fixes #17333
- Builds upon #16219

## Code changes

The `ITranslatorConnector` interface is now properly typed.

## User-facing changes

None

None

### DIFF
--- a/packages/translation/src/tokens.ts
+++ b/packages/translation/src/tokens.ts
@@ -76,7 +76,7 @@ export interface ILanguageData {
 }
 
 /*
- * Language list is the reply returned when no specific language
+ * List of available languages.
  */
 export interface ILanguageList {
   /**
--- a/packages/translation/src/tokens.ts
+++ b/packages/translation/src/tokens.ts
@@ -10,20 +10,102 @@ import { Token } from '@lumino/coreutils';
 import { requestTranslationsAPI } from './server';
 
 /**
- * Application default locale
+ * Application default locale.
  */
 export const DEFAULT_LANGUAGE_CODE = 'en';
 
-/*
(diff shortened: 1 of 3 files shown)


packages/translation/src/tokens.ts
```diff
@@ -10,20 +10,102 @@ import { Token } from '@lumino/coreutils';
...
+  displayName: string;
+  /**
+   * Display name of the language in the language itself.
+   */
+  nativeName: string;
+}
+
 /*
- * Translation
+ * Language list is the reply returned when no specific language
```
krassowski, line 79:
```suggestion
 * List of language data.
```

jtpio (approved):
Thanks!

PR #17415, opened 2025-03-19
Create "Kernels" section, split statusbar settings

### LINKED ISSUE #17406: Create a single "Kernel" settings section
> Maybe this should be just "Kernel settings" and later we can also fold this orphan in there:
>
> image
> 
> My reasoning is:
> - the terminals are not shown in status bar unless any is open
> - we can rework the behaviour of `showStatusBarItem` to only affect kernels
> - we could add another `showStatusBarItem` for terminal specifically, as we already have "Terminal" settings section:
>
> image
>
> Basically I would like to decouple these so that we do not have "terminals and kernels" section in settings just because the "running sessions" encompasses both in code.
>
> CC @afshin @jtpio?
>
> _Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/pull/17363#discussion_r1995272164_

krassowski:
## References

Fixes #17406

## Code changes

- renames `session-settings` created in #17363 (merged yesterday, not released) to `kernel-settings`
- add setting for showing status bar item in terminal settings

### Unified Kernels section in Settings

kernels

### New setting in terminals

image

## Follow-up work:

- move "kernel dialogs" into "Kernels settings" (this would go into 5.0 unless we can migrate user settings)
- maybe invert the relation: make `@jupyterlab/apputils-extension:running-sessions-status` provide a token and then the terminals plugin register its preference for showing or not showing terminal count.

### DIFF
--- a/packages/apputils-extension/src/index.ts
+++ b/packages/apputils-extension/src/index.ts
@@ -728,10 +728,24 @@ const sanitizer: JupyterFrontEndPlugin<IRenderMime.ISanitizer> = {
   }
 };
 
+/*
+ * A plugin owning the kernel settings
+ */
+export const kernelSettings: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/apputils-extension:kernels-settings',
+  description: 'Reserves the name for kernel settings.',
+  autoStart: true,
+  requires: [ISettingRegistry],
+  activate: (_app: JupyterFrontEnd, settingRegistry: ISettingRegistry) => {
+    void settingRegistry.load(kernelSettings.id);
+  }
+};
+
 /**
  * Export the plugins as default.
  */
 const plugins: JupyterFrontEndPlugin<any>[] = [
+  kernelSettings,
   announcements,
   kernelStatus,
   licensesClient,
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -217,7 +217,7 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
           ];
 
         disposable?.dispose();
-        if (showKernels || showTerminals) {
+        if (showKernels || showTerminals !== false) {
           disposable = registerItem({
             showKernels,
             showTerminals
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -225,14 +225,17 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
         }
       };
 
+      const kernelsPluginId = '@jupyterlab/apputils-extension:kernels-settings';
+      const terminalPluginId = '@jupyterlab/terminal-extension:plugin';
+
       void Promise.all([
-        // terminal settings may be missing if terminal plugin is not included
-        settingRegistry
-          .load('@jupyterlab/apputils-extension:kernels-settings')
-          .catch(() => undefined),
-        settingRegistry
-          .load('@jupyterlab/terminal-extension:plugin')
-          .catch(() => undefined)
+        // Settings may be missing if the respective plugins are not enabled/included.
+        kernelsPluginId in settingRegistry.plugins
+          ? settingRegistry.load(kernelsPluginId).catch(() => undefined)
+          : Promise.resolve(undefined),
+        terminalPluginId in settingRegistry.plugins
+          ? settingRegistry.load(terminalPluginId).catch(() => undefined)
+          : Promise.resolve(undefined)
       ]).then(([kernelSettings, terminalSettings]) => {
         onSettingsUpdated(kernelSettings, terminalSettings);
         if (kernelSettings) {
--- a/packages/apputils-extension/src/statusbarplugin.ts
+++ b/packages/apputils-extension/src/statusbarplugin.ts
@@ -148,7 +148,10 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
     translator: ITranslator,
     settingRegistry: ISettingRegistry | null
   ) => {
-    const createStatusItem = () => {
+    const createStatusItem = (options: {
+      showKernels: boolean;
+      showTerminals?: boolean;
+    }) => {
       const item = new RunningSessions({
         onClick: () => app.shell.activateById('jp-running-sessions'),
         onKeyDown: (event: KeyboardEvent<HTMLImageElement>) => {
@@ -163,7 +166,8 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
           }
         },
         serviceManager: app.serviceManager,
-        translator
+        translator,
+        ...options
       });
 
       item.model.sessions = Array.from(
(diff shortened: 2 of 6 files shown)


packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -188,25 +195,63 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
...
-          settings.changed.connect(onSettingsUpdated);
-        });
+      void Promise.all([
+        // terminal settings may be missing if terminal plugin is not included
+        settingRegistry
+          .load('@jupyterlab/apputils-extension:kernels-settings')
+          .catch(() => undefined),
+        settingRegistry
+          .load('@jupyterlab/terminal-extension:plugin')
+          .catch(() => undefined)
```
krassowski, line 235: This leads to a console warning in `app` and `federated` examples tests as this plugin is not included and thus a the job fails:

```
    tornado.web.HTTPError: HTTP 404: Not Found (Schema not found: /home/runner/work/jupyterlab/jupyterlab/examples/app/build/schemas/@jupyterlab/terminal-extension/plugin.json)
[W 2025-03-19 11:45:31.457 AppTest] 404 GET /foo/lab/api/settings/@jupyterlab/terminal-extension:plugin?1742384731424 (bb717af38d23460db6000a477810a24d@::1) 13.81ms referer=http://localhost:8888/foo/lab
```

It feels like we would need to check if this setting exists first.

krassowski: > It feels like we would need to check if this setting exists first.

Or should I already make this plugin return a token with `showTerminals` setter and optionally require it in the terminal plugin?

Any opinions?

krassowski: So I went for checking if the plugin exists rather than for adding a new token. This is because adding a new token here:
- would mean we cannot remove it later without breaking compatibility which is larger comittment
- is a bit awkward as we do not have tokens representing a settings object elsewhere, so I wanted to have a bit more time to think about it and alternatives

This is what I was thinking:

```ts
/**
 * The running session status indicator model.
 */
export const IRunningSessionsStatusModel = new Token<IRunningSessionsStatusModel>(
  '@jupyterlab/apputils:IRunningSessionsStatusModel',
  'A model allowing to configure running sessions statusbar content.'
);


/**
 * The running session status indicator model.
 */
export interface IRunningSessionsStatusModel {
  /**
   * Whether to show kernels, true by default.
   */
  showKernels?: boolean;

  /**
   * Whether to show terminals.
   *
   * The default is true if one or more terminals are open, false otherwise.
   */
  showTerminals?: boolean;
}
```

and in implementation `showKernels` and `showTerminals` would have setters with side-effect of re-rendering the statusbar item.

brichet:
> * maybe invert the relation: make `@jupyterlab/apputils-extension:running-sessions-status` provide a token and then the terminals plugin register its preference for showing or not showing terminal count.

This options seems more reasonable to me, and would clarify the implementation, I think.
The new `@jupyterlab/apputils-extension:kernels-settings` plugin (which currently does nothing AFAIU) could also require this plugin to update its state, with a `showKernel` setter.

brichet:
Just saw your last comment, sorry for the noise.

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -188,25 +195,66 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
...
-        .then(settings => {
-          onSettingsUpdated(settings);
-          settings.changed.connect(onSettingsUpdated);
-        });
+      const kernelsPluginId = '@jupyterlab/apputils-extension:kernels-settings';
+      const terminalPluginId = '@jupyterlab/terminal-extension:plugin';
+
+      void Promise.all([
+        // Settings may be missing if the respective plugins are not enabled/included.
+        kernelsPluginId in settingRegistry.plugins
```
brichet, line 233: Can't we use `app.isPluginActivated(kernelsPluginId)` to know if the plugin is included and activated ?

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -188,25 +195,66 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
...
       let disposable: IDisposable;
       const onSettingsUpdated = (
-        settings: ISettingRegistry.ISettings
+        kernelSettings?: ISettingRegistry.ISettings,
+        terminalsSettings?: ISettingRegistry.ISettings
       ): void => {
-        const showStatusBarItem = settings.get('showStatusBarItem')
-          .composite as boolean;
+        const showTerminalsMap = {
+          'if-any': undefined,
```
brichet, line 202: If the `showTerminals` setting is set to 'if-any' and the `showKernels` set to `false`, the `RunningSession` item is disposed.

 [record-2025-03-21_12.33.30.webm](https://github.com/user-attachments/assets/f17a4b46-5de3-4f69-9635-68081c31a8e8)

brichet (commented):
Thanks @krassowski, I have some comments below.

packages/apputils-extension/src/statusbarplugin.ts
krassowski: I think that `isPluginActivated` may be false for terminals initially because activation happens in order to resolve tokens, right? Since there is no token relation the order would be random.

packages/apputils-extension/src/statusbarplugin.ts
```diff
@@ -188,25 +195,66 @@ export const runningSessionsStatus: JupyterFrontEndPlugin<void> = {
...
+              | 'if-any'
+              | 'never'
+              | 'always'
+              | undefined) ?? 'if-any'
+          ];
 
         disposable?.dispose();
-        if (showStatusBarItem) {
-          disposable = registerItem();
+        if (showKernels || showTerminals) {
```
krassowski, line 220:
```suggestion
        if (showKernels || showTerminals !== false) {
```

packages/apputils-extension/src/index.ts
```diff
@@ -728,10 +728,24 @@ const sanitizer: JupyterFrontEndPlugin<IRenderMime.ISanitizer> = {
...
+/*
+ * A plugin owning the kernel settings
+ */
+export const kernelSettings: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/apputils-extension:kernels-settings',
+  description: 'Reserves the name for kernel settings.',
+  autoStart: true,
+  requires: [ISettingRegistry],
+  activate: (_app: JupyterFrontEnd, settingRegistry: ISettingRegistry) => {
+    void settingRegistry.load(kernelSettings.id);
```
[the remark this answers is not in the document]
krassowski: No the `load()` call is not strictly necessary. That said if settings schema name was changed this should make the tests fail. I think there is no harm in leaving it in, and then for 4.5 we would use the token approach (where this plugin would be responsible for setting the `showKernels` value on a new token).

packages/apputils-extension/src/statusbarplugin.ts
krassowski: Thank you, 08d2daa should fix it

packages/apputils-extension/src/statusbarplugin.ts
brichet: As a reminder, for a future PR with the token approach, we could define the `showTerminals` as `boolean | 'if-any'` to clarify the state, instead of using `underfined` as a third boolean state.

brichet (approved):
Looks good, thanks @krassowski.

Issue #17386, opened 2025-03-12
Too many "kernel" lines in the tooltip


krassowski:
## Description

![Image](https://github.com/user-attachments/assets/c8452312-70e0-4905-b3c9-00f6f026377c)

## Reproduce

1. Create a new notebook
2. Hover over the notebook in file browser

## Expected behavior

Only one line

## Context

- Browser and version: Chrome
- JupyterLab version: 4.4.0b0

krassowski:
The hover text gets created here:

https://github.com/jupyterlab/jupyterlab/blob/e5dbd320e591a50647d614da2a4e6daaf99ce3d7/packages/filebrowser/src/listing.ts#L3230-L3268

but this does not include the `Kernel: ` part which only gets added in:

https://github.com/jupyterlab/jupyterlab/blob/e5dbd320e591a50647d614da2a4e6daaf99ce3d7/packages/filebrowser/src/listing.ts#L1007-L1024

PR #17421, opened 2025-03-21
Fix too many kernel info in filebrowser tooltip

itsmevichu:
Fixes - https://github.com/jupyterlab/jupyterlab/issues/17386


packages/filebrowser/src/listing.ts
```diff
@@ -968,6 +968,7 @@ export class DirListing extends Widget {
   }
 
   // Update item nodes based on widget state.
+  private _lastRenderedState = new WeakMap<HTMLElement, string>();
```
krassowski, line 971: Can you move it down next to other private properties?

krassowski (approved):
Thank you @itsmevichu!

> I noticed that no addition to the tooltip was made when testing on Binder, but this was due to https://github.com/jupyterlab/jupyterlab/issues/17425.

Issue #17349, opened 2025-02-27
DocumentSearch extension finds matches within unsupported element leading to disappearing text


achhina:
## Description

When you have an embedded `<svg>` tag within your cell output the application search matches content inside the tag. This results in an unsupported `<mark>` tag being wrapped around the text, which doesn't render and makes your content appear to have missing text. I've attached a small video as an example with a developer build of the most recent commit off main.

https://github.com/user-attachments/assets/c9885a5c-7259-499b-88ef-3b23eb2e8c46

## Reproduce

You can reproduce by embedding a `svg` tag with some text, like the following cell.

```python
from IPython.display import display, HTML

svg = """
<svg height="30" width="200" xmlns="http://www.w3.org/2000/svg">
  <text x="10" y="20" font-family="Arial" font-size="16" fill="white">Some Text</text>
</svg>
"""

display(HTML(svg))
```

This was originally found by rendering a Plotly plot with a `notebook` renderer, and a user trying to search for the name of a subplot title. Unfortunately it took the user quite some time to realize that there was not an issue with how they were rendering their plot, but with their searched query stripping the text from their plot.

## Expected behavior

Ideally I'd expect the searched text within a `svg` to be wrapped with a `tspan` and some CSS fill styles to emulate a `mark` style. Otherwise, I'd expect an unsupported element to not be matched. However, it does appear that `SVG` is listed as an unsupported element in the `HTMLSearchEngine` class?

https://github.com/jupyterlab/jupyterlab/blob/1c64e5e00c3842eb7694190ed80a31685dcdeffa/packages/documentsearch/src/providers/genericsearchprovider.ts#L67

## Context

- Operating System and version: macOS Sequoia 15.2 - Intel chip
- Browser and version: Firefox 135.0.1
- JupyterLab version: 4.4.0b0

<details><summary>Troubleshoot Output</summary>
<pre>
❯ jupyter troubleshoot

[notice] A new release of pip is available: 24.0 -> 25.0.1
[notice] To update, run: pip install --upgrade pip
$PATH:
        /Users/achhina/projects/jupyterlab/.venv/bin
        /Users/achhina/bin
        /Users/achhina/.pyenv/shims
        /Users/achhina/.nix-profile/bin
        /nix/var/nix/profiles/default/bin
        /usr/local/bin
        /System/Cryptexes/App/usr/bin
        /usr/bin
        /bin
        /usr/sbin
        /sbin
[...]

PR #17429, opened 2025-03-26
Fix document search matching text within unsupported SVG tags

achhina:
## References
Fixes https://github.com/jupyterlab/jupyterlab/issues/17349

## Code changes

I added `svg` in lowercase to the list of unsupported elements, along with some tests and comments that explain why checking the uppercase qualified name for SVG's is not always sufficient to identify an element.
## User-facing changes

Previously search candidates would match and be highlighted within SVGs which lead to the fragment not rendering and appearing as if it was disappearing.

The disappearing text video can be seen in the original issue, and the following is a video with the proposed changes.

https://github.com/user-attachments/assets/8b01b699-1a7a-43fc-be75-e06476d61870

I don't believe there should be any backward incompatible changes, however would it make sense to backport this fix?

### DIFF
--- a/packages/documentsearch/src/providers/genericsearchprovider.ts
+++ b/packages/documentsearch/src/providers/genericsearchprovider.ts
@@ -20,6 +20,17 @@ export class HTMLSearchEngine {
   /**
    * We choose opt out as most node types should be searched (e.g. script).
    * Even nodes like <data>, could have textContent we care about.
+   *
+   * Note:
+   * We will be checking each node's nodeName attribute in the Node interface.
+   * ref: https://dom.spec.whatwg.org/#dom-node-nodename
+   *
+   * Specifically for Element nodes, where nodeName is stated as the
+   * 'HTML-uppercased qualified name'. However, that does not mean that HTML
+   * elements qualified name is guaranteed to be uppercased even when the
+   * content type is HTML, i.e. XML tags like <svg>. This only applies when the
+   * node's namespace is in HTML and the node document is a HTML document.
+   * ref: https://dom.spec.whatwg.org/#element-html-uppercased-qualified-name
    */
   static UNSUPPORTED_ELEMENTS = {
     // https://developer.mozilla.org/en-US/docs/Web/HTML/Element#Document_metadata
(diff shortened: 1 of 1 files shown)


krassowski:
Thank you @achhina for investigating and proposing a fix.

What do you think about instead upper-casing `parentElement.nodeName` in:

https://github.com/jupyterlab/jupyterlab/blob/85c82eba1caa7e28a0d818c0840e13756c1b1256/packages/documentsearch/src/providers/genericsearchprovider.ts#L97

to make it a generic fix for any element? Or do you think it only affects SVGs?

achhina:
> Thank you @achhina for investigating and proposing a fix.
> 
> What do you think about instead upper-casing `parentElement.nodeName` in:
> 
> https://github.com/jupyterlab/jupyterlab/blob/85c82eba1caa7e28a0d818c0840e13756c1b1256/packages/documentsearch/src/providers/genericsearchprovider.ts#L97
> 
> to make it a generic fix for any element? Or do you think it only affects SVGs?

Thank you for reviewing @krassowski. I originally added a separate function that does this check, which takes in a `Node` and transforms `Node.nodeName` to uppercase before checking within the unsupported elements. I ended up removing it due to some concerns of performance and it being invoked incrementally on every node in large notebooks. However, thinking about this some more it was probably premature due to how cheap the transform is and function inlining. Which also lines up with a rough test I just did for a large notebook with over 20k elements and seeing negligible difference.

I went ahead and added the function back, and opted for a function to save having to remember to do this uppercasing if in the future additional checks would be required for unsupported elements. Ideally I would have liked to mark `UNSUPPORTED_ELEMENTS` private to force using `isUnsupportedNode`, but considering it's public and exported I think that would count as public API breakage? However, happy to remove the function and perform the uppercase within the `search` function if that's preferred!

krassowski:
> Ideally I would have liked to mark UNSUPPORTED_ELEMENTS private to force using isUnsupportedNode, but considering it's public and exported I think that would count as public API breakage? However, happy to remove the function and perform the uppercase within the search function if that's preferred!

You could follow the following pattern:

```ts
namespace Private {
  export function isUnsupportedNode() {
  }
}
```

and then use it as `Private.isUnsupportedNode()`.

I think your concern about performance is a valid one.

Looking more into it, the MDN documentation make it clear that `nodeName` is:

> The value of [Element.tagName](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName), that is the uppercase name of the element tag if an HTML element, or the lowercase element tag if an XML element (like a SVG or MathML element).

It looks like searching in MathML does not break it, though highlighting only works partially:

image

Maybe a better approach from the user point of view would be to not reject on SVGs but fix the `<mark>` to be valid in a given context?

krassowski:
I have now more confidence in the `svg: true` approach as a short-term fix, so if you prefer to revert back to it I would be happy to merge this PR now. Still, it looks like we could support correct highlighting for SVGs and MathML, which we can track separately.

achhina:
> I have now more confidence in the `svg: true` approach as a short-term fix, so if you prefer to revert back to it I would be happy to merge this PR now. Still, it looks like we could support correct highlighting for SVGs and MathML, which we can track separately.

Reverted! Agreed about supporting highlighting for these tags - would you know if there is an existing issue to provide support for those two, and if someone is already working on that?

> > Ideally I would have liked to mark UNSUPPORTED_ELEMENTS private to force using isUnsupportedNode, but considering it's public and exported I think that would count as public API breakage? However, happy to remove the function and perform the uppercase within the search function if that's preferred!
> 
> You could follow the following pattern:
> 
> ```ts
> namespace Private {
>   export function isUnsupportedNode() {
>   }
> }
> ```
> 
> and then use it as `Private.isUnsupportedNode()`.
> 

Ah, I think this is moot now after the revert, but I had wanted to make `UNSUPPORTED_ELEMENTS` private to force usage of `isUnsupportedNode`.

krassowski (approved):
Thank you @achhina!

krassowski:
> would you know if there is an existing issue to provide support for those two, and if someone is already working on that?

No, I believe there is no other issue for that, feel welcome to create it and work on it :)

achhina:
> > would you know if there is an existing issue to provide support for those two, and if someone is already working on that?
> 
> 
> 
> No, I believe there is no other issue for that, feel welcome to create it and work on it :)

https://github.com/jupyterlab/jupyterlab/issues/17459 - done!

I noticed the CI failed here, but it appears to be for workspaces and unrelated?

krassowski:
Yes, it was unrelated - it passed on restart :)

krassowski:

PR #17430, opened 2025-03-26
Make `ILabShell` optional in the logconsole extension

jtpio:
## References

Noticed when trying to add support for the `down` area in Notebook 7.

Current the logconsole plugin has a hard requirement on `ILabShell`, which makes reusing the plugin downstream difficult.

## Code changes

Handle the case where `ILabShell` is missing.

## User-facing changes

None for JupyterLab users.

Chances to get the log console to work for Notebook users.

Using this patch downstream in Notebook:

image

None

### DIFF
--- a/packages/logconsole-extension/src/index.tsx
+++ b/packages/logconsole-extension/src/index.tsx
@@ -17,6 +17,7 @@ import {
   WidgetTracker
 } from '@jupyterlab/apputils';
 import { IChangedArgs } from '@jupyterlab/coreutils';
+import { DocumentWidget } from '@jupyterlab/docregistry';
 import {
   ILoggerRegistry,
   LogConsolePanel,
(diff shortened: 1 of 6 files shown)


packages/logconsole-extension/src/index.tsx
```diff
@@ -118,6 +125,18 @@ function activateLogConsole(
     translator
   });
 
+  const getCurrentWidgetPath = () => {
+    const currentWidget = app.shell.currentWidget;
+    if (labShell?.currentPath) {
+      return labShell.currentPath;
+    }
+    // For other shells, set the source to the current widget path
+    if (currentWidget && currentWidget instanceof DocumentWidget) {
```
jtpio, line 134: For reference, this follows the same approach used by `ILabShell`:

https://github.com/jupyterlab/jupyterlab/blob/85c82eba1caa7e28a0d818c0840e13756c1b1256/packages/application/src/shell.ts#L1424-L1426

krassowski (approved):
Thank you @jtpio!

PR #17437, opened 2025-03-27
Fix usage of `ITerminalAPIClient` in `TerminalConnection`

jtpio:
## References

As noticed in https://github.com/jupyterlite/terminal/pull/49#issuecomment-2758684746 thanks to @ianthomas23.

## Code changes

Fix usage of `ITerminalAPIClient` in `TerminalConnection`

## User-facing changes

None

None

### DIFF
--- a/packages/services/src/terminal/default.ts
+++ b/packages/services/src/terminal/default.ts
@@ -10,7 +10,7 @@ import { ISignal, Signal } from '@lumino/signaling';
 import { ServerConnection } from '..';
 
 import * as Terminal from './terminal';
-import { shutdownTerminal, TERMINAL_SERVICE_URL } from './restapi';
+import { TerminalAPIClient } from './restapi';
 
 /**
  * An implementation of a terminal interface.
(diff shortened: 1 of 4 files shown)


packages/services/src/terminal/default.ts
```diff
@@ -381,16 +388,10 @@ export class TerminalConnection implements Terminal.ITerminalConnection {
...
   private _reconnectAttempt = 0;
   private _pendingMessages: Terminal.IMessage[] = [];
+  private _terminalAPIClient: Terminal.ITerminalAPIClient;
 }
 
 namespace Private {
-  /**
-   * Get the url for a terminal.
-   */
-  export function getTermUrl(baseUrl: string, name: string): string {
```
jtpio, line 390: Removing this private function as it doesn't seem to be used anywhere in the code base.

ianthomas23 (approved):
I've confirmed locally that this fixes the problem with the JupyterLite terminal. Thanks @jtpio.

krassowski (approved):
Thank you @jtpio and @ianthomas23!

jtpio:
Nice, thanks for the review!

Issue #17427, opened 2025-03-26
No visual indication where the selection would be dropped when dragging in a code editor


jtpio:
## Description

Selecting text and dragging it somewhere else in a code editor does not show any visual indication where the selection would be dropped when dragging.

## Reproduce

Select some text in a code editor (file editor or notebook cell), and try to move it elsewhere:

https://github.com/user-attachments/assets/0913abf2-d2d0-4a2c-b697-76bef247b79f

## Expected behavior

Trying with the Codemirror demo on https://codemirror.net/try/.

We see the cursor moves as the selection is being moved:

https://github.com/user-attachments/assets/59abf480-b4f6-4c4b-8b68-c6b6c0dcab5e

## Context

- Operating System and version: MacOS 15.3.2
- Browser and version: Chrome 134
- JupyterLab version: `main` / `4.4.0rc0`: 85c82eba1caa7e28a0d818c0840e13756c1b1256

MUFFANUJ:
Seems a fair problem, would love to work on this!

PR #17438, opened 2025-03-28
Fix visual indication for drag and drop in editor

MUFFANUJ:
## References
Fixes - #17427 

After changes 
The cursor moves as the selection is being moved as expected.

https://github.com/user-attachments/assets/a45dd26d-9084-474e-979f-1cb98e11fad9

### DIFF
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -25,6 +25,7 @@ import {
   IEditorLanguageRegistry,
   IExtensionsHandler
 } from './token';
+import { dropCursor } from '@codemirror/view';
 
 /**
  * The class name added to CodeMirrorWidget instances.
@@ -786,6 +787,10 @@ namespace Private {
     doc?: string
   ): EditorView {
     const extensions = editorConfig.getInitialExtensions();
+    // Check if dropCursor is enabled in the config (defaulting to true)
+    if (editorConfig.getOption('dropCursor') !== false) {
+      extensions.push(dropCursor());
+    }
     extensions.push(...additionalExtensions);
     const view = new EditorView({
       state: EditorState.create({
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -88,6 +88,7 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
         // We need to set the order to high, otherwise the keybinding for ArrowUp/ArrowDown
         // will process the event shunting our edge detection code.
         Prec.high(onKeyDown),
+        dropCursor(),
         updateListener,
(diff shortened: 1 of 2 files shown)


MUFFANUJ:
@meeseeksdev tag bug

jtpio:
Nice, thanks for working on this!

packages/codemirror/src/editor.ts
```diff
@@ -25,6 +25,7 @@ import {
   IEditorLanguageRegistry,
   IExtensionsHandler
 } from './token';
+import { dropCursor } from '@codemirror/view';
```
jtpio, line 28: Maybe this import can be grouped with the existing `@codemirror/view` import above?

packages/codemirror/src/editor.ts
```diff
@@ -87,6 +88,7 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
         // We need to set the order to high, otherwise the keybinding for ArrowUp/ArrowDown
         // will process the event shunting our edge detection code.
         Prec.high(onKeyDown),
+        dropCursor(),
```
jtpio, line 91: Wondering if this extension should be placed next to the other default extensions, for example here?

https://github.com/jupyterlab/jupyterlab/blob/5ce9e8ccc779373149b837cf0431d1748b51feeb/packages/codemirror/src/extension.ts#L604-L609

MUFFANUJ: You are right thanks for this suggestion!

packages/codemirror/src/extension.ts
```diff
@@ -776,6 +777,15 @@ export namespace EditorExtensionRegistry {
           title: trans.__('Line Wrap')
         }
       }),
+      Object.freeze({
+        name: 'dropCursor',
+        default: true,
+        factory: () => createConditionalExtension(dropCursor()),
+        schema: {
+          type: 'boolean',
+          title: trans.__('Drop Cursor')
```
krassowski, line 786: Backporting this will (should) warn about new translatable string.

I think it is a fine idea to have opt-out in case if this causes issues, though I think it would be also fine to always include. What you all think?

krassowski (approved):
Thank you @MUFFANUJ.

I will merge it into 4.4.0 but I will not backport for now as it is not clear how to approach it.

packages/codemirror/src/extension.ts
krassowski: > By opt-out would that meant not including the dropCursor extension if the string is localized in the given language?

No, I meant opt-out as having this as a setting in general, regardless of the localization. I was suggesting that not having a setting would be fine too.

PR #17440, opened 2025-03-29
Fix read-only indicator not showing in RTC docprovider

Darshan808:
### Description:
This PR fixes the issue where the read-only indicator was not displayed in the toolbar when opening read-only files in collaborative mode. Additionally, attempting to save such files did not trigger a warning popup but instead resulted in an error in the console.

### Issue:
- In collaborative mode, the read-only indicator and warning popup were blocked.

- The writable attribute was always set to false (I'm not sure why this was done)

### Fix:
The writable attribute is now determined based on the file system permissions.

The read-only indicator and warning popup are now displayed even in collaborative mode when opening read-only files.

### Note:
This fix applies to file-system read-only permissions and not user permissions. While anyone with access to the collaborative session can still modify the read-only file which are visible to other users of session, they cannot save it to disk.

### DIFF
--- a/packages/docmanager-extension/src/index.tsx
+++ b/packages/docmanager-extension/src/index.tsx
@@ -888,10 +888,7 @@ function addCommands(
           if (saveInProgress.has(context)) {
             return;
           }
-          if (
-            !context.contentsModel?.writable &&
-            !context.model.collaborative
-          ) {
+          if (!context.contentsModel?.writable) {
             let type = (args._luminoEvent as ReadonlyPartialJSONObject)?.type;
             if (args._luminoEvent && type === 'keybinding') {
               readonlyNotification(context.path);
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -481,12 +481,11 @@ export class Context<
   private _updateContentsModel(
     model: Contents.IModel | Omit<Contents.IModel, 'content'>
   ): void {
-    const writable = model.writable && !this._model.collaborative;
     const newModel: Omit<Contents.IModel, 'content'> = {
       path: model.path,
       name: model.name,
       type: model.type,
-      writable,
+      writable: model.writable,
       created: model.created,
       last_modified: model.last_modified,
       mimetype: model.mimetype,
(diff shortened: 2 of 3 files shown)


Darshan808:
> The writable attribute was always set to false (I'm not sure why this was done)

Why is this done in collaborative mode? I feel like I might be missing something.

davidbrochart:
Thanks for looking into this @Darshan808.
In RTC the write attribute is set to false because saving is automatically done in the backend. So you need to restore that for read-write files, but I think your fix is fine for read-only files. This will also need a fix in jupyter-collaboration to check if a file is read-only before trying to save (currently it fails at saving).

krassowski:
Also see https://github.com/jupyterlab/jupyterlab/pull/16900 which already added a condition on `collaborative` attribute which short-circuits saving.

> This will also need a fix in jupyter-collaboration to check if a file is read-only before trying to save (currently it fails at saving).

:+1:

krassowski:
There already is this property:

https://github.com/jupyterlab/jupyterlab/blob/af8ad58b6cc07e0589650b5f8e025a21563390ef/packages/docregistry/src/context.ts#L227-L232

For auto-save I think we would want to use it here (instead of `writable`):

https://github.com/jupyterlab/jupyterlab/blob/968b1a8989a9d37d4620ff83f3899cbff019aab4/packages/docmanager/src/savehandler.ts#L114-L119

Darshan808:
> For auto-save I think we would want to use it here (instead of writable)

Yes, this approach would be better, but the `canSave()` getter is currently `protected`. Would there be any issue in making it `public`?

Darshan808:
> This will also need a fix in jupyter-collaboration to check if a file is read-only before trying to save (currently it fails at saving).

https://github.com/jupyterlab/jupyter-collaboration/pull/457 fixes this

krassowski (approved):
This looks good to me, thank you.

As a follow-up we could add a test for save handler bailing out if model is neither collaborative nor writable somewhere here:
 https://github.com/jupyterlab/jupyterlab/blob/bb9dca346cd25530e86bd66ab338000c5e781b25/packages/docmanager/test/savehandler.spec.ts#L45

PR #17442, opened 2025-03-31
Fix CommandPalette gets re-rendered even if hidden

### LINKED ISSUE #13677: Command palette gets re-rendered even if hidden
## Description

Command palette gets re-rendered even if hidden.

## Reproduce

1. Open command palette and two tabs
2. Close the command palette
3. Go to inspector right click on command palette element, select "break on" > "subtree modifications"
4. Switch tabs
5. See that command palette fragments get re-rendered via lumino virtual DOM

Screenshot from 2022-12-29 13-23-55

It is not clear to me why this update appears. Of note the migration to `requestAnimationFrame` in lumino seems to make the debugging/profiling a bit more confusing as I don't immediately see if this was a genuine animation frame request or just lumino messaging.

## Expected behavior

Updating command palette on tab switch is expected because different commands will become available in different contexts (e.g. text editor/notebook), but re-rendering it while it is hidden is not expected.

In a bigger picture we could ask: should command palette get detached rather than hidden. I think hiding is ok because there is ever only one command palette, but requires a better hygiene on when a re-render is triggered.

## Context

- JupyterLab version: 4.0.0a31

fcollonval:
## References

Fixes #13677

## Code changes

- Bump lumino to include the new algorithm of `Widget.isVisible` to be consistent if the widget is within an hidden widget
- Simplify the code of `ListWidget` in `@jupyterlab/running` that was hitting a similar issue.

## User-facing changes

Less loaded application - but the gain should not be that much

None


fcollonval:
I set the target to 4.4.0. Should we backport it to 4.3.x dispite the bump to a new minor version of `@lumino/wdigets` - although it does have the same API?

krassowski:
The previous `@lumino/widgets` version had some API changes so I would not pull it into 4.3.x, but I think we can still merge it into 4.4.0.

It looks like the integrity test is failing.

krassowski (approved):
Thank you @fcollonval!

fcollonval:
Thanks for fixing the integration test @krassowski

PR #17444, opened 2025-04-01
Do not try to prettier `.mypy_cache`

krassowski:
## References

Running `jlpm lint` can take ages if there is `.mypy_cache` in any of the subdirectories. This adds it to ignored patterns.

## Code changes

None

## User-facing changes

None

None


jtpio (approved):
Thanks!

PR #17457, opened 2025-04-08
Fix `@jupyterlab/services` import

jtpio:
## References

Noticed while browsing the code base.

## Code changes

Import from `@jupyterlab/services` instead of `@jupyterlab/services/src/kernel/kernel', for consistency with the rest of the code base.

## User-facing changes

None

None


krassowski (approved):
Thank you @jtpio!

PR #17466, opened 2025-04-10
Fix incorrect modifications in collaborative save events

### LINKED ISSUE #17462: Tab title on hover do not correspond to the right notebook
## Description

Tab title on hover do not correspond to the right notebook

## Reproduce

1. Install jupyterlab 4.4.0 and jupyter-collaboration 4.0
2. Create a few notebooks
3. Hover over the tabs
4. See the path of the notebook is wrong

Good:

Image

Wrong (wrong directory and name)

Image

Also wrong (wrong directory)

Image

This does not seem to happen when `jupyter-collaboration` is not installed.

## Expected behavior

Titles are pointing to the correct notebook.

## Context

- Browser and version: Chrome
- JupyterLab version: 4.4.0

Darshan808:
## Fixes #17462

https://github.com/jupyterlab/jupyterlab/pull/16695 introduced an enhancement to update the contents model of a file when the hash value is changed by the server. However, the signal is currently broadcasted to all open files, causing every file’s contents model to be overwritten with the data from the newly saved file.

This leads to several issues:
- Incorrect filenames shown on hover
- Unexpected behavior in read-only files
- Other potential inconsistencies in code which relies on contents model of files

This PR addresses the problem by adding a check to ensure that the signal corresponds to the correct file. If the file does not match, the contents model will not be updated.

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -418,6 +418,11 @@ export class Context<
     change: Contents.IChangedArgs
   ): void {
     if (change.type === 'save' && this._model.collaborative) {
+      //skip if the change isn't related to current file
+      if (this._contentsModel?.path !== change.newValue?.path) {
+        return;
+      }
+
       // Update the contents model with the new values provided on save.
       // This is needed for save operations performed on the server-side
       // by the collaborative drive which needs to update the `hash`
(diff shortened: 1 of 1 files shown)


packages/docregistry/src/context.ts
```diff
@@ -418,6 +418,11 @@ export class Context<
     change: Contents.IChangedArgs
   ): void {
     if (change.type === 'save' && this._model.collaborative) {
+      //skip if the change isn't related to current file
```
krassowski, line 421:
```suggestion
      // Skip if the change isn't related to current file.
```

krassowski (approved):
Thank you @Darshan808!

krassowski:

PR #17469, opened 2025-04-12
Follow changes to stream outputs

davidbrochart:
## References

Together with https://github.com/jupyter-server/jupyverse/pull/480, this fixes server-side execution in Jupyverse (but it's a general bug fix).

## Code changes

Changes to a cell stream output must be observed if initially present in the model, not just when a new stream output is added.

## User-facing changes

In particular, this fixes server-side execution, e.g. when reloading a page with a long-running cell execution.

None.

### DIFF
--- a/packages/outputarea/src/widget.ts
+++ b/packages/outputarea/src/widget.ts
@@ -119,6 +119,17 @@ export class OutputArea extends Widget {
     ) {
       const output = model.get(i);
       this._insertOutput(i, output);
+      if (output.type === 'stream') {
+        // This is a stream output, follow changes to the text.
+        output.streamText!.changed.connect(
+          (
+            sender: IObservableString,
+            event: IObservableString.IChangedArgs
+          ) => {
+            this._setOutput(i, output);
+          }
+        );
+      }
     }
     model.changed.connect(this.onModelChanged, this);
     model.stateChanged.connect(this.onStateChanged, this);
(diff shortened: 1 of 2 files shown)


davidbrochart:
It looks like UI tests fail on `main` too?

krassowski:
> It looks like UI tests fail on `main` too?

I guess the problematic test is very flaky as they passed yesterday:

image

I opened https://github.com/jupyterlab/jupyterlab/issues/17479 to track it

davidbrochart:
https://github.com/jupyterlab/jupyterlab/pull/17469/commits/9ac19e4aedcc4a3995ce1b050e7040c06e2d76f3 adds a test that fails on `main` (see #17481), and passes here.

krassowski (approved):
Thank you @davidbrochart!

packages/outputarea/src/widget.ts
```diff
@@ -119,6 +119,17 @@ export class OutputArea extends Widget {
...
       const output = model.get(i);
       this._insertOutput(i, output);
+      if (output.type === 'stream') {
+        // This is a stream output, follow changes to the text.
+        output.streamText!.changed.connect(
+          (
+            sender: IObservableString,
+            event: IObservableString.IChangedArgs
+          ) => {
+            this._setOutput(i, output);
```
cben, line 129: Is it safe to refer to numeric `i` here, or should it be some kind of stable reference?  
This runs later as output arrives/changes(?), meanwhile user could insert/delete cells above current cell?  [tried that, observed no bugs]

But hmm this is not indexing of cells, right?  Is it indexing lines of output?  Chunks of output of given stdout/stderr name?

Issue #17460, opened 2025-04-08
[a11y][Violation][1.1.1 Non-text Content] Multiple SVG elements has no accessible names


nkn2022:
## Description

I am running an accessibility test suite with Jupyter using the **IBM Equal Access Accessibility Checker** tool on Chrome browser. Few issues found on the `data-icon="ui-components:XX` are missing to reach a first formal level of compliance. So kindly help look into these. 

Issues reported were violating **WCAG 2.2** requirements. 
Ref: https://www.ibm.com/able/requirements/checker-rule-sets

Listing of violations categorized based on the checkpoint. 

****Violation of Rule:** `svg_graphics_labelled`**

> List of issues with elements that violate this rule
> > **Issue 1** 
> > The SVG element has no accessible name
> > **Sample code references(not showing all):** 
> > `<svg class="" data-icon="ui-components:save" width="32" viewBox="0 0 32 32" height="32" xmlns="http://www.w3.org/2000/svg">`
> > `<svg data-icon="ui-components:caret-down-empty" viewBox="0 0 18 18" width="16" xmlns="http://www.w3.org/2000/svg">`
> > `<svg data-icon-id="02358d44-6b12-4c0a-96fc-3c3c8147d4c4" data-icon="ui-components:ellipses" viewBox="0 0 24 24" width="16" xmlns="http://www.w3.org/2000/svg">`
...

Possible Fixes: 

> What to do
> Ensure the non-decorative SVG element has an accessible name that is not empty:
> 
> Add an aria-labelledby attribute to the element that points to visible text on the page that is a meaningful label.
> Or, add an aria-label attribute to the element.
> Or, add a direct child <title> element.
> Or, add an xlink:title attribute on a link.
> Or, for text container elements, add the text content.
> Or, only if the design cannot have a visible label, use the title attribute to provide a label.

## Reproduce

- Install Chrome extension for IBM accessibility checker tool. Ref: https://www.ibm.com/able/toolkit/verify/automated
- Open JupyterLab UI on Chrome browser. Open `notebook` file to edit.
- Run the tool and we can see the above `Violation`s reported.

## Expected behavior

No violations are reported on the scanning to reach a first formal level of compliance.

## Context

- Operating System and version: MacOS Sonoma Version 14.7.3
- Browser and version: Chrome Version 134.+
- JupyterLab version: Version 4.3.6

krassowski:
[...]

PR #17472, opened 2025-04-14
Add `aria-hidden` to icons (`<svg>` elements) which do not have a `title`

nkn2022:
## References

-  Fixes [[a11y][Violation][1.1.1 Non-text Content] Multiple SVG elements has no accessible names](https://github.com/jupyterlab/jupyterlab/issues/17460)

## Code changes

`LabIcon` generates `svg` elements that do not have an accessible name. All the reported issues are on the images that are decorative so should include `aria-hidden = true` to mark the them decorative. [Ref](https://www.w3schools.com/accessibility/accessibility_meaningful_images.php)

## User-facing changes

No changes.

NA


packages/ui-components/src/icon/labicon.tsx
```diff
@@ -581,6 +581,9 @@ export class LabIcon implements LabIcon.ILabIcon, VirtualElement.IRenderer {
 
       if (title) {
         Private.setTitleSvg(svgElement, title);
+      } else {
+        // mark as a decorative
+        svgElement.setAttribute('aria-hidden', 'true');
```
krassowski, line 586: The CI fails with:

```
error >> Warning: Invalid ARIA attribute `%s`. Did you mean `%s`?%s ariaHidden aria-hidden 
```

Not sure why yet

nkn2022: Sorry, my mistake, the warning appears after this change. Something was messed up in my local cache. 
Not really sure why this happens.

krassowski: After digging a bit, I think it has to do with this logic:

https://github.com/jupyterlab/jupyterlab/blob/cf8fb54f3cf6c8a46c9006720f5e4b15eff76814/packages/ui-components/src/icon/labicon.tsx#L491-L510

https://github.com/jupyterlab/jupyterlab/blob/cf8fb54f3cf6c8a46c9006720f5e4b15eff76814/packages/ui-components/src/icon/labicon.tsx#L405-L418

https://github.com/jupyterlab/jupyterlab/blob/cf8fb54f3cf6c8a46c9006720f5e4b15eff76814/packages/ui-components/src/utils.ts#L62-L90

I think if you add a  special clause in `getReactAttrs` to not convert any attribute starting with `aria-` to camel case it should work, as react expects all ARIA attributes to be hyphen-cased rater than camelCase (https://legacy.reactjs.org/docs/accessibility.html#wai-aria).

nkn2022:
@krassowski Thank you for the detailed review. Sorry for the delayed response. My points below.

Extension owners already have an option to set `title`. If this is not set we will assume it is decorative. I think this should suffice. 

For the edge cases you mentioned, I think it will be good to tackle as a separate issue as they address different aspect of accessibility. 

I checked the toggle menu icon and see that it is treated as decorative and have `aria-hidden` property set with this PR which is correct. 
image

Let me know if we can proceed.

krassowski (approved):
Thank you @nkn2022! I opened https://github.com/jupyterlab/jupyterlab/issues/17501 to track the menu checkmark issue; I will merge this PR into 4.5.0 branch so that there is time to tackle the menu checkmark before release.

krassowski:
I think https://github.com/jupyterlab/jupyterlab/issues/17501 is only not reported as violation because of the limitation of the automated checker, right? Having no testing from actual users with accessibility needs I do not feel comfortable backporting this without more testing just yet, but once we get more testing I am happy to backport to 4.x branch.

nkn2022:
@krassowski I think I understand your concern. The `checkmark` here has a meaning, it is not just a decorative icon. correct? 
If that is the case, let us not merge this PR. We should look into an alternative fix or a way to handle this scenario. What do you say?

nkn2022:
@krassowski We can fix this by adding `aria-selected="true/false"` to the item with `role="menuitem"`. This roles will have selection state. Need to find a solution for this.

krassowski:
Yes, the fix could go in `lumino` or in this repo since we are overriding it here:

https://github.com/jupyterlab/jupyterlab/blob/f0585d9aa4c55def95df05febf6b2dda3dba58e7/packages/ui-components/src/icon/widgets/menusvg.ts#L96

https://github.com/jupyterlab/jupyterlab/blob/f0585d9aa4c55def95df05febf6b2dda3dba58e7/packages/ui-components/src/icon/widgets/menusvg.ts#L170-L173

nkn2022:
> Yes, the fix could go in `lumino` or in this repo since we are overriding it here:
> 
> https://github.com/jupyterlab/jupyterlab/blob/f0585d9aa4c55def95df05febf6b2dda3dba58e7/packages/ui-components/src/icon/widgets/menusvg.ts#L96
> 
> https://github.com/jupyterlab/jupyterlab/blob/f0585d9aa4c55def95df05febf6b2dda3dba58e7/packages/ui-components/src/icon/widgets/menusvg.ts#L170-L173

@krassowski I see that this api is creating the div that contains the `svg` element. 
image

We need to add`aria-selected` property to its parent element `<li>` which has the role `'menuitem'`. So this change should go into the lumino repo. Specifically, this section of the API, where the aria attributes are updated. 
https://github.com/jupyterlab/lumino/blob/0e5c140cd2aaa9b176179420602297efad815584/packages/widgets/src/menu.ts#L1368-L1372

nkn2022:
@krassowski In my testing the fix with my mac voice over and I did not see that it was announcing the selection with `aria-selected =true` setting on the `menuitem`. Reading up some more on this issue, I found that the right way to fix this is to use a different 'role' itself for such items. 
Ref: 
https://www.w3.org/TR/wai-aria/states_and_properties#introstates
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemcheckbox_role

I can create a PR for the [issue](https://github.com/jupyterlab/jupyterlab/issues/17501) based on this and you can review it and see if this works.

nkn2022:
@krassowski Should we wait until we get the [lumino fix](https://github.com/jupyterlab/lumino/pull/753) into this PR? Or can we merge this now?

krassowski:
I think we can merge and just coordinate that this and the lumino update gets into the same release. I will look into getting lumino fix released tomorrow

nkn2022:
Sounds good. Thanks.

krassowski:
For the record, I released a new `lumino` version https://github.com/jupyterlab/lumino/releases/tag/v2025.4.30 and will open a PR to update the versions used by jupyterlab tomorrow.

Issue #10422, opened 2021-06-15
extra_labextensions_path ordering


mlucool:
### Problem

I have a set of extensions that can be found in a read-only path that I set in `extra_labextensions_path` (love this feature). Sometimes I one want do development on one of them, without losing the ability to use others (of course, I could reinstall). Following the prebuilt directions, I use `jupyter-labextension develop . --overwrite` to link my extension in my venv. If run lab without `extra_labextensions_path` everything works as expected. If I then use `extra_labextensions_path`, my extension gets shadowed with the non-dev version.

### Proposed Solution

There should be some way to enforce ordering on loading (maybe there is and I am not aware). One idea I tried was `
--LabApp.extra_labextensions_path=/path/to/extras/share/jupyter --LabApp.extra_labextensions_path=/path/to/dev/share/jupyter` (the latter is already loaded by default by lab). I suspect there are a lot of good solutions here if we add a flag of some type.

mlucool:
I tried this and it did not work. I set a breakpoint at jupyterlab_server/config.py:42 and only saw it called once (which is surprising if this is how it must get the list) and
```
/jupyterlab/commands.py(1183)_get_app_info()
-> info["federated_extensions"] = get_federated_extensions(self.labextensions_path)
```
passed in `[]`.

Did you try setting a breakpoint? Maybe my setup is wrong

PR #17476, opened 2025-04-16
Reorder `extra_labextensions_path` to favor defaults

peytondmurray:
## References

Closes https://github.com/jupyterlab/jupyterlab/issues/10422

## Code changes

This PR changes the order of precedence of the default labextension paths and the optional `extra_labextensions_path`.

## User-facing changes

While this seems like a minor change, `jupyterlab_server` [uses this ordering to get a list of the federated extensions](https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L40), and when two extensions have the same name, only the first one is returned. What this means is that if a user has configured `extra_labextensions_path`, and that path contains an extension which is also in any of the default locations, the one on the `extra_labextensions_path` is not used.

This is useful in particular in cases where the user has configured `extra_labextensions_path`, but then wants to go develop one of their extensions and runs `jupyter labextension develop . --overwrite`. This will create a symlink in the default location, meaning that with this change the extension that is under development will take precedence and the one  in `extra_labextensions_path` with the same name will be ignored.


peytondmurray:
Going to mark this as ready for review, but I think I'll also write a test for `jupyter_server` to ensure that the overwrite behavior I mentioned above is tested there.

krassowski (approved):
Thank you @peytondmurray!

PR #17477, opened 2025-04-16
Fix token extraction for incomplete paths

DenisaCG:
The token returned by the `getTokenAt` function when looking at a syntax tree ending in an incomplete path is erroneous, returning a warning sign as the type, an empty value and the already given offset: 
```
{value: '', offset: 8, type: '⚠' }
```

This causes the path auto-completion in `jupyterlab-lsp` to not consider the already typed prefix when inserting the selected completion item, as the incorrect token computed when fetching the `LSP` suggestions [here](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/4628ce6bfcf1910c98b6f575d0e89e4f805c7670/packages/jupyterlab-lsp/src/features/completion/provider.ts#L198-L218), causes the completions to not have their `insertText` recomputed to eliminate the overlap with the token prefix [here](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/4628ce6bfcf1910c98b6f575d0e89e4f805c7670/packages/jupyterlab-lsp/src/features/completion/provider.ts#L322-L418).

## References
Fixes: https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1119.

## Code changes
An extra check was added to the `getTokenAt` function in the `codemirror` package. If the relevant leaf found from the syntax tree is erroneous, so it has the name as  `⚠` and the properties `to` and `from` are equal,  we look at the parent of the node, and if it exists, we consider it the new current node to compute the token from.

This solution was chosen as the syntax tree of the lines of code ending with an incomplete path create a child of the last node as the error one shown above.  

## User-facing changes

**Before** - when choosing a completion for the path, the typed prefix is not considered:

[Before.webm](https://github.com/user-attachments/assets/4c48726d-f3bd-49bb-918f-057a84b6950e)

**After** - when choosing a completion for the path, the typed prefix is considered:

[After.webm](https://github.com/user-attachments/assets/35587a46-a545-4b82-9813-0024b319a352)

None.

cc: @trungleduc

### DIFF
--- a/packages/codemirror/src/editor.ts
+++ b/packages/codemirror/src/editor.ts
@@ -534,10 +534,14 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
           }
           // If the relevant leaf token has been found, stop iterating.
           if (offset >= ref.from && offset <= ref.to) {
+            let currentNode = ref;
+            if (ref.name === '⚠' && ref.from === ref.to && ref.node.parent) {
+              currentNode = ref.node.parent;
+            }
             token = {
-              value: this.state.sliceDoc(ref.from, ref.to),
-              offset: ref.from,
-              type: ref.name
+              value: this.state.sliceDoc(currentNode.from, currentNode.to),
+              offset: currentNode.from,
+              type: currentNode.name
             };
             return false;
           }


krassowski:
Ready for review?

DenisaCG:
> Ready for review?

Yes, thanks for the suggestions.

packages/codemirror/src/editor.ts
```diff
@@ -534,10 +534,16 @@ export class CodeMirrorEditor implements CodeEditor.IEditor {
           }
           // If the relevant leaf token has been found, stop iterating.
           if (offset >= ref.from && offset <= ref.to) {
+            let currentNode = ref;
+            // The syntax tree of the code lines ending with an incomplete path create an erronous
```
trungleduc, line 538:
```suggestion
            // The syntax tree of the code lines ending with an incomplete string creates an erronous
```

krassowski (approved):
Thank you @DenisaCG!

Issue #17478, opened 2025-04-17
Add an `export` method to `NbConvertManager`


jtpio:
Currently the `@jupyterlab/notebook-extension:export` plugin gets the nbconvert URL here:

https://github.com/jupyterlab/jupyterlab/blob/cf437ae54ca7ef920059605ac8e406fb5452a692/packages/notebook-extension/src/index.ts#L612-L616

But the `PageConfig.getNBConvertURL()` function does not seem to be doing much, and mostly makes use of the `baseUrl` page config option to construct the URL:

https://github.com/jupyterlab/jupyterlab/blob/cf437ae54ca7ef920059605ac8e406fb5452a692/packages/coreutils/src/pageconfig.ts#L210-L229

So instead of first creating the `url`, and then calling `window.open` from the plugin, maybe it could make sense to move the URL creation and download logic to `NbConvertManager`:

https://github.com/jupyterlab/jupyterlab/blob/cf437ae54ca7ef920059605ac8e406fb5452a692/packages/services/src/nbconvert/index.ts#L18

Instead of the two steps, maybe a single `export` method would be enough. The method would then become available from plugins via the following:

```ts
app.serviceManager.nbconvert.export({format, path})
```

`PageConfig.getNBConvertURL()` would be kept for backwards compatibility, but marked as deprecated.

PR #17482, opened 2025-04-17
Add an optional `exportAs` method to `NbConvert.IManager`

jtpio:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/17478

## Code changes

- [x] Add an optional `export` method to `NbConvert.IManager`
- [x] Update `@jupyterlab/notebook-extension:plugin` accordingly
- [x] Mention the change of the `NbConvert.IManager` definition in the migration guide, which does not directly extend `NbConvertManager` anymore

## User-facing changes

None for JupyterLab users.

But this will allow adding in-browser exporters to JupyterLite, as mentioned in https://github.com/jupyterlite/jupyterlite/pull/1625#issuecomment-2812961794

The optional `export` method should not be breaking.

However this PR also take the opportunity to fix the definition of the `NbConvert.IManager` interface, so downstream that want to implement a `NbConvert.IManager` don't need to provide the exact `NbConvertManager`. This is likely fine as unlikely to be extended by third-party extensions at the moment, but still worth documenting.

### DIFF
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -603,7 +603,7 @@ export const exportPlugin: JupyterFrontEndPlugin<void> = {
           ? trans.__('Save and Export Notebook: %1', formatLabel)
           : formatLabel;
       },
-      execute: args => {
+      execute: async args => {
         const current = getCurrent(tracker, shell, args);
 
         if (!current) {
@@ -621,15 +621,10 @@ export const exportPlugin: JupyterFrontEndPlugin<void> = {
         };
 
         if (context.model.dirty && !context.model.readOnly) {
-          return context.save().then(() => {
-            void services.nbconvert.exportAs?.(exportOptions);
-          });
+          await context.save();
         }
 
-        return new Promise<void>(resolve => {
-          void services.nbconvert.exportAs?.(exportOptions);
-          resolve(undefined);
-        });
+        return services.nbconvert.exportAs?.(exportOptions);
       },
       isEnabled
     });
--- a/packages/notebook-extension/src/index.ts
+++ b/packages/notebook-extension/src/index.ts
@@ -93,6 +93,7 @@ import {
   IRenderMime,
   IRenderMimeRegistry
 } from '@jupyterlab/rendermime';
+import { NbConvert } from '@jupyterlab/services';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { IStateDB } from '@jupyterlab/statedb';
 import { IStatusBar } from '@jupyterlab/statusbar';
@@ -611,10 +612,12 @@ export const exportPlugin: JupyterFrontEndPlugin<void> = {
 
         const { context } = current;
 
-        const exportOptions = {
+        const exportOptions: NbConvert.IExportOptions = {
           format: args['format'] as string,
           path: current.context.path,
-          download: true
+          exporterOptions: {
+            download: true
+          }
         };
 
         if (context.model.dirty && !context.model.readOnly) {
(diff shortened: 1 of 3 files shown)


packages/services/src/nbconvert/index.ts
```diff
@@ -76,6 +81,32 @@ export class NbConvertManager {
...
 
+  /**
+   * Export a notebook to a given format.
+   *
+   * @param options - The export options.
+   * @param options.format - The export format (e.g., 'html', 'pdf').
+   * @param options.path - The path to the notebook to export.
+   * @param options.download - Whether to download the file or open it in a new tab. Defaults to false.
+   */
+  async export(options: {
```
bollwyvl, line 92: given `export` has some overloaded meaning in ESM/TS, and this is intended to be overloaded, maybe it's worth some more tweaks:

- consider a slightly more distinguishing name, e.g. `exportAs` 
- make the options an `IExportOptions` so that overloads have something concrete to point at instead of a list of positional args
- potentially allow per-`options.exporterOptions` (`download?` might just be the first example in the simple case)

jtpio: Thanks for the review 👍 

I pushed a couple of changes to implement these ideas.

packages/services/src/nbconvert/index.ts
```diff
@@ -95,22 +123,72 @@ export namespace NbConvertManager {
...
-/**
- * A namespace for builder API interfaces.
- */
-export namespace NbConvert {
   /**
-   * The interface for the build manager.
+   * The interface for the nbconvert export options.
    */
-  export interface IManager extends NbConvertManager {}
+  export interface IExportOptions {
```
bollwyvl, line 150: :rocket:

packages/services/src/nbconvert/index.ts
```diff
@@ -76,16 +81,39 @@ export class NbConvertManager {
...
 
-  protected _requestingFormats: PromiseDelegate<NbConvertManager.IExportFormats> | null;
-  protected _exportFormats: NbConvertManager.IExportFormats | null = null;
+  /**
+   * Export a notebook to a given format.
+   *
+   * @param options - The export options.
+   * @param options.format - The export format (e.g., 'html', 'pdf').
+   * @param options.path - The path to the notebook to export.
+   * @param options.download - Whether to download the file or open it in a new tab. Defaults to false.
```
krassowski, line 90:
```suggestion
   * @param exporterOptions.download - Whether to download the file or open it in a new tab. Defaults to false.
```

packages/services/src/nbconvert/index.ts
```diff
@@ -95,22 +123,72 @@ export namespace NbConvertManager {
...
+     */
+    getExportFormats(force?: boolean): Promise<NbConvert.IExportFormats>;
+
+    /**
+     * Export a notebook to a given format.
+     *
+     * @param options - The export options.
+     * @param options.format - The export format (e.g., 'html', 'pdf').
+     * @param options.path - The path to the notebook to export.
+     * @param options.download - Whether to download the file or open it in a new tab. Defaults to false.
```
krassowski, line 190:
```suggestion
     * @param exporterOptions.download - Whether to download the file or open it in a new tab. Defaults to false.
```

packages/notebook-extension/src/index.ts
```diff
@@ -609,21 +610,24 @@ export const exportPlugin: JupyterFrontEndPlugin<void> = {
...
         if (context.model.dirty && !context.model.readOnly) {
           return context.save().then(() => {
-            window.open(url, '_blank', 'noopener');
+            void services.nbconvert.exportAs?.(exportOptions);
           });
         }
 
         return new Promise<void>(resolve => {
-          window.open(url, '_blank', 'noopener');
+          void services.nbconvert.exportAs?.(exportOptions);
```
krassowski, line 630: What is the rationale for `void` over await here?

jtpio: I think it was just to keep the existing surrounding `return new Promise<void>`.

But maybe we could remove it now and simply return `services.nbconvert.exportAs?.(exportOptions)` ?

krassowski (approved):
Thank you @jtpio!

PR #17487, opened 2025-04-18
Add new CSS rule for expanding linked output to 100% height

SylvainCorlay:
This allows outputs such as ipyleaflet or JupyterGIS maps to make use of the full height of the linked output panel.

### DIFF
--- a/packages/outputarea/style/base.css
+++ b/packages/outputarea/style/base.css
@@ -264,6 +264,10 @@ body.lm-mod-override-cursor .jp-OutputArea-output.jp-mod-isolated::before {
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


mfisher87:
Tested with JupyterGIS, this enables our widget to consistently fill the whole space in a sidecar and "Create new view for cell output" windows.

krassowski:
Do you have some before/after screenshots?

I think this can be merged into 4.5.0 once we branch off. But if you would like to backport it, I wonder if we should test this with more widget libraries/output types?

krassowski (approved):
It would be useful to see a screenshot, but I trust this is generally useful based on feedback from downstream.

Issue #17488, opened 2025-04-18
Signal when cells are pasted in the Notebook


brichet:
### Problem

From a third party extension, it may be interesting to know that a new cell has been pasted, and has previously been copied or cut, to handle the metadata for example.
Currently I don't think that there is a way to retrieve that information.

### Proposed Solution

Add a signal in the Notebook widget when cells are pasted, including whether it has been cut or copied from the current Notebook, or comes from another Notebook.

I was thinking about updating the `lastClipboardInteraction` [setter](https://github.com/jupyterlab/jupyterlab/blob/ea23ac60be1ae91b0feb4a773bf46857b76f1bb0/packages/notebook/src/widget.ts#L1761) as following:

```ts
  set lastClipboardInteraction(newValue: 'copy' | 'cut' | 'paste' | null) {
    const clipboard = Clipboard.getInstance();
    if (clipboard.hasData(JUPYTER_CELL_MIME)) {
      if (newValue === 'copy' || newValue === 'cut') {
        // Save locally the copied/cut cells
        this._localCopy = clipboard.getData(JUPYTER_CELL_MIME) as nbformat.IBaseCell[];
      } else if (newValue === 'paste') {
        const pasted = clipboard.getData(JUPYTER_CELL_MIME) as nbformat.IBaseCell[];

        // Check if the current clipboard match the previously copied/cut cells.
        const isLocal = ArrayExt.shallowEqual(pasted, this._localCopy, JSONExt.deepEqual);
        
        // Emit a signal with the previous interaction, normally 'cut' or 'copy' if the
        // pasted cells come from this Notebook, null otherwise.
        this._cellsPasted.emit({
          action: (isLocal ? this._lastClipboardInteraction : null),
          count: pasted.length
        });
      }
    }
    this._lastClipboardInteraction = newValue;
  }
```

That way third party extension should be able to know which cells has been pasted (using the current cell and the count), and if it follows a copy or cut in the same Notebook.

### Additional context

Related issues:
- https://github.com/jupyterlab/jupyterlab/issues/13228
- https://github.com/jupyter/nbgrader/issues/1083

jtpio:
@brichet looks like this can now be closed as fixed by https://github.com/jupyterlab/jupyterlab/pull/17506?

PR #17506, opened 2025-04-25
Add a signal when cells are pasted in the Notebook

brichet:
This PR adds a signal when cells are pasted to the notebook.

This signal contains an "interaction" object:

```ts
{
  // 'null' if the pasted cell doesn't provide from the same Notebook.
  // Otherwise it should be one of `copy` or `cut`.
  action: 'copy' | 'cut' | 'paste' | null;

  // Number of cells pasted.
  count: number; 
}
```

## References

Fixes:
- https://github.com/jupyterlab/jupyterlab/issues/17488

Related to
- https://github.com/jupyterlab/jupyterlab/issues/13228
- https://github.com/jupyter/nbgrader/issues/1083

## Code changes

- Add a new signal to the `Notebook` widget, which emits when the `lastClipboardInteraction` is set to `'paste'`.
- Add a new private variable `_localCopy` to the notebook widget, which get the last copied (or cut) cell.
- Add unit tests

## User-facing changes

None

None

### DIFF
--- a/packages/notebook/src/widget.ts
+++ b/packages/notebook/src/widget.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { DOMUtils } from '@jupyterlab/apputils';
+import { Clipboard, DOMUtils } from '@jupyterlab/apputils';
 import {
   Cell,
   CodeCell,
@@ -22,7 +22,7 @@ import { TableOfContentsUtils } from '@jupyterlab/toc';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { WindowedList } from '@jupyterlab/ui-components';
 import { ArrayExt, findIndex } from '@lumino/algorithm';
-import { MimeData } from '@lumino/coreutils';
+import { JSONExt, MimeData } from '@lumino/coreutils';
 import { ElementExt } from '@lumino/domutils';
 import { Drag } from '@lumino/dragdrop';
 import { Message } from '@lumino/messaging';
@@ -1621,6 +1621,13 @@ export class Notebook extends StaticNotebook {
     return this._selectionChanged;
   }
 
+  get cellsPasted(): ISignal<
+    this,
+    { action: 'copy' | 'cut' | 'paste' | null; count: number }
+  > {
+    return this._cellsPasted;
+  }
+
   /**
    * The interactivity mode of the notebook.
    */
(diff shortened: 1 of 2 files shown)


krassowski:
FYI there is a PR which rewires the paste actions to use system clipboard, https://github.com/jupyterlab/jupyterlab/pull/17445, this signal may need to be added for that one too or this PR may need modifications one the other one gets merged.

krassowski:
https://github.com/jupyterlab/jupyterlab/pull/17445 is now merged

brichet:
Thanks, I'll rebase as soon as possible

packages/notebook/src/widget.ts
```diff
@@ -1621,6 +1621,13 @@ export class Notebook extends StaticNotebook {
     return this._selectionChanged;
   }
 
+  get cellsPasted(): ISignal<
```
krassowski, line 1624: Can you add add a docstring?

packages/notebook/src/widget.ts
```diff
@@ -1621,6 +1621,13 @@ export class Notebook extends StaticNotebook {
...
+    this,
+    { action: 'copy' | 'cut' | 'paste' | null; count: number }
```
krassowski, line 1626: Should this be an interface with documentaiton for each of the members? It is not immediately obvious what either the `action` nor `count` represents. Should `action` be called `lastInteraction` or `previousInteraction` or `previousAction` instead?

packages/notebook/test/actions.spec.ts
```diff
@@ -1698,6 +1698,117 @@ describe('@jupyterlab/notebook', () => {
         NotebookActions.undo(widget);
         expect(widget.widgets.length).toBe(count - 2);
       });
+
+      it('should emit a signal with cut action', async () => {
+        let signals: {
+          action: 'copy' | 'cut' | 'paste' | null;
+          count: number;
+        }[] = [];
```
krassowski, line 1706: If that was an interface, the test would have fewer repetitive bits like this too.

packages/notebook/src/widget.ts
```diff
@@ -1759,6 +1766,38 @@ export class Notebook extends StaticNotebook {
...
+
+            // Check if the current clipboard match the previously copied/cut cells.
+            const isLocal = ArrayExt.shallowEqual(
+              pasted,
+              this._localCopy,
+              JSONExt.deepEqual
+            );
+
+            // Emit a signal with the previous interaction, normally 'cut' or 'copy' if the
+            // pasted cells come from this Notebook, null otherwise.
```
krassowski, line 1788: Would be great to expose this in the documentation (via docstring).

brichet:
Thanks @krassowski for the comments, I addressed them in the last commit.

krassowski (approved):
Thank you @brichet.

I am wonder if we should have a more generic signal that handles paste/cut/copy, but maybe these could be added separately as individual signals.

brichet:
> I am wonder if we should have a more generic signal that handles paste/cut/copy, but maybe these could be added separately as individual signals.

Yes, I asked myself the same question, too, but did not find a use case for the `copy` and `cut` signal. 
And the data emitted would probably not be the same, as the `previousInteraction` doesn't really make sense in the context of `copy` or `cut`, but the cell contents might.

packages/notebook/src/widget.ts
```diff
@@ -3261,6 +3300,29 @@ export namespace Notebook {
      */
     export interface IOptions extends StaticNotebook.ContentFactory.IOptions {}
   }
+
+  /**
+   * An interface for the paste cell event.
+   */
+  export interface IPasteCells {
```
krassowski, line 3307: Came back to merge it, but here is a nitpick: should it be

```suggestion
  export interface IPastedCells {
```

?

brichet: I updated it.

superlevure:
Any chance we could back-port this great addition to `4.x.x`?

brichet:
> Any chance we could back-port this great addition to `4.x.x`?

I think that we can do that. Maybe in 4.4.x should be enough ?

superlevure:
That'd be perfect yes 🙏

brichet:
@meeseeksdev please backport to 4.4.x

Issue #14619, opened 2023-05-31
Unable to disable autosave in collaborative mode


goyalyashpal:
continued from https://github.com/jupyterlab/jupyterlab/discussions/14516

## Description
_Describe the bug clearly and concisely. Include screenshots if possible_

The changes i am making are getting automatically saved always in jupyter lab 4 (all the --pre, 4.0, and 4.0.1)
_despite_ having the option of "Autosave Documents < Settings" in jupyter lab turned off (shown as blank in the gutter)

https://github.com/jupyterlab/jupyterlab/assets/19423063/0a37c971-a3e8-40bb-8046-9b7599958af3

## Reproduce

1. launch jupyter lab >= 4.0
2. open any file (you have write access to) in jupyter lab
3. make sure the auto save documents is not checked
4. make any changes
5. see the "unsaved indicator" - the filled disc - in the file's tab in place of close tab 'x' button
6. leave it for a few seconds

the unsaved indicating disc disappears (i.e. changes are saved to file) to be replaced by closing x

## Expected behavior

file should not be autosaved

## Context

- Operating System and version: Win 10 22H2
- Browser and version: Vivaldi 6.0.2979.18 (Stable channel) (64-bit) UA: Chrome/112.0.0.0
- JupyterLab version: v4.0.1

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.

```
C:\Users\RESUSER>jupyter troubleshoot
$PATH:
        C:\Program Files\ImageMagick-7.1.0-Q16-HDRI
        C:\Program Files\Common Files\Oracle\Java\javapath
        C:\Program Files (x86)\Common Files\Oracle\Java\javapath
        C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\iCLS\
        C:\Program Files\Intel\Intel(R) Management Engine Components\iCLS\
        C:\Windows\system32
        C:\Windows
        C:\Windows\System32\Wbem
        C:\Windows\System32\WindowsPowerShell\v1.0\
        C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\DAL
        C:\Program Files\Intel\Intel(R) Management Engine Components\DAL
        C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\IPT
        C:\Program Files\Intel\Intel(R) Management Engine Components\IPT
[...]

PR #17508, opened 2025-04-25
Enable save in collaborative mode

Darshan808:
## References #14619 

When `jupyter-collaboration` is installed, files were always autosaved regardless of the autosave setting and manual saving is disabled. This PR restores manual save functionality in RTC by re-enabling the save action and removing the tooltip.

This is part of a broader fix and relates to the corresponding PR in [`jupyter-collaboration`](https://github.com/jupyterlab/jupyter-collaboration/pull/479).

### DIFF
--- a/packages/docregistry/src/context.ts
+++ b/packages/docregistry/src/context.ts
@@ -948,8 +948,7 @@ or load the version on disk (revert)?`,
     }
   }
 
-  private _createSaveOptions(): Partial<Contents.IModel> &
-    Contents.IContentProvisionOptions {
+  private _createSaveOptions(): Partial<Contents.IModel> {
     let content: PartialJSONValue = null;
     if (this._factory.fileFormat === 'json') {
       content = this._model.toJSON();
@@ -963,8 +962,7 @@ or load the version on disk (revert)?`,
     return {
       type: this._factory.contentType,
       format: this._factory.fileFormat,
-      content,
-      contentProviderId: this._contentProviderId
+      content
     };
   }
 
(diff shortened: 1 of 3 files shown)


packages/docregistry/src/context.ts
```diff
@@ -969,7 +963,8 @@ or load the version on disk (revert)?`,
     return {
       type: this._factory.contentType,
       format: this._factory.fileFormat,
-      content
+      content,
+      contentProviderId: this._contentProviderId
```
krassowski, line 967: Why is this needed?

Darshan808:
The `save` method of Drive requires a `contentProviderId` to correctly identify the current content provider. 
https://github.com/jupyterlab/jupyterlab/blob/71a67241b4dd8b1bcf8425403661283a3ad273a5/packages/services/src/contents/index.ts#L1375-L1382

However, in the existing code, the `contentProviderId` was not being passed down, causing it to always be `undefined`. As a result, the default provider was always returned during saving. 

This issue did not surface in normal mode since falling back to the default provider masked the problem. However, in RTC mode, saving requires the correct `contentProviderId` to be explicitly provided, which exposed the bug.

krassowski (approved):
Approving since https://github.com/jupyterlab/jupyter-collaboration/pull/479 was just merged.

When I tested this locally one thing which feels odd (but needs to be addressed on `jupyter-collaboration` side) is that when saving the dirty indicator goes away immediately, but the actual save on disk happens half a second to a second later. I wonder if we should add an await for a confirmation that save succeeded when sending manual save message.

Issue #15267, opened 2023-10-13
Cannot enable line numbers in notebook cells


twopwood:
## Description

Line numbers in any type of cells (code, markdown, raw)  in notebook files do not persist.  I can turn them on with _View_ -> _Show Line Numbers_ but this only works for a given notebook file; if I close it or open a new notebook file, line numbers are turned off and the menu item is unchecked.

When I go to the Settings Editor, and select Notebook, line numbers are disabled despite the fact that I have no files yet in `JUPYTER_CONFIG_DIR/lab/user-settings/@jupyterlab`.
![Screenshot_20231012_151844](https://github.com/jupyterlab/jupyterlab/assets/89099104/15272fbd-e7d9-4b04-921c-50ed07f9c271)
If I click the Line Numbers box, a red "Restore to Defaults" button appears and if I click that, the Line Numbers box becomes unchecked, despite that being the default setting.
![Screenshot_20231012_151955](https://github.com/jupyterlab/jupyterlab/assets/89099104/dc0c5d2c-f5e6-46b6-a9b9-07b55ff58f1a)
The file `JUPYTER_CONFIG_DIR/lab/user-settings/@jupyterlab/notebook-extension/tracker.jupyterlab-settings` appears to be updating as I change settings.  Even if I manually change the file or use the Advanced Settings Editor and set
```
    "codeCellConfig": {
        "lineNumbers": true
    },
```
line numbers still do not appear.  The only way to turn them on (temporarily) is with _View_ -> _Show Line Numbers_.  And if I change any settings in the Settings Editor Notebook settings (i.e. not just the Line Number checkbox) and go back to the open notebook tab, the line numbers are off again.

The Line Wrap setting also cannot be enabled but it is also grayed out in the _View_ menu.  The Notebook settings for Markdown and Raw cells also have the same issue.  All of the other Notebook settings seem to work as expected.  There do not seem to be any issues with line numbers on other file types (e.g. text files).

Overall it seems that something is overriding the Line Number settings.  They are turned off (even though this is not the default setting) in the Notebook settings even on a new install with an empty JUPYTER_CONFIG_DIR.

This issue occurs across multiple distros and browsers.  It even occurs in JupyterLab Desktop.  This issue did not occur in JupyterLab 3.x.

## Reproduce

Clean install of jupyterlab:
`conda create -n jlab-test --override-channels --strict-channel-priority -c conda-forge -c nodefaults jupyterlab`
`conda activate jlab-test `
[...]

PR #17526, opened 2025-05-01
Upgrade CodeMirror packages to latest

### LINKED ISSUE #17464: Not able to fold strings in code cells in Jupyterlab 4
## Description

When trying out Jupyterlab 4.4.0, I notice that there is no option to fold strings that could be folded in Jupyterlab 3.3.6

## Reproduce

1. Turn on Code Folding in `Settings -> Notebook -> Code Cell Configuration -> Code Folding`
2. Refresh page (may not be needed)
3. Create a notebook with the following cell, and try to fold cell by clicking the mini triangle
```python
query = f"""
 SELECT * FROM TABLE
"""
```

## Expected behavior

Code folding option available for strings on Jupyterlab 4.4.0

## Context

Working version:
- Jupyterlab version: 3.6.8

Broken version:
- Jupyterlab version: 4.4.0

https://github.com/user-attachments/assets/7fa62f50-683e-4511-96f5-8233c438f7d0

krassowski:
## References

- Fixes #17464 

## Code changes

None

## User-facing changes

Selected changes, click in the links for full changelog:
- `@codemirror/autocomplete` 6.18.3 → [6.18.6](https://github.com/codemirror/autocomplete/blob/main/CHANGELOG.md#6186-2025-02-12)
  - no changes relevant to JupyterLab
- `@codemirror/commands` 6.7.1 → [6.8.1](https://github.com/codemirror/commands/blob/main/CHANGELOG.md#681-2025-03-31)
  - > Fix an issue where creating a comment for a line that starts an inner language would use the comment style from the outer language.
- `@codemirror/state` 6.5.0 → [6.5.2](https://github.com/codemirror/state/blob/main/CHANGELOG.md#652-2025-02-03)
  - bug fixes for issues without know manifestation in JupyterLab
- `@codemirror/view` 6.35.3 → [6.36.6](https://github.com/codemirror/view/blob/main/CHANGELOG.md#6366-2025-04-24)
   - > Fix an issue where the editor would inappropriately scroll when editing near the bottom of the document with line wrapping enabled, in some cases.
   - > Fix an issue that caused unnecessary transactions on focus change.
   - > Fix an issue where some browsers wouldn't enable context menu paste when clicking on placeholder text.
   - > Fix an issue where cursor height would unnecessarily be based on a placeholder node's dimensions, and thus be off from the text height.
  - > Fix an issue where scrolling down to a range higher than the viewport could in some situations fail to scroll to the proper position.
   - > Don't include the window scrollbars in the space available for displaying tooltips.
  - > Fix an issue where some kinds of relayouts could put the editor in a state where it believed it wasn't in window, preventing relayout, though it in fact was.
   - > Fix a crash in `MatchDecorator` when updating matches at the end of the document.
   - > Make selection rectangles verticaly align precisely, rather than introducing a slight overlap.
   - We may need to use this one for windowing:
      > View updates now have a `viewportMoved` flag that is only true when a viewport change originated from something other than mapping the viewport over a document change.
- `@codemirror/lang-javascript` 6.2.2 → [6.2.3](https://github.com/codemirror/lang-javascript/blob/main/CHANGELOG.md#623-2025-02-12)
  - no changes relevant to JupyterLab
- `@codemirror/lang-python` 6.1.7 → [6.2.0](https://github.com/codemirror/lang-python/blob/main/CHANGELOG.md#620-2025-04-25):
   - docstrings can be folded (as in JupyterLab 3.x)
  - indentation is respected in new lines of subscripts (as in JupyterLab 3.x)
- `@codemirror/lang-markdown 6.3.1 → [6.3.2](https://github.com/codemirror/lang-markdown/blob/main/CHANGELOG.md#632-2025-01-09)
  - > Make Markdown-specific commands return false inside fenced code.
  - > Fix an infinite loop caused by `insertNewlineContinueMarkup`.
- `@codemirror/language` 6.10.6 → [6.11.0](https://github.com/codemirror/language/blob/main/CHANGELOG.md#6110-2025-03-13)
   - > Fix an issue where indentation for a stream language would fail to work when the parse covered only part of the document, far from the start.
- `@codemirror/legacy-modes` 6.4.2 → [6.5.1](https://github.com/codemirror/legacy-modes/blob/main/CHANGELOG.md#651-2025-04-22)
   - > Make Pascal keyword matching case-insensitive.
   - > Fix a bug in tokenizing multiline strings in xQuery.
- `@codemirror/search` 6.5.8 → [6.5.10](https://github.com/codemirror/search/blob/main/CHANGELOG.md#6510-2025-02-26)
  - no changes relevant to JupyterLab

| Before (no code folding for docstrings) | After |
|--|--|
| image | image |

None


brichet (approved):
Thanks @krassowski for the detailed changes of upgrading `CodeMirror`.
Indeed it fixes the string folding as expected.

brichet:
While testing this, I observed some confusion between the same settings in `CodeMirror` and `Notebook -> Code Cell Configuration`.
It is not related to this PR, but I haven't find an issue about this (I can open one if there is not).

In the following record I played with `Code Folding` and `Line Numbers` in both settings sections, and it appears that
- unchecking a checkbox in `Notebook -> Code Cell Configuration` is not reflected in Notebook
- the line numbers are not displayed sometime while it is checked in both section

[record-2025-05-02_11.35.16.webm](https://github.com/user-attachments/assets/868bb846-0e5f-4d66-b827-cc4be044a3ff)

I wonder if there is another place where these settings could be updated.

krassowski:
I think some of it is covered by https://github.com/jupyterlab/jupyterlab/issues/15267, but in general, I find it hard to reason about how the three places which define the CodeMirror editor behaviour (CodeMirror settings, Notebook settings, Settings available in menu in UI) interact with each other.

krassowski:
Thank you for the reviews!

PR #17527, opened 2025-05-01
Fix incorrect ARIA labels in the `MainAreaWidget`

dnlzrgz:
## References

Issue: https://github.com/jupyterlab/jupyterlab/issues/13045
PRs: https://github.com/jupyterlab/jupyterlab/pull/17155, https://github.com/jupyterlab/jupyterlab/pull/15137

## Code changes

- Changed the aria-label for `content.node` to "main area content" as a more generic default, replacing "notebook content."
- Changed the aria-label for `content.toolbar` to "main area toolbar" as a more useful default, replacing "notebook actions."
- Checked the test to address and correct any instances of improper behavior.

## User-facing changes

Improved ARIA labels for the main content area and toolbar to better reflect their purpose.

No backwards-incompatible changes identified.


packages/apputils/src/mainareawidget.ts
```diff
@@ -44,11 +44,21 @@ export class MainAreaWidget<T extends Widget = Widget>
     const trans = (options.translator || nullTranslator).load('jupyterlab');
     const content = (this._content = options.content);
     content.node.setAttribute('role', 'region');
-    content.node.setAttribute('aria-label', trans.__('notebook content'));
+    content.node.setAttribute(
+      'aria-label',
+      this._widget
+        ? this._widget.getAriaLabel('content')
```
krassowski, line 50: I think `getAriaLabel` does not exist?

dnlzrgz: Oh! I'm really sorry. Since the original feedback only mentioned the failing test, I assumed the code changes were correct and only the test needed updating

It's okay then if I update the `aria-label` attributes like this? (as mentioned in the issue)

```
content.node.setAttribute('aria-label', trans.__('main area content'));
//...
toolbar.node.setAttribute('aria-label', trans.__('main area toolbar'));
```

And update the related assertions?

krassowski: I think yes, this would be better than the current state.

Issue #17507, opened 2025-04-25
Code lines in large cells randomly disappear/are not rendered 


invisibleroads:
## Description

1. Cells sometimes partially render or do not render when there are downstream cells that are more than one screen.
2. The commands for inserting a cell above or below sometimes insert the cell in unexpected places.
3. The SHIFT-TAB keyboard shortcut for showing function tooltips sometimes does not work.

## Reproduce

1. Make multiple code cells that are longer than one screen, then work towards the bottom of the notebook. Scroll back up and sometimes the cells that are at the top of the notebook stop rendering, requiring a refresh of the notebook.
2. I'm not sure how to reproduce the insert cell above or below behavior. I just know that sometimes using the keyboard shortcut "ESC a" and "ESC b" inserts the cell in unexpected places. When this happens, then the keyboard shortcuts "ESC k" and "ESC j" skip cells.
3. There are times when the SHIFT-TAB keyboard shortcut stops working. I can't pinpoint exactly when.

## Expected behavior

1. All code cells should be visible when scrolling up or down.
2. "ESC a" should insert a code cell above the current cell and "ESC b" should insert a code cell below the current cell.
3. After a function has been imported into the namespace, typing the name of the function and then pressing SHIFT-TAB should show a tooltip with the function signature.

## Context

- Operating System and version: Fedora 41
- Browser and version: Chrome Version 133.0.6943.53 (Official Build) (64-bit)
- JupyterLab version: 4.4.1

FlynnOConnell:
Unbearable. Makes jupyter lab unusable :( I can't edit anything longer than 10 lines without it going black. Need to refresh the page every 2 minutes

![Image](https://github.com/user-attachments/assets/a89be375-7ac2-43ba-92a8-0eacaca1aaeb)

FlynnOConnell:
https://github.com/user-attachments/assets/586e10f1-d749-4231-951e-05baf9266283

analisis1data:
I have the same issue

Operating System and version: Windows 11
Browser and version: Firefox 137.0.2 (64-bit)
JupyterLab version: 4.4.1

krassowski:
@analisis1data do you have the same issue as described by @FlynnOConnell or one of the issues described by @invisibleroads? Because now it feels like we have four different issues conflated in a single github issue making it really hard to manage

krassowski:
@invisibleroads 

1. All code cells should be visible when scrolling up or down.

[...]

PR #17528, opened 2025-05-01
Fix rendering of long cell editors in windowed notebook

krassowski:
## References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/17507

## Code changes

Does not delay the intersection observer callbacks for CodeMirror editor so that the lines can be rendered as user scrolls through the notebook

## User-facing changes

All lines render again,

None

### DIFF
--- a/packages/notebook/src/windowing.ts
+++ b/packages/notebook/src/windowing.ts
@@ -37,6 +37,16 @@ function isInScrollingNotebook(element: Element | null): boolean {
 }
 
 /**
+ * Check whether the element is part of a CodeMirror editor.
+ */
+function isCodeMirrorElement(element: Element | null): boolean {
+  if (!element) {
+    return false;
+  }
+  return !!element.closest('.cm-editor');
+}
+
+/**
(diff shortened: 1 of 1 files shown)


krassowski:
@andrii-i would you be able to help testing/review? You could test on Binder.

andrii-i (approved):
Thank you @krassowski. This PR fixes the large code cell scrolling problem. I tested both on Binder and locally.

paulmromer:
I had a notebook with a large code cell that would eventually fail to render when I used full mode for windowing and scrolled in the cell.  

With this same cell and full mode for windowing, I have not been able to reproduce this failure on the new Binder version.

nick-shapavalau:
Fixed the issue for me, huge thanks!

brianmapes:
For us grateful half-lost regular users, can someone say how to get the fix? 
just conda update jupyterlab? or mamba?

krassowski:
JupyterLab 4.4.3 is on conda-forge if that is what you are asking.

PR #17538, opened 2025-05-02
Add updatable fields in user manager

brichet:
This PR add a `updatableFields` property to the user manager.
This property is optional, and will be populated by the server API after https://github.com/jupyter-server/jupyter_server/pull/1518

## References

Related to:
- https://github.com/jupyterlab/jupyterlab/issues/14327
- https://github.com/jupyterlab/jupyter-collaboration/issues/350

## Code changes

- add a new type `UpdatableUserField` in `User` namespace.
- add a new private variable in `UserManager`, and a getter. This variable can be populated when refreshing the user information from the server, but the information is not required.
- Update the tests that mock user.

## User-facing changes

None

None

### DIFF
--- a/packages/services/src/testutils.ts
+++ b/packages/services/src/testutils.ts
@@ -764,6 +764,7 @@ export class FakeUserManager extends BaseManager implements User.IManager {
 
   private _identity: User.IIdentity;
   private _permissions: ReadonlyJSONObject;
+  private _updatableFields: User.UpdatableUserField[];
 
   private _userChanged = new Signal<this, User.IUser>(this);
   private _connectionFailure = new Signal<this, Error>(this);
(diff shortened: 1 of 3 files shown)


packages/services/src/testutils.ts
```diff
@@ -774,7 +775,8 @@ export class FakeUserManager extends BaseManager implements User.IManager {
   constructor(
     options: UserManager.IOptions = {},
     identity: User.IIdentity,
-    permissions: ReadonlyJSONObject
+    permissions: ReadonlyJSONObject,
+    updatableFields: User.UpdatableUserField[] = []
```
krassowski, line 779: I wonder, have you explored if this could be encoded as a permission instead of a new field? It feels like user would be granted permission to change its avatar, or nickname so I would be just wary of duplicating concepts; if there are good reasons not to use permissions then its fine to add it as-is.

brichet: I updated https://github.com/jupyter-server/jupyter_server/pull/1518 with this change.
I converted this PR to draft, waiting for https://github.com/jupyter-server/jupyter_server/pull/1518 to be merged, then I'll close it.

brichet:
Closing this PR, it is not relevant anymore after the update of https://github.com/jupyter-server/jupyter_server/pull/1518, which has been merged.

Thanks again @krassowski for the suggestion.

PR #17540, opened 2025-05-05
Fixing Settings Editor Filtered View's Defaulting Behavior in SettingsFormEditor

### LINKED ISSUE #16697: Changing searched settings reset other settings in the same section to default
## Description
When you search for some settings and change them, all other settings within the same section are reset back to default.

The issue is also there when you change multiple settings instead of one. The settings file also shows no other settings so it isn't an UI-only error. Here is a video showcasing the problem:

https://github.com/user-attachments/assets/11ac0914-e0c6-40cc-9841-4b6522de504c

When the other settings are reset, the UI shows blank checkboxes and empty inputs even if the defaults are ticked checkboxes or there is a default value. Restarting JupyterLab Desktop will make the default settings shown correctly (but it doesn't revert back the customized settings that are already reset). The blank UI could be a separate issue.

## Reproduce
1. Open Settings
2. Change a few settings in one section e.g. File Browser
3. Search for settings in the section
4. Change the settings in the search results page (one or multiple settings)
5. All settings other than the changed ones (in step 4) will be reset back to default value (including the ones changed in step 2)

## Expected behavior
Changing searched settings will retain the state of other settings in the same section.

## Context
- Browser and version: JupyterLab Desktop 4.2.1-1 (latest)
- JupyterLab version: 4.2.4 (latest)

yukiiii04:
## References
Fixes #16697 

## Problem Description
As described in the referenced issue, when users interact with a filtered view of the settings editor (e.g., after performing a search), only the matching settings fields are rendered. If a user modifies a setting while a filter is active, the component saves only the filtered subset of the settings (formData), discarding any previously configured settings that were not visible at the time.

This results in a significant data loss issue: unrendered fields, even if previously customized, are removed from the saved configuration, effectively resetting them to defaults. This behavior violates user expectations, especially in workflows where filtering is used to find and tweak individual settings.

## Code changes

The previous implementation of the _onChange method directly replaced this._formData with e._formData which only contains the fields that are currently visible in the UI. This causes all settings not currently visible to be lost, as described in #16697. 

In our implementation, we create a deep clone of the existing form data and only update fields changed in the current view. We iterate through e.formData keys and apply the changed fields while keeping all other settings not visible in the current UI.

## User-facing changes
None in terms of updating UI, but the backend change update makes sure that when the user selects an filtered item (checkbox or input), the previously selected setting preferences do not reset.

N/A

## Notes
The change is fully contained within SettingsFormEditor and does not affect the plugin registry, plugin list, or external APIs. No breaking changes were introduced.

### DIFF
--- a/packages/settingeditor/src/SettingsFormEditor.tsx
+++ b/packages/settingeditor/src/SettingsFormEditor.tsx
@@ -239,15 +239,37 @@ export class SettingsFormEditor extends React.Component<
     );
   }
 
+  //LYNN ADDITION !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   private _onChange = (e: IChangeEvent<ReadonlyPartialJSONObject>): void => {
     this.props.hasError(e.errors.length !== 0);
-    this._formData = e.formData as ReadonlyJSONObject;
+
+    // Create a deep copy of the current form data to work with
+    const updatedFormData = JSONExt.deepCopy(
+      this._formData as PartialJSONObject
+    );
+
+    // Only update fields that are actually present in the filtered view
+    if (e.formData) {
+      // Safely iterate over the keys in e.formData
+      Object.keys(e.formData).forEach(key => {
+        // Use type assertion to tell TypeScript this is safe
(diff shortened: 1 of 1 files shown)


krassowski:
@meeseeksdev please backport to 4.4.x

Issue #16205, opened 2024-04-19
File size display in file browser should use correct units


msanchezst:
### Problem

We have noticed that the file size displayed in the Jupyterlab file browser appears to be using an incorrect unit. The file sizes are reported in MiB but the displayed units are MB,GB, and so on. This is confusing as the actual size is not accurately represented. 

### Proposed Solution

I would recommend updating the file browser to display the file sizes using the correct units (MiB) to match the actual file sizes reported by the OS.  

<img width="1796" alt="Screenshot" src="https://github.com/jupyterlab/jupyterlab/assets/106176704/3e92610b-23fe-45fe-9b85-2efb2dff3fae">

In the screenshot above, the units displayed in the File Explorer should be MiB

### Additional context

Has this issue been evaluated in the past and decided to report the file sizes in this way? If so, I'm curious to understand the reasoning behind that decision.

krassowski:
Quick look: I think that the file manager in `jupyter-server` by default returns size in bytes, based on [this code](https://github.com/jupyter-server/jupyter_server/blob/af1342df16963b952374b9cb110db2121423f24f/jupyter_server/services/contents/filemanager.py#L230-L243) referencing `st_size` which normally is in bytes.

Then on frontend this is formatted here

https://github.com/jupyterlab/jupyterlab/blob/b3f0b44b1136e50b757ad23db4b02d44ccd80740/packages/filebrowser/src/listing.ts#L2695-L2700

Using this function:

https://github.com/jupyterlab/jupyterlab/blob/b3f0b44b1136e50b757ad23db4b02d44ccd80740/packages/filebrowser/src/listing.ts#L3007-L3027

krassowski:
Does it make more sense to show MiB vs MB? What do most file browsers show? Real-estate-wise MB is shorter. Should it be configurable?

msanchezst:
@krassowski thank you for your quick response!  Based on this:

https://github.com/jupyterlab/jupyterlab/blob/b3f0b44b1136e50b757ad23db4b02d44ccd80740/packages/filebrowser/src/listing.ts#L2695-L2700

it looks to me that the value set for `k` in the `formatFileSize`, the formatted file size is converted to Bi Units (KiB,MiB,GiB), but the labels here are incorrect: 

https://github.com/jupyterlab/jupyterlab/blob/b3f0b44b1136e50b757ad23db4b02d44ccd80740/packages/filebrowser/src/listing.ts#L3020

[...]

PR #17556, opened 2025-05-08
Auto-open uploaded files with size and settings control

### LINKED ISSUE #17493: Auto-open uploaded notebook/file after upload
### Problem

When uploading a notebook to work on, it is hard to find it after it was uploaded.

### Proposed Solution

When user uses the upload button either:
- open the uploaded notebook/file
- display a notification with a button allowing to open the notebook/file

Questions:
- a) should we open all file types or only notebooks and text files?
- b) should we display the notification about upload complete for both upload button and drag-and-drop?
- c) when user uploads multiple files should we open them all?
- d) if user uploads a large file which could freeze the UI when opened, should we still open it? Should there be a size limit for auto-open?
- e) should it be opt-in or opt-out?

### Additional context

- Google Colab opens the uploaded notebook straightaway. Colab only allows to upload one notebook at a time.
- Google Drive shows a little notification in the right corner allowing to open the uploaded file or to display its location:
   Image

Darshan808:
### Description:

This PR adds a new plugin that improves the UX of file uploads in the File Browser:

- Automatically opens uploaded files if it's a single file and within a size threshold.
- Shows a notification with “Open All” for multiple uploads or large files.
 image
- Introduces two new user settings under filebrowser:
  - `autoOpenUploads` (default: `true`)
  - `maxAutoOpenSizeMB` (default: `50`)

Since `FileBrowserModel` does not distinguish between single and multiple file uploads, I added a new `filesUploaded` signal in the `Uploader` class to emit upload results, which this plugin listens to.

### DIFF
--- a/galata/test/galata/filebrowser.spec.ts
+++ b/galata/test/galata/filebrowser.spec.ts
@@ -2,6 +2,8 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { expect, test } from '@jupyterlab/galata';
+import * as fs from 'fs';
+import * as path from 'path';
 
 const DEFAULT_NAME = 'Untitled.ipynb';
 
@@ -44,5 +46,74 @@ test.describe('filebrowser helper', () => {
     await expect(
       page.getByRole('main').getByRole('tab', { name: DEFAULT_NAME })
     ).toHaveCount(2);
+  });
+});
+
+test.describe('upload auto-open behavior', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.activity.closeAll();
+  });
+
+  test('small notebook uploads and opens automatically', async ({
+    page,
+    tmpPath
+  }) => {
+    const nbName = 'small.ipynb';
+    const nbPath = path.join(tmpPath, nbName);
+    fs.mkdirSync(tmpPath, { recursive: true });
+    fs.writeFileSync(
+      nbPath,
+      JSON.stringify({
+        cells: [],
+        metadata: {},
+        nbformat: 4,
+        nbformat_minor: 5
+      })
+    );
+
+    await page.filebrowser.openDirectory(tmpPath);
+
+    // Intercept the file chooser that the Upload button triggers
+    const [fileChooser] = await Promise.all([
+      page.waitForEvent('filechooser'),
+      page.getByTitle('Upload').click()
+    ]);
+    await fileChooser.setFiles(nbPath);
+    await page
+      .locator('.jp-Dialog')
+      .getByRole('button', { name: 'Select Kernel', exact: true })
+      .click();
+    expect(await page.activity.isTabActive(nbName)).toBe(true);
+  });
+
+  test('large notebook (>50 MB) uploads but does not auto-open (notification shown)', async ({
+    page,
+    tmpPath
+  }) => {
+    const bigName = 'big.ipynb';
+    const bigPath = path.join(tmpPath, bigName);
+    const minimal = JSON.stringify({
+      cells: [],
+      metadata: {},
+      nbformat: 4,
+      nbformat_minor: 5
+    });
+    const repeat = Math.ceil((52 * 1024 * 1024) / minimal.length);
+    fs.mkdirSync(tmpPath, { recursive: true });
+    fs.writeFileSync(bigPath, minimal.repeat(repeat));
+
+    await page.filebrowser.openDirectory(tmpPath);
+
+    const [fileChooser] = await Promise.all([
+      page.waitForEvent('filechooser'),
+      page.getByTitle('Upload').click()
+    ]);
+    await fileChooser.setFiles(bigPath);
+
+    expect(await page.activity.isTabActive(bigName)).toBe(false);
+    await page.getByRole('button', { name: 'Upload', exact: true }).click();
+    //Wait for upload to finish
+    await page.waitForTimeout(1500);
+    await expect(page.getByText(/Uploaded big\.ipynb.*/)).toBeVisible();
   });
 });
--- a/packages/filebrowser-extension/schema/browser.json
+++ b/packages/filebrowser-extension/schema/browser.json
@@ -242,6 +242,18 @@
       "title": "Navigate files and directories with single click",
       "description": "Whether to allow single click selection on file browser items",
       "default": false
+    },
+    "autoOpenUploads": {
+      "type": "boolean",
+      "title": "Auto-open uploads",
+      "description": "Automatically open uploaded notebooks",
+      "default": true
+    },
+    "maxAutoOpenSizeMB": {
+      "type": "number",
+      "title": "Max auto-open size",
+      "description": "Maximum file size (MB) to auto-open",
+      "default": 50
     }
   },
   "additionalProperties": false,
--- a/packages/filebrowser-extension/src/index.ts
+++ b/packages/filebrowser-extension/src/index.ts
@@ -20,6 +20,7 @@ import {
   ICommandPalette,
   InputDialog,
   IToolbarWidgetRegistry,
+  Notification,
   setToolbar,
   showErrorMessage,
   WidgetTracker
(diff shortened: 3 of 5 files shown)


packages/filebrowser-extension/schema/browser.json
```diff
@@ -242,6 +242,18 @@
...
+    "autoOpenUploads": {
+      "type": "boolean",
+      "title": "Auto-open uploads",
+      "description": "Automatically open uploaded notebooks",
+      "default": true
+    },
+    "maxAutoOpenSizeMB": {
+      "type": "number",
+      "title": "Max auto-open size",
+      "description": "Maximum file size (MB) to auto-open",
```
krassowski, line 255:
```suggestion
      "description": "Maximum file size (MB) to auto-open files on upload",
```

packages/filebrowser-extension/schema/browser.json
```diff
@@ -242,6 +242,18 @@
       "title": "Navigate files and directories with single click",
       "description": "Whether to allow single click selection on file browser items",
       "default": false
+    },
+    "autoOpenUploads": {
+      "type": "boolean",
+      "title": "Auto-open uploads",
+      "description": "Automatically open uploaded notebooks",
```
krassowski, line 249: Is it notebooks and files?

```suggestion
      "description": "Automatically open uploaded notebooks and files",
```

packages/filebrowser/src/upload.ts
```diff
@@ -34,6 +36,13 @@ export class Uploader extends ToolbarButton {
     this.addClass('jp-id-upload');
   }
 
+  /**
+   * A signal emitted with file info when a batch of upload completes .
```
krassowski, line 40:
```suggestion
   * A signal emitted with file info when a batch of upload completes.
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,114 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+        } else {
+          // multi-file case
+          Notification.emit(
+            `Upload Complete (${models.length} files)`,
+            'info',
+            {
+              autoClose: 5000,
+              actions: [
+                {
+                  label: 'Open All',
```
krassowski, line 991: Needs translation wrapper.

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,114 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+                file.size
+                  ? ` (${(file.size / (1024 * 1024)).toFixed(1)} MB)`
+                  : ''
+              }`,
+              'info',
+              {
+                autoClose: 5000,
+                actions: [
+                  {
+                    label: 'Open File',
```
krassowski, line 966: Needs translation wrapper

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,114 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+            // open immediately
+            void app.commands
+              .execute('docmanager:open', { path: file.path })
+              .catch(err => {
+                void showErrorMessage(`Opening ${file.name} failed`, err);
+              });
+          } else {
+            // notify and offer an “Open” button
+            Notification.emit(
+              `Uploaded ${file.name}${
```
krassowski, line 956: Need translation

krassowski (commented):
I wonder if we need to be more selective with auto-opening files; for example, should we avoid auto-opening `.zip`/`.gz` and other archives and binary files? These do not work well with plain JupyterLab when opened with the default editor and just show infinite spinner:

image

Also, I wonder if there should be different size limits for files and notebooks. It is reasonable to open a 200 MB notebook, but a 200 MB txt file would likely choke the editor.

Darshan808:
>  I wonder if there should be different size limits for files and notebooks. It is reasonable to open a 200 MB notebook, but a 200 MB txt file would likely choke the editor.

We could definitely consider adding different size limits for notebooks and files, but I did a quick check and interestingly, I didn't observe much difference in the loading times between a 200 MB notebook and a 200 MB text file. In fact, the text file loaded slightly faster in my tests:

- 200 MB notebook: ~4.6 seconds
- 200 MB text file: ~3.8 seconds

Curious to hear your thoughts on this.

Darshan808:
These are the files on which auto-open won't be applied:
```python
const isBinaryFile = ['.zip', '.gz', '.tar', '.rar', '.7z', '.exe', '.bin', '.iso'];
```
Any suggestions on which others to add or are these fine?

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,140 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+            }
+            return acc;
+          },
+          []
+        );
+        // Check if the uploaded file has an allowed extension
+        const fileExt = models[0].name
+          .toLowerCase()
+          .slice(models[0].name.lastIndexOf('.'));
+        const isAllowedFileType = allowedExtensions.includes(fileExt);
```
krassowski, line 958: Instead of `lastIndexOf` we may need to loop over them and check suffixes because the extensions may be composite, e.g. `.table.json` as per https://jupyterlab.readthedocs.io/en/4.1.x/api/interfaces/docregistry.DocumentRegistry.IFileType.html#extensions

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,140 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+        | undefined;
+      if (!uploader) {
+        console.warn('Uploader widget not found');
+        return;
+      }
+
+      uploader.filesUploaded.connect((_sender, models) => {
+        // single-file case
+        // Get all allowed extensions from default file types
+        const defaultFileTypes = DocumentRegistry.getDefaultFileTypes();
```
krassowski, line 944: Not sure if we should use only the default file types or request a token and get all file types, including these registered by extensions. What do you think?

Darshan808: I couldn't find any specific token to get those, but I used `app.docRegistry.fileTypes()` instead. From what I can tell, it does include the file types registered by other extensions as well. Let me know if there's a better approach!

galata/test/galata/filebrowser.spec.ts
```diff
@@ -46,3 +48,72 @@ test.describe('filebrowser helper', () => {
...
+      page.waitForEvent('filechooser'),
+      page.getByTitle('Upload').click()
+    ]);
+    await fileChooser.setFiles(bigPath);
+
+    expect(await page.activity.isTabActive(bigName)).toBe(false);
+    await page.getByRole('button', { name: 'Upload', exact: true }).click();
+    //Wait for upload to finish
+    await page.waitForTimeout(1500);
+    await expect(page.getByText(/Uploaded big\.ipynb.*/)).toBeVisible();
```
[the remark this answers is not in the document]
krassowski: It looks like it is still failing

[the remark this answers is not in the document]
Darshan808: It is passing locally, but not sure why its failing on CI. I'll try increasing timeout and see whether it'll help or not.

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,140 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+                void showErrorMessage(`Opening ${file.name} failed`, err);
+              });
+          } else {
+            // notify and offer an “Open” button
+            Notification.emit(
+              trans.__(
+                'Uploaded %1%2',
+                file.name,
+                file.size
+                  ? trans.__(' (%1 MB)', (file.size / (1024 * 1024)).toFixed(1))
```
krassowski, line 984: For small files, should we show the size in KB instead? Currently I sometimes see 0.0MB which is somewhat confusing. Or maybe just drop the size if it is negligible?

image

There is a private utility in `@filebrowser/listing` which we could export and reuse (though its implementation is disputed https://github.com/jupyterlab/jupyterlab/issues/16205).

https://github.com/jupyterlab/jupyterlab/blob/b3f0b44b1136e50b757ad23db4b02d44ccd80740/packages/filebrowser/src/listing.ts#L3007-L3027

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,140 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+        ) {
+          const file = models[0];
+          if (
+            autoOpen &&
+            file.size &&
+            file.size <= maxSize &&
+            isAllowedFileType
+          ) {
+            // open immediately
+            void app.commands
```
krassowski, line 972: The `void` here may be unnecessary since there is `.catch`, right? Similar in two more places below.

```suggestion
            app.commands
```

packages/filebrowser-extension/schema/browser.json
```diff
@@ -242,6 +242,18 @@
       "title": "Navigate files and directories with single click",
       "description": "Whether to allow single click selection on file browser items",
       "default": false
+    },
+    "autoOpenUploads": {
+      "type": "boolean",
+      "title": "Auto-open uploads",
+      "description": "Automatically open uploaded notebooks and files",
```
krassowski, line 249: Let's make it explicit that non-supported file types will not be opened:

```suggestion
      "description": "Automatically open uploaded notebooks and supported files",
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -903,6 +904,140 @@ const openUrlPlugin: JupyterFrontEndPlugin<void> = {
...
+                    }
+                  }
+                ]
+              }
+            );
+          }
+        } else {
+          // multi-file case
+          Notification.emit(
+            trans.__('Upload Complete (%1 files)', models.length),
```
krassowski, line 1011: I am not 100% sure, but it feels more appropriate to use sentence case here:

```suggestion
            trans.__('Upload complete (%1 files)', models.length),
```

krassowski (commented):
Tested it again, it is looking good to me @Darshan808! Just a few final suggestions before we merge.

packages/filebrowser-extension/src/index.ts
Darshan808: This implementation works well for us. Would it make sense to move it out of the `Private` namespace and export it?

PR #17559, opened 2025-05-08
Use hatch to get the version

### LINKED ISSUE #17558: Translatable strings check needs to be updated (no `setuptools` anymore)
This job: https://github.com/jupyterlab/jupyterlab/actions/runs/14906199682/job/41869168954?pr=17557 failed for a wrong reason., It should have failed due to translatable string modification, but it failed because of `setuptools` not being installed:

```
++ python setup.py --version
  File "/home/runner/work/jupyterlab/jupyterlab/setup.py", line 5, in <module>
    __import__("setuptools").setup()
    ^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'setuptools'
+ [[ '' =~ ^[0-9]+\.[0-9]+\.0(\.dev|a|b|rc)[0-9]+$ ]]
+ exit 1
```

We need to update:

https://github.com/jupyterlab/jupyterlab/blob/8401ef9900476b4d9445e19a41fd0a3ebadda125/.github/workflows/check-i18n.yml#L42-L47

krassowski:
## References

Fixes #17558 

## Code changes

Replaces `python setup.py --version` with `hatch version`

To test this change, run:

```
if [[ ( $(hatch version) =~ ^[0-9]+\.[0-9]+\.0(\.dev|a|b|rc)[0-9]+$ )]]; then echo "pre-release"; fi
```

on both `main` and `4.4.x ` branch and see that it correctly prints `pre-release` on `main` and nothing on `4.4.x`

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/check-i18n.yml
+++ b/.github/workflows/check-i18n.yml
@@ -42,7 +42,7 @@ jobs:
             # Set job status as failed if
             # - branch is not "main"
             # - and current version is not a pre release of a zeroed patch version
-            if [[ "${GITHUB_BASE_REF}" != "main" && ! ( $(python setup.py --version) =~ ^[0-9]+\.[0-9]+\.0(\.dev|a|b|rc)[0-9]+$ ) ]]; then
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

PR #17563, opened 2025-05-13
Add a `selectionChanged` signal to the file browser

### LINKED ISSUE #14598: Emit a signal when the file browser selected items changes
### Problem

Currently there is no easy way to know when the list of selected items in the file browser changes.

### Proposed Solution

Add a new `selectionChanged` signal on the `FileBrowser`. The args could contain the list of selected items or `void` since they can already be retrieved with `browser.selectedItems()`.

### Additional context

This would directly help Notebook 7 in the case of adding buttons to the file browser toolbar when multiple items are selected: https://github.com/jupyter/notebook/pull/6888

jtpio:
## References

Fixes #14598 

## Code changes

- [x] Add a `selectionChanged` signal to the file browser and emit it when the file browser selection changes
- [x] Add tests

## User-facing changes

None for regular users.

This will mostly be useful for Notebook 7 which displays some action buttons based on the current selection, and had to workaround that for now in https://github.com/jupyter/notebook/pull/6888

None.

Of note, the handling of the selection could have been moved to a proper class (owning the signal and the selected path), had `selection` not been `protected` in `DirListing`. But since it's `protected` and may be used in derived classes, better not change it and keep it as it is.

### DIFF
--- a/docs/source/extension/extension_migration.rst
+++ b/docs/source/extension/extension_migration.rst
@@ -9,6 +9,15 @@ Extension Migration Guide
 JupyterLab 4.4 to 4.5
 ---------------------
 
+File Browser updates
+^^^^^^^^^^^^^^^^^^^^
+
+The file browser now emits a ``selectionChanged`` signal when the selected files and folders change.
+
+If you maintain an extension which implements a subclass of ``DirListing`` that changes selection, it is recommended to
+emit the ``selectionChanged`` signal when the selection changes. This will ensure that any listeners to the
+``selectionChanged`` signal are notified of the change.
+
 API updates
 ^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


krassowski:
> had `selection` not been `protected` in `DirListing`. But since it's `protected` and may be used in derived classes

Should this PR also add a note to extension migration guide recommending that if someone implements a subclass of `DirListing` which changes `selection` they now should emit `selectionChanged`? This could be relevant to extensions like `jupyterlab-unfold`.

jtpio:
Thanks @krassowski for the review and suggestion!

That makes sense, I'll push a change for this now.

krassowski:
Does toggling the selection on Ctrl + space need to emit `selectionChanged` too?

https://github.com/jupyterlab/jupyterlab/blob/cafbf1758d9b91f3f8ea638a0b0471a96b8ea605/packages/filebrowser/src/listing.ts#L1743-L1749

jtpio:
> Does toggling the selection on Ctrl + space need to emit `selectionChanged` too?

I think so, looks like this one was missed, thanks for catching it!

krassowski (approved):
Thank you @jtpio!

Issue #16690, opened 2024-08-18
Ctrl + A in file browser listing should select all files


krassowski:
### Problem

Moving or downloading multiple files is cumbersome as I need to select all by clicking multiple files, using shift + click or using the checkbox. It should be easier. 

### Proposed Solution

All UI Linux file browsers I know support Ctrl + A shortcut for selecting all. I think this is also the case for Windows. Not sure what shortcut is used on Mac.

### Additional context

It should be tested that the shortcut is not active when the file name is edited as in that case Ctrl + A should select everything in the edit field.

We would want to add a new command, say `filebrowser:select-all` and then a shortcut here:

https://github.com/jupyterlab/jupyterlab/blob/35bab6e06d8e4f347219f91a4f0315a4169dfe09/packages/filebrowser-extension/schema/browser.json#L143-L184

JasonWeill:
<kbd>CMD</kbd>+<kbd>A</kbd> does the same thing in Finder, the macOS file browser.

JasonWeill:
From 2024-09-04 frontends meeting: if all files are selected, and the user presses <kbd>ENTER</kbd> or <kbd>RETURN</kbd>, should we display a warning dialog asking the user whether they really want to open a lot of files at once? This is particularly important in notebook or in JupyterLab in single-document mode, where every file opens in its own tab/window, so closing a large number of windows or tabs may be cumbersome. This might merit a separate issue.

PR #17571, opened 2025-05-15
Add select all command for file browser listing

RRosio:
## References

This aims to address https://github.com/jupyterlab/jupyterlab/issues/16690

## Code changes

- Register select all command as a shortcut and context menu item
   - Adding the context menu option was not mentioned in the original issue, but I followed the pattern I saw for other commands. However, I can remove it if this is not the desired behavior. I did wonder if it might be redundant given that the DirListing checkbox already exists.
- Adds a `selectAll` function to the `DirListing` and `FileBrowser` classes

## User-facing changes

The user will now see the Select All option, and its corresponding shortcut, in the context menu after right-clicking on a file. 

https://github.com/user-attachments/assets/da8c6ff4-47dd-438e-99c3-2474dc40745f

None to my knowledge

### DIFF
--- a/packages/filebrowser-extension/src/index.ts
+++ b/packages/filebrowser-extension/src/index.ts
@@ -87,6 +87,8 @@ namespace CommandIDs {
 
   export const duplicate = 'filebrowser:duplicate';
 
+  export const selectAll = 'filebrowser:select-all';
+
   // For main browser only.
   export const hideBrowser = 'filebrowser:hide-main';
 
@@ -967,6 +969,18 @@ function addCommands(
     },
     icon: copyIcon.bindprops({ stylesheet: 'menuItem' }),
     label: trans.__('Duplicate')
+  });
+
+  commands.addCommand(CommandIDs.selectAll, {
+    execute: () => {
+      const widget = tracker.currentWidget;
+
+      if (widget) {
+        return widget.selectAll();
+      }
+    },
+    icon: copyIcon.bindprops({ stylesheet: 'menuItem' }),
+    label: trans.__('Select All')
   });
 
   commands.addCommand(CommandIDs.goToPath, {
--- a/packages/filebrowser/src/browser.ts
+++ b/packages/filebrowser/src/browser.ts
@@ -423,6 +423,13 @@ export class FileBrowser extends SidePanel {
   }
 
   /**
+   * Select all listing items.
+   */
+  selectAll(): void {
+    this.listing.selectAll();
+  }
+
+  /**
    * Download the currently selected item(s).
    */
   download(): Promise<void> {
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -515,6 +515,21 @@ export class DirListing extends Widget {
   }
 
   /**
+   * Select all directory listing items
+   */
+  selectAll(): void {
+    const items = this._model.items();
+    const newSelection: { [key: string]: boolean } = {};
+
+    for (const item of items) {
+      newSelection[item.path] = true;
+    }
+
+    this.selection = newSelection;
+    this.update();
+  }
+
+  /**
    * Download the currently selected item(s).
    */
   async download(): Promise<void> {
--- a/packages/filebrowser-extension/schema/browser.json
+++ b/packages/filebrowser-extension/schema/browser.json
@@ -92,6 +92,11 @@
         "rank": 12
       },
       {
+        "command": "filebrowser:select-all",
+        "selector": ".jp-DirListing-item",
+        "rank": 12.5
+      },
+      {
         "command": "filebrowser:copy-path",
         "selector": ".jp-DirListing-item[data-isdir]",
         "rank": 14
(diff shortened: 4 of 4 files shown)


RRosio:
For jupyterlab visual regression tests, these context-menu tests could be addressed with snapshot updates
```
[jupyterlab] › test/jupyterlab/contextmenu.test.ts:50:7 › Application Context Menu › Open file browser context menu on folder 
[jupyterlab] › test/jupyterlab/contextmenu.test.ts:66:7 › Application Context Menu › Open file browser context menu on file 
[jupyterlab] › test/jupyterlab/contextmenu.test.ts:81:7 › Application Context Menu › Open file browser context menu on notebook with kernel 
```

- The windowed-notebook I'm not too sure about... It doesn't seem to have been failing/flaky as of late, so I can investigate the cause a bit more.
   -`[jupyterlab] › test/jupyterlab/windowed-notebook.test.ts:75:5 › should not update height when hiding `

For the documentation visual regression tests, all three may be addressed with snapshot updates
```
[documentation] › test/documentation/customization.test.ts:89:7 › Default › should use default context menu 
[documentation] › test/documentation/general.test.ts:299:7 › General › Shareable link ──────────
[documentation] › test/documentation/general.test.ts:640:7 › General › Open With ───────────────
```

If keeping the context menu item for `Select All` is preferred, I can ping the bot to update the snapshots.

packages/filebrowser/src/listing.ts
```diff
@@ -514,6 +514,21 @@ export class DirListing extends Widget {
...
+   */
+  selectAll(): void {
+    const items = this._model.items();
+    const newSelection: { [key: string]: boolean } = {};
+
+    for (const item of items) {
+      newSelection[item.path] = true;
+    }
+
+    this.selection = newSelection;
```
krassowski, line 528: Following https://github.com/jupyterlab/jupyterlab/pull/17563 this should also emit `selectionChanged` signal:

```suggestion
    this.selection = newSelection;
    this._selectionChanged.emit()
```

packages/filebrowser/src/listing.ts
```diff
@@ -514,6 +514,21 @@ export class DirListing extends Widget {
       });
   }
 
+  /**
+   * Select all directory listing items
```
krassowski, line 518: Is this?

```suggestion
   * Select all listing items in the current directory.
```

I wonder, how it would behave with `jupyterlab-unfold` installed.

packages/filebrowser/src/listing.ts
```diff
@@ -514,6 +514,22 @@ export class DirListing extends Widget {
       });
   }
 
+  /**
+   * Select all listing items in the current directory.
+   */
+  selectAll(): void {
```
krassowski: Can we make it an async method? It does not need to be one as of today, but I have in the back of my mind that we were talking about lazy fetching of contents for large directories and in these cases we may need to call API async, it will be easier to make this change if the methods are already `async`.

```suggestion
  async selectAll(): Promise<void> {
```

Also, most of the surrounding methods (`download`, `goUp`, `duplicate`, `delete`, etc) are already `async`.

packages/filebrowser/src/browser.ts
```diff
@@ -422,6 +422,13 @@ export class FileBrowser extends SidePanel {
     return this.listing.duplicate();
   }
 
+  /**
+   * Select all listing items.
+   */
+  selectAll(): void {
+    this.listing.selectAll();
```
krassowski:
```suggestion
  selectAll(): Promise<void> {
    return this.listing.selectAll();
```

packages/filebrowser/src/listing.ts
```diff
@@ -514,6 +514,22 @@ export class DirListing extends Widget {
...
+    const items = this._model.items();
+    const newSelection: { [key: string]: boolean } = {};
+
+    for (const item of items) {
+      newSelection[item.path] = true;
+    }
+
+    this.selection = newSelection;
+    this._selectionChanged.emit()
```
krassowski:
```
 Property '_selectionChanged' does not exist on type 'DirListing'.
```

I think this needs a merge with `main`.

packages/filebrowser-extension/src/index.ts
```diff
@@ -1103,6 +1105,18 @@ function addCommands(
...
 
+  commands.addCommand(CommandIDs.selectAll, {
+    execute: () => {
+      const widget = tracker.currentWidget;
+
+      if (widget) {
+        return widget.selectAll();
+      }
+    },
+    icon: copyIcon.bindprops({ stylesheet: 'menuItem' }),
```
krassowski: Select all should not have a copy icon - it looks like this was a copy-paste error.

```suggestion
```

packages/filebrowser-extension/src/index.ts
```diff
@@ -1103,6 +1105,17 @@ function addCommands(
...
 
+  commands.addCommand(CommandIDs.selectAll, {
+    execute: () => {
+      const widget = tracker.currentWidget;
+
+      if (widget) {
+        return widget.selectAll();
+      }
+    },
+    label: trans.__('Select All')
```
krassowski: Picking up `describedBy` noise from #17649.

```suggestion
    label: trans.__('Select All'),
    describedBy: {
      args: {
        type: 'object',
        properties: {}
      }
    }
```

PR #17577, opened 2025-05-19
Ensure strings with yellow background in tracebacks are visible

### LINKED ISSUE #17576: Unreadable strings in tracebacks (yellow text on yellow background)
## Description
Strings in tracebacks are formatted with yellow text, but there are cases where they will also have a yellow background, which makes them impossible to read.
This affects the light, dark and dark (high contrast) themes.

Image

There exists a CSS rule in the dark themes to set the text black (introduced in #13249, and later copied with the addition of the high contrast theme), but it currently does not have precedence for the string (with yellow foreground color) due to the use of `:where`. It does apply correctly to the function names and parenthesis, which would be white if the rule is manually disabled:
Image)

## Reproduce
Any python code rendering strings in a traceback where they would have a yellow background, such as:
```python
def test(string):
    raise

test('hello')
```

## Expected behavior
If the text was black on yellow background, it would be very easy to read.

I see a few ways of addressing this problem, other suggestions are welcome:
1. Add `!important` to the CSS rule fixing the illegible text in the dark themes and copy this rule in the light theme CSS.
2. Remove the `:where` CSS function to increase its precedence and copy this rule to the light theme.
3. **Recommended**:  Remove the `:where` CSS function to increase its precedence and move this rule in the base CSS file rather than duplicating it in the three separate theme files
4. _More complex_: Prevent the creation of HTML elements with both `ansi-yellow-fg`  and `ansi-yellow-bg` classes.

## Context

- Operating System and version: Ubuntu 24.04
- Browser and version: Firefox 138.0
- JupyterLab version: 4.5.0a0 (also tested and reproduced with 3.6.8) 

---

I would be happy to open a PR with one of the simpler fixes.

# 🐍 🤹‍♂

EtiennePelletier:
## References
Fixes #17576

## Code changes

As the bug is not not specific to any of the three built-in themes, I opted to move the existing rule in the base CSS file, and remove the `:where` CSS function that reduced its specificity. I believe this is a better solution than using `!important` or duplicating the same rule across multiple theme files. In all cases, black text on a yellow background is very readable. Feedback is welcome!

## User-facing changes

Users will now be able to read strings that are part of exceptions where content has a yellow background.

image

# :snake: :man_juggling:

### DIFF
--- a/packages/rendermime/style/base.css
+++ b/packages/rendermime/style/base.css
@@ -213,6 +213,11 @@
   padding-top: var(--jp-code-padding);
 }
 
+/* fix illegible yellow text with yellow background in exception stacktrace */
+.jp-RenderedText pre .ansi-yellow-bg.ansi-yellow-fg {
+  color: black;
+}
+
 /*-----------------------------------------------------------------------------
 | RenderedLatex
 |----------------------------------------------------------------------------*/


krassowski (approved):
Tested to confirm that this does not make it worse on IPython 8.x and does improve it on IPython 9.x

| IPython 9 | IPython 8 |
|--|--|
| image | image |

krassowski:
Though it's a pity that numbers are no longer highlighted. This is why there was `:where` in the first place. We can either ship this patch with a small degradation of highlighting but with the strings being legible in 9.x, or wait for a patch in IPython.

krassowski:
It looks like it needs a lint pass:

```
[warn] packages/rendermime/style/base.css
```

EtiennePelletier:
Thank you for the advice. Considering the issue will likely be fixed upstream in IPython, I also think it would make sense to patch only the yellow fg+bg edge case within Jupyterlab. I will work on this and amend the pull request.

EtiennePelletier:
Here's the new before/after with IPython 9 with the modified PR, only the strings are affected:
image

krassowski (approved):
Thank you @EtiennePelletier!

Issue #16746, opened 2024-09-05
Horizontal line - created invalid empty header in TOC


stefansimik:
## Description

Issue is very easy and reproducible:

Simply, when one write "---" (3 dashes) in Markdown, it creates horizontal line. This works fine.
Problem is, that it also creates invalid empty header in TOC.

See on the picture:

![CleanShot 2024-09-05 at 11 11 42@2x](https://github.com/user-attachments/assets/747f7a2a-bd95-430f-880e-29c35584ee35)

## Reproduce

Very easily, simply write 3 dashes anywhere (as simple paragraph = enter before + after)

## Expected behavior

When there are only 3 dashes - it should not create empty header in TOC.

Nobody wants to really create header with empty content (that would not make any sense), 
so this special scenario should be always treaded like intent to write horizontal line (not intent to write header).

## Context

- Operating System and version: MacOS Sonome v.14.6.1   (M1 chip)
- Browser and version: Chrome 128.0.6613.113 (Official Build) (arm64)
- JupyterLab version: 4.2.5

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/opt/anaconda3/envs/mlcourse/bin
	/opt/anaconda3/condabin
	/usr/local/bin
	/System/Cryptexes/App/usr/bin
	/usr/bin
	/bin
	/usr/sbin
	/sbin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin
	/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
	/Users/stesim/Applications/quarto/bin
	/Users/stesim/Library/Application Support/JetBrains/Toolbox/scripts

sys.path:
	/opt/anaconda3/envs/mlcourse/bin
	/opt/anaconda3/envs/mlcourse/lib/python312.zip
	/opt/anaconda3/envs/mlcourse/lib/python3.12
	/opt/anaconda3/envs/mlcourse/lib/python3.12/lib-dynload
	/opt/anaconda3/envs/mlcourse/lib/python3.12/site-packages

sys.executable:
	/opt/anaconda3/envs/mlcourse/bin/python

sys.version:
	3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:32:50) [Clang 16.0.6 ]

platform.platform():
[...]

PR #17583, opened 2025-05-20
Enhancing the TOC logic

itsmevichu:
> I think we could have a reasonable performance without increasing the API surface with the following approach:
> - populate ToC asynchronously cell by cell
> - for markdown cells render them to HTML in full; if they are rendered in notebook maybe just reuse the rendered content, if not do that async or in a worker
> - parse headings out of generated HTML
> - do not parse the markdown cell again until its content or type changed
_Originally posted by @krassowski in https://github.com/jupyterlab/jupyterlab/issues/16456#issuecomment-2165630236_

This PR enhances the overall TOC logic as per the above comment.

Fixes the below issues:
- https://github.com/jupyterlab/jupyterlab/issues/16371
- https://github.com/jupyterlab/jupyterlab/issues/16746
- https://github.com/jupyterlab/jupyterlab/issues/15902


itsmevichu:
@krassowski @fcollonval Please take a look at the code and share any comments. I'll update or add unit tests after your review.

packages/fileeditor/src/toc/markdown.ts
```diff
@@ -34,30 +46,40 @@ export class MarkdownTableOfContentsModel extends TableOfContentsModel<
...
 export class MarkdownTableOfContentsFactory extends EditorTableOfContentsFactory {
+  parser: IMarkdownParser | null;
+
+  constructor(
+    tracker: WidgetTracker<IDocumentWidget<FileEditor>>,
+    parser: IMarkdownParser | null
+  ) {
+    super(tracker);
+    this.parser = parser;
+  }
```
krassowski, line 81: Can we make it protected?

```suggestion
  constructor(
    tracker: WidgetTracker<IDocumentWidget<FileEditor>>,
    protected parser: IMarkdownParser | null
  ) {
    super(tracker);
  }
```

packages/cells/src/widget.ts
```diff
@@ -324,6 +324,9 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
     return new Array<Cell.IHeading>();
```
krassowski, line 324: Do we want to deprecate the old getters?

packages/cells/src/widget.ts
```diff
@@ -324,6 +324,9 @@ export class Cell<T extends ICellModel = ICellModel> extends Widget {
...
   }
 
+  public async getHeadings(): Promise<Cell.IHeading[]> {
```
krassowski, line 327: Let's add a docstring explaining that this is preferred over `get heading`.

packages/fileeditor/src/toc/markdown.ts
```diff
@@ -18,6 +20,16 @@ export class MarkdownTableOfContentsModel extends TableOfContentsModel<
   IEditorHeading,
   IDocumentWidget<FileEditor, DocumentRegistry.IModel>
 > {
+  protected parser: IMarkdownParser | null;
+
+  constructor(
+    widget: IDocumentWidget<FileEditor>,
+    parser: IMarkdownParser | null,
+    configuration?: TableOfContents.IConfig | undefined
```
krassowski, line 28: Can you move `parser` to the end of the arguments list so that we don't break compatibility with downstream code which used `new MarkdownTableOfContentsModel(widget, configuration)`?

packages/toc/src/utils/markdown.ts
```diff
@@ -62,88 +62,41 @@ export async function getHeadingId(
...
-      let frontmatterEndLineIdx = lineIdx + 1;
-      frontmatterEndLineIdx < lines.length;
-      frontmatterEndLineIdx++
-    ) {
-      if (lines[frontmatterEndLineIdx] === '---') {
-        lineIdx = frontmatterEndLineIdx + 1;
-        break;
-      }
-    }
+export async function getHeadings(
```
krassowski, line 65: Can you rename this to `parseHeadings` and leave the old `getHeadings` as-is for backward compatibility, but mark the old `getHeadings` as deprecated in favour of async `parseHeadings`?

krassowski (commented):
This looks good to me, thank you @itsmevichu! Just a few suggestions to allow releasing this in a minor version by making sure we are not introducing any breaking API changes.

jtpio:
Many thanks @itsmevichu for working on this!

This seems to indeed be fixing many issues related to the table of contents.

For example this:

```
<h1 style="color: orange;">Test title, <code style="background-color: transparent;">foo</code> and <code style="background-color: transparent;">bar</code></h1>
```

Doesn't render correctly in 4.4.3:

image


And renders fine with this PR 🎉 

image



Wondering if there is a way to have this PR be a bug fix instead of targeting 4.5.0?

packages/fileeditor-extension/src/index.ts
```diff
@@ -89,7 +90,8 @@ const plugin: JupyterFrontEndPlugin<IEditorTracker> = {
     IEditorLanguageRegistry,
     IEditorThemeRegistry,
     IDefaultFileBrowser,
-    ISettingRegistry
+    ISettingRegistry,
+    IMarkdownParser
```
krassowski: This should be in `optional`, not in `requires`.

krassowski: This is the cause of the remaining failures on CI I believe (hopefully only this):

```
Parsed an error from text content: undefined TypeError: translator.load is not a function
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:plugin' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/debugger-extension:files' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:cursor-position' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:completer' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:language-server' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:editor-syntax-status' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/fileeditor-extension:tab-space-status' failed to activate.
Parsed an error from text content: undefined Plugin '@jupyterlab/tooltip-extension:files' failed to activate.
```

This is because `requires` + `optional` get concatenated and the order of arguments in `activate` must match the order in the concatenated arrays.

I will push a fix.

krassowski: Resolved in 2f120993dd9be3f23b7651db5ef4c0e8be3a7590

krassowski:
The visual test failures appear related. All `Collapsible Headings` tests are either failing or flaky, and `Table of Contents › Notebook context menu` seems to be failing with the required heading not showing up.

krassowski:
So I think the issue is that `maybeCreateCollapseButton` gets called when `headings` is still an empty array (since it now gets populated async). We need some kind of a signal which fires when headings get asynchronously updated

https://github.com/jupyterlab/jupyterlab/blob/2f120993dd9be3f23b7651db5ef4c0e8be3a7590/packages/cells/src/widget.ts#L2368-L2369

https://github.com/jupyterlab/jupyterlab/blob/2f120993dd9be3f23b7651db5ef4c0e8be3a7590/packages/cells/src/widget.ts#L2196-L2198

itsmevichu:
Instead of adding a new signal, I just waited on `cell.getHeadings().then(...)` before rendering the collapse logic. That way, it only shows up once the headings are actually resolved.

Also in the UI visualization tests, I've added a wait to allow the collapse behavior to finish before assertions.

krassowski:
Thank you for your work on it, great progress!

It looks like two more tests are failing consistently:

```
    [jupyterlab] › test/jupyterlab/notebook-edit.test.ts:123:7 › Notebook Edit › Move cells up ─────
    [jupyterlab] › test/jupyterlab/toc.test.ts:64:7 › Table of Contents › Notebook context menu ────
```

In `Move cells up` the collapse triangle does not show up in time, I guess this could be solved with a small timeout, in `Notebook context menu` the `HTML title` is not visible in the ToC and does not show up within 1 minute which seems more serious.

itsmevichu:
Thanks! I had already addressed those issues but forgot to push the changes 😅.

krassowski:
Hmm.. It feels like there is a distinct lag about one second until the ToC gets updated which is why there is ~1s timeout required in tests. It looks like it comes from this place:

https://github.com/jupyterlab/jupyterlab/blob/22562319f14b87e46df9d984f47bf587593f52f3/packages/toc/src/factory.ts#L11-L70

I wonder, did refactor to async made it slower, or was it slow all along?

We can do that separately but I would probably vote to decrease the timeout to maybe 200ms or so.

itsmevichu:
Async is actually causing a bit of delay, which is noticeable in the heading collapse logic, but it’s not really visible in the ToC because the timeout there was already set to 1000ms.
After changing the timeout to 200ms, I noticed a good improvement. The headings get resolved quicker, so the collapse logic works smoothly now since it will fetch from the cache.

So, I’ve now removed the timeouts from the tests;

itsmevichu:

Issue #16890, opened 2024-10-23
[a11y][Violation][4.1.2 Name, Role, Value] Accessibility issues reported on JupyterLab UI using IBM Equal Access Accessibility Checker tool


nkn2022:
## Description

I am running an accessibility test suite with Jupyter using the **IBM Equal Access Accessibility Checker** tool on Chrome browser. The results look good overall, only a few things are missing to reach a first formal level of compliance. So kindly help look into these. 

Issues reported were violating **WCAG 2.2** requirements. 
Ref: https://www.ibm.com/able/requirements/checker-rule-sets

Listing of violations categorized based on the checkpoints. 
Full report (includes information on element locations where issue were found): 
[Accessibility_Report-Untitled.ipynb - JupyterLab.xlsx](https://github.com/user-attachments/files/17383160/Accessibility_Report-Untitled.ipynb.-.JupyterLab.xlsx)

# Category: `4.1.2 Name, Role, Value`
Ref: https://www.ibm.com/able/requirements/requirements/?version=v7_3#4_1_2

****Violation of Rule:** `aria_attribute_required`**

> List of issues with elements that violate this rule with a screenshots
> > **Issue 1** 
> > Element with 'switch' role does not have the required ARIA attribute(s): 'aria-checked'
> > **Element location:** 
> > `<button title="Simple Interface (⇧ ⌘ D)" role="switch" class="jp-switch">`
> > **Screenshot:**
> > <img width="468" alt="image" src="https://github.com/user-attachments/assets/0d33bd73-a40e-4f24-8b47-b7b2f64ce7fc">

****Violation of Rule:** `element_tabbable_role_valid`**

> List of issues with elements that violate this rule with a screenshots
> > **Issue 1** 
> > The tabbable element's role 'none' is not a widget role
> > **Element location:** 
> > `<jp-button minimal="" appearance="stealth" current-value="" tabindex="0" title="New Launcher (⇧ ⌘ L)" data-command="launcher:create" aria-label="New Launcher (⇧ ⌘ L)" aria-disabled="false" class="jp-ToolbarButtonComponent">`
> > **Screenshot:**
> > <img width="468" alt="image" src="https://github.com/user-attachments/assets/5d8350aa-b485-4980-9820-34c7204eb9ab">

> > **Issue 2**
> > The tabbable element's role 'none' is not a widget role
> > **Element location:** 
> > `<span tabindex="0" class="jp-DirListing-itemText">`
> > **Screenshot:**
[...]

PR #17591, opened 2025-05-23
Fixes for accessibility issues reported on missing ARIA roles

nkn2022:
## References
[[a11y][Violation][4.1.2 Name, Role, Value] Missing ARIA roles and attributes for some JupyterLab elements](https://github.com/jupyterlab/jupyterlab/issues/16890) 
[[a11y][Violation][4.1.2 Name, Role, Value] Some Notebook UI elements missing roles and labels](https://github.com/jupyter/notebook/issues/7488)

## Code changes

- Add missing roles for multiple components
- Add missing aria-checked attribute to the switch role component

## User-facing changes

Users using screen readers will see that the roles and labels defined on these components.

The issues fixed are the following (See original [issue](https://github.com/jupyterlab/jupyterlab/issues/16890) description for details on these issues)
****Violation of Rule:** `aria_attribute_required`**
> > Element with 'switch' role does not have the required ARIA attribute(s): 'aria-checked'
> > **Issue 1** 

****Violation of Rule:** `element_tabbable_role_valid`**
> > The tabbable element does not have a valid widget role
> > The tabbable element's role 'none' is not a widget role
> > **Issue 2**
> > **Issue 6**
> > **Issue 7**
> > **Issue 8** 

NA


packages/ui-components/src/components/toolbar.tsx
```diff
@@ -861,6 +861,7 @@ export function ToolbarButtonComponent(
 
   return (
     <Button
+      role="button"
```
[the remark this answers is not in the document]
krassowski: So it looks like before and after two nested elements had the `aria-label="Refresh List"` but before only one of them had `button` role; now both will have a button role. It feels to me like one should be hidden away from aria altogether, am I correct?

| Before | After |
|--|--|
| image | image |

packages/codemirror/src/editor.ts
```diff
@@ -800,6 +800,13 @@ namespace Private {
       }),
       parent: host
     });
+    if (
+      view.contentDOM.getAttribute('role') === 'textbox' &&
+      view.contentDOM.getAttribute('contenteditable') === 'true'
+    ) {
+      view.contentDOM.setAttribute('aria-label', 'Enter code here');
+      view.contentDOM.setAttribute('tabIndex', '0');
```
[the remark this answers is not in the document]
nkn2022: Thanks for the info. I see you answered my question in the issue. 
I was expecting the focus to happen on a `Tab`. After this I was able to use `Esc` to get out of this input field. So thought this would work. But this is not needed.
I see that  when the focus is on the cell area and if I press `Enter` the focus is moved to the input field. Is this documented somewhere?
I will remove the `tabIndex` setting from above code.

packages/cells/src/widget.ts
```diff
@@ -1091,6 +1091,7 @@ export class CodeCell extends Cell<ICodeCellModel> {
         ? trans.__('Code Cell Content')
         : trans.__('Code Cell Content with Output');
     this.node.setAttribute('aria-label', ariaLabel);
+    this.node.setAttribute('role', 'region');
```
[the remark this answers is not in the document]
krassowski: I don't think we should change page down/up behaviour here, that would be breaking for users.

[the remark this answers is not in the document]
nkn2022: I will revert this change for now and see if @tonyfast has any suggestions. Thanks!

packages/ui-components/src/components/toolbar.tsx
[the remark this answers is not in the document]
nkn2022: Yes, the actual `button` element is not accessible via keyboard. So do not get focused. 
`jp-button` is the focused element and the one that performs the action. So we need the `role` setting there. The test should be fixed to get the right element.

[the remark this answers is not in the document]
nkn2022: I have tried a fix for the test failure. I am not able to run the test on my local, guess I am missing some setting. I will wait and check on the PR build.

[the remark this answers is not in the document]
nkn2022: @krassowski From discussions in the last frontend meeting, I understand we want to remove the internal `button`. I think that would be a cleaner fix. If that button element is not accessible then switching it to a `span` element should suffice. Since this is a major change I assume you want to  handle this also as a separate PR. So I will remove this change from this PR.

nkn2022:
@krassowski I have removed some of the changes that we thought would be best to handle separately. 
Please review the rest of the changes. 
Thanks!

nkn2022:
@krassowski any comments/suggestions for this PR? Its been a while. wondering how to proceed.

krassowski:
Thanks for your patience. I plan to review this PR over the upcoming weekend. Feel free to ping me on Monday if you don't hear from me by then.

packages/codemirror/src/editor.ts
```diff
@@ -800,6 +800,12 @@ namespace Private {
       }),
       parent: host
     });
+    if (
+      view.contentDOM.getAttribute('role') === 'textbox' &&
+      view.contentDOM.getAttribute('contenteditable') === 'true'
+    ) {
+      view.contentDOM.setAttribute('aria-label', 'Enter code here');
+    }
```
krassowski: - this only runs when editor is created, so if the cell is later changed to read-only it may have a wrong label
- there is not translation wrapper so it will read in English for say French students

I wonder if this should be a dedicated CodeMirror extension. Maybe this can be tackled in a separate PR?

packages/codeeditor/src/lineCol.tsx
```diff
@@ -248,6 +248,7 @@ function LineColComponent(
   };
   return (
     <TextItem
+      role="button"
```
krassowski, line 251: Maybe also:

```suggestion
      role="button"
      aria-haspopup="true"
```

Since clicking on it triggers:

image

packages/filebrowser/src/listing.ts
```diff
@@ -1105,6 +1105,9 @@ export class DirListing extends Widget {
         // Must check if the name node is there because it gets replaced by the
         // edit node when editing the name of the file or directory.
         nameNode.tabIndex = i === this._focusIndex ? 0 : -1;
+        if (nameNode.tabIndex === 0 && !nameNode.role) {
+          nameNode.role = 'button';
+        }
```
krassowski: This role gets applied to the current button but then never removed. This means that a user who moved over half the of files with down arrow key will see half of the filebrowser nodes as buttons and the other as not buttons. If this is required, should it:

- instead be assigned in:

https://github.com/jupyterlab/jupyterlab/blob/f072715a3e289d1f69fa56d8fbec887d484da69d/packages/filebrowser/src/listing.ts#L3477-L3487

- or should it be removed when `nameNode.tabIndex` is not zero?

packages/fileeditor/src/syntaxstatus.tsx
```diff
@@ -46,7 +46,18 @@ namespace EditorSyntaxComponent {
 function EditorSyntaxComponent(
   props: EditorSyntaxComponent.IProps
 ): React.ReactElement<EditorSyntaxComponent.IProps> {
-  return <TextItem source={props.language} onClick={props.handleClick} />;
+  return (
+    <TextItem
+      role="button"
```
krassowski, line 51: I could not reach this by keyboard, does it need tabindex?:

```suggestion
      role="button"
      aria-haspopup="true"
      tabIndex={0}
```

Also, does it need `aria-haspopup="true"` as it opens a popup:

image

packages/fileeditor/src/tabspacestatus.tsx
```diff
@@ -49,7 +49,13 @@ function TabSpaceComponent(
       : trans.__('Tab Indent');
   return (
     <TextItem
+      role="button"
```
krassowski, line 52: Same as with others:

```suggestion
      role="button"
      aria-haspopup="true"
      tabIndex={0}
```

At this point I wonder if we should have a wrapper around `TextItem` say `PopupButtonItem` as this is a frequent use case. And ideally it should also have `aria-controls` pointing to the popup.

packages/codeeditor/src/lineCol.tsx
nkn2022: Good suggestion. I have committed this.

packages/fileeditor/src/tabspacestatus.tsx
nkn2022: @krassowski I do not see many items having popups to add a new wrapper class. We could just pass in the props as needed from the caller. 
When testing I noticed that the DOM order is reversed for the "right" section of the footer toolbar. So the tab focus will go in reverse order(shown in video attached). Also, I added `tabIndex` to all elements as they contain some data that could be announced by the screen readers. Do you have any thoughts on this? Should we only allow tabbing to elements that have an action? 

https://github.com/user-attachments/assets/76f6cb8c-6278-426d-a1d0-f347d37742a4

packages/fileeditor/src/syntaxstatus.tsx
nkn2022: How can I get this button enabled?

packages/filebrowser/src/listing.ts
nkn2022: Thanks for finding this. I see that this is happening. I will need to remove it when the `tabIndex` is not set to zero. This is only needed when the `tabIndex` is set to `0` so has to be set at the place we have now where this is set.

packages/codemirror/src/editor.ts
nkn2022: Yes, this is being set for all the editors. 
But I noticed that this issue is not being reported in the latest scans. So I will remove this code change from this PR.

packages/fileeditor/src/tabspacestatus.tsx
krassowski: Good point about ordering, we can discuss that in a new issue.

> Also, I added `tabIndex` to all elements as they contain some data that could be announced by the screen readers. Do you have any thoughts on this? Should we only allow tabbing to elements that have an action?

I think the conventional thing is to only apply it on interactive elements, but I am not sure here.

nkn2022: Reading these docs I understand we only need to allow focus on interactive elements. So will keep it so. 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Keyboard

packages/fileeditor/src/syntaxstatus.tsx
nkn2022: Updated the code to include tabIndex.

packages/statusbar/src/components/text.tsx
```diff
@@ -32,7 +32,11 @@ export function TextItem(
   const { title, source, className, ...rest } = props;
   return (
     <span
-      className={`jp-StatusBar-TextItem ${className}`}
```
nkn2022, line 35: The `className` is not always available and this was including `undefined` in the class name. So updated this to fix that.

nkn2022:
@krassowski I have updated the PR with all the changes we discussed above. Request for a re-review.

packages/codemirror/src/editor.ts
nkn2022: I have removed this code change.

krassowski (approved):
Thank you @nkn2022! This exposed an issue with tab order, I opened https://github.com/jupyterlab/jupyterlab/issues/17680 to track it.

CC @tonyfast as I promised to ping you once this is done.

krassowski:
@nkn2022 will merging this PR close one of the linked issues, or do both include additional work not fully addressed by this PR?

packages/ui-components/src/components/toolbar.tsx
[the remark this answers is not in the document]
nkn2022: @krassowski This needs an issue.

nkn2022:
@krassowski The `jp-button` role problem is something that affects a lot of components. So we need an issue for this. 
After that we can merge and close both issues.

krassowski:
Can you create a new issue for `jp-button`?

nkn2022:
> Can you create a new issue for `jp-button`?

Sure will do. Thanks for merging.

nkn2022:
@krassowski I realized that we haven't closed the original accessibility issue with this PR merge as the button ariaLabel issue reported there still persists. Could we work on the suggested `jp-button` changes using the same issue: https://github.com/jupyterlab/jupyterlab/issues/16890? 
The change is required to fix the accessibility issue so make sense to use the same. Let me know if you have a different thought. Also, would it be possible to assign someone to work on it soon?

PR #17596, opened 2025-05-27
Emit `selectionChanged` on model refresh

jtpio:
## References

Follow-up to https://github.com/jupyterlab/jupyterlab/pull/17563

Noticed in https://github.com/jupyter/notebook/pull/7656

## Code changes

Emit `selectionChanged` on model refresh

## User-facing changes

None

None

### DIFF
--- a/packages/filebrowser/src/listing.ts
+++ b/packages/filebrowser/src/listing.ts
@@ -715,9 +715,11 @@ export class DirListing extends Widget {
   /**
    * Clear the selected items.
    */
-  clearSelectedItems(): void {
+  clearSelectedItems(emit = true): void {
     this.selection = Object.create(null);
-    this._selectionChanged.emit();
+    if (emit) {
+      this._selectionChanged.emit();
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/filebrowser/src/listing.ts
```diff
@@ -2475,6 +2475,10 @@ export class DirListing extends Widget {
         this.selection[path] = true;
       }
     }
+    // Emit the signal once for all items that were selected.
+    if (existing.length > 0) {
+      this._selectionChanged.emit();
```
krassowski, line 2480: Should it check if `existing != this.selection`? If I read it right the change could be due to a file/folder being removed in a model update, but a lot of the time the code above would effectively be a no-op, right?

jtpio: I think the main issue is the `this.clearSelectedItems();` call above which emits the `selectionChanged`. So the signal is first emitted, the selection is then empty because `this.clearSelectedItems()` clears the selection, so the signal has to be emitted again after the selection is restored?

krassowski: Could a solution be adding a new argument and calling it like `clearSelectedItems(emit=false)`?

krassowski: Just checking back here.

jtpio: Right, that sounds like a fine approach.

I'll try to get back to this PR some time soon, unless you would like to pick it up @krassowski no problem.

jtpio: Just pushed a new commit to follow this idea.

